PR #4885, opened 2020-05-03
Fixes for clang warnings for Ubuntu 20.04

Pentarctagon:
The first commit changes two uses of `python` -> `python3`, since 20.04 doesn't come with the `python` link by default.

Clang fix for `-Wdeprecated-copy` adds a copy constructor due to it warning about having an assignment operator defined but no copy constructor defined is deprecated.

Clang fix `-Wstring-plus-int` silences a warning from the lua source code stating that `<string> + <int>` does nothing.  This sounds... incorrect, but the lua code isn't ours and it doesn't seem to be breaking anything.

Everything else is for `-Wrange-loop-construct` warnings.

### DIFF
--- a/src/addon/info.hpp
+++ b/src/addon/info.hpp
@@ -87,6 +87,8 @@ struct addon_info
 		this->read(cfg);
 	}
 
+	addon_info(const addon_info&) = default;
+
 	addon_info& operator=(const addon_info& o) {
 		if(this != &o) {
 			this->id = o.id;
--- a/src/reports.cpp
+++ b/src/reports.cpp
@@ -526,8 +526,7 @@ static config unit_advancement_options(const unit* u)
 {
 	if (!u) return config();
 	config res;
-	typedef std::pair<std::string, std::string> pair_string;
-	for (const pair_string &ps : u->advancement_icons()) {
+	for (const auto& ps : u->advancement_icons()) {
 		add_image(res, ps.first, ps.second);
 	}
 	return res;
@@ -873,8 +872,7 @@ static int attack_info(reports::context & rc, const attack_type &at, config &res
 			}
 		}
 
-		typedef std::pair<int, std::set<std::string>> resist_units;
-		for (const resist_units &resist : resistances) {
+		for (const auto& resist : resistances) {
 			int damage_with_resistance = round_damage(specials_damage, damage_multiplier * resist.first, damage_divisor);
 			tooltip << "<b>" << damage_with_resistance << "</b>  "
 				<< "<span color='" << attack_info_percent_color(resist.first-100).to_hex_string() << "'>"
--- a/po/CMakeLists.txt
+++ b/po/CMakeLists.txt
@@ -167,7 +167,7 @@ if(ENABLE_POT_UPDATE_TARGET)
 	add_custom_command(
 			# Same idea with a dummy file as in pot-update: we never make it, and run the command each time.
 			OUTPUT ${PROJECT_SOURCE_DIR}/po/po_stat.dummy
-			COMMAND python utils/po_stat.py
+			COMMAND python3 utils/po_stat.py
 			--update-cfg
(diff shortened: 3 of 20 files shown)


src/addon/info.hpp
```diff
@@ -87,6 +87,27 @@ struct addon_info
...
+		, tags(o.tags)
+		, locales(o.locales)
+		, core(o.core)
+		, depends(o.depends)
+		, feedback_url(o.feedback_url)
+		, updated(o.updated)
+		, created(o.created)
+		, local_only(o.local_only)
+	{}
+
```
gfgtdf, line 110: why not `= default` here instead

src/reports.cpp
```diff
@@ -874,7 +874,7 @@ static int attack_info(reports::context & rc, const attack_type &at, config &res
 		}
 
 		typedef std::pair<int, std::set<std::string>> resist_units;
-		for (const resist_units &resist : resistances) {
+		for (const resist_units resist : resistances) {
```
gfgtdf, line 877: resist_units is a quote nontrivial type to copy, you shouldn't copy it here, if the old code copies it we should figure out why that is instead of removing the `&`

gfgtdf: actualyl here is why i does a copy: you need `<const int, ` instead of `<int, ` in the line above, please add that const and add the `&` back.

src/addon/info.hpp
Pentarctagon: I didn't realize that was a thing I could do.  I'll change that.

src/reports.cpp
Pentarctagon: Updated as well.

And for reference, the full text of the warning was:
```
src/reports.cpp:877:28: error: loop variable 'resist' has type 'const resist_units &' (aka 'const pair<int, set<basic_string<char> > > &') but is initialized with type 'std::pair<const int, std::set<std::__cxx11::basic_string<char>, std::less<std::__cxx11::basic_string<char> >, std::allocator<std::__cxx11::basic_string<char> > > >' resulting in a copy [-Werror,-Wrange-loop-construct]
                for (const resist_units &resist : resistances) {
                                         ^
src/reports.cpp:877:8: note: use non-reference type 'resist_units' (aka 'pair<int, set<basic_string<char> > >') to keep the copy or type 'const std::pair<const int, std::set<std::__cxx11::basic_string<char>, std::less<std::__cxx11::basic_string<char> >, std::allocator<std::__cxx11::basic_string<char> > > > &' to prevent copying
                for (const resist_units &resist : resistances) {
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
So it was previously being copied.

src/reports.cpp
```diff
@@ -527,7 +527,7 @@ static config unit_advancement_options(const unit* u)
 	if (!u) return config();
 	config res;
 	typedef std::pair<std::string, std::string> pair_string;
-	for (const pair_string &ps : u->advancement_icons()) {
+	for (const pair_string ps : u->advancement_icons()) {
```
gfgtdf, line 530: same here, assinga  const in the line above woudl be better than removing the `&`

gfgtdf: (alternativeley you can remove the line above altogether and just wrtie `for (const auto& ps : ...`)

Pentarctagon: I'll use auto.  Using a typedef like that seems unnecessarily verbose.

gfgtdf:
I didnt look at the python part but the C++ part looks good to me now

AI0867:
There appear to be systems where python 3 is installed as `/usr/bin/python`, and no `python3` symlink is provided. While they're probably less used than ubuntu, I'm not sure if replacing the name like this is the best choice.

Pentarctagon:
Nearly all of Wesnoth's python scripts are already using the shebang of `#!/usr/bin/env python3` so they also won't/haven't been working on those systems for a couple years now at least.

PR #4899, opened 2020-05-10
Add 2p - Minervan Gate map and scenario

a-nop:
This map is [discussed in a forum thread](https://forums.wesnoth.org/viewtopic.php?f=15&t=52475).

I've created a scenario based on the other maps.  I added localizations for English only, assuming there is some automation to generate the rest?


a-nop:
Oops, looks like it failed a first interaction check.  See the forum thread for more interaction.

Pentarctagon:
Having two equals signs in `user_team_name==` is incorrect.  You also shouldn't be modifying the .po file directly like that.

Otherwise, you should first be making this into an add-on and focus on getting player feedback before proposing a map for mainline.

a-nop:
Thanks for the feedback!

I've fixed the typos, made the scenario strings explicit and removed the .po changes.

[The wiki suggests I should distribute a single map on the forums instead of as an add-on](https://wiki.wesnoth.org/Distributing_content) so I'm doing that for now, I will see about contributing it to some other add-on.

Pentarctagon:
I'll correct the wiki page then.  The forums are a good place to post map screenshots and get feedback, but making it available on the add-ons server is the best way to distribute any content.

ElderoZ:
"Add 2p - Minervan Gate map and scenario"

No.

soliton-:
With some exceptions (Isar's Cross for example) the multiplayer maps coming with the game should be well balanced. To prove that your map should be extensively tested before being considered for mainline.

Feel free to suggest your map again once that testing was done and preferably several skilled players agree that the map is balanced.

ElderoZ:
The maps that have been added to the mainline mappool have been played on extensively for years.

Issue #4877, opened 2020-04-27
Window incorrectly resized when dismissing units


GuidoBartoli:
### Game and System Information
Wesnoth 1.4.11 installed with flatpak on Linux Mint 19.3

### Describe the bug
When the recall dialog appear, if you have some units and start to dismiss them, when you dismiss the second to last one, the window is resized and the old one is kept in the background. The result are two super imposed windows with double borders caused by the different sizes. When the last unit is dismissed or the dialog closed, both windows disappear.

### To Reproduce

1. Load a game with some units available for recall
2. Open the recall dialog with ALT+R or with right click
3. Begin dismissing units until you remain with two units
4. Dismiss the one with the longest name
5. The window is redrawn smaller on top of the old one

### Expected behavior
The window should resize while the old one should disappear.

### Screenshots
Before:
![before](https://user-images.githubusercontent.com/16103676/80371541-5e7bdd00-8892-11ea-99ad-9efcab6bec9d.jpg)

After:
![after](https://user-images.githubusercontent.com/16103676/80371561-663b8180-8892-11ea-8388-2dd11f086cc3.jpg)

AI0867:
Looks like the window shrinks because `resilient` is shorter than `intelligent`, so less space is needed, but the map isn't redrawn after the window decides to shrink.

gfgtdf:
It's be better if the dialog  wouldn't resize at all here though.

GuidoBartoli:
Actually, it looks like a new smaller window is drawn on top of the previous one, so you see the border of the larger one underneath. It should simply use only one window and resize it if needed.

Pentarctagon:
Is there a way to tell a dialog to never resize itself?

Pentarctagon:
[An example save file](https://github.com/wesnoth/wesnoth/files/4618481/AToTB-The.Chase-Auto-Save1.gz):
* Dismiss Hann
* Dismiss Faren
* Dismiss Efran
* Dismiss Meowoc - the dialog now shows up as ![this](https://user-images.githubusercontent.com/3030250/81746342-edc8e900-946b-11ea-8161-e14871abaec2.png).

gfgtdf:
> Is there a way to tell a dialog to never resize itself?

I am not sure, to understand that one would first need to check why it rebuilds itself in the first place (in particular whether someone calls invalidate_layout or whtger maybe the Window is even fully recreated).

[...]

PR #4900, opened 2020-05-10
Fixed a bug with rezising in the recall unit window

RFeuerhack:
As described in issue #4877 sometimes the recall unit window would layer on top of another one after dismissing a unit. I noticed that only happens if you dismiss a unit that you aren't asked whether you are sure to dismiss said unit. by adding a else case and always asking if the user is sure to dismiss the unit I fixed that problem. I'm not sure if that is the best solution for this problem, but so far I haven't come up with a better fix.

### DIFF
--- a/src/gui/dialogs/unit_recall.cpp
+++ b/src/gui/dialogs/unit_recall.cpp
@@ -362,6 +362,8 @@ void unit_recall::dismiss_unit(window& window)
 		message << _("This unit is close to advancing a level.") << " " << (u.gender() == unit_race::MALE
 		         ? _("Do you really want to dismiss him?")
 		         : _("Do you really want to dismiss her?"));
+	}else {
+		message << _("Are you sure you want to dismiss this unit?");
 	}
 
 	if(!message.str().empty()) {


Pentarctagon:
I don't think that adding an extra popup for users to click through.  Ideally either the recall dialog wouldn't resize in the first place, or whatever causes the screen to redraw when closing the confirmation popup you've added should instead be done automatically.

gfgtdf:
> I noticed that only happens if you dismiss a unit that you aren't asked whether you are sure to dismiss said unit.

This could also just be a coincidence during your testing though. Is there any reason why adding this dialog should make it redraw the background map?

RFeuerhack:
I'm not sure what u mean by redrawing the background map, but I agree that it could be a coincidence  even though I tested enough in my opinion and adding said dialog fixed it in all cases for me.

Pentarctagon:
> Is there any reason why adding this dialog should make it redraw the background map?

In [window.cpp](https://github.com/wesnoth/wesnoth/blob/9fad4670a685317757d37198454d8d0e251782cb/src/gui/widgets/window.cpp#L595-L600) it looks like there's some logic about restoring what's been displayed, so that could be it.

RFeuerhack:
Thank you,  I will look into that.

Pentarctagon:
@RFeuerhack are you still looking into this?

Wedge009:
Doesn't look like it. Could be closed.

Wedge009:
Doesn't look like it. Could be closed.

Issue #4879, opened 2020-04-28
UtBS Dark Assassin declared an "Orc"


eyalroz:
Wesnoth 1.15.3
Under the Burning Suns
Multiple scenarios, but specifically - Out of the Frying Pan

The Dark Assassin's race is shown as "Orc" - before and after he reveals himself. That doesn't make sense. Before, it should be unknown, and afterwards he should be an elf. (By "afterwards" I mean during dialogue, he's no longer active as a unit).

doofus-01:
This should no longer be an issue, as of #4906

Issue #4875, opened 2020-04-26
Dark Assassin drawn with curved dagger, said to have a scythe


eyalroz:
Wesnoth 1.15.3
Scenario 4: Descending into Darkness

In the scenario, Kaleh is attack by a "Dark Assassin". First I should say that dressing him up like an Orcish Assassin is somewhat unbecoming, but - that's a matter of taste I suppose.

The issue is that the assassin is drawn to be holding a curved dagger with a sort of a curved hook at its bottom - certainly not a scythe. But the combat dialogue suggest he has a scythe. Make it a dagger please, if there isn't something more fitting.

Wedge009:
Perhaps a case of the unit attributes not being updated along with the sprite re-design. Compare the unit sprites between 1.10 and 1.12 (which is also used for 1.14):

https://units.wesnoth.org/1.10/mainline/en_US/Dark%20Assassin1.html
https://units.wesnoth.org/1.12/mainline/en_US/Dark%20Assassin1.html

eyalroz:
Actually, examining the sprite again, I note that:

* Combat dialog - scythe
* Large illustration - dagger
* Sprite - a blade which could be interpreted in more than one way

doofus-01:
> First I should say that dressing him up like an Orcish Assassin is somewhat unbecoming, but - that's a matter of taste I suppose.

I agree, the unit data is also in an `/orcs` directory, he's basically an orc copy.  I'll look into getting him a unique sprite.

doofus-01:
This should no longer be an issue, after #4906

PR #4906, opened 2020-05-18
UtBS - units - Replacing inconsistent Dark Assassin unit/sprite data

doofus-01:
#4875 and #4879 mention the inconsistency of the Dark Assassin, seems like a placeholder that was never updated.  This PR is to address that.

Just base sprite for now, but I plan to (minimally) animate the new unit and wire into the relevant scenarios.


doofus-01:
The animations would not cut it for a default era sprite, but I think they are good enough for a minor character here.

Issue #4909, opened 2020-05-21
Map not drawn on smaller resolutions


gfgtdf:
When i try to start any game on a rather small resolution (which i mostly do to workaround #4908) the map is not drawn also the white border in the minimap only shows a very flat box.

![image](https://user-images.githubusercontent.com/4893136/82510411-b6f96f80-9b0a-11ea-9456-eb3b52ff3c5c.png)

Pentarctagon:
This one I suspect may be caused by #4851.

gfgtdf:
doing `git revert 43de778c6aad87^..f3a472ab963a5f` seems to fix this issue aswell.

gwhitney:
Apologies to all -- investigating now -- please can you let me know what resolution this occurs at? And also what roughly your pixel pitch is?

gwhitney:
Just posted PR #4910 to address this. gfgtdf if you could test whether that PR resolves this symptom I'd appreciate it. I may need more info to get at #4908, which I recognize is the more fundamental issue for you, but figured I should fix what I could do quickly first. Thanks!

Pentarctagon:
Confirmed locally that I could reproduce the issue, and the related PR fixes the issue.

PR #4851, opened 2020-04-15
Allow theme resolutions based on an arbitrary base resolution.

gwhitney:
Prior to this change, the 1024x768 base resolution (found in the default theme)
  was hard-coded into theme.cpp; resolutions beyond that were always scaled as if they
  had been defined on a 1024x768 base resolution, even if the width and height specified
  in the theme file were different. Combined with the scaling arithmetic in theme.cpp,
  this assumption resulted in bugs such as panels disappearing whenever their width in the
  base resolution specified in the theme exceeded 1024 pixels or their height exceeded 768
  pixels. This change records the base resolution from the theme in the theme and all of
  its objects, so the scaling computation can be done correctly.

This change would resolve issue #4850 -- see also the prototype high res theme posted at http://forums.wesnoth.org/viewtopic.php?p=654226#p654226

PR #4910, opened 2020-05-21
Fix issue #4909.

gwhitney:
As Pentarctagon suggested, before PR #4851 the lower resolutions of the default theme were still calculating off of a base resolution of 1024x768. After PR #4851, they were calculating off of their specified resolutions. This is more sensible, but given the scaling code and the old theme specification, could cause the main map to disappear via a bug complementary to the one that PR #4851 fixed (that the main map was disappearing when the resolutions were *too high*). This PR tweaks the rescaling to handle these cases when the screen resolution is small, and tweaks the definition of the 800x600 resolution of the default theme to specify the main map at a reasonable size for that resolution (formerly it was specified as larger than the screen (!) which was only working due to the quirks of location and size scaling).

### DIFF
--- a/data/themes/_initial.cfg
+++ b/data/themes/_initial.cfg
@@ -234,7 +234,7 @@
     [change]
         id=minimap-panel
         image=themes/classic/minimap-800.png
-        rect="842,=+6,+181,+177"
+        rect="=,=+6,+181,+177"
     [/change]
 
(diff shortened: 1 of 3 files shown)


Pentarctagon:
I was able to reproduce the issue myself, and confirm this fixes it.  Thanks!

Issue #4908, opened 2020-05-20
text is very tiny


gfgtdf:
on my latest 1.15.3 build a lot of dialogs currently look like this:

![image](https://user-images.githubusercontent.com/4893136/82508719-3f294600-9b06-11ea-8f3a-51a5fb6e0267.png)

when actually, there are supposed to look like this: 

![image](https://user-images.githubusercontent.com/4893136/82508789-6e3fb780-9b06-11ea-8c80-0166629c5356.png)

This only happens when the window is large enough. when i make the window smaller this problem goes away

Pentarctagon:
If you revert the commits from #4860, does the issue go away?

gfgtdf:
running `git revert 43de778c6aad87^..f3a472ab963a5f` and compiling will take a while

Wedge009:
That was my first thought too, that it might be related to the recent scaling efforts - before reading the description I thought it was a duplicate of the 'too small at super-high resolutions' issue.

gfgtdf:
Yes it does away if i revert the commits from my last post.

Pentarctagon:
@gwhitney Can you take a look at this and #4909?  If not, or if a solution is likely to take weeks or more, then these will need to be reverted on master in the mean time.

@gfgtdf The specific monitor resolution and pixel pitch, as well as the resolutions you changed Wesnoth to in order to trigger these bugs, will be useful to know.

gwhitney:
Looking right now; and yes, the monitor/window resolutions and pixel pitch would be very helpful to know.

gwhitney:
[...]

PR #4860, opened 2020-04-20
Gui2 font scaling

gwhitney:
This is pull request is independent of, but related to, #4851. That one addresses the fonts and layout controlled by the theme. This one allows for, and implements, the ability to scale the text font size and layout of GUI2 visual elements according to the screen pixel pitch. Specifically, over 1200x900 pixel dimensions, text_font_size is computed proportional to the pixel pitch. Also certain dialogs are allowed to take up a certain fraction of the screen size, rather than be limited to a numerically fixed number of pixels high or wide. The primary mechanism for facilitating these scalings is to allow the values for the keys text_font_size, maximum_width, and maximum_height to be formulas rather than just numerical constants. In addition, the SDL pixel pitch is exposed in the variable screen_pitch_microns, within GUI2.

PR #4911, opened 2020-05-21
Work around issue #4908.

gwhitney:
PR #4860 scales fonts by the observed screen pitch in microns when the main window is 1200x900 or larger. However, it seems based on the issue report #4908 that in some cases, the reported pixel pitch is not correct, and the scaling was shrinking fonts intolerably. This PR ensures that the scaling will never decrease font sizes, although it does not address why or under what circumstances the screen_pitch_microns may be incorrect.

### DIFF
--- a/data/gui/widget/label_default.cfg
+++ b/data/gui/widget/label_default.cfg
@@ -97,7 +97,7 @@
 # dpi. It came out looking pretty decent on my 90-micron monitor.
 
 #define _GUI_SCALE_RES SIZE
-	"(floor({SIZE} * 265 * 2 / (3 * screen_pitch_microns)))"
+	"(max({SIZE}, floor({SIZE} * 265 * 2 / (3 * screen_pitch_microns))))"
 #enddef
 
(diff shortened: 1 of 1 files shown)


soliton-:
Thanks for the quick response!

Pentarctagon:
Window resolution and screen pitch wouldn't be a bad thing to have included in the information we [tell people to provide](https://forums.wesnoth.org/viewtopic.php?f=4&t=25660) when reporting a bug(point 2).  See `generate_plain_text_report()` in `game_version_dialog.cpp` specifically.  Screen pitch in particular isn't something I'm even sure how to get easily - I ended up just going back to Newegg to get the monitor product ID and then googled the monitor specs.

gwhitney:
Hmm, sorry for being dense, but I don't know how to get to the game_version_dialog in the game to see what it does. Let me know, and let me know if you want a PR to add the reported dpi scale factors to the info it provides, if that was the drift of your last comment and if that's sensible.

Pentarctagon:
On the main menu's lower left corner, there's a button that pops up a dialog with the build information.  On the lower left corner of that dialog, there's a button with the hover "Copy the full report to clipboard".

And yeah, a PR would be good - definitely easier to have it automatically collected here than having to ask users to search and find it.

PR #4913, opened 2020-05-21
Fix divide by zero when SDL can't get DPI for scale factor.

fluffbeast:
SDL_GetDisplayDPI  returns zero if it can't determine the DPI scale factor.

This fix will fall back to a 1.0 scale factor instead of dividing by zero when this happens.


Pentarctagon:
Please put
```
int returncode;
returncode = SDL_GetDisplayDPI(window->get_display_index(), nullptr, &hdpi, &vdpi);
```
on one line as well - no reason to split that over two lines.

Pentarctagon:
Merged, thanks for the fix!

PR #4915, opened 2020-05-22
add weapon abilities in unit_type analysis for simulation

newfrenchy83:
This changes must be add abilities like UtBs formation to be added in unit_type simulation but not the 'aura' abilities(Httt initiative).

### DIFF
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -295,6 +295,7 @@ battle_context_unit_stats::battle_context_unit_stats(const unit_type* u_type,
 	// Get the weapon characteristics as appropriate.
 	auto ctx = weapon->specials_context(*u_type, map_location::null_location(), attacking);
 	boost::optional<decltype(ctx)> opp_ctx;
+
 	if(opp_weapon) {
 		opp_ctx.emplace(opp_weapon->specials_context(*opp_type, map_location::null_location(), !attacking));
 	}
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -319,7 +319,7 @@ battle_context_unit_stats::battle_context_unit_stats(const unit_type* u_type,
 	if(plagues) {
 		plague_type = (*plague_specials.front().ability_cfg)["type"].str();
 		if(plague_type.empty()) {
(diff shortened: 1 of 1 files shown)


soliton-:
Is there a reason why this is not relevant anymore?

newfrenchy83:
@soliton- i fear what apply_to=oponent cause a crash with that.

soliton-:
That's very little to go by. Does that mean there is a bug in the ability handling code? Surely this does not only affect the damage calculation that was being modified here. Is there a bug report about this crash?

newfrenchy83:
@soliton-  in fact, i think what this bugs was caused by a bug fixed by https://github.com/wesnoth/wesnoth/commit/3b9325647d9de6122b186b7728405cde5ef87fa9, i retry and all fine, perhaps i will reopen this PR

newfrenchy83:
@soliton- i play some and tested abilities used like special with this change, but none bugs. i will reopen this PR but i frustrated to don't find the cause of this crashes who is genertaed one time only.

gfgtdf:
does this actually work? I'd expect that (since if there is no unit ser in the attack context setter) that get_special_bool returns exactly the same values as bool_ability.

newfrenchy83:
> 
> 
> does this actually work? I'd expect that (since if there is no unit ser in the attack context setter) that get_special_bool returns exactly the same values as bool_ability.

I expect but how know this, i don't see this.

gfgtdf:
> Did you test that this actually changes anything?

newfrenchy83:
I don't see how testing that, alas?

soliton-:
Make up a test case where the ability will produce a significant difference between two attack targets or so. Like without considering the ability there is a 100% hit chance and with it is really 0%. Then you should see a difference in AI behaviour with and without this PR.

newfrenchy83:
all that what i know on this function came to https://github.com/wesnoth/wesnoth/commit/fb4361d666b8b0e169544217e33f4287a747719c. i don't know how see diffrences in game.

gfgtdf:
I think its also conlear what you mean by "must be add abilities" in the commit message

PR #4916, opened 2020-05-23
Add current video settings to the game_config::full_build_report.

gwhitney:
As requested by Pentarctagon in #4911.


Pentarctagon:
Thanks for adding this!

An example of the additional output for anyone interested when run from ingame:
```
Current video settings
======================

Current pixel resolution: 1920x1055
Refresh rate: 144
SDL reports: 91.8418x91.7458 dots per inch.
```

And then when run from `--report` as expected:
```
Current video settings
======================

No video initialized.
```

PR #4918, opened 2020-05-24
Scale buttons and toggle buttons with screen pitch also.

gwhitney:
This is a follow-on to PR #4860 -- it extends the scaling initiated in that PR to most of the buttons and toggle buttons in the game. Somewhat mysteriously, the "Menu," "Actions,", and "End Turn" buttons in the main gameplay window are still not scaling. Hopefully that can be tracked down in the future.

### DIFF
--- a/data/gui/macros/_initial.cfg
+++ b/data/gui/macros/_initial.cfg
@@ -349,6 +349,28 @@
 	hint_image = "icons/action/zoomdefault_25.png~FL(horiz)"
 #enddef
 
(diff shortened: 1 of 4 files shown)


gwhitney:
Note further that this a a .cfg-only commit; no source code is changed.

Pentarctagon:
I'm not seeing any errors, so I'll merge this in a couple days if no other comments.

Pentarctagon:
Thanks!

PR #4922, opened 2020-05-26
Append the timestamp to the save/replay file name.

Pentarctagon:


src/savegame.cpp
```diff
@@ -451,6 +433,11 @@ bool savegame::save_game(const std::string& filename)
 		if (filename_.empty())
 			filename_ = filename;
 
+		time_t t = std::time(nullptr);
+		tm tm = *std::localtime(&t);
+		auto time = std::put_time(&tm, "%Y%m%d-%H%M%S");
+		filename_ = (formatter() << filename_ << '_' << time).str();
```
soliton-, line 439: This appends the date each time this function is called regardless if the saved filename in the member variable is used or not. That does not seem like a good idea? Seems like the previous timestamp needs to be replaced.

soliton-:
In general I'm not sure how annoying it is not being able to specify (or know in advance) the complete filename of saves anymore. Perhaps when doing automated AI games where you want to analyze the replay afterwards? Surely something we can work around when/if such features are requested.

I would not remove the overwrite checks. Saving twice in a second does not sound too far fetched (though it doesn't sound useful either) but more importantly you can change the date/time on your device and produce same filenames that way as well. Certainly that is not very likely but I would not leave potentially overwriting a save to chance. Also the code is already there and doesn't seem particularly complex.

src/savegame.cpp
Pentarctagon: I tested re-saving the game/replay multiple times, and it's not an issue, since the file name gets re-created from scratch each time anyway.

Pentarctagon:
I'll re-add the checks then.

src/savegame.cpp
soliton-: What is the if empty check above for then? Dead code?

Pentarctagon: I believe so.  The two places where `save_game()` is called both set `filename_` to a value.  That check also doesn't account for both `filename_` and `filename` being empty, even though the `filename` argument defaults to being empty.

soliton-: Ok, just confusing code that should be cleaned up at some point then. (and maybe a quick history check what the intention was...)

PR #4925, opened 2020-05-29
attack code fixes

gfgtdf:
wml code that changes the number of strikes during an attack should now work properly.
In particular this fixes an segfault reported on the forum when attacks were removed during an event.

I have to think about it first, whether this is really a good idea, because it also changes when the number of strikes is calcualted.


Pentarctagon:
So this would fix #4927 I assume?

gfgtdf:
Yes but i though there was also a isse with dioscussion about whether we shodul reavaluate the number of strikes during an attack or not which this is alsoa  about but i cannot find that one

Pentarctagon:
@CelticMinstrel and-or @stevecotton could one of you take a look at this and see if it's something that can be merged?  Otherwise I'll close this PR as well.

CelticMinstrel:
At a glance I _think_ it looks fine for merging? It would probably need to be tested, at the very least.

PR #4926, opened 2020-05-29
don't hardcode plague contraint

gfgtdf:
as the title says.

There is no reason why we should have the 'plague doesn't work on villages' restruction encoded into the cpp engine

This is not really a priority of mine i justfound this weird when looking at the c++ code

### DIFF
--- a/data/campaigns/Secrets_of_the_Ancients/utils/zombie-utils.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/utils/zombie-utils.cfg
@@ -6,6 +6,11 @@
         name= _ "plague"
         description= _ "When a unit is killed by a Plague attack, that unit is replaced with a Walking Corpse on the same side as the unit with the Plague attack. This doesn’t work on Undead or units in villages."
         type=null # prevents normal corpse, and allows SotA code to handle it
+        [filter_opponent]
+            [filter_location]
+                terrain="*^V*"
+            [/filter_location]
+        [/filter_opponent]
     [/plague]
 #enddef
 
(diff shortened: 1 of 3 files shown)


ProditorMagnus:
I believe reasoning for restriction was to not need to handle village ownership.

It will need to be announced as API change.

gfgtdf:
> I believe reasoning for restriction was to not need to handle village ownership.

Oh hmm makes sense. will think about it


> It will need to be announced as API change.

Actuall im not sure here, see 1.14 code the https://github.com/wesnoth/wesnoth/blob/1.14/src/units/abilities.cpp#L1039 only checks for the id=plague key of the special not of the special not the tagname, for for umc abiities that usually ahvea  differnt id, this 'new' code shodul match the 1.14 code much better

doofus-01:
This is a pretty big change.  You fixed it for core and SotA, but this is a subtle screw to UMC not using the core plague macro.  Would it be difficult to make the village terrain status check a key, with default to current behavior?

gfgtdf:
> This is a pretty big change. You fixed it for core and SotA like I said before from looking at the 1.14 code I don't think it is

doofus-01:
Have you tested that?  I just tested in 1.14.11, and UMC plague-type ability automatically prevents spawning in a village.

gfgtdf:
actually i checked this again, i and was wrong neither 1.14 nor the code after this  commit  (considering the only the cpp changes) will make plague work on villages since it is additionally  checked in the attack code.

i still don't really like that way this and other abiities have special checks in abilities.cpp, will think about what to do.

Pentarctagon:
Is this something that should be done still, or would it make sense to just close this PR?

ProditorMagnus:
I think at this state this is not useful. If explicitly adding support for plague of villages, it should also give key to have village change/not change ownership because of plague.

I do not remember anyone who has asked to have plague which works on village.

CelticMinstrel:
I think the spirit of this change is good, but as others have said, it needs to be done in a backwards-compatible way

Pentarctagon:
Alright, closing then.

PR #4930, opened 2020-06-01
Fix [filter_weapon] that doesn't work in filter_opponent

newfrenchy83:
@gfgtdf without other_attack_ in special_active_impl, the opponent's attacks cannot be filtered and the use of filter_weapon filter_opponent always returns a null result.


gfgtdf:
oh yes of course, dunno why i made this stupid  mistake

PR #4933, opened 2020-06-02
Add the ability to output preprocessed WML as XML.

Pentarctagon:
This commit:
* Renames config's debug() method to as_text() to be more descriptive.
* Adds two as_xml methods to config.
* Also writes the XML out when using the --preprocess command line option.

### DIFF
--- a/src/actions/undo.cpp
+++ b/src/actions/undo.cpp
@@ -300,11 +300,11 @@ void undo_list::read(const config & cfg)
 			}
 		} catch (const bad_lexical_cast &) {
 			ERR_NG << "Error when parsing undo list from config: bad lexical cast." << std::endl;
-			ERR_NG << "config was: " << child.debug() << std::endl;
+			ERR_NG << "config was: " << child.as_text() << std::endl;
 			ERR_NG << "Skipping this undo action..." << std::endl;
 		} catch (const config::error& e) {
 			ERR_NG << "Error when parsing undo list from config: " << e.what() << std::endl;
-			ERR_NG << "config was: " << child.debug() << std::endl;
+			ERR_NG << "config was: " << child.as_text() << std::endl;
 			ERR_NG << "Skipping this undo action..." << std::endl;
 		}
 	}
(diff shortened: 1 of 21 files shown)


gfgtdf:
hmm i am not  against this, but i also don't see the usecase right now.

gfgtdf:
since i don't know the usecase the implementation looks fine, however it doesn't preserver translatable strings, since it just calls .str ()

Pentarctagon:
> hmm i am not against this, but i also don't see the usecase right now.

The idea would be the add JSON later, which would hopefully be equally easy, and then get rid of wmlparser.

> EDIT: i don't think making it print the xml by default with --preprocess is a good idea, better make it an extra flag

Sure.

> EDIT2: are lt and amp really the onylo 2 caracters in wxml tha nees escaping ?in wml

In attributes there are a few others, but in text it's just those two.  I validated the output with `xmllint`, which at least for the case I tried found no errors.

> since i don't know the usecase the implementation looks fine, however it doesn't preserver translatable strings, since it just calls .str () whether this is a problem and whats the best approach to fix it woudl be probably depends on the usegase.

I saw that as well, but I wasn't sure what (if anything) could be done about it. `val.second` was preserving the leading `_`, whereas both `str()` and `t_str()` aren't.

Pentarctagon:
Closing as using the `debug()` method as a starting point is not a correct approach.

gfgtdf:
> I saw that as well, but I wasn't sure what (if anything) could be done about it.

if we just think about simple atrnaltable string that are not composition of multiple trsalatable strings we could make store the textdomain as a attribute like `<name td="wesnoth-utbs">Kaleh</name>` to mark translatable strings (then the text would be the untranslated string), however like i said, this only works for simple translatable strings,


for more complicated case different syntaxes would be thinkable

AI0867:
It's doable, but you'd need to turn every attribute that can take a translatable string (which is either most or all of them) into an element instead of an attribute:

    [foo]
        #textdomain wes-foo
        text= _"foo" +
        #textdomain wes-bar
            _"bar"
    [/foo]

would become

    <foo>
        <text>
            <translatable domain="wes-foo">foo</translatable>
            <translatable domain="wes-bar">bar</translatable>
        </text>
    </foo>

PR #4937, opened 2020-06-06
UtBS - some minor polishing

doofus-01:
Minor map edits, unit placement edits, and polish for scenarios 4 and 5.


doofus-01:
I don't believe there is anything controversial here, so I'll merge it now.

PR #4938, opened 2020-06-06
Ancient Stone masonry

doofus-01:
I was going through UtBS campaign, it occurred to me some ancient/weathered stone floor and wall terrains could make a nice addition.

Terrain - "Ias" - stone floor variation
Also plan to put an ancient stone wall variation in here, but just the floor for now.

Screenshot_20200606


doofus-01:
First pass at the walls.  None of the north-facing images are done yet, but they are minor.

Screenshot_20200606_2

Issue #4945, opened 2020-06-10
A unit is broken in scenario three of Under the Burning Suns [spoilers]


123chess456:
### Game and System Information
Version 1.15.3, Windows, downloaded from Wesnoth.org

### Describe the bug
In Scenario 3 of Under the Burning Suns, Garak, a Quenoth Champion, is given the "righteous rage" ability on both of his attacks. The ability is described as basically being the same as berserk from a Dwarf Berserker, yet Garak's attacks function normally and do not "berserk".

### To Reproduce

1. Play Under the Burning Suns
2. Get to Scenario 3
3. Attack a unit with Garak
4. Garak will oftentimes leave the enemy still alive

### Expected behavior
Garak should either kill the enemy or die doing so.

soliton-:
`righteous rage` only works on the champions. Since presumably the description did not make that clear feel free to suggest an improved description.

PR #4947, opened 2020-06-10
Changes description of "righteous rage"

123chess456:
In Under the Burning Sun scenario 3, Garak is given the ability "righteous rage" on both of his attacks. This effectively makes those attacks behave the same as berserk when Garak is fighting an enemy "champion", i.e. two undead units named Grak and Zur. I believe this was unclear and raised #4945 because I didn't understand that this ability only applies to fights against Grak and Zur, so I've made this PR to hopefully make the description a little clearer.


Pentarctagon:
@nemaara Is this good to merge?

Issue #4949, opened 2020-06-14
UtBS - bad filtering in Sc.8


doofus-01:
### Game and System Information
 - What version of the game are you running?
1.15.3+dev
 - Where did you download it from? (Steam, Mac App Store, our website, built from source, etc.)
source
 - If you're using a custom build, what compiler did you use, and what commit did you build at?
8cffcb930f8
 - What OS are you running?
Linux
### Describe the bug
In UtBS Sc. 8 "Out of the Frying Pan", there is a point where moving one of two special units to a trigger point causes Nym to be the stored unit in `{CHECK_EXPLORER}`, and subsequent WML 

            [message]
                speaker=$explorer.id
                message= _ "Look, daylight! I think we finally made it out of the caves!"
            [/message]

            [find_path]
                [traveler]
                    id=Nym
                [/traveler]
                [destination]
                    x=$explorer.x
                    y=$explorer.y
                [/destination]
                allow_multiple_turns=yes
                variable=nym_path
            [/find_path]

            [if]
                [variable]
                    name=unit.id
                    not_equals=Nym
                [/variable]
                [and]
                    [variable]
                        name=nym_path.hexes
                        numerical_equals=0
                    [/variable]
                [/and]
                [then]
                    # Nym is blocked by deep water
                    [message]
                        speaker=Nym
                        message= _ "Kaleh? Zhul? I’m still stuck back here..."
                        image=portraits/nym_moody.png
                    [/message]

                    [endlevel]
                        result=defeat
                    [/endlevel]

                    [return]
                    [/return]
                [/then]
            [/if]

causes Nym to give both dialogs, then game over.  If `explorer` is not Nym (i.e. a regular unit, not Dust Devil or Flesh Golem), then all is well.

Changing the `unit.id` filter above to `explorer.id` seems to fix the issue.  I intend to fix this in an incoming PR, but I am logging it here, just in case that PR doesn't work out.

doofus-01:
I believe this is fixed now, from #4950

PR #4950, opened 2020-06-14
Some additions and work on UtBS

doofus-01:
includes the fix to #4949


doofus-01:
The dialog changes in Scenarios 8 and 9 are primarily to get rid of the merman constantly saying "my master".  Eloh also needs to be at least somewhat charismatic, not throwing childish tantrums.  Otherwise, I tried to keep a light touch on any string changes.  If that stuff gets reverted/rejected, hopefully those things can be re-addressed.

As for the commander portrait, I'm not sure if it's better to have that just be for Durstang, or for the unit type, since there is more than one commander in the campaign.  Ideally, someone will make a better portrait, and that can be for Durstang, who has a fair amount of dialog.

Issue #2180, opened 2017-11-07
Preprocessor: unable to traverse "up"


vgaming:
Preprocessor cannot traverse "up" with the `..` notation. It becomes impossible to use relative paths AND directories.

For example, consider the following directory layout:

```
.
├── a
│   └── a.cfg
├── b
│   └── b.cfg
├── _info.cfg
└── _main.cfg
```

Now, if you're inside `a.cfg`, it's currently impossible to reference `b.cfg` by a relative path.

Please add the `..` notation? I understand the security concern of course. But then again, we might just check that resulting address lives inside `~add-ons`. Therefore the scope of files to be accessed is the same as with the `~add-ons` notation.

Pentarctagon:
See #1033

vgaming:
@Pentarctagon interesting. Well, the linked PR is not accepted yet, and it's rather big.

Maybe a smaller one can be made such as those proposed in comments? Like just resolving the full path and checking whether it's inside ~add-ons?

GregoryLundberg:
The problem with chasing `../` is it's rife with security issues which have to be handled carefully. That is why the PR was not accepted.

We're pushing to 1.14 and, while we're not, strictly speaking, in a feature freeze, I'd suggest that something this major should wait until after 1.14 ships.

I've been through this before with other software. If you want it fast, you need to reserve a CVE or four to handle the mistakes.

So, patience. There's an open PR on it and it's not going to be closed .. just delayed.

vgaming:
@GregoryLundberg Argee, these kind of changes should be done with extreme caution. Clearly defined boundaries help a lot here though ("only allow accessing files inside ~add-ons dir"). Anyway, understood.

CelticMinstrel:
@vgaming My intention was to push #1033 (most likely with much further review) _after_ 1.14 has been released. So it's not going to be in 1.14, but it will be happening eventually.

vgaming:
@CelticMinstrel yeah, I understood. Gregory said the same thing basically (feature freeze). No problem.

Wedge009:
1.14 was long since released, now 1.16 is imminent...

PR #1033, opened 2017-05-07
Allow paths with ".." in macro inclusions

CelticMinstrel:
I have made this safe (I think) by resolving any ".." components manually while the path is still a relative path to the gamedata or userdata directory. If there are ".." components left over that can't be resolved (ie, at the beginning of the path), then it's an error. Otherwise, resolution then proceeds as normal, depending on whether it's a WML path (which is also used by Lua) or binary path.

I think this could use an audit to determine if it's truly safe. I did add some unit tests to make sure it would be safe, but I'm not sure if it covers every possible case.

With this it should be possible to make any addon not depend on the name of its folder.

PR #4953, opened 2020-06-15
[WIP] allow ./../ in wml file inclusion

gfgtdf:
to do that the Preprocessor code was refactored to use the new "wml_path" class which basically splits the path into two parts of which the second might be traversed upwards but the first part not, security is guaranteed exactly the same way as before since we still call the same get_wml_location function.

This in particular allows s me  to make  the WC2 code easier by since i can hopefully do something like `#ifhave  ./../../campaigns to check whether the code  is currently inside the mainline directory or not. `

### DIFF
--- a/data/multiplayer/scenarios/Wesbench_Shroud_Walk.cfg
+++ b/data/multiplayer/scenarios/Wesbench_Shroud_Walk.cfg
@@ -2,7 +2,7 @@
 
 #ifdef DEBUG_MODE
 [binary_path]
-    path=data/campaigns/Heir_To_The_Throne
+    path=campaigns/Heir_To_The_Throne
 [/binary_path]
 [multiplayer]
     id=wesbench_shroud_walk
--- a/src/binary_path.cpp
+++ b/src/binary_path.cpp
@@ -1,0 +1,206 @@
+/*
+   Copyright (C) 2020 by the Battle for Wesnoth Project https://www.wesnoth.org/
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY.
+
+   See the COPYING file for more details.
+*/
+
+/**
+ * @file
+ * File-IO
+ */
+#define GETTEXT_DOMAIN "wesnoth-lib"
+
+#include "filesystem.hpp"
+#include "binary_path.hpp"
+
+#include "config.hpp"
+#include "deprecation.hpp"
+#include "game_config.hpp"
+#include "game_version.hpp"
+#include "gettext.hpp"
+#include "log.hpp"
+
+#include <boost/algorithm/string.hpp>
+#include <boost/filesystem.hpp>
+#include <boost/filesystem/fstream.hpp>
+#include "game_config_view.hpp"
+
+
+static lg::log_domain log_filesystem("filesystem");
+#define DBG_FS LOG_STREAM(debug, log_filesystem)
+#define LOG_FS LOG_STREAM(info, log_filesystem)
+#define WRN_FS LOG_STREAM(warn, log_filesystem)
+#define ERR_FS LOG_STREAM(err, log_filesystem)
+
+namespace bfs = boost::filesystem;
+using boost::system::error_code;
+
+
+/**
+ *  The paths manager is responsible for recording the various paths
+ *  that binary files may be located at.
+ *  It should be passed a config object which holds binary path information.
+ *  This is in the format
+ *@verbatim
+ *    [binary_path]
+ *      path=<path>
+ *    [/binary_path]
+ *  Binaries will be searched for in [wesnoth-path]/data/<path>/images/
+ *@endverbatim
+ */
+namespace
+{
+std::set<wml_path> binary_paths;
+
+typedef std::map<std::string, std::vector<wml_path>> paths_map;
+paths_map binary_paths_cache;
+
+} // namespace
+
+static void init_binary_paths()
+{
+	if(binary_paths.empty()) {
+		binary_paths.insert(wml_path::from_absolute(game_config::path + "/"));
+	}
+}
+
+namespace filesystem {
+
+binary_paths_manager::binary_paths_manager()
+{
+}
+
+binary_paths_manager::binary_paths_manager(const game_config_view& cfg)
+{
+	set_paths(cfg);
+}
+
+binary_paths_manager::~binary_paths_manager()
+{
+	cleanup();
+}
+
+void binary_paths_manager::set_paths(const game_config_view& cfg)
+{
+	cleanup();
+	init_binary_paths();
+
+	for(const config& bp : cfg.child_range("binary_path")) {
+		std::string path = bp["path"].str();
+		if(path.find("..") != std::string::npos) {
+			ERR_FS << "Invalid binary path '" << path << "'\n";
+			continue;
+		}
+
+		if(!path.empty() && path.back() != '/') {
+			path += "/";
+		}
+		if(binary_paths.count(wml_path(path)) == 0) {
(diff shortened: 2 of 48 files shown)


gfgtdf:
This code works but it still needs to be cleaned up a bit.

src/wml_path.hpp
```diff
@@ -0,0 +1,69 @@
...
+	const std::string& safe_path() const { return path_; }
+	///
+	bool exists() const;
+private:
+	wml_path(utils::string_view base, utils::string_view path);
+	std::string base_;
+	std::string path_;
+};
+}
+using wml_path = filesystem::wml_path;
```
shikadiqueen, line 69: What's the point of declaring `wml_path` in the `filesystem` ns if there's a `using` directive following the declaration?

gfgtdf: Well at First I wanted it to be in the filesystem namespace but then I didn't want to have to write filesystem:: everywhere. Soo I I should probably decide what I want and change the file accordingly.

CelticMinstrel:
I haven't looked over the diff yet but… Does this also apply to Lua `require` and `dofile`?

gfgtdf:
> I haven't looked over the diff yet but… Does this also apply to Lua require and dofile?

dofile and require do already support ".." in 1.14

CelticMinstrel:
It would seem rather unfortunate for ".." to work differently in Lua than in macro inclusions though… maybe the Lua support should use this system too?

data/multiplayer/scenarios/Wesbench_Shroud_Walk.cfg
```diff
@@ -2,7 +2,7 @@
 
 #ifdef DEBUG_MODE
 [binary_path]
-    path=data/campaigns/Heir_To_The_Throne
+    path=campaigns/Heir_To_The_Throne
```
CelticMinstrel, line 5: Removing the `data/` prefix is a very suspicious change. I'm not fundamentally against it, but it just seems like something that would break every add-on's binary path... unless of course it only affects mainline paths? If it does affect add-on paths, is there compatibility code to ensure add-ons still using it will continue to function?

src/binary_path.cpp
```diff
@@ -0,0 +1,206 @@
...
+/**
+ *  The paths manager is responsible for recording the various paths
+ *  that binary files may be located at.
+ *  It should be passed a config object which holds binary path information.
+ *  This is in the format
+ *@verbatim
+ *    [binary_path]
+ *      path=<path>
+ *    [/binary_path]
+ *  Binaries will be searched for in [wesnoth-path]/data/<path>/images/
```
CelticMinstrel, line 55: This comment seems wrong; `images` should probably be replaced with `<kind>` or something similar, right?

src/binary_path.hpp
```diff
@@ -0,0 +1,100 @@
...
+/**
+ *  The paths manager is responsible for recording the various paths
+ *  that binary files may be located at.
+ *  It should be passed a config object which holds binary path information.
+ *  This is in the format
+ *@verbatim
+ *    [binary_path]
+ *      path=<path>
+ *    [/binary_path]
+ *  Binaries will be searched for in [wesnoth-path]/data/<path>/images/
```
CelticMinstrel, line 52: Same issue as the previous comment. I realize this was copy-pasted from elsewhere but you may as well fix it at the same time, right?

src/editor/map/map_context.cpp
```diff
@@ -318,7 +318,8 @@ void map_context::load_scenario(const game_config_view& game_config)
 	config scenario;
 
 	try {
-		read(scenario, *(preprocess_file(filename_)));
+		//todo: maybe make filename_ a wml_path ?
```
CelticMinstrel, line 321: This seems like a good idea? Are you planning to do it later in this pull request?

src/filesystem_common.cpp
```diff
@@ -142,7 +143,7 @@ std::string read_map(const std::string& name)
 	if(map_location.empty()) {
 		// If this is an add-on or campaign that's set the [binary_path] for its image directory,
 		// automatically check for a sibling maps directory.
-		map_location = get_binary_file_location("maps", name);
+		map_location = get_binary_file_location("maps", name).get_abolute_path();
```
CelticMinstrel, line 146: This is misspelled, it should be `get_absolute_path`.

src/serialization/preprocessor.cpp
```diff
@@ -740,32 +740,30 @@ bool operator!=(char lhs, preprocessor_data::token_desc::TOKEN_TYPE rhs)
...
+		for(const wml_path& fname : files_) {
+			std::size_t space_pos = fname.safe_path().rfind(' ');
 
-			if(cpos != std::string::npos && cpos >= symbol_index) {
-				std::stringstream ss;
-				ss << "Found filename containing whitespace: '" << filesystem::base_name(fname)
-				<< "' in included directory '" << name << "'.\nThe included symbol probably looks similar to '"
-				<< filesystem::directory_name(fname.substr(symbol_index)) << "'";
+			if(space_pos != std::string::npos) {
+				int schraeger_pos = fname.safe_path().rfind('/');
```
CelticMinstrel, line 758: Is "shraeger" the word for "slash" in your native language? I think it'd be better to just call it "slash"... although I guess it's fairly self-explanatory in this case so maybe it doesn't matter.

src/wesnoth.cpp
```diff
@@ -346,7 +346,8 @@ static int handle_validate_command(const std::string& file, abstract_validator&
 	}
 	std::cout << "Validating " << file << " against schema " << validator.name_ << std::endl;
 	lg::set_strict_severity(0);
-	filesystem::scoped_istream stream = preprocess_file(file, &defines_map);
+	//TODO: maybe also allow 'safe' paths here ?
```
CelticMinstrel, line 349: This is typically used to read the mainline _main.cfg (or maybe an add-on's _main.cfg), so if not allowing safe paths means that it produces different results from actually loading the data into the game, then it should definitely allow safe paths (probably by default).

CelticMinstrel (commented):
It's nice that you've taken the time to find a better way to do this than my older PR! I can see a few little issues but nothing major, although I haven't really tried to figure out how the thing works, so who knows if I missed something.

data/multiplayer/scenarios/Wesbench_Shroud_Walk.cfg
gfgtdf: I don't know for sure wheterh  added it but there should definitley a compatibility path

gfgtdf:
> It's nice that you've taken the time to find a better way to do this than my older PR! I can see a few little issues but nothing major, although I haven't really tried to figure out how the thing works, so who knows if I missed something.

This pr doesn't work complteley yet (it makes the game complteley unplayable and breaka lot of images), i think in particular the last commit the changes to  [binary_path] seem to break something.

CelticMinstrel:
Well, that's unfortunate… but hopefully it can be fixed?

gfgtdf:
> but hopefully it can be fixed?

Yes it can and even if not the fiorst part can be merged without the changes to [binary_path]

Pentarctagon:
@CelticMinstrel Any opinion on whether this PR should be left open or not?

CelticMinstrel:
Well… from what I remember and reading through my old comments, it seems like it looked promising but wasn't quite ready, so… I guess the only reason to leave it open would be as a sort of example or base if someone else comes along to implement the same thing later?

Pentarctagon:
I'm going to close it then - it doesn't need to be open for people to look at it later as an example.

Wedge009:
I think #1033 was closed in favour of this one, but then it sounded like celmin wanted to redo that one as well anyway.

CelticMinstrel:
What we should probably be doing is something similar to this: https://www.icculus.org/physfs/

That one explicitly forbids `".."` though, otherwise I'd recommend just directly using it…

Is there an issue tracking this?

Wedge009:
I think this is tracked in #2180.

PR #4884, opened 2020-05-03
Announcements and mariadbpp

Pentarctagon:
Moved from #4867 since it will be a bit until this can be merged.

PR #4954, opened 2020-06-16
Send scenario/era/modification source add-ons and addon versions to wesnothd.

Pentarctagon:
This adds the additional attributes under the multiplayer information sent to wesnothd:
* mp_scenario_addon_id
* mp_scenario_addon_version (host's version)
* mp_era_addon_id
* mp_era_addon_version (host's version)
* active_mod_addon_ids
* active_mod_versions (host's versions)

---

This doesn't contain the part to actually insert the data, since I don't want to create a conflict with #4884.

### DIFF
--- a/src/game_initialization/mp_game_utils.cpp
+++ b/src/game_initialization/mp_game_utils.cpp
@@ -80,6 +80,8 @@ config initial_level_config(saved_game& state)
 
 	config level = state.to_config();
 	add_multiplayer_classification(level.child_or_add("multiplayer"), state);
+	level.child("multiplayer")["mp_scenario_addon_id"] = level.child("scenario")["addon_id"].str("N/A");
+	level.child("multiplayer")["mp_scenario_addon_version"] = state.to_config().child("scenario")["addon_version"].str("N/A");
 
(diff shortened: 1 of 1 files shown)


gfgtdf:
What the intention here? just to store it into the database ?

Pentarctagon:
Yes.

gfgtdf:
I think it'd be nice if the mp server would  automaticially check whether all addons used are up to date, for that it'd need to ask the addon server probably. Maybe it'd be easiest if the mp server just fetched a list of all addon versions from the addon server every hour.

Pentarctagon:
That would additionally require:
* The client to tell wesnothd all add-ons the user has installed and their version on login, rather than just the ones that will be used in the current game.  That wouldn't entirely supersede this change, which also links the add-on+version to the particular scenario/era/modifications being used.
* Some way to check what the current version of the user's add-ons are.  IMO this would be best implemented by giving campaignd database support and on upload inserting/updating a row in a new database table that would have the add-on ID and version, which wesnothd could then simply query.

So while I do think it's something worth doing, it's also not something I'd work on implementing at least until after #4884 is hopefully able to be merged in early July.

PR #4963, opened 2020-06-22
WIP only load active add-ons fixes #4832

### LINKED ISSUE #4832: Disable unused addon during a game
So currently it is possible that global [trait], [units], [lua] etc tagshave an effect on the current gamestate even if they are dfeined from in addons that are not active n the current game. This casues OOS and i'd be better of this was not possibel.

I think the enginr should bahve as if these addons wouldn't exist during a game unless a part of that addon is explicitly used in the current game:

When we start a game the engine should gather a list of all addon_id that contain the used [campaign], [multiplayer], [era], [modification], [resource], and hide all content from other addons. 

Clearly all addons shodul be active during the loadingscrren/multiplayer_create etc screens, to show all installed campaigns/eras etc.

The difficultiers when implementing this include:
1) when the config is loaded for a game one cannot say at that point what addons are involced since for example [scenario]s might contain [load_resource]
2) _A lot_ of places use the global config object, so an approach of replacing the global config with a a set of configs for one for each  addons, might require some refactoring

gfgtdf:
With this, the engine will ignore all global tags defined by addons that are not used in the current game, The main purpose is to solve OOS. The addons that are uses in the current gam are the addons that defines one the the [campaign],[modification] etc. tags that are uses in the current game.

### DIFF
--- a/src/game_classification.cpp
+++ b/src/game_classification.cpp
@@ -18,6 +18,9 @@
 #include "log.hpp"
 #include "serialization/string_utils.hpp"
 #include "game_version.hpp"
+#include "game_config_manager.hpp"
+
+#include <list>
 
 static lg::log_domain log_engine("engine");
 #define ERR_NG LOG_STREAM(err, log_engine)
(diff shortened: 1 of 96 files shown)


src/game_classification.cpp
```diff
@@ -91,3 +98,65 @@ bool game_classification::is_normal_mp_game() const
...
+	}
+
+	if(!campaign.empty()) {
+		mods.emplace_back("campaign", campaign);
+	}
+	while(!mods.empty()) {
+
+		const modevents_entry& current = mods.front();
+		if(current.type == "resource") {
+			if(loaded_resources.insert(current.id).second) {
```
stevecotton, line 145: This seems buggy, if it's meant to do what I think it's meant to do. I think it's meant to skip the rest of the loop for things that have previously been inserted in to loaded_resources; but instead it skips rest of the loop *unless* the id is already in loaded_resources.

gfgtdf: hmm yes you are right

PR #4970, opened 2020-06-26
Add Earthy Rockbound Cave terrain

shikadiqueen:
Adds an earthy version of the Rockbound Cave terrain I made over half a decade ago for my add-on campaigns Invasion from the Unknown and After the Storm.

image

The colour shift I used doesn't 100% match the one eleazar used for the Earthy Cave Floor terrain, but I believe I may be one of the very few people who will ever notice the difference and personally it doesn't bother me. Adding this terrain variation is long overdue and the images can be changed later if necessary.

(See also the forum thread: <https://r.wesnoth.org/t52616>)


AI0867:
If this is just a color-shift, what's the rationale for adding new binaries over using IPFs?

Of course, given that these are years old, you probably no longer have the exact color shift lying around, but it may be worth thinking about this for future color-shifted terrain.

I don't see any color-issues in the screenshot above.

shikadiqueen:
You could ask the exact same question for all the other earthy cave terrain variations.

AI0867:
True. I'm asking it here because this is an active PR.

doofus-01:
I'll look into this more later, but for now two thoughts:

 1.  Adding IPF parameters to terrain macros would be a good thing, but let's leave that to a separate PR.

 2.  Including these binaries leaves the door open for upgrading them to something more unique later, even if that "later" spills over into stable releases.  Unique images are preferable to color-shifts, but no reason for that to be a blocker.  (And this applies to a few other terrains as well, not just Uhe.)

PR #4972, opened 2020-06-27
Scale fixed-size dialogue window capped by screen size percentage

### LINKED ISSUE #4969: Load Game dialog is abnormally large
Compared to version 1.14, the Load Game dialog takes up a disproportionate amount of space, almost filling up the entire screen. I would assume this change is not intentional since it goes opposite to previous (and future, such as issue #4817) efforts to limit line width across the board for easier reading and tidier layout.

image

The culprit is PR #4860 / commit f3a472ab963a5ff04b738ccb90de1c151039c1b6 specifically, as reverting it fixes the issue:

image

gwhitney:
Changes the GUI_WINDOW_PERC_FIXED_SIZED_CENTERED macro used by the load game dialogue to scale the nominal pixel size of the window by exactly the same scaling used for font sizes at large pixel dimensions, but (a) capped by a percentage of the screen size and (b) no smaller that it would have been under older code. Provision (b) is important to avoid the dialogue becoming any smaller on lower-pixel-dimension screen sizes where the label font scaling is not in effect.

Resolves #4969

### DIFF
--- a/data/gui/macros/_initial.cfg
+++ b/data/gui/macros/_initial.cfg
@@ -248,16 +248,6 @@
 	y = "(floor((screen_height - window_height) / 2))"
 #enddef
 
-#define GUI_WINDOW_PERC_FIXED_SIZE_CENTERED PERC_W PERC_H WIDTH HEIGHT
-	automatic_placement = false
-
-	width = "(max(screen_width * {PERC_W} / 100, min(screen_width, {WIDTH})))"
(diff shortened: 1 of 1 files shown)


shikadiqueen:
Works for me on my configuration.

```
Current video settings
======================

SDL video drivers:       [x11] wayland dummy
Window size:             1920x1080
Screen refresh rate:     60
Screen dots per inch:    92.538902x92.675674
Screen dpi scale factor: 0.963947x0.965372
```

PR #4974, opened 2020-06-28
UtBS and Quenoth rebalance

nemaara:
@Hejnewar I've started this PR, please feel free to put any comments you have. This is only a first draft, I haven't playtested it at all yet.


gfgtdf:
Im courious, whats the main intentions here? And what are your critery for balancung (in particular i'd assume that some untis will be stronger in some scenarios while others are stronger in others, so i wonder how you estimate which units need changes and which ones doesn't)

Hejnewar:
I will start with comments on L1s:
Let's start with the most powerful unit Quenoth have: Scout even without disengage this unit would be the best scout in mainline by fair margin, reason for that is not durability but damage. In comparison with elvish scout he can take on average similar amount of damage but can dishout significantly more damage in return. Disengage doesnt feel as strong in campaign, but still can make them very hard to kill. Additional -1/-1 melee/ranged will bring them closer to mainline level (ignoring their ability that makes the op almost always in MP.)

Tauroch. Reason of tauroch strength was his incredeble durability, high hp + resistances + dodge created quite a monster, without so high dodge this unit really isn't that great and also is slowest Quenoth unit (no quick, always 5 mp). Now why exactly durability made this unit so good (damage isnt really that exceptional there are mainline unit that have higher damage for similar cost)? At some point hex efficiency comes into play, and average units simply can't kill unit with that high durability and any amount of protection in one turn, thus making them incredibly safe units that just won't be dying much and will be dealing damage for longer period of time before they have to go and heal. I would most likely go with no dmg nerf, 30% flat /40% sand, 6mp amd just maybe -1 or 2 hp. 

Fighter. In case of this unit the more of them you have the better they are and that is exactly why they are not great in this campaign or at least in latter part of it. Iirc L1 unit prices are really high later on and thus you cant get many of them as support for higher level ones and when I can get Sun Sylphs investing a lot of xp into fighters feels just bad.
(For MP 1 fighter - bad, 2 fighters - ok, 3 fighters - very op, 4+ fighters - almost indestructible triangle wall formation.) 
Honestly I'm not quite sure what changes they need, on one hand they are too strong in numbers but its hard to get enough of them in this campaign to reach that opness. Will think more about them. 

Mystic is just weak. Daze is weaker than slow and even with slow she would still be weaker than shaman, but I still get the one by one just to level them up. I dont think that there is need to nerf her. 

L2 next I don't want to make this post too long and learning from past I will do things not all at once. ;)

nemaara:
No longer a PR just to adjust the units, I'm rebalancing the campaign here too.

nemaara:
@Hejnewar I've completed the first balancing run, please feel free to try it out now. I think it's still a tad easy, but should be much more of a challenge than before.

stevecotton:
I last played this campaign around 1.15.0, but thought the difficulty was OK as-is back then, and I was only playing on normal difficulty. The rebalance seems to be both nerfing the player's units and boosting the difficulty for all difficulty levels - I think at least easy and normal shouldn't be made more difficult.

nemaara:
Maybe you can provide some replays so I can gauge how well you did? I'm adjusting things based on Hejne's replays right now.

stevecotton:
The only map that I wrote notes on at the time was A Subterranean Struggle, https://r.wesnoth.org/p647258 . That's probably the most useful, as I replayed it multiple times.

Attached here are two other replays, both played from my first success (with the new Quenoth) on A Subterranean Struggle.

[Blood is thicker than water](https://github.com/wesnoth/wesnoth/files/4924043/UtBS-playthrough-Blut.ist.dicker.als.Wasser.Aufzeichnung.abspielen.gz): started with 372 gold (only 303 of which gets used at the start) and knowledge of this scenario's scripted events. The replay has sync errors. For the first stage, one less recall would probably have made no difference (the pathfinder Nisen), but the rest of the troops are needed. In the second stage, Grog was lucky to survive turn 17; and on turn 19 the single merman that I sent on the back route to the avatar lands all 3 hits, survives the enemy turn, and makes the kill on turn 20. I have troops ready in position for stage 3, and finish on turn 32, by which time the replay is badly out of sync.

[Zocthanol](https://github.com/wesnoth/wesnoth/files/4924026/UtBS-playthrough-Die.Schlacht.um.die.Insel.Zocthanol.Aufzeichnung.abspielen.gz): started with 379 gold and finished on 26/40, but the middle of the battle seems close to a loss around turn 15. I think 1 or 2 less recalls would have caused me to lose.

nemaara:
I pushed a commit that reduces the difficulty on easy and normal.

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
```diff
@@ -21,8 +21,9 @@
         id=Kaleh
         type=Quenoth Youth
         canrecruit=yes
-        gold=200
-        {INCOME 6 4 2}
+        {GOLD 220 180 140}
+        {INCOME 4 3 2}
+        village_gold=1
```
stevecotton, line 26: Please give the player a hint about the lowered value of villages.

nemaara: Hmm, I'm pretty sure that's unnecessary since the income is displayed at the top anyway.

gfgtdf: i agree with @stevecotton here, since the top bar does only displays the current income, not the per village income. But isnt 1 the default anyways ?

nemaara: The default is 2. I don't think it's necessary because the per village income isn't really vital for a player to know. If they really want to, they can look at scenario settings to see it, but the real stat that matters for strategy is your aggregate income from all sources.

At least to me, I think that making a special note about village income is the same thing as making a special note about base income (which is also unnecessary).

nemaara:
I'll merge this tomorrow if there are no further comments.

PR #4977, opened 2020-06-30
Remove the UMC IDE plugin.

Pentarctagon:
The reasons being:
* It's been unmaintained for several years now, so any bugs that might be reported against it will almost certainly not be fixed.
* Having been unmaintained for so long, even if someone wanted to create a custom Wesnoth IDE plugin, it's unlikely this would be a good foundation to continue with anymore.
* There are at this point better alternatives that people are actually using and maintaining (such as the VSCode plugin).
* Having this be present can give the impression that this IDE plugin is something UMC authors should still be using, which I think makes it actively harmful. Case in point, the current version of this available on SourceForge (2.0.1) was downloaded twice as of the current week; however the most recent version (2.0.2 in the changelog, or 2.0.3 based on some git commits) is not available on SourceForge at all, nobody apparently ever uploading it.


shikadiqueen:
I am okay with this (I sort of tried to get us to either drop it or find a new maintainer for it during the Wesmere website transition) but do note that there's also `eclipse.wesnoth.org` to deal with.

Pentarctagon:
Would that be over in the wesmere repository?

Pentarctagon:
Alright, I can dig through the wiki and try to find all the references there.  For wesnoth/website, would I just open a PR to delete the `eclipse.wesnoth.org` folder there?  I don't really know much about wesnoth/website.

shikadiqueen:
Well, no, deleting that is going to create a bunch of 404s all of a sudden and that's probably not ideal from a user's standpoint. Mainly what I'm trying to say is that *something* probably needs to be done about that in a transition phase before we delete eclipse.w.o entirely.

Pentarctagon:
Could eclipse.w.o be set to redirect to https://forums.wesnoth.org/viewtopic.php?f=21&t=13799 ?

shikadiqueen:
I believe that's going to break something with the plugin itself since it phones home to check for updates IIRC?

(Yes, I love how I basically don't know what we're dealing with here.)

Pentarctagon:
It looks like it'd be http://eclipse.wesnoth.org/updates, though if I try to go there in a browser it tells me I don't have permission to access that.  I also don't know if it's something it does automatically or if the plugin itself would break if it couldn't do that anymore (I'd assume not though, since otherwise it would be impossible to use without an internet connection).

Pentarctagon:
Actually it looks like it doesn't work now anyway.  The instructions say to use the URL to install the plugin, and it gives an error instead:
```
An error occurred while collecting items to be installed
session context was:(profile=epp.package.java, phase=org.eclipse.equinox.internal.p2.engine.phases.Collect, operand=, action=).
Artifact not found: http://eclipse.wesnoth.org/updates/plugins/org.eclipse.help.appserver_3.1.400.v20110425.jar.
http://eclipse.wesnoth.org/updates/plugins/org.eclipse.help.appserver_3.1.400.v20110425.jar
Artifact not found: http://eclipse.wesnoth.org/updates/plugins/org.mortbay.jetty.server_6.1.23.v201012071420.jar.
http://eclipse.wesnoth.org/updates/plugins/org.mortbay.jetty.server_6.1.23.v201012071420.jar
Artifact not found: http://eclipse.wesnoth.org/updates/plugins/org.mortbay.jetty.util_6.1.23.v201012071420.jar.
http://eclipse.wesnoth.org/updates/plugins/org.mortbay.jetty.util_6.1.23.v201012071420.jar
```

soliton-:
I agree that at this point it is actively harmful to make it look like the UMC IDE plugin is supported.

Perhaps @timotei can say something to this in general and how best to phase this out.

timotei:
@Pentarctagon These are the contents of that update URL: https://github.com/wesnoth/website/tree/master/eclipse.wesnoth.org/updates. Not having list permissions of that URL does not prevent eclipse from working properly, since it reads some specific paths.

The installation fails (I just tried it) because of some missing plugins that are part of the newer Eclipse versions. 

I would need to re-release a new version supporting them, but, given that nobody said anything for a long time (last reply in the related forum thread was in 2016), I think indeed it might be a good idea to remove it. I don't even know if anyone is still using the app - since he standalone version seems to start, but not sure if it's still compatible with the latest versions of Wesnoth (last "officially" supported one was 1.9)

I also think that alternatives such as VS Code or other extensible editors are a better fit than the (heavy) Eclipse platform.

W/r/t the website, we can just make it redirect to that forum thread from the index, and just keep the updates URL content as is to prevent it breaking for existing (if any?) users.

Looks good to me 👍 :shipit:

Pentarctagon:
@timotei Thanks!

@shikadiqueen I did a few searches of the wiki and fixed the couple direct links to eclipse.w.o that I found.  Once this is merged I'll also remove the wesnoth-umcplugin folder over on SourceForge.  So probably it'd be best to update wesnoth/website first, deploy that, create the redirect for eclipse.w.o, then merge this PR and remove the SourceForge files?

gfgtdf:
Well i don't know anything about this, since everyone including the original author agrees to deleting this then i also think it shoudl be deleted.

shikadiqueen:
@Pentarctagon I can set up the redirect immediately without having to wait for any PRs. But is it really necessary to delete the files from SF.net? We don't delete past versions of Wesnoth after a new one is released either — the whole thing is meant to be an archive of past releases.

Pentarctagon:
Mainly I was thinking that the previous versions of wesnoth make it rather obvious that they're older releases based on the version number - wesnoth 1.12 is obviously an older release than 1.14 - whereas the the "wesnoth-umcplugin" folder doesn't have that context.  So at a minimum I'd like to add "-unsupported" to the folder name, or maybe move it into the wesnoth-1.10 folder.

timotei:
@Pentarctagon I wouldn't delete the files either, unless really needed for various reasons, since the app itself still works fine (just the newer eclipse support is broken).

For the context, isn't what the suffix "umcplugin" provides?

Pentarctagon:
IMO it indicates what it is, but it doesn't indicate that it should no longer be used with current versions of wesnoth.

Pentarctagon:
I've added a [notice](https://sourceforge.net/projects/wesnoth/files/wesnoth-umcplugin/_IMPORTANT%20NOTICE_/) stating it's no longer supported.

@shikadiqueen Should I open a PR removing the non-update related files from wesnoth/website?

shikadiqueen:
I think I can just delete it myself once you decide to formally pull the plug, since there's stuff I need to tweak on the Apache side of things myself regardless.

So, to sum it up:

* The entirety of `eclipse.w.o` except for `/updates/` should become a redirect to `https://forums.wesnoth.org/viewtopic.php?f=21&t=13799`
* `/eclipse.wesnoth.org` should be deleted from the wesnoth/website repository, the `/updates/` directory left intact **on the live server only** for compatibility with previous installs for an unspecified period of time.

Is this correct?

Pentarctagon:
Yep, that's correct.

shikadiqueen:
Okay, went ahead and made the eclipse.w.o changes live.

PR #4979, opened 2020-07-01
Remove VC14 (VS2015/VS2017 project files).

Pentarctagon:
Reasons:
* These projectfiles are still 32-bit, whereas all other builds for all other OSes are now 64-bit.
* Relatedly, these projectfiles are dependent on the libraries kept at the aquileia/external repository.
* VS2019 is still listed as supporting Windows 7, so there is a minimal likelihood of developers not being able to use VS2019.
* The VS2019 projectfiles instead use vcpkg to get the required dependencies, which is easier to setup than the aquileia/external prebuilt libraries.
* It's one less thing that needs to be updated whenever source files are added/moved/removed.
* It's two fewer jobs that Travis needs to run, which means Travis builds will finish more quickly.


Wedge009:
I'm okay with this, but I thought some (a few - one?) were still using older VS projects. Even I was using the older version I generally keep my own set of project files anyway (based on the ones in the repository, of course).

gfgtdf:
unless we have  developer in out team that still used visual studio 2017 im fine with this.

Pentarctagon:
Alright, merged then (also I'm getting tired of the travis build failed notifications).

Issue #4981, opened 2020-07-02
add a function affect_distant to the abilities


newfrenchy83:
### Game and System Information
 - What version of the game are you running?master
 - Where did you download it from? (Steam, Mac App Store, our website, built from source, etc.)github
 - If you're using a custom build, what compiler did you use, and what commit did you build at?gcc8.1.0
 - What OS are you running?windows 7

### Describe the desired feature
I would like to add a to engine function so that the abilities can affect the remote units of the owner of that ability and not just the adjacent units. I will propose a request in relation to this.

### Screenshots
If applicable, add screenshots or mockups to help explain your request.

PR #4982, opened 2020-07-02
add a function [affect_distant] to the abilities

newfrenchy83:
resolve https://github.com/wesnoth/wesnoth/issues/4981

Since attempts to create remote abilities exist in the mainline campaign(DiD), I propose a hardcoded solution for devellopement branch( i doubt what this is accepted for 1.16, too late for that).


Pentarctagon:
gfgtdf - for the C++ part.
nemaara - to review that the DiD part still works as intended.

gfgtdf:
tbh i am a little worried about performance if we check every unit within in a range. also a maximum range of 10 seems to a be a little randomly chosen.

Furthermore from looking at the code i'm not convinced that it actually works.

And lastly your code must be refactored more, the fact that you added many codes that looks very similar to each and are probably copied with little modifications tells me that there is room for refactoring/improvement.

newfrenchy83:
For me 10 is great maximum range, but perhaps it is lowesr, or must be greatest?

I will work on refactored code also.

newfrenchy83:
@gfgtdf i pull an first simplification od the code, and can you said me what max_radius will be ade quate for you?

gfgtdf:
> and can you said me what max_radius will be ade quate for you?

i think it would be better if there was no max range, for that you'd probably have to cache which units have any abiltiy with [affect_distant] and then iterate over those units when you want to check the ability.

newfrenchy83:
> 
> 
> > and can you said me what max_radius will be ade quate for you?
> 
> i think it would be better if there was no max range, for that you'd probably have to cache which units have any abiltiy with [affect_distant] and then iterate over those units when you want to check the ability.

And how i do that?

nemaara:
Since I have no ability to test this patch on my own, I think it's better if you don't touch DiD's code and open a new PR later or let me do it later. Or, we can merge it in without testing, then I'll test it once it's in master and fix it as necessary.

gfgtdf:
> And how i do that?

well one approach would  be, you track the a unit abilities are changed,and whenever that happes you update a max_ability_radius attribute of the unit, when you check foir affect_distance you iterate over all in untis map and chack that attribute befroe ou do anything else.

newfrenchy83:
this approach seem more difficult to make, and i fear what i use more cpu. i think what i will close this PR and abandon.

gfgtdf:
:/ your choice  . maybe when you are wiser you can try again :p

newfrenchy83:
Not yet, i have an idea for resolve max_radiis value and and search in all map.

gfgtdf:
No you don't do `get_tiles_in_radius` for that,  you just iterate over `units`  (`for(const unit& u : units()) {  ... }`)

newfrenchy83:
@gfgtdf it is done, and thanks for your help.

PR #4984, opened 2020-07-05
Addon titles and descriptions made translatable

kabachuha:
**Addon titles and descriptions made translatable**

Hello, wesnoth devs!

The international wesnoth players had long been troubled by not being able to see what the addon is about without installing it or using external tools. So I decided to make it possible to see addon titles and descriptions in the addon manager.

I've done through `[translation]` tag modification because sending .mo file for the entire campaign is unnecesearily, and .mo and .po workflow just for .pbl file is redundant.

The feature is done on the `master` branch, but I'd like backporting it to 1.14 in after it's tested.

The original tag contains only the `language` attribute, the new one has the following structure. (If the declaration is abscent, but the .mo file is present, it'll be added with `supported=yes` attribute.):
```
[translation]
	description="localized description"
	language="ru" #locale code
	supported=yes #is set to yes if the corresponding .mo file has been found, otherwise no. If the client receives the tag without the `supported` attribute, it's assumed as `yes` for backward compatibility.
	title="localized title"
[/translation]
```

To upload the description translation insert it into the .pbl file.

<details>
  <summary>The example .pbl file</summary>
  
  ```
	author="kabachuha"
	title="Hello World"
	icon="units/monsters/gryphon.png"
	version="0.1.0"
	description="Checking the translation of the addon!"
	type="campaign"
	email="artemkhrapov2001@yandex.ru"
	passphrase="iloveasaltycucumber"
	translate=true

	[translation]
		language="ru"
		title="Здравствуй, мир!"
		description="Проверка перевода описания аддона"
	[/translation]

	[translation]
		language="zh_CN"
		title="你好，世界!"
		description="检查插件的翻译 (translated online)"
	[/translation]
  ```
  Notice: My testing addon also has a .mo file for the Japanese translation which isn't declared in this file. But you'll see it in the server database with `supported=yes`.
  
</details>

The deal with the `supported` attribute is to protect against an edge case when the translation is declared without the .mo file (as pointed out by Iris (@shikadiqueen) and against duplicate translations.

The server stores the addon info as usual except for the appended `translation` tags.

<details>
  <summary>The server.cfg from my testing addon server</summary>
  
  ```
	compress_level=6
	[campaigns]
		[campaign]
			author="kabachuha"
			description="Checking the translation of the addon!"
			downloads=1
			email="artemkhrapov2001@yandex.ru"
			filename="data/Hello_World"
			icon="units/monsters/gryphon.png"
			name="Hello_World"
			original_timestamp=1593893863
			passhash="hs3F0Q1wNL2b2xWTn5ODS1"
			passsalt="nfOLTKWsrjS/s6v2"
			size=776987
			timestamp=1593899194
			title="Hello World"
			translate=true
			type="campaign"
			upload_ip="127.0.0.1"
			uploads=7
			version="0.1.0"
			[translation]
				description="Проверка перевода описания аддона"
				language="ru"
				supported=yes
				title="Здравствуй, мир!"
			[/translation]
			[translation]
				description="检查插件的翻译 (translated online)"
				language="zh_CN"
				supported=no
				title="你好，世界!"
			[/translation]
			[translation]
				language="ja"
				supported=yes
			[/translation]
		[/campaign]
		[campaign]
			author="kabachuha"
			description="Addon with no translation for Russian!"
			downloads=0
			email="artemkhrapov2001@yandex.ru"
			filename="data/Status_Quo_Addon"
			icon="units/monsters/gryphon.png"
			name="Status_Quo_Addon"
			original_timestamp=1593894760
			passhash="E6D1VGjAb5bXqxK/KAWKH0"
			passsalt="oMGowk3uSHic9QY0"
			size=3656
			timestamp=1593894760
			title="Status Quo"
			translate=true
			type="campaign"
			upload_ip="127.0.0.1"
			uploads=1
			version="0.1.0"
			[translation]
				description="检查插件的翻译 (translated online)"
				language="zh_CN"
				supported=no
				title="你好，世界!"
			[/translation]
		[/campaign]
	[/campaigns]
  ```
  
</details>

The client side will check if the translation exists for the current locale and will use the translated title (the original title is shown slightly below the title for players who had known the addon by the English name) and the translated description in that case.

Also, I made possible to search both for the original and the translated strings in the manager.

***Demos***

<details>
  <summary>Pictures</summary>

The changes are fully backwards compatible with the old server/client.

backward_compatible

Addons on my pet server.

2020-07-05-005141_1920x1080_scrot

Chinese translation is rendered fine too.

2020-07-05-005233_1920x1080_scrot

Uploadable addons are also translated and the search is possible.

2020-07-05-004724_1920x1080_scrot

</details>

***Known issue***

The addon's name may not be fully shown in the dependecies window if it's very long. (According to Iris and Vultraz, its the fault of the gui even without the translation, so it belongs to another issue).

### DIFF
--- a/src/addon/info.cpp
+++ b/src/addon/info.cpp
@@ -117,10 +117,10 @@ void addon_info::write(config& cfg) const
 	cfg["uploads"] = this->uploads;
 	cfg["type"] = get_addon_type_string(this->type);
 
-	for(std::pair<std::string, addon_info_translation> element : this->info_translations) {
-		config* locale = &cfg.add_child("translation");
-		(*locale)["language"] = element.first;
-		element.second.write(*locale);
+	for(const auto& element : this->info_translations) {
+		config& locale = cfg.add_child("translation");
+		locale["language"] = element.first;
+		element.second.write(locale);
 	}
 
 	cfg["core"] = this->core;
@@ -158,9 +158,7 @@ addon_info_translation addon_info::translated_info() const
 	std::string locale = get_language().localename;
 
 	if(locale != "en_US") {
-		std::map<std::string, addon_info_translation>::const_iterator info;
-
-		info = info_translations.find(locale);
+		auto info = info_translations.find(locale);
 		if(info != info_translations.end()) {
 			return info->second;
 		}
--- a/src/addon/info.cpp
+++ b/src/addon/info.cpp
@@ -18,6 +18,7 @@
 #include "config.hpp"
 #include "font/pango/escape.hpp"
 #include "gettext.hpp"
+#include "language.hpp"
 #include "picture.hpp"
 #include "log.hpp"
 #include "serialization/string_utils.hpp"
@@ -57,6 +58,22 @@ namespace {
 	}
 }
 
+void addon_info_translation::read(const config& cfg)
+{
+	this->language = cfg["language"].str();
+	this->supported = cfg["supported"].to_bool(true);
+	this->title = cfg["title"].str();
+	this->description = cfg["description"].str();
+}
+
+void addon_info_translation::write(config& cfg) const
+{
+	cfg["language"] = this->language;
+	cfg["supported"] = this->supported;
+	cfg["title"] = this->title;
+	cfg["description"] = this->description;
+}
+
 void addon_info::read(const config& cfg)
 {
 	this->id = cfg["name"].str();
@@ -73,7 +90,9 @@ void addon_info::read(const config& cfg)
 	const config::const_child_itors& locales_as_configs = cfg.child_range("translation");
 
 	for(const config& locale : locales_as_configs) {
-		this->locales.push_back(locale["language"].str());
+		if(locale["supported"].to_bool(true))
+			this->locales.push_back(locale["language"].str());
+		this->info_translations.push_back(addon_info_translation(locale));
 	}
 
 	this->core = cfg["core"].str();
@@ -100,8 +119,8 @@ void addon_info::write(config& cfg) const
 	cfg["uploads"] = this->uploads;
 	cfg["type"] = get_addon_type_string(this->type);
 
-	for (const std::string& locale_id : this->locales) {
-		cfg.add_child("translation")["language"] = locale_id;
+	for(const addon_info_translation& locale : this->info_translations) {
+		locale.write(cfg.add_child("translation"));
 	}
 
 	cfg["core"] = this->core;
@@ -130,6 +149,68 @@ std::string addon_info::display_title() const
 	} else {
 		return font::escape_text(this->title);
 	}
+}
+
+addon_info_translation addon_info_translation::invalid = {"en_US", false, "invalid_addon!", ""};
+
+addon_info_translation addon_info::translated_info() const
+{
+	std::string locale = get_language().localename;
+
+	if(locale != "en_US") {
+		for(const addon_info_translation& info : this->info_translations) {
+			if(info.language == locale || info.language == locale.substr(0, 2)) {
+				return info;
+			}
+		}
+	}
+
+	return addon_info_translation::invalid;
+}
+
+std::string addon_info::display_title_translated() const
+{
+	addon_info_translation info = this->translated_info();
+
+	if(info.valid()) {
+		return info.title;
+	}
+
+	return "";
+}
+
+std::string addon_info::display_title_translated_or_original() const
+{
+	std::string title = display_title_translated();
+	return title.empty() ? display_title() : title;
+}
+
+std::string addon_info::description_translated() const
+{
+	addon_info_translation info = this->translated_info();
+
+	if(info.valid()) {
+		return info.description;
+	}
+
+	return this->description;
+}
+
+std::string addon_info::display_title_full() const
+{
+	std::string local_title = display_title_translated();
+	if(local_title.empty())
+		return display_title();
+	return local_title + " (" + display_title() + ")";
+}
+
+std::string addon_info::display_title_full_shift() const
+{
+	std::string local_title = display_title_translated();
+	if(local_title.empty())
+		return display_title();
+	return local_title + "\n"
+		+ "<small>(" + display_title() + ")</small>";
 }
 
 std::string addon_info::display_icon() const
--- a/src/addon/info.cpp
+++ b/src/addon/info.cpp
@@ -117,7 +117,7 @@ void addon_info::write(config& cfg) const
 	cfg["uploads"] = this->uploads;
 	cfg["type"] = get_addon_type_string(this->type);
 
-	for(const std::pair<std::string, addon_info_translation> &element : this->info_translations) {
+	for(std::pair<std::string, addon_info_translation> element : this->info_translations) {
 		config* locale = &cfg.add_child("translation");
 		(*locale)["language"] = element.first;
 		element.second.write(*locale);
--- a/src/addon/info.cpp
+++ b/src/addon/info.cpp
@@ -60,7 +60,6 @@ namespace {
 
 void addon_info_translation::read(const config& cfg)
 {
-	this->language = cfg["language"].str();
 	this->supported = cfg["supported"].to_bool(true);
 	this->title = cfg["title"].str();
 	this->description = cfg["description"].str();
@@ -68,7 +67,6 @@ void addon_info_translation::read(const config& cfg)
 
 void addon_info_translation::write(config& cfg) const
 {
-	cfg["language"] = this->language;
 	cfg["supported"] = this->supported;
 	cfg["title"] = this->title;
 	cfg["description"] = this->description;
@@ -91,8 +89,8 @@ void addon_info::read(const config& cfg)
 
 	for(const config& locale : locales_as_configs) {
 		if(locale["supported"].to_bool(true))
-			this->locales.push_back(locale["language"].str());
-		this->info_translations.push_back(addon_info_translation(locale));
+			this->locales.emplace_back(locale["language"].str());
+		this->info_translations.emplace(locale["language"].str(), addon_info_translation(locale));
 	}
 
 	this->core = cfg["core"].str();
@@ -119,8 +117,10 @@ void addon_info::write(config& cfg) const
 	cfg["uploads"] = this->uploads;
 	cfg["type"] = get_addon_type_string(this->type);
 
-	for(const addon_info_translation& locale : this->info_translations) {
-		locale.write(cfg.add_child("translation"));
+	for(const std::pair<std::string, addon_info_translation> &element : this->info_translations) {
+		config* locale = &cfg.add_child("translation");
+		(*locale)["language"] = element.first;
+		element.second.write(*locale);
 	}
 
 	cfg["core"] = this->core;
@@ -151,17 +151,23 @@ std::string addon_info::display_title() const
 	}
 }
 
-addon_info_translation addon_info_translation::invalid = {"en_US", false, "invalid_addon!", ""};
+addon_info_translation addon_info_translation::invalid = {false, "", ""};
 
 addon_info_translation addon_info::translated_info() const
 {
 	std::string locale = get_language().localename;
 
 	if(locale != "en_US") {
-		for(const addon_info_translation& info : this->info_translations) {
-			if(info.language == locale || info.language == locale.substr(0, 2)) {
-				return info;
-			}
+		std::map<std::string, addon_info_translation>::const_iterator info;
+
+		info = info_translations.find(locale);
+		if(info != info_translations.end()) {
+			return info->second;
+		}
+
+		info = info_translations.find(locale.substr(0, 2));
+		if(info != info_translations.end()) {
+			return info->second;
 		}
 	}
 
@@ -189,7 +195,7 @@ std::string addon_info::description_translated() const
 {
 	addon_info_translation info = this->translated_info();
 
-	if(info.valid()) {
+	if(info.valid() && !info.description.empty()) {
 		return info.description;
 	}
 
@@ -202,15 +208,6 @@ std::string addon_info::display_title_full() const
 	if(local_title.empty())
 		return display_title();
 	return local_title + " (" + display_title() + ")";
-}
-
-std::string addon_info::display_title_full_shift() const
-{
-	std::string local_title = display_title_translated();
-	if(local_title.empty())
-		return display_title();
-	return local_title + "\n"
-		+ "<small>(" + display_title() + ")</small>";
 }
 
 std::string addon_info::display_icon() const
--- a/src/addon/info.hpp
+++ b/src/addon/info.hpp
@@ -22,9 +22,66 @@
 #include <set>
 #include <map>
 
+struct addon_info_translation;
 struct addon_info;
 class config;
 typedef std::map<std::string, addon_info> addons_list;
+
+struct addon_info_translation
+{
+	static addon_info_translation invalid;
+
+	std::string language;
+	bool supported;
+	std::string title;
+	std::string description;
+
+	addon_info_translation()
+		: language()
+		, supported(true)
+		, title()
+		, description()
+	{}
+
+	addon_info_translation(std::string lang, bool sup, std::string titl, std::string desc)
+		: language(lang)
+		, supported(sup)
+		, title(titl)
+		, description(desc)
+	{
+	}
+
+	explicit addon_info_translation(const config& cfg)
+		: language()
+		, supported(true)
+		, title()
+		, description()
+	{
+		this->read(cfg);
+	}
+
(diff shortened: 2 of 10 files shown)


src/addon/info.hpp
```diff
@@ -22,10 +22,67 @@
...
+	}
+
+	void read(const config& cfg);
+
+	void write(config& cfg) const;
+
+	bool valid()
+	{
+		return title != "invalid_addon!";
+	}
```
shikadiqueen, line 83: There's technically nothing stopping me from titling my add-on `invalid_addon!` in either the original language or translations, is there?

src/addon/info.cpp
```diff
@@ -132,6 +151,68 @@ std::string addon_info::display_title() const
...
+}
+
+std::string addon_info::display_title_full_shift() const
+{
+	std::string local_title = display_title_translated();
+	if(local_title.empty())
+		return display_title();
+	return local_title + "\n"
+		+ "<small>(" + display_title() + ")</small>";
+}
```
shikadiqueen, line 214: I'm not entirely sure these methods belong here, at least not in the current form. Both revolve around GUI implementation details, especially `display_title_full_shift()`, whose purpose is not self-evident from the name and there are no Doxygen comments attached to it. The second method at least seems like it belongs somewhere deep in the GUI2 Add-on Manager dialog's implementation instead.

(I'm also not entirely sure if using a linebreak and Pango markup to achieve whatever the method is trying to achieve here *is* a good idea in the first place, that's more @Vultraz's thing I guess.)

src/addon/info.cpp
```diff
@@ -73,7 +90,9 @@ void addon_info::read(const config& cfg)
 	const config::const_child_itors& locales_as_configs = cfg.child_range("translation");
 
 	for(const config& locale : locales_as_configs) {
-		this->locales.push_back(locale["language"].str());
+		if(locale["supported"].to_bool(true))
+			this->locales.push_back(locale["language"].str());
+		this->info_translations.push_back(addon_info_translation(locale));
```
shikadiqueen, line 95: In general, `emplace_back` is preferred over `push_back` for containers that support it to avoid the extra copy, but this is not strictly an issue on my book.

src/addon/info.cpp
```diff
@@ -132,6 +151,68 @@ std::string addon_info::display_title() const
...
+
+addon_info_translation addon_info::translated_info() const
+{
+	std::string locale = get_language().localename;
+
+	if(locale != "en_US") {
+		for(const addon_info_translation& info : this->info_translations) {
+			if(info.language == locale || info.language == locale.substr(0, 2)) {
+				return info;
+			}
```
shikadiqueen, line 164: Is there a reason why this seems to accept `[translation]`s for language variants that may or may not be actually compatible (such as the different Chinese variants)? Also noting that the order in which they are found is not guaranteed AFAICT — if the .pbl file is preseeded with `[translation]` blocks then I figure the author can make it so any other variant is selected here instead of the language that the player is actually using, even if both are supported.

src/campaign_server/addon_utils.cpp
```diff
@@ -92,23 +92,43 @@ std::string format_addon_feedback_url(const std::string& format, const config& p
...
+		if(locale) {
+			locale["supported"] = true;
+		}
+		else {
+			addon.add_child("translation")["language"] = locale_id; //Doing this to circumvent a weird validation bug
+			addon.find_child("translation", "language", locale_id)["supported"] = true;
+		}
+	}
+}
+
```
shikadiqueen, line 108: A couple of style issues:
* `locale_id` should be a const reference here. The check for its emptiness makes up most of the function body instead of returning early if it passes.
* The `else` branch performs excessive WML scans. What's going on here is that you need to create the WML node in line 103 (`add_child()`), take its reference, and then do all further operations (adding the `language` and `supported` attribute values) with that reference.
* Based on that, the comment in there is also redundant. The validation bug you were hitting is the infamous "missing WML child yet unchecked for" error — if `find_child()` does not actually find the node you were looking for it returns a dummy WML node that can't be used in any meaningful way beyond checking its validity (the `bool` conversion). Since you already figured that out and are creating the child accordingly, you don't need to point it out.

src/campaign_server/campaign_server.cpp
```diff
@@ -893,9 +898,12 @@ void server::handle_upload(const server::request& req)
 		data["icon"] = (*campaign)["icon"];
 		data["type"] = (*campaign)["type"];
 		data["tags"] = (*campaign)["tags"];
-		(*campaign).clear_children("translation");
 		find_translations(data, *campaign);
 
+		for(const config& locale_params : (*campaign).child_range("translation")) {
+			data.add_child("translation", locale_params);//Do we need it?
+		}
+
```
shikadiqueen: To answer your question: no, we don't need it. There's no reason for most of the metadata to be copied into `data` (the on-disk archive's WML), but we seem to do that for historical reasons anyway. Probably best not to waste time with it in this PR.

src/campaign_server/addon_utils.cpp
```diff
@@ -92,23 +92,43 @@ std::string format_addon_feedback_url(const std::string& format, const config& p
...
 		} else {
 			find_translations(dir, addon);
 		}
 	}
+
+	for(config& locale : addon.child_range("translation")) {
+		if(!locale["supported"].to_bool(false)) {
+			locale["supported"] = false;
+		}
+	}
```
shikadiqueen, line 131: I assume this is intended to prevent people from inputting crap into `translation.supported` values, but it just makes `false` values become... `false`. It would make more sense to go over all `translation` nodes in the caller (to make the semantics of `find_translation` less oblique) and **delete** the `supported` attributes as well **all unrecognized data** (including other WML children nodes) in a first pass at the start of the function, before looking at the archive's contents.

The reason for deleting all unrecognized data is that otherwise it could well become possible for add-on authors to break clients in the future. As well as the fact that they could just abuse this to deliver extraneous data that the client won't process and increase the size of the `[request_campaign_list]` response for no good reason.

src/campaign_server/campaign_server.cpp
```diff
@@ -623,7 +623,7 @@ void server::handle_request_campaign_list(const server::request& req)
 
 			for(const config& j : i.child_range("translation"))
 			{
-				if(j["language"] == lang) {
+				if(j["language"] == lang && j["supported"].to_bool(true)) {
```
shikadiqueen, line 626: `supported` should default to `false` in all contexts (especially this one) instead of `true`. This can also be achieved by calling `to_bool()` without an argument.

src/campaign_server/campaign_server.cpp
```diff
@@ -881,6 +881,11 @@ void server::handle_upload(const server::request& req)
 			(*campaign).add_child("feedback", url_params);
 		}
 
+		(*campaign).clear_children("translation");
+		for(const config& locale_params : upload.child_range("translation")) {
+			(*campaign).add_child("translation", locale_params);
+		}
+
```
shikadiqueen, line 888: This is the best place to perform the sanitation pass I refer to in my comment on `find_translations()`.

src/campaign_server/addon_utils.cpp
shikadiqueen: Accidentally posted this with obsolete pointers the first time around. I want to re-emphasize that `find_translation` shouldn't be in charge of sanitizing the input, that should be the caller's responsibility.

src/campaign_server/campaign_server.cpp
kabachuha: Well, thanks for clarification with this question. I'll remove it then.

src/addon/info.cpp
kabachuha: I think it can be done with two loops in that case. The first will look for the full locale, the second for the short version.

src/addon/info.cpp
kabachuha: Ok.

src/addon/info.hpp
kabachuha: Well, technically yes. Would you suggest adding a filter on the server side or check whether it's invalid in some other way?

src/campaign_server/campaign_server.cpp
kabachuha: I believe it's made for the backwards compatibility, so if the `supported` attribute is abscent, it will be assumed as true. Newer addons will always have either `supported=yes or no` variable.

src/campaign_server/addon_utils.cpp
kabachuha: 1. Very well.
2,3. Initially, I tried
```
locale = addon.add_child("translation");
locale["language"] = locale_id;
locale["supported"] = true;
```

But it crashed on the first line due to the validation error because the `=` operator had been overloaded with a validation check.
I understand from your suggestion that I should try referring it with a pointer instead? 
Something like this: `locale_pointer = &addon.add_child("translation");` ?

src/campaign_server/campaign_server.cpp
kabachuha: Ok

src/campaign_server/addon_utils.cpp
kabachuha: Good idea, thanks! I'll make it.

src/addon/info.cpp
kabachuha: Well, `display_title_translated_or_original()` and `display_title_full_shift()` can be moved to the classes where they are referrenced. However, `display_title_full()` is used in quite a few various places to indicate the name of the addon, so I doubt if I should move it somewhere.

src/campaign_server/campaign_server.cpp
kabachuha: I think with the sanity check `j["supported"].to_bool(true)` will be redundant, so ok.

src/campaign_server/addon_utils.cpp
kabachuha: I'll add the emptiness check to the sanitation check, then.

src/addon/info.cpp
kabachuha: Ok, will use it.

src/campaign_server/campaign_server.cpp
shikadiqueen: That's a good point, I forgot that `supported` is going to be entirely absent from campaignd configs initially. In that case it should default to `false` across the board and it should be guaranteed to be present in the config for new/updated uploads.

src/campaign_server/addon_utils.cpp
shikadiqueen: The reason you can't assign to `locale` if the WML node is absent is that in C++ references are immutable. Once they are initialized, they are pointing to whatever they were initialized with (even if it's destroyed, this creates a dangling reference, which leads to undefined behaviour). So you need to take a new reference from the return value of `add_child()` -- otherwise you'd be attempting to call the assignment operator for the original (dummy) instance with the result of `add_child()` as its argument.

src/campaign_server/campaign_server.cpp
kabachuha: I think it will be defaulted to true on the client. But the server will guarantee it for the new ones as you said.

src/addon/info.cpp
shikadiqueen: I agree `display_title_full()` probably should remain where it is since it's used by the add-ons client class.

src/addon/info.cpp
shikadiqueen: The fact that you need to do any passes at all shows that the the design may not be terribly optimal (not that it matters too much in this code path since it's client-side and not a particularly hot one). Personally I would've made `addon_info::info_translations` a `std::map` indexed by locale ids to save myself the trouble of having to go over the full list every time I'm looking for a particular entry.

src/addon/info.hpp
shikadiqueen: AFAICT the best validity criterion here (as in the only one that doesn't require contorting the implementation of anything else around it) really is checking if the translated title is empty.

src/campaign_server/campaign_server.cpp
shikadiqueen: pffft, I meant it should default to `true`, yes. Otherwise what I said doesn't make any sense.

shikadiqueen: Also, regarding the client side: it should also default to `true`, since if the server doesn't send it it means it's an old server. Which means that it will certainly only emit the `translations` tags for translations that have catalogues.

src/addon/info.hpp
kabachuha: Ok, very well.

kabachuha:
I think I made all the requested changes. Also tested them with my pet server, everything looks fine so far.

Pentarctagon:
One of the travis jobs is failing: https://travis-ci.com/github/wesnoth/wesnoth/jobs/358117144#L707

kabachuha:
Fixed it, @Pentarctagon.

src/addon/info.cpp
```diff
@@ -117,7 +117,7 @@ void addon_info::write(config& cfg) const
 	cfg["uploads"] = this->uploads;
 	cfg["type"] = get_addon_type_string(this->type);
 
-	for(const std::pair<std::string, addon_info_translation> &element : this->info_translations) {
+	for(std::pair<std::string, addon_info_translation> element : this->info_translations) {
```
shikadiqueen, line 120: For reasons I don't quite understand only one compiler (clang 7) took fault to the original. Ideally this should be declared as `const auto& element` to avoid the copies and for readability (we all know the value type for `info_translations`, it's not necessary to specify it explicitly).

src/addon/info.cpp
```diff
@@ -119,8 +117,10 @@ void addon_info::write(config& cfg) const
 	cfg["uploads"] = this->uploads;
 	cfg["type"] = get_addon_type_string(this->type);
 
-	for(const addon_info_translation& locale : this->info_translations) {
-		locale.write(cfg.add_child("translation"));
+	for(const std::pair<std::string, addon_info_translation> &element : this->info_translations) {
+		config* locale = &cfg.add_child("translation");
+		(*locale)["language"] = element.first;
+		element.second.write(*locale);
```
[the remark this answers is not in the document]
kabachuha: Done.

src/addon/info.cpp
```diff
@@ -151,17 +151,23 @@ std::string addon_info::display_title() const
...
+		std::map<std::string, addon_info_translation>::const_iterator info;
+
+		info = info_translations.find(locale);
+		if(info != info_translations.end()) {
+			return info->second;
+		}
+
+		info = info_translations.find(locale.substr(0, 2));
+		if(info != info_translations.end()) {
+			return info->second;
```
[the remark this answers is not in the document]
kabachuha: Ok.

src/campaign_server/campaign_server.cpp
```diff
@@ -883,7 +883,18 @@ void server::handle_upload(const server::request& req)
...
+				(*locale)["language"] = locale_params["language"].str();
+				(*locale)["supported"] = false;
+
+				if(!locale_params["title"].empty()) {
+					(*locale)["title"] = locale_params["title"].str();
+				}
+				if(!locale_params["description"].empty()) {
+					(*locale)["description"] = locale_params["description"].str();
+				}
+			}
```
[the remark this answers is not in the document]
kabachuha: Thanks! Changed it to a reference.

src/addon/info.cpp
kabachuha: Ok, now it's working well with that compiler.

src/campaign_server/addon_utils.cpp
```diff
@@ -92,18 +92,14 @@ std::string format_addon_feedback_url(const std::string& format, const config& p
...
-		else {
-			addon.add_child("translation")["language"] = locale_id; //Doing this to circumvent a weird validation bug
-			addon.find_child("translation", "language", locale_id)["supported"] = true;
-		}
+	config* locale = &addon.find_child("translation", "language", locale_id);
+	if(!*locale) {
+		locale = &addon.add_child("translation");
+		(*locale)["language"] = locale_id;
 	}
+	(*locale)["supported"] = true;
```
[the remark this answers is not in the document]
kabachuha: Ok, I made `locale_id` a const reference. However, I left the pointer untouched in this place because the reference reassignment had caused the server crash.

<details>
  <summary>The old crashing code</summary>

```
void support_translation(config& addon, const std::string& locale_id)
{
	config* locale = &addon.find_child("translation", "language", locale_id);
	if(!*locale) {
		locale = &addon.add_child("translation");
		(*locale)["language"] = locale_id;
	}
	(*locale)["supported"] = true;
}
```

</details>

PR #4985, opened 2020-07-05
use shared_ptr for unit_ptr

gfgtdf:
This make the code a bit simpler, furthermore it
allows us to use weak_ptr to untsi shodul we ever
need that.

### DIFF
--- a/src/units/unit.hpp
+++ b/src/units/unit.hpp
@@ -210,7 +210,7 @@ public:
 
 	unit_ptr clone() const
 	{
-		return unit_ptr(new unit(*this));
+		return unit_ptr(std::shared_ptr<unit>(new unit(*this)));
 	}
 
 	//unit_ptr shared_from_this()
--- a/src/units/unit.hpp
+++ b/src/units/unit.hpp
@@ -126,7 +126,7 @@ private:
 /**
  * This class represents a *single* unit of a specific type.
  */
-class unit
+class unit : public std::enable_shared_from_this<unit>
 {
 public:
 	/**
@@ -213,10 +213,10 @@ public:
 		return unit_ptr(new unit(*this));
 	}
 
-	unit_ptr shared_from_this()
-	{
-		return unit_ptr(this);
-	}
+	//unit_ptr shared_from_this()
+	//{
+	//	return unit_ptr(this);
+	//}
 
 	virtual ~unit();
 
@@ -1779,21 +1779,11 @@ public:
 	 */
 	unit& mark_clone(bool is_temporary);
 
-	/** @} */
-
-	long ref_count() const
-	{
-		return ref_count_;
-	}
-
-	friend void intrusive_ptr_add_ref(const unit*);
-	friend void intrusive_ptr_release(const unit*);
 
 	void set_appearance_changed(bool value) { appearance_changed_ = value; }
 	bool appearance_changed() const { return appearance_changed_; }
 
 protected:
-	mutable long ref_count_; // used by intrusive_ptr
 
 private:
 	map_location loc_;
--- a/src/units/unit.hpp
+++ b/src/units/unit.hpp
@@ -145,7 +145,13 @@ private:
 	// Copy constructor
 	unit(const unit& u);
 
-	unit();
+
+	struct unit_ctor_t {};
+public:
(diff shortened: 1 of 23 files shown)


gfgtdf:
okay this doesn't work compltey yet

CelticMinstrel:
I don't think it's correct to replace unit references with unit pointers. You should use pointer or reference based on whether the function is equipped to deal with a null. The ones you changed from reference to pointer should probably be changed back.

I didn't look at this in detail yet, but did you use `std::enable_shared_from_this` for unit? If so I would recommend exposing the `shared_from_this()` function publicly (perhaps by overloading the address operator to return it).

gfgtdf:
hmm converting to unit_ptr to refernces and back is not omething i realyl liek to be honest. I'm thinking about adding a new class 'shared_reference' that's basicially  shared_ptr wrapper that cannot be assigned nullptr.

CelticMinstrel:
> hmm converting to unit_ptr to refernces and back is not omething i realyl liek to be honest.

Why not? If it works for normal pointers, why shouldn't we allow it to work for shared pointers? I think a shared_reference class would be overkill for this, honestly. That said... I can't say it's _wrong_. There _is_ one minor downside to taking the reference, which is that it's not an owning reference. I think that usually won't be an issue, but... with a shared_reference type class you could make it an owning reference. At the very least, I would prefer a shared_reference over replacing all those references by pointers.

gfgtdf:
okay there were some cases in the naimation code that converted between unit_ptr and unit* so i had to use shared_from_this anyways if i didnt want to to a really large refactor.. But it should fork fine now

CelticMinstrel:
I definitely like overloading operator& to return shared_from_this(), but I suppose I'm probably in the minority there…? Either way, get_shared_ptr() is perfectly reasonable too. Maybe a tad long, but whatever.

gfgtdf:
> I definitely like overloading operator& to return shared_from_this(),

there are major problems with overloading operator &, for example in case where unit is incomplete, it will still return a noprmal pointer so the result of `&u` depends on whether unit.hpp was included or not.

stevecotton:
Doesn't build with GCC 10, it needs an `#include <stdexcept>` in src/utils/shared_reference.hpp.

Also, there's no documentation for the shared_reference class, other than the commit message.

CelticMinstrel:
The shared_reference class looks fairly self-explanatory to me - it's literally a non-nullable shared_ptr.

Again I might be in the minority, but I probably would have written `operator&` instead of `operator shared_ptr`. I realize that there's a common wisdom that overloading the address-of operator is "bad", but my understanding is it's bad because you can change the semantics of it. Overloading it to return a `shared_ptr` does not change the semantics, so in my opinion it's a reasonable thing to do.

> there are major problems with overloading operator &, for example in case where unit is incomplete, it will still return a noprmal pointer so the result of `&u` depends on whether unit.hpp was included or not.

This _is_ a legitimate issue that I hadn't thought of. However, I don't think it would allow incorrect code to compile in many cases (if any). If you're taking the address of a unit, chances are you're going to pass it to a function that takes a `unit_ptr` (or return it as a `unit_ptr`), so that will fail to compile because `unit*` is not implicitly convertible to `unit_ptr`. If you're not doing that, you're probably doing something that otherwise requires `unit` to be complete, such as accessing a member of it.

So I would classify it as a minor problem, rather than a major problem, though there could be some case I haven't thought of that could change my mind.

soliton-:
I think it would go a long way to simply add a single line description instead of requiring everyone to read the implementation.

src/units/unit.hpp
```diff
@@ -210,7 +210,7 @@ class unit : public std::enable_shared_from_this<unit>
 
 	unit_ptr clone() const
 	{
-		return unit_ptr(new unit(*this));
+		return unit_ptr(std::shared_ptr<unit>(new unit(*this)));
```
AI0867, line 213: `return unit_ptr(std::make_shared<unit>(*this))`?

gfgtdf: yes i should do this, the reason i added that line is that i temporarily, replaced unit_ptr with a wrapper for std::shared_ptr<unit> without the constructor that takes 1 unit* to detect all cases hen unit* is converted into unit_ptr

PR #4986, opened 2020-07-06
Create Unit UI improvements & unit type variation selection

shikadiqueen:
This is a bit of a messy pull request that touches both backend code (the synced commands functionality) and UI code (context menu, Create Unit) bringing a few improvements to the Create Unit dialog.

Firstly, smaller aspects of Create Unit should be more streamlined by removing redundant labels and caption text.

Secondly, the dialog now supports selecting the unit type variation for the unit that will be created.

image

In order to implement this functionality I had to touch a surprising amount of code in `menu_events.cpp`, the synced commands functionality, as well as a couple key methods of the `unit` class. As far as I can tell there shouldn't be unexpected side-effects for these changes, but just in case I want @gfgtdf's review in addition to @Vultraz's anyway since it's not code I am well familiarized with at the moment.

(Also, hopefully the UI implementation side of things conforms to the current UI coding style.)

### DIFF
--- a/data/gui/window/unit_create.cfg
+++ b/data/gui/window/unit_create.cfg
@@ -201,10 +201,32 @@
 
 										[column]
 											grow_factor = 1
-											horizontal_grow = true
+											horizontal_alignment = "right"
 											[grid]
 												[row]
 													grow_factor=0
+													[column]
+														border = "all"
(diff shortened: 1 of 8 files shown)


gfgtdf:
looks good to me

PR #4988, opened 2020-07-09
Various fixes

Pentarctagon:
Tested locally:
* Starting a game with and without mods using 1.14.13 (Steam release).
* Starting a game with and without mods using 1.14.13+dev.
* Loading a saved game in MP.

### DIFF
--- a/src/server/server.cpp
+++ b/src/server/server.cpp
@@ -1779,8 +1779,14 @@ void server::handle_player_in_game(socket_ptr socket, std::shared_ptr<simple_wml
 			const std::vector<std::string> mods = utils::split(m["active_mods"].to_string());
 			const std::vector<std::string> mod_sources = utils::split(m["active_mod_versions"].to_string());
 			const std::vector<std::string> mod_versions = utils::split(m["active_mod_addon_ids"].to_string());
-			if(mods.size() != 0) {
-				for(unsigned int i = 0; i < mods.size(); i++){
+			const bool mod_info_present = mods.size() == mod_sources.size() && mod_sources.size() == mod_versions.size();
+			if(!mod_info_present && (!mod_sources.empty() || !mod_versions.empty())) {
+				WRN_SERVER << "mod info mismatch for game " << g.id() << ", " << g.db_id() << " - mods: '" << m["active_mods"].to_string() << "', mod_versions: '" << m["active_mod_versions"].to_string() << "', mod_sources: '" << m["active_mod_addon_ids"].to_string() << "'";
+			}
+			for(unsigned int i = 0; i < mods.size(); i++){
+				if(!mod_info_present) {
+					user_handler_->db_insert_modification_info(uuid_, g.db_id(), mods[i], "", "");
+				} else {
 					user_handler_->db_insert_modification_info(uuid_, g.db_id(), mods[i], mod_sources[i], mod_versions[i]);
 				}
 			}
--- a/src/game_initialization/mp_game_utils.cpp
+++ b/src/game_initialization/mp_game_utils.cpp
@@ -80,8 +80,6 @@ config initial_level_config(saved_game& state)
 
 	config level = state.to_config();
 	add_multiplayer_classification(level.child_or_add("multiplayer"), state);
-	level.child("multiplayer")["mp_scenario_addon_id"] = level.child("scenario")["addon_id"].str();
-	level.child("multiplayer")["mp_scenario_addon_version"] = state.to_config().child("scenario")["addon_version"].str();
 
 	// [multiplayer] mp_era= should be persistent over saves.
 	std::string era = params.mp_era;
@@ -97,6 +95,17 @@ config initial_level_config(saved_game& state)
 
 	const config& game_config = game_config_manager::get()->game_config();
 	const config& era_cfg = game_config.find_child("era", "id", era);
+
+	if(const config& temp_cfg = game_config.find_child("scenario", "id", level.child("multiplayer")["mp_scenario"])) {
+		level.child("multiplayer")["mp_scenario_addon_id"] = temp_cfg["addon_id"].str();
+		level.child("multiplayer")["mp_scenario_addon_version"] = temp_cfg["addon_version"].str();
+	} else if(const config& temp_cfg = game_config.find_child("multiplayer", "id", level.child("multiplayer")["mp_scenario"])) {
+		level.child("multiplayer")["mp_scenario_addon_id"] = temp_cfg["addon_id"].str();
+		level.child("multiplayer")["mp_scenario_addon_version"] = temp_cfg["addon_version"].str();
+	} else {
+		level.child("multiplayer")["mp_scenario_addon_id"] = "";
+		level.child("multiplayer")["mp_scenario_addon_version"] = "";
+	}
 
 	if(!era_cfg) {
 		if(!params.saved_game) {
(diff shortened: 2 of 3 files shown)


gfgtdf:
I'd be better to use level.child_or_empty() here to make absoluteley sure that it cannot give a manditory child missing error,

Pentarctagon:
Updated.

src/server/server.cpp
```diff
@@ -1779,8 +1779,11 @@ void server::handle_player_in_game(socket_ptr socket, std::shared_ptr<simple_wml
 			const std::vector<std::string> mods = utils::split(m["active_mods"].to_string());
 			const std::vector<std::string> mod_sources = utils::split(m["active_mod_versions"].to_string());
 			const std::vector<std::string> mod_versions = utils::split(m["active_mod_addon_ids"].to_string());
-			if(mods.size() != 0) {
-				for(unsigned int i = 0; i < mods.size(); i++){
+			const bool mod_info_present = mods.size() == mod_sources.size() && mod_sources.size() == mod_versions.size();
+			for(unsigned int i = 0; i < mods.size(); i++){
+				if(!mod_info_present) {
+					user_handler_->db_insert_modification_info(uuid_, g.db_id(), mods[i], "", "");
```
soliton-, line 1785: Are these inserts useful? Perhaps just log a warning about inconsistent mod info?

soliton-: Ah wait, `active_mods` contains something, ok. I suppose we can investigate inconsistent mod info from the DB then. No need for a log message.

soliton-: Might still be useful to log what is in `active_mod_versions` and `active_mod_addon_ids`.

Pentarctagon: Though would it be useful, given that at the moment every released 1.14 client would be logging it?

soliton-: Because versions and ids are empty? We wouldn’t need to log those since there’s no additional info there.

Pentarctagon: What would be the need to add logging here, then?

gfgtdf:
Well the part in mp_game_utils.cpp looks good, in the sense that it reliably fixes the crash, i'm still unsure whether that was the correct place for this code in the first place, especially considering that [multiplayer][addon] already exists.

Pentarctagon:
Would it make more sense to look up the scenario add-on and add-on version via `game_config_manager`, like how it's done for the era and any modifications?

gfgtdf:
no what i mean is: we already send the [multiplayer][addon] tags to inform other players which addons need to be downloaded. And the code you added in mp_game_utils.cpp sends the same information again. It's probably better to reuse the old information instead of sending the ame information twice to the server (this in particular applies to 1.15, in 1.14 it might be better not to risk changing the existing code ).



As i mentioned  before on discord, [addon] contains onyl information about addons with require_download=yes (whcih is the default fo campaigns and eras btw) this could however be changed.

Pentarctagon:
So you mean that all add-ons (modifications, era, scenario/campaign) should have this information sent in [multiplayer][addon]?  I'd assume that wouldn't be backward compatible with earlier clients then, since they wouldn't know that not all [addon]s need to be downloaded.

Pentarctagon:
Which I guess would be done by modifying `src/saved_game.cpp`'s `check_require_scenario()` and `load_mod()` methods to, instead of skipping created the [addon] tag entirely when something isn't required, always create an [addon], add attributes indicating whether they are required or not, and skipping the call to the `update_addon_requirements()` method for things that aren't required.

Then there'd be somewhere else that needs to be updated to only look for [addon] entries with `required=yes` instead of looking at all [addon] entries.

gfgtdf:
> I'd assume that wouldn't be backward compatible with earlier clients then

Yes exactly that's why  made  the comment about 1.14/1.15

> Then there'd be somewhere else that needs to be updated to only look for [addon] entries with required=yes instead of looking at all [addon] entries.

I personaly do i'ta  bit differently  instead of addinga  new required= attribute, we should just omit the min_version attribute for addons that don't require installation, change the code that reads [addon] accordings. (to make 'no min_version' mean no_required)


It's this code btw https://github.com/wesnoth/wesnoth/blob/45fec655c2e13b4655f154a1de85b31225184b44/src/game_initialization/lobby_data.cpp#L469 . Not sure why it operates on config objects rather than mp_game_settigns objects.

Pentarctagon:
For this PR then, I'd prefer to use `game_config_manager` like I said above, since that would also address the issue of the source add-on ID and version missing from games reloaded in MP.  The exception to that being if someone tried to reload a game in MP using a scenario they no longer have installed, but I'd assume that having that happen and also still work is a rare occasion.

For 1.15, I'll do the [multiplayer][addon] approach, though I would prefer to have a separate `required` attribute for readability reasons unless you're strongly opposed - someone else coming to look through this would probably more easily understand `required=yes|no` vs `min_version` being present or not.

gfgtdf:
> For 1.15, I'll do the [multiplayer][addon] approach, though I would prefer to have a separate required attribute for readability reasons unless you're strongly opposed - someone else coming to look through this would probably more easily understand required=yes|no vs min_version being present or not.

Well okay but keep in mind that multiplayer/era/etc with require_download=no should  not increase the min_version for their add-on (this is important becasue one addon might supply many of era/multiplayer/campaign and oynl soem fo them might have require_download and might require different min_versions)

>  The exception to that being if someone tried to reload a game in MP using a scenario they no longer have installed, but I'd assume that having that happen and also still work is a rare occasion.

I think the more interresting question is what we should use when the wml from the savefile is form an older addon version, but the host uses a newer version.

Pentarctagon:
> Well okay but keep in mind that multiplayer/era/etc with require_download=no should not increase the min_version for their add-on (this is important becasue one addon might supply many of era/multiplayer/campaign and oynl soem fo them might have require_download and might require different min_versions)

For add-ons that aren't required, I would skip straight to https://github.com/wesnoth/wesnoth/blob/master/src/mp_game_settings.cpp#L195, correct?

> I think the more interresting question is what we should use when the wml from the savefile is form an older addon version, but the host uses a newer version.

To maintain full accuracy would require tracking the versions of all add-ons that are part of the save file (including dependencies, which could themselves potentially be added or removed), which I doubt is worth the effort or added complexity.

gfgtdf:
> For add-ons that aren't required, I would skip straight to https://github.com/wesnoth/wesnoth/blob/master/src/mp_game_settings.cpp#L195, correct?

im not sure what you mean here

Pentarctagon:
If the add-on isn't found and isn't required, then all that needs to be done is `addons.emplace(cfg["id"].str(), new_data);`.

src/server/server.cpp
soliton-: I'm saying that it might be useful to log what is in `active_mod_versions` and `active_mod_addon_ids` when the sizes are not all equal, so we can figure out if something goes wrong and hopefully also see why.

My assumption is that when you said "at the moment every released 1.14 client would be logging" it means those two keys are empty. In that case logging would not be useful though so that case should be excluded from logging.

Pentarctagon: Alright, updated.

soliton-: I think the logging should be outside the loop? Also please add the game IDs so we can reference the DB entry to the log message.

Something like "mod info mismatch in game ... mods: ..."

Pentarctagon: Updated.

Issue #3265, opened 2018-06-17
Crash when showing the credits if they're too long


Konrad22:
I don't know what is causing it. Reloading and ending again doesn't lead to the crash.

sevu:
I am afraid that this is hard to troubleshoot without more info.
A backtrace or stacktrace would probably help, but ...
 if you are on windows, I don't know how one can get it
 if you are on linux, one can get it with `corecumpctl`

Konrad22:
[wesnoth-20180720-195331-5016.log](https://github.com/wesnoth/wesnoth/files/2215288/wesnoth-20180720-195331-5016.log)
Maybe this can help?

gfgtdf:
ok the logs shows pango erros in a >100000 lines credit text and then gives a 'out of memory'. This makes it seem somewhat likely the out of memeory is related to the lots of text in the credits screen.

Konrad22:
Is the memory, which is avaible for something like this depended on my pc or does it depend on wesnoth?

CelticMinstrel:
Memory available would depend on the computer, not on Wesnoth, but if Wesnoth is allocating a ridiculous amount of memory, then that needs to be fixed somehow.

Konrad22:
How do I check wheter Wesnoth is allocating a ridiculous amount of memory instead of my pc just being bad?

gfgtdf:
> Memory available would depend on the computer, not on Wesnoth

usually this is not the case. Wesnoth releases (at least on windows) is a 32bit exacutable with a maximum of 2gb ram, even if your computer has a lot more ram. On the other hand if the 'real' memeory would go rare then the behviour (at least on windows) would be quite differnt, first the pc would try to swap everying around making everythign extremely slow, then at some point you get a erro message like 'Your computer is low on meemory please close procecces or i will' (not sure on the exact wording here)


In any case this is clearly an bug in wesnoth. I somehow remember that there was at some point a discussion about the screidsscreen allocating one big surface for the scrdits screen, so if we assume each textline to be 20 pixels and the screensize to be 1000 pixels 3000 lines of text and each pixel to be encoded with 4 bytes thats 240,000,000 bytes.

Which is a lot (and clearly too much) but usually not enough to cause an out of memory.

Which campaign did you play when you get this error? (and with mods or not)

Konrad22:
[...]

PR #4992, opened 2020-07-13
Throw instead of crashing when handling extremely long text

stevecotton:
Both of these still result in the program quitting, they just make it a graceful quit with a hopefully useful diagnostic. 

The length of text required to trigger these is excessive; in one case the trigger is to make (width * height * 32 bits per pixel) exceed 2 gigabytes, for debugging I added a comment to the credits with 30000 short lines and a single 12000 character line. These changes are inspired by bug #3265, but I think the already-merged fix to that bug handles all sane input, the changes in this PR handle edge cases that need deliberately crafted input to trigger them.

### DIFF
--- a/src/font/text.cpp
+++ b/src/font/text.cpp
@@ -683,6 +683,17 @@ void pango_text::rerender(const bool force)
 		cairo_format_t format = CAIRO_FORMAT_ARGB32;
 		const unsigned stride = cairo_format_stride_for_width(format, width);
 
+		// TODO: a sane value should be chosen for this arbitrary limit. The limit currently merely prevents arithmetic
+		// overflow when calculating (stride * height), and still allows this function to allocate a 2 gigabyte surface.
+		//
+		// Making the limit match the amount that can be handled by a single call to render() would allow this function to be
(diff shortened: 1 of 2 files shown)


gfgtdf:
If you are adding that exception, you should add more information, i actually did a similar thing although only in the 1.14 branch https://github.com/wesnoth/wesnoth/commit/6cdf6525e77a614dd745b3e75c410e2bec9d944f#diff-05a06dec7741cfcf562aed3765cf1617

Also from the commit message it sounds like you only considered the case of canvas.cpp but there are probably also other places that call  surface::surface(int w, int h), did you check that they also still work correctly in case of a bis surface?

stevecotton:
Thanks for the feedback, and I see that I'd also missed the unit test failures. I'll rework this and reopen the PR when it's ready for another review.

stevecotton:
I've opened a separate PR, #5042, for the text.cpp/text.hpp part of this.

I still intend to investigate all callers of surface::surface(w,h) and then reopen this PR for the surface change, but that's a big task very low priority, so I doubt it'll be done soon.

Issue #2964, opened 2018-04-25
'todo:' notes in help pages for Scenario Editor


gfgtdf:
some help pages, in particular those about the editor contain are incomplete and contain 'todo:' comments.

stevecotton:
I'd like to re-enable the editor help pages for the 1.15 branch, with the idea that they'll either be improved or hidden again before 1.16.0.

stevecotton:
The remaining todos in the help are only about the scenario editor - retitling the issue to match that.

jostephd:
Marked blocker for the same reason as #4280: We shouldn't release 1.16.0 with todos in the help.

Current list:

```
git grep -ni todo data/core/editor/
data/core/editor/help.cfg:133:    text= "<img>src=icons/action/editor-tool-label_60.png align=left box=yes</img>" + _"TODO"
data/core/editor/help.cfg:141:    text= "<img>src=icons/action/editor-tool-item_60.png align=left box=yes</img>" + _"TODO"
data/core/editor/help.cfg:159:    text= "<img>src=icons/action/editor-tool-village_60.png align=left box=yes</img>" + _"TODO"
data/core/editor/help.cfg:167:    text= "<img>src=icons/action/editor-tool-unit_60.png align=left box=yes</img>" + _"TODO"
data/core/editor/help.cfg:299:    text= _ "TODO"
data/core/editor/help.cfg:349:TODO.>>
data/core/editor/terrain-groups.cfg:8:#TODO
```

stevecotton:
@fendrin what's to be done in the TODO in terrain-groups.cfg, please?

It was added in 169c74ba55048f12731b86b0a537a99f649af3c5, which also adds `[terrain-group]id=autumn` to that file. My guess is that the TODO was just a reminder to add that terrain-group, and you forgot to remove the TODO comment.

Re: jostephd's list:
* for four of them, the subject is obvious
* line 299 is the Time Schedule Editor
* line 349 is about the file format that the Scenario Editor mode uses - should probably include the hints in https://github.com/wesnoth/wesnoth/issues/1380#issuecomment-528818324

jostephd:
If we don't hear otherwise, I suggest to just remove the TODO in terrain-groups.cfg.

@stevecotton Do you plan to work on this further before string freeze? I assume these todos were made user-visible again by your PR, #4414.

stevecotton:
[...]

Issue #1380, opened 2017-05-08
Map editor does not preserve unknown properties (GNA #21859)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?21859)** by dindoz on 2014-03-23

_ Made a map_scenario.cfg with the editor for being able to use all cool new option 

_ generate map and add some unit and side

_ save the map_scenario

_ open the map_scenario.cfg with Kwrite for exemple.

_ add some capacity to some units or side that cant be added by the editor.

_ save the file, open it with the editor and save the map_scenario.cfg.

: All capacity manually added are gone...

(Reproduced on ubuntu 13.10)
Release: 1.11.11
Priority: 5 - Normal
Severity: 3 - Normal

wesnoth-bugs:
**Modified on 2014-03-24**

**dindoz added attachment:** [aaa.cfg](http://gna.org/bugs/download.php?file_id=20427)

wesnoth-bugs:
**Modified on 2014-06-05**

**fendrin changed status:** None -> Postponed

Wedge009:
I don't profess to know much about how the internal editor is used, but I would expect that if one is doing some manual work on a map file that the editor doesn't handle, then of course re-saving that file within the editor would discard all the manual work.

I'm inclined to treat this as invalid - what sort of things do folks need to do that can't be applied with the internal editor?

gfgtdf:
> what sort of things do folks need to do that can't be applied with the internal editor?

well there is some stuff like unit variables or modifications/objects, but all thouse things can also be added via [modify_unit] in events i think.

The real problem here is how poorly the editor is documented, you are not supposed to modify the editor generatd file, instead  it is iirc supposed  to used is to include the editor file in the real scenario like
```
[scenario]
  ## edito generated scenario file
  {~/add-ons/my_addon/maps/scenario1.cfg}
  [event]
    ## do stuff
  [/event]
[/scenario]
```

jostephd:
> The real problem here is how poorly the editor is documented, you are not supposed to modify the editor generatd file, instead it is iirc supposed to used is to include the editor file in the real scenario like

In this case, let's:

1. Make the cfg files generated by the editor start with a hard-to-miss comment saying `Do not edit this file manually. Changes will be lost. Do this instead: (here put the code @gfgtdf posted)`

2. Let's make the editor error out if the file it loaded has a key or tag that the editor doesn't know how to generate.

[...]

PR #4999, opened 2020-07-19
Add help topics for the Editor (as it stands in the 1.14 branch)

stevecotton:
Add help topics for the Editor, as it stands in the 1.14 branch

This commit is on the master branch. There have been no changes at all to the editor documentation in the 1.14 branch, while there have been several updates to the master branch that would be good to backport if the help is displayed in the 1.14 branch - therefore my plan is to get this set of changes reviewed in the master branch, and then simply pull the entire file into the 1.14 branch along with enabling the editor documentation (a small change to data/core/help.cfg, as done in commit ed611f94385f608512da55d691d3c9753240d5d5).

A couple of warnings about the scenario mode being buggy are added to the user-visible text, find these by looking for <bold> tags within the text.

<strike>Original draft PR description follows:

I'm hoping that someone who uses the scenario editor will take over and write
the documentation. Without the knowledge of someone who uses the editor, the
only option is to leave a gap in the documentation; this commit is merely
hiding the problem to remove the "Blocker" tag from issue #2964.

This leaves dead links in the "Editor Tools" page, which I hope will be fixed
by someone else adding the relevant pages, rather than having to just remove
that section.</strike>

### DIFF
--- a/data/core/editor/help.cfg
+++ b/data/core/editor/help.cfg
@@ -3,7 +3,7 @@
 [section]
     id=editor
     title= _ "Map and Scenario Editor"
-    topics=..editor, editor_modes, editor_toolkit, map_format
+    topics=..editor, editor_map_format, editor_separate_events_file
     sections=editor_mode_terrain, editor_mode_scenario
     sort_topics=no
 [/section]
@@ -18,7 +18,7 @@
 [section]
     id=editor_mode_scenario
     title= _ "Scenario Editor"
-    topics=editor_named_area, editor_playlist, editor_tool_label, editor_tool_item, editor_tool_village, editor_tool_unit
+    topics=..editor_mode_scenario, editor_tool_label, editor_tool_scenery, editor_tool_unit, editor_tool_village, editor_time_schedule, editor_playlist
     sort_topics=no
 [/section]
 
@@ -105,7 +105,7 @@ The paste tool also has some clipboard-manipulation functions:" +
 [topic]
     id=editor_tool_starting
     title= _ "Starting Locations Tool"
-    text= "<img>src=icons/action/editor-tool-starting-position_60.png align=left box=yes</img>" + _"Defines the side leader starting position
+    text= "<img>src=icons/action/editor-tool-starting-position_60.png align=left box=yes</img>" + _ "Defines the side leader starting position.
 
 This tool sets the side leaders’ default starting locations, and named special locations. Both types of location are enabled in both <ref>dst='..editor_mode_terrain' text='Terrain Editor'</ref> and Scenario Editor modes. The location names are shown as a list in the editor palette, clicking on the map will place that name on a hex, each location can only be placed on a single hex, and the editor will only allow one location per hex.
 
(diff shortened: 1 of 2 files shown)


gfgtdf:
It should be noted that not instead of including the editors .cfg  files via macro inclusion they can (or should) not added simply via the map_file attribute as in `map_file="map2.cfg"`.

stevecotton:
@gfgtdf I'm not sure I've understood your comment, it seems to have a "not" too many. I read the trick about including the editor-generated .cfg in a comment in another Github issue, but hadn't tried to use it until now; now that I've tried to use it, it seems a lot more complex than I thought.

I don't know whether to continue this PR, or whether to ask for help on the forums about whether this is really the best way to use the scenario editor.

data/core/editor/help.cfg
```diff
@@ -337,13 +382,19 @@ The <italic>text='map_file'</italic> method is preferred over using a preprocess
 
 The map data can stored as part of a scenario’s .cfg file, directly in the <italic>text='map_data'</italic> attribute. In other words, in the place that the preprocessor would include it when using the preprocessor-include method.
 
-<bold>text='Using Embedded Format in Terrain Mode'</bold>
+<header>text='Using Embedded Format in Terrain Mode'</header>
 
 If you are editing the map and not using the Scenario Mode support, then it’s trivial to move the data to a native map file before opening it in the editor. This conversion is recommended — the editor supports editing the content of map_data while leaving everything else in the file untouched, but this is rarely-used code. Maps opened this way are marked (E) in the Window menu.>> + "
```
stevecotton, line 387: Found some reference documentation for this, https://github.com/wesnoth/wesnoth/blob/ff0eecc7c7e8e896e1e53439a103d3991fd97289/src/editor/map/map_context.cpp#L120-L139

gfgtdf:
> @gfgtdf I'm not sure I've understood your comment, it seems to have a "not" too many. I read the trick about including the editor-generated .cfg in a comment in #1380 (comment) , but hadn't tried to use it until now; now that I've tried to use it, it seems a lot more complex than I thought.

I sometimes accidentally write 'not' instead of 'now'. The other comment is outdated and described the bahviour that was reccomended for 1.14, in 1.15 you just write  `map_file="scenario1.cfg"`

stevecotton:
>  in 1.15 you just write `map_file="scenario1.cfg"`

Doesn't work for me. Although `map_file` is preferred for the maps created by the terrain editor, it doesn't seem to have support for the WML-with-embedded-`map_data` format that the scenario editor creates.

stevecotton:
@Pentarctagon I see that you've roadmapped removing most of the scenario editor for 1.15.11 in March 2021. I fully agree with that; the effort that went into this PR has made me doubt that anyone is really using it.

A minor request, let's have a discussion about how the file handling should work. I feel we need an extended map format, which the .cfg format described in this PR is acting as; I suggest it should contain the information that would be useful for using the same map for multiple scenarios:
* the terrain
* named areas (but not assigning any related time schedules to those areas)
* named hexes (which can already be placed in the current terrain-only maps using the misnamed "starting locations" tool)
* terrain-like scenery (windmills, braziers, etc)

Pentarctagon:
I've no objection to discussing which features to keep and how to continue supporting them.

gfgtdf:
> Doesn't work for me. Although map_file is preferred for the maps created by the terrain editor, it doesn't seem to have support for the WML-with-embedded-map_data format that the scenario editor creates.

Hmm this is a bug then.

gfgtdf:
pushed a fix https://github.com/wesnoth/wesnoth/commit/67b4fb9ce087864d144f347d758f98254d747489 can you try again ?

gfgtdf:
So what i think about the scenario editor:
I think we should keep it, but see it as a extension of the map editor to add new information like areas and scenery instead of an editor to create full scenarios. And using with with map_file allows is to be used exactly like that.

I also think the format that it generated, which is the same as the [scenario] format is already the best format we can find.

I think it's important that by default the scenario editor does not add information that the user didn't explicitly add, for example currently it afaik adds the default `[time]`s to it and the `experience_modifier=70`. This should be changed.

Furthermore the i of certain features must clearly be improved.

Lastly if we are really done we could merge the 'special_locations' and the 'areas' feature into one.  The main reason why i added special_locations in the first place is that while its technicialyl inferiour to the 'areas' tool, the areas tool simply wasn't usable t that time for multiple reasons. and i had doubt that it ever would. Even more i was 100% sure that i wouldn't become usable before 1.14.

Pentarctagon:
Separate from which existing functionality to keep/remove, one thought I had about this is that rather than trying to create a separate standalone scenario cfg in the editor userdata directory to have the user choose an add-on name and then automatically create a skeleton add-on directory structure in the add-ons directory.

So if the user enters an add-on name of "MyFirstAddon", then:
* A minimal _main.cfg would be created containing the [binary_path] for MyFirstAddon. (if not already existing)
* The sub-directories would be created: maps, images, macros, scenarios, events. (if not already existing)
* The plain map data would be saved to, for example, `maps/MyFirstMap.map`.
* A scenario file `scenarios/MyFirstMap.cfg` which would include only the minimum required information to work (basically just any scenario settings, the side definitions, and the event file include).
* An "event file" would be written to `events/MyFirstMap.cfg`.  This would contain the editor-created events for placing units, etc.

This would both keep things separated by what they're doing, as well as give users a head start on creating an actual add-on that could be distributed if they want.

data/core/editor/help.cfg
```diff
@@ -280,6 +280,51 @@ To get started, refer to the <ref>dst='editor_tool_paint' text='Paint Tool'</ref
...
+
+# wmllint: markcheck off
+[topic]
+    id=editor_separate_events_file
+    title= _ "Using a separate file for WML events"
+    text= _ <<When loading a .cfg file, the scenario editor understands files created by the scenario editor, but is likely to have difficulty with files that have been edited by hand.
+
+Files created by the scenario editor omit the opening [scenario] and closing [/scenario] tags. If you are creating a campaign (or other add-on), then those tags need to be added, there is more detail about this in the <ref>dst='map_format' text='map file format'</ref> documentation.
+
+One workflow is to created a separate WML file, also with the .cfg extension, which uses the WML preprocessor to include the editor-created file. This separate file contains both the [scenario] tag and any hand-edited WML such as events. With this workflow, the add-on’s file structure could look like this:
```
Pentarctagon, line 320: > One workflow is to create**d** a separate WML file

Should just be "create".

data/core/editor/help.cfg
```diff
@@ -280,6 +280,51 @@ To get started, refer to the <ref>dst='editor_tool_paint' text='Paint Tool'</ref
 [/topic]
 # wmllint: markcheck on
 
+# wmllint: markcheck off
+[topic]
+    id=..editor_mode_scenario
+    title= _ "Scenario Editor"
+    text= _ "The scenario editor mode adds support for some map-related WML features, such as areas and scenery items. Most scenarios will still require additional WML to be written using a different tool, the scenario editor does not support scripting the scenario’s events." + "
```
Pentarctagon, line 287: > Most scenarios will still require additional WML to be written using a different tool, the scenario editor does not support scripting the scenario’s events.

This should be separated by a semi-colon rather than a comma.

data/core/editor/help.cfg
```diff
@@ -280,6 +280,51 @@ To get started, refer to the <ref>dst='editor_tool_paint' text='Paint Tool'</ref
...
+[/topic]
+# wmllint: markcheck on
+
+# wmllint: markcheck off
+[topic]
+    id=editor_separate_events_file
+    title= _ "Using a separate file for WML events"
+    text= _ <<When loading a .cfg file, the scenario editor understands files created by the scenario editor, but is likely to have difficulty with files that have been edited by hand.
+
+Files created by the scenario editor omit the opening [scenario] and closing [/scenario] tags. If you are creating a campaign (or other add-on), then those tags need to be added, there is more detail about this in the <ref>dst='map_format' text='map file format'</ref> documentation.
```
Pentarctagon, line 318: > If you are creating a campaign (or other add-on), then those tags need to be added, there is more detail about this in the <ref>dst='map_format' text='map file format'</ref> documentation.

The second comma should be a period or a semicolon.

stevecotton:
> pushed a fix [67b4fb9](https://github.com/wesnoth/wesnoth/commit/67b4fb9ce087864d144f347d758f98254d747489) can you try again ?

Thanks, that works now. I had a bug in my local testing too - I was trying to use a relative path, "./file.cfg", which works for preprocessor inclusion but not for map_file.

CelticMinstrel:
> Separate from which existing functionality to keep/remove, one thought I had about this is that rather than trying to create a separate standalone scenario cfg in the editor userdata directory to have the user choose an add-on name and then automatically create a skeleton add-on directory structure in the add-ons directory.

This is an interesting idea, but it'll make it harder for people who just want to make a custom map and play on it with a few friends, won't it? Saving the maps to the editor userdata directory means you can just load one up and play on it without any extra work.

> The sub-directories would be created: maps, images, macros, scenarios, events. (if not already existing)

Events? I don't think any of the mainline campaigns have that folder, do they?

> An "event file" would be written to `events/MyFirstMap.cfg`. This would contain the editor-created events for placing units, etc.

Uhh… what? Why would it need events to place units? Units can be placed directly in the `[side]` tag.

> This would both keep things separated by what they're doing, as well as give users a head start on creating an actual add-on that could be distributed if they want.

The downside of what you described is that it makes it harder for the user to later load up the map and edit it. It's solvable, but something that needs to be carefully considered. How does the user load the map into the editor? Do they load the scenario file? The events file? The map file? If the latter, how does the scenario editor know it's not just a plain map?

Pentarctagon:
> This is an interesting idea, but it'll make it harder for people who just want to make a custom map and play on it with a few friends, won't it? Saving the maps to the editor userdata directory means you can just load one up and play on it without any extra work.

I don't see how?  They save the scenario in the editor and then select it in the game creation screen.

> Events? I don't think any of the mainline campaigns have that folder, do they?

I don't see how that's really relevant.

> Uhh… what? Why would it need events to place units? Units can be placed directly in the [side] tag.

Both are equivalent, and using events can be made more flexible.

> The downside of what you described is that it makes it harder for the user to later load up the map and edit it. It's solvable, but something that needs to be carefully considered. How does the user load the map into the editor? Do they load the scenario file? The events file? The map file? If the latter, how does the scenario editor know it's not just a plain map?

Any of those files.  And I want to remove the distinction between there being a "scenario editor" and a "map editor", so there's just a scenario editor.

CelticMinstrel:
> I don't see how? They save the scenario in the editor and then select it in the game creation screen.

And it will work in multiplayer?

I think there _is_ some difference between loading a custom map and loading a scenario. I'm not quite sure what it is or if it matters, though.

> I don't see how that's really relevant.

Maybe not, it's just a bit strange to add a folder that (as far as I know) no-one else uses.

> Both are equivalent, 

I don't think that's true. As I recall, using `[unit]` in the `[side]` tag has the effect of "recall this unit or create it if it doesn't exist". That's definitely not true of using `[unit]` in `[event]`.

> and using events can be made more flexible.

You mean additional features could be added to the scenario editor that create events?

> Any of those files.

So, if they load the events or map file, you need to be able to detect that this is part of a scenario and load up the scenario instead. But how can you tell if a map is part of a scenario, exactly? You could jump up one level and look for a file `scenarios/map_name.cfg`, but I think that would give false positives on every mainline campaign. Or would you use the presence of an events file as an indicator? Maybe that could work…

> And I want to remove the distinction between there being a "scenario editor" and a "map editor", so there's just a scenario editor.

Yeah, I agree, as long as there's still a way to save _just_ the map and not a full scenario.

Pentarctagon:
> And it will work in multiplayer?
> 
> I think there is some difference between loading a custom map and loading a scenario. I'm not quite sure what it is or if it matters, though.

Why wouldn't it work?

> I don't think that's true. As I recall, using [unit] in the [side] tag has the effect of "recall this unit or create it if it doesn't exist". That's definitely not true of using [unit] in [event].

This is only the case for units don't have an x,y coordinate set, which isn't the case for units created in the editor.

> You mean additional features could be added to the scenario editor that create events?

I mean that it would be easier in the future to allow setting an event for something to happen on.

> So, if they load the events or map file, you need to be able to detect that this is part of a scenario and load up the scenario instead. But how can you tell if a map is part of a scenario, exactly? You could jump up one level and look for a file scenarios/map_name.cfg, but I think that would give false positives on every mainline campaign. Or would you use the presence of an events file as an indicator? Maybe that could work…

I'm not sure how that would give false positives for mainline campaign maps?  For anything mainline you'd probably just break it if you open and then resave the scenario file.

CelticMinstrel:
> This is only the case for units don't have an x,y coordinate set, which isn't the case for units created in the editor.

Really? That seems weird in itself. Why would the presence or absence of a location control something unrelated like whether to check the recall list?

> I'm not sure how that would give false positives for mainline campaign maps?

Maybe I'm remembering wrong but I thought all the mainline campaigns named their maps and scenarios to match.

Pentarctagon:
> Really? That seems weird in itself. Why would the presence or absence of a location control something unrelated like whether to check the recall list?

It's what the wiki says, at least.

stevecotton:
The current draft probably won't fit 1.15's eventual design, but does describe what 1.14 does. I'm wondering whether to take the usable documentation (and the parts about the terrain editor that are now documented in 1.15) and put them on either:

* 1.14's in-game help (and make that section non-hidden again)
* the Wiki
* forum thread https://r.wesnoth.org/t53724

Pentarctagon:
I'd say 1.14's in-game help.  That way it'll be easy to forward-port depending on what happens.

stevecotton:
The branch name is now completely the opposite of what this is, but to change the branch name I'd need to create a new PR, I'd prefer to keep the existing PR with the comments about file format docs.

> I'd say 1.14's in-game help. That way it'll be easy to forward-port depending on what happens.

I agree, however my plan is to review the PR against the master branch first, and then copy the entire file into the 1.14 branch along with enabling the editor documentation (a small change to data/core/help.cfg, as done in commit ed611f94385f608512da55d691d3c9753240d5d5). There have been no changes at all to the editor documentation in the 1.14 branch, while there have been several other updates to the master branch that would be good to backport if the help is displayed in the 1.14 branch. It also means there's no mention of the sound-source tool, but that seems self-explanatory.

A couple of warnings about the scenario mode being buggy are added to the user-visible text, find these by looking for `<bold>` tags within the text.

data/core/editor/help.cfg
```diff
@@ -105,7 +105,7 @@ The paste tool also has some clipboard-manipulation functions:" +
 [topic]
     id=editor_tool_starting
     title= _ "Starting Locations Tool"
-    text= "<img>src=icons/action/editor-tool-starting-position_60.png align=left box=yes</img>" + _"Defines the side leader starting position
+    text= "<img>src=icons/action/editor-tool-starting-position_60.png align=left box=yes</img>" + _ "Defines the side leader starting position
```
Pentarctagon, line 108: Missing a period.

data/core/editor/help.cfg
```diff
@@ -122,151 +122,208 @@ Named locations can be accessed from WML using the Standard Location Filter’s
...
-    id=editor_tool_item
-    title= _ "Item Tool"
-    text= "<img>src=icons/action/editor-tool-item_60.png align=left box=yes</img>" + _"TODO"
+    id=editor_tool_scenery
+    title= _ "Item Tool (Scenery Tool)"
+    text= "<img>src=icons/action/editor-tool-item_60.png align=left box=yes</img>" + _ "The Item Tool allows placing decorations such as windmills, bookcases and monoliths. Multiple items can be placed on the same hex.
+
+<bold>text='Note:'</bold> the tool doesn’t support deleting items once placed, nor does it support undo. Mistakes can currently only be fixed by editing the generated WML file.
+
+This tool is only available in Scenario Mode, the decorations are not part of the terrain and are implemented in the scenario using WML’s <italic>text='[item]'</italic> tag."
```
Pentarctagon, line 149: Should be a semicolon instead of a comma.

data/core/editor/help.cfg
```diff
@@ -122,151 +122,208 @@ Named locations can be accessed from WML using the Standard Location Filter’s
...
     id=editor_tool_label
     title= _ "Label Tool"
-    text= "<img>src=icons/action/editor-tool-label_60.png align=left box=yes</img>" + _"TODO"
+    text= "<img>src=icons/action/editor-tool-label_60.png align=left box=yes</img>" + _ "Put text labels on the map.
+
+• Left-click will open a dialog box to create a new label or edit an existing one.
+• Right-click deletes.
+• Drag-and-drop with the left mouse button moves labels.
+
+This tool is only available in Scenario Mode, the decorations are implemented in the scenario using WML’s <italic>text='[label]'</italic> tag."
```
Pentarctagon, line 135: Should be a semicolon instead of a comma.

data/core/editor/help.cfg
```diff
@@ -122,151 +122,208 @@ Named locations can be accessed from WML using the Standard Location Filter’s
...
 [topic]
     id=editor_tool_village
     title= _ "Village Ownership Tool"
-    text= "<img>src=icons/action/editor-tool-village_60.png align=left box=yes</img>" + _"TODO"
+    text= "<img>src=icons/action/editor-tool-village_60.png align=left box=yes</img>" + _ "This tool assigns ownership of villages at the start of a scenario. The villages must first be placed on the terrain with the <ref>dst='editor_tool_paint' text='Paint Tool'</ref>.
+
+• Left-click will assign the village to the currently-selected side.
+• Right-click will set the village back to unowned.
+
+This tool is only available in Scenario Mode, ownership information is implemented using WML’s <italic>text='[side][village]'</italic> tag."
```
Pentarctagon, line 164: Should be a semicolon instead of a comma.

Pentarctagon: Also those are two tags, so "tag" should be plural.

data/core/editor/help.cfg
```diff
@@ -122,151 +122,208 @@ Named locations can be accessed from WML using the Standard Location Filter’s
...
     id=editor_tool_unit
     title= _ "Unit Tool"
-    text= "<img>src=icons/action/editor-tool-unit_60.png align=left box=yes</img>" + _"TODO"
+    text= "<img>src=icons/action/editor-tool-unit_60.png align=left box=yes</img>" + _ "Place units belonging to the currently-selected side.
+
+• Left-click will place a unit.
+• Left drag-and-drop will move an already-placed unit.
+• Various operations are added to the right-click menu when the hex contains a unit.
+
+This tool is only available in Scenario Mode, it adds WML <italic>text='[unit]'</italic> tags to the appropriate <italic>text='[side]'</italic>."
```
Pentarctagon, line 180: Should be a semicolon instead of a comma.

data/core/editor/help.cfg
```diff
@@ -122,151 +122,208 @@ Named locations can be accessed from WML using the Standard Location Filter’s
...
+To use the tool:
 
-This tool can be used to abstract between the implementation of an effect and the map specific setting.
-This is a very powerful mechanism since it allows generic scenario code to work with different maps providing the needed named locations."
+• select hexes using the <ref>dst='editor_tool_select' text='select tool'</ref>
+• in the Areas menu, select Add New Area
+• then in the Areas menu, select Save Selection to Area
+• then in the Areas menu, select Rename Selected Area and choose a name for the area
+
+This tool is only available in Scenario Mode, it adds WML <italic>text='[time_area]'</italic> tags to the scenario. Although the tag’s name implies time, it is now more generic and can be used for other purposes without needing to change the time-of-day schedule in the area."
```
Pentarctagon, line 202: Should be a semicolon instead of a comma.

data/core/editor/help.cfg
```diff
@@ -122,151 +122,208 @@ Named locations can be accessed from WML using the Standard Location Filter’s
...
 # wmllint: markcheck off
+# wmllint: unbalanced-on
 [topic]
     id=editor_playlist
     title= _ "Playlist Manager"
-    text= "<img>src=icons/action/playlist_30.png align=left box=yes</img>" + _ "Saves a list of music tracks defining a random playlist to the scenario.
+    text= "<img>src=icons/action/playlist_30.png align=left box=yes</img>" + _ "Shows a list of music tracks known to the editor, with toggle-boxes to enable them.
 
-Have a look at the addon server for easy to use additional music tracks."
+This tool is only available in Scenario Mode, it adds WML <italic>text='[music]'</italic> tags to the scenario."
```
Pentarctagon, line 214: Should be a semicolon instead of a comma.

data/core/editor/help.cfg
```diff
@@ -122,151 +122,208 @@ Named locations can be accessed from WML using the Standard Location Filter’s
...
+<bold>text='Warning: the Scenario Mode is buggy in 1.14.'</bold> Improving it is one of the main blockers for the 1.16 release, however in 1.14 its quality is far below that of the game and the terrain mode." + "
 
-<header>text='What you do *not* get'</header>
+" + _ "<header>text='What you do *not* get'</header>
 
 • What-you-see-is-what-you-get
-The editor is not a WYSIWYG application.
 
-Because which exact graphic tile represents a terrain in the map depends on all terrain rules loaded (which is different between the editor and each other use case) the map won’t look exactly the same.
+The editor is not a WYSIWYG application. Because which exact graphic tile represents a terrain in the map depends on all terrain rules loaded (which is different between the editor and each other use case) the map won’t look exactly the same in the game as it does in the editor.
```
Pentarctagon, line 241: I don't think using the WYSIWYG acronym really adds anything useful here, or the "What-you-see-is-what-you-get" text above it.

data/core/editor/help.cfg
stevecotton: Ended up rewriting it as `This tool is only available in Scenario Mode; ownership information is stored by adding WML <italic>text='[village]'</italic> tags to the appropriate <italic>text='[side]'</italic>."`

If this is the biggest change needed then I'm happy.

Pentarctagon (approved):

PR #5001, opened 2020-07-22
Warn about manually editing files created by the scenario editor

stevecotton:
This comment will be added at the start of any .cfg file written by the
scenario editor. Given that the only target audience is authors using text
editors, it's not translatable and uses plain ascii characters.

This is the easy part of #1380.


gfgtdf:
looks good to me

Pentarctagon:
I say change "probably shouldn't" to just "shouldn't".

PR #5003, opened 2020-07-23
[1.14] Warn about manually editing files created by the scenario editor

stevecotton:
Backport of #5001 to the stable branch.

This comment will be added at the start of any .cfg file written by the
scenario editor. Given that the only target audience is authors using text
editors, it's not translatable and uses plain ascii characters.

(cherry picked from commit 11d702225b5999919d742997e866a777ca29a3c3)


stevecotton:
Will force-push to resolve a conflict in the changelog, wait for the build, and then merge.

PR #5005, opened 2020-07-25
Avoid prematurely generating bad color range/palette definitions

irydacea:
(PR against 1.14 because master refuses to read custom colour range/palette definitions at all right now? The forward port is trivial, though.)

Previously it was possible for the game to try to coerce any named color range/palette that hasn't been defined by WML yet into a range/palette with RGB `#000000` as its first entry, since `strtol()` (used by the `color_t::from_hex_string()` function) will return 0 if the input is not a valid hexadecimal number.

Now, normally this condition was *not* reached because `color_t::from_hex_string()` throws `std::invalid_argument` if the input
isn't exactly 6 chars (as in char, not Unicode characters) long. The problem is, in English, "yellow" is exactly 6 chars long.

The existing code suggests that we should be throwing in this case, and this is exactly what was already being done in the general case, so just do that immediately when:

 1. A named color range doesn't exist and it doesn't have at least one comma.
 2. A named color palette doesn't exist and it has characters that aren't hexadecimal digits or commas.

Reasoning:

 1. While `color_range`'s type definition in the engine suggests that only the first colour value is needed, this doesn't make much sense from the WML API standpoint so we might as well be stricter than the `color_range` constructor here.

 2. As for palettes, there's a possibility that someone may want to perform a palette substitution with exactly one colour (e.g. `~PAL(FF00FF>000000)`), so we just make sure that all characters in the "name" are hexadecimal digits. This means that using "DEADED" as a name might be a bad idea if the palette somehow finds itself being used before being defined (it will effectively create a new palette with `#DEADED` as its only entry) but it's either that or making the single colour case surprisingly clunky to write.

Ideally, we should be resetting colour info entirely (the `game_config::reset_color_info()` function) every time we enter a game,
but for *some* reason that's not done at the moment. It would also make more sense to allow `[color_range]` and `[color_palette]` definitions to be overridden under certain circumstances, but currently that's not the case (the first defined element with a given id remains a constant for the entire game session) and it's also more work to change that.

(Wow, I'm doing a terrible job at selling this commit, aren't I?)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,10 @@
    * Do not list nameless abilities in the Abilities and Ability Upgrades lists in unit
      descriptions in Help (issue #3060).
    * Fixed disabled buttons using the pressed highlight color.
+ ### WML engine
+   * Do not attempt to use names of [color_range]s as inline color range definitions if
+     they don't have commas, or names of [color_palette]s as inline palette definitions
+     if they have characters that aren't commas or hexadecimal digits.
  ### Miscellaneous and bug fixes
    * Made wmllint recognize [remove_time_area] in order to avoid spurious warnings about
      unit ids.
--- a/src/color.cpp
+++ b/src/color.cpp
@@ -60,8 +60,8 @@ color_t color_t::from_rgb_string(const std::string& c)
 
 color_t color_t::from_hex_string(const std::string& c)
 {
-	if(c.length() != 6) {
-		throw std::invalid_argument("Color hex string should be exactly 6 digits");
+	if(c.length() != 6 || c.find_first_not_of("0123456789AaBbCcDdEeFf") != std::string::npos) {
+		throw std::invalid_argument("Color string must be exactly 6 hexadecimal digits");
 	}
 
 	unsigned long temp_c = std::strtol(c.c_str(), nullptr, 16);
--- a/src/game_config.cpp
+++ b/src/game_config.cpp
@@ -530,6 +530,16 @@ const color_range& color_info(const std::string& name)
 		return i->second;
 	}
 
+	// If the name string has commas, it's an inline-defined color range. Otherwise
+	// it's the name of a color range that doesn't exist (e.g. looking at saves from
+	// an add-on that's not loaded yet).
+
+	if(name.find(',') == std::string::npos) {
+		throw config::error(_("Color range not defined: ") + name);
+	}
+
+	LOG_NG << "attempting to parse \"" << name << "\" as an inline color range definition\n";
+
 	std::vector<color_t> temp;
 	for(const auto& s : utils::split(name)) {
 		try {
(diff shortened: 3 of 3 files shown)


gfgtdf:
It looks like tc_info throws for an invalid color like `aaaaaq` but returns an empty color `aaaaaaa` (7 digits).

Why not move the `find_first_not_of("0123456789ABCDEF")` check into the from_hex_string function and make it returns std::invalid argument just like for too long strings ?

shikadiqueen:
I forgot that hex digits can be lowercase, I need to fix that.

I'm not sure how many other things use `color_t::from_hex_string()` that could potentially break if I make it stricter.

gfgtdf:
> I'm not sure how many other things use color_t::from_hex_string() that could potentially break if I make it stricter.

I think it's quite safe to assume that no code really relies  on from_hex_string interpreting invalid hex string as 0 iff they have size 6.

shikadiqueen:
Updated PR. The first commit now handles lowercase hexadecimal digits. Also added a commit making `color_t::from_hex_string()` stricter.

However, I'd like to point out that the latter is used in more places than I originally expected:

```
03:48:03 shadowm@hanacore ~/src/wesnoth-1.14 git:bugfix/yellow-color-range-1.14 % fgrep -nRI from_hex_string src
src/color.hpp:102:      static color_t from_hex_string(const std::string& c);
src/game_config.cpp:403:                                color_vec.push_back(color_t::from_hex_string(s));
src/game_config.cpp:477:                                temp.push_back(color_t::from_hex_string(s));
src/game_config.cpp:506:                                        temp.push_back(color_t::from_hex_string(s));
src/game_config.cpp:546:                        temp.push_back(color_t::from_hex_string(s));
src/game_config.cpp:577:                        temp.push_back(color_t::from_hex_string(s));
src/color.cpp:61:color_t color_t::from_hex_string(const std::string& c)
src/gui/core/canvas.cpp:1252:   , link_color_(cfg["text_link_color"], color_t::from_hex_string("ffff00"))
src/gui/widgets/label.cpp:49:   , link_color_(color_t::from_hex_string("ffff00"))
src/gui/widgets/label.cpp:259:  , link_color(cfg["link_color"].empty() ? color_t::from_hex_string("ffff00") : color_t::from_rgba_string(cfg["link_color"].str()))
src/gui/widgets/styled_widget.cpp:182:  return color_t::from_hex_string("ffff00");
```

> It looks like tc_info throws for an invalid color like `aaaaaq`(invalid character) but returns an empty for an invalid color like `aaaaaaa` (7 digits).

That's existing behaviour. Do you think it should throw in both cases?

gfgtdf:
>  That's existing behaviour. Do you think it should throw in both cases?

My idea was actually not to throw at all (if we remove the added code in `tc_info`, then it will fallback to the `} catch(const std::invalid_argument&) {` catch below because of the added code in from_hex_string), however the only reason for that opinion is that it also didn't throw before.

I mean you _can_ make it throw but then you should also make sure that codes that call tc_info can handle that (and not just crash to desktop), because tc_info didn't throw before.

CelticMinstrel:
Note that this is against 1.14, so the indicated lack of conflicts is wrong. There likely will be conflicts if merging this to master.

Issue #4724, opened 2020-01-26
Desktop notifications show literal markup


ProditorMagnus:
### Game and System Information
 - Same as #4723

### Describe the bug
Notifications show markup such as `<b>` https://i.vgy.me/bS2Ihp.png

### To Reproduce
Same as #4723

### Expected behavior
Notification does not contain `<b>`

### Screenshots
https://i.vgy.me/bS2Ihp.png

Pentarctagon:
Removed those tags since it happens on Linux as well.

PR #5006, opened 2020-07-26
Fixed markup in notifications with regex

philippalexanderhaendler:
This is a fix for the issue #4724

### DIFF
--- a/src/mp_ui_alerts.cpp
+++ b/src/mp_ui_alerts.cpp
@@ -28,6 +28,7 @@
 
 #include <string>
 #include <vector>
+#include <regex>
 
 namespace mp_ui_alerts {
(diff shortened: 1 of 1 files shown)


soliton-:
It would seem a better idea to call that function with an unformatted scenario name in the first place.

Pentarctagon:
@philippalexanderhaendler I haven't looked and the code, but is it possible to send the unformatted text to the function rather than adding the formatting and then later removing it?

philippalexanderhaendler:
@Pentarctagon This was more like a quick fix than a long term solution. Passing the unformatted text would be the way to go, you're right. I'm gonna look at a way to do this after my exams are done :)

PR #5016, opened 2020-07-30
Add lua attack_type:active_specials() api

gfgtdf:
this api can be used to query active attack and make them work 'just like mainline abilities'

This also contains a ability-related fix.

As a "testcase" this adds the new [feeding] ability implemented in lua:
> Now feeding is reconized by tagname instead of
id. This means in particular that umc authors
can use [feeding] abilities with filters now.
>
> Also [feeding] is now technicially a weapon
special, but sicne weapon speciasl can be used
in [abilities] it will work as before.
>
> Lastly [feeding] now takes a value, for example
`[feeding] value=(opponent.level)` should
increase the units max hp by the opponents level
(not tested).

### DIFF
--- a/data/core/macros/abilities.cfg
+++ b/data/core/macros/abilities.cfg
@@ -320,15 +320,17 @@ Enemy units cannot see this unit while it is in deep water, except if they have 
 #enddef
 
 #define ABILITY_FEEDING
+    # technically this is a weapon special, but like all weaponspecials it can nowdays be used in [abilities] as well.
     # Canned definition of the Feeding ability to be included in an
     # [abilities] clause.
     # This works because of data/lua/feeding.lua
-    [dummy]
+    [feeding]
         id=feeding
+        value=1
         name= _ "feeding"
         female_name= _ "female^feeding"
         description=_ "This unit gains 1 hitpoint added to its maximum whenever it kills a unit, except units that are immune to plague."
-    [/dummy]
+    [/feeding]
 #enddef
 
 #weapons specials
--- a/data/lua/feeding.lua
+++ b/data/lua/feeding.lua
@@ -14,21 +14,36 @@ on_event("die", function()
 	local u_killer = wesnoth.units.get(ec.x2, ec.y2)
 	local u_victim = wesnoth.units.get(ec.x1, ec.y1)
 
-	if not u_killer or not u_killer:matches { ability = "feeding" } then
+	if not u_killer or not u_killer.active_attack then
 		return
 	end
+
+	local feeding = u_killer.active_attack:active_specials {
+		tag = "feeding",
+		base_value = 0,
+	}
+
+	if feeding.value <= 0 then
+		return
+	end
+
 	if not u_victim or u_victim:matches { status = "unplagueable" } then
 		return
 	end
+
+	-- po: Floating text shown when a unit with the "feeding" ability gets a kill
+	local text = _(_"+$value max HP", _"+$value max HP", feeding.value)
+	local text = text:vformat { value = feeding.value }
+
 	local u_killer_cfg = u_killer.__cfg
 	for i,v in ipairs(wml.get_child(u_killer_cfg, "modifications"))do
 		if v[1] == "object" and v[2].feeding == true then
 			local effect = wml.get_child(v[2], "effect")
-			effect.increase_total = effect.increase_total + 1
-			u_killer_cfg.max_hitpoints = u_killer_cfg.max_hitpoints + 1
-			u_killer_cfg.hitpoints = u_killer_cfg.hitpoints + 1
+			effect.increase_total = effect.increase_total + feeding.value
+			u_killer_cfg.max_hitpoints = u_killer_cfg.max_hitpoints + feeding.value
+			u_killer_cfg.hitpoints = u_killer_cfg.hitpoints + feeding.value
 			wesnoth.units.to_map(u_killer_cfg)
-			wesnoth.interface.float_label(ec.x2, ec.y2, _ "+1 max HP", "0,255,0")
+			wesnoth.interface.float_label(ec.x2, ec.y2, text, "0,255,0")
 			return
 		end
 	end
@@ -37,9 +52,9 @@ on_event("die", function()
 		feeding = true,
 		T.effect {
 			apply_to = "hitpoints",
-			increase_total = 1,
+			increase_total = feeding.value,
 		},
 	})
-	u_killer.hitpoints = u_killer.hitpoints + 1
-	wesnoth.interface.float_label(ec.x2, ec.y2, _ "+1 max HP", "0,255,0")
+	u_killer.hitpoints = u_killer.hitpoints + feeding.value
+	wesnoth.interface.float_label(ec.x2, ec.y2, text, "0,255,0")
 end)
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -725,7 +725,6 @@ private:
 	};
 
 	bool perform_hit(bool, statistics::attack_context&);
-	void fire_event(const std::string& n);
 	void refresh_bc();
 
 	/** Structure holding unit info used in the attack action. */
@@ -750,6 +749,13 @@ private:
 
 		std::string dump();
 	};
+
+	/// returns true if the wml aborted the attack
+	bool fire_event(const std::string& n, unit_info& primary, unit_info& secondary);
+	/// returns true if the wml aborted the attack
+	bool fire_event(const std::string& n);
+	/// calls fire_event and aborts the attack if needed.
+	void fire_mid_attack_event(const std::string& n);
 
 	/**
 	 * Used in perform_hit to confirm a replay is in sync.
@@ -846,48 +852,47 @@ attack::attack(const map_location& attacker,
 	}
 }
 
-void attack::fire_event(const std::string& n)
+bool attack::fire_event(const std::string& n)
+{
+	fire_event(n, a_, d_);
+}
+
+bool attack::fire_event(const std::string& n, unit_info& primary, unit_info& secondary)
 {
 	LOG_NG << "attack: firing '" << n << "' event\n";
 
+	bool primary_is_attacker = &primary == &a_;
+	bool secondary_is_attacker = &secondary == &a_;
+
 	// prepare the event data for weapon filtering
 	config ev_data;
-	config& a_weapon_cfg = ev_data.add_child("first");
-	config& d_weapon_cfg = ev_data.add_child("second");
 
 	// Need these to ensure weapon filters work correctly
-	boost::optional<attack_type::specials_context_t> a_ctx, d_ctx;
+	boost::optional<attack_type::specials_context_t> first_ctx;
+	boost::optional<attack_type::specials_context_t> second_ctx;
 
-	if(a_.stats_->weapon != nullptr && a_.valid()) {
-		if(d_.stats_->weapon != nullptr && d_.valid()) {
-			a_ctx.emplace(a_.stats_->weapon->specials_context(nullptr, nullptr, a_.loc_, d_.loc_, true, d_.stats_->weapon));
-		} else {
-			a_ctx.emplace(a_.stats_->weapon->specials_context(nullptr, a_.loc_, true));
+	auto set_context = [&](config& weapon_cfg, boost::optional<attack_type::specials_context_t>& ctx, unit_info& u1, unit_info& u2, bool is_attacker) {
+		if(u1.stats_->weapon != nullptr && u1.valid()) {
+			if(u2.stats_->weapon != nullptr && u2.valid()) {
+				ctx.emplace(u1.stats_->weapon->specials_context(nullptr, nullptr, u1.loc_, u2.loc_, is_attacker, u2.stats_->weapon));
+			} else {
+				ctx.emplace(u1.stats_->weapon->specials_context(nullptr, u1.loc_, is_attacker));
+			}
+			u1.stats_->weapon->write(weapon_cfg);
 		}
-		a_.stats_->weapon->write(a_weapon_cfg);
-	}
 
-	if(d_.stats_->weapon != nullptr && d_.valid()) {
-		if(a_.stats_->weapon != nullptr && a_.valid()) {
-			d_ctx.emplace(d_.stats_->weapon->specials_context(nullptr, nullptr, d_.loc_, a_.loc_, false, a_.stats_->weapon));
-		} else {
-			d_ctx.emplace(d_.stats_->weapon->specials_context(nullptr, d_.loc_, false));
+		if(weapon_cfg["name"].empty()) {
+			weapon_cfg["name"] = "none";
 		}
-		d_.stats_->weapon->write(d_weapon_cfg);
-	}
+	};
 
-	if(a_weapon_cfg["name"].empty()) {
-		a_weapon_cfg["name"] = "none";
-	}
-
-	if(d_weapon_cfg["name"].empty()) {
-		d_weapon_cfg["name"] = "none";
-	}
+	set_context(ev_data.add_child("first"), first_ctx, primary, secondary, primary_is_attacker);
+	set_context(ev_data.add_child("second"), second_ctx, secondary, primary, secondary_is_attacker);
 
 	if(n == "attack_end") {
 		// We want to fire attack_end event in any case! Even if one of units was removed by WML.
-		resources::game_events->pump().fire(n, a_.loc_, d_.loc_, ev_data);
-		return;
+		resources::game_events->pump().fire(n, primary.loc_, secondary.loc_, ev_data);
+		return false;
 	}
 
 	// damage_inflicted is set in these two events.
@@ -896,16 +901,22 @@ void attack::fire_event(const std::string& n)
 		ev_data["damage_inflicted"] = resources::gamedata->get_variable("damage_inflicted");
 	}
 
-	const int defender_side = d_.get_unit().side();
-
 	bool wml_aborted;
 	std::tie(std::ignore, wml_aborted) = resources::game_events->pump().fire(n,
-		game_events::entity_location(a_.loc_, a_.id_),
-		game_events::entity_location(d_.loc_, d_.id_), ev_data);
+		game_events::entity_location(primary.loc_, primary.id_),
+		game_events::entity_location(secondary.loc_, secondary.id_), ev_data);
 
 	// The event could have killed either the attacker or
 	// defender, so we have to make sure they still exist.
 	refresh_bc();
+	return wml_aborted;
+}
+
+
+void attack::fire_mid_attack_event(const std::string& n)
+{
+	const int defender_side = d_.get_unit().side();
+	bool wml_aborted = fire_event(n);
 
 	if(wml_aborted || !a_.valid() || !d_.valid()
 		|| !resources::gameboard->get_team(a_.get_unit().side()).is_enemy(d_.get_unit().side())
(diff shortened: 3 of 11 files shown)


gfgtdf:
I wonder whether we should  keep the "not unplagueable" condition in feeding.lua or move it to a filter in ABILITY_FEEDING

data/core/macros/abilities.cfg
```diff
@@ -320,15 +320,17 @@ Enemy units cannot see this unit while it is in deep water, except if they have
 #enddef
 
 #define ABILITY_FEEDING
+    # technicialyl this is a weapon special, but like all weaponspecials it can nowdays be used in [abilities] aswell.
```
stevecotton, line 323: Missing a space in "as well", and there are swapped letters in "technically".

data/lua/feeding.lua
```diff
@@ -14,21 +14,33 @@ on_event("die", function()
...
+
+	if feeding.value <= 0 then
+		return
+	end
+
 	if not u_victim or u_victim:matches { status = "unplagueable" } then
 		return
 	end
+
+	local text = wesnoth.format(_"+$value max HP", { value = feeding.value})
```
stevecotton, line 34: This should probably now use wmlxgettext's feature for translatable plurals, and also have a comment for translators.

src/actions/attack.cpp
```diff
@@ -742,13 +742,18 @@ class attack
...
+		const battle_context_unit_stats* stats_;
+
 		unit_info(const map_location& loc, int weapon, unit_map& units);
 		unit& get_unit();
 		bool valid();
 
 		std::string dump();
 	};
 
+	//returns true if the wml aborted the attack
```
stevecotton, line 754: This documentation doesn't say much about the function.

Both of the places that call this function ignore the return code, which doesn't seem to fit with the documentation.

gfgtdf: > Both of the places that call this function ignore the return code, which doesn't seem to fit with the documentation

no fire_event(1) doesn't ignore it

data/core/macros/abilities.cfg
gfgtdf: will fix, thx

data/lua/feeding.lua
gfgtdf: i don't really know how to do that bu i can try

stevecotton:
Please add some unit tests. The existing tests suggest some interesting edge-cases to try: `alice_kills_bob_on_retal` and the `_levelup variations`.

gfgtdf:
> Please add some unit tests.

I was going to update the feeding unit test to update to the new feeding implementation.

> The existing tests suggest some interesting edge-cases to try: alice_kills_bob_on_retal and the _levelup variations.

I'm not sure what you mean here

data/lua/feeding.lua
gfgtdf: i tried but coudn't find a lua api for that.

stevecotton: It wasn't documented on the Wiki (now fixed), but comments starting `po:` are supported in the same way as the WML ones.
https://wiki.wesnoth.org/GettextForWesnothDevelopers#Marking_up_strings_in_Lua

Please forget my suggestion about supporting plurals. I thought some languages might need a plural of "HP",  but that seems wrong now.

stevecotton:
> I was going to update the feeding unit test to update to the new feeding implementation.

+1

> > The existing tests suggest some interesting edge-cases to try: alice_kills_bob_on_retal and the _levelup variations.
> 
> I'm not sure what you mean here

I mean, test that the right unit gets the bonus if a defender with feeding kills an attacker, and check what happens if the unit levels up at the same time.

data/lua/feeding.lua
gfgtdf: > It wasn't documented on the Wiki (now fixed), but comments starting po: are supported in the same way as the WML ones.

Thx for the help, but i I wasn't talking about the comment, i was talking about the plurals.

gfgtdf: oh ne3vermind the wiki seems to also explain that :/ , but it looks wrong? it seems to call string.format with a string that expects variable subsitution in wml style.

gfgtdf: also that wiki page is a little hard to find, i'd expect at least a ling here https://wiki.wesnoth.org/LuaWML/Display#wesnoth.textdomain (which is where i'd look for such lua api)

stevecotton: > looks wrong?

It's copied from data/lua/wml/objectives.lua , so I expect it to work.

CelticMinstrel:
Just for the record, plurals in Lua are `_("singular string", "plural string", value)`… so pretty much the same as in C++.

I haven't looked at the actual PR, maybe I will later…

gfgtdf:
so i'm quite happy with the current state, one thing i'm unsure about is whether the active_attack shodul be a 'getter property' as it is not or a function (`unit.active_attack` vs `unit:active_attack()` )  the advantage of the second one would be that a bit more flexible (for example we cann add parameter if we every need to)

CelticMinstrel:
1. So, I think active_specials should actually be two different functions rather than one function that returns totally different things based on the input. Perhaps active_specials returns the data, and special_value or similar returns the value. You could still have both of them use the same code (for example, the return_data / return_value variables could be replaced by a single template parameter, or you could use the lua_cpp API and std::bind.)
2. A third possibility for active_attack is to make it just return the index of the attack. I'm not sure if that would help for future expansion though. Can you think of any reason you'd _want_ to ever add a parameter?
3. Just to clarify, the value in the returned active specials data is the calculated value, not the formula, right?

gfgtdf:
> So, I think active_specials should actually be two different functions

I disagree, sometime you want to calculate both things at the same time, (the value and the cfgs) and in that case having two functions would mean that the engine would need to evaluate twice which specials are active and which are not.


> Just to clarify, the value in the returned active specials data is the calculated value, not the formula, right?

Its the value, i'm not sure what you mean by formula here though.

> Can you think of any reason you'd want to ever add a parameter?

Not really, but who knows? Maybe one day well support group attacks where 3 untis fights each so in each round each unit  hits every other unit, and uses different attack against each of the different  units ..... :p

CelticMinstrel:
> I disagree, sometime you want to calculate both things at the same time, (the value and the cfgs) and in that case having two functions would mean that the engine would need to evaluate twice which specials are active and which are not.

I do kinda understand your reasoning, but I think it's better to present a clean interface rather than prematurely optimize for the someone needing both pieces of information, a situation which I don't think will be very common.

Have you actually benchmarked the time it takes to calculate active specials? I can't imagine it to be significant in many situations.

Furthermore, couldn't there be a function that, given the list of specials, calculates the composite value from it without recalculating the list? So that would satisfy uncommon cases that need both.

> Its the value, i'm not sure what you mean by formula here though.

Like if you wrote `value="(opponent.level)"` as in your example, does it return the unit's level or the string `"(opponent.level)"`?

gfgtdf:
> Furthermore, couldn't there be a function that, given the list of specials, calculates the composite value from it without recalculating the list? So that would satisfy uncommon cases that need both.

This was indeed my first idea when i added this api, but the problem is that the 'ability list' class contains unsafe references so it cannot easily be exposed to lua without major refactoring.

> Like if you wrote value="(opponent.level)" as in your example, does it return the unit's level or the string "(opponent.level)"?

It returns the level, remember that the function does't return the value for a single weapon special, but the cumulative value of all active specials with that tagname. (in most cases there will probably only be one, but just like [leadership] there can for example be multiple units with such an ability that effects others.) 

> Have you actually benchmarked the time it takes to calculate active specials?

No but its clear that for this function calculating which specials are active is the main part of the work.
Furthermore calling them multiple times might have other problems, for example if you for some reason have a nondeterministic filter, (using dice operator) then calling the function twice might returns different values.


> I do kinda understand your reasoning, but I think it's better to present a clean interface rather than prematurely optimize for the someone needing both pieces of information, a situation which I don't think will be very common.

I don' think the interface is that unclean, one function returning multiple values (if they are calculated the same way) is not that uncommon.

CelticMinstrel:
> This was indeed my first idea when i added this api, but the problem is that the 'ability list' class contains unsafe references so it cannot easily be exposed to lua without major refactoring.

That's a bit different from what I meant... I meant you could have an ability_list function that returns the list, and another function that takes that list and calculates the composite value from it.

> Furthermore calling them multiple times might have other problems, for example if you for some reason have a nondeterministic filter, (using dice operator) then calling the function twice might returns different values.

Sure, but the solution to that is: don't use a nondeterministic filter.

> I don' think the interface is that unclean, one function returning multiple values (if they are calculated the same way) is not that uncommon.

It might be common in languages like C, but that doesn't make it a good design. Given a choice between…

```lua
local stuff = unit:ability_list{tag = 'feeding', value = true}
```

…and…

```lua
local stuff = unit:ability_value "feeding"
```

I don't think anyone would suggest the first is a cleaner interface. I know I would definitely prefer the second version by far. I could also settle for something like this:

```lua
local stuff = (unit:ability_list "feeding"):composite()
```

And when only one of the two forms is needed, with Lua you can two get Lua functions out of a single C++ function. I still think the case where both are needed would be uncommon, but you could always add a third Lua function that calls the same underlying C++ function if you really think that's necessary. To me, that would be vastly preferable to the API you've implemented here.

* active_abilities("tag") -- returns the list of active abilities
* ability_value("tag") -- returns the composite value of the ability
* active_abilities_with_value("tag") -- returns both list and value

Pentarctagon:
-Cross-posting a bit from Discord-

If I'm reading this correctly, if `base_value` is provided, it returns only the value attribute of the tagname; if `data` is provided, then it returns a table containing the cfg of each ability with that tagname plus the `teacher_loc` and `student_loc` attributes? And if neither are provided it does nothing?

If that's correct, then:
* it sounds like one function that's doing two separate things (returning the value of the first matching tagname vs returning the full contents of all matching tagnames plus a couple other attributes), which IMO isn't ideal since it's an additional detail to need to remember, especially since UMC authors aren't usually coders themselves.  So having two functions that each do their specific thing would be cleaner, unless there's a reason to need to have both pieces of functionality be part of the same function.
* more generally than just here as well, if the function requires one or more arguments to do anything at all, and it's possible to know with certainty if those arguments are missing, then a warning should be displayed. So if `data` and `base_value` are both missing, then it should be treated the same as the tagname being missing.
* if nothing else, `base_value` and `data` should be more descriptive, like `return_value` and `return_abilities`.

gfgtdf:
> That's a bit different from what I meant... I meant you could have an ability_list function that returns the list, and another function that takes that list and calculates the composite value from it.

That's exactly what i meant, but  as i said, you cannot easily create a lua object that wraps an ability_list (c++) object without refactoring, since the C++ ability_list contains raw pointers to other (c++) (config) objects, and those might be freed during the lifetime of the lua ability_list object.


> `local stuff = (unit:ability_list "feeding"):composite()`

As i said above: you cannot implement a function like `ability_list` without refactoring.

> It might be common in languages like C, but that doesn't make it a good design

I disagree, especially as soon as a function has a lot of ( in particular multiple optional) arguments taking a single table is much cleaner than taking a list of arguments. And this api was designed so that we are able to add more parameter later if we should need them: For example some people might want to get specials by id instead of tag. Also people might want a parameter to set the context.




> And if neither are provided it does nothing?

This part is wrong, It still calculates the mount of abilities that matches, (in particular  whether one matched at all), in some cases this is all you need (boolean abilities like poison/slow).

> if nothing else, base_value and data should be more descriptive, like return_value and return_abilities.

base_value is used for a base value of the ability though, `return_value` makes it sound like its a boolean.

> it sounds like one function that's doing two separate things ..

these two things, well see the discussion before on this topic, basicially these things are for the most part the result of the same calculation, so calculating then separately would be wasted calculation time.

Pentarctagon:
> these two things, well see the discussion before on this topic, basicially these things are for the most part the result of the same calculation, so calculating then separately would be wasted calculation time.

Are the calculations expensive enough that doing them once vs twice matters, though?  And relatedly, does returning just the count and the base_value really need to be its own functionality, rather than just adding it to each ability returned along with `teacher_loc` and `student_loc`?

gfgtdf:
> Are the calculations expensive enough that doing them once vs twice matters

Yes these calculations are usually the majority of the calculation time of that function.


> rather than just adding the base_value as an attribute to each ability returned along with teacher_loc and student_loc?

I really don't think this make any sense, this value is a propertly of the whole 'ability list' not foa  single ability tag.,

Pentarctagon:
To try putting it a different way, my main concern here isn't with the API itself, but with how it's being presented to UMC authors.  If I'm understanding correctly, and based on:
> It returns the level, remember that the function does't return the value for a single weapon special, but the cumulative value of all active specials with that tagname. (in most cases there will probably only be one, but just like [leadership] there can for example be multiple units with such an ability that effects others.)

When the attribute `base_value` is provided, it returns the sum of all the `base_value`s for matching weapon specials/abilities, and returns it as the `value` attribute?

gfgtdf:
> When the attribute base_value is provided, it returns the sum of all the base_values for matching weapon specials/abilities, and returns it as the value attribute?

No it returns the 'value' of an given ability (tag), see https://wiki.wesnoth.org/AbilitiesWML#Common_keys_and_tags_for_specials_with_a_value,

Pentarctagon:
Given the name `base_value` though, I was thinking it meant the same thing as would be affected by [filter_base_value].

gfgtdf:
> Given the name base_value though, I was thinking it meant the same thing as is used by [filter_base_value].

Yes it does.

gfgtdf:
by the ability tag have no base_value propertly, also nothing is summed, if you have two [chance_to_hit] specials with value=70 in each your unit also won't get a chance_to_hit of 140 .

Pentarctagon:
> > Given the name base_value though, I was thinking it meant the same thing as is used by [filter_base_value].
> Yes it does.

I'm really confused, then.  [filter_base_value] doesn't filter on the `value` attribute, it filters on the value of whatever is being modified by the `value` attribute.  For example:
```
[damage]
    id=a
    name="a"
    description=""
    value=1
    multiply=2
    [filter_base_value]
        greater_than=15
    [/filter_base_value]
[/damage]
```
would only be active on units whose original attack deals 15 or more damage.

gfgtdf:
> I'm really confused, then. [filter_base_value] doesn't filter on the `value` attribute, it filters on the value of whatever is being modified by the `value` attribute. For example:
> 
> ```
> [damage]
>     id=a
>     name="a"
>     description=""
>     value=1
>     multiply=2
>     [filter_base_value]
>         greater_than=15
>     [/filter_base_value]
> [/damage]
> ```
> 
> would only be active on units whose original attack deals 15 or more damage.

Yes that's true, what exactly is your question ?

Pentarctagon:
Does `base_value` return the value of the attack or the value of the [damage]?

> > Given the name base_value though, I was thinking it meant the same thing as is used by [filter_base_value].
> 
> Yes it does.
sounds like the former, but
> No it returns the 'value' of an given ability (tag), see https://wiki.wesnoth.org/AbilitiesWML#Common_keys_and_tags_for_specials_with_a_value,
sounds like the latter.

gfgtdf:
> ```
> [damage]
>     id=a
>     name="a"
>     description=""
>     value=1
>     multiply=2
>     [filter_base_value]
>         greater_than=15
>     [/filter_base_value]
> [/damage]
> ```

Your example is very uncommon since it'd first set the value to 1 and then multiplay it by 2 so and attack with dealing more than 15 damage would deal 2 damange. You coudl aswell just write `value=2`

> Does base_value return the value of the attack or the value of the [damage]?

You give the function the base_value (`atttack;active_specials { tag = "damange, base_value= 15}`) and it will calculate the (modified) value for you. (`atttack;active_specials({ tag = "damange, "value= 17}).value`) should return 2 for if it has a the special you just wrote

Pentarctagon:
Ah, alright, that makes more sense.

Pentarctagon:
My last question then is regarding:

> And this api was designed so that we are able to add more parameter later if we should need them: For example some people might want to get specials by id instead of tag. Also people might want a parameter to set the context.

Will the API stay clean in the future if additional parameters are added?  It's not difficult to imagine it getting messy if several years from now there end up being a variety of different parameters that can be passed, especially if they're all technically optional.

gfgtdf:
> Will the API stay clean in the future if additional parameters are added

Well if we want to add something later, we will then think about what will be the best way to add it. Maye we will then come to the conlusion that it better to add another function.

> That's exactly what i meant, but as i said, you cannot easily create a lua object that wraps an ability_list (c++) object without refactoring, since the C++ ability_list contains raw pointers to other (c++) (config) objects, and those might be freed during the lifetime of the lua ability_list object.

So i thought it about it some more, i think the refactoring  needed isn't that big, we just need to swicth unit unit::abilities_ type from `config` to somethign like `vector<shared_ptr<config>>` and make ability_list hold shared (or weak) pointers. I'm still unsure though, whether we should do it that way, since it'd somehow make it harder to change things in the c++ code later if we expose it too much to lua.

Pentarctagon:
> Well if we want to add something later, we will then think about what will be the best way to add it. Maye we will then come to the conlusion that it better to add another function.

I think part of what CelticMinstrel was getting at as well was that it would be good to consider that now so that we're confident the API will remain consistent, rather than potentially having this function for retrieving some information and then later deciding additional related information should be retrieved by separate functions.

gfgtdf:
What i said was a general comment that applies to add api, i currently see no other information that we could want to return.

CelticMinstrel:
> That's exactly what i meant, but as i said, you cannot easily create a lua object that wraps an ability_list (c++) object without refactoring, since the C++ ability_list contains raw pointers to other (c++) (config) objects, and those might be freed during the lifetime of the lua ability_list object.

I never said anything about wrapping the C++ ability list. Why are you so hung up on that? My suggestion was that the Lua ability list could be used to calculate the same thing.

> I disagree, especially as soon as a function has a lot of ( in particular multiple optional) arguments taking a single table is much cleaner than taking a list of arguments.

This isn't false, but it doesn't change that the particular arguments your proposed API _does_ take aren't good.

> And this api was designed so that we are able to add more parameter later if we should need them: For example some people might want to get specials by id instead of tag. Also people might want a parameter to set the context.

Sure, but that's all beside the point, which is that the parameters that your proposed API already takes _are not good_.

> these two things, well see the discussion before on this topic, basicially these things are for the most part the result of the same calculation, so calculating then separately would be wasted calculation time.

1. This is premature optimization, unless you've benchmarked it.
2. Even if it's a reasonable argument, you can get past it without having the function return different data depending on its parameters.

If the data returned is different, the function should be different. From what you've said so far, it sounds like there's actually _three_ different things it returns, so there should be _three_ different functions, possibly four.

* One returning just a boolean
* One returning the value
* One returning the list
* Maybe one returning both the value and the list

> You give the function the base_value (`atttack;active_specials { tag = "damange, base_value= 15}`) and it will calculate the (modified) value for you. (`atttack;active_specials({ tag = "damange, _basevalue= 17}).value`) should return 2 for if it has a the special you just wrote

So now I'm confused. Why is `base_value` a number? Shouldn't the function take the base value from the unit?

src/units/attack_type.cpp
```diff
@@ -259,6 +259,10 @@ bool attack_type::matches_filter(const config& filter) const
 	return matches;
 }
 
+bool attack_type::is_active() const
+{
+	return self_ || other_ || self_loc_ != map_location::null_location() || other_loc_ != map_location::null_location();
```
stevecotton, line 264: What does is_active mean? There's no documentation, and from the name I would expect this to check whether it's an "offense-only" attack, a "defense-only" attack, etc.

Pentarctagon:
Closing given the various remaining open questions and gfgtdf no longer being around to answer them.

PR #5020, opened 2020-08-01
Make sure to reset the logging level to the default in tests that change the logging level.

### LINKED ISSUE #4898: Boost unit tests logging changes when LTO is enabled
After adding the following code to the start of the function [here](https://github.com/wesnoth/wesnoth/blob/6eeab88c5e448051a60fd27ccd40a6e627274c6a/src/gui/auxiliary/iterator/policy_order.hpp#L79):
```
if(lg::debug().dont_log(gui2::log_gui_iterator))
{
    std::cout << "debug not enabled" << std::endl;
}
```

[This](https://github.com/wesnoth/wesnoth/files/4605285/noLTO.log) is the output of the non-LTO run, and [this](https://github.com/wesnoth/wesnoth/files/4605286/LTO.log) is the output for the LTO run.

Requires GCC 9, since with GCC 7 the boost unit tests segfault.

Pentarctagon:
Fixes #4898

### DIFF
--- a/src/tests/gui/iterator.cpp
+++ b/src/tests/gui/iterator.cpp
@@ -276,4 +276,6 @@ BOOST_AUTO_TEST_CASE(test_gui2_iterator)
 
 	test_control();
 	test_grid();
+
+	lg::set_log_domain_severity("all", lg::warn());
 }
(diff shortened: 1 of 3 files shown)


gfgtdf:
looks good to me

PR #5021, opened 2020-08-01
An add-on's min_version should be unset to indicate the latest version is needed.

### LINKED ISSUE #5004: MP min add-on version logic incorrect.
Relevant comment [here](https://github.com/wesnoth/wesnoth/pull/4884#discussion_r460077853).  Essentially add-ons with no minimum version specified should default to requiring the latest version of the add-ons.

Creating this issue so it will be fixed separately from #4884, since it's an existing bug.

Pentarctagon:
This changes the min_version logic to:
1) Unset the min_version if the add-on already exists with a set min_version.
2) Only check and potentially update the min_version if the add-on already has a min_version set.

Fixes #5004

### DIFF
--- a/src/mp_game_settings.cpp
+++ b/src/mp_game_settings.cpp
@@ -180,38 +180,44 @@ void mp_game_settings::update_addon_requirements(const config & cfg) {
 		return;
 	}
 
-	mp_game_settings::addon_version_info new_data(cfg);
+	mp_game_settings::addon_version_info new_addon_data(cfg);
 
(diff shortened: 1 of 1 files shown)


gfgtdf:
I think this is not 100% correct, of the previous addon  (`addon`) had required=false, it should be overwritten by `new_data` even though it has `addon.min_version` unset.

Pentarctagon:
I think the pushed update should cover that.

gfgtdf:
I think this shoudl work, but looks complicated let me try to find an alternative patch.

gfgtdf:
made an alternative patch: #5022

PR #5022, opened 2020-08-02
alterntivive min_addon_version fix

### LINKED ISSUE #5004: MP min add-on version logic incorrect.

gfgtdf:
Fixes #5004

Now we have the following invariant: addon.min_version is set if and only if addon.required.

### DIFF
--- a/src/mp_game_settings.cpp
+++ b/src/mp_game_settings.cpp
@@ -182,11 +182,21 @@ void mp_game_settings::update_addon_requirements(const config & cfg) {
 
 	mp_game_settings::addon_version_info new_data(cfg);
 
+	
+	if(!new_data.required) {
+		new_data.min_version = boost::none;
+	}
+	else if(new_data.required && !new_data.min_version) {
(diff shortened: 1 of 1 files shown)


Pentarctagon:
Wouldn't this unset the min_version if the last add-on item being processed isn't required?

gfgtdf:
> Wouldn't this unset the min_version if the last add-on item being processed isn't required?

Hmm i don't think so, the only code that sets calls `addon.min_version` in case of `new_data.required`

Pentarctagon:
I'll go ahead and fix this up and merge an updated PR if there are no issues raised with it.

Pentarctagon:
Re-done in #5998

PR #4856, opened 2020-04-18
Add a new attribute that checks calls special_bool for weapon specials abilities were filtered like weapon specials

newfrenchy83:
This PR resolves https://github.com/wesnoth/wesnoth/issues/4840 by adding two new attributes to [filter_weapon] for special weapons implemented in [abilities] is also filtered.

Issue #4840, opened 2020-04-06
a way to query active specials for a weapon


gfgtdf:
With the newly added weapon speciasl as abilities feature in 1.15, the special_tag_active attribute does not longer return the same value that is used in the actual attack calculation to check whether a weapon is really under the effect of an ability making it harder to implement weapon specials in wml properly, We should either
1) make the implementation of special_tag_active call the special bool function
Or
2) Add a new attribute that checks calls special_bool

Im currently unsure which of these is the best way though.

newfrenchy83:
i already tried to resolve that but yes, only the true weapon_specials were identified by special_tag_active, probably because weapon-special must be implemented inside [specials] tag for that.

gfgtdf:
Wee so what it your opinion on changing this line 

https://github.com/wesnoth/wesnoth/blob/e053566a15287d14b3ac8fb0ec05ddc2bfc7e910/src/units/attack_type.cpp#L207


to call attack->ability_bool() instead ?

newfrenchy83:
i already try that, but if id or tag are not in [special] tags, it is undectectable. for filter_weapon or filter_attack, it is only the wml code wrting inside [attack] who can be filtered, the remain code inside unit_type can't be filtered.

I tried to emulate get_special_boll in ability_bool but without changes.

The ability like weapon_special can't be filtered like true weapon_specials, and i search an alternate to [filter_weapon] for resolve that but in best case, it will be partial.

newfrenchy83:
i think what i did findest an solution but i must test before.

newfrenchy83:
@gfgtdf i resoved this issue with add a new attribute that checks calls special_bool

gfgtdf:
I worked on something that fits _my own_ usecase  better in https://github.com/wesnoth/wesnoth/pull/5016 basicialyl we add a lua api function that allows me to not only check whether any ability/special with a certain tagname is active but also to to query information about them. This only supports querying abilities by tagname (not by id) yet, but i don't think this is a problem.

newfrenchy83:
> 
> 
> I worked on something that fits _my own_ usecase better in #5016 basicialyl we add a lua api function that allows me to not only check whether any ability/special with a certain tagname is active but also to to query information about them. This only supports querying abilities by tagname (not by id) yet, but i don't think this is a problem.

except what if [filter_student/opponent/eetc..], apply_to or active_on were used and condition not matches, weapon abilities will be seen like  active

PR #5023, opened 2020-08-02
extent special_id_active and special_type active to abilities weapon and add two attributes for [specials] only

newfrenchy83:
it is the same resolve of https://github.com/wesnoth/wesnoth/issues/4840 but with the number of commit in https://github.com/wesnoth/wesnoth/pull/4856 and the errrored build, i prefer put a new PR


gfgtdf:
> i prefer put a new PR

I todl you this alrady, but: don't do this, learn how to qualy multipel commits into one usualyl usng `git rebase -i`

newfrenchy83:
@gfgtdf could you review this PR, please?

newfrenchy83:

CelticMinstrel:
The schema changes in the diff look suspect…

newfrenchy83:
> The schema changes in the diff look suspect…

could you be more explicit?

src/units/abilities.cpp
```diff
@@ -1146,13 +1168,188 @@ unit_ability_list attack_type::get_special_ability(const std::string& ability) c
...
+				}
+			} else if(special_id && !special_tags){
+				if ( get_special_children_id(special_id_matches, it->abilities(), special) ) {
+					return true;
+				}
+			} else if(!special_id && special_tags){
+				if ( get_special_children_tags(special_tag_matches, it->abilities(), special) ) {
+					return true;
+				}
+			}
```
kabachuha, line 1238: @newfrenchy83 can you please put this piece of code in a separate function? The same code as here is used four times in the PR.

src/units/abilities.cpp
```diff
@@ -1146,13 +1168,188 @@ unit_ability_list attack_type::get_special_ability(const std::string& ability) c
...
+					abil_list.append((*self).get_abilities(entry.tag_name, self_loc_));
+					for(unit_ability_list::iterator i = abil_list.begin(); i != abil_list.end();) {
+						if(!special_active(*i->ability_cfg, AFFECT_SELF, entry.tag_name, true, "filter_student")) {
+							i = abil_list.erase(i);
+						} else {
+							++i;
+						}
+					}
+				}
+			}
```
kabachuha, line 1252: This section also contains a heavily duplicated code.

src/units/abilities.cpp
```diff
@@ -1146,13 +1168,188 @@ unit_ability_list attack_type::get_special_ability(const std::string& ability) c
...
+		return true;
+	}
+	return false;
+}
+
+bool attack_type::bool_ability(const std::string& ability, bool simple_check, bool special_id, bool special_tags) const
+{
+	bool abil_bool = get_special_bool(ability, simple_check, special_id, special_tags) || get_special_ability_bool(ability, special_id, special_tags);
 	return abil_bool;
 }
```
kabachuha, line 1354: Why assigning the result to a variable if you can return it directly?

CelticMinstrel:
> could you be more explicit?

Not really? It's adding some new accepted keys and they look like keys that should not be accepted. I have no idea what they do or what they're for or why they're named the way they're named. I haven't looked at the PR in detail for the record, just noticed that the schema portion seems to suggest there's something that (at best) needs work.

newfrenchy83:
@CelticMinstrel these keys are equivalent to special_id_active and special_type_active used in [filter_weapon] and animation [filter_attack] (but not in [event][filter_attack]), but appplied to abilities equivelent weapon and no to regular weapon specials, and are added to shema to same manner what originals keys.
For the names, perhaps could you suggest other names what the actuals, i know what i what i am not easy for the names.

CelticMinstrel:
Okay, so let me get this straight:

* `special_type_active` is true if the weapon being used has a `[specials]` subtag of the given type
* `wp_ability_type_active` is true if the unit is under the influence of an `[abilities]` subtag of the given type that affects the weapon being used.

Is this accurate?

If that's the situation, then I'd like to know if there's a reason to distinguish between those two situations. Are there cases where you want wp_ability_type_active but not special_type_active? I see you added some uses of wp_ability_type active; would they still work if they checked either wp_ability_type_active or special_type_active? Like so:

```
wp_ability_type_active=thing
[or]
  special_type_active=thing
[/or]
```

Similarly, are there any cases you can imagine where you'd want to check special_type_active but exclude wp_ability_type_active?

newfrenchy83:
> 
> 
> Okay, so let me get this straight:
> 
>     * `special_type_active` is true if the weapon being used has a `[specials]` subtag of the given type
> 
>     * `wp_ability_type_active` is true if the unit is under the influence of an `[abilities]` subtag of the given type that affects the weapon being used.
> 
> 
> Is this accurate?
> 
Yes
> If that's the situation, then I'd like to know if there's a reason to distinguish between those two situations. Are there cases where you want wp_ability_type_active but not special_type_active? I see you added some uses of wp_ability_type active; would they still work if they checked either wp_ability_type_active or special_type_active? Like so:
> 
> ```
> wp_ability_type_active=thing
> [or]
>   special_type_active=thing
> [/or]
> ```
> 
> Similarly, are there any cases you can imagine where you'd want to check special_type_active but exclude wp_ability_type_active?

The principal reason is what if i want use special_type_active for both specials and abilities subtags, the game crash. same if it not case to exclude wp_ability_type_active, i do not choice what to using two attributes for the game function.

src/units/abilities.cpp
```diff
@@ -1111,11 +1111,33 @@ namespace { // Helpers for attack_type::special_active()
 
 unit_ability_list attack_type::list_ability(const std::string& ability) const
 {
+	const unit_map& units = display::get_singleton()->get_units();
+
+	unit_const_ptr self = self_;
+	unit_const_ptr other = other_;
```
CelticMinstrel, line 1117: Is there a reason to copy these pointers? If it's just because you don't like the trailing underscore, then I'd say that's not a good enough reason.

src/units/abilities.cpp
```diff
@@ -1146,14 +1168,168 @@ unit_ability_list attack_type::get_special_ability(const std::string& ability) c
...
+			if ( &*it == self.get() )
+				continue;
+
+			if(get_ability_children(special_tag_matches, special_id_matches, it->abilities(), special, special_id , special_tags)){
+				return true;
+			}
+		}
+		if(special_tags){
+			for(const special_match& entry : special_tag_matches) {
+				if(entry.tag_name !="heals" && entry.tag_name !="regenerate" && entry.tag_name !="skirmisher" && entry.tag_name !="teleport" && entry.tag_name !="hides"){
```
CelticMinstrel, line 1242: I think it would be nicer to put this list of excluded tags into a `std::set` and then just call `excluded_tags.count(entry.tag_name) == 0`.

CelticMinstrel (commented):
> The principal reason is what if i want use special_type_active for both specials and abilities subtags, the game crash. same if it not case to exclude wp_ability_type_active, i do not choice what to using two attributes for the game function.

Hmm, okay, that's interesting. Why does it crash, exactly? I do think it would be better to just use the existing keys if possible, unless someone has a use-case for keeping them separate.

src/units/abilities.cpp
newfrenchy83: @CelticMinstrel it is because https://github.com/wesnoth/wesnoth/commit/f12d51a2f9af13e5edd4c44a32b2ee8c0ea15862 of @gfgtdf , but i don't know if is fix list_ability.

src/units/abilities.cpp
newfrenchy83: it is done

src/units/abilities.cpp
newfrenchy83: it is fixed

newfrenchy83: the copy of pointer was for is compatibility with https://github.com/wesnoth/wesnoth/pull/5016/commits/589780410797425176719572bbc85ce7d09b0698

kabachuha:
@newfrenchy83 can you please make a rebase for some of your commits. Commit messages like `update file_XXX.cpp` are not descriptive at all. Also, as @gfgtdf noted, you can squash some of them together not to fill the history with numerous, but insignificant changes. And please get rid of `Merge branch 'master'`, use a `rebase` and force-push instead. If you don't know how to use commandline git for some reason there is a quite comprehensive guide on the Wesnoth wiki.
https://wiki.wesnoth.org/Git_for_Wesnoth_Crash_Course

newfrenchy83:
> 
> 
> @newfrenchy83 can you please make a rebase for some of your commits. Commit messages like `update file_XXX.cpp` are not descriptive at all. Also, as @gfgtdf noted, you can squash some of them together not to fill the history with numerous, but insignificant changes. And please get rid of `Merge branch 'master'` commits, use a `rebase` and force-push instead when you want to keep up with the upstream. If you don't know how to use commandline git for some reason there is a quite comprehensive guide on the Wesnoth wiki.
> https://wiki.wesnoth.org/Git_for_Wesnoth_Crash_Course

Alas the first step, clone my fork in local is difficualt fore me because my net conexion is 1M/s in the best case.

CelticMinstrel:
We can always hit the Squash button on GitHub, as long as it's not something super-complicated that really _should_ be spread across multiple commits (which this doesn't seem like it is).

I'd prefer if `excluded_tags` was declared as `static const`.

Also, what's the current situation on the WML attributes? Is it still impossible to merge `wp_ability_` and `special_` tags? I still dislike splitting them in this way… is there actually a use-case for it, I wonder…? (That last question is for everyone really, not just newfrenchy.)

newfrenchy83:
> Also, what's the current situation on the WML attributes? Is it still impossible to merge `wp_ability_` and `special_` tags? I still dislike splitting them in this way… is there actually a use-case for it, I wonder…? (That last question is for everyone really, not just newfrenchy.)

Alas, yes, it is still impossible to merge wp_ability and special_ tags, if i suse bool_ability instead to get_special_bool, the game crash, if i change "wp_ability_" to "special_" then only the regulars specials are filtered. For the use case, the initiative abilities could be an example, i want to insert an leadership for call an leader_anim when initiative is active; in this case, i don't want what iitiative was active if unit using a regular firststrike attack, and i exclude all firstrike special_type.

CelticMinstrel:
The fact that merging them makes the game crash suggests that something else is wrong. I would think there's another way to prevent the crash.

stevecotton:
@newfrenchy83 I'd like to try debugging the crash. Which is the most recent commit in this PR that crashes in that way (alternative, please would you push a separate branch that shows the crash), and what are the steps to reproduce it?

newfrenchy83:
> 
> 
> @newfrenchy83 I'd like to try debugging the crash. Which is the most recent commit in this PR that crashes in that way (alternatively, please would you push a separate branch that shows the crash), and what are the steps to reproduce it?

I find how merge attribute without crash the game, i pull that in new PR or in this PR?

newfrenchy83:
In fact, no, it's not fixed.

newfrenchy83:
> 
> 
> @newfrenchy83 I'd like to try debugging the crash. Which is the most recent commit in this PR that crashes in that way (alternatively, please would you push a separate branch that shows the crash), and what are the steps to reproduce it?

i create a PR https://github.com/newfrenchy83/wesnoth/pull/31 to describe the change for merging the tattributes for what you test the condition of crash,

in Httt, in debug mode, create Princess unit in side enemy of your side, an Elvish Champion to same side of Princess, and place the cursor on Champion and the game crah.

stevecotton:
The crash is an infinite recursion with the following callstack:
* (anonymous namespace)::special_unit_matches
* attack_type::special_active_impl
* attack_type::impl_check_abilities
* attack_type::check_abilities
* attack_type::get_special_ability_bool
* attack_type::bool_ability
* matches_simple_filter
* attack_type::matches_filter
* attack_type::matches_filter

My feeling is that separating wp_ability_id_active and special_id_active isn't the solution - it prevents the crash with a specific pair of units, but what stops the crash with another pair of units?

newfrenchy83:
> 
> 
> The crash is an infinite recursion with the following callstack:
> 
>     * (anonymous namespace)::special_unit_matches
> 
>     * attack_type::special_active_impl
> 
>     * attack_type::impl_check_abilities
> 
>     * attack_type::check_abilities
> 
>     * attack_type::get_special_ability_bool
> 
>     * attack_type::bool_ability
> 
>     * matches_simple_filter
> 
>     * attack_type::matches_filter
> 
>     * attack_type::matches_filter
> 
> 
> My feeling is that separating wp_ability_id_active and special_id_active isn't the solution - it prevents the crash with a specific pair of units, but what stops the crash with another pair of units?

But then why, when i use separate attributes, it prevent the crash and play leading_anim, but never with special_ same if i clear regular special_ of matches_simple_filter?

newfrenchy83:
@CelticMinstrel and @stevecotton , and yet  best if i replace both "wp_ability_id_active" and "special_id_active" by "bspecial_id_active", the leading-anim is played without crash nor bugs. Is the "special" word in attribute name who is cause of conflict?

newfrenchy83:
But if i want test a regular speical with new attribute name, the filter don't matches. The code function with attribute name different only, and if i using "special_" for name of attribute for abilities, the game crash.

newfrenchy83:
@CelticMinstrel , @stevecotton 
that's it, I figured out what was wrong.
When I merged attributes with bool_ability, I forgot to remove
[filter_student]
[filter_weapon]
[not]
special_type_active = firststrike

initiative ability, which causes it to filter itself, causing the game to endlessly loop and crash.
This incident is for me the proof that the attributes must remain separate, I had implemented this filter to exclude special weapons of the firstrike type so that leading_anim is only triggered when using an attack benefiting from the 'initiative' ability and not the one already having firststrike.
if the attributes are merged then I will have to forgo the use of leadership animations (leading_anim or helper_anim) for aura type abilities applied to weapons, and there may be other cases in the future where separation will be necessary.


Could you retest attribute merge without [filter_student]
[filter_weapon]
[not]
special_type_active = firststrike
to check? Thank you.

stevecotton:
> Could you retest attribute merge without [a WML-only change] to check?

Sorry, but my feeling is that it shouldn't merge like this. If we know that the C++ can crash, and all it takes to trigger it is a custom unit with some WML, then there's still a C++ bug that needs to be fixed. In fact, we should probably add a unit test based on that WML.

newfrenchy83:
> 
> 
> > Could you retest attribute merge without [a WML-only change] to check?
> 
> Sorry, but my feeling is that it shouldn't merge like this. If we know that the C++ can crash, and all it takes to trigger it is a custom unit with some WML, then there's still a C++ bug that needs to be fixed. In fact, we should probably add a unit test based on that WML.

Except what the bugs exist also in regular weapon specials, like when you using  [filter_self]
[filter_weapon]
[not]
special_type_active = firststrike
in firststrike weapon special, the game crash also. this problem already there before what i work on abilities used like weapon,

newfrenchy83:
@CelticMinstrel @stevecotton could you review https://github.com/newfrenchy83/wesnoth/pull/32 please before what i decide to merge in this PR?

Pentarctagon:
If there's a bug that's causing a crash, the crash should be fixed rather than adding new WML attributes to work around the crash.

newfrenchy83:
> 
> 
> If there's a bug that's causing a crash, the crash should be fixed rather than adding new WML attributes to work around the crash.

when an weapon-special exclude himself via [filter_weapon], we have an infinite recursive, how doing for escape to the loop?

And create an option for select abilities or specials only is not only for work around the crash same if i accept to merge attribute in special_ but also for wml develloper are the choice to not filter abilities and specials systematicelly but have the choice.

CelticMinstrel:
From an API standpoint, I definitely think this is preferable, though I'm still confused at why you'd _want_ the ability to exclude specials or abilities in this filter.

Can we rename `exclude_tags` to `special_exclude_tags` to make it more obvious that the two keys are related?

I think the crash still needs to be fixed _somehow_, however.

stevecotton:
Thinking about how to fix the crash: we could put a hard limit on the number of levels of recursion.

* What to do when that limit is hit. Hardcode "yes", "no" or a WML error?
* What should the limit be? Do we try and work out a worst case, or just choose a huge limit?
    * When I was testing the crash, the stack was (IIRC) about 2000 recursions deep, so simply choosing 100 as the hard limit might be reasonable.
    * Thinking of possible worst cases, I can think of at least 10 units being involved: UtBS's formation ability, combined with an attack ability that works similarly.

Pentarctagon:
Maybe I'm misunderstanding the problem, but if the infinite recursion is happening when a weapon filter excludes the weapon it's on, shouldn't it stop there? It's already excluded.

newfrenchy83:
> 
> 
> Maybe I'm misunderstanding the problem, but if the infinite recursion is happening when a weapon filter excludes the weapon it's on, shouldn't it just stop there? It's already excluded.

I think the same thing but @stevecotton think what other case could exist.

kabachuha (changes requested):
Hello, @newfrenchy83 !

I understand your difficulties with the command line git, but I can't stand the current set of the commit names, so I made a rebase of your pr repository on my fork and all that's left is to merge it with your `patch-13` branch.

How I changed your commits (only names, code is the same) [rebased branch on the fork](https://github.com/kabachuha/wesnoth/tree/fix-newfrenchy):

  ```
	reword a6fd636ea70 Add a new attribute that checks calls special_bool for weapon specials abilities were filtered like weapon specials
	squash bfd0941fb3e Update weapon.cfg
	squash 3674ff563bb Update animation-utils.cfg
	squash e1b6847a1fc add abilities for call anim for weapon like abilities
	squash de809a8f428 leading_anim can now called for initiative ability
	squash e8db90e0334 Update abilities.cpp
	squash 30d5cdfea31 Update attack_type.cpp
	squash 0bafcaf242e Update attack_type.hpp
	pick 0fca6dbf3d2 add new attribute to udisplay unit_attack
	reword 776f5bb3d7b Update unit.hpp
	squash 41c034db659 fix attack_type::list_ability and attack_type::get_special_ability_bool
	squash 0209ad7681e fix shema validation
	squash e46cc8454a1 fix after https://github.com/wesnoth/wesnoth/pull/4985
	fixup 0f66ca562e4 fix udisplay
	fixup 71d35e3fe85 fix abilities.cpp
	reword ac174637ebf Update animation-utils.cfg
	fixup 7ac1a9fda05 Fix indentation in animation-utils
	pick a12963403b7 add static get_ability_children
	reword 304c6f979a1 add case where wp_ability_type_active is exlude
	reword 40007fa78ec using exluded_tags list
	pick 58406ba9d90 remove code unused
	fixup ff9de1453f5 Update abilities.cpp
	reword 03580a10856 ix an error
	reword dbd72534af5 Create function chek abilities (#30)
	squash 0ade133e81d use 'static' for excluded_tags declaration
	pick 4db8cc5df5b add special_only and  ability_only options
	reword 962cd8ff94d add [filter_specials] and [filter_abilities] (#34)

	# Rebase b8f03c40e6f..962cd8ff94d onto b8f03c40e6f (27 commands)
	#
	# Commands:
	# p, pick <commit> = use commit
	# r, reword <commit> = use commit, but edit the commit message
	# e, edit <commit> = use commit, but stop for amending
	# s, squash <commit> = use commit, but meld into previous commit
	# f, fixup <commit> = like "squash", but discard this commit's log message
	# x, exec <command> = run command (the rest of the line) using shell
	# d, drop <commit> = remove commit
	# l, label <label> = label current HEAD with a name
	# t, reset <label> = reset HEAD to a label
	# m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]
	# .       create a merge commit using the original merge commit's
	# .       message (or the oneline, if no original merge commit was
	# .       specified). Use -c <commit> to reword the commit message.
	#
	# These lines can be re-ordered; they are executed from top to bottom.
	#
	# If you remove a line here THAT COMMIT WILL BE LOST.
	#
	#	However, if you remove everything, the rebase will be aborted.
	#
	#	
	# Note that empty commits are commented out

  ```

Some commits were reworded as following:

  ```
	add a new attribute checking that special_bool calls for weapon specials abilities are filtered like weapon specials

	add new attribute to udisplay unit_attack

	add a way to query unit abilities

	use a second attack filter in helping anim

	add static get_ability_children

	add a case when wp_ability_type_active is excluded

	use a list of excluded tags

	remove code unused

	fix a display error

	create an abilities check function

	add special_only and ability_only options

	add [filter_specials] and [filter_abilities]

  ```

However, GitHub doesn't allow the pull requests to have force-push commits in them, so you'd have to use a little command line to merge the changes — alas, ç'est la vie — an adequate development workflow is virtually impossible without it.

I do understand your problems with the bandwidth — I, too, have around 1 MiB/s of speed and today I had my internet connection go off like five times.

Luckily for us, git has some command line options that allows you to reduce the network and storage usage _dramatically_. Firstly, you can clone only *one* branch selected with a `--single-branch` option. Then you can cut off the unneeded history with a shallow copy option that allows cloning only the latest N commits: `--depth N`.

2020-10-29-131523_816x590_scrot

If you look at my screen, you may notice that the total download size has been reduced to only 463 MiB — whereas the usual wesnoth clone size is around 3-4 gigabytes.

So, I suggest you to try the following with your git (based on [What should I do if I’m in a bad situation?](https://stackoverflow.com/questions/9646167/clean-up-a-fork-and-restart-it-from-the-upstream):

0. Clone your repo with the network optimization list (I think 300 commits should be enough, but you can estimate it yourself):

```sh
git clone --single-branch --branch patch-13 --depth 300 https://github.com/newfrenchy83/wesnoth.git
```

1. Add my fork with the rebase of your branch as an remote source (wesnoth is still `upstream`)

```sh
git remote add kabachuha-fix https://github.com/kabachuha/wesnoth.git
```

2. Fetch the commits history from it:

```sh
git fetch kabachuha-fix
```

3. Override your patch branch with my rebase branch:

```sh
git reset --hard kabachuha-fix/fix-newfrenchy
```

4. Finally, force push your remote branch on GitHub:

```sh
git push origin patch-13 --force
```

5. Have a cake while enjoying the embellished commit history :) !

If you have any questions or in case something goes wrong, please, write here or DM me on the forums (forum username is `kabachuha`). I really appreciate if you read this text and if you will put at least some of it into practice. After all, remember — bad internet is not an excuse to neglect the workflow.

newfrenchy83:
@kabachuha it is done

kabachuha (dismissed):
Looks fine to me.

stevecotton:
@newfrenchy83 I have the same question as @kabachuha - why are you keeping both this and #5239 open?

newfrenchy83:
> 
> 
> @newfrenchy83 I have the same question as @kabachuha - why are you keeping both this and #5239 open?

https://github.com/keepwesnoth/wesnoth/pull/5239 is the same except what [filter_specials/abilities] don't implemented, @CelticMinstrel was septical for utility of these tags or discrimination of specials/abilities source, i keep both open because i don't know what choice(with or without tags) will be doing.

stevecotton:
> #5239 is the same except what [filter_specials/abilities] don't implemented

Oh, I see. Git can make that a lot easier: have a commit specifically for that change, and then your two branches can share all of their commits except for that one.

A history visualisation tool is helpful. Git's command-line tools normally come with one called `gitk`, you can see all of the branches with `gitk --all &`.

First get both of your branches to start from the same point. When @kabachuha did the rebase for you, he updated it to start from b8f03c40e6f25ad0fd2ae3c6c1d7360c76270ef0, which is more recent than where the patch-14 branch started from. So checkout patch-14 and `git rebase b8f03c40e6f25ad0fd2ae3c6c1d7360c76270ef0`.

Now compare the two branches, `git diff patch-14 patch-13` - you'll see the small amount of code for the filter_specials/abilities, as well as some changes that you've made to one branch but forgotten to make to the other, such as renaming the first argument of get_special_bool().

This is also a good time to sort out the branch names :D With patch-14 checked out, let's create a new branch named "separate_specials_filter" and check that out now. Then:

    git diff --name-only patch-13
The output should show that all the changes are in the src/ and data/ folders

    git checkout patch-13 src/ data/
Sets the contents of those folders to be what's in patch-13. This overwrites all of your local changes, but as the two branches are so similar that's exactly what we want.

    git commit data/campaigns/Heir_To_The_Throne/ data/schema/filters/weapon.cfg src/units/attack_type.cpp -m 'Use separate filters for specials and abilities'
Makes a single commit with the changes between the two branches (excluding the bits that I think we should exclude).

You can now either work on the branch without the separate filter (in which case you can rebase the separate filter on top of your work), or you can work on the branch with the separate filter (in which case interactive-rebase will let you later reorder the commits so the the separate-filter one is the last one). Either way, you only have to do each change once instead of copying between branches.

You can then push the whole branch, and mention in the PR that it can be reviewed both without and without the controversial commit.

### Code review:

Having reviewed just the difference between the branches, it seems too small for all of changes that are specific to separate filters to be in that diff. In other words, I think both branches contain code that should only be in the separate-filter branch.

### Notes:

There's also a `git reflog`, for finding your old work if all of this branch-hopping goes wrong.

When I've got a big PR like this one which isn't getting approved, I use a similar trick but separate out the bits which improve the already-merged code without adding new features, and rebase to make that the first commit on my branch. I open a PR just for that first commit, wait for it to merge, and then repeat the process because each time the controversial PR gets smaller.

This whole long text must have already been written somewhere, but my web-searching skills couldn't find it among the other pages about git-rebase.

newfrenchy83:
@CelticMinstrel i have removed [filter_abilities] of the code

newfrenchy83:
@CelticMinstrel  [filter_abilities] removed and buildi passed.

newfrenchy83:
@CelticMinstrel @Pentarctagon i request an advise, what it is the best for you, using a [filter_specials] tags or using special_only_id/type_active?

Pentarctagon:
Is there a single summary somewhere of exactly what this is doing and why?  It's been long enough and the above comments are confusing enough that I'm not sure.

newfrenchy83:
> 
> 
> Is there a single summary somewhere of exactly what this is doing and why? It's been long enough and the above comments are confusing enough that I'm not sure.

Normal special weapons have filters that allow like special_id_active to use them as trigger criteria for another weapon or event (like filter_attack in unit animation), the idea here is to extend special_id / type_active to abilities used as special weapons for example to call a leading_anim with a firststrike ability (Httt), however I want to be able to filter the normal specials independently using either a [filter_specials] tag or a special_only_id / type_active (for example a chance_to_hit ability has a filter that excludes normal chance_to_hit but not those contained in [abilities] to avoid an infinite loop or to avoid leading_anim being called by specials with the same id as the ability associated with the animation).

Finally, these attributes unlike special_id / type without the activity filter, cannot be used in [event] [filter_attack]

Pentarctagon:
> however I want to be able to filter the normal specials independently using either a [filter_specials] tag or a special_only_id / type_active (for example a chance_to_hit ability has a filter that excludes normal chance_to_hit but not those contained in [abilities] to avoid an infinite loop or to avoid leading_anim being called by specials with the same id as the ability associated with the animation).

What's the use case for this?  Whether an attack has a [firststrike] (or any other special) applied by an ability to all attacks or by a weapon special to a specific attack, the end result is the same - the weapon is considered as having that special applied.

newfrenchy83:
> What's the use case for this? Whether an attack has a [firststrike] (or any other special) applied by an ability to all attacks or by a weapon special to a specific attack, the end result is the same - the weapon is considered as having that special applied.

Let us take the example of a so-called anti-firstrike ability or weapon, i.e. a firststrike which is triggered only in the presence of an opponent benefiting from a firststrike type ability but not a weapon. special, if the id's are different which will probably be the case then special_id_active will be sufficient on the other hand if the id's are the same then it is necessary to distinguish the sources, with
[filter_opponent]
special_id_active = firststrike
[not]
[filter_specials]
special_id_active = firststrike
or
[not]
[filter_specials]
special_type_active = firststrike

or when faced with a large number of abilities or weapons of the same type the WML developer may want to use special_type_active while limiting or excluding the usual special.

These cases can be very rare but for me it justifies maintaining this option of discrimination between the two sources because there can always be situations where filtering the two sources can cause problems.

Pentarctagon:
> Let us take the example of a so-called anti-firstrike ability or weapon, i.e. a firststrike which is triggered only in the presence of an opponent benefiting from a firststrike type ability but not a weapon special.

Right, I'm asking why anyone would want to do that.  It would result in the anti-firststrike ability being used against units with the firststrike ability and not used against units with the firststrike weapon special, which just sounds confusing, especially to players who don't know these sorts of details.

newfrenchy83:
> Right, I'm asking why anyone would want to do that. It would result in the anti-firststrike ability being used against units with the firststrike ability and not used against units with the firststrike weapon special, which just sounds confusing, especially to players who don't know these sorts of details.

The ability will be type leadership(affect adjacent allies) and the'anti-firststrike affect the communication teacher-student, but not an unit who can use firststrike by itself without teacher help, and then for what this unit don't affected by 'anti-firstrike' if he was under 'firststrike leadership' in same time it will be necesssary to add 

[filter_opponent]
special_id_active=firstrike # ability affected
[not]
[filter_specials]
special_type_active= firststrike # unit with special and ability in same time don't affected

Pentarctagon:
I don't really see the difference to be honest.  Even if the unit is getting firststrike from a leadership-like ability instead of having firststrike by itself, it still has firststrike.

newfrenchy83:
> 
> 
> I don't really see the difference to be honest. Even if the unit is getting firststrike from a leadership-like ability instead of having firststrike by itself, it still has firststrike.

Imagine that we can scramble the exchanges between the leader and his student, in this case the student, deprived of his leader's instructions, cannot anticipate the attack as would a unit which could do it by himself, the antifirstrike blocks these exchanges so that the target does not see the attack coming before receiving or dodging the first hit

newfrenchy83:
@Pentarctagon perhaps you must review https://github.com/wesnoth/wesnoth/pull/5239 instead, this is the same except what [filter_specials] don't implemented

Pentarctagon:
> Imagine that we can scramble the exchanges between the leader and his student, in this case the student, deprived of his leader's instructions, cannot anticipate the attack as would a unit which could do it by himself, the antifirstrike blocks these exchanges so that the target does not see the attack coming before receiving or dodging the first hit

That seems like such a rare use case that I don't think it needs built in support, especially since there are already ways to accomplish the same thing.

> @Pentarctagon perhaps you must review #5239 instead, this is the same except what [filter_specials] don't implemented

I'd much rather you create a single PR for a single thing that's being done and make updates to that one PR when changes are needed, instead of there being similar but somewhat different versions of the code in multiple PRs.  I mean this just in general, not specifically for these PRs.

Pentarctagon:
Closing this one in favor of #5239 then.

PR #5024, opened 2020-08-02
Fix SDL2 2.0.12 change in SDL_GetDisplayDPI

### LINKED ISSUE #5019: [macOS] Support or workaround SDL2 2.0.12 change in SDL_GetDisplayDPI function
SDL2 2.0.12 changed `SDL_GetDisplayDPI` function for macOS. It now returns DPI multiplied by screen's scale factor.

Visual diff of change (SDL2 2.0.10 on the left and SDL2 2.0.12 on the right):

@Pentarctagon found this additional info about when this change happened in SDL's mercurial repository:
```
changeset: 13313:3b03741c0095 to src/video/cocoa/SDL_cocoamodes.m
commit message:
cocoa: SDL_GetDisplayDPI() should account for Retina displays.

Fixes Bugzilla #4856.

bug: https://bugzilla.libsdl.org/show_bug.cgi?id=4856
```

Problem is reproducible only in `master` branch and only with retina macOS.

Until better solution [I'll workaround](https://github.com/hrubymar10/MacCompileStuff/commit/8d3c9e0820029cdd38990df2db5cca787d284830) this issue by patching SDL2 library provided with [MacCompileStuff](https://github.com/hrubymar10/MacCompileStuff). I'll simply revert this change.

hrubymar10:
As always, I want to ask somebody for code review. I am not C++ master but I tried my best.

Also I have to note that this code is probably correct but I cannot tell it for 100%. As always, `it works for me`.

Also wait for travis CI builds before merging.

Fixes #5019

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -639,6 +639,7 @@ for env in [test_env, client_env, env]:
         env.Append(FRAMEWORKS = "Cocoa")            # Cocoa GUI
         env.Append(FRAMEWORKS = "Security")         # commonCrypto (after OpenSSL replacement on Mac)
         env.Append(FRAMEWORKS = "IOKit")            # IOKit
+        env.Append(FRAMEWORKS = "CoreGraphics")     # CoreGraphics
 
 if not env['static_test']:
     test_env.Append(CPPDEFINES = "BOOST_TEST_DYN_LINK")
--- a/projectfiles/Xcode/The Battle for Wesnoth.xcodeproj/project.pbxproj
+++ b/projectfiles/Xcode/The Battle for Wesnoth.xcodeproj/project.pbxproj
@@ -91,7 +91,6 @@
 		46685CA4219D53E90009CFFE /* tokenizer.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 46685C7B219D51890009CFFE /* tokenizer.cpp */; };
 		46685CA5219D57F60009CFFE /* battery_info.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 46181DD12119F74C00908BC2 /* battery_info.cpp */; };
 		46685CA6219D63530009CFFE /* apple_battery_info.mm in Sources */ = {isa = PBXBuildFile; fileRef = 46181DCE2119F73A00908BC2 /* apple_battery_info.mm */; };
-		46685CA7219D63960009CFFE /* IOKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 46BCC904211A1E68003BDB8C /* IOKit.framework */; };
 		46685CA8219D63FA0009CFFE /* preprocessor.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 46685C6E219D51870009CFFE /* preprocessor.cpp */; };
 		46685CA9219D63FE0009CFFE /* binary_or_text.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 46685C72219D51880009CFFE /* binary_or_text.cpp */; };
 		46685CAA219D64050009CFFE /* parser.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 46685C74219D51880009CFFE /* parser.cpp */; };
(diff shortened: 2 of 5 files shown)


Pentarctagon:
As long as this builds, and you can confirm it works with unmodified SDL 2.0.10 and 2.0.12, I'm fine with it.

shikadiqueen:
Minor thing: we generally do not raise the indentation level within C++ `namespace` blocks since it leads to nonsense like this:

```
namespace ns1 {
    namespace ns2 {
        namespace ns3 {
            /* 90% of a 5,000 lines long file at indentation level 3 */
        }
    }
}
```

AI0867:
The version check checks that all three version components are greater than or equal to the corresponding components of 2.0.12
That will work for patch releases, but not for minor or major releases.

I believe wesnoth has a proper version class now (game_version.hpp) that you can uses for comparisons.

hrubymar10:
> As long as this builds, and you can confirm it works with unmodified SDL 2.0.10 and 2.0.12, I'm fine with it.
@Pentarctagon Yes, it works with both. It has runtime check for SDL version


> Minor thing: we generally do not raise the indentation level within C++ `namespace` blocks since it leads to nonsense like this:
> 
> ```
> namespace ns1 {
>     namespace ns2 {
>         namespace ns3 {
>             /* 90% of a 5,000 lines long file at indentation level 3 */
>         }
>     }
> }
> ```
@shikadiqueen I just copy&pasted structure from existing file. I see your point there but if we want to "fix" that, then we should fox that everywhere.


> The version check checks that all three version components are greater than or equal to the corresponding components of 2.0.12
> That will work for patch releases, but not for minor or major releases.
> 
> I believe wesnoth has a proper version class now (game_version.hpp) that you can uses for comparisons.
@AI0867 I don't follow. If there will be for example SDL v3.1.2 then major version is 3, which is greater then 2 so whole if becames true and it will apply the patch.

Regarding to the `game_version.hpp`. I know about that componnent but AFAIK it requires version in string format which SDL doesn't have and I thought that it will be unnecesarry complicated to format it using for example printf.

shikadiqueen:
> @shikadiqueen I just copy&pasted structure from existing file. I see your point there but if we want to "fix" that, then we should fox that everywhere.

I am aware that there are a few places that do indent namespaces, but almost all new code sdoesn't.

Pentarctagon:
> I don't follow. If there will be for example SDL v3.1.2 then major version is 3, which is greater then 2 so whole if becames true and it will apply the patch.
> 
> Regarding to the game_version.hpp. I know about that componnent but AFAIK it requires version in string format which SDL doesn't have and I thought that it will be unnecesarry complicated to format it using for example printf.

It wouldn't evaluate to true for 3.1.2, since it would check 3 > 2 (true) and 1 > 0 (true) and 2 > 12 (false).

`game_version.hpp` also has a [constructor that takes ints](https://github.com/wesnoth/wesnoth/blob/5ba26b914384ea2c1ccbb6a18fb9692f71295dbc/src/game_version.hpp#L51) (the last two arguments have default values and so are optional).

hrubymar10:
> > I don't follow. If there will be for example SDL v3.1.2 then major version is 3, which is greater then 2 so whole if becames true and it will apply the patch.
> > Regarding to the game_version.hpp. I know about that componnent but AFAIK it requires version in string format which SDL doesn't have and I thought that it will be unnecesarry complicated to format it using for example printf.
> 
> It wouldn't evaluate to true for 3.1.2, since it would check 3 > 2 (true) and 1 > 0 (true) and 2 > 12 (false).
> 
> `game_version.hpp` also has a [constructor that takes ints](https://github.com/wesnoth/wesnoth/blob/5ba26b914384ea2c1ccbb6a18fb9692f71295dbc/src/game_version.hpp#L51) (the last two arguments have default values and so are optional).

Oh I see 😅 I meant to use or`||` instead of and `&&`

Also now I see that wesnoth's `game_version` will be better.

Pentarctagon:
I'll merge this in the next day or two unless anything else comes up then.

PR #5025, opened 2020-08-02
Various minor graphics fixes/updates

doofus-01:
For now includes
1.  Adult version of scorpion portrait, so the scorpion and scorpling don't share a portrait
2.  Revised naga dirkfang portrait, in anticipation of LordBob's completion of the unit line
3.  dwarf castle floor transitions (only needed if mixed with non-dwarf castles)


doofus-01:
LordBob is OK with the scorpion and dirkfang now, so I'm declaring this finished.

Issue #5030, opened 2020-08-05
Game fails to show JPG images


DisherProject:
### Game and System Information
 - What version of the game are you running? 
I tried both 1.14 and 1.15.3
 - Where did you download it from? 
I built it from source
 - If you're using a custom build, what compiler did you use, and what commit did you build at?
I used the VC2019 project with vcpkg, as described [here](https://github.com/wesnoth/wesnoth/blob/master/projectfiles/VC16/README.md)
 - What OS are you running?
Windows 10

### Describe the bug
The game successfully loads all the images, but it seems unable to display the JPG ones. This also causes Wesnoth to crash at the end of the first story part of TRoW.

### To Reproduce
Just start the game: you'll see the background looking like this:

![Cattura](https://user-images.githubusercontent.com/29802612/89444457-020e4080-d752-11ea-990c-5798d7214574.PNG)

You can also try starting TSG or TRoW, and you'll notice that no pictures are shown during the story. On the other hand, in AToTB everything is fine.

### Additional context
Here is some information from the logger. Only the JPG images seem to be problematic here.
```
20200805 19:25:17 debug filesystem: Looking for 'maps/background.jpg'.
20200805 19:25:17 debug filesystem:   checking 'C:\Users\Stefano\Documents\My Games\Wesnoth1.15/images/'
20200805 19:25:17 debug filesystem:   checking 'C:\Users\Stefano\Documents\wesnoth/images/'
20200805 19:25:17 debug filesystem:   checking 'C:\Users\Stefano\Documents\My Games\Wesnoth1.15/data/core/images/'
20200805 19:25:17 debug filesystem:   checking 'C:\Users\Stefano\Documents\wesnoth/data/core/images/'
20200805 19:25:17 debug filesystem:   found at 'C:\Users\Stefano\Documents\wesnoth/data/core/images/maps/background.jpg'
20200805 19:25:17 debug filesystem:   checking 'C:\Users\Stefano\Documents\My Games\Wesnoth1.15/'
20200805 19:25:17 debug filesystem:   checking 'C:\Users\Stefano\Documents\wesnoth/'
20200805 19:25:17 debug filesystem:   not found
20200805 19:25:17 info filesystem: Streaming C:\Users\Stefano\Documents\wesnoth/data/core/images/maps/background.jpg for reading.
20200805 19:25:17 error display: could not open image 'maps/background.jpg'
[...]

PR #5031, opened 2020-08-05
Add JPG support to VC++2019 build

DisherProject:
Related to #5030 

I have updated the list of the dependencies and the Visual Studio project, so that they also include `sdl2-image[libjpeg-turbo]:x64-windows`. This further library solved the issue.


Pentarctagon:
It looks like travis failed since it hadn't built and cached the jpeg library, so I've cleared the cache and trigger a manually run.  Once that's done I'll re-run the failed job here, and at that point I'm expecting it'll pass.

Pentarctagon:
@DisherProject You'll need to update the dependencies [here](https://github.com/wesnoth/wesnoth/blob/master/utils/travis/steps/script.sh#L28) as well.

DisherProject:
Thank you, i missed that

Pentarctagon:
As a note for the future: having for example `fixes #5030` in your commit message will automatically close #5030 when the PR is merged.

PR #5038, opened 2020-08-07
Campaignd storage rework and incremental updates/uploads

kabachuha:
# Campaignd storage rework and incremental update packs

Hello, Wesnoth devs! The goal of this pull request is to solve multiple problems considering the `campaignd` (addons server) as well as to lay a path for some new useful features.

## Addon storage rework
Firstly, it divides the addon storage monofile into a set of folders containing the metadata, full installation packs and incremental update packs. That allows to save disk writing usage because only addons with modified data will be written when the server saves its configuration. Also, the runtime metadata model has changed from list (`config`) to hash-map (`std::unordered_map`) in order to guarantee a constant time access on average instead of checking every element's name. Moreover, that allows storing multiple addon versions, so the client will be able to grab the latest version satisfying its `min_wesnoth_version` (see @gfgtdf 's proposal) or migrate to a previous version in case of breaking changes. (_see future plans_) Old addons are to be converted on the first server setup.

The new addon storage structure on server:

```
server.cfg <- server configuration
data/ADDON_NAME_1/
data/ADDON_NAME_1/addon.cfg <- add-on title, icon, description, translations info, etc. (metadata) for the current version, including information on past versions and their packs
data/ADDON_NAME_1/update_pack_HASHHASHHASHHASHASH1.gz <- cached update pack from version X to version X+1
data/ADDON_NAME_1/update_pack_HASHHASHHASHHASHASH2.gz <- cached update pack from version X to version X+2
data/ADDON_NAME_1/update_pack_HASHHASHHASHHASHASH3.gz <- cached update pack from version X+2 to version X+3
data/ADDON_NAME_1/full_pack_HASHHASHHASHHASHASH4.gz <- the full WML pack at version X, no metadata
data/ADDON_NAME_1/full_pack_HASHHASHHASHHASHASH5.gz <- the full WML pack at version X+1, no metadata
data/ADDON_NAME_1/full_pack_HASHHASHHASHHASHASH6.gz <- the full WML pack at version X+2, no metadata
data/ADDON_NAME_1/full_pack_HASHHASHHASHHASHASH7.gz <- the full WML pack at version X+3, no metadata
```

The addon's metadata config gained two children: version and update_pack. There can be multible `[version]` and `[update_pack]` tags in the same metadata.

At the moment they consists of these attributes:

```
[version]
	filename="/full_pack_HASHHASHHASHHASHASH4.gz"
	version="0.1.1"
[/version]
[update_pack]
	expire= # the expiration timestamp. Outdated update packs will be removed and replaced with the new ones if they are still needed.
	filename="/update_pack_HASHHASHHASHHASHASH3.gz"
	from="0.1.0" # the previous version
	to="0.1.1" # the new version
[/update_pack]
```

The `[version]` tag may be supplied with `min_wesnoth_version` attribute later (see _future plans_ and @gfgtdf).

## Incremental update packs

When a new addon upload is received by server and it's addon has already been published, the server will generate an update pack from the previous version to the new one (and from this new version to the next one if we're uploading an interim version, it's also possible). The update pack contains a list of files to be removed from the local version and a list (old style addon archive) of files to be installed. The client is now sending the version of the local addon if it's present and the desired version (if it is empty or not found, the latest version will be selected). Then the server will send a chain of addended updatepacks for each version gap (for example, if the client has version 0.1.0 and server has 0.1.1 and 0.1.2, and the client wants 0.1.2, we'll send 0.1.0->0.1.1 and 0.1.1->0.1.2 update packs combined). If it failed to find the initial version (or it was not specified like in the case with old clients), the server will send a full pack.

The justifications for incremenal updates is simple as said in an excerpt from https://wiki.wesnoth.org/NotSoEasyCoding:

```
Right now, whenever a new add-on version is published, players must download (or conversely, upload) the whole contents, even if the differences between both versions amount to a single line change. This is not a problem with small add-ons, but there is a substantial number of large (> 10 MiB) add-ons on the server. Large downloads and uploads are inconvenient for players on slow or metered connections, and also increase the server's traffic requirements. 
```

## Incremental upload packs
Now it's also possible to upload addons the same incremental way. When a client wants to update an addon on the server, the game request the list of file hashes from it and then forms the upload pack by checking the difference between the checksums. That addon hash list is generated during the upload process and then it's directly sent to the client. When the pack arrives on the server it is integrated into the update pack sequence and the next version full pack is formed. This mechanism will lower both the network and the CPU usage for the server.

## Future plans

This PR is generally completed (I'm looking for a review), but it can be adjusted now or with the future PR. I perceive the next steps as following:

- Add client side changes detection using hash verification with a client-side generated checksum on download or with a hash request from the server. If the changes has been detected, we can throw a warning message box.
- ~~Add partial **uploads** via client-generated (but server-validated) update packs (will require adding hash request from the server to the protocol).~~ (Done)
- Add client side version selection based on its `min_wesnoth_version` attribute (see @gfgtdf 's min version proposal in the roadmap).
- Add a gui list of avaible addon versions.
- Attach a changelog ?
- Cosmetic stuff

I think this rework may be added to the `1.16` roadmap along with @gfgtdf 's suggestion.


gfgtdf:
Soo how about the first obvious question:  is disk usage a problem? looking at http://add-ons.wesnoth.org/1.14/ there is for example the addon " The Ravagers" which is 60mnb of size and gas 35 uploads which would then already be 2bg if we store all version.

Pentarctagon:
Currently the campaignd VM is using 28 GB out of 39 GB allocated to it.  The server as a whole (campaignd, wesnothd, database, website, etc) is only using 990 GB out of 2.7 TB, so I assume there wouldn't be an issue with giving campaignd more disk space.  Some could also potentially be reclaimed by only keeping the latest versions of add-ons uploaded to archived instances (aka 1.10 and earlier currently).

Would this incremental storage create problems for the units database (units.wesnoth.org)?

gfgtdf:
So what is the client ui currently like? Can the client chose a version to download or are the old versions only kept to create the diffs?

Also if the client can choopse a averison, is there an option for an addon author to disabel that for certain version, in particular when a version is really buggy and you just don't want anyone to downlaod them?

kabachuha:
The client gui was not changed in the current version: it just tries to download an update for the addon and receives the update packs if it's found, or the full pack otherwise. The version selection option exists in the request form, and I'm planning to implement the selection gui if what we have now will work well enough. Your suggestions are quite good, though, probably they will work together with the full-pack population control.

gfgtdf:
> it just tries to download an update for the addon and receives the update packs if it's found, or the full pack otherwise

Wait i thought the update pack will be generated 'on demand' ? If not, when will the update pack be generated ?

kabachuha:
The update packs are generated on upload.

gfgtdf:
How many upload packs are generated? just one (from previous version to the current  one)?

kabachuha:
Yes, from the previous version to the current one if we're sending the newest version, or from the prev. version to the sent version and from the sent version to the next one if the version number is somewhere in the middle.

gfgtdf:
> and from the sent version to the next one if the version number is somewhere in the middle.

Okay i don't think we should talk too much about the rare cases that the uploaded version had an older version numbr than the one on the server.

> Yes, from the previous version to the current one

Hmm okay sounds good, i was fearing it'd generate is for every previous version to the uploaded one which could be slow.

gfgtdf:
I mean an alternative approach woudl be to don't generate the update packs when we upload a addon but instead when we download it (generate it or use it if it already exists) and then remove old ones maybe after one month. But i don't think this really matter in most cases.

So from what you said it looks good will look at the code soon

kabachuha:
Later, the client generated upload packs will be made possible, so it will speed up things.

src/server/campaignd/server.cpp
```diff
@@ -668,29 +753,110 @@ void server::handle_request_campaign_list(const server::request& req)
...
+						config update_pack;
+						filesystem::scoped_istream in = filesystem::istream_file(campaign["filename"].str() + pack["filename"].str());
+						read_gz(update_pack, *in);
+						if(update_pack) {
+							pack_data.append(update_pack);
+							size += filesystem::file_size(campaign["filename"].str() + pack["filename"].str());
+						} else {
+							WRN_CS << "Unable to create an update pack sequence from version (" << from << ") to ("
+								   << to << ") for the addon '" << req.cfg["name"].str() << "'. A full pack will be sent instead!\n";
+							pack_data = config::get_invalid();
```
gfgtdf, line 814: This looks  wrong, assigning the invalid config to a config with just raise an error, also that break will just scape form the  for loop not form the outer while  loop

src/server/campaignd/addon_utils.cpp
```diff
@@ -152,4 +153,74 @@ void add_license(config& cfg)
...
+			}
+		}
+	}
+
+	for(const config& d : to.child_range("dir")) {
+		const config& origin_dir = from.find_child("dir", "name", d["name"]);
+		if(origin_dir) {
+			config& dir = pack.add_child("dir");
+			write_difference(dir, origin_dir, d, with_content);
+		}
```
gfgtdf, line 197: this looks liek it'll ignore newly added directories.

src/server/campaignd/server.cpp
```diff
@@ -898,30 +1064,113 @@ void server::handle_upload(const server::request& req)
...
 		data["name"] = "";
-		data["campaign_name"] = (*campaign)["name"];
-		data["author"] = (*campaign)["author"];
-		data["description"] = (*campaign)["description"];
-		data["version"] = (*campaign)["version"];
-		data["timestamp"] = (*campaign)["timestamp"];
-		data["original_timestamp"] = (*campaign)["original_timestamp"];
-		data["icon"] = (*campaign)["icon"];
-		data["type"] = (*campaign)["type"];
-		data["tags"] = (*campaign)["tags"];
```
gfgtdf, line 911: why did these get removed?

kabachuha: To not duplicate the addon meta in the full packs, and this data is never used by the client.

src/server/campaignd/server.cpp
kabachuha: Ok, I'll test it and fix the loop with a check in the end/start, probably.

src/server/campaignd/addon_utils.cpp
kabachuha: Will probably fix it by feeding an empty dir if the directory is not found.

src/server/campaignd/server.cpp
gfgtdf: hmm, maybe it better to duplicate  this data for robustness though (so that if the config file gets lost one still knows what these files are by looking inside them) not sure. @shikadiqueen an opinion on this?

src/server/campaignd/addon_utils.cpp
kabachuha: Done.

gfgtdf: I'm wondering, instead of calling write_difference again, wouldn't it have teh same effect to just add the whole of d? `pack.add_child("dir", d);` (instead of calling write_difference)


Also It seems like this cannot delete directories: if the funciton is called with with_content=no then, it will detect that all files here are 'missing' but it probably  won't be able to delete the directory.

src/server/campaignd/server.cpp
```diff
@@ -811,14 +812,14 @@ void server::handle_request_campaign(const server::request& req)
...
-							pack_data = config::get_invalid();
+							failed = true;
 							break;
 						}
 					}
 				}
 			}
 
-			if(pack_data && !pack_data.empty()) {
+			if(!failed && pack_data && !pack_data.empty()) {
```
gfgtdf, line 822: It makes no sense  to check `pack_data` here, since `pack_data` is an actual object instead of a reference this is never `false`. (the bool operator of the config class compares the this pointer to the config class, )

src/server/campaignd/addon_utils.cpp
kabachuha: The first point is worth considering, indeed.

The list of files to be removed is written in a `removelist` with `with_content=no`.  If the directory after purging the files is empty, it gets removed:
https://github.com/wesnoth/wesnoth/blob/3764ac90322a883a2d6910a89707744100c26c6b/src/addon/manager.cpp#L318-L320

gfgtdf: >  If the directory after purging the files is empty, it gets removed:

hmm so it impossible to upload addons with empty directories ?

kabachuha: They may be possible in a full pack, but if a directory was containing files, but as a result of an update pack it got empty (the purge check goes across the update pack removelist), it will be deleted.

kabachuha: Also, the proposal to just append the whole directory if it's new, won't work correctly in the case of `with_content=no` when we need to make a list of files without content.

src/server/campaignd/server.cpp
kabachuha: Very well, removed in https://github.com/wesnoth/wesnoth/pull/5038/commits/44135f3930d7d381eb6c3769d21e2985736861ad .

src/server/campaignd/addon_utils.cpp
gfgtdf: > They may be possible in a full pack, but if a directory was containing files and as a result of an update pack it got empty (the purge check goes across the update pack removelist), it will be deleted

So technicially the addon manager might delete empty directories unexpectedly, i don't have a problem with this since i don't see a use-case for empty directories, but its definitely a change that should be documented.

kabachuha: Fine, I can something about it on the wiki after the changes are out.

src/server/campaignd/server.cpp
shikadiqueen: This is a really annoying one because we need to explicitly duplicate almost all of the metadata. It's true there's *some* value in keeping the metadata around (I used it in combination with server logs and a previous backup to restore the add-ons server to a working state a couple of years ago after a major accident), but for the purpose of this PR I find it's an inconvenient distraction. Maybe we can evaluate re-adding it at a later time with cleaner code.

src/server/campaignd/addon_utils.cpp
shikadiqueen: There's nowhere in the wiki where people who need to know about this (who??) would see it, really. It's not important anyway.

If this involves `#ifhave` or the equivalent Lua thing, perhaps just drop a note somewhere in there telling people not to rely on the existence of empty directories.

AI0867: If they're using git for version control, empty directories aren't saved either, so that should reduce the incidence of the problem.

src/server/campaignd/server.cpp
```diff
@@ -668,29 +753,111 @@ void server::handle_request_campaign_list(const server::request& req)
...
+			full_pack += version->second["filename"].str();
+		} else {
+			WRN_CS << "The selected version (" << to << ") for the addon '" << req.cfg["name"].str()
+				   << "' has not been found or it is unspecified! Sending the latest version instead!\n";
+			to = version_map.rbegin()->first;
+			full_pack += version_map.rbegin()->second["filename"].str();
+		}
+
+		// Negotiate an update pack if possible
+		if(!from.empty() && version_map.count(version_info(from)) != 0) {
```
shikadiqueen, line 788: If `version_map` is a `std::map`, the result of `count()` can only be 1 or 0. Please use `!empty()` for clarity.

src/server/campaignd/server.cpp
```diff
@@ -898,30 +1065,113 @@ void server::handle_upload(const server::request& req)
...
 		{
-			filesystem::atomic_commit campaign_file(filename);
+			filesystem::atomic_commit campaign_file(filename + version_cfg["filename"].str());
 			config_writer writer(*campaign_file.ostream(), true, compress_level_);
 			writer.write(data);
 			campaign_file.commit();
 		}
 
-		(*campaign)["size"] = filesystem::file_size(filename);
+		(*campaign)["size"] = filesystem::file_size(filename + version_cfg["filename"].str()) + filesystem::file_size(filename + "/addon.cfg");
```
shikadiqueen, line 1101: There's basically no point to this change since the `size` attribute is primarily used to inform the user of the add-on download size and the size of the metadata is both negligible and irrelevant, since the client receives the full metadata in the `[request_campaign_list]` response instead.

This should've been in a separate commit since it's both orthogonal to the actual commit ("Split addons meta from the server config") and completely changes existing behaviour.

src/server/campaignd/server.cpp
```diff
@@ -220,6 +222,73 @@ void server::load_config()
...
+	// Convert all legacy addons to the new format on load
+	if(cfg_.has_child("campaigns")) {
+		WRN_CS << "Old format addons have been detected in the config! They will be converted to the new file format! ";
+		config& campaigns = cfg_.child("campaigns");
+		WRN_CS << campaigns.child_count("campaign") << " entries to be processed.\n";
+		for(config& campaign : campaigns.child_range("campaign")) {
+			const std::string& addon_id = campaign["name"].str();
+			const std::string& addon_file = campaign["filename"].str();
+			if(get_addon(addon_id)) {
+				ERR_CS << "The addon '" << addon_id
```
shikadiqueen, line 250: This is an abnormal state condition. At this point the server needs to die with an error instead of engaging in destructive behaviour since it means someone messed something up in either the code or the filesystem.

src/server/campaignd/server.cpp
```diff
@@ -220,6 +222,73 @@ void server::load_config()
...
+					   << "' already exists in the new form! The old version will be ignored and its file deleted!\n";
+				filesystem::delete_file(filesystem::normalize_path(addon_file));
+				continue;
+			}
+			if(std::find(legacy_addons.begin(), legacy_addons.end(), addon_id) == legacy_addons.end()) {
+				ERR_CS << "No file has been found for the legacy addon '" << addon_id
```
shikadiqueen, line 256: Server needs to die here as well.

src/server/campaignd/server.cpp
```diff
@@ -220,6 +222,73 @@ void server::load_config()
...
+					   << "'. Please, check the file structure!\n";
+				continue;
+			}
+
+			config data;
+			in = filesystem::istream_file(filesystem::normalize_path(addon_file));
+			read_gz(data, *in);
+			if(!data) {
+				ERR_CS << "Couldn't read the content file for the legacy addon '" << addon_id << "'!\n";
```
shikadiqueen, line 265: Server needs to die here as well.

src/server/campaignd/server.cpp
```diff
@@ -220,6 +222,73 @@ void server::load_config()
...
+	std::vector<std::string> legacy_addons, dirs;
+	filesystem::get_files_in_dir("data", &legacy_addons, &dirs);
+	config meta;
+	for(const std::string& addon_dir : dirs) {
+		in = filesystem::istream_file(filesystem::normalize_path("data/" + addon_dir + "/addon.cfg"));
+		read(meta, *in);
+		if(meta) {
+			addons_.emplace(meta["name"].str(), meta);
+		} else {
+			WRN_CS << "Failed to load addon from dir '" << addon_dir << "'\n";
```
shikadiqueen, line 237: The server needs to die here instead of issuing a warning that absolutely nobody will see. If this condition is reached odds are that something is horribly wrong with the filesystem or its contents. Proceeding in this state is bound to result in data or metadata loss later.

src/server/campaignd/server.cpp
```diff
@@ -898,30 +1065,113 @@ void server::handle_upload(const server::request& req)
...
-		data["icon"] = (*campaign)["icon"];
-		data["type"] = (*campaign)["type"];
-		data["tags"] = (*campaign)["tags"];
 		find_translations(data, *campaign);
 
 		add_license(data);
+		
+		const std::string& new_version = (*campaign)["version"].str();
+		//#TODO: probably make hash from new_version + required_wesnoth_version ?
+		const std::string& file_hash = utils::md5(new_version).hex_digest();
```
shikadiqueen, line 1075: Why is the hash used here calculated from the version number instead of the archive contents?

src/server/campaignd/server.cpp
```diff
@@ -515,40 +598,37 @@ void server::send_error(const std::string& msg, const std::string& extra_data, s
...
-			found = true;
-			break;
-		}
-
-		++pos;
+	auto addon = addons_.find(id);
+	if(addon != addons_.end()) {
+		return addons_.at(id);
+	} else {
+		return config::get_invalid();
```
shikadiqueen, line 607: You just obtained a valid iterator to the item you were looking for and instead of dereferencing said iterator you're searching for the same item all over again.

src/server/campaignd/server.cpp
```diff
@@ -668,29 +753,111 @@ void server::handle_request_campaign_list(const server::request& req)
...
+	// The desired version is selected on the client side considering
+	// the min_wesnoth_version in [version] children of the addon info (#TODO: gfgtdf's part)
+	std::string to = req.cfg["version"].str();
+	const std::string& from = req.cfg["from_version"].str();
+	std::string full_pack = campaign["filename"].str();
+
+	auto version_map = get_version_map(campaign);
+
+	if(version_map.empty()) {
+		// Old format ?
```
shikadiqueen, line 772: If this is not supposed to happen then the server needs to abort this operation with an error.

src/server/campaignd/server.cpp
```diff
@@ -668,29 +753,111 @@ void server::handle_request_campaign_list(const server::request& req)
...
+			WRN_CS << "The selected version (" << to << ") for the addon '" << req.cfg["name"].str()
+				   << "' has not been found or it is unspecified! Sending the latest version instead!\n";
+			to = version_map.rbegin()->first;
+			full_pack += version_map.rbegin()->second["filename"].str();
+		}
+
+		// Negotiate an update pack if possible
+		if(!from.empty() && version_map.count(version_info(from)) != 0) {
+			config pack_data;
+			// Make a line of consequential updates beginning from the old version to the new one.
```
shikadiqueen, line 790: I believe the word you want here is either "consecutive" or "sequential", not "consequential".

src/server/campaignd/server.cpp
```diff
@@ -668,29 +753,111 @@ void server::handle_request_campaign_list(const server::request& req)
...
+					if(pack["from"].str() == prev_version["version"].str()
+							&& pack["to"].str() == next_version["version"].str()) {
+						config update_pack;
+						filesystem::scoped_istream in = filesystem::istream_file(campaign["filename"].str() + pack["filename"].str());
+						read_gz(update_pack, *in);
+						if(update_pack) {
+							pack_data.append(update_pack);
+							size += filesystem::file_size(campaign["filename"].str() + pack["filename"].str());
+						} else {
+							WRN_CS << "Unable to create an update pack sequence from version (" << from << ") to ("
```
shikadiqueen, line 813: Don't you mean "find" rather than "create" here?

src/server/campaignd/server.cpp
shikadiqueen: I rescind the previous comment. Obviously we can't use unsanitized input to locate files because otherwise we could get into serious trouble.

src/server/campaignd/server.cpp
kabachuha: Ok, added the error message.

src/server/campaignd/server.cpp
kabachuha: Thanks! Changed to addon->second .

src/server/campaignd/server.cpp
kabachuha: Added errors in such destructive situations in ba2bbbab21c7871c3e000138803af061648ad92e.

src/server/campaignd/server.cpp
```diff
@@ -247,23 +247,19 @@ void server::load_config()
...
-				continue;
+				throw filesystem::io_exception("The addon '" + addon_id
+					   + "' already exists in the new form! Possible code or filesystem interference!\n");
 			}
 			if(std::find(legacy_addons.begin(), legacy_addons.end(), addon_id) == legacy_addons.end()) {
-				ERR_CS << "No file has been found for the legacy addon '" << addon_id
-					   << "'. Please, check the file structure!\n";
-				continue;
+				throw filesystem::io_exception("No file has been found for the legacy addon '" + addon_id
+					   + "'. Please, check the file structure!\n");
```
shikadiqueen, line 255: This doesn't matter for the PR, but in the future avoid using un-assertive language like "please" when addressing either the player or (in this case) server administrators over stuff that demands their immediate action.

src/server/campaignd/server.cpp
```diff
@@ -769,10 +765,8 @@ void server::handle_request_campaign(const server::request& req)
 	auto version_map = get_version_map(campaign);
 
 	if(version_map.empty()) {
-		// Old format ?
-		ERR_CS << "The (" + to + ") version of the addon '" << req.cfg["name"].str()
-			   << "' is stored in the old format (bug?)! Trying to send a legacy full-pack.";
-		to = campaign["version"].str();
+		send_error("No versions of the addon '" + req.cfg["name"].str() + "' have been found by the server!", req.sock);
```
shikadiqueen, line 768: Quoting myself from Discord:

> "No versions of the addon 'foo' have been found by the server!" really should be "No versions of the add-on 'foo' are available on the server." in my opinion

(CC @Vultraz)

kabachuha: Changed the line wording, Vultraz approved it in a Discord message.

shikadiqueen:
Looks good to me.

kabachuha:
Added a mechanism for clientside upload pack generation which will reduce the server CPU and memory usage. Largely untested yet.

kabachuha:
Tested it on my local server today and fixed some spotted bugs. Now it's seemingly working as planned when handling upload packs.

CelticMinstrel:
I have two questions about this.

1. What if you upload a version of the addon with the _same_ version number as the latest version? Will it completely replace the current version, both upload pack and full pack?
2. Does this diff only at the filesystem level, or does it support file-level patches as well? That is, if a file has changed, will the diff contain a remove of the old file and an add of the new file, or will it contain a patch to transform the old to the new?

kabachuha:
@CelticMinstrel 
1. Yes, it will override the matching version and regenerate the upload pack for it. However, the uploadable addons may be required to have their version incremented in the future, as proposed by @shikadiqueen in https://github.com/wesnoth/wesnoth/issues/5079 .
2. It makes only full file replacements because the diff processing will put much more strain on the server (it will have to send back the whole addon to compute the diffs if a client generate diffs or it will have to generate them by itself what is quite an expensive operation; by contrast, in the PR it sends only the precomputed list of file hashes which is much smaller by definition and easier to handle) and create potential file content conflicts. There is not much use in the diffs to reduce the file size in our case as the text files are quite well compressed by gzip, so the relative growth compared with compressed diffs won't be significant, and the large binary files like images and sounds are generally not diffable anyway (or with serious complications).

src/server/campaignd/addon_utils.cpp
stevecotton: `filesystem::dir_size(dir)` ought to (as in, I'm looking at it now and think it already does) return 0 rather than -1 for an empty directory.

If you already have a setup for testing this, please would you? It would take a bit of setting up to test myself, as `purge_addon` and `purge_dir` are only called when upgrading an add-on with removed files; they aren't called using the "Remove add-ons" dialog, nor when deleting an add-on via the add-on manager's `-` button.

Pentarctagon: What needs to be tested? Just that dir_size returns 0 on an empty directory?

stevecotton: Good point, I can call it from somewhere else to test.

* It returns 0 for an empty directory
* It throws `boost::filesystem::filesystem_error` for a path that points to a nonexistent directory

The behavior of purge_addon and purge_dir also need to be tested, because it's definitely not expecting either of the above results. The first issue is that it probably leaves empty directories (which is trivial), however errors deleting a file are likely to leave the add-on in a partially-installed state (admittedly it probably already wasn't in the right state anyway).

PR #5039, opened 2020-08-09
Add more fauna resources

doofus-01:
Wesnoth has some fauna like giant rats and spiders, but the selection is pretty limited.

First commit to this is the tusker line by thespaceinvader a while back (needs some updates):
https://forums.wesnoth.org/viewtopic.php?p=338501#p338501

More to come, just wanted to get this PR kicked off.
EDIT:  Forum thread for this https://forums.wesnoth.org/viewtopic.php?f=9&t=52894


sevu:
Making an animals race might be an idea, to have two races instead of one crowded one.

data/core/units/monsters/Tusker_Tusklet.cfg
```diff
@@ -0,0 +1,99 @@
...
+            image="units/monsters/tusker/tusklet-water.png"
+            duration=50
+        [/frame]
+    [/standing_anim]
+#    ellipse="misc/ellipse-nozoc"
+    hitpoints=25
+    movement_type=orcishfoot
+    movement=4
+    experience=26
+    {AMLA_DEFAULT}
```
sevu, line 27: It has advances_to and {AMLA_DEFAULT}

soliton-:
I don't think an animals race would help. What would you put as animal and what as monster?

I think it's fine as is: either make more specific races for units that have several features in common or add them as `race=monster`.

data/core/units/monsters/Tusker_Tusklet.cfg
doofus-01: I did a pretty minimal update to the original files, to get the images to work.  We'll have to make several adjustments to these .cfg files.

doofus-01:
I agree it would be nice to not lump them all in with "monster", also agree using a second category "animal" would be problematic.  Is there any downside to having more specific races?

CelticMinstrel:
A possible compromise would be to make races for broad types of animals, though it's not clear how broad or specific such categories would need to be... "reptile" and "mammal" is probably a little too broad...

sevu:
Having too many races, like races with just one unit line, is the other extreme, crowing the race list instead of the unit list. Somewhere one must draw a line, into which group a unit line goes, even if it's vague.
Linking two help pages to each other in the description can be done additionally.

Another term which came to mind instead of animals would be `creatures`.

soliton-:
If the list is too crowded now then perhaps propose to group some races together. As @CelticMinstrel said there can potentially be broader race definitions. Otherwise I don't think there is a real solution unless we just want to have less unit types or somehow make up a taxonomy tree...

An arbitrary split of the catch-all monster race we have now into multiple generic races like "animal" or "creature" does not seem useful. It'll just be confusing to try to remember if some unit was a monster or animal or creature. (And if there is no reason to remember that then why do the split in the first place.)

doofus-01:
Falcons, gryphons, bats, and ogres have their own race already, seems we've made our decision.  I guess they can just go into monsters if they are solo units (no variations or other levels), otherwise, they get a unique race.

CelticMinstrel:
I'm confused, why is the falcon listed as new? It looks like it should be a move...?

doofus-01:
What?

data/core/units/monsters/Icemonax.cfg
```diff
@@ -0,0 +1,93 @@
...
+    movement=4
+    experience=26
+    level=0
+    alignment=chaotic
+    advances_to=null
+    {AMLA_DEFAULT}
+    undead_variation=mounted
+    cost=10
+    usage=fighter
+    description=_ "Icemonax are a strange beast of the frozen north, best described as a large, armored ground squirrel.  Their armor accummulates frost and snow, even in more temperate climates.  Most Icemonax are curious, often wandering into places they do not belong.  Despite this, efforts to study and domesticate them have failed, as they die soon after capture."
```
stevecotton, line 34: This should only use a single-space after each full-stop.

data/core/units/monsters/Icemonax.cfg
```diff
@@ -0,0 +1,93 @@
...
+#            end=150
+#            image="units/monsters/tusker/tusklet-moving.png"
+#        [/frame]
+#    [/movement_anim]
+    {DEFENSE_ANIM "units/monsters/glypto/young-glypto-defend2.png" "units/monsters/glypto/young-glypto-defend1.png" tusker-hit.ogg }
+    [defense]
+        village=50
+        frozen=50
+    [/defense]
+    [movement]
```
stevecotton, line 48: This should be `movement_costs`.

The Travis job is reporting this and some other WML validation errors.

CelticMinstrel:
**Me:**
> I'm confused, why is the falcon listed as new? It looks like it should be a move...?

@doofus-01:
> What?

According to the diff, you've added a new graphic for the falcon and the elder falcon, but both of those already had graphics. I _think_ you meant to move it, but forgot to delete the original?

As of right now there are two (identical, I believe) versions of the falcon graphic - one in units/monsters/ and one in units/monsters/falcon/.

doofus-01:
> According to the diff, you've added a new graphic for the falcon and the elder falcon, but both of those already had graphics. I think you meant to move it, but forgot to delete the original?

Oh, right.  It's a different image, I haven't bothered much yet with git bookkeeping, pleasing travis, etc.  Just distractions at this point.

CelticMinstrel:
> Oh, right. It's a different image, I haven't bothered much yet with git bookkeeping, pleasing travis, etc. Just distractions at this point.

So you've _changed_ the falcon images then, correct?

It's not a priority right now obviously but if the old images are unused they should definitely be deleted before this is merged (and if this is squashed, that'll likely cause git to recognize it as a move).

doofus-01:
This fauna project overall isn't done, but I think I'll go ahead and merge this PR if Travis CI doesn't find anything horrible.

PR #5040, opened 2020-08-09
Have mainline content send [multiplayer][addon] info as well.

Pentarctagon:
This commit:
* sends the era and scenario as part of a [multiplayer][addon] in order to make it consistent with how add-ons' information is sent to wesnothd. this means all of the game's content information is now available in the [multiplayer][addon] data.
* correctly covers the case of there potentially being modifications added to mainline at some point. the current code looks at the [multiplayer]mp_scenario= and [multiplayer]mp_era attributes to cover the case of a mainline era/scenario being used, but looks for modifications only in the [multiplayer][addon] data, which wouldn't have any entry for a mainline modification right now if one were added.

### DIFF
--- a/src/saved_game.cpp
+++ b/src/saved_game.cpp
@@ -296,47 +296,43 @@ void saved_game::expand_scenario()
 
 void saved_game::check_require_scenario()
 {
-	if(starting_point_["addon_id"].empty()) {
-		return;
-	}
-
-	config required_scenario;
-
-	required_scenario["id"] = starting_point_["addon_id"];
-	required_scenario["name"] = starting_point_["addon_title"];
-	required_scenario["version"] = starting_point_["addon_version"];
-	required_scenario["min_version"] = starting_point_["addon_min_version"];
-	required_scenario["required"] = starting_point_["require_scenario"].to_bool(false);
-	config& content = required_scenario.add_child("content");
+	const std::string version_default = starting_point_["addon_id"].empty() ? game_config::wesnoth_version.str() : "";
+	config scenario;
+	scenario["id"] = starting_point_["addon_id"].str("mainline");
+	scenario["name"] = starting_point_["addon_title"].str("mainline");
+	scenario["version"] = starting_point_["addon_version"].str(version_default);
+	scenario["min_version"] = starting_point_["addon_min_version"];
+	scenario["required"] = starting_point_["require_scenario"].to_bool(false);
+	config& content = scenario.add_child("content");
 	content["id"] = starting_point_["id"];
 	content["type"] = "scenario";
 
-	mp_settings_.update_addon_requirements(required_scenario);
+	mp_settings_.update_addon_requirements(scenario);
 }
 
-void saved_game::load_mod(const std::string& type, const std::string& id, size_t pos)
+// "non scenario" at the time of writing this meaning any era, campaign, mods, or resources (see expand_mp_events() below).
+void saved_game::load_non_scenario(const std::string& type, const std::string& id, size_t pos)
 {
 	if(const config& cfg = game_config_manager::get()->game_config().find_child(type, "id", id)) {
 		// Note the addon_id if this mod is required to play the game in mp.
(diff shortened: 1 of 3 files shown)


gfgtdf:
im wondering whether we maybe should use the version wesnoth in the version attribute for this.

Pentarctagon:
I didn't since that information is already available by looking at the host's [CLIENT_VERSION](https://github.com/wesnoth/wesnoth/blob/ee6d86c20b815b62836cb24f6bd49a4cb9e66bc1/utils/mp-server/table_definitions.sql#L101), though including it here as well wouldn't hurt anything either.

src/saved_game.cpp
```diff
@@ -296,47 +296,40 @@ void saved_game::expand_scenario()
...
-	required_scenario["id"] = starting_point_["addon_id"];
-	required_scenario["name"] = starting_point_["addon_title"];
-	required_scenario["version"] = starting_point_["addon_version"];
-	required_scenario["min_version"] = starting_point_["addon_min_version"];
-	required_scenario["required"] = starting_point_["require_scenario"].to_bool(false);
-	config& content = required_scenario.add_child("content");
+	config scenario;
+	scenario["id"] = starting_point_["addon_id"].str("mainline");
+	scenario["name"] = starting_point_["addon_title"].str("mainline");
+	scenario["version"] = starting_point_["addon_version"];
```
soliton-, line 302: Perhaps this should default to the mainline/wesnoth version? (Same thing a little further down.)

src/saved_game.cpp
```diff
@@ -296,47 +296,40 @@ void saved_game::expand_scenario()
...
+	config& content = scenario.add_child("content");
 	content["id"] = starting_point_["id"];
 	content["type"] = "scenario";
 
-	mp_settings_.update_addon_requirements(required_scenario);
+	mp_settings_.update_addon_requirements(scenario);
 }
 
-void saved_game::load_mod(const std::string& type, const std::string& id, size_t pos)
+void saved_game::load_non_scenario(const std::string& type, const std::string& id, size_t pos)
```
soliton-, line 312: I think this should have a better name or at least a short description what a "non-scenario" is. (campaign/era/modification/resource/more?)

src/saved_game.cpp
Pentarctagon: See my answer to gfgtdf just above.

src/saved_game.cpp
Pentarctagon: Currently it's loading the mods, era, campaign, and resources.  I'm not sure what a better name would be, though I think `load_non_scenario` is more accurate than `load_mod`.

src/saved_game.cpp
soliton-: If that's what gfgtdf said then I agree.

src/saved_game.cpp
soliton-: So a description could be: "non-scenario" here means modification, era, campaign or resource.

Anyway I've approved since these are just minor suggestions. Feel free to disregard if you don't think they're useful.

src/saved_game.cpp
Pentarctagon: gfgtdf said he was wondering whether wesnoth's version should be used, so if you agree with that, then... should I add it? (not trying to be difficult; I'd just like something definitive one way or the other)

src/saved_game.cpp
Pentarctagon: I'll add it as a comment then.

src/saved_game.cpp
soliton-: I would definitely do it that way.

Pentarctagon: Alright, updated.

gfgtdf:
tbh the main reason why i suggested addon the wesnoth version there was that having no version there would make  https://github.com/wesnoth/wesnoth/pull/5022 more complicated.

Looks good to me.

PR #1925, opened 2017-08-28
Preprocessor cleanup

Vultraz:
Ignore the first commit (and don't view the unified diff, you won't get anything out of it). Mostly some work to make the memory management of the preprocessor saner, as requested by @jyrkive. Currently crashes on startup :(

Issue #1744, opened 2017-05-31
Text rendering output bad when copied with alpha blending


Vultraz:
During my work on the accelerated rendering branch, I've once again run into the issue that plagued me when I attempted to use textures for text in the GUI2 canvas along with a surface renderer: **text looks like crap if it's copied with alpha blending**:

![this text looks like shit](https://cloud.githubusercontent.com/assets/3558659/26613141/0aa2f902-4605-11e7-9959-ec972deb1603.JPG)

For some reason, the text looks fine on buttons (and only buttons - in other tests before I managed to generally restrict the effect to semi-transparent surfaces).

This doesn't occur when copying the text directly to the screen renderer (`SDL_RenderCopy`) or copying the pixel data verbatim to a texture (`SDL_UpdateTexture`). However, the former conflicts with my design of having draw results cached to a texture, and the latter is impossible to use since having no blending at all cases the fully transparent area of the surface to overwrite whatever data is underneath them. The same thing occurs when disabling alpha blending (`SDL_BLENDMODE_NONE`) on either the text surface or a texture created from that surface.

This is related to #1568. The underlying cause is probably the same. This issue isn't by any means new; it's likely the reason we have the custom `blit_surface` function. I'm guessing #1568 simply covers cases it wasn't able to cover.

I'm really not sure how to solve this. A stopgap measure is to call `pango_cairo_show_layout` twice, but I'm not sure that's really what we want to do.

And for the record, no, this isn't a premultipled vs postmultipled alpha problem (at least, not entirely). Our rendering code already does a pre-to-post multiplied alpha conversion on the cairo text output.

Wedge009:
Upon seeing the bug title, I also thought it might be related to #1568. I don't think we want to call show layout twice - I remember you already removed that kind of thing on Windows for the last Cairo text rendering issue.

Speaking of which, is this only on Windows or does it happen on other OSs?

Vultraz:
Not sure, but given that `blit_surface` is used on all platforms I'd say it's probably not Windows-specific.

Wedge009:
Comparing (by eye) between Linux and Windows (MSVC), I don't see this issue. Windows looks closer to what is shown in #5618 (after change applied), so assuming that PR has resolved this issue.

Issue #1568, opened 2017-05-08
GUI2 canvas text looks bad when rendered on a semi-transparent bg (GNA #24959)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?24959)** by vultraz on 2016-08-10

gui2::tcontrol widgets like tlabel seems to call font::ttext directly, while gui2 canvas operations call gui2::ttext (which then calls font::ttext). The output of the latter seems to be worse than the output of the former.

The latter also doesn't seem to handle newline characters correctly.

Needs investigation.

(Reproduced on Windows 10)
Release: All
Priority: 5 - Normal
Severity: 3 - Normal

wesnoth-bugs:
**Modified on 2016-08-11**

**wedge009 wrote:**

> Is there a good example of a dialogue I should be looking at to check this?

wesnoth-bugs:
**Modified on 2016-08-11**

**vultraz wrote:**

> Compare, for example, the text in the headings of the status dialog to the text in the list. The former looks markedly worse, IMO.

wesnoth-bugs:
**Modified on 2016-08-12**

**wedge009 wrote:**

> It's not an obvious difference to me. Perhaps the font size difference may also be making an impact? Do you think it's another Windows-specific text rendering issue? Any difference that's there is even harder to see in Linux. If I had to pick one, overall I'd say the Linux rendering looks nicer.
> 
> 
> 
> I had a look at the ttext references, to see if maybe I could experiment with using one over the other, but I find it all so scattered and hard to read. ):
> 
> (file #28272, file #28273)

**wedge009 added attachment:** [linux.png](http://gna.org/bugs/download.php?file_id=28273)

**wedge009 added attachment:** [windows.png](http://gna.org/bugs/download.php?file_id=28272)

wesnoth-bugs:
**Modified on 2016-08-18**

**vultraz wrote:**

> Hmmm. Could you compare at any text you can type in text boxes, such as the username field in mp connect?

wesnoth-bugs:
**Modified on 2016-08-19**

**wedge009 added attachment:** [mp_linux.png](http://gna.org/bugs/download.php?file_id=28414)

**wedge009 wrote:**

[...]

PR #5042, opened 2020-08-10
text.cpp: throw if rendering needs more than 2GB, fix antialiasing with stride > width

stevecotton:
The main change here is avoid a crash, but there's also a fix for a possible UI issue with text looking bad. I've made the two changes in a single commit, because if we backport either we might as well backport both (in my opinion).

The effect of text looking bad might only occur with certain combinations of OS, debug/release builds, and potentially graphics card and driver; I don't know whether it ever occurs. The effect is similar to  #1568, #1744 and #1925, to the extent that I'd expect any reports of it to be treated as duplicates of those bugs, however it doesn't quite match the screenshots in those bugs; when this bug is the root cause.

Simulated screenshot, showing what happens if Cairo pads the stride to twice the width - the first half of the text (vertically) looks OK. For this bug to cause the screenshots in #1568, #1744 and #1925 would need an excessive amount of padding, which is why I doubt it's the cause of those bugs.
half_stride_malin

Refactor text.cpp's buffer alloc, throw instead of crashing if it needs more than 2GB

The limit is arbitrary, and 2 gigabytes is very large. The new TODO comment
in the .cpp file gives a possible future refactor which would reduce the limit.

The old early-return for surface_buffer_.empty() moves upwards and changes to
testing for zeros before the allocation; it now also acts as a guard for
division by zero in the (height > int::max() / stride) test.

The loop around from_cairo_format() relied on stride being exactly
sizeof(uint32_t) * width, an assumption which would break if
cairo_format_stride_for_width added padding.

### DIFF
--- a/src/font/text.cpp
+++ b/src/font/text.cpp
@@ -681,14 +681,30 @@ void pango_text::rerender(const bool force)
 		if(maximum_height_ > 0) { height = std::min(height, maximum_height_); }
 
 		cairo_format_t format = CAIRO_FORMAT_ARGB32;
-		const unsigned stride = cairo_format_stride_for_width(format, width);
+		const int stride = cairo_format_stride_for_width(format, width);
 
-		this->create_surface_buffer(stride * height);
-
-		if (surface_buffer_.empty()) {
+		// The width and stride can be zero if the text is empty or the stride can be negative to indicate an error from
+		// Cairo. Width isn't tested here because it's implied by stride.
+		if(stride <= 0 || height <= 0) {
 			surface_ = surface(0, 0);
+			surface_buffer_.clear();
 			return;
 		}
+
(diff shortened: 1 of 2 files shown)


AI0867:
I don't know what the `cairo_format_stride_for_width` function does exactly, but it looks like you're relying on overflow, and that it not overflow too far. I suppose that's all you can really do with the output if it returns int though. It might be a good idea to sanity-check the inputs?

stevecotton:
`cairo_format_stride_for_width` returns "the appropriate stride to use given the desired format and width, or -1 if either the format is invalid or the width too large". So negative values are an error flag, rather than the result of overflow.

stevecotton:
@Vultraz , @AI0867 : ping, please

AI0867:
@stevecotton looks okay then. Sorry for the (very) late reply. I've been inundated with mail lately.

Issue #4280, opened 2019-08-26
Remove two TODO comments from the help text


jostephd:
Remove the two TODO comments added in #4217. I need some input about _how_ to fix this, see https://github.com/wesnoth/wesnoth/pull/4217#issuecomment-524574853

This issue is just about those two comments. For other todo comments in the help see #2964

- [x] Fix `At all difficulties except “Beginner” (TODO: and “Easy”?),`
- [x] Fix `TODO: link to a forum thread`
- [x] Fix `At least for the 1.15.x series before 1.16.0` (see next comment)
- [x] Does `In the mainline campaigns, on hard` mean "on each campaign's hardest difficulty"? Even if the hardest difficulty is "Nightmare"?

stevecotton:
Thanks for logging this, it's probably a good ticket to fix the save-loading page related "refer to the forums" line:

https://github.com/wesnoth/wesnoth/blob/c4ffade49aa4efc2a410e031539a791c005c3b30/data/core/help.cfg#L289-L294

jostephd:
Agreed, added a checklist to the first post. Also marked as blocker, not in any technical sense but because it's the sort of thing that should be fixed before a final release.

jostephd:
This is a string freeze blocker. Could someone take this issue on and fix it? That is, either draft a PR, or start a discussion first to figure out what the PR should say?

stevecotton:
We've already had a discussion without reaching consensus.
https://forums.wesnoth.org/viewtopic.php?f=3&t=18640

jostephd:
I remember that, but even though we didn't reach consensus, we still need to fix the todos before string freeze, one way or the other. For example, maybe we can agree that "winnable without foreknowledge" is always true when difficulty is at most "Normal" and level is at most "Intermediate", and leave the expectation of higher difficulties/levels unspecified. Or maybe we'd just revise those bullets into recommended rules of thumb for creating UMC campaigns. Or something.

stevecotton:
@slavrenyuk has brought up a similar discussion on Discord https://www.wesnoth.org/irclogs/2019/11/%23wesnoth-dev.2019-11-06.log

CelticMinstrel:
Suggestion: Remove the TODO notes from the string and add them instead as a comment near the string. That way, at least they don't block the string freeze.

stevecotton:
[...]

PR #5076, opened 2020-08-12
Help about losing: remove reference to specific difficulties

stevecotton:
Part of resolving issue #4280.

This won't be true of all scenarios in all campaigns, but as a general rule it seems fine.


CelticMinstrel:
I don't really have anything to say on the content of this change, sorry.

soliton-:
If this means giving up on the concept that the easiest difficulty (at least in easy campaigns) should be beatable with no carryover, I find that unfortunate.

stevecotton:
> If this means giving up on the concept that the easiest difficulty (at least in easy campaigns) should be beatable with no carryover, I find that unfortunate.

The discussion threads about difficulty didn't reach a consensus. I feel it was right to put a TODO in the docs while the discussion was ongoing; but that TODO is no longer being discussed, and the docs are the wrong place to try to reopen it.

soliton-:
> The discussion threads about difficulty didn't reach a consensus. I feel it was right to put a TODO in the docs while the discussion was ongoing; but that TODO is no longer being discussed, and the docs are the wrong place to try to reopen it.

I wasn't aware there was no consensus on the easiest difficulty of easy campaigns and the TODO certainly did not reflect that either. Maybe I missed some discussion.

Pentarctagon:
I'd say that putting a TODO in text that users will see is always not the correct thing to do - to the user it looks like an indicator of poor quality, and for the developers it risks the TODO ending up forgotten.

Pentarctagon:
As far as whether beating Easy campaigns on the Easy difficulty setting should be possible without considering gold carryover, I think it makes sense for that to be possible.

stevecotton:
Thanks, I think for the moment it's a case of "neither Beginner nor Easy", at least for HttT.

The finale of HttT on beginner difficulty gives the player 400 gold (plus carryover), and the enemy sides start with a total of 950 gold and all the villages. I think it's safe to say that a player whose skill level gets them there with no carryover and no veterans isn't going to win.

AToTB's finale is definitely doable on beginner difficulty. There's an ally side that's almost balanced with the enemy, and with just a little assistance from the player the ally will win.

I don't know what to make of TSG's final scenario (of the elf branch). The scenario needs a unit (that's not the player's leader) to reach a certain hex on or after turn 11. It's easy with an understanding of ZoC and baiting the AI to chase your leader, but those skills should be enough to get there with carryover and a recall list.

soliton-:
> I think it's safe to say that a player whose skill level gets them there with no carryover and no veterans isn't going to win.

FTR, I cannot remember a complaint about that scenario on the easiest difficulty. If the player having less gold is the definite metric by which we can tell that the scenario is not beatable there seems to be a trivial way to fix that. I do disagree with that metric though. It seems to assume that the player is incapable of learning anything. Especially on the last scenario you just have to find some way to beat this scenario.

If a player made it this far and cannot beat this scenario it seems like an excellent place to learn some more tactics to manage that. If the player is in any way interested in wesnoth game play I find it hard to believe that it's "safe to say" that they will have no chance to beat this scenario nor do I believe that this is the intention of the setup.

stevecotton:
Clarification - it wasn't a complaint about the difficulty, just a decision about this line of the documentation. The line in question talks about both carryover and recall list - in the case of the last scenario of HttT, even minimal gold should allow circa 20 recalls, all of them likely to be max level.

soliton-:
I understood that it was not a difficulty complaint but supposed to be a reason for the change in this PR. As explained above I do not agree with the reasoning or the documentation change.

Pentarctagon:
I suspect even a skilled player would have some difficulty on the final scenario of HttT on Easy with no extra gold or a recall list.  Is it *possible* to beat?  Probably.  But beating a scenario where you start with half the gold of your opponents, no village income or any good way to obtain more than a couple villages, and only able to recruit L1 elves to fight L1+L2 humans and orcs on flat terrain would be exceptionally unlikely to be accomplished by a player whose tactics up to this point did not result in any carryover gold or a recall list to utilize.

soliton-:
Seems like a fun challenge but I guess showing that it is possible won't change what the intention is. I'd like to know a definitive answer to that since it sounds like I'll have to adjust the advice I've given the last decade.

So from now on if a player does not manage to beat a scenario in an easy campaign on the easiest difficulty the suggestion is to replay previous scenarios and not adjust the difficulty?

Pentarctagon:
The text still says "If you find yourself in this situation, you may try to go back a scenario or two and win them more convincingly, or change to an easier difficulty" though?

Also, if a player is playing an Easy campaign on the easiest difficulty, what would they adjust the difficulty to?

soliton-:
I meant the campaign maintainer would adjust the difficulty if they get such reports. :wink:

Pentarctagon:
Right.  And likewise, if nemaara says that right now Wesnoth's campaigns are all beatable on Beginner and potentially Easy difficulties without any carryover gold or recalls, then the help should be updated (again) to reflect that.  What I'm opposed to is having the help text say that all of Wesnoth's campaigns are all beatable on Beginner and potentially Easy difficulties without any carryover gold or recalls when they actually aren't, even if the goal is that they should be.

PR #5077, opened 2020-08-12
Help dealing with losses: don't link to the forums

stevecotton:
This is part of resolving #4280.


shikadiqueen:
A better question is why Wesnoth of all games needs this information to be included in the game itself.

CelticMinstrel:
As for the link itself, I think it makes sense to just leave it as-is for translations? The translators could note that it links to an English-language forum but otherwise leave it untouched.

stevecotton:
Thanks to everyone who commented, I've deleted the line and removed the "WIP" from the title.

Issue #1129, opened 2017-05-08
Slow should stay in effect until it's slower's turn again, but does not (GNA #11393)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?11393)** by anonymous on 2008-03-28

In 2on2 MP games the "slow" effect is removed once the slowed unit's side ends its turn. Instead, it should stay in effect until it is the slower's turn again.

Example:

Assume players p1 to p4 with team1 being p1,p4 and team2 being p2,p3. Now p3 slows a unit of p1 on its turn. Slow stays effective during p4's turn and during p1's turn now. But once p1 has ended its turn, slow is already removed, leaving p2 without the possibility to take advantage of the team effort. Currently, the "slow" effect can only be used by the other team players if they move after the slower's side.

However it would make more sense to me if turn order would not matter here.

The attachment is a replay showing the problem. In turn2, Green has slowed a red Gryphon. But when Blue is about to move (turn3), slow is already gone.

(Reproduced on Linux)
Release: 1.4 stable
Priority: 5 - Normal
Severity: 3 - Normal

gfgtdf:
> In 2on2 MP games the "slow" effect is removed once the slowed unit's side ends its turn. Instead, it should stay in effect until it is the slower's turn again.

Yes this is the same behvaiour as in sp, changing this woudl make the game more complicated since the game and the players would have to remmber who slowed the unit. I don't think this will be changed.

ProditorMagnus:
Implemented as https://forums.wesnoth.org/viewtopic.php?t=42668

gfgtdf:
Anyone with an opinion on this should we just reject this?

Pentarctagon:
IMO the duration of the slows effect should be consistent regardless of who did the slowing.  Having it be dependent on which side did the slowing feels like an unnecessary complexity, and also it's not game logic that's as obvious to the player due to the current inconsistency.

@Hejnewar - any opinion?  This would effectively be a sort of buff to slows, though since this would only affect maps with 3+ players my assumption is it wouldn't be too much of an impact.

gfgtdf:
> IMO the duration of the slows effect should be consistent regardless of who did the slowing. Having it be dependent on which side did the slowing feels like an unnecessary complexity, and also it's not game logic that's as obvious to the player due to the current inconsistency.

Im not quite sure what you mean here you first  talk about complexity, and then about the current sitatuion, so are you trying to say that the _current_ way is more complex?

[...]

PR #5082, opened 2020-08-15
make slow duration to new turn instead to end turn

newfrenchy83:
tis fix https://github.com/wesnoth/wesnoth/issues/1129 issue


gfgtdf:
No, it doesn't implement what was suggested in that issue, which is to make the slow effect last until the turn of the unit that slowed the units (not the unit that was slowed.)

In fact this pr will remove the effect of slow completely since the unit will no longer be slowed  in the next  turn.

Issue #4480, opened 2019-10-17
poedit warnings for translatable strings "HINT^ line starting with space"


stevecotton:
Poedit does some sanity checks, and warns in the UI if the original and translation differ in:
* starting with whitespace
* ending with whitespace
* ending with a `.`

The new special notes all trigger warnings, because `SPECIAL_NOTE^ This unit is able to move through solid stone walls.`, etc should be translated starting with a space. Poedit doesn't know about Wesnoth's hinting method.

jostephd:
We should change the special note strings so they don't start with spaces.

We should also not use leading spaces in translated strings. Some translations didn't keep the leading space where the English string had one.

stevecotton:
What about strings that start with newlines? If it needs a change, I'd like to update #4414's new documentation before the translators work on it.

jostephd:
I think the best form would be to have neither leading whitespace nor trailing whitespace (where by "whitespace" I mean spaces, tabs, newlines, etc) in translated strings. Making the change only for strings that haven't been processed by a pot-update sounds like a good idea.

stevecotton:
Logged #4484, "pot-update broken by #4414's editor documentation", although I'm hoping @AncientLich will take it as a feature-request for wmlxgettext.

Wedge009:
With respect to poedit's checks, they are useful, but Wesnoth's conventions do indeed often cause false alerts on these.

#4305 predates this report and seems to resolve the issues around those checks - is there still more work to do on this, for Special Notes or other things?

stevecotton:
There's exactly two left, according to a search of the .pot files for anything that has a caret followed by whitespace. It's just a co-incidence that both are special notes.

* UtBS's `SPECIAL_NOTE^ Unlike normal Shadows, Ethereal Shadows cannot hide at night.`
* The Silver Mage's special note, which is already fixed on master and in the list for backporting (#5979)

Expect a new PR to referencing this issue soon.

Wedge009:
Hooray, thanks.

PR #5083, opened 2020-08-15
Save-loading: remove "under discussion", add hint about accepting losses

stevecotton:
This removes some subtle todo text that didn't have a explicit "TODO" in
the text, although it was tracked as part of #4280. The discussion ended
without a conclusion about whether those statements were correct for all
mainline campaigns. This PR resolves two of the four subtasks of #4280.

The existing text in this help topic represented paragraphs by having
translatable strings that start with whitespace (two newlines), as described in
issue #4480. As this commit already needed to change all but two of the
paragraphs, I've switched them all to using non-translatable paragraph
separators.  Even though the whole save-loading topic is new in 1.15, it has
already been translated in some languages, and making this consistent means two
unnecessary fuzzy strings. Sorry to the translators, but I think it's
worthwhile. @ivanovic , this is why I've requested your review.

### DIFF
--- a/data/core/help.cfg
+++ b/data/core/help.cfg
@@ -285,18 +285,16 @@ Normally you can undo a unit’s movement, as long as an event with a randomized
     [topic]
         id=saveload
         title= _ "Save-loading"
-        text= _ "Random numbers are part of Wesnoth, attacks can fail and units can die due to bad luck. This is an expected part of the game, and one of the aims of the game is that a sufficiently skilled player should be able to complete all of the mainline campaigns, on hard, without needing to save-load. Going back to a previous turn to try a different strategy is a part of learning the game, but we recommend against reloading merely to try the same strategy again while hoping for better luck." + _ "
+        text= _ "Random numbers are part of Wesnoth, attacks can fail and units can die due to bad luck — this is an expected part of the game. Going back to a previous turn to try a different strategy is a part of learning the game, but we recommend against reloading merely to try the same strategy again while hoping for better luck." + "
 
-One of the challenges of the game is to work out how to protect your heroes. Small risks quickly build up: if you have five important units, and they each have just a 1% chance of death each turn, you can <italic>text='expect'</italic> one of them to die about every 20 turns. So, you’ll rarely make it through a scenario without having it happen." + _ "
+" + _ "One of the challenges of the game is to work out how to protect your heroes. Small risks quickly build up: if you have five important units, and they each have just a 1% chance of death each turn, you can <italic>text='expect'</italic> one of them to die about every 20 turns. So, you’ll rarely make it through a scenario without having it happen." + "
 
-The following points are under discussion on the forums. At least for the 1.15.x series before 1.16.0 please assume that the forum threads are more authoritative than the text on this page." + _ "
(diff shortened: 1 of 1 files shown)


data/core/help.cfg
```diff
@@ -285,18 +285,15 @@ Normally you can undo a unit’s movement, as long as an event with a randomized
...
+" + _ "Small risks build up for your non-hero units too — however these units don’t have to survive. While the story and dialogue usually aim to give you some emotional bond with your troops, the game mechanics reward cold-blooded thoughts such as “even if those units survive, I’m not going to recall them in any future scenarios, ...”, and the even colder “..., and that will reduce my upkeep”." + "
 
-In the mainline campaigns, on hard:
-• A sufficiently skilled player will have a good chance to win each scenario on the first time that they see that scenario, without foreknowledge of what will happen.
-• This assumes that the player picks up all of the clues given in the dialogue.
-• There are more experience points available than on easy, and the player will be able to train enough troops to cope with losing some L2 or L3 units." + _ "
+" + _ "The difficulty levels are balanced assuming that the player won’t save-load. In the mainline campaigns, harder difficulties generally have more enemies and thus more experience points available; this leads to them being balanced assuming that the player will be able to train enough troops to cope with losing some L2 or L3 units." + "
```
Pentarctagon, line 294: I'd change L2 and L3 to "level two" and "level three", unless the former way is consistent with existing text elsewhere.

data/core/help.cfg
```diff
@@ -285,18 +285,15 @@ Normally you can undo a unit’s movement, as long as an event with a randomized
...
+" + _ "One of the challenges of the game is to work out how to protect your heroes. Small risks quickly build up: if you have five important units, and they each have just a 1% chance of death each turn, you can <italic>text='expect'</italic> one of them to die about every 20 turns. So, you’ll rarely make it through a scenario without having it happen." + "
 
-The following points are under discussion on the forums. At least for the 1.15.x series before 1.16.0 please assume that the forum threads are more authoritative than the text on this page." + _ "
+" + _ "Small risks build up for your non-hero units too — however these units don’t have to survive. While the story and dialogue usually aim to give you some emotional bond with your troops, the game mechanics reward cold-blooded thoughts such as “even if those units survive, I’m not going to recall them in any future scenarios, ...”, and the even colder “..., and that will reduce my upkeep”." + "
```
Pentarctagon, line 292: I'd change:

> the game mechanics reward cold-blooded thoughts such as “even if those units survive, I’m not going to recall them in any future scenarios, ...”, and the even colder “..., and that will reduce my upkeep”

to:

"the game mechanics also often reward purely practical actions, such as making sure the first units to die are ones you have no intention of recalling in future scenarios."

stevecotton: I've pushed a new draft that goes in a different direction: "the game mechanics reward remembering that it’s just a game — there is rarely any penalty for letting these units die, other than being unable to recall them in future scenarios. Your battle-plan should distribute the risks based on which troops you want to recall later."

Pentarctagon: "battle-plan" shouldn't have a dash in it, but otherwise I'm good with that.

stevecotton: Resolved in the version pushed to master.

stevecotton:
Merged - it shows as closed because I pushed to master along with a commit updating the changelog.

Issue #4004, opened 2019-03-28
Some replays don't work (replay menu doesn't appear/oos error)


Konrad22:
My replay from TLU Sc0 (Badass) and both mine and @nemaara 's replay from AtS3 S1 don't work for some reason. (The oos error doesn't happen for me, only for nemaara, no idea why.)
Since it happens in two unrelated add-ons, it might be a wesnoth engine bug.

[https://github.com/project-ethea/After_the_Storm/issues/68]

[AtS-3-Beyond her Smile replay.gz](https://github.com/wesnoth/wesnoth/files/3018741/AtS-3-Beyond.her.Smile.replay.gz)
[TLU-Morning in al-Kamija Aufzeichnung abspielen.gz](https://github.com/wesnoth/wesnoth/files/3018742/TLU-Morning.in.al-Kamija.Aufzeichnung.abspielen.gz)

[https://forums.wesnoth.org/viewtopic.php?p=639284#p639284]

gfgtdf:
https://github.com/wesnoth/wesnoth/blob/1.14.6/src/replay_controller.cpp#L77

https://github.com/wesnoth/wesnoth/blob/1.14.6/src/display.cpp#L279

stevecotton:
Both of these replays use `[change_theme]`, and it seems there's no trigger for replay_controller to attach itself to the new theme's theme_reset_event.

gfgtdf:
Yes that's what I was trying to say I. My previous post, I think the easiest way to fix it would be just, in the line that recreates.tye theme to move over all event handles from the old theme object to the new theme object

shikadiqueen:
Another possibility would be to simply ignore `[change_theme]` actions when in replay mode (no idea what should be done for multiplayer).

PR #5086, opened 2020-08-17
fire theme_reset event after theme was changed

gfgtdf:
probably not the prettiest fix, 4004

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -287,8 +287,13 @@ display::~display()
 	resources::fake_units = nullptr;
 }
 
-void display::set_theme(config theme_cfg) {
+void display::set_theme(config theme_cfg)
+{
+	auto reset_event = theme_.theme_reset_event();
 	theme_ = theme(theme_cfg, screen_.screen_area());
+	theme_.theme_reset_event() = reset_event;
+	theme_.theme_reset_event().notify_observers();
+
 	builder_->set_draw_border(theme_.border().show_border);
 	menu_buttons_.clear();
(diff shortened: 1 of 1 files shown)


shikadiqueen:
What does this do in practice though?

gfgtdf:
as described on https://github.com/wesnoth/wesnoth/issues/4004  previous the replay_controller listenes to a reset_replay event that is destroyed  in that line, so we try to carry to  over to the new theme object.

shikadiqueen:
Doesn't that gloss over the fact that the relevant themes (`Cutscene_Minimal` and `Cutscene`) as well as any other potential candidates for use with `[change_theme]`, do not or might not provide any replay controls, so if for whatever reason you are meant to stay with a different theme than a replay-suitable default you are going to lose the ability to control the replay anyway?

gfgtdf:
hmm i thought the code in replay_cotnroller  would just take the [replay] tag from a global config  and not from the theme itself, so that it'd alway be available. I might have been wrong here.

In that case, can't we just add the the replay theme by using the same for as for the default theme to all of these themes?

shikadiqueen:
That I can't tell you since I'm not well versed in theme WML or the underlying infrastructure. I just wanted to point out again that even if the themes in mainline are patched accordingly by hand, that doesn't mean that it will be done by WML authors when writing their own custom themes as well.

Pentarctagon:
Closing as gfgtdf is no longer around and this fix is incomplete.

PR #5091, opened 2020-08-20
add moveto_item event

gfgtdf:
this event is fired whena unit steps on a [item], this basicially moves
the implementation of the wc2_drop_pickup event to mainline to reduce code duplication


stevecotton:
It looks as if there's interesting stuff here, IIUC a way for the `moveto_item` callback to set a `taken` variable which has similar effects as calling `remove_item`. However, there's no documentation for it, and there's no new unit test to check that it works - I think both are needed before this merges.

gfgtdf:
> However, there's no documentation for it

Usualyl these thigns get documented in the wiki after they are merged.

stevecotton:
Having documentation in the pull request makes it possible to review whether code's side-effects match the docs, and it can then be copied to the wiki after merging. This is mentioned in the https://wiki.wesnoth.org/PatchSubmissionGuidelines , a page that admittedly seems out of date (it still refers to the players_changelog).

gfgtdf:
I decided to do it a different way which also allows me to remove this code duplication https://github.com/wesnoth/wesnoth/commit/2a5ba5ace74af327bf73afdfbb72e60a6f1a2b6c

stevecotton:
Remember to update the Wiki with the new Lua function.

gfgtdf:
hmm i don't know where to be honest since i couldn't find the locations where the other items.lua functions are documented.

stevecotton:
The other functions in items.lua seem to be implementations of WML tags, documented in InterfaceActionsWML. WML has a `store_items` function, which could probably be refactored to use this.

> Usualyl these thigns get documented in the wiki after they are merged.

I've documented `[store_items]item_name=`, 17 months after it was merged. Stuff like this is why I push for developers to write documentation before merging, so it's quick to cut&paste afterwards.

gfgtdf:
> The other functions in items.lua seem to be implementations of WML tags, 

It also exports functions like wesnoth.interface.remove_item

> WML has a store_items function, which could probably be refactored to use this.

I dont think this would  make it easier since this function converts the lua wml table to a lua table and then if we implemed store_item on top of it we'd need to convert iot back

stevecotton:
Where should UMC developers using Lua look to find documentation of these new functions?

> It also exports functions like wesnoth.interface.remove_item

I included wesnoth.interface.remove_item in "implementation of wml tags". While there's nothing (AFAIK) that documents the existence of the Lua function, it has the exact same name as WML tag, and the exact same quirk (that the single parameter for matching is tested against all of image, name and halo). While it would be better if it was documented, any UMC developer wondering what the Lua equivalent of WML's remove_item is has a good chance of finding it - either by guessing or by checking how the WML tag is implemented.

On the other hand, that file also exports wesnoth.interface.place_halo, for which there's no caller anywhere in mainline, nor documentation on the Wiki. How are UMC authors expected to find it, or wesnoth.interface.get_items?

gfgtdf:
> hmm i don't know where to be honest since i couldn't find the locations where the other items.lua functions are documented.

I just randomly stumbled over the answer to this question https://wiki.wesnoth.org/LuaWML/Tiles#items.remove

CelticMinstrel:
I think the _correct_ place for this documentation is [LuaAPI/wesnoth/interface](https://wiki.wesnoth.org/LuaAPI/wesnoth/interface). We should be moving all the documentation over to the LuaAPI pages anyway.

Also, I think allowing for location filters to check items as I mentioned in my previous comment is also something that should be done; is there another existing feature request for it?

gfgtdf:
> is there another existing feature request for it?

No

PR #5092, opened 2020-08-20
Update the tournaments text every hour automatically.

Pentarctagon:


src/server/wesnothd/server.cpp
```diff
@@ -1004,7 +1025,7 @@ void server::add_player(socket_ptr socket, const wesnothd::player& player)
 		send_server_message(socket, motd_+'\n'+announcements_, "motd");
```
soliton-, line 1025: Tournament announcements are sent with the motd now and I'd keep it that way.

soliton- (approved):

PR #5094, opened 2020-08-22
update DID_SHADOW_VEIL_HANDLER in DID campaign

newfrenchy83:
simplify DID_SHADOW_VEIL_HANDLER  in DiD and implement an did_shadow_veil_helper active in radius 5 of shadow_veil ability

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
+++ b/data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
@@ -140,7 +140,7 @@
 #define ABILITY_SHADOW_VEIL_HELPER
     [hides]
         id=did_shadow_veil_helper
-		[filter]
+        [filter]
             [filter_location]
                 radius=5
                 [filter]
@@ -150,23 +150,23 @@
                             side=$unit.side
                         [/allied_with]
                     [/filter_side]
-				[/filter]
+                [/filter]
             [/filter_location]
-		[/filter]
+        [/filter]
     [/hides]
 #enddef
 
 #define DID_SHADOW_VEIL_HANDLER
     [event]
-	    id=did_shadow_veil_unit_placed
+        id=did_shadow_veil_unit_placed
         name=unit placed
         first_time_only=no
         [filter]
             side=1
             race=undead
-			[not]
-			    ability=did_shadow_veil_helper
-			[/not]
+            [not]
+                ability=did_shadow_veil_helper
+            [/not]
         [/filter]
         [filter_condition]
             [have_unit]
@@ -174,18 +174,18 @@
                 side=1
             [/have_unit]
         [/filter_condition]
-		[modify_unit]
-		    [filter]
-		        x,y=$x1,$y1
-		        [not]
-		            ability=did_shadow_veil_helper
-					[or]
-					    ability=did_shadow_veil
-					[/or]
-		        [/not]
-		    [/filter]
+        [modify_unit]
+            [filter]
+                x,y=$x1,$y1
+                [not]
+                    ability=did_shadow_veil_helper
+                    [or]
+                        ability=did_shadow_veil
+                    [/or]
+                [/not]
+            [/filter]
             [object]
-			    duration=scenario
+                duration=scenario
                 [effect]
                     apply_to=new_ability
                     [abilities]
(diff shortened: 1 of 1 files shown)


nemaara:
Not sure if this is the fault of my original code or this update, but somehow the ability doesn't work on skeletons. I'm not sure why though, it seems to me like they should pass the filter. Might be an engine issue too or something.

CelticMinstrel:
This change does not look good. Removing the `ABILITY_SHADOW_VEIL_HELPER` macro is probably not the right approach, and replacing an `[object]` with a `[set_variable]` is almost definitely not the right approach. I don't fully understand what's _supposed_ to happen here, but wouldn't something like this work better?

```
#define ABILITY_SHADOW_VEIL_HELPER
  [hides]
        id=did_shadow_veil_helper
        affect_self=yes # Maybe this is not needed? I'm not sure.
        name= _ "veiled"
        description= _ "This unit is hidden."
        [filter_location]
            radius=5
            [filter]
                ability=did_shadow_veil
                [filter_side]
                    [allied_with]
                        side=$unit.side
                    [/allied_with]
                [/filter_side]
        [/filter_location]
    [/hides]
#enddef
```

Then the actual implementation could either go back to the original:

```
[object]
    take_only_once=no
    silent=yes
    duration=forever
    [filter]
        x,y=$x1,$y1
    [/filter]
    [effect]
        apply_to=new_ability
        [abilities]
            {ABILITY_SHADOW_VEIL_HELPER}
        [/abilities]
    [/effect]
[/object]
```

Or use `[modify_unit]`:

```
[modify_unit]
    [object]
        [filter]
            x,y=$x1,$y1
        [/filter]
        [effect]
            apply_to=new_ability
            [abilities]
                {ABILITY_SHADOW_VEIL_HELPER}
            [/abilities]
        [/effect]
    [/object]
[/modify_unit]
```

Is there any reason why this wouldn't work?

newfrenchy83:
> 
> 
> This change does not look good. Removing the `ABILITY_SHADOW_VEIL_HELPER` macro is probably not the right approach, and replacing an `[object]` with a `[set_variable]` is almost definitely not the right approach. I don't fully understand what's _supposed_ to happen here, but wouldn't something like this work better?
> 
> ```
> #define ABILITY_SHADOW_VEIL_HELPER
>   [hides]
>         id=did_shadow_veil_helper
>         affect_self=yes # Maybe this is not needed? I'm not sure.
>         name= _ "veiled"
>         description= _ "This unit is hidden."
>         [filter_location]
>             radius=5
>             [filter]
>                 ability=did_shadow_veil
>                 [filter_side]
>                     [allied_with]
>                         side=$unit.side
>                     [/allied_with]
>                 [/filter_side]
>         [/filter_location]
>     [/hides]
> #enddef
> ```
> 
> Then the actual implementation could either go back to the original:
> 
> ```
> [object]
>     take_only_once=no
>     silent=yes
>     duration=forever
>     [filter]
>         x,y=$x1,$y1
>     [/filter]
>     [effect]
>         apply_to=new_ability
>         [abilities]
>             {ABILITY_SHADOW_VEIL_HELPER}
>         [/abilities]
>     [/effect]
> [/object]
> ```
> 
> Or use `[modify_unit]`:
> 
> ```
> [modify_unit]
>     [object]
>         [filter]
>             x,y=$x1,$y1
>         [/filter]
>         [effect]
>             apply_to=new_ability
>             [abilities]
>                 {ABILITY_SHADOW_VEIL_HELPER}
>             [/abilities]
>         [/effect]
>     [/object]
> [/modify_unit]
> ```
> 
> Is there any reason why this wouldn't work?

@gfgtdf this approach are better, but using same object for many unit will make savegames heavy and slow the game but like my first approeach in engine will can be worst, i will be using your code, but replace duration=forever by duration=scenario

newfrenchy83:
> 
> 
> Not sure if this is the fault of my original code or this update, but somehow the ability doesn't work on skeletons. I'm not sure why though, it seems to me like they should pass the filter. Might be an engine issue too or something.

@nemaara the skeletons have already an hides ability in water, i think what my update create a bug, i must change that.

gfgtdf:
> @gfgtdf this approach are better, but using same object for many unit will make savegames heavy and slow the game but like my first approeach in engine will can be worst, i will be using your code, but replace duration=forever by duration=scenario

This was celmins code, not mine, but i agree with what he said, and also about using duration=scenario, furthermroe i'd probalby use something like [filter][not] ability=did_shadow_veil_helper in the [modify_unit] tag to amke sure the object is no applied twice.

newfrenchy83:
@gfgtdf , @CelticMinstrel and @nemaara it is done.

CelticMinstrel:
Just one more thing - why did you remove the name and description for the "veiled" ability? Is there a reason for that?

newfrenchy83:
@CelticMinstrel because unit is 'veiled' what if he is in radius=5 of Mal Keshar when he have shadow_veil, if unit have description of veiled ability then what he is not hidden, this will be contracditory, and unuse.

CelticMinstrel:
I have no idea what you just said, but I assume the purpose of the name/description is to indicate to the player that the opponent can't see that unit, or something to that effect.

newfrenchy83:
@CelticMinstrel  Using the description to indicate that the unit is hidden or not makes sense when using event name = moveto to add / remove the ability depending on the position, but with unit placed the description remains active even when the unit is visible, and the invisibility icon is sufficient and appears on the map when a hex is selected within range of the unit with did_shadow_veil, whereas with the current configuration, it is only after the movement made that the unit becomes invisible, and you have to measure the distance each time to know what will happen. That's why I preferred to update it to make it look more 'natural' like with ambush ability.

CelticMinstrel:
Well, it still seems wrong for a PR focusing on mechanics to be removing flavour text, but I guess it's mostly up to @nemaara whether that's okay.

nemaara:
It's fine to remove it because it's already indicated by the hidden status. As long as the ability works and the code looks fine, then it doesn't matter to me about changing the implementation.

PR #5100, opened 2020-08-26
WC: Implement internationalization with a mainline-style textdomain

shikadiqueen:
Although World Conquest was added to mainline in 1.15.4, apparently people forgot that mainline content needs translations. 😛

This PR takes care of renaming WC's textdomain to follow the format used for mainline textdomains, as well as setting up the necessary infrastructure for generating and updating catalogues. Based on discussions from around the time WoV was added, I understand that it's expected that the first pot update (included in this PR) will preseed the British English catalogue with the source strings. The Shavian script catalogue also gets preseeded since it's technically an English locale, but I blanked it out like it was done with WoV before since, you know, it uses a completely different and incompatible script.

I'm awaiting confirmation from @gfgtdf that the new textdomain name is appropriate and that there are no pre-existing translations of the add-on before considering the PR complete so I can merge it along with a gettext.w.o update. I'd also like @ivanovic and @loonycyborg to confirm that I didn't miss anything wrt CMake and SCons, and @Wedge009 to confirm that the preseeding of the British English catalogue is still desirable.


gfgtdf:
> 'm awaiting confirmation from @gfgtdf that the new textdomain name is appropriate and that there are no pre-existing translations of the add-on before considering the PR complete so I can merge it along with a gettext.w.o update

I think the name is appropriate and there are no pre-existing translations of the add-on

gfgtdf:
I want to note that, there are still some strings in the training and artifact code that i want to change, basically the campaigns often builds strings like "+<number> defence on <terrain>" and i can either use string concatination (`_"+" .. percentage .. _" defence on " .. terrain `) or wesnoth.vformat ( `_("+$number percentage on $terrain"):vformat { number= number, terrain= terrain} `) for it.


I'm not yet sure which of these two is the better way. Because: i think the  `vformat` way is more elegant but on the other hand the engine supports concatination of strings natively, so if i use the concatination  method, the resulting objects will stay tstring  objects.

shikadiqueen:
Format strings should always be used instead of concatenation when available, to avoid making translators' lives impossible. Remember that most languages do not use the exact same grammatical order as English, so reordering of construct components like that is often necessary.

Also "defence" is British English. Mainline uses US English modulo lore exceptions ("sceptre", "spectre").

Wedge009:
Yes, I've tried to leave the base strings as US dialect. In most places I saw 'defense'.

loonycyborg (approved):

gfgtdf (approved):

ivanovic (approved):

Wedge009 (approved):

PR #5102, opened 2020-08-26
Update World Conquest II Strings

Wedge009:
**Before** merging this I have some questions about some phrases in `data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help.lua` that I didn't know how to decipher (from a native English speaker's point of view):

Line 39:
> Carryover is 15%, comunitary and avoid negative amounts.

I don't know what 'comunitary' is supposed to mean. Communal? Also what does 'carry-over avoid negative amounts' supposed to mean? Gold will always be non-negative at the start of a scenario?

Line 45:
> Every chance will do different dice rolls

This is in the context of units' training. Every chance of what? Every time a new unit is recruited there will be a different chance of being trained?

Line 68 (partially edited):
> The World Conquest II era consists of factions that are build of pairs of mainline factions of which one has a healer available (Drakes, Rebels and Loyalists), and one does not (Orcs, Dwarves and Undead) the recruit list is also organized in pairs so that you sometimes have to recruit a different unit before you can recruit the units that you want. The available heroes, desertes, and random leaders also depend on your factions, the items you can get do not depend on the faction you choose.

I didn't understand what was meant by 'factions that are build of pairs of mainline factions'. Also... 'desertes'?

I'm guessing @gfgtdf as author (?) can help with the review. I found these while reviewing #5100 for @shikadiqueen.

### DIFF
--- a/data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
+++ b/data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
@@ -2,11 +2,11 @@ local _ = wesnoth.textdomain 'wesnoth-wc'
 
 type_infos = {
 	["default"] = {
-		founddialogue = _"You guys look like you could use some help. Mind if I join in? It's been a while since I had a good fight!",
+		founddialogue = _"You look like you could use some help. Mind if I join in? It’s been a while since I had a good fight!",
 		reply = _ "Excellent. We could always use more help.",
 	},
 	["Orcish Grunt"] = { 
-		founddialogue=_"'bout time. Been forever since I had a good fight, eh?",
+		founddialogue=_"’bout time. Been forever since I had a good fight, eh?",
 		image="units/orcs/grunt.png",
 		name="Orcish Grunt",
 	},
@@ -21,7 +21,7 @@ type_infos = {
 		name="Orcish Archer",
 	},
 	["Orcish Assassin"] = { 
-		founddialogue=_"Heya, boss. Looks like you've got some guys need killing.",
+		founddialogue=_"Hey boss. Looks like you’ve got some prey that need killing.",
 		image="units/orcs/assassin.png",
 		name="Orcish Assassin",
 	},
@@ -31,11 +31,11 @@ type_infos = {
 		name="Wolf Rider",
 		alt_reply = { { 
 			race="wose",
-			reply=_"Ok. But take care where your dog does its stuff.",
+			reply=_"Very well. But take care where your dog does its stuff.",
 		} }
 	},
 	["Orcish Leader"] = { 
-		founddialogue=_"Heh, looks like you whelps might be getting in over your heads. Good thing I'm here now to win this fight for ya, huh?",
+		founddialogue=_"Heh, looks like you whelps might be getting in over your heads. Good thing I’m here now to win this fight for ya, huh?",
 		image="units/orcs/leader.png",
 		name="Orcish Leader",
 	},
@@ -58,7 +58,7 @@ type_infos = {
 		name="Elvish Fighter",
 	},
 	["Elvish Archer"] = { 
-		founddialogue=_"You guys look like you could use some help. Mind if I join in? It's been a while since I had a good fight!",
+		founddialogue=_"You look like you could use some help. Mind if I join in? It’s been a while since I had a good fight!",
 		image="units/elves-wood/archer.png",
 		name="Elvish Archer",
 	},
@@ -68,7 +68,7 @@ type_infos = {
 		name="Elvish Shaman",
 		alt_reply = { { 
 			race="elf,wose",
-			reply=_"Yea, flower power!",
+			reply=_"Yeah, flower power!",
 		} }
 	},
 	["Elvish Scout"] = { 
@@ -100,7 +100,7 @@ type_infos = {
 		} }
 	},
 	["Cavalryman"] = { 
-		founddialogue=_"You're not from around here, but I seem to find myself between employers at the moment and I'm not picky. I'll fight for you, if you'll have me.",
+		founddialogue=_"You’re not from around here, but I seem to find myself between employers at the moment and I’m not picky. I’ll fight for you, if you’ll have me.",
 		image="units/human-loyalists/cavalryman/cavalryman.png~CROP(14,14,72,72)",
 		name="Cavalryman",
 		alt_reply = { { 
@@ -123,12 +123,12 @@ type_infos = {
 		name="Spearman",
 	},
 	["Fencer"] = { 
-		founddialogue=_"Looks like you're a bit down on your luck, my friends. But now that I am here, there's nothing to worry about!",
+		founddialogue=_"Looks like you’re a bit down on your luck, my friends. But now that I am here, there’s nothing to worry about!",
 		image="units/human-loyalists/fencer.png",
 		name="Fencer",
 	},
 	["Heavy Infantryman"] = { 
-		founddialogue=_"Finally reinforcements are here! I've been pinned down for days. Help me fight my way out of here and I'll gladly follow you!",
+		founddialogue=_"Finally reinforcements are here! I’ve been pinned down for days. Help me fight my way out of here and I’ll gladly follow you!",
 		image="units/human-loyalists/heavyinfantry.png",
 		name="Heavy Infantryman",
 	},
@@ -138,7 +138,7 @@ type_infos = {
 		name="Bowman",
 	},
 	["Sergeant"] = { 
-		founddialogue=_"You're not from around here, but I seem to find myself between employers at the moment and I'm not picky. I'll fight for you, if you'll have me.",
+		founddialogue=_"You’re not from around here, but I seem to find myself between employers at the moment and I’m not picky. I’ll fight for you, if you’ll have me.",
 		image="units/human-loyalists/sergeant.png",
 		name="Sergeant",
 	},
@@ -157,12 +157,12 @@ type_infos = {
 		} }
 	},
 	["Dwarvish Fighter"] = { 
-		founddialogue=_"Having trouble, eh? Never worry, lads, we'll sort 'em out soon enough!",
+		founddialogue=_"Having trouble, eh? Never worry, lads, we’ll sort ’em out soon enough!",
 		image="units/dwarves/fighter.png",
 		name="Dwarvish Fighter",
 	},
 	["Thief"] = { 
-		founddialogue=_"You've got me, guv, it's a fair cop! Just lemme work for you instead. You won't regret it, guv, I promise!",
+		founddialogue=_"You’ve got me, guv, it’s a fair cop! Just lemme work for you instead. You won’t regret it, guv, I promise!",
 		image="units/human-outlaws/thief.png",
 		name="Thief",
 	},
@@ -172,7 +172,7 @@ type_infos = {
 		name="Dwarvish Thunderer",
 	},
 	["Poacher"] = { 
-		founddialogue=_"What, you want my help? A guy like me? Huh, that's rich. Oh well... let's give it a shot, eh?",
+		founddialogue=_"What, you want my help? A guy like me? Huh, that’s rich. Oh well... let’s give it a shot, eh?",
 		image="units/human-outlaws/poacher.png",
 		name="Poacher",
 	},
@@ -182,7 +182,7 @@ type_infos = {
 		name="Dwarvish Guardsman",
 	},
 	["Footpad"] = { 
-		founddialogue=_"Hey, hey, easy there! I done nothin' to hurt you. We're all friends here, right? Looks like you might be in a tight spot, but don't worry. No one's better at getting out of tight spots than me, boss!",
+		founddialogue=_"Hey, hey, easy there! I done nothin’ to hurt you. We’re all friends here, right? Looks like you might be in a tight spot, but don’t worry. No one’s better at getting out of tight spots than me, boss!",
 		image="units/human-outlaws/footpad.png",
 		name="Footpad",
 	},
@@ -199,7 +199,7 @@ type_infos = {
 		} }
 	},
 	["Gryphon Rider"] = { 
-		founddialogue=_"Need a hand? Me an' me bird can get just about anywheres you need.",
+		founddialogue=_"Need a hand? Me an’ me bird can get just about anywheres you need.",
 		image="units/dwarves/gryphon-rider.png",
 		name="Gryphon Rider",
 		alt_reply = { { 
@@ -211,7 +211,7 @@ type_infos = {
 		} }
 	},
 	["Dwarvish Scout"] = { 
-		founddialogue=_"Having trouble, eh? Never worry, lads, we'll sort 'em out soon enough!",
+		founddialogue=_"Having trouble, eh? Never worry, lads, we’ll sort ’em out soon enough!",
 		image="units/dwarves/scout.png",
 		name="Dwarvish Scout",
 	},
@@ -231,7 +231,7 @@ type_infos = {
 		name="Drake Burner",
 		alt_reply = { { 
 			race="drake",
-			reply=_"Perfect. We can always use more fire power.",
+			reply=_"Perfect. We can always use more firepower.",
 		} }
 	},
 	["Saurian Augur"] = { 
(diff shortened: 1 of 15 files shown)


shikadiqueen:
I'm not familiarized with the campaign myself, but from a cursory glance at the file I'm guessing that "deserters" was meant where it says "desertes".

Wedge009:
Deserters was my first thought too, but it didn't make sense to me in the context of heroes and leaders (units). Are deserters part of the new mechanics in this campaign?

gfgtdf:
> Are deserters part of the new mechanics in this campaign?

Yes, you can find them in bonus points and they'll join your army. They are a special type of heroes basicially.

gfgtdf:
> I don't know what 'comunitary' is supposed to mean. Communal? 

It means the carryover is evenly split over the players, if say the players end a scenario with 30,100, and 200 then it will take the sum, devide it by 3 and then take   15% of that so in the next scenario so in out case every player will start the next scenario with (330)/3 * 0.15 = 16 extra gold.  

> Also what does 'carry-over avoid negative amounts' supposed to mean? Gold will always be non-negative at the start of a scenario?

That you cannot carry over negative amounts of gold.

> Every chance will do different dice rolls

think It basically means that it works as one would expect: the probabilities to get ertain training are independent of each other.

> I didn't understand what was meant by 'factions that are build of pairs of mainline factions'.

Well each factions in the WC2 era lets you recruits the units of two mainline factions, for example "The Cult" faction allows you to recruit loyalists and undead units.

stevecotton:
It still needs conversion of apostrophes to typographical-style ones.

data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help.lua
```diff
@@ -53,19 +53,19 @@ function wesnoth.wml_actions.wc2_show_wocopedia(cfg)
 		local str_cat_feedback = _ "Feedback"
 		local str_des_feedback =
 			_ "<b>Feedback</b>:\n" ..
-			_ "For feedback plase either post in the Word conquest II thread in the official wesnoth forum https://r.wesnoth.org/t39651 or file an issue at github https://github.com/gfgtdf/World_Conquest_II/issues .\n\n" ..
+			_ "For feedback please either post in the World Conquest II thread in the official Wesnoth forum https://r.wesnoth.org/t39651 or file an issue at GitHub https://github.com/gfgtdf/World_Conquest_II/issues .\n\n" ..
```
soliton-, line 56: This should presumably be rewritten for mainline if it's still useful at all.

data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help.lua
```diff
@@ -53,19 +53,19 @@ function wesnoth.wml_actions.wc2_show_wocopedia(cfg)
...
 		local str_cat_era = _ "Factions"
-		local str_des_era = _ "<b>Factions</b>\n The Word Conquest 2 era consists of faction that are build of pairs of mainline faction of which at one has a healer available (Drakes, Rebels and Loyalists), and one does not (Orcs, Dwarves and Undead) the recruilist is also organized in pairs so that you sometimes have to recruit a different units before you can recruit the units that you want. The available heroes, desertes and random leaders also depend on your factions, the items you can get do not depend on the faction you choose."
+		local str_des_era = _ "<b>Factions</b>\n The World Conquest II era consists of factions that are build of pairs of mainline faction of which at one has a healer available (Drakes, Rebels and Loyalists), and one does not (Orcs, Dwarves and Undead) the recruit list is also organized in pairs so that you sometimes have to recruit a different unit before you can recruit the units that you want. The available heroes, desertes, and random leaders also depend on your factions, the items you can get do not depend on the faction you choose."
```
soliton-, line 68: I think the 2/II should be removed in all user-facing strings.

Wedge009: I recall the discussion in #wesnoth-dev about WC vs WC2 but my focus here is simply on updating things that seem/look nonsensical to me. Mainly spelling errors but also things that I just didn't understand as an English speaker. (As an aside, WC2 means something completely different to me altogether.)

data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help.lua
Wedge009: Not within the scope of this PR.

Wedge009:
Thanks everyone for the input. I've updated the PR with typographical apostrophes and revised the WoCopedia Help based on @gfgtdf's clarification. I believe it's ready for full review and (if accepted) merge now.

data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
```diff
@@ -2,11 +2,11 @@ local _ = wesnoth.textdomain 'wesnoth-World_Conquest'
 
 type_infos = {
 	["default"] = {
-		founddialogue = _"You guys look like you could use some help. Mind if I join in? It's been a while since I had a good fight!",
+		founddialogue = _"You guys look like you could use some help. Mind if I join in? It’s been a while since I had a good fight!",
 		reply = _ "Excellent. We could always use more help.",
 	},
 	["Orcish Grunt"] = { 
-		founddialogue=_"'bout time. Been forever since I had a good fight, eh?",
+		founddialogue=_"’bout time. Been forever since I had a good fight, eh?",
```
stevecotton, line 9: It's not an issue for the current PR to solve, but wmlxgettext doesn't pick up any hints about context from the coding style.

Commenting on this line because, seeing only the string without context, I thought this sounded like a Fencer rather than an Orc.

data/campaigns/World_Conquest/resources/strings/effects.cfg
```diff
@@ -82,7 +82,7 @@ _ "terrain^water" #enddef
 _ "improved defenses" #enddef
 
 #define STR_LIVING
-_ "no undead^consumed by a living beign:" #enddef
+_ "no undead^consumed by a living being:" #enddef
 
 #define STR_UP_TO VALUE
 _ "up to"+" {VALUE}% " #enddef
```
stevecotton, line 88: The character between "up" and "to" is a non-breaking space, rather than a normal space.

data/campaigns/World_Conquest/lua/map/scenario_utils/plot.lua
```diff
@@ -45,29 +45,29 @@ function add_plot(scenario, scenario_num, nplayers)
...
 		end_message("1,2,3", false, _ "We win! Truly ours will be a glorious empire!")
 		end_message("1,2,3", true, _ "Do not get hasty! To reach the great continents of the east we must subdue the savage coast of Moragdu, notorious for its corsairs, before we can advance further.")
 	elseif scenario_num == 3 then
-		start_message("4", true, _ "What's this? It is a long time since strangers have set foot upon Moragdu--and in such great numbers too!")
+		start_message("4", true, _ "What’s this? It is a long time since strangers have set foot upon Moragdu--and in such great numbers too!")
 		start_message("5", true, _ "The time of prophecies is upon us! The destroyers have arrived from beyond the sea!")
-		start_message("6", true, _ "I don't know what the lizard is babbling about, but we'll sort these 'destroyers' out quickly enough.")
+		start_message("6", true, _ "I don’t know what the lizard is babbling about, but we’ll sort these ‘destroyers’ out quickly enough.")
 
 		end_message("1,2,3", true, _ "Victory is ours! Let us take sail in search of new lands to conquer!")
```
stevecotton, line 58: Should "take sail" be "set sail"?

data/campaigns/World_Conquest/lua/map/scenario_utils/plot.lua
```diff
@@ -45,29 +45,29 @@ function add_plot(scenario, scenario_num, nplayers)
...
-		start_message("4", true, _ "What's this?! Foreign invaders have set foot upon our shores!")
-		start_message("5", true, _ "We'll send them back to where they came from quick enough. All troops, to me!")
+		start_message("4", true, _ "What’s this?! Foreign invaders have set foot upon our shores!")
+		start_message("5", true, _ "We’ll send them back to where they came from quick enough. All troops, to me!")
 
 		end_message("1,2,3", false, _ "We win! Truly ours will be a glorious empire!")
 		end_message("1,2,3", true, _ "Do not get hasty! To reach the great continents of the east we must subdue the savage coast of Moragdu, notorious for its corsairs, before we can advance further.")
 	elseif scenario_num == 3 then
-		start_message("4", true, _ "What's this? It is a long time since strangers have set foot upon Moragdu--and in such great numbers too!")
+		start_message("4", true, _ "What’s this? It is a long time since strangers have set foot upon Moragdu--and in such great numbers too!")
```
stevecotton, line 54: That double-hyphen should change. IIUC, the replacement should be a spaced em dash.

data/campaigns/World_Conquest/lua/map/scenario_utils/plot.lua
```diff
@@ -45,29 +45,29 @@ function add_plot(scenario, scenario_num, nplayers)
...
 		start_message("4", true, _ "Word has come to our island of your victories. Your army is impressive, but it will avail you not.")
 		start_message("6", true, _ "Indeed. You come this far, and no farther!")
-		start_message(r, true, _ "Hah! We didn't come all this way just to give up and go home. We will destroy you as we have destroyed all that have stood before us!")
+		start_message(r, true, _ "Hah! We didn’t come all this way just to give up and go home. We will destroy you as we have destroyed all that have stood before us!")
 		start_message("7", true, _ "So be it. We stand squarely behind our allies.")
-		start_message("2,3", true, _ "You are not the only ones with allies! We'll push on till the end!")
+		start_message("2,3", true, _ "You are not the only ones with allies! We’ll push on till the end!")
 
 		end_message("1,2,3", true, _ "Onwards to the ships!")
 		end_message("1,2,3", false, _ "We have travelled great oceans and brought low mighty empires. Should we not wait until regain our forces?")
```
stevecotton, line 69: Should we not wait until *we* regain our forces?

I think adding the missing word helps, but it still doesn't sound right. Maybe "until our forces regain their strength".

stevecotton (commented):
In the Git commit message, please leave a blank line between the first and third lines, otherwise some Git programs concatenate all the lines of the first paragraph together (and that's way too long for some UIs).

World_Conquest/lua/campaign/objectives.lua line 10: missing a hyphen in "rightclick".

Wedge009:
Sorry, I normally do that but it was my first time using `git rebase` from the command-line.

Regarding right-click, I must have missed that though I already hyphenated it for GB locale. I notice a tendency for concatenation of words in US locale anyway.

data/campaigns/World_Conquest/resources/strings/effects.cfg
Wedge009: Maybe that was deliberate? I don't know. Changed, as requested.

data/campaigns/World_Conquest/lua/map/scenario_utils/plot.lua
Wedge009: I assumed it was a nautical term, like 'take in the sails', but yes, probably makes more sense to be set sail.

data/campaigns/World_Conquest/lua/map/scenario_utils/plot.lua
Wedge009: Wasn't focused on punctuation as much as spelling, plus everyone seems to follow different conventions. My preference is for spaced en dash while others use non-spaced em dash, but I'll change per your suggestion.

data/campaigns/World_Conquest/lua/map/scenario_utils/plot.lua
Wedge009: Thanks for highlighting this, I didn't read everything super-thoroughly. Agreed on your rewording.

data/campaigns/World_Conquest/resources/strings/effects.cfg
```diff
@@ -82,10 +82,10 @@ _ "terrain^water" #enddef
 _ "improved defenses" #enddef
 
 #define STR_LIVING
-_ "no undead^consumed by a living beign:" #enddef
+_ "no undead^consumed by a living being:" #enddef
 
 #define STR_UP_TO VALUE
-_ "up to"+" {VALUE}% " #enddef
+_ "up to"+" {VALUE}% " #enddef
```
gfgtdf, line 88: Whats the reason for this change ?

Wedge009: @stevecotton remarked it stood out. What's the reason for the non-break?

gfgtdf: To make the dialog where this text appears in looks better,  there are quite a few other places where this addons uses nonbreaking strings, and it does that intentionally. ( https://github.com/gfgtdf/World_Conquest_II/commit/6c352fb216853308b1b8c71e5870a1675ca0a889 ) .

gfgtdf: Actually i'm not 100% sure where  utils/gameplay/training.cfg  still has them becasue i changed tehse strigns at some point and i migth ahve accidentatlyl dropped them here (which  which be bad because that's the main place where they are needed.)

data/campaigns/World_Conquest/resources/strings/effects.cfg
Wedge009: See @gfgtdf's remarks below.

gfgtdf:
I'm not exactly sure what to do about the nonbreakable string in particular maybe we should adda  comment or translators might oversee them, but lets leave that to a seperate pr/issue.

Wedge009:
Sure, after all, I understand you're the author. This only started as a means to fix obvious spelling mistakes and somehow it's become a full string review...

shikadiqueen:
Is it not possible to use the XML escaping syntax for non-breakable space (`&#160;`) instead? Otherwise whoever next edits that code could accidentally replace it with regular 0x20 spaces.

shikadiqueen:
A couple of things:

* The PR doesn't fix several more instances of "Ok" used in UI code (sometimes not translated). At least one of them (WoCopedia) should use "Close" instead since the dialog does not have a "Cancel" option alongside it.
* I am not entirely sure about the replacement of "Ok" with "Okay" in game prose (as opposed to UI). So far we had made an active effort to avoid this and other modern words in prose.

Wedge009:
Again, I was only approaching this from the perspective of an English speaker. In my experience (and others may disagree), 'Ok' doesn't make sense or at the very least I would consider to be non-standard. Can either have 'OK' (especially in dialogue buttons like 'OK' vs 'Cancel) or 'Okay/okay' in prose. I vaguely recall writing style guidelines but that wasn't what I had in mind when I made those changes. Nonetheless, I'll try to review this tomorrow.

Wedge009:
Have replaced the 'Ok' in dialogues and prose. Hope that's satisfactory - let me know if I missed any.

Regarding 'OK' vs 'Close', is it okay to have label as 'Close' while leaving the ID as 'ok'? This is also beyond the scope of the PR so I have left it as-is for now.

shikadiqueen:
> Regarding 'OK' vs 'Close', is it okay to have label as 'Close' while leaving the ID as 'ok'?

Yes, and this is in fact required since the UI engine handles special knowledge about widgets with id of `ok`.

Wedge009:
Looking at the text overall, it seems to take a more informal approach as compared with the normal, story-based campaigns. Nonetheless, given the 'modern' language use brought up previously I removed the use of 'guys' as the only update since the last push a few days ago.

Just to clarify, I replaced 'heya boss' with 'hey boss' not as a greeting (as seems to be used a lot in current times) but as the older meaning (at least the one I am accustomed to) of the use of 'hey' as a means to gain attention. Perhaps impolitely, especially in the context of a superior, but these are orcs and discipline isn't what they are normally know for so I think it's okay.

shikadiqueen:
Would it be appropriate to have @nemaara take a look as well since they handle string-heavy content in SP?

Wedge009:
If she has time, sure. Just keeping in mind that, for my work here, I'm mainly concerned about correcting spelling errors and such, not rewriting the entire thing to be 'up to standard'.

Wedge009:
Haven't heard anything for a few days - any objections before merging? Does it have to be approved by someone specific under the new contract?

PR #5105, opened 2020-08-26
Resend the announcements and server info text when rejoining the lobby.

### LINKED ISSUE #5095: Server Information bug.
 - 1.15.4
 - Steam
 - Windows 10

### Describe the bug
Bug

### To Reproduce
1. Log in
2. Enter any game
3. Quit to Main Menu
4. Check Information panel

### Expected behavior
Przechwytywanie

Pentarctagon:
Fixes #5095

### DIFF
--- a/src/gui/dialogs/multiplayer/lobby.cpp
+++ b/src/gui/dialogs/multiplayer/lobby.cpp
@@ -116,6 +116,9 @@ bool mp_lobby::logout_prompt()
 	return show_prompt(_("Do you really want to log out?"));
 }
 
+std::string mp_lobby::announcements_ = "";
+std::string mp_lobby::server_information_ = "";
+
 mp_lobby::mp_lobby(const game_config_view& game_config, mp::lobby_info& info, wesnothd_connection &connection)
 	: quit_confirmation(&mp_lobby::logout_prompt)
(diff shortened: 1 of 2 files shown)


soliton-:
I'm not a fan of re-sending the same info to fix this in principle. I'd rather send the info/announcements again when they actually changed.

If we must do a server-side fix first then it should probably rather be done next to the player_record::enter_lobby calls (stuff like `player_connections_.get<game_t>().modify(it, std::bind(&player_record::enter_lobby, _1));`) to make sure we get all cases.

gfgtdf:
> I'm not a fan of re-sending the same info to fix this in principle. I'd rather send the info/announcements again when they actually changed.

I agree here and at least for 1.15 at first sight i see no reason why we shouldn't be able  to fix  this on clientside.

Pentarctagon:
Alright, I'll look at that then.  The main thing I wasn't sure about with a client-side fix was where to store the information rather than in `mp_lobby` in `lobby.*pp` as is done currently.

Pentarctagon:
Updated to make the variables static, so their values are preserved through multiple instances of `mp_lobby`.

soliton- (approved):

Issue #5097, opened 2020-08-23
Crash when loading save file after defeat in local multiplayer game


IceTyp:
### Game and System Information
-1.15.4
-Steam
-Ubuntu 20.04

### Describe the bug
When I tried to load a save file after my leader is defeated in a local multiplayer game, BfW crashes.

### To Reproduce
1. Start a local multiplayer game. (I tried it with World Conquest 1p and also with a Random map against the AI.)
2. Let your leader be defeated.
3. After the Defeat message appears, press space
4. Press Ctrl + o to load a game.
5. Choose any save file and press Enter to load it.

The game will crash.

### Savefiles
Alternatively to step 1, you could also open this save file, end your turn, and proceed with step 2.
[Random map-Auto-Save6.gz](https://github.com/wesnoth/wesnoth/files/5114398/Random.map-Auto-Save6.gz)

### Additional context
I have not tried this for non-random maps. 
I could not reproduce this bug in the campaign DiD.

Pentarctagon:
Ouch.  An even simpler way to reproduce:
* Load the save game
* Quit to the main menu
* Load the same save game
* Wesnoth crashes

The crash is caused by the assert [here](https://github.com/wesnoth/wesnoth/blob/5e5df83db6937f7e140003104c66a3020e4934e1/src/saved_game.cpp#L658) being triggered.

A bisect points to commit 5759404121215730cbd4ec899061317923e0ec6e being the cause.

The backtrace is [here](https://pastebin.com/wUtpjbdE).

Pentarctagon:
As another note, loading a save from within an ongoing game (aka not after being defeated or after going back to the main menu) does not trigger the crash.

Pentarctagon:
Fixed in 5b5a60d0c4969260bdd0bfea59a5e12f074e13d6.

stevecotton:
Reopening - 5b5a60d makes the WML unit tests assert.

The Travis build of 5b5a60d itself failed before that point, due to the utf-8 check on the missing `RELEASE_NOTES`. The crash in the unit tests is visible in https://travis-ci.com/github/wesnoth/wesnoth/jobs/378132819

```
Thread 1 "wesnoth" received signal SIGABRT, Aborted.
[...]

PR #5106, opened 2020-08-28
new lua gui2 widget userdata

gfgtdf:
lua now has a widget userdata that can be used to set/get
widgets properties as one would expect, a lot of the
set/get_dialog_xy function were converted into
modifiable properties of the widget userdata. This
in particular allows us to get rid of the strange
'path to widget' type of arguments of gui2 functions.

This is not as good as i could be though, i'm still unsure about
quite some of the name of the properties, also this does not
yet allow multiple dialog obejcts, furthermore it currently
generates a lot of compiler wanrings


stevecotton:
Please fix the current master branch before developing new functionality on top. All of the unit tests are crashing because of https://github.com/wesnoth/wesnoth/issues/5097#issuecomment-681304026 , which needs a better fix to the bug introduced by your 5759404121215730cbd4ec899061317923e0ec6e.

data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help.lua
```diff
@@ -68,49 +49,71 @@ function wesnoth.wml_actions.wc2_show_wocopedia(cfg)
 		local str_des_era = _ "<b>Factions</b>\n The Word Conquest 2 era consists of faction that are build of pairs of mainline faction of which at one has a healer available (Drakes, Rebels and Loyalists), and one does not (Orcs, Dwarves and Undead) the recruilist is also organized in pairs so that you sometimes have to recruit a different units before you can recruit the units that you want. The available heroes, desertes and random leaders also depend on your factions, the items you can get do not depend on the faction you choose."
 		local str_cat_settings = _ "Settings"
 
+
+
+		local root_node = dialog:find("treeview_topics")
+		local details = dialog:find("details")
+
+		function gui.widget.add_help_page(parent_node, args)
```
CelticMinstrel, line 57: This probably shouldn't be shoved into the `gui.widget` namespace. Is there a really good reason for it? If not, I'd probably suggest making it a local function. On the other hand, even if there _is_ a really good reason, I wonder if we could come up with a better way?

data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help_dialog.lua
```diff
@@ -93,13 +94,13 @@ return {
 							horizontal_grow = true,
 							vertical_grow = true,
 							T.tree_view {
-								id = "left_tree",
+								id = "treeview_topics",
```
CelticMinstrel, line 97: This isn't directly related to the new API, but I think it might be much cleaner if you put the dialog definigion into an actual WML file and loaded it in with `wml.load`... unless it's dynamically-built in a way that would prevent this?

src/scripting/lua_kernel_base.cpp
```diff
@@ -448,21 +450,21 @@ lua_kernel_base::lua_kernel_base()
...
-		{ "set_dialog_active",        &lua_gui2::intf_set_dialog_active		},
-		{ "set_dialog_visible",       &lua_gui2::intf_set_dialog_visible    },
-		{ "add_dialog_tree_node",     &lua_gui2::intf_add_dialog_tree_node	},
-		{ "set_dialog_callback",      &lua_gui2::intf_set_dialog_callback	},
-		{ "set_dialog_canvas",        &lua_gui2::intf_set_dialog_canvas		},
-		{ "set_dialog_focus",         &lua_gui2::intf_set_dialog_focus      },
-		{ "set_dialog_markup",        &lua_gui2::intf_set_dialog_markup		},
-		{ "set_dialog_value",         &lua_gui2::intf_set_dialog_value		},
-		{ "remove_dialog_item",       &lua_gui2::intf_remove_dialog_item    },
+		//{ "get_dialog_value",         &lua_gui2::intf_get_dialog_value		},
```
CelticMinstrel, line 453: These should just be removed entirely, not left commented out.

src/scripting/lua_widget.cpp
```diff
@@ -0,0 +1,205 @@
...
+}
+
+
+bool luaW_iswidget(lua_State* L, int index)
+{
+	return luaL_testudata(L, index, widgetKey) != nullptr;
+}
+
+
+static void luaW_pushwidgettablecontainer(lua_State* L)
```
CelticMinstrel, line 68: These next few functions are kinda confusing. Could we have some sort of explanation of what they're doing, maybe some code comments?

src/scripting/push_check.hpp
```diff
@@ -72,6 +76,21 @@ namespace lua_check_impl
 	template<typename T>
 	using remove_constref = std::remove_const_t<std::remove_reference_t<std::remove_const_t<T>>>;
 
+
+	template<typename T>
```
CelticMinstrel, line 80: This seems sorta useful, but is there any way to add a version of `lua_push` for it as well?

src/scripting/lua_widget_attributes.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+
+/// CLASSIC
+
+WIDGET_GETTER("value_compat,selected_index", int, gui2::listbox)
+{
+	return w.get_selected_row() + 1;
+}
+
+/* idea
```
CelticMinstrel, line 178: This seems like a good idea! Was there some technical reason why it didn't work?

src/scripting/lua_widget_attributes.cpp
```diff
@@ -0,0 +1,527 @@
...
+	if(const unit_type* ut = luaW_tounittype(L, value.index)) {
+		w.set_displayed_type(*ut);
+	} else if(unit* u = luaW_tounit(L, value.index)) {
+		w.set_displayed_unit(*u);
+	} else {
+		luaW_type_error(L, value.index, "unit or unit type");
+	}
+}
+
+WIDGET_GETTER("item_count", int, gui2::multi_page)
```
CelticMinstrel, line 303: I think the `#` operator should also be overloaded to return this value. You can probably just make the operator work for all widgets, returning `item_count` if it exists or 1 otherwise.

src/scripting/lua_widget_attributes.cpp
```diff
@@ -0,0 +1,527 @@
...
+	return w.get_page_count();
+}
+
+WIDGET_SETTER("use_markup", bool, gui2::styled_widget)
+{
+	w.set_use_markup(value);
+}
+
+//TODO: while i think this shortcut is useful, i'm not that happy about
+//      the name since  it changes 'label' and not 'text', the first one
```
CelticMinstrel, line 314: You could just change it to `marked_up_label` then? Or for something a little shorter, maybe `rich_label`?

src/scripting/lua_widget_attributes.cpp
```diff
@@ -0,0 +1,527 @@
...
+			}
+			break;
+		default:
+			luaW_type_error(L, value.index, "boolean or string");
+	}
+
+	w.set_visible(flag);
+
+	//if(flag == visibility::hidden) {
+	//	// HACK: this is needed to force the widget to be repainted immediately
```
CelticMinstrel, line 378: I'm assuming this is a hack that became unnecessary? In that case, why is it even still here?

src/scripting/lua_widget_attributes.cpp
```diff
@@ -0,0 +1,527 @@
...
+WIDGET_SETTER("value_compat,unfolded", bool, gui2::tree_view_node)
+{
+	if(value) {
+		w.unfold();
+	} else {
+		w.fold();
+	}
+}
+
+WIDGET_SETTER("value_compat,unit", lua_index_raw, gui2::unit_preview_pane)
```
CelticMinstrel, line 292: Why is there a setter for this but no getter? That seems like a pretty enormous oversight. The unit preview pane isn't all that useful if you can't even get the unit displayed!

src/scripting/lua_widget_attributes.cpp
```diff
@@ -0,0 +1,527 @@
...
+	w.set_active(value);
+}
+
+WIDGET_SETTER("tooltip", t_string, gui2::styled_widget)
+{
+	w.set_tooltip(value);
+}
+
+
+WIDGET_SETTER("callback", lua_index_raw, gui2::widget)
```
CelticMinstrel, line 334: I don't like calling this `callback`, because the actual meaning of it depends on the widget, doesn't it? I'd prefer descriptive names beginning with `on`, for example `onchange` or `onclick`.

Or is this a compatibility layer similar to `value_compat`? In that case, I can't help wondering why it doesn't also have `_compat` in its name... plus, if that's the case, you shouldn't've used it in the zombie dialog.

src/scripting/lua_widget_attributes.cpp
```diff
@@ -0,0 +1,527 @@
...
+	if(!wd) {
+		throw std::invalid_argument("the widget has no window assigned");
+	}
+	lua_pushvalue(L, value.index);
+	if (!luaW_setwidgetcallback(L, &w, wd, "on_left_click")) {
+		connect_signal_notify_modified(w, std::bind(&dialog_callback, L, lua_ptr<gui2::widget>(w), "callback"));
+	}
+}
+
+WIDGET_SETTER("on_button_click", lua_index_raw, gui2::widget)
```
CelticMinstrel, line 450: I don't understand the difference between `on_left_click` and `on_button_click`. They sound like they both do the same thing?

src/scripting/lua_widget_attributes.cpp
```diff
@@ -0,0 +1,527 @@
...
+	if(it != setters.end()) {
+		for(const auto& func : it->second) {
+			if(func(L, 3, w)) {
+				return 0;
+			}
+		}
+		ERR_LUA << "none of "<< it->second.size() << " setters macthed\n";
+	}
+	else {
+		ERR_LUA << "unknown peopertly id : " << str << " #known properties="  << setters.size() << "\n";
```
CelticMinstrel, line 521: This is just one of many typos in your code... it's one thing to have a typo in a comment, but you should really spell-check log messages and error messages at least.

CelticMinstrel (commented):
I realize this is a bit late for a review, but there are a number of small issues that I've noticed in this new API, as indicated by the individual review comments.

On top of that, it really desperately needs detailed documentation, preferably on [LuaAPI/Gui](https://wiki.wesnoth.org/LuaAPI/gui) and a new page [LuaAPI/Gui/Widget](https://wiki.wesnoth.org/LuaAPI/gui/widget).

On the whole though, it seems like a major improvement over the previous API, so thanks for taking the time to implement it!

data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help.lua
gfgtdf: gui.widget is (more or less) the widget metatale , this way i can use it like widget:add_help_page.  I think its fine the way its done here tbh, its true that this function only works inside this file but i don't really see this as a problem.

src/scripting/lua_widget_attributes.cpp
gfgtdf: No i just commented it out to make the code easier for refactoring  i intended to check whether its still needed to bring it back in.  Also i'm not really a fan of unneeded invalidate_layout calls since they ,, well change the dialogs layout while its open which can be really confusing for the user.

I'd probably rather add a invalidate_layout function to the gui2 api. so that thare are less unconditional invaliadate_layout calls the the umc author cannot control.

src/scripting/lua_widget_attributes.cpp
gfgtdf: > Why is there a setter for this but no getter? That seems like a pretty enormous oversight. The unit preview pane isn't all that useful if you can't even get the unit displayed!

What? i think you are missing something here the unit previosu pane just shows a unit,  its a not a widget to select a unit.that is is has no user interaction, so there is a no point in such a getter.

src/scripting/lua_widget_attributes.cpp
gfgtdf: > Or is this a compatibility layer similar to value_compat? In

I _think_ it is but i'm not 100% sure

src/scripting/lua_widget_attributes.cpp
gfgtdf: On left_click is whenever  the user clicks on the widget (this works for widgets) while on button_click only works on widgets and depends on the specific type of button, in particular for "repeating_button" it will fire the event multiepl times if the user keeps pressing the button

src/scripting/lua_widget_attributes.cpp
gfgtdf: hm ye i guess i should fix them

src/scripting/lua_widget.cpp
gfgtdf: They are the implementation of the callbacks, basically they associate to each (c++) widget a lua table that contains their callbacl functions

src/scripting/lua_widget_attributes.cpp
gfgtdf: Hmm not really i t can be one with a bit i just wans't sure whwther its useful, and i prefer to rather not to add something in such cases.

src/scripting/lua_widget_attributes.cpp
gfgtdf: hmm yes i actually had the same thought, just an hour ago

src/scripting/lua_widget_attributes.cpp
gfgtdf: hmm bit isnt it the text on the label that is marked up and not the label ?

data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help.lua
CelticMinstrel: Ahh, I was wondering if that was the case. It's still not a great idea to be adding stuff to the widget metatable, but I guess it's okay since it's in a campaign rather than a multiplayer addon? Unless of course it's an MP campaign?

Maybe just assign `gui.widget.add_help_page = nil` in the postshow? At least then it won't leak out into any other addon, which wouldn't matter if this is single-player but does set a better example for anyone who _does_ want to do something similar in multiplayer. Calling it outside the scope of this dialog would never work anyway, right?

src/scripting/lua_widget.cpp
CelticMinstrel: Ah, okay... I think the functions are just poorly named, then...

src/scripting/lua_widget_attributes.cpp
CelticMinstrel: Hmm. Well, it's pretty easy to do it in Lua with the existing API anyway, so probably not that important. You'd just do `widget[widget.selected_index]`, right?

src/scripting/lua_widget_attributes.cpp
CelticMinstrel: I'm not sure, I was just responding to the content of the comment. It was just a random suggestion; if it doesn't make sense, don't use it?

I'd think of the label and the text of the label to be the same thing, personally. Not sure if that makes sense in GUI2.

src/scripting/lua_widget_attributes.cpp
CelticMinstrel: Adding a way to explicitly invalidate the layout in the GUI2 API seems alright to me. I'm not quite sure if I'd want to call it `invalidate_layout` but I can't think of anything better either, so I guess that'll do.

src/scripting/lua_widget_attributes.cpp
CelticMinstrel: I would argue that there's always a point in a getter. If you have a getter, you don't need to store the unit in a separate location. If you have a list box full of unit preview panes, you don't need to keep a separate list of the units. Keeping a separate list can always risk the two representations getting out of sync if you're not careful. I'd prefer to just get the unit directly from the preview pane that was selected in the list box.

Is there a technical reason that makes it difficult to implement a getter here? If there is, then it's not the end of the world to not have one, but if there's no technical reason that makes it hard, I think it should just exist.

The same goes for other read-only widgets. For example, the text on a button should have a getter. I didn't check whether all of these _do_ have getters, maybe they do already.

src/scripting/lua_widget_attributes.cpp
CelticMinstrel: So, I notice it's used in the zombie dialog. Is there a way to not do that? Or do you perhaps need to add more `on_xxx` callbacks? The zombie dialog seems to be doing an "on select" event, not sure what the actual underlying event is though.

src/scripting/lua_widget_attributes.cpp
CelticMinstrel: Hmm. I think I'd recommend renaming them to something like `on_mouse_down` (or `on_mouse_up`, whichever more accurately represents it) for the one that applies to any widget and `on_click` for the one that only applies to buttons.

src/scripting/lua_widget_attributes.cpp
gfgtdf: i think the on_modifed event is fired in this case.

src/scripting/lua_widget_attributes.cpp
gfgtdf: Well, then you'd need to store the 'unit' object somewhere inside the lua gui2 implementation which i still think is more work than its worth.

src/scripting/lua_widget.cpp
gfgtdf: well "luaW_pushwidgettable" pushed the table that is associated to a certain widget, inernally the data structure is 
{-- all windows  { --windowtable { --widget table} }} so the window table contains all widgettables for this window. and for each of these tables there are helper function that push them to the lua stack

src/scripting/lua_widget_attributes.cpp
CelticMinstrel: Does it needed to be stored in the Lua implementation? Does the C++ widget not provide a getter already? If there's no getter in C++, then you might be right that it's more work than it's worth. I'd still prefer to have it but I can live without it.

data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help_dialog.lua
gfgtdf: dunno, i like lua so personally wont take the time to do that.. , i don't think its dynamicially  created though.

data/campaigns/World_Conquest/lua/game_mechanics/wocopedia/help.lua
gfgtdf: > rather than a multiplayer addon? Unless

Its unrelated to single or multiplayer in particular since addons work in single and multi players.

> gui.widget.add_help_page = nil in the postshow?


well i'm not against doing this but i also don't see this as a problem.

data/campaigns/Secrets_of_the_Ancients/utils/zombie_recruit_dialog.lua
```diff
@@ -106,15 +107,16 @@ local function preshow()
...
+            list_item.unit_type.label   = afford_color_span_start .. unit_type.name .. " " .. z.sota_variation .. afford_color_span_end
+            list_item.unit_sprite.label = unit_type.image .. "~RC(magenta>red)"
+            list_item.unit_cost.label   = afford_color_span_start .. unit_type.cost .. afford_color_span_end
+
+            listedZombies[ dialog.unit_list.size ] = z.type
         end
-        zArrayIndex = zArrayIndex + 1
     end
-    wesnoth.set_dialog_value(1, "unit_list")
+    dialog.unit_list.selectzed_index = 1
```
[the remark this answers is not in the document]
gfgtdf: fixed in master

src/scripting/lua_widget_attributes.cpp
gfgtdf: Yes

src/scripting/lua_widget_attributes.cpp
gfgtdf: > I'd think of the label and the text of the label to be the same thing, personally. Not sure if that makes sense in GUI2.

Yes me too, im not sure whether we can though, if i undestood gui2 correctly widgets like  for example textboxes can have a label attached to them. And then these widgets do really have a  "label" and a "text"

src/scripting/lua_widget_attributes.cpp
gfgtdf: > 'm not quite sure if I'd want to call it invalidate_layout but

Yes i agree but i cannot think of a better name

src/scripting/lua_widget_attributes.cpp
gfgtdf: > Does the C++ widget not provide a getter already?

No, iirc the unit previw pane setter just sets all the child widgets to the information it extracts to the unit, meaning it does not  store the unit.

CelticMinstrel: I see. Well, I guess we can live without a getter for this attribute.

PR #5109, opened 2020-08-29
add-ons manager translated titles and description also for System default language

walodar:
The nice new feature that allows translated add-on titles and descriptions in the add-ons manager works well when the language preference is explicitly set to some supported language, but not when "System default language" is selected, because this is represented by an empty string in language.hpp. This pull request fixes this by getting the effective locale (the actual translation target language) from gettext.

### DIFF
--- a/src/addon/info.cpp
+++ b/src/addon/info.cpp
@@ -18,10 +18,11 @@
 #include "config.hpp"
 #include "font/pango/escape.hpp"
 #include "gettext.hpp"
-#include "language.hpp"
 #include "picture.hpp"
 #include "log.hpp"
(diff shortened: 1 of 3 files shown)


kabachuha:
If it will be merged with the master branch, someone also should backport the fix to 1.14, since the feature is in that version as well.

walodar:
Thank you. I have gladly backported these changes and opened the new additional pull request #5116 to the 1.14 branch.

gfgtdf (approved):

Pentarctagon:
I'll merge this and the backport tomorrow then, unless anything else comes up.

PR #5110, opened 2020-08-30
Help browser: fix behavior of clicking on icons of terrain sections

stevecotton:
This commit uses the UI's calculated indent for the UI's event handling, and
removes section::level. The code generating the help sections still has its own
"level", which is used to detect excessive recursion.

This fixes a UI quirk that I only noticed while working out what section::level
was for, while considering the subtopics needed for implementing a taxonomy
system. The help browser calculates indentation for its tree-view based on the
data tree given to it. However, the logic for deciding whether a mouse-click
was on the icon or the text used section::level, rather than any value
calculated in the UI classes. While section::level was often the correct
number, it was wrong for the terrain types' sections.


gfgtdf:
I don't know anything about that part all of the code so all i can say is that i see anything obviously wrong with it.

gfgtdf (approved):

PR #5123, opened 2020-09-03
Show the "loyal" trait in help

### LINKED ISSUE #1935: Trait loyal & aged bugs in Help
In the traits section in help:

loyal is listed as must_have
aged is missing

Aged used to be showing, but it's likely the changes to a generated trait list caused it to be dropped.
Took a look, couldn't figure out what was going on.

stevecotton:
Fixes #1935 for normal users, expanding on aac0941's fix which only showed
these abiities when Wesnoth was in debug mode.

The removed enum member "NON_REVEALING_DESCRIPTION" was unused, and all
existing callers effectively used description_type() as a boolean.

This does mean that the "aged" trait will be shown to all users, even if
they haven't played HttT, however that seems reasonable.

### DIFF
--- a/src/help/help_impl.cpp
+++ b/src/help/help_impl.cpp
@@ -680,7 +680,8 @@ std::vector<topic> generate_trait_topics(const bool sort_generated)
 	for (const unit_type_data::unit_type_map::value_type &i : unit_types.types())
 	{
 		const unit_type &type = i.second;
-		if (description_type(type) == FULL_DESCRIPTION) {
+		const auto desc_type = description_type(type);
+		if (desc_type == FULL_DESCRIPTION || desc_type == HIDDEN_BUT_SHOW_MACROS) {
 			if (config::const_child_itors traits = type.possible_traits()) {
 				for (const config & trait : traits) {
 					const std::string trait_id = trait["id"];
@@ -1015,6 +1016,12 @@ UNIT_DESCRIPTION_TYPE description_type(const unit_type &type)
 	if (encountered_units.find(type.id()) != encountered_units.end()) {
 		return FULL_DESCRIPTION;
 	}
+
+	// See the docs of HIDDEN_BUT_SHOW_MACROS
+	if (type.id() == "Fog Clearer") {
+		return HIDDEN_BUT_SHOW_MACROS;
+	}
+
 	return NO_DESCRIPTION;
 }
 
--- a/src/help/help_impl.hpp
+++ b/src/help/help_impl.hpp
@@ -229,7 +229,20 @@ void generate_races_sections(const config *help_cfg, section &sec, int level);
 void generate_terrain_sections(section &sec, int level);
 std::vector<topic> generate_unit_topics(const bool, const std::string& race);
 void generate_unit_sections(const config *help_cfg, section &sec, int level, const bool, const std::string& race);
-enum UNIT_DESCRIPTION_TYPE {FULL_DESCRIPTION, NO_DESCRIPTION, NON_REVEALING_DESCRIPTION};
+enum UNIT_DESCRIPTION_TYPE {
+	FULL_DESCRIPTION,
+	/** Ignore this unit for documentation purposes. */
+	NO_DESCRIPTION,
+	/**
+	 * Although the unit itself is hidden, traits reachable via this unit are not hidden.
+	 *
+	 * This is a bug workaround - traits are defined by WML macros, and therefore the help
+	 * system has to use a place where that macro is instanciated to provide the documentation.
+	 * None of the normal unit types has the "loyal" trait, but there is a hidden unit which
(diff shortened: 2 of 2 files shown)


gfgtdf:
You say in the comment that it's a workaroung fora  bug, so how would your ideal solution  would look like ?


Also you are hardcoding the fog Clearer id, im not sure whether that's a good thing to do, in particular because i'm quite sure that adding a trait (or maybe more likely: an ability) to a dummy unit to show it in the help  is something that has been done before by other umc authors.

stevecotton:
> something that has been done before by other umc authors.

That was already broken in 1.14. Note: if `Fog Clearer` is in your `encountered_units` then then loyal trait is visible, which caused some confusion during my testing.

> how would your ideal solution would look

* Add an alternative way to expand those macros in to the help, either by allowing `[trait]` inside a help `[topic]` or with a new WML tag. Sounds like more work to implement and more complexity.
* Add a new attribute to `[unit_type]`, which is only going to be used in the Fog Clearer. Sounds like that would add a little overhead to all the unit parsing.

In both cases, I think the downsides are small, but that they're worse than just hardcoding Fog Clearer into this place.

gfgtdf:
> That was already broken in 1.14. Note: if Fog Clearer is in your encountered_units then then loyal trait is visible, which caused some confusion during my testing.

Probably umc authors assumed it worked for the very same reason as your resting was confusing 

> or with a new WML tag

I realyl prefer this alternative, ideally these traits can the be also queried by lua.

stevecotton:
> > or with a new WML tag
> 
> I realyl prefer this alternative, ideally these traits can the be also queried by lua.

While I also like it, I think that's a feature request that's going to snowball into a huge feature request. I'd add refactoring the way that the trait macros work, so that the preprocessor doesn't create lots of copies of the help text, hopefully saving some startup time, saving some memory and making the elusive cache-corruption bug even rarer.

Anyone implementing that feature would delete the code added by this PR, but that seems a trivial addition to their workload.

soliton-:
I haven't looked too closely into the help but it seems to already generate WML internally. Maybe we can just pre-generate the help WML for the loyal trait and feed it to the help in addition to the internally generated WML?

stevecotton:
@soliton- , please would you give more details about that idea? How would it know which traits to treat specially, without hardcoding it somewhere?

soliton-:
Pre-generating the help WML, i.e. just putting it in some cfg file, is hard coding.

stevecotton:
What would that approach gain over the magic-monster-id approach that's already implemented in this PR?

If you were discussing the approaches for the big feature request then I've misunderstood, I'm assuming you're suggesting alternatives for the immediate bugfix.

soliton-:
I was responding to the new tag idea mentioning that you may not even need a new tag. The help already generates WML internally so you might just use that format. As I said though I haven't looked closely what exactly the generated help WML looks like.

stevecotton:
@soliton- would you be okay with me merging this PR as-is, and leaving anything more complex to the bigger feature request (which I'm not volunteering to work on)?

gfgtdf:
> I hadn't quite figured out how global traits would be specified… 

The simplest solution would be to have a ling global_traits=strong,quick,resitant,intelligent somewhere in the [game_config]

CelticMinstrel:
> The simplest solution would be to have a ling global_traits=strong,quick,resitant,intelligent somewhere in the [game_config]

I'm not sure if that solution works. Do addons need to be able to add global traits? If not, putting them in game_config is fine, but otherwise…

gfgtdf:
> Do addons need to be able to add global traits?


I'd say no, there is no usecase for this.

Tbh if i were to write to do it from scrath now i'd remove the idea of 'global traits' alltogether and just put all available traits in the [race]s that want them, but that'd be too much of a compatibility break at this point.

CelticMinstrel:
> Tbh if i were to write to do it from scrath now i'd remove the idea of 'global traits' alltogether and just put all available traits in the [race]s that want them, but that'd be too much of a compatibility break at this point.

I think it's actually doable - after all, `ignore_global_traits` exists, so the global traits can continue to work as normal (for backwards compatibility) but all mainline races now use `ignore_global_traits`.

> I'd say no, there is no usecase for this.

I can't think of a use-case but I suspect it's possible in 1.14. I wonder if there are any addons that do it though? If it's possible but no-one knew and no-one exploited it, then breaking it doesn't matter.

stevecotton:
While I welcome this discussion, please could it happen in the new feature request instead of this PR?

BTW, does anyone object to:
* Merging this PR as-is to get the functionality working in 1.15.5?
* Backporting it to 1.14?

CelticMinstrel:
I don't really like the fix. I think backporting it is fine at least, not sure about merging it.

PR #5127, opened 2020-09-04
Fix inverted logic in working out which resources are active

### LINKED ISSUE #5126: Add-ons loaded with [load_resource] are treated as inactive in 1.15.4
 - 1.15.4+dev (local build of a2b98687b7a88fecbf70a4630fbdc1f2872098ea)

### Describe the bug
The feature added in #4963 to only load active add-ons is buggy with respect to `[resource]` and `[load_resource]`. It looks like inverted logic on a conditional block that's meant to skip add-ons that have already been seen, instead it skips add-ons the first time that they're seen.

### To Reproduce
See https://github.com/ProditorMagnus/Ageless-for-1-14/issues/27

stevecotton:
This line is meant to skip the rest of the loop for things that have previously
been inserted in to loaded_resources; but instead it skipped the rest of the
loop unless the id was already in loaded_resources. Fixes #5126.

### DIFF
--- a/src/game_classification.cpp
+++ b/src/game_classification.cpp
@@ -142,7 +142,7 @@ std::set<std::string> game_classification::active_addons(const std::string& scen
 
 		const modevents_entry& current = mods.front();
 		if(current.type == "resource") {
-			if(loaded_resources.insert(current.id).second) {
+			if(!loaded_resources.insert(current.id).second) {
 				mods.pop_front();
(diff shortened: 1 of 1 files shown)


gfgtdf:
Hmm i thought i fixed this already, but doesn't look like it...

Yes looks good to me.

gfgtdf (approved):

PR #5130, opened 2020-09-04
Add 'Delay Advancements [modification]

gfgtdf:
Depsite being a [modification] this is actuall implemented
in the C++ code the [modification] just sets a flag.

With this modification active all advancements are delayed
until its the players (that controls the unit in question) turn
again.

This is meant as an alterntive to the preset advancement
type of mod.  want to play around with this alternative mod a a bit in 1.15
maybe we will remove it later if we find ot that it wasn't a good idea.

The problbme this is supposed to fix is that advancemnts
are done randomly during the enemeies turn. It has a few
advantages over the preset advancemnt approach:
1) It can easily handle Amlas and normal advancement
2) It doesn't need special code to handle the case that
   A unit advances multiple times. During an enemies turn.
3) It doesn't break in case that other wml code changed
   The advancements of a unit after the pre-selected advancement was chosen
4) It might be difficult for the user to know in advance which units
have a chance to advance during the next turn, if for example multiple of enemy units
unexpectedly attack one of the players units.
 

It also has a disadvantage: it changes the rules of the
game quite a bit, in particular if the units heals form
an advancement (which us usually the case), since now
with this 'healing the unit on advancement by retaliation'
during the enemies turn can no longer happen. I still
think its worth to think about this and test it though.

### DIFF
--- a/src/play_controller.cpp
+++ b/src/play_controller.cpp
@@ -534,7 +534,7 @@ void play_controller::do_init_side()
 		bool delay_advancements = std::find(active_mods.begin(), active_mods.end(), "delay_advancements") != active_mods.end();
 
 		for(unit &u : resources::gameboard->units()) {
-			if(u.side() == current_side()) {
+			if(delay_advancements && u.side() == current_side()) {
 				advance_unit_at(u.get_location());
 			}
 		}
--- a/src/play_controller.cpp
+++ b/src/play_controller.cpp
@@ -20,6 +20,7 @@
 
 #include "play_controller.hpp"
 
+#include "actions/advancement.hpp"
 #include "actions/create.hpp"
 #include "actions/heal.hpp"
 #include "actions/undo.hpp"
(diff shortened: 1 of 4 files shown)


src/play_controller.cpp
```diff
@@ -528,6 +529,16 @@ void play_controller::do_init_side()
 
 	// Make sure vision is accurate.
 	actions::clear_shroud(current_side(), true);
+	{
+		const auto& active_mods = get_saved_game().classification().active_mods;
+		bool delay_advancements = std::find(active_mods.begin(), active_mods.end(), "delay_advancements") != active_mods.end();
+
+		for(unit &u : resources::gameboard->units()) {
```
AI0867, line 536: Why is this loop run unconditionally, when only with delay_advancements active will any units be advances on start-of-turn?

gfgtdf: because that's 2 lines less i the c++ code for the same effect,  (and the performance implications are negletible)

AI0867: It's 2 lines less at the cost of complicating the conditional, and iterating over all units an extra time at the start of every side turn (regardless of how low that cost may be).

gfgtdf: well 'at the start of every side turn' is really nothing, keep in mind that every [store_unit] is slower, that we iterate over all units at the end of every user action to check whether the game ended, that we iterate over all units everytimes the player moves the mouse to calculate the resistances of enemies against the units attack in the right-side panel.

Pentarctagon:
> With this modification active all advancements are delayed until its the players (that controls the unit in question) turn **again**.

Question about the "again" specifically - does that mean if a unit gets the required xp to level up when it's currently their side's turn that the unit will need to wait a whole round to level up?

gfgtdf:
> Question about the "again" specifically - does that mean if a unit gets the required xp to level up when it's currently their side's turn that the unit will need to wait a whole round to level up?

No

Pentarctagon:
Alright.  Can this be merged prior to 1.15.6, so that it's available for players to try out?

PR #5131, opened 2020-09-04
misc wml cleanups

gfgtdf:
minor wml changes.

In particular we replace cases of multiple {MODIFY_UNIT} with one [modify_unit] call.

Also we add a few comments

### DIFF
--- a/data/campaigns/Delfadors_Memoirs/utils/sides.cfg
+++ b/data/campaigns/Delfadors_Memoirs/utils/sides.cfg
@@ -487,11 +487,11 @@
     {CLEAR_VARIABLE wesnoth_victors}
     # Kalenz will lead in next scenario
     {UNMAKE_HERO Kalenz}
-    {MODIFY_UNIT id=Kalenz canrecruit yes}
     [modify_unit]
         [filter]
             id=Kalenz
         [/filter]
+        canrecruit=yes
         x,y=recall,recall
     [/modify_unit]
 #enddef
--- a/data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
@@ -360,6 +360,7 @@
 
         {CLEAR_VARIABLE malin}
 
+        # FIXME: allow_undo like this causes corrupt replays.
         [allow_undo][/allow_undo]
     [/event]
 
@@ -393,6 +394,7 @@
 
         {CLEAR_VARIABLE malin}
 
+        # FIXME: allow_undo like this causes corrupt replays.
         [allow_undo][/allow_undo]
     [/event]
 
--- a/data/campaigns/Eastern_Invasion/scenarios/05_Northern_Outpost.cfg
+++ b/data/campaigns/Eastern_Invasion/scenarios/05_Northern_Outpost.cfg
@@ -254,7 +254,16 @@
 
         {DACYN_CASTING_SPELL_ANIMATION_OBJECT}
 
-        {MODIFY_UNIT id=Dacyn variables.casting_the_spell yes}
+        [modify_unit]
+            [filter]
+                id=Dacyn
+            [/filter]
+
+            [set_variable]
+                name=casting_the_spell
+                value=yes
+            [/set_variable]
+        [/modify_unit]
     [/event]
 
     # Dacyn remains busy until the bandits are defeated
@@ -351,7 +360,14 @@
 
         {VARIABLE shodrano_killed yes}
 
-        {MODIFY_UNIT id=Dacyn variables.casting_the_spell no}
+        [modify_unit]
+            [filter]
+                id=Dacyn
+            [/filter]
+            [clear_variable]
+                name=casting_the_spell
+            [/clear_variable]
+        [/modify_unit]
 
         [if]
             [have_unit]
@@ -404,7 +420,14 @@
     [event]
         name=victory
 
-        {MODIFY_UNIT id=Dacyn variables.casting_the_spell no}
+        [modify_unit]
+            [filter]
+                id=Dacyn
+            [/filter]
+            [clear_variable]
+                name=casting_the_spell
+            [/clear_variable]
+        [/modify_unit]
 
         [message]
             speaker=Owaec
--- a/data/campaigns/The_South_Guard/scenarios/05_Choice_In_The_Fog.cfg
+++ b/data/campaigns/The_South_Guard/scenarios/05_Choice_In_The_Fog.cfg
@@ -864,6 +864,7 @@
 
                             [show_objectives][/show_objectives]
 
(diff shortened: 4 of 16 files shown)


stevecotton:
The up-to-date documentation for running the WML validation locally is the first section https://wiki.wesnoth.org/ValidationFAQ , however there's also plenty of out-of-date "validation" pages on the Wiki too (including the later sections of that very page). Please would you try it and give feedback about what needs improvement there?

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
```diff
@@ -360,6 +360,7 @@
 
         {CLEAR_VARIABLE malin}
 
+        # FIXME: allow_undo like this causes corrupt replays.
         [allow_undo][/allow_undo]
```
soliton-, line 364: Why not just remove these since they're clearly bogus?

gfgtdf: I'm not sure what the correct approach is here:
1) remove the [allow_undo]
2) remove the xp gain
3) make it somehow work with [on_undo]

soliton-: Ah, [on_undo] could be an option, yeah.

stevecotton: For this one, I think removing the [allow_undo] is the best choice. The hex it's on is en-route from the starting castle to a village that's perfectly placed for the leader to capture on the first turn, so I'd expect the player to pick this up and then immediately grab the village and end turn.

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
```diff
@@ -393,6 +394,7 @@
 
         {CLEAR_VARIABLE malin}
 
+        # FIXME: allow_undo like this causes corrupt replays.
```
stevecotton, line 397: Removing the [allow_undo] on this feels right to me, but it doesn't have the definite reason that the other one does.

@nemaara as this gives a significant amount of XP, I think there should be at least some hint if the player sends a walking corpse to investigate it.

nemaara: All of the [allow-undo]s can just be removed I think.

data/campaigns/Eastern_Invasion/scenarios/05_Northern_Outpost.cfg
```diff
@@ -254,7 +254,16 @@
...
+            [filter]
+                id=Dacyn
+            [/filter]
+
+            [set_variable]
+                name=casting_the_spell
+                value=yes
+            [/set_variable]
+        [/modify_unit]
     [/event]
```
stevecotton, line 267: Unless we're going to deprecate the macro, I think the three changes in this file shouldn't be done - leaving them as one-liners is easier to read.

data/campaigns/The_South_Guard/scenarios/05_Choice_In_The_Fog.cfg
```diff
@@ -864,6 +864,7 @@
 
                             [show_objectives][/show_objectives]
 
+                            # TODO: what is the effect of these lines? CLEAR_VARIABLE cannot undo an [effect]...
```
stevecotton, line 867: I guess it rebuilds the unit, thus maybe working around the broken use of TRAIT_LOYAL inside a `[modify_unit][modifications]` block.

I think #5634 fixes it properly, and that PR also takes your improvement for TSG S02 from the current PR.

Pentarctagon:
Is this still something that should be kept open?

CelticMinstrel:
With a little cleanup and _maybe_ a deprecation message for `{MODIFY_UNIT}`, I think this could probably be merged into master…

data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
```diff
@@ -840,26 +840,17 @@
...
-                [unstore_unit]
-                    variable=this_item
-                [/unstore_unit]
-            [/do]
-        [/foreach]
-        {CLEAR_VARIABLE bandits}
+            [status]
+                guardian=no
+            [/status]
+        [/modify_unit]
```
stevecotton, line 853: This one is still worth doing, but needs testing. The old code changes both `ai_special` and `status.guardian`, why does the new code only change `status.guardian`?

data/campaigns/Delfadors_Memoirs/utils/sides.cfg
```diff
@@ -487,11 +487,11 @@
     {CLEAR_VARIABLE wesnoth_victors}
     # Kalenz will lead in next scenario
     {UNMAKE_HERO Kalenz}
-    {MODIFY_UNIT id=Kalenz canrecruit yes}
     [modify_unit]
         [filter]
             id=Kalenz
         [/filter]
+        canrecruit=yes
```
stevecotton, line 494: Cherry-picking this one seems good.

stevecotton (commented):
I think this should be closed. I already cherry-picked some of the changes, and there's a couple more than I think are improvements (commented on in this review) which I'd be happy if someone cherry-picked. The rest seem like stuff that shouldn't be done - expanding a one-liner to a larger block of text slightly reduces the readability of the code.

data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
CelticMinstrel: `ai_special=guardian` is a very old feature that was replaced with the `guardian` status and does nothing but set it to yes. Setting `ai_special` to anything other than `guardian` has no effect.

CelticMinstrel:
Expanding a one-liner to a block _may_ affect readability, so if there are some like that, sure. I did see some where a group of `{MODIFY_UNIT}` statements are replaced with a single `[modify_unit]` block though, which is definitely an improvement; and there was even one place where a loop was replaced by a single `[modify_unit]` block.

stevecotton:
> Expanding a one-liner to a block _may_ affect readability, so if there are some like that, sure. I did see some where a group of `{MODIFY_UNIT}` statements are replaced with a single `[modify_unit]` block though, which is definitely an improvement;

Even 2 or 3 `MODIFY_UNIT` lines can be more readable than a unified `[modify_unit]` block. Less efficient, but I don't see a place where this is performance-critical code.

> and there was even one place where a loop was replaced by a single `[modify_unit]` block.

If you're referring to TSG S02 then the change should be to fix #5635 instead.

CelticMinstrel:
That's true, then neither a loop nor a modify_unit would be needed. I don't think I really agree on your first point though.

Issue #5122, opened 2020-09-03
Compilation of mushroom terrain related graphical issues.


Hejnewar:
Issue 1 - For most people new mushroom terrain simply looks worse than the previous one. I strongly encourage to look at them in game on MP maps and in campaigns.
Before:
![Before](https://user-images.githubusercontent.com/41817546/92088416-213dd380-edcd-11ea-8ed8-a8fc1e734d52.JPG)
After:
![After](https://user-images.githubusercontent.com/41817546/92088420-21d66a00-edcd-11ea-8852-a3f48e03321c.JPG)
It also starts to cover swamp terrain and this looks awful (I see potential in this one high grass terrain / ornament could look good):
![This](https://user-images.githubusercontent.com/41817546/92091844-9a3f2a00-edd1-11ea-85a7-efba8d6df49c.JPG)

Issue 2 - New mushroom terrain doesn't cover hill terrain that exists underneath it perfectly.
![Cover](https://user-images.githubusercontent.com/41817546/92089545-978f0580-edce-11ea-91a6-22473df8dde0.JPG) ![MushroomCover](https://user-images.githubusercontent.com/41817546/92088477-387cc100-edcd-11ea-9657-747efaf5bebe.JPG) 

Issue 3 -Transitions work weirdly. They sometimes apply and sometimes they don't.
![Shrooms in watter](https://user-images.githubusercontent.com/41817546/92088616-6235e800-edcd-11ea-8349-cd1df8336ec8.JPG) ![In viodd](https://user-images.githubusercontent.com/41817546/92090528-d3769a80-edcf-11ea-968b-e616a62cd65f.JPG) ![In water 3](https://user-images.githubusercontent.com/41817546/92089689-c7d6a400-edce-11ea-84ff-b1eb4fc3381f.JPG)   
In water but in line:
![In line](https://user-images.githubusercontent.com/41817546/92089007-eb4d1f00-edcd-11ea-8de5-465187c2be8f.JPG)

Issue 4 - This is also a problem in 1.14 but transitions don't work well with few other terrains.
With walls and with keeps (dune keep for reference):
![Keep transitions also 1 14](https://user-images.githubusercontent.com/41817546/92088883-c35dbb80-edcd-11ea-9fba-80210eab21fb.JPG) ![Shrooms and wall](https://user-images.githubusercontent.com/41817546/92088850-b9d45380-edcd-11ea-8ff1-1774c9cf4727.JPG)

[...]

PR #5136, opened 2020-09-07
make mushrooms-small smaller

doofus-01:
To fix the spillover shown in #5122 

@Hejnewar - I am not able to reproduce your third image in that issue.  Can you post that again with the terrain code overlay?


Hejnewar:
Sure. 
Map: Swamp of Dread. 
Map

I believe @Pentarctagon is already working on cases like this one in #5137

doofus-01:
OK, so it's not an issue with ^Tf or Tb.  Thanks.

Issue #5135, opened 2020-09-06
wc:  deprecated fungus terrain


gfgtdf:
wc still uses the old fungus terrain

Pentarctagon:
`grep -rl "\^Uf" | wc -l` gives 157 files as still using the old terrain, so it's looks like most of mainline still hasn't been updated to the new terrain, rather than just WC.

stevecotton:
I think this transition is a low priority - even if all of mainline is updated, 1.16 and probably 1.18 have to keep support for `^Uf` for UMC.

Pentarctagon:
I don't think there's really any reason/need to remove support for ^Uf entirely, but if something is deprecated then mainline shouldn't be using it.

PR #5137, opened 2020-09-07
[WIP] Mass convert of *^Uf terrain to the balance equivalent Tb^Tf terrain.

Pentarctagon:
---

This is a mass-conversion of all remaining *^Uf terrain to Tb^Tf, since if we can't make sure that something that's deprecated isn't used by mainline anymore then it shouldn't have been deprecated in the first place.  It does not touch WC, since I assume @gfgtdf has his own plans to handle WC (#5135).  Marked WIP since I also noticed along the way that the Xol and Xuce terrains are deprecated but also still used in a few places, so I intend to review all the deprecated terrains and make sure they aren't used in any mainline maps anymore.

@gfgtdf I've requested your review as well to sanity check the changes to `data/multiplayer/scenarios/test_lua_mapgen.lua` in particular.


stevecotton:
While it will be a lot of effort, I'd prefer this to be done with a manual review of each map, rather than an bulk search-and-replace. Some of the ^Uf tiles should be Tb^Tf, but others would be better as Uu^Tf or Hh^Tf. Leaving the existing ^Uf tiles as-is both tells us which maps haven't been reviewed yet, and during a map review helps track which hexes have already been considered.

The evil path of Northern Rebirth should probably be removed, rather than being part of this.

stevecotton:
For Xuce, the only instances are some leftovers (that aren't visible to the player) in HttT's 14_Plunging_Into_the_Darkness.map . That also has ^Uf mushrooms, but only in a secret area.

data/campaigns/Under_the_Burning_Suns/utils/animation_macros.cfg
```diff
@@ -22,7 +22,5 @@ misc/blank-hex.png~BLIT(units/quenoth/fire/flame-color.png,{X},{Y2})~MASK(units/
 
 # Not all unwalkable, but still supposed to provide some visual cue that movement is different.
 # This is only used for animation, it's the list of terrains on which the Sun Sylph will fly.
-# Both the old ^Uf* and new ^Tf* mushrooms are supported here - if the Quenoth elves are moved
-# to core then it seems sensible to still have the old behavior on old maps.
```
stevecotton, line 26: I think this should stay in, it applies to old UMC maps even when there are no ^Ufs in mainline.

gfgtdf:
data/multiplayer/scenarios/test_lua_mapgen.lua is basciaiyll the WC2 maritime mapgenerator in a one-file version, in fact i consider removing it now that WC2 is mainlined.

data/campaigns/Under_the_Burning_Suns/utils/animation_macros.cfg
Pentarctagon: Alright, I'll push an update.

Pentarctagon:
@gfgtdf is there anything that still uses it, or can I delete it here?

gfgtdf:
> @gfgtdf is there anything that still uses it, or can I delete it here?

No it's a test as the name says, it was never used by any other code.

gfgtdf:
> While it will be a lot of effort, I'd prefer this to be done with a manual review of each map

I agree on this part, in particular since i think on a lot of these  of these maps the balance effect of these is negligible,  So it'd probably be better to use the old graphocs that the scenario designer had in mind in the first place.

Pentarctagon:
I'm fine with someone manually reviewing the maps as long as it's done (read: all *^Uf is removed from mainline maps) prior to 1.16's release.  Which would by default mean @nemaara for SP and then probably @Hejnewar et al for MP.

Hejnewar:
Yeah, I think it's better to do this this way, . It's gonna take some work and time but better looks (hopefully) will be worth it.

soliton-:
I think in campaigns the chance is low that the mixed mushroom terrain would break something. It's likely used primarily for the aesthetics.

For the balanced multiplayer maps on the other hand it likely would since it is often used precisely for its gameplay properties that the mixed terrain does not have.

Pentarctagon:
@nemaara then - will you be able to do the review of *^Uf terrain usage in SP maps?

nemaara:
I'm confused, does this PR leave the tiles looking the same as before, or do they look different? If it's the same look then SP doesn't need a manual review and a mass change is fine.

Pentarctagon:
The look is the same as it was previously in 1.15, but different from 1.14.  For example, the Slavers map from DW-

1.14 (Steam):
DW-Slavers-1 14-Steam
1.15 (Steam):
DW-Slavers-1 15-Steam
1.15 (With PR applied):
DW-Slavers-1 15-With-PR

soliton-:
Mass conversion for SP should use the mixed terrain that keeps the aesthetics not the gameplay properties.

Pentarctagon:
Ultimately which way to handle it is up to nemaara.  The only restriction I'm placing on it though is that, one way or another, the deprecated mushroom terrain won't be used by mainline maps in 1.16.

nemaara:
I agree with Soliton. Let's keep the mushrooms with the same look as 1.14.

Pentarctagon:
Alright, will update.

Pentarctagon:
@nemaara I've changed the SP maps to use the overlay instead.

@gfgtdf I went ahead and deleted the test_lua_mapgen.lua and test_lua_mapgen.cfg files.

@Hejnewar I've opened #5145 for the MP maps that still use `^Uf`.

data/multiplayer/scenarios/2p_Elensefar_Courtyard.cfg
```diff
@@ -96,7 +96,7 @@
             [/or]
             [or]
                 x,y=32-36,9-19
-                terrain=R*,Gg^Emf,Rp^Uf,Rp^Dr,Rd^Uf
+                terrain=R*,Gg^Emf,Tb^Tf,Rp^Dr
```
stevecotton, line 99: This needs to be in #5145 , along with the change to the corresponding map.

data/multiplayer/scenarios/ANL_utils/ANL_worker_options.cfg
```diff
@@ -473,7 +473,7 @@
                             image = "terrain/cave/floor.png"
                             label = "<span color='green'>" + _ "Harvest Mushrooms" + "</span>"
                             description = "<span size='small'>" + _ "Earns: 3g" + "</span>"
-                            {ANL_SHOW_IF ("Peasant") "Uu^Uf"}
+                            {ANL_SHOW_IF ("Peasant") "Tb^Tf"}
```
stevecotton, line 476: I think this needs to be Uu^Tf. @sevu, please would you confirm?

The corresponding map change isn't in this commit, it was already done in 353dce142ebc55ff954e7de66171b80fb51bfffa. Looks like the effect is that peasants can't harvest mushrooms in 1.15.4.

gfgtdf:
> @gfgtdf I went ahead and deleted the test_lua_mapgen.lua and test_lua_mapgen.cfg files.

I think it'd be better to to do in a seperate  commit though deletinga  a file in a commit named "Mass convert of *^Uf terrain for SP." make it looks like  you accidentally deleted it.

data/multiplayer/scenarios/2p_Elensefar_Courtyard.cfg
Pentarctagon: Done.

Pentarctagon:
Moved the file deletions to a separate commit.

gfgtdf:
The commit message now kinda imples that there used at some point or at least supposed to be used. To make it clear, they were tests, like the unit [test]s for specific features, they were never intended to be used by any other code.

Pentarctagon:
"unused" doesn't have any implication of what they were used for previously.  It means just that - nobody uses them for anything anymore.

data/multiplayer/scenarios/ANL_utils/ANL_worker_options.cfg
sevu: Yes. Easy fix is changing it to `Uu^Tf`. Better fix is using `*^Tf` and using 29 lines farther
[terrain]
x,y=$x1,$y1
terrain="^"
layer=overlay
[/terrain]
@Pentarctagon can change it or remove it from the commit

Pentarctagon: @sevu You mean changing `{MODIFY_TERRAIN "Uu" $x1 $y1}` to
```
[terrain]
x,y=$x1,$y1
terrain="^"
layer=overlay
[/terrain]
```
?

Pentarctagon:
@nemaara Can you approve, if there's nothing else?

data/multiplayer/scenarios/ANL_utils/ANL_worker_options.cfg
sevu: yes

gfgtdf:
He said if it doesn't change that visuals but it does. That fact that the visuals were already broken in some earlier 1.15.x version is not really important her in my opinion. What matter i that they look different than when they were drawn/written.

Pentarctagon:
I don't follow.

Pentarctagon:
Soliton said that for SP the mushrooms should keep the same look as in 1.14, nemaara agreed, and so I implemented that.

data/multiplayer/scenarios/ANL_utils/ANL_worker_options.cfg
Pentarctagon: Alright, updated.

gfgtdf:
oh okay my bad then, i just looked at the first file  which happend to be the ai test sceanrio which still converted "Uu^Uf" to "Tb^Tf" so i assumed this pr still changed everything to Tb^Tb

Pentarctagon:
Yeah, stuff like that (the pathfinding unit test as well for example) I kept coverted to Tb^Tf since aesthetics definitely aren't relevant there.

nemaara (approved):

PR #5138, opened 2020-09-07
WIP, completely untested: Remove the evil path of Northern Rebirth

stevecotton:
@Pentarctagon , @gfgtdf This isn't ready for merging (because it's untested), but it (slightly) simplifies your #5137 and #5131 .

Its inclusion in mainline is causing extra work in tasks roadmapped for 1.15.x.
It has some of the deprecated ^Uf mushrooms, and some use of the {MODIFY_UNIT}
macro.

If it comes back then it will likely start as UMC.


nemaara (approved):
As long as the campaign doesn't crash by removing these scenarios it seems fine to me.

stevecotton:
Thanks for the feedback, will do some testing and (hopefully) merge afterwards.

cooljeanius:
> 
> If it comes back then it will likely start as UMC.

There's already Northern Rebirth Remake on the add-ons server that has the evil branch

stevecotton:
Thanks @cooljeanius , I've added that to the draft release notes. https://wiki.wesnoth.org/Release_Notes_1.15.5

Issue #4876, opened 2020-04-26
In UtBS, Wesnothpedia says Elves have pale skin


eyalroz:
Wesnoth 1.15.3
Any scenario

The Wesnothpedia section on Elves discusses the regular elves, despite the individual unit links pointing to the Quenoth variants. The general text for elves should be altered as well - or there should be a split between Elves and Quenoth Elves (even better)

Wedge009:
Is 'Wesnothpedia' the in-game help? I wonder if there's a mix-up in how the pages are generated - if you temporarily rename your preferences file (a lot of global state information, such as units you've encountered, is also stored here) so you are effectively starting with a 'blank slate', does the issue still occur?

eyalroz:
> Is 'Wesnothpedia' the in-game help? 

Yes.

> if you temporarily rename your preferences file

Which preference file, exactly?

Wedge009:
_The_ preferences file: https://wiki.wesnoth.org/Editingwesnoth

eyalroz:
@Wedge009 : I my userdata direction (`wesnoth --config-path 2>&1 | tail -1`) there is no preferences file as described at the link.

soliton-:
It’s --userconfig-path nowadays.

stevecotton:
There's no `[race]` tag anywhere in data/campaigns/UtBS. The Quenoth units all have `race=elf`, which gets the `[race]id=elf` from data/core/units.cfg; and a `[hide_help]` is used to hide the pre-Fall elves.

I agree with the suggestion of making the Quenoth elves a separate race. This isn't an offer to do the work, however I am intending to play the campaign again soon with #4974, so might test.

eyalroz:
> It’s --userconfig-path nowadays.

No preferences file there either (under `$HOME/local/share/wesnoth/1.15`).

soliton-:
Sounds like you are using --userconfig-dir or --config-dir with a different path when you start wesnoth or your installation is broken.

Btw, it's odd that the path you report says `local` and not `.local`. Haven't seen that before.

eyalroz:
> Sounds like you are using --userconfig-dir or --config-dir with a different path when you start wesnoth or your installation is broken.

I'm not, and - it isn't; at least, not that I know of...

Also - it's `.local`, I mis-copied.

eyalroz:
[...]

PR #5143, opened 2020-09-09
Show [race]help_taxonomy= as "This is a group of units, all of whom are $help_taxonomy"

stevecotton:
This allows the help browser to add links between [race]s. I've used the word "group" rather
than "subrace",  it seems less likely to have political overtones.

The code comments talk about both Dunefolk and Quenoth Elves - of these two, only the
Dunefolk's data is changed in this commit, the Quenoth still have race=elf at the moment.

Draft documentation for the Wiki:

```
* '''help_taxonomy''': {{DevFeature1.15|?}} in the help browser, show this
race as a group of units from another race; the value of this attribute should
be the other race's '''id''' attribute. This only affects the help browser, for
all other purposes (such as WML filters) the two races are completely separate.

How this is visualised in the help browser is a GUI design decision, this attribute
merely tells the engine that the relationship exists.
```

The PR provides a visualisation, even though the proposed Wiki-doc is non-specific.
Here's three vertically-sliced screenshots:
wesnoth_help_taxonomy

### DIFF
--- a/data/core/units.cfg
+++ b/data/core/units.cfg
@@ -209,6 +209,7 @@ The eastern provinces of Wesnoth, known as the Clan Homelands, have a geography 
         male_name= _"race^Dunefolk Human"
         female_name= _"race+female^Dunefolk Human"
         plural_name= _"race+plural^Dunefolk"
+        help_taxonomy=human
         description= _"An offshoot of a forgotten nomadic civilization, the Dunefolk lay claim to the river valleys and oases of the Sandy Wastes. How they came to inhabit this far corner of the Great Continent is unknown. Their legends tell of many long and perilous travels through far-flung lands, but the true origin of their people is a topic of endless and heated debate among even the most erudite of their scholars.
 
 Whatever their origin, the Dunefolk have thrived. Bustling cities stand proudly in the largest fertile regions. Skilled artisans, fine smiths, and wealthy merchants form the backbone of the urban economy. Each of these cities also enjoys a degree of independence less common in the more centralized nations to the north, many even maintaining their own standing armies. In times of need, however, each and all rally to a higher authority designated to protect the superior interest of the nation.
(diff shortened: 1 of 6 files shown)


CelticMinstrel:
1. I don't think the draft wiki documentation belongs in the commit message?
2. Shouldn't this also be added to the Quenoth race definition? Or is that a separate PR since it's a campaign?

stevecotton:
The build failure is a missing optimisation combined with Werror; I'll fix it in the next upload.

> I don't think the draft wiki documentation belongs in the commit message?

Why not? It's also a description of the functionality added by the commit.

> Shouldn't this also be added to the Quenoth race definition? Or is that a separate PR since it's a campaign?

Separate PR because, although I've been talking about the Quenoth as a separate race, they're currently using `race=elf`. The details of that are in the WIP commit that's getting automatically linked to issue #4876.

CelticMinstrel:
> Why not? It's also a description of the functionality added by the commit.

Sure, but it's not formatted in a way that's sensible for a commit message. And it includes MediaWiki markup which just seems silly.

stevecotton:
> Sure, but it's not formatted in a way that's sensible for a commit message. And it includes MediaWiki markup which just seems silly.

Please would you give an example of how to write it better? Knowledge of MediaWiki should be common for Wesnoth devs - anyone adding new WML attributes should be updating the Wiki themselves.

CelticMinstrel:
I don't even think you need any of that in the commit message, honestly. Everything above the "Draft documentation for wiki" is basically sufficient for a commit message. Maybe it could be a touch more detailed, like "Add [race]help_taxonomy= to group races in the help browser" or something like that. The draft wiki documentation can go on the PR but it doesn't fit with what a commit message _is_.

I mean, it's not a deal-breaker or something if you really want to put it in the commit message. It just doesn't make any sense to me to see it there.

stevecotton:
Plans for merging: I'm intending to leave the PR open until Sunday, and then merge unless an issue is found.

About the commit message, part of it is communication with myself. I tried different UX ideas, cherry-picked and rebased this along with the [race]id=quenoth changes. So the commit message is partially to tell myself what this version of the code is meant to do.

stevecotton:
Still has a final todo for me - working out how to phrase this for the release-notes.

PR #5146, opened 2020-09-11
Allow the player to recruit, even when some recruitable units aren't known

### LINKED ISSUE #5144: One missing unit in recruit list makes entire recruit list unavailable
 - 7036af787f9e4e871a0e5a1647670fa09535ea47 and later

### Describe the bug
When recruit list contains id which is not known unit, it is not possible to recruit other units either.

### To Reproduce
https://forums.wesnoth.org/viewtopic.php?p=658078#p658078

### Expected behavior
Being able to recruit the units which are known.

stevecotton:
Fixes #5144 "one missing unit in recruit list makes entire recruit list
unavailable". The player will see an error dialog every time they try to
recruit, that seems good given that the limited recruitment options are likely
to severely affect game balance.

The `std::vector<t_string> unknown_units` uses t_string instead of std::string
because that's the class needed for format_conjunct_list. Those items aren't
translated - they're the values of [unit_type]id=, but the corresponding
[unit_type] is missing, so we can't get the translatable [unit_type]name=.

### DIFF
--- a/src/menu_events.cpp
+++ b/src/menu_events.cpp
@@ -259,16 +259,29 @@ void menu_handler::recruit(int side_num, const map_location& last_hex)
 
 	std::set<std::string> recruits = actions::get_recruits(side_num, last_hex);
 
+	std::vector<t_string> unknown_units;
 	for(const auto& recruit : recruits) {
 		const unit_type* type = unit_types.find(recruit);
 		if(!type) {
 			ERR_NG << "could not find unit '" << recruit << "'" << std::endl;
-			return;
+			unknown_units.emplace_back(recruit);
+			continue;
 		}
 
 		map_location ignored;
 		map_location recruit_hex = last_hex;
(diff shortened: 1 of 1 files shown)


gfgtdf:
This shows the error message overytime the user tries to recruit, i think it'd be better if it was only shown once.

stevecotton:
> it'd be better if it was only shown once

Why? Please give a reason rather than just an opinion; I've already put my reasoning in the commit message.

gfgtdf:
> Why? Please give a reason rather than just an opinion; I've already put my reasoning in the commit message.

Okay i just looked at the code.

Because you are not telling the player anything  new and it will probably be annoying at some point, also this error is not worse than any other type of error quite the opposite: while this type of error has a good chance that the scenario is still playable and just a bit more difficult, while for example erros in a scenarios wml code , have a good chance to make a scenario technically unwinnable.


That said it's not like i care that much about it, an error is an error, and if we need a annoying repeating  message to make the  user report a bug in the campaigns thread about it, then i'm fine with it.

PR #5150, opened 2020-09-13
Item images contributed by Blarumyrran on Discord

doofus-01:
I trust Blarumyrran knows what he's doing, but I still want to see these in game before merging this.
I also don't want it to be forgotten, so I'll add the 1.16 milestone.


CelticMinstrel:
For staff.png, the old and new are a very different design, and I like both of them. Maybe Blarumyrran's staff could become staff2.png or something?

The same is also true of staff-magic.png, but I don't like the old one as much in that case.

doofus-01:
opened a newer version #5821

Issue #5074, opened 2020-08-10
UtBS dialogue touchup


Pentarctagon:
It's not certain whether this will be able to be finished in time for 1.15/1.16, but this is the issue to keep track of it.

Taken from the [1.15 Roadmap](https://wiki.wesnoth.org/1.15_Roadmap#1.15.14_.2806.2F19.2F2021.29_.28Beta_1.29).

Pentarctagon:
@nemaara Should I close this or move it to the 1.17 milestone?

nemaara:
I'd say close, supposing I remain SP maintainer I think tackling DM/EI first is more pertinent and I'll try my hand at it before UtBS.

Pentarctagon:
Alright, closing then.

PR #5151, opened 2020-09-15
UtBS: Make "quenoth" a separate race to "elf" (fixes #4876)

### LINKED ISSUE #4876: In UtBS, Wesnothpedia says Elves have pale skin

stevecotton:
# Main commit

UtBS: Make "quenoth" a separate race to "elf" (fixes #4876)

This is motivated by reorganising the help pages to split the Quenoth from the
"pointy ears, pale skin" help text in data/core/units.cfg. It also means that
there's no reason to [hide_help] for the default era elvish units. The new help
text is just a short placeholder for now.

The new race has [race]help_taxonomy=elf.

If the player loads a mid-campaign save then their recalls and heroes will have
race=elf while new recruits have race=quenoth.  For that reason, I've changed
the scenarios' event triggers to recognise both races.

When levelling up, units with race=elf may change to race=quenoth.

Opening the help for a unit will show the help as if the unit has race=quenoth,
even when right-clicking on a race=elf unit. This is good.

For a few of these, I think another race might be better - they're converted to
Quenoth here, as they were Elves before. The Corrupted Elf is should probably be
considered undead, and the Divine Avatar should possibly be something else, more
like an elemental than a living creature. That those two are currently included
in the elves does have the side-effect of hiding a plot spoiler about the Dark
Assassin who is also included in the elves (the spoiler being that this one is
correct, he really is an elf).

# Second commit

UtBS S02: Make the warning about the tentacles an enter_hex instead of a moveto

If you try to move straight into the water, the explorer should stop at the
lake edge and comment about where they're going. Trivial change, mainly because
I was using a fast unit for debugging and wondered why I hadn't triggered the
lake monsters.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/_main.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/_main.cfg
@@ -182,7 +182,7 @@
         first_time_only=no
 
         [filter]
-            race=elf
+            race=quenoth,elf
         [/filter]
 
         [filter_condition]
@@ -213,7 +213,7 @@
 [/campaign]
 # wmllint: validate-on
 
-#Append Units
+#Append Units, including the movetypes and race definition
 [units]
     {UTBS_INCLUDE units/units.cfg}
     {UTBS_INCLUDE units/quenoth}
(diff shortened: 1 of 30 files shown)


stevecotton:
Going to force-push to fix a commit message, with the same code as before. The second commit should say "UtBS S06b" instead of "UtBS S02".

nemaara:
I'm assuming this works as in it doesn't break the game coz of any code changes or anything.

About the race text, I kind of find it a bit meh. Maybe we could try having something a bit mysterious? It doesn't have to be long or anything, just less direct comparison with elves.

stevecotton:
I've given it a test play-through, and think there's no bugs. Admittedly I was using debug mode from scenario 3 onwards, giving characters unlimited movement points and using debug-kill to run around the map instead of fighting.

> race text is a bit meh

While I fully agree with you, I'd prefer to leave that for another PR, and discussion in #4876 seemed okay with that plan. (Okay, full disclosure is that I'd like to drop the task on you in #5074).

PR #5154, opened 2020-09-16
Change most remaining cases of using async_send_doc to send_to_player for wesnothd

Pentarctagon:


soliton-:
I would rename `send_to_player` to `async_send_doc_queued` or so to better reflect what it does.

It might also better fit into `send_receive_wml_helpers.ipp` where `async_send_doc` and the other `document` handling is. Maybe @loonycyborg has an opinion on that. I don't really know why it's devided like that.

loonycyborg:
Only reason it's divided like that is because send_to_player existed before server_base was factored out. And I saw write queue as specific to wesnothd, in campaignd it does only one send at the time anyway since pattern there is request->response->request->response.. In wesnothd afaict  most of potential for interleaved writes comes from PMs between players and any failures to queue might be hard to detect since messages will be mostly quite small and won't be split into multiple low level writes by async_write(). I'm fine with promoting this to lower-level utilities though.

loonycyborg (approved):

loonycyborg:
Also this PR didn't appear among my review requested notifications in github for some reason.

Pentarctagon:
send_to_player() is also used in wesnothd's game.cpp, so I think leaving it in server_base makes sense.

Pentarctagon:
@soliton- I've updated the PR to rename the function.

soliton- (approved):

PR #5157, opened 2020-09-19
A minor Polish translation fix

PiotrMakarewicz:
I would normally have sent it to ivanovic but it's so close to the release of 1.15.5 that he may not be in time to commit it.


Pentarctagon:
I also didn't see this prior to 1.15.5 being tagged, so it's a moot point now.

Issue #5158, opened 2020-09-19
Improving unit placed events


sevu:
Some tags fire as side-effect `unit placed` events.
That's due to implementation details, might be good to change that.

It's know to happen for:
- [x] move_unit, probably move_unit_fake too
- [x] unpetrify

are there more?

It's either due to the the Lua statement `u:extract()` or `u:to_map()`. Exists there an alternative which can be used?

<s>Edit: … could be as easy as `u:to_map(nil, nil, false)`</s>

Is as easy as `u:to_map(u.x, u.y, false)`

CelticMinstrel:
I'm not 100% sure but I think calling `to_map()` on a unit that's already on the map might also not fire an event? So if the `extract()` isn't strictly necessary, removing it _might_ make a difference.

sevu:
In the case of wesnoth.interface.move_unit_fake:
Without the extract the unit sprite remains until the unit reached it's goal and is removed then. But still fires an unit placed event.

For [unpetrify], without the extract() it causes a coredump, guess that's a bug by itself? The backtrace is with 1.15.5.

For other places I wonder why it is needed.

```
#0  std::__uniq_ptr_impl<unit_animation_component, std::default_delete<unit_animation_component> >::_M_ptr() const (this=0x548) at /usr/include/c++/10.2.0/bits/unique_ptr.h:173
#1  0x000056011ef77b8a in std::unique_ptr<unit_animation_component, std::default_delete<unit_animation_component> >::get() const (this=0x548) at /usr/include/c++/10.2.0/bits/unique_ptr.h:422
#2  0x000056011ef730ec in std::unique_ptr<unit_animation_component, std::default_delete<unit_animation_component> >::operator*() const (this=0x548) at /usr/include/c++/10.2.0/bits/unique_ptr.h:408
#3  0x000056011ef71c90 in unit::anim_comp() const (this=0x0) at src/units/unit.hpp:1532
#4  0x000056011ef5e5b9 in game_lua_kernel::intf_put_unit(lua_State*) (this=0x7f129c00fda0, L=0x7f129c010078) at src/scripting/game_lua_kernel.cpp:2184
#5  0x000056011ef76a8b in dispatch<&game_lua_kernel::intf_put_unit>(lua_State*) (L=0x7f129c010078) at src/scripting/game_lua_kernel.cpp:150
[...]

PR #5159, opened 2020-09-20
avoiding unit placed events

sevu:
Guess after all this years I touch the Lua stuff behind the API …

One commit for #5158, other I found while testing for e3e80b4.

This all comes down to this comment:
https://github.com/wesnoth/wesnoth/blob/de9311e2622a9603834113cfd7f3fa2edb4de344/src/scripting/game_lua_kernel.cpp#L2583-L2584

It's the same method `[unstore_unit]fire_events=…` uses internally, just undocumented.

@CelticMinstrel, @gfgtdf can this go in as is?

### DIFF
--- a/data/lua/diversion.lua
+++ b/data/lua/diversion.lua
@@ -58,7 +58,7 @@ local status_anim_update = function(is_undo)
 		    changed_something = true
                     ec_unit.status.diversion = true
                     ec_unit:extract()
-                    ec_unit:to_map()
+                    ec_unit:to_map(false)
                     wesnoth.wml_actions.animate_unit {
                             flag = "launching",
                             with_bars = true,
@@ -79,7 +79,7 @@ local status_anim_update = function(is_undo)
 		    changed_something = true
                     ec_unit.status.diversion = false
                     ec_unit:extract()
-                    ec_unit:to_map()
+                    ec_unit:to_map(false)
                     wesnoth.wml_actions.animate_unit {
                             flag = "landing",
                             with_bars = true,
--- a/data/lua/feeding.lua
+++ b/data/lua/feeding.lua
@@ -26,7 +26,7 @@ wesnoth.game_events.add_repeating("die", function()
 			effect.increase_total = effect.increase_total + 1
 			u_killer_cfg.max_hitpoints = u_killer_cfg.max_hitpoints + 1
 			u_killer_cfg.hitpoints = u_killer_cfg.hitpoints + 1
-			wesnoth.units.to_map(u_killer_cfg)
+			wesnoth.units.to_map(u_killer_cfg, false)
 			wesnoth.interface.float_label(ec.x2, ec.y2, _ "+1 max HP", "0,255,0")
 			return
 		end
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -498,7 +498,7 @@ function wml_actions.petrify(cfg)
 		unit.status.petrified = true
 		-- Extract unit and put it back to update animation (not needed for recall units)
 		unit:extract()
-		unit:to_map()
+		unit:to_map(false)
 	end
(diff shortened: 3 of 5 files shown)


CelticMinstrel:
I don't think changing `wml_actions` to `wesnoth.wml_actions was necessary in `move_unit_fake`, was it? Surely there was a `local wml_actions = wesnoth.wml_actions` earlier in the file?

The second commit seems fine, assuming that the change to core.lua was a function that moves a unit (I didn't look at the full file).

gfgtdf:
I'm not against this patch, but i think that it doesn't fix the problem its supposed to fix, even with this, other umc code or other newly mainline code can accidentatlyl fire unit_placed events. I think the best way to make sure that this doesn't cause problem is to mak addon authors aware of the fact that unit_placed events might be firered rather too often than too rarly and that the should make their code more robust against unit_placed  events that are fired multipel times if they use unit_placed events.

sevu:
It's true, someday somebody might run into it again. The place where somebody might look before writing such a function would be the LauAPI page of tomap(). I think it's best we document it and add some kind of disclaimer there. I could make another note that the behaviour of this function might change like with undocumented things.

It's better to get unit placed events work reliably. They combine prerecruit and prerecall, but also work for units placed with debug mode and plague - basically the perfect solution if it wouldn't be for this. There's not many people who would write code breaking this.

gfgtdf:
> It's better to get unit placed events work reliably

Umc will fire unit_places events, also i'm also sure that your patch also doesn't catch all mainline cases of 'unexpected' unit_placed events  (ai code, scenario specific lua/wml code ...).  If a code that uses unit placed events breaks by unit places event being  fired multiple times i _really_ recommend you fix the code that uses unit placed instead of relying  the it will now fire just once that , the wiki also explicitly says that "can fire an unlimited number of times for the same unit" about unit placed events.

> basically the perfect solution

perfect solution for what?



I am also not sure whether not fireing  on [unit_move] is really a good change, since  people might be using unit placed events to track the locations of a unit and since afaik [move_unit] doesn't fire moveto events  their code might be relying on unit_placed when keeing track on location changes via [move_unit], in particular since the wiki also explicitly mentions that [move_unit] fires unit places events.

CelticMinstrel:
> since people might be using unit placed events to track the locations of a unit

If people are doing this, then that's their mistake. Unit placed is about _placing_ units, not moving them. It should never fire if a unit on the map is moved to a location on the map. This can likely be fixed in the actual implementation of the to_map function.

If someone wants to track the location of a unit, they should use moveto or hex entered events. We should also ensure that `[move_unit]` fires (or at least, _can_ fire) these events.

> in particular since the wiki also explicitly mentions that [move_unit] fires unit places events.

Well, with this PR we'd just have to edit it to retract that statement. An event called "unit placed" should only fire when a unit is placed. Moving a unit is a completely different action and should fire a separate event.

It's fine to note that the event might redundantly fire for the same unit sometimes. However, whenever the engine _can_ identify that the event has already fired for a given unit, then it should refrain from firing it again.

Pentarctagon:
@CelticMinstrel any thoughts on merging this?

CelticMinstrel:
The first two commits look good. The third commit looks wrong – it defines a function that's clearly meant to capture the `current_unit` variable, but it's outside the scope in which `current_unit` exists, so it's probably going to throw an "unknown variable" error?

Pentarctagon:
@sevu Are you around to look into this?

sevu:
Currently not, but later this month. IIRC the third comment worked during testing … maybe because it is a sub-function?

sevu:
Updated and tested. One commit has been fixed in master and is dropped.

CelticMinstrel:
Looks fine to me.

Pentarctagon:
This can be merged then?

Issue #5155, opened 2020-09-17
Orb color for UtBS's disengage ability (unit can move but can't attack)


stevecotton:
UtBS and some UMC abilities allow a unit to attack and then move, although it can't attack again. Currently the unit gets the partially-moved orb color, the same as a unit that can still attack. I'm wondering about adding a new orb color specifically for this state, shown in this prototype image with white and teal:
![wesnoth_disengage_white_teal](https://user-images.githubusercontent.com/101462/93460858-2ed27d80-f8e4-11ea-938c-8bc3982ffb19.png)

The orb colors themselves are configurable in advanced preferences, the main change here would be to add the extra state to the orb-choosing code. Adding documentation for this would overlap with #4455.

Now that I have an issue number, a commit with some some very early experimental code will get pushed and auto-linked to this issue by Github.

@Pentarctagon: I'd be interested in putting this on the 1.15.7 roadmap.
@nemaara: as UtBS is the main driver for this, how do you feel about it?

cooljeanius:
This always confused me when playing thru UtBS and a new orb color would definitely help; thanks for filing this!

AI0867:
I'm not opposed to this, but do consider the case where `max_attacks` is larger than 1. In that case, you may end up with `0<attacks_left<max_attacks`. I would say that as the unit can still attack, it should have a yellow orb, and only once `attacks_left` has hit 0, should the orb use the new color, but other opinions could be defensible.

Somewhat related question: Is it `max_attacks` or `attacks` or both? [UnitTypeWML](https://wiki.wesnoth.org/UnitTypeWML) says `attacks`, [SingleUnitWML](https://wiki.wesnoth.org/SingleUnitWML) says `max_attacks` and I thought the keys were supposed to match these days.

Pentarctagon:
@stevecotton given UtBS is the main reason for this, I'll add it to the roadmap if nemaara is good with the idea.

CelticMinstrel:
So we have an orb colour when a unit can attack and move, and an orb colour when a unit cannot attack or move. This new one is for when a unit can move but not attack. Is there one for when a unit can attack but not move? I think that currently uses the partial colour, right? Should we add another new colour for that state?

stevecotton:
> consider the case where max_attacks is larger than 1. In that case, you may end up with 0<attacks_left<max_attacks. I would say that as the unit can still attack, it should have a yellow orb, and only once attacks_left has hit 0, should the orb use the new color

[...]

Issue #1424, opened 2017-05-08
MP - Orb's color change for allied units not able to move/do anything more (GNA #22964)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?22964)** by anonymous on 2014-11-17

In a MP campaign like LoW the cooperation is an important part.

When the situation is complicated and needs some time to be played, it's easy to be lost about what is already done and what can still be done because allied unit's orb's color is always blue. This leads to an impossibility to help, or ask help after some moves in an easy fashion.

An other orb color for allied units that are not able to move/do anything more will help.

Can it be possible to have that feature implemented?

This was asked and quoted as doable [in that post](://forums.wesnoth.org/viewtopic.php?f#12&t39570&start=45#p577765)

(Reproduced on linux mint 17)
Release: 1.11.19
Priority: 5 - Normal
Severity: 1 - Wish

stevecotton:
I don't know for certain, but I guess the reason this isn't implemented is to hide the AI's tricks during AI turns (see #4428). I'm currently looking at the orb-drawing code, and considering adding an option to show normal colors for allied units; will post the proof-of-concept for that, even if it doesn't work out.

soliton-:
I don't think hiding AI tricks is the reason nor do I think should this reveal AI tricks that that would be a big deal. The benefit would be for allied sides played by a human player.

CelticMinstrel:
Even if it _does_ reveal AI tricks and this is undesirable, you could just make it not apply to AI sides at least by default. Other than allied sides played by a human player, I suppose this might be useful for an allied side that you can give commands to. Although the core system of events for doing that was removed at some point, there's nothing to prevent an add-on campaign from rolling their own system for it.

stevecotton:
@Pentarctagon I'd like to put this change into 1.15.6; the draft PR includes the disengage orb too (#5155), however the two features can be separated if desired.

Pentarctagon:
I've no objections, at least.

stevecotton:
Implemented in #5180, no idea why that didn't auto-link.

PR #5162, opened 2020-09-24
WIP: Orb refactor, disengage orb and allies' orbs

stevecotton:
This is work-in-progress, particularly for the first of the three
commits. I've done a lot of moving code around, and it doesn't
feel right yet, so I'd welcome all suggestions, even ones that
throw these changes out and implement the two features in different ways.

The first commit is refactoring with no new features, and then the other
two are features using the refactored code's new abilities.

I'm expecting build failures on MacOS, as this adds a new .hpp and .cpp
file without adding them to the XCode project.

Refactoring the orb-coloring code, adding new units/orb_status.hpp
---

Instead of creating lots of image::locators and calling a family of
similarly-named preferences functions, this refactor encapsulates that logic
and uses an enum.

WIP: This needs the new orb_status files adding to the MacOS build.

WIP: Does it make sense to add the orb_status files instead of just
putting these bits in an existing file?

WIP: Should the pref_show_orb and get_orb_color functions move into
src/preferences/general.hpp?

WIP: Some of the changes here are just to reform and reorder lists, ready
to add the disengaged orb to those lists. Should they instead be left to
the disengage commit?

WIP: Add a new orb color for the "disengaged" state (#5155)
---

This will match the mounted Quenoth units' "disengage" skill, when they
can still move but can't attack. It should also trigger for some UMC abilities
that get extra moves after a character attacks.

WIP: The select_orb_colors dialog's layout no longer fits its window size.

WIP: Should addig the extra color (teal) be done here? I've ended up using
the existing white color for the default disengage orb.

WIP: During allies' turns, their units' orbs show the move status (#1424)
---

WIP: the orb-refactor commit that this depends on is WIP.
WIP: check for todos and WIPs, including further down this commit message.

Instead of using the allied orb during their turn, the unmoved, moved and
partial orbs are shown. The player's own units will be shown with the moved orb
during the ally's turn, which matches the previous behavior. This is intended for
co-operative MP campaigns, where it will make it easier for players to discuss
possible moves. The UX is also visible in SP scenarios with allied sides, HttT's
first scenario is an easy place to see what it does.

WIP: test this paragraph. When one player owns multiple sides, or when multiple
players are hot-seating, the usual handling still applies - the orbs are drawn
as if the one that has most recently had its turn is the only side that the
player owns.

The disengaged orb will also be supported, if that feature merges. The code in
this commit needs no further changes to support OrbStatus::disengaged.

### DIFF
--- a/projectfiles/CodeBlocks/wesnoth.cbp
+++ b/projectfiles/CodeBlocks/wesnoth.cbp
@@ -1169,6 +1169,8 @@
 		<Unit filename="../../src/units/make.hpp" />
 		<Unit filename="../../src/units/map.cpp" />
 		<Unit filename="../../src/units/map.hpp" />
+		<Unit filename="../../src/units/orb_status.cpp" />
+		<Unit filename="../../src/units/orb_status.hpp" />
 		<Unit filename="../../src/units/ptr.hpp" />
 		<Unit filename="../../src/units/race.cpp" />
 		<Unit filename="../../src/units/race.hpp" />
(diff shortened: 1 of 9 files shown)


CelticMinstrel:
If this is all WIP, why is it not opened as a draft?

stevecotton:
Thanks for the suggestion - I hadn't used the draft feature before and didn't think of it.

stevecotton:
I'm going to open a new PR for the new (no longer WIP) implementation of #1424.

PR #5163, opened 2020-09-24
Fix #4961 (halo animations sometimes freeze)

### LINKED ISSUE #4961: [item]halo= ... animations (sometimes) freeze
The Battle for Wesnoth version 1.15.3+dev (1ff6b7bbc4a-Modified)
Running on openSUSE Leap 15.1
Distribution channel: Default

#### Libraries
Boost:             1.66
OpenSSL/libcrypto: 1.1.0i (runtime 1.1.0i)
Cairo:             1.15.10 (runtime 1.15.10)
Pango:             1.40.13 (runtime 1.40.13)
SDL:               2.0.8 (runtime 2.0.8)
SDL_mixer:         2.0.2 (runtime 2.0.2)
SDL_ttf:           2.0.14 (runtime 2.0.14)

#### Features
JPEG screenshots:               yes
Lua console completion:         yes
Legacy bidirectional rendering: yes
D-Bus notifications back end:   yes

#### Current video settings
SDL video drivers:       [x11] wayland KMSDRM dummy
Window size:             1920x1015
Screen refresh rate:     60
Screen dots per inch:    128.000000x128.186920
Screen dpi scale factor: 1.333333x1.335280

### Describe the bug
Animation using the `[item]halo=...` key isn't updated unless there is something else causing a redraw, at least in some circumstances.  Specifically, load the attached save-file from UtBS, and observe the central monster's halo.  It is supposed to blink, but it only does that if I move the mouse around.  If you debug-create a unit that has a standing animation (Sun Sylph on cave terrain, for example), the monster halo blinks as intended.

Other circumstances, such as the dancing mage in the center of MP map Aethermaw, do not seem to have this issue, possibly because of animated terrain.

[UtBS-The Final Confrontation-Auto-Save4_20200621-102722.gz](https://github.com/wesnoth/wesnoth/files/4810418/UtBS-The.Final.Confrontation-Auto-Save4_20200621-102722.gz)

DisherProject:
As [this](https://github.com/wesnoth/wesnoth/blob/master/src/display.cpp#L2521-L2524) comment says, we are only redrawing haloes if anything else in the map is invalidating at least one hex.

`halo_man_->unrender(invalidated_)` actually invalidates the hexes covered by haloes which need to be updated, so i'm just moving the call outside the if-clause.

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -2512,16 +2512,12 @@ void display::draw(bool update,bool force) {
 
 	draw_init();
 	pre_draw();
-	// invalidate all that needs to be invalidated
+	// invalidate animated terrain, units and haloes
 	invalidate_animations();
 
 	if(!get_map().empty()) {
 		//int simulate_delay = 0;
 
-		/*
-		 * draw_invalidated() also invalidates the halos, so also needs to be
-		 * ran if invalidated_.empty() == true.
-		 */
 		if(!invalidated_.empty()) {
 			draw_invalidated();
 			invalidated_.clear();
@@ -3131,6 +3127,8 @@ void display::invalidate_animations()
 			new_inval |=  u->anim_comp().invalidate(*this);
 		}
 	} while (new_inval);
+
+	halo_man_->unrender(invalidated_);
 }
 
 void display::reset_standing_animations()
--- a/src/game_display.cpp
+++ b/src/game_display.cpp
@@ -226,7 +226,6 @@ void game_display::post_draw() {
 
 void game_display::draw_invalidated()
 {
-	halo_man_->unrender(invalidated_);
 	display::draw_invalidated();
 	if (fake_unit_man_->empty()) {
 		return;


stevecotton:
I'm not familiar with this code, but it looks like a minimal change on top of code that's already too complex.

However, it also looks like a change that increases the messiness of the code - moving a draws-to-screen call into a sets-invalidation-flags function. So overall I'm slightly against merging this.

CelticMinstrel:
For some reason, #4961 doesn't seem to be linked from this PR (though at least the commit is), so this comment endeavours to fix that.

DisherProject:
> I'm not familiar with this code, but it looks like a minimal change on top of code that's already too complex.

Do you mean that it is likely to create some new issues?

> However, it also looks like a change that increases the messiness of the code - moving a draws-to-screen call into a sets-invalidation-flags function. So overall I'm slightly against merging this

I could split the `halo_manager::unrender` method into two functions: the former would just invalidate the haloes, the latter would effectively unrender them. Then, i could call the invalidation [here](https://github.com/wesnoth/wesnoth/pull/5163/commits/7aacf94dddeed3669ffe99136c3346977a154d60#diff-c69d99f823238fb59e4c6efdf3a9b5563ad606396fb51bb644871ba129454fa5R3131) and the redraw [here](https://github.com/wesnoth/wesnoth/pull/5163/commits/7aacf94dddeed3669ffe99136c3346977a154d60#diff-1ed1a70269b778cd638100875bab830983a2f0e1104d23098fbd1d01f4bb41c3L229).

stevecotton:
> Do you mean that it is likely to create some new issues?

I'm worried that it will, yes.

However, I'm unfamiliar with the code and not wanting to learn a new code section, I don't know what should be done with it.

Pentarctagon:
@mesilliac would this be made irrelevant by the rendering overhaul going on?

mesilliac:
Hmm. Eventually yes but not yet. From a quick look this could probably just be merged, as it's a small change. I'll take a look at it later.

mesilliac:
Verified that it works and the change is okay. Merged

PR #5164, opened 2020-09-25
Run the wmlindent and wmllint Makefile targets from data/tools 

cooljeanius:
I was just messing around seeing what different Makefile targets in different directories did, and it looks like the ones that ran `wmlindent` and `wmllint` from the data/tools directory did some useful tweaks, so maybe let's merge those? I might have screwed up some stuff when doing my own merge from master when I pulled to bring my own branch up-to-date, so if so, I'm sorry about that and can try to fix it.


Pentarctagon:
Some of it looks fine; things like adding `image=wesnoth-icon.png` to messages without an image attribute is definitely incorrect though.

cooljeanius:
> Some of it looks fine; things like adding `image=wesnoth-icon.png` to messages without an image attribute is definitely incorrect though.

Is there a way to tell `wmllint` to stop doing that? I had this problem with some UMC, too...

Pentarctagon:
To be honest I'm not sure when that would ever be correct, since `image` isn't a required attribute, so I'd say the fix would be to remove whatever's doing that from wmllint.

gfgtdf:
> To be honest I'm not sure when that would ever be correct

The reason is probably that at some point image defaulted to wesnoth-icon.png if omitted and wmlint is a tool fo backwards compatbility iirc.

stevecotton:
I think wmllint is named incorrectly - it's more of a convert-to-new-version tool than a lint tool.

.gitignore
```diff
@@ -202,6 +202,7 @@ fonts/.uuid
 *~
 .deps
 .dirstamp
+*.cfg-bak
```
stevecotton, line 205: Can we treat wmllint's behavior of creating backup files as a bug, that has no place in a world with version-control tools? (That would probably need the default change to just linting, without making unexpected changes to files).

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
```diff
@@ -384,7 +386,7 @@
         {VARIABLE_OP malin.experience add 8}
         [floating_text]
             x,y=$malin.x,$malin.y
-            text="<span color='#800080'>" + "+8 exp" + "</span>"
+            text=_ "<span color='#800080'>" + "+8 exp" + "</span>"
```
stevecotton, line 389: @nemaara looks like we should put an underscore on this line, but applying to the "+8 exp" string rather than the color strings.

Might just make the whole line one translatable string, as we already throw marked-up text at the translators.

.gitignore
cooljeanius: That would probably require changing the behavior of the `-r` flag to `wmllint` as well, since it relies on those files

Pentarctagon:
@gfgtdf still, it's not a valid thing for it to be doing anymore.

.gitignore
CelticMinstrel: Or at least have a flag to disable the backups behaviour.

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
```diff
@@ -351,7 +352,7 @@
         {VARIABLE_OP malin.experience add 4}
         [floating_text]
             x,y=$malin.x,$malin.y
-            text="<span color='#800080'>" + "+4 exp" + "</span>"
+            text=_ "<span color='#800080'>" + "+4 exp" + "</span>"
```
CelticMinstrel, line 355: I think the underscore should instead go after the first `+`.

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
CelticMinstrel: I think the underscore should indeed go after the first `+`. Throwing markup text at translators is perhaps okay for complex text such as the help document, but I don't think it should be done in avoidable situations like this.

data/campaigns/Descent_Into_Darkness/scenarios/03_A_Haunting_in_Winter.cfg
```diff
@@ -1029,21 +1029,21 @@
                 {VARIABLE_OP malin.experience add 30}
                 [floating_text]
                     x,y=$malin.x,$malin.y
-                    text="<span color='#800080'>" + "+30 exp" + "</span>"
+                    text=_ "<span color='#800080'>" + "+30 exp" + "</span>"
```
CelticMinstrel, line 1032: I think the underscore should instead go after the first `+`, here and for all similar instances (there are quite a few in this file).

data/campaigns/Descent_Into_Darkness/scenarios/06_Return_to_Parthyn.cfg
```diff
@@ -496,21 +497,21 @@
         {VARIABLE_OP malin.experience add 12}
         [floating_text]
             x,y=$malin.x,$malin.y
-            text="<span color='#800080'>" + "+12 exp" + "</span>"
+            text=_ "<span color='#800080'>" + "+12 exp" + "</span>"
```
CelticMinstrel, line 500: Again, I think the underscore should instead go after the first `+`.

data/campaigns/Heir_To_The_Throne/scenarios/23_Test_of_the_Clans.cfg
```diff
@@ -27,7 +27,7 @@
 				if not cfg.silent then
 					local splash = _("Still $units_to_slay clan member to defeat!", "Still $units_to_slay clan members to defeat!", n)
 					wesnoth.wml_actions.print{
-						text = splash:vformat{units_to_slay = n},
+						text =_  splash:vformat{units_to_slay = n},
```
CelticMinstrel, line 30: This is dead wrong, as this is Lua code rather than WML. It will result in an error at runtime.

@Elvish-Hunter Is it easy to prevent wmllint from touching embedded Lua code?

data/campaigns/Secrets_of_the_Ancients/utils/zombie-utils.cfg
```diff
@@ -39,7 +39,7 @@
             name=zombies
             mode=merge
             [split]
-                list= _ "(Bat),
+                list= "(Bat),
```
CelticMinstrel, line 42: This looks like it's just terrible code. This is not a good way to implement this, at all. I would say this code just needs a major overhaul. In particular, all the unit names in this list are already translated in the core game data - why does SotA need to have its own version of each one?

I've opened #5170 to track this as a separate issue. In the meantime, the removal of these two translation marks should be reverted.

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
```diff
@@ -575,9 +575,9 @@
 
         [message]
             speaker=Scylla
-            message= _ "Yesss, we remember the oath...  Sssslay them all!" # wmllint: no spellcheck
-# leaving this here because it might have been a reference to the alien?  It really isn't clear.
-#            message= _ "Sssslay them all! In the name of the Ssscaled One!" # wmllint: no spellcheck
+            message= _ "Yesss, we remember the oath... Sssslay them all!" # wmllint: no spellcheck
+            # leaving this here because it might have been a reference to the alien?  It really isn't clear.
+            #            message= _ "Sssslay them all! In the name of the Ssscaled One!" # wmllint: no spellcheck
```
CelticMinstrel, line 580: The whitespace _after_ this `#` character should be replaced with a single space. @Elvish-Hunter I wonder, is this automatable in wmllint?

data/core/terrain-graphics/new-macros.cfg
```diff
@@ -47,13 +47,13 @@ base
     # Places {IMAGESTEM}[1-11].png as overlay tiles.
 
 #arg LAYER
-0#endarg
+    0#endarg
```
CelticMinstrel, line 50: I don't think this should be indented. Perhaps we need to add some new rules to wmllindent?

data/core/terrain-graphics/new-macros.cfg
```diff
@@ -3529,7 +3529,7 @@ wall#endarg
             type={TERRAINLIST}
         [/tile]
 
-#        rotations=bl
+        #        rotations=bl
```
CelticMinstrel, line 3532: Here's another case where there's excessive whitespace _after_ the comment character. (Or rather, soneone added a comment character before the indentation rather than after it.)

CelticMinstrel (changes requested):
This run reveals both a number of issues in the actual WML source and also some issues with the wmllint and wmlindent tools.

I'm not sure if running wmlindent on the schema directory is okay. I know that the gui directory uses a different indentation standard (for some reason), and it looks like the schema directory shares that difference, so perhaps it shouldn't be run there? That's not really my decision to make, mind you, just a thought. Of course, the main reason the schema directory shared that difference in the first place is because it was almost entirely crafted by me.

I'm not so sure that all those additions of `image=wesnoth-icon.png` are wrong. It's not incorrect to use it for narration, though it could be redundant, perhaps? If it _is_ considered wrong (possibly due to being redundant), then the rule should just be removed from wmllint.

There are two changes here that actually break things. Either wmllint should be fixed, or comments should be added to tell wmllint to not touch those lines. The two changes both have specific comments and involve translation marks - adding one in Lua code where it doesn't belong, and removing one used in an unconventional place but for good reason. Because of these to changes, I would say this PR _cannot_ be merged in its current state. Despite that, it's still useful to _see_ what sorts of changes wmllint did and where it went wrong, so thanks for submitting the PR!

data/campaigns/Descent_Into_Darkness/scenarios/06_Return_to_Parthyn.cfg
CelticMinstrel: Incidentally, this seems like a situation that wmllint _might_ be able to detect automatically? If a line consists of three strings concatenated by `+` where the first and third string both begin with `<` and end with `>`, then it's probably this situation. In fact, maybe that could even be extended to an arbitrary sequence of strings concatenated by `+`.

@Elvish-Hunter Any thoughts on this idea?

Pentarctagon:
> I'm not sure if running wmlindent on the schema directory is okay. I know that the gui directory uses a different indentation standard (for some reason), and it looks like the schema directory shares that difference, so perhaps it shouldn't be run there? That's not really my decision to make, mind you, just a thought. Of course, the main reason the schema directory shared that difference in the first place is because it was almost entirely crafted by me.

IMO, unless there's a specific reason to *not* use the same indentation standard as everywhere else, then the same standard should be used.

> I'm not so sure that all those additions of image=wesnoth-icon.png are wrong. It's not incorrect to use it for narration, though it could be redundant, perhaps? If it is considered wrong (possibly due to being redundant), then the rule should just be removed from wmllint.

I'd say it's wrong for the reason that not having any image in a [message] is a valid stylistic choice for a campaign author to make (I'm thinking in particular of some DiD's texts), and a script whose purpose is supposed to help with fixing compatibility issues shouldn't be making extraneous changes that have no impact on compatibility.

gfgtdf:
> @gfgtdf still, it's not a valid thing for it to be doing anymore.

Yes i agree but i think this is more a general flawa of wmlint: What changes it does does currently not depend on the version of the "source" meaning there is no way to tell if to do only "1.10 to 1.14" changes for example. It will always treat and addon as if it came from wesnoth 1.0 and apply all chaqnges to preserve behviour if possible.

CelticMinstrel:
> IMO, unless there's a specific reason to not use the same indentation standard as everywhere else, then the same standard should be used.

Well, _is_ there a specific reason when it comes to the gui directory?

Pentarctagon:
Unless for some absurd reason the GUI WML is tab sensitive like python, then no.

Pentarctagon:
> Yes i agree but i think this is more a general flawa of wmlint: What changes it does does currently not depend on the version of the "source" meaning there is no way to tell if to do only "1.10 to 1.14" changes for example. It will always treat and addon as if it came from wesnoth 1.0 and apply all chaqnges to preserve behviour if possible.

I don't think that's a valid thing to be doing anyway, to be honest.  wmllint should end up doing the correct behavior to update the WML to what's correct for the version of Wesnoth it came with.  So (and I'm just using arbitrary versions here), if it were valid to set the `image=` attribute on all [message]s from 1.2 to 1.6, then the wmllint from 1.7 and later versions should have that logic removed.  It doesn't make a lot of sense to me that wmllint from 1.14 would support upgrading only to 1.10 or 1.12 for example.

gfgtdf:
> I don't think that's a valid thing to be doing anyway, to be honest. wmllint should end up doing the correct behavior to update the WML to what's correct for the version of Wesnoth it came with

Well if we assume that the addon was written for a very old wesnoth version then this is exactly what it does.


> It doesn't make a lot of sense to me that wmllint from 1.14 would support upgrading only to 1.10 or 1.12 for example.

I didn't suggest that, i suggested being able to specify the version where that addon came from not that target version to whcih version to port.

>  So (and I'm just using arbitrary versions here), if it were valid to set the image= attribute on all [message]s from 1.2 to 1.6, then the wmllint from 1.7 and later versions should have that logic removed.

I'm not sure what you mean here do you suggest to remove all compatibility code form wmlint on each new dev release and let it only handle change from the last stable version to the current stable version?

Pentarctagon:
I mean that if a change wmllint is making is only correct through version 1.6, for example, then it shouldn't keep making that same change for versions that come after 1.6.

gfgtdf:
But then porting older addons code would no longer be correct.

gfgtdf:
I mean tbh i don't know whether it applies to the concrete case, and i'm totally fine with removing the code to add wesnoth-icon.png here. But my concern still applies: take for example the [terrain_mask] tag, im quite sure that it's behavior over time, and if we want to change  in a future update we have no way to tell wmlint how it should interpret the 'old' tag.

CelticMinstrel:
I don't think it's too unreasonable to assume that any addon being ported by wmllint comes from the previous stable version of the game. If that assumption is false, then they can first run it through the version of wmllint that came with the previous stable version of the game, right? It should keep any changes that are going through a deprecation cycle, of course, but perhaps ancient changes like adding this image could be removed?

Pentarctagon:
@gfgtdf to clarify, I'm not saying wmllint should get rid of steps needed to provide an upgrade for very old add-ons.  So if to get an older add-on working wmllint needs to change A -> B, B -> C, C -> D, then the A and B should should still be kept.

If Wesnoth goes back to expecting C though, then the C -> D step should be removed from wmllint.

gfgtdf:
Well yes of course but i thought that that was not the case we were talking about here. As i understood it, we were assuming the situation that in an old wesnoth version [message]image= defaulted to wesnoth-icon.png and some wesnoth version changed the code to make it default to no image, and then added this compatibility code to preserve this behavior for old addons.

Pentarctagon:
Ah, alright.  If that's the case, then this would just be a one off decision of "wmllint should make this particular change anymore".

.gitignore
AI0867: There are probably still plenty of UMC that are developed without the assistance of version control. They probably rely on this feature for so far as they actually use the tool.

AI0867:
When it comes to old versions to upgrade from, wmllint has repeatedly been cleaned of its more ancient upgradepaths, with the argument that if you really want to port an ancient add-on, you can use ancient versions of wmllint to do the first steps. Unfortunately, wmllint still combines several behaviours: 
 * correct mistakes in current WML 
 * upgrade from previous stable WML to current WML
   * this is not one monolithic upgrade in the case of the development versions, especially because WML is not versioned
 * possibly upgrade even earlier stables to current WML

Wmllint rules *are* required to only modify things if it's unambiguously the older version or an incorrect usage, and the WML could not have been produced by itself or an author writing correct modern WML. If this is no longer the case, we should perhaps reinstate that rule.

Regarding the image= rule, that at least used to be actual policy, and was probably started when we moved from dialog-style [message] to the current style. I'm not sure if it was also a default from the GUI-engine.

In the past, I believe it was also a rule that the all the WML included in a release had to wmllint clean (that is, wmllint would make no changes to it, nor complain about anything). This PR seems to be evidence that that rule is no longer used.

gfgtdf:
> . I'm not sure if it was also a default from the GUI-engine.

I looked at the code and it indeed didn't seem to be the case. When i wrote that supposition i assumed that the main(only) usecase for wmlint was to upgrade older addons

gfgtdf:
After looking at this pr content i can say that the majority of the  changes, here are either questionable to just wrong. and that the pr can definitely not be merged as-is. Also if want to do this seperate the imlindend form the wmllint commits to mkae it easioer to ewview

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
sevu: There was an wmllint magic comment to suppress this. I think for this case, it was to add at the end of the same line
` # wmllint: noconvert`

They are described in the comment text of the wmllint python script.

.gitignore
sevu: There's an option to remove the backup files instead: -c / --clean

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
```diff
@@ -339,6 +339,7 @@
         [message]
             speaker=narrator
             message= _ "Ignorance is the parent of fear."
+            image=wesnoth-icon.png
```
sevu, line 342: It's possible to shut up wmllint while not providing a picture, with
`image=none` (special keyword since 1.14.x)

sevu:
> > Some of it looks fine; things like adding `image=wesnoth-icon.png` to messages without an image attribute is definitely incorrect though.
> 
> Is there a way to tell `wmllint` to stop doing that? I had this problem with some UMC, too...

Yes, `image=none`, new since 1.14.

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
nemaara: No idea what the right way is but sure making it translatable is totally fine.

stevecotton:
> Yes, image=none, new since 1.14.

I assumed that was for stopping a unit image appearing for messages said by units. If it's for working around a feature in wmllint then I think that's a bug in wmllint.

AI0867:
> > Yes, image=none, new since 1.14.
> 
> I assumed that was for stopping a unit image appearing for messages said by units. If it's for working around a feature in wmllint then I think that's a bug in wmllint.

The whole idea of the current `[message]` UI is to have units appear to be speaking on the sides. The shield is the convention for the narrator. What is the usecase for not having any image at all?

At the time, not having an image specified was considered to be an oversight by the UMC dev.
If the speaker was a unit, the portrait would be used. If the speaker was the narrator, the shield should be used unless there was a more specific thing to show instead.

stevecotton:
What about the shield should make the player think of a narrator? Looking through the DiD changes in the current PR, pictures of books and monoliths might fit the messages, but a shield with swords and the SoF doesn't.

If we want to say that every message should have either an image or an explicit "none", then we should put that in the validation schema for mainline.

CelticMinstrel:
That shield also happens to be the Wesnoth logo, which is probably something to do with the reasoning for using it with narration.

AI0867:
> That shield also happens to be the Wesnoth logo, which is probably something to do with the reasoning for using it with narration.

It's indeed the wesnoth logo. It means "It is the game that is speaking to you now, not any of its characters.".

I still don't know what the usecase is for a `[message]` without an image. Without it, the `image=none` feature sounds like a misguided workaround.

I just checked which files in mainline are missing `image=wesnoth-icon.png` according to wmllint, and the sample I checked are all scenarios that have been recently rewritten, modified or imported. In other words, wmllint is actually informing us of lacking tests and maintenance here.

Pentarctagon:
If you want all of the player's focus to be on the text, you wouldn't add an image since that would also draw their attention.

stevecotton:
> It means "It is the game that is speaking to you now, not any of its characters.".

How should the player know that it means that? What about that image conveys that message?

CelticMinstrel:
Maybe the fact that it's literally the game's logo?

CelticMinstrel:
Note: I think that if we want the Wesnoth logo to be the default for narration, then it would make more sense to actually make it the default rather than getting wmllint to add it everywhere. The implementation of `[message]` could easily set the image to the Wesnoth logo if it finds `speaker=narrator` with no image.

Pentarctagon:
> Note: I think that if we want the Wesnoth logo to be the default for narration, then it would make more sense to actually make it the default rather than getting wmllint to add it everywhere. The implementation of [message] could easily set the image to the Wesnoth logo if it finds speaker=narrator with no image.

I agree with this, FWIW.  Though I assume that would then add some work for @nemaara to add `image=none` in a bunch of places.

nemaara:
Speaking as a UMC author now, I would be pretty annoyed at being forced to have all my messages have an explicit image=none. Maybe it would be a blanket find and replace thing, but it still seems like a bunch of unnecessary work from the UMC side just so that some mainline script stops complaining. I also tend to think that a default no image makes more sense than the wesnoth logo.

CelticMinstrel:
> Though I assume that would then add some work for @nemaara to add `image=none` in a bunch of places.

Well, not exactly? If the mainline policy is to use the shield for narration, then she doesn't have to do anything; it's the current state (where nothing is used) that's wrong. _If_ the mainline policy is to use the shield for narration, it probably shouldn't use `image=none` _anywhere_ really. I mean there might be an _occasional_ exception, but…

Unless you mean in her add-ons, in which case yeah.

data/campaigns/Descent_Into_Darkness/scenarios/03_A_Haunting_in_Winter.cfg
cooljeanius: actually this whole section could probably be simplified so that the same thing isn't repeated 3 times; mind if I do it that way instead?

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
cooljeanius: done in ad68ef0

data/campaigns/Heir_To_The_Throne/scenarios/23_Test_of_the_Clans.cfg
cooljeanius: reverted in ad68ef0

data/campaigns/Secrets_of_the_Ancients/utils/zombie-utils.cfg
cooljeanius: reverted removal of translation marks in ad68ef0

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
cooljeanius: I generally prefer ifdef-ing out code over commenting out code, but in the meantime, I've fixed the whitespace in ad68ef0

Pentarctagon:
I mean that, given she's the SP lead, mainline's policy for that is really up to her.  My own opinion as well is that having the wesnoth logo makes sense for cases where it's actually the game "speaking" to the player about the game itself (such as giving a hint, explaining some sort of scenario/campaign mechanic, that sort of thing) rather than as a default for all cases.

CelticMinstrel:
Fair enough.

gfgtdf (changes requested):
As i said before this cannot be mnerged as-is since its one  big monster commits not only hard too review but also makes it hard to understand for anyone that reads wesnoth commits history. 

At least you have to split it into a commit that does the whitspeace changes and another commit that doe the wmllint changes.

cooljeanius:
> As i said before this cannot be mnerged as-is since its one big monster commits not only hard too review but also makes it hard to understand for anyone that reads wesnoth commits history.
> 
> At least you have to split it into a commit that does the whitspeace changes and another commit that doe the wmllint changes.

How do I do that? Do I have to start over from a new branch and a new pull request? Or is there a way to do so within this branch?

gfgtdf:
> How do I do that? Do I have to start over from a new branch and a new pull request? Or is there a way to do so within this branch?

You can force push, but it depends a on what git tool you are using the git commandline programm or some gui programm?

cooljeanius:
> > How do I do that? Do I have to start over from a new branch and a new pull request? Or is there a way to do so within this branch?
> 
> You can force push, but it depends a on what git tool you are using the git commandline programm or some gui programm?

A combination of the two; I use the GitHub GUI app for basic stuff and the command-line program for anything else

gfgtdf:
you can make a backup of your stuff with "git branch", then do `git reset HEAD~4 --hard` to undo the last 4 commits then, run wmlindend then commit then do the other changes and comit again, finaly push with "push  -f" to overwriote your pull request

cooljeanius:
> you can make a backup of your stuff with "git branch", then do `git reset HEAD~4 --hard` to undo the last 4 commits then, run wmlindend then commit then do the other changes and comit again, finaly push with "push -f" to overwriote your pull request

ok done

gfgtdf:
I think its best if we discuss the wmlindent changes first, then merge then and then discuss the second part, so some notes about the wmlindent changes::

1) do we all agree that we want to use spaces over tabs in the schema config files? I personally don't have a strong opinion on this but in case that someone wants to work on these files (keep them updated in the future) she can veto this change.


2) the 
```
#arg NAME
value #endarg
```
should follow the same rules as the 
```
#define NAME
value #enddef
```
meaning `value` should not be indented.

CelticMinstrel:
I'm against 1 but in favour of 2. However, if 1 _is_ implemented, then I think we should go all the way and also change the GUI config files to be space-indented. (Actually, 2 spaces might be better for those files… since the indentation can get really, really deep… although using tabs also means anyone can choose their own indentation size.)

More regarding 2: by inserting indentation there, it likely changes the actual semantics of the code - I seem to recall that whitespace is significant within preprocessor definitions, and it's also significant within quoted text, so depending on where the macro is used, an extra tab or four spaces in the argument default might give a different result.

Pentarctagon:
I'm pretty indifferent to tabs vs spaces (and if spaces, then exactly how many spaces to use).  So whichever way that goes, I just want it to be consistently applied to all WML.

gfgtdf:
We certainly shouldn't change mainline wesnoth indentation style, because it will ruin git blame for some time. for the same reason i'm against changing the indentation style of the gui2 wml.

I really think that having two different styles is a smaller problem than having git blame ruined by a 'change all indentation' style commits, in particular when the two areas are cleanly split like the gui2 wml from the rest or the schema wml.

gfgtdf:
@cooljeanius i'm wondering what command have did you run exactly for the wmlindent changes ?

cooljeanius:
> @cooljeanius i'm wondering what command have did you run exactly for the wmlindent changes ?

`make reindent` in data/tools; the target is in data/tools/Makefile

CelticMinstrel:
> because it will ruin git blame for some time

I'm pretty sure git blame can easily give you the second-most-recent change to the file, rather than the most recent one, so I don't see it being ruined by a full-scale indentation update. And GitHub's blame interface in particular makes it really easy to keep jumping back in time.

Pentarctagon:
It needs a relatively recent version of git, but there's [--ignore-rev](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revltrevgt).

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
```diff
@@ -372,6 +373,7 @@
         [message]
             speaker=narrator
             message= _ "A monolith of pale stone, luminous and bright, but cold and brittle to the touch."
+            image=wesnoth-icon.png
```
nemaara, line 376: Unless it's specified, these should all be image=none.

nemaara (changes requested):

data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
cooljeanius: by "these should all" do you mean all the monoliths, all the instances in DiD, or all instances in the entire PR?

nemaara: Everything in DiD. If it doesn't specify an image in the original, assume it's none.

cooljeanius: ok, I'm pretty sure I replaced all the icons introduced by 0165e25 with 488dc65; there were some instances where the icon was already there beforehand that I left alone, though.

Pentarctagon:
One remaining problem for wmlindent is that it indents the contents between `#arg`/``#endarg`, which is incorrect since that is sensitive to leading whitespace (as gtgtdf noted earlier).  Otherwise, I don't think ruining git blame is an issue, since it can be told to ignore certain commits.

gfgtdf:
> since it can be told to ignore certain commits.

github webpage balme (which is the only thing i use actually )cannot

CelticMinstrel:
> github webpage balme (which is the only thing i use actually )cannot

Sure it can. If you find the blame for the line of interest points to a commit that does nothing useful (such as reformatting), then you can click the little symbol between the date and the line number to jump to a blame prior to that commit.

gfgtdf:
> Sure it can. If you find the blame for the line of interest points to a commit that does nothing useful (such as reformatting), then you can click the little symbol between the date and the line number to jump to a blame prior to that commit.


That's something else than what pent desribed, and one very annoying click in particular since  git(hub) blame for a huge file might take some quite time to load.

CelticMinstrel:
It's true that it's less convenient, but it works. You implied that it was not possible. It's not like this is going to be a change from the current blame workflow, anyway — there are plenty of other "nothing" commits that just reformat code or the like, so you'll need to deal with this regardless of whether we run wmlindent.

It would definitely be better to not submit misindented code, true, but unless we have some sort of task on Travis that runs wmlindent to verify everything's good, there _will_ be misindented code submitted because someone missed it.

Pentarctagon:
Adding a step to an existing job to validate that the indentation is good would be pretty easy, but everything existing right now would first need to be able to pass wmlindent without being changed.

CelticMinstrel:
Incidentally, I tried running wmlindent locally on data/ and it failed at some point with an encoding error. Did anyone else get that? It just made a whole lot of changes and then suddenly crashed.

I was looking into how to make it not indent the arg stuff, no luck yet unfortunately. Maybe someone else has a better idea of how to do it.

gfgtdf:
So i did the wmlindent changes  myself in 2f234a8823f93db665ed31521aac449cc214e197 b21c110f304d82bf982203d7cb922aba08a17686 and commit between, so this can be free for discussion about  the wmlint changes again.

Pentarctagon:
I tried running it via GUI.pyw and it gave several warnings about files that ended with a non-zero indent, as well as eventually the error: `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x93 in position 0: invalid start byte`.

gfgtdf:
i reccomend you to rebase/redo this on latest master so that the diff works again

stevecotton:
> UnicodeDecodeError: 'utf-8' codec can't decode byte 0x93 in position 0: invalid start byte

That's likely to be the CVE-2018-1999023 tests. One of those files already has a .bin extension, the other two are .cfgs (I thought they were renamed, but it seems not).

CelticMinstrel:
> So i did the wmlindent changes myself

I assume you did it in a way that doesn't reindent the `#arg`/`#endarg`?

gfgtdf:
> I assume you did it in a way that doesn't reindent the #arg/#endarg?

You can just check that by looking at my commits...

cooljeanius:
> i reccomend you to rebase/redo this on latest master so that the diff works again

Actually never mind, I've changed my mind as to how I think I want to go about doing this: instead of doing a giant pull request that affects everything, I think I'm just going to make sure to run `wmllint` on just the files I'll be changing otherwise as part of the pull requests I'll submit for changes to individual campaigns

data/core/terrain-graphics/new-macros.cfg
cooljeanius: I think @gfgtdf got it when doing b21c110

data/core/terrain-graphics/new-macros.cfg
cooljeanius: well anyways the `wmlindent` parts are done now

data/campaigns/Descent_Into_Darkness/scenarios/06_Return_to_Parthyn.cfg
hexagonrecursion: Sounds too specific. How often would this rule be relevant?

data/campaigns/Heir_To_The_Throne/scenarios/23_Test_of_the_Clans.cfg
cooljeanius: > This is dead wrong, as this is Lua code rather than WML. It will result in an error at runtime.
> 
> @Elvish-Hunter Is it easy to prevent wmllint from touching embedded Lua code?

@hexagonrecursion has a fix that does that in #5470

PR #5165, opened 2020-09-25
Apply the deprecation standard to the new GUI2 API

CelticMinstrel:
This also removes backwards-compatibility.lua altogether.
It's no longer correct in the general case to add backwards compatibility code to this file, so removing it entirely avoids situations where compatibility code is incorrectly added there.

This is ready to merge pending testing; I need to actually build master before I can test it.


gfgtdf:
the comment on wesnoth.fire says explicitly that its not deprecated

CelticMinstrel:
Alright, I can change it to not be deprecated. No problem. I just thought from some of your comments elsewhere that we intended to deprecate it. I think your words were something like "it should always work without using wesnoth.fire"?

gfgtdf:
> . I think your words were something like "it should always work without using wesnoth.fire"?

well yes as long as you don't want wml variable substitution.  wesnoth.fire calls the wml_action  but applies wml variable substitution (which i think is rarely useful but some peopel seem to use it)

gfgtdf:
I also think we should convert all mainline content that uses the old gui2 api to the new gui2 api before we deprecate the old one.

CelticMinstrel:
I think deprecating first and converting afterwards is better than converting first and then deprecating. That said, doing both at the same time is even better. Are there uses of the old GUI2 API in master? I can convert them in this PR if you want me to.

Pentarctagon:
The travis log mentions a few other issues by the way.

CelticMinstrel:
That updates all the instances I could find outside of WorldConquest, although I'm not confident that all the conversions are correct. Do you @gfgtdf want to handle the uses of GUI2 API in WorldConquest?

@Pentarctagon I'll check Travis at some point but I'll probably try to test locally first.

Pentarctagon:
You'll need to run the WML unit tests locally, in that case.

data/scenario-test.cfg
```diff
@@ -1261,32 +1261,33 @@ local result = wesnoth.synchronize_choice(function()
...
             local ut = wesnoth.unit_types[v].__cfg
-            wesnoth.set_dialog_value(ut.name, "the_list", i, "the_label")
-            wesnoth.set_dialog_value(ut.image, "the_list", i, "the_icon")
+            dialog:find("the_list", i, "the_label").label = ut.name
+            dialog:find("the_list", i, "the_icon").label = ut.image
         end
-        wesnoth.set_dialog_value(2, "the_list")
+        dialog.the_list.selected_index = 2
         select()
+        -- TODO: How does this convert to the new API? It seems to be missing an argument...?
```
gfgtdf, line 1281: The window itself is also a widget sao this probably just sets the canvas of  the window so probably just:
```
dialog:set_canvas(2, {
            T.rectangle { x = 20, y = 20, w = 20, h = 20, fill_color= "0,0,255,255" },
            T.rectangle { x = 30, y = 30, w = 20, h = 20, fill_color = "255,0,0,255" }
})
```

CelticMinstrel:
@gfgtdf Thanks for converting the invest dialog. Besides that, it looks like the other uses in World Conquest are pretty simple, so I've got those conversions here; do they look correct?

PR #5167, opened 2020-09-25
Changes in new units.

Hejnewar:
These changes are supposed to make new units look natural (and balanced). For example Icemonax should have cold resistance and Gorer (Level 1 unit) shouldn't deal 26-2 damage even if with charge.

### DIFF
--- a/data/core/units/monsters/Ant.cfg
+++ b/data/core/units/monsters/Ant.cfg
@@ -7,24 +7,33 @@
     profile="portraits/monsters/giant_ant.png"
     {DEFENSE_ANIM "units/monsters/ant/ant-defend.png" "units/monsters/ant/ant.png" hiss.wav }
     hitpoints=22
-    movement_type=mountainfoot
+    movement_type=orcishfoot
     movement=4
     experience=50
-    level=1
-    alignment=chaotic
+    level=0
+    alignment=neutral
     advances_to=null
     {AMLA_DEFAULT}
-    cost=12
+    cost=7
     usage=fighter
     description= _ "Giant Ants are common in almost any environment, from caverns deep under the earth to the tops of tall mountains. Though normally not hostile, they can bite at close range."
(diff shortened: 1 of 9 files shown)


Pentarctagon:
For the future as well, more descriptive commit messages are helpful.  So what you have in the PR comment here would ideally be in the commit message itself.

soliton-:
A wmlindent run might be good before making a PR to avoid introducing indentation errors.

Pentarctagon:
I'm wondering if a wmlindent run would be better done as part of the release process actually, rather than leaving it to individual PRs.

CelticMinstrel:
I think it's better to run it for each commit so that the indentation _never_ becomes wrong. Putting it on a git hook might be nice, although that's a little dangerous since it sometimes produces incorrect results...

soliton-:
I think educating people about proper indentation and the tools that can help with it is better than periodically cleaning up afterwards.

Indeed a hook that would reject indentation errors would be neat. There are already ways to override wmlindent perhaps there would need to be more.

Pentarctagon:
Potentially it could work where wmllindent is run and then the build fails if `git status` shows files as being modified.

CelticMinstrel:
To be clear, a hook that simply runs wmlindent and applies the result is a little dangerous. If it instead rejected indentation errors, then that would be okay. That said, is there even a way to check in a hook to the repository?

Pentarctagon:
Not sure.  @Vultraz would you know?

soliton-:
I don't think you can install server-side git hooks. Presumably one would have to do some github actions stuff...

CelticMinstrel:
I wouldn't expect Vultraz to know that… but I don't expect it to be possible in any case. That said, anything that runs it and alerts if it produces changes could be useful; and if that's all it's doing it could totally be a GitHub Action or a Travis task or the like.

PR #5173, opened 2020-09-26
Wall terrain graphics (Xo*) work 

doofus-01:
trying to clean up some terrain graphics glitches, and possibly add some transitions


doofus-01:
Screenshot_20200927_
I started out trying to make these plants an overlay of the walls, like the bones in the catacombs, but it started to get too complicated, so it's just another wall variation.  But upside is that now the flowers ^Efm can be drawn on them.

Pentarctagon:
Those walls look amazing!

doofus-01:
Thanks.
I give up trying to understand what Travis is complaining about, so I'll probably merge this soon.

Pentarctagon:
It looks like a WML schema validation error: `Invalid value 'Xo*,100' in key 'type=' in tag [tile]`

doofus-01:
Yeah, I don't know what that means.  I thought it might have something to do with tabs being in there, but replacing those with spaces didn't change the error.

PR #5174, opened 2020-09-26
Remove the wesnoth-ai textdoman, move a few strings to wesnoth-lib (#4669)

### LINKED ISSUE #4669: i18n: Remove wesnoth-ai textdomain
Currently, Wesnoth ships with a whole textdomain dedicated to localized strings revolving around the game's AI engine and the AI demos/test scenarios. As of the current master branch tip (84ab0601173b2b359c3b075dfda73b7639660802), the string count for this textdomain is 378.

The problem is that other than a few strings near the start of the catalogue template, none of the other strings have any value for regular players since they are sourced from AI demos/test scenarios in `data/ai/micro_ais/` and `data/ai/scenarios/`. None of them will be displayed to people who aren't tinkering with WML or contributing to AI development as part of the Wesnoth dev team, and some of the language used in them is very technical in nature.

Since version 1.14.0, translated string count is used to determine whether a translation should even be listed in the Language menu to players by default, so it's not an option for translators to skip strings that dont' crop up in normal gameplay. The majority of the textdomain thus constitutes unnecessary workload for translators (especially so for new translations and abandoned translations in need of extensive review). In my experience as the former maintainer of the Spanish translation in 2008/2009, every single string counts when you're pressured to reach 100% string count before the first gold release of a new stable series.

Thus my proposal is to move the strings that actually have some relevance to the user to `wesnoth` or `wesnoth-lib` (ignoring for a moment the fact that both of these textdomains really ought to be either merged or shuffled around) and exclude the rest of the sources of the wesnoth-ai textdomain from wmlxgettext passes. This can be done in master and optionally backported to stable if it's deemed to be a low-risk change.

stevecotton:
Rename old AI to "Old Default AI (1.14 version)", as discussed in https://github.com/wesnoth/wesnoth/issues/4669#issuecomment-692699415

This whole textdomain is dedicated to localized strings revolving around the
game's AI engine and the AI demos/test scenarios, with around 370 strings in
it. Fixes #4669.

Other than a few strings near the start of the catalogue template, none of these
have any value for regular players since they are sourced from AI demos and
test scenarios in data/ai/micro_ais/ and data/ai/scenarios/. None of them will
be displayed to people who aren't tinkering with WML or contributing to AI
development as part of the Wesnoth dev team, and some of the language used in
them is very technical in nature.

This commit leaves the "#textdomain wesnoth-ai" lines in the AI demos, but
drops the .po and .pot catalogues for wesnoth-ai. The few strings that are
seen by regular players move to the wesnoth-lib textdomain.


stevecotton:
@Pentarctagon this is on the roadmap for 1.15.10, but it's ready now and I've just seen that our Catalan translator has spent time translating those 370 strings. Does it sound reasonable to put out a mail to the translator list this afternoon, saying that there's no need to translate this textdomain?

Pentarctagon:
I think that'd make sense.

Is there a reason to leave the `#textdomain wesnoth-ai` lines in the AI demos btw?

mattsc (approved):
Since I was asked to review: I only have a conceptual understanding of how translations work, so I can't comment on the implementation, but removing the textdomain is fine by me.

stevecotton:
> Is there a reason to leave the `#textdomain wesnoth-ai` lines in the AI demos btw?

When wmlxgettext encounters a file with no textdomain, it behaves as if it the file started with `#textdomain wesnoth`. At least, that's what it does when run from `scons pot-update`.

Issue #4455, opened 2019-10-12
Add a help page for gold/silver crowns and loyal marker


stevecotton:
To my knowledge, the convention is: 
* yellow crown for leaders 
* silver crown for (loyal?) units who cause you to lose the scenario
* bronze marker for loyal units

__the above originally posted by @sevu in #4452__

We shouldn't need to say "to an experienced dev's knowledge", we should have a page in the help's Gameplay section, as we do with "Orbs" and the hp/experience colouring.

jostephd:
What about ellipses? What does it means when a unit has the seven-pointed star ellipse? And what's the distinction between the two kinds of seven-pointed star ellipses?

![2019-10-12-213248_130x124_scrot](https://user-images.githubusercontent.com/24784687/66707919-d9870700-ed37-11e9-82d7-774eb83f831c.png)

(That's _Blackwater Port_ after `:unit type=Fog Clearer` on Konrad and Delfador. Ignore the orb colors.)

stevecotton:
Added to the roadmap for 1.15.6.

stevecotton:
Currently blocked by PR #5180. I want to add this to the same help page as the orbs, and am waiting for that PR's revision of the orbs page to merge.

CelticMinstrel:
This can be closed, right?

PR #5180, opened 2020-10-01
Orb refactor and ally orb colors during allies' turns (#1424)

stevecotton:
Note: the build will fail on MacOS, as the new orb_status files still need
adding to the MacOS build.

Instead of using the allied orb during their turn, the unmoved, moved and
partial orbs are shown. The player's own units will be shown with the moved orb
during the ally's turn, which matches the previous behavior. This is intended for
co-operative MP campaigns, where it will make it easier for players to discuss
possible moves. The UX is also visible in SP scenarios with allied sides, HttT's
first scenario is an easy place to see what it does.

Updated the documentation about orb colors. In these docs I've moved all of
the images to a single line above the bullet-pointed list, as otherwise the
layout looked bad once a line wrapped - the first line appeared
horizontally-aligned with the center of the image, the second started under the
image.

Added documentation about the delay in multiplayer games, as the purpose of this
orb-coloring feature is to assist discussion in multiplayer games.

Refactoring the orb-coloring code, adding new units/orb_status.hpp

Instead of creating lots of image::locators and calling a family of
similarly-named preferences functions, this refactor encapsulates that logic
and uses an enum.

An assert has been added to `unit_drawer`'s constructor. There are two callers
- the `display` class guards it with an explicit check before constructing
unit_drawer. The `game_display` class doesn't have an explicit check, but it's
clear that other code in that class assumes the teams are already valid (and
would crash if they weren't).

Other issues connected with this PR
---

Issue #5155's proposed orb_status::disengaged has been tested with this, and
builds on the refactor in the first of this PR's two commits.

The orbs' help page suggests that there should be documentation about the crowns
too, however I want to separate any work on that from this PR. Issue #4455 tracks
the info about crowns.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
  ### Language and i18n
  ### Lua API
  ### Multiplayer
+   * During allies' turns, use orb colors to show which ones can still move (issue #1424). Also enabled for allied AI sides in singleplayer.
  ### Terrain
  ### Units
  ### User interface
--- a/data/core/help.cfg
+++ b/data/core/help.cfg
@@ -277,7 +277,11 @@ To see where the enemy can move to during their next turn, press Ctrl-v or Cmd-v
         title= _ "Shroud and Fog of War"
         text= _ "In some scenarios, parts of the map will be hidden from you. There are two mechanisms that can be used separately or together. The <italic>text='shroud'</italic> hides both the terrain and any units at a location. However, once it is cleared, you can always see that location. The <italic>text='fog of war'</italic> only hides units and ownership of villages (other than by you or your allies). The fog of war is cleared temporarily when you have units nearby, but returns when they leave. Both the shroud and the fog of war are cleared by units. Each unit clears locations adjacent to those within one turn’s move (ignoring zones of control and enemy units).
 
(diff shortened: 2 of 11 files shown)


Pentarctagon:
Which color did you end up going with for orb_status::disengaged?

stevecotton:
> Which color did you end up going with for orb_status::disengaged?

I left it out of this PR - the refactor does the groundwork, but the extra bits for the new color are in [whichever commit most recently auto-linked itself to issue #5155].

stevecotton:
@gfgtdf as you're working on multiplayer advancements, please would you test and review this?

The build failure is merely because it adds a new file, which needs someone to add to Xcode. I'm intending to merge as-is, complete with that failure.

stevecotton:
@Hejnewar, please would you comment on the Xcode workflow? Is in easier for you if I merge and then you update afterwards?

As discussed on IRC, I'll add Vultraz to the reviews - going to push to fix the merge conflict in the changelog first.

soliton-:
I assume you meant to talk to our mac packager @hrubymar10. In general it's ok to commit new files and let people with access to build systems where it's not just adding the new files to a list to fix those later.

stevecotton:
@hrubymar10 if you've time to update the Xcode build before release then I think this could be merged, otherwise I'll postpone it until after 1.15.6 has released.

hrubymar10:
> @hrubymar10 if you've time to update the Xcode build before release then I think this could be merged, otherwise I'll postpone it until after 1.15.6 has released.
> 
> Edit: ah, 1.15.6 has already been tagged. Definitely going to postpone then `:D`

I am sorry but for some reason I got email just 5 hours ago so thi PR didn't make it to 1.15.6 . 😔 I believe that this could be merged now @stevecotton and I'll update Xcode project for it once it will be in master .

PR #5181, opened 2020-10-01
Changes in new monster units regarding experience and advancements

IceTyp:
Giant Stoat (L0), Giant Ant (L0), Gorer (L1), Cave Bear (L2) and Great Icemonax (L2) have no advancements.
I changed their experience to 50 times level (25 for L0) to match with similar units in mainline.

Tusklet and Gorer are of the same unit line. So I gave Tusklet the advancement to Gorer.

There are some things I am not sure about:
Should Giant Ant (L0) advance to Fire Ant (L1)? I don't think so because they appear in different faunas corresponding to the [forum post](https://forums.wesnoth.org/viewtopic.php?f=9&t=52894) of doofus-01.

Icemonax (L0) and Great Icemonax (L2) are obviously of the same unit line. But their levels differ by two. Hence, an advancement of Icemonax to Great Icemonax would not make much sense at this time.
Will there be a L1 unit between those or will one of them be changed to L1?

### DIFF
--- a/data/core/about.cfg
+++ b/data/core/about.cfg
@@ -1305,6 +1305,9 @@
         name = "holius-fr"
     [/entry]
     [entry]
+        name = "IceTyp"
+    [/entry]
+    [entry]
         name = "ilyapopov"
     [/entry]
     [entry]
--- a/data/core/units/monsters/Ant.cfg
+++ b/data/core/units/monsters/Ant.cfg
@@ -9,7 +9,7 @@
     hitpoints=22
     movement_type=orcishfoot
     movement=4
-    experience=50
+    experience=25
     level=0
     alignment=neutral
     advances_to=null
--- a/data/core/units/monsters/Bear.cfg
+++ b/data/core/units/monsters/Bear.cfg
@@ -10,7 +10,7 @@
     hitpoints=72
     movement_type=mountainfoot
     movement=6
-    experience=80
+    experience=100
     level=2
     alignment=chaotic
     advances_to=null
--- a/data/core/units/monsters/Giant_Stoat.cfg
+++ b/data/core/units/monsters/Giant_Stoat.cfg
@@ -10,7 +10,7 @@
     hitpoints=23
     movement_type=elusivefoot
     movement=6
-    experience=50
+    experience=25
     level=0
     alignment=chaotic
     advances_to=null
--- a/data/core/units/monsters/Icemonax_Greater.cfg
+++ b/data/core/units/monsters/Icemonax_Greater.cfg
@@ -23,7 +23,7 @@
     hitpoints=56
     movement_type=largefoot
     movement=5
-    experience=80
+    experience=100
     level=2
     alignment=neutral
     advances_to=null
--- a/data/core/units/monsters/Tusker_Gorer.cfg
+++ b/data/core/units/monsters/Tusker_Gorer.cfg
@@ -23,7 +23,7 @@
     hitpoints=35
     movement_type=rodentfoot
     movement=7
-    experience=150
+    experience=50
     {AMLA_DEFAULT}
     level=1
     alignment=neutral
--- a/data/core/units/monsters/Tusker_Tusklet.cfg
+++ b/data/core/units/monsters/Tusker_Tusklet.cfg
@@ -24,10 +24,9 @@
     movement_type=rodentfoot
     movement=5
     experience=26
-    {AMLA_DEFAULT}
     level=0
     alignment=neutral
-    advances_to=null
+    advances_to=Gorer
     undead_variation=mounted
     cost=10
     usage=fighter


Pentarctagon:
@Hejnewar Thoughts?

CelticMinstrel:
Just double-checking but… these _do_ all have `{AMLA_DEFAULT}`, right? Other than the tusklet obviously.

Hejnewar:
Looks fine to me. This shouldn't affect anything in meaningful way.

Pentarctagon:
@CelticMinstrel aside from the Tusklet, yes.

@doofus-01 For

> Icemonax (L0) and Great Icemonax (L2) are obviously of the same unit line. But their levels differ by two. Hence, an advancement of Icemonax to Great Icemonax would not make much sense at this time.
> Will there be a L1 unit between those or will one of them be changed to L1?

Were you planning on making a L1 unit, or are they meant to be independent (one doesn't advance to the other)?

doofus-01:
I was thinking the L0 wouldn't advance to the "Great" version, but if that really disappoints people, it can always be reconsidered.  Similarly, I didn't envision the ants advancing one to the other.

The concept of monsters "upgrading" in the middle of a fight seems a bit odd, or at least more of a stretch than the sudden boost in social status and resources implied for the military/humanoid units.  I guess it already happens for bats, but no reason to highlight it further.  Maybe there can be some non-repeating AMLA that distinguishes the veteran monsters?

Hejnewar:
Wesnoth is inconsistent on that field. There are 4 monster / animal units that advance and fit each other (bat / falcon / mudcrawler / scorpion) and only one that doesn't (wolf) every other monster / animal doesn't have anything it could advance to or from. 

I changed ant from L1 to L0 because of statistics, it just didn't fit in L1, it cost there would be too low. I would be against Ants leveling to each other too, after all ant worker will always be ant worker, it won't grow up to be big ant warrior. For animals they mostly level normally if they have anything to level to as showcased above.

doofus-01:
Falcon was and bat is a member of a mainline MP faction, so there's that aspect.  We could pretty easily drop the falcon-> elder falcon advancement.  As for the mudcrawler, who knows what makes sense for an animated mud glob.  If the wolf should advance because of the riders, that's fine, it wouldn't be the first compromise of concepts made in this game.

ElderoZ:
It does seem natural that monster units like giant rats, giant ants, fire guardians or scorpions don't advance and are stuck at one level. I think it does make sense for monsters, they have always been like that but besides that it does make them feel primitive, and i never really thought that each type of a monster unit really needs a bigger, stronger version that would be higher level. It does take some from them being "primitive".

A unit that is low level, like level 1 or lvl 0 and cant advance does feel primitive because its so low level and can't become better/stronger with experience. This is kinda a trait for all monsters. There are exceptions like mudcrawlers, but these are magical creations, and advancing from lvl 0 to lvl 1 max still afterall feels primitive, from lvl 1 to lvl 2 not so much.

doofus said that it would be weird for a monster to change during fight, and it does indeed feel a bit weird, though for me its probably more because I think that a primitive creature like a small monster shouldn't really change.

IceTyp:
There is no unit, where advancements during a battle make sense from a logical perspective. How should a Peasant with enough experience find a sword and a bow and turn into a Bowman in the middle of a battle?
But advancements are a **critical aspect of Wesnoth** and game mechanics do not need to be realistic. 

For me, unit trees consisting of many units show that this unit is more adaptive. Most of those units have some form of intelligence and they can speak and work together. 

Monsters on the other hand are "dumb". High level monsters (e.g. Giant Spider, Yeti) have no access to advancements. 
There are some monsters with advancements. But most of those unit lines start with level 0 (e.g. bats, falcons, scorpions) or consist only of two units. So unit trees of monsters have much less variety.

I would not give each unit line advancements. But I would allow advancements where it is already available.

At least for consistency, I would give advancements not only to a few monsters like bats, scorpions and Mudcrawlers but also to wolves, the Tusklet, and the Icemonax.

ElderoZ:
Advancements are not really a critical element of monster units though.

Monsters not having advancements is pretty consistent in itself.

I think for an unit to classify as a "monster" unit it must be:
- Generally it is used in campaigns like a monster unit, which usually means it being units without a leader that can be found in caves or some "wild" places in scenarios, that usually are in guard mode and move and attack only when they can reach an unit.
- It doesn't really belong to any faction
- It appears more primitive than other units, which usually means its mindless, and can't really advance, if its stuck at some level then it does appear more primitive than units that can advance, it seems like a good trait for a monster to look more like a monster.

Hejnewar:
It would be kinda weird for me if the same unit in faction could advance but without it, it couldn't. Primitivity is not really what I look at when I play. They are just units for me. Advancing is too unreal anyway. 

Some weird reasons:
It unnecessarily expands "Monsters" section on units.
I can't force feed monster unit in order to get more xp from scenario.

doofus-01:
ElderoZ and Hejnewar stated some of my thoughts, especially the parts about "monsters" not playing the same role as some other unit like a Spearman (or a Bat, because of MP), and unnecessarily expanding the "monsters" section.  Time and resources are limited, I don't want to waste them on monster levels at the expense of monster variety/quality.

A peasant getting a field promotion to a spearman isn't all that outlandish.  If it seems impossible for a given scenario, that is the fault of the scenario author.

CelticMinstrel:
I don't see any particular reason not to make monsters advance in those situations where there _is_ a bigger or stronger version of the same monster that has a higher level. I also don't see any reason to force monsters to have an advancement tree just for the sake of having one.

If it were up to me, I'd make wolves advance. I'm not as sure about the icemonax as it currently stands though, since skipping a level in an advancement is pretty unprecedented. I suppose one of the two existing units could always change levels so that they're only one level apart…

More importantly, I think arguments from realism are more or less specious. Wesnoth is a game that's already highly unrealistic, after all. Most of the game mechanics aren't very realistic, so I don't think there's much point in shoehorning them into realism. That said, I don't think monsters advancing is _all_ that unrealistic, either - in most cases these are a case of juveniles reaching maturity. Sure, that's not the sort of thing you'd expect to happen in the middle of a battlefield, but nor is a peasant somehow finding a bow or spear somewhere and being promoted to a foot soldier. Normally such a promotion would happen _between_ battles, not in the middle of a battle.

Pentarctagon:
As far as the two questions then:

> Should Giant Ant (L0) advance to Fire Ant (L1)? I don't think so because they appear in different faunas corresponding to the forum post of doofus-01.

These should remain separate unit lines.

> Icemonax (L0) and Great Icemonax (L2) are obviously of the same unit line. But their levels differ by two. Hence, an advancement of Icemonax to Great Icemonax would not make much sense at this time.
> Will there be a L1 unit between those or will one of them be changed to L1?

These two can be left as-is as well.  If a L1 intermediate unit gets added at some point, then it would make sense to have them advance to each other, but for right now it's not really a problem that they don't either.

cooljeanius:
> > Icemonax (L0) and Great Icemonax (L2) are obviously of the same unit line. But their levels differ by two. Hence, an advancement of Icemonax to Great Icemonax would not make much sense at this time.
> > Will there be a L1 unit between those or will one of them be changed to L1?
> 
> These two can be left as-is as well. If a L1 intermediate unit gets added at some point, then it would make sense to have them advance to each other, but for right now it's not really a problem that they don't either.

I would very much appreciate a L1 intermediate icemonax for one of my campaigns, if that's possible.

PR #5182, opened 2020-10-02
Use the new animals in the "animals" micro-ai demo

stevecotton:
Several new animals were added in PR #5039's cbad6f42492be5fcbf8c29a3d10cbc43598ba32c.

The micro-ai demo can be accessed by starting Wesnoth with the command-line
arguments `--test animals`. The scenario uses some stand-ins for animals that
weren't in mainline; the new bears and tuskers fit straight in. For the sheep,
the icemonax's graphics are a bit more sheep-shaped than the troll whelp's.

This demo doesn't need to be balanced, the changed unit stats are no problem.

UtBS's dustboks aren't in core, otherwise I'd use them as the deer stand-in.

@doofus-01 thanks for these new resources. Are the unit ids final (as in, should
I postpone this change in case they get renamed)?

### DIFF
--- a/data/ai/micro_ais/scenarios/animals.cfg
+++ b/data/ai/micro_ais/scenarios/animals.cfg
@@ -3,7 +3,7 @@
 #define ANIMAL_AI_DESCRIPTIONS1
     _"<u>General</u>: These AIs are set up to simulate (to some extent) how these animals behave in real life. This includes that they are animals, meaning that they are not super smart. As an example, the wolves generally hunt in a pack, but are easily distracted by prey coming into range. They are also decent, but not great at cornering deer. For the most part, this is intentional.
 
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
> UtBS's dustboks aren't in core, otherwise I'd use them as the deer stand-in.

It's orthogonal to this PR, but I think they should be moved to core. The UtBS animals could maybe even be moved separately from the Quenoth elves (I know some people wanted to hold off on moving the Quenoth until they're more finalized).

mattsc (approved):
Looks good to me, thanks for doing this.  For the record, while it is true that the scenario is not and does not need to be balanced, there are some subtleties that need to be taken into account.  For example, the herding does not work very well if the "sheep" are faster than the "sheep dogs".  The same is true for the tuskers and tusklets.  None of this is an issue here, I am just stating it in case more changes will be made in the future.

doofus-01 (approved):
It looks reasonable to me, thanks.  I don't anticipate changing their IDs.  There's some question about the advancements, but I don't think that matters here.

PR #5186, opened 2020-10-04
[Proposal] Extended Outro dialog with campaign-specific credits

Vultraz:
An idea to improve the credits experience. This doesn't do anything to deal with the larger issue over the massive surface in the credits dialog, but I think it presents the campaign-specific credits better.

In short, it extends the Outro sequence past the "The End" text to include the campaign name followed by the campaign-specific credits with each section split into groups of no more than 5 names at a time. The long-form credits are **no longer shown** after.

For instance:
image

### DIFF
--- a/src/gui/dialogs/outro.cpp
+++ b/src/gui/dialogs/outro.cpp
@@ -50,13 +50,14 @@ outro::outro(const game_classification& info)
 
 	// We only show the end text and the title if credits were turned off
 	if(info.end_credits) {
-		const auto& credits = about::get_credits_data();
-		const auto campaign_credits = std::find_if(credits.begin(), credits.end(),
-			[&info](const about::credits_group& group) { return group.id == info.campaign; });
+		for(const auto& about : about::get_campaign_credits(info.campaign)->sections) {
+			if(about.names.empty()) {
+				continue;
+			}
 
-		for(const about::credits_group::about_group& about : campaign_credits->sections) {
 			// Split the names into chunks of 5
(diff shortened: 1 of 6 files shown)


shikadiqueen:
I feel like I've seen this somewhere before. Perhaps in one particular add-on.

Is this a good idea in the general case, though? Credits for any single campaign can be arbitrarily long.

Pentarctagon:
@nemaara as well - any opinion on this?

nemaara:
Seems okay to me, as long as it's only for the campaign credits and not the entire credits list that follows after that.

src/gui/dialogs/outro.cpp
```diff
@@ -31,26 +32,62 @@ namespace dialogs
...
+			for(std::size_t i = 0; i < num_chunks; ++i) {
+				std::stringstream ss;
+
+				// Only include section title on first chunk
+				if(i == 0) {
+					ss << about.title << "\n";
+				}
+
+				for(std::size_t k = i * chunk_size; k < std::min<unsigned>((i + 1) * chunk_size, num_names); ++k) {
+					ss << "\n<span size='xx-small'>" << about.names[k].first << "</span>";
```
soliton-, line 72: Looks like there is a `font::escape_text()` call missing here.
```suggestion
					ss << "\n<span size='xx-small'>" << font::escape_text(about.names[k].first) << "</span>";
```

Vultraz: Added this to the central handling in about.cpp.

PR #5188, opened 2020-10-04
Refactor translatable strings for the SotA zombie recruit dialog

CelticMinstrel:
I think it's easier for translators if each string is actually just one string. Using "split" on a translatable string just unnecessarily complicates things.

I also fixed a bug in `[set_variable]` which was referenced by a comment in the SotA zombie-utils file.

### DIFF
--- a/data/campaigns/Secrets_of_the_Ancients/utils/zombie-utils.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/utils/zombie-utils.cfg
@@ -13,81 +13,97 @@
 #define ZOMBIE_INIT
     [event]
         name=prestart
+        #wmllint: markcheck off
+        # TODO: Most of the sota_variation strings could probably reuse race names from core and just add the parentheses in the Lua code
+        # sota_variation will be used in the zombie recruit dialog
         [set_variables]
             name=zombies
-            [split]
-                list="SotA Flying Corpse_Bat,
-                SotA Walking Corpse_Rat,
-                SotA Walking Corpse_Wolf,
-                SotA Walking Corpse_Human,
-                SotA Walking Corpse_Mounted,
-                SotA Walking Corpse_Wolf Rider,
-                SotA Swimming Corpse,
-                SotA Walking Corpse_Saurian,
-                SotA Walking Corpse_Wose,
-                SotA Walking Corpse_Goblin,
-                SotA Flying Corpse_Gryphon,
-                SotA Walking Corpse_Spider,
-                SotA Walking Corpse_Troll,
-                SotA Walking Corpse_Dwarf"
-                key=type
-                separator=","
-            [/split]
+            [value]
+                type=SotA Flying Corpse_Bat
+                sota_variation= _ "(Bat)"
+                allow_recruit_message= _ "You can now raise bat corpses!"
+                allow_recruit=no
+            [/value]
+            [value]
+                type=SotA Walking Corpse_Rat
+                sota_variation= _ "(Rat)"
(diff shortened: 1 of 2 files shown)


data/campaigns/Secrets_of_the_Ancients/utils/zombie-utils.cfg
```diff
@@ -104,47 +120,19 @@
...
-    # Remove the trailing comma from the string.
-    [set_variable]
-        name={VARIABLE}
-        value="$(substring( '${VARIABLE}', 0, (length('${VARIABLE}')-1) ))"
-    [/set_variable]
 
     # This *should* work, but doesn't as of 1.14.5. It causes a lua error because of
     # concatenation of a boolean, presumably because the strings are all either
     # "yes" or "no". If that changes, this code can be uncommented, and the code
     # above removed.
```
stevecotton, line 127: Out of date comment

data/campaigns/Secrets_of_the_Ancients/utils/zombie-utils.cfg
```diff
@@ -13,81 +13,97 @@
...
+            [/value]
+            [value]
+                type=SotA Walking Corpse_Wolf Rider
+                sota_variation= _ "(Wolf Rider)"
+                allow_recruit_message= _ "You can now raise wolf rider corpses!"
+                allow_recruit=no
+            [/value]
+            [value]
+                type=SotA Swimming Corpse
+                # po: The ^ at the end means that this string is not used in English. Since there is only one swimmer corpse, it would be redundant to label it in English. You may translate this string (without the ^) if it makes more sense in your language to do so.
```
stevecotton, line 58: While what the text is saying makes sense to me, it doesn't say where this will be used, and I had to playtest (in debug mode, for speed) to find out how these strings fit in. I still don't understand why these are separate strings, instead of just being included in the `[unit_type]name=` string.

Having a string translated to an empty string is bad for the translation stats - it will look like an untranslated string, and that's what gettext.wesnoth.org and poedit will treat it as.

Suggestion, add something like this:
These sota_variation strings are shown in the recruit dialog, appended to the name of the unit type. For example, "Flying Corpse (Bat)", "Walking Corpse (Human)" and "Walking Corpse (Rat)". There is only one "Swimming Corpse".

stevecotton: I think changing it to "(Merfolk)" would fit, both from translation and from UX perspective. The player is definitely going to kill some Merfolk in S10, and the first scenario with Nagas is S11.

Just noticed a plot hole: when Ras-Tabahn kills Merfolk in S10, Ardonna gets the knowledge before they meet up again. Haven't playtested to check that, I'm just assuming it from the mechanics.

CelticMinstrel: This isn't really related, but I think it'd be nice if merfolk and naga had separate walking corpse variations. After all, the merfolk are more fish-like while nagas are more snake-like - they're not all that similar.

ivanovic:
> I think it's easier for translators if each string is actually just one string. Using "split" on a translatable string just unnecessarily complicates things.
> 
> I also fixed a bug in `[set_variable]` which was referenced by a comment in the SotA zombie-utils file.

Yes, a string should be as complete for a translator as possible since languages are strange like that. Please do not try to be smart to save strings since translators will hate you for making their life even more miserable. Better to have the option to translate a very similar string 10 times differently if required (the reason why we e.g. have the unit descriptions as female and male version).

Issue #5067, opened 2020-08-10
Revamp the Ghoul line portraits


Pentarctagon:
Taken from the [1.15 Roadmap](https://wiki.wesnoth.org/1.15_Roadmap#1.15.9_.2801.2F16.2F2021.29).

doofus-01:
We should consider removing this from the roadmap.  It's not the sort of thing that should be done late in the cycle, and I don't see getting this done soon.

Pentarctagon:
Alright, removed.

PR #5189, opened 2020-10-04
SotA - revisions to units and possibly terrain

doofus-01:
This PR is mostly for some preliminary work to ghoul updates (#5067), plus whatever else I stumble across in SotA.  No intention to revise balance or text, except as it relates to the ghouls.

Revision to the undead sea captain was because I didn't see a skeleton matching the portrait, the "fresh body" concept, or the opening dialog of the _Blackwater_ scenario.  Maybe the zombie captain can turn into a skeleton captain later, if it really affects the later scenarios.

The intro of _Becalmed_ scenario contained the origin story of the ghouls, so I revised that.  The images are placeholders, but it is currently too verbose to just show the book image.

Other things under consideration:
 - ~~Touch-up to the ship graphics (shadows, mostly)~~ Second thoughts on this.
 - Move the graveyard fence to either a fence or wall terrain - **Done**
 - Fix the terrain codes so they don't use "y", which is reserved for UMC. - **Done**


doofus-01:
This is the first commit for the proposed iron fence.  The graphics aren't great but will be improved once the sizing is finalized - just want to make sure it looks like a real barrier and not a child's playpen.
Screenshot_20201010_

stevecotton:
> just want to make sure it looks like a real barrier and not a child's playpen

Sorry to be negative, but for me it passes into the realm of "that fence is the right size", but then "if it's like UK burial customs, those gates are too narrow for a coffin with pallbearers on each side".

doofus-01:
The gates are just the existing rusty gates.

data/campaigns/Secrets_of_the_Ancients/scenarios/02_Dark_Business.cfg
```diff
@@ -593,13 +593,13 @@ of Healing"
     [/event]
 
     # This macro animates guards moving out of the guardhouse. Each guard moves to the
-    # nearest unoccupied, walkable, hex, without an ^Nhyr overlay. ^Nhyr ("Not here") marks
+    # nearest unoccupied, walkable, hex, without an ^NohS overlay. ^Nhyr ("Not here") marks
```
stevecotton, line 596: There's were two `Nhyr`'s on this line, still need to change one.

doofus-01:
I'd posted this earlier, but it seems to have been removed (maybe by another annoying bot?), it's an update of the iron fence and how it interacts with the existing gates.  
Screenshot_20201011_

Vultraz:
This looks *excellent*! Perhaps it should be a core terrain?

doofus-01:
Thanks.  It's implemented in the PR as a core terrain, though intended to replace the SotA graveyard fence overlay image.

Among other things, it needs a little more testing and probably correction to mesh well with the other likely terrains.

doofus-01:
The fences work with most reasonable neighboring terrains now.  There are a few glitches still, but I think they will be difficult to fix without making very specific hack rules that will be difficult to maintain.  One that is a layering issue, possibly fixable, is the cave wall overlapping the post in the bottom right corner.  The tent is an issue that I don't see a way to fix.
Screenshot_20201017_

doofus-01:
That last update just exchanged the big graveyard fence for the new *^Eqf terrain.  That by itself isn't an improvement, but at least now the map layout can be altered.

nemaara:
So this says it needs my review but I don't have a great way to test it. Realistically it seems like since this is mostly graphics updates probably doofus knows what he's doing and it seems fine to me.

doofus-01:
@nemaara - I think it needs your review for the "lore change" regarding ghouls origin in the intro of Scenario 4.  As it is, I think it's a first draft.  I'll merge this so it's easier to test, it will be revised based on feedback.

PR #5196, opened 2020-10-08
Removed Registered Users Only game config option (resolves #4980)

### LINKED ISSUE #4980: "Registered Users Only" checkbox is checked with no option to clear it on lan servers with no user registration mechanism
1.14.13

website for mac, flatpack for linux

not custom

 - What OS are you running?
linux and mac

### Describe the bug
When hosting a lan server on linux box, the linux and mac clients are not able to uncheck the "Registered Users Only" option. Even though the server does not support registered users

### To Reproduce
1. From a mac or linux client, log into the official servers and create a game with the option "Registered Users Only" checked.
2. leave the game
3. log out
4. Start a locally hosted lan server
5. log into the local lan server from mac or linux client
6. Creating a game will have the  "Registered Users Only" checkbox checked on by default, with no way to uncheck it.

### Expected behavior
If the client is connecting to a local lan server that does not support registered users, the "Registered Users Only" checkbox should not be checked by default, even if the last game created by the client had the option turned on.

Vultraz:
Closes #4980. As @Pentarctagon suggested. Opening a PR to confirm removing that bit from wesnothd is ok.

### DIFF
--- a/data/gui/window/mp_create_game.cfg
+++ b/data/gui/window/mp_create_game.cfg
@@ -933,10 +933,6 @@
 {_GUI_MP_CREATE_MAIN_TOGGLE "observers" _"Observers" _"Allow users who are not playing to watch the game"}
 #enddef
 
-#define _GUI_MP_CREATE_REGISTERED_TOGGLE
-{_GUI_MP_CREATE_MAIN_TOGGLE "registered_users" _"Registered Users Only" _"Allow only registered users to join the game"}
-#enddef
-
 #define _GUI_MP_CREATE_SYNC_TOGGLE
 {_GUI_MP_CREATE_MAIN_TOGGLE "strict_sync" _"Strict Synchronization" _"Stricter checks for out-of-sync errors"}
 #enddef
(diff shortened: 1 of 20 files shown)


CelticMinstrel:
I don't think this feature should be removed. Hiding it if the server requires it by default is a good idea, sure, but I don't like removing it altogether.

Pentarctagon:
@CelticMinstrel Why should the feature be kept, then?

Pentarctagon:
(Also, Soliton is the wesnothd maintainer)

soliton-:
I don't think the option is particularly useful anymore. We're unlikely to go back to a mixed registration setup on wesnoth.org and other servers are unlikely to have a mixed registration setup to begin with.

soliton- (approved):

CelticMinstrel:
> Why should the feature be kept, then?

Because the official server is not the only server. Removing an option because the official server doesn't need it is silly. While it's true that _most_ unofficial servers would probably be small affairs not requiring registration, who are we to constrain that possibility? The forum registration handler is part of the Wesnoth source code after all; anyone could use it if they wanted to.

Pentarctagon:
If the only reason to keep a feature is that there's a near, but not quite, 0% chance it's still useful, then I don't consider that to be a strong argument for keeping something that's actively causing a problem for users we do actually know about.

PR #5198, opened 2020-10-09
Thread-safe logging

CelticMinstrel:
The draft #5176 included a wish for thread-safe logging, which I think would be a good idea in general, so I tried implementing the thing I was describing in that thread.

This works in the most basic sense, which is to say it compiles on my computer with no errors. I have not yet done any testing whatsoever, not even just running the game to see if logging even slightly works. Thus, the draft status of this PR.

### DIFF
--- a/src/formatter.hpp
+++ b/src/formatter.hpp
@@ -67,6 +67,31 @@ public:
 	{
 		return stream_.str();
 	}
+	
+	// Support manipulators
+    formatter& operator<<(std::ostream&(*fn)(std::ostream&)) &
+	{
+        fn(stream_);
+		return *this;
+    }
+	
+    formatter&& operator<<(std::ostream&(*fn)(std::ostream&)) &&
+	{
+		fn(stream_);
+		return std::move(*this);
+    }
+	
+    formatter& operator<<(std::ios_base&(*fn)(std::ios_base&)) &
+	{
+		fn(stream_);
+		return *this;
+    }
+
+    formatter&& operator<<(std::ios_base&(*fn)(std::ios_base&)) &&
+	{
+		fn(stream_);
+		return std::move(*this);
+    }
 
 private:
 	std::ostringstream stream_;
--- a/src/game_config_manager.cpp
+++ b/src/game_config_manager.cpp
@@ -134,7 +134,7 @@ void game_config_manager::load_game_config_with_loadscreen(FORCE_RELOAD_CONFIG f
 	boost::optional<std::set<std::string>> active_addons)
 {
 	if (!lg::info().dont_log(log_config)) {
-		auto& out = lg::info()(log_config);
+		auto out = formatter();
 		out << "load_game_config: defines:";
 		for(const auto& pair : cache_.get_preproc_map()) {
 			out << pair.first << ",";
@@ -149,7 +149,8 @@ void game_config_manager::load_game_config_with_loadscreen(FORCE_RELOAD_CONFIG f
 			out << "\n Everything:";
 		}
 		out << "\n";
-	}
+		FORCE_LOG_TO(lg::info(), log_config) << out.str();
+	} 
 
 
 	game_config::scoped_preproc_define debug_mode("DEBUG_MODE",
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -26,6 +26,7 @@
 #include <map>
 #include <sstream>
 #include <ctime>
+#include <mutex>
 
 #include "global.hpp"
 
@@ -44,6 +45,7 @@ static std::ostream null_ostream(new null_streambuf);
 static int indent = 0;
 static bool timestamp = true;
 static bool precise_timestamp = false;
+static std::mutex log_mutex;
 
 static boost::posix_time::time_facet facet("%Y%m%d %H:%M:%S%F ");
 static std::ostream *output_stream = nullptr;
@@ -207,7 +209,7 @@ static void print_precise_timestamp(std::ostream & out) noexcept
 	} catch(...) {}
 }
 
-std::ostream &logger::operator()(const log_domain& domain, bool show_names, bool do_indent) const
+log_in_progress logger::operator()(const log_domain& domain, bool show_names, bool do_indent) const
 {
 	if (severity_ > domain.domain_->second) {
 		return null_ostream;
@@ -218,33 +220,58 @@ std::ostream &logger::operator()(const log_domain& domain, bool show_names, bool
 			std::cerr << ss.str() << std::endl;
 			strict_threw_ = true;
 		}
-		std::ostream& stream = output();
+		log_in_progress stream = output();
 		if(do_indent) {
-			for(int i = 0; i != indent; ++i)
-				stream << "  ";
-			}
+			stream.set_indent(indent);
+		}
 		if (timestamp) {
-			if(precise_timestamp) {
-				print_precise_timestamp(stream);
-			} else {
-				stream << get_timestamp(std::time(nullptr));
-			}
+			stream.enable_timestamp();
 		}
 		if (show_names) {
-			stream << name_ << ' ' << domain.domain_->first << ": ";
+			stream.set_prefix(formatter() << name_ << ' ' << domain.domain_->first << ": ");
 		}
 		return stream;
 	}
 }
 
-void scope_logger::do_log_entry(const log_domain& domain, const std::string& str) noexcept
+log_in_progress::log_in_progress(std::ostream& stream)
+	: stream_(stream)
+{}
+
+void log_in_progress::operator|(formatter&& message)
 {
-	output_ = &debug()(domain, false, true);
+	std::lock_guard<std::mutex> lock(log_mutex);
+	for(int i = 0; i < indent; ++i)
+		stream_ << "  ";
+	if(timestamp_) {
+		if(precise_timestamp) {
+			print_precise_timestamp(stream_);
+		} else {
+			stream_ << get_timestamp(std::time(nullptr));
+		}
+	}
+	stream_ << prefix_ << message.str();
+}
+
+void log_in_progress::set_indent(int level) {
+	indent_ = level;
+}
+
+void log_in_progress::enable_timestamp() {
+	timestamp_ = true;
+}
+
+void log_in_progress::set_prefix(const std::string& prefix) {
+	prefix_ = prefix;
+}
+
+void scope_logger::do_log_entry(const std::string& str) noexcept
+{
 	str_ = str;
 	try {
 		ticks_ = boost::posix_time::microsec_clock::local_time();
 	} catch(...) {}
-	(*output_) << "{ BEGIN: " << str_ << "\n";
+	debug()(domain_, false, true) | formatter() << "{ BEGIN: " << str_ << "\n";
 	++indent;
 }
 
@@ -255,15 +282,10 @@ void scope_logger::do_log_exit() noexcept
 		ticks = (boost::posix_time::microsec_clock::local_time() - ticks_).total_milliseconds();
 	} catch(...) {}
 	--indent;
-	do_indent();
-	if (timestamp) (*output_) << get_timestamp(std::time(nullptr));
-	(*output_) << "} END: " << str_ << " (took " << ticks << "ms)\n";
-}
-
-void scope_logger::do_indent() const
-{
-	for(int i = 0; i != indent; ++i)
-		(*output_) << "  ";
+	auto output = debug()(domain_, false, true);
+	output.set_indent(indent);
+	if(timestamp) output.enable_timestamp();
+	output | formatter() << "} END: " << str_ << " (took " << ticks << "ms)\n";
 }
 
 std::stringstream& wml_error()
--- a/src/log.hpp
+++ b/src/log.hpp
@@ -58,6 +58,7 @@
 #include <utility>
 
 #include <boost/date_time/posix_time/posix_time_types.hpp>
+#include "formatter.hpp"
 
 using boost::posix_time::ptime;
 
@@ -112,12 +113,32 @@ void set_strict_severity(int severity);
 void set_strict_severity(const logger &lg);
 bool broke_strict();
 
+// A little "magic" to surround the logging operation in a mutex.
+// This works by capturing the output first to a stringstream formatter, then
+// locking a mutex and dumping it to the stream all in one go.
+// By doing this we can avoid rare deadlocks if a function whose output is streamed
+// calls logging of its own.
+// We overload operator| only because it has lower precedence than operator<<
+// Any other lower-precedence operator would have worked just as well.
+class log_in_progress {
+	std::ostream& stream_;
+	int indent_ = 0;
+	bool timestamp_ = false;
+	std::string prefix_;
+public:
+	log_in_progress(std::ostream& stream);
+	void operator|(formatter&& message);
+	void set_indent(int level);
+	void enable_timestamp();
+	void set_prefix(const std::string& prefix);
+};
+
 class logger {
 	char const *name_;
 	int severity_;
 public:
 	logger(char const *name, int severity): name_(name), severity_(severity) {}
-	std::ostream &operator()(const log_domain& domain,
+	log_in_progress operator()(const log_domain& domain,
 		bool show_names = true, bool do_indent = false) const;
 
 	bool dont_log(const log_domain& domain) const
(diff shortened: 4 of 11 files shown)


src/game_config_manager.cpp
```diff
@@ -129,12 +129,13 @@ bool map_includes(const preproc_map& general, const preproc_map& special)
 }
 } // end anonymous namespace
 
+#define LOG_GC_DEFINES FORCE_LOG_TO(out, log_config)
```
stevecotton, line 132: What does this line do?

CelticMinstrel: That line doesn't do anything - it's a relic of an alternate adjustment I considered making to the code in this function.

Pentarctagon:
Several of the travis jobs failed, it looks like.

Pentarctagon:
It's still failing - it just looks like it succeeded since with [ci skip] as the final commit it only ran the labeler task, which succeeded.

CelticMinstrel:
Any idea what the newest failure means? Seems like a weird link error in GUI2 or something? Or maybe I misread the log.

Pentarctagon:
For Windows it's just the build taking too long for there to be time for the unit tests to complete.

For the others, I'm not sure.  It looks like a linker error, but the only commonality I can see so far is that the failed builds are using the debug configuration and that it's happening across xcode, scons, and cmake.

soliton-:
Looks to me like this must be some errors from not completely rebuilding everything. Surely changes in log.?pp require a recompile of pretty much everything.

The PR looks pretty good to me. If the build issues are not there when building locally by hand I suggest to just merge this and see if there really is an issue then.

CelticMinstrel:
Okay, then I'll run a few tests to make sure it all works, and then I'll force-push, wait another day or so just in case, and merge it.

CelticMinstrel:
Just for information, this is the test I ran. The following diff:

```diff
diff --git a/src/wesnoth.cpp b/src/wesnoth.cpp
index 3bcda3f9550..94a76510346 100644
--- a/src/wesnoth.cpp
+++ b/src/wesnoth.cpp
@@ -799,6 +799,19 @@ static int do_gameloop(const std::vector<std::string>& args)
        plugins_context plugins("titlescreen", callbacks, accessors);

        plugins.set_callback("exit", [](const config& cfg) { safe_exit(cfg["code"].to_int(0)); }, false);
+       
+       // Quick test of threaded logging...
+       std::vector<std::future<void>> test;
+       for(int thread = 1; thread <= 10; thread++) {
+               LOG_GENERAL << "Starting thread " << thread << std::endl;
+               test.push_back(std::async(std::launch::async, [thread]() {
+                       LOG_GENERAL << "Thread " << thread << " has ID " << std::this_thread::get_id() << std::endl;
+                       std::this_thread::sleep_for(std::chrono::milliseconds(rand() % 100));
+                       for(int msg = 1; msg <= 10; msg++) {
+                               LOG_GENERAL << "Thread " << thread << " message " << msg << std::endl;
+                       }
+               }));
+       }

        for(;;) {
                // reset the TC, since a game can modify it, and it may be used
```

Then run as follows (obviously omit `--wconsole` if not on Windows):

```
.\wesnoth.exe --wconsole --debug --data-dir=. --log-info=general
```

The output I saw on my sample run was:

```
20201012 12:28:33 info general: Starting thread 1
20201012 12:28:33 info general: Starting thread 2
20201012 12:28:33 info general: Thread 1 has ID 9032
20201012 12:28:33 info general: Starting thread 3
20201012 12:28:33 info general: Thread 2 has ID 9520
20201012 12:28:33 info general: Starting thread 4
20201012 12:28:33 info general: Thread 3 has ID 9244
20201012 12:28:33 info general: Starting thread 5
20201012 12:28:33 info general: Thread 4 has ID 10568
20201012 12:28:33 info general: Starting thread 6
20201012 12:28:33 info general: Thread 5 has ID 10860
20201012 12:28:33 info general: Starting thread 7
20201012 12:28:33 info general: Thread 6 has ID 5840
20201012 12:28:33 info general: Starting thread 8
20201012 12:28:33 info general: Thread 7 has ID 11180
20201012 12:28:33 info general: Starting thread 9
20201012 12:28:33 info general: Thread 8 has ID 11016
20201012 12:28:33 info general: Starting thread 10
20201012 12:28:33 info general: Thread 9 has ID 10872
20201012 12:28:33 info general: Thread 10 has ID 9152
20201012 12:28:33 info general: Thread 1 message 1
20201012 12:28:33 info general: Thread 1 message 2
20201012 12:28:33 info general: Thread 1 message 3
20201012 12:28:33 info general: Thread 3 message 1
20201012 12:28:33 info general: Thread 4 message 1
20201012 12:28:33 info general: Thread 2 message 1
20201012 12:28:33 info general: Thread 1 message 4
20201012 12:28:33 info general: Thread 3 message 2
20201012 12:28:33 info general: Thread 6 message 1
20201012 12:28:33 info general: Thread 5 message 1
20201012 12:28:33 info general: Thread 7 message 1
20201012 12:28:33 info general: Thread 4 message 2
20201012 12:28:33 info general: Thread 2 message 2
20201012 12:28:33 info general: Thread 1 message 5
20201012 12:28:33 info general: Thread 3 message 3
20201012 12:28:33 info general: Thread 6 message 2
20201012 12:28:33 info general: Thread 5 message 2
20201012 12:28:33 info general: Thread 7 message 2
20201012 12:28:33 info general: Thread 8 message 1
20201012 12:28:33 info general: Thread 10 message 1
20201012 12:28:33 info general: Thread 9 message 1
20201012 12:28:33 info general: Thread 4 message 3
20201012 12:28:33 info general: Thread 2 message 3
20201012 12:28:33 info general: Thread 1 message 6
20201012 12:28:33 info general: Thread 3 message 4
20201012 12:28:33 info general: Thread 6 message 3
20201012 12:28:33 info general: Thread 5 message 3
20201012 12:28:33 info general: Thread 7 message 3
20201012 12:28:33 info general: Thread 8 message 2
20201012 12:28:33 info general: Thread 10 message 2
20201012 12:28:33 info general: Thread 9 message 2
20201012 12:28:33 info general: Thread 4 message 4
20201012 12:28:33 info general: Thread 2 message 4
20201012 12:28:33 info general: Thread 1 message 7
20201012 12:28:33 info general: Thread 3 message 5
20201012 12:28:33 info general: Thread 6 message 4
20201012 12:28:33 info general: Thread 5 message 4
20201012 12:28:33 info general: Thread 7 message 4
20201012 12:28:33 info general: Thread 8 message 3
20201012 12:28:33 info general: Thread 10 message 3
20201012 12:28:33 info general: Thread 9 message 3
20201012 12:28:33 info general: Thread 4 message 5
20201012 12:28:33 info general: Thread 2 message 5
20201012 12:28:33 info general: Thread 1 message 8
20201012 12:28:33 info general: Thread 3 message 6
20201012 12:28:33 info general: Thread 6 message 5
20201012 12:28:33 info general: Thread 5 message 5
20201012 12:28:33 info general: Thread 7 message 5
20201012 12:28:33 info general: Thread 8 message 4
20201012 12:28:33 info general: Thread 10 message 4
20201012 12:28:33 info general: Thread 9 message 4
20201012 12:28:33 info general: Thread 4 message 6
20201012 12:28:33 info general: Thread 2 message 6
20201012 12:28:33 info general: Thread 1 message 9
20201012 12:28:33 info general: Thread 3 message 7
20201012 12:28:33 info general: Thread 6 message 6
20201012 12:28:33 info general: Thread 5 message 6
20201012 12:28:33 info general: Thread 7 message 6
20201012 12:28:33 info general: Thread 8 message 5
20201012 12:28:33 info general: Thread 10 message 5
20201012 12:28:33 info general: Thread 9 message 5
20201012 12:28:33 info general: Thread 4 message 7
20201012 12:28:33 info general: Thread 2 message 7
20201012 12:28:33 info general: Thread 1 message 10
20201012 12:28:33 info general: Thread 3 message 8
20201012 12:28:33 info general: Thread 6 message 7
20201012 12:28:33 info general: Thread 5 message 7
20201012 12:28:33 info general: Thread 7 message 7
20201012 12:28:33 info general: Thread 8 message 6
20201012 12:28:33 info general: Thread 10 message 6
20201012 12:28:33 info general: Thread 9 message 6
20201012 12:28:33 info general: Thread 4 message 8
20201012 12:28:33 info general: Thread 2 message 8
20201012 12:28:33 info general: Thread 3 message 9
20201012 12:28:33 info general: Thread 6 message 8
20201012 12:28:33 info general: Thread 5 message 8
20201012 12:28:33 info general: Thread 7 message 8
20201012 12:28:33 info general: Thread 8 message 7
20201012 12:28:33 info general: Thread 10 message 7
20201012 12:28:33 info general: Thread 9 message 7
20201012 12:28:33 info general: Thread 4 message 9
20201012 12:28:33 info general: Thread 2 message 9
20201012 12:28:33 info general: Thread 3 message 10
20201012 12:28:33 info general: Thread 6 message 9
20201012 12:28:33 info general: Thread 5 message 9
20201012 12:28:33 info general: Thread 7 message 9
20201012 12:28:33 info general: Thread 8 message 8
20201012 12:28:33 info general: Thread 10 message 8
20201012 12:28:33 info general: Thread 9 message 8
20201012 12:28:33 info general: Thread 4 message 10
20201012 12:28:33 info general: Thread 2 message 10
20201012 12:28:33 info general: Thread 6 message 10
20201012 12:28:33 info general: Thread 5 message 10
20201012 12:28:33 info general: Thread 7 message 10
20201012 12:28:33 info general: Thread 8 message 9
20201012 12:28:33 info general: Thread 10 message 9
20201012 12:28:33 info general: Thread 10 message 10
20201012 12:28:33 info general: Thread 8 message 10
20201012 12:28:33 info general: Thread 9 message 9
20201012 12:28:33 info general: Thread 9 message 10
```

Of course, the output won't be exactly the same for everyone, but as you can see, lines are not getting interleaved.

Pentarctagon:
I'm able to reproduce the travis linking failure locally, so it's not an issue with travis.

CelticMinstrel:
Any idea what's causing it? Does it seem like it might be caused by a partial rebuild like someone suggested on IRC/Discord?

Pentarctagon:
I cleaned out all existing build-generated files, rebuilt from scratch, and it still happened, so the problem isn't a partial rebuild at least.

Pentarctagon:
The first error of `home/wesnoth-travis/src/gui/core/event/distributor.cpp:461: undefined reference to `gui2::event::mouse_button_event_types_wrapper<(gui2::event::ui_event)8, (gui2::event::ui_event)9, (gui2::event::ui_event)10, (gui2::event::ui_event)11, (gui2::event::ui_event)12, (gui2::event::ui_event)13>::sdl_button_down_event'` does point to debug-only logging though: https://github.com/wesnoth/wesnoth/blob/30d6be0be5daaa4b9a7cf4675178347a494c4ebc/src/gui/core/event/distributor.cpp#L461

CelticMinstrel:
Hmm, that's really odd. It seems to be declared here:

https://github.com/wesnoth/wesnoth/blob/30d6be0be5daaa4b9a7cf4675178347a494c4ebc/src/gui/core/event/distributor.hpp#L170

So I don't see how it can possibly be a link error. In fact, it shouldn't even have linkage? It's a static const variable of enum type.

Pentarctagon:
A quick google seems to indicate that getting a linker error or not depending on optimization level isn't an entirely unusual thing to happen, though I'm not sure what would be causing it to happen in this case.

src/formatter.hpp
```diff
@@ -67,6 +67,31 @@ class formatter
...
+	{
+        fn(stream_);
+		return *this;
+    }
+
+    formatter&& operator<<(std::ios_base&(*fn)(std::ios_base&)) &&
+	{
+        fn(stream_);
+		return std::move(*this);
+    }
```
stevecotton, line 94: The indentation here is mixed up, please run it through clang-format.

src/log.hpp
```diff
@@ -112,12 +113,32 @@ void set_strict_severity(int severity);
 void set_strict_severity(const logger &lg);
 bool broke_strict();
 
+// A little "magic" to surround the logging operation in a mutex.
+// This works by capturing the output first to a stringstream formatter, then
+// locking a mutex and dumping it to the stream all in one go.
+// By doing this we can avoid rare deadlocks if a function whose output is streamed
+// calls logging of its own.
+// We overload operator| only because it has lower precedence than operator<<
+// Any other lower-precedence operator would have worked just as well.
```
stevecotton, line 122: Why does it need to be an operator, instead of a normal function?

Pentarctagon:
As an aside, something I wasn't sure about and was reminded about by stevecotton's question above - is there a reason the function doing the actual logging is `operator()` instead of a regular function as well?

src/log.hpp
CelticMinstrel: In order to support the streaming syntax. If it were a regular function, all log statements would be forced to end with a closing parenthesis.

CelticMinstrel:
I've wondered why that was an operator, too. There doesn't seem to be any reason for it that I can see, since it's not really part of an exposed API and doesn't seem to be necessary to support any particular syntactic form.

src/log.hpp
Pentarctagon: And the reason for that is to avoid needing to adjust the syntax of every log statement currently in the codebase, right?

soliton-: Yes, that would otherwise need to happen.

stevecotton: I think an earlier line of this comment block is from an earlier draft, and doesn't match the code under review - that's the "This works by capturing the output first to a stringstream formatter" line. If that line was correct, then I think the output could be done in the formatter's destructor, removing the need for the operator overload.

soliton-:
So if the errors are real my first try would be to revert the TST_GUI_I macro and see if that fixes it.

src/log.cpp
```diff
@@ -218,33 +220,58 @@ std::ostream &logger::operator()(const log_domain& domain, bool show_names, bool
...
+
+void log_in_progress::enable_timestamp() {
+	timestamp_ = true;
+}
+
+void log_in_progress::set_prefix(const std::string& prefix) {
+	prefix_ = prefix;
+}
+
+void scope_logger::do_log_entry(const std::string& str) noexcept
```
stevecotton, line 268: `noexcept` on a function that's doing I/O and manipulating strings? It was there before, but I think that indicates a bug in the existing code.

stevecotton:
> I don't see how it can possibly be a link error. In fact, it shouldn't even have linkage? It's a static const variable of enum type.

That looks like a weird use of templates, and while it says "static const" it's in a template that has multiple instances. Maybe refactor the code to be less weird, and see if the compiler is happier about it?

BTW, compiling locally with GCC 10, the linker errors come with another message that's easier to find info on: "warning: creating DT_TEXTREL in a PIE", which leads over to https://stackoverflow.com/questions/43367427/32-bit-absolute-addresses-no-longer-allowed-in-x86-64-linux

src/log.hpp
CelticMinstrel: > And the reason for that is to avoid needing to adjust the syntax of every log statement currently in the codebase, right?

Pretty much. Also I kinda like the streaming syntax.

> If that line was correct, then I think the output could be done in the formatter's destructor, removing the need for the operator overload.

I think this would be a possible setup, yes, but I chose to use the pre-existing quick formatter instead of implementing a separate custom thing for the logging mechanism.

src/log.cpp
CelticMinstrel: Yeah, I'm not really sure. I agree that it seems suspect, but it's probably out-of-scope for this PR in any case.

CelticMinstrel:
> Maybe refactor the code to be less weird, and see if the compiler is happier about it?

I believe @Vultraz has plans here, so maybe I'll let him do that and rebase and see if that helps.

src/formatter.hpp
CelticMinstrel: I think the force-push should fix this.

Vultraz:
If you're referring to the boost::mpl craziness in the GUI2 dispatcher, yes, that is gone now.

src/log.hpp
Pentarctagon: One option that keeps the streaming syntax (but wouldn't avoid needing to update the existing logging statements) would be to use the macro to pass an unnamed ostringstream into the logger function.  For example:
```
#define MTEST(STREAM) test(static_cast<std::ostringstream&&>(std::ostringstream() << STREAM));

void test(std::ostringstream oss)
{
	std::cout << oss.str();
}

MTEST("hello " << "world" << "\n")
```

Pentarctagon:
It looks like the error remains.

PR #5199, opened 2020-10-09
Add the 'Add-ons' tag to the pull-request labeler.

kabachuha:
Added support for the `Add-ons` tag to the pull-request labeler.
The file changes in the following directories will be marked as related to the addons system:
  - src/server/campaignd/**
  - src/addon/**
  - src/gui/dialogs/addon/**

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,3 +1,8 @@
+Add-ons:
+  - src/server/campaignd/*
+  - src/addon/*
+  - src/gui/dialogs/addon/*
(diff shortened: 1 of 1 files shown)


soliton-:
While there are currently no subdirs I would suggest to use `**` just in case that changes.

kabachuha:
/**/*, so to check any possible subdirs, or just /**?

soliton-:
`/**` should already include all subdirs and files.

Issue #5194, opened 2020-10-06
When you change the language to Chinese, some words will disappear


lishaoxia1985:
<img width="1299" alt="YueR" src="https://user-images.githubusercontent.com/49801619/95154281-0d1e3500-07c4-11eb-8139-0349dba08ae4.png">

The bug is this:
 1. Open the app in English
 2. Change the language to Chinese, then some words will disappear
 3. If you reopen the app in Chinese, the game will show normally

I advise that all the language use the font "wqy-microhei", its license is under Apache2 and GPLv3. This is website: 
https://sourceforge.net/projects/wqy/files/wqy-microhei/0.2.0-beta/

Pentarctagon:
A screenshot would be useful, as well as the information listed [here](https://forums.wesnoth.org/viewtopic.php?f=4&t=25660).

Beyond that, if the text displays correctly after restarting, it's unlikely to be an issue with the font itself.

stevecotton:
Reproducible with 1.14.9, 1.14 (built from a recent Git commit) and current master:

The Battle for Wesnoth version 1.15.5+dev (at 87469933bc)
Running on Debian GNU/Linux bullseye/sid
Distribution channel: Default

Game paths
==========

Data dir:        /home/USER/development/other-source/wesnoth/main
User config dir: /home/USER/.config/wesnoth1.15/
User data dir:   /home/USER/.local/share/wesnoth/1.15
Saves dir:       /home/USER/.local/share/wesnoth/1.15/saves
Add-ons dir:     /home/USER/.local/share/wesnoth/1.15/data/add-ons
Cache dir:       /home/USER/.cache/wesnoth

Libraries
=========

Boost:             1.71
OpenSSL/libcrypto: 1.1.1g (runtime 1.1.1g)
Cairo:             1.16.0 (runtime 1.16.0)
Pango:             1.46.2 (runtime 1.46.2)
SDL:               2.0.12 (runtime 2.0.12)
SDL_image:         2.0.5 (runtime 2.0.5)
SDL_mixer:         2.0.4 (runtime 2.0.4)
SDL_ttf:           2.0.15 (runtime 2.0.15)

Features
========

JPEG screenshots:               yes
Lua console completion:         yes
Legacy bidirectional rendering: yes
D-Bus notifications back end:   yes

Current video settings
======================

SDL video drivers:       [x11] wayland KMSDRM dummy
Window size:             1024x768
Screen refresh rate:     60
[...]

PR #5200, opened 2020-10-09
Check which glyphs are in each font, instead of using a list hardcoded in WML

stevecotton:
This is a "it's known to be broken, so let's rip it out and put in a minimal
replacement" change. Updated raw pointers to smart pointers just because,
updated the docs a bit, and ended up with a big change.

Fixes the most visible part of #5194, where Chinese needed DroidSansFallbackFull
to be loaded before DroidSansJapanese. The removed code in
`char_block_map::insert` and `char_block_map::compress` had a bug that
triggered when one font had a contiguous range of codepoints that was a
superset of several ranges in another font - this meant it treated the first
font containing U+4E00 as having the whole CJK Unified Ideographs block.

Remove the now-unused font codepoints WML. There is no schema change for this,
it seems the data/hardwired/fonts.cfg file isn't checked by the validation.

Optimise calling set_font_list with the same list (but possibly reordered), by
reusing the already-loaded fonts.


Vultraz:
I can't say this is an area of the code I really understand, so if you've tested it and it works it's a go from me.

src/font/sdl_ttf.cpp
```diff
@@ -43,122 +43,82 @@ static lg::log_domain log_font("font");
...
-			if (/* next.first */ i->first <= last) {
-				insert(first, i->first - 1, id);
-				return;
-			}
-		}
-		cbmap.emplace(first, block_t(last, id));
-	}
-	/**
-	 * Compresses map by merging consecutive ranges with the same font, even
-	 * if there is some unassociated ranges in-between.
```
AI0867, line 101: So, this bit of functionality here was the cause of the bug?

stevecotton: Yup. I think I found more bugs as the refactor grew, but this was the base of the Chinese characters being invisible.

PR #5203, opened 2020-10-11
Update editor cfg file to reflect earlier changes in scaling

### LINKED ISSUE #5193: Editor graphically a mess at 800x600
Local Linux builds of 1.15.4 and current master

### Describe the bug
The editor does not display correctly at 800x600 (aka the smallest resolution selectable in the preferences). The map overlaps with the editor's sidebar palette, and what is visible changes in response to mouseovers.

The bug isn't present in 1.14 - while that has #2618, it's a lesser bug than the 1.15 situation. The two versions in the screenshots have the same bugs, they look different because I scrolled the map in only one of them.

wesnoth_editor_800x600

gwhitney:
Commit 43de778 made all themes scale from their base dimensions rather
  than from an arbitrary hard-coded 1024x768 pixel size. As a result, the
  800x600 layout for the editor needs its main map explicitly sized --
  the 843-pixel width no longer fits in 800x600, and it isn't being scaled
  down from 1024. (But once it is explicitly set for the 800x600 layout,
  it will scale up for pixel widths between 800 and 1024). This commit
  adds such an explicit sizing, analogous to the one found in default.cfg.

  Resolves #5193.

### DIFF
--- a/data/themes/editor.cfg
+++ b/data/themes/editor.cfg
@@ -1004,6 +1004,11 @@
         width=800
         height=600
 
+        [change]
+            id=main-map
+            rect="0,+7,+619,600"
+        [/change]
+
         [remove]
             id=top_button_file3
         [/remove]


gwhitney:
OK, wow, that wins the award for my fastest-merged PR ever. Thanks, Glen

PR #5206, opened 2020-10-12
Fix tip for Oasis healing

vgaming:


vgaming:
It seems that a lot of people don't really know about Oasis healing, only village healing. Hopefully mentioning Oasis in the tactics tips would help

data/tips.cfg
```diff
@@ -73,7 +73,7 @@
     source= _ "<i>― The Wesnoth Tactical Guide</i>"
 [/tip]
 [tip]
-    text= _ "Poisoned units will lose eight hitpoints every turn until they are one point from dying. They will then remain at death’s door until the poison is removed. Healers can prevent a poisoned unit from deteriorating, but the poison will not be removed unless the unit regenerates, or begins a turn in a village or next to a unit with the <i>Cures</i> ability."
+    text= _ "Poisoned units will lose eight hitpoints every turn until they are one point from dying. They will then remain at death’s door until the poison is removed. Healers can prevent a poisoned unit from deteriorating, but the poison will not be removed unless the unit regenerates, or begins a turn in a village, oasis, or next to a unit with the <i>Cures</i> ability."
```
stevecotton, line 76: Should there be an "an" before "oasis"?

vgaming: Good point, thanks. I've changed the wording and pushed an amended commit

stevecotton:
Looks good to me. I'd merge now, but I like to leave the CI jobs to finish first.

vgaming:
@stevecotton definitely! Too bad we're not on GitLab where a button exists to "Merge when pipeline succeeds" :)

CelticMinstrel:
I don't think this is grammatically correct: `begins a turn in a village, an oasis, or next to a unit with…`

Consider it as a logical expression. Before the change, it was correct — the unit begins a turn (in a village) or (next to a unit). After the change, however, it's confused — the unit begins a turn (in (a village), (an oasis)) or (next to a unit). Analyzed that way, you can see that there are in fact two separate disjunctions, and one of them is missing the key word that marks it as being a disjunction. There are two possible fixes that I know of: either add "or", or add "in":

* `begins a turn in a village or an oasis, or next to a unit with…`
* `begins a turn in a village, in an oasis, or next to a unit with…`

In the first variation you could probably get away with omitting "an" as well.

vgaming:
@CelticMinstrel English is not my native language, but I think you're right! Thanks for pointing the tricky part out (and requiring time).. My intention is just to educate people on Oasis better.

I'll change the wording to the one you described (the second choice, more boring but unambiguous).

vgaming:
I've force-pushed again to leave a cleaner history in git

Pentarctagon:
I assume this can be merged, then?

PR #5207, opened 2020-10-12
Messenger Micro AI: add [avoid] tag functionality

mattsc:
This is complete as far as I am concerned, I just want to give a couple days for people to comment and for @Hejnewar to test.  I see no reason why this should not go into 1.15.6.

### DIFF
--- a/data/ai/lua/ai_helper.lua
+++ b/data/ai/lua/ai_helper.lua
@@ -739,11 +739,12 @@ function ai_helper.get_locations_no_borders(location_filter)
     return locs
 end
 
-function ai_helper.get_closest_location(hex, location_filter, unit)
+function ai_helper.get_closest_location(hex, location_filter, unit, avoid_map)
     -- Get the location closest to @hex (in format { x, y })
     -- that matches @location_filter (in WML table format)
     -- @unit can be passed as an optional third parameter, in which case the
     -- terrain needs to be passable for that unit
+    -- @avoid_map (location set): if given, the hexes in avoid_map are excluded
     -- Returns nil if no terrain matching the filter was found
 
     -- Find the maximum distance from 'hex' that's possible on the map
@@ -779,6 +780,14 @@ function ai_helper.get_closest_location(hex, location_filter, unit)
         end
 
         local locs = wesnoth.map.find(loc_filter)
+
+        if avoid_map then
+            for i = #locs,1,-1 do
+                if avoid_map:get(locs[i][1], locs[i][2]) then
+                    table.remove(locs, i)
+                end
+            end
+        end
 
         if unit then
             for _,loc in ipairs(locs) do
--- a/data/ai/lua/ai_helper.lua
+++ b/data/ai/lua/ai_helper.lua
@@ -295,6 +295,9 @@ function ai_helper.robust_move_and_attack(ai, src, dst, target_loc, cfg)
     --     parameter is true, a partial move is done instead.
     --   weapon: The number (starting at 1) of the attack weapon to be used.
     --     If omitted, the best weapon is automatically selected.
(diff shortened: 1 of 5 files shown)


CelticMinstrel:
The only thing I'm wondering is whether changing `find_path_with_shroud` to `find_path_with_avoid` means it's now ignoring shroud when previously it didn't.

mattsc:
Yeah, I was wondering that also until I reminded myself of what the code does.  `find_path_with_avoid` actually calls `find_path_with_shroud` [here](https://github.com/wesnoth/wesnoth/blob/master/data/ai/lua/ai_helper.lua#L1888).

mattsc:
There are still some issues with this, most likely with the ai_helper functions rather than the MAI code itself.  I'll need to work on this some more, so this will not be ready for 1.15.6.

mattsc:
I need some help...  That last commit 6014b94 works just fine for what it is supposed to do, but it also triggers an assert in the Hang Out MAI:

`Assertion failed: (use_undo || (!resources::undo_stack->can_redo() && !resources::undo_stack->can_undo())), function run, file /Users/mattsc/mats/pers/Wesnoth/wesnoth/src/synced_context.cpp, line 57.`

I have no idea why this is happening, why it is only happening for this MAI and not others, what it means in the first place, and whether it's an AI problem or an engine bug.

What I do know is that it is triggered by [this line](https://github.com/wesnoth/wesnoth/blob/master/data/ai/micro_ais/cas/ca_hang_out.lua#L111), and that it does not happen when I delete the `AH.checked_move()` 3 lines earlier.

To reproduce, simply start the Hang Out MAI test scenario (either from the CL with `-thang_out` or from the title screen with the hotkey) and end the player's turn.  The crash happens the moment the first footpad moves.

Any help would be much appreciated!

CelticMinstrel:
Hmm, I guess this is intentional, but this only supports a single container variable per AI... essentially just giving you a namespace where you can store stuff, right? I guess improving on that might be out of scope here though.

The other thing I'm noticing is that you're passing the full variable name in the synced command config. I think I would recommend including only the `ai_id` and have the synced command itself construct the variable name from that. This means someone can't use the command to cheat any variable they want. (Well, any _container_ variable at least.)

I don't have any insight on the assert, though.

mattsc:
> Hmm, I guess this is intentional, but this only supports a single container variable per AI... essentially just giving you a namespace where you can store stuff, right? I guess improving on that might be out of scope here though.

Yes, that is intentional (and unchanged from previous behavior).  We can improve on that at some other time if we come up with a use case for it, but I agree, it is beyond the scope of this PR.

> The other thing I'm noticing is that you're passing the full variable name in the synced command config. I think I would recommend including only the `ai_id` and have the synced command itself construct the variable name from that. This means someone can't use the command to cheat any variable they want. (Well, any _container_ variable at least.)

Sure, that's not a problem.  I can do that.

mattsc:
> The other thing I'm noticing is that you're passing the full variable name in the synced command config. I think I would recommend including only the ai_id and have the synced command itself construct the variable name from that. This means someone can't use the command to cheat any variable they want. (Well, any container variable at least.)

d5e3173 does this now.

I'll look into what causes the assert next.  I would really like to merge this before the 1.15.11 release though.  If I don't figure this out until then, I will back the synced_command (replay fix) commit back out and merge the rest.

mattsc:
As expected, the crash is not due to a problem with the code in this PR.  It is separate and fixed by #5595.  However, that is only one possibility to fix it, so I am going to let the PR sit there for a few days to see if somebody thinks it should be done differently.  Once #5595 is merged, we can merge this PR also.

mattsc:
I've backed out the part with the synced_command and squashed things.  I'll just wait for the tests to finish and then I'll merge this.

PR #5208, opened 2020-10-13
revert Redesign force cth macro

newfrenchy83:
revert https://github.com/wesnoth/wesnoth/commit/866420adf85ec1da02f565f2efd1933bf27419c4#diff-ca2b2f3f52116e047fc943b701f7798e0008c99644d9edf7ade1daf66ae61b7d

when i replaced 1.12/1.14 version of FORCE CTH, the weapon abilities overwrited the regular cth specials but now the cth mike marksman or magical don't affected, i must place the 1.14 version of macro for fixing that.

### DIFF
--- a/data/core/macros/utils.cfg
+++ b/data/core/macros/utils.cfg
@@ -513,8 +513,6 @@
     #!         greater_than=10
     #!     [/variable]
     #! )}
-    # WARNING : if an unit must disappear in mid-battle before attack end was fired, don't use [unstore_unit] but recreate unit or cth modifications
-    # persist. Or don't use this macro.
     [event]
         name=attack
         first_time_only=no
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
I don't quite understand. What is the issue this is supposed to fix? The marksman and magical abilities shouldn't be directly affected by force chance to hit, they should just be overwhelmed by it… if magical and marksman are limiting the chance to hit even under force chance to hit, then yeah that's a bug, but this doesn't seem like the correct solution in that case. Or is there some other issue?

newfrenchy83:
@CelticMinstrel If i use force cth macro to force cth value to 60 or less ( 0 by example) on an attack with marksman or magical weapon
Active, then the value given by weapon specials not changed and the unit enemy can be hit. This is the bug what i created with using of ability instead to remain on 1.12/14 version of macro.

CelticMinstrel:
I think I'd prefer if we found a way to change this that still involves using an ability, because the old code that overwrites your marksman ability seems pretty dirty.

newfrenchy83:
@CelticMinstrel can it using object for remove all cth specials by tags instead of id ?

newfrenchy83:
@CelticMinstrel It must overwrite specials chance_to_hit if their value exceed macro cth value, and it must doing that without clear other specials, and the old code seem the one conveniant, using abilities is unuse in that case.

PR #5209, opened 2020-10-13
Refactoring the templates in distributor.cpp (makes PR #5198 build)

stevecotton:
This is part-way though a refactor, but I'm not committing myself to completing
it; publishing for the comments and discussion.

This solves PR #5198's compile errors about undefined references, but the
solution here is to expand the templates into the multiple hardcoded copies
block starting with `switch(events_.sdl_button_down_event)`, which is the
first big block of changes in the diff.

### DIFF
--- a/src/gui/core/event/distributor.cpp
+++ b/src/gui/core/event/distributor.cpp
@@ -357,47 +357,92 @@ void mouse_motion::stop_hover_timer()
 
 #undef LOG_HEADER
 #define LOG_HEADER                                                             \
-	"distributor mouse button " << name_ << " [" << owner_.id() << "]: "
+	"distributor mouse button " << events_.name << " [" << owner_.id() << "]: "
 
-template<typename T>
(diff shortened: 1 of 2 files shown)


Vultraz:
This looks like a reasonable change.

CelticMinstrel:
I don't see any particular issues with this change, though that switch sure is ugly...

Mind if I cherry-pick this into my PR?

Also @Vultraz - do you have any further improvements to suggest? I didn't think of anything but I haven't put much thought into it either...

stevecotton:
If you want to cherry-pick this, please feel free. Many thanks for the feedback; I looked for further improvements but felt it needed a general redesign.

Vultraz:
@CelticMinstrel not presently. I'll wait and see what the final PR looks like.

stevecotton:
I don't want to commit to the time for a major rewrite of these .cpp and .hpp files. At the moment I think the current state of the PR is better than the existing codebase, but that improving from here is going to be a lot of effort.

PR #5213, opened 2020-10-16
add overwrite_specials attribute to abilities like specials for superced other specials(abilities or traditionnal)

newfrenchy83:
if force cth macro is used for impose an value less what value of cth specials, the macro, the value usd will be the cth specials value.
this PR is an alternate soluce to https://github.com/wesnoth/wesnoth/pull/5208 PR


Pentarctagon:
So essentially, the current situation is that if a unit has both a weapon special and an ability that modify an attack's chance to hit, the weapon special takes precedence. This this adds an `overwrite_specials` attribute which will force the ability to precedence?

CelticMinstrel:
I assume setting `cumulative=yes` in the CTH macro does not work?

If that's true, then I tentatively support this, as it's pretty close to the sort of solution I was envisioning.

newfrenchy83:
> 
> 
> I assume setting `cumulative=yes` in the CTH macro does not work?
> 
> If that's true, then I tentatively support this, as it's pretty close to the sort of solution I was envisioning.

if cumulative=yes in CTH macro then if base_value geater_than cth value, base_value will be used instead of value macro.

If you speak of marrksman abilities, in the case of cth abilities with value=50 and cumulative=no, marsksman special value=60 and cumulative=yes with base_value=80%, then chance_to_hit will be 80% in normal condition, or 50% if overwrite_specials=yes

CelticMinstrel:
> if cumulative=yes in CTH macro then if base_value geater_than cth value, base_value will be used instead of value macro.

Can you explain what this means with an example? I think I get it but I'm not quite sure.

newfrenchy83:
@CelticMinstrel if a marsksman attack an unit human in water, his chance to hit will be to 80%, this is the base value of cth without specials or abilities esuivalent, if with cumulative=yes the greater_value will be used, ans same if an aura abilities cth to value=50, and cumulative=no is adjacent, thentheresult is unchanged.

CelticMinstrel:
So it sounds like using `cumulative=yes` only works if you want to force a high chance to hit, but doesn't work if you want to force a low chance to hit (eg 30%)?

newfrenchy83:
@CelticMinstrel Yes

newfrenchy83:
@CelticMinstrel i commited change for what abilities with overwrite_special superced also other abilities like specials

Pentarctagon:
It looks like someone else has run into this problem: https://forums.wesnoth.org/viewtopic.php?f=4&t=53607

@newfrenchy83 could you add a unit test for this as well?  Something like giving a unit a weapon special that has 100% chance to hit, using FORCE_CHANCE_TO_HIT to make it 0%, and then confirming that the attack missed?

newfrenchy83:
> 
> 
> It looks like someone else has run into this problem: https://forums.wesnoth.org/viewtopic.php?f=4&t=53607
> 
> @newfrenchy83 could you add a unit test for this as well? Something like giving a unit a weapon special that has 100% chance to hit, using FORCE_CHANCE_TO_HIT to make it 0%, and then confirming that the attack missed?

This is done

Pentarctagon:
Sorry, I mean a test like what's in the `data/tests/scenarios` directory, which get run as part of the WML unit tests each commit.

newfrenchy83:
> 
> 
> Sorry, I mean a test like what's in the `data/tests/scenarios` directory, which get run as part of the WML unit tests each commit.

added a test file for force cth macro in `data/tests/scenarios` directory

newfrenchy83:
@Pentarctagon i has forget to add test to wml_unit test.

newfrenchy83:
@Pentarctagon unit test added and validated.

Pentarctagon:
For the `attack_type::list_ability` method, is there a reason to create `abil_over_list` and `abil_over_other_list`?  It looks like it would be equivalent to add the check for `*(i->ability_cfg)["overwrite_specials"].to_bool()` to the existing if check when looping over `abil_list` and `(*i->ability_cfg)["overwrite_specials"].to_bool()` to the existing if check when looping over `abil_other_list`.

For the `attack_type::get_special_ability` method, it looks like you could replace `overwrite = true;` with `return abil_list;` and then get rid of the `overwrite` variable.

newfrenchy83:
> 
> 
> For the `attack_type::list_ability` method, is there a reason to create `abil_over_list` and `abil_over_other_list`? It looks like it would be equivalent to add the check for `*(i->ability_cfg)["overwrite_specials"].to_bool()` to the existing if check when looping over `abil_list` and `(*i->ability_cfg)["overwrite_specials"].to_bool()` to the existing if check when looping over `abil_other_list`.
> 
> For the `attack_type::get_special_ability` method, it looks like you could replace `overwrite = true;` with `return abil_list;` and then get rid of the `overwrite` variable.

i fixed the defaults in the writing.

Pentarctagon:
Why are `abil_list` and `abil_other_list` getting repopulated if they're empty?

newfrenchy83:
> 
> 
> Why are `abil_list` and `abil_other_list` getting repopulated if they're empty?

Because the list is necessary for the ability to work, the process must therefore be done in two stages for each list, first test overwrite_specials with the other filters to verify both that the ability is active and that it has priority over the other abilities of the list, if overwrite_specials = false for all the abilities, then the list is empty and all the abilities are inactive, so that the abilities work normally without priority on the specials it is necessary to recreate the list but without overwrite_specials in filters, if any abilities are active, then the chosen value will be the highest in the final list in get_special_abilities.

Pentarctagon:
@CelticMinstrel I think this makes sense to me; any opinion on whether it can be merged?

newfrenchy83:
> 
> 
> @CelticMinstrel I think this makes sense to me; any opinion on whether it can be merged?

@Pentarctagon I fear what @CelticMinstrel don't reply immediately, my last comment to him was 17 day ago and he don't reply yet.

Pentarctagon:
I pinged him on Discord/IRC and he said he should be able to have a look in the next few days.

CelticMinstrel:
Just based on the comments, this makes sense. I haven't taken a look at the code, but I should be able to glance over that tomorrow.

data/scenario-test.cfg
```diff
@@ -4380,6 +4397,34 @@ unplagueable: $wml_unit.status.unplagueable"
...
+        name=attacker hits
+        first_time_only=yes
+        [filter]
+            id=cth_tester
+        [/filter]
+
+        [message]
+            speaker=narrator
+            image=wesnoth-icon.png
+            message="FORCE_CHANCE_TO_HIT macro don't working fine. "
```
CelticMinstrel, line 4411: If this test is going to exist, we need to make the messages actually comprehensible.

CelticMinstrel (approved):
Other than that one issue in the test scenario, I don't see any issues with this PR.

Pentarctagon:
@newfrenchy83 Please change that to `Attacker hit despite having 0% chance to hit.`.

data/scenario-test.cfg
```diff
@@ -231,8 +231,25 @@ Xu, Xu, Qxu, Qxu, Ql, Ql, Ql, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Gg, Gg, Gg, Gg, Gg
...
+            [modifications]
+                [object]
+                    [effect]
+                        apply_to=attack
+                        [set_specials]
+                            mode=append
+                            [chance_to_hit]
+                                id=chance100
+                                name="Chance 100%"
+                                description="This attack has 100% to hit the cible."
```
Pentarctagon, line 246: "cible" should be "target"?

newfrenchy83:
i removed testing in scenario-test.cfg

PR #5214, opened 2020-10-16
Make use of std::chrono consistent everywhere

Pentarctagon:


CelticMinstrel:
My only complaint is that I'd prefer to see `1'000'000` instead of `1000000`, as it makes the code clearer. I'm _pretty_ sure that's in C++14? If not, it's not the end of the world.

Pentarctagon:
It compiled locally, so I think so, but let's see what travis says.

CelticMinstrel:
Hmm… come to think of it, instead of `std::string(6 - fractional.size(), '0')`, shouldn't you use `std::setfill('0') << std::setw(6)`? This requires including `<iomanip>` but I think it's cleaner than manually constructing a fill string.

The only minor downside is that `setfill` changes stream state, so it may be better to restore it after, something like:

```
char fill = out.fill('0');
out << ... << std::setw(6) << micros;
out.fill(fill);
```

Still needs `<iomanip>` for `setw`.

Pentarctagon:
For simple, one-off case like this, using string is fine (IMO).

CelticMinstrel:
I think it's better to keep things consistent and always use the stream fill character when you want a stream fill, but… as long as it produces the correct result, I suppose it's fine.

Pentarctagon:
The way I was thinking of it was "padding a string" rather than "filling a stream", so that's probably the difference.

CelticMinstrel:
That only makes sense if the thing you need to pad is a string in the first place, but here, it's an integer…

Pentarctagon:
True.

soliton- (approved):

PR #5215, opened 2020-10-18
SoF S9 - some cleanup to the final scenario

doofus-01:
- slight update to the map
 - multiple ways to get the orc to break his tablet


Pentarctagon:
The WML schema validation is currently failing due to `Invalid key 'remove=' in tag [event]`.

doofus-01:
Well, then something else needs cleanup.
https://wiki.wesnoth.org/EventWML#remove

stevecotton:
/me self-assigns "deprecate [event]remove=yes in favor of [remove_event]"

stevecotton:
The flashing glyphs around 27,20 are drawn on a layer above the units.

The bridges around 34,6 are very dark, it looks as if the orcs are floating in mid-air.

nemaara (approved):
As long as the code runs this change seems fine.

PR #5216, opened 2020-10-18
Deprecate [event]remove=yes, recommend [remove_event]

stevecotton:
The remove attribute for [event] was never added to the validation schema,
and using [remove_event] seems easier to search WML for. Let's just recommend
the new key and deprecate the old one.

Tested with the current implementation of #5215

### DIFF
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -766,6 +766,7 @@ end
 
 function wml_actions.event(cfg)
 	if cfg.remove then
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Looks good to me. Since it was never in the schema, I assume all uses have already been removed from mainline, right?

Pentarctagon:
Any existing uses would've been giving the same same schema validation error as #5215 is getting.

Pentarctagon (approved):

PR #5218, opened 2020-10-19
Fix initiative don't affect allied side in Httt

newfrenchy83:


soliton-:
Is the idea that the description mentions "friendly units" and that should mean allies are included?

newfrenchy83:
@soliton- yes, in 1.14 macro we have  [filter_side]
        [allied_with]
            [has_unit]
                id="Li'sar"
            [/has_unit]
        [/allied_with]
    [/filter_side]

initiative must affect all allies of Li'sar and not only her own side, but i forget to put affect_allies=yes in 1.15

soliton-:
Ah, a regression from 1.14.

PR #5219, opened 2020-10-20
Add docs about crowns and ellipses (fixes #4455)

### LINKED ISSUE #4455: Add a help page for gold/silver crowns and loyal marker

stevecotton:
The image added here is a grass background to draw units on via Image Path
Functions. For the documentation, a square image looks better than a hexagon
that has black borders.


stevecotton:
Force pushed because I forgot to run optipng, and forgot to run it with the most aggressive options.. Have done that now.

Pentarctagon:
Should the po hint be below the second set of images as well?

data/core/help.cfg
```diff
@@ -390,7 +390,58 @@ If a strike is determined to hit, it will always do at least 1 point of damage.
...
+            "<img>src='help/empty-square.png~BLIT(misc/ellipse-top.png)~BLIT(misc/ellipse-bottom.png)~BLIT(units/human-loyalists/spearman.png~TC(1, magenta))~BLIT(misc/loyal-icon.png)'</img>" +
+            "<img>src='help/empty-square.png~BLIT(misc/ellipse-leader-top.png)~BLIT(misc/ellipse-leader-bottom.png)~BLIT(units/human-magi/white-mage.png~TC(1, magenta))~BLIT(misc/leader-expendable.png)'</img>" + "
+
+" + _ "• Leaders are shown with a <bold>text='gold crown'</bold>" + "
+" + _ "• Heroes have a <bold>text='blueish-silver crown'</bold>. Losing any of these units is likely to result in an immediate defeat." + "
+" + _ "• <ref>dst='traits_loyal' text='Loyal units'</ref> have the <bold>text='loyal icon'</bold>. Most campaigns will give you a few loyals, however they never occur randomly — they are always characters that are a part of the narration, never random recruits. Losing them doesn’t result in automatic defeat." + "
+" + _ "• Expendable leaders are shown with a <bold>text='silver crown'</bold>. Rarely encountered in campaigns, these units can recruit but losing them doesn’t automatically lead to a defeat." + "
```
stevecotton, line 437: Half this text needs to be removed - I copied it to a separate paragraph, but forgot to trim the original paragraph.

stevecotton:
> Should the po hint be below the second set of images as well?

I thought it was necessary here as context for translating "From left to right", but not for the text under the second set of images.

The next push will have a rewrite of most of the text.

CelticMinstrel:
Your two po notes look like they are utterly useless, because they refer to a "paragraph" (technically a bulleted list) but are completely divorced from the paragraph they refer to due to it being split up into one string per bulleted item and having the po note only on the first string. The note also seems too vague to really be useful even if that issue was fixed (eg by merging the bulleted list into one string or by duplicating the po note for every line).

I'm not sure what the fix for this is, honestly. Maybe the po note simply isn't useful at all and can be removed. Maybe it should be clarified and repeated for every bulleted item, or clarified differently for each item.

I'm also a _little_ dubious of including the bullets in the translatable string. There's probably lots of precedent for it, and I'm not sure what's _better_, but personally I'd move the bullet out into an untranslatable portion. Probably need more opinions on this before actually making a change.

stevecotton:
At least in poedit, the strings tend to be presented to the translator in the same order that the occur in the WML; so a note of the first line of a list is going to have been seen recently by anyone doing the first translation of this paragraph.

stevecotton:
> I'm also a _little_ dubious of including the bullets in the translatable string. There's probably lots of precedent for it, and I'm not sure what's _better_, but personally I'd move the bullet out into an untranslatable portion. Probably need more opinions on this before actually making a change.

@ivanovic and @shikadiqueen , please could we have your feedback on this?

There is some precedent, but it seems to be from only 3 authors (and I'm one of those 3).

shikadiqueen:
Just keep the bullets. Splitting translatable strings up is generally very confusing unless people are willing to read the source file side-by-side while translating (which IME is something people rarely do).

stevecotton:
Thanks for the feedback. The build failure is a simple timeout on the Windows build.

data/core/help.cfg
```diff
@@ -390,7 +390,61 @@ If a strike is determined to hit, it will always do at least 1 point of damage.
...
+            "<img>src='help/empty-square.png~BLIT(misc/ellipse-top.png)~BLIT(misc/ellipse-bottom.png)~BLIT(units/human-loyalists/spearman.png~TC(1, magenta))~BLIT(misc/loyal-icon.png)'</img>" + "
+
+" +
+            #po: The help page has a set of images above this paragraph, in the same order as the first set of bullet points
+            _ "• Leaders are shown with a <bold>text='gold crown'</bold>.
+• Heroes have a <bold>text='blueish-silver crown'</bold>.
+• Expendable leaders are shown with a <bold>text='silver crown'</bold>.
+• <ref>dst='traits_loyal' text='Loyal units'</ref> have the <bold>text='loyal icon'</bold>.
+
+Of these categories, only leaders are built into the basic game rules and game engine. The others are found in campaigns, and are characters that form part of the narration. The exact rules for them are set by their campaign, which might deviate from these standard expectations about them:
```
Pentarctagon, line 441: The "and game engine" should be removed. It's not useful information to the large majority of people who'd be reading this help entry.

CelticMinstrel: It also kinda sounds redundant - the game engine is the implementation of the game rules, so it's almost like you're saying the same thing twice?

stevecotton: The information that I want to convey here is that units with canrecruit=yes alway have this exact ellipse (or its noZoC variant), and AFAIK WML can't override that with a different shape.

Pentarctagon: The point being though that people who would want to know that information won't be reading the in-game help page to find it, they'd look on the wiki.

Pentarctagon (approved):

PR #5224, opened 2020-10-23
Bump minimum gcc requirement to 5.4.

Pentarctagon:


Pentarctagon:
Mostly opening this just to get a travis build done, but also it looks like there are a couple defines that can be cleaned up from `global.hpp`.

PR #5226, opened 2020-10-24
campaignd: Implement basic command line interface

shikadiqueen:
Adds a basic command line interface to campaignd implementing options to change logging configuration and set the server dir and config file path. This uses Boost.program_options, so some changes to the SCons build recipe were required (CMake links wesnothd and campaignd against Boost.program_options regardless already).

This changes campaignd's startup behaviour slightly so that campaignd ensures that the directory containing the config file exists before doing anything else. Normally this wouldn't be noticeable since the default behaviour continues to be to use `./server.cfg` (pwd should exist, right?), but it's important in the event that the user passes `-c` with a nonexistent path.

Additionally, it moves the argv repackaging logic out of wesnoth.cpp, in particular all the Windows-specific code. I would like to make wesnothd use it as well later since it presumably also incorrectly assumes `argv` contains UTF-8 data on Windows like the game proper used to do.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -372,6 +372,7 @@ if env["prereqs"]:
         conf.CheckBoost("iostreams", require_version = boost_version) & \
         conf.CheckBoostIostreamsGZip() & \
         conf.CheckBoostIostreamsBZip2() & \
+        conf.CheckBoost("program_options", require_version = boost_version) & \
         conf.CheckBoost("random", require_version = boost_version) & \
         conf.CheckBoost("smart_ptr", header_only = True) & \
 	CheckAsio(conf) & \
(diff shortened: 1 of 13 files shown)


shikadiqueen:
Requesting review from @gfgtdf for the code involving the filesystem API, as well as `read_argv()` in src/commandline_argv.cpp.

I'd also like to know if he had a particular reason not to use [`CommandLineToArgvW()`](https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw) back in the day instead of hand-rolling the Windows command line parser.

shikadiqueen:
Since it's been a while and the changes should *not* result in any behaviour changes for the game, I'm going to merge this as is. Hopefully I can check with gfgtdf about using `CommandLineToArgvW()` at a later time.

PR #5227, opened 2020-10-24
Some schema fixes

CelticMinstrel:
The commit descriptions should mostly be self-explanatory; I'm only opening this as a pull request because it adds something to the Travis task. This can be merged as soon as Travis passes (assuming of course that it _does_ pass).

### DIFF
--- a/data/schema/units/specials.cfg
+++ b/data/schema/units/specials.cfg
@@ -19,6 +19,27 @@
 	{FILTER_TAG "filter_defender" unit {FILTER_TAG "filter_weapon" weapon ()}}
 	{WML_MERGE_KEYS}
 [/tag]
+# A few specials inheriting from ~generic~ are included here so that unit abilities can then inherit from them.
+[tag]
+	name="firststrike"
+	max=infinite
+	super="units/unit_type/attack/specials/~generic~"
+[/tag]
+[tag]
+	name="poison"
+	max=infinite
+	super="units/unit_type/attack/specials/~generic~"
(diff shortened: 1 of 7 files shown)


src/serialization/schema_validator.hpp
```diff
@@ -112,20 +111,21 @@ class schema_validator : public abstract_validator
 		}
 	};
 
+	/** Controls the way to print errors. */
+	bool create_exceptions_;
```
stevecotton, line 115: Needs a Wreorder warning to be fixed in the .cpp file

data/schema/units/specials.cfg
```diff
@@ -19,6 +19,27 @@
 	{FILTER_TAG "filter_defender" unit {FILTER_TAG "filter_weapon" weapon ()}}
 	{WML_MERGE_KEYS}
 [/tag]
+# A few specials inheriting from ~generic~ are included here so that unit abilities can then inherit from them.
```
stevecotton, line 22: This sounds good, but the corresponding changes to the abilities' validation seems to be missing.

CelticMinstrel: These were added because ability tags were already referencing them as super-tags, but they didn't exist, so that gave an error in the schema self-validator.

src/serialization/schema_validator.hpp
CelticMinstrel: Fixed in 38c099d.

PR #5229, opened 2020-10-24
LoW S07: Revise the dialogue, as the dwarves don't turn up until the end

stevecotton:
Change dialogue to address the bug reported in https://r.wesnoth.org/t53438

I'm expecting this to need some changes, but it seemed sense to start with a PR instead of an issue.

### DIFF
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter2/07_Elves_Last_Stand.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter2/07_Elves_Last_Stand.cfg
@@ -434,6 +434,20 @@
 
     ### OLURF ###
     [event]
+        name=turn 7
+
+        # Hint to the player that the dwarves aren't going to turn up until the end
+        [message]
+            id=Landar
+            message= _ "Those dwarves must still be protecting our gold — on their land."
+        [/message]
+        [message]
+            id=Kalenz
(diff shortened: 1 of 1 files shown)


joeljpa:
Thanks for putting this on my behalf once again.

I think maybe it'll also be good to have Kalenz or Landar complain or remark in the middle of the battle about the dwarves' non-arrival? Isn't it is implied in the previous campaign that Olurf's being paid to come for this very battle. A regular first-time player will be anticipating their arrival due to this and it'll help clarify that.

Generic dialogue examples - maybe like at early game "Where's Olurf?" to middle-late game "Look's like he won't be making it, Elves fight on!"

stevecotton:
> I think maybe it'll also be good to have Kalenz or Landar complain or remark in the middle of the battle about the dwarves' non-arrival?

I've added some on turn 7, but feel it's fairly poor dialogue:
* Landar: Those dwarves must still be protecting our gold — on their land.
* Kalenz: Yes, it seems we must defend the Kalian alone.

I chose turn 7 because it's the afternoon after the start of battle, was wondering about commenting that it's already past noon.

Another option would be to have a troll run in from the north, reporting that the dwarves are approaching. I like that more than this, but it would change the whole reason why Olurf's troops are delayed, and probably need other dialogue to be revised.

nemaara (approved):

PR #5231, opened 2020-10-24
boost::regex to std::regex take two

Vultraz:
Testing what Travis does with this

### DIFF
--- a/src/units/types.cpp
+++ b/src/units/types.cpp
@@ -32,7 +32,7 @@
 #include "gui/auxiliary/typed_formula.hpp"
 #include "gui/dialogs/loading_screen.hpp"
 
-#include <boost/regex.hpp>
+#include <regex>
 #include <boost/range/algorithm_ext/erase.hpp>
 
 #include <array>
@@ -870,7 +870,7 @@ void throw_base_unit_recursion_error(const std::vector<std::string>& base_tree, 
 	throw config::error(ss.str());
 }
 
-const boost::regex fai_identifier("[a-zA-Z_]+");
+const std::regex fai_identifier("[a-zA-Z_]+");
 
 template<typename MoveT>
 void patch_movetype(
@@ -887,8 +887,8 @@ void patch_movetype(
 	gui2::typed_formula<int> formula(formula_str);
 	wfl::map_formula_callable original;
 
-	boost::sregex_iterator m(formula_str.begin(), formula_str.end(), fai_identifier);
-	for(const boost::sregex_iterator::value_type& p : std::make_pair(m, boost::sregex_iterator())) {
+	std::sregex_iterator m(formula_str.begin(), formula_str.end(), fai_identifier);
+	for(const std::sregex_iterator::value_type& p : std::make_pair(m, std::sregex_iterator())) {
 		const std::string var_name = p.str();
 
 		wfl::variant val(original_cfg[var_name].to_int(default_val));
--- a/src/ai/composite/component.cpp
+++ b/src/ai/composite/component.cpp
@@ -26,7 +26,7 @@
 
 #include "ai/formula/ai.hpp"
 
-#include <boost/regex.hpp>
+#include <regex>
 
 namespace pathfind {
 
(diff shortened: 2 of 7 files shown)


CelticMinstrel:
It looks like the chief reason why `std::regex` can't just be dropped in as a replacement for `boost::regex` comes down to the dot-all flag - `std::regex` simply doesn't support it, and we _rely_ on it in several places. The dot-all flag simply controls whether the `.` can match newline characters - normally it doesn't, but with dot-all enabled, it does.

So, I'll go over each instance one at a time…

* The AI case is splitting up a component path of the form `ɑ[i].b[j].c[k]`. It's hard to be sure since the actual regex is very opaque and confusing, but I wouldn't anticipate any trouble; I'd still recommend double-checking that that code still works, though.
* It's hard to tell with the map_context stuff, but I _suspect_ that might be taking the entire contents of a file as a string and searching for various strings in it. If that's the case, it'll definitely be prone to the dot-all flag. Most of those regexes look like they expect to match a single line, but the first one looking for `map_data` would need to match a map directly embedded into the scenario file, which will fail without dot-all. The macro case could theoretically also fail without dot-all, though I _think_ it's looking for a file inclusion rather than an actual macro, so in that case it would be fine.
* Also in the map_context file, changing `match_dot_not_null` to `match_not_null` seems suspicious. The `match_not_null` flag means that an empty sequence doesn't match, but `match_dot_not_null` sounds like it's supposed to be a flag preventing the `.` from matching a null character? I don't know for sure though.
* The regex in the location_palette is trivial and would pose no issue.
* The schema not only relies on dot-all but also takes the regex from essentially user input. This makes it the biggest sticking point, as it's the only place where it would be tricky to simply modify the regex to avoid relying on dot-all.
* The WFL identifier regex in unit types is trivial and would pose no issue.

Other than the schema, it would probably be possible to make it work with `std::regex` instead of `boost::regex` without too much trouble - the regexes themselves would need some adjusting. With the schema though, there are a _lot_ of regexes that you'd need to go through, and it's certainly going to make them uglier and harder to understand. It's really unfortunate that `std::regex` didn't include a dot-all flag, but… without that, I think we can't really use it for anything complicated. And using `std::regex` in some places and `boost::regex` in others would just be confusing and probably cause more problems than it solves.

Pentarctagon:
It looks like in C++17 there's a [multiline option](https://en.cppreference.com/w/cpp/regex/syntax_option_type) that might help?  The description there isn't very useful though.

CelticMinstrel:
Sorry, no - multiline is a completely different option. It governs the meaning of `$` and `^`, not of `.`. It's good that C++17 added it, but unfortunately it won't help here.

soliton-:
Well, for the places where it's necessary we could implement the option ourselves by replacing dots with `(.|\\r|\\n)` or so. That would need some care to not replace it if the dot is escaped or in a bracket expression etc...

It seems like it's not really worth it to switch to std::regex for us right now.

src/units/types.cpp
```diff
@@ -887,8 +887,8 @@ void patch_movetype(
 	gui2::typed_formula<int> formula(formula_str);
 	wfl::map_formula_callable original;
 
-	boost::sregex_iterator m(formula_str.begin(), formula_str.end(), fai_identifier);
-	for(const boost::sregex_iterator::value_type& p : std::make_pair(m, boost::sregex_iterator())) {
+	std::sregex_iterator m(formula_str.begin(), formula_str.end(), fai_identifier);
+	for(const std::sregex_iterator::value_type& p : std::make_pair(m, std::sregex_iterator())) {
```
stevecotton, line 891: I'm going to make a merge-conflict to your PR here, but the resolution is simply going to be removing this file from the PR.

Commit d922914a2042976cf17c9e5322b59d5c52b4eb95 took two bits of code, neither of which needed a regex, and combined them into a single block that did need a regex. Part of fixing #5308 will be reverting that.

Pentarctagon:
I'm going to go ahead and close this, both since it's fairly old by now as well as the discovery during #6291 that std::regex's implementation for CJK languages is currently so amazingly bad (at least for some compilers) that's it's essentially unusable.

PR #5234, opened 2020-10-25
Fauna Revisited

doofus-01:
Some revisions to the recently committed fauna.  Mostly to include graphics work, but some minor conceptual changes as well.
 

- [x] stoat attack animations and portrait
- [x] bear portrait
- [x] icemonax attack animations and portraits
- [x] falcon attack animation and portrait


doofus-01:
There's still more that could be done, but I think it's better to get what's completed merged.  There will be more efforts in the future.  I'll wait for Travis CI before merging.

doofus-01:
Greater Icemonax still needs claws attack anim, but I'll leave that for the future.

PR #5239, opened 2020-10-26
Extent special_id-active and special_type_active to abilities

newfrenchy83:
I know, you said many time to use command line instead of create new PR, but like my net connection is low(1Mb/s) i do not choice 
for resolve https://github.com/wesnoth/wesnoth/issues/4840 with the choice of make the implementation of special_tag_active call the special bool function with a bonuse

### DIFF
--- a/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
+++ b/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
@@ -5,7 +5,6 @@
 # - through [attacks]
 # - through [effect] increase_attacks
 
-{GENERIC_UNIT_TEST "swarm_disables_upgrades_with_abilities_adjacent" (
 #define FILTER_BOB
 [filter_student]
     [filter_weapon]
@@ -37,147 +36,161 @@ affect_allies=yes
 [affect_adjacent]
 [/affect_adjacent]
 #enddef
+
+#define ALICE_ATTACK_BOB SPECIAL_ID
+[unit]
+    id=alex
+    name=_"Alex"
+    x,y=12,4
+    type=Elvish Hero
+    side=1
+    [/unit]
+    [unit]
+    id=ben
+    name=_"Ben"
+    x,y=14,3
+    type=Orcish Warrior
+    side=2
+[/unit]
+[modify_unit]
+    [filter]
+    [/filter]
+    max_hitpoints=100
+    hitpoints=100
+    attacks_left=1
+[/modify_unit]
+[object]
+    silent=yes
+    [effect]
+        apply_to=new_ability
+        [abilities]
+			[swarm]
+				swarm_attacks_max=1
+				swarm_attacks_min=1
+				{FILTER_BOB}
+			[/swarm]
+			[attacks]
+				value=10
+				{FILTER_BOB}
+			[/attacks]
+			[attacks]
+				add=13
+				{FILTER_BOB}
+			[/attacks]
+			[damage]
+				value=1
+				{FILTER_BOB}
+			[/damage]
+			[chance_to_hit]
+				value=100
+				{FILTER_BOB}
+			[/chance_to_hit]
+		[/abilities]
+    [/effect]
+    [filter]
+        id=bob
+    [/filter]
+[/object]
+[object]
+    silent=yes
+    [effect]
+        apply_to=new_ability
+        [abilities]
+			[attacks]
+				value=10
+				{FILTER_ALICE}
+			[/attacks]
+			[damage]
+				value=1
+				{FILTER_ALICE}
+			[/damage]
+			[chance_to_hit]
+				{SPECIAL_ID}
+				value=100
+				{FILTER_ALICE}
+			[/chance_to_hit]
+				[/abilities]
+	[/effect]
+    [filter]
+        id=alice
+    [/filter]
+[/object]
+[object]
+    silent=yes
+    [filter]
+        id=bob
+    [/filter]
+    [effect]
+        apply_to=attack
+        increase_attacks=12
+    [/effect]
+[/object]
+
+[store_unit]
+    [filter]
+        id=alice
+    [/filter]
+    variable=a
+    kill=yes
+[/store_unit]
+[store_unit]
+    [filter]
+        id=bob
+    [/filter]
+    variable=b
+[/store_unit]
+[unstore_unit]
+    variable=a
+    find_vacant=yes
+    x,y=$b.x,$b.y
+[/unstore_unit]
+[store_unit]
+    [filter]
+        id=alice
+    [/filter]
+    variable=a
+[/store_unit]
+
+[do_command]
+    [attack]
+        weapon=0
+        defender_weapon=0
+        [source]
+            x,y=$a.x,$a.y
+        [/source]
+        [destination]
+            x,y=$b.x,$b.y
+        [/destination]
+    [/attack]
+[/do_command]
+[store_unit]
+    [filter]
+        id=alice
+    [/filter]
+    variable=a
+[/store_unit]
+[store_unit]
+    [filter]
+        id=bob
+    [/filter]
+    variable=b
+[/store_unit]
+{ASSERT ({VARIABLE_CONDITIONAL a.hitpoints equals 99})}
+{ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 90})}
+#enddef
+
+{GENERIC_UNIT_TEST "swarm_disables_upgrades_with_abilities_adjacent" (
     [event]
         name=start
-        [unit]
-        id=alex
-        name=_"Alex"
-        x,y=12,4
-        type=Elvish Hero
-        side=1
-        [/unit]
-        [unit]
-        id=ben
-        name=_"Ben"
-        x,y=14,3
-        type=Orcish Warrior
-        side=2
-        [/unit]
-        [modify_unit]
-            [filter]
-            [/filter]
-            max_hitpoints=100
-            hitpoints=100
-            attacks_left=1
-        [/modify_unit]
-        [object]
-            silent=yes
-            [effect]
-                apply_to=new_ability
-                [abilities]
-                    [swarm]
-                        swarm_attacks_max=1
-                        swarm_attacks_min=1
-                        {FILTER_BOB}
-                    [/swarm]
-                    [attacks]
-                        value=10
-                        {FILTER_BOB}
-                    [/attacks]
-                    [attacks]
-                        add=13
-                        {FILTER_BOB}
-                    [/attacks]
-                    [damage]
-                        value=1
-                        {FILTER_BOB}
-                    [/damage]
-                    [chance_to_hit]
-                        value=100
-                        {FILTER_BOB}
-                    [/chance_to_hit]
-                [/abilities]
-            [/effect]
-            [filter]
-                id=ben
-            [/filter]
-        [/object]
-        [object]
-            silent=yes
-            [effect]
-                apply_to=new_ability
-                [abilities]
-                    [attacks]
-                        value=10
-                        {FILTER_ALICE}
-                    [/attacks]
-                    [damage]
-                        value=1
-                        {FILTER_ALICE}
-                    [/damage]
-                    [chance_to_hit]
-                        id=test_cth
-                        value=100
-                        {FILTER_ALICE}
-                    [/chance_to_hit]
-                [/abilities]
-            [/effect]
-            [filter]
-                id=alex
-            [/filter]
-        [/object]
-        [object]
-            silent=yes
-            [filter]
-                id=bob
-            [/filter]
-            [effect]
-                apply_to=attack
-                increase_attacks=12
-            [/effect]
-        [/object]
-
-        [store_unit]
-            [filter]
-                id=alice
-            [/filter]
-            variable=a
-            kill=yes
-        [/store_unit]
-        [store_unit]
-            [filter]
-                id=bob
-            [/filter]
-            variable=b
-        [/store_unit]
-        [unstore_unit]
-            variable=a
-            find_vacant=yes
-            x,y=12,3
-        [/unstore_unit]
-        [store_unit]
-            [filter]
-                id=alice
-            [/filter]
-            variable=a
-        [/store_unit]
-
-        [do_command]
-            [attack]
-                weapon=0
-                defender_weapon=0
-                [source]
-                    x,y=$a.x,$a.y
-                [/source]
-                [destination]
-                    x,y=$b.x,$b.y
-                [/destination]
-            [/attack]
-        [/do_command]
-        [store_unit]
-            [filter]
-                id=alice
-            [/filter]
-            variable=a
-        [/store_unit]
-        [store_unit]
-            [filter]
-                id=bob
-            [/filter]
-            variable=b
-        [/store_unit]
-        {ASSERT ({VARIABLE_CONDITIONAL a.hitpoints equals 99})}
-        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 90})}
+        {ALICE_ATTACK_BOB id=test_cth}
         {SUCCEED}
+    [/event]
+)}
+
+{GENERIC_UNIT_TEST "swarm_disables_upgrades_with_abilities_adjacent_fail" (
+    [event]
+        name=start
+        {ALICE_ATTACK_BOB ()}
+        {FAIL}
     [/event]
 )}
--- a/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
+++ b/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
@@ -1,0 +1,183 @@
+# This unit test defines a WML object based implementation of the "unupgradable" ability
+# https://github.com/ProditorMagnus/Ageless-for-1-14/blob/52c1eaf31722bb58046a1b459d4f29daa8d88487/data/general_data/weapon_specials/unupgradable.cfg
+# and checks that it works. What is being tested here is that
+# [swarm] with swarm_attacks_max=swarm_attacks_min prevents strike changes
+# - through [attacks]
+# - through [effect] increase_attacks
+
+{GENERIC_UNIT_TEST "swarm_disables_upgrades_with_abilities_adjacent" (
+#define FILTER_BOB
+[filter_student]
+    [filter_weapon]
+        type=blade
+    [/filter_weapon]
+[/filter_student]
+[filter_opponent]
+    race=elf
+    [filter_weapon]
+        special_id_active=test_cth #for testing if special_id_active work with abilities used like weapons
+    [/filter_weapon]
+[/filter_opponent]
+affect_self=no
+affect_allies=yes
+[affect_adjacent]
+[/affect_adjacent]
+#enddef
+#define FILTER_ALICE
+[filter_student]
+    [filter_weapon]
+        type=blade,pierce
+    [/filter_weapon]
+[/filter_student]
+[filter_opponent]
+    race=orc
+[/filter_opponent]
+affect_self=no
+affect_allies=yes
+[affect_adjacent]
+[/affect_adjacent]
+#enddef
+    [event]
+        name=start
+        [unit]
+        id=alex
+        name=_"Alex"
+        x,y=12,4
+        type=Elvish Hero
+        side=1
+        [/unit]
+        [unit]
+        id=ben
+        name=_"Ben"
+        x,y=14,4
+        type=Orcish Warrior
+        side=2
+        [/unit]
+        [modify_unit]
+            [filter]
+            [/filter]
+            max_hitpoints=100
+            hitpoints=100
+            attacks_left=1
+        [/modify_unit]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=new_ability
+                [abilities]
+                    [swarm]
+                        swarm_attacks_max=1
+                        swarm_attacks_min=1
+                        {FILTER_BOB}
+                    [/swarm]
+                    [attacks]
+                        value=10
+                        {FILTER_BOB}
+                    [/attacks]
+                    [attacks]
+                        add=13
+                        {FILTER_BOB}
+                    [/attacks]
+                    [damage]
+                        value=1
+                        {FILTER_BOB}
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                        {FILTER_BOB}
+                    [/chance_to_hit]
+                [/abilities]
+            [/effect]
+            [filter]
+                id=ben
+            [/filter]
+        [/object]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=new_ability
+                [abilities]
+                    [attacks]
+                        value=10
+                        {FILTER_ALICE}
+                    [/attacks]
+                    [damage]
+                        value=1
+                        {FILTER_ALICE}
+                    [/damage]
+                    [chance_to_hit]
+                        id=test_cth
+                        value=100
+                        {FILTER_ALICE}
+                    [/chance_to_hit]
+                [/abilities]
+            [/effect]
+            [filter]
+                id=alex
+            [/filter]
+        [/object]
+        [object]
+            silent=yes
+            [filter]
+                id=bob
+            [/filter]
+            [effect]
+                apply_to=attack
+                increase_attacks=12
+            [/effect]
+        [/object]
+
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+            kill=yes
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        [unstore_unit]
+            variable=a
+            find_vacant=yes
+            x,y=12,3
+        [/unstore_unit]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+
+        [do_command]
+            [attack]
+                weapon=0
+                defender_weapon=0
+                [source]
+                    x,y=$a.x,$a.y
+                [/source]
+                [destination]
+                    x,y=$b.x,$b.y
+                [/destination]
+            [/attack]
+        [/do_command]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        {ASSERT ({VARIABLE_CONDITIONAL a.hitpoints equals 99})}
+        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 90})}
+        {SUCCEED}
+    [/event]
+)}
--- a/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
+++ b/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
@@ -49,7 +49,7 @@ affect_allies=yes
         [unit]
         id=ben
         name=_"Ben"
-        x,y=14,4
+        x,y=14,3
         type=Orcish Warrior
         side=2
         [/unit]
--- a/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
+++ b/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
@@ -194,3 +194,7 @@ affect_allies=yes
         {FAIL}
     [/event]
 )}
+
+#undef FILTER_BOB
+#undef FILTER_ALICE
+#undef ALICE_ATTACK_BOB
--- a/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
+++ b/data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
@@ -64,28 +64,28 @@ affect_allies=yes
     [effect]
         apply_to=new_ability
         [abilities]
-			[swarm]
-				swarm_attacks_max=1
-				swarm_attacks_min=1
-				{FILTER_BOB}
-			[/swarm]
-			[attacks]
-				value=10
-				{FILTER_BOB}
-			[/attacks]
-			[attacks]
-				add=13
-				{FILTER_BOB}
-			[/attacks]
-			[damage]
-				value=1
-				{FILTER_BOB}
-			[/damage]
-			[chance_to_hit]
-				value=100
-				{FILTER_BOB}
-			[/chance_to_hit]
-		[/abilities]
+            [swarm]
+                swarm_attacks_max=1
+                swarm_attacks_min=1
+                {FILTER_BOB}
+            [/swarm]
+            [attacks]
+                value=10
+                {FILTER_BOB}
+            [/attacks]
+            [attacks]
+                add=13
+                {FILTER_BOB}
+            [/attacks]
+            [damage]
+                value=1
+                {FILTER_BOB}
+            [/damage]
+            [chance_to_hit]
+                value=100
+                {FILTER_BOB}
+            [/chance_to_hit]
+        [/abilities]
     [/effect]
     [filter]
         id=bob
@@ -96,21 +96,21 @@ affect_allies=yes
     [effect]
         apply_to=new_ability
         [abilities]
-			[attacks]
-				value=10
-				{FILTER_ALICE}
-			[/attacks]
-			[damage]
-				value=1
-				{FILTER_ALICE}
-			[/damage]
-			[chance_to_hit]
-				{SPECIAL_ID}
-				value=100
-				{FILTER_ALICE}
-			[/chance_to_hit]
-				[/abilities]
-	[/effect]
+            [attacks]
+                value=10
+                {FILTER_ALICE}
+            [/attacks]
+            [damage]
+                value=1
+                {FILTER_ALICE}
+            [/damage]
+            [chance_to_hit]
+                id={SPECIAL_ID}
+                value=100
+                {FILTER_ALICE}
+            [/chance_to_hit]
+        [/abilities]
+    [/effect]
     [filter]
         id=alice
     [/filter]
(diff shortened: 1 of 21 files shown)


CelticMinstrel:
I don't think this really helps the underlying issue with the previous PR. In fact it goes right back to keeping specials and abilities of the same type entirely separate.

newfrenchy83:
@CelticMinstrel i see it like an mix of two choice, when special_id_active put directly in [filter_weapon], specials and abilities are both researched, but if attribute pulled inside [filter_specials/abilities] specials and abilities are separate, and like speration will be used less what merged, the WML develloper don't haveto using different attributes names.

CelticMinstrel:
I still don't think it makes sense to separate these two concepts at all. From my perspective, they are both the exact same thing, just coming from a different source.

newfrenchy83:
If discriminate the source can be useful, i know also what this cases can be rare, i will remove filter_specials/abilities for now.

kabachuha:
@newfrenchy83 , please, check my comment on #5023 . While you can get away with the creation of copied pull-requests — and this pr is evidently a duplicate of #5023 — you're breaking the commit history and the workflow for the others with a heap of unnamed commits (like `Update X.cpp`), commits with mistakes or commits with only one file changed when they are connected by a single purpose. If you are willing to make any significant long-term contributions to wesnoth in the future, you may want to learn how to organize the proper commit history. And if the network is an issue, there are some options that can relieve it as I described it in the review.

stevecotton (changes requested):
I feel that the feature of extra animation isn't worth the added code complexity; it probably needs a refactor of the existing code to simplify it before adding new features.

src/units/attack_type.hpp
```diff
@@ -90,7 +90,8 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 	/// Returns list who contains list_ability and get_specials list for each ability type
 	unit_ability_list get_special_ability(const std::string& ability) const;
 	///return an boolean value for abilities like poison slow firstrike or petrifies
-	bool bool_ability(const std::string& ability) const;
+	bool get_special_ability_bool(const std::string& special, bool special_id=true, bool special_tags=true) const;
+	bool bool_ability(const std::string& ability, bool simple_check=false, bool special_id=true, bool special_tags=true) const;
```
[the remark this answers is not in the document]
newfrenchy83: @stevecotton Ok for simple_check, but for special_id and special_tags, i prefer what it doing later object of sperarate PR because parameter exist also in get_special_bool

stevecotton (commented):
Thanks for fixing the specific issue that I raised, but I feel code complexity is the big issue here. I feel that the feature of extra animation isn't worth the added code complexity; it probably needs a refactor of the existing code to simplify it before adding new features. It's not an issue with your changes, rather it's an issue with the existing code. At the same time, I don't want to become the maintainer of these files, which means that I'm the wrong person to either approve or disapprove.

In terms of maintainability, it has three different getters, all of whose names are some variation on "special ability", all of which take similar arguments, and all of which do something different. This makes for hard-to-read code where mistakes are more likely to slip through code reviews.
* bool bool_ability(const std::string& ability, bool special_id=true, bool special_tags=true) const;
* bool get_special_ability_bool(const std::string& special, bool special_id=true, bool special_tags=true) const;
* bool get_special_bool(const std::string& special, bool simple_check=false, bool special_id=true, bool special_tags=true) const;

newfrenchy83:
@CelticMinstrel like dicriminate source can be useful, like for don't call leading_anim with special_type_active don't filtered, i close this PR

CelticMinstrel:
The filter for a leading animation would use the `ability=` key in filters, so it has nothing to do with this. A leadership `chance_to_hit` is both an ability and a special, so it should match `ability=chance_to_hit` on the unit that owns it and also `special=chance_to_hit` on the unit affected by it.

newfrenchy83:
except what like that ability=chance_to-hit will match ability same if not active, and filter and attribute specific to weapon don't match. with filter, it is unit what is filtered, and don't attack used like in filter_weapon or anim filter_attack.

CelticMinstrel:
If you want to know when it's active, use `ability_active=` instead of `ability=`. If you want to know about the weapon of the supported unit, a `[filter_adjacent]` with `[has_attack]special_active=` should work. (I don't know if it _does_ work, but `special_active=` _should_ only match the weapon being used in an attack, which would mean it never matches any weapon on a unit not involved in the attack; of course, in the case of a `chance_to_hit` ability, it would match _every_ weapon on the supported unit.)

newfrenchy83:
> 
> 
> If you want to know when it's active, use `ability_active=` instead of `ability=`. If you want to know about the weapon of the supported unit, a `[filter_adjacent]` with `[has_attack]special_active=` should work. (I don't know if it _does_ work, but `special_active=` _should_ only match the weapon being used in an attack, which would mean it never matches any weapon on a unit not involved in the attack; of course, in the case of a `chance_to_hit` ability, it would match _every_ weapon on the supported unit.)

[has_attack] is used for filtered the units who _own_ the attack filtered and not the weapon _used_, else i don't have to implemente filter_(second_)weapon tags in leadership and resistance abilities, before to think to implemented specials weapon in abilities tags and same if special_active= could worked, (i don't know), exlude an weapon  special with [has_attack] is like exlude all unit who own the weapon, same if has other attack don't compatible with the filters.
ability_active don't exist, ability_type_active only, and [filter_student/opponent/etc],apply_to and active_on not used by this attribute in [filter]. For me, the actual attribute are incomplete and more complicated for the wml develloper what to include [filter_specials/abilities] tags inside [filter_weapon] or animate [filter_attack] for discriminate the source of weapon specials  and not the unit who possessed this special or abilitie

ex: first case, with and without [filter_specials]
[filter_weapon]
[not]
special_type_active=filtered , only type other what 'filtered'(abilires or specials tags) matched.

[filter_weapon]
[not]
[filter_sepcials]
special_type_active=filtered, only type other what 'filtered' and inside [specials] matched, but if a unit who has attack 'filtered' but use another attack, this attack match.

second case, with attribute existent

[filter_weapon]
[not]
special_type_active=filtered , only type other what 'filtered'(abilires or specials tags) matched.

and for exlude specials only

[filter_adjacent]
[filter]
[has_attack]
[not]
special_type_active=filtered (special_active work?), unit who own 'filtered' don't match same if use another attack

or for exlude abilities
[filter]
[not]
ability_type_active='filtered' if unit or weapon don't matche [filter_student/etc] or apply_to or active_on don't matches, 'filtered' is 'active' then it mustn't be active.

CelticMinstrel:
> [has_attack] is used for filtered the units who own the attack filtered and not the weapon _used_

Yes, but `special_active` is only ever true on a weapon that is being used, so placing it in `[has_attack]` should exclude any weapons not actually being used in the attack.

I think it's a good idea to add tags similar to `[filter_specials]` and `[filter_abilities]`, but `[filter_abilities]` belongs in a unit filter, not a weapon filter. Abilities have nothing to do with weapons; weapons only have specials. If a unit is affected by a leadership chance_to_hit special, then `special_type_active=chance_to_hit` should be true for all that unit's weapons, and `ability_type_active=chance_to_hit` should _only_ be true for the unit with the leadership ability. The leader has the ability active, and the student has the special active; these are two different things.

I think a `[filter_specials]` tag would need three attributes: `active=yes` (or no), `type=`, and `id=`. If both type and ID are present, then both must match the same special. The `[filter_abilities]` tag would be the same but in a unit filter, rather than a weapon filter.

newfrenchy83:
but special_active don't work inside has_attack, when i test aura abilities who supported unitsd with poison weapon only with that, the ability supported also units who don't have poison specials, implemented filter_abilities inside unit filter will be same effect.

newfrenchy83:
replace [filter_weapon] by [filter_abilities] inside [filter_student/opponent/etc], ok, i think could doing this, but not in egular [filter]

newfrenchy83:
and if i cannot put [filter_abilities] inside [filter_attack] animation, then i cannot use for discriminate source for leading_anim.

newfrenchy83:
> I think it's a good idea to add tags similar to `[filter_specials]` and `[filter_abilities]`, but `[filter_abilities]` belongs in a unit filter, not a weapon filter. Abilities have nothing to do with weapons; weapons only have specials. If a unit is affected by a leadership chance_to_hit special, then `special_type_active=chance_to_hit` should be true for all that unit's weapons, and `ability_type_active=chance_to_hit` should _only_ be true for the unit with the leadership ability. The leader has the ability active, and the student has the special active; these are two different things.
> 
true but it can said alos student has special active, and support of ability active, rename [filter_abilities] to [filter_support] will be more sense and more simple what implemente tags inside [filter]

CelticMinstrel:
> but special_active don't work inside has_attack

Fair enough. I don't know if it would be tricky to make it work in this context.

> and if i cannot put [filter_abilities] inside [filter_attack] animation, then i cannot use for discriminate source for leading_anim.

This sounds like something possibly important, but I'm not sure what you mean. Can you try to explain in more detail?

newfrenchy83:
for imlemente [filter_abilities] outside [filter_weapon/attack] i do create an function equivalent to attack_type::matches_filter and static simple_filter only for atttibutes inside [filter_abilities] and implemente this in all files where matches_filter will be already implemented, this is more complicatd what to implement directly inside [filter_weapon/attack] and use already implemented attact-type::matches_filter, i repeat this, rename [filter_specials/abilities] to [internal/external_special] is more simple and more readable ,tags inside filter_weapon/attack for me point special who support attack( owned by attack like regular specials or owned outside like special inside abilities) what [filter_abilities] was in same parent what [filter_specials] recall what theses tags assume same function(filter source of special who support student/oponent/etc.

CelticMinstrel:
I didn't understand a word of that, but `[filter_abilities]` has nothing to do with attacks and doesn't belong in an attack filter.

newfrenchy83:
> 
> 
> I didn't understand a word of that, but `[filter_abilities]` has nothing to do with attacks and doesn't belong in an attack filter.

If we follow your reasoning, special_id / type_active should never designate an ability like chance_to_hit leadership, which is not possessed by the attack designated by [filter_weapon] or, this is precisely the subject of my two PRs, So I'm asking you the question, why is it only when I want to use [filter_abilities] to specifically designate special weapons in [abilities] that this is a problem?

For my previous comment, I will try to rephrase:

To implement [filter_abilities] independently of [filter_weapon / attack] in abilities / specials or animation I have to create a function equivalent to attack_type :: matches_filter then implement the function call wherever the original function is located which is more complicated to put implemented and that I am reluctant to do for a simple story of writing WML code. You say that [filter_abilities] has nothing to do with attacks, I'm saying that as long as abilities other than [leadership] grant the same attack bonuses as special weapons this distinction is no longer as absolute, [filter_abilities] special_type_active = is equivalent to special_type_active but only for specials encoded outside of the attack concerned and if you accept the extension of special_id_type_active to abilities equivalent to special weapons, you should accept the rest otherwise you are in contradiction with the reasoning that [abilities] has nothing to do with attacks and that ability filters cannot be implemented in [filter_weapon], since special_id / type_active becomes an ability filter in addition to an ability filter. special.

Finally, if it's the terms [filter_specials] and [filter_abilities] that are confusing, renaming them to [filter_internal] and [filter_external] would perhaps be clearer considering that the chance_to_hit implemented in [abilities] is a special encoded outside of [attack] but which counts as much as [specials] in [attack].

I hope I was clearer this time.

CelticMinstrel:
Somewhat clearer. I'm still not convinced that `[filter_abilities]` in an attack is useful, though, and it seems to me that you're approaching the problem you're facing from the wrong angle. What is the problem you're trying to solve by adding a `[filter_abilities]` or `wp_ability_type=` or whatever you wish to call it? I would like to see a complete example… either an example that doesn't work without these changes and (in your opinion) needs these changes to work, or an example of how you'd _use_ these changes to get the filter you need.

newfrenchy83:
A problem that can only be solved by using [filter_abilities], probably not, but one that could be solved more easily, or at least less laborious to use for a wml encoder could be this purely theoretical example:

Imagine a special of the firststrike type designed to only activate against an opponent benefiting from another leadership type firststrike but not the so-called 'traditional' firststrike, if the campaign offers only one, or even two such leaderships. , just list them with special_id_active, since it will be unlikely that their id is the same special firststrike that we know, but if we are dealing with a cracked who multiplies the leadership of this type in his campaign (filtering by distance, type of attack, etc ...), then listing them all in the write-heavy filter, and in this case using [filter_abilities] special_type_active = can become interesting to lighten the wml code of the filter.

The other case that I imagine would be even more improbable but with the abilities, all the special weapons in the core can have an equivalent leadership, in this case how will the coder name them, if he thinks to give charge leadership an id = charge_leadeship then special_id_active can be used directly to differentiate them in [filter_weapon] animation filter_attack, on the other hand if the encoder keeps the same id as for the special weapon loads then use of [filter_specials / abilities] to differentiate them becomes useful .

CelticMinstrel:
Thanks, but I was asking for a code example…

newfrenchy83:
@CelticMinstrel ok, i have not need of of [filter_abilities], because [filter_sepcials] tags must be sufficient, but i want what this [filter_specials] remain.

newfrenchy83:
@CelticMinstrel this PR could be possibly merged before 1.16, please?

Pentarctagon:
Something more general than this PR that I've been wondering for while: what was the original motivation for adding "weapon specials as abilities"?  The end result of giving the unit a "weapon special as an ability" is that it's functionally identical to a unit that has that weapon special applied to every attack it possesses, right?

newfrenchy83:
> 
> 
> Something more general than this PR that I've been wondering for while: what was the original motivation for adding "weapon specials as abilities"? The end result of giving the unit a "weapon special as an ability" is that it's functionally identical to a unit that has that weapon special applied to every attack it possesses, right? So for example, in combat a unit with the firststrike ability works the same way as if all of that unit's attacks had the firststrike weapon special.

As I explained at the time, implementing a special weapon in [abilities] offers two advantages: first, that of being able to affect all the opponent's weapons with apply_to = opponent, even when the 'possessing unit only has melee or ranged type attacks (' formation 'abilities in Utbs), the other is the leadership type interaction (initiative ability in Httt) or terror (' diversion ' abilities in core), with more diverse leadership than just increasing attack power. Before, it would have been necessary to pass a rather heavy wml (or even lua) encoding (see code in the Loti extension), which was quite rebarbative.

All that's missing to make this almost as complete as a real special weapon is expanding the special_id_type_active attributes to the abilities used as weapons.

newfrenchy83:
@Pentarctagon @CelticMinstrel someone could review this PR please?

Pentarctagon:
I'm not at all familiar with this area of the code and there's another thing I want to get in for 1.15.10, so it may be a couple weeks until I can get to this.

Pentarctagon:
Can you add doxygen doc comments to the new functions?  So with a description, `@param`, and `@return`.

newfrenchy83:
In.cpp or hpp files ?

Pentarctagon:
hpp files.

src/units/udisplay.cpp
```diff
@@ -620,6 +621,13 @@ void unit_attack(display * disp, game_board & board,
 	assert(def.valid());
 	unit &defender = *def;
 	int def_hitpoints = defender.hitpoints();
+	const_attack_ptr weapon = attack.shared_from_this();
+	auto ctx = weapon->specials_context(attacker.shared_from_this(), defender.shared_from_this(), a, b, attacking, secondary_attack);
+	std::optional<decltype(ctx)> opp_ctx;
```
Pentarctagon, line 626: What is `ctx`/`opp_ctx` created here?  They don't seem to be referenced anywhere else.

src/units/udisplay.cpp
```diff
@@ -640,33 +648,43 @@ void unit_attack(display * disp, game_board & board,
...
 	animator.add_animation(defender.shared_from_this(), defender_anim, def->get_location(), true, text, {255, 0, 0});
 
-	for(const unit_ability& ability : attacker.get_abilities_weapons("leadership", attack.shared_from_this(), secondary_attack)) {
-		if(ability.teacher_loc == a) {
-			continue;
-		}
-
-		if(ability.teacher_loc == b) {
-			continue;
+	const std::vector<std::string> leader_tags{"leadership", "damage", "chance_to_hit", "berserk", "swarm", "drains", "heal_on_hit", "plague", "slow", "petrifies", "firststrike", "poison"};
```
Pentarctagon, line 660: This set of tags is used in multiple places, so they should be available from a single location rather than having multiple separate lists.

src/units/abilities.cpp
```diff
@@ -1171,14 +1171,207 @@ unit_ability_list attack_type::get_special_ability(const std::string& ability) c
 	return abil_list;
 }
 
-bool attack_type::bool_ability(const std::string& ability) const
+static bool get_ability_children(std::vector<special_match>& tag_result,
```
Pentarctagon, line 1174: Needs doxygen doc comment.

src/units/unit.hpp
```diff
@@ -1649,6 +1649,25 @@ class unit : public std::enable_shared_from_this<unit>
...
+	 * @param tag_name name of ability type checked.
+	 * @param loc location of the unit checked.
+	 * @param from unit adjacent to @a this is checked in case of [affect_adjacent] abilities.
+	 * @param dir direction to research a unit adjacent to @a this.
+	 * @param weapon the attack used by unit checked in this function.
+	 * @param opp_weapon the attack used by opponent to unit checked.
+	 */
+	bool get_adj_ability_bool_weapon(const config& special, const std::string& tag_name, int dir, const map_location& loc, const unit& from, const_attack_ptr weapon=nullptr, const_attack_ptr opp_weapon = nullptr) const;
+
+	bool get_self_ability_bool_weapon(const config& special, const std::string& tag_name, const map_location& loc, const_attack_ptr weapon = nullptr, const_attack_ptr opp_weapon = nullptr) const;
```
Pentarctagon, line 1664: Needs doxygen doc comment.

src/units/unit.hpp
```diff
@@ -1649,6 +1649,25 @@ class unit : public std::enable_shared_from_this<unit>
...
+
+	bool get_self_ability_bool(const config& special, const std::string& tag_name, const map_location& loc) const;
```
Pentarctagon, line 1666: Needs doxygen doc comment.

src/units/unit.hpp
```diff
@@ -1649,6 +1649,25 @@ class unit : public std::enable_shared_from_this<unit>
...
+
+	bool get_adj_ability_bool(const config& special, const std::string& tag_name, int dir, const map_location& loc, const unit& from) const;
```
Pentarctagon, line 1668: Needs doxygen doc comment.

src/units/attack_type.hpp
```diff
@@ -92,8 +92,14 @@ class attack_type : public std::enable_shared_from_this<attack_type>
...
-	/** return an boolean value for abilities like poison slow firstrike or petrifies */
-	bool bool_ability(const std::string& ability) const;
+	/** return an boolean value for abilities like poison slow firstrike or petrifies
+	 * @return True if the special @a special is active.
+	 * @param special The special being checked.
+	 * @param special_id If true, match @a special against the @c id of special tags.
+	 * @param special_tags If true, match @a special against the tag name of special tags.
+	 */
+	bool get_special_ability_bool(const std::string& special, bool special_id=true, bool special_tags=true) const;
+	bool bool_ability(const std::string& ability, bool special_id=true, bool special_tags=true) const;
```
Pentarctagon, line 102: Needs doxygen doc comment.

src/units/attack_type.hpp
```diff
@@ -113,9 +119,60 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 
 	// Configured as a bit field, in case that is useful.
 	enum AFFECTS { AFFECT_SELF=1, AFFECT_OTHER=2, AFFECT_EITHER=3 };
+	bool check_self_abilities(const config& cfg, const std::string& special) const;
```
Pentarctagon, line 122: Needs doxygen doc comment.

src/units/attack_type.hpp
```diff
@@ -113,9 +119,60 @@ class attack_type : public std::enable_shared_from_this<attack_type>
...
+	bool check_adj_abilities(const config& cfg, const std::string& special, int dir, const unit& from) const;
```
Pentarctagon, line 123: Needs doxygen doc comment.

src/units/udisplay.cpp
```diff
@@ -640,33 +648,43 @@ void unit_attack(display * disp, game_board & board,
...
+	if(!leader_tags.empty()) {
```
Pentarctagon, line 661: In what situation could `leader_tags` be empty here?

Pentarctagon:
So for the changes to `udisplay.cpp` lines 660-685, that's to set the leading animation for the attacker, but now for abilities besides leadership, correct?

newfrenchy83:
> 
> 
> So for the changes to `udisplay.cpp` lines 660-685, that's to set the leading animation for the attacker, but now for abilities besides leadership, correct?

yes, but only abilities used like weapon, because can be used like 'leadership' for cth,poison, etc.. instead of simply damge of attack

src/units/udisplay.cpp
newfrenchy83: if an ability is other type of what leadership and other in leader_tags like heals or hides

src/units/udisplay.cpp
newfrenchy83: i can't use weapon_leadership for what abilities used like weapon call [leading_anim] else the condition of activity don't be checked correctly.

src/units/udisplay.cpp
Pentarctagon: Unless I'm completely missing something here, isn't `leader_tags` just a const vector that's always initialized to the same list of values?

Pentarctagon:
> yes, but only abilities used like weapon, because can be used like 'leadership' for cth,poison, etc.. instead of simply damge of attack

And it wouldn't cause any issues to have a unit with leadership for multiple types of weapon special abilities at the same time?

src/units/udisplay.cpp
newfrenchy83: special_id_type_active don't detectspecial for animation without ctx/op_ctx

src/units/udisplay.cpp
newfrenchy83: > 
> 
> > yes, but only abilities used like weapon, because can be used like 'leadership' for cth,poison, etc.. instead of simply damge of attack
> 
> And it wouldn't cause any issues to have a unit with leadership for multiple types of weapon special abilities at the same time?

the reply is in https://github.com/newfrenchy83/wesnoth/pull/61 PR

newfrenchy83:
'And it wouldn't cause any issues to have a unit with leadership for multiple types of weapon special abilities at the same time?'

the reply is in https://github.com/newfrenchy83/wesnoth/pull/61 PR

src/units/attack_type.hpp
newfrenchy83: it is done

src/units/abilities.cpp
```diff
@@ -1171,14 +1171,216 @@ unit_ability_list attack_type::get_special_ability(const std::string& ability) c
...
+}
+/**
+ * Returns whether or not @a *this has a special ability with a tag or id equal to
+ * @a special. the Check is for a special ability
+ * active in the current context (see set_specials_context), including
+ * specials obtained from the opponent's attack.
+ */
+bool attack_type::get_special_ability_bool(const std::string& special, bool special_id, bool special_tags) const
+{
+	static std::set<std::string> included_tags{"damage", "chance_to_hit", "berserk", "swarm", "drains", "heal_on_hit", "plague", "slow", "petrifies", "firststrike", "poison"};
```
Pentarctagon, line 1276: Making it static doesn't change the fact that this set of values is defined in multiple places.  `{"damage", "chance_to_hit", "berserk", "swarm", "drains", "heal_on_hit", "plague", "slow", "petrifies", "firststrike", "poison"}` should appear *once* in all the code rather than in three separate places like it does now.

One other option would be to use MAKE_ENUM to define the possible values and then check for matches with the `string_to_enum(str_val, enum_default)` method.

src/units/abilities.cpp
```diff
@@ -1171,14 +1171,216 @@ unit_ability_list attack_type::get_special_ability(const std::string& ability) c
...
+	const unit_map& units = display::get_singleton()->get_units();
+	assert(display::get_singleton());
```
Pentarctagon, line 1278: The assert should come before you try to use the singleton's output.

Pentarctagon:
Please also add one or more WML unit tests showing that this is working as intended.

Pentarctagon:
For the `get_ability_children` function in abilities.cpp:

Change:
```
if ( get_special_children(tag_result, id_result, parent, id) ) {
	return true;
}
```
to simply `get_special_children(tag_result, id_result, parent, id)`.

Change:
```
if ( get_special_children_id(id_result, parent, id) ) {
	return true;
}
```
to simply `get_special_children_id(id_result, parent, id)`.

Change:
```
if ( get_special_children_tags(tag_result, parent, id) ) {
	return true;
}
```
to simply `get_special_children_tags(tag_result, parent, id)`.

Also change it to return void rather than bool.

---

Those functions will only ever return true if the optional `just_peeking` argument is set to true, which it isn't, in which there's no need for these if-checks (since the methods always return false) nor any need to return a boolean (since it will likewise always be false).

src/units/abilities.cpp
newfrenchy83: MAKE_ENUM? imust define a case for each weapon_type?

Pentarctagon: You would make a NOT_FOUND enum value or similar, then do `if(string_to_enum(str_val, NOT_FOUND) == NOT_FOUND)`

src/units/abilities.cpp
newfrenchy83: it's fixed

newfrenchy83:
> 
> 
> Please also add one or more WML unit tests showing that this is working as intended.

it is done.

src/units/abilities.cpp
newfrenchy83: i declared std::set<std::string> included_tags in unit.hpp instead.

src/units/abilities.cpp
```diff
@@ -1171,14 +1171,197 @@ unit_ability_list attack_type::get_special_ability(const std::string& ability) c
...
-	bool abil_bool = get_special_bool(ability);
-	unit_ability_list abil = list_ability(ability);
-	if(!abil.empty()) {
-		abil_bool = true;
+	if(tag_name == "leadership" && leader_bool){
+		if((*u).get_self_ability_bool_weapon(special, tag_name, loc, self_attack, other_attack)) {
+			return true;
+		}
+	}
+	if((*u).checking_tags().count(tag_name) != 0){
```
Pentarctagon, line 1231: This check isn't needed since it can never be 0.

src/units/abilities.cpp
```diff
@@ -1171,14 +1171,197 @@ unit_ability_list attack_type::get_special_ability(const std::string& ability) c
...
+
+bool unit::get_adj_ability_bool(const config& special, const std::string& tag_name, int dir, const map_location& loc, const unit& from) const
+{
+	const auto adjacent = get_adjacent_tiles(loc);
+	return (affects_side(special, side(), from.side()) && from.ability_active(tag_name, special, adjacent[dir]) && ability_affects_adjacent(tag_name,  special, dir, loc, from));
+}
+
+bool unit::get_self_ability_bool_weapon(const config& special, const std::string& tag_name, const map_location& loc, const_attack_ptr weapon, const_attack_ptr opp_weapon) const
+{
+	return (get_self_ability_bool(special, tag_name, loc) && ability_affects_weapon(special, weapon, false) && ability_affects_weapon(special, opp_weapon, true));
```
Pentarctagon, line 1211: Why does this need to check both `weapon` and `opp_weapon` for the special?

newfrenchy83: this sub function is used for cheking [leadership] in special_id/type_active or this ability can't checke opponent but student and teacher only, and with [filter_weapon] and [filter_second_weapon], the weapon of student and opponent, then yes it need check weapon and opp_weapon for checking acitivtie of [leadership] abilities (filter diffrent of weapon special and abilities used like weapon

Pentarctagon: I'm not sure I understand.  Wouldn't most cases be that the unit attacking and the unit defending don't have the same ability affecting their weapons?

newfrenchy83: What I'm trying to say is that since leadership can filter the weapons of the unit under leadership as well as those of the adversary, special_id / type_active must be able to differentiate between active leadership and one that does not. would not be because one of the filtered weapons does not match like for an special active or not with [filter_student] or [filter_opponent]

Pentarctagon: Alright, so for `unit::get_self_ability_bool_weapon`, it only gets called when the tag being checked is [leadership] and when `leader_bool` is set to true, which as of right now only happens in the `if(self_)` part of `attack_type::get_special_ability_bool`.  So the [leadership] check would only happen for the attacker.

Then for the `unit::get_self_ability_bool_weapon` method itself, it checks:
* `get_self_ability_bool(special, tag_name, loc)` - does the unit have this ability active right now and is the unit affected by it?
* `ability_affects_weapon(special, weapon, false)` - does this ability affect the attacker's weapon (by checking [filter_weapon])?
* `ability_affects_weapon(special, opp_weapon, true)` - does this ability affect the defender's weapon (by checking [filter_second_weapon])?

Is that correct?

newfrenchy83: > 
> 
> Alright, so for `unit::get_self_ability_bool_weapon`, it only gets called when the tag being checked is [leadership] and when `leader_bool` is set to true, which as of right now only happens in the `if(self_)` part of `attack_type::get_special_ability_bool`. So the [leadership] check would only happen for the primary unit.
> 
> Then for the `unit::get_self_ability_bool_weapon` method itself, it checks:
> 
>     * `get_self_ability_bool(special, tag_name, loc)` - does the unit have this ability active right now and is the unit affected by it?
> 
>     * `ability_affects_weapon(special, weapon, false)` - does this ability affect the primary unit's weapon (by checking [filter_weapon])?
> 
>     * `ability_affects_weapon(special, opp_weapon, true)` - does this ability affect the secondary unit's weapon (by checking [filter_second_weapon])?
> 
> 
> Is that correct?

1. yes
2.  yes
3. yes

Pentarctagon: Alright, so, what would be an example of a situation where the primary unit's leadership ability affects both the primary unit's weapon and the secondary unit's weapon?

newfrenchy83: i make error for 3, leadership don't affect secondary weapon, but is conditonnal by this presence, but this the always the unit student who is affected by leadership. the first example is a magical leadership who work for magical weapon only, (use [filter_weapon]special_id_active=magical),
the second example is leadership who don't work if opponent use fire cold or weapon who afect the teacher(use [filter_second_weapon][not]type=cold). the ability_affect_weapon name can wrong the reader and all excuses for that.

Pentarctagon: So given the following definitions:
* Teacher (unit with leadership ability)
* Primary unit (the unit attacking)
* Secondary unit (the unit defending)

The first example is of a teacher that has a leadership ability that only gives its bonuses to the primary unit's attacks that have the Magical weapon special.

The second example is of a teacher that has a leadership ability that only gives its bonuses to the primary unit's attacks that have the Magical weapon special, and only if the secondary unit isn't using an attack that has the Cold or Fire damage type.

Is that correct?

newfrenchy83: Correct, but i didn't said what [filter_weapon]special_id_active=magical was implemented in second example

Pentarctagon: Alright.  I think it'd be good for there to be new, separate WML unit tests that check the correctness of different cases where this can be used.

newfrenchy83: you must stoping to use attacker and defender for define self/student and opponent in weapon special, attacker is the unit who begin the fight and apply_to=attacker aplied to unit who begin fight, student or opponent.
primary unit:(the unit student of Teacher, in leadership weapon filter by [filter_weapon])
secondary unit (the unit opponent to student weapon filtered by [filter_second_weapon] in leadership abilities)
with leadership ever active in offense and defense, using attacker and defender is confusing.

newfrenchy83: test added for abilities affect_adjacent with or without [leadership] abilities

Pentarctagon:
I *think* I'm good with this.  The only other thing I'd like to have added would be some negative test cases for non-leadership and leadership situations, so for example filtering for `special_id_active=test_cth` when the opponent doesn't have that active.

data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
```diff
@@ -0,0 +1,199 @@
...
+    [event]
+        name=start
+        {ALICE_ATTACK_BOB test_cth}
+    [/event]
+)}
+
+{GENERIC_UNIT_TEST "swarm_disables_upgrades_with_abilities_adjacent_fail" (
+    [event]
+        name=start
+        {ALICE_ATTACK_BOB test_cth_fail}
```
stevecotton, line 193: I think this is testing that an ability with a different name doesn't get matched by the filter for `test_cth`. Please move the final `{ASSERT ...}` and `{SUCCEED}` outside of the macro, and test exactly the values that are expected when the attack does get upgraded instead of testing a `fail`.

* That makes it a more rigorous test, checking the expected result exactly.
* It makes it clearer what has broken if it fails.
* The test runs a bit faster - all the "returns 0" tests in wml_test_schedule are run together, but the other ones need a separate instance of Wesnoth for each one.

data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
```diff
@@ -37,147 +36,161 @@ affect_allies=yes
...
+				{FILTER_ALICE}
+			[/damage]
+			[chance_to_hit]
+				{SPECIAL_ID}
+				value=100
+				{FILTER_ALICE}
+			[/chance_to_hit]
+				[/abilities]
+	[/effect]
+    [filter]
```
kabachuha: The WML tags are misaligned.

data/test/scenarios/swarm_disables_upgrades_with_abilities_adjacent.cfg
newfrenchy83: i had reverted add test fail

kabachuha:
@newfrenchy83 please, don't forget to squash all those failed commits with rebase, so the context is seen more clearly.
image

newfrenchy83:
testing fail finalised but how i will rebasecommit history now?

newfrenchy83:
i close this PR and i pen https://github.com/wesnoth/wesnoth/pull/5541 in replacement

Issue #5041, opened 2020-08-10
Draw text on images in IntroWML, useable for place-name labels on the journey-tracker maps


stevecotton:
> When maps are shown in IntroWML (the `[story]` tag), there's already support for drawing journey-tracking images on top of the map. For the non-creased maps, IntroWML could add support for text boxes.

I plan to implement that, and have added it to the roadmap for 1.15.9. I feel it would be useful for UMC, but wouldn't expect it to meet the standards for use in mainline's campaigns. Maybe we could later add a "skew the map, including labels, to give the nice perspective" later on.

(Originally posted in https://github.com/wesnoth/wesnoth/issues/5011#issuecomment-669941897 , now split to a separate issue corresponding to the roadmap item.)

CelticMinstrel:
Note: If you do include the schema change, you'll need to remove the link to `scenario/story/image` under `[part]`, since there's now an actual tag definition there in its place.

The schema validator should also reject that, mind you… if the same tag is defined multiple times it should be able to detect that and print an error… but that's a separate issue.

stevecotton:
> The schema validator should also reject that, mind you… if the same tag is defined multiple times it should be able to detect that and print an error… but that's a separate issue.

Nice, thanks for adding this in #5227. I've rebased and will push an update here when that PR merges.

stevecotton:
Example image from the current WIP implementation:
![wesnoth_introwml_dw_map](https://user-images.githubusercontent.com/101462/115557230-cef4e880-a2b1-11eb-8de5-7d558ef18b00.png)

AI0867:
Based on this image, I would suggest an (optional) contrasting shadow and/or an opaque block behind the text. You can see real legibility problems for the labels on the mountains. As these labels will be translated, placing them better is no guarantee for legibility in other languages.

CelticMinstrel:
Like Al0867 said, you need some sort of outline to make the text consistently legible. A semi-transparent background (like the removed background on the title) would also work, though I think I prefer a proper outline.

stevecotton:
I might reintroduce this once the 1.18 development cycle starts, but feel it's too late for 1.16.

* Unless each letter can be hand-aligned by an artist then this won't have the quality for mainline campaigns' maps.
* Campaigns can already put overlay images on maps (this is how the journey-tracker's tracks work). The feature would make them easier to translate, but that only matters if the campaign's locations have different names in different languages. Probably not going to be an issue for many UMC campaigns.
[...]

PR #5242, opened 2020-10-26
IntroWML code cleanup and fix for [image]scaled attribute

### LINKED ISSUE #5223: Does anyone want [story][part][image]scaled=yes ?
This is about the `[story][part][image]` tag, not the main `[image]` tag. This tag is documented on the IntroWML page, but is possibly only used by the macros `OLD_JOURNEY`, `NEW_JOURNEY`, `NEW_BATTLE`, etc.

My question is, does anyone want to keep its `scaled` attribute? This undocumented attribute defaults to false, but if you add `scaled=yes` then the image is scaled up to the width and height of the background - not to the same ratio as the background, but to the full size. So if you have a 1000 pixel wide background in an 800 pixel wide window, then `scaled=yes` means each journey-marker is an 800 pixel-wide blob.

The bug is already present in 1.14, haven't checked 1.12 yet. It would be easy to fix, but as no-one's noticed this bug then I'd prefer to just delete it.

Note: the `scaled` attribute doesn't control the positioning of the image. The x and y co-ordinates will still be scaled so that a journey marker that's meant to be on top of a map location is still correctly placed when the background image is scaled.

stevecotton:
Prior to this commit, the scaled= attribute was undocumented and had what I assume was a bug - when enabled the image was scaled up to the width and height of the background - not to the same ratio as the background, but to the full size. So if you had a 1000 pixel wide background in an 800 pixel wide window, then scaled=yes meant each journey-marker was an 800 pixel-wide blob.

However, if anyone was using it with background-size images as overlays then this commit will break that usage - that use case is still supported by having multiple [background_layer]s.

Closes #5223, which was a question about whether to fix or simply remove the scaled attribute. Given that [background_layer]scale= and [image]scaled= have different meanings (the background_layer one should and does cause it to be full-window), I've kept the 'd' on the end of 'scaled' too.

Use a loop instead of recursion. This cleanup is prompted by #5041 (drawing map labels in IntroWML), which will add more items to be drawn in this loop.  The old implementation would not have triggered tail-recursion optimisations, as the recursion site wasn't the final code in the function.

Correct documentation in the .hpp file about when the delay takes effect (this isn't a behavior change, just a documentation correction).

Review the schema, and remove attributes that aren't supported by the code.  Many of these attributes are supported for `background_layer` but not `image`.

### DIFF
--- a/src/gui/dialogs/story_viewer.cpp
+++ b/src/gui/dialogs/story_viewer.cpp
@@ -319,63 +319,58 @@ void story_viewer::display_part(window& window)
 void story_viewer::draw_floating_image(window& window, floating_image_list::const_iterator image_iter, int this_part_index)
 {
 	const auto& images = current_part_->get_floating_images();
-
-	// If the current part has changed or we're out of images to draw, exit the draw loop.
-	if((this_part_index != part_index_) || (image_iter == images.end())) {
-		timer_id_ = 0;
-		return;
-	}
-
-	const auto& floating_image = *image_iter;
-
-	std::ostringstream x_ss;
-	std::ostringstream y_ss;
-
-	// Floating images are scaled by the same factor as the background.
-	x_ss << "(trunc(fi_ref_x * base_scale_x) + base_origin.x";
-	y_ss << "(trunc(fi_ref_y * base_scale_y) + base_origin.y";
-
-	if(floating_image.centered()) {
-		x_ss << " - (image_original_width  / 2)";
-		y_ss << " - (image_original_height / 2)";
-	}
-
-	x_ss << " where fi_ref_x = " << floating_image.ref_x() << ")";
-	y_ss << " where fi_ref_y = " << floating_image.ref_y() << ")";
-
-	config cfg, image;
-
-	image["x"] = x_ss.str();
-	image["y"] = y_ss.str();
-	image["w"] = floating_image.autoscale() ? "(width)"  : "(image_width)";
-	image["h"] = floating_image.autoscale() ? "(height)" : "(image_height)";
-	image["name"] = floating_image.file();
-
-	// TODO: implement handling of the tiling options.
-	//image["resize_mode"] = "tile_centered"
-
-	cfg.add_child("image", std::move(image));
-
 	canvas& window_canvas = window.get_canvas(0);
 
-	// Needed to make the background redraw correctly.
-	window_canvas.append_cfg(cfg);
-	window_canvas.set_is_dirty(true);
+	// If the current part has changed or we're out of images to draw, exit the draw loop.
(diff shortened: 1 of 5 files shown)


src/gui/dialogs/story_viewer.cpp
```diff
@@ -319,63 +319,52 @@ void story_viewer::display_part(window& window)
...
-	image["y"] = y_ss.str();
-	image["w"] = floating_image.autoscale() ? "(width)"  : "(image_width)";
-	image["h"] = floating_image.autoscale() ? "(height)" : "(image_height)";
-	image["name"] = floating_image.file();
+		std::ostringstream x_ss;
+		std::ostringstream y_ss;
 
-	// TODO: implement handling of the tiling options.
-	//image["resize_mode"] = "tile_centered"
+		// Floating images are scaled by the same factor as the background.
```
stevecotton, line 332: Just before creating this PR, I noticed that this should say "Floating image's locations are scaled ...". Hence opening it as a draft for now.

stevecotton:
The two PRs that have autolinked above fix issues found by the schema changes. As part of my local build I'd checked some of the campaigns against the schema, but not all of them.

CelticMinstrel:
Hmm, I'm kinda having second thoughts about this now. Based on #5223 it was clear that scaling an `[image]` wasn't useful with the current implementation, but I think it actually _is_ useful to be able to scale floating images on larger screens. The problem was that it was scaling to the same size as the background, which isn't correct - it should take the same scale factor that was used on the background and apply it to the image as well. If the background is scaled up 1.2x, then the floating image should also be scaled up 1.2x.

An example of a case where you might want to do this is if the background represents a _setting_ and the floating image is a _character_ that needs to move relative to the setting. Keeping the relative scale the same between background and floating seems correct for that case.

stevecotton:
Thanks for the feedback, I've changed it to support the scaled= attribute.

CelticMinstrel:
Seems good. It does seem a little weird for `[background_layer]` to support `scale=` while `[image]` supports `scaled=` though… should we maybe make it consistent?

stevecotton:
[background_layer]scale= and [image]scaled= have different meanings, so I think it's better for them to have different names. The background_layer one should and does cause that layer to be full-window.

CelticMinstrel:
Hmm, maybe, but those are still awfully similar names…

stevecotton:
Maybe `resize_with_background` or just `resize`?

CelticMinstrel:
Yeah, maybe something like that would help avoid confusion…

stevecotton:
Planning to merge this today, if there are no more comments.

stevecotton:
The wiki is currently down for maintenance - added this to the undocumented list as a reminder to myself.

PR #5244, opened 2020-10-27
SotA S04: Remove some [image]scale=no attributes

stevecotton:
While [background_layer] has a scale attribute, [story][part][image] doesn't. A schema update for that will follow in PR #5242.


sigurdfdragon (commented):
Noted.

stevecotton:
@sigurdfdragon does "Noted" mean "merge" or "don't merge"?

stevecotton:
I thought I was going to re-add "scale" to the schema to address https://github.com/wesnoth/wesnoth/pull/5242#issuecomment-716937581 . However, I think now it makes sense to call the supported one "scaled", which means I need these "scale=no" lines to be removed otherwise they'll fail the schema validation in #5242.

sigurdfdragon:
@stevecotton It means neither.

PR #5245, opened 2020-10-27
DM and LoW: Remove '[story][part]delay=' attributes

stevecotton:
None of these had any effect, either on 1.15 or 1.14. I doubt that they ever
had an effect, because if these story screens had 4 second or even 8 second
delays then it would surely have been treated as a bug.

A schema update for this will follow in PR #5242.

### DIFF
--- a/data/campaigns/Delfadors_Memoirs/scenarios/14_Shadows.cfg
+++ b/data/campaigns/Delfadors_Memoirs/scenarios/14_Shadows.cfg
@@ -15,7 +15,6 @@
     [story]
         [part]
             story= _ "Kalenz and his troops fared swiftly south and then westward along the north bank of the Great River. They successfully evaded the orcs, but would soon find there were grimmer foes awaiting them."
-            delay=4000
         [/part]
     [/story]
     {DM_TRACK {KALENZ_STAGE2}}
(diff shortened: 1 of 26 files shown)


CelticMinstrel:
This might be worth testing this on 1.12, as it's probably a regression from then.

I think it's quite likely that the unit on those is supposed to be "ticks" rather than "milliseconds". Assuming 60 ticks per second that would mean they were intended to be 1 or 2 second delays, not 4 or 8 seconds.

stevecotton:
1.12's story screens were in general much slower than 1.14; 1.12 drew the journey markers before showing the story text, and faded the story text in from top to bottom. Personally I find 1.14's responsiveness to be a UI improvement.

However, these delay= attributes don't seem to add any additional delay on top of the 1.12 UI design, and I didn't notice any code to support them (but only took a quick look).

CelticMinstrel:
Okay.

I think the idea of allowing a delay on specific images is nice (for using the story screen as a sort of cutscene builder) but if it didn't function even in 1.12, then I don't have any problem removing the attribute for the time being. It gives us a chance to reconsider the time unit, if nothing else…

Pentarctagon (approved):

PR #5246, opened 2020-10-27
Fix [store_unit_defense] and add a unit test for it

stevecotton:
See https://r.wesnoth.org/t53448

### DIFF
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -953,16 +953,33 @@ function wesnoth.wml_actions.cancel_action(cfg)
 end
 
 function wesnoth.wml_actions.store_unit_defense(cfg)
+	wesnoth.deprecated_message("[store_unit_defense]", 3, "1.17.0", "This function returns the chance to be hit, high values represent bad defenses. Using [store_unit_defense_on] is recommended instead.")
+
 	local unit = wesnoth.units.find_on_map(cfg)[1] or wml.error "[store_unit_defense]'s filter didn't match any unit"
 	local terrain = cfg.terrain
 	local defense
 
 	if terrain then
-		defense = units:defense(terrain)
+		defense = unit:chance_to_be_hit(terrain)
 	elseif cfg.loc_x and cfg.loc_y then
-		defense = units:defense(wesnoth.get_terrain(cfg.loc_x, cfg.loc_y))
+		defense = unit:chance_to_be_hit(wesnoth.get_terrain(cfg.loc_x, cfg.loc_y))
 	else
-		defense = units:defense(wesnoth.get_terrain(unit.x, unit.y))
+		defense = unit:chance_to_be_hit(wesnoth.get_terrain(unit.x, unit.y))
+	end
+	wml.variables[cfg.variable or "terrain_defense"] = defense
+end
+
+function wesnoth.wml_actions.store_unit_defense_on(cfg)
+	local unit = wesnoth.units.find_on_map(cfg)[1] or wml.error "[store_unit_defense_on]'s filter didn't match any unit"
+	local terrain = cfg.terrain
+	local defense
+
+	if terrain then
+		defense = unit:defense_on(terrain)
+	elseif cfg.loc_x and cfg.loc_y then
+		defense = unit:defense_on(wesnoth.get_terrain(cfg.loc_x, cfg.loc_y))
+	else
+		defense = unit:defense_on(wesnoth.get_terrain(unit.x, unit.y))
 	end
 	wml.variables[cfg.variable or "terrain_defense"] = defense
 end
--- a/data/schema/core/actionwml.cfg
+++ b/data/schema/core/actionwml.cfg
@@ -1098,6 +1098,15 @@
 		{SIMPLE_KEY variable string}
 	[/tag]
 	[tag]
+		name="store_unit_defense_on"
+		max=infinite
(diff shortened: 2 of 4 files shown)


stevecotton:
Having opened the PR, I've just noticed a log message that wesnoth.units.defense has been deprecated in favour of wesnoth.units.defense_on. My feeling is that we should similarly deprecate [store_unit_defense] in favour of the (not yet implemented) [store_unit_defense_on] or [store_defense_on].

Pentarctagon (approved):

CelticMinstrel:
Is there a need to deprecate the tag? You could also just implement it in terns of `wesnoth.units.defense_on` and add some option to the tag to check defense on a different terrain type than it's standing on.

stevecotton:
I may have misunderstood the difference between units.defense and units.defense_on. AFAICS from writing this PR and a brief skim of the history, wesnoth.units.defense and [store_unit_defense] return lower-is-better values; for example a footpad on castle gets a value of 30 to represent 70% defense.

AFAICS, wesnoth.units.defense_on  is just an odd name for the function that returns what a coder probably expects, a value of 70 to represent 70% defense. That's the reason why I'd deprecate [store_unit_defense].

stevecotton:
New idea: add a "higher_is_better" attribute.

For compatibility with 1.14, return the chance to be hit rather than the defense, for example 30 would be returned for a unit on 70% terrain.

Setting higher_is_better=yes makes it return 70 for a unit on 70% defense terrain. I suggest that should become the default in 1.17.

Pentarctagon:
While I understand the reasoning, I'm not sure it justifies an eventual silent compatibility break.

CelticMinstrel:
As you can see from https://github.com/wesnoth/wesnoth/blob/master/data/lua/core.lua#L826 and https://github.com/wesnoth/wesnoth/blob/master/data/lua/core.lua#L601 …

So as you noticed, the deprecated function actually returns the chance to be hit on a given terrain type, not the defense value. Thus to avoid the deprecation warning the technically correct solution is to call `wesnoth.units.chance_to_be_hit` instead of `wesnoth.units.defense`… though since you're offering a `higher_is_better` option, there's probably no reason to bother switching on two functions that are explicitly defined as inverses of each other – your latest implementation seems fine.

I think a compatibility break like stevecotton suggests is a good idea, but if we're gonna do that we definitely need a deprecation warning if someone omits the `higher_is_better` key. I think the timeline would look something like this:

* 1.16 - If `higher_is_better` is omitted, give a level 3 deprecation warning saying that the default will change, and default to whatever makes it compatible with 1.14.
* 1.18 - If `higher_is_better` is omitted, probably give a level 3 or maybe even 4 deprecation message, but it defaults to the opposite of 1.16. If it's present, give a level 1 deprecation warning saying it may be removed in the future.
* 1.20 - No message if it's omitted, defaulting to the same as 1.18. The warning if present could maybe be increased to level 2 but doesn't have to be.

In short, for 1.16 and 1.18, we'd encourage people to _always_ use the key; we could even set it to be a mandatory key in the schema. In 1.20, we'd encourage people to omit it once more, and make it optional in the schema.

Side note: I'm not too keen on the name "higher_is_better", but if it's only intended to smooth over an eventual compatibility breakage, I guess the name doesn't matter that much…? I can't think of any _really good_ alternatives, either.

Pentarctagon:
If there's already a function that provides the same functionality, then why would we deprecate anything and force UMC authors to change their code?

stevecotton:
> If there's already a function that provides the same functionality, then why would we deprecate anything and force UMC authors to change their code?

There's already such a function at the Lua API level, but not at the WML API level. However, the reason for deprecating (for both Lua and WML) is that "defense" is the right name for a function that returns the opposite value:

* `defense_on` returns defense on a given terrain, higher values mean less likely to be hit.
* `chance_to_be_hit` returns defense on a given terrain, higher values mean more likely to be hit.
* `defense` returns the chance to be hit.

At the moment `[store_unit_defense]` isn't used in mainline, and it's taken 6 months from 1.15.3's release to a bug report about it being broken, so I think there's few UMC authors dependent on the old behavior.

Pentarctagon:
> defense_on returns defense on a given terrain, higher values mean less likely to be hit.
> chance_to_be_hit returns defense on a given terrain, higher values mean more likely to be hit.

How is this incorrect though?  70% defense means the unit has a 70% chance to dodge the attack, and 70% chance to hit means exactly that.

Pentarctagon:
Also, fewer than half of the add-ons on 1.14 have been ported to 1.15 so far, so I don't think the length of time it took to report the bug can be used as an indicator here since it still works in 1.14.

stevecotton:
> > defense_on returns defense on a given terrain, higher values mean less likely to be hit.
> > chance_to_be_hit returns defense on a given terrain, higher values mean more likely to be hit.
> 
> How is this incorrect though? 70% defense means the unit has a 70% chance to dodge the attack, and 70% chance to hit means exactly that.

Those two are fine, it's the third one that's deprecated.

I need to push a new version to fix the build failure (or maybe I fix it as a bug in the run_wml_tests script). Used a non-ASCII typographical quotation mark in a deprecation message.

Pentarctagon:
I would prefer the [store_unit_defense_on] new tag approach, then.

stevecotton:
Done using the new tag approach.

Pentarctagon (approved):

PR #5247, opened 2020-10-28
Add bounds checks for unit dismiss and recall (fixes #5171)

### LINKED ISSUE #5171: Crash when clicking dismiss when recall list is filtered and no units match the filter.
The description and a simple test case to reproduce the issue is at https://forums.wesnoth.org/viewtopic.php?f=4&t=53284

The backtrace I got when reproducing the crash is at https://pastebin.com/QcUvRm56

seanyeh:
Turns out that https://github.com/wesnoth/wesnoth/issues/5171 is caused when index is -1, when no unit is selected in the list. I don't know if we want some certain error message to show, but otherwise this fix will basically make the buttons do nothing if no unit is selected.

To reproduce error: open recall window in any game and type in search filter so no units match. Clicking "dismiss" or "rename" will segfault

### DIFF
--- a/src/gui/dialogs/unit_recall.cpp
+++ b/src/gui/dialogs/unit_recall.cpp
@@ -313,6 +313,10 @@ void unit_recall::rename_unit(window& window)
 	listbox& list = find_widget<listbox>(&window, "recall_list", false);
 
 	const int index = list.get_selected_row();
+	if (index == -1) {
+		return;
+	}
+
 	unit& selected_unit = const_cast<unit&>(*recall_list_[index].get());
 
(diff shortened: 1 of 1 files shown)


shikadiqueen:
As gfgtdf said in the original issue, we'd also want the buttons to remain greyed out while the selection is invalid.

seanyeh:
@shikadiqueen Good point, I've added that. I guess now technically we don't need the index checks...should we still keep them to be safe?

shikadiqueen:
I generally do that in my own code as a safeguard for completely unexpected situations, so I'd say yes.

Issue #1082, opened 2017-05-08
Can't disable chat / set chat lines to 0 for replay mode (GNA #7202)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?7202)** by anonymous on 2006-09-27

Hi.

Now that bug #6672 is fixed and you can access preferences while watching replay, it would be nice if you would have the option to  set the "Chat Lines" slider to 0 on the Multiplayer Preferences tab, to prevent chat from obstructing the game view.

(If you could access the command line in replay mode, you could :clear, but I don't think this is really all that necessary, and it would be an altogether different feature request. It wouldn't really be quite as useful for just this purpose, although I'm sure many would use that option too.)

Thanks.

(Reproduced on all)
Release: 1.3-svn
Priority: 5 - Normal
Severity: 3 - Normal

wesnoth-bugs:
**Modified on 2006-10-29**

**anonymous wrote:**

> No real need to limit this to &gt;=1 in gameplay either. I can imagine someone wanting to only observe a game &quot;live&quot; without having the chat obstruct the view.
> 
> 
> 
> See also f.r. / bug #7254 for a related issue.

wesnoth-bugs:
**Modified on 2016-02-27**

**ancestral wrote:**

> This 9 1/2-old feature request is still absolutely valid, whether it�s implemented as a check box in the replay UI, or in preferences, or just being able to set chat lines to 0.
> 
> 
> 
> I am working on the new Wesnoth trailer, and going through replays is an excellent way to recording video captures of in-game action. In-game chat makes this less desirable. I think there's also value to squelching the chat if someone wishes to share their replay with others, but is sensitive to the chat that occurred.

PR #5248, opened 2020-10-28
Allow users to set chat lines to 0.

Hejnewar:
Allow users to set chat lines to 0. #1082 fix.

This is really simple fix making me wonder if its correct, I tested it and it seems to work as intended without any bugs. If this wasn't done because of some internal reason or good of players this can be closed.


ElderoZ:
That would be a good change

Issue #2817, opened 2018-04-02
chat focus loss when room owner re-assignes sides


vgaming:
Steps:
* create a game on MP server 
* join the game with another client (second client)
* start typing anything in game chat with the second client (not room owner)
* let first client (room owner) re-assign a side. Like choosing between "Network Player" and "Empty" and player names

What you'll observe at this point: chat focus on the second client gets lost. Focus goes elsewhere right while you typed something. (Thus breaking your typing.)

Vultraz:
Honestly, this is expected. Most apps would do that.

CelticMinstrel:
Stealing focus is never expected and is always bad. The only reason most apps do it is because it's not actually an easy problem to solve. For this particular case it's about a text field focus, so... how do you determine whether the user is currently typing into the text field? How do you ensure that the text field remains focused? It may be inevitable that events temporarily remove the focus, so what you'd have to do is ensure it gets restored before keyboard event handling continues.

I think this is a good thing to fix, eventually.

CelticMinstrel:
Yeah, it shouldn't have to, so if it can be fixed in a way that doesn't remove the focus in the first place, that's even better.

ProditorMagnus:
Also happens when someone joins or leaves game.

jostephd:
This fixes it, but makes it so it's not possible to use up arrow / down arrow to scroll the sides list. It also makes it so the chat box is focused when that dialog is entered.

```
diff --git a/src/gui/dialogs/multiplayer/mp_join_game.cpp b/src/gui/dialogs/multiplayer/mp_join_game.cpp
index 760f9b9eb23..e36c296ff97 100644
--- a/src/gui/dialogs/multiplayer/mp_join_game.cpp
+++ b/src/gui/dialogs/multiplayer/mp_join_game.cpp
@@ -339,7 +339,7 @@ void mp_join_game::generate_side_list(window& window)
 
 	tree_view& tree = find_widget<tree_view>(&window, "side_list", false);
 
-	window.keyboard_capture(&tree);
+	window.keyboard_capture(&find_widget<chatbox>(&window, "chat", false));
 
 	tree.clear();
 	team_tree_map_.clear();
```

vgaming:
I think that's pretty good. I am curious of course if we can insert some additional easy (yet maintainable) hack to detect up/down keystrokes and do the appropriate action, but even in the proposed form, I think it's worth it.

I would merge.

soliton-:
[...]

PR #4737, opened 2020-02-10
fix chat focus loss when re-assigning sides

vgaming:
Fixes GH-2817

Patch originally provided by @jostephd at github: https://github.com/wesnoth/wesnoth/issues/2817

PR #5252, opened 2020-10-31
Don't lose text focus on mp player join (fixes #3715)

### LINKED ISSUE #3715: 1.14.5 MP Lobby text input loses focus when new message arrives
Wesnoth v1.14.5 (MP at Official server)
Windows 7 x64

Join to any hosted game. When you write something and server message comes (for example - "Someone joined game") command line loses focus. So you need to click on it to continue writing.  
Very annoying.

seanyeh:
https://github.com/wesnoth/wesnoth/issues/3715

To reproduce:
1. Player 1 hosts a multiplayer game
2. Player 2 joins the game. Player 2 focuses the chat input
3. Player 3 joins the game. Player 2 is no longer focused on the chat input

Does anyone know why we set keyboard focus here?
https://github.com/wesnoth/wesnoth/blob/6aa6c010e77eeded27dccbb200af06214363937f/src/gui/dialogs/multiplayer/mp_join_game.cpp#L348
I've looked through the commit history and it was there since the original version of the file. This takes away focus from the chat input when a player joins, and doesn't look like it does anything useful. But I could be wrong - I'm pretty new to the codebase.

### DIFF
--- a/src/gui/dialogs/multiplayer/mp_join_game.cpp
+++ b/src/gui/dialogs/multiplayer/mp_join_game.cpp
@@ -345,8 +345,6 @@ void mp_join_game::generate_side_list(window& window)
 
 	tree_view& tree = find_widget<tree_view>(&window, "side_list", false);
 
-	window.keyboard_capture(&tree);
-
 	tree.clear();
 	team_tree_map_.clear();
 	const std::map<std::string, string_map> empty_map;


Pentarctagon:
Thanks for taking a look at this bug as well as the others :)

seanyeh:
@Pentarctagon no problem, had some free time these few days so this has been something fun to do. i've been a huge fan of wesnoth for a long time

Vultraz:
HMMMM. I can't remember why I added that line. Probably to allow for keyboard nav of the tree but... that doesn't seem that useful. 🤔

soliton-:
So I assume this is a superior solution to #4737? Presumably it also fixes #2817 then.

seanyeh:
@soliton- Just went through the reproduction steps in #2817 - looks like this fixed that as well

vgaming:
@seanyeh yaaay, thanks for fixing! This is really nicer this way

PR #5253, opened 2020-11-01
AToTB: Changing a difficulty name

Konrad22:
'Knight' to 'Grand Knight' so it matches the sprite used.


CelticMinstrel:
I don't really have a problem with this change, but it does seem kinda pointless…

PR #5254, opened 2020-11-01
Fix dragging labels in editor (fixes #3302)

### LINKED ISSUE #3302: Moving labels in editor causes back to titlescreen
To move a label, use drag and drop:
move_a_label

seanyeh:
https://github.com/wesnoth/wesnoth/issues/3302

Currently in the scenario editor, dragging a label causes a crash. I rewrote the dragging functionality to be similar to how we do it with units, hope that's ok

### DIFF
--- a/src/editor/action/mouse/mouse_action_map_label.cpp
+++ b/src/editor/action/mouse/mouse_action_map_label.cpp
@@ -32,33 +32,15 @@ editor_action* mouse_action_map_label::click_left(editor_display& disp, int x, i
 	click_ = true;
 	map_location hex = disp.hex_clicked_on(x, y);
 	clicked_on_ = hex;
-	last_draged_ = hex;
 	return nullptr;
 }
 
 editor_action* mouse_action_map_label::drag_left(editor_display& disp, int x, int y
-		, bool& partial, editor_action* /*last_undo*/)
+		, bool& /*partial*/, editor_action* /*last_undo*/)
 {
 	map_location hex = disp.hex_clicked_on(x, y);
-
-	/* Cursor is still on old hex field */
-	if (hex == last_draged_)
-		return nullptr;
-	click_ = false;
-
-	editor_action_chain* chain = nullptr;
-	const terrain_label* label = disp.get_controller().get_current_map_context().get_labels().get_label(last_draged_);
-
-
-	if (label) {
-		partial = true;
-		chain = new editor_action_chain(new editor_action_label_delete(last_draged_));
-		chain->append_action(new editor_action_label(hex, label->text(), label->team_name(), label->color(),
-			label->visible_in_fog(), label->visible_in_shroud(), label->immutable(), label->category()));
-	}
-
-	last_draged_ = hex;
-	return chain;
+	click_ = (hex == clicked_on_);
+	return nullptr;
 }
 
 editor_action* mouse_action_map_label::up_left(editor_display& disp, int x, int y)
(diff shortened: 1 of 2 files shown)


src/editor/action/action_label.cpp
```diff
@@ -88,4 +88,42 @@ void editor_action_label_delete::perform_without_undo(map_context& mc) const
...
 }
 
+IMPLEMENT_ACTION(label_replace)
+
+editor_action* editor_action_label_replace::perform(map_context& mc) const
+{
+	editor_action_ptr undo(new editor_action_label_replace(new_loc_, loc_));
+
+	perform_without_undo(mc);
+	return undo.release();
```
stevecotton, line 98: A task for a separate PR, but this method should be returning the std::unique_ptr instead of returning an owning raw pointer.

src/editor/action/action_label.hpp
```diff
@@ -97,4 +97,27 @@ class editor_action_label_delete : public editor_action_location
...
+
+	/** Inherited from editor_action, implemented by IMPLEMENT_ACTION. */
+	editor_action_label_replace* clone() const;
+
+	editor_action* perform(map_context& mc) const;
+
+	void perform_without_undo(map_context& mc) const;
+
+	/** Inherited from editor_action, implemented by IMPLEMENT_ACTION. */
+	const std::string& get_name() const;
```
stevecotton, line 117: The four const methods can and should all have the `override` keyword too.

src/editor/action/action_label.cpp
```diff
@@ -88,4 +88,42 @@ void editor_action_label_delete::perform_without_undo(map_context& mc) const
...
+{
+	const terrain_label* old_label = mc.get_labels().get_label(loc_);
+
+	// Do nothing if label does not exist in original location
+	if (!old_label) {
+		return;
+	}
+
+	// Move label to new location by creating one in the new location
+	// with the same properties and clearing the old location
```
stevecotton, line 111: This doesn't undo correctly.

* Create labels 'a' and 'b'
* Drag 'a' on top of 'b'
* Try to undo

The result is that 'a' moves back to where it originally was, but 'b' doesn't reappear.

stevecotton (commented):
The build failure was just an unrelated build breakage from changes in campaignd that were merged into master.

src/editor/action/action_label.cpp
seanyeh: Thanks, you're right, this is something I completely overlooked. I was still having some trouble understanding how these editor actions worked, but I think I have a better understanding now. I've changed my approach to be simpler and similar to the original approach, but moving the logic to `drag_end_left` (when the drag is released).

src/editor/action/mouse/mouse_action_map_label.cpp
```diff
@@ -108,6 +90,47 @@ editor_action* mouse_action_map_label::up_right(editor_display& disp, int x, int
...
+		.get_label(clicked_on_);
+
+	// Return nullptr if the dragged label doesn't exist
+	if (!dragged_label) {
+		return nullptr;
+	}
+
+	// Create action chain to:
+	// 1. Delete label in initial hex
+	// 2. Delete label in target hex
```
stevecotton, line 114: Please say something on this line like "... if it exists, will be a no-op if it doesn't"

stevecotton:
+1, Looks good to me.

I think the editor_action framework should be updated to use std::unique_ptr in place of raw pointers, but that's a separate task.

Vultraz:
@stevecotton if you've confirmed this works can be merged

PR #5255, opened 2020-11-02
AToTB S02: Rewrite the script with the elves thinking the baddies are grey mages

stevecotton:
@Earth-Cake and @Konrad22 
This is about 10 lines of translatable text, not sure if it's okay to backport to 1.14.
The storyline itself is going to have large changes by 1.18 (maybe 1.16), so I'm not sure if
this will ever be in a stable release. But having had the idea, I thought I'd make a PR for it.

If a player loads a start-of-S02 save without the `who_slew_mordak variable` then it will still
work, showing the "I saw the killing blow with my own eyes" version.

Explanation for what's going on (put in the commit message because this bit is non-canon):

There are some "grey mages" in the Grey Woods, who appear in the Liberty
campaign. Most of the details of these were removed from 1.15, but in 1.14's
Liberty S06 they don't tolerate necromancers. SotA mentions an ancient elvish
civil war in these woods, with the ghosts still haunting the battlefield. I'm
assuming a general truce between elves and grey mages, on the understanding
that the grey mages are doing exorcisms of ghosts from the ancient war.

### DIFF
--- a/data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
+++ b/data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
@@ -34,7 +34,7 @@
             story=_ "16 V, 363 YW
 Excerpt from the journal of Arvith of Maghre
 
-We’ve been searching three days for Baran, and turned up nothing. My best hunch was to head north into the borderlands, where the necromancer’s minions could safely hide; everywhere else is more farmland. At first I thought the search might be useless, but late in the first day we found a set of tracks. Some of them had been made by skeletal feet.
+We’ve been searching three days for Baran, and turned up nothing. My best hunch was to head north into the borderlands, where the necromancer’s minions could safely hide; everywhere else is more farmland. At first I thought the search might be useless, but late in the first day we found a set of tracks. Some of them had been made by skeletal feet, although after their first campsite the tracks were merely those of men carrying heavy loads.
 
 We’re close enough to be certain now: those tracks are heading into the Grey Woods. No one from Maghre or any of the other villages has gone into that forest in living memory. Stories have been passed down for generations warning against it. Supposedly the place is haunted by lost souls who hunger for the living, and anyone who dies there is doomed to join them."
         [/part]
@@ -173,7 +173,7 @@ Besides... I want my brother back."
         id=Muff Toras
         name= _ "Muff Toras"
         unrenamable=yes
-        x,y=9,4
+        x,y=9,5
 
         [ai]
             passive_leader=yes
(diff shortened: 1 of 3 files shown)


Konrad22:
Personally I don't see any conflicts with the 1.14 lore. But I do think that you should pay more attention to the case of Elves or even the elvish leader staying alive until you unmask the Necromancers. Otherwise it will feel left incomplete whenever someone actually manages to do so.
TL;DR Keeping Elves alive can make you miss out on story without any replacement.

stevecotton:
I've added an elvish scout riding in (regardless of whether any other elves are alive), saying "I heard sounds of battle, but did not expect to see this. I will inform the Council, they can work out who the necromancers are.", and riding off quickly.

Konrad - thanks, I meant to ping you as you'd recently commented about this in the forum threads. I should have pinged LordLewis too, but Github's auto-complete isn't picking up based on that name; if you'd ping him it would be appreciated by me.

data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
```diff
@@ -445,11 +450,87 @@ Besides... I want my brother back."
...
 
+        [move_unit]
+            id=Muff Toras
+            to_x,to_y=9,4
+            force_scroll=yes
+        [/move_unit]
+
         [message]
             speaker=Muff Toras
             message= _ "Curses! If they had been an hour slower, our master’s reinforcements would already be here to meet us."
```
Earth-Cake, line 461: Although this is not a part of the change, this line does seem a bit illogical. If the reinforcements were travelling towards there, it's unlikely that Arvith wouldn't get to fight them, as they'd be on his way, as far as I know. And IIRC there is nothing that even mentions these reinforcements later on.

Earth-Cake:
@Lordlewis6

Earth-Cake (approved):

data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
stevecotton: Good catch. I think this wording would be better, but it needs corresponding changes to the objectives and `time over` events (which both reference these reinforcements too) :

```
# po: There are no undead on the map yet, but Muff Toras summons some immediately after saying this
message= _ "Curses! If they had been a few hours slower, I might have been able to cover my tracks without arousing the elves’ suspicions. Still, no need for the disguise any more."
```

stevecotton:
Going to force-push to add a changelog message, without changing the other files.

data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
Earth-Cake: I agree, it looks good. As for the turn limit, I guess unlimited could be fine. Or the time limit could remain the same, as they need to find Baran as soon as possible.

Konrad22 (approved):

nemaara (approved):

PR #5258, opened 2020-11-03
Remove a few bits of never implemented editor features.

Pentarctagon:
The sound source tool in particular, since showing a button to players whose only function is to tell them it's not implemented makes no sense.


stevecotton:
This could remove every mention of "sound" from data/core/editor/help.cfg, but if it's not done in this PR then it can be done in #4999.

CelticMinstrel:
It's good to remove them from the theme when they're not implemented, but I think we should definitely implement these things too. Can we maybe open an issue for each of the things that's not implemented and removed in this PR, so we don't forget about them?

I'd probably also leave in the hotkey definitions (the details can be commented out so they don't appear in the hotkey list)… they could be added if the things are implemented later, but that's more work, and as long as they're not shown to the player, I don't think they'll get in the way?

Both the sound source tool and map rotation would definitely be useful, I think. I'm not sure what "editor selection generate" is, maybe that one's not useful.

stevecotton:
We currently have 977 open issues, I feel it would be better to put ideas on a Wiki page unless someone steps forward committing to work on them in time for 1.18.

CelticMinstrel:
I think issues are a better place for suggestions than the wiki, and keeping the total count of issues low doesn't seem like a reasonable argument to me.

Pentarctagon:
@stevecotton I'll update the help here then.

@CelticMinstrel The UI button for the soundsource tool was last touched in 2013, and even back then its tooltip said that it wasn't implemented, so I very much doubt that anyone is going to step forward suddenly now to implement any of these.  The other two were literally never more than an entry in an enum and a case statement that always returned false.  Opening a single issue to list out all the things we'd like to have implemented in the scenario editor would be fine, but opening issues to implement things that have already been left unimplemented for the better part of a decade doesn't seem useful.

Pentarctagon:
Also, putting things in the wiki to keep the issue count low here on github isn't really accomplishing anything other than pushing those things under the rug where they're more likely to be forgotten.

CelticMinstrel:
I don't see the difference between one issue and two issues in this case. The sound source tool would be useful, and rotating a map would be useful. What's wrong with opening a different issue for each?

Pentarctagon:
This is by far not the only thing I'm going to be doing with the editor, so I'm expecting more things to come up during that and it would be convenient to have all the suggestions in one place.

CelticMinstrel:
Well, I guess either way works; I'd prefer separate issues but if you want to make a combined one I guess it'll do…

PR #5260, opened 2020-11-05
Convert the editor_action framework to use std::unique_ptr instead of raw pointers

stevecotton:
When returning std::unique_ptr instead of raw pointers, the return type can't be a co-variant, which is why the clone() functions are all being changed to return the base class.   

Most of the dozen or so uses of editor_action_ptr where replaced with the `auto` keyword, but the reason that I didn't use in it the new code was that I find `std::unique_ptr<editor_action>` quicker to read than remembering if editor_action_ptr is a unique_ptr, a shared_ptr or a plain raw pointer.

### DIFF
--- a/src/editor/action/action.cpp
+++ b/src/editor/action/action.cpp
@@ -60,11 +60,11 @@ std::string editor_action::get_description() const
 	return "Unknown action";
 }
 
-editor_action* editor_action::perform(map_context& mc) const
+std::unique_ptr<editor_action> editor_action::perform(map_context& mc) const
 {
-	editor_action_ptr undo(new editor_action_whole_map(mc.map()));
+	auto undo = std::make_unique<editor_action_whole_map>(mc.map());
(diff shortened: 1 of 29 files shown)


Vultraz:
Might be better to use the `editor_action_ptr` alias consistently. Intellisence in vscode can reveal the prt type.

stevecotton:
The build failure is a "not enough time to run tests" one.

stevecotton:
I plan to merge this tomorrow morning, unless it gets a negative review.

src/editor/action/action.cpp
```diff
@@ -260,24 +250,24 @@ editor_action* editor_action_starting_position::perform(map_context& mc) const
 		// action_chain.
 		editor_action_chain* undo_chain = new editor_action_chain();
```
stevecotton, line 251: This line should also use a unique_ptr

PR #5261, opened 2020-11-05
Only show notifications for visible mp games (fixes #4723)

### LINKED ISSUE #4723: Desktop notifications ignore current game filter
 - windows steam 1.14.9

### Describe the bug
After adding game filter by text, I still get notification about game I will not see after switching to wesnoth - example https://i.vgy.me/lbp2Hb.png

### To Reproduce
- Add game filter which would not match newly created game
- Switch to different program
- Create new game from different computer

### Expected behavior
No desktop notification since created game does not match filter

### Screenshots
https://i.vgy.me/lbp2Hb.png

seanyeh:
https://github.com/wesnoth/wesnoth/issues/4723

Don't know if this approach is best, but I broke out a `is_game_visible` method from `apply_game_filter` to test if an individual game is visible.
I had originally considered adding some sort of check in
https://github.com/wesnoth/wesnoth/blob/1cdc4ef530835876621f74b681646271fc4646b0/src/gui/dialogs/multiplayer/lobby.cpp#L310-L313
 but it proved tricky because the game filter isn't actually applied until https://github.com/wesnoth/wesnoth/blob/1cdc4ef530835876621f74b681646271fc4646b0/src/gui/dialogs/multiplayer/lobby.cpp#L397 Making an additional call to `apply_game_filter` beforehand worked but didn't seem like a good idea to call it so many times. (the logic in `lobby.cpp` is messy enough already)

### DIFF
--- a/src/game_initialization/lobby_info.cpp
+++ b/src/game_initialization/lobby_info.cpp
@@ -343,6 +343,24 @@ void lobby_info::make_games_vector()
 	games_visibility_.flip();
 }
 
+bool lobby_info::is_game_visible(const game_info& game) {
+	bool show = true;
+
+	for(const auto& filter_func : game_filters_) {
+		show = filter_func(game);
+
+		if(!show) {
+			break;
+		}
+	}
+
+	if(game_filter_invert_) {
+		show = !show;
+	}
+
+	return show;
+}
+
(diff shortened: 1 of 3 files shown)


stevecotton (approved):
Looks good to me, and a quick test worked.

PR #5264, opened 2020-11-07
Colored attack type icons, resistances and movement costs. Reordered resistances.

walodar:
Using colors to make relevant information stand out at a quick glance.
The six attack type icons are recolored using the six equidistant hues cyan, green, yellow, red, magenta and blue, with the same saturation and lightness as the old icons. The images are optimized with woptipng.
Resistances and movement costs are colored using a gradient from red to green, like the defense values.
The resistance table in the help browser now also shows the attack type icons.
Previously, the resistance table was always ordered alphabetically by their original English names (not their translation, unlike the terrain modifiers). Now, the order is set to blade - pierce - impact - fire - arcane - cold, this corresponds with their frequency of occurrence and makes the attack type icons form an unrolled color wheel.
These changes are made in all relevant areas, including the help browser, tooltips, the sidebar and dialogs.

Example:
comparison


CelticMinstrel:
I like the idea, but I'm not sure I like the choice of colour assignments; the physical ones seem weird, at least. Also, I feel like fire and cold belong together, rather than with arcane in between. There's no _reason_ for putting the icons in colour wheel order other than that there was no other reason to put them in a different order, right? I doubt many people would even notice that the order is the same as a colour wheel.

ElderoZ:
"Now, the order is set to blade - pierce - impact - fire - arcane - cold, this corresponds with their frequency of occurrence and makes the attack type icons form an unrolled color wheel."

This is very arbitrary, way more arbitrary than the current setup. And im pretty sure its not even true. Another rather obvious point is that it will change depending on the era you are playing.

About colored resistances, now, thats a lot of colors, colored terrains defense is good as the colors are more easily distingushable from each other and you basically are working with just a few set colors, in practice you will see only 6 colors (maybe 7, there are a few units with 10% defense), for defense from 20% to 70%. Here you would need colors for resistances from like -50% to 70% which is 12 colors, double the amount. And it shows, the color differences between resistances changing by 10% aren't very easy to see at first glance.

walodar:
Thanks for the interest. I have no strong arguments for the color choice of the physical attack types. One could maybe say that cyan for blade fits cold steel and green for pierce fits wooden arrows or spears, but mostly, it's because I see no good reason against this choice and it makes for a neat arrangement.

Red for fire and blue for cold is in my opinion more necessary, perhaps somewhat unfortunately. I can understand wanting to have fire and cold next to each other, but personally, I would prefer not to break the order for it.

More importantly, and also regarding the second comment, the order reflects the number of attacks of that type over all units, so resistances listed earlier are in some sense more crucial.

The point that this order can change for custom eras is a valid one, but I do not understand the remark claiming that it is not correct for mainline content. Using the unit list from https://units.wesnoth.org/trunk/mainline/en_US/mainline.html and considering that the Bowman line is listed twice, searching for the different attack types yields:
blade: 164
pierce: 123
impact: 108
fire: 33
arcane: 20
cold: 13

2 of the 13 cold attacks belong to the brand-new Icemonax and Greater Icemonax units that do not yet appear in any mainline campaign or multiplayer era, so I think cold should be last.
This order of attack types is also valid for the mainline multiplayer eras.

Finally, I personally do not think it's bad that the color differences are small. Looking at the current situation for the Heavy Infantryman, there is no color difference from 0% to 20%, but a radical shift from 0% to -10%. Also for the Lich, I think it's worth emphasizing that -10% is not so bad, but -40% is quite a bit worse.

Hejnewar:
I also think that fire and cold belong toghether. 

"More importantly, and also regarding the second comment, the order reflects the number of attacks of that type over all units, so resistances listed earlier are in some sense more crucial."

Not really, you can have 20 weak attacks and 13 very strong attacks for example. In case like that resistance versus 13 very strong attacks will be more crucial. That's just example tho. Another thing would be default mp games, there arcane type is pretty much non existent for example.

CelticMinstrel:
I don't think it makes sense to order the damage types based on how common they are on mainline units. It's fine for the three physical damage types, because there's no other argument I can think of on how to order them, but for the energy damage types, I really think fire and cold belong together.

I definitely agree that red for fire and either blue or cyan for cold is necessary. Purple for arcane makes sense too, because purple is a "fancy" colour, so it's suitable for magical damage. Your arguments for blade and pierce seem as good as anything unless there are other concerns, so sure, I guess that's fine. That leaves yellow for impact… I can't really think of anything better, so whatever? (And the only other option for an impact colour is probably orange anyway, which isn't any better or worse than yellow.)

walodar:
Hi, okay, I have now modified the order to blade - pierce - impact - fire - cold - arcane, as requested. Thanks all!

comparisonColdBeforeArcane

ElderoZ:
"The point that this order can change for custom eras is a valid one, but I do not understand the remark claiming that it is not correct for mainline content. Using the unit list from https://units.wesnoth.org/trunk/mainline/en_US/mainline.html and considering that the Bowman line is listed twice, searching for the different attack types yields:
blade: 164
pierce: 123
impact: 108
fire: 33
arcane: 20
cold: 13

2 of the 13 cold attacks belong to the brand-new Icemonax and Great Icemonax units that do not yet appear in any mainline campaign or multiplayer era, so I think cold should be last.
This order of attack types is also valid for the mainline multiplayer eras."

So you just counted the number of attacks of a specific type in all units, for levels? Thats a very basic approach and doesn't at all reflect how often you will see certain attacks in the game, its basically not a representation of anything meaningful, for an average player that representation would make absolutely 0 sense, I bet nobody has done what you did here, ie just counted the number of all attacks. Alphabetical sorting at least represents something meaningful, it makes sense.
You can't even recruit most of these units, level 3s and 2s are usually not recruitable. Also some attacks like say bowman's melee are way weaker. Moreover this in no way reflects how often some units are recruited and how often you would see them on the battlefield

"Finally, I personally do not think it's bad that the color differences are small. Looking at the current situation for the Heavy Infantryman, there is no color difference from 0% to 20%, but a radical shift from 0% to -10%. Also for the Lich, I think it's worth emphasizing that -10% is not so bad, but -40% is quite a bit worse."

Well, if they are so small, isnt it a similiar situation to the one you talked about where 0 and 20% have the same color? It could be argued that the radical shift for negative resistances is helpful, because it helps warn people about certain types of attacks that are effective against a unit.

CelticMinstrel:
Just for the record, I personally have no issue using the gradient for the resistance colours.

I was just wondering… what about custom damage types? Will those be shown with this change? (For that matter, were they shown _before_ this change?)

walodar:
Hi, okay, I agree that counting the number of attacks is a basic approach, but I do think that there exists a correlation. I also find the notion that some types belong together mentioned by the other commentators appealing. In my opinion the benefits of an alphabetical order increase greatly with the length of a list, and since we have only six entries here, I find the suggested order with its associated "meanings" more appropriate in this case.

Sorry for not being clear concerning the resistance colors. What I meant was that I like that with a color gradient, there are no sudden large jumps for small differences, like from 0 to -10%, but large differences become more and more apparent.

I don't know if the game supports custom damage types in general, but the changes should not affect this. My comparison operator works for all strings, those other than the default six are sorted behind them (and then with the < operator).

ElderoZ:
"Hi, okay, I agree that counting the number of attacks is a basic approach, but I do think that there exists a correlation. I also find the notion that some types belong together mentioned by the other commentators appealing. In my opinion the benefits of an alphabetical order increase greatly with the length of a list, and since we have only six entries here, I find the suggested order with its associated "meanings" more appropriate in this case."

It would make sense to literally noone, thats my biggest problem with it. The criterion used is so obscure and so meaningless, I can't think of a single thing that would correlate with the number of attacks of a certain type in an era, units used, damage dealt, balance, number of units with a certain attack, nothing.

"Sorry for not being clear concerning the resistance colors. What I meant was that I like that with a color gradient, there are no sudden large jumps for small differences, like from 0 to -10%, but large differences become more and more apparent."

Fair, its a convention to separate resistances into groups, and I think it makes more sense than adding so many very similiar unnecessary colors so there are no large jumps. But at the same time the differences in resistances on units in general as a whole are rather small, large differences in resistances are more of an exception than a rule.


"I don't know if the game supports custom damage types in general, but the changes should not affect this. My comparison operator works for all strings, those other than the default six are sorted behind them (and then with the < operator)."

"I was just wondering… what about custom damage types? Will those be shown with this change? (For that matter, were they shown before this change?)"

It does, and the colors work for them like for all other resistances,

CelticMinstrel:
> I don't know if the game supports custom damage types in general

It does support custom damage types, but I wonder if it supports setting an icon for them…? Are the icons for the default damage types recoloured using magenta, or are they actually those colours in the file?

ElderoZ:
"It does support custom damage types, but I wonder if it supports setting an icon for them…? Are the icons for the default damage types recoloured using magenta, or are they actually those colours in the file?"

I tested it using Guru of Magic from eoma, damagetype's icon in help is image not found and it doesn't show in GUI. So it wouldn't work it seems.

Well, im not really a fan of representing damage types as icons and not words anyway.

walodar:
Hi, thanks for the continued interest. I found one more argument for the suggested damage type order: The online unit list has been using this order since Wesnoth 1.4, so for more than 12 years. Example: https://units.wesnoth.org/1.4/C/Merman%20Fighter.html

For the help browser (help_topic_generators.cpp), the icons in the resistance table work the same way as the existing ones from the list of attacks some lines of code above. The icons are searched for in the directory images/icons/profiles/ .
With today's additional commit, the handling is now a bit more graceful, missing images are replaced by a blank image scaled to the same size, the formatting of the tables is preserved, also in the sidebar and in dialogs.

The image files for the icons are in the actual final colors, there is no intermediate step with recoloring from magenta.

walodar:
Hi, one other thing about missing damage type icons: For damage type "x", the accessed image is "icons/profiles/x.png", but like for attack icons, this directory structure does not necessarily have to be in the main game directory.

For your concrete example with Era of Magic and its icon-less new damage type "secret", creating a file named "secret.png" in the directory "_userdata_/data/add-ons/Era_of_Magic_Resources/images/icons/profiles/" is all that's required.
(All the folders except for the last subfolder "profiles" already exist after the installation of the add-on "Era of Magic - Resources", and Era of Magic's binary_path is already set to include "data/add-ons/Era_of_Magic_Resources/images".)

Pentarctagon:
Was there ever a final decision about whether this is something that could be merged or not?

@Vultraz in particular.

CelticMinstrel:
I think the idea is good but I'm a bit unsure about the implementation…

Wedge009:
This seems like a nice addition and complements the icons (fairly) recently added for terrains. I also agree with what's settled on the ordering of the attack types, particularly the grouping of the physical vs energy attack types. I have no strong opinion on the colouring.

Unless there's specific concerns about the implementation I think it's worth merging.

Pentarctagon:
@CelticMinstrel @stevecotton any thoughts on getting this merged?

CelticMinstrel:
My comment from last year still stands.

Pentarctagon:
Can you be more specific about what should be improved in the implementation?

CelticMinstrel:
Um… I can't think of anything specific? I'm just not entirely sure what it's doing, nor am I familiar with this part of the code, and it looks a little messy or something.

stevecotton:
Feedback from someone who's colorblind (or who knows how to assess usability for colorblindness) would help. I'm worries that the icons for blade and pierce are already similar (bottom-left to top-right horizontal slash), and that putting them on blue and green backgrounds makes them more similar.

For the movement points, I think there should be a single palette for all units, instead of having `4` be redder for Heavy Infantry compared to more mobile units, otherwise it's harder to compare possible recruits. The palette for the HI seems better to me than the reduced variance in the Lich's palette. Removing white from the color palettes for numbers seems to be reducing the ease of distinguishing numbers by color, personally I find the old palette as good as the new HI palette.

For the resistance colors, I think the colors should be spread more in the range that units are likely to have. Instead of spreading the colors across the full -100 to +100 range, maybe just -60 to +60 for the full spectrum.

I think there's utility in having a big difference between 0 and -10, as a resistance of zero or more allows thinking "the two biggest hitters on the enemy side can do a maximum of 40hp, so any of my units with more than that and only two exposed sides will survive" - having a negative resistance  breaks that train of thought.

CelticMinstrel:
> Feedback from someone who's colorblind (or who knows how to assess usability for colorblindness) would help.

I think this could be dodged somewhat by using the recolouring system to apply the colour to the icons, while the icons themselves on disk are all magenta or some other fixed colour palette. That would open the door to having this colour scheme for non-colour-blind people and a different one for the colour-blind.

> For the resistance colors, I think the colors should be spread more in the range that units are likely to have. Instead of spreading the colors across the full -100 to +100 range, maybe just -60 to +60 for the full spectrum.

So basically, if a unit _does_ have ±100, it would use the exact same colour as ±60?

walodar:
Hi, thank you. Of course I agree that things should be as nice as possible for colorblind people. Currently, all attack type icons use the same shade of red, so at the very least, the situation wouldn't be worse with the new icons.

Changing the code to use the recoloring system for icons in magenta would be possible, it would be necessary to do this everywhere where icons are used (like the sidebar or the recruitment dialog), and the target colors would also have to be specified somewhere, perhaps in team-colors.cfg.

Custom damage type icons (see my comment from 11 Nov 2020) would then also require more than an image file.

By the way, here is a screenshot of the recruitment dialog:

recruitment

For the color of the movement costs on different terrains, your suggestion of using a fixed color for every number would illustrate the direct information that is also given by the numerical value of that number.

In my proposal, the color is relative to the movement points of that particular unit, where pure yellow means moving in that terrain costs half of the unit's movement points, and even more reddish colors mean the unit won't be able to move more than one hex per turn if it has to cross multiple hexes of that terrain.

When hovering over the movement costs in the sidebar, the colors in the tooltip reflect the actual stats of that specific unit, so the costs for a quick Heavy Infantryman will be more greenish.

For the resistance colors, personally I think the color differences are large enough, also between 0% and -10%, and resistances of -100% and 100% do occur for mainline units (for example https://www.wesnoth.org/units/trunk/mainline/en_US/Mudcrawler.html and https://www.wesnoth.org/units/trunk/mainline/en_US/Fire%20Dragon.html ).

But I would be willing to change any of these things, I just wanted to explain my thoughts a bit.

CelticMinstrel:
> Currently, all attack type icons use the same shade of red, so at the very least, the situation wouldn't be worse with the new icons.
> 
> Changing the code to use the recoloring system for icons in magenta would be possible, it would be necessary to do this everywhere where icons are used (like the sidebar or the recruitment dialog), and the target colors would also have to be specified somewhere, perhaps in team-colors.cfg.
> 
> Custom damage type icons (see my comment from 11 Nov 2020) would then also require more than an image file.

Hmm, you make some good points here. In that case, I think we can probably consider potentially making the colours customizable as a separate change that can be done after this is merged. Colour-blindness is certainly something to think about it, but as you said, they're all red currently, so this change shouldn't make it any worse than it already was for colour-blind people.

walodar:
Hi, thanks. It's been a few days again, may I please ask which changes I have to make so this pull-request can be approved? It's been open for a long time, and it would be awesome if this could be merged.

Pentarctagon:
@stevecotton can confirm, but barring any further comments then what he last said still stands in terms of colors, aside from color blind support - that'd definitely be nice to have, but is also a larger change that's outside the scope of this PR.

stevecotton:
Don't worry about the color-blind support, I agree that that should be for another PR.

> So basically, if a unit does have ±100, it would use the exact same colour as ±60?

That's my suggestion, with the idea that +60 is already so large that the player can ignore a damage source, and -60 already forces the player to be especially alert to where a unit such as a white mage can move.

> In my proposal, the color is relative to the movement points of that particular unit, where pure yellow means moving in that terrain costs half of the unit's movement points, and even more reddish colors mean the unit won't be able to move more than one hex per turn if it has to cross multiple hexes of that terrain.

@CelticMinstrel and @max-torch, do you have thoughts on this point? My feeling is that the player's thoughts will already be either
* "I'm looking at a scout, it'll have a movement radius roughly *[mental image]* big", and so the information to communicate on the details view is "the movement radius will be half-size on these terrains". Here, relative scaling would color 1, 2 and 3 cost terrains as shades of green, but "a scout can still move over a lot of these hexes" doesn't communicate much to the player.
* "I'm looking at a tank, it's already a very small movement radius". In either relative or absolute scaling there's a lot of visible differentiation between 1 and 2 cost terrains.
* "I'm in a cave with humans/elves, quick units get an extra hex of movement". Here relative scaling on the 5mp vs 6mp units could be helpful, but in this situation I'd expect the player to be looking for the word "quick" or just looking at the highlighted hexes when selecting a unit on the map, rather than looking at the sidebar's "mp" tooltip or the details view.

Putting an additional display showing how many hexes that is might be an option:
image

walodar:
Hi, thank you. This new commit should include the requested changes.

The color of the movement costs is now fixed for all units, and the same as it was for the Heavy Infantryman. The nice new idea with the number of hexes is also included.

For the resistances, the end points of the full color spectrum are now reached at -60% and +60%, which increases the color differences for every intermediary step.

Addionally, I had a simple idea that should make the icons more distinguishable for colorblind people: I mirrored the inner part of the blade icon, so while the pierce icon still shows a bottom-left to top-right line, for the new blade icon it is now bottom-right to top-left. If we imagine the enemy is in the top-right direction, this could also be somewhat appropriate for the mode of attack.

Below is a screenshot for the Lich and the Heavy Infantryman. Thanks again!

comparison20220723

CelticMinstrel:
The only thing I don't like is scaling down the terrain alpha map for the hex indicator.

I'm wondering whether it makes sense to use U+2B23 (HORIZONTAL HEXAGON) ⬣ for this. The character does exist in DejaVu Sans but not in Lato. Using a font would probably make it look better, but might make it dependent on translations to include a fallback that contains the character.

If that doesn't seem acceptable, I think we should at least use an image sized specifically for this purpose.

max-torch:
**On the resistance color scale**
Okay, so what was done about the resistance colors was that the continuous color scale, which is red as lowest and green as highest, was remapped, such that now -60 is the reddest and +60 is the greenest. hmmm, well, I guess that can make sense if truly the range of possible values spans from -60 to +60. 

A question that comes to my mind is, should frequency of certain resistances be taken into account? Like for instance, if you will very rarely ever see, let's say -60, like for example if only one unit had it (I don't know how many units) , should we just make both -60 and the next value after -60 both equally be the reddest? Or should we absolutely map the color scale regardless of frequency of appearance of certain resistances?

**On the blade and pierce icons**
If it's just my opinion I like it. It doesn't look off for a normal person while is considerate for a color-blind person.

**On Hex symbols beside the movement costs**
I'm trying to put myself in the shoes of a new player for this. I like how it further engages your visual sense and perception to ingest the information about how much a unit can move. I'm worried about players seeing this before they learn about the quick trait... would there be a way to convey the effect of _quick_? Or is it implicitly understood that these are just the standard range of hexes for the unmodified with _quick_ traits unit?
Maybe a button beside the "Movement cost" word that if clicked, will adjust the number of hexes if the unit has _quick_, and the button only appears on units that can have quick?
I agree with celmin on making the hex symbol look nicer.

**On how to take this comment**
These comments are not a statement of requirement such that "if you follow all the changes in this comment, that's what we're asking for and then this is good for merging". I am raising these points, and I am wondering if others have consensus on them, and if there is then further discussion can say that we confirm that we want these things and then the PR author can go and put in the work to implement them.

walodar:
Hi, thanks for the new comments.

The Unicode hexagon is a great idea, the last commit now uses it.

Since it is only a string now, the number of hexes is now also shown in the tooltips in the sidebar or the recruitment dialog.

The hexagons are printed without selecting a specific font, and are not marked as translatable. On my Linux system, this works well for any chosen language, I hope that's true everywhere.

Regarding the last comment: The resistance color scale is probably a matter of personal preference, for me, simply always scaling from -60% to 60% is fine, but I would still be willing to make changes.

That the number of hexes is now also shown in the tooltip for any unit should hopefully alleviate the mentioned issue with the quick trait a bit.

The screenshot below focuses on the Bowman, who normally has 5 MP. From left to right we have a part of the help page, the MP tooltip in the recruitment dialog, and the MP tooltip in the sidebar for a Bowman with the trait quick, which gives him 6 MP.

Both for the color scale and the movement costs, I am not sure it would be good to make the help page even more complicated. Other traits also affect some values, though probably with less eye-catching consequences.

Thank you again!

bowman

CelticMinstrel:
> Or is it implicitly understood that these are just the standard range of hexes for the unmodified with _quick_ traits unit?

I think this is the case. When you view unit details in help, as far as I know it only shows details of the unit archetype, not the specific unit instance. If it did show the specific unit instance, that should be made clear somewhere.



> Maybe a button beside the "Movement cost" word that if clicked, will adjust the number of hexes if the unit has _quick_, and the button only appears on units that can have quick?

Adding buttons in the help would require an entirely new system, and I don't think that's worthwhile for just this one use-case.

Pentarctagon:
I sounds like this is good to go at this point then?

walodar:
Hi, this question was probably not addressed to me, but just in case: For me, this would be good to go. I would be very happy if this could be merged.

Pentarctagon:
It wasn't directed at anyone in particular - if all comments have been addressed and nobody has anything else, then I'll probably merge this in the next day or two.

stevecotton:
I'm generally feeling this is good and ready to merge, just a slight reservation about what happens when a unit has 10 or more movement points along with a defense cap. However, my feeling is we should merge and then look at that edge-case later, it seems that the Gryphon Master is the only 10mp unit in core, and there aren't any more mobile units in Ageless.

Gryphon Master (unit in core with 10 moves and a defense cap):
image

Changing that to 28 max moves:
image

Vultraz (approved):

Pentarctagon:
What happens if a newline is inserted every 5 movement points?

walodar:
Hi, thanks for the comments, and for discovering this issue.

Inserting newline characters would place the next hexagons at the beginning of a new row in the table, I think that would not be ideal.

Especially since fortunately, there is a lot of usable width left anyway.

I have taken the liberty of adding a small improvement, the width of all columns should now be correctly automatically adjusted.

The screenshots show the help page for the Gryphon Master with 10 MP and also for a modified Gryphon with 28 MP, like above.

GryphonMaster

Gryphon28MP

Pentarctagon:
The usable width could be significantly reduced depending on the window size though.

Pentarctagon:
Anyway, it's a minor issue that can be addressed later, as stevecotton said.

Thanks for continuing with this despite how long it took for us to get back around to it!

grz0:
@walodar I reported a small issue related to your PR - #6936

CelticMinstrel:
You're sure that's caused by this PR?

Wedge009:
I don't think it is - see my remarks in #6936.

eyalroz:
Having just noticed this discussion, I have to say, as a player, that - overall, this is a change for the worse. A bit of color hinting - not strong colors - for the resistances may arguably have been a good idea, and I'm not even sure about that. All of the rest - I disapprove, and most especially the movement cost hexes as argued at the linked issue.

> Using colors to make relevant information stand out at a quick glance.

Nope, it makes information be drowned out. Hurts the eyes and make the player want to look away. At a quick glance I notice less than I did before.

PR #5266, opened 2020-11-08
Add addon version selection dropdown list

kabachuha:
Replaced the version display label in the addons manager with a dropdown list displaying all the downloadable versions in descending order with the latest version chosen by default. The selector is frozen for the pubishable addons on their .pbl set version.

The tooltips with useful data like the upload time and the pack size will be added later (not in this pr) as they require some work on the server side.

Tested on my Wesnoth instance.

### DIFF
--- a/data/gui/window/addon_manager.cfg
+++ b/data/gui/window/addon_manager.cfg
@@ -125,16 +125,13 @@
 									[column]
 										border = "left,top,bottom"
 										border_size = 5
-										horizontal_alignment = "left"
-
-										[scroll_label]
-											id = "version"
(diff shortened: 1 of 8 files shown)


Pentarctagon:
Before and after screenshots would be useful.

Vultraz:
Yes, a screenshot would be useful.

kabachuha:
@Vultraz The screenshots:

Before changes. You can see a plaintext version label to the right of "Version:".



The addons manager with the dropdown added.



How versions are listed in it.



Installed the previous version of the addon by selecting it from the list and clicking "install".

Issue #5273, opened 2020-11-11
FR: Support per-campaign WML unit tests


stevecotton:
Feature request: unit tests for campaigns' WML

Illustrative example: SotA has macros `SAVE_ZOMBIE_LIST` and `LOAD_ZOMBIE_LIST`. Bug #5256 could have easily been caught by a unit test, because the root cause is that `{SAVE_ZOMBIE_LIST variable_name}` ignores the `variable_name` and thus overwrites the wrong variable.

I'd like to be able to have a unit-test file that's loaded with the appropriate campaign's WML available, so that it can unit test stuff in the campaign's utils folder.

Pentarctagon:
Campaign WML can already be made available by starting wesnoth while providing the particular campaign define, I think?

soliton-:
Issues like that could also trigger a warning in the preprocessor. A macro with an unused parameter seems suspicious to begin with. Not sure if it even could be an error or there are some legitimate use cases.

stevecotton:
About to open a PR adding this (posting it here because I'm still checking it compiles before pushing it):

```
	if(cmdline_opts.preproc_define || cmdline_opts.preprocess_input_macros || cmdline_opts.preprocess_path) {
		// It would be good if this was supported for running tests too, possibly for other uses.
		// For the moment show an error message instead of leaving the user wondering why it doesn't work.
		std::cerr << "That --preprocess-* option is only supported when using --preprocess or --validate-wml.\n";
		return -1;
	}
```

soliton-:
`run_wml_tests --additional_arg=--campaign=my_campaign --additional_arg=--campaign-difficulty=NORMAL --additional_arg=--campaign-scenario=my_campaign_unit_test`

Does that do anything useful? (Given an actual campaign and scenario ID.)

stevecotton:
@soliton- in do_gameloop(), the play_test() and unit_test() paths run and quit before we reach the call to goto_campaign(). Those options currently won't affect anything run via the unit test support.

stevecotton:
The unit-test framework also seems to run without the terrain macros enabled. I tried dropping `#define CAMPAIGN_SECRETS_OF_THE_ANCIENTS` into the campaign's _main.cfg, but running with `--unit-test` gots errors about `NEW:FENCE` not being defined.

stevecotton:
Clarification of what's blocking this: it's mainly about preprocessing, rather than campaigns.

[...]

PR #5274, opened 2020-11-11
Show an error message when --preprocess-defines is ignored

stevecotton:
While making it work would be better, this is a minimial change to at least
get the developer back on track instead of wasting time trying to work out
why putting this in a campaign's _main.cfg file doesn't work:

    #ifdef ENABLE_CAMPAIGN_TESTS
        {./test/}
    #endif

Context for this is in #5273. @Pentarctagon thanks for the suggestion, but don't worry I was already trying this and wasting the time before you posted it.


stevecotton:
The build-error is just that a build took too long, AFAICS it happened while setting up the build env before building the code changed in this PR.

PR #5275, opened 2020-11-11
SotA: Fix Ardonna's zombie list after Rastabahn's scenarios

### LINKED ISSUE #5256: SotA - Rastaban zombie recruit list overwrites Ardonna list
The Battle for Wesnoth version 1.15.6+dev (9e91ddb1fee-Modified)
### Describe the bug
Zombie recruit list for Ardonna in scenario "Battleground" is the list from Rastaban in previous scenario.  No wolves or rats, but there are swimmers.

### To Reproduce
Just playing through the campaign, I don't think I did anything special.  Start-of-scenario savefile for "Battleground" is attached.

### Expected behavior
Zombie recruit list would be the same as the last Ardonna scenario.

### Savefiles
[SotA-Battleground.gz](https://github.com/wesnoth/wesnoth/files/5472364/SotA-Battleground.gz)

stevecotton:
Fixes #5256, which was that `{SAVE_ZOMBIE_LIST rastabahn_zombies}` overwrote
the ardonna_zombies variable.

Move the MERGE_ZOMBIE_LIST code to be next to the LOAD and SAVE code, as any
change to one of these functions should probably be done to all of them.

There's a unit test for this in a commit that's auto-attached itself to #5273, but there isn't the framework support to run it - instead it needs #define statements pushed into the toplevel data/_main.cfg file.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
    * Secrets of the Ancients:
      * Revisions to Bone Captain
      * Scenario 2 uses new Iron Fence terrain, in preparation for potential map revisions.
+     * While the two leaders are separated, only the currently active one learns new zombie types.
    * Sceptre of Fire:
      * Some revisions/bugfixes to Scenario 9 "Caverns of Flame"
  ### Lua API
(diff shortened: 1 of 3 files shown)


doofus-01:
Thanks for fixing, I'll test it out tomorrow.

doofus-01 (approved):
Tested with Battleground scenario, the recruit list has been fixed.

PR #5276, opened 2020-11-11
SotA: Don't concatenate strings in the recruit dialog

stevecotton:
Instead of trying to concatenate translatable strings, just make a single full
string to be shown in the recruitment dialog. This came from trying to update
the i18n hints, and realising my suggest of making the swimmers "Merfolk" in
f6dc83375 was causing problems when they could also be nagas.

These strings are stored in a variable at the start of the campaign (in
ZOMBIE_INIT), so won't appear when loading an existing save game. This is why
the Lua script keeps some support for the old names (although they're not going
to be translated in that case).

@ivanovic and @shikadiqueen I believe this is better for translators, but is it?
I'd also like to ask and apologise to our Polish translator, who has already translated
the 1.15.6 changes from f6dc83375.

### DIFF
--- a/data/campaigns/Secrets_of_the_Ancients/units/SotA_Corpses.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/units/SotA_Corpses.cfg
@@ -84,10 +84,13 @@
 
 #textdomain wesnoth-sota
 [unit_type]
-    # This unit appears in the recruit list as soon as the player has access to more than
-    # one type of walking corpse. Trying to recruit it launches the zombie recruit dialog.
     id=SotA Walking Corpse_Dummy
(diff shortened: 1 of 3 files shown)


stevecotton:
Positive feedback from the translators' thread: https://r.wesnoth.org/p659939

stevecotton:
@shikadiqueen and @ivanovic, I'd like to get this in in time for the next pot-update, please would you review it?

ivanovic (approved):

CelticMinstrel:
Just for the record since you requested my review, I think this is fine.

shikadiqueen:
Seems fine to me.

shikadiqueen (approved):

PR #5280, opened 2020-11-12
SoF: fix #5278 in scenario 4 and a small fix for scenario 2p5

### LINKED ISSUE #5278: Item placement bug in SoF S4, possibly because of [store_locations]
The bug was reported for Wesnoth 1.15.4 here https://forums.wesnoth.org/viewtopic.php?p=658095#p658095
I have confirmed the behavior on 1.15.6+dev and have confirmed that the behavior does not occur on 1.15.3 (IIRC, it also didn't occur for the version(s) of 1.15.3+dev I was working on at the time).

### Describe the bug
The placement of the coal and gold locations in SoF scenario 4 doesn't work as intended anymore. Coal always gets placed on the keeps of the respective sides leaders that serve as filter anchors, while gold never gets placed at all (presumably because no keep location is matching its distance filter). _This currently makes the scenario uncompleteable_ (without debug).

### To Reproduce
Load the attached save any number of times and remove shroud via debug.

### Expected behavior
Edit: important context: the scenario map is in part generated randomly, which leads to the issue at hand.
Until some point in the 1.15.4 development, the items were still placed correctly at a random location somewhere in the area specified by the respective filters (starting at ~ line 400 of the scenarios code). I have tested that reverting my changes to that filter (commit 7e00c45) does not impact the behavior on 1.15.6+dev.
I assume the bug lies in the behavior of the `[store_locations]` action or something related and not the scenario code.

### Savefiles
[SoF-Gathering Materials.gz](https://github.com/wesnoth/wesnoth/files/5527156/SoF-Gathering.Materials.gz)

### Additional context
I also took a look at the data state in some save files. For the one provided in the forum thread linked above (with the 2 coal deposits already picked up by the player), resource locations are saved as follows. Note the weird values for gold in the first case and the occurrence of Strings in both cases.

```
	[coal_1]
			x=0
			y=0
		[/coal_1]
		[coal_2]
			x=0
			y=0
		[/coal_2]
		[gold_1]
			x="-1].x"
			y="-1].y"
		[/gold_1]
```

In one of the scenario starts I created via the attached file, the locations for turn 1 are given as:

```
		[coal_1]
			x=32
			y=33
		[/coal_1]
		[coal_2]
			x=10
			y=8
		[/coal_2]
		[gold_1]
			x=""
			y=""
		[/gold_1]
```

IskarJarak:
SoF S2p5:
-  increased the turn limit by 1 because the scenario felt _very_ tight on time even on easy difficulty (and my changes in 1.15.4 slightly increased minimum travel distance).
- the victory event filters were only triggered by Rugnur's moveto events; this is inconsistent with the scenario objectives and could also lead to costing the player a full turn if Rugnur wasn't the last hero moved to the tunnels end. I changed the event to trigger on any player movement while the heroes reached their destination. Tested to work for all three heroes.

SoF S4:
- reverted my commit from 1.15.4 to fix #5278 
- re-added the functionality of that commit and extended it to also encompass rails
- tested this extensively...

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,8 @@
      * Revisions to Bone Captain
      * Scenario 2 uses new Iron Fence terrain, in preparation for potential map revisions.
    * Sceptre of Fire:
+     * Scenario 2p5: increased the turn limit by 1 and enhanced victory filter conditions
+     * Scenario 4: fixed a bug that prevented resources from spawning and added some small flavor adjustments
      * Some revisions/bugfixes to Scenario 9 "Caverns of Flame"
  ### Lua API
  ### Multiplayer
--- a/data/campaigns/Sceptre_of_Fire/scenarios/2p5_Reaching_the_Runecrafter.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/2p5_Reaching_the_Runecrafter.cfg
@@ -2,7 +2,7 @@
 [scenario]
     name= _ "Reaching the Runecrafter"
     id=2p5_Reaching_the_Runecrafter
-    turns=15
+    turns=16
     map_file=2p5_Reaching_the_Runecrafter.map
     next_scenario=3_Searching_for_the_Runecrafter
     victory_when_enemies_defeated=no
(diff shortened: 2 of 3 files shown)


nemaara (approved):
As long as nothing breaks with these changes they seem fine.

Issue #2932, opened 2018-04-21
make a scenario editor tool for defining multihex areas without tod assigned.


gfgtdf:
NEW POST:

Since [time_area] does not necessarily contains time information anymore it'd be nice to rename it to just [area]

OLD POST: 

there is a 'tip' that says 

> Use the map editor's <i>Areas</i> menu to define regions later addressed by event or ai coding.

but currently the editors area feature can only add time areas not general areas that do not change the tod of those hexes, making it not very usable for that case. 

We also have the 'special locations' feature that can add single hexes. but that cannot really hande multihex areas.

It wodul be nice if the wml devs would not have to need to deal with 3 paralelly exiustent systems so it'd be nice if we could maybe extend one of the systems mentioned abive to hande this case.

CelticMinstrel:
I was actually wondering earlier if it's possible to have a `[time_area]` that actually doesn't change the underlying schedule. For example, what happens if the `[time_area]` contains no time tags? It might use the underlying schedule or it might use a null schedule, and I think the former would make more sense.

gfgtdf:
> For example, what happens if the [time_area] contains no time tags? It might use the underlying schedule or it might use a null schedule

Last time i tested, which is quite some time ago it used the null scedule, and i'm not aware of any changes to that.

gfgtdf:
hmm maybe we should also rename it to just [area] then ?

CelticMinstrel:
I'm not sure. I don't particularly mind it being called `[time_area]` even though it may not actually contain a schedule, but I suppose people could get confused by this. On the other hand, I'm not keen on `[area]` when it does contain a schedule.

gfgtdf:
See also https://github.com/wesnoth/wesnoth/pull/4601

stevecotton:
Thinking about whether to rename `[time_area]` to `[area]`, I prefer the longer name, as it's not a generic word and is therefore easier to search for in C++ and Lua code. If we renamed it, I'd suggest `[named_area]`, but in this case I think simply not renaming it is best - I think we'd end up supporting both names forever, as there would be little reason to remove the deprecated one.

Whatever the name is, I think the same name should work both with and without a schedule. The WML author might be deciding whether to implement something via moveto events or simply put in a schedule, and it doesn't make sense to change the tag just to decide whether or not to recognise the schedule.

[...]

PR #5282, opened 2020-11-14
Editor - add top-level WML tag to generated scenarios and change file type detection

Pentarctagon:
Also removes a few typedefs that were defined and then not used at all or only used once.

### DIFF
--- a/src/editor/map/context_manager.cpp
+++ b/src/editor/map/context_manager.cpp
@@ -907,8 +907,7 @@ void context_manager::load_map(const std::string& filename, bool new_context)
 		}
 
 		if(get_map_context().is_embedded()) {
-			const std::string& msg = _("Loaded embedded map data");
-			gui2::show_transient_message(_("Map loaded from scenario"), msg);
+			LOG_ED << "Loaded embedded map data\n";
 		} else {
 			if(get_map_context().get_filename() != filename) {
 				if(get_map_context().get_map_data_key().empty()) {
@@ -916,14 +915,7 @@ void context_manager::load_map(const std::string& filename, bool new_context)
 						<< filename << " -> " << get_map_context().get_filename()
 						<< " with no apparent scenario load\n";
 				} else {
-					utils::string_map symbols;
-					symbols["old"] = filename;
-					const std::string& msg = _("Loaded referenced map file:\n$new");
-					symbols["new"] = get_map_context().get_filename();
-					symbols["map_data"] = get_map_context().get_map_data_key();
-					gui2::show_transient_message(_("Map loaded from scenario"),
-						//TODO: msg is already translated does vgettext make sense?
-						VGETTEXT(msg.c_str(), symbols));
+					LOG_ED << "Loaded referenced map file:\n"+get_map_context().get_filename();
 				}
 			}
 		}
--- a/src/editor/map/context_manager.cpp
+++ b/src/editor/map/context_manager.cpp
@@ -669,7 +669,7 @@ void context_manager::save_scenario_as_dialog()
 	dlg.set_title(_("Save Scenario As"))
 	   .set_save_mode(true)
 	   .set_path(input_name)
-	   .set_extension(".cfg");
+	   .set_extension(".ed.cfg");
 
 	if(!dlg.show()) {
 		return;
--- a/src/editor/map/map_context.cpp
+++ b/src/editor/map/map_context.cpp
@@ -30,6 +30,7 @@
 #include "terrain/type_data.hpp"
 #include "units/unit.hpp"
 #include "game_config_view.hpp"
+#include "gui/dialogs/transient_message.hpp"
 
 #include <boost/regex.hpp>
 
@@ -128,14 +129,14 @@ map_context::map_context(const game_config_view& game_config, const std::string&
 	 * 2. The file contains an editor generated scenario file.
 	 *    * embedded_ = false
 	 *    * pure_map_ = false
-	 * 3. A scenario embedding the map
+	 * 3. The map file is referenced by map_data={MACRO_ARGUMENT}.
+	 *    * embedded_ = false
+	 *    * pure_map_ = true
+	 * 4. A scenario embedding the map
 	 *    * embedded_ = true
 	 *    * pure_map_ = true
 	 *    The data/scenario-test.cfg for example.
 	 *    The map is written back to the file.
-	 * 4. The map file is referenced by map_data={MACRO_ARGUEMENT}.
-	 *    * embedded_ = false
-	 *    * pure_map_ = true
 	 */
 
 	log_scope2(log_editor, "Loading file " + filename);
@@ -154,12 +155,7 @@ map_context::map_context(const game_config_view& game_config, const std::string&
 	}
 
 	// 1.0 Pure map data
-	boost::regex rexpression_map_data(R"""(map_data\s*=\s*"(.+?)")""");
-	boost::smatch matched_map_data;
-
-	if(!boost::regex_search(
-		file_string, matched_map_data, rexpression_map_data, boost::regex_constants::match_not_dot_null)
-	) {
+	if(filesystem::ends_with(filename, ".map") || filesystem::ends_with(filename, ".mask")) {
 		map_ = editor_map::from_string(game_config, file_string); // throws on error
 		pure_map_ = true;
 
@@ -167,63 +163,64 @@ map_context::map_context(const game_config_view& game_config, const std::string&
 		return;
 	}
 
-	const std::string& map_data = matched_map_data[1];
-
-	boost::regex rexpression_macro(R"""(\{(.+?)\})""");
-	boost::smatch matched_macro;
-
-// TODO: think about if there are any better ways to determine an editor generated scenario file vs a user generated scenario file
-//       now that it can't be done by checking for the presence of a top-level tag
-//       replace file contents inspection with extension checks
-//       * if .map -> pure map file
-//       * if .ed.cfg or file path containts "/editor/" -> editor generated scenario
-//       * if .cfg -> player created scenario
-	if(!boost::regex_search(map_data, matched_macro, rexpression_macro)) {
-		// We have a map_data string but no macro ---> embedded or scenario
-
-		if(file_string.find(editor_generated) == 0) {
-			// 2.0 Editor generated scenario
-			LOG_ED << "Loading generated scenario file from editor prior to top-level tags having been added" << std::endl;
-			try {
-				load_scenario(game_config);
-			} catch(const config::error& e) {
-				// We already caught and rethrew this exception in load_scenario
-				throw editor_map_load_exception("load_scenario", e.message);
-			}
-		} else {
-			// 3.0 Embedded map from non-editor generated scenario
-			LOG_ED << "Loading embedded map file from non-editor generated file" << std::endl;
-			embedded_ = true;
-			pure_map_ = true;
-			map_ = editor_map::from_string(game_config, map_data);
+	// 2.0 Editor generated scenario
+	if(filename.find("/editor/") != std::string::npos || filesystem::ends_with(filename, ".ed.cfg")) {
+		LOG_ED << "Loading editor generated scenario file" << std::endl;
+		try {
+			load_scenario(game_config);
+		} catch(const config::error& e) {
+			// We already caught and rethrew this exception in load_scenario
+			throw editor_map_load_exception("load_scenario", e.message);
 		}
+		add_to_recent_files();
+		return;
+	}
+
+	// 3.0 Macro referenced pure map
+	std::size_t map_data_macro_start = file_string.find("map_data=\"{");
+	if(filesystem::ends_with(filename, ".cfg") && map_data_macro_start != std::string::npos) {
+		std::size_t map_data_macro_end = file_string.find("}\"", map_data_macro_start+11);
+		const std::string& macro_argument = file_string.substr(map_data_macro_start+11, map_data_macro_end-(map_data_macro_start+11));
+
+		LOG_ED << "Map looks like a scenario, trying {" << macro_argument << "}" << std::endl;
+
+		std::string new_filename = filesystem::get_wml_location(macro_argument,
+			filesystem::directory_name(filesystem::get_short_wml_path(filename_)));
+
+		if(new_filename.empty()) {
+			std::string message = _("The map file looks like a scenario, "
+									"but the map_data value does not point to an existing file")
+								+ std::string("\n") + macro_argument;
+			throw editor_map_load_exception(filename, message);
+		}
+
+		LOG_ED << "New filename is: " << new_filename << std::endl;
+
+		filename_ = new_filename;
+		file_string = filesystem::read_file(filename_);
+		map_ = editor_map::from_string(game_config, file_string);
+		pure_map_ = true;
 
 		add_to_recent_files();
 		return;
 	}
 
-	// 4.0 Macro referenced pure map
-	const std::string& macro_argument = matched_macro[1];
-	LOG_ED << "Map looks like a scenario, trying {" << macro_argument << "}" << std::endl;
+	// 4.0 Embedded map from non-editor generated scenario
+	std::size_t map_data_embedded_start = file_string.find("map_data=\"");
+	if(filesystem::ends_with(filename, ".cfg") && map_data_embedded_start != std::string::npos) {
+		std::size_t map_data_embedded_end = file_string.find("\"", map_data_embedded_start+10);
+		const std::string& map_data = file_string.substr(map_data_embedded_start+10, map_data_embedded_end-(map_data_embedded_start+10));
 
-	std::string new_filename = filesystem::get_wml_location(macro_argument,
-		filesystem::directory_name(filesystem::get_short_wml_path(filename_)));
-
-	if(new_filename.empty()) {
-		std::string message = _("The map file looks like a scenario, "
-								"but the map_data value does not point to an existing file")
-							  + std::string("\n") + macro_argument;
-		throw editor_map_load_exception(filename, message);
+		LOG_ED << "Loading embedded map file from non-editor generated file" << std::endl;
+		embedded_ = true;
+		pure_map_ = true;
+		map_ = editor_map::from_string(game_config, map_data);
+		add_to_recent_files();
+		return;
 	}
 
-	LOG_ED << "New filename is: " << new_filename << std::endl;
-
-	filename_ = new_filename;
-	file_string = filesystem::read_file(filename_);
-	map_ = editor_map::from_string(game_config, file_string);
-	pure_map_ = true;
-
-	add_to_recent_files();
+	ERR_ED << "Unable to load file: " << filename << std::endl;
+	gui2::show_transient_message(_("Failed to load file"), _("Attempted to load an unknown type of file. Filenames must end in either '.map', '.cfg', or '.ed.cfg'."));
 }
 
 map_context::~map_context()
@@ -593,6 +590,11 @@ bool map_context::save_scenario()
 
 	if(scenario_id_.empty()) {
 		scenario_id_ = filesystem::base_name(filename_, true);
+		// base_name removes everything after the last period, not the first period
+		// so the .ed part needs to be removed manually
+		if(filesystem::ends_with(scenario_id_, ".ed")) {
+			scenario_id_ = filesystem::base_name(scenario_id_, true);
+		}
 	}
 
 	if(scenario_name_.empty()) {
@@ -602,7 +604,7 @@ bool map_context::save_scenario()
 	try {
 		std::stringstream wml_stream;
 		wml_stream
-			<< editor_generated << "\n"
+			<< "# This file was generated using the scenario editor.\n"
 			<< "#\n"
 			<< "# If you edit this file by hand, then you shouldn't use the\n"
 			<< "# scenario editor on it afterwards. The editor completely\n"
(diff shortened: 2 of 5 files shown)


src/game_initialization/create_engine.cpp
```diff
@@ -741,7 +746,7 @@ void create_engine::init_all_levels()
 		}
 
 		if(type == "sp" || type.empty() || (type == "hybrid" && !mp)) {
-			campaign_ptr new_sp_campaign(new campaign(data));
+			std::shared_ptr<campaign> new_sp_campaign(new campaign(data));
```
Vultraz, line 749: Should be `make_shared`

src/game_initialization/create_engine.cpp
```diff
@@ -692,7 +692,12 @@ void create_engine::init_all_levels()
 				continue;
 			}
 
-			scenario_ptr new_scenario(new scenario(data));
+			// previous editor format had no top-level tag present
+			if(data.has_child("multiplayer"))
+			{
+				data = data.child("multiplayer");
+			}
+			std::shared_ptr<scenario> new_scenario(new scenario(data));
```
Vultraz, line 700: Should be `make_shared`

Vultraz (changes requested):

CelticMinstrel:
I think maybe it should support choosing between a "multiplayer" tag and a "scenario" tag.

Pentarctagon:
[multiplayer] is essentially [scenario] except it supports some additional attributes and maybe has some different defaults though, I think?  The current editor setup also can't be used with any single player campaigns since, at a minimum, it would remove the `next_scenario` attribute every time the scenario was saved.

src/editor/map/map_context.cpp
```diff
@@ -122,21 +123,20 @@ map_context::map_context(const game_config_view& game_config, const std::string&
...
 	 *    0.2 Map file empty
-	 *    0.3 No valid data
 	 * 1. It's a file containing only pure map data.
 	 *    * embedded_ = false
 	 *    * pure_map_ = true
-	 * 2. A scenario embedding the map
+	 * 2. The file contains an editor generated scenario file.
+	 *    * embedded_ = false
+	 *    * pure_map_ = false
+	 * 3. The map file is referenced by map_data={MACRO_ARGUEMENT}.
```
stevecotton, line 132: Only one E in 'argument'

src/editor/map/map_context.cpp
```diff
@@ -155,71 +155,72 @@ map_context::map_context(const game_config_view& game_config, const std::string&
 	}
 
 	// 1.0 Pure map data
-	boost::regex rexpression_map_data(R"""(map_data\s*=\s*"(.+?)")""");
-	boost::smatch matched_map_data;
-
-	if(!boost::regex_search(
-		file_string, matched_map_data, rexpression_map_data, boost::regex_constants::match_not_dot_null)
-	) {
+	if(filename.rfind(".map")+4 == filename.size()) {
```
stevecotton, line 158: While string::ends_with() won't be available until C++20, we already have filesystem::ends_with().

src/editor/map/map_context.cpp
```diff
@@ -586,6 +590,11 @@ bool map_context::save_scenario()
 
 	if(scenario_id_.empty()) {
 		scenario_id_ = filesystem::base_name(filename_, true);
+		// base_name removes everything after the last period, not the first period
+		// so the .ed part needs to be removed manually
+		if(scenario_id_.find(".ed")+3 == scenario_id_.size()) {
+			scenario_id_ = scenario_id_.substr(0, scenario_id_.find(".ed"));
```
stevecotton, line 596: This line could use filesystem::base_name again

src/editor/map/map_context.cpp
```diff
@@ -595,7 +604,7 @@ bool map_context::save_scenario()
 	try {
 		std::stringstream wml_stream;
 		wml_stream
-			<< "# This file was generated using the scenario editor.\n"
+			<< editor_generated << "\n"
```
stevecotton, line 607: Why does exactly one line of this move to the .hpp file? Why not all or none?

Pentarctagon: What happened was basically:
0) "It's silly that the editor isn't writing out a valid scenario cfg."
1) I added the [multiplayer] tag.
2) I realized that the presence of a top-level tag was being used by the editor to determine if it was an editor generated scenario or not, which wouldn't work anymore.
3) I decided to instead use the presence of that comment string to determine if the file was an editor generated scenario or not.
4) I decided that was also horrible, and I should just use file extensions instead.

Short version - I should just move it back to the cpp file with the rest of that comment block.

Pentarctagon:
The above feedback has been addressed in the last push.

CelticMinstrel:
> The current editor setup also can't be used with any single player campaigns since, at a minimum, it would remove the `next_scenario` attribute every time the scenario was saved.

The 1.14 scenario editor works for _any_ campaign - you simply transclude the scenario file into your own scenario file which adds the tags and any extra attributes required. Even if the scenario editor is only going to be for multiplayer though, it still needs to support `next_scenario`.

I think the ability to keep generated stuff and hand-written stuff completely separate is good. That can still be done when adding an enclosing tag as you do here, but it means the tag name needs to be selectable (scenario or multiplayer). Then instead of writing this:

```
[scenario]
    {scenarios/my-scenario.cfg}
    # events and stuff
[/scenario]
```

You would write this:

```
{scenarios/my-scenario.cfg}

[+scenario]
    # events and stuff
[/scenario]
```

If you're _not_ going to keep generated and hand-written stuff separate, then the scenario editor needs to write back any unknown keys and tags verbatim, without altering or removing them. (Order of keys doesn't actually matter, so shuffling that is sorta okay, though not ideal.)

> Use file extensions rather than file contents to determine how to load a file in the editor.

I don't object to this in principal, but if you're going to do it, please also recognize .mask as a map file… either that or fall back on the old content checking if the extension isn't recognized.

CelticMinstrel:
> [multiplayer] is essentially [scenario] except it supports some additional attributes and maybe has some different defaults though, I think?

Oh right, I wouldn't call this accurate either. One tag is used by `[campaign]`, the other by the multiplayer UI. They create the scenario in two different scopes. It's true that `[multiplayer]` supports some additional attributes and may have some different defaults, but that's not the key difference between them.

gfgtdf:
> I think the ability to keep generated stuff and hand-written stuff completely separate is good.

I strongly support this one, in particular since even with this, the editor will probably still discard hand written changes in particular comments.


> That can still be done when adding an enclosing tag as you do here, but it means the tag name needs to be selectable (scenario or multiplayer). Then instead of writing this:...

in 1.15 i'd actually use `[scenario ]map_file=scenarios/my-scenario.cfg` instead of  `[scenario] {scenarios/my-scenario.cfg}`

which has a similar effect but does some fixed in parituclar when the ediot .cfg hand sides in them, and as far i i can see this pr  also breaks  the `[scenario ]map_file=scenarios/my-scenario.cfg` syntax

Pentarctagon:
> I don't object to this in principal, but if you're going to do it, please also recognize .mask as a map file… either that or fall back on the old content checking if the extension isn't recognized.

Done.

> The 1.14 scenario editor works for any campaign - you simply transclude the scenario file into your own scenario file which adds the tags and any extra attributes required. Even if the scenario editor is only going to be for multiplayer though, it still needs to support next_scenario.

I have a hard time imagining that anyone who knows what transcluding is or needs to use it would be using the scenario editor's limited functionality in the first place instead of writing that small amount of WML themselves.

> in 1.15 i'd actually use [scenario ]map_file=scenarios/my-scenario.cfg instead of [scenario] {scenarios/my-scenario.cfg}

Why would providing a **scenario** cfg to an attribute called **map**_file even be a thing that's supported?

> as far i i can see this pr also breaks the [scenario ]map_file=scenarios/my-scenario.cfg syntax

It would, but again, I don't understand why that's a thing that anyone would expect to work like that.

Pentarctagon:
> If you're not going to keep generated and hand-written stuff separate, then the scenario editor needs to write back any unknown keys and tags verbatim, without altering or removing them. (Order of keys doesn't actually matter, so shuffling that is sorta okay, though not ideal.)

Not overwriting unknown WML when writing the scenario file back out is something I'd like to add, though as far as I know there's no way around the limitations of comments getting removed as well as any macros getting expanded.

Pentarctagon:
More generally though, if the expectation for the editor is that UMC authors continue to use it to manage the few things it can handle while also wrapping all their handwritten stuff around it, then IMO that sounds like an incredibly broken workflow.  My assumption has always been that people would use the editor for very simple things, and then stop using the scenario editor part of it if/when they started needing to hand-write WML.  If that's not the case, then I guess I'm not actually sure what I want to change about it anymore.

src/editor/map/context_manager.cpp
```diff
@@ -907,23 +907,15 @@ void context_manager::load_map(const std::string& filename, bool new_context)
...
 				} else {
-					utils::string_map symbols;
-					symbols["old"] = filename;
-					const std::string& msg = _("Loaded referenced map file:\n$new");
-					symbols["new"] = get_map_context().get_filename();
-					symbols["map_data"] = get_map_context().get_map_data_key();
-					gui2::show_transient_message(_("Map loaded from scenario"),
-						//TODO: msg is already translated does vgettext make sense?
-						VGETTEXT(msg.c_str(), symbols));
+					LOG_ED << "Loaded referenced map file:\n"+get_map_context().get_filename();
```
stevecotton, line 918: Should probably use operator<< instead of operator+

src/editor/map/map_context.cpp
```diff
@@ -122,21 +123,20 @@ map_context::map_context(const game_config_view& game_config, const std::string&
 	 * 0. Not a scenario or map file.
 	 *    0.1 File not found
 	 *    0.2 Map file empty
-	 *    0.3 No valid data
 	 * 1. It's a file containing only pure map data.
 	 *    * embedded_ = false
 	 *    * pure_map_ = true
-	 * 2. A scenario embedding the map
+	 * 2. The file contains an editor generated scenario file.
+	 *    * embedded_ = false
```
stevecotton, line 130: Either here or at the declaration of `embedded_`, please document more clearly what this means.

The current doc for embedded_ is "Whether the map context refers to a map embedded in a scenario file.", I think clearer would be "True if the file that will be written to by clicking Save contains data that shouldn't be overwritten by the editor, in addition to data that can be generated by the editor. False if the complete file should be overwritten by the serialized version of the parts that the editor understands."

src/game_initialization/create_engine.cpp
```diff
@@ -741,7 +746,7 @@ void create_engine::init_all_levels()
 		}
 
 		if(type == "sp" || type.empty() || (type == "hybrid" && !mp)) {
-			campaign_ptr new_sp_campaign(new campaign(data));
+			std::shared_ptr<campaign> new_sp_campaign = std::make_shared<campaign>(data);
```
stevecotton, line 749: Can use `auto` here.

I haven't re-read the discussion about what the editor should do yet - I'll probably add more comments when I've had to think about it, but that won't be in the next couple of days.

src/editor/map/map_context.cpp
Pentarctagon: I would say "Whether or not the map's terrain information is directly part of the scenario cfg file or exists as a separate '.map' file."

CelticMinstrel:
> More generally though, if the expectation for the editor is that UMC authors continue to use it to manage the few things it can handle while also wrapping all their handwritten stuff around it, then IMO that sounds like an incredibly broken workflow. My assumption has always been that people would use the editor for very simple things, and then stop using the scenario editor part of it if/when they started needing to hand-write WML. If that's not the case, then I guess I'm not actually sure what I want to change about it anymore.

Well, the few things it can handle are actually pretty useful to have a UI for - placing units and time areas and such. I don't think it's a good idea to make the scenario editor work in a way that's incompatible with handwritten WML, because if someone has already made their scenario in the editor and then find they need to handwrite some WML, they'll end up with something that can no longer be safely edited in the scenario editor.

> Not overwriting unknown WML when writing the scenario file back out is something I'd like to add, though as far as I know there's no way around the limitations of comments getting removed as well as any macros getting expanded.

Comments are another argument for keeping the generated parts isolated into a separate file.

> Why would providing a scenario cfg to an attribute called map_file even be a thing that's supported?

I actually agree that this doesn't make much sense on the surface, although it's not totally nonsensical - the scenario config _also_ includes an embedded map, after all. Still, it might be better to use a different attribute or something for it…

If the recommended way of using a scenario file together with handwritten WML becomes adding `scenario_file=path/to/generated-scenario.cfg`, then I think it actually makes sense to invent a new enclosing tag for generated scenario files - instead of `[scenario]` or `[multiplayer]`, maybe call it `[map]` or something? I'm not sure what's a good name, but I don't see that the choice of scenario vs multiplayer actually matters when it's not going to be directly loaded into the engine as part of parsing a `_main.cfg`. If it's only loaded by either gathering files in the editor folder _or_ by including it in another scenario tag, then the enclosing tag could be basically anything you want.

src/editor/map/map_context.cpp
stevecotton: It's being set to `false` here, but the terrain information is being saved in the .cfg file.

Pentarctagon: Then: "Whether the map data is from a non-editor generated scenario file."

stevecotton: But what does that mean? As in, if someone's reviewing a conditional statement based on this, how do they decide which behavior is the "should do this for true" and which is the "should do this for false"?

Pentarctagon: Which part of the description isn't clear? "should do this for true" would be when dealing with a scenario that's hand-written, and "should do this for false" would be in the opposite case.

gfgtdf:
> Why would providing a scenario cfg to an attribute called map_file even be a thing that's supported?

because the .cfg generates by the editor are  map data just with additional information in particular [(time_)area]s . I don't have a problem with the renaming them to .map as well (the .map would be used needed for old and new map files)

> More generally though, if the expectation for the editor is that UMC authors continue to use it to manage the few things it can handle while also wrapping all their handwritten stuff around it, then IMO that sounds like an incredibly broken workflow.


How so? having the map editor to edit map and map related things and doing the rest by hand sounds like a very normal workflow to me, basicially what we did for ages except that the map files contain a little more information.

 
> My assumption has always been that people would use the editor for very simple things, and then stop using the scenario editor part of it if/when they started needing to hand-write WML. If that's not the case-


This one the other hand sounds broken to be, since its very likely that you'll want to edit the map at some later point which is then not possible.


> then I guess I'm not actually sure what I want to change about it anymore

I really suggest the following:

1) make the editor always generate a config file that contains just map_data= and optional other attributes like [time_area]
2) integrate all tools that make sense in the normal editor, in particular the [time_area] tool. maybe name rename what we currently call .cfg to .map . So that basically the current 'scenario editor' becomes the default mode. and the 'map only' mode is removed.
3) deprecate {.../map.map} in favor of map_file=.../map.map which can handle 'old' and 'new' maps
4) (This is would be nice but is a lot of work an probably won't happen) Add a 'scenario editor' that can generate whole scenarios or even addons ideally  with seperate .map and .cfg files 

> then I think it actually makes sense to invent a new enclosing tag for generated scenario files maybe call it [map] or something?

Maybe, No opinion on this one.

Pentarctagon:
A map file and a scenario file are two specific, different things though.  A map file has the `.map` extension and contains terrain information; a scenario file has the `.cfg` extension, contains WML/lua, as well as either embedded map data or references a `.map` file.

stevecotton:
As a scenario-author, I want to be able to use the scenario editor for things that should fit the terrain. For example, if I resize the map then I want the starting locations to move along with the castle-keep hexes that they're on.

LoW S03 and S07 have a map with lots of labels on, it would be good if those could be part of the map instead of being part of the scenario, again so that the editor can handle moving them if the map is resized.

AToTB S03 (Guarded Castle) has several more examples of things that should move with terrain:

Being able to place the guard units and scenery items would be good.

Being able to place the location for the "second wave" event would be good (this is supported in 1.14's `.map` files by the location_id, although that's not used in this map).

About half of the map is indoors, with a complex shape for that set of hexes. Resizing the map or adding an extra wing to the castle would need corresponding changes to the time area, and it would be nice to be able to do that visually instead of manually coding this:
https://github.com/wesnoth/wesnoth/blob/1cd0cc7d260fb09981ee3a7bbbe4b5fc39a1e59b/data/campaigns/Two_Brothers/scenarios/03_Guarded_Castle.cfg#L230-L233

Being able to define an area for a moveto trigger would be helpful, particularly for this event (the comment about it being exactly 40 hexes doesn't match the code, but ignore that):
https://github.com/wesnoth/wesnoth/blob/1cd0cc7d260fb09981ee3a7bbbe4b5fc39a1e59b/data/campaigns/Two_Brothers/scenarios/03_Guarded_Castle.cfg#L754-L762

gfgtdf:
> A map file and a scenario file are two specific, different things though. A map file has the `.map` extension and contains terrain information; a scenario file has the `.cfg` extension, contains WML/lua, as well as either embedded map data or references a `.map` file. That you can place something on a map doesn't automatically make it part of the map data.

You are here comparing the usual (handwritten) scenario .cfg files to the .map files generated by the editor, and currently the .cfg files by the map editor are something quite different than  what is  handwritten .cfg (this is most easily seen by the fact thath the the .cfg files generated by the editor lack the [scenario]/[multiplayer] tags). the .cfg extension of the editor just indicates that is uses the wml format, i still say the most natural way to see it is as an 'extended map file'

> Being able to define an area for a moveto trigger would be helpful


You can use the area= key in standard location filters, and define the [time_area] in the editor, (see also https://github.com/wesnoth/wesnoth/issues/2932 for renaming [time_area] to just [area])

Pentarctagon:
> You are here comparing the usual (handwritten) scenario .cfg files to the .map files generated by the editor, and currently the .cfg files by the map editor are something quite different than what is handwritten .cfg (this is most easily seen by the fact thath the the .cfg files generated by the editor lack the [scenario]/[multiplayer] tags). the .cfg extension of the editor just indicates that is uses the wml format, i still say the most natural way to see it is as an 'extended map file'

It's just a regular scenario cfg though, aside from missing a top-level WML tag.  I don't see how that makes it an "extended map file".

gfgtdf:
> It's just a regular scenario cfg though

No its since it doesn't contain events, lua code, variables and a probably quite some other stuff, I am not even sure sure if its playable at all if you don't add sides to it. It just has some attributes that scenario files also have.

Just becasue it contains a lof of keys that are also contained in [scenario] it doesnt maek it a full scenario, i


>  I  don't see how that makes it an "extended map file".

Because it only contains map-related information (or rather: should contain. In particular i support removing other data form in the default configuration, like information about [side]s), and because we define it like that, and because map_file= can load it. Or to be more precise: because it makes sense to see it like that, in particular because the 'editor for terrain+areas+labels included via map_file' workflow is easier to use and to understand than the 'editor generates scenario file which then cannot be edited after it was edited by hand' workflow


Also we are arguing about words and definitions here.

So far my main suggestion was to merge the normal mode with the scenario mode and 'extend' the editor data (the content of map files, not the content of the map_data attribute!) to contain area information etc (and switch from a raw mp to a wml based format which contains a map_data attribute among other attributes, this is realy easy to do since it in fact what the scenario generator already does), and i think so far most people agree with me. (also iirc this was already suggested many years ago but was hard to execute because the default way to include maps was via macro inclusion, now with map_file it's easier)

We can additionally support a real scenario editor, that supports 'as much features as possible', but i personally i it'd be nice to have but im not sure whether its the best was to spend our time.

Pentarctagon:
I guess it feels pretty misleading then for it to be called a scenario editor and for the in-game menu items to be "New Scenario"/"Save Scenario As...".

Anyway, is there any issue with using file extensions instead of file contents here?  If that's also problematic, then I'll probably go ahead and close the PR.

CelticMinstrel:
> I am not even sure sure if its playable at all if you don't add sides to it.

Editor scenario files do contain sides, so they're playable. Plus, even if they didn't contain sides, the game is able to simply add sides to them, right? It works for plain maps, so I'd expect it to also work for "extended maps".

> Anyway, is there any issue with using file extensions instead of file contents here?

I don't have a problem with this, at least. I suppose it might break if some rare person called their maps something other than a .map or .mask extension, but they can just rename them.

> In particular i support removing other data form in the default configuration, like information about [side]s

Sides are kinda an edge case, as they can also contain units, and in fact I'd actually expect units placed in the scenario editor to be part of the side rather than in a prestart event (which conveniently groups the units by their side).

There's other information that's not really map-related though, such as the music selection…

stevecotton:
> Anyway, is there any issue with using file extensions instead of file contents here? If that's also problematic, then I'll probably go ahead and close the PR.

I vote for closing without merging, although I'd leave a branch containing these changes for reference. Improving the situation is necessary for 1.16, but merging this would add an additional ".ed.cfg with a [multiplayer] tag" variation that the eventual 1.16 code has to handle.

Pentarctagon:
> I vote for closing without merging, although I'd leave a branch containing these changes for reference. Improving the situation is necessary for 1.16, but merging this would add an additional ".ed.cfg with a [multiplayer] tag" variation that the eventual 1.16 code has to handle.

I mean just the extension changes without the [multiplayer] tag being added.

stevecotton:
> I mean just the extension changes without the [multiplayer] tag being added.

I think it's more likely to end up with ".map" used for 1.16 rather than ".ed.cfg".

PR #5285, opened 2020-11-15
Make saved_game::expand_scenario() look correct, and add a const

stevecotton:
This doesn't change the behavior - the changed line is in a conditional block
which tells us exactly which of scenario_id()'s code paths will happen, and it
will still end up using carryover_["next_scenario"]. The change is just making
it consistent with the way that other lines in expand_scenario() find out which
scenario id to use; it's preparing to load the next scenario, so it should use
that scenario's id to load any metadata, not the current scenario's id.

Also, scenario_id() should be const. This helps with refactoring out the
existence of a de-facto singleton instance of saved_game.

### DIFF
--- a/src/saved_game.cpp
+++ b/src/saved_game.cpp
@@ -269,7 +269,7 @@ void saved_game::set_defaults()
 void saved_game::expand_scenario()
 {
 	if(this->starting_point_type_ == STARTING_POINT_NONE && !has_carryover_expanded_) {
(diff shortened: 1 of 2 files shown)


stevecotton:
There's only one line of logging from the failing build, seems unrelated:

    An error occurred while generating the build script.

Vultraz:
Looks logically sound to me.

gfgtdf:
looks good to me.

gfgtdf (approved):

PR #5286, opened 2020-11-15
animal zombie variations

doofus-01:
The recently added fauna units and some existing core units, such as the rat, need better undead variations.  Maybe everything doesn't get a custom zombie, but all should get something that sort of makes sense.

Let's start with moving the SotA rat zombie to core.

EDIT:  A to-do list wouldn't hurt.

These are goals for this PR
- [x] move zombie rat to core and add portrait

- [x] zombie falcon (anim frames and portraits)

- [x] fixing gryphon zombie movement anims

- [x] zombie Gorer/Tusklet

These will probably be in a future PR, but we''ll see how it goes.
- [ ] zombie Bear

- [ ] zombie Yeti

- [ ] zombie Croc

- [ ] zombie water/sea serpent

- [ ] zombie cuttlefish


doofus-01:
That last commit was not really anything to do with undead, but living version should probably come before zombie version.

doofus-01:
I've got to eat the elephant one bite at a time or I'll give up, so the secondary checklist will have to be in a future PR.  The CI errors don't seem to have anything to do with this, so I'll go ahead and merge.  (I'll be able to fix it in the next couple days if that's wrong.)

Issue #5132, opened 2020-09-05
Removing need to hash password before sending to wesnothd/campaignd from the client


Pentarctagon:
A problem that's come up with the current implementation of logging into the official MP server with the forum username and password is that the wesnoth client needs to know how to hash the password it's sending, since all communication between wesnothd/campaignd is unencrypted.  This is an issue, since the hashing algorithm used by the forum software tends to change between releases which then requires altering which hashing algorithm the forums software uses so that the hashed password sent by the wesnoth client can be compared to the hashed password stored in the forum database.

For example, if the wesnoth client only knows how to hash passwords with MD5, and a forum upgrade changes the default password hashing to use bcrypt, then obviously the MD5 hash and the bcrypt hash aren't going to match.  This has come up again recently with phpbb now changing from preferring bcrypt to preferring argon2.

Short version: the wesnoth client shouldn't need to know how to create a hash that matches the phpbb forum software.

---

Based on some previous discussions, it sounds like the best solution would be to OpenSSL(?) to encrypt communication done between wesnothd/campaignd and the client, which would remove the need to have the password hashed by the client.

loonycyborg:
Asio has built-in support for ssl: https://www.boost.org/doc/libs/1_74_0/doc/html/boost_asio/overview/ssl.html
Most likely it will be matter of just using that extra api. But also matter of certificates requires research. From what I gather there is no general way to properly make use of system certificate authorities with asio in cross-platform manner atm. Perhaps we can supply our own certs since protocol that will be carried by the ssl layer is specific to wesnoth too?

Pentarctagon:
If we provide our own certs, one issue would be that they would need to stay valid for the life of the current stable branch and the next stable branch.

Pentarctagon:
@loonycyborg given the issues this causes when upgrading the forums, is it something you'd be able to investigate sometime during the 1.15 development cycle?

loonycyborg:
Maybe. Still would like to hear more comments about this from other people. From @GregoryLundberg for example.

Pentarctagon:
I don't think he's still around.  Regardless though, I'll probably give it a try in the next couple months and see if maybe I can get it to work.

PR #5287, opened 2020-11-17
WIP: Encrypt client/server communication using OpenSSL with Boost Asio.

Pentarctagon:
Currently implemented are the MP client connection and wesnothd+campaignd since they both use server_base.  I don't know how to run campaignd locally, but for wesnothd I was able to connect, disconnect, and play a short game between two players using my local instance.  I also confirmed that SSL/TLS was definitely being used via `echo "" | openssl s_client -connect localhost:11111` (11111 is the port my local wesnothd uses), the only issue reported being that the certificate used is self-signed.

Things left to do:

- [x] Implement for the add-ons server client.
- [x] See if I can find out why `shutdown: stream truncated` is printed out whenever a [shutdown()](https://github.com/boostorg/beast/issues/995#issuecomment-359507329) is called on the socket now, or whether it can be [ignored](https://github.com/boostorg/beast/issues/824).
- [ ] How to handle LAN servers/clients? - Probably this will need a runtime check to not use SSL for LAN.  Also this would apply for unofficial servers as well?
- [x] Travis certs - see if it works to generate the necessary files during the travis runs rather than having them in the repo.
- [ ] Fix travis.
- [ ] Make sure OpenSSL is ready and able to be packaged as part of the macOS release.
- [ ] Testing.
- [ ] Certs - How do we get or create the actual certs that will be used?
- [ ] A couple other TODO comments I left myself.

This is also the first part of addressing #5132.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -362,4 +362,7 @@ jobs:
           export CFG=Release
           export CACHE_DIR=~/build-cache
           export PATH="/mnt/c/ProgramData/chocolatey/lib/SQLite/tools/sqlite-tools-win32-x86-3330000:$PATH"
+          sudo apt update
+          sudo apt install dos2unix
+          dos2unix .github/workflows/ci-scripts/windows.sh
           ./.github/workflows/ci-scripts/windows.sh
--- a/.github/workflows/ci-scripts/windows.sh
+++ b/.github/workflows/ci-scripts/windows.sh
@@ -3,16 +3,17 @@ echo "python3: $(which python3)"
 echo "sqlite3.exe: $(which sqlite3.exe)"
 echo "cmd.exe: $(which cmd.exe)"
 echo "MSBuild.exe: $(which MSBuild.exe)"
+echo "dos2unix: $(which dos2unix)"
 START_DIR="$PWD"
-ls
+dir
 cd ..
-ls
+dir
 git clone --depth=1 https://github.com/microsoft/vcpkg.git vcpkg
 cd vcpkg
 cmd.exe //C bootstrap-vcpkg.bat
 cmd.exe //C 'C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsall.bat' amd64 '&&' vcpkg integrate install
 alias make="make -j2"
-ls
+dir
 ./vcpkg.exe install sdl2:x64-windows sdl2-image:x64-windows sdl2-image[libjpeg-turbo]:x64-windows sdl2-mixer:x64-windows sdl2-ttf:x64-windows bzip2:x64-windows zlib:x64-windows pango:x64-windows cairo:x64-windows fontconfig:x64-windows libvorbis:x64-windows libogg:x64-windows boost-filesystem:x64-windows boost-iostreams:x64-windows boost-locale:x64-windows boost-random:x64-windows boost-regex:x64-windows boost-asio:x64-windows boost-program-options:x64-windows boost-system:x64-windows boost-thread:x64-windows boost-bimap:x64-windows boost-multi-array:x64-windows boost-ptr-container:x64-windows boost-logic:x64-windows boost-format:x64-windows
 rm -R downloads
 rm -R buildtrees


Vultraz:
Perhaps this is a dumb question, but why do we save the password, hash, and send it each time instead of keeping an OAuth token?

Pentarctagon:
I'd guess it's because it was easier to send a hashed password than implement OAuth support, assuming OAuth was even around to be supported back when forum login support was added (OAuth 1.0 was released in 2010).

Vultraz:
Right, but at the time we just saved plaintext passwords. 😬 Would a token be better now?

Pentarctagon:
I'm not sure; I've not really looked into how it's used.  Partially it'd depend on how much work it'd be to implement and partially there'd need to be a way to link the token to a forum user ID, given off the top of my head we still need to make sure usernames are unique, determine which users get mod rights, and be able to accurately populate the `player_info` table.

loonycyborg:
We can gen our own cert, distribute public key with wesnoth and require it to work when "connect to official server" option is used. And let any other cert with other servers, potentially letting people add own trusted certs.

loonycyborg:
So I assume verify mode should be ssl::verify_fail_if_no_peer_cert for connecting to official servers and load_verify_file() should be used to load pem file appropriate to requested server. Though I'm not sure how exactly we make own auth authorities yet.

loonycyborg:
https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/example/cpp03/ssl/server.cpp shows an example of server. Serverside verify_* functions are only useful to authenticate clients afaict.

loonycyborg:
I think we should create own CA, create own cert for each official server and sign them all with that CA. Then have client use our CA when authenticating official servers.

loonycyborg:
Alternatively perhaps we can leverage already existing letsencrypt setup.

Pentarctagon:
Is it possible to have letsencrypt certificates have a long enough expiration period?  For some reason I was thinking they were required to have a pretty short validity period.

loonycyborg:
letsencrypt certs are auto-renewed and can be validated by OS provided certificate store. It's only matter of making letsencrypt sign wesnothd's certs in addition to website certs. Though getting to OS certs can be hard on some configurations(android? wolfssl?).

Pentarctagon:
Does auto-renewing require any update to the certificate that clients would have, or would they be able to keep using the same one for as long as the certificate keeps getting renewed?

loonycyborg:
Only server cert will have to be updated. Clients merely verify signature and check whether certificate hostname matches the hostname it's trying to connect to. https://www.boost.org/doc/libs/1_74_0/doc/html/boost_asio/reference/ssl__host_name_verification.html would need to be used. This relies on server's cert being signed by a certificate authority trusted by the system, and letsencrypt enables that, just like with website's https.

Pentarctagon:
Alright, that's good then.

As far as getting the OS provided certs, what I've been able to find is:
* Linux - set_default_verify_paths() should work.
* Windows - set_default_verify_paths() might work, otherwise something like [this](https://stackoverflow.com/a/40710806) might be needed.
* macOS - I didn't find anything one way or the other.
* iOS/Android/alternatives to OpenSSL - I'm not sure.  For iOS and Android probably the worst case is they'd also need custom code to load the OS certificates, but for things like WolfSSL I really have no idea.

Pentarctagon:
Though it looks like that method you linked isn't available until Boost 1.73.

loonycyborg:
Seems it was given a clearer name: https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ssl__rfc2818_verification.html

Pentarctagon:
Aren't those different things, since the description says they are based on different RFCs?

loonycyborg:
Most likely that rfc number just wasn't specific enough. rfc2818 describes entire http over tls protocol.

loonycyborg:
According to asio changelog:

> Added the ssl::host_name_verification class, which is a drop-in replacement for ssl::rfc2818_verification. The ssl::rfc2818_verification class has been marked as deprecated. As a consequence of this change, SSL support now depends on functions that were introduced in OpenSSL 1.0.2.

Pentarctagon:
Huh, alright then. It would've been nice for the documentation to say that, I suppose.

soliton-:
Is there any provision so far to continue to support unencrypted communication?

Pentarctagon:
It's the 3rd item in the checklist in the first post.

soliton-:
Excellent. The check would ideally be in the handshake so we can continue to support redirecting clients based on version. Maybe the encryption would only begin once past that part.

Pentarctagon:
Unecrypted uses `boost::asio::ip::tcp::socket` whereas encrypted uses `boost::asio::ssl::stream<boost::asio::ip::tcp::socket>`, which aren't interchangeable, so most likely it'll need one of each and then need to choose which one based on whether encryption is enabled or not.

Pentarctagon:
@soliton- also, there is specificaly `socket->async_handshake(boost::asio::ssl::stream_base::server, ...)` that needs to be done for encrypted connections to work.  I don't know technically how redirecting is handled currently though.

PR #5288, opened 2020-11-18
campaignd: Fix broken Windows support

shikadiqueen:
This creates temporary files with the `FILE_SHARE_DELETE` sharing mode on Windows when using `filesystem::atomic_commit`,  thus allowing us to rename them to their final location without having to close them beforehand. Without these changes, campaignd gets really upset when trying to rename files created with `atomic_commit` since Windows by default will deny deletion/renaming of open files.

**UPDATE 2020-11-20:** Now it also makes it so the temporary files overwrite the destination in a POSIX-like fashion by using `SetFileInformationByHandle()`.

~~... Or at least so the theory goes.~~

~~`Wine (5.0) is being temperamental and campaignd doesn't work properly on it, with client requests not going completely through for some mysterious reason, and even occasionally crashing with a segmentation fault somewhere in Wine land. Maybe our network code does something that Wine *really* doesn't like. So for the time being, this code is almost completely untested. I may try using MSVC++ instead at some point when I feel like it, or someone else can do it for me.`~~

**Testing procedure:**

* Run `campaignd.exe` so it will generate its initial configuration (`server.cfg`) and data dir in the current working path (or somewhere else if using the `-s <PATH>` command line switch). Make sure it doesn't log any errors involving `server.cfg`.
* Restart `campaignd.exe`. Again make sure it doesn't print any errors involving `server.cfg`.
* Upload an add-on with the Wesnoth client.
* Reupload the same add-on, modified.
* Verify that no errors involving the add-on were printed and that the modified version is what clients get when installing it fresh.

**Caveats:**

~~This is not in a mergeable state right now due to lack of testing. Additionally, the `rename()` call in `atomic_commit::commit()` needs to be replaced with `_wrename()` for Unicode path safety.~~

~~A simpler alternative to calling `CreateFile()` ourselves would be to reset the underlying `ostream` pointer in `filesystem::atomic_commit::commit()` before renaming the temporary file to force it to be closed first. However, this would mean that callers would have to take care not to use any existing references to the `ostream` following a call to `commit()`.~~

### DIFF
--- a/src/server/campaignd/fs_commit.cpp
+++ b/src/server/campaignd/fs_commit.cpp
@@ -21,10 +21,24 @@
 #include <cstdio>
 #include <cstring>
 
-#ifndef _WIN32
-#include <unistd.h>
 #include <boost/iostreams/device/file_descriptor.hpp>
 #include <boost/iostreams/stream.hpp>
+
+#ifndef _WIN32
+
+#include <unistd.h>
+
+#else
+
+#include "formatter.hpp"
+#include "serialization/unicode_cast.hpp"
+
+#include <boost/system/error_code.hpp>
+#include <boost/filesystem.hpp>
+
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+
 #endif
 
 static lg::log_domain log_filesystem("filesystem");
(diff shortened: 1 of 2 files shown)


stevecotton:
Github supports a "convert to draft" (at the bottom of the reviewers list) for stuff that shouldn't be merged.

shikadiqueen:
Addressed all the initial issues after a long evening fighting the Windows API until I finally found out I was calling `filesystem::ostream_file()` in the fallback path instead of `filesystem::<unnamed>::ostream_file_with_delete()`.

PR #5290, opened 2020-11-19
Add a unit test for the behavior of [unit]id= recalling units

stevecotton:
Answers a question raised in PR #4999; the test also passes on 1.14.


CelticMinstrel:
Huh? That works? I actually thought that only `[unit]` in `[side]` had that behaviour.

Pentarctagon (approved):

CelticMinstrel:
Note: if this can be verified to work in `[side]` too, that would probably be even better.

stevecotton:
@CelticMinstrel see the commit that's auto-linked itself above. The problem is that you need the unit to be on the recall list before the start of the scenario - maybe a `preload` event could do it?

PR #5292, opened 2020-11-19
Wrap owning editor_action pointers in unique_ptr

AI0867:
It was [recently noted](https://github.com/wesnoth/wesnoth/pull/5254#discussion_r516381049) that the editor works by passing raw owning `editor_action*` around. This PR changes that.

### DIFF
--- a/src/editor/action/action.cpp
+++ b/src/editor/action/action.cpp
@@ -60,11 +60,11 @@ std::string editor_action::get_description() const
 	return "Unknown action";
 }
 
-editor_action* editor_action::perform(map_context& mc) const
+editor_action_ptr editor_action::perform(map_context& mc) const
 {
 	editor_action_ptr undo(new editor_action_whole_map(mc.map()));
 	perform_without_undo(mc);
-	return undo.release();
+	return undo;
 }
 
 IMPLEMENT_ACTION(whole_map)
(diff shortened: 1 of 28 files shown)


Pentarctagon:
This is similar to #5260 then?

src/editor/action/action.cpp
```diff
@@ -260,11 +253,11 @@ editor_action* editor_action_starting_position::perform(map_context& mc) const
 		// action_chain.
 		editor_action_chain* undo_chain = new editor_action_chain();
```
stevecotton, line 254: This line should use unique_ptr too, but both I (in #5260) and AI0867 missed it.  (I'm reviewing a diff between the two PRs.)

stevecotton:
I think #5260 is the better of the two PRs, there are some changes in it that aren't in this one:

* writing `std::unique_ptr` repeatedly instead of using the alias, I think this had consensus support on `#wesnoth-dev`
* more use of `auto`
* using `override` and removing corresponding `/** Inherited from editor_action, implemented by IMPLEMENT_ACTION. */` comments
* removing commented-out code from mouse_action_item.cpp which seems to be leftovers from the creation of that file

AI0867:
Oops, I hadn't noticed that other PR.

I messed around with various ways to do the covariant return types, before I decided they weren't actually necessary, which is why the override keyword is missing.

stevecotton:
Any objection to me merging #5260?

AI0867:
I have none.

Pentarctagon:
Closing then as #5260 has been merged.

PR #5299, opened 2020-11-22
Encrypt client/server communication using OpenSSL with Boost Asio.

Pentarctagon:
Re-PR of #5287.

Currently implemented are the MP client connection and wesnothd+campaignd since they both use server_base.  I don't know how to run campaignd locally, but for wesnothd I was able to connect, disconnect, and play a short game between two players using my local instance.  I also confirmed that SSL/TLS was definitely being used via `echo "" | openssl s_client -connect localhost:11111` (11111 is the port my local wesnothd uses), the only issue reported being that the certificate used is self-signed.

Things left to do:

- [x] Implement for the add-ons server client.
- [x] See if I can find out why `shutdown: stream truncated` is printed out whenever a [shutdown()](https://github.com/boostorg/beast/issues/995#issuecomment-359507329) is called on the socket now, or whether it can be [ignored](https://github.com/boostorg/beast/issues/824).
- [ ] How to support either SSL or non-SSL?
- [x] Travis certs - see if it works to generate the necessary files during the travis runs rather than having them in the repo.
- [x] Fix the CI.
- [x] Make sure OpenSSL is ready and able to be packaged as part of the macOS release.
- [ ] Testing. In particular, it will need to be tested if macOS and Windows clients will need any OS-specific code to be able to load the system-provided certs.
- [x] Certs - How do we get or create the actual certs that will be used?  [See this comment.](https://github.com/wesnoth/wesnoth/pull/5299#issuecomment-734408639)
- [ ] A couple other TODO comments I left myself.

This is also the first part of addressing #5132.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -209,6 +209,12 @@ fonts/.uuid
 data/**/*.xcf
 data/**/*.psd
 
+# SSL certificates
+certs/*.pem
+certs/*.crt
+certs/*.csr
+certs/*.key
+
 # uncategorized
 revision*
 config.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,7 @@ if(NOT APPLE)
 else()
 	set(CRYPTO_LIBRARY "-framework Security")
 endif()
+find_package(OpenSSL REQUIRED)
 
 find_package(Boost ${BOOST_VERSION} REQUIRED COMPONENTS iostreams program_options regex system thread random)
 
--- a/SConstruct
+++ b/SConstruct
@@ -391,6 +391,7 @@ if env["prereqs"]:
     if(env["PLATFORM"] != 'darwin'):
         # Otherwise, use Security.framework
         have_server_prereqs = have_server_prereqs & conf.CheckLib("libcrypto")
+        have_server_prereqs = have_server_prereqs & conf.CheckLib("libssl")
 
     env = conf.Finish()
 
--- a/certs/README.md
+++ b/certs/README.md
@@ -1,0 +1,4 @@
+About
+=====
+
+This directory is for the certificates/keys needed for encrypting communication between the wesnoth client and wesnothd/campaignd.
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -75,6 +75,7 @@ endif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 set(game-external-libs
 	${common-external-libs}
 	${CRYPTO_LIBRARY}
+	${OPENSSL_SSL_LIBRARY}
 	${Boost_SYSTEM_LIBRARIES}
 	${Boost_RANDOM_LIBRARY}
 	${Boost_THREAD_LIBRARY}
@@ -101,6 +102,7 @@ set(server-external-libs
 	${Boost_SYSTEM_LIBRARIES}
 	${Boost_RANDOM_LIBRARY}
 	${CRYPTO_LIBRARY}
+	${OPENSSL_SSL_LIBRARY}
 	${MYSQL_LIBS}
 	-lpthread
 )
--- a/src/network_asio.cpp
+++ b/src/network_asio.cpp
@@ -57,8 +57,9 @@ using boost::system::system_error;
 
 connection::connection(const std::string& host, const std::string& service)
 	: io_service_()
+	, ssl_ctx_(boost::asio::ssl::context::sslv23)
 	, resolver_(io_service_)
-	, socket_(io_service_)
+	, socket_(io_service_, ssl_ctx_)
 	, done_(false)
 	, write_buf_()
 	, read_buf_()
@@ -69,8 +70,12 @@ connection::connection(const std::string& host, const std::string& service)
 	, bytes_to_read_(0)
 	, bytes_read_(0)
 {
-	resolver_.async_resolve(
-		boost::asio::ip::tcp::resolver::query(host, service), std::bind(&connection::handle_resolve, this, _1, _2));
+	ssl_ctx_.set_default_verify_paths();
+	ssl_ctx_.load_verify_file("certs/"+host+".crt");
+	socket_.set_verify_mode(boost::asio::ssl::verify_fail_if_no_peer_cert);
+	socket_.set_verify_callback(boost::asio::ssl::rfc2818_verification(host));
+
+	resolver_.async_resolve(resolver::query(host, service), std::bind(&connection::handle_resolve, this, _1, _2));
 
 	LOG_NW << "Resolving hostname: " << host << '\n';
 }
@@ -86,7 +91,7 @@ void connection::handle_resolve(const boost::system::error_code& ec, resolver::i
 
 void connection::connect(resolver::iterator iterator)
 {
-	socket_.async_connect(*iterator, std::bind(&connection::handle_connect, this, _1, iterator));
+	boost::asio::async_connect(socket_.lowest_layer(), iterator, std::bind(&connection::handle_connect, this, _1, iterator));
 	LOG_NW << "Connecting to " << iterator->endpoint().address() << '\n';
 }
 
@@ -94,10 +99,14 @@ void connection::handle_connect(const boost::system::error_code& ec, resolver::i
 {
 	if(ec) {
 		WRN_NW << "Failed to connect to " << iterator->endpoint().address() << ": " << ec.message() << '\n';
-		socket_.close();
+		// TODO: need to test this part somehow
+		//       not sure if this works, or if I need to use shutdown() and create an entirely new socket
+		//       though the intention is to try multiple IPs through the same ssl context, so this sounds correct at least
+		socket_.lowest_layer().close();
 
 		if(++iterator == resolver::iterator()) {
 			ERR_NW << "Tried all IPs. Giving up" << std::endl;
+			socket_.shutdown();
 			throw system_error(ec);
 		} else {
 			connect(iterator);
@@ -112,11 +121,19 @@ void connection::handshake()
 {
 	static const uint32_t handshake = 0;
 
-	boost::asio::async_write(socket_, boost::asio::buffer(reinterpret_cast<const char*>(&handshake), 4),
-		std::bind(&connection::handle_write, this, _1, _2));
+	socket_.async_handshake(boost::asio::ssl::stream_base::client,
+		[this](const boost::system::error_code& hs_ec)
+		{
+			if(hs_ec) {
+				throw system_error(hs_ec);
+			} else {
+				boost::asio::async_write(socket_, boost::asio::buffer(reinterpret_cast<const char*>(&handshake), 4),
+					std::bind(&connection::handle_write, this, _1, _2));
 
-	boost::asio::async_read(socket_, boost::asio::buffer(&handshake_response_.binary, 4),
-		std::bind(&connection::handle_handshake, this, _1));
+				boost::asio::async_read(socket_, boost::asio::buffer(&handshake_response_.binary, 4),
+					std::bind(&connection::handle_handshake, this, _1));
+			}
+		});
 }
 
 void connection::handle_handshake(const boost::system::error_code& ec)
@@ -156,7 +173,7 @@ void connection::transfer(const config& request, config& response)
 
 void connection::cancel()
 {
-	if(socket_.is_open()) {
+	if(socket_.lowest_layer().is_open()) {
 		boost::system::error_code ec;
 
 #ifdef _MSC_VER
@@ -165,7 +182,7 @@ void connection::cancel()
 #pragma warning(push)
 #pragma warning(disable:4996)
 #endif
-		socket_.cancel(ec);
+		socket_.lowest_layer().cancel(ec);
 #ifdef _MSC_VER
 #pragma warning(pop)
 #endif
--- a/src/network_asio.hpp
+++ b/src/network_asio.hpp
@@ -33,6 +33,7 @@
 #include "exceptions.hpp"
 
 #include <boost/asio.hpp>
+#include <boost/asio/ssl.hpp>
 
 class config;
 
@@ -120,10 +121,12 @@ public:
 private:
 	boost::asio::io_service io_service_;
 
+	boost::asio::ssl::context ssl_ctx_;
+
 	typedef boost::asio::ip::tcp::resolver resolver;
 	resolver resolver_;
 
-	typedef boost::asio::ip::tcp::socket socket;
+	typedef boost::asio::ssl::stream<boost::asio::ip::tcp::socket> socket;
 	socket socket_;
 
 	bool done_;
--- a/src/server/campaignd/server.cpp
+++ b/src/server/campaignd/server.cpp
@@ -305,6 +305,15 @@ void server::load_config()
 	filesystem::scoped_istream in = filesystem::istream_file(cfg_file_);
 	read(cfg_, *in);
 
+	// ssl setup
+	ssl_ = cfg_["ssl"].to_bool();
+	if(ssl_) {
+		setup_ssl(cfg_["crt"], cfg_["private_key"], cfg_["dhparam"]);
+		LOG_CS << "SSL enabled\n";
+	} else {
+		LOG_CS << "SSL not enabled\n";
+	}
+
 	read_only_ = cfg_["read_only"].to_bool(false);
 
 	if(read_only_) {
--- a/src/server/common/send_receive_wml_helpers.ipp
+++ b/src/server/common/send_receive_wml_helpers.ipp
@@ -107,8 +107,8 @@ struct sendfile_op
 	{
 		// Put the underlying socket into non-blocking mode.
 		if (!ec)
-			if (!sock_->native_non_blocking())
-				sock_->native_non_blocking(true, ec);
+			if (!sock_->lowest_layer().native_non_blocking())
+				sock_->lowest_layer().native_non_blocking(true, ec);
 
 		if (!ec)
 		{
@@ -116,7 +116,7 @@ struct sendfile_op
 			{
 				// Try the system call.
 				errno = 0;
-				int n = ::sendfile(sock_->native_handle(), fd_, &offset_, 65536);
+				int n = ::sendfile(sock_->lowest_layer().native_handle(), fd_, &offset_, 65536);
 				ec = boost::system::error_code(n < 0 ? errno : 0,
 											   boost::asio::error::get_system_category());
 				total_bytes_transferred_ += ec ? 0 : n;
@@ -187,7 +187,7 @@ struct sendfile_op
 		bool failed = false;
 		if (!pending_)
 		{
-			BOOL success = TransmitFile(sock_->native_handle(), file_, 0, 0, &overlap_, nullptr, 0);
+			BOOL success = TransmitFile(sock_->lowest_layer().native_handle(), file_, 0, 0, &overlap_, nullptr, 0);
 			if (!success)
 			{
 				int winsock_ec = WSAGetLastError();
--- a/src/server/common/server_base.cpp
+++ b/src/server/common/server_base.cpp
@@ -32,10 +32,13 @@ static lg::log_domain log_config("config");
 
 #include "server/common/send_receive_wml_helpers.ipp"
 
+// TODO: test what happens when server gets unencrypted but expects encrypted and vice versa
+// TODO: move setup out of constructor so that options can be configured in the wesnothd/campaignd config files
 server_base::server_base(unsigned short port, bool keep_alive) :
 	port_(port),
 	keep_alive_(keep_alive),
 	io_service_(),
+	ssl_ctx_(boost::asio::ssl::context::sslv23),
 	acceptor_v6_(io_service_),
 	acceptor_v4_(io_service_),
 	#ifndef _WIN32
@@ -44,6 +47,17 @@ server_base::server_base(unsigned short port, bool keep_alive) :
 	#endif
 	sigs_(io_service_, SIGINT, SIGTERM)
 {
+}
+
+void server_base::setup_ssl(const std::string& crt, const std::string& private_key, const std::string& dhparam)
+{
+	ssl_ctx_.set_options(boost::asio::ssl::context::default_workarounds
+        | boost::asio::ssl::context::no_sslv2
+        | boost::asio::ssl::context::no_sslv3
+        | boost::asio::ssl::context::single_dh_use);
+	ssl_ctx_.use_certificate_chain_file(crt);
+	ssl_ctx_.use_private_key_file(private_key, boost::asio::ssl::context::pem);
+	ssl_ctx_.use_tmp_dh_file(dhparam);
 }
 
 void server_base::setup_acceptor(boost::asio::ip::tcp::acceptor& acceptor, boost::asio::ip::tcp::endpoint endpoint)
@@ -79,8 +93,8 @@ void server_base::start_server()
 
 void server_base::serve(boost::asio::ip::tcp::acceptor& acceptor)
 {
-	socket_ptr socket = std::make_shared<boost::asio::ip::tcp::socket>(io_service_);
-	acceptor.async_accept(*socket, [&acceptor, socket, this](const boost::system::error_code& error){
+	socket_ptr socket = std::make_shared<boost::asio::ssl::stream<boost::asio::ip::tcp::socket>>(io_service_, ssl_ctx_);
+	acceptor.async_accept(socket->lowest_layer(), [&acceptor, socket, this](const boost::system::error_code& error){
 		this->accept_connection(acceptor, error, socket);
 	});
 }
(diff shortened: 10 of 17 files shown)


loonycyborg:
I'll look into using our current letsencrypt setup to possibly provide and autorenew certs for wesnothd and campaignd. Advantage of this is that clients never require any certificates beside those provided by OS to verify connection and we won't have to deal with certificate expiry as long as auto-renewal on server works.

Pentarctagon:
Great, thanks.

Pentarctagon:
So I think the main thing left to implement is supporting either SSL and non-SSL connections.  The main problem I'm running into for being able to support either one or the other with a run-time check is that the regular TCP socket is `boost::asio::ip::tcp::socket` while the SSL layer is`boost::asio::ssl::stream<boost::asio::ip::tcp::socket>`, and there's no common parent class or (as far as I know) any other way that would allow passing either into the same function.

For the MP/add-ons server clients I don't think this is much of an issue since their use of the socket is limited enough that a few if-statements should be workable.

For wesnothd and campaignd though, that's not workable since there are significantly more uses of the socket, the socket is passed around to various functions, and is used as a key in the `player_connections` boost::multi_index.  What is much more limited though are instances where TCP-specific functions are called (in the current PR, anywhere that was changed to call `lowest_layer()`), so I'm thinking that the way to go for wesnothd and campaignd would be to make SSL vs non-SSL be determined by a compile time define.  So for example:
```
#ifdef USE_SSL
typedef std::shared_ptr<boost::asio::ssl::stream<boost::asio::ip::tcp::socket>> socket_ptr;
#else
typedef std::shared_ptr<boost::asio::ip::tcp::socket> socket_ptr;
#endif
```
and as an example of a place that would need the `lowest_layer()` difference between SSL and non-SSL:
```
#ifdef USE_SSL
setsockopt(socket->lowest_layer().native_handle(), IPPROTO_TCP, TCP_KEEPALIVE, &timeout, sizeof(timeout));
#else
setsockopt(socket->native_handle(), IPPROTO_TCP, TCP_KEEPALIVE, &timeout, sizeof(timeout));
#endif
```

Pentarctagon:
(pinging @soliton- and @Vultraz as well since they were on the previous PR)

loonycyborg:
Maybe always use ssl but don't do verification unless it makes sense like in case of official servers?

loonycyborg:
I think by default it should do host name validation if server specified by domain name and do no validation on ips. There should be facility to add known servers and ability to set whether validation is required or validation cert for each separately. 

Also this might lead to alternative for password authentication. Use certs to authenticate clients instead. Though server would need to know clients in advance. Probably it will require an option on the forum letting users to upload their certs, then wesnothd can read it from db.

Pentarctagon:
The two use cases I am aware of for not having SSL enabled are:
* The LAN server that's easily available from the main menu - if SSL is enabled at all, we'd need to ship a default private key and certificate for the server/client to use in order to satisfy OpenSSL, even if they won't provide any actual security (which I suppose isn't needed in this case either).
* [Alternate MP servers](https://wiki.wesnoth.org/MultiplayerServers#Alternate_third-party_servers) - Though they could of course use SSL and create certificates (and definitely should if their players are using a password to login).

I don't know if @soliton- or @shikadiqueen have others in mind.

soliton-:
I would like to keep the initial part of the protocol the same so that we can keep running a redirector server for all versions.

In principal this should be possible. Do everything as before until you get to the login procedure and do the ssl stuff then (or not) if both sides support it.

No clue how easy that is to accomplish with asio though.

Pentarctagon:
> I would like to keep the initial part of the protocol the same so that we can keep running a redirector server for all versions.
> 
> In principle this should be possible. Do everything as before until you get to the login procedure and do the ssl stuff then (or not) if both sides support it.
> 
> No clue how easy that is to accomplish with asio though.

At a minimum it would require duplicating some functions in server_base and send_receive_wml_helpers so they'd be able to work with either an SSL or non-SSL connection, as well as there would need to be a second handshake done and socket created for SSL once it got to the login part.  Also some additional logic would be needed on the client side for them to know at which point to start using SSL (if enabled for a particular server) since wesnothd_connection has no knowledge currently of whether login has occurred.

loonycyborg:
With this PR I managed to make host based authentication work by adding the following to wesnothd's config file:
`
ssl=true
crt=/path/to/fullchain.pem
private_key=/path/to/privkey.pem
dhparam=certs/dhparam.pem
`
fullchain.pem and privkey.pem come from acme.sh:
`
acme.sh --install-cert -d <server-domain-name> --key-file /path/to/privkey.pem --fullchain-file /path/to/fullchain.pem
`
This can be adapted to any other acme client as long as you figure out what their equivalents --key-file and --fullchain-file.
This works with this PR with system certs if connecting to server by name and domain name matches to what acme got cert for.
For this way load_verify_file() call shouldn't be used on client.

loonycyborg:
Conditional compilation for ssl might be ok on server. But in any case clients will have to support both.

Vultraz:
Since I really have no idea how this stuff works, does this bring us closer to not having to save the password and using tokens instead?

Pentarctagon:
@Vultraz No.

@AI0867 The client will support using either, so everyone needing to upgrade is not an issue since they'd need 1.16 (eventually) anyway.  The question is about the server implementation.

Pentarctagon:
@soliton- Also, I have no idea how the redirector works.  How does it currently determine which wesnothd instance to redirect a player to?

loonycyborg:
Yes, it's usual wesnothd, only it redirects clients to other versions based on version string they send at early stage during handshake.

Pentarctagon:
Alright, so it would be https://github.com/wesnoth/wesnoth/blob/753590beb3f240f29a6fd1a528e4ebb3fed3289c/src/server/wesnothd/server.cpp#L641-L655 rather than something entirely independent then?  If so, I don't think SSL vs non-SSL would have any affect, would it?

loonycyborg:
if it supports ssl only it wouldn't be able to handle old clients. So it would be unencrypted.

Pentarctagon:
Alright, so it wouldn't be able to redirect between 1.14 and 1.16, for example.  Is it setup where all connections go to one wesnothd instance and then get redirected to another instance if needed, or something like that?  I wonder if it would be possible to detect if it's encrypted, if so immediately redirect to 1.14, and then let the 1.14 instance handle any additional redirects that might be needed (ie: 1.15 -> 1.14 -> 1.12).

I'm not bothered at all by the first part prior to the login not being encrypted. I'd just like to avoid needing to refactor how the sockets are handled if I don't need to.

loonycyborg:
I think original redirecting instance can stay unencrypted. But other instances will use whatever is appropriate to them assuming client can adapt.

Pentarctagon:
Out of curiosity, why are clients redirected instead of giving each wesnothd instance its own port?

Vultraz:
Wait, they do have their own port. I can connect to a specific wesnothd instance by specifying the port...

Pentarctagon:
> Wait, they do have their own port. I can connect to a specific wesnothd instance by specifying the port...

If you go to the Connect to Server dialog the official servers are listed as being on port 15000, from `data/game_config.cfg`, on at least 1.12, 1.14, and master.

Vultraz:
You can still connect to the 1.14 instance via 1.15 by connecting to `server.wesnoth.org:14999`, for example.

Pentarctagon:
Right.  I'm wondering why 1.14's game_config.cfg isn't set to 14999 directly though.  Unless there's a reason that the 1.14 port would need to be changed to something other than 14999 but the 15000 port would be left as-is, it seems strange to me that a redirector would be needed instead of going directly to the correct port.

shikadiqueen:
I'm fairly sure (Soliton should be able to confirm since he's been working with the MP servers for ages) that the whole point of the redirector is that: 1) not all servers need to use the exact same ports for each version; 2) we should be able to reassign ports as needed (which was probably more important back in the day when wesnoth.org relied on donated hosting).

(It also makes things incredibly easier at release time compared to the campaignd business. At least these days we've settled on a port numbering scheme for that instead of rotating ports in the 15001 - 15009 range like before.)

Pentarctagon:
Alright.  So if the redirector has to stay and support non-SSL requests, probably what will need to happen instead of the `#ifdef`s is every current direct use of `socket_ptr` in all of the server classes will need to be moved to an encapulating class (named something like network_connection) which will then be responsible for deciding whether it needs to be dealing with a regular TCP object or an SSL stream and based on that which version of the boost functions to call.

---

In the case of a redirect being done, everything would stay unencrypted and would be handled as today aside from needing the network_connection class instead of directly using the socket_ptr.

---

In the case of no redirect being done, instead of sending a redirect response, the server would need to send something to the client telling it to start a new, SSL-enabled connection (if the server is configured to use SSL).  It seems like a fairly terrible idea to send an unencrypted message from the server to tell the client to start using encryption, but I'm not sure what the alternative is since the client has no way of knowing whether it's going to get redirected to a different wesnothd instance or continue to logging in.  I also so far haven't been able to find a way to pre-emptively check if an incoming connection is using SSL or not, which would be needed in order to handle the initial communication that happens through the redirect stage vs a post-redirect SSL-enabled connection.

The client would also need to use a non-SSL connection through the part where a redirect might happen and then start a new SSL connection after it gets the redirect.

One potential option that would make this a lot less horrible if it works is [this SO answer](https://stackoverflow.com/a/11699839) which says that you're able to use the plain TCP socket to read and write data up until you do the SSL handshake.  If true, then it sounds like the redirecting could potentially be done in `server_base::accept_connection`.

---

Separate from anything else, it'd also be good to exit at startup if wesnothd is configured to require an account to login but doesn't have SSL configured to be used, to avoid any potential for passwords to accidentally be sent over the network unencrypted.

Pentarctagon:
Alternatively, would it make any sense to turn the redirector into its own tiny executable where all it does is check the version sent and reply with the redirect information?  Then the redirector could always be built with SSL disabled, wesnothd and campaignd could be built either way, and then I *think* the main issue remaining would be how to have the client know whether it needs to use SSL or not for a particular server and port.

loonycyborg:
Most likely changing the way redirector works would be disruptive for our current setup. Starting ssl handshake based on what's in version string after receiving it might be better idea. Revamping the way infrastructure with multiple versions of wesnothd/campaignd works should be separate project.

AI0867:
> Most likely changing the way redirector works would be disruptive for our current setup. Starting ssl handshake based on what's in version string after receiving it might be better idea. Revamping the way infrastructure with multiple versions of wesnothd/campaignd works should be separate project.

The redirection works by returning a WML tag containing the hostname and port to redirect to if the client does not have a version the server accepts. Whatever is listening at port 15000 could be changed to a tiny program doing just that without disrupting much, just the fact that master would no longer connect to that port without redirection.

In fact, I wrote a bit of python some years ago that could do the job, though it could do with some rewriting: https://github.com/AI0867/wesnet-tools/blob/master/server_redirector.py

soliton-:
As Iris said the redirector setup is convenient and flexible. It is not the end-all-be-all but if possible I'd like to keep it.

@Pentarctagon Everything you said in https://github.com/wesnoth/wesnoth/pull/5299#issuecomment-734652946 sounds good, except I don't think it's an issue to negotiate whether to use encryption in an unencrypted fashion (infact how else would you do that?) and that SO post sounds promising indeed.

The redirector could be replaced by a smaller program that just does redirecting not sure that really helps much other than resource usage perhaps though. And it would mean maintaining that smaller program as well. (It should be able to read wesnothd style config files with preprocessor directives for example.)

Pentarctagon:
> Most likely changing the way redirector works would be disruptive for our current setup. Starting ssl handshake based on what's in version string after receiving it might be better idea. Revamping the way infrastructure with multiple versions of wesnothd/campaignd works should be separate project.

I don't mean changing how it works at all.  I just mean that the dozen or so lines of wesnothd's code that currently handle redirection would be broken out and run as-is, just they'd be running as a standalone executable rather than within wesnothd.

> As Iris said the redirector setup is convenient and flexible. It is not the end-all-be-all but if possible I'd like to keep it.
> 
> @Pentarctagon Everything you said in #5299 (comment) sounds good, except I don't think it's an issue to negotiate whether to use encryption in an unencrypted fashion (infact how else would you do that?) and that SO post sounds promising indeed.
> 
> The redirector could be replaced by a smaller program that just does redirecting not sure that really helps much other than resource usage perhaps though. And it would mean maintaining that smaller program as well. (It should be able to read wesnothd style config files with preprocessor directives for example.)

I don't really have much of an opinion about the redirector in and of itself, honestly.  It's that needing to refactor all socket_ptr usage to be behind a class acting as an abstraction layer over using TCP vs SSL is a lot more work, would need a lot more testing to make sure pretty much everything else still works, and would create a much larger risk of something *not* working if the only reason to need to do that is the redirector, given that the redirector part is a very tiny part of wesnothd.  Whether that means removing the redirector entirely, moving the C++ implementation to an independent executable, using a python implementation of it, or even something else entirely is more or less the same to me as long as it avoids needing to be able to support accepting both SSL and non-SSL connections.

loonycyborg:
wrt level of indirection wesnothd already has player_record class, socket_ptr is merely used as an index to find appropriate player record. It can be extended to support both ssl and unencrypted clients at the same time. I guess record can have hold both low level socket and socket for ssl and provide helper to provide appropriate socket based on what caller wants and whether socket is ssl/non-ssl.

loonycyborg:
btw can this be made a PR withing wesnoth/wesnoth? I might want to add some commits here too.

Pentarctagon:
I'm thinking more of things like the send_receive_wml_helpers for example - its functions wouldn't be able to take an argument of `socket_ptr` or directly use `socket_ptr`s anymore since it would need to be able to be used both with TCP and SSL connections and those types aren't interchangeable.  The redirector uses the same send/receive functions from there as everything else, and this also happens before a player logs in and gets added to a `player_record`.

Pentarctagon:
> btw can this be made a PR within wesnoth/wesnoth? I might want to add some commits here too.

Sure, though I want to get the SSL/non-SSL approach for the server decided before anything else is done.

loonycyborg:
> I'm thinking more of things like the send_receive_wml_helpers for example - its functions wouldn't be able to take an argument of `socket_ptr` or directly use `socket_ptr`s anymore since it would need to be able to be used both with TCP and SSL connections and those types aren't interchangeable. The redirector uses the same send/receive functions from there as everything else, and this also happens before a player logs in and gets added to a `player_record`.

Functions in send_receive_wml_helpers can be made templated on socket type too, then they will work on any socket type that supports async_write() etc.

Pentarctagon:
That wouldn't work for cases that need to call `lowest_layer()` in order for an SSL connection to do TCP-specific things though.

loonycyborg:
Actually most likely there is a way to check the socket type with template metaprogramming and do either directly do things or call lowest_layer, depending on type.

Pentarctagon:
It probably is, but it still seems to me to be a "fitting a square peg in a round hole" sort of situation.  With the right tools you can make it work, but it won't be a good fit.

loonycyborg:
To me it seems like a better alternative to conditional compilation at least. And it's definitely asio way of doing things. It prefers template metaprogramming to subclassing, that's why there is no common base class for ssl and plain socket. But both types in fact implement same "concept", there's just no language support for expressing this directly yet, at least in c++14. That's what concepts c++20 feature is about.

Pentarctagon:
In some situations sure, but in this case it doesn't seem like a worthwhile tradeoff to make the code and the initial client<->server interaction more complex in order to avoid some simple `#ifdef`s.

loonycyborg:
Also, like soliton noted it's not simple to just tear out redirect part of the code because it also relies on config file parsing complete with use of preprocessor for includes. You could still use conditional compilation approach but delay ssl handshake until after redirect part is passed. Before that don't refer to socket type as socket_ptr but do another typedef for lowest layer socket.

Pentarctagon:
> Also, like soliton noted it's not simple to just tear out redirect part of the code because it also relies on config file parsing complete with use of preprocessor for includes.

It does, but there is already existing code for that - I wouldn't need to do anything new to be able to read the config file other than maybe move wesnothd's `server::read_config` into server_base.

> You could still use conditional compilation approach but delay ssl handshake until after redirect part is passed. Before that don't refer to socket type as socket_ptr but do another typedef for lowest layer socket.

That's more or less what the SO answer I linked above was suggesting, I think.  I'm not sure if that would actually work though, since it sounds strange to initialize the socket as `ssl::stream<tcp>` with an SSL context, get and use the plain TCP socket to communicate, and then only after that do the SSL handshake and continue using it as an encrypted stream.

Whether it works or not is a moot question though if the redirector can be made into a standalone executable though.

loonycyborg:
> That's more or less what the SO answer I linked above was suggesting, I think. I'm not sure if that would actually work though, since it sounds strange to initialize the socket as `ssl::stream<tcp>` with an SSL context, get and use the plain TCP socket to communicate, and then only after that do the SSL handshake and continue using it as an encrypted stream.

I guess you can also get plain tcp socket from plain tcp acceptor, and later when you decide to use ssl construct ssl::stream
 from said plain tcp socket. ssl::stream has perfect forwarding in its constructor, so if you pass existing tcp socket to it will count as copy constructor, right?

Pentarctagon:
I'm not sure.  If it ends up that the redirector needs to stay as a part of wesnothd, then it'd be the first thing I'd test to see if it works or not.

Pentarctagon:
@loonycyborg I've added the branch at https://github.com/wesnoth/wesnoth/tree/openssl

@soliton- Given the sticking point is the redirector which is part of wesnothd, and you're the wesnothd maintainer, the decision of how to handle it is ultimately up to you.

Another option as well would be to let the redirector use SSL, which would mean no changes to it, but it wouldn't support redirecting between 1.14 and the eventual 1.16.  The issue would then resolve itself when 1.18 is released and 1.14 turned off, since at that point all of the active wesnothd instances would be using SSL.

loonycyborg:
On second thought redirection isn't a problem at all with current way our infra works. The wesnothd instance that handles redirects isn't used for anything else so it can remain unencrypted. However it can tell clients to use ssl to connect to server it redirects them too. This assumes clients can have both ssl and non-ssl, which would be needed for unencrypted LAN servers and the like.

Pentarctagon:
Yeah, having a standalone redirector isn't relevant if that's the case, since there's already a wesnothd instance doing the exact same thing.

Pentarctagon:
Also, as far as the wesnothd client goes, these are the cases I've thought of that it would need handle and how it would determine whether to use SSL or not: https://github.com/wesnoth/wesnoth/blob/ef45567c924af217c334ca4f7da62aa015f41d92/src/wesnothd_connection.cpp#L76-L86

Let me know if there are any cases that are missing or incorrect.

loonycyborg:
Might be a good idea to always check that if client gets redirect from official server it's redirected to a domain in game_config.cfg . Otherwise it could be used for MitM attack: fake redirector directs client to fake ssl server on attacker's domain where they have certbot installed.

AI0867:
> Might be a good idea to always check that if client gets redirect from official server it's redirected to a domain in game_config.cfg . Otherwise it could be used for MitM attack: fake redirector directs client to fake ssl server on attacker's domain where they have certbot installed.

It might be a good idea to implement something analogous to HSTS.

loonycyborg:
This can be done only once server can support both ssl and non-ssl. For now extra domain check is better than nothing.

Pentarctagon:
> Might be a good idea to always check that if client gets redirect from official server it's redirected to a domain in game_config.cfg . Otherwise it could be used for MitM attack: fake redirector directs client to fake ssl server on attacker's domain where they have certbot installed.

This would prevent unofficial servers from using a redirector though, and along with that prevent them from being able to tell the client to use SSL.

Also, if the redirector getting Man in the Middle attacked is something that's a concern, then IMO it'd be best to do what I suggested above:

> Another option as well would be to let the redirector use SSL, which would mean no changes to it, but it wouldn't support redirecting between 1.14 and the eventual 1.16. The issue would then resolve itself when 1.18 is released and 1.14 turned off, since at that point all of the active wesnothd instances would be using SSL.

loonycyborg:
> This would prevent unofficial servers from using a redirector though, and along with that prevent them from being able to tell the client to use SSL.

It will not because my idea was to enforce this only if original address is that of official server. Basically if original server is known as official then all redirects should be to domain known as official server domains.

Pentarctagon:
So it would be that a server listed in game_config.cfg wouldn't be allowed to redirect to a server not in game_config.cfg?  Or that one server wouldn't be allowed to redirect any different server, only a different port on the same server?

loonycyborg:
Option one. Option two would work for unofficial servers too though, might be better. Wouldn't let us to redirect to different official domains though but we're not doing it anyway.

soliton-:
So to recap:
The client needs runtime encryption support to connect to ssl enabled and disabled servers. Encryption or not can probably be decided before connecting.
The server needs runtime encryption support as a config option. With this encryption or not is decided at start-up time. (No real point in supporting a change at config reload time.)

Thus all involved network code needs to be abstracted to support encryption or not before connecting.

Additionally to allow the redirection setup on server.w.o to support ssl-capable and older versions it'd be nice to leave the handshake up to the redirection decision unconditionally unencrypted. For this the network code (client and server side) would need to be able to delay the encryption handling up to that point.

The alternative would be to use a new network port for 1.16+. Perhaps abandoning the redirection approach (not removing the feature, just not using it in the normal operation of server.w.o) and keep using a new port for each new wesnoth version like we do with campaignd.

So regarding the technical details, currently we're using `boost::asio::ip::tcp::socket` which for encryption would change to `boost::asio::ssl::stream<boost::asio::ip::tcp::socket>`. The underlying `tcp::socket` would still be accessible and could be used as before, right? Then `async_handshake` is called to setup ssl? And that could as well be called at any later time?

Anything important I missed? Anything I got wrong?

Pentarctagon:
> The client needs runtime encryption support to connect to ssl enabled and disabled servers. Encryption or not can probably be decided before connecting.

Correct.

> The server needs runtime encryption support as a config option. With this encryption or not is decided at start-up time. (No real point in supporting a change at config reload time.)
> 
> Additionally to allow the redirection setup on server.w.o to support ssl-capable and older versions it'd be nice to leave the handshake up to the redirection decision unconditionally unencrypted. For this the network code (client and server side) would need to be able to delay the encryption handling up to that point.

For wesnothd, since redirection is done by a wesnothd instance that does nothing else but redirection, SSL would be a compile-time setting.  The wesnothd redirector instance would be built with it disabled and the other instances that players play games on would be built with it enabled.

> The alternative would be to use a new network port for 1.16+. Perhaps abandoning the redirection approach (not removing the feature, just not using it in the normal operation of server.w.o) and keep using a new port for each new wesnoth version like we do with campaignd.

The one additional thing to note is that the current setup with a redirector could be reintroduced with 1.18 since at that point all wesnothd instances would have been built with SSL enabled.

> So regarding the technical details, currently we're using boost::asio::ip::tcp::socket which for encryption would change to boost::asio::ssl::stream<boost::asio::ip::tcp::socket>. The underlying tcp::socket would still be accessible and could be used as before, right? Then async_handshake is called to setup ssl? And that could as well be called at any later time?

That's what the SO answer said, but that's the only place I was able to find that suggested doing that as a solution.  Whether it works or not is currently untested.  The underlying socket is available by calling `lowest_layer()`.

soliton-:
> > The server needs runtime encryption support as a config option. With this encryption or not is decided at start-up time. (No real point in supporting a change at config reload time.)
> > Additionally to allow the redirection setup on server.w.o to support ssl-capable and older versions it'd be nice to leave the handshake up to the redirection decision unconditionally unencrypted. For this the network code (client and server side) would need to be able to delay the encryption handling up to that point.
> 
> For wesnothd, since redirection is done by a wesnothd instance that does nothing else but redirection, SSL would be a compile-time setting. The wesnothd redirector instance would be built with it disabled and the other instances that players play games on would be built with it enabled.
> 
The redirector is currently the trunk/master instance of wesnothd. You can also connect to it without getting redirected. Also every other wesnothd instance also redirects you to the appropriate version if you connect with the wrong one. But regardless I think it's not a great idea to have different builds for ssl and non-ssl wesnothd from a usability side and it does not seem that helpful from the implementation side either. The code still needs to deal with both cases.

Pentarctagon:
> Also every other wesnothd instance also redirects you to the appropriate version if you connect with the wrong one.

I'd expect that someone directly connecting to the wrong wesnothd instance is very rare since most players would just be clicking Join Official Server, so I think that the 1.14 instance not being able to redirect someone to the 1.16 instance and vice versa is a non-issue.

> But regardless I think it's not a great idea to have different builds for ssl and non-ssl wesnothd from a usability side and it does not seem that helpful from the implementation side either. The code still needs to deal with both cases.

It's simpler to add support with a compile time option for the servers rather than a run time option, which loonycyborg has done: 67d2040156e67291d03ca57af8f4122ec42d8b45

loonycyborg:
Changing server to support ssl and non-ssl at the same time will require all functions that use socket_ptr to be revamped to use template arg instead.

Pentarctagon:
@soliton- At this point it's really up to you as far as how important it is for the servers to be able to support both SSL and non-SSL based on a run time setting vs a compile time option.  If you consider it to be a very important feature, then fine, it can be done, but to me at least the additional work and code complexity doesn't match up with the benefit it would provide.

soliton-:
I do not consider multiple builds to get SSL and non-SSL enabled servers as reasonable.

Also I don't think it's that big of a deal to adjust all socket_ptr uses. Most if not all of it should just be a mechanical search and replace. And as it seems to me with that we should get the issue of an unconditionally unencrypted handshake fairly easily as well and thus keep backwards compatibility to continue redirecting or at least show a readable error to older clients.

loonycyborg:
I think it can work nicely. master server that works in non-ssl mode and only redirects new connections both to new and legacy servers and new 1.16+ servers that work in ssl only mode which 1.16+ clients will support.

loonycyborg:
That said we most likely will get rid of this as compile option for server eventually but that would be done better after revamping it to use coroutines because that would drastically decrease amount of handlers needing the change.

loonycyborg:
Most likely using boost.coroutine won't require even a boost version requirement bump, it's quite an old library.

loonycyborg:
Also need to settle on a protocol how it will be decided to use ssl in connection. Have server send special value/WML tag at some point? Have separate port range for ssl servers?

Pentarctagon:
Given the redirector will be staying unencrypted, and if it's going to be setup that there's a wesnothd instance that does only redirecting, then the only case where the client would need to assume encrypted communication from the start would be if the server name matches an official server but the port doesn't (aka someone is connecting directly to an official server).

Pentarctagon:
Though actually, given that every wesnothd instance is going to have the redirector code present, I guess every initial connection would be non-ssl and then wesnothd will need to either redirect or tell the client whether to use ssl or not.

soliton-:
> Also need to settle on a protocol how it will be decided to use ssl in connection. Have server send special value/WML tag at some point? Have separate port range for ssl servers?

http://wiki.wesnoth.org/MultiplayerServerWML
The client version response could be extended by a capabilities/features key that contains SSL.
Then the server could indicate SSL use in the mustlogin request. Either based on the client feature or just based on its settings. It should send an error to clients that are not SSL capable if the server is setup to do forum auth.

Pentarctagon:
Closing since this isn't the PR that will be used for this.

PR #5301, opened 2020-11-22
Test config::add_child

CrawlCycle:
This commit tests the three config::add_child in src/config.cpp.
The config::add_child are among the most popular methods and
have at least 150 usages.

### DIFF
--- a/src/tests/test_config.cpp
+++ b/src/tests/test_config.cpp
@@ -15,14 +15,17 @@
 #define GETTEXT_DOMAIN "wesnoth-test"
 
 #include <boost/test/unit_test.hpp>
+// Work around a Boost<1.67 bug fixed in 1.67: Include test_case.hpp after
+// unit_tests.hpp. https://svn.boost.org/trac10/ticket/13387
+#include <boost/test/data/test_case.hpp> // for parametrized test
 #include <cmath>
 
 #include "config.hpp"
 #include "variable_info.hpp"
 
-BOOST_AUTO_TEST_SUITE ( test_config )
+BOOST_AUTO_TEST_SUITE(test_config)
 
-BOOST_AUTO_TEST_CASE ( test_config_attribute_value )
+BOOST_AUTO_TEST_CASE(test_config_attribute_value)
 {
 	config c;
 	const config& cc = c;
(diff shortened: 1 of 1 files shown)


CrawlCycle:
Parametrized test and Boost version
-----------------------------------
Boost 1.58 and below has a macro `BOOST_PARAM_TEST_CASE` that
1. Don't automatically register the test
2. Only allow one parameter
Boost 1.59 and above has a `BOOST_DATA_TEST_CASE` macro that is much better.

https://github.com/wesnoth/wesnoth/blob/master/INSTALL.md says that
Wesnoth requires Boost libraries >= 1.56.0 

Ubuntu 16.04 ships with Boost 1.58 and will reach end of life on April 2021.

Workaround
----------
Therefore, I rewrite my parametrized unit test as a loop.
This work-around won't identify each parametrization of the test.
When one of the parametrization fails, we need to run 
`boost_unit_tests --log_level=message` to print messages
that helps to identify which parameter fails the test.

Macro in Boost 1.59 and above
--------------------
Example usage:

```c++
#include <boost/test/unit_test.hpp>
// Include test_case.hpp after unit_tests.hpp to work around
// https://svn.boost.org/trac10/ticket/13387
#include <boost/test/data/test_case.hpp>  // for parametrized test

namespace bdata = boost::unit_test::data;

BOOST_DATA_TEST_CASE (
    	test_add_child__key_value,
    	bdata::make({true, false}) * bdata::make({"A", "B", "C"}),
    	test_lvalue_ref, new_key_in )  // 2 * 3 = 6 cases
{
...
```

Boost would identify the parametrizations of the test in the example as:

    test_case_name/_1
    test_case_name/_2
    test_case_name/_3
    test_case_name/_4
    test_case_name/_5
    test_case_name/_6

I wish it would add the name and value of the parameters...

Pentarctagon:
Wesnoth's current standard is to support the current and previous Ubuntu LTS, which right now would be 20.04 and 18.04.  So I don't see an issue with increasing the minimum Boost version to 1.59 if it's needed.

CrawlCycle:
I think I want to use `BOOST_DATA_TEST_CASE` of Boost 1.59+. I will revert the last commit.

The documentation says boost will report the sample (the value of the parameters) upon test failure:
"in case of error, the context ... is reported ... This context contains the *sample* for which the test failed... " -- [Boost 1.59 documentation](https://www.boost.org/doc/libs/1_59_0/libs/test/doc/html/boost_test/tests_organization/test_cases/test_case_generation/datasets_auto_registration.html).

src/tests/test_config.cpp
```diff
@@ -286,4 +289,58 @@ BOOST_AUTO_TEST_CASE ( test_variable_info )
 	}
 }
 
+/**
+ * Test if config::add_child(key) correctly adds a child with the key.
+ * @pre The constructors of config are correct.
+ */
+BOOST_AUTO_TEST_CASE ( test_add_child__key )
```
stevecotton, line 296: In C++, identifies with a double-underscore are reserved, even when it's in the middle of the identifier.

CrawlCycle: I fixed the name and pass the test parameters directly into the test code.

src/tests/test_config.cpp
```diff
@@ -15,6 +15,9 @@
 #define GETTEXT_DOMAIN "wesnoth-test"
 
 #include <boost/test/unit_test.hpp>
+// Include test_case.hpp after unit_tests.hpp to work around
```
stevecotton, line 18: Please add that this was fixed in 1.67, so we know when we can update it

src/tests/test_config.cpp
```diff
@@ -286,4 +289,51 @@ BOOST_AUTO_TEST_CASE ( test_variable_info )
...
+BOOST_AUTO_TEST_CASE (
+		add_child_ThisIsEmpty_newKey_AppendEmptyChildToThis_ReturnNewChild )
+{
+	config actual;
+	const config new_child = actual.add_child("A");
+	const config expected("A");
+	// Assert add_child adds the child correctly
+	BOOST_CHECK_EQUAL (actual, expected);
+	// Assert the new child returned by config::add_child is correct
+	BOOST_CHECK_EQUAL (new_child, config());
```
stevecotton, line 301: Your indentation and spacing is better than the result of running clang-format on this file, however running clang-format and then cherry-picking the good differences would be better than both. For example, our style is not to have a space before the bracket in `if(` or `config(`, so it makes sense to omit it from `BOOST_CHECK_EQUAL(` too.

src/tests/test_config.cpp
```diff
@@ -286,4 +289,51 @@ BOOST_AUTO_TEST_CASE ( test_variable_info )
...
+BOOST_DATA_TEST_CASE (
+		add_child_ThisIsNotEmpty_newOrExistingKey_lOrRValue_AppendNewChildToThis_ReturnNewChild,
+		bdata::make<std::string>({"A", "B", "C"})
+		* bdata::make<std::string>({"lvalue_ref", "rvalue_ref"}),
+		key, update_ref_kind)  // 3 * 2 = 6 cases
+{
+	// Data for testing base.add_child(key, update)
+	const config base {"A", config(), "a", 1};   // [A][/A] a = 1
+	const config update {"B", config(), "b", 2}; // [B][/B] b = 2
+	// Make expected
```
stevecotton, line 314: Please move the after-the-semicolon comments to this line, so it's more like

    // Expected structure is [A][/A] a=1 [new_key][B][/B] b=2 [/new_key]

src/tests/test_config.cpp
```diff
@@ -286,4 +289,51 @@ BOOST_AUTO_TEST_CASE ( test_variable_info )
...
+
+	// Make actual
+	config actual(base);
+	config new_child;
+	if (update_ref_kind == std::string("lvalue_ref"))
+		new_child = actual.add_child(key, update);  // lvalue reference
+	else
+		new_child = actual.add_child(key, config(update)); // rvalue ref.
+
+	// Assert base.add_child(update) == expected
```
stevecotton, line 331: Some of these comments aren't needed, they're obvious from the code

src/tests/test_config.cpp
CrawlCycle: I make my IDE use clang-format to format the code.

stevecotton (approved):
Looks good, are you ready for it to be merged?

CrawlCycle:
Yup. I am ready for it to be merged. Thanks for the review.

Issue #1352, opened 2017-05-08
Nested #define's (GNA #21343)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?21343)** by laerne on 2013-12-13

This code :

#define F Y Z

  #define G A B

  {H Y A B}

  #enddef

    {K G {Z}}

  #undef G

#enddef

fails with "Macro/file 'Z' is missing at a/_main.cfg:10", the macro parameters seems to have been undefined with the first #enddef

At least one of the following behavior should happen :

* Define should be nestable.

* inside a #define, lines beginning by #define are comments.

I'm not hiding that I truly prefer the former solution, because it allows simulate lambda function for callback macros (As show in the first example where K is a macro taking the name of a macro).

(Reproduced on Any)
Release: 1.10.6
Priority: 3 - Low
Severity: 1 - Wish

wesnoth-bugs:
**Modified on 2013-12-13**

**laerne added attachment:** [_main.cfg](http://gna.org/bugs/download.php?file_id=19465)

wesnoth-bugs:
**Modified on 2013-12-14**

**jamit wrote:**

> Just observing: if we were to take your example code, and apply the rule "inside a #define, lines beginning by #define are comments", then stripping out that comment line would leave
> 
> 
> 
> #define F Y Z
> 
> {H Y A B}
> 
> #enddef
> 
> {K G {Z}}
> 
> #undef G
> 
> #enddef
> 
> 
> 
> This looks like it would fail with "Macro/file 'Z' is missing at a/_main.cfg:9". Not exactly an improvement in the end behavior.

wesnoth-bugs:
**Modified on 2013-12-16**

**jamit wrote:**

> I looked at the source and ran a few tests, and I think the second solution is already in place: inside a #define, lines beginning with "#define" *are* comments. So I guess that means this issue is already resolved? (I mean, one of the proposed solutions is in place, so what else would be needed to close this?)

wesnoth-bugs:
**Modified on 2013-12-16**

**laerne wrote:**

> I believe this should converted into a feature request for the first solution and not a bug.
> 
> 
> 
> What I mean with the second solution, is that the full block within beginning with #define (and not only the lines...) are comment.  But I realize this is not easier than the first solution, so it is better to have the first as a feature request, I think.

wesnoth-bugs:
**Modified on 2013-12-16**

[...]

PR #5303, opened 2020-11-23
Some refactoring of preprocessor macro definition parsing

CelticMinstrel:
While trying to fix the schema test failure (by stripping out comments from the macro definition), I did some refactoring of the logic so it looks more sane and (in my opinion) easier to understand. Although the attempt at filtering out comments ended in failure, I figured the refactoring could still be worthwhile.

### DIFF
--- a/src/serialization/preprocessor.cpp
+++ b/src/serialization/preprocessor.cpp
@@ -54,6 +54,12 @@ static std::string preprocessor_error_detail_prefix = "\n    ";
 
 static const char OUTPUT_SEPARATOR = '\xFE';
 
+static std::set<std::string> known_directives = {
+	"define", "enddef", "arg", "endarg", "undef",
+	"ifdef", "ifndef", "ifhave", "ifnhave", "ifver", "ifnver", "endif",
+	"error", "warning", "deprecated"
+};
+
 // get filename associated to this code
 static std::string get_filename(const std::string& file_code)
 {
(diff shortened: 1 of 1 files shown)


src/serialization/preprocessor.cpp
```diff
@@ -1156,112 +1162,131 @@ bool preprocessor_data::get_chunk()
 
 			std::string symbol = items.front();
 			items.erase(items.begin());
-			int found_arg = 0, found_enddef = 0, found_deprecate = 0;
+			// int found_arg = 0, found_enddef = 0, found_deprecate = 0, found_if;
```
stevecotton, line 1165: Leftover line that could be deleted?

src/serialization/preprocessor.cpp
```diff
@@ -1156,112 +1162,131 @@ bool preprocessor_data::get_chunk()
...
+							deprecation_detail += '\n';
 						}
+						deprecation_detail += read_rest_of_line();
+						skip_eol();
+					} else if(directive == "enddef") {
+						buffer.erase(found_directive - 1, std::string::npos);
+						found_enddef = true;
+						break;
+					} else if(directive == "define") {
+						// TODO: Maybe add support for this? This would fill feature request #21343
```
stevecotton, line 1278: GNA 21343 is Github #1352. However, I think the comment can just be deleted as "wontfix" sounds right to me.

src/serialization/preprocessor.cpp
CelticMinstrel: Yup.

PR #5305, opened 2020-11-23
SotA: Fix translation hints

stevecotton:
They need to be on the line above the string, instead of being a comment
following on the same line as the string.


stevecotton:
Can be cleanly cherry-picked to 1.14.

ivanovic (approved):

stevecotton:
Merged by pushing branches so that I could retitle it to "SotA: fix some more translation hints", as the parent commit is "Sota: add translation hints". It's showing as closed instead of merged because I pushed the branches in the wrong order.

PR #5309, opened 2020-11-23
SotA S04: Run wmllint, which fixed some double-spaces in strings

stevecotton:
These texts are only in the master branch, no need to backport to 1.14.


stevecotton:
@sevu although I added Doofus as the reviewer, I think either of you could sanity-check and merge it. Please would you have a look?

PR #5310, opened 2020-11-23
Document the public API of src/config.hpp

CrawlCycle:


CrawlCycle:
I think tests that verify the correctness of the documentation would be nice.

PR #5312, opened 2020-11-24
Test features of src/serialization/preprocessor.hpp

CrawlCycle:


CrawlCycle:
I have zero idea about the meaning of the return value of 
```c++
/**
 * Function to use the WML preprocessor on a file.
 *
 * @param defines                 A map of symbols defined.
 *
 * @returns                       The resulting preprocessed file data.
 */
filesystem::scoped_istream preprocess_file(const std::string& fname, preproc_map* defines = nullptr);
```

Here is the output of my test:
```
--- The input ----
#define Macro
ABC
#enddef
{Macro}
{Macro}{Macro}
{Macro}{Macro}{Macro}

--- Call preprocess_file ----
--- The output ----
�line 1 dd
�line 1 dd 1 dd
ABC
�line 1 dd




�line 1 dd 5 dd
ABC
�line 5 dd
�line 1 dd 5 dd
ABC
�line 5 dd

�line 1 dd 6 dd
ABC
�line 6 dd
�line 1 dd 6 dd
ABC
�line 6 dd
�line 1 dd 6 dd
ABC
�line 6 dd


--- The preproc_map ----
1st: Macro
2nd: value: ABC
 arguments: dd

```

PR #5313, opened 2020-11-24
SotA S10: Add some translation hints

stevecotton:
@sevu, I was wondering if the de version of the line should be "Wart ihr noch nie angeln?". It sounds okay to me with "du", but I'm not a native speaker.


sevu (approved):
Yes, German version should be "Wart ihr noch nie angeln?", in the case when both persons are addressed. The version with "du" is for singular.

Btw, a third way exists, in the case one wants to sound very formal and overly fancy, one can use "Wart Ihr noch nie angeln?" (with capital Ihr) to address a single person in plural form.
Though in the translation files are many cases where it is falsely written in lowercase.

PR #5318, opened 2020-11-27
NR S2_1: remove rails from castle hexes; fixes #5317

### LINKED ISSUE #5317: NR S2: rails displayed on castle (discussion/not a bug).
@doofus-01 

Not sure if this applies to other cave scenarios as well, but the display of rails through a castle looks weird to me. Here is a screenshot from the scenario:

rails_on_keep

Here is a contrasting screenshot from SoF, where rails don't get placed on the same tile as castle/keep:

keep_on_rails

In the case of the first screenshot you _could_ argue that the castle is just walls around a dirt field and then it kind of makes sense, but having the palisades on top of the tracks looks a) strange, b) impractical, and c) unstable to me. The patches of light dirt to contrast the surrounding dark dirt from the castle dirt also don't help.

Thinking about it again, I am a little torn if hiding the rails below the castle would make for better visuals though.

IskarJarak:
Removed rails from various castle hexes on the map, as that looked wrong.


sevu:
The encampments could have been set afterwards when the dwarves were gone. Both would work.

IskarJarak:
@sevu see the discussion on the referenced issue, this is about the visuals of rails displayed on castle hexes, not about lore justifications. The feedback was that it looked weird and that removal would be easier than creating new transitions. (though the issue is pretty new, feel free to join in 😄 )

Issue #5308, opened 2020-11-23
[resistance_defaults] and [terrain_defaults] are broken


CelticMinstrel:
Assuming the existence of a unit with a `type=sonic` attack, the following code works on 1.14.4 but does not work on 1.15 master:

```
[resistance_defaults]
	id=sonic
	default="(impact)"
[/resistance_defaults]
```

**What you should see:** If you place any unit, highlight it, and press D, it should show sonic resistance equal to their impact resistance.

**What you actually see:** Sonic resistance is not shown at all unless the unit has a movetype that explicitly specifies it.

Note: I didn't test, but it's possibly that `[terrain_defaults]` is similarly broken.

Pentarctagon:
> 1.14.4

Did you mean 1.14.14?

CelticMinstrel:
No, I haven't actually installed the latest stable version for several months, so it is indeed 1.14.4.

Pentarctagon:
It looks like it broke somewhere between 1.14.7 and 1.14.9 (1.14.8 doesn't exist).

Pentarctagon:
Removed the Blocker: Point label, seeing as this has been broken for multiple releases already without being noticed or reported.

Pentarctagon:
@stevecotton it looks like this is caused by fcd0ceda37793f79ec59e76a2f621fb07af64c5b

stevecotton:
Self-assigned and will investigate.

Does anyone have a good idea how to write a unit test for this, without adding an extra resistance to the units in all tests?

stevecotton:
The commit that's just auto-linked above fixes part of this, which was caused by a string being "resistance" in one location and "resistances" in another. I'll continue debugging.

This now works (giving 70% defense):

        [resistance_defaults]
            id="special_res_for_test"
            default="30"
        [/resistance_defaults]
    
But using formulas still doesn't. It creates the new type but with the default value, not the value from arcane:

        [resistance_defaults]
            id="copy_of_arcane"
            default="(arcane)"
        [/resistance_defaults]

stevecotton:
@Pentarctagon I'm wondering if it should be fixed or left as-is on the 1.14 branch. Either is going to be a change from a previous 1.14 release, meaning it could OOS a game - leaving it as-is means the OOS only occurs if a player has a version from before September 2019, which fixing it would give OOS if 1.14.15 was played with 1.14.14.

stevecotton:
[...]

PR #5319, opened 2020-11-27
Unit test for resistances, could do with Lua advice

stevecotton:
Add a unit test for resistances (using the standard 6 damage types)

I'm working on a bugfix for [resistance_defaults] (bug #5308), for which the
unit test is going to be out-of-tree until we support some method of testing
it without affecting other tests. The version here is the part of that which
can be in-tree, because it doesn't change the core unit's stats.

This uses some Lua to check that the units have only the expected
resistance types, but I feel there should be a better way to write it.

### DIFF
--- a/data/test/scenarios/test_resistances.cfg
+++ b/data/test/scenarios/test_resistances.cfg
@@ -1,0 +1,46 @@
+# wmllint: no translatables
+
+{GENERIC_UNIT_TEST "test_resistances" (
+    [event]
+        name = start
+
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=alice
+        [/store_unit]
+
+        # Elvish archers have a slight malus against arcane
+        {ASSERT {VARIABLE_CONDITIONAL alice.resistance.arcane equals 110}}
+        {ASSERT {VARIABLE_CONDITIONAL alice.resistance.impact equals 100}}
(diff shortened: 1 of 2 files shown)


stevecotton:
The Lua works, so IMO this can be merged as-is. I think there must be a better way to do it, but it's functional and only in a unit-test.

CelticMinstrel:
Looks good to me. Is it useful to do the same for movement costs and defense values?

stevecotton:
> Is it useful to do the same for movement costs and defense values?

They're fairly similar to this, as long as it's an out-of-tree test I don't think there's much value in adding the in-tree shadow of it. There's also already a test for movement costs that tests the pathfinding code and cumulative costs after moving over several mixed-terrain hexes, so it's an area that already has test coverage.

Testing [terrain_defaults] just needs lines like this added to the file in this PR:

```
        {ASSERT {VARIABLE_CONDITIONAL alice.movement_costs.hills equals 2}}
        {ASSERT {VARIABLE_CONDITIONAL alice.movement_costs.special_terrain_for_test equals 1}}
        {ASSERT {VARIABLE_CONDITIONAL bob.movement_costs.hills equals 1}}
        {ASSERT {VARIABLE_CONDITIONAL bob.movement_costs.special_terrain_for_test equals 3}}
```

PR #5323, opened 2020-11-28
units - undead variation for sea serpent(ish) units

doofus-01:
I see this working for Sea Serpent, Water Serpent, and Swamp Lizard (Crocodile), and similar UMC.  It's not going to be a perfect match, but it's better than the shark-head merman thing we've currently got (which is fine for mermen, nagas, whatever humanoid hybrids).

- [x] walking corpse sprite frames

- [x] soulless sprite frames

- [x] portrait


doofus-01:
I don't think the CI errors have anything to do with this.

CelticMinstrel:
I still think separate merfolk and naga variations would be better though - merfolk are fish-like but nagas are serpentine.

PR #5326, opened 2020-11-30
Add '/roll N' command to games and game lobbies.

### LINKED ISSUE #5221: Add "/roll N" command to chat rooms
In any chat room user could execute `/roll Number` command to produce **publicly visible** random number from 1 to N.

Like:
/roll 5
_wesnothd:_ 3

### For what a reason?
This feature could help to choose something between at least 2 people (duels), or 2 teams (2 vs 2 games) in multiplayer games.

Example: 
2 teams could not decide who should play at 1&4 slots - this random generation could help them to do fairly and random choose.

kabachuha:
Adds '/roll N' command which gets a random number between 1 and N and prints it as "$username rolled a dice [1 - $N] and got a $value." to the game creation lobby chat or just to the sender otherwise (so not to flood the main server chat and ensure a sense of privacy).

Closes #5221 .

### DIFF
--- a/src/server/wesnothd/server.cpp
+++ b/src/server/wesnothd/server.cpp
@@ -211,7 +211,7 @@ const std::string help_msg =
 	" dul|deny_unregistered_login [yes|no], kick <mask> [<reason>],"
 	" k[ick]ban <mask> <time> <reason>, help, games, metrics,"
 	" netstats [all], [lobby]msg <message>, motd [<message>],"
-	" pm|privatemsg <nickname> <message>, requests, sample, searchlog <mask>,"
+	" pm|privatemsg <nickname> <message>, requests, roll <sides>, sample, searchlog <mask>,"
 	" signout, stats, status [<mask>], stopgame <nick> [<reason>], unban <ipmask>\n"
 	"Specific strings (those not in between <> like the command names)"
 	" are case insensitive.";
@@ -226,6 +226,7 @@ server::server(int port,
 	, ip_log_()
 	, failed_logins_()
 	, user_handler_(nullptr)
+	, die_(static_cast<unsigned>(std::time(nullptr)))
 #ifndef _WIN32
 	, input_path_()
 #endif
@@ -385,6 +386,7 @@ void server::setup_handlers()
 	SETUP_HANDLER("version", &server::version_handler);
 	SETUP_HANDLER("metrics", &server::metrics_handler);
 	SETUP_HANDLER("requests", &server::requests_handler);
+	SETUP_HANDLER("roll", &server::roll_handler);
 	SETUP_HANDLER("games", &server::games_handler);
 	SETUP_HANDLER("wml", &server::wml_handler);
 	SETUP_HANDLER("netstats", &server::netstats_handler);
@@ -1152,7 +1154,7 @@ void server::handle_query(socket_ptr socket, simple_wml::node& query)
 
 	const std::string& query_help_msg =
 		"Available commands are: adminmsg <msg>, help, games, metrics,"
-		" motd, netstats [all], requests, sample, stats, status, version, wml.";
+		" motd, netstats [all], requests, roll <sides>, sample, stats, status, version, wml.";
 
 	// Commands a player may issue.
 	if(command == "status") {
@@ -1167,6 +1169,7 @@ void server::handle_query(socket_ptr socket, simple_wml::node& query)
 		command == "netstats all" ||
 		command.compare(0, 7, "version") == 0 ||
 		command == "requests" ||
+		command.compare(0, 4, "roll") == 0 ||
 		command == "sample" ||
 		command == "stats" ||
 		command == "status " + player.name() ||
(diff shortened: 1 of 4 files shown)


Pentarctagon:
What happens if the player enters a number larger than can be stored in an int?

kabachuha:
@Pentarctagon Now that I changed the function to the more safe variant, it will print a warning message that the value is too big. (starting with 2,147,483,648)

AI0867:
"Dice" is plural. The singular is "die".

src/server/wesnothd/server.hpp
```diff
@@ -207,6 +207,7 @@ class server : public server_base
 	void stats_handler(const std::string &, const std::string &, std::string &, std::ostringstream *);
 	void metrics_handler(const std::string &, const std::string &, std::string &, std::ostringstream *);
 	void requests_handler(const std::string &, const std::string &, std::string &, std::ostringstream *);
+	void roll_handler(const std::string&, const std::string&, std::string&, std::ostringstream*);
```
AI0867, line 210: On the other lines there's a space between the base type and the reference or pointer modifier. (`std::string &` rather than `std::string&`)

kabachuha: Sorry, the autoformatter.

src/server/wesnothd/server.cpp
```diff
@@ -2131,6 +2133,56 @@ void server::requests_handler(const std::string& /*issuer_name*/,
...
+		*out << "The upper value of the toss is too big!";
+		return;
+	}
+
+	if(N < 1) {
+		*out << "The upper value of the toss must not be lower than 1!";
+		return;
+	}
+	static randomness::mt_rng dice;
+	std::string value = std::to_string(dice.get_next_random() % N + 1);
```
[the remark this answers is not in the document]
CelticMinstrel: Use `std::uniform_int_distribution` to generate it?

CelticMinstrel:
This is... definitely something. I'm not opposed to the basic idea of having a dice-roll command, but... I guess this just seems really startlingly hyperspecialized? Not that I'm suggesting we should do a whole complex dice-evaluation parser, mind you... but supporting only a single die with nothing special like modifiers or the like just seems like a pointless endeavour.

Don't take this as an argument that it shouldn't be included in its current state, but... I'm a little baffled at how simplistic it is.

Incidentally, if it were decided that allowing expressions _is_ a good idea, it's not actually required to write a new parser - you could take the argument to be Wesnoth Formula Language code and just compile and evaluate it. That would allow stuff like `/roll d6` and `/roll 2d4` and `/roll 3d6+4`, for example. Now, whether this is actually a _good_ idea... I don't know. After all, it would also allow crazy-complicated formulas... and I'm also unsure how safe it is to allow players to type in an arbitrary formula.

src/server/wesnothd/server.cpp
```diff
@@ -211,7 +211,7 @@ const std::string help_msg =
 	" dul|deny_unregistered_login [yes|no], kick <mask> [<reason>],"
 	" k[ick]ban <mask> <time> <reason>, help, games, metrics,"
 	" netstats [all], [lobby]msg <message>, motd [<message>],"
-	" pm|privatemsg <nickname> <message>, requests, sample, searchlog <mask>,"
+	" pm|privatemsg <nickname> <message>, requests, roll, sample, searchlog <mask>,"
```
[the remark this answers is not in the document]
kabachuha: I don't think the defaulting would be useful. It can confuse people who might run the command in hurry forgetting to put a number and getting a result in overlooked bounds.

The argument `<sides>` note ought to be set indeed.

kabachuha:
@soliton- corrected most of the noted moments. However, rate limitting can be probably left out for now because the players can see those messages only in the game lobbies, which are easy controllable by the host, and the calculations are not expensive by themselves.

kabachuha:
@CelticMinstrel I think the proposal of entering computational formulae as a roll argument would unnecesseary complicate this action, which is supposed to just help people to choose an item from a set, and increase its runtime consts.

src/server/wesnothd/server.cpp
```diff
@@ -2131,6 +2134,49 @@ void server::requests_handler(const std::string& /*issuer_name*/,
...
+	auto player_ptr = player_connections_.get<name_t>().find(issuer_name);
+	if(player_ptr == player_connections_.get<name_t>().end()) {
+		return;
+	}
+
+	auto g_ptr = player_ptr->get_game();
+	if(g_ptr && g_ptr->is_member(player_ptr->socket())) {
+		g_ptr->send_server_message_to_all(issuer_name + " rolled a die [1 - " + parameters + "] and got " + value + ".", player_ptr->socket());
+	} else {
+		*out << " (The result is shown to the others only in game lobbies.)";
```
soliton-, line 2176: This is not entirely accurate since it's also shown once the game is started, right?

Perhaps should be: " (The result is shown to others only in a game.)"

src/server/wesnothd/server.cpp
```diff
@@ -2131,6 +2134,49 @@ void server::requests_handler(const std::string& /*issuer_name*/,
...
+	}
+
+	if(N < 1) {
+		*out << "The die cannot have less than 1 side!";
+		return;
+	}
+	std::uniform_int_distribution<int> dice_distro(1, N);
+	std::string value = std::to_string(dice_distro(die_));
+
+	*out << "You rolled a die [1 - " + parameters + "] and got " + value + ".";
```
soliton-, line 2165: I think this should say: "and got a " ...

src/server/wesnothd/server.cpp
```diff
@@ -2131,6 +2134,49 @@ void server::requests_handler(const std::string& /*issuer_name*/,
...
+
+	auto player_ptr = player_connections_.get<name_t>().find(issuer_name);
+	if(player_ptr == player_connections_.get<name_t>().end()) {
+		return;
+	}
+
+	auto g_ptr = player_ptr->get_game();
+	if(g_ptr && g_ptr->is_member(player_ptr->socket())) {
+		g_ptr->send_server_message_to_all(issuer_name + " rolled a die [1 - " + parameters + "] and got " + value + ".", player_ptr->socket());
```
soliton-, line 2174: As above.

soliton- (changes requested):

soliton-:
> However, rate limitting can be probably left out for now because the players can see those messages only in the game lobbies, which are easy controllable by the host, and the calculations are not expensive by themselves.

Yeah, it's not a big deal.

kabachuha:
@soliton- Ok, applied those wording suggestions.

src/server/wesnothd/server.cpp
```diff
@@ -2131,6 +2134,49 @@ void server::requests_handler(const std::string& /*issuer_name*/,
...
+
+	*out << "You rolled a die [1 - " + parameters + "] and got a " + value + ".";
+
+	auto player_ptr = player_connections_.get<name_t>().find(issuer_name);
+	if(player_ptr == player_connections_.get<name_t>().end()) {
+		return;
+	}
+
+	auto g_ptr = player_ptr->get_game();
+	if(g_ptr && g_ptr->is_member(player_ptr->socket())) {
```
soliton-, line 2173: The `is_member` call seems useless. If `g_ptr` is valid the player is part of the game.

kabachuha: It's being also checked at
 https://github.com/wesnoth/wesnoth/blob/680559037a4145be376833b0e055cf0e6ac2369a/src/server/wesnothd/server.cpp#L1126-1130 , I'm not quite sure about this.

soliton-: Yes, it is useless there as well. `is_member` checks for players or observers. There is nothing more in a game.

kabachuha: Fair enough, put the fix in a new commit just in case.

soliton-:
One last thing... the commit message still says game lobbies instead of games... :wink:

kabachuha:
@soliton- Renamed it.

soliton-:
Thanks!

amokmen:
@kabachuha and all others: thank you!

Just compiled and tested, working as a charm! :)

Issue #5325, opened 2020-11-30
jamming: is it used, would anyone know if it has broken for ages?


stevecotton:
Units have movement points, vision points (default to mp) and jamming points (default zero). Movetypes can have separate costs for movement, vision and jamming. I'm touching the movetype code, and am ensuring that support still exists for jamming, but have come to the conclusion that:

* I wouldn't be able to tell if jamming is working as intended or not
* The only units using it seem to be in data/scenario-test.cfg, where there's no hint about what it should do
* The documentation appears to be @fendrin's https://r.wesnoth.org/t36716 and the question https://r.wesnoth.org/t37503
* If I wanted to know more about it, then based on Git history (#2486 and #5308) the person to ask will be me.

So a question - is it used, and would anyone know if it was broken?

Pentarctagon:
It's also documented on the wiki under [unit_type].

Hejnewar:
I'm using it in my era and I should be able to tell if it was broken.

stevecotton:
Thanks - knowing that the given version works correctly, I can write some units tests for the stuff that I'm changing and improve the docs. Would ideally like to have an example showing which hexes will be affected, similar to https://wiki.wesnoth.org/FindPathExplanation

PR #5327, opened 2020-11-30
Fix [resistance_defaults] and [terrain_defaults] (fixes #5308)

### LINKED ISSUE #5308: [resistance_defaults] and [terrain_defaults] are broken

stevecotton:
The draft version included the testing which I'm going to keep out-of-tree, that's now in my stevecotton:debugging/testing_i2486 branch.

These now work:

	[resistance_defaults]
		id="special_res_for_test"
		default="30"
	[/resistance_defaults]

	[resistance_defaults]
		id="copy_of_arcane"
		default="(arcane)"
	[/resistance_defaults]

and so do these:

	[terrain_defaults]
		id="special_terrain_for_test"
		[movement_costs]
			default="(swamp_water + 1)"
			orcishfoot="(vision_costs.swamp_water * 2)"
		[/movement_costs]
	[/terrain_defaults]
	[terrain_defaults]
		id="special_terrain_for_test"
		[defense]
			default="(20 + 7 * movement_costs.special_terrain_for_test)"
		[/defense]
	[/terrain_defaults]

For [terrain_defaults], I've approached it as a new feature rather than a
simple fix. The subtags now use the same names as the [movetype] subtags and
[effect]'s `apply_to` attribute, so [terrain_defaults][movement_costs] instead
of [terrain_defaults][movement].

The formula handling will now recognise "resistance", "movement_costs",
"vision_costs", "jamming_costs" and "defense". For [resistance_defaults], the
formula will recognise both "(arcane)" and "(resistance.arcane)" as equivalent,
similarly for [terrain_defaults] "(swamp_water)" is a shorthand for whichever
subtag is being patched.

A [terrain_defaults] tag may use data added in a previous [terrain_defaults],
as in the examples above where the second tag's [defense] is based on the first
tag's [movement_costs], this gives orcish grunts on the special terrain a 62%
chance to be hit. However, if the [movement_costs] and [defense] were in a
single [terrain_defaults] tag then the result would be implementation defined,
because no guarantee is made of the order in which the children of the tag are
processed.

The schema gets fixed for [resistance_defaults] and [terrain_defaults], as it
only allowed one instance of each tag. The subtags of [terrain_defaults]
already had the new names.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,8 +13,9 @@
    * Some standing/bobbing animations now filtered for low HP (depicting exhaustion) (PR #5388)
  ### User interface
  ### WML Engine
-   * New [set_variable] options: reverse=yes, min=list, max=list
+   * New `[set_variable]` options: reverse=yes, min=list, max=list
  ### Miscellaneous and Bug Fixes
+   * Fixed `[terrain_defaults]` and `[resistance_defaults]` (issue #5308).
 
 ## Version 1.15.8
  ### Add-ons client
--- a/data/schema/units/movetypes.cfg
+++ b/data/schema/units/movetypes.cfg
@@ -27,6 +27,8 @@
 	[/tag]
 [/tag]
 
+# For adding a new terrain type or resistance type, forgetting to add a default is probably an error.
+# Vision and jamming are exceptions to this, as they fallback to the movement costs.
 #define MOVETYPE_PATCHING
 	{REQUIRED_KEY default f_int}
 	{ANY_KEY f_int}
@@ -34,12 +36,14 @@
 
 [tag]
 	name="resistance_defaults"
+	max=infinite
 	{REQUIRED_KEY id string}
 	{MOVETYPE_PATCHING}
 [/tag]
 
 [tag]
 	name="terrain_defaults"
+	max=infinite
 	{REQUIRED_KEY id string}
 	[tag]
 		name="defense"
@@ -51,10 +55,10 @@
 	[/tag]
 	[tag]
 		name="jamming_costs"
-		{MOVETYPE_PATCHING}
+		{ANY_KEY f_int}
 	[/tag]
 	[tag]
 		name="vision_costs"
-		{MOVETYPE_PATCHING}
+		{ANY_KEY f_int}
 	[/tag]
 [/tag]
--- a/src/movetype.cpp
+++ b/src/movetype.cpp
@@ -885,6 +885,9 @@ void movetype::merge(const config & new_cfg, const std::string & applies_to, boo
 	else if(applies_to == "resistance") {
 		resist_.merge(new_cfg, overwrite);
 	}
+	else {
+		ERR_CF << "movetype::merge with unknown applies_to: " << applies_to << std::endl;
+	}
 }
 
 /**
--- a/src/units/types.cpp
+++ b/src/units/types.cpp
@@ -19,6 +19,7 @@
 
 #include "units/types.hpp"
 
+#include "formula/callable_objects.hpp"
 #include "game_config.hpp"
 #include "game_errors.hpp" //thrown sometimes
 //#include "gettext.hpp"
@@ -32,7 +33,6 @@
 #include "gui/auxiliary/typed_formula.hpp"
 #include "gui/dialogs/loading_screen.hpp"
 
-#include <boost/regex.hpp>
 #include <boost/range/algorithm_ext/erase.hpp>
 
 #include <array>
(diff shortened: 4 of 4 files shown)


knyghtmare:
Yeah, the suggested WML API for both the '[resistance_defaults]` and '[terrain_defaults]' are acceptable to me. Thank you for taking the time to address and fix the problem.

CelticMinstrel (commented):
> then the result would be implementation defined.

Implementation defined is for when someone is writing a specification that others are expected to implement. That's not the case in Wesnoth, so I think we should define what it means.

Since using multiple tags allows basing them on each other, I'd say the logical choice would be that multiple subtags always read the value as of before the current tag modification. In other words, with this code:

```
[terrain_defaults]
    id="test_terrain1"
    [movement_costs]
        default="(movement_costs.swamp_water + 1)"
    [/movement_costs]
    [defense]
        default="(20 * movement_costs.swamp_water)"
    [/defense]
[/terrain_defaults]
[terrain_defaults]
    id="test_terrain2"
    [movement_costs]
        default="(movement_costs.swamp_water + 1)"
    [/movement_costs]
[/terrain_defaults]
[terrain_defaults]
    id="test_terrain2"
    [defense]
        default="(20 + movement_costs.swamp_water)"
    [/defense]
[/terrain_defaults]
```

I think it's reasonable to expect the defense on test_terrain2 to be 10 more than the defense on test_terrain1.

> for [terrain_defaults] that isn't provided and the old-style "(swamp_water)" now needs to be written as "(movement_costs.swamp_water)" - while the old style could be supported, the [terrain_defaults] tag already needs subtags and those subtags are already changed from 1.14's names.

Two comments on this. First, I'm not a fan of the greater length of the names here... I can probably live with it, but I imagine it's going to be extra confusing for people trying to figure out whether or not to put an S on `movement_costs` - on a WFL unit, it's singular. If you're changing things for consistency, there's a lot more than just this...

Secondly, I think it would make sense for the current tag to be the default, so if you write just "swamp_water" in the `[defense]` tag, then it's equivalent to "defense.swamp_water", but if you write it in the `[movement_costs]` tag, then it's equivalent to "movement_costs.swamp_water". I figure basing on the same attribute would be the most common case, after all (although basing jamming or vision on movement might also be fairly common, I suppose).

stevecotton:
> Implementation defined is for when someone is writing a specification that others are expected to implement. That's not the case in Wesnoth, so I think we should define what it means.

I think of "implementation defined" as "the specification leaves the details to the implementation, and the implementation might be quite complex with odd edge cases". While it could be defined exactly here as there's only one implementation, I'd prefer to allow refactoring to change the behavior in this edge case, and I don't see any benefit to WML authors from making the specification tighter, it would only save them a few extra lines for `[/terrain_defaults][terrain_defaults]id=...`. On the C++ side however, it makes a huge difference between having to calculate all the formulas, store the results, and then apply them in order; between the previous version and the just-pushed one there was a day of debugging until I realised that `cumulative_values`'s values were only being read after the vision and jamming values had been accumulated.

> `example code`
> I think it's reasonable to expect the defense on test_terrain2 to be 10 more than the defense on test_terrain1.

I think there's something missing in your example code - all values seem to depend only on the swamp_water values, which aren't changed by anything in that example code.

> it would make sense for the current tag to be the default, so if you write just "swamp_water"

The new push changes the behavior to do this.

> MacOS build failure

Seems to be missing libraries, unrelated to this PR.

CelticMinstrel:
> I think there's something missing in your example code - all values seem to depend only on the swamp_water values, which aren't changed by anything in that example code.

You're right - the defense for both terrain types was supposed to depend on the movement cost of itself:

```
[terrain_defaults]
    id="test_terrain1"
    [movement_costs]
        default="(movement_costs.swamp_water + 1)"
    [/movement_costs]
    [defense]
        default="(20 * movement_costs.test_terrain1)"
    [/defense]
[/terrain_defaults]
[terrain_defaults]
    id="test_terrain2"
    [movement_costs]
        default="(movement_costs.swamp_water + 1)"
    [/movement_costs]
[/terrain_defaults]
[terrain_defaults]
    id="test_terrain2"
    [defense]
        default="(20 * movement_costs.test_terrain2)"
    [/defense]
[/terrain_defaults]
```

stevecotton:
> Since using multiple tags allows basing them on each other, I'd say the logical choice would be that multiple subtags always read the value as of before the current tag modification. In other words, with [example code directly above]

I'm still not following this. For `test_terrain1`, if it's being calculated based on the value before the current `[terrain_defaults]` tag then I'd expect one of these results:
* `(20 * 0)`, based on there being no value for movement_costs.test_terrain1
* `(20 * 99)`, based on the default for an unreachable terrain
* a formula error resulting in the default value being used

CelticMinstrel:
Yeah, that makes sense, but there is a fourth possibility as well:

* The movetype already had an entry for `test_terrain1`, so the defense is twenty times that value, rather than being based on the swamp water value.

Mind you, now that I say it, I'm not sure we actually care about that case…

stevecotton:
@CelticMinstrel I think this is ready for merging, please would you comment if you disagree?

The build failure was the MacOS build configuration that's fixed in c7643e0785bdb5c4466e6defff17f921aea6b075. I'm about to push a no-change rebase to retrigger the CI. (I tried the CI's "retry" button for this, but it seems to use the original merge-PR-with-master version of the source instead of merging the PR with the current source).

For the Wiki documentation, I could rephrase "implementation defined" as "unsupported".

soliton-:
I think saying that the processing order is undefined or that you can't rely on it is understandable.

CelticMinstrel:
Why is the `default` key no longer required in the schema? Is there a good reason for this, or is it just because the code never _actually_ requires it?

> For the Wiki documentation, I could rephrase "implementation defined" as "unsupported".

Yeah, "unsupported" is definitely better than "implementation defined".

stevecotton:
> Why is the `default` key no longer required in the schema? Is there a good reason for this, or is it just because the code never _actually_ requires it?

I thought it would be useful for buffing existing move types, for example "orcs get extra vision over water in this campaign", and also because my unit test was itself failing the schema-test. However, I'm now wondering if it would be better to require an explicit declaration that the default should be the old value, and tell UMC authors deliberately want no default to use an `#ifndef SCHEMA_SHOULD_SKIP_THIS` around it*.

One thing I haven't tested is the effect of patching mixed terrains, for example an overlay where the `[terrain_type]` gives it worst-of (`_bas`, `fungus`) and then a `[terrain_defaults]` lets merfolk swim under it when it's on deep water**. These would have a problem with the default value - at the time the patching is done, we're working with only the basic terrain types and don't even know which mixed terrains we'll encounter on the map.

`*` I'm still hoping to make the schema a standard and easily-accessible tool for UMC authors.
`**` There might be a better way to do this terrain, by just creating a `[terrain_type]` for it's deep-water variation.

CelticMinstrel:
> I thought it would be useful for buffing existing move types, for example "orcs get extra vision over water in this campaign"

Ah, that actually _does_ sound like a good idea. I don't think there'd be any difference between explicitly setting `default="(this_terrain)"` and having no default, would there? (Where "this_terrain" is the terrain being patched.)

I think the engine reduces the terrain type to an archetype before looking up the value in the movetype, so probably patching something that's not an archetype just won't work?

stevecotton:
> I don't think there'd be any difference between explicitly setting `default="(this_terrain)"` and having no default, would there? (Where "this_terrain" is the terrain being patched.)

The only difference that I can think of is that vision and jamming costs would record a value instead of being "fall back to movement costs"; they wouldn't update if the movement costs were changed afterwards. I'm now wondering if the schema should make `default` optional for `vision_costs` and `jamming_costs`, but mandatory for the other types.

> I think the engine reduces the terrain type to an archetype before looking up the value in the movetype, so probably patching something that's not an archetype just won't work?

Correct. I'd forgotten that the movetype code did that.

CelticMinstrel:
> I'm now wondering if the schema should make default optional for `vision_costs` and `jamming_costs`, but mandatory for the other types.

Hmm, making it optional for `vision_costs` makes sense, at least, since vision defaults to movement by default… but why `jamming_costs` too? Doesn't jamming default to 0 if never set? Or am I misunderstanding how jamming works?

stevecotton:
I'm not sure that I understand jamming correctly, but #5325 says that's it's currently working, so I'm assuming that the refactor which caused #5308 didn't change the logic.

As with movement and vision, there's a `jamming` which is the number of jamming points (defaults to zero) and per-terrain `jamming_costs` which fall back to the vision costs. Some of the discussion threads said that `jamming_costs` default to `UNREACHABLE` rather than the vision costs, but this doesn't seem to be the case.

To my imagination, jamming in a fantasy setting should work differently to a sci-fi setting. In fantasy, I assume jamming would be fog or smoke and would be more difficult on terrains with low vision costs; on the other hand in sci-fi, sending out noise to jam long-range sensors would probably be easier on terrains with low vision costs. So the jamming that we have is the sci-fi version.

stevecotton:
@CelticMinstrel I'm planning to merge tomorrow unless there are further comments.

PR #5329, opened 2020-12-01
Converted uses of boost::optional to utils::optional

Vultraz:
Coverts usecases of `boost::optional` to `utils::optional`. Doesn't touch the few optional references since those aren't supported with std::optional except via `std::reference_wrapper` which adds an extra layer of indirection that makes conditional compilation difficult.

### DIFF
--- a/src/play_controller.hpp
+++ b/src/play_controller.hpp
@@ -25,6 +25,7 @@
 #include "terrain/type_data.hpp"
 #include "tod_manager.hpp"
 #include "game_state.hpp"
+#include "utils/optional_fwd.hpp"
 
 #include <set>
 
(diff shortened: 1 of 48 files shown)


src/play_controller.hpp
```diff
@@ -132,10 +133,14 @@ class play_controller : public controller_base, public events::observer, public
 		gamestate().end_level_data_ = data;
 	}
 	void reset_end_level_data() {
-		gamestate().end_level_data_ = boost::none;
+		gamestate().end_level_data_.reset();
 	}
 	bool is_regular_game_end() const {
-		return gamestate().end_level_data_.get_ptr() != nullptr;
+#if defined HAVE_CXX17 || BOOST_VERSION >= 106800
```
Vultraz, line 139: I realized I could have done `#ifdef HAVE_CXX17 ||` but is this ok?

CelticMinstrel: The parentheses are optional here, yes.

PR #5333, opened 2020-12-02
MP: unique background graphics for Isar's Cross

sevu:
Mainlines one of the two maps from VMP.
The graphics were already reviewed by @doofus-01.

TerrainGraphicsWML is black magic, this is mostly the code from VMP, without the need for a separate terrain code. Is there anything which could be done better about it or the wording of the text?
For storing the images, I put them into core, because unlike images elsewhere, "terrain/" is silently prefixed to the path. To store them somewhere else a new binary path would be needed.

I leave this open for a few days..


data/multiplayer/scenarios/4p_Isars_Cross.cfg
```diff
@@ -59,4 +59,116 @@
         gold=75
         fog=yes
     [/side]
+
+    [options]
+        [checkbox]
+            id=classical_graphics
+            default=no
+            # po: This string will not only be used for Isar’s Cross.
```
stevecotton, line 67: This hint doesn't say much about what the string to translate means. Does "classical" mean that the map will have orchestral music, the turn-timer replaced by an image of a conductor who waves their white baton as time runs out, and on the seventh turn the start-of-turn notification sound replaced with an unexpected cannon blast?

Vultraz: Suggest `Simple background` for a name.

stevecotton: I think the `name` should be written with title case (capital letters at the start of words), as title case is used for the "General" and "Game Settings" tabs.

Re-reading my previous comment, the joke about "classical" doesn't seem funny any more. Sorry for posting it.

Instead of a `#po` hint that it will be used for other maps, I'd suggest making it specific to this map for now.
```
    name=_ "isars^Disable Background Image"
    # po: Tooltip for the Isar's Cross map, makes the map look as it did in Wesnoth 1.14 and earlier
    description=_ "Show a blank area outside the map, instead of the floating island with cliffs."
```

doofus-01:
> TerrainGraphicsWML is black magic, this is mostly the code from VMP, without the need for a separate terrain code. Is there anything which could be done better about it or the wording of the text?

I will take a look.  I suspect there is a way to make the total file-size smaller.

data/multiplayer/scenarios/4p_Isars_Cross.cfg
sevu: Classical Graphics was my attempt to positively paraphrase "Don't use fancy graphics", but it's sounds a lot better do be direct about what it does.

Pentarctagon:
Any chance this could be ready for 1.15.9?

doofus-01:
I forgot about this, but if it is a priority, I think it can be ready for 1/16.

Pentarctagon:
It would be nice to get it in, I think, but not necessarily a priority for now.

doofus-01:
I looked into this, and think it would probably be simple to reduce the size with the layered file used to produce these PNGs, but from just the PNGs themselves, it would require manual massaging.

It is possible to convert the backgrounds to JPEGs, then use ~MASK() IPF to restore the transparency, but that really slows down the initial display.  I guess you could cut out a strip around the transparent hole, save that as a PNG and put the rest as a JPEG at a lower layer.

Pentarctagon:
@inferno8 Would you be able to give doofus-01 the original layered file?

inferno8:
@Pentarctagon Yes. I've just sent @doofus-01 a PM with the file.

doofus-01:
@inferno8 @Pentarctagon - I got the file, I'll look into it this weekend.  Thanks.

doofus-01 (approved):

PR #5338, opened 2020-12-03
campaignd: [server_id] request

shikadiqueen:
This new commands allows clients to request some identification and feature capability information from the server, intended to allow clients to find out what features the server supports and enable or disable UI elements accordingly while preserving compatibility where desired.

Currently this results in a response like this:

```
[server_id]
    version="1.15.7+dev (c7643e0785b-Modified)"
    cap="auth:legacy,delta"
[/server_id]
```

More information might be added later, such as a website URL and a link to a web listing for the particular server (e.g. <https://add-ons.wesnoth.org/1.16/>), more details on add-on licensing, or even doing away with the need to send `[request_terms]` before displaying the prompt on the client side.

Note: clients assume `cap=auth:legacy` for servers that can't respond to the query in this version. This accurately reflects the situation for Wesnoth 1.14 and earlier. The 1.15.x server will be updated once this feature goes live so that 1.15.8+ clients don't assume that delta packs are not supported even though they very much are since 1.15.6.

Also note that the client support is an early WIP right now. Ideally the code paths that prod the server to find out if delta uploads/downloads are supported will be replaced with a capability check instead. The API for this (`addons_client::server_supports()`) is not in the patchset yet.

### DIFF
--- a/data/gui/window/addon_license_prompt.cfg
+++ b/data/gui/window/addon_license_prompt.cfg
@@ -1,0 +1,123 @@
+#textdomain wesnoth-lib
+###
+### Definition of the Add-ons Manager license promnpt dialog
+###
+
+[window]
+	id = "addon_license_prompt"
+	description = "Add-ons Manager content license prompt displayed during uploads."
+
+	[resolution]
+		definition = "default"
+
+		automatic_placement = true
+		vertical_placement = "center"
+		horizontal_placement = "center"
+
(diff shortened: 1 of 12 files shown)


shikadiqueen:
The segmentation fault issue is fixed in commit ab770d1cbab23871a28bc20a74c859037f987eaf in master.

PR #5341, opened 2020-12-05
Use coroutines via boost::asio::spawn to improve overall code structure in servers

loonycyborg:
This will add dependency on boost.coroutine and boost.context which in turn supports only particular arches and OSes, but all platforms that wesnoth would be realistically run on are supported afaict.


Pentarctagon:
My main question would be whether these libraries are available on ARM macs and on iOS (@hrubymar10 and @singalen).

Vultraz:
Not being really familiar with coroutines I'm not sure I can offer much feedback here. 🤔

src/server/common/send_receive_wml_helpers.ipp
```diff
@@ -35,6 +35,36 @@
...
+		{
+			uint32_t size;
+			char buf[4];
+		} data_size;
+		data_size.size = htonl(s.size());
+
+		std::vector<boost::asio::const_buffer> buffers;
+
+		buffers.push_back(boost::asio::buffer(data_size.buf, 4));
+		buffers.push_back(boost::asio::buffer(s.begin(), s.size()));
```
Vultraz, line 54: Recommend initializer list.

Pentarctagon:
One thing I noticed as well when updating the docker images for coroutine is that there were warnings saying that coroutine is deprecated and coroutine2 should be used instead.  Is that at all relevant to this PR?

loonycyborg:
This PR relies on boost::asio::spawn and boost::asio::yield_context. They aren't deprecated. I suppose they can just swap underlying implementation from coroutine1 to coroutine2. If that ever becomes a problem there are also third party implementations of spawn() we could reuse.

Vultraz:
Should not we use `coroutines2` anyway?

loonycyborg:
They're not supported by asio atm, so would need to use custom code for that at least. This can always be done later.

soliton-:
Why squash the whole thing?

loonycyborg:
My thought was to reduce number of small commits to reduce the possibility that potential git bisect hits invalid/useless state.

loonycyborg:
Also all those commits logically belong together and all commit messages preserved anyway.

soliton-:
The "preserved" commit message is a mess with little chance to figure out what's what. It looks like a simple chronological list of commit messages. I suppose we can hope to stay on github so we can look at this PR still.

PR #5344, opened 2020-12-06
various work on new fauna

doofus-01:
miscellaneous additions/fixes to the ongoing wesnoth fauna project


doofus-01:
I believe the CI errors are from the parent, not anything in here.  If that's wrong, I can fix it tomorrow.

PR #5349, opened 2020-12-09
Wesnothd: added a login response, including admin status

Vultraz:
This is instead of having the server send a message after the login is complete and before the initial gamelist.
Since we wait at the loading screen for the latter (e41534d114437b26c735915f530df2cbc28e6c15), this has needed
to be handled here (2b4ae206ab796934899a628be6d3808a4ccd6c60), but it doesn't make much sense to be handled this
way. Many years ago this message was meant to show up in chat, but after the former change that no longer happened,
(IIRC) and after the latter change it most certainly won't.

This adds a proper [login_response] response from the server on a successful login which includes an `is_admin`
key containing the user's admin status.


soliton- (approved):
I would suggest to not call it admin on the client side either. (I know that is a pre-existing mess.) Otherwise looks good to me.

Issue #4543, opened 2019-11-03
Campaign Menu GUI design 


nemaara:
I'm posting this as a separate issue so we can keep track of things in one place, as per our discussion on discord (tagging @Pentarctagon, @stevecotton, @Vultraz). Basically this is for cleaning up the campaign menu a little bit and moving the Tutorial here as well. Items to discuss or take care of are:

1. Putting Tutorial as the first entry on the campaign list (thus doing away with the button in the title screen).
2. Making "save random seed" and "PRNG" modifications in the base screen (and removing the tick box for save random seed)
3. Removing campaign filtering (done, see https://github.com/wesnoth/wesnoth/commit/c7f2d7a22b707573900b9319e581da1956e0d168).

Pentarctagon:
The filtering may not be a good thing to get rid of, given what @Konrad22 said on Discord about him actually having 100+ campaigns installed.

shikadiqueen:
The Search box (which did briefly go away) is now in the top right corner sharing a row with the dialog caption, since this seems to make more sense as a general design approach across the board and makes it less obtrusive overall (commit d53ca602b8df5cc302cd1d5d502654823dc67415).

shikadiqueen:
Also worth noting is that I set custom labels for the OK buttons (now "Play") on the Campaigns menu proper and the difficulty selection dialog (commit 46dbbc06c967bfc8ee5acb6525e1f1bd96b53005). Somehow I was under the impression this already was the case before — or perhaps I intended to implement the same change early in 1.13.x and forgot about it entirely.

The reason I'm mentioning this here is that I feel it would make sense to have as part of a more unified design approach with these major dialogs a larger "primary action" button (in this case Play) like it's currently done in the MP Create Game screens and the Add-ons Manager.

nemaara:
As discussed on discord, it might be nice to make the campaign image a backdrop and make room to embed the difficulty menu into the campaign menu itself. So as an idea, we could have the red circled area be the difficulty menu, the blue circled area contains the campaign text, and the image is the background behind it.

![image](https://user-images.githubusercontent.com/13598479/68319098-c263dc80-008b-11ea-8639-7423686da4eb.png)

stevecotton:
What's the value to the player of moving the difficulty select into the campaign menu?

[...]

PR #5358, opened 2020-12-15
Campaigns menu overhaul

shikadiqueen:
This overhauls the game's Campaigns menu following some of the directions previously suggested in issue #4543.

Requesting reviews from @Vultraz for the UI layout and code changes, @gfgtdf for the engine changes, and @Pentarctagon and @nemaara for the core changes to the single-player campaign UI flow.

## User interface changes

Screenshot_20201215_002624
Screenshot_20201215_003209
Screenshot_20201215_003039

* The main area of the dialog is rearranged into four distinct sections:
  * The caption row, with the dialog caption and search box
  * The main panel, with the list of campaigns on the left, the campaign description and image on the right, and an image background.
  * The settings row, with the Modifications, Combat (RNG) and Difficulty menus arranged in that order.
  * The control row, with the Play and Cancel buttons.
* The toggle buttons serving as the headers for the list of campaigns have their text centered in order to make the list header seem more tidy in general.
* The **Difficulty menu** is integrated into the dialog in the form of a dropdown menu button. The old standalone menu remains available for use by Load Game and others that may need it.
* The Save random seed option is replaced with a new **Combat** dropdown menu button allowing the player to choose between Default RNG, Predictable RNG (identical to Save random seed), and Biased RNG (identical to the Advanced Preferences experimental PRNG option, **removed** by this PR).

## WML changes

* A new attribute, `background=<image path>` is added to top-level `[campaign]` to allow specifying a custom background image for the main panel. In the absence of this, a default image is used (currently `story/landscape-battlefield_nohumans.jpg`).
* The `random_mode=` attribute in game classification has a new possible value, `biased`, which selects the predictable RNG as with the Save random seed option before but also enables the RNG bias logic in the attack engine (previously the experimental PRNG option). The logic in the `attack` constructor may now be slightly redundant since it shouldn't be possible for `random_mode` to be `biased` and `randomness::generator->is_networked()` to return true (@gfgtdf may want to confirm this).
* `toggle_button` widgets gain new definitions: `listbox_header_centered` (unused), `listbox_header_bg` (unused), `listbox_header_centered_bg` (used by this PR for the campaign list header row).
* `panel` widgets gain new definitions: `panel_listbox_bg` (used by this PR for the campaign list header row), `campaign_menu_bg` (used by this PR for the main campaign menu panel).

## Known issues

* The Difficulty menu button has to compose labels from the label proper specified in `[difficulty]` WML as well as the description in order for the labels to make sense (e.g. for HttT, the difficulty labels would be the rather meaningless "Fighter", "Commander", "Lord" otherwise). The problem arises from the possibility of `[difficulty]` descriptions using Pango markup that will look plain wrong on the menu button proper's label. We don't seem to have a function to unescape markup entities **and** strip Pango markup at the moment, so this is left unsolved in the PR for the time being.
* The Combat menu button has overlong descriptions resulting in a messy dropdown display as seen in the third screenshot above. I'm not entirely sure what to do about this without losing the intended conveyance of information — tooltips perhaps?
* The dialog has layout issues on low resolution (< 1024 px horizontal) modes and other situations where overlong text crops up.on the game settings row at the bottom. I want to fix this, **ideally before merging**.
* There is no way to set modification options from the dialog itself. If this is wanted it could be worked on at a later date, preferably with the help of someone who knows better how modifications work (such as @gfgtdf).
* The idea suggested in issue #4543 of adding a separate tab for tutorials is not implemented yet. There does not seem to be a use case for it yet either since the tutorial only has two scenarios at the moment. I also largely ignore if it's possible to instruct the game to skip to a scenario other than `first_scenario`. I would greatly appreciate some input on this particular point for this PR and others.
* Mainline campaigns have no custom backgrounds defined yet. Below is attached a patch I used to test the feature only, with no real thematic coherence intended ­— save perhaps for HttT, where my best option was the antiquated 1.0 story art used for the Battle for Wesnoth scenario.
* Campaign images (often portraits) are currently allowed to jump around even when the text doesn't take enough space for them to need to do so. Perhaps anchor them to the bottom of the right pane? (Note that @Vultraz seems to be of the opinion that we should do away with the images themselves. I do not agree with this.)
* Campaign descriptions may suffer from text contrast issues when a sufficiently bright background is selected. Currently this is partly alleviated by using a dark translucent overlay on the background (see the `campaign_menu_bg` definition for `panel`). I don't feel we can do anything about this short of adding shadows or (preferably) outlines to the text.
* Long campaign descriptions predictably cause a vertical scrollbar to appear. `multi_page` remembers the scrollbar position for each page and I can't quite decide if this is desirable behaviour or not.

### Original Experimental PRNG option patch

See commit 3f06813fcf597b63271e26c9a3a9aa1480020cca and the [forum thread](https://forums.wesnoth.org/viewtopic.php?f=6&t=48160), both by Dave.

### Background image test patch

See https://gist.github.com/shikadiqueen/e66d433d9ee0d6dd4915f986d6658c38.

### DIFF
--- a/data/advanced_preferences.cfg
+++ b/data/advanced_preferences.cfg
@@ -167,15 +167,5 @@
     default=no
 [/advanced_preference]
 
-[advanced_preference]
-    field=use_prng
-    name= _ "Use experimental PRNG combat"
-    description= _ "Enables more determinstic chance-to-hit calculations. This is an experimental feature designed to bring the observed hit/miss rate more in line with the displayed percentages.
-
-Note: this option only affects singleplayer, and the ‘Save random seed’ option must also be enabled when creating a game for this to work."
-    type=boolean
-    default=no
-[/advanced_preference]
-
 #ifdef __UNUSED__
 #endif
--- a/data/gui/widget/panel_campaign_menu.cfg
+++ b/data/gui/widget/panel_campaign_menu.cfg
@@ -1,0 +1,79 @@
+#textdomain wesnoth-lib
+###
+### Panel with a configurable background (via label text) for the Campaigns menu.
+###
+
+#define _GUI_RESOLUTION RESOLUTION BORDER
+	[resolution]
+
+		{RESOLUTION}
+
+		left_border = {BORDER}
+		right_border = {BORDER}
+		top_border = {BORDER}
+		bottom_border = {BORDER}
+
+		[background]
+
+			[draw]
+
+				[image]
+					name = "(if(text = '', 'story/landscape-battlefield_nohumans.jpg', text) .. '~O(0.25)')"
+					x = "(max(0, (width - (width / (image_original_width/image_original_height))) / 2))"
+					w = "(width / (image_original_width/image_original_height))"
+					h = "(height * (image_original_width/image_original_height))"
+				[/image]
+
+				[rectangle]
+					x = 0
+					y = 0
+					w = "(width)"
+					h = "(height)"
+
+					border_thickness = 3
+					border_color = "16, 22, 35, 255"
+				[/rectangle]
+
+				[rectangle]
+					x = 1
+					y = 1
+					w = "(width - 2)"
+					h = "(height - 2)"
+
+					border_thickness = 1
+					border_color = {GUI__BORDER_COLOR_DARK}
+				[/rectangle]
+
+				[image]
+					x = 3
+					y = 3
+					w = "(width - 6)"
+					h = "(height - 6)"
+					name = "dialogs/translucent54-background.png"
+				[/image]
+
+			[/draw]
+
+		[/background]
+
+		[foreground]
+
+			[draw]
+
+			[/draw]
+
+		[/foreground]
+
+	[/resolution]
+#enddef
+
+[panel_definition]
+
+	id = "campaign_menu_bg"
+	description = "Definition of the dynamic background panel for the Campaigns menu"
+
+	{_GUI_RESOLUTION () 3}
+
+[/panel_definition]
+
+#undef _GUI_RESOLUTION
--- a/data/gui/widget/panel_campaign_menu.cfg
+++ b/data/gui/widget/panel_campaign_menu.cfg
@@ -32,6 +32,8 @@
 
 					border_thickness = 3
 					border_color = "16, 22, 35, 255"
+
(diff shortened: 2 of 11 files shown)


nemaara (approved):
Looks pretty nice.

Vultraz:
One think that immediately stand out to me is that Modifications section. It should be aligned with the left window edge.

soliton-:
Perhaps the combat descriptions can be moved to a linked help page instead.

shikadiqueen:
> Perhaps the combat descriptions can be moved to a linked help page instead.

I thought about this, but that would mean adding a new help page and hoping for people to click on a somewhat unrelated element in the dialog (a new Help button) to figure out what the option does.

On the other hand, it turns out that tooltips do work with dropdown menu items as one would expect them to do without any issues:

Screenshot_20201215_131214

Pentarctagon (approved):

Pentarctagon:
@nemaara Once this is merged, can you take a look as far as adding backgrounds for the mainline campaigns?

Vultraz (approved):

shikadiqueen:
After consulting with @Pentarctagon and @Vultraz, we've decided to merge as-is. The PR is still here for @gfgtdf's reference if he becomes able to review the engine changes at a later time.

AI0867:
I dislike the description of the Biased RNG. It seems to imply that the regular RNG doesn't obey the true probabilities, even though it doesn't actually say that.

I'm not sure how to word it better though.

CelticMinstrel:
Now that this has been merged, I have two comments.

* First, I agree with @AI0867 on the description of the "Biased RNG". I'm not 100% how it works, but I did glance at the original commit and it looks similar to DotA's system where the chance to hit starts _below_ the target chance but increases with every miss, resetting once there's a hit. I don't think "biased" is at all an accurate description of this - biased gives the impression that it's altering the probabilities, but if it _is_ similar to what DotA does, then it's not altering the probabilities themselves, only the distribution of successes and failures. Perhaps "Smoothed RNG" or something similar? I'm not happy with that either but it seems less misleading than "Biased RNG".
* It seems like there's no longer any way to use the experimental PRNG without enabling Save Random Seed, is that correct? I'd prefer them to be two separate options. They have nothing to do with each other, after all.

PR #5360, opened 2020-12-15
Unit: removed use of boost::ptr_vector

Vultraz:
There really seems no reason to use this. Also removes a stray use of `boost::copy` in `game_lua_kernel`.

### DIFF
--- a/src/scripting/game_lua_kernel.cpp
+++ b/src/scripting/game_lua_kernel.cpp
@@ -111,8 +111,6 @@
 #include "deprecation.hpp"
 
 #include <functional>               // for bind_t, bind
-#include <boost/range/algorithm/copy.hpp>    // boost::copy
-#include <boost/range/adaptors.hpp>     // boost::adaptors::filtered
 #include <array>
 #include <cassert>                      // for assert
 #include <cstring>                      // for strcmp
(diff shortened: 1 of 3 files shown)


soliton-:
You can see the commit that introduced it for the reason: https://github.com/wesnoth/wesnoth/commit/d3615be

CelticMinstrel:
Even after reading that commit message, I still don't understand why it was done this way. Did gfgtdf think it would be somehow an optimization? I don't really think it is though?

The commit message also confirmed that it's pre-C++11, based on its reference to BOOST_FOREACH.

I honestly can't see any way in which using ptr_vector is better than this change - with ptr_vector, you have the same number of copies or moves, but you _also_ have allocations that shouldn't really be needed. That's also not the intended use of ptr_vector as far as I'm aware - config is not a polymorphic type.

AI0867:
> I honestly can't see any way in which using ptr_vector is better than this change - with ptr_vector, you have the same number of copies or moves, but you _also_ have allocations that shouldn't really be needed. That's also not the intended use of ptr_vector as far as I'm aware - config is not a polymorphic type.

You have the same number, but a pointer is a lot smaller than the object itself, which makes reordering things faster. I have no idea if the contents of this vector are reshuffled a lot, but that could be a reason.

PR #5364, opened 2020-12-15
Remove the editor option to assign a recruit list to a unit.

Pentarctagon:
The UI implementation of this is bad enough that I consider its presence to actively a detriment to anyone trying to use the editor.


CelticMinstrel:
It's a useful feature though, isn't it? Assuming it sets the `extra_recruit` key… but if the UI is terrible it might be better removed than left, at least until someone has time to make a better UI…

PR #5365, opened 2020-12-16
Remove the ability to change a unit's ID from the editor.

Pentarctagon:
The reasoning being:
* Currently the editor will silently not save changes to the ID that match what the editor would assign - for example changing Mage-1 to Mage-99.
* If that were fixed, it's also quite trivial to have the editor create two units with identical IDs - place a Mage (assigned Mage-1), rename to Mage-2, place another Mage -> there are now two units with the ID Mage-2.
* More generally, given that many users of the scenario editor aren't at all familiar with WML and that the wiki says that having multiple units with the same ID "may produce unpredictable results in many cases", I feel like this isn't functionality that *should* be exposed in the editor.


CelticMinstrel:
I think having a human-readable ID is pretty important for events, though - how would you expect people to reference a placed unit if they can't change the ID? I mean, sure, they can use the generated ID, but that doesn't make for very readable code.

The editor could easily both give an error if you try to assign an ID that's used by another unit and skip a sequence ID if it finds it to be already used by an existing unit, which would fix the second two issues (though it would be more complicated).

Pentarctagon:
I don't think people are likely to be using the editor in that way, given its limited ability to place units.  It might be useful to open a forum poll asking how people are using the editor though.

cooljeanius:
> I don't think people are likely to be using the editor in that way, given its limited ability to place units. It might be useful to open a forum poll asking how people are using the editor though.

Link to said poll is here, for reference: https://forums.wesnoth.org/viewtopic.php?f=6&t=53724

stevecotton:
Where does the logic for "silently not saving" the ID happen?

While I agree with @CelticMinstrel that this GUI could be a nice feature, I think it's better to remove it now; if someone wants to write the error-checking then it would be trivial for them to re-add the GUI too.

Pentarctagon:
> Where does the logic for "silently not saving" the ID happen?

I don't know where it is, just that if I change a unit ID from something like the example, the unit's ID is simply not present in the scenario cfg that's written to disk.

stevecotton:
Not saving such IDs was added by 95540af9bf0ad0c67342c139ab3381ae954d5505, with the explanation "these default ids might cause bugs later since the scenario might generate units with identical ids".

How about changing the dialog box so that the OK button and default action is disabled when the ID contains a hyphen?

CelticMinstrel:
I think 95540af9bf0ad0c67342c139ab3381ae954d5505 is actually wrong and should just be reverted. Instead, the scenario editor should check whenever it generates a new ID that there isn't already a unit with that ID on the map.

In fact, does 95540af9bf0ad0c67342c139ab3381ae954d5505 even actually fix the bug it purports to fix? If the issue is that the scenario editor generates units in a sequence, then what happens when you load an existing scenario? Maybe it alters its ID sequence number based on the units it finds? But if it just checked at unit placement time to verify that the ID doesn't exist, I think that would be a lot simpler.

Issue #5359, opened 2020-12-15
Boost: 2 deprecated headers


sevu:
Just saw that now when redirecting stdout but not stderr:
```
In Datei, eingebunden von /usr/include/boost/config/header_deprecated.hpp:18,
von /usr/include/boost/function_output_iterator.hpp:12,
von src/units/unit.cpp:57:
/usr/include/boost/function_output_iterator.hpp:14:1: Anmerkung: »#pragma message: This header is deprecated. Use <boost/iterator/function_output_iterator.hpp> instead.«
14 | BOOST_HEADER_DEPRECATED("<boost/iterator/function_output_iterator.hpp>")
| ^~~~~~~~~~~~~~~~~~~~~~~
In Datei, eingebunden von /usr/include/boost/smart_ptr/detail/sp_thread_sleep.hpp:22,
von /usr/include/boost/smart_ptr/detail/yield_k.hpp:23,
von /usr/include/boost/smart_ptr/detail/spinlock_gcc_atomic.hpp:14,
von /usr/include/boost/smart_ptr/detail/spinlock.hpp:42,
von /usr/include/boost/smart_ptr/detail/spinlock_pool.hpp:25,
von /usr/include/boost/smart_ptr/shared_ptr.hpp:29,
von /usr/include/boost/shared_ptr.hpp:17,
von /usr/include/boost/iostreams/device/file_descriptor.hpp:28,
von src/filesystem.cpp:36:
/usr/include/boost/system/windows_error.hpp:16:3: Anmerkung: »#pragma message: This header is deprecated and is slated for removal. If you want it retained, please open an issue in github.com/boostorg/system.«
16 |   BOOST_PRAGMA_MESSAGE("This header is deprecated and is slated for removal."
|   ^~~~~~~~~~~~~~~~~~~~
```

Boost version is 1.75.0 / build is from master.

Vultraz:
Both headers have been removed.

PR #5367, opened 2020-12-16
FS: removed use of Windows-specific error code

Vultraz:
The OS-specific error code headers were deprecated in Boost 1.75 (see #5359). This code was first added back in 2013
(see 037e1c12997fe92a2aed36c8b2a05ff31696e8de) but testing now on Windows 10 using Boost 1.73 at least shows that
`no_such_file_or_directory` is returned for both missing files and paths. Seems unnecessary to keep this case around
especially as the header is deprecated going forward.

### DIFF
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -35,7 +35,6 @@
 #include <boost/filesystem/fstream.hpp>
 #include <boost/iostreams/device/file_descriptor.hpp>
 #include <boost/iostreams/stream.hpp>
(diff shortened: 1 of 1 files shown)


Wedge009:
Makes sense to me, but it depends what the project has decided to support in terms of Boost versions. What's the current minimum expected Boost for Wesnoth?

Pentarctagon:
Minimum is 1.65 currently since that's what Ubuntu 18.04 has available, but in practice for Windows it's whatever vcpkg.

PR #993, opened 2017-04-21
[WIP] Improvements to the Lua map generator

CelticMinstrel:

PR #5370, opened 2020-12-18
cave_map_generator.lua: fix transform flip

virtualghetto:
Reported on the forum.
https://forums.wesnoth.org/viewtopic.php?f=4&t=53679


CelticMinstrel:
Hmm… I'm not quite sure what this is doing, and those removes and inserts look rather weird. Can you perhaps explain? I'm guessing it has something to do with the fact that a map is a hex grid, not a rectilinear grid, but…

virtualghetto:
The forum post includes a demo for testing this patch.
The attached images show the map[] fix, but they don't show the map.locations fix.
You will notice that the units are not placed on their correct keep positions in prestart.

In the Lua cave generator, the transform attribute is supposed to flip the generated map along the x or y axis. Currently it doesn't work, this patch is supposed to fix that.

When the map is generated there are two parts, the map[] array will have the generated terrains like Uu, Xu, Kud. And the map.locations will have the item_location ids like '1' or 'sceptre'.

The transform is broken on the map[] level because it is going through the array line as a whole and not stopping mid-point.
For example, it swaps map[0] and map[999], then map[1] and map[998] .... and when the index i becomes greater than j it doesn't stop. So it will swap what was already swapped once again, and returns everything to its original position.
i.e., it will swap map[998] and map[1], followed by map[999] and map[0] again.
That's what the break is for in the patch.

The second part of the fix is swapping the map.locations.
Unfortunately trying the same method on map.locations did not work because some indices in map.locations will be nil and have no value. A straight forward swap based on indices like map.locations[] did not work. It will quit wesnoth.

So, the patch is checking if we have map.locations defined, then it is retrieving both old and new locations.
The i and j will either have some value like '1', '2', 'sceptre' or they will be nil.
If it is nil, then don't try to remove it from the old position or insert it into the new position. This also will quit wesnoth.
If it is not nil, then remove it from the old position and do an insert of that value to the new position.

virtualghetto:
Just in case, I changed the break check to be over the x and y values instead of the index values.

CelticMinstrel:
Okay, I understand the issue now and how this addresses it, and it all makes sense. I still have one concern, though. This part looks possibly wrong:

```lua
if x > map.w - x - 1 then
	break
end
map[i], map[j] = map[j], map[i]
if map.locations then
```

Doesn't this mean that locations on the right half of the map won't be fixed up? Perhaps the logic should go more like this:

```lua
if x < map.w - x then
	map[i], map[j] = map[j], map[i]
end
if map.locations then
```

The same concern also applies to the vertical flip logic.

virtualghetto:
> Doesn't this mean that locations on the right half of the map won't be fixed up? Perhaps the logic should go more like this:
> 
> ```lua
> if x < map.w - x then
> 	map[i], map[j] = map[j], map[i]
> end
> if map.locations then
> ```

The right half does get fixed up. It is done simultaneously with the left half.
The point of the break is to not swap the right half two times over.
We start from the extreme edges and move inwards from both left and right directions.
`i` takes care of the left half, while `j` takes care of the right half.
Once the inflection happens, we stop by breaking.

The break applies to both map[] and map.locations. So if we are to go with your suggestion, the map.locations would have to be moved inside that `if` as well.
> ```lua
> if x < map.w - x then
> 	map[i], map[j] = map[j], map[i]
>      if map.locations then
>         --do map.locations swapping
>      end
> end
> ```

I need to think about how to use loc_set:iter for swapping or if it is even applicable here.
I will also test that `-1` in `if x > map.w - x - 1 then` to see if the right most edge is getting ignored or if `map.w -1` is indeed the right most valid location.

CelticMinstrel:
> So if we are to go with your suggestion, the map.locations would have to be moved inside that `if` as well.

Yeah, there's definitely no point in doing it that way if both need to be moved.

I took another look, and I can see how the locations are swapped now. If it's to be done this way though, I think it still needs two changes.

1. That `if` and `break` is pretty silly - you can just adjust the upper bound on the `for`!
2. The local i and j in the location handling is confusing, I think. Maybe try calling those variables left / right / top / bottom? Or maybe you can come up with better names.

If done with `loc_set:iter` then point 1 above still applies, but in this case the logic would be swapping both x and y for one location at a time, so point 2 becomes irrelevant.

Does anyone else have opinions on iterating the location set vs dealing with locations at the same time as terrain?
@Pentarctagon @Vultraz @gfgtdf

virtualghetto:
> I will also test that `-1` in `if x > map.w - x - 1 then` to see if the right most edge is getting ignored or if `map.w -1` is indeed the right most valid location.

I just test this. map.w-1 is indeed the right most edge, removing the -1 exited wesnoth.

There is another problem with flip_x and flip_y that everyone should be aware of.
On their own they shouldn't be considered reliable because of the asymmetrical nature of the wesnoth board. They could result in a broken up or impassable passage.
flip_xy on the other hand seems to be more reliable, but still problematic if near the edge of the map.

Note, the mainline HttT scenario 17 does a flip_x transform. Maybe that needs to be changed to flip_xy, or not do a transform at all.

No transform. The keep is placed on the bottom-right edge 22,22
no_flip

flip_x
flip_x

flip_y
flip_y

flip_xy
flip_xy

virtualghetto:
I think it would be best if the transform was performed on the original `[chamber]` x,y values before the map is generated rather than on the end result. 
This would ensure the generated keeps and passages don't get deformed, and we don't have to worry about messing with map.locations.

The translation would go above [Line 61 in cave_map_generator.lua](https://github.com/wesnoth/wesnoth/blob/master/data/lua/cave_map_generator.lua#L61) with something that in effect does the following depending on the transform_chance and transform type.

```lua
x = map.w - x - 1 -- for flip_x
y = map.h - y - 1 -- for flip_y
```

CelticMinstrel:
I believe the flip can still be achieved after the map is generated… the math is probably just more complicated due to the hex representation, and I'm not sure what it would actually look like.

Does your latest version avoid the issues from the screenshots while still flipping the map correctly?

virtualghetto:
I don't know what the math would look like either.

These screenshots are from the attached demo with the cave_map_generator.lua changes applied. The castle and passages look fine, and each leader is at the correct starting position.

[_main.cfg.txt](https://github.com/wesnoth/wesnoth/files/5731534/_main.cfg.txt)

No transform, keep 1 in at 22,22
no_flip

flip_x
flip_x

flip_y
flip_y

flip_xy
flip_xy

virtualghetto:
Should I make a cleaner pull request for this one? The change is no longer in mapgen_helper.lua and the commits show merges master.

CelticMinstrel:
If you want to force-push that's fine, but we can also just use "Squash and Merge" here on GitHub.

CelticMinstrel:
So, I just took another look at this patch, and I think it's valid as-is, but I still have one concern. I'd like to be able to easily add additional transform types, such as "rotate clockwise". I think it could be done by storing just the name of the transform instead of two flip booleans and calling a function in the mapgen_helper module, similar to how the old method worked. Are you interested in implementing that? If not, I can cherry-pick this and add that functionality on top of it.

virtualghetto:
I'm fine with you cherry-picking or fixing it in any way you see fit.
You don't have to merge this.
Thanks for offering, but I won't try to implement the extra functionality. (I don't know how to, tbh)

Pentarctagon:
@CelticMinstrel was this cherry-picked/should this be closed?

CelticMinstrel:
I'll probably roll this into #993 too.

Pentarctagon:
@CelticMinstrel same question as before.

CelticMinstrel:
It looks like this fix _has_ been merged into #993 in a different form. It could be either merged or closed. If merged it would conflict with #993 but I think resolving the conflict would be to just take the #993 version.

Pentarctagon:
Alright, closing then.

PR #5371, opened 2020-12-18
SotBE S09: add user_team_name

celerini:


nemaara (changes requested):
Given that you're playing orcs too, it's probably better to name them something else.

nemaara (approved):

PR #5378, opened 2020-12-20
Include architecture info into wesnoth's informations

hrubymar10:
## Description
Because of universal wesnoth 1.15.8 package I wanted to have easy way to determine which arch user use. So I added this exact patch from commit cbe72f5 into official 1.15.8 package. Now it is good time to discuss it and maybe change it for all platforms.

## Ping
@shikadiqueen
@Pentarctagon

### DIFF
--- a/src/desktop/apple_version.mm
+++ b/src/desktop/apple_version.mm
@@ -55,6 +55,12 @@ namespace apple {
 		version_string += [[version_array objectAtIndex:1] UTF8String];
 		version_string += " (";
 		version_string += [[version_array objectAtIndex:3] UTF8String];
+		
+#if defined(__aarch64__)
+		version_string += " arm64";
+#elif defined(__x86_64__)
+		version_string += " x86_64";
+#endif
 
 		return version_string;
 	}


CelticMinstrel:
If we're going to do this, I wonder if it should detect less common architectures too? I know Intel/ARM are basically the only options for Mac and Windows, but I think Linux can install on some other architectures, though I don't know how common that is… still, even if we don't detect them, perhaps there should at least be a fallback "unknown architecture" case.

Pentarctagon:
This is Apple only anyway I believe.

shikadiqueen:
I just pushed two changes that let us display the *runtime* architecture on Windows (commit 15356f1db67b705686462bb077d0a39bc1f78791) and POSIX (9b1ea3078da2b2d874774bc631648d90a0f78433) platforms like so:

image

I would expect that for our official use case here with macOS, the runtime will always match the build architecture (except in 1.14.x which I understand can't be built for Apple's ARM yet), but you can probably confirm by rebasing the patch on top of current master -- I'd expect to see it say "arm64 arm64" or so.

It may be preferable to add another entry for the build architecture instead of embedding that into the OS version string.

hrubymar10:
@shikadiqueen 's works for both ARM64 (where it shows arm64) and Rosetta2 (where it shows x86_64). So I'll close this now as it was replaced by better solution.

Closing

PR #5379, opened 2020-12-20
Fix segfaults and a few other issues in wesnoth.find_path

CelticMinstrel:
This is what I have so far on the matter... I _think_ it addresses every possible case, but I'm quite not sure and I haven't yet tested it.

This includes an API change, deprecating the use of `viewing_side=0` to specify that fog, shroud, and hiding are ignored. Instead, you now use `see_all=true` for this purpose. The reason for this can be found in the issue discussion.

### DIFF
--- a/src/scripting/game_lua_kernel.cpp
+++ b/src/scripting/game_lua_kernel.cpp
@@ -1697,11 +1697,13 @@ int game_lua_kernel::intf_find_path(lua_State *L)
 	int arg = 1;
 	map_location src, dst;
 	const unit* u = nullptr;
+	int viewing_side = 0;
 
 	if (lua_isuserdata(L, arg))
 	{
 		u = &luaW_checkunit(L, arg);
 		src = u->get_location();
+		viewing_side = u->side();
 		++arg;
 	}
 	else
@@ -1710,6 +1712,7 @@ int game_lua_kernel::intf_find_path(lua_State *L)
 		unit_map::const_unit_iterator ui = units().find(src);
 		if (ui.valid()) {
 			u = ui.get_shared_ptr().get();
+			viewing_side = u->side();
 		}
 		++arg;
 	}
@@ -1723,7 +1726,6 @@ int game_lua_kernel::intf_find_path(lua_State *L)
 		return luaL_argerror(L, arg - 2, "invalid location");
 
 	const gamemap &map = board().map();
-	int viewing_side = 0;
 	bool ignore_units = false, see_all = false, ignore_teleport = false;
 	double stop_at = 10000;
 	std::unique_ptr<pathfind::cost_calculator> calc;
@@ -1731,10 +1733,10 @@ int game_lua_kernel::intf_find_path(lua_State *L)
 	if (lua_istable(L, arg))
 	{
 		ignore_units = luaW_table_get_def<bool>(L, arg, "ignore_units", false);
-
+		see_all = luaW_table_get_def<bool>(L, arg, "see_all", false);
 		ignore_teleport = luaW_table_get_def<bool>(L, arg, "ignore_teleport", false);
 
-		stop_at = luaW_table_get_def<double>(L, arg, "stop_at", stop_at);
+		stop_at = luaW_table_get_def<double>(L, arg, "max_cost", luaW_table_get_def<double>(L, arg, "stop_at", stop_at));
 
 
 		lua_pushstring(L, "viewing_side");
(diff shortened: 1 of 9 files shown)


src/scripting/game_lua_kernel.cpp
```diff
@@ -1759,20 +1766,22 @@ int game_lua_kernel::intf_find_path(lua_State *L)
...
 	}
 
 	if (!calc) {
-		if (!u) return luaL_argerror(L, 1, "unit not found");
+		if(!u) {
+			return luaL_argerror(L, 1, "unit not found OR custom cost function not provided");
+		}
 
-		calc.reset(new pathfind::shortest_path_calculator(*u, viewing_team,
+		calc.reset(new pathfind::shortest_path_calculator(*u, board().get_team(viewing_side),
```
Pentarctagon, line 1784: This would throw if `viewing_side` were 0, wouldn't it?

src/scripting/game_lua_kernel.cpp
```diff
@@ -1759,20 +1766,22 @@ int game_lua_kernel::intf_find_path(lua_State *L)
...
+			return luaL_error(L, "wesnoth.find_path: ignore_teleport=false requires a valid viewing_side");
+		} else {
+			teleport_locations = pathfind::get_teleport_locations(*u, board().get_team(viewing_side), see_all, ignore_units);
+		}
 	}
 
 	if (!calc) {
-		if (!u) return luaL_argerror(L, 1, "unit not found");
+		if(!u) {
+			return luaL_argerror(L, 1, "unit not found OR custom cost function not provided");
```
Pentarctagon, line 1781: This should be "AND" rather than "OR".

src/scripting/game_lua_kernel.cpp
```diff
@@ -1723,26 +1726,30 @@ int game_lua_kernel::intf_find_path(lua_State *L)
...
 
 	if (lua_istable(L, arg))
 	{
 		ignore_units = luaW_table_get_def<bool>(L, arg, "ignore_units", false);
-
+		see_all = luaW_table_get_def<bool>(L, arg, "see_all", false);
 		ignore_teleport = luaW_table_get_def<bool>(L, arg, "ignore_teleport", false);
 
-		stop_at = luaW_table_get_def<double>(L, arg, "stop_at", stop_at);
+		stop_at = luaW_table_get_def<double>(L, arg, "max_cost", luaW_table_get_def<double>(L, arg, "stop_at", stop_at));
```
CelticMinstrel, line 1739: After looking at the forum thread, I realized `stop_at` was an error introduced in eee9214e9a42b3681a10a42471399dc98a363521, so I think I'll drop support for it if there's no objections.

src/scripting/game_lua_kernel.cpp
CelticMinstrel: Yes, but it only reaches here if there's a valid unit, in which case `viewing_side` cannot be 0.

It's set in two places - if there's a unit, it's set to the side of that unit. Then if `viewing_side` is explicitly set in options, it will be used instead if it's valid.

src/scripting/game_lua_kernel.cpp
CelticMinstrel: My reasoning for making it "OR" is that providing either is sufficient to make the error go away. It's true that both are not provided when this message is shown, but both are not required.

src/scripting/game_lua_kernel.cpp
Pentarctagon: Alright, I see it now.  I was thinking that it could get here in the case of a provided location that doesn't have a unit on it, but in that case `!u` would be true and it would show the error instead.

Pentarctagon:
It looks like the CI deprecation failing due to the deprecation message being triggered.

CelticMinstrel:
This quashes most of the deprecated uses, but I'm not sure how to deal with the stuff in ai_helper. The issue is that the options table could be passed through a number of layers of calls before finally reaching `wesnoth.find_path`, and I'm not quite sure where it needs to be updated to use `ignore_visibility=true` in place of `viewing_side=0`.

Broadly speaking, there's two options - make the substitution just before calling `wesnoth.find_path` (so callers to ai_helper functions still use `viewing_side=0`); or identify all uses that need to be updated elsewhere and update them (so ai_helper functions no longer treat `viewing_side=0` as a special case).

It looks like some of the changes in the last commit already updated some of those call sites, so perhaps the second method is better? Any thoughts, @mattsc?

mattsc:
Barring mistakes, 62d24b2 should take care of what needs to be done for the Lua AIs.  It's infeasible to test all possible combinations of uses of the ai_helper functions, but I have done a lot of it, so most of this should work.  That should include uses of `wesnoth.find_reach`, although `ignore_visibility` still needs to be implemented for that function, so I have not been able to fully test those yet.

mattsc:
I've tested `ignore_visibility` with `wesnoth.find_reach()` and modified the last ai_helper function (`next_hop`) that wasn't done yet.  All seems to be working as expected now, as far as I can tell.

mattsc (approved):

Pentarctagon:
It looks like there's still a deprecation message that needs to be fixed.

CelticMinstrel:
You mean there's something using a deprecated API, or I need to add a deprecation message somewhere?

Pentarctagon:
Many WML unit tests are failing because of the message:

> error deprecation: wesnoth.find_reach with viewing_side=0 (or an invalid side) has been deprecated and will be removed in version 1.17.0.

CelticMinstrel:
It looks like all such errors are coming from use of `ai_helper` functions - @mattsc?

I did spot one potential use of the deprecated form that wasn't, but it shouldn't be the cause of any of the unit test failures, and furthermore I'm not sure how best to handle it. Basically, `[store_reachable_locations]` will result in a deprecated call if the WML sets `viewing_side=0`. The same is not true of `[find_path]` (it doesn't seem to expose a viewing side option).

PR #5380, opened 2020-12-20
Add CI jobs for macOS ARM configurations.

Pentarctagon:


hrubymar10:
This PR is wrong. If you don't disable x86_64 in arm64 task and arm64 in x86_64 then it will produce both slices by default. https://github.com/wesnoth/wesnoth/runs/1586061956#step:3:5094

I think that at least DEBUG build which produces artefacts should be universal, so what about two separate release builds and one universal DEBUG build?

Pentarctagon:
It would depend on what the goal is really.  If you're going to be producing universal builds for both arches then that's what the CI should be doing, or if you're not then the CI shouldn't be either.

hrubymar10:
I will produce universal builds for both archs.

So we can have two separate Release builds (to paralelize them) but one universal Debug build.

Pentarctagon:
The point of CI though is exactly to check that the builds aren't broken in the particular environment.  Not too long enough there was an issue where the Release builds were succeeding but the Debug builds were failing with a compiler error, for example.  If/when Github Actions decides to cut back on what they offer for free with their CI, then that would be the time to decide what we can do without.  For now though, we're allowed to have 20 jobs running in parallel, so whether there's 14 jobs or 15 jobs is not a huge difference.

hrubymar10:
Point is that if build will fail just for example for arm64 and will work for x86_64 it may be difficult to understand what error is for somebody who don’t understand xcodebuild. Also as author of failed commit will probably have x86_64 machine it will work for him what will make it so much more confusing. But, that is my opinion. I can also be wrong with that at this may never happen.

Pentarctagon:
That's actually the kind of thing the CI is supposed to reveal though - if commit X succeeds on x86_64 but fails on ARM64, then that means there's something that needs to be fixed for it to work on ARM64.  A similar example would be vultraz's recent fights with GCC on Linux where the macOS, Windows, and some Linux jobs succeeded, but other Linux jobs (usually the Steam Runtime one) failed.

PR #5386, opened 2020-12-23
Send the player forum ids to the client on the player list.

Pentarctagon:


Vultraz (changes requested):
Could we have both the UID and the forum profile URL? There might be other uses for the UID in and of itself.

Pentarctagon:
@Vultraz updated.

Vultraz (approved):

soliton-:
I would rather suggest to send the profile URL template on login. Sending it in the user list seems like excessive duplication.

Pentarctagon:
So it would be added to `login_response_` you mean?

soliton-:
Yeah, `[join_lobby]` could get  attributes `forum_user_id` and `forum_profile_url_template` or so containing `forums.w.o....<USER_ID>` or without the placeholder and you'd just append the ID to get the URL. Then the client can construct its own profile URL and those of others from the user list.

Pentarctagon:
While it will work either way for this, would it make sense to set up a URL shortener for the profile URLs?  So they could be `https://r.wesnoth.org/u######` instead of `https://forums.wesnoth.org/memberlist.php?mode=viewprofile&u=######`.

soliton-:
That redirect already exists. 😉

Pentarctagon:
Alright, updated.

src/server/wesnothd/player.cpp
```diff
@@ -34,6 +34,7 @@ wesnothd::player::player(const std::string& n, simple_wml::node& cfg,
 	cfg_.set_attr_dup("name", n.c_str());
 	cfg_.set_attr("registered", registered ? "yes" : "no");
 	cfg_.set_attr("moderator", moderator ? "yes" : "no");
+	cfg.set_attr_int("forum_id", id);
```
soliton-, line 37: Does this work correctly? You're setting on the function parameter and not the member.

Pentarctagon: It does - I tested it by printing out the data sent to the client and the Profile button now in the MP lobby correctly creates the link using the forum ID.  It's a node reference to data owned by `games_and_users_list_`, so both refer to the same thing, but still definitely a typo.

Issue #5383, opened 2020-12-22
HttT Sc. Crossroads,  Ambushers are visible and active


rmj555:
### Game and System Information
 - What version of the game are you running?1.15.8
 - Where did you download it from? (Steam, Mac App Store, our website, built from source, etc.)Sourceforge
 - If you're using a custom build, what compiler did you use, and what commit did you build at?
 - What OS are you running?Windows10

### Describe the bug
The ambushers should be invisible until next to them, but are visible and active.

### To Reproduce

1. First..
2. Then...
3. Lastly...

### Savefiles
If applicable, attach savefiles to help explain your problem. (To find savefiles, start Wesnoth, click on the `i` button at the bottom left corner of the title screen, and open the `Saved games` directory.)

### Additional context
Add any other context (compiler logs, console output, etc) about the problem here.

Issue #5385, opened 2020-12-23
Holy water doesn't work


rmj555:
1.15.8
Windows 10
Sourceforge

Holy water did make a melee weapon arcane in either HttT or TRoW

rmj555:
Paladins don't have arcane weapons.

newfrenchy83:
@rmj555 paladins have already a arcane sword, and fore me holy water work without 'fix' what i have added in my PR, and the unit who are already an arcane attack are exluded by the filter in macros/items.cfg line 271

CelticMinstrel:
I'm not sure what point either of you are trying to say about paladins, but let me state for the record what _should_ happen if a paladin tries to take holy water: it should say the paladin cannot use it.

sevu:
That _should_ already be the case.

AI0867:
A paladin also has a lance though. Arguably, holy water should be usable to make that weapon arcane too.

rmj555:
The problem, it seems, is not that a weapon does not become arcane, but that the weapon is not described as arcane.  I had two royal guards; one hit skeletons at 12-4, the other at 5-4.

soliton-:
Described where, @rmj555?

If you're talking about the help pages then those describe the unit type.

rmj555:
When attacking you get a choice of weapons. The melee weapon is not described as arcane.

soliton-:
Can anyone reproduce this or can you provide a save that shows the issue?

doofus-01:
Weapons aren't described as 'arcane' or 'pierce' in the attack dialog.  Specials get a string, but 'arcane' isn't a special.  Is that a bad GUI decision?  Maybe, I don't know.  But strictly speaking, if OP is talking about the attack dialog, it's not a bug.

rmj555:
When I put the cursor over the unit, the unit description on the right does not say the weapon is arcane.  It does so in 1.14.

CelticMinstrel:
> A paladin also has a lance though. Arguably, holy water should be usable to make that weapon arcane too.

Ah, does it? In that case, I guess the correct filter is something like

```
[has_weapon]
    range=melee
    [not]
        type=arcane
    [/not]
[/has_weapon]
```

CelticMinstrel:
[...]

PR #5387, opened 2020-12-23
ambusher don't workin httt 07_Crosroad.cfg

newfrenchy83:
fix https://github.com/wesnoth/wesnoth/issues/5383

### DIFF
--- a/data/campaigns/Heir_To_The_Throne/scenarios/07_Crossroads.cfg
+++ b/data/campaigns/Heir_To_The_Throne/scenarios/07_Crossroads.cfg
@@ -327,6 +327,7 @@
 
                     [object]
                         id=one_time_ambush
+                        take_only_once=no
 
                         [filter]
                             x,y=$ambush_location.x,$ambush_location.y
--- a/data/core/macros/items.cfg
+++ b/data/core/macros/items.cfg
@@ -270,9 +270,7 @@
         side=1
         [has_attack]
             range=melee
-            [not]
-                type=arcane
-            [/not]
+            type=blade,pierce,impact,fire,cold
         [/has_attack]
     ) items/holy-water.png
     _"Sprinkling this water on melee weapons grants them the <i>arcane</i> damage type until the end of the current scenario. Should $unit.name use it?"
@@ -280,7 +278,6 @@
     _"holy water^Leave it"
     _"$unit.name has no need for this Holy Water! Let another take it." (
         [object]
-            id={ID}
             name= _ "Holy Water"
             image=items/holy-water.png
             duration=scenario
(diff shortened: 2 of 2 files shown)


CelticMinstrel:
…_how_ can it possibly be this simple!? Is there some other bug here?

newfrenchy83:
two bugs here, the holy water because no id for object( an and httt crossroad(same id and specify take_only_once=no for applied to all ambusher, for trident, bug is potential if two same items in same scenario. If another bugs, then it must review all work of @nemaara in the core and campaign

Pentarctagon:
As a side note: In the commit message I believe you need to say put fix or fixes in front of each issue being addressed.  So for example: fix #5385 and fix #5383.

Since in #5385 you can see it says you **linked** a pull request whereas #5383 says only that you mentioned the issue.

CelticMinstrel:
So… am I to understand that these items work if you have only one of them in the scenario, but don't work if you have both in the same scenario?

newfrenchy83:
> 
> 
> So… am I to understand that these items work if you have only one of them in the scenario, but don't work if you have both in the same scenario?

if these item work with same [object] yes because same [object] applied one time only by default, for same items work both, each object must have an different id, or, for be applied to many unit like in 
HttT Sc. Crossroads, specify take_only_once=no.

CelticMinstrel:
But according to the wiki, not specifying an ID should have the same effect as `take_only_once=no`, unless I misinterpreted it…

newfrenchy83:
> 
> 
> But according to the wiki, not specifying an ID should have the same effect as `take_only_once=no`, unless I misinterpreted it…

Maybe, but if an id is necessary like in 'Crossroad' then  `take_only_once=no must be specified, for holy water, i must look if holy water don't work in master version.

newfrenchy83:
in fact the https://github.com/wesnoth/wesnoth/issues/5385 issue seem concern only the paladins, and yes otherwise, holy water items work normally without id specified, i must resolve thsi problems specifically.

CelticMinstrel:
This change seems wrong too.

* "not type=arcane" strikes me as the correct logic
* Changing it as you did in this commit breaks any case where a campaign defines a custom damage type
* Disallowing a paladin to use the holy water is correct, since their lance is already arcane.

newfrenchy83:
No, the sword only is already arcane but spear is pierce, but is if you said what paladins musn't have item then i remove that.

newfrenchy83:
> 
> 
> This change seems wrong too.
> 
>     * "not type=arcane" strikes me as the correct logic
> 
>     * Changing it as you did in this commit breaks any case where a campaign defines a custom damage type
> 
>     * Disallowing a paladin to use the holy water is correct, since their lance is already arcane.

i reverted that, Crossroad ambusher is the only fixing here now.

newfrenchy83:
@CelticMinstrel i think what this PR is complete now.

sevu:
When I made that change, I probably tested the unit with the take_only_once case …

sevu:
Actually, no.
The difference to TSG is, that the object there is embedded into the unit. That works despite there being implicitly `take_only_once=yes` …

PR #5388, opened 2020-12-26
Revisions to bobbing animations

doofus-01:
I've never been happy with the heaving motions in the standing/bobbing animations of some of the sprites, for example the spearman and orc archer.  However, the heavy breathing might make sense if the unit was physically failing, so in this PR I've got the filter to be < 1/2 HP-max (just to pick a number).

In principle, a similar distinction/filter could be made for healthy units, like the saurians dancing (just as an example of an existing standing animation that isn't heaving).

Filtering standing animations like this would multiply the needed animations by 2-3X, but we were never going to get all standing sprites animated anyway, and this would make future contributions more conceptually clear (dancing-->subtle shifting-->heaving).

In this PR, I've initially added the Orc Grunt and Archer.  Before committing, I'd move other existing animations similarly, and add some new standing animations.  But only if the concept makes sense, or if it would help the concept make sense.


Pentarctagon:
Would this affect idle animations at all?

doofus-01:
It would prevent idle animations from firing if the unit is "wounded", but otherwise it wouldn't.

At this point in the PR, a spearman with >1/2 HP just stands there, and occasionally does the idle animation.  A spearman with <= 1/2 HP starts heaving and never does the idle animation.

1/2 HP is probably too high, but good for testing.

doofus-01:
Now the orcish grunt is how I see this going, at least for heaving animations - there's still room for non-wounded standing animation.

If there are no objections, I will finish reviewing which existing bobbing animations are wounded heaving vs which are something else (like the ghosts or saurians aren't heaving), then run wmlindent and merge this.

Pentarctagon:
Makes sense to me. I'd also say if it comes up that the regular idle animations are a higher priority since they'll be the more commonly seen animations of the two types.

PR #5389, opened 2020-12-26
allow 'traditional' special weapons to take priority over others

newfrenchy83:
Until then, if the abilities used as weapons can, thanks to overwrite_specials take priority over other abilities of the same type and even over traditional special weapons, the reverse could not be true, this commit should allow to rectify this imbalance in including overwrite_special in special 'traditional' weapons


CelticMinstrel:
Right now my only complaint about this is the key should be `overwrite_abilities` instead of `overwrite_specials`.

I also have a question. What happens when you have an ability with `overwrite_specials` and also a special with `overwrite_abilities`?

newfrenchy83:
@CelticMinstrel To simplify my explanation, I will keep the same attribute for both, if an ability used as a weapon and a special weapon of the same type both have overwrite_specials then the general rule of comparison applies for them, when value = is used for both, the higher value is used. On the other hand, if two different attributes are used, the priority definition between two special weapons of the same type but from different sources is no longer done. The easiest way for coding is to define an identical attribute for both sources.

newfrenchy83:
@CelticMinstrel don't forget what for my https://github.com/wesnoth/wesnoth/pull/5023 PR, i want extent special_id/type_active to abilities used like weapons, and the attribute has 'special' in his name, perhaps use 'overwrite=' will be more logic.

CelticMinstrel:
> The easiest way for coding is to define an identical attribute for both sources.

This does make sense, but on the other hand, I think it's confusing to someone writing these weapon specials, because when used in a weapon special, it's not overwriting other weapon specials, right?

> perhaps use 'overwrite=' will be more logic.

No, that just begs the question, "overwrites _what_"?

newfrenchy83:
> This does make sense, but on the other hand, I think it's confusing to someone writing these weapon specials, because when used in a weapon special, it's not overwriting other weapon specials, right?
I implemented overwrite_special inside get_specials for what if two weapon_specials of same type affect one same attack, the overwrite_specials give priority to one, like overwrite_special give priority to abilities iver other abilities of same type and specials

> 
> No, that just begs the question, "overwrites _what_"?

then specials in 'overwrite_specials' must be read in the largest signification: traditionnal special and abilities used like weapon specials, or 'overwrite_others' but 'others what?' you will said.

CelticMinstrel:
> what if two weapon_specials of same type affect one same attack

I think this'll be a fairly unlikely edge case… and `overwrite_specials` seems odd even for that meaning.

newfrenchy83:
wi

> 
> 
> > what if two weapon_specials of same type affect one same attack
> 
> I think this'll be a fairly unlikely edge case… and `overwrite_specials` seems odd even for that meaning.

with one specials with apply_to=self and another owned byt opponent with apply_to=opponent, what the last used overwrite_specials attribute for superced value of first, i can be more sense and for 'overwrite_others' instead of 'overwrite_specials', what you advice?

newfrenchy83:
@CelticMinstrel i renamed overwrite_specials to overwrite_others for use same attribute for abilities and specials

PR #5391, opened 2020-12-28
Remove python script converting custom doc syntax into MediaWiki pages.

Pentarctagon:
WIP: either delete the existing doc comments or convert them to doxygen.

---

The GUI2 wiki pages are currently generated automatically by the `wiki_grabber.py` script.  However they have not been updated in quite a while and based on previous discussion are of questionable value anymore in the first place.  Remaining TODO:

- [ ] Delete the wiki pages that are part of https://wiki.wesnoth.org/Category:GUI_WML_Reference (last update seems to have been a bit over 5 years ago).
- [ ] Convert the existing doc comments to doxygen.
- [ ] Fix all the warnings currently shown when running doxygen.
- [ ] Add a CI check to cause a failure when warnings reappear in the future.


Pentarctagon:
@Vultraz in particular given you're the UI maintainer - thoughts?

shikadiqueen:
Removing documentation, however incomplete it may be, would be a massive step backwards.

Vultraz:
Definitely in favor of converting stuff to regular doc comments.

Pentarctagon:
Alright, will do that.

Pentarctagon:
Also, there are quite literally hundreds of warnings being output when I run doxygen locally, so I'm going to fix those in this PR as well.  Then hopefully be able to add a travis check to make sure warnings are known and dealt with in the future.

shikadiqueen:
That sounds like material for an entirely separate PR to me, otherwise it will be a review nightmare.

Pentarctagon:
They would be separate commits so they would be viewable independently, but I can make a separate PR if that would be easier.

PR #5400, opened 2020-12-30
Remove ICCP chunk from PNGs

doofus-01:
There were libpng warnings for some images, it wasn't always clear which ones.  This might not be the best way to fix it, but it seems to work.  I will have to re-optimize some of the files before this is merged, if it ever is.


Pentarctagon:
What is ICCP anyway?  I've seen warnings about there being bad profiles seemingly at random, but was never sure on what that actually meant in practice.

doofus-01:
It's an embedded color profile, http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.iCCP
Can't say I fully undertand it all, but it's data that's apparently not needed, unless we are professional photo publishers and really need fine control.  I know I'm to blame for some of the warnings, as I didn't realize GIMP was exporting with this crap by default.

doofus-01:
This wasn't the way to fix the warnings, maybe I'll try again some day.  For now, I'll just take care of whatever little file I stumble upon.

PR #5404, opened 2021-01-02
North-facing sprites

doofus-01:
This PR is for some minor touch-ups to old sprites, and addition of north-facing frames.

North-facing frames might make unit recognition more difficult if not done well, but I think it does more to fight the sliding statue effect than the bobbing animations I am trying to remove or reduce.


doofus-01:
Well, I didn't make as much progress on this as I'd hoped - it ended up being mostly orcish archer.  Better to get feedback sooner rather than later, so I'll merge this if CI passes.

ghype:
image path to old poacher-attack.png could be removed as no need for it anymore.

doofus-01:
True, that could be removed.  Or we could use it for an alternate animation.

PR #5406, opened 2021-01-02
Store content names for game history viewer.

Pentarctagon:


Vultraz:
What content names are being stored here? Addon names?

Pentarctagon:
No, it'd be the scenario/era/modification names.  Currently only the ID is stored, which isn't what would be best to display in the game history viewer.

loonycyborg (approved):

Vultraz (approved):

src/saved_game.cpp
```diff
@@ -330,6 +331,7 @@ void saved_game::load_non_scenario(const std::string& type, const std::string& i
 		non_scenario["required"] = cfg[require_attr].to_bool(require_default);
 		config& content = non_scenario.add_child("content");
 		content["id"] = id;
+		content["name"] = cfg["addon_title"].str(cfg["name"].str("Unknown"));
```
soliton-, line 334: Wouldn't it be more helpful to fall back to the addon_id than "Unknown"?

Pentarctagon: The ID is already stored though, and the ID and name can be the same value even when both are actually present, so if needed there wouldn't be any way to reliably check in the future for all cases where the name wasn't found.

soliton-: Ok but then why not leave it empty instead? Now you cannot decide if the name was empty or "Unknown" so when displaying to the user you'd like to fall back to the ID that seems awkward.

Pentarctagon: Updated then.

PR #5409, opened 2021-01-05
[1.14] Re-enable the editor help, add docs about keyboard modifiers and the scenario editor

### LINKED ISSUE #2964: 'todo:' notes in help pages for Scenario Editor

stevecotton:
A couple of warnings about the scenario mode being buggy are added to the
user-visible text, find these by looking for <bold> tags within the text.

This takes all the changes to the editor documentation that were made to master
up to and including a4711ebad9. There have been no changes at all to the editor
documentation in the 1.14 branch, while there have been several updates to the
master branch that would be good to backport if the help is displayed in the
1.14 branch - this commit simply pulls the entire file into the 1.14 branch
along with enabling the editor documentation (a small change to
data/core/help.cfg, as done in commit ed611f9438).

The sound source tool isn't mentioned at all in this documentation. There is a
button for it in the 1.14 UI, but that button's only function is to say that
it's not implemented, so documentation seems unnecessary.

Fixes #2964 for the 1.14 branch.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,10 @@
      * S09: Name the enemy team.
    * The South Guard:
      * S06b: Fix some enemies’ ambush abilities (issue #5383).
+ ### Editor
+   * Re-enabled topics for the terrain editor in the in-game help browser
+   * Added help topics for the scenario editor’s tools
+   * Added documentation about the files written by the editor
  ### Translations
    * Updated translations: Catalan, Chinese (Traditional), Portuguese (Brazil), Spanish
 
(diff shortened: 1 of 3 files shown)


Pentarctagon:
1.14 isn't in a string freeze, just FYI.

Pentarctagon (approved):

stevecotton:
> 1.14 isn't in a string freeze, just FYI.

Thanks. In that case, would getting it in right now be best?

Pinging @ivanovic too, as this is going to push a lot of text to the translators.

Pentarctagon:
Now is fine.  Given 1.14's pace, I wouldn't expect it to be in another string freeze for another couple months.

PR #5410, opened 2021-01-06
Add support for asynchronously querying a player's game history for display on the client.

Pentarctagon:
---

The server expects a request of (`search_for` optional):
```
[game_history_request]
    offset=NN
    search_for="a_username"
[/game_history_request]
```
and will send a response with a [game_history_result] per game in the form of:
```
[game_history_result]
    game_name="game name"
    reload=yes|no
    game_start="some timestamp"
    scenario_name="some scenario name"
    scenario_id="some scenario id"
    era_name="some era name"
    era_id="some era id"
    [player]
        name="p1"
        faction="drakes"
    [/player]
    [player]
        name="p2"
        faction="loyalists"
    [/player]
    [modification]
        name="some mod name 1"
    [/modification]
    [modification]
        name="some mod name 2"
    [/modification]
    [modification]
        id="some_mod_id_1"
    [/modification]
    [modification]
        id="some_mod_id_2"
    [/modification]
[/game_history_result]
```

Also, I'm thinking it would make sense at this point to add an index to the game table's `START_TIME` (for the `order by` clause) and the player table's USER_ID (since it's used in the `exists` part of the statement)

### DIFF
--- a/src/server/common/dbconn.cpp
+++ b/src/server/common/dbconn.cpp
@@ -16,6 +16,7 @@
 #include "server/common/dbconn.hpp"
 #include "server/common/resultsets/tournaments.hpp"
 #include "server/common/resultsets/ban_check.hpp"
+#include "server/common/resultsets/game_history.hpp"
 #include "log.hpp"
 
 static lg::log_domain log_sql_handler("sql_executor");
@@ -106,6 +107,65 @@ std::string dbconn::get_tournaments()
 	{
 		log_sql_exception("Could not retrieve the tournaments!", e);
 		return "";
+	}
+}
+
+std::unique_ptr<simple_wml::document> dbconn::get_game_history(int player_id, int offset)
+{
+	try
+	{
+		std::string game_history_query = "select "
+"  game.GAME_NAME, "
+"  game.RELOAD, "
+"  game.START_TIME, "
+"  GROUP_CONCAT(CONCAT(player.USER_NAME, ':', player.FACTION)) as PLAYERS, "
+"  IFNULL(scenario.NAME, '') as SCENARIO_NAME, "
+"  IFNULL(scenario.ID, '') as SCENARIO_ID, "
+"  IFNULL(era.NAME, '') as ERA_NAME, "
+"  IFNULL(era.ID, '') as ERA_ID, "
+"  IFNULL(GROUP_CONCAT(distinct mods.NAME, '') as MODIFICATION_NAMES, "
+"  IFNULL(GROUP_CONCAT(distinct mods.ID), '') as MODIFICATION_IDS, "
+"  concat('https://replays.wesnoth.org/', substring(game.INSTANCE_VERSION, 1, 4), '/', year(game.END_TIME), '/', lpad(month(game.END_TIME), 2, '0'), '/', lpad(day(game.END_TIME), 2, '0'), '/', game.REPLAY_NAME) as REPLAY_URL "
+"from "+db_game_info_table_+" game "
+"inner join "+db_game_player_info_table_+" player "
+"   on exists "
+"  ( "
+"    select 1 "
+"    from "+db_game_player_info_table_+" player1 "
+"    where game.INSTANCE_UUID = player1.INSTANCE_UUID "
+"      and game.GAME_ID = player1.GAME_ID "
+"      and player1.USER_ID = ? "
+"  ) "
+"  and game.INSTANCE_UUID = player.INSTANCE_UUID "
+"  and game.GAME_ID = player.GAME_ID "
+"  and player.USER_ID != -1 "
+"  and game.END_TIME is not NULL "
+"inner join "+db_game_content_info_table_+" scenario "
+"   on scenario.TYPE = 'scenario' "
+"  and scenario.INSTANCE_UUID = game.INSTANCE_UUID "
+"  and scenario.GAME_ID = game.GAME_ID "
+"inner join "+db_game_content_info_table_+" era "
+"   on era.TYPE = 'era' "
+"  and era.INSTANCE_UUID = game.INSTANCE_UUID "
+"  and era.GAME_ID = game.GAME_ID "
+"left join "+db_game_content_info_table_+" mods "
+"   on mods.TYPE = 'modification' "
+"  and mods.INSTANCE_UUID = game.INSTANCE_UUID "
+"  and mods.GAME_ID = game.GAME_ID "
+"group by game.INSTANCE_UUID, game.GAME_ID "
+"order by game.START_TIME desc "
+"limit 11 offset ? ";
+
+		game_history gh;
+		get_complex_results(create_connection(), gh, game_history_query, player_id, offset);
+		return gh.to_doc();
+	}
+	catch(const mariadb::exception::base& e)
+	{
+		log_sql_exception("Could not retrieve the game history for forum ID `"+std::to_string(player_id)+"`!", e);
+		auto doc = std::make_unique<simple_wml::document>();
+		doc->set_attr("error", "Error retrieving game history.");
+		return doc;
 	}
 }
 
--- a/src/server/common/forum_user_handler.cpp
+++ b/src/server/common/forum_user_handler.cpp
@@ -207,6 +207,14 @@ std::string fuh::get_tournaments(){
 	return conn_.get_tournaments();
 }
 
+void fuh::async_get_and_send_game_history(boost::asio::io_service& io_service, server_base& s_base, socket_ptr player_socket, int player_id, int offset) {
+	boost::asio::post([this, &s_base, player_socket, player_id, offset, &io_service] { 
+		boost::asio::post(io_service, [player_socket, &s_base, doc = conn_.get_game_history(player_id, offset)]{
+			s_base.async_send_doc_queued(player_socket, *doc);
+		});
+	 });
+}
+
 void fuh::db_insert_game_info(const std::string& uuid, int game_id, const std::string& version, const std::string& name, int reload, int observers, int is_public, int has_password){
 	conn_.insert_game_info(uuid, game_id, version, name, reload, observers, is_public, has_password);
 }
--- a/src/server/common/forum_user_handler.hpp
+++ b/src/server/common/forum_user_handler.hpp
@@ -121,6 +121,18 @@ public:
 	std::string get_tournaments();
 
 	/**
+	 * Runs an asynchronous query to fetch the user's game history data.
+	 * The result is then posted back to the main boost::asio thread to be sent to the requesting player.
+	 * 
+	 * @param io_service The boost io_service to use to post the query results back to the main boost::asio thread.
+	 * @param s_base The server instance the player is connected to.
+	 * @param player_socket The socket use to communicate with the player's client.
+	 * @param player_id The forum ID of the player to get the game history for.
+	 * @param offset Where to start returning rows to the client from the query results.
+	 */
+	void async_get_and_send_game_history(boost::asio::io_service& io_service, server_base& s_base, socket_ptr player_socket, int player_id, int offset);
+
+	/**
 	 * Inserts game related information.
 	 * 
 	 * @param uuid The value returned by @ref get_uuid().
--- a/src/server/common/resultsets/game_history.cpp
+++ b/src/server/common/resultsets/game_history.cpp
@@ -1,0 +1,87 @@
+/*
+   Part of the Battle for Wesnoth Project https://www.wesnoth.org/
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY.
+
+   See the COPYING file for more details.
+*/
+
+#ifdef HAVE_MYSQLPP
+
+#include "server/common/resultsets/game_history.hpp"
+#include "serialization/string_utils.hpp"
+#include "log.hpp"
+
+static lg::log_domain log_sql_handler("sql_executor");
+#define ERR_SQL LOG_STREAM(err, log_sql_handler)
+
+void game_history::read(mariadb::result_set_ref rslt)
+{
+    while(rslt->next())
+    {
+        result r;
+        r.game_name = rslt->get_string("GAME_NAME");
+        r.reload = rslt->get_boolean("RELOAD");
+        r.game_start = rslt->get_date_time("START_TIME").str();
+        r.scenario_name = rslt->get_string("SCENARIO_NAME");
+        r.scenario_id = rslt->get_string("SCENARIO_ID");
+        r.era_name = rslt->get_string("ERA_NAME");
+        r.era_id = rslt->get_string("ERA_ID");
+        for(const auto& player_info : utils::split(rslt->get_string("PLAYERS")))
+        {
+            std::vector<std::string> info = utils::split(player_info, ':');
+            if(info.size() == 2)
+            {
+                r.players.emplace_back(player{ info[0], info[1] });
+            }
+            else
+            {
+                ERR_SQL << "Expected player information to split into two fields, instead found the value `" << player_info << "`." << std::endl;
+            }
+        }
+        r.modification_names = utils::split(rslt->get_string("MODIFICATION_NAMES"));
+        r.modification_ids = utils::split(rslt->get_string("MODIFICATION_IDS"));
+        r.replay_url = rslt->get_string("REPLAY_URL");
+        results.push_back(std::move(r));
+    }
+}
+
+std::unique_ptr<simple_wml::document> game_history::to_doc()
+{
+    auto doc = std::make_unique<simple_wml::document>();
+    for(const auto& result : results)
+    {
+        simple_wml::node& ghr = doc->root().add_child("game_history_result");
+        ghr.set_attr_dup("game_name", result.game_name.c_str());
+        ghr.set_attr_int("reload", result.reload);
+        ghr.set_attr_dup("game_start", result.game_start.c_str());
+        ghr.set_attr_dup("scenario_name", result.scenario_name.c_str());
+        ghr.set_attr_dup("scenario_id", result.scenario_id.c_str());
+        ghr.set_attr_dup("era_name", result.era_name.c_str());
+        ghr.set_attr_dup("era_id", result.era_id.c_str());
+        for(const auto& player : result.players)
+        {
+            simple_wml::node& p = ghr.add_child("player");
+            p.set_attr_dup("name", player.name.c_str());
+            p.set_attr_dup("faction", player.faction.c_str());
+        }
+        for(const auto& mod : result.modification_names)
+        {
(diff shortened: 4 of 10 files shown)


src/server/common/dbconn.cpp
```diff
@@ -109,6 +110,61 @@ std::string dbconn::get_tournaments()
...
+std::unique_ptr<simple_wml::document> dbconn::get_game_history(int player_id, int offset)
+{
+	try
+	{
+		std::string game_history_query = "select "
+"  game.GAME_NAME, "
+"  game.RELOAD, "
+"  game.START_TIME, "
+"  GROUP_CONCAT(CONCAT(player.USER_NAME, ':', player.FACTION)) as PLAYERS, "
+"  IFNULL(scenario.NAME, scenario.ID) as SCENARIO, "
```
soliton-, line 122: Probably better to provide the client with all info and let it decide what to display.

For example the client might want to display "not set" in the user's language and the ID in a tooltip or so.

src/server/common/dbconn.cpp
```diff
@@ -109,6 +110,61 @@ std::string dbconn::get_tournaments()
...
+"   on era.TYPE = 'era' "
+"  and era.INSTANCE_UUID = game.INSTANCE_UUID "
+"  and era.GAME_ID = game.GAME_ID "
+"left join wesnothd_game_content_info mods "
+"   on mods.TYPE = 'modification' "
+"  and mods.INSTANCE_UUID = game.INSTANCE_UUID "
+"  and mods.GAME_ID = game.GAME_ID "
+"group by game.INSTANCE_UUID, game.GAME_ID "
+"order by game.START_TIME desc "
+"limit 10 offset ? ";
```
soliton-, line 153: There appear to be performance considerations with high offsets so we should be careful to make sure we have no such issues or limit the offset amount.

src/server/common/dbconn.cpp
Pentarctagon: Alright, will do.

src/server/common/dbconn.cpp
Pentarctagon: The potential performance issue as I understand it is that it can't just jump straight to the offset, rather it needs to also find everything else before that first.  Putting the index on the START_TIME column used in the `order by` should be enough, I think.  By the time we're in the range where even an indexed column has performance issues we'd be beyond the number of games any single person would've been able to play.

Vultraz (changes requested):
Suggesting adding an optional `user_id` key to `game_history_request` that takes the forum id of the user and falling back to the current player if absent. That would allow calling this for arbitrary players if that was desired at some point.

Pentarctagon:
A username would be more useful, since `player_connections_` allows directly searching by name, whereas the user ID is somewhat more out of the way to get to.

The client implementation of `[game_history_request]` is also not part of this PR, to be clear.

Vultraz:
I can handle the client side.

Pentarctagon:
I've added the `search_for` attribute for the username that will have their game history searched for.

src/server/common/dbconn.cpp
```diff
@@ -109,6 +110,63 @@ std::string dbconn::get_tournaments()
...
+
+		game_history gh;
+		get_complex_results(create_connection(), gh, game_history_query, player_id, offset);
+		return gh.to_doc();
+	}
+	catch(const mariadb::exception::base& e)
+	{
+		log_sql_exception("Could not retrieve the game history for user `"+std::to_string(player_id)+"`!", e);
+		std::unique_ptr<simple_wml::document> doc;
+		doc->set_attr_dup("error", "Error retrieving game history.");
```
Vultraz, line 165: This looks like nullptr access.

Pentarctagon: Fixed.

src/server/common/resultsets/game_history.cpp
```diff
@@ -0,0 +1,73 @@
...
+            p.set_attr_dup("name", player.name.c_str());
+            p.set_attr_dup("faction", player.faction.c_str());
+        }
+        for(const auto& mod : result.modifications)
+        {
+            simple_wml::node& m = ghr.add_child("modification");
+            m.set_attr_dup("name", mod.c_str());
+        }
+    }
+    return doc.clone();
```
Vultraz, line 70: How come we clone the doc here? Couldn't `doc` be created as a unique_ptr on the heap and just returned?

src/server/wesnothd/server.cpp
```diff
@@ -1807,7 +1807,17 @@ void server::handle_player_in_game(socket_ptr socket, simple_wml::document& data
...
+	} else if(data.child("game_history_request")) {
+		if(user_handler_) {
+			int offset = data.child("game_history_request")->attr("offset").to_int();
+			std::string player_name = data.child("game_history_request")->attr("search_for").to_string();
+			int player_id = user_handler_->get_forum_id(player_name);
+
+			LOG_SERVER << "Querying game history for player `" << player_name << "`." << std::endl;
+			user_handler_->get_game_history(io_service_, *this, socket, player_id, offset);
+		}
+		return;
```
Vultraz, line 1819: It doesn't seem this result is ever actually sent to player 🤔

src/server/common/resultsets/game_history.cpp
Pentarctagon: Mostly just it's still kind of weird to be using a unique_ptr to be able to return something.  Updated.

src/server/wesnothd/server.cpp
Pentarctagon: That happens in the `forum_user_handler`'s method.

src/server/common/resultsets/game_history.cpp
Vultraz: Hmm... You could make it return a `document` and add a move ctor/assignment op for when that function is actually used (since the `document` copy ctor is deleted).

src/server/common/forum_user_handler.cpp
```diff
@@ -207,6 +207,15 @@ std::string fuh::get_tournaments(){
 	return conn_.get_tournaments();
 }
 
+void fuh::get_game_history(boost::asio::io_service& io_service, server_base& s_base, socket_ptr player_socket, int player_id, int offset) {
+	boost::asio::post([this, &s_base, player_socket, player_id, offset, &io_service] { 
+		auto doc = conn_.get_game_history(player_id, offset);
+		boost::asio::post(io_service, [player_socket, &s_base, doc = std::move(doc)]{
```
Vultraz, line 213: No need to move this into the lambda. Can just do:
```cpp
boost::asio::post(io_service, [player_socket, &s_base, doc = conn_.get_game_history(player_id, offset)] {
```

src/server/common/resultsets/game_history.cpp
Pentarctagon: I'll leave that to someone else - this works, and I'm fine with not needing to try to figure out how to do that for simple_wml.

src/server/common/forum_user_handler.cpp
Pentarctagon: Updated as well.

Vultraz (approved):

Vultraz:
Actually, question - are both nested `post` calls still needed then? What does having two inside each other do?

Pentarctagon:
The outer `post` isn't provided an io_service, so it gets executed on a different boost thread pool asynchronously from everything else wesnothd is doing, which lets wesnoth do other things while the query runs.

The inner `post` is provided the same io_service that wesnothd is using, so it gets executed synchronously on the same thread as everything else wesnothd is processing, which is then why no additional thread safety logic is needed around the call to `async_send_doc_queued()`.

src/server/common/dbconn.cpp
soliton-: True, I assume just using a high offset without a high number of games being in the DB should not be an issue and as you say we're not likely to get such a high games count.

src/server/common/resultsets/game_history.cpp
```diff
@@ -0,0 +1,73 @@
...
+        r.scenario_id = rslt->get_string("SCENARIO_ID");
+        r.era_name = rslt->get_string("ERA_NAME");
+        r.era_id = rslt->get_string("ERA_ID");
+        for(const auto& player_info : utils::split(rslt->get_string("PLAYERS")))
+        {
+            std::vector<std::string> info = utils::split(player_info, ':');
+            if(info.size() == 2)
+            {
+                r.players.emplace_back(player{ info[0], info[1] });
+            }
```
soliton-, line 37: Perhaps this should be an assert or there should be an error message if the size is not 2.

src/server/common/dbconn.cpp
```diff
@@ -109,6 +110,64 @@ std::string dbconn::get_tournaments()
...
+		std::string game_history_query = "select "
+"  game.GAME_NAME, "
+"  game.RELOAD, "
+"  game.START_TIME, "
+"  GROUP_CONCAT(CONCAT(player.USER_NAME, ':', player.FACTION)) as PLAYERS, "
+"  IFNULL(scenario.NAME, '') as SCENARIO_NAME, "
+"  IFNULL(scenario.ID, '') as SCENARIO_ID, "
+"  IFNULL(era.NAME, '') as ERA_NAME, "
+"  IFNULL(era.ID, '') as ERA_ID, "
+"  IFNULL(GROUP_CONCAT(distinct mods.NAME), IFNULL(GROUP_CONCAT(distinct mods.ID), '')) as MODIFICATIONS, "
```
soliton-, line 126: Perhaps this should also be split in name and ID so the client can decide how to display this?

src/server/wesnothd/server.cpp
```diff
@@ -1807,7 +1807,17 @@ void server::handle_player_in_game(socket_ptr socket, simple_wml::document& data
 	} else if(data.child("stop_updates")) {
 		g.send_data(data, socket);
 		return;
-		// Data to ignore.
+	} else if(data.child("game_history_request")) {
+		if(user_handler_) {
+			int offset = data.child("game_history_request")->attr("offset").to_int();
+			std::string player_name = data.child("game_history_request")->attr("search_for").to_string();
+			int player_id = user_handler_->get_forum_id(player_name);
```
soliton-, line 1814: This should get the ID from the player object if no `search_for` was given. And probably also if search_for references an online player.

src/server/wesnothd/server.cpp
```diff
@@ -1807,7 +1807,17 @@ void server::handle_player_in_game(socket_ptr socket, simple_wml::document& data
...
+
+			LOG_SERVER << "Querying game history for player `" << player_name << "`." << std::endl;
```
soliton-, line 1816: Would be nice to include the ID as well, IMO.

src/server/common/forum_user_handler.hpp
```diff
@@ -120,6 +120,18 @@ class fuh : public user_handler
...
+	 * Runs an asynchronous query to fetch the user's game history data.
+	 * The result is then posted back to the main boost::asio thread to be sent to the requesting player.
+	 * 
+	 * @param io_service The boost io_service to use to post the query results back to the main boost::asio thread.
+	 * @param s_base The server instance the player is connected to.
+	 * @param player_socket The socket use to communicate with the player's client.
+	 * @param player_id The forum ID of the player to get the game history for.
+	 * @param offset Where to start returning rows to the client from the query results.
+	 */
+	void get_game_history(boost::asio::io_service& io_service, server_base& s_base, socket_ptr player_socket, int player_id, int offset);
```
soliton-, line 133: Should rather be called `async_get_and_send_...` or so.

soliton- (changes requested):

Vultraz:
Another thought: is there a way to get the total number of games, or to know in the response that there are no more games to fetch?

Pentarctagon:
Probably the simplest would be to fetch 11 rows but only display 10 rows.  Then if there's ever 10 or fewer rows in a response, that would mean there's nothing left to fetch.  Otherwise the query will be slower since it needs to find and count all rows that could possibly be returned by the query instead of being able to stop early due to the limit+offset.

src/server/common/resultsets/game_history.cpp
Pentarctagon: Which would you prefer?

src/server/wesnothd/server.cpp
Pentarctagon: Is there a case where no `search_for` would be given?

src/server/common/dbconn.cpp
```diff
@@ -109,6 +110,63 @@ std::string dbconn::get_tournaments()
...
+"order by game.START_TIME desc "
+"limit 10 offset ? ";
+
+		game_history gh;
+		get_complex_results(create_connection(), gh, game_history_query, player_id, offset);
+		return gh.to_doc();
+	}
+	catch(const mariadb::exception::base& e)
+	{
+		log_sql_exception("Could not retrieve the game history for user `"+std::to_string(player_id)+"`!", e);
```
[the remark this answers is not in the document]
Pentarctagon: Updated.

src/server/common/dbconn.cpp
```diff
@@ -109,6 +110,64 @@ std::string dbconn::get_tournaments()
...
+		auto doc = std::make_unique<simple_wml::document>();
+		doc->set_attr_dup("error", "Error retrieving game history.");
```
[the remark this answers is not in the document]
Pentarctagon: Updated.

src/server/common/dbconn.cpp
Pentarctagon: Updated.

src/server/wesnothd/server.cpp

src/server/common/forum_user_handler.hpp
Pentarctagon: Updated.

src/server/common/resultsets/game_history.cpp
soliton-: I think so far it can only happen if someone edits the DB manually? Still I'd rather be prepared for it to happen and print an error.

src/server/wesnothd/server.cpp
soliton-: I assume `search_for` is given if you want the history for another player?

There should also be an early abort when `get_forum_id` returns 0.

Pentarctagon: Searching for yourself would also make sense, I think.  Or the client could just directly provide the ID since it knows its own forum ID at this point.

src/server/common/resultsets/game_history.cpp
Pentarctagon: Added an error logging statement.

src/server/wesnothd/server.cpp
Vultraz: Searching for yourself would be the primary usecase (at least initially). Since the forum ID is now known for everyone client-side it would be good to support passing that instead of (or in addition) to username.

Pentarctagon: So essentially:
* if the forum ID attribute is present -> use that
* else if the player being searched for is online -> get the forum ID from wesnothd's player info
* else get the forum ID by querying the database, making sure it didn't return a zero before actually running the query

Vultraz: Sure. Or we could just make the ID attribute mandatory. Either works.

src/server/common/resultsets/game_history.cpp
```diff
@@ -0,0 +1,87 @@
...
+            else
+            {
+                ERR_SQL << "Expected player information to split into two fields, instead found `" << info.size() << "` fields." << std::endl;
```
soliton-, line 44: The error should give enough info to find the issue in the DB. Though not sure if we have access to the query data here? Could at least show the offending string then.

src/server/wesnothd/server.cpp
soliton-: I would not give either `forum_id` or `search_for` when asking for own game history.

Is there even a need for both `forum_id` and `search_for`? I would just pick one. Probably `search_for` so you can search for users you just know the name of. Which is basically what Vultraz suggested initially just with name instead of forum ID.

src/server/common/resultsets/game_history.cpp
```diff
@@ -0,0 +1,87 @@
...
+            }
+        }
+        r.modification_names = utils::split(rslt->get_string("MODIFICATION_NAMES"));
+        r.modification_ids = utils::split(rslt->get_string("MODIFICATION_IDS"));
+        r.replay_url = rslt->get_string("REPLAY_URL");
+        results.emplace_back(r);
```
Vultraz, line 50: Suggest using `push_back(std::move(r))` here to make use of the implicitly-declared move ctor for `result` rather than making a copy.

src/server/wesnothd/server.cpp
Pentarctagon: So in that case:
* if `search_for` is not present -> get forum ID from the requesting player's info
* else if `search_for` is present -> search the players currently online and if found get the forum ID from their player info
* else -> query the forum database for their forum ID

src/server/common/resultsets/game_history.cpp
Pentarctagon: Only the result set is available here.  I've added the string to the error output.

src/server/common/resultsets/game_history.cpp
Pentarctagon: Updated.

src/server/wesnothd/server.cpp
Vultraz: Why the restriction of needed to be online, though?

soliton-: The last case is for offline players.

Pentarctagon: Alright, updated.

src/server/wesnothd/server.cpp
```diff
@@ -1807,7 +1807,33 @@ void server::handle_player_in_game(socket_ptr socket, simple_wml::document& data
 	} else if(data.child("stop_updates")) {
 		g.send_data(data, socket);
 		return;
-		// Data to ignore.
+	} else if(data.child("game_history_request")) {
```
soliton-, line 1810: You could assign the child in one go same as a couple lines up and then use that instead of calling `child()` again.

soliton- (approved):
Nice work!

src/server/wesnothd/server.cpp

Pentarctagon:
@Vultraz last question then - did you want to go with the "fetch 11 rows" approach?

Pentarctagon:
@soliton- also, are you good with me adding those two indexes?

Vultraz:
> @Vultraz did you want to go with the "fetch 11 rows" approach?

I don't see why not. Is there any other alternative?

Pentarctagon:
Not one that would avoid needing to at least count the total number of rows the query would return, which would affect how quickly the query completes.

Vultraz:
Alright, then we can go with that.

soliton-:
> @soliton- also, are you good with me adding those two indexes?

Very much so.

Pentarctagon:
Alright, added those to the PR, and I'll go ahead and add them to the actual database as well.

PR #5413, opened 2021-01-08
Make wesnothd use iterators instead of socket_ptrs to keep track of players

loonycyborg:
Use `player_iterator` typedef instead of `socket_ptr` to refer to players once they're logged in. `player_iterator` points directly to player's record in `player_connections_` container allowing for access to any stats associated with player without searching by socket. Also this paves way to potentially remembering disconnected players for purpose of implementing player reconnect feature.

Interface conventions:
- `player_iterator` arguments should always refer to player existing in `player_connections_`. Potential past-the-end iterators should be checked for before passing the `player_iterator` to other functions.
- `utils::optional<player_iterator>` should be used in cases where actual player may not exist.

### DIFF
--- a/src/server/wesnothd/game.cpp
+++ b/src/server/wesnothd/game.cpp
@@ -74,12 +74,6 @@ namespace wesnothd
 int game::id_num = 1;
 int game::db_id_num = 1;
 
-void game::missing_user(player_iterator /*socket*/, const std::string& func) const
-{
-	WRN_GAME << func << "(): Could not find user (socket:\t<some C++ pointer>"
-			 << ") in player_info_ in game:\t\"" << name_ << "\" (" << id_ << ", " << db_id_ << ")\n";
-}
-
 game::game(wesnothd::server& server, player_connections& player_connections,
 		player_iterator host,
 		const std::string& name,
--- a/src/server/wesnothd/game.cpp
+++ b/src/server/wesnothd/game.cpp
@@ -1927,10 +1927,11 @@ std::string game::debug_sides_info() const
 utils::optional<player_iterator> game::find_user(const simple_wml::string_span& name)
 {
 	auto player { player_connections_.get<name_t>().find(name.to_string()) };
-	if(player != player_connections_.get<name_t>().end())
+	if(player != player_connections_.get<name_t>().end()) {
 		return player_connections_.project<0>(player);
-	else
+	} else {
 		return {};
+	}
 }
 
 void game::send_and_record_server_message(const char* message, utils::optional<player_iterator> exclude)
--- a/src/server/wesnothd/game.cpp
+++ b/src/server/wesnothd/game.cpp
@@ -74,14 +74,14 @@ namespace wesnothd
 int game::id_num = 1;
 int game::db_id_num = 1;
 
-void game::missing_user(socket_ptr /*socket*/, const std::string& func) const
+void game::missing_user(player_iterator /*socket*/, const std::string& func) const
 {
 	WRN_GAME << func << "(): Could not find user (socket:\t<some C++ pointer>"
 			 << ") in player_info_ in game:\t\"" << name_ << "\" (" << id_ << ", " << db_id_ << ")\n";
 }
 
 game::game(wesnothd::server& server, player_connections& player_connections,
-		const socket_ptr& host,
+		player_iterator host,
 		const std::string& name,
 		bool save_replays,
 		const std::string& replay_save_path)
@@ -115,18 +115,11 @@ game::game(wesnothd::server& server, player_connections& player_connections,
 	, rng_()
 	, last_choice_request_id_(-1) /* or maybe 0 ? it shouldn't matter*/
 {
-	assert(owner_);
 	players_.push_back(owner_);
 
-	const auto iter = player_connections_.find(owner_);
-	if(iter == player_connections_.end()) {
-		missing_user(owner_, __func__);
-		return;
-	}
-
 	// Mark the host as unavailable in the lobby.
-	iter->info().mark_available(id_, name_);
-	iter->info().set_status(player::PLAYING);
+	owner_->info().mark_available(id_, name_);
+	owner_->info().set_status(player::PLAYING);
 }
 
 game::~game()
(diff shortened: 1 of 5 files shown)


Vultraz:
Looks all good to me, save a few minor formatting issues.

src/server/wesnothd/game.hpp
```diff
@@ -20,15 +20,17 @@
 #include "server/common/simple_wml.hpp"
 #include "utils/make_enum.hpp"
 
+#include <utils/optional_fwd.hpp>
```
Vultraz, line 23: Not a system header, shouldn't use brackets.

src/server/wesnothd/server.hpp
```diff
@@ -24,6 +24,7 @@
 #include "server/wesnothd/player_connection.hpp"
 
 #include <boost/asio/steady_timer.hpp>
+#include <utils/optional_fwd.hpp>
```
Vultraz, line 27: Not a system header, shouldn't use brackets.

src/server/wesnothd/server.hpp
```diff
@@ -190,7 +193,7 @@ class server : public server_base
 
 	void delete_game(int, const std::string& reason="");
 
-	void update_game_in_lobby(const wesnothd::game& g, const socket_ptr& exclude=socket_ptr());
+	void update_game_in_lobby(const game& g, utils::optional< wesnothd::player_iterator > exclude = {});
```
Vultraz, line 196: Excessive spacing in template, also probably don't need to specify `wesnothd`.

src/server/wesnothd/game.cpp
```diff
@@ -1980,42 +1917,41 @@ std::string game::debug_sides_info() const
...
+utils::optional<player_iterator> game::find_user(const simple_wml::string_span& name)
 {
-	const auto iter = player_connections_.get<name_t>().find(name.to_string());
-	if(iter != player_connections_.get<name_t>().end()) {
-		return iter->socket();
-	} else {
-		return socket_ptr();
-	}
+	auto player { player_connections_.get<name_t>().find(name.to_string()) };
+	if(player != player_connections_.get<name_t>().end())
```
Vultraz, line 1930: Block brackets got removed.

Vultraz (changes requested):

src/server/wesnothd/game.cpp
```diff
@@ -74,14 +74,14 @@ namespace wesnothd
 int game::id_num = 1;
 int game::db_id_num = 1;
 
-void game::missing_user(socket_ptr /*socket*/, const std::string& func) const
+void game::missing_user(player_iterator /*socket*/, const std::string& func) const
```
Pentarctagon, line 77: Maybe at this point just remove this argument and update the warning text to not mention the socket?

src/server/wesnothd/game.cpp
```diff
@@ -177,37 +170,26 @@ bool game::is_muted_observer(const socket_ptr& player) const
...
-	if(iter != player_connections_.end()) {
-		return iter->info().name();
-	}
-
-	return "(unknown)";
+	return iter->info().name();
 }
 
-std::string game::list_users(user_vector users, const std::string& func) const
+std::string game::list_users(user_vector users, const std::string& /*func*/) const
```
Pentarctagon, line 183: Might as well remove the `func` argument here too.

src/server/wesnothd/game.cpp
loonycyborg: Also considering to remove this entire function.

Vultraz (approved):

src/server/wesnothd/game.cpp
```diff
@@ -522,49 +487,45 @@ void game::transfer_side_control(const socket_ptr& sock, const simple_wml::node&
...
 		msg << "You can't give away side " << side_num << ". It's controlled by '" << old_player_name << "' not you.";
 		DBG_GAME << msg.str() << "\n";
-		send_server_message(msg.str(), sock);
+		send_server_message(msg.str(), player);
 		return;
 	}
 
 	// find the player that is passed control
-	socket_ptr newplayer = find_user(newplayer_name);
+	auto newplayer { find_user(newplayer_name) };
```
soliton-, line 524: Is there an issue with using `auto newplayer = find_user(newplayer_name)`?

loonycyborg: Getting in the habit of using brace initialization for clarity, to avoid the "most vexing parse" issue.

loonycyborg: I mean since we now can use brace initialization there's no need for copy-initialization.

soliton-: Is this a general change we'll employ in the complete code base? Should this be added to the coding standards/guide lines?

Vultraz: Isn't this still copy initialization, just now with the copy constructor rather than the copy assignment operator?

AI0867: Aren't both copy constructors? `classname foo = bar;` does not default-construct foo followed by copy assignment, it calls the copy constructor directly. That is, it is very much not the same as `classname foo; foo = bar;`.

That means that `auto foo = bar;` and `auto foo { bar };` should call the same constructor, but there might be different parsing details. (maybe initializer lists could get involved)

loonycyborg: It should generally work in the same way due to copy elision. I'm still a bit torn on what of those two ways I prefer. I find that assignment annoying conceptually because it is not real assignment. Brace initialization shows the intent better. But I don't insist that entire codebase should be revamped to brace initialization. Assignment looks closer to classic C ways of doing things and using it might be a habit that is hard to get rid of.

Vultraz: Personally, I tend to prefer assignment initialization just because it's more akin to how most languages do things, but I have no objection to brace-init.

Also, I must say, once one starts debating copy elision that's a sign we've descended into the C++ weeds 😂

soliton-: I would suggest to decide one way or another or describe clear rules on when to use one or the other. I don't think it's useful to just mix and match arbitrarily.

soliton- (approved):

Issue #5414, opened 2021-01-09
WFL variables aren't getting assigned correctly in [damage] special.


virtualghetto:
### Game and System Information
 - What version of the game are you running?
 - 1.15.8
 - If you're using a custom build, what compiler did you use, and what commit did you build at?
 - tag 1.15.8
 - What OS are you running?
 - linux

### Describe the bug
The WFL variables described [here]( https://wiki.wesnoth.org/AbilitiesWML#Available_formula_variables_in_Abilities_and_Weapon_Sepcials) 
`self`, `student`, `attacker`, `defender`, `other`
all point to `self` when accessed in the `[damage]` weapon special context.

Also, it would also be nice if the current damage base value is assigned a WFL variable for further operations.

newfrenchy83:
i look at https://github.com/wesnoth/wesnoth/commit/782f9ca98a7c40f8705aad795ce8d2fa62db261a#diff-eb9c8f8b8122951ad515f39be0e4d8550c4f22421024154f55bdef508de85897R473 commit and i think what the issue come to there, @gfgtdf could you review that?

PR #5415, opened 2021-01-10
fix WFL variables aren't getting assigned correctly 

newfrenchy83:
fix attacker and defender formula pointed both' self'. Fix https://github.com/wesnoth/wesnoth/issues/5414 issue.


Vultraz:
I think it would be clearer to just flip the sides instead  of negating the conditional:
```cpp
is_attacker_ ? other_ : self_
```

newfrenchy83:
> 
> 
> I think it would be clearer to just flip the sides instead of negating the conditional:
> 
> ```c++
> is_attacker_ ? other_ : self_
> ```

it's done

PR #5432, opened 2021-01-14
Silence wmllint error about a missing campaign id

hexagonrecursion:
This PR fixes the following wmllint error:
```
"../../data/campaigns/Dead_Water/_main.cfg", line 76: campaign requires an ID attribute but has none
```

This wmllint error was caused by the fact that the campaign definition is split into two and wmllint is not smart enough to understand that.

https://github.com/wesnoth/wesnoth/blob/92fedf8f2aa7a4b83fe0e59e7e0e4f2c16e3c207/data/campaigns/Dead_Water/_main.cfg#L7-L8
https://github.com/wesnoth/wesnoth/blob/92fedf8f2aa7a4b83fe0e59e7e0e4f2c16e3c207/data/campaigns/Dead_Water/_main.cfg#L74-L76

### DIFF
--- a/data/campaigns/Dead_Water/_main.cfg
+++ b/data/campaigns/Dead_Water/_main.cfg
@@ -71,9 +71,12 @@
     {campaigns/Dead_Water/units}
 [/units]
 
+# wmllint bug: wmllint is not smart enough to realise that the campaign already has an id
+# wmllint: validate-off
 [+campaign]
     {ENABLE_KRAKEN}
 [/campaign]
+# wmllint: validate-on
 
 {campaigns/Dead_Water/scenarios}
 


soliton-:
I would suggest for wmllint to ignore [+campaign] instead.

hexagonrecursion:
Want to try fixing that? Beware:
* there are **NO unit tests** (as far as I can tell)
* translator() is a very long function
* tagstack is a global variable
* despite its name validate_on_pop() does more than just validating: it mutates the tagstack
https://github.com/wesnoth/wesnoth/blob/75271a824657b0be63369ed4fc55cf0cb3579429/data/tools/wmllint#L2774-L2776
https://github.com/wesnoth/wesnoth/blob/75271a824657b0be63369ed4fc55cf0cb3579429/data/tools/wmllint#L968-L974

soliton-:
To fix it I'd add info about + tags to the tagstack and use that to skip any tests that test for some required attribute since that's just not going to work well anywhere.

For this specific case the easiest is to just remove the silly `[+campaign]` use entirely though.

PR #5434, opened 2021-01-15
Add [effect]apply_to_vision= attribute to [effect]apply_to=movement

### LINKED ISSUE #3356: [1.15] EffectWML's interaction between vision and movement
In Wesnoth 1.14, the interactions between `[effect]apply_to=movement` and `[effect]apply_to=vision` depend on the order of the operations. Consider a unit with 5 moves and default vision:
* It has (effectively) 5 moves and 5 vp.
* After (mp + 1) (vp + 2), it will have 6 moves and 8 vp.
* After (vp + 2) (mp + 1), it will have 6 moves and 7 vp.

This is because vision defaults to -1 which means "the same as the max movement points". Using any effect with apply_to=vision sets the number of vision points, and from then on they're no longer affected by the movement points.

I suggest adding an `affects_vision` key to `apply_to=movement` with options
* `yes` (default for increase) - if vision is not -1, then the increase will be applied to vision as well
* what to call this, maybe `as_before` - the vision variable is ignored, so the effects are the same as Wesnoth 1.14.
* `no` - if vision is -1, sets it to the current movement before the effect is applied

Please would someone add the [low-priority] tag?

stevecotton:
Add attribute also_apply_to_vision to [effect]apply_to=movement

When processing [effect]apply_to=movement, if also_apply_to_vision is 'yes'
(which is the default) then the vision points will change by the same amount.
Previously, it sometimes affected vision and sometimes didn't; for most cases
where it makes a difference, I expect it to be a change from unanticipated
behavior to expected behavior.

Please refer to the new unit test added in this commit for more detailed docs;
that test is also a rough draft for the Wiki update needed when this merges.

The reason it sometimes affected vision was that the special value of -1 vision
points was interpreted as "use the value of the max movement points instead".
The special value of -1 is still supported and frequently used, and refactor that
is out of scope for this commit - it's easy to check when the code path changed
in this commit is used, however it's considerably more complex to find all
routes that create a unit with vision set to -1.

I'm expecting one add-on, Rebirth In Nature, to need a large update for this;
as well as a trivial change needed to the Add Creature Pack.

There are several mods that have their own handling for vision, recalculate the
values frequently, and are expected to continue working as before, as they'll
just overwrite any changes that the engine makes.
* Proper Flying Mod / Proper Vision Mod
* LotI
* Blessed Altar Mod
* Shards Era's `CALCULATE_WEAPONS_ONLY` macro
* Castle of Evil Spirit

Rebirth In Nature is an RPG-style campaign that expects the player's unit to
have vision separated from movement along with right-click menus that change
the character's movement, this will probably need `also_apply_to_vision=no` in
many places. OTOH, its `item_id=mobility` uses `apply_to=movement` followed by
`apply_to=vision` to implement in 1.14 what this change will make default
behavior in 1.16.

There are a few sharpshooter units which have vision better than their
movement. If given a buff which boosts their movement, this will give
them a buff to vision too; likely a change towards expected behavior.
* Archaic Era's Royal Ranger
* Ageless Era's Royal Ranger and Dwarvish Forest Sniper
* Eastern Europe at War's Yacht and Great Yacht
* Era of More Units's Lone Wolf
* Southernerns has several sharpshooters
* WWII Era has several sharpshooters
* War of Dominions has some guard towers with very low movement

Silver Age has abilities that modify vision based on time-of-day, but unlike
the Proper Flying Mod and LotI these are done by adding and removing objects
that add or subtract one vision point, these should be unaffected by this
change. Unrelated to that, it declares values for `[unit_type]vision=` even
when the value is the same as max moves - like the sharpshooters this means
that buffs to their movement wouldn't have buffed their vision but now do.

There are few units with vision less than movement. Generally these seem to be
missile weapons that are represented as units, and they probably won't meet any
apply_to=movement statements - even if a movement power-up is available, the
player would probably choose a different unit.

The Add Creature Pack has two creatures (the Cactose Elder and Carnivore Fatal
Plant) that have reduced vision, along with AMLAs that increase those units'
movement; these AMLAs will need to use `also_apply_to_vision=no`. These two
creatures are used in Castle of Evil Spirit, but they won't get enough
experience to get an AMLA there.

Fixes #3356.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -27,6 +27,7 @@
    * Modify implementation of overwrite_specials attribute for replace yes/no parameter by none/one_side/both_sides and select abilities used like weapons and specials who must be overwrited(owned by fighter where special applied or both)
    * Add a `ability_id_active` attribute to `[filter]`
    * `[terrain_mask]` now accepts `mask_file` as an alternative to an inline mask. The file is loaded from the same place as `map_file` in the `[scenario]` tag (ie, a maps/ subdirectory of your binary path). Anyone who prefers to keep masks separate from regular maps is free to make a subdirectory for their masks (or just keep all their masks inline).
+   * `[effect]apply_to=movement` now always affects vision too, except when given the `apply_to_vision=no` attribute
  ### Miscellaneous and Bug Fixes
    * More optimization in the UI drawing code, fixes the crash displaying the full credits (issue #5043).
    * Made GUI.pyw compatible with Python 3.9 (issue #5719).
--- a/data/schema/units/modifications.cfg
+++ b/data/schema/units/modifications.cfg
@@ -51,7 +51,13 @@
 				[/tag]
 			[/case]
 			[case]
-				value=movement,vision,jamming,experience,max_experience,recall_cost
+				value=movement
+				{SIMPLE_KEY increase s_int_percent}
+				{SIMPLE_KEY set s_int}
+				{DEFAULT_KEY apply_to_vision s_bool yes}
+			[/case]
+			[case]
+				value=vision,jamming,experience,max_experience,recall_cost
 				{SIMPLE_KEY increase s_int_percent}
 				{SIMPLE_KEY set s_int}
 			[/case]
--- a/data/test/scenarios/test_effect_on_vision.cfg
+++ b/data/test/scenarios/test_effect_on_vision.cfg
@@ -134,6 +134,91 @@
     [/event]
 )}
 
+# Test that changing the max_moves has a corresponding effect on vision points, even when the vision points have already been set to a separate value.
+# This behavior is different to that in 1.14.
+{GENERIC_UNIT_TEST "effect_move_separated_vision" (
+    [event]
+        name = side 1 turn 1
+
+        # Ignore the default values for an Elvish Archer, this test needs to be able to determine whether increase=50% was calculated based on the
+        # max movement points or the vision points, so it needs the values to be different.
+        [modify_unit]
+            [filter]
+                id=alice
+            [/filter]
+            [effect]
+                apply_to=movement
+                set=15
+            [/effect]
+            [effect]
+                apply_to=vision
+                set=3
+            [/effect]
+        [/modify_unit]
+        {ASSERT_ALICE_STAT max_moves equals 15}
+        {ASSERT_VISION_EFFECTIVELY equals 3}
+        {ASSERT_ALICE_STAT jamming equals 0}
+
+        # Increasing movement should affect vision
+        [modify_unit]
+            [filter]
+                id=alice
+            [/filter]
+            [effect]
+                apply_to=movement
+                increase=1
+            [/effect]
+        [/modify_unit]
+        {ASSERT_ALICE_STAT max_moves equals 16}
+        {ASSERT_VISION_EFFECTIVELY equals 4}
+        {ASSERT_ALICE_STAT jamming equals 0}
+
+        # Increasing movement by 50% increases vision by (50% of movement) not by (50% of vision)
+        [modify_unit]
+            [filter]
+                id=alice
+            [/filter]
+            [effect]
+                apply_to=movement
+                increase=50%
+            [/effect]
+        [/modify_unit]
+        {ASSERT_ALICE_STAT max_moves equals 24}
+        {ASSERT_VISION_EFFECTIVELY equals 12}
+        {ASSERT_ALICE_STAT jamming equals 0}
+
+        # Decreasing movement should affect vision
+        [modify_unit]
+            [filter]
+                id=alice
+            [/filter]
+            [effect]
+                apply_to=movement
+                set=16
+            [/effect]
+        [/modify_unit]
+        {ASSERT_ALICE_STAT max_moves equals 16}
+        {ASSERT_VISION_EFFECTIVELY equals 4}
+        {ASSERT_ALICE_STAT jamming equals 0}
+
+        # Vision shouldn't go below zero
+        [modify_unit]
+            [filter]
+                id=alice
+            [/filter]
+            [effect]
+                apply_to=movement
+                increase=-50%
+            [/effect]
+        [/modify_unit]
+        {ASSERT_ALICE_STAT max_moves equals 8}
+        {ASSERT_VISION_EFFECTIVELY equals 0}
+        {ASSERT_ALICE_STAT jamming equals 0}
+
+        {SUCCEED}
+    [/event]
+)}
+
 # Test that apply_to=vision works
 {GENERIC_UNIT_TEST "effect_vision" (
     [event]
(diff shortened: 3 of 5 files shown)


stevecotton:
The CI test failure was because I included a test that's for the intended follow-up refactor; the refactor to not use -1 at all.

I'll drop the list of add-ons from the commit message for the next push. Current status of checking add-ons:

Add-ons that would need to change for this
---

For both the Proper Flying Mod and Silver Age, I'll post comments in their discussion threads before making the PR non-draft.

### Proper Flying Mod

The Proper Flying Mod completely separates vision from movement, giving most terrains a vision_cost of 1 but giving most units less vision points than movement points. The number of vision points is recalculated whenever the time-of-day changes.

With 1.14's behavior, any game with PFM enabled would effectively have `affect_vision` default to `no`. Picking up an item from another add-on that gives additional movement will temporarily give an unusually good vision boost, however as the vision is reset when the ToD changes that's likely to be overwritten before the player's next turn.

### Silver_Age_demo

This declares the vision value in its [unit_type] tags, even when they're the same as the max_moves. There are many abilities here that modify vision based on time-of-day.

### Rebirth In Nature

Rebirth In Nature's `item_id=mobility` has an effect that uses `apply_to=movement` followed by `apply_to=vision`. That ought to be simply a case of adding `affect_vision=no` to get code that works both with and without the change, although once the 1.14 version is no longer needed it could just be `affect_vision=yes` and remove the `[effect]apply_to=vision`.

As this RPG-style campaign expects the player's unit to have vision separated from movement, various of the other abilities should probably have an explicit `affect_vision=no`.

### add_creature_pack

This has two creatures (the Cactose Elder and Carnivore Fatal Plant) that have reduced vision, along with AMLAs that increase those units' movement. These will need an explicit affect_vision=no.

These two creatures are used in Castle of Evil Spirit, but they won't get enough experience to get an AMLA there.

### Bob's RPG Era

Bob's RPG Era has a couple of items that increase vision, but only until end-of-turn. The separation of vp from mp would be reverted when the unit is rebuilt at the end of the turn.

This one's almost in the "should work the same" category, the only edge case is when a map allows a player to pick up a vision boost followed on the same turn by a movement boost, but the map has been designed to only allow it in that order - the result is an extra vision boost for the current turn only.

The following add-ons do something with vision, but should work the same with this change
---

* Genesis S15 uses [modify_unit] to set both max_moves and vision, and then reverts the change near the end of the level. There's a cutscene between the revert and the end, otherwise it would probably be better to use an [object]duration=scenario. Either way, it shouldn't be a problem for apply_to=vision.
* Legend_of_the_Invincibles_beta handles multiple items providing vision bonuses, and recalculates vision based on `max_moves`.
* Blessed_Altar_Mod handles multiple items providing vision bonuses, and recalculates vision based on `max_moves`.
* Shards_Era's `CALCULATE_WEAPONS_ONLY` macro uses the same code as Blessed_Altar_Mod. Did the code originally come from LotI?
* Diplomatic_Mission has an AMLA with apply_to=movement and an AMLA with apply_to=vision, but these can only be taken in that order. I assume the player won't get 2 AMLAs before getting the movement ring in S02.
* Return_from_Captivity limits all units' vision from S08 onwards (and redoes this in recruit,recall,post_advance events). The event seems to be missing from S10, not sure if that's intentional.
* Ageless_Survival_Race has a power-up that gives +1mp +2vp, in that order.

These add-ons are likely to change to use the new feature
---

* SXC currently doesn't have any specific-to-vision code yet, it will have some added for its 1.15 port.
* Tunnel Rats has some units that are made inactive by setting their movement to zero followed by giving them vision - this could be done more cleanly affect_vision=no when removing the movement. The units with explicit vision are enemies, they don't survive the scenario.

These add-ons have units with explicitly set vision values, but shouldn't have a problem with this change:
---

### Sharpshooters

The following units have vision better than their movement. When combined with other add-ons, movement boosts for sharpshooters should probably increase their vision, this commit would cause that change.

* Archaic Era has a sharpshooter (the Royal Ranger) with increased vision, and separately other units whose custom AMLA increases movement, but no units that have both.
* Ageless Era has two sharpshooters with increased vision (Royal Ranger and Dwarvish Forest Sniper), and separately other units whose custom AMLA increases movement, but no units that have both.
* Eastern_Europe_at_War has units (the Yacht and Great Yacht) with increased vision, and separately other units whose custom AMLA increases movement, but no units that have both.
* Era_of_more_units has a scout (the Lone Wolf)
* Southernerns has sharpshooters
* WWII_Era has sharpshooters

### Blind units

Generally these are missiles that are represented as units. These units probably won't meet any apply_to=movement statements - even if a movement power-up is available, the player would probably choose a different unit.

* Antar has an "illusion" unit that shouldn't be able to see anything
* Lazersquad has missile units
* Chaoz has a missle unit

### Others with explicit vision values

* Defense of Elensefar wants the player to be able to highlight their lighthouses' vision range, and messes about with movement to do that. Nothing related to the affect_vision change.
* A_Random_RPG_Adventure allows the DM to modify any unit's stats, any change would be to the DM's input (a human player's input). Having asked in the forum thread, changing the vision is intended to be for effects such as roleplaying blindness or low vision.
* Wild_Frontiers has a lighthouse with vision different to movement, no problems with the affect_vision change.

WIP: check the following
---

* DiD S09 - opened issue #5440.
* SXRPG does have specific-to-vision code, and may need modifications.
* Castle of Evil Sprit sets player vision (and it's tied to the amount of lamp oil available as well as movement).
* War_of_Dominions
* WesCraft
* Conquestplus_space
* Conquestplus
* The_Ravagers. It has the same AMLAs as Diplomatic_Mission, but has more to check too.

stevecotton:
Rebasing to solve the changelog conflict.

Pentarctagon:
I'd prefer just `apply_to_vision` instead of `also_apply_to_vision`.

Also, was the WIP from your previous comment completed?

stevecotton:
The next push will change it to `apply_to_vision`.

Of the WIPs in the previous comment:
* DiD's #5400 has been fixed.
* SXRPG didn't need modifications. It has some abilities that reduce vision_costs, but doesn't make any changes to vision points (except via apply_to=movement).
* Castle of Evil Spirit's oil-lamp calculation is now in the "will just overwrite any changes that the engine makes" group.
* Castle of Evil Spirit's forum thread also covers the Add Creature Pack, https://r.wesnoth.org/p664287
* War_of_Dominions is in the "sharpshooters" group.
* WesCraft, Conquestplus and Conquestplus_space are all in the "missile weapons that are represented as units" group.
* The_Ravagers has a side-game of chess, where all units can lose vision - at the end of that side-game, all the chess pieces are killed, so there's no unit that will have movement changed after its vision is changed.

stevecotton:
> The next push will change it to apply_to_vision.

I said that 8 days ago, and then didn't notice that Github was showing the next push above that message (I guess it displays it based on commit time). Please could I have a re-review?

stevecotton:
Force-pushed to resolve a conflict in the changelog

PR #5435, opened 2021-01-15
wmllint: Ignore location_id when checking for shroud and fog

hexagonrecursion:
The following line was causing a wmllint error because it contains _s:
```
Irs, book_start Icr^Ii, Irs
```
```
"../../data/campaigns/Descent_Into_Darkness/maps/07c_A_Small_Favor3.map", line 10: warning, fog or shroud in map file
```

### DIFF
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -2884,8 +2884,8 @@ def translator(filename, mapxforms, textxform):
                     else:
                         fields = [x for x in line]
                     outmap.append(fields)
-                    terrain_strings = (f.split()[-1] for f in fields)
                     if not maskwarn and maptype == 'map':
+                        terrain_strings = (f.split()[-1] for f in fields)
                         if any(re.search('_s|_f(?!me)', t) for t in terrain_strings):
                             print('"%s", line %d: warning, fog or shroud in map file' \
                                   % (filename, lineno+1))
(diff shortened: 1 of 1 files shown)


data/tools/wmllint
```diff
@@ -2884,10 +2884,12 @@ def translator(filename, mapxforms, textxform):
                     else:
                         fields = [x for x in line]
                     outmap.append(fields)
-                    if not maskwarn and maptype == 'map' and re.search('_s|_f(?!me)', line):
-                        print('"%s", line %d: warning, fog or shroud in map file' \
-                              % (filename, lineno+1))
-                        maskwarn = True
+                    terrain_strings = (f.split()[-1] for f in fields)
```
stevecotton, line 2887: I think this line should be moved inside the outermost conditional

stevecotton:
Thanks for fixing this. BTW, `book_start` is a location_id, not a note.

data/tools/wmllint
hexagonrecursion: It took me way too long to figure out which conditional you are referring to as "outermost". I though you meant this one
https://github.com/wesnoth/wesnoth/blob/6a0baa563f8b3a6aede975debd657e6402ec2af5/data/tools/wmllint#L2829
and was trying to understand why or how. But now I see that you were referring to
```py
if not maskwarn and maptype == 'map':
```

hexagonrecursion: Done

stevecotton (approved):
Looks good to me.

I think this should be merged as soon as the CI build is fixed (a unrelated problem with the vcpkg setup).

Pentarctagon:
Though it hasn't been done the last couple of days, python script changes really should be reviewed by Elvish-Hunter as well.

PR #5437, opened 2021-01-15
Quote an attribute value

hexagonrecursion:
~~1. According to this https://github.com/wesnoth/wesnoth/commit/6405507b2408fc79511c0210a7a39689d56a4b6f  `[message]` with `speaker=narrator` is supposed to have `image=wesnoth-icon.png`~~

2. https://github.com/wesnoth/wesnoth/blob/75271a824657b0be63369ed4fc55cf0cb3579429/data/tools/wmllint#L2583


Pentarctagon:
As of the discussion in #5164, it's questionable whether this is actually desirable.

hexagonrecursion:
Ok. This leaves only the quotes.

PR #5441, opened 2021-01-16
Use gcc-9 in the Scout runtime instead of gcc-5.

Pentarctagon:
The gcc-9 shipped with the Scout runtime forces the use of -static-libstdc++ which is what allows this to work. However, LTO is not yet able to be enabled since as of right now it simply crashes immediately with `free(): invalid pointer`.

---

Note 1: Information about the Scout-provided `gcc-9` comes from https://github.com/ValveSoftware/steam-for-linux/issues/7430#issuecomment-761068435
Note 2: The updated docker image has not yet been uploaded for this PR.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -169,8 +169,8 @@ jobs:
           export IMAGE=steamrt
           export NLS=false
           export TOOL=scons
-          export CC=gcc-5
-          export CXX=g++-5
+          export CC=gcc-9
+          export CXX=g++-9
           export CXX_STD=14
           export CFG=release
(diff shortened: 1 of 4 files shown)


loonycyborg (approved):
Using newer gcc is really nice. Since we build boost manually anyway and that's our only C++ library dependency any C++ abi changes are easy for us to deal with.

Pentarctagon:
Alright, I'll get this merged later today then, and then open another PR for defaulting to C++17 afterwards.

Pentarctagon:
FYI I updated the docker image to the latest Scout runtime image as well.

Issue #5416, opened 2021-01-10
Terrain graphics background layer


doofus-01:
Using images on the off-map regions, like the new Isar's-Cross-in-the-sky, is nice, but could be better if we could use a solid JPEG background.  The problem is, base terrains are drawn on the absolute lowest layer, so you can't put anything under them.  There are two ways to get around this (without changing anything in the engine):
1.  carefully replace the "*" in the [terrain-graphics] map with "." wherever a desired base terrain graphic should be.
2. move the base terrain layer up one, to -999, and reserve layer=-1000 for backgrounds.

I think option 2 is the better one, but I'll need some time to make sure it doesn't cause some issue somewhere.  There may also be some other problem I'm not thinking of.

I'll try to get this in to 1.15 some time, starting with Isar's Cross.

soliton-:
Option 2 sounds better to me too.

doofus-01:
Maybe with #6254, this is worth looking at again.  (Flagging this for my future self.)

CelticMinstrel:
Why is -1000 the lowest layer exactly…?

doofus-01:
It's not, I mixed something up.  There was some other issue, but I don't remember.  It'll come back when/if I or anyone else works on this.

doofus-01:
The webp conversion makes whatever was valid here, if anything, very low priority.  If there is still anything to do around this, a fresh start wouldn't hurt.

PR #5450, opened 2021-01-16
Large JPG backgrounds

doofus-01:
Dealing with #5416


doofus-01:
Screenshot_20210116
With the core modifications above, I've got the Isar's Cross graphics "working" with <100 kB JPG background (off-map, the clouds) that has TOD variations, with one 1.2 MB border PNG, however as the attached image shows, the border PNG can't be used as-is this way, will need some redrawing.

Possibly more important than the file size, the background can be animated.  Using IPFs seems to be a bad idea, but a set of JPGs looks manageable.  It's still somewhat expensive, but it can be toggled like the whole fancy background is.

doofus-01:
Screenshot_20210117
Got it working with a modified cliff transition, can then add smaller images for the hanging earth, waterfalls, etc. (the wedges hanging down in the screenshot are placeholders)

doofus-01:
That last update makes Sevu/inferno8 version of Isar's the default, and simple version or this version other options.

doofus-01 (commented):
Was this really needed?  Layer -1000 doesn't seem to actually be a limit, was there something else preventing lower background images?

doofus-01:
There may be some assets here, but future work will need a fresh start.

PR #5451, opened 2021-01-17
Hide warnings from GCC 10 if compiling with LTO

sevu:
That way we could hide the warning where necessary, without specifying for all build steps.
Should we?


stevecotton:
Is this still necessary with Lua 5.3.6?

sevu:
Yes

loonycyborg (approved):

PR #5456, opened 2021-01-17
GUI2/Dispatcher: refactored [dis]connect_signal to use constexpr-if

Vultraz:
Opening as a PR for comments on the design, as well as waiting until we get the MacOS requirements sorted. Not 100% this is the correct use of `std::is_convertible`, but it does seem to work.

### DIFF
--- a/src/gui/core/event/dispatcher.hpp
+++ b/src/gui/core/event/dispatcher.hpp
@@ -17,10 +17,11 @@
 
 #include "gui/core/event/handler.hpp"
 #include "hotkey/hotkey_command.hpp"
-#include <functional>
 
 #include <SDL2/SDL_events.h>
 
+#include <cassert>
+#include <functional>
 #include <list>
 #include <map>
 #include <type_traits>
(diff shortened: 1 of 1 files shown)


stevecotton:
The error messages generated by both GCC and Clang are uninformative - they only hint at which .cpp file has the failing call:

```
In file included from src/actions/advancement.cpp:29:
In file included from src/gui/dialogs/unit_advance.hpp:16:
In file included from src/gui/dialogs/modal_dialog.hpp:18:
In file included from src/gui/core/static_registry.hpp:18:
In file included from src/gui/core/window_builder.hpp:19:
In file included from src/gui/widgets/grid.hpp:17:
In file included from src/gui/widgets/widget.hpp:17:
src/gui/core/event/dispatcher.hpp:526:4: error: static_assert failed "No matching signal queue found for event"
                        static_assert(false, "No matching signal queue found for event");
                        ^             ~~~~~
src/gui/core/event/dispatcher.hpp:570:4: error: static_assert failed "No matching signal queue found for event"
                        static_assert(false, "No matching signal queue found for event");
                        ^             ~~~~~
```

stevecotton:
> Opening as a PR for comments on the design

Will this check the callback function's signature at compile time, or will std::bind() manipulate the signature to fit and leave the error to be detected at runtime?

Vultraz:
For reference, this is why I couldn't use `static_assert(false)`: https://devblogs.microsoft.com/oldnewthing/20200311-00/?p=103553

src/gui/core/event/dispatcher.hpp
```diff
@@ -484,310 +485,98 @@ class dispatcher
...
+		} else if constexpr(is_notification_event(E)) {
+			VALIDATE_AND_ADD_TO_QUEUE(signal_notification)
+		} else if constexpr(is_message_event(E)) {
+			VALIDATE_AND_ADD_TO_QUEUE(signal_message)
+		} else if constexpr(is_raw_event_event(E)) {
+			VALIDATE_AND_ADD_TO_QUEUE(signal_raw_event)
+		} else if constexpr(is_text_input_event(E)) {
+			VALIDATE_AND_ADD_TO_QUEUE(signal_text_input)
+		} else {
+			assert(false && "No matching signal queue found for event");
```
stevecotton, line 528: Has this replaced a compile-time check with a runtime check?

stevecotton:
> For reference, this is why I couldn't use `static_assert(false)`: https://devblogs.microsoft.com/oldnewthing/20200311-00/?p=103553

https://en.cppreference.com/w/cpp/language/if includes an example of how to do it as a compile-time check, using dependent_false_v<T>.

PR #5459, opened 2021-01-18
Unit: converted upkeep handling to std::variant

Vultraz:
The upkeep_parser_visitor is still boost::variant-compatible since it's used by
const_attribute_value::apply_visitor, and configs still use boost::variant under the hood.

Additionally, these visitors are used in multiple places, so I can't convert them to lambas in-class.

Finally, I used brace-init for the upkeep types and visitors to make it clearer these are not functions.

### DIFF
--- a/src/actions/advancement.cpp
+++ b/src/actions/advancement.cpp
@@ -372,9 +372,11 @@ void advance_unit(map_location loc, const advancement_option &advance_to, bool f
 	std::vector<int> not_seeing = actions::get_sides_not_seeing(*u);
 
 	// Create the advanced unit.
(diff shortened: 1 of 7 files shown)


Vultraz:
Damn. Looks like we need to support MacOS 10.14 at minimum for this to work. :(

stevecotton:
I don't see any advantage here, over just using the Boost version for all platforms.

Vultraz:
@stevecotton Main advantage is that most of the work transitioning to the standard API is already done. Later on, one only needs to replace `utils` with `std`. It's also just better to use the STL when possible, IMO.

PR #5462, opened 2021-01-18
DiD: add wmllint: recognize Jaime

hexagonrecursion:
This PR silences a wmllint error (unknown 'Jaime' referred to by id) in DiD caused by an intentional scenario design choice.

wmllint checks that every unit referred to in [event], [message] etc is first introduced by [unit], [recall] etc. It does this to catch errors where a scenario author forgets to recall a unit. Jaime is spawned by an easter egg in 07a_A_Small_Favor. Jaime may be on the recall list for this scenario if the egg was triggered and if Jaime survives 07a, 07b and 07c. I assume that in this case the scenario author intentionally counts on the player manually recalling Jaime instead of using [recall].

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
@@ -565,6 +565,12 @@
         [/message]
     [/event]
 
+    # wmllint checks that every unit referred to in [event], [message] etc is first introduced by [unit], [recall] etc.
+    # It does this to catch errors where a scenario author forgets to recall a unit.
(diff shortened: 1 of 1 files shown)


stevecotton:
There's two CI failures - the first one is that some other files in DiD aren't passing the wmlindent test, and the second is th #5463 crash during the Boost tests.

```
make: Entering directory '/home/wesnoth-CI/data/tools'
wmlindent: ../../data/campaigns/Descent_Into_Darkness/scenarios/07c_A_Small_Favor3.cfg changed
wmlindent: ../../data/campaigns/Descent_Into_Darkness/scenarios/07b_A_Small_Favor2.cfg changed
make: Leaving directory '/home/wesnoth-CI/data/tools'
WML indentation check failed! (checkindent)
```

PR #5470, opened 2021-01-19
wmllint: Stop looking for wml tags inside [lua]

hexagonrecursion:
wmllint should not search for wml tags and keys inside [lua].
I caught it trying to add a translation mark in 23_Test_of_the_Clans inside a [lua] tag.

### DIFF
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -1896,6 +1896,7 @@ to be called on their own".format(filename, num))
     in_map_generator = False
     in_name_generator = False
     in_candidate_action = False
+    in_lua = False
     storeid = None
     storevar = None
     ignoreable = False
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Completely ignoring tags inside `[lua]` doesn't seem like the correct option, because there's the `[args]` subtag which can contain arbitrary data.

soliton-:
What checks do you expect wmllint to perform on arbitrary data?

CelticMinstrel:
It can still check that tags are balanced in arbitrary data, right?

soliton-:
Probably. I would not call that wmllint's strong suit though.

Elvish-Hunter:
> Completely ignoring tags inside `[lua]` doesn't seem like the correct option, because there's the `[args]` subtag which can contain arbitrary data.

Exactly. Instead of just ignoring everything that's inside a `[lua]` tag, a better approach might be still ignoring everything inside `[lua]`, but not if there's an `[args]` tag open.
Granted, it might also be possible for `[args]` to have another `[lua]` tag inside of it... In this case (which, I hope, should be extremely rare) the flag system currently used in that chunk of code won't be enough, and you'll have to at least keep count of how many `[lua]` and `[args]` are currently open.

soliton-:
I would not complicate wmllint's code on the off chance that it may detect an unbalanced tag in `[lua]`. We have actual parsers for that. Though if it's simple to do why not...

Elvish-Hunter:
> I would not complicate wmllint's code on the off chance that it may detect an unbalanced tag in [lua]. We have actual parsers for that. Though if it's simple to do why not...

Let's keep thing simple and skip the tag balancing part, then. My proposal is to add an `in_lua_args` flag that is appropriately set when finding an `[args]` tag. The `if in_lua:` line (1914) should then become `if in_lua and not in_lua_args:`, so wmllint can check the content of the `[args]` tags. Once this is done, I'll approve this PR.

hexagonrecursion:
I am no longer interested in continuing this PR. Anyone feel free to pick this up.

PR #5473, opened 2021-01-19
wmllint: fix textdomain errors

hexagonrecursion:
Ibanovic wants all #textdomain declrations in line 1: c10cf2b68b6

Several files were missing an explicit textdomain declaration.
Three of them actually have translated strings that defaulted to "wesnoth" textdomain:
- data/campaigns/World_Conquest/resources/data/training.cfg
- data/campaigns/World_Conquest/era/campaign/heroes.cfg
- data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg

### DIFF
--- a/data/campaigns/World_Conquest/era/campaign/heroes.cfg
+++ b/data/campaigns/World_Conquest/era/campaign/heroes.cfg
@@ -1,4 +1,4 @@
-##############################################################
+#textdomain wesnoth-wc
 
 #define WORLD_CONQUEST_II_ERA_HEROES_TRAITS
     [trait_extra]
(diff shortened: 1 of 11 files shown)


data/campaigns/World_Conquest/era/campaign/heroes.cfg
```diff
@@ -253,7 +254,6 @@
     [/trait]
 #enddef
 
-#textdomain wesnoth-multiplayer
```
soliton-, line 256: Surely this should stay here and the rest of the file be wesnoth-wc.

PR #5479, opened 2021-01-20
Add wmllint: display on/off where needed

hexagonrecursion:
Wrap wmllint display on/wmllint display off around several messages where manual line breaks are clearly inserted intentionally.


soliton-:
I'm not convinced this test is doing anything useful. I really doubt this ever caught an error. I'd rather remove it.

PR #5481, opened 2021-01-20
wmllint: remove check for nonstandard word-wrap style

hexagonrecursion:
> I'm not convinced this test is doing anything useful. I really doubt this ever caught an error. I'd rather remove it.

@soliton- in https://github.com/wesnoth/wesnoth/pull/5479#issuecomment-763696081

While I'm at it I also removed an orphaned comment.

### DIFF
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -122,11 +122,6 @@
 # Note that this will also disable stack-based validation on the span
 # of lines they enclose.
 #
-# You can suppress warnings about newlines in messages (and attempts to
-# repair them) with "wmllint: display on", and re-enable them with
-# "wmllint: display off".  The repair attempts (only) may also be
-# suppressed with the --stringfreeze option.
-#
 # A special comment "# wmllint: notecheck off" will disable checking unit types
 # for consistency between abilities/weapon specials and usage of special notes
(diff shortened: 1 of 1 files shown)


soliton-:
Thanks. Let's see if there are objections. If there are none also all display wmllint comments should be removed.

Elvish-Hunter:
By doing a git blame and looking at ESR's commit that introduced this check (15187c0), there were indeed two incorrectly placed newlines that were caught and fixed. I suppose this might've been useful because some programs make it hard to guess whether you actually typed a newline, it's the editor that wrapped the line or the editor placed a newline for you (this last behavior seems to be one of the possible options in vim).
Do you think that fixing an editor's behavior is a sufficient reason to keep this check?

Pentarctagon:
Not especially.

soliton-:
IMO that commit nicely shows how this check just means questionable edits on existing strings and plenty magic comments to turn the check off in many more places.

Elvish-Hunter:
Since nobody is interested in keeping this check, I'm going to merge this PR. However, as soliton correctly pointed out, all the `# wmllint: display on/off` magic comments will have to be removed from mainline content, since they won't do anything anymore.

PR #5491, opened 2021-01-22
Change mainline advancement modification

Pentarctagon:
The two commits explain their own reasoning.  Beyond that, the remaining question would be whether an advancement planning mod would:
* be a mainlining of the existing Plan Unit Advance modification; or
* be something brand new

The first option also of course depends on whether @vgaming is fine with their modification being mainlined or not.

### DIFF
--- a/data/lua/as_text.lua
+++ b/data/lua/as_text.lua
@@ -1,0 +1,62 @@
+-- NOTE: the string output from here is intended solely as an aid in debugging; it should never be taken and used as an input to anything else.
+
+-- escaping takes 3/4 of the time, but we can't avoid it...
+local function escape(str)
+	str = string.gsub(str, "%c", "")
+	str = string.gsub(str, "[\\\"]", "\\%0")
+	return str
+end
+
+local function add_table_key(obj, buffer)
+	local _type = type(obj)
+	if _type == "string" then
+		buffer[#buffer + 1] = escape(obj)
+	elseif _type == "number" then
+		buffer[#buffer + 1] = obj
+	elseif _type == "boolean" then
+		buffer[#buffer + 1] = tostring(obj)
+	else
+		buffer[#buffer + 1] = '???' .. _type .. '???'
+	end
+end
+
+local function format_any_value(obj, buffer)
+	local _type = type(obj)
+	if _type == "table" then
+		buffer[#buffer + 1] = '{'
+		buffer[#buffer + 1] = '"' -- needs to be separate for empty tables {}
+		for key, value in pairs(obj) do
+			add_table_key(key, buffer)
+			buffer[#buffer + 1] = '":'
+			format_any_value(value, buffer)
+			buffer[#buffer + 1] = ',"'
+		end
+		buffer[#buffer] = '}' -- note the overwrite
+	elseif _type == "string" then
+		buffer[#buffer + 1] = '"' .. escape(obj) .. '"'
+	elseif _type == "boolean" or _type == "number" then
+		buffer[#buffer + 1] = tostring(obj)
+	elseif _type == "userdata" then
+		buffer[#buffer + 1] = '"' .. escape(tostring(obj)) .. '"'
+	else
+		buffer[#buffer + 1] = '"???' .. _type .. '???"'
+	end
+end
+
+local function value_to_text(obj)
+	if obj == nil then return "null" else
+		local buffer = {}
+		format_any_value(obj, buffer)
+		return table.concat(buffer)
+	end
+end
+
+function wesnoth.as_text(...)
+	local result = {}
+	local n = 1
+	for _, v in ipairs({ ... }) do
+		result[n] = value_to_text(v)
+		n = n + 1
+	end
+	return table.concat(result, "\t")
+end
--- a/data/modifications.cfg
+++ b/data/modifications.cfg
@@ -1,5 +1,0 @@
-[modification]
-    id = "delay_advancements"
-    name = _ "Delay Advancements"
-    type = "hybrid"
-[/modification]
--- a/data/modifications.cfg
+++ b/data/modifications.cfg
@@ -1,0 +1,32 @@
+#textdomain wesnoth
+#ifdef MULTIPLAYER
+
+[modification]
+    id=plan_unit_advance
+    name=_"Plan Unit Advance"
+    description=_"When playing a multiplayer game, you do not control what a unit advances if they advances on another player's turn. With this modification you can set what your units advance to beforehand either for a specific unit or for all units of the same type."
+
+    [options]
+        [checkbox]
+            id="pickadvance_force_choice"  # WARNING: do not change this ID because other maps are relying on it
+            name=_"Force advancement planning"
+            description=_"You will be asked a question on choosing advancement whenever an undecided unit appears.
+
+Some eras and scenarios may automatically enable this option."
+            default=no
+        [/checkbox]
+    [/options]
+
+    [event]
+        name=preload
+        first_time_only=no
+        [lua]
+            code= {./modifications/pick_advance/dialog.lua}
+        [/lua]
+        [lua]
+            code= {./modifications/pick_advance/main.lua}
+        [/lua]
+    [/event]
+[/modification]
+
+#endif
--- a/data/modifications/pick_advance/dialog.lua
+++ b/data/modifications/pick_advance/dialog.lua
@@ -1,0 +1,129 @@
+-- << pickadvance_dialog
+
+pickadvance = {}
+local T = wesnoth.require("lua/helper.lua").set_wml_tag_metatable {}
+local _ = wesnoth.textdomain "wesnoth"
+
+local function filter_false(arr)
+	local result = {}
+	for _, v in ipairs(arr) do
+		if v ~= false then
+			result[#result + 1] = v
+		end
+	end
+	return result
+end
+
+
+function pickadvance.show_dialog_unsynchronized(advance_info, unit)
+-- dialog exit codes --
+	local reset_code = -3
+	local single_unit_code = -1
+	local all_units_code = 1
+--
+	local unit_type_options = advance_info.type_advances
+	local options = {}
+	for _, ut in ipairs(unit_type_options) do
+		options[#options + 1] = wesnoth.unit_types[ut]
+	end
+
+	local unit_override_one = (advance_info.unit_override or {})[2] == nil
+		and (advance_info.unit_override or {})[1] or nil
+	local game_override_one = (advance_info.game_override or {})[2] == nil
+		and (advance_info.game_override or {})[1] or nil
+
+	local description_row = T.row {
+		T.column { T.label { use_markup = true, label = _"Plan advance:" } },
+	}
+
+	local list_sub_row = T.row {
+			T.column { T.image { id = "the_icon" } },
+			T.column { grow_factor = 0, T.label { use_markup = true, id = "the_label" } },
+			T.column { grow_factor = 1, T.spacer {} },
+		}
+
+	local toggle_panel = T.toggle_panel { return_value = single_unit_code, T.grid { list_sub_row } }
+
+	local list_definition = T.list_definition { T.row { T.column { horizontal_grow = true, toggle_panel } } }
+
+	local listbox = T.listbox { id = "the_list", list_definition, has_minimum = true }
+
+	local reset_button = T.button {
+		return_value = reset_code,
+		label = _"Reset",
+		tooltip = _"Reset advancements to default"
+	}
+	local unit_button = T.button {
+		return_value = single_unit_code,
+		label = _"Save",
+		tooltip = _"Save the advancement for this unit only"
+	}
+	local recruits_subbutton = T.button {
+		return_value = all_units_code,
+		label = _"Save (all)",
+		tooltip = _"Save the advancement for all units of this type"
+	}
+	local recruits_button = not unit.canrecruit and T.row { T.column { horizontal_grow = true, recruits_subbutton } }
+
+-- main dialog definition
+	local dialog = {
+		T.tooltip { id = "tooltip_large" },
+		T.helptip { id = "tooltip_large" },
+		T.grid(filter_false {
+			T.row { T.column { T.spacer { width = 250 } } },
+			description_row,
+			T.row { T.column { horizontal_grow = true, listbox } },
+			T.row { T.column { horizontal_grow = true, unit_button } },
+			recruits_button,
+			T.row { T.column { horizontal_grow = true, (unit_override_one or game_override_one) and reset_button or T.spacer { width = 250 } } },
+		})
+	}
+
+-- dialog preshow function
+	local function preshow()
+		for i, advance_type in ipairs(options) do
+			local text = advance_type.name
+			if advance_type.id == game_override_one then
+				text = text .. _" (chosen, all)"
+			elseif advance_type.id == unit_override_one then
+				text = text .. _" (chosen)"
+			end
+			wesnoth.set_dialog_value(text, "the_list", i, "the_label")
+			local img = advance_type.__cfg.image
+			wesnoth.set_dialog_value(img or "misc/blank-hex.png", "the_list", i, "the_icon")
+		end
+
+		wesnoth.set_dialog_focus("the_list")
+
+		local function select()
+			local i = wesnoth.get_dialog_value "the_list"
+			if i > 0 then
+				local img = options[i].__cfg.image
+				wesnoth.set_dialog_value(img or "misc/blank-hex.png", "the_list", i, "the_icon")
+			end
+		end
+		wesnoth.set_dialog_callback(select, "the_list")
+	end
+
+-- dialog postshow function
+	local item_result
+	local function postshow()
+		item_result = wesnoth.get_dialog_value("the_list")
+	end
+
+	local dialog_exit_code = wesnoth.show_dialog(dialog, preshow, postshow)
+
+-- determine the choice made
+	local is_reset = dialog_exit_code == reset_code
+	local is_ok = dialog_exit_code >= single_unit_code and item_result >= 1
+	local game_scope = dialog_exit_code == all_units_code
+	return {
+		is_unit_override = is_reset or is_ok,
+		unit_override = is_ok and options[item_result].id or is_reset and table.concat(unit_type_options, ","),
+
+		is_game_override = is_reset or game_scope,
+		game_override = game_scope and options[item_result].id or nil,
+	}
+end
+
+-- >>
--- a/data/modifications/pick_advance/main.lua
+++ b/data/modifications/pick_advance/main.lua
@@ -1,0 +1,250 @@
+-- << pick_advance/main.lua
+
+local on_event = wesnoth.require("lua/on_event.lua")
+local T = wesnoth.require("lua/helper.lua").set_wml_tag_metatable {}
+local _ = wesnoth.textdomain "wesnoth"
+
+wesnoth.wml_actions.set_menu_item {
+	id = "pickadvance",
+	description = _"Pick Advance",
+	T.show_if {
+		T.lua {
+			code = "return pickadvance.menu_available()"
+		},
+	},
+	T.command {
+		T.lua {
+			code = "pickadvance.pick_advance()"
+		}
+	}
+}
+
+-- replace any non-alphanumeric characters with an underscore
+local function clean_type_func(unit_type)
+	return string.gsub(unit_type, "[^a-zA-Z0-9]", "_")
+end
+
+-- splits a comma delimited string of unit types
+-- returns a table of unit types that aren't blank, "null", and that exist
+local function split_comma_units(string_to_split)
+	local result = {}
+	local n = 1
+	for s in string.gmatch(string_to_split or "", "[^,]+") do
+		if s ~= "" and s ~= "null" and wesnoth.unit_types[s] then
+			result[n] = s
+			n = n + 1
+		end
+	end
+	return result
+end
+
+-- returns a table of the original unit types
+--         a comma delimited string containing the same values
+local function original_advances(unit)
+	local clean_type = clean_type_func(unit.type)
+	local variable = unit.variables["pickadvance_orig_" .. clean_type] or ""
+	return split_comma_units(variable), clean_type_func(variable)
+end
+
+-- replace the unit's current advancements with the new set of units via object/effect
+local function set_advances(unit, array)
+	wesnoth.add_modification(unit, "object", {
+		id = "pickadvance",
+		take_only_once = false,
+		T.effect {
+			apply_to = "new_advancement",
+			replace = true,
+			types = table.concat(array, ",")
+		}
+	})
+end
+
+-- for table "arr" containing sets of [index,unit_type]
+-- return table containing sets of [unit_type,true]
+local function array_to_set(arr)
+	local result = {}
+	for _, v in ipairs(arr) do
+		result[v] = true
+	end
+	return result
+end
+
+-- for table "arr" containing sets of [unit_type,true]
+-- return table containing sets of [index,unit_type]
+local function array_filter(arr, func)
+	local result = {}
+	for _, v in ipairs(arr) do
+		if func(v) then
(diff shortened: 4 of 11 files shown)


vgaming:
Hi, do I understand it correctly that we're considering the two commits above, PLUS adding some Lua/CPP source code to allow pre-selecting advances if they happen on enemy turn?

Depending on that, and having maintained the current "Plan Unit Advance" add-on, I could give some implementation/feedback details further.

Overall, I'm fine if wesnoth will make this feature internal (and thus make PUA obsolete). Removing the clutter and making people's playing experience smoother is definitely good by me :)

soliton-:
If you want to change advancement to default to the first one I would suggest to make sure that the first one is the most sensible choice.

Pentarctagon:
@vgaming What's being considered is either: (a) your modification as-is; or (b) a separate modification.  Implementing it in C++ wasn't something I was considering.

@soliton- I'll take a look.

vgaming:
@soliton- oh right, that makes a lot of sense! That's how the add-on should then work, I think.

@Pentarctagon All clear, all good :+1: 

Unrelated: not sure if we should randomize the advancement options for units at startup then :thinking: I guess we can return to this question when implementing the Lua part.

Pentarctagon:
> Unrelated: not sure if we should randomize the advancement options for units at startup then thinking I guess we can return to this question when implementing the Lua part.

That would effectively negate the change to not randomly choose an advancement - choosing a random advancement from a predefined list is the same as always choosing the first advancement from a randomly ordered list.

vgaming:
@Pentarctagon not exactly. First, an add-on can then conveniently set a preference for advancing if it happens on enemy turn while not preventing the user to make a choice if advance happens on their turn. Second, the randomization code can be moved from C++ to Lua, which I believe is a good thing because the CPP core can be more lightweight (hopefully more maintainable?). Third, but that is really minor, randomization at advance time is slightly different than randomization at recruit time if you reload your game (but again that can be disregarded).

vgaming:
@Pentarctagon and generally, this question is for later when if/after we accept the PR. It can obviously go both ways, the current PR doesn't prevent any future decision here.

Pentarctagon:
In any case I suppose, running `grep -Iirn "advances_to=.*,.*"` gives the results:
```
data/campaigns/Secrets_of_the_Ancients/units/Delinquent.cfg:15:              advances_to=Footpad,SotA Dark Adept
data/campaigns/Secrets_of_the_Ancients/units/Ruffian.cfg:8:                  advances_to=Footpad,Thug,SotA Dark Adept
data/campaigns/Under_the_Burning_Suns/units/undead/Ethereal_Ghost.cfg:15:    advances_to=EWraith,EShadow
data/campaigns/Under_the_Burning_Suns/units/quenoth/Fighter.cfg:20:          advances_to=Quenoth Warrior,Quenoth Flanker
data/campaigns/Under_the_Burning_Suns/units/quenoth/Scout.cfg:15:            advances_to=Quenoth Pathfinder,Quenoth Archer
data/campaigns/Under_the_Burning_Suns/units/quenoth/Mystic.cfg:16:           advances_to=Quenoth Shaman,Quenoth Sun Singer
data/campaigns/Under_the_Burning_Suns/units/quenoth/Tauroch_Rider.cfg:24:    advances_to=Tauroch Vanguard,Tauroch Stalwart
data/campaigns/Liberty/units/Villagers.cfg:182:                              advances_to=Huntsman_Peasant,Ranger
data/campaigns/Dead_Water/units/Citizen.cfg:14:                              advances_to=Merman Brawler,Merman Fighter,Merman Hunter
data/core/units/dunefolk/Rider.cfg:17:                                       advances_to=Dune Raider,Dune Horse Archer,Dune Sunderer
data/core/units/dunefolk/Herbalist.cfg:22:                                   advances_to=Dune Apothecary,Dune Alchemist
data/core/units/dunefolk/Skirmisher.cfg:18:                                  advances_to=Dune Strider, Dune Falconer
data/core/units/dunefolk/Soldier.cfg:17:                                     advances_to=Dune Swordsman,Dune Captain,Dune Spearguard
data/core/units/merfolk/Hunter.cfg:15:                                       advances_to=Merman Spearman,Merman Netcaster
data/core/units/merfolk/Warrior.cfg:14:                                      advances_to=Merman Triton,Merman Hoplite
data/core/units/merfolk/Initiate.cfg:15:                                     advances_to=Mermaid Priestess,Mermaid Enchantress
data/core/units/undead/Spirit_Ghost.cfg:14:                                  advances_to=Wraith,Shadow
data/core/units/undead/Skeleton.cfg:14:                                      advances_to=Revenant,Deathblade
data/core/units/undead/Necro_Dark_Sorcerer.cfg:15:                           advances_to=Lich,Necromancer
data/core/units/elves/Fighter.cfg:15:                                        advances_to=Elvish Captain,Elvish Hero
data/core/units/elves/Archer.cfg:16:                                         advances_to=Elvish Ranger,Elvish Marksman
data/core/units/elves/Shaman.cfg:16:                                         advances_to=Elvish Druid,Elvish Sorceress
data/core/units/nagas/Dirkfang.cfg:15:                                       advances_to=Naga Ophidian,Naga Ringcaster
data/core/units/humans/Mage.cfg:16:                                          advances_to=White Mage,Red Mage
data/core/units/humans/Horseman.cfg:14:                                      advances_to=Knight,Lancer
data/core/units/humans/Mage_Red.cfg:15:                                      advances_to=Arch Mage,Silver Mage
data/core/units/humans/Peasant.cfg:15:                                       advances_to=Spearman,Bowman
data/core/units/humans/Woodsman.cfg:15:                                      advances_to=Bowman,Poacher
data/core/units/humans/Outlaw_Ruffian.cfg:14:                                advances_to=Thug,Footpad
data/core/units/humans/Horse_Knight.cfg:15:                                  advances_to=Paladin,Grand Knight
data/core/units/humans/Woodsman_Trapper.cfg:14:                              advances_to=Huntsman, Ranger
data/core/units/humans/Loyalist_Spearman.cfg:15:                             advances_to=Swordsman,Pikeman,Javelineer
data/core/units/drakes/Burner.cfg:14:                                        advances_to=Fire Drake,Drake Flare
data/core/units/drakes/Clasher.cfg:13:                                       advances_to=Drake Thrasher,Drake Arbiter
data/core/units/trolls/Whelp.cfg:14:                                         advances_to=Troll,Troll Rocklobber
data/core/units/saurians/Augur.cfg:14:                                       advances_to=Saurian Oracle,Saurian Soothsayer
data/core/units/goblins/Wolf_Rider.cfg:27:                                   advances_to=Goblin Knight,Goblin Pillager
data/core/units/goblins/Spearman.cfg:14:                                     advances_to=Goblin Impaler,Goblin Rouser
```
I'll take a closer look today or tomorrow when I have time, but from glancing through the list there's nothing that currently stands out - the two I would have changed being the Whelp (if the Troll Rocklobber had been first) and the Spearman (if the Javelineer had been first), but those are already not the case.
Also @Hejnewar do you have any thoughts/opinions on this?

Hejnewar:
Not really, currently as I recall advancements are in the same order that they are in unit file and that is also good because it's an order that people are used to. At the same time PUA is clear enough that changing order shouldnt affect it too negativley.

Pentarctagon:
@soliton- The current advancements list looks good to me.

Pentarctagon:
@vgaming What's the history/purpose behind:
```
local function array_to_set(arr)
	local result = {}
	for _, v in ipairs(arr) do
		result[v] = true
	end
	return result
end

local function array_filter(arr, func)
	local result = {}
	for _, v in ipairs(arr) do
		if func(v) then
			result[#result + 1] = v
		end
	end
	return result
end

--- works as anti-cheat and fixes tricky bugs in [male]/[female]/undead variation overrides
local function filter_overrides(unit, overrides)
	local possible_advances_array = original_advances(unit)
	local possible_advances = array_to_set(possible_advances_array)
	local filtered = array_filter(overrides, function(e) return possible_advances[e] end)
	return #filtered > 0 and filtered or possible_advances_array
end
```
What sort of bugs were there and how does it work as an anti-cheat?

vgaming:
@Pentarctagon hi,

I don't remember about the undead bug, should've written it down. Undead was the key how I discovered the bug IIRC. A way to re-find the bug could be to try level zombies of different variations without this fix. One possible idea of the reason: maybe PUA would otherwise set advancements for your, say, "saurian zombie" upgrades to the one you specified for "human zombies".

I remember about anti-cheat. The reason for this check is how `overrides` type of data is constructed. It originates from a `synchronize_choice` function call here: https://github.com/vgaming/pick_advance/blob/master/pick_advance/lua/main.lua#L136
Now what if the user has overridden the contents of `synchronize_choice` locally, e.g. to set their units to advance to lvl5 dragon? Because it's a local choice, players would just get a String denoting the desired advance back (lvl5 dragon in this example). This would result in any unit being able to advance to any other.

I do realize that using this cheat would require:
* a software developer, because you have to read the code and understand these details
* knowing that the fact of cheating would be immediately obvious to any observer

Still, in such case, I wanted to prevent a bad image for wesnoth where a hacker could do literally everything.

vgaming:
Function `array_to_set` transforms a table like this:
`{ [1] = "a", [2] = "b", [3] = "c"}` (which could also be written in Lua as `{"a", "b", "c"}`)
into
`{ ["a"] = true, ["b"] = true, ["c"] = true }`

Function `array_filter` filters an array of values by using a function (predicate). For example, given an array
`arr = {1, 2, 3, 4, 5}`
and a function `func = function(x) return x > 2  end`
calling `array_filter(arr, func)` should give the result `{3, 4, 5}`.

Note that without syntactic sugar, array `{1, 2, 3, 4, 5}` is `{[1] = 1, [2] = 2, [3] = 3, [4] = 4, [5] = 5}` and array `{3, 4, 5}` is `{ [1] = 3, [2] = 4, [3] = 5}`.

Pentarctagon:
Alright, thanks.

And for the `on_event`s, what was the reason for setting their priority to -91?

vgaming:
@Pentarctagon I think I wanted to execute this piece of code early (earlier than many other add-ons code) https://wiki.wesnoth.org/LuaWML:Events#on_event.lua

I didn't make any tests suggesting that need. I'd consider removing it. Can do that in the already existing PUA to give it some time for feedback BTW. Done: https://github.com/vgaming/pick_advance/commit/41e7fffbfbdba1e04d003114fe56fe7d6b10e4a7

Pentarctagon:
I'm not really bothered whether they're there or not; I was just curious if it had been needed to solve some issue that had come up previously.

Pentarctagon:
Remaining questions:
* @vgaming I added a bunch of comments, so can you check and make sure they're correct?
* @CelticMinstrel Would the json_format functions be useful as a general utility rather than specific to this modification?
* @Vultraz Would you be able to review the UI of the advancement choice dialog?
---
Also, as far as translations go, how would that be handled?  There are quite a few places where string concatenation is done (though mostly involving newlines to increase button height), and otherwise there are two strings that are (I assume) able to be translated by the definition and use of `local translate = wesnoth.textdomain "wesnoth"` but otherwise no strings displayed on the UI are preceded by an underscore.

data/modifications/pick_advance/dialog.lua
```diff
@@ -0,0 +1,168 @@
...
+end
+
+
+function pickadvance.show_dialog_unsynchronized(advance_info, unit)
+-- dialog exit codes --
+  local help_code = -4
+  local reset_code = -3
+  local single_unit_code = -1
+  local all_units_code = 1
+--
```
vgaming, line 28: Nice idea to document dialog exit codes on top!
On a practical side, should we use tabs/spaces uniformly? These 4 lines are written with two spaces, below everything is in tabs.

data/modifications/pick_advance/dialog.lua
```diff
@@ -0,0 +1,168 @@
...
+-- dialog exit codes --
+  local help_code = -4
+  local reset_code = -3
+  local single_unit_code = -1
+  local all_units_code = 1
+--    
+	local label = "Plan advance:"
+
+	local unit_type_options = advance_info.type_advances
+	--print_as_json("advances for", unit.type, unit_type_options)
```
vgaming, line 32: TODO: delete this if `print_as_json` won't go into mainline

data/modifications/pick_advance/dialog.lua
```diff
@@ -0,0 +1,168 @@
...
+	local options = {}
+	for _, ut in ipairs(unit_type_options) do
+		options[#options + 1] = wesnoth.unit_types[ut]
+	end
+	local show_images = true
```
vgaming, line 37: Not sure what I intended to do here, but I never finished really. Let's remove this constant and its one usage below

data/modifications/pick_advance/dialog.lua
```diff
@@ -0,0 +1,168 @@
...
+	local unit_override_one = (advance_info.unit_override or {})[2] == nil
+		and (advance_info.unit_override or {})[1] or nil
+	local game_override_one = (advance_info.game_override or {})[2] == nil
+		and (advance_info.game_override or {})[1] or nil
+
+	local description_row = T.row {
+		T.column { T.label { use_markup = true, label = label } },
+	}
+
+	local list_sub_row
```
vgaming, line 48: With `show_images` inlined (see above), this would also be simplified.

data/modifications/pick_advance/dialog.lua
```diff
@@ -0,0 +1,168 @@
...
+	local toggle_panel = T.toggle_panel { return_value = single_unit_code, T.grid { list_sub_row } }
+
+	local list_definition = T.list_definition { T.row { T.column { horizontal_grow = true, toggle_panel } } }
+
+	local listbox = T.listbox { id = "the_list", list_definition, has_minimum = true }
+
+	local reset_button = T.button {
+		return_value = reset_code,
+		label = "\n" .. translate("Reset") .. "\n",
+		tooltip = "Reset advancements to default"
```
vgaming, line 70: I wonder if the quality of the addon is sufficient for Wesnoth's quality standards. I mean, this text is untranslated. I never figured out how to do that properly and efficiently. I think adding the addon to mainline would still be a clear improvement over the current situation, just making the choice explicit here.

data/modifications/pick_advance/dialog.lua
```diff
@@ -0,0 +1,168 @@
...
+	local is_ok = dialog_exit_code >= single_unit_code and item_result >= 1
+	--print(string.format("Button %s pressed (%s). Item %s selected: %s", dialog_exit_code, is_ok and "ok" or "not ok", item_result, options[item_result].id))
+	local game_scope = dialog_exit_code == all_units_code
+	return {
+		is_unit_override = is_reset or is_ok,
+		unit_override = is_ok and options[item_result].id or is_reset and table.concat(unit_type_options, ","),
+
+		is_game_override = is_reset or game_scope,
+		game_override = game_scope and options[item_result].id or nil,
+	}
```
vgaming, line 165: ! Important !
The current add-on was written with the intention of shrinking the possible advancement options into one (because otherwise wesnoth would randomize).

With the current PR (https://github.com/wesnoth/wesnoth/pull/5491), we could change the logic to *reorder* the possible advancement list so that user's choice comes first. The user would then still be able to manually pick an advance if it happens at their turn.

I've received comments from people that asked for the add-on to work this way already. I could try help doing this rewrite.

data/modifications/pick_advance/dialog.lua
vgaming: Oh sorry, I now see that this question was raised by @Pentarctagon already. Resolving it here to de-dup.

data/modifications/pick_advance/dialog.lua
Pentarctagon: Thanks, fixed - by default my text editor is set to use spaces instead.

data/modifications/pick_advance/dialog.lua
Pentarctagon: Done.

data/modifications/pick_advance/dialog.lua
Pentarctagon: Done.

data/modifications/pick_advance/dialog.lua
Pentarctagon: That sounds like a good idea to me.

Pentarctagon:
Updated to add translation support - @sevu could you double check whether this is correct?

Also, I've removed the help message since the same information is now shown on the tooltip when hovering over the respective buttons.

Pentarctagon:
The table-to-json functionality has been moved to mainline and renamed to `wesnoth.as_text(...)`.  A unit test has also been added to validate its output.

data/modifications/pick_advance/dialog.lua
```diff
@@ -0,0 +1,131 @@
+-- << pickadvance_dialog
+
+-- #textdomain wesnoth
```
sevu, line 3: Not needed, line 7 is enough.

data/lua/as_text.lua
```diff
@@ -0,0 +1,66 @@
...
+			buffer[#buffer + 1] = '":'
+			format_any_value(value, buffer)
+			buffer[#buffer + 1] = ',"'
+		end
+		buffer[#buffer] = '}' -- note the overwrite
+	elseif _type == "string" then
+		buffer[#buffer + 1] = '"' .. escape(obj) .. '"'
+	elseif _type == "boolean" or _type == "number" then
+		buffer[#buffer + 1] = tostring(obj)
+	elseif _type == "userdata" then
```
sevu, line 43: Should there be a case for translatable strings in the list? I think they had their own type or they are userdata.

data/modifications.cfg
```diff
@@ -1,5 +1,35 @@
+#ifdef MULTIPLAYER
+
 [modification]
-    id = "delay_advancements"
-    name = _ "Delay Advancements"
-    type = "hybrid"
+    id=plan_unit_advance
+    name="Plan Unit Advance"
+    description="In multiplayer game, you do not control unit advances if they happen on enemy turn.
+With this add-on, you can set them beforehand:
```
sevu, line 7: These two strings without _,
and the textdomain could be specified in line 1.

add-on -> extension?

data/modifications/pick_advance/main.lua
```diff
@@ -0,0 +1,249 @@
+-- << pick_advance/main.lua
+
+local on_event = wesnoth.require("lua/on_event.lua")
+local T = wesnoth.require("lua/helper.lua").set_wml_tag_metatable {}
+
+wesnoth.wml_actions.set_menu_item {
+	id = "pickadvance",
+	description = "Pick Advance",
```
sevu, line 8: needs _, and due to that the textdomain needs be declared too.

This file uses _ already for something else, but I think it does not clash.

sevu (commented):
Looks quite good, see the comments.

Wonder in which textdomain the strings should go. wesnoth-multiplayer? wesnoth-lib? wesnoth?

data/modifications/pick_advance/main.lua
vgaming: `_` is really not the best approach though, it means to mark a variable as unused in Lua... For example, for tools like `luacheck` it is the default convention and you don't get `unused variable` warnings for underscore variables. If you're really, really not interesting in an argument in Lua, you use `_`, just the underscore.

If I would use Lua best practices in Wesnoth, I'd name the function `translate` or `i18n` (or what's the better word from translation engine perspective?), e.g.:

```
local translate = wesnoth.textdomain "wesnoth"
local something = translate "Sword"
local unit = translate("Pikeman")
```

Pentarctagon: The convention for Wesnoth's codebase though is to use an underscore, since that mimics how it's done with WML translations.

data/lua/as_text.lua
Pentarctagon: According to [here](https://wiki.wesnoth.org/LuaAPI#Wesnoth_Types) they'd be userdata.

data/modifications/pick_advance/dialog.lua
Pentarctagon: Updated.

data/modifications.cfg
Pentarctagon: Updated.

data/modifications/pick_advance/dialog.lua
Pentarctagon: @vgaming to clarify - was this something you were planning to look into in the near future (within the next week or so)?  Either way is fine - there's no reason a second PR can't be made for updating the implementation - but if not I'm probably going to go ahead and merge this with the current implementation.

Pentarctagon:
I'll merge this in a few days barring any further comments/issues.

data/modifications/pick_advance/main.lua
sevu: On the Lua side in can be called anything. Using something else than _ is a matter of adding support in wmlxgettext to also recognize it.

vgaming: @sevu I'd even say wmlxgettext doesn't need to be changed at all, it's a matter of local variable name `_` or `translate` or else. I guess we can close this part of the discussion as offtopic though. I should raise a separate issue for that, if I'd want to change the current Lua guidelines in wesnoth

data/modifications/pick_advance/dialog.lua
vgaming: Hey, I'm struggling between really wanting to do it, and not having time for it at the moment.

To explain the scope at least:
* We need to make the add-on reorder the advancement list, not shrinking it. (Q: should we re-check if other addons added new stuff to units?)
* We probably could eliminate or change the menu options. Right now the possibilities/options are: "assign for this unit", "assign for this unit and all new recruits". With the deterministic changes in mind, however, this could always default to changing recruits as well. It could be the worded as something like "pick advance for Mage" and it would assign advance for all Mages unassigned yet, plus all new Mage recruits.

vgaming: Leaving it at the discretion of the PR author to change it or define the deadline. Both make sense.

Pentarctagon: I think I'd favor going ahead and merging it as-is so it'd be available for 1.15.10's release; changing the implementation to make use of reordering the advancements list instead would definitely be nice to have, but it's also a lower priority than getting the modification mainlined.

> should we re-check if other addons added new stuff to units?

I don't think it'd be necessary, unless you have an idea for it?

> We probably could eliminate or change the menu options. Right now the possibilities/options are: "assign for this unit", "assign for this unit and all new recruits". With the deterministic changes in mind, however, this could always default to changing recruits as well. It could be the worded as something like "pick advance for Mage" and it would assign advance for all Mages unassigned yet, plus all new Mage recruits.

I think that changing the advancement for just a single unit is probably something people would want to do.  Assigning it to every unit of that type where they haven't made an explicit choice would also somewhat go against one of the reasons for making advancements non-random: if they like the default, they don't need to change the advancement.

data/lua/as_text.lua
```diff
@@ -0,0 +1,66 @@
+-- NOTE: the string output from here is intended solely as an aid in debugging; it should never be taken and used as an input to anything else.
+
+local delete_chars = string.char(00, 01, 02, 03, 04, 05, 06, 07, 08, 09, 10, 11, 12, 13, 14, 15,
+	16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31) -- https://www.ietf.org/rfc/rfc4627.txt
+local delete_regexp = "[" .. delete_chars .. "]"
+
+-- escaping takes 3/4 of the time, but we can't avoid it...
+local function escape(str)
+	str = string.gsub(str, delete_regexp, "")
```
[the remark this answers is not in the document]
Pentarctagon: Updated to `"%c"`.  Performance shouldn't be an issue here I don't think since this is only meant to be used for debugging output, and it's not exceptionally slow as-is.

data/lua/as_text.lua
```diff
@@ -0,0 +1,66 @@
...
+	str = string.gsub(str, '\\', '\\\\')
+	return string.gsub(str, '"', '\\"')
```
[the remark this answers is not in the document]
Pentarctagon: Updated.

PR #5492, opened 2021-01-22
[WIP] Add missing spellings to the dictionary

hexagonrecursion:
currently wmllint produces a a ton of spelling warnings. I'm going through them and so far most have turned out to be false positives.


hexagonrecursion:
I'm not sure about this:

https://github.com/wesnoth/wesnoth/blob/c9c120481545b2177dfbd7fbeb273f83c6ba4e2e/data/core/units/dunefolk/Alchemist.cfg#L25

> However, in contrast to their brethrens’ focus on herbs for medicine and healing ...

Since plural is "brethren" shouldn't it be "brethren's"?

hexagonrecursion:
https://github.com/wesnoth/wesnoth/blob/c9c120481545b2177dfbd7fbeb273f83c6ba4e2e/data/core/units/orcs/Slurbow.cfg#L20
> often with a multi-ply arc of laminate wood or bone

Is "multi-ply" a real word?

hexagonrecursion:
https://github.com/wesnoth/wesnoth/blob/c9c120481545b2177dfbd7fbeb273f83c6ba4e2e/data/core/units/undead/Corpse_Necrophage.cfg#L18
> But the most revolting fact about these fratures ...

Is "fratures" a real word?

hexagonrecursion:
https://github.com/wesnoth/wesnoth/blob/b8d7b96d7b3213a52d7d35d681cabecc6e35331e/data/campaigns/Descent_Into_Darkness/units/Malin_Keshar.cfg#L253

Is "impetuity" a real word?

Pentarctagon:
I feel like this is probably another case where it'd be better to just remove the spellcheck feature from wmllint than add wmllint directives in all the hundreds of places it gives a false positive.

hexagonrecursion:
It did catch half a dozen or so misspellings. By that logic we should delete wmllint completely - in [my experience so far](https://github.com/wesnoth/wesnoth/pulls?page=2&q=+is%3Apr+author%3Ahexagonrecursion) the wast majority of its errors were false positives.

hexagonrecursion:
This PR is close to completion, but if it's likely that it won't be merged than I'm not motivated to continue working on it.

Pentarctagon:
In this case I'm wondering not because there are false positives, but rather because there's *so many* false positives.  So it's more a question of whether the spellcheck functionality is worth the overhead of needing to tell wmllint about all of them.

@soliton- , @Elvish-Hunter - any opinions?

soliton-:
I think this can certainly have more merit than the newline check. I'm not someone writing a lot of text in WML though. Perhaps @nemaara or other UMC content creators can weigh in.

The thing to remember with all the false positives in mainline is that wmllint is not only for mainline (where lots of people look over changes and actual issues will get caught anyway) but also for UMC.

Also IMO the recent effort to make mainline wmllint-clean should end in adding a wmllint check to CI to keep it this way for the future.

Pentarctagon:
Barring nemaara or someone else objecting, I suppose there's no harm in keeping the spellchecker given the future CI check then.

nemaara:
I'm not the right person to ask about this given that I don't use and have no clue what wmllint does.

hexagonrecursion:
@nemaara Wmllint indeed does to many things (IMHO). The functionality in question is spellchecking: when a PR containing changes to wml files is opened (or updated) all user visible text will be checked for misspelled words and all suspected misspellings will be reported reported. New words can be added globally, locally to a directory or locally to a file. When a character's name matches the unit's id (a very common case) this is also automatically recognized as a valid spelling. Do you think making the continuous integration automatically check the spelling of user-visible strings is a good idea? I'm even considering implementing a feature that makes it automatically post a comment with a list of all misspellings and links to the lines containing them.

Pentarctagon:
@nemaara The way it would affect you is that misspellings (whether actual misspelled words or false positives) would cause a CI failure to be reported until it's fixed, either by correcting the spelling (if a word is actually misspelled) or adding a wmllint directive so that wmllint knows of the word (this PR has numerous examples of this).

This PR does not add the CI check itself, however.

hexagonrecursion:
I am closing this as I have no intention of continuing working on this PR. I have cherry-picked just the typo fixes into a separate PR #5525

CelticMinstrel:
> So it's more a question of whether the spellcheck functionality is worth the overhead of needing to tell wmllint about all of them.

I think spellcheck functionality is useful to have, though it's a little unfortunate that it's not enabled on a default wmllint. I'm less sure about the _method_ of adding new spellings for Wesnoth, though.

It looks like a few of the words added could even be viewed as upstream issues with the spell-checker itself. For example, "wolves'll" should be seen as valid without any specific add, because "wolves" is in the dictionary and "'ll" is a valid clitic. Same applies to any other uses of contracted auxiliary verbs, and maybe also common prefixes like "un-".

Another odd one I noticed is "mayhap" - I think that should be "mayhaps"?

I didn't look through everything in this PR, so I don't know if there are more cases like that; perhaps I'll take a full look later. One thing I did notice is that a lot of the "local spellings" are in fact eye dialect, which is pretty unfortunate; I'd almost suggest it's worth having an separate dictionary just for that, to keep it self-consistent. Like, have a whole list of the "eye dialect" words somewhere and add a WML directive `# wmllint: import local spellings dwarf-dialect.txt`.

hexagonrecursion:
>  it's a little unfortunate that it's not enabled on a default wmllint.

It is enabled in wmllint by default. CI just does not run wmllint at all. There are hundreds of wmllint warnings in mainline. Mostly false positives in my experience.

CelticMinstrel:
> It is enabled in wmllint by default.

Only if pyenchant(?) is installed, which is not the case by default.

PR #5497, opened 2021-01-26
Legend of Wesmere: update Cliffs of Thoria difficulty

cooljeanius:
So, to be honest, I wanted to edit a lot more scenarios in Legend of Wesmere when I played thru it, but the fact that it's also playable in multiplayer kind of scared me off. I did manage to edit one scenario in it, though: Cliffs of Thoria. It's a pretty simple edit, just tweaking some turns/gold/income/aggression values.

### DIFF
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter3/10_Cliffs_of_Thoria.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter3/10_Cliffs_of_Thoria.cfg
@@ -112,8 +112,8 @@
         side=5
         allow_player=no
         canrecruit=yes
-        {GOLD 220 300 380}
-        {INCOME 2 7 12}
+        {GOLD 200 300 400}
+        {INCOME 4 8 12}
         team_name=trolls
         user_team_name= _ "Enemies"
 #ifdef EASY
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter3/10_Cliffs_of_Thoria.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter3/10_Cliffs_of_Thoria.cfg
@@ -10,7 +10,7 @@
     experience_modifier=100
     name= _ "Cliffs of Thoria"
     {LOW_MAP 10_Cliffs_of_Thoria.map}
-    {TURNS 32 30 28}
+    {TURNS 36 30 28}
 
     {DEFAULT_SCHEDULE}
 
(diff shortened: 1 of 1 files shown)


Pentarctagon:
By all accounts I've heard, LoW is pretty terrible at being an MP campaign.  So IMO if you wanted to get really in depth with changing it, then (barring @nemaara saying otherwise) it'd be fine to turn it into a regular SP campaign at the same time.

cooljeanius:
> By all accounts I've heard, LoW is pretty terrible at being an MP campaign. So IMO if you wanted to get really in depth with changing it, then (barring @nemaara saying otherwise) it'd be fine to turn it into a regular SP campaign at the same time.

The thing is, the fact that it's playable in multiplayer affects the single-player design, too. The whole thing about having multiple separate leaders, each with their own recruits and recall lists, really only makes sense (at least to me) as a concession to the multiplayer mode. So I'd want to consider ripping that out, as well, if I were to remove the multiplayer mode. I don't think I do actually want to remove it, though, as I think I have actually seen people say they've had fun with it in multiplayer on Discord. So I'll just change this one scenario for now instead.

data/campaigns/Legend_of_Wesmere/scenarios/chapter3/10_Cliffs_of_Thoria.cfg
```diff
@@ -112,8 +112,8 @@
         side=5
         allow_player=no
         canrecruit=yes
-        {GOLD 260 320 380}
-        {INCOME 4 8 12}
+        {GOLD 220 300 380}
+        {INCOME 2 7 12}
```
nemaara, line 116: A little too drastic on the income change I think. You could make the gold values {200 300 400} here I guess, keeping the income as in the original.

nemaara:
I did a relatively recent rebalance on LoW, so I'm thinking it should be reasonably balanced (granted probably the easier difficulties can go a bit easier). Multiplayer is actually pretty simple to balance around, you can't simply divide the SP gold by the number of sides, you need to give a bit of extra gold since some interactions (notably leadership) don't really work between sides. Also, low gold values for MP is just not fun since people will only get 1-2 units each. 

I'm not comfortable removing the MP aspect of LoW, I'd rather that be a decision from Pentarctagon and/or the board.

data/campaigns/Legend_of_Wesmere/scenarios/chapter3/10_Cliffs_of_Thoria.cfg
cooljeanius: > A little too drastic on the income change I think. You could make the gold values {200 300 400} here I guess, keeping the income as in the original.

ok done in 241f2ca

Pentarctagon:
Well, I've already said I'd be fine with it being turned into an SP campaign :P

cooljeanius:
> Well, I've already said I'd be fine with it being turned into an SP campaign :P

"would be fine with" is different from "this is what we're going to do." And anyways that's outside the scope of this PR. Let's have a forum thread for discussing such a removal if it's something people are serious about.

PR #5500, opened 2021-01-29
CMake: avoid unnecessarily remaking all .mo files on each build

stevecotton:
All the .mo files had a dependency on the target directory, which meant that
every change to the timestamp of that directory caused all the files to be
remade; and each file remade caused a change to the directory's timestamp ...

Since CMake 3.0, "cmake -E make_directory" has been deprecated in favor of
"file(MAKE_DIRECTORY ...)", so use that instead of the custom command. This
means the directories are created when the cmake config is generated, instead
of during the subsequent make steps.

Note: the mo-update target which builds the .mo files is only run when the
build is configured with ENABLE_NLS=ON. Both before and after this change,
none of the .mo files are rebuilt when ENABLE_NLS=OFF.


Pentarctagon:
Minor nitpick - can the `file` command be put all on one line?

stevecotton:
> Minor nitpick - can the file command be put all on one line?

Done.

stevecotton:
@ivanovic please would you review this?

ivanovic:
Sorry, got no clue if this change makes sense or not. If you are confident about it, feel free to go ahead.

PR #5502, opened 2021-01-30
NR S13: make all player sides have turns in sequence

### LINKED ISSUE #5396: NR S13, please switch Hamels (AI) and Krashs (player) sides
 - 1.15.7
 - Steam
 - Windows 10

NR S13 is a huge scenario with about 9 sides, 3 of them being under control of the player. I would like for those 3 sides to take their turns one after another, without an AI side inbetween. That way the player can do his part and then leave for a couple of minutes until it's the players turn again. 
Right now it's like this: Player -> 5x AI -> Player -> AI -> Player
I'd like to have instead: Player -> 6x AI -> 2x Player

### Screenshots
grafik
grafik

hexagonrecursion:
Fixes #5396

was: Player -> 5x AI -> Player -> AI -> Player
now: Player -> 6x AI -> 2x Player

### DIFF
--- a/data/campaigns/Northern_Rebirth/scenarios/13a_Showdown.cfg
+++ b/data/campaigns/Northern_Rebirth/scenarios/13a_Showdown.cfg
@@ -314,8 +314,7 @@
         canrecruit=yes
         side=8
         controller=human
-        {GOLD 1200 1000 800}
-        {INCOME 30 20 10}
+        gold=400
         recruit=Dwarvish Steelclad,Dwarvish Thunderguard,Dwarvish Stalwart
         team_name=knalgans
         user_team_name=_"Alliance"
--- a/data/campaigns/Northern_Rebirth/scenarios/13a_Showdown.cfg
+++ b/data/campaigns/Northern_Rebirth/scenarios/13a_Showdown.cfg
@@ -306,7 +306,6 @@
     # STARTING_VILLAGES works due to defeat_condition & having a starting location
     {STARTING_VILLAGES 7 10}
 
-    # Allied AI
     [side]
         type=Dwarvish Lord
         id=Hamel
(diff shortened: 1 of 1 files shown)


hexagonrecursion:
I don't like how manual and error-prone this is. Is there a way to change the turn order without having to change the side number everywhere?

hexagonrecursion:
One other question: will this break saves and replays?

cooljeanius:
I don't think this is really an improvement... breaking up all the AI turns with player turns helps keep the player more engaged in that scenario, because otherwise the AI just takes too long.

nemaara:
> I don't think this is really an improvement... breaking up all the AI turns with player turns helps keep the player more engaged in that scenario, because otherwise the AI just takes too long.

Well I think this is a problem with NR's design more than an issue with side order. Anyway, beside the point. I'm wondering if it would be better to let the player actually play Hamel's side. Any thoughts?

soliton-:
> I'm wondering if it would be better to let the player actually play Hamel's side. Any thoughts?

If the AI is incapable of playing an allied side I'd rather take over and play it myself unless it's really important for story reasons or something.

cooljeanius:
> > I'm wondering if it would be better to let the player actually play Hamel's side. Any thoughts?
> 
> If the AI is incapable of playing an allied side I'd rather take over and play it myself unless it's really important for story reasons or something.

It's not that the AI is *incapable* of playing Hamel's side, it's just that it's boring to let it do so. When I last played NR the issue was actually that the AI played Hamel's side *too* well, defeating Rakshas before I could even get close to him with any of my other sides. 
(...which I guess could also be solved by making Hamel's side player-playable, as the player could just hold back with Hamel's side in that case)

soliton-:
I was referring to the highest difficulty where you have to rush to Hamel to not lose the game early.

nemaara:
Yea, I think let's make Hamel a player side instead of swapping the sides around. I think that's preferable anyway, regardless of the issue in question.

hexagonrecursion:
@nemaara done

nemaara:
Erm well you're going to have to make the side have less gold I think. More in line with what the other sides have, just slightly more since Hamel wasn't there the last couple scenarios.

hexagonrecursion:
Done

nemaara (approved):

Issue #1038, opened 2017-05-08
Add section in help explaining the use of the current set of hotkeys


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?4067)** by anonymous on 2005-04-15

Many players seem to live in ignorant bliss about the use of hotkeys, and this is aggravated by the fact that the MANUAL seems to be constantly out of date on this particular field.

IMO, this could be remedied by adding a help section going through and explaining in slightly wordier manner than the curt texts given in Preferences for what the hotkeys do.

Preferably this section should read off the current setting of hotkeys, and display that in help.

This section could be located under Gameplay.

(Reproduced on All)
Release: 0.9.0
Priority: 1 - Later
Severity: 2 - Minor

wesnoth-bugs:
**Modified on 2005-10-04**

**ott wrote:**

> See https://savannah.nongnu.org/bugs/index.php?func=detailitem&amp;item_id=12697 for Savannah history.

CelticMinstrel:
Waiting on #994.

singalen:
An option for this one.

As [wiki rightfully points out](https://wiki.wesnoth.org/HotKeysSystem#Future_plans), different keyboard layouts can have different hotkeys. This should be mostly covered by help localization.

Plus, Mac has different hotkey system. Would be nice to make the hotkeys help depend on current platform.

jostephd:
#994 has been replaced by #3653.

CelticMinstrel:
However, it would be possible to implement this without #3653, if someone wanted to.

jostephd:
Updated labels accordingly.

hexagonrecursion:
> the MANUAL seems to be constantly out of date

Solution: delete this information from the manual.

> IMO, this could be remedied by adding a help section going through and explaining in slightly wordier manner than the curt texts given in Preferences for what the hotkeys do.

Having to look up documentation is poor user experience. A better option would be to fit all the relevant info into the UI somehow.

Pentarctagon:
Would it be possible to instead just make the current hotkey rows in the preferences dialog have hover text with a more detailed description?  Having to go all the way from the preferences (when not sure what something does) to the help page (to read the description of what said thing does) and then back to the preferences would take far too much time especially if there are multiple preferences they want to check.

PR #5503, opened 2021-01-30
Delete the list of hotkeys from the manual

hexagonrecursion:
As pointed out in #1038 the list of hotkeys is too hard to keep up to date.


hexagonrecursion:
The following are currently not listed in preferences->hotkeys
- [ ] Escape (Exit game, exit menu, cancel message)

hexagonrecursion:
What precisely is the difference between
m              Message another player (in multiplayer)
and
Alt+m          Message everyone in the game (in multiplayer)

stevecotton:
> What precisely is the difference between "m Message another player (in multiplayer)" and "Alt+m Message everyone in the game (in multiplayer)"?

It unchecks the "to team only" checkbox, if I recall correctly.

src/hotkey/hotkey_command.cpp
```diff
@@ -115,7 +115,7 @@ std::array<hotkey_command_temp, HOTKEY_NULL - 1> master_hotkey_list {{
 	{ HOTKEY_CREATE_UNIT, "createunit", N_("Create Unit (Debug!)"), false, scope_game, HKCAT_DEBUG, "" },
 	{ HOTKEY_CHANGE_SIDE, "changeside", N_("Change Side (Debug!)"), false, scope_game, HKCAT_DEBUG, "" },
 	{ HOTKEY_KILL_UNIT, "killunit", N_("Kill Unit (Debug!)"), false, scope_game, HKCAT_DEBUG, "" },
-	{ HOTKEY_PREFERENCES, "preferences", N_("Preferences"), false, scope_game | scope_editor | scope_main, HKCAT_GENERAL, "" },
+	{ HOTKEY_PREFERENCES, "preferences", N_("Go to Preferences menu"), false, scope_game | scope_editor | scope_main, HKCAT_GENERAL, "" },
```
stevecotton, line 118: Maybe "Open Preferences Screen"?

I think the 'm' in 'menu' should be a capital letter.

src/hotkey/hotkey_command.cpp
```diff
@@ -276,7 +276,7 @@ std::array<hotkey_command_temp, HOTKEY_NULL - 1> master_hotkey_list {{
 	{ HOTKEY_USER_CMD, "command", N_("Enter User Command"), false, scope_game, HKCAT_CHAT, "" },
 	{ HOTKEY_CUSTOM_CMD, "customcommand", N_("Custom Command"), false, scope_game, HKCAT_CHAT, "" },
 	{ HOTKEY_AI_FORMULA, "aiformula", N_("Run Formula"), false, scope_game, HKCAT_DEBUG, "" },
-	{ HOTKEY_CLEAR_MSG, "clearmessages", N_("Clear Messages"), false, scope_game, HKCAT_CHAT, "" },
+	{ HOTKEY_CLEAR_MSG, "clearmessages", N_("Clear Chat"), false, scope_game, HKCAT_CHAT, "" },
```
stevecotton, line 279: It does also clear WML and Lua error messages that have been shown at the top-left of the screen.

Pentarctagon:
F5 does, or definitely should, work - it's used to update the WML loaded by the game from the main menu rather than needing to restart Wesnoth.

stevecotton:
Instead of deleting this section, I'm thinking we should delete whole file, after a last check that everything in it is already in the in-game help.

hexagonrecursion:
> > What precisely is the difference between "m Message another player (in multiplayer)" and "Alt+m Message everyone in the game (in multiplayer)"?
> 
> It unchecks the "to team only" checkbox, if I recall correctly.

* I do not see a checkbox.
* In my testing (1.15.8) all three send a message to every player. Will try reproducing on master

stevecotton:
Urgh, just wasted an hour or so trying to work out when that was removed, only to find that the checkbox is only shown in MP games when there's another player on the same team who's connected via the network. If the checkbox isn't shown then the message is sent to all players, no matter which shortcut was used for opening the text input prompt.

That's controlled by the `has_friends()` call here, and it's been this way since Wesnoth 1.1.3 (circa 15 years ago):
https://github.com/wesnoth/wesnoth/blob/a9247060abd63203a17fedd73366ef1e55c5a0a8/src/menu_events.cpp#L205-L212

CelticMinstrel:
One think you might be missing about F5 is that it only works at the title screen (and I think in the map editor as well). It doesn't work during a game.

The Escape key can't be rebound, as I recall, so perhaps it alone deserves special mention in the game manual. (And I don't support deleting the game manual, for the record.)

I'm not sure about 1-7… it could be possible to allow customizing them, but that'd be kinda a pain since it's seven (or maybe even nine?) different hotkeys, one for each possible range. There could maybe be a new mechanic for exposing this information instead, possibly involving the scroll wheel or something.

Maybe "Scroll to Leader" would be better than "Center on Leader"?

hexagonrecursion:
> Maybe "Scroll to Leader" would be better than "Center on Leader"?

Done.

> I don't support deleting the game manual

I'd like to focus on this one section for now.

stevecotton:
I think this is good and should be merged, but feel I'm the wrong person to review it. I'm only listed as a reviewer because I previously left review comments.

Pentarctagon:
@hexagonrecursion can you squash the cpp changes?

hexagonrecursion:
Done

Pentarctagon:
The change itself looks pretty straight forward, so I'll merge this tomorrow unless there are objections.

PR #5512, opened 2021-02-03
fix codeblock projectfiles and update instructions

newfrenchy83:


projectfiles/CodeBlocks/README.md
```diff
@@ -9,18 +9,16 @@
...
     checkbox for this when choosing components to install).
 
-    NOTE: You must make sure to download the 32-bit version.
-    Building a 64-bit Wesnoth executable on Windows is currently not supported and will
-    fail with the SDK package provided in the next step.
+    NOTE: You must make sure to download the 64-bit version.
+
 	
-	To resolve out of memory errors when creating debug builds, follow these steps:
+	if you download 32 bit version and what you have memory errors when creating debug builds, you can follow these steps:
```
Pentarctagon, line 21: "if you download 32 bit version and what you have"
should be
"If you download the 32 bit version and you have out of"

projectfiles/CodeBlocks/README.md
```diff
@@ -37,6 +35,8 @@
 	
 	C:\masm32\bineditbin.exe /LARGEADDRESSAWARE cc1plus.exe
 	```
+	
+	if you use 64-bit version, then you cannot use MASM32, who don't have 64 bit version, and you cannot make debug build.
```
Pentarctagon, line 39: "if you use 64-bit version, then you cannot use MASM32, who don't have 64 bit version, and you cannot make debug build."
should be
"If you use the 64-bit version then you cannot use MASM32 since it doesn't have a 64 bit version, which means you cannot make a debug build."

projectfiles/CodeBlocks/README.md
newfrenchy83: i's done

PR #5513, opened 2021-02-03
change User to Custom to unify the help with the game

celerini:
I see no "User Maps" under the list box when loading previously saved map, but the "Custom Maps" appears here.


stevecotton:
Thank you for fixing this, and my apologies for adding this bug.

It's also been a valuable lesson to me about the amount of time that I keep unmerged branches - it seems I originally wrote these docs back in 2017.

celerini:
Thanks for your thanks (you're welcome). I hit it when translating.

Meanwhile I almost finished translation of the wesnoth-help domain and noticed two other things:
- the message "To load a map..." doesn't appear in any po file (results in one untranslatable sentence in Scenario Editor)
- the "Named Areas" is altogether missing in 1.15.9, nor dst='editor_named_area' is clickable in ToD... section, so I can translate it but it's not to player's profit

Issue #5519, opened 2021-02-05
[unstore_unit] doesn't work with location_id


Hejnewar:
### Describe the bug
[unstore_unit] doesn't work with location_id if location_id is provided unit will be unstored in recall list instead of place on the map regardless of its stored position. 

### Expected behavior
Unit is unstored in correct location.

Pentarctagon:
Can you provide a code example where this happens?

Hejnewar:
```
#define UNSTORE_LOC ID LOC
    [unstore_unit]
        variable = unit_{ID}
        find_vacant = yes
        location_id = {LOC}
    [/unstore_unit]
    {CLEAR_VARIABLE unit_{ID}}
#enddef
```

doofus-01:
That doesn't look like a very complete example.

Hejnewar:
```
    [event]
        name = moveto
        first_time_only = no
        
        [store_unit]
            [filter]
                find_in = unit
            [/filter]
            variable = stored_unit
            kill = yes
        [/store_unit]
        [unstore_unit]
            variable = stored_unit
            location_id = 1
        [/unstore_unit]
        [if]
            [have_unit]
                find_in = unit
                [not]
                    x,y = recall,recall
                [/not]
                count = 0
            [/have_unit]
            [then]
                [inspect]
                [/inspect]
                [recall]
                    find_in = unit
                    location_id = 2
                [/recall]
            [/then]
        [/if]
    [/event]
```

How to use: put in any scenario with defined starting positions for sides 1 and 2

PR #5521, opened 2021-02-06
Fixed unstore_unit lua code (issue #5519)

MrTitainin:
x,y were both binding to whole table instead of their expected values
#5519 should now be solved, although there still might be problem with location_id in teleport code

### DIFF
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -839,7 +839,8 @@ wml_actions.unstore_unit = function(cfg)
 	local x = cfg.x or unit.x or -1
 	local y = cfg.y or unit.y or -1
 	if cfg.location_id then
-		x, y = wesnoth.special_locations[cfg.location_id]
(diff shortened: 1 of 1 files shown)


Hejnewar:
Tested - works fine for me.

soliton-:
Can you fix the call for teleport a couple lines down as well?

CelticMinstrel:
I see, whoever wrote that (was it me?) must've thought that special_locations returns two values rather than a table. This fix seems fine, if a little ugly. Could also use … I think it's table.unpack?
```
x, y = table.unpack(wesnoth.special_locations[whatever])
```
It doesn't really matter, though.

MrTitainin:
This should address both your concerns.

Vultraz:
Just to confirm, `table.unpack` worked fine?

Hejnewar:
@Vultraz I tested that version too and I didn't notice anything wrong with it so I assume that it works fine.

PR #5523, opened 2021-02-07
Show a two-color orb for the "disengaged" state (#5155)

### LINKED ISSUE #5155: Orb color for UtBS's disengage ability (unit can move but can't attack)

stevecotton:
As the disengaged state is part-way between the "partial" and "moved" states,
the orb has parts in each color. On the minimap these units are shown in the
partial color (which is also the color that would be used before this change).

This will match the mounted Quenoth units' "disengage" skill, when they
can still move but can't attack. It should also trigger for some UMC abilities
that get extra moves after a character attacks.

During testing, I found that TSG allows some of the bandits to attack on the
first turn of the bandit branch. There's no gameplay change there, but the orbs
make it much clearer that some units can still attack.

I think there are already too many preferences for orbs, so reused the existing
settings for the colors. A new "show disengaged orb" preference is added, which
when disabled shows the old partial orb instead.

Update the orb and ellipse sections of doc/manual/.

Closes #5155.


doofus-01:
Can the red part be darker?  I think the change is a bit too subtle, judging from the help image.  Sorry I don't remember how to test a PR and I won't figure it out right now, but something like this image might work?
orb-two-color

stevecotton:
Like this, or should it go all the way down to match the full-red orb? While there are no units with crowns and disengage in UtBS, I'm worried that a crown would cover up most of the non-red part, and I generally think it's a feature that it's a bit brighter than the orb for units that can't move.

wesnoth_disengage_darker_red

The image in the help is a little off - I cut&pasted the orb with UtBS's sunlight onto the right color grass to fit with the other orbs' images.

doofus-01:
I would have said all the way to full-red darkness, but that's a good point about the crowns covering up the non-red...

I think we do want to have partially dark, to say "partially stopped/empty", and high contrast would make it easier to distinguish at a glance.  But maybe the boundary needs to be somewhere else.

This discussion is forcing me to notice that the yellow orb and green orb look exactly the same.  I don't play competitive, so I don't care, but if ~5% of people also have some colorblindness issue, someone will care, and that was what the customizable colors were for.  Or at least one of the reasons.  So, I think there should be something in addition to changing red/green colors, for example a harsher light-dark map.

stevecotton:
I'll try it with the color boundary cutting both the light and the dark parts of the orb, instead of having the color boundary aligned with the light boundary.

For colorblind support, maybe we should have an option to show icons instead of orbs - the code for that would be similar to the code for showing the two-color orb.

stevecotton:
wesnoth_disengage_tennis_ball

A tennis-ball inspired seam.

Notes about the image creation:
* create a color range `to_ellipse_red` with `rgb=FF0000,FF0000,000000,FF0000`
* `wesnoth --render-image 'misc/orb.png~RC(magenta>to_ellipse_red)' images/misc/orb-ellipse-red.png`
* open the `orb.png` and `orb-ellipse-red.png` images as layers in Gimp, add a layer mask to both of them
* use the layer mask to get each pixel from exactly one of the layers

doofus-01:
The last image looks good to me.  An option to change the orbs to little icons or something else simple and not color-centric sounds like it could work.

stevecotton:
Thanks for the feedback. I think it's ready to merge, but will leave it open a couple more days - there's no rush for this.

nemaara:
Personally I think the half colored orbs look kind of funky, but I don't currently have a better suggestion and this is something for doofus to decide anyway.

The presence of a disengaged orb is fine with me, so long as we're consistent with when it's displayed (i.e. if a unit can move but has no attacks left, the disengaged orb is always displayed, so for example it applies to both UtBS and TSG). I think this seems to be the case with what you have here so I don't have any other comments.

nemaara (approved):

stevecotton:
@Lordbob there's an `orb@2x.png` which I haven't created an `orb-two-color@2x.png` companion for; are the `@2x.png` images still used?

Latest push is just because I forgot to run `optipng`, and looking at the log for `orb@2x.png` reminded me to do it.

doofus-01 (approved):
I took this for a test-drive and it seems good to me.  The image looks fine for the current default colors, but if that needs to be revised in the future, it shouldn't be a problem.

cooljeanius:
> wesnoth_disengage_tennis_ball
> 
> A tennis-ball inspired seam.

Is it just me or does the red half look bigger than the yellow half in this image?

Zireael07:
@cooljeanius: Perspective fooling you imho.

CelticMinstrel:
Did anyone try this with a loyal disengaged unit? I wonder if the yellow is still visible then. Or what about a disengaged leader?

stevecotton:
> Did anyone try this with a loyal disengaged unit? I wonder if the yellow is still visible then. Or what about a disengaged leader?

You can test that by:
* moving a unit (so it's not a green orb)
* `:debug`
* `:unit attacks_left=0`

wesnoth_disengage_swamp_excursion

AI0867:
Does this also work in reverse for units that have no moves left but can still attack?

There is a similar conceptual difference between a unit that has moved partially and can attack and a unit that has moved fully and can attack.

CelticMinstrel:
I also mentioned that, in the corresponding issue thread.

AI0867:
I don't really see it, so I did some testing, and got the following results.

    moves\attacks | 0   | some   | all
    --------------+-----+--------+-----
    0             | red | orange | orange
    some          | O/R | orange | orange
    all           | O/R | orange | green

stevecotton:
I've done a prototype for an "engaged" orb, it's autolinked above.

I'm not completely opposed to the idea, but feel the engaged state is something that's already visible on the map without needing an extra orb - a unit with a partial orb that's adjacent to an enemy is in the engaged state (with the edge case that it's a non-skirmisher unit). With the disengage state, it's typical to move the unit out of the way so that another unit can attack, and so without the disengage orb there's no visual clue of whether the unit has already attacked.

AI0867:
Going by the code, that would result in the following table?

```
moves\attacks | 0   | some   | all
--------------+-----+--------+-----
0             | red | R/O    | R/O
some          | O/R | orange | orange
all           | O/R | orange | green
```
I guess we can interpret this is orange/red being independent, but green being the "unit fully fresh" state.


One could consider (though it would be a much bigger change) to fully decouple the two, giving us the following grid:

```
moves\attacks | 0   | some   | all
--------------+-----+--------+-----
0             | red | R/O    | R/G
some          | O/R | orange | O/G
all           | G/R | G/O    | green
```

PR #5524, opened 2021-02-08
Edit Scepter of Fire difficulty

cooljeanius:
This is my attempt to apply patches from [my general mods collection](https://github.com/cooljeanius/wesnoth_mods/tree/master/campaigns/Sceptre_Of_Fire). Basically, I needed more turns in Gathering Materials, and I needed to make it easier for Alanin to escape in Towards the Caves. Tested on 1.14, not on 1.15 though.

### DIFF
--- a/data/campaigns/Sceptre_of_Fire/scenarios/4_Gathering_Materials.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/4_Gathering_Materials.cfg
@@ -206,13 +206,6 @@
         {FLAG_VARIANT knalgan}
         team_name=dwarves
         user_team_name= _ "Dwarves"
-        # When I first edited this scenario I didn't realize that
-        # the default gold when left unspecified is 100, so my specification
-        # of {GOLD 50 30 10} was actually an unwitting reduction...
-        # I am fixing that by just adding an extra hundreds digit, but that
-        # is untested...
-        {GOLD 150 130 110}
-        {INCOME 5 3 1}
     [/side]
     # these AI sides were set to not be on the same side. Assuming that is intentional, let's give them a little distinction and variety
     [side]
@@ -240,7 +233,7 @@
         recruit=Troll Whelp,Young Ogre,Troll Rocklobber
         team_name=trolls1
         user_team_name=_"Resident Trolls"
-        {GOLD 60 120 180}
+        {GOLD 75 125 175}
     [/side]
     [side]
 #ifdef HARD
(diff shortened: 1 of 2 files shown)


data/campaigns/Sceptre_of_Fire/scenarios/4_Gathering_Materials.cfg
```diff
@@ -233,7 +240,7 @@
         recruit=Troll Whelp,Young Ogre,Troll Rocklobber
         team_name=trolls1
         user_team_name=_"Resident Trolls"
-        {GOLD 75 125 175}
+        {GOLD 60 120 180}
```
stevecotton, line 243: Could this make the later levels harder instead of easier? On easy, this looks like at least 36 xp less available to the player (one less troll whelp per team, multiplied by three teams).

cooljeanius: There are more turns on which the player can grind them though

doofus-01:
I should just point out that SoF got major revisions in 1.15, so revisions to 1.14 might not carry directly.  SoF probably needs adjustments, but BfW 1.14 won't be the stable version past October if all goes according to plan.  So maybe efforts here should be directed to the dev version.

cooljeanius:
> I should just point out that SoF got major revisions in 1.15, so revisions to 1.14 might not carry directly. SoF probably needs adjustments, but BfW 1.14 won't be the stable version past October if all goes according to plan. So maybe efforts here should be directed to the dev version.

This pull request is against 1.15 and I got it to apply successfully.

data/campaigns/Sceptre_of_Fire/scenarios/4_Gathering_Materials.cfg
```diff
@@ -206,6 +206,13 @@
         {FLAG_VARIANT knalgan}
         team_name=dwarves
         user_team_name= _ "Dwarves"
+        # When I first edited this scenario I didn't realize that
```
nemaara, line 209: I just played S4, I'm pretty sure these changes aren't really in the direction I want to go with this scenario. I think I'd like to keep the lower turn limit and simply give you more gold if necessary on easier difficulties. How about I handle this scenario in my rebalance pass? S6 looks okay to me though.

cooljeanius: ok... so do you want me to revert the S4 changes then? Or can we keep them and just have you overwrite them when you do your own rebalance pass?

nemaara: I think it doesn't make sense to commit them to begin with if I'm going to overwrite them anyway so I think it's better to not include them in this PR.

cooljeanius: > I think it doesn't make sense to commit them to begin with if I'm going to overwrite them anyway so I think it's better to not include them in this PR.

ok reverted in c9f8a61

PR #5531, opened 2021-02-10
Make the preferences window big enough for HDPI fonts

### LINKED ISSUE #5185: Preferences window broken on high resolutions
Current master has a little bigger texts and buttons on high resolutions(bigger than 1200x900), so preferences do not fit in the 800x600 window anymore and needs at least 900x650 window.

Maybe insted of `GUI_WINDOW_FIXED_SIZE_CENTERED ` preferences window should use `GUI_WINDOW_PERC_FIXED_SIZE_CENTERED` ? For example:

```diff
#data/gui/window/preferences.cfg
[window]
	id = "preferences_dialog"
	description = "Main game preferences dialog."
	[resolution]
		definition = "default"

-		{GUI_WINDOW_FIXED_SIZE_CENTERED 800 600 }
+		{GUI_WINDOW_PERC_FIXED_SIZE_CENTERED 75 70 800 600 }
# Note: 75% of 1200 is 800. Height percent was chosen by checking  look ща preferences window on resolution of 1280x960
```

Resulting preferences window:
image

_Originally posted by @vgaming in https://github.com/wesnoth/wesnoth/issues/2219#issuecomment-662097214_
> Today-s `master` branch with 4k  (note the unnecessary scroll bars)
> Could be unrelated from 4k.
> 
>

stevecotton:
The hardcoded size of the window was too small for the large fonts on HDPI
displays, this changes to almost-automatic sizing. It leaves an upper limit for
the height of 80% of the main window size, so that there's a border at top and
bottom. Fixes #5185

Buttons that were placed at the bottom of pages move up so that they aren't
too far from the other controls. The sub-tab-selector buttons on the
multiplayer tab move above the sub-tabs themselves.

With HDPI text, the checkboxes are placed a little too high to align with
the font, but that's a cosmetic detail; this commit fixes the usability issue.

wesnoth_preferences_4k_comparison

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -25,6 +25,7 @@
    * Added a "disengaged" orb, shown instead of the partially-moved orb for units that can move but can't attack.
    * Added information about the build's (not runtime) target CPU architecture to the game version info dialog and --report.
    * Added terminal-style command history browsing with up-down keys for in-game consoles used by debug mode, ai and search floating textboxes.
+   * Made the preferences dialog larger on HDPI screens; this fixes a usability issue with HDPI font sizes (issue #5185).
  ### WML Engine
    * Extent 'special_id_active' and 'special_type_active' to abilities used like weapon and to [leadership] abilities.
    * abilities used like weapon can call [leading_anim] now.
--- a/data/gui/window/preferences.cfg
+++ b/data/gui/window/preferences.cfg
@@ -129,7 +129,8 @@
 	[resolution]
 		definition = "default"
 
-		{GUI_WINDOW_FIXED_SIZE_CENTERED 800 600}
+		# The hotkey list will always want to be as tall as possible, which causes a lot of blank space on the other tabs. Limit the height so that the "General" tab fits the content tightly with non-HDPI fonts, and looks reasonable with HDPI fonts.
+		maximum_height = "({GUI_SCALE_RESOLUTION 600})"
 
 		[linked_group]
 			id = "page_icon"
--- a/data/gui/window/preferences/01_general.cfg
+++ b/data/gui/window/preferences/01_general.cfg
@@ -225,6 +225,8 @@
 [layer]
 
 	[row]
+		grow_factor = 0
+
 		[column]
 			horizontal_grow = true
 			vertical_alignment = "top"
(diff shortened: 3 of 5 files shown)


soliton-:
Can't say much to the implementation but the result looks like a clear improvement.

Pentarctagon:
Why is there a ton of empty space on the bottom of the last two images?

stevecotton:
> Why is there a ton of empty space on the bottom of the last two images? it's definitely better, but it isn't the best look still.

The hotkeys tab would like as much height as it can get, and the extra space gets used on the advanced preferences tab too.

Pentarctagon:
Is there any way to shrink it for tabs that don't need as much space as they can get?

stevecotton:
> Is there any way to shrink it for tabs that don't need as much space as they can get?

Changing the `maximum_height` of the whole window based on the size of the main window is easy, but note the 1280x1024 hdpi image in my examples (the top middle one), which is the smallest resolution at which the HDPI fonts take effect. The General tab is already large enough that the Close button is partially cut off, so I wouldn't want the window to be smaller. (The screenshot shows the Display tab, but it's the General tab that's needing the space there.)

In GUI2 I can't see a way to set the maximum height of a listbox, nor any other maximum height setting for things other than the window. @Vultraz, is there something that I'm missing?

The extra space is probably less distracting for HDPI users - we're already used to GUIs where rows or columns fit in less of the screen than the program author expected.

shikadiqueen:
The massive extra space gives it a rather amateurish look, much like the short scrolling bars that show up in several circumstances. I'm sure there's a better approach for fixing this.

stevecotton:
I'm considering a maximum height based on font size instead of pixels, but would still like feedback from the GUI2 devs about whether there's a better way to stop the listboxes being the size-defining widgets.

stevecotton:
Still a similar approach, but tuned so that the dialog fits the content tightly with non-HDPI fonts, and there's less extra space with HDPI fonts. I've taken a brief look at the GUI2 layout mechanism, but decided that I don't want to take on the task of changing it.

wesnoth_preferences_4k_comparison_v2

stevecotton:
Having pinged about the latest images on IRC and waited for comments, I'm intending to merge this soon (after rebasing and letting the CI run).

PR #5535, opened 2021-02-13
Upgrade to Lua 5.4.2

Vultraz:
Upgrades from 5.3.6 to 5.4.2.


Vultraz:
@CelticMinstrel can you look at the change to lua_resume, see if the change is correct? https://www.lua.org/manual/5.4/manual.html#lua_resume

CelticMinstrel:
I'm going to look at the diff too, but for anyone else looking at this, be sure to check out these links.

https://www.lua.org/manual/5.4/readme.html#changes
https://www.lua.org/manual/5.4/manual.html#8

CelticMinstrel:
I think the lua_resume change is correct, but I think since we have this new information we should check that the correct number of results is returned… namely however many results the subsequent code expects.

mattsc:
From Discord/irc: The Lua 5.4 changes should be fine for our AI code.  We do use `math.random` (random number generation inside AI evaluation functions needs to be unsynced), but the change should be no problem.  And I don't remember the AI code using anything that would be affected by the rest.

mattsc (approved):

PR #5536, opened 2021-02-13
Use to-be-closed variables to scope WML variables in tag definitions

CelticMinstrel:
With this you no longer have to worry about forgetting to end the variable scope - Lua will do it for you automatically. This is a PR because I've only done a basic smoke test (making sure there are no errors when starting a game; for this test I chose Aethermaw). I didn't confirm that the variable is correctly set and unset, though I have no reason to expect it wouldn't be; I did do some testing to verify how to-be-closed variables work.

### DIFF
--- a/data/lua/wml-flow.lua
+++ b/data/lua/wml-flow.lua
@@ -111,7 +111,7 @@ wesnoth.wml_actions["for"] = function(cfg)
 		return
 	end
 	local i_var = cfg.variable or "i"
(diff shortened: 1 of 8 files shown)


stevecotton:
> I didn't confirm that the variable is correctly set and unset

Please add a WML-based unit test for this.

CelticMinstrel:
> Please add a WML-based unit test for this.

Done.

PR #4580, opened 2019-11-18
Refactor the game map to permit exposing it to Lua

CelticMinstrel:
Basically what this does is add `wesnoth.get_map()` which returns the same thing as the mapgen kernel's `wesnoth.create_map()` only referring to the actual map currently in use by the game.

I haven't had a chance to test it yet, but I'm opening it up for code review now.

Issue #4459, opened 2019-10-13
`[music] ms_after/ms_before=` poorly understood and used in a nonsensical fashion


shikadiqueen:
The `[music] ms_after=` and `ms_before=` options appear to be (incorrectly documented)[https://wiki.wesnoth.org/MusicListWML], and they are also used in mainline in ways that don't seem to make any sense.

### `ms_before`

If I take a look at `DEFAULT_MUSIC_PLAYLIST` (which seems to be the only thing Secrets of the Ancients ever uses as of version 1.14.9), all tracks have `ms_before=12000` set (that's 12 seconds), which doesn’t make any sense from an artistic standpoint even if `ms_before` did whatever it is the author expected it to do.

Now, what `ms_before` does do in practice is **not** whatever the wiki says — if you use `immediate,play_once=yes,yes` to play a track that's **shorter** than `ms_before`, then the game will wait the difference between the next track's `ms_after` and the current track's length before playing the next track. The only way to observe this with `DEFAULT_MUSIC_PLAYLIST` is playing a track shorter than 12 seconds, for example by overwriting `silence.ogg` with a 2 seconds long file:

* `:lua wesnoth.wml_actions.music { name = "silence.ogg", immediate = true, play_once = true }`

If you allow `DEFAULT_MUSIC_PLAYLIST` to advance naturally instead, you'll notice there are no artifical gaps between tracks. When a track ends, the next one plays, no extra delays involved. This seems to make sense to me other than the fact that the macro has that useless `ms_before` specified for all tracks.

### `ms_after`

This one is probably going to be a point of contention. Whatever the wiki says about it is false. What actually happens in practice is that if a music track is forcefully switched using `immediate=yes`, there will be a fade-out for `ms_after` milliseconds (or presumably less if the remaining duration of the track is less than `ms_after`). This makes `immediate=yes` nigh useless in situations where `ms_after` is specified for the currently-playing track — it might as well be called `immediate_but_wait` instead (see bug #4458).

[...]

Issue #4458, opened 2019-10-13
[music] immediate=yes may block waiting for the previous track to fade out


shikadiqueen:
In Wesnoth 1.14, `[music] immediate=yes` doesn't ignore the previous track's `ms_after` setting, which is often set to 2000 milliseconds for the first track in a scenario's initial playlist because of the `SCENARIO_MUSIC` macro (for some reason). This means that it may block for 2 seconds waiting for the previous track to be faded out first.

This is not ideal since it's a change from 1.12 and other previous stable series where the change was decidedly immediate, and affects macros such as `INCIDENTAL_MUSIC` where the new track is expected to play instantly.

*Steps to reproduce:*

In the in-game console:
* `:debug`
* `:lua wesnoth.music_list.current.ms_after = 2000; wesnoth.wml_actions.music { name = "silence.ogg", immediate = true }`
* `silence.ogg` is provided as an example. Any valid name different from the current track's should work.

shikadiqueen:
Do note that after further testing for the purpose of submitting #4459, fixing this bug would render `ms_after` entirely meaningless. As far as I'm concerned that's not necessarily an issue since it's currently documented and used as if it did something that it doesn't do. If anything the issue is that it does anything at all and that there's no way to nullify it for one of the most common use cases of `immediate=yes`.

Issue #4030, opened 2019-04-18
Music ms_after hangs scenario story screen


slavrenyuk:
I will describe 2 buggy use cases below. The bug only affects story screen, during a scenario `ms_after` works well.

**a) ms_after setting for music tag directly under campaign tag hangs story starting**
Steps to reproduce
1) Go to `<Wesnoth directory>/data/core/macros/sound-utils.cfg`
2) In `SCENARIO_MUSIC` macro add `ms_after=10000` for intro music "revelation.ogg"
3) Launch the game, start "Son of the Black-Eye" campaign (it uses that macro)
4) There is a 10 second lag before you can move through the story pages

**b) ms_after setting for music in a prestart event hangs story ending**
Steps to reproduce
1) Go to `<Wesnoth directory>/data/core/macros/sound-utils.cfg`
2) In `SCENARIO_MUSIC` macro change `ms_after=2000` with `ms_after=10000`  for a prestart event music
3) Launch the game, start "Son of the Black-Eye" campaign (it uses that macro)
4) Finish reading the story (or skip it)
4) There is a 10 second lag before the scenario starts

jyrkive:
As unfortunate as it is, this is currently by design. Hence the enhancement label.

`Mix_FadeOutMusic()` blocks the main thread during the fade-out in this line: https://github.com/wesnoth/wesnoth/blob/aae617f7a83b274b2323fb493caa2cdfb78d1485/src/sound.cpp#L754

SDL_Mixer isn't documented to be thread safe, so we can't move that call into a worker thread.

The only way to keep fade-outs from hanging the main thread would be a full rewrite of the sound engine.

slavrenyuk:
Thanks for the quick reply.

jyrkive:
Sure thing. (It's easy to reply quickly because we get real-time notifications when new issues are filed, and I'm familiar with sound code.)

slavrenyuk:
`ms_before` doesn't cause similar bug because of this line? https://github.com/wesnoth/wesnoth/blob/9cccfd73b0128e467912fca3470d52265622bd15/src/sound.cpp#L814

There is still something strange about `ms_after`. If I set `ms_after` for the intro music, it hangs before the intro music starts to play. And if I set `ms_after` in the prestart event, it hangs before that music starts to play. Doesn't it mean that the sound engine takes `ms_after` of the wrong track?

jyrkive:
> `ms_before` doesn't cause similar bug because of this line?

[...]

PR #5537, opened 2021-02-13
Various Lua reorganization work

CelticMinstrel:
I'm going to use this pull request for most of the API reorganization, but #4580 covers the game map refactor, and some other larger bits may get their own pull request.

For now, I'm interested in feedback on whether these two new modules are even a good idea. The versions module is intended to replace `wesnoth.compare_versions` while the mathx module is used as a dumping ground for various math-ish functions that were in the Wesnoth API.


soliton-:
`mathx` seems like an obscure name. Why not something like `utils.math...` or `math.utils...` or so?

CelticMinstrel:
There is no pre-existing utils module, and `mathx` mirrors the already existing `stringx`, which is similar in that it also adds functions to a core Lua module.

soliton-:
`stringx` sounds similarly obscure to me. Is there some precedent for this in lua or did we just make that up?

CelticMinstrel:
I don't know of any precedent in Lua specifically. I think there's precedent somewhere for <module>x being an extended version of <module>, possibly in Java or something.

vgaming:
Hey, I've just encountered `stringx` randomly while reading code and playing around with Wesnoth codebase. What is `stringx` aimed for? Is it copied from anywhere? Can we add it to https://wiki.wesnoth.org/LuaWML before releasing?

I think it's essential to have documentation for those type of things

CelticMinstrel:
It's already fully documented at [LuaAPI](https://wiki.wesnoth.org/LuaAPI/stringx); all the documentation in the LuaWML namespace is being slowly migrated there as well.

vgaming:
@CelticMinstrel oh sorry, I forgot about this section. Thanks for explaining!

CelticMinstrel:
I'd like to get this merged pretty soon. Does anyone have any comments on my choices?

Pentarctagon:
Unit tests would be good to add.

src/filesystem.hpp
```diff
@@ -428,7 +428,7 @@ std::string get_short_wml_path(const std::string &filename);
  *   units/konrad-fighter.png ->
```
stevecotton, line 428: Remember to update the comments too. The line that needs changes is above this, on line 425.

src/scripting/game_lua_kernel.cpp
```diff
@@ -1132,22 +1134,26 @@ int game_lua_kernel::impl_game_config_get(lua_State *L)
 	char const *m = luaL_checkstring(L, 2);
 
 	// Find the corresponding attribute.
-	return_int_attrib("last_turn", tod_man().number_of_turns());
+	return_int_attrib_deprecated("last_turn", "wesnoth.game_config", INDEFINITE, "11.7", "Use wesnoth.scenario.turns instead", tod_man().number_of_turns());
```
stevecotton, line 1137: 11.7

src/scripting/game_lua_kernel.cpp
```diff
@@ -1864,42 +1952,144 @@ int game_lua_kernel::intf_find_cost_map(lua_State *L)
...
+	}
+	int handle_idx = lua_gettop(L);
+	
+	if(*handle != 0) {
+		font::remove_floating_label(*handle);
+	}
+	
 	const SDL_Rect& rect = game_display_->map_outside_area();
-
+	int x = rect.x + rect.w / 2 + loc.wml_x(), y = rect.y + rect.h / 2 + loc.wml_y();
```
stevecotton, line 2067: This would be easier to read if it was 2 lines.

src/scripting/lua_common.hpp
```diff
@@ -207,13 +207,23 @@ int luaW_pcall_internal(lua_State *L, int nArgs, int nRets);
...
+
 #define return_tstring_attrib(name, accessor) \
 do { \
 	if (strcmp(m, (name)) == 0) { \
 		luaW_pushtstring(L, (accessor)); \
 		return 1; \
 	} \
 } while(false)
+#define return_tstring_attrib_deprecated(name, prefix, level, version, msg, accessor) \
+	return_deprecated_attrib(return_tstring_attrib, name, accessor, prefix, level, version, msg)
```
stevecotton, line 226: Many of the new `#defines` here for deprecation helpers seem unused - I think we shouldn't add them unless we're going to use them.

src/filesystem.hpp
CelticMinstrel: You mean the word "image"?

src/scripting/lua_common.hpp
CelticMinstrel: It seemed easier to add all of them and then they could just be used as needed, rather than having to add each new one as needed.

src/filesystem.hpp
CelticMinstrel: You mean the word "image"?

data/lua/core/_initial.lua
```diff
@@ -73,9 +73,4 @@ end
 
 unpack = wesnoth.deprecate_api('unpack', 'table.unpack', 3, '1.17', table.unpack)
 math.pow = wesnoth.deprecate_api('math.pow', '^', 3, '1.17', function(a,b) return a ^ b end)
-if wesnoth.kernel_type() == "Game Lua Kernel" then
-	-- wesnoth.wml_actions.music doesn't exist yet at this point, so create a helper function instead.
-	wesnoth.set_music = wesnoth.deprecate_api('wesnoth.set_music', 'wesnoth.music_list', 1, nil, function(cfg)
-		wesnoth.wml_actions.music(cfg)
-	end)
```
Pentarctagon, line 80: I'm a little confused about this - if `wesnoth.wml_actions.music` doesn't actually exist, what was this doing?

data/lua/core/filesystem.lua
```diff
@@ -0,0 +1,16 @@
...
+--[========[File Handling]========]
+print("Loading filesystem module...")
+
+filesystem.asset_type = {
+    IMAGE = 'images',
+    SOUND = 'sounds',
+    MUSIC = 'music',
+    MAP = 'maps',
+    TRANSLATION = '',
+}
```
Pentarctagon, line 11: Are these really the only asset types?  Also, where is this used?  I just did a page search here and it doesn't show `asset_type` as being used by anything.  Also also, is there a reason for TRANSLATION to be an empty string?

data/lua/core/wml.lua
```diff
@@ -189,14 +189,14 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 	--- Calling wesnoth.fire isn't the same as calling wesnoth.wml_actions[name] due to the passed vconfig userdata
```
Pentarctagon, line 189: This isn't called `wesnoth.fire` anymore.

Pentarctagon:
As mentioned in Discord then, the new modules added are:
* mathx
* versions
* audio
* filesystem
* wesnoth.scenario

And it looks like the deprecations being done are:
* wesnoth.compare_versions --> versions.compare
* wesnoth.play_sound --> wesnoth.audio.play
* wesnoth.sound_volume --> wesnoth.audio.volume
* wesnoth.add_sound_source --> wesnoth.audio.sources
* wesnoth.get_sound_source --> wesnoth.audio.sources
* wesnoth.remove_sound_source --> wesnoth.audio.sources
* wesnoth.music_list --> wesnoth.audio.music_list
* wesnoth.set_music --> wesnoth.audio.music_list
* wesnoth.have_file --> filesystem.have_file
* wesnoth.read_file --> filesystem.read_file
* wesnoth.canonical_path --> filesystem.canonical_path
* wesnoth.get_image_size --> filesystem.image_size
* wesnoth.open_help --> gui.open_help
* wesnoth.color_adjust --> wesnoth.interface.color_adjust
* wesnoth.print --> wesnoth.interface.add_floating_label
* wesnoth.random --> mathx.random
* wesnoth.get_time_stamp --> mathx.current_timestamp - **this doesn't really seem like it's related to math. Also it'd be clearer to call it something like `millis_since_sdl_init`.**
* wesnoth.create_side --> wesnoth.sides.create
* wesnoth.modify_ai --> wesnoth.sides.add|delete|change_ai_component - **I'm not really sure what this means.**
* wesnoth.create_animator --> wesnoth.units.create_animator
* wesnoth.create_weapon --> wesnoth.units.create_weapon
* wesnoth.get_variable --> wml.variables
* wesnoth.set_variable --> wml.variables
* wesnoth.get_all_vars --> wml.all_variables
* wesnoth.fire --> wml.fire
* wesnoth.eval_conditional --> wml.eval_conditional
* helper.rand --> mathx.random_choice
* helper.round --> mathx.round
* helper.shuffle --> mathx.shuffle

data/lua/core/_initial.lua
CelticMinstrel: Emphasis on "at this point". It's not that it doesn't exist; it's just loaded later than this file, so _at the point when this file is executed_, it doesn't exist; but by the time anything calls this function, it _will_ exist.

data/lua/core/filesystem.lua
CelticMinstrel: The asset type can basically be any string, these are just the standard ones. I'm unsure about `TRANSLATION`; I _thought_ I recalled making the po loader use the binary path with an empty string, but I couldn't find where that was in the code. Even if I'm right, perhaps it doesn't make sense for it to be here, since 1) it doesn't really make sense for someone to want to load a translation file and 2) if the add-on happened to use mo files, it wouldn't even work anyway.

So I'll probably delete the `TRANSLATION` one.

data/lua/core/wml.lua
CelticMinstrel: Oh, good point! I'll fix that.

CelticMinstrel:
> wesnoth.get_time_stamp --> mathx.current_timestamp - **this doesn't really seem like it's related to math. Also it'd be clearer to call it something like millis_since_sdl_init.**

Yeah, I too thought it might be a bit of a stretch to put it in `mathx`, but I couldn't think of a better place, either. You do make a good point that the name is misleading; timestamp implies something that can be interpreted as a date, but this is just a monotonic frame accumulator. Your alternate name suggestion sounds both overly long and overly descriptive (there's no reason to include `sdl` in the name), maybe something more like `get_global_frame_count`?

> wesnoth.modify_ai --> wesnoth.sides.add|delete|change_ai_component - **I'm not really sure what this means.**

This one's not new in this PR anyway, it already exists in master… but about what it means, it just adds, modifies, or changes a component in the RCA AI system. A component is just a piece of the AI - an aspect, candidate action, or stage.


Should I take your lack of detailed comments as overall approval of the API?

Pentarctagon:
> Yeah, I too thought it might be a bit of a stretch to put it in mathx, but I couldn't think of a better place, either. You do make a good point that the name is misleading; timestamp implies something that can be interpreted as a date, but this is just a monotonic frame accumulator. Your alternate name suggestion sounds both overly long and overly descriptive (there's no reason to include sdl in the name), maybe something more like get_global_frame_count?

I'm not sure what frame count would refer to here - the description of the function on the wiki says that it's the number of milliseconds since SDL was initialized, not the number of frames that have been displayed since the UI was first displayed.  `get_global_frame_count` is also actually longer than `millis_since_sdl_init`.

> This one's not new in this PR anyway, it already exists in master… but about what it means, it just adds, modifies, or changes a component in the RCA AI system. A component is just a piece of the AI - an aspect, candidate action, or stage.

So `add|delete|change_ai_component` doesn't mean literally that, rather it means `add_ai_component|delete_ai_component|change_ai_component`?

> Should I take your lack of detailed comments as overall approval of the API?

Overall yes.

CelticMinstrel:
> So `add|delete|change_ai_component` doesn't mean literally that, rather it means `add_ai_component|delete_ai_component|change_ai_component`?

Correct.

> I'm not sure what frame count would refer to here - the description of the function on the wiki says that it's the number of milliseconds since SDL was initialized, not the number of frames that have been displayed since the UI was first displayed. `get_global_frame_count` is also actually longer than `millis_since_sdl_init`.

True, I guess frames isn't quite accurate… but the "millis" abbreviation and the inclusion of "sdl" both don't sit well with me. If I go with the basic meaning of your name but address those two issues, then… `ms_since_init`?

Pentarctagon:
I guess my only question would be if there's a difference between Wesnoth starting and the UI initializing.

CelticMinstrel:
You mean like, a time difference between the entry to main() and the SDL_init() call? I have no idea if there's a difference, but even if there is, I doubt it's important. The value you get back from this timestamp function is meaningless except when compared to itself - in other words, the only meaningful use of the timestamp is to compare it with other timestamps within the same session, or subtract it to get a duration.

Pentarctagon:
So ms_since_init would work then I suppose.  And since it's from when SDL is initialized, what about putting it under the `interface` module?

CelticMinstrel:
I'm not really sure how that fits, given what other stuff is in `interface`, but if you really prefer it there, then whatever… it's not like there exists any really good place for it… are you sure you want it moved there?

Pentarctagon:
To me it's more related to the interface than math given it's the milliseconds since SDL was initialized, so yeah I'd prefer it there.

src/filesystem.hpp
stevecotton: More than just one word - the documentation needs to explain what the new argument means.

The example in the documentation could do with words too, and needs to be an example of how to use the current version of the function, not just the previous version of the function.

stevecotton:
IMO (and of course I'm not the person who gets to set the rules on what's too large):

I think this is too large an API change to get into 1.16, given that 1.15.14 is the API freeze release, and this PR includes new APIs. Even if it goes into 1.15.13, the APIs are unlikely to be used by other developers until the freeze - if we find anything that we then want to change, it'll need a decision to either ship as-is, or to thaw the freeze.

At this point, I think you should choose parts of this and make small PRs to get those parts into 1.15.13. Any API that's already changed during the 1.15 branch would be a priority, as that means UMC won't need to be changed again between 1.16 and 1.18.

Pentarctagon:
The full API freeze starts with the first RC - I've updated the roadmap wiki page to state that explicitly.

As far as whether this is too big to go in now, my opinion is that we're either in a feature freeze or we're not, and I'd prefer to keep that clear distinction.  There would also be, at a minimum assuming only 1 RC release is needed, 4 months between the start of the feature freeze and 1.16, so I'm not too worried about the amount of time being too little.

Partially also this is based on the assumption that nobody is going to try and deprecate these new APIs or renamed APIs in 1.17 (which I'd oppose), so I would somewhat like to just get this batch of deprecations done and over with rather than a whole bunch of new deprecations being added into 1.17 right at the start.

stevecotton:
For some reason the CI build didn't run on the last push - it's only run the Labeler Triage check.

The Boost unit tests still don't build, the two lines changed in the last commit need `"images", ` adding to them.

The WML unit tests succeed.

data/ai/lua/ai_helper.lua
```diff
@@ -61,7 +61,7 @@ end
 
 function ai_helper.done_eval_messages(start_time, ca_name)
     ca_name = ca_name or 'unknown'
-    local dt = wesnoth.get_time_stamp() / 1000. - start_time
+    local dt = wesnoth.ms_since_init() / 1000. - start_time
```
stevecotton, line 64: This local variable appears to be unused.

stevecotton:
I tried reenabling World Conquest, but it seems something in this PR is breaking it.

With current master (07a4be462bc5777a36c3fe6ba535421cf4b06954) + my local branches for Pango fixes and IntroWML, I can create a WC game.

With this branch, also based on 07a4be462bc5777a36c3fe6ba535421cf4b06954 + this PR, the multiplayer create-game dialog says the map size for WC is -2x-2. Pressing "Create Game" gives the error "The selected game has no sides!"

data/lua/core/wml.lua
```diff
@@ -359,6 +362,49 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 			wml.array_access.set(key, value)
 		end
 	})
+	
+	--[========[Global persistent variables]========]
```
stevecotton, line 366: This is code moving from World Conquest, but is it generally useful code, and is it presenting the right API for general use? I don't think it is, I think this should be left in WC for the moment, and left for when there's time to properly design an API for 1.17.

* It doesn't seem to have any documentation on the Wiki yet.
* The `immediate=true` makes sense to me, but means it doesn't provide all of the documentated behavior of the WML tag.
* While this might just be that I'm not used to Lua coding conventions, the way that the namespace is accessed via `global_vars.wc2.skip_pickup_dialog` feels a bit odd.

src/scripting/lua_kernel_base.cpp
```diff
@@ -471,22 +483,16 @@ lua_kernel_base::lua_kernel_base()
...
 		{ "require",                  &dispatch<&lua_kernel_base::intf_require>          },
 		{ "kernel_type",              &dispatch<&lua_kernel_base::intf_kernel_type>          },
 		{ "compile_formula",          &lua_formula_bridge::intf_compile_formula},
 		{ "eval_formula",             &lua_formula_bridge::intf_eval_formula},
 		{ "name_generator",           &intf_name_generator           },
-		{ "random",                   &intf_random                   },
 		{ "log",                      &intf_log                      },
-		{ "get_image_size",           &intf_get_image_size           },
-		{ "get_time_stamp",           &intf_get_time_stamp           },
+		{ "ms_since_init",            &intf_get_time_stamp           },
```
stevecotton, line 495: Why isn't the C++ function getting renamed?

CelticMinstrel: There's no particular reason, I just didn't bother. I can rename it if you prefer.

data/ai/lua/ai_helper.lua
CelticMinstrel: I don't think that has anything to do with this PR though…

data/lua/core/wml.lua
CelticMinstrel: I moved it because I thought it _did_ look like a good, generally useful API. It's true that it's not exposing the transactional nature of PersistenceWML though…

src/scripting/lua_kernel_base.cpp
Pentarctagon: Please keep names consistent across lua and C++ usage.

data/ai/lua/ai_helper.lua
mattsc: Yeah, that's a leftover from a previous version of that function.  Feel free to remove it, or I can do so separately.  In the latter case, I'll wait until after this has been merged though, so that it doesn't cause conflicts.

src/scripting/lua_kernel_base.cpp
CelticMinstrel: Done

data/lua/core/wml.lua
stevecotton: I feel we'll want to change the interface again during 1.17, and as the existing code is campaign-specific it doesn't need to be in the "get all the deprecations done now so that there aren't more deprecations in 1.17" group.

Why rush to make this a general API in 1.16 instead of waiting until we have time to design it slowly?

data/lua/core/audio.lua
```diff
@@ -0,0 +1,30 @@
...
+	wesnoth.add_sound_source = wesnoth.deprecate_api('wesnoth.add_sound_source', 'wesnoth.audio.sources', 1, nil, function(cfg)
+		wesnoth.audio.sources[cfg.id] = cfg
+	end, 'Assign a config to the sound source ID')
+	wesnoth.get_sound_source = wesnoth.deprecate_api('wesnoth.get_sound_source', 'wesnoth.audio.sources', 1, nil, function(id)
+		return wesnoth.audio.sources[id]
+	end, 'Index by the sound source ID')
+	wesnoth.remove_sound_source = wesnoth.deprecate_api('wesnoth.remove_sound_source', 'wesnoth.audio.sources', 1, nil, function(id)
+		wesnoth.audio.sources[id] = nil
+	end, 'Assign nil to the sound source ID')
+	wesnoth.music_list = wesnoth.deprecate_api('wesnoth.music_list', 'wesnoth.audio.music_list', 1, nil, setmetatable({}, {
```
stevecotton, line 21: Given #4459, #4030 and #4458, is wesnoth.audio.music_list ready? It seems there's going to be further changes to that API.

src/scripting/game_lua_kernel.cpp
```diff
@@ -4238,6 +4226,12 @@ game_lua_kernel::game_lua_kernel(game_state & gs, play_controller & pc, reports
 	luaL_setfuncs(L, map_callbacks, 0);
 	lua_pop(L, 1);
 
+	// Add open_help to the GUI module
```
stevecotton, line 4229: Why do this here, instead of moving it to lua_gui2.cpp?

Given the naming of the other gui functions, `show_help` seems more consistent than `open_help`.

data/lua/core/gui.lua
```diff
@@ -154,3 +154,6 @@ wesnoth.get_dialog_value = wesnoth.deprecate_api('wesnoth.get_dialog_value', val
 wesnoth.add_dialog_tree_node = wesnoth.deprecate_api('wesnoth.add_dialog_tree_node', '<widget>:add_item_of_type', 1, nil, add_dialog_tree_node)
 wesnoth.remove_dialog_item = wesnoth.deprecate_api('wesnoth.remove_dialog_item', '<widget>:remove_items_at', 1, nil, remove_dialog_item)
 wesnoth.show_dialog = wesnoth.deprecate_api('wesnoth.show_dialog', 'gui.show_dialog', 1, nil, gui.show_dialog)
+if wesnoth.kernel_type() == "Game Lua Kernel" then
```
stevecotton, line 157: Why restrict specifically this to Game Lua Kernel, what's different compared to `show_dialog`?

src/scripting/lua_kernel_base.cpp
```diff
@@ -95,6 +94,63 @@ static int intf_compare_versions(lua_State* L)
...
+	}
+	version_info vers(major, minor, rev, sep[0], special);
+	lua_push(L, vers.str());
+	return 1;
+}
+
+/**
+ * Returns the current Wesnoth version
+ */
+static int intf_current_version(lua_State* L) {
```
stevecotton, line 149: These version functions should have unit tests.

data/lua/core/gui.lua
CelticMinstrel: That's a very good question. It has always been restricted to the Game Lua Kernel, so I simply didn't change that, but I'm not sure that there's a good reason for that…

src/scripting/game_lua_kernel.cpp
CelticMinstrel: Both good points.

data/lua/core/audio.lua
CelticMinstrel: I'm not sure if there will be changes, but it's ready for use in its current form.

src/scripting/lua_kernel_base.cpp
CelticMinstrel: Obviously I can't unit test current_version though.

I guess I could copy the version tests from the Boost unit tests and convert them to Lua…

data/lua/core/audio.lua
stevecotton: IIUC from the IRC discussion, a reason for accepting a lot of deprecations now is to reduce the number of deprecations between 1.16 and 1.18. This sounds as if bits of the newly-added API will get deprecated before 1.18 is released.

CelticMinstrel: Well, as far as I know there are no current plans to make any of the API changes noted in those issues, so I don't think it makes sense to hold off on this change on the grounds that someone _might_ hypothetically make more changes in the future (a future which might not even _be_ in the 1.17 cycle - it might end up waiting for 1.19 for example).

src/scripting/lua_kernel_base.cpp
stevecotton: A unit test would at least check that it doesn't crash and returns a plausible version-like string.

data/lua/core/audio.lua
stevecotton: @Pentarctagon please would you give a decision about this?

src/help/help.cpp
```diff
@@ -87,16 +87,14 @@ void show_unit_description(const unit_type &t)
 		help::show_unit_help(t.id(), t.show_variations_in_help(), hide_help);
 }
 
-help_manager::help_manager(const game_config_view *cfg) //, gamemap *_map)
+static game_config_view dummy_view;
+help_manager::help_manager(const game_config_view *cfg)
+	: guard(game_cfg, cfg == nullptr ? &dummy_view : cfg)
```
stevecotton, line 92: What can this `guard` do, other than re-introducing a dangling pointer at a later point in time?

The old code was easier to read, and I think it would behave the same unless a dangling pointer had already been created before this method was called. If that dangling pointer does exist, then this just makes debugging harder.

CelticMinstrel: This guard makes help managers nestable. If you had two help managers active simultaneously, with the old code the pointer would be nulled out when the inner one exits, meaning an attempt to show help after that could cause a crash. With this, it'll be nulled out when the outer one exits, and if by some chance there were two different game configs (I don't think there ever would be though), the outer one would be restored when the inner help manager exits.

data/lua/core/audio.lua
Pentarctagon: What I want to avoid in 1.17 and onward is deprecating something simply to rename it, move it, re-order its arguments, or any other sort of superficial change, since 1.15 is already a lost cause in that regard anyway.  If an API needs to have a breaking change made to it because the current implementation is actually functionally broken somehow, then I think that's fine inasmuch as it's better to have a working API.

stevecotton: So renaming it before 1.16 and then replacing it with a different set of functions after 1.16 is accepted? Okay, I've misunderstood then.

src/help/help.cpp
stevecotton: Replied to in the other PR, I still consider this an issue in this PR.

Pentarctagon: In what situation would there be two help managers active?

data/lua/core/audio.lua
Pentarctagon: I'm not sure how that would be different than renaming?

Pentarctagon: The proposal in the bug reports more or less comes down to "we need to replace our use of SDL_Mixer entirely", which would be a lot of work and which nobody has said they have any intention of doing.

The argument "we shouldn't to <thing> because someone else might at some point maybe do something else" isn't a very convincing one to me. While I do think there's too much API churn, accepting that argument would make it very hard to do a lot of things, or anything at all if taken to its extreme.

Pentarctagon:
@CelticMinstrel Can you add an `experimental` module as well?

src/help/help.cpp
CelticMinstrel: If you call `gui.show_help()`, because in order to ensure the help exists no matter where it's called from, it needs to create a help manager.

CelticMinstrel:
Adding a new module is pretty easy, but what are you expecting me to put in there?

data/lua/core/wml.lua
CelticMinstrel: It's already a general API in 1.14, so your argument doesn't really make sense…

Pentarctagon:
I don't have anything specific in mind, but essentially it'd just be a place for anything brand new that you're not confident you wouldn't want to change in a future release.  Also, during 1.17, I'd like to take anything that isn't documented and either document it or move it to the experimental module.  Then for everything that's *not* experimental, enforce a higher bar for any changes that would require UMC authors to update their code.

CelticMinstrel:
So, for now, just adding it as an empty module would suffice?

Pentarctagon:
I'd say `wesnoth.experimental` and then maybe `wesnoth.internal` for things that are internal implementation details that exist but also shouldn't be used or relied on.  Why would `wml.get_variable` be one of those though?

data/lua/core/wml.lua
stevecotton: What was already a general API in 1.14? The Lua code is being copied from World Conquest, so it wasn't general Lua API before.

CelticMinstrel:
Because the official API for accessing variables is now `wml.variables`; `wml.get_variable` has thus become an internal implementation detail of `wml.variables`.

I'm not sure I like the idea of creating `wesnoth.internal` as a dumping ground for things like that… especially since there aren't very many of them… but it's certainly one possibility, and does make it quite obvious what you shouldn't be using…

data/lua/core/wml.lua
CelticMinstrel: I don't remember my thoughts when I originally posted that, but I may have been referring to the WML API?

What about putting this in the "experimental" module that Pentarctagon suggested? Then it can be generally available in 1.16 but also more easily iterated on if we decide on a better design.

stevecotton: As it doesn't need any C++ changes, I don't see what it gives to UMC authors. My alternative would be for UMC authors in 1.14/1.16 to just cut-and-paste the block from WC into their own add-on's utils files, which will still work unchanged in 1.18 even if we add a nicer API for 1.18.

data/lua/wml-tags.lua
```diff
@@ -967,7 +967,9 @@ function wml_actions.terrain_mask(cfg)
 		rules[#rules + 1] = rule
 	end
 	if cfg.mask_file then
-		mask = wesnoth.read_file(cfg.mask_file)
+		local resolved = filesystem.resolve_asset(filesystem.asset_type.MAP, cfg.mask_file)
```
stevecotton, line 970: Mainline seems inconsistent about whether map-masks should be in a "maps" or a "masks" folder. SoF uses "maps", while Liberty, LoW and HttT use "masks".

I think "masks" makes sense, but please let's move this to a separate PR where we can discuss conventions and whether the editor needs changes at leisure, without needing to affect the timing of the reorg PR.

src/scripting/game_lua_kernel.cpp
```diff
@@ -4197,10 +4169,18 @@ game_lua_kernel::game_lua_kernel(game_state & gs, play_controller & pc, reports
...
-	lua_pushcfunction(L, &lua_common::intf_tovconfig);
-	lua_setfield(L, -2, "tovconfig");
+	static luaL_Reg const wml_callbacks[] {
+		{"tovconfig", &lua_common::intf_tovconfig},
+		{"eval_conditional", &intf_eval_conditional},
+		// These aren't actually part of the API - they're used internally by the variable metatable.
+		{ "get_variable", &dispatch<&game_lua_kernel::intf_get_variable>},
+		{ "set_variable", &dispatch<&game_lua_kernel::intf_set_variable>},
+		{ "get_all_vars", &dispatch<&game_lua_kernel::intf_get_all_vars>},
+		{ nullptr, nullptr }
```
stevecotton, line 4181: Having all of the commits in mainline be self-contained is helpful when in future a developer is using `git bisect`.

There's currently a 10-commit window where this table lacks its sentinel - when rebasing, please squash fix commits like this into the commit that needs them.

data/lua/core/wml.lua
Pentarctagon: It's definitely much more visible.  As it is, the only way any UMC author would know to do that would be if they found the functionality themselves somehow or were told by someone else who had found it.  It also provides an opportunity for people to provide feedback on something that will probably become a non-experimental API at some point.

stevecotton:
`[sound_source]` is currently broken. On a build with this merged into current master (so b740153d9261ec02702ea219dcddf6fc5183dfe3 + 59ac7a20bf1197902ef69209a2df574d19a24006), the following error message is printed at the start of Eastern Invasion S01:

    lua/wml-tags.lua:845: bad argument #3 to 'newindex' (sound source expected, got vconfig)

data/lua/wml-tags.lua
CelticMinstrel: I have the opposite opinion; a mask is a type of map, so I think it makes more sense to put them in the same folder. There's always the option of putting them in a subfolder if you really want them separate.

data/lua/core/wml.lua
CelticMinstrel: Alright, I moved it to experimental.

stevecotton:
There will be both `wesnoth.interface.float_label` and `wesnoth.interface.add_floating_label`, which despite the similar names are different implementations.

The arguments for `add_floating_label` aren't documented in the C++, which is a break from the coding style in this file. The code also seems to be trying to accept options in multiple orders (and treating hexadecimal strings as a different thing to numbers), which is adding unnecessary complexity.

src/help/help.cpp
stevecotton: @CelticMinstrel, will you now remove this?

CelticMinstrel:
> There will be both `wesnoth.interface.float_label` and `wesnoth.interface.add_floating_label`, which despite the similar names are different implementations.

Ahh, good point! I missed that… hmm…

src/help/help.cpp
stevecotton: Just realised that nesting the help like that would be a UX regression.

The help browser already has a back button; from a UX perspective, opening a second help page already does nest. Once it's nested there are two buttons available - one to unnest by one level, and another to immediately exit the help browser regardless of how many levels are nested.

The next question is what would trigger a second help manager to open, and again that would be a UX regression if it was anything other than the UI that the user was interacting with. From a UX perspective, I think the only way `gui.show_help()` should be able to open a second help page would be if we added a feature that allowed Lua or ActionWML for interactive help pages; and even in that case, the correct UX would be to open a new page in the existing help browser, so that the back button works.

CelticMinstrel: Your comment makes no sense whatsoever. The help manager has nothing to do with UX; all it does is make sure help has been parsed and generated topics have been generated.

Additionally, I tested it and there doesn't seem to be any weird behaviour with the back button. It works as expected.

Some parts of your comment would be relevant if a help topic were able to contain links or buttons that execute Lua code which could theoretically call `gui.show_help` again, but since that's not possible right now, I think it's something we can worry about if it becomes possible in the future.

stevecotton: You said several times "your comment makes no sense" in the RAII thread. Please learn from that thread.

@Pentarctagon does it make sense to you?

Pentarctagon: It sounds like this change essentially results in a no-op, since there's currently no way to have two help managers existing at the same time right now.  In which case I would say that this specific change should be removed until/unless it becomes possible in the future.

CelticMinstrel: It's not a no-op, because this commit _creates_ a situation where there can be two help managers existing at the same time. Now, I _could_ go back to exposing help only in the game kernel, making this become unnecessary; but I think it should be done anyway, as the previous help manager seems like a crash waiting to happen.

From my perspective, there's no reason _not_ to do this change. It has no impact on _anything_, except making the rest of the commit safe and crash-free. None of the arguments you made against using `guard_value` in the other thread apply here, either - there, I replaced one resource management idiom with another, and you were reasonably concerned that there might be unforeseen consequences from that change; but here, the code was _already_ using the `guard_value` idiom, just in a non-reentrant way.

There is technically a third option, which is to remove the help manager altogether and make all the help code directly access the game config whenever help is shown. I have no idea what unforeseen consequences that might have, though.

> You said several times "your comment makes no sense" in the RAII thread.

I said it doesn't make sense because it doesn't make sense. Your comment was about things that have no relationship whatsoever with this change. The help manager has nothing to do with any user interface code; it's just a scope guard, nothing more.

> Please learn from that thread.

That thread was a completely different situation. We apparently had a different definition of what RAII is, so we ended up talking past each other for awhile until that discrepancy was brought to light. That's not what happened here. Anything I might learn from that thread, while likely useful in some capacity, doesn't apply here.

-----

Just to be clear, the following two requests are wholly incompatible. I can do one of them, but not both.

1. Expose `gui.show_help` in lua_kernel_base.
2. Revert this change to the help manager.

stevecotton: Okay, I misunderstood how long the lifecycle of the main help_manager is. I'm now re-reviewing this.

PR #5538, opened 2021-02-13
Command history for in-game consoles

### LINKED ISSUE #1542: implement optional command-line history feature without GNU readline/history dependency (GNA #24569)
**[Original submission](https://gna.org/bugs/index.php?24569)** by gfgtdf on 2016-04-07

readline/history has limited support on windows:

1) It doesnt support utf8 filepaths on windows

2) The examples from their webpage make the programm crash if compiled with msvc.

3) It imposstible to extract the offical download files on windows becasue of invalid filename.

(1) is the main problem here becasue it means that we won't be able to write history to disk on some windows sytems. ( (2) and (3) are just to show that readline is clearly not meant as a platform independent libary)

In a plattform indenpendent software like wesnoth, libraries like readline should never be used.

Ideally we woudl just improve gui2's own histpry implementation and use that then too in the lua console.

(Reproduced on windows)
Release: 1.13+dev
Priority: 5 - Normal
Severity: 3 - Normal

kabachuha:
Now it's possible to browse the command history with up/down keys in in-game consoles such as used by debug mode, ai and search floating textboxes. It stores all the entered commands in chronological order, but already existing items will be moved to the end instead. If the player jumps to the previous command with some unenterred text in the textbox, it will also be stored. Chat messages are not stored like this to avoid spam.

I added a video with a little showcase below.

Closes #1542 .

https://user-images.githubusercontent.com/14872007/108371136-7e5c0300-720e-11eb-9039-bea0882c196d.mp4

(or this link if it doesn't show on github)
https://drive.google.com/file/d/18oYwpW3vEs2VP5OOh2iAXgyiS2Pjb-Er/view?usp=sharing

### DIFF
--- a/src/floating_textbox.cpp
+++ b/src/floating_textbox.cpp
@@ -33,7 +33,8 @@ namespace gui{
 		check_(nullptr),
 		mode_(TEXTBOX_NONE),
 		label_string_(),
-		label_(0)
+		label_(0),
+		command_history_()
 	{}
 
 	void floating_textbox::close(game_display& gui)
@@ -142,5 +143,19 @@ namespace gui{
 					events::chat_handler::MESSAGE_PRIVATE, false);
 		}
 		box_->set_text(text);
+	}
+
+	void floating_textbox::memorize_command(const std::string& command)
+	{
+		if(command.empty()) {
+			return;
+		}
+
+		auto prev = std::find(command_history_.begin(), command_history_.end(), command);
+
+		if(prev != command_history_.end()) {
+			command_history_.erase(prev);
+		}
+		command_history_.emplace_back(command);
 	}
 }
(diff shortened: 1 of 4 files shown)


cooljeanius:
for searchability purposes, this is functionality similar to that of readline or libedit.

Pentarctagon:
I'd rather it be called command history than command memory.  The former is the far more common terminology for this sort of feature.

kabachuha:
@Pentarctagon , renamed it.

CelticMinstrel:
Does this work only for the command console or also for chat, formula, and search? If it does work, do all four have separate history? If it doesn't, is it easy to extend to them?

kabachuha:
@CelticMinstrel yes, they all have a shared command history except for the chat (disabled to avoid spam) since only one instance of floating_textbox, which it is attached to, is constructed within playsingle_controller.

CelticMinstrel:
Separate history would definitely be preferable, but I guess this is better than nothing. Thanks for implementing it.

PR #5547, opened 2021-02-15
UtBS - new terrain-graphics rules for flood water

### LINKED ISSUE #5539: UtBS: Terrain transition issues with wooden floors in Out of the Frying Pan
In UtBS scenario 8 (Out of the Frying Pan) there is a dungeon room that uses the wooden planks floor terrain, which normally has an orderly transition to water with custom decorated edges. Someone decided that for this particular application of it in UtBS that wasn't enough, however, and added code to replace those transitions with what's apparently supposed to be water blending into the floor, kind of like sandy beaches.

The problem is that the code in question doesn't work in practice on either 1.14 or master, creating a broken terrain look instead:

image

Furthermore, on master some of the tiles in the room use a different terrain (Old Wooden Floor `Ior` instead of Basic Wooden Floor `Iwr`) that's not accounted for by the code, resulting in a mixture of both broken and default transitions instead of the intended result.

The code can be found in `data/campaigns/Under_the_Burning_Suns/utils/terrain_graphics.cfg` guarded behind `#ifndef EDITOR`.

doofus-01:
Fixes #5539
Makes the flood water in Sc8 have a fading overlap on floor terrains, instead of the usual banks and edges.
I think it works based on what I see in the editor, but still need to test in scenario.

I'm not sure if this should be back-ported to BfW 1.14, or maybe it's better to just remove the custom rules and live with the default transitions.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
@@ -4181,7 +4181,7 @@
         [/store_locations]
 
         [terrain]
-            terrain=Wwg
+            terrain=Wwgz
 
             [filter_adjacent_location]
                 find_in=shallow_to_deep_locs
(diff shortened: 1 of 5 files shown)


doofus-01:
It seems to work in scenario, I'll merge this in a bit.
Screenshot_20210215

PR #5548, opened 2021-02-15
Take display zoom into account when rendering unit animation frames

### LINKED ISSUE #5508: animation frame keys x,y do not scale with zoom level
 The Battle for Wesnoth version 1.15.9+dev (63c71a010af-Modified) x86_64

#### Libraries
Boost:             1.66
OpenSSL/libcrypto: 1.1.1d (runtime 1.1.1d)
Pango:             1.44.7 (runtime 1.44.7)
SDL:               2.0.8 (runtime 2.0.8)
SDL_mixer:         2.0.2 (runtime 2.0.2)
SDL_ttf:           2.0.14 (runtime 2.0.14)

gcc & scons
 - What OS are you running?
Linux

### Describe the bug
The x,y keys are ignored in animation frames.  It's easiest to see in the dunefolk raider, the torch flame is centered on the sprite image.  I believe many of the missile frames are also affected, for example the footpad's rock.

halo_x,halo_y do seem to still work.

I tried this on 1.15.8+dev, forgot to record which commit but from around New Year's, and it seemed to have the same issue.  I'll add the blocker label, but this seems like something that would have been noticed before so it may just be a quirk on my system.

EDIT:  It looks like x,y do still work outside the [frame] tags.
### Screenshots
Screenshot_000
The sidebar image shows where the flame is supposed to be.

shikadiqueen:
**Note:** I have tested the patch against the bug's test case, but I need to go over my own files to see if I can find a more elaborate test case involving progressive parameters.

It was already considered for registering halos, just not for rendering regular frames. This commit touches the halo code a little so we don't need to call get_zoom_factor() multiple times per frame.

Closes #5508.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -24,6 +24,7 @@
    * Added information about the build's (not runtime) target CPU architecture to the game version info dialog and --report.
  ### WML Engine
  ### Miscellaneous and Bug Fixes
+   * Fixed display zoom not being taken into account when using the `x`, `y`, `directional_x` and `directional_y` attributes in unit animations.
 
 ## Version 1.15.9
  ### Add-ons server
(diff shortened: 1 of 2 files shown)


shikadiqueen:
Okay, I didn't find any useful test cases in my own add-ons, but it turns out the Dunefolk Falconer and Fire Dragon use `x` and `directional_x`, respectively. Both appear to work incorrectly before the patch and correctly with the patch, when zoomed in.

Wedge009:
Any idea if something similar needs to apply to resolve #5974? I'm trying to see how zoom affects unit positioning.

PR #5551, opened 2021-02-16
units - Walking Corpse beast-rider (taurus) variation

doofus-01:
This initial commit has just the base frame, I'll be working on the other frames and the portrait.  I'm not sure about the stats, but those would be easy to change later, assuming this is merged within the next couple months.
Screenshot_20210215_3


doofus-01:
I don't think the CI failures are from anything I did, it looks like all commit to master at the time I started this PR were failing.

PR #5556, opened 2021-02-17
 upgrades codeblock to Lua 5.4.2 

newfrenchy83:


soliton-:
Is there a reason to switch to c++20 when wesnoth only requires c++17? Seems like a good way to accidentally try to introduce c++20 only constructs...

newfrenchy83:
> 
> 
> Is there a reason to switch to c++20 when wesnoth only requires c++17? Seems like a good way to accidentally try to introduce c++20 only constructs...

I will revert that then.

PR #5561, opened 2021-02-19
addon/client: Add file operation progress dialog for installing add-ons

irydacea:
This implements an add-on extraction progress dialog that does not actually run its own event loop, allowing the caller to take ownership of it (`display()` static method) and update its state using a callback function object.

The dialog is roughly identical to the network progress dialog, except for the greyed out Cancel button. This is intentional so that the dialogs can be transitioned more or less seamlessly into each other by the add-ons client code.

image

In order to avoid stalling the extraction process in UI code, the callback is invoked for every single file, but the dialog's progress update method places a time restriction on GUI2 API calls of 120 milliseconds -- this is a good throttle interval that allows add-ons to be extracted in about the same amount of time as before while still updating the progress bar smoothly enough for add-ons that take longer than that.

(This is not the most trivial code to test, so it is suggested to add a sleep/delay API call in `unarchive_file()` in src/addon/manager.cpp to introduce artificial delays.)

One issue with this code, however, is that because `modeless_dialog` doesn't execute its own event loop, the only way to get the dialog to be updated is to force a draw event in ourselves via the new `gui2::dialogs::modeless_dialog::force_redraw()` method. This is really a side-effect of my design choice here to run the dialog in the middle of a blocking operation instead of somewhere where events are being processed normally. I'm not entirely sure if the draw events would be pushed even in that case, however.


AI0867:
What is the `SDL_Delay` needed for? Won't the drawing actually happen otherwise?

shikadiqueen:
To avoid busy-looping if the method is called from a loop.

Pentarctagon:
@mesilliac is the implementation of this still applicable given all the rendering changes?

mesilliac:
The changes to `modeless_dialog` would no longer work - but they're also no longer necessary, it can just call `window_->queue_redraw()` to update the display.

Apart from that it looks like it should work.

mesilliac:
Seems to be working correctly after reverting the modeless_dialog commit and adding a tweak to use the new drawing system.

I tested with a delay in `unarchive_file` as suggested and it works as expected.

Pentarctagon:
@soliton- or @CelticMinstrel can you push an xcode projectfile update?

Otherwise yeah, I don't see why not.

PR #5541, opened 2021-02-14
Extent special_id_active and special_type_active to abilities

newfrenchy83:
with the number of commit and conversation in https://github.com/wesnoth/wesnoth/pull/5239 it is more simle to add another PR for fix https://github.com/wesnoth/wesnoth/issues/4840 issue

PR #5562, opened 2021-02-19
Replace [leading_anim] callable by abilities used like weapons by add of [teaching_anim]

newfrenchy83:
Now what abilities used like weapon could call[leading_anim] it is more carefull to limit the call
in core unit to [leadership] only for don't have unwanted call of animation if abilites special of leader type are added by [modifications]
in these units.


stevecotton:
Would a similar change be needed to many units in add-on eras?

I haven't understood why this is needed, but if it is needed in the add-ons too then I think it's a bug in the already-merged change to `[leading_anim]`'s behavior.

CelticMinstrel:
I think the issue is that `[leading_anim]` is now called in a lot more places than it used to be, but whether this is the best solution, I'm not sure. Since `[leading_anim]` is an aliased tag, we could always just make it imply that filter; someone wanting a leadership animation for some other ability would then use `[animation]apply_to=leading` and provide their own filter. Then again, no other alias implies a filter, so I'm not sure if this is a good idea either.

stevecotton:
Given there's 3000 units in Ageless Era that have a `[leading_anim]`, how many of them will need to change?

Why is `[leading_anim]` being used at all for a non-leadership ability? According to the Wiki, `[leading_anim]` is only used for leadership, which is out of date.

I didn't keep track of the previous PR, but from #5541 I guess the `[leading_anim]` tag is being used for a gives-initiative ability. In which case, why not create an `[initiative_anim]` tag instead?

newfrenchy83:
> 
> 
> Given there's 3000 units in Ageless Era that have a `[leading_anim]`, how many of them will need to change?
> 
> Why is `[leading_anim]` being used at all for a non-leadership ability? According to the Wiki, `[leading_anim]` is only used for leadership, which is out of date.
> 
> I didn't keep track of the previous PR, but from #5541 I guess the `[leading_anim]` tag is being used for a gives-initiative ability. In which case, why not create an `[initiative_anim]` tag instead?

i removed call of [leading_anim] by abilities used like weapons.

CelticMinstrel:
I don't quite understand the diff but... those abilities will still trigger an animation, right? Even if it's not `[leading_anim]`.

newfrenchy83:
> 
> 
> I don't quite understand the diff but... those abilities will still trigger an animation, right? Even if it's not `[leading_anim]`.

No, it is more carefull what it have [leading_anim] called by [leadership] only, and not abilities used like weapon trigger animation, finally.

CelticMinstrel:
That's not the question. The question is: do those leadership-like abilities trigger an animation on the leader?

newfrenchy83:
> 
> 
> That's not the question. The question is: do those leadership-like abilities trigger an animation on the leader?

The reply is yes. And this is what i tryed to resolve, but remove that is the best . 

If two animation leader are called in same time(two [leading_anim] or two tags different), it will alternate of anim in fight, and if one animation called what if no [leadership] active by example, if another leader what teacher is present, then this animation will not be played. In these condition, what [leadership] was the only to called animation of leading ([leading_anim] remain the best option.

CelticMinstrel:
So if they don't call a `[leading_anim]`, what type of animation do they call?

newfrenchy83:
> 
> 
> So if they don't call a `[leading_anim]`, what type of animation do they call?

If this PRis merged, none. nor [leading_anim] nor [initiative_anim] what i can create. But now, [leading_anim] is the only animation callable by these abilities special.

AI0867:
`[leading_anim]` shouldn't be special, it should simply be an `[animation]` with a default filter for leadership abilities. Is its use in WC2 simply a misunderstanding by the WML author, or is there a bug in the animation engine?

newfrenchy83:
> 
> 
> `[leading_anim]` shouldn't be special, it should simply be an `[animation]` with a default filter for leadership abilities. Is its use in WC2 simply a misunderstanding by the WML author, or is there a bug in the animation engine?

[leading_anim] don't called by weapon specials but by [leadership] or by abilities used like specials who affect adjacent unit( specials in [abilities]) and i want repair this error and what [leading_anim] called by [leadership] only.

AI0867:
> > `[leading_anim]` shouldn't be special, it should simply be an `[animation]` with a default filter for leadership abilities. Is its use in WC2 simply a misunderstanding by the WML author, or is there a bug in the animation engine?
> 
> [leading_anim] don't called by weapon specials but by [leadership] or by abilities used like specials who affect adjacent unit( specials in [abilities]) and i want repair this error and what [leading_anim] called by [leadership] only.

I believe you're saying that `[leading_anim]` should only be used by `[leadership]`? I agree. I was just wondering why the author of all these animations chose to use that tag instead of a more appropriate one. It might be a misunderstanding or an engine problem.

newfrenchy83:
> 
> 
> > > `[leading_anim]` shouldn't be special, it should simply be an `[animation]` with a default filter for leadership abilities. Is its use in WC2 simply a misunderstanding by the WML author, or is there a bug in the animation engine?
> > 
> > 
> > [leading_anim] don't called by weapon specials but by [leadership] or by abilities used like specials who affect adjacent unit( specials in [abilities]) and i want repair this error and what [leading_anim] called by [leadership] only.
> 
> I believe you're saying that `[leading_anim]` should only be used by `[leadership]`? I agree. I was just wondering why the author of all these animations chose to use that tag instead of a more appropriate one. It might be a misunderstanding or an engine problem.

@CelticMinstrel and @AI0867 The author, is me, and yes i done an error, what i corrected in add [special_anim] called by special abilities used like leadership, and only if [leadership] or [resistance] of teacher are inactive

CelticMinstrel:
Now we just need to agree on the name. I don't really like `[special_anim]` although I can kinda see why you chose it. What about `[teaching_anim]`? Or is that too confusing? We _do_ have a student/teacher distinction in the filters so it might make sense but not sure...

newfrenchy83:
ok for (teaching_anil]

newfrenchy83:
@CelticMinstrel  It's done.

newfrenchy83:
> 
> 
> Now we just need to agree on the name. I don't really like `[special_anim]` although I can kinda see why you chose it. What about `[teaching_anim]`? Or is that too confusing? We _do_ have a student/teacher distinction in the filters so it might make sense but not sure...

i replaced [special_anim] by [teaching_anim], this PR canbe pulled now.

newfrenchy83:
@Pentarctagon coud you review this PR, please?

Pentarctagon:
So if I'm understanding correctly, with this change there is a new tag [teaching_anim] added that is used when a unit with a leadership-like ability "teaches" adjacent units a weapon special?

newfrenchy83:
> 
> 
> So if I'm understanding correctly, with this change there is a new tag [teaching_anim] added that is used when a unit with a leadership-like ability "teaches" adjacent units a weapon special?

yes, but only if teacher don't use a [leadership] ability (if apply_to=self) or a [resistance](if apply_to=opponent) ability because they can call [leading_anim] or [helper_anim].

src/units/animation.cpp
```diff
@@ -1346,6 +1346,34 @@ void unit_animator::add_animation(unit_const_ptr animated_unit
...
+	if(!animated_unit){
+		return false;
+	}
+
+	display* disp = display::get_singleton();
+
+	anim_elem tmp;
+	tmp.my_unit = std::move(animated_unit);
+	tmp.src = src;
+	tmp.animation = tmp.my_unit->anim_comp().choose_animation(*disp, src, event, dst, value, hit_type, attack, second_attack, value2);
```
Pentarctagon, line 1368: Is there a reason to not just do `animated_unit->anim_comp().choose_animation(*disp, src, event, dst, value, hit_type, attack, second_attack, value2);`?

newfrenchy83: No, i fixed this.

src/units/animation.cpp
```diff
@@ -1345,6 +1346,29 @@ void unit_animator::add_animation(unit_const_ptr animated_unit
...
+	}
+
+	display* disp = display::get_singleton();
+
+	if(animated_unit->anim_comp().choose_animation(*disp, src, event, dst, value, hit_type, attack, second_attack, value2)){
+		return true;
+	}
+
+	return false;
+}
```
Pentarctagon, line 1370: I think this can be even more simply written as:
```
return animated_unit
	&& animated_unit->anim_comp().choose_animation(*display::get_singleton(), src, event, dst, value, hit_type, attack, second_attack, value2);
```

src/units/animation.cpp
```diff
@@ -1345,6 +1346,29 @@ void unit_animator::add_animation(unit_const_ptr animated_unit
...
 
+bool unit_animator::has_animation(unit_const_ptr animated_unit
+		, const std::string& event
+		, const map_location &src
+		, const map_location &dst
+		, const int value
+		, const unit_animation::hit_type hit_type
+		, const_attack_ptr attack
+		, const_attack_ptr second_attack
+		, int value2) const
```
Pentarctagon, line 1357: Please add a doxygen comment for this method.

newfrenchy83: doxygen added

src/units/animation.cpp
newfrenchy83: it is done

Pentarctagon:
I think this is fine.  I'll leave it open a couple days in case there are other comments.

Pentarctagon:
@newfrenchy83 can you fix the merge conflict?

newfrenchy83:
> 
> 
> @newfrenchy83 can you fix the merge conflict?

it is done

PR #5563, opened 2021-02-19
Improve the Load dialog's UX when there are no save files or corrupted files

### LINKED ISSUE #5517: 'Load Game' window should not be shown if there is no saved game
The Battle for Wesnoth version 1.15.9+dev (c42bf114-Clean) x86_64

### Describe the bug
The current behaviour is a bit weird, you can interact with elements inside the window even if there is no save to load.

### To Reproduce

1. Remove all saved games
2. Click 'Load'

### Expected behavior
'Load Game' window should not be shown when there is no saved game.
There should be a different dialogue window clearly stating that there is no saved game to load.
Screenshot_20210204_194732

stevecotton:
Re-add the popup that appears when "Load" is pressed on the title screen if
there are no files, but now check for files from previous versions too.

Improve handling of switching from a version with files to a version without
files, disabling buttons on the load dialog when there are no files in the
current directory.

This adds a \todo about the error path at the start of evaluate_summary_string,
this path is reached by creating empty files in the save dir. The dialog's state
is reasonable, so it seemed a low priority and I don't want to introduce new
bugs by refactoring this path now.

Fixes #5517

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,8 @@
  ### Translations
  ### Units
  ### User interface
+   * Improved the Load dialog when there are no saved games for the current version, and also when there are corrupted files.
+   * Re-added the pop-up when there are no saved games at all (issue #5517).
  ### WML Engine
    * Standard Location Filters now support gives_income=yes|no to make it simpler to match villages regardless of owner
  ### Miscellaneous and Bug Fixes
--- a/src/gui/dialogs/game_load.cpp
+++ b/src/gui/dialogs/game_load.cpp
@@ -58,6 +58,27 @@ namespace gui2::dialogs
 
 REGISTER_DIALOG(game_load)
 
+bool game_load::execute(const game_config_view& cache_config, savegame::load_game_metadata& data)
+{
+	if(savegame::save_index_class::default_saves_dir()->get_saves_list().empty()) {
+		bool found_files = false;
+		for(const auto& dir : filesystem::find_other_version_saves_dirs()) {
+			if(!found_files) {
+				// this needs to be a shared_ptr because get_saves_list() uses shared_from_this
+				auto index = std::make_shared<savegame::save_index_class>(dir.path);
+				found_files = !index->get_saves_list().empty();
+			}
+		}
+
+		if(!found_files) {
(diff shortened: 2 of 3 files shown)


stevecotton:
I haven't added anything to the changelog for this, I think it's just a small change.

Pentarctagon (changes requested):
Two issues that I noticed after testing locally:
* When deleting the last file for the current version, the Load Game dialog closes automatically.  I think it should stay open instead since the player might want to go look at another version's save files - I'd guess that the previous assumption when this dialog was designed was that once there are no save files left to display there's no longer any reason to show the dialog, but that assumption no longer necessarily holds true.
* When switching from the current version's saves to another version's saves the Delete button becomes disabled, but it should be allowed to delete another version's saves from the current version.  This happens regardless of whether or not the current version has any save files present.

stevecotton:
> When deleting the last file for the current version, it should stay open

Done.

> When switching to another version's saves the Delete button becomes disabled

I feel this should be done in a separate PR, as the logic of whether to deactivate the button is in `save_index_class::read_only()` and there's a corresponding guard in `save_index_class::delete_game()`. If I recall correctly from writing b66bdb8a1, that was added thinking about whether to create new saved games a directory where the current version of Wesnoth shouldn't update the save index, rather than thinking about deleting files.

However, once we start deleting files in the other versions' directories there will be futher questions of what should be done when the last file in the directory is deleted:
* should the directory itself be deleted?
* does there need to be more logic in `filesystem::find_other_version_saves_dirs()` to ignore empty directories?

Pentarctagon:
I'm fine with another PR.  For the two questions, I would say that the answer to both is no - for the first because if the player ever goes looking for the saves directory they then wouldn't be able to find it anymore, and for the second because I'd guess we'd get people asking why a particular version's directory isn't listed if we decide to hide it.

Pentarctagon (changes requested):
The popup doesn't seem to be appearing when I have no save files at all.  Tested with empty save directories present for both 1.15 and 1.14.

stevecotton:
Fixed, it will now ignore empty directories when deciding whether to show the pop-up.

Pentarctagon (approved):

PR #5566, opened 2021-02-21
Fixup WC2 validation (combination of 41b2b0b with e47b759)

stevecotton:
The combination of disabling and validating WC2 meant that the build failed,
so add another #def that will enable it for validation.


CelticMinstrel:
Well, I guess this is fine. The other option would be to re-enable it until the next release.

Pentarctagon (approved):

PR #5567, opened 2021-02-22
Support for TLS-encrypted connections

loonycyborg:
This adds backwards-compatible support for using TLS both to wesnothd and campaignd as well as appropriate clientside support.

To enable TLS support in servers extra config options to their config files are added:

tls_enabled=true
tls_fullchain="path/to/fullchain.pem"
tls_private_key="path/to/privkey.pem"
tls_dh="path/to/dhparam.pem"

fullchain and private key files can be generated via let's encrypt using an acme protocol client. Acme domain must match server's domain name clients will connect to. dh file can be generated with appropriate openssl command.

if `tls_enabled=false` server will still accept unencrypted connections from clients both with and without this PR merged.
In any case servers are compatible with clients without this PR merged, at least as basic unencrypted connectivity is concerned.

Client always supports TLS and will use it whenever appropriate, using unencrypted connection if TLS cannot be used. Client  uses RFC 2818 host verification. So client won't try to use TLS when connecting to raw IP, or when connecting to loopback network interface.


src/network_asio.cpp
```diff
@@ -78,12 +80,15 @@ connection::connection(const std::string& host, const std::string& service)
...
 	, bytes_read_(0)
 {
-#if BOOST_VERSION >= 106600
-	resolver_.async_resolve(host, service,
-#else
-	resolver_.async_resolve(boost::asio::ip::tcp::resolver::query(host, service),
-#endif
-		std::bind(&connection::handle_resolve, this, std::placeholders::_1, std::placeholders::_2));
+	boost::system::error_code ec;
+	auto result = resolver_.resolve(host, service, boost::asio::ip::resolver_query_base::numeric_host, ec);
```
Vultraz, line 84: Don't you still need this BOOST_VERSION conditional?

src/server/campaignd/server.cpp
```diff
@@ -449,45 +449,60 @@ void server::load_config()
...
 
 	LOG_CS << "Loaded addons metadata. " << addons_.size() << " addons found.\n";
+
+	load_tls_config(cfg_);
 }
 
 std::ostream& operator<<(std::ostream& o, const server::request& r)
 {
-	o << '[' << r.addr << ' ' << r.cmd << "] ";
+	o << (utils::get_if<tls_socket_ptr>(&r.sock) ? "TLS" : "") << '[' << r.addr << ' ' << r.cmd << "] ";
```
Vultraz, line 458: Here I recommend `utils::holds_alternative` (alias for [`std::holds_alternative`](https://en.cppreference.com/w/cpp/utility/variant/holds_alternative)) since you're not using the value, only checking what type it holds.

src/network_asio.cpp
```diff
@@ -109,28 +114,91 @@ void connection::handle_connect(const boost::system::error_code& ec, endpoint en
...
+		done_ = true;
+	}
+}
+
+void connection::fallback_to_unencrypted()
+{
+	assert(use_tls_ == true);
+	use_tls_ = false;
+
+	boost::asio::ip::tcp::endpoint endpoint { utils::get<raw_socket>(socket_).remote_endpoint() };
```
Vultraz, line 197: How come you don't use `visit` here like you do in other places (and on the next few lines)?

src/server/common/server_base.cpp
```diff
@@ -261,11 +306,41 @@ void server_base::coro_send_doc(socket_ptr socket, simple_wml::document& doc, bo
...
 
-void server_base::coro_send_file(socket_ptr socket, const std::string& filename, boost::asio::yield_context yield)
+template<class SocketPtr> void server_base::coro_send_file(SocketPtr socket, const std::string& filename, boost::asio::yield_context yield)
 {
+	// We fallback to userspace if using TLS socket because sendfile is not aware of TLS state
+	// TODO: keep in mind possibility of using KTLS instead. This seem to be available only in openssl3 branch for now
+	if constexpr (utils::decayed_is_same<tls_socket_ptr, decltype(socket)>) {
+		coro_send_file_userspace(socket, filename, yield);
+		return;
+	} else {
```
Vultraz, line 342: Do you need this `else` block? If so, it isn't indented, nor is the `return` needed in the first block. Also wonder if this might not be better as a template specialization rather than a constexpr-if check.

`decltype` isn't needed, you can just use `SocketPtr` directly.

src/server/common/server_base.cpp
```diff
@@ -317,12 +392,17 @@ void server_base::coro_send_file(socket_ptr socket, const std::string& filename,
...
 
 #elif defined(_WIN32)
 
-void server_base::coro_send_file(socket_ptr socket, const std::string& filename, boost::asio::yield_context yield)
+template<class SocketPtr> void server_base::coro_send_file(SocketPtr socket, const std::string& filename, boost::asio::yield_context yield)
 {
+	if constexpr (utils::decayed_is_same<tls_socket_ptr, decltype(socket)>) {
+		coro_send_file_userspace(socket, filename, yield);
+		return;
+	} else {
```
Vultraz, line 405: Likewise here as above.

src/wesnothd_connection.cpp
```diff
@@ -160,12 +178,70 @@ void wesnothd_connection::handle_handshake(const error_code& ec)
...
+	}
+}
+
+// worker thread
+void wesnothd_connection::fallback_to_unencrypted()
+{
+	assert(use_tls_ == true);
+	use_tls_ = false;
+
+	boost::asio::ip::tcp::endpoint endpoint { utils::get<raw_socket>(socket_).remote_endpoint() };
```
Vultraz, line 240: Likewise here, why not use a visitor?

src/wesnothd_connection.hpp
```diff
@@ -147,8 +149,15 @@ class wesnothd_connection
...
-	typedef boost::asio::ip::tcp::socket socket;
-	socket socket_;
+	boost::asio::ssl::context tls_context_;
+
+	std::string host_;
+	std::string service_;
+	typedef boost::asio::ip::tcp::socket raw_socket;
+	typedef boost::asio::ssl::stream<raw_socket> tls_socket;
+	typedef utils::variant<raw_socket, tls_socket> any_socket;
+	bool use_tls_ = true;
```
Vultraz, line 159: I think if you use inline initialization you need to provide an inline initializer for all members. I would not be opposed to that, though.

Vultraz (changes requested):

src/network_asio.cpp
loonycyborg: socket_ is guaranteed to always have raw_socket at this point.

Vultraz: Right, but you can still use a visitor. You just happen to know what the type passed in will be.

src/wesnothd_connection.hpp
Vultraz: Though in this case it would probably just be easier to move this to the initializer list.

src/network_asio.cpp
loonycyborg: But get is simpler. And visitor just handle both cases generating unneeded extra code for tls branch.

Pentarctagon:
It looks like the Ubuntu 18.04 and macOS jobs are failing, so I'd guess there some issue relating to C++17?

loonycyborg:
> It looks like the Ubuntu 18.04 and macOS jobs are failing, so I'd guess there some issue relating to C++17?

The issue is that boost.variant has no .emplace()

loonycyborg:
While in case of Ubuntu 18.04 it's mostly due to old version of asio requiring yet more conditional compilation.

Pentarctagon (approved):

Pentarctagon:
Iris and Soliton can definitely give a more in depth look at how this affects their areas, but I'm good with this.

Pentarctagon:
Also, once merged, we do need to make sure at least one person using each Linux, macOS, and Windows tests out the add-ons server and MP server.

loonycyborg:
Which documentation should be updated to contain new config file switches?

Pentarctagon:
campaignd has [this](https://wiki.wesnoth.org/Campaignd#server.cfg); I don't know if wesnothd has similar documentation somewhere or not.  Adding doxygen comments to the respective member variables would be good too.

shikadiqueen:
I'm curious exactly what is supposed to happen if SIGHUP or the `reload` command are sent to campaignd and the TLS configuration WML has changed from the original values. Note that `wesnothd` also accepts SIGHUP with the same semantics.

Other than that, no objections on the campaignd side, although I haven't tested the PR yet. Could you provide me with fool-proof instructions for generating and configuring a certificate with campaignd?

loonycyborg:
`reload` will reinitialize tls context used for new connections at very least. I think it shouldn't break existing connections or change their cert/tls context but this needs testing. To generate cert for wesnothd/campaignd own DNS name is needed, I have a free DNS name from freedns.afraid.org and just used acme.sh to get cert for it.

shikadiqueen:
Does that mean you can't generate and self-sign a certificate using openssl or something instead?

loonycyborg:
I suppose you can make own cert, and make your system trust it. But that cert would need to have appropriate field for domain name so host-based authentication would work.

shikadiqueen:
I'm not sure how to test this then.

loonycyborg:
You can most definitely test unencrypted path.

Pentarctagon:
Is testing only without encryption enough?

src/network_asio.cpp
```diff
@@ -104,42 +118,99 @@ void connection::handle_connect(const boost::system::error_code& ec, endpoint en
...
+			// immediate disconnect likely means old server not supporting TLS handshake code
+			fallback_to_unencrypted();
+			return;
+		}
+
 		throw system_error(ec);
 	}
+	
+	if(use_tls_) {
+		if(handshake_response_.num == 0xFFFFFFFFU) {
```
soliton-, line 158: Would be nice to document the special handshake values we have so far and what they mean. And/or a little overview documentation on the handshake procedure. There is one [here](http://wiki.wesnoth.org/MultiplayerServerWML#The_handshake) which we can update but I think in the sources somewhere would be good too.

src/server/campaignd/server.cpp
```diff
@@ -466,45 +466,60 @@ void server::load_config()
...
 	}
 #endif
+
+	load_tls_config(cfg_);
 }
 
 std::ostream& operator<<(std::ostream& o, const server::request& r)
 {
-	o << '[' << r.addr << ' ' << r.cmd << "] ";
+	o << (utils::holds_alternative<tls_socket_ptr>(r.sock) ? "TLS" : "") << '[' << r.addr << ' ' << r.cmd << "] ";
```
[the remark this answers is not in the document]
shikadiqueen: I was in fact planning to change it to `[+xxx.xxx.xxx.xxx request_command_name]` at some point afterwards.

loonycyborg:
> Is testing only without encryption enough?

Only was implying that unencrypted codepath needs testing too. Eventually we'll have TLS on server, at that point everyone will be able to test it fully.

src/network_asio.cpp
```diff
@@ -78,23 +77,38 @@ connection::connection(const std::string& host, const std::string& service)
...
 
+connection::~connection()
+{
+	if(auto socket = utils::get_if<tls_socket>(&socket_)) {
+		boost::system::error_code ec;
+		// this sends close_notify for secure connection shutdown
+		(*socket)->async_shutdown([](const boost::system::error_code&) {} );
+		const char buffer[] = "";
+		// this write is needed to trigger immediate close instead of waiting for other side's close_notify
+		boost::asio::write(**socket, boost::asio::buffer(buffer, 0), ec);
```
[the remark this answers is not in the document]
loonycyborg: Yes, that was recommended way when there is no need to keep connection after TLS shutdown. If disconnect was originally triggered by other side going offline then they will never send close_notify making us wait for timeout.

src/network_asio.hpp
```diff
@@ -103,6 +102,14 @@ class connection
 		return done_;
 	}
 
+	/** True if connection is currently using TLS and thus is allowed to send cleartext passwords or auth tokens */
+	bool using_tls() const
+	{
+		// Calling this function before connection is ready may return wrong result
+		assert(done_);
```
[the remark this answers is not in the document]
loonycyborg: This variable signals when it it is possible to initiate new high level operations, so basically not when connection is being established and not when high level operation is already in progress. I agree that current way is kinda confusing but improving this is out of scope of this PR.

src/server/campaignd/server.cpp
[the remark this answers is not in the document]
loonycyborg: Can you show an example of how log messages for connections with TLS should look?

src/server/common/server_base.cpp
```diff
@@ -128,19 +130,57 @@ void server_base::serve(boost::asio::yield_context yield, boost::asio::ip::tcp::
...
+
+	async_read(*socket, boost::asio::buffer(protocol_version.buf), yield[error]);
 	if(check_error(error, socket))
 		return;
 
-	if(memcmp(handshake.get(), "\0\0\0\0", 4) != 0) {
-		ERR_SERVER << client_address(socket) << "\tincorrect handshake\n";
-		return;
-	}
+	switch(ntohl(protocol_version.number)) {
```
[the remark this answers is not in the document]
loonycyborg: Only reason to use union here is to avoid warning from gcc about type-punning. It's consistent with other places in the code that use same trick and writing/reading to different union members is the core of it.

src/network_asio.cpp
loonycyborg: The question is where? In server_base probably?

src/server/campaignd/server.cpp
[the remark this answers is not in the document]
loonycyborg: Probably would be best if way to show TLS being used is same in both wesnothd and campaignd.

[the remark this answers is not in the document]
loonycyborg: Could add "+" prefix to all log messages if it's considered clear enough.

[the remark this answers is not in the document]
soliton-: Prefixing IPs with a plus in wesnothd as well is fine with me.

src/server/common/server_base.cpp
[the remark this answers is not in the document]
soliton-: https://en.cppreference.com/w/cpp/language/union#Explanation says: 

> It's undefined behavior to read from the member of the union that wasn't most recently written.

So unless all usable compilers feature a non-standard extension that makes this defined it doesn't seem like something we should do.

src/network_asio.cpp
soliton-: Well, in the documentation of a function that does the handshake or the class that contains that function. So server_base sounds good.

src/server/common/server_base.cpp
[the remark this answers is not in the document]
loonycyborg: My solution was influenced by https://stackoverflow.com/questions/98650/what-is-the-strict-aliasing-rule or some precursor of it. We might find another solution but it's out of scope for this PR.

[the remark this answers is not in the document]
AI0867: > https://en.cppreference.com/w/cpp/language/union#Explanation says:
> 
> > It's undefined behavior to read from the member of the union that wasn't most recently written.
> 
> So unless all usable compilers feature a non-standard extension that makes this defined it doesn't seem like something we should do.

When it comes to sockets-related things, that appears to essentially be the case.

Nevertheless, relying on union type-punning, host byte order and `ntohl`/`htonl` is silly when you can simply explicitly define the protocol byte order instead.

soliton- (approved):
I would still like to see documentation added and the out of scope issues dealt with eventually. Apart from that this looks good to me.

src/server/common/server_base.cpp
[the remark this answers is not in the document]
loonycyborg: And how exactly would such explicit definition be implemented? You mean comparing exact byte patterns instead of casting them to numbers?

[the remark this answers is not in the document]
AI0867: No, you do convert them to numbers, but without ntohl or type-punning.

Something like `uint32 addr = uint32(byte0) << 24 | uint32(byte1) << 16 | uint32(byte2) << 8 | uint32(byte3);` where each byte is a uint8, explicitly defines the conversion without any type-punning or byte order dependencies.

Wrap it and the inverse in a function, and you're done.

You can then do stuff like 

    read_into(buf)
    switch (big_endian_uint32(buf)) {

with hopefully better names than I came up with.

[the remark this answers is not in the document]
loonycyborg: That just seems more error prone to me. Sending a single number in network byte order is equally good protocol definition and bit sizes are fixed too due to use of appropriate types. htonl/ntohl calls are more maintainable than manual bit shifts.

Pentarctagon:
It looks like some trailing whitespace was introduced, which is why the CI failed on the last commit.

loonycyborg:
I don't see where though. Generally github diff highlights them.

shikadiqueen (approved):

src/server/campaignd/server.cpp
```diff
@@ -1143,9 +1164,11 @@ void server::handle_request_campaign(const server::request& req)
 		}
 
 		LOG_CS << req << "Sending add-on '" << name << "' version: " << to << " size: " << full_pack_size / 1024 << " KiB\n";
-		boost::system::error_code ec;
-		coro_send_file(req.sock, full_pack_path, req.yield[ec]);
-		if(check_error(ec, req.sock)) return;
+		utils::visit([this, &req, &full_pack_path](auto&& socket) {
+			boost::system::error_code ec;
+			coro_send_file(socket, full_pack_path, req.yield[ec]);
+			if(check_error(ec, socket)) return;
```
Vultraz, line 1170: Why a return statement at the end of the function?

src/server/campaignd/server.cpp
```diff
@@ -1197,9 +1220,11 @@ void server::handle_request_campaign_hash(const server::request& req)
 		}
 
 		LOG_CS << req << "Sending add-on hash index for '" << req.cfg["name"] << "' size: " << file_size / 1024 << " KiB\n";
-		boost::system::error_code ec;
-		coro_send_file(req.sock, path, req.yield[ec]);
-		if(check_error(ec, req.sock)) return;
+		utils::visit([this, &path, &req](auto&& socket) {
+			boost::system::error_code ec;
+			coro_send_file(socket, path, req.yield[ec]);
+			if(check_error(ec, socket)) return;
```
Vultraz, line 1226: Likewise here.

src/server/campaignd/server.cpp
loonycyborg: Because it was standard pattern in code before? But yeah it probably indicates some confusion here. Actually I'm not even sure about the best way to make those visitors interrupt calling function.

loonycyborg: If I throw instead then exception will actually originate from asio's event loop and would need additional handling to prevent them from stopping the loop.

loonycyborg: I fixed those cases of unintentionally ignoring errors. Though I'll most likely eventually make error handling everywhere exception based.

PR #5571, opened 2021-02-26
Add a Wesnoth MSVC visualization file for config and t_string

CelticMinstrel:
This should greatly improve the experience of debugging Wesnoth code in MSVC. It also works in VSCode if you configure it to run MSBuild, as I have done.

The natvis file added in this commit needs to be added as a reference in the Visual Studio project file in order to work in Visual Studio (according to the documentation). Since I don't have Visual Studio and don't know what the MSBuild code would look like to add this, I've left it to someone else who does.

### DIFF
--- a/src/tstring.cpp
+++ b/src/tstring.cpp
@@ -42,10 +42,10 @@ const char UNTRANSLATABLE_PART = 0x02;
 const char TEXTDOMAIN_SEPARATOR = 0x03;
 const char ID_TRANSLATABLE_PART = 0x04;
 const char PLURAL_PART = 0x05;
-
-std::vector<std::string> id_to_textdomain;
-std::map<std::string, unsigned int> textdomain_to_id;
 }
+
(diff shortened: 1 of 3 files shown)


src/tstring.cpp
```diff
@@ -42,11 +42,11 @@ const char UNTRANSLATABLE_PART = 0x02;
 const char TEXTDOMAIN_SEPARATOR = 0x03;
 const char ID_TRANSLATABLE_PART = 0x04;
 const char PLURAL_PART = 0x05;
-
-std::vector<std::string> id_to_textdomain;
-std::map<std::string, unsigned int> textdomain_to_id;
 }
 
+std::vector<std::string> t_string_base::id_to_textdomain;
```
Vultraz, line 47: Use `static inline` please, if possible.

Vultraz (changes requested):

Vultraz (approved):

PR #5573, opened 2021-02-27
AI: Improve behavior of the retreat_injured CA

mattsc:
The previous version of the CA had two massive problems:

- The threshold for retreating was based on the `caution` aspect.  This would make almost fully healthy units retreat if caution was set to a large value for other reasons.  This is fixed by introducing a dedicated aspect `retreat_factor`.  In addition, the retreat threshold is now based on the unit's hitpoints, rather than its level.
- In the presence of enemies, units sometimes retreated all the way to the edge of the map.  This is fixed by improving the retreat location evaluation, and by introducing an new aspect `retreat_enemy_weight`.  The new behavior is as follows:
  - Enemy threats are much less important, but their weight can be increased using `retreat_enemy_weight`
  - By default, only healing locations are considered as potential retreat locations; this can be overridden by setting `retreat_enemy_weight` to negative values, in which case all hexes are considered (and the previous behavior of running away from enemies may be restored, depending on the absolute value of `retreat_enemy_weight`
  - If `retreat_enemy_weight` is set to zero, enemies are not taken into account at all

In the process, I also found and fixed a bug that resulted in hexes next to healers (own and allied) not being counted as healing locations


mattsc:
I consider this complete at this time.  I am going to let it sit here for a few days to see if there are comments.  I'll also link to this PR in the [forum post discussing the issues](https://forums.wesnoth.org/viewtopic.php?f=6&t=53614).

This should definitely be merged before 1.15.11 is released

src/ai/lua/core.cpp
```diff
@@ -498,6 +498,14 @@ static int cfun_ai_get_recruitment_pattern(lua_State *L)
 	return 1;
 }
 
+static int cfun_ai_get_retreat_factor(lua_State *L)
```
CelticMinstrel, line 509: I would've suggested omitting this, since these are supposed to be deprecated and this one even _is_ deprecated...

src/ai/lua/core.cpp
```diff
@@ -498,6 +498,14 @@ static int cfun_ai_get_recruitment_pattern(lua_State *L)
 	return 1;
 }
 
+static int cfun_ai_get_retreat_enemy_weight(lua_State *L)
```
CelticMinstrel, line 501: Same here, kinda silly to add something already deprecated.

src/ai/lua/core.cpp
mattsc: Taken care of in 200f5dd913

src/ai/lua/core.cpp
mattsc: Also done in 200f5dd913

Issue #2710, opened 2018-03-22
User config and data migration across branches


irydacea:
One common complaint about Wesnoth's development approach is that players are expected to lose all their user configuration and data (add-ons, saves) when upgrading from one stable branch to the next one, or migrate them by hand and brace for the issues that can result from:

* Trying to run Wesnoth with preferences from an older version (which *usually* works just fine, but issue #2612 is an example of how this can go wrong due to oversights)
* Using obsolete add-ons (which usually have been updated to the newer stable branch, but the player has to go to the Add-ons Manager to reinstall them by hand or upgrade them at once if they copied over their `data/add-ons` dir themselves)
* Loading older saves (start-of-scenario saves *usually* work just fine, but this is not always the case if the attached WML in units has changed in the meantime, e.g. abilities and other data that isn't defined/overridden by the engine)

Even though it's easy to shrug it off and just decide that the clean-slate approach for new stable branches is part of the project's development philosophy somehow, once Wesnoth is released on Steam we can expect more vocal members of the audience (e.g. game reviewers) to complain rather loudly once the 1.16 stable series lands and people suddenly have lost all their gameplay progress and profile configuration from 1.14 unless they go out of their way to read instructions on a wiki/forum post and do file management operations to get their stuff back -- or just give up, depending on the case.

So, in my opinion, going forward, some form of automated or semi-automated migration task will be necessary to avoid this issue from cropping up every time and demotivating players who've spent countless hours playing the game to get 100% SP campaign completion or what have you.

A potential implementation for this would be as follows:

* On first start-up, if Wesnoth detects its user config and data dir are **both** empty or non-existent (obvious sign of the first launch of the new branch), it should look for the dir(s) for the previous stable branch and offer the user the option to migrate their profile from the previous version. If the user declines, then the game continues with a clean slate like it currently does.
* If the user opts to migrate their profile, then saves, preferences, and credentials are copied (or moved?) over to the new version's profile. The game should then gather a list of add-ons installed on the old version (currently this is a trivial directory enumerate operation), connect to the add-ons server, and see which add-ons are available for the new version and offer automatically downloading them.
* Before finishing and loading the preferences file, the game should perform any syntax/structure changes that may be required on the new profile's copy of it. If for some reason the process fails, inform the user accordingly and just proceed with fresh preferences (but this should never happen).
[...]

PR #5574, opened 2021-02-27
Unsubtly warn about subtle problems loading saved games from old versions

stevecotton:
This connects to issue #2710, although the main reasoning about it is in #5517.
The difficulty balance isn't the only problem with loading old saved games, but
it's a simple example to explain.

This warning is intended for the change between 1.14 and 1.16, but will be shown
to anyone loading 1.15.x saves into 1.15.y - that's something that they should
expect in a dev branch.

### DIFF
--- a/src/savegame.cpp
+++ b/src/savegame.cpp
@@ -257,11 +257,22 @@ bool loadgame::check_version_compatibility(const version_info& save_version)
 
 	if(preferences::confirm_load_save_from_different_version()) {
 		const std::string message
-			= _("This save is from a different version of the game ($version_number|). Do you wish to try to load it?");
+			= _("This save is from a different version of the game ($version_number|), and might not work with this "
+				"version.\n"
+				"\n"
+				"<b>Warning:</b> saves in the middle of campaigns are especially likely to fail, and you should either "
+				"use the old version or restart the campaign. Even when a saved game seems to load successfully, "
+				"subtler aspects like gameplay balance and story progression could be impacted. The difficulty, the "
+				"challenge, the <i>fun</i> may be missing.\n"
+				"\n"
+				"For example, the campaign may have been rebalanced with fewer enemies in the early scenarios, but "
+				"expecting your recall list to have correspondingly less experience in the late scenarios.\n"
+				"\n"
+				"Do you wish to continue?");
 		utils::string_map symbols;
 		symbols["version_number"] = save_version.str();
 		const int res = gui2::show_message(_("Load Game"), utils::interpolate_variables_into_string(message, &symbols),
-			gui2::dialogs::message::yes_no_buttons);
+			gui2::dialogs::message::yes_no_buttons, true);
 		return res == gui2::retval::OK;
 	}
 
--- a/changelog.md
+++ b/changelog.md
@@ -24,6 +24,7 @@
    * Improved the Load dialog when there are no saved games for the current version, and also when there are corrupted files.
    * Re-added the pop-up when there are no saved games at all (issue #5517).
    * Fixed resource leak in the Pango text rendering pipeline when using the characters_per_line constraint.
+   * Make the warning about loading saves from old versions much clearer.
  ### WML Engine
    * Standard Location Filters now support gives_income=yes|no to make it simpler to match villages regardless of owner
  ### Miscellaneous and Bug Fixes


Pentarctagon:
This warning shouldn't be worded to be so specific to campaigns - every save file can have the same kinds of issues.

shikadiqueen:
> <b>Warning:</b> for saves in the middle of campaigns, you should either use the old version or restart the campaign. Even when a saved game seems to load successfully, there are many ways in which it could go subtly wrong. The difficulty, the challenge, the <i>fun</i> may be missing.
>
> For example, the difficulty levels may have been rebalanced — if you find it’s too easy, it might just be that you’ve become a better player since you last played the campaign, but it might also be that your recalls are stronger than the current implementation of the campaign expects.
>
> Do you wish to try to load it?

I'm not a fan of using contractions in the Wesnoth UI and I purposefully avoid it at all times, even though Vultraz seems to have a different mind about it. It *is* a very unusually informal piece as evidenced by the end of the second paragraph.

I do feel like the whole third paragraph could be done away with to shorten the full message a bit while still conveying the basic points:

> <b>Warning:</b> for saves in the middle of campaigns, you should either use the old version or restart the campaign. Even when a saved game seems to load successfully, subtler aspects like gameplay balance and story progression could be impacted. The difficulty, the challenge, the <i>fun</i> may be missing.

shikadiqueen:
> This warning shouldn't be worded to be so specific to campaigns - every save file can have the same kinds of issues.

This is true, but I reckon the odds that someone will load a non-campaign non-replay save for an older version of the game are extremely low.

Pentarctagon:
> This is true, but I reckon the odds that someone will load a non-campaign non-replay save for an older version of the game are extremely low.

Given that the Steam update from 1.14.x to 1.16.0 will be automatic, I don't think that's a good assumption to make.

stevecotton:
It should be possible to check if the save is part of a campaign, and only show this version of the message if it is. However, I'm not sure adding more logic here is worthwhile, particularly because it means more code paths to test.

The reason that the message focused on campaigns is that they're more likely to have problems that don't manifest until later, when the data from the saved game is combined with the next scenario's content. They're also more likely to have a player in the middle of a game, having already spent time building up their recall list.

I think @shikadiqueen's revised second paragraph is good, and will gladly take it for the next push.

While I can shorten the third paragraph, I think it shouldn't be cut entirely. We're asking players to throw away their progress on the campaign, an example saying why it can't be automatically fixed might help convince them instead of a dialog that gets forgotten followed by a negative review that the new version is too easy / too hard.

AI0867:
We might also consider adding a popup on first run of a new version, when a saves from an old version are detected, to explain what has happened, and how to go back to an old version to finish your campaigns.

stevecotton:
New wording:

<b>Warning:</b> for saves in the middle of campaigns, you should either use the old version or restart the campaign. Even when a saved game seems to load successfully, subtler aspects like gameplay balance and story progression could be impacted. The difficulty, the challenge, the <i>fun</i> may be missing.

For example, the campaign may have been rebalanced with fewer enemies in the early scenarios, but expecting your recall list to have correspondingly less experience in the late scenarios.

src/savegame.cpp
```diff
@@ -256,12 +256,21 @@ bool loadgame::check_version_compatibility(const version_info& save_version)
...
+			"This save is from a different version of the game ($version_number|).\n"
+			"\n"
+			"<b>Warning:</b> for saves in the middle of campaigns, you should either use the old version or restart "
+			"the campaign. Even when a saved game seems to load successfully, subtler aspects like gameplay balance "
+			"and story progression could be impacted. The difficulty, the challenge, the <i>fun</i> may be missing.\n"
+			"\n"
+			"For example, the campaign may have been rebalanced with fewer enemies in the early scenarios, but "
+			"expecting your recall list to have correspondingly less experience in the late scenarios.\n"
+			"\n"
+			"Do you wish to try to load it?");
```
Pentarctagon, line 269: I'd prefer "Do you wish to continue?".

Pentarctagon:
> It should be possible to check if the save is part of a campaign, and only show this version of the message if it is. However, I'm not sure adding more logic here is worthwhile, particularly because it means more code paths to test.

That misses the point - errors and other incompatibilities/issues aren't campaign-specific problems in the first place.  The warning should be worded so that it's clear that these issues affect all save files from other versions, not just campaigns.  Otherwise players are going to see it and think that if they aren't continuing a campaign then everything should be fine.

stevecotton:
> That misses the point - errors and other incompatibilities/issues aren't campaign-specific problems in the first place.

If they get an OOS dialog during the first turn then I think the wording in this dialog doesn't matter. 

I'm failing to see what else should be covered:

* not a campaign
* doesn't have an obvious error in the first few turns
* isn't too obscure to warn about
* doesn't give the player a decent amount of fun despite the bug

There might be an event in a later turn that fails, but that could also happen due to a bug in the scenario.

Pentarctagon:
As an example then, if a scenario tries to spawn a unit using the Dunefolk names that were deprecated in 1.14 and removed in 1.15, then the players will not get any indication of issues until the game tries to actually spawn the units at which point they will get an error and no units will spawn.

There are also longer single scenarios that players could spend many hours playing before they reach a bug caused by loading the save from 1.14 using 1.16.  Campaigns are **more** affected by compatibility issues, but **all** save files can be affected by them.  The warning message needs to reflect that.

stevecotton:
I doubt that any warning message would help in those situations - the player is still going to try loading it and then play until the error shows up.

I'm not sure restarting on 1.16 would always help either - the add-on's author might have uploaded to 1.16 after basic testing, but before finding such a bug that only occurs late in the game.

Pentarctagon:
> I doubt that any warning message would help in those situations - the player is still going to try loading it and then play until the error shows up.

If that's what you believe players will do regardless of a warning, then what's the purpose of this PR?

stevecotton:
With a message starting "Warning: for saves in the middle of campaigns,", I believe a lot of players will pay attention to the message. By making it generic enough to cover any saved game, I believe it the warning becomes boilerplate to click though without reading.

Pentarctagon:
How about:
```
Warning: Any save created by another version of Wesnoth may not work.
For saves in the middle of campaigns especially, you should either use the old version or restart 
the campaign. Even when a saved game seems to load successfully, subtler aspects like gameplay balance 
and story progression could be impacted. The difficulty, the challenge, the fun may be missing.

For example, the campaign may have been rebalanced with fewer enemies in the early scenarios, but 
expecting your recall list to have correspondingly less experience in the late scenarios.
```

stevecotton:
I'd like to keep "campaign" close to the bold "Warning:". How about putting the "might not work" on the first line instead?

This save is from a different version of the game (`$version_number|`), and might not work with this version.

<b>Warning:</b> saves in the middle of campaigns are especially likely to fail, and you should either use the old version or restart the campaign. Even when a saved game seems to load successfully, subtler aspects like gameplay balance and story progression could be impacted. The difficulty, the challenge, the <i>fun</i> may be missing.

For example, the campaign may have been rebalanced with fewer enemies in the early scenarios, but expecting your recall list to have correspondingly less experience in the late scenarios.

Do you wish to continue?

Pentarctagon:
Looks fine to me.

stevecotton:
Force-pushed to fix the merge conflict in the changelog.

I think this is ready to merge, @shikadiqueen are you ok with the current wording?

shikadiqueen:
It works I guess.

stevecotton:
Thanks - in that case, even if we reword it later, I think it's time to merge.

PR #5577, opened 2021-02-28
Some enhancements to the WML test runner

CelticMinstrel:
This is a bunch of updates to the WML test runner that I did while trying to figure out the test failures in #4580. The general motivation is that I was unable to get the output from the test that failed or even identify _which_ test that failed when running on Windows.

The main purpose of this PR is to verify that the runner still works on POSIX systems, but I'm also open to review comments.

* The first commit just fixes the new command-line argument I recently added on master - apparently I forgot to test it and it didn't work, just crashed the runner.
* In order to reduce clutter when displaying test output, I've added a `--nobanner` flag to Wesnoth. This could be probably achieved with `tail` or similar tools on POSIX systems, but that won't work on Wesnoth. The flag suppresses all or most output up to the "Checking lua scripts..." line.
* Showing the numeric test return code is kind of silly, especially when you're using Python enums that know their own name, so now a failed test shows the actual meaning of the return codes (in addition to the number)
* The test runner now uses a pipe instead of a temporary file to capture the output. If you pass `-v` (which previously did nothing), you can see the full output from every test. This avoids cluttering the repository root directory with all the test result files. It doesn't use the `capture_output` parameter that the code comment mentioned because that captures stdout and stderr separately whereas I wanted to capture them merged together.
* The test runner was unable to capture the output from the test on Windows, due to the log redirection, so I've added `--wnoredirect` to Windows builds to disable it. I've also added `--wnoconsole` to disable the automatic console attachment if you're using a flag that normally implies it; I suspect that wasn't actually needed for my purpose but I figure it can't hurt to leave it in.
* The test runner can now limit the size of a test batch. If we ever run into a problem where we have so many tests that the batch times out, we can use this. At the moment, I don't think we have enough tests for it to matter.
* The test runner claimed to skip tests that expect timeout when timeout is disabled, but it didn't actually do so. This is fixed.
* The test runner now prints a summary of tests passed and failed at the end. It also includes a count of tests that crashed (due to a signal) or were skipped (due to expecting timeout when timeout is disabled).
* When told to run multiple tests, the Wesnoth executable exits on the first failure. The test runner now determines which test fails and starts a new batch to continue where it left off.

### DIFF
--- a/run_wml_tests
+++ b/run_wml_tests
@@ -34,6 +34,30 @@ class TestCase:
     def __str__(self):
         return "TestCase<{status}, {name}>".format(status=self.status, name=self.name)
 
+class TestResultAccumulator:
+    passed = 0
+    skipped = 0
+    failed = 0
+    crashed = 0
+    
+    def __init__(self, total):
+        self.total = total
+    
+    def pass_test(self, n = 1):
+        self.passed = self.passed + n
+    
+    def skip_test(self, n = 1):
+        self.skipped = self.skipped + n
+    
+    def fail_test(self, n = 1):
+        self.failed = self.failed + n
+    
+    def crash_test(self, n = 1):
+        self.crashed = self.failed + n
+    
+    def __bool_(self):
+        return self.passed + self.skipped == self.total
+
 class TestListParser:
     """Each line in the list of tests should be formatted:
         <expected return code><space><name of unit test scenario>
@@ -63,7 +87,7 @@ class TestListParser:
             if self.verbose > 1:
                 print(t)
             test_list.append(t)
-        return batcher(test_list)
+        return batcher(test_list), TestResultAccumulator(len(test_list))
 
 def run_with_rerun_for_sdl_video(args, timeout):
     """A wrapper for subprocess.run with a workaround for the issue of travis+18.04
@@ -120,7 +144,7 @@ class WesnothRunner:
         if self.verbose > 1:
             print("Options that will be used for all Wesnoth instances:", repr(self.common_args))
 
-    def run_tests(self, test_list):
+    def run_tests(self, test_list, test_summary):
         """Run all of the tests in a single instance of Wesnoth"""
         if len(test_list) == 0:
             raise ValueError("Running an empty test list")
@@ -135,6 +159,7 @@ class WesnothRunner:
             args.append(test.name)
         if self.timeout == 0:
             if test.status == UnitTestResult.TIMEOUT:
+                test_summary.skip_test()
                 print('Skipping test', test_list[0].name, 'because timeout is disabled')
                 return
             timeout = None
@@ -159,6 +184,14 @@ class WesnothRunner:
             print(res.stdout.decode('utf-8'))
             if self.verbose > 1:
                 print("Result:", res.returncode)
+        returned_result = UnitTestResult(res.returncode)
+        num_passed = 0
+        if test_list[0].status == UnitTestResult.PASS:
+            num_passed = res.stdout.count(b"PASS TEST")
+            test_summary.pass_test(num_passed)
+        elif returned_result == expected_result:
+            num_passed = 1
+            test_summary.pass_test()
         if res.returncode < 0:
             print("Wesnoth exited because of signal", -res.returncode)
             if options.backtrace:
@@ -166,12 +199,15 @@ class WesnothRunner:
                 gdb_args = ["gdb", "-q", "-batch", "-ex", "start", "-ex", "continue", "-ex", "bt", "-ex", "quit", "--args"]
                 gdb_args.extend(args)
                 subprocess.run(gdb_args, timeout=240)
+            test_summary.crash_test()
+            test_summary.skip_test(len(test_list) - num_passed - 1)
             raise UnexpectedTestStatusException()
-        returned_result = UnitTestResult(res.returncode)
         if returned_result != expected_result:
             if self.verbose == 0:
                 print(res.stdout.decode('utf-8'))
             print("Failure, Wesnoth returned", returned_result, "but we expected", expected_result)
+            test_summary.fail_test()
+            test_summary.skip_test(len(test_list) - num_passed - 1)
             raise UnexpectedTestStatusException()
 
 def test_batcher(test_list):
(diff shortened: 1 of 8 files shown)


run_wml_tests
```diff
@@ -159,19 +184,25 @@ class WesnothRunner:
             print(res.stdout.decode('utf-8'))
             if self.verbose > 1:
                 print("Result:", res.returncode)
+        num_passed = res.stdout.count(b"PASS TEST")
+        test_summary.pass_test(num_passed)
```
CelticMinstrel, line 188: I just realized this won't work for tests that are _expected_ to fail... posting to remind me to fix it.

src/wesnoth.cpp
```diff
@@ -87,6 +87,7 @@
 
 #include <boost/iostreams/filtering_stream.hpp> // for filtering_stream
 #include <boost/program_options/errors.hpp>     // for error
+#include <boost/optional.hpp>
```
Vultraz, line 90: Why boost::optional?

Vultraz (changes requested):

run_wml_tests
```diff
@@ -153,24 +177,37 @@ class WesnothRunner:
...
+            test_summary.skip_test(len(test_list) - num_passed - 1)
             raise UnexpectedTestStatusException()
-        if res.returncode != expected_result.value:
-            with open(get_output_filename(args), "r") as output:
-                for line in output.readlines():
-                	print(line)
-            print("Failure, Wesnoth returned", res.returncode, "but we expected", expected_result.value)
+        if returned_result != expected_result:
+            if self.verbose == 0:
+                print(res.stdout.decode('utf-8'))
```
stevecotton, line 207: This was already printed on line 186.

run_wml_tests
```diff
@@ -202,6 +251,8 @@ if __name__ == '__main__':
         help="New timer value to use for batched tests, instead of 300s as default.")
     ap.add_argument("-bd", "--batch-disable", action="store_true",
         help="Disable test batching, may be useful if debugging a small subset of tests.")
```
stevecotton, line 253: Is the `--batch-disable` option still useful, could it be done with `--batch-max=1` instead?

src/wesnoth.cpp
CelticMinstrel: Because all the other files I could find also used boost::optional. It's possible I was working on a branch before you replaced them somehow? Or I accidentally searched on the 1.14 branch, unlikely though that sounds... In any case, I can replace it.

run_wml_tests
CelticMinstrel: Your line number is off, but... the first print is only if -v is passed, this one is if -v was not passed.

run_wml_tests
CelticMinstrel: I was thinking about that too... I think `--batch-disable` is slightly more optimal than `--batch-max=1` with the current code, but that's easily fixed. I was also considering keeping `--batch-disable` but making it an alias for `--batch-max=1` instead of a separate flag.

run_wml_tests
CelticMinstrel: In other words, without -v it only prints the full output when the test fails; with -v it _always_ prints the full output.

src/wesnoth.cpp
Vultraz: Yeah, please use `std::optional` unless you're using a reference type.

CelticMinstrel: Yeah, it was the use of `boost::optional` on references that confused me, as it turns out. Thus why I replaced them all with `utils::optional_reference` on master...  now the only place `boost::optional` is still used is in spirit_po.

CelticMinstrel:
Okay, now I just need to figure out why this breaks the tests on non-Windows builds and fix it without breaking them once more on Windows builds.

stevecotton:
> Okay, now I just need to figure out why this breaks the tests on non-Windows builds and fix it without breaking them once more on Windows builds.

In the `except subprocess.TimeoutExpired as t:` block, `t.output`, `t.stdout` and `t.stderr` are all `None` (testing on Debian Unstable).

CelticMinstrel:
Hmm. The documentation said that the `TimeoutExpired` exception is supposed to include the program's output if it had been captured... but maybe that's a new feature and the Linux versions we support are running a too-old Python? Do we know specifically which version of Python needs to be supported?

But on the other hand... if they're None it means they exist. If they didn't exist you'd get `AttributeError` trying to access them... am I missing something else?

stevecotton:
> If they didn't exist you'd get `AttributeError` trying to access them... am I missing something else?

The error is a couple of lines later, when you call `None.decode ('utf-8')`.

CelticMinstrel:
Yes, I understand that, but the question is, why is it None when the `subprocess.run` call specified to capture the output?

stevecotton:
It's not an old version of Python, I'm testing with 3.9.2.

It might be considered a bug, there's an `if` in subprocess.py that checks `if stdout_seq` instead of `if stdout_seq is not None`, so treats a process that doesn't print any output as if it wasn't capturing stdout; however bug or not this PR needs to cope with it.

```
> /usr/lib/python3.9/subprocess.py(153)__init__()
-> self.cmd = cmd
(Pdb) list
148  	
149  	    Attributes:
150  	        cmd, output, stdout, stderr, timeout
151  	    """
152 B	    def __init__(self, cmd, timeout, output=None, stderr=None):
153  ->	        self.cmd = cmd
154  	        self.timeout = timeout
155  	        self.output = output
156  	        self.stderr = stderr
157  	
158  	    def __str__(self):
(Pdb) p output
None
(Pdb) p stderr
None
(Pdb) up
> /usr/lib/python3.9/subprocess.py(1178)_check_timeout()
-> raise TimeoutExpired(
(Pdb) list
1173 	                       skip_check_and_raise=False):
1174 	        """Convenience for checking if a timeout has expired."""
1175 	        if endtime is None:
1176 	            return
1177 	        if skip_check_and_raise or _time() > endtime:
1178 ->	            raise TimeoutExpired(
1179 	                    self.args, orig_timeout,
1180 	                    output=b''.join(stdout_seq) if stdout_seq else None,
1181 	                    stderr=b''.join(stderr_seq) if stderr_seq else None)
1182 	
1183 	
(Pdb) p stdout_seq
[]
(Pdb) p stderr_seq
None
```

CelticMinstrel:
Okay, looks like the tests pass now!

PR #5579, opened 2021-03-01
Support [terrain_defaults]'s original subtag names

stevecotton:
This is preparation for backporting 0ba433203e, with the idea that both sets
of names will be supported in both branches as part of fixing #5308.

No deprecation messages are added. While config::get_attribute_value() has a
config::get_old_attribute_value(), there isn't currently a similar utility for
config::child(); maybe I should add one now, but it feels too large a change
for a backport.


stevecotton:
The build failure is just a lack of disk-space.

CelticMinstrel (approved):

PR #5582, opened 2021-03-02
Handle empty children in patch_movetype

stevecotton:
In movetype::write, the children might not be created if that child has an
empty config; there might be no child at all for movetypes such as "none". The
movetype "none" is a real movetype defined in units.cfg for units that
shouldn't be able to move.

The original_cfg local variable was only used once anyway, removing it and the
comment made the code clearer.

This wasn't caught in the 1.15 testing, but with 1.14 it resulted in a
"mandatory WML child missing" warning.

### DIFF
--- a/src/units/types.cpp
+++ b/src/units/types.cpp
@@ -885,12 +885,8 @@ void patch_movetype(movetype& mt,
 	config mt_cfg;
 	mt.write(mt_cfg);
 
-	// original_cfg can be blank, if a movetype similar to "none" exists.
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
I think that code should have used `child_or_empty`… your new code is fine too, although I guess it makes the line a bit long but whatever.

stevecotton:
> I think that code should have used `child_or_empty`

Thanks, I changed it to use that.

PR #5583, opened 2021-03-02
[1.14] Fix [resistance_defaults] and [terrain_defaults] (issue #5308)

### LINKED ISSUE #5308: [resistance_defaults] and [terrain_defaults] are broken

stevecotton:
The bug introduced in fcd0ceda is fixed by ensuring that movetype::merge() is
called with the strings "movement_costs", "vision_costs" and "jamming_costs"
instead of "movement", "vision" and "jamming". Closes #5308.

Both [terrain_defaults][movement] and [terrain_defaults][movement_costs] are
supported and equivalent, as added to the master branch in 3522eb2c.

These now work again:

    [resistance_defaults]
        id="special_res_for_test"
        default="30"
    [/resistance_defaults]

    [resistance_defaults]
        id="copy_of_arcane"
        default="(arcane)"
    [/resistance_defaults]

    [terrain_defaults]
        id="special_terrain_for_test"
        [movement]
            default="(swamp_water + 1)"
            orcishfoot="(swamp_water * 2)"
        [/movement]
    [/terrain_defaults]

Formulas can now access other parts of the movetype too, as this allowed the
code to be the same as in 1.15:

    [terrain_defaults]
        id="special_terrain_for_test"
        [movement_costs]
            default="(swamp_water + 1)"
            orcishfoot="(vision_costs.swamp_water * 2)"
        [/movement_costs]
    [/terrain_defaults]
    [terrain_defaults]
        id="special_terrain_for_test"
        [defense]
            default="(20 + 7 * movement_costs.special_terrain_for_test)"
        [/defense]
    [/terrain_defaults]

The formula handling will recognise "resistance", "movement_costs",
"vision_costs", "jamming_costs" and "defense". For [resistance_defaults], the
formula will recognise both "(arcane)" and "(resistance.arcane)" as equivalent,
similarly for [terrain_defaults] "(swamp_water)" is a shorthand for whichever
subtag is being patched.

A [terrain_defaults] tag may use data added in a previous [terrain_defaults],
as in the examples above where the second tag's [defense] is based on the first
tag's [movement_costs], this gives orcish grunts on the special terrain a 62%
chance to be hit. However, relying on data in the same [terrain_defaults] that
creates or changes it is unsupported - if the [movement_costs] and [defense]
were in a single [terrain_defaults] tag then the result would be implementation
defined, because no guarantee is made of the order in which the children of the
tag are processed.

The unit tests for [terrain_defaults] and [resistance_defaults] must stay
out-of-tree until we support some method of testing them without affecting the
core units' stats during other tests. The test added here is limited to
checking the ways that WML can read the stats, it's also a framework for the
out-of-tree parts to use.

This is a squashed cherry pick of these commits from 1.15:
* 72863e578ba2153520488f096f88d024d7795340 (the unit test)
* 0ba433203e508478fff7f79f7fcb64d186ba40e7
* 3522eb2c2a13c7d78480117a83f55946a0dee2d4 (adding the aliases)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -17,6 +17,7 @@
    * Added support to wmlxgettext for double-quote characters in translatable raw strings
    * Fixed an error message and the AI leader potentially not moving when it cannot reach a keep because it's occupied by an allied unit
    * Fixed display zoom not being taken into account when using the `x`, `y`, `directional_x` and `directional_y` attributes in unit animations.
+   * Fixed `[terrain_defaults]` and `[resistance_defaults]` (issue #5308).
 
 ## Version 1.14.15
  ### Add-ons client
--- a/data/test/scenarios/test_resistances.cfg
+++ b/data/test/scenarios/test_resistances.cfg
@@ -1,0 +1,46 @@
+# wmllint: no translatables
+
+{GENERIC_UNIT_TEST "test_resistances" (
+    [event]
+        name = start
+
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=alice
+        [/store_unit]
+
+        # Elvish archers have a slight malus against arcane
+        {ASSERT {VARIABLE_CONDITIONAL alice.resistance.arcane equals 110}}
+        {ASSERT {VARIABLE_CONDITIONAL alice.resistance.impact equals 100}}
+        {ASSERT {VARIABLE_CONDITIONAL alice.resistance.length equals 1}}
+
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=bob
+        [/store_unit]
+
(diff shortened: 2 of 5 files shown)


stevecotton:
Draft for the moment, I think it will be ready once #5582 has been cherry-picked on top.

CelticMinstrel (approved):
I don't see any issues with this.

PR #5587, opened 2021-03-05
Fix several Lua issues, mostly unintentional globals

CelticMinstrel:
Submitting as a PR because this touches a lot of AI code and I'm not 100% confident all these globals were indeed unintentional.

There's some typos and stuff fixed in here too.

### DIFF
--- a/data/ai/lua/generic_recruit_engine.lua
+++ b/data/ai/lua/generic_recruit_engine.lua
@@ -29,6 +29,7 @@ return {
         local AH = wesnoth.require "ai/lua/ai_helper.lua"
         local M = wesnoth.map
         local LS = wesnoth.require "location_set"
+        local recruit_lib = {}
 
         local recruit_data = {}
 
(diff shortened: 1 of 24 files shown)


data/ai/lua/generic_recruit_engine.lua
```diff
@@ -330,12 +331,10 @@ return {
 
             my_hp = my_hp + sum_gold_for_sides({{"allied_with", {side = wesnoth.current.side} }})*2.3
             enemy_hp = enemy_hp+sum_gold_for_sides({{"enemy_of", {side = wesnoth.current.side} }})*2.3
-            hp_ratio = my_hp/(enemy_hp + 1e-6)
-
-            return hp_ratio
+            return my_hp / enemy_hp
```
mattsc, line 334: Did you remove the `1e-6` here intentionally?  It was obviously there to prevent a division by zero, and while that is very unlikely to happen, it is not entirely impossible, I think.

mattsc:
As I have said in a couple other places, I cannot test things right now because my build currently crashes, but I don't see any issues with any part of the AI code that would need to be tested.

mattsc (approved):

data/ai/lua/generic_recruit_engine.lua
CelticMinstrel: Ah, I speculated that it was to prevent truncating division and removed it on the basis that division doesn't truncate now, but if it had a different purpose, I can put it back… although, I'm not sure it's even _sensible_? It's going to produce an `hp_ratio` a million times larger than `my_hp`.

Issue #5065, opened 2020-08-10
Add a forum_auth=yes|no attribute to _server.pbl


Pentarctagon:
This would allow authenticating via the add-on author's forum username and password. Add-on authors utilizing this then wouldn't need to provide a separate email or passphrase value.

Taken from the [1.15 Roadmap](https://wiki.wesnoth.org/1.15_Roadmap#1.15.9_.2801.2F16.2F2021.29).

Issue #5061, opened 2020-08-10
Store add-on upload history in the database


Pentarctagon:
This would make it very easy to, for example, let users know if a new version of an add-on they're using is available.

Taken from the [1.15 Roadmap](https://wiki.wesnoth.org/1.15_Roadmap#1.15.8_.2812.2F19.2F2020.29).

Issue #5060, opened 2020-08-10
Add SQL support to campaignd using mariadbpp


Pentarctagon:
Taken from the [1.15 Roadmap](https://wiki.wesnoth.org/1.15_Roadmap#1.15.8_.2812.2F19.2F2020.29).

PR #5588, opened 2021-03-06
Adds optional database support to campaignd.

Pentarctagon:
Also uses that support to insert data into the new "addon_info" table when an add-on is uploaded as well as adds validation of the topic_id to ensure it's a valid int before inserting.

One use case for this would be to be able to check how many people are using outdated add-ons when playing online multiplayer on the primary official server.  Others would be to let players know when they create/join a game with an outdated version of one or more add-ons that will be used as well as adding a way to open an add-on's feedback thread in their browser from the current game they are playing.

Addresses #5060
Addresses #5061

---

Implementing #5065 would be dependent on whenever #5567 is merged, since doesn't make sense to implement client-side hashing only to soon remove it when TLS support is added.

### DIFF
--- a/src/server/campaignd/server.cpp
+++ b/src/server/campaignd/server.cpp
@@ -1372,6 +1372,21 @@ ADDON_CHECK_STATUS server::validate_addon(const server::request& req, config*& e
 		return ADDON_CHECK_STATUS::UNEXPECTED_DELTA;
 	}
 
+	if(const config& url_params = upload.child("feedback")) {
+		try {
+			int topic_id = std::stoi(url_params["topic_id"].str("0"));
+			if(user_handler_ && topic_id != 0) {
+				if(!user_handler_->db_topic_id_exists(topic_id)) {
+					LOG_CS << "Validation error: feedback topic ID does not exist in forum database\n";
+					return ADDON_CHECK_STATUS::FEEDBACK_TOPIC_ID_NOT_FOUND;
+				}
+			}
+		} catch(...) {
+			LOG_CS << "Validation error: feedback topic ID is not a valid number\n";
+			return ADDON_CHECK_STATUS::BAD_FEEDBACK_TOPIC_ID;
+		}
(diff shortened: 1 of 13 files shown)


Pentarctagon:
@shikadiqueen the changes you requested on Discord have been made.

src/server/campaignd/server.cpp
```diff
@@ -1426,8 +1452,15 @@ void server::handle_upload(const server::request& req)
...
 	addon.clear_children("feedback");
+	int topic_id = 0;
 	if(const config& url_params = upload.child("feedback")) {
 		addon.add_child("feedback", url_params);
+		// already validated that this can be converted to an int in validate_addon()
+		topic_id = url_params["topic_id"].to_int();
+	}
+
+	if(user_handler_) {
+		user_handler_->db_insert_addon_info(server_id_, name, addon["title"].str(), addon["type"].str(), addon["version"].str(), false, topic_id);
```
shikadiqueen, line 1463: Ehhh, I didn't really add any code to ensure `server_id_` isn't empty. In fact, the client expects it to potentially be empty/not present. What do you think should be done about this?

Pentarctagon: What's the use case for it being empty?

shikadiqueen: Not having to set any configuration when booting up a throwaway test instance.

Pentarctagon: Then I'd say the solution is to log an error and then exit on startup if database support is enabled but the ID is missing.

shikadiqueen: Then you know what to do. :p

Pentarctagon: Alright, updated.

shikadiqueen:
Seems good to me.

shikadiqueen (approved):

PR #5589, opened 2021-03-06
scorpion variations

### LINKED ISSUE #4076: Sand Scuttler and Giant Scorpion
I remember there being an old Giant Scorpion portrait that was removed not because of quality but rather as the old sprite didn't seem menacing enough to match the portrait. With the introduction of the Giant Scorpling though, I think the new Giant Scorpion portrait could be used for the Level 0 and the old portrait for the Level 1 Giant Scorpion.
Further, akin to the Dread Bat, which uses a recolour of the Blood Bat (hardcoded in the .cfg file), the Scorpion portraits could be tweaked to match the Sand Scuttler by making the hue more brown.

Thanks in advance 😃

doofus-01:
Some variations for scorpion.  I'll add to list as things progress. 

- derivative portrait for sand scuttler EDIT: this would close #4076
- zombie sprite frames and portrait
- new unit rock scorpion - base sprite, defense anim, and portrait


doofus-01:
I'm going to merge this.  There's more to do, but I need to get sync'd with what's current.

PR #5591, opened 2021-03-07
Fixed single player campaign number of players marker

### LINKED ISSUE #5361: Singleplayer Campaigns show number of players as 2
### Describe the bug
Singleplayer campaigns claim they are for two players when started in multiplayer.

### To Reproduce
1. Select "Multiplayer"
2. Select "Local Game"
3. Choose "Singleplayer Campaigns" in the scenario drop box

The number of players is shown as 2 for every singleplayer campaign.

### Expected behavior
It should show the number of players as 1.

krazyfencer:
The problem with the map generation showing 'Players: 2' on the local multiplayer generation of a single player campaign isn't a problem with the engine, but rather that all of the single player campaigns have not defined a min_players and max_players setting within their config.  Note that a campaign's min_players and max_players default to 2 (see src/game_initialization/create_engine.cpp).

Modified single-player campaign configurations to include min_players=1 and max_players=1.
This allows the correct number of players to be shown on the generation of a local
singleplayer campaign.

Please note that this is my first wesnoth update and one of my first PRs, so let me know if I clowned something up.

fixes #5361

### DIFF
--- a/data/campaigns/Dead_Water/_main.cfg
+++ b/data/campaigns/Dead_Water/_main.cfg
@@ -14,8 +14,6 @@
     abbrev= _ "DW"
     define=CAMPAIGN_DEAD_WATER
     first_scenario=01_Invasion
-    min_players=1
-    max_players=1
 
     {CAMPAIGN_DIFFICULTY EASY      "data/campaigns/Dead_Water/images/units/merfolk/citizen.png~RC(magenta>red)" ( _ "Citizen") ( _ "Beginner")}
     {CAMPAIGN_DIFFICULTY NORMAL    "units/merfolk/fighter.png~RC(magenta>red)" ( _ "Fighter") ( _ "Normal")} {DEFAULT_DIFFICULTY}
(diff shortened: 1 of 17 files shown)


Pentarctagon:
I think it would make more sense to update the default from 2 to 1 instead.  The default of two doesn't make much sense even for multiplayer campaigns since there can be more than two players for those as well, so it's really something that the campaign author needs to set themselves.

CelticMinstrel:
Maybe it makes sense for the min to default to 1 while the max defaults to 2? Or is it the max that causes "Players:2" to show at all?

krazyfencer:
In gui/dialogs/multiplayer/mp_create_game.cpp, for a campaign or sp campaign, if max_players > min_players, the label text changes from Players: to 'number of players^$min to $max'.   So if we make min=1, max=2 for the sp campaigns, they will all indicate that 1-2 players are allowed.
So I'm guessing campaigns should be defaulted to min=1, max=1?  And if a campaign supports multiplayer like 'Legend of Wesmere', they would just have to specify min_players, max_players in the config.

Pentarctagon:
Merged - thanks for fixing this!

PR #5594, opened 2021-03-07
modifications to pango_text for invalid markup

### LINKED ISSUE #5043: Crash when very long Pango text contains invalid markup
I introduced a new bug in #3265's 012069440e9976461ae2fc98937c080d7c44671b , when fixing the handling of very long texts. If the text is both long enough to trigger the `while (start_of_line != text_.cend())` loop and contains invalid markup then the call within the loop to `set_markup()` can call `set_text()`, thus changing the string that the loop is iterating over.

@Pentarctagon while this isn't on the roadmap, I assume it's a blocker and intend to fix it by 1.15.4 (probably) or 1.15.5 (definitely).

This was found while testing PR #5042 . That PR doesn't affect this bug, but the refactor suggested in it would be one way to fix it.

krazyfencer:
The problem here is that the set_markup method in src/font/text.cpp directly calls set_text() when invalid markup is encountered, however when handling line-by-line rendering, this causes two problems to happen.  First, it invalidates text_, which the processing is currently looping through.  Second, text_ is set to the line data, not the entire text.

I believe the solution is for set_markup not to modify the text via set_text when invalid markup is encountered.  It still reports the invalid text to debug, but it simply returns false indicating that setting the markup was unsuccessful.  The two places set_markup are called (set_text and rerender) have been modified to notice the failure of set_markup and use pango_layout_set_text in these cases to directly render the broken markup as text.

This will fix the loop invalidation, but I have changed the behaviour of wrapping the broken markup text in the string: '"The text contains invalid Pango markup: "'.  This can be added back in if you want, but it seemed odd to be doing this while silently dropping invalid utf-8 in other places in the code.  It seemed to make more sense to report the errors to the error streams and make the gui output as close to the requested value as possible.

fixes #5043

### DIFF
--- a/src/font/text.cpp
+++ b/src/font/text.cpp
@@ -294,11 +294,13 @@ bool pango_text::set_text(const std::string& text, const bool markedup)
 					<< " text '" << text
 					<< "' contains invalid utf-8, trimmed the invalid parts.\n";
 		}
-		if(markedup) {
-			if(!this->set_markup(narrow, *layout_)) {
-				return false;
-			}
+		if(markedup && !this->set_markup(narrow, *layout_)) {
+			// we were told to use mark up, but the text has invalid markup in it
+			markedup_text_ = false;
 		} else {
+			markedup_text_ = markedup;
+		}
+		if(!markedup_text_) {
 			/*
 			 * pango_layout_set_text after pango_layout_set_markup might
 			 * leave the layout in an undefined state regarding markup so
@@ -309,7 +311,6 @@ bool pango_text::set_text(const std::string& text, const bool markedup)
 		}
 		text_ = narrow;
 		length_ = wide.size();
-		markedup_text_ = markedup;
 		calculation_dirty_ = true;
 		surface_dirty_ = true;
 	}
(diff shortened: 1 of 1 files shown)


stevecotton:
Thanks for doing this, I'll review tomorrow. However, I'd prefer to just rip out the line-by-line rendering, assuming that end_credits.cpp is the only called that will want to display such a large amount of text.

krazyfencer:
No problem, just let me know if there's anything I can do to help.  If you plan on just pulling the line-by-line code from the original source, you may want to also look at the relationship between set_text and set_markup.  Basically if set_text() is called with invalid markup, it will call set_markup() to validate the markup text.  When it fails, set_markup calls set_text() to assign the text wrapped in an error message, with markup==false.  The thing is that this set_text() call will set the text_ member, and then as the stack unwinds, the original call to set_text() will return false (as the set_markup() call will fail).
So when somebody calls set_text(), it will return false, saying it didn't work, however the text contents of the object (and various flags) have actually been modified.  Unsure if that's a huge deal, just pointing it out.

stevecotton:
I think the parts outside of the line-by-line are good changes. There's a little of of extra whitespace in comments to clear up, but otherwise I like it. @Vultraz?

Something seems to be wrong with the line-by-line itself - when viewing the credits it seems to be repeating parts of the text. However, that's better than crashing, I'm not sure if it's an existing bug, and it's a part that I want to remove anyway so doesn't seem worth debugging.

Vultraz:
@shikadiqueen is also doing work on text rendering right now so let's get her feedback too

shikadiqueen:
I'm fairly sure the reasoning behind the "The text contains invalid Pango markup" lead at the time was making sure that add-on authors could clearly see that something was wrong with their input -- specifically, reducing the chances that they would miss it when skimming through text. For example, if I have two long paragraphs of text in a story screen and somewhere in between there are italicized words, there's a very good chance I'll miss that the markup is broken if the only indication of it is a `<i>foo</i>` and a message in stderr that the majority of content authors will miss (sadly).

Do you want me to look at the code in detail? The line-by-line rendering code is something I did see recently but deliberately decided to ignore because it sounds awful -- in particular since it's the Pango-based equivalent of the labyrinthine mess I just worked on excising off the codebase along with SDL_ttf.

Pentarctagon:
I think it'd be good to keep showing the error message in a place more visible than the logs.  If adding it to the text with the invalid markup is an issue, then maybe it could be displayed up in the chat area where other things (like lua errors) are displayed.

stevecotton:
@krazyfencer will you update this to re-add the "text contains invalid Pango markup", please?

Alternatively, I'd be happy to take over if you want keep your own time for something else.

stevecotton:
My current plan is to make this PR obsolete via #5602.

stevecotton:
Made obsolete by #5682.

PR #5595, opened 2021-03-07
Fix potential crash when running commands in synced context

mattsc:
Function run() at the beginning of this same file asserts when a command is configured not to be undoable (use_undo=false), but the undo stack is not empty (resources::undo_stack->can_undo() returns true).  Thus, the undo stack needs to be cleared before executing commands that are not supposed to be undoable.

Most commands do this, but it is not necessarily guaranteed for commands that are otherwise unsynched and get put into synced context here.  An example are the AI recall/recruit actions.  Thus, we need to clear the undo stack before trying to run such a command.


mattsc:
This is the reason for the crash happening in #5207.  The custom_synced_command there always has `use_undo=true`.  This is hard-coded by invoking synced commands through using `[do_command]` from ActionWML (which is something that we might also want to change, but it is separate from this issue).  If an AI recruit action is then executed afterward, the game crashes with the mentioned assert.

soliton-:
I'm not sure why the assert is there if just clearing the undo stack is the solution. I would have thought the solution is to clear the undo stack in some AI code. This looks like generic code that might run in AI or human turns.

soliton-:
Does clearing the undo stack trigger sending the moves to the server/remote sides in online games? Otherwise I expect there will be OOS if only the local side commits to the undoable moves and remote sides get the synced context stuff before the undoable moves. (And that is what I would guess is the reason for the assert.)

mattsc:
I had some more discussion with @soliton- on this and the bottom line is that I just don't understand this part of the code and functionality well enough to be certain that this won't cause other problems.  I would have to do a lot more digging into this and I don't have the time for that at the moment.  So I am not going to merge this for the time being.

As for #5207, I am going to separate out the part that causes the assert into a separate PR and merge the rest.

mattsc:
The part of #5207 that caused the assert has now been separated into its own PR #5609, and the rest of 5207 has been merged.

mattsc:
This is still open and I don't have the knowledge needed to decide how to proceed.  I see three options:

1. Use the fix I originally suggested above - but see the follow-up discussion.
2. Have AI recruit/recall actions clear the undo stack before they are being executed.  Problem with that is that they currently don't have access to the undo stack, and I don't know if that's like that for a reason or simply because so far there was no reason for this.
3. Allow [Lua function `invoke_synced_command()`](https://github.com/wesnoth/wesnoth/commit/23aac4f72fafc8c8a0991a614255b9532b3615a1#diff-07cbc854496f2facc551dd5184013a7c84ee0f5d6eaa528e8b581ca0947d6e1fR3971) to take an additional argument `use_undo` when being forwarded to [Action WML do_command](https://github.com/wesnoth/wesnoth/blob/master/src/game_events/action_wml.cpp#L337).

I have no idea which of those would cause any additional problems.  The last option probably has the least potential for unwanted side effects, but ...

@CelticMinstrel Pinging you since you are the last one, other than @gfgtdf, who has worked on that part of the code, as far as I can see.  In fact, the link to game_lua_kernel.cpp is not to the file, but to your commit in which you forward `invoke_synced_command` to `[do_command]` (although that was not a change of behavior, only of implementation).

mattsc:
Well, I have done what I can about this at this point.  So unless somebody has new input, this will just have to sit here together with the bug fix that triggered it.

soliton-:
https://github.com/wesnoth/wesnoth/blob/afded0db323d7625baebf406cb114d66bbb72ad7/src/synced_context.hpp#L54

I noticed this comment about use_undo.

I would still say that clearing the undo stack (or not filling it) in AI turns is the best idea but short of that it might be easier to check if this is an AI turn and use that info instead of use_undo.

mattsc:
@soliton-  I am finally getting back to this ...

> I would still say that clearing the undo stack (or not filling it) in AI turns is the best idea

I believe that you are right - and that the only actions for which this is necessary are recruiting and recalling as, for example, AI unit moves [pass nullptr to the move action](https://github.com/wesnoth/wesnoth/blob/master/src/ai/actions.cpp#L477) and thus take care of this already.  Also, recruiting/recalling [already set use_undo=false](https://github.com/wesnoth/wesnoth/blob/master/src/ai/actions.cpp#L659) (which is what causes the assert in the first place), so the only thing left to do is to clear the undo stack for those two actions.

The suggested fix is in 418da41 - I have tested that that fixes the assert encountered in #5207 and #5609.

soliton-:
Makes sense and looks good to me. (not that I looked any deeper into this)

Remember to squash when merging to get rid of the previous fix and revert commits.

mattsc:
Thanks for the cheer.  🙂 

I just squashed and force pushed.  I needed to change the commit message also, since this fix is slightly different from the one I imagined originally, and I was too lazy to do that last night.

I'll wait until the tests pass, and then merge this.

PR #5598, opened 2021-03-10
Make hotkeys work in any layout

### LINKED ISSUE #5421: Single key hotkeys don't work in a cyrillic layout
Hotkeys that include only one latin character don't work when a cyrillic layout is enabled. E.g. s (statistics), n (next unit), p (planning mode). The hotkeys that include ctrl or alt along with a latin character (e.g. ctrl+j, ctrl+u) work properly.

renom:
This makes wesnoth hotkeys be bound to scancodes instead of keycodes and inputted text.
Fixes #5421.


soliton-:
What happens in layouts where `:` is not produced by `shift+;`? This seems to move us back to not being able to tell someone how to use the game without knowing their keyboard layout and probably default hotkeys not working at all.

renom:
@soliton- that means you should press the `;` button along with the shift button. No matter whether there's a colon will be input or not.

renom:
@CelticMinstrel that's because you're on latin layout(s). This PR is dedicated to the setup that a player has to switch between a latin and non-latin ones.

soliton-:
@renom Why would we want that? We want the hotkey to be `:` no matter what you have to press on your keyboard to get it. Besides on my keyboard layout for example I have to press shift to get `;` so the hotkey makes no sense that way and won't work right.

renom:
> @renom Why would we want that?

See above. You don't care because #5421 doesn't affect you. For me it's much worse that I have to switch to the english layout each time when I need to use a hotkey, or create a cyrillic duplicate for an each single hotkey, than the fact I cannot create a symbol-based hotkey. Anyway, I think it could be resolved somehow. At worse it could be hardcoded (to have the colon hotkey the way you want).

renom:
@soliton- it's displayed as "shift+;" in the settings, not as ":". So technically you won't see ":" anywhere if you merge this pr. Are you really going to create a hotkey setup for every single language in the world?

renom:
Btw, you display "shift+latin_letter" as "shift+latin_letter". For example "shift+a" becomes "shift+a", not "A". What's the matter with "shift+;"? Also, "shift+cyrillic_letter" is possibly buggy, as it's displayed as "shift+uppercase_letter". See the screenshot:
wesnoth

src/hotkey/hotkey_item.cpp
```diff
@@ -288,19 +273,8 @@ bool hotkey_keyboard::matches_helper(const SDL_Event &event) const
 	unsigned int mods = sdl_get_mods();
 	const hotkey_command& command = get_hotkey_command(get_command());
```
kabachuha, line 274: Forgot to remove the command variable.

kabachuha, line 274: Forgot to remove the command variable.

src/hotkey/hotkey_item.cpp
```diff
@@ -483,11 +457,7 @@ bool is_hotkeyable_event(const SDL_Event &event) {
 
 	unsigned mods = sdl_get_mods();
```
kabachuha, line 458: Also this is no longer needed.

stevecotton:
What happens if two different keys on the keyboard can generate the same symbol? For Latin layouts this sometimes happens on the keys around the enter and shift keys, where some hardware makes enter or shift thinner and adds an extra key in the additional space.

What happens if the user switches between two layouts that have the same keys in different places, such as Dvorak / Qwerty / Qwertz / Azerty? Dvorak layouts also have a lot of variance in where the symbol keys are, someone could switch between programmer's variant and a variant with easy access to accented characters.

Are there other open-source games handling these keys in the way that you want? It would be useful to look at a working implementation for reference.

> Are you really going to create a hotkey setup for every single language in the world?

For the non-Latin languages that Wesnoth is translated into, it seems a reasonable thing to do. I think it would make sense, even if the hotkeys ended up on different physical keys depending on which charset was in use.

renom:
@stevecotton did you study the source code or just wrote this immediately?

stevecotton:
That was based on a quick skim of the code. I've now taken more time to test it, and found two problems where I expected to find one.

### Keys with SDL_SCANCODE_UNKNOWN

The conversion can't handle hotkeys bound to `<` or `>`. For both of these `SDL_GetScancodeFromName` returns `SDL_SCANCODE_UNKNOWN`, and so they don't work at all; based on https://wiki.libsdl.org/SDL_Scancode I think this is a feature of SDL and not a quirk caused by my keyboard layout.

### Two physical keys can generate the same symbol

My keyboard has the following keys, using Linux IBus' English (Dvorak) layout.
* scancode 25: `,` and `<`
* scancode 26: `.` and `>`
* scancode 94: `<` and `>`

I tried binding `shift+<`, `shift+>` and `<` to hotkeys. Using the current master version, both methods of entering `shift+>` work and trigger the same hotkey. With the convert-to-scancodes in this PR, even if `SDL_GetScancodeFromName` returned a scancode then one of the two methods of entering that combination would break.

renom:
@stevecotton I've added a workaround to handle unsupported scancodes by SDL_TextInputEvent. Could you test it please? (I'm unable to test, because my keyboard doesn't have any buttons that unsupported by SDL_KeyboardEvent). Anyway, I think it'd better be fixed in SDL. Also, I intentionally didn't make ":" and others (i.e. all the buttons except for latin-letter buttons) work this way (being handled by an input symbol instead of a button code, so that the game would remember ":" instead of "shift+;" for example). To make them work this way simply uncomment `CKey::is_uncomposable` calls at the 158 and 278 lines. It may be helpful for testing purposes.
PS I don't like "shift+;" being remembered as ":" simply because ":" could be on another button in another layout (in the russian layout on my keyboard it's on button "6" for example). The idea of this pr is to make a hotkey use the same button regardless which layout is enabled at the moment.

stevecotton:
Results:
* unshifted key 96 works, triggering the `<` hotkey
* shift + key 25 should trigger the `shift+<` hotkey, but instead triggers the `<` one
* neither method of generating `shift+>` works

renom:
@stevecotton `CKey::is_uncomposable` isn't used until you uncomment this.

renom:
Question to the wesnoth team - would it be merged or not? Now it works fine in my opinion. If you find a bug I'll try to fix it. I see you guys have different opinions on this (IIUC, what @stevecotton wrote is a matter of the design, so it's not a bug). Who will make the final decision?

stevecotton:
96 was a typo. By `shift+>` I meant either of `shift + key 26` or `shift + key 94`.

Pentarctagon:
So basically it's a question of whether hotkeys should be based on the keys' physical location (scancode, implemented in this PR) vs the character the key represents (keycode, current implementation)?

CelticMinstrel:
I think a system that handles both scancode-based keys and unicode-based keys would be needed to satisfy every use case; using unicode means that we can stick to the idea that : is the command key and typing : works no matter where it is on your keyboard, but using scancodes means the hotkey doesn't change if you switch layouts. On the other hand, making it an option might be too confusing… also, how are you supposed to display a scancode to the user? It'll be super confusing if the scancode is "Q" but that key actually produces (and is labelled as) "A".

renom:
@CelticMinstrel are there anywhere in the world keyboards made without an english layout? Anyway, even in the current implementation the latin letters are displayed in the hotkey settings unless you edit them.

renom:
Also, check this:
image
The hotkeys are displayed in latin regardless the language.

CelticMinstrel:
> are there anywhere in the world keyboards made without an english layout?

What exactly are you asking here? Certainly not all keyboards have their keys physically laid out in the QWERTY layout… I assume there are keyboards that don't even have Latin letters on them at all but I'm not sure about that one.

stevecotton:
The standard keybindings are broken on non-Qwerty layouts. For example, if I press `u` to undo a move then on my Dvorak I'm pressing the physical key that's `f` on a Qwerty; instead of undoing the move the command window pops up because `f` is the hotkey for `run formula`.

Also, saving custom hotkeys to the preferences file doesn't work for key 94. The preferences file gains a `[hotkey]` section, but it doesn't record which key it should be bound to:
```
[hotkey]
	alt=no
	cmd=no
	command="editor-brush-sw-ne"
	ctrl=no
	disabled=no
	shift=yes
[/hotkey]
```

renom:
SDL's keycode and layout events don't work as it's intended (at least on my pc).
1. The SDL_KEYMAPCHANGED event never occurs. This event should occur when the layout is switched.
2. Field `key.keysym.sym` (type SDL_Keycode) of SDL_Event, which is used when the event type is SDL_KeyboardEvent, sticks to the single layout, the one that's enabled when wesnoth is started, so that regardless whether the layout is changed or not afterwards, all the keycodes are obtained like the layout has never been changed (e.g. I start the game when the russian layout is enabled, then I switch to the english one, but when SDL_KeyboardEvent occurs, `key.keysym.sym` is still filled with russian letter keycodes).
I tested both on Wayland and X11, also tested outside wesnoth.

renom:
I was thinking of displaying hotkeys depending on the current layout, but probably it's impossible for now.

Pentarctagon:
Some of the CI jobs are failing due to a doxygen warnings, so that's an issue.  The main problem of course is that I was added as a reviewer and then the question I asked got ignored.

renom:
> SDL's keycode and layout events don't work as it's intended (at least on my pc).
> 
> 1. The `SDL_KEYMAPCHANGED` event never occurs. This event is supposed to occur when the layout is switched.
> 2. Field `key.keysym.sym` (type `SDL_Keycode`) of `SDL_Event`, which is used when the event type is `SDL_KeyboardEvent`, sticks to the single layout, the one that's enabled when wesnoth is started, so that regardless whether the layout is changed or not afterwards, all the keycodes are obtained like the layout has never been changed (e.g. I start the game when the russian layout is enabled, then I switch to the english one, but when `SDL_KeyboardEvent` occurs, `key.keysym.sym` is still filled with russian letter keycodes).
> 
> I tested both on Wayland and X11, also tested outside wesnoth.
> 
> Edit: the same way, `SDL_GetKeyFromScancode` is supposed to return a keycode depending on the current layout, but always uses keycodes of the starting layout.

Reported this to the SDL's GitHub:
https://github.com/libsdl-org/SDL/issues/4187

cooljeanius:
looks like some conflicts have arisen... @renom, could you look into resolving them?

Pentarctagon:
Closing due to age.

PR #5599, opened 2021-03-10
ui: Drop SDL_ttf and FriBidi dependencies in GUI1 and GUI2 code

shikadiqueen:
This patch series eliminates all dependencies on SDL_ttf and FriBidi in the Wesnoth codebase, including a forgotten dependency on SDL_ttf by GUI2 for obtaining font measurements, used by the textbox widget.

The bulk of the PR deals with reimplementing a few SDL_ttf wrapper APIs using Pango in `src/font/sdl_ttf_compat.cpp`. This includes some of the code used for the Help browser's layout strategy, which remains largely unchanged for the time being, providing identical functionality as before.

For the sake of simplicity, however, GUI1 markup is not implemented in this PR and it is instead dropped entirely, aside from the delimiters used to build GUI1 menus (such as the Help browser's menu). GUI1 markup became obsolete and was entirely phased out over the years leading up to 1.14 with the gradual adoption of GUI2 across the board since Wesnoth 1.6, and is not missed by anyone who wanted full control of how text is rendered due to its issues with variable line spans. Everyone agrees that GUI1 markup was a kluge and not a proper solution to the problem it was trying to solve. Once again, I must remind people that Help markup is not GUI1 markup. Help markup stays. GUI1 markup goes away.

As far as I know, there should be no functional differences compared to master. However, I would *greatly appreciate* double-checking from CJK translators either now or after the next development release as to whether the word-wrapping in the Help system is still sane. It *should* be, given how most of the work is done by Pango, which is a mature text layout API used by Gnome and others; but you never know when it comes to these things.

**Note:** Going forward, all the APIs in `sdl_ttf_compat.cpp` are meant to be removed. They are only supposed to exist for as long as their callers' architecture remains unchanged. For GUI1 widgets, that probably means forever. For the Help system, that means until a GUI2-based version is written, or the current one altered further to utilize Pango to its fullest potential.


stevecotton:
As Fribidi is involved, just a quick note that right-to-left support has been broken for some time; #5600 is an old bug and is not a regression added by this PR.

Pentarctagon:
It would probably make sense to send an email to the translators ML about testing things out.

shikadiqueen:
> > It would probably make sense to send an email to the translators ML about testing things out.

Translators generally aren't tech-savvy people, so emailing them before a release might not yield very useful results.

shikadiqueen:
For the record, I'm not going to update the Xcode project files myself because the game is going to be unbuildable without the addition of `sdl_ttf_compat.cpp` to it anyway.

Pentarctagon:
My general assumption for anything that needs it is that hrubymar can handle xcode project updates.

Pentarctagon:
I should be able to take a closer look at this on Sunday btw, though you and vultraz obviously understand the code in a more meaningful way.  It's kind of nice to be removing dependencies for once though.

stevecotton:
There's still an indirect dependency on FriBidi, as Pango itself depends on it. I still think this PR is a good change though.

PR #5601, opened 2021-03-11
Remove strong trait from Elvish Lady trait pool.

knyghtmare:
She is a cutscene unit but has no melee attack. Inheriting the _strong_ trait would not benefit her in anyway.

This does not impact anything in particular but looks rather odd when you view the in-game help. Thus, I have fixed it.
Screenshot 2021-03-11 044516

### DIFF
--- a/data/core/units/elves/Lady.cfg
+++ b/data/core/units/elves/Lady.cfg
@@ -8,6 +8,11 @@
     small_profile="portraits/elves/lady.png~CROP(0,20,380,380)"
     profile="portraits/elves/lady.png"
     hitpoints=47
+    ignore_race_traits=yes
+    {TRAIT_QUICK}
+    {TRAIT_INTELLIGENT}
+    {TRAIT_RESILIENT}
+    {TRAIT_DEXTROUS}
     movement_type=woodland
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
By your argument, the dexterous trait also gives her no benefit. Are you going to remove that too? That said, I'd rather not remove either and just give her some attacks and maybe a new sprite…

knyghtmare:
> By your argument, the dexterous trait also gives her no benefit. Are you going to remove that too? That said, I'd rather not remove either and just give her some attacks and maybe a new sprite…

She has a ranged attack and that benefits from `{TRAIT_DEXTROUS}`. This trait is also a racial trait for elves so I think it should stay.
Removing only the _strong_ trait from her trait pool seems better since she has no melee attack.

AI0867:
Related: Dark Adepts also cannot be strong, though their (living) advancements can.

CelticMinstrel:
Oh? Someone added an attack to her at some point? I either didn't know or had forgotten.

Dark Adepts probably can't be strong for MP reasons… that doesn't really apply to the Elvish Lady.

PR #5602, opened 2021-03-12
Rendering only the visible area in text_shape

### LINKED ISSUE #5043: Crash when very long Pango text contains invalid markup

stevecotton:
Merge #5681's canvas viewport support with #5682's pango_text viewport support. Both of those seemed large enough to split into their own PRs, this is the small change that needs both together; if the canvas PR merges first then I could roll this into the pango_text PR.

Fixes #5043.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -18,6 +18,7 @@
  ### WML Engine
    * Modify implementation of overwrite_specials attribute for replace yes/no parameter by none/one_side/both_sides and select abilities used like weapons and specials who must be overwrited(owned by fighter where special applied or both)
  ### Miscellaneous and Bug Fixes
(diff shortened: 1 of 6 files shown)


stevecotton:
I'm going to roll this into the other PR.

PR #5605, opened 2021-03-12
Add the server id to wesnothd as well.

Pentarctagon:
The version when wesnothd is compiled may not match the client versions it serves when switching to a new stable branch, so that value must come from the server config. This also means that the value must be updated manually to match the intended version.


soliton- (approved):

soliton-:
What should the id be set to for dev versions? 1.15 or 1.15.X?

Pentarctagon:
I'd say just 1.15 is fine, since that's the important part, and it also means not needing to update that for every point release.

PR #5607, opened 2021-03-13
Fix a crash where a WFL object could go out of scope and be destroyed before the formula was finished with it

CelticMinstrel:


mattsc:
The crash in the lurkers test scenario is gone and the FAI lurkers behave as they should, so I am going to approve this.

I can't say anything about the C++ part.  Some of the tests failed though.

mattsc (approved):

src/ai/default/contexts.hpp
```diff
@@ -47,6 +47,7 @@ struct target {
 
 class attack_analysis : public wfl::action_callable
 {
+	attack_analysis* clone() const override {return new attack_analysis(*this);}
```
stevecotton, line 50: Please use unique_ptr or shared_ptr instead of raw pointers. I had to do a similar conversion in b63b8a959a7b217228d44029466486e135cfdfb5 in order to start refactoring the editor code.

That would mean that the return type couldn't be co-variant, so all these clone() methods would have to return `unique_ptr<wfl::action_callable>` instead of `unique_ptr<attack_analysis>`, however does the caller ever need to know the exact subclass?

stevecotton (changes requested):
There's no documentation for the `safe_call_result` class, please would you add some to explain what it's meant to make safer?

src/ai/default/contexts.hpp
CelticMinstrel: Are you sure `unique_ptr` is correct though? This pointer is only going to be wrapped in a `shared_ptr`. You _are_ correct that there's no need for covariance here, but given that this is a _private_ clone function and the pointer is immediately wrapped in `shared_ptr`, it kinda feels wrong to use a `unique_ptr`.

CelticMinstrel:
I'm pretty sure `safe_call_result` has _something_ to do with the `safe_call()` function, but I'm not quite sure how that works, as `safe_call()` I think returns an action, not a `safe_call_result`. It might actually be more accurate to remove `safe_` from its name since it seems to somehow indicate the _result_ of an action that has been executed, but I'm not quite sure how that works.

src/ai/default/contexts.hpp
stevecotton: I did say "unique_ptr or shared_ptr"

CelticMinstrel: The result of this function _is_ wrapped in a `shared_ptr` already, and the function is private so it shouldn't be possible to call it anywhere else.

stevecotton: The smart pointers are tools to help with tracking ownership, using std::make_... instead of `new` makes the code more readable. Particularly in a PR that's fixing an object lifecycle bug, and where there's still a unit test failure to debug, you should be aiming for better code readability.

Returning shared_ptr and using std::make_shared is more efficient than allocating an object with `new` and then creating a shared_ptr to wrap it. You don't like the idea of returning a unique_ptr and then wrapping it, but using unique_ptr should be the same overhead as returning a raw pointer and then wrapping it - both of them are more expensive than using std::make_shared.

I recommend watching one of Herb Stutter's lectures about smart pointers.

src/formula/callable.hpp
```diff
@@ -172,11 +154,20 @@ class formula_callable
...
-	mutable std::set<callable_die_subscriber*> dtor_notify;
-
 private:
+	virtual formula_callable* clone() const = 0;
+	const_formula_callable_ptr to_ptr() const
+	{
+		if(weak_from_this().use_count() > 0) {
+			return shared_from_this();
+		}
+		return const_formula_callable_ptr(clone());
```
stevecotton, line 164: This looks like a source of obscure bugs.

If the caller has a pointer to this, why does it want to keep a pointer instead of just computing the result now? I think the only answer to that is "because something will set more variables later, which may change the value" - therefore silently making a copy here means that something will get the wrong value.

CelticMinstrel: There shouldn't be anything setting values on a callable in the middle of formula evaluation (other than temporary callables created as part of the evaluation), so I don't think it's _likely_ that this is a problem, but I'm not able to rule it out, unfortunately.

Ideally there would be no way to construct a formula_callable that's _not_ tracked by a shared_ptr, but that gets quite tricky…

CelticMinstrel:
So yeah, I have no idea why the tests fail, so I'm not sure if we should merge this.

Should we at least cherry-pick 714f411? This commit just changes the WFL code; it doesn't fix the underlying issue in the WFL engine, but it does make it not crash.

@mattsc What do you think?

stevecotton:
I'm confused by the commit message of 1bdcdb1 - it says it's "exactly the same", but does it also prevent a crash?

CelticMinstrel:
Basically there's a bug somewhere in the WFL or FAI engine which allows the contents of the mapped list to go out of scope and get destroyed after the mapping operation is complete. If you were to do `map(x, self)` on a list whose contents doesn't trigger this bug, you'll end up with the same list.

soliton-:
@CelticMinstrel any point in keeping the branch?

PR #5617, opened 2021-03-15
Simplify pango_text: surface is always dirty, remove unused functionality

stevecotton:
Every instance of pango_text had its own surface and dirty bit for caching the
result of rendering. However, pango_text is used as a singleton, so as soon as
there are two strings on screen then they're overwriting each others' cached result
and the dirty bit is always set. The loading screen is probably the only screen
where the dirty bit could be useful, as there's only one string on that screen.

The other parts removed were simply unused:
* the `force` options
* insert_unicode
* outlines

Co-ordination with the other currently open pango_text PRs:
* #5594 and this can be merged in either order, and won't have merge conflicts
* #5602 will have merge conflicts with both this and #5594, I'll keep rebasing it until the other PRs have merged

### DIFF
--- a/src/font/text.cpp
+++ b/src/font/text.cpp
@@ -53,7 +53,6 @@ pango_text::pango_text()
 	, font_size_(14)
 	, font_style_(STYLE_NORMAL)
 	, foreground_color_() // solid white
-	, add_outline_(false)
 	, maximum_width_(-1)
 	, characters_per_line_(0)
 	, maximum_height_(-1)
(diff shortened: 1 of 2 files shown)


shikadiqueen:
`pango_text` is **not** a singleton at the moment. The SDL_ttf compatibility API creates its own private instance and `gui2::styled_widget` among others create their own private instances for measurement purposes.

shikadiqueen:
GUI2 repeatedly calculates text extents and even renders text for figuring out widget dimensions. I want to double-check what's going on with that. Don't merge this without further input on that front.

stevecotton:
The text_box_base class does have its own instance of pango_text, and invalidates the basic assumption that I based this PR on - it can assume that no other user is altering the contents of the text.

The change in this PR is probably still OK, but it's not the obviously-correct change that I thought it was.

PR #5618, opened 2021-03-15
font: Work around chromatic aberration issue with Windows ClearType

shikadiqueen:
There's a long-standing issue with our usage of Cairo/ClearType on Windows that creates very noticeable chromatic aberration of rendered text (see the visual reference below) due to Wesnoth not respecting the industry recommendation of not using subpixel hinting when rendering text on transparent surfaces -- the latter being quite unfortunately something we always do because `pango_text` only generates transparent surfaces and most of the time has its output blitted onto more transparent surfaces.

This patch disables subpixel hinting, causing Cairo to render wispy text because of gamma issues when not rendering text directly onto an opaque surface. However, it also makes the implementation of pango_text overblend all transparent pixels to compensate like it was previously done before the revert in commit c91b200b7a10bf82fb13417a1267fcdf0fc4b2c1. This version uses alpha blending directly instead of forcing Pango to overdraw the entire text layout, however.

Overblending by 75% seems good enough for our purposes.

(Requesting @Vultraz's review since like all `pango_text`-related changes this is very much a UI related change and he's the current UI maintainer.)

## Visual reference

**Before:**

pango-text-default-1
pango-text-default-2

**After (175% alpha):**

pango-text-corrected-175-1
pango-text-corrected-175-2

**After (300% alpha):**

pango-text-corrected-1
pango-text-corrected-2

### DIFF
--- a/src/font/text.cpp
+++ b/src/font/text.cpp
@@ -81,7 +81,9 @@ pango_text::pango_text()
 	cairo_font_options_t *fo = cairo_font_options_create();
 	cairo_font_options_set_hint_style(fo, CAIRO_HINT_STYLE_FULL);
 	cairo_font_options_set_hint_metrics(fo, CAIRO_HINT_METRICS_ON);
-	cairo_font_options_set_antialias(fo, CAIRO_ANTIALIAS_DEFAULT);
+	// Always use grayscale AA, particularly on Windows where ClearType subpixel hinting
+	// will result in colour fringing otherwise. See from_cairo_format() further below.
+	cairo_font_options_set_antialias(fo, CAIRO_ANTIALIAS_GRAY);
 
 	pango_cairo_context_set_font_options(context_.get(), fo);
 	cairo_font_options_destroy(fo);
@@ -633,6 +635,17 @@ static void from_cairo_format(uint32_t & c)
 	unpremultiply(r, div);
 	unpremultiply(g, div);
 	unpremultiply(b, div);
+
+#ifdef _WIN32
+	// Grayscale AA with ClearType results in wispy unreadable text because of gamma issues
+	// that would normally be solved by rendering directly onto the destination surface without
+	// alpha blending. However, since the current game engine design would never allow us to do
+	// that, we work around that by increasing alpha at the expense of AA accuracy (which is
+	// not particularly noticeable if you don't know what you're looking for anyway).
+	if(a < 255) {
+		a = std::clamp<unsigned>(unsigned(a) * 1.75, 0, 255);
+	}
+#endif
 
 	c = (static_cast<uint32_t>(a) << 24) | (static_cast<uint32_t>(r) << 16) | (static_cast<uint32_t>(g) << 8) | static_cast<uint32_t>(b);
 }


src/font/text.cpp
```diff
@@ -618,6 +620,21 @@ static void unpremultiply(uint8_t & value, const unsigned div) {
...
+	if(a == 255) {
+		return;
+	}
+
+	unsigned tmp_r = 1 + r * 255;
+	r = (tmp_r + (tmp_r >> 8)) >> 8;
+	unsigned tmp_g = 1 + g * 255;
+	g = (tmp_g + (tmp_g >> 8)) >> 8;
+	unsigned tmp_b = 1 + b * 255;
+	b = (tmp_b + (tmp_b >> 8)) >> 8;
```
stevecotton, line 634: AFAICS the non-alpha channels always end up with the same values at the end as they have at the start.

Plotting it with Python:
```
>>> for r in range(0, 256):
...     tmp_r = 1 + r * 255
...     print (r - ((tmp_r + (tmp_r >> 8)) >> 8))
... 
0
lots more zeros
```

shikadiqueen: You're not wrong. The concept itself is dumb since obviously you can't mix (r, g, b) and (r, g, b) and expect something different in return. I'm sleepy.

shikadiqueen:
Updated and greatly simplified the patch after octalot pointed out some redundant maths I came up with while falling asleep.

However, I also made it so the alpha is ramped up to 175% instead of 300%, which seems to give out results more consistent with Linux and less BOLD, BRIGHT AND IN YOUR FACE.

stevecotton (approved):

shikadiqueen:
I forgot to point out a question that could arise from this, as well as its answer.

We already have `fonts/fonts.conf` for use with fontconfig on macOS and Linux, instructing the usage of greyscale AA with the rendering backend instead of subpixel AA. So the obvious question is why we don't use that on Windows as well -- in fact, I asked that question before myself. Loonycyborg and gfgtdf did some research into it and there is, in fact, [a working patch set that uses fontconfig](https://github.com/wesnoth/wesnoth/compare/use_fontconfig).

The problem is that the patch set limits our compatibility with older Windows versions greatly because of the fact that internally fontconfig uses ANSI versions of Windows and CRT APIs, which breaks when dealing with font paths containing Unicode characters. This is a border case for Wesnoth that has previously been known to render the game completely unusable before the last 1.12 release candidate, which saw the introduction of related fixes targeting SDL 1.2 thanks to SDL providing the necessary infrastructure to let Wesnoth use wide-character Unicode APIs instead. Fontconfig has no such thing, however, so the only working solution found involved using a new manifest option [available only since Windows 10 version 1903](https://docs.microsoft.com/en-us/windows/uwp/design/globalizing/use-utf8-code-page#set-a-process-code-page-to-utf-8).

Since it's not our intention to drop compatibility with older versions of Windows at this time, continuing not to use fontconfig and adjusting the alpha channel ourselves seems like the least impactful solution to the same issue. Since Wesnoth is not a graphics editor, I'm pretty sure we can afford some minor AA inaccuracy for the sake of not locking people out prematurely.

Wedge009:
I remember the long-standing issue of Windows builds rendering text that seemed 'green' or otherwise didn't look as 'nice' compared with other builds. This does seem to help a lot in that regard.

It sounds a bit annoying, but I agree with the approach regarding the fontconfig situation.

PR #5624, opened 2021-03-19
Add a prompt to allow entering an add-on's password at upload time.

Pentarctagon:
This is an alternative to needing to store the plaintext password in the _server.pbl.

### DIFF
--- a/src/addon/client.cpp
+++ b/src/addon/client.cpp
@@ -21,6 +21,7 @@
 #include "font/pango/escape.hpp"
 #include "formula/string_utils.hpp"
 #include "gettext.hpp"
+#include "gui/dialogs/addon/addon_auth.hpp"
 #include "gui/dialogs/addon/install_dependencies.hpp"
 #include "gui/dialogs/message.hpp"
 #include "gui/widgets/retval.hpp"
@@ -169,20 +170,6 @@ bool addons_client::upload_addon(const std::string& id, std::string& response_me
 		return false;
 	}
 
-	std::string passphrase = cfg["passphrase"];
-	// generate a random passphrase and write it to disk
-	// if the .pbl file doesn't provide one already
-	if(passphrase.empty()) {
-		passphrase.resize(16);
-		for(std::size_t n = 0; n < passphrase.size(); ++n) {
-			passphrase[n] = randomness::generator->get_random_int('a', 'z');
-		}
-		cfg["passphrase"] = passphrase;
-		set_addon_pbl_info(id, cfg);
-
-		LOG_ADDONS << "automatically generated an initial passphrase for " << id << '\n';
-	}
-
 	cfg["name"] = id;
 
 	config addon_data;
@@ -287,6 +274,16 @@ bool addons_client::delete_remote_addon(const std::string& id, std::string& resp
 
 	config request_buf, response_buf;
 	config& request_body = request_buf.add_child("delete");
+
+	// the passphrase isn't provided, prompt for it
+	if(cfg["passphrase"].empty()) {
+		if(!gui2::dialogs::addon_auth::execute(cfg)) {
+			config dummy;
+			config& error = dummy.add_child("error");
+			error["message"] = "Password not provided.";
+			return !update_last_error(dummy);
+		}
+	}
 
 	request_body["name"] = id;
 	request_body["passphrase"] = cfg["passphrase"];
--- a/src/gui/dialogs/addon/addon_auth.cpp
+++ b/src/gui/dialogs/addon/addon_auth.cpp
@@ -1,0 +1,48 @@
+/*
+	Part of the Battle for Wesnoth Project https://www.wesnoth.org/
+
+	This program is free software; you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation; either version 2 of the License, or
+	(at your option) any later version.
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY.
+
+	See the COPYING file for more details.
+*/
+
+#define GETTEXT_DOMAIN "wesnoth-lib"
+
+#include <functional>
+
+#include "gui/dialogs/addon/addon_auth.hpp"
+
+#include "gui/auxiliary/find_widget.hpp"
+#include "gui/widgets/password_box.hpp"
+#include "gui/widgets/window.hpp"
+
+namespace gui2::dialogs
+{
+
+REGISTER_DIALOG(addon_auth)
+
+addon_auth::addon_auth(config& cfg)
+	: cfg_(cfg)
+{
+	
+}
+
+void addon_auth::pre_show(window& win)
+{
+	win.add_to_tab_order(find_widget<password_box>(&win, "password", false, true));
+}
+
+void addon_auth::post_show(window& win)
+{
+	if(get_retval() == gui2::retval::OK)
+	{
+		cfg_["passphrase"] = find_widget<password_box>(&win, "password", false).get_real_value();
+	}
+}
+
+} // namespace dialogs
--- a/src/gui/dialogs/addon/addon_auth.hpp
+++ b/src/gui/dialogs/addon/addon_auth.hpp
@@ -1,0 +1,50 @@
+/*
+	Part of the Battle for Wesnoth Project https://www.wesnoth.org/
+
+	This program is free software; you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation; either version 2 of the License, or
+	(at your option) any later version.
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY.
+
+	See the COPYING file for more details.
+*/
+
+#pragma once
+
+#include "gui/dialogs/modal_dialog.hpp"
+
+namespace gui2::dialogs
+{
+
+/**
+ * @ingroup GUIWindowDefinitionWML
+ *
+ * This shows the dialog to provide a password when uploading an add-on.
+ * 
+ * Key               |Type           |Mandatory|Description
+ * ------------------|---------------|---------|-----------
+ * password          | text_box      |yes      |The password used to verify the uploader.
+ */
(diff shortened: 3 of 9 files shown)


src/tests/gui/test_gui2.cpp
```diff
@@ -569,6 +571,15 @@ BOOST_AUTO_TEST_CASE(test_make_test_fake)
 
 namespace {
 
+template<>
+struct dialog_tester<addon_auth>
+{
+	addon_auth* create()
```
stevecotton, line 577: Please return std::unique_ptr or std::shared_ptr instead of a raw owning pointer.

src/gui/dialogs/addon/addon_auth.hpp
```diff
@@ -0,0 +1,57 @@
...
+   std::string get_password();
+
+private:
+	/** Inherited from modal_dialog, implemented by REGISTER_DIALOG. */
+	virtual const std::string& window_id() const override;
+
+	/** Inherited from modal_dialog. */
+	virtual void pre_show(window& window) override;
+
+	/** Inherited from modal_dialog. */
```
stevecotton, line 48: Doxygen generates better docs without this comment - without it, Doxygen will copy the docs from modal_dialog.

src/gui/dialogs/addon/addon_auth.hpp
```diff
@@ -0,0 +1,57 @@
...
+	virtual void post_show(window& window) override;
+
+	field_text* password_;
+
+   std::string password_text_;
```
stevecotton, line 53: Inconsistent indentation on this line.

src/tests/gui/test_gui2.cpp
Pentarctagon: The tests are currently set up to want a `T*`.  Changing that is outside the scope of this PR.

src/gui/dialogs/addon/addon_auth.hpp
Pentarctagon: Fixed.

src/gui/dialogs/addon/addon_auth.hpp
Pentarctagon: This is a common problem with a lot of different dialogs then, and all should be fixed rather than just this one, which is also outside the scope of this PR.  The boost tests in your previous comment I don't really feel the need to address to be honest, but I'll push a separate commit to mass remove these comments after this PR is merged.

src/gui/dialogs/addon/addon_auth.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "gui/widgets/window.hpp"
+
+namespace gui2::dialogs
+{
+
+REGISTER_DIALOG(addon_auth)
+
+addon_auth::addon_auth()
+{
+	password_ = register_text("password", true);
```
Vultraz, line 29: What do you do with this?

Vultraz (changes requested):

src/gui/dialogs/addon/addon_auth.cpp
Pentarctagon: It's the password field used to enter the password to send when uploading an add-on, instead of taking it from the _server.pbl.

stevecotton:
There's also an automated passphrase generation in client.cpp - does this make that dead code? (I'd fully support that, I feel that it would be better to direct the user to documentation than to autogenerate a passphrase.)

https://github.com/wesnoth/wesnoth/blob/f1c37f64accbafe8dd52e638d50a8c70172f8abd/src/addon/client.cpp#L172-L184

Pentarctagon:
I suppose it would be, since the passphrase won't ever be blank one way or another after this PR.

src/gui/dialogs/addon/addon_auth.cpp
Vultraz: If you don't use the `password_` field you don't need to save a pointer in this class.

src/gui/dialogs/addon/addon_auth.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+void addon_auth::pre_show(window& win)
+{
+	win.add_to_tab_order(find_widget<text_box>(&win, "password", false, true));
+}
+
+void addon_auth::post_show(window& win)
+{
+   password_text_ = find_widget<password_box>(&win, "password", false).get_real_value();
```
Vultraz, line 39: Is it ok to keep the raw password value?

src/gui/dialogs/addon/addon_auth.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * Key               |Type           |Mandatory|Description
+ * ------------------|---------------|---------|-----------
+ * password          | text_box      |yes      |The password used to verify the uploader.
+ */
+class addon_auth : public modal_dialog
+{
+public:
+	addon_auth();
+
+   std::string get_password();
```
Vultraz, line 39: Bad indent.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -738,6 +739,21 @@ void addon_manager::publish_addon(const addon_info& addon)
 		}
 	}
 
+	// the passphrase isn't provided, prompt for it
+	if(cfg["passphrase"].str().empty()) {
```
Vultraz, line 743: Why not just `check `cfg["passphrase"].empty()`?

src/gui/dialogs/addon/manager.cpp
```diff
@@ -738,6 +739,21 @@ void addon_manager::publish_addon(const addon_info& addon)
...
+		gui2::dialogs::addon_auth auth;
+		auth.show();
+
+		switch(auth.get_retval()) {
+			case gui2::retval::OK:
+				cfg["passphrase"] = auth.get_password();
```
Vultraz, line 749: Instead of using a getter, I suggest taking `cfg["password"]` as a reference and setting that directly in post_show.

src/gui/dialogs/addon/addon_auth.cpp
Pentarctagon: Alright, will fix.

src/gui/dialogs/addon/addon_auth.cpp
Pentarctagon: Is there a reason it wouldn't be?

src/gui/dialogs/addon/addon_auth.hpp
Pentarctagon: Will fix.

src/gui/dialogs/addon/manager.cpp
Pentarctagon: No idea, will fix.

src/gui/dialogs/addon/manager.cpp
Pentarctagon: Updated.

src/addon/client.cpp
```diff
@@ -288,6 +275,11 @@ bool addons_client::delete_remote_addon(const std::string& id, std::string& resp
 	config request_buf, response_buf;
 	config& request_body = request_buf.add_child("delete");
 
+	// the passphrase isn't provided, prompt for it
+	if(cfg["passphrase"].empty()) {
+		gui2::dialogs::addon_auth::execute(cfg);
```
Vultraz, line 280: If you want to exit at this point you need to check whether `execute` returns false.

src/gui/dialogs/addon/addon_auth.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "gui/widgets/password_box.hpp"
+#include "gui/widgets/window.hpp"
+
+namespace gui2::dialogs
+{
+
+REGISTER_DIALOG(addon_auth)
+
+addon_auth::addon_auth(config& cfg)
+: cfg_(cfg)
```
Vultraz, line 30: Bad indent

src/gui/dialogs/addon/addon_auth.hpp
```diff
@@ -0,0 +1,55 @@
...
+	See the COPYING file for more details.
+*/
+
+#pragma once
+
+#include "gui/dialogs/modal_dialog.hpp"
+
+namespace gui2
+{
+class field_text;
```
Vultraz, line 20: Unnecessary forward definition, please remove and collapse the namespaces.

src/gui/dialogs/addon/addon_auth.cpp
```diff
@@ -0,0 +1,48 @@
...
+{
+	register_text("password", true);
```
Vultraz, line 32: I just noticed you don't actually set an initial value here. You shouldn't need this line at all.

src/gui/dialogs/addon/addon_auth.hpp
```diff
@@ -0,0 +1,55 @@
...
+	/** Inherited from modal_dialog, implemented by REGISTER_DIALOG. */
+	virtual const std::string& window_id() const override;
+
+	/** Inherited from modal_dialog. */
+	virtual void pre_show(window& window) override;
+
+	/** Inherited from modal_dialog. */
+	virtual void post_show(window& window) override;
+
+	config& cfg_;
```
Vultraz, line 51: I had been thinking of keeping a single `config_attribute_value` reference, but this works too.

src/gui/dialogs/addon/addon_auth.cpp
Pentarctagon: Fixed.

src/gui/dialogs/addon/addon_auth.hpp
Pentarctagon: Fixed.

src/gui/dialogs/addon/addon_auth.cpp
Pentarctagon: Fixed.

src/addon/client.cpp
Pentarctagon: Updated.

Vultraz (approved):

Issue #4145, opened 2019-07-01
Eastern Invasion


jostephd:
- [x] There's a bunch of ASCII quotes `'` that should be changed to typographic quotes `’`

- [x] S2: On the first playthrough a player might not look for the amulet, the objectives say `Get moving quickly.`

    #4420 unshrouds the village near the amulet. Is that enough?

- [x] ~S4c: The [custom tod schedule](https://github.com/wesnoth/wesnoth/blob/c94abfaad1c24ad29c5be39b5baad9808b7ced38/data/campaigns/Eastern_Invasion/scenarios/04c_Mal-Ravanals_Capital.cfg#L10-L18) should be mentioned in dialog or in a `[note]` in the objectives~ ab900b0093503809d91fc9c48ead19e388b63fe7

- [ ] S4c: In the new map there's a huge luck element about where the Paladin gets placed. The southwestern location is much easier to rescue than the southeastern location.

- [ ] S4c: the Death Knight has a 20-hex keep but [recruits up to 6 units per turn](https://github.com/wesnoth/wesnoth/blob/c94abfaad1c24ad29c5be39b5baad9808b7ced38/data/campaigns/Eastern_Invasion/scenarios/04c_Mal-Ravanals_Capital.cfg#L135) despite having money for more. That doesn't make any sense in-world.

- [x] ~~S7a: [hide_unit](https://github.com/wesnoth/wesnoth/blob/c94abfaad1c24ad29c5be39b5baad9808b7ced38/data/campaigns/Eastern_Invasion/scenarios/07a_The_Crossing.cfg#L139-L141) should be replaced with invisibility: https://github.com/wesnoth/wesnoth/issues/1357#issuecomment-299798847~~ - #4420

- [x] ~~S7a: side 2 should be hidden until it arrives.~~ - #4420

- [x] S7a: During side 1 turn 1, once side 2 is hidden, it'll be obvious in the status table that side 2 is missing. Maybe that's ok.

- [x] S7a: when Grug joins the party, there should be dialog or a narrator message informing the player that two more Ogres were added to the recall list as well - #4420 adds a narrator message

- [x] S7a: Grug still has `upkeep="full"` after it joins Gweddry (probably due to using `[modifications]` in `[modify_unit][modifications][effect]apply_to=loyal`, see #4137, and https://github.com/wesnoth/wesnoth/issues/4137#issuecomment-522270226)

- [x] Bone Knight has no attack icon for the "trample" attack

[...]

PR #5626, opened 2021-03-20
EI S7a S10 S11: Minor fixes

stevecotton:
S7a: Improve how side 2 makes its entrance. Don't use [hide_unit], that's not what it's for.

EI S10: Make recall undoable: this event shows a warning about recalled units having low / no movement through mountainous terrain.

EI S10: Switch shroud to fog. The scenario was otherwise generally unwinnable on the first attempt. Although the extra turns in added in ab900b0093503809d91fc9c48ead19e388b63fe7 help with that, it's still worth changing to fog instead of shroud.

EI S11: Annotate bonus objectives as such.

Part of #4145.

### DIFF
--- a/data/campaigns/Eastern_Invasion/scenarios/07a_The_Crossing.cfg
+++ b/data/campaigns/Eastern_Invasion/scenarios/07a_The_Crossing.cfg
@@ -136,9 +136,13 @@
             id=Owaec
         [/recall]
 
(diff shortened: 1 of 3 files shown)


nemaara (approved):

jostephd:
@stevecotton Thanks!

PR #5627, opened 2021-03-20
EI S07b: Wake the sea creatures with noise instead of magic

stevecotton:
This comes from a discussion about which archaic-sounding pronoun should be
used instead of "you" in "Creatures from the deep, I summon you to destroy this
foe". The conclusion with that was that "thee" or "ye" would sound better, but
"you" is the correct archaic pronoun.

The .ogg file is copied from Northern Rebirth.

Part of #4145.


data/campaigns/Eastern_Invasion/scenarios/07b_Undead_Crossing.cfg
```diff
@@ -131,44 +131,78 @@
 
         [message]
             speaker=Mal-un-Karad
-            message= _ "Creatures from the deep, I summon thee to destroy this foe!"
+            # po: the speaker is a necromancer, but he's about to summon some sea creatures
+            message= _ "Time to wake my living reinforcements!"
```
nemaara, line 135: Doesn't need "living", but otherwise looks fine.

stevecotton:
@nemaara do you want to check this again before I merge?

nemaara:
Looks fine.

PR #5628, opened 2021-03-20
Rename functions managing special weapons to improve readability for engine devellopper

newfrenchy83:
When I was looking at the special weapons extension to [abilities] tags, I didn't understand the concept of lists at first, until someone added the .append function to add a list to a list. other, but I kept the names I used for the functions. Now that this work is completed, I would like the function calls to find their original names (get_specials and get_special_bool) among other things to clarify the reading for a developer.

### DIFF
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1359,7 +1359,7 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 	return false;
 }
 
-bool attack_type::has_special_or_abilities(const std::string& special, bool special_id, bool special_tags) const
+bool attack_type::has_special_or_ability(const std::string& special, bool special_id, bool special_tags) const
 {
 	return (has_special(special, false, special_id, special_tags) || has_weapon_ability(special, special_id, special_tags));
 }
@@ -1456,7 +1456,7 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
 		return false;
 	}
 	if (tag_name == "firststrike" && !is_attacker && other_attack &&
-		other_attack->has_special_or_abilities("firststrike")) {
+		other_attack->has_special_or_ability("firststrike")) {
 		return false;
 	}
 
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -693,7 +693,7 @@ namespace {
  * active in the current context (see set_specials_context), including
  * specials obtained from the opponent's attack.
  */
-bool attack_type::get_special_bool_temp(const std::string& special, bool simple_check, bool special_id, bool special_tags) const
+bool attack_type::get_special_bool(const std::string& special, bool simple_check, bool special_id, bool special_tags) const
 {
 	{
 		std::vector<special_match> special_tag_matches;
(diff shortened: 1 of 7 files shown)


Pentarctagon:
Adding "_old" to a method name generally isn't a good way to improve readability unless there's some doxygen documentation to go with it that explains what that means and why it's the "old" way of doing something.

newfrenchy83:
> 
> 
> Adding "_old" to a method name generally isn't a good way to improve readability unless there's some doxygen documentation to go with it that explains what that means and why it's the "old" way of doing something.

i will replace "old" by "temp"

newfrenchy83:
@Pentarctagon i will use"without_abilities" instead

CelticMinstrel:
So "_without_abilities" means it doesn't check if someone is affected by a leadership-like ability that would grant that tag? Why does that even exist?

newfrenchy83:
> 
> 
> So "_without_abilities" means it doesn't check if someone is affected by a leadership-like ability that would grant that tag? Why does that even exist?

the function _without_abilities are the old function who check specials in [specials] renamed for what get_specials and get_special_bool point the globals specials function with check both abilities and specials tags and if the code don't writed in one function, this because case of function without abilities are called in the code (attack_type.cpp and special_id/type  without active, or the plague special in attack.cpp, or in ai where simple_check=true and not compatible with abilities)

CelticMinstrel:
> this because case of function without abilities are called in the code (attack_type.cpp and special_id/type without active, or the plague special in attack.cpp, or in ai where simple_check=true and not compatible wit

Why?

newfrenchy83:
i replace "without_abilities" by "_temp" and i midify code for these function don't called outside of get_specials and get_special_bool

Pentarctagon:
I mean, "_temp" isn't really any better than "_old".

newfrenchy83:
> 
> 
> I mean, "_temp" isn't really any better than "_old".

have you suggestion?

Pentarctagon:
I'd say that if the method checks just specials, then it should have "specials" in the name.  If it checks specials and abilities, then it should have "specials_and_abilities" in the name.

newfrenchy83:
> 
> 
> I'd say that if the method checks just specials, then it should have "specials" in the name. If it checks specials and abilities, then it should have "specials_and_abilities" in the name.

it's done.

CelticMinstrel:
I'd still like to know why there are places that check specials but not abilities…

Pentarctagon:
As another change, I'd prefer methods that are named like `get_..._bool` be changed to `has_...`.  So for example change `get_special_and_abilities_bool` to `has_special_or_abilities`.

newfrenchy83:
> 
> 
> As another change, I'd prefer methods that are named like `get_..._bool` be changed to `has_...`. So for example change `get_special_and_abilities_bool` to `has_special_or_abilities`.

only get_..._bool, not get... ?

Pentarctagon:
The methods that don't end in "_bool" return special/ability information rather than a boolean, correct?

newfrenchy83:
> 
> 
> The methods that don't end in "_bool" return special/ability information rather than a boolean, correct?

yes

Pentarctagon:
Right, so for methods that return a boolean, it's generally better form for the method name to be phrased somewhat like a yes or no question.  As an example, in `utils/math` there's the `is_even` and `is_odd` methods, rather than something like `get_even_bool` and `get_odd_bool`.

newfrenchy83:
@Pentarctagon done.

src/units/abilities.cpp
```diff
@@ -1359,9 +1359,9 @@ bool attack_type::get_special_ability_bool(const std::string& special, bool spec
 	return false;
 }
 
-bool attack_type::bool_ability(const std::string& special, bool special_id, bool special_tags) const
+bool attack_type::has_special_and_abilities(const std::string& special, bool special_id, bool special_tags) const
```
soliton-, line 1362: Surely this should be called `has_special_or_ability`. The implementation clearly uses or.

newfrenchy83: It's done

soliton-: If you want to go with plural it should be `has_specials_or_abilities`. To me it doesn't look like this is about multiple specials or abilities though. The input is a single string.

Issue #4137, opened 2019-06-28
modify_unit doesn't apply the "loyal" effect


jostephd:
`[modify_unit][modifications][effect]apply_to=loyal` doesn't set set the unit's upkeep to "loyal". Seen in EI S7a where Grug still has `upkeep="full"` after it joins Gweddry.

gfgtdf:
> `[modify_unit][modifications][effect]`

This is wrong syntax it's is supposed to be `[modify_unit][object][effect]`

jostephd:
I checked again and EI S7a does `[modify_unit][modifications][trait][effect]apply_to=loyal`. I omitted the [trait] in the first post. Sorry about that.

https://wiki.wesnoth.org/DirectActionsWML#.5Bmodify_unit.5D says "Accepts generally the syntax inside of wml unit variables created by [store_unit] which can be viewed in a savefile or by using the inspect command." The :inspect command in S9 shows `[modifications][object][effect]apply_to=loyal`. That's not exactly what you wrote but I assume it would work too.

zookeeper and I spoke about this yesterday. If I understood correctly he said the WML code in the scenario used to work. @ln-zookeeper, is that right?

ln-zookeeper:
I don't know if that code actually used to work (IIRC it was added in 1.13). But yes, the wiki does specifically mention that [trait] should be added **without** [modifications], so if that solves the problem then the current WML is in fact mistaken; the documentation doesn't really explicitly say what the current WML should do, after all.

gfgtdf:
> EI S7a does `[modify_unit][modifications][trait][effect]apply_to=loyal`

I'm not 100% sure what this does, it probably does a config-merge on the [modifications] tag ( meging the apply_to loyal with the first tarit that the uni has) but that's just a guess, also does to the config based implementation of modufy_unit it will probably not effect the units upkeep= field since Thai will be oweritten with the original upkeep= value when in unit in unstored.


In any case I recommend not to use ` [modify_unit][modifications]`

CelticMinstrel:
Yeah, that'll do a config merge on the modifications tag, but that does mean it should actually work, as the loyal trait is now in the modifications and `[modify_unit]` unstores the unit at the end, which if I recall correctly should trigger a rebuild.

That said, it's probably better to omit the `[modifications]` tag in this case.

(Actually, I'm not quite sure what the default behaviour is, but is there a chance that the existing code overwrites the first effect tag in the first trait of the unit?)

[...]

Issue #5632, opened 2021-03-22
[UtBS] S1: Quenoth Mystic still spawns on a deep water hex. 


knyghtmare:
### Game and System Information
 - What version of the game are you running?
 - 1.15.11+dev
 - Where did you download it from? (Steam, Mac App Store, our website, built from source, etc.)
 - Github Clone
 - If you're using a custom build, what compiler did you use, and what commit did you build at?
 - msys2 + latest commit
 - What OS are you running?
 - Windows 10

### Describe the bug
![Screenshot 2021-03-23 053222](https://user-images.githubusercontent.com/5283677/112071191-90c4b600-8b99-11eb-8d89-f8022bede1c7.png)
Screenshot showing the Quenoth Mystic spawning on a deep water hex. Also, I stated this bug before and was told it was fixed...and yet here it is...again.

### To Reproduce
1. load my autosave
[UtBS-The Morning After-Auto-Save13.gz](https://github.com/wesnoth/wesnoth/files/6185889/UtBS-The.Morning.After-Auto-Save13.gz)
2. Move the Tauroch rider unit to the island.

### Expected behavior
The Quenoth mystic at 23,27 should be somewhere else on the island, maybe 24,28?

stevecotton:
Simple solution that lead to further questions - add `passable=yes` to the code that spawns them, and set the `x` and `y` to the location of the tree itself, so that they'll spawn in a vacant hex around it, even if the player waded across the water from the north instead of using the bridge.

This lead on the question of whether `NAMED_GENERIC_UNIT` includes the `passable` flag. It doesn't, therefore I'm thinking of not using it for these units.

Next question:

```
    # Creates a generic unit of TYPE belonging to SIDE at X,Y, which has a
    # generated gender and traits (just like a recruited unit).
```

The uses of `NAMED_GENERIC_UNIT` UtBS S01 scenario seem unusual - that macro is documented as choosing a random gender, but in UtBS S01 it's being used to create units with names that look as if they should have specific genders. In other scenarios it's used with generic names such as "Criminal", "Westin Guard", "Merman Guard", "Goblin Skirmisher", etc.

However, although most of those are gender-neutral in English, they're not gender-neutral in all languages - TSG S02 has a workaround where, after using this macro, it then finds the female ones and sets their name to "female^Criminal" (see PR #5634); better would be to have a macro that takes both the male and female names.

[...]

PR #5634, opened 2021-03-23
TSG S05: Fix use of [modify_unit]

stevecotton:
TSG S05: Fix Urza's upkeep when taking the bandit branch. The `[trait]` shouldn't be inside a `[modifications]` block, as explained in #4137.

<strike>TSG S02: Simplify [store_unit]+[foreach] to [modify_unit]</strike>

### DIFF
--- a/data/campaigns/The_South_Guard/scenarios/05_Choice_In_The_Fog.cfg
+++ b/data/campaigns/The_South_Guard/scenarios/05_Choice_In_The_Fog.cfg
@@ -795,9 +795,7 @@
                                 moves=$this_unit.max_moves
                                 profile=portraits/urza-afalas.png
                                 canrecruit=no
-                                [modifications]
-                                    {TRAIT_LOYAL_HERO}
-                                [/modifications]
+                                {TRAIT_LOYAL_HERO}
                             [/modify_unit]
 
                             [modify_unit]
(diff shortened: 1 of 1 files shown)


data/campaigns/The_South_Guard/scenarios/02_Proven_by_the_Sword.cfg
```diff
@@ -154,26 +154,14 @@
...
-                {VARIABLE this_item.name _"female^Criminal"}
-                [unstore_unit]
-                    variable=this_item
-                    find_vacant=no
-                [/unstore_unit]
-            [/do]
-        [/foreach]
-        {CLEAR_VARIABLE namechange}
+            name= _"female^Criminal"
+        [/modify_unit]
```
stevecotton, line 164: Instead of doing this, let's deprecate the `NAMED_GENERIC_UNIT` macro and replace it with one that takes both a male and a female name. The macro generates units with a random gender, and with the exception of UtBS S01 (see issue #5632), it's being used with names that might be gendered in non-English languages, so it ought to support `female_name`.

It's only used in TSG and UtBS, with names such as "Criminal", "Westin Guard", "Merman Guard", "Goblin Skirmisher", etc.

stevecotton:
Decided to remove the S02 change from this PR, so it's now only the improvement to S05.

CelticMinstrel (approved):

Issue #4978, opened 2020-06-30
Setting a unit's facing does not redraw the unit to be facing the new direction


nemaara:
### Game and System Information
 - What version of the game are you running?
1.15.3

### Describe the bug
As the title says. The following block of code does nothing in 1.15.3, where as it works in 1.14.
`        [modify_unit]
            [filter]
                id=Yumi
            [/filter]
            facing=ne
        [/modify_unit]
`

Edit: turns out the unit's facing actually does change in WML (the gamestate inspector shows that this works) but the game doesn't draw it properly on the screen.

### Expected behavior
I expect my unit to turn when that block of code executes, but she doesn't. 

### Savefiles
The event at the end of the turn in the following save should make my unit (Yumi) turn several times during the cutscene, but she doesn't in 1.15, while she does in 1.14.

[Gns-1-Frozen Skies-Auto-Save3.gz](https://github.com/wesnoth/wesnoth/files/4850265/Gns-1-Frozen.Skies-Auto-Save3.gz)

gfgtdf:
You say that it doesn't work on 1.15.3, have you also tested master?

nemaara:
No because I can't test master (or maybe I can but I forgot how).

Pentarctagon:
You can download the latest executable from https://sourceforge.net/projects/bfw-travis-upload/files/

nemaara:
This bug appears to be in master as well.

gfgtdf:
Okay seems to be an Invalidate_hex call missing somewhere

gfgtdf:
> Okay seems to be an Invalidate_hex call missing somewhere

Actualyl i think this might not be sufficient, seems like  facing_ is cliseley tired to the animation code.

CelticMinstrel:
Does this result from a Lua API bug, or is it using store/unstore to change the facing of a unit?

It's true that facing is closely tied to the animation code, but as far as I recall, that's limited to setting it when an attack or movement animation plays. It _should_ work to arbitrarily change a unit's facing as long as they are simply standing on a single hex. Did you try the invalidate_hex call yet?

gfgtdf:
> Does this result from a Lua API bug, or is it using store/unstore to change the facing of a unit?

Its a lua api bug

> Did you try the invalidate_hex call yet?

Yes i did and it didn't work.

CelticMinstrel:
[...]

PR #5640, opened 2021-03-26
Use the slow path for [modify_unit]facing=

stevecotton:
Fixes bug #4978, that changing the facing wasn't updating the display.


CelticMinstrel:
Why? Shouldn't you be fixing this in `impl_unit_get` rather than forcing the slow path?

stevecotton:
I'm withdrawing myself from this, as I now see that I don't understand this part of the codebase. Although the code change here makes the graphics update, I don't understand why the fast path doesn't work.

The code in `impl_unit_set` already calls `unit::set_facing`.

PR #5641, opened 2021-03-26
modify overwrite_specials attributefor select the specials and weapon abilities who must be overwrited

newfrenchy83:
The overwrite_specials attribute is limites to abilities/specials owned by fighter or teacher, the specials/abilities who affect fighter but owned by opponent(apply_to=opponent/both) not affected.
the new attribute is created for fix that bugbut also select the specials afected, if is apply_to=self or opponent or both

### DIFF
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1300,18 +1300,18 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 	assert(display::get_singleton());
 	const unit_map& units = display::get_singleton()->get_units();
 	if(self_){
-		std::vector<special_match> special_tag_matches;
-		std::vector<special_match> special_id_matches;
-		get_ability_children(special_tag_matches, special_id_matches, (*self_).abilities(), special, special_id , special_tags);
+		std::vector<special_match> special_tag_matches_self;
+		std::vector<special_match> special_id_matches_self;
+		get_ability_children(special_tag_matches_self, special_id_matches_self, (*self_).abilities(), special, special_id , special_tags);
 		if(special_tags){
-			for(const special_match& entry : special_tag_matches) {
+			for(const special_match& entry : special_tag_matches_self) {
 				if(check_self_abilities(*entry.cfg, entry.tag_name)){
 					return true;
 				}
 			}
 		}
 		if(special_id){
-			for(const special_match& entry : special_id_matches) {
+			for(const special_match& entry : special_id_matches_self) {
 				if(check_self_abilities(*entry.cfg, entry.tag_name)){
 					return true;
 				}
@@ -1326,16 +1326,18 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 			if ( &*it == self_.get() )
 				continue;
 
-			get_ability_children(special_tag_matches, special_id_matches, it->abilities(), special, special_id , special_tags);
+			std::vector<special_match> special_tag_matches_adj;
+			std::vector<special_match> special_id_matches_adj;
+			get_ability_children(special_tag_matches_adj, special_id_matches_adj, it->abilities(), special, special_id , special_tags);
 			if(special_tags){
-				for(const special_match& entry : special_tag_matches) {
+				for(const special_match& entry : special_tag_matches_adj) {
 					if(check_adj_abilities(*entry.cfg, entry.tag_name, i , *it)){
 						return true;
 					}
 				}
 			}
 			if(special_id){
-				for(const special_match& entry : special_id_matches) {
+				for(const special_match& entry : special_id_matches_adj) {
 					if(check_adj_abilities(*entry.cfg, entry.tag_name, i , *it)){
 						return true;
 					}
@@ -1345,11 +1347,11 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 	}
 
 	if(other_){
-		std::vector<special_match> special_tag_matches;
-		std::vector<special_match> special_id_matches;
-		get_ability_children(special_tag_matches, special_id_matches, (*other_).abilities(), special, special_id , special_tags);
+		std::vector<special_match> special_tag_matches_other;
+		std::vector<special_match> special_id_matches_other;
+		get_ability_children(special_tag_matches_other, special_id_matches_other, (*other_).abilities(), special, special_id , special_tags);
 		if(special_tags){
-			for(const special_match& entry : special_tag_matches) {
+			for(const special_match& entry : special_tag_matches_other) {
 				if(check_self_abilities_impl(other_attack_, shared_from_this(), *entry.cfg, other_, other_loc_, AFFECT_OTHER, entry.tag_name)){
 					return true;
 				}
@@ -1357,7 +1359,7 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 		}
 
 		if(special_id){
-			for(const special_match& entry : special_id_matches) {
+			for(const special_match& entry : special_id_matches_other) {
 				if(check_self_abilities_impl(other_attack_, shared_from_this(), *entry.cfg, other_, other_loc_, AFFECT_OTHER, entry.tag_name)){
 					return true;
 				}
@@ -1372,9 +1374,11 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 			if ( &*it == other_.get() )
 				continue;
 
-			get_ability_children(special_tag_matches, special_id_matches, it->abilities(), special, special_id , special_tags);
+			std::vector<special_match> special_tag_matches_oadj;
+			std::vector<special_match> special_id_matches_oadj;
+			get_ability_children(special_tag_matches_oadj, special_id_matches_oadj, it->abilities(), special, special_id , special_tags);
 			if(special_tags){
-				for(const special_match& entry : special_tag_matches) {
+				for(const special_match& entry : special_tag_matches_oadj) {
 					if(check_adj_abilities_impl(other_attack_, shared_from_this(), *entry.cfg, other_, *it, i, other_loc_, AFFECT_OTHER, entry.tag_name)){
 						return true;
 					}
@@ -1382,7 +1386,7 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 			}
 
 			if(special_id){
-				for(const special_match& entry : special_id_matches) {
+				for(const special_match& entry : special_id_matches_oadj) {
 					if(check_adj_abilities_impl(other_attack_, shared_from_this(), *entry.cfg, other_, *it, i, other_loc_, AFFECT_OTHER, entry.tag_name)){
 						return true;
 					}
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1115,21 +1115,10 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
 	if(self_) {
 		abil_list.append((*self_).get_abilities(ability, self_loc_));
 		for(unit_ability_list::iterator i = abil_list.begin(); i != abil_list.end();) {
-			if(!((*i->ability_cfg)["overwrite_specials"].to_bool() && special_active(*i->ability_cfg, AFFECT_SELF, ability, true, "filter_student"))) {
+			if(!special_active(*i->ability_cfg, AFFECT_SELF, ability, true, "filter_student")) {
 				i = abil_list.erase(i);
 			} else {
 				++i;
-			}
-		}
-
-		if(abil_list.empty()){
-			abil_list.append((*self_).get_abilities(ability, self_loc_));
-			for(unit_ability_list::iterator i = abil_list.begin(); i != abil_list.end();) {
-				if(!special_active(*i->ability_cfg, AFFECT_SELF, ability, true, "filter_student")) {
-					i = abil_list.erase(i);
-				} else {
-					++i;
-				}
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


Pentarctagon:
Is there a reason to not just change the implementation of the existing attribute?

newfrenchy83:
> 
> 
> Is there a reason to not just change the implementation of the existing attribute?

i revrt to 'overwrite_specials' name and make some refactoring of code.

Pentarctagon:
What is `overwrite_apply`?

newfrenchy83:
> 
> 
> What is `overwrite_apply`?

In shema DEFAULT_KEY is used for define the parameter who can applied for overwrite_specials and onmy none/one_side/both_sides are valid. these key are defined in game_config.cfg to same matter who for apply_to or active_on, but i can rename 'overwrite_apply' to 'ability_overwrite' then.

Pentarctagon:
Why would anything need to be set in the game_config?  This isn't something that would need to be set globally is it?  There's also nothing actually looking for the value of `ability_overwrite` in this PR or in current master, so how would it even do anything?

newfrenchy83:
> 
> 
> Why would anything need to be set in the game_config? This isn't something that would need to be set globally is it? There's also nothing actually looking for the value of `ability_overwrite` in this PR or in current master, so how would it even do anything?

Define this directly in shema ability.cfg? I'm not certain

newfrenchy83:
> 
> 
> Why would anything need to be set in the game_config? This isn't something that would need to be set globally is it? There's also nothing actually looking for the value of `ability_overwrite` in this PR or in current master, so how would it even do anything?

This is not quite true, FORCE_CHANCE_TO_HIT uses overwrite_specials = both_sides, and this macro is used in some tests. It is true that otherwise the only abilities currently present do not use this attribute, but that can change with someone who would like to implement a magical cth type leadership which imposes a value of 70% even for attacks benefiting from a special cth with 80%, but that would be affected by a 'formation' type ability possessed by the opponent, then it could use overwrite_specials = one_side, but if this value is not set then checking fail.

Pentarctagon:
Can you remove ability_overwrite?

newfrenchy83:
> 
> 
> Can you remove ability_overwrite?

No, ithink no, how define the value what overwrite_specials can use?

Pentarctagon:
Oooh, alright, I misunderstood.  You're defining a brand new type in the WML schema that only accepts those three values.

Keeping it in `data/schema/game_config.cfg` is the correct choice then, I think.

Pentarctagon:
Why is plague handled specially in `attack_type::get_specials_and_abilities`?

newfrenchy83:
> 
> 
> Why is plague handled specially in `attack_type::get_specials_and_abilities`?

for what overwrite_specials could be used on [plague], if an [plague] abilitie use overwrite_specials specials [plague] wwill be overwrited, else it be back to specials used by default, if the default configuration remain in attack overwrite_specials don't usable for [plague]

src/units/abilities.cpp
```diff
@@ -1115,60 +1130,91 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
...
+		if(overwrite_self){
+			for(unit_ability_list::iterator i = temp_list.begin(); i != temp_list.end();) {
+				bool overwrite = false;
+				if(filter_self == "filter_student"){
+					overwrite = overwrite_special_affects(*i->ability_cfg);
+				}
+				if(special_active(*i->ability_cfg, AFFECT_SELF, ability, true, filter_self) && !overwrite) {
+					i = temp_list.erase(i);
+				} else {
+				++i;
```
Pentarctagon, line 1212: Missing indentation.

src/units/abilities.cpp
```diff
@@ -1115,60 +1130,91 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
...
+			if(filter_self == "filter_student"){
+				overwrite = overwrite_special_affects(*i->ability_cfg);
+			}
+			if((special_active(*i->ability_cfg, AFFECT_SELF, ability, true, filter_self) && !overwrite) || (special_active_impl(other_attack_, shared_from_this(), *i->ability_cfg, AFFECT_OTHER, ability, true, filter_self) && !overwrite)) {
+				i = temp_list.erase(i);
+			} else {
+				++i;
+			}
+		}
+	} else if(!overwrite_either){
```
Pentarctagon, line 1202: I think you can just check `overwrite_self` here?  `overwrite_either` is a boolean and you already checked the true case above, so checking that it's false at this point doesn't really do anything.

Actually this could just be an `else` rather than an `else if`.

src/units/abilities.cpp
```diff
@@ -1107,6 +1107,21 @@ namespace { // Helpers for attack_type::special_active()
 //beneficiary unit does not have a corresponding weapon
 //(defense against ranged weapons abilities for a unit that only has melee attacks)
 
+static bool overwrite_special_affects(const config& special)
+{
+	//log_scope("overwrite_special_affects_self");
```
Pentarctagon, line 1112: Is there a reason to keep this line?

src/units/abilities.cpp
```diff
@@ -1107,6 +1107,21 @@ namespace { // Helpers for attack_type::special_active()
...
+	const std::string& apply_to = special["overwrite_specials"];
+	if ( apply_to.empty() )
+		return false;
+	if ( apply_to == "none" )
+		return false;
+	if ( apply_to == "one_side" )
+		return true;
+	if ( apply_to == "both_sides" )
+		return true;
+	return false;
```
Pentarctagon, line 1122: This method can be simplified to `return apply_to == "one_side" || apply_to == "both_sides"`.

Pentarctagon:
So for `attack_type::overwrite_special_checking`, what that does is:
* Check whether there's an `overwrite_specials` attribute for "one_side" and "both_sides" using `abil_list`.
* If there's no `overwrite_specials` on any ability, return `temp_list` unchanged.
* If `overwrite_specials` is set to "both_sides", remove any ability special from `temp_list` that isn't active and either affecting the unit itself or the other unit, or doesn't have the `overwrite_specials` attribute set.
* If `overwrite_specials` is set to "one_side", remove any ability special from `temp_list` that isn't active and affecting the unit itself, or doesn't have the `overwrite_specials` attribute set.

Correct?

src/units/abilities.cpp
newfrenchy83: fixed

src/units/abilities.cpp
```diff
@@ -1115,60 +1130,91 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
...
-	return abil_list;
+	if(!overwrite_either && !overwrite_self){
+		return temp_list;
+	} else if(overwrite_either){
+		for(unit_ability_list::iterator i = temp_list.begin(); i != temp_list.end();) {
+			bool overwrite = false;
+			if(filter_self == "filter_student"){
+				overwrite = overwrite_special_affects(*i->ability_cfg);
+			}
+			if((special_active(*i->ability_cfg, AFFECT_SELF, ability, true, filter_self) && !overwrite) || (special_active_impl(other_attack_, shared_from_this(), *i->ability_cfg, AFFECT_OTHER, ability, true, filter_self) && !overwrite)) {
```
Pentarctagon, line 1191: This can be shortened to `!overwrite && (special_active(*i->ability_cfg, AFFECT_SELF, ability, true, filter_self) || special_active_impl(other_attack_, shared_from_this(), *i->ability_cfg, AFFECT_OTHER, ability, true, filter_self))`.

src/units/abilities.cpp
```diff
@@ -1115,60 +1130,91 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
...
+			}
+		}
+	} else if(!overwrite_either){
+		if(overwrite_self){
+			for(unit_ability_list::iterator i = temp_list.begin(); i != temp_list.end();) {
+				bool overwrite = false;
+				if(filter_self == "filter_student"){
+					overwrite = overwrite_special_affects(*i->ability_cfg);
+				}
+				if(special_active(*i->ability_cfg, AFFECT_SELF, ability, true, filter_self) && !overwrite) {
```
Pentarctagon: Please put `!overwrite` first - it's faster to check a boolean value than to call the `special_active(...)` method and if `overwrite` is true then the call to `special_active(...)` will be skipped.

This is also just a general good practice - when doing conditional checks, but the faster checks first.

newfrenchy83:
> 
> 
> So for `attack_type::overwrite_special_checking`, what that does is:
> 
>     * Check whether there's an `overwrite_specials` attribute for "one_side" and "both_sides" using `abil_list`.
> 
>     * If there's no `overwrite_specials` on any ability, return `temp_list` unchanged.
> 
>     * If `overwrite_specials` is set to "both_sides", remove any ability special from `temp_list` that isn't active and either affecting the unit itself or the other unit, or doesn't have the `overwrite_specials` attribute set.
> 
>     * If `overwrite_specials` is set to "one_side", remove any ability special from `temp_list` that isn't active and affecting the unit itself, or doesn't have the `overwrite_specials` attribute set.
> 
> 
> Correct?

Yes, but for both_sides , this either temp_list with apply_to=self and apply_to=opponent owned by enemy who affect self, for one_side, only apply_to=self owned by unit( or teacher). and specials never use overwrites_specials attributes, only abilities.

newfrenchy83:
@Pentarctagon and when temp_list is checked, all abilities or specials inactive already erased or not added before in get_specials or get_weapon_ability. special_active is used for diffrence apply_to=sel/opponent/attacker/defender, i use an already existent function, for that.

src/units/attack_type.hpp
```diff
@@ -125,6 +125,13 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 
 	// Configured as a bit field, in case that is useful.
 	enum AFFECTS { AFFECT_SELF=1, AFFECT_OTHER=2, AFFECT_EITHER=3 };
+	/** overwrite_special_checking : return an unit_ability_list list after checking presence or not of overwrite_specials
+	 * @param ability The special ability type who is being checked.
+	 * @param temp_list the list checked and returned.
+	 * @param abil_list list checked for verify presence of overwrite_specials .
+	 * @param filter_self name of [filter_"self/student"] if is abilities or specials who are checked
+	 */
+	unit_ability_list overwrite_special_checking(const std::string& ability, unit_ability_list temp_list, unit_ability_list abil_list, const std::string& filter_self = "filter_self") const;
```
Pentarctagon, line 134: Is there a reason to pass `abil_list` and `temp_list` by value instead of by reference and returning void?  `abil_list` could also be a const reference if I'm not mistaken.

newfrenchy83: temp_list is the onmy value to must be returned. for abil_list could be a const refrence, i must test that before.

newfrenchy83: i cannot compile with abil_list like const refrence

newfrenchy83: abil_list is const refrence now, but i will not use void.

Pentarctagon: You have `const unit_ability_list abil_list`, which should be `const unit_ability_list& abil_list` (missing `&`).

src/units/abilities.cpp
```diff
@@ -1115,62 +1115,100 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
...
 }
 
+static bool overwrite_special_affects(const config& special)
+{
+	const std::string& apply_to = special["overwrite_specials"];
+	if ( apply_to.empty() || apply_to == "none")
+		return false;
+	if ( apply_to == "one_side" || apply_to == "both_sides")
+		return true;
+	return false;
```
Pentarctagon, line 1165: I'd very much prefer this be just `return apply_to == "one_side" || apply_to == "both_sides"`.

src/units/abilities.cpp
```diff
@@ -1115,62 +1115,99 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
...
-	abil_list.append(get_specials(special));
 	return abil_list;
 }
 
+static bool overwrite_special_affects(const config& special)
+{
+	const std::string& apply_to = special["overwrite_specials"];
+	if ( apply_to.empty() || apply_to == "none"){
+		return false;
+	}
```
Pentarctagon, line 1163: Please delete the above three lines.

newfrenchy83: Done

src/units/abilities.cpp
```diff
@@ -1115,62 +1115,99 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
...
+	return (apply_to == "one_side" || apply_to == "both_sides");
+}
+
+unit_ability_list attack_type::overwrite_special_checking(const std::string& ability, unit_ability_list temp_list, const unit_ability_list& abil_list, const std::string& filter_self) const
+{
+	bool overwrite_self = false;
+	bool overwrite_either = false;
+
+	for(auto& i : abil_list) {
```
Pentarctagon, line 1169: This should be `const auto& i`.

newfrenchy83: This is added

Pentarctagon:
I think I'm good with this now, but I'll leave it open a couple more days to see if anyone else comments.

newfrenchy83:
> 
> 
> I think I'm good with this now, but I'll leave it open a couple more days to see if anyone else comments.

Excuse me but i pull a bug fix in this PR.

Pentarctagon:
For the fix, what is the current incorrect behavior, and what is the new behavior with the fix?

newfrenchy83:
> 
> 
> For the fix, what is the current incorrect behavior, and what is the new behavior with the fix?

This bug is rare enough that I may have missed it so far;
when in Httt, The Princess is adjacent to an enemy unit with [leadership], and their students clash with the Princess's student in defense, the 'initiative' ability does not work.
This is because std :: vector <special_match> special_tag / id_matches is not reset for each adjacent unit, which leads to an accumulation of data which interferes and disrupts the functioning of the bool type abilities.

With the repair, 'initiative' works fine even in this case.

src/units/abilities.cpp
```diff
@@ -1326,6 +1326,8 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 			if ( &*it == self_.get() )
 				continue;
 
+			std::vector<special_match> special_tag_matches;
+			std::vector<special_match> special_id_matches;
```
Pentarctagon, line 1330: I would prefer that these are given different names rather than shadowing an existing name.  That way things like this would be a compiler error instead.

So for example call these `special_tag_matches_adj` and `special_id_matches_adj`, then call the variables on lines 1303-1304 `special_tag_matches_self` and `special_id_matches_self`.

Pentarctagon:
Can you squash the previous two commits with the initial bugfix commit?

newfrenchy83:
> 
> 
> Can you squash the previous two commits with the initial bugfix commit?

it's done

PR #5644, opened 2021-03-28
pango_text: remove unused force options and single-character insertion

stevecotton:
The single-character insertion seems unlikely to be used in future. While
pango_text isn't a singleton, a few shared instances are created and then used
in a singleton-like way, with each caller assuming that it received it in an
unknown state and always calling set_text(), set_font_size(), set_font_style(),
etc.

This is a subset of the changes in #5617.

### DIFF
--- a/src/font/text.cpp
+++ b/src/font/text.cpp
@@ -140,17 +140,6 @@ unsigned pango_text::insert_text(const unsigned offset, const std::string& text)
 	return len;
 }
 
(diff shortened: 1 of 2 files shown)


Vultraz (approved):
No objections from me. @shikadiqueen ?

shikadiqueen:
Seems okay to me.

shikadiqueen (approved):

PR #5646, opened 2021-03-29
[1.15] Make WFL's error-reporting machinery thread_local

### LINKED ISSUE #5629: Random wesnoth crashes during loading screen
 - 1.15.11
 - Steam
 - Windows 10

### Describe the bug
Wesnoth is randomly crashing during loading of game / mp (trying to enter local or mp server) / campaigns (trying to enter campaign selection screen). Game is not stopped or without answer, it's process is ended completely and instantly.

### To Reproduce
I have no idea how to reproduce it. It just happens from time to time usually during first launch of wesnoth. After certain section loads properly it doesn't crash again unless I reset my PC.

stevecotton:
To report the WFL call stack in the event of errors, a static stack is used to
record that call-stack. This wasn't thread-safe, and it intermittently crashed
when the game config was movetype patching (the other WFL thread involved is
the GUI2 layout code).

Also a Rule of Three fix for the call_stack_manager class. While this wasn't
the cause of the crash, it could also have lead to a call_stack.pop_back() on
an empty stack.

Fixes #5629.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -17,6 +17,7 @@
  ### WML Engine
  ### Miscellaneous and Bug Fixes
    * Fixed units with max movement set to zero being given one max movement point by `[unstore_unit]` or when loading a saved game
+   * Fixed an intermittent crash on the loading screen (issue #5629)
 
 ## Version 1.15.11
  ### AI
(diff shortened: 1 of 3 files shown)


CelticMinstrel (approved):
Seems fine to me.

stevecotton:
@Vultraz, please would you review this soon? It's only an intermittent crash, but could also cause memory corruption leading to heisenbugs.

PR #5655, opened 2021-04-02
Fix a typo

hexagonrecursion:
singular: brother; plural: brethren; plural possessive: brethren’s
granted: it is possible that this word is an exception. I'm not an expert


CelticMinstrel:
It's correct to use 's as the plural possessive when the plural does not itself end in S. I don't think there's any exceptions to this.

Issue #5651, opened 2021-03-31
Dark Forecast Has Become Impossible on Default (1.15.11)


anakayub:
### Game and System Information
 - What version of the game are you running? 1.15.11
 - Where did you download it from? (Steam, Mac App Store, our website, built from source, etc.) Website
 - If you're using a custom build, what compiler did you use, and what commit did you build at? Default compile
 - What OS are you running? MacOS Big Sur

### Describe the desired feature
In the current version, it seems that Dark Forecast has become impossible to beat in its default settings. The spawns are now too frequent; usually the next spawn occurs before finishing off the current one. Spawns are also too powerful (as they appear before any proper recovery/healing); their value usually approaches/equals the value of the players' units. To add insult to injury, they appear usually right before their optimal time of day, making their counterattacks devastating. I suggest some form of adjustment somewhere to make beating this mainline scenario...possible. I've tried over the past 3 weeks and this was the only time I reached turn 30 with this exact faction combo.

### Screenshots
If applicable, add screenshots or mockups to help explain your request.

<img width="1792" alt="Screenshot 2021-03-31 at 10 57 31 PM" src="https://user-images.githubusercontent.com/42091252/113183729-ecbcc800-9286-11eb-8d5d-1fc3404dc846.png">
The value of the spawn here probably even exceeds the units I have left

[https://replays.wesnoth.org/1.15/2021/03/31/2p__Dark_Forecast_(Survival)_Turn_30_(35).bz2](url)
My last attempt, played on the MP server.

knyghtmare:
I didn't win but I didn't died either. It's rather...unwinnable.
[2p — Dark Forecast (Survival)-Auto-Save59.gz](https://github.com/wesnoth/wesnoth/files/6238747/2p.Dark.Forecast.Survival.-Auto-Save59.gz)
[2p — Dark Forecast (Survival)-Auto-Save58.gz](https://github.com/wesnoth/wesnoth/files/6238748/2p.Dark.Forecast.Survival.-Auto-Save58.gz)
![Screenshot 2021-04-01 002808](https://user-images.githubusercontent.com/5283677/113194199-acf3e180-9282-11eb-891c-dd6186467007.png)
![Screenshot 2021-04-01 002837](https://user-images.githubusercontent.com/5283677/113194205-afeed200-9282-11eb-8c96-88aca7a9c5f1.png)
[...]

PR #5656, opened 2021-04-02
Fix for weather change cheat in dark forecast

### LINKED ISSUE #5653: [Bug] [MP] Dark Forecast exploit in code allows player to prevent (cheat) weather change effects
 - Wesnoth 1.15.11+dev
 - Github
 - v1.15.11+dev (289ac354a07-Modified)
 - What OS are you running?
 - Windows 10

### Describe the bug
In line 453 of https://github.com/wesnoth/wesnoth/blob/289ac354a07b2d5ac05481e5e949eae5505190cd/data/multiplayer/scenarios/2p_Dark_Forecast.lua#L453
The weather will only change for a `side 3 turn` event. Thus, if side 3 is left empty, and player enters 1p-mode as side 4, he/she can prevent the weather effects from taking place.

### To Reproduce

1. Enter MP Create Game and select `2p - Dark Forecast`
2. Set side 3 as Empty (`Nobody`)
3. Play as Side 4.
4. Weather change will never happen.
5. Enjoy the exploit.

### Expected behavior
Weather change every 3 turns.

### Savefiles
Replay as `Rebels` of `Default Era`
[2p — Dark Forecast (Survival) replay 20210402-023207.gz](https://github.com/wesnoth/wesnoth/files/6246382/2p.Dark.Forecast.Survival.replay.20210402-023207.gz)

### Additional context
Discovered after probing the code for a fix on #5651

knyghtmare:
Adjusted the weather change effect from `side 3 turn` to `side 1 turn`. Side 1 is never empty so it would bypass the side 3 as empty cheat which impeded weather change.

Fixes #5653 

Also, I have playtested under the _cheat active_ conditions and found the weather change to working now.

### DIFF
--- a/data/multiplayer/scenarios/2p_Dark_Forecast.lua
+++ b/data/multiplayer/scenarios/2p_Dark_Forecast.lua
@@ -449,7 +449,13 @@ local function weather_map(name)
 	wesnoth.redraw {}
 end
 
--- change weather at side 3 turns, TODO: consider the case that side 3 is empty.
+-- change weather at side 1 turns
+-- initially this was set for every side 3 turns
+-- which allowed a cheat to be used by players 
+-- by setting side 3 as Empty and picking a faction 
+-- optimised for the default/basic map for side 4
+-- which caused this weather change featur to never
+-- trigger
 on_event("side 1 turn", function()
 	-- get next weather event
 	local weather_event = wml.variables["weather_event[0]"]
--- a/data/multiplayer/scenarios/2p_Dark_Forecast.lua
+++ b/data/multiplayer/scenarios/2p_Dark_Forecast.lua
@@ -454,8 +454,7 @@ end
 -- which allowed a cheat to be used by players 
 -- by setting side 3 as Empty and picking a faction 
 -- optimised for the default/basic map for side 4
--- which caused this weather change featur to never
--- trigger
+-- which caused this weather change feature to never trigger
 on_event("side 1 turn", function()
 	-- get next weather event
 	local weather_event = wml.variables["weather_event[0]"]
--- a/data/multiplayer/scenarios/2p_Dark_Forecast.lua
+++ b/data/multiplayer/scenarios/2p_Dark_Forecast.lua
@@ -450,7 +450,7 @@ local function weather_map(name)
 end
 
 -- change weather at side 3 turns, TODO: consider the case that side 3 is empty.
-on_event("side 3 turn", function()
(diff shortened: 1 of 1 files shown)


Pentarctagon:
The comment line immediately above that should be updated to reflect this change.

knyghtmare:
> The comment line immediately above that should be updated to reflect this change.

Done.

knyghtmare:
> Just as a stylistic thing, can you break comment lines on natural pauses (ie: sentences) rather than keeping it to ~8 words max per comment line? It's easier to read without the artificial breaks in the word flow like that, and even on tiny screens there's line wrapping built in to pretty much every editor these days.

Updated the comment to 4 sentences. One explains what the code does. The rest clarifies why such a change was made.

CelticMinstrel:
Just wondering, but with this change, couldn't you still cheat by setting side 1 to empty?

knyghtmare:
> Just wondering, but with this change, couldn't you still cheat by setting side 1 to empty?

Side 1 is the aggressive side which controls all enemy units, and has `allow_player=no` set. Players can never set side 1 as empty.

CelticMinstrel:
Ahhh, okay.

soliton-:
So now the weather changes in the enemy turn and the player has potentially no way to put their units on good terrain?

knyghtmare:
> So now the weather changes in the enemy turn and the player has potentially no way to put their units on good terrain?
Enemy sides are 1 and 2. So, the player does get chance to place units on their turn. I did play with the new patched Dark Forecast a few times. The terrain change was no issue, but the low income and village gold were more prominent loosing factors.

soliton-:
Weather used to change at the start of the player's turn. Thus the player could place their units according to the new terrain. Now weather changes after the player placed their units at the start of the enemy's turn. Am I missing something?

knyghtmare:
> Weather used to change at the start of the player's turn. Thus the player could place their units according to the new terrain. Now weather changes after the player placed their units at the start of the enemy's turn. Am I missing something?

Good point. Perhaps another solution would be to enable `side_name=_"player"` for sides 3 and 4 and then name the event as `player_turn` (okay, this might cause weather change to happen twice in one turn...now that I think about it). Currently setting the terrain change on side 1 turns, sort of increases difficulty in some cases, and should be addressed #5651

AI0867:
Is side 2 always occupied? If so, one might use a `side 2 turn` that then sets a `side turn` event, which should then trigger on the first side after side 2.

PR #5658, opened 2021-04-02
Dark Forecast Lua file comment Typo fix

knyghtmare:
"cauculate" does not seem to be any existing word and auto-corrects to "calculate" which seems more appropriate here.


Pentarctagon:
Could you fix create_timed_spaws -> create_timed_spawns while you're at it?

knyghtmare:
> Could you fix create_timed_spaws -> create_timed_spawns while you're at it?

Done.

PR #5662, opened 2021-04-04
remove reference to tusker, replace with boar or piglet

doofus-01:
Mainline will no longer step on UMC use of "Tuskers".  Still using the tusker/gorer sprites for now.


CelticMinstrel:
Well, I don't really see the point  of this, but I don't really have a problem with it either…

PR #5663, opened 2021-04-04
Refactor the game event system to be based on Lua at the core

CelticMinstrel:
All events are now just a Lua function and a package of arguments that are passed to the function. A typical event added via WML (ie, with the `[event]` tag) is the function `wesnoth.wml_actions.command` with the contained ActionWML tags passed as the argument, so they work just the same as before. However, you can now add events in two other ways:

* ~Instead of using ActionWML, you can write the entire event inline as Lua code using `[event]code=`. Although this is similar to just using an event containing a single `[lua]` tag, there are some differences. The code is compiled when the event is added, rather than when it is executed, and the argument passed to the event code is a copy of the `[event]` tag with filter data and event name and ID stripped out.~
* You can now register an event directly from Lua using the new `wesnoth.game_events.add` function, which supercedes the `on_event()` system that World Conquest and a few other things depend on. Unlike `on_event()`, this interface supports everything that `[event]` does, such as `first_time_only=true` or giving the event an ID that will allow other events to remove it. It also supports some things that `[event]` doesn't, such as using an arbitrary Lua function to handle the event (although if you do this, the event becomes unserializable, just the same as if you used `on_event()`) or marking the event as a menu item event (meaning it will be removed if the menu item with the same ID is removed).

The other significant change in this PR is a refactor of the event filtering system, which would make it very easy to add a new type of filter if we wanted to in the future (although I'm not anticipating any new types). As part of that, two new ways of filtering in events have been added:

* The `filter_formula` key in `[event]` takes a WFL formula that determines whether the event should trigger. It can access all the event context info (like `loc1`, `event_id`, `unit1`, etc) as well as the game state (such as `map`, `units`, `turn`, and `time_of_day`). In particular, it exposes event context info that nothing else does in `event_data`, although there may not be any way to actually set that data yet in `[fire_event]`.
* ~The `filter` key in `[event]` takes Lua code that determines whether the event should trigger. If there is a `[filter_args]` tag, its contents is passed to this function.~
* In addition, when adding an event from Lua, you can pass an arbitrary Lua function as the filter. Doing this makes the event unserializable, of course.

One other thing I've done here is to divorce menu items from events at the Lua level. This was already the case in C++, and it seemed to make for a cleaner API. The relevant functions were never documented, so this is probably a minor thing, but basically: `wesnoth.interface.set_menu_item` adds or updates a menu item but does not make it do anything - the `[command]` tag is ignored. To make it do something, you need to use `wesnoth.game_events.add` to create the event that the menu item will trigger. Of course, `[set_menu_item]` still does both for you.

### DIFF
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -146,10 +146,12 @@ function wml_actions.fire_event(cfg)
 
 	local w1 = wml.get_child(cfg, "primary_attack")
 	local w2 = wml.get_child(cfg, "secondary_attack")
-	if w2 then w1 = w1 or {} end
+	local data = wml.get_child(cfg, "data") or {}
+	if w1 then table.insert(data, wml.tag.first(w1)) end
+	if w2 then table.insert(data, wml.tag.second(w2)) end
 
-	if cfg.id and cfg.id ~= "" then wesnoth.fire_event_by_id(cfg.id, x1, y1, x2, y2, w1, w2)
-	elseif cfg.name and cfg.name ~= "" then wesnoth.fire_event(cfg.name, x1, y1, x2, y2, w1, w2)
+	if cfg.id and cfg.id ~= "" then wesnoth.game_events.fire_by_id(cfg.id, x1, y1, x2, y2, data)
+	elseif cfg.name and cfg.name ~= "" then wesnoth.game_events.fire(cfg.name, x1, y1, x2, y2, data)
 	end
 end
 
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -665,7 +665,7 @@ end
 
 function wml_actions.set_menu_item(cfg)
 	wesnoth.interface.set_menu_item(cfg.id, cfg)
-	wesnoth.game_events.add(cfg.id, wml_actions.command, true)
+	wesnoth.game_events.add_menu(cfg.id, wml_actions.command)
 end
 
 function wml_actions.place_shroud(cfg)
@@ -730,9 +730,7 @@ function wml_actions.event(cfg)
 		wesnoth.deprecated_message("[event]remove=yes", 2, "1.17.0", "Use [remove_event] instead of [event]remove=yes")
 		wml_actions.remove_event(cfg)
 	else
-		local delay = cfg.delayed_variable_substitution
-		if delay == nil then delay = true end
-		wesnoth.game_events.add(delay, cfg)
+		wesnoth.game_events.add_wml(cfg)
 	end
 end
 
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -663,6 +663,7 @@ end
 
 function wml_actions.set_menu_item(cfg)
 	wesnoth.interface.set_menu_item(cfg.id, cfg)
+	wesnoth.game_events.add(cfg.id, wml_actions.command, true)
 end
 
 function wml_actions.place_shroud(cfg)
@@ -727,7 +728,9 @@ function wml_actions.event(cfg)
 		wesnoth.deprecated_message("[event]remove=yes", 2, "1.17.0", "Use [remove_event] instead of [event]remove=yes")
 		wml_actions.remove_event(cfg)
 	else
-		wesnoth.add_event_handler(cfg)
+		local delay = cfg.delayed_variable_substitution
+		if delay == nil then delay = true end
+		wesnoth.game_events.add(delay, cfg)
 	end
 end
 
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -740,7 +740,7 @@ function wml_actions.remove_event(cfg)
 	local id = cfg.id or wml.error("[remove_event] missing required id= key")
 
 	for _,w in ipairs(id:split()) do
-		wesnoth.remove_event_handler(w)
+		wesnoth.game_events.remove(w)
 	end
 end
 
--- a/data/multiplayer/scenarios/2p_Isle_of_Mists.lua
+++ b/data/multiplayer/scenarios/2p_Isle_of_Mists.lua
@@ -1,6 +1,6 @@
 local _ = wesnoth.textdomain 'wesnoth-multiplayer'
 local T = wml.tag
-local on_event = wesnoth.require("on_event")
+local on_event = wesnoth.game_events.add_repeating
 
 local random_spawns = {
 	{
--- a/data/test/scenarios/events-test_actions.cfg
+++ b/data/test/scenarios/events-test_actions.cfg
@@ -8,13 +8,13 @@
 )}
 
 # Test that dynamically adding an event from Lua works
-# There's two different ways of calling game_events.add, so test both
+# There's two different ways of adding events, so test both
 {GENERIC_UNIT_TEST "event_test_lua" (
     [event]
         name=preload
         [lua]
             code=<<
-                wesnoth.game_events.add('new turn', function()
+                wesnoth.game_events.add_repeating('new turn', function()
                     unit_test.succeed()
                 end)
             >>
--- a/data/test/scenarios/events-test_actions.cfg
+++ b/data/test/scenarios/events-test_actions.cfg
@@ -1,0 +1,99 @@
+
+# This is kinda a duplicate of a basic unit test system sanity test, but... it seems to fit here as well...
+{GENERIC_UNIT_TEST "event_test_action_wml" (
+    [event]
+        name=start
+        {SUCCEED}
+    [/event]
+)}
+
+# Test that dynamically adding an event from Lua works
+# There's two different ways of calling game_events.add, so test both
+{GENERIC_UNIT_TEST "event_test_lua" (
+    [event]
+        name=preload
+        [lua]
+            code=<<
+                wesnoth.game_events.add('new turn', function()
+                    unit_test.succeed()
+                end)
+            >>
+        [/lua]
+    [/event]
+)}
+{GENERIC_UNIT_TEST "event_test_lua_advanced" (
+    [event]
+        name=preload
+        [lua]
+            code=<<
+                wesnoth.game_events.add{
+                    name = 'new turn',
+                    action = function()
+                        unit_test.succeed()
+                    end
+                }
+            >>
+        [/lua]
+    [/event]
+)}
+
+# Test that first_time_only works correctly in Lua events
+{GENERIC_UNIT_TEST "event_test_lua_repeat" (
+    [event]
+        name=preload
+        [lua]
+            code=<<
+                wml.variables.only_once = 0
+                wml.variables.should_repeat = 0
+                wesnoth.game_events.add{
+                    name = 'new turn',
+                    first_time_only = true,
+                    action = function()
+                        wml.variables.only_once = wml.variables.only_once + 1
+                    end
+                }
+                wesnoth.game_events.add{
+                    name = 'new turn',
+                    first_time_only = false,
+                    action = function()
+                        wml.variables.should_repeat = wml.variables.should_repeat + 1
+                    end
+                }
+            >>
+        [/lua]
+    [/event]
+    [event]
+        name=side 1 turn 1
+        [end_turn][/end_turn]
+    [/event]
+    [event]
+        name=side 2 turn 1
+        [end_turn][/end_turn]
+    [/event]
+    [event]
+        name=side 1 turn 2
+        [end_turn][/end_turn]
+    [/event]
+    [event]
+        name=side 2 turn 2
+        {ASSERT ({VARIABLE_CONDITIONAL only_once equals 1})}
+        {ASSERT ({VARIABLE_CONDITIONAL should_repeat equals 2})}
+        {SUCCEED}
+    [/event]
+)}
+
+# Verify that a warning is produced if adding an event that would break saved games
+{GENERIC_UNIT_TEST "event_test_lua_break_save" (
+    [event]
+        name=start
+        [lua]
+            code=<<
+                wesnoth.game_events.add{
+                    name = 'new turn',
+                    action = function() end
+                }
+                unit_test.succeed()
+            >>
+        [/lua]
+    [/event]
+)}
--- a/data/test/scenarios/events-test_filters.cfg
+++ b/data/test/scenarios/events-test_filters.cfg
@@ -230,3 +230,119 @@
         {RETURN ({VARIABLE_CONDITIONAL triggers equals 4})}
     [/event]
 )}
+
+{GENERIC_UNIT_TEST event_test_filter_wfl (
+    [event]
+        name=start
+        {VARIABLE triggers 0}
+    [/event]
+    [event]
+        name=side turn
+        first_time_only=no
+        filter_formula="turn_number = side_number"
+        {VARIABLE_OP triggers add 1}
+        {ASSERT ({VARIABLE_CONDITIONAL side_number equals $turn_number})}
+    [/event]
+    [event]
+        name=turn 3
+        {RETURN ({VARIABLE_CONDITIONAL triggers equals 2})}
+    [/event]
+    [event]
+        name=side turn
+        first_time_only=no
+        [end_turn][/end_turn]
+    [/event]
+)}
+
+{GENERIC_UNIT_TEST event_test_filter_wfl2 (
+    [event]
+        name=start
+        {VARIABLE triggers 0}
+        [do_command]
+            [move]
+                x=7,7,6,5,4,3,3,3
+                y=3,4,4,5,4,4,3,2
+            [/move]
+        [/do_command]
+        {RETURN ({VARIABLE_CONDITIONAL triggers equals 3})}
+    [/event]
+    [event]
+        name=enter hex
+        first_time_only=no
+        filter_formula="loc.x = loc.y"
+        {VARIABLE_OP triggers add 1}
+    [/event]
+)}
+
+{GENERIC_UNIT_TEST event_test_filter_lua_serializable (
+    [event]
+        name=start
+        {VARIABLE triggers 0}
+    [/event]
+    [event]
+        name=side turn
+        first_time_only=no
+        [filter_condition]
+            [lua]
+                code=<<return wesnoth.current.turn == wesnoth.current.side>>
+            [/lua]
+        [/filter_condition]
+        {VARIABLE_OP triggers add 1}
+        {ASSERT ({VARIABLE_CONDITIONAL side_number equals $turn_number})}
+    [/event]
+    [event]
+        name=turn 3
+        {RETURN ({VARIABLE_CONDITIONAL triggers equals 2})}
+    [/event]
+    [event]
+        name=side turn
+        first_time_only=no
+        [end_turn][/end_turn]
+    [/event]
+)}
+
+{GENERIC_UNIT_TEST event_test_filter_lua_dynamic (
+    [event]
+        name=preload
+        [lua]
+            code=<<
+                wml.variables.triggers = 0
+                wesnoth.game_events.add{
+                    name = 'side turn',
+                    first_time_only = false,
+                    filter = function()
+                        return wesnoth.current.turn == wesnoth.current.side
+                    end,
+                    action = function()
+                        wml.variables.triggers = wml.variables.triggers + 1
+                        unit_test.assert_equal(wesnoth.current.turn, wesnoth.current.side, 'filter correctly passed')
+                    end
+                }
+            >>
+        [/lua]
+    [/event]
+    [event]
+        name=side 2 turn 3
+        {RETURN ({VARIABLE_CONDITIONAL triggers equals 2})}
+    [/event]
+    [event]
+        name=side turn
+        first_time_only=no
+        [end_turn][/end_turn]
+    [/event]
+)}
+
+{GENERIC_UNIT_TEST event_test_filter_lua_dynamic_break_save (
+    [event]
+        name=start
+        [lua]
+            code=<<
+                wesnoth.game_events.add{
+                    name = 'new turn',
+                    filter = function() end
+                }
+                unit_test.succeed()
+            >>
+        [/lua]
+    [/event]
+)}
--- a/src/formula/callable_objects.cpp
+++ b/src/formula/callable_objects.cpp
@@ -30,6 +30,9 @@
 #include "game_board.hpp"
 #include "game_version.hpp"
 #include "resources.hpp"
+#include "tod_manager.hpp"
+#include "play_controller.hpp"
+#include "game_events/pump.hpp"
 
 static lg::log_domain log_scripting_formula("scripting/formula");
 #define LOG_SF LOG_STREAM(info, log_scripting_formula)
@@ -890,6 +893,89 @@ void safe_call_result::get_inputs(formula_input_vector& inputs) const
 	if(current_unit_location_ != map_location()) {
 		add_input(inputs, "current_loc");
 	}
+}
+
+void gamestate_callable::get_inputs(formula_input_vector &inputs) const
+{
+	add_input(inputs, "turn_number");
+	add_input(inputs, "time_of_day");
+	add_input(inputs, "side_number");
+	add_input(inputs, "sides");
+	add_input(inputs, "units");
+	add_input(inputs, "map");
+}
+
+variant gamestate_callable::get_value(const std::string &key) const
+{
+	if(key == "turn_number") {
+		return variant(resources::tod_manager->turn());
+	} else if(key == "time_of_day") {
+		return variant(resources::tod_manager->get_time_of_day().id);
+	} else if(key == "side_number") {
+		return variant(resources::controller->current_side());
+	} else if(key == "sides") {
+		std::vector<variant> vars;
+		for(const auto& team : resources::gameboard->teams()) {
+			vars.emplace_back(std::make_shared<team_callable>(team));
+		}
+		return variant(vars);
+	} else if(key == "units") {
+		std::vector<variant> vars;
+		for(const auto& unit : resources::gameboard->units()) {
+			vars.emplace_back(std::make_shared<unit_callable>(unit));
+		}
+		return variant(vars);
+	} else if(key == "map") {
+		return variant(std::make_shared<gamemap_callable>(*resources::gameboard));
+	}
+
+	return variant();
+}
+
+void event_callable::get_inputs(formula_input_vector &inputs) const
+{
+	add_input(inputs, "event");
+	add_input(inputs, "event_id");
+	add_input(inputs, "event_data");
+	add_input(inputs, "loc");
+	add_input(inputs, "unit");
+	add_input(inputs, "weapon");
+	add_input(inputs, "second_loc");
+	add_input(inputs, "second_unit");
+	add_input(inputs, "second_weapon");
+}
+
+variant event_callable::get_value(const std::string &key) const
+{
+	if(key == "event") {
+		return variant(event_info.name);
+	} else if(key == "event_id") {
+		return variant(event_info.id);
+	} else if(key == "loc") {
+		return variant(std::make_shared<location_callable>(event_info.loc1));
+	} else if(key == "second_loc") {
+		return variant(std::make_shared<location_callable>(event_info.loc2));
+	} else if(key == "event_data") {
+		return variant(std::make_shared<config_callable>(event_info.data));
+	} else if(key == "unit") {
+		if(auto u1 = event_info.loc1.get_unit()) {
+			return variant(std::make_shared<unit_callable>(*u1));
+		}
+	} else if(key == "second_unit") {
+		if(auto u2 = event_info.loc2.get_unit()) {
+			return variant(std::make_shared<unit_callable>(*u2));
+		}
+	} else if(key == "weapon") {
+		if(event_info.data.has_child("first")) {
+			return variant(std::make_shared<attack_type_callable>(attack_type(event_info.data.child("first"))));
+		}
+	} else if(key == "second_weapon") {
+		if(event_info.data.has_child("second")) {
+			return variant(std::make_shared<attack_type_callable>(attack_type(event_info.data.child("second"))));
+		}
+	}
+
+	return variant();
 }
 
 } // namespace wfl
--- a/src/formula/callable_objects.hpp
+++ b/src/formula/callable_objects.hpp
@@ -26,9 +26,32 @@ class team;
 class terrain_type;
 class unit;
 class unit_type;
+namespace game_events {struct queued_event;}
 
 namespace wfl
 {
+
+/** An object representing the state of the game, providing access to the map and basic information */
+class gamestate_callable : public formula_callable
+{
+public:
+	// Currently it accesses all its state through the resources namespace, so nothing is passed into it.
+
+	void get_inputs(formula_input_vector& inputs) const override;
+	variant get_value(const std::string& key) const override;
+};
+
+/** An object representing the state of the current event; equivalent to Lua's wesnoth.current.event_context */
+class event_callable : public formula_callable
+{
+public:
+	event_callable(const game_events::queued_event& event) : event_info(event) {}
+
+	void get_inputs(formula_input_vector& inputs) const override;
+	variant get_value(const std::string& key) const override;
+private:
+	const game_events::queued_event& event_info;
+};
 
 class terrain_callable : public formula_callable
 {
--- a/src/game_events/entity_location.cpp
+++ b/src/game_events/entity_location.cpp
@@ -89,8 +89,7 @@ bool entity_location::matches_unit(const unit_map::const_iterator & un_it) const
  * the unit is required to additionally match the unit that was supplied
  * when this was constructed.
  */
-bool entity_location::matches_unit_filter(const unit_map::const_iterator & un_it,
-                                          const vconfig & filter) const
+bool entity_location::matches_unit_filter(const unit_map::const_iterator & un_it, const unit_filter& filter) const
 {
 	if ( !un_it.valid() )
 		return false;
@@ -101,7 +100,7 @@ bool entity_location::matches_unit_filter(const unit_map::const_iterator & un_it
 
 	// Filter the unit at the filter location (should be the unit's
 	// location if no special filter location was specified).
-	return unit_filter(filter).matches(*un_it, filter_loc_)  &&
+	return filter.matches(*un_it, filter_loc_)  &&
 	       matches_unit(un_it);
 }
 
--- a/src/game_events/handlers.cpp
+++ b/src/game_events/handlers.cpp
@@ -40,13 +40,14 @@ namespace game_events
 {
 /* ** event_handler ** */
 
-event_handler::event_handler(config&& cfg, bool imi, const std::vector<std::string>& types)
+event_handler::event_handler(config&& cfg, bool imi, const std::vector<std::string>& types, game_lua_kernel& lk)
 	: first_time_only_(cfg["first_time_only"].to_bool(true))
 	, is_menu_item_(imi)
 	, disabled_(false)
 	, cfg_(cfg)
 	, types_(types)
 {
+	event_ref_ = lk.save_wml_event(cfg);
 }
 
 void event_handler::disable()
@@ -69,7 +70,7 @@ void event_handler::handle_event(const queued_event& event_info, game_lua_kernel
 		disable();
 	}
 
-	lk.run_wml_action("command", vconfig(cfg_, false), event_info);
+	lk.run_wml_event(event_ref_, vconfig(cfg_, false), event_info);
 	sound::commit_music_changes();
 }
 
--- a/src/game_events/handlers.cpp
+++ b/src/game_events/handlers.cpp
@@ -24,6 +24,8 @@
 #include "game_events/pump.hpp"
 #include "game_events/manager_impl.hpp" // for standardize_name
 
+#include "formula/callable_objects.hpp"
+#include "formula/formula.hpp"
 #include "formula/string_utils.hpp"
 #include "game_board.hpp"
 #include "game_data.hpp"
@@ -211,6 +213,28 @@ private:
 	bool first_;
 };
 
+struct filter_formula : public event_filter {
+	filter_formula(const std::string& formula) : formula_(formula) {}
+	bool operator()(const queued_event& event_info) const override
+	{
+		wfl::gamestate_callable gs;
+		wfl::event_callable evt(event_info);
+		wfl::formula_callable_with_backup data(evt, gs);
+		return formula_.evaluate(data).as_bool();
+	}
+	void serialize(config& cfg) const override
+	{
+		std::string code = formula_.str();
+		if(cfg.has_attribute("filter_formula")) {
+			// This will probably never happen in practice, but handle it just in case it somehow can
+			code = "(" + cfg["filter_formula"].str() + ") and (" + code + ")";
+		}
+		cfg["filter_formula"] = code;
+	}
+private:
+	wfl::formula formula_;
+};
+
 void event_handler::read_filters(const config &cfg)
 {
 	for(auto filter : cfg.all_children_range()) {
@@ -228,6 +252,9 @@ void event_handler::read_filters(const config &cfg)
 		} else if(filter.key == "filter_second_attack") {
 			add_filter(std::make_unique<filter_attack>(vcfg, false));
 		}
+	}
+	if(cfg.has_attribute("filter_formula")) {
+		add_filter(std::make_unique<filter_formula>(cfg["filter_formula"]));
 	}
 }
 
--- a/src/game_events/handlers.cpp
+++ b/src/game_events/handlers.cpp
@@ -20,11 +20,20 @@
  */
 
 #include "game_events/handlers.hpp"
+#include "game_events/conditional_wml.hpp"
+#include "game_events/pump.hpp"
+#include "game_events/manager_impl.hpp" // for standardize_name
 
+#include "formula/string_utils.hpp"
+#include "game_board.hpp"
 #include "game_data.hpp"
 #include "log.hpp"
+#include "play_controller.hpp"
+#include "resources.hpp"
 #include "scripting/game_lua_kernel.hpp"
+#include "side_filter.hpp"
 #include "sound.hpp"
+#include "units/filter.hpp"
 #include "variable.hpp"
 
 static lg::log_domain log_engine("engine");
@@ -40,14 +49,44 @@ namespace game_events
 {
 /* ** event_handler ** */
 
-event_handler::event_handler(config&& cfg, bool imi, const std::vector<std::string>& types, game_lua_kernel& lk)
-	: first_time_only_(cfg["first_time_only"].to_bool(true))
-	, is_menu_item_(imi)
+event_handler::event_handler(const std::string& types, const std::string& id)
+	: first_time_only_(true)
+	, is_menu_item_(false)
 	, disabled_(false)
-	, cfg_(cfg)
+	, is_lua_(false)
+	, id_(id)
 	, types_(types)
+{}
+
+std::vector<std::string> event_handler::names(const variable_set* vars) const
 {
-	event_ref_ = lk.save_wml_event(cfg);
+	std::string names = types_;
+
+	// Do some standardization on the name field.
+	// Split the name field and standardize each one individually.
+	// This ensures they're all valid for by-name lookup.
+	std::vector<std::string> standardized_names;
+	for(std::string single_name : utils::split(names)) {
+		if(utils::might_contain_variables(single_name)) {
+			if(!vars) {
+				// If we don't have gamedata, we can't interpolate variables, so there's
+				// no way the name will match. Move on to the next one in that case.
+				continue;
+			}
+			single_name = utils::interpolate_variables_into_string(single_name, *vars);
+		}
+		// Variable interpolation could've introduced additional commas, so split again.
+		for(const std::string& subname : utils::split(single_name)) {
+			standardized_names.emplace_back(event_handlers::standardize_name(subname));
+		}
+	}
+
+	return standardized_names;
+}
+
+bool event_handler::empty() const
+{
+	return args_.empty();
 }
 
 void event_handler::disable()
@@ -63,15 +102,150 @@ void event_handler::handle_event(const queued_event& event_info, game_lua_kernel
 	}
 
 	if(is_menu_item_) {
-		DBG_NG << cfg_["name"] << " will now invoke the following command(s):\n" << cfg_;
+		DBG_NG << "menu item " << id_ << " will now invoke the following command(s):\n" << args_;
 	}
 
 	if(first_time_only_) {
 		disable();
 	}
 
-	lk.run_wml_event(event_ref_, vconfig(cfg_, false), event_info);
+	lk.run_wml_event(event_ref_, vconfig(args_, false), event_info);
 	sound::commit_music_changes();
 }
+
+bool event_handler::filter_event(const queued_event& ev) const
+{
+	return std::all_of(filters_.begin(), filters_.end(), [&ev](const auto& filter) {
+		return (*filter)(ev);
+	});
+}
+
+void event_handler::write_config(config &cfg) const
+{
+	if(disabled_) {
+		WRN_NG << "Tried to serialize disabled event, skipping";
+		return;
+	}
+	if(is_lua_) {
+		WRN_NG << "Skipping serialization of an event bound to Lua code";
+		return;
+	}
+	if(!types_.empty()) cfg["name"] = types_;
+	if(!id_.empty()) cfg["id"] = id_;
+	cfg["first_time_only"] = first_time_only_;
+	for(const auto& filter : filters_) {
+		filter->serialize(cfg);
+	}
+	cfg.append(args_);
+}
+
+void event_filter::serialize(config&) const
+{
+	WRN_NG << "Tried to serialize an event with a filter that cannot be serialized!";
+}
+
+struct filter_condition : public event_filter {
+	filter_condition(const vconfig& cfg) : cfg_(cfg.make_safe()) {}
+	bool operator()(const queued_event&) const override
+	{
+		return conditional_passed(cfg_);
+	}
+	void serialize(config& cfg) const override
+	{
+		cfg.add_child("filter_condition", cfg_.get_config());
+	}
+private:
+	vconfig cfg_;
+};
+
+struct filter_side : public event_filter {
+	filter_side(const vconfig& cfg) : ssf_(cfg.make_safe(), &resources::controller->gamestate()) {}
+	bool operator()(const queued_event&) const override
+	{
+		return ssf_.match(resources::controller->current_side());
+	}
+	void serialize(config& cfg) const override
+	{
+		cfg.add_child("filter_side", ssf_.get_config());
+	}
+private:
+	side_filter ssf_;
+};
+
+struct filter_unit : public event_filter {
+	filter_unit(const vconfig& cfg, bool first) : suf_(cfg.make_safe()), first_(first) {}
+	bool operator()(const queued_event& event_info) const override
+	{
+		const auto& loc = first_ ? event_info.loc1 : event_info.loc2;
+		auto unit = resources::gameboard->units().find(loc);
+		return loc.matches_unit_filter(unit, suf_);
+	}
+	void serialize(config& cfg) const override
+	{
+		cfg.add_child(first_ ? "filter" : "filter_second", suf_.to_config());
+	}
+private:
+	unit_filter suf_;
+	bool first_;
+};
+
+struct filter_attack : public event_filter {
+	filter_attack(const vconfig& cfg, bool first) : swf_(cfg.make_safe()), first_(first) {}
+	bool operator()(const queued_event& event_info) const override
+	{
+		const unit_map& units = resources::gameboard->units();
+		const auto& loc = first_ ? event_info.loc1 : event_info.loc2;
+		auto unit = units.find(loc);
+		if(unit != units.end() && loc.matches_unit(unit)) {
+			const config& attack = event_info.data.child(first_ ? "first" : "second");
+			return swf_.empty() || matches_special_filter(attack, swf_);
+		}
+		return false;
+	}
+	void serialize(config& cfg) const override
+	{
+		cfg.add_child(first_ ? "filter_attack" : "filter_second_attack", swf_.get_config());
+	}
+private:
+	vconfig swf_;
+	bool first_;
+};
+
+void event_handler::read_filters(const config &cfg)
+{
+	for(auto filter : cfg.all_children_range()) {
+		vconfig vcfg(filter.cfg);
+		if(filter.key == "filter_condition") {
+			add_filter(std::make_unique<filter_condition>(vcfg));
+		} else if(filter.key == "filter_side") {
+			add_filter(std::make_unique<filter_side>(vcfg));
+		} else if(filter.key == "filter") {
+			add_filter(std::make_unique<filter_unit>(vcfg, true));
+		} else if(filter.key == "filter_attack") {
+			add_filter(std::make_unique<filter_attack>(vcfg, true));
+		} else if(filter.key == "filter_second") {
+			add_filter(std::make_unique<filter_unit>(vcfg, false));
+		} else if(filter.key == "filter_second_attack") {
+			add_filter(std::make_unique<filter_attack>(vcfg, false));
+		}
+	}
+}
+
+void event_handler::add_filter(std::unique_ptr<event_filter>&& filter)
+{
+	filters_.push_back(std::move(filter));
+}
+
+void event_handler::register_wml_event(game_lua_kernel &lk)
+{
+	event_ref_ = lk.save_wml_event();
+}
+
+void event_handler::set_event_ref(int idx)
+{
+	event_ref_ = idx;
+	is_lua_ = true;
+}
+
 
 } // end namespace game_events
--- a/src/game_events/handlers.cpp
+++ b/src/game_events/handlers.cpp
@@ -278,22 +278,61 @@ private:
 	wfl::formula formula_;
 };
 
+static std::unique_ptr<event_filter> make_filter(const std::string& key, const vconfig& contents)
+{
+	if(key == "filter_condition") {
+		return std::make_unique<filter_condition>(contents);
+	} else if(key == "filter_side") {
+		return std::make_unique<filter_side>(contents);
+	} else if(key == "filter") {
+		return std::make_unique<filter_unit>(contents, true);
+	} else if(key == "filter_attack") {
+		return std::make_unique<filter_attack>(contents, true);
+	} else if(key == "filter_second") {
+		return std::make_unique<filter_unit>(contents, false);
+	} else if(key == "filter_second_attack") {
+		return std::make_unique<filter_attack>(contents, false);
+	}
+	return nullptr;
+}
+
+/**
+ * This is a dynamic wrapper for any filter type, specified via [insert_tag].
+ * It loads the filter contents from a variable and forwards it to the appropriate filter class.
+ */
+struct filter_dynamic : public event_filter {
+	filter_dynamic(const std::string& tag, const std::string& var) : tag_(tag), var_(var) {}
+	bool operator()(const queued_event& event_info) const override
+	{
+		variable_access_const variable(var_, resources::gamedata->get_variables());
+		if(!variable.exists_as_container()) return false;
+		if(auto filter = make_filter(tag_, vconfig(variable.as_container()))) {
+			return (*filter)(event_info);
+		}
+		return false;
+	}
+	void serialize(config& cfg) const override
+	{
+		auto tag = cfg.add_child("insert_tag");
+		tag["name"] = tag_;
+		tag["variable"] = var_;
+	}
+	bool can_serialize() const override
+	{
+		return true;
+	}
+private:
+	std::string tag_, var_;
+};
+
 void event_handler::read_filters(const config &cfg)
 {
 	for(auto filter : cfg.all_children_range()) {
 		vconfig vcfg(filter.cfg);
-		if(filter.key == "filter_condition") {
-			add_filter(std::make_unique<filter_condition>(vcfg));
-		} else if(filter.key == "filter_side") {
-			add_filter(std::make_unique<filter_side>(vcfg));
-		} else if(filter.key == "filter") {
-			add_filter(std::make_unique<filter_unit>(vcfg, true));
-		} else if(filter.key == "filter_attack") {
-			add_filter(std::make_unique<filter_attack>(vcfg, true));
-		} else if(filter.key == "filter_second") {
-			add_filter(std::make_unique<filter_unit>(vcfg, false));
-		} else if(filter.key == "filter_second_attack") {
-			add_filter(std::make_unique<filter_attack>(vcfg, false));
+		if(auto filter_ptr = make_filter(filter.key, vcfg)) {
+			add_filter(std::move(filter_ptr));
+		} else if(filter.key == "insert_tag" && make_filter(vcfg["name"], vconfig::empty_vconfig())) {
+			add_filter(std::make_unique<filter_dynamic>(vcfg["name"], vcfg["variable"]));
 		}
 	}
 	if(cfg.has_attribute("filter_formula")) {
--- a/src/game_events/handlers.cpp
+++ b/src/game_events/handlers.cpp
@@ -128,8 +128,26 @@ void event_handler::write_config(config &cfg) const
 		WRN_NG << "Tried to serialize disabled event, skipping";
 		return;
 	}
+	static const char* log_append_preload = " - this will not break saves since it was registered during or before preload\n";
+	static const char* log_append_postload = " - this will break saves because it was registered after preload\n";
 	if(is_lua_) {
-		WRN_NG << "Skipping serialization of an event bound to Lua code";
+		static const char* log = "Skipping serialization of an event with action bound to Lua code";
+		if(has_preloaded_){
+			WRN_NG << log << log_append_postload;
+			lg::log_to_chat() << log << log_append_postload;
+		} else {
+			LOG_NG << log << log_append_preload;
+		}
+		return;
+	}
+	if(!std::all_of(filters_.begin(), filters_.end(), std::mem_fn(&event_filter::can_serialize))) {
+		static const char* log = "Skipping serialization of an event with filter bound to Lua code";
+		if(has_preloaded_) {
+			WRN_NG << log << log_append_postload;
+			lg::log_to_chat() << log << log_append_postload;
+		} else {
+			LOG_NG << log << log_append_preload;
+		}
 		return;
 	}
 	if(!types_.empty()) cfg["name"] = types_;
@@ -146,6 +164,11 @@ void event_filter::serialize(config&) const
 	WRN_NG << "Tried to serialize an event with a filter that cannot be serialized!";
 }
 
+bool event_filter::can_serialize() const
+{
+	return false;
+}
+
 struct filter_condition : public event_filter {
 	filter_condition(const vconfig& cfg) : cfg_(cfg.make_safe()) {}
 	bool operator()(const queued_event&) const override
@@ -155,6 +178,10 @@ struct filter_condition : public event_filter {
 	void serialize(config& cfg) const override
 	{
 		cfg.add_child("filter_condition", cfg_.get_config());
+	}
+	bool can_serialize() const override
+	{
+		return true;
 	}
 private:
 	vconfig cfg_;
@@ -169,6 +196,10 @@ struct filter_side : public event_filter {
 	void serialize(config& cfg) const override
 	{
 		cfg.add_child("filter_side", ssf_.get_config());
+	}
+	bool can_serialize() const override
+	{
+		return true;
 	}
 private:
 	side_filter ssf_;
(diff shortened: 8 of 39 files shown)


Pentarctagon:
Unit tests proving out the different aspects of this functionality would be very beneficial.

Pentarctagon:
Adding doxygen doc comments to methods is also a good habit for us to all get into, I think.

src/formula/callable_objects.hpp
```diff
@@ -25,10 +25,31 @@ class team;
...
 class unit_type;
+namespace game_events {class queued_event;}
 
 namespace wfl
 {
 
+class gamestate_callable : public formula_callable
+{
+public:
+	// TODO: Constructor
```
stevecotton, line 36: Without a constructor, is there any data that's accessible via this object?

Could the `world`-related bits be removed from the current PR, and added as a separate change later?

src/game_events/handlers.cpp
```diff
@@ -41,13 +52,49 @@ namespace game_events
...
+			if(utils::might_contain_variables(single_name) && !vars) {
+				// If we don't have gamedata, we can't interpolate variables, so there's
+				// no way the name will match. Move on to the next one in that case.
+				single_name.clear();
+			} else {
+				single_name = event_handlers::standardize_name(single_name);
+			}
+		}
+
+		assert(!standardized_names.empty());
```
stevecotton, line 86: How do we know that UMC can't trigger this assert? It looks as if it might trigger when there's a `$` in a name.

src/game_events/handlers.cpp
```diff
@@ -63,15 +110,210 @@ void event_handler::handle_event(const queued_event& event_info, game_lua_kernel
...
+		return resources::lua_kernel->run_wml_event(event_ref_, args_, event_info, &result) && result;
+	}
+	void serialize(config& cfg) const override
+	{
+		if(cfg.has_attribute(("filter"))) {
+			WRN_NG << "Somehow an event has two filter= attributes";
+		}
+		cfg["filter"] = code_;
+	}
+	~filter_lua()
```
stevecotton, line 261: Remember the Rule of Three, otherwise there are many ways to get `clear_wml_event` calls without corresponding `save_wml_event` calls.

src/formula/callable_objects.hpp
CelticMinstrel: It's accessing all its data through the `resources` namespace, so yeah, there's data accessible via this object.

src/game_events/handlers.cpp
CelticMinstrel: This code was copy-pasted from elsewhere and then altered, so it's possible that this assertion just isn't correct anymore.

src/game_events/handlers.cpp
CelticMinstrel: You mean having a move constructor, right? I think the existing use is safe due to guaranteed copy elision, but I guess it can't hurt to be extra certain.

I'm pretty sure an extra unnecessary `clear_wml_event` call on the same reference is safe, though; but if it you called `run_wml_event` on a reference that was already cleared, that might be bad.

stevecotton: Mainly it's about the copy constructor and copy assignment. Rule of Three is that the compiler generates those for you unless you tell it not to, and so the destructor can run more often than the user-defined constructor does.

The move constructor and move assignment aren't automatically generated if there's a user-defined destructor, but for backward compatibility the copy versions still are.

CelticMinstrel: So then what you want me to do is delete the copy constructor?

stevecotton: Both the copy constructor and the copy assignment.

src/game_events/manager_impl.cpp
```diff
@@ -89,68 +88,55 @@ handler_list& event_handlers::get(const std::string& name)
...
+	// Register the new handler.
 	// Do note active_ holds the main shared_ptr, and the other three containers
 	// construct weak_ptrs from the shared one.
-	DBG_EH << "inserting event handler for name=" << name << " with id=" << id << "\n";
-	active_.emplace_back(new event_handler(std::move(event_cfg), is_menu_item, standardized_names));
+	DBG_EH << "inserting event handler for name=" << names << " with id=" << id << "\n";
+	active_.emplace_back(handler);
 
 	//
 	//  !! event_cfg is invalid past this point! DO NOT USE!
```
stevecotton, line 132: This comment can be removed, the code it's referring to is gone.

src/game_events/manager_impl.hpp
```diff
@@ -20,8 +20,26 @@
...
+	handler_ptr handler_;
+public:
+	pending_event_handler(event_handlers& list, handler_ptr handler)
+		: list_(list)
+		, handler_(handler)
+	{}
+	bool valid() const {return handler_.get();}
+	event_handler* operator->() {return handler_.get();}
+	event_handler& operator*() {return *handler_;}
+	~pending_event_handler();
```
stevecotton, line 40: Rule of Three, although for this one I think it should be moveable but not copyable.

src/scripting/lua_kernel_base.cpp
```diff
@@ -711,7 +711,7 @@ bool lua_kernel_base::protected_call(lua_State * L, int nArgs, int nRets, error_
 bool lua_kernel_base::load_string(char const * prog, const std::string& name, error_handler e_h)
 {
 	// pass 't' to prevent loading bytecode which is unsafe and can be used to escape the sandbox.
-	int errcode = luaL_loadbufferx(mState, prog, strlen(prog), name.empty() ? name.c_str() : prog, "t");
+	int errcode = luaL_loadbufferx(mState, prog, strlen(prog), name.empty() ? prog : name.c_str(), "t");
```
stevecotton, line 714: This line looks serious enough to have a bug report and PR on its own. A unit test that would fail with the old code would also be good.

data/lua/wml-tags.lua
```diff
@@ -148,10 +148,12 @@ function wml_actions.fire_event(cfg)
...
 	local w2 = wml.get_child(cfg, "secondary_attack")
-	if w2 then w1 = w1 or {} end
+	local data = wml.get_child(cfg, "data") or {}
+	if w1 then table.insert(data, wml.tag.first(w1)) end
+	if w2 then table.insert(data, wml.tag.second(w2)) end
 
-	if cfg.id and cfg.id ~= "" then wesnoth.fire_event_by_id(cfg.id, x1, y1, x2, y2, w1, w2)
-	elseif cfg.name and cfg.name ~= "" then wesnoth.fire_event(cfg.name, x1, y1, x2, y2, w1, w2)
+	if cfg.id and cfg.id ~= "" then wesnoth.game_events.fire_event_by_id(cfg.id, x1, y1, x2, y2, data)
+	elseif cfg.name and cfg.name ~= "" then wesnoth.gamee_events.fire_event(cfg.name, x1, y1, x2, y2, data)
```
stevecotton, line 156: typo "gamee"

src/scripting/lua_kernel_base.cpp
CelticMinstrel: It's not all that serious since this is just name shown in error messages, and the bug is just that our code doesn't use the name it's supposed to use.

But I can extract it into a separate commit, sure, and maybe add a unit test if I can think of a way to do it.

CelticMinstrel:
I think this addresses _most_ of the concerns people mentioned. I still have a few things to finish though:

* [ ] delayed_variable_substitution=yes seems to be broken
* [ ] [remove_event] seems to be broken
* [ ] make pending_event_handler moveable
* [ ] extract the code chunk name fix

src/scripting/lua_kernel_base.cpp
CelticMinstrel: I did manage to make a unit test for this, but it's reliant on this PR, so I don't think I can extract the unit test to a separate PR.

src/formula/callable_objects.cpp
```diff
@@ -849,4 +860,87 @@ void safe_call_result::get_inputs(formula_input_vector& inputs) const
...
+		}
+	} else if(key == "unit2") {
+		if(auto u2 = event_info.loc2.get_unit()) {
+			return variant(std::make_shared<unit_callable>(*u2));
+		}
+	} else if(key == "attack1") {
+		if(event_info.data.has_child("first")) {
+			return variant(std::make_shared<attack_type_callable>(attack_type(event_info.data.child("first"))));
+		}
+	} else if(key == "attack2") {
```
stevecotton, line 937: Is this creating a set of new names for existing concepts, similar to "repeating" vs "first_time_only"?

stevecotton (changes requested):
There still isn't enough documentation in this PR to enable the PR to be reviewed.

Change requested: add the documentation.

src/formula/callable_objects.cpp
CelticMinstrel: Maybe? I guess in wesnoth.current they're called `weapon` and `second_weapon`…

CelticMinstrel:
The documentation goes on the wiki though, right? It's not part of the PR.

src/formula/callable_objects.cpp
stevecotton: I think the API should be consistent, as discussed in https://www.wesnoth.org/irclogs/2021/04/%23wesnoth-dev.2021-04-11.log between 12:00 and 19:45.

While searching for an example I noticed there was no unit test for formulas using these values.

stevecotton:
There's more than just API docs to add. I think docs should be written like the game Taboo - try to explain what it does without repeating the words in the class name, function name, argument types or argument names; that often acts as a design review.

The `gamestate_callable` class could do with an explanation of what it is, and that it doesn't need any state passed to the constructor.

The `event_callable` and `pending_event_handler` classes could do with docs too, so could `event_filter`.

src/formula/callable_objects.cpp
Pentarctagon: Please don't use different names for existing functionality and concepts - even though this particular functionality is new, it doesn't exist in a vacuum, and should follow existing names even if you don't think they're completely ideal.

Otherwise in a few years someone else is probably going to try and propose changing these new names to be consistent with the existing names anyway, and I don't think any UMC authors are going to be happy to deal with the names starting off different between the APIs and then anyone who uses to new API having to do a bunch of boilerplate work to change the names later.

CelticMinstrel:
Does that help?

src/formula/callable_objects.cpp
CelticMinstrel: > While searching for an example I noticed there was no unit test for formulas using these values.

Okay, added. It's a good thing you mentioned this, because those values actually didn't work due to them not being valid WFL identifiers.

src/formula/callable_objects.cpp
```diff
@@ -849,4 +860,87 @@ void safe_call_result::get_inputs(formula_input_vector& inputs) const
 	}
 }
 
+void gamestate_callable::get_inputs(formula_input_vector &inputs) const
+{
+	add_input(inputs, "turn");
+	add_input(inputs, "time_of_day");
+	add_input(inputs, "current_side");
```
CelticMinstrel, line 867: I'm wondering about what I should call the turn and side numbers here. In Lua they are `wesnoth.current.turn` and `wesnoth.current.side`, while in WML they are `$turn_number` and `$side_number`. That suggests three possibilities for WFL:

* `turn` and `side` (I can't remember if there was a reason I didn't use `side`…)
* `current_turn` and `current_side` (like Lua but that dot becomes an underscore)
* `turn_number` and `side_number` (like WML without the `$`)

Pentarctagon: I'd say `turn_number` and `side_number`.

CelticMinstrel:
I drafted up some documentation for the [Lua API](https://wiki.wesnoth.org/LuaAPI/wesnoth/game_events) on the wiki.

I realize it's perhaps a little tucked away in this PR, but does anyone have any comments on the Lua `unit_test` API that I've added here?

stevecotton:
> I realize it's perhaps a little tucked away in this PR, but does anyone have any comments on the Lua `unit_test` API that I've added here?

I'll do a review this evening, but was going to ask if you'd split this into a separate PR.

src/scripting/lua_kernel_base.cpp
stevecotton: Thanks for creating the test. Sorry for suggesting a separate PR for this change, I misunderstood what it did.

data/test/scenarios/test_assert.cfg
```diff
@@ -27,3 +27,68 @@
...
+        code=<<
+            unit_test.assert_equal('foo', 'bar', 'string equality')
+            unit_test.assert_not_equal('foo', 'foo', 'string inequality')
+            unit_test.assert_less('foo', 'bar', 'string ordering')
+            unit_test.assert_equal(1, 2, 'numeric equality')
+            unit_test.assert_not_equal(1, 1, 'numeric inequality')
+            unit_test.assert_greater(1, 2, 'numeric greater-than')
+            unit_test.assert_less(2, 1, 'numeric less-than')
+            unit_test.assert_in_range(1, 2, 5, 'numeric in-range')
+            unit_test.assert_approx_equal(1, 1.001, 0.0001, 'numeric approx-equal')
```
stevecotton, line 90: This has to be split into 9 separate tests, otherwise it only proves that at least one of the checks works.

data/test/scenarios/test_return.cfg
```diff
@@ -15,3 +15,17 @@
...
+{GENERIC_UNIT_TEST "test_lua_return" (
+    [event]
+        name = start
+        code=<<unit_test.pass_if(true, 'return true')>>
+    [/event]
+)}
+{GENERIC_UNIT_TEST "test_lua_return_fail" (
+    [event]
+        name = start
+        code=<<unit_test.pass_if(false, 'return false')>>
```
stevecotton, line 29: I'd prefer `unit_test.succeed()` for the success cases, and replacing the `unit_test.pass_if(err:find('test')` with `unit_test.assert_something(err:find('test'))`.

For WML it made sense to have macros for conditionals, but that's because the WML syntax for conditional statements is bad.

src/utils/guard_value.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+   See the COPYING file for more details.
+*/
+
+#pragma once
+
+namespace utils {
+
+template<typename T>
+class guard_value {
```
stevecotton, line 19: Needs documentation. Please don't wait for a review comment on each class.

stevecotton:
The CI builds are failing, and for the GCC builds the first failure is a really easy-to-fix issue that's been in the branch since 0562b9a2e3bba01dc64b646a6c2e2c72af2403ed. There's more issues to come after that, as there are (different) bugs reported by Clang, the schema and by `run_wml_tests`; however my main point is that the CI tools could be helping us here, if you address their complaints now instead of waiting until after people have reviewed.

src/scripting/game_lua_kernel.hpp
```diff
@@ -57,8 +58,12 @@ class game_lua_kernel : public lua_kernel_base
 	tod_manager & tod_man();
 
 	config level_lua_;
+	int EVENT_TABLE;
+	bool has_preloaded_ = false;
 
 	std::stack<game_events::queued_event const * > queued_events_;
+	
+	utils::guard_value<game_lua_kernel*> resources_ptr_;
```
stevecotton, line 66: If there are two lua kernels, how do we know which order they'll be destroyed in?

CelticMinstrel:
> I'll do a review this evening, but was going to ask if you'd split this into a separate PR.

I _think_ that's possible. It feels a little pointless when nothing would be using it, though.

data/test/scenarios/test_return.cfg
CelticMinstrel: So basically, remove `pass_if` and `fail_if`? The `pass_if` function was intended to be equivalent to the `{RETURN}` macro.

src/utils/guard_value.hpp
CelticMinstrel: I think this class is pretty self-explanatory, but sure, I'll add a comment.

src/scripting/game_lua_kernel.hpp
CelticMinstrel: There aren't two Lua kernels.

CelticMinstrel:
The two schema commits are duplicated from master, so they'll go away once I rebase.

CelticMinstrel:
Looks like all the remaining CI failures are various WML unit tests, and I suspect they are likely to be related to `delayed_variable_substitution`…

stevecotton:
> > I'll do a review this evening, but was going to ask if you'd split this into a separate PR.
> 
> I _think_ that's possible. It feels a little pointless when nothing would be using it, though.

At the moment it seems the rate-limit on Wesnoth development is a shortage of reviewers' time, and I'm expecting the current PR end up with a lot of comments and requests to change stuff. Each self-contained bit that can be split off and reviewed in isolation is likely to merge faster, and also to make this PR smaller.

src/scripting/lua_kernel_base.cpp
CelticMinstrel: To be clear, I meant the test is reliant on this PR, not that the change is. Even if I extracted the change to a separate PR, the test wouldn't be part of the PR.

src/scripting/game_lua_kernel.hpp
stevecotton: Then why do we need to store the old value of `resources::lua_kernel`, instead of knowing that it must be `nullptr`?
And why do it here, instead of doing it in the caller that's creating and destroying the instance of `game_lua_kernel`?

Basically, for every change in a071adc601fa68252713a87c53ff44e0d7bf9512 I'm wondering why it's being done this way.

stevecotton:
The schema is still failing, on
* `[event]code=`
* `[event]filter_formula=`

CelticMinstrel:
Oh right, I guess I spoke too soon on #5709 - the _existing_ schema commits will go away but I also need some new ones.

stevecotton:
> Instead of using ActionWML, you can write the entire event inline as Lua code using `[event]code=`.

What advantage does this feature give? If the author doesn't have use much Lua then they could just use `[event][lua]code=`; but if they have a lot of Lua then it seems likely to me that they'd have one WML event that called `dofile`, and register all of their events directly from the Lua file.

CelticMinstrel:
> What advantage does this feature give?

It doesn't directly give any advantage over using `[lua]` in ActionWML, but the new internals make it trivial to implement and if ActionWML were ever deprecated it would become the only way to write an event.

> if they have a lot of Lua then it seems likely to me that they'd have one WML event that called `dofile`, and register all of their events directly from the Lua file.

I don't think this is actually the best option. Any non-recurring events are better registered in WML so that they can serialize correctly. While it's possibly to do it in pure Lua, you'd need to track extra state to determine whether the event has fired yet.

stevecotton:
The CI says the CI runs had a green status, but below is an extract from the CI log. This PR is broken, and the CI script is broken (at the moment I'm testing whether reverting 99384b038c3eebc70595d5f1d052499239d88f80 restores the CI reporting).

```
Error (strict mode, strict_level = 1): wesnoth reported on channel error engine
20210415 05:59:49 error engine: failed to auto-store $unit at (-999,-999)
20210415 05:59:49 error engine: failed to auto-store $second_unit at (-999,-999)
20210415 05:59:49 error engine: failed to auto-store $unit at (-999,-999)
20210415 05:59:49 error engine: failed to auto-store $second_unit at (-999,-999)
FAIL TEST (BROKE STRICT): event_test_auto_variables_damage
Failure, Wesnoth returned 5 FAIL_BROKE_STRICT for event_test_auto_variables_weapons but we expected 0 PASS
Error (strict mode, strict_level = 1): wesnoth reported on channel error engine
20210415 05:59:51 error engine: failed to auto-store $unit at (-999,-999)
20210415 05:59:51 error engine: failed to auto-store $second_unit at (-999,-999)
20210415 05:59:51 error engine: failed to auto-store $unit at (-999,-999)
20210415 05:59:51 error engine: failed to auto-store $second_unit at (-999,-999)
FAIL TEST (BROKE STRICT): event_test_filter_attack
Failure, Wesnoth returned 5 FAIL_BROKE_STRICT for event_test_filter_unit but we expected 0 PASS
Error (strict mode, strict_level = 1): wesnoth reported on channel warning wml
20210415 06:00:11 warning wml: The following conditional test unexpectedly failed:
	[variable]
		equals=""
		name="unit.facing"
	[/variable]
	
	Interpolated to:
	[variable]
		equals=""
		name="unit.facing"
	[/variable]
	
	Note: The variable unit.facing currently has the value ne.
FAIL TEST: recruit_facing_enemy_one
Failure, Wesnoth returned 1 FAIL for test_force_chance_to_hit_macro but we expected 0 PASS
Error (strict mode, strict_level = 1): wesnoth reported on channel warning wml
20210415 06:00:12 warning wml: The following conditional test unexpectedly failed:
	[variable]
		equals=""
		name="unit.facing"
	[/variable]
	
	Interpolated to:
	[variable]
		equals=""
		name="unit.facing"
	[/variable]
	
	Note: The variable unit.facing currently has the value ne.
FAIL TEST: recruit_facing_enemy_two
Failure, Wesnoth returned 1 FAIL for heal but we expected 0 PASS
Error (strict mode, strict_level = 1): wesnoth reported on channel warning wml
20210415 06:00:14 warning wml: The following conditional test unexpectedly failed:
	[variable]
		equals=""
		name="unit.facing"
	[/variable]
	
	Interpolated to:
	[variable]
		equals=""
		name="unit.facing"
	[/variable]
	
	Note: The variable unit.facing currently has the value n.
FAIL TEST: recruit_facing_leader
Failure, Wesnoth returned 1 FAIL for heal but we expected 0 PASS
Error (strict mode, strict_level = 1): wesnoth reported on channel error scripting/lua
20210415 06:00:17 error scripting/lua: Lua Error: When executing, Lua runtime error: lua/core/map.lua:11: attempt to index a nil value (local 'x')
stack traceback:
	lua/core/map.lua:11: in field 'read_location'
	lua/core/interface.lua:71: in function <lua/core/interface.lua:69>
	(...tail calls...)
	[string "..."]:7: in local 'bytecode'
	lua/wml-tags.lua:268: in local 'cmd'
	[C]: in ?
Failure, Wesnoth returned 2 TIMEOUT for check_interrupts_return_nested but we expected 0 PASS
Error (strict mode, strict_level = 1): wesnoth reported on channel error scripting/lua
20210415 06:05:15 error scripting/lua: Lua Error: When executing, Lua runtime error: ~wml:[for] has a step of 0!
stack traceback:
	[C]: in function '.error'
	lua/core/wml.lua:186: in function 'wml.error'
	lua/wml-flow.lua:105: in local 'cmd'
	[C]: in ?
FAIL TEST (BROKE STRICT): forloop_all_zero
Failure, Wesnoth returned 5 FAIL_BROKE_STRICT for check_interrupts_case but we expected 0 PASS
Error (strict mode, strict_level = 1): wesnoth reported on channel error scripting/lua
20210415 06:05:16 error scripting/lua: Lua Error: When executing, Lua runtime error: ~wml:[for] has a step of 0!
stack traceback:
	[C]: in function '.error'
	lua/core/wml.lua:186: in function 'wml.error'
	lua/wml-flow.lua:105: in local 'cmd'
	[C]: in ?
FAIL TEST (BROKE STRICT): forloop_step_zero
Failure, Wesnoth returned 5 FAIL_BROKE_STRICT for heal but we expected 0 PASS
WML Unit Tests Result: 266 of 274 tests passed
    (266 passed, 8 failed)
```

CelticMinstrel:
I had actually noticed these test failures already and was starting to investigate them… what I _hadn't_ noticed is that the WML test runner is incorrectly reporting a zero exit status. It _could_ be caused by the linked commit; if you do choose to revert it, I think I have another idea of how to fix the bug it fixed.

CelticMinstrel:
This fixes _some_ of the test failures, but I'm very confused by two of the remaining ones - the two for-loop tests that are failing look like they're _supposed_ to fail? And yet, they somehow pass on master? How?

data/lua/core/unit_test.lua
```diff
@@ -0,0 +1,146 @@
...
+					n = n + 1
+				end
+			end
+		end
+        return tostring(val)
+    end
+
+    function unit_test.log(prefix, message)
+        std_print(prefix .. ': ' .. message)
+    end
```
Pentarctagon, line 49: Inconsistent indenting for log and tostring.

Pentarctagon: Actually just a lot of weird indentation in this file overall.

src/game_events/entity_location.cpp
```diff
@@ -89,18 +89,18 @@ bool entity_location::matches_unit(const unit_map::const_iterator & un_it) const
  * when this was constructed.
  */
 bool entity_location::matches_unit_filter(const unit_map::const_iterator & un_it,
-                                          const vconfig & filter) const
+										  const unit_filter& filter) const
```
Pentarctagon, line 92: Can this just be put on one line?

data/lua/core/unit_test.lua
CelticMinstrel: I've fixed this indentation in #5708

Pentarctagon:
For events written entirely in WML, those should continue to work as-is with no need to change anything (aka nothing has been removed or deprecated there), correct?

Otherwise, for deprecations there are:
* `wesnoth.set_menu_item` --> `wesnoth.interface.set_menu_item`, though it looks like the more exact replacement would be to call `wml_actions.set_menu_item` directly?
* `wesnoth.add_event_handler` --> `wesnoth.game_events.add`
* `wesnoth.remove_event_handler` --> `wesnoth.game_events.remove`
* `wesnoth.fire_event` --> `wesnoth.game_events.fire`, and also it looks like the 4th and 5th arguments have been combined?
* `wesnoth.fire_event_by_id` --> `wesnoth.game_events.fire_by_id`
* `on_event` --> `wesnoth.game_events.add`

And then newly added is the `unit_test` module and a bunch of new WFL values that can be used.

src/formula/callable_objects.hpp
```diff
@@ -25,10 +25,34 @@ class team;
 class terrain_type;
 class unit;
 class unit_type;
+namespace game_events {struct queued_event;}
 
 namespace wfl
 {
 
+// An object representing the state of the game, providing access to the map and basic information
```
Pentarctagon, line 33: This should be a doxygen comment.

src/formula/callable_objects.hpp
```diff
@@ -25,10 +25,34 @@ class team;
...
+class gamestate_callable : public formula_callable
+{
+public:
+	// TODO: Constructor
```
Pentarctagon, line 37: This TODO should be resolved before merging.

src/formula/callable_objects.hpp
```diff
@@ -25,10 +25,34 @@ class team;
...
+	// Currently it accesses all its state through the resources namespace, so nothing is passed into it.
+	
+	void get_inputs(formula_input_vector& inputs) const override;
+	variant get_value(const std::string& key) const override;
+};
+
+// An object representing the state of the current event; equivalent to Lua's wesnoth.current.event_context
```
Pentarctagon, line 44: Doxygen comment.

src/game_events/handlers.hpp
```diff
@@ -29,17 +29,33 @@
...
+	/** Returns true if it is possible to serialize the filter into a config. */
+	virtual bool can_serialize() const;
+	virtual ~event_filter() = default;
+	event_filter() = default;
+private:
+	event_filter(const event_filter&) = delete;
+	event_filter& operator=(const event_filter&) = delete;
+};
 
 class event_handler
```
Pentarctagon, line 52: Doxygen comments on things that have been newly added would be good.

src/game_events/manager.hpp
```diff
@@ -54,19 +56,22 @@ class manager
...
-	void read_scenario(const config& scenario_cfg);
+	void read_scenario(const config& scenario_cfg, game_lua_kernel& lk);
 	~manager();
 
-	/** Create an event handler. */
-	void add_event_handler(const config& handler, bool is_menu_item = false);
+	/** Create an event handler from an [event] tag. */
+	void add_event_handler_from_wml(const config& handler, game_lua_kernel& lk, bool is_menu_item = false);
+	/** Create an empty event handler. Expects the caller to finish setting up the event.
+	 */
```
Pentarctagon, line 65: This comment should all be on the same line.

src/game_events/manager_impl.hpp
```diff
@@ -20,8 +20,37 @@
...
+	pending_event_handler(const pending_event_handler&) = delete;
+	pending_event_handler& operator=(const pending_event_handler&) = delete;
+	// It's move-constructible, but there's no way to make it move-assignable since it contains a reference...
+	pending_event_handler& operator=(pending_event_handler&&) = delete;
+public:
+	pending_event_handler(event_handlers& list, handler_ptr handler)
+		: list_(list)
+		, handler_(handler)
+	{}
+	/// Check if this handler is valid.
```
Pentarctagon, line 45: Please use the `/** words */` style for single line comments.

CelticMinstrel:
> For events written entirely in WML, those should continue to work as-is with no need to change anything (aka nothing has been removed or deprecated there), correct?

Yes, ActionWML is still fully-functional and not deprecated. This system means we _could_ deprecate it in the future if we wished, but I don't think now's a good time to take such a drastic step. (Maybe there _is_ no good time, I'm not sure.)

> `wesnoth.set_menu_item` --> `wesnoth.interface.set_menu_item`, though it looks like the more exact replacement would be to call `wml_actions.set_menu_item` directly

Yes, that's correct; and what `wesnoth.wml_actions.set_menu_item` does is just calle `wesnoth.interface.set_menu_item` followed by `wesnoth.game_events.add`. Note that `wesnoth.set_menu_item` was actually undocumented (and listed as "should not use"), so I possibly could've not bothered with a deprecation message, but… it can't hurt, right?

> `wesnoth.fire_event` --> `wesnoth.game_events.fire`, and also it looks like the 4th and 5th arguments have been combined?

Correct, and the same for `fire_by_id` (combining the 4th and 5th arguments). The rationale for this combination is that splitting that into two arguments both obscured what actually happens and prevented access to some things that should actually work; with the new form, you can now perfectly fake an `attacker hits` or a village `capture` event, for example. The old form did not permit this.

> And then newly added is the `unit_test` module and a bunch of new WFL values that can be used.

Correct, though the `unit_test` module has been split out into #5708 .

src/formula/callable_objects.hpp
CelticMinstrel: I was kinda unsure what to do here since it's more of a WFL API than a C++ API, but I can make it doxygen if you prefer.

src/formula/callable_objects.hpp
CelticMinstrel: Is simply deleting it sufficient? I put it there thinking there must be some object I need to pass to it to encompass the game state, but then I ended up reading everything from the `resources` namespace instead.

src/formula/callable_objects.hpp
Pentarctagon: My thought is that even if it's implementing WFL, it's still C++ code that needs to be understood by anyone who needs to understand it later.

src/formula/callable_objects.hpp
Pentarctagon: That'd be fine.

stevecotton:
> > > Instead of using ActionWML, you can write the entire event inline as Lua code using [event]code=.
> >
> > What advantage does this feature give?
> 
> It doesn't directly give any advantage over using `[lua]` in ActionWML, but the new internals make it trivial to implement and if ActionWML were ever deprecated it would become the only way to write an event.

Those don't sound like advantages for WML authors. I think it has a two major disadvantages:
* it means that using Lua code inconsistently sometimes needs the `[lua]` tag and sometimes doesn't
* if the WML author uses `code=` somewhere in an `[event]` that also has WML, omitting the `[lua]` tag means the WML code is silently ignored. Even if we add warnings, it's still going to be another thing for WML authors to have to debug.

```
# Test whether WML code runs when followed by Lua in a [lua] tag, and whether a
# failing {ASSERT} in the WML unit test framework triggers a fail when the test
# finishes via the Lua unit test framework.
#
# The expected test result is a FAIL, triggered by the {ASSERT} line.
{GENERIC_UNIT_TEST "test_mixed_wml_lua_1" (
    [event]
        name = start
        {ASSERT ([not][true][/true][/not])}
        [lua]
            code=<<
                unit_test.pass_if(true, 'return true')
            >>
        [/lua]
    [/event]
)}

# This looks similar to test_mixed_wml_lua_1, but misses the [lua] tag.
# 
# This is now a Lua event - the test result is PASS, because the {ASSERT} line
# doesn't get run.
{GENERIC_UNIT_TEST "test_mixed_wml_lua_2" (
    [event]
        name = start
        {ASSERT ([not][true][/true][/not])}
        code=<<
            unit_test.pass_if(true, 'return true')
        >>
    [/event]
)}
```

stevecotton:
> Hmm… you do seem to have a point there… I wonder if using a different attribute would help? Like [event]action= instead of code? That's also more consistent with the Lua API here.

What advantage would the ability to write `[event]action=` give, and who would gain from that?

My main complaint about `code=` without the `[lua]` tag isn't what it's called, it's that from my point of view it adds problems without giving any advantages, and if it merges before the 1.16 branch splits then we'll have to support it until at least 1.19.

CelticMinstrel:
Like I said, it doesn't give any direct advantage over `[lua]code=` from an add-on developer's perspective; it's just an obvious outcome of this refactoring. If you can assign a Lua function as an event handler, assigning Lua code is basically the same thing.

It _does_ give an advantage if the ultimate end goal is to deprecate ActionWML though, because it's a way of writing an event that doesn't touch ActionWML event slightly (a `[lua]` tag is still ActionWML). I'm not sure if deprecating ActionWML is even something to strive for, but WML as a coding language never has been great, and there are even a few outstanding problems that are really hard to solve in WML (such as plural strings).

It would be possible to disable the feature right now and only enable it if and when ActionWML is actually deprecated; all that means is commenting out the line that checks if the attribute exists. It seems kind of pointless to do that, though.

Or… would you prefer that I check the `[event]` tag's children and create the equivalent of the current `[event]code=` if and only if it has exactly one child that's a `[lua]` tag? That creates problems of its own, though, because the event handler function needs to passed some arguments. If it's an ActionWML event, the arguments are obviously the contents of the `[event]` tag. However, if it were a code event defined by `[event][lua]code=`, the arguments would have to be the `[args]` subtag of `[lua]` (to avoid breaking existing events), which isn't hard to do but seems like an inconsistency in the implementation.

> My main complaint about `code=` without the `[lua]` tag isn't what it's called, it's that from my point of view it adds problems without giving any advantages

At least, the second disadvantage you listed is rendered far less severe by changing the name, as it's no longer an issue of accidentally forgetting an enclosing tag and breaking your ActionWML event. The first issue you listed is… possibly just complaining that there would exist a way to use Lua code that doesn't involve a `[lua]` tag? It's a pretty minor problem as long as the attribute is renamed.

Do you see additional problems that the feature creates as well?

Pentarctagon:
Given that WML ignores unknown attributes, the current functionality would be the reverse of what would happen with this PR - forgetting the put [lua] tags around the `code=<<>>` attribute would result in the lua code silently doing nothing, correct?

CelticMinstrel:
If you're saying that currently forgetting the enclosing `[lua]` tag means the code does nothing… then yes, that's what happens currently on master.

Pentarctagon:
So the situation then is that for the invalid code:
```
[event]
name=start
code = <<...lua code...>>
[some_wml]
...
[/some_wml]
[/event]
```

One current master, the lua code is silently ignored and the WML is executed; after this change, the lua code is executed and a WML error is displayed in-game?

CelticMinstrel:
There's no error shown either before or after this change, but apart from that, you are correct. It wouldn't make sense to show a WML error in the general case, although I suppose I could check if the code contains `...` and show an error if it doesn't and there are unknown things in the `[event]`.

Pentarctagon:
It doesn't sound like the situation of something getting ignored is made any worse by this then, at least.

stevecotton:
It switches from (the bit being ignored is where the error is) to (the bit being ignored is where the error isn't).

CelticMinstrel:
I'm not quite sure I understand what you're saying… it sort of makes sense and it sort of doesn't…

As a side note: if `code=` (or `action=` if we prefer) exists but isn't valid Lua code, you're still going to get an error, although it might not be the _most_ comprehensible possible error. That can be improved, though.

CelticMinstrel:
There's still some failing tests that I need to figure out, but I think I've addressed all the comments so far. Anything else I need to do?

src/scripting/game_lua_kernel.hpp
CelticMinstrel: The `resources_ptr_` member is a RAII construct that assigns the Lua kernel pointer to the `resources::lua_kernel` on destruction and nulls out the `resources::lua_kernel` Lua kernel pointer on destruction.

We don't actually _need_ to store the old value of `resources::lua_kernel` since it should always be `nullptr`, but doing it this way allows the `guard_value` class to be useful for other kinds of situations.

CelticMinstrel:
Comments on [the API documentation](https://wiki.wesnoth.org/LuaAPI/wesnoth/game_events) are also welcome.

src/scripting/game_lua_kernel.hpp
stevecotton: This doesn't look anything like RAII. It would be a little bit like RAII if `guard_value`'s constructor were to throw when the original value wasn't nullptr, but there's nothing like RAII here at the moment.

My previous comment from April 15th still awaits an answer.

CelticMinstrel: That comment doesn't make any sense whatsoever. The `guard_value` class is an RAII class for a scoped temporary change to a variable. It would be downright nonsensical for it to throw based on the prior content of the variable.

I definitely answered your first question. For the second one… putting the `guard_value` in the Lua kernel means that `resources::lua_kernel` is non-null; if and only if the Lua kernel exists. If the `guard_value` was in the kernel's owner, that guarantee would be broken.

data/test/scenarios/test_lua_name.cfg
```diff
@@ -0,0 +1,45 @@
...
+            local ret, err = pcall(function()
+                local some_variable = bar
+            end)
+            unit_test.assert_contains(err, 'test', 'error contains test')
+            unit_test.succeed()
+        >>
+    [/event]
+)}
+
+{GENERIC_UNIT_TEST test_lua_name (
```
stevecotton, line 22: Tests should have documentation about what they're testing, otherwise it makes it hard to review whether the test is correct. This test is going to be broken by later changes in this PR, but let's get to that later.

Having asked again for advantages of `[event]action=` over `[event][lua]code=`, I got the response in IRC
> It does allow you to make a "safe" event that catches and handles errors, seems possibly helpful in unit tests at least.

which implied that this feature could only be done with `[event]action=`. It can also be done with `[event][lua]code=`, noting that the already-present `[lua]` tag needs to go inside the new `[lua]` tag's `[args]`.

On to why this test is going to break before this PR is complete: the `[lua]` tag that's already present looks like ActionWML. While he's been more open to `[event]action=` that I, @Pentarctagon has said that it needs to generate a warning in that case.

CelticMinstrel: The use in this test is not one that generates a warning. A warning is generated if both of the following conditions are satisfied:

* The Lua code does not contain the string `...`
* There are any tags in the event (excluding filter tags)

In this case, the Lua code _does_ contain `...`, so there's no warning. The warning is already implemented, and there is even a test that ensures it triggers. We could perhaps improve on this even further (it would be nice if `...` in a string or comment didn't count) but that might be a bit tricky.

> It can also be done with `[event][lua]code=`, noting that the already-present `[lua]` tag needs to go inside the new `[lua]` tag's `[args]`.

Ah yes, you're right. In fact, there's at least one test that already does this, possibly one of the for-loop tests? I don't quite remember which one it is, but it does exist.

src/scripting/game_lua_kernel.hpp
stevecotton: What is the resource being acquired? If I understand you, the resource is the memory location; however in that case it being non-nullptr means that the resource is already being used - therefore if you're saying this is RAII, it needs to react because the resource is already taken.

If that's not it, then please identify the "resource" in this class' design.

CelticMinstrel: > however in that case it being non-nullptr means that the resource is already being used

That doesn't make any sense. It's not even necessarily a pointer. It could be an integer, or a string, or literally anything.

I've already explained its purpose. It's for when you need to _temporarily_ change the value of a variable and want to ensure that the old value gets restored regardless of what exit path is taken. In a sense, it's a special case of the ON_SCOPE_EXIT utility.

I'm not sure there's a "resource" to identify. It could be the memory location, but we're not acquiring exclusive access to it or anything. Or maybe it could be the temporary value stored to that memory location.

stevecotton: > I'm not sure there's a "resource" to identify.

You said it's RAII, Resource Acquisition Is Initialisation. The comments of mine that you're saying don't make sense are based on trying to fit it to RAII, because you said it was.

What doesn't make sense is the design of `guard_value` itself.

Above was about the class, below is about the caller.

The way it's being used in the specific case (and in the similar one in the other PR) is to wrap a raw pointer, a non-owning global which is relying on some mechanism to ensure it doesn't become dangling. That's a messy construct which should be redesigned, but putting a temporary-value wrapper around it works in the cases that it already works in, and is still going to break in the cases that it already breaks in. It has a edge-case that it improves (if something else in game_lua_kernel's constructor throws), and an edge-case that it makes worse (if two game_lua_kernels get constructed). On top of that, the class isn't a good fit for the edge-case that it does improve.

CelticMinstrel: > That's a messy construct which should be redesigned

Sure, but deprecating the `resources` namespace is extremely out of scope for either PR.

> putting a temporary-value wrapper around it works in the cases that it already works in, and is still going to break in the cases that it already breaks in.

I don't know what you're talking about. By using the `guard_value`, it still works in all the cases that it already worked in, and also works in some cases that it previously broke in. Can you point to _any_ case that would actually break it now?

>  an edge-case that it makes worse (if two game_lua_kernels get constructed)

The only way this could happen is in code of the form:

```cpp
auto lk1 = new game_lua_kernel(...);
auto lk2 = new game_lua_kernel(...);
delete lk1;
delete lk2
```

And I don't think we have a need to support that kind of use-case. That would also break with the old version of this code, so on the whole the `guard_value` improves the situation: it doesn't break anything new, and it makes some things work that previously didn't.

The only thing I'd maybe consider doing differently is making the `guard_value` be a sibling of the `game_lua_kernel` instead of a child. I don't think it makes much difference in practice, however, unless the constructor called something that accessed the kernel via the `resources` namespace.

> On top of that, the class isn't a good fit for the edge-case that it does improve.

I don't agree, but if you have alternative suggestions, I'm listening.

stevecotton: Are you now agreeing with me that it's not RAII?

Just leave the code where it was in play_controller.cpp. Leave it unless and until you want to do the proper cleanup in the `resources` namespace.

CelticMinstrel: Reverting that change is not an option. The whole reason I did it was to fix a crash, which it did successfully.

stevecotton: What are the steps to reproduce the crash, so that I can look at a better way to fix it?

If you want to call this RAII, identify what the R is, otherwise it doesn't make any sense.

CelticMinstrel: I don't remember exactly, but I think it was something pretty much impossible to miss, like when you start the scenario, or when an event triggers, or maybe when the scenario ends. I doubt there is a better way to fix it, other than eliminating the use of the `resources` namespace in the offending location.

stevecotton: Can you remember any more details about the bug?

I tried reverting that bit of the code, and successfully played my currently-in-progress HttT game from a mid-scenario save through to the second turn of the next scenario. So it doesn't seem to be an impossible-to-miss bug, and I'm interested in whether it indicates a bug elsewhere in this PR.

CelticMinstrel: Possibly it only occurs if the scenario adds an event using `wesnoth.game_events.add{}` with at least one of `filter=` or `action=` set to a Lua function? I had added some quick events in the tutorial that do things like that.

src/scripting/game_lua_kernel.cpp
```diff
@@ -3248,16 +3246,135 @@ int game_lua_kernel::intf_log_replay(lua_State* L)
 	return 0;
 }
 
-/** Adding new events */
+struct lua_event_filter : public game_events::event_filter
```
stevecotton, line 3249: The `lua_event_filter` and `filter_lua` classes seem to be almost the same, however they take different approaches to accessing the lua_kernel. This one stores a reference, the other one accesses it via `resources::`.

I'm not saying which of those is better, but I'm interested in why they're inconsistent.

CelticMinstrel: They basically are the same… I guess the only reason it exists twice is because they are file-local classes.

One is used for filter code placed in an `[event]` tag, while the other is used when you add an event from Lua and pass a function as the filter.

src/scripting/game_lua_kernel.hpp
stevecotton: I wrote a long comment, and think I then clicked on "Resolve" instead of "Comment" afterwards. If you've seen this comment before, that's why it's being reposted; and it's why this got resolved and then unresolved.

The `event_test_filter_lua_serializable` unit test was able to reproduce the bug, and the call-stack of the crash is:
```
#0  in game_lua_kernel::clear_wml_event(int) (with this==nullptr)
#1  in game_events::filter_lua::~filter_lua()
#13 in std::vector<std::shared_ptr<game_events::event_filter>, std::allocator<std::shared_ptr<game_events::event_filter> > >::~vector()
#14 in game_events::event_handler::~event_handler()
#29 in game_events::event_handlers::~event_handlers()
#32 in game_events::manager::~manager()
#35 in game_state::~game_state()
#39 in play_controller::~play_controller()
```

With the old code, play_controller's destructor set the `resources::` pointers to nullptr early in the destruction. The game_events::manager and lua kernel are both owned by the game_state class, and what the guard_value change did was to move the `resources::lua_kernel = nullptr` slightly later in the destruction sequence. The old way that all the `resources::` pointers were cleared early in the destruction sequence seems to be a design choice, that from that point onwards the components would be more isolated.

Changing to having the components call each other during the destruction seems a design change. I don't know whether it's a large or a small change, but I think finding out enough to be sure about it has to be part of this PR. Redesigning the `resources::` pointers might not need to be in the scope of this PR, but having enough knowledge about the mechanism to be sure about it has to be in here. I'm worried that this PR is putting in a new design that doesn't mesh with the old one, without a clear description of what the old design was or how they interact.

Can we be sure that none of the other `resources::` pointers are accessed during the destruction? I'm not sure that we can even be certain that clear_wml_event() won't cause them to be accessed - it removes a reference count, but the Lua object might have a finalizer. For example, src/scripting/lua_terrainfilter.cpp has a finalizer (although it seems to be a no-op), and a Lua based filter that accesses another resource in its finalizer seems plausible, maybe to clear a cache.

src/scripting/game_lua_kernel.cpp
stevecotton: This links into the comment that I clicked "resolve" on instead of posting - storing a reference vs making the `resources::lua_kernel` non-nullptr are two different ways to access the kernel during play_controller's destruction sequence; both of them seem to be changing from the existing design.

src/scripting/game_lua_kernel.hpp
Pentarctagon: Is the use of `guard_value` something that can be made into a separate PR, or does the rest of this PR depend on it?  If this is the last thing or nearly the last thing that's remaining to be resolved, then it'd probably to be better to get everything else in and then look at this on its own.

Pentarctagon:
So what's the status of this now?

CelticMinstrel:
I was contemplating a few changes based on Iris's feedback, and I think there's also an issue with error reporting that I still need to resolve.

stevecotton:
The questions from https://github.com/wesnoth/wesnoth/pull/5663#discussion_r628692199 are still unresolved; does anyone understand the existing design sufficiently to push this into the API-freeze release?

CelticMinstrel:
Looks like the CI failures are just whitespace, so I'll incorporate that into my next force-push.

data/multiplayer/scenarios/2p_Isle_of_Mists.lua
```diff
@@ -1,6 +1,6 @@
 local _ = wesnoth.textdomain 'wesnoth-multiplayer'
 local T = wml.tag
-local on_event = wesnoth.require("on_event")
+local on_event = wesnoth.game_events.on_repeating
```
soliton-, line 3: This was surely meant to be `add_repeating`.

CelticMinstrel:
So, 3c9e18b removes the guard_value thing that @stevecotton found controversial for some reason (since it also removes the feature that made it necessary). Therefore, I think this can now be merged assuming CI passes.

data/test/scenarios/events-test_actions.cfg
```diff
@@ -0,0 +1,89 @@
...
+# This is kinda a duplicate of a basic unit test system sanity test, but... it seems to fit here as well...
+{GENERIC_UNIT_TEST "event_test_action_wml" (
+    [event]
+        name=start
+        {SUCCEED}
+    [/event]
+)}
+
+# Test that dynamically adding an event from Lua works
+# There's two different ways of adding events, so test both
```
stevecotton, line 11: I acknowledge these were written before @Pentarctagon's new style for test comments, but please update the test files to the  `API(s) being tested:` style.

A `# wmllint: no translatables` would be good too.

src/game_events/handlers.hpp
stevecotton: The documentation still seems to be missing.

src/game_events/manager.cpp
```diff
@@ -190,24 +227,8 @@ void manager::execute_on_events(const std::string& event_id, manager::event_func
...
-
-					matches = standardized_event_id ==
-						event_handlers::standardize_name(utils::interpolate_variables_into_string(name, *gd));
-				}
-				else {
-					matches = standardized_event_id == name;
-				}
-
-				if (matches) {
+			for(const std::string& name : handler->names(gd)) {
```
stevecotton, line 230: The old code checked for `gd != nullptr`, the new code looks as if it will unconditionally try to dereference. What guards against that?

src/game_events/manager.hpp
```diff
@@ -55,21 +57,23 @@ class manager
...
+	pending_event_handler add_event_handler_from_lua(const std::string& name, const std::string& id, bool repeat = false, bool is_menu_item = false);
 
 	/** Removes an event handler. */
 	void remove_event_handler(const std::string& id);
 
 	/** Gets an event handler by ID */
 	const handler_ptr get_event_handler_by_id(const std::string& id);
 
-	void add_events(const config::const_child_itors& cfgs, const std::string& type = std::string());
+	void add_events(const config::const_child_itors& cfgs, game_lua_kernel& lk, const std::string& type = std::string());
```
stevecotton, line 74: Although `type` isn't new, please add a comment saying what it is.

src/game_events/manager.hpp
```diff
@@ -55,21 +57,23 @@ class manager
...
 
-	void write_events(config& cfg) const;
+	void write_events(config& cfg, bool include_nonserializable=false) const;
```
stevecotton, line 76: Needs a comment to explain what serializing a non-serializable thing will do.

src/game_events/manager_impl.cpp
```diff
@@ -94,70 +93,51 @@ handler_list& event_handlers::get(const std::string& name)
...
 
 		if(find_it != id_map_.end() && !find_it->second.expired()) {
 			DBG_EH << "ignoring event handler for name='" << name << "' with id '" << id << "'\n";
-			return;
+			return {*this, nullptr};
 		}
 	}
 
 	if(name.empty() && id.empty()) {
 		lg::log_to_chat() << "[event] is missing name or id field\n";
```
stevecotton, line 109: Why remove the debugging here?

This looks like an incorrectly resolved merge-conflict with #6619's bfd0ef9af7aace4d753521fb29e9ad7b9c9419c7.

CelticMinstrel: If this is about the code that dumps the entire content of the event, it's not removed. The event content is no longer available in this function, so that debug message was moved into the caller.

src/game_events/manager.cpp
CelticMinstrel: "some places" is an extremely unhelpful comment.

src/game_events/manager.hpp
CelticMinstrel: I don't really think that belongs here, but… it's the unit type that contains the event, if any.

CelticMinstrel:
You can ignore 96af39c, I was planning to merge that separately after this PR is finished.

src/game_events/manager_impl.cpp
stevecotton: I don't see where it moved to, and the logging seems to be missing when retesting #6619.

Steps:
1. In EI S01, add `{EXTRA_SCENARIO_MUSIC nunc_dimittis.ogg}` immediately after the opening `[scenario]` tag.
2. Run Wesnoth with `--log-debug=wml`, and start the scenario.
3. The "[event] is missing name or id field" is shown in chat
4. No additional log is printed, neither on the Lua console nor the command-line console.

src/game_events/manager.cpp
stevecotton: Where those "some places" are depends on how the unconditional dereference is fixed. Either make `handler->names` handle a potential nullptr, or make it take a reference instead.

src/game_events/manager.hpp
stevecotton: There's no hint in the current docs that this expects a unit type, so it definitely needs documentation.

CelticMinstrel: I don't disagree. It just has nothing whatsoever to do with this PR.

src/game_events/manager.cpp
CelticMinstrel: Regardless, I addressed this in 28d555c.

src/game_events/manager_impl.cpp
CelticMinstrel: The message is there and works as expected, with one minor caveat. It moved to the `event_handler` logdomain. Is that good enough? Should I change the logdomain for that one message?

stevecotton: It needs to be useful for debugging - in the old code it had a message printed to the console at log level error or warning, with additional information printed immediately after that if used with `--log-debug=wml`.

With the new code, using `--log-warning=event_handler` doesn't print anything. Using `--log-debug=event_handler` prints 673 lines, with these looking very in-place among them; this doesn't even look like an error message:
```
Checking lua scripts... ok
20220704 00:55:38 debug event_handler: Content of that event:
[music]
	append = yes
	name = nunc_dimittis.ogg
[/music]

20220704 00:55:38 debug event_handler: Registered WML event 'prestart' (first time only) with the following actions:
```

src/game_events/manager.hpp
stevecotton: So, was the check for `utils::might_contain_variables(name)` just dead code? I mean, there was a code path specifically for events that occurred when `gd` was `nullptr`, including a specific path which would allow ones that didn't need interpolation to still happen.

src/game_events/manager.cpp
stevecotton: So, was the check for `utils::might_contain_variables(name)` just dead code? I mean, there was a code path specifically for events that occurred when `gd` was `nullptr`, including a specific path which would allow ones that didn't need interpolation to still happen.

src/scripting/game_lua_kernel.hpp
stevecotton: Resolved because the `guard_value` was removed.

src/game_events/manager.cpp
CelticMinstrel: ???

What are you even talking about? That check is still there in `handler->names`, protected by a null check on the passed-in `variable_set`.

stevecotton: I missed the check inside `handler->names`, there wasn't an unconditional dereference, sorry about that part.

28d555c appears to be incorrect, and it's 28d555c that caused the "dead code" question. It means we can't reach the `utils::might_contain_variables(name)` line with `gd` being `nullptr`, which means the moved-but-still-present conditional block for `gd == nullptr && !utils::might_contain_variables(single_name)` is dead code. The question about whether that conditional block is actually useful remains.

Also, in the new code (but not the old) the check for `utils::might_contain_variables(single_name)` is being done after the variables have been interpolated.

src/game_events/manager.hpp
stevecotton: Thanks for adding the comment, but please make it Doxygen-style (as in, a Doxygen doc for the function).

src/game_events/manager.cpp
CelticMinstrel: > Also, in the new code (but not the old) the check for `utils::might_contain_variables(single_name)` is being done after the variables have been interpolated.

???

It's literally "if might contain variables then interpolate variables"...

CelticMinstrel: Okay, it took me a bit to figure out what you were saying, but… basically, the first check is wrong. It should split, _then_ interpolate, right?

src/game_events/manager_impl.cpp
CelticMinstrel: Okay, so the correct fix is:

1. Move it to the WML logdomain?
2. Print the error to the console in the event_handler logdomain?
3. Both, in the WML logdomain?

stevecotton: > Okay, so the correct fix is: ... ?

I know that I push for "more documentation" often, but in this case I'd say the correctness of the fix should be judged by writing a reply to https://r.wesnoth.org/p672974 that starts "From 1.17.6, use ...". Either of these sound good:

- use `--log-debug=wml`, just as before.
- use `--log-info=event_handler` instead.

Given the amount of output from `--log-debug=event_handler`, I think the current message has to either move to the `wml` logdomain or be promoted to `info` severity in the `event_handler` logdomain. In either case, print the one-line error to the console in the same logdomain too.

src/game_events/handlers.hpp
stevecotton: Although docs were added to `event_filter`, they're still missing from the same file's `event_handler`. While that existed before, most of its methods are new additions.

src/game_events/manager.cpp
stevecotton: I'm assuming that's been done, at least it looks done from a quick check.

src/game_events/manager.hpp
stevecotton: Still unresolved. On the same subject, please document the situation for `event_handler::write_config` too.

src/game_events/manager_impl.cpp
stevecotton: Thanks for resolving.

src/scripting/game_lua_kernel.cpp
```diff
@@ -622,12 +625,16 @@ int game_lua_kernel::intf_fire_event(lua_State *L, const bool by_id)
 		}
 	}
 
-	if (!lua_isnoneornil(L, pos)) {
-		data.add_child("first", luaW_checkconfig(L, pos));
+	luaW_toconfig(L, pos, data);
+
+	// Support WML names for some common data
+	if(data.has_child("primary_attack")) {
```
stevecotton, line 631: This seems to be missing a unit test.

Pentarctagon:
It looks like all that remains from the review comments are some documentation and unit tests?

CelticMinstrel:
Is that what's missing? Which documentation and which unit tests? Are you expecting me to write an entire new unit test suite for master or what?

There was also some talk about which commits should be squashed with what. I'm not sure if there was ever a conclusion on that.

src/scripting/game_lua_kernel.cpp
stevecotton: Github doesn't send any notification when an emoji is added as a response.

What's the confusion? AFAIK all the existing unit tests were passing before the code for handling the `primary_attack` key was added, and they're still all passing afterwards, which suggests there's no unit test for it.

Pentarctagon:
I just went and looked through the comments that weren't marked as resolved.

CelticMinstrel:
This comment thread is not marked resolved but presumably should be, right?

https://github.com/wesnoth/wesnoth/pull/5663#discussion_r628692684

CelticMinstrel:
Also, what about the two "remove X" commits? Did you want me to squash them away? I left them as remove commits mainly because I thought we might later want to revert them to get the feature back, but (assuming they do squash cleanly) I could always push those reverts to a separate branch to preserve the code.

Pentarctagon:
I'd say squash them away.
* "Remove [event]code=" looks simple enough to re-implement if needed even without the commit.
* "Remove [event]filter=" has an equivalent way of doing the same thing already.

CelticMinstrel:
And… what about "Fix WML errors showing as Lua errors"? There was some suggestion that it could be squashed into the original commit, though that would create a few conflicts to resolve (but not tricky ones).

Pentarctagon:
If it creates conflicts, then I'd say leave it.

src/scripting/game_lua_kernel.cpp
stevecotton: Resolving because the `filter_lua` class no longer exists.

src/scripting/game_lua_kernel.cpp
stevecotton: Is there another way to test it? I think it's not much more effort than cut&pasting an existing test.

The code being commented on seems to disagree with the emmylua docs about what `first` and `second` are - here they're attacks, but in the emmylua they're units (or locations of units).

CelticMinstrel: Okay, I can understand how that's confusing.

The `first` and `second` in the EmmyLua docs are indeed units. The `first` and `second` here are tags of the `data` argument.

In C++, an event handler is passed three pieces of information – two locations, and a config containing arbitrary additional data. The locations are the units. For attack events, the weapons are passed as `[first]` and `[second]` child tags in the config.

So, what code here does is that if the WML table contains a `[primary_attack]` tag, it converts it to the `[first]` tag that the event handler internally expects. Similarly, `[secondary_attack]` is converted to `[second]`. The choice of tags here is based on the tags used in the `[fire_event]` tag, for consistency; the `[first]` and `[second]` tags can be thought of as an internal detail.

src/game_events/handlers.cpp
```diff
@@ -62,15 +104,263 @@ void event_handler::handle_event(const queued_event& event_info, game_lua_kernel
...
+		return std::make_unique<filter_attack>(contents, true);
+	} else if(key == "filter_second") {
+		return std::make_unique<filter_unit>(contents, false);
+	} else if(key == "filter_second_attack") {
+		return std::make_unique<filter_attack>(contents, false);
+	}
+	return nullptr;
+}
+
+struct filter_dynamic : public event_filter {
```
stevecotton, line 308: Please document `filter_dynamic`, what it does isn't obvious from the name.

From the Wiki docs for `[insert_tag]` I'd expect the value of the variable to be evaluated once, at the time that the event is created. If I'm understanding this code correctly, then it gets reevaluated each time the filter runs - I can see usecases that would enable, but it would break WML that constructed the filter with something unit-specific during a `first_time_only=no` event, `[insert_tag]`'d it, and then overwrote it when that `first_time_only=no` event was triggered again by a different unit.

src/game_events/handlers.hpp
```diff
@@ -66,17 +82,61 @@ class event_handler
...
+	void register_wml_event(game_lua_kernel& lk);
+	void set_event_ref(int idx, bool has_preloaded);
+
 private:
 	bool first_time_only_;
 	bool is_menu_item_;
 	bool disabled_;
-	config cfg_;
-	std::vector<std::string> types_;
+	bool is_lua_;
```
stevecotton, line 134: Documentation of these new variables would be good.

Particularly, what does `has_preloaded_` mean? AFAICS it seems to be "whether the lua kernel had already finished preloading when it last called `set_event_ref`", and I'm wondering why that previous state matters.

src/game_events/handlers.cpp
CelticMinstrel: That's not how events are evaluated in the engine. The content of an `[event]` tag is divided into two distinct parts: the event definition (including name, ID, and filters), and the ActionWML. Variables in the event definition are substituted when the event triggers, _not_ when the event handler is added.

Your example of something this would break would already break in 1.16.x. Prior to the commit that introduced this class, there was a unit test failure because such dynamic filters no longer worked; this fixes that unit test failure.

src/game_events/handlers.hpp
CelticMinstrel: It matters because events registered after preload need to be serializable in order for saved games to work, but events registered _in_ preload do _not_ need to be serializable, because the preload will re-register them when you load a saved game.

Therefore, we only issue a warning about non-serializable events if they're registered after preload.

data/lua/wml-tags.lua
```diff
@@ -727,15 +730,17 @@ function wml_actions.event(cfg)
 		wesnoth.deprecated_message("[event]remove=yes", 2, "1.17.0", "Use [remove_event] instead of [event]remove=yes")
 		wml_actions.remove_event(cfg)
 	else
-		wesnoth.add_event_handler(cfg)
+		local delay = cfg.delayed_variable_substitution
+		if delay == nil then delay = true end
+		wesnoth.game_events.add_wml(delay, cfg)
```
stevecotton, line 735: Given that `cfg.delayed_variable_substitution` is also handled in the C++, is there a reason to not use just `wesnoth.game_events.add_wml(cfg)`?

I'd prefer to have the C++ only take a fixed number of arguments if possible - I think that was discussed in Discord for a different PR (after this code was already in the current PR).

CelticMinstrel: There isn't any reason. It was just slightly simpler because that's what the previous API (where everything was `wesnoth.game_events.add`) did.

I disagree with the general approach of only taking a fixed number of arguments, but I don't mind in this specific case; I can change it.

data/test/scenarios/events-test_filters.cfg
```diff
@@ -230,3 +230,119 @@
...
+        [end_turn][/end_turn]
+    [/event]
+)}
+
+{GENERIC_UNIT_TEST event_test_filter_lua_dynamic (
+    [event]
+        name=preload
+        [lua]
+            code=<<
+                wml.variables.triggers = 0
```
stevecotton, line 309: Should also test that an event with `first_time_only=true` only triggers once. This correctly succeeds, so the code under test works:

```
diff --git a/data/test/scenarios/events-test_actions.cfg b/data/test/scenarios/events-test_actions.cfg
index 75448bad0f2..1071f2b4493 100644
--- a/data/test/scenarios/events-test_actions.cfg
+++ b/data/test/scenarios/events-test_actions.cfg
@@ -43,12 +43,20 @@
         name=preload
         [lua]
             code=<<
-                wml.variables.triggers = 0
+                wml.variables.only_once = 0
+                wml.variables.should_repeat = 0
+                wesnoth.game_events.add{
+                    name = 'new turn',
+                    first_time_only = true,
+                    action = function()
+                        wml.variables.only_once = wml.variables.only_once + 1
+                    end
+                }
                 wesnoth.game_events.add{
                     name = 'new turn',
                     first_time_only = false,
                     action = function()
-                        wml.variables.triggers = wml.variables.triggers + 1
+                        wml.variables.should_repeat = wml.variables.should_repeat + 1
                     end
                 }
             >>
@@ -68,7 +76,9 @@
     [/event]
     [event]
         name=side 2 turn 2
-        {RETURN ({VARIABLE_CONDITIONAL triggers equals 2})}
+        {ASSERT ({VARIABLE_CONDITIONAL only_once equals 1})}
+        {ASSERT ({VARIABLE_CONDITIONAL should_repeat equals 2})}
+        {SUCCEED}
     [/event]
 )}
```

src/scripting/game_lua_kernel.cpp
stevecotton: Ah, ok.

Would it be sensible to remove any `data["first"]` and `data["second"]` tags added by a confused WML or Lua author?

data/test/scenarios/events-test_filters.cfg
CelticMinstrel: I'm a little bit confused. Is this a diff that I should just apply? Or is the idea that this diff produces an additional test, so we have the version without the diff applied _and_ the version with the diff applied?

stevecotton: I meant it as something that could be just applied. Splitting them into separate tests would be fine too.

src/game_events/handlers.cpp
stevecotton: Still no documentation for this class in the .cpp/.hpp files.

The test was added in #6785, will comment about the Wiki there / in IRC.

src/game_events/handlers.hpp
stevecotton: I think the name should be clearer about it recording the state at creation time, maybe `created_in_preload_`, but anyway this is resolved.

stevecotton (approved):
The unresolved comment requests documentation for `filter_dynamic`, however that might be replaced by a feature-request to change the behavior for `[event][insert_tag]name=filter`.

PR #5666, opened 2021-04-04
Update the animals Micro-AI demo with boars and piglets

stevecotton:
This adjusts to the change in e7b5996e988747c2b590852b4b461abe48857526.

The micro ai itself is unchanged, it's still controlled by `tusker_type` and
`tusklet_type`, it simply needed the new unit ids to be used in this scenario.

The demo can be run with `wesnoth -t animals`.


doofus-01 (approved):
Seems to work.  The reference to tuskers in the description is fine as that hints at how to use the micro-ai.  Once the sprites are updated, it might be a little less intuitive for some of us, but I'm not sure how much of an issue that really is.

mattsc (approved):
Since it does not change the MAI itself, this is fine as far as I am concerned.

PR #5671, opened 2021-04-05
Various fauna-related things

doofus-01:
For now, just rock scorpion pincer attack animation


CelticMinstrel:
What exactly is the problem with the old sprites? I'm not convinced the new ones are an improvement… and are you replacing all the tuskers or only the piglet?

Hejnewar:
I did tweak water spray to look more like poison spray. (I remember similar one from some UMC but couldn't find it.) I think it could be better for Rock Scorpion than water spray. 

poisonspray

doofus-01:
> I did tweak water spray to look more like poison spray. (I remember similar one from some UMC but couldn't find it.) I think it could be better for Rock Scorpion than water spray.

A separate poison spray image could be nice to have, but not strictly necessary for now.  If we do use one, it would need to be more differentiated.  Otherwise, I believe the IPF method works.

doofus-01:
I'll merge this soon.  Missed 1.15.12, maybe that's good, we'll see.

PR #5672, opened 2021-04-05
Make the debug clock look right

stevecotton:
This is the only widget in mainline using GUI2's `[draw][circle]` tag. I'm
currently working on a change to the C++ `text_shape` class, and want to
be able to regression-test the placement of `[circle]`s after that change.

I'd be okay with deprecating both `[circle]` and `[round_rectangle]` as neither
seems to be used elsewhere, but this close to feature-freeze I'd prefer to keep
them in 1.16 rather than suggest removing them now.

To see the widget, start the game with `wesnoth --clock`, and then click the
button underneath "Quit" on the main menu.

Before and after:
wesnoth_debug_clock_alignment

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -17,6 +17,7 @@
    * Added logs folder path to the Game Version dialog on Windows.
    * Made Accelerated Speed toggle announcement not stackable over itself or other announcements.
    * Minor adjustments to the sidebar's minimap area in the default game theme to reclaim empty and uneven space back. Note that this WILL break custom WML themes that use the same sidebar graphics as the default theme.
+   * Minor adjustment to the placement of numerals on the debug clock.
  ### WML Engine
  ### Miscellaneous and Bug Fixes
    * Fixed units with max movement set to zero being given one max movement point by `[unstore_unit]` or when loading a saved game
--- a/data/gui/window/debug_clock.cfg
+++ b/data/gui/window/debug_clock.cfg
@@ -7,8 +7,8 @@
 
 #define _GUI_TICK_MARK ANGLE TEXT
 [text]
-	x = "(round(125 + (100.0 * sin({ANGLE}))))"
-	y = "(round(125 - (100.0 * cos({ANGLE}))))"
+	x = "(round(125 + (100.0 * sin({ANGLE})) - (text_width / 2)))"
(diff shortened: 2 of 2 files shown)


CelticMinstrel:
For the record, `[round_rectangle]` _is_ used in one place, for the shadow under the story screen title. I thought it was also used for buttons but apparently I was wrong… in any case it's also a generally useful primitive so I wouldn't recommend deprecating it.

Personally, I don't think this change needs to be mentioned in this changelog.

stevecotton:
> For the record, `[round_rectangle]` _is_ used in one place, for the shadow under the story screen title. I thought it was also used for buttons but apparently I was wrong… in any case it's also a generally useful primitive so I wouldn't recommend deprecating it.

Not since e369e10eee094068bcb64434633fdc49044a85ed.

CelticMinstrel:
Well, that commit is obviously wrong in the first place, so… it should be reverted and fixed properly. It is absolutely essential to have some sort of background behind the story screen title. Without that, the text will inevitably be unreadable sometimes.

stevecotton:
@CelticMinstrel, we should just remove `[round_rectangle]` now. Either an immediate level 4 deprecation or a level 3 with a note to look more closely at what it draws:

1. It's not a useful primitive in its current state because the rounded corners that it draws look bad. See the image below.
2. Better alternatives can be written in WML instead of C++, and are already used in mainline: https://github.com/wesnoth/wesnoth/blob/master/data/gui/macros/button.cfg
3. The first person to notice that it draws the bad corners isn't trying to use it; instead I'm trying to test that my fix to #5043 doesn't cause a regression in it.

To see what it does draw, revert 120aa0e6caf6876afa9f44dea8c1b6cd9efb28c4, revert e369e10eee094068bcb64434633fdc49044a85ed and then change the `fill_color`s in data/gui/widget/label_story_screen_title.cfg to something bright. It looks like this:
image

CelticMinstrel:
I think it makes more sense to fix it rather than remove it. And that WML example you link is absolutely horrible, so holding it up as a "better alternative" seems rather questionable. It would be much better if that WML example could just _use_ `[round_rectangle]`. I understand there are some reasons why it can't (eg, top and bottom borders are a different colour, right?), but maybe those reasons can be addressed.

PR #5681, opened 2021-04-07
Only render the visible area in canvas.cpp

stevecotton:
If a widget is partially hidden or off-screen, this commit makes canvas.cpp
optimise and only draw the visible part.

Many of the asserts in canvas.cpp are removed, as the surface is no longer the
full canvas. The SDL_Render* functions do support the out-of-bounds and
negative values for x and y that are now passed to them, these are tested by
SDL2's testviewport.c.

Scrolling by a single pixel will force canvas::draw to do the full work of
redrawing the canvas. I had considered rendering a few of the off-screen lines
too, however it seems this isn't optimisable because the dirty flag is already
set on each redraw - that can be traced to window.cpp's push_draw_event()
causing canvas::set_is_dirty(true) to be called.

Refactor handling of the x, y, w, h variables with a common rect_bounded_shape
class, so that there are less code paths that might have unnoticed bugs.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,7 @@
  ### User interface
  ### WML Engine
  ### Miscellaneous and Bug Fixes
+   * More optimizations in the UI drawing code, these shouldn't have visible effects (PR #5681).
    * Made GUI.pyw compatible with Python 3.9 (issue #5719).
 
 ## Version 1.15.12
--- a/src/gui/core/canvas.cpp
+++ b/src/gui/core/canvas.cpp
@@ -50,9 +50,6 @@ static void set_renderer_color(SDL_Renderer* renderer, color_t color)
 /**
  * Draws a line on a surface.
  *
- * @pre                   The caller needs to make sure the entire line fits on
- *                        the @p surface.
- * @pre                   @p x2 >= @p x1
  * @pre                   The @p surface is locked.
  *
  * @param canvas          The canvas to draw upon, the caller should lock the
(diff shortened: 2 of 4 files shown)


shikadiqueen:
I don't know (yet) what happened here but the rendering quality of text in multiple places is seriously degraded on Linux/X11. This is a very big no-no as it stands:

image

shikadiqueen:
Unmodified code for comparison:

image

shikadiqueen:
There is also an issue where if you launch HttT and progress the story screen all the way to the end (journey map), the background becomes black while the battle icon is flashing on the screen.

stevecotton:
Thanks for catching those - I've moved the PR back to draft state while I debug them. The text looks OK with HDPI, and somehow I missed it.

With the text looking thin, I'm worried that I've introduced a *bugfix* rather than a bug. The old text looks very similar to the result of printing the text twice on the same surface, and this PR added some extra cleanup if an old surface was reused, as well as discarding old surfaces if the size of the viewport changed.

shikadiqueen:
> With the text looking thin, I'm worried that I've introduced a bugfix rather than a bug. The old text looks very similar to the result of printing the text twice on the same surface

This is extremely unlikely based on my observations during the removal of SDL_ttf. I can assure you at no point do I overrender text in the transitional API used for GUI1.

stevecotton:
Thanks for the hint, it saved me a lot of time. The font problem turned out to be swapping a `blit_surface` for an `sdl_blit`.

I have a bit more testing to do before reopening the PR.

stevecotton:
The build failure was a network error after building - the build stage itself succeeded.

CelticMinstrel:
I don't have any issues with this, although I didn't look at it in great detail. As long as it works, I think it's probably fine.

stevecotton:
Please review this, the graphics problem has already been fixed.

The recent force-pushes have only been to move to the post-1.15.12 changelog and to correct a grammar error in a comment.

shikadiqueen:
No obvious regressions observed. Again I don't want to look at the code too closely since it's icky GUI2 implementation details.

Issue #4959, opened 2020-06-21
Credits / about out of sync between 1.14 and git master


ivanovic:
### Game and System Information
 - irrelevant

### Describe the bug
The credits between 1.14 and master are significantly out of sync. Some data was added to the 1.14 credits only, other data was added to master only. For the information on the website we always use master assuming that changes to stable are forward ported to master as well.

### To Reproduce

1. get a checkout of 1.14 and master
2. copy data/core/about.cfg from 1.14 to master
3. run git diff on master, resulting diff: https://pastebin.com/2jR7rLRy

### Expected behavior
No extra credits added in diff in master, especially existing name entries are identical. Removals are expected though since master might contain more content than 1.14. But names should not be gone of the obviously updated name entries be reverted.

This might also be an issue for the about sections in the various campaigns _main.cfg files. This issue also affected the translation credits (especially German translation) and was fixed there in master.

### Screenshots
N/A

### Savefiles
N/A

### Additional context
resulting diff: https://pastebin.com/2jR7rLRy

stevecotton:
The campaigns still need checking, but the main about.cfg file was already mostly in sync. After backporting two corrections to 1.14 (the number in newfrenchy83's name, and iKevinY changing to Kevin Yap), every entry in 1.14 lines up with one in master.

In master, "Astrid Halberkamp" should change to "Astrid Halberkamp (bumbadadabum)".

stevecotton:
> In master, "Astrid Halberkamp" should change to "Astrid Halberkamp (bumbadadabum)".

@Asheviere, would you prefer it if the nickname was removed from your credits in 1.14? I was about to add it to master, but then saw that it had been deliberately removed in 287ec15c101b030af956d640f04328803c5360e1.

stevecotton:
The campaigns are up to date, although if we backport removing Astrid's nickname then there's an instance in Eastern Invasion's _main.cfg too.

stevecotton:
All done.

PR #5682, opened 2021-04-07
Rendering only a viewport area in pango_text

stevecotton:
This is meant to be an optimisation, which will reduce memory usage when
rendering text. It doesn't seem to reduce the amount of work done by Pango, but
the drawing will be clipped in the Cairo library.

The line-by-line workaround for rendering huge areas of text is dropped, as we
now only expect to be rendering screen-sized or smaller parts of the text.
Rendering huge texts (greater than 64kpixels high) requires the caller of
pango_text::render() to change to calling pango_text::render(SDL_Rect) instead,
which will be handled in a separate commit - fixing the known crash when
opening the credits screen will require that commit on top of this one.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -18,7 +18,7 @@
  ### WML Engine
    * Modify implementation of overwrite_specials attribute for replace yes/no parameter by none/one_side/both_sides and select abilities used like weapons and specials who must be overwrited(owned by fighter where special applied or both)
  ### Miscellaneous and Bug Fixes
-   * More optimizations in the UI drawing code, these shouldn't have visible effects (PR #5681).
+   * More optimization in the UI drawing code, fixes the crash displaying the full credits (issue #5043).
    * Made GUI.pyw compatible with Python 3.9 (issue #5719).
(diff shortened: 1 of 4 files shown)


stevecotton:
Put into draft state waiting for #5736 to merge.

stevecotton:
@Pentarctagon if there's no feedback then I intend to merge this today. I feel it's better to have the credits screen working in the next release, so that everyone who wants to can check that they're in their desired place in the list before the string-freeze.

I'm also planning to make a separate PR moving myself up the credits list from the Miscellaneous Contributors section, so might as well update anyone else's entry too.

Pentarctagon:
If you're going to be updating a bunch of the credits anyway, could you take a look at #4959 at the same time as well?

stevecotton:
> It doesn't seem to reduce the amount of work done by Pango, but the drawing will be clipped in the Cairo library.

The upstream wishlist-issue for that is https://gitlab.gnome.org/GNOME/pango/-/issues/79

Issue #5616, opened 2021-03-15
CI: WML unit tests results make it very hard to find errors


stevecotton:
@CelticMinstrel the WML unit test script's results are currently very confusing. Below is part of the output of the c611128 validation (which passes, none of these are problems). However, this makes it very hard to work out why the CI failed on other runs - there's one other error, somewhere.

```
Running test test_move_fail_6
Error (strict mode, strict_level = 1): wesnoth reported on channel warning replay
20210314 23:50:58 warning replay: Warning: Path data contained something which could not be parsed to a sequence of locations:
 config = x = 16,15,14,13,12,11
y = 3,3,3,3,3,bock

FAIL TEST (BROKE STRICT): test_move_fail_6

Running test test_store_unit_defense_deprecated
Error (strict mode, strict_level = 1): wesnoth reported on channel error deprecation
20210314 23:50:59 error deprecation: [store_unit_defense] has been deprecated and will be removed in version 1.17.0.
  This function returns the chance to be hit, high values represent bad defenses. Using [store_unit_defense_on] is recommended instead.
20210314 23:50:59 error deprecation: [store_unit_defense] has been deprecated and will be removed in version 1.17.0.
  This function returns the chance to be hit, high values represent bad defenses. Using [store_unit_defense_on] is recommended instead.
20210314 23:50:59 error deprecation: [store_unit_defense] has been deprecated and will be removed in version 1.17.0.
  This function returns the chance to be hit, high values represent bad defenses. Using [store_unit_defense_on] is recommended instead.
20210314 23:50:59 error deprecation: [store_unit_defense] has been deprecated and will be removed in version 1.17.0.
  This function returns the chance to be hit, high values represent bad defenses. Using [store_unit_defense_on] is recommended instead.
20210314 23:50:59 error deprecation: [store_unit_defense] has been deprecated and will be removed in version 1.17.0.
  This function returns the chance to be hit, high values represent bad defenses. Using [store_unit_defense_on] is recommended instead.
20210314 23:50:59 error deprecation: [store_unit_defense] has been deprecated and will be removed in version 1.17.0.
  This function returns the chance to be hit, high values represent bad defenses. Using [store_unit_defense_on] is recommended instead.
FAIL TEST (BROKE STRICT): test_store_unit_defense_deprecated

Running test alice_kills_bob
PASS TEST (VICTORY): alice_kills_bob

Running test bob_kills_alice_on_retal
[...]

PR #5688, opened 2021-04-08
More unit tests for events

CelticMinstrel:
This just exists to ensure CI passes, but if anyone wants to comment on the new tests, feel free.

### DIFF
--- a/data/test/scenarios/event_handlers_in_events.cfg
+++ b/data/test/scenarios/event_handlers_in_events.cfg
@@ -233,3 +233,92 @@
         {RETURN ({VARIABLE_CONDITIONAL pass_test equals 1})}
     [/event]
 )}
+
+{GENERIC_UNIT_TEST "event_handers_in_events__delayed" (
+    [event]
+        name=prestart
+        {VARIABLE pass_test 0}
+
+        [event]
+            name=start
+            delayed_variable_substitution=yes
+            {VARIABLE pass_test "$($pass_test + 1)"}
+        [/event]
+        
+        {VARIABLE pass_test 1}
+    [/event]
+    [event]
+        name=turn 1
+        {RETURN ({VARIABLE_CONDITIONAL pass_test equals 2})}
+    [/event]
+)}
+
+{GENERIC_UNIT_TEST "event_handlers_in_events__dynamic_filter" (
+    [event]
+        name=start
+        {VARIABLE pass_test 0}
+        [set_variables]
+            name=filter
+            [value]
+                race=elf
+            [/value]
+        [/set_variables]
+        {UNIT 1 "Orcish Grunt" 1 1 ()}
+        {UNIT 1 "Elvish Archer" 2 2 ()}
+        [event]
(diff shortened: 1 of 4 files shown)


data/test/macros/wml_unit_test_macros.cfg
```diff
@@ -88,6 +88,7 @@ Elvish Archer#endarg
                 name = "Alice"
                 type = {SIDE1_LEADER}
                 id=alice
+                random_traits=no
```
stevecotton, line 91: Why is this line necessary? Alice and Bob already don't get random traits, so if this line is needed then it indicates an API change that's subtle but which will affect UMC.

CelticMinstrel: They don't get random traits? How do you know? I can remove it if it's redundant.

stevecotton: I ran a test within the GUI (ran it with `--test`) and checked.

CelticMinstrel: Why do they not get random traits if they don't have this line…?

stevecotton:
> Why do they not get random traits if they don't have this line…?

I'm not sure, but think it's something about only generating deterministic numbers during scenario creation, thus no random numbers for the traits. Or it might be because they're leaders.

> Note: I don't see how this change would affect UMC though.

If someone has balanced their scenario assuming that the enemy leader doesn't have the `quick` trait then it might break. I can remember an example with a leader capturing a village, and then on turn 2 moving to a tiny castle that was closer than their main base.

One of the things that the unit tests do is to ensure that things don't change during a stable branch. If someone finds the reason that Alice and Bob don't get traits and 'fixes' it then, without this line, the unit tests will notice that Alice randomly has more MP.

CelticMinstrel:
Oh, come to think of it, it's probably because they're leaders. Okay, I'll remove that line then.

CelticMinstrel:
The CI test failures aren't in the WML unit tests, so I'm merging this.

stevecotton:
> The CI test failures aren't in the WML unit tests, so I'm merging this.

Yes they are, they're your bug. After your opposition in #5616 to making the CI logs easier to read, it's disappointing to see that you didn't read them before merging.

```
20210409 14:27:32 error validation: Invalid key 'accuracy=' in tag [attack]
at test/scenarios/events-test_autovariables.cfg:252
    included from test/_main.cfg:29
    included from data/_main.cfg:68
20210409 14:27:32 error validation: Extra tag [recruit]; there may only be 1 [recruit] in [do_command]
at core/macros/utils.cfg:152
    included from core/_main.cfg:19
    included from data/_main.cfg:18
20210409 14:27:32 error validation: Invalid key 'accuracy=' in tag [attack]
at test/scenarios/events-test_filters.cfg:156
    included from test/_main.cfg:29
    included from data/_main.cfg:68
validation failed
Test validation complete!  Success: No
```

PR #5690, opened 2021-04-08
Use the database to store and retrieve login history for the searchlog command.

Pentarctagon:
This allows the history to be persisted across restarts, whereas right now it's lost.

### DIFF
--- a/src/server/common/dbconn.cpp
+++ b/src/server/common/dbconn.cpp
@@ -17,7 +17,9 @@
 #include "server/common/resultsets/tournaments.hpp"
 #include "server/common/resultsets/ban_check.hpp"
 #include "server/common/resultsets/game_history.hpp"
+
 #include "log.hpp"
+#include "serialization/unicode.hpp"
 
 static lg::log_domain log_sql_handler("sql_executor");
 #define ERR_SQL LOG_STREAM(err, log_sql_handler)
@@ -36,6 +38,7 @@ dbconn::dbconn(const config& c)
 	, db_tournament_query_(c["db_tournament_query"].str())
 	, db_topics_table_(c["db_topics_table"].str())
 	, db_addon_info_table_(c["db_addon_info_table"].str())
+	, db_connection_history_table_(c["db_connection_history_table"].str())
 {
 	try
 	{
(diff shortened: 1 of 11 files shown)


src/server/common/dbconn.cpp
```diff
@@ -372,6 +375,74 @@ void dbconn::insert_addon_info(const std::string& instance_version, const std::s
...
+	{
+		log_sql_exception("Unable to insert login row user `"+username+"` and ip `"+ip+"`.", e);
+	}
+}
+
+void dbconn::update_logout(const std::string& username, const std::string& ip)
+{
+	try
+	{
+		modify(connection_, "UPDATE `"+db_connection_history_table_+"` SET LOGOUT_TIME = CURRENT_TIMESTAMP WHERE USER_NAME = ? AND IP = ? and LOGOUT_TIME IS NULL",
```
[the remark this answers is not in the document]
Pentarctagon: I'm not sure - I'd need to check if mariadbpp supports returning a generated ID from an insert query.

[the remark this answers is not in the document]
Pentarctagon: Or actually it looks like that's what `insert()` already does, so I just need to test it.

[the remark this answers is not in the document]
Pentarctagon: @soliton- Updated.

src/server/wesnothd/server.cpp
```diff
@@ -2813,33 +2773,41 @@ void server::searchlog_handler(const std::string& /*issuer_name*/,
...
 			}
 		}
-	}
 
-	if(!found_something) {
-		*out << "\nNo match found.";
+		if(!found_something) {
+			*out << "\nNo match found.";
+		}
+	} else {
```
soliton-, line 2805: I suppose there is some wildcard conversion needed? Our own wildcard matching works on `*` and `?` while SQL LIKE uses `%` and `_`. `_` is a bit problematic since it's allowed in user names. I would suggest to just convert `*` to `%` since it's the most useful anyway and escape `_`.

soliton- (changes requested):

src/server/common/dbconn.cpp
```diff
@@ -372,6 +375,76 @@ void dbconn::insert_addon_info(const std::string& instance_version, const std::s
...
+	{
+		log_sql_exception("Unable to update login row `"+std::to_string(login_id)+"`.", e);
+	}
+}
+
+void dbconn::get_users_for_ip(const std::string& ip, std::ostringstream* out)
+{
+	try
+	{
+		mariadb::result_set_ref rslt = select(connection_, "SELECT USER_NAME, IP, date_format(LOGIN_TIME, '%Y/%m/%d %h:%i:%s'), coalesce(date_format(LOGOUT_TIME, '%Y/%m/%d %h:%i:%s'), '(not set)') FROM `"+db_connection_history_table_+"` WHERE IP LIKE ? order by LOGIN_TIME",
```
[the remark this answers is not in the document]
Pentarctagon: I think `text%` has always been able to use an index.  `%text%` used to not be able to use an index, but should support indexing as of a few years ago at least.

Pentarctagon:
The previous comments should all be addressed now.

src/server/common/forum_user_handler.hpp
```diff
@@ -221,6 +221,40 @@ class fuh : public user_handler
...
+	/**
+	 * Updates the database for when a player logs out.
+	 * 
+	 * @param login_id The generated ID that uniquely identifies the row to be updated.
+	 */
+	void db_update_logout(unsigned long long login_id);
+
+	/**
+	 * Searches for all players that logged in using the ip address.
+	 * The ip address provided can be a partial address.
```
soliton-, line 241: I'd rather say it can use wildcards since if you just hand in a partial address it's just not going to match.

soliton- (approved):

src/server/common/forum_user_handler.hpp
Pentarctagon: Updated.

src/server/common/forum_user_handler.hpp
```diff
@@ -221,6 +221,40 @@ class fuh : public user_handler
...
+	/**
+	 * Updates the database for when a player logs out.
+	 * 
+	 * @param login_id The generated ID that uniquely identifies the row to be updated.
+	 */
+	void db_update_logout(unsigned long long login_id);
+
+	/**
+	 * Searches for all players that logged in using the ip address.
+	 * The '*' wildcard can be used to search for partial ip addresses.
```
soliton-, line 241: These functions are called after the wildcard conversion so they take SQL wildcards.

Pentarctagon: Updated.

PR #5693, opened 2021-04-09
Fix better_attack evaluation for poison attacks

mattsc:
Poison is added to the attack damage when comparing attacks by the same unit with different weapons, but it should only be counted as a bonus if the enemy survives, otherwise it serves no purpose.

### DIFF
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -518,11 +518,11 @@ bool battle_context::better_combat(const combatant& us_a,
 		return true;
 	}
 
-	// Add poison to calculations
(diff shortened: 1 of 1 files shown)


stevecotton:
I'm updating the changelog and assuming that this affects both the UI in the attack dialog and also the choices by the AI, is that correct?

mattsc:
> I'm updating the changelog and assuming that this affects both the UI in the attack dialog and also the choices by the AI, is that correct?

That is correct, yes.  Thanks for taking care of the changelog entry.

PR #5694, opened 2021-04-10
When checking [variable] comparison attributes, return after checking the first one.

Pentarctagon:
The current implementation technically allows for there to be multiple comparison attributes present and it will return false if any of them are false or return true if all of them are true. This changes the implementation to only check the first comparison attribute found and return true or false based on that check, which is how the [variable] tag is documented as working on the wiki.

### DIFF
--- a/src/game_events/conditional_wml.cpp
+++ b/src/game_events/conditional_wml.cpp
@@ -101,7 +101,7 @@ namespace builtin_conditions {
 			if (values.has_attribute(name)) { \
 				std::string attr_str = values[name].str(); \
 				std::string str_value = value.str(); \
-				if (!(test)) return false; \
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Hmm… is it possible to warn if there are multiple comparison attributes? Maybe just as simple as checking the total number of attributes?

Pentarctagon:
Does vconfig have a way to get a count of the attributes?

Pentarctagon:
Merged given the conversation on Discord - I'd still like to know whether vconfig can give a count of attributes or not though.

CelticMinstrel:
I'm not sure about vconfig, but config definitely has it, and you can get the underlying config from a vconfig.

PR #5697, opened 2021-04-10
Simplify canvas: renderer lifetime and only one list of shapes

stevecotton:
The renderer only needs to exist during draw(). The old implementation kept the
old renderer until the next call to draw(), but would then usually destroy the
old one and create a new one.

Canvas kept separate lists of what had already been drawn and what needed to be
drawn, but whenever changes happened all of the drawn_shapes_ were put back
into shapes_. No point in having two separate lists.

The list of shape needs pointers to support polymorphism, but doesn't need
shared_ptrs.

This is some cleanup to make #5681 slightly smaller.

### DIFF
--- a/src/gui/core/canvas.cpp
+++ b/src/gui/core/canvas.cpp
@@ -798,7 +798,6 @@ void text_shape::draw(surface& canvas,
 
 canvas::canvas()
 	: shapes_()
-	, drawn_shapes_()
 	, blur_depth_(0)
 	, w_(0)
 	, h_(0)
@@ -811,7 +810,6 @@ canvas::canvas()
 
 canvas::canvas(canvas&& c) noexcept
 	: shapes_(std::move(c.shapes_))
-	, drawn_shapes_(std::move(c.drawn_shapes_))
 	, blur_depth_(c.blur_depth_)
 	, w_(c.w_)
 	, h_(c.h_)
@@ -856,10 +854,6 @@ void canvas::draw(const bool force)
 		shape->draw(canvas_, renderer.get(), variables_);
 	}
 
-	// The shapes have been drawn and the draw result has been cached. Clear the list.
-	std::copy(shapes_.begin(), shapes_.end(), std::back_inserter(drawn_shapes_));
-	shapes_.clear();
-
 	SDL_RenderPresent(renderer.get());
 
 	is_dirty_ = false;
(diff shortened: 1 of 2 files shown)


src/gui/core/canvas.cpp
```diff
@@ -844,33 +834,27 @@ void canvas::draw(const bool force)
 		variables_.add("height", wfl::variant(h_));
 	}
 
+	auto renderer = std::unique_ptr<SDL_Renderer, decltype(&SDL_DestroyRenderer)> {nullptr, &SDL_DestroyRenderer};
```
CelticMinstrel, line 837: Is it possible to simplify this by leveraging template deduction guides? If `std::unique_ptr` has these properly configured then I would expect the following to work:

```
auto renderer = std::unique_ptr(SDL_CreateSoftwareRenderer(canvas_), &SDL_DestroyRenderer);
```

Or maybe even:

```
std::unique_ptr renderer(SDL_CreateSoftwareRenderer(canvas_), &SDL_DestroyRenderer);
```

On a related note, there doesn't seem to be any reason to create the renderer in the if statement, as both branches create it in the same way.

stevecotton: Depending on which branch we're going to take, `canvas_` might be null at this point. So to move the `SDL_CreateSoftwareRenderer` call out of the `if`, the code would end looking as if it had 4 code paths even though it's only 2:

```
const bool must_clear_canvas = canvas_;
if(!canvas_) {
    DBG_GUI_D << "Canvas: create new empty canvas.\n";
    canvas_ = surface(w_, h_);
}
std::unique_ptr renderer(SDL_CreateSoftwareRenderer(canvas_), &SDL_DestroyRenderer);
if(must_clear_canvas) {
    DBG_GUI_D << "Canvas: clearing cached canvas.\n";
    SDL_SetRenderDrawColor(renderer.get(), 0, 0, 0, 0);
    SDL_RenderClear(renderer.get());
}
```

CelticMinstrel: Hmm… I can't decide whether that's okay or not…

Even if it's not, there's also the possibility of casting nullptr:

```
std::unique_ptr renderer(static_cast<SDL_Renderer*>(nullptr), &SDL_DestroyRenderer);
```

Alternatively, don't rely on deduction guides but also don't use auto:

```
std::unique_ptr<SDL_Renderer> renderer(nullptr, &SDL_DestroyRenderer);
```

stevecotton: GCC 10 refuses to compile either of those, are you sure they're valid?

CelticMinstrel: I thought they would be…? Do the two from my other post work? I don't think there's any semantic difference between the second one in my first post and the first in my second post. Unless it doesn't like casting nullptr to a specific pointer type, but that seems unlikely.

stevecotton: https://stackoverflow.com/questions/51109767/why-cant-unique-ptrs-template-arguments-be-deduced

AI0867: So the answer is pretty much no. `std::make_unique` might help, but then you can't pass `SDL_DestroyRenderer`.

stevecotton: I wondered `auto renderer = create_auto_deleter<SDL_DestroyRenderer>();` was possible, with the template deducing the type `SDL_Renderer` from deleter's arguments. I think there is, but didn't reach a working template before deciding that the current code is going to be easier to read.

CelticMinstrel: So in other words, none of my examples work, huh? That's unfortunate…

CelticMinstrel (approved):

stevecotton:
@Vultraz should I wait or merge, please?

Issue #4179, opened 2019-07-17
vision, [store_reachable_locations] and units with vision different to movement


stevecotton:
`[store_reachable_locations]` supports storing `range=vision`, which as implement and documented on the Wiki stores "the vision range (movement range ignoring Zone of Control + 1 hex)."

That assumption is broken by the feature that units can have [vision_costs] separate to their [movement_costs] or vision_points different to their max movement points. There isn't currently a WML function for storing the vision of such units.

CelticMinstrel:
Suggestion: Add support for `range=movement` and `range=jamming` to use movement and jamming costs+points respectively. Then make `range=vision` use the vision costs+points.

That might be a slight compatibility break, mind you... maybe leave `range=vision` as-is for 1.14 and also add `range=sight` for actual vision points?

Maybe also a mode that uses vision but also accounts for enemy jamming.

stevecotton:
The existing `viewing_side` key could also trigger accounting for enemy jamming (I haven't checked if it already does).

I don't understand jamming yet, but I think it's more than a list of locations. IIUC it's also the number of jamming points that are left over when that hex is reached, with the left-over points being the penalty that's applied to enemy vision.

stevecotton:
Added to the roadmap for 1.15.9.

PR #5704, opened 2021-04-14
scenario-test: Add a right-click "show vision range" option

stevecotton:
Make "show attack range" not overwrite the array for cleaning up
the "show movement range" command's display.

Add a label and tooltip about the enemy unit with jamming.

The new option shows the result of [store_reachable_locations]range=vision.
There are known issues (in #4179) with that for:
* units with vp different to mp
* units with vision costs different to movement costs
* jamming by enemy units

### DIFF
--- a/data/scenario-test.cfg
+++ b/data/scenario-test.cfg
@@ -620,6 +620,13 @@ Adjacent own units of lower level will do more damage in battle. When a unit adj
     [/side]
 
     [label]
+        x,y=12,13
+        text="Jamming Orc"
+        tooltip="This unit’s ability reduces vision, that’s why 13,13 is shrouded at the start."
+        visible_in_shroud=yes
+    [/label]
+
+    [label]
         x,y=20,12
         text="Enhance Vision"
         tooltip="Set the unit's vision to 15."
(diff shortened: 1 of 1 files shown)


data/scenario-test.cfg
```diff
@@ -619,6 +619,13 @@ Adjacent own units of lower level will do more damage in battle. When a unit adj
         [/unit]
     [/side]
 
+    [label]
+        x,y=12,13
+        text="Jamming Orc"
+        tooltip="This unit’s ability reduces vision, that’s why 13,13 is shrouded at the start."
```
CelticMinstrel, line 625: I think I'd say "is initially shrouded" rather than "is shrouded at the start" but I guess it's not a big deal, both work.

CelticMinstrel:
This seems fine to me.

data/scenario-test.cfg
stevecotton: I like that and it does sound better for native speakers. However, I'm going to keep it as-is because I think keeping the language simpler is better for an obscure tooltip in a dev-only scenario.

PR #5705, opened 2021-04-14
Make [store_reachable_locations]range=vision calculate vision

### LINKED ISSUE #4179: vision, [store_reachable_locations] and units with vision different to movement

stevecotton:
Previously it calculated max movement, and then added the adjacent hexes. This
version should correctly handle:
* units with vp different to max mp
* units with vision costs different to movement costs
* jamming by enemy units

An interactive test showing the before and after results is in #5704
Fixes #4179

### DIFF
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -1,5 +1,4 @@
 local helper = wesnoth.require "helper"
-local location_set = wesnoth.require "location_set"
 local utils = wesnoth.require "wml-utils"
 local wml_actions = wesnoth.wml_actions
 local T = wml.tag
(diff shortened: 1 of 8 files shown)


CelticMinstrel:
I don't see any issues with this, but it's also not an area I'm really familiar with so I could be missing something.

`wesnoth.find_vision_range` is probably going to move into a pathfinding module or something, so it feels a bit silly to be adding it here only to move it soon after, but I guess it's fine.

stevecotton:
Should anyone else review it?

Working on this, I get the feeling that no-one is familiar with it. Particularly I feel that I'll now be the person that people come to if they want to know what 'jamming' is.

PR #5708, opened 2021-04-15
Add a Lua unit_test module only available in [test] scenarios

CelticMinstrel:
This is for future use, especially in #5663 - it was requested that it be extracted to a separate PR, so here it is.


stevecotton:
Please bring in the tests of this code from data/test/scenarios/test_assert.cfg too, although I guess they would need to use `[lua]` tags instead of `[event]code=`.

CelticMinstrel:
As far as I know, all the tests actually using the `unit_test` module are reliant on the events PR to work… I suppose I could convert some of the most basic ones (test_assert and also test_return) to use a `[lua]` tag though as you mention, and maybe also use it for some other pre-existing Lua unit tests…

CelticMinstrel:
I could rebase, but I suspect we'll be using squash and merge anyway, right? In which case there's no point.

Those are all the existing unit tests I could find where I thought it made sense to use the new module. I'm kinda glad I looked too, as it resulted in three new functions in the `unit_test` module.

data/test/scenarios/test_return.cfg
```diff
@@ -15,3 +15,21 @@
...
     [/event]
 )}
+
+# Same as the above two, but using the Lua unit_test module
+{GENERIC_UNIT_TEST "test_lua_return" (
+    [event]
+        name = start
+        [lua]
+            code=<<unit_test.pass_if(true, 'return true')>>
+        [/lua]]
```
stevecotton, line 25: There's an extra `]` on this line and on line 33, which means the preprocessor fails before any tests run.

stevecotton (changes requested):
Having fixed the syntax error, the tests time out during the `lua_wml_tagnames` test.

CelticMinstrel:
Okay, all tests pass… and given the basic sanity unit test cases, I don't _think_ it's likely that those passes represent a failure of the system rather than a genuine pass. Any other comments before this is merged?

stevecotton:
CI output lines 1642-1645 for the ubuntu-2004-scons-gcc-release-lto build:
```
Executing WML tests
Failure, Wesnoth returned 2 TIMEOUT for test_wml_conditionals but we expected 0 PASS
WML Unit Tests Result: 259 of 260 tests passed
    (259 passed, 1 failed)
```

Despite that, the CI says the run passed; this means there's a problem with the code under test and a problem with the script running the test. I tried to debug a timeout about 24 hours ago when testing this PR - there's definitely a timeout bug introduced in this PR.

CelticMinstrel:
I don't see how it's possible for a test that this doesn't even touch to fail because of this PR, but I'll take a look at it.

stevecotton:
Thanks for fixing the off by one error when the script prints the name of the failing test; the test that's timing out is really `test_lua_tagnames`. I reindented that one yesteday while debugging, but my reindent isn't the problem - it's still passing on master.

CelticMinstrel:
Okay, I fixed the broken test (it was a syntax error, not sure why it didn't show in the log) and added some documentation. Does this seem ready to merge now? I figured I'd squash it when it's merged rather than force-pushing.

data/lua/core/unit_test.lua
```diff
@@ -0,0 +1,204 @@
...
+			linger_mode = true
+		}
+	end
+
+	--! Fail the test unconditionally
+	function unit_test.fail()
+		unit_test.finish('fail')
+	end
+
+	--! Pass the test unconditionally
```
stevecotton, line 21: What happens if neither `pass()` nor `fail()` is called? Will it be like `{SUCCEED}`, where the test ends up running endlessly?

What happens if an assert fails first? Does Lua execution skip the `pass()`, does the `pass()` override the earlier status, or does `pass()` conditionally not cause the test to pass?
```
unit_test.assert(x == 1, "Wrong value for x")
unit_test.pass()
```

data/test/scenarios/as_text.cfg
```diff
@@ -26,42 +26,29 @@
...
+                unit_test.assert_contains(wml_tab_txt, '{"1":{"1":"one","2":{"1":{"1":"first","2":{')
+                unit_test.assert_contains(wml_tab_txt, '"a":1')
+                unit_test.assert_contains(wml_tab_txt, '"b":5')
+                unit_test.assert_contains(wml_tab_txt, '"c":true')
+                unit_test.assert_contains(wml_tab_txt, ',"2":{"1":"two","2":{"1":{"1":"second","2":{')
+                unit_test.assert_contains(wml_tab_txt, '"x":9')
+                unit_test.assert_contains(wml_tab_txt, '"y":3')
+                unit_test.assert_contains(wml_tab_txt, '"z":false')
 
                 -- Pass the test. Doesn't do anything if any of the above assertions has failed.
```
stevecotton, line 50: This comment does not match the current documentation of `unit_test.pass()`.

data/test/scenarios/test_assert.cfg
```diff
@@ -27,3 +27,167 @@
...
+        [/lua]
+    [/event]
+)}
+{GENERIC_UNIT_TEST "test_lua_assert_fail" (
+    [event]
+        name = start
+        [lua]
+            code=<<
+                unit_test.assert(false, 'assert false')
+                unit_test.pass_if(true, 'return true')
```
stevecotton, line 49: What is this test testing? If it's testing `unit_test.assert`, then why is it using a `pass_if` instead of `pass`?

stevecotton (changes requested):
What do we need to do to make the unit tests a normal part of developers' local workflows; instead of, as has happened again, the CI getting fed code that doesn't run?

Also, `run_wml_tests` crashed on the line that was changed in 9535ed926c2bae38d07b81b710fc017fd7924b53.

data/test/scenarios/test_assert.cfg
```diff
@@ -27,3 +27,167 @@
...
+            code=<<
+                unit_test.assert_equal('foo', 'foo', 'string equality')
+                unit_test.assert_not_equal('foo', 'bar', 'string inequality')
+                unit_test.assert_greater('foo', 'bar', 'string ordering')
+                unit_test.assert_equal(1, 1, 'numeric equality')
+                unit_test.assert_not_equal(1, 2, 'numeric inequality')
+                unit_test.assert_greater(2, 1, 'numeric greater-than')
+                unit_test.assert_less(1, 2, 'numeric less-than')
+                unit_test.assert_in_range(1, 0, 2, 'numeric in-range')
+                unit_test.assert_approx_equal(1, 1.0001, 0.0001, 'numeric approx-equal')
```
stevecotton, line 80: Here the difference between the two values is the same as the delta, so the result of this should be uncertain.

Instead you should test a value that's certain - such as `(1, 1.0001, 0.001)`, where the difference is an order of magnitude less than the delta.

data/lua/core/unit_test.lua
CelticMinstrel: If neither is called, then the test will never terminate.

These use `[endlevel]` under the hood, and as I understand it, the first `[endlevel]` wins and any subsequent ones do nothing. Thus, an assert failing means a subsequent pass does nothing.

data/test/scenarios/test_assert.cfg
CelticMinstrel: It's a direct conversion of `test_assert` from the same file. I'm not sure why that test asserts and returns instead of asserting and succeeding.

stevecotton: I feel the aim should be to test the Lua unit test framework, so it's more important to work out what it should test rather than to try and copy existing tests.

data/lua/core/unit_test.lua
stevecotton: The documentation should reflect that.

Keeping compatibility with the quirks of the WML unit test framework might help when we replace that (I assume we'll port the old macros to be based on the Lua unit test framework), but a new implementation should be an opportunity to change to a better-understood implementation.

data/test/scenarios/test_assert.cfg
CelticMinstrel: Well, ultimately the choice of `pass` or `pass_if` here makes no real difference… so I guess I'll change it.

stevecotton:
What would you need to make the unit tests a normal part of your local workflow; instead of, as has happened again, stuff getting uploaded that fails the tests on the CI?

CelticMinstrel:
The error I'm seeing in the CI seems to be the bug with `run_wml_tests` that I think your own PR #5715 was addressing. I was running the tests with a local bugfix (as mentioned elsewhere, initializing `num_passed` to 0 instead of 1), all the tests passed. So it seems likely that if we merge your PR, the CI on this PR will also pass.

> What would you need to make the unit tests a normal part of your local workflow

It's pretty much impossible to make them a normal part of my workflow - they just take too long to complete for them to be added as a step in the Xcode build. Still, obviously I would be running them manually when I'm doing something that directly relates to them.

stevecotton:
The unit tests could easily run during the time that you should be spending manually reviewing your own code before uploading it for others to review. You should spend more time reviewing and documenting; the currently failing test illustrates my point that you should be asking yourself what each test is testing.

stevecotton:
> One of those pass_if(true) needed to be that way

That's not the change that's needed, although it might make the CI pass.

There are 3 tests of the WML unit test framework, but that stems from the WML test framework having macros to work out the limitations of WML's syntax. The number of tests of the Lua unit test framework, along with their documentation, should show that you've designed the new framework for Lua's syntax - instead it's looking as if you've just cut, pasted, and poked it until the CI passes.

CelticMinstrel:
> That's not the change that's needed

So what do you think is needed then?

data/lua/core/unit_test.lua
Pentarctagon: I'm not sure why the WML unit test framework would be going away?  While many WML tags are implemented in lua under the hood, WML is still its own API that needs its own tests.

data/test/scenarios/test_assert.cfg
```diff
@@ -27,3 +27,276 @@
...
+        name = start
+        [lua]
+            code=<<
+                unit_test.assert(true, 'assert true')
+                unit_test.pass_if(false, 'return false')
+            >>
+        [/lua]
+    [/event]
+)}
+
```
stevecotton, line 65: I think there should be a few tests for sanity-checking the framework with multiple events. From the current implementation of finish() we can be sure this works, but it seems worth the sanity-check in case someone refactors the framework.

```
# Sanity check that a fail() in a fired event causes the test to fail
{GENERIC_UNIT_TEST "test_lua_fail_in_fired_event" (
    [event]
        id = fail_event
        [lua]
            code=<<
                unit_test.fail()
            >>
        [/lua]
    [/event]
    [event]
        name = start
        [fire_event]
            id = fail_event
        [/fire_event]
        [lua]
            code=<<
                unit_test.pass()
            >>
        [/lua]
    [/event]
)}
```

stevecotton (changes requested):

data/lua/core/unit_test.lua
```diff
@@ -0,0 +1,212 @@
...
+		if type(val) == 'string' then
+			-- Strings are output quoted but without escaped_split
+			-- Pick a quote style that the string doesn't contain
+			if val:find("'") == nil then
+				return "'" .. val .. "'"
+			elseif val:find('"') == nil then
+				return '"' .. val .. '"'
+			else
+				local n = 0;
+				-- This loop is guaranteed to end eventually, although it could run a very large number of times on particularly pathological input.
```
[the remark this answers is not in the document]
CelticMinstrel: I completely forgot that that exists… I'll switch to that, then.

data/lua/core/unit_test.lua
```diff
@@ -0,0 +1,212 @@
...
+	--! Causes the test to complete with a specific result.
+	--! This is mostly an internal detail but might very occasionally be useful to produce a result other than pass or fail
+	function unit_test.finish(result)
+		wesnoth.wml_actions.endlevel{
+			test_result = result,
+			linger_mode = true
+		}
+	end
+
+	--! End the test in unconditional failure
```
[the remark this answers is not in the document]
CelticMinstrel: I'm not actually 100% sure, but I think it will run events that are currently queued but will exit as soon as the event queue is empty. For example, if this is a "new turn" event, the "turn refresh" will still execute.

It definitely doesn't exit the current event.

data/lua/core/unit_test.lua
```diff
@@ -0,0 +1,212 @@
...
+		return tostring(val)
+	end
+
+	--! Output a log message to the test output
+	function unit_test.log(prefix, message)
+		std_print(prefix .. ': ' .. message)
+	end
+
+	--! End the test in failure if condition is true and success otherwise
+	--! Can be thought of as 'return not condition'
```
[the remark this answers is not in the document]
CelticMinstrel: I think this is a little more coherent when the test has a single assertion and whether it passes or fails is exactly the same as the assert condition. However, I suppose it does end up being equivalent to an assert plus a pass.

data/lua/core/unit_test.lua
```diff
@@ -0,0 +1,212 @@
...
+
+	--! Fail the test with a message unless the condition is true
+	function unit_test.assert(condition, message)
+		if not condition then
+			unit_test.log('Assertion failed', message)
+			unit_test.fail()
+		end
+	end
+
+	--! Fail the test with a message unless the condition is false
```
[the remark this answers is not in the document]
CelticMinstrel: The reason to use this would be similar to the reason to use eg `assert_equal` - so that the negation can be included in the message if the assertion fails. Admittedly it's not _quite_ as useful for negation as it is for equality… do you think it's worth keeping?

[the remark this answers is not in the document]
stevecotton: For new APIs, I'm strongly in the "less is better" camp.

data/lua/core/unit_test.lua
stevecotton: This comment is a wishlist, not a request for the current PR.

> I'm not sure why the WML unit test framework would be going away?

The WML unit test framework has a mechanism (the `ended` flag) to stop logging asserts after an assert has already failed. It would be nice to add similar functionality to the Lua unit test framework. It would then be nice to have the two unit test frameworks interact with each other, so that logging stops on the first failed assert even when a test mixes the two frameworks. Conclusion: in 1.18 I expect the API (well the macro names) of the WML unit test framework to still be there, but for those macros to have become a wrapper around the Lua unit test framework.

CelticMinstrel: That `ended` flag seems like something that's better left optional… there might be some asserts where you want to keep going (because if multiple asserts fail you can fix them all at once) and others where you don't (because the assertion failure guarantees that later asserts also fail).

CelticMinstrel:
Any further comments?

data/lua/core/unit_test.lua
```diff
@@ -0,0 +1,166 @@
...
+			linger_mode = true
+		}
+	end
+
+	--! End the test in unconditional failure
+	function unit_test.fail()
+		unit_test.finish('fail')
+	end
+
+	--! End the test in success if all assertions passed, otherwise in failure
```
stevecotton, line 21: Please document somewhere that the reason for needing this is so that a test can be spread over multiple events (even over multiple turns).

data/test/scenarios/test_assert.cfg
```diff
@@ -27,3 +27,291 @@
...
+        name = start
+        [lua]
+            code=<<
+                unit_test.assert(false, 'assert false')
+                unit_test.succeed()
+            >>
+        [/lua]
+    [/event]
+)}
+# Sanity check that a fail() in a fired event causes the test to fail
```
stevecotton, line 58: Should have a blank line before this

data/test/scenarios/test_return.cfg
```diff
@@ -15,3 +15,37 @@
         {RETURN ([not][true][/true][/not])}
     [/event]
 )}
+
+# Same as the above two, but using the Lua unit_test module
```
stevecotton, line 19: It's good that there are now four tests, but the documentation no longer matches.

stevecotton (changes requested):

data/lua/core/unit_test.lua
```diff
@@ -0,0 +1,166 @@
+
+--[========[Unit Test Helpers]========]
+
+if rawget(_G, 'unit_test') ~= nil then
+	print("Loading unit_test module...")
+
```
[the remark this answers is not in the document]
CelticMinstrel: It's on blank lines, so who cares.

Pentarctagon:
> I still think this needs to be discovered and added to the documentation.
> 
> @Pentarctagon if I'm pushing too hard on this point please say.

I can't seem to be able to reply directly to this, but what do you mean by "this needs to be discovered"?

CelticMinstrel:
> I can't seem to be able to reply directly to this, but what do you mean by "this needs to be discovered"?

@stevecotton's comment is in response to my comment - you can view the whole thread if you click on the Changed tab.

data/lua/core/unit_test.lua
stevecotton: Marking this conversation as resolved - the initial point about more information is still open but is also covered by https://github.com/wesnoth/wesnoth/pull/5708#discussion_r616272336

data/lua/core/unit_test.lua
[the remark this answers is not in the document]
stevecotton: Please clean it up now, instead of creating noise in a subsequent general "clean up whitespace" commit.

stevecotton:
It's almost looking ready, just the whitespace cleanup left.

CelticMinstrel:
Then I'll merge it sometime tomorrow.

PR #5710, opened 2021-04-15
Scope exit utilities

CelticMinstrel:
This is for future use, especially in #5663 - since it wasn't directly related to events, I extracted it to a separate PR.

I'll merge this in a day or two unless there are any comments (for example, maybe someone has a better name than `guard_value`?).

### DIFF
--- a/projectfiles/VC16/wesnoth.vcxproj
+++ b/projectfiles/VC16/wesnoth.vcxproj
@@ -4135,6 +4135,7 @@
     <ClInclude Include="..\..\src\utils\parse_network_address.hpp" />
     <ClInclude Include="..\..\src\utils\reference_counter.hpp" />
     <ClInclude Include="..\..\src\utils\scope_exit.hpp" />
+    <ClInclude Include="..\..\src\utils\guard_value.hpp" />
     <ClInclude Include="..\..\src\variable.hpp" />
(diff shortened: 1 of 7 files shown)


stevecotton:
Aaaaargh I missed an important detail in my suggestions. Splitting the Lua unit test module into a separate PR made sense to me, partly because it had tests that would be included in the PR.

For these scope exit utils, it doesn't feel like a complete PR. There's nothing creating an instance of `guard_value` here, and while ON_SCOPE_EXIT is used I don't see anything that would notice during the CI steps if it didn't work.

CelticMinstrel:
Well, you didn't suggest splitting this out, I just did it on my own. I almost just pushed it to master directly but then suddenly thought maybe there's a better name for `guard_value` and decided to open it as a PR instead.

And I guess it would be pretty easy to add Boost unit tests for these, if we want.

stevecotton:
Discussion of this is happening in https://github.com/wesnoth/wesnoth/pull/5663#discussion_r613075671

IMO it was merged too fast, it's not RAII, and it should be reverted.

CelticMinstrel:
Just because you don't understand its design and use doesn't mean it should be reverted. How the heck can it make dangling pointers more likely? There's no chance of a dangling pointers with its use in #5663.

PR #5713, opened 2021-04-15
Fix teaching_anim not implmmented in variation_id=sceptre of Httt Princess and Battle Princess

newfrenchy83:
I have forget to pull this in variation scpeter of unit_type


stevecotton:
Code looks good to me, and it works in game with the leading animation being used instead if the supported unit is of lower level. (The leading animation has a shiny spark, the initiative one doesn't).

The Battle Princess' animation for both leading and initiative looks odd to me, too much waving the sword about, but that's not caused by this PR. I'd approve, but I think I've the wrong person to do the review.

doofus-01:
I'll look at it this weekend.

Pentarctagon:
@doofus-01 were you able to take a look at this?

doofus-01:
Sorry, I'd forgotten about this, but looking at it now.

doofus-01 (approved):
It looks like it should be fine.  It would be better if there were additional animation frames, both leading/initiative and regular/sceptre, but that can be a separate issue.

PR #5716, opened 2021-04-17
Xcode: Disable hardened runtime on debug builds

CelticMinstrel:
These are the changes I needed to make to the project in order to build. I'm open to suggestions if there's a better way.


hrubymar10:
macOS 11 Big Sur at least Apple Silicon versions requires all apps to be signed by at least Ad Hoc self signed certificate. That's the reason why I can't remove code signing from Xcode project.

Please try use latest Xcode 12 and macOS 11 Big Sur and report back if project builds for you.

hrubymar10 (changes requested):
https://github.com/wesnoth/wesnoth/pull/5716#issuecomment-821990489

CelticMinstrel:
There's probably a way to use - instead of blank for the code signing and still have it work; I think I've seen it work on other projects. I'll investigate when I have time.

I _am_ using Xcode 12 for the record, but I'm on Catalina still.

CelticMinstrel:
@hrubymar10 Okay, this works after all.

It seems the real issue with the code signing is that the option itself is buggy; it _should_ trigger a full rebuild if you change it, but instead it just gives an error unless I do a clean build.

CelticMinstrel:
Further data: it seems that the WML unit test system does not work with ad-hoc code signing enabled, so we probably _do_ need to either disable it for debug builds, or find a way to allow the signed executable to be launched as a command-line tool. (That might even help with the Boost unit tests too?)

hrubymar10:
> Further data: it seems that the WML unit test system does not work with ad-hoc code signing enabled, so we probably _do_ need to either disable it for debug builds, or find a way to allow the signed executable to be launched as a command-line tool. (That might even help with the Boost unit tests too?)

what does ti do or what it doesn't? I would say that it doesn't work because of bad rpaths instead of codesigning problems.

CelticMinstrel:
That's possible. When I run the WML unit tests using ad-hoc signing, every run crashes instantly. I can try running a single instance later to see if there's any output that the WML test runner is suppressing, but it's _supposed_ to include the output if a run crashes…

hrubymar10:
Fixed by 4832912cca79b264061d23af53d3f487219352af

PR #5717, opened 2021-04-18
Fix save_index issues

### LINKED ISSUE #5702: Improve autosave performance
a2c795bfd7a22c6196f2b5bd1386edaf1523e541 on Linux

While playing Wesnoth I've noticed that after the end of a turn game display remains frozen for about 2-3 seconds.
I was suspecting that this lag is caused by the autosave feature, so I disabled autosaves and the lag at the end of a turn disappeared.
I measured how much time is spend in `autosave_savegame::autosave()` and in `write_save_index()` (gzip compression was enabled):

~~~
20210413 02:11:10 info engine: Milliseconds to write_save_index(): 625
20210413 02:11:10 info engine: Milliseconds to write_save_index(): 629
20210413 02:11:10 info engine: Milliseconds to save DiD-Endless Night-Auto-Save30.gz: 1369
20210413 02:11:10 info engine: Deleting savegame 'DiD-Endless Night-Auto-Save24.gz'
20210413 02:11:11 info engine: Milliseconds to write_save_index(): 634
20210413 02:11:11 info engine: Milliseconds to autosave: 2008
~~~

`sum of time in write_save_index() = 1888 ms (avg 629 ms)`
`autosave time - write_save_index() = 120 ms`

If four `write_save_index()` calls could be replaced with only one at the end of `autosave_savegame::autosave()` then autosave time is reduced to:
`120 ms + 629 ms = 749 ms`
This is only 37% of the original autosave time.

In short, this feature request is about reducing the number of intermediate writes of the save index to the disk.  This would result in improved autosave performance. 

EDIT: please read comment 3.

grz0:
This pull request fixes some issues related to save_index:
* fixes a bug in save_index_class::remove() method.
  The method incorrectly tries to remove an attribute from root config instead of removing 'save' children.
  In effect, save files were not being removed from the save_index file.
* removes call to save_index_manager_->remove() in savegame::finish_save_game
  This call to remove() is incorrect and has no effect except causing unneeded call to write_save_index().
* introduces a new clean_up_index() method in save_index_class.
  The method removes non-existing save files from the index. Clean up is triggered only during the first
  synchronisation of the index to disc and only if the index has more save files than found on disk.

Fixes  #5702.

### DIFF
--- a/src/save_index.cpp
+++ b/src/save_index.cpp
@@ -70,7 +70,7 @@ void save_index_class::rebuild(const std::string& name, const std::time_t& modif
 void save_index_class::remove(const std::string& name)
 {
 	config& root = data();
-	root.remove_attribute(name);
+	root.remove_children("save", [&name](const config& d) { return name == d["save"]; });
 	write_save_index();
 }
 
@@ -97,6 +97,22 @@ const std::string& save_index_class::dir() const
 	return dir_;
 }
 
+void save_index_class::clean_up_index()
+{
+	config &root = data();
+
+	std::vector<std::string> filenames;
+	filesystem::get_files_in_dir(dir(), &filenames);
+
+	if(root.all_children_count() > filenames.size()) {
+		root.remove_children("save", [&filenames](const config& d)
+			{
+				return std::find(filenames.begin(), filenames.end(), d["save"]) == filenames.end();
+			}
+		);
+	}
+}
+
 void save_index_class::write_save_index()
 {
 	log_scope("write_save_index()");
@@ -104,6 +120,11 @@ void save_index_class::write_save_index()
 	if(read_only_) {
 		LOG_SAVE << "no-op: read_only instance";
 		return;
+	}
+
+	if(clean_up_index_) {
+		clean_up_index();
+		clean_up_index_ = false;
 	}
 
 	try {
@@ -126,6 +147,7 @@ save_index_class::save_index_class(const std::string& dir)
 	, modified_()
 	, dir_(dir)
 	, read_only_(true)
+	, clean_up_index_(true)
 {
 }
 
--- a/src/save_index.hpp
+++ b/src/save_index.hpp
@@ -124,6 +124,8 @@ private:
 	config& data();
 
 	static void fix_leader_image_path(config& data);
+	/** Deletes non-existent save files from the index. */
+	void clean_up_index();
 
 	bool loaded_;
 	config data_;
@@ -134,5 +136,7 @@ private:
 	 * write_save_index() and delete() are no-ops.
 	 */
 	bool read_only_;
+	/** Flag to only run the clean_up_index method once. */
+	bool clean_up_index_;
 };
 } // end of namespace savegame
--- a/src/savegame.cpp
+++ b/src/savegame.cpp
@@ -541,8 +541,6 @@ void savegame::finish_save_game(const config_writer& out)
 		if(!out.good()) {
 			throw game::save_game_failed(_("Could not write to file"));
 		}
-
-		save_index_manager_->remove(gamestate_.classification().label);
 	} catch(const filesystem::io_exception& e) {
 		throw game::save_game_failed(e.what());
 	}


stevecotton:
I'm not sure if this PR is right or wrong, but the call stack worries me. A call to `write_game_to_disk(filename="HttT-The Bay of Pearls-Auto-Save1.gz"`, results in a call to `remove(name="HttT-The Bay of Pearls")`, which suggests it will either hit nothing or it will hit more indices than expected.

```
#0  savegame::save_index_class::remove(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    (this=0x555558e1da10, name="HttT-The Bay of Pearls") at <src_tree>/src/save_index.cpp:71
#1  0x0000555555a40320 in savegame::savegame::finish_save_game(config_writer const&) (this=0x7fffffffb6a0, out=...)
    at <src_tree>/src/savegame.cpp:545
#2  0x0000555555a3fddd in savegame::savegame::write_game_to_disk(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    (this=0x7fffffffb6a0, filename="HttT-The Bay of Pearls-Auto-Save1.gz") at <src_tree>/src/savegame.cpp:515
#3  0x0000555555a3f7ca in savegame::savegame::save_game(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
    (this=0x7fffffffb6a0, filename="") at <src_tree>/src/savegame.cpp:471
#4  0x0000555555a3eff2 in savegame::savegame::save_game_automatic(bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (this=0x7fffffffb6a0, ask_for_overwrite=false, filename="") at <src_tree>/src/savegame.cpp:381
#5  0x0000555555a40d9e in savegame::autosave_savegame::autosave(bool, int, int)
    (this=0x7fffffffb6a0, disable_autosave=false, autosave_max=10, infinite_autosaves=61) at <src_tree>/src/savegame.cpp:616
#6  0x00005555561afc63 in playsingle_controller::before_human_turn() (this=0x7fffffffc450)
    at <src_tree>/src/playsingle_controller.cpp:456
```

grz0:
I can't reproduce this crash.  `savegame::savegame::finish_save_game()` calls  `save_index_class::remove()` with `HttT-The Bay of Pearls` as argument but game does not crash.

I think that this call to `remove()` is incorrect:
`save_index_manager_->remove(gamestate_.classification().label);`
What should be the result of this call? Should it remove all saves with a given label from the index?
I'm guessing that the intention of this code was to trigger a rebuild of the index if a new save file has the same name as the old one. If I'm right then we can remove this line because `savegame::save_game()` already called `save_index_manager_->rebuild(filename_)`.

stevecotton:
Sorry for the confusion - it didn't crash, it's just a stack trace from a debugging breakpoint.

grz0:
OK. I will remove this call in the next commit, it is incorrect and has no effect except triggering unneeded `write_save_index()` which increases lag during autosaving.
BTW, I think that this PR should be labelled as a bugfix, not as an enhancement because `save_index_class::remove()` does not work as intended.

src/save_index.hpp
```diff
@@ -124,6 +124,8 @@ class save_index_class : public std::enable_shared_from_this<save_index_class>
 	config& data();
 
 	static void fix_leader_image_path(config& data);
+	/** Deletes non-existent save files from the index **/
+	void clean_up_index(void);
```
Pentarctagon, line 128: What's the difference between `void clean_up_index(void);` and `void clean_up_index();` given this is C++ rather than C?

src/save_index.cpp
```diff
@@ -106,6 +121,11 @@ void save_index_class::write_save_index()
 		return;
 	}
 
+	if(clean_up_index_){
```
stevecotton, line 124: Should have a space between ')' and '{'.

src/save_index.hpp
```diff
@@ -134,5 +136,6 @@ class save_index_class : public std::enable_shared_from_this<save_index_class>
 	 * write_save_index() and delete() are no-ops.
 	 */
 	bool read_only_;
+	bool clean_up_index_;
```
stevecotton, line 139: A comment would be good - while the commit message is good, assume people will later be looking at this code without referring to the Git history.

Maybe `/** Flag to only run the clean_up_index method once */`

src/save_index.hpp
```diff
@@ -124,6 +124,8 @@ class save_index_class : public std::enable_shared_from_this<save_index_class>
 	config& data();
 
 	static void fix_leader_image_path(config& data);
+	/** Deletes non-existent save files from the index. **/
```
stevecotton, line 127: Although Doxygen comments need `/**` at the start, they only need `*/` at the end.

stevecotton (approved):
When force-pushing, please also rebase on the current master and squash the commits.

We can do that in Github when merging, so it's not a big issue, but it's good to do if you're force-pushing anyway.

grz0:
@stevecotton should I change something in this PR? It has been approved by you but it is not merged, so it is not clear to me if we are waiting for more approvals or this PR requires changes.

stevecotton:
@Pentarctagon, do you have any more comments? I'm happy with this merging now.

Pentarctagon:
The doxygen style comment was never addressed.

grz0:
I fixed doxygen comments. I also rebased and squashed commits as suggested previously.

Pentarctagon:
Given that this says it's improving performance, has a test been done to measure the performance difference with this code change?

grz0:
> Given that this says it's improving performance, has a test been done to measure the performance difference with this code change?
I measured how long it takes to execute autosave_savegame::autosave() function.

Test with my original save_index file, enabled gzip compression and without patch:
`20210421 21:43:45 info engine: Milliseconds to autosave: 3494`

The size of this file is 37 megabytes, it contains 7940 entries.
This huge size of the save_index files is caused by not working remove() method (save_index file grows infinitely).

Test on patched wesnoth, after save_index auto cleanup, gzip enabled:
`20210421 21:54:56 info engine: Milliseconds to autosave: 311`

The original save_index file has been reduced to 4 megabytes and 1010 entries.
It takes about 7500 milliseconds to clean up this huge save_index, but this needs to be done only once.

So, 3494ms vs 311ms.

The above was tested on wesnoth-debug. In #5702 I measured that it takes about 2000 milliseconds to complete autosave on release build and original save_index.

Pentarctagon:
Alright, sounds good to me then.

Pentarctagon (approved):

Issue #4436, opened 2019-10-08
After 1.15.15, check for `%d` or `%s` in translated strings


stevecotton:
Sounds like we should collect somewhere the stuff to tell translators in the next announcement: there's also ...
_Originally posted by @jostephd in https://github.com/wesnoth/wesnoth/issues/4165#issuecomment-539296026_

- [ ] #4165 strings changing from `%d` to `$percent|` and `$turns_remaining`
- [ ] #4390 Replace hot-key markers with initials #4390 
- [ ] Roadmap for the "Irdya Canon" rewrite

sevu:
- [ ] 0dc80ecb5298 Added to several message . This messages appear in this tiny window which is shown when e.g. you can't recall units.

Wedge009:
I don't remember if we need to make a note regarding the Special Notes bullet points. @CelticMinstrel did #4350 so maybe it's okay.

jostephd:
I think we do, since #4297 is outstanding.

Wedge009:
Request users to use 'right quote' instead of 'straight quote' for apostrophes, where possible. Mark-up formatting like in #4507 breaks when 'straight quotes' are used.

jostephd:
Can the French translation use `text="foo'bar"` with double instead of single quotes?

Wedge009:
Not sure what you mean by that, but I think no. The issue is not specific to French - I suppose it could even happen in English (but I haven't seen any instances of that yet).

jostephd:
The French translation currently uses `<italic>text='ponts au-dessus d'eau peu profonde'</italic>`. Could it use `<italic>text="ponts au-dessus d'eau peu profonde"</italic>` instead? Note that the double quotes would have to be escaped (with backslashes, I guess) in the po file.

My point is that the three single quotes are a _syntax error_ in the French translation. Yes, there are things Wesnoth could do better at all, see https://github.com/wesnoth/wesnoth/issues/4507#issuecomment-546584629, but this issue is just about what we should tell translators.

CelticMinstrel:
> Note that the double quotes would have to be escaped (with backslashes, I guess) in the po file.

They'd also need to be escaped (doubled) in the WML file.

CelticMinstrel:
I doubt most translators manually edit the po file - the tools they use should escape any quotes for them.

stevecotton:
[...]

PR #5720, opened 2021-04-20
UtBS: Use $amount instead of %d in the teaching ability

stevecotton:
Part of #4436


CelticMinstrel (approved):
There wasn't really a need to split it into multiple lines, but it doesn't hurt either…

PR #5723, opened 2021-04-20
Multiplayer turns over dialog: stop using %d and %s in translatable strings

### LINKED ISSUE #4274: Translatable string "Sides %s and %d are tied." has hardcoded conjunct list
Low-priority, as having a match end in a 3-way tie sounds a rare occurrence; the reason for logging the issue is mainly that having the heterogeneous `%s and %d` looked wrong.

https://github.com/wesnoth/wesnoth/blob/b4c65f17964617f9d5fba84117fd9cc927a25682/data/multiplayer/eras.lua#L93

If sides 1 and 2 tie, then this will have "1" in the %s and 2 in the %d.
If sides 1, 2 and 3 tie, then this will have "1, 2" in the %s and 3 in the %d.

So the translation doesn't know how many sides there are, nor can it do a non-English-style format_conjunct_list.

stevecotton:
This should probably be changed into a table of GUI2 widgets, but this commit
is an improvement on what was here before. It removes the broken conjunct list
formatting for 3-way ties. However, it also means that there are 5 lines per
language to translate again.

Fixes #4274.

### DIFF
--- a/data/multiplayer/eras.lua
+++ b/data/multiplayer/eras.lua
@@ -42,25 +42,20 @@ res.turns_over_advantage = function()
 	local income_factor = 5
 
 	local winning_sides = {}
-	local tie = true
 	local total_score = -1
 	local side_comparison = ""
-	local color = "#000000"
+	local winners_color = "#000000"
 	for side, team in all_sides() do
 		if not team.__cfg.hidden then
-			local r, g, b = 255, 255, 255
-			if     team.__cfg.color == 1 then r, g, b = 255,   0,   0
-			elseif team.__cfg.color == 2 then r, g, b =   0,   0, 255
-			elseif team.__cfg.color == 3 then r, g, b =   0, 255,   0
-			elseif team.__cfg.color == 4 then r, g, b = 155,  48, 255
-			elseif team.__cfg.color == 5 then r, g, b =   0,   0,   0
-			elseif team.__cfg.color == 6 then r, g, b = 165,  42,  42
-			elseif team.__cfg.color == 7 then r, g, b = 255, 165,   0
-			elseif team.__cfg.color == 8 then r, g, b = 255, 255, 255
-			elseif team.__cfg.color == 9 then r, g, b =   0, 128, 128 end
+			-- The translatable strings include support for coloring the side names, but here we just use white. The team colors accessible to Lua
+			-- could be "red", "blue", etc (which are also supported by Pango's markup), but could also be user-defined names from a color changer mod.
+			-- Logged as bug #5722.
+			local side_color = "#ffffff"
 			if # wesnoth.units.find_on_map( { side = side } ) == 0 then
-				side_comparison = side_comparison .. string.format( tostring( _ "<span strikethrough='true' foreground='#%02x%02x%02x'>Side %d</span>") .. "\n",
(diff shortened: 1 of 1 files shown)


data/multiplayer/eras.lua
```diff
@@ -42,25 +42,20 @@ res.turns_over_advantage = function()
...
-			elseif team.__cfg.color == 4 then r, g, b = 155,  48, 255
-			elseif team.__cfg.color == 5 then r, g, b =   0,   0,   0
-			elseif team.__cfg.color == 6 then r, g, b = 165,  42,  42
-			elseif team.__cfg.color == 7 then r, g, b = 255, 165,   0
-			elseif team.__cfg.color == 8 then r, g, b = 255, 255, 255
-			elseif team.__cfg.color == 9 then r, g, b =   0, 128, 128 end
+			-- The translatable strings include support for coloring the side names, but here we just use white. The team colors accessible to Lua
+			-- could be "red", "blue", etc (which are also supported by Pango's markup), but could also be user-defined names from a color changer mod.
+			-- Logged as bug #5722.
+			local side_color = "#ffffff"
```
CelticMinstrel, line 53: You know, the API needed to just directly get the colour already exists…

```lua
local side_color = string.format("#%02x%02x%02x", table.unpack(wesnoth.colors[team.color].mid))
```

Or if you uncomment https://github.com/wesnoth/wesnoth/tree/master/src/scripting/lua_color.cpp#L121-L125:

```lua
local side_color = wesnoth.colors[team.color].pango_hex
```

data/multiplayer/eras.lua
```diff
@@ -73,26 +68,36 @@ res.turns_over_advantage = function()
...
+		local comparison_text = _ "Sides $side_number and $other_side_number are tied."
+		side_comparison = side_comparison .. "\n" .. comparison_text:vformat{side_number = winning_sides[1], other_side_number = winning_sides[2]}
 	else
-		side_comparison = side_comparison .. string.format( "\n" .. tostring( _ "<span foreground='%s'>Side %d</span> has the advantage."), color, winning_sides[1])
+		local winners = stringx.format_conjunct_list("", winning_sides)
+		-- po: In the end-of-match summary, three or more teams have all tied for the best score. $winners contains the result of formatting the conjunct list.
+		local comparison_text = _ "Sides $winners are tied."
+		side_comparison = side_comparison .. "\n" .. comparison_text:vformat{winners = winners}
 	end
 	wesnoth.fire("message", { message = side_comparison, speaker = "narrator", image = "wesnoth-icon.png"})
```
CelticMinstrel, line 102: Maybe this is out-of-scope for this PR, but I _really_ think this should use `gui.show_popup` instead of the current `gui.show_narration`.

data/multiplayer/eras.lua
stevecotton: Thanks. Using `pango_hex` looks like a nice encapsulation, but I think it should be done in a separate PR.

data/multiplayer/eras.lua
stevecotton: Once the CI is running again I'll push an update with this.

CelticMinstrel (approved):

PR #5725, opened 2021-04-20
fix text don't readable because duration not specified

newfrenchy83:
When a enemy is killed in 23_Test_of_the_Clans.cfg of Httt, the show of number to enmy to vanquish is too breve for be read.


nemaara:
Make it 1.5 or 2 seconds please.

newfrenchy83:
> 
> 
> Make it 1.5 or 2 seconds please.

Done

nemaara (approved):

PR #5726, opened 2021-04-21
Add IntroWML support for text labels on maps

stevecotton:
Opening as a draft to discuss whether this feature should be included in 1.16, the roadmap issue for it is #5041.

My plan is to merge the API change (the first commit) in time for 1.15.13, which would allow UMC to hand-code labels using it. The trackplacer change still needs work, but changes to that tool could still be done in the 1.16 branch. Making a journey in the current version of trackplacer will give the right coordinates to cut & paste for the label.

This adds a `[story][part][image]label=` attribute, allowing the `[image]` tag
to draw either images, text, or both. Putting both functionalities on the same
tag has tradeoffs - the tag name is now a bit inaccurate, however it allows
having landmarks with captions on the map such as

	[image]
		x,y=1030,391
		file="map_icons/dwarven_door.png"
		label=_ "map_label^Kal Kartha"
	[/image]

For the label, the x and y coordinates aren't the exact placement of the text,
instead they're the location of the place that the text should label.

I'm not expecting the IntroWML labels to reach mainline campaign quality, they're not going to be anywhere near a hand-kerned text by an artist fitting the lettering to the features of the map. I'm hoping that they'd be used by UMC.

Example image - Dead Water's map using only WML labels
---

The italic 'd' on 'Bildheld' gets cut, which is probably a bug in pango_text's width calculation. That's a problem which I would expect to fix as part of this.
wesnoth_introwml_dw_map
The text scales with the image size. I've cropped the edges off the larger screenshot.


CelticMinstrel:
This needs a way to draw outlined text to guarantee legibility.

stevecotton:
> This needs a way to draw outlined text to guarantee legibility.

Agreed, but that's going to get this involved in the queue of PRs changing `src/font/text.*pp` and `src/gui/core/canvas*pp`.

CelticMinstrel:
Do you have outlined text in that queue? If it's in the works, I think it's okay to merge this before that's ready.

src/gui/core/canvas_private.hpp
```diff
@@ -446,6 +447,9 @@ class text_shape : public canvas::shape {
 
 	/** The maximum height for the text. */
 	typed_formula<int> maximum_height_;
+
+	/** Whether to draw a border around the text. */
+	typed_formula<bool> outline_;
```
CelticMinstrel, line 452: I think it may make more sense for this to be an integer (representing the thickness of the outline)? I'm not sure _how_ useful it would be though…

stevecotton:
I might reintroduce this once the 1.18 development cycle starts, but feel it's too late for 1.16.

* Unless each letter can be hand-aligned by an artist then this won't have the quality for mainline campaigns' maps.
* Campaigns can already put overlay images on maps (this is how the journey-tracker's tracks work). The feature would make them easier to translate, but that only matters if the campaign's locations have different names in different languages. Probably not going to be an issue for many UMC campaigns.
* The text outlines need more work, and discussion about how their width and color are controlled. If it was merged now, we'd probably want to change the API in 1.18.

CelticMinstrel:
Even if this isn't going to be in 1.16, why does that mean it needs to be closed? Would the implementation have to drastically change in the new version?

stevecotton:
I closed the PR to tidy up the waiting-for-review PR list. I think it's useful to keep that list as small as possible, at least while 1.16's beta testing phase is happening.

CelticMinstrel:
It's been three years, any plans to revisit this?

stevecotton:
I've put in to the roadmap for March (deliberately not putting many commitments on it this time).

PR #5734, opened 2021-04-24
Don't check SDL version on each surface::free_surface() call

grz0:
While running Wesnoth under profiler I've noticed that  `free_surface()` always checks and compares SDL version to decide if
a workaround for SDL2.0.6 bug should be applied. This results in dynamic memory (de)allocations.
This check can be done only once and stored in a static variable, thus removing these memory allocations from the function which is called frequently in the rendering path.

Alternatively, the minimal SDL2 version could be bumped to 2.0.7, this would also allow removing some compile-time checks from the code.

### DIFF
--- a/src/sdl/surface.cpp
+++ b/src/sdl/surface.cpp
@@ -24,6 +24,10 @@ const SDL_PixelFormat surface::neutral_pixel_format = []() {
 #endif
 }();
 
+const bool surface::sdl_freesurface_bug_ = []() {
+	return sdl_get_version() == version_info(2, 0, 6);
+}();
+
 surface::surface(SDL_Surface* surf)
 	: surface_(surf)
 {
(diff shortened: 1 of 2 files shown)


shikadiqueen:
The code seems good to me. Not sure if @Pentarctagon or anyone else has an opinion on the minimum SDL version requirement. I brought up on Discord that it's suspiciously low for software that has seen several other things (compilers and Boost) have their requirements bumped to relatively recent versions.

Pentarctagon:
Ubuntu 18.04 has 2.0.8, so I don't think there'd be an issue with increasing the minimum required version.  Every platform we distribute Wesnoth on ourselves already uses a much higher version as well.

grz0:
OK, so I'm closing this PR and I will open a new issue for SDL version bump.

shikadiqueen:
Out of curiosity, would you be able to provide specific numbers for Wesnoth's performance both before and after getting rid of the SDL version check? It would be nice to have this information available for future reference.

grz0:
> Out of curiosity, would you be able to provide specific numbers for Wesnoth's performance both before and after getting rid of the SDL version check? It would be nice to have this information available for future reference.

`perf top` reports ~0.40% overhead when the game is idle. I don't know if this has any noticeable impact, but it was easy to get rid of this function call so I created this PR.
image

PR #5736, opened 2021-04-24
Bump the minimum SDL2 version to 2.0.8 and remove workarounds which are no longer required

### LINKED ISSUE #5735: Bump the minimum SDL version for 1.15.x
As per discussion in #5734  - Wesnoth requires a pretty low version of SDL in comparison to required Boost and compilers versions.
Bumping the minimum SDL version would allow removing some workarounds for SDL bugs from the code.

grz0:
This PR bumps the minimum SDL2 version to 2.0.8.
Now we can remove:
* workarounds for a bug in SDL_FreeSurface() on SDL2.0.6
* surface::clear_without_free() method, now it has no users
* compile-time checks for SDL_CreateRGBSurfaceWithFormatFrom() availability
* a workaround for pre-SDL2.0.5 bug in SDL_BlitSurface()

`CVideo::~CVideo()` had the following version check:
`sdl_get_version() >= version_info(2, 0, 6)`
but it seems to me that it was incorrect, it should be `==`.
I think this is a workaround for the same issue as in `surface::free_surface()`, so I removed it.

Closes #5735


Pentarctagon:
This looks alright to me.  @Vultraz / @shikadiqueen any thoughts?

grz0:
I squashed some commits in this PR and reworded messages. No other changes.

grz0:
@Pentarctagon @shikadiqueen I have found yet another SDL2 workaround which probably could be removed.

in `display::scroll()`:
```
		// This is a workaround for a SDL2 bug when blitting on overlapping surfaces. The bug
		// only strikes during scrolling, but will then duplicate textures across the entire map.
		surface screen_copy = screen.clone();
```
This bug seems to be fixed in SDL2.0.5:
https://github.com/libsdl-org/SDL/commit/aae28e3ec15b44b285af952075dbc9c1993eb54f

call to `screen.clone()` results in memory (de)allocation and software blitting inside SDL.
I recompiled Wesnoth without this workaround and I don't see any issues while scrolling and CPU usage seems to be a bit lower.

Pentarctagon:
You're referring to https://github.com/wesnoth/wesnoth/blob/e74abf05676f3346b80bd914d81844133b6d85de/src/display.cpp#L1956-L1974 correct?

grz0:
Yes, I'm referring to this code. This looks like a workaround for the SDL2 bug which I mentioned earlier.

Pentarctagon:
It looks like that copying was introduced in 1c17ecaa753753db71f6f26d2d6c05a1e20f83d3 as part of the move from SDL 1.2 to SDL 2.  I'd be fine with removing it as well, as long as scrolling does still work as expected meaning: scrolling with a scrollbar in the game UI, in WML [message]s, and on a game map.

shikadiqueen:
Code seems okay. I haven't tested it yet though.

Removing the cloning of the screen framebuffer seems like a potentially quite impactful change to performance. There is a colossal memory block hiding behind the `SDL_Surface` after all.

grz0:
OK, so I will update this PR with a commit that removes copying the screen's surface.
I'm not sure yet what to do with `SDL_SetSurfaceBlendMode()`, probably it can be also removed.

> There is a colossal memory block hiding behind the SDL_Surface after all.
In my case this results in allocation of 8M memory block.

grz0:
I pushed a new commit that removes screen surface copying. 
`SDL_SetSurfaceBlendMode()` can't be removed, we need to set `SDL_BLENDMODE_NONE` and restore the previous `SDL_BLENDMODE_BLEND` blending mode. Otherwise, there are artefacts caused by blending.

shikadiqueen:
No problems here (SDL 2.0.14).

By the way for @Pentarctagon's reference, the impacted code path is **only** used when scrolling the main gamemap viewport, not for any other kind of scrolling.

grz0:
Thanks for testing.
I don't see any other workarounds which could be removed, so I don't plan to make any further changes to this PR.

Pentarctagon:
This is good to be merged then?

stevecotton:
The Windows build failures are #5741, and are unrelated to this. The Flatpak build failure could be a rate-limited download, it's also unrelated to this.

grz0:
I was curious what is the effect of removal of surface copying, so I measured CPU usage while scrolling on 200x200 grass map.
60sec avg CPU usage:
 * ~51% on the previous Flatpak build
 * ~31% on the current Flatpak build

stevecotton:
Just noticed a graphics issue that's not caused by this; mentioning it here as this PR is the one that people will blame first.

When resizing Wesnoth's window, the map gamemap flickers with the existing pixels being laid out with the wrong stride size, it then redraws correctly. This isn't new, it was there in 1.15.12.

shikadiqueen:
> When resizing Wesnoth's window, the map gamemap flickers with the existing pixels being laid out with the wrong stride size, it then redraws correctly. This isn't new, it was there in 1.15.12.
I'm sorry to say but this bug has been around ever since SDL 2 became the default in 1.13.x.

PR #5738, opened 2021-04-25
horse things

doofus-01:
- dunefolk cataphract touchups
 - basic horse
 - White Horse variation
 - zombie horses

Other horse things to come, as time permits.


doofus-01:
> Whitespace and WML indentation check failed! (checkindent)

I'd like to merge this so people test it before we reach RCs, but if I run wmlindent, it breaks the macro for the IPF used in Horse.cfg and Horse_Black.cfg.  Is there a way to escape the CI checks?

doofus-01:
There is more work to do here, but I'm going to commit this now to get new content in sooner rather than later.

PR #5739, opened 2021-04-25
add 'ability_id_active' attribute to filter

newfrenchy83:
until now, only the tpe of ability could be filtered with activity, and id only for unit who have ability, same if she's inactive.

### DIFF
--- a/src/units/filter.cpp
+++ b/src/units/filter.cpp
@@ -428,7 +428,7 @@ void unit_filter_compound::fill(vconfig cfg)
 						const unit_map::const_iterator it = units.find(adjacent[i]);
 						if (it == units.end() || it->incapacitated())
 							continue;
-						if ( &*it == (args.u.shared_from_this()).get() )
+						if (&*it == (args.u.shared_from_this()).get())
 							continue;
 
 						std::vector<ability_match> ability_id_matches_adj;
--- a/src/units/filter.cpp
+++ b/src/units/filter.cpp
@@ -16,6 +16,8 @@
 
 #include "log.hpp"
 
+#include "display.hpp"
+#include "display_context.hpp"
 #include "config.hpp"
 #include "game_data.hpp"
 #include "map/map.hpp"
@@ -288,6 +290,26 @@ void unit_filter_compound::create_attribute(const config::attribute_value v, C c
 	}
 }
 
+namespace {
+
+	struct ability_match
+	{
+		std::string tag_name;
+		const config* cfg;
+	};
+
+	void get_ability_children_id(std::vector<ability_match>& id_result,
+	                           const config& parent, const std::string& id) {
+		for (const config::any_child &sp : parent.all_children_range())
+		{
+			if(sp.cfg["id"] == id) {
+				ability_match special = { sp.key, &sp.cfg };
+				id_result.push_back(special);
+			}
+		}
+	}
+}
+
 void unit_filter_compound::fill(vconfig cfg)
 	{
 		const config& literal = cfg.get_config();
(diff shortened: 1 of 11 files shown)


CelticMinstrel:
Seems fine to me.

Pentarctagon:
A WML unit test for this would be good.

newfrenchy83:
> 
> 
> A WML unit test for this would be good.

Done, i await checking to be complete

CelticMinstrel:
Those unit tests don't look right, and they're definitely misnamed. You're not testing that filters on an event work correctly, you're testing that filters themselves work; so the general logic should be:

1. Create a unit with the ability
2. Assert [have_unit] with the particular filter

And the tests should not have "event" in the name.

newfrenchy83:
> 
> 
> Those unit tests don't look right, and they're definitely misnamed. You're not testing that filters on an event work correctly, you're testing that filters themselves work; so the general logic should be:
> 
>     1. Create a unit with the ability
> 
>     2. Assert [have_unit] with the particular filter
> 
> 
> And the tests should not have "event" in the name.

I fixed this.

data/test/scenarios/events-test_filters.cfg
```diff
@@ -148,6 +148,157 @@
...
+        name=start
+        {VARIABLE triggers 0}
+        [unit]
+            id=alex
+            name=_"Alex"
+            x,y=12,4
+            type=Elvish Archer
+            side=1
+        [/unit]
+        [object]
```
CelticMinstrel, line 162: This needs to have a `[filter]` tag to match the unit it applies to.

You can probably just use `id=alice` or `id=bob`, no need to create a third unit, is there?

data/test/scenarios/events-test_filters.cfg
```diff
@@ -148,6 +148,157 @@
...
+            silent=yes
```
CelticMinstrel, line 234: This also needs a `[filter]`  tag.

CelticMinstrel (commented):
That looks better as a unit test, but now you need to move it to a more appropriate file. This unit test does not belong in events-test_filters.cfg.

Pentarctagon:
Also, for the tests that start with `swarm_disables_upgrades_with_abilities`, please update the description so that they say what they're doing rather than being copies of each other.  Either a separate commit here or a separate PR is fine.

newfrenchy83:
> 
> 
> Also, for the tests that start with `swarm_disables_upgrades_with_abilities`, please update the description so that they say what they're doing rather than being copies of each other. Either a separate commit here or a separate PR is fine.

description updated in this PR

CelticMinstrel:
That's better.

newfrenchy83:
@Pentarctagon don't forget this PR please.

src/units/filter.cpp
```diff
@@ -386,6 +408,42 @@ void unit_filter_compound::fill(vconfig cfg)
...
+							return true;
+						}
+					}
+
+					const auto adjacent = get_adjacent_tiles(args.loc);
+					for(unsigned i = 0; i < adjacent.size(); ++i) {
+						const unit_map::const_iterator it = units.find(adjacent[i]);
+						if (it == units.end() || it->incapacitated())
+							continue;
+						if ( &*it == (args.u.shared_from_this()).get() )
```
Pentarctagon, line 431: Please remove the extra spaces, so it's `if (&*it == (args.u.shared_from_this()).get())` to keep this consistent.

newfrenchy83: It done

Issue #5741, opened 2021-04-27
Windows CI builds broken with Pango 1.48


stevecotton:
From the CI log for ac9f6da, which is the last version of master that successfully built on the Windows CI:
```
The following packages will be built and installed:
pango[core]:x64-windows -> 1.40.11#9
```

From more recent builds:
```
The following packages will be built and installed:
pango[core]:x64-windows -> 1.48.0
...
fatal error C1083: Cannot open include file: 'pango/pango.h': No such file or directory (compiling source file ..\..\src\show_dialog.cpp) [D:\a\wesnoth\wesnoth\projectfiles\VC16\wesnoth.vcxproj]
```

stevecotton:
Presumably a problem with vcpkg, as the timing corresponds with https://github.com/microsoft/vcpkg/commit/546813ae7b9e2873dd3d38e78b27ac5582feae10

Pentarctagon:
Opened https://github.com/microsoft/vcpkg/issues/17537 for this.

stevecotton:
The workaround has been merged, but this issue can stay open as a reminder to revert bfe1f51.

Pentarctagon:
It looks like this will be open for a while, since upstream has no intention of fixing the problem and I don't think anyone currently involved in Wesnoth is knowledgeable of both cmake and Visual Studio.  The hack solution there sounds like even if it does work it would frequently break, given it involves version number specific directories.

Pentarctagon:
For reference as well, 75b73043e3349a02b7aa64371858d8550e4f966b was when MSVC support was previously removed due to cmake generally and MSVC specifically being in poor shape.  Given the MSVC-specific things that *were* there though, maybe there's not a lot that would need to be done to get cmake into a minimal working form.

CelticMinstrel:
Bit late to notice this, but at least one person on the vcpkg issue seems to think it should be fixed…

Pentarctagon:
Though thinking it should be fixed and actually fixing it are very different things.  Plus given that one of the vcpkg devs actively doesn't care if they break the integration, it's pretty likely that this sort of problem would keep happening.

PR #5742, opened 2021-04-27
Keep the Windows CIs on a version of vcpkg that builds Wesnoth

stevecotton:
Quick workaround for #5741, the Windows builds weren't finding Pango's header
files when used with vcpkg version 546813ae7b9e2873dd3d38e78b27ac5582feae10
([glib up to gtk] update and make it work with meson).

Hopefully we can revert this workaround soon.

(Note: don't merge until the CI build has passed)

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -317,7 +317,8 @@ jobs:
       - name: DOS1
         shell: cmd
         run: |
-          git clone --depth=1 https://github.com/microsoft/vcpkg.git ../vcpkg
+          git clone --shallow-since=2021-04-25 https://github.com/microsoft/vcpkg.git ../vcpkg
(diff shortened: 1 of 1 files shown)


stevecotton:
Build results: the windows release build passed, the windows debug build reached the linking phase and then had a error trying to link SDL2_ttf.lib, which I assume will be fixed by rebasing on 3858737ac3684e0656770fabb92724b05d5f74ed.

On both CIs, that means that it found the Pango headers and so this does seem to work around #5742.

Pentarctagon (approved):

stevecotton:
I'm now assuming that the SDL2_ttf problem on windows-release is a09ffa9fd81764032374c224b6ad8c81e515a857. The expectation for the current build is therefore for windows-debug to pass and windows-release to fail; if that happens then I'll just merge.

PR #5744, opened 2021-04-27
Add wesnoth.colors[name].pango_color and use it in the multiplayer turns-over dialog

### LINKED ISSUE #5722: Multiplayer turns over advantage doesn't use team colors
Local builds of 1.14.16 and 1.15.12

### Describe the bug
The multiplayer "turns over advantage" screen is supposed to team-color the sides' names. Instead it shows them all in white.

There's a hardcoded list of the RGB values for colors "1", "2", "3", ..., "9"; however the Lua code is receiving the RGB colors "red", "blue", etc.

### To Reproduce
1. Start a local multiplayer
2. In the create game screen, set the turn limit to 1 turn
3. Start the game
4. End turn for each side
5. A dialog pops up when the turns run out

### Screenshots
This is what it's meant to look like. If you try to reproduce it, you'll find that all the names are white instead.

The teams here are all using their default colors, which have names that are also recognised by Pango's rendering code. Notably Ravana's color-changer mod is not activated, which is why I'm creating a bug report instead of opening a PR with a simple fix.

wesnoth_turns_over_with_color

### Hardcoded list
https://github.com/wesnoth/wesnoth/blob/1b0d63f3057a268b94bfee0fb383a609a8e7a50b/data/multiplayer/eras.lua#L51-L60

stevecotton:
The first of the two commits already has PR #5723, this PR will be for the second one once that merges. It's mainly about the unit test and whether `pango_color` is the right name.

Fixes #5722.

With Ravana's color changer mod:
wesnoth_turns_over_with_popup_color

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,8 @@
  ### Campaigns
  ### Editor
  ### Multiplayer
+ ### Lua API
+   * Added `pango_color` to the `wesnoth.colors` table, for easy use in formatted text.
  ### Packaging
    * Increased minimum required version of SDL to 2.0.8 (PR #5736).
  ### Terrain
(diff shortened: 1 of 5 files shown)


CelticMinstrel (approved):
Looks good to me! It really does look better with `show_popup`, too!

stevecotton:
The Windows build failures were network errors.

PR #5749, opened 2021-04-29
Make command-line option --language never require --all-translations

stevecotton:
When the user supplies an exact locale identifier on the command line,
it doesn't make sense to check whether it's an incomplete translation.


stevecotton:
The two builds that have failed are currently only showing "this check failed" without any logs about why. I suspect those two might be checking Doxygen changes, but I'm going to rerun them and hope they at least give logs.

PR #5752, opened 2021-04-30
Remove the old PNG optimisation script

stevecotton:
A Python version was added in b3fe980d709add4fb8b3fb88faad946bfe55e477, the
main reason for removing the old one is that it's inconsistent to have two
tools where one does almost (but not quite) the same as the other.

Both versions check that the old and new images look the same. However, if any
tool changed the pixels' values then old script would discard the changes from
the other tools too. The new script checks the image after each tool is run, so
that the optimisations from the other tools can still be kept.

The indexed2rgb.sh script had a call to utils/wesnoth-optipng. It's deleted
instead of updated because both wesnoth-optipng itself (since 17c242a9f) and
the new woptipng both do what indexed2rgb.sh did (they all run ImageMagick with
option "png:color-type=6").


stevecotton:
Once the CI has run, I intend to merge this. The force-push is just to put a changelog entry in the same commit.

PR #5753, opened 2021-04-30
Run woptipng on images with bad iCCP, and associated files

stevecotton:
The files that had bad iCC profiles were the following subset. I ran woptipng
on the files that were checked in at the same time too.
* data/core/images/portraits/monsters/woodland_boar.png
* data/core/images/portraits/undead/zombie-boar.png
* data/core/images/terrain/castle/troll/tile.png
* data/core/images/units/human-loyalists/heavyinfantry-attack-8.png
* data/core/images/units/human-outlaws/bandit-idle-1.png
* data/core/images/units/human-outlaws/bandit-idle-2.png
* data/core/images/units/human-outlaws/bandit-idle-3.png
* data/core/images/units/human-outlaws/bandit-idle-4.png
* data/core/images/units/human-outlaws/bandit-idle-5.png
* data/core/images/units/human-outlaws/bandit-idle-6.png
* data/core/images/units/human-outlaws/trapper-bow-attack1.png
* data/core/images/units/human-outlaws/trapper-bow-attack2.png
* data/core/images/units/human-outlaws/trapper-bow-attack3.png
* data/core/images/units/human-outlaws/trapper-bow-attack4.png
* data/core/images/units/human-outlaws/trapper-bow-attack5.png
* data/core/images/units/human-outlaws/trapper-bow-attack6.png
* data/core/images/units/human-outlaws/trapper-bow-attack7.png
* data/core/images/units/human-outlaws/trapper-bow-defend.png
* data/core/images/units/human-outlaws/trapper-bow.png
* data/core/images/units/human-outlaws/trapper.png

PR #5754, opened 2021-04-30
Make woptipng's debug output about bigger images clearer

stevecotton:
If a tool made the image bigger, the debug message was the same as if the tool
had made the image's pixel change. Clarify that into two separate messages.

As each verify_images() call checks that the file became smaller, the removed
print statement would never be reached.

### DIFF
--- a/utils/woptipng.py
+++ b/utils/woptipng.py
@@ -23,6 +23,7 @@
 from multiprocessing import Pool
 import multiprocessing # cpu count
 from PIL import Image as PIL # compare images
+import enum
 import subprocess # launch advdef, optipng, imagemagick
 import os # os rename, niceness
 import shutil # copy files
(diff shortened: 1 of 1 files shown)


stevecotton:
@doofus-01 said in https://github.com/wesnoth/wesnoth/pull/5753#pullrequestreview-650042088
> I'd tried to fix this at one point, but didn't use the right imagemagick. If there are particular woptipng options needed, can you document that somewhere?

Putting this back into draft, to add better documentation about the size threshold.

utils/woptipng.py
```diff
@@ -184,8 +187,6 @@ def optimize_image(image):
         size_delta = size_after - size_initial
         perc_delta = (size_delta/size_initial) *100
```
stevecotton, line 188: This whole function could do with a review and tidy-up. There's a `while` loop that always runs exactly once, and afterwards the conditional statements have a mixture of re-reading `size_after` from the filesystem but using the already-calculated `perc_delta`.

utils/woptipng.py
```diff
@@ -102,7 +95,10 @@ def verify_images(source_img, new_img, transform):
         os.rename(new_img, source_img) # move new image to old image // os.rename(src, dest)
     else: # we can't os.rename(image_after, image_before) because that would leave us with no source for the next transform
         shutil.copy(source_img, new_img) # override new image with old image // shutil.copy(src, dest)
-        debugprint(("TRANSFORMATION unsuccessfull: + " + transform + ", REVERTING " + source_img))
+        if not no_change:
```
soliton-, line 98: I would try to avoid the double negative and switch the if around or invert the `no_change` variable.

utils/woptipng.py
```diff
@@ -154,70 +150,102 @@ def check_progs():
...
-            f.write(initial_file_content)
+    summary_string = None
+    status = None
+    if size_after == size_initial:
+        # probably already optimized with this script
+        status = ProcessingStatus.UNCHANGED
+    elif perc_delta*-1 < THRESHOLD:
+        # changed size, but exceeds the threshold
+        if size_after < size_initial:
+            summary_string = "not replacing {image}, as the reduction was less that the threshold. Changed from {size_initial} to {size_after}, {size_delta}b, {perc_delta}%"
```
soliton-, line 197: that -> than

soliton- (approved):
I'm not particularly familiar with python but it looks good to me.

utils/woptipng.py
stevecotton: Going to rename it `pixels_identical`.

PR #5760, opened 2021-05-03
Rewrite the WML unit test for sighted events

stevecotton:
The old version's [event]name=sighted had a {RETURN} statement in it,
which meant the test stopped on the first sighting. Most of the test
wasn't testing anything.

### DIFF
--- a/data/test/scenarios/sighted_events.cfg
+++ b/data/test/scenarios/sighted_events.cfg
@@ -1,133 +1,176 @@
+# wmllint: no translatables
+
 # This test checks that sighted events are being triggered
-# at the appropriate time. An elvish archer moves towards
-# and discovers an orcish grunt.
-# We pass if the elvish archer is the discoverer, and the
-# variable "sighted_should_trigger" is set.
-# The second test checks what happens if the orc is discovered
-# on the elf's turn.
+# correctly due to units moving - whether that's the viewer
+# moving and revealing hexes in the fog, or whether that's
+# the enemy units moving into the already-revealed hexes.
+[test]
+    name = "Unit Test sighted_on_move"
+    map_file=test/maps/sighted_events.map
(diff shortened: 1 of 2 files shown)


CelticMinstrel:
Assuming that the new test covers everything the old pair of tests did and more, the name you chose for it seems to be rather misleading.

stevecotton:
> Assuming that the new test covers everything the old pair of tests did and more, the name you chose for it seems to be rather misleading.

How so? It's about checking that appearing out of the fog triggers a sighted event, but that movement that doesn't reveal someone out of the fog doesn't trigger the event.

CelticMinstrel:
You're planning to add even more tests, then?

I was just noticing that you removed two tests and added one, so that suggests the new one covers everything the old ones did, but it didn't seem like the name for the new one was general enough to cover what the old ones did.

I think `sighted_on_move` is a good name if it's specifically testing sighting based on movement.

Sighting based on movement is kinda two things though, right? It could happen if someone moves into your field of vision, or if you move and your new field of vision uncovers a new unit. Does it test both of those?

stevecotton:
> You're planning to add even more tests, then?

I'm not committing to doing that, it's only likely if a question comes up that the test would answer. 

> I was just noticing that you removed two tests and added one, so that suggests the new one covers everything the old ones did, but it didn't seem like the name for the new one was general enough to cover what the old ones did.

I think I've covered everything that the old one covered or was intended to cover, but getting a second opinion on that is part of the reason for the review.

> Sighting based on movement is kinda two things though, right? It could happen if someone moves into your field of vision, or if you move and your new field of vision uncovers a new unit. Does it test both of those?

As well as changing the name, the force-push just now improved the comments a bit to clarify that it's checking both of those cases.

Hejnewar (approved):

PR #5762, opened 2021-05-06
Use cmake to create VS project files.

### LINKED ISSUE #5741: Windows CI builds broken with Pango 1.48

Pentarctagon:
Due to an upstream change in vcpkg that breaks the simple integration previously available with `vcpkg integrate install`, building using that setup method is no longer possible. In order to work correctly, cmake must instead be used to generate the VS project files, since that is able to integrate with vcpkg, since vcpkg also uses cmake to build all the library dependencies.

An additional benefit of this is that it will no longer be necessary to separately update the VS project files since it will read the same source_lists files as cmake (on linux) and scons do.

This also enables running the WML unit tests on Windows with this in order to confirm that a valid wesnoth.exe is in fact being generated as well as fixes building the boost unit tests.

The warning level for both release and debug builds are now at level three, the remaining warnings have been fixed, and therefore strict builds have been enabled - any warning will now cause the build to fail, just like for the linux jobs.

Known issues:
* The boost unit tests don't actually run successfully - they fail on CI at least with an exit code on 201 - however I don't know if this is a real problem or just a problem with running headless on CI.
* The debug build doesn't quite work since the executables are built against the non-debug dlls but cmake copies over the debug dlls into the output directory. For now this can be worked around by copying the release dlls into the debug directory.
* The instructions in INSTALL.md are not very good since I don't use Windows and thus can't write anything more detailed. Ideally someone who uses Windows can add more detailed step by step instructions at some point.

Fixes #5741


stevecotton:
Hmm - the answer to "why is config_cache.cpp including an SDL header" is that it uses it for `#ifdef MOUSE_TOUCH_EMULATION`, added in aedc081b2eb2b4e519bc59cc28eeaee27c872a0a. Commenting here because removing that `#include` will silently introduce a bug while appearing to compile successfully`*`.

`*` I haven't tested that compilation, just assuming it'll work.

Pentarctagon:
I'm more confused about why it only fails the scons+clang+debug job and nothing else, but I also don't see why that include is needed in the commit you linked?  What does including `SDL2_platform.h` have to do with
```
#if defined(MOUSE_TOUCH_EMULATION) || defined(__IPHONEOS__)
	target["IPHONEOS"] = preproc_define();
#endif
```
?

stevecotton:
Not sure why it's just that platform, but that header provides a `#define __IPHONEOS__ 1` on the appropriate platform.

Pentarctagon:
It looks like it would be better to just directly check for `TARGET_OS_IPHONE` rather than including the SDL header (which does that check) and then checking for what SDL defines, given there's nothing here that's related to SDL that's used.

PR #5763, opened 2021-05-06
add [immune] abilitiy who protect unit affected again weapon- specials of poison,slow, drains or plague type

newfrenchy83:
An issue for me with the core specials is what it can't be canceled without use a wml lua eencoding, and what in attack_prediction window, specials is showed active. An solution for addon develloper is to create specials similar but active under condition, but for core units, the issue remain. I propose to add new abilities for can canceled specials directly and simplify the work of addon develloper. I hpe what this idea is pertinent.


Pentarctagon:
I suppose I can see the use of this, but it also feels like it could get really confusing.  Maybe others have an opinion on this.

newfrenchy83:
> 
> 
> I suppose I can see the use of this, but it also feels like it could get really confusing. Maybe others have an opinion on this.

change the name of tags could help to had best view?

CelticMinstrel:
I'm not sure protection against status effects needs to be an ability at all. You can already protect against poison just by giving the nonpoisonable status effect (I think that's the correct name?), which can also be done as a trait or object if that's what you want (and doing it as a trait certainly allows the user to be informed of the resistance). Adding a similar case for slow seems alright though.

data/schema/units/abilities.cfg
```diff
@@ -133,10 +144,23 @@
 {BASED_ON_SPECIAL "poison"}
 {BASED_ON_SPECIAL "slow"}
 {BASED_ON_SPECIAL "petrifies"}
+{IMMUNE_TO_SPECIAL "unattacks"}
+{IMMUNE_TO_SPECIAL "unswarm"}
```
stevecotton, line 148: It makes sense for consistency with the other ones, but "immune to swarm" is a bit weird.

Having severed all but one head of the great hydra, the foolish hero threw his shield to the ground and beckoned the stunned beast on. "Come weakling, do not let the legends claim that I merely slew an injured beast. Have five goes at biting me, as if you still had all of your heads."

src/units/abilities.cpp
```diff
@@ -1481,20 +1525,53 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
...
+		return false;
+	}
+	if (tag_name == "berserk" && other &&
+		other->get_ability_bool_anti_weapons("unberserk", other_loc, !is_attacker, other_attack, self_attack)) {
 		return false;
 	}
-	if (tag_name == "firststrike" && !is_attacker && other_attack &&
-		other_attack->has_special_or_ability("firststrike")) {
+	if (tag_name == "firststrike" && !is_attacker && ((other_attack &&
+		other_attack->has_special_or_ability("firststrike")) || (other && other->get_ability_bool_anti_weapons("unfirststrike", other_loc, !is_attacker, other_attack, self_attack)))) {
```
stevecotton, line 1574: I haven't checked if the called functions handle this, but what happens when:
* attacker has firststrike
* defender has firststrike and unfirststrike

I think that should let the defender strike first, but think it's going to fail on the `other_attack->has_special_or_ability("firststrike")` call.

stevecotton:
I feel it's going to be confusing, but I also feel it'll lead to an arms-race of "unblockable super poison" etc.

It must have been discussed in the archives of the ideas forum, it's included in the frequently proposed ideas list. https://forums.wesnoth.org/viewtopic.php?t=34904#w0fpi29

src/units/abilities.cpp
newfrenchy83: if defender and attacker has firsttrike then, the fight played like if none has firststrike, and with unfirsttsrike ability, it's like attacker don't have special, same if defender don't have firststike

newfrenchy83:
> 
> 
> I feel it's going to be confusing, but I also feel it'll lead to an arms-race of "unblockable super poison" etc.
> 
> It must have been discussed in the archives of the ideas forum, it's included in the frequently proposed ideas list. https://forums.wesnoth.org/viewtopic.php?t=34904#w0fpi29

ok, i abandon this then.

stevecotton:
Sorry, I may have been too quick to link to the frequently-proposed ideas. It seems that one was last discussed around 10 years ago, so leaving this PR open to find out the opinions of the current development team is probably worthwhile.

CelticMinstrel:
Okay, so I just read the commit message and didn't look at the details, and assumed this was about immunity to status effects because the commit message only mentioned slows and poison.

For immunity against abilities, I think it's even less advisable. First of all, what does that even _mean_? Immunity to status effects (slows, poison, petrify) makes sense, but doing it as three separate abilities doesn't. (And immunity to poison is already possible.) Immunity to drain makes sense… and is also already possible. Immunity to plague makes sense… I'm not sure if this one is currently possible.

But the others? What does it mean to be immune to a "damage" or "attacks" ability? What does it mean to be immune to a "chance_to_hit" ability? What does it mean to be immune to a "swarms" or "berserk" or "firststrike" ability? If it's just circumstances that means the ability won't apply to a particular target (here I'm assuming an ability that targets others, not the unit itself), then that's _also_ already possible, isn't it? Just write the filter on the ability to exclude the required circumstance.

In short… I think everything this purports to do is already possible, so I'm not seeing the point of the PR.

If you disagree that it's already possible, perhaps you can enlighten me.

newfrenchy83:
> 
> There is _one_ thing that I can think of that might make sense in this vein:
> 
> ```
> [immune]
>     status=fear
>     # other ability tags and keys
> [/immune]
> ```
> 
> When anything tries to add the "fear" status to a unit affected by this ability, the status is not added.

I's done for offical satus(poison,slow,drains,plague and petrifies) but not for extra status.

changelog.md
```diff
@@ -18,6 +18,7 @@
  ### WML Engine
    * Modify implementation of overwrite_specials attribute for replace yes/no parameter by none/one_side/both_sides and select abilities used like weapons and specials who must be overwrited(owned by fighter where special applied or both)
    * Add a 'ability_id_active' attribute to [filter]
+   * Add [immune] abilities who have similar effect what undrainable or unpoisonable status, and aplied also on slow plague or petrifies.
```
CelticMinstrel, line 21: Typo: "applied" has two Ps. Also, use "which" instead of "who" and replace "what" with either "to" or "as", and a comma after "slow".

CelticMinstrel:
So, it seems like you completely missed the point of my previous comment… I do think bundling immunity into a single ability is better than having each one with its own anti-ability, but you're still using the same ability-based approach to immunity.

> use immune to fear is not possible, because the weapon specials must be using wml or lua code for add 'fear' status

Immunity to a status effect logically means that whenever the status effect is applied, by any means, your immunity is checked and if you're immune it decides not to apply it. This approach should work equally well for built-in statuses (poison, slows, petrified) as it does for custom statuses such as fear or stunned.

But you do make a point that there's no weapon special that actually applies a status… but immunity to a status doesn't explicitly require that, either.

-----

Question: Is it possible/plausible to handle an immunity in the `unit::set_state` function instead of in the abilities system? That would allow it to work for custom states as easily as for built-in states, although it would obviously not work for plague or drain immunity.

I think _maybe_ plague immunity should actually be a separate tag; I'm not sure what to call it (my best idea currently is `[resist_plague]` or `[plague_resist]`), but I think it should take a list of unit types inside it. This allows creating an ability that makes you immune to some types of plague but not others, so it's an improvement on the existing state where plague immunity is all-encompassing. Another idea in this vein is that the resistance might not be absolute - it could grant a 60% chance of resisting plague, for example.

Since drain immunity is already possible, I think it should also be removed from this, unless you can think of a reason for it to be a separate tag (similar to the plague idea above). I suppose the "chance to resist" _might_ be enough reason to keep it, but I'm not sure.

Note: This is not a request for changes; it is only opening discussion. Please let us discuss how this should look before making any further changes to this pull request. There are other possibilities that I haven't mentioned, and it would be a waste of your time if you implemented this version and then we decided a different version was better, right?

newfrenchy83:
I already knew what added this type of feature will be complicated, create immune to status change will be more difficult and willbe object of anoteher PR. This time a close this PR and be back or not if i find solution.

PR #5765, opened 2021-05-07
Filtering by tags in the add-on manager

stevecotton:
Add a new drop-down filter to the add-ons manager, based on PblWML's
`tags` attribute.

The existing PblWML specification includes a comma-separated list of tags.
Until now there are no official tag names, but some of the 1.14 add-ons already
have some unofficial ones. If this feature is included in 1.16 then I assume
many UMC authors will use the newly-official tags.

For a larger add-on list than the dev server has, use port 15014 which will
connect to the 1.14 server instead (in the connect dialog change the address to
addons.wesnoth.org:15014).

### DIFF
--- a/data/gui/window/addon_manager.cfg
+++ b/data/gui/window/addon_manager.cfg
@@ -718,6 +718,32 @@
 
 				[label]
 					definition = "default"
+					label = _ "Tags:"
+				[/label]
+
+			[/column]
+
+			[column]
+				grow_factor = 0
+				border = "all"
+				border_size = 5
+
+				horizontal_alignment = "left"
+
+				[multimenu_button]
+					id = "tag_filter"
+					definition = "default"
+				[/multimenu_button]
+			[/column]
+
+			[column]
+				grow_factor = 0
+				border = "all"
+				border_size = 5
+				horizontal_alignment = "left"
+
+				[label]
+					definition = "default"
 					label = _ "Order:"
 				[/label]
 			[/column]
--- a/src/gui/dialogs/addon/manager.cpp
+++ b/src/gui/dialogs/addon/manager.cpp
@@ -206,6 +206,40 @@ const std::vector<addon_manager::addon_order> addon_manager::all_orders_{
 	[](const addon_info& a, const addon_info& b) { return a.created > b.created; }}
 };
 
+namespace
+{
+struct addon_tag
+{
+	/** Text to match against addon_info.tags() */
+	std::string id;
+	/** What to show in the filter's drop-down list */
+	std::string label;
+	/** Shown when hovering over an entry in the filter's drop-down list */
+	std::string tooltip;
+};
+
+const std::vector<addon_tag> tag_filter_types_{
+	{"cooperative", N_("addon_tag^Cooperative"),
+		// TRANSLATORS: tooltip in the drop-down menu for filtering add-ons
+		N_("addon_tag^All human players are on the same team, versus the AI")},
+	{"cosmetic", N_("addon_tag^Cosmetic"),
+		// TRANSLATORS: tooltip in the drop-down menu for filtering add-ons
+		N_("addon_tag^These make the game look different, without changing gameplay")},
+	{"difficulty", N_("addon_tag^Difficulty"),
+		// TRANSLATORS: tooltip in the drop-down menu for filtering add-ons
+		N_("addon_tag^Can make campaigns easier or harder")},
+	{"rng", N_("addon_tag^RNG"),
+		// TRANSLATORS: tooltip in the drop-down menu for filtering add-ons
+		N_("addon_tag^Modify the randomness in the combat mechanics, or remove it entirely")},
+	{"survival", N_("addon_tag^Survival"),
+		// TRANSLATORS: tooltip in the drop-down menu for filtering add-ons
+		N_("addon_tag^Fight against waves of enemies")},
+	{"terraforming", N_("addon_tag^Terraforming"),
+		// TRANSLATORS: tooltip in the drop-down menu for filtering add-ons
+		N_("addon_tag^Players can change the terrain")},
+};
+};
+
 addon_manager::addon_manager(addons_client& client)
 	: orders_()
 	, cfg_()
@@ -309,6 +343,22 @@ void addon_manager::pre_show(window& window)
 	connect_signal_notify_modified(status_filter,
 		std::bind(&addon_manager::apply_filters, this));
 
+	// The tag filter
+	auto& tag_filter = find_widget<multimenu_button>(&window, "tag_filter", false);
+
+	std::vector<config> tag_filter_entries;
+	for(const auto& f : tag_filter_types_) {
+		tag_filter_entries.emplace_back("label", t_string(f.label, GETTEXT_DOMAIN), "checkbox", false);
+		if(!f.tooltip.empty()) {
+			tag_filter_entries.back()["tooltip"] = t_string(f.tooltip, GETTEXT_DOMAIN);
+		}
+	}
+
+	tag_filter.set_values(tag_filter_entries);
+
+	connect_signal_notify_modified(tag_filter, std::bind(&addon_manager::apply_filters, this));
+
+	// The type filter
 	multimenu_button& type_filter = find_widget<multimenu_button>(&window, "type_filter", false);
 
 	std::vector<config> type_filter_entries;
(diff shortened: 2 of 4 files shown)


knyghtmare:
I would like this, but would UMC authors require any set list of tags, or are they allowed to go "wild" with the tags?

stevecotton:
Thanks for reminding me that I should have included this in the original description. Although UMC authors could add any tags they want, with this implementation only a few would be recognised by this filter box; it's a comma-separated list and unrecognised tags would simply be ignored. Other ways to see the tags would still work, but AFAICS there's nothing in the client's existing UI that shows the tags.

For example, Undead Empire has tags {`sevu`,`building`,`constructing`,`terraforming`,`anl`,`sim`,`city`,`survival`}. With this prototype, that means it appears in the both "Terraforming" and "Survival" lists.

Advance Wesnoth Wars has tags {`Advance_Wesnoth_Wars`, `advance wars`,`wargroove`, `xp`, `leader`, `heal`,  `healer`, `healing`, `promote`, `leader`, `recruitment`, `recruit`, `rng`, `prng`, `deterministic`, `random`, `randomless`, `no random`, `randomness`, `auto`, `passive`, `unknown`, `rpg`, `advanced`, `advancement`, `status`, `campaign`, `multiplayer`, `scenario`, `mod`, `mods`, `pack`, `predictive`, `deterministic`, `combat`, `attack`, `debug`, `terrain`, `pvp`, `pve`, `pvpve`, `combat`, `combats`, `modern`, `squad`, `troop`, `swarm`, `amla`, `level`, `hybrid`, `ninja`, `stealth`, `ambush`, `epic`, `boss`, `squad`, `gifted`, `survival`, `french`, `eagle`, `unitmarker`}. With this prototype that means it appears in the "RNG" and "Survival" lists.

Having a fixed list means this doesn't need logic to decide which tags to show, and it will allow the engine to include translations of the tags names along with tooltips:
wesnoth_tag_filter

stevecotton:
Does this do roughly what you want, @knyghtmare?

knyghtmare:
> Does this do roughly what you want, @knyghtmare?

I guess so. More or less. You can test this in the trunk add-on server too.

shikadiqueen:
I'm not @Vultraz but I'm pretty sure we'd want to allow people to select multiple tags at once, as well as some way to display the tags under the add-on description.

stevecotton:
I'm assuming that new tags could be added to the list of recognised tags during a stable branch, outside of string freeze periods.

> I'm pretty sure we'd want to allow people to select multiple tags at once

Given that an add-on can have multiple tags, if it should fit in more than one category then I'd expect the author to add both tags, so I don't see a reason for the UI to have an "xxx OR yyy" option. In the opposite direction I'm not expecting there to be so many add-ons that a user would want to filter for add-ons that only have "xxx AND yyy".

The "Type" drop-down is the opposite, for example a player might want "SP/MP Campaigns" plus "MP Campaigns".

shikadiqueen:
> Given that an add-on can have multiple tags, if it should fit in more than one category then I'd expect the author to add both tags, so I don't see a reason for the UI to have an "xxx OR yyy" option.

The filter options work on **multiple** add-ons at once -- the full server list. I don't see a reason for the UI **not** to allow people to list add-ons containing at least one of a set of tags chosen by the player.

shikadiqueen:
> The main issue here is how to display a list of strings that: (1) aren't translated, but that might match strings that we have translations for

Gotta document in PblWML what tags the game considers to have a built-in meaning/translation then. Which was sort of the plan all along.

> (2) could have whitespace in the data
> (3) could have and in the data (or whatever other marker format_conjunct_list uses)

Automatically-formatted lists of tags anywhere else wouldn't use "and". You should stick to commas here as long as we don't have a better way to distinguish between single and multi-word items like giving each tag a coloured background. Now, if you mean what to do with tags that themselves have "and" in them, just don't worry about them. Ideally people would use ampersands but that's kind of too much to ask from them.

(This whole system wasn't very well thought-out tbh and I just implemented it to appease someone who kept asking for it without much of a design in mind, hoping that improvements and new iterations would come into being organically like everything else since Wesnoth 0.1.)

data/gui/window/addon_manager.cfg
```diff
@@ -678,6 +710,32 @@
...
 
+			[column]
+				grow_factor = 0
+				border = "all"
+				border_size = 5
+				horizontal_alignment = "left"
+
+				[label]
+					definition = "default"
+					label = _ "Tag:"
```
shikadiqueen, line 721: Should be "Tags:" here as well.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -206,6 +206,27 @@ const std::vector<addon_manager::addon_order> addon_manager::all_orders_{
...
+	std::string tooltip;
+};
+
+const std::vector<addon_tag> tag_filter_types_ {
+	{"cooperative", N_("addon_tag^Cooperative"), N_("The players are all on the same team, versus the AI")},
+	{"cosmetic", N_("addon_tag^Cosmetic"), N_("These make the game look different, without changing gameplay")},
+	{"difficulty", N_("addon_tag^Difficulty"), N_("Can make campaigns easier or harder")},
+	{"rng", N_("addon_tag^RNG"), N_("Affect how luck and randomness occur")},
+	{"survival", N_("addon_tag^Survival"), N_("Fight against waves of enemies")},
+	{"terraforming", N_("addon_tag^Terraforming"), N_("Your units can change the terrain")},
```
shikadiqueen, line 226: Maybe "Players" rather than "Your units", since that's a broader and still correct description of how you could implement the same mechanic.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -206,6 +206,27 @@ const std::vector<addon_manager::addon_order> addon_manager::all_orders_{
...
+	std::string label;
+	/** Shown when hovering over an entry in the filter's drop-down list */
+	std::string tooltip;
+};
+
+const std::vector<addon_tag> tag_filter_types_ {
+	{"cooperative", N_("addon_tag^Cooperative"), N_("The players are all on the same team, versus the AI")},
+	{"cosmetic", N_("addon_tag^Cosmetic"), N_("These make the game look different, without changing gameplay")},
+	{"difficulty", N_("addon_tag^Difficulty"), N_("Can make campaigns easier or harder")},
+	{"rng", N_("addon_tag^RNG"), N_("Affect how luck and randomness occur")},
```
shikadiqueen, line 224: I don't feel too keen on talking about "luck and randomness" (what's the difference between either?). Perhaps a better description would be "Affect how chance-to-hit impacts combat outcomes"? Not too sure about this either to be honest.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -206,6 +206,27 @@ const std::vector<addon_manager::addon_order> addon_manager::all_orders_{
...
+	/** Text to match against addon_info.tags() */
+	std::string id;
+	/** What to show in the filter's drop-down list */
+	std::string label;
+	/** Shown when hovering over an entry in the filter's drop-down list */
+	std::string tooltip;
+};
+
+const std::vector<addon_tag> tag_filter_types_ {
+	{"cooperative", N_("addon_tag^Cooperative"), N_("The players are all on the same team, versus the AI")},
```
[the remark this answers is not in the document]
stevecotton: I'll take that rewording.

When I checked a couple of days ago, the 1.14 server had 2 with `coop`, 1 with `cooperative` and ` with `cooperation`. As we are going to be expressly telling UMC authors which words will be treated specially, my preference is for the unambiguous longer word.

src/gui/dialogs/addon/manager.cpp
stevecotton: I agree it needs to change, but not sure what to.

stevecotton:
I'd like to get both this and #5774 (showing the tags in the UI) in today, please.

The two PRs will have a trivial merge conflict in changelog.md. Whichever one goes in first, I want to do a manual rebase anyway to check that addon_manager.cfg merges correctly (that Git's fuzzy patch system has put the "Tags:" line at the bottom of that block).

src/gui/dialogs/addon/manager.cpp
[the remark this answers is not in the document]
stevecotton: I'm assuming that `cooperative` is still okay, and that this wasn't a required-to-change comment.

shikadiqueen (approved):

src/gui/dialogs/addon/manager.cpp
stevecotton: Going for "Modify the randomness in the combat mechanics, or remove it entirely"

PR #5770, opened 2021-05-09
Credits: hide the blank section for Friulian

stevecotton:
When an [about] section has a title but no [entry] tags, it's automatically
hidden in the credits. The entry with name="" was causing the title to show
up with a blank section under it.

Asturian and Friulian were removed in ad7699bffc16f7c982c95f1989ad6f932a1580b4
because they'd been added to the translation framework but hadn't received any
translated strings.

The reason that I'm leaving these in rather than completely removing them is
that the string "Friulian Translation" has already been translated into almost
all of the languages that Wesnoth supports. "Asturian Translation" has been
translated into around half of them, although the fuzzy-string support has put
the translation of "Estonian Translation" into the other half.

### DIFF
--- a/data/core/about_i18n.cfg
+++ b/data/core/about_i18n.cfg
@@ -706,9 +706,6 @@ sort=yes
 # no strings committed so far
 [about]
     title = _"Friulian Translation" # wmllint: no spellcheck
-    [entry]
-        name = ""
-    [/entry]
 [/about]
 
 [about]


shikadiqueen:
Instead of copy-pasting comments around, how about updating the [documentation](https://wiki.wesnoth.org/CampaignWML#Campaign_credits) instead?

stevecotton:
I'll take on the task of updating the docs, but I think copy-pasting a single line three times makes sense here. I assume a translator is going to look near their specific language in the WML file itself, even though they'll look at all the translatable strings in poedit.

It also (deliberately) means there will be a conflict to stop Friulian's empty entry getting copied in again when someone next syncs from 1.14 to master.

shikadiqueen:
> It also (deliberately) means there will be a conflict to stop Friulian's empty entry getting copied in again when someone next syncs from 1.14 to master.
Why not remove it from 1.14?

stevecotton:
> Why not remove it from 1.14?

Will do, and this is now smaller so I feel OK about merging without waiting for more feedback.

PR #5773, opened 2021-05-10
Run woptipng on the new horses

stevecotton:
57 of 57 files optimized, 678061 bytes reduced to 586300 bytes; -91761 bytes, -13.53%
This was running with the -t0 option, so that it considers any reduction good, even
if it's less than 10% on the big portrait images.

@doofus-01 sorry for not catching this before #5738 merged.


soliton-:
Perhaps we could make some CI check for this.

It'd be great to make sure images are optimized before first commit since each time re-committing binary files blows up the repository step by step.

stevecotton:
> Perhaps we could make some CI check for this.

I'm not sure we can, PNG optimisation seems slower than C++ compilation. It might be possible to have a verification option for woptipng - which doesn't guarantee that it finds the best optimisation, but just detects if the quick optimisations manage to save something (and then warns that the full tool needs to be run).

PR #5774, opened 2021-05-11
Add-ons manager: show tags in the details panel

stevecotton:
Show the contents of PblWML's "tags" attribute in the add-ons dialog. They're
shown at the bottom of the right-hand panel (only if the window size is large
enough to trigger that dialog's two-panel layout).

This started as part of #5765, I've split it because I think this part is ready to
merge, but the filtering-by-tags part needs an inspiration about the descriptive
strings.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 ## Version 1.15.12+dev
  ### Add-ons client
+   * The details panel now shows the list of tags in each add-on.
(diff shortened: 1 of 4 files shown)


Pentarctagon:
A screenshot would be useful.

shikadiqueen (approved):
Seems good to me.

PR #5775, opened 2021-05-11
Make callers of help::show_help not need a help_manager

stevecotton:
Most of the places that created a help_manager did so just to call show_help on
the next line. Move knowledge of the internals out of those callers and into
help.cpp.

This leaves editor_controller and play_controller with knowledge of help_manager,
however both of those classes handle ownership of the game_config too.

@CelticMinstrel, I still have some `\todo` comments to check, but I think this is a better way to handle help_manager.


stevecotton:
The `\todo` comments are all done, I think this is ready to go in.

CelticMinstrel:
Though I find this kind of pointless, I also don't have any particular issues with it. It doesn't seem any better than the previous approach, but it doesn't seem any worse either.

PR #5781, opened 2021-05-13
Another round of Lua reorganization work

CelticMinstrel:
This currently contains the floating label change (already seen in the previous pull request) as well as a proposed new dir() function. Most likely, both of these will be subject to some renames before they're ready for merge.

I also plan to add modules for pathfinding and ~schedule management~ in this PR. They should be straight renames with no new functionality.


CelticMinstrel:
After examining schedule management in more detail, I decided it would need to be more than just a straight rename, so I'll open a new pull request for it. It's possible it might not make it into 1.16, though I'll try to get there if I can.

CelticMinstrel:
I think the unit tests are failing because deprecation messages cause strict failure. I'll try running them locally without strict.

src/scripting/game_lua_kernel.cpp
```diff
@@ -4253,6 +4252,7 @@ game_lua_kernel::game_lua_kernel(game_state & gs, play_controller & pc, reports
 		{ "append_ai", &intf_append_ai },
 		{ "debug_ai", &intf_debug_ai },
 		{ "switch_ai", &intf_switch_ai },
+		{ "teleport", &dispatch<&game_lua_kernel::intf_teleport> },
```
stevecotton, line 4255: This is in `side_callbacks`, `unit_callbacks` is about 20 lines above.

stevecotton:
If I'm understanding the error correctly, then the error message and names of the arguments to `wesnoth.deprecate_api` are wrong.

```
-- Marks a function or subtable as deprecated.
-- Parameters:
---- elem_name: the full name of the element being deprecated (including the module)
---- replacement: the name of the element that will replace it (including the module)
---- level: deprecation level (1-4)
---- version: the version at which the element may be removed (level 2 or 3 only)
---- Set to nil if deprecation level is 1 or 4
---- elem: The actual element being deprecated
---- detail_msg: An optional message to add to the deprecation message
function wesnoth.deprecate_api(elem_name, replacement, level, version, elem, detail_msg)

	...

	else
		wesnoth.log('warn', "Attempted to deprecate something that is not a table or function: " ..
			elem_name .. " -> " .. replacement .. ", where " .. elem_name .. " = " .. tostring(elem))
```

Should those arguments really be called `(elem_name, replacement_name, level, version, replacement, detail_msg)`, and the error message in this case end `where ".. replacement_name .." = " .. tostring(replacement)`?

stevecotton:
While I made the most noise about splitting up the previous PR, I wasn't the only one who said to split this PR. The sooner it gets split into smaller ready-to-merge bits, the sooner those bits get merged, and so the sooner any bugs such as #5816 get found.

```
20210514 00:08:11<+wesdiscordbot> <Pentarctagon> it'll be easier to have it be multiple PRs if you start with it that way.
20210514 00:08:28< celticminstrel> It won't be any easier at this point.
20210514 00:08:51< celticminstrel> Cherry-picking either of the two current commits to another branch is trivial.
20210514 00:09:05< celticminstrel> And the same will almost certainly be true of the subsequent two commits.
20210514 00:09:39< celticminstrel> Once I start replacing deprecated uses in master, that might change, but for now, I don't see any hurry to split things up.
20210514 00:11:07< celticminstrel> It's quite a bit easier to not split it up. Switching branches can be a pain, especially since the Xcode project doesn't work without some customizations.
20210514 00:11:38< celticminstrel> So even if it's going to be split up, I'd rather put it off at least until it's actually ready.
20210514 00:12:03<+wesdiscordbot> <Iris> have you considered at some point that you try to do way too many things in a single pr at the same time?
20210514 00:12:25< celticminstrel> lua_reorg is one thing in a certain sense, so…
20210514 00:13:57<+wesdiscordbot> <Pentarctagon> it was made of like 25 smaller things
```

stevecotton:
> wesnoth.unit_types and wesnoth.terrain_types are now available in map generation scripts

What uses the newly-available functions during map generation, or in other words what tests this new functionality?

data/lua/wml-tags.lua
```diff
@@ -767,8 +767,48 @@ function wml_actions.redraw(cfg)
 	wesnoth.redraw(cfg, clear_shroud)
 end
 
+local wml_floating_label = {valid = false}
 function wml_actions.print(cfg)
-	wesnoth.print(cfg)
+	wesnoth.deprecation_message('[print]', 1, nil, 'use [floating_label] instead')
```
stevecotton, line 772:
```
20210530 06:18:39 error scripting/lua: lua/wml-tags.lua:772: attempt to call a nil value (field 'deprecation_message')
stack traceback:
	lua/wml-tags.lua:772: in local 'cmd'
```

CelticMinstrel: Moved to #5837

CelticMinstrel:
> What uses the newly-available functions during map generation, or in other words what tests this new functionality?

Moved to #5838

src/scripting/game_lua_kernel.cpp
```diff
@@ -4260,6 +4248,13 @@ game_lua_kernel::game_lua_kernel(game_state & gs, play_controller & pc, reports
 		{"zoom", &dispatch<&game_lua_kernel::intf_zoom>},
 		{"clear_menu_item", &dispatch<&game_lua_kernel::intf_clear_menu_item>},
 		{"set_menu_item", &dispatch<&game_lua_kernel::intf_set_menu_item>},
+		{"allow_end_turn", &dispatch<&game_lua_kernel::intf_allow_end_turn>},
+		{"allow_undo", &dispatch<&game_lua_kernel::intf_allow_undo>},
+		{"cancel_action", &dispatch<&game_lua_kernel::intf_cancel_action>},
```
stevecotton, line 4253: I'm hesitant about classing `allow_undo` and `cancel_action` as "functions for querying or manipulating the in-game UI".

* `cancel_action` seems more like a unit-placement function, putting the unit in the hex that it's currently moving through.
* `allow_undo`'s effect on the event stack will affect the UI, but it feels like a property of the `[event]` saying that the `[event]` doesn't affect game state.

CelticMinstrel: Hmm… I definitely see your point. Perhaps allow_undo's deprecation should instead be integrated into the events PR using a new API.

I don't see `cancel_action` as a unit function. It also feels more event-related, actually, so maybe it should also be moved into game_events as part of the events PR?

stevecotton: Classing them as event-related sounds good to me.

Redesigning this bit of the API sounds good too.

data/lua/core/_initial.lua
```diff
@@ -107,3 +107,14 @@ wesnoth.set_end_campaign_text = wesnoth.deprecate_api('wesnoth.set_end_campaign_
 		wesnoth.scenario.end_text_duration = d
 	end
 end)
+
+if wesnoth.kernel_type() ~= 'Application Lua Kernel' then
```
stevecotton, line 111: The C++ has `intf_find_path` for the mapgen and game kernels, but not the application kernel.

src/scripting/mapgen_lua_kernel.cpp
```diff
@@ -249,6 +249,18 @@ mapgen_lua_kernel::mapgen_lua_kernel(const config* vars)
 	lua_pop(L, 1);
 	assert(lua_gettop(L) == 0);
 
+	// Create the paths module
+	cmd_log_ << "Adding paths module...\n";
+	static luaL_Reg const path_callbacks[] {
+		{ "find_path",                 &intf_find_path                          },
```
stevecotton, line 255: Although they have the same name, `game_lua_kernel::intf_find_path` and `mapgen_lua_kernel::intf_find_path` expect different argument lists. This seems a good time to give the mapgen one a different name.

src/scripting/mapgen_lua_kernel.cpp
```diff
@@ -249,6 +249,18 @@ mapgen_lua_kernel::mapgen_lua_kernel(const config* vars)
 	lua_pop(L, 1);
 	assert(lua_gettop(L) == 0);
 
+	// Create the paths module
+	cmd_log_ << "Adding paths module...\n";
```
stevecotton, line 253: This lacks the corresponding removal of `wesnoth.find_path`, which is still there on line 223.

Line 224 is also interesting, that's now bug #5846.

data/lua/core/_initial.lua
CelticMinstrel: That's… why this line exists?

src/scripting/mapgen_lua_kernel.cpp
CelticMinstrel: I thought one of those argument lists was sort of a subset of the other…? I'll take a closer look though.

data/lua/core/_initial.lua
stevecotton: Oops, I misread `~=` as `==`

stevecotton:
This doesn't yet include updating the places in mainline that call these functions.

src/scripting/mapgen_lua_kernel.cpp
CelticMinstrel: The difference pretty much amounts to an extra parameter in `mapgen_lua_kernel::intf_find_path` as well as it not supporting the options table (you must directly pass a function).

I think the two APIs could just be unified - add the extra options in `mapgen_lua_kernal` to `game_lua_kernel` and at least some of the extra options in `game_lua_kernel` to `mapgen_lua_kernel`. That may be out of scope for this PR, but I think I'll at least make it support an options table.

data/ai/lua/ai_helper.lua
```diff
@@ -13,7 +13,7 @@ local M = wesnoth.map
 -- With their default settings, the ai_helper functions use the vision a player of
 -- the respective side would see, that is, they assume no knowledge of invisible
 -- units. This can be influenced with the 'viewing_side' and 'ignore_visibility' parameters,
--- which work in the same way as they do in wesnoth.find_reach() and wesnoth.find_path():
+-- which work in the same way as they do in wesnoth.find_reach() and wesnoth.paths.find_path():
```
stevecotton, line 16: `wesnoth.find_reach()` also moved to `wesnoth.paths`

data/ai/lua/battle_calcs.lua
```diff
@@ -1510,7 +1510,7 @@ function battle_calcs.get_attack_combos_subset(units, enemy, cfg)
             if (not blocked_hexes:get(xa, ya) or ((xa == unit.x) and (ya == unit.y))) then
 
                 -- Check whether the unit can get to the hex
-                -- wesnoth.map.distance_between() is much faster than wesnoth.find_path()
+                -- wesnoth.map.distance_between() is much faster than wesnoth.paths.find_pathd_path()
```
stevecotton, line 1513: Typo

src/scripting/game_lua_kernel.cpp
```diff
@@ -2694,7 +2700,9 @@ namespace
 			lua_pushnumber(L, side);
 			if (luaW_pcall(L, 1, 1, false)) {
 				if(!luaW_toconfig(L, -1, cfg)) {
-					lua_kernel_base::get_lua_kernel<game_lua_kernel>(L).log_error("function returned to wesnoth.synchronize_choice a table which was partially invalid");
+					static const char* msg = "function returned to wesnoth.synchronize_choice a table which was partially invalid";
```
stevecotton, line 2703: Uses deprecated name of function

src/scripting/mapgen_lua_kernel.cpp
```diff
@@ -157,34 +157,55 @@ static int intf_default_generate_height_map(lua_State *L)
 }
 /**
  * Finds a path between two locations.
- * - Args 1,2: source location.
- * - Args 3,4: destination.
- * - Arg 5: cost function
- * - Args 6,7 size of map.
- * - Arg 8 include border.
+ * - Args 1: source location.
+ * - Args 2: destination.
```
stevecotton, line 161: This is meant to still work as the deprecated `wesnoth.find_path`, there's no wrapper function around it in `data/lua/core/_initial.lua`.

CelticMinstrel: I'm not sure what you're getting at? There's an assignment for `wesnoth.find_path = wesnoth.paths.find_path` on line 112. The arguments to the new function are the same but with two extra features - the first argument can be a location table (rather than the first two arguments being x and y) and the last argument can be an options table instead of passing the cost function and other options all as separate parameters.

So, any code written against the current API still works.

stevecotton: > the first argument can be a location table

Then the documentation doesn't match the function.

CelticMinstrel: It does match. The first argument is a location. The computer might see it as one argument (a table with two values) or as two arguments (separate integers), but it's logically a single argument.

And this equivalence holds true basically everywhere you can pass a location in the Lua API; all I did was add another case of it.

stevecotton:
When I read in IRC that "all except 5852 has already had some review" then I feel you're saying that I've reviewed it. I've actually stopped before completing the review, because I feel it's no longer productive to continue.

Please split this into separate PRs, so that the parts that are already ready to merge can merge. Otherwise reviewers are having to repeatedly look through old stuff to reach the new stuff; or maybe they're waiting until the two of us have finished going back and forth before they take a look themselves.

Several mainline campaigns trigger deprecation warnings. If this were split into smaller PRs then it would be much easier to see which changes to the Lua API lack corresponding changes in mainline; take the lack of detail in this paragraph's opening sentence as an expression of my frustration at the size of this PR.

CelticMinstrel:
You can always review individual commits one at a time, can't you? There's no need to review all pull requests in one block. Sometimes it makes sense to, sometimes it doesn't.

I think I've tracked down the rest of the deprecated uses that need to be replaced now.

src/scripting/game_lua_kernel.cpp
```diff
@@ -4272,6 +4261,37 @@ game_lua_kernel::game_lua_kernel(game_state & gs, play_controller & pc, reports
...
+	lua_setfield(L, -2, "paths");
+	lua_pop(L, 1);
+	
+	// Create the sync module
+	cmd_log_ << "Adding sync module...\n";
+	static luaL_Reg const sync_callbacks[] {
+		{ "invoke_command",    &intf_invoke_synced_command },
+		{ "run_unsynced",      &intf_do_unsynced },
+		{ "evaluate",          &intf_synchronize_choice },
+		{ "multi_evaluate",    &intf_synchronize_choices },
```
Pentarctagon, line 4287: I'd prefer `multi_evaluate` to be `evaluate_multiple`.

Pentarctagon:
As far as the PR as a whole, I'd like:
* The first two commits (path-related stuff) as a PR.
* The second two commits (wesnoth.interface stuff) as a PR.
* The last commit (sync module) as a PR.
* Moving wesnoth.teleport (5th commit) and wesnoth.get_traits (6th commit) seem simple enough to just go ahead and cherry-pick to master.

src/scripting/game_lua_kernel.cpp
CelticMinstrel: That's funny, because that was my first thought, but then I changed it to this.

CelticMinstrel:
See #5861, #5862, and #5863.

data/lua/core/interface.lua
```diff
@@ -53,6 +53,11 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 	wesnoth.zoom = wesnoth.deprecate_api('wesnoth.zoom', 'wesnoth.interface.zoom', 1, nil, wesnoth.interface.zoom)
 	wesnoth.gamestate_inspector = wesnoth.deprecate_api('wesnoth.gamestate_inspector', 'gui.show_inspector', 1, nil, gui.show_inspector)
 	wesnoth.color_adjust = wesnoth.deprecate_api('wesnoth.color_adjust', 'wesnoth.interface.color_adjust', 1, nil, function(cfg) wesnoth.interface.color_adjust(cfg.red, cfg.green, cfg.blue) end)
+	wesnoth.allow_end_turn = wesnoth.deprecate_api('wesnoth.allow_end_turn', 'wesnoth.interface.allow_end_turn', 1, nil, wesnoth.interface.allow_end_turn)
+	wesnoth.clear_messages = wesnoth.deprecate_api('wesnoth.clear_messages', 'wesnoth.interface.clear_messages', 1, nil, wesnoth.interface.clear_messages)
+	wesnoth.end_turn = wesnoth.deprecate_api('wesnoth.end_turn', 'wesnoth.interface.end_turn', 1, nil, wesnoth.interface.end_turn)
+	wesnoth.get_viewing_side = wesnoth.deprecate_api('wesnoth.get_viewing_side', 'wesnoth.interface.get_viewing_side', 1, nil, wesnoth.interface.get_viewing_side)
+	wesnoth.message = wesnoth.deprecate_api('wesnoth.message', 'wesnoth.interface.add_message', 1, nil, wesnoth.interface.add_message)
```
stevecotton, line 60: Maybe make this `add_chat_message` to clarify further that it's unrelated to `[message]`?

CelticMinstrel: I'm not opposed to that, but it seems that if this one is called `add_chat_message` then the one that clears it should be `clear_chat_messages`, which means that `clear_chat_messages` also clears WML error messages. Is that okay? (Of course, it only clears them from chat; they'd still be in the Lua console.)

stevecotton: I agree with the logic, and think the name `clear_chat_messages` while a little inaccurate will be clear to devs about which part of the UI it clears.

I do wonder if there should be a Lua function for clearing the chat area at all - it sounds like something that should only be triggered by the player choosing to do it.

src/scripting/game_lua_kernel.cpp
stevecotton: If they're not interface functions, why is this PR still moving them into `wesnoth.interface`?

CelticMinstrel: Huh? I thought I already reverted that bit… did I miss something?

data/lua/core/interface.lua
CelticMinstrel: You might have a point? It's not used in mainline anywhere, either…

stevecotton (approved):

PR #5783, opened 2021-05-14
lua/gui2: Restore layout invalidation hack when setting widget labels

### LINKED ISSUE #5755: lua/gui2: Rebuilding listboxes on 1.15.11+ breaks them
(I've been aware of this issue for a while but I'm only now filing a bug report, hence 1.15.11+.)

The following 1.14-compatible Lua code displays a simple dialog box with a list containing three items, along with a button to rebuild the list:

```lua
local T = wml.tag

local dd = {
    T.helptip { id = "tooltip_large" },
    T.tooltip { id = "tooltip_large" },

    T.grid {
        T.row {
            T.column {
                T.grid {
                    T.row {
                        T.column {
                            border = "all",
                            border_size = 5,
                            T.button {
                                id = "ok",
                                label = "Close"
                            }
                        },
                        T.column {
                            border = "all",
                            border_size = 5,
                            T.button {
                                id = "refresh",
                                label = "Refresh"
                            }
                        },
                    }
                }
            }
        },
        T.row {
            T.column {
                border = "all",
                border_size = 5,
                T.listbox {
                    id = "list",
                    T.list_definition {
                        T.row {
                            T.column {
                                vertical_grow = true,
                                horizontal_grow = true,
                                T.toggle_panel {
                                    T.grid {
                                        T.row {
                                            T.column {
                                                horizontal_grow = true,
                                                grow_factor = 1,
                                                border = "all",
                                                border_size = 5,
                                                T.label {
                                                    id = "text"
                                                }
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}

wesnoth.show_dialog(dd, function()
    local function refresh()
        wesnoth.remove_dialog_item(1, 0, "list")

        wesnoth.set_dialog_value("Foo", "list", 1, "text")
        wesnoth.set_dialog_value("Bar", "list", 2, "text")
        wesnoth.set_dialog_value("Baz", "list", 3, "text")
    end

    wesnoth.set_dialog_callback(refresh, "refresh")

    refresh()
end)
```

On 1.14.x, this code works without any issue. However, on 1.15.x, the listbox becomes non-functional and invisible as soon as Refresh is pressed for the first time, resulting in this error in stderr:

```
20210501 03:24:37 error gui/layout: grid [_content_grid] place: Failed to place a grid, we have 39,30 space but we need 39,93 space. This happened at a grid with the id '_content_grid' in a 'N4gui27listboxE' with the id 'list' in a 'N4gui24gridE' with the id '_window_content_grid' in a 'N4gui24gridE' with the id '_content_grid' in a 'N4gui215scrollbar_panelE' with the id '' in a 'N4gui24gridE' with the id '' in a 'N4gui26windowE' with the id ''.
```

While investigating the original issue in my add-on I found out that this happens there specifically if a list is cleared and rebuilt with less items than the amount it would need to cause the listbox's vertical size to grow. If more items are inserted than there's available vertical space for, the listbox grows and it remains functional. However, in the minimal test case the number of items inserted seems to have no bearing on whether the listbox remains functional or not — it just breaks. It could be because in the test case here the dialog's vertical size is not constrained.

shikadiqueen:
Fixes #5755.

This is one of several layout invalidation hacks used in the old Lua
GUI2 API lost in the implementation of the new API (which the old API
now wraps).

It turns out that in the previous API (see 1.14 for reference if you
need easy access to the source) there were several calls to
invalidate_layout() that weren't there just to look pretty.

For wesnoth.set_dialog_value:

 * One prior to setting the value of a text_box
 * One prior to setting the value of a unit_preview_pane
 * One prior to setting the value of a styled_widget
 * One after changing a widget's visibility to hidden

All of these were removed by the new API implementation, with the last
one ending up commented out, restored now in commit
8989eb430311092e66177bc09e1b1c344000d5aa. All the other ones are
entirely gone, however.

It turns out that the dialog layout invalidation done for the
styled_widget case was essential to get listboxes working correctly
after clearing and adding rows to them (which was automatically done
right before in the implementation of wesnoth.set_dialog_value by the
find_widget() helper function). It was done for every single listbox
item, but without it the listbox would become completely useless under
certain conditions:

  20210501 03:24:37 error gui/layout: grid [_content_grid] place: Failed to place a grid, we have 39,30 space but we need 39,93 space. This happened at a grid with the id '_content_grid' in a 'N4gui27listboxE' with the id 'list' in a 'N4gui24gridE' with the id '_window_content_grid' in a 'N4gui24gridE' with the id '_content_grid' in a 'N4gui215scrollbar_panelE' with the id '' in a 'N4gui24gridE' with the id '' in a 'N4gui26windowE' with the id ''.

Now, it seems horribly inefficient to do this for every single new row
added to the listbox, but this solution *works*. We might want to
restore the hacks for the other 2 cases as well -- or figure out a
cleaner solution for the aforementioned bug.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,7 @@
    * Add a more advanced way of manipulating version strings via `wesnoth.version` constructor function.
    * Add a `wesnoth.scenario` data module that consolidates everything there is to know about the current scenario. Much of this information was previously available through `wesnoth.game_config`, but there are some new details too.
    * Fixed a regression introduced in 1.15.5 causing GUI2 widgets set to a hidden state to remain visible but not able to be interacted with.
+   * Fixed a regression introduced in 1.15.5 causing GUI2 listboxes to break when cleared and re-filled under certain circumstances.
  ### Packaging
    * Increased minimum required version of SDL to 2.0.8 (PR #5736).
  ### Terrain
--- a/src/scripting/lua_widget_attributes.cpp
+++ b/src/scripting/lua_widget_attributes.cpp
@@ -397,6 +397,10 @@ WIDGET_SETTER("visible", lua_index_raw, gui2::styled_widget)
 //must be last
 WIDGET_SETTER("value_compat,label", t_string, gui2::styled_widget)
 {
+	gui2::window* window = w.get_window();
+	if(window) {
+		window->invalidate_layout();
+	}
 	w.set_label(value);
 }
 


stevecotton:
> Now, it seems horribly inefficient to do this for every single new row added to the listbox, but this solution /works/.

+1 from me

CelticMinstrel:
Yeah… if it works, then it works. Seems fine to me. Sure, it might be better to find a better solution, but getting it working is probably more important.

PR #5786, opened 2021-05-15
adding seahorse graphics

doofus-01:
I can't run woptipng, and optipng doesn't do much with this.  I might merge this later, but the image should cool off here for a while in case it needs fixing.


doofus-01:
Thanks to @stevecotton wiki update, I was able to get woptipng.py to run.  I'll commit this once CI passes.

PR #5787, opened 2021-05-16
Icemonax advances to Greater Icemonax

ForestDragon-wesnoth:
I think it is perfectly reasonable to be able to have icemonax advance to greater icemonax. This makes the units feel like a proper unit line, and gives UMC authors more freedom how to use the unit. To compensate for the fact that a lvl0 unit advances to lvl2, I set the exp quite high (53) to make this not something abusable

### DIFF
--- a/data/core/units/monsters/Icemonax.cfg
+++ b/data/core/units/monsters/Icemonax.cfg
@@ -23,11 +23,10 @@
     hitpoints=35
     movement_type=largefoot
     movement=4
-    experience=25
+    experience=53
     level=0
     alignment=neutral
-    advances_to=null
-    {AMLA_DEFAULT}
+    advances_to=Great Icemonax
     undead_variation=wolf
     cost=11
     usage=fighter


doofus-01:
It's already pretty easy for UMC authors to allow the advancement, not sure I see the "more freedom".  The level 0 -> 2 seems more of an issue.

The wolf doesn't advance to great wolf, and people have been fine with that for a long time.  Why this push here?

ForestDragon-wesnoth:
> It's already pretty easy for UMC authors to allow the advancement, not sure I see the "more freedom". 
Yeah, you make a good point there
> The level 0 -> 2 seems more of an issue.
well, that isn't entirely unthinkable in my opinion, and will make the unit feel more unique from other monsters
> The wolf doesn't advance to great wolf
honestly that is very bad too in my opinion
> and people have been fine with that for a long time. Why this push here?
well, me and some others in the discord server have been discussing that some of the locked advancements (like icemonax, arcanister, or the wolves) should not be locked as they are, but should be advanceable normally

ForestDragon-wesnoth:
> It's already pretty easy for UMC authors to allow the advancement, not sure I see the "more freedom". 

Yeah, you make a good point there

> The level 0 -> 2 seems more of an issue.

well, that isn't entirely unthinkable in my opinion, and will make the unit feel more unique from other monsters

> The wolf doesn't advance to great wolf

honestly that is very bad too in my opinion

> and people have been fine with that for a long time. Why this push here?

well, me and some others in the discord server have been discussing that some of the locked advancements (like icemonax, arcanister, or the wolves) should not be locked as they are, but should be advanceable normally

knyghtmare:
> well, me and some others in the discord server have been discussing that some of the locked advancements (like icemonax, arcanister, or the wolves) should not be locked as they are, but should be advanceable normally

The Icemonax's advancement seems odd to me. True, the high XP requirement does seem to counter a lvl 0 advancing to a lvl 2, but I feel as though an "adolescent Icemonax" seems missing and should be considered (for 1.17.x). However, the current state seems fine as well, it could just grown significantly larger during a hibernation cycle of sorts.

I do feel the wolves should have advancement enabled by default. Most people don't use them since most don't know how to enable the advancements as I have seen.

The Arcanister one...I am not sure about this one as it might totally imbalance Sceptre of Fire. Better to just have it as a "Thursagan" only unit.

doofus-01:
There is the macros file `optional_unit_advancements.cfg` that deals with this for the other units.  I'm not sure how useful UMC creators find that, but it's apparently been the solution to this and seems like a sane compromise.

ForestDragon-wesnoth:
I can understand some advancements being locked, like ancient lich/runesmith, but others like wolves I honestly can't see the point of having locked by default. mainline monster advancements are inconsistent as hell - mudcralwers can advance, wolves can't, bats can advance, icemonax can't.

> I'm not sure how useful UMC creators find that
well, it's better than having to use [base_unit] for advancements, but the problem is not with the feature (macros locking/unlocking advancements) itself imo, bur rather which unit paths are locked.

ForestDragon-wesnoth:
> I do feel the wolves should have advancement enabled by default. Most people don't use them since most don't know how to enable the advancements as I have seen.

agreed. on top of that it's also weird how wolves can grow up on advancement when there's a goblin riding them, but not when they are by themself

CelticMinstrel:
I'm dubious about the whole skipping a level thing, so I'd prefer this as an optional advancement.

> on top of that it's also weird how wolves can grow up on advancement when there's a goblin riding them, but not when they are by themself

That's probably the goblin getting a new wolf, rather than their wolf growing up.

Wolf advancement might be disabled by default because the three levels are technically different species.

ForestDragon-wesnoth:
> I'm dubious about the whole skipping a level thing, so I'd prefer this as an optional advancement.
fair enough. that is a reasonable compromise.
> That's probably the goblin getting a new wolf, rather than their wolf growing up.
makes sense,

cooljeanius:
I think having an intermediate Level 1 icemonax would be better than having a Level 0 advance straight to a Level 2

cooljeanius:
(this is the last thing I'm still keeping cooljeanius/Son_Of_The_Black_Eye_Easy_Edition#6 open for, btw)

Wedge009:
I'm not familiar with either the icemonax or the wolf (in terms of advancement) but I also agree there should be an intermediate L1 stage if one desires an advancement from L0 to L2. I don't think it's a matter of 'uniqueness' but of setting/breaking expectations in terms of the game mechanics and unit conventions. At this point I don't see any reason for there to be a direct advancement from L0 to L2.

cooljeanius:
so wait I still think the "add an intermediate icemonax" idea is worthwhile... should I open a separate issue for that?

cooljeanius:
> Yeah, that'd be an entirely new unit rather than just editing the advancement of an existing unit.

ok I opened #6805

PR #5789, opened 2021-05-16
Last new terrains for 1.16

doofus-01:
For now, it's snowy (^Feta) and dead oak (*^Feth) variations of great tree.  There might or might not be anything else, we'll see.


doofus-01:
I was looking into snowy bridges, when I realized (despite what I hinted at on the forums), there does need to be some frozen road terrain.  A variation on the clean cobbles was the most straightforward.  In the shot below, there are some patches of Rrc,at the north side of the chasm and near the more SW bridge, but the rest is Rra.

Screenshot_20210522_a

CelticMinstrel:
I'm amused by the terrain code for the snowy great tree…

doofus-01:
> I'm amused by the terrain code for the snowy great tree…

Well, as long as giant blocks of cheese don't become a terrain type, I think we're safe.  It does bring to mind the trope about travelers lost in the dessert.

PR #5794, opened 2021-05-18
update codeblock projectfiles andfix lua_mathx not compiled with gcc 8.0 on windows7(in codeblock project)

newfrenchy83:
When i compile this file with gcc called with codeblock editor i have line 59: 'std::round is not a member of std'

### DIFF
--- a/src/scripting/lua_mathx.cpp
+++ b/src/scripting/lua_mathx.cpp
@@ -22,8 +22,8 @@
 #include "lua/lua.h"
 #include "lua/lualib.h"
 
-#include <stdio.h>
-#include <math.h>
+#include <cstdio>
+#include <cmath>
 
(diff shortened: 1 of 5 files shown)


newfrenchy83:
@Pentarctagon  could you review this PR, this is a fix for gcc 8.0 compilation

src/scripting/lua_mathx.cpp
```diff
@@ -22,6 +22,9 @@
 #include "lua/lua.h"
 #include "lua/lualib.h"
 
+#include <stdio.h>
+#include <math.h>
```
Pentarctagon, line 26: These should be `<cstdio>` and `<cmath>`.

newfrenchy83: It's fixed

PR #5796, opened 2021-05-20
Hash passwords server-side instead of client-side.

Pentarctagon:
This removes the need for the Wesnoth client to need to know how to hash the password to match what is expected by the forum database.

---

This does of course depend on TLS support being enabled and working first.


src/server/wesnothd/server.cpp
```diff
@@ -974,46 +973,93 @@ template<class SocketPtr> bool server::authenticate(
...
+			return std::make_pair(
+				utils::md5(hash.base64_digest(), outer_salt).base64_digest(),
+				nonce
+			);
+		} catch(const utils::hash_error& err) {
+			ERR_SERVER << "bcrypt hash failed: " << err.what() << std::endl;
+			return std::make_pair("", "");
+		}
+	} else {
+		ERR_SERVER << "Unable to determine how to has the password for user: " << user << std::endl;
```
loonycyborg, line 1049: Does it mean "how to <i>pass</i>" the password?

Pentarctagon: Whoops, should be "hash".

loonycyborg:
Currently second layer of hashing we do is only useful for unencrypted passing of password. If server does double hashing then it becomes entirely redundant. I guess could do it like this for now and simplify later though.

Pentarctagon:
Yeah, for now I just focused on moving the client-side code to the server, not whether parts of it are still really needed or not.

src/server/common/server_base.cpp
```diff
@@ -571,6 +586,106 @@ void server_base::load_tls_config(const config& cfg)
...
+			if(outer_salt.size() != 32) {
+				throw utils::hash_error("salt wrong size");
+			}
+
+			return std::make_pair(
+				utils::md5(hash.base64_digest(), outer_salt).base64_digest(),
+				nonce
+			);
+		} catch(const utils::hash_error& err) {
+			ERR_SERVER << "bcrypt hash failed: " << err.what() << std::endl;
```
soliton-, line 680: Should at least contain the username.

src/server/wesnothd/server.cpp
```diff
@@ -880,17 +880,23 @@ template<class SocketPtr> bool server::authenticate(
...
-	// applies the nonce to the valid hash and compares the results
+	// - Server sends client a password request (if TLS/database support is enabled)
+	// - Client sends the plaintext password
+	// - Server receives plaintext password, hashes it, and compares it to the password in the forum database
 
 	registered = false;
 
 	if(user_handler_) {
+		const std::string salt = user_handler_->extract_salt(username);
+		if(salt.length() < 2) {
```
soliton-, line 891: Where does the 2 come from?

src/game_initialization/multiplayer.cpp
```diff
@@ -442,7 +401,9 @@ std::unique_ptr<wesnothd_connection> mp_manager::open_connection(std::string hos
 
 				const std::string ec = (*error)["error_code"];
 
-				if(ec == MP_MUST_LOGIN) {
+				if(is_pw_request && !conn->using_tls()) {
+					error_message = _("The remote server requested a password while using an insecure connection.");
```
soliton-, line 405: `is_pw_request` is also false if the password is empty even if there was a password request.

soliton- (changes requested):

src/server/wesnothd/server.cpp
Pentarctagon: Later it checks whether `salt[1] == '2'`, which is UB if the salt is length 0 or length 1.  This check is also done here just so I don't need to make the hash_password() method templated just to send the error message back.

src/server/common/server_base.cpp
Pentarctagon: Added.

src/game_initialization/multiplayer.cpp
Pentarctagon: Fixed.

soliton-: I would have actually fixed the variable instead. It has a misleading name that already tripped you up once.

Pentarctagon: I decided to just delete the variable and update the if-statement's comments.

src/server/wesnothd/server.cpp
soliton-: `hash_password` has a lot of error conditions anyway. Wouldn't it just be one more case where an empty pair is returned? Or is it clear that if the salt is a single character the "login to forum" hint is more appropriate?

Pentarctagon: I don't know which is more appropriate, I'm just preserving the existing functionality of sending that error message back to the user.  If that's not needed anymore, then I can delete that error message and instead let it fall back to the default "Password hashing failed" error message.

soliton-: But existing functionality is to send this error message only if `salt` is empty. Now it is also sent if `salt` contains 1 character.

Pentarctagon: So I should change it check for salt being empty?

soliton-: I think so. That would get rid of the magic number 2 that can only be explained by code in another function which does its own check anyway.

soliton- (approved):

loonycyborg (approved):

PR #5798, opened 2021-05-21
An easier way of setting special notes in the most common use-case

CelticMinstrel:
Abilities and weapon specials now have a `special_note` attribute that automatically gets prepended to the list of special notes. The `[special_note]` tag is still supported for special notes not directly linked to an ability.

This deprecates all but 3 of the `{NOTE_*}` macros and even updates wmllint. The horses were missing `{NOTE_DEFENSE_CAP}` so I also added that.


data/core/macros/special-notes.cfg
```diff
@@ -126,171 +132,115 @@ _"This unit has a defense cap on certain terrain types — it cannot achieve a h
...
-        note={INTERNAL:SPECIAL_NOTES_FIRSTSTRIKE}
-    [/special_note]
+#deprecated 3 1.17 This note is now added automatically.
 #enddef
 
 #define NOTE_POISON
-    [special_note]
-        note={INTERNAL:SPECIAL_NOTES_POISON}
-    [/special_note]
+#deprecated 3 1.17 This note is now added automatically.
```
stevecotton, line 243: I think this is actually `DEP_LEVEL::REMOVED` with a post-removal stub instead of `DEP_LEVEL::FOR_REMOVAL`.

data/core/macros/special-notes.cfg
```diff
@@ -299,134 +249,128 @@ _"This unit has a defense cap on certain terrain types — it cannot achieve a h
...
 # Deprecated versions here
 
 #define SPECIAL_NOTES_SPIRIT
-#deprecated 1 Use {NOTE_SPIRIT} instead (note the singular), which generates a [special_note] tag.
+#deprecated 3 1.17 Use {NOTE_SPIRIT} instead (note the singular), which generates a [special_note] tag.
 {INTERNAL:SPECIAL_NOTES_SPIRIT}#enddef
 
 #define SPECIAL_NOTES_ARCANE
-#deprecated 1 Use {NOTE_ARCANE} instead (note the singular), which generates a [special_note] tag.
+#deprecated 3 1.17 Use {NOTE_ARCANE} instead (note the singular), which generates a [special_note] tag.
```
stevecotton, line 259: This seems closer to `DEP_LEVEL::PREEMPTIVE`, a simple maintenance wrapper can be left in the code indefinitely and just means text appears in a different place on the help page.

The thing that's prompted me to look at this is that 3000 errors are reported when rebuilding the cache with Ageless Era installed. `DEP_LEVEL::PREEMPTIVE` would mean they were only warnings and that they would only be reported in the 1.17 branch, but `DEP_LEVEL::FOR_REMOVAL` means they'll be errors and reported in the 1.16 branch.

@ProditorMagnus: heads-up.

ProditorMagnus: Looks like simple case of removing all notes which show error.

data/tools/wmllint
```diff
@@ -1073,42 +1073,9 @@ def standard_unit_filter():
...
-    ("{WEAPON_SPECIAL_CHARGE}", "{NOTE_CHARGE}"),
-    ("{WEAPON_SPECIAL_DRAIN}", "{NOTE_DRAIN}"),
-    ("{WEAPON_SPECIAL_FIRSTSTRIKE}", "{NOTE_FIRSTSTRIKE}"),
-    ("{WEAPON_SPECIAL_POISON}", "{NOTE_POISON}"),
-    ("{WEAPON_SPECIAL_STUN}", "{NOTE_STUN}"),
-    ("{WEAPON_SPECIAL_SHOCK}", "{NOTE_SHOCK}"),
-    ("{WEAPON_SPECIAL_DAZE}", "{NOTE_DAZE}"),
+    ("movement_type=mounted", "{NOTE_DEFENSE_CAP}"),
+    ("movement_type=undeadspirit", "{NOTE_SPIRIT}"),
+    ("type=arcane", "{NOTE_ARCANE}"),
```
ProditorMagnus, line 1078: Any reason why arcane is not automatically added?

ProditorMagnus: Though admittedly that note seems like ancient leftover without any actual purpose.

data/core/macros/special-notes.cfg
CelticMinstrel: So you're saying I should change it from 3 to 4?

data/tools/wmllint
CelticMinstrel: You're asking me why I didn't make these three notes get automatically added?

That's because they're not linked to a weapon special or unit ability. One is linked to a damage type, the other two to movement types. I _could_ make the latter two be added automatically but the former doesn't have any "central" place to put a special note, so I would have to create such a place which is a significantly larger task.

I do plan to do both those things at some point, but not for 1.16.

data/core/macros/special-notes.cfg
CelticMinstrel: Technically, leaving this in now means the text will be shown twice on the help page, but that's still just a cosmetic issue, so sure.

What are these errors you're referring to? I didn't remove any macros, so there shouldn't be any errors; at most there should be tons of deprecation warnings. Are you referring to those as errors?

stevecotton: This doesn't seem to show the text twice, and the deprecation message says that it should be replaced with `{NOTE_ARCANE}` rather than removed.

By errors and warnings, I mean the difference between `lg::warn()` and `lg::err()`.
https://github.com/wesnoth/wesnoth/blob/faac6939f42a86cd81b79cba20104d4ef58b7ad8/src/deprecation.cpp#L42-L56

data/tools/wmllint
stevecotton: > I do plan to do both those things at some point, but not for 1.16.

Then `{NOTE_ARCANE}` should itself deprecated, and `{SPECIAL_NOTES_ARCANE}`'s deprecation message should be changed because it's currently telling UMC authors to use a deprecated macro.

data/core/macros/special-notes.cfg
stevecotton: I think so.

data/tools/wmllint
CelticMinstrel: Deprecating `{NOTE_ARCANE}` before its substitute exists doesn't seem like the correct answer…

stevecotton: So follow that logic on to whether it's right to deprecate `{SPECIAL_NOTES_ARCANE}` at this point.

CelticMinstrel: Well, it was already deprecated from the previous change that _added_ `{NOTE_ARCANE}`, and furthermore, someone using `{SPECIAL_NOTES_ARCANE}` might end up with two "Special Notes" sections, so I don't think leaving it undeprecated is a good idea.

stevecotton: So what is a UMC developer meant to do? This conversation keeps reminding me of @shikadiqueen's complaints that UMC devs keep being made to do busy-work because the core devs aren't planning.

Suggestions:
* Make the deprecation note on `{SPECIAL_NOTES_ARCANE}` say "In *version* this special note will be added automatically; there is no reason to change to `{NOTE_ARCANE}` as that will become a no-op in the same version". Change the deprecation level to match that.
* Deprecate neither tag until a long-term solution is available.

CelticMinstrel: I don't think either solution is viable. Use of `{NOTE_ARCANE}` is required to fit in with the auto-added notes from weapons and abilities, and the version when it will be replaced is unknown. It might be 1.17, or it might not be. I think the correct action for a UMC developer _is_ to use `{NOTE_ARCANE}`, although using neither is also an option.

Elvish-Hunter: Adding these three note pairs isn't really that useful, because they are already handled elsewhere in wmllint: https://github.com/wesnoth/wesnoth/blob/f1fdbc81e2a383234f1dd59c85a97383d82a7585/data/tools/wmllint#L1486-L1531.

CelticMinstrel: Huh, okay. I added them because it looked like my previous special notes commit had incorrectly removed them, but if they're handled separately, I guess I'll take them out again.

data/core/macros/special-notes.cfg
CelticMinstrel: Just noting that if you were to leave in the old `SPECIAL_NOTES` macro, you _would_ get the text displaying twice, as this example shows.

image

data/core/macros/special-notes.cfg
```diff
@@ -113,6 +113,12 @@ _"This unit has a defense cap on certain terrain types — it cannot achieve a h
     [/special_note]
 #enddef
 
+#define NOTE_ABILITY_REMOVE ABILITY
+    [+{ABILITY}]
+        special_note=""
```
stevecotton, line 118: On the White Mage's and Elvish Druid's help pages, there's a blank bullet point where the second `[heals]` ability's special note should go.

Note: the `special_note=""` of the NOTE_ABILITY_REMOVE macro is also cut&pasted without using the macro into somewhere in abilities.cfg.

data/core/macros/special-notes.cfg
stevecotton: That suggests the deprecation level on the `SPECIAL_NOTES` macro itself is too low. But that example is for one of the auto-detected abilities, not `{SPECIAL_NOTES_ARCANE}`.

data/tools/wmllint
stevecotton: > You're asking me why I didn't make these three notes get automatically added?
> 
> That's because they're not linked to a weapon special or unit ability. One is linked to a damage type, the other two to movement types. I _could_ make the latter two be added automatically but the former doesn't have any "central" place to put a special note, so I would have to create such a place which is a significantly larger task.

I'm coming to the conclusion that (doing the full task) would be easier than (handling the deprecation properly with the task half-done). It would also avoid making busy-work for UMC creators.

> version when it will be replaced is unknown. It might be 1.17, or it might not be.

In the same way that opening a PR makes an implied commitment of your own time to progressing it, I feel merging this PR should have an implied commitment to the follow-up task in 1.17. I'm feeling quite invested myself having spent time on the review; if you want me to then I'll take over getting the full task done for 1.15.14.

CelticMinstrel: Does that mean you're offering to implement:

* A `special_note` key in `[units][movetype]` (side note: there might be other places it could be supported if you wanted, eg in `[race]`, but that's not required for these specific three notes)
* A `[damage_type]` tag (probably under `[units]`? not sure) that includes `special_note` and also supercedes `[language]type_$TYPEID` as well as specifying things like damage icon and colour

If you want to commit to that, then I'm fine with it.

data/core/macros/special-notes.cfg
CelticMinstrel: Right, in that case, you'd only get a double note if you actually included both `{SPECIAL_NOTES_ARCANE}` and `{NOTE_ARCANE}`.

data/tools/wmllint
Elvish-Hunter: > Huh, okay. I added them because it looked like my previous special notes commit had incorrectly removed them, but if they're handled separately, I guess I'll take them out again.
> 
> Do you think I should remove this whole system as well, or just leave the list empty by default?

For now, just leave the list empty; once this PR is merged, I'll handle its removal, if needed.

src/units/unit.cpp
```diff
@@ -2633,6 +2633,29 @@ void unit::clear_changed_attributes()
 	}
 }
 
+std::vector<t_string> unit::unit_special_notes() const {
```
stevecotton, line 2636: It seems this function is never called, the special notes are only accessed by the help_topic_generator, which only uses the functions in units/types.cpp.

src/units/types.cpp
```diff
@@ -493,11 +493,43 @@ t_string unit_type::unit_description() const
...
+		}
+	}
+	for(const auto& attack : attacks()) {
+		for(const config::any_child ability : attack.specials().all_children_range()) {
+			if(ability.cfg.has_attribute("special_note")) {
+				return true;
+			}
+		}
+	}
+	return false;
```
stevecotton, line 509: I think the has_special_notes() function can simply be removed, as the only caller is the help_topic_generator, and if this returns true then it must call special_notes() anyway. So it could just call special_notes() and check whether the vector is empty.

My interest here is working out what to do with `NOTE_DEFENSE_CAP` - there's already logic in help_topic_generator.cpp to show the cap column if and only if it has data; that special note could be auto-added in the generator.

src/units/unit.cpp
CelticMinstrel: Hmm… I think there _should_ be some place where this is called though, even if it doesn't exist currently… after all, there's support for an individual unit to have custom special notes.

data/core/macros/special-notes.cfg
stevecotton: This was fixed in the next push.

src/units/unit.cpp
stevecotton: Probably, but as it needs to change in this PR and it's currently an untested code path, I'd prefer to just have a comment saying to consider copying it from types.cpp when someone has a test-case.

src/units/types.cpp
stevecotton: This comment is as much notes to myself as a reply, still trying to work out what fits better.

At the moment the few translatable strings in src/units/*.cpp are all in textdomain `wesnoth`, while both special-notes.cfg and help_topic_generators.cpp are in `wesnoth-help`. I'd prefer to keep it in the same domain as the other notes.

For the current PR, there should be logic somewhere so that a unit with two arcane attacks only gets one copy of the note on its help page. In future, but not for this PR, it might be nice to have a way to hide some special notes that the user is familiar with. Those two bits of logic could fit together, OTOH they're probably trivial logic (if not trivial UI) anyway.

src/units/unit.cpp
CelticMinstrel: That would make more sense if not for the fact that it's basically a clone of the equivalent function in `unit_type`.

src/units/types.cpp
CelticMinstrel: > For the current PR, there should be logic somewhere so that a unit with two arcane attacks only gets one copy of the note on its help page.

What are you talking about? The only way you could get two copies of the note with the current PR is is you included both `{SPECIAL_NOTES_ARCANE}` and `{NOTE_ARCANE}`, which you're not supposed to do.

The logic you describe would only become important if the arcane note were automatically added.

src/units/unit.cpp
stevecotton: Why would it make sense to have an unused and untested codepath instead of a comment?

CelticMinstrel: If it's essentially copy-paste of a used and tested codepath then it's not much different from it itself being used and tested.

irydacea:
So if I understand correctly:

* `special_note=` is part of abilities and weapon specials. If present, any unit with the ability/special in question gets the note added to its help page in an unspecified order*.
* `[special_note]` (existing in 1.15.x) is part of unit types. If present, the unit gets the note added to its help page in an unspecified order*.
* For notes that can be bound to an ability or special, the new `NOTE_FOO` macros no longer do anything other than emit deprecation warnings hinting at their upcoming removal in 1.17.0.
* For notes that can't be bound to an ability or special, we continue to recommend and use a handful of `NOTE_FOO` macros.
* Old `SPECIAL_NOTES_FOO` macros continue to add text. Is the reason for this that some people might have used them outside of the intended context? Also, this way, a unit that has Magical and hasn't had its `SPECIAL_NOTES+SPECIAL_NOTES_MAGICAL` removed when ported to 1.16 will have its description look roughly like this: `[...] Special notes: has magical [...] Special notes: has magical" (note the double notes).
* Stuff about `wmllint` that I'm sure E_H can handle without me having to look at it.

(* I'm sure the order is well specified in the C++, it's just that I haven't look at it yet.)

Is all this correct?

CelticMinstrel:
All of that is correct.

src/units/unit.cpp
stevecotton: @irydacea, please can we have your opinion on this? I think it would be better to just delete the unused code.

data/core/macros/special-notes.cfg
stevecotton: The reason that I haven't marked this resolved is that I still think the DEP_LEVEL on `SPECIAL_NOTES` ought to be higher.

src/units/unit.cpp
irydacea: Which function is used for the sidebar description tooltips? Considering that units having different attributes than their unit types is actually an extremely common thing.

data/tools/wmllint
stevecotton: > Does that mean you're offering to implement:

I think my offer is:
* Add `[units][movetype][special_note]` to replace `NOTE_SPIRIT`.
* Replace `NOTE_DEFENSE_CAP` with something that's auto-added. If we don't agree about where the auto-adding code should go, I'll move the note to `[units][movetype][special_note]` for the first PR and then do a follow-up PR changing it to auto-adding.
* While agreeing that it's a hack, add `[language]special_note_damage_type_arcane` and auto-add `special_note_damage_type_$TYPEID` to any unit with a `TYPEID` attack.

I like your idea of replacing `[language]type=$TYPEID` with a `[damage_type]` tag, but feel it's too big a change to get in to 1.16. While I can see some UMC authors using `[language]special_note_damage_type_electric` etc, it will be far fewer authors and far fewer locations that `NOTE_SPIRIT`; so I'd be happy about adding that for 1.16 and then deprecating it in 1.18.

src/units/unit.cpp
stevecotton: Thanks for that question, I've opened PR #5808 to include the special notes in the tooltip. Haven't tested it with units that have different attributes to their unit types, but handling that seems more suited to the current PR (and if it needed any changes then it would add a merge conflict with the current PR).

@CelticMinstrel the cut & pasted code in the current PR has a bug that causes duplicated special notes for attacks, an example unit that shows this is the Lancer.

data/tools/wmllint
CelticMinstrel: I think that's pretty much an acceptable compromise, but why `[units][movetype][special_note]` instead of [units][movetype]special_note=`?

src/units/unit.cpp
CelticMinstrel: Okay, so I'll need to call `std::unique` too…

data/tools/wmllint
stevecotton: I hadn't thought it over when I posted, but the UTBS units with NOTE_ETHEREAL have two special notes that should probably both move to their movetype.

CelticMinstrel: I didn't look into what ethereal really connects to… is it present exactly when (if and only if) a specific movetype is used? If so, then sure.

stevecotton: Having looked more at NOTE_ETHEREAL, there isn't a corresponding movetype. It's used in a set of specialised ghosts, which have the base unit's movetype and then some custom movement costs on top.

stevecotton:
Should I add comments here for things which I think should change, and which I've already changed in #5811?

stevecotton:
I think the whole of #5811 is now ready to merge; it includes a rebased (to fix the merge conflict) copy of this PR. Unless this one gets new changes (other than just a rebase), I think it should close and the focus should be on getting #5811 in.

data/tools/wmllint
stevecotton: > For now, just leave the list empty; once this PR is merged, I'll handle its removal, if needed.

@Elvish-Hunter: it ended up merging as #5811 instead of #5798, but please would you do the update for wmllint?

PR #5800, opened 2021-05-21
Move translation mark validation from wmllint to schema

CelticMinstrel:
With this, the schema validator can now distinguish between a translatable and a non-translatable string.

* The `t_string` type, which must be translatable, is now a schema built-in type (the only built-in).
* Type definitions now support an `allow_translatable` key, which defaults to false, meaning the value _must not_ be translatable. If set to true, the value is allowed to be translatable. Translatable values in this key will still be matched against the pattern.


data/gui/window/network_transmission.cfg
```diff
@@ -95,7 +95,6 @@
 									definition = "gold_small"
 									# NOTE: needed to avoid explicit calls to invalidate_layout()
```
stevecotton, line 96: This leaves a dangling comment that makes no sense without the line that it referred to.

data/core/terrain.cfg
```diff
@@ -1276,7 +1276,9 @@ Occasionally caves are <italic>text='illuminated'</italic>."
     #editor_image=cave/earthy-wall-hewn-tile
     id=cavewall_earthy_hewn
     name= _ "Cave Wall"
-    editor_name="Reinforced Earthy Cave Wall" # wmllint: ignore
+#ifndef SCHEMA_SHOULD_SKIP_THIS
+    editor_name="Reinforced Earthy Cave Wall"
```
stevecotton, line 1280: This should be translatable.

data/core/units/dunefolk/Wyvern_Rider.cfg
```diff
@@ -14,7 +14,7 @@
     {AMLA_DEFAULT}
     cost=102
     usage=scout
-    description=""
+    #description=""
```
stevecotton, line 17: I feel a `[unit_type]` that lacks a description should fail validation.

utils/CI/schema_validation.sh
```diff
@@ -8,7 +8,7 @@ function validate_core()
     echo "------"
     echo "Validating $NAME..."
     
-    ./wesnoth --validate data/_main.cfg &> temp.log || SUCCESS="No"
+    ./wesnoth --validate data/_main.cfg --preprocess-defines=SCHEMA_SHOULD_SKIP_THIS &> temp.log || SUCCESS="No"
```
soliton-, line 11: I'd rather call the define something like `SCHEMA_VALIDATION`.

data/core/terrain.cfg
CelticMinstrel: The reason I left it untranslated was because it's deprecated… although if there's a non-deprecated terrain with the same name, I may as well just make it translated…

data/core/units/dunefolk/Wyvern_Rider.cfg
CelticMinstrel: My thought was that maybe I should just make it consider an empty string without translation mark to be a valid t_string. Does that make sense?

CelticMinstrel:
I'm thinking I'll squash this on merging, so I split the commits up into easier-to-review chunks in case anyone wants to limit their review to a particular subset.

data/schema/core/conditionalwml.cfg
```diff
@@ -25,17 +25,17 @@
...
+		{SIMPLE_KEY equals any}
+		{SIMPLE_KEY not_equals any}
+		{SIMPLE_KEY numerical_equals s_int}
+		{SIMPLE_KEY numerical_not_equals s_int}
+		{SIMPLE_KEY greater_than s_int}
+		{SIMPLE_KEY greater_than_equal_to s_int}
+		{SIMPLE_KEY less_than s_int}
+		{SIMPLE_KEY less_than_equal_to s_int}
+		{SIMPLE_KEY boolean_equals s_bool}
+		{SIMPLE_KEY boolean_not_equals s_bool}
```
CelticMinstrel, line 38: Changing these to s_int and s_bool should maybe be a separate commit? It has nothing to do with the tstring stuff.

data/schema/types/basic.cfg
```diff
@@ -25,14 +25,14 @@
...
-	name=string
+	name=any
 	value=".*"
+	allow_translatable=yes
 [/type]
 [type]
-	name=t_string
-	value="_?.*"
+	name=string
+	value=".*"
```
CelticMinstrel, line 34: Note to self: I should swap the order of these types for a better diff.

data/campaigns/tutorial/scenarios/01_Tutorial_part_1.cfg
```diff
@@ -138,8 +138,63 @@
...
+        #  [/set_menu_item]
+        #  [lua]
+        #      code=<<
+        #          wesnoth.interface.set_menu_item(
+        #              'test',
+        #              {description="Test"},
+        #              function() gui.alert("Even this works!") end
+        #          )
+        #      >>
+        #  [/lua]
```
CelticMinstrel, line 197: Whoops, this is old test code from the events PR, it shouldn't be in here.

data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
```diff
@@ -714,9 +713,8 @@
 
                 [floating_text]
                     x,y=$cunit.x,$cunit.y
-                    # wmllint: markcheck off - nothing to translate here
-                    text="<span color='#ff0000'>" + "2" + "</span>"
-                    # wmllint: markcheck on
+                    color=255,0,0
+                    text=2
```
stevecotton, line 717: While all the places changed in this PR use a single color, this wouldn't be able to handle more complex formatting. For example, a hitpoints-to-gold or hitpoints-to-magic_points conversion that has a red negative and a yellow or blue positive number.

data/campaigns/Heir_To_The_Throne/scenarios/17_Scepter_of_Fire.cfg
```diff
@@ -6,7 +6,7 @@
     map_generation=lua
     [generator]
         id="cavegen"
-        config_name="Lua Cave Generator"
+        config_name=_"Sceptre of Fire Caves"
```
stevecotton, line 9: Where is this shown?

data/test/macros/wml_unit_test_macros.cfg
```diff
@@ -71,7 +71,7 @@ human#endarg
 Elvish Archer#endarg
 
     [test]
-        name = "Unit Test " + {NAME}
+        name = _ "Unit Test " + {NAME}
```
stevecotton, line 74: This doesn't feel like a good change - if the schema is checking for a mark, but isn't complaining about the untranslatable `NAME`.

stevecotton:
I think the changes to Dead Water and SotBE are bugfixes that should be cherry-picked.

data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
CelticMinstrel: I don't really get your example, but you're correct that more complex formatting could just include the translation mark.

data/campaigns/Heir_To_The_Throne/scenarios/17_Scepter_of_Fire.cfg
CelticMinstrel: I don't think it's shown anywhere, because this file is not parsed in map editor mode.

If it _was_ parsed in map editor mode, you would see it in the Generate Map dialog.

data/test/macros/wml_unit_test_macros.cfg
CelticMinstrel: A value is considered translatable if any segment of it is translatable, and it's fairly common to concatenate variously-translatable parts in this way (for example, untranslatable formatting codes and translatable content). This at least avoids the trap of putting a macro inside the translatable string, and since this is in wesnoth-test, it won't be sent to the translators anyway.

data/gui/window/debug_clock.cfg
```diff
@@ -149,7 +149,7 @@
 						minimum_value = 0
 						maximum_value = 3599
 						step_size = 1
-						label = 3599
+						label = _ "3599"
```
stevecotton, line 152: Should `[slider]label=` be translatable? It looks like it needs to be a number.

data/test/maps/pathfind_1.cfg
```diff
@@ -61,7 +61,6 @@ victory_when_enemies_defeated=yes
     lawful_bonus=0
     mask=""
     #textdomain wesnoth-help
```
stevecotton, line 63: Is this line still necessary?

data/gui/window/debug_clock.cfg
CelticMinstrel: I think it doesn't _need_ to be a number but usually will be… it also won't usually be set in the WML, since it needs to update dynamically as the slider moves.

This is the debug clock, so I don't really care if it's a bit weird, but I can change something if that seems like a good idea.

data/test/maps/pathfind_1.cfg
CelticMinstrel: Nope!

utils/CI/schema_validation.sh
CelticMinstrel: Done, and I also defined it in the C++ instead of in the CI script.

src/theme.cpp
```diff
@@ -431,7 +431,7 @@ theme::label::label()
 
 theme::label::label(std::size_t sw, std::size_t sh, const config& cfg)
 	: object(sw, sh, cfg)
-	, text_(cfg["prefix"].str() + cfg["text"].str() + cfg["postfix"].str())
+	, text_(cfg["prefix"].str() + cfg["prefix_literal"].str() + cfg["text"].str() + cfg["postfix_literal"].str() + cfg["postfix"].str())
```
Pentarctagon, line 434: What is prefix_literal and postfix_literal?

src/gui/dialogs/network_transmission.cpp
```diff
@@ -113,6 +113,9 @@ void network_transmission::pre_show(window& window)
 		subtitle_label.set_use_markup(true);
 	}
 
+	// NOTE: needed to avoid explicit calls to invalidate_layout()
+	// in network_transmission::pump_monitor::process()
+	find_widget<label>(&window, "numeric_progress", false).set_label(" ");
```
Pentarctagon, line 118: What is this doing?

src/theme.cpp
CelticMinstrel: It's pretty much exactly what it says – a literal, untranslated string that prefixes or postfixes the main string.

src/gui/dialogs/network_transmission.cpp
CelticMinstrel: I honestly have no idea. The WML set the label to a single space with this comment, but that makes it fail validation, so I figured setting it in the C++ would have the same effect.

src/theme.cpp
Pentarctagon: Why is that being added with this PR?

src/gui/dialogs/network_transmission.cpp
Pentarctagon: Given there's no longer any explicit call to invalidate_layout in network_transmission::pump_monitor::process(), is it even still needed?

src/theme.cpp
CelticMinstrel: It was so that certain things in the theme didn't have to be translated. In particular I think there were several cases of strings consisting of only spaces.

src/gui/dialogs/network_transmission.cpp
CelticMinstrel: That's a good question! The comment suggests removing this would mean we need to add explicit calls there, but I have no idea why a single space would make that difference.

Pentarctagon: Looks like https://github.com/wesnoth/wesnoth/commit/34d45d3ee5ff4d3ec939e5ae0272a292eefce60f added this in WML.

Pentarctagon (approved):
Overall I guess this seems like pretty uncontroversial change.  If this is ever removed from wmllint as well, then at that point I'd expect there to be more discussion about schema being accessible, etc.

PR #5804, opened 2021-05-22
 Create a [anti_special] ability who negate specifically some specials

newfrenchy83:
With this ability, it is possible to negate the effects of specials without ipmlement filter in this special
and apply that to core specials or abilities used like specials.
Another advantage is the [affect_adjacent] but also to affect a special unique without affect special of same type who could be used in same attack.
an exemple can be to cancel the effect of magical chance_to_hit without affect the marsksman abilities or to create different type of poisons


newfrenchy83:
@CelticMinstrel  @Pentarctagon  could you review this PR, please?

CelticMinstrel:
I still don't see how this adds anything that can't be done already.

newfrenchy83:
> 
> 
> I still don't see how this adds anything that can't be done already.

[anti_special] can replace overwrite_specials while being more flexible and can also be used to neutralize the opponent's weapons selectively, but also condition the use, for example for poisons, we can use filter_opponent to neutralize only the poison orcs, considering that the target has the antidote against this precise mixture, but not against a toxin used by other fighters (spiders or scorpions), in this case filtering the race in anti_special avoids being used [event ] to heal the target after being hit, either to implement a special [poison] programmed to be inactive against a unit affected by ability and have to duplicate core units to do so or to use [object] ( using object for a large number of units can end up slowing down the game).

For overwrite_specials, one of the things that bothered me was that a special weapon would stay marked as active even when overwrited by an ability used as a special, either through the UI or the special weapon filters. , which can be annoying in some cases (a special weapon that activates when the opponent uses a weapon supposed to be neutralized), or relevant in others.

So yes, [anti_special] doesn't do anything that can't be done already, but it makes it easier to do, with less heavy lua or wml coding, and that's also the reason why I implemented the abilities used as weapons, allowing teaching without heavy programming as in Loti, we could already do it, but the code was heavier to write.

Pentarctagon:
The abilities/weapon specials code can't just keep getting more and more and more complex though.

newfrenchy83:
> 
> 
> The abilities/weapon specials engine code can't just keep getting more and more and more complex though.

I understand, and i will forget this.

PR #5805, opened 2021-05-23
A few WFL updates

CelticMinstrel:
Opening a PR to ensure CI passes; assuming it does, I'll merge tomorrow.

I don't expect much controversy in this PR, but it:

* Deprecates "side" in WFL formulas in favour of "side_number", which is consistently 1-based
* Adds tod_bonus and base_tod_bonus gamestate functions
* Adds "terrain" to unit objects, which looks up the terrain info for the unit's current location

### DIFF
--- a/src/formula/callable_objects.cpp
+++ b/src/formula/callable_objects.cpp
@@ -26,7 +26,9 @@
 #include "log.hpp"
 #include "recall_list_manager.hpp"
 #include "deprecation.hpp"
-#include "version.hpp"
+#include "game_board.hpp"
+#include "game_version.hpp"
+#include "resources.hpp"
 
 static lg::log_domain log_scripting_formula("scripting/formula");
 #define LOG_SF LOG_STREAM(info, log_scripting_formula)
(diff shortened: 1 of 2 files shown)


src/formula/callable_objects.cpp
```diff
@@ -688,6 +704,9 @@ void team_callable::get_inputs(formula_input_vector& inputs) const
 variant team_callable::get_value(const std::string& key) const
 {
 	if(key == "side") {
+		deprecated_message("team.side", DEP_LEVEL::FOR_REMOVAL, version_info("1.17"), "Use side_number instead.");
+		return variant(team_.side());
+	} else if(key == "side_number") {
```
stevecotton, line 709: Calling it `side` seems more consistent with SideWML.

stevecotton (commented):
Unit tests would be good.

While deprecating the zero-based data is probably good, the documentation on the Wiki doesn't warn about these being zero-based. Updating the Wiki would be good.

src/formula/callable_objects.cpp
CelticMinstrel: I deprecated this one because I think it's confusing if it's called "side_number" in one place but "side" in another. I could lower the level on this one, though.

PR #5808, opened 2021-05-26
Include the new-style special notes in the sidebar's unit description

stevecotton:
_@irydacea asked in https://github.com/wesnoth/wesnoth/pull/5798#discussion_r639428602_:
> Which function is used for the sidebar description tooltips? Considering that singular units having different attributes than their unit types is actually an extremely common thing.

There seems to be some bug about triggering this tooltip - for me hovering the mouse over the sidebar reliably shows other tooltips, but is very intermittent with this one. However, the subject of this PR is the content of the tooltip when it does show.

### DIFF
--- a/src/reports.cpp
+++ b/src/reports.cpp
@@ -192,6 +192,12 @@ static config unit_type(const unit* u)
 	str << u->type_name();
 	tooltip << _("Type: ") << "<b>" << u->type_name() << "</b>\n"
 		<< u->unit_description();
+	if(const auto& notes = u->unit_special_notes(); !notes.empty()) {
+		tooltip << "\n\n" << _("Special Notes:") << '\n';
+		for(const auto& note : notes) {
+			tooltip << "• " << note << '\n';
+		}
+	}
 	return text_report(str.str(), tooltip.str(), has_variations_prefix + "unit_" + u->type_id());
 }
 REPORT_GENERATOR(unit_type, rc)
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,7 @@
  ### User interface
  ### WML Engine
  ### Miscellaneous and Bug Fixes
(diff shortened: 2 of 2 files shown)


CelticMinstrel:
Looks good… the only other thing I'd recommend is to take care that `_("Special Notes")` is in the correct textdomain, which I suspect is currently _not_ the case.

stevecotton:
For a two-word string, I feel it isn't worth using a textdomain other than the one that reports.cpp defaults to.

CelticMinstrel:
This is more of a "we shouldn't move the string to another textdomain" concern. If reports.cpp and special-notes.cfg use the same textdomain already, then it's fine.

stevecotton:
They're not the same textdomain, so it indeed would cause a duplication. However, there's already some duplication between textdomains, and while that does add to the translators' workload, I feel a two-word string isn't the place to worry about that.

irydacea:
No problem with the trivial duplicate. This is more common than you'd think, and "fixing" it with a verbose gettext call is likely to cause issues in the future if someone changes the string and forgets to update other references to it.

src/reports.cpp
```diff
@@ -192,6 +192,13 @@ static config unit_type(const unit* u)
 	str << u->type_name();
 	tooltip << _("Type: ") << "<b>" << u->type_name() << "</b>\n"
 		<< u->unit_description();
+	std::vector<t_string> notes = u->unit_special_notes();
```
Vultraz, line 195: Suggest
```cpp
if(const auto notes = u->unit_special_notes(); !notes.empty())
```
since it's only used once

Vultraz (changes requested):

src/reports.cpp
stevecotton: Done.

PR #5811, opened 2021-05-28
Refactor special notes for abilities, attack types, movetypes and weapon specials

stevecotton:
Many unit's files are changed here, but the main files to review are the C++ and those under data/core/macros.

An easier way of setting special notes in the most common use-cases. Text given
in the following attributes will be collected and added to the special notes
for units and unit types:

* [ability tag name]special_note=
* [language]special_note_damage_type_TYPE=
* [movetype][special_note]note=
* [attack][specials][special tag name]special_note=

It's no longer necessary to put these notes in each unit_type's .cfg file, and
the macros for doing so are now deprecated.

C++ changes
-----

Simplify both unit_type::special_notes and unit::unit_special_notes. Add
utils::stable_unique, similar to std::unique but accepts non-ordered input and
preserves the order in the output.

Remove unit_type::has_special_notes() - callers can instead call
special_notes() and then check if the returned vector is empty, which removes
the need for duplicating code in unit_type.

Trade-off: the new [language]special_note_damage_type_TYPE is likely deprecated in 1.19.
-----

Adding [language]special_note_damage_type_TYPE= uses the same existing design
as [language]type_TYPE=, however both are hacks that don't fit the general
style of WML. It could be better to define a new [damage_type] tag that
supercedes both and also provides a place for specifying the damage icon;
however that won't be done in time for the API freeze for 1.16.

Doing it in the way that this commit does it is a hack, but it's one where
replacing it with the better solution in 1.18 will affect very few UMCs (only
those that define additional damage types). Even in the UMCs that would be
affected, it would likely only be a few changes in a single central file.

Trade-off: NOTE_DEFENSE_CAP is not auto-added
-----

It might be better to auto-add NOTE_DEFENSE_CAP when movetype.cpp detects that
the type has capped values. However as NOTE_SPIRIT already requires
[movetype][special_note], it's simple to use the same mechanism. If we decide
to change it to being auto-added, the current commit greatly reduces the number
of places that would need to change again, as it's now in the [movetype]
instead of the many [unit_type]s using that movetype.


src/units/types.hpp
```diff
@@ -440,3 +448,13 @@ struct unit_experience_accelerator {
 private:
 	int old_value_;
 };
+
+/**
+ * Common logic for unit_type::special_notes() and unit::special_notes(). Adds
+ * any notes from the sources given as arguments, and filters out duplicates.
+ *
+ * \todo: move this out of the global namespace
```
stevecotton, line 456: Where should this move to?

stevecotton:
I'm keeping it in draft mode because I want to do another re-review of it myself, but I think this is close to ready.

src/units/types.hpp
CelticMinstrel: I can't think of a good place for it.

CelticMinstrel:
I intend to merge #5798 before you merge this, so maybe you shouldn't mention the first half of the work in your commit message…

stevecotton:
> I intend to merge #5798 before you merge this, so maybe you shouldn't mention the first half of the work in your commit message…

I squashed all of it (including all of #5798) into one commit, I don't think it makes sense to split it:
* Around 60 of the unit files have changes on adjacent lines, for example removing `{NOTE_MAGICAL}` and removing `{NOTE_ARCANE}`.
* The C++ changes are different, with this branch adding utils::stable_unique and deleting unit_type::has_special_notes.

stevecotton:
The latest push adds _NO_NOTES versions of some of the healing macros. Instead of needing to treat a blank note as a removed note, this allows the following (diff against the implementation in #5798):

```
 #define ABILITY_CURES
     # Canned definition of the cure ability (which entails heal+8) to be
     # included in an [abilities] clause..
-    {ABILITY_UNPOISON}
+    {INTERNAL:ABILITY_UNPOISON_NO_NOTES}
+    {INTERNAL:ABILITY_EXTRA_HEAL_NO_NOTES}
     [+heals]
         special_note={INTERNAL:SPECIAL_NOTES_CURES}
     [/heals]
-    {ABILITY_EXTRA_HEAL}
-    {NOTE_ABILITY_REMOVE heals}
 #enddef
 ```

CelticMinstrel:
> I don't think it makes sense to split it:

It's a question of preserving authorship information.

stevecotton:
> It's a question of preserving authorship information.

If you want me to put your name in Git's author field then I'd be happy to.

However, the code in these PRs is going in different directions - if both PRs go in then this one's diff will look like a removal of #5798's implementation.

If you really want split by author, then I suggest making a commit with this PR's versions of src/utils/general.hpp, src/help/help_topic_generators.cpp and src/tests/test_util.cpp; and then rebasing #5798 with those changes already available.

CelticMinstrel:
> However, the code in these PRs is going in different directions

I don't think that's true in general. Your work is mostly built on mine, just factoring out the logic even more than I did. Your `combine_special_notes` function is almost identical to my `special_notes` function.

…that said, I took a look at it just now and spotted what appears to be a regression relative to #5798. Your implementation does not exclude blank entries from the special notes. I think it's correct to do this for two reasons:

* A special note might be added by a macro, where someone might pass an empty string as the note. There shouldn't be an empty bullet as a result.
* The strategy I used for the heal abilities was to set the special note to empty. (`[special_note]remove=yes' would also work for that, but it doesn't fit well in the macro format and would make the macro less widely usable.)

> if both PRs go in then this one's diff will look like a removal of #5798's implementation.

Yeah, I don't think "removal" is the right word here, but even if it were an outright removal, I don't think that's a problem.

stevecotton:
> A special note might be added by a macro, where someone might pass an empty string as the note. There shouldn't be an empty bullet as a result.

Passing an empty string in that case sounds like a bug in the place using the macro, and not displaying the blank note (as a hint to fix the place using the macro) sounds like a bug in the engine, IMO.

> The strategy I used for the heal abilities was to set the special note to empty.

I think https://github.com/wesnoth/wesnoth/pull/5811#issuecomment-850977449 is a better way to achieve that.

CelticMinstrel:
> Passing an empty string in that case sounds like a bug in the place using the macro, and not displaying the blank note (as a hint to fix the place using the macro) sounds like a bug in the engine, IMO.

I disagree. In such a situation, fixing the macro (either definition or call site) is probably impossible. I ran into some cases like this in #5800, where (if I recall correctly) a macro created a widget with a tooltip, but one of the call sites wanted to omit the tooltip. The only possibility in such a situation is to pass an empty string to the macro; it's not possible for the macro to omit the attribute altogether based on input.

> I think #5811 (comment) is a better way to achieve that.

It does solve it without relying on an empty note, sure. I don't think it's necessarily a better solution, but it works.

stevecotton:
Summarizing IRC discussion, the support for not displaying blank notes will be re-added.

stevecotton:
This is now ready to merge, IMO. I've done basic testing on it, checking the notes for white mages, knights, UtBS' fairy units, UtBS' ethereal ghosts, and in SXC checked the notes for white mages that lose their arcane attack, as well as units with multiple arcane attacks.

Note that the sidebar tooltip shows the notes for the current unit, but opening the help-browser by right-clicking on a unit shows the notes for generic units of that type. These can be different if units have `[modifications]`, and this is deliberate.

Interaction with the #5798: the first commit of this one is #5798 rebased to resolve the merge conflict that prevents that one from merging. Unless there are more changes done to #5798 (other than just rebasing), then this one should merge and #5798 should then close as obsolete. The changes in this one are now separated again with one commit per author.

CelticMinstrel (approved):

PR #5815, opened 2021-05-29
AToTB S02: The necromancer is leaving, not waiting for reinforcements

stevecotton:
Also change a line from the elves, so that it changes from sympathy to defense
instead of a sudden swing from sympathy to aggression.

@Earth-Cake, @Konrad22, @Lordlewis6 and @nemaara: this is a follow-up to #5255. I'm assuming that AToTB won't be replaced with a rewritten version in time for 1.16, so want to improve up the existing dialogue.

### DIFF
--- a/data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
+++ b/data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
@@ -263,7 +263,7 @@ Besides... I want my brother back."
                         boolean_equals=yes
                     [/variable]
                 [/show_if]
-                description= _ "Kill the Dark Adept before his reinforcements arrive"
+                description= _ "Kill the Dark Adept before more elves arrive"
                 condition=win
             [/objective]
             [objective]
@@ -391,7 +391,7 @@ Besides... I want my brother back."
             speaker=Nil-Galion
             message= _ "We heard the sounds of battle three days ago, but the foul reek of death was unmistakable as the prisoner passed us; if your brother really is the prisoner then he is also a necromancer. Remember your brother as we all hope he was, and know that the prisoner’s fate will be revenge as dire as you could wish.
 
-Either way, nothing will be gained by your intrusion into our woods. Now advance no further, or you will die."
+Either way, nothing will be gained by your intrusion into our woods. Now advance no further, for we will defend our land."
         [/message]
 
         [message]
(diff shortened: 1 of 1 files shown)


data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
```diff
@@ -458,7 +458,8 @@ Either way, nothing will be gained by your intrusion into our woods. Now advance
 
         [message]
             speaker=Muff Toras
-            message= _ "Curses! If they had been an hour slower, our master’s reinforcements would already be here to meet us."
+            # po: There are no undead on the map yet, but Muff Toras summons some immediately after saying this
+            message= _ "Curses! Had they been a few hours slower, I might covered our tracks without arousing the elves’ suspicions. Still, no need for the disguise any more."
```
nemaara, line 462:
```suggestion
            message= _ "Curses! Had they been a few hours slower, I might have covered our tracks without arousing the elves’ suspicions. Still, no need for the disguise any more."
```

data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
```diff
@@ -699,15 +700,24 @@ Either way, nothing will be gained by your intrusion into our woods. Now advance
...
+            [then]
+                [message]
+                    speaker=Arvith
+                    message= _ "These elves have delayed us too long, the kidnappers’ trail will have faded by now."
+                [/message]
+            [/then]
+            [else]
+                [message]
+                    speaker=Arvith
+                    message= _ "More elves are arriving; even if they believe me, the kidnappers’ trail is sure to fade before we can follow."
```
nemaara, line 717:
```suggestion
                    message= _ "More elves are arriving! The kidnappers’ trail is sure to fade before we can follow."
```

nemaara (commented):
Second one is only a suggestion, take it or leave it is fine with me either way.

data/campaigns/Two_Brothers/scenarios/02_The_Chase.cfg
stevecotton: Thanks. I'm going to leave the "even if they believe me" in, because it doesn't imply that Arvith could defeat the elvish reinforcements. I'm assuming a few scouts would slow down Arvith, and a large number of reinforcements would follow.

I guess the dark adept (who gets released alive at the start of S03) finds out that there's an army of elves following Arvith and waiting for the Council's decision; but that's not our protagonists' problem.

PR #5818, opened 2021-05-29
Add a dir() function to improve API discoverability

CelticMinstrel:
This is intended to allow someone to explore the API right in the Lua console. The current commit implements the actual function, but there's a lot more work to support it throughout the API, which is why this is a draft. I don't expect this to be finished for 1.16.

### DIFF
--- a/src/scripting/lua_kernel_base.cpp
+++ b/src/scripting/lua_kernel_base.cpp
@@ -47,6 +47,7 @@
 #include "utils/name_generator.hpp"
 #include "utils/markov_generator.hpp"
 #include "utils/context_free_grammar_generator.hpp"
+#include "utils/scope_exit.hpp"
 
 #include <cstring>
 #include <exception>
@@ -54,6 +55,7 @@
 #include <string>
 #include <sstream>
 #include <vector>
+#include <numeric>
 
 #include "lua/lauxlib.h"
 #include "lua/lua.h"
@@ -446,6 +448,96 @@ static int intf_get_language(lua_State* L)
 	return 1;
 }
 
+static void dir_meta_helper(lua_State* L, std::vector<std::string>& keys)
+{
+	switch(luaL_getmetafield(L, -1, "__dir")) {
+		case LUA_TFUNCTION:
+			lua_call(L, 0, 1);
+			[[fallthrough]];
+		case LUA_TTABLE:
+			auto dir_keys = lua_check<std::vector<std::string>>(L, -1);
+			std::copy(dir_keys.begin(), dir_keys.end(), std::back_inserter(keys));
+			break;
+	}
+}
+
+/**
+ * Prints out a list of keys available in an object.
+ * A list of keys is gathered from the following sources:
+ * - For a table, all keys defined in the table
+ * - Any keys accessible through the metatable chain (if __index on the metatable is a table)
+ * - The output of the __dir metafunction
+ * The list is then sorted alphabetically and formatted into columns.
+ * - Arg 1: Any object
+ * - Arg 2: (optional) Whether to include write-only keys (passed on to the __dir metamethod)
+ * - Arg 3: (optional) Function to use for output; defaults to _G.print
+ */
+static int intf_object_dir(lua_State* L)
+{
+	if(lua_isnil(L, 1)) return luaL_argerror(L, 1, "Can't dir() nil");
+	if(lua_isboolean(L, 2)) {
+		include_write_only = luaW_toboolean(L, 2);
+	}
+	if(!lua_isfunction(L, -1)) {
+		luaW_getglobal(L, "print");
+	}
+	int fcn_idx = lua_gettop(L);
+	std::vector<std::string> keys;
+	if(lua_istable(L, 1)) {
+		// Walk the metatable chain (as long as __index is a table)...
+		// If we reach an __index that's a function, check for a __dir metafunction.
+		int save_top = lua_gettop(L);
+		lua_pushvalue(L, 1);
+		ON_SCOPE_EXIT(&) {
+			lua_settop(L, save_top);
+		};
+		do {
+			int table_idx = lua_absindex(L, -1);
+			for(lua_pushnil(L); lua_next(L, table_idx); lua_pop(L, 1)) {
+				if(lua_isstring(L, -2)) {
+					keys.push_back(lua_tostring(L,-2));
+				}
+			}
+			// Two possible exit cases:
+			// 1. getmetafield returns TNIL because there is no __index
+			// In this case, the stack is unchanged, so the while condition is still true.
+			// 2. The __index is not a table
+			// In this case, obviously the while condition fails
+			if(luaL_getmetafield(L, table_idx, "__index") == LUA_TNIL) break;
+		} while(lua_istable(L, -1));
+		if(lua_isfunction(L, -1)) {
+			lua_pop(L, 1);
+			dir_meta_helper(L, keys);
+		}
+	} else if(lua_isuserdata(L, 1)) {
+		dir_meta_helper(L, keys);
+	}
+	// Sort and remove any duplicates
+	std::sort(keys.begin(), keys.end());
+	keys.erase(std::unique(keys.begin(), keys.end()), keys.end());
+	size_t max_len = std::accumulate(keys.begin(), keys.end(), 0, [](size_t max, const std::string& next) {
+		return std::max(max, next.size());
+	});
+	// Let's limit to about 80 characters of total width with minimum 3 characters padding between columns
+	static const size_t MAX_WIDTH = 80, COL_PADDING = 3;
+	size_t n_cols = (MAX_WIDTH + COL_PADDING) / (max_len + COL_PADDING);
+	size_t col_width = max_len + COL_PADDING;
+	size_t n_rows = ceil(keys.size() / double(n_cols));
+	for(size_t i = 0; i < n_rows; i++) {
+		std::ostringstream line;
+		line.fill(' ');
+		line.setf(std::ios::left);
+		for(size_t j = 0; j < n_cols && j + (i * n_cols) < keys.size(); j++) {
+			line.width(col_width);
+			line << keys[j + i * n_cols];
+		}
+		lua_pushvalue(L, fcn_idx);
+		lua_push(L, line.str());
+		lua_call(L, 1, 0);
+	}
+	return 0;
+}
+
 // End Callback implementations
 
 // Template which allows to push member functions to the lua kernel base into lua as C functions, using a shim
@@ -517,6 +609,10 @@ lua_kernel_base::lua_kernel_base()
 	lua_setglobal(L, "dofile");
 	lua_pushnil(L);
 	lua_setglobal(L, "loadfile");
+
+	// Add dir()
+	lua_pushcfunction(L, intf_object_dir);
+	lua_setglobal(L, "dir");
 
 	// Store the error handler.
 	cmd_log_ << "Adding error handler...\n";


stevecotton:
Does this interact with the already-implemented tab completion?

CelticMinstrel:
It currently doesn't… I _was_ thinking that unifying them might be nice though.

src/scripting/lua_kernel_base.cpp
```diff
@@ -446,6 +448,100 @@ static int intf_get_language(lua_State* L)
...
+			break;
+	}
+}
+
+/**
+ * Prints out a list of keys available in an object.
+ * A list of keys is gathered from the following sources:
+ * - For a table, all keys defined in the table
+ * - Any keys accessible through the metatable chain (if __index on the metatable is a table)
+ * - The output of the __dir metafunction
```
stevecotton, line 472: How will this filter out deprecated stuff?

Replying to https://github.com/wesnoth/wesnoth/pull/5965#issuecomment-885810718 :
> Renaming something because the old name is misleading is an accomplishment in and of itself

Removing the old name is an accomplishment, but merely adding a new name gives people more things to look at. It will improve the readability of code that uses the new name, but its affect on the task of writing new code is not always beneficial.

CelticMinstrel: That's a good point. Perhaps I could add a `__dir_exclude` metafunction/table, which lists keys to exclude from the dir listing even if they exist in the table.

CelticMinstrel: I implemented something a bit different in the "Exclude deprecated things from dir()" commit (not direct-linking it because the hash could evolve as I force-push). Does that look sufficient to you? There could in theory be a few deprecated things that are caught by that system but they should be rare.

The implementation is likely inefficient, but as this is intended for use in the Lua console to discover API (particularly experimental API), I don't think that matters much.

CelticMinstrel:
Note: I merged part of new_rendering in so I could actually test stuff. That'll be merged to master before this is done and I'll force-push it into oblivion at that time, so just ignore those extra commits.

src/scripting/lua_color.cpp
```diff
@@ -126,11 +126,28 @@ static int impl_color_get(lua_State *L)
...
+	lua_push(L, keys);
+	return 1;
+}
+
 static int impl_color_set(lua_State *L)
 {
 	return luaL_argerror(L, 2, "color objects canot be modified");
 }
 
+static int impl_colors_dir(lua_State* L)
```
stevecotton, line 141: I'd prefer to have more than a one-character difference from `impl_color_dir`. Even if it's trivial in this case, it's something that I'd like to be a general coding guideline.

Maybe `impl_colors_table_dir`?

src/scripting/lua_kernel_base.cpp
```diff
@@ -501,6 +511,179 @@ static int intf_get_language(lua_State* L)
...
+			break;
+		case LUA_TTABLE:
+			auto dir_keys = lua_check<std::vector<std::string>>(L, -1);
+			std::copy(dir_keys.begin(), dir_keys.end(), std::back_inserter(keys));
+			break;
+	}
+	lua_pop(L, 1);
+}
+
+// These two are separate functions so I can use a protected call on it to catch errors.
```
stevecotton, line 531: It's not obvious what the "other one" of "these two" is. Please name it explicitly.

src/scripting/lua_kernel_base.cpp
```diff
@@ -501,6 +511,179 @@ static int intf_get_language(lua_State* L)
...
+			lua_pushcfunction(L, impl_get_dir_suffix);
+			lua_pushvalue(L, 1);
+			const auto& key = keys[j + i * n_cols];
+			lua_pushlstring(L, key.c_str(), key.size());
+			std::string suffix = " !"; // Exclamation mark to indicate an error
+			if(lua_pcall(L, 2, 1, 0) == LUA_OK) {
+				suffix = luaL_checkstring(L, -1);
+			}
+			// This weird calculation is because width counts in bytes, not code points
+			// Since the suffix is a Unicode character, that messes up the alignment
```
stevecotton, line 675: Can we be sure that there are no multibyte characters in the Lua names?

CelticMinstrel: That's a good point, as it's technically a possibility… however, Lua doesn't allow Unicode identifiers, and in 99% of cases these will be identifiers that are being listed. I'm not really sure if it's worth making the same correction for the actual identifiers…

src/scripting/lua_kernel_base.cpp
```diff
@@ -501,6 +511,179 @@ static int intf_get_language(lua_State* L)
...
+ * Prints out a list of keys available in an object.
+ * A list of keys is gathered from the following sources:
+ * - For a table, all keys defined in the table
+ * - Any keys accessible through the metatable chain (if __index on the metatable is a table)
+ * - The output of the __dir metafunction
+ * - Filtering out any keys beginning with two underscores
+ * - Filtering out any keys for which object[key].__deprecated exists and is true
+ * The list is then sorted alphabetically and formatted into columns.
+ * - Arg 1: Any object
+ * - Arg 3: (optional) Function to use for output; defaults to _G.print
```
stevecotton, line 584: Missing the documentation for arg 2.

src/scripting/lua_kernel_base.cpp
```diff
@@ -501,6 +511,179 @@ static int intf_get_language(lua_State* L)
...
+ */
+static int intf_object_dir(lua_State* L)
+{
+	if(lua_isnil(L, 1)) return luaL_argerror(L, 1, "Can't dir() nil");
+	if(!lua_isfunction(L, -1)) {
```
stevecotton, line 589: Why does this use `-1` instead of just using `3`?

stevecotton:
A name as generic as `dir` shouldn't be added to the global namespace.

src/scripting/lua_kernel_base.cpp
stevecotton: ok

src/scripting/lua_kernel_base.cpp
```diff
@@ -501,6 +511,179 @@ static int intf_get_language(lua_State* L)
...
+			lua_pushboolean(L, deprecated);
+			return 1;
+		}
+		lua_pop(L, 1);
+	}
+	lua_pushboolean(L, false);
+	return 1;
+}
+
+static int impl_get_dir_suffix(lua_State*L)
```
stevecotton, line 549: Are these suffixes a convention, either from Lua or another language?

If they're not a convention: for functions, the suffix ` ()` would seem more natural to me than ` ƒ`, although both make sense. I'm not seeing the logic that associates daggers with tables.

General comment on the whole PR: more code-comments would help. Knowing what each function is meant to do makes reviewing both easier and better (because the reviewers don't make as many incorrect assumptions).

CelticMinstrel: I chose dagger for table because it resembles a T, and florin for functions because it resembles an F. An italicized lower-case F (which the florin basically is) _is_ commonly used for functions in mathematics, which influenced my decision.

src/scripting/lua_widget_attributes.cpp
```diff
@@ -574,4 +574,47 @@ int impl_widget_set(lua_State* L)
 	ERR_LUA << "invalid modifiable property of '" <<  typeid(w).name()<< "' widget:" << str << "\n";
 	return luaL_argerror(L, 2, "invalid modifiable property of widget");
 }
+
+int impl_widget_dir(lua_State* L)
```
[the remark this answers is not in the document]
CelticMinstrel: `gui.show_dialog` works from the Lua console, so I tested it via `gui.show_dialog(dialog_wml, function(window) dir(whatever) end)`.

For my test case I loaded the tutorial character selection dialog into `dialog_wml`, and I also added an ID to every widget in the tutorial to prove that that aspect of it works correctly.

data/lua/core/wml.lua
```diff
@@ -345,6 +366,7 @@ if wesnoth.kernel_type() ~= "Application Lua Kernel" then
 
 	local root_variable_mt = {
 		__metatable = "WML variables proxy",
+		__dir = dir_vars(true),
```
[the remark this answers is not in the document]
CelticMinstrel: I think it would be the other way around – the tab completion could be made to use the dir mechanism instead of what it currently does. That doesn't mean calling the dir() function but rather using the underlying implementation that dir() uses. It also means tab-completion will not reveal deprecated functions, but that might be fine?

CelticMinstrel:
> A name as generic as `dir` shouldn't be added to the global namespace.

I don't really agree; I feel that the global namespace is precisely where this belongs. It doesn't really fit in the wesnoth namespace or any of our other namespaces. That said, if the namespacing is the only blocker, I'm open to moving it.

stevecotton:
> It doesn't really fit in the wesnoth namespace or any of our other namespaces.

Given that it's easy for anyone to alias `dir = wesnoth.debugging_utils_or_something.dir`, please move it out of global anyway. Having thought it over, I also suggest giving it a name that can be searched for - try typing `dir` into the forum and wiki's search boxes, and there will be lots of results for other things.

CelticMinstrel:
Um, if you're going to suggest a rename, please give _specific_ suggestions rather than "maybe this isn't really a good name". Otherwise I'm going to stick with the current name.

data/lua/core/wml.lua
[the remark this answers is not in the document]
stevecotton: Have you already looked at the existing mechanism, and if so please would you give an overview of what existing functions are gaining parallel implementations in this PR?

(I'll have a look for myself, but the review can be quicker if I you give me a quick-start on that.)

CelticMinstrel:
I'm still waiting to see if anyone has any suggestions for a different qualified name for this function. If no-one makes any specific suggestions, I'll just merge at some point with the current name (or maybe moved into wesnoth namespace since people seem to feel strongly about that).

src/scripting/lua_kernel_base.cpp
CelticMinstrel: Just for the record, I don't actually have any special attachment to these suffixes. If someone has better suggestions, I'm open to changing them.

src/scripting/lua_kernel_base.cpp
```diff
@@ -501,6 +511,180 @@ static int intf_get_language(lua_State* L)
...
+			dir_meta_helper(L, keys);
+		}
+	} else if(lua_isuserdata(L, 1) && !lua_islightuserdata(L, 1)) {
+		lua_pushvalue(L, 1);
+		dir_meta_helper(L, keys);
+		lua_pop(L, 1);
+	}
+	// Sort and remove any duplicates
+	std::sort(keys.begin(), keys.end());
+	auto final = std::unique(keys.begin(), keys.end());
```
stevecotton, line 628: While `final` isn't a reserved keyword in this context, I'd prefer to rename it to something that isn't a reserved word in any context. Maybe "removed"?

stevecotton:
> I'm still waiting to see if anyone has any suggestions for a different qualified name for this function.

* `list_attributes` would work well on the Wiki, but still gets many unrelated matches on the forums.
* `print_attributes` seems good in both.

Given that you can alias `dir = wesnoth.print_attributes`, I see no reason to put it in the global namespace.

CelticMinstrel:
> Given that you can alias `dir = wesnoth.print_attributes`, I see no reason to put it in the global namespace.

There is one reason:

> Note that the name is the same as the WFL function that does the same thing, so I think that's one argument for keeping the name.

Would you object to adding it as `wesnoth.print_attributes` and also adding a line `dir = wesnoth.print_attributes` in core.lua? Or if not in core.lua, then automatically aliasing it whenever you open the Lua console?

CelticMinstrel:
I believe this is ready for merge now, unless you wanted me to wire it into the tab-completion before merging.

stevecotton:
> > Note that the name is the same as the WFL function that does the same thing, so I think that's one argument for keeping the name.
>
> Would you object to adding it as `wesnoth.print_attributes` and also adding a line `dir = wesnoth.print_attributes` in core.lua? Or if not in core.lua, then automatically aliasing it whenever you open the Lua console?

Look at how much unrelated stuff turns up when searching the Wiki for `dir` though.

I would object, because it's easy to add but hard to remove - removing it would take a deprecation cycle.

> Also, to clarify the tab-completion situation… are you saying you intend to work on wiring this into it once I merge this, or did you want me to do that?

I meant that I intended to review, but wanted your knowledge as a head-start.

CelticMinstrel:
> removing it would take a deprecation cycle.

Not if it's _only_ in the Lua console. That's not API that can be used in Lua loaded into the scenario. You claim that aliasing it manually is "easy", and that's technically true, but it's still annoying to have to do it every single time you open a game. And the name "dir" is already used in WFL for a function that does the exact same thing. Sure it's not very searchable, but we can probably add something to the startup log in the Lua console if that's the only problem.

> I meant that I intended to review, but wanted your knowledge as a head-start.

So you _are_ planning to work on that…?

PR #5819, opened 2021-05-29
add unslowable status to spectral or fire elemental units

newfrenchy83:
the units make of fire, or ghost don't must be slowable by attacks like bolas or entangle, because same if these attacks damage these units,
their non-corporal nature make impossible to slow their movements.

### DIFF
--- a/data/core/macros/traits.cfg
+++ b/data/core/macros/traits.cfg
@@ -73,6 +73,36 @@ Undead units generally have undead as their only trait. Since undead units are t
     [/trait]
 #enddef
 
+#define TRAIT_ETHERAL_UNDEAD
+    # Units with trait Undead cannot be drained, poisoned, or plagued.
+    [trait]
+        id=etheral_undead
+        availability="musthave"
+        male_name= _ "undead"
+        female_name= _ "female^undead"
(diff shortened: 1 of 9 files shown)


Pentarctagon:
The WML indentation check failed on the CI.

Otherwise this sounds potentially undesirable because of the effect it would have on mainline balance, even if it technically is more realistic.

doofus-01:
This is also eroding the KISS principle.  I'm not saying it breaks anything, but it does add complication.

nemaara:
The ghosts are definitely too much of a balance break to add. I'm unsure about fire elementals, but I think they might be fine. I'd like to know what Hejne thinks regarding them, but the ghosts are a no from me.

Hejnewar:
Yeah, its no form me. It's unnesessary complication for players. 

I could also argue that at least ghosts and jinn are pretty material, for example shadow is using blade damage type and that type is very much material and they look pretty material on it's portrait too, they also can't go thru walls (except in UtBS). Jinn just has a lot of material parts that could be caught in net and slowed. Fire units, someone could probably make an argument that it would take them a while to burn that net or something and their resistances make them look very corporal given that they don't resist any of physical damage types.

PR #5821, opened 2021-05-30
add items from Blarumyrran and run woptipng on the directory

doofus-01:


CelticMinstrel:
Running woptipng in the same commit makes it impossible to tell what's actually new here… maybe it would be better to run woptipng on master, commit that, then add the new icons?

doofus-01:
You can look at the rich diff.

CelticMinstrel:
That really doesn't help. I mean, sure, you can notice the ones that are brand-new and those that have a very obvious change, but clicking "rich diff" on every image is a huge pain and you'll never know whether one of those that _doesn't_ show an obvious difference is an update or a woptipng alteration.

soliton-:
Certainly anyone trying to figure out attribution from this might as well give up. 2 commits would be much better.

PR #5822, opened 2021-05-31
SoF - final scenarios

doofus-01:
Trying to clean up final scenarios.  
 
- Sometimes the orc with the tablet manages to avoid all the triggering events.  I think the updated AI goals help avoid that, but it's hard to test
- Minimal cleanup to the epilogue
- added @nemaara to the credits
- (not done yert) - the "Dragon" scenario is odd, I think replacing the bats with some sort of flying reptile would make it less odd.  New units would be campaign-specific, not part of the mainline roster.


data/campaigns/Sceptre_of_Fire/scenarios/Epilogue.cfg
```diff
@@ -227,11 +227,11 @@
         [/message]
         [message]
             speaker=Haldric II
-            message= _ "Your advice seems good to me. I will take it. As for you, and the gryphon, you are free to go. Alanin, you’ve served in the army long enough to retire, and even if you hadn’t, you’ve seen enough combat. So, go back to your village with your well-earned pension."
+            message= _ "Your advice is grating to my ears, not what I wish to hear...  And yet, I see the wisdom in it, and I agree.  Alanin, you’ve served in the army long enough to retire, and even if you hadn’t, you’ve seen enough combat. Go back to your village, and I shall see that you and your family live well and enjoy great success."
```
stevecotton, line 230: There's a double-space in here, and there's also one in the "Aye, but it's no use" line.

nemaara:
I think the wyvern idea is neat. I had the dragon in DiD have fire guardian minions to abstractly represent the dragon using fire magic in a way. It seems neat if we could have different units or abilities surrounding each mainline dragon to give them all different flavors/personalities.

data/campaigns/Sceptre_of_Fire/scenarios/Epilogue.cfg
doofus-01: Sore subject, but I'll fix it, thanks.

doofus-01:
> It seems neat if we could have different units or abilities surrounding each mainline dragon to give them all different flavors/personalities.

That's a good point.  The fire dragon as a unit needs some work, we should revise it in 1.17, and keep this in mind.

doofus-01:
The CI error about `code` not being a valid key to `[goal]` may or may not be correct, but it conflicts with the wiki, and `code` didn't cause any error when tested (hard to tell if AI did what it was supposed to, it seemed to though).

https://wiki.wesnoth.org/LuaAI#Lua_Goals_and_Targets

data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
```diff
@@ -88,10 +89,21 @@
...
                 value=1
             [/goal]
+            [goal]
+                name=lua_goal
+                engine=lua
+                code = <<
+                    local turn = wesnoth.current.turn
+                    local targets = {}
+                    targets[1] = { value = 2, type = 2, loc = { x = 29, y = 7 } }
+                    targets[2] = { value = turn * 2 , type = 2, loc = { x = 19, y = 17 } }
```
CelticMinstrel, line 103: Please use strings for the target type in both of these lines (in this case that would be `type = 'explicit'`).

doofus-01:
Well, I think I did that wrong.  Dammit git.

PR #5833, opened 2021-06-05
Schema: Add lua_goal support

CelticMinstrel:
@doofus-01 Try pulling this and see if that fixes your validation issue. Pushing it to your PR would be fine if you're unable to validate locally.

### DIFF
--- a/data/schema/ai/goal.cfg
+++ b/data/schema/ai/goal.cfg
@@ -27,6 +27,10 @@
 					{FILTER_TAG criteria location ()}
 					{SIMPLE_KEY protect_radius int}
 				[/case]
+                [case]
+                    value=lua_goal
+                    {SIMPLE_KEY code string}
+                [/case]
 			[/switch]
 		[/else]
 	[/if]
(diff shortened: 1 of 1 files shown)


doofus-01:
I'm getting the same validation error.

CelticMinstrel:
Ahh, I see the issue. This wasn't the correct fix in the first place.

CelticMinstrel:
@doofus-01 Try that change instead.

doofus-01:
How do I test that locally, is it `perl wml.pm something-something`?

CelticMinstrel:
If you're asking how to run the validation locally, it would be something like `./wesnoth --data-dir=. --validate=data/_main.cfg --preprocess-defines=$DEFINE,$DIFFICULTY` where `$DEFINE` and `$DIFFICULTY` are replaced with the campaign and difficult defines respectively. (That line comes from utils/CI/schema_validation.sh.)

doofus-01:
Thanks, that works.  `validation succeeded`

PR #5834, opened 2021-06-05
Add a unit-test for the backstab ability

stevecotton:


CelticMinstrel (approved):
I don't see any issues with this.

The terms "ally" and "enemy" in the test names are relative to the unit being attacked, right?

stevecotton:
Thanks, and yes on the "ally" and "enemy" being relative to Bob.

PR #5837, opened 2021-06-07
Add wesnoth.interface.add_floating_label as a replacement for wesnoth.print

CelticMinstrel:
This returns a label handle which allows you to remove, reposition, or replace the label later.

In addition to all the features of wesnoth.print, you can now specify where the label appears onscreen,
as well as a fadeout time separate from the duration.

### DIFF
--- a/data/campaigns/tutorial/utils/utils.cfg
+++ b/data/campaigns/tutorial/utils/utils.cfg
@@ -5,7 +5,7 @@
         text={STRING}
         size=18
         duration=unlimited
-        red,green,blue=255,255,255
+        color=255,255,255
     [/print]
 #enddef
 
--- a/data/campaigns/tutorial/utils/utils.cfg
+++ b/data/campaigns/tutorial/utils/utils.cfg
@@ -1,18 +1,18 @@
 #textdomain wesnoth-tutorial
 
 #define PRINT STRING
-    [print]
+    [floating_label]
         text={STRING}
         size=18
         duration=unlimited
         red,green,blue=255,255,255
-    [/print]
+    [/floating_label]
 #enddef
 
 #define CLEAR_PRINT
-    [print]
+    [floating_label]
         text=""	# wmllint: ignore
-    [/print]
+    [/floating_label]
 #enddef
 
 #define GENDER MALE_WML FEMALE_WML
--- a/data/campaigns/tutorial/utils/utils.cfg
+++ b/data/campaigns/tutorial/utils/utils.cfg
@@ -1,18 +1,18 @@
 #textdomain wesnoth-tutorial
 
 #define PRINT STRING
-    [floating_label]
+    [print]
         text={STRING}
         size=18
         duration=unlimited
         red,green,blue=255,255,255
-    [/floating_label]
+    [/print]
 #enddef
 
 #define CLEAR_PRINT
-    [floating_label]
+    [print]
         text=""	# wmllint: ignore
-    [/floating_label]
+    [/print]
 #enddef
 
 #define GENDER MALE_WML FEMALE_WML
--- a/data/lua/core/interface.lua
+++ b/data/lua/core/interface.lua
@@ -61,7 +61,7 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 	wesnoth.get_viewing_side = wesnoth.deprecate_api('wesnoth.get_viewing_side', 'wesnoth.interface.get_viewing_side', 1, nil, wesnoth.interface.get_viewing_side)
 	wesnoth.message = wesnoth.deprecate_api('wesnoth.message', 'wesnoth.interface.add_chat_message', 1, nil, wesnoth.interface.add_chat_message)
 	-- wesnoth.wml_actions.print doesn't exist yet at this point, so create a helper function instead.
-	wesnoth.print = wesnoth.deprecate_api('wesnoth.print', 'wesnoth.interface.add_floating_label', 1, nil, function(cfg)
+	wesnoth.print = wesnoth.deprecate_api('wesnoth.print', 'wesnoth.interface.add_overlay_text', 1, nil, function(cfg)
 		wesnoth.wml_actions.print(cfg)
 	end)
 	-- No deprecation for these since since they're not actually public API yet
--- a/data/lua/core/interface.lua
+++ b/data/lua/core/interface.lua
@@ -60,6 +60,10 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 	wesnoth.end_turn = wesnoth.deprecate_api('wesnoth.end_turn', 'wesnoth.interface.end_turn', 1, nil, wesnoth.interface.end_turn)
 	wesnoth.get_viewing_side = wesnoth.deprecate_api('wesnoth.get_viewing_side', 'wesnoth.interface.get_viewing_side', 1, nil, wesnoth.interface.get_viewing_side)
 	wesnoth.message = wesnoth.deprecate_api('wesnoth.message', 'wesnoth.interface.add_chat_message', 1, nil, wesnoth.interface.add_chat_message)
+	-- wesnoth.wml_actions.print doesn't exist yet at this point, so create a helper function instead.
+	wesnoth.print = wesnoth.deprecate_api('wesnoth.print', 'wesnoth.interface.add_floating_label', 1, nil, function(cfg)
+		wesnoth.wml_actions.print(cfg)
+	end)
 	-- No deprecation for these since since they're not actually public API yet
 	wesnoth.set_menu_item = wesnoth.interface.set_menu_item
 	wesnoth.clear_menu_item = wesnoth.interface.clear_menu_item
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -5,11 +5,11 @@
     map_file=test/maps/generic_unit_test.map
     id = overlay_text_demo
     turns = unlimited
-    suppress_end_turn_confirmation = yes
     
     [side]
         side=1
         controller = human
+        suppress_end_turn_confirmation = yes
         [leader]
             name = _ "Alice"
             type = Elvish Archer
@@ -19,6 +19,7 @@
     [side]
         side=2
         controller = human
+        suppress_end_turn_confirmation = yes
         [leader]
             name = _ "Bob"
             type = Orcish Grunt
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -88,7 +88,7 @@
             code=<<
                 left_label_id = wesnoth.interface.add_overlay_text(
                     {"Left Label", size = 42, color = "ff8000"},
-                    {"unlimited", 500},
+                    {duration = "unlimited", 500},
                     -100, -60
                 )
             >>
@@ -109,7 +109,7 @@
             code=<<
                 right_label_id = wesnoth.interface.add_overlay_text(
                     {"Right Label", size = 42, color = {0, 128, 255}},
-                    {"unlimited", 500},
+                    {duration = "unlimited", 500},
                     100, -60
                 )
             >>
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -10,7 +10,7 @@
         side=1
         controller = human
         [leader]
-            name = "Alice"
+            name = _ "Alice"
             type = Elvish Archer
             id=alice
         [/leader]
@@ -19,7 +19,7 @@
         side=2
         controller = human
         [leader]
-            name = "Bob"
+            name = _ "Bob"
             type = Orcish Grunt
             id=bob
         [/leader]
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -14,6 +14,9 @@
             name = _ "Alice"
             type = Elvish Archer
             id=alice
+            [modifications]
+                {TRAIT_QUICK}
+            [/modifications]
         [/leader]
     [/side]
     [side]
@@ -24,6 +27,9 @@
             name = _ "Bob"
             type = Orcish Grunt
             id=bob
+            [modifications]
+                {TRAIT_QUICK}
+            [/modifications]
         [/leader]
     [/side]
     [event]
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -38,8 +38,14 @@
             code=<<
                 left_label_id = {valid = false}
                 right_label_id = {valid = false}
-                top_label_id = left_label_id
-                bottom_label_id = left_label_id
+                top_label_id = {valid = false}
+                bottom_label_id = {valid = false}
+                wesnoth.interface.add_overlay_text("A test scenario demonstrating the add_overlay_text API. Move one of the units to the labelled hexes to add, move, or remove labels at the edge of the screen.", {
+                    size = 22, color = "ff8000",
+                    duration = "unlimited",
+                    max_width = 300,
+                    valign = "top", halign = "right"
+                })
             >>
         [/lua]
     [/event]
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -88,7 +88,7 @@
             code=<<
                 left_label_id = wesnoth.interface.add_overlay_text(
                     {"Left Label", size = 42, color = "ff8000"},
-                    "unlimited",
+                    {"unlimited", 500},
                     -100, -60
                 )
             >>
@@ -109,7 +109,7 @@
             code=<<
                 right_label_id = wesnoth.interface.add_overlay_text(
                     {"Right Label", size = 42, color = {0, 128, 255}},
-                    "unlimited",
+                    {"unlimited", 500},
                     100, -60
                 )
             >>
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -5,6 +5,7 @@
     map_file=test/maps/generic_unit_test.map
     id = overlay_text_demo
     turns = unlimited
+    suppress_end_turn_confirmation = yes
     
     [side]
         side=1
@@ -86,11 +87,12 @@
         [/filter]
         [lua]
             code=<<
-                left_label_id = wesnoth.interface.add_overlay_text(
-                    {"Left Label", size = 42, color = "ff8000"},
-                    {duration = "unlimited", 500},
-                    -100, -60
-                )
+                left_label_id = wesnoth.interface.add_overlay_text("Left Label", {
+                    size = 42, color = "ff8000",
+                    duration = "unlimited", fade_time = 500,
+                    location = {100, -60},
+                    halign = "left"
+                })
             >>
         [/lua]
     [/event]
@@ -107,11 +109,12 @@
         [/filter]
         [lua]
             code=<<
-                right_label_id = wesnoth.interface.add_overlay_text(
-                    {"Right Label", size = 42, color = {0, 128, 255}},
-                    {duration = "unlimited", 500},
-                    100, -60
-                )
+                right_label_id = wesnoth.interface.add_overlay_text("Right Label", {
+                    size = 42, color = {0, 128, 255},
+                    duration = "unlimited", fade_time = 500,
+                    location = {100, -60},
+                    halign = "right"
+                })
             >>
         [/lua]
     [/event]
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -87,7 +87,7 @@
         [lua]
             code=<<
                 left_label_id = wesnoth.interface.add_overlay_text(
-                    {"Left Label", 42, "ff8000"},
+                    {"Left Label", size = 42, color = "ff8000"},
                     "infinity",
                     -100, -60
                 )
@@ -108,7 +108,7 @@
         [lua]
             code=<<
                 right_label_id = wesnoth.interface.add_overlay_text(
-                    {"Right Label", 42, "0080ff"},
+                    {"Right Label", size = 42, color = {0, 128, 255}},
                     "infinity",
                     100, -60
                 )
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -88,7 +88,7 @@
             code=<<
                 left_label_id = wesnoth.interface.add_overlay_text(
                     {"Left Label", size = 42, color = "ff8000"},
-                    "infinity",
+                    "unlimited",
                     -100, -60
                 )
             >>
@@ -109,7 +109,7 @@
             code=<<
                 right_label_id = wesnoth.interface.add_overlay_text(
                     {"Right Label", size = 42, color = {0, 128, 255}},
-                    "infinity",
+                    "unlimited",
                     100, -60
                 )
             >>
--- a/data/test/scenarios/overlay_text_demo.cfg
+++ b/data/test/scenarios/overlay_text_demo.cfg
@@ -37,7 +37,7 @@
         [lua]
             code=<<
                 left_label_id = {valid = false}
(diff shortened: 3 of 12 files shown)


stevecotton:
> This returns a label handle which allows you to remove, reposition, or replace the label later.

Is there a demo showing this in action?

CelticMinstrel:
There's no demo; I tested by launching the tutorial and running the commands from the console. Shouldn't be too hard to make a demo, I guess?

stevecotton:
> Specifically 5781, 5837, and 5852.
> All except 5852 has already had some review.

Just because someone's asked for a demo doesn't mean that they've reviewed the code. I want unit tests and/or a demo *before* I review the code.

Although this was reviewed back in #5537, the comments raised there don't seem to have been addressed:

> There will be both `wesnoth.interface.float_label` and `wesnoth.interface.add_floating_label`, which despite the similar names are different implementations.
>
> The arguments for `add_floating_label` aren't documented in the C++, which is a break from the coding style in this file. The code also seems to be trying to accept options in multiple orders (and treating hexadecimal strings as a different thing to numbers), which is adding unnecessary complexity.

CelticMinstrel:
I didn't address the name similarity simply because I haven't been able to come up with a better name. I can add documentation comments to `add_floating_label`. (There are other functions that lack this documentation too, this should probably be addressed someday.)

Pentarctagon:
> trying to accept options in multiple orders

sounds just it would just end up confusing for UMC authors too.

CelticMinstrel:
> sounds just it would just end up confusing for UMC authors too.

That's not what it's doing. It presents a simple API  (`add_floating_label( "text", duration, location)` ) but allows the first argument to also be an options table (to control text size and color) and the same for the second argument (letting you specify the fade-out duration separately).

Regardless of the form you choose, the first argument is the text, the second argument is the duration, and the final argument is the location.

stevecotton:
This fails schema validation now that #5800 has merged.

> > trying to accept options in multiple orders
> 
> That's not what it's doing.

Passing a table as the first argument:

* `{"Label", "ff8000", 42}` is equivalent to `{"Label", 42, "ff8000", 42}`
* `{"Label", "42", "ff8000"}` gives the error "color hex string should be exactly 6 digits"
* `{"Label", 42, 0x0080ff}` crashes out "Caught general 'St12length_error' exception: Text is too long to render", because it's interpreting the color as a size

Passing a table as the second argument also has a confusing issue:

* `{duration = "infinity"}` doesn't trigger a warning, but the label fades out after 2 seconds

The demo doesn't include full coverage of the features in this PR.

CelticMinstrel:
> The demo doesn't include full coverage of the features in this PR.

It wasn't really meant to have full coverage… it does cover most of the new features though? I guess there's no label with a separate duration and fade time, but that's the only thing I can think of.

stevecotton:
Instead of having several arguments that can contain tables of optional values, why not have the first argument always be the mandatory text string, and the second argument be a table of all optional values?

About code coverage - if there's functionality that isn't needed by the demo and isn't needed by the `[print]` tag's implementation, is it necessary to add that functionality now?

CelticMinstrel:
> Instead of having several arguments that can contain tables of optional values, why not have the first argument always be the mandatory text string, and the second argument be a table of all optional values?

Well, I like how this method groups the options into related sets - the first argument describes the text, the second argument describes the duration, and the last argument describes the positioning. Having a single options table for everything would certainly be a possibility though.

> About code coverage - if there's functionality that isn't needed by the demo and isn't needed by the `[print]` tag's implementation, is it necessary to add that functionality now?

I think it's useful functionality. Even though it's not covered by the demo, I'm pretty sure I did some tests back when I implemented it, to make sure it works. I can do some more, just to make sure.

stevecotton:
> Even though it's not covered by the demo, I'm pretty sure I did some tests back when I implemented it, to make sure it works. I can do some more, just to make sure.

Thanks for the honesty, but I'd have more confidence if the demo code had full coverage.

The more functionality that you add, the harder it becomes for someone to later try and unify this with the `[floating_text]` tag and the `[label]` tag, and also the more time that UMC authors are likely to waste working out which of those tags does what they want. I'd much prefer to keep `[print]` as simple as possible; that also makes it more likely that someone in future will agree to adding a function with the name that you want, and with the full functionality that you want.

Something that I'd really like to see is a guide for UMC authors about which of `[floating_text]`, `[label]` and `[print]` they should use. What are the use-cases where the new `[print]` functionality is the best choice?

CelticMinstrel:
> Thanks for the honesty, but I'd have more confidence if the demo code had full coverage.

Well, the tutorial covers that particular functionality, and it appears to work there.

data/campaigns/tutorial/utils/utils.cfg
```diff
@@ -5,7 +5,8 @@
         text={STRING}
         size=18
         duration=10000
```
stevecotton, line 7: Whether or not this PR goes in, I feel that the tutorial should probably use the default values. The default duration is currently 50ms, which I consider to be a bug (introduced in a9d9e48c72e70fd41ae9759e3894752eb9708596).

stevecotton:
Please respond to my "... What are the use-cases where the new [print] functionality is the best choice?" question.

> Well, the tutorial covers that particular functionality, and it appears to work there.

Murphy's Law (build is master + f0cf798ec1155c2497130b95f211091fc6765d6f)
```
error scripting/lua: lua/wml-tags.lua:795: bad argument #2 to 'add_overlay_text' (that duration should be integer or 'infinity' expected, got table)
stack traceback:
	[C]: in field 'add_overlay_text'
	lua/wml-tags.lua:795: in local 'cmd'
```

CelticMinstrel:
Is that from the tutorial, or from manually calling the function (eg in the Lua console)? Because I distinctly remember a lack of error messages when I tested the tutorial.

data/campaigns/tutorial/utils/utils.cfg
CelticMinstrel: I'm not sure if it makes sense for the tutorial to use the default values. I guess if it depends if your expected common use case is the same as how the tutorial uses it?

stevecotton:
It's from the tutorial.

data/campaigns/tutorial/utils/utils.cfg
stevecotton: 50ms doesn't make sense at all.

CelticMinstrel:
Well, I'm not getting that error at all. I would expect to see that error if you tried to do something like `{duration = 'moo'}`.

stevecotton:
Please respond to my "... What are the use-cases where the new [print] functionality is the best choice?" question.

I retested and reproduced that error on 616f4657f917c3aced09f2f2321024184bca8707.

* Start the tutorial
* Choose which character to play
* Click on the character
* Error message appears (and the `[print]`'d text does not appear)

CelticMinstrel:
* A separate fade time just makes sense most of the time - gradually fading out over the full duration is quite weird.
* Setting an offset would be useful if you want to have several types of messages appear on different parts of the screen.

That's the only new functionality added to `[print]`; the Lua API supports moving an existing label and placing multiple labels onscreen at the same time, but `[print]` doesn't.

stevecotton:
> * A separate fade time just makes sense most of the time - gradually fading out over the full duration is quite weird.

Nothing requires `[print]` to have a fade-out at all, and it doesn't in the current master.

> * Setting an offset would be useful if you want to have several types of messages appear on different parts of the screen.

If they're meant to align to things on the map then something that takes a location would be better than something that takes an offset. If the intention is to place things somewhere on the user's screen then an API like IntroWML that adjusts to the screen size would be better than using pixel offsets.

In short, I don't see a use case which would justify making `[print]` more complex. It's probably time to ask @Pentarctagon to decide if this PR should go in.

CelticMinstrel:
> Nothing requires `[print]` to have a fade-out at all, and it doesn't in the current master.

False. In current master, `[print]`'s fade-out is its duration. Just try waiting for ten seconds in the tutorial.

> If they're meant to align to things on the map then something that takes a location would be better than something that takes an offset.

Yeah, I was thinking about that too… I'd like to add support for linking the location to a map location but I'm not really sure how that would work…

> If the intention is to place things somewhere on the user's screen then an API like IntroWML that adjusts to the screen size would be better than using pixel offsets.

This was more my intention. Yeah, you do have a point, maybe I should scale the offset by the resolution?

stevecotton:
> > > A separate fade time just makes sense most of the time - gradually fading out over the full duration is quite weird.
> >
> > Nothing requires `[print]` to have a fade-out at all, and it doesn't in the current master.
> 
> False. In current master, `[print]`'s fade-out is its duration. Just try waiting for ten seconds in the tutorial.

What's false? In the tutorial, the text appears for 10 seconds and then disappears. The alpha value doesn't appear to change during the 10 seconds; it certainly doesn't gradually fade out over the full duration.

> Yeah, I was thinking about that too… I'd like to add support for linking the location to a map location but I'm not really sure how that would work…
> ...
> This was more my intention. Yeah, you do have a point, maybe I should scale the offset by the resolution?

The question that I'm more interested in is whether this is for 1.16 or 1.18. My feeling is that this needs to be postponed, and it needs to start with a decision about which problems it's going to solve, followed by deciding whether `[print]` is the right tag to change.

Pentarctagon:
Ideally anything that would break backwards compatibility is done for 1.16 rather than delayed for 1.18, since as I've said I want to significantly reduce the amount of changes needed between stable branches for things like a function getting deprecated and renamed.

CelticMinstrel:
> What's false?

I might be thinking of 1.14 rather than master?

data/campaigns/tutorial/utils/utils.cfg
stevecotton: Resolved in #5868.

stevecotton:
> Ideally anything that would break backwards compatibility is done for 1.16 rather than delayed for 1.18

I think that needs this to be split into a deprecation PR, and a separate adding-new-functionality PR.

CelticMinstrel:
It doesn't split very easily though…

CelticMinstrel:
Okay, so reading through the thread here again, it seems like all the complaints fall into three categories:

1. Stuff about `[print]`
2. Stuff about the format of the API
3. Stuff about the demo scenario

Point (1) is irrelevant, as this PR is not about `[print]`. If it makes things easier, I can just delete the extra keys from `[print]` for now (we can always add them later if they're wanted). I do think that adding a fade time to `[print]` is nice to have, but it's not the point of this PR, and the new offset keys in `[print]` don't even have a current use-case. **Should I go ahead and delete those extra keys for now?**

Point 2 is just a matter of some minor reorganizing of the code if we want to change the API. I like the fact that the current API divides all the possible arguments into categories such that the first argument specifies everything about the text, the second argument specifies everything about the timing, and the third argument specifies everything about the positioning. The alternative, where the first argument is just the text string and the second argument is an options table combining text. timing, and positioning arguments, is not a huge change, although I'm not entirely convinced it's a good one. **Is my argument for the current API sound? Should I change it?**

Point 3, I'd have to take a look to see what's missing, but… certainly I could add more stuff to the demo scenario. **Let me know if there's something specific that you think is important to be there.**

data/lua/core/interface.lua
```diff
@@ -60,6 +60,10 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 	wesnoth.end_turn = wesnoth.deprecate_api('wesnoth.end_turn', 'wesnoth.interface.end_turn', 1, nil, wesnoth.interface.end_turn)
 	wesnoth.get_viewing_side = wesnoth.deprecate_api('wesnoth.get_viewing_side', 'wesnoth.interface.get_viewing_side', 1, nil, wesnoth.interface.get_viewing_side)
 	wesnoth.message = wesnoth.deprecate_api('wesnoth.message', 'wesnoth.interface.add_chat_message', 1, nil, wesnoth.interface.add_chat_message)
+	-- wesnoth.wml_actions.print doesn't exist yet at this point, so create a helper function instead.
+	wesnoth.print = wesnoth.deprecate_api('wesnoth.print', 'wesnoth.interface.add_overlay_text', 1, nil, function(cfg)
```
stevecotton, line 64: IIRC, the plan was for no new Lua deprecations in 1.18.

stevecotton:
> Okay, so reading through the thread here again, ...

What are the current goals of the PR? Please give some use cases from the perspective of players or UMC authors, so it's possible to check whether the PR would meet those goals.

Something that I'd really like to see is a guide for UMC authors, recommending which of the multiple ways to display to text on screen they should use. What are the use-cases where the new functionality is the best choice?

> 3. Stuff about the demo scenario

I think this should come from the use-cases, and ideally be in scenarios that players see rather than the dev-only test scenario. It should be visible enough so that, if it gets broken before RC1, someone will notice it before the 1.18.0 release.

data/lua/core/interface.lua
Pentarctagon: I'd hesitate to call it a deprecation if it's just a helper wrapper for something that doesn't exist right now anyway (assuming I'm understanding the comment right.

CelticMinstrel: The comment means that this code runs before the `print` function is defined.

We can certainly void the deprecation and hold off to 1.20 to deprecate it, but that creates the weird situation of having two functions in two different places that do essentially the same thing, so I don't think that's a great idea.

CelticMinstrel:
> What are the current goals of the PR?

This PR is part of the Lua reorganization project, so its _primary_ goal is converting old, bad APIs into a nicer form.

The extra API features mainly represent parts of the underlying C++ API that were previously not exposed to Lua but seem useful. The only exception to that is the fade out, which previously didn't work in a sane way (if at all, I don't quite remember).

* Offset - if you know you want to have two messages on the screen simultaneously, the offset lets you do that without causing problems.
* Fade - fading out at the end just makes it look nicer, in my opinion.
* Moving a label - I don't have a specific use-case for this one, and I _think_ you could get the same effect by deleting the label and placing an identical one in the new location, but it seems like a useful convenience if all you want to do is reposition it.

I think that covers all the new features added here.

Pentarctagon:
IMO the features sound fine and the deprecation is... fine, I guess, since it's just a single one. Just keep in mind that with 1.16 I did announce that there wouldn't be any more mass deprecations/renamings done.

CelticMinstrel:
I think I had about four deprecations in mind for 1.18, so that probably doesn't count as "mass deprecations"? We can talk about each one as it comes up though.

data/test/scenarios/overlay_text_demo.cfg
```diff
@@ -0,0 +1,162 @@
...
+        name=moveto
+        first_time_only=no
+        [filter]
+            x,y=15,6
+        [/filter]
+        [lua]
+            code=<<
+                right_label_id = wesnoth.interface.add_overlay_text(
+                    {"Right Label", size = 42, color = {0, 128, 255}},
+                    {"unlimited", 500},
```
stevecotton, line 112: This line doesn't set a duration, although either of the following would work here:
* `"unlimited",`
* `{duration = "unlimited", 500},`

stevecotton:
> Offset - if you know you want to have two messages on the screen simultaneously, the offset lets you do that without causing problems.

I think there's no good UX which would be served by that.

The tutorial needs to get the user to do specific actions, and it puts a message in the middle of the screen. It's a bad UX, but we accept it because at that point the user is on rails while learning the interface - if they don't do that exact action then we put up a message about undoing the action, and lock out the "end turn" button.

To put a message near the center of the screen, I think it's a terrible UX unless it's either a short-duration message, or is as important to the user as the tutorial's "do exactly this or get softlocked" messages. Having two such messages on screen simultaneously suggests two independent "do exactly this as your next action or get softlocked" messages - if they're dependent on each other then a single message such as "move both Kalenz and Landar near the signpost" would cover both situations.

The API isn't right for putting things near the edge of the screen, because that would require checking the window size, calculating the offset from the center, and then checking that the window isn't resized afterwards. For the UtBS scenario, I think putting a persistent "Merfolk left to save: $number" near the edge of the display would be better than the current UX of using a temporary message.

The added Lua code has something that uses a `map_location loc{0, 0, wml_loc()};` but the intention is undocumented. The effect would seem to be that it would take the current screen coordinates of that hex, but then attach the label to those screen coordinates, so scrolling the screen would move the label away from its original hex. I'm assuming that this is leftover code, as it doesn't seem to fit with the offsets idea.

CelticMinstrel:
> The added Lua code has something that uses a `map_location loc{0, 0, wml_loc()};` but the intention is undocumented.

It's not a map location. It's just using the `map_location` struct because that conveniently already handles grabbing two coordinates, but the numbers read into that struct are taken as screen coordinates.

stevecotton:
> > The added Lua code has something that uses a `map_location loc{0, 0, wml_loc()};` but the intention is undocumented.
> 
> It's not a map location. It's just using the `map_location` struct because that conveniently already handles grabbing two coordinates, but the numbers read into that struct are taken as screen coordinates.

That needs some comments to make the code more understandable.

CelticMinstrel:
Looks like the build failure is literally the one line that was mysteriously not working for me. I can delete it, but… why _does_ that not work?

data/test/scenarios/overlay_text_demo.cfg
```diff
@@ -0,0 +1,165 @@
+#textdomain wesnoth-test
+
+[test]
+    name = _ "Overlay Text Demo"
+    map_file=test/maps/generic_unit_test.map
+    id = overlay_text_demo
+    turns = unlimited
+    suppress_end_turn_confirmation = yes
```
stevecotton, line 8: This is a per-`[side]` setting.

CelticMinstrel:
Okay… so does the changed API and the added `[hv]align` keys address most of your concerns? I'll add "top" and "bottom" labels to the test scenario, but is there anything else I should add as well?

CelticMinstrel:
Noticed a bug – labels don't reposition correctly if you switch resolution. I feel like this should be a solved problem in the engine though… or do other built-in labels do the same thing?

src/scripting/game_lua_kernel.cpp
```diff
@@ -2281,47 +2281,177 @@ int game_lua_kernel::intf_find_cost_map(lua_State *L)
...
+ * Returns: label handle
+ */
+int game_lua_kernel::intf_set_floating_label(lua_State* L, bool spawn)
+{
+	t_string text = luaW_checktstring(L, 1);
+	int size = font::SIZE_SMALL;
 	color_t color = font::LABEL_COLOR;
+	int lifetime = 2'000, fadeout = 100;
+	font::ALIGN alignment = font::ALIGN::CENTER_ALIGN, vertical_alignment = font::ALIGN::CENTER_ALIGN;
+	// This is actually a relative screen location, but map_location already supports
```
stevecotton, line 2338: Please say somewhere in this comment that it's measured in pixels, not hexes.

stevecotton:
> Noticed a bug – labels don't reposition correctly if you switch resolution. I feel like this should be a solved problem in the engine though… or do other built-in labels do the same thing?

I feel investigating that, or working out how to connect the implementation of this to the window-resize, is part of this PR. Should I review before that's implemented?

CelticMinstrel:
I'd like a review of the API at least. If you don't want to comb through the source code when you're not sure what might change for the bugfix, that seems fine to me.

CelticMinstrel:
Okay, so I tested in 1.14 and 1.16, and the bug about labels not adjusting when you switch resolutions is already present, so I think that bug shouldn't be a blocker for merging this PR and should instead be opened as a separate issue.

data/test/scenarios/overlay_text_demo.cfg
```diff
@@ -0,0 +1,270 @@
...
+                {TRAIT_QUICK}
+            [/modifications]
+        [/leader]
+    [/side]
+    [event]
+        name=preload
+        [lua]
+            code=<<
+                left_label_id = {valid = false}
+                right_label_id = left_label_id
```
stevecotton, line 40: While copying the values like this works, it would be more readable to just put `{valid = false}` on each line.

data/test/scenarios/overlay_text_demo.cfg
```diff
@@ -0,0 +1,270 @@
...
+                top_label_id = left_label_id
+                bottom_label_id = left_label_id
```
stevecotton, line 42: Wondering about the WML API that would be built over this, and how it would fit with multiple labels. I think the current `[print]` behavior of always replacing a previous label should be applied in a per-location fashion, so `[add_overlay_text]valign=bottom` would replace the previous `[add_overlay_text]valign=bottom` etc.

Or, if there isn't a WML wrapper, will it always need a prestart event like this one?
```
#define INIT_HINT_STRING_SUPPORT
    # Should be used in [event]name=prestart, or another event that triggers before the first hint.
    [lua]
        code=<<
            hint_label = {valid = false}
        >>
    [/lua]
#enddef
```

Context: I've been wondering about this as a solution to the unread walls of text in the tutorial.
```
#define SHOW_HINT_STRING TEXT
    # Similar to PRINT_STRING, but puts the text at the edge (top or bottom) of the screen.
    # Used in the tutorial for info that can be left for reading until another hint is triggered.
    #
    # There is no interaction between PRINT_STRING and SHOW_HINT_STRING - they don't cancel each
    # other's messages.
    [lua]
        code=<<
            hint_label = wesnoth.interface.add_overlay_text((...).text, {
                size = 24,
                color = {255, 255, 255},
                duration = "unlimited",
                valign = "bottom"
            })
        >>
        [args]
            text={TEXT}
        [/args]
    [/lua]
#enddef
```

src/scripting/game_lua_kernel.cpp
```diff
@@ -2281,47 +2281,177 @@ int game_lua_kernel::intf_find_cost_map(lua_State *L)
...
+		if(luaW_tableget(L, 2, "size")) {
+			size = luaL_checkinteger(L, -1);
+		}
+		if(luaW_tableget(L, 2, "color")) {
+			if(lua_isstring(L, -1)) {
+				color = color_t::from_hex_string(lua_tostring(L, -1));
+			} else {
+				auto vec = lua_check<std::vector<int>>(L, -1);
+				if(vec.size() != 3) {
+					return luaL_error(L, "floating label text color should be a hex string or an array of 3 integers");
```
stevecotton, line 2352: Alpha channel?

src/scripting/game_lua_kernel.cpp
```diff
@@ -2281,47 +2281,177 @@ int game_lua_kernel::intf_find_cost_map(lua_State *L)
...
+	// everything needed to read in a pair of coordinates.
+	// Depending on the chosen alignment, it may be relative to centre, an edge centre, or a corner.
+	map_location loc{0, 0, wml_loc()};
+	if(lua_istable(L, 2)) {
+		if(luaW_tableget(L, 2, "size")) {
+			size = luaL_checkinteger(L, -1);
+		}
+		if(luaW_tableget(L, 2, "color")) {
+			if(lua_isstring(L, -1)) {
+				color = color_t::from_hex_string(lua_tostring(L, -1));
```
stevecotton, line 2348: I'd prefer having only one way to provide arguments, so only accepting an RGB (or RGBA) vector for the color value.

data/test/scenarios/overlay_text_demo.cfg
CelticMinstrel: There is no WML API _yet_ but I assume there would be at some point.

src/scripting/game_lua_kernel.cpp
CelticMinstrel: Converting between the two formats is non-trivial, so I think it's totally worth it to support both.

src/scripting/game_lua_kernel.cpp
CelticMinstrel: That is a good question… I'm not even sure if alpha channel would work, but I also can't think of a reason why it wouldn't…

stevecotton:
Text at the bottom of the screen is going to overlap with `[message]` boxes, so maybe that area isn't useful. Text at the top seems promising, if it's large font and isn't in a multiplayer scenario (so there's no chat log).

I tried using this as an improvement for the tutorial's walls-of-text, by converting some of the narrator messages to overlays. I doubt that's what you expected, but it's one of the UXs that I think this could be useful for. It would either need much less text to support a bigger font, or it would need a background similar to the one for `[message]`.

tutorial_add_overlay

Code is in the https://github.com/stevecotton/wesnoth/commits/tutorial_overlay_text branch

CelticMinstrel:
> It would either need much less text to support a bigger font, or it would need a background similar to the one for `[message]`.

I think the floating label system does support a background, so that could be added to the Lua API either in this PR or in a later PR.

PR #5838, opened 2021-06-07
wesnoth.unit_types and wesnoth.terrain_types are now available in map generation scripts

CelticMinstrel:
For example, this could allow you to tailor a map to the movement of a specific unit type, or modify terrain based on properties rather than just the terrain code.


CelticMinstrel:
> What uses the newly-available functions during map generation, or in other words what tests this new functionality?

Nothing yet. I can definitely think of use-cases, but none have been written yet.

It's fine if this doesn't go into 1.16. Perhaps when I finally get back to #993 some use-cases for this can be included.

stevecotton (changes requested):
> > What uses the newly-available functions during map generation, or in other words what tests this new functionality?
> 
> Nothing yet. I can definitely think of use-cases, but none have been written yet.

Then I think it's not ready for merging, and most definitely not 10 days before API freeze.

Pentarctagon:
Any update on this?

CelticMinstrel:
It's on the roadmap for… I think the release after next? Or more generally, once the events PR is finished.

stevecotton:
The conflict suggests that #8801 should be applied to whereever that code moved to.

PR #5839, opened 2021-06-07
Add addons to wesnoth.game_config

CelticMinstrel:
This adds:
- wesnoth.game_config.eras
- wesnoth.game_config.campaigns
- wesnoth.game_config.modifications
- wesnoth.game_config.resources

Each is an ID to config mapping of the particular addon type.

wesnoth.game_config.eras previously existed as just an array of all era IDs, so this is an incompatible change. However, the old API was never documented, and the comment said it was specifically for @sigurdfdragon who approved this change (although he couldn't remember why he used it).

I'm a little unsure if there are use-cases for this. It most likely won't show all installed add-ons, since @gfgtdf made changes awhile back to only load active add-ons. At most, it would show all eras, campaigns, modifications, and resources associated with an active add-on, regardless of whether said era, modification, or resource is actually active, plus all built-in core eras, modifications, and resources. The same is probably true of campaigns, but they normally don't load at all in multiplayer.


Pentarctagon:
> I'm a little unsure if there are use-cases for this.

Does this even need to be added in the first place, then?

CelticMinstrel:
Well, @sigurdfdragon _used_ to have a use-case for it but seems to have forgotten what it was. So… yeah, I'm not sure.

Pentarctagon:
Given that, it sounds like it would be better to remove the old `wesnoth.game_config.eras` API, if anything.

CelticMinstrel:
Yeah, we should probably remove the old undocumented API regardless of whether this new version is added.

CelticMinstrel:
I just realized, this should also deprecate `wesnoth.get_era` (which _is_ [documented](https://wiki.wesnoth.org/LuaWML/Misc#wesnoth.get_era)) and `wesnoth.get_resource` (which isn't).

stevecotton:
> Given that, it sounds like it would be better to remove the old `wesnoth.game_config.eras` API, if anything.

That removal has merged in 2f4868351b61106bdb9a162fc2b1154b3b725977. I think this PR should be closed without merging further changes.

CelticMinstrel:
`wesnoth.get_era` and `wesnoth.get_resource` still exist and probably have the same problems as `wesnoth.game_config.eras`.

Pentarctagon:
`wesnoth.get_resource` isn't documented either, I don't think, so it could also be removed.  `wesnoth.get_era` though should probably be deprecated first, with no replacement.

sigurdfdragon:
If/when this is merged, will there still be a way for a add-on to get the user selected era? 
And if the user selected era=default_era, then fetch the age_of_heroes era by referencing the id?

Pentarctagon:
What do you mean by "fetch the age_of_heroes era"?

CelticMinstrel:
@sigurdfdragon The _active_ era is currently available via `wesnoth.scenario.era` (only in a multiplayer game). Active modifications and resources are similarly available. This is the era config, not just the ID.

> `wesnoth.get_resource` isn't documented either, I don't think, so it could also be removed. `wesnoth.get_era` though should probably be deprecated first, with no replacement.

Sounds good to me… note though that `wesnoth.get_era` is used in at least two addons. Perhaps those uses can be replaced with `wesnoth.scenario.era` though, I'm not sure.

sigurdfdragon:
> What do you mean by "fetch the age_of_heroes era"?
use a lua function to have access to all of the wml of that [era] when it is not the active era.

@CelticMinstrel 
wesnoth.scenario.era making the active era config available will work as a replacement for get_era, at least for the one add-on I'm familiar with.

Pentarctagon:
> use a lua function to have access to all of the wml of that [era] when it is not the active era.

What do you use it for?

sigurdfdragon:
In the 'Random Campaign' add-on, if the player launches it with the default era selected:
In the first 3 scenarios the player is presented with enemy factions from default era
in scenarios 4-6 they are presented with enemy factions from age of heroes.
Because the mainline eras id's are known, RC directly gets the data for those eras.

As opposed to the case of the player selecting something like imperial era, where since it is not certain what the id of the aoh style version of imperial era would be, or even that it exists. For those cases, RC derives an aoh style era from the active era by looking through the advancement tree of all the recruits in all the factions of the active era.

CelticMinstrel:
@ForestDragon-wesnoth seems to have come up with another use-case that this could support – to check the version of other add-ons that you depend on. (Although, it looks like the API doesn't currently include versions, but I'm fairly sure that can be added.)

(Whether this is the best way to check that, I'm not sure, but it's a possibility.)

Pentarctagon:
That sounds reasonable. Would it also make sense to add a specific function to get just the version information? So a `wesnoth.game_config.era_version` or similar.

CelticMinstrel:
Probably? Something like that. I'm not exactly sure what that would look like though.

Vultraz:
Any plans for this to go anywhere?

PR #5841, opened 2021-06-07
When starting or loading a game, warn if the next scenario is unknown

### LINKED ISSUE #5530: Warn on loading a game if the next_scenario doesn't exist
When the user loads a game, I suggest the engine should check that the next_scenario is a known scenario id, and show a warning if it isn't. This would be a part of #2710, but could also be useful for changes when upgrading UMC campaigns.

### Example with a non-branching path

Warning: the next scenario is missing

Although you can play this scenario, it seems that there will be a "unknown scenario" error instead of the next scenario. This is likely to be a bug in the campaign, or the result of loading an old save game that's incompatible with the current version of the campaign.

The missing scenario's id is `07b_caves`.

### Example with a branching path

Warning: depending on the branch you take, the next scenario may be missing

Although you can play this scenario, it seems that some of the ways to complete it will lead to an "unknown scenario" error instead of the next scenario. This is likely to be a bug in the campaign, or the result of loading an old save game that's incompatible with the current version of the campaign.

The following branches are expected to be ok:
* winning the entire campaign
* `19a_swamp_route`
* `19b_mountain_route`

The following branches will lead to an error:
* `19c_river_route`

stevecotton:
Check that the scenario's next_scenario= exists, and display a warning
if it would lead to an "Unknown Scenario" error. This also checks any
[endlevel] tags in events.

The error message includes the id of the missing scenario. If there
are multiple branches and some should be okay then a slightly
different message is shown saying which ones should work, hopefully
the ids will be something that hints to the player how to reach the
working branches.

The warnings' translatable strings could have plural and singular
forms, but these errors should be rare enough that it seems excessive
to have those separate plural and singular forms.

Fixes #5530.

wesnoth_unknown_next_scenario
wesnoth_unknown_and_known_next_scenarios_800x600

### DIFF
--- a/src/play_controller.cpp
+++ b/src/play_controller.cpp
@@ -36,6 +36,7 @@
 #include "game_state.hpp"
 #include "gettext.hpp"
 #include "gui/dialogs/loading_screen.hpp"
+#include "gui/dialogs/message.hpp"      // for show_error_message
 #include "gui/dialogs/transient_message.hpp"
 #include "hotkey/command_executor.hpp"
 #include "hotkey/hotkey_handler.hpp"
@@ -64,6 +65,7 @@
 #include "units/id.hpp"
 #include "units/types.hpp"
 #include "units/unit.hpp"
+#include "utils/general.hpp"
 #include "whiteboard/manager.hpp"
 
 #include <functional>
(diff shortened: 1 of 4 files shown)


src/play_controller.cpp
```diff
@@ -1196,6 +1198,85 @@ void play_controller::start_game()
...
+		message = _("There will be an “unknown scenario” error instead of the next scenario. This is likely to be a bug in the campaign, or the result of loading an old saved game that’s incompatible with the current version of the campaign.\n\nThese scenarios are missing:\n\n$unknown_list|");
+	} else {
+		// TRANSLATORS: This is an error that will hopefully only be seen by UMC authors and by players who have already said "okay" to a "loading saves from an old version might not work" dialog. It should be rare enough that there aren't separate strings for singular vs plural in either of the lists.
+		message = _("Some of the ways to win this scenario will lead to an “unknown scenario” error instead of the next scenario. This is likely to be a bug in the campaign, or the result of loading an old saved game that’s incompatible with the current version of the campaign.\n\nThe following branches are expected to be ok:\n\n$known_list|\nThe following branches will lead to an error:\n\n$unknown_list|");
```
stevecotton, line 1260: I still need a better way to give these long lines to `gettext`.

stevecotton: Ah, gettext supports string concatenation. Easier than I thought, not sure why I didn't try that before pushing the original version.

CelticMinstrel: By string concatenation do you mean `_("some text\n"  "more text\n)`?

stevecotton: Yes.

stevecotton:
Added an assumption that any value with a `$` in it is a WML variable which will be successfully interpolated later, and a unit test to check various scenarios that shouldn't cause false-positives.

Reworded the dialog itself, but would be grateful for better phrasing. I don't expect the current wording to be the final draft.

It would be easy to add more versions depending on whether the user had loaded a mid-scenario save, as the calling function knows that; but I don't think the extra clarity justifies more translatable text. Then again, I could split it up by paragraph and make each paragraph support singular/plural.
wesnoth_unknown_next_scenario
wesnoth_unknown_and_known_next_scenarios_800x600

stevecotton:
I had another go at the wording, including splitting up singulars from plurals.

The warning about replays and mid-scenario saves is only shown when loading a mid-scenario save. I'm hoping to merge this as-is, and fix it to show for replays too during the string freeze.

There's rough edges around the note that this could be the last scenario. When there are other working branches, I think it's clearer to put in it the bullet-pointed list of possible good choices, but I'm not sure whether to count it as a branch both for the heading of that list and for the choice of "Warning: old save or unwinnable scenario" vs "Warning: broken campaign branches".

unknown_scenario_1_0
unknown_scenario_1_0_last
unknown_scenario_2_1_last_reload

Pentarctagon:
IMO they're too wordy; it should just say something like:
> One or more of the possible next scenarios are missing and you may not be able to finish this campaign. Please report the following missing scenarios to the campaign's author:
> * scen_one
> * scen_two
> 
> Once this is fixed, you will need to restart this scenario.

stevecotton:
I'm okay about changing to the short wording, but one of the motivations both for this dialog's existence and its message were that a lot of mainline campaigns are going to break when people migrate from 1.14 to 1.16. The response to those being reported will be that it's not a bug.

Pentarctagon (approved):

doofus-01:
Is there a way for UMC authors to disable this?  I've got generated scenarios, not all possible values exist but it does not mean the player will encounter a broken branch.  I don't want to have to rewrite things just to avoid this message box.

I asked this on the forums as well, but here is probably better.

CelticMinstrel:
Don't generated scenarios still need to have a recognizable ID? I'm confused at how your described situation even makes sense.

doofus-01:
The situation makes sense.  All the generated scenarios have recognizable IDs.  The macro calling for the next scenario can put together IDs that don't exist, but under conditions that don't matter.

CelticMinstrel:
I see.

Well, the _simple_ fix would be to just create an empty dummy scenario for every possible non-existent ID, but I assume that's too much work.

Have you verified that your situation _does_ trigger the warning? If so, maybe there's some way for the warning to detect that those non-existent ones are in fact unreachable?

doofus-01:
Yeah, creating dummy scenarios will be whack-a-mole.

Yes, I came across this in testing fixes to the LuaAPI deprecation thing, I wouldn't bring it up just because I browsed GitHub.  This nanny message is to protect the players from careless UMC authors, it shouldn't require extra meta-code. 

Showing the relevant code will be difficult.  It's for the (many years old) add-on _Bad Moon Rising_ https://github.com/doofus-01/wesnoth-Bad_Moon_Rising 
I've no doubt there is a way to clean up the code to avoid this, but that's a really discouraging project.  A toggle to disable this hostile message isn't all that crazy.

CelticMinstrel:
It's difficult for me to figure out exactly where this macro stuff is in the add-on… I can see various macros containing generator code, but none of that seems particularly relevant.

soliton-:
Since this check cannot be free of false positives a way to explicitly disable it makes a lot of sense.

stevecotton:
There's already a unit-test that shows things that won't be treated as positives, even if the scenario doesn't exist.

For BMR, the error dialog can be disabled by interpolating a blank variable into the scenario id:
```
         [endlevel]
-            next_scenario={SCENARIO}
+            next_scenario={SCENARIO}"$disable_unknown_scenario_error|"
             {CONTINUE}
             replay_save=no
         [/endlevel]
```
which I tried combined with (possibly unnecessarily, given WML's handling of unset variables) setting that to a blank string in a `[event]name=prestart`.

When testing, I'd installed the 1.14 version of the add-on, so trying to trigger those `[endlevel]` tags was blocked by an error about `store_time_of_day`.

doofus-01:
> When testing, I'd installed the 1.14 version of the add-on, so trying to trigger those [endlevel] tags was blocked by an error about store_time_of_day.

Thanks for checking.  I believe that is a Wesnoth bug fixed in https://github.com/wesnoth/wesnoth/commit/b05215d8aa66b088b30f133ee341e0e76b42857e
At least, I was recently getting that error consistently too, until updating yesterday.

doofus-01:
I can confirm appending the empty variable works.  Seems a little obscure, hopefully it is a stable method.

Issue #5840, opened 2021-06-07
User actions in lobby are executed by index and can apply to the wrong user


vgaming:
### Game and System Information
 - What version of the game are you running? = 1.14.16
 - Where did you download it from? = ArchLinux
 - What OS are you running? = ArchLinux

### Describe the bug
When I try to execute an action by selecting a user in lobby, it can be applied to the wrong user if the original target disconnects from the server

### To Reproduce
1. Open Multiplayer lobby (log in)
2. Select a target user by double-clicking on somebody. A pop-up should appear
3. Without closing the popup, wait till the target user disconnects (if the target user is your second nickname, then just disconnect with the second nickname from the server)
4. With the popup window still open, execute some action like Add Friend or Send Private Message (tested on `Send Private Message`)

### Expected behavior
The action should be applied to the user I originally selected

What actually happens: the action is applied to the alphabetically next user. Note that this also applies to moderation commands, like "kick" or "ban". It is hard to undo these commands as well.

### Screenshots
![2021 06 07_09:20:19](https://user-images.githubusercontent.com/33452702/120975517-9af65980-c771-11eb-88d8-417fa4b711ff.png)

soliton-:
Can you make:
https://github.com/wesnoth/wesnoth/blob/957496560b4b9eb8fe7d5e41db334a9f3435e7ea/src/gui/dialogs/multiplayer/lobby.cpp#L1173-L1179

into:
``` c++
 delay_playerlist_update_ = true;
 lobby_delay_gamelist_update_guard g(*this); 

 lobby_player_info dlg(*chatbox_, *info, lobby_info_); 
 dlg.show(); 
```
and see if that fixes it?

vgaming:
@soliton- I've built the latest `master` to have a baseline for comparison, but on `master` the bug is not present.

Any opinions on what to do about the issue?

soliton-:
Are you sure it's not present? How did you test?

vgaming:
@soliton- I've started two `master` branch clients. I've connected both of them to the online server. With one of them, I've "selected" (popup window) the other. Then I disconnected the other. Then in the popup screen of the first client, I've made actions like "send message". A chat was opened with the "other" nickname even though that nickname has already disconnected.

Note that there was no alphabetically "next" person on the server, maybe it matters. I'll try swapping the roles of the nicknames. Maybe it also matters that the only other player is myself. I could try starting up a third client to test..

soliton-:
[...]

PR #5842, opened 2021-06-07
Add option to pre-populate wesnothd's user list with dummy users.

Pentarctagon:
None of the dummy players have an actual connection or client, so any attempt to interact with them will fail. The main usecase for this right now is testing the performance of the MP lobby UI updates when there are many players online.

### DIFF
--- a/src/server/wesnothd/server.cpp
+++ b/src/server/wesnothd/server.cpp
@@ -259,6 +259,8 @@ server::server(int port,
 	, cmd_handlers_()
 	, timer_(io_service_)
 	, lan_server_timer_(io_service_)
+	, dummy_player_timer_(io_service_)
+	, dummy_player_timer_interval_(30)
 {
 	setup_handlers();
 	load_config();
@@ -539,6 +541,24 @@ void server::load_config()
 #endif
 
 	load_tls_config(cfg_);
+
+	if(cfg_["dummy_player_count"].to_int() > 0) {
+		for(int i = 0; i < cfg_["dummy_player_count"].to_int(); i++) {
+			simple_wml::node& dummy_user = games_and_users_list_.root().add_child_at("user", i);
+			dummy_user.set_attr_dup("available", "yes");
+			dummy_user.set_attr_int("forum_id", i);
+			dummy_user.set_attr_int("game_id", 0);
+			dummy_user.set_attr_dup("location", "");
+			dummy_user.set_attr_dup("moderator", "no");
+			dummy_user.set_attr_dup("name", ("player"+std::to_string(i)).c_str());
+			dummy_user.set_attr_dup("registered", "yes");
+			dummy_user.set_attr_dup("status", "lobby");
+		}
+		if(cfg_["dummy_player_timer_interval"].to_int() > 0) {
+			dummy_player_timer_interval_ = cfg_["dummy_player_timer_interval"].to_int();
+		}
+		start_dummy_player_updates();
+	}
 }
 
 bool server::ip_exceeds_connection_limit(const std::string& ip) const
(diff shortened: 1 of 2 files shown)


src/server/wesnothd/server.cpp
```diff
@@ -590,6 +606,53 @@ void server::dump_stats(const boost::system::error_code& ec)
 	start_dump_stats();
 }
 
+void server::start_dummy_player_updates()
+{
+#if BOOST_VERSION >= 106600
```
Vultraz, line 611: Didn't we already bump past 1.66 min?

Pentarctagon: Yes, for the TLS PR.

src/server/wesnothd/server.cpp
```diff
@@ -590,6 +606,53 @@ void server::dump_stats(const boost::system::error_code& ec)
...
+{
+	if(ec) {
+		ERR_SERVER << "Error waiting for dummy player timer: " << ec.message() << "\n";
+		return;
+	}
+
+	int size = games_and_users_list_.root().children("user").size();
+	std::cout << "player count: " << games_and_users_list_.root().children("user").size() << std::endl;
+	if(size % 2 == 0) {
+		simple_wml::node* dummy_user = games_and_users_list_.root().children("user").at(size-1);
```
soliton-, line 629: Would be even nicer if this would cycle through users by removing from the start and adding "player<counter++>" to the end. So it'd start with removing "player0" and then add "player500", remove "player1" and add "player501" etc. That way bugs with issuing commands to the wrong user (like #5840) for example are easy to spot.

Just if you like this is fine as is, too.

soliton- (approved):

src/server/wesnothd/server.cpp
Pentarctagon: Given there's unrelated code that still uses that #if check as well, I'll fix all of them in a separate commit after this PR is merged.

src/server/wesnothd/server.cpp
```diff
@@ -590,6 +606,53 @@ void server::dump_stats(const boost::system::error_code& ec)
 	start_dump_stats();
 }
 
+void server::start_dummy_player_updates()
+{
+#if BOOST_VERSION >= 106600
+	dummy_player_timer_.expires_after(std::chrono::seconds(30));
```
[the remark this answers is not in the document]
Pentarctagon: Done.

src/server/wesnothd/server.cpp
```diff
@@ -590,6 +606,53 @@ void server::dump_stats(const boost::system::error_code& ec)
...
+
+void server::dummy_player_updates(const boost::system::error_code& ec)
+{
+	if(ec) {
+		ERR_SERVER << "Error waiting for dummy player timer: " << ec.message() << "\n";
+		return;
+	}
+
+	int size = games_and_users_list_.root().children("user").size();
+	std::cout << "player count: " << games_and_users_list_.root().children("user").size() << std::endl;
```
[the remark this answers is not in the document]
Pentarctagon: Done.

src/server/wesnothd/server.cpp
Pentarctagon: There's some strange issue going on where
```
dummy_player_count_max_++;
simple_wml::node& dummy_user = games_and_users_list_.root().add_child_at("user", size-1);
dummy_user.set_attr_dup("available", "yes");
dummy_user.set_attr_int("forum_id", dummy_player_count_max_);
dummy_user.set_attr_int("game_id", 0);
dummy_user.set_attr_dup("location", "");
dummy_user.set_attr_dup("moderator", "no");
dummy_user.set_attr_dup("name", ("player"+std::to_string(dummy_player_count_max_)).c_str());
dummy_user.set_attr_dup("registered", "yes");
dummy_user.set_attr_dup("status", "lobby");
```
is somehow duplicating the player name that I myself used to log in (in this case p2), rather than being player501, player502, etc.  So I'd rather leave this for whenever a future change is done to also enable that testing in the first place, since as-is just this wouldn't be enough to allow that testing.

PR #5845, opened 2021-06-08
Combine profile build into release build.

Pentarctagon:
---

Just letting the CI complete to make sure none of the configuration options broke.


stevecotton:
What advantage does this give?

Pentarctagon:
It's now possible to enable profiling with everything the release build is able to do.  For example, it's not possible currently to enable LTO on a profile build, and it makes more sense to add the profiler flags onto the release build when needed rather than duplicatively implement things for a separate profile build option.

Issue #5578, opened 2021-03-01
MP lobby - poor performance


Pentarctagon:
While it's been a well known problem for a while, it's apparently never been explicitly documented: once the number of players on the official multiplayer server reaches 150+, the responsiveness of the lobby can degrade to the point where typing messages and clicking buttons is noticeably delayed.  This is a major problem if we want Wesnoth's popularity to increase - people won't keep playing the game if every time more people start playing the lobby starts badly lagging.

From what I understand the cause of the lagging is because the tree view containing the list of player names is very slow to update, and right now is being cleared and replaced entirely each time anyway.

Tagging @Vultraz and @shikadiqueen for additional information as well.

Pentarctagon:
A suggestion brought up on Discord was to only update the UI on a timer (every 5 seconds?) instead of on every change like is currently done.  While it's of course better to keep the information displayed as up to date as possible, it becomes rather irrelevant if said UI can't actually handle updating that frequently.

Pentarctagon:
As an update to the above comment - it was determined that the UI doesn't actually update more than once per second even if it receives updates more frequently than that.  So the issue is definitely the time it takes for the UI to update itself once there are a certain number of players online rather than it trying to update more frequently than it can handle.

Also, for trying to replicate the lagging encountered without needing to wait for a time when 150+ players are online, there's https://github.com/renom/fastbot that could probably be used to replicate that.  Also potentially an option could be to see if there's a way to pre-populate the player list with a set of dummy players, though I'm not sure yet whether that's actually possible or not.

soliton-:
It might be easiest to use `utils/wesnothd.pl` to create a wesnothd test instance which produces whatever kind of player join/quit pattern and the according userlist update messages.

soliton-:
It's a wesnothd implementation in perl. It's not ready to use for what I suggested. One would have to implement the user activity and possibly also update a thing or two since this is quite old and there probably were a couple protocol changes... Actually it seems this still uses binary WML so nevermind. Probably too much work to update then.

I thought it would be nice to have something in a scripting language so it's simpler to quickly code up different test scenarios. But it looks like hacking in a dummy/demo mode into wesnothd proper is easier.

Using a simple client/bot implementation is certainly also an option but might cost quite a bit of resources if you want to simulate 300 users or so. Would be very flexible as well though.

Pentarctagon:
Is there any reason to keep wesnothd.pl around, in that case?

[...]

PR #5848, opened 2021-06-09
Slight optimisation of canvas::set_variables using std::move

stevecotton:
Doesn't make much difference for release builds built with GCC,
it takes between 1% and 7% off the cost of std::update_canvas().
Calculated using the time taken by gui2::widget::get_best_size()
as a reference.

It's a tiny improvement, but it's also a tiny change, and the
improvement is in the loop that's MP lobby's performance problem,
as described in issue #5578.

### DIFF
--- a/src/formula/callable.hpp
+++ b/src/formula/callable.hpp
@@ -255,6 +255,12 @@ public:
 		return *this;
 	}
 
+	map_formula_callable& add(const std::string& key, variant&& value)
+	{
+		values_[key] = std::move(value);
+		return *this;
+	}
+
 	void set_fallback(const_formula_callable_ptr fallback)
 	{
 		fallback_ = fallback;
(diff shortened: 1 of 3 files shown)


CelticMinstrel (approved):
I don't really see any issues with this change, assuming `variant&&` can still handle most of the common use-cases, which I guess is the case since you didn't change any use-cases.

I _am_ confused at why one function moved from the source file to the header but I guess it doesn't matter.

Pentarctagon (approved):

stevecotton:
> I _am_ confused at why one function moved from the source file to the header but I guess it doesn't matter.

It's just that it became a one-liner in 362eebc6f047a2e93c9dbe7f09eb29f57b8288bc, and I think the compiler might inline it if it's in the header.

PR #5849, opened 2021-06-09
Split styled_widget::update_canvas into ..._size

stevecotton:
Try and reduce GUI2's overhead of creating wfl::variants, by making a smaller function that only updates the size. Removes about three-quarters of the biggest box on those profile graphs in issue #5578, but avoids calling a virtual function that several subclasses have overridden.

WIP: Needs a careful check of the subclasses that haven't been changed in this commit, to see if any others should be changed. Going to open a draft PR as-is, to save anyone else looking at the same optimisation.

This is not the fix for #5578, the real issue there is how often this function gets called. But this will still be a general GUI2 optimisation, so worthwhile even without that issue.


Pentarctagon:
@stevecotton is this something you're still looking into?

stevecotton:
No point in keeping this open, it's not going to get to the top of my todo list anytime soon.

PR #5851, opened 2021-06-11
Fix bug in AI state loading that could cause an error when the AI is initialized from a saved game

CelticMinstrel:
It's kind of hard to explain but basically the code was doing something like this:

```lua
if ai then
  return ai
else
  return wesnoth.debug_ai(side)
end
```

And `wesnoth.debug_ai(side)` is definitely not the same thing as `ai`.


mattsc:
I cannot claim that I fully understand the changes to the code, but I tested the situation that previously caused errors due to the incompletely initialized AI after reloading.  The error is gone and all the functionality is there.  So as far as I can tell, this fixes the problem.

mattsc (approved):

PR #5855, opened 2021-06-12
Last minute Naga things

doofus-01:
- [x] Add the guardian/mace nagas from UtBS to core (+ adding LordBob's portrait for them)
- [x] Revise the Dirkfang-Ophidian branch to something without the bow
- [x] Revise descriptions (and names?) for the mace & ophidian nagas
- [x] Revise mace naga sprites so they aren't just the regular fighters holding different things in their hands (base frames only, but animations can be added later)


doofus-01:
I moved some things out of a campaign and the github-actions bot didn't care, it only noticed when I moved something into a campaign.

doofus-01:
I don't know what LordBob has planned for the Ophidian thrown weapon in his portrait, but for now I went with a type of javelin.  The sprites are very ambiguous as to what the thrown weapon is.  Since the thrown weapon probably needs to be nerfed, I added parry to the melee, to give them differentiation.

Whatever the artwork ends up being, we need to get attacks numbers straight soon.  @Hejnewar - need your input on this (parry for melee, does ranged need to be nerfed)

This image has Ophidians and my first draft at a Guardian (level 1 mace naga).  Apparently the artwork has a little more time, but unit config changes are near last call.
Screenshot_20210612_nagas

CelticMinstrel:
How exactly do they make it more deadly by adding lead? I realize lead is toxic, I'm just missing how that actually applies here (and anyway, they don't have poison on that attack).

Also… it's not that there's anything _fundamentally_ wrong about the idea of them getting their lead via scavenging, but…  I think it's nicer if they can be a _little_ more self-sufficient. It's not _that_ hard to mine some galena and melt it down to extract the lead – in fact, since nagas are amphibious, they could have access to submarine deposits that few other species can reach, and they could haul it to the shore for the smelting.

Mind you, using lead as a poison is also quite weird, as there are plenty of other applications for it. Admittedly, some of those applications aren't good ones (due to the toxicity), but they're probably better than using it as a poison, and if a poison is what they want then there are far more effective sources.

doofus-01:
It's a bit of a reach, but I was thinking more of weight and momentum rather than poison.  Small javelins don't really jibe with the existing numbers, unless there is something extra.  LordBob says he'll look into this stuff after Monday, for the portrait, but I won't be available then, so this is a placeholder.

Hejnewar:
Ok I see, problem with parry on swiming units is that they would have 80% of dodge on reef. On land it wouldn't be a big deal but water control units are mostly melee units (actually no matter if melee or ranged parry would make some faction have bad time), increase form 60% to 70% is 33% increase in tankiness of a unit but from 60% to 80% is 100%. Nothing is going to kill this unit in reasonable timeframe. I actually kind of like the concept but it's really not good place for it, not without reason 80% dodge and higher is really uncommon / nonexistent. 

Javelineer deals 11-2, 12-2 / 13-3 seems passable as javelines. 

Also because you move these nages to mainline here are correct prices, Im aware that they are recruited in UtBS that is why I wont be pushing for including them in this release but I still wanted to post them:
Naga Hunter - 19g. 
Naga Guardian - 15g.
Naga Warden - 25g.
Naga Sentinel - 44g.

CelticMinstrel:
> It's a bit of a reach, but I was thinking more of weight and momentum rather than poison.

I see… that does make sense, to a degree, although I'm not sure if it makes sense for a _ranged_ weapon.

I still question the choice of saying they get their lead from ship ballasts though (which you've misspelled by the way).

Hejnewar:
> What if you set the reef's 70% to defense cap mode? Would that prevent the 80% dodge, or does parry ignore defense cap?

Parry does ignore defense cap (it reduces enemy chance to hit according to wiki). Also it could be a bit unintuitive for players, also it does not show up in level up / unit creation tool tip if I remember correctly, also it's not at all explained in game.

doofus-01:
> Also it could be a bit unintuitive for players, also it does not show up in level up / unit creation tool tip if I remember correctly, also it's not at all explained in game.

If I understand what you mean by creation tool tip, there are a lot of things not shown there.  Many existing abilities or specials can be called unintuitive.  If it's not explained, it can be explained, using a dummy special if nothing else.

As for the 80%, the defenses can also be adjusted.

CelticMinstrel:
> If it's not explained, it can be explained, using a dummy special if nothing else.

Pretty sure parry is explained as a tooltip at least - if the unit has a weapon with parry, it'll automatically be listed like a weapon special.

Hejnewar:
> Pretty sure parry is explained as a tooltip at least - if the unit has a weapon with parry, it'll automatically be listed like a weapon special.

I opened FR for that #5860.

> If I understand what you mean by creation tool tip, there are a lot of things not shown there. Many existing abilities or specials can be called unintuitive. If it's not explained, it can be explained, using a dummy special if nothing else.

I would say that this is much more unintuitive than 1.14 mushrom grove. In mp in 1.14 you either look at tooltip or wait for like minute for unit page to load..

> As for the 80%, the defenses can also be adjusted.

Seems gimmicky, nerfing unit to add ability to make it have the same thing it had but only in certain scenarios. Well... That is probably the reason why both parry and accuracy are not used much anywhere. At least parry has potential to make for example shield that is 0-0 ranged attack with parry for unit without ranged attack.

CelticMinstrel:
> At least parry has potential to make for example shield that is 0-0 ranged attack with parry for unit without ranged attack.

You know, that's not a bad idea… and isn't there a dunefolk unit that uses a shield?

doofus-01:
> Well... That is probably the reason why both parry and accuracy are not used much anywhere.

Looking into this more, it would have been better if parry really worked defensively, but I forgot it's always in effect.  So yeah, not too useful for something like this.  Never mind.

doofus-01:
> Also because you move these nages to mainline here are correct prices, Im aware that they are recruited in UtBS that is why I wont be pushing for including them in this release but I still wanted to post them:
Naga Hunter - 19g.
Naga Guardian - 15g.
Naga Warden - 25g.
Naga Sentinel - 44g.

If the names can be changed, then the old UtBS ones could be base_unit variations, keeping all the old configurations, including the prices, for the campaign.  Warden and Sentinel are already used by other races, yet aren't so common like Warrior.  Synonyms for 'guard' that make it a little more obvious what the progression is could be helpful here.

Not moving the hunter, just the maces.

data/core/macros/abilities.cfg
```diff
@@ -481,6 +481,21 @@ Enemy units cannot see this unit while it is in deep water, except if they have
...
 #enddef
 
+#define WEAPON_SPECIAL_DEFLECT
+    # Canned definition of the Deflect ability to be included in a
+    # [specials] clause.
+    [chance_to_hit]
+        id=deflect
+        name= _ "deflect"
+        description= _ "When used defensively, this attack reduces the opponent's chance to hit by 10%."
+        special_note={INTERNAL:SPECIAL_NOTES_DEFLECT}
```
CelticMinstrel, line 491: The `INTERNAL` macros are just a deprecation pathway to avoid writing the strings out twice. For new abilities, you should put the string here where it's used.

data/core/units/nagas/Sicarius.cfg
```diff
@@ -16,15 +16,13 @@
...
+    description= _ "While they sometimes still stand in as sellswords for potential Dunefolk allies, Naga Sicarii are more often the keepers of trade routes and resource beds close to waterways. For the right fee or exchange of goods, a Sicarius will guarantee safe travel or free access to valuable supplies in his territory. Rubbed the wrong way and a Sicarius becomes a fearsome foe, not because of their personal strength in combat, but instead due to the numerous allies that he can call upon to quash any potential nuisance. Though perfectly capable warriors when the time is necessary, these experienced mercenaries know perfectly well that the best methods for generating money are those that do not place themselves in danger.  When they do find themselves in danger, their manipulative cunning is put to good use, for they can anticipate most enemy strikes, and use their heavy, curved blade to deflect or distract at the critical moment."
```
CelticMinstrel, line 19: I think "most" sounds a bit strong; how about "many" instead?

doofus-01: Sure, I was just trying to make it not be "all".  This is compatible with marksman (makes it 50%) but I'm not sure how this interacts with noncumulative (magical), need to test a bit more.

data/core/macros/abilities.cfg
doofus-01: OK, I was just following all the other notes, but I think I understand.

doofus-01:
There is still work to do, but as far as the stuff due before 6/20, I think this is completed.  I will merge once it passes CI.

data/core/macros/abilities.cfg
```diff
@@ -534,6 +534,19 @@ Enemy units cannot see this unit while it is in deep water, except if they have
...
 #enddef
 
+#define WEAPON_SPECIAL_ABSORB AMT
+    # Canned definition of the Absorb ability to be included in a
+    # [specials] clause.
+    [damage]
+        id=absorb
+        name= _ "absorb"
+        description= _ "This attack reduces the damage suffered from an enemy strike."
+        special_note= _ "This attack puts the unit in good defensive position, and it absorbs some of the damage dealt by an enemy strike."
```
CelticMinstrel, line 544: This special note seems like it might be better as the description, and the special note should be written from the perspective of the unit rather than the attack.

doofus-01: OK.  The special notes could probably be completely removed, but that's out of scope here.

AI0867:
> > It's a bit of a reach, but I was thinking more of weight and momentum rather than poison.
> 
> I see… that does make sense, to a degree, although I'm not sure if it makes sense for a _ranged_ weapon.

For the same reason bullets are made mostly out of lead. Lead is 50% more dense than iron, and has 10+ times the density of wood or bone. This increases the mass-to-surface-area ratio and allows you to retain the momentum for longer. This is even more important under water than in the air.

PR #5748, opened 2021-04-28
[MP] Isle of Mists - a new 1p/2p Survival Scenario. 

knyghtmare:
Description:
_Isle of Mists is a survival scenario for solitaire or two-player team-based play against randomly-spawned AI units. Victory is achieved by surviving and defeating all enemy waves. During the course of play, the terrain will change based on random weather effects. You need to use the default map settings for the scenario to work right._

This PR contains:
1. Adds a new multiplayer survival scenario.
2. Lua (similar to Dark Forecast one) code file used for the enemy waves and weather change events
3. Comes with 6 weather conditions (6 maps)

Features:
1. Default difficulty is set at `0`.
2. Has 1p/2p modes
3. 5 final waves
4. 8 difficulty levels, ranging from `-30` to `+60`, with `+60` being catered especially for veteran/skilled players.

Replays: 
Default difficulty | 2player mode | Drakes + Rebels
[2p — Isle of Mists (Survival) replay 20210429-050016.gz](https://github.com/wesnoth/wesnoth/files/6395359/2p.Isle.of.Mists.Survival.replay.20210429-050016.gz)

Reviewers:
I would like to request reviews from nemaara, Pentarctagon and whoever is involved in MP scenario content development.

PR #5856, opened 2021-06-12
Balance changes for 1.15.14

Hejnewar:
For changelog:
Dune Apothecary experience changed from 100 to 65.
Dune Captain experience changed from 86 to 75.
Dune Horse Archer experience changed form 70 to 65.
Naga Sicarius hp changed form 55 to 53 and melee damage changed form 9 to 6, new special "deflect" added to melee attack.
Fire Guardian can now level into Fire Wraith.
Revised statistics of all animal horses in order for them to fit better within lore and not violate PIPLB. Bay Horse can now level into Great Horse.

Jinn - a lot changed within JInn, but the most impactful one is resistance change, it should allow for more wide use of this unit.
Fire Wraith cost changed from 25 ot 22.

Horses were changed a lot here are the most important changes:
Bay Horse can now level into Great Horse. 
White Horse now has resistances and movement type very similar to elvish horses, it works well with its description. 
Dark Horse now has "whinny" attack (in my opinion animation of this attack looks better on Dark Horse than it does on Black Horse), Dark Horse no longer violates RIPLB while leveling up.

I tried to make animal horses as distinct as possible while still staying horses in core. Bay and Great Horses are just plain old horses, Dark and Black are a bit nightmareish but still horses, White one is closest to elvish ones.

### DIFF
--- a/data/core/units/monsters/Jinn.cfg
+++ b/data/core/units/monsters/Jinn.cfg
@@ -11,17 +11,24 @@ units/monsters/jinn#enddef
     image="{IMG_PATH_TEMP}/jinn.png"
     hitpoints=58
     movement_type=spirit
-    movement=8
+    movement=6
     experience=100
     level=2
     alignment=liminal
     advances_to=null
     {AMLA_DEFAULT}
-    cost=56
+    cost=54
     usage=archer
     description=""
     die_sound={SOUND_LIST:HUMAN_OLD_DIE}
     undead_variation=null
+    [resistance]
+        blade=80
+        pierce=70
+        impact=100
+        fire=80
+        cold=100
+    [/resistance]
     [defend]
         start_time=-126
         [frame]
@@ -101,7 +108,7 @@ units/monsters/jinn#enddef
         icon=attacks/lightning.png
         type=fire
         range=ranged
-        damage=18
+        damage=20
         number=1
         [specials]
             {WEAPON_SPECIAL_MAGICAL}
--- a/data/core/units/monsters/Jinn.cfg
+++ b/data/core/units/monsters/Jinn.cfg
@@ -17,7 +17,7 @@ units/monsters/jinn#enddef
     alignment=liminal
     advances_to=null
     {AMLA_DEFAULT}
-    cost=54
+    cost=53
     usage=archer
     description=""
     die_sound={SOUND_LIST:HUMAN_OLD_DIE}
@@ -27,7 +27,7 @@ units/monsters/jinn#enddef
         pierce=70
         impact=100
         fire=80
-        cold=100
+        cold=120
     [/resistance]
     [defend]
         start_time=-126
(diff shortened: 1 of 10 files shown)


data/core/units/monsters/Jinn.cfg
```diff
@@ -11,17 +11,24 @@ units/monsters/jinn#enddef
...
     description=""
     die_sound={SOUND_LIST:HUMAN_OLD_DIE}
     undead_variation=null
+    [resistance]
+        blade=80
+        pierce=70
+        impact=100
+        fire=80
+        cold=100
+    [/resistance]
```
CelticMinstrel, line 31: I think I'd raise cold a little, to 110 or maybe even 120. A vulnerability to cold seems reasonable for a desert-dwelling creature, doesn't it?

Other than that I don't have any particular objections to these resistances.

data/core/units/monsters/Jinn.cfg
```diff
@@ -101,7 +108,7 @@ units/monsters/jinn#enddef
         icon=attacks/lightning.png
         type=fire
         range=ranged
-        damage=18
+        damage=20
```
CelticMinstrel, line 111: 20 seems really high… it _is_ a level 2 unit but I wonder if it might be better to split this damage across two strikes?

CelticMinstrel:
So I guess you decided a flame aura wasn't a good fit?

data/core/units/monsters/Jinn.cfg
Hejnewar: I was thinking about it but after reading comment about animation I decided to leave it as one.

data/core/units/monsters/Jinn.cfg
Hejnewar: Reducing cold resistance is fine. I will do that later.

Hejnewar:
Honestly I'm recently either busy or tired and I will be like that until after or very close to 1.15.14 release and I still have to review #5748. Thinking about it now I would say that it would fit Fire Wraith better, either as constant effect, effect scaling with hp or even one time nova after losing x% of hp after which unit would lose its armour and turn into much more offensive one but as I'm right now I'm really not willing to engage with anything more complicated than simple changes, even thinking about horses took me more time than it should because I really wanted them to be distinct and horses and fitting.

CelticMinstrel:
Yeah, it would _definitely_ fit the Fire Wraith… but I also just wish the Jinn had at least _one_ thing that's distinctly fiery…

Also, looking at the full Jinn definition, I just realized… does it make sense for the wind blast to be impact? Is this just a blast of wind, or is it a blast of wind filled with sand? In the latter case, it _might_ make sense to be pierce.

Hejnewar:
It does make sense for wind blast to be impact. It's not called sandstorm so I assume that it's just that, wind blast.

Hejnewar:
Updated first post with changes. Some units just didn't look good with their current xp costs and some looked like were left at old xp cost for some reason so I fixed them. Naga recived some changes in order to accomodate new special without any price changes, yes it still is that strong and Jinn got -20% cold resistance.

Hejnewar:
Meh, well I will take care of that latter then.

CelticMinstrel:
I realize it's extremely late to be asking this but… what's a "sicarius"?

Hejnewar:
I remember that I checked that sometime ago in google and it was ruffian / killer / murderer in latin.

Issue #2500, opened 2018-02-13
Wesnoth commands report


ProditorMagnus:
I marked with box those that I think need action.

Missing help entry
- [x] 	benchmark
- [x] 	theme

Incomplete help
- 	friend accepts note after name
- 	report alias is query adminmsg, in addition to just adminmsg

Unneeded
- 	controller - Lua and WML have access to it
- [x] 	details
- [x] 	drop
- [x] 	register
- [x] 	set

Room commands do nothing on 1.13, regression - though they have extremely limited use even in 1.12
- [x] 	join
- [x] 	names
- [x] 	part
- [x] 	room
- [x] 	room_query - topic at least doesnt work on 1.12 either
- [x] 	rooms

Untested
- [x] 	create - alternate for create unit menu?
- [x] 	throw - I guess same as fire_event as called from lua

Misc
- [ ] 	control/give_control - needs GUI access as well for 1.14
- 	"q" does not give help message that it is supposed to - when writing out query, or "q ", then that works
- 	log - unsure what domains and levels exist
- 	refresh - not sure what it does
- 	whiteboard, whiteboard_options - I guess first one is same as planning mode, second one if it still works should be reachable from same menu
- 	alias does not work when used with /
- [x] 	benchmark and fps are duplicates

stevecotton:
Half of those can't be fixed during a string freeze. :(

ProditorMagnus:
Only give_control GUI access is such that would be needed for 1.14.

soliton-:
Perhaps not so great to have that many different issues in one report. Thanks for the overview nevertheless.

CelticMinstrel:
If the GUI access can be done using only existing strings, sure?

doofus-01:
I think "refresh" regenerates the image cache.  It's useful for drawing/testing terrain or sprite images, not sure a regular player would have reason to use it though.  

I think there used to be a command "sunset" that could have gone in the same category, but it doesn't seem to work anymore.  Maybe I'm miss-remembering the name.

ProditorMagnus:
Sunset I reported long time ago, gna 22979.

Follow-up Comment #2:

Sunset was removed in 1.13.7.

Follow-up Comment #1:

I'm not convinced this is actually a bug.

This appears to be an extremely old debugging aid for developers, the purpose
is to help visualize what pixels are being redrawn during screen refresh
events. It might also be useful to see this during the main menu and such,
[...]

PR #5857, opened 2021-06-13
Add a method to clear and then add multiple items to a tree view.

Pentarctagon:
This addresses the lobby performance issue as far as the playerlist updating. Previously for every player an item was added and the size of the tree view recalculated, which led to an increasingly unresponsive UI when the number of players in the lobby got to around 150+. With this change, the replace_children() method adds all the players and then after that recalculates the tree view's size once.

---

With a -O0 build, it now takes a fraction of a second to update the playerlist, down from over 30 seconds previously, when adding/removing a player from a playerlist with 500 entries.

For reference with #5578, the profile with this change applied:
prof

### DIFF
--- a/src/gui/dialogs/multiplayer/lobby.cpp
+++ b/src/gui/dialogs/multiplayer/lobby.cpp
@@ -104,7 +104,7 @@ void sub_player_list::update_player_count_label()
 void player_list::init(window& w)
 {
 	active_game.init(w, _("Selected Game"), true);
-	other_rooms.init(w, _("Lobby"), true);
+	lobby_players.init(w, _("Lobby"), true);
 	other_games.init(w, _("Other Games"));
 
 	tree = find_widget<tree_view>(&w, "player_tree", false, true);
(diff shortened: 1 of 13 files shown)


src/gui/dialogs/multiplayer/lobby.cpp
```diff
@@ -693,12 +694,16 @@ void mp_lobby::update_playerlist()
 		tree_group_field["use_markup"] = "true";
 		tree_group_item["name"] = tree_group_field;
 
-		tree_view_node& player = target_list->tree->add_child("player", tree_group_item);
+		//tree_view_node& player = target_list->tree->add_child("player", tree_group_item);
```
CelticMinstrel, line 697: I assume commented code will be removed before merging, right?

src/gui/dialogs/multiplayer/lobby.cpp
```diff
@@ -693,12 +694,16 @@ void mp_lobby::update_playerlist()
...
 
-		connect_signal_mouse_left_double_click(find_widget<toggle_panel>(&player, "tree_view_node_label", false),
-			std::bind(&mp_lobby::user_dialog_callback, this, &user));
+		//connect_signal_mouse_left_double_click(find_widget<toggle_panel>(&player, "tree_view_node_label", false),
```
CelticMinstrel, line 699: Is this missing in the new implementation? I don't see anything that connects signals.

src/gui/widgets/tree_view.cpp
```diff
@@ -59,7 +59,7 @@ tree_view_node& tree_view::add_node(
 	return get_root_node().add_child(id, data, index);
 }
 
-std::pair<tree_view_node::ptr_t, int> tree_view::remove_node(tree_view_node* node)
+std::pair<std::shared_ptr<tree_view_node>, int> tree_view::remove_node(tree_view_node* node)
```
CelticMinstrel, line 62: Why did the type change here? It looks like you just removed a typedef… why?

src/gui/dialogs/multiplayer/lobby.cpp
Pentarctagon: Yes.

src/gui/dialogs/multiplayer/lobby.cpp
Pentarctagon: Yeah, it's currently incomplete.

src/gui/widgets/tree_view.cpp
Pentarctagon: Because it was barely used and `ptr_t` is so undescriptive that it just ends up taking more time to remember/check what that type actually is than it will ever save anyone in the slightly reduced typing.

Pentarctagon:
It should be mostly good now.  Additionally it removes the last bits of rooms support since it's related, which addresses part of #2500.

As a separate note as well, while the playerlist refresh is now much faster, scrolling the list itself is quite laggy.

CelticMinstrel:
Seems strange that this removes the "group players" setting. Also seems strange that that setting doesn't actually do anything. Am I missing something? Is it _not_ supposed to switch between the tree-based view (selected game / lobby / other games) and a flat user list?

Pentarctagon:
I'm not sure where it's said that it doesn't do anything.  It shouldn't be a huge problem to re-add it though if needed.

CelticMinstrel:
> I'm not sure where it's said that it doesn't do anything.

It doesn't say anywhere that it doesn't do anything, but when you removed it in that push, I only saw a single reference to it, which is something that does nothing because rooms don't actually exist.

Pentarctagon:
The group players setting does currently combine the currently selected game's players, the players in the lobby, and the players in other games into a single list.  I removed it because it's an advanced preference that's been defaulted to off for ~12 years, so the odds that any significant number of people are using it anymore is very low.

CelticMinstrel:
Hmm, so those three lines you removed were all that was necessary for players to not be grouped?

Well, whatever… it looks strange but I guess it doesn't matter.

soliton-:
I don't see how the default of an option suggests how many people change it. I would remove that option just because it seems quite useless.

Would be great if it were done in a separate commit though so that the commits are more focused on a single change.

Pentarctagon:
Split it into a separate commit.

PR #5858, opened 2021-06-13
Tutorial S02 and tips: Mention holding shift for accelerated speed

stevecotton:
One of these uses bold for the menu's name, the other uses italic, because that
seems more consistent with the other hints in the same context.

Re-attach the comment about upkeep to the correct [event], and move the comment
about the easter-egg event to the start of its [event].

### DIFF
--- a/data/campaigns/tutorial/scenarios/02_Tutorial_part_2.cfg
+++ b/data/campaigns/tutorial/scenarios/02_Tutorial_part_2.cfg
@@ -405,7 +405,7 @@ If either Shaman advances to become a Druid, then she’ll be able to heal adjac
         [/message]
     [/event]
 
-    # Mention upkeep and income
+    # Easter egg if you recruit all Shamans
     [event]
         name=recruit,recall
         [filter]
@@ -425,7 +425,6 @@ If either Shaman advances to become a Druid, then she’ll be able to heal adjac
 
         {CLEAR_PRINT}
 
-        # Easter egg if you recruit all Shamans
         [message]
             speaker=Galdrad
             message= _ "You certainly have a fondness for shamans, I see. They will provide a lot of healing, but not much offensive potential."
@@ -449,6 +448,7 @@ If either Shaman advances to become a Druid, then she’ll be able to heal adjac
         [/disallow_end_turn]
     [/event]
 
+    # Mention upkeep and income; or remind about undo if capturing a village without recruiting
     [event]
         name=capture
         first_time_only=no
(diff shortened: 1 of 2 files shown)


stevecotton:
About to force-push, it will be the same content but added to a different place in tips.cfg.

I put the new tip next to the "Read the Hotkeys list in the Preferences menu." because they were both about the same menu. I'm now thinking about making a "delete some tips" PR, which would have a merge conflict.

CelticMinstrel:
Just double-checking, but is this key rebindable? I looked through the list of hotkeys in the preferences and didn't see it, but I'm not confident I didn't just miss it.

Certainly it's the sole case of an action bound to _just_ a modifier, so it might be a special case…

stevecotton:
It's not rebindable, it's hardcoded here:
https://github.com/wesnoth/wesnoth/blob/2df6235efd646fea909cf181314171bdd751504e/src/display.cpp#L2399-L2404

Added @irydacea as a reviewer, as she's considering turning acceleration on by default.

irydacea:
> Added @irydacea as a reviewer, as she's considering turning acceleration on by default.

Removing myself because what I said on the forums is more of a historical thing and I'm not interested in having that debate with the art team again.

nemaara:
The way you've worded it isn't always true if the acceleration factor is set to less than 1. I'm not sure what the use case of that is for regular players (I sometimes use it to look at my animations really carefully but not for gameplay), but it is there after all.

stevecotton:
> isn't always true if the acceleration factor is set to less than 1

I could reword it, but making it strictly accurate is either going to make it longer or make it less obvious that it can (normally) speed things up.

data/campaigns/tutorial/scenarios/02_Tutorial_part_2.cfg
```diff
@@ -1013,6 +1013,17 @@ Rest-healing is an exception to the rule — if a unit doesn’t do anything for
...
     [/event]
 
+    [event]
+        name=turn 12
+
+        [message]
+            speaker=narrator
+            caption= _ "Speeding Up Animations"
+            image=wesnoth-icon.png
+            message= _ "Holding down <b>shift</b> will make the animations for moving and fighting quicker. The <b>acceleration factor</b> can be set, and turned on by default, in the <b>Preferences</b> menu’s <b>General</b> tab."
```
nemaara, line 1023:
```suggestion
            message= _ "Holding down <b>shift</b> will make the animations for moving and fighting quicker. The <b>acceleration factor</b> can be set and turned on by default in the <b>Preferences</b> menu’s <b>General</b> tab."
```

nemaara (approved):

nemaara:
I'm not sure, I'll leave that up to you. Otherwise I've just suggested removing the commas in the current text since I feel like it reads better to me. It's not a big deal, it can be merged as is if you like it.

PR #5861, opened 2021-06-15
Move the end level data accessors to wesnoth.scenario

CelticMinstrel:


CelticMinstrel:
Since this passed I guess it's ready for review?

CelticMinstrel:
That's weird… the code is pretty much the same, so why has FAIL turned into FAIL_BY_DEFEAT!?

CelticMinstrel:
Now that CI passes, this is ready for review.

src/scripting/game_lua_kernel.cpp
```diff
@@ -1188,6 +1188,61 @@ namespace {
 	}
 }
 
+static int impl_end_level_data_get(lua_State* L)
+{
+	const end_level_data& data = *static_cast<end_level_data*>(lua_touserdata(L, 1));
+	const char* m = luaL_checkstring(L, 2);
```
stevecotton, line 1194: What is `m` used for?

src/scripting/game_lua_kernel.cpp
```diff
@@ -1188,6 +1188,61 @@ namespace {
...
+		end_level_data& data_;
+		play_controller& pc_;
+	};
+}
+
+int game_lua_kernel::impl_end_level_data_set(lua_State* L)
+{
+	end_level_data& data = *static_cast<end_level_data*>(lua_touserdata(L, 1));
+	const char* m = luaL_checkstring(L, 2);
+	end_level_committer commit(data, play_controller_);
```
stevecotton, line 1227: What error state is being avoided by using this committer object instead of just calling set_end_level_data directly?

src/scripting/game_lua_kernel.cpp
CelticMinstrel: That's just the key being looked up. Every getter function calls it m… I think it's because the macros expect that?

I didn't write this code for the record, it was just copy-pasted from further down the file.

src/scripting/game_lua_kernel.cpp
CelticMinstrel: The `modify_*` macros include a return statement, so the only alternative to this would be adding a `set_end_level_data` call to every one of them.

stevecotton: Oh, so this gets called multiple times, once to set each member of the object. I'm still on the learning curve for this code.

stevecotton (approved):

PR #5866, opened 2021-06-16
Add support for forum_auth when uploading to campaignd.

Pentarctagon:
---

Re-PR of #5836 so it's off my fork's master branch since it won't make it for 1.16.0.

@irydacea When you have some time, I'd appreciate an in depth description of what you're looking for in terms of UI changes and how the password would be stored.  I know we talked about it a bit on Discord, but I'm still having a hard time coming up with an implementation.


Pentarctagon:
@irydacea Coming back to this, now that 1.16 is out, can you elaborate on what you'd like to see here?

PR #5870, opened 2021-06-19
gui2/addon_manager: Add TLS connection indicator

irydacea:
This implements the TLS status indicator for the add-ons manager UI.

Since this is a rather trivial feature with no real impact on anything else, I would like to work on the MP version of it and get it merged during the feature freeze reusing the strings from this PR to avoid bothering translators. For this purpose, however, I'd like input on the tooltips below asap since I don't think they are very good at the moment.

### Without TLS, small window

Screenshot_20210618_203236
Tooltip:
> Traffic between the game and the server is not secure

### With TLS, big window

Screenshot_20210618_203342
Tooltip:
> Traffic between the game and the server is encrypted


irydacea:
Incidentally, I'm open to using the top right corner of the dialog instead, although personally I would prefer to reserve it for search/filtering options across the board.

loonycyborg (approved):

Vultraz (approved):

Pentarctagon (approved):

PR #5871, opened 2021-06-19
Remove the wesnoth-test textdomain

CelticMinstrel:
This is related to #5800 and (assuming I didn't miss anything) should ensure that nothing from this textdomain ever gets translated.

This allows us to mark things translatable without having to violate schema validation that checks whether things should be translatable.


stevecotton:
Not sure who should review, but this looks OK to me (just comparing it to a42600ffbe09049170fd66a1d5eab2a4e9dd69f6).

PR #5875, opened 2021-06-20
[UtBS] dustbok description for 1.16

doofus-01:
as advertised, nothing more.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/units/quenoth/Dustbok.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/quenoth/Dustbok.cfg
@@ -15,7 +15,7 @@
     advances_to=null
     cost=12
     usage=scout
-    description= ""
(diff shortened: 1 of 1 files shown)


data/campaigns/Under_the_Burning_Suns/units/quenoth/Dustbok.cfg
```diff
@@ -15,7 +15,7 @@
     advances_to=null
     cost=12
     usage=scout
-    description= ""
+    description= "Dustboks are nimble, graceful creatures whose timid and peaceful nature stands in sharp contrast to the harsh, barren land in which they live.  Because the scarred, scorched land cannot support herds or even territorial neighbors, dustboks are solitary wanderers and yet still social, curious, and willing to cooperate with anyone if it leads to food.  While this behavior has led to the death of many foolish dustbok, it has made them excellent partners to the Quenoth Elves, who value their speed and agility."
```
nemaara, line 18:
```suggestion
    description= "Dustboks are nimble, graceful creatures whose timid and peaceful nature stands in sharp contrast to the harsh land in which they live.  Because the scorched land cannot support herds or even territorial neighbors, dustboks are solitary wanderers and yet still social, curious, and willing to cooperate with anyone if it leads to food.  While this behavior has led to the death of many a foolish dustbok, it has made them excellent partners to the Quenoth Elves, who value their speed and agility."
```

Bit too many adjectives at the start.

PR #5876, opened 2021-06-20
1.16 graphics for new(ish) monsters

doofus-01:
- revise fire wraith halo
- add LordBob's portraits for Jinn and Fire ghosts


doofus-01:
Probably more is needed, but I won't get it done in the next few hours, so I'll be merging this soon.

PR #5880, opened 2021-06-20
1.16 - some naga animations

doofus-01:
Just attack/defend for guard and defend for shield guard.


soliton-:
It would be great to squash and merge such changes to get rid of the "ran woptipng" commit or rather the unoptimized images. 

Pretty much anytime there are "fixup" commits in a PR these should be squashed before merging either by doing that locally and force pushing or using the github function if one resulting commit is fine.

PR #5882, opened 2021-06-21
Use the built-in difficulty selection for World Conquest

CelticMinstrel:
Since there seems to be no way to put difficulty selection on a single scenario, this moves it to the MP campaigns section.

There are some kinks to work out with this still. I think possibly it may even end up collapsing the three WC scenarios into one? However, it seems to work at least on a basic level, so if anyone wants to try it out and point out anything that's broken that I missed…

Opened against master because this _does_ add one minor translatable string. For 1.16.0 we could probably just leave it untranslatable, since it's only a campaign abbreviation.


data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg
```diff
@@ -22,17 +22,64 @@ _ "World Conquest 3p" #enddef
...
+        description = {WC_II_CAMPAIGN_DESC_{PLAYERS}P}
+        first_scenario = "WC_II_{PLAYERS}p{ID_SUFFIX}"
+        type = mp
+        abbrev = _ "WC"
+        {WC2_CAMPAIGN_DIFFICULTY VERY_EASY {WC2_HUMAN_DIFFICULTY human-peasants/peasant purple} _"Peasant" _"Very easy" 6 2 2 10 yes 0}
+        {WC2_CAMPAIGN_DIFFICULTY EASY {WC2_HUMAN_DIFFICULTY human-loyalists/sergeant black} _"Sergeant" _"Easy" 7 3 2 7 yes 5}
+        {WC2_CAMPAIGN_DIFFICULTY NORMAL {WC2_HUMAN_DIFFICULTY human-loyalists/lieutenant brown} _"Lieutenant" _"Normal" 8 4 2 5 yes 10} {DEFAULT_DIFFICULTY}
+        {WC2_CAMPAIGN_DIFFICULTY HARD {WC2_HUMAN_DIFFICULTY human-loyalists/general orange} _"General" _"Hard" 8 5 2 2 no 13}
+        {WC2_CAMPAIGN_DIFFICULTY VERY_HARD {WC2_HUMAN_DIFFICULTY human-loyalists/marshal white} _"Grand Marshal" _"Very hard" 9 6 2 1 no 17}
+        {WC2_CAMPAIGN_DIFFICULTY NIGHTMARE {WC2_NIGHTMARE_DIFFICULTY} _"Nightmare" _"Expert" 9 7 1 0 no 20}
```
Pentarctagon, line 72: The rest I assume are existing strings, but are "Very easy" and "Very hard" already existing?

CelticMinstrel: Oh right, there's these too.

The one I was thinking of was `abbrev=_"WC"` but yeah, I guess there's three new strings so far.

Even though the others _are_ existing strings as far as I know, some of them might not be in the WC textdomain specifically, so I'll need to make sure to add some #textdomain statements if this is to go into 1.16.0.

knyghtmare: can "very easy" and "very hard" be renamed to "Novice/Rookie/Beginner" and "Challenging/Difficult" respectively?

knyghtmare:
https://github.com/wesnoth/wesnoth/blob/ee672e8a80b2acd8d13576848edbc64074340932/data/campaigns/World_Conquest/config.cfg#L17-L36

Not Sure if you would patch this or not, but is add-on mode necessary for something which is already in the mainline campaigns directory?

https://github.com/wesnoth/wesnoth/blob/ee672e8a80b2acd8d13576848edbc64074340932/data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg#L60-L66

Seems to be missing two/three attributes:
```
icon = 
background = 
image = 
```
Not sure if you would be addressing these, but I felt like pointing it out.

data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg
CelticMinstrel: Sure.

CelticMinstrel:
> Not Sure if you would patch this or not, but is add-on mode necessary for something which is already in the mainline campaigns directory?

It's not necessary, no. I may remove it at some point but not as part of this PR.

> Seems to be missing two/three attributes:

Yes, I'm aware of that. Suggestions as to what to use there would be welcome.

knyghtmare:
> > Seems to be missing two/three attributes:
> 
> Yes, I'm aware of that. Suggestions as to what to use there would be welcome.

image

I have used the 6 main MP factions for image icons and IPFs. Will they do, or do we need to spice them up? I can PR to this branch or give the code here.

I tried using attack icons as backgrounds but that went terribly wrong so I left them with the unit sprites.

as for the image, I am open to anything that fits into "tropical world domination"

image

CelticMinstrel:
Note: I'm not sure that it will continue to be three separate campaigns, split by the number of players. If I can do it as one campaign where you select the player count separately, I probably will.

That aside, I have no particular objection to any of your proposed icons. I guess a PR to this branch is the logical choice for someone who doesn't have push access, but it may be better to wait until I've figured out whether combining them will actually work.

knyghtmare:
> Note: I'm not sure that it will continue to be three separate campaigns, split by the number of players. If I can do it as one campaign where you select the player count separately, I probably will.

I do agree with having it as one campaign "World Conquest" would be the ideal solution.
```
    [campaign]
        # specifics for multiplayer
        allow_era_choice=yes
        min_players=1
        max_players=3

        # add-on specific
        require_campaign=yes
    [/campaign]
```

> but it may be better to wait until I've figured out whether combining them will actually work.

Yes, it is no rush. Gives ample time to churn up a good campaign icon. 

As for the `background=` one for WC, I believe that the MP Create Game does not support it, so it's only for SP campaigns.

CelticMinstrel:
I'm going to mark it as ready for review because it seems to be 90% working now… however, when playing 1-player or 2-player, there are still two empty sides visible that should be hidden. They're marked as hidden, so I don't understand why they're visible.

I abandoned the idea of making it a single campaign after discovering that `[option]` tags are not visible to the map generator and map generator settings are not available on a `[campaign]`. Those are the only two methods I can think of to pass that information to the map generator. I think for master we should _make_ `[option]` tags be visible and merge the three campaigns into one, but for 1.16 I don't think that's an option.

data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg
```diff
@@ -22,22 +22,89 @@ _ "World Conquest 3p" #enddef
...
+        id = {ID}
+        name={NAME}
+        allow_new_game=no
         scenario_generation=lua
+        experience_modifier=100
+        victory_when_enemies_defeated=yes
+        carryover_percentage=0
+        carryover_report=no
+        carryover_add=no
+        force_lock_settings=yes
```
CelticMinstrel, line 104: I moved these from Lua to WML as part of trying to figure out why the hidden sides show… so maybe it's better not to move them.

data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg
```diff
@@ -22,22 +22,89 @@ _ "World Conquest 3p" #enddef
...
+        define = CAMPAIGN_WC_{PLAYERS}P
         name= {WC_II_CAMPAIGN_NAME_{PLAYERS}P}
         description = {WC_II_CAMPAIGN_DESC_{PLAYERS}P}
-        allow_new_game=yes
-        require_scenario=no
+        first_scenario = "WC_II_{PLAYERS}p"
+        min_players={PLAYERS}
+        max_players={PLAYERS}
+        type = mp
+        abbrev = _ "WC"
```
CelticMinstrel, line 73: As far as I can tell, this is the sole _new_ string in here. I think leaving it untranslated for 1.16.0 is probably an acceptable compromise? It could become translated as of 1.16.1.

Side note: Should it include the player count?

data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg
```diff
@@ -22,22 +22,89 @@ _ "World Conquest 3p" #enddef
...
+        {WC2_CAMPAIGN_DIFFICULTY VERY_EASY {WC2_HUMAN_DIFFICULTY human-peasants/peasant purple} {STR_PEASANT} {STR_BEGINNER} 6 2 2 10 yes 0}
+        {WC2_CAMPAIGN_DIFFICULTY EASY {WC2_HUMAN_DIFFICULTY human-loyalists/sergeant black} {STR_SERGEANT} {STR_EASY} 7 3 2 7 yes 5}
+        {WC2_CAMPAIGN_DIFFICULTY NORMAL {WC2_HUMAN_DIFFICULTY human-loyalists/lieutenant brown} {STR_LIEUTENANT} {STR_MEDIUM} 8 4 2 5 yes 10} {DEFAULT_DIFFICULTY}
+        {WC2_CAMPAIGN_DIFFICULTY HARD {WC2_HUMAN_DIFFICULTY human-loyalists/general orange} {STR_GENERAL} {STR_HARD} 8 5 2 2 no 13}
+        # Challenging is a string that exists in mainline but only in campaigns so it probably counts as adding a string?
+        #textdomain wesnoth-httt
+        {WC2_CAMPAIGN_DIFFICULTY VERY_HARD {WC2_HUMAN_DIFFICULTY human-loyalists/marshal white} {STR_GRAND_MARSHAL} _"Challenging" 9 6 2 1 no 17}
+        # Expert is a string that was supposed to be in wesnoth-wc but accidentally got added to wesnoth-units instead
+        #textdomain wesnoth-units
+        {WC2_CAMPAIGN_DIFFICULTY NIGHTMARE {WC2_NIGHTMARE_DIFFICULTY} {STR_NIGHTMARE} _"Expert" 9 7 1 0 no 20}
```
CelticMinstrel, line 84: This string was in the removed difficulty.lua but was placed in the wrong textdomain. For master the textdomain declaration should be removed and it just goes in wesnoth-wc.

data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg
```diff
@@ -22,22 +22,89 @@ _ "World Conquest 3p" #enddef
...
+        type = mp
+        abbrev = _ "WC"
+        require_campaign=no
+        {WC2_CAMPAIGN_DIFFICULTY VERY_EASY {WC2_HUMAN_DIFFICULTY human-peasants/peasant purple} {STR_PEASANT} {STR_BEGINNER} 6 2 2 10 yes 0}
+        {WC2_CAMPAIGN_DIFFICULTY EASY {WC2_HUMAN_DIFFICULTY human-loyalists/sergeant black} {STR_SERGEANT} {STR_EASY} 7 3 2 7 yes 5}
+        {WC2_CAMPAIGN_DIFFICULTY NORMAL {WC2_HUMAN_DIFFICULTY human-loyalists/lieutenant brown} {STR_LIEUTENANT} {STR_MEDIUM} 8 4 2 5 yes 10} {DEFAULT_DIFFICULTY}
+        {WC2_CAMPAIGN_DIFFICULTY HARD {WC2_HUMAN_DIFFICULTY human-loyalists/general orange} {STR_GENERAL} {STR_HARD} 8 5 2 2 no 13}
+        # Challenging is a string that exists in mainline but only in campaigns so it probably counts as adding a string?
+        #textdomain wesnoth-httt
+        {WC2_CAMPAIGN_DIFFICULTY VERY_HARD {WC2_HUMAN_DIFFICULTY human-loyalists/marshal white} {STR_GRAND_MARSHAL} _"Challenging" 9 6 2 1 no 17}
```
CelticMinstrel, line 81: This string exists in most mainline campaigns. I chose HttT more or less at random to reference it from. For master however it should just be added to wesnoth-wc.

data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg
```diff
@@ -22,22 +22,89 @@ _ "World Conquest 3p" #enddef
...
         name= {WC_II_CAMPAIGN_NAME_{PLAYERS}P}
         description = {WC_II_CAMPAIGN_DESC_{PLAYERS}P}
-        allow_new_game=yes
-        require_scenario=no
+        first_scenario = "WC_II_{PLAYERS}p"
+        min_players={PLAYERS}
+        max_players={PLAYERS}
+        type = mp
+        abbrev = _ "WC"
+        require_campaign=no
```
CelticMinstrel, line 74: I suspect this is redundant, as it's designed for add-ons but this is mainline content. Can I get some confirmation here?

knyghtmare:
I tried to load this but:
image

data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg
knyghtmare: > Side note: Should it include the player count?

I think the abbrev should be dependent on the number of players, so `WC1p`, 'WC2p' and 'WC3p' would be nice. WC for all might become confusing when viewing save files.

knyghtmare:
Okay, as in alternate, I GitZipped the World Conquest dir from this branch "wc_difficulty" and dumped it all inside data/campaigns. It loads fine now, and it's all good. Would need a follow up for the icons/images but they can be added in later.

data/campaigns/World_Conquest/scenarios/WC_II_scenario.cfg
knyghtmare: According to the wiki:
> require_campaign: Whether clients are required to have this campaign installed beforehand to be allowed join a game using this campaign. Possible values 'yes' (the default) and 'no'.

I think, leaving it as `require_campaign=yes` or just not mentioning it would be good. Since this is for mainline, it would default to `yes` anyways.

Issue #2325, opened 2017-12-26
a wml tag [query_location] that queries a location from the user


gfgtdf:
migrated from easycoding, no my own idea:

> * [query_location] - queries a location from the user
> ** variable
> ** [filter_location]
> ''Regarding [query_location]: probably a better name would be choose_location. This is used when a wml event is running and needs to get a target location hex from the user. For example, you right click on a catapult unit and choose the menu item "fire Catapult." Now a [message] says to click on any target within 5 hexes. Now [choose_location] is encountered and the mouse cursor changes to a target sign (possibly customizable). If the user's mouse drifts outside of the allowed locations it changes to a cancel icon. If the operation is cancelled then the variable is cleared and no location stored. --Sapient''
>
>''This can be implemented in lua by using wesnoth.game_events.on_mouse_action and wesnoth.syncronize_choice''

CelticMinstrel:
Wait, we actually have the ability to do this now within the Lua API?

gfgtdf:
we should, we have wesnoth.game_events.on_mouse_action whihc basically fires whenever a mode is slicked anywhere (similar to select events, but also works on on-unit hexed) and sync_choice to send it to others.

ProditorMagnus:
wesnoth.game_events.on_mouse_action has not been documented anywhere. Trying to find out how it works.

ProditorMagnus:
Seems it is fired on leftclick on hex with 1.13.x+.

gfgtdf:
> wesnoth.game_events.on_mouse_action has not been documented anywhere. Trying to find out how it works.

it migth be buggy though, in paricular it'S current implementation might nreak if a user reassings the mouse1/mouse2 hotkeys in the hotkey menu.

ProditorMagnus:
Adding extra select unit hotkey (to mouse1) makes it so that both ways fire this.

gfgtdf:
ok i was bored so i made a sample implementation

```
function wesnoth.wml_actions.query_location(cfg)
	local res = wesnoth.synchronize_choice("location", function()
		local old_callback = wesnoth.game_events.on_mouse_action
		local xy = nil

		function wesnoth.game_events.on_mouse_action(x,y)

			if not xy then
				xy = { x = x, y = y }
			end

			if old_callback then
				--TODO: do we really want this?
				old_callback(x ,y)
			end
		end

		while not xy do
			while not xy do
				wesnoth.delay(10)
			end
			
[...]

PR #5892, opened 2021-06-24
Try to fix the issue with wesnoth.game_events.on_mouse_action

CelticMinstrel:
As described on the wiki, the action can trigger on the wrong hex if the mouse has moved during the callback.

In addition to directly fixing this, I've added a way for the callback itself to have some control over what happens, such as cancelling the action entirely or providing an alternate hex of its choosing.


src/scripting/game_lua_kernel.cpp
```diff
@@ -4570,6 +4570,8 @@ game_lua_kernel::game_lua_kernel(game_state & gs, play_controller & pc, reports
 	lua_setfield(L, -2, "__index");
 	lua_pushcfunction(L, &dispatch<&game_lua_kernel::impl_theme_items_set>);
 	lua_setfield(L, -2, "__newindex");
+	lua_pushliteral(L, "game display");
+	lua_setfield(L, -2, "__metatable");
```
CelticMinstrel, line 4574: This wasn't supposed to be in this commit…

CelticMinstrel:
So, any comments on this? I'd like to merge it soon(ish).

Pentarctagon:
I feel like I remember gfgtdf merging something related to this same issue? I can't seem to find it though so maybe I'm imagining things.

src/scripting/game_lua_kernel.cpp
```diff
@@ -5183,15 +5183,16 @@ void game_lua_kernel::mouse_over_hex_callback(const map_location& loc)
...
 	if (!luaW_getglobal(L, "wesnoth", "game_events", "on_mouse_action")) {
 		return;
 	}
+	int ret = lua_gettop(L);
 	lua_push(L, loc.wml_x());
 	lua_push(L, loc.wml_y());
-	luaW_pcall(L, 2, 0, false);
-	return;
+	luaW_pcall(L, 2, 2, false);
+	luaW_tolocation(L, ret, loc);
```
stevecotton, line 5197: What is expected to have happened here - is the on_mouse_action event expected to modify the location in UMC WML / Lua code?

CelticMinstrel:
> I feel like I remember gfgtdf merging something related to this same issue?

I don't remember anything like that, though maybe there was an unmerged PR?

src/scripting/game_lua_kernel.cpp
stevecotton: What would be the usecase for that? I'm not seeing a reason why the game would have a better idea of where the user intended to click than the user themselves.

CelticMinstrel: Once the callback returns, the game will proceed with the standard effect of clicking the specified hex. By returning an invalid hex (eg -9,-9), this can be prevented.

As for returning a valid hex other than the one clicked, I can't think of anything concrete right now. I think there _is_ use-case, though – if there was some mechanic where two hexes are "paired", it might make sense to have a handler where you click one of them but the action is instead performed on the other.

stevecotton: Having it say that it's handled the click makes sense to me.

However, I don't see the purpose of "paired" hexes instead of using the existing options to select units and hexes.
Generally, I feel it would better to avoid having an extra edge-case to support in the engine.

CelticMinstrel: Returning a different location would simulate an actual click on that location, so the action would depend on what's on that location and the unit (if any) that's currently selected. I'm not quite sure whether a "paired" hexes mechanic would require an actual simulated click, though.

stevecotton: But what UX does that enable? The Lua code can easily find out what's on that other hex, it presumably chooses that other hex based on something that the Lua code knows about the scenario, so why doesn't the Lua just call whatever code it wants to directly?

The answers that you've given so far read to me as "we can add more complexity" without saying what benefits come from that complexity. When there are possible benefits but no concrete example of what would use them, I'd suggest putting a "decide whether to do this" task on the roadmap, along with encouraging the UMC authors to look at the roadmap and comment if they would use the potential features.

CelticMinstrel:
Okay, so for now I've force-pushed it, but I'm going to run some tests, possibly add stuff to the test scenario, and perhaps change the return type to boolean instead of location.

CelticMinstrel:
Assuming the CI passes, I think this is ready to merge now.

src/mouse_events.cpp
```diff
@@ -785,17 +787,23 @@ void mouse_handler::select_or_action(bool browse)
 	wb::future_map_if_active planned_unit_map;
 
 	if(game_lua_kernel* lk = pc_.gamestate().lua_kernel_.get()) {
-		lk->select_hex_callback(last_hex_);
+		lk->select_hex_callback(clicked_hex);
+		// The callback will typically modify clicked_hex, so make sure it's still valid
+		if(!pc_.get_map().on_board(clicked_hex)) return;
```
stevecotton, line 792: Please put the `return` on its own line. In the middle of a function like this, inside a conditional block, isn't an obvious place to look for a check-or-return statement.

src/mouse_events.cpp
```diff
@@ -785,17 +787,23 @@ void mouse_handler::select_or_action(bool browse)
...
 
 	if(clicked_u && (!selected_u || selected_u->side() != side_num_ ||
 	  (clicked_u->side() == side_num_ && clicked_u->id() != selected_u->id()))
 	) {
-		select_hex(last_hex_, false);
+		select_hex(clicked_hex, false);
 	} else {
+		// There's a small chance the last hex has changed since this function began.
+		// Make sure the move applies to the hex that was originally clicked!
+		// TODO: Will this cause problems with previous_hex_?
```
stevecotton, line 805: This TODO still needs investigation.

src/mouse_events.cpp
```diff
@@ -785,17 +787,23 @@ void mouse_handler::select_or_action(bool browse)
 	wb::future_map_if_active planned_unit_map;
 
 	if(game_lua_kernel* lk = pc_.gamestate().lua_kernel_.get()) {
-		lk->select_hex_callback(last_hex_);
+		lk->select_hex_callback(clicked_hex);
+		// The callback will typically modify clicked_hex, so make sure it's still valid
```
stevecotton, line 791: Comment needs an update.

src/mouse_events.cpp
CelticMinstrel: What a pain…

src/mouse_events.cpp
CelticMinstrel: In fact, it still does modify `clicked_hex`, because I took the quickest route to changing the return value of the callback – it just sets the location to an invalid one if the callback returns true.

I'll still modify it, because I'm not sure that's a "typical" use-case.

src/mouse_events.cpp
CelticMinstrel: I'm not sure how to test it, but the only place it's used seems to be in `mouse_handler::current_unit_attacks_from`…

stevecotton: So what does that mean? Could the user see that the red attack-from arrow is coming in from one direction, click OK to attack, and then have the unit attack from a different direction instead?

stevecotton:
There should be something in mainline that uses the feature of cancelling the action. While `data/scenario-test.cfg` is a bit of a dumping ground, it could go there if nowhere else fits.

src/mouse_events.cpp
CelticMinstrel: I'm not sure, but… that seems like something that could be tested…

CelticMinstrel:
A simple "do you really want to move here" should do the trick, I think. I'll add that.

src/mouse_events.cpp
CelticMinstrel: It looks like this can indeed happen. Tested with the following callback:

```lua
wesnoth.game_events.on_mouse_action = function()
    wesnoth.interface.delay(500)
end
```

Start an attack, move the mouse randomly, and the unit moves to the wrong hex and the attack never actually starts. Even if the wrong hex happens to still be next to the attacked unit, it doesn't attack.

stevecotton:
Thanks for adding it to the test scenario, but the main question for me is still [what UX does this enable?](#r805408552)

Overriding the left-mouse button breaks drag-and-move starting from the affected hex, and while I haven't played the game on iPhone, I believe that's a major input method on that platform. It doesn't prevent drag-and-move ending on the affected hex, so "do you really want to move here" is a broken example - it's also an example that can already be done via `[enter_hex]`, `[exit_hex]` and `[cancel_action]`.

For the test case, I'd prefer something that describes what should happen, so that a future tester or developer can get an easy answer to the question "is that doing what it's meant to do, has it been broken in the meantime". For example (had it affected the other mouse button instead), label the hex "Right-clicking on this shows a prompt", and then have the question be "Click Yes to show the normal right-click menu, or No to suppress it".

src/mouse_events.cpp
stevecotton: That seems like it needs to be fixed as part of the "action can trigger on the wrong hex" bug.

Pentarctagon:
I don't think the link in your above comment works?

stevecotton:
It's the last comment in the review on Feb 5th, but that's now marked as resolved (and it doesn't auto-expand when clicking the link).

CelticMinstrel:
[This link seems to work.](#discussion_r805408552)

I couldn't think of a concrete use-case, which is why I changed it to just a true/false return for now. If we think of a concrete use-case later, we can add the ability to return a location then.

CelticMinstrel:
I changed the test case to one that I feel is more representative, and I think it also demonstrates why cancelling the action is useful. If there are no further objections by the end of the week, I'll probably go ahead and merge this.

stevecotton:
Same question as before, what UX does this enable?
Is this a good way to implement that UX?
Working backwards from the example code, I don't see a feature that this is a good implementation of.

Assuming that the intended feature is #2325 "a wml tag [query_location] that queries a location from the user" then this is not a good implementation:
* The UMC needs to include Lua code with the implementation details.
* The engine does not know that the user is being asked to select a hex, meaning there isn't a possibility of adding a better UI in the engine.
* Any localisation needs to be in handled in the UMC.
* The user can save the game while being asked to select a hex; the engine doesn't know anything special is happening.
    - With the example code, the spying doesn't happen when the game is reloaded.
    - In practice that would probably mean that the UMC gets into an unexpected state, because an `[event]name=moveto` such as "step onto a one-shot rune, choose which hex to affect" is forced by this implementation to be seen by the engine as two separate events.
    - (Note: If you start `wesnoth -t` and then load a game twice, Wesnoth exits. That's unrelated to this, it's how the test mode and mid-game loading mechanism works.)
* Synchronisation in multiplayer should be part of `#2325`, which suggests the engine should know that both parts are one event.

This can't be used more generally than #2325 either. The implementation only allows one callback to be registered, and the example code deregisters the callback with `wesnoth.game_events.on_mouse_action = nil`. This means that a `[modification]` using this would be removed by the scenario code, which implies that it should only be used as a "capture the very next click after this event", thus limiting it to only being used for `[query_location]`'s UX.

I think this PR should be reduced to just the bugfix, without adding a new feature. I think it also needs to include a fix for https://github.com/wesnoth/wesnoth/pull/5892#pullrequestreview-889336828 too.

CelticMinstrel:
> I think it also needs to include a fix for [#5892 (review)](https://github.com/wesnoth/wesnoth/pull/5892#pullrequestreview-889336828) too.

That's covered by 6f6d08252a953add86b41c0b9f44aefaefa197c4.

> This can't be used more generally than #2325 either. The implementation only allows one callback to be registered, and the example code deregisters the callback with `wesnoth.game_events.on_mouse_action = nil`. This means that a `[modification]` using this would be removed by the scenario code, which implies that it should only be used as a "capture the very next click after this event", thus limiting it to only being used for `[query_location]`'s UX.

I'm not sure what you're trying to say here. If you need multiple callbacks for some reason, it's not hard to do that by either making the callback call its predecessor or making a wrapper callback that calls all your callbacks. If you need to capture several clicks, the callback can register a new callback instead of just deregistering itself (and "deregistering" in a real use-case would probably mean reregistering the predecessor callback, rather than assigning nil). There are any number of ways this can be used other than the sample code I added to the test scenario.

Also, none of your complaints in that post have anything to do with the new feature added in this pull request. You're complaining about the existing API, not the small new feature I've added to it. While most of your points are good ones, the long-term solution to them would be to design a new API; but I don't think that's a reason not to make the existing one a little better… unless you're offering to design and implement a better API in the 1.17 development cycle.

(There are other issues with the existing API besides what you mentioned in that post – it triggers when you click the minimap, it can't handle right-click or drag-and-drop, etc – but I don't think those are a reason not to add this small new feature either.)

src/mouse_events.cpp
```diff
@@ -785,17 +789,26 @@ void mouse_handler::select_or_action(bool browse)
...
 	  (clicked_u->side() == side_num_ && clicked_u->id() != selected_u->id()))
 	) {
-		select_hex(last_hex_, false);
+		select_hex(clicked_hex, false);
 	} else {
+		// There's a small chance the last hex has changed since this function began.
+		// Make sure the move applies to the hex that was originally clicked!
+		utils::guard_value clicked(last_hex_, clicked_hex);
+		utils::guard_value prev(previous_hex_, attack_hex);
+		utils::guard_value route(current_route_, saved_route);
```
stevecotton, line 811: Looks like `move_action` is so complex that the easiest way to handle this is to still call it, but override three of its side effects after the call. My instinct for this part of the codebase is that that's probably true, but it's also why I'm so adverse to additional complexity in this area.

It's going to take me a lot of time to review these three lines' interaction with the other side effects of `move_action`, I can't commit to it this week.

stevecotton (changes requested):
Same question as before, what UX does this enable?
Is this a good way to implement that UX?
The reason that I'm putting the "changes requested" flag on this PR is because it still lacks an answer to those questions, and so any cost of added complexity outweighs the benefits identified.

Which add-ons currently use `on_mouse_action`, or would use it with this additional feature? Of the ones that I have installed, none of them currently use it.

src/mouse_events.cpp
stevecotton: @Pentarctagon, thanks for the offer to help with the reviews. I feel that the way the unit selection and `previous_hex`, `clicked_hex`, `attack_hex`, `saved_route`, etc are used in the existing code is spaghetti, and that adding `utils::guard_value` around the call to `move_action` is making the spaghetti worse. Please could I offload to you the task of working out if this is a good change, and maybe working out if there's a long-term plan to improve this?

stevecotton:
@CelticMinstrel and I both seem to feel this is waiting on the other person to respond to the feature part. For my perspective, the answers to "what UX does this enable? Is this a good way to implement that UX?" is that we've identified a UX that should be fixed with a different API, but nothing that would benefit from the added complexity of the feature part of this PR. Therefore this PR should be reduced to a pure bugfix.

The Era of High Sorcery is the only identified user of on_mouse_action, and it uses a workaround of reselecting a unit after the on_mouse_action event. Therefore it doesn't benefit from flagging the event as handled by the callback.

Lots of add-ons might use a `[query_location]`, but `on_mouse_action` is not a good API for UMC authors, as illustrated by the complexity of EoHS' code.

If there's another UX that's in favour of adding the feature, then I think it's the PR author's task to identify that UX.

[Logs of IRC discussion, between 4am and 7am](https://irclogs.wesnoth.org/2022/03/%23wesnoth-dev.2022-03-20.log)

CelticMinstrel:
I would argue that the "new feature" is a part of the bugfix. I would've just changed it, but I thought that might cause compatibility issues, although I'd be willing to do that _instead_ of adding the return value check if you think that's better.

src/mouse_events.cpp
Pentarctagon: Would you be able to explain more what you mean by "Looks like move_action is so complex that the easiest way to handle this is to still call it, but override three of its side effects after the call"?

This doesn't seem like a result of complexity, but simply that no matter how complex it is or how long it takes to execute, there's a chance someone manages to click something else half way through.

Pentarctagon:
So I guess, barring an explanation for https://github.com/wesnoth/wesnoth/pull/5892#pullrequestreview-906770514, that bit seems fine to me?

src/mouse_events.cpp
```diff
@@ -785,17 +789,26 @@ void mouse_handler::select_or_action(bool browse)
...
+			return;
+		}
 	}
 
-	unit_map::iterator clicked_u = find_unit(last_hex_);
+	unit_map::iterator clicked_u = find_unit(clicked_hex);
 	unit_map::iterator selected_u = find_unit(selected_hex_);
 
 	if(clicked_u && (!selected_u || selected_u->side() != side_num_ ||
 	  (clicked_u->side() == side_num_ && clicked_u->id() != selected_u->id()))
```
stevecotton, line 803: What is this `if` meant to check? I think some of the intended logic would check whether both units are on the same side, but the implemented logic is checking if (they're on the same side and also that side is the currently active side).

I'd speculate that should be checking if they're enemies of each other instead, so that attack predictions can be shown.

src/mouse_events.cpp
stevecotton: Sorry for taking so long to respond to this:

I'm not complaining about the "make sure the move applies to the hex that was originally clicked", I'm complaining that:
* The three values overwritten before calling `move_action` are a subset of the ones that control `move_action`.
* After `move_action` returns, a subset of the values that it changes are overwritten by the destructors of `utils::guard_value`.
* My gut feeling is that the new code isn't covering all edge-cases of the existing code.

@CelticMinstrel, why is `guard_value` used here? If `move_action` changes the values, why should this overwrite them after it returns? Why are these three specific member variables overwritten, but no others?

Walking through `move_action`:
* There are a lot of edge-cases, testing is unlikely to even cover all of the ones that can happen without the whiteboard.
* There's logic using `selected_hex_`, which may have changed since the start of `select_or_action` because `mouse_motion` can change it. In the current PR's code there's some checks about which team the unit (if any) on `selected_hex_` is on, but that still leaves a wide possibility for `selected_hex_` to have changed since the start of `select_or_action`.
* There's logic involving `last_hex_` and `selected_hex_`, where exactly one of those two has been overwritten in `select_or_action` to have the value that it had when `select_or_action` was called.
* Several paths through `move_action` call `select_hex`, a function which can change many of the class' member variables, a superset of the ones that will be restored by the `guard_value` destructors.

stevecotton:
Still waiting on @CelticMinstrel to identify a UX for the feature.

Pentarctagon:
If one of the main blockers for getting this merged is deciding whether there's a reason to let actions be canceled, maybe that can be split out to a separate PR so the rest of this can get merged separately.

Pentarctagon:
> I would argue that the "new feature" is a part of the bugfix. I would've just changed it, but I thought that might cause compatibility issues, although I'd be willing to do that instead of adding the return value check if you think that's better.

What would the compatibility issues be? At least based on what stevecotton said, there don't appear to be any add-ons changing it would affect, so breaking compatibility here wouldn't have much of an impact.

CelticMinstrel:
The compatibility issues would be that if something relied on the old behaviour, it would stop working. The comment you're quoting is from before we checked if add-ons actually use the API.

src/mouse_events.cpp
CelticMinstrel: This may be a good question but it's not really in scope for this PR…

Pentarctagon:
Ah, my bad then.

So to try to summarize things, the remaining questions are:
* Why would UMC authors want to be able to cancel a click action?
* https://github.com/wesnoth/wesnoth/pull/5892#discussion_r871798906
Correct?

src/mouse_events.cpp
CelticMinstrel: > Why are these three specific member variables overwritten, but no others?

They are the ones I identified as being necessary to overwrite. If I need to add another one, of course I can do that.

CelticMinstrel:
> * Why would UMC authors want to be able to cancel a click action?

Generally speaking, if a UMC author is using this API, it's to make something special happen when you click, in place of selecting/deselecting a unit or setting up a move or attack action. So, cancelling the action should logically have been the default behaviour, although there could also be reason for not cancelling it too (perhaps you're just monitoring what the player is doing or something?).

src/mouse_events.cpp
stevecotton: Please share your knowledge about why that's the necessary set - that knowledge would be really helpful for reviewing.

stevecotton:
> Generally speaking, if a UMC author is using this API, it's to make something special happen when you click, in place of selecting/deselecting a unit or setting up a move or attack action. So, cancelling the action should logically have been the default behaviour, although there could also be reason for not cancelling it too (perhaps you're just monitoring what the player is doing or something?).

This doesn't address the points in https://github.com/wesnoth/wesnoth/pull/5892#issuecomment-1079417919 , which I assume @Pentarctagon's one-line summary was referring to.

src/mouse_events.cpp
CelticMinstrel: From what I could tell, these are the variables that determine what action is taken as a result of the click. Under normal circumstances, they'd never change during the execution of this function, but if the callback takes some time to execute, they _might_ change… which was what this was supposed to fix.

CelticMinstrel:
I don't see how that makes any sense. I answered Pentarctagon's question, which doesn't seem to have any relation to the post you linked.

Pentarctagon:
I was referring to https://github.com/wesnoth/wesnoth/pull/5892#issuecomment-1125411418

src/mouse_events.cpp
stevecotton: My analysis on May 12th shows that your set is incomplete, there are more variables that determine what action is taken as a result of the click. However, that doesn't mean my analysis is the full set, it's just sufficient to review and say that you, as PR author, need to do a complete analysis.

You've also only responded to a single part of my questions from May 12th. If `move_action` changes the values, why should this overwrite them after it returns?

CelticMinstrel:
Okay, sure, but you didn't really do a good job of indicating that… and anyway, I don't think that comment is relevant. This is essentially fixing a bug, and the new feature is backwards compatibility for the bugfix. What does user experience have to do with anything?

The main use for this API is to temporarily replace the mouse action, so cancelling it is the preferred behaviour, and EoHS is the primary example of this; the fact that it employs a workaround for the bug is not, in my opinion, a valid reason to not fix the bug.

Maybe there could be uses that don't want to replace the mouse action (monitoring what the player is doing, maybe? not sure this API would be the ideal way to implement that, but it's a possibility), so it makes sense to include backwards compatibility instead of just outright changing it. If preferred, I suppose I could just outright change it, but honestly I think it's better not to.

Basically, as far as I'm concerned, this is totally ready for merging, and unless someone identifies an actual bug in the implementation, I have no plans to adjust it further. I really don't understand what octalot wants from me… perfection or something? It's an improvement on what exists in 1.16, so what's the issue? Sure, the API is a bit strange, and we could design a much better API to cover the same use-case, but unless someone's proposing to do this _right now_, I don't think it's an argument against merging this.

src/mouse_events.cpp
```diff
@@ -785,17 +789,26 @@ void mouse_handler::select_or_action(bool browse)
 	wb::future_map_if_active planned_unit_map;
 
 	if(game_lua_kernel* lk = pc_.gamestate().lua_kernel_.get()) {
-		lk->select_hex_callback(last_hex_);
+		lk->select_hex_callback(clicked_hex);
+		// The callback may modify clicked_hex, so make sure it's still valid
```
Pentarctagon, line 793: This is no longer true, I believe?

src/mouse_events.cpp
Pentarctagon: Isn't the reason for that the whole reason this PR was opened in the first place? To fix the issue of the action triggering on the wrong hex if the player moved the mouse?

stevecotton:
While [the API documentation](https://wiki.wesnoth.org/LuaAPI/wesnoth/game_events#wesnoth.game_events.on_mouse_action) already has a lot of comments about this function, I think we should deprecate it and add this to them: "EoHS (Era of High Sorcery) uses this, but has its own wrapper to work around the problems, including (if possible) queuing the data from the callback and performing it later. When triggered from EoHS's dialogs, for example when summoning, the processing is queued and runs in the same context that showed the dialog. Instead of trying to copy that code, it's probably better to ask the dev team to create a better API for your usecase."

That should be added even if this PR is merged. The reason that I keep asking what UX this PR enables is that I don't believe it's an incremental improvement - I believe it offers zero benefit because the only caller (other than a trivial case in the testing scenario) already has a library of workarounds written around it.

CelticMinstrel:
> Instead of trying to copy that code, it's probably better to ask the dev team to create a better API for your usecase.

This doesn't strike me as a good thing to _ever_ say on the wiki.

Pentarctagon:
I suppose deprecating it because it's barely used and not worth the effort and/or risk to fix the current implementation would essentially accomplish the same thing.

"Deprecated indefinitely with no replacement" would be a much better way to word it though.

Pentarctagon:
@CelticMinstrel the specific wording aside, what's your opinion on instead marking this API as deprecated?

CelticMinstrel:
I'd prefer not to deprecate it without a replacement…

Pentarctagon:
Alright, well, it seems we're at an impasse here. So I think the best option (for now) is to mark this as deprecated indefinitely with no replacement, since it's apparently only used by one UMC which has its own suite of code to make it usable, making the impact of deprecating it very minimal. If someone comes back later and says that the current API is fine, just buggy, then we can always undeprecate it and revisit this PR.

soliton-:
As a side note: `on_mouse_action` is one of the events that can crash wesnoth, see #10061.

PR #5903, opened 2021-06-27
various animations

doofus-01:
whatever animations I get done today

- jinn attacks

PR #5905, opened 2021-06-28
Jinn and horse graphics

doofus-01:
This is to add #5903 to master.


CelticMinstrel:
I think this is against 1.16, not master?

doofus-01:
Yep, I was hoping there was some way to just apply the relevant changes, but it's not looking good.

CelticMinstrel:
You need to create a branch based on 1.16 and then cherry-pick the specific commits.

doofus-01:
This won't be the only thing I need to port to 1.17, and rather than struggling with git each time, I'll just save them for later as a batch.  I'll close this and open an issue (#5963)

PR #5911, opened 2021-06-29
Possible replacement for the depreciated AI_CONTROLLER

Adamant14:
The deprecated AI Controller is removed from all mainline campaign scenarios without any replacement. Is that really a good idea? I think it's not. There was a good reason for creating and adding the AI_CONTROLLER.  If one plays scenarios in which you are dependent on AI-controlled allies, then it can be quite useful if you as a player have the opportunity to at least protect the AI side leader from suicidal actions. Otherwise there is often nothing left but to play the scenario again from the start, which is annoying. Especially when this can easily be avoided with the help of this simple and self-explanatory tool from Beetlenaut.
It is simple, and it is not as complicated, and as inscrutable for the simple gamer as the old AI controller. And this simplicity is what defines it. Most of the time it doesn't take more than that. Just three very simple commands:
1. DO NOT commit suicide!
2. Go here!
3. Do what ever you want!

I suggest not to remove the AI controller without a replacement, but replacing it with [Beetlenauts AI_Ally_Control](https://forums.wesnoth.org/viewtopic.php?f=8&t=54533).


CelticMinstrel:
I don't think any of these commits contain the actual change in the PR title.

cooljeanius:
so yeah I find the AI controller useful and will be copying it manually into the WML folder for campaigns I maintain; agreed that something looks wrong with this PR, though...

knyghtmare:
@Adamant14 
I had this same issue due to fetching and pulling, spiraling out of control.
The solution is:
1. close this PR
2. make a new branch in your fork (which is identical to the current wesnoth's master branch)
3. add the Improved AI controller files by @beetlenaut inside the macros directory.
4. make a new pull request.

cooljeanius:
yeah please close this; since it's based on the wrong branch, it means we keep getting notifications for commits unrelated to the actual issue here

PR #5913, opened 2021-07-02
Revert "Updated a few places that didn't handle locations in the unified way"

stevecotton:
While it might be a good idea, the code as committed fails the unit tests.
Reverting these changes also fixes the whitespace errors.

This reverts commit 5a9c24c4e79d624e34aa475949a4aa49b1984322.

### DIFF
--- a/src/ai/lua/core.cpp
+++ b/src/ai/lua/core.cpp
@@ -275,7 +275,10 @@ static int ai_recruit(lua_State *L, bool exec)
 	const char *unit_name = luaL_checkstring(L, 1);
 	int side = get_readonly_context(L).get_side();
 	map_location where;
-	luaW_tolocation(L, 2, where);
+	if (!lua_isnoneornil(L, 2)) {
+		where.set_wml_x(lua_tonumber(L, 2));
+		where.set_wml_y(lua_tonumber(L, 3));
+	}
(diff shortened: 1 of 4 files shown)


stevecotton:
The Windows-only build failure is this, which points to e555e0681db1e85df6dcb42808984e15b167807d being broken too.
```
   D:\a\wesnoth\wesnoth\src\gui\dialogs\statistics_dialog.cpp(164,1): error C3493: 'shift' cannot be implicitly captured because no default capture mode has been specified [D:\a\wesnoth\wesnoth\src\wesnoth-client.vcxproj]
  D:\a\wesnoth\wesnoth\src\gui\dialogs\statistics_dialog.cpp(170,18): error C2064: term does not evaluate to a function taking 2 arguments [D:\a\wesnoth\wesnoth\src\wesnoth-client.vcxproj]
  D:\a\wesnoth\wesnoth\src\gui\dialogs\statistics_dialog.cpp(180,19): error C2064: term does not evaluate to a function taking 2 arguments [D:\a\wesnoth\wesnoth\src\wesnoth-client.vcxproj]
```

CelticMinstrel:
I'm already working on a fix, so this will cause conflicts with that…

PR #5915, opened 2021-07-03
Remove the 10ms delay before closing a window

stevecotton:
This delay is to stop the loop running continously, it isn't necessary to do
that when we're about to exit the loop anyway. I'm unsure whether this makes
any noticeable difference, but it seems reasonable to remove delays when
possible.

I'm tempted to change the enclosing loop to a `do ... while` instead of a `for`
that has an empty clause.


irydacea:
The change makes sense to me.

> I'm tempted to change the enclosing loop to a do ... while instead of a for
that has an empty clause.

I don't think the compiler cares about the difference and this isn't code that people look at very often.

PR #5918, opened 2021-07-03
Add a prompt to migrate data from previous versions of wesnoth.

Pentarctagon:
Currently migrated, if existing:
* Add-ons (redownloaded if present on new add-ons server)
* Preferences file
* Credentials file

---

Currently incompletely tested, but mostly seems to be working.  The current (I assume pretty terrible) prompt UI:
Screenshot from 2021-07-03 13-42-52

One thing I'm also unsure about is whether/how to have the prompt show up *after* the titlescreen is displayed rather than immediately before.

### DIFF
--- a/data/gui/window/migrate_version_selection.cfg
+++ b/data/gui/window/migrate_version_selection.cfg
@@ -1,0 +1,111 @@
+#textdomain wesnoth-lib
+###
+### Definition of the window to ask which Wesnoth version to migrate data from
+###
+
+[window]
+	id = "migrate_version_selection"
+	description = "Dialog to select a previous version of Wesnoth to migrate data from."
+
+	[resolution]
+		definition = "default"
+
+		[linked_group]
+			id = "version_group"
+			fixed_width = true
+		[/linked_group]
+
+		[linked_group]
+			id = "radio_options"
+			fixed_width = true
+		[/linked_group]
+
+		[tooltip]
+			id = "tooltip"
+		[/tooltip]
+
+		[helptip]
+			id = "tooltip"
+		[/helptip]
+
+		[grid]
+			[row]
+				[column]
+
+					[label]
+						definition = "default"
+						label = _ "Choose which version of Wesnoth to import your data from:"
+					[/label]
+
+				[/column]
+			[/row]
+			[row]
+				[column]
+					[listbox]
+						id = "versions_listbox"
+						definition = "default"
+						horizontal_scrollbar_mode = "never"
+
+						[list_definition]
+							[row]
+								[column]
+									vertical_grow = true
+									horizontal_grow = true
+									[toggle_panel]
+										definition = "default"
+										[grid]
+											[row]
+												[column]
+													grow_factor = 1
+													horizontal_alignment = "left"
+
+													[label]
+														id = "version_label"
+														definition = "default"
+													[/label]
+
+												[/column]
+											[/row]
+										[/grid]
+									[/toggle_panel]
+								[/column]
+							[/row]
+						[/list_definition]
+					[/listbox]
+				[/column]
+			[/row]
+			[row]
+				[column]
+					[grid]
+						[row]
+							[column]
+								grow_factor = 1
+								border = "all"
+
+								[button]
+									id = "ok"
+									definition = "default"
+
+									label = _ "OK"
+								[/button]
+
+							[/column]
+							[column]
+								grow_factor = 1
+								border = "all"
+
+								[button]
+									id = "cancel"
+									definition = "default"
+
+									label = _ "Skip"
+								[/button]
+
+							[/column]
+						[/row]
+					[/grid]
+				[/column]
+			[/row]
+        [/grid]
+    [/resolution]
+[/window]
--- a/src/filesystem.hpp
+++ b/src/filesystem.hpp
@@ -121,14 +121,13 @@ static const blacklist_pattern_list default_blacklist{
 };
 
 /**
- * Populates 'files' with all the files and
- * 'dirs' with all the directories in dir.
- * If files or dirs are nullptr they will not be used.
- *
- * mode: determines whether the entire path or just the filename is retrieved.
- * filter: determines if we skip images and sounds directories
- * reorder: triggers the special handling of _main.cfg and _final.cfg
- * checksum: can be used to store checksum info
+ * @param dir The directory to examine.
+ * @param[out] files The files in @a dir. Won't be used if nullptr.
+ * @param[out] dirs The directories in @a dir. Won't be used if nullptr.
+ * @param mode Determines whether the entire path or just the filename is retrieved.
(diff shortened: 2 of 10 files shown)


CelticMinstrel:
This UI seems like you can't decide between using radiobuttons and using a listbox… you should pick only one. (A dropdown menu, ie menu_button, could be another option.)

Probably your listbox's toggle_button just uses the wrong definition?

src/filesystem.hpp
```diff
@@ -121,14 +121,13 @@ static const blacklist_pattern_list default_blacklist{
...
- * filter: determines if we skip images and sounds directories
- * reorder: triggers the special handling of _main.cfg and _final.cfg
- * checksum: can be used to store checksum info
+ * @param dir The directory to examine.
+ * @param files The files in @a dir. Won't be used if nullptr.
+ * @param dirs The directories in @a dir. Won't be used if nullptr.
+ * @param mode Determines whether the entire path or just the filename is retrieved.
+ * @param filter Determines if we skip images and sounds directories.
+ * @param reorder Triggers the special handling of _main.cfg and _final.cfg.
+ * @param checksum Can be used to store checksum info.
```
CelticMinstrel, line 130: This updated doc comment seems to be simultaneously an upgrade and a downgrade. It's an upgrade because it uses proper doxygen syntax, but a downgraded because it loses the genera description.

Also, I think you should mark the output parameters as such.

```suggestion
 * Get a list of all files and/or directories in a given directory.
 * 
 * @param dir The directory to examine.
 * @param[out] files The files in @a dir. Won't be used if nullptr.
 * @param[out] dirs The directories in @a dir. Won't be used if nullptr.
 * @param mode Determines whether the entire path or just the filename is retrieved.
 * @param filter Determines if we skip images and sounds directories.
 * @param reorder Triggers the special handling of _main.cfg and _final.cfg.
 * @param[out] checksum Can be used to store checksum info.
```

src/filesystem.hpp
```diff
@@ -208,6 +207,13 @@ filesystem::scoped_istream istream_file(const std::string& fname, bool treat_fai
 filesystem::scoped_ostream ostream_file(const std::string& fname, std::ios_base::openmode mode = std::ios_base::binary, bool create_directory = true);
 /** Throws io_exception if an error occurs. */
 void write_file(const std::string& fname, const std::string& data);
+/**
+ * Read a file and then writes it back out.
```
CelticMinstrel, line 211: Do we really need to describe the implementation detail? I think it should be fine to just say "Copies a file" or similar.

src/serialization/string_utils.cpp
```diff
@@ -865,4 +865,18 @@ void ellipsis_truncate(std::string& str, const std::size_t size)
 	}
 }
 
+std::string replace_all(const std::string& src, const std::string& from, const std::string& to)
```
CelticMinstrel, line 868: I think there's a `boost::replace_all` or similar which may already do exactly what you need.

data/gui/window/migrate_version_selection.cfg
```diff
@@ -0,0 +1,122 @@
...
+														id = "version_label"
+														definition = "default"
+													[/label]
+
+												[/column]
+
+												[column]
+													grow_factor = 1
+													horizontal_alignment = "center"
+													[toggle_button]
```
CelticMinstrel, line 72: There's no reason to have a toggle button here… I'd just remove it altogether. If a radiobutton is desired it could be added to the toggle panel definition.

src/wesconfig.h
```diff
@@ -40,11 +40,12 @@
   #undef VERSION
 #endif
 
-#define VERSION "1.15.14+dev"
+// TODO: undo changes to this file before merge
```
CelticMinstrel, line 43: Don't forget!

CelticMinstrel (commented):
Overall I think it looks good. Obviously the UI needs work but you already knew that.

Pentarctagon:
> This UI seems like you can't decide between using radiobuttons and using a listbox… you should pick only one. (A dropdown menu, ie menu_button, could be another option.)

My understanding is that to have a dynamic set of UI elements that are all visible at the same time, a listbox is needed so that rows can be added as needed to it.  The example I started with being log_settings.cpp and logging.cfg.

src/filesystem.hpp
Pentarctagon: The same information is present in the `@param` descriptions, unless I'm missing something.  I'll add [out] though.

src/filesystem.hpp
Pentarctagon: At that point there really isn't much of a reason to say even that - it doesn't tell the reader anything they wouldn't already have known from just the method name.

data/gui/window/migrate_version_selection.cfg
Pentarctagon: I don't understand - a toggle button is essentially a radio button/checkbox widget isn't it?

src/serialization/string_utils.cpp
Pentarctagon: Will update.

CelticMinstrel:
That _might_ be technically false but in practice, yes. The log settings use both because each log entry actually has options within itself; here however the list itself is the option, so you don't really need a button.

data/gui/window/migrate_version_selection.cfg
CelticMinstrel: Yes, but a radio button doesn't make sense in this context. The list itself is your choice. If you mean this to be a checkbox rather than a radio button (to migrate from potentially multiple version), then I suppose that could make sense, though I imagine making that actually work is quite a bit harder.

src/filesystem.hpp
CelticMinstrel: That's true…

src/filesystem.hpp
CelticMinstrel: I just think there should be a summary line on all functions. Maybe an exception can be made for extremely obvious cases like `copy_file`, but this definitely isn't one of those.

Pentarctagon:
Ah, alright, that makes sense.

src/wesconfig.h
stevecotton: I find it easier to remember if I put changes like this in a separate commit. That would also make the other changes easier to cherry-pick to a master-based branch for testing.

src/filesystem.hpp
Pentarctagon: It's also relevant in case anyone tries to use it to copy a large file - if for whatever reason someone wants to a copy a large file, the current implementation will read the entire thing into memory before writing it out.

src/wesconfig.h
stevecotton: /me prods @Pentarctagon with this "put changes like this in a separate commit" idea

PR #5921, opened 2021-07-04
Add a prompt to migrate data from previous versions of wesnoth.

Pentarctagon:
Currently migrated, if existing:
* Add-ons (redownloaded if present on new add-ons server)
* Preferences file
* Credentials file

---

Re-PR of #5918 so it's on a mainline branch.

Note to self: make sure to review the comments there as well.

### DIFF
--- a/data/gui/window/migrate_version_selection.cfg
+++ b/data/gui/window/migrate_version_selection.cfg
@@ -1,0 +1,137 @@
+#textdomain wesnoth-lib
+###
+### Definition of the window to ask which Wesnoth version to migrate data from
+###
+
+[window]
+	id = "migrate_version_selection"
+	description = "Dialog to select a previous version of Wesnoth to migrate data from."
+
+	[resolution]
+		definition = "default"
+
+		[linked_group]
+			id = "version_group"
+			fixed_width = true
+		[/linked_group]
+
+		[linked_group]
+			id = "radio_options"
+			fixed_width = true
+		[/linked_group]
+
+		[tooltip]
+			id = "tooltip"
+		[/tooltip]
+
+		[helptip]
+			id = "tooltip"
+		[/helptip]
+
+		[grid]
+			[row]
+				[column]
+					horizontal_alignment = "left"
+					border = "all"
+					border_size = 5
+
+					[label]
+						definition = "title"
+						label = _ "Data Migration"
+					[/label]
+
+				[/column]
+			[/row]
+			[row]
+				[column]
+					horizontal_alignment = "left"
+					border = "all"
+					border_size = 5
+
+					[label]
+						definition = "default"
+						label = _ "Choose which version of Wesnoth to import your data from.
+Some settings will not take effect until Wesnoth is restarted."
+					[/label]
+
+				[/column]
+			[/row]
+			[row]
+				[column]
+					horizontal_grow = true
+					border = "all"
+					border_size = 5
+					
+					[listbox]
+						id = "versions_listbox"
+						definition = "default"
+						horizontal_scrollbar_mode = "never"
+
+						[list_definition]
+							[row]
+								[column]
+									vertical_grow = true
+									horizontal_grow = true
+									[toggle_panel]
+										definition = "default"
+										[grid]
+											[row]
+												[column]
+													grow_factor = 1
+													horizontal_alignment = "left"
+													border = "all"
+													border_size = 5
+
+													[label]
+														id = "version_label"
+														definition = "default"
+													[/label]
+
+												[/column]
+											[/row]
+										[/grid]
+									[/toggle_panel]
+								[/column]
+							[/row]
+						[/list_definition]
+					[/listbox]
+				[/column]
+			[/row]
+			[row]
+				[column]
+					horizontal_alignment = "right"
+					[grid]
+						[row]
+							[column]
+								grow_factor = 1
+								border = "all"
+								border_size = 5
+
+								[button]
+									id = "ok"
+									definition = "default"
+
+									label = _ "OK"
+								[/button]
+
+							[/column]
+							[column]
+								grow_factor = 1
+								border = "all"
+								border_size = 5
+
+								[button]
+									id = "cancel"
+									definition = "default"
+
+									label = _ "Skip"
+								[/button]
+
+							[/column]
+						[/row]
+					[/grid]
+				[/column]
+			[/row]
+		[/grid]
+	[/resolution]
+[/window]
--- a/src/addon/manager_ui.cpp
+++ b/src/addon/manager_ui.cpp
@@ -270,18 +270,29 @@ bool ad_hoc_addon_fetch_session(const std::vector<std::string>& addon_ids)
 		}
 
 		bool return_value = true;
-		for(const std::string & addon_id : addon_ids) {
+		std::ostringstream os;
+		for(const std::string& addon_id : addon_ids) {
 			addons_list::const_iterator it = addons.find(addon_id);
 			if(it != addons.end()) {
 				const addon_info& addon = it->second;
-				addons_client::install_result res = client.install_addon_with_checks(addons, addon);
-				return_value = return_value && (res.outcome == addons_client::install_outcome::success);
+				// don't redownload in case it was already downloaded for being another add-on's dependency
+				if(!filesystem::file_exists(filesystem::get_addons_dir()+"/"+addon_id)) {
+					addons_client::install_result res = client.install_addon_with_checks(addons, addon);
+					return_value = return_value && (res.outcome == addons_client::install_outcome::success);
+				}
 			} else {
-				utils::string_map symbols;
-				symbols["addon_id"] = addon_id;
-				gui2::show_error_message(VGETTEXT("Could not find an add-on matching id $addon_id on the add-on server.", symbols));
+				if(!return_value) {
+					os << ", ";
+				}
+				os << addon_id;
 				return_value = false;
 			}
+		}
+
+		if(!return_value) {
+			utils::string_map symbols;
+			symbols["addon_ids"] = os.str();
+			gui2::show_error_message(VGETTEXT("Could not find add-ons matching the ids $addon_ids on the add-on server.", symbols));
 		}
 
 		return return_value;
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -23,7 +23,6 @@
 #include "config.hpp"
 #include "deprecation.hpp"
 #include "game_config.hpp"
-#include "game_version.hpp"
 #include "gettext.hpp"
 #include "log.hpp"
 #include "serialization/string_utils.hpp"
@@ -510,14 +509,14 @@ std::string get_next_filename(const std::string& name, const std::string& extens
 
 static bfs::path user_data_dir, user_config_dir, cache_dir;
 
-static const std::string get_version_path_suffix(const version_info& version)
+const std::string get_version_path_suffix(const version_info& version)
 {
 	std::ostringstream s;
 	s << version.major_version() << '.' << version.minor_version();
 	return s.str();
 }
 
-static const std::string& get_version_path_suffix()
+const std::string& get_version_path_suffix()
 {
 	static std::string suffix;
 
@@ -568,6 +567,9 @@ static void setup_user_data_dir()
 #if defined(__APPLE__) && !defined(__IPHONEOS__)
 	migrate_apple_config_directory_for_unsandboxed_builds();
 #endif
+	if(!file_exists(user_data_dir)) {
+		game_config::check_migration = true;
+	}
 
 	if(!create_directory_if_missing_recursive(user_data_dir)) {
 		ERR_FS << "could not open or create user data directory at " << user_data_dir.string() << '\n';
@@ -1078,6 +1080,11 @@ void write_file(const std::string& fname, const std::string& data)
 			throw io_exception("Error writing to file: '" + fname + "'");
 		}
 	}
+}
+
+void copy_file(const std::string& src, const std::string& dest)
+{
+	write_file(dest, read_file(src));
 }
 
 bool create_directory_if_missing(const std::string& dirname)
--- a/src/filesystem.hpp
+++ b/src/filesystem.hpp
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "exceptions.hpp"
+#include "game_version.hpp"
 
 class config;
 class game_config_view;
@@ -121,14 +122,15 @@ static const blacklist_pattern_list default_blacklist{
 };
 
 /**
- * Populates 'files' with all the files and
- * 'dirs' with all the directories in dir.
- * If files or dirs are nullptr they will not be used.
+ * Get a list of all files and/or directories in a given directory.
  *
- * mode: determines whether the entire path or just the filename is retrieved.
- * filter: determines if we skip images and sounds directories
- * reorder: triggers the special handling of _main.cfg and _final.cfg
- * checksum: can be used to store checksum info
+ * @param dir The directory to examine.
+ * @param[out] files The files in @a dir. Won't be used if nullptr.
+ * @param[out] dirs The directories in @a dir. Won't be used if nullptr.
+ * @param mode Determines whether the entire path or just the filename is retrieved.
+ * @param filter Determines if we skip images and sounds directories.
+ * @param reorder Triggers the special handling of _main.cfg and _final.cfg.
+ * @param[out] checksum Can be used to store checksum info.
  */
 void get_files_in_dir(const std::string &dir,
                       std::vector<std::string>* files,
@@ -149,6 +151,8 @@ std::string get_saves_dir();
 std::string get_intl_dir();
 std::string get_screenshot_dir();
 std::string get_addons_dir();
+const std::string get_version_path_suffix(const version_info& version);
+const std::string& get_version_path_suffix();
 
 /**
  * Get the next free filename using "name + number (3 digits) + extension"
@@ -208,6 +212,13 @@ filesystem::scoped_istream istream_file(const std::string& fname, bool treat_fai
 filesystem::scoped_ostream ostream_file(const std::string& fname, std::ios_base::openmode mode = std::ios_base::binary, bool create_directory = true);
 /** Throws io_exception if an error occurs. */
 void write_file(const std::string& fname, const std::string& data);
+/**
+ * Read a file and then writes it back out.
+ *
+ * @param src The source file.
+ * @param dest The destination of the copied file.
+ */
+void copy_file(const std::string& src, const std::string& dest);
 
 std::string read_map(const std::string& name);
 
--- a/src/gui/dialogs/migrate_version_selection.cpp
+++ b/src/gui/dialogs/migrate_version_selection.cpp
@@ -1,0 +1,123 @@
+/*
+   Copyright (C) 2008 - 2021
+   Part of the Battle for Wesnoth Project https://www.wesnoth.org/
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY.
+
+   See the COPYING file for more details.
+*/
+
+#define GETTEXT_DOMAIN "wesnoth-lib"
+
+#include "gui/dialogs/migrate_version_selection.hpp"
+
+#include "addon/manager_ui.hpp"
+#include "filesystem.hpp"
+#include "game_version.hpp"
+#include "gettext.hpp"
+#include "gui/auxiliary/find_widget.hpp"
+#include "gui/dialogs/message.hpp"
+#include "gui/widgets/listbox.hpp"
+#include "gui/widgets/window.hpp"
+#include "preferences/credentials.hpp"
+#include "preferences/game.hpp"
+
+#include <boost/algorithm/string.hpp>
+
+static lg::log_domain log_version_migration{"gui/dialogs/migrate_version_selection"};
+#define ERR_LOG_VERSION_MIGRATION LOG_STREAM(err, log_version_migration)
+#define WRN_LOG_VERSION_MIGRATION LOG_STREAM(warn, log_version_migration)
+#define LOG_LOG_VERSION_MIGRATION LOG_STREAM(info, log_version_migration)
+#define DBG_LOG_VERSION_MIGRATION LOG_STREAM(debug, log_version_migration)
+
+namespace gui2::dialogs
+{
+REGISTER_DIALOG(migrate_version_selection)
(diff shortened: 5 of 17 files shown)


src/filesystem.hpp
```diff
@@ -121,14 +121,13 @@ static const blacklist_pattern_list default_blacklist{
...
 /**
- * Populates 'files' with all the files and
- * 'dirs' with all the directories in dir.
- * If files or dirs are nullptr they will not be used.
- *
- * mode: determines whether the entire path or just the filename is retrieved.
- * filter: determines if we skip images and sounds directories
- * reorder: triggers the special handling of _main.cfg and _final.cfg
- * checksum: can be used to store checksum info
+ * @param dir The directory to examine.
```
CelticMinstrel, line 124: I still think this doc comment should include a brief summary.

src/gui/dialogs/migrate_version_selection.cpp
```diff
@@ -0,0 +1,123 @@
...
+    if(mig.versions_.size() > 0)
+    {
+        mig.show();
+    }
+}
+
+migrate_version_selection::migrate_version_selection()
+{
+    version_info current_version = game_config::wesnoth_version;
+    std::string current_version_str = std::to_string(current_version.major_version())+"."+std::to_string(current_version.minor_version());
```
stevecotton, line 51: These lines get repeated several times in this file, and they duplicate code that's already in filesystem.cpp's (currently private) `get_version_path_suffix` functions.

src/gui/dialogs/migrate_version_selection.hpp
```diff
@@ -0,0 +1,43 @@
...
+public:
+    migrate_version_selection();
+    static void execute();
+
+private:
+	virtual void pre_show(window& window) override;
+	virtual void post_show(window& window) override;
+	virtual const std::string& window_id() const override;
+
+    std::vector<std::string> versions_;
```
stevecotton, line 40: Inconsistent whitespace.

src/gui/dialogs/migrate_version_selection.cpp
Pentarctagon: Huh, alright. I did look for something already existing like that, but only in `version_info` rather than in `filesystem`.

src/filesystem.hpp
Pentarctagon: So just something like `Loads a list of files and-or directories from @a dir`?

src/gui/dialogs/migrate_version_selection.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    for(unsigned int i = 1; i < current_version.minor_version(); i++)
+    {
+        std::string previous_version_str = std::to_string(current_version.major_version())+"."+std::to_string(current_version.minor_version()-i);
+        std::string previous_addons_dir = boost::replace_all_copy(filesystem::get_addons_dir(), current_version_str, previous_version_str);
+
+        if(filesystem::file_exists(previous_addons_dir))
```
stevecotton, line 58: Needs a sanity check that `previous_addons_dir != filesystem::get_addons_dir()`.

That can happen with the command-line options `--userconfig-dir` `--userdata-dir`. I can't remember if there's any platform where the default userconfig-dir doesn't have the version number in.

src/filesystem.hpp
Pentarctagon: Never mind, I see you had a general description on your previous comment.

src/gui/dialogs/migrate_version_selection.cpp
Pentarctagon: As a more general question, since I'm much more familiar with `static` in Java than C++: is there a reason that `filesystem.cpp` has essentially a bunch of private static functions in it?  I've also been running into various compiler/linker errors related to this, and examples I find online seem to all be based on static methods of a class, whereas here there is no "filesystem" class that these are attached to.

src/filesystem.hpp
CelticMinstrel: Either mine or yours sounds fine.

src/gui/dialogs/migrate_version_selection.cpp
CelticMinstrel: When used outside a class or function, `static` means something is local to the current file – you can't reference it from another file.

src/gui/dialogs/migrate_version_selection.cpp
```diff
@@ -0,0 +1,123 @@
+/*
+   Copyright (C) 2008 - 2018 by Jörg Hinrichs <joerg.hinrichs@alice-dsl.de>
```
stevecotton, line 2: This line is an old cut&paste.

src/gui/dialogs/migrate_version_selection.cpp
stevecotton: This use of `static` is the C-style equivalent of an putting the function in an anonymous namespace; I probably should have used the `namespace` keyword. When I added that code, I didn't consider that it might be used by the migration code (sorry).

src/gui/dialogs/migrate_version_selection.cpp
CelticMinstrel: I usually just delete the name and email and write "by The Battle for Wesnoth Project", essentially merging the two lines. Up to you though.

src/gui/dialogs/migrate_version_selection.cpp
Pentarctagon: Well, the migration code wasn't going to exist for quite a long time yet anyway.  As long as there isn't a reason to keep it static though, I'll make it non-static and accessible to the migration dialog code.

Pentarctagon:
Alright, I believe all the previous comments should have been addressed now.  So that leaves the remaining testing and then the UI updates.

@Vultraz would you prefer I try to make the UI better in this PR, or would you rather to just do it yourself?

Pentarctagon:
For the UI/UX, right now the adhoc connection displays a popup for each individual add-on that's not found, which I imagine could end up being quite a lot popups in this use case.  So would it be better to just not show a message for add-ons not found, or to try and consolidate all add-ons that aren't found into a list on a single popup?

CelticMinstrel:
Consolidating them seems essential for the user experience. How exactly is add-on migration handled, anyway? Does it download the new version from the new add-ons server if possible? If not, does this prompt you mentioned allow for copying it ad-hoc into the new add-ons directory?

Pentarctagon:
It uses the existing `ad_hoc_addon_fetch_session(...)` functionality which is currently used by the MP lobby to download missing add-ons when joining a game.  For that you provide a `std::vector<std::string>>` of the add-on IDs (in this case a listing of the folders in the old version's add-ons directory) - add-ons it finds on the new add-ons server are downloaded, while a popup saying it couldn't be found is displayed per add-on that isn't found.

Copying add-ons from the old version to the new versions seems inadvisable though, and will likely result in UMC authors getting error reports for things they haven't even ported yet.

CelticMinstrel:
I think copying the add-on over will work in at least 50% of cases, but if that option is supported we should definitely make it _very_ clear that the user is doing this "at their own risk", so to speak.

Is there a way to re-trigger the migration at a later time, so that add-ons that failed to migrate the first time might succeed the second time?

Pentarctagon:
Right now it only triggers if Wesnoth had to create brand new userdata directories when it launched.

Pentarctagon:
The last force push has fixed the UI to list out the add-on IDs all in a single popup rather than a single popup per add-on that wasn't found.

The preferences and credentials file copying is also tested and working... somewhat.  The actual copying of the file works, but for some reason it currently will correctly reload and persist only some of the preferences, but not all of them.  For example, the contents of any [server] tags are persisted, but not the contents of any [completed_campaigns] or [history] tags.  The `encountered_terrain_list` attribute also gets reset back to empty for some reason.  Once I get that figured out though, I think all that remains is the UI's appearance.

Additionally, many settings are not applied immediately, such as fullscreen (needs a restart) and sounds being enabled/disabled (requires a restart or F5 on the main menu).  I'm not sure if there's a good way to apply these immediately, though.

CelticMinstrel:
There's definitely a way. I don't know if it's a _good_ way though.

Pentarctagon:
Alright, looks like it's because different parts of the preferences are actually loaded in two places `preferences/general` and `preferences/game`.

CelticMinstrel:
I was kinda wondering… is there any reason to want to migrate from two different versions at the same time? Maybe it's just not worth the effort…

Pentarctagon:
I'd guess that there are people who may want to, but then you also need to deal with deciding which takes priority when there are identical add-ons and-or preferences set between them which is the part I don't think is worth the effort.

CelticMinstrel:
Yeah… that makes sense.

Pentarctagon:
Alright, it should all be good now, aside from the UI.  I've also added a popup after migrating is complete saying that some setting won't take effect until Wesnoth is restarted (ie: fullscreen vs non-fullscreen).

I'm going to investigate the credentials handler not being able to handle storing multiple users and if it's simple to fix I'll include it here, otherwise I'll make a separate PR for it.

Pentarctagon:
Multiple credentials actually seem to be working for me locally at least.  If I log into my local wesnothd as "p1" and then "p2", then restart Wesnoth, I'm no longer prompted for a password when I use either of those usernames.

Pentarctagon:
So yeah, this is ready for review now, I believe.

@Vultraz let me know if you'd rather review and give me instructions on how to change the UI, or if it'd be better if it's merged as-is and then you can fix it yourself.

data/gui/window/migrate_version_selection.cfg
```diff
@@ -0,0 +1,111 @@
...
+		[/tooltip]
+
+		[helptip]
+			id = "tooltip"
+		[/helptip]
+
+		[grid]
+			[row]
+				[column]
+
```
Vultraz, line 34: Add:
```
horizontal_alignment = "left"
border = "all"
border_size = 5
```

data/gui/window/migrate_version_selection.cfg
```diff
@@ -0,0 +1,111 @@
...
+					[label]
+						definition = "default"
+						label = _ "Choose which version of Wesnoth to import your data from:"
+					[/label]
+
+				[/column]
+			[/row]
+			[row]
+				[column]
+					[listbox]
```
Vultraz, line 44: Add
```
horizontal_grow = true
border = "all"
border_size = 5
```

data/gui/window/migrate_version_selection.cfg
```diff
@@ -0,0 +1,111 @@
...
+									vertical_grow = true
+									horizontal_grow = true
+									[toggle_panel]
+										definition = "default"
+										[grid]
+											[row]
+												[column]
+													grow_factor = 1
+													horizontal_alignment = "left"
+
```
Vultraz, line 61: Add
```
border = "all"
border_size = 5
```

data/gui/window/migrate_version_selection.cfg
```diff
@@ -0,0 +1,111 @@
...
+				[/column]
+			[/row]
+			[row]
+				[column]
+					[grid]
+						[row]
+							[column]
+								grow_factor = 1
+								border = "all"
+
```
Vultraz, line 84: Add
```
border_size = 5
```

data/gui/window/migrate_version_selection.cfg
```diff
@@ -0,0 +1,111 @@
...
+									definition = "default"
+
+									label = _ "OK"
+								[/button]
+
+							[/column]
+							[column]
+								grow_factor = 1
+								border = "all"
+
```
Vultraz, line 96: Add
```
border_size = 5
```

data/gui/window/migrate_version_selection.cfg
```diff
@@ -0,0 +1,111 @@
...
+										[/grid]
+									[/toggle_panel]
+								[/column]
+							[/row]
+						[/list_definition]
+					[/listbox]
+				[/column]
+			[/row]
+			[row]
+				[column]
```
Vultraz, line 78: Add
```
horizontal_alignment = "right"
```

src/gui/dialogs/migrate_version_selection.cpp
```diff
@@ -0,0 +1,126 @@
...
+            versions_.push_back(previous_version_str);
+        }
+    }
+}
+
+void migrate_version_selection::pre_show(window& window)
+{
+    listbox& version_list = find_widget<listbox>(&window, "versions_listbox", false);
+
+    for(unsigned int i = 0; i < versions_.size(); i++)
```
Vultraz, line 71: Please use a range-for loop `for(const std::string& ver : versions_)`

src/gui/dialogs/migrate_version_selection.cpp
```diff
@@ -0,0 +1,126 @@
...
+        item_label["label"] = version;
+        data["version_label"] = item_label;
+
+        version_list.add_row(data);
+    }
+}
+
+void migrate_version_selection::post_show(window& window)
+{
+    if(this->get_retval() == gui2::OK)
```
Vultraz, line 86: No need for `this`.

src/gui/dialogs/migrate_version_selection.hpp
```diff
@@ -0,0 +1,43 @@
...
+*/
+
+#pragma once
+
+#include "gui/dialogs/modal_dialog.hpp"
+
+namespace gui2
+{
+
+namespace dialogs
```
Vultraz, line 22: Can collapse these namespace declarations (`namespace gu2::dialogs {}`).

src/wesnoth.cpp
```diff
@@ -779,6 +780,14 @@ static int do_gameloop(const std::vector<std::string>& args)
 
 	game_config_manager config_manager(cmdline_opts);
 
+	if(game_config::check_migration) {
+		game_config::check_migration = false;
+		gui2::dialogs::migrate_version_selection::execute();
+		gui2::show_message("",
```
Vultraz, line 786: Probably better to put this in the prompt itself, especially since there's a case where the dialog may not show at all.

src/gui/dialogs/migrate_version_selection.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+#define GETTEXT_DOMAIN "wesnoth-lib"
+
+#include "gui/dialogs/migrate_version_selection.hpp"
+
+#include "addon/manager_ui.hpp"
+#include "filesystem.hpp"
+#include "game_version.hpp"
+#include "gui/auxiliary/find_widget.hpp"
+#include "gui/dialogs/migrate_version_selection.hpp"
```
Vultraz, line 23: Duplicate include, remove.

Vultraz (changes requested):

src/filesystem.cpp
```diff
@@ -510,14 +509,14 @@ std::string get_next_filename(const std::string& name, const std::string& extens
...
-static const std::string get_version_path_suffix(const version_info& version)
+std::string get_version_path_suffix(const version_info& version)
 {
 	std::ostringstream s;
 	s << version.major_version() << '.' << version.minor_version();
 	return s.str();
 }
 
-static const std::string& get_version_path_suffix()
+std::string& get_version_path_suffix()
```
stevecotton, line 519: Either this needs to be const, or it needs to lose the behavior of caching a static string.

src/gui/dialogs/migrate_version_selection.cpp
```diff
@@ -0,0 +1,130 @@
+/*
+   Copyright (C) 2008 - 2018
```
stevecotton, line 2: Either drop the years entirely, or put 2021 in here.

src/gui/dialogs/migrate_version_selection.cpp
```diff
@@ -0,0 +1,130 @@
...
+        mig.show();
+    }
+}
+
+migrate_version_selection::migrate_version_selection()
+{
+    version_info current_version = game_config::wesnoth_version;
+    std::string current_version_str = filesystem::get_version_path_suffix();
+
+    for(unsigned int i = 1; i < current_version.minor_version(); i++)
```
stevecotton, line 56: As it's a new file, please run it through clang-format.

src/addon/manager_ui.cpp
```diff
@@ -270,20 +270,28 @@ bool ad_hoc_addon_fetch_session(const std::vector<std::string>& addon_ids)
...
+				}
+				os << addon_id;
 				return_value = false;
 			}
 		}
 
+		if(!return_value) {
+			utils::string_map symbols;
+			symbols["addon_id"] = os.str();
+			gui2::show_error_message(VGETTEXT("Could not find an add-on matching the ids $addon_id on the add-on server.", symbols));
```
stevecotton, line 292: There's a mix of singular and plural here.

stevecotton:
If an add-on has dependencies, there's a prompt about whether to download them even though they should be automatically installed (as the dependency was itself installed in the old version).

src/gui/dialogs/migrate_version_selection.cpp
CelticMinstrel: If you do that though, do a quick manual check of the file to make sure it didn't do anything stupid to the formatting.

Pentarctagon:
I'm not sure I follow the scenario you're describing.  Are you saying that IftU was migrated, then when migrating something else that listed IftU as a dependency you were prompted to download it again?

stevecotton:
Starting with these in the old version's add-ons folder:

```
After_the_Storm/  Archaic_Era/	AtS_Music/  Era_of_Magic/  Era_of_Magic_Resources/  IftU_Music/ Invasion_from_the_Unknown/

$ grep 'dependencies' data/add-ons/*/_info.cfg 
data/add-ons/After_the_Storm/_info.cfg:	dependencies="AtS_Music,IftU_Music"
data/add-ons/Archaic_Era/_info.cfg:	dependencies=""
data/add-ons/AtS_Music/_info.cfg:	dependencies="After_the_Storm"
data/add-ons/Era_of_Magic/_info.cfg:	dependencies="Era_of_Magic_Resources"
data/add-ons/Era_of_Magic_Resources/_info.cfg:	dependencies=""
data/add-ons/IftU_Music/_info.cfg:	dependencies="Invasion_from_the_Unknown"
data/add-ons/Invasion_from_the_Unknown/_info.cfg:	dependencies="IftU_Music"
```

* Start a migration
* An "Install Dependencies" dialog asks to install "After the Storm Music", "Invasion from the Unknown" and "Invasion from the Unknown Music".
* Click "Yes"
* Download of AtS Music
* Download of IftU
* Download of IftU Music
* Download of AtS
* Download of Archaic Era
* An "Install Dependencies" dialog asks to install "After the Storm", "Invasion from the Unknown" and "Invasion from the Unknown Music". Each has the yellow text "Installed, upgradable" under the name, and the version fields all display "0.0.0" as the current version.
* Click "Yes"
* Download of AtS
* Download of IftU
* Download of IftU Music
* Download of AtS Music
* An "Install Dependencies" dialog asks to install "Era of Magic - Resources"
* Click "Yes"
* Download of EoM Resources
* Download of EoF
* Download of EoM Resources (yes, again)
* An "Install Dependencies" dialog asks to install "Invasion from the Unknown". It has the yellow text "Installed, upgradable" under the name, and the version field displays "0.0.0" as the current version.
* Click "Yes"
* Download of IftU
* Download of IftU Music
* An "Install Dependencies" dialog asks to install "Invasion from the Unknown Music". It has the yellow text "Installed, upgradable" under the name, and the version field displays "0.0.0" as the current version.
* Click "Yes"
* Download of IftU Music
* Download of IftU

Pentarctagon:
It sounds like the solution for this case is for it to skip any add-ons that already exist in the 1.16 add-ons directory since here we already know that anything present is the latest version.

src/gui/dialogs/migrate_version_selection.cpp
Pentarctagon: I'm not sure why it's doing any of the things it's currently configured to do anyway, so I'm not sure what would be a case of it doing something we don't want it to do.

CelticMinstrel: I'm not sure either, but it can happen. There will always be cases where clang-format gets things wrong, either because it's a case we didn't think of and thus didn't explicitly configure correctly, or because it's an unusual case where we actually _want_ non-standard formatting.

Automated formatting is fine and all, but blinding committing what it did without even double-checking is not a good idea.

src/filesystem.cpp

src/gui/dialogs/migrate_version_selection.cpp
Pentarctagon: Updated.  The years from my understanding are technically required, unfortunately, otherwise it'd be nice to not have them there at all.

src/addon/manager_ui.cpp
Pentarctagon: Updated.

Pentarctagon:
I added a check to not treat add-ons downloaded but not refreshed in the version cache as needing to be upgraded, and another check so that if an add-on gets downloaded as a dependency before it gets downloaded from the list of add-on IDs it doesn't get downloaded twice.

src/gui/dialogs/migrate_version_selection.cpp
Pentarctagon: Also, is the current .clang-format even something that we want applied without discussing what the format should be?  It exists, but clearly hasn't been updated in quite a while (it still lists the C++ language version as C++11), and I have no idea currently what changes it's set to make or if they're actually desirable.

Pentarctagon: I suppose also that it doesn't really make a lot of sense to me to only apply clang-format to one file at a time.  Code formatting isn't really something that consistent across the rest of the codebase anyway, so only applying here and there occasionally feels like it'll just end up as an instance of https://xkcd.com/927/

stevecotton: A new file is code that wasn't there before, so running the code formatter on it doesn't add extra changes to the history. Running the code formatter on existing files does generate extra changes.

The main thing I noticed in this file was the placement of opening braces being inconsistent with other Wesnoth files. Our format may be a bit odd, but the current codebase is consistent enough to notice.

Pentarctagon: Updated, I suppose.  I'm not a fan of how it adjusts comments especially though, and the line length limit generally.

stevecotton:
The dependency handling is much better, thanks. It would be even better if it didn't prompt when the old version already had the dependency installed (so it's already in the list of add-ons to add).

For the prompt about which version to add, how about changing the dialog to just a column of buttons?
* `[Import settings and add-ons from 1.15]`
* `[Import settings and add-ons from 1.14]`
* `[Start with a clean slate]`

Pentarctagon:
Do that would require adding some special handling for this specific case, which I don't really want to do.

IMO having buttons doesn't really improve it, and I also don't want to add any more new strings either.

PR #5922, opened 2021-07-04
Fix untranslatable string in advancefrom deprecated message

fujimo-t:
I noticed that second chunk is not marked to translatable while working 1.16 translation.

The target branch is set to master because I heard 1.16 is string freezed.
Is it possible to merge this also 1.16 and re-run pot update?


stevecotton:
I think this one shouldn't be translatable at all, as it's an error that will only make sense to WML authors. @ivanovic?

CelticMinstrel:
Our generic "X has been deprecated" message is translatable, so I think it can make sense to make "use Y instead" translatable too… but probably it should be a single string, like "use $elem instead" and substitute in the replacement?

That doesn't _quite_ work here, mind you, but other deprecation messages have an untranslatable portion with additional details, so perhaps the same could be done here.

```cpp
interpolate_into_string(
  _("Use $elem instead."),
  {"elem", "[modify_unit_type]\n" + modify_unit_type.debug() + "\n [/modify_unit_type]"
) + "This goes in the [campaign] or [multiplayer] tag rather than in the [unit_type]."
```

However, either way it's probably something that has to wait for 1.16.1?

fujimo-t:
Indeed, using substitution looks to better and this untranslatable thing is not important for most players.

Wedge009:
This seems to be hanging around unnecessarily, given it's a small change. If waiting for 1.16.1, it can just be added to the back-port list, no?

Does fujimo need help adapting celmin's suggestion?

fujimo-t:
I'm sorry for late reply.
I had left GitHub for a while.

I will try to fix by string interpolation.

stevecotton:
This message never triggers, the code to trigger it is broken and looks for `[units][advancefrom]` instead of `[unit_type][advancefrom]`. This is issue #6204.

If we're going to make it a translatable player-readable string, I suggest phrasing it as `Error: [advancefrom] does not work, "Cuttle Fish" units will not be able to advance to "Kraken"`. Even if the unit names aren't translated, that's probably the info the player needs to decide whether to wait for the campaign author to fix it.

stevecotton:
Many thanks to you @fujimo-t, but as I've already written other C++ changes from investigating #6204 it felt logical to open a new PR that supersedes this one.

Issue #1397, opened 2017-05-08
wmlindent cannot handle more than one closer on the same line (GNA #22264)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?22264)** by shadowmaster on 2014-07-02

When provided with a file containing more than one closer string on the same line (such as "[/foo])}" instead of "[/foo]" and ")" on separate lines), wmlindent will decrement the indentation level only once instead of multiple times for that line as one would expect, resulting in it increasing the indentation for everything that follows.

For example:

```

{FOO (

    [tag]

    [/tag])} # Two closers here.



[test]

[/test]

```

... is handled by wmlindent as follows:

```

{FOO (

    [tag]

    [/tag])} # wmlindent sees only one closer here!



    [test] # Indentation level incremented by the bug above.

    [/test]

```

I've attached a sample input file where the results of this oversight become obvious from line 5 onwards in the output.

(Reproduced on *)
Release: 1.13.0-dev, 1.10.7+dev
Priority: 3 - Low
Severity: 3 - Normal

wesnoth-bugs:
**Modified on 2014-07-02**

**shadowmaster added attachment:** [indent-test](http://gna.org/bugs/download.php?file_id=21259)

wesnoth-bugs:
**Modified on 2015-07-26**

**shadowmaster changed item group:** WML -> WML Tools

stevecotton:
It can't handle an opener on the same line as something else either. Example, formatted as wmlindent will format it:
```
       {ASSERT ([foo]
       bar = "baz"
   [/foo])}
```

cooljeanius:
oh yeah I've noticed this, too; it's easy enough to work around manually by just adding extra newlines yourself, though

CelticMinstrel:
@Elvish-Hunter Does this seem readily fixable? It doesn't sound too hard but I'm not familiar with the `wmlindent` code so who knows…

Elvish-Hunter:
I actually tried to fix it once, but the result wasn't good: my fix altered the indentation in another part of the code that had some unusual syntax (I believe it as a scenario in LoW), so I scrapped it.
Until I get the right idea for a good fix, if anyone finds a line or a block that confuses the script that part of the code can be wrapped inside this pair of comments: `# wmlindent: start ignoring` and `# wmlindent: stop ignoring`.

cooljeanius:
...or, they can just fix it manually by inserting a linebreak

[...]

PR #5927, opened 2021-07-05
Add a unit test for wesnoth.current.synced_state

stevecotton:
Note that this file gets formattedly badly by wmlindent, however it seems that wmlindent already skips the `data/test/scenarios` directory, the same issue affects `data/test/scenarios/recruit_facing.cfg`. Still working out if that is #1397 or if it should be a separate bug; I think it's a bug that needs fixing in wmlindent for 1.16, as it seems to break on any use of wesnoth.sync.evaluate_single in a `.cfg` file.

### DIFF
--- a/data/test/scenarios/test_synced_state.cfg
+++ b/data/test/scenarios/test_synced_state.cfg
@@ -1,0 +1,90 @@
+# wmllint: no translatables
+
+# Check that wesnoth.current.synced_state works as documented, and that a subset of the standard events match the docs
+# in https://wiki.wesnoth.org/EventWML#Multiplayer_safety
+{GENERIC_UNIT_TEST "test_synced_state" (
+    [event]
+        name = preload
+        [lua]
+            code =<<
+                    unit_test.assert_equal(wesnoth.current.synced_state, 'preload', 'preload should have its own state')
+                >>
+        [/lua]
+    [/event]
+
+    [event]
+        name = start
+        [lua]
+            code =<<
+                    unit_test.assert_equal(wesnoth.current.synced_state, 'synced', 'start should be synchronised')
+                    local result = wesnoth.sync.evaluate_single(
+                        function()
+                            unit_test.assert_equal(wesnoth.current.synced_state, 'local_choice', 'expected value to be local_choice, within synced event')
+                            return { value = wesnoth.current.synced_state }
+                        end)
+                    unit_test.assert_equal(result.value, 'local_choice', 'wrong value returned, within synced event')
+                >>
+        [/lua]
+    [/event]
+
+    [event]
+        name=turn 1
+        [lua]
+            code =<<
+                    unit_test.assert_equal(wesnoth.current.synced_state, 'synced', '“turn X” should be synced')
+                >>
+        [/lua]
+
+        # Trigger the events below. The order that they're triggered in is checked in events-test_victory.cfg.
+        [kill]
+            side=2
+        [/kill]
+    [/event]
+
+    [event]
+        name=enemies defeated
+        [lua]
+            code =<<
+                    unit_test.assert_equal(wesnoth.current.synced_state, 'synced', 'enemies_defeated should be synced')
+                >>
+        [/lua]
+    [/event]
+
+    [event]
+        name=local_victory
+        [lua]
+            code =<<
+                    unit_test.assert_equal(wesnoth.current.synced_state, 'unsynced', 'local_victory should be unsynced')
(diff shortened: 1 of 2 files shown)


stevecotton:
The build failure an engine bug, of the code under test. Should we fix the engine, or should we document that `evaluate_single` doesn't work when called in an unsync'd context?

(I didn't realise that the test failed until the CI notified me, I'd been running the tests on a branch with a slightly earlier version of this test).

data/test/scenarios/test_synced_state.cfg
```diff
@@ -0,0 +1,86 @@
+# wmllint: no translatables
+
+# Check that wesnoth.current.synced_state works as documented, and that a subset of the standard events match the docs
+# in https://wiki.wesnoth.org/EventWML#Multiplayer_safety
+{GENERIC_UNIT_TEST "test_synced_state" (
+    [event]
+        name = preload
+        [lua]
+            code =<<
+                    unit_test.assert_equal(wesnoth.current.synced_state, 'preload', 'preload should have it’s own state')
```
CelticMinstrel, line 10: its not it's

CelticMinstrel:
Can we test a "select" event? I _think_ it would fire when you select a unit from code.

data/test/scenarios/test_synced_state.cfg
stevecotton: Oops, thanks. Fixed in my local copy, but I'd like to work out what needs fixing in the engine before pushing it again.

stevecotton:
> Can we test a "select" event? I _think_ it would fire when you select a unit from code.

Am I doing this wrong, or is this another bug (testing on master)?
```
$ u
{unit: <alice at (13,4)>}
$ u:select()
Lua Error:
  When executing, Lua runtime error: [string "return u:select()"]:1: attempt to call a nil value (method 'select')
stack traceback:
	[string "return u:select()"]:1: in main chunk
$ u:scroll_to()
Lua Error:
  When executing, Lua runtime error: [string "return u:scroll_to()"]:1: attempt to call a nil value (method 'scroll_to')
stack traceback:
	[string "return u:scroll_to()"]:1: in main chunk
```

CelticMinstrel:
It looks like your u is an array of units rather than a single unit, so try `u[1]:select()`.

stevecotton:
Having tried it, no. It appears that selecting a unit via Lua doesn't trigger a `name=select` event.

CelticMinstrel:
How about `u:select(true, true)`?

data/test/scenarios/test_synced_state.cfg
```diff
@@ -0,0 +1,90 @@
...
+                    unit_test.assert_equal(result.value, 'local_choice', 'wrong value returned, within synced event')
+                >>
+        [/lua]
+    [/event]
+
+    [event]
+        name=turn 1
+        [lua]
+            code =<<
+                    unit_test.assert_equal(wesnoth.current.synced_state, 'synced', 'enemies_defeated should be synced')
```
stevecotton, line 34: Force-pushing to fix the function name here ("turn X" instead of "enemies_defeated")

stevecotton:
> Should we fix the engine, or should we document that `evaluate_single` doesn't work when called in an unsync'd context?

I think the answer to that one is "document it", because it's the same behavior in 1.13 (introduced in 3b86c955616c).

@CelticMinstrel are there any changes that you want before I merge this? For the `select` test, I think it should be part of the fix to #5931, separate from this PR.

CelticMinstrel:
As far as I understand, the commit message for 3b86c955616c6b4b0fd9ca6807413924ad40c907 seems to be describing `wesnoth.synchronize_choices`, which is now `wesnoth.sync.evaluate_multiple`. Maybe this functionality was initially added to `synchronize_choice` and then gfgtdf later decided a separate function was better?

CelticMinstrel (approved):

stevecotton:
> As far as I understand, the commit message for [3b86c95](https://github.com/wesnoth/wesnoth/commit/3b86c955616c6b4b0fd9ca6807413924ad40c907) seems to be describing `wesnoth.synchronize_choices`, which is now `wesnoth.sync.evaluate_multiple`. Maybe this functionality was initially added to `synchronize_choice` and then gfgtdf later decided a separate function was better?

Although the commit message describes the `evaluate_multiple` behavior, the code in that commit is already separated. Compare lines 1077-1078 with 1125-1127.

CelticMinstrel:
It seems you're right… so the commit message is just inaccurate? That changed the error to a warning (line 1125+) but I guess it didn't make it actually work?

PR #5930, opened 2021-07-06
Miscellanous stuff for CI

CelticMinstrel:
I'll merge this as soon as CI passes.

### DIFF
--- a/src/scripting/game_lua_kernel.cpp
+++ b/src/scripting/game_lua_kernel.cpp
@@ -1670,8 +1670,9 @@ int game_lua_kernel::intf_end_turn(lua_State* L)
 	//note that next_player_number = 1, next_player_number = nteams+1 both set the next team to be the first team
 	//but the later will make the turn counter change aswell fire turn end events accoringly etc.
 	if (!lua_isnoneornil(L, 1)) {
+		int max = 2 * teams().size();
 		int npn = luaL_checkinteger(L, 1);
-		if (npn <= 0 /*TODO: || npn > 2*nteams*/) {
+		if (npn <= 0 || npn > max) {
 			return luaL_argerror(L, 1, "side number out of range");
 		}
 		resources::controller->gamestate().next_player_number_ = npn;
--- a/src/scripting/game_lua_kernel.cpp
+++ b/src/scripting/game_lua_kernel.cpp
@@ -1922,13 +1922,27 @@ int game_lua_kernel::intf_find_vision_range(lua_State *L)
 	actions::create_jamming_map(jamming_map, resources::gameboard->get_team(u->side()));
 	pathfind::vision_path res(*u, u->get_location(), jamming_map);
 
-	auto vision_set = res.edges;
-	for(auto d : res.destinations)
-	{
-		vision_set.insert(d.curr);
+	lua_createtable(L, res.destinations.size() + res.edges.size(), 0);
+	for(const auto& d : res.destinations) {
+		luaW_push_namedtuple(L, {"x", "y", "vision_left"});
+		lua_pushinteger(L, d.curr.wml_x());
+		lua_rawseti(L, -2, 1);
+		lua_pushinteger(L, d.curr.wml_y());
+		lua_rawseti(L, -2, 2);
+		lua_pushinteger(L, d.move_left);
+		lua_rawseti(L, -2, 3);
+		lua_rawseti(L, -2, lua_rawlen(L, -2) + 1);
 	}
-
-	luaW_push_locationset(L, vision_set);
+	for(const auto& e : res.edges) {
+		luaW_push_namedtuple(L, {"x", "y", "vision_left"});
+		lua_pushinteger(L, e.wml_x());
+		lua_rawseti(L, -2, 1);
+		lua_pushinteger(L, e.wml_y());
+		lua_rawseti(L, -2, 2);
+		lua_pushinteger(L, -1);
+		lua_rawseti(L, -2, 3);
+		lua_rawseti(L, -2, lua_rawlen(L, -2) + 1);
+	}
 	return 1;
 }
 
--- a/src/scripting/game_lua_kernel.cpp
+++ b/src/scripting/game_lua_kernel.cpp
@@ -1454,7 +1454,7 @@ int game_lua_kernel::impl_scenario_get(lua_State *L)
 	return_string_attrib("type", classification.campaign_type.to_string());
 	return_string_attrib("difficulty", classification.difficulty);
 	return_bool_attrib("show_credits", classification.end_credits);
-	return_string_attrib("end_text", classification.end_text);
+	return_tstring_attrib("end_text", classification.end_text);
 	return_int_attrib("end_text_duration", classification.end_text_duration);
 	if(!classification.campaign.empty()) {
 		return_cfgref_attrib("campaign", find_addon("campaign", classification.campaign));
@@ -1513,7 +1513,7 @@ int game_lua_kernel::impl_scenario_set(lua_State *L)
 
 	game_classification& classification = play_controller_.get_classification();
 	modify_bool_attrib("show_credits", classification.end_credits = value);
-	modify_string_attrib("end_text", classification.end_text = value);
+	modify_tstring_attrib("end_text", classification.end_text = value);
 	modify_int_attrib("end_text_duration", classification.end_text_duration = value);
 	if(strcmp(m, "end_level_data") == 0) {
 		vconfig cfg(luaW_checkvconfig(L, 3));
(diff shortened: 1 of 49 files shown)


stevecotton:
I thought the reason that so many Lua changes were rushed in before the 1.16 API freeze was so that the API wouldn't have any more deprecations between now and 1.18.

Also, two other issues that have been discussed multiple times in IRC:
* lack of unit tests
* including multiple commits in that wouldn't pass the CI

src/scripting/game_lua_kernel.cpp
```diff
@@ -1670,8 +1670,8 @@ int game_lua_kernel::intf_end_turn(lua_State* L)
 	//note that next_player_number = 1, next_player_number = nteams+1 both set the next team to be the first team
 	//but the later will make the turn counter change aswell fire turn end events accoringly etc.
 	if (!lua_isnoneornil(L, 1)) {
-		int npn = luaL_checkinteger(L, 1);
-		if (npn <= 0 /*TODO: || npn > 2*nteams*/) {
+		int npn = luaL_checkinteger(L, 1), max = 2 * teams().size();
+		if (npn <= 0  || npn > max) {
```
stevecotton, line 1674: Why create the `max` variable, instead of just putting `2 * teams().size()` in the `if` statement?

CelticMinstrel: Because the latter caused a comparison of differently-signed integers warning, which the CI treats as an error.

CelticMinstrel:
> I thought the reason that so many Lua changes were rushed in before the 1.16 API freeze was so that the API wouldn't have any more deprecations between now and 1.18.

That was indeed the point, but at the same time there were some deprecations that were intentionally postponed, such as the events PR and the floating labels PR. Anyway, besides those and these, I don't think there's anything much _left_ to deprecate.

Pentarctagon:
> This isn't a pull request. I'm just pushing everything I've done in the past week or whatever. And as long as the CI passes, this doesn't even matter.

Not building or not passing the tests will make it harder to bisect any bugs later, if needed.  Of course it matters.

CelticMinstrel:
It's important that each individual commit _builds_, yes. It's not important that each individual commit passes CI. I tested each of these commits as I worked on them, so I highly doubt any individual commit will fail to build, but I can double-check if you want me to.

(It's highly doubt rather than certainty because sometimes I have two changes made simultaneously but put them into individual commits.)

stevecotton:
> It's not important that each individual commit passes CI.

Of course it is, so that people can use the CI or locally run the CI scripts as part of bisecting. Do you remember the 1.15.13 release, when @Pentarctagon and I were searching for the bug somewhere in a batch of around 25 commits?

image

src/scripting/game_lua_kernel.cpp
stevecotton: Either a `static_cast<unsigned>` or putting the `int max =` on its own line would be better than having one line with two declarations separated by a comma.

CelticMinstrel:
> Of course it is, so that people can use the CI or locally run the CI scripts as part of bisecting.

Why would you do that? It's a total waste of time. If you're debugging a unit test failure, just run the unit tests. If you're debugging a schema failure, just run the schema. If you really want, you can run both always, but there's certainly no reason to run the whitespace check for _any_ bisect, and there's no need to run the multiplayer test or any of the WML unit tests if you're debugging a UI bug.

Your criteria for telling the bisect whether a commit is good or bad isn't "did the CI pass", it's "did the specific bug I'm looking for manifest". There could've been multiple bugs introduced in that commit range, some of which could be fixed already.

stevecotton:
> Your criteria for telling the bisect whether a commit is good or bad isn't "did the CI pass", it's "did the specific bug I'm looking for manifest".

Under what circumstances would those two criteria have different answers? Given the output of a build, how much of a developer's time does it take to get the answer to each of those? What changes to developers' workflows would improve the situation?

> There could've been multiple bugs introduced in that commit range, some of which could be fixed already.

What changes to developers' workflows would improve that situation, or even completely avoid it?

CelticMinstrel:
> Under what circumstances would those two criteria have different answers?

Probably reasonably often. Often the CI won't help you track the bug at all. If it's a UI bug, you need to actually launch the game and do something to reveal it. If it's a bug that's triggered in a certain campaign by a certain save file, you need to actually load up that save file and see if the bug triggers. I think whether CI passes and what it outputs will probably be irrelevant to a bisect more often than it is useful.

It's useful if what you're bisecting is specifically a CI failure (mainly, a unit test failure, or _maybe_ a schema failure). Otherwise it doesn't really tell you much of note. Maybe it's one way for you to get a build that you can test with, but that's not dependent on all the tests passing. And if you _are_ bisecting a unit test failure, you don't really need to run the schema validation and the multiplayer test and everything else. You just need to run the unit tests.

> What changes to developers' workflows would improve that situation, or even completely avoid it?

There's nothing that will completely avoid it. Bugs sneak in without being noticed. It's just something that happens – even if you do all the testing you can think of, you might have missed a case.

stevecotton:
Are you deliberately ignoring the context of the conversation?

For an issue that can be detected by the CI no-one else has to do a build at all. All it takes is a quick glance at the red and green icons on https://github.com/wesnoth/wesnoth/commits/master

Pentarctagon:
> It's useful if what you're bisecting is specifically a CI failure (mainly, a unit test failure, or maybe a schema failure). Otherwise it doesn't really tell you much of note.

Those are both very important things.  Even if the bug I'm bisecting wasn't found by the CI, I don't want to have to try and figure out if the particular commit being tested is behaving strangely because something changed with the bug I'm actually tracking down or if it's because someone committed a change that broke the WML unit tests.

src/scripting/game_lua_kernel.cpp
CelticMinstrel: I don't see what's wrong with having two declarations on the same line, but if it'll make you stop complaining, I'll put them on separate lines.

CelticMinstrel:
>  Those are both very important things.

I'm assuming by "both" you mean schema and unit test failures. Yes, they're important things, but they're not usually the kind of bug that requires bisecting.

Pentarctagon:
Did you read anything after the first sentence?

CelticMinstrel:
Even if the unit tests were broken and fixed within the bisect range, what are the odds that the breakage is directly related to the bug you're bisecting for?

Certainly if the unit tests broke because _everything_ broke, like a syntax error in a core Lua file, that could (depending on what you're bisecting for) be quite inconvenient. On the other hand, if the unit tests broke because something logged an error, it's probably not even an inconvenience.

Obviously we should strive to avoid pushing the former type of commit to master.

Pentarctagon:
Yes, the chance of a WML unit tests error (for example) affecting a bisect is low.  That's not the point though.  The point is that it easily can be, and should be, 0%.

CelticMinstrel:
> The point is that it easily can be, and should be, 0%.

That's not really true. The unit tests take forever to run. It's ridiculous to expect everyone to run them on every single commit just to make sure not even the slightest thing broke, and it's also ridiculous to expect people to run them before pushing (unless you're pushing directly to master) because the CI is going to run them anyway, so running them yourself is a total waste of time. I run them locally when I'm doing stuff that directly touches them (adding new unit tests or changes that touch a lot of stuff) or if I see a unit test failure in the CI with a non-obvious cause.

And the boost unit tests don't even run at all outside of Linux. I mean, _theoretically_ they can (the issues are all in the build process, not in the code), but no-one's bothered to figure out what's needed to make them run.

Pentarctagon:
I don't know about macOS, but the boost tests do build on Windows - we just don't have any Windows devs around and active enough to look into why they immediately fail after starting.

If you're pushing the master, then yeah, the unit tests should be run locally first as should the whitespace fixup script (...which also won't work entirely on Windows).  Even when pushing to a PR though, there shouldn't be individual commits that break and then immediately fix one or more of the tests - there should just be a single commit where the necessary changes are made that therefore would pass the tests.

CelticMinstrel:
> Even when pushing to a PR though, there shouldn't be individual commits that break and then immediately fix one or more of the tests

Sure. It makes sense that if you have a unit-test-breaking commit and a unit-test-fixing commit in the same PR should be squashed together before merging.

The boost unit tests do build on MacOS. They just don't run because of (I think) signing and/or hardening issues and I'm not sure how to deal with that.

Pentarctagon:
> Sure. It makes sense that if you have a unit-test-breaking commit and a unit-test-fixing commit in the same PR should be squashed together before merging.

That's what octalot was talking about before, I believe.  For example, "Lua API: Deprecate helper.set_wml_action_metatable" then the next commit is "Lua: Update all mainline references to the helper module".

CelticMinstrel:
I don't think that qualifies as a unit-test-breaking commit, but I did swap the order if that helps.

Personally, I think it's clearer if the commit that deprecates something and the commit that updates everything that use it are kept separate, as there could be hundreds of places that use it which will drown out the code that actually implemented the deprecation.

Pentarctagon:
Given octalot previously said
> including multiple commits in the history that wouldn't pass the CI

there were more than just that one.

Issue #5931, opened 2021-07-06
Lua unit:select(true, true) does not fire select events


stevecotton:
### Game and System Information
- Very latest version of master on Linux

### Describe the bug
According to the docs, `unit:select(true, true)` will trigger an `[event]name=select`. It doesn't.

### To Reproduce
Add this code to the unit test in PR #5927 :

```
+++ b/data/test/scenarios/test_synced_state.cfg
@@ -23,6 +23,17 @@
                             return { value = wesnoth.current.synced_state }
                         end)
                     unit_test.assert_equal(result.value, 'local_choice', 'wrong value returned, within synced event')
+                    wml.variables["start_event_happened"] = true;
+                >>
+        [/lua]
+    [/event]
+
+    [event]
+        name=select
+        [lua]
+            code =<<
+                    unit_test.assert_equal(wesnoth.current.synced_state, 'unsynced', 'select should be unsynced')
+                    wml.variables["select_event_happened"] = true;
                 >>
         [/lua]
     [/event]
@@ -32,6 +43,11 @@
         [lua]
             code =<<
                     unit_test.assert_equal(wesnoth.current.synced_state, 'synced', '“turn X” should be synced')
+
+                    local u = wesnoth.units.find_on_map({id = "alice"})
+                    u[1]:select(true, true)
+                    unit_test.assert(wml.variables["start_event_happened"], 'Skipped the start event')
+                    unit_test.assert(wml.variables["select_event_happened"], 'Skipped the select event')
                 >>
         [/lua]
 
```

Result will be
```
Assertion failed: Skipped the select event
FAIL TEST: test_synced_state
```

stevecotton:
There's a WML unit test for this in #5938, written in WML so that it can run on 1.14. The event doesn't fire on 1.14 either.

PR #5938, opened 2021-07-07
TEST FAILS: Add a unit test that [select_unit]fire_event=yes works

stevecotton:
@CelticMinstrel following on from #5927, I looked at testing `[select_unit]fire_event` or `unit:select(true, true)`. This version of the test is written in WML so that it can be run on 1.14 too.

It doesn't appear to work, nor did it work on 1.14.

Running the test interactively doesn't make it work either.

Running the test interactively and manually clicking on Alice with the mouse
does fire the event.


stevecotton:
Logged as bug #5931, so I'll close the PR and keep the bug open.

PR #5960, opened 2021-07-14
[world conquest] [1.16] update to new mushroom terrain from the deprecated mushroom terrain

knyghtmare:
addresses #5135 for the 1.16 branch
I shall do the one for master branch later on.


stevecotton:
Still needs work on https://github.com/wesnoth/wesnoth/pull/5961#issuecomment-880843547

knyghtmare:
hmm, it appears as though I need to another commit (part 3) since `git grep -rl "^Uf" | wc -l` returned 13 this time

Pentarctagon:
Is that last commit part 3 and this is completed, or is that not yet done?

knyghtmare:
Would be nice if someone ran `git grep` to check. I think I did get all of them (for this PR)

Pentarctagon:
`git grep -rl "^Uf" | wc -l` for this PR returns 0. `grep -Iirn "\^Uf"` also returns no results.

Pentarctagon:
@knyghtmare is there anything else needed for this PR or it complete?

Pentarctagon:
`grep -Irn "\^Uf"` in the WC directory also gives no results, so it looks like it.

PR #5964, opened 2021-07-18
1.16 - Troll defense animations

doofus-01:
Some of the trolls have really lazy defense animations, must have been placeholders.
(Not criticizing the previous Art Director, there's too much to keep track of for one person...)


stevecotton:
The new Grunt animation seems more like wincing or idling than defending - I feel the club should be blocking the hit instead of staying at waist height.

doofus-01:
There can be more than one animation, if someone wants to make one.  Blocking, dodging, and flinching animations would be better than one all-purpose animation, but there would need to be more people working on this.

doofus-01:
This did end up being a little underwhelming, but still a small improvement.  I'll merge once it's passed CI.

PR #5965, opened 2021-07-18
Prevents logging of duplicate deprecation messages and messages missing timestamps.

### LINKED ISSUE #5942: Some log file lines are missing timestamp
 - 1.15.14, windows

### Describe the bug
Log file contains lines such as
```
20210709 02:29:32 warning deprecation: special= has been deprecated and may be removed in version 1.17.0.; Please use special_id or special_type instead
20210709 02:29:32 warning deprecation: special= has been deprecated and may be removed in version 1.17.0.; Please use special_id or special_type instead
Invalid WML found: special= has been deprecated and may be removed in version 1.17.0.; Please use special_id or special_type instead (2)
20210709 02:29:32 warning deprecation: special= has been deprecated and may be removed in version 1.17.0.; Please use special_id or special_type instead
Invalid WML found: special= has been deprecated and may be removed in version 1.17.0.; Please use special_id or special_type instead
```
It feels like different types of messages which do not belong in same file. But more visibly some of them are missing timestamp.

### Expected behavior
Either summary messages have timestamp, or they are not logged in same file.

Pentarctagon:
Fixes #5942

---

With the code:
```
[event]
    name=start

    [lua]
        code=<<wesnoth.alert("hi")>>
    [/lua]
[/event]
```
currently it logs:
```
20210717 23:25:58 info deprecation: wesnoth.alert has been deprecated indefinitely.;  (Note: You should use gui.alert instead in new code)
Invalid WML found: wesnoth.alert has been deprecated indefinitely.;  (Note: You should use gui.alert instead in new code)
```
after this PR it will log:
`20210717 23:29:42 error wml: Invalid WML found: wesnoth.alert has been deprecated indefinitely.;  (Note: You should use gui.alert instead in new code)`


CelticMinstrel:
It feels wrong to shuffle the deprecation message out of the deprecation logdomain and into the wml logdomain…

Pentarctagon:
Unless there's a way to tell it to log everything sent to chat *except* the deprecation messages, I'm not sure what the alternative is.

ProditorMagnus:
Why "**info** deprecation" -> "**error** wml"? "deprecated indefinitely" suggests it is not especially severe.

stevecotton:
> Why "**info** deprecation" -> "**error** wml"? "deprecated indefinitely" suggests it is not especially severe.

I take the opposite observation here - why doesn't "**error** wml" trigger the **BROKE_STRICT** flag?

Pentarctagon:
Because `lg::wml_error()` isn't actually a logger, it's just a stringstream that gets read out periodically in pump.cpp so that duplicate WML error messages can be printed ingame/logged once rather than one line for every identical WML error.

Pentarctagon:
The ironic thing being, if I set wml_error() to also mark BROKE_STRICT, then now the test `check_interrupts_continue_global` fails because it errors on `error wml: Invalid WML found: [continue] found outside a loop scope!`.

Pentarctagon:
It looks like some other tests break as well, so probably the way to address this would be to not have BROKE_STRICT as its own separate result but instead have there be a BROKE_STRICT variation of other unit test results.  So for example there would be `SUCCESS`, `SUCCESS, BROKE STRICT`, `FAIL`, `FAIL, BROKE STRICT`.  Since the tests themselves are expected to give another result, but then that result currently gets overridden by BROKE_STRICT.

CelticMinstrel:
My position is still that treating a deprecation of level 1 or 2 as an error is a mistake. Deprecations such as these are supposed to exist as a fallback so that old code can continue to work for as long as possible without being updated. Level 2 should probably be a warning, but level 1 shouldn't even be that.

Certainly, the "error wml" should break strict in a unit test (probably it already does); but these also shouldn't _be_ "error wml". They should be "info deprecation" or at most "warning deprecation".

One possibility would be to have the WML error stream contain info giving the log level of each line. For example, a line beginning with "Warning:" could be treated as a warning, instead of an error. Or you could use control characters if you prefer, or even add separate WML error streams for each level (although that creates a problem of ordering).

> rather than one line for every identical WML error that was recorded since the last time it was read out.

Does it actually suppress duplicates? It doesn't seem possible from what I know of how it works.

Pentarctagon:
> Certainly, the "error wml" should break strict in a unit test (probably it already does); but these also shouldn't be "error wml". They should be "info deprecation" or at most "warning deprecation".

It currently doesn't.  The check for whether or not errors have been logged is determined by the logger's `operator()`, which the WML error messages don't use (which I assume means they are also excluded from the work you did previously to make logging thread safe).

> One possibility would be to have the WML error stream contain info giving the log level of each line. For example, a line beginning with "Warning:" could be treated as a warning, instead of an error. Or you could use control characters if you prefer, or even add separate WML error streams for each level (although that creates a problem of ordering).

That might be the best option here, as long as the default is that it's an error - that's how the existing functionality works and it also wouldn't be good design for everyone who ever writes to the WML error stream to need to know/remember to prepend a magic identifier string.

> Does it actually suppress duplicates? It doesn't seem possible from what I know of how it works.

Yes: https://github.com/wesnoth/wesnoth/blob/88873e3faec24a7020b5fad61bffa41693ec92f2/src/game_events/pump.cpp#L341-L343

CelticMinstrel:
> It currently doesn't.

But with the change in this PR, it does, right?

> That might be the best option here, as long as the default is that it's an error

Yeah, of course the default would be an error.

Pentarctagon:
> But with the change in this PR, it does, right?

It does, since part of the change is to have WML errors (eventually) be sent to a log stream rather than directly to std::cerr.

stevecotton:
> The ironic thing being, if I set wml_error() to also mark BROKE_STRICT, then now the test `check_interrupts_continue_global` fails because it errors on `error wml: Invalid WML found: [continue] found outside a loop scope!`.

My preferred solution would be to have some finer-grained way for tests to indicate to the test framework that they expect messages to be logged, and then failing any test which logs unexpectedly.

CelticMinstrel:
Yeah, that would definitely be better. Basically, let a test specify that a specific log message won't break strict.

Pentarctagon:
While that would be useful, I think adding that functionality is out of scope for this PR and should be a separate issue.  The update I pushed just now fixes the WML unit test failures by adding unit test results to indicate what the result was and whether strict was broken during the test.

CelticMinstrel:
It's still not correct for deprecation messages to not be output on the deprecation logdomain though…

Pentarctagon:
I don't see a sensible way to have the deprecation.cpp to tell pump.cpp that it already logged a particular message.  So either the message gets duplicated in the log by both the deprecation log domain and the chat message handler, or you have to pick only one of them to do the logging.  However since the chat message handler logs more than just deprecation messages, removing logging from there would also prevent other messages from being logged, so that leaves deprecation.cpp as the place to skip the duplicate logging.

CelticMinstrel:
I think it makes more sense to just _not_ log deprecation messages to chat at all. It does mean we need another place for deprecation messages that aren't called from Lua, though. (Deprecation messages called from Lua, using `wesnoth.deprecated_message`, are logged to the Lua console as warnings, using the new warning system in Lua 5.4.)

Pentarctagon:
If the deprecation messages aren't clearly made visible, then nobody is going to know to update their code until the deprecated function is removed.  Which is even more important now given the massive amount of things you deprecated during 1.15.

CelticMinstrel:
> If the deprecation messages aren't clearly made visible, then nobody is going to know to update their code until the deprecated function is removed.

No. There's no need to make _all_ deprecation messages so clearly visible. Only level 3 and 4 deprecation messagess need to be pushed into the developer's face. If we decide we want to _remove_ something that was deprecated at level 1 or 2, we'll first promote it a level so it becomes more visible.

> Which is even more important now given the massive amount of things you deprecated during 1.15.

These are nearly all Lua deprecations, which as I said are visible in the Lua console. Unfortunately, I don't think pushing non-Lua deprecations to the Lua console is viable, since there may not _be_ a Lua console at the time the deprecation message is emitted.

---

Instead of dumping all the deprecation messages into the chat, I think it would be sufficient to dump a _single_ message the first time something deprecated is triggered (per-scenario), something like "Deprecated WML or Lua found - see log or Lua console for details". Even that is only really needed if it's a level 3 or 4 deprecation.

Pentarctagon:
This PR is meant to address two things: some log messages having no timestamp and some log messages being duplicated.  The debate over which deprecation levels should be displayed and where is entirely irrelevant to this PR.

CelticMinstrel:
I'm not sure it's _entirely_ irrelevant, but that's not the main substance of my objection here. It's simply _not right_ for a deprecation of level 1 or 2 to be shown as an error message. The function you're changing has clear logic for translating deprecation level to log level, and this PR just wipes that clean to make everything an error.

Outputting `lg::wml_error` ultimately to the "error wml" stream _is_ correct, however… so I think the best quick fix is to _remove_ the logic that mirrors deprecations to `lg::wml_error`. Although you could also just leave that function unchanged, because once 1.16.0 is out, the block mirroring it to `lg::wml_error` won't even be entered unless it's explicitly enabled in advanced preferences. It would also make sense to make the version check only be the _default_, so that it can be explicitly _disabled_ even in development builds.

Pentarctagon:
The `if(log_ptr && !log_ptr->dont_log(log_deprecate))` part currently, and would continue to, prevent info level deprecation messages from being logged, wouldn't it?  Or am I misunderstanding what `dont_log` does?

Pentarctagon:
Then again I suppose the before and after example I posted shows this isn't the case.  At the top of the file
```
// Set the default severity with the second parameter.
// -1 means the default is to never log on this domain.
// 0 would mean log errors only.
// 1 would mean log errors and warnings.
// and so on and so on.
static lg::log_domain log_deprecate("deprecation", 0);
```
would suggest that only deprecation errors are logged, however something else seems to set the default level to 2 (info) rather than 0 (error).

Pentarctagon:
So to fix
> The function you're changing has clear logic for translating deprecation level to log level, and this PR just wipes that clean to make everything an error.

it should change from
```
if((preferences::get("show_deprecation", false) || game_config::wesnoth_version.is_dev_version()) {
```
to
```
if((preferences::get("show_deprecation", false) || game_config::wesnoth_version.is_dev_version())
	&& out_log.get_severity() == lg::severity::LG_ERROR) {
```

CelticMinstrel:
> The `if(log_ptr && !log_ptr->dont_log(log_deprecate))` part currently, and would continue to, prevent info level deprecation messages from being logged, wouldn't it? Or am I misunderstanding what `dont_log` does?

`dont_log` is the underlying check function that controls _all_ of Wesnoth's log functions, to determine whether to output the message based on the level currently assigned to the given logdomain. So it won't categorically prevent info level messages from being logged.

> however something else seems to set the default level to 2 (info) rather than 0 (error).

Are you running with `--debug`? It looks like this block would have the effect you're seeing:

https://github.com/wesnoth/wesnoth/blob/e5b4a39242846a2dbcc6044f793db151f0f5ce92/src/game_config.cpp#L116-L128

Actually, in retrospect, I think boosting it by 2 is rather excessive; 1 seems sufficient to me, although we could also remove this code altogether.

> So to fix […] it should change from […] to […]

I was thinking of something more like:

```c++
if(preferences::get("show_deprecation", game_config::wesnoth_version.is_dev_version())) {
```

Possibly with a `!` in front of is_dev_version? I think it's correct without it but not 100% sure.

Pentarctagon:
> I was thinking of something more like:
> if(preferences::get("show_deprecation", game_config::wesnoth_version.is_dev_version()))

How is that any different than what it's doing today?

Also, thinking about it a bit more, I don't actually know what you mean by
> The function you're changing has clear logic for translating deprecation level to log level, and this PR just wipes that clean to make everything an error.

The original logic was https://github.com/wesnoth/wesnoth/blob/a1f44d8f6448eda20ced1af8ec042b2f9b2f787f/src/deprecation.cpp#L78-L85

I updated it later to https://github.com/wesnoth/wesnoth/blob/124adb308f402cfaecfaf60e587a707520fc8045/src/deprecation.cpp#L78-L86

And here I'm updating it to https://github.com/wesnoth/wesnoth/blob/4c719457b9490955567204d0fb738aacf33795ae/src/deprecation.cpp#L78-L87

This PR does nothing to "make everything an error" - it has for a long time already treated deprecation messages that get displayed ingame as WML errors, the only thing this PR does is prevent a duplicate message from also being logged from the deprecation log domain.

CelticMinstrel:
> Also, thinking about it a bit more, I don't actually know what you mean by

The logic I'm referring to is further up the function, in the switch statement. The change to this function doesn't itself make deprecation messages always an error; it's also adding a timestamp to the WML error message that does it.

> How is that any different than what it's doing today?

It means that the `is_dev_version` check is a default, not an override.

In the current form, a user's preference is ignored in dev versions, and deprecation messages are _always_ dumped to the chat.

With my suggested change, a user who doesn't change the setting will see the deprecation messages in dev versions but not in stable versions. However, if a user goes into preferences to change the setting, it will be respected.

There are three possible results when looking up the "show_deprecation" key from preferences - it could be missing, true, or false. My suggested change only uses the `is_dev_version` result if it's missing.

> the only thing this PR does is prevent a duplicate message from also being logged from the deprecation log domain.

Which defeats the point of even _having_ a deprecation logdomain, if it doesn't contain all the deprecation messages.

Pentarctagon:
I think the only options at this point then are:
* Continue to have the duplicate log messages.
* Require that logging to terminal/disk and displaying the message ingame are two separate operations everywhere.

CelticMinstrel:
I think the duplicates are unfortunate, but not _too_ bad as long as they can be controlled by the advanced preference.

Pentarctagon:
I kinda think having it as two separate methods that need to be used would be better, if only because then it's clearly separating what's being done.

Pentarctagon:
I'll be merging this tomorrow if there are no other comments.

stevecotton:
> With my suggested change, a user who doesn't change the setting will see the deprecation messages in dev versions but not in stable versions. However, if a user goes into preferences to change the setting, it will be respected.
> 
> There are three possible results when looking up the "show_deprecation" key from preferences - it could be missing, true, or false. My suggested change only uses the `is_dev_version` result if it's missing.

I feel the old behavior was a feature rather than a bug.

The migrate-preferences feature can also copy that setting between branches.

stevecotton:
This PR seems okay, but also seems to be adding complexity to fix a bug that'll disappear on when we get a non-dev release number anyway.

CelticMinstrel:
> I feel the old behavior was a feature rather than a bug.

Our handling of deprecation messages makes them _too_ visible in my opinion. It's a good state for someone who wants to update their addon (or mainline) to avoid using _any_ deprecated things, but it's a bad state for a player or even an add-on developer not explicitly working on updating deprecated things. Deprecation messages come in four levels for a _reason_. They're not necessarily supposed to mean "you must update this immediately or your code will break". A level 1 deprecation is a warning that "this is obsolete and you shouldn't use it in new code but it's not going to hurt anything if you leave it in your old code".

stevecotton:
The old code (without this PR) already supports those levels. Please can you describe what you'd change *compared to the UI that 1.16 will have, when is_dev_version() returns false*, without this PR?

Pentarctagon:
My own opinion is that all deprecations should be done with the intention that people actually start using the non-deprecated code eventually.  If the deprecation will never break (deprecated indefinitely) and is effectively invisible (only logged to a log file that a lot of people don't seem to know even exists) then I don't understand the point of deprecating the original function in the first place.

CelticMinstrel:
> it's not actually accomplishing anything

Sure it is. Having the API actually make sense is a good goal in and of itself. Renaming something because the old name is misleading is an accomplishment in and of itself, and it's good to encourage people to only use the new name in new code, but there's also no need to break everything that uses the old name in such a case. That's what "deprecated indefinitely" is for.

I used level 1 for most of the deprecations in 1.16, mostly just out of habit, but in retrospect, some of them should've been level 2. I can bump them up on master but I assume it's too late to bump them up for 1.16.

Pentarctagon:
I'm not talking about breaking things, I mean that if by default deprecated indefinitely only gets put into the log file, then you're barely accomplishing anything by deprecating it since only a tiny fraction of UMC authors will actually see it.  Even more so on platforms that only have terminal output rather than a log file.

CelticMinstrel:
That's true; for Lua deprecations they'll mostly show in the Lua console, but non-Lua ones won't. Having a similar central place to view deprecation messages could maybe be useful…

ProditorMagnus:
I see lowest deprecation messages as something user intentionally should ask for, similar to deciding to run wmllint.

Pentarctagon:
That's essentially what starting Wesnoth with --debug does, though that's probably even less known than the log file.

Pentarctagon:
In any case, if there aren't any objections to things this PR in terms of what's logged vs not logged, I'm going to merge this tomorrow sometime.

PR #5968, opened 2021-07-18
[1.16] Tutorial S01: Don't let the first Quintain attack on turn 2

stevecotton:
A player can follow Dacyn's instructions to the letter, yet get
killed if they choose a village 2 hexes away from the Quintain on
turn 2.

There are better suggestions about handling this in
https://r.wesnoth.org/t54644 , however during the string-freeze
let's simply make the Quintain stay on its starting spot on turn 1.

Add a CLEAR_PRINT because the player has just completed the
"End your turn" instruction.

Given the release timing, I'm opening this as a 1.16 PR, intending
to cherry-pick to master afterwards.

### DIFF
--- a/data/campaigns/tutorial/scenarios/01_Tutorial_part_1.cfg
+++ b/data/campaigns/tutorial/scenarios/01_Tutorial_part_1.cfg
@@ -621,6 +621,19 @@
         {PRINT (_"End your turn")}
     [/event]
 
+    # On the Quintain's first turn, it's always next to the player, but
+    # sometimes the AI chooses to move before attacking. If it moves onto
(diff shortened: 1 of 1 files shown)


Pentarctagon:
If there are better ways to fix this, wouldn't it be better to implement one of the better ways in master?

stevecotton:
Even if there will eventually be a better fix in master, I think cherry-picking makes sense.

* It means a small improvement to master arrives quickly.
* It makes clear that any subsequent change to master is intended to superceed this.
* It makes it easy to cherry-pick the subsequent improvement from master to 1.16.x when the string-freeze is over.

Pentarctagon (approved):

Issue #4531, opened 2019-10-31
Title screen GUI design


nemaara:
Making an issue for this finally, since I've been talking about it for a while: here would be some general things about improving the title screen and in game GUIs (which it really would be nice to improve). I'll split the in game GUI discussion into a separate issue.

One fundamental thing in GUI design is using layout, sizing, and coloring to draw attention to elements with different importance. The (functionally) most important elements also need to be the largest and stand out the most in the design. Another important thing is how much information you have in the GUI: every extra piece added dilutes all the other information contained in the display. In general, in the base display, only the most important things need to be there, and extra information can be shoved into side menus to avoid cluttering the base display. **_The primary goal is ease of use and understanding (not how much information you can cram into one space)._** With this in mind, let's look at our title screen.

![image](https://user-images.githubusercontent.com/13598479/67983640-7f24ec00-fbfb-11e9-9546-5ad9b0f0bf06.png)

As can be seen from the part circled in red, we have 10 (!) buttons on the right all with the same size and coloring with absolutely nothing to differentiate them. Supposing that I were a new player opening up the game for the first time, there is nothing visually to direct me to what I want to do, which is just to play something. The first thing I'm going to think about when opening the game isn't to fix my preferences nor is it to open the help menu. I just want to play the game first, then see if I need help or to fix some options (say, playing without music if it bothers me). 

Next, circled in blue, we have the "tips of the day" section, which is basically the same size as the 10 buttons menu, meaning that the GUI gives it more or less has the same importance as the 10 buttons on the right combined. Obviously, however, tips of the day are not nearly as important as most of the single buttons on the right, let alone all of them combined.

Therefore, when considering the functionality of every section and button on the title screen, we should think about emphasizing only the most important 4-5 buttons, and leave the others to small icons off to the side. "Language", for example, is just a part of preferences and can be merged into that. "Tutorial" can be merged into "Campaigns" in some form. "Add-ons" and "Map Editor" are part of content creation and can be merged into a category as well. "Credits" is definitely not as important to players as any of the above, and probably should be a small button or icon off to the side. Under this design, we'd have "Singleplayer", "Multiplayer", "Content Creation", "Preferences", and optionally a quit button. Note that "Preferences" in many games is also provided as a small icon rather than a separate button.

[...]

PR #5969, opened 2021-07-19
[1.16] Title screen: move "change language" back to the text-button area

stevecotton:
Okay for the string-freeze - the necessary string was re-added in fc50458b.

No forward-port required, as master will have a full title screen redesign in
issue #4531. This commit is reverting acc96b16, which is a part of #4531,
because the redesign has paused at a place that causes problems with large
screens - the button added in acc96b16 moves further from the center of the
screen as the screen size increases, and it doesn't scale with screen size. On
an 800x600 window it's easily noticeable, but on 2560x1440 it's less obvious
than either Dan Tonk or Weldyn.

### DIFF
--- a/data/gui/window/title_screen.cfg
+++ b/data/gui/window/title_screen.cfg
@@ -254,6 +254,7 @@ where
 			{_GUI_BUTTON "addons" _"Add-ons" _"Download usermade campaigns, eras, or map packs"}
 			{_GUI_BUTTON "editor" _"Map Editor" _"Start the map editor"}
 			{_GUI_BUTTON "preferences" _"Preferences" _"Configure the game’s settings"}
+			{_GUI_BUTTON "language" _"Language" _"Change the language"}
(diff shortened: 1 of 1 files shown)


nemaara (commented):
Not my call, ask Iris and Vultraz.

Vultraz:
I have no objection but also don't see the huge necessity of this. The button seems fine as is.

CelticMinstrel:
If this is just intended as a temporary fix for usability issues on larger screens, there's also the option of having _both_ buttons. It's redundant, but it won't hurt anything.

stevecotton:
> I have no strong objection but also don't see the huge necessity of this. The button seems fine as is. I've vote not to merge.

Is that because the button seems visible enough, or because Wesnoth should detect the system language?

Vultraz:
It should definitely detect (and does) the system language, AFAIK.

stevecotton:
The comments seem to be slightly to the "don't merge" side of neutral, so closing.

Issue #1135, opened 2017-05-08
Support for gettext plurals in WML (GNA #11811)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?11811)** by vladimirslavik on 2008-06-10

Apart from providing the basic translation with _"", it might be useful to have full support for plurals as well.

There are two possible ways to achieve this:

1)

Adding some keys like message_plural and plural_var to all places where is text that can be seen. The possible candidates for such extension are:

[message] message, [print] text, [label] text, [objectives] [objective] text

Other places are probably not worth attention as they aren't likely to show anything with variable number.

Additionally something like [evaluate_plural] (could be part of [set_variable]?) for retrieving already converted string could be useful (although I can't really see UMC authors learning how to use that).

Example:

```

[message]

  message = _"I have $count unit"

  message_plural = _"I have $count units"

  plural_var = $count

[/message]

```

Advantages: Already possible with current WML system without invasive changes.

Disadvantages: Only directly "showable" strings can be pluralized. Quite a lot harder on wml parser that extracts translatable stuff.

2)

Essentially extending representation of wml objects with "strings" that can be a compound of three parts - singular, plural and variable for calculation. The syntax can be anything... eg. _("sg", "pl", $var)

Example:

```

[message]

  message = _("I have $count unit", "I have $count units", $count)

[/message]

```

Advantages: Slightly more intuitive usage, can be parts of string concatenation like a+b+c.

Disadvantages: Big change to the way wml strings are - new type of "string" that cannot be mixed ad hoc with normal strings without other effects - how to store / when evaluate the expression if it's a part of something like a+b+c?

(Reproduced on Windows XP)
Release: any
Priority: 5 - Normal
Severity: 2 - Minor

wesnoth-bugs:
**Modified on 2008-06-12**

**mordante wrote:**

> I really like this idea, I share your concern with learning UMC creators to use it. On the other hand you can see the difference in most English sentences as well, it's no as hidden as the genders. But if authors have problems with it, translators might help with posting suggestions on the forum.
> 
> 
> 
> I think proposal 1 looks the best, only the name 'plural_var' sounds a bit confusing, maybe just call it 'count'.

wesnoth-bugs:
**Modified on 2008-06-24**

**knarf wrote:**

[...]

PR #5979, opened 2021-07-24
Remove a leading space from the Silver Mage's [special_note]

stevecotton:
The strings in special-notes.cfg don't have a leading space,
and this one causes a warning in poedit.

### DIFF
--- a/data/core/units/humans/Mage_Silver.cfg
+++ b/data/core/units/humans/Mage_Silver.cfg
@@ -22,7 +22,7 @@ They have, in fact, their own order amongst the ranks of magi, an order which wi
 
 Silver magi are often more physically adept than other magi, and their skills are of undeniable use on the battlefield, if one can manage to induce the mage to apply them."
     [special_note]
-        note= _"SPECIAL_NOTE^ Silver Magi are well-attuned to their magical natures and are highly resistant to non-physical damage."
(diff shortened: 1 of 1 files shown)


nemaara (approved):

AncientLich:
@stevecotton Sorry I need a summary of issues you find in wmlxgettext. Not sure about how much time I will need because I don't have so much free time available and, moreover, I currently not have a proper "home" to work at the moment (only a small room, shared with my girlfriend... the intention is to buy a house, but I will need some months before finding and obtaining it).
So it could be very hard for me, in this part of my life, to grant the maintainance to wmlxgettext... I will try to do my best, but I can't make promises.

After saying that, I need a summary of all the problems you found (I am a bit lost on fragmented reports) and I will start to think how to solve all those issues together, when I will be able to do it.
Sorry if I answer so slowly....

stevecotton:
@AncientLich I'll keep a summary of anything new that turns up, but AFAIK there's no issues needing attention. All I can think of are the low-priority #5989 and the "I suggest we document and close this" #1135.

AncientLich:
About #5969 I am almost sure that the issue is generated by msgmerge and not wmlxgettext, so probably this issue cannot be fixed easly, and probably could require an entire new tool that acts after msgmerge... easier to tell it than to do it, I am afraid.

The second feature is probably out of my skills... not sure if I can improve wmlxgettext to be so clever to manage that new scanario...

Issue #5973, opened 2021-07-21
Backport string changes after string freeze


Pentarctagon:
Backport 1541a31a5689949fdc98db83b4c63029b85444de to 1.16.1.

stevecotton:
Backport #5979, if it merges

stevecotton:
Backport #5985, if it merges

stevecotton:
Backport #5987, if it merges

kabachuha:
Backport https://github.com/wesnoth/wesnoth/commit/d445cb36dc1e15e7aecb75a979bea762bf122ba4 to 1.16.1.

kabachuha:
Backport https://github.com/wesnoth/wesnoth/commit/3de79c15fbf7862f6fa2e69225cc30fcc67c6ba9

kabachuha:
Backport c098140

Pentarctagon:
cbd0d6bd9ae27d1b459b334dbaf95ad0de14b828

Pentarctagon:
04bc431e04b9d6f3d0b443a61c3a83d2977dffd8

kabachuha:
Also backport https://github.com/wesnoth/wesnoth/commit/536d6a226cd4ed76b034ef06a521e56eef309c68 as it allows the usage of unit names in the text of HttT without workarounds in some languages.

P.S. if you don't mind I'll pin this issue as it has almost left the first page of issues on my screen and it's important for any string changes that have been intended to be in 1.16 (1.16.1 after the unfreeze).

Wedge009:
71ebeb8eb4da49e4c980ffb906e3c017f8e68863

Wedge009:
Backport #6060 if it's merged.

Wedge009:
#6067 6067

Wedge009:
#6073

Wedge009:
#6084

stevecotton:
#6085

Wedge009:
#6091

Wedge009:
#6095

Wedge009:
#6101

Wedge009:
#6108

Wedge009:
da3eac4ecb26a6f841ed0749d2787d5ee4c42055

Wedge009:
#6129 6129

Wedge009:
#6143

Wedge009:
#6152

Wedge009:
[...]

PR #5981, opened 2021-07-25
IoM: Add translation hint about the "Healing Keep"

stevecotton:
@knyghtmare I think this is needed because we're translating the role of that hex, rather than a terrain that acts like both a keep and an oasis.

### DIFF
--- a/data/multiplayer/scenarios/2p_Isle_of_Mists.cfg
+++ b/data/multiplayer/scenarios/2p_Isle_of_Mists.cfg
@@ -186,6 +186,8 @@
 
         [label]
             x,y = 16,9
+            # po: In the scenario Isle of Mists, there's an allied AI unit with heals+8.
+            # po: This label is added to the hex that she starts on, and she doesn't move.
             text = _"Healing Keep"
         [/label]
 


knyghtmare:
> both a keep

I don't think it works like a keep. The 6 adjacent hexes are all "flat" terrain, so players cannot use 16,9 as a recruitment station. The oasis role does not apply either as well, the hex itself does nothing, the Elvish Lady heals+8/cures 6 adjacent units at the start of side 5's turn.

stevecotton:
I agree with you, I gave that as a false interpretation which this hint tries to avoid, but that puzzles me more about this. The word "keep" is literally in the text to be translated - the hint is to suggest translating it as "person who heals" instead of "terrain that's a keep and and oasis".

Pentarctagon:
She is a healer standing on a keep.

stevecotton:
Are there comments on the change itself? IIUC, all the comments so far are discussing text that's on Github but outside the Git commit itself.

Pentarctagon:
Did something come up indicating that a hint is needed?

stevecotton:
I was doing some of the easy bits of the German translation, met this text without the context of the scenario, and realised that I'd read it as the name of a terrain-type.

Pentarctagon:
How are po hints handled if the same text is used in two places within the same textdomain?  It seems like a bad idea to add scenario-specific context in a note whose text could end up being re-used elsewhere in the same textdomain if that same note would effectively be used for all uses of the text.

Though I'm also not sure what the difference here is - the original English text isn't making any distinction between whether it's the effect hex or the effect of a unit on the hex, so why would the German translation be doing so?

stevecotton:
> It seems like a bad idea to add scenario-specific context in a note whose text could end up being re-used elsewhere in the same textdomain if that same note would effectively be used for all uses of the text.

True in all respects, but changing it to "isle of mists^Healing Keep" would break string-freeze.

> so why would the German translation be doing so?

Because a literal translation would be much longer ("Heilender Burgfried" vs "Heilerin"), and would poke into other hexes on the map.
@Byteron should I have done it as "Heilender\nBurgfried" instead?

Wedge009:
This seems like another small change that's been hanging around for a while. Might a solution be to just change 'Healing Keep' to something like 'Healer' and flag it as another string change to back-port later? (#5973)

Pentarctagon:
@stevecotton Are we waiting on a response from @Byteron for this?

stevecotton:
I think I'll close without merging - if it really is a problem then another translation team will flag it.

PR #5986, opened 2021-07-26
Copyright update.

Pentarctagon:
This additionally:
* Makes all copyright notices identical aside from the starting year for Wesnoth-specific source files. Files not included: mariadbpp, lua, spirit po, xbrz, and bcrypt (crypt_blowfish).
* Removes all attribution from the files, since the vast majority of them are outdated or seemingly just outright incorrect. For example, I would guess that Dave is no longer the sole author of the majority of Wesnoth's current code.


stevecotton:
What are the aim and the target audience of this?

IANAL, but I feel doing this at the start of a stable branch doesn't give any protection - if either wording is weaker then someone wanting to avoid the license can simply take the weaker version.

If the aim is to improve the behavior of Vim's auto-detection of tabstops then that seems useful. (Vim sometimes looks at the indentation of the copyright, and guesses that our code style uses three spaces for indentation.)

From the perspective of Git history, this is going to be an annoying commit that appears in any filtered-by-filename view of the history.

Pentarctagon:
The aim is that the copyright year is outdated and should be kept up to date.  Making everything identical then makes it easier to script next time.

Pentarctagon:
Also worth noting that the license of these files doesn't change - they're GPL v2+ before this commit and remain so.  So there is no "weaker wording" or anything like that, especially since I don't have the authority to change the copyright of most of these files in the first place.

irydacea:
Removing the attributions doesn't seem like a good move at all.

Pentarctagon:
They don't give up to date or even accurate information is nearly all cases.  Whether the attributions are there or not, if someone wants to know which people have contributed to the file, the answer will always be "check the git history and see if you can figure out who everyone is".

irydacea:
It's more of a principle thing, really. I don't think I've ever heard of this ever being done anywhere else before outside of a copyright transfer via CLA or the like.

Pentarctagon:
What about having it say something generic like "by multiple authors"?

irydacea:
The point isn't what it will say but rather what it used to say.

Pentarctagon:
I don't follow.  It didn't previously say anything useful.

irydacea:
Copyright statements generally don't get removed regardless of their perceived usefulness is what I'm saying.

irydacea:
It's been brought to my attention that this PR may in fact be in violation of [US copyright law](https://www.law.cornell.edu/uscode/text/17/1202).

> (b) Removal or Alteration of Copyright Management Information.—No person shall, without the authority of the copyright owner or the law—
> (1) intentionally remove or alter any copyright management information,
> (2) distribute or import for distribution copyright management information knowing that the copyright management information has been removed or altered without authority of the copyright owner or the law, or
> (3) distribute, import for distribution, or publicly perform works, copies of works, or phonorecords, knowing that copyright management information has been removed or altered without authority of the copyright owner or the law,
> knowing, or, with respect to civil remedies under section 1203, having reasonable grounds to know, that it will induce, enable, facilitate, or conceal an infringement of any right under this title.

Pentarctagon:
Though that would also seem to imply that we can't update the years in the copyright notice either without the agreement of all the people who contributed to the file, which for most source files would likely mean we just can't update it.

irydacea:
That's definitely a bit of a grey area but it's still less questionable than removing names outright.

irydacea:
Also, for reference, I do not consent to my name being removed.

Pentarctagon:
Well, I guess I'm off to re-add a whole bunch of wrong information back...

CelticMinstrel:
I think this argument is kind of stupid, honestly. Then again, I'm one of the people who never added my name to the header in the first place…

I think Pentarctagon probably has a good point on the copyright information being wrong in some cases, and that it should probably be removed in those cases. I think I've seen cases where someone copy-pasted the copyright header from another file into their new file without updating the attribution, so if you can prove that the attributed person is not the person who first created the file, I think removing the attribution is the correct choice.

Having attribution at all is kind of stupid, because it's attaching the person's name to something that will evolve beyond their original code and, eventually, may not even resemble the code they originally wrote.

Suggestion: What about tacking an "et al" on the end of any attribution? This makes it subtly less wrong than leaving the attribution untouched.

irydacea:
"This argument is kind of stupid" is not a terribly persuasive counterargument to make, especially when like you said you specifically chose not to assign copyright to yourself for new code. This isn't the right place for an anti-copyright political campaign, at any rate.

It's also worth remembering that copyright does not necessarily equal authorship, and generally speaking the people who need to know about either understand this.

Pentarctagon:
> I think I've seen cases where someone copy-pasted the copyright header from another file into their new file without updating the attribution

Also there's https://github.com/wesnoth/wesnoth/blob/674fda85b7d74ea45d7db4ef842616097b58f0a1/src/boilerplate-header.cpp

Pentarctagon:
Also, uh...
https://github.com/wesnoth/wesnoth/blob/674fda85b7d74ea45d7db4ef842616097b58f0a1/src/desktop/apple_notification.mm#L2
What.

CelticMinstrel:
> "This argument is kind of stupid" is not a terribly persuasive counterargument to make

I wasn't really intending to make an argument, just stating an opinion. I _did_ give proper reasoning on the parts where I _was_ making an argument - the second and final paragraphs.

stevecotton:
> Also, uh...
> 
> https://github.com/wesnoth/wesnoth/blob/674fda85b7d74ea45d7db4ef842616097b58f0a1/src/desktop/apple_notification.mm#L2
> 
> What.

Probably a GSoC student, or someone working for Google when that firm were giving 20% of their employees' time to other projects. The latter would be copyright to Google, surprisingly I can't find in a web-search whether GSoC assigned the copyright to Google.

GSoC with multiple years would be the result of people running scripts that bump the year counter.

stevecotton:
I feel you should ask @Alarantalara before stripping that copyright statement.

(Posted as a new comment, because I don't know whether Github pings people added when an old comment is edited.)

PR #4390, opened 2019-09-27
Replace hot-key markers with initials

Wedge009:
Suggested in https://github.com/wesnoth/wesnoth/issues/1152#issuecomment-535022785 and mitigates #1499.

PR #5987, opened 2021-07-26
Resolve title-screen vs main menu inconsistency

Wedge009:
Discussed in #4390 and originally introduced in 7c8c3114.

I suppose can back-port to 1.16 when there's no string freeze.

### DIFF
--- a/src/gui/dialogs/preferences_dialog.cpp
+++ b/src/gui/dialogs/preferences_dialog.cpp
@@ -761,8 +761,8 @@ listbox& preferences_dialog::setup_hotkey_list()
 	t_string& row_is_g_markup = row_data["lbl_is_game"]["use_markup"];
 	t_string& row_is_e        = row_data["lbl_is_editor"]["label"];
 	t_string& row_is_e_markup = row_data["lbl_is_editor"]["use_markup"];
-	t_string& row_is_t        = row_data["lbl_is_titlescreen"]["label"];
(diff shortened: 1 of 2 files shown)


src/gui/dialogs/preferences_dialog.cpp
```diff
@@ -761,8 +761,8 @@ listbox& preferences_dialog::setup_hotkey_list()
 	t_string& row_is_g_markup = row_data["lbl_is_game"]["use_markup"];
 	t_string& row_is_e        = row_data["lbl_is_editor"]["label"];
 	t_string& row_is_e_markup = row_data["lbl_is_editor"]["use_markup"];
-	t_string& row_is_t        = row_data["lbl_is_titlescreen"]["label"];
-	t_string& row_is_t_markup = row_data["lbl_is_titlescreen"]["use_markup"];
+	t_string& row_is_t        = row_data["lbl_is_mainmenu"]["label"];
+	t_string& row_is_t_markup = row_data["lbl_is_mainmenu"]["use_markup"];
```
stevecotton, line 765: There's a `t` in the variable names.

I'm not sure if the `lbl_is_titlescreen` needs to change - it already isn't a user-facing string.

Wedge009: Thanks for catching that - I didn't include `*.cfg` in my text search.

Even if it's not user-visible, source code should be consistent. Inconsistency will leave open the possibility for confusion later on, otherwise.

Should be updated in a few minutes.

PR #5988, opened 2021-07-26
Add a script to automatically update the copyright notice in source files.

Pentarctagon:


utils/update_copyright.py
```diff
@@ -0,0 +1,52 @@
...
+import re
+import sys
+
+previous_year = str(date.today().year-1)
+if len(sys.argv) == 2:
+    previous_year = sys.argv[1]
+
+new_year = str(date.today().year)
+
+previous_notice = "^\tCopyright \(C\) 20[0-9]{2} - "+previous_year+"$"
```
CelticMinstrel, line 19: Is this really right? It only matches start years in the range [2000,2100), right?

Also as a side note, I generally recommend using raw string syntax for regular expressions.

utils/update_copyright.py
```diff
@@ -0,0 +1,52 @@
...
+            skip = True
+            break
+    if skip:
+        continue
+
+    for file in files:
+        if re.search(extensions, file):
+            lines = []
+            found = False
+            with open(root+"/"+file, 'r') as f:
```
CelticMinstrel, line 39: You should use `os.path.join` to build paths.

utils/update_copyright.py
Pentarctagon: Wesnoth was initially created in 2003, so it wouldn't make much sense for there to be copyright notices for earlier years.

CelticMinstrel: Sure, but not supporting 2100 and above is also kind of silly when it's trivial to support it. I have no idea if Wesnoth will still be around in 2100, but all you have to do is change this to `[0-9]{4}` and it automatically works if it still is.

Pentarctagon: Mainly it's to catch some cases of typoes, like if someone typed 2201 instead of 2021.

Pentarctagon: Also for some reason
`r"^	Copyright (C) 20[0-9]{2} - "+previous_year+r"$"`
does not work while
`"^\tCopyright \(C\) 20[0-9]{2} - "+previous_year+"$"`
does work.

soliton-: Well, space and tab are different characters. If you want to match any whitespace you could use `\s`. `()` denotes a capture group but you want to match literal parentheses so you need to escape.

Pentarctagon: If I copy/paste that character into a text editor it gets treated as a tab.  No idea why it only shows up here as taking up a single space.

PR #5990, opened 2021-07-27
Apply "replay old saves" patch to the 1.14 branch

cooljeanius:
This applies the patch from this forums thread to 1.14: https://forums.wesnoth.org/viewtopic.php?f=10&t=53824&start=15
Note that I realize that 1.14 is closed, and that there won't actually be any further changes to it, but I just wanted to get this open so we can see what a successful application of the patch looks like, so that we can then use that for getting it rebased upon a more up-to-date branch next. Hence, I am marking this as a draft.


data/core/macros/names.cfg
```diff
@@ -173,6 +174,7 @@ suffix_male=af|al|lih|bair|sur|bi|lah|at|sim|ma|ih|yl|iq|raj|mam|man|ya|zid|riya
 #enddef
 
 #define LIZARD_NAMES
+    # Saurian name definition: no Context-free_grammar: male_name_generator/female_name_generator missing
```
CelticMinstrel, line 177: There _is_  context-free grammar for saurians - see the name_generator key. I'm not sure whether it's actually used though, could be a bug there.

cooljeanius: yeah that part of the patch gets rejected when trying to apply it against trunk; maybe it was added sometime after 1.14?

src/savegame.cpp
```diff
@@ -635,69 +643,1183 @@ void ingame_savegame::write_game(config_writer &out) {
...
+std::string append_suffix_to_unit_type(std::string unit_type, config& cfg)
+{
+	const std::string suffix = convert_version_to_suffix(cfg);
+	return unit_type + suffix;
+}
+
+void update_start_unit_types(config& replay_start, config& cfg)
+{
+	const std::string suffix = convert_version_to_suffix(cfg);
+	if(not suffix.empty())
```
CelticMinstrel, line 680: We prefer to use symbols rather than names for operators; in this case it should be `!`. Not sure if there are other instances of this; see iso646.h for a list of search terms to replace.

src/savegame.cpp
```diff
@@ -635,69 +643,1183 @@ void ingame_savegame::write_game(config_writer &out) {
...
+				attack["defender_type"] = attack["defender_type"] + suffix;
+			}
+		}
+	}
+}
+
+void feral_trait_1_8_enter_hex(config& event)
+{
+	std::stringstream ss;
+	ss << "		name=enter_hex\n";
```
CelticMinstrel, line 776: These sorts of passages don't really belong in the source. The WML in each such function should be extracted into a .cfg file which the function then loads by name.

src/savegame.cpp
```diff
@@ -635,69 +643,1183 @@ void ingame_savegame::write_game(config_writer &out) {
...
+		{
+			if(event["name"] == "prestart")
+			{
+				if(config& fire_event = event.child("fire_event"))
+				{
+					if(fire_event["name"] == "place_units")
+					{
+						config& lua = event.child("lua");
+
+						std::string changes[] = {"if searched == actual then;if actual:sub(1, #searched) == searched  then",
```
CelticMinstrel, line 1103: This seems like a place where extracting to an external .lua file would be more readable, though I'm not quite sure what that file would look like…

src/savegame.cpp
```diff
@@ -635,69 +643,1183 @@ void ingame_savegame::write_game(config_writer &out) {
...
+					++count;
+					if(config& results = random.child("results"))
+					{
+						if(results["dies"] == "yes")
+						{
+							remove_additional_numbers = true;
+							break;
+						}
+					}
+					//if(remove_additional_numbers) command.remove_child("random", count);
```
CelticMinstrel, line 1476: Don't leave commented code in.

CelticMinstrel:
All the new functions in savegame.cpp that deal with converting really old saves should be extracted into one or more additional source files to avoid bloating the main savegame.cpp file. The only code for old saves that belongs in savegame.cpp is the main `convert_old_saves` function and any code for converting saves from 1.14 or 1.15.

You could make a separate savegame_1_x.cpp file for every version supported, but I imagine that might be a bit overly tedious. A single savegame_old.cpp file might be acceptable but would probably also end up getting bloated. I'm not sure what a good solution is here.

-----

I don't like the inclusion of all the old unit definitions here. They're a vast number of mostly-tiny files. I'm not entirely sure how to improve the situation, but one thing that could help is merging together smaller files, so that most of the old-version unit files end up with one file per race plus one or two larger files.

I also don't like that all the old unit definitions are automatically loaded. I think it should be possible to load them only if needed.

-----

This is just a general summary of my thoughts, and I might come up with more complaints or suggestions later.

src/savegame.cpp
```diff
@@ -635,69 +643,1183 @@ void ingame_savegame::write_game(config_writer &out) {
...
+			}
+		}
+	}
+}
+
+void update_pre_recruited_loyals(config& cfg)
+{
+	const std::string suffix = convert_version_to_suffix(cfg);
+	if(suffix.empty()) return;
+	if(config& replay = cfg.child("replay_start"))
```
CelticMinstrel, line 1136: I suspect many or most of these functions could be handled more cleanly with a WML patch file (of the sort created by `wesnoth --diff a b`). That'll obviously be quite a bit of work to create, but it would offload much of the burden of the update from the source code. It might even be more readable, though WML patch files aren't super readable either.

cooljeanius:
> All the new functions in savegame.cpp that deal with converting really old saves should be extracted into one or more additional source files to avoid bloating the main savegame.cpp file. The only code for old saves that belongs in savegame.cpp is the main `convert_old_saves` function and any code for converting saves from 1.14 or 1.15.
> 
> You could make a separate savegame_1_x.cpp file for every version supported, but I imagine that might be a bit overly tedious. A single savegame_old.cpp file might be acceptable but would probably also end up getting bloated. I'm not sure what a good solution is here.
> 
> I don't like the inclusion of all the old unit definitions here. They're a vast number of mostly-tiny files. I'm not entirely sure how to improve the situation, but one thing that could help is merging together smaller files, so that most of the old-version unit files end up with one file per race plus one or two larger files.
> 
> I also don't like that all the old unit definitions are automatically loaded. I think it should be possible to load them only if needed.
> 
> This is just a general summary of my thoughts, and I might come up with more complaints or suggestions later.

So... how should I go about trying to address your suggestions? Should I continue working on these changes on this branch, or should I focus on trying to get the patch to apply cleanly against master first before addressing your concerns?

CelticMinstrel:
> should I focus on trying to get the patch to apply cleanly against master first before addressing your concerns?

I'd say this is probably more important. Make sure it works at all before trying to address any of my concerns.

knyghtmare:
I thought 1.14 branch was not getting any more releases?

cooljeanius:
> I thought 1.14 branch was not getting any more releases?

it's not; I just based this off of that because it's where it applied cleanly, so I could have something to look at. Hence why I marked this as a draft.

CelticMinstrel:
For the record, you can change the base branch later by clicking Edit at the top of the page (the same place you change the title).

cooljeanius:
> For the record, you can change the base branch later by clicking Edit at the top of the page (the same place you change the title).

OK, so I just tried that, but it added so many extra commits that it became harder to review...

CelticMinstrel:
Yeah, if you do change the base branch, you need to actually change the commits to be based on that branch, usually by rebasing.

cooljeanius:
demario has a fresh version of this patch that would require reapplying from scratch, so I'm going to give up on this pull request and try opening a new one instead (not immediately tho; I might want to wait and see what else demario does)

PR #5991, opened 2021-07-27
Re-add attribution to copyright notices.

Pentarctagon:
These are mostly useless and outdated/wrong, but apparently it's probably illegal to remove them.

### DIFF
--- a/src/about.cpp
+++ b/src/about.cpp
@@ -1,5 +1,6 @@
 /*
 	Copyright (C) 2003 - 2021
+	by David White <dave@whitevine.net>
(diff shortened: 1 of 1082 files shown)


irydacea:
Why are some of them broken up after the year into a new line and others aren't?

Pentarctagon:
The line that ends with the year is the one that the script updates.

irydacea:
Is there really no better option so that it doesn't look like someone press enter too early?

Pentarctagon:
I don't see how it makes any difference to be honest.

CelticMinstrel:
There's definitely another option (make the script a bit more robust), but I also don't think it makes any difference.

Issue #5926, opened 2021-07-05
Tutorial: replay shows character selection dialog


stevecotton:
An issue similar to #5334 affects the tutorial (tested on the master branch's current commit). The character-selection dialog pops up during a replay, followed (if you select the other character) by an OOS.

From https://github.com/wesnoth/wesnoth/issues/5334#issuecomment-738131757 :

It's a problem with the campaign code, not the core `show_dialog`. Just added the bold bits to show_dialog's docs:

This function should be called in conjunction with `#wesnoth.synchronize_choice`, in order to ensure that only one client displays the dialog and that the other ones recover the same input values from this single client. **Even in single-player campaigns, synchronization is necessary otherwise the replay will be broken.**

Pentarctagon:
Would it be possible/make sense to have this be synchronized by default?  It seems like a strange API that will by default break all replays that use it.

CelticMinstrel:
There's no reason to synchronize a dialog that provides only information and no choice. Since `gui.show_dialog` can easily create either type of dialog, I think it makes sense to let the designer decide whether it needs to be synced. I don't think there's a way of automatically detecting whether a dialog offers a choice to the player.

Pentarctagon:
Is there a downside to synchronizing by default?  Sure, synchronizing a dialog that has no choice on it does nothing, but it also prevents bugs like these from happening at all.

stevecotton:
I'd be more tempted to write a nice wrapper than to try changing the behavior right now.

stevecotton:
@CelticMinstrel as you've worked on the tutorial-based example, are you going to fix the tutorial and SotA too?

There are also dialogs that shouldn't be synchronized; admittedly these are likely to be rare enough that making them add a non-default option seems reasonable

 Ageless has a right-click option so that players can read the changelog when it's not their turn:
https://github.com/ProditorMagnus/Ageless-for-1-14/blob/master/data/era_events/unsynced_menu.cfg

I'm not planning much further development on SXC, but if I did then I'd be looking at making the initial turn's shopping event asynchronous, instead of needing each player to wait in turn.

CelticMinstrel:
I can fix the tutorial. I'm not quite sure about SotA but I can look into that too.

stevecotton:
> I can fix the tutorial. I'm not quite sure about SotA but I can look into that too.

Will you have time to do these, or should I take them?

CelticMinstrel:
[...]

Issue #5334, opened 2020-12-03
SotA: Zombie dialog interrupts replays


stevecotton:
### Game and System Information
 - Reported by @doofus-01 in https://r.wesnoth.org/p660620
 - Reproducible on 1.15.7+dev built from 7c4ce9dd0fb50e8b16876f27538322b242126747

### Describe the bug
When the player recruited a zombie in the replay, the zombie dialog pops up and asks the replay-viewer which zombie to recruit.

### To Reproduce
* Load the savefile attached to @doofus-01 's forum post
* Play the first turn

### Other stuff
As #5237 notes, this dialog isn't necessary for SotA but is a nice example of how to write a Lua/GUI2 dialog - no matter whether we fix it or remove it, we should ensure that there is a fully-working example to point UMC developers at. @Pentarctagon, I think this should be a blocker for a new release.

Time issues: I can't commit more time to Wesnoth this month, and assume neither @beetlenaut nor @gfgtdf can either.

Pentarctagon:
Is this a problem with that particular dialog or with the `show_dialog` function?

stevecotton:
It's a problem with the campaign code, not the core `show_dialog`. Just added the bold bits to show_dialog's docs:

This function should be called in conjunction with `#wesnoth.synchronize_choice`, in order to ensure that only one client displays the dialog and that the other ones recover the same input values from this single client. **Even in single-player campaigns, synchronization is necessary otherwise the replay will be broken.**

Pentarctagon:
Based on https://github.com/wesnoth/wesnoth/blob/7bdd72886ddd27c431d94e342f95834a117513ec/data/campaigns/Secrets_of_the_Ancients/utils/zombie_recruit_dialog.lua#L152 this was apparently committed while known to not work, in fact.

From a quick look it doesn't seem as though the fix is especially difficult, at least.

Pentarctagon:
@stevecotton Also the version should be 1.15.7+dev, seeing as 1.15.8 hasn't been release yet :P

stevecotton:
@CelticMinstrel does 726b5458f961f7e2c02de91b66b90dbfed5ddcfd fix this?

I've taken a quick look at the code and think the issue is still open, but haven't spent time testing yet.

CelticMinstrel:
That commit doesn't really change anything; it just moves the WML code out to a WML file for easier editing.

stevecotton:
This one can (and should) still be fixed during the string freeze.

[...]

PR #5994, opened 2021-07-28
Tutorial S01 and SotA: Synchronise calls to gui.show_dialog

stevecotton:
Two related bugfixes, put into one PR because, if we find things to change on one, there may be things to change on the other one too. Both will backport cleanly and are already tested on the 1.16 branch.

Tutorial S01: Synchronise the choice of character
---

This avoids showing the dialog during replays, and thus avoids an
OOS if the person watching the replay chooses the other character.

Fixes (on master) #5926.

SotA: Synchronise the zombie recruitment dialog
---
    
This dialog works by setting variables in the postshow function,
so the code had to be moved around to ensure that those variables
were synchronised, not just the return value of the dialog.
    
The message about not having enough gold isn't necessary for the
replay, so that gets moved into the synchronised code too.

Fixes (on master) #5334.

### DIFF
--- a/data/campaigns/Secrets_of_the_Ancients/utils/zombie_recruit_dialog.lua
+++ b/data/campaigns/Secrets_of_the_Ancients/utils/zombie_recruit_dialog.lua
@@ -87,12 +87,17 @@ wml.variables["recruitedZombieType"] = "cancel" -- default value
 if zExists==false then
     gui.show_prompt("", _ "There are no corpses available.", "")
 else
-    --TODO: this seems not to be synced.
-    local returned = gui.show_dialog(wml.get_child(zombie_recruit_dialog, 'resolution'), preshow, postshow)
-    if  returned ~= -2 and sides[1].gold  < recruitCost then
-        gui.show_prompt("", _ "You do not have enough gold to recruit that unit", "")
-    elseif returned ~= -2 and (sides[1].gold ) >= recruitCost then
-        wml.variables["recruitedZombieType"] = recruitedType
-        wml.variables["recruitedZombieCost"] = recruitCost
-    end
+    local result = wesnoth.sync.evaluate_single(function()
+        local res = gui.show_dialog(wml.get_child(zombie_recruit_dialog, 'resolution'), preshow, postshow)
+        if res == -2 then
+            return {recruitCost = 0, recruitedType = "cancel"}
+        end
+        if sides[1].gold  < recruitCost then
+            gui.show_prompt("", _ "You do not have enough gold to recruit that unit", "")
+            return {recruitCost = 0, recruitedType = "cancel"}
+        end
+        return {recruitCost = recruitCost, recruitedType = recruitedType}
+    end)
+    wml.variables["recruitedZombieType"] = result.recruitedType
+    wml.variables["recruitedZombieCost"] = result.recruitCost
 end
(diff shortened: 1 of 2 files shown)


stevecotton:
The build failure seems to be a problem with the build itself - there's no log messages from the build step.

The SotA fix is now ready, and I think both should go in one PR - if we find something wrong with the SotA fix then it might be something that needs the Tutorial fix to change too.

data/campaigns/Secrets_of_the_Ancients/utils/zombie_recruit_dialog.lua
```diff
@@ -87,12 +87,18 @@ wml.variables["recruitedZombieType"] = "cancel" -- default value
...
-    end
+    local result = wesnoth.sync.evaluate_single(
+        function()
+            local res = gui.show_dialog(wml.get_child(zombie_recruit_dialog, 'resolution'), preshow, postshow)
+            if res == -2 then
+                return {recruitCost = 0, recruitedType = "cancel"}
+            end
+            if sides[1].gold  < recruitCost then
+                gui.show_prompt("", _ "You do not have enough gold to recruit that unit", "")
+                return {recruitCost = 0, recruitedType = "cancel"}
```
CelticMinstrel, line 98: Does putting "cancel" in the variable cause potential problems anywhere else?

data/campaigns/Secrets_of_the_Ancients/utils/zombie_recruit_dialog.lua
```diff
@@ -87,12 +87,18 @@ wml.variables["recruitedZombieType"] = "cancel" -- default value
...
+            end
+            return {recruitCost = recruitCost, recruitedType = recruitedType}
+        end)
```
CelticMinstrel, line 101: This indentation _feels_ unbalanced. I'd probably prefer to move `function()` to line 90 and push everything back one level. Alternatively, maybe moving the closing parenthesis to its own line?

data/campaigns/Secrets_of_the_Ancients/utils/zombie_recruit_dialog.lua
```diff
@@ -87,12 +87,18 @@ wml.variables["recruitedZombieType"] = "cancel" -- default value
...
-        gui.show_prompt("", _ "You do not have enough gold to recruit that unit", "")
-    elseif returned ~= -2 and (sides[1].gold ) >= recruitCost then
-        wml.variables["recruitedZombieType"] = recruitedType
-        wml.variables["recruitedZombieCost"] = recruitCost
-    end
+    local result = wesnoth.sync.evaluate_single(
+        function()
+            local res = gui.show_dialog(wml.get_child(zombie_recruit_dialog, 'resolution'), preshow, postshow)
+            if res == -2 then
+                return {recruitCost = 0, recruitedType = "cancel"}
```
CelticMinstrel, line 94: Feels slightly unfortunate that this line is duplicated… but it's only one line so I guess it's fine.

stevecotton:
It's already used as a guard value on line 85.

Pentarctagon:
Both of the fixes were tested, I assume?

stevecotton:
> Both of the fixes were tested, I assume?

Yes, on both master and 1.16.

data/campaigns/Secrets_of_the_Ancients/utils/zombie_recruit_dialog.lua
stevecotton: It's already used as a guard value on line 85.

Pentarctagon (approved):

data/campaigns/Secrets_of_the_Ancients/utils/zombie_recruit_dialog.lua
stevecotton: The latest push does this whitespace change (only).

Issue #5544, opened 2021-02-15
HDPI fonts and broken dialogs during the new-player journey


stevecotton:
### Describe the bug
I'm going through the new-player journey, seeing what looks wrong on a HDPI screen up to the point that the player completes the tutorial and starts the next campaign. This carries on from #5185, which was that the preferences dialog doesn't fit properly with HDPI font sizes.

- [ ] the campaigns menu looks bad (horizontal scrollbar) but is usable for selecting a campaign by title, if you already know which one to play
- [ ] the campaigns menu is far too small for deciding which campaign to play - only 6 fit in the left-hand pane, and for most campaigns the description doesn't fit in the right-hand pane
- [ ] the tutorial's Select Character dialog is too small, cutting the buttons off
- [ ] the tutorial's middle-of-screen print statements are too small
- [ ] in the Attack dialog's Damage Calculations subdialog, everything is readable but the bar-graphs are small with very small text
- [ ] the "Keep" label overlaps with print statements such as the "Move Li'sar to the keep" and "Attack the quintain with your fighters" - probably because the map is centered with a large border around the map. The large border is understandable, but the overlapping text should be fixed; this will get worse when the print statements' text size is improved.
- [ ] in tutorial S2, the Recruit dialog's left-hand pane gets a vertical scrollbar when selecting a shaman - it's too small for a unit with 1 ability and 2 attacks
- [ ] the Advance Unit dialog is small, similar to the Recruit dialog
- [ ] at the start of the next campaign, clicking on most of the story screen does nothing. The story screen's "Skip" and arrow buttons are small and far from the centers of attention, so the mouse feels unresponsive. Clicking on the picture should probably advance the story, the same as pressing the spacebar (might be handled as part of #4817 )
- [ ] in tutorial S2, "I hope my unit can survive the counterattack" is shown even if the orc dies during that combat

### Game and System Information
Local build of 1.15.9+dev, Debian Linux.

Full details in https://github.com/wesnoth/wesnoth/issues/5185#issuecomment-774196328 , the monitor settings are:
SDL video drivers: [x11] wayland KMSDRM KMSDRM_LEGACY dummy
Window size: 3840x2078
Screen refresh rate: 60
Screen dots per inch: 281.895966x282.804138
Screen dpi scale factor: 2.936416x2.945876

stevecotton:
[...]

PR #5996, opened 2021-07-30
Tutorial S01: HDPI support

stevecotton:
Remove the maximum size for the character-select dialog, the window will
automatically choose the minimum size that fits the text.

Extend the map 1 hex south. This means that, if the full map fits on screen,
the horizontal center of the map (where the print statements appear) is between
the keep and the south village, thus their labels don't overlap the print
statements.

This backports cleanly (and has been tested on 1.16), it fixes 2 of the points in #5544.

Placement of `[print]` on HDPI:
image

The character selection still looks good in a minimum-sized window:
image

### DIFF
--- a/data/campaigns/tutorial/gui/character_selection.cfg
+++ b/data/campaigns/tutorial/gui/character_selection.cfg
@@ -1,8 +1,6 @@
 #textdomain wesnoth-tutorial
 
 [resolution]
-    maximum_height = 250
-    maximum_width = 400
     [helptip]
         id="tooltip_large"
     [/helptip]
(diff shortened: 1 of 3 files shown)


cooljeanius:
Looks good! Are changes like these going to be required all throughout the scenario files?

stevecotton:
Thanks. Hopefully this is the only scenario that's going to need a map-size change.

Many of the GUI dialogs need some fixing for HDPI, but those are mostly parts of core; the tutorial one of the few campaigns that define their own GUI dialogs. SotA's walking-corpse recruitment dialog would benefit from being bigger, but it doesn't look bad as-is. I haven't tested World Conquest yet, but it seems that is already doing something with a `#define FORMULA_GUI_SCALE_RESOLUTION`.

Pentarctagon (approved):

PR #6000, opened 2021-07-31
Update Dirkfang.cfg

MXB2001:
Sound plays at -800 long before the missile anim begins and ends. I believe this should fix it to play at the correct time.

### DIFF
--- a/data/core/units/nagas/Dirkfang.cfg
+++ b/data/core/units/nagas/Dirkfang.cfg
@@ -63,6 +63,7 @@
         [/filter_attack]
         start_time=-800
         missile_start_time=-150
+        sound_start_time=-150
         [if]
             hits=yes
             [missile_frame]
@@ -72,8 +73,10 @@
             [/missile_frame]
             [frame]
                 image="units/nagas/dirkfang-throw[1~10].png:[90,100,120,150,150,100,100,130,130,120]"
-                sound=throwing-knife.ogg
             [/frame]
+            [sound_frame]
+                sound=throwing-knife.ogg
+            [/sound_frame]
         [/if]
         [else]
             hits=no
@@ -86,8 +89,10 @@
             [/missile_frame]
             [frame]
                 image="units/nagas/dirkfang-throw[1~10].png:[90,100,120,150,150,100,100,130,130,120]"
-                sound=throwing-knife-miss.ogg
             [/frame]
+            [sound_frame]
+                sound=throwing-knife-miss.ogg
+            [/sound_frame]
         [/else]
     [/attack_anim]
     [female]


stevecotton:
The CI failures are an unrelated error that's since been fixed on master, so shouldn't matter for merging this PR. I've checked locally that `wmlindent` and `wmllint` don't change the file.

The sound effect sounds to me like metal on metal, so the sound of a weapon being unsheathed rather than being thrown. For this reason, having it play at -800 seems better to me. However, -150 looks good to me too as my brain just goes "that's a sound effect associated with the projectile".

doofus-01:
I don't remember what `throwing-knife.ogg` sounds like, but assuming that's good, can't this just use the `SOUND:HIT_AND_MISS` macro, like the melee attack?  Would be more consistent, and mass updating would be easier.  I think the current (pre-PR) config is just sloppy copy-paste of some old code.

stevecotton:
@doofus-01 I like the suggestion to use the macro, however my main question is about the sound itself.

* The Orcish Assassin's throwing animation is only two frames, with the delay between unsheathing and throwing being just 100ms.

* The Dirkfang has a 650ms sequence of preparing for the throw. I've just realised that the weapon isn't unsheathed during that, the unit already has it in-hand in its idle pose.

MXB2001:
Yes the reason I submitted this was watching the unit in action. It plays the sound, which is followed by lengthy silence as the animation wraps up. It's just like a badly lip synched foreign film. : - ) I tested the change on my own copy of the game now, and it looks and sounds ok.

stevecotton:
Yes, but my point was that `throwing_knife.ogg` seems to be the sound of unsheathing the weapon, not the sound of it traveling through the air. Syncing it to the projectile's movement would be like syncing dialog to the wrong actor's lips.

The Ringcaster and Zephyr are using `hatchet.wav`, which sounds more like something traveling through the air.

MXB2001:
Hmm, never noticed. I hate to have to contradict you though... : - ) I just checked the files out and there are two problems with the your theory. One is that there is a miss file... That implies a throwing sound. Second is that on listening I can heard that you are correct, the sound is of a knife being unsheathed but directly followed by the sound of the knife rotating through the air followed by a hit thunk for the hit file and nothing for the miss file. As you said the guy is holding the chakram and not unsheathing it so this was definitely not the right sound file. I also think hatchet would likely have been a better choice. In any case, I offer my quick and dirty fix as a stop gap measure until someone can overhaul it properly.

doofus-01:
Overhaul is a bit of an overstatement, but we can certainly fix it.  I'll try to get some work on animations done this weekend, including this.  Thanks.

doofus-01:
The timing should be fixed now, even if the exact sound isn't agreed on yet.

PR #6012, opened 2021-08-08
August 8th animations

doofus-01:
- try to fix dirkfang chakri sound (#6000)
- ringcaster defense anim
- ringcaster melee attack anim
- ringcaster ranged attack anim


stevecotton:
Whichever sound these use, the Zephyr should probably use it too.

However, the point that I've said repeatedly in #6000 is that `throwing-knife.ogg` isn't the right sound.

doofus-01:
> However, the point that I've said repeatedly in #6000 is that throwing-knife.ogg isn't the right sound.

I saw that, but until someone provides the right sound, it seems OK to me.

doofus-01:
I won't get to the Zephyr today, so this is basically just a Ringcaster animation PR.  There will have to be a future naga animation PR.  I asked LordBob about the Ophidian portraits, so we know what the ranged weapon is really supposed to be.  If I don't hear back within the next week or two, I'll have to run with the short javelin thing (jarid).

I'll merge this PR once CI passes.

PR #6014, opened 2021-08-11
[world conquest] [master/1.17] Replacement of khalifate unit types with dunefolk unit types

knyghtmare:
```
Arif -> Dune Soldier

Hakim -> Dune Herbalist
Jundi -> Dune Rover

Khaiyal -> Dune Piercer

Naffat -> Dune Burner

Rami -> Dune Rider
```

I have also substituted the "Dune Piercer" with "Dune Rider", since it's a non-existent unit in 1.16 and master.

### DIFF
--- a/data/campaigns/World_Conquest/era/factions/The_Gang.cfg
+++ b/data/campaigns/World_Conquest/era/factions/The_Gang.cfg
@@ -13,7 +13,7 @@
             commanders=Rebels,Northerners
             heroes=Drakes,Undead_All
             deserters=Loyalists,Knalgans,Dune Herbalist
-            deserters_names={STR_HAKIM}+", "+{STR_LOYALISTS}+{STR_AND}+{STR_KNALGAN}
+            deserters_names={STR_DUNE_HERBALIST}+", "+{STR_LOYALISTS}+{STR_AND}+{STR_KNALGAN}
             {WC_II_PAIR "Troll Whelp" "Elvish Fighter"}
             {WC_II_PAIR "Orcish Grunt" "Orcish Grunt"}
             {WC_II_PAIR "Elvish Archer" "Elvish Archer"}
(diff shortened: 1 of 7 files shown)


Pentarctagon:
Give the textdomains are the same and that those macros are apparently used only in one place, could you also just delete the macros and directly put the unit names there.  So `{STR_DUNE_SOLDIER}` to `_ "Dune Soldier"` for example.

Maybe there was some specific reason why gfgtdf did things like this, but having macros for things that are only used once and especially for tiny things like this really only serves to make the code harder to understand since now you need to look at multiple files instead of just one file.

CelticMinstrel:
In order to do what you suggest, every use of one of these macros (potentially) changes from 1 line to 3 lines, because you need to change the textdomain, use the string, and then change the textdomain back.

By the way, `{STR_AND}` is not valid either. I'd recommend replacing it with `", "` for now until we can do it properly.

Pentarctagon:
`#textdomain wesnoth-units` is set above both the unit name macros and at the top of the faction files.

Pentarctagon:
I suppose it might be better to wait until later to fix all the string concatenation and single use macros.  It looks like this is ready to merge otherwise?

knyghtmare:
> I suppose it might be better to wait until later to fix all the string concatenation and single use macros. It looks like this is ready to merge otherwise?

At the current state, it just replaces the Dunefolk references in the defines and a bug where a non-existent unit type was being added in if the era was in use.

I do intend to make a follow-up commit to fix the strings and all.

CelticMinstrel:
Is it worth making a follow-up commit to fix just these strings? There's a much bigger issue with strings in WC, and it probably can't be solved for 1.16.0, so it might make sense to do all the string fixes at the same time.

knyghtmare:
> There's a much bigger issue with strings in WC, and it probably can't be solved for 1.16.0, so it might make sense to do all the string fixes at the same time.

I do agree. The primary concern should be all the issues in the "Enable World Conquest" project.

Pentarctagon:
I'll go ahead and merge/backport this later today then unless there are objections.

Issue #5228, opened 2020-10-24
variable 'wc2_artifacts' must be assigned before being used


doofus-01:
### Game and System Information
The Battle for Wesnoth version 1.15.6+dev (9e91ddb1fee-Modified)
Distribution channel: Default

### Describe the bug
Started a local MP game, chose World Conquest for the era.  Recruited a couple times, right-clicked on one of the units, got a Lua error:
``
error scripting/lua: campaigns/World_Conquest/lua/era/era.lua:357: variable 'wc2_artifacts' must be assigned before being used
stack traceback:
        [C]: in function '.error'
        lua/ilua.lua:136: in metamethod '__index'
        campaigns/World_Conquest/lua/era/era.lua:357: in function <campaigns/World_Conquest/lua/era/era.lua:349>
``

Pentarctagon:
@doofus-01 does this error still appear?

knyghtmare:
Confirmed on: Battle for Wesnoth v1.15.14+dev (8679f0573f8-Modified) x86_64

20210626 02:48:57 error scripting/lua: campaigns/World_Conquest/lua/era/era.lua:360: variable 'wc2_artifacts' must be assigned before being used
stack traceback:
	[C]: in function '.error'
	lua/ilua.lua:136: in metamethod 'index'
	campaigns/World_Conquest/lua/era/era.lua:360: in function <campaigns/World_Conquest/lua/era/era.lua:352>

CelticMinstrel:
Hmm… it's assigned in game_mechanics/_load.lua, so there are four possible problems that I can think of:

1. An error occurs in shared_utils/wml_converter.lua, game_mechanics/ability_events.lua, or game_mechanics/artifacts.lua
2. game_mechanics/artifacts.lua returns nil
3. Something later assigns nil to the variable
4. game_mechanics/_load.lua is never called

The second or third seem the most likely…

Pentarctagon:
It looks like `#4` is the cause?  Adding just something like `wesnoth.log("error", "hi")` to the top of `_load.lua` results in "hi" being logged when starting a WC scenario but not when starting a regular scenario with the WC era.

`_load.lua` is called from:
```
lua/main.lua:4:wesnoth.dofile("./game_mechanics/_load.lua")
lua/campaign_main.lua:7:wesnoth.dofile("./game_mechanics/_load.lua")
```
`main.lua` is then only ever called from the WC scenario file:
```
[...]

PR #6015, opened 2021-08-11
[world conquest] World Conquest Era is only available for the campaign

knyghtmare:
Until a proper solution is found for #5228 , this will have to do. It suppresses the appearance of the era in the era selection list for other scenarios, preventing the bug from happening.
Hmm, I guess the [campaign] tag should support the `allow_era=` key in future but for the moment, this will do.

Players can play WC with Default, Default+Dunefolk, Age of Heroes, AOH+Dunefolk and UMC eras for the time being.

### DIFF
--- a/data/campaigns/World_Conquest/era/era.cfg
+++ b/data/campaigns/World_Conquest/era/era.cfg
@@ -8,6 +8,10 @@
         id= "{STR_ERA_ID_WC_II}"
         name= {STR_ERA_NAME_WC_II}
         description= {STR_ERA_DESCRIPTION_WC_II}
+
+        allow_scenario=WC_II_1p,WC_II_2p,WC_II_3p
+        hide_help=yes
+
         require_era=no
         # addon_min_version="8.2"
 


ProditorMagnus:
It should use [load_resource] like other addons.

ProditorMagnus:
https://github.com/wesnoth/wesnoth/blob/4e276e91cea8a3952d97fbc66e5d1821b15f3597/data/campaigns/World_Conquest/era/era.cfg#L36 suggests campaign and era should work separately.

knyghtmare:
> https://github.com/wesnoth/wesnoth/blob/4e276e91cea8a3952d97fbc66e5d1821b15f3597/data/campaigns/World_Conquest/era/era.cfg#L36
> 
> suggests campaign and era should work separately.

That is what is commented. On playtesting, bug #5228 happens. Better to keep the era disabled for the 1.16 cycle.

Pentarctagon:
The CI failure is due to the WML schema missing an [era] attribute, which has been fixed in 1dcfee8fb5cf901768b697dc8088c202cb6abd9f.

PR #6020, opened 2021-08-14
Format times in the add-on manager using utils::format_time_summary

stevecotton:
As it's a date, this needs to be a translatable string, and one that can be backported to 1.16 (so using a string that's already in 1.16). Possible choices for doing that are:
* using the utils::format_time_summary() function
* using one of the strings already used by the utils::format_time_summary() function
* using the strings from src/save_index.cpp's save_info::format_time_local()

DRAFT: The layout is broken in a way that causes these dates to often get ellipsed. That isn't a new issue introduced by the main (second) commit here, it's already an issue in 873946142fe73f4bf01d758e5cc4ad54182503a1 which will be addressed by @Vultraz's layout changes.

### DIFF
--- a/src/gui/dialogs/addon/manager.cpp
+++ b/src/gui/dialogs/addon/manager.cpp
@@ -956,8 +956,8 @@ static std::string format_addon_time(std::time_t time)
 		std::ostringstream ss;
 
 		const char* format = preferences::use_twelve_hour_clock_format()
-			? "%B %d %Y, %I:%M %p"
-			: "%B %d %Y, %H:%M";
+			? "%Y-%m-%d %I:%M %p"
+			: "%Y-%m-%d %H:%M";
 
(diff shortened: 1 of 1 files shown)


Wedge009:
I appreciate this effort. I just did a check of what it looked like in 1.14 and it used ISO standard of YYYY-MM-DD HH:MM (according to 12 vs 24 hour advanced preference). I was surprised to check master and see the date as the American MMM DD YYYY, HH:MM.

With your change, I see DD MMM, so lost the year and time but I assume that's due to the limitation of re-using existing date strings. In all cases I tested against en_GB internal language selection.

Wedge009:
Is this still necessary? Since I wound up doing #6248 and c7cab90d5d3f385e42dcb48ef60a38c9b086bad5...

stevecotton:
Your change fixed it, no need for this.

PR #6029, opened 2021-08-16
A few new AI features

CelticMinstrel:
Mostly AI stuff, partly related to the eventual removal of FormulaAI.

Without FormulaAI, there was no longer a way to set actions for a specific unit. This adds a new way to do that. It's essentially just syntax sugar but it should be useful sometimes.

Mostly not back-portable, but probably all the commits with "fix" in the message can be backported.


mattsc:
I am not going to have time to go through the code in detail and/or actually test things in the near future.  However, I assume that you have tested that everything works and as this is for 1.17, I am not worried about that.  There is lots of time to shake out potential bugs.  I did go over the commits and looked through what they are meant to do (and had a quick scan over the code) and I am fine with the intent of these changes.  [As discussed on Discord/irc, there are some things that I would not have considered worth the effort if I had to do them, but now that they are done, there is not reason not to include them either.]

So if you're okay with this level of review, I'll approve this PR.

Issue #5891, opened 2021-06-24
World Conquest: final scenario incorrectly tries to generate another scenario


Pentarctagon:
After the “final battle” scenario is won, WC tries to generate an additional scenario despite the fact that this is supposed to the the end of the campaign which causes a crash to main menu.

Reported by @knyghtmare

knyghtmare:
Confirmed on Battle for Wesnoth v1.15.14+dev (b05215d8aa6-Modified) x86_64
```cs
20210712 00:46:53 info deprecation: wesnoth.set_end_campaign_text has been deprecated indefinitely.;  (Note: You should use wesnoth.scenario.end_text, wesnoth.scenario.end_text_duration instead in new code)
Invalid WML found: wesnoth.set_end_campaign_text has been deprecated indefinitely.;  (Note: You should use wesnoth.scenario.end_text, wesnoth.scenario.end_text_duration instead in new code)
~add-ons/World_Conquest/lua/map/main.lua:42: bad argument #-1 to 'dofile' (file not found)	stack traceback:
	[string ""]:8: in function <[string ""]:8>
	[C]: in field 'dofile'
	~add-ons/World_Conquest/lua/map/main.lua:42: in function '.wc_ii_generate_scenario'
	(...tail calls...)
	[C]: in function '.xpcall'
	[string ""]:8: in main chunk
nil
20210712 00:46:56 error mapgen: Error when running lua_map_generator create_scenario.
The generator was: World Conquest 1p
bad return value:
  expected a config (table), found a nil20210712 00:46:56 error general: Map generator error: Error when running lua_map_generator create_scenario.
The generator was: World Conquest 1p
bad return value:
  expected a config (table), found a nil
```

stevecotton:
Looks like a core bug rather than a campaign bug - `[endlevel]end_text=` isn't deprecated, but over in data/lua/wml/endlevel.cfg we have:

```
	if end_text or end_text_duration then
		wesnoth.set_end_campaign_text(end_text or "", end_text_duration)
	end
```

stevecotton:
The campaign is fixed in #6040 (master, will merge when CI passes) and #6039 (1.16).
The deprecation warning is fixed in 9daa10a9f27c5a95520e871417bbd72aa52aa688 for master, which doesn't yet have a backport.

stevecotton:
I regret trying to handle the deprecation issues in this bug report, and have split them to #6042.

The current issue is fixed by #6040 (master) and #6039 (1.16).

PR #6039, opened 2021-08-23
WC: End the campaign at the end of the final scenario

stevecotton:
Fixes the campaign part of #5891 (which is the main part of that bug).


CelticMinstrel:
Wow, that's an amazingly simple fix. Cool!

PR #6040, opened 2021-08-23
WC: End the campaign at the end of the final scenario

stevecotton:
Fixes the campaign part of #5891.

(forward-port of #6039)


Pentarctagon:
What do you mean by "the campaign part"?  Is there another part of WC that would be affected by this?

stevecotton:
I meant "not the deprecation warning".

Issue #3600, opened 2018-10-04
Terrain rendered on top of unit


GregoryLundberg:
From https://forums.wesnoth.org/viewtopic.php?f=4&t=48903

The wooded terrain to the SE overlays the unit. Example from OP:
![alt text](https://forums.wesnoth.org/download/file.php?id=82001 "Terrain overlays unit")
Note how the green pine tree from the SE tile overlays the wolf's head.

doofus-01:
I think this is fixed, but someone else should confirm (1.16, not master)

doofus-01:
If this is duplicate of #1153, it is fixed in 1.16 (still needs to be ported to master)

Issue #1153, opened 2017-05-08
split forest graphics iinto two layers in order to handle some edge cases better (GNA #13242)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?13242)** by anonymous on 2009-03-22

After the scout was slowed by an elven shaman, his face was obscured behind a forest tile. Screen shot provided shows the end result. The artifact cleared up as soon as the scout moved.

(Reproduced on Ubuntu)
Release: 1.6
Priority: 3 - Low
Severity: 1 - Wish

wesnoth-bugs:
**Modified on 2009-03-22**

**alink wrote:**

> No screenshot attached :-/

wesnoth-bugs:
**Modified on 2009-03-24**

**ai0867 wrote:**

> The shield of the right dwarf fighter shows what is probably the same thing. Whether this is a bug or not I don't know.
> 
> 
> 
> http://wesnoth.org/start/1.6/images/start-4-full.jpg

wesnoth-bugs:
**Modified on 2009-04-10**

**boucman wrote:**

> hmm
> 
> 
> 
> this "works as designed" I'm not sure if/how to fix that
> 
> 
> 
> basically, to convey a feeling of "depth" some tiles are drawn in front of others (think of the elven tree for example) 
> 
> 
> 
> in that case the overlay of the bordering tile overspill a bit too far, but the tile is considered in front of the uni and its overlay is thus drawn in front.
> 
> 
> 
> this is an artistic bug, but I'm not sure it's worth fixing. I will forward to zookeeper for the time being, but he will probably mark as "won't fix"

wesnoth-bugs:
**Modified on 2009-04-14**

**zookeeper changed status:** None -> Postponed

**zookeeper wrote:**

> It would be possible to fix these glitches for terrains such as forests. However, it'd require splitting off all the forest graphics into at least two parts so that the top and bottom parts of the tile could layer differently and thus for example the offending little tree in this case would get drawn behind the unit.
> 
> 
> 
> However, it'd be a pretty big task for relatively little gain, so as a workaround I'd suggest just moving any trees which are far away from the hex (such as that one little tree) a bit closer to their hex to prevent them from overlapping with adjacent units in such a noticeable manner.
> 
> 
> 
> I'll mark this as postponed, since we might actually implement the proper fixes one day.

wesnoth-bugs:
**Modified on 2009-10-29**

[...]

PR #6050, opened 2021-08-28
lower isolated pine trees within image

doofus-01:
addresses #1153 (and #3600) in a simpler way than introducing new layers


doofus-01:
I didn't squash this based on previous feedback, but the Goldilocks Zone is not obvious to me.

PR #6053, opened 2021-08-29
[world conquest] [1.16] [lua] Check for wc2_scenario value before executing wc2_start_units function 

### LINKED ISSUE #5900: [world conquest] [BUG] faction recruit list gets appended by recruits of another faction of the same era and Leader unit inherits heroic trait for each scenario and it stacks
 - Battle for Wesnoth v1.15.14+dev (8679f0573f8-Modified) x86_64
 - Github
 - Msys2
 - What OS are you running?
 - Windows 10

### Describe the bug
After each successive scenario, the recruit list is appended with the recruits of another faction of the same era. Additionally, the leader unit inherits "trait heroic" for every scenario, and the effects stack.

### Expected behavior
Recruit list stays the same. 

### Screenshots
Screenshot 1: My recruit list is of the Drakes faction from Default Era.
SS1
Screenshot 2: My recruit list is now Loyalists and Drakes.
Screenshot 2021-06-27 202228
Screenshot 3: My recruit list has now Drakes, Northerners and Loyalists.
SS3
Screenshot 4: My recruit list has all factions except Knalgan Alliance.
SS4
Screenshot 5: Leader unit inherits "heroic" trait for each scenario. It can be clearly seen how many times it has stacked.
image

knyghtmare:
What this does is just checks the value of wc2_scenario before executing the function, and executes it for the first scenario only as intended.
Same as #6052, Fixes #5900 for 1.16

### DIFF
--- a/data/campaigns/World_Conquest/lua/campaign/scenario.lua
+++ b/data/campaigns/World_Conquest/lua/campaign/scenario.lua
@@ -66,9 +66,11 @@ end)
 -- NOTE: this is a bit fragile, in particualr it breaks if difficulty_selection happens before the prestart event above.
 on_event("wc2_start", function(cx)
 	for side_num = 1, wml.variables.wc2_player_count do
-		wesnoth.wml_actions.wc2_start_units {
-			side = side_num
-		}
+		if wml.variables.wc2_scenario == 1 then
+			wesnoth.wml_actions.wc2_start_units {
+				side = side_num
+			}
+		end
 	end
 
 	if wml.variables.wc2_difficulty.extra_training then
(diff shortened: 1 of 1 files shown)


data/campaigns/World_Conquest/lua/campaign/scenario.lua
```diff
@@ -66,9 +66,11 @@ end)
...
 on_event("wc2_start", function(cx)
 	for side_num = 1, wml.variables.wc2_player_count do
-		wesnoth.wml_actions.wc2_start_units {
-			side = side_num
-		}
+		if wml.variables.wc2_scenario == 1 then
+			wesnoth.wml_actions.wc2_start_units {
+				side = side_num
+			}
+		end
```
stevecotton, line 73: Either way works and there's no practical problem here, but I think doing `if ... then; for ... do` would be better style than `for ... do; if ... then`, because the condition being checked is the same each time through the loop.

knyghtmare:
Why did one check fail? Kind of confused

Pentarctagon:
Every once in a while the CI just decides it doesn't feel like running one of the macOS jobs and instead just marks it as failed.  Like in this case if you click on the Details link, you can see that when it got to the third step of actually building Wesnoth, it just failed without actually doing anything.

Pentarctagon:
In any case, is this good to merge?

knyghtmare:
Yes, it is.

Issue #5108, opened 2020-08-29
World Conquest: treasure unobtainable if bonus point is on the southern boundary


MarcusAichmayr:
### Game and System Information
 - 1.15.4
 - Steam
 - Ubuntu 20.04

### Describe the bug
If a bonus point is on the southern boundary of the map and you obtain a treasure there, then the treasure will be unreachable.
This is because the treasure is placed on the area directly south to the bonus point and this area is not reachable by any unit.

### To Reproduce
To reproduce this issue, the generated map must have a bonus point on the southern boundary of the map.
By visiting it, an event will be triggered. If this is a treasure event, then the treasure will be placed on an unreachable area as described above.
Note that randomly another event might be triggered instead. If this is the case, you can reload the corresponding save file to visit the bonus point again.

### Screenshots
![image](https://user-images.githubusercontent.com/70106364/91634168-933da380-e9ee-11ea-9295-c68401bf6879.png)

### Savefiles
The Footpad cannot reach the area of the treasure south of him.
[WC II 1 - Start Turn 2_treasure unobtainable.gz](https://github.com/wesnoth/wesnoth/files/5145051/WC.II.1.-.Start.Turn.2_treasure.unobtainable.gz)

This save file can be used to trigger the event by visiting the bonus point _Tarigolin's Route_. Note that another event might be triggered instead of the treasure event. In this case, you can reload the save file and try again to trigger the treasure event.
[WC II 1 - Start-Auto-Save2.gz](https://github.com/wesnoth/wesnoth/files/5145050/WC.II.1.-.Start-Auto-Save2.gz)

gfgtdf:
Hmm okay then we have two obvious ways to fix it 
1) don't spawn bonus points near the southern border, 
2) find a free location when a bonus point is discovered.

IceTyp:
I am not sure if this is already implemented, but we should also ensure that the location where the treasure is placed is not impassable, unwalkable, or Deep Water or something like that.

Wedge009:
The use of hexes can make checking boundary conditions tricky. I recall a similar issue with villages, #1390.

Pentarctagon:
Was this ever fixed?

knyghtmare:
Still persists for Battle for Wesnoth v1.15.14+dev (8679f0573f8-Modified) x86_64
![image](https://user-images.githubusercontent.com/5283677/123485155-5e5ca600-d62b-11eb-88db-385de002d427.png)

[...]

PR #6055, opened 2021-08-29
[1.16] Fix an off-by-one in mapgen's x and y filters, and add unit tests

stevecotton:
Fixes the currently suspected root cause of #5108, but requires further testing
to confirm that it fixes that issue.

Closes #6501 (the question of how to test the mapgen filter).

Opening the 1.16 PR first, will forward-port afterwards.
Ping @knyghtmare

### DIFF
--- a/data/test/scenarios/lua_map_find.cfg
+++ b/data/test/scenarios/lua_map_find.cfg
@@ -1,0 +1,35 @@
+#textdomain wesnoth-test
+
+# Tests for the game kernel's version of w.map.find.
+{GENERIC_UNIT_TEST "lua_map_find" (
+    [event]
+        name = start
+        [lua]
+            code = <<
+                local gg_locs = wesnoth.map.find({terrain = "Gg"})
+                unit_test.assert_equal(#gg_locs, 13, "unexpected number of 'Gg' terrains found")
+
+                local gg_kov_locs = wesnoth.map.find({terrain = "Gg^Kov"})
+                unit_test.assert_equal(#gg_kov_locs, 149, "unexpected number of 'Gg^Kov' terrains found")
+
+                -- "keep_locs" and "castle_locs", ignoring that the whole map has the keep overlay
+                local castle_locs = wesnoth.map.find({terrain = "C*^*"})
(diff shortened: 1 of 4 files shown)


knyghtmare:
> Ping @knyghtmare
I have begun to run tests. I will load the WC scenario 20-30 times and see if the bug is present or not and upload the excel file here.

knyghtmare:
My testing results. These were the occurrences I got, and the bug seems to have been fixed.
image
image

Excel File of 20 trials.
[Bug-5108-test.xlsx](https://github.com/wesnoth/wesnoth/files/7073007/Bug-5108-test.xlsx)

CelticMinstrel:
The fix _looks_ good at a glance, but I am uncertain about one thing. Does the filter used in WC separately exclude border hexes? Because if it not, it looks like your change now means the items can be placed on the top or left border of the map. If that's not correct, the check should be changed from `>= 0` to `> 0`.

stevecotton:
The bounds checks here are just to ensure the C++ doesn't access an invalid index into the bitset. Filtering the top and left borders hexes is done by the bitset itself.

The code in WC does exclude the borders, it filters for `"2-" .. map.width-3`.

CelticMinstrel:
That sure is ugly though… probably we should at least change that to `map.playable_width-2` for symmetry… and I guess it's intentional, but it prevents placing them not only _on_ the border but also adjacent to the border.

stevecotton:
Changing it to use map.playable_width is scope creep for this PR, because this PR is fixing an API bug; it doesn't need to change any of the campaign files.

Changing map.width-3 to map. playable_width-2 sounds like an off-by-one error too. Assuming the border is the same size on both sides, the difference should be even.

CelticMinstrel:
> Changing map.width-3 to map. playable_width-2 sounds like an off-by-one error too.

Hmm… you're probably right, actually; `playable_width` should be 2 less than `width`. I guess we just have to live with the asymmetry.

stevecotton:
I think it's ready to merge, but waiting for a review approval.

CelticMinstrel (approved):

PR #6058, opened 2021-09-01
Fix an off-by-one in mapgen's x and y filters, and add unit tests

### LINKED ISSUE #5108: World Conquest: treasure unobtainable if bonus point is on the southern boundary

stevecotton:
Forward-port of 6055, fixes #5108 and closes #6501 (the question of how to test the mapgen filter).

(cherry picked from commit 30f50bac681fda1566f5cee5aad7be6b65cb0fdb)

### DIFF
--- a/data/test/scenarios/mapgen_filter.cfg
+++ b/data/test/scenarios/mapgen_filter.cfg
@@ -1,0 +1,103 @@
+#textdomain wesnoth-test
+
+# Check that the mapgen kernel's wesnoth.map.filter correctly handles "terrain" values.
+# This will load the generic_unit_test map, change all of the hexes that are between Alice's and
+# Bob's castles to void, and then check that the map looks as expected.
+{GENERIC_UNIT_TEST "mapgen_filter_terrain" (
+    map_generation=lua
+    map_data=""
+    map_file=""
+    [generator]
+        id = "string only shown in error messages (id)"
+        config_name = _ "string only shown in error messages (config_name)"
+        create_map = <<
+            local template_map_data = filesystem.read_file("test/maps/generic_unit_test.map")
+            local map = wesnoth.map.create(template_map_data)
+
+            local filter_hexes_to_change = wesnoth.map.filter(
+                {"any",
+                    {"terrain", "C*^*"},
+                    {"terrain", "K*^*"}
+                }
+            )
+            for i,loc in ipairs(map:find(filter_hexes_to_change)) do
+                map[loc] = "Xv"
+            end
+
+            return map.data
+        >>
+    [/generator]
+
+    # The mechanism used to test the results of the generator is itself tested in lua_map_find.
+    [event]
+        name = start
+        [lua]
+            code = <<
+                local gg_locs = wesnoth.map.find({terrain = "Gg"})
+                unit_test.assert_equal(#gg_locs, 13, "unexpected number of 'Gg' terrains found")
+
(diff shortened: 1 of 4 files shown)


data/test/scenarios/mapgen_filter.cfg
```diff
@@ -0,0 +1,103 @@
+#textdomain wesnoth-test
+
+# Check that the mapgen kernel's wesnoth.map.filter correctly handles "terrain" values.
+# This will load the generic_unit_test map, change all of the hexes that are between Alice's and
+# Bob's castles to void, and then check that the map looks as expected.
+{GENERIC_UNIT_TEST "mapgen_filter_terrain" (
+    map_generation=lua
+    map_data=""
+    map_file=""
```
CelticMinstrel, line 9: I think these are supposed to be omitted when making a generated map…

stevecotton: Oops, I documented them on the Wiki and then forgot about it. Having them be empty strings is accepted by the C++, but as there's no way to unset an attribute in WML the only way to remove them would be to replace the use of GENERIC_UNIT_TEST. Is that worth doing?

CelticMinstrel: Ahhh, so they're here to null out the default from `GENERIC_UNIT_TEST`? As long as it doesn't cause any problems with the map generator, I think that should be fine. The only thing is, I brought it up because I thought I noticed someone recently saying it _does_ cause problems… although maybe that doesn't apply in unit tests?

stevecotton: > I thought I noticed someone recently saying it *does* cause problems

Could you be thinking of my recent Wiki edit that says [setting it to a *non-empty* value causes implementation-defined behavior](https://wiki.wesnoth.org/index.php?title=MapGeneratorWML&curid=1446&diff=68460&oldid=68335)? Having a non-empty `map_data` skips map generation, and at this point any non-empty `map_file` will already have been read into `map_data`:

https://github.com/wesnoth/wesnoth/blob/ef0da66ec9be823f73490fd11434fae0edafdb00/src/saved_game.cpp#L498-L506

CelticMinstrel: That's probably what I was thinking of, which means there's no problem here at all.

PR #5382, opened 2020-12-21
Tweak Delfador's Memoirs difficulty

cooljeanius:
This campaign mostly had a good difficulty level, there were just a few speedbumps. Specifically, Swamps of Illuven. I kind of went overboard with my changes to Swamps of Illuven at first (making Garrath actually honor his deal on EASY was kind of gamebreaking!), so I then had to go back and add a bunch of other events to reduce the benefit of having Garrath as an ally. Outside of that scenario, though, most of my changes were just basic tweaks to gold and turns amounts. I was kind of bothered by the starting gold being left implicit in many places, so I went and made sure to explicitly specify it in as many places as I felt was necessary. Marking this as a draft because the S06 changes are rather drastic and I am totally open to doing them in a different way. Also it's possible that some of my diffs didn't exactly apply entirely cleanly, so it might be worth trying to apply the patches from https://github.com/cooljeanius/wesnoth_mods/tree/master/campaigns/Delfadors_Memoirs with some different arguments to `patch(1)` to see if I missed anything. 
cc @nemaara

PR #6060, opened 2021-09-05
Avoid contemporary phrasing

Wedge009:
Introduced in #5382, discussion seems to agree that we should avoid this.

### DIFF
--- a/data/campaigns/Delfadors_Memoirs/scenarios/06_Swamps_of_Illuven.cfg
+++ b/data/campaigns/Delfadors_Memoirs/scenarios/06_Swamps_of_Illuven.cfg
@@ -185,7 +185,7 @@
                     [/message]
                     [message]
                         speaker=Lionel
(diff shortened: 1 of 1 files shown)


data/campaigns/Delfadors_Memoirs/scenarios/06_Swamps_of_Illuven.cfg
```diff
@@ -185,7 +185,7 @@
                     [/message]
                     [message]
                         speaker=Lionel
-                        message=_"While it appears we lucked out this time, I fear for the travelers for whom this man does NOT keep his word..."
+                        message=_"While it appears we were lucky this time, I fear for the travelers for whom this man does NOT keep his word..."
```
nemaara, line 188:
```suggestion
                        message=_"While it appears we were fortunate this time, I fear for the travelers for whom this man does NOT keep his word..."
```

nemaara (approved):

PR #6069, opened 2021-09-08
Ignore auto-generated Steam file in save directory

Wedge009:
I haven't seen anyone else remark about this, which surprises me. But the `steam_autocloud.vdf` file that constantly gets updated and shown in my game load dialogue bothers me enough to make this small change.

I don't know if it's worthwhile ignoring any file that doesn't follow `*.gz` or other expected file format, but that's beyond the scope of this PR. In such a case, we may as well follow what seems to be customary in regular application load dialogues and have a filter under or next to name field, with `*.gz` as the default filter.

I also don't know if this change is small enough to be allowed into 1.16 branch, hence the back-port reminder label.

PS Any attempt to load `steam_autocloud.vdf` results in an error message 'The file you have tried to load is corrupt', so I can't imagine a normal real-world case where anyone would want to leave this file visible in the load dialogue.

### DIFF
--- a/src/save_index.cpp
+++ b/src/save_index.cpp
@@ -241,9 +241,8 @@ public:
 	}
 
 private:
-	/* 'The file steam_autocloud.vdf will be created in each location specified by your Steamworks cloud paths.
-		This file is used by Steam, and can be ignored by your game.'
-		Reference: https://partner.steamgames.com/doc/features/cloud */
+	/* Steam documentation indicates games can ignore its auto-generated 'steam_autocloud.vdf'.
+	   Reference: https://partner.steamgames.com/doc/features/cloud (under Steam Auto-Cloud section as of September 2021) */
 	const std::vector<std::string> files_to_ignore_ = std::vector {std::string {"steam_autocloud.vdf"}};
 };
 
--- a/src/save_index.cpp
+++ b/src/save_index.cpp
@@ -213,6 +213,7 @@ std::shared_ptr<save_index_class> save_index_class::default_saves_dir()
 	return instance;
 }
 
+/** Filter file names based on input string. */
 class filename_filter
 {
 public:
@@ -230,6 +231,22 @@ private:
 	std::string filter_;
 };
 
+/** Ignore certain files - in particular, the auto-generated Steam cloud record. */
+class filename_ignore
+{
+public:
+	bool operator()(const std::string& filename) const
+	{
+		return std::find(files_to_ignore_.begin(), files_to_ignore_.end(), filename) != files_to_ignore_.end();
+	}
+
+private:
+	/* 'The file steam_autocloud.vdf will be created in each location specified by your Steamworks cloud paths.
+		This file is used by Steam, and can be ignored by your game.'
+		Reference: https://partner.steamgames.com/doc/features/cloud */
+	const std::vector<std::string> files_to_ignore_ = std::vector {std::string {"steam_autocloud.vdf"}};
+};
+
 /** Get a list of available saves. */
 std::vector<save_info> save_index_class::get_saves_list(const std::string* filter)
 {
@@ -238,6 +255,7 @@ std::vector<save_info> save_index_class::get_saves_list(const std::string* filte
 	std::vector<std::string> filenames;
 	filesystem::get_files_in_dir(dir(), &filenames);
 
+	filenames.erase(std::remove_if(filenames.begin(), filenames.end(), filename_ignore()), filenames.end());
 	if(filter) {
 		filenames.erase(
 			std::remove_if(filenames.begin(), filenames.end(), filename_filter(*filter)), filenames.end());


src/gui/dialogs/game_load.cpp
```diff
@@ -54,6 +54,11 @@ static lg::log_domain log_gameloaddlg{"gui/dialogs/game_load_dialog"};
 #define LOG_GAMELOADDLG   LOG_STREAM(info,  log_gameloaddlg)
 #define DBG_GAMELOADDLG   LOG_STREAM(debug, log_gameloaddlg)
 
+// 'The file steam_autocloud.vdf will be created in each location specified by your Steamworks cloud paths.
+//  This file file is used by Steam, and can be ignored by your game.'
```
soliton-, line 58:
```suggestion
//  This file is used by Steam, and can be ignored by your game.'
```

Wedge009: (facepalm)

Thanks for that.

Pentarctagon:
> I don't know if it's worthwhile ignoring any file that doesn't follow *.gz or other expected file format, but that's beyond the scope of this PR. In such a case, we may as well follow what seems to be customary in regular application load dialogues and have a filter under or next to name field, with *.gz as the default filter.

Uncompressed save files are an option that can be enabled in the advanced preferences, as is using bzip2 instead of gzip.  I don't know why we need those options available to be honest, but they do currently exist.

Wedge009:
Yes, I remembered there were other options, hence the thoughts on having a filter in the load dialogue just as many other applications do. Could select between default gzip (or whatever the current preference happened to be), bzip2, or 'All Files'. But then, I suppose uncompressed save files have no extension at all, do they?

Anyway, as I mentioned, that's outside the scope of my intention. All I want to do here is get rid of the `steam_autocloud.vdf` that's constantly getting in the way.

stevecotton:
I like the intention, and sorry for not reviewing earlier. I think the PR as-is is already better than the status quo, and support putting in it 1.16.

This should probably also affect the `"There are no saved games to load."` check in `game_load::execute`, however that would make it a much bigger change.

It would be nicer instead of an `#define` to use something like
```
namespace {
    /** 'The file steam_autocloud.vdf will be created in each location specified by your Steamworks cloud paths.
     *  This file is used by Steam, and can be ignored by your game.'
     * Reference: https://partner.steamgames.com/doc/features/cloud
     */
    const auto files_to_ignore = std::string {"steam_autocloud.vdf"};
}
```

Wedge009:
I don't recall seeing any 'no games to load' messages lately, but I'll double-check that and see what I can do to accommodate it.

I can make the update you suggested, but just wondering what the benefit of doing it that way is supposed to be. For simple things I was told that preprocessor directives are quicker and simpler, but maybe that's just an 'old fashioned' way of doing things - I'll accept I'm not familiar with newer conventions.

Wedge009:
After re-reading #5517, I think I don't get the 'no games to load' message because even though my 1.15 and 1.17 directories have no saved games, my 1.14 directory does (and plenty of them). So all I see - at least initially - is a blank dialogue.

Wedge009:
To accommodate the 'no games to load' feature, I realised I should be ignoring files at _retrieval_ of the save game list (`save_index.cpp`) rather than just at the display of those saved games (`game_load.cpp`). I've updated my PR accordingly, but please advise if there's anything I could do better. I've tested both with empty save directories (other than the `steam_autocloud.vdf`) as well as non-empty.

PS I hope it's acceptable for 1.16 branch - I could argue that the presence of the Steam file is a 'bug'. (:

src/save_index.cpp
```diff
@@ -230,6 +231,22 @@ class filename_filter
...
+{
+public:
+	bool operator()(const std::string& filename) const
+	{
+		return std::find(files_to_ignore_.begin(), files_to_ignore_.end(), filename) != files_to_ignore_.end();
+	}
+
+private:
+	/* 'The file steam_autocloud.vdf will be created in each location specified by your Steamworks cloud paths.
+		This file is used by Steam, and can be ignored by your game.'
```
stevecotton, line 245: Maybe paraphrase this two lines instead of copying them verbatim? I don't think it matters, but means no-one will wonder about the copyright.

stevecotton:
Looks good to me, and thanks for seeing that I meant to make the ignore list a `std::vector<std::string>` but forgot that in my example.

The thing about avoiding the preprocessor was generally a style thing of trying to avoid the preprocessor if possible.

src/save_index.cpp
Wedge009: Why would copyright be a concern? I was simply trying to show that the Steam documentation makes it clear that we (the game) can ignore this auto-generated file and I assume that most people don't bother clicking through links (plus on-line documentation can change too). I can change this, but I don't see a compelling reason to do so given what I was trying to do here.

Wedge009:
That's good - the stand-alone string would have been easier but I eventually worked out how to deal with the general vector case.

Funny, maybe it comes from my prior C++ background (pre-C++11 days) where using the preprocessor as much as possible to reduce run-time burden was seen as a good thing...

src/save_index.cpp
stevecotton: Copyright is a concern because we distribute Wesnoth; I don't think such a small quote matters but copyright's something where it's easier to avoid the question than to work out what the answer to the question is.

How about "Metadata that's automatically created by the Steam cloud-save framework, not a game that can be loaded"?

Wedge009: My question was more along the lines of how copyright is a concern since I was only quoting public documentation - doesn't the US have something called 'fair use' as well? Plus this is only commentary, not actual code.

But if it seems this is such a cause for concern I'll rephrase this shortly.

stevecotton:
Remember to squash when merging :D

stevecotton (approved):

Wedge009:
Thank you. @Pentarctagon do you feel comfortable to let this go to 1.16 or would you prefer to wait till after release?

Pentarctagon:
I'm fine with it going in 1.16.

Wedge009:
Thanks, I'll merge shortly then.

PR #6073, opened 2021-09-11
Rename kick attack for horses and wyverns

### LINKED ISSUE #6071: Issue in translating "kick" unit attack. New string required.
### Situation

As of my knowledge, the string for the "kick" attack is present in the following units: **Horses** (bay, dark, white unit ranges) the **Wild Wyvern**, and the **Orcish Nightblade**. See screenshot below.

image

### Issue

When translating the string "kick" to Portuguese (Portugal) a problem arises from the fact the language has separate words for "human kicks" and "animal kicks". A kick delivered by a human foot will always be called _pontapé_. A kick delivered by an animal is usually called _coice_. This means that picking either of the options for translation will always end up in at least one inadequate translation because there is only one string to cover all three attacks. There are also no "neutral" Portuguese words I can use in this circumstance (such as the Portuguese equivalent for the words "hit" or "slam"), because the attack icon for horses show a picture of a hoof, and the attack animation for the Nightblade displays a flying kick. Therefore a new string is necessary for the horse attacks in order to ensure a decent translation.

### Proposed solution

- Create a new dedicated string for the horse attack called "hooves", since it's in line with other animal attacks like "claw", "tusk", etc.
- Replace the Wyvern attack name with "slam", which is already used in Sky Drakes and currently uses the same icon as the Wyvern "kick". See picture below:

image

Wedge009:
Makes attack name consistent with other weapon attacks and also avoids translation issues (resolves #6071).

### DIFF
--- a/data/core/units/monsters/Horse.cfg
+++ b/data/core/units/monsters/Horse.cfg
@@ -28,8 +28,8 @@
     [/resistance]
     {DEFENSE_ANIM "units/monsters/horse/horse-defend-2.png" "units/monsters/horse/horse-defend-1.png" {SOUND_LIST:HORSE_HIT} }
     [attack]
-        name=kick
-        description=_"kick"
+        name=hooves
+        description=_"hooves"
         icon=attacks/hoof.png
         type=impact
         range=melee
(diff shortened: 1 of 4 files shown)


sheepo99:
I think the plural "hooves" would perhaps sound a bit better. When horses kick they do so with both legs.

CelticMinstrel:
I was thinking the same thing.

Wedge009:
I was thinking that attacks were mainly singular, but then I noticed theire's throwing knives. I also thought claw-based attacks were just 'claw', but they are indeed 'claws'. I'll update this shortly.

Issue #6046, opened 2021-08-24
UtBS S3, Garaks blessing includes steadfast, which is useless for him


Konrad22:
### Game and System Information
 - What version of the game are you running? 1.15.15
 - Where did you download it from? (Steam, Mac App Store, our website, built from source, etc.) Steam
 - What OS are you running? Windows 10

### Describe the 'bug'
Garak has -10% arcane and 0% for other resistances, which is why 'steadfast' is useless and serves only to confuse the player.

### To Reproduce

1. First load the save file, check his stats and abilities.

### Expected behavior
No steadfast, and/or maybe some ability that raises his resistances.

### Savefiles
[UtBS-A Stirring in the Night-Auto-Save4.gz](https://github.com/wesnoth/wesnoth/files/7038818/UtBS-A.Stirring.in.the.Night-Auto-Save4.gz)

knyghtmare:
> No steadfast, and/or maybe some ability that raises his resistances.

Maybe some ability which gives +20% resistances on offence?
Or, some ability which gives +10% resistances on defence?

Actually, no. This would mean more strings...with string freeze in effect...so...no steadfast seems the best option.

Discontinuum:
Add him some armour with the blessing xD and the steadfast will become useful

Konrad22:
Reopening this, as the resistance ability special for Garak is just a useless as the steadfast was. The ability triggers only if Garak has more than 0% resistance, which he doesn't have vs any dmg type. Same problem as steadfast.

soliton-:
The filter_base_value part seems like it should just be removed, `max_value` already caps it to 50% and I see no reason to special case vulnerabilites. Of course the description would need to be adjusted.

Btw, if you remove the C parts of your URL (`#L699C1` -> `#L699`) and just keep it to full lines then github will show the linked range itself.

PR #6075, opened 2021-09-11
Replace steadfast with resistance abilities special for Garak

newfrenchy83:
if Garak have 0% or -10% resistance only then steadfeast don't usable here. Fix https://github.com/wesnoth/wesnoth/issues/6046


nemaara (approved):

nemaara:
I'm assuming this has been tested and works. Looks fine to me.

PR #6078, opened 2021-09-12
1.16 - terrain graphics variations and transitions

doofus-01:
tents_001
- camp/tent village variations
- Great Dead Tree variations


doofus-01:
I intend to get a decent `Irs`-`W*` transition, like how the wooden floor has a believable transition.  Not sure if I'll get it done in time.
Screenshot_001

doofus-01:
Screenshot_Irs-W_20210919
I'm not sure this is going to make it into 1.16.0.  I think I addressed most of the transitions (even if more work is needed), but there are still a few corner cases that aren't obscure enough to ignore.

PR #6087, opened 2021-09-15
fix possible bug in attack prediction windows

newfrenchy83:
if an weapon like abilitie with apply_to=opponent is used, then name appear in attack

### DIFF
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -864,7 +864,7 @@ std::string attack_type::weapon_specials(bool only_active, bool is_backstab) con
 	if(self_){
 		std::set<std::string> checking_name;
 		for (const config::any_child sp : (*self_).abilities().all_children_range()){
-			const bool active = check_self_abilities_impl(shared_from_this(), other_attack_, sp.cfg, self_, self_loc_, AFFECT_EITHER, sp.key);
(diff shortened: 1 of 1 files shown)


Pentarctagon:
Can you provide a screenshot of the issue?

newfrenchy83:
> 
> 
> Can you provide a screenshot of the issue?

bug_attack_window

newfrenchy83:
it si when absorb implemented in [abilities] instead of [specials]( in specials, it is normal)

newfrenchy83:
bug_fixed
ans when it si fixed

stevecotton:
Should it not appear on the opponent's attack?

newfrenchy83:
> 
> 
> Should it not appear on the opponent's attack?

I have already tried to implement it but this randomly generates another bug, the display of the percentage of cth may not appear in the opponent's attack, and I cannot find a solution because the character random affects the relevance of the tests

Pentarctagon:
If I'm understanding, the issue is that when a unit has an ability that's a weapon special (ie: firststrike as an ability), and when that ability affects the opponent, the ability's name is shown on the attacking unit's weapons in the attack dialog?

newfrenchy83:
> 
> 
> If I'm understanding, the issue is that when a unit has an ability that's a weapon special (ie: firststrike as an ability), and when that ability affects the opponent, the ability's name is shown on the attacking unit's weapons in the attack dialog?

Yes, exactly.

PR #6089, opened 2021-09-16
Switch to hard-coded colours instead of relying on pango names

### LINKED ISSUE #4348: Make tooltips use brighter green
Some tooltips, at least the time of day and traits tooltips, use a dark green color. @ProditorMagnus reports this is platform specific.

See https://forums.wesnoth.org/viewtopic.php?p=647304#p642778 and following posts (including the spoiler)

The dark green should be made brighter, probably #00ff00 which is what windows already uses. Also make sure the red is legible.

Wedge009:
Resolves #4348

Pango changed its definition of 'green' in version 1.32.2 to align with CSS specifications (see #4348 for details). This change focuses only on instances of `<span color="green">`, leaving instances of named colours that don't use 'green'. For example:
https://github.com/wesnoth/wesnoth/blob/f8a41624e3022947b6d7d9ac6b0311c4a7d1fee4/src/gui/dialogs/gamestate_inspector.cpp#L598

**Problem**: I cannot see how to use this approach with `<format>` so it breaks time-of-day bonus colour consistency. Then again, I'm not sure if the green used in `<format>` is the same as the green used in `<span>`...
https://github.com/wesnoth/wesnoth/blob/f8a41624e3022947b6d7d9ac6b0311c4a7d1fee4/src/help/help_impl.cpp#L398-L402
tod-schedule

Also:
https://github.com/wesnoth/wesnoth/blob/f8a41624e3022947b6d7d9ac6b0311c4a7d1fee4/src/help/help_topic_generators.cpp#L63-L68
Plus coloured text for 'vision' and 'jamming' that I can't find in the GUI.

**Questions**:
* How might `<format>` colours be updated? (Edit: Is it a case of Help being stuck on GUI1?)
* Where is the formula debugger used? I didn't change this one because I can't find where to test it: https://github.com/wesnoth/wesnoth/blob/f8a41624e3022947b6d7d9ac6b0311c4a7d1fee4/src/gui/dialogs/formula_debugger.cpp#L49-L51

**Comparisons**:
intelligent-beforeintelligent-after
quick-beforequick-after
resilient-beforeresilient-after
strong-beforestrong-after
tod-beforetod-after

### DIFF
--- a/src/gui/dialogs/formula_debugger.cpp
+++ b/src/gui/dialogs/formula_debugger.cpp
@@ -46,8 +46,8 @@ void formula_debugger::pre_show(window& window)
 		for(int d = 0; d < c; ++d) {
 			stack_text << indent;
 		}
-		stack_text << "#<span color=\"green\">" << i.counter()
-				   << "</span>: \"<span color=\"green\">" << font::escape_text(i.name())
+		stack_text << "#<span color=\"#00ff00\">" << i.counter()
+				   << "</span>: \"<span color=\"#00ff00\">" << font::escape_text(i.name())
 				   << "</span>\": (" << font::escape_text(i.str()) << ") " << std::endl;
 		++c;
 	}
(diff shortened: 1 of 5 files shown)


cooljeanius:
speaking of which, could the list of colors be documented somewhere on the wiki? Last time I looked I couldn't find any such list...

Wedge009:
Suggest you don't side-track the PR, but what are you looking for exactly? Any time something is documented separately from code, there's the risk of divergence. That's a general principle, not specific to Wesnoth. Even commentary within code can diverge from what the code itself is saying if it's not maintained properly.

CelticMinstrel:
> Where is the formula debugger used? I didn't change this one because I can't find where to test it:

Run the game and play any map in debug mode. (I use --debug on the command-line but enabling debug mode with :debug should also be sufficient.) Then press F and you get a console where you can type in a WFL formula to evaluate. If that formula contains a call to the `debug()` function, then the formula debugger will appear. (If you don't enable debug mode, it won't appear.)

Wedge009:
Thank you. Updated formula debugger as well now:
debugger-beforedebugger-after

Vultraz (approved):

PR #6090, opened 2021-09-16
UtBS: Silence the Dust Devil

### LINKED ISSUE #4892: UtbS Battle for Zocthanol Isle - Dust Devil speaks?!
Wesnoth 1.15.3
Under the Burning Suns
Battle for Zocthanol Isle

The unit which defeats the Undead leader finds a golden key around its neck and says as much. Except... Dust Devils don't speak. There should be a check on which unit has dispatched the Draug, and if it's the Dust Devil - have an adjacent unit speak instead; or perhaps Kaleh.

Wedge009:
Use existing macro to avoid situations where Nym's otherwise-mute Dust Devil pet might speak.

Resolves #4892.

Note: I'm not familiar with WML so I'm not sure if I've done this 'right'. But for a quick test, focusing on the Draug mentioned in the aforementioned issue, I did the following (GUI Debug Tools wasn't working for me - maybe not yet working for 1.17.0+dev):
* Start UtBS campaign and debug-skip to Battle for Zocthanol Isle.
* Debug-remove fog and shroud.
* Debug-create Mage of Light adjacent to Kelur and attack to bring its HP down (high probability of all three attacks hitting).
* Debug-create Dust Devil and use it to finish off Kelur.

Before: Dust Devil remarks about the key left behind.
After: Nym speaks on the Dust Devil's behalf.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/06a_In_the_Tunnels_of_Trolls.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/06a_In_the_Tunnels_of_Trolls.cfg
@@ -1458,10 +1458,12 @@
             message= _ "Argh! Curse you! May you never live to see daylight again!"
         [/message]
 
+        {CHECK_SPEAKER}
         [message]
-            speaker=second_unit
+            speaker=$speaking_unit.id
             message= _ "And so, at last, it ends."
         [/message]
+        {CLEAR_VARIABLE speaking_unit}
 
         [message]
             side=4
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
@@ -759,10 +759,12 @@
             message= _ "Ha, you’re trapped. I’ve got you right where I want you, and this time no one is gonna save you."
         [/message]
 
+        {CHECK_SPEAKER}
         [message]
-            speaker=second_unit
+            speaker=$speaking_unit.id
             message= _ "How did a troll get stuck all the way back here?"
         [/message]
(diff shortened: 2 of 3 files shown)


nemaara:
Hmm are you sure this works? I don't see the {CHECK_SPEAKER} macro being added anywhere and I don't remember it being in the original game?

Wedge009:
It worked for the one test I did as explained in the original post, plus I wouldn't have known to do this if I didn't have examples to follow in the rest of the campaign - initially I tried to do my own checks manually via `[if]`, which worked but was clumsy, until I found the macro. As per the request in #4892 I also assigned the speech to Kaleh, until I noticed in S2 that the Dust Devil was Nym's pet, so it makes sense that she speaks on its behalf.

Source:
https://github.com/wesnoth/wesnoth/blob/41bf3741b816c9e701cd4526bfe53de5a5e2dd89/data/campaigns/Under_the_Burning_Suns/utils/dialog-macros.cfg#L38-L71

Pre-existing example:
https://github.com/wesnoth/wesnoth/blob/41bf3741b816c9e701cd4526bfe53de5a5e2dd89/data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg#L1171-L1176

nemaara:
Ah! Alright.

nemaara (approved):

Wedge009:
Thanks. I'd really like for someone else to test or at least check this before merging because as I said above I'm not familiar with WML. Not sure if @eyalroz is able but it's their request and they're active.

However, if it does do the intended job, I hope @Pentarctagon can approve its suitability for 1.16 branch, but if we can't establish its correctness then perhaps it's best to wait till 1.16.1.

Pentarctagon:
I don't see why it couldn't go into 1.16 as well.

nemaara:
Just checked it, it also works for me. Should be mergeable.

Wedge009:
Great, thank you both.

PR #6092, opened 2021-09-16
Only display difficulty for SP campaigns (resolves #5321)

### LINKED ISSUE #5321: Load game dialogue: shows difficulty for MP scenarios
But there are exist no difficulties in MP scenarios. It shows »Medium« as fallback. Maybe the whole line can be skipped.
load

Wedge009:
There are four save game types encoded:
* SP campaigns (internally called 'scenario')
* Multiplayer
* Tutorial
* Test scenario

I think Tutorial type is the old style tutorial from 1.14 and earlier which was treated as separate from SP campaigns. From 1.15 the Tutorial is included in the SP campaigns list and seems to be treated as such because with this change, MP and Test scenario save games will no longer have difficulty displayed in the save game preview of the game load dialogue, but SP campaigns and the Tutorial will.

### DIFF
--- a/src/gui/dialogs/game_load.cpp
+++ b/src/gui/dialogs/game_load.cpp
@@ -427,8 +427,15 @@ void game_load::evaluate_summary_string(std::stringstream& str, const config& cf
 		str << _("Scenario start");
 	}
 
-	str << "\n" << _("Difficulty: ")
-		<< difficulty_human_str;
+	// Only makes sense to display difficulty for SP campaigns:
+	try {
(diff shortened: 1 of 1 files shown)


stevecotton:
What about LoW's multiplayer mode, which has a difficulty selection using the standard mechanism?

Wedge009:
Good point. MP campaigns do seem to have an actual difficulty included (although WC's `UNTLB` notation doesn't look very user-friendly to me). I'll have to see how I can distinguish campaign vs scenario for MP save game type.

Wedge009:
I don't think this is going to be as straightforward to resolve as I first thought...

PR #6094, opened 2021-09-16
i18n: Fix dsngettext() not stripping caret annotations from plural forms

### LINKED ISSUE #6093: Untranslated string prefixes are not automatically stripped from gettext plural forms
When the game looks up plural form translations and falls back to the C locale string, the caret `prefix^` will only be removed from the singular form if present, and never the plural form.

Example using Lua:

Screenshot_20210916_135907

This affects both 1.16 and master.

**EDIT:** Upon further examination, this appears to be the original intended behaviour. However, this doesn't mean it's a good idea -- see #6094.

irydacea:
This affects operation of `dsngettext()` and its users (e.g. `_n()` and `VNGETTEXT()`) specifically when the plural version of a string is chosen.

Apparently back in 2007 in commit 4852283ab3e1916c66d7d6660d61f9dc2cae38c6, this function and its late sibling were made to only strip carets from singular forms under the probable assumption that disambiguation won't be needed for plurals because of the way GNU gettext pairs them with their singular msgid under a single unique entry. Like everything else related to our gettext functionality, this wasn't documented in the C++ code.

However, there's no guarantee that this will remain the case forever if we some day decide to replace the GNU gettext format with something without this guarantee. It's also likely to confuse more people later down the line (especially now that the Lua API exposes this functionality).

Closes #6093.

### DIFF
--- a/src/gettext.cpp
+++ b/src/gettext.cpp
@@ -411,15 +411,33 @@ std::string dsgettext (const char * domainname, const char *msgid)
 	return msgval;
 }
 
+namespace {
+
+inline const char* is_unlocalized_string2(const std::string& str, const char* singular, const char* plural)
+{
+	if (str == singular) {
+		return singular;
+	}
+
+	if (str == plural) {
+		return plural;
+	}
+
+	return nullptr;
+}
+
+}
+
 std::string dsngettext (const char * domainname, const char *singular, const char *plural, int n)
 {
 	//TODO: only the next line needs to be in the lock.
 	std::scoped_lock lock(get_mutex());
 	std::string msgval = bl::dngettext(domainname, singular, plural, n, get_manager().get_locale());
-	if (msgval == singular) {
-		const char* firsthat = std::strchr (singular, '^');
+	auto original = is_unlocalized_string2(msgval, singular, plural);
+	if (original) {
+		const char* firsthat = std::strchr (original, '^');
 		if (firsthat == nullptr)
-			msgval = singular;
+			msgval = original;
 		else
 			msgval = firsthat + 1;
 	}


src/gettext.cpp
```diff
@@ -411,13 +411,31 @@ std::string dsgettext (const char * domainname, const char *msgid)
...
 	//TODO: only the next line needs to be in the lock.
 	std::scoped_lock lock(get_mutex());
 	std::string msgval = bl::dngettext(domainname, singular, plural, n, get_manager().get_locale());
-	if (msgval == singular) {
-		const char* firsthat = std::strchr (singular, '^');
+	auto original = is_unlocalized_string2(msgval, singular, plural);
+	if (original) {
+		const char* firsthat = std::strchr (original, '^');
 		if (firsthat == nullptr)
 			msgval = singular;
```
stevecotton, line 440: Looks like this needs changing to `msgval = original;`

src/gettext.cpp
```diff
@@ -411,15 +411,33 @@ std::string dsgettext (const char * domainname, const char *msgid)
 	return msgval;
 }
 
+namespace {
+
+inline const char* is_unlocalized_string2(const std::string& str, const char* singular, const char* plural)
```
stevecotton, line 416: Why use `inline` in a .cpp file (as opposed to in an .hpp file)?

irydacea: Personal preference. That's just my way of saying "this tiny function could as well be written into its caller but i wanted it to have a cute name" to people reading the code.

soliton-:
I agree it's a good idea to handle the caret notation for plural strings the same as singular for symmetry and being least surprising even if it's not strictly necessary.

Issue #6036, opened 2021-08-20
utils::format_timespan() is too buggy to contact the translators about


stevecotton:
`utils::format_timespan(std::time_t)` is a function for converting times into strings such as "3 months and 20 seconds".

Several things to note about this:

* It uses a hardcoded assumption of 30 days per month, despite being able to generate strings with years and seconds.
* It's only called from one place, which is to display ban durations to people banned from the multiplayer server.
* It doesn't use [gettext's pluralisation rules](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html), instead it hardcodes the bug described at the top of that page.
* A string such as "3 months and 20 seconds" lacks the concept of significant figures.

The Spanish and French translations omit the number from the translation of `timespan^$num years`, but given the problems in the C++ I think that's the least of the issues here.

irydacea:
Hi, author of the function here. I wrote it as part of fixing #3766 in PR #3904.

Could you clarify what you mean by "lacks the concept of significant figures"?

> It uses a hardcoded assumption of 30 days per month

I thought about this part very hard and concluded that for the initial purpose of the function there was absolutely no need to get into calendar specifics, since most people tend to call 30 days a month anyway to keep things simple. Bear in mind that if you wanted to change this assumption, you would need a precise reference point as a new function argument in order to be able to tell what the individual lengths of the months encompassed by the specified period of time are (in the Gregorian calendar at least).

> It doesn't use gettext's pluralisation rules, instead it hardcodes the bug described at the top of that page.

This is intentional. In order to keep the code clean, a look-up table is used for the format strings, which means that they must not be **translated** strings, while `_n()` only ever returns translated strings. I did consider using gettext plurals but found that rearranging the code for that purpose would make it considerably worse for a fairly nonexistent benefit.

stevecotton:
> Could you clarify what you mean by "lacks the concept of significant figures"?

I mean that it's usual to round numbers a few decimal places, and in a similar way I think that it's odd to include units much smaller than the largest one. "2 months and 10 seconds" is 2.0000039 months; when the largest unit is months then it feels odd to me to include any unit smaller than days. It also feels odd to go to a precision smaller than days when people expect "1 month" to vary by +/- a few days anyway.

Will think more about the code when I'm more awake tomorrow.

irydacea:
[...]

PR #6097, opened 2021-09-17
Make utils::format_timespan() use gettext plurals, simpler output by default

irydacea:
See #6036.

This also adds a new `N_n()` no-op marker to `gettext.hpp` that is to `_n()` as `N_()` is to `_()`.

**NOTE:**

This is **NOT** a string-freeze-compliant change. Changing the strings from being standalone to being part of a plural form set causes them to be automatically fuzzied by the pot-update target. Merging this in 1.16 will have to wait until after 1.16.0 is released.


stevecotton:
I feel keeping the `detailed` option is unnecessary, but as it's already implemented it seems reasonable to keep it.

stevecotton (approved):

Issue #5190, opened 2020-10-04
[remove_object] does not restore status


doofus-01:
### Game and System Information
 - What version of the game are you running?
The Battle for Wesnoth version 1.15.5+dev (97c87443c10-Modified)
Distribution channel: Default

### Describe the bug
[remove_object] does not remove the status effects applied by the object.  Similar to #3222, maybe the same problem. 

### To Reproduce

Add the WML below to the first scenario of HttT.  When Konrad moves, the dialogs will fire and his sprite will change colors as expected, but looking at unit data with `:inspect` shows the status effects are still there after the object is removed.
```
   [event]
        name=moveto
        [filter]
            id=Konrad
        [/filter]
        [message]
            speaker=Konrad
            message= _ "I'm undead!"
        [/message]
        [object]
            id=skull_ring
            [effect]
                apply_to=status
                add=undrainable
            [/effect]
            [effect]
                apply_to=status
                add=unpoisonable
            [/effect]
            [effect]
                apply_to=status
                add=unhealable
            [/effect]
            [effect]
                apply_to=status
                add=unplagueable
            [/effect]
            [effect]
                apply_to=status
                add=not_living
            [/effect]
            [effect]
                apply_to=image_mod
                replace="~CS(-120,-120,-120)"
            [/effect]
        [/object]
        [event]
            name=moveto
            [filter]
                id=Konrad
            [/filter]
            [message]
                speaker=Konrad
                message= _ "I'm alive again!"
            [/message]
            [remove_object]
                object_id=skull_ring
            [/remove_object]            
         [/event]
    [/event]
```

stevecotton:
Although I agree it's a bug, I think we're too close to release to fix it for 1.16.

gfgtdf:
While i agree that this is a bug It's not easy for find a good way to fix this, since tracking whether a status should be reset in cases could turn out to be really hard, since there are usually statuses that shouldn't be reset (slowed, poisoned) because they are applied in combat, and status that should be reset 'unplaguable' 'undead' etc that are usually applied via modifications

CelticMinstrel:
The main problem is that at the point when the object is rebuilt, the information on which status should be reset is gone – that information is on the object that was removed, so it's no longer on the unit.

vgaming:
Given the problems outlined, I guess the only technically correct solution is to deprecate status changes in effects? E.g. effects are removable with the current API, so they should only have things that _can_ be removed.

[...]

PR #6102, opened 2021-09-17
Clear dehydration status on victory (resolves #5687)

### LINKED ISSUE #5687: UtBS: 'dehydrated' icon is displayed in the right panel even in the next scenario
The Battle for Wesnoth version 1.15.11+dev (0575bde-Clean) x86_64
Running on Freedesktop.org 20.08.7 (Flatpak runtime) x86_64

### Describe the bug
'dehydrated' icon is incorrectly displayed in the right panel even if a unit is not dehydrated

### To Reproduce
1. Start UtBS campaign
2. Skip/win the first scenario
3. In the second scenario dehydrate one of the units then skip/win the scenario while the unit is dehydrated
4. In the third scenario check status of the unit in the right panel. The unit will remain marked as dehydrated during the whole campaign. 

### Expected behavior
'dehydrated' icon should not be displayed in the right panel if a unit is not dehydrated

### Screenshots
(Third scenario) In the right panel a unit is incorrectly marked as dehydrated:
image

Wedge009:
Even though units are not dehydrated in subsequent scenarios, the dehydration icon may otherwise still appear in the side-bar without this.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/utils/dehydration-utils.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/utils/dehydration-utils.cfg
@@ -303,4 +303,28 @@ Hd, Dd*, Dd^E*, Rd #enddef
             object_id=dehydration_overlay
         [/remove_object]
     [/event]
+
+    [event]
+        name="victory"
+
+        [modify_unit]
+            [filter]
+                status=dehydrated
+            [/filter]
+
+            [status]
+                dehydrated=no
+            [/status]
+        [/modify_unit]
+
+        [modify_unit]
+            [filter]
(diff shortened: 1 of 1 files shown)


stevecotton:
I think this is a good implementation.

It would be nicer if `[remove_object]` removed the status, but that's a known bug which I think we're too close to release to fix (#5190).

nemaara (approved):

Wedge009:
I followed the gist of what was reported in #5687 and it seemed to work for me - I'm not too familiar with the details of WML generally. Any objection to this being included in 1.16?

Pentarctagon:
Nope.

Wedge009:
Thank you.

PR #6104, opened 2021-09-18
Editor: Work-around crash situation

### LINKED ISSUE #5883: It's possible to have scenario-specific tools active when editing a map, which can cause a crash
### Describe the bug

If you have a regular map and a scenario map open, the tool isn't reset when you close the scenario map. Attempting to use the tool then causes a crash.

### To Reproduce

1. Open the map editor
2. Modify a few hexes, then create a new scenario with the file menu.
3. Select the units tool (or probably, any scenario-only tool)
4. Close the scenario from the file menu (close map)
5. You are now looking at the map from earlier, but the units tool is still active.
6. Click anywhere on the map, and watch Wesnoth crash.

Something else to consider: is it possible for this to happen when simply switching maps via the window menu?

### Expected behavior

If you switch to a map from a scenario for any reason, and the current tool is a scenario-only tool, it should revert you to the regular terrain tool.

Wedge009:
Work-around crash situation caused by using scenario-specific tools in map context (closes #5883)

### DIFF
--- a/src/editor/controller/editor_controller.cpp
+++ b/src/editor/controller/editor_controller.cpp
@@ -877,6 +877,10 @@ bool editor_controller::do_execute_command(const hotkey::hotkey_command& cmd, in
 		// map specific
 		case HOTKEY_EDITOR_MAP_CLOSE:
 			context_manager_->close_current_context();
+			// Copy behaviour from when switching windows to always reset the active tool to the Paint Tool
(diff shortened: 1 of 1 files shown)


Pentarctagon:
Any reason not to merge and backport this before RC1?

Wedge009:
Just waiting for someone to say it's okay to do so. I see 4 thumbs up and no objections so I can do this shortly.

Issue #5743, opened 2021-04-27
Unit checksum code needlessly checks weapon special names, which can cause OOS


ForestDragon-wesnoth:
### Game and System Information
 - Wesnoth 1.14.16, Steam, Windows 10

### Describe the bug
the unit checksum code, when calculating weapon specials, considers weapon special names part of the checksum. therefore, if the name is changed locally, it causes OOS

### Expected behavior
Since weapon special names are a text string that has no actual impact on gameplay, it should not be counted in the checksum in my opinion, as it causes problems for people using my macro-based translation

### Proposed fix:
replace the following code in units.cpp:
https://github.com/wesnoth/wesnoth/blob/fa720141587392938ce65fcaae1827b467bc397c/src/units/unit.cpp#L2678-L2682
with this:
```C++
		for(const config& spec : att.child_range("specials")) {
			config& child_spec = child.add_child("specials", spec);

			child_spec.recursive_clear_value("description");
			child_spec.recursive_clear_value("description_inactive");
			child_spec.recursive_clear_value("name");
			child_spec.recursive_clear_value("name_inactive");
		}
```

soliton-:
Suggestion from ForestDragon-wesnoth to reproduce the issue:

1. install Great Steppe Era, and Great Steppe Era Russian Translation (the way the translation is done with macros can cause changed checksum)
2. play a match with GSE (specifics don't matter, as long as you have a save file with the translated macros) and save a replay
3. uninstall the translation add-on
4. reboot the game
5. load replay

PR #6106, opened 2021-09-18
Fix Issue 5743 surronding unit checksum code caring about weapon spec…

cpranga:
…ial names, which causes OOS when changed locally; fix issue by clearing out respective values by following later code.

### DIFF
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -2734,6 +2734,9 @@ std::string get_checksum(const unit& u)
 			config& child_spec = child.add_child("specials", spec);
 
 			child_spec.recursive_clear_value("description");
+			child_spec.recursive_clear_value("description_inactive");
+			child_spec.recursive_clear_value("name");
+			child_spec.recursive_clear_value("name_inactive");
 		}
 	}
 


cpranga:
Fixes #5743

Pentarctagon:
As a note for the future, to have the issue linked so that merging the PR closes the issue, you need to put `fixes #NNNN` in the text of the commit message.

Wedge009:
I don't normally play MP so I tend not to consider OoS issues. If I understand the original issue correctly, the reporter wants to ignore these fields in the check-sum because they have the potential to vary even though there's no effect game-play. I suppose that makes sense, but I don't really understand the details around that. Has this been tested? (Is it even possible to test easily?)

@soliton- You looked at the original issue #5743...

soliton-:
Looks fine to me.

Would be good to fix the commit message format though. The first line should be a short summary, followed by an empty line, followed by a more detailed description. Something like:

> Improve unit checksum code
> 
> Clear out unimportant attributes from weapon specials. Fixes #5743.

Pentarctagon:
Any chance of getting this merged and backported before RC1 (in about 20 hours)?

cpranga:
Thanks for your patience; I've been sick recently. I was able to force push the updated commit message, but I was unable to test with the listed example; I'm still learning how to do stuff, and other commitments prevented me from doing too much

Pentarctagon:
Alright, no worries :)

soliton- (approved):
Looks good to me. I'm going to merge for master without test since it's simple enough but it'd be nice to get this tested into 1.16 before the next release as well.

PR #6110, opened 2021-09-22
extent addition of the names of the abilities used as weapons in the attack dialog window to apply_to=opponent

newfrenchy83:
If a weapon with apply_to=opponent is used, this name will appear in opponent attack affected.


newfrenchy83:
and screenscaps before and after modifications
without_modifications

and after
with_modification

Pentarctagon:
Isn't this the thing you said you weren't sure how to do in #6087 ?
> I have already tried to implement it but this randomly generates another bug, the display of the percentage of cth may not appear in the opponent's attack, and I cannot find a solution because the character random affects the relevance of the tests

newfrenchy83:
https://github.com/Pentarctagon It's true and that's why it took me a long time, before sending this request, because I was just trying to find a solution, if the names of the abilities used as weapons are placed in the line after the names 'traditional' special weapons, then the display of the chance to hit is not altered, FYI this problem also manifested itself with apply_to = self and therefore is not specific to apply_to = opponent as I initially thought.
Separating special weapons from abilities into two strings separated by a newline seems to remove the problem that appeared in Quenoth fighters when damage was 10 (two characters) instead of 9.

newfrenchy83:
@Pentarctagon could you review this PR,please?

Pentarctagon:
Please change the name to fully spell it out.

stevecotton:
Sorry for being so late to notice this, but I think this is a bad change that should be reverted.

This image looks as if the blue unit has an active buff called "formation", but in terms of combat what's really happening is that the red unit has an advantage because there's a second red unit activating red's formation ability.
image

If there's a second blue unit, the situation is reversed, but that's also confusing. In both cases, the buff's name appears on the disadvantaged side's stats.

newfrenchy83:
> 
> 
> Sorry for being so late to notice this, but I think this is a bad change that should be reverted.
> 
> This image looks as if the blue unit has an active buff called "formation", but in terms of combat what's really happening is that the red unit has an advantage because there's a second red unit activating red's formation ability. 
> 
> If there's a second blue unit, the situation is reversed, but that's also confusing. In both cases, the buff's name appears on the disadvantaged side's stats.

in the case of ability used like special, when apply_to=opponent, this is attack of unit affected who appear with 'formation' name to said, and what player undetstand what blue unit has disavantage because of formation abilities.

Pentarctagon:
From the discussion on Discord, the problem is that this causes a UX issue due to the Formation ability being implemented in a different way than it's described (the text says it increases the unit's defense, but in reality it's decreasing the opponent's chance to hit). So it's not really clear that the blue unit in the above screenshot is being affected by the Formation ability because of the enemy they're attacking vs because they themselves have the Formation ability.

@newfrenchy83 If the Formation ability were implemented like a regular weapon special rather than as an ability, then the red unit in the above screenshot would have "formation" listed under its attack rather than the blue unit, correct?

newfrenchy83:
> 
> 
> From the discussion on Discord, the problem is that this causes a UX issue due to the Formation ability being implemented in a different way than it's described (the text says it increases the unit's defense, but in reality it's decreasing the opponent's chance to hit). So it's not really clear that the blue unit in the above screenshot is being affected by the Formation ability because of the enemy they're attacking vs because they themselves have the Formation ability.
> 
> @newfrenchy83 If the Formation ability were implemented like a regular weapon special rather than as an ability, then the red unit in the above screenshot would have "formation" listed under its attack rather than the blue unit, correct?

Correct

newfrenchy83:
i work to replace "name"by special_effect tags for describe the effect on attack instead to add name of ability

PR #6111, opened 2021-09-22
extent addition of the names of the abilities used as weapons in the attack dialog window to apply_to=opponent(1.16)

newfrenchy83:
If a weapon with apply_to=opponent is used, this name will appear in opponent attack affected.


newfrenchy83:
@Pentarctagon i hope what this PR will be commited for 1.16.1.

Pentarctagon:
Ah, right.

PR #6114, opened 2021-09-23
Main menu language button v3

irydacea:
Since @stevecotton expressed interest in the iteration of the Language button I mentioned [in the middle of the discussion in #4531](https://github.com/wesnoth/wesnoth/issues/4531#issuecomment-842665100), I decided to package that and that alone into a PR that could potentially be merged into 1.16 without any string changes whatsoever.

Screenshot_20210923_120951

This new iteration of the button displays the name of the currently selected (or *guessed*, if it's System Default) language as its label, while retaining the icon for aesthetic purposes. It also increases the bottom and edge margins for that UI grid row to make it slightly more elegant than the previous "everything stuck in the corner" approach. There are otherwise no changes in functionalityhere.

image

image

image

(Paging @Vultraz and @nemaara as well.)

CC #4531

### DIFF
--- a/src/gui/dialogs/title_screen.cpp
+++ b/src/gui/dialogs/title_screen.cpp
@@ -38,6 +38,7 @@
 #include "gui/dialogs/preferences_dialog.hpp"
 #include "gui/dialogs/screenshot_notification.hpp"
 #include "gui/dialogs/simple_item_selector.hpp"
+#include "language.hpp"
 #include "log.hpp"
 #include "preferences/game.hpp"
 //#define DEBUG_TOOLTIP
@@ -58,6 +59,8 @@
 
 #include <algorithm>
 #include <functional>
+
+#include <boost/algorithm/string/erase.hpp>
 
 static lg::log_domain log_config("config");
 #define ERR_CF LOG_STREAM(err, log_config)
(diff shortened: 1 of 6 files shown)


irydacea:
Side note: the English language menu entries are the only ones that use square brackets instead of round parentheses anywhere (e.g. `English [US]` instead of `English (US)`). I'm not sure how I feel about this.

Vultraz (approved):

nemaara (approved):

Wedge009:
As mentioned in #wesnoth-dev, I think this is a great improvement. I also don't think `[]` vs `()` is a big issue - either looks fine to me.

irydacea:
To me it seems annoyingly inconsistent and also unusual in a broader context (i.e. outside of wesnoth).

Wedge009:
I suppose what I'm saying is that it's not worth cancelling the whole change over this issue. Outside of Wesnoth I've seen `[]` used as well, but I suppose it would be nice to keep it consistent if possible.

irydacea:
I'd never think of cancelling this over that. Plus it's a trivial WML change.

src/gui/dialogs/title_screen.cpp
```diff
@@ -311,6 +314,25 @@ void title_screen::pre_show(window& win)
 		}
 	});
 
+	auto* lang_button = find_widget<button>(&win, "language", false, false);
+
+	if(lang_button) {
```
stevecotton, line 319: Could use a C++17 style declaration of the variable inside the conditional.

stevecotton:
Looks good to me (sorry about not reviewing until I saw the ping on IRC, I thought this was only waiting for Vultraz).

irydacea:
Decided to remove the (currently) unused button variations to make this more appealing for 1.16.

Also replaced the aforementioned punctuation in the English translation names.

Issue #4258, opened 2019-08-21
1.15 should leadership affect allied sides


stevecotton:
In 1.15.0 and 1.15.1, `[leadership]` affects units on allied sides in addition to units on the same side. This was an unintentional change caused by #4257, but the question now is whether to make it match the old behavior, or whether to keep the new behavior.

stevecotton:
Tagged `MP` as it will affect balance in 2v2 matches.

gfgtdf:
I think it'd be better if leadership affected allied sides, (iirc i suggested this change when we were talking about the LoW mp version, where leadersip is quite weak because of this, but iirc our plan was to remove LoW eventually so that probably not the strongest point here), i'm not really a mp expert but i'd expect the 1.14 leadership behaviour kinda discourages mixing troops (and instead each player fighting on their own flank).

Also i'd assume that leadership is already weaker in 2vs2 than in a 1vs1 match with the same amount of treep simply becasue you cannot move your leadership unit along with the attacks or retreat him after if the leadersip unit and the other units are on different sides.

It'd be nice to hear more about the impact on 2vs2 because maybe there is something i didnt see maybe some strategy of one team having  laedersip unit and the other team recruiting lvl0s woudl be too strong? Hard to say from here.

Lastly even if there is a reason why leadersip is too good in 2vs2, we can also do a compromise like only increasing up to 25% for allied units.

soliton-:
I would not change how default leadership works and needlessly break UMC. If we want allied leadership that's already been possible.

I don't think allied leadership will change balance too much. As gfgtdf pointed out it's hard to use intentionally so it's not going to be applicable very often. I would certainly not make it more complicated by having allied leadership work differently.

Another criteria is does it make sense? I think usually it does not.

Konrad22:
(It's probably not important, but this change also affects 'inspire' from THoT.)

jostephd:
> It'd be nice to hear more about the impact on 2vs2 because maybe there is something i didnt see maybe some strategy of one team having laedersip unit and the other team recruiting lvl0s woudl be too strong? Hard to say from here.

Ask on the forum? There are probably players who could weigh in on this who don't follow github.

sevu:
The impact is relatively low. I'd see is as a nice side-effect of the code change.

soliton-:
It'd help if you explain why it's nice. That's what this issue is about.

Perhaps I should repeat that it was already possible before. The code change was not needed to make it happen.

[...]

PR #6117, opened 2021-09-26
Simplifying the description of leadership

Discontinuum:
I think that the current description of leadership is too wordy and hard to understand for no reason. So my proposal is to rewrite it more concise. I propose this variant as a starter, but it's not final by any means and I'm glad to hear any further suggestion on that

### DIFF
--- a/data/core/macros/abilities.cfg
+++ b/data/core/macros/abilities.cfg
@@ -198,9 +198,9 @@ This ability will not cure an affected unit of poison, however, only delay its e
         cumulative=no
         name= _ "leadership"
         female_name= _ "female^leadership"
-        description= _ "This unit can lead your own units that are next to it, making them fight better.
+        description= _ "This unit can lead your troops in battle.
 
-Adjacent own units of lower level will do more damage in battle. When a unit adjacent to, of a lower level than, and on the same side as a unit with Leadership engages in combat, its attacks do 25% more damage times the difference in their levels."
+All your adjacent troops of a lower level deal 25% more damage times the difference of their levels."
         special_note={INTERNAL:SPECIAL_NOTES_LEADERSHIP}
         affect_self=no
         [affect_adjacent]


data/core/macros/abilities.cfg
```diff
@@ -198,9 +198,9 @@ This ability will not cure an affected unit of poison, however, only delay its e
         cumulative=no
         name= _ "leadership"
         female_name= _ "female^leadership"
-        description= _ "This unit can lead your own units that are next to it, making them fight better.
+        description= _ "This unit can lead your troops in battle.
```
kabachuha, line 201: what about "adjacent troops"? I think an indication what the units are only one hex away is needed to avoid confusion.

Discontinuum: It's described in the next line, "adjacent troops"

Wedge009:
How is it hard to understand? Also, I suspect the detail on _own units_ is necessary because of a possible future change in #4258.

Discontinuum:
I think this kind of structure is not very simple 

Like, it's not what you can read and understand instantly, you need to stop and think a bit to figure it out. It seems too redundant to be clear. 

Also its side effects are poor translations. Not enough translators are brave enough to put it simple irrespective to what in the text, so usually they try to follow the original text closely and the result is artificial and hard to read.

About "own", maybe, I don't know about proposals. It's a minor detail, to add "own", my point is in simplifying the whole structure.

AI0867:
> I think this kind of structure is not very simple
> 
> Like, it's not what you can read and understand instantly, you need to stop and think a bit to figure it out (it might seem simple with underlining, but without it I doubt). It seems too redundant to be clear.

No part of that sentence is redundant. Every adjective (phrase) explains a different part of how leadership works. It is indeed that complicated.

A leadership unit affects units that are

- adjacent to it
- on the same side (team or side number?)
- of lower level

by an amount that scales by the difference in levels.

A bulleted list like this might be clearer though.

sheepo99:
> I think this kind of structure is not very simple

I agree that this is pretty poor writing.

Discontinuum:
> No part of that sentence is redundant. Every adjective (phrase) explains a different part of how leadership works. It is indeed that complicated.

It's redundant in a sense that it tried to be too much explicit. To express these ideas you basically need three things: "adjacent", "your" ("your own", "your allied", whatever), "lower level". But that text is tried to add each time the relation to a leadership unit: "adjacent to *that unit*", "the same side as *that unit*", "of a lower level than *that unit*". But it clogs the resulting sentence too much and it doesn't become clearer. No need to always highlight that its relation to *that unit*, it's graspable from the context.

> A bulleted list like this might be clearer though.
I thought about it too, in a single line that's just impossible to scan instantly. But I still believe it's worth considering to rewrite it more concise.

Wedge009:
Suggest rewriting the whole description if it's a problem rather than trying to just fix things piece-meal.

AI0867:
It tries to be explicit because otherwise people will gloss over parts of the sentence and come complain in the forums about leadership not working because of a requirement they didn't read.

One might say "adjacent, lower-level units of your own side", but that's already pretty much the first sentence of the paragraph, so I agree there is some redundancy there.

The rewrite doesn't look bad to me, but we might want to run it by some people unfamiliar with it, to see if they get all the information out of it.

CelticMinstrel:
Hmm, I _think_ it looks good? It seems a bit unclear which units are meant with "difference in levels", but the old description wasn't any better in that respect.

max-torch:
> All your adjacent troops

Adjacent to what? To each other? You have to specify _adjacent to a unit with leadership_. There is actually an ability that affects troops _adjacent to each other_ called _formation_.
> All your adjacent troops

I think the word "units" was more consistent with existing wesnoth writing and terminology.
> deal 25% more damage times the difference of their levels."

In my opinion, "their" in this case is an unclear pronoun. Does "their" mean among the adjacent units or the units and the unit with leadership? For example, you could have a lvl2 leader having both a lvl0 and a lvl1 leader adjacent to it. So we have to be explicit that "their" is the leader and each unit.

CelticMinstrel:
Then maybe "times their level difference with the leader"?

max-torch:
> Then maybe "times their level difference with the leader"?

Clearer. yes.

Discontinuum:
> Adjacent to what? To each other? 

I understand this feeling that we should try to explain everything but still even if we absolutely have to do it, we need to think hard about how to make it in the most concise way. There is a fine borderland here between too obscure and too detail, in both cases some players will gloss over and always somebody asks questions about how it or another ability works. So for a starting variation I decided not to try to explain things deductible from the context. Some people may not understand, but some will not understand it anyway, be it short or long, concise or wordy. Something in-between we should find

data/core/macros/abilities.cfg
```diff
@@ -198,9 +198,9 @@ This ability will not cure an affected unit of poison, however, only delay its e
...
 
-Adjacent own units of lower level will do more damage in battle. When a unit adjacent to, of a lower level than, and on the same side as a unit with Leadership engages in combat, its attacks do 25% more damage times the difference in their levels."
+All your adjacent troops of a lower level deal 25% more damage times the difference of their levels."
```
max-torch, line 203: Alright, how does this sound:
"Units with leadership increase the damage of each player-owned adjacent lower-level unit by 25% for each difference in level."
Which sounds better for the last clause "for each difference in level" or "times the difference of their levels"?

max-torch:
> > Adjacent to what? To each other?
> 
> I understand this feeling that we should try to explain everything but still even if we absolutely have to do it, we need to think hard about how to make it in the most concise way. There is a fine borderland here between too obscure and too detail, in both cases some players will gloss over and always somebody asks questions about how it or another ability works. So for a starting variation I decided not to try to explain things deductible from the context. Some people may not understand, but some will not understand it anyway, be it short or long, concise or wordy. Something in-between we should find

I actually agree with this, I don't know why I was being so overly explicit previously, I think my brain was on coding mode, which requires that everything be explicit.

data/core/macros/abilities.cfg
Discontinuum: That might be a good touch. 

I wonder if "a unit with leadership increases" is a bit better, so  both the leader unit and the underling unit will be singular and "their" in "times the difference of their levels" will clearly relate to the only possible pair in the context of this phrase: the difference of the leader and the underling levels.

data/core/macros/abilities.cfg
nemaara, line 201:
```suggestion
        description= _ "This unit can lead other troops in battle.
```

Needs to not have "your", enemy units can have leadership too (yes I know the original was not exactly right).

data/core/macros/abilities.cfg
nemaara, line 203:
```suggestion
All adjacent lower-level units from the same side deal 25% more damage for each difference in level."
```

nemaara (changes requested):

Issue #6123, opened 2021-09-27
SotA S04 Becalmed: Add note that the player can move between decks


max-torch:
### Text Issues: Typos, grammar mistakes, suggested improvements

* English

* Secrets of the Ancients S04 Becalmed

### Issue
When you first play the scenario, when the movable hexes of a unit are highlighted, you might not think to check that there are also movable hexes at the lower deck and later be surprised that the enemy is popping out of the stairs (it happened to me). More likely to occur in small resolutions or zoomed and small screens, where you don't see the lower deck on screen.

* Please quote the text verbatim. You can use `**bold**` to point out the mistaken part:
The scenario objectives have the following notes:
```
            [note]
                description= _ "You can recruit from anywhere on the two highest decks."
            [/note]
            [note]
                description= _ "Barrels act as villages."
            [/note]
            [note]
                description= _ "Right-click to simplify the ship graphics."
            [/note]
```

* _(optional)_ What do you propose to change the text to?
Add one more note that says
```
            [note]
                description= _ "Player and enemy units can move between lower and upper decks of the ship through the stairs. "
            [/note]
```
### Reference Image
At the default zoom level, player will not see that there are highlighted movable hexes at the lower deck unless they scroll down.
![image](https://user-images.githubusercontent.com/41433185/134909072-8c9a6407-3523-416f-8c2f-e9e1b10b9335.png)

Issue #6121, opened 2021-09-27
SotA S2 Dark Business: Add additional note about villages and guardhouse scenario text


max-torch:
## Text Issues: Typos, grammar mistakes, suggested improvements
* English
* Secrets of the Ancients S2 Dark Business

## First Issue
When the Guardhouse is in the player's possession the ability to recruit skeletons is granted. And this ability is lost if the guardhouse is recaptured by the enemy. The exact details of this may not be obvious. For example, I thought you needed to have any of your units actually occupying the guardhouse to gain the ability - I didn't know the condition was that it merely had to be in your possession.

### Existing text
This is the instruction we get about this:
            `message= _ "I would do better if I had weapons, and I bet I can find some in the guardhouse. Go take it, my companions."`
and
            `message= _ "You can now recruit skeletons and skeleton archers!"`

### Proposal
Let's make it clear that you gain the ability when the guardhouse is in your possession. Add more notes to the Objectives (new note highlighted bold):
            [note]
                description= _ "Right-click on a grave adjacent to your leader to raise a unit."
            [/note]
            [note]
                description= _ "Each grave can be used only once."
            [/note]
            [note]
                description= _ "This scenario takes place entirely at night."
            [/note]
            **[note]
                description= _ "You will gain an advantage if the village labeled 'Guardhouse' is in your possession."
            [/note]**

## Second Issue
You are given the option to recruit or recall a bat, for 20 gold. It's not clear why if you don't realize there are villages _outside_ of the graveyard. If you like, you can make this dialog only appear on easy and normal difficulty.

### Existing text
            speaker=Ardonna
            message= _ "Hey, come back! I didn’t mean to frighten you."
        [/message]
        [message]
            speaker=Ardonna
            message= _ "Maybe I could attract it if I jingled some coins."

### Proposal
            speaker=Ardonna
            message= _ "Hey, come back! I didn’t mean to frighten you."
        [/message]
        [message]
            speaker=Ardonna
            message= _ "Maybe I could attract it if I jingled some coins. (There are villages outside of the graveyard)."

###Reference Image
![image](https://user-images.githubusercontent.com/41433185/134900321-0d602d8f-52df-4e40-bf41-124af438edb3.png)

cooljeanius:
I don't think "(There are villages outside of the graveyard)" necessarily makes sense for Ardonna to say; I think that would make more sense as a narrator message, or a hint in the objectives
(I didn't notice them either, tbh)

[...]

Issue #6120, opened 2021-09-27
[SotA Dialog] S06 Following the Shadow: provide better dialog hint about tomato surprise


max-torch:
### Text Issues: Typos, grammar mistakes, suggested improvements

* English language
* Secrets of the Ancients S06 Following the Shadow

### The Issue
We have this commentary at the top of the scenario cfg
```
# A large number of bats spawn in the cave under these two conditions:
# (1) Six non-bat units are in the cave (Bats don't stomp on the floor.)
# (2) Ardonna has been in the cave for two turns in a row. (That gets her solidly into
#     the cave, but not close to the exit. If she is alone, it's the player's fault when
#     the bats kill her.)
```
Sure the player should have to be careful and even expect tomato surprises, but I doubt that it's a strong case to expect the player to expect such a tomato surprise that involves an entire army of units instantly spawning on all sides of the leader (around the whole cave) resulting in an instant defeat. Even in other campaigns where there have been tomatoes or bats spawning, for example, The Mage in Dead Water, the surprise units still somewhat come from in front of the player.

* Please quote the text verbatim. You can use `**bold**` to point out the mistaken part:
The dialog does give caution and gives some form of warning:
           ` message= _ "If there is one bat, there are certainly more. We should be careful."`
But given the nature and degree of the tomato surprise, it justifies further hinting. Make it a clear i-told-you-so.

* _(optional)_ What do you propose to change the text to?
 ` message= _ "If there is one bat, there are certainly more. We should be careful. It would be foolish to explore this cave alone."`

### Reference Image
![image](https://user-images.githubusercontent.com/41433185/134896361-979ec408-7886-4d84-84f6-5696429e3f86.png)

Wedge009:
This one I remember from 1.14 and I didn't find it much fun either.

cooljeanius:
Agreed, the surprise bats got me, too, when I played this (I'm on 1.14)... I think the proposed text change would be an improvement. 

cc @beetlenaut

max-torch:
> This one I remember from 1.14 and I didn't find it much fun either.
> 
> Edit: By the way, since you're giving feedback on the campaign design itself here and in #6119, you may be better off doing so in https://forums.wesnoth.org/viewtopic.php?t=40545.

I will make Git issues and then I will post on the forums and provide Git issue links on the posts.

knyghtmare:
[...]

Issue #6119, opened 2021-09-27
[SotA Final Scenario Dialog] The protagonist says the previous scenario was 'easy'


max-torch:
### Text Issues: Typos, grammar mistakes, suggested improvements

* English Language Text

* SotA Final Scenario

* This [dialog](https://github.com/wesnoth/wesnoth/blob/ca90694141d8e2c5d2d1ea8afcfb071e9a439f12/data/campaigns/Secrets_of_the_Ancients/scenarios/21_Against_the_World.cfg#L632-L639) appears when Dwarves emerge onto the map:
```
        [message]
            speaker=Ras-Tabahn
            message= _ "Look! The dwarves circled around in front of us."
        [/message]
        [message]
            speaker=Ardonna
            message= _ "That’s why getting through Knalga was so easy."
        [/message]
```
The issues with this are:
* The second-to-the-last scenario, North Knalga, is one of the most challenging scenarios in the entire campaign.
* The amount of dwarves that appear in the final scenario is small and looks like a small scout force compared to the massive army of dwarves you fight in the previous scenario.

* _(optional)_ What do you propose to change the text to?
```
        [message]
            speaker=Ardonna
            message= _ "On the bright side, if they did not choose to split their forces and surprise us here we might have faced greater opposition back in Knalga!"
        [/message]
```

### Reference Images
Back in Knalga, first batch of dwarves:
![image](https://user-images.githubusercontent.com/41433185/134893061-33dd1d67-2d25-4d25-a198-ef8d839935c5.png)
Back in Knalga, _second_ batch of dwarves:
![image](https://user-images.githubusercontent.com/41433185/134893888-73eaea23-383b-4157-87b7-546267a59d00.png)

In the Ford of Abez, one small batch of dwarves:
![image](https://user-images.githubusercontent.com/41433185/134893334-ac3f1115-915e-4d5c-bf6e-8f442e4842da.png)

Wedge009:
I haven't played the 1.15 revision to SotA yet, but I wonder if it's meant to be a sarcastic remark or just reflection of what's supposed to be in the story. I often find - in games generally, not specifically Wesnoth - that game-play and story don't always have to line-up.

cooljeanius:
yeah I thought that seemed off when I finally got around to beating SotA recently, too (via cheating, but still)...

cc @beetlenaut

knyghtmare:
Perhaps this word change?
```
        [message]
            speaker=Ardonna
[...]

PR #6129, opened 2021-09-28
SotA Proofread Pass

max-torch:
Only proofreading. No major edits to tone, meaning, or dialog. The `wmlxgettext` module was used so that no text would be overlooked. The writing is already high quality and there was not much to nitpick.

### DIFF
--- a/data/campaigns/Secrets_of_the_Ancients/scenarios/14_Entering_the_Northlands.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/scenarios/14_Entering_the_Northlands.cfg
@@ -28,7 +28,7 @@
             title_alignment=left
             story= _ "24 IV, 23 YW
 
-We reached the Ford of Abez as the dawn light spread behind the mountain peaks to the north. As we waded across, the dark water reached to our waists, which made for slow going. I dared not imagine what manner of tentacled creature lurked below, but we completed the crossing without incident. The water was chill with snowmelt, so our teeth were chattering when we reached the far shore. The rising sun was warm though, and had us dried in short order. Despite this, Ras-Tabahn was not well off. He collapsed in the sand and went into a paroxysm of coughing the like of which I had never seen. When he could speak again, he assured me that it was only on account of the cold, and I should not be unduly troubled. I am uneasy all the same."
+We reached the Ford of Abez as the dawn light spread behind the mountain peaks to the north. As we waded across, the dark water reached up to our waists, which made for slow going. I dared not imagine what manner of tentacled creature lurked below, but we completed the crossing without incident. The water was chill with snowmelt, so our teeth were chattering when we reached the far shore. The rising sun was warm though, and had us dried in short order. Despite this, Ras-Tabahn was not well. He collapsed in the sand and went into a paroxysm of coughing the like of which I had never seen. When he could speak again, he assured me that it was only on account of the cold, and I should not be unduly troubled. I am uneasy all the same."
         [/part]
         {JOURNEY_PART 14}
     [/story]
--- a/data/campaigns/Secrets_of_the_Ancients/scenarios/14_Entering_the_Northlands.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/scenarios/14_Entering_the_Northlands.cfg
@@ -28,7 +28,7 @@
             title_alignment=left
             story= _ "24 IV, 23 YW
 
-We reached the Ford of Abez as the dawn light spread behind the mountain peaks to the north. As we waded across, the dark water reached up to our waists, which made for slow going. I dared not imagine what manner of tentacled creature lurked below, but we completed the crossing without incident. The water was chill with snowmelt, so our teeth were chattering when we reached the far shore. The rising sun was warm though, and had us dried in short order. Despite this, Ras-Tabahn was not well. He collapsed in the sand and went into a paroxysm of coughing the like of which I had never seen. When he could speak again, he assured me that it was only on account of the cold, and I should not be unduly troubled. I am uneasy all the same."
+We reached the Ford of Abez as the dawn light spread behind the mountain peaks to the north. As we waded across, the dark water reached to our waists, which made for slow going. I dared not imagine what manner of tentacled creature lurked below, but we completed the crossing without incident. The water was chill with snowmelt, so our teeth were chattering when we reached the far shore. The rising sun was warm though, and had us dried in short order. Despite this, Ras-Tabahn was not well. He collapsed in the sand and went into a paroxysm of coughing the like of which I had never seen. When he could speak again, he assured me that it was only on account of the cold, and I should not be unduly troubled. I am uneasy all the same."
         [/part]
         {JOURNEY_PART 14}
     [/story]
--- a/data/campaigns/Secrets_of_the_Ancients/scenarios/15_Mountain_Pass.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/scenarios/15_Mountain_Pass.cfg
@@ -22,7 +22,7 @@
             #po: "consumption" is an old word for tuberculosis
             story= _ "6 V, 23 YW
 
-We have reached a pass through the mountains that Ras-Tabahn knew of. The air is cool in these upper reaches. Mist rises from the snow in the afternoon, and the icicles drip the purest water from rock overhangs. I find it all refreshing, but Ras-Tabahn’s coughing fits are frequent. Sometimes he is reduced to speaking in short phrases due to lack of breath. Of course, he is quite old, fifty at least, but I am also certain that he is ill. His thinness leads me to believe it is consumption, but he refuses to discuss it, preferring to deny the severity of his problem."
+We have reached a pass through the mountains that Ras-Tabahn knew of. The air is cool in these upper reaches. Mist rises from the snow in the afternoon, and the icicles drip the purest water from rock overhangs. I find it all refreshing, but Ras-Tabahn’s coughing fits are frequent. Sometimes he is reduced to speaking in short phrases due to lack of breath. Of course, he is quite old — at least fifty — but I am also certain that he is ill. His thinness leads me to believe it is consumption, but he refuses to discuss it, preferring to deny the severity of his problem."
         [/part]
         {JOURNEY_PART 15}
     [/story]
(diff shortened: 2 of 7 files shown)


cooljeanius:
While using `wmlxgettext`, were there any strings that you thought could be clarified for translators by adding `#po:` comments that'd go in the potfile? Examples: unclear pronoun reference, unclear speaker gender, unclear number for things that can be either singular or plural, etc.

max-torch:
> While using `wmlxgettext`, were there any strings that you thought could be clarified for translators by adding `#po:` comments that'd go in the potfile? Examples: unclear pronoun reference, unclear speaker gender, unclear number for things that can be either singular or plural, etc.

I wasn't keeping an eye out for this, so I would have to go through everything again to see if there are any additional translator notes that would be good to add - but what I can say was that the existing translator notes were already pretty good here - genders, plurals, archaic words, were among the existing #po notes.

data/campaigns/Secrets_of_the_Ancients/scenarios/18_Abandoned_Outpost.cfg
```diff
@@ -426,7 +426,7 @@ Now the process is complete for both of us. I posted a guard so nothing could ha
 
         [message]
             speaker=Ardonna
-            message= _ "We have conquered death! Let us bring the news back to Wesnoth. We should to let them know that nobody ever need die again."
+            message= _ "We have conquered death! Let us bring the news back to Wesnoth. We should - to let them know that nobody ever need die again."
```
Wedge009, line 429: It's hard to tell what the character is, but it looks like a regular hyphen to me - should be an em dash according to https://wiki.wesnoth.org/Typography_Style_Guide

Wedge009:
Overall looks good to me. Saves me from doing the same when I replay SotA, which I'm not planning to do until 1.16. I'm just not sure if all the additional commas in S15 are necessary - I'm not super-strong on the grammatical rules there - but it's not incorrect from what I can see.

data/campaigns/Secrets_of_the_Ancients/scenarios/18_Abandoned_Outpost.cfg
CelticMinstrel: This sentence sounds awkward regardless of whether there's a dash there…

data/campaigns/Secrets_of_the_Ancients/scenarios/15_Mountain_Pass.cfg
```diff
@@ -22,7 +22,7 @@
...
             story= _ "6 V, 23 YW
 
-We have reached a pass through the mountains that Ras-Tabahn knew of. The air is cool in these upper reaches. Mist rises from the snow in the afternoon, and the icicles drip the purest water from rock overhangs. I find it all refreshing, but Ras-Tabahn’s coughing fits are frequent. Sometimes he is reduced to speaking in short phrases due to lack of breath. Of course he is quite old — 50 at least — but I am also certain that he is ill. His thinness leads me to believe it is consumption, but he refuses to discuss it, preferring to deny the severity of his problem."
+We have reached a pass through the mountains that Ras-Tabahn knew of. The air is cool in these upper reaches. Mist rises from the snow in the afternoon, and the icicles drip the purest water from rock overhangs. I find it all refreshing, but Ras-Tabahn’s coughing fits are frequent. Sometimes he is reduced to speaking in short phrases due to lack of breath. Of course, he is quite old, fifty at least, but I am also certain that he is ill. His thinness leads me to believe it is consumption, but he refuses to discuss it, preferring to deny the severity of his problem."
```
CelticMinstrel, line 25: I think I prefer this with the dashes. The added comma after "of course" and writing out "fifty" in full are good though.

Wedge009: Works for me, and resolves what I mentioned earlier in the general comments.

data/campaigns/Secrets_of_the_Ancients/scenarios/14_Entering_the_Northlands.cfg
```diff
@@ -28,7 +28,7 @@
...
+We reached the Ford of Abez as the dawn light spread behind the mountain peaks to the north. As we waded across, the dark water reached up to our waists, which made for slow going. I dared not imagine what manner of tentacled creature lurked below, but we completed the crossing without incident. The water was chill with snowmelt, so our teeth were chattering when we reached the far shore. The rising sun was warm though, and had us dried in short order. Despite this, Ras-Tabahn was not well. He collapsed in the sand and went into a paroxysm of coughing the like of which I had never seen. When he could speak again, he assured me that it was only on account of the cold, and I should not be unduly troubled. I am uneasy all the same."
```
nemaara, line 31: This line isn't necessary to change, I think the original is fine.

data/campaigns/Secrets_of_the_Ancients/scenarios/15_Mountain_Pass.cfg
nemaara: Agree with celmin.

data/campaigns/Secrets_of_the_Ancients/scenarios/18_Abandoned_Outpost.cfg
nemaara: If you want the dash you can get rid of the "we should" and merge it with the previous sentence. This edit otherwise doesn't work for me.

nemaara (changes requested):

data/campaigns/Secrets_of_the_Ancients/scenarios/14_Entering_the_Northlands.cfg
Wedge009: Often I find 'up' can be redundant but perhaps not in this case. But it's okay for me without it.

The other change I think is a good one. If you don't like it, how about 'Ras-Tabahn was not faring well'?

nemaara: I'd rather have Ardonna talk plainly. She seems like she wouldn't use extra words when it's not necessary.

Wedge009: In that case, all the more reason to just say 'Ras-Tabahn was not well'.

data/campaigns/Secrets_of_the_Ancients/scenarios/18_Abandoned_Outpost.cfg
CelticMinstrel: Without the dash, you'd probably need to drop "to", as in "we should let"…

data/campaigns/Secrets_of_the_Ancients/scenarios/14_Entering_the_Northlands.cfg
max-torch: Will accept removing the 'up', but can we remove the 'off'? My understanding is that anytime you say 'Well off' it means 'rich'.

data/campaigns/Secrets_of_the_Ancients/scenarios/18_Abandoned_Outpost.cfg
cooljeanius: How about a colon?

max-torch: Will change the text to this:
"We have conquered death! Let us bring the news back to Wesnoth! Let all in this world know that nobody ever need die again."
> image

"ever need die" is grammatically incorrect but we will permit it for tone or archaism.

data/campaigns/Secrets_of_the_Ancients/scenarios/14_Entering_the_Northlands.cfg
nemaara: Oops, I was super tired while reading this apparently. Yeah, please remove the off.

Wedge009: That is already done, so marking as resolved. I don't think there are any outstanding changes pending.

data/campaigns/Secrets_of_the_Ancients/scenarios/18_Abandoned_Outpost.cfg
CelticMinstrel: > "ever need die" is grammatically incorrect but we will permit it for tone or archaism.

How is it grammatically incorrect? The phrasing "nobody need die again" is grammatically valid; at most the placement of "ever" may be strange. It doesn't feel wrong like this, but it might fit better if it were placed between "need" and "die".

> How about a colon?

Ehhh, I dunno. A colon is almost never correct in speech, as far as I know. There's probably some corner cases where it applies, but I'm not sure this is one of them…

nemaara (approved):

data/campaigns/Secrets_of_the_Ancients/scenarios/18_Abandoned_Outpost.cfg
max-torch: > How is it grammatically incorrect?

Hmm, yeah. I just said it was grammatically incorrrect because - Grammarly, ugh. I try not to rely on it and only use for QA reasons, at the end of everything, and with a grain of salt. Anyway, you can see the screenshot where it was underlined red so you see why I ended up saying it was incorrect.

CelticMinstrel: It's something of an unusual construction, so it wouldn't surprise me if the grammar checker doesn't understand it. But that's also part of the reason to keep it here, for the archaic flavour (it was likely more common in the past).

Vultraz (approved):

Wedge009:
@max-torch Anything you still want to change before merging?

max-torch:
> @max-torch Anything you still want to change before merging?

I soon plan on dealing with my other suggestions #6123, #6121, #6120, and #6119. I suppose those can go into a separate pull request than this, since those changes are additions that are beyond just proofreading?

Wedge009:
Yeah, probably better to keep it separate since those four mainly deal with game-play elements. Although, can't merge at the moment anyway, due to the failing check (looks like a transient failure, though?).

Issue #6125, opened 2021-09-27
UtBS S2 Across the Harsh Sands Pinnacle Rock is covered during dialog


max-torch:
### Game and System Information
 - 1.15.18
 - Steam
 - Windows 10

### Describe the bug
At the start of the scenario during the dialog with Garak, the map scrolls to the top to show Pinnacle Rock, while Garak explains that we should head to it. The lower the resolution or smaller the screen or zoom level, the worse the issue gets, as the dialog covers almost the entire Pinnacle Rock.

### To Reproduce
Load S2 of Under the Burning Suns

### Expected behavior
The map should scroll to Pinnacle Rock, stopping when Pinnacle Rock is precisely centered onscreen and not covered, completely or partially, by dialog.

### Screenshots
![image](https://user-images.githubusercontent.com/41433185/134913709-ccaf45cf-162a-458d-8fbb-a924320d52aa.png)

### Savefiles
Savefile of start of scenario save of S2: 
[UtBS-Across the Harsh Sands.gz](https://github.com/wesnoth/wesnoth/files/7236400/UtBS-Across.the.Harsh.Sands.gz)

### Additional context
The problem gets worse at lower resolutions/zoom levels.

Wedge009:
Contrary to the report, I could only replicate the issue at _high_ zoom levels. 2560x1440 monitor.

max-torch:
We can say "the degree of the problems varies depending on screen size, resolution, zoom levels."

Wedge009:
Closed via #6135.

PR #6135, opened 2021-09-30
Pinnacle rock map label covered during dialog fix

Descacharrado:
Issue #6125 - UtBS S2 Across the Harsh Sands Pinnacle Rock is covered during dialog fix. Now the camera will scroll to the tile where the "Pinnacle Rock" map label is.


Wedge009:
Well spotted - perhaps the rock was moved at some point but the `scroll_to` co-ordinates weren't updated.

@max-torch Do you have time to check this?

Wedge009:
I checked it before the tests finished and it looked right to me. Thanks.

max-torch:
> @max-torch Do you have time to check this?
I can also confirm it's working for me.

Issue #6122, opened 2021-09-27
[FR] [1.17/1.18] Allow Scenario Editor's unit tool to Include Unit Variations.


knyghtmare:
Currently, the scenario editor has an unit placement tool which does not have any option to switch between unit variations or add them. I think UMC authors using the scenario editor would greatly appreciate such an feature. Maybe it can be added inside the right-click menu when the placed unit is on the map, like how facing, generated unit ID, and upkeep checkbox is added.

FrostBlade5:
I just hope in future updates they will add those unit-variations into the map-editor.
Would be amazing.

FrostBlade5:
Unit Variations: Walking Corpse and Soulless:
https://i.postimg.cc/CxX1WSx5/4.png

CelticMinstrel:
I think it would be preferable to add unit variations as separate entries to the units palette. There are a lot more unit variations in Wesnoth besides the walking corpses - wolves, scorpions, and horses are some examples of units with variations.

FrostBlade5:
Exactly what came to my mind. Another unit palette for variations only.
I dont know how many variations exist in total. personally im only interested to the zombie/infected/undead variations.
I think those alone could fill up a whole palette.

Not sure if every single unit has its own undead-variant but im sure alot of them do.

CelticMinstrel:
That's not what I meant at all. I don't think there should be a separate palette to pick variations; the scenario editor should just treat variations the same as base units. So for example, you'd see the three wolf variants right beside each other in the monsters section, and the undead section would have every type of walking corpse alongside the ghouls and skeletons.

cooljeanius:
Does #7719 address this at all?

Pentarctagon:
It does not.

PR #6138, opened 2021-10-01
1.16 - Map-Editor missing Units

FrostBlade5:
Map-Editor missing Units

Hello, im here to make a Suggestion!
Excuse me if the Topic does not fit to this Thread, feel free to move it.
Personally i think this Suggestion fits into Developement.

When i open the ingame Map-Editor and create a new Scenario i am able
to place all types of Unites, expect the unit-list is incomplete.
See Screenshot: https://postimg.cc/bsxXgVM3 (i beleve those are called unit-variations)

Let me explain: when a Undead-unit kills another unit, instead of dying
they become infected and will also turn into an undead unit.

Those are some of the Units im talking about: https://postimg.cc/yD37PNVz
Can you add the missing Units into the Map-Editor, please?
Would be awesome to be able and use those units in the map-editor!!

I have already reported this request on the wesnoth forum here: https://forums.wesnoth.org/viewtopic.php?f=6&t=54937
But since nobody so far has replyed what is going to happen i will try my luck here on github.


Pentarctagon:
I assume this was meant to be opened as an issue rather than a PR.

FrostBlade5:
> 
> 
> I assume this was meant to be opened as an issue rather than a PR.

Well, i never used Github im new basically dont know whats the difference. Pull requests might be more important? In this case, for me personally id love to be able an use unit variations in the map editor, thats all.

An issue regarding this topic has already been opened by knyghtmare: https://github.com/wesnoth/wesnoth/issues/6122

My question is, will this issue dissapear within all the other 900 issues or will someone take care of it? I honestly beleve it is a very doable thing with less effort, really.

Pentarctagon:
Issues are for reporting bugs and making feature requests.  PRs are for actually implementing those bug fixes or features.

PR #6143, opened 2021-10-02
TRoW Proofread Pass

max-torch:
For the campaign The Rise of Wesnoth. For Approval. Minor edits and proofreads. `wmlxgettext` was used so as to ensure all text was reviewed.

### DIFF
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/01_A_Summer_of_Storms.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/01_A_Summer_of_Storms.cfg
@@ -144,7 +144,7 @@
         name=start
         [message]
             speaker=narrator
-            message= _ "The trouble seems to have finally reached to the heart of the Isle, to the isolated lands of King Eldaric IV."
+            message= _ "The trouble seems to have finally reached the heart of the Isle, to the isolated lands of King Eldaric IV."
             image=wesnoth-icon.png
         [/message]
         [message]
@@ -157,7 +157,7 @@
         [/message]
         [message]
             speaker=King Eldaric IV
-            message= _ "You’re showing initiative, son! I’m proud of you! Yes, you may lead our forces to battle, it is time — but I’ll stay near to keep an eye on you. There is more to this raid than meets the eye, I think."
+            message= _ "You’re showing initiative, son! I’m proud of you! Yes, you may lead our forces in battle, it is time — but I’ll stay near to keep an eye on you. There is more to this raid than meets the eye, I think."
         [/message]
         [message]
             speaker=Prince Haldric
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/06_Temple_in_the_Deep.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/06_Temple_in_the_Deep.cfg
@@ -205,7 +205,7 @@
             [then]
                 [message]
                     speaker=narrator
-                    message= _ "You already have the Fire Ruby."
+                    message= _ "Haldric already has the Ruby of Fire."
                     image=wesnoth-icon.png
                 [/message]
 
@@ -229,7 +229,7 @@
 
                         [message]
                             speaker=narrator
-                            message= _ "As you open the chest you see it, the Ruby of Fire. It is the size of an apple, and burns with an internal fire, which is refracted through its faces. You can feel the power flowing from it..."
+                            message= _ "As Haldric opens the chest he sees it, the Ruby of Fire. It is the size of an apple, and burns with an internal fire, which is refracted through its faces. He can feel the power flowing from it..."
                             image=wesnoth-icon.png
                         [/message]
 
@@ -289,7 +289,7 @@
         name=victory
         [message]
             speaker=Prince Haldric
-            message= _ "I’m glad that’s over! We have the Ruby of Fire, and that Lich-Lord is now a pile of dust, let’s get out of these catacombs!"
+            message= _ "I’m glad that’s over! We have the Ruby of Fire, and that Lich-Lord is now a pile of dust. Let’s get out of these catacombs!"
         [/message]
         {CLEAR_VARIABLE Have_Ruby}
     [/event]
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/07_Return_to_Oldwood.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/07_Return_to_Oldwood.cfg
@@ -117,7 +117,7 @@
         [/message]
         [message]
             speaker=Prince Haldric
-            message= _ "Yes, it was a tough battle, but we prevailed. Now, I’m having a problem. I know not what I should do next."
+            message= _ "Yes, it was a tough battle, but we prevailed. Now, I have a problem. I know not what I should do next."
         [/message]
         [message]
             speaker=Prince Haldric
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/08_Clearwater_Port.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/08_Clearwater_Port.cfg
@@ -918,7 +918,7 @@
         name=victory
         [message]
             speaker=Prince Haldric
-            message= _ "We’ve escaped from the orcs before we were trapped by the ice! Now, on to Southbay."
+            message= _ "We’ve escaped from the orcs before getting trapped by the ice! Now, on to Southbay."
         [/message]
 
         [message]
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/09_Fallen_Lich_Point.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/09_Fallen_Lich_Point.cfg
@@ -244,7 +244,7 @@
 
         [message]
             speaker=narrator
-            message= _ "Prince Haldric has arrived at Fallen Lich Point, to retrieve the Lich-Lord Caror’s Book of Fire and Darkness, and flee into the sewers of Southbay."
+            message= _ "Prince Haldric has arrived at Fallen Lich Point, to retrieve Lich-Lord Caror’s Book of Fire and Darkness, and flee into the sewers of Southbay."
             image=wesnoth-icon.png
         [/message]
 
@@ -344,7 +344,7 @@
         [/filter]
         [message]
             speaker=narrator
-            message= _ "This monolith was erected by me, ― (<i>chipped away</i>), first Mage of the good people of the Green Isle. By its power the Lich-Lord is bound in stone. To end the spell a noble of the line of Kings should utter the following..."
+            message= _ "This monolith was erected by me, ― (<i>chipped away</i>), first Mage of the good people of the Green Isle. By its power, the Lich-Lord is bound in stone. To end the spell, a noble of the line of Kings should utter the following..."
             image=wesnoth-icon.png
         [/message]
 
@@ -388,7 +388,7 @@
 
                                     [message]
                                         speaker=Prince Haldric
-                                        message= _ "The lich is free! Let’s bash him and grab that book. That sounds like a job for you, Lady Jessene!"
+                                        message= _ "The lich is free! Let’s bash him and grab that book. This sounds like a job for you, Lady Jessene!"
                                     [/message]
                                     [message]
                                         speaker=Lady Jessene
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/11_Southbay_in_Winter.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/11_Southbay_in_Winter.cfg
@@ -182,7 +182,7 @@
         [/message]
         [message]
             speaker=King Addroran IX
-            message= _ "Hmm. There is wisdom in your words, and my end should be in this place. We can hold out for the rest of the winter here. When the orcs last came it was fall and the harvest was in. We can hunt the great schools of fish that live under the ice. You might even be able to depart in well provisioned ships!"
+            message= _ "Hmm. There is wisdom in your words, and my end should be in this place. We can hold out for the rest of the winter here. When the orcs last came it was fall and the harvest was in. We can hunt the great schools of fish that live under the ice. You might even be able to depart in well-provisioned ships!"
         [/message]
         [message]
             speaker=Lady Jessene
@@ -190,7 +190,7 @@
         [/message]
         [message]
             speaker=King Addroran IX
-            message= _ "No. Flee, and flee east. It is my duty to hold this city as long as I can for the sake of all who can be evacuated. That means that it shall fall on Haldric and you to lead the evacuation."
+            message= _ "No. Flee, and flee east. It is my duty to hold this city as long as I can for the sake of all who can be evacuated. This means that it shall fall on Haldric and you to lead the evacuation."
         [/message]
         [message]
             speaker=Prince Haldric
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/16_The_Kalian.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/16_The_Kalian.cfg
@@ -227,7 +227,7 @@
 
 #define ISLE
     [option]
-        label= _ "Let’s put those souls to rest on the cursed isle!"
+        label= _ "Let’s put those souls on the cursed isle to rest!"
 
         [show_if]
             [variable]
@@ -313,7 +313,7 @@
         [then]
             [message]
                 speaker="Lord El'Isomithir"
-                message= _ "Long ago a clan of elves on the Isle of Tears fell under the sway of a dark curse. Their souls still haunt that place and no elf will go there. You should clear this isle, and put their souls to rest."
+                message= _ "Long ago, a clan of elves on the Isle of Tears fell under the sway of a dark curse. Their souls still haunt that place and no elf will go there. You should clear this isle, and put their souls to rest."
             [/message]
         [/then]
     [/if]
@@ -341,7 +341,7 @@
             value="Isle"
             [message]
                 speaker="Lord El'Isomithir"
-                message= _ "Finally the souls of our poor kin may rest. Thank you."
+                message= _ "Finally, the souls of our poor kin may rest. Thank you."
             [/message]
         [/case]
         [case]
@@ -433,7 +433,7 @@
                 [/message]
                 [message]
                     speaker=Lord Aryad
-                    message= _ "Well then, human. Each of us lords has a specific quest for you. If you complete them all you will be granted all of the plains in our domain and the hills south of the Great River, if not you will be forced to depart. So which quest do you wish to undertake first?"
+                    message= _ "Well then, human. Each of us lords has a specific quest for you. If you complete them all you will be granted all of the plains in our domain and the hills south of the Great River, if not you will be forced to depart. So, which quest do you wish to undertake first?"
                 [/message]
                 {QUESTS}
 
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/21_The_Plan.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/21_The_Plan.cfg
@@ -109,11 +109,11 @@
 
         [message]
             speaker=Lady Jessene
-            message= _ "So what is your plan?"
+            message= _ "So, what is your plan?"
         [/message]
         [message]
             speaker=Prince Haldric
-            message= _ "We’re going to convince Jevyan that we gave the elves the Ruby of Fire to secure our place in this new land. Then if the orcs return, hopefully they’ll go looking for our not-so-loyal elven allies first."
+            message= _ "We’re going to convince Jevyan that we gave the elves the Ruby of Fire to secure our place in this new land. Then, if the orcs return, hopefully they’ll go looking for our not-so-loyal elven allies first."
         [/message]
         [message]
             speaker=Lady Jessene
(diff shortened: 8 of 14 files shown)


data/campaigns/The_Rise_Of_Wesnoth/scenarios/07_Return_to_Oldwood.cfg
```diff
@@ -117,7 +117,7 @@
         [/message]
         [message]
             speaker=Prince Haldric
-            message= _ "Yes, it was a tough battle, but we prevailed. Now, I’m having a problem. I know not what I should do next."
+            message= _ "Yes, it was a tough battle, but we prevailed. Now, I have a problem. I know not what I should do next."
```
knyghtmare, line 120: > I know not what I should do next.

Wouldn't "I do not know what I should do next" be better?

data/campaigns/The_Rise_Of_Wesnoth/units/Wesfolk_Outcast.cfg
```diff
@@ -18,7 +18,7 @@
     advances_to=Wesfolk Lady
     cost=18
     usage=mixed fighter
-    description= _ "Born in the Wesfolk aristocracy, she and her people were excluded from society by their lords, who betrayed their loyalty when a war against Haldric’s people was being lost. This outcast still conserves her nobleness in her veins and on the battlefield she has earned valuable experience, which she can lend to her people for turning a fight in their favor."
+    description= _ "Born in the Wesfolk aristocracy, she and her people were excluded from society by their lords, who betrayed their loyalty when a war against Haldric’s people was being lost. This outcast still conserves her nobleness in her veins, and on the battlefield she has earned valuable experience, which she can lend to her people for turning a fight in their favor."
```
knyghtmare, line 21: Was the single comma that necessary?

data/campaigns/The_Rise_Of_Wesnoth/scenarios/07_Return_to_Oldwood.cfg
max-torch: Not necessarily.
Have you heard the phrase "You know not what you are asking?"

data/campaigns/The_Rise_Of_Wesnoth/units/Wesfolk_Outcast.cfg
max-torch: image
independent clause.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/09_Fallen_Lich_Point.cfg
```diff
@@ -344,7 +344,7 @@
         [/filter]
         [message]
             speaker=narrator
-            message= _ "This monolith was erected by me, ― (<i>chipped away</i>), first Mage of the good people of the Green Isle. By its power the Lich-Lord is bound in stone. To end the spell a noble of the line of Kings should utter the following..."
+            message= _ "This monolith was erected by me, ― (<i>chipped away</i>), first Mage of the good people of the Green Isle. By its power, the Lich-Lord is bound in stone. To end the spell, a noble of the line of Kings should utter the following..."
```
Wedge009, line 347: I must have missed this in my recent review. I wonder if 'line of kings' should be lower-case.

Wedge009:
Overall seems okay to me. I just wonder about the change from second-person to third-person when Haldric finds the Ruby. Perhaps it should be phrased differently for third-person.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/09_Fallen_Lich_Point.cfg
max-torch: One could argue yes. There is a rule that says honorifics such as judge, king, etc, should be lower case if it does not follow a name. He is King Eldaric. A king. Meet Prince Haldric. A prince must fight. Want me to change king to lower case?

Wedge009: If you could, yes.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/07_Return_to_Oldwood.cfg
CelticMinstrel: > Wouldn't "I do not know what I should do next" be better?

They are both correct grammar and mean the same thing, but the existing phrasing is archaic.

data/campaigns/The_Rise_Of_Wesnoth/units/Wesfolk_Outcast.cfg
CelticMinstrel: I agree with this comma. Without it, it lends to parsing it like this, which then leads you into a grammatical dead end (indicated by the asterisk):

> This outcast still conserves her nobleness [in her veins] and [on the battlefield] *she has earned valuable experience, which she can lend to her people for turning a fight in their favor.

There's one other way I can see to solve this:

> This outcast still conserves her nobleness in her veins and on the battlefield; she has earned valuable experience, which she can lend to her people for turning a fight in their favor.

However, I think the meaning implied by the added comma is probably closer to the intention than the different meaning conveyed by adding that semicolon.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/09_Fallen_Lich_Point.cfg
CelticMinstrel: Wouldn't that also apply to "mage" in this line?

data/campaigns/The_Rise_Of_Wesnoth/scenarios/01_A_Summer_of_Storms.cfg
```diff
@@ -157,7 +157,7 @@
         [/message]
         [message]
             speaker=King Eldaric IV
-            message= _ "You’re showing initiative, son! I’m proud of you! Yes, you may lead our forces to battle, it is time — but I’ll stay near to keep an eye on you. There is more to this raid than meets the eye, I think."
+            message= _ "You’re showing initiative, son! I’m proud of you! Yes, you may lead our forces in battle, it is time — but I’ll stay near to keep an eye on you. There is more to this raid than meets the eye, I think."
```
CelticMinstrel, line 160: I question this replacement; it seems to change the meaning and possibly not in a good way.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/06_Temple_in_the_Deep.cfg
```diff
@@ -229,7 +229,7 @@
 
                         [message]
                             speaker=narrator
-                            message= _ "As you open the chest you see it, the Ruby of Fire. It is the size of an apple, and burns with an internal fire, which is refracted through its faces. You can feel the power flowing from it..."
+                            message= _ "As Haldric opens the chest he sees it, the Ruby of Fire. It is the size of an apple, and burns with an internal fire, which is refracted through its faces. He can feel the power flowing from it..."
```
CelticMinstrel, line 232: I think I have a few punctuation complaints here too.

* Shouldn't there be a comma after "chest"?
* The comma before "Ruby of Fire" doesn't seem right. I think it should be a dash or maybe even a colon. (And I don't suggest colons lightly; they're usually the wrong choice.)
* I'm not sure about the comma after "apple". It seems to me that it could be omitted, maybe?

data/campaigns/The_Rise_Of_Wesnoth/scenarios/08_Clearwater_Port.cfg
```diff
@@ -918,7 +918,7 @@
         name=victory
         [message]
             speaker=Prince Haldric
-            message= _ "We’ve escaped from the orcs before we were trapped by the ice! Now, on to Southbay."
+            message= _ "We’ve escaped from the orcs before getting trapped by the ice! Now, on to Southbay."
```
CelticMinstrel, line 921: I don't think I like this choice of word. It could be "being" or "becoming" instead, which might be a little better, but I wonder if there's something even better than those…

data/campaigns/The_Rise_Of_Wesnoth/scenarios/11_Southbay_in_Winter.cfg
```diff
@@ -182,15 +182,15 @@
...
+            message= _ "Hmm. There is wisdom in your words, and my end should be in this place. We can hold out for the rest of the winter here. When the orcs last came it was fall and the harvest was in. We can hunt the great schools of fish that live under the ice. You might even be able to depart in well-provisioned ships!"
         [/message]
         [message]
             speaker=Lady Jessene
             message= _ "Don’t you intend to come with us?"
         [/message]
         [message]
             speaker=King Addroran IX
-            message= _ "No. Flee, and flee east. It is my duty to hold this city as long as I can for the sake of all who can be evacuated. That means that it shall fall on Haldric and you to lead the evacuation."
+            message= _ "No. Flee, and flee east. It is my duty to hold this city as long as I can for the sake of all who can be evacuated. This means that it shall fall on Haldric and you to lead the evacuation."
```
CelticMinstrel, line 193: I think "that" is more correct in this line.

If you made the change because "that" appears twice in the same sentence, you could just remove the second instance, though personally I don't think that's an issue.

Unrelated side note: maybe also "upon Haldric"?

data/campaigns/The_Rise_Of_Wesnoth/scenarios/16_The_Kalian.cfg
```diff
@@ -433,7 +433,7 @@
                 [/message]
                 [message]
                     speaker=Lord Aryad
-                    message= _ "Well then, human. Each of us lords has a specific quest for you. If you complete them all you will be granted all of the plains in our domain and the hills south of the Great River, if not you will be forced to depart. So which quest do you wish to undertake first?"
+                    message= _ "Well then, human. Each of us lords has a specific quest for you. If you complete them all you will be granted all of the plains in our domain and the hills south of the Great River, if not you will be forced to depart. So, which quest do you wish to undertake first?"
```
CelticMinstrel, line 436: Additionally, I think "Great River" should be followed by a semicolon.

data/campaigns/The_Rise_Of_Wesnoth/units/Wesfolk_Leader.cfg
```diff
@@ -19,7 +19,7 @@
     {AMLA_DEFAULT}
     cost=60
     usage=mixed fighter
-    description= _ "Born in the Wesfolk aristocracy, she and her people were excluded from society by their lords, who betrayed their loyalty when a war against Haldric’s people was being lost. This outcast still conserves her nobleness in her veins and on the battlefield she has earned valuable experience, which continues increasing with time, as does her natural leadership."
+    description= _ "Born in the Wesfolk aristocracy, she and her people were excluded from society by their lords, who betrayed their loyalty when a war against Haldric’s people was being lost. This outcast still conserves her nobleness in her veins, and on the battlefield she has earned valuable experience, which continues to increase with time, as does her natural leadership."
```
CelticMinstrel, line 22: This looks good, but I notice an awkward passive in here too - "was being lost". Is there anything we can do about that? It's not the end of the world, but it sure is awkward.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/09_Fallen_Lich_Point.cfg
Wedge009: I didn't even see that... It seems to be a title bestowed upon the inscriber, in which case I might argue for it to be 'First Mage'.

CelticMinstrel: Yeah, I'd agree that either both words should be capitlized or neither.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/01_A_Summer_of_Storms.cfg
max-torch: Rationale is it matches the line that comes immediately before it.
https://github.com/wesnoth/wesnoth/blob/dd3baa488150bfc8d8cd0105da9083009cb29ca0/data/campaigns/The_Rise_Of_Wesnoth/scenarios/01_A_Summer_of_Storms.cfg#L154-L161
            message= _ "Father, I am of age now. May I lead our forces in battle?"

data/campaigns/The_Rise_Of_Wesnoth/scenarios/06_Temple_in_the_Deep.cfg
max-torch: >Shouldn't there be a comma after "chest"?

Sounds good to me since the introductory clause is long.

> The comma before "Ruby of Fire" doesn't seem right. I think it should be a dash or maybe even a colon. (And I don't suggest colons lightly; they're usually the wrong choice.)

I'll go with an em dash for this one.

> I'm not sure about the comma after "apple". It seems to me that it could be omitted, maybe?
We can omit. The clause after apple is a dependent clause after all.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/08_Clearwater_Port.cfg
max-torch: "ending up"?

Wedge009: No, I'd vote for being/becoming in this case.

max-torch: "winding up"?

max-torch: Settled on "We’ve escaped from the orcs before the winter ice could trap us!"

data/campaigns/The_Rise_Of_Wesnoth/scenarios/11_Southbay_in_Winter.cfg
max-torch: I followed the suggestion here (celmin). A google search shows me that upon and on are practically interchangeable and don't make much difference in meaning. Keeping the original text.

data/campaigns/The_Rise_Of_Wesnoth/units/Wesfolk_Leader.cfg
max-torch: Settled on "when they started losing a war against Haldric's people"

data/campaigns/The_Rise_Of_Wesnoth/scenarios/08_Clearwater_Port.cfg
Wedge009: Much better than the 'up' suggestions.

data/campaigns/The_Rise_Of_Wesnoth/units/Wesfolk_Leader.cfg
```diff
@@ -19,7 +19,7 @@
     {AMLA_DEFAULT}
     cost=60
     usage=mixed fighter
-    description= _ "Born in the Wesfolk aristocracy, she and her people were excluded from society by their lords, who betrayed their loyalty when a war against Haldric’s people was being lost. This outcast still conserves her nobleness in her veins and on the battlefield she has earned valuable experience, which continues increasing with time, as does her natural leadership."
+    description= _ "Born in the Wesfolk aristocracy, she and her people were excluded from society by their lords, who betrayed their loyalty when they started losing a war against Haldric's people. This outcast still conserves her nobleness in her veins, and on the battlefield she has earned valuable experience, which continues to increase with time, as does her natural leadership."
```
Wedge009, line 22: I believe this is fine, but 'Haldric's people' needs to use a 'curly' apostrophe.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/01_A_Summer_of_Storms.cfg
CelticMinstrel: Yeah, I guess that does make sense. I might've changed Haldric's preposition instead, though. But I'm not entirely sure which is better…

data/campaigns/The_Rise_Of_Wesnoth/scenarios/11_Southbay_in_Winter.cfg
CelticMinstrel: Yeah, "upon" doesn't change the meaning, just gives a different feel to it. More formal maybe? But it's not really an issue.

data/campaigns/The_Rise_Of_Wesnoth/units/Wesfolk_Leader.cfg
CelticMinstrel: I think I might prefer an infinitive rather than a gerund here as well - "when they started to lose".

data/campaigns/The_Rise_Of_Wesnoth/scenarios/01_A_Summer_of_Storms.cfg
max-torch: I propose making both lines "lead our forces into battle". This good?

data/campaigns/The_Rise_Of_Wesnoth/scenarios/21_The_Plan.cfg
```diff
@@ -109,11 +109,11 @@
 
         [message]
             speaker=Lady Jessene
-            message= _ "So what is your plan?"
+            message= _ "So, what is your plan?"
```
nemaara, line 112: I think this one is better without a comma, the others are fine though.

nemaara (approved):

nemaara (approved):

nemaara:
Merging in a day or two unless there are more comments.

nemaara:
Well I swear I hit squash before merging but I guess it didn't work...

data/campaigns/The_Rise_Of_Wesnoth/scenarios/09_Fallen_Lich_Point.cfg
```diff
@@ -344,7 +344,7 @@
         [/filter]
         [message]
             speaker=narrator
-            message= _ "This monolith was erected by me, ― (<i>chipped away</i>), first Mage of the good people of the Green Isle. By its power the Lich-Lord is bound in stone. To end the spell a noble of the line of Kings should utter the following..."
+            message= _ "This monolith was erected by me, ― (<i>chipped away</i>), First Mage of the good people of the Green Isle. By its power, the Lich-Lord is bound in stone. To end the spell, a noble of the line of kings should utter the following..."
```
stevecotton, line 347: I'm translating this, and I want to check my understanding of it, because I think the meaning has subtly changed:
* "First Mage" would be either the highest ranking mage in the army, or the one appointed to the Royal Court.
* "first Mage" would specifically be the first (chronologically) to occupy that role.

So the "first Mage" would also be the "First Mage", but their successor would be "second First Mage".

data/campaigns/The_Rise_Of_Wesnoth/units/Wesfolk_Leader.cfg
```diff
@@ -19,7 +19,7 @@
     {AMLA_DEFAULT}
     cost=60
     usage=mixed fighter
-    description= _ "Born in the Wesfolk aristocracy, she and her people were excluded from society by their lords, who betrayed their loyalty when a war against Haldric’s people was being lost. This outcast still conserves her nobleness in her veins and on the battlefield she has earned valuable experience, which continues increasing with time, as does her natural leadership."
+    description= _ "Born in the Wesfolk aristocracy, she and her people were excluded from society by their lords, who betrayed their loyalty when they started to lose a war against Haldric’s people. This outcast still conserves her nobleness in her veins, and on the battlefield she has earned valuable experience, which continues to increase with time, as does her natural leadership."
```
stevecotton, line 22: Was the intention to make this change to the L2 Wesfolk_Lady too? The L2 and L3 previously had exactly the same text.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/09_Fallen_Lich_Point.cfg
max-torch: I think the lore and also TRoW is slated to have an upcoming rework? Meaning you can probably choose your interpretation since it is yet to be finalized.

data/campaigns/The_Rise_Of_Wesnoth/units/Wesfolk_Leader.cfg
max-torch: Only the first change, "when they started to lose a war" has the intention to also make the change to also the Wesfolk Lady and Wesfolk Outcast. (A mistake that they were not all changed).
Regarding experience increasing with time, that is meant to be different.

data/campaigns/The_Rise_Of_Wesnoth/scenarios/09_Fallen_Lich_Point.cfg
Wedge009: I think I suggested this capitalisation. It's difficult to know what the intent of the original text was. My objection was the capitalisation of Mage on its own, to me it didn't make sense to do so. It would only make sense if First Mage was a title of some sort otherwise it should all be lower-case as in 'first mage of the good people...'. Whether the First was meant as the chronological first mage of the Green Isle, or First as in a position of primacy (eg like 'First Lady' in the US), I don't really know.

As max mentioned, it's subject to being rewritten anyway. Personally I read it as the chronological First Mage, as in, whoever inscribed this was the first formally-recognised magic wielder and there isn't necessarily an implication that it's a position of authority that gets handed down to successors.

Issue #6140, opened 2021-10-02
TRoW S8 Clearwater Port: Smoothen Naga Reveal


max-torch:
### Game and System Information
 - 1.15.18
 - Steam
 - Windows 10

### Describe the desired feature
The issue is that once the naga pop up on the island at the bottom, immediately the map scrolls away from them, before you can even make out exactly what happened. The desired feature is to either make the map linger on the nagas longer when they pop up, or make the nagas come in from the side of the map and stop on their designated hexes.

### Screenshots
See how hard it is to see that the naga have shown up.
![trownaga(4)](https://user-images.githubusercontent.com/41433185/135706755-95257691-cecf-4c9e-acba-f88648936193.gif)

The gif preview is somewhat compressed, which was necessary to meet exactly the 10MB file limit. You can zoom in to more clearly see the issue. You can also check out the replay:
[TRoW-Clearwater Port replay 20210915-213016.gz](https://github.com/wesnoth/wesnoth/files/7271002/TRoW-Clearwater.Port.replay.20210915-213016.gz)

Wedge009:
I don't know much about WML, but I wonder if there's a way to pause/wait for a few seconds before initiating the scroll away.

Wedge009:
What's the event that prompts the second movement? I can only see the `scroll_to` for the naga introduction:
https://github.com/wesnoth/wesnoth/blob/dd3baa488150bfc8d8cd0105da9083009cb29ca0/data/campaigns/The_Rise_Of_Wesnoth/scenarios/08_Clearwater_Port.cfg#L392-L394

max-torch:
I believe that event is merely the game focusing its attention on the next unit that moves. Like when the game scrolls to a speaker of dialog, a map scroll that does not need to be explicitly defined in WML code.

Wedge009:
Maybe giving one of the naga an introductory taunt might serve to pause the view.

knyghtmare:
> What's the event that prompts the second movement? I can only see the `scroll_to` for the naga introduction:
> 
> https://github.com/wesnoth/wesnoth/blob/dd3baa488150bfc8d8cd0105da9083009cb29ca0/data/campaigns/The_Rise_Of_Wesnoth/scenarios/08_Clearwater_Port.cfg#L392-L394

Would adding a time delay here resolve this?

Wedge009:
As discussed in the PR, this issue only occurs if you have units scheduled to continue moving after the naga appears. It's not a problem if this is not the case.

PR #6144, opened 2021-10-02
TRoW S08 - Add a way of smoothing the appearance of Naga in the map in case the player premoves a unit.

Descacharrado:
Delay time can be changed. One second felt okey imo tho (sorry for the quality of the attached file).
Fix for issue #6140

https://user-images.githubusercontent.com/30196839/135730846-952fe630-27b9-4bfb-828e-78e3b04150db.mp4

### DIFF
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/08_Clearwater_Port.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/08_Clearwater_Port.cfg
@@ -411,6 +411,9 @@
         {LOYAL_UNIT 3 (Naga Fighter) 30 31}
         {LOYAL_UNIT 3 (Naga Fighter) 27 33}
 #endif
+        [delay]
+            time=1000
+        [/delay]
     [/event]
 
     [event]


Wedge009:
This works for me - I just wasn't sure of the precise WML to do this. The other option I thought of was to add a line for the naga but that would involve some writing and approvals, plus complications with string freezes, etc.

BTW, if you specify the GitHub issue in your commit message as per https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue it will close the issue automatically upon merge.

@max-torch Are you satisfied with 1 second? I think it should be sufficient.

max-torch:
Would it be a good idea to coauthor descacharrado? He came up with the exact same solution at the exact same time and it's on record at #6140.
image

max-torch:
Just want to clarify that it happens regardless of any "premoving". It happens all the time.

Descacharrado:
> Just want to clarify that it happens regardless of any "premoving". It happens all the time.

That is weird because I could not replicate the issue without premoving a unit. If there was nothing to move the camera after the naga spawn it would just lock on the naga for me, which is the expected result.

Descacharrado:
> BTW, if you specify the GitHub issue in your commit message as per https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue it will close the issue automatically upon merge.

Thank you, will take into account in future PRs

Wedge009:
Confirmed: the underlying issue is only a problem if you have units scheduled to continue moving before the event triggers. I confirm this in both 1.14.17 and master.

max-torch:
> Confirmed: the underlying issue is only a problem if you have units scheduled to continue moving before the event triggers. I confirm this in both 1.14.17 and master.

Okay I stand corrected. Well then, at the very least, it's a case in all replays - so it will make replays look nicer.

nemaara:
If I merge this do I need to port it to 1.16 too?

PR #6161, opened 2021-10-05
Install certs so CI can download scons over https.

Pentarctagon:


Wedge009:
Oh, this resolved the flatpak check failing?

Pentarctagon:
Yep.  Why it suddenly started happening at all will probably forever remain a mystery, but this fixes it.

PR #6164, opened 2021-10-06
Refactor handling of allies in Sceptre of Fire

Wedge009:
With four potential allies in the campaign, there is a lot of dialogue duplication. This refactoring attempts to consolidate the text. Additionally, the ally ID was recorded in the variable `ally_name` which can be confusing (it was to me, at least), so I renamed it to `ally_id` instead.

Unless I made a mistake somewhere, the dialogue itself should not be changed, this is just to avoid redundant duplication. Please keep in mind I'm still very much a novice with WML so be gentle in pointing out any errors. Thanks.

PS I tested a little bit but haven't done a thorough checking of S7b (where `ally_id` is introduced) onwards, for all four allies.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/07a_Dealing_with_Dwarves.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/07a_Dealing_with_Dwarves.cfg
@@ -144,14 +144,14 @@
 
             [then]
                 [set_variable]
-                    name=ally_name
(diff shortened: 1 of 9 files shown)


Wedge009:
WTH, completely wrong campaign...

CelticMinstrel:
You can't just rename the PR?

Wedge009:
I needed to update the commit message too.

Issue #6158, opened 2021-10-05
Refactor ally dialogue in UtBS


Wedge009:
#6155 highlights that the way the dialogue of a potential ally is handled is not good. Basically it appears in many cases strings are printed in quadruplicate to accommodate the four different potential ally characters. Suggest refactoring this not just to avoid the redundant strings and corresponding translation effort but also to avoid complications with female characters as @CelticMinstrel mentioned.

MultiSeptim:
As a translator (czech), I don't mind multiple similar text strings (because they are easy to translate), but the inappropriate use of variable names, which are difficult to incorporate into translation. And Czech still uses Latin, I can't imagine what the troubles will be in other scriptures.

Wedge009:
Re-opening due to reversion of #6165 via d5b95e621d5a11e6fecbd301117e5ca639164825. I suspect it's unlikely that avoiding text redundancy and avoiding variables for translation efforts can be resolved together.

I want to highlight that the refactoring was _not_ the only place where variables are used in translated text. So I consider that the question of what to do in those circumstances may also be relevant. A couple of examples:
https://github.com/wesnoth/wesnoth/blob/d5b95e621d5a11e6fecbd301117e5ca639164825/data/campaigns/Dead_Water/scenarios/10_The_Flaming_Sword.cfg#L272-L275

https://github.com/wesnoth/wesnoth/blob/d5b95e621d5a11e6fecbd301117e5ca639164825/data/campaigns/Heir_To_The_Throne/scenarios/20b_Underground_Channels.cfg#L612-L613

For more background, see the discussion in https://github.com/hrubymar10/wesnoth-cs/pull/209.

CelticMinstrel:
Reopened due to a discussion in hrubymar10/wesnoth-cs#209.

stevecotton:
The revert or revert+rework will need backporting. Sorry, but I think d5b95e6 reverted too much stuff, as it included reverting the variable-name change from `ally_name` to `ally_id`, and people are going to be mid-playthrough on 1.16.x with a save file that now includes the new variable name.

To backport, it needs to be using `ally_id`, but also have separate translatable strings for each ally-specific message.

Wedge009:
> I think [d5b95e6](https://github.com/wesnoth/wesnoth/commit/d5b95e621d5a11e6fecbd301117e5ca639164825) reverted too much stuff, as it included reverting the variable-name change from `ally_name` to `ally_id`.

[...]

Issue #6155, opened 2021-10-05
UtBS S09: Untranslated `$ally_name`, which way should we go?


stevecotton:
A bug introduced by dac45d3eb52fba72898a6604941e7e6b20d1cbe1, the commit that's just entered master.

In this line, `$ally_name` contains a unit-id rather than a translated name. It is always one of the hero units "Grog", "Nog", "Rogrimir" or "Jarl", and in many languages the unit-id and the name are the same; however they're different in several languages including `bg`, `el`, `zh_CN` and `zh_TW`.

Options are to use four strings and the `MESSAGE_DEPEND_ON_ALLY` macro, or to just omit the name from the message.

stevecotton:
Clarification: the issue is that Grog's called "Grog", "Грог", "果洛格", etc depending on the language, but `$ally_name` will contain "Grog" irrespective of language.

Wedge009:
Is another option to split the string so that the `$ally_name` is not part of the translation? I'd be very surprised if this is the only place where variables are used in dialogue.

Hmm, I just had a look at `MESSAGE_DEPEND_ON_ALLY` in UtBS, it does look like the original author/s had to rely on quadruplicate strings to accommodate the different potential allies... Ugh, messy, but should probably do the same thing to accommodate this string too.

stevecotton:
In 13 years one player has noticed the capitalisation bug in "which way should we go?", I think we should just remove the name from that text.

max-torch:
> In 13 years one player has noticed the capitalisation bug in "which way should we go?", I think we should just remove the name from that text.

Or, we could say "only one player has bothered to report". :)

CelticMinstrel:
I think the entire way the ally works is suboptimal. Instead of having an `$ally_name` variable, it should set a role. That's beside the point, though. If you want to put the ally's actual name in the text, all you need is a `[store_unit]` call above the message to put the unit in a variable. Then use `$the_variable.name`.

Wedge009:
It does indeed appear the way the ally dialogue is implemented isn't great, but that's another issue.

Wedge009:
> If you want to put the ally's actual name in the text, all you need is a `[store_unit]` call above the message to put the unit in a variable. Then use `$the_variable.name`.

@CelticMinstrel That sounds like it will resolve the quadruplication, but the issue with the name translation still remains doesn't it? Or do you mean to keep the `$the_variable.name` outside of the translated string?

[...]

PR #6165, opened 2021-10-06
Refactor handling of allies in UtBS

Wedge009:
With four potential allies in the campaign, there is a lot of dialogue duplication. This refactoring attempts to consolidate the text. Additionally, the ally ID was recorded in the variable `ally_name` which can be confusing (it was to me, at least), so I renamed it to `ally_id` instead.

Unless I made a mistake somewhere, the dialogue itself should not be changed, this is just to avoid redundant duplication. Please keep in mind I'm still very much a novice with WML so be gentle in pointing out any errors. Thanks.

PS I tested a little bit but haven't done a thorough checking of S7 (where `ally_id` is introduced) onwards, for all four allies.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/07b_Talking_with_Trolls.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/07b_Talking_with_Trolls.cfg
@@ -142,14 +142,14 @@
 
             [then]
                 [set_variable]
-                    name=ally_name
+                    name=ally_id
                     value=Grog
                 [/set_variable]
             [/then]
 
             [else]
                 [set_variable]
-                    name=ally_name
+                    name=ally_id
                     value=Nog
                 [/set_variable]
 
@@ -356,26 +356,11 @@
             message= _ "Darmog has never been above ground, but Darmog understand your story. A leader must protect and care for his people. Every people deserve to find their own home. If we can help you we will."
         [/message]
 
-        [if]
-            [variable]
-                name=ally_name
-                equals=Grog
-            [/variable]
-
-            [then]
-                [message]
-                    speaker=Spiritual Advisor
-                    message= _ "We may be able to help you find a way back to the sunlit lands. In our temples we do keep records of the past. We have not walked above the earth for many many generations, not since the darkness drove us underground. But we are masters of the underground lands, and we have explored many tunnels. Recently one of our scouts found a path that leads north back to the sunlit lands, I think it may be the way you are trying to go. In reward for your achievements, we will help you. Grog has volunteered to protect you and lead you back to the sunlight lands."
-                [/message]
-            [/then]
-
-            [else]
-                [message]
-                    speaker=Spiritual Advisor
-                    message= _ "We may be able to help you find a way back to the sunlit lands. In our temples we do keep records of the past. We have not walked above the earth for many many generations, not since the darkness drove us underground. But we are masters of the underground lands, and we have explored many tunnels. Recently one of our scouts found a path that leads north back to the sunlit lands, I think it may be the way you are trying to go. In reward for your achievements, we will help you. Nog has volunteered to protect you and lead you back to the sunlight lands."
-                [/message]
-            [/else]
-        [/if]
+        {SET_ALLY_UNIT}
+        [message]
+            speaker=Spiritual Advisor
+            message= _ "We may be able to help you find a way back to the sunlit lands. In our temples we do keep records of the past. We have not walked above the earth for many many generations, not since the darkness drove us underground. But we are masters of the underground lands, and we have explored many tunnels. Recently one of our scouts found a path that leads north back to the sunlit lands, I think it may be the way you are trying to go. In reward for your achievements, we will help you. $ally_unit.name has volunteered to protect you and lead you back to the sunlight lands."
+        [/message]
 
         [if]
             [variable]
@@ -404,26 +389,11 @@
         [/if]
         {CLEAR_VARIABLE saved_grog}
 
-        [if]
-            [variable]
-                name=ally_name
-                equals=Grog
-            [/variable]
-
-            [then]
-                [message]
-                    speaker=Kaleh
-                    message= _ "Thank you very much for your help. We were worried about getting lost in all these twisting tunnels. And we would be honored to have you come with us, Grog."
-                [/message]
-            [/then]
-
-            [else]
-                [message]
-                    speaker=Kaleh
-                    message= _ "Thank you very much for your help. We were worried about getting lost in all these twisting tunnels. And we would be honored to have you come with us, Nog."
-                [/message]
-            [/else]
-        [/if]
+        [message]
+            speaker=Kaleh
+            message= _ "Thank you very much for your help. We were worried about getting lost in all these twisting tunnels. And we would be honored to have you come with us, $ally_unit.name."
+        [/message]
+        {CLEAR_VARIABLE ally_unit}
 
         [if]
             [have_unit]
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg
@@ -289,7 +289,7 @@
 
         # recall dwarf/troll
         [recall]
-            id=$ally_name
+            id=$ally_id
         [/recall]
         # wmllint: recognize Grog
         # wmllint: recognize Nog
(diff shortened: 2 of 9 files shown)


CelticMinstrel:
That's… way larger than I expected. It's pretty much the kind of thing I was thinking of, though. I'm just not 100% sure that this is truly a viable option… it should be fine in all languages I know anything about, but I can't be certain that there exists no language in which it would be a problem.

For example: theoretically, there could be a language where "of <name>" is translated with <name> declined in a genitive case, which may not be as simple as a straight suffix or prefix. I don't know if this hypothetical language exists, but it wouldn't surprise me if it does.

Wedge009:
In for a penny, in for a pound - once I start something I find it hard to stop...

I know you pointed out potential issues with `female_message` in #6155 but in this particular case all characters are male. As for other potential issues - wouldn't it be a problem even without this change? Or are you thinking that explicitly typing out all four cases would avoid those potential issues? I admit I'm not a super-grammatical expert so aspects that are foreign to English would be lost on me.

Anyway, given the unknowns with other languages, maybe it's better to just leave this change in master and not back-port to 1.16.

stevecotton:
While it's a good change, I think it shouldn't be backported. As a general rule (including for this change), I think any string changes that shouldn't be backported should also not be added to master until 1.16.1 has released, so that they can't conflict with any string changes that do get backported.

Conclusion: I think this PR should be postponed.

data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg
```diff
@@ -1842,31 +1838,27 @@
...
+                [message]
+                    speaker=$ally_id
+                    message= _ "The chieftain told me to serve you, and it will take more than this to shatter my confidence in you, lad. But those elves must be out on a different island, how will we cross the deep water?"
+                [/message]
+            [/then]
+
+            [else]
+                [message]
+                    speaker=$ally_id
+                    message= _ "Great Leader told $ally_unit.name to serve you, and so $ally_unit.name will still follow your command. But other elves must be out on a separate island. How will we cross deep water?"
```
stevecotton, line 1857: With the names taken out, lines like this should have a `# po: speaker is a troll` hint.

data/campaigns/Under_the_Burning_Suns/scenarios/07b_Talking_with_Trolls.cfg
```diff
@@ -356,26 +356,11 @@
...
-                [/message]
-            [/else]
-        [/if]
+        {SET_ALLY_UNIT}
+        [message]
+            speaker=Spiritual Advisor
+            message= _ "We may be able to help you find a way back to the sunlit lands. In our temples we do keep records of the past. We have not walked above the earth for many many generations, not since the darkness drove us underground. But we are masters of the underground lands, and we have explored many tunnels. Recently one of our scouts found a path that leads north back to the sunlit lands, I think it may be the way you are trying to go. In reward for your achievements, we will help you. $ally_unit.name has volunteered to protect you and lead you back to the sunlight lands."
```
stevecotton, line 362: With the names taken out lines like this should have a `# po: speaker is a troll`. However, that points out a new problem - the `en_US` text isn't written as troll speech.

Wedge009: What do you mean by that? That not all lines spoken by trolls sound like the stereotypical 'Grog smash!' bad grammar? I noticed that too, when the author sometimes uses the same lines for both ally races, dwarves and trolls. But as I said in the original post, I haven't made (or intended to make) any dialogue changes here, just consolidate the existing lines.

Wedge009:
> While it's a good change, I think it shouldn't be backported. As a general rule (including for this change), I think any string changes that shouldn't be backported should also not be added to master until 1.16.1 has released, so that they can't conflict with any string changes that do get backported.
> 
> Conclusion: I think this PR should be postponed.

I had no intention of back-porting anything before the string freeze is over and when I realised I wasn't actually changing any lines I saw no reason to back-port it to 1.16 after all.

But I'm not sure I understand your point about 1.16.1 - there are already a number of changes in master that are pending back-porting, as you're no doubt aware. At this point I'm willing to postpone till at least 1.16.0 but I'm not sure I see the justification to wait further still.

data/campaigns/Under_the_Burning_Suns/scenarios/07b_Talking_with_Trolls.cfg
stevecotton: That's what I mean - a spiritual advisor should will be more learned than the average, but at the same time a troll spiritual advisor probably shouldn't sound exactly the same as a dwarf spiritual advisor.

It's an issue in the existing text, but seems something that should be fixed for 1.18.

stevecotton:
I agree that there's no point postponing past the time that the strings from master have been backported to the 1.16 branch. My mistake to say 1.16.1 instead.

data/campaigns/Under_the_Burning_Suns/scenarios/07b_Talking_with_Trolls.cfg
Wedge009: The dwarves don't have a spiritual adviser - allying with the dwarves has Kaleh and friends speaking with a dwarvish king in S7a.

Anyway, I'm glad you agree the shared lines (between dwarf and troll allies) are a separate issue.

data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg
Wedge009: Done.

data/campaigns/Under_the_Burning_Suns/scenarios/07b_Talking_with_Trolls.cfg
Wedge009: Didn't add the `po`hint here because the adviser is referring to the ally, not to himself, with the `$ally_unit.name` mention.

Wedge009:
Save files contain a snapshot of the scenario WML, but I suppose it could be a problem when moving to the next scenario after `ally_name` has changed to `ally_id`. On the other hand, aren't breaking changes expected between major versions?

Still, I suppose it's not hard to do, but I'm not familiar with the precise syntax. I'm assuming it's something along the lines of:

```
[if]
	[???]
		name=ally_name
	[/???]

	[then]
		[set_variable]
			name=ally_id
			value=$ally_name
		[/set_variable]
		{CLEAR_VARIABLE ally_name}
	[/then]
[/if]
```
Can you advise, please?

Wedge009:
Any idea why the ubuntu checks failed?

stevecotton:
> Any idea why the ubuntu checks failed?

Because `{ALLY_NAME_COMPATIBILITY}` needs to be inside an `[event]`. The actual error is
```
Validating data/_main.cfg against schema /home/wesnoth-CI/data/schema/game_config.cfg
Error (strict mode, strict_level = 0): wesnoth reported on channel error validation
20211007 02:37:43 error validation: Tag [if] may not be used in [scenario]
at campaigns/Under_the_Burning_Suns/utils/dialog-macros.cfg:86
    included from campaigns/Under_the_Burning_Suns/_main.cfg:158
    included from data/_main.cfg:38
    included from campaigns/Under_the_Burning_Suns/_main.cfg:168
    included from data/_main.cfg:38
    included from campaigns/Under_the_Burning_Suns/scenarios/12_The_Final_Confrontation.cfg:47
    included from campaigns/Under_the_Burning_Suns/_main.cfg:158
    included from data/_main.cfg:38
    included from campaigns/Under_the_Burning_Suns/_main.cfg:227
    included from data/_main.cfg:38
validation failed
```

Wedge009:
Thanks. I find it strange that the error was specific to the ubuntu checks, plus I didn't know that was a syntax rule.

stevecotton:
> For S12, I had the macro outside of any `[events]`s because the `[side]` declarations made references to `$ally_name`/`$ally_id`. Is it sufficient to have the macro in the 'prestart' `[event]`?

`prestart` is too late; I don't know if `preload` would be early enough (would need to test).

stevecotton:
I tested setting a variable in `preload`, but this is not early enough to be used in `[side]`. @CelticMinstrel is there a way S12's ai settings could be handled?

Wedge009:
I think I will separate this into the original intent of the PR (refactoring) and the subsequent compatibility request.

Wedge009:
Split the compatibility changes into #6301. This now just contains the original refactoring.

Wedge009:
Note that this was reverted in d5b95e621d5a11e6fecbd301117e5ca639164825 and #6158 re-opened.

PR #6178, opened 2021-10-08
DM Proofread Pass

max-torch:
Proofread and Copy Edit pass of Delfador's Memoirs. `wmlxgettext` was used to review all text.

### DIFF
--- a/data/campaigns/Delfadors_Memoirs/scenarios/22_Face_of_the_Enemy.cfg
+++ b/data/campaigns/Delfadors_Memoirs/scenarios/22_Face_of_the_Enemy.cfg
@@ -149,7 +149,7 @@
         [/message]
         [message]
             speaker="Sagus"
-            message= _ "I am the one whose place you took in the court of Wesnoth. But my daughter brought me what you seek, and I have acquired its power."
+            message= _ "I am the one whose place you took in the court of Wesnoth. But my daughter brought me what you seek, and I have gained its power."
         [/message]
         [message]
             speaker=Chantal


max-torch:
Fourth Draft is up.

data/campaigns/Delfadors_Memoirs/scenarios/02_This_Valley_Belongs_to_Me.cfg
```diff
@@ -65,7 +65,7 @@
             story = _ "Delfador’s wandering time with Methor took them to many strange places, both within and beyond the lands of men. They studied and practiced, applying magecraft to aid those they traveled among and to deepen their understanding of the hidden currents and powers of the world."
         [/part]
         [part]
-            story = _ "But the elder mage’s health was taxed by the rigors of the journey. A year after the two left Alduin, Methor settled in a tranquil valley at the edge of the Dulatus Hills, not far southeast of the capitol at Weldyn. He opened a small school of his own, and with Delfador assisting him he began training a handful of youths who might in time be worthy to enter the great Academy on Alduin."
+            story = _ "But the elder mage’s health was taxed by the rigors of the journey. A year after the two left Alduin, Methor settled in a tranquil valley at the edge of the Dulatus Hills, not far southeast of the capitol at Weldyn. He opened a small school of his own; and with Delfador assisting him, he began training a handful of youths who might in time be worthy to enter the great Academy on Alduin."
```
Wedge009, line 68: @max-torch @nemaara I'm sorry I missed the opportunity to correct this in this big review. With respect to:

> ...the capitol at Weldyn.

As I understand it, there's the US Capitol Building in the US capital, but it's incorrect - though seemingly rather common in US English - to use 'capitol' when 'capital' should actually be used instead, as is the case here. Any objection to changing this once the 1.16.2 string freeze is over?

max-torch: I wouldn't object. @Wedge009

data/campaigns/Delfadors_Memoirs/scenarios/16_Dark_Sky_Over_Weldyn.cfg
```diff
@@ -104,7 +104,7 @@
...
-            story={CAPTION ( _ "Delfador")} + _ "There are three things we must do. First, we must organize our defenses as best we can. Second, we need to ask the elves to create a diversion, to distract and delay Iliah-Malal. The elves will not be able to fight the undead head on on the field, but they can harass them by striking from forests where the undead dare not follow. And last, I have been studying and I think I now know how to close Iliah-Malal’s portal to the Land of the Dead. I will need a small escort, but this must be done immediately."
+            story={CAPTION ( _ "Delfador")} + _ "There are three things we must do. First, we must organize our defenses as best we can. Second, we need to ask the elves to create a diversion that will distract and delay Iliah-Malal. The elves will not be able to fight the undead head-on on the field, but they can harass them by striking from forests where the undead dare not follow. And last, I have been studying the Book of Crelanu and I think I now know how to close Iliah-Malal’s portal to the Land of the Dead. I will need a small escort, but this must be done immediately."
```
Wedge009, line 107: > The elves will not be able to fight the undead head-on on the field...

I know it was written this way in the original text before this PR, but re-reviewing this while updating localisation I think this might be better phrased as 'head-on **in** the field'. I recognise there's a subtle difference in meaning there, but at least it avoids the clumsy-sounding repetition of 'on'.

Any opinions on this?

data/campaigns/Delfadors_Memoirs/scenarios/23_Epilogue.cfg
```diff
@@ -8,13 +8,13 @@
     [story]
         [part]
             music=battle-epic.ogg
-            story=_ "Delfador returned to Weldyn, but not before secreting the Book of Crelanu in a place known only to him. To his surprise, Asheviere made no further mention of the events in the manor. But not long after, orcish raids increased to a degree that threatened the Kingdom. The King mustered his armies to deal with the orcish menace, leading to the great battle at the Ford of Abez that was lost through betrayal."
+            story=_ "Delfador returned to Weldyn, but not before secreting the Book of Crelanu to a place known only to him. To his surprise, Asheviere made no further mention of the events in the manor. But not long after, orcish raids increased to a degree that threatened the Kingdom. The King mustered his armies to deal with the orcish menace, leading to the great battle at the Ford of Abez that was lost through betrayal."
```
Wedge009, line 11: > To his surprise, Asheviere made no further mention of the events in the manor.

Another thing I just noticed (that was already present in the original text before this PR). Would it make sense to drop the 'further' in this sentence? As it stands, it implies to me that Asheviere had made some mention of the events in the manor (Delfador's attack on Sagus) but is no longer discussing it. Without the 'further' I interpret this as no mention at all, which seems to make more sense. It's a subtle difference, I know, but I thought it worth bringing up if this is supposed to be a big revision of the prose (forthcoming rewrite of the campaign notwithstanding).

data/campaigns/Delfadors_Memoirs/scenarios/16_Dark_Sky_Over_Weldyn.cfg
cooljeanius: I think either way is fine; the point about avoiding repetition does have me leaning towards "in" though

CelticMinstrel: I don't really think there's a need to avoid repetition when grammar requires it (a common case is two consecutive "that"s), and this seems like it's such a case. Unless "in" better fits the intended meaning, I'd leave it as "on".

Wedge009: I'm aware of those cases, but those are quite rare. Even disregarding the repetition aspect, I think it reads better to me with 'in', but perhaps I'm in the minority here. If people are comfortable to leave it as-is, so be it.

max-torch: https://owl.purdue.edu/owl/general_writing/grammar/prepositions/prepositions_of_location_at_in_on.html
That article explains it well.
One of the example sentences is "Three players are practicing on the field"

Wedge009: I was thinking of a different context of 'field', not like playing on a football field but fighting out in the field.

But no matter, I've already said I'm in the minority here.

data/campaigns/Delfadors_Memoirs/scenarios/23_Epilogue.cfg
Wedge009: Since I had no feedback on this suggestion (in contrast with the other one) I'm assuming there's no objections and updated this in 0c13523544d36138dd58a780d2dd4197ba5bef8b (I didn't think it was big enough to warrant a full PR).

data/campaigns/Delfadors_Memoirs/scenarios/16_Dark_Sky_Over_Weldyn.cfg
max-torch: https://glosbe.com/en/en/to%20meet%20on%20the%20battlefield%20%28idiom%29
Here's an site that has cited usage of fighting or meeting out _on_ a battlefield.
I suppose the field in the context of the dialog is indeed a battlefield.

Wedge009: I'm not disputing the usage of 'on'. I am simply saying that in the context I am thinking of 'in' is more appropriate - which clearly isn't how others are reading it. Like field-testing a new weapon or whatever.

I can't see why this is generating so much discussion when the other suggestion is arguably more significant. Makes me wish I never brought it up, should have just changed it without asking.

PR #6181, opened 2021-10-09
SoF- S6 - fix #6146 and related side-storing issue causing auto-defeat

### LINKED ISSUE #6146: Sceptre of Fire "already have this one" bug
- v 1.15.16
- downloaded from: 
   https://www.wesnoth.org/#download -> Development
   (the link has already changed to that of the newer version by the time I report this, so I don't know the exact link)
- Win 8

### Describe the bug
When I was starting the scenario The Dragon, I tried to buff my dwarves with Thursagan's Swiftness Rune. The dwarves rejected them, claiming they 'already have this one' although they obviously did not. I have tested a few times and it appeared the bug affected dwarves I had recalled during Towards the Caves and the Swiftness Rune alone, which was also the only rune I used in that scenario. That was as if the effects have been cancelled but the internal status had not been updated. 
Also, I am not sure if this is intended, but Rugnur was not eligible to the runes altogether in The Dragon (no blue rune spinning above him and no interface when clicked). 

### To Reproduce
(My guess)
1. Use Runes on Dwarves in Towards the Caves
2. Use the same Runes on the same dwarves in The Dragon

### Expected behavior
"I already have this one!"

### Screenshots
(none)

### Savefiles
[SoF-The Dragon-Auto-Save1.gz](https://github.com/wesnoth/wesnoth/files/7273425/SoF-The.Dragon-Auto-Save1.gz)

doofus-01:


Wedge009:
Does that mean you managed to resolve #6146 after all or will there still be potential problems?

doofus-01:
Yes, this should fix both problems in that issue.

Wedge009:
Great!

PR #6182, opened 2021-10-10
SoF S4 Make mining gold and coal clearer

### LINKED ISSUE #6154: [SoF] S4 Gathering Materials: Make mining gold and coal clearer
### First Proposal
In the scenario Gathering Materials from Sceptre of Fire.
I propose adding a narrator-hint after this line:
https://github.com/wesnoth/wesnoth/blob/dd3baa488150bfc8d8cd0105da9083009cb29ca0/data/campaigns/Sceptre_of_Fire/scenarios/4_Gathering_Materials.cfg#L705
Something which says "Dwarvish Miners are the  only units that can fetch and carry gold and coal."

#### Rationale
Yes, the existing dialog states this somewhat, but it can't hurt to improve clarity by restating the intention from a meta-perspective, further eliminating any remote chance of misinterpretation.

Alternatively, this could be an objective note instead of narrator-hint. Or both if desired.

### Second Proposal
Add an event that triggers when a hex containing gold and coal is revealed. For the event, scroll the map to center the gold/coal hex onscreen and make the hex border/flash blink using a colored border highlight. Have the unit that reveals the hex say something like "I found coal!" and "I found gold!"

#### Rationale
The player might not be aware when and where gold and coal have already been discovered - putting the mouse on the hex containing the gold and coal only says the terrain and does not confirm it is the gold/coal. The only way to confirm is to actually send a unit to it, which can be annoying if the player's units are already very far. Additionally, coal can be difficult to spot on the map due to its low contrast with the cave terrain.

#### Screenshot
Coal has been revealed on the map but I was completely unaware.
image

### Third Proposal
Make Scenario Objectives concise. Scenario objectives is not the place for any kind of story context such as "to create the sceptre", and the second objective is more of a note than an objective.
Current objectives:
image

### Task List
- [x] Add narrator-hint (or objectives note) about using miners
- [ ] Add event that highlights gold and coal upon discovering it on the map
- [x] Make Scenario Objective concise

Replay attached (version 1.15.18):
[SoF-Gathering Materials replay 20211004-164355.gz](https://github.com/wesnoth/wesnoth/files/7284487/SoF-Gathering.Materials.replay.20211004-164355.gz)

max-torch:
Resolves #6154.

### Changes
- Adds a narrator-hint about the Dwarvish Miner
- Adds an event that highlights gold and coal upon sighting it on the map
- Rewords Scenario Objectives to be more concise

Test run replay on 1.15.18:
[Highlight_gold_coal.gz](https://github.com/wesnoth/wesnoth/files/7318073/Highlight_gold_coal.gz)
The replay only demonstrates the sighted event and doesn't include the narrator-hint and reworded objectives.
The testing of the feature was on 1.15.18 because there were Lua errors that messed things up in the latest 1.17 master branch.

### Screenshot
image

### Method of Hex highlight

A hidden Dwarvish Fighter unit was placed on each of the gold and coal hexes. A `sighted` event was used to trigger the display of a message where `speaker=unit` so that the hex would get a red border. A `fire_event` containing `kill` was used to remove the hidden unit after the message. The hidden units were assigned to Side 2 but they were given the loyal trait so that they would not interfere with Side 2's income. The units were `petrified` so they would stay in place.

**UPDATE: If you scroll below and follow the conversation, a Map Labels implementation was applied instead of a hex sighted implementation.**


cooljeanius:
some campaigns use "fake" units specifically for this purpose instead of using existing regular units; see for example `core/units/Fog_Clearer.cfg` in mainline, or the "Pillager Attractor" in TEG: https://github.com/cooljeanius/The_Earths_Gut/blob/master/units/fake/Pillager_Attractor.cfg
...or FtF's "invisible" unit: https://github.com/cooljeanius/Flight_Freedom/blob/2d38eb52446984ef052e37c68fe2805d9d051eae/units/fake_units.cfg#L231-L249

max-torch:
Oh dear, I just realized that even in the screenshot we can see the unit name which needs to be removed.

max-torch:
I went ahead and converted the added narrator-hint
> Dwarvish Miners can extract and carry gold and coal.

to objective notes. This avoids having to sacrifice player immersion. I also felt this is what @doofus-01 would want.

max-torch:
> The WML validation fails; also the indentation is being changed by wmlindent.
> 
> ```
> Validating data/_main.cfg against schema /home/wesnoth-CI/data/schema/game_config.cfg
> Error (strict mode, strict_level = 0): wesnoth reported on channel error validation
> 20211012 14:27:59 error validation: Invalid key 'hidden=' in tag [unit]
> at campaigns/Sceptre_of_Fire/scenarios/4_Gathering_Materials.cfg:713
>     included from campaigns/Sceptre_of_Fire/_main.cfg:128
>     included from data/_main.cfg:38
> 20211012 14:27:59 error validation: Invalid key 'hidden=' in tag [unit]
> at campaigns/Sceptre_of_Fire/scenarios/4_Gathering_Materials.cfg:727
>     included from campaigns/Sceptre_of_Fire/_main.cfg:128
>     included from data/_main.cfg:38
> 20211012 14:27:59 error validation: Invalid key 'hidden=' in tag [unit]
> at campaigns/Sceptre_of_Fire/scenarios/4_Gathering_Materials.cfg:741
>     included from campaigns/Sceptre_of_Fire/_main.cfg:128
>     included from data/_main.cfg:38
> validation failed
> ```

It turns out use of `hidden` is not allowed. Well, is it okay if I use the fog clearer unit from core instead, because of the blank sprite image? Also, I will run wmllint and wmlindent on the cfg file before I resubmit the next fix for this.

max-torch:
Regarding, the latest commit I pushed:
> The WML validation fails; also the indentation is being changed by wmlindent.

I ran wmlindent. All it did was remove a single blank line. Should be okay now.

To resolve WML validation:
The issue was the use of `hidden=yes` in my PR. The fix was to use the Fog Clearer unit instead of a Dwarvish Fighter. However, the Fog Clearer is a unit that is marked for deprecation (from core I presume), so I made a copy of the Fog Clearer inside of the SoF directory and appended 'SoF' and '_SoF' to its id and filename respectively so that this feature will not fail if and when Fog Clearer is removed from core.

No more WML validation issues and wmlindent issues are introduced by this PR. If the team approves of the use of fog clearer for this, and the copying of fog clearer into the SoF directory, then this is good to go.

AI0867:
The Fog Clearer has been deprecated since 2012, because it was replaced with the `[lift_fog]` and `[reset_fog]` tags. Is there a reason those don't work here?

doofus-01:
> I went ahead and converted the added narrator-hint
    Dwarvish Miners can extract and carry gold and coal.
to objective notes. This avoids having to sacrifice player immersion. I also felt this is what @doofus-01 would want.

Sounds good to me.

I wonder if a lot of the issue the fog clearer and possibly any other visual changes would attempt to address could be solved with simply using a map label?  There could be a message explaining it in the context (right-click) menu, but it wouldn't be required.

AI0867:
> > The Fog Clearer has been deprecated since 2012, because it was replaced with the `[lift_fog]` and `[reset_fog]` tags. Is there a reason those don't work here?
> 
> I am not using fog clearer to lift fog. I am using it to make the `sighted` event function as a 'hex sighted' event. In this scenario, I am using it to trigger dialog when the player sights gold or coal.

Wouldn't something like this work for that?
```
[event]
    name=enter_hex
    [filter]
        side=1
    [/filter]
    [filter_condition]
        [have_location]
            x,y=$coal1.x,$coal1.y
            [filter_vision]
                side=1
            [/filter_vision]
        [/have_location]
    [/filter_condition]
    ...
[/event]
```

cooljeanius:
> > > The Fog Clearer has been deprecated since 2012, because it was replaced with the `[lift_fog]` and `[reset_fog]` tags. Is there a reason those don't work here?
> > 
> > 
> > I am not using fog clearer to lift fog. I am using it to make the `sighted` event function as a 'hex sighted' event. In this scenario, I am using it to trigger dialog when the player sights gold or coal.
> 
> Wouldn't something like this work for that?
> 
> ```
> [event]
>     name=enter_hex
>     [filter]
>         side=1
>     [/filter]
>     [filter_condition]
>         [have_location]
>             x,y=$coal1.x,$coal1.y
>             [filter_vision]
>                 side=1
>             [/filter_vision]
>         [/have_location]
>     [/filter_condition]
>     ...
> [/event]
> ```

I've found that `enter hex`/`exit hex` events can really slow down the game, personally, since they have to be checked so often...

max-torch:
Map Labels are now in use. No more hex sighted event. Please see updated changed files. Please see if current state of PR is acceptable.

Screenshot:
image

doofus-01:
I haven't used labels much, can you clear them, I guess by setting the text key to a (non-translated) empty string?

There might still be some small technical things to do, but conceptually this seems good to me.

max-torch:
> Now it fails CI because of the blank labels. Either validation needs to be fixed or there needs to be some documented way to clear the labels. EDIT: Of course, fixing validation would be outside of this PR

How does one confirm that it's the blank labels that is causing the CI to fail?

max-torch:
Thanks to WhiteWolf, I found the {REMOVE_LABEL X Y} macro. Tested working on the scenario. Replay attached. Let's hope the CI passes.
[removelabelsmacro.gz](https://github.com/wesnoth/wesnoth/files/7360681/removelabelsmacro.gz)

Wedge009:
After playing SoF recently (without these changes) I think this is a welcome addition. I had a _really_ hard time spotting the gold and coal, _especially_ the dark coal...

PR #6189, opened 2021-10-13
DM S07 Fix for player having to pass turn in order to meet alternative win condition (#6173)

### LINKED ISSUE #6173: DM S7 Night in the Swamp: Still required to press End Turn after meeting victory condition
 - 1.15.18
- Windows 10

### Describe the bug
First, I need to disclaim that I am aware of 20fd6d80ba8394c303be956924317b486fe370b9.
After moving to the hex of the last stone to destroy it, and having no enemies around, the scenario does not trigger victory until the player presses End Turn, and the victory plays out at the start of the next turn. This behavior is unlike other victory conditions in other scenarios throughout the campaigns.

### To Reproduce
Accomplish the alternative objective of DM S7.

### Expected behavior
It should be detected immediately once the player satisfies the objectives so that the victory screen can immediately appear. It should not be checked only at the start of a turn. **Similar to how the game does not require you to End Turn after you have killed an enemy leader for scenarios where the objective is to kill an enemy leader.**

### Screenshots
Victory Condition has been met. Turn 7. Scenario has not yet ended.
image
Victory Condition is now checked. Had to End Turn. You can see in screenshot it is Turn 8. Replay buttons are now greyed out.
image

### Savefiles
[DM-Night in the Swamp replay 20211006-191338.gz](https://github.com/wesnoth/wesnoth/files/7303282/DM-Night.in.the.Swamp.replay.20211006-191338.gz)

### Additional Context
Secrets of the Ancients S5 Blackwater also has a victory condition requiring the player to kill all enemy units, and you dont have to press End Turn to get the victory.

Descacharrado:
As stevecotton suggested a good way of tackling this kind of win condition is to have a counter which counts to 2 on EASY and NORMAL difficulty and 3 on HARD (Line 133). Every time a generator is destroyed one is added to the counter variable called generators_destroyed (L 126-129). The edited macro would now work like this:

- If the player destroys all the generators and there are no enemy units remaining the scenario would end (L 135 - 152)
- If the player destroys all the generators and there ARE enemy units remaining, then the objectives are updated and shown to the player (now the optional win condition is to defeat the remaining enemy units, L 154-179) and a child event (L 181-205) is spawned, which makes it so if the player defeats all enemy units he wins.
- Before any [endlevel] tag there has to be a {CLEAR_VARIABLE} clearing the created variable.

This has been tested on my machine and works in any case (in 1.16.0 too), but further testing and review would be beneficial.

EDIT: As suggested, the only use of this macro occurs in scenario 7, which made the definition inside the campaign.cfg possible. If the macro was wanted to be used somewhere else too many arguments would need to be added (and they would be pretty obscure, just imagine passing objectives, speakers and strings as arguments).

( Closes #6173 )


Wedge009:
This works for me, for both the situation when there are enemies present after destroying the generators and when there are not any. If you fix the capitalisation and indentation on the comment, and update the `#ifdef` condition, I think this will be suitable for both master and 1.16 branches.

@stevecotton If you have time can you see if this matches what you had in mind when you made the suggestion?

Descacharrado:
> Edit: A reminder - if you want this PR to close the issue, you need to use one of the closing keywords along with the issue number in your commit message, eg `closes #6173`.

I think just commenting it (as I did in the bottom of the first comment) works too, because if you visit the Issue this can be seen:

Sin título

Idk tho, I'm new to committing here.

kabachuha:
Not a big problem, just for the variable wording. I think "spawners" is a more commonly accepted word than "generators" in level design.

Descacharrado:
The macro having objectives could be solved, as was said, by including it in the scenario.cfg file or we could pass additional WML as an argument to the macro, to keep it usable for other scenarios in the future. I like the idea of scenario-specific macros but I like modular coding better. Anyways, if no there are no further comments I will push what I think is best in a few days.

data/campaigns/Delfadors_Memoirs/scenarios/07_Night_in_the_Swamp.cfg
```diff
@@ -130,7 +130,7 @@
         [if]
             [variable]
                 name=generators_destroyed
-                greater_than={ON_DIFFICULTY 1 2 2}
+                equals={ON_DIFFICULTY 2 2 3}
```
Wedge009, line 133: Is this correct? I'm no expert with WML but it looks to me like it's changing from:
Easy: > 1
Normal: > 2
Hard: > 2

to:
Easy: == 2
Normal: == 2
Hard: == 3

I suppose `generators_destroyed` is a variable you've introduced with the PR as a whole, so I'm guessing this means your original implementation for Normal was incorrect, since the number of generators is 2, 2, 3 for the respective difficulties.

Descacharrado: > this means your original implementation for Normal was incorrect.
That statement is correct, there are 2 generators on EASY and NORMAL and 3 on HARD and while testing the commit 362ebc0 I found that the game was not finishing on NORMAL so I decided to fix it and change greater_than to equals as that made more sense imo.

In conclusion, the lines erased in 362ebc0 from generators.cfg are not the exact same as the ones I added to the scenario.cfg so I added a bug, which then I had to patch up on 5eaba0f.

Wedge009: Was just worried at first that you were changing something but then afterwards noticed the bug for Normal. In general (from a programming perspective) `>` or `>=` is safer practice than just testing for equality but I'm okay with leaving this PR as using `equals`.

I'll try to retest before merging, once the string freeze is lifted, but if it's largely the same (functionally) as what you originally submitted a few weeks ago that should be good.

Wedge009:
Retest looked good to me, both for destroying the stones with enemies remaining as well as destroying the stones last.

PR #6192, opened 2021-10-14
TRoW S19 Chest is automatically collected at the end of the scenario

### LINKED ISSUE #6141: TRoW S19 The Vanguard: Ensure the player receives the gold the orcs drop before ending the level
 - 1.15.18
 - Steam
 - Windows 10

In the beginning, the orcs' dialog talks about a chest of gold that they drop:
image
Later, the player has to move a unit onto the hex containing the chest of gold to get it:
image
**It is not 100% obvious** that this wooden box sprite is the aforementioned chest of gold, and **the player can finish the scenario by killing the leader without ever picking up the gold**. 

It seems that players are meant to get this chest as it's referenced in the dialog, and is placed right close to the leader, on the way to approaching the enemy leader, who you have to defeat to finish the scenario.

The desired feature is to make the player receive gold as an event triggered by killing the troll leader and remove the chest from the map. This is consistent with what occurs in other scenarios in the campaign, such as in the scenario Troll Hole, where the player finds gold after killing an enemy leader. Let the unit that delivers the killing blow say something like "I found the chest of gold the orcs dropped nearby somewhere on the battlefield. It seems the trolls were trying to retrieve it. Some _friends_ these orcs are."

Descacharrado:
The S19's gold chest close to the troll enemy leader will now be collected at the end of the scenario if the player misses it (which is unlikely but could happen). A new dialogue may be necessary in line 704.

The expected behaviour of the changes is:

- If the player moves a unit during the scenario to the chest, it opens and cannot be opened again and at the end of the scenario nothing will happen.
- If the player misses the chest, Haldric (Or the closest ally unit, if that is ever wanted to be implemented) moves to the chest and opens it, and the scenario ends.

(Closes #6141)

### DIFF
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/19_The_Vanguard.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/19_The_Vanguard.cfg
@@ -318,12 +318,14 @@
 
     [event]
         name=get chest
+
         [sound]
             name=open-chest.wav
         [/sound]
         [store_unit]
             [filter]
                 x,y=$xx1,$yy1
+                side=1
             [/filter]
             variable=chest_unit_temp
         [/store_unit]
@@ -331,6 +333,7 @@
             speaker=$chest_unit_temp.id
             message= _ "I’ve found the orcs’ chest! It’s filled with gold."
         [/message]
+
         {CLEAR_VARIABLE chest_unit_temp}
 
         [remove_item]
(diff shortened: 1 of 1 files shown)


stevecotton:
I think it should have new dialog, but feel it doesn't need the animation of the unit moving to the chest.

Does the current implementation still work if there's an enemy unit on the hex when the `victory` event fires?

nemaara (approved):

Issue #6196, opened 2021-10-16
UtBS S6b In_the_Domain_of_Dwarves Bugs and Issues


max-torch:
### Game and System Information
 - 1.15.18
 - Steam
 - Windows 10

#### Describe the bug
If the dust devil or Elyssa triggers the Dwarf Sergeant dialog, the dwarves call the dust devil or Elyssa a troll.

#### To Reproduce

1. Trigger the dwarf sergeant dialog with the dust devil or Elyssa, and keep other units away to make sure only the dust devil was recognized.
2. Dwarves believe they see a troll.

#### Expected behavior
Dwarves should see a dust devil and say "what in the Earth's Guts is that?"
Upon seeing Elyssa, the Dwarves could say "What's a little girl like you doing here? Are you lost?"

#### Screenshots
![image](https://user-images.githubusercontent.com/41433185/137583348-86f8c4db-1ac6-4b2c-a763-b4f48f984794.png)

#### Savefile
Dust Devil Replay: 
[dust devil.gz](https://github.com/wesnoth/wesnoth/files/7357345/dust.devil.gz)

#### Describe the bug
A unit can get stuck in an impassable cave wall (either 16,31 or 23,30) during the scripted event where the Dwarvish Stalwart seals the tunnel.

#### To Reproduce

1. Send a unit close enough to the Dwarvish Stalwart (17,31 or 22,30) but far enough not to trigger the event.
2. End Turn.
3. Move the unit next to the Dwarvish Stalwart.
4. After the dialog, the unit is stuck inside a cave wall.

#### Expected behavior
No units should suddenly be getting stuck in walls. If the player manages to get beside the dwarf in one turn, I suggest changing the dialog to say "Woah, where did you come from! Get out of there, you're standing on the fuse! You sneaky little elf/troll/floating mound of dust/mage!" and then the wall cave-in never occurs, the player is rewarded for quickly sneaking up on the guard.

#### Screenshots
Unit is one turn away:
![image](https://user-images.githubusercontent.com/41433185/137583750-40783b78-fd00-48ee-91f1-e6075f7e6c4c.png)
Unit is now stuck:
![image](https://user-images.githubusercontent.com/41433185/137583764-d9296de5-7734-4c19-9322-12fd2b7279c4.png)

#### Savefile
[unit can get stuck.gz](https://github.com/wesnoth/wesnoth/files/7357385/unit.can.get.stuck.gz)

#### Describe the bug
[...]

PR #6198, opened 2021-10-16
UtBS S06b Fix Dwarf Sergeant confusing a dust devil or human for a troll

Descacharrado:
Fix for one of the tasks in Issue #6196. Used a [switch] instead of [if]-[elses] for clarity. **⚠New strings have been added ⚠** and have to be revisioned, make sure string freeze is lifted before merging/backporting.

Dust devil (monster) string: 

Dust devil 1
Dust devil 2

Elyssa (human) string:

Human 1
Human 2

In case other race is added to previous scenarios in the future:

Other 1
Other 2

Troll and Quenoth elf strings are unchanged.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
@@ -694,7 +694,7 @@
                 value=human
                 [message]
                     x,y=10,34
-                    message= _ "No, It’s a... human?"
+                    message= _ "No, it’s a... human?"
                 [/message]
                 [modify_unit]
                     [filter]
@@ -712,7 +712,7 @@
                 [message]
                     x,y=10,34
                     #po: "it" is referring to the dust devil
-                    message= _ "No, It’s a... what is that?"
+                    message= _ "No, it’s a... what is that?"
                 [/message]
                 [modify_unit]
                     [filter]
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
@@ -654,18 +654,10 @@
             message= _ "Oh let me guess, a big nasty troll, right? And when I turn around you flee like the cowards you are. Do you think I’m stupid enough to fall for that one?"
         [/message]
 
-        [if]
-            [variable]
-                name=unit.race
-                equals=quenoth
-            [/variable]
-            [or]
-                [variable]
-                    name=unit.race
-                    equals=elf
-                [/variable]
-            [/or]
-            [then]
+        [switch]
+            variable=unit.race
+            [case]
+                value=quenoth
                 [message]
                     x,y=9,36
                     message= _ "No, it’s an elf!"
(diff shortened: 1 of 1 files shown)


Wedge009:
Haven't tested this but code makes sense to me, and good on you for adding a default case too.

stevecotton (approved):

data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
```diff
@@ -696,8 +689,60 @@
...
+                [modify_unit]
+                    [filter]
+                        id=Dwarf Sergeant
+                    [/filter]
+                    facing=se
+                [/modify_unit]
+                [message]
+                    speaker=Dwarf Sergeant
+                    #po: "it" is referring to the dust devil
+                    message= _ "What in the Earth’s Guts is that? Kill it!"
```
Wedge009, line 726: I'm sorry I missed this in the original review - heck, I was the one who merged this - but it doesn't look right to have 'Earth's Guts' capitalised. Are we referring to something specific here? I think these should both be lower-case.

Descacharrado: I have given it some thought, it also looks to me that both of them should be lower case. At first, I was thinking that at least Earth should be upper case, but probably there is no "Earth" in wesnoth, only "earth". As far as I know, we are referring to the underground, so maybe we should add that in the po hint aswell. How shall we proceed?

Wedge009:
Capitalised Earth would be referring to our planet, which no-one in Irdya should really know about. (: As you described, earth can refer to the soil/ground/dirt.

If you don't want to go to the trouble of making another PR just for this small change, you can put your revised po hint here and I can update some time tomorrow (I'm about to go to bed soon). Otherwise, just go ahead and make a PR as normal.

I'm thinking something along the lines of:
```
#po: 'earth's guts' is just a reference to the Dust Devil's elemental nature
message= _ "What in the earth’s guts is that? Kill it!"
```

Feel free to make your own suggestions.

data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
Wedge009: Capitalised Earth would be referring to our planet, which no-one in Irdya should really know about. (: As you described, earth can refer to the soil/ground/dirt.

If you don't want to go to the trouble of making another PR just for this small change, you can put your revised po hint here and I can update some time tomorrow (I'm about to go to bed soon). Otherwise, just go ahead and make a PR as normal.

I'm thinking something along the lines of:
```
#po: 'earth's guts' is just a reference to the Dust Devil's elemental nature
message= _ "What in the earth’s guts is that? Kill it!"
```

Feel free to make your own suggestions.

Descacharrado: I still have the branch of this PR, if I were to edit it there, would that be fine? Anyways this is my suggestion:
```
#po: 'earth’s guts' references the underground. What is being talked about is the dust devil.
message= _ "What in the earth’s guts is that? Kill it!"
```

Wedge009: I think since this PR is already merged even if you added to the original branch you submitted the request from you'd have to create a new PR anyway.

Thanks, I'll update this in a short while.

Issue #6197, opened 2021-10-16
UtBS S6b In_the_Domain_of_Dwarves - Improve Amulet Easter Egg Logic


max-torch:
### Game and System Information
 - 1.15.18
 - Steam
 - Windows 10

### Suggestion 1
Make sure Grog doesn't die in the fight where he is introduced. It happens probably 1 out of every 5 games. Perhaps use scripted forced hits and misses.

#### Screenshots
Before even getting close. Grog has already taken so many hits and is about to die. His health bars is practically empty.
![image](https://user-images.githubusercontent.com/41433185/137586369-915ba0e3-93aa-4abd-bce7-fa636b89d1ab.png)

#### Savefile
[Grog death replay.gz](https://github.com/wesnoth/wesnoth/files/7357489/Grog.death.replay.gz)

#### Rationale
Practically all players will restart once they see Grog die like this.

#### Forum complaints of this
The complaint is inside the spoiler tag:
- https://forums.wesnoth.org/viewtopic.php?p=581506#p581506

### Suggestion 2
Improve amulet easter egg logic. Currently, when the player kills the dwarf hermit the player finds an amulet. Somehow, this amulet is teleported all the way across the lake to the unit that steps on the rune. I have two suggestions to improve this:
- Place a rune on the hermit island. Make a unit wearing the amulet step on that rune at the same time the unit across the lake also steps on their rune become the requirement that opens the way to the crypt.
- Make the unit that steps on the rune ask "can we examine that amulet for clues?" and then the person who found the amulet says "there is writing on the amulet which says 'speak friend'". And then the unit across the lake also says 'speak friend' and is surprised to see the way to the crypt open.

#### Forum complaints of this
- https://forums.wesnoth.org/viewtopic.php?p=624948#p624948
- https://forums.wesnoth.org/viewtopic.php?p=498263#p498263

### Suggestion 3
When the dwarf sentinel on the bridge dies, and triggers the sealing of the south tunnel that leads to the dwarf chieftain, it's possible for the player to not see what exactly occurred. We get that some detonation took place but it's not clear where. I suggest scrolling the map to the spot where the tunnel gets closed, lifting the fog, then making the explosion happen which closes the tunnel, so that the player can clearly see it.

#### Screenshot
We learn that a charge has been detonated, but we have no idea where and for what purpose. It lacks context when it is introduced.
[...]

PR #6210, opened 2021-10-18
UtBS S06b Grog spawn overhaul

Descacharrado:
Addresses bug 3 in #6196 and suggestion 1 in #6197.

Addressing bug 3: Previously, Grog spawned when a unit stepped on (6-8,28-34) making a unit with more than 6 mp able to see that there is no Grog at (7,28) from 9,35 (or further if the unit has even more than 7 mp).

(outdated) The expected behaviour now is for Grog to be there the entire time (spawns in the prestart event, Lines 398 to 413), fighting with the dwarf without the possibility of either of them dying **before** he is discovered, as their chance to hit would be null (Lines 751 to 762). Everytime they fight, their xp is reset and 8 hp is drained from Grog (so he does not go above 25 hp before being found, Lines 765 to 851). Once Grog is discovered (event line 853), the variable defined in line 291 is cleared (Line 867), and thus none of the events previously described would make effect.

Addressing suggestion 1: previously, one could get unlucky and have Grog die (which is kind of common).

(outdated) This can be reduced in the easier difficulties as seen in line 868, making it less likely for grog to die in turns 2+ after he is discovered. (Has to be kept in mind that, even if recondite, the chance for him to die will always exist using this method), if it is really wanted for him not to die on, lets suppose, the first three turns he is discovered, more turn-by-turn code should be added or more hp should be added to grog in line 407 (I have already tested this option, but Grog always crushed the dwarf by himself, making the player's tactical choices feel irrelevant).

(update) Grog will now spawn at the start of the scenario, simulate fighting with the dwarf the entire time (even if off-screen, behind shroud) and when he is found, the experience restrictions would be lifted, so if many more turns pass, the dwarf would advance before a troll which most likely would have less than 25 hp, making it very easy for him to die.


max-torch:
Can I ask what the easiest way to test the changes in this PR is?

The only method I know of at the moment is to load a replay of the previous scenario, press continue play so that a start of scenario save of this scenario gets generated, and then play the scenario until reaching Grog. What would be a more professional way?

Descacharrado:
> Can I ask what the easiest way to test the changes in this PR is?
> 
> The only method I know of at the moment is to load a replay of the previous scenario, press continue play so that a start of scenario save of this scenario gets generated, and then play the scenario until reaching Grog. What would be a more professional way?
> 
> I already checked out the pull request. Also, is it possible to test this in 1.17 latest commit or would you have to copy the cfg to the 1.15.18 wesnoth to test this?

use `git-pr number remote` to check out a PR.

Also to test the changes faster, once you have the cfg in place (by creating a new local branch with the PR changes or by copying it) you can start the campaign and use the following commands, by pressing `:` ingame.

- `debug`
- `choose_level` -> select the level with the changes

Other useful commands to test and debug are

- fog / shroud (removes fog / shroud, on command line)
- inspect (to inspect variables, on command line)
- shift+c (creates a new unit, ingame)
- shift + k (kills a unit, ingame)

nemaara (approved):

Descacharrado:
Resolved conflicts in a812d31

Issue #5133, opened 2020-09-06
[modify_unit]x,y=recall problby won't work


gfgtdf:
[modify_unit]x,y=recall problby won't work

During https://github.com/wesnoth/wesnoth/pull/5131 i noticed that wml does that sometimes.

Edit with explanation, now that another report of this has clarified it for me: trying to move a unit from the map to a side's recall list with `[modify_unit]x,y=recall,recall` or between sides' recall lists with `[modify_unit]side=2` will not work with the fast-path implementation. The old slow-path implementation would store and unstore the unit, using the fast path means we don't get any of the special-case handling that's in the unstore-unit code.

The effect of `[modify_unit]x,y=` to move a unit to an occupied map-hex is also likely to have changed. Tests are needed.

The `[modify_unit]side=` is reported in https://r.wesnoth.org/t54977
A more trivial effect of the fast-path, that the animation code isn't notified about the changes, is logged as #4978.

stevecotton:
I searched but could only find one instance of this. It's a macro that happens at end-of-scenario, but if it doesn't work then the schema should be improved to catch it.

https://github.com/wesnoth/wesnoth/blob/54f86f0c9b1c53cb2d855c47deccd566dfb3f1bd/data/campaigns/Delfadors_Memoirs/utils/sides.cfg#L445-L448

gfgtdf:
> , but if it doesn't work

The problem is that it did probably work in 1.14 but we changtes the [modify_unit] implementation in 1.15

CelticMinstrel:
This sounds like it should be relatively simple to fix if true - just add a special case in `[modify_unit]` if x = "recall".

cooljeanius:
Does this apply only to cases where the recall list is explicitly specified, or also to cases where the recall list just happens to get caught in the filter implicitly?

stevecotton:
Fixed by removing the fast path in #6223.

PR #6223, opened 2021-10-24
Remove the fast path of [modify_unit]

stevecotton:
Always use the "slow path", because it stores and unstores the unit, triggering
the desired side effects of unstoring a unit.

Fixes issue #5133, and tests that with the new unit test.

Fixes bug #4978, that changing the facing wasn't updating the display.

This corresponds to 1.16's commit 13c5d8a96ead8d2a7f0435f60b8ec7f7e0af0972,
that commit merely disabled the fast path to be a minimal change.


stevecotton:
Please take the role of `somebody` and log them yourself.

stevecotton:
Is there a reason to wait longer before merging this?

Pentarctagon:
Not afaik.

stevecotton:
Stable has #6222's 13c5d8a96ead8d2a7f0435f60b8ec7f7e0af0972 instead, which was the minimal change to disable the fast path. Is something still broken in stable?

CelticMinstrel:
…did I ever create the two new issues as mentioned above?

gfgtdf:
> So, from that PR it seems that the main advantage of the "fast path" wasn't speed, rather it was smaller save files

Its wasn't exactly about the savefiles size but rather about not storing redundant information in them. The slow path which uses [unstore_unit] overwrites the whole unit, then everything is stored in the savefile.  Now when in a newer wesnoth version or addon version for example changes changes the definition of an ability (like it happend with backstab in wensoth 1.14 or 1.16) The unit form the savefile would still use the old ability definition, leading to bugs being hard to fix because the bad code is stored in the savefiles, or as in case of the "backstab chance" annoying  "deprecated" messages.

When the fast path is used, the game realizes that for example only the units hp is changed, and doesn't store unit units abilities in  the savefile, so when the game is loaded in a newer wesnoth or addon version changes to the units ability (or other stat changes of the unit type) take effect immidiately.

PR #6235, opened 2021-10-26
changes for wc4p 

sevu:
Changes to WC2 to support a 4th side.


knyghtmare:
Okay, after trying several times, I will be debugging my way through the scenarios since the 4-player mode is too much to manage (as in its frustrating to manage 4 sides and reminds me of NR...which i dislike). I will be checking for bugs mostly, especially the 4th player being denied a keep/castle in one rare case and maybe some other things such as one player unfortunately having their castle place right next to the enemy base and may be overwhelmed due to the enemy unit count. Normally, one of the players dies in my playthroughs and I get to the 3player mode...

knyghtmare (changes requested):
Okay, finally finished 4 player mode. Took a while.
### Feedback/Review:
1. There is a very rare chance in the first scenario where player 4 is denied a keep and castle.
2. In first scenario, one player has a terrible misfortune of being positioned nearest to enemy keep (like 10 hexes away?) and is mostly likely to meet an untimely end.
3. In later scenarios (with 4-6 enemy sides), one of the sides always has a chance of being "_sandwiched_" by two enemy sides where survival chances are negligible.

soliton-:
I would suggest to factor out the changes done here that have nothing to do with 4 player support and check/commit them separately.

Regarding 4 player support I would suggest a more generic approach instead of adding a bunch of checks for a fourth player at random places and support any number of players as much as is reasonable instead or else we'll do this again the next time someone thinks we need support for 5 players. Although given that WC has existed for more than a decade now and nobody has made a version with more than 3 players before as far as I'm aware... the question is also if there is a reasonable need for this at all.

PR #6241, opened 2021-10-30
Use the CI to generate artifacts that can be used for releases

Pentarctagon:


Pentarctagon:
@loonycyborg Does this look good for the Windows/Steam Linux artifacts?

.github/workflows/ci-scripts/ubuntu.sh
```diff
@@ -12,14 +12,29 @@ echo "CFG: $CFG"
...
+		./make_steam_build || exit 1
+		tar -cf steambuild.tar steambuild
+		mv steambuild.tar ~/steambuild.tar
+elif [ "$IMAGE" == "mingw" ]; then
+		git archive --format=tar HEAD > wesnoth.tar
+		mv wesnoth.tar ~/wesnoth.tar
+		cd utils/dockerbuilds/
+		./make_mingw_build || exit 1
+		cd mingwbuild
+		mv $(find . -maxdepth 1 -regex "./wesnoth.*-win64.exe") ~/wesnoth-installer.exe
```
soliton-, line 26: This should be written as:
```suggestion
		mv ./wesnoth*-win64.exe ~/wesnoth-installer.exe
```

loonycyborg:
Not sure what exactly it's doing about windows artifacts. Resulting mingwbuild dir contains the installer, wesnoth(d).exe and dlls as well as "transtations" dir are the steam release. Everything else there shouldn't be used for upload.

Pentarctagon:
I don't follow - there is no artifact created from the mingwbuild folder.

loonycyborg:
The dir you move installer from also should contain exes and dlls and translations for uploading as steam windows release.

Pentarctagon:
Those would be the same as would be extracted for the itch.io upload, right?  If so it'd probably more sense to get them that way than having them included as something additional to download.

loonycyborg:
Yes but this would also include particular files like data/, so extracting installer seemed like safer option to me, to make sure that full set is same in both releases.

Pentarctagon:
Aren't the same files uploaded to both Steam and itch.io for Windows (aside from the dist file)?

loonycyborg:
Yes, but steam upload is made according to filters in steam .vdf files, which aren't even committed atm.

Pentarctagon:
So two questions then:
* Can the vdf files be committed?
* How does that affect using the files extracted from the standalone executable for the Steam upload?  As long as the same files are available either way then the filters would continue working the same as they currently do wouldn't they?

loonycyborg:
They can be committed yes, but filters there are only useful for steamcmd anyway. Though core issue here is to specify this full list in one place, or at least in as few places as possible. With current way list is duplicated: 1 in container that makes installer, 2 steam vdf file. Since itch upload is run outside of the container it would either need a third specification of the list or reuse windows installer. Making itchio script read steam vdf files would be more pain than it's worth.

Pentarctagon:
This is not making any sense to me.  The standalone installer is generated which contains all the files that needs to be uploaded to both itch.io and Steam.  The itch.io script extracts the contents that need to be uploaded and uploads them with butler.

Why would you not be able to just take those same files that were uploaded to itch.io and upload them to Steam with steamcmd?

loonycyborg:
They have some files both generated from the container and from game's original directory, data/ and the like. But not all files from there should be uploaded, for example not attic/. Particular subset of things that should be uploaded can change in the future and it's best to maintain it in few places.

Pentarctagon:
Right, so I'm imagining it like:
* Download the standalone installer artifact > upload to SourceForge
* Extract the needed files from the standalone installer to a folder somewhere > upload to itch.io with the script
* Use steamcmd from that same folder to upload to steam
I don't know why directories like attic would be even tangentially involved here.

loonycyborg:
I guess this way would work too, just need to modify vdf to draw from unpacked installer instead of from both git checkout and mingbuild/ dir.

loonycyborg:
In fact I'm doing this current way because I've set up steam stuff before I knew how to unpack those installers :P

loonycyborg:
Also could you keep same name(with version number) when moving the installer?

Pentarctagon:
I'm not entirely sure.  Github Actions apparently uses the same variable for both the branch and the tag, so I'm not sure without having this actually run on a tagged commit whether it will use the tag or if only the branch name is available.

loonycyborg:
Maybe just keep original name that is set according to wesconfig.h?

Pentarctagon:
And get it from `wesnoth --version` or is there a better way to get that?

loonycyborg:
Yes, don't change file name when moving. I guess running a regex on wesconfig.h would work too.

Pentarctagon:
It'd need to be with wesconfig I think, since the mingw output executable wouldn't be able to be run at all even to just get the version.

Pentarctagon:
@loonycyborg the version is in the filenames now.

loonycyborg:
Why it can't be `mv ./wesnoth*-win64.exe ~/`?

loonycyborg (approved):

Issue #3352, opened 2018-07-19
wmlindent crashes when sending SIGINT


sevu:
Aborting the program with `ctrl+c` shuts it down with a crash.
$ wmlindent 
```
^CTraceback (most recent call last):
  File "/home/space/git/wesnoth/data/tools/wmlindent", line 323, in <module>
    arguments, "|".join(exclude))
  File "/home/space/git/wesnoth/data/tools/wmlindent", line 249, in convertor
    linefilter("standard input", sys.stdin, sys.stdout)
  File "/home/space/git/wesnoth/data/tools/wmlindent", line 322, in <lambda>
    convertor(lambda n, f1, f2: reindent(n, f1, f2),
  File "/home/space/git/wesnoth/data/tools/wmlindent", line 118, in reindent
    for line in infp:
KeyboardInterrupt
```

ProditorMagnus:
I would say that is expected.

sevu:
I would say it should exit without Traceback.

GregoryLundberg:
SIGINT is the interrupt signal. The terminal sends it to the foreground process when the user presses ctrl-c. The default behavior is to terminate the process, but it can be caught or ignored. The intention is to provide a mechanism for an orderly, graceful shutdown.

SIGQUIT is the dump core signal. The terminal sends it to the foreground process when the user presses ctrl-\. The default behavior is to terminate the process and dump core, but it can be caught or ignored. The intention is to provide a mechanism for the user to abort the process. You can look at SIGINT as "user-initiated happy termination" and SIGQUIT as "user-initiated unhappy termination."

SIGTERM is the termination signal. The default behavior is to terminate the process, but it also can be caught or ignored. The intention is to kill the process, gracefully or not, but to first allow it a chance to cleanup.

SIGKILL is the kill signal. The only behavior is to kill the process, immediately. As the process cannot catch the signal, it cannot cleanup, and thus this is a signal of last resort.

soliton-:
This is the standard behaviour of python scripts.

CelticMinstrel:
This is normal, but I think it might be good for the program to catch this exception and terminate silently (or with a simple one-line message) rather than with a traceback. It probably applies to the other Python tools too (wmllint, wmlscope, wmlxgettext).

PR #6242, opened 2021-10-31
KeyboardInterrupt Handlers for wmltools

max-torch:
Closes https://github.com/wesnoth/wesnoth/issues/3352.

wmllint and wmlscope already had KeyboardInterrupt Handlers. I just edited their displayed messages for consistency. wmlindent and wmlxgettext did not have interrupt handlers and it had to be defined in their scripts.

wmllint and wmlscope use try-except clauses to define the interrupt routine.
I used a listener, `signal.signal`, to handle the interrupt routine so that I wouldn't have to wrap all the existing code in a try-except clause and change all the indentation.

Approver, please try running all four scripts and then pressing ctrl-c to abort so that you can verify the new behavior.

### DIFF
--- a/data/tools/wmlindent
+++ b/data/tools/wmlindent
@@ -61,7 +61,7 @@ if there's an indent open at end of file or if a closer occurs with
 indent already zero; these two conditions strongly suggest unbalanced WML.
 """
 
-import sys, os, argparse, filecmp, re, codecs
+import sys, os, argparse, filecmp, re, codecs, signal
 from wesnoth import wmltools3 as wmltools
 
 closer_prefixes = ["{NEXT "]
@@ -299,7 +299,16 @@ def convertor(linefilter, arglist, exclude):
                 print("wmlindent: no WML file found, exiting", file=sys.stderr)
                 sys.exit(1)
 
+def sigint_handler(signal, frame):
+    """This function defines what happens when the SIGINT signal is encountered by pressing ctrl-c during runtime. 
+    When ctrl-c is pressed, a one-line message is displayed and Python exits with Status 0, which refers to successful termination.
+    This overrides Python's default behavior of displaying a traceback when ctrl-c is pressed.
+    """
+    print ('Aborted by pressing ctrl-c')
+    sys.exit(0)
+
 if __name__ == '__main__':
+    signal.signal(signal.SIGINT, sigint_handler)
     parser = argparse.ArgumentParser(
         formatter_class=argparse.RawTextHelpFormatter
         )
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -3504,6 +3504,6 @@ In your case, your system interprets your arguments as:
         if failed_any_dirs:
             sys.exit(1)
     except KeyboardInterrupt:
-        print("Aborted")
+        print("Aborted by pressing ctrl-c")
 
 # wmllint ends here
--- a/data/tools/wmlscope
+++ b/data/tools/wmlscope
@@ -552,6 +552,6 @@ directories are given, all files under the current directory are checked.""")
                 xref.unresdump()
                 xref.duplicates(exportonly=True)
     except KeyboardInterrupt:
-        print("wmlscope: aborted.", file=sys.stderr)
+        print("Aborted by pressing ctrl-c", file=sys.stderr)
 
 # wmlscope ends here
--- a/data/tools/wmlxgettext
+++ b/data/tools/wmlxgettext
@@ -34,6 +34,7 @@
 import os
 import re
 import sys
+import signal
 import warnings
 import argparse
 from datetime import datetime
@@ -162,6 +163,7 @@ def commandline(args):
 
 
 def main():
+    signal.signal(signal.SIGINT, sigint_handler)
     args = commandline(sys.argv[1:])
     pywmlx.ansi_setEnabled(args.text_col)
     pywmlx.wincol_setEnabled(args.text_col)
@@ -261,6 +263,16 @@ def main():
         outfile.close()
     if args.debugmode:
         fdebug.close()
+
+
+
+def sigint_handler(signal, frame):
+    """This function defines what happens when the SIGINT signal is encountered by pressing ctrl-c during runtime. 
+    When ctrl-c is pressed, a one-line message is displayed and Python exits with Status 0, which refers to successful termination.
+    This overrides Python's default behavior of displaying a traceback when ctrl-c is pressed.
+    """
+    print ('Aborted by pressing ctrl-c')
+    sys.exit(0)
 
 
 


data/tools/wmlindent
```diff
@@ -299,7 +299,16 @@ def convertor(linefilter, arglist, exclude):
                 print("wmlindent: no WML file found, exiting", file=sys.stderr)
                 sys.exit(1)
 
+def sigint_handler(signal, frame):
+    """This function defines what happens when the SIGINT signal is encountered by pressing ctrl-c during runtime. 
+    When ctrl-c is pressed, a one-line message is displayed and Python exits with Status 0, which refers to successful termination.
+    This overrides Python's default behavior of displaying a traceback when ctrl-c is pressed.
+    """
+    print ('Aborted by pressing ctrl-c')
```
knyghtmare, line 307: Is this os independent? Normally, I have to do cmd+c for the same thing on macos.

max-torch: It's the same message on all OS as of latest commit in the PR. Would it suffice if i change the message to say 
>Aborted using Keyboard Interrupt

?

cooljeanius: > Is this os independent? Normally, I have to do cmd+c for the same thing on macos.

Cmd-C is only the equivalent of ctrl-c when other OSes are using ctrl-c to mean "copy"; it's still ctrl-c when it's used to mean "kill this process running in a terminal" though

CelticMinstrel: > Is this os independent? Normally, I have to do cmd+c for the same thing on macos.

???

Ctrl+C for interrupt is still Ctrl+C on Mac (although Cmd+period also works).

Elvish-Hunter:
My main worry with this PR was that the `signal` module might not have worked correctly on Windows. Luckily this is not the case, so I'm merging this PR and cherry-picking to 1.16 as well.

data/tools/wmlindent
soliton-: Still a bit odd to reference ctrl-c specifically since it is just one way to send an INT signal.

data/tools/wmlindent
```diff
@@ -299,7 +299,16 @@ def convertor(linefilter, arglist, exclude):
...
+    sys.exit(0)
```
soliton-, line 308: Exiting with 0 seems very questionable.

data/tools/wmlindent
max-torch: What's the other way to send INT?

data/tools/wmlindent
max-torch: What is the Exit Status Number to use?

data/tools/wmlindent
soliton-: You use kill to send signals. In general the syscall or on the command line there is a kill command and your shell probably has it as a builtin as well.

data/tools/wmlindent
soliton-: At least something that is not 0. Ideally the process should kill itself and not just exit so that a shell like bash will handle the case of pressing ctrl-c on a terminal correctly. Further reading: https://www.cons.org/cracauer/sigint.html

cooljeanius: I dunno about Python, but in C at least there's macros you're supposed to use for the argument to `exit()`, rather than the raw numbers

soliton-: This has nothing to do with using signal numbers or names. The point is that terminating (probably in the middle of a run) on a signal should not count as a success (i.e. exit status 0). The exit status should indicate failure and ideally it should also indicate that the termination is because of the signal which cannot be done at all purely with the exit status, hence the process killing itself.

data/tools/wmlindent
max-torch: If you're talking about `kill` command that could even be sent without using keyboard interrupt. In that case does "Runtime Interrupted by User" sound better?

data/tools/wmlindent
max-torch: I used 0 because I thought it counted as a success since it was what the user wanted. I can make this exit status 1.

soliton-: There is nothing fancy about calling an external command from an interactive shell or a script. That is pretty much the main thing you do. wmlindent for example is called from a CI script on every commit. If someone now runs the CI script themselves and wants to abort with ctrl-c when wmlindent happens to be running then nothing will abort except wmlindent. They'll have to press ctrl-c again and if more commands behave this way again and again...

Exit status 0 meaning success and anything else failure is the same on Windows. It's a pretty universal convention, not just for UNIX.

max-torch: How do I run the CI script? I want to compare behavior of aborting the CI script before and after this PR.

max-torch:
I was talking with ChatGPT about [#3352](https://github.com/wesnoth/wesnoth/issues/3352), and it gave an interesting answer that is different than what I, the PR author, had implemented here:

>Yes, it is possible to suppress the traceback when interrupting a Python script using SIGINT (which is the signal that is sent to a process when you press CTRL+C on the command line). To do this, you can use the signal module to register a signal handler that catches the SIGINT signal and suppresses the traceback.
>
>Here's an example of how you might do this:

```python
import signal

def handle_sigint(signum, frame):
    # Do nothing, just suppress the traceback
    pass

signal.signal(signal.SIGINT, handle_sigint)

# Main part of the script goes here
```
>This will register the handle_sigint() function as a signal handler for the SIGINT signal. When the script receives the SIGINT signal (i.e., when you press CTRL+C), the signal handler will be called and will suppress the traceback.
>
>It's worth noting that this approach will not prevent the script from being interrupted when CTRL+C is pressed. It will simply suppress the traceback that is normally displayed when the KeyboardInterrupt exception is raised.

I am aware that responses of ChatGPT are not guaranteed to be correct, but I have found that a large number of its answers are, in fact, correct upon verification. I still believe all responses it gives should be verified.

If this solution of ChatGPT is correct, it may be a more elegant solution than what I had provided in this PR and it even reduces the headache of wondering what the appropriate Exit Status Number to use for `sys.exit()`

Elvish-Hunter:
> If this solution of ChatGPT is correct, it may be a more elegant solution than what I had provided in this PR

Removing the *'Aborted by pressing ctrl-c'* message doesn't sound like a good idea. Also, that chunk of code looks like it'll just disable `Ctrl + c` instead of merely suppressing the output.

> it even reduces the headache of wondering what the appropriate Exit Status Number to use for sys.exit().

Closing a program with `Ctrl + c` is always considered an abnormal termination, so it should always return an exit code different from 0. The page at https://tldp.org/LDP/abs/html/exitcodes.html even suggest for this specific case to set the exit code to 130...

PR #6247, opened 2021-11-01
Fixes for broken gui.get_user_choice() function

Elvish-Hunter:
While updating the Wesnoth Lua Pack for the 1.16 series, I noticed that the `gui.get_user_choice()` function is unusable: all it does is throwing the following error.

    error scripting/lua: lua/core/gui.lua:38: bad argument #2 to 'insert' (number expected, got function)

This is caused by three commas that shouldn't be there, so the first commit of this PR removes them.
After having fixed this, I checked the function with the test scenario: while it was supposed to show an image column and two text columns, it still used the old syntax (the one with `&` and `=`, to be clear), instead of the current `image`, `label`, `description` and `default` keys used by `[option]` tags. Since support for the old syntax has been removed, the text shown to the user was just like this:

    *&items/potion-blue.png=Nice blue=Surely you’ll like that one

The second and third commit of this PR add support for the currently used keys (which can be in a table or vconfig object) and update the test scenario, respectively.

### DIFF
--- a/data/lua/core/gui.lua
+++ b/data/lua/core/gui.lua
@@ -35,10 +35,20 @@ function gui.get_user_choice(attr, options)
 		msg[k] = attr[k]
 	end
 	for k,v in ipairs(options) do
-		table.insert(msg, wml.tag.option { message = v,
-			wml.tag.command { wml.tag.lua {
-				code = string.format("gui.__user_choice_helper(%d)", k)
-			}}})
+		if type(v) == "table" or type(v) == "userdata" then
+			table.insert(msg, wml.tag.option { image = v.image,
+				label = v.label,
+				description = v.description,
+				default = v.default,
+				wml.tag.command { wml.tag.lua {
+					code = string.format("gui.__user_choice_helper(%d)", k)
+				}}})
+		elseif type(v) == "string" then
+			table.insert(msg, wml.tag.option { message = v,
+				wml.tag.command { wml.tag.lua {
+					code = string.format("gui.__user_choice_helper(%d)", k)
+				}}})
+		end
 	end
 	wesnoth.wml_actions.message(msg)
 	gui.__user_choice_helper = nil
--- a/data/lua/core/gui.lua
+++ b/data/lua/core/gui.lua
@@ -35,8 +35,8 @@ function gui.get_user_choice(attr, options)
 		msg[k] = attr[k]
 	end
 	for k,v in ipairs(options) do
-		table.insert(msg, wml.tag.option, { message = v,
-			wml.tag.command, { wml.tag.lua, {
+		table.insert(msg, wml.tag.option { message = v,
+			wml.tag.command { wml.tag.lua {
 				code = string.format("gui.__user_choice_helper(%d)", k)
 			}}})
 	end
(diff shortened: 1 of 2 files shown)


CelticMinstrel (approved):
This all looks good.

stevecotton:
There's a deprecated `helper.get_user_choice` that passes through to this one. If the new implementation doesn't accept the old parameters, that suggests the deprecation helper is still broken.

CelticMinstrel:
For master I'd just remove the deprecated one. If it's backported, ideally you'd make it work somehow.

Elvish-Hunter:
Since version 1.16.1 is being released tomorrow and I haven't had much time this week to do more work on it, for now I've just backported 57b2d71 there (as 603430c).

Wedge009:
Been a month since this was last updated and it looks like something was fixed for 1.16. Any reason why this is still not merged to master despite being approved?

Pentarctagon:
It sounds like the deprecation helper is still broken?

Wedge009:
Oh okay. Was thrown off by celmin's earlier approval.

Elvish-Hunter:
I decided to close this PR because I just pushed to master 5b18f2d (which includes 57b2d71 and improves 855b3a7 by covering every possible Lua data type) and 38855ba (which is similar to 93b2977).
About `helper.get_user_choice()`, this is just a thin wrapper around `gui.get_user_choice()`. The old DescriptionWML syntax (the one which used asterisks, equal signs and ampersands) was already deprecated in the 1.14 series: to support it, I'd need to re-implement in Lua something that we already removed from the C++ engine. It doesn't seem a good idea to "fix" this one.
@Pentarctagon considering the feature freeze, should I backport 5b18f2d to 1.16?

Pentarctagon:
In terms of UMC, what would backporting it affect?

stevecotton:
Note: 5b18f2d needed a whitespace fixup to pass the CI.

Elvish-Hunter:
> In terms of UMC, what would backporting it affect?

In 1.16.0, the function was completely broken.
In 1.16.1, it only supports data types that can be cast to strings, which means there's no way to use the image column and the second text column (this requires using a table or a WML table, since the old syntax has been removed from the engine), so it can still be considered broken.
In 1.16.2, it'll finally work properly; of course, UMC authors should use `wesnoth.current_version()` to check that their add-on is being run at least on Wesnoth 1.16.2. It's also interesting to note that `gui.get_user_choice()` isn't currently documented in the LuaAPI/gui page: as a direct consequence of this, a grep on my local mirror of the 1.16 server shows that there are only three add-ons using it, which are 1The_Great_Steppe_Era, Era_of_High_Sorcery (both as `helper.get_user_choice()`) and Wesnoth_Lua_Pack (this one is under my direct control anyway).

Pentarctagon:
I'd say it's fine to backport, in that case.

Elvish-Hunter:
Backported as b643d25 and 77e4a79.

Elvish-Hunter:
I also documented it in the LuaAPI/gui wiki page: https://wiki.wesnoth.org/LuaAPI/gui#gui.get_user_choice.

PR #6248, opened 2021-11-01
Add translation hints for time format strings

### LINKED ISSUE #6246: strftime po hints
These strings (and probably a couple that I've missed) should have translation hints:

```
src/format_time_summary.cpp-			// save is from current year
src/format_time_summary.cpp:			format_string = _("%b %d");
--
src/format_time_summary.cpp-		// save is from a different year
src/format_time_summary.cpp:		format_string = _("%b %d %Y");
--
src/save_index.cpp-	if(std::tm* tm_l = std::localtime(&modified())) {
src/save_index.cpp-		const std::string format = preferences::use_twelve_hour_clock_format()
src/save_index.cpp:			? _("%a %b %d %I:%M %p %Y")
src/save_index.cpp:			: _("%a %b %d %H:%M %Y");
```

This one should be translatable too:
```
src/gui/dialogs/addon/manager.cpp-
src/gui/dialogs/addon/manager.cpp-		const char* format = preferences::use_twelve_hour_clock_format()
src/gui/dialogs/addon/manager.cpp:			? "%B %d %Y, %I:%M %p"
src/gui/dialogs/addon/manager.cpp:			: "%B %d %Y, %H:%M";
```

Wedge009:
Resolves #6246

### DIFF
--- a/src/format_time_summary.cpp
+++ b/src/format_time_summary.cpp
@@ -44,25 +44,31 @@ std::string format_time_summary(std::time_t t) {
 		if(days_apart == 0) {
 			// save is from today
 			if(preferences::use_twelve_hour_clock_format() == false) {
+				// TRANSLATORS: 24-hour time, eg '13:59'
 				format_string = _("%H:%M");
 			}
 			else {
+				// TRANSLATORS: 12-hour time, eg '1:59 PM'
 				format_string = _("%I:%M %p");
 			}
 		} else if(days_apart > 0 && days_apart <= current_time.tm_wday) {
 			// save is from this week
 			if(preferences::use_twelve_hour_clock_format() == false) {
+				// TRANSLATORS: Day of week + 24-hour time, eg 'Sunday, 13:59'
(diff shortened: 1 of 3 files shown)


stevecotton:
It looks good, but it doesn't backport cleanly because of the other changes between 1.16 and master. Not sure what to do about that, the best option might be to merge it and make a hand-edited backport for 1.16.2.

Wedge009:
No worries, when the string freeze lifts I can do the same/similar thing for 1.16.

stevecotton (approved):

stevecotton:
Just realised that most of this doesn't need to wait for the string freeze to be backported. With the exception of the one that's converted to translatable strings, all of the other po hints could go into 1.16.1 and help any translation team that's working before 1.16.2 comes out.

Wedge009:
Oh, good point.

Wedge009:
92e8e1cd86b0e57c2d5f4e51b55a538c8ca419f4

PR #6251, opened 2021-11-03
Check for potential duplicates before inserting.

Pentarctagon:
---

I had already fixed this issue on the client side a while ago, but perhaps WCII has another way of triggering it.

### DIFF
--- a/src/server/wesnothd/server.cpp
+++ b/src/server/wesnothd/server.cpp
@@ -1676,9 +1676,19 @@ void server::handle_player_in_game(player_iterator p, simple_wml::document& data
 			const simple_wml::node& m = *g.level().root().child("multiplayer");
 			DBG_SERVER << simple_wml::node_to_string(m) << '\n';
 			// [addon] info handling
+			std::set<std::string> uniqueness_checker;
 			for(const auto& addon : m.children("addon")) {
 				for(const auto& content : addon->children("content")) {
-					user_handler_->db_insert_game_content_info(uuid_, g.db_id(), content->attr("type").to_string(), content->attr("name").to_string(), content->attr("id").to_string(), addon->attr("id").to_string(), addon->attr("version").to_string());
+					const std::string type = content->attr("type").to_string();
+					const std::string content_id = content->attr("id").to_string();
+					// the UUID is identical for the entirety of the server instance and the game ID is identical for all rows for this game
+					// therefore only need to be checking against the two values that would change with each row
+					if(uniqueness_checker.count(type+content_id) == 0) {
+						uniqueness_checker.insert(type+content_id);
+						user_handler_->db_insert_game_content_info(uuid_, g.db_id(), content->attr("type").to_string(), content->attr("name").to_string(), content->attr("id").to_string(), addon->attr("id").to_string(), addon->attr("version").to_string());
+					} else {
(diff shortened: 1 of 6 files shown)


src/server/wesnothd/server.cpp
```diff
@@ -1676,9 +1676,19 @@ void server::handle_player_in_game(player_iterator p, simple_wml::document& data
...
 			for(const auto& addon : m.children("addon")) {
 				for(const auto& content : addon->children("content")) {
-					user_handler_->db_insert_game_content_info(uuid_, g.db_id(), content->attr("type").to_string(), content->attr("name").to_string(), content->attr("id").to_string(), addon->attr("id").to_string(), addon->attr("version").to_string());
+					const std::string type = content->attr("type").to_string();
+					const std::string content_id = content->attr("id").to_string();
+					// the UUID is identical for the entirety of the server instance and the game ID is identical for all rows for this game
+					// therefore only need to be checking against the two values that would change with each row
+					if(uniqueness_checker.count(type+content_id) == 0) {
+						uniqueness_checker.insert(type+content_id);
+						user_handler_->db_insert_game_content_info(uuid_, g.db_id(), content->attr("type").to_string(), content->attr("name").to_string(), content->attr("id").to_string(), addon->attr("id").to_string(), addon->attr("version").to_string());
```
soliton-, line 1688: I'd use those new variables here now as well.

src/server/wesnothd/server.cpp
```diff
@@ -1676,9 +1676,19 @@ void server::handle_player_in_game(player_iterator p, simple_wml::document& data
...
 			// [addon] info handling
+			std::set<std::string> uniqueness_checker;
 			for(const auto& addon : m.children("addon")) {
 				for(const auto& content : addon->children("content")) {
-					user_handler_->db_insert_game_content_info(uuid_, g.db_id(), content->attr("type").to_string(), content->attr("name").to_string(), content->attr("id").to_string(), addon->attr("id").to_string(), addon->attr("version").to_string());
+					const std::string type = content->attr("type").to_string();
+					const std::string content_id = content->attr("id").to_string();
+					// the UUID is identical for the entirety of the server instance and the game ID is identical for all rows for this game
+					// therefore only need to be checking against the two values that would change with each row
+					if(uniqueness_checker.count(type+content_id) == 0) {
```
soliton-, line 1686: I think you could also just insert and check the return value. No big need to optimize this though...

soliton- (approved):
Looks good!

Pentarctagon:
@soliton- I updated it to check the returned number of modified rows, since that is a simpler approach. Also in doing so I noticed that the documentation of the `dbconn::modify` method was incorrect - it said it returned the number of rows modified, however it was actually returning the ID generated from an AUTO_INCREMENT column (or 0 if there was none).

As a result I have also renamed the existing `dbconn::modify` method to `dbconn::modify_get_gid` and changed the `dbconn::modify` method method to use mariadbpp's execute() method rather than insert().  This won't affect anything else since both insert() and execute() have the exact same implementation, differing only in the value they return.

soliton-:
Is `gid` for generated ID? Sounds too much like group ID to me. I'd rather just use `modify_get_id`.

Pentarctagon:
It is, and it's been updated.

soliton-:
Do you intend more changes? From my side this is good to go in.

Pentarctagon:
Ah, I was waiting for you to say if you were good with it; I suppose I should've re-requested approval on the PR.  I'll go ahead and merge/backport this in a bit then.

PR #6254, opened 2021-11-03
Convert Isar's Cross map background images to webp format

inferno8:
I've reduced the size of Isar's Cross background images up to 77% (from 5.18Mb to 1.16Mb) by using the WebP image format instead of PNG. I used lossy compression with quality setting of 90, so there is no degradation in perceptible visual quality. 

Does Wesnoth support such format on all platforms? I am running BfW on Windows 10 and it works fine for me (the map background graphics are visible), but I am not sure about other OSes.


Pentarctagon:
Is there a reason to use webp rather than jpg?

inferno8:
Jpg does not support transparency. WebP offers file sizes that are around 30% smaller than JPEG without a quality gap. It also provides transparency (alpha channel) like PNG.

irydacea:
Wesnoth supports everything SDL_image supports. The issue is that we currently do not **guarantee** that webp support is available since only PNG and JPEG support are tested by scons and cmake and mainline does not use any other formats in any highly-visible places either.

Pentarctagon:
It looks like the Steam Linux runtime likely wouldn't support webp, given its [list of packages](https://repo.steampowered.com/steamrt-images-scout/snapshots/0.20211102.0/com.valvesoftware.SteamRuntime.Platform-amd64,i386-scout.manifest.dpkg) doesn't have `libwebp`.

Wedge009:
I also generally don't like the idea of using lossy compression - unless someone keeps the lossless source somewhere (yes, I know there are JPGs in use). If we are already using lossless images, I'd prefer to keep it that way unless such images are unreasonably large in file size relative to alternatives.

inferno8:
Ok, I see. So the best option for images with transparency at the moment is to stick with PNGs (and maybe TIFFs, although the former is guaranteed to work on all platforms). I suppose the alternative solution to reduce the file size in that case would be to use 2 separate files - one for a main image and the second one for a dedicated alpha mask and then use Image Path Functions to render the combined image properly. I wonder if IPF can handle huge images like these...

CelticMinstrel:
> I wonder if IPF can handle huge images like these...

I don't know for certain, but I suspect it would be very inefficient to use an IPF for that.

There may be another options that doesn't require transparency, however. If I understand correctly, these images are placed as terrain graphics, and the transparency is essentially a "hole" through which the main portion of the map shows. Would it be possible to place the image in a lower layer than all other terrain graphics? Or do parts of it _need_ to render above the terrain? Even if they do need to though, it may be possible to split it into a main background portion with no transparency that renders underneath, plus several smaller pieces with transparency that render above the main terrain.

inferno8:
> Would it be possible to place the image in a lower layer than all other terrain graphics?

I am afraid not. Here's an example of a fully opaque jpg background (I replaced transparent pixels with white color for testing). 
isar-layer-test
The layer parameter has been set to -1000. Grass and other basic tiles are no longer visible. Setting the layer to -1001 completely disables rendering of the background (I guess all basic tiles are rendered on the layer -1000, even chasms, so they are 'on top' of the background in that particular case).

To sum up, using jpgs to render the image below other terrain graphics is, in my opinion, currently impossible. Without transparency the now visible portion of the image, where the "hole" once was, is rendered on top of basic tiles like grass, sand and chasm effectively covering them, which is really bad.

Btw, using IPF for masking causes serious lag, so CelticMinstrel was right.

Theoretically there is one other option - to reduce the color palette, but this will surely degrade the quality of these graphics, so I am not happy about it.

I wish the Steam linux runtime supported WebP, but it is not surprising that it doesn't - the format is relatively new and still not so common.

I am afraid these PNG backgrounds will have to stay as they are for the time being, unless someone comes up with better solution on how to deal with them in the future.

Pentarctagon:
Looking at the Windows installer, it's pulling in libwebp already automatically, so that's probably why it works on Windows.  Most linux distros also have libwebp available by now.  So it'd be a question of:
* Updating the vcpkg dependencies to have libwebp.
* Updating the MacCompileStuff to have libwebp.
* Either getting libwebp into the Scout runtime used by Steam Linux or providing it ourselves.

irydacea:
Worth keeping in mind that SDL_image also needs to have been compiled to use libwebp. Just dropping libwebp.so into a folder isn't good enough if SDL_image has no idea it's supposed to use it.

irydacea:
For the record, I would support this move as a long term goal (depending on how much we need to bribe Valve to get support into the Steam runtime). On my own add-ons it seems like using **lossless** WebP for unit portraits has the potential to yield quite substantial data savings around 20%:

```
15:20:31 shadowm@hanacore ~/src/projects/Naia/images git:master % du -b portraits/* portraits.webp/* 
239212  portraits/dark-knight.png
214870  portraits/demon-regular+female.png
196620  portraits/demon-regular.png
196200  portraits/elynia.png
166330  portraits/elyssa.png
171346  portraits/galas.png
79679   portraits/giant-ant.png
119620  portraits/igor.png
316804  portraits/mal-hekuba.png
208229  portraits/mal-keshar.png
182230  portraits.webp/dark-knight.webp
167412  portraits.webp/demon-regular+female.webp
143646  portraits.webp/demon-regular.webp
155022  portraits.webp/elynia.webp
118744  portraits.webp/elyssa.webp
124684  portraits.webp/galas.webp
63652   portraits.webp/giant-ant.webp
92904   portraits.webp/igor.webp
215426  portraits.webp/mal-hekuba.webp
162860  portraits.webp/mal-keshar.webp
```

Also, it's true that WebP is relatively new and lossy graphics are more inconvenient to edit for those who absolutely need to edit every single thing in existence, but hear me out:

* We have a dedicated repository for lossless versions of assets that normally don't need to be shipped with the game, in the form of [https://github.com/wesnoth/resources]. The gargantuan SVG source of the Wesnoth branding lives there among other things.
* Lossy WebP use on the web has become increasingly more common especially among large platforms because all modern web browsers have supported it for years. Granted, they do have an option that Wesnoth doesn't have which is the ability to fallback to JPEG when the user's browser is known to be unable to render WebP or fails to do so.
* Lossless WebP is an option that still yields big savings as seen above. I'd expect lossless (or very high-quality lossy) WebP to be used instead of PNG for portraits, while using (lower-quality) lossy WebP in JPEG's stead in story images and such.

Pentarctagon:
I can open an issue over at https://github.com/ValveSoftware/steam-runtime/issues asking about the possibility of webp support sometime today or tomorrow then, in that case.

Wedge009:
I didn't know webp had lossless options too - in that case I'd be in favour of using it where it makes sense and the dependency issues can be resolved. It looks like in the same way webm is an open alternative to MPEG, webp is an alternative to JPEG.

inferno8:
@Pentarctagon: thanks!

I ran some tests yesterday and it looks like WebP is extremely effective for portraits. Even with lossless compression it managed to cut the size of the entire folder by 1/3 (!). But lossy compression is where it really shines. Setting the output quality to 95% (which is a very high value) it reached astonishing... 72% size reduction. By reducing the quality further down to 90% you can get additional 5% reduction. And it still looks better than most jpgs of the same quality setting.

Here's a chart with the results of my tests:
webp-portraits-chart

As you can see, the WebP format can be used to effectively reduce the size of the game's images, thus greatly decreasing the download size for players (which is always welcome).

BfW has become quite 'bulky' in recent years (more than 0,5GB - typical 2D games are much more lightweight). That's why I highly recommend implementing solutions like these in the future, especially that there is a high chance everyone will appreciate this.

Pentarctagon:
I'm wondering now whether the same ~1/3 size reduction would also be true for smaller images like terrain, units, etc seeing as `data/core/images/` is 123.4 MBs currently.

CelticMinstrel:
This could cause problems if people are using IPFs to recolour portraits though… does anyone do that?

inferno8:
I analyzed terrain and units folders. Here are the results:

webp-terrain-chart
Using lossless webp for terrain tiles one can gain 28% more space (-10.3MB). Lossy compression affects colors, so it may visually break transitions (for example each grass tile variation may have a slightly different hue or saturation, not matching borders of other tiles). So in case of terrain images, I'd go with lossless webp to avoid any kind of degradation.

When converting unit images I used lossless compression only for obvious reasons (we are dealing with a strict TC palette here).
webp-units-chart
The output images were smaller by almost 37% which is more than 1/3 (we get additional 3.2MB of free space). This would probably apply to other folders too. 

I estimate at least 20% size reduction of the entire images folder when using lossless for everything.

irydacea:
> This could cause problems if people are using IPFs to recolour portraits though… does anyone do that?

I would not feel an ounce of remorse for breaking that kind of use case because mainline  portrait art was never made with such a thing in mind, in the artistic (not technical) sense.

irydacea:
> When converting unit images I used lossless compression only for obvious reasons (we are dealing with a strict TC palette here).

Regardless of the team colouring algorithm in use, all pixel art must always remain lossless because it's the only way to preserve its fidelity. Unlike portrait and story art it *is* made with each pixel's final rendering result in mind.

CelticMinstrel:
> mainline portrait art was never made with such a thing in mind, in the artistic (not technical) sense.

True, true. If anyone does it, or if we think anyone might do it, I'd at least suggest mentioning that it can't be relied on anymore.

Pentarctagon:
Webp support is implemented in the steam-runtime for Linux and it should just be a matter of time until it's generally available: https://github.com/ValveSoftware/steam-runtime/issues/475#issuecomment-989973960

I tested it out using a webp version of the Arch Mage sprite, following the instructions for using the Soldier runtime with webp support currently added, and it seems to work - with the only webp version present the Arch Mage's sprite was still displayed as normal in-game.

The webp image file size actually increased to 10 KB vs the current png file's 1.6 KB, but that was just from me exporting the png to webp from Gimp.

Pentarctagon:
From the OP:
> I used lossy compression with quality setting of 90, so there is no degradation in perceptible visual quality.

Wedge009:
So it's lossy. 'Perceptible' is subjective.

Pentarctagon:
So to update the list from earlier:
* Updating the vcpkg dependencies to have libwebp - This can be done fairly trivially, so I plan to do it myself in the near future.
* Updating the MacCompileStuff scripts to build libwebp - @hrubymar10 can MacCompileStuff be updated to include libwebp as well as build SDL2_image with webp support? Also the xcode projectfiles.
* Either getting libwebp into the Scout runtime used by Steam Linux or providing it ourselves - most recent comment is that it'll be available mid-January at the earliest.
* Check that the Flathub version has the correct libraries - @loonycyborg Do you know how to check whether the Flathub runtime has libwebp as well as webp support enabled in SDL2_image?

---

Other TODOs:
* Decide which images should be converted to webp, actually convert them, and then update `wmllint` to handle the image file renaming.
* Update scons and cmake to handle libwebp. @newfrenchy83 I assume the CodeBlocks projectfiles will need to be updated for this as well.

newfrenchy83:
> So to update the list from earlier:
> 
>     * Updating the vcpkg dependencies to have libwebp - This can be done fairly trivially, so I plan to do it myself in the near future.
> 
>     * Updating the MacCompileStuff scripts to build libwebp - @hrubymar10 can MacCompileStuff be updated to include libwebp as well as build SDL2_image with webp support? Also the xcode projectfiles.
> 
>     * Either getting libwebp into the Scout runtime used by Steam Linux or providing it ourselves - most recent comment is that it'll be available mid-January at the earliest.
> 
>     * Check that the Flathub version has the correct libraries - @loonycyborg Do you know how to check whether the Flathub runtime has libwebp as well as webp support enabled in SDL2_image?
> 
> 
> Other TODOs:
> 
>     * Decide which images should be converted to webp, actually convert them, and then update `wmllint` to handle the image file renaming.
> 
>     * Update scons and cmake to handle libwebp. @newfrenchy83 I assume the CodeBlocks projectfiles will need to be updated for this as well.

For compilation with gcc, the precompilded libwebp libraries exist in .lib only. Change the extension to .a is authorize the compile with mingw or must be use the linux precompiled libraries?

Pentarctagon:
I don't follow - what is your question about?

newfrenchy83:
> I don't follow - what is your question about?

forget this, i already had this problem before, i must refind solution by myself, excuse me.

Wedge009:
Just had Steam client updates propagated, notes seem to indicate webp support has been added: https://store.steampowered.com/oldnews/102669

Pentarctagon:
Is that the regular client or the beta client?

Wedge009:
I assume regular - I don't subscribe to anything beta on Steam.

Pentarctagon:
Huh, well that would be pretty fast then.  After I push #6385 then I'll push an updated steam runtime image to Wesnoth's docker hub, and then open a PR with the webp support enabled. If the CI job for the steam runtime passes there, then it will confirm the webp support for steam-linux is generally available.

loonycyborg:
Wrt flatpak current freedesktop runtime does have webp library installed.

Pentarctagon:
I believe everything is now ready with webp support. So I think this can be merged now, aside from the changelog file conflict.

Issue #6096, opened 2021-09-17
Hovering with the cursor over certain unit attributes does not display explanation box when window is maximized or in fullscreen


IceTyp:
### Game and System Information

 - Wesnoth 1.15.17
 - Steam
 - Ubuntu 20.04

### Describe the bug

When hovering with the cursor over certain unit attributes, the explanation box does not appear. This seems to happen only if the window is **maximized** or in **fullscreen**.
Note that the cursor is not displayed in the screenshots.

The following window is **not maximized**. When hovering over _nightstalk_ the explanation box appears (as it should):

![Hovering over ability not maximized window](https://user-images.githubusercontent.com/90767623/133762992-0e65103f-b7c9-4790-a49e-92bec5d8cfb8.png)

In contrary, the next window is **maximized**. The explanation box does not appear when hovering over _nightstalk_:

![Hovering over ability maximized window](https://user-images.githubusercontent.com/90767623/133762648-0362333c-017a-4b38-b168-48c470cb72b4.png)

Here, the window is in **fullscreen**. The explanation box does not appear when hovering over _nightstalk_:

![Hovering over ability fullscreen](https://user-images.githubusercontent.com/90767623/133765299-c618d43a-2c16-4701-8f5e-060368ab45a3.png)

### Other unit attributes

The explanation box of the following elements in the above picture is not displayed when hovering over them (when the window is maximized or in fullscreen):
- _Shadow_
- _nightstalk_
- _skirmisher_
- _Undead_ (Race)

In the above picture, the following elements are not affected by this bug:
- _chaotic (+0%)_
- _undead_ (Trait)
- _8x3 claws_
- _backstab_

Note that there is a similar behavior with other units.

### Further explanations

I have also installed Wesnoth 1.14.14 on the same system (with flatpak). There, this bug does not occur, that is, the explanation box is displayed when hovering over any of the listed attributes (even when the window is maximized or in fullscreen).

Wedge009:
@stevecotton and I noticed something similar while discussing #6007. Are you sure you only observe this with a maximised window?

IceTyp:
For me it only happens when the window is maximized or in fullscreen.
It does not happen when it is left aligned, right aligned or something else.

Pentarctagon:
It's definitely related to the window size somehow. I started a Sulla's Ruins game with a Lieutenant:
* 1280x720 - both Lawful and Leadership tooltips appear on hover.
[...]

PR #6256, opened 2021-11-04
[1.16] Fix tooltips on the sidebar

### LINKED ISSUE #6007: [1.16] The unit panel tooltip is not shown on names and descriptions
 - 1.15.15
 - The official site/SourceForge
 - Not a custom build
 - What OS are you running?
 - Windows 10

### Describe the bug
A tooltip doesn't show up when mouse hovers over a long unit name on the right in-scenario panel. It makes long names impossible to read unless they show up in a dialog which doesn't bound to happen.

### To Reproduce

1. First, load a scenario.
2. Then, find or spawn a unit with a long name that doesn't fit the right panel.
3. Lastly, hover a mouse over the top entry that is the name in bold white characters.
4. Observe that there is no tooltip.

### Expected behavior
A full-name tooltip used to be shown in 1.14.

### Screenshots
The localized fort guard name is cropped by the right edge and a tooptip with the full information is not available although there is the cursor over it.

no_tooltip

### Additional context
The fix will probably bear some regression as the behaviour was correct in 1.14.

stevecotton:
Fix iterator-use-after-invalidation in tooltips
---

It's undefined behavior to access an iterator after calling std::map::erase(),
the correct method is to use the new iterator that erase() returns instead.

Fix pango_text::set_maximum_height
---

Change back to handling the height in our code instead of expecting Pango to
handle it. Pango is still expected to handle the maximum width, as Pango needs
to handle the word-wrapping.

While writing f282eb7, I had noticed that set_maximum_height called
PangoLayout's pango_layout_set_height, and removed the apparently unnecessary
enforcement of the maximum height from our wrapper code. The documentation
added in this commit explains why that was incorrect.

Fixes #6007 and #6096, both issues that some tooltips on the sidebar stopped
working. The text on that sidebar is densely packed, and while this commit
makes no visual change, it removes a few blank pixels of padding below the
text. That extra padding was sufficient for the next line's area to overlap,
making tooltips::add_tooltip remove the previously-added tooltip.

### DIFF
--- a/src/font/text.cpp
+++ b/src/font/text.cpp
@@ -398,6 +398,11 @@ pango_text& pango_text::set_maximum_height(int height, bool multiline)
 	if(height != maximum_height_) {
 		// assert(context_);
 
+		// The maximum height is handled in this class' calculate_size() method.
+		//
+		// Although we also pass it to PangoLayout if multiline is true, the documentation of pango_layout_set_height
+		// makes me wonder whether we should avoid that function completely. For example, "at least one line is included
+		// in each paragraph regardless" and "may be changed in future, file a bug if you rely on the current behavior".
 		pango_layout_set_height(layout_.get(), !multiline ? -1 : height * PANGO_SCALE);
 		maximum_height_ = height;
 		calculation_dirty_ = true;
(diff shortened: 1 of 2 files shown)


Pentarctagon:
Was it tested that there are no issues with the sidebar tooltips at a variety of resolutions?

stevecotton:
Have now found an issue going down to 800x600. At that resolution, the side-flag is eclipsed (even if the unit has a shorter name), and the tooltips for both the full name and the flag are missing; the movement stats and defense stats (on the line above the flag) can sometimes also be missing at 800x600.
image

Having a width of 830 or more seems to make the movement and defense stats reliable.
Having a height of 750 or more seems to make the full name reliable.
Having a height of 750 or more seems to make the flag's ellipse reliable. When the flag is ellipsed, the tooltip seems to only be on the ellipse itself, hovering over the flag image doesn't work.

I think this PR is still a huge improvement.

Pentarctagon (approved):

PR #6259, opened 2021-11-05
Added ability to add setter callbacks on preferences

Vultraz:
A simple callback interface for pref setters required by @irydacea. This could honestly be used to replace functions like `set_music`, allowing one to simply call `set` and having the callback handle the side effects.

### DIFF
--- a/src/preferences/general.cpp
+++ b/src/preferences/general.cpp
@@ -53,7 +53,7 @@ bool no_preferences_save = false;
 bool fps = false;
 
 config prefs;
-std::map<std::string, std::function<void(const config::attribute_value&)>> callbacks {};
+std::map<std::string, preferences::setter_callback> callbacks {};
 }
 
 namespace preferences {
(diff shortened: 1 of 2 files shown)


irydacea:
It's kind of your responsibility to prescribe where the new API should be used (hi Doxygen). Also, pasting what I said on Discord here:

> sure could do it in the preferences dialog implementation but then that means the preferences dialog sets global state beyond its purview which is supposed to be the preferences flags alone
> and when the dialog is closed, then what? that state remains
> that's pretty odd don't you think?

stevecotton:
Even after reading the Discord logs, I haven't worked out which preferences this feature is wanted for, but very few of the existing preferences seem candidates; they affect code paths that simply check the preference each time. I'm wondering if it would be better to use the existing `events::generic_event` class and create an event specifically for preferences that should trigger immediate changes elsewhere in Wesnoth.

irydacea:
> Even after reading the Discord logs, I haven't worked out which preferences this feature is wanted for

A screen saver toggle to control SDL's screen saver inhibition feature.

Pentarctagon:
@Vultraz is this something you're still going to pursue?

Vultraz:
I'm not sure what to do with this anymore, and given what Iris said about there are probably better ways to control state than this.

Issue #6269, opened 2021-11-07
Leadership in 1.16 has a bug


MXB2001:
### Game and System Information
 - What version of the game are you running?
1.16.0

 - Where did you download it from? (Steam, Mac App Store, our website, built from source, etc.)
 flathub

 - What OS are you running?
Linux Debian 9

### Describe the bug
When the leadership tag is used with a value set to something negative and cumulative is set to yes the value used is not the negative value but 0.

### To Reproduce

1. First..
```
#define ABILITY_NOCTURNAL
    [leadership]
        id=nocturnal
        name=_ "nocturnal"
        cumulative=yes
        value=25
        affect_self=yes
        [filter]
            [filter_location]
                time_of_day=chaotic
            [/filter_location]
        [/filter]
    [/leadership]
    [leadership]
        id=nocturnal
        cumulative=yes
        value=-25
        affect_self=yes
        [filter]
            [filter_location]
                time_of_day=lawful
            [/filter_location]
        [/filter]
    [/leadership]
#enddef
```

2. Then...
Run a test game with a unit that has this ability.

3. Lastly...
Observe that the damage during night gets the leadership bonus while during the day the -25% is not applied

### Expected behavior
For negative leadership values to work when cumulative=yes (it works when set to no - I have tested)

Wedge009:
I'm asking, not telling: is that how leadership is intended to work? This looks like a non-standard way to try to use it - I'm just wondering if negative values were necessarily considered in the design because there are often other issues with negative values (I'm thinking of #4232) as an example.

newfrenchy83:
I'm asking if this ability tested in 1.14 version?

newfrenchy83:
Why using cumulative=yes in leadership, when it is used in marksman specials, thhe highest value of cth is used(if base_value is greater_ than vlue of special, then base value applied), i think what it is the same here,  during the day, this is the time_of_day bonus who is applied when cumulative=yes because this is the bonus base value. For cumulative=no ,the this is final_value= tod_bonus + leadership_bonus but for cumulative=yes this is : if tod_bonus>leadership bonus then value=tod_bonus.

newfrenchy83:
like said in https://forums.wesnoth.org/viewtopic.php?t=55092 the bugs was fixed in 1.14 but be back in 1.16.

newfrenchy83:
i pull a PR for fix that.

newfrenchy83:
[...]

PR #6270, opened 2021-11-07
fix bug in leadership abilities in master

newfrenchy83:
Fix https://github.com/wesnoth/wesnoth/issues/6269 issue


Pentarctagon:
Please add a unit test for this situation.

stevecotton:
What does this do to the speed of the AI? It seems that get_cumulative_value uses a cached value, but this calculates it with multiple nested loops each time.

I haven't tested, asking because I don't know.

newfrenchy83:
> 
> 
> Please add a unit test for this situation.

test case added

newfrenchy83:
> 
> 
> What does this do to the speed of the AI? It seems that get_cumulative_value uses a cached value, but this calculates it with multiple nested loops each time.
> 
> I haven't tested, asking because I don't know.

i not decide with this PR or added 'add' and sub to schema validation now what it can be used instead of cumulative=yes in leadership

Pentarctagon:
Interestingly, though no idea if it's related or not: https://forums.wesnoth.org/viewtopic.php?t=55094

newfrenchy83:
> 
> 
> What does this do to the speed of the AI? It seems that get_cumulative_value uses a cached value, but this calculates it with multiple nested loops each time.
> 
> I haven't tested, asking because I don't know.

for limit eventual decreased of AI speed, i implement for what abil.highest("value")( and lowest also) called what if [leadership] with cumulative=yes is present in the list, else if default calculation who is used. and 'add' and 'sub' can be used in same time.

newfrenchy83:
@Pentarctagon I'll try to explain what this PR does:
first, all the abilities [leadership] and only those there are checked to verify if any of them has the attribute "cumulative = yes"
if this is the case then the method of calculating 'value' in abilities.cpp is abandoned in favor of the old method used for leadership in version 1.14, and this calculation is only performed once.
However, calculations with 'add', 'sub', 'multiply' or 'divide' remain active in case a leadership uses one of these attributes instead of 'value'.
Compatibility is not broken while fixing an issue that appeared in version 1.15, but too little reported before for me to understand the true usefulness of the cumulative option in [leadership] and I apologize for it.

newfrenchy83:
@Pentarctagon review https://github.com/wesnoth/wesnoth/pull/6281 in same time, please.

newfrenchy83:
someone could review this PR?

Pentarctagon:
Unless there are further comments in response to https://github.com/wesnoth/wesnoth/issues/6269#issuecomment-970588681 I am planning on closing the issue and the related PRs since it seems to actually be working as intended now.

newfrenchy83:
> Unless there are further comments in response to [#6269 (comment)](https://github.com/wesnoth/wesnoth/issues/6269#issuecomment-970588681) I am planning on closing the issue and the related PRs since it seems to actually be working as intended now.

only if you use add or sub insteat of  value with cumulative=yes. else, the using of cumulative=yes remain diffrent in 1.16 what in 1.14.

PR #6273, opened 2021-11-07
Fix leadership bug for 1.16

newfrenchy83:


Pentarctagon:
This wouldn't be able to be fixed in 1.16 since it would break compatibility with existing clients.

PR #6274, opened 2021-11-07
use "special_effect" instead of 'name' in attack prediction windows for abilities used like weapon

newfrenchy83:


Pentarctagon:
This can't go into 1.16 since it adds a new feature to the API.

PR #6275, opened 2021-11-07
color 'name' in attack prediction windows for abilities used like weapon based on benefit(master)

newfrenchy83:


CelticMinstrel (changes requested):
You also need a female variant of this key (something like `female_special_effect`).

I'm not entirely happy with the actual name of the key (`special_effect`) but I'll let others decide whether it needs changing

newfrenchy83:
> 
> 
> You also need a female variant of this key (something like `female_special_effect`).
> 
> I'm not entirely happy with the actual name of the key (`special_effect`) but I'll let others decide whether it needs changing

i'm not convinced for female variant bacause this is attack who is affected, not unit.

CelticMinstrel:
> i'm not convinced for female variant bacause this is attack who is affected, not unit.

But the attack is owned by a unit. Also, there's already a `female_name`, which presumably is used for display when the attack is owned by a female unit, so this should be consistent with that.

newfrenchy83:
@CelticMinstrel female vriant added

src/units/abilities.cpp
```diff
@@ -899,16 +900,20 @@ std::string attack_type::weapon_specials(bool only_active, bool is_backstab) con
...
 
-				add_name(weapon_abilities, active, sp, checking_name);
+				add_name(weapon_abilities, active, sp, checking_name, gender);
 			}
 		}
 	}
 
 	if(other_){
+		unit_race::GENDER gender = other_->gender();
+		if(self_){
```
CelticMinstrel, line 910: Is this check even correct? You checked the "have a self" case just above here… but if you have a self _and_ an other, this means it uses the gender of the self rather than that of the other. Is that correct though?

stevecotton:
I feel this still isn't the correct UX - that it's presenting information in the place that's logical to someone who is thinking about the implementation, but not presenting it in the place that's logical to the player.

Other parts of the UX have tooltips giving more information, maybe a tooltip on the chance-to-hit could show
```
60% terrain
-10% enemy's formation ability
```

Suggestion from @Pentarctagon in the Discord:
> I don't suppose it could be something like having a red down arrow to the left of the name of abilities that are having a negative affect of some kind?
> ie: ▼formation
> since the unit is under the effect of their opponent's Formation ability and it is affecting their attack
> so it seems like it's not that the information itself is irrelevant, just that it's presentation is confusing

src/units/abilities.cpp
newfrenchy83: Correct,self_ is the unit who own attack affected by abilitie, and 'special_efffect' describe the effect on attack.

this is the attack of self who has 'decreased presision', in attack prediction windows and not the unit itself, this is why idon't understand the idea of female variant because for me; attack is a female name and never male (in french)

newfrenchy83:
@Pentarctagon formation abilitie implemented in unit who have melee attack only, but who must defend against all range of attack, if formation imlmented in specials, then only attack melee of opponent will be decreaded.

Or implmente cth in opponent like in 1.14, but the use of abilite used like specials is also t don't have to make that.

src/units/abilities.cpp
CelticMinstrel: I'm not 100% sure that it makes sense to have male and female strings here; however, it should be consistent. Either both name and special effects have male/female versions, or neither do.

Pentarctagon:
Please post a screenshot of what the UI looks like with the change as well.

newfrenchy83:
> 
> 
> Please post a screenshot of what the UI looks like with the change as well.
color_name_abilitie_red

if formation increase instead to decrease cth opponent, the color will be green :
color_name_abilitie_green

src/units/abilities.cpp
newfrenchy83: i remove special_effect attribute

Pentarctagon:
Is stevecotton's suggestion of a tooltip feasible here?  Also the up/down unicode arrow from my suggestion?

newfrenchy83:
> 
> 
> Is stevecotton's suggestion of a tooltip feasible here? Also the up/down unicode arrow from my suggestion?

for stevecotton suggestion, i don't think could doing that? and for the up/down arrow, the idea isn't bad, but i'm not sure to know implement instead o color, who remain more visible wht arrow.

newfrenchy83:
https://github.com/wesnoth/wesnoth/pull/6280 is 1.16 equivalent to thi PR, and i use color for don't add new feature who will not be implement 1n 1.16 versions

newfrenchy83:
> 
> 
> It wouldn't be instead of the color, it would be in addition to it. And I think it should be possible just by prepending the unicode escape sequence?

i tried, but i failed. C++ don't know the unicode for arrow, when i put the unicode, i obtain rectangular symbol

CelticMinstrel:
That's probably a problem with the font, not C++. Did you write it as `'\u25BC'` and `'\u25B2'`? Or you could also try `'\u2B63'` and `'\u2B61'`. As long as the works and draws _something_ on the screen, C++ is handling it just fine; if you're still not seeing an arrow, it's a font problem.

A tooltip should be possible, but that's done at a different level from the changes in this PR – it needs to be done in the UI code.

newfrenchy83:
> 
> 
> That's probably a problem with the font, not C++. Did you write it as `'\u25BC'` and `'\u25B2'`? Or you could also try `'\u2B63'` and `'\u2B61'`. As long as it works and draws _something_ on the screen, C++ is handling it just fine; if you're still not seeing an arrow, it's a font problem.
> 
> A tooltip should be possible, but that's done at a different level from the changes in this PR – it needs to be done in the UI code.

it's a font problem.

Pentarctagon:
So to confirm, the unicode up/down arrow isn't something that is supported with whatever font is used here?

newfrenchy83:
> 
> 
> So to confirm, the unicode up/down arrow isn't something that is supported with whatever font is used here?

Yes

Pentarctagon:
@stevecotton Do you have any further thoughts given the issues with implementing the tooltip and the up/down arrows?

newfrenchy83:
I think to one thing : it will be a good idea to extent the good_color to boolean weapon like abilities?

PR #6277, opened 2021-11-07
fix cumulative=yes cancel negative value of some abilities(1.16)

newfrenchy83:
suggest an other resolve to https://github.com/wesnoth/wesnoth/issues/6269 issue


Pentarctagon:
How does this fix the issue?  All it adds are some attributes to schema and a unit test.

newfrenchy83:
> 
> 
> How does this fix the issue? All it adds are some attributes to schema and a unit test.

excuse me, this is not fix the issue, but suggest other way to resolve.

Pentarctagon:
For 1.16 the fix would probably be to have Formation be a regular weapon special that's added onto each weapon, rather than an ability that automatically adds the weapon special to all of the unit's attacks.

newfrenchy83:
> 
> 
> For 1.16 the fix would probably be to have Formation be a regular weapon special that's added onto each weapon, rather than an ability that automatically adds the weapon special to all of the unit's attacks.

Uh?

Pentarctagon:
Commented in the wrong PR, whoops.

newfrenchy83:
i added a fix for cumulative=yes, for in case of negative value, this is the lowest value and not the highest who was used

PR #6278, opened 2021-11-07
revert extent addition of the names of the abilities used as weapons in the attack dialog window to apply_to=opponent(1.16)

newfrenchy83:


newfrenchy83:
@Pentarctagon formation abilitie implemented in unit who have melee attack only, but who must defend against all range of attack, if formation imlmented in specials, then only attack melee of opponent will be decreaded.

Or implmente cth in opponent like in 1.14, but the use of abilite used like specials is also t don't have to make that.

PR #6279, opened 2021-11-07
HttT S05b - Added cracked temples

### LINKED ISSUE #5576: HttT Isle of the Damned: temples look recently built
Terrain description for the ancient temples is currently just "dirt flat". Konrad refers to them as "ancient temples".
They don't look ancient, though. Perhaps just **"temple"** would do?

Descacharrado:
Added some cracks to the temples so the scenario dialogue is more fitting (see #5576).

Result can be seen in the following image (top to bottom temple-cracked1, 3, 2):
A

Changes can definitely be made on the location of each temple, the using of a normal temple (temple1.png) in some spot or the modification (by me or whomever) of the images.

(Closes #5576).


Wedge009:
As mentioned before, looks good to me. Thanks for this effort!

stevecotton:
The images look good, but a couple of change requests:

* `utils/woptipng.py` can compress those .png files better.
* The same temples appear in TRoW S17d too, and should change there tool.

Descacharrado:
> * The same temples appear in TRoW S17d too, and should change there too.

From #5576:

> well the temples were also there in the scenario in The Rise of Wesnoth that also takes place there, and that was 517 years before HttT, so I'd say that qualifies as "ancient"

TRoW S17d occurs a couple hundred years before HttT, which is why (imo) TRoW's temples should keep their look.

Wedge009:
I think that's reasonable.

Should probably squash this when merging - if it's okay to merge.

Wedge009:
I reckon this can go to 1.16 too, if it's accepted.

nemaara (approved):
Looks good but I don't have git on this comp so will merge and port to 1.16 when I get to my other computer.

Issue #6284, opened 2021-11-11
Add-ons server keeps undefined properties at their previous value upon update


vgaming:
### Game and System Information
 - What version of the game are you running? wesnoth_addon_manager version 170ec9ba139e240ddbb445e102c6112a2ea6b6fd  2021-09-25
 - Where did you download it from? `git clone`
 - What OS are you running? ArchLinux

### Describe the bug
Context: I have an add-on (Creep_Wars_Dev) that I try to update. It previously had a dependency on `pick_advance`, and now I want to remove it as it's a built-in wesnoth dependency now.

The bug happens when I update the add-on using wesnoth_addon_manager. I'm using this _server.pbl: https://github.com/vgaming/CreepWars/blob/8a417ccf1ec12e589079845674490165d7ff82f3/Creep_War_Dev/_server.pbl  which has no `dependencies` defined.
When I update my addon, the add-on store says that my addon is accepted. The version is updated accordingly, as seen in the game client. **However, the dependency does not get removed and remains to be the old one**.

In other words, updating an addon with not-specified `dependencies` did not cause removal of the dependencies. The only thing that worked was to submit this update: https://github.com/vgaming/CreepWars/commit/1d2358c492052e3f3ab416e26fe23545dedc1719  after which everything worked as expected

### Steps To Reproduce

1. Create a new add-on with a dependency specified
2. Use wesnoth_addon_manager to publish the addon
3. Remove the dependency from the add-on
4. Update the addon
5. Observe from the game client that the add-on still requires the old dependency

### Expected behavior
After publishing an add-on that **does not specify** any dependencies, dependencies should be removed by the addons server

### Screenshots
![2021 11 11_18:35:46e](https://user-images.githubusercontent.com/33452702/141343518-ce960c89-1198-46a7-8c2d-c9b412ea6b91.png)

soliton-:
Official add-on server instances have been restarted with the fix.

PR #6289, opened 2021-11-12
campaignd: Drop removed pbl attributes on upload if applicable

irydacea:
This adds a `config::copy_or_remove_attributes()` method with the semantics needed to solve issue #6284 in campaignd, and uses it instead of `config::copy_attributes()` in campaignd. The latter is now unused in the overall codebase, but I'm leaving it intact with new documentation in case it's useful to someone else in the future.

### DIFF
--- a/src/config.hpp
+++ b/src/config.hpp
@@ -525,6 +525,12 @@ public:
 		}
 	}
 
+	/**
+	 * Copies attributes that exist in the source config.
+	 *
+	 * @param from Source config to copy attributes from.
+	 * @param keys Attribute names.
+	 */
 	template<typename... T>
 	void copy_attributes(const config& from, T... keys)
 	{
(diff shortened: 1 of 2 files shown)


stevecotton:
IIUC the new function is equivalent to the following, and the following seems clearer to me.
```
addon.clear();
addon.copy_attributes(...);
```

stevecotton:
@soliton- I'd support either merging this or (assuming it's equivalent) the `addon.clear()` alternative.

soliton-:
It's only equivalent if all attributes in `addon` are listed and that's not the case. The idea is that those attributes that are listed are the ones the uploader is responsible for but nothing else. You can see in the diff context that there are potentially more attributes in `addon` than are listed.

Yeah, this should be merged.

Issue #6288, opened 2021-11-12
Wesnoth fails to remember multiple account passwords


vgaming:
### Game and System Information
wesnoth-1.16.1 running on ArchLinux, installed from ArchLinux repository

### Describe the bug
Wesnoth seems to have bugs with remembering online account passwords. Passwords are prompted for accounts that already were submitted password for memoization, and the size of the `credentials` files grows unpredictably

### Steps to reproduce, Scenario 1
I've migrated from wesnoth-1.14 to wesnoth-1.16. Each time I start wesnoth and try to log in as "vasili", wesnoth prompts me for the password, even though I have the option to remember the password checked. Each time I close back wesnoth, `credentials` file grows by about 50 bytes.

1. Use my already existing `credentials` file... (I know it's only reproducible by me, but it's reproducible. See Scenario2 below for Steps that can be done by anybody.)
2. Start wesnoth, try to log in
3. Be prompted for the password. Enter it in, log in to main wesnoth online server
4. Close wesnoth. Observe that the `credentials` file grew by ~50 bytes
5. Start wesnoth again, try to log in, observe that password need to be entered again

### Steps to reproduce, Scenario 2
1. Remove `credentials` file (or temporarily rename it)
2. Start wesnoth, log in as "vasili" (I'm not sure if names matter, I guess you can replace with your own, except if alphabetical order or some hashing matters). Observed at this point: you are asked for the pass as expected, and the `credentials` file is around 50 bytes after closing wesnoth.
3. Start wesnoth, log in as "vasya2". Observed at this point: again you are asked for the pass as expected, and the `credentials` file grows by another 50 bytes to a total of ~100 bytes.
4. Start wesnoth, log in as "vasya2" (again). This time, you would expect wesnoth to already remember your password. It doesn't happen, however. Observed behavior is: you are asked for the pass again, and the `credentials` file grows to around 150 bytes (should've stayed at ~100 bytes at this point).
5. Start wesnoth, log in as "vasya2" (again). This time, with the `credentials` file being ~150 bytes in size, wesnoth will let you through without password prompt, as would be expected. The `credentials` file does not grow in size when closing wesnoth. This behavior should've been on previous step, but is only observed now.

### Screenshots
![2021 11 12_08:48:07](https://user-images.githubusercontent.com/33452702/141429871-58c03f4c-aaba-48ee-b307-185e01feb9eb.png)

### Additional context
There are no relevant logs being made in the console when I log in.

[...]

PR #6290, opened 2021-11-12
Fix multi-account password saving

vgaming:
Fixes GH-6288, and hopefully makes the code
more readable / easier to understand.

### DIFF
--- a/src/preferences/credentials.cpp
+++ b/src/preferences/credentials.cpp
@@ -37,6 +37,7 @@
 #endif
 
 static lg::log_domain log_config("config");
+#define DBG_CFG LOG_STREAM(debug , log_config)
 #define ERR_CFG LOG_STREAM(err , log_config)
 
 class secure_buffer : public std::vector<unsigned char>
@@ -154,6 +155,7 @@ namespace preferences
 
 	std::string password(const std::string& server, const std::string& login)
 	{
+		DBG_CFG << "Retrieving password for server: '" << server << "', login: '" << login << "'\n";
 		auto login_clean = login;
 		boost::trim(login_clean);
 
@@ -177,6 +179,7 @@ namespace preferences
 
 	void set_password(const std::string& server, const std::string& login, const std::string& key)
 	{
+		DBG_CFG << "Setting password for server: '" << server << "', login: '" << login << "'\n";
 		auto login_clean = login;
 		boost::trim(login_clean);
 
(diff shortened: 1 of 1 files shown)


vgaming:
I tested the new code locally, it works both for single-account use case and for storing/loading multiple accounts

vgaming:
I also hate to admit, but I cannot pinpoint *where exactly* the old code was broken. In my perception that was just messy code that is really hard to understand, with unneeded additional variable `offset`, triple resizes that adjust for escaped and not escaped keys... A lot of unnecessary complexity. In the issue GH-6288 itself, however, it was observed that `credentials` file is not split correctly, with the `CREDENTIAL_SEPARATOR` obviously broken. I can copy more of the relevant IRC discussions here if there's a question

Wedge009:
As discussed in #wesnoth-dev, it is indeed strange that it didn't expose the issue for multi-account credential saving for you in 1.14. But it does look like replacing manual `std::copy` and `offset` tracking is a good thing and if it solves the bug for you and @Hejnewar.

It looks like only @CelticMinstrel has done any meaningful work on this code, so I'll suggest he review this before merging.

vgaming:
UPD: When using this branch on an old corrupted `credentials` file from wesnoth-1.16.1, the passwords have still to be entered once, but they're saved successfully and for the user it works as expected. (This PR doesn't really change reading of the credentials file.) What remains however is that the `credentials` file will keep those extra broken/unneeded entries, even with the new PR. They'll just stay as garbage around, without use.

vgaming:
To provide a bit of context for people who weren't today in IRC. Without this PR, for single-account credentials file, the entry in credentials file looks like:  server: server.wesnoth.org:15000, login: vasili. For multi-account credentials file, without the PR, corrupted entries will be created upon save, such as:  server: server.wesnoth.org:15000, login: login: vasili@server.wesnoth.org:15000=�����;sa��vasya2   (the login spans much further than it should, including the server and probably password of the previous/next entries)

src/preferences/credentials.cpp
```diff
@@ -230,21 +233,15 @@ namespace preferences
 			filesystem::delete_file(filesystem::get_credentials_file());
 			return;
 		}
-		secure_buffer credentials_data(1, CREDENTIAL_SEPARATOR);
-		std::size_t offset = 1;
+		secure_buffer credentials_data(0, CREDENTIAL_SEPARATOR);
```
CelticMinstrel, line 236: `(0, CREDENTIAL_SEPARATOR)` does nothing - just remove it.

CelticMinstrel (changes requested):
Looks good, just one minor change to make.

src/preferences/credentials.cpp
vgaming: Hey, I was expecting that -- I just newb enough in C++ to not know the alternative.. So what do I write then, `credentials_data();` or `credentials_data;` ?

vgaming: UPD: neverming, found it, the compiler is pretty unambiguous on what I should do anyway :sweat_smile:

vgaming: Thanks for the review! Addressed the comment

CelticMinstrel (approved):

AI0867:
The only potential bug I see in the old code is that the second resize (for the difference in size between escaped and non-escaped keys) doesn't pad the buffer with CREDENTIAL_SEPARATOR.

The new code looks like a definite improvement. It might be a bit slower due to not pre-allocating the buffer, but this is hardly a performance-critical path.

vgaming:
@AI0867 totally agree. Also, for this code path, I'd prefer readability, maintainability and less bugs, over performance optimization %insert-donald-knuth-quote-here%

AI0867:
Mostly unrelated, but I'm unsure if I should open a new issue for this: I noticed a TODO to do key stretching on the autogenerated key. This seems like a good idea, but unfortunately, it comes with a whole host of backwards-compatibility issues, especially if we want to be able to allow older versions of wesnoth to keep using the same credentials file.

Should we open a new issue for that to think about what a good design would even look like?

vgaming:
@AI0867 I don't know, I can't really say. Normally I always try to prioritize fixing bugs, and only then doing new features. But that is subjective of course.

PR #6293, opened 2021-11-14
Check whether to display lobby joins in the game creation screen.

### LINKED ISSUE #6234: "Show joins of friends only" does nothing
 - 1.16
 - Steam
 
### Describe the bug
"Show joins of friends only" selected in Multiplayer settings does nothing and still shows all lobby joins.

### To Reproduce

1. Set "Show joins of friends only" selected in Multiplayer settings
2. Create a game lobby (do not start the game)
3. Watch lobby messages

### Expected behavior
I only get messages about friends joining.

Pentarctagon:
Fixes #6234

### DIFF
--- a/src/gui/widgets/chatbox.cpp
+++ b/src/gui/widgets/chatbox.cpp
@@ -582,6 +582,8 @@ void chatbox::process_message(const ::config& data, bool whisper /*= false*/)
 	if(whisper) {
 		add_whisper_received(sender, message);
 	} else {
+		if (!preferences::parse_should_show_lobby_join(sender, message)) return;
+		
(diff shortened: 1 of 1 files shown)


Wedge009:
I assume this does the job but I find it interesting that it wasn't required in 1.14 (unless the bug was also present there). Maybe a product of various lobby refactoring...?

Pentarctagon:
Yeah my assumption is that something was refactored and now this code is handling this network message instead of what was handling it previously.

Vultraz:
I'm fine with this as a stopgap but I say the message should be removed entirely and the notifications handled another way.

PR #6296, opened 2021-11-15
Fix [store_unit]'s stored coordinates for recalls

### LINKED ISSUE #6295: [modify_unit] returns all units put to the recall list with [put_to_recall_list] on the battlefield
 - What version of the game are you running? 1.16.0 from flatpak

### Describe the bug
[modify_unit] tag places back all units that have been put to the recall list in this scenario with [put_to_recall_list] on hexes where they stood before being put to the recall list. If several units were put to the recall list from the same hex, only the last of them appears, others are lost (they aren't on the map and no more in the recall list)

### To Reproduce
Dunno how much it is a convenient way to show this, but I made a showcase campaign (WML Guide with the first scenario altered)
[WML_Guide_bug_showcase.zip](https://github.com/wesnoth/wesnoth/files/7534509/WML_Guide_bug_showcase.zip)

1. Install manually the showcase addon campaign
2. Open "WML guide" campaign. You will have menu items "To Recall", "Win" and "Modify" in the first scenario
3. Recruit a unit on some hex, then put it to the recall list with the menu item, then recruit another unit on the same hex and put it to the recall list again.
4. Call [modify_unit] (either with "Modify" or with "Win")
5. Notice how all units put to the recall list returned to their places. And if several units were on the same hex, notice that all but one are now lost (not on the map and not in the recall list)

### Expected behavior
Units get modified but don't reappear on the map from the recall list

### LINKED ISSUE #6315: [store_unit] doesn't replace x,y of units to "recall", "recall" when you store units from the recall list
Most likely connected with #6295 

If you have units in the recall list with coords not equal to 0,0 (that happens if you just put units to the recall list with [put_to_recall_list] or if you saved units from the map to a variable and then unstored them into the recall list), [store_unit] doesn't replace their coords to "recall", "recall", as a result [unstore_unit] places them on the map. So the unit that had to be unstored back to the recall list appears on the map.

In 1.14 [store_unit] changed x,y to recall, recall if you stored units from the recall list, and after needed manipulations [unstore_unit] put it back to the recall list.

stevecotton:
Units on the recall list might have x,y coordinates that are on the map, which
therefore need to be replaced with "recall,recall" within [store_unit]. The
existing code created a temporary variable, changed the coordinates, and then
returned the unchanged original instead of the temporary.

Add a new test that `[put_to_recall_list]` followed by `[modify_unit]`
doesn't move the unit back to the map.

Fixes #6295
Fixes #6315

### DIFF
--- a/data/lua/wml-tags.lua
+++ b/data/lua/wml-tags.lua
@@ -403,7 +403,7 @@ function wml_actions.store_unit(cfg)
 			ucfg.x = 'recall'
 			ucfg.y = 'recall'
 		end
-		utils.vwriter.write(writer, u.__cfg)
+		utils.vwriter.write(writer, ucfg)
 		if kill_units then u:erase() end
 	end
 end
--- a/data/test/scenarios/put_to_recall_and_modify.cfg
+++ b/data/test/scenarios/put_to_recall_and_modify.cfg
@@ -1,10 +1,11 @@
 #textdomain wesnoth-test
 
 #####
-# API(s) being tested: [modify_unit] on the recall list, [role][auto_recall]
+# API(s) being tested: [put_to_recall_list], [modify_unit], [role][auto_recall]
 ##
 # Actions:
-# Create Charlie, belonging to side 1, on the recall list.
+# Create Charlie, belonging to side 1, on the map.
+# Repose Charlie on the recall list.
 # Assign Charlie to side 2.
 # Recall Charlie next to his side's leader.
 ##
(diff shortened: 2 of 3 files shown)


Pentarctagon:
This looks like something that could be reasonably easily bisected?

stevecotton:
Good idea, running the bisection now.

stevecotton:
It turned out to be a one-line fix.

CelticMinstrel:
This doesn't look write… isn't `ucfg` undefined when you pass it to `vwriter.write`?

stevecotton:
> This doesn't look right… isn't `ucfg` undefined when you pass it to `vwriter.write`?

Have a look at the file in an editor, not just Github's diff. For some reason, Github's rendering of the diff makes tabs on unchanged lines wider than tabs on changed lines.

CelticMinstrel:
Ohhh, I get it, the `end` that looks like it ends the `for` actually ends the `if`… so this looks fine, then.

PR #6297, opened 2021-11-15
Eastern Invasion and Liberty Bone Knight attack image

Descacharrado:
From #4145:

> [ ] Bone Knight has no attack icon for the "trample" attack

Bone Knight's image was added to core.

In-game it looks like this:

Bone knight trample image in-game

EDIT: removed kick-orcish.png in f79848e5 because this commit inspired doofus to work on better icons on #6339


Wedge009:
Thanks for taking this initiative. Overall it makes sense to me. Are the images yours? I'm not sure how the skeletal hoof icon looks all that different from the normal hoof icon. Also, is the Bone Knight only used in EI? If not, might it make sense to have the icon in core as well?

Note to whoever merges: please remember to squash.

Wedge009:
I think making it whiter would help - the brown (which I suppose is meant to give an 'aged' look) makes it look like a normal hoof to me.

Wedge009:
Looks good to me! And good to see Liberty benefits too. It's not my decision, but I approve. (:

doofus-01 (changes requested):
I'll work on a more detailed review this weekend, but for now:
1. hoof-skeletal.png looks fine, it can go in
2. kick-orcish.png needs work before merging

Wedge009:
I forgot to mention that the kick icon is quite dark. I also thought we had kick before, but looking at #6071 again, it is indeed currently missing.

doofus-01:
So about that kick image:
Execution - Too narrow color range, needs some highlights.  Context matters, it probably wouldn't look as bad in game as it does in github.com, but it still needs highlights.  Compare it to `fist.png`, for example.
Design - It looks like a small, bare foot?  That might make sense for a goblin or zombie, I'm not sure it does for the nightblade.

orc-kick
That's my relatively quick take, it still needs some work but is hopefully worth more than words.  In general, stick with the mainline (non-magical) attack icons as models, and using more materials (leather, metal, etc.) can make it more interesting.

Descacharrado:
That is way better, I would vote on putting that in.

Wedge009:
@doofus-01 Since it's just the skeletal hoof being added now is it okay to merge this?

doofus-01:
It is OK to merge now, thanks.

PR #6301, opened 2021-11-16
Refactor handling of allies in UtBS (plus compatibility)

Wedge009:
Depends on / is super-set of #6165, which refactors ally dialogue, and attempts to (but incompletely) handle the variable name correction from `$ally_name` to `$ally_id`. The original variable name was confusing because the values it held referred to the `id` of the ally, not the `name`.

Outstanding issue: The final scenario does not work properly because the variable `$ally_id` is used in the `[side]` declaration.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
@@ -317,6 +317,7 @@
         [/recall]
 
         # recall dwarf/troll ally
+        {ALLY_NAME_COMPATIBILITY}
         [recall]
             id=$ally_id
             x,y=51,43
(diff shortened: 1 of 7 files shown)


nemaara:
Is this mergeable? Or do you want to fix the last scenario first.

Wedge009:
From #6165, it was @stevecotton's suggestion to try to make the variable change less of a problem for those in the middle of playing through the campaign - even though arguably major version changes allow for breaking changes in campaign code. Strictly speaking, the variable name change is not necessary, but since `id` and `name` are different attributes of a `unit`, leaving it as `$ally_name` was really confusing to me.

The way I implemented the compatibility was just with an `[if]` block but unfortunately it doesn't work as intended for the final scenario where the variable is used in the `[side]` declaration - it appears the way WML works is that the condition can only be run inside an `[event]` block. Apparently even the `preload` event  is not early enough to make the variable name change effective. @CelticMinstrel suggested that we can try to use a raw `[lua]` block to work around the issue but I'm still quite weak on WML stuff, so need some help working out how to implement that.

nemaara:
Okay one of them can merge it then since this is more of a code related thing.

Wedge009:
Abandoning due to reversion of #6165.

PR #6302, opened 2021-11-16
UtBS Hide Dark Assassin's race until uncloacked

### LINKED ISSUE #6213: UtBS: Keep race of Dark Assassin hidden from the player
 - 1.15.18
 - Steam
 - Windows 10

Display the race of the Dark Assassin as ?????, at least until it is revealed that the assassin is an elf.

### Rationale
In the story, the characters (and supposed to be the player as well) get surprised when they discover that the assassin is an elf. However, the player can already know that the assassin is an elf from seeing the Race as Quenoth Elf in the sidebar and in the Help Encyclopedia Unit Description.

### Screenshots
This is supposed to be a surprise reveal:
image
But it's not because of this:
image
and this:
image

Descacharrado:
Created a new race which displays the Unknown Unit help page (Lines 14-26 in units.cfg).

~~The only problem that can be seen is that Dark assassin is triplicated (as it is in the code)~~ I defined just one cloacked dark assassin and one uncloaked one and then modified via wml in the scenario cfg file.

Changed the base unit to be the Dark Assassin Uncloaked so less code is used overall (and less changes have to be made to the attack and hp stats).

Q: Why did I delete the hide_help=yes property?  A: It makes the player not be able to read the units description, which I though was not intended as it describes best the behaviour while cloaked:
Why not hide_help=yes

So now players will be able to see the unit's description (clicking on top of the units type) without being spoiled and the ingame race will now be displayed as follows:
Dark assassin cloaked
Dark assassin uncloaked

(Closes #6213)


knyghtmare (changes requested):
Given that the units Dark Assassin 1, 2 and 3 are just 11-20 lines of code...wouldn't it be better to have all 4 of them in a single file? Especially since 3 of them do have translatable strings but have a textdomain.

Descacharrado:
I agree, it will require some changes in scenario files, but I think it is completely doable.

Descacharrado:
With this last commit I have reduced the unit files to only two (I could not do it with only one unit cfg file) and removed the #textdomain from the unit cfg that had no strings. I have tested the changes, and all worked as intended, but further testing would be wise.

Some questions before merging come to mind:
- Is removing the #textdomain correct in Dark_Assassin_Cloaked.cfg?
- Should the units.cfg string be changed from "This unit is unknown..." to "This race is unknown..."?
- What does "no newline at end of file" mean, should I add a new line at the end of wesnoth cfg files?

knyghtmare (commented):
What I meant was clump the unit file codes in a singular file but I see you have taken it a step further. Good.

knyghtmare:
> Is removing the #textdomain correct in Dark_Assassin_Cloaked.cfg?

There aren't any translatable strings there so I guess it's fine?

> Should the units.cfg string be changed from This unit is unknown... to This race is unknown...?

I would go with "The race of this unit cannot be revealed yet..."

knyghtmare (approved):

nemaara (approved):

Issue #6264, opened 2021-11-06
Sidebar tooltips for name and side not showing in 800x600 resolution


stevecotton:
This is the last remaining part of bugs #6007 and #6096. I think it's going to turn out to have a different root cause (seems specific to the low-resolution layout).

At 800x600 resolution, the side-flag is ellipsed (even if the unit has a shorter name), and the tooltips for both the full name and the flag are missing; the movement stats and defense stats (on the line above the flag) can sometimes also be missing at 800x600.
![image](https://user-images.githubusercontent.com/101462/140539196-76f8716c-a728-4ff2-a9b1-6221caf72acc.png)

Having a width of 830 or more seems to make the movement and defense stats reliable.
Having a height of 750 or more seems to make the full name reliable.
Having a height of 750 or more seems to make the flag's ellipse reliable. When the flag is ellipsed, the tooltip seems to only be on the ellipse itself, hovering over the flag image doesn't work.

stevecotton:
For sides with an empty name, there's a different reason that there's no tooltip for the flag; in that case it might be the correct behavior. https://github.com/wesnoth/wesnoth/blob/508c9406f1b87e86faf04e290b70ed028ad4fd67/src/reports.cpp#L234-L238

One such side is the mudcrawlers' side in UtBS S01, which is the scenario where I was testing this; this explains some of the intermittent behavior.

PR #6304, opened 2021-11-17
ui: Fix 800x600 theme's sidebar tooltips

stevecotton:
Two changes in the theme config for 1024x600, which is inherited by the config
for 800x600. Both are the same line but otherwise only loosely related to each
other; together these fix issue #6264, which was that trigger areas for some
tooltips overlapped.

Reduce the width for the unit's name, thus giving the side-flag and side-number
(which get the remaining horizontal space) enough space to display themselves.
128 pixels wide is the same as the name gets in larger themes, even though
those larger themes use larger font sizes. This means that the side number
doesn't get ellipsed, at least for games with up to 9 sides, with the
side-effect that the tooltip can be displayed when hovering over the flag.

Move the name, and thus the flag and side-number down by 1 pixel, so that they
don't overlap with the tooltip trigger areas for movement points and terrain
defense. The name's location is based on the placement of `unit-box-botleft`;
the theme for larger screen-sizes similarly adds 5 pixels of padding because
the `unit-box` is only 72x72 for the image, and the text areas for movement
points and defense go lower than that.

This file is currently identical in 1.16 and master, so forward-porting it will be
a simple cherry-pick.

image

### DIFF
--- a/data/themes/default.cfg
+++ b/data/themes/default.cfg
@@ -557,7 +557,7 @@
         [change]
             id=unit-description
             font_size={DEFAULT_FONT_SMALL}
-            rect="=,+0,+140,+{DEFAULT_FONT_SMALL_HEIGHT}"
+            rect="=,+1,+128,+{DEFAULT_FONT_SMALL_HEIGHT}"
         [/change]
         [change]
             id=unit-type


Wedge009 (approved):
This is tricky to test because not every unit exhibited missing tool-tips. But for the few that I could find, the change seemed to improve the situation regarding tool-tip availability.

stevecotton:
> This is tricky to test because not every unit exhibited missing tool-tips.

It's confusingly intermittent, because it isn't just dependent on which unit is selected. It's dependent on the order in which the information on the sidebar changed, as there's an optimisation that only updates the display when particular information changes.

* If you select a unit that's the same side as the previously selected unit then the tooltip on the flag doesn't refresh.
* The terrain that your mouse passes over between selecting a unit and the sidebar matters - when it only passes over terrain with the same defense stats for that unit then the tooltip doesn't refresh.

(Tooltips should always work once this PR merges, whether or not the bullet-points above apply.)

Wedge009:
Was simply pointing out the intermittent nature of the issue makes it hard to do before/after comparisons. I imagine that's why you sought extra feedback for this otherwise small change.

soliton-:
Would it be possible to emit a warning when tooltip areas overlap or is that used on purpose as well? Pretty sure this isn't the first time such an issue was overlooked for ages because it's so difficult to track down.

stevecotton:
> Would it be possible to emit a warning when tooltip areas overlap or is that used on purpose as well?

Possibly, but at least when I was debugging it didn't seem easy to filter out the false positives. Filtering out ones that have the same top-left corner is easy, and is in the commit that's just auto-linked above. Filtering out matching right-hand corners for right-justified text didn't solve the problem either (and I dropped that code as it wasn't helping my debugging).

In display::refresh_report, first the area for the report is cleared (in the commit that's just auto-linked above, this is on the line after the new "Removing tooltips that overlap with" log), and then a loop that calculates offsets of `x` can add multiple areas from a single report. For example, "cures, heals+8" has two areas with two tooltips, yet even for the "cures" part the left edge is a few pixels to the right of the left edge of `rect` when that log is printed.

The code in `clear_tooltips(const SDLRect&)` doesn't know which tip is which (although it does have access to the text of the old tip, via `i`, and the caller has the name of the new tip). When debugging 6096 I just kept an
image-editor open with a screenshot and used it to look up the coordinates.

stevecotton:
@Vultraz I added you as a reviewer because this is touching the UI. Should I wait or just merge?

Wedge009:
He seems to be busy with studies these days, I waited four months for #5987 (which was your suggestion anyway) before eventually merging after the string freeze ended.

Vultraz (approved):
I'm not familiar enough with the theme code to really comment here, but if whatever issue you have is fixed by this change it's good with me.

PR #6308, opened 2021-11-19
Add code comments stating what each unit test is testing.

Pentarctagon:
---

This PR is the start of three steps I'm going go through as I'm able:
1) Add code documentation to the existing unit tests.
2) Reorganize the existing unit tests into directories based on which APIs they are testing.
3) Add more tests to fill in the areas that don't have any unit test coverage.

### DIFF
--- a/data/test/scenarios/ability_id_not_active_test.cfg
+++ b/data/test/scenarios/ability_id_not_active_test.cfg
@@ -4,15 +4,15 @@
 # API(s) being tested: [have_unit]ability_id_active
 ##
 # Actions:
-# Give Alice an unconditionally active [hides] ability.
-# Recruit a Peasant and check that a unit exists with an active [hides] ability.
-# Recruit a Woodsman and check that a unit exists with an active [hides] ability.
+# Give Alice a [hides] ability active only on forests.
+# Recruit a Peasant and check that no unit exists with an active [hides] ability.
+# Recruit a Woodsman and check that no unit exists with an active [hides] ability.
 ##
 # Expected end state:
-# [have_unit]ability_id_active triggers on each recruit
+# [not][have_unit]ability_id_active does not trigger on either recruit
 #####
 
-{GENERIC_UNIT_TEST test_ability_id_active (
+{GENERIC_UNIT_TEST test_ability_id_not_active (
     [event]
         name=start
         {VARIABLE triggers 0}
@@ -26,6 +26,11 @@
                 [abilities]
                     [hides]
                         id=hides_test
+                        [filter]
+                            [filter_location]
+                                terrain=*^F*
+                            [/filter_location]
+                        [/filter]
                     [/hides]
                 [/abilities]
             [/effect]
@@ -56,9 +61,11 @@
             type=Peasant
         [/filter]
         {ASSERT (
-            [have_unit]
-                ability_id_active=hides_test
-            [/have_unit]
+            [not]
+                [have_unit]
+                    ability_id_active=hides_test
+                [/have_unit]
+            [/not]
         )}
         {VARIABLE_OP triggers add 1}
     [/event]
@@ -68,9 +75,11 @@
             type=Woodsman
         [/filter]
         {ASSERT (
-            [have_unit]
-                ability_id_active=hides_test
-            [/have_unit]
+            [not]
+                [have_unit]
+                    ability_id_active=hides_test
+                [/have_unit]
+            [/not]
         )}
         {VARIABLE_OP triggers add 1}
     [/event]
--- a/data/test/scenarios/event_handlers_in_events.cfg
+++ b/data/test/scenarios/event_handlers_in_events.cfg
@@ -1,3 +1,13 @@
+#####
+# API(s) being tested: [event][fire_event]
+##
+# Actions:
+# Define a custom event
+# Fire the custom event from within another event with [fire_event] using its name
+##
+# Expected end state:
+# The custom event has executed
+#####
 {GENERIC_UNIT_TEST "event_handlers_in_events_1" (
     [event]
         name=test
@@ -13,6 +23,16 @@
     [/event]
 )}
 
+#####
+# API(s) being tested: [event][fire_event],[event][event]
+##
+# Actions:
+# Define a custom event nested within another event
+# Fire the custom event from within the same event than the custom event was defined in with [fire_event] using its name
+##
+# Expected end state:
+# The custom event has executed
+#####
 {GENERIC_UNIT_TEST "event_handlers_in_events_2" (
     [event]
         name=start
@@ -29,6 +49,17 @@
     [/event]
 )}
 
+#####
+# API(s) being tested: [event][fire_event],[event][event]
+##
+# Actions:
+# Define a custom event nested within another event
+# The nested event is defined first in the WML
+# Fire the custom event from within a different event than the custom event was defined in with [fire_event] using its name
+##
+# Expected end state:
+# The custom event has executed
+#####
 {GENERIC_UNIT_TEST "event_handlers_in_events_3" (
     [event]
         name=start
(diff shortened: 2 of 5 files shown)


Wedge009:
Given these tests seem to have a lot of parts, it's probably necessary to have some explanatory comments but something I was taught (that makes sense to me) is to try to pack as much of the commentary into the test name itself (even if that means longer test names). Lest the commentary and the test code diverge and you wind up having to maintain two things - the code and the commentary together.

data/test/scenarios/ability_id_active_test.cfg
```diff
@@ -1,3 +1,19 @@
...
+# API(s) being tested: [have_unit]ability_id_active
+##
+# Starting state:
+# Side 1 leader Alice (Elvish Archer)
+# Side 2 leader Bob (Orcish Grunt)
+##
+# Actions:
+# Give Alice an unconditionally active [hides] ability.
+# Recruit a Peasant and check that a unit exists with an active [hides] ability.
+# Recruit a Woodsman and check that a unit exists with an active [hides] ability.
```
stevecotton, line 11: While you're not changing the test here, the test itself seems really weird - as if it's testing for an existing unit (Alice) having her data corrupted by the recruiting mechanism.

Pentarctagon:
@Wedge009 While I agree generally, these unit tests (shouldn't be) things that change often, if at all.  Otherwise it means that we're frequently changing the results the APIs give (whether intentionally or not), which is something we should be trying to avoid so UMC doesn't break.

data/test/scenarios/ability_id_active_test.cfg
Pentarctagon: Yeah I was wondering that myself.  It seems like it'd be an equivalent test to just add the ability and then immediately afterwards check whether the ability is active.

Wedge009:
Oh, okay, I didn't know there were external things relying on them.

Pentarctagon:
In a perfect world, we'd have complete coverage and so any time an API changed we'd either find out when the CI failed or could look back at the file history for a test to see who changed it and why.  Obviously we don't have that now, but that's the ideal.

data/test/scenarios/backstab.cfg
```diff
@@ -82,14 +82,80 @@
...
+#####
+# API(s) being tested: [damage]+[filter_opponent]formula
+##
+# Starting state:
+# Side 1 leader Alice (Elvish Archer)
+# Side 2 leader Bob (Orcish Grunt)
+##
+# Actions:
+# Bob is put on Void terrain (0% defense) and given 100 hp.
+# A Rogue is spawned north, southeast, and southwest of Bob.
```
stevecotton, line 112: They're spawned south, northwest and northeast of Bob (on the hexes that Bob is `n,se,sw` of).

data/test/scenarios/backstab.cfg
```diff
@@ -82,14 +82,80 @@
...
+# Side 1 leader Alice (Elvish Archer)
+# Side 2 leader Bob (Orcish Grunt)
+##
+# Actions:
+# Bob is put on Void terrain (0% defense) and given 100 hp.
+# A Rogue is spawned north and south of Bob.
+# A Rogue attacks Bob.
+##
+# Expected end state:
+# Bob gets hit all three times and ends with 100-(6 * 3 * 2) hp.
```
stevecotton, line 99: I feel this adds a lot of lines of text, but the only new information added is that it's testing `[damage]+[filter_opponent]formula`; and it requires a lot more reading to compare the differences between tests. It also means that checking for missing test cases requires scanning a lot more vertical space.

For understanding and maintaining this set of tests, I think the common docs in lines 4-9 are more useful, and I feel the GENERIC_UNIT_TEST should probably be documented in a single central location.

There are a couple of ways to force 100% chance to hit, of which using void terrain is merely one of them. Putting the exact method in the docs here means more changes if someone refactors the mechanism.

Pentarctagon: That was the point - each test should be documented as to what it's doing individually.

stevecotton: But what advantages do we gain from that, compared to the disadvantages above?

Pentarctagon: I think the parts about GENERIC_UNIT_TEST and the method of making the unit have 0% defense are true.  I don't feel that lines 4-9 fulfill the same purpose as the comment blocks I've added though.

data/test/scenarios/ability_id_not_active_test.cfg
```diff
@@ -0,0 +1,86 @@
...
+# Actions:
+# Give Alice a [hides] ability active only on forests.
+# Recruit a Peasant and check that no unit exists with an active [hides] ability.
+# Recruit a Woodsman and check that no unit exists with an active [hides] ability.
+##
+# Expected end state:
+# [not][have_unit]ability_id_active does not trigger on either recruit
+#####
+
+# wmllint: no translatables
```
stevecotton, line 13: I think this line should be the first line of the file, because the `#textdomain` should be the first line of any file with translatables.

data/test/scenarios/backstab.cfg
stevecotton: Instead of documenting the mechanics of the test, how about changing two of those lines to:
* "Bob is made to have 0% defense but given enough HP to survive."
* "The backstab ability was active, as shown by the amount of damage that Bob takes."

stevecotton: > I guess that logic implies that these lines should just have `yes` or `no` instead of `(6 * 3 * 2)`.

I've created #6473 with that change to the logic, and started with the version of the comments from this PR.

stevecotton:
I keep skipping over this because it's still a draft - if it's rebased then I'll either approve or have only trivial comments left.

stevecotton:
I think the basic format of the documentation is now decided, so I think there's an advantage to merging small sets of files instead of waiting until the full set is ready - it gives a set of examples that can be used when writing new tests.

data/test/scenarios/event_handlers_in_events.cfg
```diff
@@ -48,6 +79,17 @@
     [/event]
 )}
 
+#####
+# API(s) being tested: [event][fire_event],[event][event]
+##
+# Actions:
+# Define a custom event nested within another event
+# The nested event is defined second in the WML
+# Fire the custom event from within the same event that the custom event was defined in with [fire_event] using its name
```
stevecotton, line 88: It's not "within the same event that the custom event was defined in".

This three lines might be clearer as
* Attempt to fire a custom event before that custom event has been defined
* Define the custom event

data/test/scenarios/event_handlers_in_events.cfg
```diff
@@ -266,6 +382,15 @@
     [/event]
 )}
 
+#####
+# API(s) being tested: [event][insert_tag]
```
stevecotton, line 386: It's a bug in the test rather than the documentation, but this test can pass regardless of the filter, because `first_time_only` defaults to `yes`.

stevecotton:
Generally looks good.

Pentarctagon:
Updated to address the comments.

data/test/scenarios/event_handlers_in_events.cfg
stevecotton: It should have `first_time_only=no`, not `yes`; however I've just tested and the filter works both ways. Explanation:

```
##
# Actions:
# Define a post_advance event which includes a filter created via [insert_tag]
# Two units are created, exactly one of which matches the filter
# Both units are advanced a level
##
# Expected end state:
# The event was triggered exactly once
##
```

data/test/scenarios/ability_id_active_test.cfg
```diff
@@ -1,3 +1,15 @@
+#####
```
stevecotton, line 1: Please put the `# wmllint: no translatables` in this file on the first line too.

I'm okay with not adding it if a file doesn't already have it, because that would require checking whether it's accurate or not. I just want to ensure that any new comments are added below it.

data/test/scenarios/event_handlers_in_events.cfg
Pentarctagon: Updated.

data/test/scenarios/event_handlers_in_events.cfg
```diff
@@ -280,6 +406,7 @@
         {UNIT 1 "Elvish Archer" 2 2 ()}
         [event]
             name=post_advance
+            first_time_only=yes
```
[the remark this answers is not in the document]
Pentarctagon: Updated.

PR #6067, opened 2021-09-07
Revise capitalisation

Wedge009:
Felt compelled to review capitalisation after noticing quite a bit of inconsistency, even within a single scenario, let alone across a campaign or different campaigns...

As per https://wiki.wesnoth.org/ProseStyle, race names are generally lower-case unless part of unit or place names, or a specific group (such as 'Quenoth Elves'):
* Elf/Elves/Elvish/Elven
* Drakes
* Dwarf/Dwarves/Dwarvish/Dwarven
* Goblins
* Gryphons
* Humans
* Mer/Merfolk
* Nagas
* Ogres
* Orcs
* Saurians
* Trolls
* Woses

I substitute 'human' for 'men' on occasion to make it clear it's a reference to the race, not males (non-humans sometimes say 'men' when speaking to fellow male soldiers, for example).

Capitalisation of titles is a little less straightforward. Clearly cases such as King Haldric, Queen Asheviere, Prince of Wesnoth, Great Chief, etc, are capitalised.
However, use of a title alone can be valid with or without capital, eg 'the king/King'. I try to be consistent and consider the context:
* Title is used in place of the name, eg compare 'by the King's order' with 'by Garard's order'
* Would the speaker show respect to the person concerned as their leader, or a foreign dignitary? Contrast, 'Asheviere, that vile queen'.
* Compare 'my king' with 'my father' (latter is not usually capitalised in everyday English)
* Occasionally, when the speaker may not otherwise know a character's name or the correct protocol to address, may say something like 'Sir Elf'.

I apologise if I run counter to previous revisions - removal of 'princess' as a direct address in Northern Rebirth, for example, also dealt with the capitalisation issue, if I recall correctly.

Miscellaneous changes:
* Magi, as plural of mage, is not usually capitalised.
* Council may be capitalised when referring to a specific instance, such as the Elvish High Council in the Ka'lian, but not in instances such as 'making council'.

The review focused solely on campaign prose - there may be cases in other data files where capitalisation needs revising too.

PR #6309, opened 2021-11-19
UtBS: Use a lower-case letter for "taurochs"

stevecotton:
It's the name of a species, so lower-casing it is consistent with #6067.

I intend to backport this too, I spotted it while resolving fuzzy strings.


Wedge009:
It's interesting because Wesnoth seems to make an exception to the lower-case policy for unit names, eg [Fire Drake](https://units.wesnoth.org/1.16/mainline/en_US/Fire%20Drake.html). However, all of these changes appear to be referring to taurochs, the animal in general not the unit specifically. So it makes sense to me to make the change.

knyghtmare (approved):
Seems rather trivial but yeah, it's okay.

nemaara (approved):

Wedge009:
One failing test for a string change... looks like a failure of the test itself, though?

stevecotton:
> One failing test for a string change... looks like a failure of the test itself, though?

Looks like the build broke long before it would have reached this file - seems there's no logs from any stage after checkout.

PR #6312, opened 2021-11-19
color name of ability like weapon in attack prediction window and specifie when owned by opponent(master)

newfrenchy83:
in case of specials abilities with value, the name will be colored after the benefit or not of special.
color_name_abilitie_red
if abilitie give disaventage

color_name_abilitie_green

if give advantage.

with_modification

if 'value' used or boolean

### DIFF
--- a/po/wesnoth-lib/fr.po
+++ b/po/wesnoth-lib/fr.po
@@ -8103,9 +8103,11 @@ msgstr ""
 "Un terrain représenté par une chaîne de plus de 4 caractères a été trouvé, "
 "il s'agit de :"
 
-#: src/units/abilities.cpp:850
-msgid "affected by: "
-msgstr "Affectée par : "
+#: src/units/abilities.cpp:922
+#: src/units/abilities.cpp:937
+msgid ""
+"Opponent:\n"
+msgstr "Adversaire :\n"
 
 #: src/wml_exception.cpp:62
 msgid ""
--- a/po/wesnoth-lib/fr.po
+++ b/po/wesnoth-lib/fr.po
@@ -8103,6 +8103,10 @@ msgstr ""
 "Un terrain représenté par une chaîne de plus de 4 caractères a été trouvé, "
 "il s'agit de :"
 
+#: src/units/abilities.cpp:850
+msgid "affected by: "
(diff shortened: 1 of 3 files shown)


stevecotton:
Sorry to keep leaving the same comment, but I still feel that's the wrong place in the UX to put something that (as far as the player is concerned) doesn't belong to the right-hand unit.

newfrenchy83:
> Sorry to keep leaving the same comment, but I still feel that's the wrong place in the UX to put something that (as far as the player is concerned) doesn't belong to the right-hand unit.

and if the plyer unit attack with weapon ranged ait with formation, where 'formation' appear(unit with formation are attack melle only)? describe effect can't be done without new freature, and this not possible in 1.16 now. The only other option is to remove that of attack prediction windows

newfrenchy83:
@stevecotton it can be resolved by
affected_by
:

newfrenchy83:
@stevecotton I study another option that should be better for you by indicating whether the lability applies to the unit or its opponent and I get this:
AFFECT_EITHER
Would that suit you better?

Pentarctagon:
I don't suppose it's possibe to have it be implemented in a way that isn't the opposite of how we want it to be displayed?

newfrenchy83:
@Pentarctagon Répondez-vous à mon dernier commentaire ou au commentaire précédent?

Pentarctagon:
It's a general comment.  The issue seems to really be that an ability is described as buffing one unit but it's implemented by debuffing the other unit.

newfrenchy83:
> It's a general comment. The issue seems to really be that an ability is described as buffing one unit but it's implemented by debuffing the other unit.

the only solution would then be to remove the abilities that apply to the opponent and leave only those with apply_to = self (or attacker and defender when it is self)

newfrenchy83:
@stevecotton i knew what description for 'formation' is not compatible with addition of name in attack prediction and i already pull two PR for remove formation of window, but i use this for exemple for that:
opponentstring

the origin of abilitie is specified, the player will be understand (i hope!)what the attack cth decreased because of defense abilities or his opponent. After, 'formation' is only a exemple, but this method could be interresting for other abilities.

PR #6313, opened 2021-11-19
Capitalise references to the Academy of Magic on Isle of Alduin

### LINKED ISSUE #6310: Revise Capitalization and referencing of the Academy of Magic in Alduin
In several scenarios and campaigns, the Academy of Magic in Alduin is referenced and featured. The way it is referenced and spelled and typed is inconsistent throughout.
Variations may include:
- Academy
- Academy of Magic
- great Academy
- Great Academy
- great Academy of Magic
- Great Academy of Magic

Just to name a few campaigns that have it we have:
- Secrets of the Ancients
- Heir to the Throne
- Delfador's Memoirs
- A Tale of Two Brothers

Wedge009:
Resolves #6310.

### DIFF
--- a/data/campaigns/Secrets_of_the_Ancients/scenarios/02_Dark_Business.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/scenarios/02_Dark_Business.cfg
@@ -760,7 +760,7 @@ of Healing"
         [/message]
         [message]
             speaker=Glant
-            message= _ "I don’t know, but this evening a mage from the academy was here talking to the mayor, my brother says."
+            message= _ "I don’t know, but this evening a mage from the Academy was here talking to the mayor, my brother says."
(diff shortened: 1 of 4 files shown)


knyghtmare (approved):
A change as minor as this would been better as a direct commit, but oh well.

Wedge009:
Have to let these be approved by nemaara plus better to err on the side of caution. I still think the underlying issue report is a non-issue but it's worth making these consistent with the rest of the prose.

rubyFeedback:
> A change as minor as this would been better as a direct commit, but oh well.

But others may disagree at some point in time, so IMO Wedge009's more
cautious approach seems better. See it more as a good thing - a small,
innocent, harmless change where nothing can break! :D

nemaara (approved):

Issue #6291, opened 2021-11-14
Regex processing during MP server connection takes excessively long time with CJK languages


Wedge009:
Reporting on @fujimo-t's behalf (at least I think it is their report) from #wesnoth-dev discussion at 2021-11-14 ~02:30-03:00 UTC.

### Game and System Information
Official Windows 1.16.1 release

### Describe the bug
Connection to MP server appears to hang when Japanese or Chinese language is selected. (2021-11-21 update: it just takes an excessively long time.) Seems only applicable to Windows build, official Linux release does not exhibit this problem.

### To Reproduce
1. Start Wesnoth 1.16.1 on Windows
2. Select Japanese language
3. Log-in to official MP server
4. Observe the connection screen remains indefinitely (does not even reach the enter credentials stage).

### Expected behaviour
Successful log-in to MP server in a 'reasonable' time (eg within a minute).

### Additional context
I have confirmed this, though have not yet tested any versions besides official 1.16.1 releases, nor any other languages besides `en_GB`.

2021-11-21 update: appears to be an upstream bug in regex handling.

fujimo-t:
Thank you for open this issue.

I can reproduce it by self-build according to the [make_mingw_build](https://github.com/wesnoth/wesnoth/blob/master/utils/dockerbuilds/make_mingw_build) script but can't by Visual Studio + vcpkg.
(Both source is same 1.16.1 tag.) 

I built by the mingw with debug option and confirmed "Invalid parameter" messages repeated many time while the hanging.
[invalid-parameter-messages.txt](https://github.com/wesnoth/wesnoth/files/7533178/invalid-parameter-messages.txt)

Then I set a breakpoint on _invalid_parameter, It seems the message is from std::regex constructor in src/utils/parse_network_address.cpp.
[backtrace.txt](https://github.com/wesnoth/wesnoth/files/7533183/backtrace.txt)

Pentarctagon:
What's the value of the `address` argument when this is happening?

fujimo-t:
From backtrace.txt:
> #26 0x000000000073abd5 in parse_network_address (address="add-ons.wesnoth.org", default_port="15016")
    at /wesnoth/src/utils/parse_network_address.cpp:24

It seems no problem.

Wedge009:
I confirm 1.16.1+dev build compiled with MSVC does not exhibit the issue. I even tested both with and without generated `mo` translation files, both log-ins were successful.

[...]

PR #6316, opened 2021-11-21
Replace std::regex by boost to temporary workaround for Mingw-w64's bug

fujimo-t:
Temporary workaround for #6291.

It is an upstream bug but seems to take a long time to fundamental resolve, such as fix the bug at the upstream or move release toolchain to MSVC.

### DIFF
--- a/src/editor/palette/location_palette.cpp
+++ b/src/editor/palette/location_palette.cpp
@@ -29,7 +29,7 @@
 
 #include "formula/string_utils.hpp"
 
-#include <regex>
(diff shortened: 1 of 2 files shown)


fujimo-t:
Also this is prevent #5231 until fundamental resolve.

Wedge009:
Even if this ends up being a long-term 'temporary' solution, as long as it maintains functionality I'd be in favour of it.

Pentarctagon (approved):

PR #6321, opened 2021-11-22
CMake: run xgettext on the .cpp file with GETTEXT_DOMAIN "wesnoth"

stevecotton:
Make the search for translatable strings include install_dependencies.cpp in the POTFILES.in file.

Exactly one (plural) string has been missing from .pot files build by CMake, presumably since 2017.

### DIFF
--- a/cmake/update_pot_source_dependencies.cmake
+++ b/cmake/update_pot_source_dependencies.cmake
@@ -20,23 +20,29 @@ if(DOMAIN STREQUAL ${DEFAULT_DOMAIN})
 	add_custom_command(
 		OUTPUT ${PROJECT_SOURCE_DIR}/po/${DOMAIN}/POTFILES.in
 
-		# Create an empty new one, to be sure it will exist.
-		COMMAND ${CMAKE_COMMAND}
-				-E touch ${PROJECT_SOURCE_DIR}/po/${DOMAIN}/POTFILES.in
-
-		# Find all cpp files which are not in a .git directory.
+		# Find all cpp files which are not in a .git directory, check their
+		# textdomain, and write the list of matching files to POTFILES.in.
 		COMMAND find src  -name .git -prune -o -name '*.[hc]pp' -print |
 				sort |
 				while read file\; do
 					# If the file doesn't contain a GETTEXT_DOMAIN
 					# definition it should be added to the default domain.
-					if ! grep \"^\#define  *GETTEXT_DOMAIN\"
+					if ! grep '^\#define  *GETTEXT_DOMAIN'
 							$$file > /dev/null 2>&1\; then
 
-						echo $$file >>
-							${PROJECT_SOURCE_DIR}/po/${DOMAIN}/POTFILES.in \;
+						echo $$file \;
+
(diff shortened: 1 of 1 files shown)


cmake/update_pot_source_dependencies.cmake
```diff
@@ -35,6 +35,15 @@ if(DOMAIN STREQUAL ${DEFAULT_DOMAIN})
 
 						echo $$file >>
 							${PROJECT_SOURCE_DIR}/po/${DOMAIN}/POTFILES.in \;
+
+					# While files don't need a GETTEXT_DOMAIN to be included in
+					# the default domain, accept files that contain a
+					# GETTEXT_DOMAIN definition for the default domain too.
+					elif grep \"^\#define  *GETTEXT_DOMAIN  *\\\"${DOMAIN}\\\"\"
```
soliton-, line 42:
```suggestion
					elif grep '^#define  *GETTEXT_DOMAIN  *\"${DOMAIN}\"'
```
It seems to me like this should work as well and is more readable.

stevecotton: The backslash before the `#` was definitely necessary, but I took the rest of the suggestion.

soliton-: Ah, thanks for checking. Must be cmake syntax then since neither shell nor grep care.

cmake/update_pot_source_dependencies.cmake
```diff
@@ -30,13 +30,22 @@ if(DOMAIN STREQUAL ${DEFAULT_DOMAIN})
...
+					# While files don't need a GETTEXT_DOMAIN to be included in
+					# the default domain, accept files that contain a
+					# GETTEXT_DOMAIN definition for the default domain too.
+					elif grep '^\#define  *GETTEXT_DOMAIN  *\"${DOMAIN}\"'
+							$$file > /dev/null 2>&1\; then
+
+						echo $$file \;
 					fi
-				done
+				done >>
```
soliton-, line 46: Presumably it can be a non-append redirection now? Or is the file expected to be there already?

stevecotton: True, it could be. The file's created by a `touch` command in this cmake file.

I'm not sure if we'd need to worry about Bash's `noclobber` option, which makes it an error to `>` to an already-existing file.

soliton-: Right, just before this command. Since now there is no chance the file won't exist that is probably useless.

Based on the touch command `noclobber` is clearly not on so far and this should not read the user's .bashrc or so either.

stevecotton: I don't understand. Why does the touch command tell us anything about `noclobber`?

soliton-: Sorry, I misunderstood. I'd be surprised if `noclobber` is enabled by default anywhere. That'd break lots of scripts. If you're really worried you could make it `>|`.

stevecotton: Changed it to that, and removed the `touch` lines.

PR #6322, opened 2021-11-22
wmlxgettext: sort-by-filename before processing files

### LINKED ISSUE #5989: wmlxgettext: unnecessary reordering of `#.` lines
Each pot-update has many changes that are simply `#.` comment lines being reordered, it would be better if the tools had a stable sort of these lines. In contrast, the `#:` lines do seem to have a stable sort.

An example from the pre-1.15.15 pot update (df6e679f136142db1e012cff1aca64294b41086a):
```
@@ -4491,10 +4491,10 @@ msgstr ""
 msgid "Enemies"
 msgstr ""
 
-#. [side]: id=King Thurongar, type=Dwarvish Lord
-#. [side]: id=Melusand, type=Mermaid Diviner
 #. [side]
+#. [side]: id=King Thurongar, type=Dwarvish Lord
 #. [side]: id=Great Leader Darmog, type=Great Troll
+#. [side]: id=Melusand, type=Mermaid Diviner
 #: data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg:73
 #: data/campaigns/Under_the_Burning_Suns/scenarios/07a_Dealing_with_Dwarves.cfg:52
 #: data/campaigns/Under_the_Burning_Suns/scenarios/07b_Talking_with_Trolls.cfg:53
```

stevecotton:
Add a sort inside wmlxgettext so that the order of filenames on the command
line doesn't affect the output. The sort is run unconditionally, because I
don't see a usecase for not enabling it. For consistency with xgettext and
msgcat, if there was an option to enable it then it would be called
`--sort-by-file`.

The aim of this change is to reduce the number of changes in .po files caused
by a pot-update. When multiple comments refer to the same translatable string,
separate runs of pot-update could generate them in a random order. If I've
understood the root cause correctly, wmlxgettext itself was deterministic, but
the order of the files on wmlxgettext's command line was not deterministic.

For testing, I tried reversing the sort order; the resulting reordering hit
lines that commonly swap order during pot-updates.

This affects the ordering of the lines beginning `#.` that are generated from
`#po:` comments and also those auto-generated from `[message]speaker=`, etc.

@AncientLich, it's only a one-liner, but it seems polite to ask you first.

Fixes #5989.

### DIFF
--- a/data/tools/wmlxgettext
+++ b/data/tools/wmlxgettext
@@ -211,6 +211,7 @@ def main():
     # if args.recursive is True and args.filelist is None:
     else:
         startPath, filelist = pywmlx.autof.autoscan(startPath)
+    filelist.sort()
     for fileno, fx in enumerate(filelist):
         fname = os.path.join(startPath, os.path.normpath(fx))
         is_file = os.path.isfile(fname)


stevecotton:
I'm planning to merge on 30th Nov (if not earlier), so that @max-torch's #6338 can take over the modification of this file.

max-torch:
> I'm planning to merge on 30th Nov (if not earlier), so that @max-torch's #6338 can take over the modification of this file.

In any case, this is easy enough to copy in if i encounter merge conflicts.

cooljeanius:
This change should probably be noted somewhere on the wiki, if it hasn't already been

stevecotton:
There doesn't seem to be much detail of wmlxgettext on the Wiki. Short of adding full documentation there, the only change that I can see would be to add "Use the latest version, it stable-sorts the files to create smaller diffs" to https://wiki.wesnoth.org/GettextForWesnothDevelopers .

PR #6323, opened 2021-11-22
Luacheck

Pentarctagon:


.luacheckrc
```diff
@@ -0,0 +1,17 @@
...
+max_line_length=false
+max_code_line_length=false
+max_string_line_length=false
+max_comment_line_length=false
+-- show the warning/error codes as well
+codes=true
+-- don't show files with no issues
+quiet=1
+-- don't show undefined variable usage
+-- there are thousands of warnings here because luacheck is unaware of Wesnoth's lua envvironment and has no way to check which have been loaded
```
stevecotton, line 11: Spellcheck "envvironment"

data/ai/lua/ca_village_hunt.lua
```diff
@@ -60,16 +60,16 @@ function ca_village_hunt:evaluation(cfg, data, filter_own)
         return 0
     end
 
-    local avoid_map = AH.get_avoid_map(ai, nil, true)
+    local avoid_map2 = AH.get_avoid_map(ai, nil, true)
```
Wedge009, line 63: Without knowing anything about what these changes are doing, is there something more descriptive we can use than just appending numbers? Is it just to make the variables unique and for some reason we cannot re-use existing variables?

Pentarctagon: The warning being fixed is about a variable shadowing another, which isn't good both for readability as well as if you remove the second declaration of the variable then all the code that was previously using it will now silently start using the firs declaration rather than warning about using an undefined variable.

Wedge009: Hmm, okay. Generally I try to avoid calling things `x1`, `x2`, `x3`, etc, but I suppose it'd be a bit much to try to give everything sensible names here.

cooljeanius:
This is probably outside the scope of this PR, but I'm wondering if it would be possible to have `wmllint` run `luacheck` on any lua files it finds in an add-on (if `luacheck` is available) when a UMC author runs it on their add-on directory?

data/ai/lua/ca_village_hunt.lua
stevecotton: I think it would be better to have the config for luacheck, but for the moment have it as something that developers (backed up by code-reviewers) run manually, with a long-term goal of looking carefully at all of the warnings.

With the code that we're commenting on, I think both `avoid_map1` and `avoid_map2` will have the same contents:
```
    local avoid_map1 = LS.of_pairs(ai.aspects.avoid)
    local avoid_map2 = AH.get_avoid_map(ai, nil, true)
```
but that `get_avoid_map(ai, nil, true)` appears to also return `LS.of_pairs(ai.aspects.avoid)`.

stevecotton:
> I'm wondering if it would be possible to have `wmllint` run `luacheck` on any lua files it finds in an add-on (if `luacheck` is available) when a UMC author runs it on their add-on directory?

Maybe give it a separate tab in the maintenance tools' GUI.pyw? `luacheck` will itself already search a directory recursively for files, so it's easy to run from the command line too.

data/ai/lua/ca_village_hunt.lua
Pentarctagon: I don't see how those are related? The goal here isn't to evaluate every single one of hundreds of warnings to determine whether the current code isn't ideal in some way, it's to address the existing warnings so that the check can be run in the CI to prevent even more of the same types of problems from being added.

data/ai/lua/ai_helper.lua
```diff
@@ -516,7 +516,10 @@ end
 function ai_helper.split(str, sep)
     -- Split string @str into a table using the delimiter @sep (default: ',')
 
-    local sep, fields = sep or ",", {}
```
soliton-, line 519: Is the issue here really the `or` or that it is done together with the comma operator?

Pentarctagon: The issue was that it was declaring `local sep` which shadowed the `sep` function argument.

data/ai/lua/ca_village_hunt.lua
stevecotton: You're right - I was imagining a dream-world where everyone rigorously opened PRs for all changes.

data/ai/micro_ais/cas/ca_hunter.lua
```diff
@@ -60,8 +60,7 @@ function ca_hunter:execution(cfg)
     -- If hunting_status is not set for the hunter -> default behavior -> random wander
     if (not hunter_vars.hunting_status) then
         -- Hunter gets a new goal if none exist or on any move with 10% random chance
-        local rand = math.random(10)
-        if (not hunter_vars.goal_x) or (rand == 1) then
+        if (not hunter_vars.goal_x) or (math.random(10) == 1) then
```
stevecotton, line 63: Lua uses short-circuit evaluation [confirmation](http://www.lua.org/manual/5.4/manual.html#3.4.5). Could the reordering and short-circuit here cause OOS in replays?

data/campaigns/Descent_Into_Darkness/scenarios/09_Descent_into_Darkness.cfg
```diff
@@ -34,7 +34,7 @@
             {STORYTXT_BACKGROUND book.jpg}
         [/part]
         [part]
-            story= _ "So reads the book that Malin has reclaimed. <i>“The spells of necromancy can free the spirit from the limitations of the flesh, but only once the soul has been unbound from the body. The necromancer must make the necessary incantations with his dying breaths, then conquer his own spirit much the same way he binds the spirits of others. Because he retains his own will, however, the lich can call upon the awesome powers of the spirit world.”</i>"
+            story= _ "So reads the book that Malin has reclaimed. <i>“The spells of necromancy can free the spirit from the limitations of the flesh, but only once the soul has been unbound from the body. The necromancer must make the necessary incantations with his dying breaths, then conquer his own spirit in much the same way he binds the spirits of others. Because he retains his own will, however, the lich can call upon the awesome powers of the spirit world.”</i>"
```
stevecotton, line 37: Caught a necromancer trying to sneak in with the moon-language changes!

cooljeanius: wait this is something that `luacheck` caught?

data/campaigns/Under_the_Burning_Suns/units/quenoth/Tauroch.cfg
```diff
@@ -15,7 +15,7 @@
     advances_to=null
     cost=12
     usage=fighter
-    description= _ "Taurochs are huge, hardened beasts of the desert, considered untamable by most surviving races. However, the elves’ affinity with nature has allowed them to form an unlikely bond with these unruly animals. While Taurochs can easily carry or pull heavy loads, their thick hides and unwavering resolution also make them formidable mounts in battle."
+    description= _ "Taurochs are huge, hardened beasts of the desert, considered untamable by most surviving races. However, the elves’ affinity with nature has allowed them to form an unlikely bond with these unruly animals. While taurochs can easily carry or pull heavy loads, their thick hides and unwavering resolution also make them formidable mounts in battle."
```
stevecotton, line 18: Commit c4492327306cfdd81c34848219e57cb738868788 also seems to be here in error.

data/campaigns/Descent_Into_Darkness/scenarios/09_Descent_into_Darkness.cfg
Wedge009: What? This is something I did in 3922d04b446f9f4bb06441ca4644adf25b35fc69, to revert a mistake in 19a42e3ead717dc1808162168585abe8755df4a4. Maybe Pent's branch isn't forked from current master?

data/campaigns/World_Conquest/lua/map/postgeneration/2A_Springs.lua
```diff
@@ -155,7 +155,7 @@ function world_conquest_tek_map_decoration_2a()
 	while #terrain_to_change > 0 and mathx.random(10) ~= 1 do
 		local loc = terrain_to_change[mathx.random(#terrain_to_change)]
 		map[loc] = "^Efm"
-		local terrain_to_change = wct_store_possible_flowers("G*^Fet")
+		terrain_to_change = wct_store_possible_flowers("G*^Fet")
```
stevecotton, line 158: This looks like a bugfix, I think it will change the behavior and needs testing.

data/campaigns/Descent_Into_Darkness/scenarios/09_Descent_into_Darkness.cfg
Pentarctagon: I'm really confused with what git thinks it's doing right now.

Pentarctagon: Should be good now.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Tauroch.cfg
Pentarctagon: Also this should be fixed as well.

data/ai/micro_ais/cas/ca_hunter.lua
Pentarctagon: The first condition would either be evaluated to true or false the same in the initial play-through vs the replay, so it shouldn't?

data/campaigns/Descent_Into_Darkness/scenarios/09_Descent_into_Darkness.cfg
Wedge009: Yep.

data/campaigns/World_Conquest/lua/map/postgeneration/2A_Springs.lua
Pentarctagon: What would be the test case here?  The only difference now is that it will exit the loop early if there's no terrain left to change rather than waiting for `mathx.random(10)` to return 1.

data/ai/micro_ais/cas/ca_hunter.lua
stevecotton: You're right, I was confused. Couldn't happen on a stable branch.

Faulty logic: while the WML would be loaded from the replay, the Lua code is probably read from the game data instead. But that's broken logic, as long as this is only on a development branch when getting a new version is expected to OOS some replays.

data/campaigns/World_Conquest/lua/map/postgeneration/2A_Springs.lua
stevecotton: Ah yeah, didn't realise how little it could change.

data/ai/lua/ai_helper.lua
soliton-: I see. Can't say the new code is more readable personally.

Since `sep` is only used once I'd just put the `or` there. Or if you want to keep the assignment just don't use local?

Pentarctagon: Alright, changed back to being a single line.

Pentarctagon:
Is anyone still reviewing this, or are there any objections to merging?

The only remaining thing I'm not 100% certain about is whether it makes sense to ignore all undefined global variable warnings or not.  The current output of not ignoring those is https://pastebin.com/YeWK2vin - a large number of the warnings are about things that are clearly actually present and could be marked as ignored explicitly (such as `accessing undefined variable wesnoth`), but there are also a ton of WC and map generator specific things, as well as many instances of `accessing undefined variable f` which I'm assuming is a convention and actually does work given how widely it's used.

Wedge009:
Would be nice not to have those warnings but unless someone takes the time to analyse them all it may be better to leave it as-is for now. Assuming this change provides the desired improvement, maybe merge it and note an issue for the remaining warnings.

Pentarctagon:
Partially I'm also now just curious about what this `f` thing that's used everywhere even is. Is it the same thing being accessed in a bunch of different places, or a bunch of different things? What functionality is attached to it? Usually I'd expect something called `f` to be some sort of short lived function object/pointer/etc rather than something global(?) and long-lived. Probably going to take a look at that later.

stevecotton:
I think your answer is in eb1bd1af6cfa3e510e9ef5fae72f966567059969, although that still leaves it as a global.

Pentarctagon:
Alright, well, for my next PR I think I'm going to just mass replace `f` with `wesnoth.map.filter_tags`.  Or more likely do `filter_tags = wesnoth.map.filter_tags` and replace `f` with `filter_tags`. Having that as a global variable called `f` is bizarre.

Wedge009:
Unless it's something really obvious and conventional, like `i=0` in a `for` loop, I prefer to avoid abbreviating variable names so that sounds like a good idea. Readability > brevity, in my opinion.

stevecotton:
> Alright, well, for my next PR I think I'm going to just mass replace `f` with `wesnoth.map.filter_tags`. Or more likely do `filter_tags = wesnoth.map.filter_tags` and replace `f` with `filter_tags`. Having that as a global variable called just `f` is bizarre.

I suggest delegating that to @knyghtmare, or at least I suggest such a code cleanup should be part of the work to add four-player support.

Pentarctagon:
The 4 player mode is already present in the master branch.

stevecotton:
> Is anyone still reviewing this, or are there any objections to merging?

Just finishing looking through the files, and it all looks good (or at least an improvement of code with existing problems) to me.

CelticMinstrel:
> Alright, well, for my next PR I think I'm going to just mass replace `f` with `wesnoth.map.filter_tags`. Or more likely do `filter_tags = wesnoth.map.filter_tags` and replace `f` with `filter_tags`. Having that as a global variable called just `f` is bizarre.

I think my suggestion would be:

* Change it to an uppercase F - that better matches with other places that use shortened variable names.
* Declare it as a local in every file that uses it, instead of making it a global.

vgaming:
As a short-term solution, I'd propose to take a look at per-file and per-path overrides: https://luacheck.readthedocs.io/en/stable/config.html#per-file-and-per-path-overrides

This way you can essentially allow the `f` global on those specific files for now, and otherwise enable the check in `luacheck`. Once we address the `f` problem, we can remove the exclusion rule

PR #6324, opened 2021-11-22
fix 'add' and 'sub' not added in leadership shema validation

newfrenchy83:
Like add and 'sub'can be used in [leadership] abilities  now, it must add that to schema validation


stevecotton:
Do we know when `add` and `sub` were added to leadership, and if it was added to other abilities too?

newfrenchy83:
> Do we know when `add` and `sub` were added to leadership, and if it was added to other abilities too?

since 1.15.0 the 2 jan 2019 with https://github.com/wesnoth/wesnoth/commit/7114174789e850bbd7829a353e7d885791c0a664#diff-eb9c8f8b8122951ad515f39be0e4d8550c4f22421024154f55bdef508de85897

newfrenchy83:
'add','sub',mutiply', and 'divide' already adde to [resistance] since thee beginning,

stevecotton:
Do `multiply`, `divide` and `max_value` also apply to leadership?

newfrenchy83:
> Do `multiply`, `divide` and `max_value` also apply to leadership?

'max_value', and for multiply and divide, only if base value different of 0, and this is the case only when a [leaderhip] with multiply is on same list what a regular leadership with value not equal 0, else 'value' equals zero, and 0 * 'multiply or 0/'divide' = 0. this why i add only 'add' and 'sub'  who can be used same if not regular leadership in the list.
For 'multiply'and 'divide' i could said 'yes' but it is under condition, and it is not necessary, i prefer said no.

Pentarctagon:
I think multiply and divide should be added, since they are valid attributes in every case except someone trying to multiply or divide by 0 for whatever reason.

newfrenchy83:
> I think multiply and divide should be added, since they are valid attributes in every case except someone trying to multiply or divide by 0 for whatever reason.

Done

newfrenchy83:
Someone could review this PR, (and the 1.16 equivelent) before the enxt release, please?

data/test/scenarios/test_add_or_sub_in_leadership.cfg
```diff
@@ -0,0 +1,153 @@
+# This unit test is used for verify the work of abilities used like weapons, but also of special_id_active attribute.
```
stevecotton, line 1: It's good to test both things, but it would be even better to have 2 simpler separate tests instead of a single combined test.

For the leadership test, it seems cleaner to remove the filters. Just keep the `[attacks]`, `[damage]` and `[chance_to_hit]` to make the test deterministic.

For testing `special_id_active`, it should be on something like `[attacks]` or `[damage]` where the test would reliably fail. When it's on `[chance_to_hit]`, there's a chance the RNG will make the test pass even if the `special_id_active` code is broken.

stevecotton (changes requested):
Most of these are changes that I'd like but which it would still be OK without. The hard "changes requested" is for the "there's a chance the RNG will make the test pass even if the code is broken" bit.

data/test/scenarios/test_add_or_sub_in_leadership.cfg
newfrenchy83: I is done, (i hope!)

data/test/scenarios/test_add_or_sub_in_leadership.cfg
```diff
@@ -0,0 +1,183 @@
...
+                        affect_self=yes
+                    [/leadership]
+                    [attacks]
+                        value=1
+                    [/attacks]
+                    [damage]
+                        value=60
+                    [/damage]
+                    [chance_to_hit]
+                        id=test_cth
```
stevecotton, line 31: This `test_cth` isn't needed any more.

data/test/scenarios/test_add_or_sub_in_leadership.cfg
```diff
@@ -0,0 +1,183 @@
...
+                        value=100
+                    [/chance_to_hit]
+                [/abilities]
+            [/effect]
+            [filter]
+                id=alice
+            [/filter]
+        [/object]
+        # Place Alice for fighting
```
stevecotton, line 40: This is a comment about the lines that follow it, but it has a blank-line between the comment and the code that's being commented on, combined with no blank line between it and the preceding code. That makes it harder to understand the code.

This comment applies to other parts of this code too.

data/test/scenarios/test_add_or_sub_in_leadership.cfg
stevecotton: The comment still says that this tests the `special_id_active` attribute, but I think that has already been removed. I think removing it was good, but the comment needs to be updated too.

data/test/scenarios/test_add_or_sub_in_leadership.cfg
```diff
@@ -0,0 +1,183 @@
...
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 25})}
+        {SUCCEED}
+    [/event]
+)}
+# wmlindent: opener "{GENERIC_UNIT_TEST "
+# wmlindent: closer ")}"
```
stevecotton, line 92: These aren't needed - wmlindent seems to handle it without them.

data/test/scenarios/test_add_or_sub_in_leadership.cfg
```diff
@@ -0,0 +1,153 @@
...
+                id=alice
+            [/filter]
+        [/object]
+
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+            kill=yes
```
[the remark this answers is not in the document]
stevecotton: This has been marked resolved, but doesn't seem to be resolved, please would you add a comment (in the PR review, not the code) about it.

The code for moving Alice is still mixed with storing Bob, so the `# Fight` section still depends on a variable stored in the middle of the storing-Alice code.

stevecotton:
Oops, that was meant to be a squash-and-merge.

PR #6327, opened 2021-11-22
remove Utbs 'formation' name of attack prediction windows(master)

newfrenchy83:
Like description of Utbs formation abilities is opposite of what suggest name in attack prediction windows, it best to remove that.


Pentarctagon:
Just to confirm, what does the UI look like with this changed back?

newfrenchy83:
'formatio

> Just to confirm, what does the UI look like with this changed back?

'formation' don't appear in attack prediction windows, because name implemented in [dummy] abilitie and not in abilities specials.

It is this resolve or remove code in for all abilities like weapon applied to opponent with https://github.com/wesnoth/wesnoth/pull/6318 and https://github.com/wesnoth/wesnoth/pull/6317

PR #6328, opened 2021-11-23
CMake: update the .po files before generating stats about them

### LINKED ISSUE #6320: CMake pot-update calculates data/languages stats before generating .pot files
Version 1.16.1+dev (decc9644023391d6726b8b65b0581d656c8b578a), building on Linux

The files in `data/languages/` include the stats for which translations are complete enough to be shown in the GUI without using the `--all-translations` argument. These stats have to be calculated from the .po files.

In a CMake-based build, running `make pot-update` updates those stats first, then it generates the updated .pot and .po files. Thus it has to be calculating the old stats, not the current ones.

Logging now because it's a diversion to a diversion to the issue that I'm currently working on. Very low priority, because it probably never matters; it can only calculate a higher-than-actual false value (unless the new .pot files bring back in old strings).

stevecotton:
Fixes #6320.

About whether to backport it: it's not necessary for 1.16, but it's a bugfix that I'd test along with #6321, and it's a bug that caused unexpected differences between SCons and CMake when testing #6322, so I think it's worth backporting all three PRs together.

### DIFF
--- a/po/CMakeLists.txt
+++ b/po/CMakeLists.txt
@@ -46,6 +46,10 @@ set(WMLXGETTEXT ${PROJECT_SOURCE_DIR}/data/tools/wmlxgettext)
 
 if(ENABLE_POT_UPDATE_TARGET)
 
+	# The variable pot-update-SRC accumulates a list of the dummy-targets
+	# declared in the following loops. Depending on it will cause all of the
+	# custom commands to run.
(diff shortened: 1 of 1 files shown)


soliton- (approved):
Looks good to me based on your comments (not that I know much about cmake).

Issue #6332, opened 2021-11-24
Sceptre of Fire final scenario: Handling the victory condition


Wedge009:
I originally played Sceptre of Fire all the way back in 1.4, though it looks to me like the campaign revision was only done quite recently in 1.15.x development cycle. In particular, the final scenario 'Caverns of Flame' was rewritten in #4681.

I recently replayed the campaign in 1.16.1 release but had particular trouble with the objectives for the final scenario. To release the seal on the volcano, three runic tablets need to be destroyed:
1. The first one is destroyed by the orc leader when enough of his team (or the leader himself) is killed. Thankfully the invading elves seem willing to help you complete this unless the player's side has not moved sufficiently far away enough.
https://github.com/wesnoth/wesnoth/blob/b8400b5b722a4fd022b6eabbcfb2f68bbfac7b28/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg#L467-L471
2. The second is destroyed when defeating the Troll Hero, something that seems reasonable given this team is the weakest of all the enemies.
https://github.com/wesnoth/wesnoth/blob/b8400b5b722a4fd022b6eabbcfb2f68bbfac7b28/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg#L571-L575
3. However, I had particular trouble with the final one as it requires Krawg - and only Krawg - to find the runic tablet.
https://github.com/wesnoth/wesnoth/blob/b8400b5b722a4fd022b6eabbcfb2f68bbfac7b28/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg#L660-L663

The problem I had was that I used a generic Gryphon Rider to enter that hex, since my Krawg was severely wounded by the harassing Cave Wyrmlets near the start. Nothing happened upon reaching the tablet, so I assumed real tablet was somewhere else. Eventually I had to dig through the code to find out that the trigger for the final runic tablet was Krawg alone, Gryphon Riders are not considered.

[...]

PR #4681, opened 2020-01-01
SoF - Rewrite of last scenario

doofus-01:
This doesn't change the plot in any significant way, but it should be a more credible way to end things.  There are two concerns I have:
1.  It may be too easy.  Or it may require too much "behind the scenes knowledge".  I don't know how to test that aside from getting other people to play it.
2.  The micro_ai for the special orc didn't work at all.  I don't know if it is because I was using it wrong or there was a conflict with the new filtering*. I've got a work around, but it should be replaced when possible.

*(https://wiki.wesnoth.org/Wesnoth_AI_Framework#The_.5Bcandidate_action.5D_Tag)

PR #6334, opened 2021-11-25
SoF Finale: Allow for Gryphon Riders to trigger an event on the final runic tablet

Wedge009:
Draft: needs actual dialogue that makes sense from a story perspective, explaining why it has to be Krawg to destroy the final tablet.

Relates to #6332.

Either Krawg or a Gryphon Rider/Master may trigger the first event, finding the tablet before the other two are destroyed. After they have been destroyed, Krawg and a Gryphon Rider/Master may trigger their respective event independently, though of course if Krawg goes first that's the end of the scenario.

Alternative is to trigger victory irrespective of who reaches the final tablet, but then dialogue needs to be rewritten for a Gryphon Rider/Master instead of Krawg.

### DIFF
--- a/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
@@ -658,7 +658,6 @@
         [event]
             name=moveto
             [filter]
-                id=Krawg
                 x,y=29,31
             [/filter]
             [filter_condition]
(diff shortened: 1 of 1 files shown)


stevecotton:
This has made me realise that there's another plot hole - why is Krawg the only survivor, when there are other gryphons?

Wedge009:
I also wondered about that. There's no reason why other Gryphon Riders can't have escaped... but for whatever reason only Krawg chooses to chase after Alanin.

I didn't feel like Gryphons have that great movement inside the caves, certainly not as good as in the hills and mountainside earlier on.

max-torch:
It seems like it makes a lot more sense for any Gryphon to be capable of getting the last rune. It's the least convoluted on behalf of the player experience. We can certainly make the dialog state that all Gryphons escape, and then future generations of gryphon actually know about the sceptre, which is partly why they can end up helping Konrad in HtTT. And then make it so that it is just Krawg who chooses to deliver Alanin back to Wesnoth only because they became friends, the other gryphons just go back to their nests.

max-torch:
Just realized that the other gryphons apart from Krawg have dwarf riders on them, so if we let those gryphons escape those dwarves also escape. I think thematically no dwarf is really meant to escape. What do you all think about saying that Krawg was the only gryphon that was fast enough to make it out of the volcano before being engulfed by the hot lava and gas because he was the only one that didn't have the burden of carrying a dwarf and supplies on his back?

Wedge009:
Yes, I think that's the conundrum: The story as a whole seems to be that the entire dwarf clan commissioned to create the Sceptre of Fire perish. At the moment, I can't remember how exactly the scenario was before doofus revised it in #4681 but I recall it was a lot more complex - I think basically everyone gets wiped out by the hordes of elves and orcs attacking them. Krawg was still the only one to escape, but I don't think it was done through the rune-breaking which avoided the potential situation of allowing (from a physical perspective) of Gryphon Riders to escape too.

In terms explaining how only Krawg escapes, I think the idea that only he is the one unencumbered and therefore fast enough to escape is plausible. It's certainly better than any other explanation I can think of or have seen so far.

Wedge009:
Closing in favour of #6628.

PR #6335, opened 2021-11-26
Expand reduce function to both fold and reduce

luther9:
The file `data/lua/functional.lua` contains a `reduce` function which, when receiving a falsey `identity` argument, changes it to `0`. This makes it impossible to initialize the calculation with `nil` or `false`. `functional.reduce(t, 'or', false)` will always return `0`.

While changing the function to accept any 3rd argument, I decided to rewrite it to `fold`, which always includes the init value in the calculation, and `reduce`, which only looks at the identity value if the array is empty.

Since `functional.lua` is not documented on the wiki, I wonder if maybe this change can be merged with the stable branch.

### DIFF
--- a/data/lua/functional.lua
+++ b/data/lua/functional.lua
@@ -106,15 +106,40 @@ local known_operators = {
 	['or'] = function(a, b) return a or b end,
 }
 
-function functional.reduce(input, operator, identity)
+-- If operator is a string, return the corresponding function. Otherwise, return
+-- operator.
+local function check_operator(operator)
 	if type(operator) == 'string' then
-		operator = known_operators[operator]
+		return known_operators[operator]
 	end
-	local result = identity or 0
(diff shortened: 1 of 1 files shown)


Pentarctagon:
Did you mean to make this an issue rather than a pull request?

luther9:
No, I've made a patch. Why wouldn't that be a pull request?

Pentarctagon:
Ah, I was confused about there originally being hundreds of other commits listed for some reason.

Pentarctagon:
Also worth noting that new API features can't be added to a stable branch, so this would need to be made against the master branch.

luther9:
Ok, I guess I'll have to check out master and redo this later then.

luther9:
The new version of this PR is #6337. That one is based on master, not 1.16.

PR #6336, opened 2021-11-26
Ardonna should only call her own bats 'my pet'

Wedge009:
It's possible, however unlikely, for none of Ardonna's bats to have died before killing the hostile bats in 'Following the Shadow'. In such a case, when the 'You can now raise bat corpses!' event triggers, Ardonna will still refer to the enemy bat as her 'pet' so add this condition to avoid it.


stevecotton:
Please copy something from the PR's description into either the commit message or the WML. The current comment makes me think "how could that happen?", it's good to have the answer easily available from Git.

PR #6337, opened 2021-11-26
Expand the reduce function to both fold and reduce

luther9:
The file `data/lua/functional.lua` contains a `reduce` function which, when receiving a falsey `identity` argument, changes it to `0`. This makes it impossible to initialize the calculation with `nil` or `false`. `functional.reduce(t, 'or', false)` will always return `0`.

While changing the function to accept any 3rd argument, I decided to rewrite it to `fold`, which always includes the init value in the calculation, and `reduce`, which only looks at the identity value if the array is empty.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -14,4 +14,4 @@ global=false
 unused=false
 -- excluded files due to using lua 5.4 syntax that currently gets flagged as a syntax error
 -- clear out once a newer Ubuntu LTS base is used for our docker images, which would then also have a newer luacheck available
-exclude_files={"data/lua/core/wml.lua","data/lua/wml-flow.lua","data/lua/wml/find_path.lua","data/lua/wml/harm_unit.lua","data/lua/wml/modify_unit.lua","data/lua/wml/random_placement.lua"}
+exclude_files={"data/lua/core/wml.lua","data/lua/wml-flow.lua","data/lua/wml/find_path.lua","data/lua/wml/harm_unit.lua","data/lua/wml/modify_unit.lua","data/lua/wml/random_placement.lua","data/lua/functional.lua"}
--- a/data/lua/functional.lua
+++ b/data/lua/functional.lua
@@ -126,15 +126,40 @@ local known_operators = {
 	['or'] = function(a, b) return a or b end,
 }
 
-function functional.reduce(input, operator, identity)
+-- If operator is a string, return the corresponding function. Otherwise, return
+-- operator.
+local function check_operator(operator)
 	if type(operator) == 'string' then
-		operator = known_operators[operator]
+		return known_operators[operator]
 	end
-	local result = identity or 0
-	for _, v in ipairs(input) do
-		result = operator(result, v)
+	return operator
+end
+
+-- Fold starting from the given index.
+local function fold_from_i(t, f, init, i)
+	local value <const> = t[i]
+	if value == nil then
+		return init
 	end
-	return result
+	return fold_from_i(t, f, f(init, value), i + 1)
+end
+
+-- Fold the elements of array t into a single value. operator is called as
+-- 'operator(accumulator, element)' for every element in t, where init is the
+-- initial accumulator. operator may be a function or a binary Lua operator as a
+-- string.
+function functional.fold(t, operator, init)
+	return fold_from_i(t, check_operator(operator), init, 1)
+end
+
+-- Like fold, but if t is not empty, ignore the identity argument. The first
+-- operator call will be on the first two elements.
+function functional.reduce(t, operator, identity)
+	local first <const> = t[1]
+	if first == nil then
+		return identity
+	end
+	return fold_from_i(t, check_operator(operator), first, 2)
 end
 
 function functional.take_while(input, condition)
--- a/data/lua/functional.lua
+++ b/data/lua/functional.lua
@@ -126,40 +126,28 @@ local known_operators = {
 	['or'] = function(a, b) return a or b end,
 }
 
--- If operator is a string, return the corresponding function. Otherwise, return
--- operator.
-local function check_operator(operator)
-	if type(operator) == 'string' then
-		return known_operators[operator]
-	end
-	return operator
-end
+-- Reduce the elements of array t into a single value. operator is called as
+-- 'operator(accumulator, element)' for every element in t. If a 3rd argument is
+-- provided, which may be nil, it will be used as the accumulator when calling
+-- operator on the first element. If there is no 3rd argument, the first
+-- operator call will be on the first two elements. If there is no 3rd argument
+-- and the array is empty, return nil. operator may be a function or a binary
+-- Lua operator as a string.
+function functional.reduce(t, operator, ...)
+	local f <const> = known_operators[operator] or operator
 
--- Fold starting from the given index.
-local function fold_from_i(t, f, init, i)
-	local value <const> = t[i]
-	if value == nil then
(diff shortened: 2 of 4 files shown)


Pentarctagon:
Unit tests for this functionality would be good to have - see https://github.com/wesnoth/wesnoth/blob/master/data/test/scenarios/lua_map_find.cfg for an example and https://wiki.wesnoth.org/LuaAPI/unit_test for the documentation - but let's wait to see what celticminstrel thinks of the API itself before creating those.

CelticMinstrel (commented):
Conceptually, this seems fine. However, "fold" and "reduce" are synonyms, so I think it's confusing to have both with slightly different semantics. You've also done this in a way that seems backwards-incompatible, as "reduce" now ignores the identity argument when previously it always used it.

Do you have a use-case for using the init value only if the array is empty?

luther9:
> However, "fold" and "reduce" are synonyms, so I think it's confusing to have both with slightly different semantics. You've also done this in a way that seems backwards-incompatible, as "reduce" now ignores the identity argument when previously it always used it.

They're not exactly synonyms. I figure it would be less confusing if we follow the same convention as other languages that have both fold and reduce. See [Scheme](https://www.gnu.org/software/guile/manual/html_node/SRFI_002d1-Fold-and-Map.html), [F#](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/lists#fold-and-scan-operations), [Kotlin](https://www.baeldung.com/kotlin/fold-vs-reduce). The reduce from F# and Kotlin doesn't have an init parameter, but I prefer the Scheme way, where it does have that parameter, as it's more versatile.

> Do you have a use-case for using the init value only if the array is empty?

There are a couple. Firstly, we might need to return a special value when the array is empty.

```lua
reduce(numbers, math.max)
```

Here, if the array is empty, it would be misleading to return a type of value that we expect to be in the array.

Secondly, there's performance. If the init value is an identity, it's useless to call the given function on init and the first element. Consider this code:

```lua
local function identity(x)
	return x
end

local function pipe(f, g)
	return function(x)
		return g(f(x))
	end
end

reduce(arrayOfFunctions, pipe, identity)
```

Here, the initial `pipe` call not only is unnecessary, it would add a function that does nothing to the resulting closure chain, so this might make a difference to performance.

And just to show we need *both* fold and reduce, there is a case where fold will work, but not reduce. It's basically any time the init value is not an identity:

```Lua
local function addNumToString(s, n)
	return s .. string.char(n)
end

fold(t, addNumToString, '')
```

CelticMinstrel:
> They're not exactly synonyms. I figure it would be less confusing if we follow the same convention as other languages that have both fold and reduce.

I don't think that's sufficient evidence to claim that they're not synonyms though… it's clear from the descriptions that the two functions do the same thing, just with slightly different input. On Wikipedia, "fold" and "reduce" both ultimately take you to the same page.

> Here, if the array is empty, it would be misleading to return a type of value that we expect to be in the array.

I can kinda understand this, though for your specific example I think it's sufficient to pass `-math.huge` (ie, negative infinity) as the initial value. Can you think of other examples, though? Given that the `stringx` module contains a number of functions that are essentially specialized variations of fold/reduce, some of which (eg `format_conjunct_list`) exhibit this property, I imagine most examples would come from string processing…?

> And just to show we need _both_ fold and reduce, there is a case where fold will work, but not reduce.

But (to use your terminology), is there any case where reduce will work, but not fold? Isn't the form you refer to as "fold" the most generic form of the algorithm?

Also, your example seems wrong – the init value in the example _is_ an identity

---

In any case, _if_ we were to include both functionalities under different names, I think they would need to be swapped for backwards-compatibility reasons – "reduce" takes an initial value, while "fold" does not. However, there's actually no need to have separate names, is there? The "reduce" function could use one or the other method based on whether an initial value was passed in the first place. So:

```lua
local values = {1, 2, 3, 4, 5}
local a = functional.reduce(values, '+') -- What you're calling reduce: `((((1 + 2) + 3) + 4) + 5)`
local b = functional.reduce(values, '+', 0) -- What you're calling fold: `(((((0 + 1) + 2) + 3) + 4) + 5)`
```

I do understand that you're unhappy about the case of an empty list in version `a`. Perhaps that could be handled by returning `nil` if the list is empty, so your preferred version of what you call "reduce" ends up something like:

```lua
local a = functional.reduce(values, '+') or empty_list_value
```

With the caveat of course that you need a separate if statement instead if the result of the reduce is a boolean value:

```lua
local a = functional.reduce(values, '+')
if a == nil then
  a = empty_list_value
end
```

I guess this could still create problems if `nil` was a possible output from the reduce for a non-empty list though… but even then it could be solved by checking if the list is empty before reducing:

```lua
local a
if #values == 0 then
  a = empty_list_value
else
  a = functional.reduce(values, '+')
end
```

Of course, there's also the option of making it throw if the list is empty – then your code would look something like this:

```lua
local success, a = pcall(functional.reduce, values, '+')
if not success then a = empty_list_value end
```

To sum up, I think the advantage of having a "default" argument in what you call "reduce" is minimal, as you can get the same effect easily with a couple of extra lines of code.

Also, I checked, and it _is_ possible to differentiate between the following two calls in Lua:

```lua
functional.reduce(values, fcn)
functional.reduce(values, fcn, nil)
```

So using an optional argument for the initial value doesn't rule out accepting `nil` as a valid initial value.

luther9:
I think we can agree to disagree on the first part of your post. The main issue seems to be that you don't want to provide two functions that are so similar to each other. I think I can work with that. The only question left is, if `reduce` is called with an empty table and no 3rd argument, should it return nil or throw an error? I lean toward returning nil. In either case, the caller can verify that the table is non-empty before calling `reduce`.

luther9:
I still need to write tests for this, but I've uploaded my latest commit so people can check the API.

CelticMinstrel:
Nice, you figured out the exact same method I had in mind for distinguishing a missing third argument from a nil third argument. I also see this is now recursive instead of using a loop… I think Lua has tail call optimization though so that shouldn't cause any problems, right?

I don't have any particular objections to the code now, but I do have one minor request – for Lua documentation comments, please use three dashes `--- like this`. It's the format I plan to use once I get around to pushing my documentation update. (I already went through and documented everything in that format; just haven't gotten around to pushing it yet.)

luther9:
> I also see this is now recursive instead of using a loop… I think Lua has tail call optimization though so that shouldn't cause any problems, right?

Yes, proper tail calls are documented here: https://www.lua.org/manual/5.4/manual.html#3.4.10

I'll get to work on fixing the comment and writing the tests.

CelticMinstrel:
With the new luacheck PR now merged, you'll also need to either add this file to the exclusion in .luacheckrc, or refrain from using the `<const>` syntax. I don't have a preference either way.

CelticMinstrel:
I think it seems fine to merge, but… probably the new unit test should also be added to `wml_test_schedule`?

luther9:
I found the file, but I have no idea what the numbers mean. Do I just add `0 lua_functional` to the `Lua API tests` section?

CelticMinstrel:
The numbers are something like expected result code. Most tests should be 0. Just as an example, a 1 means the system expects the test to fail (ie, will consider it a pass if it fails).

Pentarctagon:
FYI the unit test script is `run_wml_tests`, and the list of valid result values is at https://github.com/wesnoth/wesnoth/blob/master/run_wml_tests#L20-L33

PR #6338, opened 2021-11-27
[WMLTools] Improve KeyboardInterrupt Handling and Style Refactor

max-torch:
Uses a try-except clause instead of a signal listener to catch KeyboardInterrupt. Put all the contents of
```python
if __name__ == "__main__":
```
inside its own function called main() and then wrapped the call to main in try-except.

Additionally, some styling has been tweaked to be more pythonic as per PEP8 style guide.

The commits have been structured to try to make it clear everything that has been done in this PR.


max-torch:
Not yet ready for review please wait.

Pentarctagon:
@max-torch should this be closed at this point?

max-torch:
> @max-torch should this be closed at this point?

yes. closing now

PR #6339, opened 2021-11-28
1.16 - some attack icon work

doofus-01:
#6297 inspired me to look at the attack icons again.  So far, this PR adds some feet and changes some existing icons that looked off to me.  It also adds fangs that should fit the sea-serpent better than the snake fangs.

A few other things that could possibly be included:
1.  Tail for the big Icemonax, it currently uses the dragon tail
2.  Shadows currently use ghoul claws, they need their own version
3.  Possibly another spear version, less crude than the existing one.  The pike icon might be good enough for most cases (for example the spear branch in the dunefolk), so this is low priority.
4. Sometimes a single attack icon is used for two different attacks, which looks a bit odd.  For example the Giant Mudcrawler.  So, there are several icons which could use alternates, where the same thing may be depicted, but something like the camera angle is changed.


cooljeanius:
good to see the various icons for kick attacks; I'll finally be able to drop my attempt to hack one onto the Orcish Nightblade!

doofus-01:
The last push is a demo for what I meant in point 4 above.  Not sure the image is final, but the idea is there.

Descacharrado (commented):
Overall, I like your work very much but I'm not sure about the replacement of "mace-spiked" I suggest keeping both images and using the new icon for lower level heavy infantry and/or just the thug line. Not that I'm anyone to decide, just wanted to leave my feedback.

doofus-01:
Nothing wrong with thoughtful feedback.  The old mace image really does need to be fixed, but maybe I lost something in the update.

doofus-01:
That pretty much covers the things on my list.  I'll merge later unless someone has objections.

PR #6342, opened 2021-11-30
VS+Clang

Pentarctagon:
---

Not necessarily something I'm saying we should definitely do, but opening the PR to start a discussion about using Clang with Visual Studio on Windows rather than MSVC. The main idea being that it would mean we'd only need to deal with Clang and GCC rather than Clang, GCC, and MSVC, and that we'd be able to build with Clang on all the platforms we support.

### DIFF
--- a/src/desktop/version.cpp
+++ b/src/desktop/version.cpp
@@ -381,7 +381,7 @@ std::string os_version()
 			}
 	}
 
-	if(v.szCSDVersion && *v.szCSDVersion) {
+	if(*v.szCSDVersion) {
 		version += " ";
 		version += unicode_cast<std::string>(std::wstring(v.szCSDVersion));
 	}
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -73,7 +73,6 @@ jobs:
           export CACHE_DIR=/home/wesnoth-CI/build
           ./.github/workflows/ci-scripts/ubuntu.sh
 
-
   steam-runtime:
     runs-on: ubuntu-20.04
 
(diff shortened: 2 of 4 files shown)


cooljeanius:
I think it makes sense to remove MSVC from the CI, since there were issues with the CI running out of resources, but I also think support for compiling with MSVC should stick around for people compiling locally just in case that's the only option they have available to them at the time

loonycyborg:
Generally it's considered useful to have builds with as many compilers as possible because each one could have found an issue others missed. I have no love for MSVC in particular but it's still an extra compiler.

Pentarctagon:
@cooljeanius That would be a bad idea more due to VS+MSVC being the only jobs that build on Windows, and we do need to make sure commits can build there.

@loonycyborg True, though there have also been some other large projects that switched (Chrome and Firefox being the first that come to mind). MSVC is also in a bit of a weird position in that its output is never actually distributed anywhere.  GCC and Clang are used for the packages distributed across macOS and Linux, but since the Windows packages are actually cross-compiled with GCC on Linux as well, it ends up being a bit of a strange situation where Windows devs are using a compiler (MSVC) whose output is never used by any players.

src/desktop/version.cpp
```diff
@@ -381,7 +381,7 @@ std::string os_version()
 			}
 	}
 
-	if(v.szCSDVersion && *v.szCSDVersion) {
+	if(*v.szCSDVersion) {
```
soliton-, line 384: What's the reason behind this change?

Pentarctagon: Clang called it out as a warning, which strict mode convert to an error:
`address of array 'v.szCSDVersion' will always evaluate to 'true' [-Werror,-Wpointer-bool-conversion]`

Pentarctagon: The other warnings beside the microsoft one that've been added to the cmake script would also need to have more targeted fixes, either by telling it to ignore the warning for specific lines (usually boost includes) or other minor code changes.

soliton-: Ah, it's an array. Makes sense.

CelticMinstrel: I see the PR was closed, but should this change be made anyway?

Pentarctagon: Done in af5ddbbe2ce6e42943e620250072a16d1d6202ba

PR #6343, opened 2021-11-30
UtBS S02 Ogre now dissappears before Nym starts moving or talking

### LINKED ISSUE #6341: UtBS S2: Final ogre killed re-appears and lingers during Dust Devil event
1.16.1 release

### Describe the bug
In Under the Burning Suns S2 Across the Harsh Sands, there is a group of hunting ogres that ambush you near (17,49). A conversation event triggers (the one where Nym obtains the Dust Devil) when the last ogre is killed, but in 1.16 that ogre re-appears and lingers for the duration of the conversation.

### To Reproduce
Fortunately, the strange behaviour can be reproduced in debug-mode:

1. Start UtBS and use debug mode to go to the next scenario.
2. Move Kaleh and/or Nym to the area near (17,49) to trigger the appearance of the hunting ogres.
3. Debug-kill the ogres and notice that the final ogre re-appears to linger during the event where Nym obtains the Dust Devil.

### Expected behaviour
No ogres should be present (because they're all dead) during the event. This was the case for 1.14 so this seems to be a regression.

### Screenshots
ogre

[ogre.gz](https://github.com/wesnoth/wesnoth/files/7622777/ogre.gz)

Descacharrado:
As said in the comments, a die event makes the unit linger. Applied what is said in https://wiki.wesnoth.org/Eventwml#die

> Note: The primary unit is not removed from the game until the end of this event. The primary unit can still be manipulated, will block other units from taking its hex, and will still be found by standard unit filters (except [have_unit]). To prevent this behavior, you can use [kill] to remove the unit immediately.

(Closes #6341)

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
@@ -708,6 +708,12 @@
             message= _ "That’s the last of them. This looks like a hunting party, they must have a camp around here somewhere."
         [/message]
 
+        # Move Nym adjacent to the bottle. During a die event, $unit is still blocking the hex, so killing it with wml is needed.
+        [kill]
+            role="Hunting Ogre"
+            animate=no
+        [/kill]
+
         [item]
             x=$x1
             y=$y1
@@ -722,10 +728,7 @@
             message=_"Hey! Look there, they dropped something... A stone bottle, sealed. I wonder what’s inside..."
         [/message]
 
-        # Move Nym adjacent to the bottle. During a die event, $unit is still blocking the hex, so
-        # Nym probably ends up running to the hex north-west of $x1,$y1.
-        #
-        # Don't move her if she's already adjacent to it - if she landed the killing blow, moving
+        # Don't move Nym if she's already adjacent to the Ogre - if she landed the killing blow, moving
         # her now means that she doesn't get the XP. Also, the unit still has ZoC, and she'll take a
         # longer route to avoid it, which looks particularly obvious if Nym was adjacent the ogre.
         [if]
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
@@ -726,7 +726,7 @@
             message=_"Hey! Look there, they dropped something... A stone bottle, sealed. I wonder what’s inside..."
         [/message]
 
-        # Move Nym adjacent to the bottle, but don't move her if she's already adjacent to it
+        # Move Nym to the bottle, but don't move her if she's already adjacent to it
         # if she landed the killing blow, moving her now means that she doesn't get the XP.
 
         [if]
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
@@ -703,16 +703,12 @@
             [/have_unit]
         [/filter_condition]
 
-        [message]
-            speaker=Kaleh
-            message= _ "That’s the last of them. This looks like a hunting party, they must have a camp around here somewhere."
-        [/message]
-
-        # Move Nym adjacent to the bottle. During a die event, $unit is still blocking the hex, so killing it with wml is needed.
         [kill]
             role="Hunting Ogre"
-            animate=no
         [/kill]
+
+        [redraw]
+        [/redraw]
 
         [item]
             x=$x1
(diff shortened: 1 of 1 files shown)


Wedge009:
Thanks for working this out. I did some testing and to me the scene looks better if implemented this way:

1. `die` event is triggered
2. `kill` removes the last ogre
3. `item` is added where the ogre fell
4. `redraw` is triggered
5. Kaleh starts talking, followed by Nym.

So in the save file I tested with (extraneous new lines and commentary are stripped out in the save file) I settled on this:
```
name="die"
[filter]
	role="Hunting Ogre"
[/filter]
[filter_condition]
	[have_unit]
		count=0
		role="Hunting Ogre"
	[/have_unit]
[/filter_condition]
[kill]
	role="Hunting Ogre"
	animate=no
[/kill]
[item]
	image="items/potion-grey.png"
	x="$x1"
	y="$y1"
[/item]
[redraw]
[/redraw]
[message]
	message=_"That’s the last of them. This looks like a hunting party, they must have a camp around here somewhere."
	speaker="Kaleh"
[/message]
[message]
	message=_"Hey! Look there, they dropped something... A stone bottle, sealed. I wonder what’s inside..."
	speaker="Nym"
[/message]
```

Do you think you could re-order the events that way, please? Or if you disagree, happy to discuss further.

Wedge009:
The reason I did some experimentation with the ordering is because with the PR as it currently stands, the ogre still lingers when Kaleh is speaking and only disappears when Nym speaks (since the `kill` block is only run in between their respective lines).

stevecotton:
The logic of checking if Nym is already next to the ogre/bottle needs to be updated - probably to a Standard Location Filter based on `$x1, $y1, radius=1` instead of a `[filter_adjacent]$already_killed_unit.id`. If she's the one who does the killing blow, moving her means that she doesn't get the XP.

Wedge009:
That's separate to what the issue and PR is about though. Maybe add it to this PR or make the change separately?

stevecotton:
It's not separate: the ogre being killed is the `$unit` in `[filter_adjacent]id=$unit.id`, and killing it means that the `[filter_adjacent]` never matches. The filter needs to be changed to use some data that's still valid, otherwise this change breaks the XP bugfix.

Wedge009:
I see, that wasn't clear. I don't have the situation where Nym lands the killing blow so I didn't test that.

Descacharrado:
> The logic of checking if Nym is already next to the ogre/bottle needs to be updated - probably to a Standard Location Filter based on `$x1, $y1, radius=1` instead of a `[filter_adjacent]$already_killed_unit.id`. If she's the one who does the killing blow, moving her means that she doesn't get the XP.

I see, thanks for pointing that out. I confirm that would be the case. I will have a look at it and implement what I think works best.

Descacharrado:
Now it works for me in all these situations:
- Nym lands the final blow, does not move and gets the XP.
- Nym does not land the final blow but is by the Ogre, she does not move and the right unit gets the XP.
- Nym does not land the final blow and is not by the Ogre, she moves on top of the bottle.

data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
```diff
@@ -703,38 +703,40 @@
...
             message=_"Hey! Look there, they dropped something... A stone bottle, sealed. I wonder what’s inside..."
         [/message]
 
-        # Move Nym adjacent to the bottle. During a die event, $unit is still blocking the hex, so
-        # Nym probably ends up running to the hex north-west of $x1,$y1.
-        #
-        # Don't move her if she's already adjacent to it - if she landed the killing blow, moving
-        # her now means that she doesn't get the XP. Also, the unit still has ZoC, and she'll take a
-        # longer route to avoid it, which looks particularly obvious if Nym was adjacent the ogre.
+        # Move Nym adjacent to the bottle, but don't move her if she's already adjacent to it
```
stevecotton, line 729: Move Nym <strike>adjacent</strike> to the bottle, but ...

As the ogre is no longer blocking the hex, `{MOVE_UNIT id=Nym $x1 $y1}` now puts her on the same hex as the bottle. The effect looks good, this is just a request to update the comment.

stevecotton (approved):

Wedge009:
I thought it more natural to have the bottle appear when the ogre dies, not just when Nym points it out. But I suppose that's just an artistic choice, this is certainly an improvement over what we have now. I see steve has approved so if there are no other objections I'll probably merge towards the end of my day (by 09:00 UTC).

Descacharrado:
> I thought it more natural to have the bottle appear when the ogre dies, not just when Nym points it out.

I'll test if another redraw gets the job done

Wedge009:
Oh, when I looked at your updates a couple of hours ago I read something differently - maybe I was still in the process of waking up properly. The only difference I see between what you have currently and what I tested with is that the placement of the bottle and the `redraw` call are swapped. (The original code has the `redraw` call after the bottle placement too.) As long as the bottle appears appropriately, this should be fine.

Descacharrado:
Yup, just looked at it, bottle appears as one would expect, right after the ogre is killed, before anyone even talks.

Wedge009:
Sounds good, further questions/comments/objections notwithstanding.

PR #6344, opened 2021-12-01
UtBS S02 Added a ring icon for the immune unit

### LINKED ISSUE #6126: UtBS S2 Across the Harsh Sands: Add ring icon of Traveler's ring to the unit sprite
 - 1.15.18
 - Steam
 - Windows 10

Add an indicator icon on the sprite of the unit that takes the Traveler's Ring.
This feature is already implemented in Dead Water.
image
We should also implement this here.

### Screenshots
The Traveler's Ring which I gave to the unit called Nisen. On some occasions I had to check the names of all the Quenoth scout units to find him:
image

There is nothing to indicate that he now possesses the ring:
image

Descacharrado:
Effect ingame:
Sin título

As proposed in the issue, I used the silver ring from DW. 
If the image is wanted to be moved to core, I propose wesnoth\data\core\images\misc
(Closes #6126)


soliton-:
inmune -> immune

max-torch:
Does the ring icon disappear from the unit after this scenario? It makes sense for the icon to not appear in the following scenarios because there are no other situations involving thirst afterward.

knyghtmare:
> Does the ring icon disappear from the unit after this scenario? It makes sense for the icon to not appear in the following scenarios because there are no other situations involving thirst afterward. Maybe the player would have to be informed of this also? Because we will get questions from players saying "What happened to the ring?", those who don't know that the thirst mechanic will never reappear.

The object code's duration is not mentioned so it defaults to "forever", so this overlay will be staying as per the wiki documentation. Thus, the ring overlay should stay but not do anything beyond scenario 2.

knyghtmare (approved):
Nice

max-torch:
well FWIW, whether the icon disappears or not, it's still better than having no ring icon.

Descacharrado:
Ill fix the failed checks later and confirm or not if the ring dissappears.

Wedge009:
I see it seems to be a direct copy of the Dead Water icon, but I thought it offset a bit further to the right than where it is. But as max remarked it's better than having no indication, I suppose (I didn't even encounter the scorpions in my play-through a couple of days ago since I headed directly north instead of trying to reach the first oasis).

Since it's occurring in more than one location now, it'd probably be better to move it to the core location but since it's a relatively small file I suppose the duplication is okay for this case - it's only two instances. Maybe it can be moved if it's used a third time.

max-torch:
I just find it odd that thirst is talked about in dialog and in the help in such a way that makes it sound like a recurring mechanic, when in fact it is only a quirk of a single scenario.......

See, If it was made clear that the thirst thing was only a single scenario-specific one time mechanic, then it would be nice to get rid of the ring icon at the end of the level since it would have no more use, for the exact same reason that we also get rid of the thirst indicators at the end of the scenario.

Wedge009:
Did we come to a consensus about this? Was there anything further to be done here or is it okay to merge?

Descacharrado:
> Did we come to a consensus about this? Was there anything further to be done here or is it okay to merge?

I think that telling the player that there is not going to be any other thirst-mechanic scenarios would kill immersion, so I feel like this is pbbly the best way it can be handled. I could be convinced otherwise tho. And regarding the graphics, I would shift its color to a darker blue, but maybe that is such an insignificant change that it does not even matter for this PR.

max-torch:
> Did we come to a consensus about this? Was there anything further to be done here or is it okay to merge?

On my end, it's okay to merge. I can make a separate issue dedicated to what I find strange about the thirst and thirst-related things in UtBS. I just wanted those points I were making to be on record, especially here.

PR #6345, opened 2021-12-02
UtBS S4: Restrict who remarks about the dark underground tunnels

Wedge009:
By happenstance, Elyssa was my first unit to enter the caves. It sounded really weird for her to complain about the darkness and then say that she can use her fireballs to defeat the orcs.

As per the code commentary, it seemed the intention was to have this dialogue be for the elves:
https://github.com/wesnoth/wesnoth/blob/4248738a3894df4a551debd4ff6afff4e68ffe49/data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg#L746

So I added `race=quenoth` to exclude Elyssa and focus on the elves as per the original intention.

Unfortunately - or perhaps fortunately - Kaleh was my next unit in line to enter the caves, and it sounded really bad for him to complain about the darkness and then tell himself 'we have no choice'. So I added `[not]id=Kaleh[/not]` to cover that case.

Potentially we could exclude Zhul as well, but it sounds fine to me if she complains about the darkness and then reminds Kaleh about the elves' disadvantage fighting in that environment.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
@@ -752,6 +752,11 @@
                 time_of_day_id=underground
             [/filter_location]
             side=1
+            race=quenoth
+            [not]
+                # Exclude Kaleh lest he potentially talks to himself if he is the first to enter the cave
+                id=Kaleh
+            [/not]
         [/filter]
 
         [allow_undo]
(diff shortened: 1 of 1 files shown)


stevecotton:
It would make the diff bigger, so maybe not a good idea, but because of the `race=quenoth` it's no longer necessary to use `CHECK_EXPLORER` later in the same event.

I doubt that compatibility with old saves is that useful for the fourth scenario of the campaign, but `race=quenoth,elf` would be consistent with the compatibility code later in the campaign. My feeling is to keep the compatibility everywhere, then remove it all in a single commit once we're no longer backporting changes to 1.16.

Wedge009:
Yes, I did notice that - I forgot to point out that `CHECK_EXPLORER` only really accommodates the Dust Devil case and since I've restricted the race to `quenoth` it's now redundant.

I wasn't aware that race used to be `elf`. I'll add that later.

stevecotton:
Looks good to me.

PR #6347, opened 2021-12-02
UtBS Changed and documented the change of XP bar color to blue

### LINKED ISSUE #6212: UtBS: Change the Dust Devil's XP numbers to light blue instead of purple
 - 1.15.18
 - Steam
 - Windows

Change the Dust Devil's XP numbers from purple to light blue.

### Rationale
Kaleh's and Nym's XP numbers used to be purple, but later they were changed to light blue. This signals to the player that they are worth giving XP to because they have AMLAs and don't just get a full heal. Since the Dust Devil also has AMLA, it is worth giving XP to, and should have light blue XP numbers for the same reason Kaleh's and Nym's were changed to light blue.

However, if the design intent is for the Dust Devil AMLA to be more like an easter egg than a central feature, or if the AMLA gained by the dust devil is deemed too insignificant, then we can leave it as is.

Descacharrado:
Effect ingame:
If the Dust devil still has the "Taller" and "Stronger" AMLAs available, the XP is blue.
still has good amlas
If it reaches all major AMLAs, it turns purple so the player does not waste XP on him.
reaches all major amlas

Also documented the changes in the files, so it is easier to understand, for future reference, why the XP bar is blue.

(Closes #6212)

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/units/monsters/Dust_Devil.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/monsters/Dust_Devil.cfg
@@ -79,6 +79,10 @@
         id=Dust1
         max_times=4
         description= _ "a stronger Dust Devil"
+
+        # Make the XP bar blue instead of purple.
+        major_amla=yes
+
         [effect]
             apply_to=attack
             range=melee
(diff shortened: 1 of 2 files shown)


Wedge009:
Thank you. I was wondering what `major_amla` was supposed to mean.

knyghtmare (approved):

PR #6348, opened 2021-12-02
Color name of abilities like weapon in attack prediction windows for specifie source

newfrenchy83:
In case of specials abilities with value, the name will be colored after the benefit or not of special.

the mention "Teacher: " teacherstring
and "Opponent:" 
opponentstring
will be used for what player know what unit attack is affected by Teacher or by opponent.

Remove abilities applied to=opponent is not easy and i want propose that in alternative.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
@@ -171,7 +171,7 @@ A poisoned unit cannot be cured of its poison by a healer, and must seek the car
         female_name= _ "female^formation"
         description= _ "This unit gains a +10% bonus to defense when another unit with the same ability is adjacent to it. However, this cannot raise the unit’s defense above 70%."
         special_note=_"Groups of units of this type are able to shield each other in combat."
-        describe_effect=_"cth decreased"
+        describe_effect=_"CTH decreased"
         apply_to=opponent
         sub=10
         [filter_base_value]
--- a/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
@@ -171,6 +171,7 @@ A poisoned unit cannot be cured of its poison by a healer, and must seek the car
         female_name= _ "female^formation"
         description= _ "This unit gains a +10% bonus to defense when another unit with the same ability is adjacent to it. However, this cannot raise the unit’s defense above 70%."
         special_note=_"Groups of units of this type are able to shield each other in combat."
+        describe_effect=_"cth decreased"
         apply_to=opponent
         sub=10
         [filter_base_value]
--- a/data/core/macros/abilities.cfg
+++ b/data/core/macros/abilities.cfg
@@ -146,6 +146,7 @@ This ability will not cure an affected unit of poison, however, only delay its e
         female_name= _ "female^diversion"
         description= _ "If there is an enemy of the target on the opposite side of the target, this unit diverts the target’s attention and reduces its chance to hit by 20%."
         special_note=_"If there is an enemy of the target on the opposite side of the target while attacking it, this unit diverts the target's attention and reduces its chance to hit."
+        describe_effect= _ "distracted"
         affect_self=no
         affect_allies=no
         affect_enemies=yes
--- a/data/schema/units/abilities.cfg
+++ b/data/schema/units/abilities.cfg
@@ -6,6 +6,7 @@
 		super="units/unit_type/abilities/~generic~,units/unit_type/attack/specials/" + {NAME}
 		{FILTER_TAG "filter_student" unit {FILTER_TAG "filter_weapon" weapon ()}}
 		{DEFAULT_KEY overwrite_specials ability_overwrite none}
+		{SIMPLE_KEY describe_effect t_string}
 	[/tag]
 #enddef
 
--- a/po/wesnoth-lib/fr.po
+++ b/po/wesnoth-lib/fr.po
@@ -8103,14 +8103,6 @@ msgstr ""
 "Un terrain représenté par une chaîne de plus de 4 caractères a été trouvé, "
 "il s'agit de :"
 
-#: src/units/abilities.cpp:905
-msgid "Teacher:"
-msgstr "Professeur :"
-
-#: src/units/abilities.cpp:925
-msgid "Opponent:"
-msgstr "Adversaire :"
-
 #: src/wml_exception.cpp:62
 msgid ""
 "An error due to possibly invalid WML occurred\n"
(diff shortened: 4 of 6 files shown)


Wedge009:
I'm not sure if there should be any changes to the po files as I thought they are updated in a separate process after code chagnes.

Pentarctagon:
This seems like it could get cluttered pretty quickly, especially with the line wrapping in the second image.

newfrenchy83:
> This seems like it could get cluttered pretty quickly, especially with the line wrapping in the second image.

Using abbrevviation could be work.

newfrenchy83:
It is best with this:
teacherstring

and this?
opponentstring

Pentarctagon:
This wouldn't necessarily work when translations are involved, since the length of a translated string is independent of the length of the original English string.

newfrenchy83:
> This wouldn't necessarily work when translations are involved, since the length of a translated string is independent of the length of the original English string.

If the translatable terms cannot be used, then I see only two solutions: use icons, or color the names only according to the sources, (unit owner, teacher, or adversary), what do you say?

Pentarctagon:
Coloring sounds simpler I suppose.  For icons, we'd need to either already have the icon images (since Unicode isn't an option) or have them created, and then the UI config would also need to be updated.

newfrenchy83:
> Coloring sounds simpler I suppose. For icons, we'd need to either already have the icon images (since Unicode isn't an option) or have them created, and then the UI config would also need to be updated.

i doing that:
owned by unit
owned_by_unit

teached by allie
teaching

teached by enemie:
teaching_by_nenmies

and owned by opponent
opponent

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -171,6 +171,7 @@ A poisoned unit cannot be cured of its poison by a healer, and must seek the car
         female_name= _ "female^formation"
         description= _ "This unit gains a +10% bonus to defense when another unit with the same ability is adjacent to it. However, this cannot raise the unit’s defense above 70%."
         special_note=_"Groups of units of this type are able to shield each other in combat."
+        describe_effect=_"cth decreased"
```
Wedge009, line 174: I'm guessing 'cth' is 'Chance to Hit'. Why not write it out completely?

newfrenchy83: Like it must replace 'name' in attack prediction windows, describe_effect' must be have two word maximum else ithis would be too heavy, this is why i use abreviation, if you're suggestion, don't hesitate.

Wedge009: I understand English isn't your native language, but I'm not sure I understood all of that. Two word limit sounds very tight... and arbitrary. But I suppose if it really needs to be abbreviated, then capitalise CTH to make it clear (at least in English convention) that it's an acronym.

newfrenchy83: Using chance to hit in any letter would be too cumbersome in the attack prediction window but yes I should use CTH in all capitals then, unless you have a better description to come up with (one or two words to avoid line breaks)

newfrenchy83:
@Pentarctagon https://github.com/wesnoth/wesnoth/pull/6349 could be commited before 1.16.2, please?

stevecotton:
I feel we're cycling through the same designs, but not finding one that fits; either it's confusing or it's too cluttered. I think we need a GUI mock-up that we agree on first.

Let's assume that we can add new strings and redesign things, and that we could change the interface between src/units/abilities.cpp and src/gui/dialogs/attack_predictions.cpp (even if that becomes a separate PR). Putting font colors into src/units/abilities.cpp instead of the presentation file feels like overly-tight coupling of the classes; I know that you're continuing with the existing design that used grey font for inactive abilities, I think the grey font for inactive abilities is good, but I think the way that it was implemented is wrong.

My suggestion is to add tooltips to this dialog, and say that anything that's already included in the simple stats should only appear in the tooltip. We also have the "damage calculation" button, but adding tooltips would fit the rest of the UI.
* Tooltip only: leadership, because a stat such as `10x2` already includes it and tells us everything about how it will affect the fight.
* Tooltip only: formation, marksman and magical. Similarly to leadership, their effects are already represented in the chance-to-hit percentage.
* In dialog: status aliments such as slows, poison and petrifies, because they're in addition to the damage represented by the `10x2`.
* In dialog: first-strike, slows and petrifies, because these can reduce the damage taken in a way that the simple numbers don't show.
* In dialog: drain, because it's something that the simple numbers don't show.
* In dialog: growing fury, an ability in AE that increases damage on the second and third hits, because it's something that the simple numbers don't show.
* Not sure what to do with: trample, an ability in AE that moves the attacker after the attack. It's significant enough that I think it should be shown, but it doesn't affect the stats at all. Maybe the answer is that the UMC author has to make it obvious in the name of the attack.

With those rules, the existing UI for "magical" would change, it would become tooltip only.

tooltips_for_attack_prediction

It's false to assume `affects_enemy=yes` means that it should be a negative color. There are a few rare abilities that buff enemies or debuff allies; none of these affect chance-to-hit, but examples in Ageless Era are:
* `AE_mag_magicabsorb`
* `AE_mag_coldaura` (slows as well as doing damage)
* `AE_mag_heataura`
* `AE_mag_runeaura`
* `AE_YOKAI_ABILITY_UNLUCKY`
* indirectly from the "illuminates" ability or similar darkness abilities

newfrenchy83:
@stevecotton i try to implement a tooltip in attack prediction windows( in src/gui/dialogs/unit_attack.cpp) but i can work  what in map or unit description bars. I will instead using damage calulation dialogs except if someone can said how done this.

stevecotton:
> but i can work what in map or unit description bars

Sorry, I can't understand what this means. Please would you rephrase it?

I suggest only making prototyping pictures for a while and not spending time on the implementation - we've seen a lot of designs where someone objected to something in the UX, and I think we're going to have to try more UX ideas before we find the right solution. I've posted about this PR in the `#development` Discord channel, so hopefully we'll get comments and ideas from more people, and then those people won't object later.

knyghtmare:
> With those rules, the existing UI for "magical" would change, it would become tooltip only.

I think _magical_ should be present in both tooltip and in dialog. 

My reasons:

1. I know it's always a 70% chance to hit with the unit's magical attack but consider your case here: despite showing magical (70%), the CTH is 60% and then "magical" is not holding true. This would confuse a player who is starting UtBS for the first time ever or after having not played it for a long time. Such a thing would be confusing to them.

2. Same can be said for regular engagements. Players would often get a bit confused when a certain CTH is showing 70% while it's supposed to be higher (if attacking a land-based unit in shallow water). The "magical" text in the dialog reminds them that why it's 70% and not 80/90%. Thus, being a good thing. vice versa is also true.

PR #6349, opened 2021-12-02
Color name of abilities like weapon in attack prediction windows for specifie source(1.16)

newfrenchy83:
In case of specials abilities with value, the name will be colored after the benefit or not of special.

the mention "Teacher: " teacherstring
and "Opponent:"
opponentstring
will be used for what player know what unit attack is affected by Teacher or by opponent.
 I hope what this amelioration of feature can be pushed in 1.16 but i doubt.


Pentarctagon:
So this would now look like https://github.com/wesnoth/wesnoth/pull/6348#issuecomment-985771061 correct?

newfrenchy83:
> So this would now look like [#6348 (comment)](https://github.com/wesnoth/wesnoth/pull/6348#issuecomment-985771061) correct?

yes

PR #4728, opened 2020-01-31
Fix MINGW (with MSYS2) build in Windows

Adriankhl:
Previously, only scons build in MSYS2 is possible. Now cmake also works with this fix.

`cmake .. -G "MSYS Makefiles" -DCMAKE_BUILD_TYPE=Release -DARCH=native`

PR #6352, opened 2021-12-04
Fix MINGW build on Windows by splitting MSVC and WIN32

ddiefenthaler:
Fixes the MINGW build in the MSYS2 environment on Windows.

I've seen the previous PR #4728, but a more recent change to the CMakeLists-files introduced MSVC-specific compile and link flags for all WIN32 targets.
This PR splits `WIN32` and `MSVC` and keeps the `gcc`-flags if MSVC is not used.
With this I was able to build and run wesnoth in the UCRT64-environment of MSYS2 using CMake.

There is one more occurrence of `NOT WIN32` where I'm not sure if the section should apply for MSYS2.
~~I will try to add an inline comment for the relevant section.~~
Since there are no changes to this section a inline comment is not possible.
Here is a link to the section in question: [CMakeLists.txt line 81-95](https://github.com/wesnoth/wesnoth/blob/9d5271b98aebdb39631b98089f22f15d80c75a5a/CMakeLists.txt#L81-L95)
The POSIX-definition in line 94 breaks the build for me if applied on Windows.
I'm not sure if the other SDL-flags (line 90) are necessary. For me the build also worked without them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,19 +79,8 @@ find_package(Gettext)
 find_package(X11)
 
 if(NOT WIN32)
-	# needed to get some SDL2 defines in... (as of rev31694 -D_GNU_SOURCE=1 is required!)
-	if(NOT MINGW)
-		set(SDL2_CONFIG "sdl2-config" CACHE STRING "Path to sdl2-config script")
-		exec_program(${SDL2_CONFIG} ARGS "--cflags" OUTPUT_VARIABLE SDL2_CFLAGS)
-		add_definitions(${SDL2_CFLAGS})
-	else()
-		# equivalent to sdl2-config --cflags --libs
-		# since cmake cannot execute sdl2-config in msys2 shell
-		set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} -I/mingw64/include/SDL2 -Dmain=SDL_main -L/mingw64/lib -lmingw32 -lSDL2main -lSDL2 -mwindows)
-	endif()
-
-		# Use the safer `mkstemp' instead of `tmpnam' on POSIX systems.
-		add_definitions(-DLUA_USE_POSIX)
+	# Use the safer `mkstemp' instead of `tmpnam' on POSIX systems.
+	add_definitions(-DLUA_USE_POSIX)
 endif()
 
 #check for some compiler/arch specific things and export defines accordingly...
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,7 +170,7 @@ set(BINARY_PREFIX "" CACHE STRING "Prefix in front of all binaries")
 
 ### Set the environment compiler flags.
 
-if(NOT WIN32)
+if(NOT MSVC)
 	if(NOT DEFINED CXX_FLAGS_USER)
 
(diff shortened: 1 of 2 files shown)


Pentarctagon:
Maybe @Adriankhl can answer that, since nobody on the Wesnoth team uses cmake+mingw/msys2. Otherwise if it works without it, feel free to remove it.

ddiefenthaler:
Thank you for your response.
The current state should be fine then, it's deactivated by the `if(NOT WIN32)`-statement.

I dug a little bit through the history of the file and think the `if`-statement got merged at some point.
Unfortunately I don't know SDL2 that well and can't tell whether this step is still required.
It looks like all parameters are redundant through CMake and the SDL2 headers.

Anyway the line looks outdated, but I can't rule out that it isn't used by some cross compilation.
For MSYS2 it should be possible to remove the [`else`-bracket](https://github.com/wesnoth/wesnoth/blob/9d5271b98aebdb39631b98089f22f15d80c75a5a/CMakeLists.txt#L87-L90).
However, I can't say whether this is good practice for SDL2.

Pentarctagon:
So yeah, doesn't look like we're going to get a response. I'd say to just delete the entire block at https://github.com/wesnoth/wesnoth/blob/9d5271b98aebdb39631b98089f22f15d80c75a5a/CMakeLists.txt#L83-L91

ddiefenthaler:
Ok, I will prepare a commit like this.
I'm not sure if removing it completely works, since it also affects the Linux build,
but I guess the CI will tell if there are any immediate problems with it.

Still, I think it is useful to document this change somewhere, so that it can be undone if problems arise with it in the future.

For reference here are tables with the affected flags:
`sdl2-config --cflags` on Linux (Debian, Ubuntu and Arch tested):
| Flag | Description |
| ---- | ------------ |
| `-I/usr/include/SDL2` | Handeled by https://github.com/wesnoth/wesnoth/blob/9d5271b98aebdb39631b98089f22f15d80c75a5a/src/CMakeLists.txt#L21 |
| `-D_REENTRANT` | Uncertain, probably not needed on modern systems, the libc documentation marks this macro as obsolete. See https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html#index-_005fREENTRANT |

MSYS2 environment:
| Flag | Description |
| --- | --- |
| -I/mingw64/include/SDL2 | Handeled by https://github.com/wesnoth/wesnoth/blob/9d5271b98aebdb39631b98089f22f15d80c75a5a/src/CMakeLists.txt#L21 |
| -Dmain=SDL_main | Handeled by SDL_main.h from SDL2 |
| -L/mingw64/lib | Not needed, since it is the default path for libraries |
| -lmingw32 -lSDL2main -lSDL2 | Handeled by https://github.com/wesnoth/wesnoth/blob/9d5271b98aebdb39631b98089f22f15d80c75a5a/src/CMakeLists.txt#L49-L53 |
| -mwindows | Handled by cmake with `add_executable( WIN32 )` |

Adriankhl:
Hi, sorry that I didn't response for this earlier, I was quite busy with my own work.

I don't have access to Windows machine anymore so I can't test it. As far as I can remember, the build would fail without those extra lines. Therefore, as long as you can build it in msys2, then it should be good to remove those lines.

Pentarctagon:
Sounds good.  I'll wait for the CI to finish just in case, then I'll merge.

PR #6354, opened 2021-12-05
UtBS S8: Ensure humans are removed from display during their 'die' event

Wedge009:
There appears to be another situation as with #6341, but for the human attackers:
human

I tested this change directly in my current save file and it seemed to resolve the issue.

I also remove redundant phrasing and correct a commentary typo. @nemaara Please advise if you object to the former - I just find the phrasing of 'Hurry up already!' to be too modern (and redundant) compared with just 'Hurry up!'


nemaara:
I think there's not really a need to change her dialogue, there's lots of cases of more casual speech in UtBS besides this one.

nemaara (approved):

PR #6357, opened 2021-12-05
TRoW S19 Fix for the flying chocobones issue

Descacharrado:
Fixed a bug reported on discord by @Konrad22 which made the chocobones spawn over a chasm as can be seen:
Flycobones
Using passable=yes, and with the previous code implementation, no undead should spawn over a chasm.
No flybones

### DIFF
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/19_The_Vanguard.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/19_The_Vanguard.cfg
@@ -149,6 +149,21 @@
         [/ai]
     [/side]
 
+#define LOYAL_UNIT_PASS SIDE TYPE X Y
+    # Creates a unit with the Loyal trait and who can walk on where it is created
+    [unit]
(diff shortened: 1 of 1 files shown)


knyghtmare:
Wouldn't something like this be better?
```ini
#define PASSABLE_HEX
[+unit]
    passable=yes
[/unit]
#enddef
```

and 

```swift
{LOYAL_UNIT 5 (Chocobone) $|spawn_x $|spawn_y}{PASSABLE_HEX}
```

knyghtmare (commented):
If it's only for this scenario, it's missing an `#undef LOYAL_UNIT_PASS`

Descacharrado:
Done and tested. I had never used [+tag] so that is good intel.

knyghtmare (approved):

knyghtmare:
> Done and tested. I had never used [+tag] so that is good intel.

Programming essentials: you never stop learning and pick up something new everyday.

nemaara (approved):

Issue #6351, opened 2021-12-03
DiD, Alone At Last: Darken Volk can advance to Lich


Discontinuum:
It happens implicitly when [remove_trait] is used to remove "loyal" (so his unit is reassembled from the template and the change to the "advances_to" preperty lost).

The easiet way to fix: not remove his loyal trait (its description seems fitting), or specify in the comments that enabling the lich advancement is intentional (if it would really be considered fitting)

soliton-:
Single unit creation with different advances_to from the unit type:
https://github.com/wesnoth/wesnoth/blob/912625be842962704525f8fc8f9f39bbc7eb3a8e/data/campaigns/Descent_Into_Darkness/utils/characters.cfg#L38
Introduction of `[remove_trait]` which probably changed the behaviour: 26c2f941f2bcdd89528481e114c0375ad2a46271

Descacharrado:
What about using something like this, but with the dark sorcerer, as far as I know there are no dark adepts in the new DiD. https://github.com/wesnoth/wesnoth/blob/912625be842962704525f8fc8f9f39bbc7eb3a8e/data/core/macros/optional_unit_advancements.cfg#L12-L19

Wedge009:
@nemaara I suppose we can't apply this to 1.16 in case Volk has somehow advanced to Lich already?

nemaara:
I'm not sure. It would only apply to S8 so I think it's minor enough that it should go in 1.16. If it were in an earlier scenario, then it probably couldn't.

Wedge009:
I thought a similar thing. I only asked because of the commits you merged recently I think this one was master-only. I'll push it to 1.16 as well, if you haven't done it already.

nemaara:
I think it's already in 1.16.

Wedge009:
Yeah, it is. Sorry, for some reason I thought it was skipped.

PR #6358, opened 2021-12-05
fix advancement for Darken Volk

newfrenchy83:

PR #6359, opened 2021-12-05
fix Did advancement

newfrenchy83:
another fix off https://github.com/wesnoth/wesnoth/issues/6351

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/_main.cfg
+++ b/data/campaigns/Descent_Into_Darkness/_main.cfg
@@ -15,7 +15,10 @@
     define=CAMPAIGN_DESCENT
     first_scenario=01_Saving_Parthyn
 
-    {DISABLE_LICH}
+    [modify_unit_type]
+        type="Dark Sorcerer"
+        remove_advancement="Lich"
+    [/modify_unit_type]
     {CAMPAIGN_DIFFICULTY EASY   "units/undead-necromancers/adept.png~RC(magenta>black)" ( _ "Neophyte") ( _ "Easy")}
     {CAMPAIGN_DIFFICULTY NORMAL "units/undead-necromancers/dark-sorcerer.png~RC(magenta>black)" ( _ "Evoker") ( _ "Normal")} {DEFAULT_DIFFICULTY}
     {CAMPAIGN_DIFFICULTY HARD   "units/undead-necromancers/lich.png~RC(magenta>black)" ( _ "Summoner") ( _ "Challenging")}
--- a/data/campaigns/Descent_Into_Darkness/_main.cfg
+++ b/data/campaigns/Descent_Into_Darkness/_main.cfg
@@ -15,6 +15,7 @@
     define=CAMPAIGN_DESCENT
     first_scenario=01_Saving_Parthyn
 
+    {DISABLE_LICH}
     {CAMPAIGN_DIFFICULTY EASY   "units/undead-necromancers/adept.png~RC(magenta>black)" ( _ "Neophyte") ( _ "Easy")}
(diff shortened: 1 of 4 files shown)


knyghtmare (approved):

Wedge009:
Just so I understand why this is a 'fix' or improvement, what's the difference between `modify_unit_type` and `advances_to`? Is it because the lich removal is applied to all Dark Sorcerers and not just Darken Volk?

newfrenchy83:
> Just so I understand why this is a 'fix' or improvement, what's the difference between `modify_unit_type` and `advances_to`? Is it because the lich removal is applied to all Dark Sorcerers and not just Darken Volk?

While Darken Volk is the only Dark Sorcerer present in Did, the use of advances_to specifically on this character is unstable and requires @nemaara to undo the removal of the loyal trait when the wizard returns his jacket as this resets the settings and advances_to was rebuilt from unit_type, therefore modifying unit_type is more appropriate, and this method is already used in TroW to prevent Aethyr from advancing to Grand Marshal, and include the modification in the help file for unit_type

knyghtmare (approved):

Wedge009:
If this solution avoids the need for Darken Volk to keep his loyal trait, can you revert the removal of the removal (not a typo) of the loyal trait together with this PR, then? a304bc105dddde7069fd1412f56d51b20759b750 for master, eb6eb7d9627a219ea404297835a7f3c4345caa8a for 1.16.

nemaara (approved):
Any fix is fine with me, just needs a coder to approve/merge it

Wedge009:
I think @knyghtmare has approved, and we can keep Volk as no longer being loyal now. From what I understand this and #6358 makes sense to me.

knyghtmare (approved):

Discontinuum:
Sad Darken is no loyal now when switches sides, I thought it's a cool thing to do, why remove this loyal trait? I don't understand

knyghtmare:
> Sad Darken is no loyal now when switches sides, I thought it's a cool thing to do, why remove this loyal trait? I don't understand

When he switches sides, he has also gets canrecruit=yes which means leader/no upkeep. Which is like the loyal trait and more. So, having the loyal trait for a selfish character such as Volk makes little sense when he switches sides. Does that clarify your confusion?

Wedge009:
I also thought the loyal trait only really makes sense when he and Malin are working together. Now that they have separated, it doesn't make sense to keep it. And in fact, in the original state he didn't have the loyal trait - it was only restored to him to avoid the possibility of him advancing to Lich, which was then resolved properly in this PR (or so I understand).

Discontinuum:
It's not about upkeep really, it's about cool description. He is still loyal to himself when he switches

Wedge009:
That makes no sense, story selfishness has nothing to do with game-play loyalty trait.

Discontinuum:
Still the trait has this description, so it's used for the story purposes too.

Wedge009:
By your logic, every enemy leader should have loyal trait. Just leave it - the loyalty trait was only restored, briefly, as a work-around. It isn't coming back.

Discontinuum:
> By your logic, every enemy leader should have loyal trait. 
No. The only reason I want it for Darken is that it's specifically used as a story element, like this
изображение

No other enemy leader has this highlighted. 

And when I proposed it, I proposed it not as a work-around but as a permanent story thing as well.

Wedge009:
Okay, that's different. I had no idea that was even present until you pointed it out, but it is indeed there:
https://github.com/wesnoth/wesnoth/blob/82b54920d6ad7faab79281091dffebb5ee57279e/data/campaigns/Descent_Into_Darkness/utils/characters.cfg#L43-L45

I'd consider it more of an in-joke or an 'easter egg' of sorts. Unfortunately it is tied to the loyal trait mechanic and I think that does override this little reference when he turns against the player. It's only a 'loss' if you're actually astute enough to find this little quirk.

PR #6362, opened 2021-12-06
UtBS S9: Avoid 'punishing' the player for saving the rebels

### LINKED ISSUE #6229: UtBS S9 Blood is Thicker than Water: Player is punished for adhering to the protagonist's desires
 - 1.15.18
 - Steam
 - Windows 10

### The Issue
In S9 of UtBS, Kaleh, the protagonist, has to put down a rebellion. The scenario objectives get updated to say 
>Defeat Tanstafaal and Eloh

 At the same time, Kaleh speaks this line of dialog:
 >"Thank you, I’m sure you will be very useful in the shallow waters. All right people, I don’t want to kill any more than I have to. Too much blood has been spilled already. Knock ’em out, wound them, kill them only if you must. But we must stop Tanstafaal in his lunacy before he destroys us entirely."

I would like to draw your attention to "I don’t want to kill any more than I have to... kill them only if you must".
**Kaleh wants to achieve the objective while spilling as little blood as possible. He wants to avoid racking up kills.**

When the objective is accomplished, the scenario is not over. There is a new objective taking several turns to accomplish. Also, all the unkilled rebels get transferred to the player - along with all of their upkeep costs. This can seem like a reward in the short-term but is actually a punishment in the long-term because it directly takes away from gold the player takes into the next scenario, which is rather on the challenging side.
**The player is punished for spilling as little blood as possible.**

The reality of this contradiction is noted in the UtBS scenario walkthrough on the wiki:
>Whenever you destroy the apparition and capture Tanstafaal, the remaining elves will automatically switch over to your side. You might not want the extra overhead for the rest of this scenario, though. Besides, they screen Tanstafaal so well that most of them will have to be done in to get to him. Oh, and Zhul comes around after the apparition is gone.

**It is not good design to punish the player for being more aligned with the desire of the protagonist.**

### Proposal
This is more of a suggestion than a proposal, since others may come up with better ideas, or more time may be needed to brainstorm one:
Turn the rebels into allied units rather than part of the player's side so that they don't drain upkeep once the objective is accomplished.

### Screenshots
It is even possible to achieve the objective via assassination with merfolk due to the map design, but this strategy is not rewarded because you gain a lot of units:
image

The player will be punished for avoiding and ignoring the units crossing the eastern ford. The player will be rewarded for killing them. (Contrary to what the protagonist wants).
image

When there is a leader-kill chance to finish the objective the player is punished for keeping nearby rebel units alive before finishing the objective; they will just start draining gold:
image

Wedge009:
Resolves #6229.

I debug-tested this and it seemed to work as intended, but of course, please point out any mistakes or any suggestions you have.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg
@@ -2340,8 +2340,13 @@
             [/filter]
             side=1
             # If the player kept as many of the rebels alive as possible it's possible for their income to become negative. Give the unit no upkeep cost to compensate.
-            # Note that having loyal upkeep is not the same as having the loyal trait - this distinction is used to restore upkeep at the victory event.
-            upkeep=loyal
+            # Note that having loyal upkeep is not the same as having the loyal trait, no loyal overlay will be displayed.
+            [object]
+                duration=scenario
+                [effect]
+                    apply_to=loyal
+                [/effect]
+            [/object]
         [/modify_unit]
         [redraw]
             clear_shroud=yes
@@ -2748,17 +2753,6 @@
                 [/message]
             [/else]
         [/if]
-
-        # Rebel units were given zero upkeep upon joining Kaleh's side. Restore normal upkeep costs at the end of the scenario.
-        [modify_unit]
-            [filter]
-                upkeep=loyal
-                [not]
-                    trait=loyal
-                [/not]
-            [/filter]
-            upkeep=full
-        [/modify_unit]
 
         # Elves and troll/dwarf get in the ships, enemies disappear (the enemy side won't persist to the next scenario)
         [put_to_recall_list]
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg
@@ -2753,7 +2753,6 @@
                 [/message]
             [/else]
         [/if]
-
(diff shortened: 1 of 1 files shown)


Wedge009:
Hmm... I don't think I'm dealing with the loyal trait properly... but it seems to work anyway?

Descacharrado:
I would use an object with an id and then remove it, so no possible previous `upkeep=loyal` units could be affected. As it seems that the code would match any unit from any side with no upkeep but which is not loyal. https://github.com/wesnoth/wesnoth/blob/e1493388318610e3392964a64107e567d3c14eaf/data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg#L2753-L2761

Descacharrado:
> would you care to make your own contribution with what you described?

I'll have a go later

Wedge009:
I was thinking that you'd just make a fresh PR from scratch, but this looks good. Thanks.

max-torch:
Maybe also somehow communicate to the player that their upkeep is not diminished? New dialog that expresses they will fight without charging upkeep as amends for their attempted rebellion.

Wedge009:
Do you have any suggestions? Not sure how this can be communicated without sounding out of place.

'As penance we shall serve you for free... but only until we leave this wretched place...' x.x

I'd prefer to leave that as a separate enhancement. I don't think we really want to make an obvious point of this particular mechanic.

max-torch:
I suppose the latest discussion now was to just make all the units become loyal? That seems like it would work story/dialog-wise as the player will see the loyal trait indicator on the units and it makes a lot of implicit sense for them all to now be loyal.

Wedge009:
It was nemaara's decision to just make them loyal on the basis that there's no reward for the player to keep them around otherwise. I didn't think it quite worked in terms of story, given the rebels' doubt and still working through the conflict in their minds, but in the end I realised there's only one extra scenario to use them in so it didn't really matter too much. See #6365 for the discussion.

PR #6363, opened 2021-12-07
Remove the nonfunctional "title screen tutorial" hotkey

stevecotton:
The code that it used to trigger was removed in
f86168ab26b3a74835575fcc732d0d63e43bfe62.


Wedge009:
Makes sense to me. Well-spotted, too.

stevecotton:
> Edit: Something to note in the change log or would most users not even notice this?

Unlikely to notice this, I think. I only noticed it because I was looking for a different hotkey.

Vultraz (approved):

CelticMinstrel:
Yeah, I would say the hotkey was _logically_ removed when its functionality was. This is just a relic that was missed at that time.

Wedge009:
Wouldn't it make sense to remove it in 1.16 as well? steve said in #wesnoth-dev he felt it wasn't necessary to, but I think it is still something that should be tidied up.

stevecotton:
Thanks for the comments, I've now created a backport for it.

PR #6368, opened 2021-12-07
UtBS S06b Fixed weird scout AI behaviour

Descacharrado:
As commented in #6196 the pathfinder may prefer to go capture the hermit's village to fighting.
Pathfinder issue
So I added the behaviour to avoid the tunnels that lead outside (they are ai-trapped to fight, lines 189-193)
This behaviour is deleted when side 6 is activated in order to not create new AI weirdness (lines 1523-1528)


Wedge009:
Looks like the check stalled rather than actually failing a test?

Descacharrado:
> Looks like the check stalled rather than actually failing a test?

I feel like the fact that it took exactly 1h to fail is telling, yup.

nemaara (approved):
Looks like you need to fix the changelog.

Descacharrado:
> Looks like you need to fix the changelog.

I waited on this PR and #6369 for approval before fixing conflicts, in case more would show up when approved.

Pentarctagon:
@Descacharrado every once in a while the CI decides it just doesn't want to run one of the two macOS jobs for some reason. So in cases where it's just one macOS job that failed and everything else is fine, just ignore it.

PR #6369, opened 2021-12-07
UtBS S06b Look at the blown charges and make sure no units get stuck

Descacharrado:
After the charges are blown, the camera looks at the terrain (lines 1012 and 1089).
If a fast enough unit would get stuck in a wall, it gets teleported back (lines 1090-1103). There cannot be any other unit where the stuck one would get teleported, as that other unit would have triggered the event.

If this and #6368 are merged, #6196 could be closed.


Wedge009:
This shifts the unit immediately outside of the hex where the charges blow. I suppose it's impossible for another unit to be at the destination hex because that would be within the sight range of the dwarf and trigger the event anyway, right?

Descacharrado:
> This shifts the unit immediately outside of the hex where the charges blow. I suppose it's impossible for another unit to be at the destination hex because that would be within the sight range of the dwarf and trigger the event anyway, right?

Right, if there was an unit there, the code would have already triggered

Wedge009:
I just remembered either tunnel can be blown shut (whichever one the player approaches first). Did you account for the other tunnel too?

Descacharrado:
> I just remembered either tunnel can be blown shut (whichever one the player approaches first). Did you account for the other tunnel too?

Weirdly, the other tunnel cannot block an unit when being shut, firstly I added the code for that tunnel (before even commiting) because of future compatibility but then I thought it maybe would be confusing.

nemaara (approved):

Descacharrado:
Can now be merged.

Wedge009:
> Weirdly, the other tunnel cannot block an unit when being shut...

I believe you but I don't quite understand what you're saying. Or maybe I'm not remembering the scenario properly. Is it only the left tunnel that can block a unit because the charges are blown in front of the guard (from Kaleh's perspective) while for the right tunnel the charges are blown behind the guard?

I'm happy to merge this once I clear this up.

Descacharrado:
>the charges are blown in front of the guard (from Kaleh's perspective) while for the right tunnel the charges are blown behind the guard?

The charges on the right are blown in the same spot where the guard is standing. Oh, I just realised I did not add the images to show how it worked, sorry.

Wedge009:
No worries, that makes sense now. Okay, I'll merge this soon. Thanks.

PR #6373, opened 2021-12-09
Add [terrain_type] to the list of top-level tags.

### LINKED ISSUE #6285: Multiplayer scenarios with custom terrains don't load for non-host players and observers
 - What version of the game are you running? 1.16.0 on flatpak on Ubuntu 20.04, my friends used Steam on Windows

### Describe the bug
When you start a game in a multiplayer scenario with custom terrains, other players (not the host) see an error window "the map could not be loaded: Unknow tile in the map <some custom terrain id>" and fail. The host, on the other hand, works fine. The same error sees everyone trying to observe such a game. (NB: all the folks have the same version of the addon in question)

### To Reproduce

1. Install any addon with a multiplayer scenario with custom terrains
2. In the lobby Create Game > Scenarios, choose a scenario in question
3. Either start a game with several people or start the game alone and let other people try to observe this game

Addons and scenarios you can try (from the 1.16 addon server): Era of Magic and scenario EoMa 4p - Zombies Challenge; Legend of the Invincibles (beta) and scenario LotI - Gladiators

### Screenshots
GladiatorsRIP
What you see for non-host players/observer in LotI - Gladiators

ZCbug

What you see in EoMa Zombies Challenge

### Additional context
Dunno if the bug is applicable to other things (campaigns, mp campaigns, etc), only tested for separate mp scenarios

Pentarctagon:
Fixes #6285

### DIFF
--- a/src/game_config_manager.cpp
+++ b/src/game_config_manager.cpp
@@ -583,7 +583,8 @@ void game_config_manager::load_addons_cfg()
 				"resource",
 				"multiplayer",
 				"scenario",
-				"campaign"
+				"campaign",
+				"terrain_type"
 			};
 
 			for(const std::string& tagname : entry_tags) {


stevecotton:
I'm not sure if it's the right fix - that's not a negative thought, I just haven't understood how it all fits together, but wanted to add the "Backport" tag as fixing that bug seems important enough to trigger the 1.16.2 release.

Pentarctagon:
Given the bad commit, my guess/understanding is that `entry_tags` is a list of top-level WML tags that it looks for and copies into the game from whichever UMCs are active. So since it didn't list the [terrain_type] tag there, it was ignoring it. I don't know why it would affect only the non-host players though.

stevecotton:
I'm still not sure I'm understanding it right, but it seems `entry_tags` is a list that gets unconditionally `std::move`'d into the game config regardless of which UMCs are active. The leftover tags are then kept in `addon_cfgs_` and only copied if their add-on is active.

However, there's still definitely something that I'm missing. One of the affected scenarios is EoMa 4p - Zombies Challenge, for which the add-on must be active enough to provide the EoMa zombie units; so I don't understand why the bug triggers with that scenario.

Pentarctagon:
It almost seems like there's an issue with *when* certain things are getting added, something like the [terrain_type] is getting added to the game config after the host sends the WML over the network to the other players.  Otherwise I'd expect this to have been failing everywhere if it was just entirely excluding [terrain_type], but that's clearly not what's happening.

stevecotton:
I tested a combination of clients with and without this change, using 1.16.1+dev (c61a65ef) with or without a cherry-pick of this PR.

In each case I'm using the scenario EoMa 4p - Zombies Challenge; as this scenario uses EoMa zombies an OOS for missing units is possible when using a non-EoMa era (the host gets zombies, but no enemies appear on other players' clients). "yes" and "no" refer to whether this patch has been applied.

* era default, host yes, client no: terrain error on client
* era default, host no, client yes: OOS error on client with (on stdout/stderr) `unknown unit type: EoMa_Walking_Corpse`
* era EoMa, host yes, client no: ok (even tried restarting the client, and it was still ok)
* era EoMa, host no, client yes: ok
* era EoMa, host no, client no: ok

At least for this scenario, because it uses units from the same add-on as the terrain, the current PR only changes the error message; this change doesn't change an unplayable situation into a playable situation.

AI0867:
Does adding `[terrain_type]` help if the related images aren't also transferred?

Discontinuum:
Wait, @stevecotton , do you mean that EoMa units aren't spawned when the era is custom even if all clients have EoMa? The scenario should work with any era, just all folks have to have EoMa

CelticMinstrel:
> Does adding `[terrain_type]` help if the related images aren't also transferred?

I think the game _is_ technically still playable if you're only missing images, though.

stevecotton:
> do you mean that EoMa units aren't spawned when the era is custom even if all clients have EoMa? The scenario should work with any era, just all folks have to have EoMa

Yes, that's another effect of the same feature that caused this `[terrain_type]` bug, and I think @Pentarctagon is treating it as part of the same bug, based on https://github.com/wesnoth/wesnoth/issues/6285#issuecomment-991948208 . As the scenario is part of the EoMa add-on, fixing the logic so that the add-on is considered to be in-use should fix both.

Aside: for a scenario in one add-on that wants to use units from a different add-on but still allow the players to choose any era, the necessary extra work for 1.16 is explained in https://github.com/ProditorMagnus/Ageless-for-1-14/issues/27 .

Pentarctagon:
Closing in the mean time since this is clearly not the right solution.

PR #6375, opened 2021-12-10
Print which campaign is causing the extra_defines warning

stevecotton:
Whether or not it's the campaign being played, this warning
gets printed if any add-on has a [campaign]extra_defines=
attribute with one of the deprecated strings (a hardcoded
list about 20 lines above the line that's changed here).


stevecotton:
I didn't add any filtering to prevent the warning if the add-on involved isn't being played, because it's simple for anyone to report the error to the campaign author; I'm expecting add-ons to get fixed quickly when it's easy to see which one is causing the warning.

knyghtmare (commented):
Finally, we shall get to see which add-ons are the culprits of spamming the screen. 
(hopefully not any of mine).

Issue #6377, opened 2021-12-10
Building fails on SunOS 5.11 (e.g. OpenIndiana) and fixes


larb0b:
### Game and System Information
 - What version of the game are you running?
    - Wesnoth 1.16.1
 - Where did you download it from? (Steam, Mac App Store, our website, built from source, etc.)
    - Built from source
 - If you're using a custom build, what compiler did you use, and what commit did you build at?
    - GCC 10.3.0, Wesnoth built from 1.16.1 release tarball
 - What OS are you running?
    - OpenIndiana 2021.10

### Describe the bug
Wesnoth does not build successfully out of the box on SunOS 5.11/Illumos platforms.

### To Reproduce

Attempt to compile on a SunOS 5.11/Illumos distribution e.g. OpenIndiana.

### Expected behavior
Wesnoth builds successfully.

### Issues and fixes
* Linking failure
  * On Solaris-based platforms, when doing things with sockets, you need to link against the socket library with `-lsocket`. If cmake or scons detect SunOS, add `-lsocket` to final linking of `wesnoth`, `wesnothd`, etc. executables.

* utsname difference
  * The declaration of `utsname u;` should be `struct utsname u;` on SunOS in `src/desktop/version.cpp`
e.g. change [this](https://github.com/wesnoth/wesnoth/blob/a202146b892249905d0306fe5777ff9802cb0bd9/src/desktop/version.cpp#L220) from
`utsname u;`
to

```
#if defined(__sun)
    struct utsname u;
#else
    utsname u;
#endif
```

### Screenshots
![image](https://user-images.githubusercontent.com/45925460/145645321-d86baf45-641c-45cf-9e4c-5b1ce87caa68.png)

After these adjustments, Wesnoth builds and runs perfectly. Tested on OpenIndiana 2021.10.

Pentarctagon:
Can you open a PR with these changes? Nobody on the dev team uses this OS, so we can't test ourselves that the change we think is what you mean is actually correct and working.

larb0b:
I would be willing to, but I'm not sure what the proper way to tell cmake and scons to add `-lsocket` for SunOS is.

Pentarctagon:
[...]

PR #6379, opened 2021-12-10
SunOS 5.11/OpenIndiana build fixes

larb0b:
With these changes, Wesnoth builds and runs on SunOS 5.11, tested on OpenIndiana 2021.10.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -637,6 +637,9 @@ for env in [test_env, client_env, env]:
         env.Append(FRAMEWORKS = "IOKit")            # IOKit
         env.Append(FRAMEWORKS = "CoreGraphics")     # CoreGraphics
 
+    if env["PLATFORM"] == 'sunos':
+        env.Append(LINKFLAGS = "-lsocket")
+
 if not env['static_test']:
     test_env.Append(CPPDEFINES = "BOOST_TEST_DYN_LINK")
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -46,6 +46,10 @@ elseif(APPLE)
 	set(common-external-libs ${common-external-libs} ${IOKIT_LIBRARY} ${SECURITY_LIBRARY})
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL "SunOS")
+	set(common-external-libs ${common-external-libs} socket)
+endif()
+
 set(game-external-libs
 	${common-external-libs}
 	${SDL2_LIBRARY}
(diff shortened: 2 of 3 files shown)


loonycyborg:
Buildsystem changes look fine to me.

Wedge009:
@Pentarctagon Any objections to merging? As you said, the only one (we're aware of so far) who is affected by #6377 and benefits from this PR is the requester.

Pentarctagon:
It looks like nothing happened to break in any of the other builds, so I'm good with it.

Pentarctagon:
@larb0b For future reference as well, if you add `Fixes #6377` to the commit message, the issue #6377 will be automatically closed when the commit is merged.

larb0b:
Ah right, forgot to do that. 

By the way, tested building Wesnoth with these patches on Solaris 11.4 and it works:

image

CelticMinstrel:
I'm not sure it's even worth using conditional compilation in the first commit, as the change should still compile on any platform.

Wedge009:
I think I understand what you mean. My guess is larb0b is being conservative in not wanting to potentially break the non-SunOS builds. There's no harm in merging this as-is and potentially including the socket library for other *NIX builds later is there?

Pentarctagon:
A code comment would be just as effective in that case.

Pentarctagon:
The CI jobs also check the configurations we ourselves use, so if it broke one of those we'd know before the PR is merged.

Wedge009:
Would you prefer such a comment before merging then?

larb0b:
> I think I understand what you mean. My guess is larb0b is being conservative in not wanting to potentially break the non-SunOS builds. There's no harm in merging this as-is and potentially including the socket library for other *NIX builds later is there?

Celtic was referring to the first commit, regarding utsname, not the socket stuff. `-lsocket` is only needed on SysV-style stuff like Solaris where it's a separate library, on most modern system's it's just part of libc. 



> I'm not sure it's even worth using conditional compilation in the first commit, as the change should still compile on any platform.
> 
> I guess it _does_ have the advantage of making it clear _why_ it's being declared in the C way but beyond that…

Yeah, it doesn't really matter.

CelticMinstrel:
> I think I understand what you mean. My guess is larb0b is being conservative in not wanting to potentially break the non-SunOS builds. There's no harm in merging this as-is and potentially including the socket library for other *NIX builds later is there?

Yeah, as larb0b said, I was referring to the `struct utsname u;` line; sorry if that was unclear.

Pentarctagon:
@larb0b So specifically, can you change the first commit to be just `struct utsname u;`?

CelticMinstrel:
Possibly with a comment noting that the struct keyword is required to compile on SunOS?

larb0b:
Done.

Pentarctagon:
Alright, I'll let this run on the CI just to be sure, then merge after it finishes.

Issue #6381, opened 2021-12-11
[terrain graphics] - sconces can have odd rules


doofus-01:
Sconces sometimes get drawn in an odd way.  In the editor, try drawing a wall running NE-SW, then overlay a sconce on the SW end.  Do the same for a wall running NW-SE.  I intend to fix this for 1.16.  I think there were some other glitches, I'll probably find them again when fixing this one.

I can't imagine anyone is relying on the asymmetric sconce behavior, but I'm posting a bug report to see if this is something that is frozen in stable.

Pentarctagon:
Graphical display glitches aren't an API, so they can be fixed in stable.

doofus-01:
Changing the behavior of a core macro could be seen as an API change.  Hopefully it isn't in this case, but it will make things look a little different, so best to ask.  I'll close this in a couple days, if there are no objections.

Pentarctagon:
As long as people don't need to make changes in their own code for it, I'd say it's still fine.

PR #6382, opened 2021-12-11
1.16 - some terrain graphics fixes

doofus-01:
This PR is for terrain-graphics issues I can fix this weekend, no particular plan outside of that.

* make X*^Efs sconce overlay rules symmetric, addresses #6381 and removes another glitch with catacombs candles in south-facing corners


doofus-01:
A screenshot of what that last commit means, the new behavior and images are the upper row
Screenshot_20211211

doofus-01:
I have no idea what those CI failures are about.

Wedge009:
A lot of tests are failing now, but I'm not sure why. It does look like a lot of error messages relating to 'invalid rotations' but it looks like you've already tried to address that.

doofus-01:
We'll see if CI passes now, but at least the invalid rotations error goes away.  Apparently, that `error engine: invalid rotations` message shows once per value in a rotations key that doesn't have six values.

PR #6385, opened 2021-12-16
Various build changes, mostly cmake

Pentarctagon:


stevecotton:
> The one webp image is included just so there's something for the scons validation to check against.

There doesn't seem to be any .webp file here. The change looks good, otherwise.

Pentarctagon:
Oh, right - I split that off since the rest of this can be merged prior to webp being enabled.

Pentarctagon:
@irydacea FYI as well, since you had looked into the SDL2 issue with the Debug build previously as well.

PR #6390, opened 2021-12-18
Get the scenario_id from the addon content list sent by the server if it's not otherwise available.

### LINKED ISSUE #6285: Multiplayer scenarios with custom terrains don't load for non-host players and observers

Pentarctagon:
Fixes #6285

### DIFF
--- a/src/server/wesnothd/server.cpp
+++ b/src/server/wesnothd/server.cpp
@@ -1760,9 +1760,9 @@ void server::handle_player_in_game(player_iterator p, simple_wml::document& data
 		g.level().root().apply_diff(*scenario_diff);
 		const simple_wml::node* cfg_change = scenario_diff->child("change_child");
 
-		// it is very likeley that the diff changes a side so this check isn't that important.
+		// it is very likely that the diff changes a side so this check isn't that important.
 		// Note that [side] is not at toplevel but inside [scenario] or [snapshot]
-		if(cfg_change /** && cfg_change->child("side") */) {
+		if(cfg_change) {
 			g.update_side_data();
 		}
 
--- a/src/game_config_manager.cpp
+++ b/src/game_config_manager.cpp
@@ -146,7 +146,7 @@ bool map_includes(const preproc_map& general, const preproc_map& special)
 } // end anonymous namespace
 
 void game_config_manager::load_game_config_with_loadscreen(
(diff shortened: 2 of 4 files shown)


soliton-:
So do we need similar logic for replays?

src/server/wesnothd/server.cpp
```diff
@@ -1760,9 +1760,9 @@ void server::handle_player_in_game(player_iterator p, simple_wml::document& data
 		g.level().root().apply_diff(*scenario_diff);
 		const simple_wml::node* cfg_change = scenario_diff->child("change_child");
 
-		// it is very likeley that the diff changes a side so this check isn't that important.
+		// it is very likely that the diff changes a side so this check isn't that important.
 		// Note that [side] is not at toplevel but inside [scenario] or [snapshot]
```
soliton-, line 1764: The comments seem to be obsolete now.

Pentarctagon: Alright, I'll remove it then.

soliton-:
That probably means for replays all addons are loaded which presumably means potential for OOS if when playing and replaying different WML is used.

Pentarctagon:
If that is the case, I think that would be a separate enhancement, beyond the currently opened issue.

PR #6392, opened 2021-12-20
Add build support for webp images.

Pentarctagon:


Pentarctagon:
@Wedge009 It looks like webp support is not yet present.

Wedge009:
That may well be, I was just reporting what I saw in Steam's release notes... maybe it was a premature announcement or I/we mis-understood what they meant by adding webp support?

Pentarctagon:
I'm guessing it was an announcement about the beta client, since that's what was said in the issue I opened.

Wedge009:
That's what I meant by premature announcement. I only use the regular releases.

loonycyborg:
Even after I enable webp support in sdl2-image the test program still exits with status code 1 for me.

loonycyborg:
Disregard that, seems I actually upgraded sdl1-image instead. Test works with actual sdl2-image.

loonycyborg:
If steamrt doesn't provide webp then we'll have to build our sdl2-image there, though maybe just installing webp would be enough. AFAIK sdl-image can find image libraries dynamically too.

Pentarctagon:
The comment from the Valve dev said that it will be generally available mid-January or later, so it's just a matter of waiting for that to be pushed. We don't need to build our own SDL2_image for the steam runtime.

Pentarctagon:
@loonycyborg The support for webp images is now available for the steam runtime, but while it appears to work in-game (I went into my steam installation of wesnoth and replace an image with the webp version of it and it still displayed fine), the conftest is saying there is no webp support. Do you have any ideas why that check wouldn't be working for the steam runtime?

loonycyborg:
What sort of issue? If I extract log file then it shows that webp config checks builds fine but fails to run.

Pentarctagon:
From the steam runtime issue:
> However, I think something might have gone wrong with the lazy-loading mechanism in SDL_image: our container has /usr/lib/x86_64-linux-gnu/libwebp.so.6, but strings /usr/lib/x86_64-linux-gnu/libSDL2_image-2.0.so.0 only mentions libwebp.so.2 (and there is no hard dependency via DT_NEEDED because SDL_image uses dlopen()). I'll look into it.

> Sorry, looks like a build-time dependency wasn't tight enough: it worked in my test environment, but not in final builds. I'll have another try at this in the next Steam beta cycle.

Pentarctagon:
It looks like the steam runtime support for webp is now working, so it's just a matter of waiting for it to be made generally available.

Wedge009:
For Windows, does libwebp support need to be built into the SDL2 library? I note you have configured `vcpkg.json` as such, plus using the default pre-built SDL2 DLLs provided by the project I don't appear to have webp images (currently only the background image on the title screen) showing even though the webp DLLs are present.

Pentarctagon:
I'm not sure - the pre-built Windows libraries haven't been used with VS for a while now.  There are options when compiling SDL2_image to enable/disable support for some formats though.

Wedge009:
My mistake - the pre-built SDL2_image DLLs already include webp support, I just wasn't including the right libwebp DLL. I have it working now, so all resolved.

Pentarctagon:
I would recommend using the current cmake configuration at some point though.

Wedge009:
Understood.

PR #6393, opened 2021-12-20
Omit line-numbers from the .po files

stevecotton:
The aim of this change is to reduce the number of changes in .po files caused
by a pot-update. For example, changes elsewhere in .cpp and .cfg files added
lots of changes just to change the line numbers.

This change is done on the final stage of creating the .pot files, so that the
same stage's `msgcat --sort-by-file` sorting still works as before, sorting by
both file and line.

Note: although the aim is reduce the number of changes in the long term, it
creates a huge diff on the very first pot-update. There are slightly more
deletions than insertions, as files with the same translatable string repeated
on multiple lines are now only listed once.

Estimated sizes for pot-update at the start of 1.16.2's string freeze. This
pot-update introduces (across all languages) 13438 new fuzzy strings, circa
60000 of the insertions are those fuzzies:

* 1.16.2 without this: 214795 insertions(+), 135146 deletions(-)
* 1.16.2 with this: 1283834 insertions(+), 1339266 deletions(-)
* 1.16.2 if this was in 1.16.1: 130824 insertions(+), 52728 deletions(-)

Estimates for master:

* 1.17.0 without this: 448590 insertions(+), 369170 deletions(-)
* 1.17.0 with this: 1356682 insertions(+), 1411589 deletions(-)

### DIFF
--- a/po/CMakeLists.txt
+++ b/po/CMakeLists.txt
@@ -96,6 +96,7 @@ if(ENABLE_POT_UPDATE_TARGET)
 
 			COMMAND ${GETTEXT_MSGCAT_EXECUTABLE}
 				--sort-by-file
+				--add-location=file
 					 ${PROJECT_SOURCE_DIR}/po/${DOMAIN}/${DOMAIN}.cpp.pot
 					 ${PROJECT_SOURCE_DIR}/po/${DOMAIN}/${DOMAIN}.wml.pot
 				--output ${PROJECT_SOURCE_DIR}/po/${DOMAIN}/${DOMAIN}.pot
--- a/po/SConscript
+++ b/po/SConscript
@@ -80,15 +80,15 @@ if "pot-update" in COMMAND_LINE_TARGETS:
         if cfgs and sources:
             env.Command(new_pot, [source_pot, wml_pot],
                 [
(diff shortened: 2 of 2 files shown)


CelticMinstrel:
This doesn't sound like a good change. At least hypothetically, those line numbers can be used by a po editor to show the context of a specific string in the source code. I'm not sure how well this actually works in practice, but… I don't think it's something we should render impossible?

Pentarctagon:
> The aim of this change is to reduce the number of changes in .po files caused by a pot-update.

What is the need behind this goal?

stevecotton:
> At least hypothetically, those line numbers can be used by a po editor to show the context of a specific string in the source code.

That's a bigger issue than I thought - I though the only .po editors that supported it were Vim and Emacs, which can open files by filename and where the users likely know their editor's search facilities. I've just found that Poedit can also do it, but Poedit 3.0.1 only accepts filenames when they have a line-number at the end - it won't even open the file if the line number is removed.

> What is the need behind this goal?

The main reason was having some local work-in-progress .po files in Git where I was cherry-picking and rebasing, and finding a lot of noise in the diff when merging. Partially it was just because of the size of the changes that happen every release.

PR #5864, opened 2021-06-15
add the names of the abilities used as weapons in the attack prediction window

newfrenchy83:
Now, if a weapon is affected by an ability used as a weapon, the name will appear as if it was a classic special.

PR #6399, opened 2021-12-23
add tooltip to unit_attack dialogs

newfrenchy83:
with this way, weapon abilities applied to stats of attacks showed in tooltips same if from opponent but only if modify damage num of attack or chance to hit stats.

i cannot capture the windows with tooltip this time.

### DIFF
--- a/data/gui/window/tooltip_floating.cfg
+++ b/data/gui/window/tooltip_floating.cfg
@@ -141,7 +141,7 @@ def reevaluate_best_size(w, s)
 
 [window]
 	id = "tooltip"
-	description = "The tooltip popup window with large tooltips, eg in the main menu."
+	description = "The tooltip popup window with floating tooltips."
 
 	[resolution]
 		definition = "tooltip"
@@ -164,6 +164,57 @@ def reevaluate_best_size(w, s)
 
 		[helptip]
 			id = "tooltip"
+		[/helptip]
+
+		[grid]
+
+			[row]
+
+				[column]
+
+					[label]
+						id = "label"
+						definition = "default_small"
+
+						use_markup = true
+						wrap = true
+					[/label]
+
+				[/column]
+
+			[/row]
+
+		[/grid]
+
+	[/resolution]
+
+[/window]
+
+[window]
+	id = "tooltip_transparent"
+	description = "The tooltip popup window with floating tooltips, and transparent background."
+
+	[resolution]
+		definition = "tooltip_transparent"
+
+		automatic_placement = false
+
+		functions = "{__GUI_WINDOW_FUNCTIONS}"
+
+		x = "{__GUI_WINDOW_X}"
+		y = "{__GUI_WINDOW_Y}"
+		width = "{__GUI_WINDOW_WIDTH}"
+		height = "{__GUI_WINDOW_HEIGHT}"
+		reevaluate_best_size = "{__GUI_WINDOW_REEVALUATE_BEST_SIZE}"
+
+		# TODO tooltips in this window make little sense.
+		# Have to think of a nice solution.
+		[tooltip]
+			id = "tooltip_transparent"
+		[/tooltip]
+
+		[helptip]
+			id = "tooltip_transparent"
 		[/helptip]
 
 		[grid]
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -839,21 +839,24 @@ std::vector<std::pair<t_string, t_string>> attack_type::special_tooltips(
  * @param[in] sp reference to ability to check
  * @param[in,out] checking_name the reference for checking if @name already added
  */
-static void add_name(std::string& weapon_abilities, bool active, const config::any_child sp, std::set<std::string>& checking_name, bool affect_adjacent)
+static void add_name(std::string& weapon_abilities, bool active, const config::any_child sp, std::set<std::string>& checking_name)
 {
 	if (active) {
 		const std::string& name = sp.cfg["name"].str();
-
 		if (!name.empty() && checking_name.count(name) == 0) {
 			checking_name.insert(name);
 			if (!weapon_abilities.empty()) weapon_abilities += ", ";
-			if (affect_adjacent) {
-				weapon_abilities += sp.cfg["affect_enemies"].to_bool() ? font::span_color(font::BAD_COLOR, name) : font::span_color(font::GOOD_COLOR, name);
-			} else {
-				weapon_abilities += font::span_color(font::BUTTON_COLOR, name);
-			}
+			weapon_abilities += font::span_color(font::BUTTON_COLOR, name);
 		}
 	}
+}
+
+static void add_string(std::string& weapon_abilities, std::string& self_abilities, const std::string temp_abilities, std::set<std::string>& checking_name)
+{
+	weapon_abilities = "";
+	checking_name.clear();
+	self_abilities += (!self_abilities.empty() && !temp_abilities.empty()) ? "\n" : "";
+	self_abilities += temp_abilities;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


CelticMinstrel:
There _must_ be a way to capture a screenshot of the tooltip. How is it different from the existing tooltip?

newfrenchy83:
> There _must_ be a way to capture a screenshot of the tooltip. How is it different from the existing tooltip?
unit_attack_tooltip


I hope what is sufficient

stevecotton:
I think tooltips work as a UI - I think we need to test with multiple abilities and probably adjust the layout a bit.

For the implementation, sorry but I think the presentation code shouldn't be in src/units/abilities.cpp. Having said that, I'm tempted to say it's something that it could be accepted as "matches the existing implementation" for 1.16, and sort out the technical debt for 1.18. OTOH, that's already the full quota of technical debt that I want to underwrite (say that I'll fix it if no-one else does) for 1.18.

To capture tooltips, Gimp's screenshot allows a delay, so you can click on Gimp's OK button and then have 5 seconds to hover the mouse and trigger the tooltip. @newfrenchy83 thanks for the screenshot, I'm only posting more because I'd already captured them:
newfrenchy_attack_tooltips

CelticMinstrel:
Okay, so… there _was_ no tooltip before and this is adding one, right?

But what I really want to know is how does the new tooltip definition differ from the original tooltip definition?

newfrenchy83:
> Okay, so… there _was_ no tooltip before and this is adding one, right?
> 
> But what I really want to know is how does the new tooltip definition differ from the original tooltip definition?

The tooltip added to the attack prediction window details what special influences such stats (no tooltip in this window before), and the abilities used as weapons and their source are included in the tooltip of the description bar of the unit, which only contains traditional specials (is that the tooltip you're talking about?).
And in 1.17, the applied abilities are absent from the window, placing them in the tooltip allows their influence in the stats to be underlined without directly cluttering the window.

When to tooltip_alter, I needed a floating tooltip with a transparent background instead of the solid rectangle used in the statistics and I could not transpose the code of the tooltips from the unit description bar

CelticMinstrel:
So basically, "tooltip_alter" is a tooltip with a transparent background?

In that case, perhaps we should call it "tooltip_transparent" instead?

newfrenchy83:
> For the implementation, sorry but I think the presentation code shouldn't be in src/units/abilities.cpp.

Except what default_value used in weapon_specials_value is defined in abilities.cpp and i will not defined all value and reference in unit_attack.cpp only for place weapon_special_value. This function must remain in abilities.cpp

data/gui/window/tooltip_floating.cfg
```diff
@@ -190,6 +190,57 @@ def reevaluate_best_size(w, s)
 
 [/window]
 
+[window]
+	id = "tooltip_transparent"
+	description = "The tooltip popup window with large tooltips, eg in the main menu."
```
CelticMinstrel, line 195: This description should be updated too.

CelticMinstrel:
For the tooltip, I notice the window definition was based on `tooltip_large` while the window itself (and placement strategy) was based on `tooltip`. It probably doesn't matter but it is a little strange.

data/gui/window/tooltip_floating.cfg
newfrenchy83: I changed the definitions

newfrenchy83:
@CelticMinstrel i think to one thing, if i use the tooltip for list all weapons specials and abilties, and what one weapon specials tradionnal must remains in unit_attack dialog?

newfrenchy83:
it must have that
wit_other_and_window_low
:

CelticMinstrel:
I'm not sure what your question is…

newfrenchy83:
> I'm not sure what your question is…

remove abilities of unit_attack dialog and leave that in tooltip only is a goog idea for you?

CelticMinstrel:
Hmm, I'm not sure. It may be fine to have some redundancy in the two places…

newfrenchy83:
i completed the tooltip with boolean abilities and fix a bug in showing for void attack for defender
tooltip_comlete1
tooltip_complete2
tooltip3
tooltip4

src/units/abilities.cpp
```diff
@@ -839,19 +839,18 @@ std::vector<std::pair<t_string, t_string>> attack_type::special_tooltips(
  * @param[in] sp reference to ability to check
  * @param[in,out] checking_name the reference for checking if @name already added
  */
-static void add_name(std::string& weapon_abilities, bool active, const config::any_child sp, std::set<std::string>& checking_name, bool affect_adjacent)
+static void add_name(std::string& weapon_abilities, bool active, const config::any_child sp, std::set<std::string>& checking_name, bool already_used, const std::string& from_str)
```
stevecotton, line 842: This takes `already_used` by value rather than by reference; the `already_used = true` later in this function only sets a local variable that isn't read afterwards.

stevecotton:
Observations using Ageless Era units:

* "attack only" (AE Centaur's Pegasus, AE Centaur's Javelinier) is shown in grey, even when it's that side's turn
* "charm" (AE's Feyborn Witch) is shown as "from teacher" instead of "from opponent" in the case that the unit with it is attacking
* "charm" (AE's Feyborn Witch) is shown as "from teacher" instead of "from opponents" in the case that another unit is attacking
* "resistant 2" (AE Centaur's Javelinier) is shown on the owning unit's tooltip, not the opponent's tooltip
* "stardust" (AE Dark Elves' Lidh) is shown in the owning unit's tooltip, not the opponent's; it also shows the number of attacks that the Lidh has rather than the number of attacks that the opponent has
* The tooltip layout will be confusing for any ability whose name starts with a number, because the damage and number of attacks lines put the name immediately after the number (and the name is a translatable string)

Also, but this is probably an AE bug:

* "charm" (AE's Feyborn Witch) isn't shown when the other unit is level 2

newfrenchy83:
> 

"charm" is a [leadership" abilities, they don't have apply_to=opponent by "affect_enemies, same if it is teacdhed to enemy, the abilitie is privided by a teacher then yes "from teacher" instead of "From opponent" because abilities apply_to=self from a teachet (affect_adjacent)for resistant 2, it is because the tradionnal weapon specials are implemented wit affect_either and i negliged to implement code for other_, i will must correct that and alos for the number confusing.

src/units/abilities.cpp
```diff
@@ -839,19 +839,19 @@ std::vector<std::pair<t_string, t_string>> attack_type::special_tooltips(
...
-
 		if (!name.empty() && checking_name.count(name) == 0) {
 			checking_name.insert(name);
 			if (!weapon_abilities.empty()) weapon_abilities += ", ";
-			if (affect_adjacent) {
-				weapon_abilities += sp.cfg["affect_enemies"].to_bool() ? font::span_color(font::BAD_COLOR, name) : font::span_color(font::GOOD_COLOR, name);
-			} else {
-				weapon_abilities += font::span_color(font::BUTTON_COLOR, name);
-			}
+			if(!already_used) {
```
CelticMinstrel, line 849: This looks strange. Why is the `affect_adjacent` logic removed? Maybe there's a good reason, but it just looks strange…

src/units/abilities.cpp
```diff
@@ -867,41 +867,123 @@ static void add_name(std::string& weapon_abilities, bool active, const config::a
...
+std::string attack_type::weapon_specials_value(const std::set<std::string> checking_tags, bool only_active, bool is_backstab) const
+{
+	//log_scope("weapon_specials_value");
+	return weapon_specials_impl(only_active, is_backstab, checking_tags, true);
+}
+
+std::string attack_type::weapon_specials_impl(bool only_active, bool is_backstab, const std::set<std::string> checking_tags, bool is_tooltip) const
+{
 	std::string res;
+	bool tag_checked;
```
CelticMinstrel, line 882: Why is this declared here? I didn't read the code carefully, but it looks like it's only used in various loops (eg, the one immediately after this) where it's unconditionally assigned before using… so I think it would make more sense to declare it separately in each loop where it's used.

CelticMinstrel:
I can see some indentation errors in this code… they should be fixed before it's merged.

src/units/abilities.cpp
newfrenchy83: with using of tooltip for indicate origin of abiliries, using colour code don't have utility.

newfrenchy83:
> I can see some indentation errors in this code… they should be fixed before it's merged.

i think what it is fixed.

src/units/abilities.cpp
CelticMinstrel: So basically you're saying that this change makes that code redundant.

newfrenchy83: yes

src/units/abilities.cpp
newfrenchy83: it is fixed

newfrenchy83:
@stevecotton the latest screenshot after least modifications:
with_italics1
with_italics2
with_italics3
with_italics4

stevecotton:
There's one other place that calls `attack_type::weapon_specials`: the implementation of [WFL's weapon filter `formula=`'s `specials` value](https://wiki.wesnoth.org/FilterWML#Filtering_Weapons) in src/formula/callable_objects.cpp. I don't know if that's used at all, and it probably should be deprecated (at least, the current form that's based on a comma-separated list).

@CelticMinstrel, what would be the best way to search UMC for uses of that WFL?

Pentarctagon:
Just "Attacks" rather than "Num of attacks" would be better wording IMO.

newfrenchy83:
> Just "Attacks" rather than "Num of attacks" would be better wording IMO.

i's done

CelticMinstrel:
> There's one other place that calls `attack_type::weapon_specials`: the implementation of [WFL's weapon filter `formula=`'s `specials` value](https://wiki.wesnoth.org/FilterWML#Filtering_Weapons) in src/formula/callable_objects.cpp. I don't know if that's used at all, and it probably should be deprecated (at least, the current form that's based on a comma-separated list).

What exactly is the problem with it? What do you mean by a "comma-separated list"? The `specials` key on a weapon object is a list of strings, not a string containing commas.

> @CelticMinstrel, what would be the best way to search UMC for uses of that WFL?

I suppose you could search for lines that contain `formula` followed by `specials`? ie `formula.*specials` in regex?

stevecotton:
> What exactly is the problem with it? What do you mean by a "comma-separated list"? The `specials` key on a weapon object is a list of strings, not a string containing commas.

* The implementation of WFL's weapon filter uses std::string::compare() to compare something from the formula with the string returned by attack_type::weapon_specials(). It's not a list comparison, it's string comparison.
* When attack_type::weapon_specials() was changed to sometimes insert Pango formatting in the string returned, the documentation wasn't updated and I suspect the caller in the WFL implementation went unnoticed. However, at this point it didn't insert Pango formatting when `only_active==false`, which means Pango formatting wasn't sent to the WFL implementation.
* On the Wiki, WFL's `specials` attribute got deprecated, but not the `formula` equivalent (although I'd say this was still a warning to anyone using the formula version).
* When attack_type::weapon_specials() was changed to sometimes return abilities from other units, fairly sure the WFL implementation went unnoticed.
* This PR is going to make it likely to return Pango formatted strings even when `only_active==false`.

@newfrenchy83 please don't take that list as a complaint about your changes - the problems here predate #5864.

> I suppose you could search for lines that contain `formula` followed by `specials`? ie `formula.*specials` in regex?

I was hoping for a ready-to-run suggestion that would catch multi-line formulas.

CelticMinstrel:
I think you're getting confused about what's what…

* The implementation of the WFL weapon filter is literally just call the formula and convert the result to bool, no string comparisons involved.
* `attack_type::weapon_specials` is not used by WFL for evaluating the formula.
* It calls `attack_type::specials` and creates a list using the ID from each special (line 110 of callable_objects.cpp).
* I can see that `attack_type::weapon_specials` _is_ called by `attack_type_callable::do_compare` (line 165), which is probably wrong… this must be what you were talking about? That affects ordering of attacks within WFL, but I'm not sure what entry points there are to it… maybe keys in a map, or a list sort function? This won't affect anyone trying to use WFL to check if a weapon has a special.
* The WFL `specials` key has not been deprecated, yes. As mentioned previously, it's a list of the IDs of all specials the weapon possesses, so you can use list manipulation functions on it. It has some problems, mind you, so maybe we _do_ want to deprecated it in favour of something new.
* The WFL implementation uses `attack_type::specials`, so if that can return specials from other units, then it's fine; if not, then yeah, that may need to be addressed somehow.

newfrenchy83:
@stevecotton Could you post a screenshot with the most suitable tooltip position for you, please?

stevecotton:
@newfrenchy83 I think having it automatically placed based on the mouse position is good (so no change needed).

Sorry for not giving feedback - I need to study the caller in the WFL implementation first.

CelticMinstrel:
Why did you request a review from me? Did you change something since I last commented that needs further input?

newfrenchy83:
> Why did you request a review from me? Did you change something since I last commented that needs further input?

if you coild review https://github.com/wesnoth/wesnoth/pull/6399/commits/5d8db17087e0b080cc762d6ed7f07dfbc5742651 commit, please, and said that what you think of that.

CelticMinstrel:
Well, I don't see anything wrong with that commit. I don't get why you're replacing wesnoth-lib with wesnoth but I don't really have a problem with it either.

newfrenchy83:
> Well, I don't see anything wrong with that commit. I don't get why you're replacing wesnoth-lib with wesnoth but I don't really have a problem with it either.

"weapon: ","Damage: " and "Attacks: " ARE ALREADY TRANSLATED IN WESNOTH TEXDOMAIN.

newfrenchy83:
Don't foreget this PR, please.

src/units/abilities.cpp
```diff
@@ -867,41 +867,118 @@ static void add_name(std::string& weapon_abilities, bool active, const config::a
...
+	//log_scope("weapon_specials_value");
+	return weapon_specials_impl(only_active, is_backstab, checking_tags, true);
+}
+
+std::string attack_type::weapon_specials_impl(bool only_active, bool is_backstab, const std::set<std::string> checking_tags, bool is_tooltip) const
+{
+	std::string res;
+	for (const config::any_child sp : specials_.all_children_range()) {
+		bool tag_checked = is_tooltip ? (checking_tags.count(sp.key) != 0) : true;
+		if(tag_checked){
```
stevecotton, line 884: Throughout this function, `is_tooltip` seems to be used for multiple things that aren't logically connected to each other.

In this line and the line above it, it seems to be relying on the caller to pass `is_tooltip==true` if and only if it wants to do filtering using the `checking_tags`. If I'm understanding that right, I'd replace these two lines with
```
		if(checking_tags.empty() || checking_tags.count(sp.key) != 0) {
```

src/units/abilities.cpp
```diff
@@ -867,41 +867,118 @@ static void add_name(std::string& weapon_abilities, bool active, const config::a
...
+			const bool active = is_tooltip ? special_active(sp.cfg, AFFECT_SELF, sp.key, is_backstab) : special_active(sp.cfg, AFFECT_EITHER, sp.key, is_backstab);
```
stevecotton, line 885: On this one, I'm not understanding why the constructing a tooltip would use different values.

src/units/abilities.cpp
```diff
@@ -867,41 +867,118 @@ static void add_name(std::string& weapon_abilities, bool active, const config::a
...
+			const std::string& default_name = !is_tooltip ? default_value(sp.cfg, "name_inactive", "name").str() : "";
+			const std::string& name = active ? sp.cfg["name"].str() : default_name;
```
stevecotton, line 887: Again, I'm not understanding why constructing a tooltip should be a reason to use a different value. That single boolean argument seems to be moving logic that should be in the caller into this function.

I think the solution should be to move the presentation into the caller - instead of returning a string here, make a `struct` just to be this function's return type. I'm thinking it would be something along the lines of:
```
	/** Part of the return structure for weapon_specials */
	struct name_and_active {
		/** Contains the special's name or name_inactive, depending on whether it's active */
		std::string name; // TODO: I assume this should be a tstring
		bool active;
	};
	struct RENAME_THIS_THING {
		std::vector<name_and_active> from_self;
		std::vector<name_and_active> from_own_teacher; // is this always 'teacher', or can it be other allies?
		std::vector<name_and_active> from_opponent;
		std::vector<name_and_active> from_enemy_teacher;
	};
```

With that, move the presentation logic into the caller. Everything about tracking whether the current section heading has already been used in the string moves there, and the code in abilities.cpp becomes simpler.

stevecotton (commented):
Sorry for taking so long to respond. At the moment I feel this code would be hard to maintain.

I don't have a clear idea of what changes would make me happy with it. Tthe comments I'm adding now are outlining a redesign; but I feel that once I see that redesign I'll want even more changes before I'm happy with it.

src/units/abilities.cpp
newfrenchy83: here is_tooltip define if all tags can be used or only one affected to a stats("chance_to_hit" for chance to hit stats) if is_toolitip is true, then only these tags can be used, else, if all tags.

src/units/abilities.cpp
newfrenchy83: In tooltip, i want what only weapon_special who affect directly stats in tooltip was showed, if applyto=opponent, name appear in attacker in windows because owner of special, but in defender tooltip only because defender stats modified by special, and AFFECT_SELF used instead of AFFECT_EITHER.

newfrenchy83: and for name string, if inactive_name can appear in window, what inactive name appear in attacker tooltip then what special active but applied to defender is confusing, and name must't appear in that case.

src/units/abilities.cpp
newfrenchy83: except what using struct is not compatible with void add_name function, and the code don't seem becomes simpler.

newfrenchy83:
i will close this PR and open another because i rewrite the code for tooltip.

soliton-:
What's the point of opening a new PR? Can you stop doing that?

newfrenchy83:
> What's the point of opening a new PR? Can you stop doing that?

sorry, but the nmber of coversation and comment are too heavy for the rework of code was readable here.

soliton-:
The conversations and comments are the context in which this PR was formed. Just because you changed the implementation all that context is now irrelevant?

To me this just looks like now if I ever want to figure out the reason why something was done in the PR I will now have to go through multiple PRs and piece things together.

If there is too much conversation needed until a PR gets to a usable state then maybe that rather means the goal should be discussed (probably on the forum) before starting the implementation.

stevecotton:
>     * I can see that `attack_type::weapon_specials` _is_ called by `attack_type_callable::do_compare` (line 165), which is probably wrong… this must be what you were talking about? That affects ordering of attacks within WFL, but I'm not sure what entry points there are to it… maybe keys in a map, or a list sort function? This won't affect anyone trying to use WFL to check if a weapon has a special.

I'm not even sure if that code is reachable, but I'm thinking of changing it to use `attack_type::specials` instead, decoupling it from these PRs. That would also mean that `attack_type::weapon_specials` is only called with `only_active=true`.

Issue #4417, opened 2019-10-01
Special notes for terrains (auto-generated from properties)


stevecotton:
When the help browser shows a terrain, it could show special notes for healing, illumination, income, and maybe defense caps, etc. This wouldn't need to be added to the WML, the engine could automatically add them based on the terrain properties.

CelticMinstrel:
It does already show these properties, I thought? But adding them to the prose could still be useful.

stevecotton:
It only shows them in debug mode.

CelticMinstrel:
Oh, really? Then yeah, they should certainly be shown outside of debug mode in some way.

stevecotton:
Healing has been covered in #6400, however I'm leaving the issue open for:
* illumination - this requires handling the min and max illumination caps too, and reading the map settings to see what the default light levels are.
* income - the amounts can be configured on a per-side basis with `[side]village_gold` and `[side]village_support`.

I think defense caps are out of scope, as they depend on the units on the map, or possibly all the units included in the current era.

Issue #4232, opened 2019-08-15
A poisoned unit on a harmful (negative heals) terrain is cured without doing harm


stevecotton:
Terrain can have a negative healing value, with the results:

- non-poisoned units take damage (this will reduce the unit to 1hp but won't kill)
- poisoned units are cured of poison, and take no damage

This looks like a one-line fix of adding a `> 0` to heal.cpp's `poison_progress`, the main effort would be the testing.
https://github.com/wesnoth/wesnoth/blob/cb5c2669ef71983a152967edc7364f4740b7675e/src/actions/heal.cpp#L93-L95

stevecotton:
I think we should treat heals and harms as separate terrain attributes, and simply say that terrains with negative values for heals= aren't supported. The existence of the current bug suggests that no-one is using this feature, and the Wiki documentation for TerrainWML says that heals= only takes positive values.

UtBS's scenario [In the Tunnels of Trolls](https://github.com/wesnoth/wesnoth/blob/master/data/campaigns/Under_the_Burning_Suns/scenarios/06a_In_the_Tunnels_of_Trolls.cfg) has a lava chamber where damaging terrain is implemented in WML. It does different amounts of damage to walking vs flying units, and it does no damage at all to the fire guardians.

Fire-based terrains seem an obvious example, and players' intuition says that drakes would take less damage from it than other races.

CelticMinstrel:
I think a more sensible approach would be to treat `heals` and `cures` as separate terrain attributes, so the former can still be negative while the latter is either a boolean or a comma-separated list of statuses.

(Though I wouldn't _terribly_ object to also having a separate `harms` key. Also as you've mentioned it would make sense to be able to apply resistances to terrain damages, so maybe an optional `damage_type` key or something.)

jostephd:
> Fire-based terrains seem an obvious example, and players' intuition says that drakes would take less damage from it than other races.

If anything, I'd expect terrain damage from a fire-based terrain to be subject to the unit's resistance/vulnerability to the `fire` damage type...

JoaquinIglesiasTurina:
I would like to address this issue.
There are different opinions whether heals should be refactored and split heals / cures.
The changes required to address the actual bug are minimal. I do not believe any changes I could make to the code base would make addressing deeper issues with the health system any harder.

However, since I am a novice C++ programmer, I am very interested on this bug, because I want to write unit tests for the heal module.

Explicitly: 
1. I would like to write tests for the heal module. 
[...]

PR #6400, opened 2021-12-24
Add a "special notes" section to terrain help pages

### LINKED ISSUE #6394: Combined terrain (i.e. Oasis) doesn't show docs in the help browser
1.16.1+dev

### Describe the bug
This applies to 

1. Open a map with an oasis, such as DW S01.
2. Right-click and open the terrain description.
3. Note that the docs for "Desert Sands / Oasis" have only "Base Terrain: Sands" as well as the moment and defense properties.
4. On the help browser's topic list, click on Desert Sands, note the docs "Deserts have a somewhat ..."
5. On the help browser's topic list, click on Oasis, note the docs "A welcome sight ..."

The suggestion here is that the docs should inherit from the components, notably `Dd^Do` should mention that oases heal like villages.

### Related

Issue #4417 would turn the healing into a special note, which would solve the case of the oasis.

stevecotton:
The notes are automatically added based on the terrain's properties, there isn't a way to add or remove them in WML.

As-is, this fixes #6394. Issue #4417 mentions some other possibilities too, but I want to postpone these:
* income is complex about upkeep, and the values can be specific to the map and the side
* illumination would need to handle min and max lighting caps too
* defense caps would entangle this with the units on the map, I don't intend to add that at all

Any questions about negative values for gives_healing() are issue #4232; this PR simply doesn't add a special note for that.

keep_in_800x600
desert_village

The next image is what's shown when choosing the terrain in the help browser's list. The main reasoning for this PR is that the "A welcome sight" line doesn't appear when right-clicking on an oasis on the map and choosing the "terrain description" menu option.
oasis

Right-clicking on an oasis on the map leads to a page similar to this one. This wording is only used for 1 hp; for 2 hp or more the wording is similar to the 8 hp one shown above (that logic may vary by language, it's the language's pluralisation rules that force two wordings for en_US).
oasis_1hp

Another terrain that doesn't appear in mainline, but seemed better to handle than not to.
keep_no_recruit_onto

### DIFF
--- a/src/help/help_topic_generators.cpp
+++ b/src/help/help_topic_generators.cpp
@@ -18,6 +18,7 @@
 #include "help/help_topic_generators.hpp"
 
 #include "font/sdl_ttf_compat.hpp"
+#include "formula/string_utils.hpp"     // for VNGETTEXT
 #include "game_config.hpp"              // for debug, menu_contract, etc
 #include "preferences/game.hpp"         // for encountered_terrains, etc
 #include "gettext.hpp"                  // for _, gettext, N_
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
…why is there a difference between these two things?

* Right-click a hex and choose "terrain description"
* Open the help browser and click on the terrain

stevecotton:
The first shows the terrain on the map, for example Dd^Do. The second shows the terrain from the .cfg file, for example ^Do.

stevecotton:
> Also, what about illumination? Maybe there should be a note for that too?

I'd prefer to make a separate PR for that, as it requires handling the min and max illumination caps too.

CelticMinstrel:
> The first shows the terrain on the map, for example Dd^Do. The second shows the terrain from the .cfg file, for example ^Do.

Okay, so they _are_ the same for terrains without an overlay?

stevecotton:
> Okay, so they _are_ the same for terrains without an overlay?

Yes

stevecotton:
Now targeting 1.17.1 (and 1.16.3), as the string-freeze for 1.16.2 has started.

CelticMinstrel:
I think this is fine. As said previously, you missed illumination, but it sounds like you plan to get to it later, so that's fine.

Issue #6283, opened 2021-11-11
1.16 OOS while watching MP-replay with Ladder era add-on 1.0.0


amokmen:
### Game and System Information
Win 10
Wesnoth 1.16 and Ladder era add-on 1.0.0
Web-site

### Describe the bug
While watching MP-replay with Ladder era add-on 1.0.0 sometimes at different turns there could be OOS. If close replay and reopen again - there would be no more OOS.

It's second time I see this bug.

1. Download and save locally MP-[replay](https://wesnoth.gamingladder.info/download-replay.php?reported_on=2021-11-10+20%3A50%3A28).
2. **Open new Wesnoth instance**.
3. Start watching replay.
4. At turn 1 would be OOS.

Probably it's add-on problem, idk.

Discontinuum:
The bug will shoot if you watch a replay IMMEDIATELY after Wesnoth start. Wesnoth throws mistakes that it doesn't know the addon's units. If you try the second time or after some activity, it will work (probably the addon will be loaded finally at that second try)

Discontinuum:
The problem encompasses all user-made eras

amokmen:
It looks like **Wesnoth engine did not use _custom units_ from properly installed add-on while watching replay at 1st time**.
So, _unknown_ units are cause of OOS.

Pentarctagon:
Does this happen if you go to Multiplayer > Local Game > Load Game?

Discontinuum:
@Pentarctagon no. When you are in Load Game of Local Game, you already see units in the left bar, so the era is loaded when you proceed to Local Game.

MXB2001:
> Does this happen if you go to Multiplayer > Local Game > Load Game?

When I try this I get "Replays not supported in Multiplayer mode" or somesuch.

The only way I can load replays is either via main screen load (the one for campaigns) or via the MP load shortcut key CTRL-O (which seems to bypass that message somehow).

amokmen:
Just noticed that **Rav_Color_Mod** add-on also did not work at first replay loading.

Pentarctagon:
I remember looking at the cache loading a while ago for some reason or another, and IIRC what it essentially does is loads a minimal amount of the WML initially so it can display the title screen faster, then once you select Local Game or any of the other options it takes the time to load up everything.  Presumably the issue here is that the "load everything" step is no longer being done when loading replays from the main menu the first time.

amokmen:
[...]

PR #6405, opened 2021-12-29
[1.16] Only show planned moves to self and allies in multiplayer

### LINKED ISSUE #6292: Multi-turn moves are visible to enemy players
- reproduced on 1.16.1+dev (e74ceb44329d416f547b17087d432522b27d2a07), Linux local build
- originally reported on 1.16.1(as tagged), by ben24626 during [a ladder match against Checkhovs_Cannon](https://replays.wesnoth.org/1.16/2021/11/12/2p__Caves_of_the_Basilisk_Turn_22_(2678).bz2)
- Note: ladder-match replays are broken by #6283.

### Describe the bug
In some circumstances, preset moves for future turns are visible to opponents and observers.

In trying to reproduce it, it seems the unit must move at least the first hex of a multi-turn move to trigger the bug. If the unit already has no movement points remaining, then setting a preset move for the next turn doesn't leak to the other players, even if it's a multi-turn move.

In the replay of Ben's game, we see this move by Checkhov, followed by Ben asking "do you have a preset move for 13,16 to move to 10,17?" "cause the computer is showing me that"
```
	[command]
		from_side=2
		sent=yes
		[move]
			x="17,16,15,14,13,12,11,10"
			y="15,15,16,15,16,16,17,17"
		[/move]
		[checkup]
			[result]
				final_hex_x=13
				final_hex_y=16
				stopped_early=no
			[/result]
			[result]
				next_unit_id=58
				random_calls=0
			[/result]
		[/checkup]
	[/command]
```

Screenshot from an attempt to reproduce it (with different units):
opponent_preset_move

stevecotton:
Two commits. The first one unifies some common code from mouse_handler::mouse_motion and mouse_handler::touch_motion.

For changelog.md
---

``` 
 ### Multiplayer
   * Fixed the originally planned move becoming visible to enemies when an "enemy sighted" event interrupts a multi-turn move (issue #6292)
   * Made planned multi-turn moves consistently visible to allies.

 ### User interface
   * Fixed "enemy sighted, press t to continue" also activating a feature that showed units' route to where the move started (issue #6292)
```

Only show planned moves to self and allies
---

Fixes #6292, that the planned final destination from multi-turn moves and
interrupted moves was visible to enemies. There are three side-numbers involved
in this logic, listed below; the GUI code was checking whether the move should
be visible to the player whose turn it is, rather to the viewing team.

* un->side(), the owner of the unit
* mouse_handler::side_num_, the side whose turn it is
* viewing_team(), the side of the player whose computer this wesnoth-client is
  running on, thus which side's shroud and fog should be applied.

The plans are still hidden for units belonging to local AI sides, to hide any
goto_x,goto_y settings by campaign authors. The edge case in multiplayer of
allied network AI sides isn't tested here, but I think it's a trivial matter
whether or not those plans are shown.

Don't leave a hex selected when using 't' to continue a move
---

Fixes the minor part of issue #6292, that functionality showing the path
to a selected hex was active after continuing an interrupted move. That
feature is completely separate to the main bug of #6292, but it looks
confusingly similar in testing. The intended way to activate this feature is
via the "Select Hex" hotkey, which doesn't have any keybinding by default, so
should only be reachable by people who want the feature.

It would be good to refactor the movement handling so that it's not split
between mouse-handling and menu-handling files, but it would also need a lot of
work and would be out-of-scope for backporting to 1.16.

### DIFF
--- a/src/menu_events.cpp
+++ b/src/menu_events.cpp
@@ -895,6 +895,7 @@ void menu_handler::move_unit_to_loc(const unit_map::iterator& ui,
 		actions::move_unit_and_record(route.steps, &pc_.get_undo_stack(), continue_move);
 	}
 
+	mousehandler.deselect_hex();
 	gui_->set_route(nullptr);
 	gui_->invalidate_game_status();
 }
--- a/src/mouse_events.cpp
+++ b/src/mouse_events.cpp
@@ -158,6 +158,9 @@ void mouse_handler::touch_motion(int x, int y, const bool browse, bool update, m
 
 	// now copy-pasting mouse_handler::mouse_motion()
 
+	// Note for anyone reconciling this code with the version in mouse_handler::mouse_motion:
+	// commit 27a40a82aeea removed the game_board& board from mouse_motion, but didn't update
+	// the corresponding code here in touch_motion.
 	game_board & board = pc_.gamestate().board_;
 
 	if(new_hex == map_location::null_location())
@@ -335,14 +338,21 @@ void mouse_handler::touch_motion(int x, int y, const bool browse, bool update, m
 		else
 			un.reset();
 	} //end planned unit map scope
+}
 
+void mouse_handler::show_reach_for_unit(const unit_ptr& un)
+{
 	if( (!selected_hex_.valid()) && un && current_paths_.destinations.empty() &&
 		 !gui().fogged(un->get_location()))
 	{
-		if (un->side() == side_num_) {
-			//unit is on our team, show path if the unit has one
+		// If the unit has a path set and is either ours or allied then show the path.
+		//
+		// Exception: allied AI sides' moves are still hidden, on the assumption that
+		// campaign authors won't want to leak goto_x,goto_y tricks to the player.
+		if(!viewing_team().is_enemy(un->side()) && !pc_.get_teams()[un->side() - 1].is_ai()) {
+			//unit is on our team or an allied team, show path if the unit has one
 			const map_location go_to = un->get_goto();
-			if(board.map().on_board(go_to)) {
+			if(pc_.get_map().on_board(go_to)) {
 				pathfind::marked_route route;
 				{ // start planned unit map scope
 					wb::future_map_if_active raii;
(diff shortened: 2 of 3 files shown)


Wedge009:
I'll try to test this next week (was travelling just before NY), at least for the SP side where I was noticing the behaviour. I haven't yet looked at the changes in detail but I didn't see anything amiss.

Wedge009:
I'm finding it hard to replicate the original issue in either SP or MP, building 1.16 branch at 1.16.1+dev 836ac469782c1c18444c087c40f1cff483586000. But I'm assuming it's still present...

stevecotton:
Replication steps:

1. Set up Isar's Cross, with everyone having Drake Warriors as leaders. The bug can be reproduced with just 2 Wesnoth clients, assigning the 2 allied sides to the same client.
2. Without recruiting, move leader 1 (starting in the south) as far north as it can go this turn (to 13,16).
3. End turn for side 1.
4. Without recruiting, try to move leader 2 (starting in the east) a multi-turn move west to 5,13.
5. The move is interrupted because of sighting leader 1. Press 't' to continue, but don't end turn.
6. On side 1's client, mouse over leader 2. You see the information leakage bug, with a planned move to 5,13.
7. On side 2's client, mouse over any visible unit. You see the selected-hex thing, with the route for that unit to 20,13.
8. End turn for side 2.
9. On side 1's client, the leakage bug will no longer be visible.

Wedge009:
Not sure what I'm doing wrong, just can't seem to find the issue any more - I even went back to 1.16.1 release.

Maybe I'll have to start playing SP again and encounter it through normal play.

stevecotton:
That is confusing, I could definitely reproduce it (and could test with both fixed and unfixed clients in the same game).

Any objection to me merging it, as I think it's better to get the fix into 1.16.2 than wait for 1.16.3? With fixes for #6283 and #6285 already in that release, I'd like it to be the standard "upgrade to this to get all the major multiplayer fixes" release.

Wedge009:
I don't have any objections, though would be good if at least someone else can confirm it resolves the reported issue.

stevecotton:
I have a confirmation from @knyghtmare that he can replicate it, @knyghtmare please would you test that this fixes it?

PR #6406, opened 2021-12-29
Support macOS 10.8-10.11 SDKs

evanmiller:
Map new constant names to their previous names after testing the SDK version. Tested against wesnoth 1.16.1.

Patch adapted from https://github.com/macports/macports-ports/pull/13437

### DIFF
--- a/src/macosx/SDLMain.mm
+++ b/src/macosx/SDLMain.mm
@@ -22,6 +22,12 @@
 #import "SDLMain.h"
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


Pentarctagon:
What issue does this fix with those versions, specifically?

evanmiller:
Undefined symbol errors (`NSEventTypeKeyDown` etc were introduced in 10.12)

Pentarctagon:
Thanks for the patch!

PR #6407, opened 2021-12-29
Only try to activate required addons after loading all addons present.

### LINKED ISSUE #6283: 1.16 OOS while watching MP-replay with Ladder era add-on 1.0.0

Pentarctagon:
The current issue is that, when loading a replay that requires addons from the main menu, it's checking for which addons can be enabled with game_classification.active_addons(...) before it reloads the cache to contain addons (such as eras) that are behind the #ifdef MULTIPLAYER define. It is of course impossible to find addon content before it has been loaded, which results in things like OOS errors due to unknown unit types.

Fixes #6283

### DIFF
--- a/src/game_classification.cpp
+++ b/src/game_classification.cpp
@@ -118,9 +118,9 @@ std::set<std::string> game_classification::active_addons(const std::string& scen
 	std::set<std::string> loaded_resources;
 	std::set<std::string> res;
 
-	std::transform(active_mods.begin(), active_mods.end(), std::back_inserter(mods),
-		[](const std::string& id) { return modevents_entry("modification", id); }
-	);
+	for(const auto& mod : active_mods) {
+		mods.emplace_back("modification", mod);
+	}
 
 	// We don't want the error message below if there is no era (= if this is a sp game).
 	if(!era_id.empty()) {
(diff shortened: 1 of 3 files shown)


Pentarctagon:
Should be good now. The main two parts:
* Changed the call to `active_addons(...)` to be after the configs for addons has actually been loaded.
* The short circuit check at https://github.com/wesnoth/wesnoth/pull/6407/files#diff-22bf60e250911c0109ba1559e91aec0538ead1d1b1ea85369cfdf22d9086594eL187-L190 has been adjusted. Removing that entirely causes the batched WML unit tests to fail due to timing out, but the existing check can't be done since, again, you can't compare whether the currently active addons are the same as the addons that need to be made active next before you've loaded said addons to be able to check for them.

PR #6421, opened 2022-01-04
Remove command prefixes on commandline

### LINKED ISSUE #1018: ingame help command help misleading
help

Descacharrado:
Commands will no longer have a prefix in the commandline, but they will keep it in the chat commands.

Closes #1018


ProditorMagnus:
It does not look correct to assume prefix is always :.

Descacharrado:
> It does not look correct to assume prefix is always :.

It does not assume prefix to be ':', it tests for a command to have ':' as a prefix. To clear things up I tested this with chat commands and game commands:

Chat commands, in order, top to bottom: (emote, :emote, /emote, /emtoe, /:emote, :/emote)
commands in chat

Game commands, in order, top to bottom: (benchmark, :benchmark, benhcmark, /benchmark, /:benchmark, :/benchmark)
commands in twodots

I may have missed somewhere you can input commands too (not just ingame by pressing : or m), in that case, tell me and I'll test those.

src/map_command_handler.hpp
```diff
@@ -191,9 +191,16 @@ class map_command_handler
 		else if (help_on_unknown_) {
 			utils::string_map symbols;
 			symbols["command"] = get_cmd();
-			symbols["help_command"] = cmd_prefix_ + "help";
-			print("help", VGETTEXT("Unknown command '$command', try $help_command "
-				"for a list of available commands.", symbols));
+			std::string input = get_data(0);
+			std::string compare = ":";
+			if (input[0] == compare[0]){
```
Pentarctagon, line 196: Is there a reason to declare `compare` rather than just directing checking `:`?

Descacharrado: I though I had tried it and it gave me an error, but looks like I did something wrong then, it's now cleaner (on 3d4d406).

Pentarctagon:
It looks good to me.

stevecotton (changes requested):
This code is also used for the chat commands, for which `cmd_prefix_` is `/` and the `help_command_` is `/help` - removing the `/` will say "help" instead. I like the idea, but the implementation needs to handle these possibilities too.

Descacharrado:
> This code is also used for the chat commands, for which `cmd_prefix_` is `/` and the `help_command_` is `/help` - removing the `/` will say "help" instead. I like the idea, but the implementation needs to handle these possibilities too.

So we would want the previous implementation for chat commands. My idea was that if the player already knew how to enter a command (either using ':' or opening the chat and using '/') they would already know that 'help' refers to /help, but it is true that it may cause confusion an make people write help in chat instead of /help. Will commit changes later.

CelticMinstrel:
> Comparing what is written to existing commands and adding suggestions

Were you planning to do this? It's not actually that hard – just compute the Levenshtein distance between the input command and each available commands and suggest those for which the result is less than a certain threshold.

Descacharrado:
> Were you planning to do this? It's not actually that hard – just compute the Levenshtein distance between the input command and each available commands and suggest those for which the result is less than a certain threshold.

I wasn't planning on doing it in this PR but I was searching for ways to do it, thanks for the guidance. I wouldnt wait for me to do that before merging tho, if I develop that succesfully I'll just open another PR if this one is closed.

src/map_command_handler.hpp
```diff
@@ -191,9 +191,21 @@ class map_command_handler
 		else if (help_on_unknown_) {
 			utils::string_map symbols;
 			symbols["command"] = get_cmd();
-			symbols["help_command"] = cmd_prefix_ + "help";
-			print("help", VGETTEXT("Unknown command '$command', try $help_command "
-				"for a list of available commands.", symbols));
+			std::string input = get_data(0);
+			if (cmd_prefix_ == "/"){
```
stevecotton, line 195: Doing it like this still means that the class treats two prefix strings specially. Instead of hardcoding strings, please check whether `input` starts with `cmd_prefix_`. We're currently using C++17, so we don't have `std::string::starts_with()`, but we do have `boost::algorithm::starts_with()`. @CelticMinstrel do we have a utils function for that (I thought so, but can't find it)?

I think the logic should check whether get_command would succeed for the suggested change, so that `:help` would get a hint to remove the `cmd_prefix_`, but `:unknown` wouldn't.

CelticMinstrel: I'm not aware of one?

Even without Boost or C++17, "starts with" is just `str.find(x) == 0`, right?

Descacharrado: > Doing it like this still means that the class treats two prefix strings specially. 

Wait, isn't that what we want? We want the `"help_command"` symbol to be `cmd_prefix_ + "help"` when it's a chat command and just `"help"` when its a "normal" command, so we need to make a distinction (unless there is another way of checking whether you are in a chat command or in a "normal" command which I do not know of). Did you mean that I should be using something similar to `input[0] == cmd_prefix_` instead of `input[0] == ':'` in line 201?

> I think the logic should check whether get_command would succeed for the suggested change, so that :help would get a hint to remove the cmd_prefix_, but :unknown wouldn't.

I feel like that is reasonable.

stevecotton: > Did you mean that I should be using something similar to input[0] == cmd_prefix_ instead of input[0] == ':' in line 201?

That's what I meant, although it might be multiple characters so it should be a starts-with check instead. Also, the translatable strings should put cmd_prefix_ in the symbols instead of including it in the string itself.

src/map_command_handler.hpp
```diff
@@ -191,9 +191,21 @@ class map_command_handler
...
+			}
+			else{
+				symbols["help_command"] = "help";
+			}
+			if (input[0] == ':'){
+				print("help", VGETTEXT("Unknown command '$command', remove ':' or try '$help_command' "
+					"for a list of available commands.", symbols));
+			}
+			else{
+				print("help", VGETTEXT("Unknown command '$command', try '$help_command' "
```
stevecotton, line 206: Something that could be left to a follow-up PR would be to convert the `'` marks to match the https://wiki.wesnoth.org/Typography_Style_Guide . There are other strings in this file that aren't involved in the current change, so I'd be happy to postpone all of these to a follow-up.

src/map_command_handler.hpp
Descacharrado: e4506d5 should handle both requests.

src/map_command_handler.hpp
Descacharrado: I would agree.

src/map_command_handler.hpp
stevecotton: It still has a hardcoded `/`, and an assumption that cmd_prefix_ is exactly 1 character.

src/map_command_handler.hpp
```diff
@@ -190,10 +190,29 @@ class map_command_handler
...
+			if (cmd_prefix_ == "/"){
+				symbols["help_command"] = cmd_prefix_ + "help";
+				symbols["command"] = cmd_prefix_ + get_cmd();
+			}
+			else{
+				symbols["help_command"] = "help";
+				symbols["command"] = get_cmd();
+			}
+			if (get_cmd().find(cmd_prefix_) == 0){
+				if (const command* c = get_command(get_data(0).erase(0,1))) {
```
stevecotton, line 203: `c` isn't used, so it shouldn't need to store this to a variable.
The `erase(0,1)` is hardcoding the assumed length of `cmd_prefix`
Why use `get_data(0)` instead of `get_cmd()`?

src/map_command_handler.hpp
```diff
@@ -190,10 +190,29 @@ class map_command_handler
...
+					print("help", VGETTEXT("Unknown command '$command', remove '$prefix' or try '$help_command' "
+						"for a list of available commands.", symbols));
+				}
+				else{
+					print("help", VGETTEXT("Unknown command '$command', try '$help_command' "
+						"for a list of available commands.", symbols));
```
stevecotton, line 209: This and the following `else` print the same message, could merge them by merging the two `if` statements that get to this line.

Descacharrado:
Drafting until complete.

Descacharrado:
Succesfully changed the hardcoded "/" for a flag. With 9de67a5, if the command prefix is ever changed, it will not be necessary to modify `map_command_handler.hpp`

src/map_command_handler.hpp
```diff
@@ -359,6 +387,11 @@ class map_command_handler
 	{
 		cmd_prefix_ = value;
 	}
+	// set command or message flag
```
stevecotton, line 390: There's no documentation about what the "command or message" flag means. Please document how this method changes the observable behavior.

Descacharrado: Hope it is now fine. Could be shorter, but I like explaining thoroughly.

Descacharrado:
Converting to draft until I resolve conflicts (because of my other PR).

Descacharrado:
Uhhhhhh I just noticed an error, wait for another commit before review.

Pentarctagon:
@stevecotton is this good to merge at this point?

stevecotton:
I think it's good to merge, will do that tomorrow unless there are more comments.

I'd prefer if the comments were in Doxygen format and reworded, but the rest of the class isn't in Doxygen format either and at the moment I can't work out a clearer way to write the text that I want to change. All of that sounds like stuff for a later PR rather than the current one.

stevecotton (approved):

Descacharrado:
> I'd prefer if the comments were in Doxygen format and reworded, but the rest of the class isn't in Doxygen format either and at the moment I can't work out a clearer way to write the text that I want to change. All of that sounds like stuff for a later PR rather than the current one.

I'll revise the format in a follow-up PR.

PR #6423, opened 2022-01-05
Use a less complicated implementation for the MAKE_ENUM of CONTROLLER.

Pentarctagon:


Vultraz (approved):

Pentarctagon:
Overall this seems to be a pretty straight forward change. The one thing that caused a problem initially was that assigning a MAKE_ENUM value to a config attribute sets the string value of it, whereas if it's a regular enum it of course sets it to its numeric value.

The benefits being this new approach is far simpler and easier to understand than trying to delve into how MAKE_ENUM works, as well as having a non-throwing way to check if a string value is a valid member of the associated enum.

Also, barring any issues with this approach, I'd probably (over the course of 1.17) convert other uses of MAKE_ENUM to this approach with the goal of eventually removing MAKE_ENUM entirely.

stevecotton:
Is there a bug that requires this to be merged soon, or can it wait until later in the dev cycle, so that it's less likely that there will be backports to 1.16.x after this merges?

> Also, barring any issues with this approach, I'd probably (over the course of 1.17) convert other uses of MAKE_ENUM to this approach with the goal of eventually removing MAKE_ENUM entirely.

Although it's easier to understand, this gives it its own handcoded class instead of sharing the common code of MAKE_ENUM. I'd prefer to have a template or even a macro, so that other enums that are converted to the new system will share common code instead of being cut&pastes.

src/gui/dialogs/game_stats.cpp
```diff
@@ -68,7 +68,7 @@ unit_const_ptr game_stats::get_leader(const int side)
 static std::string controller_name(const team& t)
 {
 	static const std::array<t_string, 3> names {{_("controller^Human"), _("controller^AI"), _("controller^Idle")}};
-	return "<span color='#808080'><small>" + names[t.controller().v] + "</small></span>";
+	return "<span color='#808080'><small>" + names[t.controller()] + "</small></span>";
```
stevecotton, line 71: The game crashes when opening the stats dialog. Note the line preceding it, vs the changed indexes in the new enum.

src/gui/dialogs/multiplayer/mp_join_game.cpp
```diff
@@ -218,15 +219,15 @@ static std::string generate_user_description(const config& side)
...
-	} else if(controller_type == "null") {
+	} else if(controller_type == side_controller::none) {
 		return _("Empty slot");
-	} else if(controller_type == "reserved") {
+	} else if(controller_type == side_controller::reserved) {
 		return VGETTEXT("Reserved for $playername", {{"playername", reservation}});
 	} else if(owner.empty()) {
 		return _("Vacant slot");
-	} else if(controller_type == "human" || controller_type == "network") {
+	} else if(controller_type == side_controller::human) {
```
stevecotton, line 230: Both the old and the new code feel wrong - the MAKE_ENUM framework supports converting to and from strings, so why isn't this using it?

src/menu_events.cpp
```diff
@@ -1489,8 +1489,8 @@ void console_handler::do_droid()
 		const bool is_droid = menu_handler_.board().get_team(side).is_droid();
 		const bool is_proxy_human = menu_handler_.board().get_team(side).is_proxy_human();
 		const bool is_ai = menu_handler_.board().get_team(side).is_ai();
-		const std::string human = team::CONTROLLER::enum_to_string(team::CONTROLLER::HUMAN);
-		const std::string ai = team::CONTROLLER::enum_to_string(team::CONTROLLER::AI);
+		const std::string human = side_controller::human;
+		const std::string ai = side_controller::ai;
```
stevecotton, line 1493: The original code seems to be making local aliases to avoid the long enum_to_string code. Now that the right-hand-side is just side_controller::human, I don't think the alias is needed.

src/enum_maps/side_controller.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+class side_controller
+{
+public:
+	enum TYPE { INVALID = 0, NONE = 1, HUMAN = 2, AI = 3, RESERVED = 4 };
+	static constexpr const char* const invalid = "invalid";
+	static constexpr const char* const none = "null";
+	static constexpr const char* const human = "human";
+	static constexpr const char* const ai = "ai";
+	static constexpr const char* const reserved = "reserved";
```
stevecotton, line 28: Having an `enum` rather than an `enum class`, combined with reusing the same names (except for a different case) for `char*` consts sounds likely to cause bugs where the wrong case is used, but the compiler silently converts `char *` to int.

src/enum_maps/side_controller.hpp
```diff
@@ -0,0 +1,49 @@
...
+			if(value == values_[i])
+			{
+				return static_cast<TYPE>(i);
+			}
+		}
+		return INVALID;
+	}
+
+private:
+	static constexpr const std::array<const char*, 5> values_ = {invalid, none, human, ai, reserved};
```
stevecotton, line 48: I feel the MAKE_ENUM macro's technique of putting each pair on the same line is better than splitting them across separate definitions for enum, strings and mapping array.

src/gui/dialogs/multiplayer/mp_join_game.cpp
Pentarctagon: I don't know why a variety of places aren't using the CONTROLLER MAKE_ENUM in places like these. Also note that the wesnothd and client MAKE_ENUMs aren't even common code - they're copy/pastes of the same few lines into two separate headers.

src/enum_maps/side_controller.hpp
Pentarctagon: How would you prefer this be done instead then?

Pentarctagon:
> Is there a bug that requires this to be merged soon, or can it wait until later in the dev cycle, so that it's less likely that there will be backports to 1.16.x after this merges?

There is not. However I would prefer this to be merged sooner than later so that it doesn't accumulate merge conflicts over the coming months.

> Although it's easier to understand, this gives it its own handcoded class instead of sharing the common code of MAKE_ENUM. I'd prefer to have a template or even a macro, so that other enums that are converted to the new system will share common code instead of being cut&pastes.

The problem with MAKE_ENUM is that having common code is not especially useful if the common code is incomprehensible without digging through a bunch of Boost macros+documentation.

Pentarctagon:
> I'm missing something here, because users of MAKE_ENUM shouldn't need to look at the innards of the macro. If new functionality could be added to the existing MAKE_ENUM, would that satisfy your needs?

I could of course simply ignore it and move on, since technically MAKE_ENUM allows supporting the usecases it's needed for, but I also think it's a severely over complicated way to implement the functionality of "let's associate an enum and a string". I don't personally think it's worth anyone's time to add more functionality to or on top of MAKE_ENUM though.

> There's already an overload string_to_enum(const string&, NAME fallback) and a bool parse(templated string) in MAKE_ENUM that don't throw.

`bool parse(templated string)` I ignored since it requires an instance of the MAKE_ENUM to exist already, which it modifies if the string is a valid value. The overloaded `string_to_enum(const string&, NAME fallback)` I missed though.

src/gui/dialogs/game_stats.cpp
Pentarctagon: Fixed.

src/menu_events.cpp
Pentarctagon: Fixed.

src/enum_maps/side_controller.hpp
Pentarctagon: Fixed.

src/enum_maps/side_controller.hpp
```diff
@@ -0,0 +1,49 @@
...
+	static constexpr const char* const reserved = "reserved";
+
+	static const char* get_text(TYPE key)
+	{
+		return values_[key];
+	}
+
+	static TYPE get_enum(const std::string& value)
+	{
+		for(unsigned int i = 0; i < values_.size(); i++)
```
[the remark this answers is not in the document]
Pentarctagon: Yeah, that's left over from when I was initially thinking of using `map`s rather than the approach I went with. Not sure what a better name would be, maybe `string_enums`?

src/enum_maps/side_controller.hpp
```diff
@@ -0,0 +1,49 @@
...
+{
+public:
+	enum TYPE { INVALID = 0, NONE = 1, HUMAN = 2, AI = 3, RESERVED = 4 };
+	static constexpr const char* const invalid = "invalid";
+	static constexpr const char* const none = "null";
+	static constexpr const char* const human = "human";
+	static constexpr const char* const ai = "ai";
+	static constexpr const char* const reserved = "reserved";
+
+	static const char* get_text(TYPE key)
```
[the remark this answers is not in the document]
Pentarctagon: Updated to `to_string()`.

For the interface, not generally, though having to do:
```
team::CONTROLLER::enum_to_string(team::CONTROLLER::AI);
```
in menu_events.cpp for example to get the string felt rather cumbersome.

src/scripting/lua_team.cpp
```diff
@@ -80,7 +80,7 @@ static int impl_side_get(lua_State *L)
 	return_string_attrib("faction", t.faction());
 	return_tstring_attrib("faction_name", t.faction_name());
 	return_string_attrib("color", t.color());
-	return_cstring_attrib("controller", t.controller().to_string().c_str());
+	return_cstring_attrib("controller", side_controller::get_string(t.controller()).c_str());
```
stevecotton, line 83: A question to possibly defer answering until a different PR: could this use `return_string_attrib` and omit the `.c_str()`? This is one of only two callers of `return_cstring_attrib`.

Pentarctagon: I'd guess probably, but I don't know for sure. Maybe @CelticMinstrel would know though.

CelticMinstrel: Yeah, I think that should work.

CelticMinstrel:
I don't quite get the point of this when `MAKE_ENUM` works perfectly well, but whatever.

Also, I don't agree with the decision to add an `INVALID` value. In my opinion, enums should only have the values that are valid. If in one place (such as `get_enum`) you need to signal that the result value is invalid, then I would just return `nullopt`.

Vultraz:
> then I would just return `nullopt`.

I second this. Now that we have `std::optional` a `nullopt` should be returned anywhere an invalid value is allowed instead of a bool or some other invalid value flag.

src/gui/dialogs/game_stats.cpp
```diff
@@ -67,8 +67,7 @@ unit_const_ptr game_stats::get_leader(const int side)
 
 static std::string controller_name(const team& t)
 {
-	static const std::array<t_string, 3> names {{_("controller^Human"), _("controller^AI"), _("controller^Idle")}};
-	return "<span color='#808080'><small>" + names[t.controller().v] + "</small></span>";
+	return "<span color='#808080'><small>" + std::string(side_controller::get_string(t.controller())) + "</small></span>";
```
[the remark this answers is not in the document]
Pentarctagon: Unintentionally; fixed.

src/scripting/lua_team.cpp
Pentarctagon: Alright, changed then.

Pentarctagon:
Updated to use std::optional instead of an INVALID enum value.

Vultraz:
Also, one last thing, you could probably use `std::string_view` instead of `const char*` (it has constexpr ctors).

Pentarctagon:
Does config handle string_view automatically, since a string_view can't be implicitly converted to string like const char* can?

This would also end up being rather cumbersome on the server-side with simple_wml since string_view doesn't guarantee its data is null terminated: https://en.cppreference.com/w/cpp/string/basic_string_view/data
> Unlike std::basic_string::data() and string literals, data() may return a pointer to a buffer that is not null-terminated. Therefore it is typically a mistake to pass data() to a routine that takes just a const CharT* and expects a null-terminated string.

While simple_wml only accepts attributes with the types of int, const char*, and string_span.

CelticMinstrel:
I'm also curious what "reserved" is for?

Vultraz:
> Does config handle string_view automatically, since a string_view can't be implicitly converted to string like const char* can?

`std::string` has a ctor that takes a `string_view`.

Pentarctagon:
IIRC "Reserved" is for when someone is hosting a reloaded game that had other human players in it.

Vultraz:
(That is, you should be able to just do `return std::string(return values_[static_cast<int>(key)]);` if it doesn't do that implicitly.)

Vultraz:
Why is that a problem?

Pentarctagon:
Everywhere that would assign it to a config attribute would need to explicitly construct a string.
`cfg["controller"] = std::string(side_controller::human)`
instead of
`cfg["controller"] = side_controller::human`

Vultraz:
That should probably be added to `config` if it's not there already.

Pentarctagon:
I think that would be something for a separate PR, especially since it would need to also address simple_wml, and I have no particular desire to try touching that.

src/gui/dialogs/game_stats.cpp
```diff
@@ -67,8 +67,8 @@ unit_const_ptr game_stats::get_leader(const int side)
 
 static std::string controller_name(const team& t)
 {
-	static const std::array<t_string, 3> names {{_("controller^Human"), _("controller^AI"), _("controller^Idle")}};
-	return "<span color='#808080'><small>" + names[t.controller().v] + "</small></span>";
+	static const std::array<t_string, 5> names {{_("controller^Idle"), _("controller^Human"), _("controller^AI"), _("controller^Reserved")}};
```
Vultraz, line 70: Why 5? 🤔

Pentarctagon: I forgot to adjust it after removing the INVALID enum value.  Is there a reason to not do just `return values_.size();` though?

src/enum_maps/base_enum.hpp
```diff
@@ -0,0 +1,52 @@
...
+		return T::values_[static_cast<int>(key)];
+	}
+
+	template<class Enum>
+	static std::optional<Enum> get_enum(const std::string& value)
+	{
+		for(unsigned int i = 0; i < T::values_.size(); i++)
+		{
+			if(value == T::values_[i])
+			{
```
stevecotton, line 35: Please run this through clang-format (but be wary, it does bad changes to the macro definition).

stevecotton:
Thanks, I like the direction that this is going in.

Adding a namespace and comments to the new template would be good.

Pentarctagon:
Done.

Vultraz:
Proposal:
```cpp
template<typename T>
struct enum_base
{
	static std::string get_string(typename T::type key)
	{
		return std::string{T::values[static_cast<int>(key)]};
	}

	static std::optional<typename T::type> get_enum(const std::string_view value)
	{
		for(unsigned int i = 0; i < T::values.size(); i++) {
			if(value == T::values[i]) {
				return static_cast<typename T::type>(i);
			}
		}

		return std::nullopt;
	}

	static constexpr std::size_t size()
	{
		return T::values.size();
	}
};

struct side_controller
{
	static constexpr std::string_view none = "null";
	static constexpr std::string_view human = "human";
	static constexpr std::string_view ai = "ai";
	static constexpr std::string_view reserved = "reserved";

	static constexpr std::array<std::string_view, 4> values{none, human, ai, reserved};
	enum class type { none = 0, human = 1, ai = 2, reserved = 3 };
};
```

This uses a data pod design instead of a derived class. `enum_base` can take any class `T` which has a `type` scoped enum and a `values` array member.

Use would be like so:
```cpp
	std::cerr << static_cast<int>(enum_base<side_controller>::get_enum("null").value()) << std::endl;
	std::cerr << enum_base<side_controller>::get_string(side_controller::type::none) << std::endl;
```

Pentarctagon:
It's a bit more typing I suppose. Also afaik string_view is effectively incompatible with simple_wml.

CelticMinstrel:
> Also afaik string_view is effectively incompatible with simple_wml.

simple_wml is literally based on string_view, right? Or rather, string_span, which is the same as string_view but hand-rolled. So it seems to me that making it compatible shouldn't be a lot of work.

Pentarctagon:
Yeah, it'd be great to use string_view instead of the custom string_span, I just really don't want to need to do that as part of this PR.

Pentarctagon:
Beyond that, I don't really have much of a preference either way, so I'll leave it up to @stevecotton .

src/string_enums/base_enum.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+namespace string_enums
+{
+/**
+ * The base template for associating string values with enum values.
+ * Implementing classes should not set custom int values for their enum.
+ * The number of enum values must match the number of elements in the @a values_ array.
+ */
+template<class T>
+struct enum_base
```
stevecotton, line 28: Please swap the order of the words either in this line, or in the filename; otherwise we have both `enum_base` and `base_enum.hpp`.

stevecotton:
At the moment I'm reviewing by porting src/tests/test_make_enum.cpp to use this. Would you consider adding more functions to the template class to better integrate with `config` and the iostreams operators?

> simple_wml

Having asked for more support for the normal `config` class, I'm going the opposite way for simple_wml. Assuming that anywhere that's using simple_wml is doing it for performance reasons, I think it's good to have `explicit` constructors making expensive operations look expensive.

Pentarctagon:
> At the moment I'm reviewing by porting src/tests/test_make_enum.cpp to use this. Would you consider adding more functions to the template class to better integrate with config and the iostreams operators?

Can you be more specific?

> Having asked for more support for the normal config class, I'm going the opposite way for simple_wml. Assuming that anywhere that's using simple_wml is doing it for performance reasons, I think it's good to have explicit constructors making expensive operations look expensive.

I'm not sure what you're referring to.

stevecotton:
> > Would you consider adding more functions to the template class to better integrate with config and the iostreams operators?
> Can you be more specific?

Adding support for
```
	std::stringstream ss;
	ss << e;
	BOOST_CHECK_EQUAL (ss.str(), "name3");
```
and
```
	config cfg;
	auto e = foo::enumname::TYPE::name;
	cfg["t1"] = e;
	cfg["t2"] = e.to_string();
```

stevecotton:
> This uses a data pod design instead of a derived class. `enum_base` can take any class `T` which has a `type` scoped enum and a `values` array member.
> 
> Use would be like so:
> 
> ```c++
> enum_base<side_controller>::get_enum("null").value() << std::endl;
> enum_base<side_controller>::get_string(side_controller::type::none) << std::endl;
> ```

I like Vultraz's suggestion, it adds type checking to `get_string` so that the compiler catches the mistake here:
```
	first_enum::get_string(first_enum::TYPE::con3); // OK
	second_enum::get_string(first_enum::TYPE::con3); // compile error (good)
```

On top of the suggestion:
```
- class enum_base
+ struct enum_base : public T
```
and then
```
-struct side_controller : string_enums::enum_base<side_controller>
+struct side_controller_defines
 {
...
-       IMPL_GET_ENUM(TYPE)
 };
+
+using side_controller = string_enums::enum_base<side_controller_defines>;
```
That's a drop-in replacement that doesn't affect the code already in this PR that uses `side_controller`.

Vultraz:
> On top of the suggestion:

What do we get with inheritance? 🤔 I like your idea of a type alias to avid having to write out `enum_base` every time, but I don't think we actually need it to inherit from `T`.

stevecotton:
The inheritance means the same name, `side_controller`, in both `side_controller::get_enum(...)` and `side_controller::TYPE::ai`.

Vultraz:
You don't need that if you have a type alias.

Vultraz:
Yeah, I the stringstream check isn't needed since we'd be using strings/char arrays/string views directly now instead of the object.

src/team.hpp
```diff
@@ -264,31 +258,35 @@ class team
...
 
 	void set_local(bool local) { info_.is_local = local; }
-	void make_human() { info_.controller = CONTROLLER::HUMAN; }
-	void make_ai() { info_.controller = CONTROLLER::AI; }
+	void make_human() { info_.controller = side_controller::TYPE::HUMAN; }
+	void make_ai() { info_.controller = side_controller::TYPE::AI; }
 	void change_controller(const std::string& new_controller) {
-		info_.controller = CONTROLLER::AI;
-		info_.controller.parse(new_controller);
+		std::optional<side_controller::TYPE> type = side_controller::get_enum(new_controller);
```
Vultraz, line 282: This (and other similar cases) is a perfect use for `optional::value_or`, since the optional will be a nullopt if `new_controller` is invalid:
```cpp
info_.controller = side_controller::get_enum(new_controller).value_or(side_controller::TYPE::AI);
```

src/team.cpp
```diff
@@ -243,17 +243,21 @@ void team::team_info::read(const config& cfg)
 		support_per_village = lexical_cast_default<int>(village_support, game_config::village_support);
 	}
 
-	controller = team::CONTROLLER::AI;
-	controller.parse(cfg["controller"].str());
+	std::optional<side_controller::TYPE> type = side_controller::get_enum(cfg["controller"].str());
```
Vultraz, line 246: `optional::value_or` can be used here too.

Vultraz: Also, while this can be replaced with a one-liner, I would recommend using `auto` instead of `std::optional<side_controller::TYPE>` other places where you need a variable.

src/string_enums/base_enum.hpp
Pentarctagon: Done.

src/team.hpp
Pentarctagon: Done.

src/team.cpp
Pentarctagon: Done.

Pentarctagon:
Also, for the assignment operator for config, that works currently since it's passed a struct that extends `enum_tag`, but is there a way to have it deduce the containing object's type from any particular `enum class` that's passed in? The `enum class` itself doesn't extend anything or have any implemented methods.

Pentarctagon:
Updated the implementation to match the vultraz+octalot changes above.  A couple comments:
* I left things as `const char*`/`const std::string&` since I don't want to deal with replacing simple_wml's string_span with string_view - while on the surface it sounds simple since they are both meant to do a similar thing, everything I've ever heard about simple_wml is that it's very difficult to work with and I don't want to spend a bunch of time figuring out how it works and chasing down obscure issues unless using string_view is considered a hard requirement of merging this.
* The `public T` of `struct enum_base : public T` is there since it's what allows using, for example, `side_controller::type::NONE` rather than `side_controller_defines::type::NONE` which while not really significant in terms of extra typing is (IMO) beneficial for consistency.

So I think what remains is the two things octalot brought up in https://github.com/wesnoth/wesnoth/pull/6423#issuecomment-1017454129:
* The first wouldn't be applicable anymore since there is no longer any instance of the struct created to be provided to the `<<` operator.
* The second may (?) be possible, but to be honest I think the simplest and most elegant solution is to just call get_string where necessary rather than try to cram it into a templated method.

CelticMinstrel:
> to be honest I think the simplest and most elegant solution is to just call get_string where necessary rather than try to cram it into a templated method.

The only possible problem with this approach is that if you _forget_ to call `get_string()`, what happens? Does it convert it to a number instead? Maybe it's fine since it's an enum class.

I definitely prefer the templated assignment operator, but if it's a compile error, I suppose it's acceptable.

Pentarctagon:
`enum class` prevents an implicit conversion to an int, so it would be a compiler error.

src/string_enums/enum_base.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include <optional>
+#include <string>
+
+namespace string_enums
+{
+/**
+ * The base template for associating string values with enum values.
+ * Implementing classes should not set custom int values for their enum.
+ * The number of enum values must match the number of elements in the @a values array.
```
CelticMinstrel, line 25: Couldn't this be enforced with a static assert? Something like…

```cpp
static_assert(T::values.size() == T::type::ENUM_MAX);
```

Although that does require adding a redundant ENUM_MAX entry to every enum, so maybe that's unacceptable…

src/string_enums/enum_base.hpp
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Convert a string into its enum equivalent.
+	 *
+	 * @param value The string value to convert.
+	 * @return The equivalent enum or std::nullopt.
+	 */
+	static std::optional<typename T::type> get_enum(const std::string value)
+	{
+		for(unsigned int i = 0; i < T::values.size(); i++) {
+			if(value == T::values[i]) {
```
CelticMinstrel, line 50: Come to think of it, this will break if the same string appears twice in the list… obviously there's not much to be done about that, but it could be added to the list of requirements in the doc comment at the top.

src/string_enums/enum_base.hpp
Pentarctagon: I had thought about something similar, but instead of doing an assert, define the array as:
`static constexpr std::array<const char*, static_cast<int>(type::ENUM_MAX)> values{ none, human, ai, reserved };`

I'd been trying to find it there was some built-in way in C++ to find the number of elements in an enum, but it doesn't look like such a feature exists.

CelticMinstrel: Yeah, the only way that I know of is adding an extra value to the enum like I mentioned.

stevecotton: > I'd been trying to find it there was some built-in way in C++ to find the number of elements in an enum, but it doesn't look like such a feature exists.

Looks like it may eventually be provided by the Reflection TS, but no compiler seems to support that yet.

src/server/wesnothd/game.hpp
```diff
@@ -15,11 +15,11 @@
 
 #pragma once
 
+#include "string_enums/side_controller.hpp"
```
stevecotton, line 18: Should sort later alphabetically, now that it's not called enum_maps.

Sorry for such a trivial comment, but small fixes are good.

src/editor/map/map_context.hpp
```diff
@@ -25,8 +25,9 @@
 #include "team.hpp"
 #include "tod_manager.hpp"
 #include "units/map.hpp"
-#include <optional>
+#include "string_enums/side_controller.hpp"
```
stevecotton, line 28: Please put this one in its alphabetical place too.

stevecotton:
I like it, but I also like MAKE_ENUM (at least, using it rather than trying to edit the Boost-based implementation). I'm generally in favour of merging it, but give everyone a couple more days to see if anything else comes to mind first.

> Also, barring any issues with this approach, I'd probably (over the course of 1.17) convert other uses of MAKE_ENUM to this approach with the goal of eventually removing MAKE_ENUM entirely.

I think you're committing yourself to that ;-D .

src/editor/map/map_context.hpp
Pentarctagon: Done.

src/server/wesnothd/game.hpp
Pentarctagon: Done.

src/string_enums/enum_base.hpp
Pentarctagon: Done.

src/string_enums/enum_base.hpp
Pentarctagon: Added.

Vultraz:
Posted this on Discord but adding it here too. We might consider this header lib: https://github.com/Neargye/magic_enum

Pentarctagon:
Discussed more there, but the main issue brought up being that it requires gcc 9+, while Wesnoth's current minimum gcc version is 7+.

PR #6425, opened 2022-01-08
[1.16.3] Fix deprecation of [advancefrom], mark as already removed

### LINKED ISSUE #6204: [advancefrom] does nothing in 1.16
Affects 1.15.18+dev, has been around since before 1.16 beta 1.

`[unit_type] [advancefrom]` in legacy WML such as the following does nothing even though the game purportedly has a compatibility path to make it emit appropriate `[modify_unit_type]` WML during add-on initialization.

```
    [advancefrom]
        unit=Dwarvish Thunderer
    [/advancefrom]
```

The add-on campaign "Invasion from the Unknown" in particular has a Dwarvish Flamethrower unit type which uses `[advancefrom]` to make it a possible advancement from the mainline L1 Dwarvish Thunderer. This works in 1.14 but not 1.16.

(Yes, I realize `[advancefrom]` is deprecated. The point is that it was supposed to be removed in 1.17.x, not 1.15.x, since `[modify_unit_type]` was first introduced in the latter.)

stevecotton:
Postponed until 1.16.3 due to string freeze.
Closes #6204.
Supersedes PR #5922, which is why it was opened as a draft before 1.16.2 released.

The existing code searched for `[units][advancefrom]` instead of
`[unit_type][advancefrom]`, so wasn't triggering. It was planned
for removed in 1.17, so simply mark it as already-removed instead.

The deprecation warning is shown on stdout/stderr, but not in the
in-game UI, so it's generally not visible to players. I think it
would be better if it was visible, but OTOH it's something that's
easy for someone to check all add-ons for with a simple grep.

### DIFF
--- a/src/game_config_manager.cpp
+++ b/src/game_config_manager.cpp
@@ -21,6 +21,7 @@
 #include "cursor.hpp"
 #include "events.hpp"
 #include "formatter.hpp"
+#include "formula/string_utils.hpp"
 #include "game_classification.hpp"
 #include "game_config.hpp"
 #include "game_version.hpp"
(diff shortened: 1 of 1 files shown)


stevecotton:
@CelticMinstrel are you happy with this, please?

CelticMinstrel:
I don't really have any objections.

PR #6428, opened 2022-01-09
UtBS S08 Kromph and Dust Devil able to trigger Eloh's event

### LINKED ISSUE #6211: UtBS S8: Kromph fails to trigger events
 - 1.15.8
(skipping the rest because irrelevant)

### Describe the bug
Kromph (a Flesh Golem the play can find in this scenario) doesn't trigger anything when leaving the caves, unlike every singly other unit. Technically he could move up to the enemy keep before the enemy leader spawns and with a bit of luck instantly kill him (because berserk).

### To Reproduce

Start the scenario, get Kromph, move him outside of the caves. Will provide a save file later so you can skip a lot of steps and turns.

### Expected behavior
Kromph triggers Elohs Event.

### Screenshots
grafik
grafik

### Savefiles
[UtBS-Out of the Frying Pan-Auto-Save40.gz](https://github.com/wesnoth/wesnoth/files/7373266/UtBS-Out.of.the.Frying.Pan-Auto-Save40.gz)
[UtBS-Out of the Frying Pan Turn 48.gz](https://github.com/wesnoth/wesnoth/files/7373265/UtBS-Out.of.the.Frying.Pan.Turn.48.gz)
[UtBS-Out of the Frying Pan Turn 51.gz](https://github.com/wesnoth/wesnoth/files/7373279/UtBS-Out.of.the.Frying.Pan.Turn.51.gz)

### Additional context
It is likely that he wasn't accounted for in the Event. Accounting for him should fix it.

Descacharrado:
The variable talking_unit is used instead of unit to set the speaker in many messages.

The talking unit that talks on behalf of the monsters can be, in this order: An elf in the map, a recalled elf, a forcedly spawned elf (Lines 3007-3067)

That unit will move to the dunes to fit the dialogue (Lines 3124-3137)

EDIT: fixed on 061a42b:

I found that code at line 3024 was not searching for units on the recall list
Also if a unit was on 19 20 previously, it would get the talker role, even if it was not intended.
Also the unit which gets created in case everything else fails could get stuck in a wall

Another thing I found testing is that skeletons in one of the events can get stuck on the wall if you walk to their spawning positions maybe fixable on another PR.

Closes #6211

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
@@ -3011,7 +3011,8 @@
                 type=Dust Devil,Flesh Golem
             [/have_unit]
             [then]
-                # if the player only has Zhul and Kaleh, a new unit has to be recalled or respawned, else reassign. Counting Nym aswell to make sure
+                # if the player only has Zhul and Kaleh, a new unit has to be recalled or respawned, else reassign. Nym should never be
+                # on the map when this is triggered but is being counted anyway to prevent future problems.
                 [role]
                     race=quenoth,elf
                     [not]
@@ -3052,7 +3053,7 @@
                     [/else]
                 [/role]
 
-                # There will always be a talker role unit here, wheter it has been recalled or "recruited"
+                # There will always be a talker role unit here, whether it has been recalled or "recruited"
                 [store_unit]
                     [filter]
                         role=talker
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
@@ -3022,6 +3022,7 @@
                     [else]
                         [if]
                             [have_unit]
+                                search_recall_list=yes
                                 x,y=recall,recall
                                 level=1
                                 race=elf,quenoth
@@ -3033,12 +3034,19 @@
                                     level=1
                                 [/recall]
                                 [role]
-                                    x,y=19,20
+                                    race=elf,quenoth
+                                    level=1
+                                    [not]
+                                        id=Zhul,Nym,Kaleh
+                                    [/not]
                                     role=talker
                                 [/role]
                             [/then]
                             [else]
                                 {UNIT 1 (Quenoth Fighter) 19 20 (role=talker)}
+                                [+unit]
+                                    passable=yes
+                                [/unit]
                             [/else]
                         [/if]
                     [/else]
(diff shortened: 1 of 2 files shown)


nemaara (approved):

nemaara:
Maybe wants @knyghtmare to look too.

knyghtmare:
> Maybe wants @knyghtmare to look too.

Seems like it can be merged in. I only commented.

data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
```diff
@@ -3002,11 +3002,77 @@
...
         [/filter]
 
+        # in case Kromph or the dust devil trigger the event the unit variable is set to track a different unit
+        [if]
+            [have_unit]
+                x,y=$x1,$y1
+                type=Dust Devil,Flesh Golem
+            [/have_unit]
+            [then]
+                # if the player only has Zhul and Kaleh, a new unit has to be recalled or respawned, else reassign. Counting Nym aswell to make sure
```
Wedge009, line 3014: aswell -> 'as well'

data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
```diff
@@ -3002,11 +3002,77 @@
...
+                                {UNIT 1 (Quenoth Fighter) 19 20 (role=talker)}
+                                [+unit]
+                                    passable=yes
+                                [/unit]
+                            [/else]
+                        [/if]
+                    [/else]
+                [/role]
+
+                # There will always be a talker role unit here, wheter it has been recalled or "recruited"
```
Wedge009, line 3055: wheter -> whether

Wedge009:
One thing I noticed while testing this is that the Dust Devil is excluded from triggering pretty much any event in S8, and probably most other scenarios too. But I suppose that generally isn't a problem except in the unusual circumstance where it travels far beyond the rest of Kaleh's party, and it only seems to be an issue in special cases such as what #6211 reports.

Anyway, the changes seem to work for me. Unfortunately the back-port will have to wait till after the string freeze since it does touch one translated string - even though no words are meaningfully changed here.

Descacharrado:
> One thing I noticed while testing this is that the Dust Devil is excluded from triggering pretty much any event in S8, and probably most other scenarios too. 

I will test all S08s events thoroughly and open an issue with all weird things I find, then proceed to fix them if no one else does, as I said:

> Skeletons in one of the events can get stuck on the wall if you walk to their spawning positions

Wedge009:
1.16.2 string freeze is over - are we waiting for anything before merging this?

Wedge009:
Ah, I just noticed this has already been approved and I did test the change beforehand, so will merge.

PR #6433, opened 2022-01-11
Fix Winged Sceptre bug in WC

Discontinuum:
In the old WC here there was wesnoth.unit_defense function, so it should be wesnoth.units.chance_to_be_hit. In the current code it's wesnoth.units.defense_on which returns actual defense (50,60,40, etc), so the code checked if the defense of the unit is better than given by the artifact and worsened it.


knyghtmare:
> In the old WC here there was wesnoth.unit_defense function, so it should be wesnoth.units.chance_to_be_hit. In the current code it's wesnoth.units.defense_on which returns actual defense (50,60,40, etc), so the code checked if the defense of the unit is better than given by the artifact and worsened it.

I have been trying to reproduce the reported winged scepter bug for a while now but my luck kept giving me Troll's blood everytime. I will test out your code and if it solves it, I will merge it in. 
I wonder if this should be backported as well? hmm...

knyghtmare (approved):
Good job!

PR #6436, opened 2022-01-11
HttT S04, S05b & S06 Role code enhancement

### LINKED ISSUE #6388: HttT S04, S05b, S06: Code around role=Advisor is a bad example
Noticed while looking for example code for the `[role]` tag in 1.16.1+dev (3cce6719e2bdbb2a1a0f7fe2a154d1064073a1a2). There's a `role=Advisor`, which has the comment below in S05b. While it works, there's a lot of odd issues that should be cleaned up; I'm putting them all in one issue because they'd all need testing together.
```
    # This appoints the Advisor role to our most advanced unit coming out of The Bay of Pearls. This would be
    # the same as the initial Advisor we would appoint if we took the path through Muff Malal's Peninsula, and
    # will be the initial Advisor recalled at the start of The Siege of Elensefar.
```

S04 assigns a unit as `role=Supporter`, which is used exactly once, in a `[message]` immediately after the assignment. The list isn't quite the same as the later scenarios' `role=Advisor`, as it includes no merfolk (and in S04 it shouldn't), but it seems odd to only use it once.

S05a seems okay. It assigns `role=Advisor` along with `[auto_recall][/auto_recall]` even though the role hasn't been assigned in a previous scenario; however in this case it's in a `prestart` event and the `[role]` tag will search the recall list, so `auto_recall` makes sense. The `type=` list includes merfolk, but that's okay as the starting encampment is on a beach.

S05b also assigns `role=Advisor`, and again includes merfolk in the `type=` list. The `ISLE_GALLEON_ARRIVE` macro will put the Advisor on-board the galleon, even if the Advisor is a merfolk. Worse, if the Advisor is a merfolk then the unit will be killed and not replaced with a copy from `iotd_recall_store`. If a merfolk gets the role then the player has probably already lost the campaign anyway (army too weak for S06), but while we're testing code changes for this issue then it's something that could be fixed along the way.

S06 has `#define ADVISOR` in the middle of an event. It works, but seems a bad coding style and messes up the indentation. Recalling a merfolk with full moves on turn 1 does give some time to get them in the water, but it still doesn't seem the best choice for the starting encampment.

Descacharrado:
>S04 assigns a unit as role=Supporter, which is used exactly once, in a [message] immediately after the assignment. The list isn't quite the same as the later scenarios' role=Advisor, as it includes no merfolk (and in S04 it shouldn't), but it seems odd to only use it once.

Now `type` is used inside `[message]` to avoid assigning a role used only once.

> S05a seems okay. It assigns role=Advisor along with [auto_recall][/auto_recall] even though the role hasn't been assigned in a previous scenario; however in this case it's in a prestart event and the [role] tag will search the recall list, so auto_recall makes sense. The type= list includes merfolk, but that's okay as the starting encampment is on a beach.

Agreed, S05a uses it properly.

> S05b also assigns role=Advisor, and again includes merfolk in the type= list. The ISLE_GALLEON_ARRIVE macro will put the Advisor on-board the galleon, even if the Advisor is a merfolk. Worse, if the Advisor is a merfolk then the unit will be killed and not replaced with a copy from iotd_recall_store. If a merfolk gets the role then the player has probably already lost the campaign anyway (army too weak for S06), but while we're testing code changes for this issue then it's something that could be fixed along the way.

Merfolk will now be unable to be the hidden advisors in S05b.

> S06 has #define ADVISOR in the middle of an event. It works, but seems a bad coding style and messes up the indentation. Recalling a merfolk with full moves on turn 1 does give some time to get them in the water, but it still doesn't seem the best choice for the starting encampment.

Moved the define outside, changed the priority list of the advisor (so merfolk will be a last resort).

Closes #6388

### DIFF
--- a/data/campaigns/Heir_To_The_Throne/scenarios/05b_Isle_of_the_Damned.cfg
+++ b/data/campaigns/Heir_To_The_Throne/scenarios/05b_Isle_of_the_Damned.cfg
@@ -57,11 +57,8 @@
 
             type="Elvish Sylph,Great Mage," +
                 "Elvish Marshal,Elvish Champion,Elvish Avenger,Elvish Sharpshooter,Elvish Shyde,Elvish Enchantress,Elvish Outrider,Paladin,Grand Knight,Mage of Light,Arch Mage,Silver Mage," +
-                "Merman Triton,Merman Hoplite,Merman Javelineer,Merman Entangler,Mermaid Diviner,Mermaid Siren," +
                 "Elvish Captain,Elvish Hero,Elvish Ranger,Elvish Marksman,Elvish Druid,Elvish Sorceress,Elvish Rider,Knight,Lancer,White Mage,Red Mage," +
-                "Merman Warrior,Merman Spearman,Merman Netcaster,Mermaid Priestess,Mermaid Enchantress," +
-                "Elvish Fighter,Elvish Archer,Elvish Shaman,Elvish Scout,Horseman,Mage," +
-                "Merman Fighter,Merman Hunter,Mermaid Initiate"
+                "Elvish Fighter,Elvish Archer,Elvish Shaman,Elvish Scout,Horseman,Mage"
             [else]
                 [unit]
                     side=1
--- a/data/campaigns/Heir_To_The_Throne/scenarios/06_The_Siege_of_Elensefar.cfg
+++ b/data/campaigns/Heir_To_The_Throne/scenarios/06_The_Siege_of_Elensefar.cfg
@@ -147,6 +147,43 @@
         {UNDERGROUND}
     [/time_area]
 
+#define ADVISOR
+    [role]
+        role=Advisor
+        [auto_recall][/auto_recall]
+
+        type="Elvish Sylph,Great Mage," +
+
+            "Elvish Marshal,Elvish Champion,Elvish Avenger,Elvish Sharpshooter,Elvish Shyde,Elvish Enchantress,Elvish Outrider," +
+            "Paladin,Grand Knight," +
+            "Mage of Light,Arch Mage,Silver Mage," +
+            "Highwayman,Fugitive,Huntsman,Ranger," +
+
+            "Elvish Captain,Elvish Hero,Elvish Ranger,Elvish Marksman,Elvish Druid,Elvish Sorceress,Elvish Rider," +
+            "Knight,Lancer," +
+            "White Mage,Red Mage," +
+            "Bandit,Outlaw,Trapper," +
+
+            "Elvish Fighter,Elvish Archer,Elvish Shaman,Elvish Scout," +
+            "Horseman," +
+            "Mage," +
+            "Thug,Footpad,Poacher" +
+
+            # Merfolk at the bottom because they are not as useful as the other options with this castle placement.
+            "Merman Triton,Merman Hoplite,Merman Javelineer,Merman Entangler,Mermaid Diviner,Mermaid Siren," +
+            "Merman Warrior,Merman Spearman,Merman Netcaster,Mermaid Priestess,Mermaid Enchantress," +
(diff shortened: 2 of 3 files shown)


Descacharrado:
There is another comma error I found, will fix later.

stevecotton:
> Advisor gets asigned on S04, after the player has won the scenario (and it does not include merfolk),

I'm confused, I can't find the code that does that assignment. So I think it's still possible to have S05a assign the role to a merfolk, and then have that merfolk chosen in S06 instead of a better suited unit who was recruited and trained in S05a.

data/campaigns/Heir_To_The_Throne/scenarios/05b_Isle_of_the_Damned.cfg
```diff
@@ -57,11 +57,8 @@
 
             type="Elvish Sylph,Great Mage," +
```
knyghtmare, line 58: I played this scenario recently and I remember that I cannot recall a single elf here. Only merfolk recalls and outlaw recruits...so why are the elves even here in the advisor creation?

data/campaigns/Heir_To_The_Throne/scenarios/06_The_Siege_of_Elensefar.cfg
```diff
@@ -174,20 +157,22 @@
...
             "Bandit,Outlaw,Trapper," +
 
             "Elvish Fighter,Elvish Archer,Elvish Shaman,Elvish Scout," +
             "Horseman," +
             "Mage," +
-            "Merman Fighter,Merman Hunter,Mermaid Initiate," +
-            "Thug,Footpad,Poacher"
+            "Thug,Footpad,Poacher" +
+
+            # Merfolk at the bottom because they are not as useful as the other options with this castle placement.
```
knyghtmare, line 172: I agree with you. Apart from S4, S5B, S11, S18 (just one level 3), S19C, S20B...Merfolk don't really have much use in HttT.

knyghtmare (commented):
Has this been tested? The code on the first file became hard to read with the reds and greens being scattered all over. Anyways, good going with the prioritization of the supporters.

Descacharrado:
> > Advisor gets asigned on S04, after the player has won the scenario (and it does not include merfolk),
> 
> I'm confused, I can't find the code that does that assignment. So I think it's still possible to have S05a assign the role to a merfolk, and then have that merfolk chosen in S06 instead of a better suited unit who was recruited and trained in S05a.

Ah no you are right, it gives a "supporter" role and not advisor, mb

data/campaigns/Heir_To_The_Throne/scenarios/05b_Isle_of_the_Damned.cfg
Descacharrado: They are given the role on prestart, before they are disallowed, that specific unit will be the one talking on behalf of the whole ship.

Descacharrado:
> Has this been tested? The code on the first file became hard to read with the reds and greens being scattered all over. Anyways, good going with the prioritization of the supporters.

I always test things before PRing, sometimes I miss thigns tho.

Descacharrado:
I think a better solution for this may be defining a macro similar to the mercenary one on AToTB which has the code from S06 (the one I moved outside of the event) something like {NEED_ADVISOR}. I'll convert to draft until I do that, last thing we want is enhancing weird code with something which still is confusing. I'll look on the other scenarios in case this same role thing happens.

stevecotton:
> Looks like it automatically reassigns (as can be seen in the images below), so I was previously completely wrong. Sorry for the missunderstanding.

I wasn't expecting that either - seems there's an undocumented `reassign` attribute which defaults to `yes`, added in 1.13.6 (b064328102594473965a2b884d5795d551bc163c).

stevecotton (approved):

stevecotton:
@knyghtmare are you happy for me to merge this?

PR #6437, opened 2022-01-11
Run the WML unit tests on macOS.

Pentarctagon:
Not really sure why they're able to run now.

---

Will merge after CI passes.

### DIFF
--- a/.github/workflows/ci-scripts/macos.sh
+++ b/.github/workflows/ci-scripts/macos.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-echo ~
-echo $PWD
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Eh? They've always been able to run on MacOS though? Although maybe the special `-p Xcode` flag is no longer necessary, that could be the difference…

Pentarctagon:
On the CI at least, they weren't able to be run since for some reason even just attempting to start the wesnoth executable would just make it hang. That was on Travis though, so either there's something different about the Github Actions environment or something was accidentally fixed in the last few years.

Issue #4667, opened 2019-12-23
OOS in replays from victory events


gfgtdf:
Synced choices from victory events cause OOS in replays, wesnoth 1.14.5, play the following scenario win with any side and see hoe the replay gives an OOS at the end:

```
[multiplayer]
    name="01_STARTUNIT_TEST"
    id= "01_STARTUNIT_TEST"
    map_data = "Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh
Gs^Vh, Gs^Vh, 2 Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh
Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh
Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh
Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh
Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh
Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, 1 Gs^Vh, Gs^Vh, Gs^Vh
Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh, Gs^Vh
"
    {DEFAULT_SCHEDULE}
    [side]
        side=1
        id= "leader1"
        controller=human
        team_name= "team1"
        type = "Elvish Archer"
        canrecruit="yes"
    [/side]
    [side]
        side=2
        id= "leader2"
        controller=human
        team_name= "team2"
        type = "Elvish Archer"
        canrecruit="yes"
    [/side]
	[event]
		name="victory"
		[set_variable]
			name="aaa"
			rand="1..2"
		[/set_variable]
	[/event]
[/multiplayer]
```

This OOS only effects replays not mp games.

stevecotton:
[...]

PR #6438, opened 2022-01-12
NR S01 Solve OOS by moving code to synced event

### LINKED ISSUE #6268: NR S1 Replay: OOS After accomplishing the final objective
 - 1.16.0
 - Steam
 - Windows 10

### Describe the bug
While watching the replay of NR S1, after the final objective is accomplished there is an OOS error.

### To Reproduce

### Expected behavior
No OOS Error.

### Screenshots
image

### Savefiles
[NR-Breaking the Chains replay 20211025-035004.gz](https://github.com/wesnoth/wesnoth/files/7491001/NR-Breaking.the.Chains.replay.20211025-035004.gz)

Descacharrado:
[This](https://wiki.wesnoth.org/EventWML#Multiplayer_safety) says the victory event is unsynced so I just moved the code to the enemies defeated event. Tested that everything works the same and replay goes all the way without OOSs.

Closes #6268

### DIFF
--- a/data/campaigns/Northern_Rebirth/scenarios/01_Breaking_the_Chains.cfg
+++ b/data/campaigns/Northern_Rebirth/scenarios/01_Breaking_the_Chains.cfg
@@ -341,15 +341,6 @@
     [event]
         name=enemies defeated
 
-        [endlevel]
-            bonus=yes
-            {NEW_GOLD_CARRYOVER 40}
-        [/endlevel]
-    [/event]
-
-    [event]
-        name=victory
-
         [message]
             speaker=Tallin
             message= _ "<i>Yes</i>! We did it! We are free!"
@@ -392,6 +383,11 @@
         [/move_unit_fake]
 
         [redraw][/redraw]
+
+        [endlevel]
+            bonus=yes
+            {NEW_GOLD_CARRYOVER 40}
+        [/endlevel]
     [/event]
 
     # Death events


knyghtmare (approved):

stevecotton:
> [This](https://wiki.wesnoth.org/EventWML#Multiplayer_safety) says the victory event is unsynced so I just moved the code to the enemies defeated event. Tested that everything works the same and replay goes all the way without OOSs.
> 
> Closes #6268

That's worrying - the docs say that `victory` **is** synced from 1.13.10 onwards, and that's part of the reason for adding a `local_victory` event. However, this bug and bugfix show that the docs are wrong.

Descacharrado:
> That's worrying - the docs say that `victory` **is** synced from 1.13.10 onwards, and that's part of the reason for adding a `local_victory` event. However, this bug and bugfix show that the docs are wrong.

Maybe the problem then resides elsewhere and these fixes (talking about #6439) should not be merged?

nemaara (approved):

Wedge009:
This and #6439 have been around for a while. I'm guessing we're stalled on whether or not network synchronisation is actually working as documented. I admit I don't know much about the network side of things and neither to I actively play MP (besides the occasional attempts to test things) nor how it relates to what seems to be a SP replay.

As steve mentioned, the wiki states that the event should be synchronised from 1.13.10 onwards. The change log for 1.13.10 (which, admittedly, may not be a comprehensive list of changes) doesn't make any mention of this apparently significant change to event handling. The only reference to the new `local_victory` event type I could find - aside from test code - is in `playsingle_controller.cpp`:
https://github.com/wesnoth/wesnoth/blob/034453d6977962b58c9e8d9d6c3a204679f0f48c/src/playsingle_controller.cpp#L309

According to git blame, @gfgtdf made that change since 1.13.5 - look at the change log for that release shows big changes... but I don't know that it necessarily includes what we're looking for here, regarding synchronisation.

Pentarctagon:
Is the victory event synced or not synced in 1.14?

If it's also not synced in 1.14, then either then wiki was updated incorrectly or it being synced was somehow broken almost immediately after being implemented, but either way making it synced would be an independent feature request to be made and discussed.

If it is synced in 1.14, then it's a bug that should be fixed in the engine rather than as it is in this PR.

Descacharrado:
> Is the victory event synced or not synced in 1.14?

I will test 1.14, and if the issue is present there aswell, I will try bisecting to see where the OOSs started occurring. I agree with this:

> If it is synced in 1.14, then it's a bug that should be fixed in the engine rather than as it is in this PR.

Pentarctagon:
Well, if 1.16.0 has the OOS and 1.14.17 has the OOS, then it's also a question of "was this ever really made to be synced?". There needs to actually be a good starting commit for a bisect.

Descacharrado:
> Well, if 1.16.0 has the OOS and 1.14.17 has the OOS, then it's also a question of "was this ever really made to be synced in the first place?". There needs to actually be a good starting commit for a bisect.

Yes, I was gonna start trying with 1.13.10 which is supposedly when the victory event was synced. Of course, If I dont find any versions on which the event was synced, it is easy to asume that it never was.

Descacharrado:
The OOS happened in 1.13.10 too

Pentarctagon:
So I guess this was never actually a thing that happened. So I'd say this fix is fine - can you also update the wiki as well though?

The local_defeat/local_victory events could probably just be removed as well, but that'd be a separate thing.

Descacharrado:
Sure, I want to test a few more things regarding this OOS issue tho, for example seeing how many instances of the same bug there are in mainline, and testing if there is a common factor. Having your approvals I'll merge.

Descacharrado:
> I think the events really are synced, and it's a bug with the replay code - #4667.

If that is the case, when solved we should/could revert this.

PR #6439, opened 2022-01-12
UtBS S08 OOS and wall-stuck units fixes

### LINKED ISSUE #6267: UtBS S8 Out of the Frying Pan Replay: OOS after accomplishing the final objective
 - 1.16.0, replay is from 1.15.18
 - Steam
  - Windows 10

### Describe the bug
While watching the replay of UtBS S8 there is an OOS error after the final objective has been accomplished.

### To Reproduce

### Expected behavior
No OOS error.

### Screenshots
image

### Savefiles
[UtBS-Out of the Frying Pan replay 20211017-215947.gz](https://github.com/wesnoth/wesnoth/files/7490942/UtBS-Out.of.the.Frying.Pan.replay.20211017-215947.gz)

Descacharrado:
Addresses my comment on #6428:
> Another thing I found testing is that skeletons in one of the events can get stuck on the wall if you walk to their spawning positions maybe fixable on another PR.

13 units could get stuck in a wall if you knew where they would spawn (and had quick sun sylphs), now they will not (fixed the same way #6357 was fixed, lines 873, 893, etc.). 
Image below is an exaggeration (but was possible with enough sun sylphs).
wall

Also you could get past a trigger (without using debug mode line 2119).
Fixed OOS at victory (similar to how #6438 was fixed, lines 3511-3781).

I'll open a PR once this is merged to add {PASSABLE_HEX} to unit-utils and remove it from TRoW S19 and here.

Closes #6267

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
@@ -870,7 +870,7 @@
             image_mods={ARMOR_SHIFT_IPF}
         [/move_unit_fake]
 
-        {NAMED_UNIT 2 (Javelineer) 33 44 (Bellerin) ( _ "Bellerin") (upkeep,role,facing=free,human_scout,sw)}
+        {NAMED_UNIT 2 (Javelineer) 33 44 (Bellerin) ( _ "Bellerin") (upkeep,role,facing=free,human_scout,sw)}{PASSABLE_HEX}
 
         [move_unit_fake]
             type=Halberdier
@@ -890,7 +890,7 @@
             image_mods={ARMOR_SHIFT_IPF}
         [/move_unit_fake]
 
-        {NAMED_UNIT 2 {ON_DIFFICULTY (Longbowman) (Longbowman) (Master Bowman)} 34 43 (Othgar) ( _ "Othgar") (upkeep,role,facing=free,human_scout,sw)}
+        {NAMED_UNIT 2 {ON_DIFFICULTY (Longbowman) (Longbowman) (Master Bowman)} 34 43 (Othgar) ( _ "Othgar") (upkeep,role,facing=free,human_scout,sw)}{PASSABLE_HEX}
 
         [message]
             speaker=Durth
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
@@ -2116,7 +2116,7 @@
         name=moveto
 
         [filter]
-            x=5-10
+            x=4-10
             y=32-37
             side=1
             [not]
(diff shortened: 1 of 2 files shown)


data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
```diff
@@ -205,6 +205,12 @@
     # add more dirt at easier difficulties
     # set starting scenario objectives
 
+#define PASSABLE_HEX
+    # Adds passable attribute to a unit
+    [+unit]
+        passable=yes
```
knyghtmare, line 211: Better to put it as:
```ini
placement=map
passable=yes
```

> use placement=map and passable=yes instead.

<https://wiki.wesnoth.org/SingleUnitWML#Creation_Options>

knyghtmare (changes requested):

data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
Descacharrado: Isn't that redundant? You already need x and y for the {UNIT} macro so placement would just put them on the map somewhere passable, which is what passable=yes does. Please expatiate.

knyghtmare: > Isn't that redundant? You already need x and y for the {UNIT} macro so placement would just put them on the map somewhere passable, which is what passable=yes does. Please expatiate.

No, that's the currently used one. the deprecated ones are `map_passable` and `leader_passable`.
The wiki mentioned both have to there (Notice the "And") so pointed it out.

Descacharrado: But I am not using `map_passable` nor `leader_passable` inside the `placement` option, im using the `passable` option. And anyways the wiki says:

> placement: How the unit should be placed: can be one value or a comma-separated list of values. Default value is 'map,leader' for a leader given directly in [side], "" otherwise. By default, 'map,recall' is implicitly appended to the end of the list.

So placement=map is default and does not have to be specified, I think.

knyghtmare: If that's the case, then I guess it shouldn't be there. I add it in for works though (to avoid any bugs popping up later)

Descacharrado: > I add it in for my works though (to avoid any bugs popping up later).

Ah, I see, well I think this is a safe bet, anyways if a bug related to this is found later and im still over here, i'll try to fix it.

knyghtmare (approved):

data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
knyghtmare: Cool

Descacharrado:
Closing and separating features because of OOS fixes standstill.

PR #6441, opened 2022-01-12
[new feature 1.17] create a filter_distant for detect unit distant and test if enemy or not( more simple what filter_location)

newfrenchy83:
use a "filter_distant" equivalent ot "filter_adjacent" is simplest to write and more dynamic what [filter_location] because is it no more necessary to specifie side of unit.

More intessting for encode events in unit_type or [+abilities] with presence of allies or enemies unit in radius.

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
+++ b/data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
@@ -137,21 +137,15 @@
     [/dummy]
 #enddef
 
-#define ABILITY_SHADOW_VEIL_HELPER SIDE
+#define ABILITY_SHADOW_VEIL_HELPER
     [hides]
         id=did_shadow_veil_helper
         [filter]
-            [filter_location]
+            [filter_distant]
                 radius=5
-                [filter]
-                    ability=did_shadow_veil
-                    [filter_side]
-                        [allied_with]
-                            side={SIDE}
-                        [/allied_with]
-                    [/filter_side]
-                [/filter]
-            [/filter_location]
+                is_enemy=no
+                ability=did_shadow_veil
+            [/filter_distant]
         [/filter]
     [/hides]
 #enddef
@@ -189,7 +183,7 @@
                 [effect]
                     apply_to=new_ability
                     [abilities]
-                        {ABILITY_SHADOW_VEIL_HELPER $unit.side}
+                        {ABILITY_SHADOW_VEIL_HELPER}
                     [/abilities]
                 [/effect]
             [/object]
@@ -224,7 +218,7 @@
                 [effect]
                     apply_to=new_ability
                     [abilities]
-                        {ABILITY_SHADOW_VEIL_HELPER $this_unit.side}
+                        {ABILITY_SHADOW_VEIL_HELPER}
                     [/abilities]
                 [/effect]
             [/object]
--- a/data/schema/filters/unit.cfg
+++ b/data/schema/filters/unit.cfg
@@ -44,6 +44,7 @@
 	{FILTER_TAG "has_attack" weapon ()}
 	{FILTER_TAG "filter_vision" vision ()}
 	{FILTER_TAG "filter_adjacent" adjacent ()}
+	{FILTER_TAG "filter_distant" distant ()}
 	{FILTER_TAG "filter_location" location ()}
 	{FILTER_TAG "filter_side" side ()}
 	{FILTER_BOOLEAN_OPS unit}
@@ -55,5 +56,14 @@
 	super="$filter_unit"
 	{SIMPLE_KEY count s_range_list}
 	{SIMPLE_KEY adjacent dir_list}
+	{SIMPLE_KEY is_enemy s_bool}
+[/tag]
+
+[tag]
+	name="$filter_distant"
+	max=0
+	super="$filter_unit"
+	{SIMPLE_KEY radius s_int}
+	{SIMPLE_KEY count s_range_list}
 	{SIMPLE_KEY is_enemy s_bool}
 [/tag]
--- a/src/units/filter.cpp
+++ b/src/units/filter.cpp
@@ -160,6 +160,44 @@ struct unit_filter_adjacent : public unit_filter_base
 	const vconfig cfg_;
 };
 
+struct unit_filter_distant : public unit_filter_base
+{
+	unit_filter_distant(const vconfig& cfg)
+		: child_(cfg)
+		, cfg_(cfg)
+	{
+	}
+
+	virtual bool matches(const unit_filter_args& args) const override
+	{
+		const unit_map& units = args.context().get_disp_context().units();
+		std::vector<map_location> surrounding;
+		int radius = cfg_["radius"].to_int(0);
+		int match_count=0;
+
+
+		get_tiles_in_radius(args.loc, radius, surrounding);
+		for(unsigned j = 0; j < surrounding.size(); ++j){
+			unit_map::const_iterator unit_itor = units.find(surrounding[j]);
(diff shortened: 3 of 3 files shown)


Pentarctagon:
This seems like it could be done by adding `is_enemy` to the existing [filter_side] tag instead of creating something new?

newfrenchy83:
> This seems like it could be done by adding `is_enemy` to the existing [filter_side] tag instead of creating something new?

it write [filter_location][filter][filter_side], with that the side allied or enemy must be specified, and i not believe what is_enemy can be added there, i will test that but i not trust

CelticMinstrel:
I don't get the point of this? The diff seems to show that this is already possible.

newfrenchy83:
> I don't get the point of this? The diff seems to show that this is already possible.

Possible but imperfect, because camp changes are not taken into account, if we have a case where a unit with latent hides becomes an enemy of the teacher, its ability will continue to take the course of the latter as a reference, (proximity with the teacher will hide the unit when it should no longer be the case), or it would be necessary to remove the ability of the unit at the time of the change of camp and possibly reimplement it. This type of implementation is not compatible for abilities calling events conditioned by the presence of a unit at a certain distance from the filtered unit, because applicable only for heroes and not for general units which can be in several different camps.
It is to facilitate the implementation of this type of ability but also of events that I propose [filter_distant], which is moreover less cumbersome to write.

newfrenchy83:

CelticMinstrel:
I still don't understand the point of this. A cursory look at the code doesn't reveal any glaring problems though/

newfrenchy83:
> I still don't understand the point of this. A cursory look at the code doesn't reveal any glaring problems though.

I know that filter_location can be used for remote unit detection, but when it comes to specifying the alliances or not of these units, it is necessary to specify which camp they are allies or enemies.
When I changed the side of an invisible unit in debug mode to make it an enemy of Mal-Keshar, it remained invisible because the condition was that Mal-Keshar had to be allied with the side the unit belonged to when the ability gave it been implemented, which had not changed.
This is what prompted me to create a [filter_distant] that works like [filter_adjacent] but for remote units, allowing the same change to result in the unit becoming visible again.
In Did, such a case never arises, but for developers of add-ons that integrate abilities based on the use of [harm_unit] to damage units adjacent to the ability user's enemy unit in combat, it becomes complicated to want to extend the concept to distant units because the specification of the reference camp prevents the implementation of such an ability in a generalist unit, hence the idea of ​​[filter_distant] to simplify their task.

newfrenchy83:
@CelticMinstrel @Pentarctagon could you review last commit of this PR please?

newfrenchy83:
> @CelticMinstrel @Pentarctagon could you review last commit of this PR please?

Forget that, the wml_trst failed, usingnew feature more interrsting what modify code for filter_location

Pentarctagon:
My previous comments still stand I believe, so barring anything else being said I think I'd lean towards closing this PR rather than merging it.

Pentarctagon:
To clarify:
* What specific use case is this blocking right now? Is there an add-on that this would benefit?
* Can this instead be added to [filter_side]? Naively, I would think that checking whether a side is an enemy or not would be something that could do.

newfrenchy83:
> To clarify:
> 
>     * What specific use case is this blocking right now? Is there an add-on that this would benefit?
> 
>     * Can this instead be added to [filter_side]? Naively, I would think that checking whether a side is an enemy or not would be something that could do.

the 'hose' ability in Legends of the Invincible (description= _ "If this attack us used, many enemies behind and around the original target are hit. It is so tiring that the unit's damage decreases by 12 after the attack, the damage penalty will half each turn.") could be see his wml code simplified.

I cannot add this to filter_side because this filter is here for verify what unit is allied or nemy or owned by side specified in filter_side, data of [filter] own [filter_location] no present in sub [filter] who own filter_side, i must use a filter_distant independant of filter_location specific to locate unit for that instead to modify a filter designed for terrain but usable for unit with a sub filter.

newfrenchy83:
Perhaps what this PR is an impasse, finally.

CelticMinstrel:
> * Naively, I would think that checking whether a side is an enemy or not would be something that could do.

Yes, that already exists in `[filter_side]`.

Pentarctagon:
> I cannot add this to filter_side because this filter is here for verify what unit is allied or nemy or owned by side specified in filter_side, data of [filter] own [filter_location] no present in sub [filter] who own filter_side, i must use a filter_distant independant of filter_location specific to locate unit for that instead to modify a filter designed for terrain but usable for unit with a sub filter.

According to https://wiki.wesnoth.org/StandardSideFilter:
> filter matches only if side number is contained in a comma-separated list. Note that as opposed to the usual inline side= key in wml action tags, which defaults to 1, **the side= key in a SSF defaults to all sides** (similarly, the id= key in SUFs defaults to all ids). This key also accepts ranges (i.e. "1,3-4,6")

So from the wiki it sounds like you could use [has_unit]'s is_enemy attribute in a [filter_side] with no sides specified?

CelticMinstrel:
I think that would result in no sides matching. The text on the linked page implies that `[has_unit]` only checks units that belong to the specific side being matched, so `is_enemy` would always be false.

If you need to check in a side filter whether one side is allied with another, you can use the `[enemy_of]` and `[allied_with]` tags, or the `[has_ally]` and `[has_enemy]` tags.

Pentarctagon:
My understanding is that the intention of this PR is essentially to be able to do those sorts of checks using the current unit's side rather than having to explicitly pass in a side number.

newfrenchy83:
> My understanding is that the intention of this PR is essentially to be able to do those sorts of checks using the current unit's side rather than having to explicitly pass in a side number, which those subtags still require.

Yes, it's that.

Pentarctagon:
On the wiki then, it says that in side filter formulas there's the `side_number` value. Would using that accomplish the same thing?

CelticMinstrel:
> My understanding is that the intention of this PR is essentially to be able to do those sorts of checks using the current unit's side rather than having to explicitly pass in a side number, which those subtags still require.

I don't understand what you're saying? A side filter is checked against either a specific side (for match) or all sides (for filter). The contents of all of those subtags are another side filter, which can contain an explicit side number but does not need to.

Pentarctagon:
In this case, the side filter is in an ability definition, which is tied to a specific unit. The goal is to have that specific unit's side be used in the filter, regardless of which side that unit is on.

Pentarctagon:
But yeah, at this point my question is whether this can be accomplished via formula already? So given this is a filter in a unit ability, can you do something like `side=$(self.side_number)`?

CelticMinstrel:
> In this case, the side filter is in an ability definition, which is tied to a specific unit. The goal is to have that specific unit's side be used in the filter, regardless of which side that unit is on.

I'm still not sure if I understand what you mean. Abilities and weapon specials do not natively support a `[filter_side]` according to the wiki, which means that if one is used it must be an element of one of the Standard Unit Filters present on the ability. That means matching units whose side matches the filter. Does not placing `[filter_side]` inside `[filter_self]` accomplish the goal of filtering on the owning unit's side?

Or if the goal is to only have it match units on allied sides, I think the following would accomplish that?

```
[filter_opponent]
  side=$other_unit.side
[/filter_opponent]
```

> So given this is a filter in a unit ability, can you do something like `side=$(self.side_number)`?

That won't do anything useful. `$()` creates a blank formula context that doesn't have access to any variables, so `self` evaluates to null. If you drop the `$()` though, that would probably work in the `formula` key and match any unit on the same side. That is to say, `formula="side = self.side_number"` probably does what you're thinking.

Pentarctagon:
The ability changed in this PR does have a (quite nested) [filter_side] in it.

From what you've said though, it sounds like it's already possible to do what newfrenchy wants to use this PR's [filter_distant] to accomplish.

CelticMinstrel:
I think it could be done using a `[filter_radius]`, probably.

Pentarctagon:
Is there any reason to keep this PR open?

PR #6443, opened 2022-01-13
Black Nightmare: add a translation hint for this unit's name

stevecotton:


Wedge009:
Huh, even in English that subtle reference would be lost on me as I've seen ghostly horses called nightmares in other games. But I agree with the change.

stevecotton:
Even though it doesn't change the translatable strings, it makes more sense to drop this in just after 1.16.2 releases rather than just before it.

doofus-01 (approved):
The comment/hint is not untrue, and if it helps the translators, it's good to have.

PR #6445, opened 2022-01-14
Support [foreach]array=this_item, nested loops using the default name

### LINKED ISSUE #6305: [foreach]array=this_item.something does not write back to the outer this_item
Low-priority because closing this as 'won't fix' seems a reasonable option. The Wiki has already been updated, logging an issue gives me a page to refer questions to.

The `[foreach]` loop makes the WML variables `this_item` and `i` locally scoped. If `readonly=no`, which is the default, then the code inside the loop can modify data through `this_item`.

The name of the `this_item` variable can be overridden, but in 1.14 it was possible to use nested loops with the default names, as in the code below. Since 1.15.10 and a0ee38a49ae0da7de6665f23a271a7a63b3c3ae7's change to data/lua/wml-flow.lua, that doesn't work, because the modified value is stored back to the current (inner) `this_item` before the inner `this_item` goes out of scope.
```
        # Store a unit to a variable
        {VARIABLE u.abilities[0].regenerate.value 4}
        {VARIABLE u.abilities[1].heals.value 4}

        [foreach]
            array=u.abilities
            [do]
                [foreach]
                    array=this_item.heals
                    [do]
                        [set_variable]
                            name=this_item.value
                            value=8
                        [/set_variable]
                    [/do]
                [/foreach]
            [/do]
        [/foreach]
```

While writing the test code, I found a separate issue with nesting using the default names: although `array=this_item.something` generally works for read-only data, using it with `array=this_item` triggers the sanity-check added in d895b1865978ef3144e650f57f3b520b271efae3. The sanity check will compare the outer `this_item.length` to the inner `this_item.length`.

Low priority because it seems there are only two add-ons that have a line matching `array=this_item.something`, and only one seems broken. The Wiki has already been updated to say this is to be avoided, which is the reverse of the old docs.
* SXCollection is the add-on that was broken, and will be fixed in that add-on.
* Era of Magic's doubleattack.cfg already uses `variable=this_item_attack` for its inner loop.

stevecotton:
Since a0ee38a49, the inner this_item was still in scope when the inner loop
writes changed data back to the outer loop's variable, which meant that
changes were silently ignored.

A lot of the changes in wml-flow.lua are just indentation because of the
extra block.

Extend the unit test for scoped variables, which includes fixing the mixed
tab-and-space indentation in that file.

Fixes #6305. I'm dithering about whether to say it's a bugfix and should go
in 1.16, or say that it's unimportant enough to only go in 1.17.

### DIFF
--- a/data/lua/wml-flow.lua
+++ b/data/lua/wml-flow.lua
@@ -179,36 +179,51 @@ function wml_actions.foreach(cfg)
 	local array = wml.array_variables[array_name]
 	if #array == 0 then return end -- empty and scalars unwanted
 	local item_name = cfg.variable or "this_item"
-	local this_item <close> = utils.scoped_var(item_name) -- if this_item is already set
 	local i_name = cfg.index_var or "i"
-	local i <close> = utils.scoped_var(i_name) -- if i is already set
-	local array_length = wml.variables[array_name .. ".length"]
 
-	for index, value in ipairs(array) do
-		-- Some protection against external modification
-		-- It's not perfect, though - it'd be nice if *any* change could be detected
-		if array_length ~= wml.variables[array_name .. ".length"] then
-			wml.error("WML array length changed during [foreach] iteration")
-		end
-		wml.variables[item_name] = value
-		-- set index variable
-		wml.variables[i_name] = index-1 -- here -1, because of WML array
-		-- perform actions
-		for do_child in wml.child_range(cfg, "do") do
-			local action = utils.handle_event_commands(do_child, "loop")
-			if action == "break" then
-				utils.set_exiting("none")
-				goto exit
-			elseif action == "continue" then
-				utils.set_exiting("none")
-				break
-			elseif action ~= "none" then
-				goto exit
+	-- Some protection against unsupported modification of the array while iterating. Changes to
+	-- via the WML variable named array_name will be ignored during the loop, and then overwritten
+	-- by the contents of "array" afterwards. The check_for_modifications is just trying to add an
+	-- error message instead of silently ignoring those changes.
+	local array_length = wml.variables[array_name .. ".length"]
+	local check_for_modifications = true
+	if array_name:find("^" .. item_name) then
+		-- Disable the check for WML such as [for]array=this_item or [for]array=this_item.abilities,
+		-- where the current item is shadowing the variable of the same name outside the loop.
+		check_for_modifications = false
+	end
+
+	do
+		local this_item <close> = utils.scoped_var(item_name)
+		local i <close> = utils.scoped_var(i_name)
+
+		for index, value in ipairs(array) do
+			wml.variables[item_name] = value
+			-- set index variable
+			wml.variables[i_name] = index-1 -- here -1, because of WML array
+			-- perform actions
+			for do_child in wml.child_range(cfg, "do") do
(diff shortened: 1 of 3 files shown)


wml_test_schedule
```diff
@@ -321,6 +321,7 @@
 0 for_end2_step2
 0 for_end-2
 0 for_end-2_step-2
+0 nested_foreach
```
stevecotton, line 324: I mixed up the names, will force-push to fix this once the local build has confirmed that I've got it right this time.

data/lua/wml-flow.lua
```diff
@@ -179,36 +179,51 @@ function wml_actions.foreach(cfg)
...
-				utils.set_exiting("none")
-				break
-			elseif action ~= "none" then
-				goto exit
+	do
+		local this_item <close> = utils.scoped_var(item_name)
+		local i <close> = utils.scoped_var(i_name)
+
+		for index, value in ipairs(array) do
+			do
```
CelticMinstrel, line 200: Why is there an extra scope here? It doesn't contain _any_ local variable, much less to-be-closed ones.

stevecotton: Thanks for catching that, fixed.

Pentarctagon:
@stevecotton is this ready to go into 1.17 at least?

cooljeanius (approved):

data/lua/wml-flow.lua
```diff
@@ -179,36 +179,50 @@ function wml_actions.foreach(cfg)
...
+
+	-- Some protection against unsupported modification of the array while iterating.
+	-- Changing the array will simply be ignored during the loop, and then overwritten
+	-- by the contents of "array" afterwards, this is simply a method of trying to add
+	-- an error message instead of silently ignoring those changes.
 	local array_length = wml.variables[array_name .. ".length"]
+	local check_for_modifications = true
+	if array_name:find("^" .. item_name) then
+		-- Disable the protection for WML such as array=this_item or array=this_item.abilities
+		check_for_modifications = false
```
CelticMinstrel, line 192: I don't quite understand this. How does skipping the array length test in this circumstance make a difference?

stevecotton: The protection is checking that the code in the loop isn't trying to modify the array via the name that exists outside the loop. However, if the name that exists outside the inner loop is `this_item` and it's shadowed by a scoped `this_item` inside the loop then the shadowing itself triggers a false positive for the protection code.

stevecotton:
This will need the tests to be moved to the new locations.

IIRC, the main reason for setting it to draft is that there were other things with higher priority - @cooljeanius, are you wanting to use this functionality, or just triaging old PRs?

cooljeanius:
> This will need the tests to be moved to the new locations.
> 
> IIRC, the main reason for setting it to draft is that there were other things with higher priority - @cooljeanius, are you wanting to use this functionality, or just triaging old PRs?

Just triaging PRs that have issues linked to them

stevecotton:
The test_scoped_vars changes have been moved to #7321, along with a clean up of those changes.

stevecotton:
> @stevecotton is this ready to go into 1.17 at least?

It's ready now.

CelticMinstrel (approved):

cooljeanius (approved):

PR #6177, opened 2021-10-07
SoF Proofread Pass and Campaign Prose edits

max-torch:
This is a proofread pass. But this also has edits that are more than just minor edits which try to improve the campaign text. `wmlxgettext` was used to review all the text.

PR #6449, opened 2022-01-16
Remove non-standard spacing (SoF's wyrms)

stevecotton:


knyghtmare (approved):
A lot of the mainline campaigns seem to be still undergoing wmllint changes. Otherwise these would have been fixed already.

stevecotton:
When we cherry-pick these, we need (at least part of) 6e6d57d3656c03ff2ab96e179b1910b1ac545ca7 too.

> A lot of the mainline campaigns seem to be still undergoing wmllint changes. Otherwise these would have been fixed already.

Thanks for the heads-up. This change was easy to do, so I can postpone this until that's done.

Wedge009:
Not sure how I missed back-porting 6e6d57d but that's done now. Thanks for spotting it.

Do we know when the wmllint changes will be done? Otherwise this could potentially be postponed indefinitely, it's such a small change after all.

Wedge009 (approved):

stevecotton:
I've duplicated @max-torch's work here, he's already done this and more (removing the unnecessary word "basically") in #6177.

PR #6452, opened 2022-01-18
1.16 - various small sprite updates

doofus-01:


doofus-01:
I guess this is mostly just "swamp lizard" animations.

Issue #6411, opened 2021-12-30
[World Conquest] [1.16.2][1.17.0] Problems stated and bugs encountered


knyghtmare:
Bugs encountered in 1.16.1+dev/1.17.0+dev

- [ ] 1. Destruction mechanic line 73 bug
I triggered this when I killed an enemy unit standing on an elven castle hex
- [ ] 2. Starter hero revision
Steam community discussions complaint
- [ ] 3. Flight staff bug
Forum Post
- [ ] 4. Starting map bad luck probe
Steam community discussions complaint (I dont think this is a bug but someone's plain bad luck)
- [ ] 5. Possible PUA bug
This needs to be probed. See reference 1.
- [ ] 6. Revision of epic trait
I dont think it's necessary but might be checked. See forum post (reference 1).

References:
1. https://forums.wesnoth.org/viewtopic.php?p=670449#p670449
2. https://steamcommunity.com/app/599390/discussions/0/3199243212170122767/

knyghtmare:
Update: I had time to review the "epic trait" and it's "always AMLA with 60 XP and raises max HP by 6."
for level 2 units, it's fine but can be a burden to level 1 units. However, players get plenty of XP to allocate to units so it's fine. Unless someone wants to offer counter argument, I will tick it off.

knyghtmare:
Update: Is there any way to trigger a specific artifact spawning? I have been trying to get winged scepter to pop up but it hasn't at all.

knyghtmare:
Update: Confirmed Bug on Winged Scepter. Forturnately, there is PR #6433 to review for this.

knyghtmare:
>  Consider if conversion of 1P World Conquest to Hybridisation should be pursued.

A very interesting case.

Technically...it is possible however we are met with a barrier of "So...how can we get an era into SP mode"?
Well, most campaigns have `era_id=default` (check save files) so there might be some way. Maybe some dirty hack. Just throwing comments on it for now.

soliton-:
What would an era mean for a campaign?

knyghtmare:
> What would an era mean for a campaign?

For an SP campaign from the campaign Menu? Nothing relevant, I think. `era_id="era_default"` is there for every SP save file.
The main request was allow WC1 to be played from the SP campaign menu (so that the player could use RNG options).
Starting the campaign from SP mod might cause an issue with the player's leader type and recruit list as it's not stated unlike in MP Create Game. This is what I meant by it.

[...]

PR #6454, opened 2022-01-18
[world conquest] Fix for HP glitch when picking up items/artifacts

knyghtmare:
Fixes 7th bug of #6411


Discontinuum:
Tbh, I don't understand how
`--rebuild unit, to reduce savefile size.
	unit:transform(unit.type)`
helps to reduce the savefile size. While it's causing the hp thing

knyghtmare:
> Tbh, I don't understand how `--rebuild unit, to reduce savefile size. unit:transform(unit.type)` helps to reduce the savefile size. While it's causing the hp thing

No clue. I have zero cpp knowledge.

stevecotton:
This was backported in cebcd653fb295951712d2e15f3772d33688eff6e, but with the possibility of OOS I'm wondering if we need to revert it :(

Issue #4344, opened 2019-09-14
Mousing over a village should show the owner side in the UI


jostephd:
When you mouse over a village, the caption under the minimap says "Enemy village" or "Allied village", but nowhere does the UI say which side owns the village; you're supposed to infer that from the flag's color and shape, but that's not always easy (see #2682). So this is a suggestion to add to the UI somewhere an identification of which side controls the village: their leader's name, if known (as in the Game Status dialog), or else their side number.

Wedge009:
Looks like relevant part is in terrain code:
https://github.com/wesnoth/wesnoth/blob/d52783a7eaf9e254b7f552c54228d007c7f5af8f/src/terrain/terrain.cpp#L183-L204
I'm not sure if team information is available there as it currently stands. There may also need to be consideration of side/team names for translations (as this has been a lingering issue in the past).

jostephd:
> I'm not sure if team information is available there as it currently stands.

Team information is _definitely_ available at the point that decides which of those strings to show, and in any case, so long as we know the x,y of the village we can use the singletons in resources.hpp to determine any info we need (controlling team, viewing team, village owner, fog, etc).

> There may also need to be consideration of side/team names for translations (as this has been a lingering issue in the past).

[side]team_name is the translatable team name. There is a member function in team.hpp that returns that value too. I don't see a problem here.

stevecotton:
I wonder if there's any UMC that defines custom values for `income_description`, `income_description_ally`, etc. The current version of #6458 would remove support for those by replacing the text with the owner's name; I'm wondering if deprecating the old feature is actually the right choice.

CelticMinstrel:
I think it's not great that there are four different income description fields, but I don't think there's any other way to make it work with localization. You can't just substitute the main income description into a string like `_"Allied $name"`, because there's no way to inflect that correctly for an arbitrary `$name` in languages that require adjective-noun agreement.

Pentarctagon:
Resolved by #7920

PR #6458, opened 2022-01-24
Show team village owner name when hovering

macabeus:
Solves https://github.com/wesnoth/wesnoth/issues/4344

Show on the side menu the village owner's name, as proposed on the above issue.

BUT I'm not sure if it's the best way to solve it because there is a little space on the side menu, so the description could be dropped frequently.

### DIFF
--- a/src/reports.cpp
+++ b/src/reports.cpp
@@ -1573,10 +1573,9 @@ REPORT_GENERATOR(terrain, rc)
 			str << map.get_terrain_info(terrain).income_description();
 		} else if (owner == viewing_side) {
 			str << map.get_terrain_info(terrain).income_description_own();
-		} else if (viewing_team.is_enemy(owner)) {
-			str << map.get_terrain_info(terrain).income_description_enemy();
 		} else {
-			str << map.get_terrain_info(terrain).income_description_ally();
+			const team &owner_team = rc.teams()[owner - 1];
+			str << VGETTEXT("$village_owner's village", {{ "village_owner", owner_team.team_name() }});
 		}
 
 		const std::string& underlying_desc = map.get_underlying_terrain_string(terrain);
--- a/src/terrain/terrain.hpp
+++ b/src/terrain/terrain.hpp
@@ -145,8 +145,6 @@ public:
 	//these descriptions are shown for the terrain in the mouse over
 	//depending on the owner or the village
 	const t_string& income_description() const { return income_description_; }
-	const t_string& income_description_ally() const { return income_description_ally_; }
-	const t_string& income_description_enemy() const { return income_description_enemy_; }
 	const t_string& income_description_own() const { return income_description_own_; }
 
 	const std::string& editor_group() const { return editor_group_; }
@@ -226,8 +224,6 @@ private:
 	int heals_;
 
 	t_string income_description_;
-	t_string income_description_ally_;
-	t_string income_description_enemy_;
 	t_string income_description_own_;
 
 	std::string editor_group_;
--- a/src/terrain/terrain.cpp
+++ b/src/terrain/terrain.cpp
@@ -56,8 +56,6 @@ terrain_type::terrain_type() :
 		min_light_(0),
 		heals_(0),
 		income_description_(),
-		income_description_ally_(),
-		income_description_enemy_(),
 		income_description_own_(),
 		editor_group_(),
 		village_(false),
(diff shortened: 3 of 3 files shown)


src/terrain/terrain.hpp
```diff
@@ -145,8 +145,6 @@ class terrain_type
 	//these descriptions are shown for the terrain in the mouse over
 	//depending on the owner or the village
 	const t_string& income_description() const { return income_description_; }
-	const t_string& income_description_ally() const { return income_description_ally_; }
-	const t_string& income_description_enemy() const { return income_description_enemy_; }
```
macabeus, line 149: At the first moment, I tried to create a method here, `income_description_other`.
But since it could not be a `const`, and `map.get_terrain_info(terrain)` returns a const, I needed to move this code completely to `reports.cpp`.

src/reports.cpp
```diff
@@ -1573,10 +1573,9 @@ REPORT_GENERATOR(terrain, rc)
 			str << map.get_terrain_info(terrain).income_description();
 		} else if (owner == viewing_side) {
 			str << map.get_terrain_info(terrain).income_description_own();
-		} else if (viewing_team.is_enemy(owner)) {
-			str << map.get_terrain_info(terrain).income_description_enemy();
 		} else {
-			str << map.get_terrain_info(terrain).income_description_ally();
+			const team &viewing_team = rc.teams()[owner - 1];
+			str << VGETTEXT("$village_owner's village", {{ "village_owner", viewing_team.team_name() }});
```
stevecotton, line 1578: "viewing team" is the wrong name for this. In a multiplayer game, the viewing_team is the side of the player whose computer this wesnoth-client is running on, thus which side's shroud and fog should be applied.

src/terrain/terrain.hpp
CelticMinstrel: This change eliminates two (maybe three) keys from the terrain WML, so it feels like a regression. I think the idea is good, but we probably need to work out a different presentation.

In particular, it appears we lose the ability for an income tile to be labelled as anything other than a village. Prior to this change, it seems like we'd have the ability to create a tile that shows as "Bank", "Allied Bank", "Enemy Bank", etc. After this change, such a tile would now be referred to as a village if I'm not mistaken.

CelticMinstrel:
I suppose the space between the map location and the terrain icons is insufficient to fit a team name, right? It does seem a bit small from the screenshot… but can't hurt to ask.

macabeus:
@CelticMinstrel 
> I suppose the space between the map location and the terrain icons is insufficient to fit a team name, right? It does seem a bit small from the screenshot… but can't hurt to ask.

I think it's small too and I don't think it's the correct place to put it.
We are never using this space, and only using it for this case would be strange imho.

> EDIT: Another possibility would be adding the owner's name to the tooltip, so you need to mouse over "Allied Village" or "Enemy Village" to see the actual owner.

I don't know if I understand your suggestion well, but I can't hover over "Allied Village"/"Enemy Village" since it'll focus on other terrain or lose the focus completely after moving the mouse.

src/terrain/terrain.hpp
macabeus: We still can translate this by `$village_owner's village`
https://github.com/wesnoth/wesnoth/blob/9e7459600f228433c8abd8eed229f78e1a2290ec/src/reports.cpp#L1578

I'm not sure how WML works on it, but if WML can change the same strings as translations do, we'll be able to update it there as well.

But yeah, it'll be a regression. It's a breaking change.

src/reports.cpp
macabeus: Fixed. Replaced it by `owner_team`

CelticMinstrel:
> I don't know if I understand your suggestion well, but I can't hover over "Allied Village"/"Enemy Village" since it'll focus on other terrain or lose the focus completely after moving the mouse.

Ah right, that's a problem… but I don't think the way you've done it here fits very well either… and putting it on a new line isn't really an option either…

I think the only way I'd be comfortable putting the owner's name on that line is if the parenthetical portion were removed. It _is_ redundant with the icons on the line above, although I'm not sure it's sufficiently redundant to justify removing it altogether.

I also think that interpolating the name into a phrase is not a good approach. Instead of "Team 2's village (Flat, Village)", I think I'd prefer to see something like "Allied village (Team 2; Flat, Village)". Of course, that doesn't fit in the available space as well…

Pentarctagon:
@Vultraz any thoughts on this?

Pentarctagon:
Closing given the lack of responses, and also the regression noted in the above open review comment.

PR #6459, opened 2022-01-24
 various small sprite updates

doofus-01:
Forward-port of #6452

* flip water serpent defense frames

* fix layering issue with lizard tail

* update sea-serpent attack icon

* NE anims for swamp lizard


Wedge009:
(Got confused by the labels: 1.16 work is already merged so this PR is the corresponding foward-port. Therefore, no 'port' label required.)

PR #6462, opened 2022-01-25
add tooltip to unit_attack dialogs

newfrenchy83:
in tooltip, the special and weapon_abilities are listed by stats (damage, attacks and chance to hit) and owner of each is specified.
replace https://github.com/wesnoth/wesnoth/pull/6399 for readibility purpose.

### DIFF
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -22,6 +22,7 @@
 #include "display_context.hpp"
 #include "font/text_formatting.hpp"
 #include "game_board.hpp"
+#include "gettext.hpp"
 #include "global.hpp"
 #include "lexical_cast.hpp"
 #include "log.hpp"
@@ -845,7 +846,7 @@ static void add_name(std::string& temp_string, bool active, const std::string na
 		if (!name.empty() && checking_name.count(name) == 0) {
 			checking_name.insert(name);
 			if (!temp_string.empty()) temp_string += ", ";
-			temp_string += name;
+			temp_string += font::span_color(font::BUTTON_COLOR, name);
 		}
 	}
 }
@@ -949,7 +950,7 @@ void attack_type::weapon_specials_impl_self(std::string& temp_string, unit_const
 		for (const config::any_child sp : self->abilities().all_children_range()){
 			bool tag_checked = (!checking_tags.empty()) ? (checking_tags.count(sp.key) != 0) : true;
 			const bool active = tag_checked && check_self_abilities_impl(self_attack, other_attack, sp.cfg, self, self_loc, whom, sp.key, leader_bool);
-			add_name(temp_string, active, font::span_color(font::BUTTON_COLOR, sp.cfg["name"].str()), checking_name);
+			add_name(temp_string, active, sp.cfg["name"].str(), checking_name);
 		}
 	}
 }
@@ -972,7 +973,7 @@ void attack_type::weapon_specials_impl_adj(std::string& temp_string, unit_const_
 				bool default_bool = (affect_adjacents == "affect_allies") ? true : false;
 				bool affect_allies = (!affect_adjacents.empty()) ? sp.cfg[affect_adjacents].to_bool(default_bool) : true;
 				const bool active = tag_checked && check_adj_abilities_impl(self_attack, other_attack, sp.cfg, self, *it, i, self_loc, whom, sp.key, leader_bool) && affect_allies;
-				add_name(temp_string, active, font::span_color(font::BUTTON_COLOR, sp.cfg["name"].str()), checking_name);
+				add_name(temp_string, active, sp.cfg["name"].str(), checking_name);
 			}
 		}
 	}
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -919,7 +919,7 @@ std::string attack_type::weapon_specials_value(const std::set<std::string> check
 	add_name_list(temp_string, weapon_abilities, checking_name, "Owned: ");
 
 	weapon_specials_impl_adj(temp_string, self_, shared_from_this(), other_attack_, self_loc_, AFFECT_SELF, checking_name, checking_tags, "affect_allies", true);
-	add_name_list(temp_string, weapon_abilities, checking_name, "Taught:: ");
+	add_name_list(temp_string, weapon_abilities, checking_name, "Taught: ");
 
 	weapon_specials_impl_adj(temp_string, self_, shared_from_this(), other_attack_, self_loc_, AFFECT_SELF, checking_name, checking_tags, "affect_enemies", true);
 	add_name_list(temp_string, weapon_abilities, checking_name, "Taught: (by an enemy): ");
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -919,10 +919,10 @@ std::string attack_type::weapon_specials_value(const std::set<std::string> check
 	add_name_list(temp_string, weapon_abilities, checking_name, "Owned: ");
 
 	weapon_specials_impl_adj(temp_string, self_, shared_from_this(), other_attack_, self_loc_, AFFECT_SELF, checking_name, checking_tags, "affect_allies", true);
-	add_name_list(temp_string, weapon_abilities, checking_name, "Teached: ");
+	add_name_list(temp_string, weapon_abilities, checking_name, "Taught:: ");
 
 	weapon_specials_impl_adj(temp_string, self_, shared_from_this(), other_attack_, self_loc_, AFFECT_SELF, checking_name, checking_tags, "affect_enemies", true);
-	add_name_list(temp_string, weapon_abilities, checking_name, "Teached (by an ennemy): ");
+	add_name_list(temp_string, weapon_abilities, checking_name, "Taught: (by an enemy): ");
 
 
 	if(other_attack_) {
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -879,14 +879,11 @@ std::string attack_type::weapon_specials(bool is_backstab) const
 	}
 	std::string temp_string;
 	std::set<std::string> checking_name;
-	if(self_){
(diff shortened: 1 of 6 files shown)


src/units/abilities.cpp
```diff
@@ -885,33 +879,118 @@ std::string attack_type::weapon_specials(bool only_active, bool is_backstab) con
...
+	} else if (!weapon_abilities.empty()){
+		res = weapon_abilities;
+	}
+	return res;
+}
+
+static void add_string(std::string& weapon_abilities, std::string& self_abilities, const std::string temp_abilities, std::set<std::string>& checking_name)
+{
+	weapon_abilities = "";
+	checking_name.clear();
```
stevecotton, line 902: What do weapon_abilities and checking_name have to do with adding a string? This seems to be part of the flow logic of the caller instead.

src/units/abilities.cpp
```diff
@@ -885,33 +879,118 @@ std::string attack_type::weapon_specials(bool only_active, bool is_backstab) con
...
+
+std::string attack_type::weapon_specials_value(const std::set<std::string> checking_tags) const
+{
+	//log_scope("weapon_specials_value");
+	std::string weapon_abilities, self_abilities, self_teacher_allies, self_teacher_enemies, opponent_abilities;
+	std::set<std::string> checking_name;
+	for (const config::any_child sp : specials_.all_children_range()) {
+		if((checking_tags.count(sp.key) != 0)){
+			const bool active = special_active(sp.cfg, AFFECT_SELF, sp.key);
+			add_name(weapon_abilities, active, sp.cfg["name"].str(), checking_name);
```
stevecotton, line 915: I can't follow the logic of this function - it seems that `weapon_abilities` is a temporary string that's used for assembling parts of other strings called `self_abilities` etc, and if that's what it is then I don't see why `weapon_abilities` is in scope for the whole function. I also can't see why it's an input parameter to the weapon_specials_*_self functions, as it seems to be an output from those.

However, this line uses it purely as an output, and IIUC the value set here would be discarded, were it not for the first weapon_specials_impl_self call taking it as an input, and then possibly concatenating other values to it before copying it to self_abilities. However, that happens in a conditional block that only happens if self_ is non-null (which might always be true, but then why is there a conditional block)?

stevecotton (changes requested):

src/units/abilities.cpp
```diff
@@ -885,33 +879,118 @@ std::string attack_type::weapon_specials(bool only_active, bool is_backstab) con
...
+			opponent_abilities += weapon_abilities;
+			add_string(weapon_abilities, self_abilities, opponent_abilities, checking_name);
+		}
+	}
+	weapon_abilities = self_abilities;
+	return weapon_abilities;
+}
+
+void attack_type::weapon_specials_impl_self(std::string& weapon_abilities, unit_const_ptr self, const_attack_ptr self_attack, const_attack_ptr other_attack, const map_location& self_loc, AFFECTS whom,
+                                     std::set<std::string>& checking_name, const std::set<std::string> checking_tags, bool leader_bool)
```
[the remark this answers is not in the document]
newfrenchy83: why, checking_tags is constant inside function, reference taht not pertinent because value not changed.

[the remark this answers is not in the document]
stevecotton: Because passing by value makes an unnecessary copy of it.

src/units/abilities.cpp
```diff
@@ -885,33 +879,118 @@ std::string attack_type::weapon_specials(bool only_active, bool is_backstab) con
...
+{
+	if(self){
```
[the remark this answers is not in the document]
newfrenchy83: excuse me for this condition, this is because i forget what this in attack dialog, self_ and other_ are ever true, but i had implement that by reflexe.

src/units/abilities.cpp
newfrenchy83: i must rename that clear_temp_string instead.

src/units/abilities.cpp
newfrenchy83: weapon_abilities used mike reference,in add_name, i cannot put other stinrg in input or i put the code itself. I will try ths option.

src/units/abilities.cpp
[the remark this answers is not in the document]
newfrenchy83: Ok

src/units/abilities.cpp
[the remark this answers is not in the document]
newfrenchy83: but i will let this condition, else i fear what building not pass checking

src/units/abilities.cpp
stevecotton: Irrespective of the names, what do these two variables have to do with this function; why are they involved in it? They're not inputs, and they don't really feel like outputs either - they both get changed, but their final value doesn't depend on any of the inputs to the function.

stevecotton: Reviewing ea7bcc0c452, I still don't understand why checking_name is involved in this function.

newfrenchy83: This is for improbable, but not impossible, cases where weapons from different sources (an ability carried by self, and another by teacher for example), would have the same name, if in the window, the name should only be mentioned only once, in the tooltip, it seems relevant to point out that the stats are influenced by two weapons possessed and taught, the example of a charge type ability can also illustrate this when two units fight with the same capacity of this type (active both in attack and defense), the player sees the attack stats of his unit influenced by his own ability but also that of his opponent. To allow this, checking_name must be cleared after each block (self, teacher or opponent) is written.

stevecotton: I think I understand, but feel it could be refactored to be simpler. If I understand the current implementation correctly:
* When `add_name_list` returns, `checking_name` is always empty. Although the `clear()` is in a conditional block, that condition will always be true except for the times that `checking_name` is already empty when `add_name_list` is entered.
* In `weapon_specials_value` the only times that `checking_name` could be non-empty on entry to a `weapon_specials_impl_*` function are the ones for the `"Self: "` and `"Opponent: "` blocks.
* There are more places that it could be non-empty in `weapon_specials`.
* In `weapon_specials_value` there's only one time that the value of `checking_name` returned from a `weapon_specials_impl_*` function is used, for the `"Opponent: "` block.

newfrenchy83: for the first point, conditionnal block not ever true, if weapon_specials_impl_* return a empty temp_string, and preffix not added to non-existent list.
. Correct "Self: " has weapon specials and abilities like weapon owned by unit in list, and opponent has all weapon special or abilities like weapon applied used by opponent(owned or teached) and applied to unit

. Correct weapon_specials apllied to windows and origin to each weapon\abilities) non specifed, a name must be addedone time only.

. Yes because opponent is the only case or two weapon_specials_impl_* are used.

stevecotton: I'm wondering if we can omit special handling for this edge-case. For the `"Self: "` case, I think it only happens if the unit has an ability called `x` and a weapon special also called `x` - in which case, showing `x, x` will look odd but possibly be easier for the player to understand than combining them into `x` when two badly-named things are active.

For the `"Other: "` string, please would you give an example of an ability that should be returned by `weapon_specials_impl_adj`? I'm wondering if those will already be in "Enemy's Teachers: ".

newfrenchy83: using "Self: " for abilities only, ok. 

For "Opponent: " a leadership charge teached to opponent but appied also to attacker must appear in in tooltip, and the "opponent: " string is here for said what damage *2by abilities used by opponent, same if teached.

newfrenchy83:
@stevecotton i removed only_active argument to attack_type::weapon_specials()

newfrenchy83:
@stevecotton PR rebased

newfrenchy83:
@stevecotton it is long time what PR not reviewed.

newfrenchy83:
@Vultraz it is a long time what this PR not reviewed, could you give an advice, see what this PR is a UI work.

Pentarctagon:
Can you post a screenshot of what it looks like with this PR's change applied?

newfrenchy83:
> Can you post a screenshot of what it looks like with this PR's change applied?

teached: (cantor encoded in Elvish Captain and leadership)
tooltip_teached

teached by an enemy(diversion ability)
tooltip_teachedenemy


used by opponent(formation ability)
tooltip_usedopponent

newfrenchy83:
@Pentarctagon  @Vultraz , i have a question, with abilities/specials showed in tooltip, show the weapon teached by enemy or used by opponent in attack prediction windows is useful, or must i  show in tooltip only for not give heavy the window itself, and show in window weapon owned or teached by allies?

Pentarctagon:
I'd say to keep it just in the tooltip for now. Given how long this PR has already been open already, let's not add more to it.

newfrenchy83:
> I'd say to keep it just in the tooltip for now. Given how long this PR has already been open already, let's not add more to it.

except what 'diversion' and formation both in window in this PR, must becommit a remove of that or don't commit anymore?

newfrenchy83:
@Pentarctagon i reformulate my question: 'diversion' and 'formation' name must appear in window(without mention of origin) or not? And if not, must commit that in this PR or i another PR after merging in master?

Pentarctagon:
I think it'd be best to see what vultraz says first.

Just as far as wording goes though, for correctness:
* "teached" should be changed to "taught" in the above screenshots.
* "enemy" is misspelled as "ennemy".
* "others aspect" should be "other aspects".

newfrenchy83:
@Pentarctagon excuse my lack of patience but i would post remording in PR, and i let what weapon owned or taught by allies in window.

Vultraz (approved):

stevecotton:
I apologise for how long it's taken me to review this, and thank @Vultraz for reviewing the UI.

My plan is to merge this evening, unless anyone has more to say. If I have any more ideas about possible refactoring, I'll create a new PR on top of this instead of blocking this PR.

PR #6463, opened 2022-01-26
Temporarily re-introduce the SPECIAL_NOTES macro

stevecotton:
While this will be deleted before 1.18, at the moment UMC authors
are still working on 1.16. While the 1.17 branch is being used for
new development in the engine, I think it's more useful to be able
to run 1.16 UMCs that test engine edge cases rather than force the
UMCs to be upgraded for 1.18's macros.

There's currently circa 1200 units using `{SPECIAL_NOTES}` in
Ageless Era, probably requiring manual checks to update them.
Another option would be to `#define SPECIAL_NOTES` in the UMC
itself, but that would likely also mean that the warning was
silenced when running on 1.16, and few of the 1200 units would
get fixed during the 1.16 cycle.

This reverts part of commit 61fa3627818c1a3fb5181a21fc651b67d17b133a.

### DIFF
--- a/data/core/macros/special-notes.cfg
+++ b/data/core/macros/special-notes.cfg
@@ -109,6 +109,12 @@ _"This unit has a defense cap on certain terrain types — it cannot achieve a h
 
 # Deprecated versions here
 
+#define SPECIAL_NOTES
+#deprecated 4 Use [special_note]note= tags instead to apply special notes, or use the {NOTE_*} macros (note the singular).
+    "
+
+"+_"Special Notes:"#enddef
+
 #define SPECIAL_NOTES_SPIRIT
 #deprecated 2 1.17 This note is now added automatically.
 {INTERNAL:SPECIAL_NOTES_SPIRIT}#enddef


stevecotton:
@CelticMinstrel. your call whether to merge or reject the PR. I've been doing this locally for testing, so thought a PR including the patch would be the best way to ask you.

stevecotton:
Thanks. I've put removing it again on the roadmap for 1.17.12 (mid January 2023).

CelticMinstrel:
When removing it, it'll probably be just deleting the file outright (though the strings therein will need to be moved to abilities.cfg).

stevecotton:
The removal was postponed further in #7209, and the macro will still exist in 1.18. The 1.17 roadmap links to this issue, which is why I'm writing the plans for 1.18 here.

The player-visible header changed from "Special Notes:" to "Special Notes (1.14-style, please update to the new list format)". Any UMC that hasn't upgraded yet already has a cosmetic bug - the help pages of units still using the {SPECIAL_NOTES} macro will include duplicate notes (assuming the expected usage of {SPECIAL_NOTES} as a heading in [unit_type]description=, which is followed by old-style notes). These are minor cosmetic bugs, which are expected to be removed as UMC gets updated.

PR #6464, opened 2022-01-27
Make the adhoc add-on download check more detailed.

### LINKED ISSUE #6383: Addons can't be updated from the multiplayer server menu
 - What version of the game are you running? Tried 1.16.0 and 1.16.1
 - Where did you download it from? flatpak and custom build from 1.16.0 tag
 - What OS are you running? Ubuntu 20

### Describe the bug
When someone is hosting a game which required some addon and you have an outdated version of this addon, the game proposes to update the addon (if you click OK). In 1.16 it does nothing. Only manual updating through Main Menu > Add-ons works.

### To Reproduce
1. Host a multiplayer game in one Wesnoth instance
2. Launch another one with an outdated version of the addon
3. Try to update the addon with this menu 
изображение

Not all addons require all users to have the same latest version to play. Addons for which this behaviour was observed: Era of Magic, The Great Steppe Era
### Expected behavior
The addon updates successfully and you can join the game

Pentarctagon:
Originally this simply checked for whether the add-on's directory existed to avoid downloading the same add-on multiple times when migrating from one major release to another. However this then causes add-ons to not get updated when joining a multiplayer game where to person joining has the add-on (so the directory exists) but needs an updated version of it.

Fixes #6383

### DIFF
--- a/src/addon/manager_ui.cpp
+++ b/src/addon/manager_ui.cpp
@@ -19,6 +19,7 @@
 #include "addon/client.hpp"
 #include "addon/info.hpp"
 #include "addon/manager.hpp"
+#include "config_cache.hpp"
 #include "filesystem.hpp"
 #include "formula/string_utils.hpp"
 #include "preferences/game.hpp"
@@ -38,6 +39,7 @@ static lg::log_domain log_filesystem("filesystem");
 static lg::log_domain log_addons_client("addons-client");
 
 #define ERR_CFG LOG_STREAM(err,   log_config)
+#define INFO_CFG LOG_STREAM(info, log_config)
 
 #define ERR_NET LOG_STREAM(err,   log_network)
 
@@ -276,11 +278,31 @@ bool ad_hoc_addon_fetch_session(const std::vector<std::string>& addon_ids)
 			addons_list::const_iterator it = addons.find(addon_id);
 			if(it != addons.end()) {
 				const addon_info& addon = it->second;
-				// don't redownload in case it was already downloaded for being another add-on's dependency
-				if(!filesystem::file_exists(filesystem::get_addons_dir()+"/"+addon_id)) {
-					addons_client::install_result res = client.install_addon_with_checks(addons, addon);
-					return_value = return_value && (res.outcome == addons_client::install_outcome::success);
+				const std::string addon_dir = filesystem::get_addons_dir()+"/"+addon_id;
+				const std::string info_cfg = addon_dir+"/_info.cfg";
+
+				// no _info.cfg, so either there's a _server.pbl or there's no version information available at all, so this add-on can be skipped
+				if(filesystem::file_exists(addon_dir) && !filesystem::file_exists(info_cfg)) {
+					INFO_CFG << "No _info.cfg exists for '" << addon_id << "', skipping update.\n";
+					continue;
 				}
+
+				// if _info.cfg exists, compare the local vs remote add-on versions to determine whether a download is needed
+				if(filesystem::file_exists(info_cfg)) {
+					game_config::config_cache& cache = game_config::config_cache::instance();
+					config info;
+					cache.get_config(info_cfg, info);
+					version_info installed_addon_version(info.child_or_empty("info")["version"]);
+
+					// if the installed version is outdated, download the most recent version from the add-ons server
+					if(installed_addon_version >= addon.current_version) {
+						continue;
+					}
+				}
+
+				// if the add-on exists locally and needs to be updated, or it doesn't exist and needs to be downloaded
+				addons_client::install_result res = client.install_addon_with_checks(addons, addon);
+				return_value = return_value && (res.outcome == addons_client::install_outcome::success);
 			} else {
 				if(!return_value) {
 					os << ", ";


Pentarctagon:
Tested locally:
* Host has add-on, joining player doesn't -> joining player prompted to download and then is able to join.
* Host has add-on, joining player has outdated version -> joining player prompted to download and then is able to join.
* Host has older version than joining player -> joining player is unable to join and is given a prompt suggesting they tell the host they need to update.

ProditorMagnus:
Reminder to test where one or more involved players are addon authors.

Pentarctagon:
It looks like that wouldn't be an issue. If the add-on versions aren't compatible and it is required, then it's not possible to join (a popup gets shown saying the host needs to update). If the add-on isn't required, then it wouldn't be checked for whether an update is needed in the first place.

src/addon/manager_ui.cpp
```diff
@@ -38,6 +39,7 @@ static lg::log_domain log_filesystem("filesystem");
 static lg::log_domain log_addons_client("addons-client");
 
 #define ERR_CFG LOG_STREAM(err,   log_config)
+#define INFO_CFG LOG_STREAM(info,   log_config)
```
stevecotton, line 42: Either make the `log_config` line up with the line above, or don't add the extra spaces.

src/addon/manager_ui.cpp
```diff
@@ -276,8 +278,29 @@ bool ad_hoc_addon_fetch_session(const std::vector<std::string>& addon_ids)
...
+				if(filesystem::file_exists(info_cfg)) {
+					game_config::config_cache& cache = game_config::config_cache::instance();
+					config info;
+					cache.get_config(info_cfg, info);
+					version_info installed_addon_version(info.child_or_empty("info")["version"]);
+
+					// if the installed version is outdated, download the most recent version from the add-ons server
+					if(installed_addon_version < addon.current_version) {
+						addons_client::install_result res = client.install_addon_with_checks(addons, addon);
+						return_value = return_value && (res.outcome == addons_client::install_outcome::success);
```
stevecotton, line 301: The contents of this conditional block are the same as the contents of the next conditional block, it would be nice to merge them.

The logic above has already used a `continue` to avoid an unnecessary download, so I think the logic here should use `continue` too.

src/addon/manager_ui.cpp
Pentarctagon: Done.

src/addon/manager_ui.cpp
Pentarctagon: Did the first part, though I'm not sure what you mean about the `continue`?

stevecotton: I mean that the logic could do this:
```
if(...)
    continue;
if(...)
    continue;
addons_client::install_stuff();
```
or it could do this:
```
bool do_update=true;
if(...)
    do_update=false;
if(...)
    do_update=false;
if(do_update==false)
    addons_client::install_stuff();
```
but please choose one or the other without mixing them.

Pentarctagon: The point of it is simply to short circuit things when there is no chance of doing either an install or an update. I could add a `do_nothing` variable and then change the if-checks that come after it to also check that variable, but that would then also mean the last part of the current logic - where it displays any add-ons which needed an install/update but weren't found on the add-ons server - would never trigger anymore.

stevecotton: Maybe I'm missing something - I'm not seeing why you do this:
```
// if the installed version is outdated, download the most recent version from the add-ons server
if(installed_addon_version < addon.current_version) {
    do_update = true;
}
```
instead of doing
```
// if the installed version is already up-to-date, skip downloading a new copy from the add-ons server
if(installed_addon_version >= addon.current_version) {
    continue;
}
```

Pentarctagon: Removing `do_update` and using `continue` would mean that the check immediately below it would be just:
```
if(!filesystem::file_exists(info_cfg)) {
	addons_client::install_result res = client.install_addon_with_checks(addons, addon);
	return_value = return_value && (res.outcome == addons_client::install_outcome::success);
}
```
which then wouldn't do an update for an outdated but existing add-on. Unless you mean getting rid of that if-check altogether?

stevecotton: AFAICS, that if-check can be removed.

Pentarctagon: Alright, updated.

src/addon/manager_ui.cpp
```diff
@@ -276,11 +278,31 @@ bool ad_hoc_addon_fetch_session(const std::vector<std::string>& addon_ids)
...
+
+				// if _info.cfg exists, compare the local vs remote add-on versions to determine whether a download is needed
+				if(filesystem::file_exists(info_cfg)) {
+					game_config::config_cache& cache = game_config::config_cache::instance();
+					config info;
+					cache.get_config(info_cfg, info);
+					version_info installed_addon_version(info.child_or_empty("info")["version"]);
+
+					// if the installed version is outdated, download the most recent version from the add-ons server
+					if(installed_addon_version < addon.current_version) {
```
stevecotton, line 298: The logic has reversed, this should be a `>=` test now.

Pentarctagon: Updated.

Issue #6466, opened 2022-01-27
Cumulative leadership doesn't work


Discontinuum:
### Game and System Information
 - What version of the game are you running? 1/16.1
 - Where did you download it from? flatpak

### Describe the bug
Cumulative leadership doesn't stack with other leaderships it should stack with (with different ids)

### To Reproduce
One of the easiest ways: get War Banner in World Conquest and place a unit near any leader and the banner-bearer. The banner provides cumulative leadership but it doesn't work

### Expected behavior
Units get increased leadership bonus

### Savefiles
[WC2-WC II 2 - Start Turn 1Cumulative.gz](https://github.com/wesnoth/wesnoth/files/7953278/WC2-WC.II.2.-.Start.Turn.1Cumulative.gz) A save of WC with War Banner and Elvish Marshal to play with

### Additional context
In 1.14 it worked

newfrenchy83:
In 1.16, cumulative mustn't be used in [leadership], instead you must replace value by 'add' or sub' for cumule value. i will try to renabled that without remove 'add' or 'sub' but i don't think what it was accepted in 1.16 or master.

Discontinuum:
Should we mention it in the docs, or has it been already mentioned?

newfrenchy83:
> Should we mention it in the docs, or has it been already mentioned?

it hasn't already mentionned, but i pull 2 PR for fix this issue, and it must wait what this was merged or not inmaster and 1.16 branch.

Discontinuum:
I mentioned it
> and it must wait what this was merged or not inmaster and 1.16 branch
In Discord @Pentarctagon said to merge the API fix only in 1.17, if I'm not mistaken

newfrenchy83:
Then mention in docs for 1.16 branch, what value and cumulative=yes must be replaced by add or sub and cumulative=no

Pentarctagon:
A change to the API can't be implemented in 1.16 since it would break compatibility with previous 1.16 releases in MP as well as replays.

Discontinuum:
>  by add or sub and cumulative=no
why cumulative=no should be in place?

newfrenchy83:
> > by add or sub and cumulative=no
> 
> why cumulative=no should be in place? I don't get it

if cumulative=yes with 'value' and value is negative, then final value will be 0. Keep cumulative=no for don't make error. No problem with add or sub but prudence

Discontinuum:
[...]

PR #6468, opened 2022-01-28
1.17 fix cumulative=yes don't function for [leadership]

newfrenchy83:
Fix https://github.com/wesnoth/wesnoth/issues/6466 issue.

### DIFF
--- a/data/test/scenarios/test_add_or_sub_in_leadership.cfg
+++ b/data/test/scenarios/test_add_or_sub_in_leadership.cfg
@@ -1,4 +1,5 @@
 # This unit test is used for verify what 'add' and 'sub' attributes work in [leadership] abilities
+# but also what 'cumulative=yes' is fixed.
 {GENERIC_UNIT_TEST "test_add_in_leadership_abilities" (
     [event]
         name=start
@@ -15,16 +16,28 @@
                 apply_to=new_ability
                 [abilities]
                     [leadership]
+                        id=leadership_no_cum
+                        value=25
+                        cumulative=no
+                        affect_self=yes
+                    [/leadership]
+                    [leadership]
                         id=leadership_pos
                         add=25
                         cumulative=no
+                        affect_self=yes
+                    [/leadership]
+                    [leadership]
+                        id=leadership_cum
+                        value=25
+                        cumulative=yes
                         affect_self=yes
                     [/leadership]
                     [attacks]
                         value=1
                     [/attacks]
                     [damage]
-                        value=60
+                        value=12
                     [/damage]
                     [chance_to_hit]
                         value=100
@@ -73,7 +86,7 @@
             [/filter]
             variable=b
         [/store_unit]
-        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 25})}
+        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 79})}
         {SUCCEED}
     [/event]
 )}
--- a/data/test/scenarios/test_add_or_sub_in_leadership.cfg
+++ b/data/test/scenarios/test_add_or_sub_in_leadership.cfg
@@ -16,9 +16,15 @@
                 apply_to=new_ability
                 [abilities]
                     [leadership]
-                        id=leadership_no_cum
+                        id=leadership_cum
                         value=25
-                        cumulative=no
+                        cumulative=yes
+                        affect_self=yes
+                    [/leadership]
+                    [leadership]#this value must be added to other leadership_cum
+                        id=leadership_cum
+                        value=25
+                        cumulative=yes
                         affect_self=yes
                     [/leadership]
                     [leadership]
@@ -27,10 +33,10 @@
                         cumulative=no
                         affect_self=yes
                     [/leadership]
-                    [leadership]
-                        id=leadership_cum
-                        value=25
-                        cumulative=yes
+                    [leadership]#this value musn't be added because already leadership_pos to value=25
+                        id=leadership_pos
+                        add=15
+                        cumulative=no
                         affect_self=yes
                     [/leadership]
                     [attacks]
@@ -86,6 +92,8 @@
             [/filter]
             variable=b
         [/store_unit]
+        #the final damage must be damage_value +(damage_value*(2*leadership_cum + highest leadership_pos))
+        # 12 + 12*(2*0.25 + 0.25) = 21. It must be remain 100hp-21=79hp
         {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 79})}
         {SUCCEED}
     [/event]
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -221,13 +221,13 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	if(drains) {
 		unit_ability_list drain_specials = weapon->get_specials_and_abilities("drains");
 		// Compute the drain percent (with 50% as the base for backward compatibility)
-		unit_abilities::effect drain_percent_effects(drain_specials, 50, backstab_pos, weapon, true);
+		unit_abilities::effect drain_percent_effects(drain_specials, 50, backstab_pos, weapon);
 		drain_percent = drain_percent_effects.get_composite_value();
 	}
 
 	// Add heal_on_hit (the drain constant)
 	unit_ability_list heal_on_hit_specials = weapon->get_specials_and_abilities("heal_on_hit");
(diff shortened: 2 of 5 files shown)


knyghtmare:
image
Seems to work. With your PR, banner + leadership is stacking now 
20% + 25% = 45%

newfrenchy83:
@Pentarctagon when you're have time to review this PR you're see what fixing of cumulative=yes in [leadership] don't disable using od add and sub in same time.

Pentarctagon:
This is definitely something that would benefit from having a unit test added to make sure this issue doesn't get reintroduced accidentally later.

newfrenchy83:
> This is definitely something that would benefit from having a unit test added to make sure this issue doesn't get reintroduced accidentally later.

test case added

src/actions/attack.cpp
```diff
@@ -1585,7 +1585,7 @@ void attack_unit_and_advance(const map_location& attacker,
 int under_leadership(const unit &u, const map_location& loc, const_attack_ptr weapon, const_attack_ptr opp_weapon)
 {
 	unit_ability_list abil = u.get_abilities_weapons("leadership", loc, weapon, opp_weapon);
-	unit_abilities::effect leader_effect(abil, 0, false);
+	unit_abilities::effect leader_effect(abil, 0, false, const_attack_ptr() , true);
```
stevecotton, line 1588: Why should is_cumulable be true for leadership, but false for any other ability?

newfrenchy83: cumulative don't work to same for leadership and other abilities like chance_to_hit or cumulative=yes using a refence base_value like valueif ref>value, then with leadership ref ever 0 but cumulative=yes is used for add value of cumulative leadership to value to other leadership, is_cumulable vriable used for disctinct leadership abilities to other in effect

stevecotton:
The build failure seems trivial - the first test timed out, but all the other passed, and they all passed on the other build machines. That happens occasionally, I assume because that test has to fill up the OS' file cache with all the data files.

I'll review as if it all passed.

newfrenchy83:
> The build failure seems trivial - the first test timed out, but all the other passed, and they all passed on the other build machines. That happens occasionally, I assume because that test has to fill up the OS' file cache with all the data files.
> 
> I'll review as if it all passed.

Could i expect PR merged before next update of 1.17?

stevecotton:
> Could i expect PR merged before next update of 1.17?

Probably. Sorry for not finding time so far.

newfrenchy83:
> > Could i expect PR merged before next update of 1.17?
> 
> Probably. Sorry for not finding time so far.
> 
> Edit: wondering if there should be more tests, but that leads into general thoughts on whether there should be a better framework for "alice and bob fight" tests to reduce the amount of code that needs to go in the individual tests.

amount of code reduced.

newfrenchy83:
@stevecotton wesnoth 1.17.2 updated. I begin to desesperate.

newfrenchy83:
@stevecotton don't forget to review this fix PR

newfrenchy83:
@stevecotton comment added in abilities.cpp

newfrenchy83:
could i expect what this PR will be commited for 1.17.4?

newfrenchy83:
@Pentarctagon or @CelticMinstrel , someone could review this PR, please, because i begin to believe what @stevecotton has forget this time.

stevecotton:
Sorry for the lack of feedback. I keep wondering "does it work with things in this order, does it work with things in that order, how". I confirm that it works with any arrangement of `cumulative=yes`/`cumulative=no` on a 4-leader test (the `COMMON_KEEP_A_B_C_D` setup added in #6627).

src/actions/attack.cpp
```diff
@@ -1586,7 +1586,7 @@ void attack_unit_and_advance(const map_location& attacker,
...
```
stevecotton, line 1589: I think all the lines in this PR should be consistent about using the same choice of `const_attack_ptr()`, `{}` or `nullptr` for the first optional argument.

src/units/abilities.cpp
```diff
@@ -1706,24 +1706,41 @@ effect::effect(const unit_ability_list& list, int def, bool backstab, const_atta
...
+		//while what is_cumulable or "cumulative" attribute are flase, "value"
+		//calculation type is usual, else , "value" is equivalent to "add" or "sub" atribute.
+		if(!is_cumulable || !cfg["cumulative"].to_bool()){
+			if (const config::attribute_value *v = cfg.get("value")) {
+				int value = get_single_ability_value(*v, def, ability, list.loc(), att, [&](const wfl::formula& formula, wfl::map_formula_callable& callable) {
+					callable.add("base_value", wfl::variant(def));
+					return formula.evaluate(callable).as_int();
+				});
+
+				int value_cum = cfg["cumulative"].to_bool() ? std::max(def, value) : value;
```
stevecotton, line 1720: This line tests the value of `cfg["cumulative"]`, but it's inside a conditional block that already checks `(!is_cumulable || !cfg["cumulative"].to_bool())`.

Under what circumstances could we have both `!is_cumulable` and yet also `cfg["cumulative"].to_bool() == true`? It shoulds like a tag that shouldn't support `cumulative=yes` having different behavior if given the value `cumulative=yes`.

stevecotton:
Given that we now have `add` and `sub`, is there still a use-case for `cumulative` that wouldn't be fulfilled better by `add` or `sub`?

src/units/abilities.cpp
newfrenchy83: [chance_to_hit] is a case where is_cumulable is false but cumulative can be true, and cumulative =no when no writen in wml code, and i don't want change this.

newfrenchy83:
> Given that we now have `add` and `sub`, is there still a use-case for `cumulative` that wouldn't be fulfilled better by `add` or `sub`?

for me, the issue is what in chance_to_hit, cumulative option wasn't an cumulative(choice of higher value, but not add two value of weapon specials), and for what after 1.18 release, i won't anymore to issue for the bug of cumulative=yes because the addon develloper don't know what must be using add or sub instead.

to was more persnonal, this is me what, in tring to change the calculationof value in [leadership] for integrate the negatives 'value' who is to origin of bug and i try to propose a real fixing, not to said of wml devellopper, "using 'add' or sub' for fix that yourselve because we don't want fix the issue what we create ourselve".

And if PR merged, cumulative option will be real utility in some weapon/abilities, where cumulative is real cumulative option.

newfrenchy83:
yes value with cumulative=yes is not more efficient what add or sub, but the devellopper who want using a cumulative leadership don't know ever what add or sub must be used, and i don't really want have reort issue in github or forum, i sense myself a few guilty.

src/actions/attack.cpp
newfrenchy83: it's done

newfrenchy83:
@stevecotton i fixed bug and pulled your test in PR

newfrenchy83:
@stevecotton testing fixed.

newfrenchy83:
@Pentarctagon like i pull stevecotton testing in PR, could you review this, please?

data/test/scenarios/test_teachers_and_specials.cfg
```diff
@@ -0,0 +1,91 @@
...
+
+{COMMON_KEEP_A_B_C_D_UNIT_TEST cumulative_leadership_same_id (
+    [event]
+        name=start
+
+        {TEST_GIVE_LEADERSHIP id=bob B 50 yes}
+        {TEST_GIVE_LEADERSHIP id=charlie B 50 yes}
+        {TEST_GIVE_LEADERSHIP id=dave B 50 yes}
+    [/event]
+)}
```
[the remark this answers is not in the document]
newfrenchy83: @stevecotton i removed your test

data/test/scenarios/test_add_or_sub_in_leadership.cfg
```diff
@@ -73,7 +86,7 @@
             [/filter]
             variable=b
         [/store_unit]
-        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 25})}
+        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 79})}
```
Pentarctagon, line 89: To confirm then, the final damage taken by the defender here is 21, which is calculated by:
* 5 damage (Elvish Archer's base damage)
* +12 (from [damage])
* +1.25 damage (from leadership_no_cum)
* +1.25 damage (from leadership_pos)
* +1.25 damage (from leadership_cum)
This adds up to 20.75, which rounds to 21. Correct?

And this is expected, since even though leadership_no_cum has `cumulative=no`, it's still cumulative with other [leadership] abilities that don't share that ID?

newfrenchy83: @Pentarctagon test modified

PR #6469, opened 2022-01-28
fix cumulative=yes don't function for [leadership](1.16)

newfrenchy83:
Fix https://github.com/wesnoth/wesnoth/issues/6466 issue.


stevecotton:
Note for anyone looking at this afterwards - as this is an API change, we decided to merge the equivalent PR to master, but not to merge the 1.16 version.

PR #6470, opened 2022-01-29
Replace make enum

Pentarctagon:
Enums:
- [x] enumname - make_enum unit test, irrelevant
- [x] another - make_enum unit test, irrelevant
- [x] ALIGNMENT_FEMALE_VARIATION - t_string
- [x] UNIT_ALIGNMENT - t_string
- [x] RANDOM_FACTION_MODE - t_string
- [x] CAMPAIGN_TYPE - string
- [x] component_availability - string
- [x] CONDITIONAL_TYPE - string
- [x] DEFEAT_CONDITION - string
- [x] EPOCH - string
- [x] grow_direction - string
- [x] hit_type - string
- [x] LEVEL_RESULT - string
- [x] PROXY_CONTROLLER - string
- [x] SAVED_GAME_MODE - string
- [x] SHARE_VISION - string
- [x] SORT_ORDER - string
- [x] STATUS - string
- [x] create_engine/TYPE - string
- [x] ai/TYPE - string
- [x] VARIANT_TYPE - string

The plan for MAKE_ENUMs that use t_strings is to implement that part the same as was already done for the [side_controller enum](https://github.com/wesnoth/wesnoth/pull/6423/files#diff-64452baca10a211dae88365da8ebc8870a3fd724cf4067030f7163ae13b938b3) (aka a static const std::array containing t_strings), though I'm not sure what the difference is between using `N_` vs `_`.

---

Testing:
- [x] ALIGNMENT_FEMALE_VARIATION - t_string
- [x] UNIT_ALIGNMENT - t_string
- [x] RANDOM_FACTION_MODE - t_string
- [x] CAMPAIGN_TYPE - string
- [x] component_availability - string
- [x] CONDITIONAL_TYPE - string
- [x] DEFEAT_CONDITION - string
- [x] EPOCH - string
- [x] grow_direction - string
- [x] hit_type - string
- [x] LEVEL_RESULT - string
- [x] PROXY_CONTROLLER - string
- [x] SAVED_GAME_MODE - string
- [x] SHARE_VISION - string
- [x] SORT_ORDER - string
- [x] STATUS - string
- [x] create_engine/TYPE - string
- [x] ai/TYPE - string
- [x] VARIANT_TYPE - string

---

Remaining:
- [x] Should the `string_enum`s all be in their own directory or kept in the directories they each are most used in?
- [x] Rename `attack_result` to `strike_result`.
- [x] Rename the formula variant values to remove the `type_` prefix.


CelticMinstrel:
> though I'm not sure what the difference is between using `N_` vs `_`.

`N_` marks it as translatable for the gettext tool, but without performing a lookup. So, if you use `N_`, you need to separately make a dynamic gettext call later in order to convert it to a `t_string`.

If it's in a header, especially if it's static, you should use `N_`.

Pentarctagon:
This is now fully implemented, I just need to now go through and test various things.

src/string_enums/ai_target.hpp
```diff
@@ -0,0 +1,34 @@
+/*
+	Copyright (C) 2008 - 2021
```
stevecotton, line 2: 2021

Pentarctagon: Since this is adding a bunch of new files, I figured I'd run the copyright update script on the repo after merging this.

Pentarctagon:
Should be good now.

projectfiles/Xcode/The Battle for Wesnoth.xcodeproj/project.pbxproj
```diff
@@ -4240,8 +4234,6 @@
 				46DF5BCC1F46173700BE6D24 /* irdya_datetime.cpp */,
 				46DF5BCB1F46173700BE6D24 /* irdya_datetime.hpp */,
 				911F471B1CAE5A7E00F47035 /* iterable_pair.hpp */,
-				EC53B04D1B23BB0E002F758F /* make_enum.cpp */,
-				91B621931B76720B00B00E0F /* make_enum.hpp */,
```
stevecotton, line 4244: Every IDE file that had `make_enum.hpp` removed should probably have `enum_base.hpp` added.

CelticMinstrel: Missing a header won't break compilation, at least.

src/ai/composite/goal.cpp
```diff
@@ -138,7 +138,7 @@ void target_unit_goal::add_targets(std::back_insert_iterator< std::vector< targe
 	for (const unit &u : resources::gameboard->units()) {
 		if (ufilt( u )) {
 			LOG_AI_GOAL << "found explicit target unit at ... " << u.get_location() << " with value: " << value() << "\n";
-			*target_list = target(u.get_location(), value(), target::TYPE::EXPLICIT);
+			*target_list = target(u.get_location(), value(), ai_target::type::XPLICIT);
```
CelticMinstrel, line 141: …why'd you drop the E here?

src/game_initialization/create_engine.cpp
```diff
@@ -28,6 +28,7 @@
 #include "saved_game.hpp"
 #include "wml_exception.hpp"
 #include "string_enums/side_controller.hpp"
+#include "string_enums/component_availability.hpp"
```
stevecotton, line 31: Alphabetically, should be one line up

src/ai/composite/goal.cpp
```diff
@@ -313,17 +313,13 @@ void lua_goal::add_targets(std::back_insert_iterator< std::vector< target >> tar
...
-		std::vector < target > targets = *(l_obj->get());
 
-		for (target tg : targets)
-		{
-			*target_list = tg;
-		}
-	} catch(const bad_enum_cast& e) {
-		ERR_AI_GOAL << "A Lua goal returned a target of an unknown type (\"" << e.value() << "\"; unfortunately, the engine cannot recover from this error. As a result, all targets returned by the goal have been lost.\n";
-	}
+	std::vector < target > targets = *(l_obj->get());
```
CelticMinstrel, line 317: Is there any way to get the error message back? Obviously the exact details are incorrect (even before, it looks like it managed to partially recover), but I think at least noting when there's an invalid target type is a good idea (and maybe don't insert the invalid type into the list at all).

src/editor/map/map_context.cpp
```diff
@@ -539,7 +539,7 @@ config map_context::to_config()
 
 		side["fog"] = t->uses_fog();
 		side["shroud"] = t->uses_shroud();
-		side["share_vision"] = t->share_vision();
+		side["share_vision"] = team_shared_vision::get_string(t->share_vision());
```
CelticMinstrel, line 542: So just theoretically, what happens if you didn't make this change? Will it still compile but produce the wrong answer?

src/game_classification.hpp
```diff
@@ -72,21 +66,21 @@ class game_classification
...
 	bool is_test() const
 	{
-		return campaign_type == CAMPAIGN_TYPE::TEST;
+		return type == campaign_type::type::TEST;
 	}
 
 	bool is_tutorial() const
 	{
-		return campaign_type == CAMPAIGN_TYPE::TUTORIAL;
+		return type == campaign_type::type::TUTORIAL;
```
CelticMinstrel, line 84: Side note: this type doesn't even exist anymore, right?

src/string_enums/enum_base.hpp
```diff
@@ -55,6 +55,20 @@ struct enum_base : public T
 		return std::nullopt;
 	}
 
+	/**
```
stevecotton, line 58: It was added in a previous PR, but I'm wondering if there should be a bounds-check in get_string() to handle `get_string(ENUM_MAX)`, which I think would be undefined behavior accessing beyond the end of `T::values`.

src/reports.cpp
```diff
@@ -394,7 +395,7 @@ REPORT_GENERATOR(unit_alignment, rc)
 	const map_location& hex = mouseover_hex.valid() ? mouseover_hex : displayed_unit_hex;
 	return unit_alignment(rc, u, hex);
 }
-REPORT_GENERATOR(selected_unit_alignment, rc)
+REPORT_GENERATOR(selected_unit, rc)
```
CelticMinstrel, line 398: ???

This doesn't look right? Why would you change the report name?

src/scripting/lua_unit.cpp
```diff
@@ -461,7 +461,13 @@ static int impl_unit_set(lua_State *L)
 	modify_vector_string_attrib("extra_recruit", u.set_recruits(value));
 	modify_vector_string_attrib("advances_to", u.set_advances_to(value));
 	if(strcmp(m, "alignment") == 0) {
-		u.set_alignment(lua_check<UNIT_ALIGNMENT>(L, 3));
+		std::string str = lua_check<std::string>(L, 3);
```
CelticMinstrel, line 464: It'd certainly be nice for `lua_check` to have an enum overload that does all this for you.

src/string_enums/defeat_condition.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <array>
+
+#include "string_enums/enum_base.hpp"
+
+struct defeat_condition_defines
+{
+	static constexpr const char* const no_leader_left = "no_leader_left";
+	static constexpr const char* const no_units_left = "no_units_left";
+	static constexpr const char* const never = "never";
+	static constexpr const char* const always = "always";
```
CelticMinstrel, line 26: …what the heck is "defeat condition always"?

src/ai/composite/goal.cpp
stevecotton: `explicit` is a C++ keyword, so the lowercase naming wouldn't work.

CelticMinstrel:
I don't see any major issues with this, but one thing I don't get is why all the string enums are collected into a single directory rather than being placed in the same directory (or even the same file) as the code that uses it (for example, `formula_type` is no longer defined in the `formula` directory). It's not like these are gigantic files, either - most are under 10 lines, I think? (That's not counting the gigantic copyright boilerhead, of course.)

src/game_classification.hpp
stevecotton: It shouldn't be in anything new, but I guess old savegames might have it in.

src/string_enums/ai_target.hpp
```diff
@@ -0,0 +1,34 @@
...
+	static constexpr const char* const village = "village";
+	static constexpr const char* const leader = "leader";
+	static constexpr const char* const xplicit = "explicit";
+	static constexpr const char* const threat = "threat";
+	static constexpr const char* const battle_aid = "battle aid";
+	static constexpr const char* const mass = "mass";
+	static constexpr const char* const support = "support";
+
+	enum class type { VILLAGE, LEADER, XPLICIT, THREAT, BATTLE_AID, MASS, SUPPORT, ENUM_MAX };
+	static constexpr std::array<const char*, static_cast<int>(type::ENUM_MAX)> values{ village, leader, xplicit, threat, battle_aid, mass, support };
```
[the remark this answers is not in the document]
stevecotton: It would be nicer to write, but I think it would also bring back the Boost macro library that MAKE_ENUM used.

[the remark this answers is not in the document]
CelticMinstrel: That doesn't rely on Boost.Preprocessor.

src/reports.cpp
stevecotton: As @CelticMinstrel says, this looks like a bug. These macros turn into ThemeWML tag names, so this makes it expect a `[selected_unit]` instead of a `[selected_unit_alignment]` in the theme.

I think anything that starts `[selected_` isn't exercised by the built-in themes, but can be used by jostephd's two-sidebar theme (and probably others).

CelticMinstrel: Seems likely this was just an overzealous search-and-replace error.

projectfiles/Xcode/The Battle for Wesnoth.xcodeproj/project.pbxproj
Pentarctagon: For xcode in particular I'm also not sure what the hash (I'm assuming it's a hash) represents, so I don't actually know how to manually add those.

src/ai/composite/goal.cpp
Pentarctagon: I wanted to keep the lowercase and uppercase names the same.

src/ai/composite/goal.cpp
Pentarctagon: I removed it since I didn't see what would have resulted in a bad_enum_cast here. `target` already has an enum member variable, so it didn't look like there was a place where it was converting from a string.

src/editor/map/map_context.cpp
Pentarctagon: It fails to compile because `enum class` isn't implicitly convertible to an int, and config doesn't have a separate assignment operator to use for this case.

src/string_enums/ai_target.hpp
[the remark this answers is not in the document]
stevecotton: Ah, for a one-line to one-line conversion, I don't think it's worth macroizing. I was wondering about something like MAKE_ENUM, where a single macro invocation would generate the whole boilerplate, including the lower-case strings and their upper-case mapping.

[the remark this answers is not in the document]
Pentarctagon: It doesn't really save a lot of typing.

Pentarctagon:
It seemed unnecessary to include sometimes a fairly large header file just to be able to use a particular string enum.

src/reports.cpp
Pentarctagon: Ah, yeah, I was actually un-renaming another report after noticing changing *that* report generator name caused the unit alignment to no longer be displayed on the sidebar, and somehow this one got caught up in that.

src/string_enums/enum_base.hpp
Pentarctagon: Should it just return a default like:
```
if(key == T::type::ENUM_MAX) {
	return "ENUM_MAX";
}
```
or something else?

CelticMinstrel:
Sometimes that does make sense, but I think there are cases where you'd have to include that other header anyway.

And that doesn't explain why they're all gathered into the `string_enums` folder rather than being included amongst the files that use them most.

src/string_enums/ai_target.hpp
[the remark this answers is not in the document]
CelticMinstrel: It's easier typing though.

src/scripting/lua_unit.cpp
CelticMinstrel: I think it wouldn't be hard to support something like `lua_check<enum_definition_list>(L, idx)`?

src/string_enums/enum_base.hpp
CelticMinstrel: `ENUM_MAX` is an invalid value, so I'd go with either an empty string or nullopt.

src/editor/map/map_context.cpp
CelticMinstrel: So basically it relies on the enum being scoped.

I guess that's fine as long as we note that the enum _must_ be scoped (I wonder if that's enforcible via `static_assert`?)

src/ai/composite/goal.cpp
CelticMinstrel: It definitely exists somewhere, because the Lua API returns the target type as a string (or an integer, which I think is deprecated).

projectfiles/Xcode/The Battle for Wesnoth.xcodeproj/project.pbxproj
CelticMinstrel: Yeah, headers should be added, but it's not a compile error if they're missing. (I think it might break the Counterparts popup to switch between associated source and header files.)

Pentarctagon:
Given C++ makes it generally a nightmare to know all of what's included unless you specifically go out of your way to look for it, IMO it's more trouble than it's worth to try and distinguish between the two cases. Also to some degree it ended up this way since that's how the first MAKE_ENUM replacement was done, and it was done that way because there was a MAKE_ENUM that was needed in both the client and wesnothd, which resulted in the same identical MAKE_ENUM declaration being copied into two separate headers.

CelticMinstrel:
I don't really agree, but whatever.

Separately, I still think they should at least be in the same directory as code they're associated with - the AI ones in the AI directory, the WFL ones in the formula directory, etc.

Pentarctagon:
My preference would be to have them grouped with each other so it's straight forward where to find any particular implementation if you're looking for it. So @stevecotton can be the tie breaker I suppose.

src/ai/composite/goal.cpp
Pentarctagon: I'm assuming the actual reading of the value happens at the point in lua_object.hpp: https://github.com/wesnoth/wesnoth/pull/6470/commits/f0b90963c4cf6cde8d21fd7f6101156056b1de84#diff-cf0d01f4533748ed6e5ea6b39b78a6c4a4d3cd7513a9b806094c6d527dc14ac1 so I can change the bad value handling and/or add logging there.

src/scripting/lua_unit.cpp
Pentarctagon: What would `enum_definition_list` be here?

src/string_enums/enum_base.hpp
Pentarctagon: It doesn't return an optional currently, so I guess an empty string would be simpler.

src/editor/map/map_context.cpp
Pentarctagon: https://stackoverflow.com/questions/15586163/c11-type-trait-to-differentiate-between-enum-class-and-regular-enum it looks like it would be something like: 
`static_assert(std::integral_constant<bool, std::is_enum<typename T::type>::value && !std::is_convertible<typename T::type, int>::value>::value, "Enum is not a scoped enum");`
which does correctly result in a compile error if an instance of it uses a regular enum.

src/gui/core/placer.cpp
```diff
@@ -16,26 +16,26 @@
 #define GETTEXT_DOMAIN "wesnoth-lib"
 
 #include "gui/core/placer.hpp"
-
-#include <cassert>
 #include "gui/core/placer/horizontal_list.hpp"
 #include "gui/core/placer/vertical_list.hpp"
+#include "string_enums/grow_direction.hpp"
```
stevecotton, line 21: This file has to be included by placer.hpp anyway, no need for it here.

While it's trivial, I'm currently looking at the dependency graphs while tiebreaking where the files should go.

stevecotton:
Although enum_base is in a namespace, none of the enums are. I think that's going to tie into the question of where the files should be, so please give me time to think it over.

src/string_enums/attack_result.hpp
```diff
@@ -0,0 +1,31 @@
...
+{
+	static constexpr const char* const hit = "hit";
+	static constexpr const char* const miss = "miss";
+	static constexpr const char* const kill = "kill";
+	static constexpr const char* const invalid = "invalid";
+
+	enum class type { HIT, MISS, KILL, INVALID, ENUM_MAX };
+	static constexpr std::array<const char*, static_cast<int>(type::ENUM_MAX)> values{ hit, miss, kill, invalid };
+};
+using attack_result = string_enums::enum_base<attack_result_defines>;
```
stevecotton, line 31: While `attack_result` is more descriptive than `hit_type`, there's already a class with the same name in src/ai/actions.hpp. The other one gets forward-declared in src/ai/game_info.hpp,

src/game_initialization/create_engine.cpp
Pentarctagon: Done.

src/ai/composite/goal.cpp
Pentarctagon: So instead of using `value_or()`, it should check if it converted correctly and if not then skip it with `continue`?

src/string_enums/enum_base.hpp
Pentarctagon: Done.

src/reports.cpp
Pentarctagon: Fixed.

src/gui/core/placer.cpp
Pentarctagon: Fixed.

src/string_enums/attack_result.hpp
Pentarctagon: I can change it to `attack_results` unless there another name you would prefer?

src/scripting/lua_unit.cpp
Pentarctagon:
```
template<typename T>
std::enable_if_t<std::is_base_of_v<enum_base, T>, std::optional<typename T::type>>
lua_check(lua_State *L, int n)
{
	std::string str = lua_check_impl::lua_check<std::string>(L, n);
	std::optional<typename T::type> val = T::get_enum(str);
	if(!val) {
		luaL_argerror(L, n, ("cannot convert " + str + " to enum.").c_str());
	}
	return val;
}
```
is the closest I've been able to get, but this doesn't work with just
`lua_check<unit_alignments>(L, 3).value_or(unit_alignments::type::NEUTRAL)`
since that wants to return an instance of `unit_alignments`. What does work is:
`lua_check_impl::lua_check<unit_alignments>(L, 3).value_or(unit_alignments::type::NEUTRAL)`

CelticMinstrel: Huh? Doesn't `lua_check<unit_alignments>(L, 3)` work with that definition?

src/ai/composite/goal.cpp
CelticMinstrel: Something like that seems reasonable, yeah.

src/scripting/lua_unit.cpp
Pentarctagon: It gives the error: `cannot convert ‘std::decay_t<string_enums::enum_base<unit_alignments_defines> >’ {aka ‘string_enums::enum_base<unit_alignments_defines>’} to ‘unit_alignments_defines::type’`

src/ai/composite/goal.cpp
Pentarctagon: Alright, should be good now.

src/scripting/lua_unit.cpp
CelticMinstrel: Okay yeah, that's because of the `lua_check` wrapper in the global namespace that then calls the `lua_check_impl` version…

You could add another overload of the wrapper too, I guess… (or instead)

src/string_enums/attack_result.hpp
CelticMinstrel: The AI one is the result of an attack, but this one is the result of a single blow within an attack. I'm not sure if that helps in naming it though…

The only other possible name I can think of right now is `attack_hit_type`.

src/scripting/lua_unit.cpp
Pentarctagon: So like:
```
template<typename T>
std::optional<typename T::type> lua_enum_check(lua_State *L, int n)
{
	//remove possible const& to make life easier for the impl namespace.
	return lua_check_impl::lua_check<T>(L, n);
}
```
?

src/string_enums/enum_base.hpp
Pentarctagon: Another option also being to use `at(int)`, which would throw instead of resulting in UB.

src/scripting/lua_unit.cpp
CelticMinstrel: Um… well, sure, I guess that works.

CelticMinstrel: Actually, I wonder if it would make sense to use `luaL_checkoption` here since that's specifically intended for enums? It won't return an `std::optional` though, but that's probably preferred since `check` functions are supposed to raise a Lua error if the argument is invalid.

In other words, you can delete the implementation function and just write this:

```lua
template<typename T>
typename T::type lua_enum_check(lua_State *L, int n)
{
	return T::get_enum(luaL_checkoption(L, n, nullptr, T::values));
}
```

That seems correct at least; I haven't tested it. It won't return at all if the value is invalid. It does rely on the enum constants being contiguous (ie, no explicit constant assignments), but I think non-contiguous enums would already be a problem in the system, right?

CelticMinstrel:
Just linking to this comment because it seems like it's easy to miss under all GitHub's arcane space-preserving magic.

https://github.com/wesnoth/wesnoth/pull/6470#discussion_r808658148

src/scripting/lua_unit.cpp
Pentarctagon: Right.

CelticMinstrel: Ah, I just noticed that [the documentation](https://www.lua.org/manual/5.4/manual.html#luaL_checkoption) states the _array_ must be null-terminated, so some kind of trickery might be needed to guarantee that's the case.

Pentarctagon: Though it looks like `luaL_checkoption`'s final argument is `const char* const*` rather than `const std::array<const char*, 4>`.

CelticMinstrel: Well yes, but the latter is trivially convertible to the former (`values.data()`). The only tricky part is the null-terminator.

I'm not sure if consecutive static variables are guaranteed to be contiguous in memory, so my personal choice for solving this would probably be a wrapper type that inserts an extra "ghost" element, something like this:

```cpp
template<size_t n>
class enum_array : public std::array<const char*, n> {
  using std::array<const char*, n>::array;
  const char* null_terminator = nullptr;
};
```

This would mean that when `luaL_checkoption` tries to read off the end of the array, it would find the `null_terminator` there and stop. (This is admittedly assuming that there's no padding; but I don't think padding would be inserted between pointers anyway.) It's less specialized than built-in `std::array` because this trick only really works with pointers.

Maybe there are other solutions. I'm not sure.

One thing I personally wouldn't do is to just make the array one longer and manually add a nullptr to the end, because that would mess up your `get_string` that's assuming all the pointers are valid. But you could still make it work as long as you account for this.

Pentarctagon: I think I'd prefer to leave it as it is currently to be honest, given it's only used in once place. If in the future it's used in more places and ends up needing a different implementation, then I'd rather address that then.

CelticMinstrel: At the very least I think it makes sense for it to return `T::type` instead of `std::optional<T::type>`.

Pentarctagon: It would need to have a default value returned if the conversion fails, in that case.

CelticMinstrel: No, because it _doesn't return at all_ if the conversion fails (it raises a Lua error instead).

Pentarctagon: Ah, my bad. Updated.

Pentarctagon:
An idea that occurred to me for this, is that instead of using ENUM_MAX on the enum to determine its size, use the count of arguments in `__VA_ARGS__` to determine its size.  Meaning, define:
```
#define ENUM_AND_ARRAY(...) \
	static constexpr int count = std::tuple_size<decltype(std::make_tuple(__VA_ARGS__))>::value; \
	enum class type { __VA_ARGS__ }; \
	static constexpr std::array<const char*, count> values{ __VA_ARGS__ };
```
(first line found from [this answer](https://stackoverflow.com/a/33349105)).

Which would then be used like:
```
struct ai_target_defines
{
	static constexpr const char* const village = "village";
	static constexpr const char* const leader = "leader";
	static constexpr const char* const xplicit = "explicit";
	static constexpr const char* const threat = "threat";
	static constexpr const char* const battle_aid = "battle aid";
	static constexpr const char* const mass = "mass";
	static constexpr const char* const support = "support";

	ENUM_AND_ARRAY(village, leader, xplicit, threat, battle_aid, mass, support)
};
using ai_target = string_enums::enum_base<ai_target_defines>;
```

The downside being you end up with the constants being the same case as the array, but it needs less typing to define the string_enum and also removes the need for ENUM_MAX (and its special handling).

CelticMinstrel:
It may also be possible to get the count from the array somehow.

Something like this:

```cpp
static constexpr std::array<const char*> values = {A, B, C};
```

I'm not sure if that works though? And even if it does I'm not sure if it's C++17.

I don't mind your macro idea too much, but having the two be the same case is a little inconvenient, as I think it means `enum_name::enum_constant` won't work – you'll need to write `enum_name::type::enum_constant`\.

Pentarctagon:
Trying to declare an std::array without providing the length doesn't work for me locally at least.

stevecotton:
C++17 allows omitting both template arguments from std::array, but doesn't allow omitting just the size: https://en.cppreference.com/w/cpp/container/array/deduction_guides

Pentarctagon:
If no one's opposed then, I'll go with:
```
#define ENUM_AND_ARRAY(...) \
	enum class type { __VA_ARGS__ }; \
	static constexpr std::array values{ __VA_ARGS__ };
```

Pentarctagon:
Also @CelticMinstrel @stevecotton - are there any other remaining issues or parts that you need to review? If not I'm going to squash most of these commits into a single one.

CelticMinstrel:
The enums are still all collected in a single directory. I'd prefer them to be in the same directory as the code they're mostly used with. I don't remember anyone else commenting about this.

stevecotton:
There's two enums that are used included from many files, and having them in their own files seems right to me.

Most of the others are included from exactly one .hpp file, and I'm intending to look at where they're used - I suspect many would either be better moved into that .hpp, or the places that include that .hpp should include just the enum directly.

CelticMinstrel:
Okay, but when an enum _is_ in its own file, I think it should be with the code that uses it most. For example, the formula type enum should be in the formula/ folder.

CelticMinstrel:
(No idea how that comment was interpreted as closing this, sorry about that.)

Pentarctagon:
side_controller in particular is an odd case to categorize like that, since it's used both in the game client as well as wesnothd. But also, if some are in their own file and some aren't, then where would the standalone files go? It wouldn't make sense to have a string_enums directory just for those, so I guess they'd have to just go in src/ along with enum_base.hpp?

src/string_enums/formula_variant.hpp
```diff
@@ -28,7 +28,6 @@ struct formula_variant_defines
 	static constexpr const char* const type_string = "string";
 	static constexpr const char* const type_map = "map";
 
-	enum class type { TYPE_NULL, TYPE_INT, TYPE_DECIMAL, TYPE_OBJECT, TYPE_LIST, TYPE_STRING, TYPE_MAP, ENUM_MAX };
-	static constexpr std::array<const char*, static_cast<int>(type::ENUM_MAX)> values{ type_null, type_int, type_decimal, type_object, type_list, type_string, type_map };
+	ENUM_AND_ARRAY(type_null, type_int, type_decimal, type_object, type_list, type_string, type_map)
```
CelticMinstrel, line 31: You know, I don't think there's any reason these all need to start with `type_`, is there? I imagine that was a relic of it once being an unscoped enum?

src/string_enums/enum_base.hpp
```diff
@@ -82,7 +83,8 @@ struct enum_base : public T
 };
 
 #define ENUM_AND_ARRAY(...) \
+	static constexpr std::size_t count = std::tuple_size<decltype(std::make_tuple(__VA_ARGS__))>::value; \
```
CelticMinstrel, line 86: So I guess the implicit template arguments didn't work?

src/string_enums/formula_variant.hpp
Pentarctagon: I also can't have a variable named just `int`, etc.

src/string_enums/enum_base.hpp
Pentarctagon: Yeah, for the mingw crosscompile job, it seemed to confuse the linker when it wasn't provided explicitly for some reason.

src/string_enums/formula_variant.hpp
CelticMinstrel: That's not an issue, because "int" is short for "integer" in the first place, so you can just de-shorten it.

Pentarctagon: I'd prefer for them to either all have the prefix or for none of them to have the prefix just for consistency.

CelticMinstrel: Yeah, that goes without saying. I don't _expect_ `null` to be a problem, but I wouldn't put it past some obscure header on some obscure OS (or even on Windows) to define `null` as a macro for a null pointer. If that _did_ happen though, we could just call it something else, like `nil` or `nothing` or something like that.

Pentarctagon: Oh, whoops, I was getting it mixed up with Java where it's just `null` instead of `nullptr`. So yeah, that (hopefully) wouldn't be a problem.

src/scripting/lua_team.cpp
```diff
@@ -82,15 +82,15 @@ static int impl_side_get(lua_State *L)
 	return_string_attrib("color", t.color());
 	return_string_attrib("controller", side_controller::get_string(t.controller()));
 	return_bool_attrib("is_local", t.is_local());
-	return_string_attrib("defeat_condition", t.defeat_condition().to_string());
-	return_string_attrib("share_vision", t.share_vision().to_string());
+	return_string_attrib("defeat_condition", defeat_condition::get_string(t.defeat_cond()));
+	return_string_attrib("share_vision", team_shared_vision::get_string(t.share_vision()));
```
stevecotton, line 86: Appears to be a duplicate of line 93.

stevecotton:
A note about the enums, organised by which files include each header. Feel free to delete this comment if it's too long, I've kept a local copy.

Allows existing code to include the new file instead of a larger .hpp
===

I think this is the only one that lets an `#include` be replaced with a smaller
one: `src/game_launcher.hpp` only needs `level_result.hpp` and not the rest of
`game_end_exceptions.hpp`.

```
src/game_end_exceptions.hpp:#include "string_enums/level_result.hpp"
```

Could be used by preferences, but currently isn't
===

```
src/game_initialization/create_engine.hpp:#include "string_enums/level_type.hpp"
src/mp_game_settings.hpp:#include "string_enums/random_faction_mode.hpp"
```


Used in src/gameboard, playcontroller and team - should those become a component?
===

```
src/team.hpp:#include "string_enums/side_proxy_controller.hpp"
```

Only used in one component
===

Only used in AI
---

```
src/ai/default/contexts.hpp:#include "string_enums/ai_target.hpp"
```

Only used in animation?
===

This one's hard to check, there's a class with the same name.

```
src/units/animation.hpp:#include "string_enums/attack_result.hpp"
```

Only used in scripting
---

```
src/scripting/plugins/manager.hpp:#include "string_enums/plugin_manager_status.hpp"
```

Only used in `game_initialization`
---

```
src/game_initialization/create_engine.cpp:#include "string_enums/component_availability.hpp"
src/game_initialization/depcheck.cpp:#include "string_enums/component_availability.hpp"
```

GUI2 cleanup candidate (issue #6540)
---

Explained in #6540.

```
src/gui/core/placer.hpp:#include "string_enums/grow_direction.hpp"
```

Variant
---

With this, my feeling is `variant.hpp` and `variant_value.hpp` ought to be reorganised to have one public file and
one header that's only used inside the implementation code for variants. The enum itself feels like an implementation
detail.

```
src/formula/variant_value.hpp:#include "string_enums/formula_variant.hpp"
```

Used in several components, but only included via a single .hpp, and only with other stuff from that .hpp
===

Used in many places:

```
src/game_classification.hpp:#include "string_enums/campaign_type.hpp"
src/mp_game_settings.hpp:#include "string_enums/saved_game_mode.hpp"
src/team.hpp:#include "string_enums/defeat_condition.hpp"
```

This one needs a special note - although `src/gui/dialogs/editor/edit_side.hpp` seems as if its `#include`s could changed from team.hpp to `team_shared_vision.hpp`, `edit_side.hpp` also depends on the `editor_team_info` class, putting a hard indirect dependency on team.hpp.

```
src/team.hpp:#include "string_enums/team_shared_vision.hpp"
```

Used in few places:

```
src/utils/irdya_datetime.hpp:#include "string_enums/wesnoth_epoch.hpp"
```

Maybe should be used in more places
===

Currently specific to one component each, but thes are general concepts that are likely to already have parallel implementions (without using `MAKE_ENUM`) in other components.

```
src/units/filter.hpp:#include "string_enums/conditional_type.hpp"
src/preferences/general.hpp:#include "string_enums/sort_order.hpp"
```

Used in many places and many includes
===

```
src/editor/map/map_context.hpp:#include "string_enums/side_controller.hpp"
src/game_board.hpp:#include "string_enums/side_controller.hpp"
src/game_initialization/connect_engine.cpp:#include "string_enums/side_controller.hpp"
src/game_initialization/create_engine.cpp:#include "string_enums/side_controller.hpp"
src/game_state.cpp:#include "string_enums/side_controller.hpp"
src/gui/dialogs/editor/edit_side.hpp:#include "string_enums/side_controller.hpp"
src/gui/dialogs/multiplayer/mp_join_game.cpp:#include "string_enums/side_controller.hpp"
src/saved_game.cpp:#include "string_enums/side_controller.hpp"
src/server/wesnothd/game.hpp:#include "string_enums/side_controller.hpp"
src/team.hpp:#include "string_enums/side_controller.hpp"
```

```
src/actions/attack.hpp:#include "string_enums/unit_alignments.hpp"
src/ai/contexts.cpp:#include "string_enums/unit_alignments.hpp"
src/reports.cpp:#include "string_enums/unit_alignments.hpp"
src/terrain/filter.cpp:#include "string_enums/unit_alignments.hpp"
src/tod_manager.cpp:#include "string_enums/unit_alignments.hpp"
src/units/types.hpp:#include "string_enums/unit_alignments.hpp"
src/units/unit.hpp:#include "string_enums/unit_alignments.hpp"
```

Pentarctagon:
For **Allows existing code to include the new file instead of a larger .hpp**, another example would be `src/gui/dialogs/multiplayer/mp_join_game.cpp` for `side_controller`. Technically it didn't *replace* an existing #include of team.hpp, but only because it was previously using hard-coded strings rather than getting the string values from the MAKE_ENUM like it should have been.

src/string_enums/attack_result.hpp
stevecotton: Suggestions: `swing_result` or `swing_type`. The animation code is already using `int swing_num` in this context.

CelticMinstrel:
I would say `unit_alignments.hpp` belongs in the units folder (although it'll go away if I get to my custom alignments proposal) and `side_controller.hpp` should be placed in the same folder as `team.hpp` (although that's unfortunately just src/ at the moment).

src/string_enums/attack_result.hpp
Pentarctagon: Maybe `strike_result`? I guess `swing_result` would tend to imply certain weapons (you don't swing things like a bow or a lance), but honestly I also don't really have a strong opinion any particular way.

Pentarctagon:
Alright, so:
* enum_base - `src/`
* ai_target - `src/ai/`
* campaign_type - `src/`
* component_availability - `src/game_initialization`
* conditional_type - `src/units/`
* defeat_condition - `src/`
* formula_variant - `src/formula/`
* grow_direction - `src/gui/`
* level_result - `src/`
* level_type - `src/game_initialization/`
* plugin_manager_status - `src/scripting/`
* random_faction_mode - `src/game_initialization/`
* saved_game_mode - `src/game_initialization/`
* side_controller - `src/`
* side_proxy_controller - `src/`
* sort_order - `src/gui/`
* strike_result - `src/units/`
* team_shared_vision - `src/`
* unit_alignments - `src/units/`
* wesnoth_epoch - `src/utils/`

src/scripting/lua_team.cpp
Pentarctagon: Removed.

CelticMinstrel:
> Alright, so:

I'm a little bit confused about the purpose of this list. The moves you describe here don't seem to have been done? Are you announcing your intention to move them to those locations and asking for feedback before you do it? (If that's the case, they seem mostly fine to me.)

Pentarctagon:
> Are you announcing your intention to move them to those locations and asking for feedback before you do it?

I'm posting what I intend to do so I don't need to move them again later.

Pentarctagon:
And done. Barring anything else coming up, this is ready to merge then.

stevecotton:
Please put these in namespaces.

Documentation would be nice-to-have, but given the assortment of stuff in this PR I think it would be more reasonable to merge without and then leave it to other developers to document stuff that's in their area. I feel we can't ask for it from other developers unless we at least have a plan for adding it to our own new stuff; I'll take a look later today and assign some tasks to myself for the documentation update.

Pentarctagon:
Should they be in the `string_enums` namespace, or other namespaces?

Pentarctagon:
To clarify: do you mean they should all be in already existing namespaces?

Also, should it be like:
```
namespace blah
{

struct ai_target_defines
{
	static constexpr const char* const village = "village";
	static constexpr const char* const leader = "leader";
	static constexpr const char* const xplicit = "explicit";
	static constexpr const char* const threat = "threat";
	static constexpr const char* const battle_aid = "battle aid";
	static constexpr const char* const mass = "mass";
	static constexpr const char* const support = "support";

	ENUM_AND_ARRAY(village, leader, xplicit, threat, battle_aid, mass, support)
};

}
using ai_target = string_enums::enum_base<blah::ai_target_defines>;
```
or
```
namespace blah
{

struct ai_target_defines
{
	static constexpr const char* const village = "village";
	static constexpr const char* const leader = "leader";
	static constexpr const char* const xplicit = "explicit";
	static constexpr const char* const threat = "threat";
	static constexpr const char* const battle_aid = "battle aid";
	static constexpr const char* const mass = "mass";
	static constexpr const char* const support = "support";

	ENUM_AND_ARRAY(village, leader, xplicit, threat, battle_aid, mass, support)
};

using ai_target = string_enums::enum_base<blah::ai_target_defines>;
}

```
?

stevecotton:
> should it be like:
I'd put as much as possible in there, particularly as the `using at_target = ...` line defines the main thing that will appear in .cpp files.

I've realised that asking for all of the string_enums to be in namespaces is asking far too much, sorry. I've pushed a branch with some changes that you're welcome to cherry-pick the first couple of, and then they get less mergeable.

* a00b1b09048ffdb25335503e7c1e74ecaafc0746 Put namespaces on the string_enums that don't need .cpp changes to do so
* 72b3ea985357c0ec8431104e0f1151ab7115ad1e Add namespace to utils::irdya_datetime and utils::wesnoth_epoch
* 46ee9850b7795ffb114b5cee8498c51dbc1fa25b see description below
* e1f4391ab1bdf34e376e80a76ea659e94c988384 The TODO comments for ones that I couldn't work out what to do with

commit 46ee9850b7795ffb114b5cee8498c51dbc1fa25b
---

MAYBE FOR MAINLINE: Put ng namespace on to some string_enums
    
Game initialisation already has the `ng` namespace. That seems to have originally been a shorthand for "engine", which isn't descriptive. However, it's currently exclusively used for declarations in src/game_initialization, so keeping it as `ng` but saying that's an acronym of "New Game" would fit quite well.
    
Many unrelated .cpp files use `#define LOG_NG LOG_STREAM(info, log_engine)`, where `NG` seems to be more "engine, or miscellanous other stuff". That makes me feel that it would be wrong to say that `namespace ng` is an acronym of New Game.

stevecotton (approved):

src/game_initialization/create_engine.cpp
```diff
@@ -19,15 +19,17 @@
...
 #include "gui/dialogs/campaign_difficulty.hpp"
 #include "log.hpp"
 #include "map/exception.hpp"
 #include "map/map.hpp"
 #include "minimap.hpp"
 #include "saved_game.hpp"
+#include "side_controller.hpp"
 #include "wml_exception.hpp"
-#include "string_enums/side_controller.hpp"
+#include "game_initialization/component_availability.hpp"
```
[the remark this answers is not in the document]
Pentarctagon: Whoops, must've done a copy/paste instead of a cut/paste when reordering it. Fixed.

Issue #2323, opened 2017-12-26
Make a "map diff" tool to visualizes changes between two version of a map


gfgtdf:
migrated from the wiki easycoding page, not my own idea.

> Maintainers of maps often make small tweaks for balance purposes. For someone who doesn't know the map like the back of their hand it may be very difficult to spot exactly what changed. There isn't any really good way to do this right now -- using the 'diff' tool at command line doesn't give good results for .map files.
>
>Such a tool would be helpful for someone who maintains a map pool containing maps made by others, also for someone who wants to browse a repository and examine balance changes on several maps, which could be instructive for someone who wants to make and balance their own maps.
>
>There's any number of ways to actually do this, one might be to have a feature in the map editor that takes the name of another map and adds map labels to indicate what hexes changed. It could also be a command accessible in debug mode when playing the game, or it could be an additional command-line argument (--diff [map1] [map2] ?) for the game. (It could be a standalone application but that's probably going to end up with much more work for you.)

CelticMinstrel:
Some text diff tools do better than plain command-line diff by the way... still, a visual diff tool would be much better. I think a side-by-side diff (with altered hexes hilited in some way on both maps) would be nicer than just adding map labels though; but of course that's also more work.

I'd say this should be available both from the command-line and from the map editor.

cooljeanius:
it'd be cool if, after creating such a tool, we could get GitHub to integrate it into their own diff viewer on this site

CelticMinstrel:
That _would_ be cool, but it'll never happen.

cooljeanius:
ok so #6472 added this for GitHub pull requests, but I think there's still some desire to have this supported within the actual Wesnoth application itself, IIRC?

PR #6472, opened 2022-01-30
Add `wesnoth-map-diff`

macabeus:
Solves https://github.com/wesnoth/wesnoth/issues/2323

This PR is a proposal to add a new command-line tool, `wesnoth-map-diff`.

They aim to be a CLI to produce an image comparing two tilemaps, used by a GitHub Action to automatically comment on a PR:
image

I'm writing it using JS because it's the language I'm most used to and be easy to develop this kind of tool.
To use it, you need to have Node.js. Then on your terminal:
```
> cd utils/wesnoth-map-diff
> npm i
> node src/index.js
```


Wedge009:
Good on you for giving this a try. If you're not finished, you can always switch the PR to a draft.

.github/workflows/map-diff.yml
```diff
@@ -0,0 +1,88 @@
...
+      id: body-comment
+      run: |
+        cd ./utils/wesnoth-map-diff
+
+        body=""
+
+        for diff_image in ${{ steps.map-diff.outputs.DIFF_IMAGES }}
+        do
+          image_link=$(curl -X POST "https://api.imgur.com/3/upload" \
+            -F "image=@\"$diff_image\"" | jq ".data.link" -r)
```
stevecotton, line 78: How often will this upload images?

.github/workflows/map-diff.yml
```diff
@@ -0,0 +1,88 @@
...
+
+          diff_images="$diff_images $output_filename"
+        done
+
+        echo "::set-output name=DIFF_IMAGES::$diff_images"
+    - name: Compress images
+      run: |
+        cd ./utils/wesnoth-map-diff
+
+        sudo apt-get install -y pngquant
```
stevecotton, line 62: There's already utils/woptipng.py

.github/workflows/map-diff.yml
macabeus: It uploads an image whenever a PR with a `.map` file is opened or updated.

.github/workflows/map-diff.yml
```diff
@@ -0,0 +1,88 @@
...
+        cd ./utils/wesnoth-map-diff
+        npm run build:prod
+    - name: Get maps
+      id: get-maps
+      run: |
+        cd ./utils/wesnoth-map-diff
+
+        git fetch --depth=1 origin ${{ github.event.pull_request.base.sha }}
+
+        MAP_PATHS=$(git diff --name-status HEAD ${{ github.event.pull_request.base.sha }} | grep ".map$" | perl -pe 's/\w\s+//')
```
soliton-, line 31: You can just as well do `git diff --name-only HEAD ${{ github.event.pull_request.base.sha }} '*.map'`

.github/workflows/map-diff.yml
```diff
@@ -0,0 +1,88 @@
...
+        flat_map_paths=${MAP_PATHS[@]}
```
soliton-, line 32: Since MAP_PATHS is not an array this does nothing useful.

.github/workflows/map-diff.yml
```diff
@@ -0,0 +1,88 @@
...
+        echo "::set-output name=MAP_PATHS::$flat_map_paths"
```
soliton-, line 33: Is it useful to split the work into multiple steps that are just more shell code? It seems to provide no benefit but require this magic to hand data over to the next step.

Without this you could handle the filepaths properly with an array instead.

.github/workflows/map-diff.yml
macabeus: I didn't notice this tool before.

But anyway, `utils/woptipng.py` isn't standalone.
It depends on a set of different packages (Pillow, advdef, `imagemagick/convert`, and optipng).

I think it's simpler just to install a single package (`pngquant`) and run it on this action.

.github/workflows/map-diff.yml
macabeus: Thanks. I didn't know the flag `--name-only`. But I still need to use `grep`:
```
➜  wesnoth-map-diff git:(feat/add-wesnoth-map-diff) ✗ git diff --name-only HEAD 514b8bca9cf '*.map'
➜  wesnoth-map-diff git:(feat/add-wesnoth-map-diff) ✗ git diff --name-only HEAD 514b8bca9cf | grep ".map$"  
data/test.map
data/test2.map
```

cooljeanius:
This seems really cool; how do I use it in my repositories for UMC campaigns that I maintain?

CelticMinstrel:
The only thing I don't quite understand is… why upload the image to Imgur when GitHub itself supports image attachments on comments?

Elvish-Hunter:
Quick question: since most of our users are on Windows, how can they run this tool on Windows systems? IMHO, a text file which explains how to do so would be appreciated.

macabeus:
@cooljeanius 
> This seems really cool; how do I use it in my repositories for UMC campaigns that I maintain?

There are two options:
- run manually on your computer. You can follow the [readme](https://github.com/macabeus/wesnoth/tree/feat/add-wesnoth-map-diff/utils/wesnoth-map-diff#readme).
- run using the [github actions](https://github.com/macabeus/wesnoth/blob/feat/add-wesnoth-map-diff/.github/workflows/map-diff.yml). It should work seamlessly just by copying and pasting.

@CelticMinstrel 
> why upload the image to Imgur when GitHub itself supports image attachments on comments?

Firstly, I tried to upload the image to GitHub. But it's harder since we need to have an authentication token.
Uploading to Imgur is easier, so I picked this solution.

@Elvish-Hunter 

There is the [readme](https://github.com/macabeus/wesnoth/tree/feat/add-wesnoth-map-diff/utils/wesnoth-map-diff#readme) explaining it. But I can try to explain in more detail how to use it on Windows.

Elvish-Hunter:
> There is the [readme](https://github.com/macabeus/wesnoth/tree/feat/add-wesnoth-map-diff/utils/wesnoth-map-diff#readme) explaining it. But I can try to explain in more detail how to use it on Windows.

@macabeus Yes, I think that some detailed instructions for our Windows users would be really useful.

macabeus:
@Elvish-Hunter I just added more detail on the readme. Please, let me know if it's better now

.github/workflows/map-diff.yml
```diff
@@ -0,0 +1,82 @@
...
+        npm run build:prod
+    - name: Get maps diff
+      id: get-maps-diff
+      run: |
+        cd ./utils/wesnoth-map-diff
+
+        ## Get Maps
+        git fetch --depth=1 origin ${{ github.event.pull_request.base.sha }}
+
+        map_paths=($(git diff --name-only HEAD ${{ github.event.pull_request.base.sha }} | grep ".map$"))
```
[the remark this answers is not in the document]
macabeus: You can reproduce the issue using my test repository:
```
> git clone git@github.com:macabeus/wesnoth.git
> cd wesnoth
> git checkout feat/add-wesnoth-map-diff-backup
> git diff --name-only HEAD 514b8bca9cf -- '*.map'
```

.github/workflows/map-diff.yml
```diff
@@ -0,0 +1,82 @@
...
+
+        for map_path in "${map_paths[@]}"
+        do
+          map_filename=$(basename "$map_path")
+          git show ${{ github.event.pull_request.base.sha }}:$map_path > $map_filename
```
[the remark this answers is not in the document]
macabeus: I didn't know `##*/` but I think `basename` is more meaningful.

Pentarctagon:
If all comments have been addressed, it sounds like this can be merged?

stevecotton:
Do any of the reviewers know enough about the Node.js ecosystem to review this? I don't, @soliton- appears to be only reviewing the shell-script part, and the other commentators seem to be asking about how to run it rather than reviewing it.

.github/workflows/map-diff.yml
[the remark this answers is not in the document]
soliton-: I cannot reproduce it there either. For me git works as documented.

```
$ git diff --name-only HEAD 514b8bca9cf -- '*.map'
data/test.map
data/test2.map
```

Anyway it's not that big a deal to keep the grep as well.

soliton-:
Yeah, I didn't and don't intend to look at the js part.

soliton-:
Since it doesn't seem we'll get anyone to review the js part I would suggest to just merge this and see how it goes.

Pentarctagon:
I'll merge tomorrow then, barring anything else coming up.

PR #6473, opened 2022-01-30
Refactor the backstab unit test

stevecotton:
Based on part of Pentarctagon's #6308, and also improving
the encapsulation of the test to take a boolean value for whether
the ability should be active instead of the expected damage total.

### DIFF
--- a/data/test/scenarios/backstab.cfg
+++ b/data/test/scenarios/backstab.cfg
@@ -1,96 +1,167 @@
 # wmllint: no translatables
+
 # Test that the backstab ability does double-damage, under the correct circumstances
 
-# Given a set of directions, create units on the opposite directions around
-# Bob's starting location. Note: the order that they spawn might not match the
-# order in the ANTIDIRECTIONS argument.
-#
-# The first unit spawned then attacks Bob, and we check that the
-# EXPECTED_DAMAGE was inflicted.
-#
-# Any commands in EXTRA_SETUP can access the new units via the spawn_points
-# array.
-#define TEST_BACKSTAB NAME ANTIDIRECTIONS EXPECTED_DAMAGE EXTRA_SETUP
-{GENERIC_UNIT_TEST {NAME} (
-    [event]
-        # Using this event instead of "start" because units get healed after "start", which makes debugging more confusing
-        name=side 1 turn 1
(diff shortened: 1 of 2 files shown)


Wedge009:
I'm not very familiar with WML tests but as a general rule I try to write self-descriptive names for the tests in preference to overly-long explanatory comments as the latter can become out of date as code/tests are updated. As I recall, back-stab doubles damage for the attacker when an attacker's ally is directly opposite the defender. But it looks like the test names are pre-existing and I assume they haven't been changed to avoid breaking tests.

It makes sense to me to have the test focus on the activation of the back-stab damage bonus vs checking arbitrary damage values in the parameters as the latter seems more prone to errors.

stevecotton:
It would be easy to change the tests' names too, please provide better suggestions.
The name just has to match in the test and in the wml_test_schedule file.

Wedge009:
I don't know what the convention is, but if I understood the tests correctly I was thinking along the lines of:
* `TEST_BACKSTAB "backstab_should_activate_with_attacker_ally_behind_defender"`
* `TEST_BACKSTAB "backstab_should_not_activate_without_attacker_ally_behind_defender"`
* `TEST_BACKSTAB "backstab_should_not_activate_with_disabled_attacker_ally_behind_defender"`
* `TEST_BACKSTAB :backstab_should_not_activate_with_defender_ally_behind_defender"`

A bit long, I know, but at least it should be clear to a newcomer what the tests are checking for.

stevecotton:
> A bit long, I know, but at least it should be clear to a newcomer what the tests are checking for.

A bit shorter (haven't pushed it yet, still pondering):

* `backstab_active_with_accomplice_behind_bob`
* `backstab_inactive_with_triangular_formation`
* `backstab_inactive_with_statue_behind_bob`
* `backstab_inactive_with_bobs_ally_behind_bob`

Wedge009:
As long as people reading can quickly and easily understand the intent of the test solely by looking at the test name. I find that's generally easier to accomplish when being a bit more verbose but whatever works.

stevecotton:
Are the ones that I posted clear to you, or not?

Wedge009:
It should be fine - definitely an improvement over the original anyway.

Wedge009 (approved):

stevecotton:
@Pentarctagon, is this OK for you?

Pentarctagon (approved):

PR #6477, opened 2022-01-31
1.16 variations 20220129

doofus-01:
I'm not completely sure these can go into 1.16.  They only change the appearance slightly and won't break anything outright, but new transition behavior between terrain codes Xos and Aa might be too much for stable?


Pentarctagon:
So the transition looks different, but there's no functionality change anywhere?

doofus-01:
Right.  Here's a screenshot.  It affects the appearance of Xos when it is next to frozen terrain.  Other Xo* aren't affected, and I can't think of a reason anyone would want the default Xos on snow/ice.  It's got an issue or two with corner cases (lower-right in the screenshot), but I think it's an improvement for most cases. 
Screenshot_20220130

Pentarctagon:
Alright, that sounds fine for 1.16.

Issue #4133, opened 2019-06-22
Loading a replay doesn't reset stats


jostephd:
1. `wesnoth -t`
2. attack something
3. save replay
4. load replay
5. At this point, the statistics dialog shows the statistics as they would be at the end of the replay. It should show all zero.
6. Pressing the Stop button resets the statistics back to all-zero (see 8d1c2780b7799de94a5dd0718943e1b0e22f690d)

The bug does not happen if I change `mid_scenario=yes` to `mid_scenario=no` in the replay savefile. If I do that I get two entries for the scenario, though (that's #4111).

I think the problem might be in this block, but I didn't debug any further:

https://github.com/wesnoth/wesnoth/blob/c0191a5a165995bc49e73ea19281c2cab780ee0d/src/game_launcher.cpp#L675-L685

Thanks to Hardwood for initial reproduction steps.

gfgtdf:
i was not able to reproduce this given the steps in the first post in current master.

jostephd:
I can reproduce in current master. In step 3, I chose `Save Replay` from the menu, not the ordinary save command. And I loaded the game directly from the mainmap, without going through the title screen.

gfgtdf:
oh ok then i tested it dirrecntly, in that case what should be done is probalby to replace the `state_.is_mid_game_save() && ` there with a 'is_start_of_scenario_save' check.

CelticMinstrel:
Apparently fixed in #7585 / 01f28b1.

PR #6479, opened 2022-01-31
reset stats on replay load

silverthefox123:
solves [#4133](https://github.com/wesnoth/wesnoth/issues/4133)


Wedge009:
If this really resolves the issue, the commentary needs to be updated as well.

(Also, the linked issue doesn't match the link text, but I blame that on GitHub - I've seen that happen before when typing issue links.)

Pentarctagon:
Has this been tested whether it causes any issues with loading saves/replays?

Issue #6475, opened 2022-01-30
[1.16.3] Check for missing credits in about.cfg


stevecotton:
A reminder task to check that new contributors have been credited. Definitely missing are:
* Bruno Macabeus (Trickster)
* Luis Miguel Iglesias (Descacharrado)

Wedge009:
Just wondering how far back we go? I notice not all of the top 100 committers (according to https://github.com/wesnoth/wesnoth/graphs/contributors) are included in `about.cfg`.

On top of @macabeus and @Descacharrado, there's also @max-torch, @knyghtmare, @czyzby, @severin, @Discontinuum, @evanmiller, @larb0b, @ddiefenthaler who seem to be absent (unless they're not credited by their GitHub alias). I only reviewed commits back as far as 2021-06-21 when 1.16 branch started. Is there a 'minimum threshold' of commits required for inclusion?

Also, not sure if he's only worked on master so far, but there's @Jonathan-Kelly for the WoF campaign too. Also, I don't see @nemaara or yumi in the credits...

max-torch:
I would be happy to be added to `about.cfg`
#6480

czyzby:
Can confirm that I didn't update the `about.cfg` in my pull request. If I also qualify, I wouldn't mind an entry in the `"Artwork and Graphics"` section along the lines of:

```ini
    [entry]
        name = "MJ (czyzby)"
        comment = "fixes and palette adjustments of the saurian sprites"
    [/entry]
```

Descacharrado:
I'd be happy to add myself in the next PR I open so you do not have to keep track of me, unless the "collecting everyone in one go" is what we are aiming to do. I also agree with Celtic:

> being added to the credits should be opt-out, not opt-in. Add everyone unless they request otherwise.

I'd say that adding the Github nickname should be the norm and then, if the contributor wants to opt out or modify their entry, they can do so similarly to what @max-torch did, opening a PR.

nemaara:
You can leave me off for now, thanks!

knyghtmare:
I got my forum name on the world conquest credits so not sure if I should added in about.cfg or not.

severin:
I'm fine either way 🙂

ddiefenthaler:
Hey, you can leave me out too.
I only made a very small contribution and not even to the game itself.
Thanks anyway!

stevecotton:
I'll gladly merge PRs for this, but it's easy to add names in a batch too.

[...]

PR #6480, opened 2022-01-31
Add max_torch to about.cfg

max-torch:
Add max_torch as miscellaneous contributor.


Wedge009:
I'd merge this, but I'm not sure how @stevecotton wants to handle #6475, whether it's worth collecting everyone in one go or just adding everyone separately.

stevecotton:
Thanks. I'll happily merge it when people add these as PRs, but will also happily do a batch afterwards too.

PR #6481, opened 2022-02-01
Update credits

czyzby:
Add missing entry to the `about.cfg` file. See #6475.

### DIFF
--- a/data/core/about.cfg
+++ b/data/core/about.cfg
@@ -439,6 +439,10 @@
         comment = "portraits (for the campaign 'The Rise of Wesnoth')"
     [/entry]
     [entry]
+        name = "MJ (czyzby)"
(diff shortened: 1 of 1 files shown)


stevecotton:
My apologies for asking you to create this PR before I asked @doofus-01 about whether the credit should go in the Artists or Miscellaneous Contributors section.

@doofus-01, please would you say which section this should be in?

czyzby:
@stevecotton I didn't realize that sprite changes can be credited in a different section. If you prefer, I can also close this PR and wait for the bulk update.

doofus-01 (approved):
The credits are a mess, I don't think the section matters much at this point.  More important is to get the "comment" key to be an accurate description, and that part looks OK to me.

Issue #5873, opened 2021-06-19
Confusing "attempted to deprecate something" error message, and related docs


stevecotton:
For example, the following line:
```
wesnoth.xxx = wesnoth.deprecate_api('wesnoth.xxx', 'wesnoth.yyy.xxx', 1, nil, wesnoth.yyy.xxx)
```
will generate the error
```
Attempted to deprecate something that is not a table or function: wesnoth.xxx -> wesnoth.yyy.xxx, where wesnoth.xxx = nil
```
However, at this point it's completely correct that `wesnoth.xxx` is `nil`. The error is really that `wesnoth.yyy.xxx` is missing.

Looking further, the documentation, names of arguments, and error messages generated by `wesnoth.deprecate_api` all seem wrong. I think these arguments should really be called `(elem_name, replacement_name, level, version, replacement, detail_msg)`, and the error message in this case end `where ".. replacement_name .." = " .. tostring(replacement)`.

```
-- Marks a function or subtable as deprecated.
-- Parameters:
---- elem_name: the full name of the element being deprecated (including the module)
---- replacement: the name of the element that will replace it (including the module)
---- level: deprecation level (1-4)
---- version: the version at which the element may be removed (level 2 or 3 only)
---- Set to nil if deprecation level is 1 or 4
---- elem: The actual element being deprecated
---- detail_msg: An optional message to add to the deprecation message
function wesnoth.deprecate_api(elem_name, replacement, level, version, elem, detail_msg)

	...

	else
		wesnoth.log('warn', "Attempted to deprecate something that is not a table or function: " ..
			elem_name .. " -> " .. replacement .. ", where " .. elem_name .. " = " .. tostring(elem))
```

CelticMinstrel:
I think the issue here is that it's not actually possible to get the name of the `elem`, so it just guesses that its name is `elem_name`. That used to be true most of the time, but lately the `elem` is actually the `replacement` instead.

It's not that the argument names are wrong, per se. The use-case implied by the current names and the use-case implied by your alternate names are both supported.

Pentarctagon:
When running the WML unit tests with the `-d` argument, this causes a lot of warnings/errors such as:
```
20210619 23:42:39 error scripting/lua: Lua Error: When executing, Lua runtime error: lua/core/_initial.lua:125: attempt to index a nil value (field 'paths')
stack traceback:
	lua/core/_initial.lua:125: in main chunk
	[C]: in field 'dofile'
[...]

PR #6483, opened 2022-02-03
Lua API: Add inline documentation for pretty much everything

CelticMinstrel:
I did this several months ago but never got around to submitting it, so there might be something missing here and there if things have changed since then.

In addition to covering things defined in Lua, this includes stub definitions with documentation of everything defined in C++, in the utils/emmylua folder.

The Lua add-on for Visual Studio Code uses this format to improve tooltips and auto-completion.

### DIFF
--- a/data/lua/core/_initial.lua
+++ b/data/lua/core/_initial.lua
@@ -13,10 +13,11 @@ local _ = wesnoth.textdomain "wesnoth"
 ---@param level '1'|'2'|'3'|'4' deprecation level (1-4)
 ---@param version string|nil the version at which the element may be removed (level 2 or 3 only)
 --- Set to nil if deprecation level is 1 or 4
----@param elem T The actual element being deprecated, ignored if level is 4.
+---@param elem T The actual element for the deprecation, ignored if level is 4.
 --- This can be the original, undeprecated element, but it does not have to be.
+--- It could also be a wrapper that presents a different API, for example.
 ---@param detail_msg? string An optional message to add to the deprecation message
----@return T elem_deprecated #A reference to the element, altered to trigger a deprecation message when used.
+---@return T elem_deprecated #A wrapper around the element, which triggers a deprecation message when used.
 --- If it is a function, the message is triggered the first time it is called.
 --- If it is a table, the message is triggered when a key is written or read on the table.
 function wesnoth.deprecate_api(elem_name, replacement_name, level, version, elem, detail_msg)
--- a/data/lua/core/_initial.lua
+++ b/data/lua/core/_initial.lua
@@ -8,14 +8,18 @@ local _ = wesnoth.textdomain "wesnoth"
 
 -- Marks a function or subtable as deprecated.
 ---@generic T
----@param elem_name string the full name of the element being deprecated (including the module)
----@param replacement_name string the name of the element that will replace it (including the module)
+---@param elem_name string the full name of the element being deprecated (including the module), to be shown in the deprecation message
+---@param replacement_name string the name of the element that will replace it (including the module), to be shown in the deprecation message
 ---@param level '1'|'2'|'3'|'4' deprecation level (1-4)
 ---@param version string|nil the version at which the element may be removed (level 2 or 3 only)
 --- Set to nil if deprecation level is 1 or 4
+--- Will be shown in the deprecation message
 ---@param elem T The actual element for the deprecation, ignored if level is 4.
---- This can be the original, undeprecated element, but it does not have to be.
+--- This can be the original, pre-deprecated element, but it does not have to be.
 --- It could also be a wrapper that presents a different API, for example.
+--- If deprecating a function, that would mean a wrapper function that calls the new API.
+--- If deprecating a table, you would need to provide a table with __index and __newindex metamethods that call the new API.
+--- This is the only argument that affects the functionality of the resulting deprecation wrapper.
 ---@param detail_msg? string An optional message to add to the deprecation message
 ---@return T elem_deprecated #A wrapper around the element, which triggers a deprecation message when used.
 --- If it is a function, the message is triggered the first time it is called.
--- a/data/lua/core/_initial.lua
+++ b/data/lua/core/_initial.lua
@@ -10,6 +10,7 @@ local _ = wesnoth.textdomain "wesnoth"
 ---@generic T
 ---@param elem_name string the full name of the element being deprecated (including the module), to be shown in the deprecation message
 ---@param replacement_name string the name of the element that will replace it (including the module), to be shown in the deprecation message
+--- Can be nil if there is not replacement, though this should be an unlikely situation
 ---@param level '1'|'2'|'3'|'4' deprecation level (1-4)
 ---@param version string|nil the version at which the element may be removed (level 2 or 3 only)
 --- Set to nil if deprecation level is 1 or 4
@@ -33,7 +34,7 @@ function wesnoth.deprecate_api(elem_name, replacement_name, level, version, elem
 	if type(level) ~= "number" or level < 1 or level > 4 then
 		local err_params = {level = level}
 		-- Note: This message is duplicated in src/deprecation.cpp
-		-- Any changes should be mirrorred there.
+		-- Any changes should be mirrored there.
 		error((_"Invalid deprecation level $level (should be 1-4)"):vformat(err_params))
 	end
 	local msg_shown = false
--- a/data/lua/core/_initial.lua
+++ b/data/lua/core/_initial.lua
@@ -7,14 +7,15 @@ print("Loading Lua core files...")
 local _ = wesnoth.textdomain "wesnoth"
 
 -- Marks a function or subtable as deprecated.
--- Parameters:
----- elem_name: the full name of the element being deprecated (including the module)
----- replacement: the name of the element that will replace it (including the module)
----- level: deprecation level (1-4)
----- version: the version at which the element may be removed (level 2 or 3 only)
----- Set to nil if deprecation level is 1 or 4
----- elem: The actual element being deprecated, ignored if level is 4
----- detail_msg: An optional message to add to the deprecation message
+---@generic T
+---@param elem_name string the full name of the element being deprecated (including the module)
+---@param replacement string the name of the element that will replace it (including the module)
+---@param level '1'|'2'|'3'|'4' deprecation level (1-4)
+---@param version string|nil the version at which the element may be removed (level 2 or 3 only)
+--- Set to nil if deprecation level is 1 or 4
+---@param elem T The actual element being deprecated, ignored if level is 4
+---@param detail_msg? string An optional message to add to the deprecation message
+---@return T
 function wesnoth.deprecate_api(elem_name, replacement, level, version, elem, detail_msg)
 	if wesnoth.game_config.strict_lua then return nil end
 	local message = detail_msg or ''
--- a/data/lua/core/_initial.lua
+++ b/data/lua/core/_initial.lua
@@ -9,18 +9,21 @@ local _ = wesnoth.textdomain "wesnoth"
 -- Marks a function or subtable as deprecated.
 ---@generic T
 ---@param elem_name string the full name of the element being deprecated (including the module)
----@param replacement string the name of the element that will replace it (including the module)
+---@param replacement_name string the name of the element that will replace it (including the module)
 ---@param level '1'|'2'|'3'|'4' deprecation level (1-4)
 ---@param version string|nil the version at which the element may be removed (level 2 or 3 only)
 --- Set to nil if deprecation level is 1 or 4
----@param elem T The actual element being deprecated, ignored if level is 4
+---@param elem T The actual element being deprecated, ignored if level is 4.
+--- This can be the original, undeprecated element, but it does not have to be.
 ---@param detail_msg? string An optional message to add to the deprecation message
----@return T
-function wesnoth.deprecate_api(elem_name, replacement, level, version, elem, detail_msg)
+---@return T elem_deprecated #A reference to the element, altered to trigger a deprecation message when used.
+--- If it is a function, the message is triggered the first time it is called.
+--- If it is a table, the message is triggered when a key is written or read on the table.
+function wesnoth.deprecate_api(elem_name, replacement_name, level, version, elem, detail_msg)
 	if wesnoth.game_config.strict_lua then return nil end
 	local message = detail_msg or ''
-	if replacement then
-		message = message .. " " .. (_"(Note: You should use $replacement instead in new code)"):vformat{replacement = replacement}
+	if replacement_name then
+		message = message .. " " .. (_"(Note: You should use $replacement instead in new code)"):vformat{replacement = replacement_name}
 	end
 	if type(level) ~= "number" or level < 1 or level > 4 then
 		local err_params = {level = level}
@@ -56,7 +59,7 @@ function wesnoth.deprecate_api(elem_name, replacement, level, version, elem, det
 		if type(old_mt) ~= "table" then
 			-- See https://github.com/wesnoth/wesnoth/issues/4584#issuecomment-555788446
 			wesnoth.log('warn', "Attempted to deprecate a table with a masked metatable: " ..
-				elem_name .. " -> " .. replacement .. ", where getmetatable(" .. elem_name .. ") = " .. tostring(old_mt))
+				elem_name .. " -> " .. replacement_name .. ", where getmetatable(" .. elem_name .. ") = " .. tostring(old_mt))
 			return elem
 		end
 		local mt = {}
@@ -80,7 +83,7 @@ function wesnoth.deprecate_api(elem_name, replacement, level, version, elem, det
 		return setmetatable({}, mt)
 	else
 		wesnoth.log('warn', "Attempted to deprecate something that is not a table or function: " ..
-			elem_name .. " -> " .. replacement .. ", where " .. elem_name .. " = " .. tostring(elem))
+			elem_name .. " -> " .. replacement_name .. ", which is " .. tostring(elem))
 	end
 	return elem
 end
--- a/data/lua/core/_initial.lua
+++ b/data/lua/core/_initial.lua
@@ -15,7 +15,7 @@ local _ = wesnoth.textdomain "wesnoth"
 ---@param version string|nil the version at which the element may be removed (level 2 or 3 only)
 --- Set to nil if deprecation level is 1 or 4
 --- Will be shown in the deprecation message
----@param elem T The actual element for the deprecation, ignored if level is 4.
+---@param elem T Implementation of the compatibility layer, ignored if level is 4.
 --- This can be the original, pre-deprecated element, but it does not have to be.
 --- It could also be a wrapper that presents a different API, for example.
 --- If deprecating a function, that would mean a wrapper function that calls the new API.
--- a/data/lua/core/gui.lua
+++ b/data/lua/core/gui.lua
@@ -2,6 +2,9 @@
 --[========[GUI2 Dialog Manipulations]========]
 print("Loading GUI module...")
 
+---Show a basic alert dialog with a single button
+---@param title string Dialog title string
+---@param msg string Detail message
 function gui.alert(title, msg)
 	if not msg then
 		msg = title;
@@ -10,6 +13,10 @@ function gui.alert(title, msg)
 	gui.show_prompt(title, msg, "ok", true)
 end
 
+---Show a basic prompt dialog with two buttons
+---@param title string Dialog title string
+---@param msg string Detail message
+---@return boolean #True if OK or Yes was clicked
 function gui.confirm(title, msg)
 	if not msg then
 		msg = title;
@@ -18,13 +25,10 @@ function gui.confirm(title, msg)
 	return gui.show_prompt(title, msg, "yes_no", true)
 end
 
---! Displays a WML message box with attributes from table @attr and options
---! from table @options.
---! @return the index of the selected option.
---! @code
---! local result = gui.get_user_choice({ speaker = "narrator" },
---!     { "Choice 1", "Choice 2" })
---! @endcode
+---Displays a WML message box with attributes from attr and options from options.
+---@param attr WML
+---@param options string[]
+---@return integer #the index of the selected option.
 function gui.get_user_choice(attr, options)
 	local result = 0
 	function gui.__user_choice_helper(i)
--- a/data/lua/core/gui.lua
+++ b/data/lua/core/gui.lua
@@ -26,8 +26,8 @@ function gui.confirm(title, msg)
 end
 
 ---Displays a WML message box with attributes from attr and options from options.
----@param attr WML
----@param options string[]
+---@param attr WML The contents of a [message] tag, without any [option]s.
+---@param options string[]|gui_narration_option_info[] A list of options to show in the message box
 ---@return integer #the index of the selected option.
 function gui.get_user_choice(attr, options)
 	local result = 0
--- a/data/lua/core/interface.lua
+++ b/data/lua/core/interface.lua
@@ -5,8 +5,11 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 
 	wesnoth.interface.select_unit = wesnoth.units.select
 
-	--! Fakes the move of a unit satisfying the given @a filter to position @a x, @a y.
-	--! @note Usable only during WML actions.
+	-- Fakes the move of a unit satisfying the given filter to position x, y.
+	-- Usable only during WML actions.
+	---@param filter WML
+	---@param to_x integer
+	---@param to_y integer
 	function wesnoth.interface.move_unit_fake(filter, to_x, to_y)
 		local moving_unit = wesnoth.units.find_on_map(filter)[1]
 		local from_x, from_y = moving_unit.x, moving_unit.y
--- a/data/lua/core/interface.lua
+++ b/data/lua/core/interface.lua
@@ -5,8 +5,8 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 
 	wesnoth.interface.select_unit = wesnoth.units.select
 
-	-- Fakes the move of a unit satisfying the given filter to position x, y.
-	-- Usable only during WML actions.
+	---Fakes the move of a unit satisfying the given filter to position x, y.
+	---Usable only during WML actions.
 	---@param filter WML
 	---@param to_x integer
 	---@param to_y integer
--- a/data/lua/core/map.lua
+++ b/data/lua/core/map.lua
@@ -39,9 +39,10 @@ if wesnoth.kernel_type() ~= "Application Lua Kernel" then
 	-- A        A^       A^B      ^        ^B
 	-- implied mode:
 	-- both     base     both     overlay  overlay
-	
-	---Adjusts a terrain code to produce one that will replace the base terrain only
-	---@param code string A terrain code
+
+	---Adjusts a terrain code to produce one that will replace the base terrain only,
+	---when the adjusted code is assigned to a terrain hex on the map
+---@param code string A terrain code
 	---@return string #The adjusted terrain code
 	function wesnoth.map.replace_base(code)
 		local base, overlay = wesnoth.map.split_terrain_code(code)
@@ -54,7 +55,8 @@ if wesnoth.kernel_type() ~= "Application Lua Kernel" then
 		end
 	end
 
-	---Adjusts a terrain code to produce one that will replace the overlay terrain only
+	---Adjusts a terrain code to produce one that will replace the overlay terrain only,
+	---when the adjusted code is assigned to a terrain hex on the map
 	---@param code string A terrain code
 	---@return string #The adjusted terrain code
 	function wesnoth.map.replace_overlay(code)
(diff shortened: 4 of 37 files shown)


data/lua/core/map.lua
```diff
@@ -120,7 +162,7 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 		elseif key == 2 then
 			return self.y
 		elseif type(key) ~= string or (#key > 0 and key[0] ~= '_') then
-			return hex_mt[key]
+			return hex_methods[key]
```
CelticMinstrel, line 165: There's an issue here, as a) I don't remember why the `~= string` test was added and b) the `~= string` test itself is wrong. I think it's out of scope for this PR though.

utils/emmylua/gui.lua
```diff
@@ -0,0 +1,226 @@
...
+---@field label tstring
+---@field description tstring
+---@class gui_narration_text_params
+---@field label tstring
+---@field text string
+---@field max_length integer
+
+---Shows a message dialog.
+---@param attributes gui_narration_params
+---@param options? string[]|gui_narration_option_info
```
CelticMinstrel, line 38: This seems slightly wrong, at minimum I missed `[]` at the end?

utils/emmylua/gui.lua
```diff
@@ -0,0 +1,226 @@
...
+---A static text label
+---@class label : simple_widget
+---A simple image
+---@class image : simple_widget
+---A simple button that triggers repeatedly if the mouse is held down
+---@class repeating_button : button
+---A generic container that lays out its children in a grid
+---@class grid : widget
+
+---@class gui.widget
```
CelticMinstrel, line 186: I'm not confident that the above covers every possible type of widget… can anyone see one I've missed?

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,267 @@
...
+---@field backstabs boolean
+---@field rounds integer
+---@field firststrike boolean
+---@field drains boolean
+---@field drain_constant integer
+---@field drain_percent number
+---@field attack_num integer
+---@field name string
+
+---Construct a function that generates random namespace
```
CelticMinstrel, line 62: … how did that end up saying "namespace" without me noticing?

data/lua/core/_initial.lua
```diff
@@ -7,14 +7,15 @@ print("Loading Lua core files...")
...
----- detail_msg: An optional message to add to the deprecation message
+---@generic T
+---@param elem_name string the full name of the element being deprecated (including the module)
+---@param replacement string the name of the element that will replace it (including the module)
+---@param level '1'|'2'|'3'|'4' deprecation level (1-4)
+---@param version string|nil the version at which the element may be removed (level 2 or 3 only)
+--- Set to nil if deprecation level is 1 or 4
+---@param elem T The actual element being deprecated, ignored if level is 4
+---@param detail_msg? string An optional message to add to the deprecation message
+---@return T
```
stevecotton, line 18: Here, `T` could do with documentation.

 #5873 has my review of the other stuff here.

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,267 @@
...
+---@type table<string, fun(cfg:WML)>
+wesnoth.wml_actions = {}
+
+---@type table<string, fun(cfg:WML):boolean
+wesnoth.wml_conditionals = {}
+
+---@type table<string, fun(unit:unit, cfg:WML)
+wesnoth.effects = {}
+
+---@type table<string, fun(cfg:WML)
```
CelticMinstrel, line 123: Looks like I have several missing `>` here…

data/lua/core/gui.lua
```diff
@@ -18,13 +25,10 @@ function gui.confirm(title, msg)
...
---! from table @options.
---! @return the index of the selected option.
---! @code
---! local result = gui.get_user_choice({ speaker = "narrator" },
---!     { "Choice 1", "Choice 2" })
---! @endcode
+---Displays a WML message box with attributes from attr and options from options.
+---@param attr WML
+---@param options string[]
+---@return integer #the index of the selected option.
```
stevecotton, line 31: Removing the example from the docs needs some documentation to replace it.

data/lua/core/interface.lua
```diff
@@ -5,8 +5,11 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
 
 	wesnoth.interface.select_unit = wesnoth.units.select
 
-	--! Fakes the move of a unit satisfying the given @a filter to position @a x, @a y.
-	--! @note Usable only during WML actions.
+	-- Fakes the move of a unit satisfying the given filter to position x, y.
+	-- Usable only during WML actions.
```
stevecotton, line 9: Should these lines start with a triple-hyphen?

utils/emmylua/wesnoth/paths.lua
```diff
@@ -0,0 +1,41 @@
+---@meta
+
+wesnoth.paths = {}
+
+---Find a good path between two hexes
+---@alias path_options {max_cost:integer, ignore_units:boolean, ignore_teleport:boolean, viewing_side:integer, width:integer, height:integer, include_borders:boolean, calculate:fun(x:integer, y:integer, cost:integer):integer}
```
CelticMinstrel, line 6: Maybe I should phrase this as a `@class` instead?

utils/emmylua/wesnoth/sides.lua
```diff
@@ -0,0 +1,174 @@
...
+---@param persistent boolean
+---@param defeat_condition string
+---@param starting_location location
+---@param recruit string[]
+---@param faction string
+---@param faction_name tstring
+---@param chose_random boolean
+---@param lost boolean
+---@param variables table<string, WML>
+---@param __cfg WMLTable
```
CelticMinstrel, line 44: …why are these all `@param` instead of `@field`…?

data/lua/core/interface.lua
CelticMinstrel: I think the triple hyphen is required for the `@` directives but not for general documentation… still, I should probably keep it consistent.

data/lua/core/map.lua
```diff
@@ -167,48 +209,75 @@ if wesnoth.kernel_type() == "Game Lua Kernel" then
...
+	function hex_methods:set_fogged(side, val)
 		if val then
 			wesnoth.sides.place_fog(side, {self})
 		else
 			wesnoth.sides.remove_fog(side, {self})
 		end
 	end
 
-	function hex_mt:label_for(who)
+	---Get a label placed by a specific side
```
stevecotton, line 248: The code in game_lua_kernel::intf_get_label looks a lot more complex than this, it seems to be "visible to" along with handling of allied sides and recognising strings as names too.

data/lua/core/gui.lua
CelticMinstrel: How's c43b07e344452925570fd0f78e833ba5a0d0dff5 look?

data/lua/core/_initial.lua
CelticMinstrel: Does fc5ea0cd372581db57614eaf0b3cad48db84f25c help?

data/lua/core/map.lua
```diff
@@ -1,10 +1,21 @@
 --[========[Map module]========]
 print("Loading map module...")
 
+---Splits a terrain code into base and overlay
+---@param code string
+---@return string #The base terrain
```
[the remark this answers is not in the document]
CelticMinstrel: Why would the base be nil?

data/lua/core/_initial.lua
stevecotton: > A reference to the element

Something like "a wrapper around" would work better for me. I can understand (and think it's accurate to say) that for functions it returns a wrapper function which prints the warning message.

> --@param elem T The actual element being deprecated, ignored if level is 4

That's true for one use-case of this function, but false and confusing for another, as discussed in #5873.

data/lua/core/gui.lua
stevecotton: Ok

data/lua/core/map.lua
```diff
@@ -28,6 +39,10 @@ if wesnoth.kernel_type() ~= "Application Lua Kernel" then
 	-- A        A^       A^B      ^        ^B
 	-- implied mode:
 	-- both     base     both     overlay  overlay
+	
+	---Adjusts a terrain code to produce one that will replace the base terrain only
```
stevecotton, line 43: I think this should be "that will replace the base terrain only, when the adjusted code is used as an argument to wesnoth.map.some_function".

There seem to be no callers of this function in mainline, so I can't check my assumption against example code.

data/lua/core/map.lua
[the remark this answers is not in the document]
stevecotton: I now think it can't be null, however that means half of `replace_base` is dead code.

data/lua/core/_initial.lua
CelticMinstrel: More clarification in 24a0a1055cd8e2e0c35ab0e44958a0463ba53dcd.

data/lua/core/map.lua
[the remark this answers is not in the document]
CelticMinstrel: No, looking at it again, I think you're right. It doesn't necessarily deal with a terrain code taken from a map – the code could be something like `^Ef`, which doesn't have a base.

stevecotton:
Untested, but I think that returns an empty string instead of nil.

CelticMinstrel:
Oh right, you're probably right. So I'll delete that last commit then.

CelticMinstrel:
Okay, I see three unresolved comments… not counting the two that are unrelated to the actual content of this PR.

data/lua/on_event.lua
```diff
@@ -21,20 +21,23 @@ wesnoth.game_events.on_event = function(eventname)
 end
 
 
-local function on_event(eventname, arg1, arg2)
+---Handle an event
+---@param eventname string
+---@param priority? number
```
stevecotton, line 26: What is the priority used for? Are lower or higher numbers higher priority?

Should either document or deprecate the possibility of having a function argument without a priority argument.

data/lua/on_event.lua
```diff
@@ -21,20 +21,23 @@ wesnoth.game_events.on_event = function(eventname)
 end
 
 
-local function on_event(eventname, arg1, arg2)
+---Handle an event
```
stevecotton, line 24: Looks like "register an event handler" rather than actually handling the event here.

data/lua/wml/items.lua
```diff
@@ -48,6 +52,23 @@ function wesnoth.interface.remove_item(x, y, name)
...
+---@field filter_team WML
+---@field visible_in_fog boolean
+---@field redraw boolean
+---@field name string
+---@field z_order integer
+---@field variables WMLTable
+
+---Get items on a given hex
+---@param x integer
+---@param y integer
```
stevecotton, line 70: I haven't tested, but can these be ranges instead?

data/lua/on_event.lua
CelticMinstrel: I'm not sure. I was planning to deprecate the whole `on_event` module anyway, once that events PR gets finished and merged…

data/lua/on_event.lua
CelticMinstrel: Good point.

data/lua/on_event.lua
CelticMinstrel: Checking the code, it looks like the list is ordered by decreasing priority.

data/lua/wml/items.lua
CelticMinstrel: They're an index to a `location_set`, so no.

data/lua/core/_initial.lua
```diff
@@ -7,19 +7,24 @@ print("Loading Lua core files...")
...
----- replacement: the name of the element that will replace it (including the module)
----- level: deprecation level (1-4)
----- version: the version at which the element may be removed (level 2 or 3 only)
----- Set to nil if deprecation level is 1 or 4
----- elem: The actual element being deprecated, ignored if level is 4
----- detail_msg: An optional message to add to the deprecation message
-function wesnoth.deprecate_api(elem_name, replacement, level, version, elem, detail_msg)
+---@generic T
+---@param elem_name string the full name of the element being deprecated (including the module)
+---@param replacement_name string the name of the element that will replace it (including the module)
```
stevecotton, line 12: Might be helpful to say which of these strings have a functional effect, and which are only used in text printed to the user / UMC author.

Some of the code handles replacement_name being empty or nil, but other parts seem to expect it to be non-empty.

data/lua/core/map.lua
```diff
@@ -86,6 +115,22 @@ if wesnoth.kernel_type() ~= "Application Lua Kernel" then
...
 if wesnoth.kernel_type() == "Game Lua Kernel" then
+	---Represents a reference to a single hex on the map
+	---@class terrain_hex
+	---@field x integer
+	---@field y integer
+	---@field fogged boolean
+	---@field shrouded boolean
+	---@field team_label? string|tstring
+	---@field global_label? string|tstring
+	---@field label? string|tstring
```
stevecotton, line 126: `label` could do with documentation about what it does when there's are multiple labels on the hex.

data/lua/core/map.lua
```diff
@@ -86,6 +115,22 @@ if wesnoth.kernel_type() ~= "Application Lua Kernel" then
...
+	---@field terrain string
+	---@field base_terrain string
+	---@field overlay_terrain string
+	---@field info terrain_info
+	---@field time_of_day time_of_day
+	---@field illuminated_time time_of_day
```
stevecotton, line 132: Please document the difference between the two time_of_day accessors.

data/lua/core/mathx.lua
```diff
@@ -1,6 +1,10 @@
 --[========[Additional mathematical functions]========]
 print("Loading mathx module...")
 
+---Pick a random choice from a list of value_second
```
stevecotton, line 4: Should `value_second` be just `values`?

data/lua/core/mathx.lua
```diff
@@ -1,6 +1,10 @@
...
+---@param possible_values string|table
```
stevecotton, line 5: I wouldn't have guessed that the string could contain ranges from this description.

data/lua/core/persistent_tags.lua
```diff
@@ -3,6 +3,12 @@
 if wesnoth.kernel_type() == "Game Lua Kernel" then
 	print("Creating persistent_tags table...")
 
+	---Defines a tag for custom saved game data.
+	---@class persistent_tag
+	---@field read fun(cfg:WMLTable)
+	---@field write fun(add:fun(WMLTable))
+
```
stevecotton, line 10: Extra blank line

data/lua/core/stringx.lua
```diff
@@ -1,11 +1,22 @@
...
+---@param str string
+---@param sep? string
+---@param esc? string
+---@return string[]
 function stringx.escaped_split(str, sep, esc)
 	esc = esc or '\\'
 	return stringx.split(str, sep, {escape = esc, strip_spaces = true, remove_empty = true})
 end
 
+---Split a string on a separator, skipping separatorss that are enclosed in quotes
```
stevecotton, line 14: `separatorss`

data/lua/core/stringx.lua
```diff
@@ -1,11 +1,22 @@
...
+---@param str string
+---@param sep? string
+---@param left? string
+---@param right? string
```
stevecotton, line 18: Documentation is unclear about what `left` and `right` are. From the first line, I'd expect this to default to quotation marks.

data/lua/core/wml.lua
```diff
@@ -157,6 +191,8 @@ function wml.literal(cfg)
 	end
 end
 
+---@param cfg WML
```
stevecotton, line 194: docs?

utils/emmylua/README.md
```diff
@@ -0,0 +1,32 @@
+
```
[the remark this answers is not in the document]
CelticMinstrel: ???

data/lua/core/map.lua
```diff
@@ -86,6 +115,22 @@ if wesnoth.kernel_type() ~= "Application Lua Kernel" then
...
+	---@field y integer
+	---@field fogged boolean
+	---@field shrouded boolean
+	---@field team_label? string|tstring
+	---@field global_label? string|tstring
+	---@field label? string|tstring
+	---@field terrain string
+	---@field base_terrain string
+	---@field overlay_terrain string
+	---@field info terrain_info
```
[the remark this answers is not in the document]
CelticMinstrel: I'll add this to the list of things to fix after this merges, as it's not a documentation issue.

data/lua/core/map.lua
CelticMinstrel: I documented this field, but do I need more on this point?

utils/emmylua/README.md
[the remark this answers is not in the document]
stevecotton: The first line of the file is blank, the text doesn't start until line 2.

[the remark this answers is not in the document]
CelticMinstrel: ???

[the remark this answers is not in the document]
stevecotton: Because it seems strange; is there meant to be a blank line at the start of the file?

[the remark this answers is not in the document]
CelticMinstrel: I don't think it's all that strange. I don't really remember, but I probably did that on purpose.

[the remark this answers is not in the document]
Pentarctagon: That said, if this is something that's meant to be standardized in Wesnoth's codebase, it should be done so as part of the whitespace checker script rather than one-off comments in a particular PR.

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,267 @@
...
+---@return "'Game Lua Kernel'" | "'Mapgen Lua Kernel'" | "'Application Lua Kernel'"
+function wesnoth.kernel_type() end
+
+---@param path string
+---@return ...
+function wesnoth.dofile(path, ...) end
+
+---@class tstring : string
+
+---Constructs a textdomain, which can be called to create translatable strings.
```
stevecotton, line 14: Should `to create translatable strings` be `to translate translatable strings`?

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,267 @@
...
+---@param defender unit
+---@param defender_weapon integer
+---@return stats_evaluation attacker_stats
+---@return stats_evaluation defender_stats
+---@return weapon_evaluation attacker_weapon
+---@return weapon_evaluation defender_weapon
+function wesnoth.simulate_combat(attacker, attacker_weapon, defender, defender_weapon) end
+
+---@class stats_evaluation
+---@field poisoned number
```
stevecotton, line 38: Should `poisoned`, `slowed` and `untouched` be booleans?

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,267 @@
...
+---@field slowed number
+---@field untouched number
+---@field average_hp number
+---@field hp_chance number[]
```
stevecotton, line 42: This may be correct, if I'm misunderstanding the syntax. Does this say it's an array, or a table that's acting as a sparse array?

IIUC, it uses possible HP numbers as indicies, and the probability of that exact HP as values.

CelticMinstrel: I remember griping about this awhile back, which means `table<number, number>` would probably be more correct here.

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,272 @@
...
+function wesnoth.textdomain(domain) end
+
+---Logs a message to the console
+---@param logger "'info'"|"'debug'"|"'warning'"|"'error'"|"'wml'"
+---@param message string
+---@param in_chat boolean
+---@overload fun(message:string, in_chat:boolean)
+function wesnoth.log(logger, message, in_chat) end
+
+---Simulate combat between two units_MP
```
stevecotton, line 26: The `units_MP` looks like a cut-and-paste typo - there's a local variable of that name somewhere, used for "units with movement points".

utils/emmylua/wesnoth.lua
stevecotton: Should that be `table<integer, number>`?

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,272 @@
...
+---@field plagues boolean Indicates whether the attack inflicts plague.
+---@field plague_type string If the attack inflicts plague, this specifies the unit type it is converted to.
+---@field backstabs boolean
+---@field rounds integer The number of rounds of combat.
+---@field firststrike boolean Indicates whether the attack always strikes first.
+---@field drains boolean Indicates whether the attack drains hit points.
+---@field drain_constant integer
+---@field drain_percent number
+---@field attack_num integer
+---@field name string
```
stevecotton, line 60: tstring ?

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,272 @@
...
+---@param definition string[] Sample names to be input to the generator
+---@param chain_size? integer The size of the Markov chain unit
+---@param max_length? integer Maximum length of the generated names
+---@return fun():string
+function wesnoth.name_generator(type, definition, chain_size, max_length) end
+
+---Compile a WFL formula into a Lua function
+---@param formula string A WFL formula
+---@return fun(variables:WML):any
+function wesnoth.compile_formula(formula, variables) end
```
stevecotton, line 79: I think this only takes 1 argument

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,272 @@
...
+---@field min color
+---@field max color
+---@field minimap color
+---@field pango_color string
+
+---@type table<string, color_range>
+wesnoth.colors = {}
+
+---@class event_context
+---@field name string
```
stevecotton, line 147: tstring ?

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,272 @@
...
+
+---Information on a specific terrain type
+---@class terrain_info
+---@field id string
+---@field name tstring
+---@field editor_name tstring
+---@field description tstring
+---@field icon string
+---@field editor_image string
+---@field light integer
```
stevecotton, line 256: The C++ classes have minimum / maximum light too, but this might be missing from the LuaAPI.

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,272 @@
...
+wesnoth.terrain_types = {}
+
+---@type table<string, unit_type>
+wesnoth.unit_types = {}
+
+---@generic T
+---@param values T[]
+---@param names string[]
+---@return table<string, T>
+function wesnoth.named_tuple(values, names) end
```
stevecotton, line 272: Can values be a heterogeneous array? Using `T` here suggest it's homogeneous.

utils/emmylua/wml.lua
```diff
@@ -0,0 +1,116 @@
...
+---@alias WML WMLTable|vconfig
+
+---A WML tag is a two-element array.
+---The first element is a string and the second element is a WML
+---@class WMLTag
+---@field tag string
+---@field contents WML
+
+---@type WMLTag
+local test = {"a", {}}
```
stevecotton, line 24: Intentional or leftover?

utils/emmylua/wesnoth/units.lua
```diff
@@ -0,0 +1,265 @@
+---@meta
+
+---Defines a weapon possessed by a unit
+---@class unit_weapon
+---@field read_only boolean
+---@field description tstring
+---@field name string
```
stevecotton, line 7: tstring ?

utils/emmylua/wesnoth/units.lua
```diff
@@ -0,0 +1,265 @@
...
+---Get the resistance of a unit againt an attack type
+---@param unit unit The unit to check
+---@param damage_type string The damage type
+---@param as_attacker? boolean Whether to consider the unit as the attacker or the defender
+---@param location? location Calculate as if the unit is on this location
+---@return number
+---@overload fun(unit:unit, damage_type:string, as_attacker?:boolean, x:integer, y:integer)
+function wesnoth.units.resistance_against(unit, damage_type, as_attacker, location) end
+
+---Get the movement of a unit on a particular terrain
```
stevecotton, line 177: Might be clearer as "get the movement cost for a unit on a particular terrain".

Similarly for vision and jamming.

stevecotton:
@CelticMinstrel , I've now finished reading all the files.

utils/emmylua/wesnoth.lua
CelticMinstrel: For some reason I thought the docs parser wouldn't accept that, but it seems like I was wrong.

utils/emmylua/wesnoth.lua
CelticMinstrel: I think it's probably an auto-complete issue. Sometimes the auto-complete just accepts the first suggestion without me noticing.

utils/emmylua/wesnoth.lua
CelticMinstrel: I'm pretty sure this is the `name` key from WML, which functionally is actually the ID of the attack.

utils/emmylua/wesnoth.lua
CelticMinstrel: I might've copy-pasted `eval_formula` and edited incorrectly.

utils/emmylua/wesnoth.lua
CelticMinstrel: Pretty sure this is the event name key from the WML, so not translatable.

utils/emmylua/wesnoth.lua
CelticMinstrel: That's a good point. I think _most_ of the actual uses are indeed homogenous, but the API itself doesn't care about the content of the array. I'm not sure there's a good way to represent that in the documentation though.

utils/emmylua/wesnoth/units.lua
CelticMinstrel: No, this is the name key from WML, so it's functionally the weapon ID.

utils/emmylua/wesnoth.lua
CelticMinstrel: Yeah, looks like those are not exposed in the LuaAPI. Think we should add them? (Not in this PR obviously.)

CelticMinstrel:
Okay, unless you have further comments regarding my replies and latest commits, I think this is ready to squash.

utils/emmylua/wesnoth.lua
stevecotton: OK. As for adding them, maybe review the C++ and the WML definitions and see if we want to change them first. That effort would fit in with the remaining documentation in #4417, seeing if there's a easier way to describe it than the one currently on the TerrainWML page.

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,272 @@
...
+---@field petrifies boolean Indicates whether the attack inflicts petrification.
+---@field plagues boolean Indicates whether the attack inflicts plague.
+---@field plague_type string If the attack inflicts plague, this specifies the unit type it is converted to.
+---@field backstabs boolean
+---@field rounds integer The number of rounds of combat.
+---@field firststrike boolean Indicates whether the attack always strikes first.
+---@field drains boolean Indicates whether the attack drains hit points.
+---@field drain_constant integer
+---@field drain_percent number
+---@field attack_num integer
```
stevecotton, line 59: There's a `deprecated` comment in the C++ for this one (in luaW_pushsimweapon).

utils/emmylua/wesnoth.lua
stevecotton: ok

utils/emmylua/wesnoth.lua
```diff
@@ -0,0 +1,272 @@
...
+---@field damage integer The damage the attack inflicts per hit.
+---@field chance_to_hit number The chance for the attack to hit.
+---@field poisons boolean Indicates whether the attack inflicts poison.
+---@field slows boolean Indicates whether the attack inflicts slow.
+---@field petrifies boolean Indicates whether the attack inflicts petrification.
+---@field plagues boolean Indicates whether the attack inflicts plague.
+---@field plague_type string If the attack inflicts plague, this specifies the unit type it is converted to.
+---@field backstabs boolean
+---@field rounds integer The number of rounds of combat.
+---@field firststrike boolean Indicates whether the attack always strikes first.
```
stevecotton, line 55: I'm not sure, but suspect many of the booleans here are whether the ability is active in the particular fight (whether the opponent is immune to poison, also has firststrike, etc).

utils/emmylua/wesnoth.lua
stevecotton: A quick search for `emmylua heterogeneous` suggests no-one's asked that yet, so accepted.

utils/emmylua/wesnoth/units.lua
stevecotton: Thanks.

stevecotton (approved):

utils/emmylua/wesnoth.lua
CelticMinstrel: Thanks for spotting that.

stevecotton (approved):

PR #6484, opened 2022-02-03
SoF: fix thursagan's arcanister advancement breaking in some cases

knyghtmare:
Fix for the following bug:
image

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,8 @@
  ### Campaigns
    * Delfador’s Memoirs
      * S14: Smoother appearance of enemies and added dialogue (#6176)
+   * Sceptre of Fire
+     * Fixed Thursagan's Arcanister advancement being "lost" after dismounting a minecart. 
    * The Rise of Wesnoth
      * S22: Fixed the possibility of a missplaced dialogue when a bridge was broken (issue #6376)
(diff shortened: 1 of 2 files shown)


stevecotton:
I'm not sure about this, but think all of `[modifications]` can be overwritten on unit rebuild. So I'm not sure about this fix.

Pentarctagon:
What would be a specific situation to test this with?

knyghtmare:
> What would be a specific situation to test this with?

As discerned on Discord, in _Gathering Materials_, get Thursagan into a cart and dismount.

Expected:
1. After dismount, max XP is 210 and blue XP bar

Bug:
2. 150 XP and purple XP bar

Pentarctagon:
And you retested that situation after applying this fix, and the bug no longer occurred?

knyghtmare:
> And you retested that situation after applying this fix, and the bug no longer occurred?

yes.

knyghtmare:
If this patches the bug, should it be merged?

stevecotton:
Looks good to me.

stevecotton:
Based on the discussion of who-should-merge-what yesterday, I should. Sorry for taking so long.

@knyghtmare please will you backport it?

PR #6488, opened 2022-02-06
1.16 Various terrain fixes

doofus-01:
some terrain graphics fixes, no new content.


doofus-01:
The craters thing allows this, with no new terrain codes
Screenshot_20220205crater

Pentarctagon:
What would happen if someone using 1.16.0-1.16.2 tried to load a map that used the new crater terrains?

doofus-01:
If map was drawn with the map editor, it would be no different.  If they had their own rules and graphics, it could possibly display differently, but would not cause a crash or "no-show".  Is there a way to grep the server for terrain-graphics defining some non-mainline `*^Dc`?

Pentarctagon:
@soliton- can, I believe.

stevecotton:
I just tested (1.16.2+dev based on 2d20b72a9e430f7), and it refused with "The game map could not be loaded: Unknown tile in map: (Gg^Dc) 'Gg^Dc'". So this has to be 1.17 only.

doofus-01:
Hrm, I must have forgotten to hit F5 or something.  This is a bug, but if it can't be fixed, it can't be fixed.  Oh well.

Issue #3059, opened 2018-05-09
Error message (possibly invalid WML) while updating Add-ons


Konrad22:
While updating my Add-ons, this happened:
![wml](https://user-images.githubusercontent.com/39015360/39801718-6f6ca0c0-536c-11e8-89ee-2fd52648da2d.png)
Wesnoth did not crash afterwards, I was just returned to the main screen.
I had an Add-on description open, when I used the 'Update all' button. (It's reproducible this way.)
And it apparently finished updating.

Konrad22:
(My case is the small(er) window.)

jyrkive:
@Konrad22 mentioned in Discord that it also occurs if he presses a key while having addon description open.

My first guess would be that the search field attempts to do something while the addon list is hidden.

Konrad22:
Small update: Enter or Esc do not trigger this. But things like alt gr + q for special symbols do trigger it.

CelticMinstrel:
Based on the error message, I think that @jyrkive is almost certainly correct in his assessment of the cause.

Possible fixes:

* Treat it as optional; if it's found to be hidden (`find_widget` returns null), don't update anything, just note that it needs to be updated when it's shown later.
* Ensure that the list is visible before working with it; then if it was originally hidden, hide it again when done.

Konrad22:
Why you this needs to be fixed beofre 1.16:
This error message will appear (and close the add-on manager) every single time the user tries to select an older version of an add-on in the add-on description.
(Why selecting an older version is possible: https://github.com/wesnoth/wesnoth/pull/5038)

CelticMinstrel:
Huh, is this still not fixed? I don't think it would be all that hard… might even be a good first issue for someone…

stevecotton:
There are multiple paths that result in the same error message. I've just posted a fix to one specific path in #5810, but I'm wondering if calling stacked_widget::set_find_in_all_layers(true) would be a more general solution.

stevecotton:
As pointed out on IRC, this is going to be much easier to trigger in 1.16, and IMO it's something that should be a blocker.

stevecotton:
Two separate causes of this were fixed in #5820, I think it makes sense to close this issue (and open a new one if there's another way to cause the issue).

PR #6490, opened 2022-02-06
Add-ons manager: handle the selected add-on changing in small-screen mode

### LINKED ISSUE #6485: WML error when pressing key Up/Down in add-on details
Generate the error by

- connect to add-on server
- select any add-on in the list
- press "Add-on Details" button
- press arrow key Up/Down[/code]

The error is:

```
error general: An error due to possibly invalid WML occurred
The error message is :
Mandatory widget 'addons' hasn't been defined.
 
When reporting the bug please include the following error message :
Condition '!must_exist || result' failed at /opt/wesnoth/wesnoth.git.master/src/gui/auxiliary/find_widget.hpp:75 in function 'find_widget'.
```
Same behavior on BFW 1.16.1.

# System settings

The Battle for Wesnoth version 1.17.0+dev x86_64
Running on Debian GNU/Linux 8 (jessie) x86_64
Distribution channel: Default

Game paths
==========

Data dir:        /opt/wesnoth/wesnoth.git.master
User config dir: /home/USER/.config/wesnoth
User data dir:   /home/USER/.local/share/wesnoth/1.17
Saves dir:       /home/USER/.local/share/wesnoth/1.17/saves
Add-ons dir:     /home/USER/.local/share/wesnoth/1.17/data/add-ons
Cache dir:       /home/USER/.cache/wesnoth

Libraries
=========

Boost:             1.68
Lua:               5.4.2
OpenSSL/libcrypto: 1.0.2l (runtime 1.0.2l)
Cairo:             1.14.0 (runtime 1.14.0)
Pango:             1.36.8 (runtime 1.36.8)
SDL:               2.0.9 (runtime 2.0.9)
SDL_image:         2.0.4 (runtime 2.0.4)
SDL_mixer:         2.0.0 (runtime 2.0.0)

Features
========

Lua console completion:       yes
D-Bus notifications back end: yes

Current video settings
======================

SDL video drivers:       [x11] dummy
Window size:             1366x702
Screen refresh rate:     60
Screen dots per inch:    112.286079x112.758377
Screen dpi scale factor: 1.169647x1.174566

stevecotton:
Fix another instance of issue #3059, where code assumes that all parts of the
add-ons manager's UI are accessible, an assumption that fails when the dialog
uses a stacked widget to handle small window sizes.

The up and down arrow keys select the previous or next add-on in the list,
even when the small-window layout is hiding the list. That feels like a feature
rather than a bug, as it's useful and an understandable UX; however it needs
the fix in this commit so that on_addon_select() doesn't throw an exception
and close the dialog. The new code is the same as the fix that 050feb623b
applied to on_selected_version_change, traversing the stacked widget to get the
info.

Fixes #6485.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 ## Version 1.16.2+dev
  ### Add-ons client
+   * Fixed: using the up or down arrow keys in small-screen mode returned to the title screen (issue #6485)
  ### Add-ons server
  ### Campaigns
    * Delfador’s Memoirs
--- a/src/gui/dialogs/addon/manager.cpp
+++ b/src/gui/dialogs/addon/manager.cpp
@@ -688,7 +688,8 @@ void addon_manager::apply_filters()
 	// In the small-screen layout, the text_box for the filter keeps keyboard focus even when the
 	// details panel is visible, which means this can be called when the list isn't visible. That
 	// causes problems both because find_widget can throw exceptions, but also because changing the
-	// filters can trigger on_addon_select and thus affect which add-on's details are shown.
+	// filters can hide the currently-shown add-on, triggering a different one to be selected in a
+	// way that would seem random unless the user realised that they were typing into a filter box.
 	//
 	// Quick workaround is to not process the new filter if the list isn't visible.
 	auto list = find_widget<addon_list>(get_window(), "addons", false, false);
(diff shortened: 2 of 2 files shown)


Wedge009 (approved):
Confirm issue resolution. I like that the details page updates along with the change of add-on selection in the background, as it would in the 'large screen' layout.

Will you be forward-porting on merge?

stevecotton:
Thanks, and I'm forward-porting it now.

PR #6491, opened 2022-02-06
fix bad Swamp-Icy_Cobble (Ss-Rra) transition

doofus-01:


stevecotton:
Before and after, looks good to me:
i6491_before_and_after

stevecotton (approved):

PR #6493, opened 2022-02-07
WoF and Drake revisions

nemaara:
Like the title says, cleans up some references in WoF and revises the mainline drake race description. This will not handle a more detailed text revision of WoF, only lore references.


knyghtmare (commented):
I am assuming the "Avalon" references were removed since they related to the paradise mentioned in the "King Arthur and the Knights of the Round Table".

"Sons of Drugis"...liked that name though...usable in UMC...

nemaara:
Yea mainline already has a policy of avoiding direct references to existing mythology/fantasy.

Jonathan-Kelly:
@nemaara
Regarding "Avalon" are we sure the policy was not just against [copyrighted material](https://forums.wesnoth.org/viewtopic.php?t=34904#w0fpi9)? Policies aside though, what is your own feeling about dropping hints in the lore which suggest Irdya and Earth may exist in the same universe, with some magical connections between them (and other) worlds? (Maybe even hinting that what the player does in Irdya could have repercussions for the real world?)

Khrakrahs and Shek'kahan's are siblings [according to SoF](https://github.com/wesnoth/wesnoth/blob/c4a7fce80a7ed3fdb2767fd2604440ce917f142a/data/campaigns/Sceptre_of_Fire/scenarios/8_The_Dragon.cfg#L400) so if you want to change this about them (?) then it might be best to adjust the line from SoF too. Also, there is another mention of this [earlier in S8](https://github.com/wesnoth/wesnoth/blob/master/data/campaigns/Winds_of_Fate/scenarios/08_Overlook.cfg#L562) (you can squash all these changes into the same commit to keep things tidy, if you do so before merging.)

@nemaara + @irydacea
Would Yechnagoth work here instead of Zhangor? Asking this because, without any connection to the crazy-evil antagonists of UtBS, I am not sure if the elvish lady's character motivations work, not even in this campy scenario. From the player's perspective, she will just be a serial-killer-politician-elf. Which may be a bit _too_ random. If removing the UtBS demons connection is the way it should be, then I think I may need to follow up with a rewrite or adjustment of this character at some point.

nemaara:
Mainline not using real world names or references has been a thing for a while. It's similar to why Lady Jessica was changed to Jessene and the whole spiel around the Dunefolk. Hinting at Irdya and Earth being in the same universe should have a very good reason behind that, which probably requires a significant amount of thought. Unless there is something you're thinking of in terms of that being a meaningful theme, I would say it's better to not have that.

Yes I want to change the dragons being brothers in SoF too, thanks for catching.

I would say it's a bit early for Yechnagoth to show up personally, but not entirely opposed. Depends on what Iris thinks as well. I did have a reason for choosing the name Vanar, but it's more of a infodump for PMs.

nemaara:
Merging in a couple days, going to try to get this in for the next release.

irydacea:
The way I envisioned things in my canon — which was also the plan for UtBS when I became its maintainer, mind you — Yechnagoth would be nothing but THE most obscure figure of myth during the mainline epoch (TRoW till the Great Fall) and known to absolutely nobody but people prying into sources of information on the previous world. She's certainly not meant to be an entity with any kind of agency or power because she's just functionally dead up to some point after the Fall. Zhangor has simply not come into existence yet, and neither has "Uria" — although I'd certainly not be opposed to there being a number of past pretenders to the name of "Uria" since it's more like a title.

nemaara:
Okay it's staying as Lord Vanar for now and we will revisit later if needed.

data/campaigns/Winds_of_Fate/scenarios/07_Harvest.cfg
```diff
@@ -487,7 +486,7 @@ My diary is very personal, Vonel!"
             speaker=narrator
             message= _ "31st day of Leaffall
 
-...those saurians may smell of the swamp they crawl from, but they shall make a worthy sacrifice to Zhangor! I have extended a formal invitation to Arinexis, a saurian ambassador, or what passes for one amongst their primitive and malodorous kind. I shall show this saurian and its entourage every courtesy and tempt them with all the delights of elven cuisine— laced with a sleeping potion of my own design..."
+...those saurians may smell of the swamp they crawl from, but they shall make a worthy sacrifice to Vanar! I have extended a formal invitation to Arinexis, a saurian ambassador, or what passes for one amongst their primitive and malodorous kind. I shall show this saurian and its entourage every courtesy and tempt them with all the delights of elven cuisine..."
```
CelticMinstrel, line 489: I don't actually care one way or another, but I'm curious why you chose to remove the sleeping potion reference.

data/campaigns/Winds_of_Fate/scenarios/09_Ancestor.cfg
```diff
@@ -719,7 +719,7 @@ I am elder of Green Swamp Clan and all my years Shek'kahan protected usss. I cou
             speaker=Khrakrahs
             image_pos=right
             mirror=yes
-            message= _ "It was nigh every season he would intrude upon my domain or me his. Each time, there was such a glorious battle between we, the Sons of Drugis, as to impart even the clouds with fright. I must not have challenged him enough that he grew weak, vulnerable. Now another great battle between us there will never be. Now there is only me."
+            message= _ "It was nigh every season he would intrude upon my domain or me his. Each time, there was such a glorious battle between we, as to impart even the clouds with fright. I must not have challenged him enough that he grew weak, vulnerable. Now another great battle between us there will never be. Now there is only me."
```
CelticMinstrel, line 722: This reads fine with the appositive "Sons of Drugis" but when you remove it, it sounds strange. Shouldn't it just be "us"? Possibly also add "so"?

…between us, so as to impart…

CelticMinstrel:
I think I like "Khatlak" better than "Avalon" anyway. The name "Avalon" just doesn't suit a fire guardian.

data/campaigns/Winds_of_Fate/scenarios/07_Harvest.cfg
nemaara: It's a bit heavy handed.

Jonathan-Kelly:
> I did have a reason for choosing the name Vanar, but it's more of a infodump for PMs.

I would be interested in taking a look at it.

Do you also have lore in mind for the name "Khatlak"? If it is just a placeholder, I would like to keep things simple for the player with the following adjustments:

_"~~Khatlak~~ **It** has come."_     (For both times this appears in S1)

_"Soon I shall peer into the far side of our Faerie Realm, ~~the side beyond Khatlak,~~ which that pretender Lady Dionli claims she fears to descry."_

data/campaigns/Winds_of_Fate/scenarios/07_Harvest.cfg
Jonathan-Kelly: Without the sleeping potion, it is unclear how she plans to capture the saurians and drakes though, no? Since she never required them to disarm and some of the saurians and drakes still fight quite well without using weaponry at all.

(This was also why she tried a couple times to get her 'guests' to eat somethings.)

CelticMinstrel: In other words, the intention is that although she _did_ use a sleeping potion, she's not coming right out and saying it… right?

nemaara: You can probably infer that she did something to the food and drink though exactly what she did you can just guess. Maybe it was a sleeping potion or maybe something else, presumably not very good for her guests though.

stevecotton:
Should the SoF change (the dragons being rivals rather than brothers) be backported? No reason why it should, I just noticed it while sanity-checking the forwardport of #6617.

nemaara:
I guess there's no need.

PR #6498, opened 2022-02-08
Replace deprecated share_maps and share_view with share_vision

### LINKED ISSUE #6496: [LoW] [deprecated code] replace share_view=yes in favour of share_vision=all
https://github.com/wesnoth/wesnoth/blob/f5ea3b0906ff5bb3f1282a33bc3419c11e1cb105/data/campaigns/Legend_of_Wesmere/utils/characters.cfg#L18

[Wiki](https://wiki.wesnoth.org/SideWML)
```
share_view: whether sides allied with this side see the units that this side sees, if they are on FoW (fog). ([Version 1.13.1 and later only](https://wiki.wesnoth.org/DevFeature)) This has been deprecated in favor of share_vision (see below).
```
Since it's been deprecated around 1.13.1, it should be updated to `share_vision=all`

Comment:
I am surprised this deprecated code has stuck so long and undetected.

Wedge009:
Closes #6496.

Not sure if I'm understanding the [documentation](https://wiki.wesnoth.org/SideWML) correctly but I interpreted `share_maps=yes` as `share_vision=shroud` and `share_view=yes` as `share_vision=all`.

Also not sure if it's worth back-porting, since there's the potential to break compatibility for games in progress. But if reviewers think it's okay, I can back-port too.


Wedge009:
Hmm, where are those ...share_view tests defined? Did it break just because I renamed the tests to match the updated code?

knyghtmare (commented):
Were all the edited scenarios checked to see if the update works as intended? Otherwise we might be preparing for a flood of bugs.

knyghtmare:
Also, it goes like this:
if share_maps (it means shroud data), so just share_vision=shroud
if share_view (it means fog), so share_vision=all
Share_vision has no specific value for "just FoW" so in those cases it should be set for "all".

In any case, the edits look fine.

Wedge009:
> Were all the edited scenarios checked to see if the update works as intended?

I checked the test scenario and a few of the campaign scenarios and didn't see anything obviously wrong but I'd still prefer for others to check this as I'm not entirely sure of 'correct' behaviour.

> Also, it goes like this: if share_maps (it means shroud data), so just share_vision=shroud if share_view (it means fog), so share_vision=all Share_vision has no specific value for "just FoW" so in those cases it should be set for "all".

That's basically how I read it too.

data/campaigns/Legend_of_Wesmere/scenarios/chapter1/03_Kalian_under_Attack.cfg
```diff
@@ -180,7 +180,7 @@
         user_team_name= _ "Enemies"
         fog=yes
         shroud=yes
-        share_view=yes
+        share_vision=all
```
stevecotton, line 183: Just a comment about the scenario, not an issue with the current PR.

It seems strange that there are settings for sharing vision between the AI sides. I can only assume that it's either a copy-and-paste from the human teams, or that there was an idea of having a human-playing-elves vs human-playing-orcs match.

Wedge009: I noticed that as well, but I thought it was to do with how the AI operates. No idea if they are normally able to share information without this explicit indicator, though I imagine they probably do.

stevecotton: Just found that the AI is more complex than I thought - some of the candidate actions do check if units are visible. I incorrectly thought it ignored fog of war completely.

No idea if that makes a different on this scenario.

stevecotton:
All looks good to me. I haven't tested in-game.

Wedge009:
Should I wait till someone else does some testing before merging?

data/campaigns/Delfadors_Memoirs/scenarios/11_Wasteland.cfg
```diff
@@ -257,7 +257,7 @@
                 [/set_recruit]
                 [modify_side]
                     side=3
-                    share_maps=yes
+                    share_vision=shroud
```
stevecotton, line 260: Maybe add a comment `# Delfador's side has fog=no, so share_vision=shroud is equivalent to share_vision=all`?

stevecotton, line 260: I've now done some testing, on both the changed version and on the current 1.16 (without this PR). The two lines are equivalent, neither has any effect.

Chantal's side has shroud=no, which I assume is why `share_vision=shroud` doesn't clear any shroud for Delfador. To share her vision in this event (but not before), her `[side]` needs both of these:
```
        shroud=yes
        share_vision=none
```

My feeling on the LoW parts of this PR is that it would be good to update the deprecated line, but leave a note for @nemaara's rewrite that it doesn't have the intended effect.

stevecotton:
I've tested all the campaign changes except for the LoW ones. The LoW ones look right, but as it's getting rewritten anyway I didn't see the need to test.

My feeling is to merge, but as it's campaign content I think the decision has to be made by @nemaara rather than me.

nemaara (approved):
Upcoming rework has no bearing on making improvements to the current version.

stevecotton:
@Wedge009 I've added a few comments about possibly adding comments, but feel free to merge as-is or with those as you feel appropriate.

data/campaigns/Delfadors_Memoirs/scenarios/11_Wasteland.cfg
Wedge009: With the editing, I'm not sure what you're wanting here. Should I be correcting something? Since this is a master-based change, I think it's fine to include corrections (as opposed to just ensuring the behaviour between the old and new syntax matches exactly).

I'll hold off on merging until this is clarified and resolved.

Wedge009: Okay, I think I understand what you mean: it appears the intention is to share Chantal's side's vision with the player after the event occurs, but it does not. I verified this for both before and after this PR's changes. I'll correct this with your suggestion.

Wedge009:
@stevecotton Thanks for your review. If you can just double-check the second commit relating to the sharing of Chantal's side's vision and advise whether or not it's okay, then I'll merge or correct accordingly.

data/campaigns/Delfadors_Memoirs/scenarios/11_Wasteland.cfg
stevecotton: I logged it as #6521, think it's probably worth fixing in 1.16 too.

stevecotton (approved):

stevecotton:
I think it makes sense to backport all of this. Now we've tested DM S11 on 1.16, I'm confident that these changes won't introduce any bugs.

Wedge009:
Back-ported with 80c92630039ab5a3c9d2ee54522198050e2defe1 + c80f83da947079b03e990f59f802a68d57f71f1b + 28ce494474e557c46844e5ff184db70b9a64aac8.

PR #6501, opened 2022-02-09
Change [filter_weapon] to not depend on attack_type::weapon_specials

stevecotton:
This changes it to use the function that attack_type::get_value already uses,
thus removing a call to attack_type::weapon_specials. The latter is being
refactored by @newfrenchy83, and questions about the effect of that refactor on
this code delayed the review on his PR.

This code is part of the object used for [filter_weapon]formula=, however the
implementation of that filter creates exactly one attack_type_callable, so
doesn't trigger this code. For testing I called it via a debugger.

### DIFF
--- a/src/formula/callable_objects.cpp
+++ b/src/formula/callable_objects.cpp
@@ -162,7 +162,20 @@ int attack_type_callable::do_compare(const formula_callable* callable) const
 		return att_->range().compare(att_callable->att_->range());
 	}
 
-	return att_->weapon_specials().compare(att_callable->att_->weapon_specials());
+	const auto self_specials = att_->specials().all_children_range();
+	const auto other_specials = att_callable->att_->specials().all_children_range();
+	if(self_specials.size() != other_specials.size()) {
+		return self_specials.size() < other_specials.size() ? -1 : 1;
+	}
+	for(std::size_t i = 0; i < self_specials.size(); ++i) {
+		const auto& s = self_specials[i].cfg["id"];
+		const auto& o = other_specials[i].cfg["id"];
+		if(s != o) {
+			return s.str().compare(o.str());
+		}
+	}
+
+	return 0;
 }
 
 unit_callable::unit_callable(const unit& u) : loc_(u.get_location()), u_(u)


stevecotton:
@newfrenchy83 sorry it's taking so long to review that PR, but it hasn't been forgotten.

I'm expecting the CI build to fail on flatpak, which is waiting for a new version of an SDK to be uploaded to dockerhub.

newfrenchy83:
@stevecotton i commited https://github.com/wesnoth/wesnoth/pull/6462/commits/27429f5e9535a8fe95a0e4e4c9a2c8642308084a for resolve that.

newfrenchy83:
> Thanks, but my motivation for writing this was to get rid of the `only_active` argument from attack_type::weapon_specials; I think this is the only place that called it with `only_active` set to `false`.

when this PR will be merged, i remove only_active argument in my PR

src/formula/callable_objects.cpp
```diff
@@ -162,7 +162,20 @@ int attack_type_callable::do_compare(const formula_callable* callable) const
...
+	const auto self_specials = att_->specials().all_children_range();
+	const auto other_specials = att_callable->att_->specials().all_children_range();
+	if(self_specials.size() != other_specials.size()) {
+		return self_specials.size() < other_specials.size() ? -1 : 1;
+	}
+	for(std::size_t i = 0; i < self_specials.size(); ++i) {
+		const auto& s = self_specials[i].cfg["id"];
+		const auto& o = other_specials[i].cfg["id"];
+		if(s != o) {
+			return s.str() < o.str() ? -1 : 1;
```
CelticMinstrel, line 174: Any reason not to just `return s.compare(o)`?

CelticMinstrel:
This seems fine to me.

src/formula/callable_objects.cpp
stevecotton: Will change to `s.str().compare(o.str())`, thanks. They're both config_attribute_value objects, so there isn't a direct compare(), but I forgot to look again for a compare() function after realising that I needed to convert them to strings.

CelticMinstrel: I realize this is a bit late now that it's merged, but you could also do:
```cpp
auto result = s.str().compare(o.str());
if(result != 0) return result;
```

stevecotton: Good idea, I'll remember that next time.

PR #6506, opened 2022-02-10
Add test scenario for teacher-based weapon specials

stevecotton:
This is intended for manual testing, the reason for writing it is to
test tooltips in the attack-predictions dialog. It has a new map, with
five sides packed into adjacent hexes.

Everyone gets a self+allies leadership based on their side number, with
their name in the ability's id and ability's name for tracking whose buffs
are affecting who.

I'm expecting to do more changes to it before considering whether it should
be merged as a standalone test, added to data/scenario-test.cfg, or discarded.

[ci skip]


stevecotton:
Closing because the ideas from this have been reworked, for example some of them became PR #6627.

PR #6507, opened 2022-02-10
remove only_active argument to weapon_specials()

newfrenchy83:
weapon_specials only called in unit_attack.cpp, only_active argument don't usable

### DIFF
--- a/src/gui/dialogs/unit_attack.cpp
+++ b/src/gui/dialogs/unit_attack.cpp
@@ -115,9 +115,9 @@ void unit_attack::pre_show(window& window)
 			attacker_itor_->get_location(), false, attacker.weapon
 		);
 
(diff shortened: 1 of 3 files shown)


stevecotton:
There's some documentation on that function that should be updated too.

If we merge this now, would you be happy to merge this exact change onto 1.17 too and then rebase the bigger PR on top? Otherwise I'd prefer to only merge things that can be applied to both 1.16 and 1.17.

newfrenchy83:
merge change into 1.16/1.17 and rebase big PR. i will update documentation.

newfrenchy83:
Documentation corrected

stevecotton:
Oops, I forgot to squash those into a single commit.

PR #6512, opened 2022-02-12
[1.16] Warning when share_view or share_maps overrides share_vision

stevecotton:
Add a new type of deprecation warning, used when a deprecated WML attribute is used alongside its replacement. Use that to log a deprecation warning if a [side] has share_vision overridden by share_view or share_maps.

While share_view and share_maps are deprecated on the Wiki, they don't trigger deprecation warnings on their own. I intend to make them trigger warnings in 1.17, but they map cleanly to the share_vision enum so the wrapper is easy to support indefinitely.

All SP uses of the deprecated attributes were removed from mainline in f983a998a2 on the 1.17 branch, which was backported to 1.16 as 80c9263003.

On 1.16, WC still has one [side]share_view=, in WC's side_definitions.lua; this has been replaced in 35aa0d1893 on the master branch. It won't trigger this warning, because it isn't used in the same tag as [side]share_vision=.

The branch name is no longer accurate, as some feature-creep has been removed.

### DIFF
--- a/src/deprecation.hpp
+++ b/src/deprecation.hpp
@@ -21,7 +21,8 @@ enum class DEP_LEVEL : uint8_t { INDEFINITE = 1, PREEMPTIVE, FOR_REMOVAL, REMOVE
 
 /**
  * Prints a message to the deprecation log domain informing players that a given feature
- * has been deprecated.
+ * has been deprecated. May show the error in the in-game chat area, but has a mechanism
+ * to avoid repeatedly showing in-game messages to players who can't fix the issue.
  *
  * @param elem_name    The name of the feature to be deprecated.
  * @param level        The deprecation level. This indicates how long the feature will
@@ -43,3 +44,17 @@ std::string deprecated_message(const std::string& elem_name,
 		DEP_LEVEL level,
 		const class version_info& version,
 		const std::string& detail = "");
+
+/**
+ * Prints a message to the deprecation log domain informing players that a deprecated
+ * feature has been used along with a new feature, and that the compatibility code
+ * doesn't support that combination.
+ *
+ * For example, if a WML tag contains both a deprecated attribute and its replacement,
+ * the compatibilty code might use one, ignore the other, and log a message.
+ *
+ * @param elem_name    The name of the feature that's deprecated.
+ * @param modern_name  The new feature that can't be used with the deprecated one.
+ */
+void unsupported_combo_message(const std::string& elem_name,
+		const std::string& modern_name);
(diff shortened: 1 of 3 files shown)


src/team.cpp
```diff
@@ -255,11 +256,6 @@ void team::team_info::read(const config& cfg)
 	// by default, persistence of a team is set depending on the controller
 	persistent = cfg["persistent"].to_bool(this->controller == CONTROLLER::HUMAN);
 
-	//========================================================
-	// END OF MESSY CODE
```
stevecotton, line 259: The "start of messy code" comment was removed in 09223f03cc.

src/config.hpp
```diff
@@ -490,13 +490,27 @@ class config
 	/**
 	 * Get a deprecated attribute without a direct substitute,
 	 * and log a deprecation message
+	 * @deprecated Use the overload that takes a version_info instead.
```
stevecotton, line 493: This is currently only used in some WFL code that's marked FOR_REMOVAL in 1.17, so this can be removed with that code.

stevecotton:
The build failures are from Doxygen saying that the overloaded function resolves ambiguously. The C++ compilers are happy with it, but I'll check Doxygen too next time I'm ready to upload this.

src/deprecation.hpp
```diff
@@ -21,7 +21,8 @@ enum class DEP_LEVEL : uint8_t { INDEFINITE = 1, PREEMPTIVE, FOR_REMOVAL, REMOVE
 
 /**
  * Prints a message to the deprecation log domain informing players that a given feature
- * has been deprecated.
+ * has been deprecated. May show the error in the in-game chat area, but has a mechanism
+ * to avoid repeatedly showing in-game messages to players who can't fix the issue.
```
stevecotton, line 25: I need to check this statement - I thought it was true, but I don't see the mechanism here, maybe it's only implemented in the Lua layers. That would explain the large number of repeated deprecation warnings about the SPECIAL_NOTES macro.

stevecotton:
This is suffering from feature-creep. I'm going to reduce the 1.16 changes to just the warning specifically for mixing deprecated and non-deprecated attributes; the rest will be 1.17 only.

src/deprecation.hpp
CelticMinstrel: I think the mechanism in question is part of `lg::wml_error`.

stevecotton: Excellent, that's definitely being used hee.

stevecotton:
The feature creep has been removed, but the branch name is no longer accurate.

stevecotton:
Might bring this back for 1.19, but the main idea of a new combo-deprecation warning would be a new string, it's late for 1.17, and definitely too late for 1.16.

PR #6515, opened 2022-02-12
world conquest: fix bezoar artifact being bugged

### LINKED ISSUE #6513: [world conquest] [artifacts] Bezoar is broken
- 1.17.0+dev / MingW64 / Windows10 / Github

### Describe the bug
Artifact is bugged. 

### To Reproduce
Equip Bezoar, see side bar, traits undead/healthy/fearless are not added as they should have been.

### Expected behavior
Not Bugged...I guess?

### Additional context
`[artifact][/artifact]` does not support `[trait][/trait]` but supports `[effect][/effect]`

knyghtmare:
Closes #6513


stevecotton:
I tested this, but it seems the unit only gets the `tenacity` ability, `unpoisonable`, `unplagueable` and 2 extra max hp. The `healthy` didn't appear (nor did the expected extra HP), nor `fearless`. The unit's current HP also didn't increase to match the extra max hp.

Please add Hejnewar as a reviewer, when it's ready.

changelog.md
```diff
@@ -10,7 +10,8 @@
    * Under the Burning Suns
      * S08: Kromph and the Dust devil now trigger a critical event (issue #6211)
    * World Conquest
-     * Fixed artifacts "draining" HP on pick-up.  
+     * Fixed artifacts "draining" HP on pick-up.
+     * Fixed Bezoar artifact being bugged and broken.
```
stevecotton, line 14: Trailing whitespace on the new line, which was on the old line in the old version.

The reason I'm commenting on this is that @Wedge009 is doing a trailing-whitespace cleanup in #6498, which I think is going to merge first.

knyghtmare: what is this "trailing whitespace" you mention? How do I patch for it here?

Wedge009: In your text editor, toggle visibility for non-printing characters. In your line 14 there's two additional spaces unnecessarily added after the full-stop - it would be best to remove that in your commit, please.

knyghtmare: Ah, I did the `Editor.renderWhitespace : "all",` option in `settings.json` and got to see 4 hidden whitespaces. Interesting.

knyghtmare:
squash merge when testing passes please.

Wedge009:
I'm not familiar with lua, but it appears the checks don't like the empty if-then statement in line 120. Is there some sort of no-op that can be placed there? Or perhaps should the condition be negated so that the `add_modification` occurs in the if statement rather than the else?

stevecotton:
> squash merge when testing passes please.

Thanks for the reminder, but you can squash your branch and then force-push to make the PR have a single commit again. That's preferable because it means you write the commit message; otherwise the commit message becomes an (editable in Github's web UI) mess of all the messages from all of the commits.

Pentarctagon:
@knyghtmare other than the `luacheck` warning, it sounds like this is good to merge?

knyghtmare:
> @knyghtmare other than the `luacheck` warning, it sounds like this is good to merge?

Yeah, I suppose it can. It fixes the WC artifact "Bezoar" by re-introducing the code which added the trait(s). Is the Luacheck warning something that should be ignored for this one?

Pentarctagon:
It causes the CI to fail, so it needs to be addressed. Though it should be as simple as changing
```
if unit:matches { wml.tag.filter_wml { wml.tag.modifications { wml.tag.trait { id = trait.id } } } } then
else
    unit:add_modification("trait", trait)
end
```
to
```
if not unit:matches { wml.tag.filter_wml { wml.tag.modifications { wml.tag.trait { id = trait.id } } } } then
    unit:add_modification("trait", trait)
end
```
?

Pentarctagon:
Thanks, will merge as soon as the CI passes.

PR #6519, opened 2022-02-13
1.16 [micro_ai] bottleneck ai

doofus-01:
This addresses an error (I'll update if I get a simple way to reproduce)
```error scripting/lua: ai/lua/ai_helper.lua:314: attempt to index a nil value (local 'dst')
stack traceback:
        ai/lua/ai_helper.lua:314: in field 'robust_move_and_attack'
        ai/micro_ais/cas/ca_bottleneck_move.lua:480: in field 'execution'
        [string "local self, params, data, filter_own = ......"]:2: in main chunk
```

I'm not sure this is a proper fix, but it seems to work, so I'll park it here for now.


knyghtmare:
What "bug" is this PR patching?

cooljeanius:
> What "bug" is this PR patching?

#6599 apparently?

doofus-01:
@mattsc - No issues on my end if you want to knock it out, I won't be doing anything here for the next few days.  Thanks.

PR #6520, opened 2022-02-15
DM S11: Correct vision sharing when Delfador finds Chantal

Wedge009:
Partial back-port of #6498 (second and final commit in that PR, as at time of writing). Originally @stevecotton's findings.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
  ### Add-ons server
  ### Campaigns
    * Delfador’s Memoirs
+     * S11: Finding Chantal will now share her side’s vision with the player, as originally intended.
      * S14: Smoother appearance of enemies and added dialogue (#6176)
    * The Rise of Wesnoth
      * S22: Fixed the possibility of a missplaced dialogue when a bridge was broken (issue #6376)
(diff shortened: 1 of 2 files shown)


stevecotton:
Ah, you got the fix in before I logged #6521, excellent. I'm not sure if it needs any rebalancing after fixing the vision.

I think this should use `share_vision`, and replace the `share_view` line too. This scenario has only instance of `share_view` or `share_maps` in the entire campaign, and `share_vision` has been supported since 1.13.2.

Wedge009:
As I explained in originally submitting #6498, I wasn't sure if switching to `share_vision` could potentially break backwards compatibility and therefore back-porting should be avoided. I would be happy to back-port all of #6498 to keep things consistent if you think it's sensible and no-one else disagrees with it. If so, that would subsume this PR.

Wedge009:
In response to your [comment in #6498](https://github.com/wesnoth/wesnoth/pull/6498#issuecomment-1040204814), now that I've finished re-reviewing all the changes made there I'm inclined to agree. Of the scenarios where the sharing is altered by an event, only DM S11 lacked the initial definition of `shroud` and `share_view`, so I don't think there are similar bugs in the other scenarios.

PR #6527, opened 2022-02-16
Upgrade to lua 5.4.4

Vultraz:


Pentarctagon:
For the static_cast, there's already https://github.com/wesnoth/wesnoth/blob/master/src/CMakeLists.txt#L99 for example - can that not be applied here for some reason?

Vultraz:
That would probably work. Didn't realize that had been added.

Pentarctagon:
The CI failures are related to doxygen warnings in heal.cpp. Which I guess aren't wrong, but also I have no idea why doxygen is complaining about them in this PR of all things.

stevecotton:
Do we absolutely need to update Lua now; could we revert this until the Doxygen bug is fixed?

FWIW, the exact chunk of Lua that seems to be triggering the bug is isolated in #6528, but it's not something that could simply be removed.

Issue #5084, opened 2020-08-16
The future of deprecation messages …


sevu:
How to handle deprecation messages in the future?
- 1.12 displayed them ingame to everybody playing it
- 1.14 hides them from everybody, except of the crazy people who start wesnoth from a terminal, or find the logs on Windows.

The old way was annoying, but it had better chances that someone gets involved and fixes it.
The current one somehow undermines the deprecation policy, as things are deprecated properly, but people won't fix it.

Maybe it's possible to display the deprecation messages dependent on whether the player has a _server.pbl in the add-on, or lacks an _info.cfg?

Current situation is, that as soon as I have other peoples add-ons installed, I get more deprecation messages than my terminal displays (I did set the buffer to 1000 lines).

In case we don't want to change anything about how / whom to display deprecation messages to, the log spam is mostly about FOREACH & NEXT. De-deprecating these two macros or dropping them already in 1.16 would be an alternative.

Pentarctagon:
I've never really liked that the warnings (and deprecation messages especially) are to a certain degree hidden.  Having those messages doesn't do anything if nobody ever sees them, and removing something that's deprecated when very few, if any, UMC authors will have seen the deprecation messages seems like it defeats the purpose of deprecating something in the first place.

So:

> Maybe it's possible to display the deprecation messages dependent on whether the player has a _server.pbl in the add-on, or lacks an _info.cfg?

would be good, IMO.  Something else would be to enable deprecation warnings by default on development branches (1.13, 1.15, etc).

CelticMinstrel:
Did 1.12 even have deprecation messages? Anyway, the original intention when I implemented the deprecation system was that they're hidden unless someone enables debug mode. Developers commonly enable debug mode, so with this policy they'll be shown in-game only to the people who need to see them. The problem with that was that some players _also_ commonly enable debug mode, but I'm pretty sure they can also be turned off in preferences? Hiding them by default is the worst option, I think. It's better to show them to everyone than hide them from everyone if those are the only two options.

I think that basing deprecation messages on the existence of a _server.pbl would be very difficult, as there's no way that I know of to associate a given deprecation message with a specific addon (and showing deprecation messages in all addons just because someone develops one addon is bad). However, it's a very good idea if it can be done.

CelticMinstrel:
[...]

PR #6532, opened 2022-02-17
Replace deprecated placement=map_overwrite

Wedge009:
I think deprecation notifications are only being shown in 1.17 but not 1.16. I noticed the `map_overwrite` deprecation message when testing the former.

I checked the change in SotA S17, no apparent regressions. I don't see any other instances of `map_overwrite`.

Info: https://wiki.wesnoth.org/SingleUnitWML#placement

### DIFF
--- a/data/campaigns/Heir_To_The_Throne/scenarios/08_The_Princess_of_Wesnoth.cfg
+++ b/data/campaigns/Heir_To_The_Throne/scenarios/08_The_Princess_of_Wesnoth.cfg
@@ -313,7 +313,8 @@
             x,y=$x1,$y1
             side=2
             type=$unit.type
-            placement=map_overwrite
+            placement=map
+            overwrite=yes
             moves=0
             attacks_left=0
             [modifications]
--- a/data/campaigns/Secrets_of_the_Ancients/scenarios/17_Mortality.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/scenarios/17_Mortality.cfg
@@ -309,7 +309,8 @@ Not far inside the cave was a ruined castle built in a style I did not recognize
         [unit]
             type=SotA Skeleton
             side=1
-            placement=map_overwrite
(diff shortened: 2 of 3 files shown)


stevecotton:
> I think deprecation notifications are only being shown in 1.17 but not 1.16.

There's a setting in advanced preferences, "show deprecation messages in-game". If it hasn't been set by the user, then it currently defaults to "no" in 1.16, "yes" in canonical 1.17.0 and 1.17.1, but "no" for 1.17.x+dev (debate about that might start up in #5084, and I've been wondering about it in #6512).

stevecotton (approved):

Wedge009:
Funnily enough I was just re-reading #5084 when you wrote that. What's 'canonical 1.17.0 and 1.17.1', though? I wasn't aware we had any releases for 1.17 yet. And I don't recall setting any deprecation message preferences so that doesn't really explain how I saw the messages in 1.17.0+dev...

Wedge009:
Update: `preferences` file is shared across versions by default on Linux, and I just confirmed I don't have 'show deprecation messages' enabled. Well, I suppose it's just as well I saw the message regardless - like with a lot of these recent deprecation-related issues, it's been deprecated since 1.13.x days.

stevecotton:
Interesting, I might have misunderstood what turns on and off the deprecation messages' default. I thought it was `is_dev_version()`:

https://github.com/wesnoth/wesnoth/blob/02cdde963b0ac548fa9d5eaedd54b6b02861c466/src/game_version.cpp#L170-L176

Wedge009:
Well, I've enabled it now so if I run into more deprecated code in the bundled campaigns, etc, I can do something about it.

Elvish-Hunter:
Regarding the deprecated `[unit] placement=` values, wmllint can automatically convert them since commits cde990a (master) and a7b22f5 (1.16), which I pushed yesterday.

PR #6467, opened 2022-01-28
[world conquest] [artifact/abilities] Make banner stack with leadership ability.

knyghtmare:
fixes banner not stacking with leadership

Addresses point 8 of #6411

PR #6535, opened 2022-02-19
[world conquest] Backport of making Banner stack with leadership

knyghtmare:
This was supposed to have been backported around 18-20 January of 2022...

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,8 @@
      * S22: Fixed the possibility of a missplaced dialogue when a bridge was broken (issue #6376)
    * Under the Burning Suns
      * S08: Kromph and the Dust devil now trigger a critical event (issue #6211)
+   * World Conquest
+     * Fix banner ability not stacking with leadership  
  ### Editor
  ### Multiplayer
  ### Lua API
--- a/changelog.md
+++ b/changelog.md
@@ -12,8 +12,6 @@
      * S22: Fixed the possibility of a missplaced dialogue when a bridge was broken (issue #6376)
    * Under the Burning Suns
      * S08: Kromph and the Dust devil now trigger a critical event (issue #6211)
-   * World Conquest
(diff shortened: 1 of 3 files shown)


stevecotton (changes requested):
This can not go to 1.16, because it will cause OOS.

Although these files are WML, they're WML that is loaded from each player's local filesystem, not shared from the host. Players with unpatched versions of Wesnoth will get different stats compared to players with this PR applied.

Apologies for saying in https://github.com/wesnoth/wesnoth/pull/6467#issuecomment-1027375025 that I'd be happy to accept this, I've now tested and confirmed that it causes OOS.

stevecotton:
For 1.16 (only), I guess we need to change WC_STR_BANNER_DESCRIPTION and the Imvest dialog strings instead, to say that it doesn't stack. Hopefully those don't cause OOS.

stevecotton:
Before the string freeze, would it make sense to add a warning that it doesn't stack?

knyghtmare:
> Before the string freeze, would it make sense to add a warning that it doesn't stack?

I would have preferred to patch up this bug given the recent player dissatisfaction with WC in its forum thread. Just telling them it doesn't stack might not really do much. However, OOS is annoying on an equal level as well.

knyghtmare:
Probably will frustrate players...but OOS...
Done with text update

stevecotton (approved):

Pentarctagon:
@knyghtmare I assume this can be merged now?

knyghtmare:
> @knyghtmare I assume this can be merged now?

I think the discussion reached a stalemate last time since it was "fixing a bug in stable and causing OOS" and "improving player experience with WC but causing save-file OOS"
https://discord.com/channels/231976805987385345/259976436490829825/944928345961926787

However, the answer to your question is: yes, it can be merged...but yeah the discussion's outcome was not clear to me.

PR #6536, opened 2022-02-19
[world conquest] update deprecated share_view to share_vision

knyghtmare:
```lua
share_view = true,
```
To:
```lua
share_vision = "all",
```


stevecotton (approved):
The CI checks are still running, so approving rather than merging immediately.

CelticMinstrel:
And now they've passed, so merging.

PR #6553, opened 2022-03-06
Piglets: add a po hint that these are wild boars' young

stevecotton:
Possibly unnecessary, but can't hurt.


doofus-01 (approved):
No objections

PR #6557, opened 2022-03-09
Minor req cleanup

Vultraz:
For CI.


stevecotton:
I think this one too, but I haven't done a build to confirm that it can be removed:
```
SConstruct:     conf.CheckBoost("thread") & \
```

Vultraz:
Ok, so the scons change breaks things??

PR #6559, opened 2022-03-12
Fix #6389.

### LINKED ISSUE #6389: Addons Tab crashes
 - What version of the game are you running? 1.16/1.17 most recent flatpaks
 - What OS are you running? Ubuntu 20

### Describe the bug
If you click addons tab, Wesnoth crashes, in the command line there is

> Ran out of memory. Aborted.

### To Reproduce
Install manually two addons from this archive [CrashersOfWesnoth.zip](https://github.com/wesnoth/wesnoth/files/7738371/CrashersOfWesnoth.zip)

### Additional context
In the archive there is an empty addon spoilt_base and an addon spoilt1 with a _server.pbl in which spoilt_base is mentioned as a dependency. It's crucial to repro. If you change the dependencies of spoilt1 to be an exisiting addon, it stops crashing. If you change the dependency to a non-exisiting addon (neither published nor locally exisitng), it also works fine. I presume the bug might be when _server.pbl has a mention of some addon which is locally there but not published.

~Also a custom build from commit 91d1296af9 doen't repro~

91d1296af9 and 3cce671, the crash occurs when you try to select the bad addon. If this addon is the first, the crash happens immediately when Addons opened. If it's not the first, to cause a crash you need to select the bad addon

Pentarctagon:
Previously this code assumed that all add-ons would have a _server.pbl or an _info.cfg. When neither was the case, it tried to dereference the begin() of an empty set, causing the crash.

### DIFF
--- a/src/addon/state.cpp
+++ b/src/addon/state.cpp
@@ -51,7 +51,11 @@ addon_tracking_info get_addon_tracking_info(const addon_info& addon)
 		} else {
 			// We normally use the _info.cfg version instead.
 			t.installed_version = get_addon_version_info(id);
-			t.remote_version = *addon.versions.begin();
+			if(addon.versions.size() > 0) {
+				t.remote_version = *addon.versions.begin();
+			} else {
+				t.remote_version = version_info(0,0,0);
+			}
 		}
 
 		if(t.remote_version == t.installed_version) {


Pentarctagon:
Tested locally and it's a pretty simple change, so I'll merge this tomorrow if there's no comments.

Vultraz:
Suggest making it an `optional`.

CelticMinstrel:
That would require checking everywhere it's accessed, though, so it would be potentially a much bigger change.

Vultraz:
image

Pentarctagon:
What would that accomplish though? Making remote_version an optional would mean I'd still need to make the same check here (to avoid the bad dereference) while also adding handling anywhere it's currently used rather than setting a default value here once.

PR #6570, opened 2022-03-22
Misc cleanups

Vultraz:
For CI.

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -962,7 +962,7 @@ gui::button::TYPE display::string_to_button_type(const std::string& type)
 
 static const std::string& get_direction(std::size_t n)
 {
-	static const std::array<std::string, 6> dirs {{ "-n", "-ne", "-se", "-s", "-sw", "-nw" }};
+	static const std::array<std::string, 6> dirs { "-n", "-ne", "-se", "-s", "-sw", "-nw" };
 	return dirs[n >= dirs.size() ? 0 : n];
 }
 
@@ -1221,18 +1221,6 @@ void display::drawing_buffer_add(const drawing_layer layer,
 {
 	drawing_buffer_.emplace_back(layer, loc, x, y, surf, clip);
 }
-
-// FIXME: temporary method. Group splitting should be made
-// public into the definition of drawing_layer
-//
-// The drawing is done per layer_group, the range per group is [low, high).
-const std::array<display::drawing_layer, 4> display::drawing_buffer_key::layer_groups {{
-	LAYER_TERRAIN_BG,
-	LAYER_UNIT_FIRST,
-	LAYER_UNIT_MOVE_DEFAULT,
-	// Make sure the movement doesn't show above fog and reachmap.
-	LAYER_REACHMAP
-}};
 
 enum {
 	// you may adjust the following when needed:
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -877,6 +877,30 @@ void display::layout_buttons()
 	}
 }
 
+namespace
+{
+gui::button::TYPE string_to_button_type(const std::string& type)
+{
+	if(type == "checkbox") {
+		return gui::button::TYPE_CHECK;
+	} else if(type == "image") {
+		return gui::button::TYPE_IMAGE;
+	} else if(type == "radiobox") {
+		return gui::button::TYPE_RADIO;
+	} else if(type == "turbo") {
+		return gui::button::TYPE_TURBO;
+	} else {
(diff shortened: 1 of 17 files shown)


src/font/font_config.cpp
```diff
@@ -87,7 +87,7 @@ namespace
 #ifdef CAIRO_HAS_WIN32_FONT
 bool is_valid_font_file(const std::string& file)
 {
-	static const std::array<std::string, 3> font_exts { ".ttf", ".ttc", ".otf" };
+	static const std::array font_exts { ".ttf", ".ttc", ".otf" };
```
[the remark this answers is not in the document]
CelticMinstrel: Using `auto` two lines down should help, depending on the API of `filesystem::ends_with`.

src/replay.cpp
```diff
@@ -106,7 +106,7 @@ static void verify(const unit_map& units, const config& cfg) {
 		u->write(u_cfg);
 
 		bool is_ok = true;
-		static const std::array<std::string, 4> fields {{"type","hitpoints","experience","side"}};
+		static const std::array fields {"type","hitpoints","experience","side"};
```
CelticMinstrel, line 109: Again you've changed the type from `std::string` to `const char*`. Same as previous comment.

src/scripting/game_lua_kernel.cpp
```diff
@@ -4726,7 +4726,7 @@ void game_lua_kernel::set_game_display(game_display * gd) {
  * elsewhere (in the C++ code).
  * Any child tags not in this list will be passed to Lua's on_load event.
  */
-static const std::array<std::string, 24> handled_file_tags {{
+static const std::array handled_file_tags {
```
CelticMinstrel, line 4729: Same thing again.

(Aside: It seems to me that the function using this could just be an `std::find` call.)

src/tests/test_serialization.cpp
```diff
@@ -57,52 +57,52 @@ BOOST_AUTO_TEST_CASE( utils_split_test )
 
 	{
 		auto split = utils::split(test_string);
-		std::array<std::string, 7> expect = {"a", "bb", "ccc  ||  d", "ee", "fff  | |  g", "hh", "iii"};
+		std::array expect = {"a", "bb", "ccc  ||  d", "ee", "fff  | |  g", "hh", "iii"};
```
CelticMinstrel, line 60: This probably no longer compiles because of the type change. My suggestion here is to just append `s` to each constant. (Note: use of that literal suffix does require `using namespace std::literals;` if I recall correctly.)

Vultraz: Other tests already did the same thing without the literal.

src/enum_base.hpp
```diff
@@ -82,9 +82,13 @@ struct enum_base : public T
...
 };
 
-#define ENUM_AND_ARRAY(...) \
-	static constexpr std::size_t count = std::tuple_size<decltype(std::make_tuple(__VA_ARGS__))>::value; \
-	enum class type { __VA_ARGS__ }; \
-	static constexpr std::array<const char*, count> values{ __VA_ARGS__ };
+#define ENUM_AND_ARRAY(...)                                                                                            \
+	enum class type { __VA_ARGS__ };                                                                                   \
+                                                                                                                       \
+	/** Provide a variable template for an array of matching size. */                                                  \
```
CelticMinstrel, line 88: This is not a variable template. A variable template is like `template<typename T> bool whatever = something<T>::value;`. Please fix the comment.

src/display.cpp
```diff
@@ -877,6 +877,30 @@ void display::layout_buttons()
...
+{
+gui::button::TYPE string_to_button_type(const std::string& type)
+{
+	if(type == "checkbox") {
+		return gui::button::TYPE_CHECK;
+	} else if(type == "image") {
+		return gui::button::TYPE_IMAGE;
+	} else if(type == "radiobox") {
+		return gui::button::TYPE_RADIO;
+	} else if(type == "turbo") {
```
CelticMinstrel, line 890: …what on earth is a "turbo"?

Vultraz: No idea!

src/enum_base.hpp
Vultraz: Ah, you're right, I meant alias template. I removed `count` since I made `values` use the alias so I figured might as well subsume it.

src/tests/test_serialization.cpp
CelticMinstrel: Oh, I misread the code. Sorry about that. This one is fine.

PR #6571, opened 2022-03-22
[1.16] Add a unit test for [move_unit]location_id with an occupied hex

stevecotton:


data/test/scenarios/test_move_unit_to_occupied_hex.cfg
```diff
@@ -0,0 +1,55 @@
...
+            [have_unit]
+                id=bob
+                x=13
+                y=3
+            [/have_unit]
+        )}
+
+        [move_unit]
+            id=alice
+            to_location=2
```
Pentarctagon, line 35: This is a location that has already been defined, I assume?

stevecotton: Numerical location ids `1`, `2`, etc correspond to the starting location for that side.

Pentarctagon (approved):

PR #6573, opened 2022-03-22
fix [firststrike] special with apply_to=opponent crashes Wesnoth 1.16.x

newfrenchy83:
Closes https://github.com/wesnoth/wesnoth/issues/6575. like reported in forum in https://forums.wesnoth.org/viewtopic.php?p=672374#p672374
if unit with "last strike" ([firststrike] special with apply_to=opponent, on any of its attacks) is targeted by other unit for attack, wesnoth crashes,
and other problems exist also for other weapon specials like poisons, when apply_to=opponent used(id two user of this specials fight, one fighter only could be poisoned)

### DIFF
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1557,28 +1557,34 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
 	temporary_facing other_facing(other, other_loc.get_relative_dir(self_loc));
 
 	// Filter poison, plague, drain, first strike
-	if (tag_name == "drains" && other && other->get_state("undrainable")) {
+	bool bool_lambda = ((whom == AFFECT_SELF) || ((whom == AFFECT_EITHER) && special_affects_self(special, is_attacker))) ? true : false;
+	unit_const_ptr lambda_unit = bool_lambda ? other : self;
+	const_attack_ptr lambda_attack = bool_lambda ? other_attack : self_attack;
+	map_location lambda_loc = bool_lambda ? other_loc : self_loc;
+	bool attacker = bool_lambda ? !is_attacker : is_attacker;
+
+	if (tag_name == "drains" && lambda_unit && lambda_unit->get_state("undrainable")) {
 		return false;
 	}
-	if (tag_name == "plague" && other &&
-		(other->get_state("unplagueable") ||
-		 resources::gameboard->map().is_village(other_loc))) {
+	if (tag_name == "plague" && lambda_unit &&
+		(lambda_unit->get_state("unplagueable") ||
+		 resources::gameboard->map().is_village(lambda_loc))) {
 		return false;
 	}
-	if (tag_name == "poison" && other &&
-		(other->get_state("unpoisonable") || other->get_state(unit::STATE_POISONED))) {
+	if (tag_name == "poison" && lambda_unit &&
+		(lambda_unit->get_state("unpoisonable") || lambda_unit->get_state(unit::STATE_POISONED))) {
 		return false;
 	}
-	if (tag_name == "firststrike" && !is_attacker && other_attack &&
-		other_attack->has_special_or_ability("firststrike")) {
+	if (tag_name == "firststrike" && attacker && lambda_attack &&
+		lambda_attack->has_special_or_ability("firststrike")) {
 		return false;
 	}
-	if (tag_name == "slow" && other &&
-		(other->get_state("unslowable") || other->get_state(unit::STATE_SLOWED))) {
+	if (tag_name == "slow" && lambda_unit &&
+		(lambda_unit->get_state("unslowable") || lambda_unit->get_state(unit::STATE_SLOWED))) {
 		return false;
 	}
-	if (tag_name == "petrifies" && other &&
-		other->get_state("unpetrifiable")) {
+	if (tag_name == "petrifies" && lambda_unit &&
+		lambda_unit->get_state("unpetrifiable")) {
 		return false;
 	}
 
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1557,10 +1557,9 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
 	temporary_facing other_facing(other, other_loc.get_relative_dir(self_loc));
 
 	// Filter poison, plague, drain, first strike
-	bool bool_lambda = ((whom == AFFECT_SELF) || ((whom == AFFECT_EITHER) && special_affects_self(special, is_attacker))) ? true : false;
-	unit_const_ptr lambda_unit = bool_lambda ? other : self;
-	const_attack_ptr lambda_attack = bool_lambda ? other_attack : self_attack;
-	bool attacker = bool_lambda ? !is_attacker : is_attacker;
+	bool bool_victim = ((whom == AFFECT_SELF) || ((whom == AFFECT_EITHER) && special_affects_self(special, is_attacker))) ? true : false;
+	const_attack_ptr victim_attack = bool_victim ? other_attack : self_attack;
+	bool attacker = bool_victim ? !is_attacker : is_attacker;
 
 	if (tag_name == "drains" && other && other->get_state("undrainable")) {
 		return false;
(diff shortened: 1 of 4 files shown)


newfrenchy83:
Closes https://github.com/wesnoth/wesnoth/issues/6575

src/units/abilities.cpp
```diff
@@ -1557,30 +1557,54 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
 	temporary_facing other_facing(other, other_loc.get_relative_dir(self_loc));
 
 	// Filter poison, plague, drain, first strike
-	if (tag_name == "drains" && other && other->get_state("undrainable")) {
+	if (tag_name == "drains" && special_affects_self(special, is_attacker) && other && other->get_state("undrainable")) {
 		return false;
 	}
-	if (tag_name == "plague" && other &&
+	if (tag_name == "drains" && special_affects_opponent(special, is_attacker) && self && self->get_state("undrainable")) {
```
stevecotton, line 1563: Should this depend on `special_active_impl`'s `whom` argument to work out which logic to check?

As the `(affects_self && other && other->get_state)` and `affects_opponent && self && self->get_state` logic is repeated many times, I'm thinking it would be better to create a local lambda function so that this `if` could become
```
if(tag_name == "drains" && lambda_name("undrainable"))
```
and later ones could become
```
if(tag_name == "poison" && lambda_name("unpoisonable", unit::STATE_POISONED))
```

newfrenchy83: changes done

stevecotton: I meant a [lambda function](https://en.cppreference.com/w/cpp/language/lambda). However, the new code is simpler than a lambda would be, so there's no need to make more changes for this.

stevecotton:
I still need time to read the code and properly understand what it's doing. I understand that the crash was caused by an infinite recursion that would only happen for firststrike, so I'm wondering about the other changes.

> other problems exist also for other weapon specials like poisons, when apply_to=opponent used(id two user of this specials fight, one fighter only could be poisoned)

Interesting, and confirmed both that that bug exists and that this fixes it.

Pentarctagon:
This sounds like something that could have unit tests created for the specials affected?

newfrenchy83:
> This sounds like something that could have unit tests created for the specials affected?

test added and validated

stevecotton:
I feel we need to fix the crash on 1.16, but I'm not sure whether the other stats should only be done on 1.17. Technically, all the ones that don't fix a crash are a potential source of OOS. That should probably be discussed on the IRC / Discord channel, no need to change the PR until there's a decision.

For the unit test, it would be good to move the new test to its own file, but I'm currently looking at reducing the amount of per-test code for similar tests.

stevecotton:
Sorry to keep asking for changes, but the conclusion on IRC was that we should only fix the crash on 1.16. Anything that doesn't crash should only change on 1.17. I'll have another look tomorrow, the fix for firststrike is working, but I want to check a few more edge cases about it (haven't quite understood why it gives the right answer in some cases).

newfrenchy83:
> Sorry to keep asking for changes, but the conclusion on IRC was that we should only fix the crash on 1.16. Anything that doesn't crash should only change on 1.17. I'll have another look tomorrow, the fix for firststrike is working, but I want to check a few more edge cases about it (haven't quite understood why it gives the right answer in some cases).

change done

newfrenchy83:
and checoking don't work yet.

src/units/abilities.cpp
```diff
@@ -1557,6 +1557,11 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
 	temporary_facing other_facing(other, other_loc.get_relative_dir(self_loc));
 
 	// Filter poison, plague, drain, first strike
+	bool bool_lambda = ((whom == AFFECT_SELF) || ((whom == AFFECT_EITHER) && special_affects_self(special, is_attacker))) ? true : false;
+	unit_const_ptr lambda_unit = bool_lambda ? other : self;
```
stevecotton, line 1561: `lambda_unit` is unused.

For the other `lambda_*` variables, I'm trying to think of suggestions for better names. For 1.17, I'd suggest `victim` "For status conditions, the unit that will be harmed by this condition".

newfrenchy83: changes done

stevecotton:
Because firststrike is different to the other abilities, I think it would readability to move it away from the status condition checks. By moving it below `// Translate our context into terms of "attacker" and "defender".`, it can use the `att_weapon` variable too:

```
// If both units have first strike then the effects cancel out. Only check
// the opponent if "whom" is the defender, otherwise this leads to infinite
// recursion.
if (tag_name == "firststrike") {
	// True if "whom" corresponds to "self", false if "whom" is "other"
	bool whom_is_self = (whom == AFFECT_SELF) || ((whom == AFFECT_EITHER) && special_affects_self(special, is_attacker));
	bool whom_is_defender = whom_is_self ? !is_attacker : is_attacker;
	if (whom_is_defender && att_weapon && att_weapon->has_special_or_ability("firststrike"))
		return false;
}
```

Sorry if this is taking over your work - I thought that I'd do a follow-up PR just to add comments to the code, but the review notes turned into code too.

stevecotton (approved):
@Pentarctagon or @CelticMinstrel, please would one of you review this (and squash-merge, I'm happy that that'll reattribute my unit-test commit to @newfrenchy83)?

I think it's good and newfrenchy deserves the full credit for it. The reason that I'm asking one of the two of you to review is I'm now looking at it with the eyes of a co-author rather than the fresh eyes of a reviewer.

newfrenchy83:
> @Pentarctagon or @CelticMinstrel, please would one of you review this (and squash-merge, I'm happy that that'll reattribute my unit-test commit to @newfrenchy83)?
> 
> I think it's good and newfrenchy deserves the full credit for it. The reason that I'm asking one of the two of you to review is I'm now looking at it with the eyes of a co-author rather than the fresh eyes of a reviewer.

review https://github.com/wesnoth/wesnoth/pull/6574 either please.

src/units/abilities.cpp
```diff
@@ -1591,6 +1587,17 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
...
+	// Filter firststrike here, if both units have first strike then the effects cancel out. Only check
+	// the opponent if "whom" is the defender, otherwise this leads to infinite
+	// recursion.
+	if (tag_name == "firststrike") {
+		// True if "whom" corresponds to "self", false if "whom" is "other"
+		bool whom_is_self = (whom == AFFECT_SELF) || ((whom == AFFECT_EITHER) && special_affects_self(special, is_attacker));
+		bool whom_is_defender = whom_is_self ? !is_attacker : is_attacker;
+		if (whom_is_defender && att_weapon && att_weapon->has_special_or_ability("firststrike"))
+			return false;
+	}
```
Pentarctagon, line 1599: Not that it should be done in this PR, but I wonder if these should be converted to a (string) enum rather than putting the weapon special tag name string directly in a bunch of places.

src/units/abilities.cpp
```diff
@@ -1556,7 +1556,7 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
 	temporary_facing self_facing(self, self_loc.get_relative_dir(other_loc));
 	temporary_facing other_facing(other, other_loc.get_relative_dir(self_loc));
 
-	// Filter poison, plague, drain, first strike
+	// Filter poison, plague, drain, slow, petrifies
```
Pentarctagon, line 1559: The parts looking at slow and petrified seem unchanged, so this comment shouldn't have removed them, I think?

src/units/abilities.cpp
```diff
@@ -1591,6 +1587,17 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
 	const_attack_ptr att_weapon = is_attacker ? self_attack : other_attack;
 	const_attack_ptr def_weapon = is_attacker ? other_attack : self_attack;
 
+	// Filter firststrike here, if both units have first strike then the effects cancel out. Only check
+	// the opponent if "whom" is the defender, otherwise this leads to infinite
+	// recursion.
```
Pentarctagon, line 1592: Put "recursion" on the line above - the first comment line is already longer anyway.

src/units/abilities.cpp
newfrenchy83: slow and petrified added recently but no specified in comment, i use this PR to fix this.

src/units/abilities.cpp
newfrenchy83: Done.

src/units/abilities.cpp
Pentarctagon: Ah, read the diff backwards. It should still mention firststrike though then?

stevecotton: This section is now the status-aliment abilities which are countered by unpoisonable, unslowable, etc, and `drain` which is handled in the same way as the status-aliment abilities.

The reason for moving firststrike later in the file was to get access to the `att_weapon` special, but it's also the case that the logic for filtering it is different to the logic in this block.

Pentarctagon (approved):

PR #6574, opened 2022-03-22
fix [firststrike] special with apply_to=opponent crashes Wesnoth 1.17.x

newfrenchy83:
like reported in forum in https://forums.wesnoth.org/viewtopic.php?p=672374#p672374
if unit with "last strike" ([firststrike] special with apply_to=opponent, on any of its attacks) is targeted by other unit for attack, wesnoth crashes,
and other problems exist also for other weapon specials like poisons, when apply_to=opponent used(id two user of this specials fight, one fighter only could be poisoned)

### DIFF
--- a/data/test/scenarios/swarm_disables_upgrades.cfg
+++ b/data/test/scenarios/swarm_disables_upgrades.cfg
@@ -129,3 +129,117 @@
         {SUCCEED}
     [/event]
 )}
+
+{GENERIC_UNIT_TEST "poison_opponent" (
+    [event]
+        name=start
+        [modify_unit]
+            [filter]
+            [/filter]
+            max_hitpoints=100
+            hitpoints=100
+            attacks_left=1
+        [/modify_unit]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [poison]
+                        apply_to=opponent
+                    [/poison]
+                    [attacks]
+                        value=1
+                    [/attacks]
+                    [damage]
+                        value=1
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                    [/chance_to_hit]
+                [/set_specials]
+            [/effect]
+            [filter]
+                id=bob
+            [/filter]
+        [/object]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [poison]
+                        apply_to=opponent
+                    [/poison]
+                    [attacks]
+                        value=1
+                    [/attacks]
+                    [damage]
+                        value=1
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                    [/chance_to_hit]
+                [/set_specials]
+            [/effect]
+            [filter]
+                id=alice
+            [/filter]
+        [/object]
+
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+            kill=yes
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        [unstore_unit]
+            variable=a
+            find_vacant=yes
+            x,y=$b.x,$b.y
+        [/unstore_unit]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+
+        [do_command]
+            [attack]
+                weapon=0
+                defender_weapon=0
+                [source]
+                    x,y=$a.x,$a.y
+                [/source]
+                [destination]
+                    x,y=$b.x,$b.y
+                [/destination]
+            [/attack]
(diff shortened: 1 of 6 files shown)


newfrenchy83:
@Pentarctagon why the checking don't work?

newfrenchy83:
@Pentarctagon test added and validated here too.

data/test/scenarios/swarm_disables_upgrades.cfg
```diff
@@ -129,3 +129,4 @@
         {SUCCEED}
     [/event]
 )}
+
```
stevecotton, line 132: Stray line

newfrenchy83: fixed

src/units/abilities.cpp
```diff
@@ -1537,29 +1537,30 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
...
-	// Filter poison, plague, drain, first strike
-	if (tag_name == "drains" && other && other->get_state("undrainable")) {
-		return false;
-	}
-	if (tag_name == "plague" && other &&
-		(other->get_state("unplagueable") ||
-		 resources::gameboard->map().is_village(other_loc))) {
+	// Filter poison, plague, drain, slow, petrifies
+	// True if "whom" corresponds to "self", false if "whom" is "other"
+	bool whom_is_self = ((whom == AFFECT_SELF) || ((whom == AFFECT_EITHER) && special_affects_self(special, is_attacker))) ? true : false;
```
stevecotton, line 1542: No need for the `? true : false`, because the part that's to the left of the `?` is already a boolean.

newfrenchy83: removed

newfrenchy83:
@Pentarctagon don't forget this PRfor master branch

Pentarctagon:
@stevecotton should I look at this one as well, or did you only mean the 1.16 version of the PR?

src/units/abilities.cpp
```diff
@@ -1537,29 +1537,30 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
...
-		return false;
-	}
-	if (tag_name == "plague" && other &&
-		(other->get_state("unplagueable") ||
-		 resources::gameboard->map().is_village(other_loc))) {
+	// Filter poison, plague, drain, slow, petrifies
+	// True if "whom" corresponds to "self", false if "whom" is "other"
+	bool whom_is_self = ((whom == AFFECT_SELF) || ((whom == AFFECT_EITHER) && special_affects_self(special, is_attacker)));
+	unit_const_ptr whom_is_unit = whom_is_self ? other : self;
+	map_location whom_is_loc = whom_is_self ? other_loc : self_loc;
```
stevecotton, line 1544: Minor, but I'd suggest `whom_unit` and `whom_loc`. The `is` was for "the unit which is `whom` **is** the same unit that's `self`".

newfrenchy83: @stevecotton done, and i added your laststrike test.

src/units/abilities.cpp
```diff
@@ -1537,29 +1537,30 @@ bool attack_type::special_active_impl(const_attack_ptr self_attack, const_attack
...
-		return false;
-	}
-	if (tag_name == "plague" && other &&
-		(other->get_state("unplagueable") ||
-		 resources::gameboard->map().is_village(other_loc))) {
+	// Filter poison, plague, drain, slow, petrifies
+	// True if "whom" corresponds to "self", false if "whom" is "other"
+	bool whom_is_self = ((whom == AFFECT_SELF) || ((whom == AFFECT_EITHER) && special_affects_self(special, is_attacker)));
+	unit_const_ptr whom_unit = whom_is_self ? other : self;
+	map_location whom_loc = whom_is_self ? other_loc : self_loc;
```
stevecotton, line 1544: Sorry, I was confused when suggesting names. `whom_unit` would make sense for the unit that is `whom`, but the unit here is the opponent of `whom`. How about `whom_is_self`, `them_unit` and `them_loc`; alternatively `whom_is_self`, `them`, `their_loc`?

So the comment could be
```
// There are only two units involved in this logic, but three pairs of names for them: {self, other},
// {attacker, defender} and {whom, them}.
```

stevecotton:
> @stevecotton should I look at this one as well, or did you only mean the 1.16 version of the PR?

I think it's good, except for the mixed-up names (sorry newfrenchy83, those were my fault). Comments would be welcome, but I think it'll be ready to merge once those names change.

It behaves as expected with the tests in PR #6582 - the ones that say "known bug in 1.16" fail and all the other pass. That's good, I'll open a PR with the 1.17 versions of those tests once this merges.

src/units/abilities.cpp
newfrenchy83: @stevecotton names changed, but i not added comment.

stevecotton (approved):
I'm sure the CI is going to pass, but will wait for it.

PR #6577, opened 2022-03-24
[DID] [Alone at Last] Ensure Spectral Servant unit-line cannot pick up the book

### LINKED ISSUE #6576: [Campaign] [DiD] The special Ghost can pickup the book which it shouldn't.

- 1.17.2+dev
- Github
- Latest Commit
- Windows 10

### Description of the bug

The special ghost line picks up the book which it should not.

1. Kill Volk
2. Move Eidolon/Phantom/Spectral Servant to Book
3. See Bug

### Expected behavior

The special ghost line should not be able to pick up the book as per the scenario set up and trigger a "too heavy for me dialogue"

knyghtmare:
Fixes #6576

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
@@ -998,7 +998,7 @@
             x=$bookX
             y=$bookY
             side=1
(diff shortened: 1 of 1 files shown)


Wedge009:
As I mentioned in #wesnoth, given that the regular ghosts are already prevented from taking the book, I agree that the special Spectral Servant should be excluded as well.

nemaara (approved):

PR #6582, opened 2022-03-26
[1.16] Unit tests for [drains], [poison] and [slows], including apply_to=opponent

stevecotton:
Here `apply_to=opponent` means that the weapon special gives the opponent the
ability, the unit that should get poisoned or slowed is the unit that has the
weapon special.

There's a known bug in 1.16, that `apply_to=opponent` check the wrong unit to
see it it's `unpoisonable`, `undrainable` etc. It also checks the wrong unit to
see if it's already poisoned or slowed, so a battle between two units that both
have reverse-poison results in at most one being poisoned.

As 1.16 has already been released, to avoid OOS the test is checking that the
current behavior's known bug is preserved. For the 1.17 branch the four tests
labelled `preserving known bug` will be changed to test the reverse.

Most of the credit for this is Newfrenchy's, as he's already written a fix
and a WML based test. This commit uses a Lua test instead to test more
combinations of statuses.

This adds a `COMMON_KEEP_A_B_UNIT_TEST` macro, which is a counterpart to the
`GENERIC_UNIT_TEST` macro that starts the leaders next to each other, ready
to attack. The `A_B` is because I'm planning a multiple-side variant too.

There's no test for [petrify], as simulate_combat doesn't provide a stat for it.

This tests only 3 of the 6 abilities whose behavior will change in 1.17's
equivalent of 1.16's 7b39b656063. That's sufficient to prevent any accidental
copy of the 1.17 fix to 1.16, and my thoughts on testing the others are:
* [firststrike]'s test is in 7b39b656063. It crashed, so is fixed in 1.16.
* [drains], [poison] and [slow] are tested here.
* [petrify] ends combat, it's also not exposed in simulate_combat's stats.
* [plague] triggers after combat ends.

### DIFF
--- a/data/test/macros/start_position_common_keep_a_b.cfg
+++ b/data/test/macros/start_position_common_keep_a_b.cfg
@@ -1,0 +1,43 @@
+#textdomain wesnoth-test
+
+##
+# Starting state:
+# Side 1 leader Alice (Elvish Archer)
+# Side 2 leader Bob (Orcish Grunt)
+#
+# Both leaders are on a single keep, adjacent to each other.
+# There is no free castle hex to recruit onto.
+##
+#define COMMON_KEEP_A_B_UNIT_TEST NAME CONTENT
+    [test]
+        name=_ "Unit Test " + {NAME}
+        map_file=test/maps/2p_single_castle.map
+        turns=unlimited
+        id={NAME}
+        random_start_time=no
+        is_unit_test=yes
+
+        {DAWN}
+
+        [side]
+            side=1
+            controller=human
+            name=_ "Alice"
+            type=Elvish Archer
+            id=alice
+            fog=no
+            team_name=West
+        [/side]
+        [side]
+            side=2
+            controller=human
+            name=_ "Bob"
+            type=Orcish Grunt
+            id=bob
+            fog=no
+            team_name=East
+        [/side]
+
+        {CONTENT}
+    [/test]
+#enddef
--- a/data/test/maps/2p_single_castle.map
+++ b/data/test/maps/2p_single_castle.map
@@ -1,0 +1,7 @@
+Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg
+Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg
+Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg
+Gg, Gg, Gg, Gg, 1 Ke, 2 Ke, Gg, Gg, Gg, Gg
+Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg
+Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg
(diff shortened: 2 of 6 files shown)


data/test/macros/wml_unit_test_macros.cfg
```diff
@@ -113,13 +165,13 @@ Elvish Archer#endarg
...
-[if]
-    [variable]
-        name={FLAG}
-        not_equals={NOT_EQUALS}
-    [/variable]
-    [then]
-        {FAIL}
-    [/then]
-[/if]
+    [if]
```
CelticMinstrel, line 168: Why randomly indenting stuff? (Or was this caused by `wmlindent`? In which case the question would be, why isn't the change already in master and why didn't the CI fail because it's not?)

stevecotton: It was caused by `wmlindent`. There's a CI exclusion in 599504cee3544ad19007913aae35b81b5375b6e7, for which the GNA bug became Github's #1397.

stevecotton:
If there's no more coments then I'll use similar code for `drains` and `slows`. The test for firststrike / laststrike ended up looking different.

stevecotton:
It now tests slows, poison and drains. The slows and poison tests are just a find-and-replace of each other, but drains is a separate test.

I've called it `reflexive_drains` but it's also the only unit test for the drains ability.

CelticMinstrel:
> There's no test for [petrify], as simulate_combat doesn't provide a stat for it.

Obviously not for 1.16, but this could (and probably should) be added.

stevecotton:
> > There's no test for [petrify], as simulate_combat doesn't provide a stat for it.
>
> Obviously not for 1.16, but this could (and probably should) be added.

We should, but even for 1.17 I think that's going to be a separate feature. At the moment, attack prediction for petrify works by running a combat simulation with damage set to the potentially petrified unit's max hp, and includes a FIXME comment about drain.
https://github.com/wesnoth/wesnoth/blob/c1d7e65cbcc69cd041819cbb9c27e9a41a54bfea/src/attack_prediction.cpp#L2076-L2085

stevecotton:
@CelticMinstrel okay to merge?

stevecotton:
Going to rebase and force-push for a trivial fixup to the commit message - there's now seven asserts for preserving the known status, not four. (Noticed now because I'm rebasing the 1.17 version).

data/test/macros/wml_unit_test_macros.cfg
stevecotton: Revised plan - I won't change `wml_unit_test_macros.cfg` in this PR.

Instead of adding the `COMMON_KEEP_A_B_UNIT_TEST` to `wml_unit_test_macros.cfg`, I want to put it in a separate file. There will then be a [separate PR](https://github.com/wesnoth/wesnoth/pull/6627) for master (possibly backported too) that moves `GENERIC_UNIT_TEST` to its own file and extracts `COMMON_KEEP_A_B_C_D_UNIT_TEST` from the 4 side setup in `firststrike_and_laststrike`.

The main reason for having these in separate files is Git's fuzzy-patch matching when merging and rebasing. Having large blocks of identical text in `wml_unit_test_macros.cfg` can mean the wrong section gets patched.

PR #6588, opened 2022-03-27
1.17 terrain transitions update

doofus-01:
This is best summarized with these before/after images.  Possibly the rubble is too busy now, maybe it should have bigger rocks, or maybe just lower contrasts.
Screenshot_20220327_before
Screenshot_20220327_after


stevecotton:
It's beginning to look like a stone path rather than rubble, IMO.

doofus-01:
> It's beginning to look like a stone path rather than rubble, IMO.

That is up to the map designer, I think?  In any case, I've made some of the rocks bigger.

I'll merge this in a few days, if there are no objections or CI failures.

Pentarctagon:
Having something that could be used for an old/ruined stone isn't a bad thing either.

stevecotton:
> > It's beginning to look like a stone path rather than rubble, IMO.
> 
> That is up to the map designer, I think? In any case, I've made some of the rocks bigger.

Looks good, thanks. Sorry for commenting without checking these in the editor first, having tested them it's easy to distinguish a line of these from a line of the flat terrains Cobbles or Cave Path.

Issue #6575, opened 2022-03-22
[firststrike] special with apply_to=opponent crashes Wesnoth 1.16.x


newfrenchy83:
### Game and System Information

- **Version:**1.17 but also reported in 1.16
- **Downloaded from:**built from source
- **Build info:**gcc
- **OS:**windows 7

### Description of the bug

like reported in https://forums.wesnoth.org/viewtopic.php?p=672374#p672374 if unit with "last strike" ([firststrike] special with apply_to=opponent, on any of its attacks) is targeted by other unit for attack then Wesnoth hangs or (according to BlakerBR) crashes to desktop without writing anything to .log file

### Steps to reproduce the behavior

1.
2.
3.

PR #6592, opened 2022-03-27
[1.16] [draft because it triggers the crash]: Add a unit test for firststrike and laststrike

stevecotton:
Unit test for issue #6575, and confirming that the fix in #6573 works. Will
rebase on top of the fix once the fix is merged.

[ci skip]


newfrenchy83:
will be more simple what i merge this in my PR for test immediately?

stevecotton:
> will be more simple what i merge this in my PR for test immediately?

That would be good. I've force-pushed because I thought of a new test, a negatestrike ability (it gives both combatants firststrike, so that they cancel out and the attacker swings first). Your fix already works with the new test too.

PR #6593, opened 2022-03-27
THoT: Remove unused Ratheln character

### LINKED ISSUE #6587: [Campaign] [SP] Clean Up THoT's code
While foraging code from this campaign and NR for the...well, you know, I have stumbled on things which seem to be obsolete code bits and files no longer in use.
Thus, I am listing them here for removal (or moving them to attic).

- [ ] Character Stat Rathlyn
- [ ] Rathlyn's  portrait

Backstory:
This character was a loyal Arch Mage given in High Pass scenario before the campaign was reworked. After the rework, he is no longer present in this campaign but his character stats define and portrait have stayed. They are no longer required.

Wedge009:
Resolves #6587


knyghtmare (approved):
It's all good. Also, might be good to move the rathelyn portrait to the https://github.com/wesnoth/resources repo

Wedge009:
What's the difference between that repository and 'attic'?

doofus-01:
I think the full-size image or layered image would go in resources, while retired "use-as-is" assets that someone might still want for some reason, like the down-scaled portrait, could go to the attic.  At least, that has been my assumption, it would be good to get that confirmed or refuted.

Wedge009:
Yeah, I was leaning towards that too, just looking at the description of the repository vs the contents of 'attic'. I was trying to find where - if anywhere else - the image might have been used in Wesnoth but it doesn't look like it's used. I'll update to move instead of delete the image.

nemaara:
Is it worth keeping his portrait around? The quality is quite a bit lower than current mainline standards.

nemaara (approved):

Wedge009:
I thought it still holds up in the current set of images, but if it's really deemed 'no good' any more, we can delete it later.

Wedge009:
Didn't notice that. Done in 79c3bfce37165705b0c1abc48d85529897c8c728.

PR #6595, opened 2022-03-29
Create MP Server Dashboard

max-torch:
A web-based dashboard for viewing and exploring mp server data.


knyghtmare (commented):
Looks cool!
Do provide some screenshots in the README.md file so users can see how it will appear as.

max-torch:
> Looks cool! Do provide some screenshots in the README.md file so users can see how it will appear as.

Where can I host the screenshot online?

knyghtmare:
> > Looks cool! Do provide some screenshots in the README.md file so users can see how it will appear as.
> 
> Where can I host the screenshot online?

I would just upload them here in a comment and then used those raw github content links in the MD file.

max-torch:
Let me add this for the record:
First, let me say that the web layout is generated on page load/refresh.
When data is fetched when the layout is generated and when filters are toggled, the entire database is queried. I'm talking
`SELECT * FROM database`
In my own testing, it's not an issue since I've been dealing with a 3 days SQL dump of the database. However, when applied to the whole database, I'm wondering how seriously it will impact performance. 

If it does, there is a refinement to address this:
Fetch the entire data only when generating the layout but not when toggling filters. Sounds easy, but in Plotly Dash, using globals is forbidden. So I can't share the data which was fetched for the layout to the filter callback functions as a global object. The implementation that Dash documentation prescribes is to use an external cache. It can be an external file that holds it, it can be held in the user's browser, or it can be an external database or caching database. The method that seems ideal for this use case is the external file method- I presume this file can be made to only exist where the app is being hosted. For an external file, the prescribed method in the documentation is to use `flask-caching`. A folder that contains cache files will hold temporary app data while the app is running.

So TLDR, if after testing this with wesnoth's server the performance when toggling filters is so bad, I will have to implement flask-caching to fix the issue. Please let me know.

Pentarctagon:
I'd also wonder about memory usage to be honest. The current database has data spanning multiple years at this point, so if you're saying it's going to be pulling in the full data set covering multiple years every time someone loads the page or changes a filter, then that'd almost certainly be a problem.

Purely for performance, there can be additional indexes added as needed. I wonder if it might be best to restrict how far back the dashboard goes to a month or a few months though.

max-torch:
> I'd also wonder about memory usage to be honest. The current database has data spanning multiple years at this point, so if you're saying it's going to be pulling in the full data set covering multiple years every time someone loads the page or changes a filter, then that'd almost certainly be a problem.
> 
> Purely for performance, there can be additional indexes added as needed. I wonder if it might be best to restrict how far back the dashboard goes to a month or a few months though.
> 
> If the necessary filtering can be implemented in SQL though, that'd be the best approach IMO.

Another thing would be to make the initial filter settings only set to the last 3 days. Then, it's the user's decision to widen the filter and fetch the heavier data. I would really like to hear from a test of this on your end, then code on this further.

Pentarctagon:
It takes ~1.2 seconds to select 597,165 rows (from `wesnothd_game_info`, which is the table name used on the real server).

max-torch:
> It takes ~1.2 seconds to select 597,165 rows (from `wesnothd_game_info`, which is the table name used on the real server).

hmm, okay at least it's not the worst with the way it's currently set up... it's in a place where it can still be used... ill look into improving the code when I get time

Pentarctagon:
Adding Elvish-Hunter, since this is a python script.

@soliton- do you know what all would be involved with eventually making this accessible on the real server?

soliton-:
No, didn't have time to look at this yet and probably won't have time the next few days either but I will definitely do so when I get a chance.

soliton-:
I tried to run it but got this error:

> Traceback (most recent call last):
  File "/home/wesnoth/dashboard/app.py", line 195, in <module>
    app.layout = prepare_layout
  File "/home/wesnoth/.local/lib/python3.9/site-packages/dash/dash.py", line 584, in layout
    layout_value = self._layout_value()
  File "/home/wesnoth/.local/lib/python3.9/site-packages/dash/dash.py", line 544, in _layout_value
    layout = self._layout() if self._layout_is_function else self._layout
  File "/home/wesnoth/dashboard/app.py", line 109, in prepare_layout
    df, min_date, max_date = query_database()
  File "/home/wesnoth/dashboard/app.py", line 65, in query_database
    df['log_TIMEDIFF_MINUTES'] = df['TIMEDIFF_MINUTES'].apply(lambda x: math.log(x))
  File "/home/wesnoth/.local/lib/python3.9/site-packages/pandas/core/series.py", line 4433, in apply
    return SeriesApply(self, func, convert_dtype, args, kwargs).apply()
  File "/home/wesnoth/.local/lib/python3.9/site-packages/pandas/core/apply.py", line 1082, in apply
    return self.apply_standard()
  File "/home/wesnoth/.local/lib/python3.9/site-packages/pandas/core/apply.py", line 1137, in apply_standard
    mapped = lib.map_infer(
  File "pandas/_libs/lib.pyx", line 2870, in pandas._libs.lib.map_infer
  File "/home/wesnoth/dashboard/app.py", line 65, in <lambda>
    df['log_TIMEDIFF_MINUTES'] = df['TIMEDIFF_MINUTES'].apply(lambda x: math.log(x))
ValueError: math domain error

Pentarctagon:
@max-torch Is this still something you're pursuing?

max-torch:
> @max-torch Is this still something you're pursuing?

Yes. i will return to this.

Pentarctagon:
@max-torch Same question as before.

max-torch:
@Slayer95
I noticed that you have contributed to Python code of Wesnoth. This is an MP Server Dashboard mainly written in Python using the Pandas library and the Plotly library that I have no time to work on, but i thought it's worth letting you know about this just in case you are interested. I can explain things about this but not work on this.

PR #6596, opened 2022-03-30
Simplify math.hpp's count_leading_zeros

stevecotton:
The old count_ones had undefined behavior in the case of arithmetic underflow,
so replace count_leading_zeros with a simpler implementation.

Note: `std::numeric_limits<signed char>::digits` excludes the sign bit from the
size - so the `for` loop gets an off-by-one error if the 'std::make_unsigned_t`
is removed.


src/utils/math.hpp
```diff
@@ -250,31 +144,19 @@ constexpr unsigned int count_leading_zeros_impl(N n, std::size_t w) {
...
 constexpr unsigned int count_leading_zeros(N n) {
-#if defined(__GNUC__) || defined(__clang__)
-	// GCC’s `__builtin_clz` returns an undefined value when called with 0
-	// as argument.
-	// [<http://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html>]
-	if (n == 0) {
-		// Return the quantity of zero bits in `n` rather than
-		// returning that undefined value.
-		return static_cast<unsigned int>(bit_width(n));
+	static_assert(std::is_integral<N>::value);
```
Vultraz, line 147: Use enable_if. also `is_integral_v`

src/utils/math.hpp
```diff
@@ -250,31 +144,19 @@ constexpr unsigned int count_leading_zeros_impl(N n, std::size_t w) {
...
+	const auto x = static_cast<std::make_unsigned_t<N>>(n);
+	constexpr auto mask = std::remove_cv_t<std::make_unsigned_t<N>>{1};
```
Vultraz, line 149: Can simplify these two `make_unsigned_t`s with `decltype(x)`

stevecotton: Do you mean keeping the first one as-is and simplifying the second one?

src/utils/math.hpp
```diff
@@ -249,32 +143,19 @@ constexpr unsigned int count_leading_zeros_impl(N n, std::size_t w) {
...
-	// GCC’s `__builtin_clz` returns an undefined value when called with 0
-	// as argument.
-	// [<http://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html>]
-	if (n == 0) {
-		// Return the quantity of zero bits in `n` rather than
-		// returning that undefined value.
-		return static_cast<unsigned int>(bit_width(n));
+constexpr std::enable_if_t<std::is_integral_v<N>, unsigned int> count_leading_zeros(N n) {
+	const auto x = static_cast<std::make_unsigned_t<N>>(n);
+	constexpr auto mask = decltype(x){1};
```
CelticMinstrel, line 148: In this type of scenario, I would avoid `auto`.

```cpp
constexpr decltype(x) mask{1};
```

Vultraz: Seconded.

src/utils/math.hpp
Vultraz: All after the `x` declaration.

stevecotton: So, the changes that are already in 8bc4e13? The only bit left is the other comment about whether to use `auto` or not?

src/utils/math.hpp
stevecotton: Changed.

src/utils/math.hpp
```diff
@@ -249,32 +143,19 @@ constexpr unsigned int count_leading_zeros_impl(N n, std::size_t w) {
...
-	// GCC’s `__builtin_clz` returns an undefined value when called with 0
-	// as argument.
-	// [<http://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html>]
-	if (n == 0) {
-		// Return the quantity of zero bits in `n` rather than
-		// returning that undefined value.
-		return static_cast<unsigned int>(bit_width(n));
+constexpr std::enable_if_t<std::is_integral_v<N>, unsigned int> count_leading_zeros(N n) {
+	const auto x = static_cast<std::make_unsigned_t<N>>(n);
+	constexpr decltype(x) mask{1};
```
Vultraz, line 148: You don't need to flag this constexpr. It's already inside a constexpr function. I'm not actually even sure what this would do...?

stevecotton: Making it constexpr says that I (as author) think it can be fully evaluated at compile time. I think there's no practical difference between constexpr and const here; the compiler is likely to optimise it away anyway.

That the containing function is constexpr doesn't affect this - `count_leading_zeros(n)` can only be evaluated at compile time if `n` is itself a compile time constant expression. For the callers that are handing utf8 data at runtime, the containing function might as well not be constexpr at all, IIUC.

Vultraz: Right, but in the compile-time case it doesn't make a difference.

Pentarctagon:
I don't believe there's anything left using `bit_width` at this point either?

stevecotton:
It seems the only thing still using `bit_width` is the unit tests for the code being changed in this PR. I could remove it in this PR, but feel it's better to merge it with the tests intact, and then follow up with a change-those-tests PR.

`count_leading_zeros` seems to itself only be called from `count_leading_ones` now. Changing it to just be a direct implementation of `count_leading_ones` would also be possible, but again would also mean changing the existing tests.

Pentarctagon (approved):

PR #6604, opened 2022-04-01
Scale map labels by zoom

Vultraz:
Simple UX improvement that occurred to me. Thoughts?

Before:
image

After:
image

### DIFF
--- a/src/map/label.cpp
+++ b/src/map/label.cpp
@@ -522,6 +522,11 @@ SDL_Rect terrain_label::get_rect() const
 	return rect;
 }
 
+static int scale_to_map_zoom(int val)
+{
+	return val * std::max(1.0, display::get_zoom_factor());
+}
+
 void terrain_label::draw()
 {
 	display* disp = display::get_singleton();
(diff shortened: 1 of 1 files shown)


stevecotton:
Sounds a good idea, but your two pictures suggest that the scaling ratio doesn't match the zoom ratio.

Vultraz:
Wdym.

stevecotton:
In one picture, the text is as wide as one side of the castle; in the other it's three times as wide as one side of the castle. So they seem to be experiencing different amounts of zoom.

Vultraz:
I mean... they're not, though.

Wedge009:
I'm not sure, but maybe the images themselves being different dimensions is making comparisons difficult. Perhaps it would be helpful to show a normal 100% view and zoomed-in view, for both 'before' and 'after' states.

In an isolated case like this it might be an improvement. Some scenarios are heavy on the labels, however, like the test scenario and defence of the Kalian. Of course, if labels don't overlap in 100% view, it should be fine when zoomed in as well, but it might be nice to see how those situations fare as well.

CelticMinstrel:
> I mean... they're not, though.

I mean… they are, though. That's literally the point of this change. I think @stevecotton may have missed that the two castles you displayed are the same zoom level. As @Wedge009 said it would be helpful to also show a 100% as reference.

stevecotton:
My mistake, I read it as "before zooming"/"after zooming" rather than "before this change"/"after this changed".

src/map/label.cpp
```diff
@@ -550,11 +555,12 @@ void terrain_label::draw()
 	bool use_markup = color_ == font::LABEL_COLOR;
 
 	font::floating_label flabel(text_.str());
+	flabel.set_font_size(scale_to_map_zoom(font::SIZE_NORMAL));
 	flabel.set_color(color_);
 	flabel.set_position(xloc, yloc);
```
stevecotton, line 560: Line 551's initialisation of `yloc` should have some scaling, otherwise the labels move north/south a bit as the zoom changes.

Vultraz:
Zoomed out. You're right, it gets too small.
image

Vultraz:
Made zoom go no lower than 1.0 and fixed the y offset.
image

PR #6609, opened 2022-04-04
THoT: Move unused images to attic

### LINKED ISSUE #6602: [Campaign] [THoT] Move portrait of Perrin (the elder mage) and righteous flame icon to Attic
Hmm...so, we're here again.

Backstory:

Perrin was the Headmaster of the Mage Academy encountered in the scenario "Mages and Drakes" in 1.14 THoT. The campaign was reworked and this scenario and this character has been removed. However, the portrait has remained (in 1.15, 1.16 and now, 1.17)
Thus, I propose the it's moved to the attic.

Similarly, 
the staff of righteous flame icon was part of some kamikaze or "final strike" mechanic of the classic THoT which has been removed as well. Thus, the icon should be moved to the attic as well. (might be useful for something else).

Seeing these assets still present makes players question what happened and ask about them such as https://forums.wesnoth.org/viewtopic.php?p=672546#p672546

### LINKED ISSUE #6607: Unit can become co-located with wall in NR Pursuit

- **Version:** 1.17.2
- **Downloaded from:** Uh... sourceforge, I guess.
- **Build info:**  GCC 10.2
- **OS:** Devuan GNU/Linux Chimaera,

### Description of the bug

In NR scenario "The Pursuit", when you enter the Great Chamber - if another unit happens to be on one of the two hexes which are closed in to become a wall - those  units co-locate with the wall hex, and can move to either direction.

That's probably not what you wanted...

Just be careful not to trigger the Great Chamber before moving some units onto the cave tiles which become wall.

### Expected behavior

Good question. Either create a wall/door which doesn't take up a full hex, closing the back units off, or alternatively, nudge them southwards. A cruel option would be to kill them off.

### Additional context

_No response_

Wedge009:
Resolves #6602


nemaara (approved):

Wedge009:
Oh darn, I just noticed I referenced the wrong GitHub issue in the read-me as well...

PR #6610, opened 2022-04-04
NR S05a_01 The Pursuit - Fixed units getting stuck in the wall before the spider chamber.

### LINKED ISSUE #6607: Unit can become co-located with wall in NR Pursuit

Descacharrado:
Stuck units will now be brought back (or pushed forward if hex is occupied). Also fixed the trigger. Will open backport PR ~soon~ when approved.

Closes #6607

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,8 @@
  ### Campaigns
    * Descent into Darkness
      * S08: Prevent Spectral Servant from picking up the book (issue #6576)
+   * Northern Rebirth
+     * S05a_01: Solve units getting stuck in the wall before the spider chamber (issue #6607)
  ### Editor
  ### Multiplayer
  ### Lua API
--- a/data/campaigns/Northern_Rebirth/scenarios/05a_01_The_Pursuit.cfg
+++ b/data/campaigns/Northern_Rebirth/scenarios/05a_01_The_Pursuit.cfg
@@ -1687,8 +1687,8 @@
         name=moveto
         [filter]
             side=1
-            x=10-18
-            y=1-6
+            x=10-18,11-17
+            y=1-6,7
         [/filter]
 
         [scroll_to]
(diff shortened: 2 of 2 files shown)


eyalroz:
Why not let them make it in? Do you think that would be too easy? The spider den is kind of a death-trap with only one unit.

Descacharrado:
> Why not let them make it in? 

I think that is the original idea
> Do you think that would be too easy? 

No, even with two units it is difficult.
> The spider den is kind of a death-trap with only one unit.

True, but it has a tactical use, which remains the same with these changes.


Anyways there is still a way of getting three units inside, it is just a bit harder to do now and no units can be on the wall.

Pentarctagon:
Seems like just some random failure, you can ignore it.

nemaara (approved):

Wedge009:
> > Why not let them make it in?
> 
> I think that is the original idea

I don't like the design, but yes, I think that was the original intent as well. Maybe one could try to sneak one of the respawning mages through...

Anyway, thanks for resolving this so fast. Tested and seems to resolve the issue, also is approved, so merging. I also tested an extreme case where units lined the entrance to the cave and the units in the wall get teleported appropriately.

I'll back-port after.

Descacharrado:
> I'll back-port after.

I have a backport branch ready to PR, dont worry too much about it ill open it soon.

Issue #1835, opened 2017-07-12
Allow viewing of previous [message]s in a dialogue sequence


DeFender1031:
One of the first things people tend to notice about the game is that if you accidentally skip a [message] by mistake, it is very hard to go back. A simple log of messages which has been shown could be kept, and navigation buttons similar to those in [story] texts could be added to [message] popups to allow navigation through previous dialogue. (The arrow keys should probably perform this operation as well.) Additionally, a menu option and hotkey could be added to bring up the most recent message and allow navigation from there when there isn't actively one on the screen. I would also propose that previous messages use a somewhat more subdued style or color scheme to visually differentiate between something that was said previously and something which is being said right now.

CelticMinstrel:
Given the way each message is logically independent, this might be rather difficult to do, but it does seem like a good idea.

ProditorMagnus:
Other than navigation, it can be easily done with Lua by changing wml_actions.message to save message, and with different tag display saved ones.

DeFender1031:
@CelticMinstrel, all it needs to a history stored somewhere of the messages that have already shown.

@ProditorMagnus, of course, but only accessible via a menu after a sequence is complete would not be the most friendly interface. If someone accidentally skips a message, they don't want to run the rest of the sequence before seeing what they missed, they want to go back immediately.

CelticMinstrel:
I was mainly referring to the arrow keys to navigate between messages on-the-fly. Either way it's certainly doable, though.

DeFender1031:
Ah, I see. Well, the arrow keys themselves already exist. Seems like it'd just be a matter of adding them to the message dialog and setting them up to navigate along the stored history, with "forward" on the current message special cased to mean "close message dialog and continue"...

CelticMinstrel:
Something like that. It's probably not quite as simple as that makes it sound, but that outlines the gist of what would need to be done.

DeFender1031:
Seems like nothing ever is with this codebase...

cooljeanius:
Many other games have features like this; I can upload screenshots of how it looks elsewhere if people would like

cooljeanius:
I think I saw that @irydacea was working on such a feature for IftU

PR #6614, opened 2022-04-06
Tutorial S02: Reduce the number of [messages]

stevecotton:
Drop many of the flavortext messages, to focus on the ones that explain
mechanics or tactics to the player.

The message about the Status Table gets removed because the player is unlikely
to benefit from it in this scenario. It might fit near the end as "you can
check how much gold the enemy has, and guess if he'll recruit next turn", but
otherwise it's just "hey, you could look at all the menu items" without any
suggestion about why the player should use it now.

Probably of interest to @max-torch too. The file is the same in 1.16 as master, so
could be backported.

### DIFF
--- a/data/campaigns/tutorial/scenarios/02_Tutorial_part_2.cfg
+++ b/data/campaigns/tutorial/scenarios/02_Tutorial_part_2.cfg
@@ -198,29 +198,7 @@
                 {CONTINUE_MSG}
         [/message]
 
-        [message]
-            speaker=student
-            message= _ "Ho, Galdrad! Has Delfador conjured something else to beat me with? A flock of scarecrows, perhaps?"
-            female_message= _ "female^Ho, Galdrad! Has Delfador conjured something else to beat me with? A flock of scarecrows, perhaps?"
-        [/message]
-
-        {GENDER (
-            [message]
-                speaker=Galdrad
-                message= _ "This is no game, Konrad! Orcs have encamped across the river. This is elven country; they are fools to enter here. We elves are fast and hard to hit in forests. You must defeat their leader so they never threaten us again. I will advise you."
-            [/message]
-        ) (
-            [message]
-                speaker=Galdrad
-                message= _ "This is no game, Li’sar! Orcs have encamped across the river. This is elven country; they are fools to enter here. We elves are fast and hard to hit in forests. You must defeat their leader so they never threaten us again. I will advise you."
-            [/message]
-        )}
-
-        [message]
-            speaker=student
-            message= _ "What should I do?"
-            female_message= _ "female^What should I do?"
-        [/message]
+        # Maybe there should be an intro for Galdrad, but there's a lot of messages in here anyway; it seems better to err on the side of "every message says something useful".
 
         {TALK_ABOUT Dumbo ( _ "First, we will have to deal with the Orcish Grunt stationed in the middle of the river. He should be little trouble.")}
 
@@ -229,27 +207,6 @@
         {TALK_ABOUT_LOC 15,4 ( _ "See this dark blue water? It’s too deep for either side to cross. The orcs could slowly wade through that narrow band of shallow lighter-blue water in the north; but we could stand on the shore and force them to fight us from the water, where they are exposed and we are protected by the forest.")}
 
         {TALK_ABOUT_LOC 17,11 ( _ "The more likely attack, then, is across the bridge. That middle island is the key: it has a village for healing injured units and forests in which we fight so well.")}
-
-        [message]
-            speaker=student
-            message= _ "Let’s go! Attack!"
-            # po: The "us" in "let's" refers to Galdrad and Li'sar.
-            female_message= _ "female^Let’s go! Attack!"
-        [/message]
-
-        {GENDER (
-            [message]
-                speaker=Galdrad
-                # po: Addressing Konrad
-                message= _ "Hang on! You need to gather your forces. Or do you intend to fight the orcs single-handedly?"
-            [/message]
-        ) (
-            [message]
-                speaker=Galdrad
-                # po: Addressing Li'sar
-                message= _ "female^Hang on! You need to gather your forces. Or do you intend to fight the orcs single-handedly?"
-            [/message]
-        )}
 
         [if]
             # Do we have a recall list at all? It’s possible that the player lost all their units in the first scenario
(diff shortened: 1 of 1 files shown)


knyghtmare:
I see two sides of this:

1. Even if it's a tutorial, it's still a campaign and should have some story in it. I like how the tutorial has been done. Makes you feel immersed in the game.

2. Not all Players would read the dialogue and prefer to skip, missing out on key instructions which has been skipped mistakenly on as scenario message text.

Wedge009:
What prompted this? I'm definitely of the mind-set that appreciates immersion and story-telling in games (eg the approach Origin Systems took with their motto 'We Create Worlds'). It's something that drew me in when I first played in 1.4.x, and I think it gives newcomers a taste of how the regular campaigns play. Just focusing on the mechanics would be very dry to me.

Looking at #6615, the overlay text looks nicer than the current chat-style printing, but how long does it last for (eg does the player have any control over how long it stays overlaid or is it displayed long-term)? Does that enhancement really make culling the text here necessary? If I'm allowed to have a preference I would keep the Tutorial text largely as-is, we've already culled other things because they were subjectively unsatisfactory.

nemaara:
I'm also hesitant to remove the flavor text completely. Making it more concise is a possibility, but altogether axing probably would make it feel too bland for people who want to read it, I think. The problem originally was that information text was also interspersed within the flavor text so that people would just skip everything. I think in PR #6615 if those get converted to overlays that will help people easily identify the flavor text vs. informational text.

stevecotton:
> What prompted this?

In no particular order
* the discussion on Discord on March 24th
* my personal dislike for the way that (mainly in S01) the characters sound like they don't want to do this tutorial - the protagonist that the player should sympathise with is saying to change things, and that doesn't seem to fit with our intention that the player follows the instructions
* watching some other games' tutorials and thinking "theirs is too long, does the same apply to ours"
* wanting to try the overlays out

> the overlay text looks nicer than the current chat-style printing, but how long does it last for (eg does the player have any control over how long it stays overlaid or is it displayed long-term)?

At the moment the player has no control over how long it's displayed, it simply stays there until the next message replaces it. Some of this is to think over and get more feedback about #5837.

CelticMinstrel:
I'm not exactly on board with removing all the flavour text either (mainly Galdrad's introduction).

Wedge009:
>     * the discussion on Discord on March 24th

If I read the right part: I think generally there will always be people who skip over stuff - whether deliberately or accidentally - and miss crucial details. Perhaps it's also a generational aspect - not that I play a lot of commercial games, but I imagine tutorials in the current generation focus a lot on the immediate detail in the interests of getting players 'into the action' as quickly as possible.

>     * my personal dislike for the way that (mainly in S01) the characters sound like they don't want to do this tutorial - the protagonist that the player should sympathise with is saying to change things, and that doesn't seem to fit with our intention that the player follows the instructions

To me that was part of the story-telling or character-building. It could also be from their previous design as more anime-style characters and this aspect hasn't yet been updated to reflect their more 'realistic' portrayals. Regarding how they 'say to change things', I don't really recall that impression - is that relating to things like the character wanting to strike the quintain with a melee attack vs ranged attack? I thought that was a good way to learn to distinguish between the options and trying it the 'wrong way' quickly teaches the player a lesson about planning attacks well.

>     * watching some other games' tutorials and thinking "theirs is too long, does the same apply to ours"

Something that isn't easy to quantify. What's too long for some may be not enough for others. Like in the aforementioned #wesnoth-dev discussion, there seemed to be some talk about what's the best post-Tutorial campaign to try, which may offer some more information about various tactics and finer details of the battle mechanics. If players 'get bored' of the Tutorial as-is that quickly, maybe Wesnoth isn't going to be in their interests anyway. Certainly I don't think the first scenario is that long, but the second one could be quite long for a new player - I remember taking a while to get through it but replaying it recently as an experienced player, it was a 'breeze'. But that length was all part of the learning experience and I would have had experience that anyway, whether in the Tutorial itself or one of the introductory campaigns.

stevecotton:
> Regarding how they 'say to change things', I don't really recall that impression - is that relating to things like the character wanting to strike the quintain with a melee attack vs ranged attack?

I mean the run up to "You want me to fight a dummy?" in S01, the sarcasm of "perhaps we should have started with a doll", "maybe I can fight in a real battle next". Then S02 opens with "Has Delfador conjured something else to beat me with? A flock of scarecrows, perhaps?"

max-torch:
> there will always be people who skip over stuff
My solution to the skipping over of crucial dialog would be to create a 'dialog log' feature. You can open it to see the previous dialog that you failed to read or forgot about.

Wedge009:
#1835. It's not an uncommon feature.

That was the other thing I was going to mention - whether or not any such changes to the message system may add complications to any sort of text log.

stevecotton:
Thanks for the feedback, I've changed the PR to a draft to reflect that it's not going to be merged.

I'm wondering about giving the player an option dialog when they first move next to Delfador (after the character selection, but before the "Good morning Delfador" message):
```
Do you want characters and story, or just a tutorial on how to play?
* I want the characters to talk to each other
* Just teach me the mechanics
```

Wedge009:
I thought about such an option too. I think it would be quite an unusual thing to do, if implemented. I think that most games that have this sort of dedicated tutorial make a creative decision one way or another - either make it in-universe but for the benefit of new players, or keep it purely instructional. Still, if you think that would be the best way to satisfy both kinds of audiences, I wouldn't object to it.

nemaara:
I don't think that's a necessary option. We should make a decision either way without asking players and for now I would say we shouldn't remove the text, only make it shorter if possible.

CelticMinstrel:
Note: I think #6615 is still worth considering independently of this.

CelticMinstrel:
> how long does it last for (eg does the player have any control over how long it stays overlaid or is it displayed long-term)?

Apparently no-one got around to answering this. The text stays until it's replaced by a new message; we _could_ limit its duration but in this particular case I don't see any reason to do so.

CelticMinstrel:
> That was the other thing I was going to mention - whether or not any such changes to the message system may add complications to any sort of text log.

Either this would exclude the hint messages from such a log (which doesn't really feel _wrong_ to me), or the hint messages would push their text into the same log area (it would probably be some sort of API like `wesnoth.interface.add_to_message_log` or something).

PR #6615, opened 2022-04-06
Proof of concept: Convert the tutorial's wall of text [message]s into overlay text

stevecotton:
The reason for opening this is to ask if it's a direction that the UI should go in.

Adds the implementation of a prototype [hint_message] tag, either using PR #5837's add_overlay_text feature, or a fallback of outputting them to the chat log. This means a few less dialogs for the user to click through, but mainly means that about one hint per turn stays visible even when skipping through the dialogs.

The first of the two commits is PR #6614.

Turn 1's hint about upkeep appears near the end, when the player captures a village.

Turn 2's hint appears in the start-of-turn dialogue, when commenting on the orc blocking the bridge.

Turn 3 onwards appear at the start of each turn.

Initially I tried the bottom of the screen, but that overlaps with the area for `[message]`s.

With PR #5873 merged too:
tutorial_add_overlay

Without that PR:
tutorial_without_add_overlay

### DIFF
--- a/data/campaigns/tutorial/scenarios/02_Tutorial_part_2.cfg
+++ b/data/campaigns/tutorial/scenarios/02_Tutorial_part_2.cfg
@@ -198,29 +198,7 @@
                 {CONTINUE_MSG}
         [/message]
 
-        [message]
-            speaker=student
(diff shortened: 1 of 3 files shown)


stevecotton:
Closing and handing over the feature to @CelticMinstrel.

PR #6616, opened 2022-04-08
Avoid using unit IDs that are no longer valid

### LINKED ISSUE #6603: wesnoth 1.16.1 crashes when loading campaign SotBE-Backhome

- **Version:** 1.16.1
- **Downloaded from:** alpine linux packages
- **Build info:**
- **OS:** alpine linux

### Description of the bug

Start wesnoth, play until  SotBE-Backhome. Enter the mission. The initial dialogue works. Just after pressing space when Grüü gets his money wesnoth segfaults:

```
% wesnoth
Battle for Wesnoth v1.16.1 x86_64
Started on Thu Mar 31 23:13:34 2022


Data directory:               /usr/share/wesnoth
User configuration directory: /home/USER/.config/wesnoth
User data directory:          /home/USER/.local/share/wesnoth/1.16
Cache directory:              /home/USER/.cache/wesnoth

Setting mode to 1920x1080
20220331 23:13:35 error general: locale set-up for icompare() is broken, falling back to std::string::compare()
20220331 23:13:35 error general: has backend: 'posix',has backend: 'std', locale: (name='c.UTF-8' country='' language='c' encoding='utf-8' variant=''),generator categories='131095'
Checking lua scripts... ok
zsh: segmentation fault  wesnoth
```

This is reproducible and every time re-loading the mission it segfaults again.

1. Play until sotbe-backhome
2. Skip through the dialogue
3. See it segfault

### Expected behavior

No segfault :)

### Additional context

This used to work in 1.14.x before

Wedge009:
Patch originally from @stevecotton. Aiming to merge before 1.16.3 is tagged (supposedly scheduled for 17th April 2022).

Resolves #6603 (crash with musl implementation of libc)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -22,6 +22,7 @@
  ### WML Engine
  ### Miscellaneous and Bug Fixes
    * Reduced the size of Isar's Cross map background images.
+   * Resolve crash on systems using musl implementation of libc (issue #6603)
 
 ## Version 1.17.2
  ### Add-ons client
--- a/src/recall_list_manager.cpp
+++ b/src/recall_list_manager.cpp
@@ -56,8 +56,10 @@ unit_const_ptr recall_list_manager::find_if_matches_id(const std::string &unit_i
  */
 void recall_list_manager::erase_if_matches_id(const std::string &unit_id)
 {
+	// using unit_id as reference has potential to cause a crash if the underlying unit becomes invald
+	// https://github.com/wesnoth/wesnoth/issues/6603
 	recall_list_.erase(std::remove_if(recall_list_.begin(), recall_list_.end(),
-		[&unit_id](const unit_ptr & ptr) { return ptr->id() == unit_id; }),
+		[unit_id](const unit_ptr & ptr) { return ptr->id() == unit_id; }),
 	                       recall_list_.end());
 }
 


Wedge009:
Just trying to understand the issue - this is a nasty bug. Is it because `erase_if_matches_id` is accepting a reference to the ID, and so when that unit is removed the ID is no longer valid? Is it the case that musl's implementation is more 'correct' or strict such that the bug is exposed using that implementation, whereas GNU's is more forgiving?

I checked that these are the only two uses of `erase_if_matches_id` in the file and the only cases where the parameter is from a unit object. Other uses of the method use a `std::string` instance as the parameter.

Regarding the existing comment 'Should it use underlying ID instead?' - I'm not sure what it means by that. If it's referring to the ID string instead of the ID from the unit object, maybe that comment should be deleted as well?

CelticMinstrel:
I mentioned this on the issue as well, but I would personally prefer to instead alter the `erase_if_matches_id` function to take the ID string by value (probably with a comment to explain why, so someone doesn't change it back later without realizing).



> Is it because `erase_if_matches_id` is accepting a reference to the ID, and so when that unit is removed the ID is no longer valid?

That appears to be the situation, yes.

Wedge009:
Oh, I overlooked that as I'm only just now starting to understand the issue.

https://github.com/wesnoth/wesnoth/blob/650f70405fffbe60648f24f8dce061575d689538/src/recall_list_manager.cpp#L57-L62
So just remove the `&` from the parameter (with explanatory comment)?

https://github.com/wesnoth/wesnoth/blob/650f70405fffbe60648f24f8dce061575d689538/src/scripting/game_lua_kernel.cpp#L2502-L2504
What about here, where it looks like the intention was to use `v->id()` as the search parameter. If the function is changed to accept string as value, would it make the cloning and reassigning between `u` and `v` redundant?

stevecotton:
> I mentioned this on the issue as well, but I would personally prefer to instead alter the `erase_if_matches_id` function to take the ID string by value (probably with a comment to explain why, so someone doesn't change it back later without realizing).

I'm wondering if that would still be undefined behavior - the copy in the callee is fine, but is the one in the caller still meant to be valid?

CelticMinstrel:
> it isn't clear to me what the reason for cloning the unit is, other than possibly to avoid the pass-by-reference issue we're now aware of.

Well, it's in `extract_unit`, the purpose of which is to remove a unit from the map while keeping a Lua-private reference to it. So if you assume removing it from the map destroys it, I think it makes sense that it's being cloned.

Wedge009:
Oh okay, when you explain it like that it makes more sense. I don't know if there's an established convention with `u` and `v`, but the single-letter variable names didn't really help me to see that and I didn't know that `u` was somehow still used after the function.

Still no word from the original reporter - maybe we'll have to miss the 1.16.3 tag if they don't respond in time?

Wedge009:
I'll update the PR either later today or tomorrow.

PR #6617, opened 2022-04-08
SoF: Consistently use lower-case after ellipses

stevecotton:
Sadly, this takes the campaign up to 100 string changes between 1.16.2 and 1.16.3. The only one of these that was already going to change for 1.16.3 is S04t's volcanic forge one.

My intention (whether or not this merges) is to send out a mail to the i18n list when Saturday's pot-update is done, identifying the [4 strings that are actual changes](https://r.wesnoth.org/p672656), and saying that all the other fuzzies are just en_US grammar changes.

I think all of these make sense as mid-sentence pauses, but a few were arguably starting a new sentence so the ellipse was acting as both an ellipse and a full stop.

Pings: @max-torch. I'd ping demario if I knew his Github username.

### DIFF
--- a/data/campaigns/Sceptre_of_Fire/scenarios/2_Closing_the_Gates.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/2_Closing_the_Gates.cfg
@@ -509,7 +509,8 @@
         [/message]
         [message]
             speaker=Baglur
-            message= _ "I think it can be sealed up somehow... Yes, look, activating that glyph seems to have closed up the gap."
+            # po: "it" is a side entrance a dwarven fortress
+            message= _ "I think it can be sealed up somehow... yes, look, activating that glyph seems to have closed up the gap."
         [/message]
     [/event]
 
--- a/data/campaigns/Sceptre_of_Fire/scenarios/2p5_Reaching_the_Runecrafter.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/2p5_Reaching_the_Runecrafter.cfg
@@ -162,7 +162,7 @@
         [/message]
         [message]
             speaker=Pidmer
-            message= _ "Um... Aren’t we needed for the salvage?"
+            message= _ "Um... aren’t we needed for the salvage?"
         [/message]
         [message]
             speaker=Bragdash
@@ -175,7 +175,8 @@
         {MOVE_UNIT id=Rugnur 13 27}
         [message]
             speaker=Rugnur
-            message= _ "... Let’s set up a base here."
+            # po: second part of "Uh, well, we’re not getting through without a fight..."
+            message= _ "... let’s set up a base here."
         [/message]
         [terrain]
             terrain=Ce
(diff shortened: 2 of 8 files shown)


knyghtmare:
> I'd ping demario if I knew his Github username.

@demario-w

Wedge009:
I still say we shouldn't let translation overhead preclude changes from being made where it makes sense to make them. Perhaps there's an argument to be made that this change isn't strictly necessary as opposed to, say, spelling or grammatical corrections. That aside, it is a good idea as courtesy to notify the translators of what's happening here.

I did a quick review and I don't see any changes I would object to. If it wasn't the weekend for me, I would have a more thorough read. Curiously, the [wiki](https://wiki.wesnoth.org/Typography_Style_Guide) suggests that if an end of sentence was intended, then a full stop would come after the ellipsis and since that guide dictates three full stops over the dedicated ellipsis character, it seems to suggest that there would be four full stops for an ellipsis at the end of a sentence, though I probably wouldn't follow that in practice.

Is it worth having nemaara's opinion on this? I didn't think I'm necessarily the one to give approval to prose changes.

Wedge009 (approved):

nemaara:
Haven't read through it but did glance. In general this is not a change to be making because ellipses sometimes represent a pause within a sentence. It depends on whether you treat the thing following the ellipses as a new sentence or not, which requires looking at the messages.

stevecotton:
While I've described it as "for consistency", it wasn't a blind search-and-replace. I think 13 should definitely should be lower case, 6 are arguable either way, and none of them should definitely be upper case.

There's a further 10 strings in the campaign that I didn't change - 8 because the next word was "I", one for a proper noun, and one that includes a question-mark before the ellipses.

nemaara (approved):
Tangentially, I think this campaign probably uses too many ellipses, do you want to get rid of some of them in this PR or just merge and deal with it later?

stevecotton:
I'd prefer to just merge, I don't want to commit myself to making further changes.

nemaara:
Merge away or I'll do it after work today. Supposing this also should go to 1.17?

Wedge009:
It was marked to be forward-ported, haven't yet checked if that's been done.

stevecotton:
Forward-ported in c7acca4ea77525fa97a0049b556acb4e67177544

Wedge009:
@sheepo99 Are you still working on Portuguese translation? Sorry i don't know how else to contact you, but just wondering if you have thoughts on the [aforementioned forum proposal](https://r.wesnoth.org/p672656). There is concern about the rate at which source strings are changing, but I'm not sure if the proposal to create a 'simple English' version of the prose is a good solution, if there is even a solution possible.

sheepo99:
Also forgot to reply to the main question, lol. Yes, you totally need an en_US translation file to keep translations stable.

knyghtmare:
> Finally, and on a personal note, this is also a consequence of the excessive amount of official campaigns in the core Wesnoth package. Problems like this would be a lot easier to manage if there was less content with more focus on quality.

Got any suggestions on which one to push to the chopping block?

knyghtmare:
Good answer. Feel free to open a new discussions thread here or on forums.

Wedge009:
> Problems like this would be a lot easier to manage if there was less content with more focus on quality.

I understand that's the intention, but obviously there will be many changes to go through before reaching that point. Bit of a chicken-or-egg problem.

sheepo99:
I don't see it as a being a chicken-or-egg issue if part of the solution is precisely to put the eggs in a different basket (namely the DLC server instead of the core package). xD

PR #6619, opened 2022-04-09
Add debug for "[event] is missing name or id"

stevecotton:
The problem with this warning was that it didn't include any clue about what
had triggered it - just that there's a broken event somewhere, and it could be
in a scenario, macro, unit_type or even from another add-on if you're using a
non-default era.

When run with `--log-debug=wml`, this makes Wesnoth print out whatever the
contents of the affected [event] tag are.

See https://r.wesnoth.org/t55584 for more details.

### DIFF
--- a/src/game_events/manager_impl.cpp
+++ b/src/game_events/manager_impl.cpp
@@ -34,6 +34,7 @@ static lg::log_domain log_event_handler("event_handler");
 
 static lg::log_domain log_wml("wml");
 #define ERR_WML LOG_STREAM(err, log_wml)
+#define DBG_WML LOG_STREAM(debug, log_wml)
 
 namespace game_events
 {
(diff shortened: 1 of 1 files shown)


knyghtmare:
[Dwarvish_Kingdom.zip](https://github.com/wesnoth/wesnoth/files/8459063/Dwarvish_Kingdom.zip)
This abandoned add-on by Ashmyr is a known-case for this "[event] is missing name or id"
Let's see if this PR works at debugging the actual cause of it.

stevecotton:
> Dwarvish_Kingdom.zip

Turns out it's the same cause as Spannerbag's campaign had - an `EXTRA_SCENARIO_MUSIC` in the first scenario. Debug output is:
```
20220410 16:30:20 error deprecation: [advancefrom] has been deprecated and removed.; Error: [advancefrom] no longer works. ‘Troll Shaman’ units will not be able to advance to ‘Troll Fire Wizard’; please ask the add-on author to use [modify_unit_type] instead.
20220410 16:30:24 error wml: [event] is missing name or id field
20220410 16:30:24 debug wml: Content of that event:
[music]
	append = yes
	name = underground.ogg
[/music]
```

knyghtmare (approved):

stevecotton:
Forward-ported in bfd0ef9af7aace4d753521fb29e9ad7b9c9419c7.

PR #6620, opened 2022-04-10
Make tutorial hint messages persistent

CelticMinstrel:
Based on @stevecotton's #6615, this moves many of the tutorial hint messages into persistent overlay text. I played through the tutorial with these changes and I think it seems pretty good, but it probably needs a little refinement.

### DIFF
--- a/data/campaigns/tutorial/lua/hint_message.lua
+++ b/data/campaigns/tutorial/lua/hint_message.lua
@@ -1,0 +1,24 @@
+
+local hint_message = {valid = false}
+function wesnoth.wml_actions.hint_message(cfg)
+    if hint_message.valid then
+        hint_message:remove()
+    end
+    if cfg.message and not cfg.remove then
+        local text = cfg.message
+        if cfg.caption and cfg.caption ~= "" then
+            text = "<b><big>" .. cfg.caption .. "</big></b>\n" .. cfg.message
+        else
+            text = cfg.message
+        end
+        hint_message = wesnoth.interface.add_overlay_text(text, {
+            size = 18,
+            location = {5,5},
+            color = {255, 255, 255},
+            duration = "unlimited",
+            max_width = "40%",
+            valign = "top",
+            halign = "left"
+        })
+    end
+end
--- a/data/campaigns/tutorial/lua/hint_message.lua
+++ b/data/campaigns/tutorial/lua/hint_message.lua
@@ -9,8 +9,6 @@ function wesnoth.wml_actions.hint_message(cfg)
         hint_text = cfg.message
         if cfg.caption and cfg.caption ~= "" then
             hint_text = "<b><big>" .. cfg.caption .. "</big></b>\n" .. hint_text
-        else
-            hint_text = cfg.message
         end
         hint_message = wesnoth.interface.add_overlay_text(hint_text, {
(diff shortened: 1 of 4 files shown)


CelticMinstrel:
Here's an example of what it looks like.

Wedge009:
I haven't tried this but as with #6615 in principle it sounds like a good idea. It probably addresses the concerns mentioned in #6614 where the instructional part of the dialogue is missed amongst the 'flavour' text that some players may skip over.

data/campaigns/tutorial/lua/hint_message.lua
```diff
@@ -0,0 +1,32 @@
...
+function wesnoth.wml_actions.hint_message(cfg)
+    if hint_message.valid then
+        hint_message:remove()
+    end
+    if cfg.message and not cfg.remove then
+        hint_text = cfg.message
+        if cfg.caption and cfg.caption ~= "" then
+            hint_text = "<b><big>" .. cfg.caption .. "</big></b>\n" .. hint_text
+        else
+            hint_text = cfg.message
```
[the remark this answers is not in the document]
CelticMinstrel: Good point, thanks for spotting that.

stevecotton:
Some of the message timings should change - for example, S01's `caption="Recruiting"` message currently displays when you move to the keep, but would fit better if it appeared at the start of the lesson, when the "I'll recruit some elves" message appears.

Some text is looking a bit cut-off (this was in a 1280x1024 Wesnoth window), only noticed it on this message.
cutoff_overlay_text

data/campaigns/tutorial/scenarios/02_Tutorial_part_2.cfg
```diff
@@ -287,12 +287,12 @@
             [/else]
         [/if]
 
-        [message]
+        [hint_message]
             speaker=narrator
             caption= _ "Recalling"
             image=wesnoth-icon.png
```
stevecotton, line 293: Why doesn't the schema validation trigger for the leftover `speaker` and `image` here?

CelticMinstrel: The `[event]` tag is set to allow any child tag inside it. The effect is that the schema validates known ActionWML tags and ignores unknown ones. The reason it's defined this way is so that adding a custom tag doesn't automatically break the schema.

Pentarctagon:
So aside from the width issue, this seems good to merge? The message timings seems unrelated to this PR's purpose though, since it's changing how a message is displayed rather than when.

CelticMinstrel:
Yes, I don't know of anything to block merging this. The width issue would be an issue in the API, not in this change. It _would_ be nice to get a review from someone other than stevecotton though.

Vultraz (approved):

data/campaigns/tutorial/scenarios/02_Tutorial_part_2.cfg
stevecotton: Still, they're leftovers that you should fix.

Pentarctagon:
@CelticMinstrel So should I click merge, or were you planning to?

CelticMinstrel:
Sure, I can do it.

Issue #1345, opened 2017-05-08
Alternate direction for "stones with sand drifts" (GNA #21018)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?21018)** by tekelili on 2013-08-11

I have found several times that embellisment "stones with sand drifts" is not always looking cool due to "face". Currently only west-east sand drifts are alllowed, a north-south variation would be helpfull :)

(Reproduced on windows)
Release: 1.10.6
Priority: 5 - Normal
Severity: 1 - Wish

Wedge009:
Not sure what is meant by 'face' here...

AI0867:
The facing of the tile, as it's directional. It sounds like this particular terrain would benefit from /\| variations.

Wedge009:
Ah, so there's only limited options for the orientation of the tile.

doofus-01:
Even if there are complaints about the new images, the situation is changed enough (in 1.17.3+) that this can be closed.

PR #6626, opened 2022-04-17
1.17 - minor terrain updates

doofus-01:
`Rrd` graphics are mostly placeholder, the graphics rules were the point for now.
Addresses #1345 - not by making \, |, / versions, but just futzing with the images.
Screenshot_20220417_
I'm not sure about the road color, but it's really easy to change for now.


doofus-01:
Screenshot_20220417_b
The bright white snow makes some of the shortcomings of the existing bridges more obvious, will need to be fixed in this PR, at least for `^Bsa*`.
The general bridge-to-castle transition scheme also needs work, not sure if this will be fixed here.

doofus-01:
I think I introduced a bug in #6606 by fixing a bad image reference, and solved it by commenting out a rule that used to always fail, but I may be missing something, so I didn't remove them.  It seems to work now, but more testing is always needed.

Issue #2219, opened 2017-11-19
High DPI (hidpi) support


dhardy:
Wesnoth is too small to be playable using my laptop's native screen resolution (276 DPI). It would be nice to have the option for pixel doubling or scaling of individual elements.

Currently there is a text-scaling option but it's very buggy (layout not adjusted so lots of clipping occurs), and maxes out at 150% (still small).

sevu:
There's a zooming function too (I use this) with I think "ctrl" + "+", or ctrl + the button where on American layout "+" is.
However, it's buggy too.

Vultraz:
HighDPI support sadly isn't as simple as flicking a switch. I've tried some basic implementations, but none have been completely successful.  Your best bet right now is to use OS app scaling.

CelticMinstrel:
It may not be as simple as flicking a switch, but a simple "draw everything scaled up" option isn't that difficult. I recently added it to another project, and if I recall correctly it only took me abotu a day to do it.

(Mind you, it's probably a bit harder in Wesnoth because it's a larger project, and it also uses SDL rather than SFML which means I'm not as sure how to do it.)

renom:
Any plans to implement this?

vgaming:
Hi, I just recently started using a 4k monitor, and I'm confirming it's a pain:
![2020 06 11_14:32:17ee](https://user-images.githubusercontent.com/33452702/84386112-10782a00-abf1-11ea-96b6-74a7fa14343c.png)
![2020 06 11_14:32:41ee](https://user-images.githubusercontent.com/33452702/84386120-13731a80-abf1-11ea-8960-432a3585de6c.png)
![2020 06 11_14:33:12ee](https://user-images.githubusercontent.com/33452702/84386125-14a44780-abf1-11ea-989e-f4a4e037565e.png)

The screenshots are made on a 3840x2160 resolution, downscaled twice so one can easily see it on a 1080p monitor.

On the last picture, you can zoom in the battlefield as @sevu correctly mentions by pressing `-` and `=`, but the chat and terrain labels will not be scaled by these keyboard shortcuts.

I wasn't able to find "text scaling" configuration at all. All pictures are from wesnoth-1.14.12.

vgaming:
If there are any HiDPI questions or tests that needs to be run - feel free to ping me, I'll try it out.

Pentarctagon:
The dpi changes that have been done by @gwhitney so far are on the master branch only, so wouldn't be in 1.14.

[...]

PR #6629, opened 2022-04-18
Add a pixel scale multiplier option, enabling integer scaling of the game to higher resolutions.

mesilliac:
I found the game unplayable on my 3840x2160 pixel display, so... i fixed it.

The primary change:
 * A new pixel scale multiplier option, ranging from 1 to 4, which scales the game by that amount before display.
 * An automatic option, that sets pixel scale to something reasonable based on window size.

What i changed to make this work:
 * The window framebuffer has been replaced (for game purposes) by a separare `SDL_Surface` for drawing onto. Everything that previously drew directly to the window framebuffer now draws to this surface.
 * The window is now rendered via texture copy. SDL handles this. I left it in software mode for the actual rendering, but theoretically hardware accelleration could be enabled as available simply by removing this restriction. SDL should choose correctly according to system.
 * When a screen flip is requested, the drawing surface the game now uses is copied manually to a rendering texture. SDL then scales this texture to the actual final output rendering surface, which Wesnoth does not currently need to care about.
 * Resize events / framebuffer updates now verify and if needed create the drawing surface and rendering texture. The rest is handled by SDL.
 * SDL's built-in capability for scaling the render texture (`SDL_RenderSetLogicalSize`) is used to automatically convert input events to the same scale as the drawing surface. Some tweaks were needed where Wesnoth overloads SDL events, particularly the window resize event, which must remain in display coordinates. This is converted to drawing coordinates in the relevant places to make everything work without further change to existing drawing or input handling.
 * a new `sdl/input.cpp` was added wrapping routines to return the mouse positon, as SDL returns this in display coordinates, not draw-space coordinates. The new `sdl::get_mouse_state` must be used in place of `SDL_GetMouseState`, but this is the only important change code-wise.
 * Some `CVideo` functions were renamed to make the coordinate space explicit. `GetSurface` is now `GetDrawingSurface`, and `screen_area` is now `draw_area`. `get_width` and `get_height` are left as-is, as from Wesnoth's point of view input coordinates are still the same as draw coordinates. Some new functions were added to get the various different draw/display/input/high-dpi-output sizes.
 * _The minimum allowable window height has been reduced from 600 to 540 pixels!_ This is so that resolutions above a new "maximum" window size of 1920x1080 can be scaled by 2. In my tests this works adequately, even with the existing 800x600 theme, but ideally a 960x540 theme should be created. 1920x1080 looks to me to be about the maximum workable resolution for Wesnoth, and for it to be scalable without fractional scaling (which looks bad), a size half-the-maximum needs to be supported. I set the maximum to 1920x1080 in `preferences/general.cpp`.
 * A few minor auxiliary functions were added. Mostly to get the preferences screen to automatically update the pixel scale when changed.

Some notes:
 * The pixel scale will never reduce beyond the minimum resolution indicated in `preferences/general.cpp`. This was previously 800x600 but i changed it to 800x540 for the reason mentioned above, and so that users running 1920x1080 can use a pixel scale of 2 if they explicitly choose to.
 * The automatic option will attempt to scale to the default resolution indicated in `preferences/general.cpp`. This is 1280x720 and is unchanged. It will try not to scale below this.
 * The automatic option will attempt to scale to ensure the (new) maximum resolution indicated in `preferences/general.cpp` is not exceeded. I have set this to 1920x1080, which feels reasonable to me, but it can easily be changed. If needed, this _will_ scale below the default resolution, but will not scale below the minimum resolution.

Potential issues with the new system:
 * SDL explicitly states that the rendering interface is not safe for multi-threaded use. Judging from what i saw of the Wesnoth codebase this may be a problem. There already seemed to be some race conditions with the old code where the game would crash if the framebuffer wasn't updated quickly enough. These problems are probably still there. In any case there should be only one routine that meaningfully touches this system - the `CVideo::flip()` function.

I have played several scenarios and browsed a variety of game systems using pixel scales of 2 and 3 at various window sizes, and everything i tested appears to work without issue.

I tested enabling hardware acceleration and that _did_ have issues. So i left it in software mode for now, which should behave roughly identically to the previous system.

And because i like pictures, here is the difference in fullscreen output at 4K resolution (3840x2160), without and with pixel scaling.
wesnoth-4k
wesnot-4k-scaledx3


mesilliac:
The failing mac build seems to just not be linking the new `sdl/input.cpp`. I added it to `source_lists/wesnoth`, but i guess there's somewhere else it needs to be added?

cooljeanius:
Nice! Would this let us close #2219 (HDPI support) or would there still be things to do for that?

src/gui/widgets/settings.cpp
```diff
@@ -51,13 +51,16 @@ std::vector<game_tip> tips;
...
+	const SDL_Rect rect = video.draw_area();
 
 	screen_width = rect.w;
 	screen_height = rect.h;
 
-	auto [scalew, scaleh] = vid.get_dpi_scale_factor();
+	// This should, perhaps, include the pixel scale multiplier.
+	// However i'm not sure what it is intended to be used for,
+	// and it doesn't appear to be currently used anywhere.
+	auto [scalew, scaleh] = video.get_dpi_scale_factor();
```
vn971, line 63: The variable seems to be used for add-ons, e.g. we set a WML variable: https://github.com/wesnoth/wesnoth/blob/master/src/gui/widgets/helper.cpp#L104

We should probably also make sure that the settings can be "live-updated" without restarting Wesnoth. Meaning that the WML variable gets updated on Scale change. I'll test it out after building. (I'd also easily see it as something that could be finished up later if it's not trivial, as having HiDPI support in itself is already great.)

src/hotkey/command_executor.cpp
```diff
@@ -33,6 +33,7 @@
 #include "show_dialog.hpp"
 #include "../resources.hpp"
 #include "../playmp_controller.hpp"
+#include "sdl/input.hpp" // get_mouse_state
```
vn971, line 36: Is that standard style in wesnoth? (I honestly have no clue, but I haven't seen it before)

cooljeanius: if it isn't already standard style, it ought to be, IMO

src/sdl/window.cpp
```diff
@@ -157,13 +166,41 @@ uint32_t window::get_flags()
 void window::set_minimum_size(int min_w, int min_h)
 {
 	SDL_SetWindowMinimumSize(window_, min_w, min_h);
+	// Can this change the size of the window?
```
vn971, line 169: Is that a TODO before merging the PR? Just checking

src/video.cpp
```diff
@@ -191,8 +194,98 @@ void CVideo::update_framebuffer()
...
+	{
+		uint32_t format = window->pixel_format();
+		int bpp = SDL_BITSPERPIXEL(format);
+
+		// Free the old drawing surface.
+		if (drawingSurface) {
+			// Except it can still be in use... what should we do about that?
+			LOG_DP << "leaking old drawing surface" << std::endl;
+			//LOG_DP << "freeing old drawing surface" << std::endl;
+			//SDL_FreeSurface(drawingSurface);
```
vn971, line 257: Is that a TODO before merging the PR? IDK whom to summon from the wesnoth team here :thinking:

src/video.cpp
```diff
@@ -220,7 +313,11 @@ void CVideo::init_window()
 		window_flags |= SDL_WINDOW_MAXIMIZED;
 	}
 
+	// TODO: fix whatever is crashing the rendering context when accelerated
+	// We can force software rendering here,
+	// but if we don't specify anything SDL should choose correctly.
 	uint32_t renderer_flags = SDL_RENDERER_SOFTWARE;
+	//uint32_t renderer_flags = 0;
```
vn971, line 320: Is this line still needed?

Pentarctagon:
> The failing mac build seems to just not be linking the new sdl/input.cpp. I added it to source_lists/wesnoth, but i guess there's somewhere else it needs to be added?

The macOS jobs use their own projectfiles rather than being generated by cmake, so those need to be updated separately by someone who has xcode installed.

Vultraz:
Nice! Really great work here! :) I think this also means our .exe needs to be properly marked HDPI-compatible to disable OS scaling.

src/gui/widgets/settings.cpp
Vultraz: It's not a WML variable, it's a WFL (formula language) variable. Considering if we should remove the `screen_pitch_microns` stuff now that we have this, since IIRC, it was an attempt to implement stopgap HDPI scaling 🤔

gwhitney: I absolutely think it's great to have a scheme to improve the playability of wesnoth on hidpi screens, and so I am all for this scaling option, but this approach is no less a stopgap than anything that's come before -- the only "real" solution is to stop "thinking" in terms of pixels internally and do all display in perceptual-scale units (degrees in field of vision, for one example, look at what modern browsers define a "px" as). That way text can be placed appropriately and in readable sizes, while the high resolution can be used for detail in the scenes and icons etc, if/when renderings at high pixel densities are available. Simply scaling all pixels is a stopgap for the former (text readability) that cuts off the latter opportunity (actually using the higher resolution for detail where it would add to the game experience).

src/hotkey/command_executor.cpp
CelticMinstrel: I have literally no idea what this comment is referring to?

src/gui/widgets/settings.cpp
Pentarctagon: Though it's also somewhat unlikely that we'd be able to overhaul/replace all the art necessary for the latter.

src/hotkey/command_executor.cpp
vn971: @CelticMinstrel I think it tries to explain what exactly of `sdl/input.hpp` will be used in this file. However, since it is not compiler-enforced anyway, I'd rather not have those. But I don't know wesnoth style sufficiently to say, thus just raising as a question.

src/video.cpp
mesilliac: Not really. It's only ever called when changing window size so it shouldn't happen frequently enough to be a problem. But i don't like it so i left the comment there.

I tried cleaning up properly but something elsewhere was still accessing the old surface and it looked like hell to debug (and would probably open a whole can of multithreading worms).

src/sdl/window.cpp
mesilliac: No, it will work fine in any case. I just wasn't sure if it was necessary to update the size variables when this was called. So i updated them anyway. If it doesn't actually change the window size the line i added can be removed. Not a huge issue either way.

src/video.cpp
mesilliac: Yeah - swapping the comment on those lines will enable hardware accelleration for platforms that support it. Ideally this should be enabled, but the line above forces software rendering in all cases.

I tried enabling it but got some crashes with an invalid GL context. It was a bit out of scope for this PR, but definitely should be looked into in the future.

src/hotkey/command_executor.cpp
mesilliac: Some files have these, some files don't. I prefer to include them but can tidy it up so it follows the style of the file it's in, if desired.

The other files that do this write it as "for get_mouse_state" etc, so i could also change it to that to make it more clear.

src/gui/widgets/settings.cpp
mesilliac: Hmm, if those specs are being made available to addons it should probably be expanded to add a couple more. Like the stuff i added to the build info page - there technically now is a "drawing size" in pixels which is what wesnoth will see and draw to, and there's a "window size" in desktop/display/screen coordinates which is used for changing resolution, and there's an "output size" in pixels of the final rendering target which may be even bigger for apple-style high dpi. I changed this "screen size" here to now refer to the "drawing size", so anything that uses that should still work at least.

Re: what gwhitney says, i agree that the basic problem is that everything works in pixels and there are (at least) three different meanings to a pixel nowdays. However i did move things in the correct direction by changing to use render textures. It should be possible (with some more work) to render some parts of the display in high-dpi, and the others in low/game-native dpi. A good place to start would be the background image (you can see in the screenshots i posted it's much higher quality in the non-scaled pic).

The biggest barrier to this actually isn't the rendering system but the input and event handling everywhere which assumes that input coordinates = render coordinates = display coordinates = output coordinates. I tried to at least make this distinction clear now in the rendering system.

mesilliac:
> Nice! Would this let us close #2219 (HDPI support) or would there still be things to do for that?

It does do exactly what the OP asked for so i'd say it closes that as originally requested.

But then again there's a lot more to actual proper high-dpi supprt, which seems to be what most of the discussion in the comments is about.

src/gui/widgets/settings.cpp
Pentarctagon: Yeah, https://github.com/wesnoth/wesnoth/tree/old_master_20181015 was actually a previous attempt to get hardware acceleration working, and IIRC it also stalled out over the same general issue.

mesilliac: Bringing it back to the code in question - i also don't actually know what this function returns for high-dpi displays. The actual dot-pitch of the physical display? Or the dot pitch in the scaled windowing system coordinates?

In my case it returns the physical DPI of the 55" display (~80dpi), which won't ever trigger high dpi handling even though the pixels appear very small from where i sit. Given that, i'd say the `screen_pitch_microns` variable should be removed in favour of checking the drawing resolution, and behaving differently if it's large. That will still work here.

> We should probably also make sure that the settings can be "live-updated" without restarting Wesnoth. Meaning that the WML variable gets updated on Scale change

I checked briefly and it looks like this should update automatically. But let me know if it doesn't and i'll try to fix it.

vgaming:
Hi, thanks a lot for the PR! Reporting some test results:

* The interface scaling overall works! :tada: It is "pixelated" though, just for awareness for reviewers. E.g. Fonts aren't actually the 4k smooth ones, but they are 1080p scaled 2x.

* The above still useful, comparing to not have scaling at all. Menu options and terrain labels finally started to be visible



* There is what seems to be a bug with fullscreen mode and the "automatic pixel scale multiplier" checkbox:


As you can see, the "[End Turn]" label on the second screenshot somewhy moved down, partially out of reach of the monitor's display

* would happily test more later!

src/hotkey/command_executor.cpp
CelticMinstrel: Oh, that. I usually only add that if it's not obvious – that is, you're including a file for some auxiliary thing it contains rather than the main thing in that file.

src/gui/widgets/settings.cpp
CelticMinstrel: I'm more or less okay with removing `screen_pitch_microns` as long as it's only used internally in GUI2 formulas.

mesilliac:
Hey thanks for the testing :). It would also be helpful to know OS and any OS-related high-dpi settings as well, as it looks like handling is different between Win/Mac/Linux, and also between X11 and Wayland on Linux. I can only test on Linux/X11 right now. I don't know much other than the basics on how other platforms do high-dpi.

Re: the font scaling yeah, they're all just scaled up for now so will be pixellated. I'd say the top layer of fonts would be the second thing to change to draw directly to the render target, after the background image. It might be complicated with things like overlapping windows, but probably just rendering the top window/dialog's fonts in high-dpi would be a good start.

Eventually the whole rendering system should stop blitting pixels to the drawing surface, and start copying textures to the render target. But this conversion either needs to be done layer-by-layer, or all-at-once. It's also much more likely to spawn bugs and compatibility issues so i didn't implement anything to help do it here.

Re: the end turn button i also noticed it being off in any small window, even with no pixel scaling. I think it might be a theme issue? I don't really know where to start there.

src/video.cpp
mesilliac: Oh, derp. Turns out it does already have a wrapper which is freeing it in the destructor. Just tested and it seems to be doing the right thing. I'll just remove this block.

src/video.cpp
mesilliac: Just a further note that i found the thread that was clearing the display and got this working, but i don't want to mix that up with this PR so i'm leaving it be for now.

Vultraz: What thread was? Only one thread should handle drawing.

mesilliac: The thread that handles the loading bar in `gui/dialogs/loading_screen.cpp` - which is passed a lambda when constructed in `play_controller::init()`. Inside that lambda it creates a `game_display`, and inside the constructor for `game_display` it called `CVideo::clear()` on the global video instance. So it ended up that the loading screen thread was clearing the screen.

When i remove the call to clear the screen from the `game_display` constructor everything seems to work. The call seems unnecessary anyway.

Vultraz: I recommend opening a separate PR for that if you're not adding it to this one.

mesilliac: Will do, i have a branch set up but it depends on this one so i'm waiting for a little more testing of this before posting it.

mesilliac:
This has been working smoothly for me. I'd say it just needs one confirmation it's working correctly on Mac and one on Windows.

I verified that the end turn button misplacement happens in small windows on the master branch as well, so it's unrelated.

Pentarctagon:
@Vultraz Can test it on Windows.
@CelticMinstrel Could you test this on macOS?

CelticMinstrel:
Although I can certainly test it, I don't believe my test would be useful since I don't have 4K monitor.

Pentarctagon:
Yeah - testing that it still works on monitors that don't need any scaling is certainly a valid test as well.

src/gui/widgets/settings.cpp
mesilliac: I'll update the comment to reflect that it should probably be deprecated. Physical DPI is never a good way of determining user-apparent pixel size, as displays do need to return accurate physical DPI info to correctly size physically-relevant things (like PDFs). I checked the calculations in `data/gui/*` and they should definitely be updated to use a different method, such as comparing to the maximum supported window size.

mesilliac: For now i updated `get_dpi_scale_factor()` as well to include pixel scale. I think this will keep behaviour approximately what it was. As i mentioned it doesn't trigger for me anyway so it's a little difficult to test.

mesilliac:
The latest commit didn't change anything critical. Just hopefully will prevent the previous high-dpi UI tweaks from applying when pixel scaling is used.

For 1080p displays a pixel scale of 2 can be forced in preferences for testing. Just knowing it doesn't break anything on a normal Mac system would be helpful.

CelticMinstrel:
I'm getting an assertion failure in SDL_DestroyTexture (called from video.cpp, line 268).

mesilliac:
That is actually really weird :/. If the texture is invalid then according to the SDL docs it should just fail silently. And it shouldn't ever be invalid.

Can you post the output when run with `--log-info=display`?

Pentarctagon:
Also it might be useful to post the build info (mainly wondering about the SDL version used).

Pentarctagon:
When I change the scale from 2 -> 3 -> 4 -> 3 -> 2 in the preferences while on the main menu, then attempt to start the tutorial campaign, it segfaults.

It does not segfault:
* On pixel scale 1 or 2 (scales 3 and 4 don't seem to change anything visually).
* If I change the scale only from 2 -> 3 -> 2.

Backtrace:
```
#1  0x00007ffff787cc75 in  () at /lib/x86_64-linux-gnu/libSDL2-2.0.so.0
#2  0x00007ffff78ad5c1 in  () at /lib/x86_64-linux-gnu/libSDL2-2.0.so.0
#3  0x00007ffff787343c in  () at /lib/x86_64-linux-gnu/libSDL2-2.0.so.0
#4  0x00007ffff784f801 in  () at /lib/x86_64-linux-gnu/libSDL2-2.0.so.0
#5  0x00007ffff7854bfb in  () at /lib/x86_64-linux-gnu/libSDL2-2.0.so.0
#6  0x00007ffff7854cef in  () at /lib/x86_64-linux-gnu/libSDL2-2.0.so.0
#7  0x0000555556aa30a9 in CVideo::flip() (this=0x55555768c750) at src/video.cpp:429
#8  0x0000555555b9055a in gui2::event::sdl_event_handler::draw() (this=0x555557da8ff0) at src/gui/core/event/handler.cpp:601
#9  0x0000555555b8fb2c in gui2::event::sdl_event_handler::handle_event(SDL_Event const&) (this=0x555557da8ff0, event=...) at src/gui/core/event/handler.cpp:421
#10 0x000055555697f1f9 in events::pump() () at src/events.cpp:740
#11 0x0000555555981c66 in gui2::window::show(bool, unsigned int) (this=0x555557bfd2e0, restore=false, auto_close_timeout=0) at src/gui/widgets/window.cpp:542
#12 0x0000555555bc684e in gui2::dialogs::modal_dialog::show(unsigned int) (this=0x7fffffffc500, auto_close_time=0) at src/gui/dialogs/modal_dialog.cpp:90
#13 0x0000555555bb719d in gui2::dialogs::loading_screen::display(std::function<void ()>) (f=...) at src/gui/dialogs/loading_screen.cpp:197
#14 0x0000555555e28a91 in play_controller::init(config const&) (this=0x7fffffffca70, level=...) at src/play_controller.cpp:210
#15 0x0000555555e26c52 in play_controller::play_controller(config const&, saved_game&, bool) (this=0x7fffffffca70, level=..., state_of_game=..., skip_replay=false) at src/play_controller.cpp:194
#16 0x0000555555e402d8 in playsingle_controller::playsingle_controller(config const&, saved_game&, bool) (this=0x7fffffffca70, level=..., state_of_game=..., skip_replay=false) at src/playsingle_controller.cpp:78
#17 0x0000555555b43a6c in campaign_controller::playsingle_scenario(end_level_data&) (this=0x7fffffffd300, end_level=...) at src/game_initialization/playcampaign.cpp:188
#18 0x0000555555b4407d in campaign_controller::play_game() (this=0x7fffffffd300) at src/game_initialization/playcampaign.cpp:278
#19 0x0000555555b4dd1b in game_launcher::launch_game(game_launcher::reload_mode) (this=0x55555768c790, reload=game_launcher::reload_mode::RELOAD_DATA) at src/game_launcher.cpp:1011
#20 0x00005555559a3b9a in do_gameloop(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)
    (args=std::vector of length 1, capacity 1 = {...}) at src/wesnoth.cpp:949
#21 0x00005555559a44ff in main(int, char**) (argc=1, argv=0x7fffffffe5f8) at src/wesnoth.cpp:1113
```
Build info:
```
The Battle for Wesnoth version 1.17.3+dev (cb685995d22-Clean) x86_64
Running on Linux Mint 20.3 x86_64
Distribution channel: Default

Game paths
==========

Data dir:        /home/USER/Desktop/wesnoth/wesnoth-dev
User config dir: /home/USER/.config/wesnoth
User data dir:   /home/USER/.local/share/wesnoth/1.17
Saves dir:       /home/USER/.local/share/wesnoth/1.17/saves
Add-ons dir:     /home/USER/.local/share/wesnoth/1.17/data/add-ons
Cache dir:       /home/USER/.cache/wesnoth

Libraries
=========

Boost:             1.71
Lua:               5.4.4
OpenSSL/libcrypto: 1.1.1f (runtime 1.1.1f)
Pango:             1.44.7 (runtime 1.44.7)
SDL:               2.0.10 (runtime 2.0.10)
SDL_mixer:         2.0.4 (runtime 2.0.4)

Features
========

Lua console completion:       yes
D-Bus notifications back end: yes

Current video settings
======================

SDL video drivers:        [x11] wayland dummy
Window size:              2560x1409
Game canvas size:         1280x704
Final render target size: 2560x1409
Screen refresh rate:      144
Screen dots per inch:     84.666664x84.666664
Screen dpi scale factor:  0.440972x0.440972

Current audio settings
======================

Audio not initialized.

Installed add-ons
=================

test1: 0.1.9
```

mesilliac:
Hmm... that segfault could actually be because of the loading screen thread clearing the display. I'll try including the fix for that here without enabling hardware acceleration.

I did manage to trigger a segfault by repeatedly messing with things and starting campaigns. I haven't managed to trigger it after including this fix.

CelticMinstrel:
> Can you post the output when run with `--log-info=display`?

Sure. Also of note is that the MacOS assertion handler apparently gives me the option to ignore the assertion and continue, after which it appears to work fine and looks exactly the same as master. Perhaps @hrubymar10's 1.17 branch (which is where I got the libs from) contains a debug build of SDL2?

```
Setting mode to 1600x900
20220423 23:50:01 info display: pixel scale: 1
20220423 23:50:01 info display: overriding logical size
20220423 23:50:01 info display:   old lsize: 0,0
20220423 23:50:01 info display:   old wsize: 1600,900
20220423 23:50:01 info display:   old osize: 1600,900
20220423 23:50:01 info display:   new lsize: 1600,900
20220423 23:50:01 info display:   new wsize: 1600,900
20220423 23:50:01 info display:   new osize: 1600,900
20220423 23:50:01 info display: creating 32bpp drawing surface with format SDL_PIXELFORMAT_ARGB8888
20220423 23:50:01 info display: destroying old drawing texture
2022-04-23 23:50:01.660974-0400 The Battle for Wesnoth[98834:38899745] WARN: 

Assertion failure at SDL_DestroyTexture_REAL (/Volumes/WD500/Users/martinhruby/xcode/wesnothMacCompileStuff_arm64/SourcesAndScripts/Sources/SDL2/SDL/src/render/SDL_render.c:3188), triggered 1 time:
  'texture && texture->magic == &texture_magic'
20220423 23:50:16 info display: creating drawing texture
Checking lua scripts... ok
20220423 23:50:28 info display: calling SDL_Quit()
20220423 23:50:28 info display: called SDL_Quit()
```

mesilliac:
Err, it looks like the problem is that i failed to initialize the drawing texture pointer. I think this should probably fix it.

CelticMinstrel:
Yup, looks like that fixes it.

mesilliac:
Great, thanks :)

Wedge009:
Just noting, as I did in #6663, that this change makes everything horribly large for me (2560x1440). Is there any possibility to make this optional/configurable? (I know there's resistance to allowing too many options for the user.)

mesilliac:
Yes it has a setting in preferences > display, so you can override the automatic scale choice there.

If you have fractional scaling for your desktop it may not be possible to get the same scale factor, as it's only allowing integer scaling. There isn't yet a consensus on how best to handle this.

Wedge009:
Oh that's a relief - thank you, I missed that.

cooljeanius:
> Yes it has a setting in preferences > display, so you can override the automatic scale choice there.
> 
> If you have fractional scaling for your desktop it may not be possible to get the same scale factor, as it's only allowing integer scaling. There isn't yet a consensus on how best to handle this.

Are there docs pointing users to this new option in preferences? e.g. release notes, help entries, etc.

Wedge009:
It's only in master for now, and not in a released version. There is a brief mention in the [change log](https://github.com/wesnoth/wesnoth/blob/master/changelog.md) but maybe this is one of the new features that needs to be highlighted in the next development release notes since it's such a visible change. Depending on how players use the development releases, or 'beta' channel in Steam, etc, and their expectations around that, such a big change may come as a surprise.

PR #4265, opened 2019-08-22
WoV: Set up for translations

stevecotton:
The next scons or cmake pot-update will create all of the empty .po files.

This pushes the Portuguese translation below the 70% threshold.

For the en_GB translation, AFAICS only "honor" and maybe "overwatch" need to change.

PR #6630, opened 2022-04-18
[WoF] Add missing wesnoth-wof.pot file

knyghtmare:
Title says it all.

This was never added as the addition of WoF was rushed for 1.17.0 (as something new had to be offered)


stevecotton:
Thanks for doing this, but refer to #4265 for an example, or to the addition of World Conquest. The set of languages that we use has changed since then, so take a LINGUAS file (, etc) from the current tree.

knyghtmare:
hmm, anything else missing?

PR #6633, opened 2022-04-22
[1.16] SoF S09: A few extra translation hints

stevecotton:
Also an idea for the translators' changelog: put a paragraph starting with `po:` in the commit message, so that someone can generate the translators' changelog from the Git changelog.

Example covering this and 2581ab6c87f24abc92c1933d790324cc22cc20e3 / f82079bcc0eb2a24fa0e0ee6559fb390480a77d7 :
```
po: In SoF S09, many strings involving Krawg (a gryphon) assumed that Krawg was
the only flying unit. Similar texts have been added that are used when the player
has gryphon riders too; most of these will be marked as fuzzy versions of the
Krawg-only strings.

### DIFF
--- a/data/campaigns/Sceptre_of_Fire/scenarios/2_Closing_the_Gates.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/2_Closing_the_Gates.cfg
@@ -509,7 +509,7 @@
         [/message]
         [message]
             speaker=Baglur
-            # po: "it" is a side entrance a dwarven fortress
+            # po: "it" is a side entrance of a dwarven fortress
             message= _ "I think it can be sealed up somehow... yes, look, activating that glyph seems to have closed up the gap."
         [/message]
     [/event]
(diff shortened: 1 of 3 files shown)


Pentarctagon:
I guess, is there a benefit to doing it this way rather than regular hints that show up in poedit?

stevecotton:
The idea of putting it in the commit message is that it's useful for a translator who had an up-to-date 1.16.2 file and is looking at the new strings, but would just be excess info for a translator who's starting with a very out-of-date or never-translated file.

Pentarctagon:
I'd say this should include the (I assume python) script to do this as well then.

stevecotton:
I wasn't planning to automate it, at least not in the near future. The plan is just to pipe the output of `git log 1.16.2..` or `git log --grep '^po:' 1.16.2..` into a text editor, the same workflow as my "add missing stuff to the changelog" task.

CelticMinstrel:
> I guess, is there a benefit to doing it this way rather than regular hints that show up in poedit?

Well, I would say that his example is extremely unsuitable as a general translation comment:

> po: In SoF S09, many strings involving Krawg (a gryphon) assumed that Krawg was
> the only flying unit. Similar texts have been added that are used when the player
> has gryphon riders too; most of these will be marked as fuzzy versions of the
> Krawg-only strings.

This only makes sense if the translator is comparing two versions. It's nonsensical for a generic po file, especially a newly-generated one.

Wedge009 (approved):
Just remarking on the hints specifically, this seems good to me.

PR #6635, opened 2022-04-23
Changes to prepare for ubuntu 22.04 CI

Pentarctagon:


Vultraz:
The "Fix range-loop-construct warning" commit should be done by using either the `s` or `sv` string literals in the array ctor.

Pentarctagon:
The what?

Vultraz:
If you want a string array:
```cpp
using namespace std::string_literals;
const std::array required {"id"s, "config_name"s, "create_map"s};
```

If you want a string_view array:
```cpp
using namespace std::string_view_literals
const std::array required {"id"sv, "config_name"sv, "create_map"sv};
```

Right now the compiler is deducing an array of `const char*`, but using the literals will make it an array of strings or string views, so you can use that type in the loop.

src/generators/lua_map_generator.cpp
```diff
@@ -37,15 +37,16 @@ lua_map_generator::lua_map_generator(const config & cfg, const config* vars)
...
+	const std::array required {"id"s, "config_name"s, "create_map"s};
+	for(const auto& req : required) {
 		if (!cfg.has_attribute(req)) {
 			if(req == "create_map" && cfg.has_attribute("create_scenario")) {
 				// One of these is required, but not both
 				continue;
 			}
 			std::string msg = "Error when constructing a lua map generator -- missing a required attribute '";
-			msg += req + "'\n";
+			msg += req + std::string("'\n");
```
Vultraz, line 49: Don't need the extra string construction but I guess it can't hurt.

src/utils/general.hpp
```diff
@@ -34,10 +34,9 @@ inline constexpr bool decayed_is_same = std::is_same_v<std::decay_t<T1>, std::de
 
 /**
  * Workaround for the fact that static_assert(false) is invalid.
- * See https://devblogs.microsoft.com/oldnewthing/20200311-00/?p=103553
  */
-template<typename>
-inline constexpr bool dependent_false_v = false;
+template<bool flag = false>
+void static_assert_false() { static_assert(flag); }
```
Vultraz, line 39: Does this work even if not marked constexpr?

src/generators/lua_map_generator.cpp
Pentarctagon: Oh, I suppose that's true now that `req` isn't a `const char*`.

src/hash.cpp
```diff
@@ -29,18 +29,13 @@ extern "C" {
 
 #ifndef __APPLE__
 
-#include <openssl/sha.h>
-#include <openssl/md5.h>
-
-static_assert(utils::md5::DIGEST_SIZE == MD5_DIGEST_LENGTH, "Constants mismatch");
-static_assert(utils::sha1::DIGEST_SIZE == SHA_DIGEST_LENGTH, "Constants mismatch");
+#include <openssl/evp.h>
```
CelticMinstrel, line 32: Why did you remove the MD5 static_assert here but not in the Apple section? You should probably remove either both or neither.

src/preferences/credentials.cpp
```diff
@@ -270,18 +270,17 @@ static secure_buffer rc4_crypt(const secure_buffer& text, const secure_buffer& k
...
-	}
+	int outlen;
+	int tmplen;
+
+	EVP_CIPHER_CTX *ctx = EVP_CIPHER_CTX_new();
+
+	// TODO: use EVP_EncryptInit_ex2 once openssl 3.0 is more widespread
+	EVP_EncryptInit_ex(ctx, EVP_rc4(), NULL, key.data(), NULL);
+	EVP_EncryptUpdate(ctx, result.data(), &outlen, text.data(), text.size());
+	EVP_EncryptFinal_ex(ctx, result.data() + outlen, &tmplen);
```
CelticMinstrel, line 281: Did you check that `outlen` will always be equal to the input length?

It seems like a plausible assumption but I'm not sure if it's a correct one.

src/gui/core/event/dispatcher.hpp
```diff
@@ -623,7 +623,8 @@ class dispatcher
 		} else if constexpr(cat == event_category::text_input) {
 			return signal_text_input_queue_;
 		} else {
-			static_assert(utils::dependent_false_v<cat>, "No matching signal queue for category");
+			// "No matching signal queue for category"
+			utils::static_assert_false();
```
CelticMinstrel, line 627: This is a negative change, as it loses the error message…

src/hash.cpp
Pentarctagon: Because the new evp header doesn't have the constant to compare against anymore, and the evp header isn't used on Apple systems.

src/gui/core/event/dispatcher.hpp
Pentarctagon: It was either this or remove it entirely, since Clang was rejecting the version that allowed an error message.

src/preferences/credentials.cpp
Pentarctagon: I don't follow. The output length is a separate value from the input length?

src/gui/core/event/dispatcher.hpp
Vultraz: Sounds like a clang bug. You're supposed to be able to use a message.

Pentarctagon: See https://github.com/wesnoth/wesnoth/runs/6146018736?check_suite_focus=true#step:3:215

Vultraz: That looks like it could have been solved with `decltype(cat)`.

src/preferences/credentials.cpp
CelticMinstrel: Yeah? I'm not sure how RC4 works, but _some_ encryption schemes require padding to bring the ciphertext to a multiple of the block size.

Pentarctagon: The previous implementation doesn't seem to, at least. It calls RC4 with a length of block_size as many times as it can, then calls it again at the end with the size of any remaining characters that aren't enough to fill a block. So there's no padding being done.

src/hash.cpp
Pentarctagon: I don't follow? Are there multiple types of MD5 hashes?

PR #6636, opened 2022-04-24
1.17 - UtBS - recolor of dwarf/phantom castle

doofus-01:
The "phantom castle" in UtBS is a barely modified dwarf castle, always struck me as either pointless (just use the core version) or inadequate.  I don't personally feel like making a custom castle for this scenario, but some IPF recolor might be worth considering.  An example so far:
Screenshot_20220424_castles
I went with blend because the lowered contrast might seem more ghostly.

I'm not sure what I think of this, maybe any severe recolor is just going to look bad, and some other ruins-type castle might be a better starting-point anyway.

Whatever change is made, if any is, would dictate what we do with the keep image (currently the dwarf throne)

pinging @nemaara because it is mainline campaign material


knyghtmare:
Hmm, I like the wall colors (blue-grey is one of favourite palettes) but the base may need some re-coloring to make it match. Sort of feels off.

doofus-01:
Yeah, the floor would need to be changed.  No reason to stick with cobbles, maybe something a little more outlandish would be better.
Screenshot_20220424_castles2

Vultraz:
Honestly, I don't thin any recolor is going to look good, because it's very obviously a recolor. 😛 If not making a new variant I suggest just using the core version.

doofus-01:
Well, tautologies are certainly true, but that's not a convincing argument. :stuck_out_tongue:

knyghtmare:
image
Does this fit the theme?
(UMC terrain pack addon)

doofus-01:
That looks familiar...  The walls are no good but the throne could be used.

cooljeanius:
> image Does this fit the theme? (UMC terrain pack addon)

Yeah I think it fits and is good

Pentarctagon:
Is this still being done?

cooljeanius:
> Is this still being done?

Or, more specifically, if someone is doing this, then who? Are we going with the one @knyghtmare proposed, or is @doofus-01 still going to come up with something else?

doofus-01:
The one knyghtmare showed was and old, aborted attempt at something I tried a long while back, I'd prefer if we didn't use that.

I might come back to this next time I go through UtBS.  Or someone else can propose something.  But none of that requires anything here, so no problem closing this.

PR #6637, opened 2022-04-25
Scroll at a consistent speed.

### LINKED ISSUE #3607: Scrolling in game / editor is erratic to the point of being annoying
Since 1.14.0, scrolling in game and in editor has been erratic. 

Steps to reproduce:
1. Open Map Editor
2. Create an empty map.
   - Size doesn't seem to matter, but let's go with 100x100
3. (Optional) Set scrolling speed to low for easier observation
4. Move the camera around the map with either mouse or arrow keys

What should happen:
The same as in 1.12: the actual scrolling speed is constant. 

What actually happens
The speed of scrolling usually follows this pattern: fast, slow, fast.

I've tested this with both Steam version and with compiled version. My system information:

- Wesnoth version: 1.14.5 from source (or 48bd45d-Modified from Steam).
- SDL2: 2.0.8
- SDL2_image: 2.0.3
- g++: 7.3.0
- OS: Kubuntu 18.04
- GPU: Intel(R) HD Graphics 520 (max core profile is 4.5 and compat profile 3.0)

I noticed that while compiling 1.14.5 from source,  the Steam build of Wesnoth didn't exhibit this behavior as strongly.

I will test this on Windows 10 soon.

As I said, this problem has been present since 1.14.0. Sorry for not reporting about this earlier!

mesilliac:
Fixes #3607

The `controller_base::handle_scroll` function is being called inside a loop that doesn't seem to be limited by anything. So if processing is fast, it loops and scrolls a lot, and if processing is slow, it doesn't.

I made it scroll by an amount proportional to the amount of time between calls to `handle_scroll`.

### DIFF
--- a/src/controller_base.cpp
+++ b/src/controller_base.cpp
@@ -43,6 +43,9 @@ controller_base::controller_base()
 	, scroll_down_(false)
 	, scroll_left_(false)
 	, scroll_right_(false)
+	, last_scroll_tick_(0)
+	, scroll_carry_x_(0.0)
+	, scroll_carry_y_(0.0)
 	, key_release_listener_(*this)
 	, last_mouse_is_touch_(false)
 	, long_touch_timer_(0)
@@ -273,7 +276,7 @@ bool controller_base::handle_scroll(int mousex, int mousey, int mouse_flags)
 		|| preferences::get("scroll_when_mouse_outside", true);
 
 	int scroll_speed = preferences::scroll_speed();
-	int dx = 0, dy = 0;
+	double dx = 0.0, dy = 0.0;
 
 	int scroll_threshold = preferences::mouse_scroll_enabled()
 		? preferences::mouse_scroll_threshold()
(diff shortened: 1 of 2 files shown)


stevecotton:
Looks and feels good to me. I'm not sure if the acceleration is necessary (as it seems to be full speed within about 3 frames), but the speed limit definitely is.

CelticMinstrel:
I vote we merge this immediately.

mesilliac:
> I'm not sure if the acceleration is necessary

The slow start is mostly so all the variables (`scrolling_`, `last_scroll_tick_` etc) get set correctly. It was simpler than writing extra handling code. The small bit of acceleration is a nice side-benefit.

Vultraz:
I like this. :) I really wish we didn't have to do this manually; optimally camera position would be something we could interpolate, but that would require a deeper refactor.

mesilliac:
Yeah, applying an ease-in to `scroll_amount` should work. It would just need another variable to track the time since scroll start and apply the easing function based on that. I wasn't sure if there was an existing easing function / system to use so i didn't try anything fancy.

Vultraz:
What about replacing `last_scroll_tick_` with a `scroll_start_` var. We don't (oddly enough) have any built-in easing functions (they should be added), so for now a helper function in the file would do.

mesilliac:
Nah it still needs it for the time delta. As there's no endpoint for mouse/keyboard scrolling, only a direction, the speed can be eased but not the absolute position.

In any case it's pushing the scope for this a bit, as it's basically just a bugfix as is.

PR #6639, opened 2022-04-26
EI S07a various fixes

Descacharrado:
Addresses some points in #4145.

Side 2 was spoiled previously in two ways:
- With the status table showing it (as commented in the linked issue)
- With enough resolution and zoomout, you could see the leader being stored.
Now the status table at turns 1 and 2 will be:
side 2 hidden
side 2 unhidden
And the leader is stored in the prestart event to ensure that it is not leaked.

Also, the player will now be notified that two more ogres are joining Gweddry (both with dialogue and visually) and Grug will have no upkeep (fixed by deleting the [modifications] tag).

Squash before merging.

P.S. If you want to test this, keep in mind you need Owaec from S05 to trigger the victory condition.


changelog.md
```diff
@@ -2,6 +2,10 @@
  ### Add-ons client
  ### Add-ons server
  ### Campaigns
+   * Eastern Invasion
+     * S07a: 
+       * Hide side 2 without spoiling it (issue #4145)
```
stevecotton, line 7: Need to be wary of the changelog being a spoiler too - maybe "better handling for side 2"?

data/campaigns/Eastern_Invasion/scenarios/07a_The_Crossing.cfg
```diff
@@ -368,7 +373,8 @@
                     [then]
                         [message]
                             speaker=Grug
-                            message= _ "Grug say join you will he."
+                            #po: "Grug say join you will he" means "I will join you" in a weird speech pattern. "Gorg and Drog too" means that two other ogres will be joining Gweddry.
```
stevecotton, line 376: There's a `Gorg` in here which should be `Gork`. Only noticed it because I was thinking "that sounds too close to Grug".

Descacharrado: Good catch!

Descacharrado:
Found a small typo, adding to the batch without changelog because it is so minor. (a46c464)

Wedge009:
I think it's generally known that trolls are often characterised with incorrect or unusual grammar or word phrasing, but I suppose the po hint can't hurt. I would just say that the speak is a troll instead of saying 'weird speech pattern', but it can be left as it is for now.

Also it's fine to update straight apostrophes to their curly equivalents without making note of it - I would say that's preferred, in fact.

Go ahead and mark the review comments as resolved if you think they have been dealt with.

Wedge009:
Sorry for the potentially silly question, but what is hard-coding the co-ordinates for Gork and Drog supposed to do? I'm assuming a setting of `recall` just allocates their position according to the standard recall behaviour.

Wedge009:
I looked through the mentions of S7a (I remember this issue being a very big list) but didn't find the connection. Is it the case that `recall` just puts the units in the recall list instead of placing them in the scenario directly? I trust the effect is what you intended, I'm just trying to understand the change, that's all.

Descacharrado:
> Is it the case that `recall` just puts the units in the recall list instead of placing them in the scenario directly?

Yes, thats it.

stevecotton:
Looks good to me, but I'm not sure who should approve. I'd guess @nemaara.

nemaara (approved):

Wedge009:
Along with #6644, will have to wait for string freeze to be over before back-porting this.

PR #6644, opened 2022-04-27
UtBS S05 Enforce encounter with the Cloaked Figure

### LINKED ISSUE #6364: UtBS S5: Enforce second encounter with 'Cloaked Figure'
Extension of #6355, and may be desirable to revert #6361 if this issue is resolved.

Encounter with the 'Cloaked Figure' is not guaranteed in S5. For story reasons it would be good to enforce the encounter in a way that doesn't intrude upon the game mechanics too much.

Descacharrado:
The Cloaked Figure will appear when enemies are defeated if it did not already. 

I tried other methods, like spawning the cloaked figure some turns after defeating an enemy leader (so the randomness is kept) or some turns after moving to certain sections, but none of those implementations felt the same (as how it is rn) and could be cheesed (so we would not be enforcing the encounter, just changing the spawning method). 

So I decided that keeping the spawning mechanism and, in the cases where the figure is not spawned (likely because of players rushing), the figure just appears.

Closes #6364 and has a revert of #6361.


knyghtmare:
Would this PR still allow players to use the "Use Keratur as a meat shield" exploit?

knyghtmare:
Would this PR still allow players to use the "Use Keratur as a meat shield" exploit?

Wedge009:
If I understand the change correctly, the assassin is triggered when the last of the enemies - whether dwarf or troll - has been defeated. A bit awkward, but it maintains the story with that character, at least.

Does victory correctly flow on from when the assassin is defeated in that case (where the dwarves/trolls are all defeated first)?

Regarding #6361, I suppose it doesn't _have_ to be reverted but at the very least the comment change there will become obsolete so may as well revert it all. You can include the reversion as part of this PR, or I (or someone else) can revert that commit if/when this PR is merged.

Descacharrado:
> Would this PR still allow players to use the "Use Keratur as a meat shield" exploit?

I'm unfamiliar with that. Please expatiate.

knyghtmare:
> > Would this PR still allow players to use the "Use Keratur as a meat shield" exploit?
> 
> I'm unfamiliar with that. Please expatiate.

Well, the encounter was meant to happen at a certain point in the heat of battle and he pops behind/adjacent Kaleh. This allows a nasty exploit where this assassin is forced to spam at a spot which puts in the middle of Kalen and the trolls/dwarves. That is, he is a meat shield in a sense. :P It's rather amusing.

Wedge009:
> Sure, I can do that, do I just change it back to what it was or do I do git revert (or whatever) in this branch?

Either works. I think if you use `git revert` it will make it slightly easier for you as it automates the commit message and makes it quite clear what is being reverted.

I also hadn't heard of the 'meat-shield' tactic nor was aware that he is still flagged as an enemy to Kaleh's enemies - probably because when I did encounter him I dispatched him as quickly as possible.

Wedge009:
Looks fine to me. Thanks.

Wedge009:
@nemaara or anyone else: Any issue or concerns against merging this?

Descacharrado:
Last commit is changelog conflicts

nemaara (approved):

Wedge009:
I think it's better to re-base your PR branch rather than try to merge the destination back into it, but it looks like squash-merge will work.

Wedge009:
Oh, darn, I just realised the string freeze for 1.16 just started. Will have to wait for the back-port, then.

PR #6646, opened 2022-04-27
Add decltype.

Pentarctagon:
Already ran it on my fork's CI, so merging immediately.


CelticMinstrel:
I'm _kinda_ surprised that works, since `decltype(evt)` looks like it should be known at instantiation time. The ideal solution is probably to use an `auto` template argument in the definition:

```cpp
template<auto>
inline constexpr bool dependent_false_v = false;
```

But I'm not sure if that actually works in C++17, and I'm also not sure if `auto` can deduce to `typename`. I _think_ it should work on some version though, so it's something to keep in mind for the future…

PR #6647, opened 2022-04-28
Editor: Add some missing terrain groupings

### LINKED ISSUE #6643: [Terrain] [Editor] Properly assign Terrain types to appropriate terrain groups
Some terrain types seem to be not properly assigned to their appropriate groups.

An example is the Elven Winter Castle. Which is supposed to be in both castle and frozen but is in castle group only. Thus, someone using a frozen terrain kit misses out on that terrain type's existence.

This is just one example. More might be present.

Wedge009:
Resolves #6643.

I'm not overly familiar with the myriad of terrains but this seems to make sense to me (I tested the changes too). I tried to review everything but it's possible there could be more groupings that could be added.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
    * Sceptre of Fire
      * S9: Allow Grypon Riders to complete the scenario (issue #6332)
  ### Editor
+   * Added some missing terrain groupings (issue #6643)
  ### Multiplayer
  ### Lua API
  ### Packaging
--- a/data/core/terrain.cfg
+++ b/data/core/terrain.cfg
@@ -472,7 +472,7 @@ Most units receive 20 to 40% defense in sand."
     string=^Efs
     aliasof=_bas
     light=25
-    editor_group=embellishments
+    editor_group=embellishments, cave
 [/terrain_type]
 
 [terrain_type]
--- a/data/core/terrain.cfg
+++ b/data/core/terrain.cfg
@@ -261,7 +261,7 @@
     editor_name= _ "Desert Cobbles"
     string=Rrd
     aliasof=Gt
-    editor_group=flat
+    editor_group=flat, desert
 [/terrain_type]
 
 [terrain_type]
@@ -407,7 +407,7 @@ Most units receive 20 to 40% defense in sand."
     editor_name= _ "Snowbits"
     string=^Esa
     aliasof=_bas
-    editor_group=embellishments
+    editor_group=embellishments, frozen
 [/terrain_type]
 
 [terrain_type]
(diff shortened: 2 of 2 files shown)


doofus-01 (approved):
Looks right, thanks.

knyghtmare (changes requested):
Some cases missing:
1. Elven Winter Castle (found in castle, and not in frozen)

Optionals: 
1. The case of "Craters" being detachable now:
image

2. Cave/Dungeon Tileset missing sconces, gates, etc.
5.  And this:
image

Maybe point number 3 can be dismissed as it's more of a personal preference (since I make dungeon maps mostly)

Context on Case 2:
Originally it was Dd^Dr, but the crater can be detached now and it takes 3 different forms now.

knyghtmare:
> > 3. And this:
> 
> ??????

point 3 Seems like another issue already reported.

CelticMinstrel:
I have no idea what you're even trying to say with point 3. A picture isn't always better than words.

knyghtmare (approved):

knyghtmare:
> I have no idea what you're even trying to say with point 3. A picture isn't always better than words.

As in might as well tackle the deprecated terrain collection issue as well. 

However, just adding point 1 is enough.

Wedge009:
> Some cases missing:
> 
>     1. Elven Winter Castle (found in castle, and not in frozen)

Um, that was the first thing I fixed since that was what you reported originally in #6643. Or is that something different? I couldn't find that one, only 'Winter Elven Castle'.

> Optionals:
> 
>     1. The case of "Craters" being detachable now:

I don't know what that means. I could only find one tile as crater - it's currently grouped under desert. I couldn't replicate your image in 1.16.2 release or current master (eb9f463632a).

>     2. Cave/Dungeon Tileset missing sconces, gates, etc.

These are grouped under embellishments - I'm not so sure it makes sense to add all the gates to the cave group as well (there are a lot of them) but I added the sconce light to it.

>     3. And this:

This is from the Secret of the Ancients group? I have no idea about that one.

CelticMinstrel:
> As in might as well tackle the deprecated terrain collection issue as well.

Eh? Those D's are all deprecated terrains? Deprecated terrains shouldn't be shown at all.



> I don't know what that means. I could only find one tile as crater - it's currently grouped under desert. I couldn't replicate your image in 1.16.2 release or current master ([eb9f463](https://github.com/wesnoth/wesnoth/commit/eb9f463632a57f2b8ebb58988268fb2238d7c780)).

I think what he's suggesting is that crater is now an overlay and should be moved (or maybe duplicated) to embellishments.

Wedge009:
I'm really not sure what to do about that - the crater stuff. Unless I have a better understanding, I'll merge what I have here soon.

knyghtmare:
> I'm really not sure what to do about that - the crater stuff. Unless I have a better understanding, I'll merge what I have here soon.

yeah, whatever changes have been done will be welcomed by map-crafters. 

As for the craters...I am a bit unsure about them as I noticed they cause some balancing issues. So, might be better opening another topic.

Wedge009:
Okay, merged, and will back-port to 1.16.

Issue #2395, opened 2018-01-22
Titlescreen flash when starting game


GregoryLundberg:
When testing #2372 I notice a flash of the titlescreen after 'Reading unit tiles ...' and before 'Building terrain rules ...'

This only occurs on a cold start of the wesnoth client. If I quit back to the titlescreen and reload the replay without quitting to the desktop, I do not see the flash.

Vultraz:
Been an issue for years. Probably has to do with the loading screen object being created twice or something? Or destroyed and recreated?

CelticMinstrel:
I think there are indeed some circumstances where the loading screen is effectively shown twice consecutively. It was more obvious on the old loading screen, which had an actual progress bar (which I am still mildly annoyed about it being removed).

Wedge009:
Why was it removed, anyway? Inaccurate representation? gui2 conversion?

Vultraz:
I didn’t like it.

Vultraz:
Well, to be more specific, it was too large and in-your-face before. I dropped it with the conversion to GUI2 since I thought we could get away with an icon only, but @CelticMinstrel insisted we have some animated part, so that’s what we have the dots. I suppose a smal progress bar could be added back, maybe, but how to determine how much to fill up at any given time I’d have to find.

Wedge009:
Even a smaller one would still be nice, but not high priority obviously. The dots at least indicate that something is busy working, I suppose.

Sorry, back to the actual issue discussion...

Wedge009:
I'm not seeing any flash in 1.14.0 (Windows or Linux). Is this still happening? It may just be that I've been running Wesnoth a lot in the past few hours so a lot of resources could still be in OS memory.

Vultraz:
The underlying issue wasn’t solved, no.

CelticMinstrel:
Pretty sure I've seen it as recently as the last release candidate.

Wedge009:
1.14.9

Wedge009:
Closed in #2395, back-ported to 1.16 in 9be14b46030ca8edaa7a2be67ec111e00c5b5537.

CelticMinstrel:
@Wedge009 That referenced issue number is wrong (it's this PR) – do you know the correct one?

Wedge009:
Sorry, fixed. Must have copied the wrong number...

PR #6648, opened 2022-04-28
Fix single-frame title-screen flicker when loading a game or starting a campaign.

mesilliac:
The title screen was being displayed for a single frame in between loading dialogs when either starting a new campaign or loading a game.

This was giving me an aneurysm.

I just turned off drawing of the title screen after these processes are confirmed. The next time the title screen needs to be drawn, it is recreated from scratch anyway, so the existing instance should never actually need to redraw.


Wedge009:
I think that was a known issue - was it formally recorded (and therefore should be closed)? Or perhaps not...

mesilliac:
I searched but didn't find an open issue for it. Could be it was small and obvious enough that nobody bothered to report it.

Wedge009:
No worries, wasn't asking you specifically. Thanks for resolving the issue!

Wedge009:
Never mind, found it in #2395.

PR #6654, opened 2022-04-29
Refactor FindHistory.cmake

bencsikandrei:
Hello!

I was super excited to look into `wesnoth`! Amazing work!

While I was getting to know the project I've read through some of the `cmake` finders and thought I may help with some modernization.

I've made `FindHistory.cmake` use lower caps keywords, added an `IMPORTED` target for it and some documentation that looks like the standard `cmake` docs.

Please let me know if this kind of stuff is interesting to you as I'll probably also look into the other `cmake` files during my journey.

Thanks and keep up the great work!
Andrei

EDIT: my editor was configured with `spaces`. I noticed all other cmake finders use tabs. I'll update accordingly

### DIFF
--- a/cmake/FindHistory.cmake
+++ b/cmake/FindHistory.cmake
@@ -1,11 +1,37 @@
-# Locale GNU history library
+#[=======================================================================[.rst:
+FindHistory
+-----------
+
+Find the GNU History includes and library.
+
+IMPORTED Targets
+^^^^^^^^^^^^^^^^
+
+This module defines the `IMPORTED` target ``History::History``, if
+History has been found.
+
+Result Variables
+^^^^^^^^^^^^^^^^
+
+This module defines the following variables:
+
+	HISTORY_FOUND         - True if history found.
+	HISTORY_INCLUDE_DIR   - Where to find readline/history.h.
+	HISTORY_LIBRARY       - Library when using history.
+
+#]=======================================================================]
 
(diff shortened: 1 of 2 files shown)


cmake/FindHistory.cmake
```diff
@@ -1,11 +1,37 @@
...
+
+#]=======================================================================]
+
+find_path(HISTORY_INCLUDE_DIR readline/history.h)
 find_library(HISTORY_LIBRARY history)
 
-# handle the QUIETLY and REQUIRED arguments and set XXX_FOUND to TRUE if all listed variables are TRUE
-INCLUDE(FindPackageHandleStandardArgs)
-FIND_PACKAGE_HANDLE_STANDARD_ARGS(History DEFAULT_MSG HISTORY_LIBRARY HISTORY_INCLUDE_DIR)
+# handle the QUIETLY and REQUIRED arguments and set ISTORY_FOUND to TRUE if all listed variables are TRUE
```
Pentarctagon, line 27: It looks like the H is missing here.

bencsikandrei: done, thanks!

Pentarctagon:
Thanks! And as a heads up if you're going to continue looking at other files, the reason there are some differences between Windows/MSVC for how libraries are added is we use vcpkg for our Windows builds which makes things like SDL2::SDL2 available (which are present otherwise).

Issue #6417, opened 2022-01-03
[1.17/1.18] Sprite palette cleanup


doofus-01:
### Game and System Information
This is an aspiration for 1.18.  An issue for the roadmap, and a reminder for myself.

### Describe the desired feature
Clean up sprites to have a set color palette, as is done in #6413 for the saurians

(This will have to be updated with specifics...)

doofus-01:
One way to tackle this is to not worry too much about the precise palette colors for now, just establish and apply something reasonable.  Then adjusting the colors later, as in #6453, should be easier in every sense.

doofus-01:
The ImageMagick process I used that seemed to work for the ogres was:
- Open the base sprite in GIMP, use that to create a separate ColorMap image by manually color-picking and drawing.  Check the `images/tools/magenta_team_color_palette.png` to make sure you didn't pick a broken TC pixel.  Also, make sure the shadow color is the only one with an intermediate alpha value
- Check the ColorMap against all the colors in the original base image by using the output of `convert ogre.png -unique-colors -scale 1000%  ogre_colors.png`  
- Once you think the ColorMap is good, batch process the other images with a command based on `convert young-ogre.png  -dither None -remap Ogre_ColorMap.png  young-ogre_remapped.png`

That could probably be optimized and improved, but I think it will get me through.

doofus-01:
That's probably all I will get done in the near future.  Most images were not visibly changed.  I caught some that were mangled, but probably missed a few.

doofus-01:
Title might need updating, but right, monsters are still to be done, and there are new units that need to be processed.  Possibly I should start a new issue, then I will close this one.

PR #6662, opened 2022-04-30
1.17 dunefolk sprite cleanup

doofus-01:
part of #6417


knyghtmare:
may i ask what method/software technique you are using to mass-update the palettes so rapidly?

doofus-01:
Bash scripts and ImageMagick.  See #6417.

PR #6665, opened 2022-05-01
Added spritsheet generator

Vultraz:
Pulls in my old spritesheet generator code from the `old_master_20181015` with minor tweaks for code changes since then. Placing here for consideration.

This takes a folder with images and writes an assembled spritesheet to disk along with a config file.


Pentarctagon:
@Vultraz are you going to merge this? At minimum, it still needs to be added to the various projectfiles/source lists and somehow be usable from the command line.

Vultraz:
Yeah, I'll get to this soon.

CelticMinstrel:
Doesn't this also need the code to load the spritesheet before it can be of any use? It's not helpful to be able to generate a spritesheet that you can't even use.

Vultraz:
I'm getting to it....

stevecotton:
Given that @doofus-01 is moving us to have metadata embedded in the image files, how will the two features interact?

doofus-01:
Credits & copyrights are the issue, not metadata.  However it all gets merged, the spritesheets are still images, so they can have metadata.  I think we just need to figure out how the info and copyrights get merged, no matter whether it's logged in the image or in a CSV.

PR #6667, opened 2022-05-01
Refactor FindTranslationTools.cmake

bencsikandrei:
Hello!

I've looked into the translation search file and it was finding a bunch of `gettext` tools in the same manner.
Used a macro to factor that code out a bit.

How I've tested: inside a docker container and I've activated the `ENABLE_POT_UPDATE_TARGET` and built like that.
During my testing I've had cmake in very verbose mode to check for the paths and how it set the variables. All LGTM.

All in all seems to work well and the code is shorter and easier to modify if needed.

The CI seems to also do what I've done so fingers crossed!

Please let me know if you'd like me to further test this change.

Thanks!
Andrei

### DIFF
--- a/cmake/FindTranslationTools.cmake
+++ b/cmake/FindTranslationTools.cmake
@@ -1,18 +1,42 @@
-# - Find the tools needed for updating the potfiles and translations
+#[=======================================================================[.rst:
+FindTranslationTools
+--------------------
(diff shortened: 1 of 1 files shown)


Pentarctagon:
Looks good, thanks!

PR #6668, opened 2022-05-01
1.17 elf sprite cleanup

doofus-01:
part of #6417


cooljeanius:
I was using some of those unused Scout sprites in at least one of my add-ons...

PR #6671, opened 2022-05-02
Refactor find vorbis file cmake

bencsikandrei:
A couple of changes for consistency with the other modules + a small commit to state the minimum requirements for Boost and CMake which were a bit behind.

As always, tested as much I could, the `link.txt` file is looking good, same as `compile_commands.json`.

I was thinking, would it be interesting to move the `FindXXX` files in a `cmake/modules` folder? That way they're separate from other cmake utilities.

Also, since I've embarked in this journey, and haven't yet seen the whole history of the project, have there been efforts to unite the SDL2 find? I'm thinking about unifying windows/*nix finders with a couple of tricks but I'd like to see if there have been other efforts. Of course, it's fine even if there weren't and it may actually be impossible, but I'd like to try.

Thanks!

### DIFF
--- a/cmake/FindVorbisFile.cmake
+++ b/cmake/FindVorbisFile.cmake
@@ -1,14 +1,39 @@
-# Locate VorbisFile
-# This module defines XXX_FOUND, XXX_INCLUDE_DIRS and XXX_LIBRARIES standard variables
-#
-# $VORBISDIR is an environment variable that would
-# correspond to the ./configure --prefix=$VORBISDIR
-# used in building Vorbis.
+#[=======================================================================[.rst:
+FindVorbisFile
+--------------
 
-# Copied from
-# http://code.google.com/p/osgaudio/source/browse/trunk/CMakeModules/FindVorbisFile.cmake
+Find the VorbisFile includes and library.
 
-SET(VORBISFILE_SEARCH_PATHS
+Environment
+^^^^^^^^^^^
+
+$ENV{VORBISDIR} is an environment variable that would correspond to
+the `./configure --prefix=$VORBISDIR` used in building Vorbis.
+
+
+IMPORTED Targets
+^^^^^^^^^^^^^^^^
+
+This module defines the `IMPORTED` target ``VorbisFile::VorbisFile``, if
+VorbisFile has been found.
+
+Result Variables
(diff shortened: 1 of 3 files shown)


cmake/FindVorbisFile.cmake
```diff
@@ -19,78 +44,103 @@ SET(VORBISFILE_SEARCH_PATHS
...
+	set(MSVC_YEAR_NAME VS2010)
+elseif(MSVC_VERSION GREATER 1499)	# >= 1500
+	set(MSVC_YEAR_NAME VS2008)
+elseif(MSVC_VERSION GREATER 1399)	# >= 1400
+	set(MSVC_YEAR_NAME VS2005)
+elseif(MSVC_VERSION GREATER 1299)	# >= 1300
+	set(MSVC_YEAR_NAME VS2003)
+elseif(MSVC_VERSION GREATER 1199)	# >= 1200
+	set(MSVC_YEAR_NAME VS6)
+endif()
```
Pentarctagon, line 59: MSVC_YEAR_NAME can just be deleted at this point, as can anything mentioning win32 for PATH_SUFFIXES.

Pentarctagon:
> Also, since I've embarked in this journey, and haven't yet seen the whole history of the project, have there been efforts to unite the SDL2 find? I'm thinking about unifying windows/*nix finders with a couple of tricks but I'd like to see if there have been other efforts. Of course, it's fine even if there weren't and it may actually be impossible, but I'd like to try.

Feel free to try, of course. It's kind of a mess though as the comments around it in the root CMakeLists.txt notes.

> I was thinking, would it be interesting to move the FindXXX files in a cmake/modules folder? That way they're separate from other cmake utilities.

I don't have a preference either way.

cmake/FindVorbisFile.cmake
```diff
@@ -36,28 +36,14 @@ http://code.google.com/p/osgaudio/source/browse/trunk/CMakeModules/FindVorbisFil
 set(VORBISFILE_SEARCH_PATHS
 	~/Library/Frameworks
 	/Library/Frameworks
-    /usr/local
+	/usr/local
```
bencsikandrei, line 39: spaces -> tabs

PR #6672, opened 2022-05-02
Various fixes about [h|c]pp

bencsikandrei:
Hello!

I'll detail the small changes I've made in the  bunch of commits I've made. On all changes, please correct me if I made assumptions that are not valid, I'm trying my best to read and understand (and test) but I'm still in the process of discovering the project.

0) There were some minor typos inside the comments of `about.hpp` and I figured they were worth committing.

1) `std::vector::insert` can be called with an empty range as it does a check internally (via its loop) so I figured the `if` was not needed

2) found a place where a `std::map` was searched twice via `[]`, doesn't matter much since it's probably called very seldom but it looks a bit better with the `.find`, I think. Please correct me on this if I'm wrong, but I figured adding the (empty) entry (via `[]`) to the map in case of not found wasn't useful. If it is, then that change is wrong.

3) AFAIK default constructors are already called so there was no need to call them in the initializer lists

4) `std::pair` was used but its official header not included (it's most probably included by vector, or string) but I think it's safer like this

Thanks again for the awesome project and for the time to review my PR!

Have a good one!


src/about.cpp
```diff
@@ -106,11 +101,15 @@ std::optional<credits_data::const_iterator> get_campaign_credits(const std::stri
...
 std::vector<std::string> get_background_images(const std::string& campaign)
 {
-	if(!campaign.empty() && !images_campaigns[campaign].empty()) {
-		return images_campaigns[campaign];
+	if(campaign.empty()) {
+		return images_general;
 	}
 
-	return images_general;
+	if(const auto it = images_campaigns.find(campaign); it == images_campaigns.cend()) {
```
stevecotton, line 108: I feel that `if(...; it != images_campaign.cend())` would be better - put the handling for there being images in the `if` block, and the fallback when there aren't images in the `else` block.

Generally, I find the old version of this function easier to read.

bencsikandrei: I'll rework it to look similar to that one while still avoiding the double operator[]

bencsikandrei: Made the change

Pentarctagon:
> EDIT: Please tell me if I'd better squash these, cause I'll do it no probs!

I'd say squash, given these are fairly small changes all to the same header+source file.

Pentarctagon:
> AFAIK default constructors are already called so there was no need to call them in the initializer lists

It's something fairly widely done in the codebase though, so unless there's a particular reason to change it, I don't think the initializer lists would need to be removed.

Otherwise it seems fine, barring any additional comment from stevecotton.

CelticMinstrel:
Personally, I wouldn't add them but I wouldn't remove them either, unless maybe I was doing a pass of removing unnecessary constructor initializers across the whole codebase. So either way is fine with me really.

bencsikandrei:
> Personally, I wouldn't add them but I wouldn't remove them either, unless maybe I was doing a pass of removing unnecessary constructor initializers across the whole codebase. So either way is fine with me really.

I'm on board with consistency, indeed the pattern is used across the project. I like your idea a lot.

Would a `clang-tidy` pass with https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-default-member-init.html sound good?

I'll make the change so that it still uses initializer lists

Pentarctagon:
I'd say open a separate PR for it, and then discussion can be done there based on what changes it makes.

PR #6676, opened 2022-05-03
UtBS S08 Fixed replay OOS by moving victory event to synced event

### LINKED ISSUE #6267: UtBS S8 Out of the Frying Pan Replay: OOS after accomplishing the final objective

Descacharrado:
Half of #6439 This implementation is different from the previous implementation (in the linked PR) because there was a simpler way of doing it and the previous implementation was buggy.

Closes #6267


CelticMinstrel:
Okay, so this is kind of confusing, but I guess the main point is… there's a gigantic cutscene in someone's "last breath" event, but "last breath" events are not synchronized? (Is that true? I'm not even sure. Nothing on the wiki suggests they're not synchronized.)

Assuming the issue _is_ indeed that "last breath" is not synchronized… does firing a different event really address the issue? But… I assume you tested it, so I guess it does fix it?

I don't really get what's going on here at all. I guess if it works and doesn't produce an out-of-sync error, that's good enough…?

Descacharrado:
It seems to be an upstream problem with replays (see https://github.com/wesnoth/wesnoth/issues/4667). In my experience it happens when **units** are **spawned** (or moved, not totally sure, maybe both) **during** (or after, or both, again not exactly sure) a **victory event**.

I did test that moving it to a non-victory event fixes it. And this implementation solves the OOS on just this scenario.

CelticMinstrel:
Ah, so the problem is the victory event, not the last breath event… I see…

Pentarctagon:
@CelticMinstrel is this fine to merge then?

CelticMinstrel:
Well… I feel like there's a better way to do it, but it's still an improvement. So sure.

PR #6681, opened 2022-05-04
GUI2: made widget initialization process more secure against memory leaks

Vultraz:
Pulling this out of the old master branch.

This commit is the followup to a similar one I did regarding window initialization. Instead
of widgets being created on the heap and not being managed by a smart pointer until they're
added to a grid, they are now always managed by a shared_ptr. To that end, this commit covers
bunch of things:

* builder_widget::build (both overloads) and all its overrides (which have now been marked
  as such) now return shared_ptr<widget>.
* The builder_grid::build() override now returns the same instead of grid* since you can't
  use covariant return types with smart pointers.
* The two implementation build helpers in builder_grid have been combined with an optional
  replacement map as the second parameter. Uses of the version that took a grid pointer could
  be easily converted to pass a reference instead.
* The pane, matrix, and viewport build() functions were removed in favor of making the ctor
  public. In case there was a deprecated ctor, that was removed.
* The viewport now keeps a widget shared_ptr instead of a reference that was then deleted
  by address-of. This was both better design and necessary to fix a crash.
* ~~build_single_widget_instance_helper and build_single_widget_instance were renamed to
  build_single_widget and build_single_widget_and_cast_to to better represent their roles
  and to indicate the latter was more a convenience extension for the latter than the other
  way around.~~

### DIFF
--- a/src/gui/widgets/grid.cpp
+++ b/src/gui/widgets/grid.cpp
@@ -67,7 +67,7 @@ unsigned grid::add_row(const unsigned count)
 	return result;
 }
 
-void grid::set_child(widget* widget,
+void grid::set_child(widget_ptr widget,
 					  const unsigned row,
 					  const unsigned col,
 					  const unsigned flags,
@@ -90,16 +90,17 @@ void grid::set_child(widget* widget,
 	cell.set_flags(flags);
 	cell.set_border_size(border_size);
 	cell.set_widget(widget);
-	if(cell.get_widget()) {
-		// make sure the new child is valid before deferring
-		cell.get_widget()->set_parent(this);
+
+	// make sure the new child is valid before deferring
+	if(gui2::widget* w = cell.get_widget()) {
+		w->set_parent(this);
 	}
 }
 
-std::unique_ptr<widget> grid::swap_child(const std::string& id,
-						   widget* w,
-						   const bool recurse,
-						   widget* new_parent)
+widget_ptr grid::swap_child(const std::string& id,
+		widget_ptr w,
+		const bool recurse,
+		widget* new_parent)
 {
 	assert(w);
 
@@ -112,7 +113,7 @@ std::unique_ptr<widget> grid::swap_child(const std::string& id,
 				grid* g = dynamic_cast<grid*>(child.get_widget());
 				if(g) {
 
-					std::unique_ptr<widget> old = g->swap_child(id, w, true);
+					widget_ptr old = g->swap_child(id, w, true);
 					if(old) {
 						return old;
 					}
@@ -123,7 +124,7 @@ std::unique_ptr<widget> grid::swap_child(const std::string& id,
 		}
 
 		// Free widget from cell and validate.
-		std::unique_ptr<widget> old = child.free_widget();
+		widget_ptr old = child.free_widget();
 		assert(old);
 
 		old->set_parent(new_parent);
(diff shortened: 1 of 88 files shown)


src/gui/widgets/grid.cpp
```diff
@@ -89,52 +89,59 @@ void grid::set_child(widget* widget,
...
 			if(recurse) {
-				// decent in the nested grids.
-				grid* g = dynamic_cast<grid*>(child.get_widget());
-				if(g) {
-
-					std::unique_ptr<widget> old = g->swap_child(id, w, true);
-					if(old) {
-						return old;
-					}
+				// Descend into nested grids. We don't handle recursion here for two reasons:
```
CelticMinstrel, line 112: This lengthy comment seems to suggest that the new code is not equivalent to the old code at all? But I have no idea what implications that would have.

Vultraz: It is, I just defer handling child grids until we've checked all widgets... though I'm just realizing the logic still doesn't work, since there may be more than one child grid...

CelticMinstrel: What do you mean, "it is"? How exactly is it equivalent?

Vultraz: It still looks at every widget and potentially sub-grids. I'm not sure why you're saying it's not equivalent at all.

CelticMinstrel: It looks like things happen in an entirely different order, and I have no idea what consequences that might have. It might give a different result in some cases, right?

src/gui/widgets/scrollbar_container.cpp
```diff
@@ -769,10 +769,11 @@ void scrollbar_container::finalize_setup()
 	}
 
 	/***** Setup the content *****/
-	content_ = build_single_widget_instance<spacer>();
+	auto content = build_single_widget_instance<spacer>();
+	content_ = content.get();
 
 	// TODO: possibly move this unique_ptr casting functionality to a helper function.
-	content_grid_.reset(dynamic_cast<grid*>(get_grid().swap_child("_content_grid", content_, true).release()));
+	content_grid_.reset(static_cast<grid*>(get_grid().swap_child("_content_grid", std::move(content), true).release()));
```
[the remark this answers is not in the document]
Vultraz: Only `shared_ptr`.

src/gui/widgets/grid.cpp
CelticMinstrel: I've looked through all the places that call `swap_child` in this diff, and while I'm not 100% sure that's all calls in the entire game, it looks like this is the only place that relies on it returning `nullptr` if nothing was swapped. So… couldn't the owning problem be solved by instead returning the passed-in widget if nothing was swapped? Something like this:

```cpp
widget* test = w.get();
std::unique_ptr<widget> old = g->swap_child(id, w, true)
if(test != old) return old;
```

Vultraz: Well, it's just looking for the widget matching the given ID, so it shouldn't matter in what order it happens.

CelticMinstrel: There could be more than one widget with the ID. There's literally nothing to enforce that IDs are unique. In fact, they're guaranteed _not_ to be unique if you're using anything like a listbox… though I don't know how relevant that is in this specific case.

Although neither the old nor the new could be called a sensible way of handling duplicate IDs, it certainly does change how they'd be handled.

Vultraz: > ```c++
> widget* test = w.get();
> std::unique_ptr<widget> old = g->swap_child(id, w, true)
> if(test != old) return old;
> ```
I like this. Thanks!

src/gui/widgets/scrollbar_container.cpp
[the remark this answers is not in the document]
Vultraz: To elaborate, `content_grid_` is a `unique_ptr<grid>`, and you can't inherently downcast from `uique_ptr<widget>`.

PR #6682, opened 2022-05-04
Use mutexes for hotkey list and image cache access.

mesilliac:
This allows `game_config_manager` to modify these directly, in stead of raising an event to switch to the main thread.

### DIFF
--- a/src/hotkey/hotkey_item.cpp
+++ b/src/hotkey/hotkey_item.cpp
@@ -28,6 +28,7 @@
 #include <SDL2/SDL.h>
 #include <key.hpp>
 #include <serialization/unicode.hpp>
+#include <mutex>
 
 
 static lg::log_domain log_config("config");
@@ -36,9 +37,13 @@ static lg::log_domain log_config("config");
 #define DBG_G  LOG_STREAM(debug, lg::general())
 #define ERR_CF LOG_STREAM(err,   log_config)
 
+using std::lock_guard, std::mutex;
+
 namespace hotkey {
 
 hotkey_list hotkeys_;
+std::mutex hotkeys_mutex_;
+
 game_config_view default_hotkey_cfg_;
 
 namespace {
@@ -327,6 +332,7 @@ void hotkey_keyboard::save_helper(config &item) const
 
 bool has_hotkey_item(const std::string& command)
 {
+	lock_guard<mutex> lock(hotkeys_mutex_);
 	for (hotkey_ptr item : hotkeys_) {
 		if (item->get_command() == command) {
 			return true;
@@ -349,6 +355,7 @@ bool hotkey_keyboard::bindings_equal_helper(hotkey_ptr other) const
 
 void del_hotkey(hotkey_ptr item)
 {
+	lock_guard<mutex> lock(hotkeys_mutex_);
 	if (!hotkeys_.empty()) {
 		hotkeys_.erase(std::remove(hotkeys_.begin(), hotkeys_.end(), item));
 	}
@@ -356,7 +363,6 @@ void del_hotkey(hotkey_ptr item)
 
 void add_hotkey(const hotkey_ptr item)
 {
-
 	if (item == hotkey_ptr()) {
 		return;
 	}
@@ -364,6 +370,7 @@ void add_hotkey(const hotkey_ptr item)
 	scope_changer scope_ch;
 	set_active_scopes(hotkey::get_hotkey_command(item->get_command()).scope);
 
+	lock_guard<mutex> lock(hotkeys_mutex_);
 	if (!hotkeys_.empty()) {
 		hotkeys_.erase(
 			std::remove_if(hotkeys_.begin(), hotkeys_.end(), [item](const hotkey::hotkey_ptr& hk) { return hk->bindings_equal(item); }),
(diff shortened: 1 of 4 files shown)


Vultraz (changes requested):
For consistently, use `std::scoped_lock` sans template (so just `std::scoped_lock locker{cache_mutex_}`)

src/hotkey/hotkey_item.hpp
```diff
@@ -34,8 +35,29 @@ typedef std::shared_ptr<hotkey_base> hotkey_ptr;
 typedef std::shared_ptr<hotkey_mouse> hotkey_mouse_ptr;
 typedef std::shared_ptr<hotkey_keyboard> hotkey_keyboard_ptr;
 
+/**
+ * A threadsafe mutex wrapper for iterating over a hotkey_list.
+ *
+ * This should likely be moved to another file.
+ */
+template <typename T, typename I>
```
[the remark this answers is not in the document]
Vultraz: Or even just give `begin()` and `end()` return type `auto`.

src/hotkey/hotkey_item.cpp
```diff
@@ -432,13 +442,15 @@ void reset_default_hotkeys()
...
-const hotkey_list& get_hotkeys()
+const threadsafe_hotkey_list get_hotkeys()
 {
-	return hotkeys_;
+	return threadsafe_hotkey_list(hotkeys_, hotkeys_mutex_);
 }
 
 void save_hotkeys(config& cfg)
 {
+	lock_guard<mutex> lock(hotkeys_mutex_);
```
stevecotton, line 452: If this lock is necessary then there's likely to be a preference-corrupting bug somewhere. The existence of this lock implies that somewhere there's code which modifies the list, and which might try to run at the same time - adding the lock removes one of these 3 possibilities, but leaves the other two possible:

* modification happens, then list is saved to `cfg`
* list is saved to `cfg`, then modification happens
* (prevented by lock) modification happens while this function is iterating over `hotkeys_`

src/hotkey/hotkey_item.cpp
```diff
@@ -389,11 +396,13 @@ void clear_hotkeys(const std::string& command)
 
 void clear_hotkeys()
 {
+	lock_guard<mutex> lock(hotkeys_mutex_);
 	hotkeys_.clear();
 }
 
 const hotkey_ptr get_hotkey(const SDL_Event &event)
 {
+	lock_guard<mutex> lock(hotkeys_mutex_);
```
stevecotton, line 405: This function returns a pointer, which is accessible after the lock is released.

src/picture.cpp
```diff
@@ -154,6 +155,9 @@ namespace
 {
 image::locator::locator_finder_t locator_finder;
 
+// Use this for anything that may modify an image cache from a thread
```
stevecotton, line 158: If the lock doesn't cover reading, how do we know that data will be valid while reader functions iterate through it?

I haven't looked in depth, but a first impression is that the mutex has been added to too few functions.

src/hotkey/hotkey_item.cpp
mesilliac: The problem mentioned in `game_config_manager` indicated that it was possible an event could modify the hotkey list at the same time as `load_hotkeys()` was being called by the worker thread running `game_config_manager::init_game_config`.

load_hotkeys() can both erase and push_back to the hotkey list, both of which can invalidate the iterator here.

I don't really know how to - or if it's possible to - ensure which things can definitely never be called while another thread is messing with the hotkey list. So i just slapped a mutex on everything that touched it...

src/hotkey/hotkey_item.cpp
mesilliac: The hotkey pointers should still be okay, as far as i understand it it's just access to the hotkey list itself that is problematic.

src/picture.cpp
mesilliac: Yes, it definitely has... the original guard was just on a single function :worried: .

So i'm kind of conflicted between doing it properly, and just leaving the mutex on that one function...

src/hotkey/hotkey_item.cpp
stevecotton: You're correct, sorry.

src/picture.cpp
mesilliac: On looking at it again i guess it does need to be done properly everywhere. That's pretty horrible.

src/hotkey/hotkey_item.cpp
CelticMinstrel: Is it possible for `load_hotkeys` to load into a local list and just swap when done? Might that be simpler?

mesilliac: I actually think it would be better to do that (swapping in stead of directly editing everything in a worker thread) for everything in those threads. But the problem again there is that if something modifies the hotkey list while it is being loaded, then the swap will clobber any changes during that time.

For all of this... i honestly have no idea why anything would ever want to be modifying hotkeys via event while the loading screen is displaying. All i have to go on is the fact that there is an existing block there. Perhaps clobbering is fine? I have no idea.

mesilliac:
I think i am going to give up on this.

I have tried several different approaches to try to get things working such that exceptions can be handled while in the loading screen, and every single one is completely and utterly horrible in some way.

The existing system is also completely horrible. But i do not have the spare time to be working on this.

From what i've seen, the way loading screen worker threads are dealt with needs a complete overhaul. There's no quick and easy fix because it was constructed unsafely in the first place.

PR #6684, opened 2022-05-04
NR: give hamel leadership

### LINKED ISSUE #6683: NR Showdown - Hamel should have leadership

**Version:** 1.17.3
**Downloaded from:**  Source, from here, I think.
- **Build info:** N/A
- **OS:** Devuan GNU/Linux Chimaera

### Description of the bug

Hamel, the Dwarf leader, is the one who trains Tallin in Leadership. Yet, when showing up in the final scenario - Showdown - he doesn't have the Leadership trait. That doesn't make sense IMHO.

N/A

### Expected behavior

Hamel should have Leadership. If you want to penalize the Dwarven faction to account for this, downgrade a unit to a lower level or reduce the Gold slightly etc.

### Additional context

_No response_

knyghtmare:
Closes #6683

### DIFF
--- a/data/campaigns/Northern_Rebirth/scenarios/02_01_Infested_Caves.cfg
+++ b/data/campaigns/Northern_Rebirth/scenarios/02_01_Infested_Caves.cfg
@@ -106,6 +106,17 @@
         name= _ "Hamel"
         profile=portraits/Hamel.webp
         canrecruit=yes
+        [modifications]
+            [object]
+                silent=yes
+                [effect]
+                    apply_to=new_ability
+                    [abilities]
+                        {ABILITY_LEADERSHIP}
+                    [/abilities]
+                [/effect]
+            [/object]
+        [/modifications]
 
         [ai]
             recruitment_pattern=fighter,fighter,fighter,mixed fighter
--- a/data/campaigns/Northern_Rebirth/scenarios/02_01_Infested_Caves.cfg
+++ b/data/campaigns/Northern_Rebirth/scenarios/02_01_Infested_Caves.cfg
@@ -107,6 +107,7 @@
         profile=portraits/Hamel.webp
         canrecruit=yes
         [modifications]
+            {TRAIT_STRONG}
             [object]
                 silent=yes
                 [effect]
--- a/data/campaigns/Northern_Rebirth/scenarios/02_02_Meeting_With_Dwarves.cfg
+++ b/data/campaigns/Northern_Rebirth/scenarios/02_02_Meeting_With_Dwarves.cfg
@@ -48,6 +48,17 @@
         name= _ "Hamel"
         profile=portraits/Hamel.webp
         canrecruit=yes
+        [modifications]
+            [object]
+                silent=yes
+                [effect]
+                    apply_to=new_ability
+                    [abilities]
+                        {ABILITY_LEADERSHIP}
+                    [/abilities]
+                [/effect]
+            [/object]
+        [/modifications]
 
         facing=se
         shroud=yes
--- a/data/campaigns/Northern_Rebirth/scenarios/02_02_Meeting_With_Dwarves.cfg
+++ b/data/campaigns/Northern_Rebirth/scenarios/02_02_Meeting_With_Dwarves.cfg
@@ -49,6 +49,7 @@
         profile=portraits/Hamel.webp
         canrecruit=yes
         [modifications]
+            {TRAIT_STRONG}
             [object]
                 silent=yes
                 [effect]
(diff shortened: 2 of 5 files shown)


Wedge009:
I searched for Hamel in the WML, his last appearance before the big show-down seems to be S4 - Clearing the Mines (which is missed in this PR, presumably because he's recalled differently to the earlier scenarios). The rest seems to be only story-related appearances.

S4 appearance:
https://github.com/wesnoth/wesnoth/blob/a8d145813db3461b67f97386c009f881e0bafb47/data/campaigns/Northern_Rebirth/scenarios/04_Clearing_the_Mines.cfg#L184

S3 appearance:
https://github.com/wesnoth/wesnoth/blob/a8d145813db3461b67f97386c009f881e0bafb47/data/campaigns/Northern_Rebirth/scenarios/03_To_the_Mines.cfg#L38-L50

I assume this doesn't upset balance too much, in this case I think this should be back-ported.

The checks seem to flag a white-space or indentation issue, but I can't see anything arising from this change specifically...

eyalroz:
In the final scenario he's not strong right now, either.

knyghtmare:
> I assume this doesn't upset balance too much, in this case I think this should be back-ported.

Level 3 leadership unit...If I remember my NR correctly, He might throw the balance of S3 and S4 out of the window since dwarven fighters/dwarven scouts/dwarven thunderers will be hitting high damage numbers thanks to +50% leadership (level difference is 2).

This will also influences his performance in S13A where his recruits are level 2s and level 1s and allows him to be a more field leader instead of a keep-seater (as in just recruits and does nothing beyond it)

Perhaps we need to think more closely at this now

eyalroz:
> This will also influences his performance in S13A where his recruits are level 2s and level 1s and allows him to be a more field leader instead of a keep-seater (as in just recruits and does nothing beyond it)

His keep is 3 hexes wide, and the initial defense line basically just that keep anyway. You could make him lose an initial free unit or a bit of money. S3 and S4 may need some thought.

knyghtmare:
image
I stand corrected. This is NOT "3-hexes wide".

eyalroz:
> This is NOT "3-hexes wide".

4 hexes tall, 3 hexes wide. There are extra fortifications outside the keep - one can't recruit in them (plus, if you put people there, you trigger the Orc Warlords near the bridge).

knyghtmare:
> > This is NOT "3-hexes wide".
> 
> 4 hexes tall, 3 hexes wide. There are extra fortifications outside the keep - one can't recruit in them (plus, if you put people there, you trigger the Orc Warlords near the bridge).

Oh i thought you meant overall castle size.

I am checking the AI code for the enemy AI. Seems Hamel is targeted by 33% of each enemy side (not Rakhas), which might be around...50 units in total? maybe more on higher difficulties.
Maybe Leadership will help him survive as I saw that stream of Konrad2 of this scenario and it seemed Hamel needed both sides 7 and 9 to send units to his side to help him survive.

eyalroz:
> Seems Hamel is targeted by 33% of each enemy side (not Rakhas), which might be around...50 units in total? maybe more on higher difficulties.

That's not what I'm seeing on the easiest difficulty. The Dwarf side is seeing the least units attacking it - no more than 5-6 or so within the first 6 turns. I'm seeing Tallin under the most fire, then Krash (Drakes), then Eryssa (Elves), then Hamel.

Wedge009:
> Also, the more important argument is whether changes to NR should be done currently (I was asked this by another contributor) when NR rework is underway...

I've said it before, just because a major redesign is 'planned' should not preclude small changes being done now. We won't know when that redesign will come - if at all - and neglecting small enhancements where it's helpful or necessary may never happen if we take this approach.

> Perhaps we need to think more closely at this [game-play balance] now

Fair enough. If it seems like it could potentially be too disruptive to the balancing aspect, then leave it just in master for now (even if it will ultimately get overridden by the major redesign).

nemaara (approved):

eyalroz:
I've never done any development work, but - do the traits/stats of characters really always get replicated in all scenarios they are in? Is there no `include "hammel.cfg"` -type directive in WML?

knyghtmare:
> I've never done any development work, but - do the traits/stats of characters really always get replicated in all scenarios they are in? Is there no include "hammel.cfg" -type directive in WML?

Normally, we have a character stat loader macro, but it depends on the campaign author/maintainer if it was necessary or not.

Wedge009:
I did a debug-test of the whole campaign and Hamel's appearances appear to have leadership applied now. I suggest this change may be worth recording in the change log. Also, I gather we decided not to back-port this to 1.16?

Issue #1549, opened 2017-05-08
Characters missing in game font due to lack of fall-back options (GNA #24621)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?24621)** by gfgtdf on 2016-04-29

Since pr #639 was merged several dialog show squares instead of the correct charactes, this can be observerd in:

1) The langugage dialog

2) Descrition text of some umc campaigns in the campaings menu

3) The hotkey preferences menu (as also shown in the discussion of pr #639 in https://cloud.githubusercontent.com/assets/4020484/14193188/b48b4800-f770-11e5-9880-ba690d934a9c.png)

(Reproduced on windows)
Release: 1.13.4+dev
Priority: 5 - Normal
Severity: 4 - Important

wesnoth-bugs:
**Modified on 2016-08-29**

**wedge009 wrote:**

> Hot-key markers have been resolved for a while now, but issue is still present in Language options. Predominately Asian languages, I believe.

wesnoth-bugs:
**Modified on 2017-03-26**

**ratarmy added attachment:** [japanese_version_info.PNG](http://gna.org/bugs/download.php?file_id=29902)

wesnoth-bugs:
**Modified on 2017-03-26**

**ratarmy added attachment:** [japanese_help.PNG](http://gna.org/bugs/download.php?file_id=29903)

wesnoth-bugs:
**Modified on 2017-03-26**

**ratarmy wrote:**

> Japanese characters are not displayed correctly in most places.
> 
> Here is a screentshot of the version information dialog in Japanese.
> 
> file #29902
> 
> 
> 
> But, in-game help is fine.
> 
> file #29903

wesnoth-bugs:
**Modified on 2017-03-26**

**wedge009 changed status:** None -> Confirmed

**wedge009 wrote:**

> Yep, I've seen that before.
> 
> 
> 
> Marking as confirmed - probably didn't have ability to do so previously. It is indeed Windows-specific - at least in standard English Windows 7 and earlier, multi-lingual font support isn't really there by default.

wesnoth-bugs:
**Modified on 2017-04-26**

**ratarmy added attachment:** [japanese_font_config_modified.PNG](http://gna.org/bugs/download.php?file_id=29944)

wesnoth-bugs:
**Modified on 2017-04-26**

**ratarmy wrote:**

[...]

Issue #6580, opened 2022-03-25
Font error 


RichardsATcn:
### Game and System Information

- **Version:* 1.16.2*
- **Downloaded from:* steam and sourcefoge*
- **Build info:* *
- **OS:* windows 10 x64*
![1648188790(1)](https://user-images.githubusercontent.com/8119891/160065019-798c7f70-7d77-4e9f-aade-41e7ec0105d7.png)
![1648188816(1)](https://user-images.githubusercontent.com/8119891/160065026-db2d237c-53bf-4b10-b66f-67216687f30e.png)
![1648188790(1)](https://user-images.githubusercontent.com/8119891/160065131-f0ae3514-e2c1-4018-839c-930f6ea6c068.png)
![1648188941(1)](https://user-images.githubusercontent.com/8119891/160065145-ce0d1e51-6e6d-41b7-9c8b-64ae24fdd717.png)
![Uploading 1648188988(1).png…]()

### Description of the bug

any language  Display Error 

### Steps to reproduce the behavior

1.
2.
3.

Wedge009:
The screens appear to be for Simplified Chinese. At first I thought it might be related to #1549 but it doesn't appear to be the case. We'll need more info, because testing 1.16.2 release I could not replicate this in Linux, Windows 10, or Windows 7.

Wedge009:
Without further information I don't think anyone can help with this and the issue could be closed.

PR #6685, opened 2022-05-04
Use fontconfig

Pentarctagon:
See #1549 for the background here. For now just creating this PR so the CI will generate a windows installer with these changes applied to see if it also fixes a potentially related issue.


Pentarctagon:
#6580 could also likely be closed by this as well.

PR #6688, opened 2022-05-06
Add command suggestions

Descacharrado:
Makes suggestions when a command is misspelled as seen below.

untitled
*works in mp chats too.

Explanation:

Other commands will be recommended when a third or less of the letters (of the shortest string, either player input or actual command) are wrong. Additions, deletions, swaps and changes all count as 1 error (1 wrong letter).

- Writing 'hlep' recommends 'help' because there is only one error and the word is four letters (1/4 <= 1/3, one swap, string_utils.cpp Line 375)
- Writing 'hel' recommends 'help' because there is only one error and the word is three letters (1/3 <= 1/3, one deletion)
- Writing 'hello' does not recommend 'help' because there are two errors and the word is four letters (2/3 > 1/3, an addition and a change)
- Writing 'aawhiteboard' does not recommend 'whiteboard', even if there are only two errors out of eleven letters (2/11 <= 1/3) because the implementation is approximated and gets "lost" after two consecutive (non-swap) errors (string_utils.hpp Line 139)

The change comes from https://github.com/wesnoth/wesnoth/pull/6421#issuecomment-1015416479 which comes from #1018

Needless to say that the implementation and thresholds are not set in stone, I even pushed all the commits so anyone can follow along the development to better understand my decisions.

### DIFF
--- a/src/formula/string_utils.cpp
+++ b/src/formula/string_utils.cpp
@@ -357,9 +357,9 @@ std::string vngettext_impl(const char* domain,
 	return msg;
 }
 
-int approximate_string_distance(const std::string &str_1, const std::string &str_2)
+int edit_distance_approx(const std::string &str_1, const std::string &str_2)
 {
-    int approximate_distance = 0;
+    int edit_distance = 0;
     if(str_1.length() == 0) {
         return str_2.length();
     }
@@ -367,32 +367,31 @@ int approximate_string_distance(const std::string &str_1, const std::string &str
         return str_1.length();
     }
     else {
-        int Lmax = 0;
         int j = 0;
-        Lmax = std::max(str_1.length(),str_2.length());
-        for(int i = 0 ; i < Lmax ; i++) {
+        int len_max = std::max(str_1.length(), str_2.length());
+        for(int i = 0; i < len_max; i++) {
             if(str_1[i] != str_2[j]){
                 //SWAP
-                if(str_1[i+1] == str_2[j] && str_1[i] == str_2[j+1]){
+                if(str_1[i+1] == str_2[j] && str_1[i] == str_2[j+1]) {
                     // No need to test the next letter
                     i++;j++;
                 }
                 //ADDITION
-                else if (str_1[i+1] == str_2[j]){
+                else if (str_1[i+1] == str_2[j]) {
                     j--;
                 }
                 //DELETION
-                else if (str_1[i] == str_2[j+1]){
+                else if (str_1[i] == str_2[j+1]) {
                     i--;
                 }
                 // CHANGE (no need to do anything, next letter MAY be successful).
-                approximate_distance++;
-                if(approximate_distance*100/std::min(str_1.length(),str_2.length()) > 33){
+                edit_distance++;
+                if(edit_distance * 100 / std::min(str_1.length(), str_2.length()) > 33) {
                     break;
                 }
             }
             j++;
         }
     }
-    return approximate_distance;
+    return edit_distance;
 }
--- a/src/formula/string_utils.cpp
+++ b/src/formula/string_utils.cpp
@@ -356,3 +356,38 @@ std::string vngettext_impl(const char* domain,
 	const std::string msg = utils::interpolate_variables_into_string(orig, &symbols);
 	return msg;
 }
+
+int approximate_string_distance(const std::string &str_1, const std::string &str_2)
+{
+    int approximate_distance = 0;
+    if(str_1.length() == 0) {
+        return str_2.length();
+    }
+    else if(str_2.length() == 0) {
+        return str_1.length();
+    }
+    else {
+        int Lmax = std::max(str_1.length(),str_2.length());
+        int j = 0;
+        for(int i = 0 ; i < Lmax ; i++) {
+            if(str_1[i] != str_2[j]){
+                //SWAP
+                if(str_1[i+1] == str_2[j] && str_1[i] == str_2[j+1]){
+                    // No need to test the next letter
+                    i++;j++;
+                }
+                //ADDITION
+                else if (str_1[i+1] == str_2[j]){
+                    j--;
+                }
+                //DELETION
+                else if (str_1[i] == str_2[j+1]){
+                    i--;
+                }
+                // CHANGE (no need to do anything, next letter MAY be successful).
+                approximate_distance++;
+            }
+        }
+    }
+    return approximate_distance;
+}
--- a/src/formula/string_utils.cpp
+++ b/src/formula/string_utils.cpp
@@ -386,6 +386,9 @@ int approximate_string_distance(const std::string &str_1, const std::string &str
                 }
                 // CHANGE (no need to do anything, next letter MAY be successful).
                 approximate_distance++;
+                if(approximate_distance > 2){
+                    break;
+                }
             }
         }
     }
--- a/src/formula/string_utils.cpp
+++ b/src/formula/string_utils.cpp
@@ -367,8 +367,14 @@ int approximate_string_distance(const std::string &str_1, const std::string &str
         return str_1.length();
     }
     else {
-        int Lmax = std::max(str_1.length(),str_2.length());
+        int Lmax = 0;
         int j = 0;
+        if (str_1.length() >= str_2.length()){
+            Lmax = str_1.length();
+        }
+        else{
+            Lmax = str_2.length();
+        }
         for(int i = 0 ; i < Lmax ; i++) {
             if(str_1[i] != str_2[j]){
                 //SWAP
@@ -390,6 +396,7 @@ int approximate_string_distance(const std::string &str_1, const std::string &str
                     break;
                 }
             }
+            j++;
         }
     }
     return approximate_distance;
--- a/src/formula/string_utils.cpp
+++ b/src/formula/string_utils.cpp
@@ -369,12 +369,7 @@ int approximate_string_distance(const std::string &str_1, const std::string &str
     else {
         int Lmax = 0;
         int j = 0;
-        if (str_1.length() >= str_2.length()){
-            Lmax = str_1.length();
-        }
-        else{
-            Lmax = str_2.length();
-        }
+        Lmax = std::max(str_1.length(),str_2.length());
         for(int i = 0 ; i < Lmax ; i++) {
             if(str_1[i] != str_2[j]){
                 //SWAP
@@ -392,7 +387,7 @@ int approximate_string_distance(const std::string &str_1, const std::string &str
                 }
                 // CHANGE (no need to do anything, next letter MAY be successful).
                 approximate_distance++;
-                if(approximate_distance > 2){
+                if(approximate_distance*100/std::min(str_1.length(),str_2.length()) > 33){
                     break;
                 }
             }
--- a/src/formula/string_utils.cpp
+++ b/src/formula/string_utils.cpp
@@ -359,39 +359,39 @@ std::string vngettext_impl(const char* domain,
 
 int edit_distance_approx(const std::string &str_1, const std::string &str_2)
 {
-    int edit_distance = 0;
-    if(str_1.length() == 0) {
-        return str_2.length();
-    }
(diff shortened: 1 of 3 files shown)


src/formula/string_utils.cpp
```diff
@@ -356,3 +356,43 @@ std::string vngettext_impl(const char* domain,
...
+{
+    int approximate_distance = 0;
+    if(str_1.length() == 0) {
+        return str_2.length();
+    }
+    else if(str_2.length() == 0) {
+        return str_1.length();
+    }
+    else {
+        int Lmax = 0;
```
CelticMinstrel, line 370: Local variables (or probably any variables) shouldn't start with an uppercase letter. I suggest `l_max` instead, or perhaps `len_max` since lowercase L might be confusing.

src/formula/string_utils.cpp
```diff
@@ -356,3 +356,43 @@ std::string vngettext_impl(const char* domain,
 	const std::string msg = utils::interpolate_variables_into_string(orig, &symbols);
 	return msg;
 }
+
+int approximate_string_distance(const std::string &str_1, const std::string &str_2)
```
CelticMinstrel, line 360: Okay… so is this [Damerau-Levenshtein distance](https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance)? That's cool, adding the transposition test. And I'm kinda surprised we didn't already have a utility like this in the engine…

Although I don't really mind this name, the shorter name `edit_distance` might be a bit nicer.

This is probably out of the scope the PR, but it'd also be nice to expose this in Lua in the `stringx` module.

src/map_command_handler.hpp
```diff
@@ -190,9 +190,30 @@ class map_command_handler
 		}
 		else if (help_on_unknown_) {
 			utils::string_map symbols;
+		    // Get the input command on a string
+            std::string string_user_ = get_cmd();
```
CelticMinstrel, line 194: If I recall correctly, Wesnoth's naming convention is to use a final underscore for member variables, _not_ for local variables. Someone else please confirm or correct me though?

src/map_command_handler.hpp
```diff
@@ -190,9 +190,30 @@ class map_command_handler
...
+            // Get all commands (even if disallowed)
+            std::vector<std::string> command_list_ = get_commands_list();
+            // Initialize the distance and the command_proposal symbol so it
+            // does not appear if no candidate is found.
+            symbols["command_proposal"] = "";
+            int distance_ = 0;
+            // Compare the input with every command (excluding alias)
+            for (typename command_map::value_type i : command_map_) {
```
CelticMinstrel, line 202: Didn't Wesnoth's style guide say no spaces before the parenthesis in control statements? I know the actual code is inconsistent though…

src/formula/string_utils.cpp
```diff
@@ -356,3 +356,43 @@ std::string vngettext_impl(const char* domain,
...
+                else if (str_1[i+1] == str_2[j]){
+                    j--;
+                }
+                //DELETION
+                else if (str_1[i] == str_2[j+1]){
+                    i--;
+                }
+                // CHANGE (no need to do anything, next letter MAY be successful).
+                approximate_distance++;
+                if(approximate_distance*100/std::min(str_1.length(),str_2.length()) > 33){
```
CelticMinstrel, line 390: Could we add spaces around the operators on this line? And maybe after the commas?

src/map_command_handler.hpp
```diff
@@ -190,9 +190,30 @@ class map_command_handler
...
+                // No need to test commands that are not enabled
+                if(is_enabled(i.second)) {
+                    distance_ = approximate_string_distance(string_user_, i.first);
+                    // Maximum of a third of the letters are wrong
+                    if (distance_*100/std::min(string_user_.length(),i.first.length()) < 34){
```
CelticMinstrel, line 207: Again, please add spaces around the operators.

It may also make it more readable to separate the calculation out into steps.

src/map_command_handler.hpp
```diff
@@ -190,9 +190,30 @@ class map_command_handler
...
+                        symbols["command_proposal"] = " did you mean '" + i.first + "'?";
+                        // If a good enough candidate is found, exit the loop.
+                        break;
+                    }
+                }
+            }
 			symbols["command"] = get_cmd();
 			symbols["help_command"] = cmd_prefix_ + "help";
-			print("help", VGETTEXT("Unknown command '$command', try $help_command "
+			print("help", VGETTEXT("Unknown command '$command',$command_proposal try $help_command "
```
CelticMinstrel, line 216: This looks translator-unfriendly. I'd recommend using an entirely separate string if there's a proposal, something like:

```suggestion
			if(has_command_proposal) {
				print("help", VGETTEXT("Unknown command '$command', did you mean '$command_proposal'? Or try $help_command "
			} else {
				print("help", VGETTEXT("Unknown command '$command', try $help_command "
			}
```

Where `has_command_proposal` would be true iff the `symbols["command_proposal"]` is populated with a command. Of course with this version `symbols["command_proposal"]` would be just the command alone, without that "did you mean" prefix.

CelticMinstrel (changes requested):
This looks great! I can't believe no-one has thought to do this before. I see a few small issues that need addressing, though (see specific comments below).

src/formula/string_utils.cpp
Descacharrado: Keep in mind that this is not the full Levenshtein's implementation, it does not do the minimum distance of the string's tails. So I would be reluctant to change the function's name to just `edit_distance` I'd say, at minimum,  `approx_edit_distance` or `edit_distance_approx`.

src/formula/string_utils.cpp
Descacharrado: Now that I gave it a thought, Lmax is only used once, so I will axe it and do `i < std::max(str_1.length(),str_2.length())` directly in the loop condition.

src/map_command_handler.hpp
Descacharrado: I copied the convention from `for (typename command_map::value_type i : command_map_) {` (same file, line 224) So I'm unsure. I'll search for the actual convention guide.

src/map_command_handler.hpp
Descacharrado: Can do. It was one of my concerns, translation issues, as I have no idea how translation works.

src/map_command_handler.hpp
Descacharrado: > It may also make it more readable to separate the calculation out into steps.

I dont really know what you mean by separating the calculation into steps... Do you mean, for example, putting `std::min(string_user_.length(),i.first.length())` into a variable first, and then using that variable in the calculation?

src/formula/string_utils.cpp
CelticMinstrel: Fair enough, either of those would be fine too, though like I said I don't really mind the current name either.

src/map_command_handler.hpp
CelticMinstrel: Yeah, I think what happened is we decided to change the style guide at some point but no-one updated existing code to follow it… either that, or we didn't _have_ a style guide and finally made one and no-one updated existing code.

src/map_command_handler.hpp
CelticMinstrel: Yeah, that's basically what I mean.

CelticMinstrel (approved):

src/map_command_handler.hpp
Descacharrado: Hmmm I wouldn't mind fitting these files to convention. Not in this PR tho. I'm gonna guess there isn't a tool like wmlindent but for C++?

src/map_command_handler.hpp
Descacharrado: Can do.

src/map_command_handler.hpp
CelticMinstrel: There is ./utils/fix_whitespace.sh, which runs wmlindent and also adjusts whitespace in C++ and Lua files, but that only covers leading whitespace (indentation), and the non-wmlindent part I believe just converts tabs to spaces or spaces to tabs rather than actually fixing misindented code.

Besides that, I think there's a clang-format specification in the repository that you could use, though I'm not sure when it was last updated (@Vultraz?).

Vultraz: `clang-format` is up-to-date.

src/map_command_handler.hpp
```diff
@@ -190,10 +190,36 @@ class map_command_handler
 		}
 		else if (help_on_unknown_) {
 			utils::string_map symbols;
+			// Get the input command on a string
+			std::string string_user = get_cmd();
+			// Initialize the distance and the command proposal bool
+			int distance = 0;
+			bool has_command_proposal = false;
+			// Compare the input with every command (excluding alias)
+			for(typename command_map::value_type i : command_map_) {
```
Vultraz, line 199: This should use a structured binding:
```cpp
for(const auto& [key, index] : command_map_)
```
(`i.first` being `key` and `i.second` being `index`)

Vultraz (changes requested):

src/map_command_handler.hpp
stevecotton: `clang-format` is up-to-date, but is also a tool with a user-unfriendly interface. For project-specific settings it doesn't take the options file on the command line, instead it has to read it in the current working directory.

1. Make sure everything's checked in to Git, in case a tonne of unrelated changes appear.
2. In a command prompt, change to Wesnoth's `src` directory (not the root, the directory with the `.clang-format` file in it)
3. Run `clang-format --style=file -i file_to_format.cpp`
    - `--style=file` is entered verbatim, don't change that for a filename
    - The `-i` makes it overwrite the file (and also means you can give multiple files in one run), otherwise it outputs to stdout

Vultraz: If you use VSCode the Format Document/Format Selection command uses `clang-format`. No need to use the command prompt.

Descacharrado: Thanks, I'll have a look at that. I had to have a look at VSCode anyways because of the Msys sdl issue so I may finally change my building tool.

src/map_command_handler.hpp
```diff
@@ -190,10 +190,45 @@ class map_command_handler
 		}
 		else if (help_on_unknown_) {
 			utils::string_map symbols;
+			// Get the input command on a string.
```
Vultraz, line 193: Please remove this comment.

src/map_command_handler.hpp
```diff
@@ -190,10 +190,45 @@ class map_command_handler
...
+			std::string string_user = get_cmd();
+			// Initialize the variables.
+			// Edit distance variable.
```
Vultraz, line 196: Please remove these two comments.

src/map_command_handler.hpp
```diff
@@ -190,10 +190,45 @@ class map_command_handler
...
+			int distance = 0;
+			// Minimum length of the two compared strings.
+			int len_min = 0;
+			// Bool used to show the command proposal if it exists.
```
Vultraz, line 200: Please remove this comment.

src/map_command_handler.hpp
```diff
@@ -190,10 +190,45 @@ class map_command_handler
...
+			bool has_command_proposal = false;
+			// Compare the input with every command (excluding alias).
+			for(const auto& [key, index] : command_map_) {
+				// No need to test commands that are not enabled.
+				if(is_enabled(index)) {
+					// Get the edit distance between input and each command.
```
Vultraz, line 206: Please remove this comment.

src/map_command_handler.hpp
```diff
@@ -190,10 +190,45 @@ class map_command_handler
...
+					distance = edit_distance_approx(string_user, key);
+					// Get the minimum length between the strings.
```
Vultraz, line 208: Please remove this comment.

Pentarctagon:
It looks like this can be merged then?

PR #6691, opened 2022-05-07
compress pngs by at least 30% 

matthiaskrgr:
compress png files via matthiaskrgr/woptipng (rust branch)

Tools used:
imagemagick / convert -strip -defined png:color-type=6
optipng -q -o5 -nb -nc -np
advpng -Z -{1-4}
oxipng --nc --np -o6 --quiet

Compression threshold: 30% (only keep files that get 30%+ smaller)

Optimized 1216 files, reduced 3755130 bytes to 1890944 bytes
3.75 mb -> 1.89 mb


Pentarctagon:
It would probably be good to have someone double check this on a different OS?  Mainly remembering 5d36e382c7919d25e987abf07c40e42e9b2e909e

Pentarctagon:
Also, relatedly, which OS do you use, and have you checked at least a few of the compressed images to make sure they don't show any strange glitches in-game?

doofus-01:
Aren't the `.png` going to be replaced with `.webp` soon anyway?  If so, is there a timeline for that?

CelticMinstrel:
Huh? I thought that was already done…?

doofus-01:
I see.  Well, it wouldn't hurt to convert things at some point, if it can be done without a lot of conflicts and breakage.  Since I touch the terrain graphics and unit sprites, I could/should take that on some day.

Back to the PR:  I think a single pass of compressing png images is of somewhat limited value at this point, since many of the units and terrains will likely get overwritten (at least by #6417, if nothing else), in addition to the webp thing.  Maybe it's still fine though, if that couple of MB are worth touching all those files.

Pentarctagon:
@matthiaskrgr are you around?

Pentarctagon:
Closing due to the lack of response.

PR #6694, opened 2022-05-08
wmllint: add missing monster ID changes

knyghtmare:
Context:

1. Gorer and Tusklet were added as is, and clashed with UMC units, so ID was changed during 1.15 cycle. Wmllint was not notified of this. Some addons using Gorer/Tusklet that time were not updated so I got Lua unknown unit bug. Also, I think better to not have these 2 lines since UMC still use IDs for this. Any comments? (#5662)

2. Giant Stoat and Frost Stoat was quickly renamed but wmllint was never notified. This might be worth adding to wmllint. (#5344) Some UMC still use deprecated Giant Stoat and aren't aware unless they see deprecated message.

### DIFF
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -969,6 +969,11 @@ linechanges = (
         ("Naga Slasher", "Naga Dirkfang"),
         ("Naga Bladewhirler", "Naga Ophidian"),
 
+        # monster unit ID changes during 1.15.x cycle
+        ("Gorer", "Woodland Boar"),
+        ("Tusklet", "Piglet"),
+        ("Giant Stoat", "Frost Stoat"),
+
         # Changed in 1.15.0: separate portrait for leader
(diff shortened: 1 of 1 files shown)


Elvish-Hunter:
Regarding the Gorer -> Woodland Boar and Tusklet -> Piglet changes, the situation is complex. When a unit changes its ID, a wmllint rule is usually added; however, in this case the change was made to prevent clashing with UMCs, and these rules are precisely going to clash against UMCs. It's probably better to remove them.
Giant Stoat -> Frost Stoat can certainly be added.

knyghtmare:
> Regarding the Gorer -> Woodland Boar and Tusklet -> Piglet changes, the situation is complex. When a unit changes its ID, a wmllint rule is usually added; however, in this case the change was made to prevent clashing with UMCs, and these rules are precisely going to clash against UMCs. It's probably better to remove them.

This is why I asked for your comment here. I guess I will just patch the UMC manually then. (The Pool of Ek).

PR #6695, opened 2022-05-09
1.17 waterworld

doofus-01:
This was inspired by a semi-recent conversation about underwater scenarios.  It doesn't directly address that, but it does lay some groundwork.  So far, this has:
1. Another water creature/monster
2. Seaweed terrain - I'm not totally sure how to deal with this, for now it is a "swamp" overlay.
3. Seashells - a simple embellishment overlay

All WIP, especially the seaweed/kelp


data/core/terrain.cfg
```diff
@@ -128,6 +128,19 @@
     editor_group=water
 [/terrain_type]
 
+[terrain_type]
+    symbol_image=water/seaweed/kelp-tile
+    id=sea_kelp
+    name= _ "Swamp"
+    editor_name= _ "Kelp Forest"
```
CelticMinstrel, line 135: Shouldn't name and editor_name be the same?

doofus-01: No, the editor names seem to be more descriptive.  I don't know that "Swamp" is necessarily the right thing here, but at least it is informative of the current situation.

CelticMinstrel: Okay, sure, maybe editor name is more descriptive, but… it doesn't seem correct for it to be completely different as it is in this case.

stevecotton:
@Pentarctagon should these all be webp instead of png?

Pentarctagon:
It depends on if they're smaller or not.

doofus-01:
Does something like #6697 make it not matter what image type is called anymore?  Otherwise, replacing the png with webp needs to be its own little project, unrelated to this.

Pentarctagon:
Another issue that came up recently as well, is it turns out that MediaWiki (or at least our version of it) currently doesn't display webp images when linked on the wiki. This would then break https://wiki.wesnoth.org/TerrainCodeTableWML, for example. So further mass conversions to webp will need to wait until that's addressed.

doofus-01:
I plan to write a couple higher level Caribe, but not for now.  The kelp is supposed to be useful for water expanses;  I don't think it adds much for a crowded coast or delta.  The other stuff is pretty minor, but hopefully still a small step of progress.

I intend to merge if this passes CI.

PR #6701, opened 2022-05-12
Remove some dead code (maybe a merge defect?)

bencsikandrei:
* small formatting changes
* small rearrange

### DIFF
--- a/src/animated.tpp
+++ b/src/animated.tpp
@@ -78,12 +78,8 @@ inline void animated<T>::start_animation(int start_time, bool cycles)
 	started_ = true;
 	last_update_tick_ = get_current_animation_tick();
 	acceleration_ = 1.0; // assume acceleration is 1, this will be fixed at first update_last_draw_time
-	start_tick_ = last_update_tick_ + static_cast<int>((starting_frame_time_ - start_time) / acceleration_);
-
+	start_tick_ = last_update_tick_ + (starting_frame_time_ - start_time);
 	cycles_ = cycles;
-	if(acceleration_ <= 0) {
-		acceleration_ = 1;
-	}
 	current_frame_key_ = 0;
 	force_next_update_ = !frames_.empty();
 }
@@ -129,8 +125,9 @@ inline void animated<T>::update_last_draw_time(double acceleration)
 		}
 	}
 
-	if(get_current_frame_end_time() < get_animation_time() && // catch up
-			get_current_frame_end_time() < get_end_time()) {  // don't go after the end
+	const int current_frame_end_time = get_current_frame_end_time();
+	// catch up && don't go after the end
+	if(current_frame_end_time < get_animation_time() && current_frame_end_time < get_end_time()) {
 		current_frame_key_++;
 	}
 }
@@ -223,7 +220,7 @@ inline int animated<T>::get_animation_time() const
 	}
 
 	int time = tick_to_time(last_update_tick_);
-	if (time > max_animation_time_ && max_animation_time_ > 0) {
+	if(time > max_animation_time_ && max_animation_time_ > 0) {
 		return max_animation_time_;
 	}
 	return time;
@@ -386,8 +383,8 @@ template<typename T>
 inline void animated<T>::set_end_time(int new_ending_time)
 {
 	int last_start_time = starting_frame_time_;
-	typename std::vector<frame>::iterator current_frame = frames_.begin();
-	while(last_start_time < new_ending_time && current_frame != frames_.end()) {
+	auto current_frame = frames_.cbegin();
+	while(last_start_time < new_ending_time && current_frame != frames_.cend()) {
 		last_start_time += current_frame->duration_;
 		++current_frame;
 	}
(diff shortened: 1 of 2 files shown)


bencsikandrei:
There's a couple of other changes that could be made in these files.

But I need some guidance, please, whether you'd like me to try them out or not.

One example is the class layout. As it is now, if I counted correctly, on a 64 bit arch the animated class takes 72B (it's virtual so, 8 bytes bonus)
        int starting_frame_time_;
        bool does_not_change_; // Optimization for 1-frame permanent animations
	bool started_;
	bool force_next_update_;
        // 1 B padding
	std::vector<frame> frames_;
	int max_animation_time_;
	int start_tick_;
	bool cycles_;
        // 7B paddding
	double acceleration_;
	int last_update_tick_;
	int current_frame_key_;

Some small rearrangements, like so:
        int starting_frame_time_;
        bool does_not_change_; // Optimization for 1-frame permanent animations
	bool started_;
	bool force_next_update_;
	bool cycles_;
	std::vector<frame> frames_;
	int max_animation_time_;
	int start_tick_;
	double acceleration_;
	int last_update_tick_;
	int current_frame_key_;

Takes it down to 64, which fits a cache line.

Is this something of interest?

Have a good one!

bencsikandrei:
In the same spirit as another PR I made, maybe a separate PR with replacing `typedef` with `using` would be interesting?

src/animated.tpp
```diff
@@ -78,12 +78,8 @@ inline void animated<T>::start_animation(int start_time, bool cycles)
 	started_ = true;
 	last_update_tick_ = get_current_animation_tick();
 	acceleration_ = 1.0; // assume acceleration is 1, this will be fixed at first update_last_draw_time
-	start_tick_ = last_update_tick_ + static_cast<int>((starting_frame_time_ - start_time) / acceleration_);
-
+	start_tick_ = last_update_tick_ + (starting_frame_time_ - start_time);
 	cycles_ = cycles;
-	if(acceleration_ <= 0) {
```
bencsikandrei, line 84: It is pretty late here, but I'm not missing something here, right?

animated is set to 1, at line 80, then never changed and then checked for <= 0

I guess it was a simple merge request gone weird

src/animated.tpp
```diff
@@ -130,7 +126,7 @@ inline void animated<T>::update_last_draw_time(double acceleration)
 	}
 
 	if(get_current_frame_end_time() < get_animation_time() && // catch up
-			get_current_frame_end_time() < get_end_time()) {  // don't go after the end
+		get_current_frame_end_time() < get_end_time()) {      // don't go after the end
```
Vultraz, line 129: Can you put these two conditions on one line.

Pentarctagon:
> Takes it down to 64, which fits a cache line.
> 
> Other things can be envisioned, like double -> float, but the alignment requirements will make it so that that doesn't help. Maybe > some of the ints (4B) can be smaller ints? Maybe I'm going too far :D
> 
> Is this something of interest?

While optimizations are good, I'd say this is overkill unless the performance improvements are measurable in some way.

src/animated.tpp
stevecotton: Your change looks good to me. Looks as if this should have been removed in 20bf7fa2e9b56df39ee53667e96200f9d4f51f82, when this function stopped taking the acceleration as a parameter.

CelticMinstrel:
Did we have a policy regarding braces on if statements? I forget. It's not that I'm opposed to adding them there, mind you… just mixing style changes with not-style changes is something I prefer not to happen. Although… these may not be style changes but they _are_ basically non-functional changes, other than making the constructor explicit.

> > Takes it down to 64, which fits a cache line.
> > Other things can be envisioned, like double -> float, but the alignment requirements will make it so that that doesn't help. Maybe > some of the ints (4B) can be smaller ints? Maybe I'm going too far :D
> > Is this something of interest?
> 
> While optimizations are good, I'd say this is overkill unless the performance improvements are measurable in some way.

Or to put it another way… profile it.

bencsikandrei:
@Pentarctagon I'll try to see the impact it has. If the struct is used very often or some vector if it iterated often it's probably worth it

@CelticMinstrel I didn't mean to change that one if, I thought I undid that one, whoops! I'm on board with consistency also, I guess it was just some reflex, working on it

@Pentarctagon @CelticMinstrel if you have some quick tips on how I can benchmark effectively and be quick about, I'd highly appreciate it, otherwise I'm gonna continue on my road to knowing the project and will figure it out eventually :D Thanks!

I'll make the changes you suggested ASAP, thanks for the feedback!

CelticMinstrel:
Just FTR, I don't mind adding the braces there… if I'm hesitant of anything, it's adding them in the same commit (or PR?) as this other stuff. It's only one change, so it's not that big a deal, but still.

Pentarctagon:
@bencsikandrei There's profiler support in cmake and scons if you're on Linux, otherwise I don't have any advice on it though.

src/animated.tpp
bencsikandrei: done

Issue #3391, opened 2018-07-28
Wesnoth terminates if an error occurs during editor reload


CelticMinstrel:
If you reload using F5 while in the editor, and a WML error occurs, Wesnoth closes. It should instead either dump you to the main menu or offer an option to try again after you've fixed the issue.

Pentarctagon:
@CelticMinstrel is this still the case after #6702?

CelticMinstrel:
I'll check, uhhh… soon? Please remind me if I don't get back to this in a few days.

Pentarctagon:
*reminds*

CelticMinstrel:
Yes, it still crashes if it encounters malformed WML. It does offer the usual "without addons" retry though. I only tried introducing an error in core WML (editor-specific stuff). I assume it doesn't crash if the error is in an add-on.

It's particularly bad to crash in this case because there could be unsaved work. I think it would be better if the WML error dialog had "Retry" and "Quit" buttons, or something like that.

Issue #3218, opened 2018-06-07
threading issues


gfgtdf:
a recent tsan run dicovered multiple objects that are aceses from both threads during the loadginscreen leading to possible crashes:

1) 
https://github.com/wesnoth/wesnoth/blob/fc2a58f6935176b70a035d3e1c03080f79687f94/src/tstring.cpp#L46
and

https://github.com/wesnoth/wesnoth/blob/fc2a58f6935176b70a035d3e1c03080f79687f94/src/tstring.cpp#L47
are changesd when a tstrign is creates which s unavoidable whiel reading the cfg files, the main screen also creates tstring object ot at least translates them during probably for showing the current state.

2) 
https://github.com/wesnoth/wesnoth/blob/2bf4d68c87b69256266b7745a9e3453693794e35/src/font/font_config.cpp#L101
and the 3 objects in the lines below re written in the worker thread at font::load_font_config() and read in the main gui thread by gui2::canvas::render() -> gui2::text_shape::draw -> ... -> font::pango_text::calculate_size()

3)
the image cahce is accessed by the ui thread to display the images of the tiltescreen and by the worker thread at
```

    #5 image::locator::init_index() src/image.cpp:282 (wesnoth-debug+0x1662831)
    #6 image::locator::locator(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) src/image.cpp:348 (wesnoth-debug+0x1662d51)
    #9 frame_builder::frame_builder(config const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) src/units/frame.cpp:110 (wesnoth-debug+0xb238cc)
    #10 unit_animation::particle::particle(config const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) src/units/animation.cpp:941 (wesnoth-debug+0xaaabab)
    #11 unit_animation::unit_animation(config const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) src/units/animation.cpp:293 (wesnoth-debug+0xa9f957)
    #16 unit_animation::add_anims(std::vector<unit_animation, std::allocator<unit_animation> >&, config const&) src/units/animation.cpp:663 (wesnoth-debug+0xaa7431)
[...]

Issue #1603, opened 2017-05-08
Game crashes when exiting the loading screen in an unusual way (GNA #25290)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?25290)** by gfgtdf on 2016-11-12

I thought there was already a bugreport for this one but i couldn't find one.

This was observed my multiple persons: when a the loadingscreen is exited by an exception (which for example when you close wesnoth during loadingscreen, but also if it closed becasue of a gui exception like 'dialog didn't fit on screen') the game sometimes crashes.

What happens in this case is that the gui::tloadscreen object is destroed but the worker thread didn't finish yet and the mainthread just detaches the workingthread which obviously can cause crashes.

There are at least 2 known approaches to fix this issue:

1) when exting during a loadingscreen by an exception we just call exit(), this should be rather easy to implement.

2) We try to exit the worker thread. Since just killing it coudl result in corrupt objects/memory, we'd need to add 'safe exit points' to the worker code e.g. calls to somethin like 'tloadscreen::maybe_throw_exit_thread_exception()'.

Another problem with 2) is that it is afaik currently not possible to add 'quit callbacks' in gui2 dialogs (to overwrite the defautl bahviour of throwing CVideo::quit())

(Reproduced on all)
Release: 1.13
Priority: 5 - Normal
Severity: 4 - Important

wesnoth-bugs:
**Modified on 2016-11-16**

**vultraz wrote:**

> The game will also crash with WML errors if you go through the full cycle of 'try load without addons' without fixing the error.

wesnoth-bugs:
**Modified on 2016-11-16**

**gfgtdf wrote:**

> > The game will also crash with WML errors if you go through the full cycle of 'try load without addons' without fixing the error.
> 
> 
> 
> Not sure what you mean here, erros casues by the actual loading (like syntax erros in the wml files) should't dont casue crashes currently. There was just a bug that was fixed in https://github.com/wesnoth/wesnoth/commit/b8467cc0c2cd3f6a59c1176ef2119f4db8365413 though.

wesnoth-bugs:
**Modified on 2016-11-16**

**gfgtdf wrote:**

> s/just a bug/such a bug

wesnoth-bugs:
**Modified on 2017-03-13**

**wedge009 changed title:** Game crashes when exiting the loadingscreen �na unusual way -> Game crashes when exiting the loading screen in an unusual way

[...]

PR #6702, opened 2022-05-13
De-thread the loading screen. Load in the main thread.

mesilliac:
The programming interface is still the same. Draw and input events are now processed when the worker function calls loading_screen::progress().

This has two main repercussions:
1. exceptions will no longer be silently eaten by the loading screen,
2. loading screen animations will not update unless progress() is called.

In order to update animations in a timely fashion... well, for me they already update so fast as to be nigh-instant. I'm not really sure what ever takes any significant time to load.

If there are actually sections of loading code that take significant time, some sort of `loading_screen::spin()` function shouild be added, that simply lets the loading screen update the spinner animation. Worker functions should call this periodically while performing long-duration tasks.

An alternate option would be to use a progress bar in stead of a spinner animation. Such that when `loading_screen::progress()` is called, it updates both the name of the loading section and the overall progress amount.

In any case, with this change it becomes the responsibility of the worker function to relinquish control momentarily so that events can be processed. As worker functions were already regularly calling `loading_screen::progress()`, with the changes i made here they do this automatically but with relatively low frequency.

The main thing that is potentially lost here is that if a loading function deadlocks for some reason, it becomes impossible to close the program via the close button. I have no idea if there is actually any chance of this happening, nor if it is really a problem. My OS offers to kill programs that are not responding to input anyway. And if there is a chance of a loading operation deadlocking, it should probably have a time-out attached.

I think doing things this way is preferable to the existing situation, and also to the major architectural overhaul that would be needed to do it properly using threads. I have previously tried four separate solutions to salvage the existing threaded implmentation, and all would be far more work than i am prepared to put in, as well as changing core game systems in dangerous ways, which i do not want to do.

### DIFF
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -470,9 +470,16 @@ static bool remove_on_resize(const SDL_Event& a)
 // TODO: I'm uncertain if this is always safe to call at static init; maybe set in main() instead?
 static const std::thread::id main_thread = std::this_thread::get_id();
 
+// this should probably be elsewhere, but as the main thread is already
+// being tracked here, this went here.
+bool is_in_main_thread()
+{
+	return std::this_thread::get_id() == main_thread;
+}
+
 void pump()
 {
-	if(std::this_thread::get_id() != main_thread) {
+	if(!is_in_main_thread()) {
 		// Can only call this on the main thread!
 		return;
 	}
@@ -868,7 +875,7 @@ void peek_for_resize()
 
 void call_in_main_thread(const std::function<void(void)>& f)
 {
-	if(std::this_thread::get_id() == main_thread) {
+	if(is_in_main_thread()) {
 		// nothing special to do if called from the main thread.
 		f();
 		return;
--- a/src/events.hpp
+++ b/src/events.hpp
@@ -126,6 +126,9 @@ void focus_handler(const sdl_handler* ptr);
 
 bool has_focus(const sdl_handler* ptr, const SDL_Event* event);
 
+// whether the currently executing thread is the main thread.
+bool is_in_main_thread();
+
 void call_in_main_thread(const std::function<void (void)>& f);
 
 //event_context objects control the handler objects that SDL events are sent
--- a/src/game_config_manager.cpp
+++ b/src/game_config_manager.cpp
@@ -47,6 +47,8 @@ static lg::log_domain log_config("config");
 #define WRN_CONFIG LOG_STREAM(warn, log_config)
 #define LOG_CONFIG LOG_STREAM(info, log_config)
 
+using gui2::dialogs::loading_screen;
+
 static game_config_manager* singleton;
 
 game_config_manager::game_config_manager(const commandline_options& cmdline_opts)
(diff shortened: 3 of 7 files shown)


Vultraz:
I strongly disagree with this direction. Certainly, there are benefits from using a single thread, but design-wise this is a step backwards.

1. We shouldn't be manually raising draw events and running the event loop. I don't know if we'll ever actually get to this point - though we are getting closer - but the event loop should be run in one spot only and dispatch draw events each go-around. We can figure out what the most optimal way to cache things *between* draws is, but we shouldn't actually be invoking drawing ourselves.
2. The `spin` function itself. It is not optimal to make animations rely on an external caller to update, especially since now we need two places to raise draw events separately. It makes the whole draw pipeline unnecessarily complicated. We should be able to go from an event loop, look at draw hooks, and know they're getting called once per frame.
3. Non-maintainability and contamination of concerns. Right now, the loading screen is one of the few places in the UI where we really have animations because of the lack of accelerated rendering and how absolutely clunky it is to work with animations with the dirty widget interface. It makes the LS animation handling unnecessarily complicated to tie the drawing code into the loading code, when they should be distinct and separate. Why should the game config manager care about whatever animation the loading screen is displaying? You could argue This method preserves that separation since `spin` only raises draw events and `draw_callback` still handles the logic, but it still means we need to keep track of where we should add `spin` calls. I don't see how this is an upgrade from mutexes. We already have `progress`. Let's not add more.
4. The multiplayer connection screen. If we switch to this method, the entire multiplayer connection workflow (https://github.com/wesnoth/wesnoth/blob/master/src/game_initialization/multiplayer.cpp#L177-L231) needs to be redesigned. Right now it relies on being able to pause the LS worker to wait on the result of a third sub-worker. I never fully finished implementing the network worker for MP (incoming data packets are already handled in their own thread) - I still want to get back to that someday - but redesigning the connection process (including everything in `open_connection` would be significant work in and of itself. And I do think it's rather elegant right now, if I do say so myself 😬

Pentarctagon:
While conceptually I do agree that having a separate thread so that loading can happen in the background while the GUI remains responsive, this has also caused its own problems (#3218, #1603, #3391 from a few minutes looking through existing issues, plus a known but apparently undocumented issue that holding F5 on the main menu will eventually crash Wesnoth).  Additionally, based what mesilliac has said from the other ways he's attempted to resolve this, there would seem to be significant other technical issues with trying to keep the current threaded implementation.

So assuming there aren't other blockers, such as potentially the multiplayer connection screen that vultraz mentioned, I'd be in favor of going back to a non-threaded implementation and then taking another look at how threading could be implemented safely (or if it's worth the risk of other issues coming up).

soliton-:
> we need to keep track of where we should add spin calls. I don't see how this is an upgrade from mutexes.

You do not see how it's easier to add a couple of function calls at appropriate places and the worst that happens if you miss some is no GUI updates for a bit in contrast to not locking a critical region and getting random crashes or worse? And once you do add all the needed mutexes? Worse performance for every code path involved whether it runs concurrently to the loading screen thread or not.

The point is avoiding running large parts of wesnoth code in parallel for little gain and significant penalties in robustness (as evidenced by quite a few bugs that have cropped up since), maintainability (avoiding future bugs) and performance and instead sprinkle function calls to update the GUI in that same code. Unless you can somehow reduce the amount of code that runs in parallel to the loading screen it seems quite difficult to call threading the better design.

Regarding the multiplayer connection code I don't see why you couldn't just add a bool flag or something to pause GUI updates similarly.

mesilliac:
As discussed on Discord previously, and as there has been no alternative provided, i'm going to push this basically as is.

I added a basic system for allowing the loading screen to animate while waiting for network data.

There will be parts that i have not marked. These will appear to "lock up". All these need is to call `loading_screen::spin()` while they are doing whatever they are doing that is taking so long.

For network connections this is usually a matter of using "wait_for" in stead of "wait", and updating the loading screen inside a loop. For file loading, this is mostly a matter of updating the loading screen inside loops that cycle many times.

Issue #6403, opened 2021-12-27
Issues found with wmltools by the vghetto port project


cooljeanius:
Thanks to @virtualghetto for this thread: https://forums.wesnoth.org/viewtopic.php?t=55180

Important posts within the thread:
- [669921](https://forums.wesnoth.org/viewtopic.php?p=669921#p669921)
- [670221](https://forums.wesnoth.org/viewtopic.php?p=670221#p670221)
- [670232](https://forums.wesnoth.org/viewtopic.php?p=670232#p670232)

Issues with wmltools found by it that I don't think are fixed yet:
- [ ] They should "remove bomb and change file formats from dos to unix" (vghetto currently uses vim to do so)
- [ ] `wmllint-1.4` may change windmill villages into regular villages (`^Vwm` -> `^Vh`), which is wrong
- [ ] `wmlindent` chokes on tags inside macro calls, like `{([stuff][/stuff])}` (I've noticed this one myself)
- [ ] `wmllint` doesn't automatically update difficulties in `_main.cfg`
- [ ] `wmllint` doesn't remove obsolete lines like `{SOUND:POISON}`, `{SOUND:SLOW}`, or `ellipse="misc/ellipse"` (`{MAGENTA_IS_THE_TEAM_COLOR}` was in this list previously, too, but that was handled in 912625b), or AI parameters like `recruitment_ignore_bad_combat`, `recruitment_ignore_bad_movement`, `attack_depth`, or `{ATTACK_DEPTH}`
- [ ] `wmllint` should do automatic replacements of parameters that previously took `true`/`false` values but now take `yes`/`no` values (e.g. `mirror`, `hide_help`, and `flies`)
- [ ] `wmllint` should rewrite AI `[target][/target]` subtags to use the new `[goal][/goal]` syntax
- [ ] `wmllint` should rewrite the objective `note=` key to use the `[note][/note]` subtag instead
- [ ] `wmllint` should rewrite `[option]message=[/option]` to use new syntax
- [ ] `wmllint` should rewrite usage of the `{FOREACH}` and `{NEXT}` macros (like `{FOREACH locations j}` and then `{NEXT j}`) to use `[for][do][/do][/for]` tags instead
- [ ] `wmllint` should add `{AMLA_DEFAULT}` to units that have `advances_to=null` and remove `{AMLA_DEFAULT}` from units that have `advances_to=something`
[...]

PR #6703, opened 2022-05-13
Wmllint: Handle removal of {SOUND:SLOW} and {SOUND:POISON}

knyghtmare:
Refer to point number five #6403


Elvish-Hunter:
So... The first thing I noticed here is that there's some code duplication going on; even the comments from the `{MAGENTA_IS_THE_TEAM_COLOR}` block were copy-pasted!
The correct approach would've been to place all the obsolete lines into a tuple and update the handling code to iterate that tuple, updating the regexp and the output message as needed. I actually already have a patch which does exactly this and I'm going to push it after the next development version is tagged.
For this reason, I'm closing this PR.

PR #6704, opened 2022-05-13
EI S01 Move the campfire and give it a little more range

### LINKED ISSUE #5766: Sound sources that are on-screen at the start of the scenario don't play
Local Linux builds of:
* 1.14.16
* 1.15.10
* Current master with various PRs merged in (this is a bug being logged in the middle of testing them)

### Describe the bug
If a `[sound_source]` is on-screen at the beginning of a scenario, it doesn't make any sound until scrolled off-screen and back on again.

1. Make Wesnoth's window large enough to have the full width of a 35-hex-wide map on-screen
2. Turn music off (probably not necessary, but makes step 4 easier)
2. Start the campaign Eastern Invasion and skip through the intro
3. You should hear the morning cockerel and turn bell
4. Check that the campfire on 33,13 is visible (without moving the map, if it isn't visible then restart with a bigger window)
5. Note that you don't hear any sound from the campfire
6. Make the window smaller, and scroll the campfire off-screen
7. Scroll back, and note that you can now hear the campfire

### Expected behavior
Crackling sound of the campfire right from the start.

### Additional context
Still reproducible even with the sound-effects volume set to 100% in the preferences (compare with #3280).
I haven't tested 1.14.3, 1.14.4, however #3280's reproduction steps suggest this bug wasn't present back then.

Descacharrado:
The campfire in S01 would not sound on certain zoom levels because of it's range. I fixed it by moving the campfire and giving all campfires in the campaign a bit more range. Tested S04b and it sounds fine.

Another issue I found while doing this is that when you change the zoom without moving the camera, the state of the sound is not updated (so if you were in max zoom, listening to the campfire and then zoomed out, the sound would not change until you moved the camera the tiniest of bits), maybe it's intentional, or an oversight.

Maybe it would be good to have an option to scale the sound source depending on zoom levels, something like:
```
[sound_source]
        scale_with_zoom=yes
        ...
[/sound_source]
```

Closes #5766


Pentarctagon:
@macabeus Why did the Map Diff job fail?

stevecotton:
Does it also depend on the screen size? I haven't grasped the details yet, but #5766 feels like a general engine bug, rather than one to fix whack-a-mole style in each scenario that can trigger it.

(Which isn't to say that doing whack-a-mole fixes isn't useful, the current PR looks like a good change.)

Descacharrado:
> I haven't grasped the details yet, but #5766 feels like a general engine bug, rather than one to fix whack-a-mole style in each scenario that can trigger it.

I tried moving the campfire near the player's keep to see if the sound would trigger: https://youtu.be/Jf5aGBACvjc 
As it does trigger, it's fair to say that the issue at hand (#5766) is scenario specific, more on that below.

> Does it also depend on the screen size? 

It depends on how much of the map is being shown, wether it is because of zoom and/or resolution. One example that may clear things out: https://www.youtube.com/watch?v=gfxl2xDP0mg
See how, when having a sound source on the edge of the map and changing only the resolution of the game (not the zoom level) the sound source stops playing. This is what I tried to convey on https://github.com/wesnoth/wesnoth/issues/5766#issuecomment-1087894757 but maybe failed to do. 

The distance to the sound source **seems** to be calculated using the hex in the center of the screen, and you cannot make the edge of the map be the center of the screen.

In conclusion, I would not say this is a bug, I would say it is scenario specific and would recommend against puting sound sources near map edges in the documentation.

macabeus:
@Pentarctagon I'm looking at why it failed 😢 
The GitHub action got to render the image: https://i.imgur.com/TtzdwUo.jpeg
But it failed to add the comment. I'm checking.

knyghtmare (approved):

nemaara (approved):

Issue #6715, opened 2022-05-17
[MacOS] Mouse and Screen UI elements are not Mapped accurately...at all


knyghtmare:
- **Version:**
- **Downloaded from:**
- **Build info:**
- **OS:**
- 1.17.4
- SourceForge
- Client Download
- MacOS Big Sur 11.6.3

**Description of Bug:**

Screen UI buttons/elements are not being accurately with the mouse and the deviations are rather significant.

<img width="768" alt="Screenshot 2022-05-17 at 12 39 15 PM" src="https://user-images.githubusercontent.com/5283677/168745411-db7f12f6-fdab-4253-86c1-da4c083a5a03.png">

<img width="815" alt="Screenshot 2022-05-17 at 12 37 58 PM" src="https://user-images.githubusercontent.com/5283677/168745216-ad02cb39-172e-4430-a6d6-4a0f6e69f217.png">

<img width="750" alt="Screenshot 2022-05-17 at 12 39 33 PM" src="https://user-images.githubusercontent.com/5283677/168745471-5fcaf906-9e93-4d31-8bf8-106d49bcabed.png">

I got curious and installed a fresh 1.17.4 client on my macbook Pro...and I got the same issue.
The mouse cursor is somewhere near the location of Knalga on the Main Menu's map screen but Multiplayer is highlighted/selected

<img width="1440" alt="Screenshot 2022-05-17 at 12 21 50 PM" src="https://user-images.githubusercontent.com/5283677/168745702-03391768-70b8-4a29-bd38-704179ac39c1.png">

Lastly, it's not just the main menu...it appears the entire Wesnoth app is affected by this.
I tried altering the display options and settings in SysPrefs but nothing restored it.

Wesnoth 1.16.2 client seems unaffected and is working as usual.

This is causing frustration and confusion from the MacOS player-base (including me), and if it's affecting _every_ Mac user, I think this is quite serious.

Note: my brother has a MacPro and he has the same issue with 1.17.4. So, 3 cases and counting...

(Is this High Priority? The app is almost unplayable on MacOS right now due to this)

Pentarctagon:
Does this happen even if the scaling is set to 1x?

Pentarctagon:
That'd make sense.

@CelticMinstrel does this also happen for you?

Pentarctagon:
I wonder if this might be at all related to the Apple-specific adjustment that was added to get_dpi() previously?
[...]

PR #6718, opened 2022-05-19
Potential fix for hardware-rendering high-dpi issues on Mac.

mesilliac:
Aimed at fixing #6715

### DIFF
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -195,11 +195,11 @@ void CVideo::update_framebuffer()
 		return;
 	}
 
-	// Find max valid pixel scale at current window size.
-	point wsize(window->get_size());
+	// Find max valid pixel scale at current output size.
+	point osize(window->get_output_size());
 	int max_scale = std::min(
-		wsize.x / preferences::min_window_width,
-		wsize.y / preferences::min_window_height);
+		osize.x / preferences::min_window_width,
+		osize.y / preferences::min_window_height);
 	max_scale = std::min(max_scale, preferences::max_pixel_scale);
 
 	// Determine best pixel scale according to preference and window size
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Just noting here what I mentioned on IRC: this appears to fix it for me.

mesilliac:
I've pushed another potential fix for the crash that was occurring when disabling pixel scale. I'm doing this blind, but that was probably the culprit...

CelticMinstrel:
Yup, looks like that fixes it.

src/video.cpp
```diff
@@ -539,8 +539,8 @@ std::pair<float, float> CVideo::get_dpi_scale_factor() const
 	if(dpi.first != 0.0f && dpi.second != 0.0f) {
 		// adjust for pixel scale
 		SDL_Point wsize = window_size();
-		dpi.first /= wsize.x / get_width();
-		dpi.second /= wsize.y / get_height();
+		dpi.first *= float(get_width()) / wsize.x;
```
Vultraz, line 542: Only thing is `static_cast` is preferred to C-style casts but that can be cleaned up separately. If this PR fixes the issue I say merge ASAP.

CelticMinstrel: This is technically "constructor-style", not "C-style". Not that that changes your point substantially.

Pentarctagon: Is there something that could be set to catch these by the compiler? For C style casts for example, there's `-Wold-style-cast`.

mesilliac: `static_cast` for a basic numeric type? Why? That just makes code messier and more difficult to read.

mesilliac: I just checked and `-Wold-style-cast` is already set and does indeed warn if you try to do something dodgy.

CelticMinstrel: Well, personally I prefer constructor-style casts for this type of thing. Clearly Vultraz (and some other people) disagree.

Pentarctagon: FTR I don't really have an opinion either way, I just prefer enforcing things as much as possible in the CI so that it's kept consistent - assuming this is meant to be a coding convention of Wesnoth's code base.

mesilliac: I don't think this makes sense as a coding convention. If the CI treats warnings as errors it will already catch dangerous casts, such as casting pointers or structs.

In cases like the one here it is an extra 13 characters (including angle brackets) each use for no benefit.

Pentarctagon:
Any reason not to go ahead and merge this?

mesilliac:
Not that i'm aware of. I just don't like merging my own PRs

PR #6722, opened 2022-05-22
1.17 some orc standing animations

doofus-01:
Some standing animations, no particular grouping.
EDIT:  It looks like it will be orcs.


doofus-01:
I'll merge once this passes CI.

PR #6723, opened 2022-05-23
Editor: Add missing items and scenery

### LINKED ISSUE #6719: [FR] [Editor - Scenario Mode] [1.17] Add missing items and scenery

Context: PR #5821 added +100 items and scenery assets into the core image directory but there is an oversight which is that said assets not being registered into editor items to be used in the scenario editor.

```ini
    [item]
        id=altar
        image=items/altar.png
        name= _ "Altar"
    [/item]
```
File to Edit: https://github.com/wesnoth/wesnoth/blob/master/data/core/editor/items.cfg

This is rather trivial in nature, but I think such oversights should not be happening...
Anyways, it's mostly just copy/pasting, editor and then cross-checking to which are missing and which are already there, thus seems quite okay for a first-timer.

Wedge009:
Resolves #6719. I didn't review #5821, I used the files that are actually present in `items` and `scenery` directories.

I'm not sure if the spelling correction for the `holly-water` ID will break anything (I couldn't find any references to it in `cfg` files). Please advise if I got anything wrong for the Lit Brazier, I'm not familiar with the animation syntax.

Back-port will need to wait for 1.16.3 string freeze to end.

### DIFF
--- a/data/core/editor/items.cfg
+++ b/data/core/editor/items.cfg
@@ -25,6 +25,12 @@
     [/item]
 
     [item]
+        id=ankh-necklace2
+        image=items/ankh-necklace2.png
+        name= _ "Ankh Necklace"
+    [/item]
+
+    [item]
         id=anvil
         image=items/anvil.png
         name= _ "Anvil"
@@ -52,6 +58,24 @@
         id=axe
         image=items/axe.png
         name= _ "Axe"
+    [/item]
+
+    [item]
+        id=axe2
+        image=items/axe-2.png
+        name= _ "Axe"
+    [/item]
+
+    [item]
+        id=axe-small
+        image=items/axe-small.png
+        name= _ "Small Axe"
+    [/item]
+
+    [item]
+        id=axe-throwing
+        image=items/axe-throwing.png
+        name= _ "Throwing Axe"
     [/item]
 
     [item]
@@ -127,8 +151,26 @@
     [/item]
 
     [item]
+        id=bow
+        image=items/bow.png
+        name= _ "Bow"
+    [/item]
+
+    [item]
+        id=bow2
+        image=items/bow-2.png
+        name= _ "Bow"
+    [/item]
+
+    [item]
         id=bow-crystal
         image=items/bow-crystal.png
+        name= _ "Crystal Bow"
+    [/item]
+
+    [item]
+        id=bow-crystal2
+        image=items/bow-crystal-2.png
         name= _ "Crystal Bow"
     [/item]
 
@@ -139,9 +181,9 @@
     [/item]
 
     [item]
-        id=bow
-        image=items/bow.png
-        name= _ "Bow"
+        id=bow-elven2
+        image=items/bow-elven-2.png
+        name= _ "Elven Bow"
     [/item]
 
     [item]
@@ -151,14 +193,39 @@
     [/item]
 
     [item]
+        id=branch
+        image=items/branch.png
+        name= _ "Branch"
+    [/item]
+
+    [item]
+        id=branch-bare
+        image=items/branch-bare.png
+        name= _ "Bare Branch"
+    [/item]
+
+    [item]
         id=brazier
         image=items/brazier.png
         name= _ "Brazier"
     [/item]
 
     [item]
+        id=brazier-lit1
+        halo=items/brazier-lit[1~8].png:100
+        image=items/brazier-lit1.png
+        name= _ "Lit Brazier"
+    [/item]
+
+    [item]
         id=buckler
         image=items/buckler.png
+        name= _ "Buckler"
+    [/item]
+
+    [item]
+        id=buckler2
+        image=items/buckler-2.png
         name= _ "Buckler"
     [/item]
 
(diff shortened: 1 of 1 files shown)


Wedge009:
Failing Windows checks seems to be an issue with vcpkg downloads, not this change.

CelticMinstrel:
I'm not 100% sure but I think that ID isn't actually used anywhere except in the editor internally. Maaaybe it might show up in saved scenario files? But I bet the editor uses the image path to identify the items in saved scenarios, rather than the ID.

knyghtmare (approved):
Items -> Missing `[item]` for `sword-short.png`

can be merged as is but the missing item's part would be nice.

CelticMinstrel:
I have no idea what you're saying in that post…

Wedge009:
I was confused at first as well. I was just missing the entry for the Short Sword - that's been added now.

data/core/editor/items.cfg
```diff
@@ -174,6 +241,18 @@
...
+    [item]
+        id=chest
+        image=items/chest.png
+        name= _ "Closed Chest"
+    [/item]
+
+    [item]
+        id=chest-open
+        image=items/chest-open.png
+        name= _ "Open Chest"
```
stevecotton, line 253: Could use a po hint that "Open" is an adjective, this isn't a menu option to open a chest.

data/core/editor/items.cfg
```diff
@@ -186,24 +265,54 @@
...
     [item]
         id=coffin-closed
         image=items/coffin-closed.png
         name= _ "Closed Coffin"
     [/item]
 
+    [item]
+        id=coffin-open
+        image=items/coffin-open.png
+        name= _ "Open Coffin"
```
stevecotton, line 289: Po hint for this one please.

data/core/editor/items.cfg
```diff
@@ -186,24 +265,54 @@
...
+    [item]
+        id=dagger-poison
+        image=items/dagger-poison.png
+        name= _ "Poison Dagger"
+    [/item]
+
+    [item]
+        id=dagger-poison-bare
+        image=items/dagger-poison-bare.png
+        name= _ "Bare Poison Dagger"
```
stevecotton, line 313: It's more "a normal dagger, which happens to match the image used for the Poison Dagger, so would look in-place in an assassin's workshop where the poison is being added". Maybe "Poison Dagger (clean)"?

Po hint also needs to say that "bare" doesn't mean the verb "to pull out a knife and threaten someone with it".

data/core/editor/items.cfg
```diff
@@ -216,6 +325,12 @@
         name= _ "Flaming Sword"
     [/item]
 
+    [item]
+        id=flame-sword-bare
+        image=items/flame-sword-bare.png
+        name= _ "Bare Flaming Sword"
```
stevecotton, line 331: It looks a lot more like an ice-sword than anything to do with flames.

knyghtmare: If we have to go add fancy words, I suggest "Unenchanted"/"Disenchanted" Flame Sword.

data/core/editor/items.cfg
```diff
@@ -437,6 +654,66 @@
...
+    [item]
+        id=torch
+        image=items/torch.png
+        name= _ "Torch"
+    [/item]
+
+    [item]
+        id=torch-bare
+        image=items/torch-bare.png
+        name= _ "Bare Torch"
```
stevecotton, line 715: "Unlit Torch" or "Torch (unlit)"

Wedge009:
If there are no further suggestions/concerns, I'll probably merge this soon.

data/core/editor/items.cfg
```diff
@@ -408,6 +585,42 @@
...
+    [item]
+        id=staff-magic-wand
+        image=items/staff-magic-wand.png
+        name= _ "Magic Wand"
+    [/item]
+
+    [item]
+        id=staff-plain
+        image=items/staff-plain.png
+        name= _ "Plain Staff"
```
CelticMinstrel, line 621: I wonder if maybe this should be changed to "quarterstaff"? Looking at it, I _think_ that's what it's meant to be, right? I'm not 100% sure though…

CelticMinstrel:
I notice there are some groups with the same name, but I guess they're just obvious variants, right?

data/core/editor/items.cfg
Wedge009: I recognise there may be a difference between the name that players see and the file naming for the sake of grouping similar items together. I wouldn't know what the original intent of each image is, though. Maybe @doofus-01 can comment on what the intention was? I note from https://opengameart.org/content/wesnoth-items that it's just labelled as a 'plain, unmagical staff', not necessarily an implication that it's meant for physical martial arts, though.

Wedge009:
You mean item1 and item2 both being called 'Item' in the user-visible string? I just followed the precedents that are already there. eg There are book variations already present that are all labelled just as 'Book'.

data/core/editor/items.cfg
doofus-01: I'd say it doesn't matter.  These are just images for content creators, they can use them in whatever way makes sense.

knyghtmare: I agree with doofus-01. The id= and name= fields do not matter here since after placement the image is WML-coded as 
```ini
[item]
    x=21
    y=21
    image="items/staff-plain.png"
[/item]
```

Wedge009: Thanks for confirming. So the important thing is the path name.

Wedge009:
Will back-port after 1.16.3 string freeze has ended.

PR #6726, opened 2022-05-23
Fix building with GCC 13.

marxin:
Fixes the following issue:
`base64.hpp:23:49: error: 'uint8_t' was not declared in this scope`

and:
`deprecation.hpp:20:22: error: found ':' in nested-name-specifier, expected '::'`

### DIFF
--- a/src/deprecation.hpp
+++ b/src/deprecation.hpp
@@ -14,6 +14,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <string>
 
 /** See https://wiki.wesnoth.org/CompatibilityStandards for more info. */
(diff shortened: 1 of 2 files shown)


src/deprecation.hpp
```diff
@@ -15,6 +15,7 @@
 #pragma once
 
 #include <string>
+#include <cstdint>
```
stevecotton, line 18: Please keep the includes in alphabetical order (unless that requires reordering a lot of already-existing lines).

marxin: Sure, fixed.

stevecotton:
Thanks, that looks good. I'm running the CI as a formality, so just a heads-up that the Windows builds are currently expected to fail due to `https://github.com/microsoft/vcpkg/issues/24876` ; if you get a notification that the build passed on every platform except the two Windows ones then there's no need to worry.

PR #6727, opened 2022-05-25
Potential fix for flickering and corrupted graphics on master branch.

mesilliac:
Changed to render everything to an offscreen texture rather than allowing SDL to do whatever it wants.

This isn't all that undesirable anyway.

From my earlier research, the only thing that doesn't support rendering to texture is some very very old graphics hardware. So likely these days, nothing.


Wedge009:
I had flickering on Linux/X11 but not Windows (AMD GPUs in both cases). This PR seems to resolve the flickering for me.

PR #6730, opened 2022-05-28
[CI TEST] nyeeeeee

Vultraz:


stevecotton:
That commit message will be completely useless for anyone looking through the history later. I'm not against commit messages being "amusing", but that can't be at the expense of usefulness.

Just "Fix mingw build" would be much better.

Vultraz:
This is not going to be the final commit message :P

PR #6733, opened 2022-05-30
Editor palettes: allow keyboard scrolling to see the last row

### LINKED ISSUE #6724: Editor: palette's scroll down button can't scroll to last row of items

- **Version:** At least 1.16.2 (current release at time of report)
- **Downloaded from:** Official Steam release + self-compiled
- **Build info:** MSVC/SCons
- **OS:** Windows and Linux

### Description of the bug

I noticed this while testing #6723 on a laptop (no scroll wheel). The items/tiles panel in the bottom-right side of the Editor window allows the user to scroll via buttons or the mouse wheel. However, I can only access all tiles with the mouse wheel scrolling method.

1. Start Wesnoth
2. Enter Editor
3. By default the 'all tiles' option should be shown in the bottom-right panel. Assuming there are more tiles than can be shown in the panel at once (I tested at 1920x1080 and 2560x1440), click the down button to scroll to the end of the available tiles. The last available tile is the 'Fake Map Border'.
4. Use the mouse wheel to scroll even further down. This exposes the two more tiles which are the _actual_ last tiles available: 'Fake Shroud' and 'Fog'.

### Expected behavior

Using the buttons to scroll down should show all available tiles/items. I suspect this inconsistency may have been missed if most developers are using a mouse wheel to scroll through the tiles.

### Additional context

I have a suspicion that there might be a mis-count in the number of items/tiles to render. The user can assign one tile each to the left-click and right-click. Could this be throwing off how the tiles/items are arranged in this panel?

stevecotton:
Fixes #6724, where scrolling with the mousewheel could reach the last row, but
scrolling with keys or buttons couldn't. Although can_scroll_up() and
can_scroll_down() already existed, can_scroll_down() didn't allow access to the
last line when that line had less than the full number of items. The mousewheel
still worked, because it ignored that logic.

Remove a feature that was seen when the last row is visible - things jumped
between columns to fill the spaces. If you have 10 items, with 4 columns and 2
visible rows then it will appear like this:
```
    a b c d
    e f g h
```
when scrolling down it used to move items sideways to fill the last line, which
seems bad UI because it made items harder to find:
```
    c d e f
    g h i j
```
the new code is simpler, and will instead show the following when scrolling down:
```
    e f g h
    i j
```

Move the layout code into adjust_size(), so that it runs one time when the number
of buttons changes. That could be separated from this commit, but the code
would still be touched in this commit (`counter_from_zero` would still be
replaced by `i`), so doing it and testing the changes together made sense.

Note for the master branch only: scrolling the palettes up and down is
noticeably laggy, which is a regression from last week. That's not caused by
this commit, we're just at a point in the rendering refactor where
`surface_restorer::update()` and thus `gui::widget::hide()` are slow. The fix
for that is already in progress and going to be in 1.17.5.

### DIFF
--- a/src/editor/palette/common_palette.hpp
+++ b/src/editor/palette/common_palette.hpp
@@ -78,7 +78,7 @@ public:
 	virtual void expand_palette_groups_menu(std::vector<config>& items, int i) = 0;
 
     //item
-	virtual int num_items() = 0;
+	virtual std::size_t num_items() = 0;
 	virtual std::size_t start_num() = 0;
 	virtual void set_start_item(std::size_t index) = 0;
 
--- a/src/editor/palette/editor_palettes.cpp
+++ b/src/editor/palette/editor_palettes.cpp
@@ -75,16 +75,19 @@ void editor_palette<Item>::expand_palette_groups_menu(std::vector<config>& items
 template<class Item>
 bool editor_palette<Item>::scroll_up()
 {
-	int decrement = item_width_;
-	if (items_start_ + num_visible_items() == num_items() && num_items() % item_width_ != 0) {
-		decrement = num_items() % item_width_;
+	bool scrolled = false;
+	if(can_scroll_up()) {
+		// This should only be reachable with items_start_ being a multiple of columns_, but guard against underflow anyway.
+		if(items_start_ < columns_) {
+			items_start_ = 0;
+		} else {
+			items_start_ -= columns_;
+		}
+		scrolled = true;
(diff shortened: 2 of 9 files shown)


Vultraz (approved):

Wedge009 (approved):
I confirm the slow-down in the editor scrolling was already present in master. I also confirm this change accomplishes what I would expect the behaviour of the tiles to be when scrolling - thanks for that.

PR #6734, opened 2022-05-31
Sync rendering system changes so far

mesilliac:
Primary visible changes:
* high-DPI fonts now work for all GUI2 items - so most of the UI. It's very nice.
* some images are also rendered in high-dpi, notably the title screen background
* the previous physical-DPI-based scaling has been disabled. Pixel scale is used exclusively for now. The old UI scaling code is still there, for future use.

Architectural changes:
* There is a new API for drawing things to the screen. It can be found in `draw.cpp/hpp`. Usage is straightforward, `draw::line(from, to)` draws a line, for example. `draw::flipped(tex, where)` draws a mirrored texture (by default horizontally, but with optional explicit direction flags). All functions drawing to the screen have been updated to use this new API. The code was consolidated from several places.
* The image cache has a stub for returning textures rather than surfaces. It doesn't actually cache textures yet, but it eventually will. This is just to assist conversion. `get_image` has been deprecated in favour of `get_surface` and `get_texture` to explicitly request that type of image storage. Where it was obvious, i have updated calls to use one or the other. Some calls to `get_image` remain, where they will be overhauled later anyway, or where the choice wasn't obvious.
* Some things have been changed from using separate mirrored or scaled textures to mirroring and scaling on-the-fly. More work remains to be done changing things over.
* GUI2 elements now draw directly to the screen, in stead of caching their canvas as a separate SDL surface.

What remains to be done:
* game display still needs to be converted
* GUI1 also needs some more converting
* image cache needs to be updated to manage textures
* there should probably be some sort of cache for rendered text, especially calculated sizes
* a new system for handling redraws without surface restorers / screen reads needs to be put in place
* a lot more

Oh, i also reimplemented map screenshots. But that will still remain broken until the game display is converted over to the new hardware accelerated rendering system, as it doesn't mix well with the old drawing system in this specific case.

### DIFF
--- a/data/gui/window/wml_message.cfg
+++ b/data/gui/window/wml_message.cfg
@@ -154,7 +154,7 @@ where calculated_width = {__GUI_IMAGE_WIDTH}
 					w = "{__GUI_IMAGE_DISPLAYED_WIDTH}"
 					h = "{__GUI_IMAGE_DISPLAYED_HEIGHT}"
 					name = "(portrait_image)"
-					vertical_mirror = "(portrait_mirror)"
+					mirror = "(portrait_mirror)"
 				[/image]
 
 			[/draw]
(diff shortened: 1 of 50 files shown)


mesilliac:
The unit tests seem to be failing, because it's trying to draw things without a window. Does it need to do that? :/ why is it doing that?

PR #6742, opened 2022-06-03
Store add-on download count in the database.

Pentarctagon:
The idea being that this could then also be included as a sheet in the multiplayer activity report, which would then be an easy way to view add-on downloads over time. This would of course not be especially useful until 1.18, since for 1.17 the database count would be starting at 0 even though existing add-ons would have a non-zero download count.

---

Assuming the CI passes, and my own local testing I'm planning to do later today works, I'll go ahead and merge this.


utils/mp-server/table_definitions.sql
```diff
@@ -134,6 +134,7 @@ create table game_content_info
 -- FORUM_AUTH: whether forum authentication is to be used when uploading
 -- UPLOADED_ON: when the addon was uploaded
 -- FEEDBACK_TOPIC: the forum topic ID where feedback for the addon can be posted, 0 if not set
+-- DOWNLOAD_COUNT: the number of times the add-on has been downloaded by players (does not count downloads from addons.wesnoth.org)
```
soliton-, line 137: should say https://addons.wesnoth.org or it's confusing.

PR #6756, opened 2022-06-08
merge some commits from the base of new_rendering, should fix problems with rendering on MacOS in master

mesilliac:
This is to address some bugs preventing the game from working on MacOS, raised on Discord.

It removes the fake compatibility framebuffer that was being used to maintain old code that expected to be able to draw directly to the screen. However it does so in a way that is likely to be very bad for performance. Or rather, the subsequent commits that bring performance back into line haven't been pulled in to this PR.

It needs some brief testing from someone with a less fast system than i have. Previously there were reports of slowness in the editor, but i don't experience it. Two things might cause slowness: scrolling around a large map in game, and scrolling the tool pallete in the map editor. If these are acceptable, then this should be merged.

Actually that editor issue might also be in current master... i'm not sure. It seems fine to me.

Most of this is converting things that used to use `SDL_Surface` to use `SDL_Texture` in stead.

There are several "TODO" comments added. Most of these are already resolved in the `new_rendering` branch.

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -895,6 +895,10 @@ const std::string& get_direction(std::size_t n)
 
 void display::create_buttons()
 {
+	if(screen_.faked()) {
+		return;
+	}
+
 	menu_buttons_.clear();
 	action_buttons_.clear();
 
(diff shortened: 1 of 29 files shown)


mesilliac:
This should now be passing all tests. (that Windows one is just failing to download something...)

I added a renderer to the fake display used for the boost unit tests. It seems to work fine. It's rendering to an offscreen texture, so theoretically it might even be able to render things for tests, if that's ever useful / relevant.

The WML validation tests are still using a headless setup with no renderer (which i think is what you get if you run wesnoth with --nogui). I got this working just by sprinkling some extra checks around, so things don't try to draw in headless mode.

I'll probably rebase and merge this along with #6760 tomorrow, assuming there are no objections.

PR #6757, opened 2022-06-08
Workflow testing

mesilliac:
This is just for messing around with CI workflows and colour diagnostics support.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -48,6 +48,7 @@ jobs:
           export CXX_STD=17
           export CFG=release
           export LTO=true
+          export CLICOLOR_FORCE=1
           ./.github/workflows/ci-scripts/ubuntu.sh
 
   ubuntu-2004-cmake-clang-debug:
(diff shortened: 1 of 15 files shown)


CelticMinstrel:
I like ff2eba36942f0daba99a8c7c289e5db589113c2a’s commit message.

CMake doesn’t seem to be colouring the progress lines green like I remember it doing when I run it from the command-line. (Not for Wesnoth but for other projects.) But errors are in red…

Are you planning to apply colours to Wesnoth's logging too or is that out of scope for this?

mesilliac:
re: the lines in green, i haven't yet found where i can force it to do that, but if i can i will.

I did get the boost unit tests, and the AI play test running with colour output for gdb. I was also considering adding colour to some other commands, yeah. It's pretty simple to have them directly output the ANSI colour codes, and that's what i usually do in my build scripts.

For the WML tests, an option to pass to wesnoth to tell it to print error messages in colour is probably the best way.

I can also add colour to the backtrace gdb runs in the WML unit tests, but i wasn't sure what that was used for. It didn't seem to trigger at all in the run i was looking at earlier that asserted.

mesilliac:
Colourful backtraces seem to be working, now that i've reenabled the docker `--tty` option. They are much, much easier to read.

I made a few things more colourful and obvious, like the headers for the different test sections (Play tests, WML validation, etc).

Some things could have more colour added. But i'll probably tidy this up and push what i have for now. Now that it's working and the mechanisms are in place, it can be extended to other parts as desired.

PR #6760, opened 2022-06-09
Colourful CI output

mesilliac:
Adds colours to CI output in a variety of places. Most notably:

* compiler error messages (this was actually already working, but now it's explicit)
* cmake progress stages
* test section headers (these were also made more distinct with a border and spacing)
* test section failure messages
* WML validation failures
* gdb backtraces
* boost test suite stages are now displayed, in colour (they weren't previously displayed at all)

The coloured outputs only really work in the linux builds. I'm not sure what the equivalent commands are on mac and windows, if any.

I also ran into a couple other problems during this, so fixed them in some additional commits.

* game::error messages weren't always being displayed on exit
* the MP Test executor was not actually running MP Tests, but duplicating Play Tests, so i fixed that as well

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -48,6 +48,7 @@ jobs:
           export CXX_STD=17
           export CFG=release
           export LTO=true
+          export CLICOLOR_FORCE=1
           ./.github/workflows/ci-scripts/ubuntu.sh
 
   ubuntu-2004-cmake-clang-debug:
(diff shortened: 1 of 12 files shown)


mesilliac:
Actually now that the boost test suites are displayed, maybe the error log shouldn't be redirected to a file then pasted afterwards... it's not checking for anything, it just cats the stderror output at the end after everything has finished.

An example with some failed tests and colourful backtraces:
https://github.com/wesnoth/wesnoth/runs/6805214607

PR #6763, opened 2022-06-11
Fix text positioning when clipped to a viewport.

mesilliac:
Should fix the lua console, and the credits screen.


Wedge009:
I confirm credits viewport no longer shrinks, with this change applied.

PR #6764, opened 2022-06-11
Fix wmlindent quote detection

CelticMinstrel:
There's been a persistent wmlindent warning in the CI logs for awhile now, so I thought I'd fix it. This prevents wmlindent from treating commented-out quotes as opening or closing a string.

I also removed several ignore directives that were clear workarounds for the bug, as well as a clearly-pointless pair in terrain-graphics.cfg. As you can see, there are no actual changes in the wmlindent run as a result, other than the removal of trailing whitespace.

### DIFF
--- a/data/tools/wmlindent
+++ b/data/tools/wmlindent
@@ -103,6 +103,29 @@ class bailout(Exception):
         self.lineno = lineno
         self.msg = msg
 
+def strip_comment(line, in_string):
+    """
+    Strip a trailing comment from a line, if present.
+    Allows for unbalanced quotes, where comment characters are literal.
+    """
+    if not line:
+        return line
+    split = line.split('"')
+    result = split[0]
+    if not in_string and '#' in result:
+        return result.split('#')[0].rstrip();
+    split.pop(0)
+    for component in split:
+        in_string = not in_string
+        result += '"'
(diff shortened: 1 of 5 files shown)


data/tools/wmlindent
```diff
@@ -103,6 +103,29 @@ class bailout(Exception):
...
         self.msg = msg
 
+def strip_comment(line, in_string):
+    """
+    Strip a trailing comment from a line, if present.
+    Allows for unbalanced quotes, where comment characters are literal.
+    """
+    if not line:
+        return line
+    split = line.split('"')
```
soliton-, line 113: Does this properly handle doubled double quotes in strings?
Like: `attr="foo ""bar"" baz"`

CelticMinstrel: I believe so, but I'm not sure if there were any of those in my testing cases…

CelticMinstrel: Just checked. Python's string.split() does indeed yield empty elements, so for example:

`some_key="This is a ""test""!!!" # With a comment at the end`

would split to 

`['some_key=', 'This is a', '', 'test', '', '!!!', ' # With a comment at the end']`

And from there the logic should work correctly.

PR #6768, opened 2022-06-13
1.17 fauna work

doofus-01:


doofus-01:
Really not feeling the Roc.  The unit itself has almost no effort invested, can we drop it?  How much does UMC use it?

doofus-01:
Thanks to f00ee60 I'm closing this out, once it passes CI.

PR #6769, opened 2022-06-13
Render most in-game text in high-dpi automatically

mesilliac:
This is partially just a compatibility wrapper for the font rendering system, so that it can return high-res textures with the intended draw-space resolution indicated. But in doing that i also had to fix some other things to get consistent results. So there are four things here:
1. Automatic high-dpi text (works for in-game UI and help browser, for example).
2. Floating game text such as damage numbers has been scaled up according to in-game zoom level for consistent size. This is in addition to also now rendering in high-dpi.
3. Outlines for floating labels have also been scaled up according both to pixel scale and zoom level.
4. Outlines for in-hex text (such as defense % when hovering a unit move) have also been scaled up.

3 and 4 are distinct because they use completely different systems to do the outline.

Everything seems to be working fine for me, but it could use some testing.

I'm not completely sure on the high-dpi game text. It might be out of place next to low-dpi sprites? It seems to look good in practice though. The examples below using both pixel scale 2 and zoom 2 are rather extreme. I expect zoom levels in practice will be much closer to 1. At a glance the text should appear similar at all zoom and pixel scaling levels, with this PR.

<details>
<summary>For some comparisons... here's what damage text in master currently looks like</summary>

pixel scale 2 zoom level 1 (master):
flzoom-master-hp-ps2z1
pixel scale 2 zoom level 2 (master):
flzoom-master-hp-ps2z2

and with this PR:
pixel scale 2 zoom level 1 (PR):
flzoom-pr-ps2z1
pixel scale 2 zoom level 2 (PR):
flzoom-pr-hp-ps2z2

static label and movement text, in current master:
pixel scale 1 zoom 1 (master):
flzoom-master-ps1z1
pixel scale 2 zoom 1 (master):
flzoom-master-ps2z1
pixel scale 2 zoom 2 (master):
flzoom-master-ps2z2

and with this PR:
pixel scale 1 zoom 1 (PR):
flzoom-pr-ps1z1
pixel scale 2 zoom 1 (PR):
flzoom-pr-ps2z1
pixel scale 2 zoom 2 (PR):
flzoom-pr-ps2z2

</details>

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -1493,8 +1493,10 @@ void display::draw_text_in_hex(const map_location& loc,
 {
 	if (text.empty()) return;
 
-	const std::size_t font_sz = static_cast<std::size_t>(font_size * get_zoom_factor());
+	const double zf = get_zoom_factor();
+	const int font_sz = int(font_size * zf);
 
+	// TODO: highdpi - use the same processing as floating_label::create_texture() and cache the effect result so it doesn't constantly rerender the same thing.
 	// TODO: highdpi - perhaps this could be a single texture with colour mod, in stead of rendering twice?
 	texture text_surf = font::pango_render_text(text, font_sz, color);
 	texture back_surf = font::pango_render_text(text, font_sz, font::BLACK_COLOR);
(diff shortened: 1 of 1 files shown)


Pentarctagon:
I'd guess most people would prefer readable if slightly out of place text vs harder to read text.

mesilliac:
I did some more testing and it seems okay. I'm going to push this, and it can be tweaked later according to any feedback.

PR #6771, opened 2022-06-14
Fix some SoTA zombie recruit dialog errors

mesilliac:
There was an error `error gui/parse: horizontal_grow and horizontal_alignment can't be combined, alignment is ignored.` when opening the recruit dialog for corpses in SoTA.

I removed horizontal_grow in favour of horizontal_alignment here, as that matches what is used in other blocks. There seems to be no visible difference.

There were also some unnecessary `..` which seemed to be attempting to concatenate strings in WML. It looks like this is a lua thing? They were visible in-game like "..Traits.." and "..Attacks..". I removed them and it seems to work fine now?

### DIFF
--- a/data/campaigns/Secrets_of_the_Ancients/gui/zombie_recruit_dialog.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/gui/zombie_recruit_dialog.cfg
@@ -78,7 +78,6 @@
                 border = "top,bottom"
                 border_size = 10
                 horizontal_alignment="left"
-                horizontal_grow=true
                 [image]
                     id = "large_unit_sprite"
                 [/image]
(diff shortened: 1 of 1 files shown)


Pentarctagon:
Yeah it looks like someone got that mixed up, or maybe missed it if the dialog was converted from lua to WML data.

data/campaigns/Secrets_of_the_Ancients/gui/zombie_recruit_dialog.cfg
```diff
@@ -168,7 +168,7 @@
                 [label]
                     definition ="default_small"
                     use_markup = "true"
-                    label = "<b>" .. _ "Traits" .. "</b>"
+                    label = "<b>" _ "Traits" "</b>"
```
CelticMinstrel, line 171: These dots should be replaced with `+`. It was probably an error from converting the dialog to Lua.

PR #6776, opened 2022-06-16
Support latin hotkeys in non-latin layouts

renom:
Different approach to solve #5421. All the latin hotkeys are supported in non-latin layouts.
In comparison with #5598, this PR isn't supposed to affect special character hotkeys.

### DIFF
--- a/src/hotkey/hotkey_item.cpp
+++ b/src/hotkey/hotkey_item.cpp
@@ -293,9 +293,9 @@ bool hotkey_keyboard::matches_helper(const SDL_Event& event) const
 	const hotkey_command& command = get_hotkey_command(get_command());
 
 	if((event.type == SDL_KEYDOWN || event.type == SDL_KEYUP)
-		&& (mods & KMOD_CTRL || mods & KMOD_ALT || mods & KMOD_GUI || command.toggle || CKey::is_uncomposable(event.key))
+		&& (mods == KMOD_NONE || mods & KMOD_CTRL || mods & KMOD_ALT || mods & KMOD_GUI || command.toggle || CKey::is_uncomposable(event.key))
 	) {
-		return event.key.keysym.sym == keycode_ && mods == mod_;
+		return (event.key.keysym.sym == keycode_ || (SDL_GetScancodeFromName(SDL_GetKeyName(event.key.keysym.sym)) == SDL_SCANCODE_UNKNOWN && SDL_GetKeyFromName(SDL_GetScancodeName(event.key.keysym.scancode)) == keycode_)) && mods == mod_;
 	}
 
 	if(event.type == SDL_TEXTINPUT && !command.toggle) {
@@ -304,7 +304,7 @@ bool hotkey_keyboard::matches_helper(const SDL_Event& event) const
 		if(text == ":" || text == "`") {
 			mods = mods & ~KMOD_SHIFT;
 		}
-		return text_ == text && utf8::size(std::string(event.text.text)) == 1 && mods == mod_;
+		return utf8::size(text) == 1 && text_ == text && mods == mod_;
 	}
 
 	return false;


src/hotkey/hotkey_item.cpp
```diff
@@ -293,9 +293,9 @@ bool hotkey_keyboard::matches_helper(const SDL_Event& event) const
 	const hotkey_command& command = get_hotkey_command(get_command());
 
 	if((event.type == SDL_KEYDOWN || event.type == SDL_KEYUP)
-		&& (mods & KMOD_CTRL || mods & KMOD_ALT || mods & KMOD_GUI || command.toggle || CKey::is_uncomposable(event.key))
+		&& (mods == KMOD_NONE || mods & KMOD_CTRL || mods & KMOD_ALT || mods & KMOD_GUI || command.toggle || CKey::is_uncomposable(event.key))
```
stevecotton, line 296: This line seems strange, it now seems to accept any combination of modifies, including no modifiers, except when the modifiers are exactly Shift.

renom: As far as I can understand, SDL_TEXTINPUT is handled [here](https://github.com/wesnoth/wesnoth/blob/master/src/hotkey/hotkey_item.cpp#L301) to make shifted hotkeys work?

stevecotton:
Adding comments to the code would be a huge help, both for the current review and for anyone looking at it in the future.

Is the comparison to `SDL_SCANCODE_UNKNOWN` meant to ensure that it doesn't trigger for anything that's on a Latin layout? Assuming it is, there's probably a lot of edge-cases to discover.

renom:
`SDL_GetScancodeFromName(SDL_GetKeyName(event.key.keysym.sym)) == SDL_SCANCODE_UNKNOWN`
If the input character can't be translated to a scancode, then it's an unknown symbol (probably a non-english layout letter or a layout-specific special character). A different condition could be here `if keycode_name != scancode_name`, but I didn't test how it'd work though (I think it could imply extra matches, for example for non-qwerty keyboards). That condition is used as an extra check before matching via a scancode name:
`SDL_GetKeyFromName(SDL_GetScancodeName(event.key.keysym.scancode)) == keycode_`
I don't known if there's a non-english dvorak layouts or something, where this approach probably won't work (because scancode names wouldn't match the actual latin symbols on the keys), but I guess Qt's and GTK's hotkeys work the same way as in this PR (for example english-letter hotkeys works fine in a russian layout without extra configuration).

renom:
Anyway, I think the game can't be aware of all the layouts on the PC, unless the user specifically switch between all of them (at least SDL API doesn't imply that).
Given that, I can't see a better approach how to detect which latin-letter is on the same button with the input non-latin letter.
#5598 solves that as in the most games (physical button hotkeys instead of letter-based hotkeys), but it wasn't merged because of having inconsistency of the hotkeys in non-qwerty layouts.

Pentarctagon:
Should #5598 be closed in favor of this PR?

stevecotton:
Sorry, this also seems to assume Qwerty layout, although in this case it's coming from here: https://stackoverflow.com/questions/41253317/dvorak-vs-qwerty-scancodes#45565134 . Testing on Debian Linux (with IBus' input handling) confirmed that this PR uses a Qwerty layout for the hotkeys when the keyboard has a Cyrillic layout.

With this PR, Latin hotkeys can override non-Latin hotkeys even when the non-Latin keyboard layout is selected, because it checks for matches in a single pass and takes the first hotkey that matches, regardless of whether it's using the fallback code. That breaks stuff for anyone who binds hotkeys for "first letter of a task, using a word in the language that I'm using, even if it's on a different physical key". Doing it in two passes needs extra code in `const hotkey_ptr get_hotkey(const SDL_Event &event)`, but is probably necessary.

Still thinking about how to progress here - it would definitely be good to fix #5421, but none of the PRs so far seem to hit the right spot.

@renom, are there any open-source games that already have a solution to this?

PR #4244, opened 2019-08-17
Revert network_transmission.cpp/hpp to its 1.14.7 version

stevecotton:
I re-ran the timing test discussed in PR #4201 (comment linked to below), and
saw similar numbers.  The old, simpler code downloads an 86.1MB addon in 25
seconds, the code with the extra thread takes 29 seconds.

https://github.com/wesnoth/wesnoth/pull/4201#issuecomment-517856090

PR #6778, opened 2022-06-19
Manual forward-port of #4244.

Pentarctagon:
This seems to have been applied to 1.14, but not 1.15/master at the time.

---

Tested locally by successfully downloading the largest add-on currently available on 1.17's add-ons server: Akladian Music


stevecotton:
Thanks - IIRC it wasn't forward-ported at the time because I expected @Vultraz to improve the threaded code instead. Fully agree with forward-porting it now.

CelticMinstrel:
Pinging #4244 here because it's not clickable in the title.

PR #6779, opened 2022-06-20
SDL: Increase requirement to 2.0.10

Wedge009:
@doofus-01 reported not being able to build master with SDL 2.0.8. [SDL_RenderCopyF](https://wiki.libsdl.org/SDL_RenderCopyF) and [SDL_RenderCopyExF](https://wiki.libsdl.org/SDL_RenderCopyExF) are only available from SDL 2.0.10 onwards.

These are the only instances of SDL 2.0.8 that I could find. I do not know if there are other places where references to SDL need to be updated.


soliton-:
https://github.com/hrubymar10/MacCompileStuff contains SDL 2.0.12 so that's fine as well.

PR #6781, opened 2022-06-20
1.17 boat sprites

doofus-01:
The boats & galleons are not pixel art.  I've moved the existing images to scenery for now, but they might need to go into terrain.


doofus-01:
I plan to add some standing animation, or at least some minimal movement somewhere.  The palettes will still need to be put through the #6417 process.

doofus-01:
These are not stellar, but the baseline for improvement was low.  Maybe campaign maintainers need to be consulted...
@nemaara and @Jonathan-Kelly - Do you have any thoughts on what's needed for boats?  These are non-standard units, mostly for campaigns, I believe.

knyghtmare (commented):
I like the new ships. Modernised.

One comment: changing IDs of these core units will cause deprecation issues in the following:

-  SotBE S6 and S11
- World Conquest
- The Rise of Wesnoth
- Heir to the Throne
- Under the Burning Suns
- Dead Water

and would need to be rectified in a follow-up PR

As for WoF, it has 2 custom ship units which were in-line with the old-style ship units: Caravel and Outrigger. They would need to modernized as well.

Jonathan-Kelly:
@doofus-01
Truth be told, I have been working on rendering some potential replacement sprites for these boats, working from scratch using Blender. Though I am not sure yet if I can meet this new tighter requirement for a small palette pixel art look.

I planned on making the boat sprites much larger (as you suggested in the WoF PR), big enough to work as a level 3, 4 or 5 unit. Following the lead of the TRoW story art, I thought a medieval Cog or Caravel type vessel might be better than the late era Galleons we have currently? Cogs and Caravels were only 20 meters from bow to stern, so (along with the larger sprite size) the sense of scale would be less far off next to infantry/cavalry type units.

That's for the "human" boat. Assuming all worked out well with that, I was thinking of doing similar for the orcs, giving them something with a lot of oar power, since that seems like their style? So maybe a Galley?

---

Anyway, it seems like what you have already done should be good enough for now. I believe only SotBE uses a galleon as an actual unit. while the others campaigns just use it as scenery or a `{fake_unit]` to drag around in story cutscenes. For WoF, ships do play an important role as real units in about 1/4 of its scenarios, but there is a chance I will find a way to improve these sprites myself, as described above.

doofus-01:
>  Though I am not sure yet if I can meet this new tighter requirement for a small palette pixel art look.

I'd just point out that sprites being pixel art is not some new requirement.  It's just that Wesnoth has a lot of legacy hodge-podge, so we had these weird boats.  If it really makes more sense for boats to be under some different mechanic, and there's a good chance that it does, it would need to be more to it than them just having blurry, down-scaled sprites.

That said, I don't see why down-scaled Blender model images can't be part of the workflow to end up with pixel art sprites.

> I planned on making the boat sprites much larger (as you suggested in the WoF PR), big enough to work as a level 3, 4 or 5 unit. Following the lead of the TRoW story art, I thought a medieval Cog or Caravel type vessel might be better than the late era Galleons we have currently? Cogs and Caravels were only 20 meters from bow to stern, so (along with the larger sprite size) the sense of scale would be less far off next to infantry/cavalry type units.

Yeah, I agree the current "Galleons" cause more dissonance than they really need to.  Cogs and Caravels sound more sensible. 

> That's for the "human" boat. Assuming all worked out well with that, I was thinking of doing similar for the orcs, giving them something with a lot of oar power, since that seems like their style? So maybe a Galley?

Makes sense to me.

spixi:
I see no reason to change the IDs. The ship units are used in many mainline and user-made campaigns. For compatibility reasons I propose to keep the old IDs.

data/core/units/boats/Pirate_Galleon.cfg
```diff
@@ -36,3 +36,12 @@
...
 [/unit_type]
+
+# trying to make more distinctions between units
+[unit_type]
+    id=Pirate Galleon
+    [base_unit]
+        id=Pirate Carrack
+    [/base_unit]
+    description= _ "Pirate galleons are deprecated, use pirate carracks instead."
+[/unit_type]
```
CelticMinstrel, line 47: This should use the DEPRECATED_UNIT macro.

data/core/units/boats/Transport_Galleon.cfg
```diff
@@ -36,3 +36,13 @@
...
+
+# renaming because "transport" was ambiguous; the distinction between Galleon and Transport Galleon was not intuitive. 
+# this is here for deprecation
+[unit_type]
+    id=Transport Galleon
+    [base_unit]
+        id=Assault Galleon
+    [/base_unit]
+        description= _ "Transport Galleons are deprecated, use Assault Galleons instead"
+[/unit_type]
```
CelticMinstrel, line 95: This too should use the DEPRECATED_UNIT macro

Jonathan-Kelly:
My apologies for taking so very long to get back to this. I finally have something which might be worth looking at:

cog

What do you think of it so far?

doofus-01:
With this and #9367, we're getting quite a collection of boat sprites!  

For that particular sprite, the overall design is interesting.  Is that a weapon on the bowsprit?  The sprite does need some darker tones and shadows, but maybe it is WIP and you will get to that.  The mast may be a bit too tall compared to the others, unless this is supposed to be some sort of special flagship.

A problem that all of these have is that they really need animations to not look weird in the animated, sloshing water.  Some of it can be faked with `y=` key bobbing, but some exaggerated recoil or roll frames would really make a difference.

Jonathan-Kelly:
> For that particular sprite, the overall design is interesting. Is that a weapon on the bowsprit?

That's a medieval style forecastle, which back then they often attached to the bowsprit. There wasn't enough room for the ballista on the forecastle so I moved it to the aftcastle.

> The mast may be a bit too tall compared to the others, unless this is supposed to be some sort of special flagship.

It follows the idea outlined back [here](https://github.com/wesnoth/wesnoth/pull/6781#issuecomment-1166839830) to replace all uses of "galleons" in mainline campaigns (most all of which are just images or fake units) with the medieval cog. But there are a few instances it would not replace:

* SotA's galleon - since its deck terrain tiles only work for a multi masted galleon and not a cog
* SotBE and WoF orcish vessels - to be replaced with an oared [galley](https://en.wikipedia.org/wiki/Galley) made using this same technique

The large size of the sprite combined with the smaller size of the medieval cog (about 20 meters stem to stern) work together to reduce the scale dissonance between ships/boats and all the other unit sprites in the game, as well as fit the older and smaller style of ships/boats depicted in TRoW's story images. Also its size is meant to fit a level 4 or 5 unit in terms of unit statistics since we have so few "boss" units for scenarios to use.

> A problem that all of these have is that they really need animations to not look weird in the animated, sloshing water.

That should all be very easy to provide, and the animations will be in all 4 facing directions (N, NE, SE, S). It's true that 3D takes 3 times longer, but at least when it comes to animations and (especially) viewing angles it pays you back somewhat.

> The sprite does need some darker tones and shadows, but maybe it is WIP and you will get to that.

Oops, I actually disabled shadows because I thought the Wesnoth sprite style did not allow too much of them. Here it is with shadows and less ambient light:

cog1

Does this look better? Just right or maybe too dark?

CelticMinstrel:
It appears to be missing submerge.

doofus-01:
> That should all be very easy to provide, and the animations will be in all 4 facing directions (N, NE, SE, S). It's true that 3D takes 3 times longer, but at least when it comes to animations and (especially) viewing angles it pays you back somewhat.

That makes sense, and does explain some misunderstanding with shadows.  Each frame will need touch-up; pixel art does require deliberate placement of each pixel, but an animation frame doesn't get as much scrutiny to detail and touch-up is much easier than drawing from scratch, so the 3D model is still a good method.  As far as shadows, what I meant was local shadow & shading on the forecastle and even the deck, not so much the big shadow from the sail.

Jonathan-Kelly:
Okay, here it is with more localized shading, removal of the sail shadow, another color palette and the submerge effect:

cog2

Hmm... thoughts?  How bright would you like me to make the sides of the hull relative to the deck?

> Each frame will need touch-up; pixel art does require deliberate placement of each pixel, but an animation frame doesn't get as much scrutiny to detail and touch-up is much easier than drawing from scratch, so the 3D model is still a good method.

In the best case scenario, I think at least all the team colored pixels will need to be placed manually.

Speaking of the animation, what frame rate and animation length do you think I should shoot for with the "standing" animation (the boat gentle rolling over the waves)?

Also, how big can/should this sprite be? Is the current fire dragon a good example of the maximum size of a (wesnoth) sprite or it is too big and will get shrunk down in its next iteration?

doofus-01:
I've given it some time but am still having a little trouble reading the forecastle, and I can't find a reference that looks like that.  But I don't know much about medieval ships.  Do you have a reference?

Frame rate depends on a lot of things (number of frames, amount of movement, sprite size, image details).  50ms per frame is surely too fast, 500ms is likely too slow, probably each frame should be 80-200ms.  Don't get too hung up on it yet, it's easy to fiddle with later.  The total length can be 

The fire dragon shouldn't be used as an example of anything; it's a good image, but it's not a great sprite and it puts the engine through "stress-tests" like layering issues and redraw artifacts.  The current height should be fine for a galleon, roughly (maybe a little too tall?).  If it seems weird having a bunch of dinghy-scale "ships", realize that these are still big sprites and scale is loose because sprites are representational.  Players already have to accept the outhouse-size villages/buildings.

Jonathan-Kelly:
Oh yeah, you can find a number of decent references on google images by searching key phrases like "cog ship", "medieval cog", "hanseatic cog", "kogge", "hanse kogge" "roundship" and such. Search engines otherwise get confused thinking you want images of gears.

[Reference 1](https://www.dhm.de/mediathek/en/web/ship-types/milestones-in-the-history-of-european-shipbuilding/04-cog/)
[Reference 2](http://www.tallship-fan.de/cgi-bin/tallship_e.pl?ACTION=DISPLAY&SCHIFFSID=1945)

kogge

Basically, the forecastle (and aftcastle) is a wooden platform surrounded by a wooden defensive wall complete with wooden [battlements](https://en.wikipedia.org/wiki/Battlement) like the stone battlements you see on top of castle turrets and walls. Archers and crossbowmen would use the battlements for cover, until they were all phased out as cannons and other firearms became common in warfare.

doofus-01:
Thanks, I see what it is now because of the discussion, but I'd never have been able to read it that way otherwise.  I'd make it shorter or broader or more wedge-shaped - something so that it is better integrated with the rest of the ship.  It should also help legibility if the interior and exterior walls were better differentiated. 

The bottom of the sail and the front edge of the sterncastle are also forming a tangent that might be obscuring whatever continuity there is between the upper decks.

Jonathan-Kelly:
Good points. I reshaped the sail and changed the view orientation slightly to make the sail and sterncastle more distinguishable from each other, and then widened the forecastle and darkened its interior walls further so the exterior walls stand out more:

cog3

Does it look better?

I guess the nuclear option would be to team color the exterior castle walls, but then they might blend into the water tiles whenever the team color is blue or cyan.

doofus-01:
The tangent has been fixed, thanks.  The overall proportions seem a bit off, both in reference to #9367 and this/#9484, but also the proportions in the N vs. S facing bow and forecastle - the forecastle is smaller in the N-facing sprite.  Depending on the intention relative to other ships, it could just be a shading issue that is saved for later.  Maybe opening a PR with the unit stats & all would be best?

Jonathan-Kelly:
No problem, I will open a separate pull request for this. Feel free to close this one as soon as you wish.

> the proportions in the N vs. S facing bow and forecastle - the forecastle is smaller in the N-facing sprite

Okay, will use orthographic projection from now on (I probably should have used that from the beginning):

cog5

The sail still looks like it changes size because it is mostly edge-on in the NE / NW facing. It will need some more reshaping.

Zireael07:
Why not assume they're sailing with the wind from the rear (downwind), which would mean the sail is pretty much 90 deg to the ship and therefore would look the same no matter the view angle?

doofus-01:
> The sail still looks like it changes size because it is mostly edge-on in the NE / NW facing. It will need some more reshaping.

I think making the mast and yard more visible (2px wide?) would help, even if it isn't proportionally accurate.  Remember that the sprites are icon representations, not shrunken simulations.  Stronger shadows would also help.

Jonathan-Kelly:
@Zireael07 sorry, do you mean that regardless of whichever direction the ship is facing, the wind is always coming from directly behind it? If so, that is currently how the sprite's sail is being handled. Or do you mean something else like [tacking](https://en.wikipedia.org/wiki/Tacking_(sailing)) the sail so that as the ship changes its facing direction the sail pivots to catch the wind better?

> I think making the mast and yard more visible (2px wide?) would help, even if it isn't proportionally accurate. Remember that the sprites are icon representations, not shrunken simulations.

Wow, that helped quite a lot (thickness of mast doubled and yard tripled):

cog6

I see what you mean about iconic rather than realistic proportions. I'll start off with a lot more exaggeration in future sprites (like the galley).

> Stronger shadows would also help.

On the sail itself? Or do you mean that the deck and inside lip of the hull needs to be darker in the ne/nw facing frame?

Zireael07:
Yeah, I was saying that the ship should be represented as though the wind is always from the rear

Jonathan-Kelly:
Thanks for clarifying, that makes sense.

soliton-:
Is this PR still useful?

Jonathan-Kelly:
I believe it is obsolete at this point.

doofus-01:
Yes, PR is obsolete.  There was some interesting discussion.  Too bad it is here and not preserved in the forums or somewhere under the control of Wesnoth.

Issue #4117, opened 2019-06-12
Duo Leader with experimental AI


sevu:
Classical AI is splitting it's resources among both leaders.
Experimental one spends all gold on one leader.

Not sure to treat this as a bug. I'm not aware of anybody using duo leaders with separate castles. In some way the behaviour could also be advantageous, by having a backup castle + leader for a side.

gfgtdf:
> I'm not aware of anybody using duo leaders

Iirc low does this, but it was just a workaround for 1.12 not supporting more than 9 sides in mp which was fixed.

Konrad22:
There are some SP campaigns add-ons which use duo (or trio) leaders with separate castles, e.g. Captured by a Nightmare.

mschoeck-zz:
I'm not sure whether this should be considered a bug either (but I think that is independent of whether it is actually used anywhere).  The AI is specifically set up to recruit using only one leader, so in that sense this is a feature.  However, which leader is chosen is (semi)random, so that could be considered a bug.

In any case, extending the ExpAI to be able to deal with multiple leaders is a perfectly reasonable feature request.  I'll have to have a look at the code to see how this could be done in a way consistent with the evaluations of this evaluations used.  I likely won't get around to that immediately though, so if somebody else wants to beat me to it, please do so.

cooljeanius:
> There are some SP campaigns add-ons which use duo (or trio) leaders with separate castles, e.g. Captured by a Nightmare.

I'm porting Dawn of Thunder to 1.16 and actually have a scenario in it where one side has 6 separate leaders (3 of which are expendable) with separate castles (at least on the easiest difficulty)

cooljeanius:
Re: "Is this a bug or enhancement?", how about splitting the difference and calling it an enhancement via labels, but a bug via issue type?

PR #6783, opened 2022-06-21
Some fixes to the Experimental AI

### LINKED ISSUE #4924: experimental_ai doesn't recruit from extra_recruit
1.14.12
Compiled from source tag 1.14.12
Debian buster

### Describe the bug
When `ai_algorithm=experimental_ai` is enabled for a side, it won't recruit from types given under `[unit] extra_recruit=`. It only recruits from types set with `[side] recruit=`

Details are in this thread.
https://forums.wesnoth.org/viewtopic.php?f=21&t=52539

mattsc:
The incentive for this change was the fact that there is a "data leakage" bug if several sides all use the Experimental AI recruiting. Because AI candidate actions are loaded with `wesnoth.require`, all the information set in the old `ca_recruit_rushers.lua` is only loaded once and thus applied to all sides. While there are other ways of fixing this problem, I've been meaning to remove the generic recruit engine in favor of simple straight CAs for a while anyway.

There is no change in behavior due to the first (refactoring) commit. The CA functions exactly as before, and `generic_recruit_engine.lua` is still there, in case some UMC uses it directly.

There are, however, a couple internal parameter changes. `leader_takes_village` and `min_turn_1_recruit` were internal parameters for the interaction between the Experimental AI's castle switch and recruit rushers CAs. Now that both CAs have full access to the AI engine's `data` variable, they are not needed any more.

The one actual change is that `score_function` has now been replaced by the usual `ca_score` parameter. Again, this is entirely internal and there is no change of behavior. In the unlikely case that some UMC actually uses the generic recruit engine with a score function, the engine is still there, as already mentioned. And for the ExpAI itself, the behavior is exactly the same, as the score function just returned a constant value anyway.

While I'm at it, I might as well also do the following as part of this PR (but I first want to see if there are any substantial objections to what I have so far):

- [x] Deprecate the generic recruit engine
- [x] When the castle switch CA is not present, l.928 of the new `ca_recruit_rushers.lua` appears to go through the `__cfg` of the AI configuration on each call, which is very inefficient. That should be avoided.
- [x] Fix #4924.  (By contrast, #4117 is more complex than I want to deal with at this time.)


mattsc:
@Alarantalara I don't know how much you are around these days, but I saw that you posted some comments recently.  If you have time, you may want to have a look too, since a lot of this is your code (although I may have introduced the problem with the data leakage, I don't remember).

data/ai/lua/ca_recruit_rushers.lua
```diff
@@ -1,50 +1,1096 @@
...
+end
+
+
+local function drainable(unit)
+    return not unit.status.undrainable
+end
+
+
+local function get_best_defense(unit)
+    local terrain_archetypes = { "Wo", "Ww", "Wwr", "Ss", "Gt", "Ds", "Ft", "Hh", "Mm", "Vi", "Ch", "Uu", "At", "Qt", "^Uf", "Xt", "Tb" }
```
Alarantalara, line 93: Might as well update the list of terrain codes while you're copying everything.
^Uf and Tb are the same terrain but ^Uf is deprecated, so it can go. I suggest using Tt, which is what both are aliases of.
Mine rails got a special code and while it's not useful in general, there might be some value in adding it (I left it out since it's more of a functional change in the suggestion, but worth considering).
```suggestion
    local terrain_archetypes = { "Wo", "Ww", "Wwr", "Ss", "Gt", "Ds", "Ft", "Hh", "Mm", "Vi", "Ch", "Uu", "At", "Qt", "Xt", "Tt" }
```

Alarantalara:
I never stopped getting mail and even read it sometimes, but hadn't been paying much attention for awhile. I started getting more involved again about a month ago, but mostly it's just been me going over all the things that changed and realizing how out of date I am. I'm having fun poking around the new features and making my unfinished ideas work again, so it's possible I might eventually do something a bit more substantial.

data/ai/lua/generic_recruit_engine.lua
```diff
@@ -21,6 +21,8 @@ return {
     --   finding the best unit types to recruit. It does not work otherwise. It assumes normal vision of the AI side (that is, it disregards
     --   hidden enemy units) for determining from which keep hex the leader should recruit and on which castle hexes to recruit new units
     init = function(ai_cas, params)
+        wesnoth.deprecated_message('ai/lua/generic_recruit_engine.lua', 3, '1.19', 'All its functionality has been moved to ai/lua/ca_recruit_rushers.lua which can now be used directly.')
```
CelticMinstrel, line 24: I think I'd just move this to the top of the file, above `return`.

mattsc: Changed and force-pushed.

mattsc:
This is done now, as far as I am concerned.  I implemented the suggestions above, added the last task (extra_recruits) and squashed commits.

I'll let this sit here for a couple days to see if there are other comments and merge it if not.

data/ai/lua/ca_recruit_rushers.lua
```diff
@@ -944,6 +944,11 @@ function ca_recruit_rushers:evaluation(cfg, data, filter_own)
         end
     end
 
+    recruit_data.recuit_types = wesnoth.sides[wesnoth.current.side].recruit
+    for _,unit_type in ipairs(leader.extra_recruit) do
+        table.insert(recruit_data.recuit_types, unit_type)
```
Alarantalara, line 949: This might cause duplicate entries if both wesnoth.sides[wesnoth.current.side].recruit and leader.extra_recruit contain the same unit.
Apart from slowing things down, depending on the level of randomness, it could also increase the chance that unit is selected since it gets two chances to have the highest score, not one.

data/ai/lua/ca_recruit_rushers.lua
```diff
@@ -944,6 +944,14 @@ function ca_recruit_rushers:evaluation(cfg, data, filter_own)
         end
     end
 
+    recruit_data.recuit_types = {}
+    for _,unit_type in ipairs(wesnoth.sides[wesnoth.current.side].recruit) do
+        recruit_data.recuit_types[unit_type] = true
```
CelticMinstrel, line 949: Missing the R in "recruit_types" here.

mattsc: Funny in that I had just noticed that myself.  And there was a more substantial problem also: the code setting up that table actually had to be pulled forward about 20 lines as it was sometimes needed earlier (a situation that had not happened in my previous tests).  Both is fixed now.

mattsc:
By the way, I should have mentioned earlier that deprecating the generic recruit engine and the new parameter `reset_cache_each_turn` are for master only, but the other four are bug fixes or efficiency improvements and should be backported to 1.16.

data/ai/lua/ca_recruit_rushers.lua
```diff
@@ -1,50 +1,1091 @@
...
+-- cfg: parameters for configuring recruitment, read from the candidate action's [args] tag
+--    ca_score: score of the candidate action
+--      (default: 196000 as defined in AI_CA_RECRUIT_RUSHERS_SCORE)
+--    randomness: a measure of randomness in recruitment
+--      higher absolute values increase randomness, with values above about 3 being close to completely random
+--      (default = 0.1)
+--    high_level_fraction: the approximate fraction of units of level 2 or higher to be recruited
+--      (default = 0)
+--     enemy_types: array of default enemy unit types to consider if there are no enemies on the map
+--      and no enemy sides exist or have recruit lists
```
mattsc, line 10: I just noticed that specifying `enemy_types` as a Lua array does not work (and has never worked) if this parameter is to be supplied by an `[args]` tag. We could convert it to a comma-separated list instead, but I suggest to simply drop this parameter altogether. It is an undocumented parameter and, as is obvious from the comment above, it is essentially never needed, and even in the rare scenario when that might happen, there is a default list of types provided that will be used instead. This is such a rare edge case that it does not seem worth it to me to keep this parameter.

mattsc:
Update: I got rid of the `enemy_types` config parameter, as discussed above. I squashed this into the first commit, which already modifies some of the other internal parameters.

data/ai/lua/ca_recruit_rushers.lua
Alarantalara: As I recall, I added it with maps like Dark Forecast in mind, where there are lots of enemies of known types, but they're all placed via scripting. Humans can play it fine after the first few games, but the AI has no way of knowing what's coming and this would help with the first round of enemies.

Of course, I never actually tried making an AI for Dark Forecast, so I never discovered the issue. I suspect I would have ended up either providing the parameter through Lua and not noticed anyway or modifiing it further to support changing spawn lists.

At any rate, unless you plan on writing a Dark Forecast AI, it seems completely reasonable to drop it.

mattsc: Thanks for the comments.  I am not planning to write an AI specifically for Dark Forecast either, so I am going to stick with dropping this for now.  We can always add something (which would likely end up being more complex than this) later, either in this CA or a custom setup, if the need arises.

PR #6785, opened 2022-06-21
Add event unit tests for dynamic filters and names

CelticMinstrel:
Mostly for CI, though I guess I need to add comments too.


Pentarctagon (approved):

stevecotton:
The behavior tested by `event_handlers_in_events__*__dynamic_filter` could be implemented without `[insert_tag]`, by using`[filter]race=$filtered_races`. I'm wondering if choosing this behavior prevents a use-case with WML constructing a more complex filter with something unit-specific during a `first_time_only=no` event, `[insert_tag]`'ing it, and then overwriting the temporary variable when that `first_time_only=no` event is triggered again by a different unit.

It seems that even Ageless Era and Era of High Sorcery don't have any events using `[insert_tag]` in an event filter, so nothing is likely to break either way.

Clarifying the Wiki docs would also be good - whichever of us gets to it first.

CelticMinstrel:
> I'm wondering if choosing this behavior prevents a use-case with WML constructing a more complex filter with something unit-specific during a `first_time_only=no` event, `[insert_tag]`'ing it, and then overwriting the temporary variable when that `first_time_only=no` event is triggered again by a different unit.

I have no idea what you're trying to say here.

Issue #6544, opened 2022-02-26
End turn button partly overflowing hidden


galegosimpatico:
<img width="803" alt="Screen Shot 2022-02-26 at 23 42 43" src="https://user-images.githubusercontent.com/53513496/155861381-a6092d6e-fa10-4039-aba6-6a765b4fe36e.png">

800x600

Wedge009:
I've seen this before but I'm not sure if it's been formally reported here previously.

galegosimpatico:
When the scenario ends it turns to a more grave "End scena".

stevecotton:
This is one of the issues in #5620; it's probably helpful to keep this as a dedicated issue for this part of that issue.

Wedge009:
At 1024x768 the End Turn/Scenario button is positioned correctly, and is given the full decoration (https://github.com/wesnoth/wesnoth/blob/master/images/buttons/large-button.png).
https://github.com/wesnoth/wesnoth/blob/70d957575881a0394835b64f50fe77b16185d252/data/themes/default.cfg#L132-L142

When the vertical resolution goes below 768 the button is reduced to the 'normal' version (https://github.com/wesnoth/wesnoth/blob/master/images/buttons/button_normal/button_H22.png) but otherwise appears to be positioned correctly.
https://github.com/wesnoth/wesnoth/blob/70d957575881a0394835b64f50fe77b16185d252/data/themes/default.cfg#L605-L609

However, when the horizontal resolution goes below 1024 the button shifts to the right as shown in the original report. I don't know enough about the GUI handling to see how/why this happens.

Wedge009:
Confirmed resolved in master, thank you.

galegosimpatico:
Nice! Thanks.

Could anyone ping again when this is deployed to Steam?

Wedge009:
I believe it will be included in the next 'beta' release, 1.17.5 I think.

mesilliac:
@galegosimpatico 1.17.5 should be up now, including this fix

Wedge009:
I don't believe there was a commit or change to specifically address this issue so I doubt there will be a back-port - I believe this was resolved as part of mesilliac's (great) work on overhauling the rendering system for 1.17. Since it's such a major change, it belongs in the development builds.

galegosimpatico:
I understand now, this patch could be written only because it's on top of other changes by mesilliac, those other changes can't be backported.

mesilliac:
[...]

Issue #5620, opened 2021-03-17
Narrow game width breaks side menu


Violetnred2:
happens on 1.15.10, but not on 1.14.15. 

happens if a window is narrower than 800 pixels or so

side menu in map editor get buttons that are misaligned, "End Scenario" button after a game ends is misaligned as well and loses it's decoration.

![wesnoth_0rAncreZqG](https://user-images.githubusercontent.com/26041224/111482644-c3efeb00-873c-11eb-94cc-ffda3a5096b5.png)
![wesnoth_vU5fFltxYu](https://user-images.githubusercontent.com/26041224/111482649-c5211800-873c-11eb-8ba9-ac0a8c9a5f56.png)

shikadiqueen:
For the record, it's intentional that on smaller sizes the End Turn button looks like a normal button instead of being the big decorated button you see at 1024x768 and larger. (Everything else is bugs, of course.)

Pentarctagon:
The minimum requirements for the window size on the website homepage are listed as 800x600, so are windows smaller than 800 pixels meant to be supported?

shikadiqueen:
No, but you can see in the screenshot that in reality the horizontal resolution is higher than 800 (and smaller than 1024).

Violetnred2:
ok, i looked at my window width this time - it is 960. 

the exact width below which the side panel break is 1032.

shikadiqueen:
Just found out this also affects the regular game theme **if** and only if the replay UI is active.

![image](https://user-images.githubusercontent.com/489895/111710979-8e74fb80-8829-11eb-86a9-11ec7530daa2.png)

shikadiqueen:
Reverting the following three commits in order shows it's another side-effect of #4851.

* 2e6a043c1a40854a08704d3614502b5b07c84fd8
* 399fd8adf96ab2dbd40f2ac4fd5cc273e002e015
* 43de778c6aad8710707680a6b59ba0adbabbe565

shikadiqueen:
For reference (I don't think I can look into this right now) the issue with the replay UI seems to stem from the aforementioned changes and related not taking into account the optional existence of the replay UI at all.

Not sure about the editor and the turn button.

Wedge009:
Had a look at the End Turn button specifically. Don't think I found anything new that hasn't already been mentioned here.
https://github.com/wesnoth/wesnoth/issues/6544#issuecomment-1053715284

mesilliac:
I fixed these all.

[...]

PR #6790, opened 2022-06-22
Back-port resolution to #6544

Wedge009:


mesilliac:
I just tested briefly, this seems to be working as expected for editor, replays, and end turn button.

I also forgot to add changelog entries, it could use one for #6544 and also for #5620

PR #6791, opened 2022-06-22
(1.17/1.18)remove deprecated backstab option

newfrenchy83:
Backstab attribute option deprecated, we could remove that, isn't it?

### DIFF
--- a/data/schema/units/specials.cfg
+++ b/data/schema/units/specials.cfg
@@ -52,7 +52,6 @@
 	{SIMPLE_KEY multiply f_real}
 	{SIMPLE_KEY divide f_real}
 	{SIMPLE_KEY cumulative bool}
-	{DEPRECATED_KEY backstab bool}
 	{FILTER_TAG "filter_base_value" base_value ()}
 [/tag]
 [tag]
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -82,8 +82,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 		bool attacking,
 		nonempty_unit_const_ptr oppp,
 		const map_location& opp_loc,
-		const_attack_ptr opp_weapon,
-		const unit_map& units)
+		const_attack_ptr opp_weapon)
 	: weapon(nullptr)
 	, attack_num(u_attack_num)
 	, is_attacker(attacking)
@@ -94,7 +93,6 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	, petrifies(false)
 	, plagues(false)
 	, poisons(false)
-	, backstab_pos(false)
 	, swarm(false)
 	, firststrike(false)
 	, disable(false)
@@ -149,7 +147,6 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	drains = !opp.get_state("undrainable") && weapon->has_special_or_ability("drains");
 	petrifies = !opp.get_state("unpetrifiable") && weapon->has_special_or_ability("petrifies");
 	poisons = !opp.get_state("unpoisonable") && weapon->has_special_or_ability("poison") && !opp.get_state(unit::STATE_POISONED);
-	backstab_pos = is_attacker && backstab_check(u_loc, opp_loc, units, resources::gameboard->teams());
 	rounds = weapon->get_specials_and_abilities("berserk").highest("value", 1).first;
 
 	firststrike = weapon->has_special_or_ability("firststrike");
@@ -180,7 +177,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	cth = std::clamp(cth, 0, 100);
 
 	unit_ability_list cth_specials = weapon->get_specials_and_abilities("chance_to_hit");
-	unit_abilities::effect cth_effects(cth_specials, cth, backstab_pos, weapon);
+	unit_abilities::effect cth_effects(cth_specials, cth, false, weapon);
 	cth = cth_effects.get_composite_value();
 
 
@@ -191,7 +188,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	chance_to_hit = std::clamp(cth, 0, 100);
 
 	// Compute base damage done with the weapon.
-	int base_damage = weapon->modified_damage(backstab_pos);
+	int base_damage = weapon->modified_damage();
 
 	// Get the damage multiplier applied to the base damage of the weapon.
 	int damage_multiplier = 100;
@@ -221,19 +218,19 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	if(drains) {
 		unit_ability_list drain_specials = weapon->get_specials_and_abilities("drains");
 		// Compute the drain percent (with 50% as the base for backward compatibility)
-		unit_abilities::effect drain_percent_effects(drain_specials, 50, backstab_pos, weapon);
+		unit_abilities::effect drain_percent_effects(drain_specials, 50, false, weapon);
 		drain_percent = drain_percent_effects.get_composite_value();
 	}
 
 	// Add heal_on_hit (the drain constant)
 	unit_ability_list heal_on_hit_specials = weapon->get_specials_and_abilities("heal_on_hit");
-	unit_abilities::effect heal_on_hit_effects(heal_on_hit_specials, 0, backstab_pos, weapon);
+	unit_abilities::effect heal_on_hit_effects(heal_on_hit_specials, 0, false, weapon);
 	drain_constant += heal_on_hit_effects.get_composite_value();
 
 	drains = drain_constant || drain_percent;
 
 	// Compute the number of blows and handle swarm.
-	weapon->modified_attacks(backstab_pos, swarm_min, swarm_max);
+	weapon->modified_attacks(swarm_min, swarm_max);
 	swarm = swarm_min != swarm_max;
 	num_blows = calc_blows(hp);
 }
@@ -255,7 +252,6 @@ battle_context_unit_stats::battle_context_unit_stats(const unit_type* u_type,
 	, petrifies(false)
 	, plagues(false)
 	, poisons(false)
-	, backstab_pos(false)
 	, swarm(false)
 	, firststrike(false)
 	, disable(false)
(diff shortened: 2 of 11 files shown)


soliton-:
Is there a specific reason to do this?

Pentarctagon:
If we say something is deprecated and then state the version it will be removed in, then we should probably remove it. Otherwise there isn't really a point to providing the version.

soliton-:
> Deprecated preemptively — This deprecation level is the one which should be aimed for the vast majority of the time. Obsolete APIs which can be implemented in terms of a simple minimal-maintenance wrapper should be deprecated, but left in the code indefinitely. At such time as an obsolete API presents itself as an active obstacle toward further development, it can be slated for removal in the next release, or even removed at will immediately, so long as it has already survived a minimum deprecation period of at least one stable release version.

Maybe the idea is that the deprecation level was wrong?

newfrenchy83:
Must i repull DEP_LEVEL::PREEMPTIVE in code?

soliton-:
> Must i repull DEP_LEVEL::PREEMPTIVE in code?

I can't tell what that means.

Can you explain why you're doing this? Are you aware of the [deprecation policy](https://wiki.wesnoth.org/CompatibilityStandards)?

CelticMinstrel:
_Was_ it deprecated in 1.15? Not in 1.13?

Pentarctagon:
Looks like it was deprecated in 1.13.5 over 6 years ago: https://github.com/wesnoth/wesnoth/commit/c761a80bd53a9537e09a7008ad9c17031e7604ca#diff-468254ac2424b59830c0606b5ab37eb3cd1863fa0394b505c7484c4e9945622f

mesilliac:
Apologies for the conflict from the indentation commit i just pushed - that was formatting only, so whatever this PR was doing should probably take precedence.

Pentarctagon:
So for the record, I'm fine with this being removed. If something is deprecated with a version it will be removed on, then that's the version it should be, well, *removed*, not the version where we all again start debating whether it should actually be removed or not. I'll leave this open for a few more days, but otherwise I don't see any reason to let this sit here open forever like it seems a lot of PRs have ended up doing.

CelticMinstrel:
> If something is deprecated with a version it will be removed on, then that's the version it should be, well, _removed_, not the version where we all again start debating whether it should actually be removed or not.

I have a slightly different take. Removing something takes effort. The version given is the minimum version at which something _can_ be removed, but there's no requirement that we _do_ remove everything marked for removal in a given version before releasing the version. So… it's not the version where it absolutely must be removed, but it's also not the version where we start debating over whether to actually remove it. There shouldn't be any debate over removing deprecated stuff in the version it was marked to be removed in; if someone chose to put in the work to do the removal, I don't see any reason not to accept it.

soliton-:
> So for the record, I'm fine with this being removed. If something is deprecated with a version it will be removed on, then that's the version it should be, well, _removed_, not the version where we all again start debating whether it should actually be removed or not.

The deprecation policy clearly says:

> Obsolete APIs which can be implemented in terms of a simple minimal-maintenance wrapper should be deprecated, but left in the code indefinitely. At such time as an obsolete API presents itself as an active obstacle toward further development, it can be slated for removal

There seems to be either a misunderstanding or disagreement with the deprecation policy or the deprecation level that was assigned. Or this actually is an obstacle to further development that hasn't been mentioned yet.

It's certainly not true that this was deprecated with a version it will be removed on. That is not how the `PREEMPTIVE` level works but the `FOR_REMOVAL` level.

> I'll leave this open for a few more days, but otherwise I don't see any reason to let this sit here open forever like it seems a lot of PRs have ended up doing.

Certainly it can just be closed unless some kind of reason is given.

CelticMinstrel:
> Obsolete APIs which can be implemented in terms of a simple minimal-maintenance wrapper

Hmm, well, this description doesn't fit this API…

And I could definitely see an argument that it presents an obstacle to further development, since it's a special case in the ability system that everything needs to continue to account for.

Pentarctagon:
> Certainly it can just be closed unless some kind of reason is given.

I'd say that CelticMinstrel's most recent post is the reason.

---

More generally though, I'm starting to question why the preemptive deprecation level exists at all much less why it's "the one which should be aimed for the vast majority of the time". The other three deprecation levels give specific information to both UMC authors and developers on what to expect:
* Indefinite - this will continue to exist forever.
* For Removal - this will be removed in version 1.XX.
* Removed - this is already gone.

Then there's this other one:
* Preemptive - same as Indefinite, except actually this may be removed on or after `<some arbitrary version>` for `<some arbitrary reason>`.

I feel like this is not particularly useful to anyone - it doesn't tell UMC authors when they need to update their code by and it doesn't tell developers when the API should be removed. It's basically just a shrug and saying "It'll be removed at some point. Maybe."

CelticMinstrel:
> * Indefinite - this will continue to exist forever.

Not quite… more like "this will continue to exist as long as it doesn't require maintenance".

> it doesn't tell UMC authors when they need to update their code by

Uhh, yes it does? They need to update their code by the specified version, other wise it _may_ stop working.

> it doesn't tell developers when the API should be removed

Okay, sure, it doesn't give developers an explicit time when it should be removed, sure. The idea is that it can be removed for almost any reason and without notification after the specified version (although removing it just for the sake of removing things should be discouraged).

soliton-:
It seems `backstab=` could simply be replaced with https://github.com/wesnoth/wesnoth/blob/d8b8c5ea8024bf51959a6520cbb5bb6fd3a1c9f7/data/core/macros/abilities.cfg#L415-L421

If that is easy to do in code then I think that would be worthwhile to actually make it "a simple minimal-maintenance wrapper".

Also seems like the deprecation notice is not great. Especially if it's clear what the exactly equivalent new WML should look like it should just mention it explicitely. No vague "Use [filter_adjacent] instead." when that is 1) not the actual tag to use but a subtag (that isn't even used anymore in the mainline implementation) and 2) it's clear what the replacement should be in its entirety.

While we don't have to bend over backwards to keep everything compatible one of wesnoth greatest assets by now is all the user made content of which not everything is kept up-to-date but occasionally old addons do get ported to new versions and IMO it is worthwhile to make that as easy as possible if we don't have to sacrifice code quality. It's a good idea to simplify internal interfaces as done here but if it's cleanly possible we should keep the exernal functionality.

Pentarctagon:
What would a simple minimal-maintenance wrapper look like here?

Pentarctagon:
Would it be to insert the [filter_opponent] formula there?

soliton-:
Yes, the idea is to put the `[filter_opponent]` WML where the `backstab` attribute is. Haven't checked how easy it is to do in code. If it's not easy then alternatively it could also be done by wmllint.

newfrenchy83:
> @newfrenchy83 Is soliton's suggestion something you'd be able to do?

i don't tkink, ,no, and i don't see interest to modify boolean backstab for put in the wml. For me formula in [filter_opponent] make the work and other possibility can be encoded. backstab attribute is deprecated and easy to replace in wml code. I propose to remove that by consequentbut else, the message is suficient.

soliton-:
If it's easy to replace in WML why can't the engine do it? Why doesn't the deprecation message say explicitly what to replace it with? It seems we're making it harder on addon authors than it needs to be and that means there is a good chance to lose content because of it.

newfrenchy83:
> If it's easy to replace in WML why can't the engine do it? Why doesn't the deprecation message say explicitly what to replace it with? It seems we're making it harder on addon authors than it needs to be and that means there is a good chance to lose content because of it.

Ah, the depreciation message? Sorry, I misunderstood and I beg your pardon for the harshness of my answer, but this PR had slipped my mind and I had imagined that I was being asked to implement the formula in filter_opponent in the 'backstab attribute. I will close this PR and open another one with the new post.

newfrenchy83:
> > If it's easy to replace in WML why can't the engine do it? Why doesn't the deprecation message say explicitly what to replace it with? It seems we're making it harder on addon authors than it needs to be and that means there is a good chance to lose content because of it.
> 
> Ah, the depreciation message? Sorry, I misunderstood and I beg your pardon for the harshness of my answer, but this PR had slipped my mind and I had imagined that I was being asked to implement the formula in filter_opponent in the 'backstab attribute. I will close this PR and open another one with the new post.

message changed in https://github.com/wesnoth/wesnoth/pull/7021 pr

stevecotton:
I merged #7024 (the PR that changes it from `PREEMPTIVE` to `FOR_REMOVAL`) thinking that there was support from @CelticMinstrel for that change, and now I'm rereading this I just imagined that. Should I revert it?

newfrenchy83:
> I merged #7024 (the PR that changes it from `PREEMPTIVE` to `FOR_REMOVAL`) thinking that there was support from @CelticMinstrel for that change, and now I'm rereading this I just imagined that. Should I revert it?

if @CelticMinstrel don't oppose to that, then you can let like this.

CelticMinstrel:
> And I could definitely see an argument that it presents an obstacle to further development, since it's a special case in the ability system that everything needs to continue to account for.

This quote from my earlier post reads like token support to present me. That is to say, I'm not strongly in favour of it but not against it either.

Removing all the backstab logic but keeping the deprecation message along with code that substitutes the formula Soliton noted would also be a reasonable alternative. In that case, you'd probably push the deprecation level in the opposite direction, to INDEFINITE.

newfrenchy83:
> > And I could definitely see an argument that it presents an obstacle to further development, since it's a special case in the ability system that everything needs to continue to account for.
> 
> This quote from my earlier post reads like token support to present me. That is to say, I'm not strongly in favour of it but not against it either.
> 
> Removing all the backstab logic but keeping the deprecation message along with code that substitutes the formula Soliton noted would also be a reasonable alternative. In that case, you'd probably push the deprecation level in the opposite direction, to INDEFINITE.

if i push level to INDEFINITE, i cannot remove the backstab logic to hard code. if remove the code, the level must be push to REMOVED to inform the wml develloper what backstab attribute don't work anymore.

CelticMinstrel:
> if i push level to INDEFINITE, i cannot remove the backstab logic to hard code. if remove the code, the level must be push to REMOVED to inform the wml develloper what backstab attribute don't work anymore.

If you remove the code, then yes, it should be changed to REMOVED. However, if you remove the code but add logic alongside the deprecation message that writes the tag @soliton- mentioned into the WML, that means backstab will continue to work even without all the supporting code, so it can be changed to INDEFINITE.

newfrenchy83:
@CelticMinstrel i cannot write the formula code in massage because a quotation marks problems in formula, this why a added refrence to abilities.cfg and WEAPON_SPECIAL_BACKSTAB for whatdevelloper know what sepcial work with this code and noanymore backstab attribute

CelticMinstrel:
That statement makes no sense whatsoever.

newfrenchy83:
@CelticMinstrel INDEFINITE pushed in https://github.com/wesnoth/wesnoth/pull/7031 PR

CelticMinstrel:
I don't know how you got that (d5535a866a0a376dffe391d651e32923d81bca1a) from what I said. That was nothing like what I said.

Pentarctagon:
@newfrenchy83 celticminstrel's statement was not about adding it to the deprecation message. It was about adding that WML into the weapon special config.

newfrenchy83:
itring but i don't find how.

CelticMinstrel:
It's definitely not a trivial thing. The current message is in a place where the config is const, so you'd need to move it to somewhere completely different, and I don't know where that place would be.

stevecotton:
I've created issue #7032 so that discussion is easier to find later, instead of being hidden from the issues' search list.

PR #6792, opened 2022-06-22
Fix retreat injured CA taking allied villages

mattsc:
There is a comment in the retreat injured CA code that taking ally side villages should be disallowed, with potential exceptions, but apparently I forgot to do that after porting this CA to the default AI. For 1.16, it is too late to do anything other than disallowing this.  That part should be done though, in order to make it consistent with previous behavior (and with player expectations).  That's the first commit in this PR, which should be backported.

The other two commits, for master only, add a new AI aspect `allow_ally_villages` which, when set, let's injured units take ally villages, but they will do so with lower priority than using own or enemy villages.  I decided against letting only "very injured" units do so in favor of a more clear-cut behavior, and also because there is already an aspect that affects how injured a unit has to be before it retreats.  Adding yet another sliding-scale parameter on top of that would be too confusing, in my opinion.


CelticMinstrel:
The amount of stuff needed to add a single new aspect is kinda crazy though…

(I assume you've already done it before and know you haven't missed anything.)

PR #6793, opened 2022-06-23
Fix Xcode warnings

mattsc:
Trying to fix the Xcode warnings.  I'm putting this up here to see what the CI run says, and also because in some cases I do not know if what I did is the preferred solution.  So let me know what needs to be changed.


mattsc:
I should have added that this is currently for the debug build only, and that another step will be to make the flags the same for debug and release builds.

projectfiles/Xcode/The Battle for Wesnoth.xcodeproj/project.pbxproj
```diff
@@ -6429,7 +6429,9 @@
 		460F21CA2858F72100EE2131 /* SignedRelease */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALWAYS_SEARCH_USER_PATHS = NO;
```
CelticMinstrel, line 6432: I'm kinda surprised this didn't just break everything...

PR #6794, opened 2022-06-23
Fix load_string() ignoring the passed chunk name

CelticMinstrel:
Complete with a unit test!
The test results in PASS_BY_VICTORY if it fails, which is a bit weird, but it does work.

There was a comment on this commit noting that it won't work without the events PR (#5663), but I suspect that comment is outdated, so I'm opening this to test the CI.

### DIFF
--- a/data/test/scenarios/test_lua_name.cfg
+++ b/data/test/scenarios/test_lua_name.cfg
@@ -1,0 +1,35 @@
+
+# Test that an error message in Lua includes the value of the tag's name key
+{GENERIC_UNIT_TEST test_lua_name (
+    [lua]
+        name=test
+        code=<<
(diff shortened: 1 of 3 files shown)


data/test/scenarios/test_lua_name.cfg
```diff
@@ -0,0 +1,18 @@
+
+# Test that an error message in Lua includes the value of the tag's name key
+{GENERIC_UNIT_TEST test_lua_name (
+    [event]
+        name=start
+        [lua]
+            name=test
```
Pentarctagon, line 7: Likely not needed, but I'd prefer the name here be something more unique than just "test".

Pentarctagon (approved):

Issue #2915, opened 2018-04-19
Very long unit animations cause the game to crash


inferno8:
Copied from:
https://forums.wesnoth.org/viewtopic.php?f=4&t=48005

This is probably very UMC-specific, but is related to the way the game handles animation so I decided to report it anyway.

Playing very long unit animations (400+ frames) leads to the following error:
`Could not create a new surface in get_surface_portion()`

The game crashes right after that or it is still running but cannot read the rest of images an animation is composed of, throwing `Surface lost` error to the console.

As far as I know, the error was observed on machines running BfW version: **1.13.14** SDL: 2.0.4 and BfW **1.12.6**, Windows 10 (10.0.16299).

I am fully aware almost no one would use huge animations like these, so this may not be relevant to the core Wesnoth. Nevertheless to help reproduce the problem I created a small test add-on in from of a multiplayer scenario for those who are interested:
[Long_Animation_Test.zip](https://github.com/wesnoth/wesnoth/files/1927082/Long_Animation_Test.zip)

Vultraz:
I... how... would you ever need such a long animation.... O_O

Vultraz:
A very quick look into this seems to indicate the problem is you're hitting Wesnoth's memory limit, as Wesnoth is still built as a 32 bit app on Windows. We're working on remedying that...

shikadiqueen:
You can get this problem on Linux too with higher numbers if you insist on building Wesnoth as an IA32 binary.

Also, it's not specific to the animation engine, it just happens to be a facilitator. The problem is that Wesnoth will try to cache every image it ever needs to load, and the images in question are *huge*, with each one taking up over 1.8 megabytes of memory without taking into account additional overhead from the attributes SDL needs to handle in memory, as well as Wesnoth's image cache structure. Have 1,000 of them and you start to run out of RAM — 32-bit Windows processes have a 2 GB limit for user mode mappings by default, and some of that memory is needed for libraries and memory-mapped files, as well as the execution stack and some miscellaneous structures.

inferno8:
> I... how... would you ever need such a long animation.... O_O

Such long animations are used in one of my add-ons to imitate cutscenes. The game does not support videos (yet) so this is a workaround. I use a dummy unit with [extra_anim] for this.

CelticMinstrel:
This may just go away with OpenGL, though if you're using screen-size images here to simulate a video, it could still be an issue…

hrubymar10:
[...]

PR #6795, opened 2022-06-24
Add options to skip adding images to the surface and texture caches

mesilliac:
I also added some diagnostics, and did some minor cleanup.

After checking the diagnostics, i set it to skip the surface cache automatically when loading textures from image files unmodified. These get cached as textures only. Sometimes this causes images to be loaded twice when the surface is needed later, but not very frequently.

I left it caching surfaces when loading IPF modified textures, as these turned out to be often reused for other image loading operations. It could be revisited after hardware-acceleration conversion is more complete.

I'm considering exposing the cache skip to WML for edge-case use of animations (see #2915 ), but it would be kind of ugly to do for halos, so i'm a bit reluctant.

Currently the point of this is mostly to avoid keeping copies of images both in graphics memory and main memory unless both are required.

### DIFF
--- a/src/picture.cpp
+++ b/src/picture.cpp
@@ -357,6 +357,18 @@ locator& locator::operator=(const locator& a)
 	return *this;
 }
 
+std::ostream& operator<<(std::ostream& s, const locator& l)
+{
+	s << l.get_filename();
+	if(!l.get_modifications().empty()) {
+		if(l.get_modifications()[0] != '~') {
+			s << '~';
+		}
+		s << l.get_modifications();
+	}
+	return s;
+}
+
 locator::value::value()
 	: type_(NONE)
 	, is_data_uri_(false)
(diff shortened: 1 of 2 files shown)


Pentarctagon:
Perhaps rather than exposing it to WML explicitly, it could automatically not cache images larger than a certain size?

mesilliac:
Hmm, that could work and would be cleaner. It might be problematic if somebody creates a huge repeating animation effect that does need to be cached. I think i heard somebody talking about animated weather effects somewhere? But i'm not sure if that was actually in-use or hypothetical.

Pentarctagon:
The only weather effect I'm aware of being in use is in Breeze Modification, which is done per-hex.

Pentarctagon:
Having it be configurable in the advanced preferences would also be good.

mesilliac:
Yep. I'll merge this as is for now with the new option, and look into applying it later.

Issue #6799, opened 2022-06-25
Rectangles are not drawn correctly


grz0:
### Game and System Information

The Battle for Wesnoth version 1.17.5+dev (666571ca-Clean) x86_64
Running on Freedesktop.org 21.08.14 (Flatpak runtime) x86_64
Distribution channel: Default

#### Libraries

Boost:             1.66
Lua:               5.4.4
OpenSSL/libcrypto: 1.1.1m (runtime 1.1.1o)
Cairo:             1.16.0 (runtime 1.16.0)
Pango:             1.48.10 (runtime 1.48.11)
SDL:               2.0.16 (runtime 2.0.16)
SDL_image:         2.0.5 (runtime 2.0.5)
SDL_mixer:         2.0.4 (runtime 2.0.4)

#### Current video settings

SDL video drivers:        [x11] wayland dummy
Window size:              1920x1015
Game canvas size:         1920x1015
Final render target size: 1920x1015
Screen refresh rate:      60

### Description of the bug

Screenshots:
![image](https://user-images.githubusercontent.com/3605541/175750841-731cf75d-82c5-4e87-afe6-0ce8eb610cb5.png)
![image](https://user-images.githubusercontent.com/3605541/175750850-2cd8d04d-0c57-44f1-a00b-5944baf85817.png)
![image](https://user-images.githubusercontent.com/3605541/175750857-5599f67a-c653-473a-bf69-4a115fcae7f8.png)

### Steps to reproduce the behavior

1. Start wesntho

mesilliac:
After some discussion on discord it looks like this is caused by an SDL bug: https://github.com/libsdl-org/SDL/issues/4001

It seems to have been introduced sometime around SDL 2.0.15 and fixed sometime around 2.0.18, and only affects some systems.

I'll look into putting an override in to draw rectangles a different way for bad SDL versions.

Wedge009:
I noticed this a while back too. Windows build via vcpkg is using SDL 2.0.22 and I don't observe the issue, but I do in Ubuntu 20.04.4, which uses 2.0.16.

Is it known what is the differentiator in only affecting 'some' systems?

mesilliac:
The upstream issue mentioned the openGL renderer specificaly having the issue, but it seemed like something they had trouble with in various forms at different times. It must be a bit of a nightmare keeping everything pixel-perfect on all the different systems.

mesilliac:
[...]

PR #6801, opened 2022-06-25
Draw rectangles as lines for some SDL versions.

mesilliac:
Aimed at fixing #6799.

According to the SDL bug linked there this was reported in SDL 2.0.15 and should have been fixed in 2.0.18. So i've set it to draw rectangles as four individual lines when linked against the affected SDL versions.

If the problem crops up again for others, the version check could be extended.

I'm assuming the problem doesn't also occur for lines.

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -87,9 +87,44 @@ void draw::set_color(const color_t& c)
 	SDL_SetRenderDrawColor(renderer(), c.r, c.g, c.b, c.a);
 }
 
+/** Some versions of SDL have a bad rectangle drawing implementation. */
+static bool sdl_bad_at_rects()
+{
+	// This could be done once at program start and cached,
+	// but it isn't all that heavy.
+	SDL_version ver;
+	SDL_GetVersion(&ver);
+	if (ver.major > 2 || ver.minor > 0) {
+		return false;
+	}
+	if (ver.patch >= 15 && ver.patch < 18) {
+		return true;
+	}
+	return false;
+}
+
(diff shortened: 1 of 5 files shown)


src/draw.cpp
```diff
@@ -87,9 +87,44 @@ void draw::set_color(const color_t& c)
...
+	SDL_version ver;
+	SDL_GetVersion(&ver);
+	if (ver.major > 2 || ver.minor > 0) {
+		return false;
+	}
+	if (ver.patch >= 15 && ver.patch < 18) {
+		return true;
+	}
+	return false;
+}
```
Pentarctagon, line 104: This seems like it should use `version_info`?

Wedge009:
This seems to work for me, though it's unfortunate that such a work-around is necessary. Thanks.

src/draw.cpp
mesilliac: Does that do SDL version handling?

It looks a bit heavy for something low level. Although i guess there's nothing that draws rectangles all that much

Pentarctagon: I mean you give it the major/minor/patch level numbers and then you can compare the objects. At a minimum though, `ver.major > 2 || ver.minor > 0` should surely be `ver.major == 2 || ver.minor == 0`.

Pentarctagon: Actually never mind, turns out I entirely misread it.

mesilliac: it not being clear is a pretty good reason to improve it... i'll add a cheap SDL-specific version check. The version_info class does several heap allocations so it's a bit much here.

Issue #6408, opened 2021-12-30
Lua error when attempting to use coward micro Ai


Hejnewar:
### Game and System Information
 - 1.16.1
 - Steam
 - Windows

### Describe the bug
![coward](https://user-images.githubusercontent.com/41817546/147756050-470d3c94-8f62-4806-88c8-d583124f4136.JPG)

### To Reproduce
Just try to use coward micro Ai with disctance set at number that is not a null value.

My current mai setup:
```
        [micro_ai]
            side = 3
            ai_type = coward
            action = add
            [filter]
            [/filter]
            seek_y = 1, 33
            [or]
                seek_x = 44
            [/or]
            distance = 10
        [/micro_ai]
```

CelticMinstrel:
I can see there's a problem somewhere in the code (this shouldn't cause the AI to crash), but your code itself is also wrong - `[or]` is not a valid tag there.

Hejnewar:
To be honest I was testing if that ```[or]``` would work in the first place. I didnt even get any error (which maybe should happen).

CelticMinstrel:
There _should_ be an error about an unrecognized tag… although it may be only in the console and thus not very visible?

I have a feeling the error you're seeing is unrelated to that though, most likely caused by having a comma-separated list in `seek_x`. The documentation doesn't say a list is supported.

Hejnewar:
Nope, I checked lua console and there is nothing about that there.

Yeah you are right that is the cause. Still should this be an error instead of information that it isn't supported?

CelticMinstrel:
If nothing else, it should be a more explanatory error that makes the actual problem clear.

mattsc:
> If nothing else, it should be a more explanatory error that makes the actual problem clear.

While this is quite simple to do for this parameter, it doesn't really make sense to do it for this parameter only and not for everything else.  Do we really want/need to do this?  Other WML/Lua code does not check either whether the parameters are the correct type and strange things and error messages can happen when that is not the case.  And also, isn't this what the schema validation is for?

CelticMinstrel:
It's always best for the error message to be as informative as possible. Of course you're right, it doesn't make sense to do it only for this parameter and not for everything else, and it doesn't make sense to do it for `[micro_ai]` and not for other WML tags, and doing it for everything could be quite a lot of work, but I do think it's best to fail fast rather than passing garbage down the line resulting in a cryptic error message later on.

[...]

PR #6802, opened 2022-06-25
Mai parms tests

mattsc:
This adds checks for the types of [micro_ai] tag parameters, addressing #6408.  It also adds warning messages if a parameter specified in a [micro_ai] tag is not valid for that MAI.

And while I was at it, I also added a couple bug fixes that I came across while working on this.

I am not planning to backport this to 1.16, as the risk of screwing something up in the stable series is too high.

### DIFF
--- a/data/ai/micro_ais/micro_ai_helper.lua
+++ b/data/ai/micro_ais/micro_ai_helper.lua
@@ -296,6 +296,61 @@ function micro_ai_helper.micro_ai_setup(cfg, CA_parms, required_keys, optional_k
             CA_cfg[k_name] = cfg[k_name]
         end
     end
+
+    -- Check whether there are invalid keys in the [micro_ai] tag
+    for k in pairs(cfg) do
+        if (not showed_deprecation_message[cfg.ai_type]) -- otherwise this produces false positives
+           and  (k ~= 'side') and (k ~= 'ai_type') and (k ~= 'action')
+           and (k ~= 'ca_id') and (k ~= 'ca_score') and (type(k) ~= 'number')
+        then
+            local is_invalid = true
+            for k_name in pairs(required_keys) do
+                if (k == k_name) then
+                    is_invalid = false
+                    break
+                end
+            end
+            if is_invalid then
+                for k_name in pairs(optional_keys) do
+                    if (k == k_name) then
+                        is_invalid = false
+                        break
+                    end
+                end
+            end
+            if is_invalid then
+                local str = "[micro_ai] tag (" .. cfg.ai_type .. ") contains invalid parameter: " .. k
+                warn(str)
+                std_print(str .. ' (see Lua console for stack trace)')
+            end
+        end
+    end
+
+    -- Check whether there are invalid tags in the [micro_ai] tag
+    for _,t in ipairs(cfg) do
+        -- [filter] is always added to [micro_ai] tags inside a unit's [ai] tag,
+        -- whether the specific MAI supports it or not
+        if (not showed_deprecation_message[cfg.ai_type]) and (t[1] ~= 'filter') then
+            local is_invalid = true
+            for k_name in pairs(required_keys) do
+                if (t[1] == k_name) then
+                    is_invalid = false
+                    break
+                end
+            end
+            if is_invalid then
+                for k_name in pairs(optional_keys) do
+                    if (t[1] == k_name) then
+                        is_invalid = false
+                        break
+                    end
+                end
+            end
+            if is_invalid then
+                str = "[micro_ai] tag (" .. cfg.ai_type .. ") contains invalid parameter: [" .. t[1] .. "]"
+                warn(str)
+                std_print(str .. ' (see Lua console for stack trace)')
+            end
         end
     end
 
--- a/data/ai/micro_ais/micro_ai_helper.lua
+++ b/data/ai/micro_ais/micro_ai_helper.lua
@@ -1,6 +1,74 @@
 local T = wml.tag
 local MAIUV = wesnoth.require "ai/micro_ais/micro_ai_unit_variables.lua"
 
+local showed_deprecation_message = {}
+
+local function show_deprecation_message(ai_type)
+    if (not showed_deprecation_message[ai_type]) then
+        wesnoth.deprecated_message('[micro_ai] tag (' .. ai_type .. '): The old syntax for required and optional parameters when creating a Micro AIs', 3, '1.19', "instead of { 'parameter' } it is now { parameter = 'parameter_type' }. See data/ai/micro_ai/mai-defs/ for examples." )
+        showed_deprecation_message[ai_type] = true
+    end
+end
+
+local function is_number(value, check_int, ai_type, name)
+    -- Allow any number for 'float' type.
+    -- Allow anything that has an integer representation for 'integer', but
+    -- issue warning if it is given in 'float' notation.
+    local number = tonumber(value)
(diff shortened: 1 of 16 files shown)


data/ai/micro_ais/micro_ai_helper.lua
```diff
@@ -3,6 +3,56 @@ local MAIUV = wesnoth.require "ai/micro_ais/micro_ai_unit_variables.lua"
...
+    -- Allow anything that has an integer representation for 'integer', but
+    -- issue warning if it is given in 'float' notation.
+    local number = tonumber(value)
+    if (not number) then
+        return false
+    elseif check_int then
+        if (number ~= math.floor(number)) then
+            return false
+        elseif (math.type(number) ~= 'integer') then
+            std_print("Warning: [micro_ai] tag (" .. ai_type .. ") parameter '" .. name .. "' must be an integer. It has an integer representation, but is provided in floating-point format.")
```
CelticMinstrel, line 17: I think it might make sense to use `warning` instead of `std_print` here? I believe that adds it to the Lua console.

data/ai/micro_ais/micro_ai_helper.lua
```diff
@@ -3,6 +3,56 @@ local MAIUV = wesnoth.require "ai/micro_ais/micro_ai_unit_variables.lua"
...
+end
+
+local function check_key_type(k_name, k_value, k_type, ai_type)
+    if k_value then
+        local is_wrong_type = false
+        if (k_type == 'float') then
+            is_wrong_type = not is_number(k_value, false, ai_type, k_name)
+        elseif (k_type == 'integer') then
+            is_wrong_type = not is_number(k_value, true, ai_type, k_name)
+        elseif (k_type == 'csv_integer') then
```
CelticMinstrel, line 31: 1. I'm not entirely sold on the name here. I wonder if "int_list" or "integer_list" would be easier to understand?

2. Maybe this is a low priority, but is there any demand for a list of real numbers?

data/ai/micro_ais/micro_ai_helper.lua
```diff
@@ -164,32 +214,42 @@ function micro_ai_helper.micro_ai_setup(cfg, CA_parms, required_keys, optional_k
...
     -- Required keys
-    for _,v in pairs(required_keys) do
-        if v:match('%[[a-zA-Z0-9_]+%]')  then
-            v = v:sub(2,-2)
-            if not wml.get_child(cfg, v) then
-                wml.error("[micro_ai] tag (" .. cfg.ai_type .. ") is missing required parameter: [" .. v .. "]")
+    for k_name,k_type in pairs(required_keys) do
+        if (k_type == 'tag') then
+            -- Check that this is not a scalar parameter
+            if cfg[k_name] then
```
CelticMinstrel, line 219: This is probably fine in the vast majority of cases, but keep in mind that it's possible to have a tag and a key with the same name. I think in that case it makes more sense to issue a warning rather than an error, if that's not too much work.

I guess it's not a priority though.

data/ai/micro_ais/micro_ai_helper.lua
```diff
@@ -164,32 +214,42 @@ function micro_ai_helper.micro_ai_setup(cfg, CA_parms, required_keys, optional_k
...
         end
     end
 
     -- Optional keys
-    for _,v in pairs(optional_keys) do
-        if v:match('%[[a-zA-Z0-9_]+%]')  then
-            v = v:sub(2,-2)
-            for child in wml.child_range(cfg, v) do
-                table.insert(CA_cfg, T[v](child))
+    for k_name,k_type in pairs(optional_keys) do
```
CelticMinstrel, line 238: Okay, sooo… I think this system (both loops) possibly counts as an API, which means this change should go through a deprecation cycle? What this means is essentially that you'd:

1. Add an `if type(k_name) = 'number'` case which uses the original code (where `k_type` is actually the key and `k_name` is ignored).
2. Add a one-time deprecation message in this case. Since this isn't a place where `deprecate_api` is possible, you'd probably do something like (pseudo-code) `if not showed_message then showed_message = true show_message() end`. The `showed_message` variable would be a file-level local variable.

CelticMinstrel (commented):
Looks good. I think my main complaints would be:

1. `warning` is new in Lua 5.4, and I think this is a great place to use it.
2. We _probably_ need a deprecation message and support for MicroAIs that have not been updated to the new style, since add-ons may contain their own custom MicroAIs.

data/ai/micro_ais/micro_ai_helper.lua
```diff
@@ -262,7 +299,8 @@ function micro_ai_helper.micro_ai_setup(cfg, CA_parms, required_keys, optional_k
 
     -- Check whether there are invalid keys in the [micro_ai] tag
     for k in pairs(cfg) do
-        if (k ~= 'side') and (k ~= 'ai_type') and (k ~= 'action')
+        if (not showed_deprecation_message) -- otherwise this produces false positives
```
CelticMinstrel, line 302: I suspect this check isn't what you want it to be. It's true if there has _ever_ been a deprecation message raised for _any_ Micro AI, not whether there has been a deprecation message for the specific Micro AI being considered.

I'm not quite sure what the best approach is here. You could just skip numeric keys (no wait, you already are? why doesn't that work?), or you could change the deprecation message to be once per Micro AI (so `showed_deprecation_message` becomes `showed_deprecation_message[ai_type]`).

mattsc: > I'm not quite sure what the best approach is here. You could just skip numeric keys (no wait, you already are? why doesn't that work?), 

Because this is the loop over the [micro_ai] tag config, not over required_key or optional_keys.

> or you could change the deprecation message to be once per Micro AI (so `showed_deprecation_message` becomes `showed_deprecation_message[ai_type]`).

Hmm.  I actually figured that getting a deprecation message only once overall was good enough.  It gets reset at each reload, after all, and deprecation messages are for UMC authors while they work on their add-ons, not something you need to worry about while you use an add-on.

However, there's a related (bigger?) issue with this, in that if one UMC creator does not bother dealing with deprecation messages, that might disable these warnings for another author.  So maybe making it MAI type specific is the way to go.

CelticMinstrel: Ohh, it shows false positives because the code within this loop doesn't handle the old required/optional keys format? Is that the reasoning?

If that's the case, it feels like it'd be better to put the "if not showed_deprecation_message" outside the loop… though that adds another level of indentation…

mattsc: > Ohh, it shows false positives because the code within this loop doesn't handle the old required/optional keys format? Is that the reasoning?

Yes.

> If that's the case, it feels like it'd be better to put the "if not showed_deprecation_message" outside the loop… though that adds another level of indentation…

Yeah, I know.  I did it this way because if I put it around the outside, then that whole chunk of code has to have its indentation changed whenever this is removed in the future, even though it is just two little pieces of conditional that need to be removed and it will look like a much bigger change than it is.  And it's not like this is a performance issue.  But it is a little "less clean" this way, so I can move it outside if you prefer.

CelticMinstrel: Ah, you have a good point there. Fair enough, we can leave it like this.

mattsc:
Okay, sounds good.
I just squashed things.  I'll go over it one more time tomorrow and then merge it unless there are further comments.

Issue #6149, opened 2021-10-03
high contrast mode


CelticMinstrel:

cooljeanius:
This was something @Zireael07 (a.k.a. "chess") suggested, and @max-torch seconded in #6133; I'm not quite sure how exactly it'd work myself, though...

Zireael07:
In most cases, high contrast mode is fairly easy to do if you work in HSV or HSL space instead of RGB. You just buff the Saturation part up.
If needed, dark outlines will usually help to separate two bright parts.

cooljeanius:
I think comparison against GitHub's own high contrast themes could be a useful place to look for inspiration:
https://github.blog/changelog/2024-08-15-high-contrast-theme-improvements/

cooljeanius:
Another tool that might possibly be useful here: https://github.com/taketwo/glasbey

Issue #6133, opened 2021-09-29
Add accessibility options for users with poor eyesight


cooljeanius:
Chess & I were brainstorming some of these ideas on Discord:
- [ ] color-blind mode
- [ ] high contrast mode
- [ ] add option to add outlines to all sprites, possibly in their team colors
- [ ] add option to outline the edges of tiles that units are standing on, possibly also in their team colors
- [ ] option to disable tile transitions

Let me know if any of these should be split off into separate bugs.

Zireael07:
3 & 4 would be awesome (I use glasses every day)

max-torch:
I would add
- [ ] Set brightness/contrast of highlighted hexes (the ones that show where a unit can move). 

It is especially hard to distinguish when you are in a cave with flat, road, cave, hill, terrain all mixed together.
Case in point:
![image](https://user-images.githubusercontent.com/41433185/135382324-9aee9ebf-cf99-49ff-82db-d2b61d5baf6c.png)
Can you easily distinguish the highlighted and unhighlighted hexes?

soliton-:
I would suggest to split the first two into separate issues and elaborate on what exactly those modes should entail. The rest seems fairly straight forward.

cooljeanius:
> I would suggest to split the first two into separate issues and elaborate on what exactly those modes should entail. The rest seems fairly straight forward.

Hm, I tried clicking the "Convert to issue" circle/button that appears when you hover over a task list item, but it doesn't seem to be working currently:
<img width="443" alt="Screen Shot 2021-10-03 at 11 34 38" src="https://user-images.githubusercontent.com/1388660/135761097-597862af-e58a-489c-b0f0-27e059445c4e.png">

CelticMinstrel:
It worked for me (perhaps it requires access to the repository), so go ahead and elaborate on the details.

cooljeanius:
Screenshots of the Discord convo that prompted this:
<img width="896" alt="Screen Shot 2021-10-03 at 12 39 49" src="https://user-images.githubusercontent.com/1388660/135763352-1ceab0c7-a000-4856-8d0f-e41559e82f3c.png">
<img width="893" alt="Screen Shot 2021-10-03 at 12 40 12" src="https://user-images.githubusercontent.com/1388660/135763360-1da9cad3-9824-4a89-a809-a94dbb45738a.png">

cooljeanius:
[...]

PR #6812, opened 2022-06-30
Address various minor highdpi TODOs

mesilliac:
Mostly these just involved checking if something was okay.

A couple unused things were removed.
1. The `hex_brightening` config option. This was not used, and from the looks of it has not been for quite some time. It was added in 2010 for testing the mouseover effect. The current effect is hardcoded and ignores this value.
2. There was a function dealing with "report" images, which actively analyzed them for transparency and automatically clipped them. This is ridiculous and i removed it some time ago. No adverse effect has been reported.

I originally thought "reports" were maybe some automatically generated graphs or something. But no. It is just the name for the automatically updated UI sections of the game display, such as the amount of village gold etc.. The icon images do not need clipping, and even if they do, it should be done via IPF.

### DIFF
--- a/data/game_config.cfg
+++ b/data/game_config.cfg
@@ -36,9 +36,6 @@
     # zoom factors must be a sorted list of ascending multipliers
     zoom_levels = 0.25, 0.33333333333333333, 0.5, 0.75, 1.0, 1.25, 1.5, 2.0, 3.0, 4.0
 
-    #temporary disable hex brightening
-    hex_brightening=1.25
-
     flag_rgb=flag_green
     unit_rgb=magenta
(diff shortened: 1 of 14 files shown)


cooljeanius:
> A couple unused things were removed.
> 
>    1. The `hex_brightening` config option. This was not used, and from the looks of it has not been for quite some time. It was added in 2010 for testing the mouseover effect. The current effect is hardcoded and ignores this value.

Wait this sounds like it could be used to address the "high contrast mode" part of #6133 perhaps? Or actually I guess that's #6149 now...

mesilliac:
It doesn't actually do anything. It was just a variable, sitting there.

From the sounds of it all it was ever used for was tweaking the brightness of the hex underneath the mouse cursor, and it seems from the original commit in 2010 like it was for testing the effect.

PR #6813, opened 2022-06-30
Inherit from SDL types for point and color_t

mesilliac:
This changes `color_t` and `point` to be thin wrappers over `SDL_Color` and `SDL_Point`.

They already had the same members, so usage is completely unchanged.

Aside from being a bit tidier to define, the practical differences are:
1. converting to and from the SDL versions no longer requires copying the data in most cases
2. color_t and point references can be passed directly to SDL functions requiring SDL_Color and SDL_Point pointers.

That second one was mostly why i wanted this.

### DIFF
--- a/src/sdl/point.hpp
+++ b/src/sdl/point.hpp
@@ -20,35 +20,15 @@
 #include <iosfwd>
 #include <tuple>
 
-/** Holds a 2D point. */
-struct point
+/** Holds a 2D point. This is a thin wrapper over SDL_Point. */
+struct point : SDL_Point
 {
-	point()
-		: x(0)
-		, y(0)
-	{
-	}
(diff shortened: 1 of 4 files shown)


src/sdl/point.hpp
```diff
@@ -20,35 +20,15 @@
 #include <iosfwd>
 #include <tuple>
 
-/** Holds a 2D point. */
-struct point
+/** Holds a 2D point. This is a thin wrapper over SDL_Point. */
+struct point : SDL_Point
```
CelticMinstrel, line 24: Doesn't this need a `public`? Or is that the default because it's a struct…?

mesilliac: That's right, it doesn't need public because it's the default for a struct.

Issue #3777, opened 2018-12-14
Error message uses absolute path


sevu:
This line prints the absolute path. Usually error messages don't do that.
https://github.com/wesnoth/wesnoth/blob/550c686e7b83e9465cd2811278c48f5c7db1069a/src/sound.cpp#L714

What does `track.file_path()` need to be replaced with?

alan-llin:
I believe it was intentional that they used `track.file_path()` for easier debugging. otherwise they would probably use `track.id()`.

CelticMinstrel:
Error messages really shouldn't ever print an absolute path since it could leak the user's login name.

ProditorMagnus:
Some more

20181231 17:17:15 error config: error reading usermade add-on 'C:\Users\Ravana\Documents\My Games\Wesnoth1.14/data/add-ons/Ageless_Era/_main.cfg'
20181231 13:12:44 error filesystem: Not opening 'C:\Users\Ravana\Documents\My Games\Wesnoth1.14/data/add-ons/Ageless_Resources/general/images/attacks/sunburst.png' due to case mismatch.

jostephd:
I don't understand why this is a problem. Errors should usually be verbose to make debugging easier, and they aren't sent to wesnothd or anything, only to the log file, so who exactly are we censoring them from?

ProditorMagnus:
I believe it is in case some people use their real name as username.

CelticMinstrel:
People paste log messages in error reports. Some people don't take care to ensure their username isn't in the log. Some people might not care, but we shouldn't create opportunities for people to accidentally include their username (which could be their real name) in a report.

For WML files, the WML-visible path (ie, relative to the data folder or the addons folder) is entirely sufficient to track down the offending file, so there's no need to give any more info than that. For assets (which includes translations), it's a bit more tricky since they're drawn from all over the place; you'd probably need to do some sort of sanitization similar to what I did in the report command-line option (the option that prints out all the versions of dependent libraries and stuff).

PR #6818, opened 2022-06-30
Sound path

mattsc:
Making this a PR only to make sure I am not missing something and CI passes.

### DIFF
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "log.hpp"
+#include <boost/algorithm/string.hpp>
 
 #include <map>
 #include <sstream>
@@ -207,6 +208,25 @@ static void print_precise_timestamp(std::ostream& out) noexcept
 	} catch(...) {}
 }
 
+std::string sanitize_log(const std::string& logstr)
+{
+	std::string str = logstr;
+
+#ifdef _WIN32
+	const char* user_name = getenv("USERNAME");
+	if(user_name != nullptr) {
+		boost::replace_all(str, std::string("\\") + user_name + "\\", "\\USER\\");
+	}
+#else
+	const char* user_name = getenv("USER");
+	if(user_name != nullptr) {
+		boost::replace_all(str, std::string("/") + user_name + "/", "/USER/");
+	}
+#endif
+
+	return str;
+}
+
 log_in_progress logger::operator()(const log_domain& domain, bool show_names, bool do_indent) const
 {
 	if (severity_ > domain.domain_->second) {
@@ -248,7 +268,7 @@ void log_in_progress::operator|(formatter&& message)
 			stream_ << get_timestamp(std::time(nullptr));
 		}
 	}
-	stream_ << prefix_ << message.str();
+	stream_ << prefix_ << sanitize_log(message.str());
 }
 
 void log_in_progress::set_indent(int level) {
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -551,11 +551,11 @@ const std::string& get_version_path_suffix()
 		if(bfs::is_directory(new_saves_dir)) {
 			if(!bfs::exists(old_saves_dir)) {
 				std::cout << "Apple developer's userdata migration: ";
-				std::cout << "symlinking " << old_saves_dir << " to " << new_saves_dir << "\n";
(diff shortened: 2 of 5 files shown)


mattsc:
Well, and to make sure it does not include another commit, I guess ...
I'll fix that.

mattsc:
Oh, I just noticed, the comments in #3777 point out other occurrences of this as well.  I'll check on those later, don't have time right this moment.

mattsc:
So ...  I've now done this for file_system.cpp and config_cache.cpp, see a0a790b12.  This removes the actual user name from a lot of output, as can be seen if you run Wesnoth with `--log-debug=filesystem` or `--log-debug=cache`.

But ...  With that, I am probably less than 20% done.  Before I continue, I'd like some input as to whether this is actually worth it.  And more so, whether we actually _want_ to do this.  It's not difficult work, but it is tedious, so I'd like to know whether it is worth continuing.

Pentarctagon:
It seems like this might be something better done in the logger itself, rather than trying to track down every individual place that might log the current user? So like have the logger check every log statement for the text `/<current user>/` and substitute it if found.

mattsc:
Hmm, yeah, good point.  I should have thought of that myself.  There are some instances that do not go through the logger, but those are comparatively few and should be easy to take care of separately.

mattsc:
Kicking myself a little here, but at least for the most part it seems to be as simple as 082d23aece

Obviously I am going to clean all of this up later, for now I just want to see if this passes CI.

Pentarctagon:
Does sanitize_path work as-is? Mainly wondering what happens when it calls normalize_path on something that's not a filesystem path string.

mattsc:
> Does sanitize_path work as-is? Mainly wondering what happens when it calls normalize_path on something that's not a filesystem path string.

It does work as is, it just does a string raplacement of the user name to "USER". If that user name does not appear, nothing happens.  Except that it also "normalizes" the path, showing the separators the same on all systems in the output, independent of what the OS uses, if I understand correctly.

So all of that has two problems.  If the user name is something that also appears in other log strings (such as, say, 'ai' or 'Konrad'), or if the separator might appear in an equation or operation, for example when the output is code.

Bottom line is, doing this by default for all strings in the logger is probably not a good idea.  We either need to detect whether the message is a directory (I don't know if we can do that reliably) or do it manually for each log message using file names or paths after all.

Pentarctagon:
I was thinking more that it would make more sense to just not use sanitize_path at all. Rather, something like:
```
void sanitize_log(std::string& logstr)
{
    #ifdef _WIN32
	    const char* user_name = getenv("USERNAME");
	    boost::replace_all(logstr, "\\"+user_name+"\\", "\\USER\\");
    #else
	    const char* user_name = getenv("USER");
	    boost::replace_all(logstr, "/"+user_name+"/", "/USER/");
    #endif
}
```

mattsc:
So there's my next attempt.  That seems to work, but I'm not sure about all that back and forth between string and const string and const char* and whatnot.  Couldn't get it to work otherwise, but I am probably just missing something simple.

src/log.cpp
```diff
@@ -207,6 +208,25 @@ static void print_precise_timestamp(std::ostream& out) noexcept
 	} catch(...) {}
 }
 
+std::string sanitize_log(const std::string& path)
```
Pentarctagon, line 211: It'd be good to call this something other than `path` and `tmp_path`, since these aren't file paths anymore.

Pentarctagon:
Overall it looks fine to me. I do wonder if we need sanitize_path at all now.

mattsc:
@Pentarctagon I think this should be ready now.  I did the following:
- Removed the griiiiit (which made it pass unit tests)
- Renamed variables as you suggested
- Went through the code and added `filesystem::sanitize_path` to output that does not go through the logger.  There is quite a bit of that, actually, incl. some that already used it and therefore does not show up in the commit.

Pentarctagon (approved):

Pentarctagon:
I do also wonder why Wesnoth has places bypassing regular logging, but I can take a look at that after this is merged.

mattsc:
> I do also wonder why Wesnoth has places bypassing regular logging, but I can take a look at that after this is merged.

Yeah, that was my take on that.

PR #6819, opened 2022-06-30
[WC] Cleanup macro+translatable string usage

### LINKED ISSUE #5114: wc: Translation issues with artifact descriptions due to string concatentation
Currently is used a lot of string concatinations, it'd be better to use vformat instead-

Pentarctagon:
I believe this should also resolve #5114, and also just generally makes it easier to see what text is being used when reading the code.


data/campaigns/World_Conquest/resources/strings/mainline.cfg
```diff
@@ -1,90 +1,40 @@
...
-
 #define WC_STR_CONCEALMENT
 _ "concealment" #enddef
 
 #define WC_STR_CURES
-_ "heals +8"+", "+_ "cures" #enddef
-
-#define WC_STR_DEFENSE
-_ "Defense" #enddef
+_ "heals +8, cures" #enddef
```
CelticMinstrel, line 21: Is this actually a core string?

CelticMinstrel (approved):

data/campaigns/World_Conquest/resources/strings/mainline.cfg
Pentarctagon: Ah, yeah, that should be in wesnoth-wc.

CelticMinstrel: …does it even belong in this file then?

Pentarctagon: Well, it was in this file before.

CelticMinstrel: Sure, but that was when it was a mainline string. Now it's not.

Pentarctagon: I mean feel free to move it somewhere else if you want.

PR #6822, opened 2022-07-02
Tweak debug representations of color_t, point and SDL_Rect

mesilliac:
These were somewhat inconsistent, and could be difficult to read. This only affects how these are written to an ostream.

It's basically as discussed in https://github.com/wesnoth/wesnoth/pull/6814#issuecomment-1171870532

The previous representations were:
```
color_t(255, 60, 140, 255) --> "255 60 140 255"
point(10,25) --> "10,25"
SDL_Rect(3, 15, 22, 89) --> "x: 3, y: 15, w: 22, h: 89"
```

The new representtions are:
```
color_t(255, 60, 140, 255) --> "#ff3c8c"
color_t(255, 60, 140, 128) --> "#ff3c8c80"
point(10,25) --> "(10,25)"
SDL_Rect(3, 15, 22, 89) --> "[3,15|22,89]"
```

The vertical bar in the rect representation is to differentiate the second pair as a size, not a position, and also (along with the square brackets) to make it clear that it's a rect, not some other 4-integer-element object.

The reasoning for this change is to make the representations both compact and clear for easy parsing in logs and debug output.

### DIFF
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -44,7 +44,7 @@ void draw::fill(
 	const SDL_Rect& area,
 	uint8_t r, uint8_t g, uint8_t b, uint8_t a)
 {
-	DBG_D << "fill " << area << " (" << color_t{r,g,b,a} << ")" << endl;
+	DBG_D << "fill " << area << ' ' << color_t{r,g,b,a} << endl;
 	SDL_SetRenderDrawColor(renderer(), r, g, b, a);
 	SDL_RenderFillRect(renderer(), &area);
 }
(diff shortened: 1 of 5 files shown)


src/draw.cpp
```diff
@@ -69,15 +69,13 @@ void draw::fill(const SDL_Rect& area)
...
-	      << " [" << r << ',' << g << ',' << b << ',' << a << ']' << endl;
+	DBG_D << "set color " << color_t{r,g,b,a} << endl;
 	SDL_SetRenderDrawColor(renderer(), r, g, b, a);
 }
 
 void draw::set_color(uint8_t r, uint8_t g, uint8_t b)
 {
-	DBG_D << "set color "
-	      << " [" << r << ',' << g << ',' << b << ']' << endl;
+	DBG_D << "set color " << color_t{r,g,b,SDL_ALPHA_OPAQUE} << endl;
```
CelticMinstrel, line 78: Doesn't it also work if you omit the `SDL_ALPHA_OPAQUE`?

mesilliac: good point, yes it does

PR #6827, opened 2022-07-02
Run Boost unit tests on MacOS CI

CelticMinstrel:
Currently this is just to test if it even works. I'm not sure whether or not it's a good idea.

### DIFF
--- a/.github/workflows/ci-scripts/macos.sh
+++ b/.github/workflows/ci-scripts/macos.sh
@@ -22,7 +22,8 @@ fi
 if [ "$CFG" == "Debug" ]; then
 	xcodebuild -project "The Battle for Wesnoth.xcodeproj" -target "unit_tests" -configuration "$CFG"
 	if [ $? == 0 ] && [ "$CFG" == "Debug" ]; then
-		./unit_tests --color_output --log_level=test_suite
+		cd 'projectfiles/Xcode/The Battle for Wesnoth.app/Content/Resources'
+		../../../unit_tests --color_output --log_level=test_suite
 		EXIT_VAL=$?
 	fi
 fi
--- a/.github/workflows/ci-scripts/macos.sh
+++ b/.github/workflows/ci-scripts/macos.sh
@@ -21,12 +21,7 @@ fi
 
 if [ "$CFG" == "Debug" ]; then
 	if [ $? == 0 ] && [ "$CFG" == "Debug" ]; then
-		echo "Entering 'build'" ; cd build ; ls
-		echo "Entering '$CFG'" ; cd $CFG ; ls
-		echo "Entering '.app'" ; cd "The Battle for Wesnoth.app" ; ls
-		echo "Entering 'Content'" ; cd Content ; ls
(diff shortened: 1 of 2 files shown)


CelticMinstrel:
Anyone know what to do about the log being truncated? There's no point in even doing this if the logs won't show.

Pentarctagon:
The full logs show if you look at the raw logs, though it's odd that it's truncating at ~10k lines when the macOS jobs that run for non-PRs already exceeds that.

It looks like part of the issue also is that it's essentially building everything 4 times? For example it you search for `/Users/runner/work/wesnoth/wesnoth/src/ai/composite/ai.cpp` in the raw logs, it shows up for:
* x86_64 for target The Battle for Wesnoth
* arm64 for target The Battle for Wesnoth
* x86_64 for target unit_tests
* arm64 for target unit_tests

CelticMinstrel:
> The full logs show if you look at the raw logs

Yeah, but those don't show colours.

Let's see if this commit works…

Pentarctagon:
From the log it looks like it's at `/Users/runner/work/wesnoth/wesnoth/projectfiles/Xcode/build/Debug/unit_tests`?

CelticMinstrel:
Okay, looks like this works now. The failure in the CI is the same failure I saw locally.

CelticMinstrel:
The failing test is now fixed on master, but should this be merged? Am I even doing it right?

.github/workflows/ci-scripts/macos.sh
```diff
@@ -19,4 +19,12 @@ if [ $EXIT_VAL == 0 ] && [ "$CFG" == "Release" ]; then
 		EXIT_VAL=$?
 fi
 
+if [ "$CFG" == "Debug" ]; then
+	if [ $? == 0 ] && [ "$CFG" == "Debug" ]; then
```
Pentarctagon, line 23: I'd say delete both these if-checks, unless there's a reason to only run the boost unit tests with the debug executable?

Pentarctagon:
It'd be good to see the boost unit tests pass on the CI before merging, just to be sure.

.github/workflows/ci-scripts/macos.sh
CelticMinstrel: I really have no idea if there's a reason, but if you want them deleted, then I'll delete them.

Pentarctagon: For the WML tests, that's only on release builds since they tend to cause failures by timing out when using the debug build.

CelticMinstrel:
Uhh… somehow it can't find the resources folder, but only on Release…?

Pentarctagon:
Does release put it in a different folder or something? The release wesnoth executable ends up in `Contents/MacOS/` rather than `Contents/Resources/` at least.

CelticMinstrel:
I'm switching to Resources because that's where the data directory is; I don't need the Wesnoth executable to run the unit tests, but the data directory needs to be in the current working directory.

I suppose I can reverse the logic and switch to the repository root instead… I'll try that.

Issue #6508, opened 2022-02-11
[Graphics] Drake Arbiter/Warden pop outside of map hex when attacking or defending.


knyghtmare:
## Wesnoth Version
1.17.0+dev

https://cdn.discordapp.com/attachments/259976436490829825/939623417882226748/IQ-Clash_of_the_Drakes_Turn_7.2.gz
(requires Inky's Quest to be installed)

Observed Here:
Video: https://www.twitch.tv/videos/1288191641
TimeStamp: 20:24

Reproduce:
1. Spawn Drake Arbiter on a ford/shallow water hex
2. Observe it being attacked and defending.
3. Observe it retaliate in attack
4. It can be seen he is levitating on water

(Not observed for Drake Clasher/Thrasher/Enforcer)

![image](https://user-images.githubusercontent.com/5283677/153634306-4943103d-bede-4343-b519-369591de1751.png)

Reported by @Konrad22

doofus-01:
I can't view the video and am not seeing the issue when I test this in mainline 1.16.  Is this specific to 1.17?

Konrad22:
> In 1.16.2, Arbiter pops out of the water, Warden looks good. ![Arbiter def](https://user-images.githubusercontent.com/30196839/153749398-6fc06fc0-af9b-46ae-b640-bc0ef62f9fe6.png)

If you compare the feet of the Wardens, you'll see that the idle ones feet are far deeper under water. It's not as obvious as the Arbiter, where the one defending is plainly above the water.

FTR, right now it doesn't trigger for me on shallow water, but it does trigger on Ford or Coastal Reef.

Descacharrado:
> If you compare the feet of the Wardens, you'll see that the idle ones feet are far deeper under water. It's not as obvious as the Arbiter, where the one defending is plainly above the water.

I have looked at the files and it looks like that happens with every single image that is not 72x72 (the norm for wesnoth unit image sizes), will open a PR soon.

PR #6518, opened 2022-02-13
Fixed Warden’s and Arbiter’s animations popping on water tiles

Descacharrado:
Will add animation gifs if needed. Needs more testing. Diff says there were deletions even tho  there were no deletions, only additions and code repositioning.

Closes #6508 
Addresses my request in #6475 (my name should be deleted from that issue when backported)

PR #6828, opened 2022-07-02
Fixed Warden’s and Arbiter’s animations popping on water tiles.

Descacharrado:
Previously, part of these drakes were popping out of the water during attack and defense animations because their images were bigger than a hex. This adds hand-picked submerge values that makes the animation feel more connected on water.

Closes https://github.com/wesnoth/wesnoth/issues/6508

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@
  ### Terrain
  ### Translations
  ### Units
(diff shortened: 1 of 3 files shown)


CelticMinstrel:
Just noting that this is continued from #6518

doofus-01 (approved):
Looks right, thanks.
EDIT:  I don't think the CI/flatpak failure is because of anything here.

Issue #6834, opened 2022-07-03
Gcc compilation fail for test when using codeblock projectfiles


newfrenchy83:
### Game and System Information

- **Version:master
- **Downloaded from:here
- **Build info:[**]compiled with gcc-8.1.0  --threads=posix --exceptions=seh --arch=x86_64 after commit (https://github.com/wesnoth/wesnoth/commit/8ed261053ddbc34ae2760deaeca98b2c00133883)
- **OS:**windows 7

### Description of the bug

the compilation of src/test/main.cpp fail.

### Steps to reproduce the behavior

load codeblocks
2load codeblock workspace in projectfiles/codeblocks folder
3. compile src/tests/main.cpp file.

### Expected behavior

file compiled normally.

### Additional context

[log.txt](https://github.com/wesnoth/wesnoth/files/9034875/log.txt)

CelticMinstrel:
This is probably not unique to CodeBlocks. I suspect it would happen in older Boost versions on any Windows computer.

newfrenchy83:
> This is probably not unique to CodeBlocks. I suspect it would happen in older Boost versions on any Windows computer.

boost 1.78 is already too old?

CelticMinstrel:
> boost 1.78 is already too old?

I really have no idea… I just vaguely recall this kind of issue arising in the past with Boost.Filesystem.

PR #6835, opened 2022-07-03
fix compilation issue in src/test/main.cpp

newfrenchy83:
Fix https://github.com/wesnoth/wesnoth/issues/6834 issue in wait of best resolving.

### DIFF
--- a/src/tests/main.cpp
+++ b/src/tests/main.cpp
@@ -61,11 +61,8 @@ std::ofstream reporter;
 struct wesnoth_global_fixture {
 	wesnoth_global_fixture()
 	{
-#ifndef __APPLE__
 		using namespace boost::unit_test;
-		reporter.open("boost_test_result.xml");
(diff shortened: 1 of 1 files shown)


src/tests/main.cpp
```diff
@@ -71,6 +75,7 @@ struct wesnoth_global_fixture {
 		}
 
 		reporter.open(file.c_str());
```
stevecotton, line 77: Is it also fixed by changing this line to use `file.string()` instead of `file.c_str()`?

CelticMinstrel:
This is very obviously not the correct fix.

While you're in that file by the way, would you mind putting `using namespace std::literals` on its own line? I apparently forgot to do that before pushing.

src/tests/main.cpp
newfrenchy83: it done

PR #6851, opened 2022-07-08
Add raise_user_interact calls to Lua CAs

mattsc:


CelticMinstrel:
Okay… this mostly seems fine. The only weird thing is that it's clearly an AI function yet is made available to non-AI code, and I'm not sure if this actually works. Before merging we should try using it in some non-AI code and see if it does its job there as well (for example, a WML event containing Lua code with a big loop).

If it doesn't work in non-AI code, I think we should either find a different way to do it that does, or move it into the AI module.

mattsc:
Good point.  I used this event (added to the Freelands MP scenario, but I don't think that is relevant information):
```
[event]
    name=moveto
    first_time_only=no

    [lua]
        code = <<
            for i = 1,1000 do
                std_print(i)
                wesnoth.interface.raise_user_interact()
                local unit = wesnoth.units.find_on_map()[1]
                wesnoth.paths.find_cost_map(unit)
            end
        >>
    [/lua]
[/event]
```
and it does work.  Also, it clearly disables UI interaction if the raise_user_interface line is not there.

stevecotton:
I suggest @CelticMinstrel is the best reviewer for this.

Even though there's a rate-limiter in the implementation of `raise_user_interact()`, I wonder a bit about whether the calls should be moved outside the innermost AI loops instead of triggering on each pathfinding operation. Have you done any timing tests?

BTW, the Wiki doesn't seem to have documentation for `wesnoth.interface.raise_user_interact`.

mattsc:
> Even though there's a rate-limiter in the implementation of `raise_user_interact()`, I wonder a bit about whether the calls should be moved outside the innermost AI loops instead of triggering on each pathfinding operation. Have you done any timing tests?

I was thinking about that too yesterday and did do some timing tests.  1000 calls took 20-30 millisec on average on my computer, or 20-30 microsec for a single call, so I decided not to worry about where exactly I place them.  If we have more than 1000 path finding operations in more than the occasional extreme case, there's a bigger problem we need to worry about in my opinion, and they will dominate the computation time needed.

> BTW, the Wiki doesn't seem to have documentation for `wesnoth.interface.raise_user_interact`.

It's on my list to add to the wiki, but I wasn't going to do that before I know that this gets actually included.

CelticMinstrel:
The only thing I'm not sure about is the name. Does `allow_user_interact` maybe sound a bit better?

mattsc:
> The only thing I'm not sure about is the name. Does `allow_user_interact` maybe sound a bit better?

I'm fine with it either way.  I just used the same name that the C++ function uses.

CelticMinstrel:
For reference, the name I used in my attempt was `allow_interact`. Adding the `user` to that does seem fine though.

I understand that the first instinct is to use the same name as the C++ function… but the C++ function is not a public API, so it's not a big deal if the name is a bit strange. If we're promoting it to a public Lua API, I think we should consider whether it's actually a good name.

(And if having different names in Lua and C++ is a concern, we can always change the C++ function's name to match whatever's decided for Lua.)

stevecotton:
We currently have `allow_end_turn`, which takes a boolean to enable/disable part of the UI; I suggest treating all names starting with `allow_` as reserved for similar setter functions.

Maybe `process_user_interaction` or similar?

mattsc:
No need to convince me that it is okay to use a different name.  I just meant that I did not give it much (any, really) thought so far, mostly because I don't care much.  So the two (or more) of you can sort it out and then I'll change it to whatever you come up with.  I do agree, now that it's been said, that something starting with "allow" sounds like there's a yes/no option to it.

CelticMinstrel:
What about `handle_user_interact`?

mattsc:
I renamed it to `handle_user_interact`, added the call to some of the Micro AI CAs, and squashed and rebased.  As far as I can tell, this is done now.

mattsc:
I have now [documented this on the wiki](https://wiki.wesnoth.org/LuaAPI/wesnoth/interface#wesnoth.interface.handle_user_interact) also. I wasn't quite sure how to describe this, so please feel free to improve on it.

CelticMinstrel:
If there are no objections, I'll probably merge this by tomorrow.

stevecotton:
Looks good to me, thanks.

PR #6855, opened 2022-07-09
A draw manager. To manage drawing to the screen.

mesilliac:
This is now basically feature-complete and fully functional.

It is a complete overhaul of the way things are drawn to the screen.

In summary:

 * objects no-longer draw by hooking a draw event. In stead they inherit from a new (mostly-)abstract base class `gui2::top_level_drawable`, and implement its interface.
 * the gamemap now renders to an offscreen buffer. This is used to implement hardware scrolling, and to redraw after halos, floating labels etc. move.
 * halos, floating labels, tooltips, and a few more things are now drawn on top of the gamemap, rather than as part of it. This allows them to be updated independently without reading pixels from the screen.
 * surface restorers have been removed. Reading pixels from the screen should now be unnecessary excepting two cases: (a) screenshots, (b) background blur. Blur is cached, and screenshots are occasional.
 * GUI2 widgets no longer keep track of dirty state. They are redrawn as necessary. Most places which previously set dirty state now queue a redraw in their part of the screen in stead.
 * A consequence is that active translucency is enabled across all UI elements, and the game display can (and does) continue to animate while menus and dialogs are showing.
 * performance is drastically increased for basically everything, most notably map scrolling, floating text, and halos.
 * CPU usage is drastically decreased. With animations disabled it is essentially zero while nothing is moving.
 * GPU usage is also minimal. The display is only flipped if something is drawn.

I haven't merged with new_rendering yet, but that shouldn't be too much trouble. For now i'm putting this up in a state that can be tested. There will be bugs. But there shouldn't be anything too major or insoluble.

Apologies for the megacommit, but it really was a mess. Easier just to commit the final cleaned version than try to preserve any separable changes.

### DIFF
--- a/src/floating_label.cpp
+++ b/src/floating_label.cpp
@@ -127,7 +127,7 @@ bool floating_label::create_texture()
 			text.set_text(text_, use_markup_);
 		}
 
-		surface foreground = text.render();
+		surface foreground = text.render_surface();
 
 		// Pixel scaling is necessary as we are manipulating the raw surface
 		const int ps = CVideo::get_singleton().get_pixel_scale();
--- a/src/floating_label.cpp
+++ b/src/floating_label.cpp
@@ -17,6 +17,7 @@
 
 #include "display.hpp"
 #include "draw.hpp"
+#include "draw_manager.hpp"
 #include "font/text.hpp"
 #include "log.hpp"
 #include "sdl/utils.hpp"
@@ -32,6 +33,11 @@ static lg::log_domain log_font("font");
 #define WRN_FT LOG_STREAM(warn, log_font)
 #define ERR_FT LOG_STREAM(err, log_font)
 
+static lg::log_domain log_display("display");
+#define ERR_DP LOG_STREAM(err, log_display)
+
+using std::endl;
+
 namespace
 {
 typedef std::map<int, font::floating_label> label_map;
@@ -41,16 +47,13 @@ int label_id = 1;
 std::stack<std::set<int>> label_contexts;
 }
 
+// TODO: draw_manager - why tf is this in namespace font?
 namespace font
 {
 floating_label::floating_label(const std::string& text, const surface& surf)
-#if 0
-	: img_(),
-#else
 	: tex_()
-	, buf_()
-	, buf_pos_()
-#endif
+	, screen_loc_()
+	, alpha_(0)
 	, fadeout_(0)
 	, time_start_(0)
 	, text_(text)
@@ -97,8 +100,13 @@ int floating_label::xpos(std::size_t width) const
 
 bool floating_label::create_texture()
 {
+	if(CVideo::get_singleton().faked()) {
+		return false;
+	}
+
+	// TODO: highdpi - this really does not need the extra indent
 	if(tex_ == nullptr) {
-		DBG_FT << "creating floating label texture" << std::endl;
+		DBG_FT << "creating floating label texture" << endl;
 		font::pango_text& text = font::get_text_renderer();
 
 		text.set_link_aware(false)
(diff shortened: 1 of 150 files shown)


src/floating_label.cpp
```diff
@@ -189,7 +189,8 @@ void floating_label::undraw()
 void floating_label::update(int time)
 {
 	if(!create_texture()) {
-		ERR_FT << "failed to create texture for floating label" << std::endl;
+		// Can't display this error because WML tests die...
+		//ERR_FT << "failed to create texture for floating label" << std::endl;
```
CelticMinstrel, line 193: Is there a way to check if the display is faked and skip creating the texture in that case? WML tests are run with implicit `--nogui` so they shouldn't be rendering anything anyway.

mesilliac: technically yes...

mesilliac:
There are still one or two minor glitches, but i think on the whole this is ready enough.

Assuming it works similarly well for other people, at least.

There is still a lot of mess and many to-do statements in the code. I prefixed them all with `TODO: draw_manager`, so they can be easily found and dealt with one-by-one.

I also added a screen fade effect to the game display. It's used to fade in at the start of a scenario, fixing some flickering.

Pentarctagon:
What's with all the `Create an empty image` text printed during the WML unit tests?

mesilliac:
Hmm. It seems that doesn't show if the tests pass, which they do for me. Looks like it comes from trying to regenerate the minimap, which was previously done during draw. I put a check there to not do that if display is faked.

mesilliac:
I have no idea why these WML tests are failing, they work for me...

irydacea:
> GUI2 widgets no longer keep track of dirty state. They are redrawn as necessary. Most places which previously set dirty state now queue a redraw in their part of the screen in stead.

Have you made sure this will not result in interesting rendering such as half-drawn widgets appearing on the screen because someone once assumed that they will be able to complete multiple partial draws before the result is actually committed to the screen?

mesilliac:
Mostly, yes.

I split all draw calls into three separate phases:
1. Layout and update - this ensures the widget layout is up to date. Animations etc can be updated here as well.
2. Render - this allows widgets to keep their own offscreen render buffer and render to it as they wish. The game display in particular uses this, but most widgets skip it and just draw on demand.
3. Expose - this is where actual draw calls happen. Calling it recursively, invalidating regions during it (setting things dirty), and modifying widget layout, are all forbidden. For the game display it just blits the right portion of its render buffer in this call, so it's very cheap.

Each vsync these are called on every top-level drawable object. Each phase individually calls the corresponding virtual function on each top-level drawable in the correct drawing order. The expose phase keeps a list of regions of the screen that need updating, and calls the whole draw stack for each region. Regions are combined so that a minimal set of draws is done.

In stead of setting a dirty flag on the widget, the region of the screen that the widget occupies is added to the list of regions to be redrawn. This ensures that all drawables in that region are updated, rather than just the widget and whatever else has checked its flag. That means for example transparency will work nicely, correctly redrawing whatever's underneath the widget container, as well as the container and widget. Anything above will get redrawn as well, so for example map animations can show thru a transparent popup message, which will redraw every time the animation under it invalidates.

Most of the work in this was splitting out all the things that were doing layout inside draw calls, and drawing inside layout, and other similarly horrible things.

I still need to go thru and make sure nothing is drawing where it shouldn't, but it's at least fairly clear now how the drawing should be done to avoid partial or incomplete draws. Widgets just have to do layout inside layout() and draw themselves inside draw(), it's pretty much that simple.

It also might do to split the animation phase out of Layout. For now it wasn't very important, as none of the UI is animated. But in the future it would be better to have an "Update" phase before "Layout" so widgets can smoothly move around, resize, modify contents, etc.

The drawing model is based on GTK's drawing model: [GTK3](https://docs.gtk.org/gtk3/drawing-model.html), [GTK4](https://docs.gtk.org/gtk4/drawing-model.html).

mesilliac:
If nobody has any major objections, i would like to merge this ASAP. I want to get it into nightlies for wider testing, and into 1.17.6 so i can address any bugs while the solutions are still fresh.

Wedge009:
To answer your question from #wesnoth-dev and here, I don't have any objections. I'll be sure to report anything I find that's amiss, as with the map preview redraw issue from the MP lobby.

PR #6856, opened 2022-07-09
Split macos build into separate steps

soliton-:


Pentarctagon:
It looks like macros.sh can be deleted then?

soliton-:
Yeah.

soliton-:
The disk image creation (hdiutil) can also go since it's not uploaded anymore or is it useful still to do it?

Issue #6745, opened 2022-06-04
[Feature Request] Make screen fade in/out affect everything on a map including units


inferno8:
### Describe the desired feature

Currently fade in and out effects are handled by delayed steps of [color_adjust] which tints the color of the screen except for units, their hp bars and labels. So basically it looks like this:
![fadetoblack-1 16](https://user-images.githubusercontent.com/35965303/171998024-52d8d4b7-6e82-42c8-8142-0d62b5303899.gif)
(BfW 1.16.x - final fade result: units stand on the black screen)
_If someone wants to quickly test this and see for themselves, they can just create a simple custom event with {FADE_TO_BLACK} and :fire it in-game:_
`	[event]
		name=fade
		{FADE_TO_BLACK}
	[/event]`

In my opinion the more "immersive" way to perform such visual transitions would be to make them affect everything players see on a map (including units, labels, etc.). Then it would look like this:
![fadetoblack-proposal](https://user-images.githubusercontent.com/35965303/171998213-51bfd096-1706-410a-824a-c8b048b038d7.gif)
(my proposal - notice the units and their hp bars are also slowly fading to black - final fade result: the perfectly black screen)
The example above could work great for a final scene of a campaign right before showing the Credits screen.

Also, this could be really useful in case of special effects like powerful spells illuminating the entire map. I am aware there are macros like FLASH_WHITE etc. but they are based on [color_adjust] which does not affect units.
An improved fade to white effect, combined with additive blend mode (if supported) could look like this:
![fadetowhite-additive](https://user-images.githubusercontent.com/35965303/172000349-c0d2457f-063c-4eec-aa91-266ff7c1d676.gif)

This kind of improved transitions could be handled by a new [screen_fade] tag supporting keys like: type (fade_in, fade_out, instant), delay (ms), red, green, blue (0-255), affect_units (yes/no), blending_mode (multiply, add), interpolation (linear, cubic, sinusoidal),  etc. for complete control. In this solution the behavior of the [color_adjust] tag would not be changed at all and the two tags could co-exist with each other.

I do realize this is a small low-priority cosmetic proposal, but still it would be really nice to have, especially that mainline campaigns and UMC often use fade effects for things like cutscenes. Improving those effects could make such cutscenes look much better in my opinion. Besides, the vast majority of games I know, use this kind of complete fading so why not have this in BfW as well?

(Note: If such feature is already present and achievable by using lua or other methods, please let me know. I just might not be aware of it.)

[...]

PR #6864, opened 2022-07-13
Add new [screen_fade] WML action

mesilliac:
This is an implementation of FR #6745 . As mentioned there it is a new interface action WML tag `[screen_fade]`, which can fade the game display to a given colour over a given duration.

It has the following format in WML:
```
[screen_fade]
	duration=(ms) #if 0 the effect is instantaneous
	red,green,blue=(0-255,0-255,0-255)
	alpha=(0-255)
[/screen_fade]
```

The fade is persistent, so after fading to a colour that colour is kept as an overlay on the game map until a fade to 0 alpha is performed.

I already implemented fading in #6855 so this is just the WML interface.

Unlike the `[color_adjust]` tag, this one also covers units and halos. And also unlike `color_adjust`, `screen_fade` is hardware accelerated.

I also added a flashy event to the test scenario to make sure that it works.


data/schema/core/actionwml.cfg
```diff
@@ -865,6 +865,15 @@
 		max=infinite
 		{COLOR_KEYS s_int}
 	[/tag]
+	[tag]
+		name="screen_fade"
+		max=infinite
+		{REQUIRED_KEY red s_int}
+		{REQUIRED_KEY green s_int}
+		{REQUIRED_KEY blue s_int}
+		{REQUIRED_KEY alpha s_int}
```
CelticMinstrel, line 874: I'd suggest making these keys optional (probably defaulting to 0, except for alpha which should maybe default to opaque, though I'm not quite sure what it even does here so I could be wrong).

src/scripting/game_lua_kernel.cpp
```diff
@@ -3724,6 +3724,19 @@ int game_lua_kernel::intf_get_color_adjust(lua_State *L)
...
 
+int game_lua_kernel::intf_screen_fade(lua_State *L)
+{
+	if(game_display_) {
+		color_t fade;
+		fade.r = luaL_checkinteger(L, 1);
+		fade.g = luaL_checkinteger(L, 2);
+		fade.b = luaL_checkinteger(L, 3);
+		fade.a = luaL_checkinteger(L, 4);
+		game_display_->fade_to(fade, luaL_checkinteger(L, 5));
```
CelticMinstrel, line 3735: This API feels quite awkward… 5 integer arguments do not exactly make for an easy-to-understand API.

I think perhaps taking the colour as a table would be easier, so the function only has two arguments. The only thing I'm not sure of is whether the table should be an array-like table (`{r, g, b, a}`) or a key-value table (`{red=r, green=g, blue=b}`).

Also, you need to add a stub with documentation in utils/emmylua/interface.lua.

Pentarctagon: Generally I prefer the latter, since then it's clear what the values in the table represent.

CelticMinstrel: I actually kinda like accepting both in this kind of scenario; if I were making an API that _returned_ a colour I'd probably use the `named_tuple` constructor for the result.

For an array-like table, I'd expect it to handle a missing alpha value but not the other components. For a key-value table… I think I'd expect it to handle missing any component.

data/schema/core/actionwml.cfg
mesilliac: For alpha, 255 fully applies the fade, so the screen will be that colour, and 0 full unapplies the fade, so the screen will be normal. Essentially it's the opacity of the fade surface that covers the display. I'm not sure which would make more sense as a default.

For the colours a 0 default would make sense.

src/scripting/game_lua_kernel.cpp
mesilliac: That would be better. I'm not sure how to do it as i haven't used lua or touched the WML interface before, so i just hacked in something that worked.

CelticMinstrel: If you're going for a key-value table, I think we have something like `luaW_table_get_default` which should help. For an array table there's `lua_check<std::vector<int>>`.

mesilliac: Sorry it commented without showing me the other comments. My only opinion is that using some sort of colour argument would be better than separating the integers, but i'm not confident doing the implementation of any of it. Help is needed here.

data/lua/wml-tags.lua
```diff
@@ -718,6 +718,11 @@ function wml_actions.color_adjust(cfg)
 	wesnoth.interface.color_adjust(cfg.red or 0, cfg.green or 0, cfg.blue or 0)
 end
 
+function wml_actions.screen_fade(cfg)
+	color = {cfg.red or 0, cfg.green or 0, cfg.blue or 0, cfg.alpha}
```
CelticMinstrel, line 722:
```suggestion
	local color = {cfg.red or 0, cfg.green or 0, cfg.blue or 0, cfg.alpha}
```

src/scripting/game_lua_kernel.cpp
```diff
@@ -3730,6 +3730,19 @@ int game_lua_kernel::intf_get_color_adjust(lua_State *L)
 	return 0;
 }
 
+int game_lua_kernel::intf_screen_fade(lua_State *L)
+{
+	if(game_display_) {
+		auto vec = lua_check<std::vector<uint8_t>>(L, 1);
+		if(vec.size() != 4) {
+			return luaL_error(L, "screen fade colour must be an array of 4 integers");
```
CelticMinstrel, line 3738:
```suggestion
			return luaW_type_error(L, 1, "array of 4 integers");
```

This makes the error be something like:

`bad argument #1 to 'screen_fade' (array of 4 integers expected, got nil)`

CelticMinstrel (changes requested):
Looks good. I just have two additional changes to propose, see below.

mesilliac:
Hmm, i applied all the suggested changes but github still says changes are requested. Aside from github's strange hangups, i think this should be ready for squash and merge...

CelticMinstrel (approved):

CelticMinstrel:
GitHub requires approval to be set manually.

Issue #1146, opened 2017-05-08
Glitch when a submerged unit stands on an [item] (GNA #13041)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?13041)** by zookeeper on 2009-02-21

When a unit stands in for example water and there's an [item] on that hex, the item image will show through the half-transparent submerged part of the unit. Instead, the item should obviously always remain behind the unit regardless of submersion.

Easy to test in Bay of Pearls in HttT: just take a mermen to the storm trident and don't pick it up.

(Reproduced on All)
Release: trunk
Priority: 3 - Low
Severity: 2 - Minor

wesnoth-bugs:
**Modified on 2009-12-28**

**boucman wrote:**

> this one is very very tricky to fix
> 
> 
> 
> the problem is that units in water are "half transparent", so we need a way to tell the item how much of itself should be transparent, which breaks the current design badly
> 
> 
> 
> no idea how to fix that one...

wesnoth-bugs:
**Modified on 2015-08-08**

**wedge009 wrote:**

> I confirm this issue is still around in 1.13.1+dev, but honestly I don't think it's that bad. Even when I first started playing Wesnoth back in 1.4.x and was playing Heir to the Throne, this glitch didn't stick out in my mind at all. True, it might be ideal to have the item hidden behind the unit in question... but on the other hand, I find it's probably useful to have the item still visible on the off chance one chooses not to pick up the item with the first unit that contacts it.
> 
> 
> 
> Given the technical effort it would take to resolve this, I think this one can be left alone. Would be nice to fix, but not really essential.

wesnoth-bugs:
**Modified on 2016-11-07**

**wedge009 changed status:** None -> Confirmed

wesnoth-bugs:
**Modified on 2016-11-07**

**wedge009 changed OS:** win xp -> All

knyghtmare:
Seems to be resolved with https://github.com/wesnoth/wesnoth/pull/6855.

Wedge009:
Hmm, does this still count as 'showing through'?
![item](https://user-images.githubusercontent.com/13679322/178650331-d6729fc9-1055-4265-83db-b2e68098ee52.png)

knyghtmare:
Seems item by item basis bug. I tested with "rock-submerged" in Dead Water.
[...]

PR #6871, opened 2022-07-14
Submerge items on water, just like units

mesilliac:
This is addressing #1146.

As per the comments in #1146 I made a change to simply use the same submerge code as units, for items. So when an item is placed on water it partially submerges. This fixes the visual appearance when a unit is swimming on top of the item, and also IMO makes items in water look a whole lot better.

There are one or two situations in which the item probably should not be submerged. It can be seen below, with the target circle. This could be fixed by using "halo" in stead of "image" in the item definition, so it draws over units in stead of under them.

It might make sense to have a separate "floating" item type, like say for debris that is actually floating on the water. There would need to be an implementation to split the item into two halves, one drawn in front of the unit and one drawn behind, as for unit ellipses. However unless there's an actual use case where this is required i don't want to speculatively implement it.

Here's the new look:
submerge

And the old look for comparison:
nosubmerge

### DIFF
--- a/data/campaigns/Heir_To_The_Throne/scenarios/04_The_Bay_of_Pearls.cfg
+++ b/data/campaigns/Heir_To_The_Throne/scenarios/04_The_Bay_of_Pearls.cfg
@@ -180,20 +180,20 @@
         {OBJ_TRIDENT_STORM 5 4 bop_stormtrident}
 
         #when certain villages are entered, Merfolk are rescued
-        {PLACE_IMAGE "units/merfolk/fighter.png~RC(magenta>red)" 5 38}
-        {PLACE_IMAGE items/cage.png 5 38}
+        {PLACE_IMAGE_SUBMERGED "units/merfolk/fighter.png~RC(magenta>red)" 5 38}
+        {PLACE_IMAGE_SUBMERGED items/cage.png 5 38}
 
-        {PLACE_IMAGE "units/merfolk/fighter.png~RC(magenta>red)" 7 35}
-        {PLACE_IMAGE items/cage.png 7 35}
+        {PLACE_IMAGE_SUBMERGED "units/merfolk/fighter.png~RC(magenta>red)" 7 35}
+        {PLACE_IMAGE_SUBMERGED items/cage.png 7 35}
 
-        {PLACE_IMAGE "units/merfolk/fighter.png~RC(magenta>red)" 11 33}
-        {PLACE_IMAGE items/cage.png 11 33}
+        {PLACE_IMAGE_SUBMERGED "units/merfolk/fighter.png~RC(magenta>red)" 11 33}
+        {PLACE_IMAGE_SUBMERGED items/cage.png 11 33}
 
-        {PLACE_IMAGE "units/merfolk/fighter.png~RC(magenta>red)" 19 23}
-        {PLACE_IMAGE items/cage.png 19 23}
+        {PLACE_IMAGE_SUBMERGED "units/merfolk/fighter.png~RC(magenta>red)" 19 23}
+        {PLACE_IMAGE_SUBMERGED items/cage.png 19 23}
 
-        {PLACE_IMAGE "units/merfolk/fighter.png~RC(magenta>red)" 3 10}
-        {PLACE_IMAGE items/cage.png 3 10}
+        {PLACE_IMAGE_SUBMERGED "units/merfolk/fighter.png~RC(magenta>red)" 3 10}
+        {PLACE_IMAGE_SUBMERGED items/cage.png 3 10}
     [/event]
 
     [event]
--- a/data/core/editor/items.cfg
+++ b/data/core/editor/items.cfg
@@ -10,204 +10,238 @@
         id=altar_evil
         image=items/altar-evil.png
         name= _ "Evil Altar"
+        submerge=1
     [/item]
 
     [item]
         id=altar
         image=items/altar.png
         name= _ "Altar"
+        submerge=1
     [/item]
 
     [item]
         id=ankh-necklace
         image=items/ankh-necklace.png
         name= _ "Ankh Necklace"
+        submerge=1
     [/item]
 
     [item]
         id=ankh-necklace2
         image=items/ankh-necklace2.png
         name= _ "Ankh Necklace"
+        submerge=1
     [/item]
 
     [item]
         id=anvil
         image=items/anvil.png
         name= _ "Anvil"
+        submerge=1
     [/item]
 
     [item]
         id=archery-target
         image=items/archery-target-right.png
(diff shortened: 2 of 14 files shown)


CelticMinstrel:
Can we see some of the items that are specifically intended for use on water, such as the shipwrecks?

mesilliac:
Shipwrecks still seem to look okay, but could use some revisiting with the effect.
shipwrecks

mesilliac:
Of course... with the boat that's *not* supposed to be sunk, there is a problem
missunken

But then when regular, boarders swim inside the boat.
missunnkenswim

In this case the unit should be standing on the boat... some way of saying "don't swim here" would probably make the most sense. And would also fix the boat.

mesilliac:
Some other comparisons from the bay of pearls HttT mission
sea-cage

trident-away

and the original issue
trident-on

CelticMinstrel:
> Of course... with the boat that's _not_ supposed to be sunk, there is a problem

Is that boat in the items folder…? I thought it was only a unit.

CelticMinstrel:
Maybe the solution is to actually use the boat unit, instead of an item. It could be placed on an allied side with the same team colour to prevent the player from having control of it, and units already have a way to disable submerge.

Unless adding a way to disable submerge on a per-item basis is actually really easy?

mesilliac:
> Unless adding a way to disable submerge on a per-item basis is actually really easy?

It's not super easy, but it's not too hard either. I don't mind doing it if that's the best solution.

Pentarctagon:
@CelticMinstrel did you mean to request review from 5 people?

CelticMinstrel:
@Pentarctagon I want some input from people who actually work on add-ons, and those 3 were the first I thought of. I added you and Vultraz just because I thought you might have an opinion too.

CelticMinstrel:
How does the whirlpool look with this change?

doofus-01:
It looks better on some things, but not all.  It makes everything look like it's a standing up cardboard cutout - things like the books and flowers would look better completely submerged, or not submerged at all and adjusted with IPFs if needed.  I'd say some sort of key to toggle or adjust this is needed.

mesilliac:
hmm, whirpools huh. That's something that should definitely be on the surface.
whirlpool-submerged

i did a quick comparison of three methods with a unit swimming there (left-to-right): (1) with the item submerge as above, (2) with an override to treat it normally (this should be the same as currently), and (3) splitting it into a back and front part like unit ellipses
whirlpool-swim

The weirdness with the submerged portion of the unit drawing over the item looks better than the cutout effect to me in all 4 cases. The whirpool image i guess is designed to take advantage of it, as it has no transparency itself, which becomes obvious when the front part is drawn in front of the unit.

The cutout also seems to be too high by 1 pixel... well that could be fixed if i did it properly.

mesilliac:
> It makes everything look like it's a standing up cardboard cutout

Yeah, it's the same for the effect on units, just less obvious.

I had some ideas of fancy effects using heightmaps that would look good, but they're not doable in plain SDL without shaders.

>  I'd say some sort of key to toggle or adjust this is needed.

I could put in a multiplier for the submerge value, like `submerge=0.0` doesn't submerge at all, `submerge=1.0` does the same as unit submerge, `submerge=2.0` submerges to twice the height where the unit submerges. `2.5` would be fully submerged in deep water as that has a submerge value of 0.4.

It might also require shifting the item placement so the waterline matches but i think that would be doable.

doofus-01:
Something like that should work for most situations, I think.  

At the end of the day, if the submerge effect just isn't working, but we can disable it in favor of IPF masks or blue/alpha-shifts, no harm done.

mesilliac:
Hmm, so correct me if i'm misunderstanding:

* items in scenarios are always fully specified, there is no item "template" that they reference by name
* the map editor has template items which it places into the scenario file as a full distinct `[item]` entry for each use

If this is how it works, then it won't be possible to make a retroactive decision on whether things should be submerged.

As such i think this feature should be disabled by default (`submerge=0`), and the item templates for the map editor should be edited to have sensible submerge values per-item. Items in existing scenarios such as Bay of Pearls can be modified manually as appropriate.

Does this sound like the right way to do it?

doofus-01:
That sounds right to me.

If the scenario editor had some editable text box, so you could change what would get written when "painted", it probably would address the problem with retroactive decisions, but that would be a whole different project.

mesilliac:
Alright i'll implement this when i have some time. It shouldn't be much change to what i already have here.

mesilliac:
So i added a multiplier for the submerge value, however, there is a bug when blitting transparent images onto other transparent images. So i was unable to adjust the position of the item to match the waterline for swimming units.

Well, i put the code in that does it anyway. But i have not used it for anything. In stead everything uses either 0 (no submerge) or 1 (terrain submerge value, same as units).

I changed the pickuppable item macros (used for things like storm tridents and health potions) to use submerge=1. This fixes the original issue.

I also added a new macro `PLACE_IMAGE_SUBMERGED` which can be used in place of `PLACE_IMAGE` for items on water that should submerge. I updated the merfolk cages in HttT:Bay of Pearls to use this new macro.

I also added submerge=1 to most of the physical items in the map editor. I passed over ones where the effect makes no sense (like runes, whirlpools, shipwrecks) and also some where the effect just looks very bad (like wells). I didn't put much effort into this, so it could still maybe use some tweaks.

If/when that blitting bug is fixed, submerge values other then 0 and 1 should also work.

Technically the waterline shift could be disabled, but then it wouldn't match swimming units and would just lead to a reoccurance of #1146 in different situations.

I didn't change any scenarios other than Bay of Pearls. There are probably other scenarios that could use the new macro, or `submerge=1` values added.

mesilliac:
The trident in BoP is now a bit more difficult to spot. I'm not sure if this is a problem. I guess usually in a game like this i'd expect there to be an occasional glint effect that drew attention to the item.

CelticMinstrel:
> i'd expect there to be an occasional glint effect that drew attention to the item.

I wonder how hard this is to actually do. I guess just am animated halo added on the item could produce that effect, right? So it could be pretty easy if we just had the assets.

mesilliac:
Yeah i think an animated halo would work just fine. It doesn't look like there's any problem with having both an image and a halo.

mesilliac:
I think this is okay as it is now. Unless there are any objections i'll merge this how i have it.

The code for non-integer submerge factors is in-place using IPF, but it won't look good unless that transparent-blit problem is solved, and i haven't set anything to use a submerge factor other than 0 or 1. 0 is still the default, so nothing in existing campaigns will submerge unless edited to do so like with Bay of Pearls and the test scenario (which i just edited).

New items placed via the editor will submerge or not according to the item.

PR #6877, opened 2022-07-16
preferences: Adjust show_deprecation default to coincide with usage

mesilliac:
If unset it has different behaviour in dev branch when checked (defaulting to yes). However this was not reflected in the actual prefs dialog which clearly said that it was "no".

This is a bit ugly... but well the default is set statically in `data/advanced_preferences.cfg`, so...

Also usage of bool preferences and how the default values work is horribly wrong everywhere, as the default value must be given by the code that tries to access the value. But they don't know the default value. So they make something up. Which leads to situations such as this one where it does not match the actual default value and wrong shit happens.

### DIFF
--- a/src/preferences/advanced.cpp
+++ b/src/preferences/advanced.cpp
@@ -15,6 +15,7 @@
 #include "preferences/advanced.hpp"
 
 #include "game_config_view.hpp"
+#include "game_version.hpp"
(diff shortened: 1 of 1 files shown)


stevecotton:
Looks good to me, should probably be reviewed by @Vultraz too.

Issue #6589, opened 2022-03-27
Having x/y offset applied to the frame is causing them to not render properly while zoomed in or out


Hejnewar:
### Game and System Information

- **Version: 1.16.2**
- **Downloaded from: Steam**
- **OS: Windows 10**

### Description of the bug

Having x/y offset applied to the frame is causing them to not render properly while zoomed in or out. See additional context.

### Steps to reproduce the behavior

1. Launch To Land Unknown
2. Zoom in or out

### Additional context

https://user-images.githubusercontent.com/41817546/160295739-4c710e00-96de-44f6-848b-fd14dee8bf52.mp4

doofus-01:
I can't see the video, what's not rendering properly?  Mainline units with y= (Sky Hunter, for example) seem to be OK, at least in 1.17.  There was an issue with x,y= keys in animations a while back, and it was fixed in #5548, but there could be other, related cases.

Hejnewar:
Its really hard for me to describe it but after graphic moves to another location it still remains where it was, unless overwritten by something (another animation, mousing over it, etc). 

![like this](https://user-images.githubusercontent.com/41817546/160298665-7242c6d1-4764-4b2a-946c-7f5851dc62f3.JPG)

doofus-01:
I've seen something like that, I think it had to do with x/y or an out-of-hex frame in the standing frame or default image, and then no explicitly specified movement animation.  In other words, the auto-generated movement animation behaves strangely.  Does the TLU unit use a custom movement animation, or the auto/default?

Hejnewar:
TLU is using that. I actually found this bug out by trying to do similar thing and TLU is just way better example of the same thing happening. 
```
        [standing_anim]
            alpha=0
            [anim_frame]
                duration=3200
                image=units/summoners-humans/carpetrider1.png
                image_mod=~SCALE(36,36)~TC(1,magenta)
                alpha=0.8
                offset=-7.0~11.0
                y=54
                auto_vflip=false
            [/anim_frame]
        [standing_anim]
```

doofus-01:
[standing_anim] without [movement_anim] sounds like the issue, in that case.  Possibly it is an engine bug, but I think an argument could be made that the engine can't know what is wanted for units with complicated standing animations.  In any case, if you explicitly write out the [movement_anim], does the issue persist?

Hejnewar:
```
    [movement_anim]
        direction=s,se,sw
        start_time=0
        [frame]
[...]

Issue #6118, opened 2021-09-27
Inconsistent cropping of images when they are larger than the hex they're on


max-torch:
### Game and System Information
 - Game version is 1.15.18
 - Steam download
 - Windows 10

### Describe the bug
When an image is placed on a hex and then scaled and cropped, the cropping of the images is inconsistent.

### To Reproduce
One way to reproduce the behavior:

1. Load the scenario SotA S20 North Knalga
2. Zoom in on the dwarf statues so that the images become larger than the hexes
3. Try scrolling back and forth around the map, removing and returning the statues from view.
4. Move the cursor over the statues
5. Repeat steps 3 and 4 to verify.

### Expected behavior
There should be no glitchy behavior. Parts of images should not reappear/disappear while you are playing and scrolling and zooming.

### Screenshots
Visual Evidence
![imagecrop5](https://user-images.githubusercontent.com/41433185/134890546-6f087341-00d5-426f-9cb8-4213b7336d63.gif)
![imagecrop2](https://user-images.githubusercontent.com/41433185/134890670-fbd8908c-9d94-45e2-9fea-b519e1de1542.gif)

### Savefiles
[SotA-North Knalga replay 20210925-131532.gz](https://github.com/wesnoth/wesnoth/files/7235452/SotA-North.Knalga.replay.20210925-131532.gz)

### Additional context
The WML code which places, scales, and crops the images in the example is [here](https://github.com/wesnoth/wesnoth/blob/ca90694141d8e2c5d2d1ea8afcfb071e9a439f12/data/campaigns/Secrets_of_the_Ancients/scenarios/20_North_Knalga.cfg#L233-L265)

AI0867:
Rather than a cropping issue, this could also be a z-order issue, that is, the game is unsure whether the [item] image should be above or below the neighbouring hex.

I suspect it's that the game does not account for an `[item]` being larger than the hex it's on, so redrawing the overlapping hex does not mark the `[item]` as dirty, and it's not redrawn over the hex.

Wedge009:
The scrolling of the first animation makes it hard to see what's going on, but at least the second example is clearer.

cooljeanius:
oh yeah I think I've noticed this with some unique villages in ANO, too

max-torch:
> The scrolling of the first animation makes it hard to see what's going on, but at least the second example is clearer.

In the first animation the scrolling is there on purpose:
[...]

PR #6878, opened 2022-07-16
1.17 (test scenario) out-of-hex unit animation tests

doofus-01:
This is for testing #6589, as well as looking at how ellipse and HP bar move with sprite in progressive y= key

### DIFF
--- a/data/scenario-test.cfg
+++ b/data/scenario-test.cfg
@@ -980,12 +980,9 @@ Adjacent own units of lower level will do more damage in battle. When a unit adj
             jamming=5
         [/unit]
         [unit]
-            # for testing out-of-hex graphical bugs
             x,y=10,7
             type="Orcish Archer"
-            # generate_name=yes
-            name="Animation Tester"
-            id=anim_tester_ooh
+            generate_name=yes
         [/unit]
         [unit]
             x,y=18,12
@@ -1160,6 +1157,16 @@ Adjacent own units of lower level will do more damage in battle. When a unit adj
         fog=yes
         shroud=yes
         share_vision=none
+        [unit]
+            # for testing out-of-hex graphical bugs
+            x,y=11,9
+            type="Orcish Archer"
+            # generate_name=yes
+            name=_ "Animation Tester"
+            id=anim_tester_ooh
+            max_moves=0
+            random_traits=no
+        [/unit]
     [/side]
     [event]
         name=turn refresh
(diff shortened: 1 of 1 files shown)


mesilliac:
I thought i couldn't get it to reproduce even with this, but now i do see some artifacts when zooming in as much as possible.

I found regular unit movement from hex-to-hex has this problem as well, just it's not normally noticable as all surrounding hexes are cleared when the unit arrives at the next hex.

I haven't found the fix yet but will keep poking at it. I guess it doesn't happen when zoomed out because the default size ends up clearing more than necessary, not less.

mesilliac:
I think i found the culprit. There was some code that wasn't accounting for zoom. I'll tidy it up and see about pushing a fix tomorrow.

mesilliac:
I committed a fix to master which fixes all the drawing problems for me at all zoom levels.

This test case should probably be merged too, but for a couple issues:
* i'm not sure what the CI failure here is about
* the map label overlaps the one to the left, it should maybe be placed on a different hex?
* the changed text colour is nice but nothing else does that so it looks a bit weird :sweat_smile:

Well those last two issues aren't all that critical.

CelticMinstrel:
The test failure is just that unit names need to be translatable to pass schema validation.

doofus-01:
Thanks for fixing the issue.  The different color text was a lazy fix for the overlapping.  There should probably also be some way to turn that animation off, without just ending the test.

I didn't know what the CI failures were about either, `No value for $TERM and no -T specified` is not for mere mortals to understand.

doofus-01:
I believe this is done now, but I noticed something that's (I think) unrelated.

This elf has what I think is the handle of a sword from the wesnoth icon blinking as the NW neighbor hex gets redrawn.  This must be some deliberate test, I guess it's failing?
```
        [unit]
            x,y=4,8
            type="Elvish Avenger"
            id=statue
            generate_name=yes
            hitpoints=4
            vision=15
            ellipse=none
            [status]
                petrified=yes
            [/status]
            overlays="wesnoth-icon.png~CROP(10,10,20,20)"
        [/unit]
```

mesilliac:
I think that might be to test the out-of-hex part of the overlay. It has the same symptoms as #6118 when you mouse over it.

All of the code related to "overlays" seems rather buggy. I can only assume sometime in the past it was all broken by a rewrite, because as the code is now, things like this could never have worked.

mesilliac (approved):

mesilliac:
I would merge this but i'm not sure whether to squash/rebase/merge, feel free to click the button :)

CelticMinstrel:
> I didn't know what the CI failures were about either, `No value for $TERM and no -T specified` is not for mere mortals to understand.

Just for the record, that was not the error message that allowed me to determine the issue. You can just ignore that message; the real error was further down in the log.

Issue #6825, opened 2022-07-02
Ghost-line sprites need review


mesilliac:
I was testing something i did and noticed that ghosts lost their transparency. Turns out it isn't my fault.
![ghost-1 17](https://user-images.githubusercontent.com/243646/176994297-8d1c990d-8bc1-478f-bfd3-d67b60579635.png)
![ghost-1 16](https://user-images.githubusercontent.com/243646/176994306-23da886a-8083-455f-9a65-1500906682b6.png)
Top is ghost in current master branch, bottom is ghost in 1.16 branch.

Ignoring the shadow, which seems to have been split out, the current ghost is missing the transparent fade-out effect that the 1.16 ghost had, where the arm and the edges of the cloak are translucent. It's especially noticable for the little wispy bits of cloak around the edges, and the trailing hand.

doofus-01:
I did this, mostly with a script and then quick manual cleanup, so there may be an issue, but the removal of translucent bits was intentional and is itself a "won't fix". 

The shadow does need to be (re)animated, but it was a lower priority than other things.

This was part of an effort to make the sprites really be pixel art.  Semitransparent pixels aren't needed; see the spectre, which was a more recent work by Jetrel.  The ghost sprite might need more editing if the change is too noticeable, but I'm not immediately seeing anything broken.

mesilliac:
Hmm. It's definitely very noticable to me. The previous ghost looked very good. This ghost, not so much...

doofus-01:
Cool.  Let me know if you have something I can work with.  If not, this should probably be closed.

Wedge009:
Any difference is very subtle. I wouldn't even notice the missing shadow (which actually makes sense for an ephemeral unit like the ghost) if you hadn't pointed it out. I looked at 1.16.3 release and master at 159d5286bd2, side-by-side. It sounds like the differences may well be intentional.

mesilliac:
It's only noticable at all if you disable standing animations. Also the shadow does seem to appear when the standing animation is playing, it's only missing in the still frame.

I would say the 75% opacity on the previously-translucent areas (specifically the back arm and back-flowing cape) should be retained. For me this is a core aspect of making the ghost appear ghostly, blending slightly into the background. I agree that generally pixel art shouldn't use alpha blending, but after looking closely there seem to be two main regions in the old sprite, one with full opacity and one with 75% opacity.

[...]

PR #6881, opened 2022-07-17
1.17 Ghost-line animations

doofus-01:
This is to address #6825, most work going into finishing the nightgaunt.


knyghtmare:
Is it intentional that the Nightgaunt's "face" is more roundish now?

PR #6882, opened 2022-07-18
Automatically append std::endl to all log messages

mesilliac:
There is no reason not to do this for general logging functions. They print names and timestamps, so cannot be sensibly concatenated without newlines.

I added a `STREAMING_LOG` replacement for `PLAIN_LOG` in the few instances relying on piecemeal output. It doesn't append the newline.

Everything else had its newline taken off, mostly via overly complex sed command.

Surprisingly this only touched a little over half the files in the project...


stevecotton:
`std::endl` isn't just a newline, it forces a flush too. As long as Wesnoth doesn't crash afterwards, I think that might be unnecessary overhead.
https://en.cppreference.com/w/cpp/io/manip/endl

stevecotton:
BTW, @mesilliac and @Pentarctagon, it seems you're both looking at this (@Pentarctagon has a work-in-progress in his `master` branch).

mesilliac:
> std::endl isn't just a newline, it forces a flush too. As long as Wesnoth doesn't crash afterwards, I think that might be unnecessary overhead.

That is exactly why i think making it `std::endl` is important. If it crashes you need those logs, and that's what the logging is for, is it not?

I have had to debug situations in the past where the necessary logging info was not provided because it was not being flushed to output before crashing. It was not pleasant.

CelticMinstrel:
Whether it's this version or Pentarctagon's, I do think this is a good idea.

Pentarctagon:
Heh, just a day or two more and I would've opened a very similar PR.

@mesilliac there are also a small number of places that are using `std::flush` which can also be removed, rather than newline or `std::endl`.

Pentarctagon (approved):

stevecotton:
I'm happy with this.

mesilliac:
Rebased, picked up some lines ending with std::flush (:+1:) and picked up some lines with trailing newlines preceded by whitespace that weren't caught by my previous regex. Also picked up an extra line from the events PR that wasn't caught in the rebase.

I'll aim to merge this as soon as 1.17.6 gets tagged to minimize divergence.

PR #6887, opened 2022-07-20
Some small improvements and fixes

mesilliac:
These are just some easy-to-pull-out fixes and improvements to base tools that i'm pulling out of my video system tidyup.

* Added some additional convenience methods on sdl/point, sdl/rect, sdl/window
* Added a single-white-pixel PNG. This is useful for drawing filled rectangles by blitting it as a texture with colour and alpha mod.
* Added an `update()` method to the top-level-drawable interface
* Improved how the draw-manager handles adding and removing TLDs, especially while iterating over them
* Allowed blit helpers to use chained method calls for setup - so you can now call `drawing_buffer_add(...).set_alpha(200).set_color(c)` etc.
* Fixed sdl/texture to properly set blend mode when creating from surface
* Fixed the `--render-image` comandline option to run without GUI and also work when --nogui is used.

For the draw manager / TLD change, i also documented a lot of how the process is supposed to work.
In summary all top-level-drawable objects (e.g. windows, tooltips, dialogs) have a chance to execute a function in each of four distinct phases every vsync frame. They are, in order:
1. `update()` - change variables, process animations, update anything that might affect layout
2. `layout()` - ensure that positioning information is up-to-date
3. `render()` - perform any offscreen rendering necessary in preparation for draw
4. `expose()` - draw to the screen

`layout()` and `expose()` are mandatory to implement. `update()` and `render()` are optional.

It already worked like this, i just added the `update()` stage.


CelticMinstrel:
> This is useful for drawing filled rectangles by blitting it as a texture with colour and alpha mod.

Huh? Isn't there already a routine for drawing a filled rectangle? Why is a texture needed for this?

mesilliac:
> Huh? Isn't there already a routine for drawing a filled rectangle? Why is a texture needed for this?

It's for the draw buffering system used by the terrain-drawing system. It queues up all the terrain, unit, etc. drawing calls, then sorts them, then draws them in order, and it doesn't support the rectangle drawing functions etc. only texture blits. It was easier to do the single-pixel-blit trick than to hack it to also perform other drawing functions.

In particular i'm using it for drawing the HP/XP bars, and the background for terrain labels. Those are the last two sticking points for hardware-accelerated game map drawing.

Actually i'll pull the terrain label one in as well, that's easily separable.

> Also, no inset/outset/whatever you want to call it for rects?

I considered adding them too, i just haven't touched anything that would use them yet. I'm generally only adding functions when i need to use them for something.

mesilliac:
pulled in the terrain-code text background thing. Previously it was creating a new SDL surface then translating it to texture individually for each one, just to draw the background rectangle behind terrain codes in the editor.

CelticMinstrel:
Huh… terrain codes aren't drawn with floating labels? I wonder if there's a reason for that…

mesilliac:
No idea... but that's how it is. This pixel trick is actually very efficient and robust for drawing rectangles and lines, so it's good to have available in any case.

PR #6889, opened 2022-07-20
update codeblock for gcc11

newfrenchy83:


projectfiles/CodeBlocks/liblua.cbp
```diff
@@ -32,6 +32,8 @@
 		<Compiler>
 			<Add option="-Wall" />
 			<Add option="-std=c++17" />
+			<Add option="-D__GLIBCXX_TYPE_INT_N_0=__int128" />
+			<Add option="-D__GLIBCXX_BITSIZE_INT_N_0=128" />
```
CelticMinstrel, line 36: Why? This does not sound like something you'd want.

newfrenchy83: This is to avoid the warning message that appears when __STRICT_ANSI__ and __GLIBCXX_TYPE_INT_N_0 are not defined with gcc11 as compiler.

PR #6890, opened 2022-07-20
Added a variation to the Royal Guard portrait to fix an anachronism

### LINKED ISSUE #6431: [Question] [Art] About the royal guard portrait
 - 1.16

Just recently realised that the royal guard portrait has a scepter of fire logo in their chest and helmet, and that portrait is used in campaigns that predate the SOF's creation. I suggest making the current portrait a variation of a new "main" portrait. I open this as a question on an issue because I may be wrong or missing something.

### Screenshots
I have made a fast "fix" of the portrait, it is very rough so I would recommend to have someone who knows what they are doing fixing this if that is wanted at all. Also the variation (w/ the SOF) should be implemented via code in the future if this is followed through.

royal-guard
royal-guard-sof

Descacharrado:
Converted the image from the issue using optiwebp

The preprocessor nesting is not pretty, I know that, but I preferred that to modifying the unit line in every affected campaign because the change will be easier to find and edit in the future.

There are some problems that arise:

1. The LoW campaign extends beyond the SoF in the timeline, so the later half of the campaign (after the SoF is created) will have Royal Guards without a sceptre drawn over their armor, maybe thats ok.
2. Maybe all campaigns up to (and including) HttT should be included and not just up to SoF, as the sceptre is not found until then.
3. This creates image redundancy, instead of this maybe it would be preferrable to have overlays to a base image, although that is something that could be done in the future once/if we want to have something drawn over the units armor (e.g. a coat of arms).

Closes #6431


CelticMinstrel (changes requested):
This change is clearly incorrect.

1. Core files shouldn't be checking campaign-specific defines.
2. Any define other than those listed will now have no portrait at all.

I think the correct approach is probably to use a unit variation and have the relevant campaigns explicitly declare that they want that variation. Or you could even not modify the unit type definition at all and just have the relevant campaigns override the portrait for those units.

Descacharrado:
I dont mind changing it to portrait override per campaign, will do later, thanks.

doofus-01:
It would be better if something was done about the blank patches.

CelticMinstrel:
> Seems like overriding the unit's portrait would require running an event everytime a RG is created, so I'll finally opt for changing the definition.

?

Are royal guards recruitable? If so, then yeah, that's not the optimal solution… but otherwise you just need to add something extra to the `[unit]` tag. (Although, if that's hidden behind a macro, it might also be a little difficult.)

Descacharrado:
> It would be better if something was done about the blank patches.

Agreed

> Are royal guards recruitable?

Spearmen are recruitable in TRoW, the solution I now have in mind is running a global `advance` event that uses an [effect] to modify the unit's portrait. For now, I'll draft the PR until I feel like I have a good enough solution.

Descacharrado:
Closing because I did not find a good solution.

Issue #6891, opened 2022-07-20
Parts of Drake Flameheart getting clipped or changing z-level at max zoom when moving the cursor back and forth


max-torch:
### Game and System Information

- **Version:** 1.17.6
- **Downloaded from:** steam
- **OS:** Windows 10
- **Resolution:** 1920x1080

### Description of the bug

Let me show the gif first:
![drakebug2](https://user-images.githubusercontent.com/41433185/180054750-09cfb394-782a-4a06-b9e3-d088163df586.gif)

As you can see, at maximum zoom, I am moving the cursor back and forth from a Drake Flameheart to other hexes around it. Sometimes parts of the unit are clipped and sometimes the parts of the castle terrain are drawn over on top of its wings.
This recording was taken in the Scenario Bilheld from the campaign Dead Water.

### Steps to reproduce the behavior

1. Have a Drake Flameheart in the middle of castle tiles.
2. Go to the maximum zoom.
3. Move the mouse cursor back and forth from the Drake to two hexes away and then back. Try different angles.

### Expected behavior

No clipping or changing of z-levels of painted graphics.

Wedge009:
You reported something similar in #6118. Do you think this could be the same issue or something else entirely? Granted, this one is after the hardware acceleration was implemented.

mesilliac:
I can't reproduce this in my current working branch, so i think one of the changes i have on my local may have fixed it.

I'll be cleaning those up and posting them probably in the next few days.

I also have a fix relating to https://github.com/wesnoth/wesnoth/issues/6589 which solves out-of-hex leftovers during animation at zoom. I'll probably try putting that up today. It might be related.

max-torch:
> You reported something similar in #6118. Do you think this could be the same issue or something else entirely? Granted, this one is after the hardware acceleration was implemented.

I have no idea if it's related. All I know is that I decided to report this, because although it has some similarity to #6118, this is happening with the new hardware acceleration.

max-torch:
In 1.17.6,
I tried doing this again, but this time selecting the drake first, which @mesilliac said to try doing. See gif:
![drakebug3](https://user-images.githubusercontent.com/41433185/180138686-432b88ed-63d6-4ee7-baac-c8a0706c0e32.gif)
Everything behaves the way it is expected to behave when the unit is selected.

[...]

PR #6893, opened 2022-07-21
Several more invasive fixes arising from general tidyup

### LINKED ISSUE #6589: Having x/y offset applied to the frame is causing them to not render properly while zoomed in or out

mesilliac:
Notable items:
* tod_hex_mask is now hardware accelerated... whatever this is. There are no examples using it in mainline so i'm not sure it works correctly. But it should.
* HP/XP bars are now rendered in hardware, using the single-pixel image from my last PR.
* Unit animation invalidation code has been updated to properly account for zoom.

Re the HP/XP bars i ended up supporting the custom bar stuff, although it's confirmed to be unused by anything. Might still be better to do this by drawing rectangles, as it would give a better result when zoomed. I had an implementation for that too, but i discarded it as broken... turns out that was due to an unrelated bug.

Re the animation invalidation, this should address #6589 and maybe #6891... i confirmed it also fixed some leftovers from movement and attacking animations when zoomed in, so it might fix other bugs as well. This fix could probably be backported to 1.16

The rest of the stuff here is general cleanup.

### DIFF
--- a/src/controller_base.hpp
+++ b/src/controller_base.hpp
@@ -148,11 +148,6 @@ protected:
 	 */
 	void handle_event(const SDL_Event& event) override;
 
-	void handle_window_event(const SDL_Event& /*event*/) override
-	{
-		// No action by default
-	}
-
 	/** Process keydown (only when the general map display does not have focus). */
 	virtual void process_focus_keydown_event(const SDL_Event& /*event*/)
 	{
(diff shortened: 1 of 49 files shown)


mesilliac:
ffs i mention in the middle of a paragraph that it "should fix" something and github decides that means i want to auto-close it? Who came up with this system

PR #6894, opened 2022-07-21
Add --log-to-file command line option.

Pentarctagon:
This option sets up writing both the cerr and cout streams to a log file, as well as pulls the log rotation logic out of log_windows.*pp into the general logger.

Note that this is meant for usage on macOS and Linux, not Windows. log_windows.*pp has additional functionality in it, as well as a fair bit related to the --wconsole option.

---

Testing locally this seems to work fine, though the fact that this wasn't done years ago makes me wonder if there's something obvious that I'm missing.

### DIFF
--- a/src/commandline_options.cpp
+++ b/src/commandline_options.cpp
@@ -257,6 +257,7 @@ commandline_options::commandline_options(const std::vector<std::string>& args)
 		("log-debug", po::value<std::string>(), "sets the severity level of the specified log domain(s) to 'debug'. Similar to --log-error.")
 		("log-none", po::value<std::string>(), "sets the severity level of the specified log domain(s) to 'none'. Similar to --log-error.")
 		("log-precise", "shows the timestamps in log output with more precision.")
+		("log-to-file", "log output is written to a file rather than to standard error.")
 		;
 
 	po::options_description multiplayer_opts("Multiplayer options");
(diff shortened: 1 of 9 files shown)


CelticMinstrel:
Only thing I'm not sure of is whether it's okay to use the C++ names for these streams.

Pentarctagon:
As the command line values?

CelticMinstrel:
Yes.

Pentarctagon:
What you prefer they be named?

CelticMinstrel:
I'm not quite sure… `-` is commonly-used to represent standard in or out, but that doesn't work if you want a way to distinguish standard out from standard error. Another option is `stdout` / `stderr`. I don't know if there are other standard names for these streams.

Pentarctagon:
I could also remove the cout and cerr options. cerr is the default anyway, and cout I added because just because it was trivial.

Pentarctagon:
The default on Windows is to mostly do what `--log-to=file` does - redirect cout and cerr to a file. So I don't think setting `--log-to=cerr` or `--log-to=cout` would do anything.

src/commandline_options.cpp
```diff
@@ -257,6 +257,7 @@ commandline_options::commandline_options(const std::vector<std::string>& args)
 		("log-debug", po::value<std::string>(), "sets the severity level of the specified log domain(s) to 'debug'. Similar to --log-error.")
 		("log-none", po::value<std::string>(), "sets the severity level of the specified log domain(s) to 'none'. Similar to --log-error.")
 		("log-precise", "shows the timestamps in log output with more precision.")
+		("log-to-file", "log output is written to a file rather than to cerr.")
```
CelticMinstrel, line 260: Even here I'd say "standard error" or similar instead of "cerr".

PR #6901, opened 2022-07-23
1.17 drake sprite cleanup

doofus-01:
part of #6417


doofus-01:
I'm assuming that CI failure was a 3rd party data transfer issue, rather than something wrong here, based on the error and also since I didn't touch any config/code.

PR #6905, opened 2022-07-24
New alternative Saurian Skirmisher upgrade

czyzby:
This pull request adds the following saurian units to the core:

* Saurian Guard: Saurian Skirmisher alternative upgrade.
* Saurian Sentinel: Saurian Guard final upgrade.

Additionally, the core Drake factions are modified as follows:

* Saurian Guard can now be a leader of the default Drakes faction.
* Saurian Sentinel can now be a leader of the Age of Heroes Drakes faction.
* Saurian Guard can now be recruited by the Age of Heroes Drakes faction.

Points of concern that should be discussed before merging this PR:

**`1)`** Naming

If there's one thing I usually struggle with, that would be naming things. I'm not sure if Spearwoman→Warden are the best fitting names for the saurian warriors. Especially since there's a Drake Warden in the same faction.

@Jonathan-Kelly has suggested Spearthrower→Javelineer - I've actually considered those before, but since I'm not sure if "spearthrower" is a valid word in the first place, I decided to go with unit names that are less focused on the spear-throwing aspect and emphasize their defensive abilities instead.

As a side note, the descriptions of the new units might have to be updated or replaced.

**EDIT:** For now, it settled on Guard→Sentinel, but I'm still open to suggestions. Spearthrower→Javelineer might be used for a campaign unit with the legacy sprites (see #6912).

**`2)`** Sprites

I did my best, but these are still mostly recolored original sprites with some stuff thrown on top. I think I mostly followed the convention for saurian upgrades, but it has to be decided whether they fit the mainline standards in the first place.

**EDIT:** @doofus-01 has proposed a revised sprite for the guard unit. Might need some time to update the animation frames. The current WIP basic sprites can be seen here:

Base frames

**`3)`** Balancing

The unit statistics were discussed with and adjusted by @Hejnewar, and I've also done some playtesting vs AI locally. Although they seem fine on paper, they have to be looked into and approved.

**EDIT**: A change of the attack types was considered from pierce to blade or impact, along with a replacement of the spear weapon with a glaive.

**`4)`** Factions

The new alternative upgrade changes the Skirmisher unit tree quite significantly. Rather than just fragile skirmishers, saurian warriors would be able to be advanced to a more defensive variant with a better ranged attack and the first strike ability. This might change how the Drakes play against factions with common access to pierce attacks such as the Loyalists.

* Should Guards be available as leaders of the default Drakes faction? With 6 speed and lack of the skirmisher ability, I believe they conform to the core leader requirements. 
* Should Guards be available in the default era at all? As an alternative, they could be an optional upgrade for campaigns and add-ons, similarly to how the Dwarvish Arcanister tree is handled.


knyghtmare:
> @Jonathan-Kelly has suggested Spearthrower→Javelineer - I've actually considered those before, but since I'm not sure if "spearthrower" is a valid word in the first place, I decided to go with unit names that are less focused on the spear-throwing aspect and emphasize their defensive abilities instead. I'm open to suggestions.

The reasoning behind these suggestions are that these are the IDs being used in the upcoming campaign candidates: the Legend of Wesmere Reborn and the War for the North (the Northern Rebirth Rework). Thus, if kept `Saurian Spearthrower` and `Saurian Javelineer`, the campaigns wont need to be patched and can resume their development unhindered.

CelticMinstrel:
> Thus, if kept `Saurian Spearthrower` and `Saurian Javelineer`, the campaigns wont need to be patched and can resume their development unhindered.

If needed to support alternate IDs and smooth porting, we can always add a stub unit that redirects to the real one, something like:
```
[unit_type]
  id=old_id
  [base_unit]
    id=new_id
  [/base_unit]
  {DEPRECATED_UNIT old_id new_id 1.19}
[/unit_type]
```

It could just go in the same file as the real unit.

knyghtmare:
> > Thus, if kept `Saurian Spearthrower` and `Saurian Javelineer`, the campaigns wont need to be patched and can resume their development unhindered.
> 
> If needed to support alternate IDs and smooth porting, we can always add a stub unit that redirects to the real one, something like:
> 
> ```
> [unit_type]
>   id=old_id
>   [base_unit]
>     id=new_id
>   [/base_unit]
>   {DEPRECATED_UNIT old_id new_id 1.19}
> [/unit_type]
> ```
> 
> It could just go in the same file as the real unit.

I dont think WftN will be needing that since I done with the Saurians and Drakes chapter of it and am drafting the final 2 levels.

czyzby:
>> @Jonathan-Kelly has suggested Spearthrower→Javelineer - I've actually considered those before, but since I'm not sure if "spearthrower" is a valid word in the first place, I decided to go with unit names that are less focused on the spear-throwing aspect and emphasize their defensive abilities instead. I'm open to suggestions.
>
> The reasoning behind these suggestions are that these are the IDs being used in the upcoming campaign candidates: the Legend of Wesmere Reborn and the War for the North (the Northern Rebirth Rework). Thus, if kept Saurian Spearthrower and Saurian Javelineer, the campaigns wont need to be patched and can resume their development unhindered.

I see. I think we should still discuss the available options, and choose names that are the most appropriate for these units - especially if we want them to be available outside of campaigns. If a change will be necessary, I can submit PRs to campaign repos with the corrected unit references, or perhaps use the unit stubs.

"Javelineer" does sound nice, but I think perhaps we should reconsider "spearthrower" to something that appears in a dictionary. :)

data/multiplayer/factions/drakes-aoh.cfg
```diff
@@ -4,7 +4,7 @@
     name= _"Drakes"
     image="units/drakes/flameheart.png"
     type=random
-    leader=Drake Flameheart,Inferno Drake,Drake Enforcer,Drake Warden,Drake Blademaster,Saurian Soothsayer,Saurian Oracle
+    leader=Drake Flameheart,Inferno Drake,Drake Enforcer,Drake Warden,Drake Blademaster,Saurian Warden,Saurian Seer,Saurian Prophet
     random_leader=Drake Flameheart,Inferno Drake,Drake Enforcer,Drake Warden,Drake Blademaster
     recruit=Drake Burner,Fire Drake,Drake Clasher,Drake Glider,Drake Fighter,Saurian Skirmisher,Saurian Augur,Drake Warrior,Drake Flare,Drake Thrasher,Sky Drake,Drake Arbiter,Saurian Ambusher,Saurian Soothsayer,Saurian Oracle
```
knyghtmare: you have forgotten to add Saurian Spearwoman as a recruit here, as stated in starting post of this PR

knyghtmare (commented):
Just one requested change. Everything else looks okay to me.

data/multiplayer/factions/drakes-aoh.cfg
czyzby: I've already amended that, refresh the changes view.

knyghtmare (approved):
Unless there are other requested changes, I give green tick.

CelticMinstrel:
> I think perhaps we should reconsider "spearthrower" to something that appears in a dictionary.

Not all valid words will be found in a dictionary. I think "spearthrower" is just fine as a name in general. I'm not sure that it's the best name for this unit, though. Still, if no-one can think of a better name, I think it'll do.

doofus-01:
> If there's one thing I usually struggle with, that would be naming things. I think Seer and Prophet follow the convention of the saurian mages naming pretty well, both being basically a synonym for clairvoyant. However, I'm not sure if Spearwoman→Warden are the best fitting names for the saurian warriors.

>Jonathan-Kelly has suggested Spearthrower→Javelineer - I've actually considered those before, but since I'm not sure if "spearthrower" is a valid word in the first place, I decided to go with unit names that are less focused on the spear-throwing aspect and emphasize their defensive abilities instead. I'm open to suggestions.

Seer & Prophet do fit the existing convention, but I still wonder if something a bit more informative is possible, something like "High Oracle" (just for illustration, not a great candidate itself).  For the skirmisher alternate, there's already a Drake Warden, so replacing that would be good.

> 2) Sprites
I did my best, but these are still mostly recolored original sprites with some stuff thrown on top. I think I mostly followed the convention for saurian upgrades (e.g. larger "crowns" as the level goes up, golden-silver weapons at Lv3), but it has to be decided whether they fit the mainline standards in the first place.
Some graphical changes that might be considered before merging:
>>        Saurian Prophet sprites were based on the recolored Oracle’s sprites. During the recolor, dark grayish blue was replaced with dark brown, which worked well for things like the armor shadows, but does not fit as well in other areas such as the body outlines. Dark blue color should be restored in some areas of the sprites where it fits better than the current palette.

>>        Additional movement of the Saurian Seer’s pauldron fringes in NE frames, similarly to how they behave in SE sprites.

I like the hat on the Prophet, but these do need revision, especially the Spearthrower (or whatever it ends up being).  Let's get the base sprite hammered out before worrying about the animations.  The Seer & Prophet animations probably can be propagated through existing assets, but the intent of the Spearthrower should be reflected in the clothing/equipment, stance, and animations.  The existing Flanker looks more defensive than the current Warden.

czyzby:
@doofus-01

> The existing Flanker looks more defensive than the current Warden.

I've also noted this in the other PR from Jonathan. Without creating new sprites, I think the best idea would be to swap the sprites with the Ambusher tree, i.e. use ones with the shield for the new defensive upgrade. However, this might cause some confusion among the players that are used to the current Ambusher look.

An entirely new sprite might be ideal, but I'm not sure if I'll be able to prepare a full set of animations without reusing some of the frames. I do like the new look that you've proposed - my only criticism would be that perhaps its right hand and shoulder are just a bit too high, and it looks too hunched and bulky compared to the others as a result. Here's a version with the arm lowered by 1px:

Spearthrower

czyzby:
@CelticMinstrel 

> Not all valid words will be found in a dictionary. I think "spearthrower" is just fine as a name in general. I'm not sure that it's the best name for this unit, though. Still, if no-one can think of a better name, I think it'll do.

What about Guardian→Javelineer or Guardian→Defender?

knyghtmare:
> @CelticMinstrel
> 
> > Not all valid words will be found in a dictionary. I think "spearthrower" is just fine as a name in general. I'm not sure that it's the best name for this unit, though. Still, if no-one can think of a better name, I think it'll do.
> 
> What about Guardian→Javelineer or Guardian→Defender?

Spearguard -> Javelineer ?

(there is already a Dune Spearguard though)

CelticMinstrel:
> What about Guardian→Javelineer or Guardian→Defender?

That first progression seems quite weird, as "guardian" has no implication of spear throwing and "javelineer" has no implication of guardian.

The second progression also feels kinda weird, or rather, it feels like a regression.

> Spearguard -> Javelineer ?

Has the same problem as Guardian -> Javelineer.

czyzby:
@CelticMinstrel 

> there is already a Dune Spearguard though

I wouldn't worry about that, there are multiple name clashes among different species (e.g., fighter, warrior, guard, archer). Drakes are a bit different, since they're from the same faction.

> That first progression seems quite weird, as "guardian" has no implication of spear throwing and "javelineer" has no implication of guardian.

On the other hand, "spearthrower" has no implication of the unit's defensive capabilities. I'm not strongly against it, and it's the name that I've originally planned for this unit, but right now I'm not sure if it's the best choice. "Spearthrower" seem to suggest that the ranged attack is the focus of the unit, but its strength lies in its versatility and melee first strike ability. "Javelineer" also does this to an extent, but it might be associated with the Loyalist Javelineer by the players, which is actually its closest equivalent among existing units. (As a side note, that might actually be a good reason to choose a different name to avoid making the units too similar.)

I think any combination of guard synonyms could work. Now that I think about it, I quite like the sound of Guard→Sentinel. Javelineer is fine too, but as you've noticed it would likely require a different set of names for the progression to make sense.

knyghtmare:
I should clarify why I (yes, I changed it from Spearwoman to Spearthrower) changed the names/ids in WftN. It's more of a ranged focused damage class unit there as oppose to its original purpose (a defensive ranged damage unit). Which is why the Saurian Spearthrower -> Saurian Javelineer has been used.

CelticMinstrel:
Guard(ian) -> Sentinel sounds good to me.

czyzby:
I've renamed the units to Guard→Sentinel, since I believe these will fit the new proposed sprites and stats better than Spearthrower→Javelineer. I've also rewritten the descriptions a bit to fit the new names. Let me know if it's OK now.

Hejnewar:
If it will be looking like that maybe some resistance changes? It doesn't fit that the unit would have the same resistances when looking like that.

Jonathan-Kelly:
> "Javelineer" does sound nice, but I think perhaps we should reconsider "spearthrower" to something that appears in a dictionary. :)

How about [Peltast](https://en.wikipedia.org/wiki/Peltast)? Considering saurians have hellenistic sounding names.

Regarding statistics, could this unit also get a ranged impact damage attack, like a sling or bola, to better counter loyalist spearman spam? Two of the loyalist spearman advancements get 40% pierce resistance, which makes them not just devastating against drakes, but fairly resistant to saurians too. This can make loyalist v drake matches rather boring by putting so much focus on deploying and outmaneuvering the loyalist spearmen, while most other units can only play minor roles. It may even be the least interesting match up since the holy-to-arcane switch so vastly improved undead v undead, many years ago.

If saurians could more strongly counter spearmen, then a greater variety of loyalist units would have to deploy to counter saurians, and in turn be countered by a greater variety of drake units.

CelticMinstrel:
> How about [Peltast](https://en.wikipedia.org/wiki/Peltast)? Considering saurians have hellenistic sounding names.

That seems a bit on the obscure side to me. I'm not wholly against it, but I think it might confuse a lot of people. (Also, on a quick glance they seem to be described as skirmishers…)

czyzby:
@Jonathan-Kelly 

> How about [Peltast](https://en.wikipedia.org/wiki/Peltast)? Considering saurians have hellenistic sounding names.

Wiki says they're named after a shield with a specific shape. Unless we want to cram in a shield on top of that huge spear, I don't think that name is a good fit. 

> Regarding statistics, could this unit also get a ranged impact damage attack, like a sling or bola, to better counter loyalist spearman spam?

I think impact attacks are rarer and weaker on average, so that means the base damage would also have to be lowered accordingly. I originally wanted this unit not to deviate too much from the original, but rather provide a sort of what-if option that represent roughly the same unit without the skirmish ability. Not sure how I feel about this one to be honest, but if people responsible for the stats (@Hejnewar?) consider it a good idea, we can consider a different attack type. Perhaps a melee spear bash in addition to the pierce attack is also an option? Extra resistances might also solidify them as a spearman counter.

@Hejnewar

> If it will be looking like that maybe some resistance changes?

I agree, the new sprite does seem more bulky and better equipped. Would you like to propose some changes to the unit statistics?

CelticMinstrel:
> I originally wanted this unit not to deviate too much from the original

It should have something to distinguish it though… a reason to choose it instead of the skirmisher. What is that distinction, exactly? (I haven't looked at the stats.)

czyzby:
First strike melee ability, stronger ranged attack, -1 speed (6), more HP and possibly higher resistances. Rather than a melee skirmisher/scout, it's more of a defensive mixed fighter. The closest equivalent would be the Loyalist Javelineer.

Hejnewar:
> Regarding statistics, could this unit also get a ranged impact damage attack, like a sling or bola, to better counter loyalist spearman spam? Two of the loyalist spearman advancements get 40% pierce resistance

While the premise is good (more variety), the arguments are pretty bad, because loyalists use swords vs drakes most of the time. (Drakes aren't the problem, saurians are, you already have enough pierce but not enough blade damage.)

> I think impact attacks are rarer and weaker on average, so that means the base damage would also have to be lowered accordingly. I originally wanted this unit not to deviate too much from the original, but rather provide a sort of what-if option that represents roughly the same unit but without the skirmish ability. Not sure how I feel about this one to be honest, but if people responsible for the stats (@Hejnewar?) consider it a good idea, we can attempt to change the attack type.

From a player perspective it is good to have more choices, having too similar choices is not actually that great, because they are still similarly good in similar situations which makes them not that meaningful for players so I can't help but support that. At the same time I think I understand your feelings a bit as I was subjected to something similar few times. :P

Don't worry too much about stats, for me it's not a lot of work at this point, impact damage is in fact often lower so maybe 9-4 instead of 12-3 would fit better but actual effect of impact on strength of a unit is low enough that if necessary I can just deal with it later. 

> Perhaps a melee spear bash in addition to the pierce attack is also an option? Extra resistances might also solidify them as a spearman counter.

Spear bash sounds a bit clunky and not very threatening to be honest. 

> I agree, the new sprite does seem more bulky and better equipped. Would you like to propose some changes to the unit statistics, or should I? It it's meant to stay on roughly on the same power level, we'll probably have to lower terrain defenses, HP or ranged damage to make up for the changes.

I have few ideas for example just increasing all physical resistances by flat amount would be alright but feel free to show your ideas, what matters to me more is payment for it, terrain defences are pretty unique for saurians but lowering the good one by 10% is a possibility still keeps some of the originality but not exactly, movement points going to 5 is also something to consider, HP and damage are fine and can be used if needed.  

I probably should also look into buffing Ambusher and Flanker as they are kinda not that great at the moment.

czyzby:
> I have few ideas for example just increasing all physical resistances by flat amount would be alright but feel free to show your ideas, what matters to me more is payment for it, terrain defences are pretty unique for saurians but lowering the good one by 10% is a possibility still keeps some of the originality but not exactly, movement points going to 5 is also something to consider, HP and damage are fine and can be used if needed.

My first thought was a flat +10% physical resistance for the Lv2 and +20% resistance for the Lv3. If you consider the Ambusher to be underpowered, maybe further changes are not necessary to balance the unit, but perhaps lowering the HP by 1 or 2 would be sufficient to offset the change. I've added the resistance changes and increased costs by +1/+2 respectively - please verify if the unit stats are correct now.

> From a player perspective it is good to have more choices, having too similar choices is not actually that great, because they are still similarly good in similar situations which makes them not that meaningful for players so I can't help but support that. At the same time I think I understand your feelings a bit as I was subjected to something similar few times. :P

If it was up to me, I'd leave the pierce spear throw ranged attack as is. I think it fits the unit thematically, and - to be completely honest - this way I can reuse most of the existing poses to provide high quality animation frames with less effort. However, I can apply this change for balancing reasons if necessary.

czyzby:
That would be the Ambusher. The original proposed sprites for the Guard/Spearthrower tree:

Lv2 Lv3

The new proposed sprites (WIP):

New

It would not be impossible to add a shield at this point, since I haven't started working on the animations yet, but I have to say I quite like the distinction from the Ambusher.

CelticMinstrel:
Okay, then yeah, it's difficult to fit an impact melee attack in with that sprite. I guess you could do a bull rush type attack though. (It would be cool if it even pushes the enemy onto another hex, though that might be too complicated for a mainline unit, and isn't really _necessary_ for a bull rush anyway. It could stun instead, for example.) The other thing I can think of if you wanted an impact attack would be sling (straight impact) or bolas (impact with slow). Ranged weapons are typically not shown in the base sprite at all (except bows), so almost anything goes.

knyghtmare:
the spear's rear end can be a quarter-staff type impact attack. that can work

> The original proposed sprites for the Guard/Spearthrower tree:

Also, since the Guard/Sentinel has decided to adopt new appearances, I think the old ones should stay as Spearthrower and Javelineer which will be a ranged damage focused unit. If not mainline, then UMC would welcome it.

czyzby:
@knyghtmare 

> Also, since the Guard/Sentinel has decided to adopt new appearances, I think the old ones should stay as Spearthrower and Javelineer which will be a ranged damage focused unit. If not mainline, then UMC would welcome it.

I don't think there's a way around it, since they don't meet mainline standards as they are. I might prepare the Saurian Pack add-on for Wesnoth 1.17 with the Spearthrower/Javelineer units as a resource for campaigns, but the truth is someone will probably just copy them sooner or later either way. 😀 If you're working on a campaign add-on, we can discuss using Saurian Pack as a dependency if you'd like.

@CelticMinstrel 

> Okay, then yeah, it's difficult to fit an impact melee attack in with that sprite. (...)

@knyghtmare 

> the spear's rear end can be a quarter-staff type impact attack. that can work

As for the attacks, I think the correct approach here is to agree on stats first and then update the sprite or the animations accordingly. For me personally, the first strike melee attack with a spear is the signature of the unit, but other than that we can discuss other options. I wouldn't want to overcomplicate the unit by introducing stun-inducing or charge-like attacks, especially since the lore says they're quite small. I feel like the first strike and increased resistances are enough in terms of abilities for a single mainline unit.

The melee impact attack with the spear pole seems like a good compromise between providing more damage type coverage for saurians, and staying true to the unit tree and initial vision. If we're considering changing the attacks at all, that is. I know it's subjective, but thematically a saurian with a sling just doesn't sit right with me, and overall feels like a downgrade from the javelins. Perhaps a blowpipe would be better, but is isn't exactly dealing impact damage either.

So, to sum up, if it makes sense in terms of the faction balance, I propose the following secondary melee attacks:

```ini
    # Guard
    [attack]
        name=spear
        description=_"spear"
        range=melee
        type=impact
        damage=8
        number=2
    [/attack]

    # Sentinel
    [attack]
        name=spear
        description=_"spear"
        range=melee
        type=impact
        damage=8
        number=3
    [/attack]
    # The icons can be chosen to reflect the spear poles.
```

Without the first strike ability and roughly 80% damage, they would mostly serve as additional coverage for opponents such as the undead, which saurians can struggle to deal with otherwise.

Jonathan-Kelly:
> (Drakes aren't the problem, saurians are, you already have enough pierce but not enough blade damage.)

> For me personally, the first strike melee attack with a spear is the signature of the unit, but other than that we can discuss other options.

How would you feel about a blade damaging [Glave](https://en.wikipedia.org/wiki/Glaive) type pole arm instead of the spear as the first strike melee weapon?

The melee attack animation might be kept similar to how it is now, with a kind of thrust-slash. Since saurians have proportionately weaker arms that provide little power swinging weapons around in wide motions, but their strong legs provide ample power for strikes that involve leaping or thrusting.

CelticMinstrel:
> the spear's rear end can be a quarter-staff type impact attack. that can work

It's definitely possible in theory, but is it safe to hold your spear like a staff? I'm not sure, it might be, but I'm just not sure.

> I know it's subjective, but thematically a saurian with a sling just doesn't sit right with me, and overall feels like a downgrade from the javelins. Perhaps a blowpipe would be better, but that weapon isn't exactly dealing impact damage either.

A blowpipe is _usually_ needles, but I don't know of any reason it couldn't instead be pellets. Although, it would be hard to justify a high damage for that.

Another possible ranged impact weapon that I can think of would be a boomerang, which would certainly make for a cool animation but I guess might not be a great fit for a spear-wielder…

czyzby:
@Jonathan-Kelly

> How would you feel about a blade damaging [Glave](https://en.wikipedia.org/wiki/Glaive) type pole arm instead of the spear as the first strike melee weapon?

That could work if we're set on diversifying their attack types. The questions are whether A) the thrust animation can be reused for a slashing weapon, B) first strike fits the first strike ability (although the long reach requirement does seem to be met). @doofus-01 Thoughts?

Hejnewar:
> How would you feel about a blade damaging [Glave](https://en.wikipedia.org/wiki/Glaive) type pole arm instead of the spear as the first strike melee weapon?

Thats actually pretty good idea. 

>  A) the thrust animation can be reused for a slashing weapon

For this just choosing correct weapon type might be enough. Thrusthing does not inherently mean that you are trying to pierce something it depends on a weapon. I even have a specyfic thing in mind but I cant find pics of it.

knyghtmare:
I think this PR has deviated extensively from its original intention. It started as a talk to add 4 new Saurian advancement and extended their limited unit variety, which is good thing. Now, it's...I don't know what it is anymore since the discussion of a new advancement-tree for the Saurian Skirmisher.

Thus, I propose the following:
1. Add Saurian Prohpet, Saurian Seer as is. They are being added level 3 advancements and AoH leaders for the Drakes faction so it's okay.
2. Saurian Spearthrower and Saurian Javelineer be added as campaign-only units. They would have a macro `{ENABLE_SPEARTHROWER}` to enable them for _Winds of Fate_, _Legends of Wesmere Reborn_ and _War for the North (NRR)_. This would satisfy the PR's intentions as well. Otherwise, these fully animated (and in-line with core saurians) 2 units would be discarded, which seemed rather hampering. Additional, their stats might be made in-line with what's being used in the campaign candidates such +1 ranged damage for both (would match both WftN and LoWR).
3. This new "Saurian Guard" -> "Saurian Sentinel" unit-line which is being discussed and intended to be available as a default alternate advancement to the Saurian Skirmisher and available as a leader in Drakes of Default Era and as recruit in AoH Drakes should be discussed and made in. a new PR. The Guard-line sounds drastically different from the Spearthrower is and should be its own unit.

czyzby:
I agree with @knyghtmare with one small exception - this PR already has some valuable discussions and feedback on the new saurian unit, so if we're still aiming to expand the mainline saurians, I'd like convert it to a WIP PR regarding the new unit, and remove unrelated files completely.

If there are no objections, I'd like to submit 2 new PRs:

* Lv3 upgrades for Oracle and Soothsayer: since they're going to be added to the mainline factions, I think the review for these should be harsher, both in terms of the sprites as well as the statistics.
* Spearthrower/Javelineer units: a redesigned upgrade for the Skirmisher focused on ranged attacks. While available in the core, it would be entirely optional and disabled in the mainline eras by default. Added with the campaigns and UMC in mind through the `{ENABLE_SPEARTHROWER}` macro. I'd also remove the first strike ability for further distinction from the upcoming Guard/Sentinel units.

czyzby:
I've modified the PR and converted it into a draft. The goal of this PR is now to design and implement a new alternative upgrade tree for the Saurian Skirmisher. The unit archetype is a _defensive mixed fighter with the first strike ability._ Comparison with the Ambusher:

* Lack of the skirmisher ability.
* Slower (6 vs 7).
* Better defensive capabilities: increased HP and resistances.
* Stronger ranged attack, with comparable albeit weaker melee attack.
* Meets the requirements for a faction leader due to its lower speed.
* _Possibly_ access to other damage types than pierce (e.g., blade melee attack).

I'm comfortable with creating the animations for the unit, especially if I can reuse the existing saurian frames (which isn't uncommon for that species to be honest) or have some reference intermediate poses for the movements. If that's alright, it would be very helpful if we could finalize the both units' basic SE and NE sprites together, @doofus-01. Once you approve the sprites, I can continue working on the animations. If there are no objections, we can go with the glaive as the main melee weapon to provide more variety in terms of the damage types. Perhaps the overall shape blade could look like `d)` in the following Wiki example (without the ornaments):

Wiki glaives

Jonathan-Kelly:
> * Lv3 upgrades for Oracle and Soothsayer: since they're going to be added to the mainline factions, I think the review for these should be harsher, both in terms of the sprites as well as the statistics.

I think these will make highly desirable additions (especially for campaigns and survival scenarios) and suspect they could reach completion and approval relatively quickly. I found they made a big improvement to WoF.

> * Spearthrower/Javelineer units: a redesigned upgrade for the Skirmisher focused on ranged attacks. While available in the core, it would be entirely optional and disabled in the mainline eras by default. Added with the campaigns and UMC in mind through the `{ENABLE_SPEARTHROWER}` macro. I'd also remove the first strike ability for further distinction from the upcoming Guard/Sentinel units.

But I think the guard and spearthrower may be redundant with each other, as they both try to fill the same role as an alternate skirmisher advancement with greater focus on ranged attacks and durability. I would rather go with whichever of these two units becomes a part of the mainline drakes MP faction, to keep things consistent.

knyghtmare:
> But I think the guard and spearthrower may be redundant with each other, as they both try to fill the same role as an alternate skirmisher advancement with greater focus on ranged attacks and durability. I would rather go with whichever of these two units becomes a part of the mainline drakes MP faction, to keep things consistent.

I think the Saurian Guard-line could be specialised more on durability by adopting a more brutish weapon selection.

She could be 1 MP slower than the skirmisher, +10 resistances to blade/impact/pierce, the melee spear can be swapped for either a glaive (blade melee) and also pierce. Or a maybe a club for impact melee.

The ranged attack doesn't have to be a spear.
I think to further distinguish her from the Spearthrower, she should adopt the use of a net (7-2 slows) or a blowgun (ranged pierce poison).
just buffing the ranged pierce spear would make her more identical to the Spearthrower which seems to be a blocker.
One further considerable is (it may be adversely taken which I think would be some people's reaction) is whether Saurian Guard should get a level 3 at all. It's still conceptual and under discussion as I see the comments. Having either slows/poison makes it rather good and utilisable so does it need the level up?

czyzby:
@Jonathan-Kelly 

> But I think the guard and spearthrower may be redundant with each other, as they both try to fill the same role as an alternate skirmisher advancement with greater focus on ranged attacks and durability. I would rather go with whichever of these two units becomes a part of the mainline drakes MP faction, to keep things consistent.

That's why the Spearthrower is turned off by default and enabled on case by case basis with a macro in the recent pull request (#6912). I agree this upgrade is a bit redundant. However, it provides just enough variety to the saurians to be useful for the campaigns, even if it's not distinct enough to make it into the mainline. If we were making all of these from scratch, I agree that Spearthrower/Javelineer would not be necessary - however, that is not the case. It's like @knyghtmare said, the alternative is simply to discard these fully animated units that could very well make it into the mainline campaigns in one way or another. From what I understand, they were already approved by some of the authors of core campaign candidates.

I've revamped the spearthrower stats for it to be more of an offensive ranged unit, without the first strike ability or increased resistances. I think there's still place for both in campaigns or custom scenarios, especially if the guard gets a different attack type - if you think about it, the Loyalist Spearman upgrades have a similar level of overlap mechanically.

Not to mention, it might take some time before (and if) the Guard and Sentinel make it into the core. Designing an entirely new unit was not my intention as the start, and in hindsight proposing the spearthrower for the default era might have been misguided, as it completely hijacked the discussion and resulted in brainstorming for a new unit. I wouldn't want to slow down campaign development. Not that I don't enjoy working on it though. 😀

@knyghtmare 

> She could be 1 MP slower than the skirmisher, +10 resistances to blade/impact/pierce, the melee spear can be swapped for either a glaive (blade melee) (...)

That's the current plan for the guard. Since it would be a core mainline unit, I don't think we should try to make it as distinctive as possible to differentiate from the Spearthrower, which might or might not be added and is designed for campaigns/UMC on top of that. Rather, we should focus on whether A) it makes sense as a Skirmisher upgrade, B) is well balanced against the Ambusher, C) does not try to do too much with too many mechanics crammed in.

Correct me if I'm wrong, but both the slowing and poisoning attacks might make the Guard a better option than the Ambusher in most situations, and would give Drakes access to new mechanics. Skirmisher is an excellent ability, but the sheer utility and versatility of the Guard with all of these extra abilities could make the Ambusher highly situational and generally underused.

If we're worried about that at all, what we can do to make it a bit more distinctive to Spearthrower without further replacing the attacks is to make the unit melee-focused. I think it should still deal more ranged damage than the Ambusher, but perhaps changing the attacks to 8-3/12-2 blade melee and 6-3/7-3 pierce ranged would be more appropriate and give more of an edge to the Spearthrower.

> One further considerable is (it may be adversely taken which I think would be some people's reaction) is whether Saurian Guard should get a level 3 at all. It's still conceptual and under discussion as I see the comments. Having either slows/poison makes it rather good and utilisable so does it need the level up?

This is something that I've also considered, even without the extra abilities. Especially since its closest equivalent - the Loyalist Javelineer - also does not have another upgrade. Again, I probably wouldn't immediately jump at poison or slow, but perhaps the steadfast ability would be a good fit to enhance its defensive capabilities. I could imagine these working quite well for a permanent level-2 unit:

```ini
    [abilities]
        {ABILITY_STEADFAST}
    [/abilities]
    [resistance]
        blade=90 # or 100?
        impact=90
        # pierce=80 by default
    [/resistance]
```

That's quite an improvement in terms of the resistances, so the other stats would probably have to be nerfed accordingly. At the very least, I think bringing the steadfast ability might require lowering the unit speed to 5.

Jonathan-Kelly:
My sense is that the best fitting alternate saurian skirmisher advancement is somewhere in between the current spearthrower and guard. By fitting, I mean:

* **Fast** and **Damaging**  _(as per the drakes faction)_
* **Terrain Defense** over **Resistances**  _(as per the saurian race)_
* Add **Strategic Variety**  _(to the drakes faction)_

From testing it in WoF, my impression was the spearthrower mainly lacked a bit of damage type variety. Which put it somewhere in between the clasher, skirmisher and augur. Perhaps all it needed was something like one (or more) of:

* (Blade) Glave Melee Attack
* (Blade+Pierce) Glave Melee Attacks
* (Blade) Assegai/Boar/Bear Spear Ranged Attack  _(additional to regular pierce throwing spears)_

---

Whereas taking it in a hulking direction with more resistances, slower movement and/or steadfast feels a bit strange for its faction, race and role as a skirmisher advancement. Focusing more on its ranged attack(s) seems like a better fit for a small and swift saurian skirmisher.

czyzby:
That's actually a good argument to increase terrain defenses over the resistances, although those could also be easily explained by the better equipment. Perhaps `steadfast` is taking it too far, but I wouldn't mind flat +10% resistances to all physical attacks, or just blade and impact to address their weaknesses. We can also adjust the defenses on some strategic terrains on top of that.

doofus-01:
There was a lot said here over the last week, sorry if I miss or gloss over something.

> An entirely new sprite might be ideal, but I'm not sure if I'll be able to prepare a full set of animations without reusing some of the frames. I do like the new look that you've proposed - my only criticisms would be that perhaps its right hand and shoulder are just a bit too high, and it looks too hunched and bulky compared to the others as a result. The javelins on the back also do not seem to conform to the style of saurians with their stronger outlines. Here's a version with the arm lowered by 1px:

The sprite I posted was more for concept, it had a lot wrong with it.  You've cleaned it up well, but we shouldn't feel locked into it yet.  The advancement would probably need more distinction, but we can worry about that later too.

> I'm comfortable with creating the animations for the unit, especially if I can reuse the existing saurian frames (which isn't uncommon for that species to be honest) or have some reference intermediate poses for the movements. If that's alright, it would be very helpful if we could finalize the both units' basic SE and NE sprites together, @doofus-01. Once you approve the sprites, I can continue working on the animations.

Recycling some of the animations can make sense to some extent, or at least using them as a base, but we don't want too many clones.  The fact that it already happens is more of a reason not to do it again.  To get this committed, you don't need to match all the skirmisher animations, that can come later. Just so long as these don't stand out for being sliding statues.

I guess further discussions need to go to the different PRs

doofus-01:
I was confused in my earlier post, didn't realize how many new branches were being proposed.  Was any consensus reached on the stats/abilities of this defensive line?  That would influence the final sprite and animations.

czyzby:
No final stats, but from what I've gathered so far:

- An extra damage type would bring much needed variety to the skirmisher tree. Replacing the pierce/spear with blade/glaive seems like a reasonable approach.
- Steadfast and extra defenses do not fit the saurian theme very well, but their defenses on some terrains might be increased by a small amount to give them some better tanking capabilities.
- I personally think the ranged throw attack should stay as is, perhaps with a change the ammunition from spears to lighter javelins to justify increased attack count. This seems most in line with the current skirmisher tree. I originally planned for the defensive branch to be a mixed fighter with roughly equivalent melee and ranged damage, so the proposed sling/web do not quite fit that pattern - I think a simple piercing attack is the way to go.

If there are no objections, I can propose stats that would fit what I've described here.

doofus-01:
Do they need to keep using a pole-arm?  If they had some falx or other sword-yet-not-sword thing, that could be part of their enhanced nimbleness, that gives them better terrain defense.  With a high ranged attack number, they could be a really annoying archer/mixed-fighter.

czyzby:
I originally designed them to be the Saurian equivalent of the Loyalist Javelineer. I think pole weapons are the way to go if we want to keep the first strike ability, but I wouldn't mind switching it up for a different kind (like halberd).

CelticMinstrel:
I do agree that firststrike implies a pole weapon, as my understanding is they're able to always strike first because they have greater reach with their weapon. More complicated pole weapons (anything other than spear or pike) don't fit quite as well, either, and halberd fits particularly poorly as a first strike weapon in my opinion. Something like a glaive might work? The key point is that it should still be chiefly a _thrusting_ weapon, which means not having an axe blade on the side.

czyzby:
Glaive could work, but I guess it comes down to whether a blade melee attack is a good idea in the first place. I believe the unit as originally proposed (with a pierce melee attack) should also still be considered:
* Skirmishers are used by drakes in larger numbers for their pierce attacks (and resistance). While Ambushers serve as a clear upgrade over the Skirmishers, they are scouts/skirmishers first and foremost. They lack the strength and durability that Guards could provide by sacrificing high mobility. Leaving Guards with a pierce melee attack would be a more faithful progression of the base unit, allowing for a stronger alternative upgrade with roughly the same counters and purpose.
* Spear, or another longer polearm with a pierced attack, is perhaps the best fit for the first strike attack thematically among all weapons.
* Blade is the most common attack type, already covered by 3 out of 4 level-1 drakes and all of their upgrades. It may seem like we're giving more options to the saurians by switching the attack type from pierce to blade, but when viewed as a whole, the drakes faction does not have a lot of options for pierced attacks to begin with - having only the Drake Clasher and Saurian Skirmisher along with their upgrades. Switching their main attack style to blade would feel like a downgrade in some match-ups.

@knyghtmare has also suggested an interesting option: impact first strike attack with a large shield. That could both fit the criteria for a diversified attack type, as well as the defensive theme of the unit. Impact attacks are also rarer than blade, with the only viable options being the level-2 Drake Thrasher and Drake Glider unit tree. The usefulness would vary from match-up to match-up, but I feel like it would be a better fit than a blade attack.

Anyway, at this point it might make sense to either do a poll on the available options (pierce spear vs blade glaive vs impact shield), or just leave the choice to the person responsible for core units design or balance. My only preference and suggestion would be not to go overboard with redesigned attacks and abilities. There's place for an axe-wielding saurian with a slowing bola ranged attack or a rogue saurian with notably higher terrain defenses, but perhaps as separate unit trees.

knyghtmare:
> has also suggested an interesting option: impact first strike attack with a large shield. That could both fit the criteria for a diversified attack type, as well as the defensive theme of the unit. Impact attacks are also rarer than blade, with the only viable options being the level-2 Drake Thrasher and Drake Glider unit tree. The usefulness would vary from match-up to match-up, but I feel like it would be a better fit than a blade attack. However, I'm not sure if a shield bash can and should have the first strike ability.

My suggested things were:
HP: ~40-42
XP: 100 (AMLA_DEFAULT)
MP: 5
same movement type
cost: 27
Resistance: +10% blade/impact/pierce

Attacks
12 - 2 heavy shield
melee impact
defend-only + first-strike
5 - 2 sling
ranged impact

CelticMinstrel:
Don't forget that you could also give it three attacks. I don't know if this is a good idea, but the shield bash for example could be an extra attack supplementing melee and ranged pierce.

czyzby:
> @knyghtmare
> 5 - 2 sling
> ranged impact

My issue with this would be that without any matching attacks, it would feel like a completely different unit rather than an upgrade. I'd be fine with the shield bash variant if it had spear throw ranged attack.

If first strike does not fit the shield attack well, perhaps we can go with a new ability that would work sort of between the first strike and reversed charge: the first melee attack made against the Saurian Guard would deal half of the damage. While not quite as effective in counter attacks as first strike, it would increase their resistances without explicitly doing so.

> @CelticMinstrel 
> Don't forget that you could also give it three attacks. I don't know if this is a good idea, but the shield bash for example could be an extra attack, not firststrike, supplementing melee and ranged pierce.

I've thought about it. We could go with a halberd main weapon with pierce first strike attack and a secondary melee blade attack. My only concern is that it would feel samey to the arbiter, but maybe it isn't a bad thing at all.

Jonathan-Kelly:
> We could go with a halberd main weapon with pierce first strike attack and a secondary melee blade attack. My only concern is that it would feel samey to the arbiter, but maybe it isn't a bad thing at all.

If we take another look at [the array of glaive designs you posted earlier](https://github.com/wesnoth/wesnoth/pull/6905#issuecomment-1195947019), we see many have pointed tips like spears. This allows them to both stab or slash with a similar jabbing motion. Which:
* Fits the saurian body plan (strong legs, weak arms).
* Gives both firststrike blade and firststrike pierce (statistically differentiating it from drake/human halberds).
* Is unique looking (visually differentiating it from halberds).
* Is the kind of weapon a saurian guard needs to ward against other saurians (as they are strong to pierce but weak to blade).
* Allows reuse of the same attack animation for both melee attacks.

There is actually a big but poorly understood difference between chopping weapons (axes, halberds) versus slicing weapons (swords, naginata). Chopping weapons require a heavy swing to deliver damage via momentum, not unlike a mace or cudgel. But lighter, swifter slicing weapons rely on edge movement across the target. Which allows them to cut with a quick jabbing motion, not unlike that used for a spear.

czyzby:
> Allows reuse of the same attack animation for both melee attacks.

I think a separate slashing animation would be nice though.

> Gives both firststrike blade and firststrike pierce (statistically differentiating it from drake/human halberds).

The first strike ability is not necessary on the more aggressive slashing attack, but otherwise I agree. A ranged attack and different resistances should be distinctive enough to differentiate from the drake unit. The more I think about this, the halberd/glaive approach with 2 melee attacks is the closest to my original vision for the Guard while simultaneously ticking the box for more diversified attack types. I'll update the unit statistics.

spixi:
Northerners got the [Slayer as a default advancement](https://github.com/wesnoth/wesnoth/commit/1ba0a11a1850b09338cde1a4c750b0b8a58a5348) and also a leader in AoH. Therefore, I have no problem to do the same with the Saurian units. In that case, the Great Troll should also be default advancement, as it is Loyalist’s only neutral unit and could help against Drakes playing a Saurian leader.

Saurian Soothsayer and Saurian Oracle should not be leader choices for AoH anymore. While Soothsayer’s Cures and Heal +8 is the best healing option Drakes have, it is still a very frail Level 2 unit. Oracle is pretty useless as a leader in AoH and outside of fun matches it is never used. Maybe we could have an advancement of the Oracle.

CelticMinstrel:
> Saurian Soothsayer and Saurian Oracle should not be leader choices for AoH anymore.

I would think this is contingent on actually adding the Seer and Prophet (or whatever they end up being called).

spixi:
> > Saurian Soothsayer and Saurian Oracle should not be leader choices for AoH anymore.
> 
> I would think this is contingent on actually adding the Seer and Prophet (or whatever they end up being called). They were mentioned earlier in this thread, and I thought there was a PR regarding them too?

Yes, that is what I mean. Northerners got Orcish Slayer, so Level 3 Saurians would be a reasonable addition. Because all these units are chaotic, so it would be also fair when AoH Loyalists can advance their Ogres.

The PR you mentioned is found [here](https://github.com/wesnoth/wesnoth/pull/6911).

knyghtmare:
> > > Saurian Soothsayer and Saurian Oracle should not be leader choices for AoH anymore.
> > 
> > 
> > I would think this is contingent on actually adding the Seer and Prophet (or whatever they end up being called). They were mentioned earlier in this thread, and I thought there was a PR regarding them too?
> 
> Yes, that is what I mean. Northerners got Orcish Slayer, so Level 3 Saurians would be a reasonable addition. Because all these units are chaotic, so it would be also fair when AoH Loyalists can advance their Ogres.
> 
> The PR you mentioned is found [here](https://github.com/wesnoth/wesnoth/pull/6911).

I think it was already planned to enable the Saurian Augur level 3 advancements as default. But this is out scope of this draft PR and is in another PR.

Hejnewar:
What is the current state of work for this PR?

doofus-01:
It will take a high priority, concerted effort by @czyzby and me to get this ready in time for 1.18 default balancing, and I don't see it happening.  We could still get this in as an advancement that is available, but not default.

czyzby:
I did start some animations for this unit, but nothing advanced yet, or matching the mainline quality. I might still look into it, but only after #6911.

Hejnewar:
People were excited about both of the PRs so thats kinda a pity.

Pentarctagon:
Is this still being pursued?

czyzby:
No, sorry, no time to finish the animations right now. If I'll get around to it, I'll focus on [updated Seer NE animations first](https://forums.wesnoth.org/viewtopic.php?t=56186) if they aren't finished.

Pentarctagon:
What's the status of this PR?

PR #6910, opened 2022-07-26
distinguish noun from verb

celerini:
I added contexts to the word Recruits to be able to translate it by proper Czech words.

### DIFF
--- a/data/gui/window/mp_faction_select.cfg
+++ b/data/gui/window/mp_faction_select.cfg
@@ -214,7 +214,7 @@
 
 							[label]
 								definition = "gold"
-								label = _ "leader^Recruits"
+								label = _ "Recruits"
 								linked_group = "details_label"
 							[/label]
 
--- a/data/gui/window/mp_faction_select.cfg
+++ b/data/gui/window/mp_faction_select.cfg
@@ -214,7 +214,7 @@
 
 							[label]
 								definition = "gold"
-								label = _ "Recruits"
+								label = _ "leader^Recruits"
 								linked_group = "details_label"
 							[/label]
 
--- a/src/gui/dialogs/statistics_dialog.cpp
+++ b/src/gui/dialogs/statistics_dialog.cpp
@@ -364,7 +364,7 @@ void statistics_dialog::update_lists()
 
 	const statistics::stats& stats = current_stats();
 
-	add_stat_row(_("number_of^Recruits"),     stats.recruits);
+	add_stat_row(_("stats^Recruits"),     stats.recruits);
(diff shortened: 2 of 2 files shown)


CelticMinstrel:
The contexts and the commit message don't really make sense to me, though I may be missing something.

Which one of those is a verb? They both look like nouns to me.

celerini:
I read the one in MP leader selection as a verb (a leader recruits units in a list below), now I see it in the Help/Eras as clear noun, my fault, sorry for the noise in the commit message.

Thus the contexts are poorely named, you are right. I do not want to polute the code with them.

But I still would like to have a context for Recruits to differentiate between stats and MP leader selection. Plain translation of Recruits in stats makes other stat names hard to translate or inconsistent (now we use verbs in passive (recruited, recalled...) in our translation).

It looks like at least German and Ukrainian translation do not use a plain noun for stats, so their MP leader selection has probably wrong word for Recruits there.

CelticMinstrel:
Yeah, I have no objection to the spirit of this pull request. We just need to make sure that the added contexts make sense, and ideally have an accurate commit message. (The commit message can be rewritten on merge though, so the contexts are the priority.)

celerini:
Next iteration. For the sake of simplicity I removed unnecessary context in MP faction select and renamed the one in the stats dialog, which can be then translated more freely.

Wedge009:
I can only find two instances of `_("Recruits")`: the statistics dialogue (this PR) and the Help pages for era factions. In both cases they are nouns.

Maybe it is better to add a po hint instead?

Pentarctagon:
So can this be merged, or should it be a po hint instead?

CelticMinstrel:
From my understanding, a po hint would not help here.

celerini:
I see three occurences of the word Recruits:

```
grep -B3 '"Recruits"' po/*/*.pot | grep "#:"
po/wesnoth-lib/wesnoth-lib.pot-#: data/gui/window/mp_faction_select.cfg:217
po/wesnoth-lib/wesnoth-lib.pot-#: src/gui/dialogs/statistics_dialog.cpp:367
po/wesnoth/wesnoth.pot-#: src/help/help_impl.cpp:706
```

I wanted to split the Recruits in wesnoth-lib in order to translate the one for stats by a different word.

I played with xgettext -c now and it merges different comments above the same word. So it looks like a po comment will not split the word Recruits for me.

Thanks to @CelticMinstrel  I used the right noun in our translation and I can live with it even though it is not optimal.

So you can freely close this PR or merge it whatever looks better for you.

PR #6875, opened 2022-07-15
Update WoF

Jonathan-Kelly:
Issues #6847 and #6706 should be addressed by these changes, as well as most all issues and change requests (regarding WoF) made by players and developers up to this point.

It looks like we now have (including WoF) three in development mainline campaigns all using the four new saurian advancements sprites created by the talented @czyzby. So I thought it might make the most sense to add these to core rather than them becoming three separate copies of campaign specific units. Also, I went a step further and hooked them up to the multiplayer drakes faction since I believe their statistics were already balanced by @Hejnewar (?) and we are still over a year away from the next major release. They might help make drakes-v-loyalists and drakes-v-drakes matches more interesting with more unit types becoming useful during these?

This adds a WoF specific unit, a powerful level 5 air elemental, which may also be worth considering in relation to #5532. The sprite is from an [older version of EoMa](https://units.wesnoth.org/1.10/Era_of_Magic/en_US/EoMa_Air_God.html) from the BfW 1.10 time period. @inferno8 let me know if you have any issue with this sprite being used in mainline. I really enjoy how ominous and otherworldly it is (the same goes for your fire and water elementals).

PR #6911, opened 2022-07-26
Saurian Augur level-3 upgrades

czyzby:
This pull requests is a part of the effort to bring units that are used by mainline campaign candidates (see #6905, #6875). It contains the level-3 upgrades of the Saurian Augur from the [Saurian Pack add-on](https://github.com/czyzby/wesnoth-saurian-pack).

The pull request consists of the following changes:

* Saurian Prophet is added as an upgrade of the Saurian Oracle.
* Saurian Seer is added as an upgrade of the Saurian Soothsayer.
* Saurian Soothsayer is replaced by the Saurian Seer as a possible leader of the Age of Heroes Drakes faction.
* Saurian Oracle is replaced by the Saurian Prophet as a possible leader of the Age of Heroes Drakes faction.
* Saurian Prophet and Saurian Soothsayer level up experience values are adjusted accordingly to support the new upgrades.

The following issues might have to be addressed:

* Since the units are added to the mainline, they have to meet the sprite quality and balancing standards.
* Unit descriptions are currently reused from the level-2 units. Custom descriptions might have to be prepared.
* Saurian Prophet sprites were based on the recolored Oracle’s sprites. During the recolor, dark grayish blue was replaced with dark brown, which worked well for things like the armor shadows, but does not fit other areas such as the body outlines where it does not complement the dark green as well. Dark blue color should be restored in some areas of the sprites where it fits better than the current palette.
* ~~Animations of the Saurian Seer’s pauldron fringes in NE frames are missing, and might be added based on the SE sprites.~~
  * New Seer animations are a work in progress.

Issue #6453, opened 2022-01-18
Saurian sprites palette adjustments


czyzby:
This issue is related to #6413 and #6417.

After the #6413 pull request, the current Saurian palette consists of:

* White, black: `#ffffff`, `#151515`
* Gold armor: `#8e5e46`, `#bf6d00`, `#ffd44a`
* Gold armor shadow, dark wood: `#581802`
* Wood (spear): `#8c3e3e`, `#e45a4c`
* Wood (staff): `#4d3f19`, `#a38637`
* Skin: `#0d4218`,  `#4d8d38`, `#98d858`, `#cff521`
* Skin (oracle): `#26882e`, `#22d22f`
* Robe, staff head: `#515249`, `#9b9c92`, `#d0d1c3`
* Robe (oracle): `#17212c`, `#3f4d5c`, `#97a7b8`, `#93c4c5`
  * _Darkest shade shared with armor._
* Armor, spear:  `#17212c`, `#2c4c6c`, `#618d8d`, `#8dc2c2`, `#c2e2e2`
* Eyes (basic): `#823818`, `#c28202`
* Eyes (upgrade): `#ad3721`, `#f88d57`
* Eyes (soothsayer): `#f93c3c`, `#ffafaf` 
* Shield (gold): `#512c00`, `#ac6e17`, `#deaa62`, `#ffe6c3`
* Shield (metal): `#263034`, `#47515d`, `#5f6b6b`, `#9aabb4`

There are a few issues with the current palette:

- Ambusher's and flanker's shields were painted after the original sprites were designed. They do not share any colors with the rest of the sprites, and are rather inconsistent with the existing similar colors (e.g., ambusher's shield and robe grays, flanker's shield and gold trims).
- The oracle's robe colors has very similar to the metal/armor palette up to the point that they could be merged.
- The palette is rather dark in general. The sprites unfortunately stand out when next to some newer graphics.
- Some colors are oversaturated, particularly the skin and skirmisher tree spear shafts. This isn't a major issue on its own, but combined with the darker palette, it further deviates from the current Wesnoth style.

This is an example of an automatic Saurian recolor that reuses some Drake colors and replaces the spear shaft palette with the augur's tree staff color:

![skirmisher](https://user-images.githubusercontent.com/11707612/147966297-4c8cc828-43db-45da-b18f-df9c40977104.png)
[...]

PR #6912, opened 2022-07-26
Saurian Spearthrower tree

czyzby:
This pull requests is a part of the effort to bring units that are used by mainline campaign candidates (see #6905, #6875). It contains an alternative upgrade of the Saurian Skirmisher from the [Saurian Pack add-on](https://github.com/czyzby/wesnoth-saurian-pack).

The pull request consists of the following changes:

* Saurian Spearthrower unit is added. This level-2 unit is a mixed fighter with a stronger ranged attack designed as an alternative upgrade of the Saurian Skirmisher.
* Saurian Javelineer unit is added as an upgrade of the Saurian Spearthrower. It's a steady upgrade to the Spearthrower, gaining stronger melee and ranged attacks.
* `ENABLE_SAURIAN_SPEARTHROWER` macro is added to allow campaigns and scenarios to turn on the Skirmisher→Spearthrower upgrade.

When compared to the Ambusher and Flanker, these units lack the skirmisher ability and deal 1 less hit in melee, but make up for it with a stronger ranged attack and slightly more HP.

During reviewing the PR in terms of the sprites and balance, please consider that these units are mostly meant for campaigns and UMC. They are _not_ available in the default eras. While the proposed stats should be looked into before merging, these units do not have to be balanced in the context of the mainline drakes faction.


knyghtmare (approved):

data/core/units/saurians/Javelineer.cfg
```diff
@@ -0,0 +1,98 @@
...
+    movement_type=lizard
+    movement=7
+    experience=150
+    level=3
+    alignment=chaotic
+    advances_to=null
+    {AMLA_DEFAULT}
+    cost=47
+    usage=archer
+    description= _ "After years of combat practice, saurians warriors that are the most talented in the skill of spear throwing might advance to the ranks of the javelineers. With better equipment and exceptional aim, saurian javelineers are not to be underestimated."
```
nemaara, line 19: This definitely needs a different description at some point, but is not necessary for this PR.

data/core/units/saurians/Javelineer.cfg
```diff
@@ -0,0 +1,98 @@
...
+    die_sound=hiss-die.wav
+    [attack]
+        name=spear
+        description=_"spear"
+        range=melee
+        type=pierce
+        damage=8
```
nemaara, line 26: I think this unit should focus even more on ranged and less melee. So weaker melee, stronger ranged, say -1 melee dmg and +1 ranged dmg, or however you want to do it.

data/core/units/saurians/Javelineer.cfg
```diff
@@ -0,0 +1,98 @@
...
+    [attack]
+        name=spear
+        description=_"spear"
+        range=ranged
+        type=pierce
+        damage=13
+        number=3
+        icon=attacks/spear-thrown.png
+        [specials]
+            {WEAPON_SPECIAL_MARKSMAN}
```
nemaara, line 39: I'm not sure this fits for saurians, but maybe there's some reasoning for it?

data/core/units/saurians/Spearthrower.cfg
```diff
@@ -0,0 +1,94 @@
...
+    hitpoints=39
+    movement_type=lizard
+    movement=7
+    experience=55
+    level=2
+    alignment=chaotic
+    advances_to=Saurian Javelineer
+    cost=23
+    usage=mixed fighter
+    description= _ "Some saurian warriors with a particularly good aim prefer to aid their allies from the back lines rather than to infiltrate the enemy formations. Their deadly spear throwing skills allow them to support other saurian warriors from afar."
```
nemaara, line 18:
```suggestion
    description= _ "Some saurians  with particularly good aim prefer to aid their allies from the backlines rather than to infiltrate enemy formations. Their deadly spear throwing skills allow them to support other saurian warriors from afar."
```

data/core/units/saurians/Spearthrower.cfg
```diff
@@ -0,0 +1,94 @@
...
+    die_sound=hiss-die.wav
+    [attack]
+        name=spear
+        description=_"spear"
+        range=melee
+        type=pierce
+        damage=6
```
nemaara, line 25: Same as above, I feel like a greater difference between melee and ranged would be better.

nemaara (changes requested):

data/core/units/saurians/Javelineer.cfg
czyzby: OK, I'll leave those as they are unless you want to propose something specific.

data/core/units/saurians/Spearthrower.cfg
czyzby: @nemaara I think that might have some slight issues that we might want to consider:

* Their melee damage will be below the Skirmisher's. Leaving it at 6x3 has a nice parallel with the Ambusher, as the Spearthrower currently deals the same damage per hit with one less strike.
* 9x3 is already above the Oracle's damage. 10x3 will start to rival Drakes, despite a lower experience requirement.

It's not necessarily wrong, I just felt like this unit could be a mixed fighter with a ranged preference that slowly advances further from the original unit archetype into an archer. If you still think the attacks should be changed, I can adjust them as you've proposed and change the unit usage to archer.

data/core/units/saurians/Javelineer.cfg
czyzby: Guess I wanted to make the final upgrade more distinctive. On second thought, I agree it doesn't fit thematically. Do you think 14x3 will be appropriate damage without the ability?

data/core/units/saurians/Javelineer.cfg
czyzby: @nemaara I think 8x3 melee attack draws an interesting parallel to the Flanker, as it deals the same damage per hit with 1 more strike. I can lower it to 7x3, but I think overall it's fitting for a unit that starts with a melee archetype.

As for the ranged damage, I agree it can be raised - especially if the marksman ability should be removed. If it was up to me, I'd probably leave melee damage as is and increase the ranged damage accordingly to 14 or 15, but it's your call.

doofus-01:
Is this the same as the "guardian" in #6905 or is this another branch?

czyzby:
> Is this the same as the "guardian" in https://github.com/wesnoth/wesnoth/pull/6905 or is this another branch?

There are 3 branches and PRs:
* Lv3 Augur upgrades. #6911 
* Spearthrower/Javelineer upgrades for Skirmisher meant for campaigns/UMC (used by WoT among others) - this one, #6912.
* New Guardian upgrade for Skirmisher. Mainline candidate for official eras. #6905

These are the only units that are not meant to be available without the `ENABLE_SAURIAN_SPEARTHROWER` macro.

data/core/units/saurians/Javelineer.cfg
nemaara: Okay that sounds fine to me but maybe @Hejnewar would also want to weigh in.

data/core/units/saurians/Spearthrower.cfg
nemaara: That's fine, leave it then.

data/core/units/saurians/Javelineer.cfg
czyzby: Removed the ability.

data/core/units/saurians/Javelineer.cfg
czyzby: I've changed the attack to 14x3 and removed the marksman ability. Let me know if that works for you.

Hejnewar: I'm trying to catch up, I believe I missed something honestly but okay, this seems okay, usually I simply don't do too much unless someone would try to do something too broken and then I stop that. :P I dont have any strong feelings about this unit so I just leave it for others. 

Actually I will mention one thing: saurians don't seem like the types to throw this strong, I think having more strikes and less damage is more friendly honestly for agility based species.

czyzby: I agree that it fits thematically. I decided to put only a few attacks for consistency with the original skirmisher tree. However, it makes sense for a ranged-focused unit to have even more throws. Current damage of 42 (14x3) is not a multiple of 4 or 5, so I'm not exactly sure what values to choose to maintain a similar power level. Do you think 10x4 is OK, or should I go with another set of values? @Hejnewar @nemaara

Hejnewar: 9-5 will land this unit pretty much in the middle of all the level 3 units when it comes to strength (and will also raise the price to 52), in comparison Flanker is second to last when it comes strength.

czyzby: I've adjusted the damage and cost, let me know if the units stats are acceptable now.

Hejnewar: Yes everything is fine.

czyzby: @nemaara Can you review the stats again?

knyghtmare:
Is this good for merge yet?

nemaara (approved):

knyghtmare:
two approvals...okay, since no objections, I will click merge within 24 hours from now.

czyzby:
I believe @doofus-01 wanted to go through all saurian unit sprites, although in this case these units will technically not make into official factions. Since they are planned to be added to future official campaigns, I think putting them in core along with the `ENABLE_SAURIAN_SPEARTHROWER` might be the best way to go.

doofus-01:
I would like to review these, but since we are still so far away from 1.18 Beta/RC, merging them now is probably fine.  I don't like the idea of having cloned units with small edits, but maybe that doesn't really describe these, let's see how it goes.

This does make me realize that Saurians didn't really get the #6417 treatment, so if you (@czyzby ) could supply an intended colormap, similar to what's in most other unit directories, that would be helpful.  If you intend a unified palette for all saurians, put it in the `images/units/saurians` directory, otherwise just putting it in the individual sub-directory is fine.  The placement of the colors in the map doesn't matter, it's currently _ad hoc_.  Thanks.

czyzby:
@doofus-01 I can provide a palette, I believe I've prepared that in a PR when I was cleaning up saurian sprites. The only thing I'm concerned of is that Ambusher's and Flanker's shields use a completely different palette that has a bit more colors and doesn't look good when recolored automatically. I think they were added or modified after the original sprites were created. Should I put these as well or only include the colors shared among other sprites?

I'll also set up a forum thread with the sprites. Been busy last couple days, but should have more time this week.

czyzby:
That's the issue with the palette: #6453.

doofus-01:
> I can provide a palette, I believe I've prepared that in a PR when I was cleaning up saurian sprites. The only thing I'm concerned of is that Ambusher's and Flanker's shields use a completely different palette that has a bit more colors and doesn't look good when recolored automatically. I think they were added or modified after the original sprites were created. Should I put these as well or only include the colors shared among other sprites?

Adding more colors to the main palette to account for the shields is OK, as long as they aren't too similar to existing colors.  What is "too similar"?  Hard to say, it's mostly a gut-level thing at this point.  Ideally we could define that better, but for now just use your best judgement.  We can do this in iterations.

czyzby:
I've tried converting the shields to existing grays and yellows, but there's too few colors to convert them all properly and honestly even when I left an extra one it didn't look very good. A distinct palette makes the shields pop out. I'm just wondering whether these colors should be acknowledged and added to the palette for all saurians, or if we want to keep them as extras for the skirmisher tree.

Jonathan-Kelly:
Has letting this advancement keep the `skirmisher` ability been considered? It seems odd that 2 of 3 advancements for a skirmisher unit (with 'skirmisher' in its name) loses the `skirmisher` ability. It would also look less weird for the Spearthrower to have no shield if it was less durable than the Ambusher (it is actually more durable as it is now) while being just as agile.

Is there any balance reason against a ranged attacker having the `skirmisher` ability?

czyzby:
Well, the Spearthrower sprite certainly looks more nimble than the Ambusher. I wouldn't be against another skirmisher unit, but it would probably call for a nerf of HP and damage output.

Jonathan-Kelly:
I agree. What do you think of the following changes:

Spearthrower
ability `skirmisher`
hitpoints 39 -> 32
melee 6x3 - - 6x3
range 9x3 -> 7x3

Javelineer
ability `skirmisher`
hitpoints 50 -> 43
melee 8x3 -> 7x3
range 9x5 -> 8x4

Alternatively, we could go full into the ranged attack specialization at the Spearthrower level, since the player made an explicit choice to go with this over the Ambusher:

Spearthrower
ability `skirmisher`
hitpoints 39 -> 32
melee 6x3 -> 6x2
range 9x3 -> 7x4

Javelineer
ability `skirmisher`
hitpoints 50 -> 43
melee 8x3 -> 7x2
range 9x5 ->10x4

Note: for damage `7` is a magic number because at the right ToD it gets a +2 boost, whereas `6` or less damage gets only a +1

czyzby:
Since it's mostly a campaign unit, I think the campaign authors should decide whether the unit would benefit from being a skirmisher or not. Especially since the stats were already approved. @knyghtmare

My concern is that the Spearthrower with the skirmisher ability will feel very samey to the Ambusher - not only because of the similar sprites, but now also the abilities. Do we really want to introduce an alternative unit upgrade that basically comes down to melee and ranged attacks being switched up? Thematically it might make sense, but I mechanically I think a more durable unit with stronger attacks will open up more options to the players.

We could hold off the merge until the final Guard's stats are chosen (#6905). That way we could adjust this unit to serve as a viable alternative. For example, if the Guard would be accepted without the first strike ability, I'd like the Spearthrower to have it even at the cost of making its melee attack weaker.

Jonathan-Kelly:
> My concern is that the Spearthrower with the skirmisher ability will feel very samey to the Ambusher - not only because of the similar sprites, but now also the abilities.

Well I mean, without a shield it looks lighter and more like a skirmisher than the Ambusher does. When testing it as part of WoF, it took me a while to realize it lacked the skirmisher ability (and was more durable). And then realizing this, it seemed the augur lines were the better option as a cost effective archer (with magic) in most circumstances.

> We could hold off the merge until the final Guard's stats are chosen

Yeah, that seems like the best way to go.

Elvenisko:
in my option.., it is a good idea to have more saurians. with or without skirmisher ability. in my faction, the southern swamplords, i have 3 branches on level 1. Saurian dark mages, saurian skirmishers and saurian fighters.

So, my fighter is a stronger version of skirmisher, without the skirmisher abiity of course.

So... I have an idea to make not 2, but three units. Leave mages and skirmishers totally unchanged, but make a new saurian line from level 1. For example:

Level 1: Saurian Fighter
Level 2: Saurian Spearthrower
Level 3: Saurian Javelineer

Differences between level 1 and 2 might be in fighting style. While Figher can be a balanced melee/ranged unit, spearthrower might be improved only in HPs and ranged attack.

czyzby:
I think a lot of players would love to see more Saurian units outside of the base 2. However, this is not the goal of this PR - or the related ones, for that matter.

Just to recap:

* Originally the proposed units were an optional, unofficial [add-on](https://github.com/czyzby/wesnoth-saurian-pack) compatible with the default era that you could turn on for multiplayer games and the campaigns.
* Since all of the units had complete animations, they made it into various campaigns including some mainline candidates.
* These units were submitted in one of the campaign-related PRs to be included as-is throughout the campaign resources.  #6875
* However, I noticed that the animation frames are outdated and unoptimized (without the latest visual fixes and palette adjusments), so I proposed to make a separate PR with the latest versions.
* I though it would be a good idea to add the units into the core instead of individual campaigns, so that they can be easily used in eras and campaigns without having to copy the assets.
* The discussion on the spearthrower tree hijacked the entire thread. I ended up splitting into several separate PRs:
  * #6911: Augur level-3 upgrades.
  * #6912: Ranged-based skirmisher upgrade for campaigns under an optional flag, so that the sprites are not wasted. (You are here.)
  * #6905: A new more defensive/mixed skirmisher upgrade with the goal of making it into the mainline. When (and whether) it will be added is another question. 
* I'm still waiting on review of the sprites from @doofus-01. I'd like to tackle the Augur sprites first, since they seem to require the least number of changes. Still, I don't want to start animating (or rather, correcting animation frames) without their approval on the base sprites first.

My goal was never to add entirely new units, and at this point I don't think we need more ideas on unit stats/archetypes/etc.

Pentarctagon:
Given doofus-01's previous statement of
> I would like to review these, but since we are still so far away from 1.18 Beta/RC, merging them now is probably fine.

does that block this PR from getting merged?

czyzby:
I think the statistics were already reviewed and approved. The sprite itself is a bit samey when compared to the Ambusher line and could probably use a redesign, but this unit is mostly there to provide some variety in campaigns and UMC - it will not be available in the mainline factions by default.

Pentarctagon:
@Hejnewar any further comments? If not, can you mark your review as approved?

Hejnewar (approved):

Pentarctagon:
Alright, so barring anything else being said, I'll go ahead and merge this in a couple days.

doofus-01:
I agree, go ahead and merge this so it is easier to test.  We can clean up anything that needs it later.

data/core/units/saurians/Javelineer.cfg
```diff
@@ -0,0 +1,95 @@
...
+    movement_type=lizard
+    movement=7
+    experience=150
+    level=3
+    alignment=chaotic
+    advances_to=null
+    {AMLA_DEFAULT}
+    cost=52
+    usage=archer
+    description= _ "After years of combat practice, saurians warriors that are the most talented in the skill of spear throwing might advance to the ranks of the javelineers. With better equipment and exceptional aim, saurian javelineers are not to be underestimated."
```
Wedge009, line 19: I understand this is just a place-holder, but is 'saurians warriors' supposed to be 'saurian warriors'?

data/core/units/saurians/Spearthrower.cfg
```diff
@@ -0,0 +1,94 @@
...
+    hitpoints=39
+    movement_type=lizard
+    movement=7
+    experience=55
+    level=2
+    alignment=chaotic
+    advances_to=Saurian Javelineer
+    cost=23
+    usage=mixed fighter
+    description= _ "Some saurians  with particularly good aim prefer to aid their allies from the backlines rather than to infiltrate enemy formations. Their deadly spear throwing skills allow them to support other saurian warriors from afar."
```
Wedge009, line 18: Double space after 'saurians', plus 'backlines' should probably 'rear' (I understand English may not be a native language, plus these are supposed to be rewritten anyway).

data/core/units/saurians/Javelineer.cfg
czyzby: That's a typo, although the entire text is a placeholder.

Issue #6907, opened 2022-07-25
`FADE_TO_BLACK` macro should probably use the new `screen_fade` system


mesilliac:
I added a smooth screen fading tag in #6864 which should be used for screen fades over the previous method of animating the `color_adjust` tag. However i haven't yet converted the macros `FADE_TO_BLACK` etc that are usually used do screen fading.

There is a small question of whether to override the existing macro to use the new system, or whether to make a new macro for the new system and keep the old macro for the old behaviour (animating `color_adjust`).

There is a visible difference, in that the old behaviour did not fade units and halos, only terrain.

The new behaviour fades everything, and so better fits the name, but if there's any scenario relying on being able to see units after doing the fade, it will break.

I'm somewhat in favour of overriding the behaviour, as it more accurately does what it says on the tin (fades to black). Some sort of `FADE_TO_BLACK_LEGACY` or `FADE_TERRAIN_TO_BLACK` could replace the old macro. But i don't know how to notify anyone possibly relying on that functionality that it has changed.

CelticMinstrel:
I'm a bit unsure about this too. First, I'm not sure whether or not it's okay to just change the behaviour of the macro. It's a relatively minor behaviour change, so it _might_ be okay, I'm just not really sure. Second, I'm not quite sure what's the best way to notify people of the change. The deprecation system is the obvious choice, but that would require deprecating the old macro (with no behaviour change presumably, since it's deprecated) and probably adding two replacement macros for the two variations in behaviour, which seems a bit less than ideal…

mesilliac:
There's also a possible future distinction between fading just the game map portion of the screen and fading the UI as well. I had planned to add an option to the `screen_fade` tag but it was nontrivial.

Perhaps adding new explicit macros and deprecating the old one is the best choice. The old macro could later be repurposed to whichever is most useful, after it passes whatever length of time is appropriate for the deprecation.

CelticMinstrel:
That's true – there's nothing preventing us from re-using a deprecated macro for some other (related) purpose instead of removing it outright.

soliton-:
If there is a simple tag now I'd say there is no need for a macro at all. Leave FADE_TO_BLACK as it is. Deprecate it if the new tag can do that same. No need for a macro if all it hides is `[screen_fade] duration=100 [/screen_fade]` or whatever.

mesilliac:
[...]

PR #6913, opened 2022-07-27
Update mainline campaigns to use screen_fade

mesilliac:
According to usage, i ended up adding the macros:

 * `SCREEN_FADE_OUT` - replaces `FADE_TO_BLACK` in most cases
 * `SCREEN_FADE_IN` - replaces `FADE_IN` in most cases
 * `SCREEN_FADE r g b duration` - convenience wrapper for `[screen_fade]` with 255 alpha. Mostly used internally.
 * `SCREEN_UNFADE duration` - convenience wrapper for `[screen_fade]` with 0 alpha. Mostly used internally.
 * `FLASH r g b actionWML` - like `FLASH_WHITE` etc but takes a colour
 * `FLASH_LIGHTNING actionWML` - flashes an appropriate colour and plays `lightning.ogg`

The `FLASH_WHITE`/`RED`/`GREEN`/`BLUE` macros were updated to use `screen_fade`.

All mainline campaigns were updated to use these or `[screen_fade]` over `[color_adjust]`, as appropriate. There are only a couple remaining uses of `color_adjust` that weren't obvious to switch, and nothing needed the old `FADE_TO_BLACK` behaviour.

### DIFF
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter4/16_The_Chief_Must_Die.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter4/16_The_Chief_Must_Die.cfg
@@ -673,21 +673,7 @@
             name=potion.ogg
         [/sound]
 
-        [color_adjust]
-            red=100
-            green=100
-            blue=100
-        [/color_adjust]
-
-        [delay]
-            time=10
-        [/delay]
-
-        [color_adjust]
-            red=0
-            green=0
-            blue=0
-        [/color_adjust]
+        {FLASH_WHITE ()}
 
         [delay]
             time=2000
--- a/data/campaigns/Liberty/scenarios/04_Unlawful_Orders.cfg
+++ b/data/campaigns/Liberty/scenarios/04_Unlawful_Orders.cfg
@@ -535,19 +535,7 @@
             [/variable]
 
             [then]
-                [color_adjust]
-                    red=255
-                    green=100
-                    blue=100
-                [/color_adjust]
-                [delay]
-                    time=10
-                [/delay]
-                [color_adjust]
-                    red=0
-                    green=0
-                    blue=0
-                [/color_adjust]
+                {FLASH 255 100 100 ()}
 
                 [terrain_mask]
                     x,y=16,33
@@ -661,19 +649,7 @@
             [/variable]
 
             [then]
-                [color_adjust]
-                    red=255
-                    green=255
-                    blue=100
-                [/color_adjust]
-                [delay]
-                    time=10
-                [/delay]
-                [color_adjust]
-                    red=0
-                    green=0
-                    blue=0
-                [/color_adjust]
+                {FLASH 255 255 100 ()}
 
                 [terrain_mask]
                     x,y=16,33
--- a/data/campaigns/Secrets_of_the_Ancients/scenarios/17_Mortality.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/scenarios/17_Mortality.cfg
@@ -322,11 +322,18 @@ Not far inside the cave was a ruined castle built in a style I did not recognize
             flag=sacrifice
         [/animate_unit]
 
-        [hide_unit]
-            # All units.
-        [/hide_unit]
+        # TODO: The skeleton should keep the final animation frame here,
+        # rather than flipping back to the standing animation for the fade.
 
-        {FADE_TO_BLACK_HOLD 1000}
+        [screen_fade]
+            red,green,blue=0,0,0
+            alpha=255
+            duration=1000
+        [/screen_fade]
+
+        [delay]
+            time=2000
+        [/delay]
 
         [fire_event]
             name=lichify  # See: sota-utils.cfg
--- a/data/campaigns/Secrets_of_the_Ancients/units/Skele_Sacrifice.cfg
+++ b/data/campaigns/Secrets_of_the_Ancients/units/Skele_Sacrifice.cfg
@@ -22,7 +22,7 @@
             layer=100
         [/frame]
         [frame]
-            image="units/undead/skeleton-se-sacrifice7.png:1000"
+            image="units/undead/skeleton-se-sacrifice7.png:500"
             x=33
             y=0
             layer=100
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/05_A_Subterranean_Struggle.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/05_A_Subterranean_Struggle.cfg
@@ -1489,27 +1489,11 @@
                     message= _ "Burn, burn and die!"
                 [/message]
 
-                [color_adjust]
-                    red,green,blue=255,0,0
-                [/color_adjust]
-
-                [redraw]
-                [/redraw]
-
-                [sound]
-                    name=fire.wav
-                [/sound]
-
-                [delay]
-                    time=100
-                [/delay]
-
-                [color_adjust]
-                    red,green,blue=0,0,0
-                [/color_adjust]
-
-                [redraw]
-                [/redraw]
+                {FLASH_RED (
+                    [sound]
+                        name=fire.wav
+                    [/sound]
+                )}
 
                 [message]
                     speaker=Dwarf Leader
@@ -1681,27 +1665,11 @@
                     message= _ "Let’s blast those monsters back to the pits they spawned from! Fire in the hole!"
                 [/message]
 
-                [color_adjust]
-                    red,green,blue=255,0,0
-                [/color_adjust]
-
-                [redraw]
-                [/redraw]
-
-                [sound]
-                    name=thunderstick.ogg
-                [/sound]
-
-                [delay]
-                    time=100
-                [/delay]
-
-                [color_adjust]
-                    red,green,blue=0,0,0
-                [/color_adjust]
-
-                [redraw]
-                [/redraw]
+                {FLASH_RED (
+                    [sound]
+                        name=thunderstick.ogg
+                    [/sound]
+                )}
 
                 [message]
                     speaker=Troll Leader
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/06a_In_the_Tunnels_of_Trolls.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/06a_In_the_Tunnels_of_Trolls.cfg
@@ -503,23 +503,12 @@
 
             # Have screen flash red
 
-            [color_adjust]
-                red,green,blue=255,0,0
-            [/color_adjust]
-
-            [redraw]
-            [/redraw]
-
-            [delay]
-                time=1000
-            [/delay]
-
-            [color_adjust]
-                red,green,blue=0,0,0
-            [/color_adjust]
-
-            [redraw]
-            [/redraw]
+            # TODO: should the explosion and shaking be inside the flash?
+            {FLASH_RED (
+                [delay]
+                    time=1000
+                [/delay]
+            )}
 
             [sound]
                 name=explosion.ogg
@@ -913,23 +902,8 @@
 
         # Have screen flash red
 
-        [color_adjust]
-            red,green,blue=255,0,0
-        [/color_adjust]
-
-        [redraw]
-        [/redraw]
-
-        [delay]
-            time=200
-        [/delay]
-
-        [color_adjust]
-            red,green,blue=0,0,0
-        [/color_adjust]
-
-        [redraw]
-        [/redraw]
+        # TODO: should the explosion and shaking be inside the flash?
+        {FLASH_RED ()}
 
         [sound]
             name=flame-big.ogg
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
@@ -522,27 +522,11 @@
 
             # Have screen flash red
 
-            [color_adjust]
-                red,green,blue=255,0,0
-            [/color_adjust]
-
-            [redraw]
-            [/redraw]
-
-            [sound]
-                name=explosion.ogg
-            [/sound]
-
-            [delay]
-                time=100
-            [/delay]
-
-            [color_adjust]
-                red,green,blue=0,0,0
-            [/color_adjust]
-
-            [redraw]
-            [/redraw]
+            {FLASH_RED (
+                [sound]
+                    name=explosion.ogg
+                [/sound]
+            )}
 
             # Have wall shake, then be replaced by dirt, then add rubble
 
@@ -1791,24 +1775,16 @@
                 message= _ "Hey, wait a minute, that amulet glows the same color as this rune. Maybe if I put the amulet on and step into the rune..."
             [/message]
 
-            [color_adjust]
-                red,green,blue=246,0,243
-            [/color_adjust]
+            {FLASH 246 0 243 (
+                {UTBS_SHAKE_SCREEN}
 
-            {UTBS_SHAKE_SCREEN}
+                [terrain]
+                    x,y=46,21
+                    terrain=Uu
+                [/terrain]
 
-            [terrain]
-                x,y=46,21
-                terrain=Uu
-            [/terrain]
-
-            {PLACE_IMAGE scenery/rubble.png 46 21}
-
-            [redraw][/redraw]
-
-            [color_adjust]
-                red,green,blue=0,0,0
-            [/color_adjust]
+                {PLACE_IMAGE scenery/rubble.png 46 21}
+            )}
 
             [remove_item]
                 x,y=45,21
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/08_Out_of_the_Frying_Pan.cfg
@@ -1287,27 +1287,11 @@
             [/not]
         [/filter]
 
-        [color_adjust]
-            red,green,blue=255,0,0
-        [/color_adjust]
-
-        [redraw]
-        [/redraw]
-
-        [sound]
-            name=magic-dark-big-miss.ogg
-        [/sound]
-
-        [delay]
-            time=100
-        [/delay]
-
-        [color_adjust]
-            red,green,blue=0,0,0
-        [/color_adjust]
-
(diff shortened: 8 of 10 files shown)


max-torch:
In #6907 @mesilliac said
> There is a visible difference, in that the old behaviour did not fade units and halos, only terrain.
>The new behaviour fades everything, and so better fits the name

There is a possible improvement that can be made in addition to the changes here. 
There is some code that is no longer needed as a consequence of the new macro behavior. For example, in the final scenario of SoF, the WML `kill` tag is used because the old fade behavior did not fade the units. The WML that kills the units during the fade to black could possibly be removed as it is now redundant with the macro. Maybe there are other cases like this in the other campaign scenarios.

data/core/macros/interface-utils.cfg
```diff
@@ -201,50 +201,78 @@
...
+
+#define SCREEN_FADE_OUT
+    {SCREEN_FADE 0 0 0 500}
+#enddef
+
+#define SCREEN_FADE_IN
+    {SCREEN_UNFADE 500}
+#enddef
+
+# Flashes the screen with a given color performing ACTION_WML.
```
CelticMinstrel, line 229: This documentation comment is not parsed out by wmlscope for whatever reason. Either it needs to be inside the `#define` or the blank line following it needs to be removed.

This is not new in this PR – it was already the case. So you don't need to change it, unless you want to.

CelticMinstrel:
I can see a number of "TODO" comments for code that needs review, so I guess that needs someone to take a look at it before merging.

Does it make more sense to split this up into one commit per campaign? (And one for the core files.)

data/core/macros/interface-utils.cfg
mesilliac: I also found the scope of this comment unclear. It's referring to several macros in bulk so it's not really directly attached to any of them. Maybe it should just be removed? The individual defines already have comments inside them.

mesilliac:
The TODOs are somewhat optional to address as the screen_fade replacements should look okay. But it would be good if someone who understands the original intention of the effect in each case could look at them and make sure it still makes sense.

Some of the effects could probably be significantly improved as well, if someone spent some time tweaking them.

I can break it up per campaign if that would make things easier. I squashed everything as it evolved several times while i was doing it, and most of the changes were similar or related.

mesilliac:
> There is some code that is no longer needed as a consequence of the new macro behavior. For example, in the final scenario of SoF, the WML kill tag is used because the old fade behavior did not fade the units
I did remove a few things like this. I found some `hide_unit` that were unnecessary for example. I didn't think of the kill tags tho. I'm not sure i know enough about WML to know if one is important or not. So that kill tag before the screen fade at the end in SoF:9 can just be removed?

max-torch:
> > There is some code that is no longer needed as a consequence of the new macro behavior. For example, in the final scenario of SoF, the WML kill tag is used because the old fade behavior did not fade the units
> 
> I did remove a few things like this. I found some `hide_unit` that were unnecessary for example. I didn't think of the kill tags tho. I'm not sure i know enough about WML to know if one is important or not. So that kill tag before the screen fade at the end in SoF:9 can just be removed?

When I look at the diff of this PR, I don't see the changes where you say you remove `hide_unit` tags. Not unless you're referring to other scenarios where you removed `hide_unit` tags?


I tested this PR on the final scenario of SoF before and after removing the following lines:
https://github.com/wesnoth/wesnoth/blob/1be1ba428251647031cd72ceed173f59ad1a0fd2/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg#L947-L951

I observed no difference in how the end event played out, units that were supposed to be gone after the last fade to black properly disappeared. I guess this means it's not just `hide_unit` but also if scenarios have `kill` that there could be redundant code that can be pruned.

*to make sure that `cfg` changes are in effect, the game has to be exited and reloaded correct? That's what I did to make sure the game was loading the old and edited scenario `cfg`.  Would be great if you can also verify (your optional discretion) that there is no difference before finalizing, since I have linked the lines removed above.


One thing I noticed was that when watching the replay, after the `SCREEN_FADE_OUT` occurs, when you click on "Reset to Beginning", the scenario resets but the map/screen is not unfaded/restored. I guess the proper behavior is for the map to be shown again in this case.

CelticMinstrel:
> So that kill tag before the screen fade at the end in SoF:9 can just be removed?

I would say removing kill tags is probably only okay if it's the final scenario, as it would otherwise affect the units that are carried over to the next scenario.

mesilliac:
There is an Epilogue scenario after SoF:9 but that's all. It's probably fine to remove the kill tag there but i'm not confident to do it myself.

> When I look at the diff of this PR, I don't see the changes where you say you remove hide_unit tags.

There was one in SoTA:17, THoT:10, WoF:08... i think that's it. I also removed a lot of `[redraw]` tags as that should no longer be necessary, as i made sure `screen_fade` ensures a full redraw after every frame, even if it does nothing. Oh and there was also a `[remove_item]` in THoT:10 which i removed, as it was also the end of the campaign and i assume the item was removed for the same reason as the units.

> I didn't test this properly yet. I checked out the PR but was running the game from an installation directory and not from the checked out directory.

Oh no :laughing:. The difference should be obvious in the tutorial after the first move, checking out Delfador's lightning flash when he summons the Quintain. Previously it was a single white frame, but now it shoud be a smooth ramp up and down for the flash. So that should be an easy test to make sure it's working.

mesilliac:
> One thing I noticed was that when watching the replay, after the SCREEN_FADE_OUT occurs, when you click on "Reset to Beginning", the scenario resets but the map/screen is not unfaded/restored. My expected behavior is for the map to be shown again in this case.

Funnily enough this one is a `color_adjust` bug. I tested if in SoF:9 and now the screen does get unfaded... but everything is red because i left that one as `color_adjust`. I wasn't sure whether the effect would match the intention if i used `screen_fade` for the red part.

max-torch:
I'm confident now that I tested the scenario properly. (I swear). I have just edited my previous comment to have the right information.

Regarding `reset to beginning`:
I can confirm that when `FADE_TO_BLACK` was used, after resetting, the map would remain black.
When `SCREEN_FADE_OUT` was used, after resetting, the map would remain red.

Either way, both are not the expected behavior. The expected behavior is that the map is restored.

mesilliac:
I set it to reset the color_adjust at the end, but it might be better to replace it with a screen_fade. I'm not sure of the intention of the effect, so i didn't replace it. Specifically i don't know if it should be visible over units or not, or if it's supposed to be a flash effect how quick it should be.

If it's just a flash effect it can be replaced with a screen fade to red before the fade to black. However it would still need an informed decision for the duration of the fade. I'm not sure if it's supposed to be instantaneous or gradual.

mesilliac:
Hmm, so i split this up by campaign to make the changes clearer. Several could use verification, from least serious to most:

* DW:10 - effect when Caladon dies needs verification that it looks okay
* EI:17b - effect when liches die needs verification that it looks okay
* HttT:24 - effect during "last breath" event needs verification that it looks okay - i did tweak this effect independently so i think it's okay, but i haven't tested it in-game
* UtBS - there were a lot of effects i changed to screen_fade. Most of them were simple flash effects and were replaced with the relevent FLASH macros, but some were custom effects using color_adjust that need verification for correctness after conversion, both in colour and in timing. It is probably also possible now to improve on the effects using the new system.

I included the SoF fixes. It could be tweaked to use screen_fade for the red but it's not critical, and would need an artistic decision on timing.

The others are mostly just converting things to use `FLASH_LIGHTNING` or `SCREEN_FADE_OUT`. The lightning flash can also maybe use tweaking, but that can now be done by tweaking the macro.

max-torch:
> I also still haven't come up with a fix for the animation frame reset during fades (SoTA:17, Tutorial). I don't know enough WML to know the right way to deal with this.
Can I ask what the issue is in the tutorial? I noticed the SoTA one but I didn't know there also was one in the tutorial.

max-torch:
I made a report that discusses the findings of my playtesting of the PR. 
https://docs.google.com/document/d/1HP-WAH2Ejvlz5qw-1wcX-CIs58J7KcZOut5DNYn35DY/edit?usp=sharing
Anyone with the link can edit and comment.

The current state of the report covers almost all the scenarios. What is missing is the scenarios from WoF and LoW, and half of the scenarios from UtBS.

I am not planning to playtest the WoF and LoW scenarios, for the reason that I haven't completed them yet, and am requesting someone else to playtest this PR for the two scenarios from those two campaigns.

When I'm done with the remaining UtBS scenarios I will make another comment on this thread.

mesilliac:
Hmm, so re: the coloured boxes around units and the river, that should have been fixed by a recent commit which should be in master - d0f9c6b8e3a628ccc488717878066cf143b15d0f - do you have the most current version of this PR? It should have been rebased to include that commit when i rearranged it so might need a force pull (the actual PR content here should all be the same but i rebased it when i rearranged the commits). This *should* cover all the cases of coloured squares appearing, but it's possible something else is going on somewhere that i didn't catch with that.

Re the brightness of the lightning effect, that can be tweaked by changing the alpha in the `FLASH_LIGHTNING` macro. It should be changed so it works there, rather than adding it as a parameter. I tried setting it to something similar to the previous effect but - on average it had less alpha, because the new effect doesn't spend all its time at the peak, but rather spends most of it ramping up and down. So it needs to be matched for the full effect, not just the peak value. I did try a much lower alpha to try to match the peak value, but it came out way too weak for a lightning flash effect.

I'll get to the other points when i have some time but first making sure you have the full current version is important. Note that that was a change in C++ so it should have updated even on old replays using screen_fade. And it shouldn't change the other points, just fix the bug with coloured boxes, but it does slightly change the peak brightness value so it might make a difference when tweaking the final value.

max-torch:
I'm in the midst of playtesting but I just want to make this comment before I forget.

In most cases, the new fade effect seems to actually be more seizure-friendly than the original, because the original color flashes have no smooth transition and are more sudden, while the new effect has continuous transition.

The only case where the new effect seems to have become more intense is the tutorial.

mesilliac:
I'll try making that one less strong, IIRC it was using a less strong colour than most of the other effects.

max-torch:
For the benefit of contextualizing this discussion: At this point, my cpp files have now been recompiled and mesiliac and I are already reproducing the same results.

Everything has now been playtested except for the scenario from LoW and from WoF, which I don't want to test because I haven't played those yet.
The findings are in the same link as the Google Doc I shared earlier in the thread.

mesilliac:
@max-torch i tweaked the lightning macro to make the effect much weaker, with a similar peak brightness to the old effect. I also tweaked the timing a lot so it flashes in quickly, then lingers, then fades out slowly. Can you check the updated effect (just in the first move of the tutorial is fine) to see if it looks better now? (ignore Delfador's flailing arms)

This lightning macro change will also affect DM:07 and HttT:14 as they all use the same macro now.

mesilliac:
The new timing might also work for Liberty. For that i assumed it was a transformation effect. If it's a "surprise!" effect it should indeed come in quicker, but i don't see anything in the comments about "surprise" like you mention in the doc?

I also thought it would be better if the screen moved to the enemy units at that point, but i don't know how to make it do so.

Re the idea of adding an alpha value to the `FLASH` macro, i left it out because at that point you might as well just use the tag directly. I figure the macro is for a cheap effect that should generally work, and more specific stuff can be tailored. So for Liberty it can be changed to do the effect manually to get the desired effect.

It might also be worth tweaking the FLASH macro so it comes in quicker in general, but it might depend on usage in other places.

mesilliac:
I've tweaked the SoF:9 fade to make the red colour a screen_fade, and i gave it a fairly long duration. From what i can see this should give a better effect, as it seems to just be a sort of "fading out everyone dies" effect?

Blerg, spoilers is the other reason i don't want to spend too much time looking at these...

mesilliac:
I've tweaked the effect in DW:10 - it's using a somewhat complex algorithm for the fade so i'm not sure i can just convert it to a simple fade. But i changed it from modifying the colour to modifying the alpha - i think this should be more in line with the original effect. This is another one i don't really want to playtest because i haven't played DW yet and it's near the top of my to-play list.

It probably could be replaced with a three-part screen fade, but it would need a lot of repeated tweaking and testing to get right and i'm not sure of the exact intention without having played it. So hopefully this fixes it acceptably.

max-torch:
> The new timing might also work for Liberty. For that i assumed it was a transformation effect. If it's a "surprise!" effect it should indeed come in quicker, but i don't see anything in the comments about "surprise" like you mention in the doc?

On a second look at the scenario cfg, I also cannot find anything that mentions that it is supposed to be a "surprise!" effect.

max-torch:
> @max-torch i tweaked the lightning macro to make the effect much weaker, with a similar peak brightness to the old effect. I also tweaked the timing a lot so it flashes in quickly, then lingers, then fades out slowly. Can you check the updated effect (just in the first move of the tutorial is fine) to see if it looks better now? (ignore Delfador's flailing arms)
> 
> This lightning macro change will also affect DM:07 and HttT:14 as they all use the same macro now.

I think the new effect looks good and pleasant.

>I've tweaked the SoF:9 fade to make the red colour a screen_fade, and i gave it a fairly long duration. From what i can see this should give a better effect, as it seems to just be a sort of "fading out everyone dies" effect?

It looks nicer because the units are now included when shading the screen.

>I've tweaked the effect in DW:10 - it's using a somewhat complex algorithm for the fade so i'm not sure i can just convert it to a >simple fade. But i changed it from modifying the colour to modifying the alpha - i think this should be more in line with the original >effect. This is another one i don't really want to playtest because i haven't played DW yet and it's near the top of my to-play list.>
>
>It probably could be replaced with a three-part screen fade, but it would need a lot of repeated tweaking and testing to get right >and i'm not sure of the exact intention without having played it. So hopefully this fixes it acceptably.

It still looks the same as before. My comment on this is that the original actually looks like a one-part screen fade that has a long duration. Current code consists of two screen fades. I will make another comment after this with my suggested change (which I tested)

max-torch:
Forgive the screenshot but I am having a hard time being precise about my suggested change on GitHub. I tested this for DW10 and I feel like this matches the original more:
image

mesilliac:
I think this might be a bug in screen fade, or in the WML algorithm. It should not be flashing twice, from what i can see. It should be looping 14 times, so changing to a 1s duration should result in a 14 second fade?

It looks like the algorithm there does create some values above 255... maybe these are wrapping to 0 and causing this

mesilliac:
I think this should fix it... it was creating some values > 255 and these must be getting wrapped. I'm not sure how to fix that to not accept incorrect values from WML, so i just changed the algorithm to pull those first two steps out of the loop.

mesilliac:
I also found the white a bit harsh so i made it a tiny bit red... that could be changed

max-torch:
> I think this should fix it...

Just pulled and tested the latest commits. DW S10 looks correct now.

mesilliac:
I think i'll commit the pieces that are confirmed okay. If i got it right these are:

* lightning effect is okay (tutorial, DM, HttT)
* DiD fades are all okay (they should all be the same)
* WoF there was just one instance of fade to black replaced, in the exact same way as DiD
* THoT the tweaked fade to black is okay
* SoF new fade out effect is okay
* DW effect is finally unbugged and now okay
* EI effect is okay
* HttT difficult to verify as old effect doesn't actually work but seems okay

I'll leave for now:
* LoW - this probably needs tweaking i'll check it later
* SoTA - i have an idea for fixing the skeleton
* Liberty - need to go over the effect again, and also check the general flash macro timing
* UtBS - i still have yet to go over this in detail. It might tie in with the flash macro speed as a lot of things now use it

Pentarctagon:
@CelticMinstrel @max-torch Is this something that's ready to be merged, or does it need further changes? Asking in particular since mesilliac hasn't been around recently so this PR is unlikely to get further updates.

CelticMinstrel:
From what I can tell it needs someone to check and potentially tweak things, to make sure the new effect still produces the desired effect for each scenario.

max-torch:
> @CelticMinstrel @max-torch Is this something that's ready to be merged, or does it need further changes? Asking in particular since mesilliac hasn't been around recently so this PR is unlikely to get further updates.

Will comment after inspecting the current state. It will require a rebuild of source code.

Wedge009:
I just checked LoW S16. The `FLASH_WHITE` macro lasts longer than the 10 ms `color_adjust` currently used, but I think it looks nicer.

I didn't see any changes for WoF S8 - perhaps this was merged separately and dropped via re-base?

max-torch:
> I just checked LoW S16. The `FLASH_WHITE` macro lasts longer than the 10 ms `color_adjust` currently used, but I think it looks nicer.
> 
> I didn't see any changes for WoF S8 - perhaps this was merged separately and dropped via re-base?

Thanks very much for checking LoW.

When I checkout the PR there is a commit that makes the change to WoF. I also just now refreshed the git pull to make sure it's the latest. Check commit history of the PR branch. See screenshot:
image

Wedge009:
It appears it's already in master (5b5bdde080d6224bc0f5338226d65ed9ea1f53cb), so outside the scope of this PR.

max-torch:
> It appears it's already in master ([5b5bdde](https://github.com/wesnoth/wesnoth/commit/5b5bdde080d6224bc0f5338226d65ed9ea1f53cb)), so outside the scope of this PR.

I think it is in the scope of this PR? It was part of the first half of this  PR that was already committed to master.
This PR is in a state where 75% of it is merged to master. From what had already been merged, I checked all the changes except for the WoF one. So I cannot comment on the WoF one if it is good.

Wedge009:
What I mean is if it's already in master, it's not in this PR, so for the purposes of decided what to do with this PR, the WoF S8 change is no longer relevant. That's not to say it shouldn't be reviewed, but if there are any issues with it, it'd be a separate issue to this PR.

mesilliac:
To confirm, the majority of this including all that was confirmed okay was merged already. I updated the PR branch so it should be clean in that the remaining commits in this branch are those that are unreviewed or unsuitable still for one reason or another. I also had the problem of not wanting to spoil campaigns so several of them (notably UtBS) i wasn't able to test well at all.

The SotA skeleton thing is still a problem. I didn't know enough WML wizardry to fix the animation. I think there's a solution, like changing the standing animation to be a chop then doing the fade while the "standing" chop animation plays, but i didn't know how to do it. The problem currently (also i think a problem with delfador's arm lifting animation in the tutorial) is that animations played from WML block processing including fades, but fades no longer block animation, so they reset to the base animation frame during the fade which is very much undesirable here.

The other effects in this PR definitely need checking and work to make better. But if someone wants to pick them up it's really not difficult to tweak the fade effects. It's just a matter of tweaking the WML and testing to make sure the effect is good/right.

PR #6914, opened 2022-07-28
[UtBS] Add series of bat screeches in the Zochtanol Isle scenario

### LINKED ISSUE #6503: [UtBS] Play the bat unit sound during dialog in S11 Zocthanol Isle
In S11 of UtBS there is this piece of dialog:
image
It would add more flavor to the scene if we make a bat sound play at the same time. The screeching sound Kaleh is talking about is being made by bats, and we do have default bat sound files in core/sounds:
bat-hit-1.ogg
bat-hit-2.ogg
bat-hit-3.ogg
bat-hit-4.ogg
I propose making several of these play one after the other to show that it is an entire host of bats that awaits in the forest. Play each bat-hit sound file in succession.

max-torch:
Resolves #6503. I tested this both with and without music enabled to the same level as the sound effects.

At worst, the player might not notice if music is on and louder than sound effects, but because the bat sound repeats 7 times it is more likely the player will notice. It also represents the fact that there are multiple bats that await.

At best, if the player falls under the demographic of players that like to play with music off, they will definitely hear the sound.

Requesting permission to merge.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/11_Battle_for_Zocthanol_Isle.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/11_Battle_for_Zocthanol_Isle.cfg
@@ -1436,6 +1436,10 @@
     # at first dusk, Kaleh comments strange noises in jungle
     [event]
         name=bat_noise_reaction
+        [sound]
+            name={SOUND_LIST:BAT_HIT}
+            repeat=6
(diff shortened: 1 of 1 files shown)


knyghtmare (approved):

knyghtmare:
All checks passed so you can merge it if you want or wait a bit if any comments come up. Your choice.

PR #6915, opened 2022-07-28
[UTBS] Remove some fungus from S6

max-torch:
Resolves one of the subitems of #6197.
Rationale for this PR is in the Issue thread.
Nemaara tagged as reviewer because they originally added the fungus. (git blame)
Knyghtmare because i believe he has a vision and focus for UtBS.


knyghtmare (approved):

nemaara (approved):
Well I actually did add it purposely as a bottleneck. But I can see it being a bit frustrating so sure we can make this change.

max-torch:
> Well I actually did add it purposely as a bottleneck. But I can see it being a bit frustrating so sure we can make this change.

It had a significant enough effect on scenario difficulty balance towards being more difficult.

PR #6920, opened 2022-07-29
Utbs s6 add scroll to detonation event

max-torch:
One of the subitems of #6197. When Jorgin dies, the map scrolls to the place where the detonation will occur. Some shroud is removed so that the player can see the event unfold.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
@@ -1240,6 +1240,23 @@
             message= _ "I couldn’t do it. Blow the backup charges! If we can’t stop them then maybe the black lake will."
         [/message]
 
+        {SCROLL_TO 12 20}
+
+        [remove_shroud]
+            side=1
+            x=11-15
+            y=19-23
+        [/remove_shroud]
+
+        [delay]
+            time=1200
+        [/delay]
+
+        [place_shroud]
+            side=1
+            x,y=11,19
+        [/place_shroud]
+
         {BACKUP_CHARGES}
 
         [remove_event]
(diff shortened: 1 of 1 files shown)


data/campaigns/Under_the_Burning_Suns/scenarios/06b_In_the_Domain_of_Dwarves.cfg
```diff
@@ -1123,6 +1131,14 @@
                     time=1000
                 [/delay]
 
+                [sound]
+                    name=thunderstick-miss.ogg
+                [/sound]
+
+                [delay]
+                    time=1800
```
nemaara, line 1139: Hmm that's a looong delay.

max-torch: Have you tested this in the scenario? I actually fine tuned that value.

nemaara: Nope, but I've made a gazillion cutscenes before. Longest delay I ever used was 1500 and that was already for a crazy long pause. I usually don't go above 1000 personally. Players aren't that patient and longer delays make them think that the game is frozen lol.

max-torch: ... There is something going on in that 1800. The duration is that length to accommodate the duration of the sound being played. It is not an empty 1800 pause. Does that make sense? It sounds good when you try it, I actually tried 1000, then 1500, then finally 1800, which was perfect.

Wedge009: If it makes any difference, the thunderstick mis-fire sound effect is 1400 ms.

knyghtmare (approved):

max-torch:
If there is no objection, I'll merge this in 4 days.

nemaara (approved):

Issue #6919, opened 2022-07-29
Feature request: "Watery chasm" terrain


cooljeanius:
### Describe the desired feature

Right now there are 2 tiers of lava: lava chasms (`Ql`) and regular lava (`Qlf`). It would be nice if this distinction existed for water tiles, too, so campaign authors can have rivers running along the bottoms of gorges... maybe go with `Qw` for the terrain code?

doofus-01:
I think this is addressed in d77a16f4153ed9441af93cfb8e5f50b9343a6470 and can be closed.  Do you agree, @cooljeanius ?

cooljeanius:
um, hm... well, it wasn't quite what I was looking for originally, but yeah I can probably get that to work, so, ok, I'll close this...

PR #6927, opened 2022-07-31
1.17 Lower terrain

doofus-01:
This is a generalized version of what was requested in #6919.  You can create some awkward situations, but that's up to the map-maker.  It's at the border instead of the hex center (like "cliffs" would need to be) by design.

Known issues:
1.  It's currently an overlay that makes terrain unwalkable, needs some refinement somehow, probably separate unwalkable and base-movement terrains - the unwalkable version with some sort of indicator like the mountain peaks have.
2.  Needs some variations, this is just the default so far (very little new art at this first commit)

Screenshot_20220730_a


doofus-01:
This now has the three-way transition with chasm and regular terrain, but still needs the water versions.
Screenshot_20220731_a

CelticMinstrel:
> still needs the water versions.

I guess that refers to how the two giant holes in the middle of the pond don't have water flowing down the sides?

doofus-01:
Yeah.  Water-low/dry-high is probably most important, but all three versions are pending.  And then there is snow.  This can be a can of worms.

doofus-01:
There are still a few things that need fixing, but I don't think they are enough to break this.  I'll merge if CI doesn't find something.
Screenshot_20220807_a

doofus-01:
That failing macos-intel (Debug) looks unrelated to whatever I did, I'll not let that stop the merge.

PR #6929, opened 2022-07-31
update codeblock test project

newfrenchy83:


Pentarctagon:
Does this need to also include the file that was removed?

CelticMinstrel:
Don't you also need to add the test_simple_wml.cpp file?

newfrenchy83:
> Don't you also need to add the test_simple_wml.cpp file?

this already done in previous PR

CelticMinstrel:
Did the previous PR also remove test_formula_ai.cpp (which was deleted)?

newfrenchy83:
> Did the previous PR also remove test_formula_ai.cpp (which was deleted)?

yes, and test_simple_wml.cpp moved to other projectfiles for test wesnothd and campaignd separately

Pentarctagon:
What is "test wesnothd and campaignd"?

CelticMinstrel:
> else the compilation fail

Why?

newfrenchy83:
@CelticMinstrel we havethat when deprecation included
[compile-fail_log.txt](https://github.com/wesnoth/wesnoth/files/9243220/compile-fail_log.txt)

CelticMinstrel:
Why is it `wesnothd_test` instead of `wesnoth_test`? Why is server_base.cpp included at all? If you look at the file lists [here](https://github.com/wesnoth/wesnoth/tree/master/source_lists), only the files in `boost_unit_tests` and the various `libwesnoth` groups should be included in the test project.

newfrenchy83:
> That sounds like your test project is incorrectly configured.
> 
> Why is it `wesnothd_test` instead of `wesnoth_test`? Why is server_base.cpp included at all? If you look at the file lists [here](https://github.com/wesnoth/wesnoth/tree/master/source_lists), only the files in `boost_unit_tests` and the various `libwesnoth` groups should be included in the test project.

i's is fixed

Pentarctagon:
On a side note, it looks like cmake supports generating CodeBlocks project files, so that might be something to look into rather than having to manually updates like this every time a source file is added or removed.

CelticMinstrel:
I see that there are some essential files removed, but… I assume those are built by the other projects and the test project just references the same object file?

newfrenchy83:
> I see that there are some essential files removed, but… I assume those are built by the other projects and the test project just references the same object file?

yes, the file in test project except test files are already compiled in other project, except if test compilation launch before the other project(out of workspace building), and i don't know how list these file in a librarie project common to wesnoth and test project by exemple.

PR #6930, opened 2022-07-31
Updates for Ubuntu 22.04 CI

Pentarctagon:
The one remaining issue is that the scons+gcc+LTO job fails when executing the boost unit tests with the error `unknown location(0): fatal error: in "Test setup": unknown type`, which I also haven't been able to reproduce locally.


data/lua/wml/find_path.lua
```diff
@@ -59,9 +59,8 @@ function wesnoth.wml_actions.find_path(cfg)
 				ignore_visibility = ignore_visibility
 			} )
 
-			if #path == 0 or cost >= 42424241 then
-				-- it's not a reachable hex. 42424242 is the high value returned for unwalkable or busy terrains
-			else
+			-- it's not a reachable hex. 42424242 is the high value returned for unwalkable or busy terrains
+			if #path ~= 0 and cost < 42424241 then
```
CelticMinstrel, line 63: This renders the comment incorrect. It was correct only because it wasn't in the `if` branch – it's now describing the situation when the condition is false, but in a way that makes it sound like it's describing the situation when the condition is true.

data/lua/wml/modify_unit.lua
```diff
@@ -10,20 +10,20 @@ function wml_actions.modify_unit(cfg)
...
 			end
 		end
 	end
 
-	local function handle_child(cfg, unit_path)
+	local function handle_child(cfg2, unit_path)
 		local children_handled = {}
-		local cfg = wml.shallow_parsed(cfg)
-		handle_attributes(cfg, unit_path)
+		local cfg3 = wml.shallow_parsed(cfg2)
```
CelticMinstrel, line 23: I don't really like adding numbers just to silence a luacheck warning.

I think `cfg2` and `cfg3` could become the same thing by removing the `local` on this line. As for what to call it… I'm not quite sure. I thought of `sub_cfg` but not entirely sure if that fits.

Wedge009: Agreed - I'm also not sure what to call them, but variable names of x1, x2, x3, etc, generally aren't that helpful in all but the most obvious of situations.

CelticMinstrel: In this case it's somewhat logical for them to have the same name, as it's a sort of recursive algorithm.

I wonder, do these local functions use anything else from the closure? It's a much bigger change, but maybe they could be extracted out of the main function as local functions at global scope.

CelticMinstrel: I took a quick look, and it appears that `handle_attributes` and `handle_child` both don't require any closure and could be moved out of the main function. `handle_unit` however does require closure.

data/lua/wml/find_path.lua
Pentarctagon: Updated.

data/lua/wml/modify_unit.lua
Pentarctagon: Updated.

data/lua/wml/find_path.lua
CelticMinstrel: What did you change…?

data/lua/wml/modify_unit.lua
```diff
@@ -10,20 +10,20 @@ function wml_actions.modify_unit(cfg)
...
+	local function handle_attributes(handle_attr_cfg, unit_path, toplevel)
+		for current_key, current_value in pairs(wml.shallow_parsed(handle_attr_cfg)) do
 			if type(current_value) ~= "table" and (not toplevel or (current_key ~= "type" and current_key ~= "mode")) then
 				wml.variables[string.format("%s.%s", unit_path, current_key)] = current_value
 			end
 		end
 	end
 
-	local function handle_child(cfg2, unit_path)
+	local function handle_child(handle_child_cfg, unit_path)
```
CelticMinstrel, line 21: This is a slight improvement, but I don't think including the function name in the argument name is a good idea either.

Pentarctagon: I really don't have any preference on what the name is, so let me know what you'd prefer I change it to.

CelticMinstrel: `child_cfg` seems reasonable, for both functions.

Pentarctagon: Alright, done.

src/formula/string_utils.hpp
```diff
@@ -49,7 +49,7 @@ inline bool might_contain_variables(const std::string &str)
  * 'str' with the equivalent symbols in the given symbol table. If 'symbols'
  * is nullptr, then game event variables will be used instead.
  */
-std::string interpolate_variables_into_string(const std::string &str, const string_map * const symbols);
+std::string interpolate_variables_into_string(const std::string &str, const std::map<std::string, t_string> * const symbols);
```
CelticMinstrel, line 52: Why was this changed? Is `string_map` not visible in this file?

src/utils/general.hpp
```diff
@@ -17,6 +17,11 @@
 #include <algorithm>
 #include <cctype>
 #include <functional>
+#include <string>
+
+#if defined(__clang__) || defined(__GNUG__)
+#include <cxxabi.h>
```
CelticMinstrel, line 23: Shouldn't this kind of thing be encapsulated in the corresponding cpp file instead?

src/formula/string_utils.hpp
Pentarctagon: Doxygen has its own source code parsing, which doesn't always exactly match what compilers will allow. So there will be cases like this where Doxygen gets confused by something that's valid C++.

src/utils/general.hpp
Pentarctagon: There isn't a `utils/general.cpp`, and I wasn't sure where else would be a better place for this kind of thing.

CelticMinstrel: If you're explicitly trying to avoid adding a new source file, I guess you could just drop it in wesnoth.cpp… but probably adding a new source file (whether general.cpp or something else) is the best option here.

src/formula/string_utils.hpp
CelticMinstrel: It doesn't seem right to compromise our code just to satisfy Doxygen. If we _have_ `string_map` we should use it everywhere it's needed, not expand it in arbitrary places. If that confuses Doxygen, we need to think of a way to make it _not_ confuse Doxygen.

src/utils/general.hpp
Pentarctagon: I assume I can create a `utils/general.cpp` without any issues.

src/formula/string_utils.hpp
Pentarctagon: I'm open to suggestions. There's no way to forcibly define certain things for Doxygen though, as far as I know.

CelticMinstrel: Okay… it's defined in the other string_utils.hpp which is literally included at the top of this file. Is Doxygen getting confused because two files have the same name…?

src/utils/general.hpp
Pentarctagon: Done.

CelticMinstrel:
That's kind of a bizarre way to determine that all dialogs have been tested…

Pentarctagon:
Kinda, yeah.

Pentarctagon:
Given all the tests are separate executions now, that seemed like the easiest way to share data between them.

CelticMinstrel:
It makes sense in the context of running the tests, but it also means that the "test_last" test will always fail if run on its own.

The other option would be to make not testing a dialog a compile-time failure…

As a side note, I'm also not a fan of using numbers in the test names. It's not _that_ hard to name them after the dialog they're testing, is it?

Pentarctagon:
> It makes sense in the context of running the tests, but it also means that the "test_last" test will always fail if run on its own.
> 
> The other option would be to make not testing a dialog a compile-time failure…

How would that be done?

> As a side note, I'm also not a fan of using numbers in the test names. It's not that hard to name them after the dialog they're testing, is it?

It's not hard, it was just faster to do it with numbers.

run_boost_tests
```diff
@@ -0,0 +1,67 @@
...
+            retry = True
+        if not retry:
+            return res
+        sdl_retries += 1
+        print("Could not initialise SDL_video error, attempt", sdl_retries)
+
+if __name__ == '__main__':
+    ap = argparse.ArgumentParser()
+    ap.add_argument("-e", "--executable", help="The name of the boost unit test executable. Default is boost_unit_tests.", default="boost_unit_tests")
+    ap.add_argument("-p", "--path", metavar="dir", help="Path to wesnoth binary. By default assume it is with this script.", default=".")
```
stevecotton, line 38: It's the boost_unit_tests binary, not the wesnoth one.

src/formula/string_utils.hpp
stevecotton: Might be https://github.com/doxygen/doxygen/issues/1398 , I've seen other problems with Doxygen getting confused about overloaded functions. Some have been fixed, but there's definitely some still-open longstanding issues.

I don't have a better suggestion.

run_boost_tests
Pentarctagon: Fixed.

CelticMinstrel:
> How would that be done?

I'm not 100% sure, but I think there may be some way to do it with templates somehow. It might require alterations/additions to the `REGISTER_DIALOG` macro too. Basically, my half-formed thought is to have a compile-time list of dialogs somehow, and iterate over them with template metaprogramming in a way that causes an untested dialog to throw an "undefined template" error.

…actually, that might be a link error rather than a compiler error?

It's possible that my idea won't even work, but I think it may be worth exploring at least.

PR #6935, opened 2022-08-02
Make modeless_dialog an actual window

mesilliac:
Not only was it conceptually confusing that a modeless dialog was not a window, but it causes several inefficiencies and general incapabilities when converting to use the new draw system.

So here i made a modeless dialog a window. This was the easy one as there are basically no modeless dialogs right now. Basically just tooltips and the debug clock.

As a note this will slightly invalidate #5561 which also uses a modeless dialog. But any changes are minimal and can be tacked on to whichever gets merged second.

The window builder finalization stuff was moved to a method of `window` so it can be called in the modeless dialog constructor.

As far as practical usage changes go... most stuff that was previously in "post_build" and "pre_show" virtual callbacks can simply go into the subclass constructor.

Modeless dialogs now require the window id to be given as a constructor parameter. This is becaues the virtual function that is defined by `REGISTER_DIALOG` is not able to be called from the `modeless_dialog` cosntructor. It is simply called from the subclass constructor and passed on.

### DIFF
--- a/src/gui/core/window_builder.cpp
+++ b/src/gui/core/window_builder.cpp
@@ -41,31 +41,7 @@ std::unique_ptr<window> build(const builder_window::window_resolution& definitio
 	// best size (if needed) after all widgets have been placed.
 	auto win = std::make_unique<window>(definition);
 	assert(win);
-
-	for(const auto& lg : definition.linked_groups) {
-		if(win->has_linked_size_group(lg.id)) {
-			t_string msg = VGETTEXT("Linked '$id' group has multiple definitions.", {{"id", lg.id}});
-
-			FAIL(msg);
-		}
-
-		win->init_linked_size_group(lg.id, lg.fixed_width, lg.fixed_height);
-	}
-
-	win->set_click_dismiss(definition.click_dismiss);
-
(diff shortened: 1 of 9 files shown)


CelticMinstrel:
> As far as practical usage changes go... most stuff that was previously in "post_build" and "pre_show" virtual callbacks can simply go into the subclass constructor.

Note that, for a hypothetical future where we have a Lua API to show a modeless dialog, we will need a `pre_show` callback, so that still needs to be a thing… admittedly, it doesn't _need_ to be called `pre_show` in C++, though.

I assume it should be fine as long as you can make the modal version of this work in the Lua API.

mesilliac:
The modal version does still have all the callbacks so should continue to work unchanged, i only removed it here because nothing was using it after refactor. It might make more sense if pre_show and post_show go in `window` rather than the subclasses, but either way modal dialog will continue to have them.

CelticMinstrel:
Looking more closely, it seems the Lua API doesn't use either a modal_dialog or a modeless_dialog. It just constructs a window, calls pre_show, shows the window (modally), and then calls post_show. (pre_show and post_show in this case are Lua functions passed to `gui.show_dialog`.)

PR #4083, opened 2019-05-16
Add pre attack event #4075

spixi:
This implements #4075

Issue #4075, opened 2019-05-13
[Feature Request] New event: pre attack


spixi:
The `attack` event happens after recalculating the number of attacks and the movement of the unit. Maybe a new `pre attack` event would be useful for modifications, which affect these values. This looks to be easily implementable.

Wedge009:
Not sure what is supposed to be tested, so removing the label.

PR #6946, opened 2022-08-07
Add pre attack event

Pentarctagon:
This addresses #4075

---

This is a re-PR of #4083 - I'll be testing and adding some unit tests for it, and if everything works then I'll merge it.

### DIFF
--- a/data/test/scenarios/events-test_pre_attack.cfg
+++ b/data/test/scenarios/events-test_pre_attack.cfg
@@ -1,0 +1,148 @@
+#####
+# API(s) being tested: [event]name=pre_attack
+##
+# Actions:
+# Move a unit and attack with it.
+##
+# Expected end state:
+# The pre_attack event is triggered.
+#####
+{GENERIC_UNIT_TEST "events-test_pre_attack_base" (
+    [event]
+        name=start
+        [do_command]
+            [move]
+                x=7,8,9,10,11,12
+                y=3,3,3,3,3,3
+            [/move]
+        [/do_command]
+        [do_command]
+            [attack]
+                weapon=0
+                defender_weapon=0
+                [source]
+                    x,y=12,3
+                [/source]
+                [destination]
+                    x,y=13,3
+                [/destination]
+            [/attack]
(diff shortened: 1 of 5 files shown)


src/actions/attack.cpp
```diff
@@ -1367,6 +1367,25 @@ void attack::perform()
...
+	}
+
+	try {
+		fire_event("pre_attack");
+	} catch(const attack_end_exception&) {
+		return;
+	}
+
 	a_.get_unit().set_facing(a_.loc_.get_relative_dir(d_.loc_));
 	d_.get_unit().set_facing(d_.loc_.get_relative_dir(a_.loc_));
```
stevecotton, line 1390: I'm not sure of the purpose of the `pre_attack` event, does it make sense for the units to face each other first, and to face each other even if an `attack_end_exception` happens?

Setting it first would allow the event to be sure that the units' facing already matched the attack direction. I guess setting it afterwards might allow a "charge bonus" if the unit's movement already made it face the correct direction.

Pentarctagon: The purpose of it is described in the FR issue.

I don't understand the significance of the direction the units are facing? My immediate thought would be it would make sense for them to be facing each other before the event is fired, since the decision to attack has already been made, even if it ends up not actually happening for some other reason.

data/test/scenarios/events-test_pre_attack.cfg
```diff
@@ -0,0 +1,194 @@
...
+                [variable]
+                    name=x2
+                    equals=13
+                [/variable]
+            [/not]
+            [then]
+                {FAIL}
+            [/then]
+        [/if]
+        [if]
```
CelticMinstrel, line 155: Each of these if statements should be reduced to an `{ASSERT}`.

Pentarctagon: Updated.

src/actions/attack.cpp
Pentarctagon: Updated.

PR #6948, opened 2022-08-07
A script to add new source files

mesilliac:
Usage: `./add_source_file <filename> [target ...]`

Targets are "wesnoth", "wesnothd", "campaignd", "lua", "tests".
It defaults to the "wesnoth" target.

The 'pbxproj' python package is required to modify the Xcode project. As may be inferred, this is a Python script.

It probably works? I don't know, really. It adds some stuff and looks vaguely correct.

It also sorts the `source_lists` files when adding to them, if that's a problem it can be made to not do that.

### DIFF
--- a/add_source_file
+++ b/add_source_file
@@ -1,5 +1,6 @@
-#!/usr/bin/python
-# For adding source files to the Battle for Wesnoth project.
+#!/usr/bin/env python3
+# encoding: utf-8
+# A script for adding source files to the Battle for Wesnoth project.
 
 import sys
 import inspect
--- a/add_source_file
+++ b/add_source_file
@@ -1,0 +1,238 @@
+#!/usr/bin/python
+# For adding source files to the Battle for Wesnoth project.
+
+import sys
+import inspect
+import uuid
+import pathlib
+
+try:
+    import pbxproj
+except:
+    print('\n'.join((
+        'This script requires the "pbxproj" module.',
+        'Install it using "pip install pbxproj"',
+    )))
+    exit(1)
+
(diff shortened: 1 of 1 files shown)


add_source_file
```diff
@@ -0,0 +1,238 @@
...
+    # or maybe targets could be checked somehow,
+    # or maybe the file could simply be completely removed and readded.
+    if project.get_files_by_name(filename.name, parent=parent_group):
+        print("  '{filename}' already found in Xcode project, skipping")
+        return
+    
+    # force is True here because otherwise a duplicate filename in
+    # a different place will block addition of the new file.
+    # the rest is just to match existing project file structure.
+    project.add_file(filename,
```
CelticMinstrel, line 127: Does this keep the files in the group sorted? I always add new files at the alphabetical location.

mesilliac: It seems to append to the end of the group. I'm not sure if there's a way to change this behaviour.

add_source_file
```diff
@@ -0,0 +1,238 @@
+#!/usr/bin/python
```
stevecotton, line 1: On non-Windows, that's python2, not python3. Copy & paste the stuff from another of our scripts:
```
#!/usr/bin/env python3
# encoding: utf-8
```

mesilliac: Hmm i thought Debian was the last remaining holdout there. Updated

Pentarctagon:
@Elvish-Hunter are you around?

Elvish-Hunter (approved):

Issue #5670, opened 2021-04-04
wesnoth crashes randomly when fps is enabled


grz0:
### Game and System Information
 - What version of the game are you running?
reproduced on the latest flatpak nightly build, and on game compiled from 3da068801be3dd8a1f762a6a8880e4b304afc430

### Describe the bug
If fps is enabled then the game may crash while scrolling after reloading save

### To Reproduce

1. start a new game on a large enough map to be able to scroll (for example Waterloo Sunset)
2. enable fps in console
3. save game
4. move the cursor to the left-upper corner
5. reload save using hotkeys, the display should be scrolled after reloading the save file
6. repeat 5, the game should crash after 3-4 attempts

### Expected behavior
Game should not crash 

### Additional context
```
Program terminated with signal SIGSEGV, Segmentation fault.

(gdb) bt
#0  0x00005571a51210b7 in display::update_display() ()
#1  0x00005571a5123d38 in display::draw_wrap(bool, bool) ()
#2  0x00005571a51324bf in display::draw(bool, bool) ()
#3  0x00005571a51321e8 in display::redraw_everything() ()
#4  0x00005571a5132540 in display::draw(bool, bool) ()
#5  0x00005571a51337cf in display::scroll_to_tile(map_location const&, display::SCROLL_TYPE, bool, bool) ()
#6  0x00005571a4988b4d in playsingle_controller::init_gui() ()
#7  0x00005571a497a475 in play_controller::start_game() ()
#8  0x00005571a4984fad in playsingle_controller::play_scenario_init() ()
#9  0x00005571a49899a5 in playsingle_controller::play_scenario(config const&) ()
#10 0x00005571a476bb5f in campaign_controller::playsingle_scenario(end_level_data&) ()
#11 0x00005571a476c2fe in campaign_controller::play_game() ()
#12 0x00005571a4773287 in game_launcher::launch_game(game_launcher::reload_mode) ()
#13 0x00005571a45668ab in main ()
```

Wedge009:
Does this still happen? I imagine things have probably improved following #6855.

Descacharrado:
It still happens in current master

grz0:
I can still reproduce on 1d63543f98480e5239f18232a797d6f9010643aa


```
(gdb) bt
#0  0x000055ccbdbd4d03 in display::update_fps_label() (this=0x55ccc31f7600) at src/display.cpp:1357
[...]

PR #6956, opened 2022-08-09
Fix crash in update_fps_label()

grz0:
update_fps_label() uses a static variable to track current frame sample.
This causes division by zero while loading a save file if the
variable was previously equal to 9.

The commit changes the variable to private member to match its lifetime
with a lifetime of frametimes_ circular buffer.

Fixes: #5670

### DIFF
--- a/src/display.hpp
+++ b/src/display.hpp
@@ -767,6 +767,7 @@ protected:
 	mutable events::generic_event scroll_event_;
 
 	boost::circular_buffer<unsigned> frametimes_; // in milliseconds
+	int current_frame_sample = 0;
 	unsigned int fps_counter_;
 	std::chrono::seconds fps_start_;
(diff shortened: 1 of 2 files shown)


Pentarctagon:
Thanks for the fix!

src/display.hpp
```diff
@@ -767,6 +767,7 @@ class display : public gui2::top_level_drawable
 	mutable events::generic_event scroll_event_;
 
 	boost::circular_buffer<unsigned> frametimes_; // in milliseconds
+	int current_frame_sample = 0;
```
soliton-, line 770: The variable name should end in underscore like the other member variables.

grz0: https://github.com/wesnoth/wesnoth/pull/6965

PR #6961, opened 2022-08-12
Font/Text: tweak line height

Vultraz:
For CI.


CelticMinstrel:
What is `PANGO_SCALE` and why is it no longer involved here? That seems possibly wrong.

PR #6963, opened 2022-08-13
[UtBS] S4 extra dialog to acknowledge effect of finding the map in S3

### LINKED ISSUE #6199: Issue with UtBS: A Stirring in the Night and the next scenario
* Is the text in English? If not, what language is it in?
Yes.

* What campaign does the text appear in? If it's not in a campaign, say where it is:
UtBS: A Stirring in the Night and the next Descending into Darkness.

* Please quote the text verbatim. You can use `**bold**` to point out the mistaken part:
Per my post and the feedback I got here: https://r.wesnoth.org/t54989

* _(optional)_ What do you propose to change the text to?
I'm going to start with what max_torch said in my forum post: "Perhaps, we could add dialog to the next scenario saying 'oh hey all the terrain here matches the map we found from that orc leader(in the previous scenario)' to make it clearer what the actual benefit was.". 
Additionally, I felt the same thing when I played it for what they said about trying to sneak part. Perhaps after she says 
"Let's see if we can sneak past them", the Goblin rider saying "Intruders! Alert the others and attack" and she saying "I spoke too soon blah blah ". Something like this would remove any idea the player had about sneaking.

max-torch:
Aims to resolve #6199.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
@@ -513,7 +513,7 @@
                 [then]
                     [message]
                         speaker=Nym
-                        message= _ "Merely a few steps in and all these goblins already know we’re here! I guess it’s not that easy to sneak around in someone else's home territory."
+                        message= _ "Merely a few steps in and all these goblins already know we’re here! I guess it’s not that easy to sneak around in someone else’s home territory."
                     [/message]
                 [/then]
             [/if]
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
@@ -349,19 +349,6 @@
     [event]
         name=start
 
-        [if]
-            [variable]
-                name=sneak_up
-                boolean_equals=yes
-            [/variable]
-            [then]
-                [message]
-                    speaker=Nym
-                    message= _ "Thanks to that map we got from the orcs last night, our scouts have been able to confirm the layout of the surrounding terrain, including the location of an orcish camp up ahead! I wonder if we could sneak around without alerting them. Oh my, what’s that far off in the distance? Are those..."
-                [/message]
-            [/then]
-        [/if]
-
         [message]
             speaker=Nym
             message= _ "Those mountains are huge! I never thought they would be so big. And what’s that white stuff on the tops of the peaks?"
@@ -371,6 +358,19 @@
             speaker=Zhul
             message= _ "I wish Garak were here, he’d know more about these lands than I do. I’ve never been up here, but I heard stories from the few who have made the journey and returned. That white stuff is called snow, Nym, and the mountains are very cold. These smaller hills aren’t as hard to cross, but they are filled with orcs and goblins. It may look peaceful now, but they hide in the many caves and tunnels beneath the surface."
         [/message]
+
+        [if]
+            [variable]
+                name=sneak_up
+                boolean_equals=yes
+            [/variable]
+            [then]
+                [message]
+                    speaker=Nym
+                    message= _ "Speaking of these lands, we may not have Garak to guide us, but all the terrain in this area perfectly matches the map we got from the orcs last night! It shows an orcish camp just up ahead! I wonder if we could sneak around without alerting them."
+                [/message]
+            [/then]
+        [/if]
 
         [message]
             speaker=Kaleh
@@ -504,6 +504,19 @@
                 speaker=Zhul
                 message= _ "Goblins are so predictable."
             [/message]
+
+            [if]
+                [variable]
+                    name=sneak_up
+                    boolean_equals=yes
+                [/variable]
+                [then]
+                    [message]
+                        speaker=Nym
+                        message= _ "Merely a few steps in and all these goblins already know we’re here! I guess it’s not that easy to sneak around in someone else's home territory."
+                    [/message]
+                [/then]
+            [/if]
         [/event]
     [/event]
 
@@ -616,21 +629,6 @@
             speaker=Kaleh
             message= _ "How can he move that fast? He is faster than any goblin rider I have ever seen. It’s almost unnatural."
         [/message]
-
-        [if]
-            [variable]
-                name=sneak_up
-                boolean_equals=yes
-            [/variable]
-            [then]
-                [message]
-                    speaker=Nym
-                    message= _ "Spotted immediately! Well, I guess I spoke too soon about trying sneaking."
-                [/message]
-            [/then]
-        [/if]
-
-        {CLEAR_VARIABLE sneak_up}
     [/event]
 
     # When Panok dies player gets ring and naga ambush is triggered if it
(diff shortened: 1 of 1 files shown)


data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
```diff
@@ -603,6 +616,21 @@
...
+
+        [if]
+            [variable]
+                name=sneak_up
+                boolean_equals=yes
+            [/variable]
+            [then]
+                [message]
+                    speaker=Nym
+                    message= _ "Spotted immediately! Well, I guess I spoke too soon about trying sneaking."
```
CelticMinstrel, line 628: "trying sneaking" isn't technically wrong I think, but it sounds weird to me. I'd go with "trying to sneak" instead.

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
```diff
@@ -342,13 +342,26 @@
...
 
+        [if]
+            [variable]
+                name=sneak_up
+                boolean_equals=yes
+            [/variable]
+            [then]
+                [message]
+                    speaker=Nym
+                    message= _ "Thanks to that map we got from the orcs last night, our scouts have been able to confirm the layout of the surrounding terrain, including the location of an orcish camp up ahead! I wonder if we could sneak around without alerting them. Oh my, what’s that far off in the distance? Are those..."
```
stevecotton, line 360: The bit from "Oh my" onwards doesn't feel like it flows into the next `[message]` to me. I'd expect it to be something about enemy units that she's seen, but in the next line she switches to talking about mountains. If big mountains are within scouting range then she should have already seen them before the scouts were sent.

stevecotton, line 360: The logic of "map -> scouts -> better knowledge of terrain" doesn't click for me. The map could have shown the scouts some vantage points, but for the scouts to have seen anything without going deeper into the map than the starting point of the next scenario we have accept that the scouts have a much larger vision range in the cutscene than they do on the actual map.

The logic of "map -> better knowledge of terrain" seems simpler, and doesn't promote the players to overthink the scouts' vision range.

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
stevecotton, line 628: Seems like this should be in `# Event 1: goblins appear in nearby foothills` rather than `# Event 3`. I'd put it where Zhul's "Goblins are so predictable" is, and hide Zhul's line if this triggers.

If it goes in event 3, it should go before Kaleh's line. Kaleh changes the subject to be about the rider's abilities.

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
```diff
@@ -603,6 +616,21 @@
...
+            [/variable]
+            [then]
+                [message]
+                    speaker=Nym
+                    message= _ "Spotted immediately! Well, I guess I spoke too soon about trying sneaking."
+                [/message]
+            [/then]
+        [/if]
+
+        {CLEAR_VARIABLE sneak_up}
```
stevecotton, line 633: (Also posted on Discord) Maybe we should make a WML coding style rule for variables that depend on the events of a previous scenario, so we only clear variables like `sneak_up` in the `victory` event? That would make it easy to add more conditional dialogue, even in events that might not be triggered in all playthroughs.

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
max-torch: I put it in event 1 because from playing the scenario, **event 3 always occurs first no matter what**, and the line given by the enemy unit **actually exclaims that he has spotted elves**.

The timing is more natural this way.

About making the line be before Kaleh's, sure i can do that.

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
max-torch: I think I understand what you're saying. This entails putting `{CLEAR_VARIABLE sneak_up} in the `victory` event, correct?

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
stevecotton: That seems like a bug - without the map, event 1 triggers first (probably on turn 2). With the map, event 3 triggers when Panok captures a village on turn 1.

Panok's comment makes little sense on turn 1, because Panok says the elves are stronger than they thought. At this point, he's only seen 1 castle full of elvish recruits/recalls.

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
stevecotton: Yes

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
max-torch: Well, it's okay I can find a way to make the dialog make sense for occurring at Event 1.

max-torch:
Please check if latest commit resolves all feedback points raised.

knyghtmare (approved):
Looks good to me.
Anyways can you adjust the commit name "Update 04_Descending_into_Darkness.cfg"? to maybe "UtBS S04: adjust variables to handle extra dialog better" or something similar?

stevecotton:
Looks good to me.

CelticMinstrel:
> Anyways can you adjust the commit name "Update 04_Descending_into_Darkness.cfg"? to maybe "UtBS S04: adjust variables to handle extra dialog better" or something similar?

Or we can just squash-merge and it goes away.

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
```diff
@@ -491,6 +504,19 @@
...
+
+            [if]
+                [variable]
+                    name=sneak_up
+                    boolean_equals=yes
+                [/variable]
+                [then]
+                    [message]
+                        speaker=Nym
+                        message= _ "Merely a few steps in and all these goblins already know we’re here! I guess it’s not that easy to sneak around in someone else's home territory."
```
Descacharrado, line 516: Typographic quote on "else's"

Descacharrado (changes requested):

Descacharrado (approved):

stevecotton:
The build failure is just because the `macos-intel-release` build was cancelled. Please would whoever cancelled it post a reason?

I'll re-run the job tomorrow if it's unexplained, because having the build-failure icon in the PR list makes it less likely that people will review it.

nemaara (approved):
At some point we should fix the dialogue lines to make the characters' personalities come out more strongly but that's probably easier to do in a batch later for some kind of rewrite.

stevecotton (approved):
Reminder: use a squash-merge instead of a rebase.

Issue #4494, opened 2019-10-20
lint warnings


jostephd:
### Game and System Information
 - What version of the game are you running?
1.15.1+dev

### Describe the bug
`make -C data/tools lint` has many warnings

### Expected behavior
`make -C data/tools lint` has no warnings

### Additional context
This step is part of the release checklist, https://wiki.wesnoth.org/ReleasingWesnoth#General_maintenance, so it shouldn't output any warnings. Or alternatively, the release process should documented which warnings can be ignored.

stevecotton:
Some of the warnings are already in other issues:
* #4382 for ANL's sides
* #4379 for defense caps and "missing forest=-70"
* #4538 for "unknown 'memoirs_elves_use_orcish_weapons'"
* #4380 for the [lua] tags in LoW S09

Stuff that's not in other issues:
- [ ] lots of warnings and missing icons in DiD
- [ ] wmllint will complain about ids used in events or objectives before the corresponding [side] tags
- [ ] wmllint complains `[side] without type attribute` if the type is inside a macro (affects LoW and SotA)
- [ ] warnings about "fog in map file" for some multiplayer maps

False positives / definite wmllint bugs:
```
"core/help.cfg", several "should not have translation mark" false positives
"core/units/humans/Loyalist_Royal_Guard.cfg", line 2: unit Royal Guard has superfluous {NOTE_*}
"core/units/humans/Loyalist_Swordsman.cfg", line 2: unit Swordsman has superfluous {NOTE_*}
"Delfadors_Memoirs/scenarios/20_Prince_of_Wesnoth.cfg", line 159: side number 1 is out of sequence (4 expected)
"Delfadors_Memoirs/scenarios/20_Prince_of_Wesnoth.cfg", line 239: side number 1 is out of sequence (4 expected)
"Northern_Rebirth/scenarios/08a_Elvish_Princess.cfg", line 217: unknown 'Stalrag' referred to by id
"Northern_Rebirth/scenarios/08a_Elvish_Princess.cfg", line 243: unknown 'Ro'Arthian' referred to by id
"Northern_Rebirth/scenarios/08a_Elvish_Princess.cfg", line 268: unknown 'Ro'Sothian' referred to by id
"Northern_Rebirth/scenarios/11a_The_Eastern_Flank.cfg", line 158: unknown 'Krash' referred to by id
[...]

PR #6966, opened 2022-08-15
Fix wmllint translation mark errors

Dou1622:
Fixes all wmllint warnings about translation mark.
See #4494


data/campaigns/Secrets_of_the_Ancients/gui/zombie_recruit_dialog.cfg
```diff
@@ -109,16 +113,24 @@
                         [column]
                             border = "right"
                             border_size = 5
+                            
+                            # wmllint: markcheck off
                             [image]
                                 label = "../../images/icons/alignments/alignment_chaotic_30.png"
                             [/image]
+                            # wmllint: markcehck on
```
CelticMinstrel, line 121: typo

data/campaigns/Under_the_Burning_Suns/units/units.cfg
```diff
@@ -16,7 +16,7 @@
     male_name= _ "Unknown"
     female_name= _ "Unknown"
     plural_name= _ "Unknown"
-    description="<img>src=units/unknown-unit.png~RC(magenta>red) align=left float=yes</img>" + _ "
+    description="<img>src=units/unknown-unit.png~RC(magenta>red) "+ _ "align=left float=yes"+"</img>"
```
CelticMinstrel, line 19: This is utterly crazy and horribly wrong. The existing code is correct, so you should probably just revert to that.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -178,7 +178,7 @@
                                                                                 [label]
                                                                                     definition = "default_large"
                                                                                     id = "training_caption"
-                                                                                    label = "default description"
+                                                                                    label = _ "default description"
```
CelticMinstrel, line 181: I would replace all of these with empty strings instead, or maybe even remove these lines entirely. There's no point throwing a "default description" string into the po files that will never be seen in the game.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -257,7 +257,7 @@
                                             border_size = 5
                                             [scroll_label]
                                                 id = "label_content"
-                                                label = "Text"
+                                                label = _ "Text"
```
CelticMinstrel, line 260: This too should not be translatable. It should be empty or absent.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -391,11 +391,13 @@
                                                 [/row]
                                                 [row]
                                                     [column]
+                                                    	  # wmllint: markcheck off
```
CelticMinstrel, line 394: Why is this markcheck off? Shouldn't it be `"<b>" + _"Recruits" + ":</b>"` or some such?

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -430,7 +432,7 @@
                                                                                     [/column]
                                                                                     [column]
                                                                                         [label]
-                                                                                            label="↔"
+                                                                                            label= _ "↔"
```
CelticMinstrel, line 435: Does anyone know what this is used for? I think if it's going to be translatable then it needs a po comment and/or a context prefix.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -456,11 +458,13 @@
                                                 [/row]
                                                 [row]
                                                     [column]
+                                                    	  # wmllint: markcheck off
```
CelticMinstrel, line 461: Again, making this translatable seems like the correct choice, rather than ignoring it.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -475,11 +479,13 @@
...
```
CelticMinstrel, line 482: Again, making this translatable seems correct, instead of ignoring it.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -494,11 +500,13 @@
                                                 [/row]
                                                 [row]
                                                     [column]
+                                                        # wmllint: markcheck off
```
CelticMinstrel, line 503: Again, translatable, not ignored.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -513,12 +521,14 @@
...
```
CelticMinstrel, line 524: Again, translatable, not ignored.

data/campaigns/World_Conquest/gui/invest_dialog.cfg
```diff
@@ -201,7 +201,7 @@
                                                         vertical_grow = true
                                                         [scroll_label]
                                                             id = "info_label"
-                                                            label = "Text"
+                                                            label = _ "Text"
```
CelticMinstrel, line 204: This should be an empty string or removed altogether.

data/campaigns/World_Conquest/gui/invest_dialog.cfg
```diff
@@ -324,7 +324,7 @@
                         [column]
                             horizontal_alignment = "right"
                             [button]
-                                label = "Ok"
+                                label = _ "Ok"
```
CelticMinstrel, line 327: This should probably be a core string from a different textdomain. I'm not quite sure where that would be found though.

data/core/terrain-graphics/deprecated-internal-corners.cfg
```diff
@@ -47,6 +47,6 @@
     [/terrain_graphics]
 #enddef
 
-# #define DISABLE_CORNER_TRANSITIONS TERRAIN
+# # define DISABLE_CORNER_TRANSITIONS TERRAIN
```
CelticMinstrel, line 50: Revert this file.

CelticMinstrel:
In addition to all my specific comments, some of these strings may require context prefixes. I'm not sure which ones would need that, though.

data/campaigns/tutorial/gui/character_selection.cfg
```diff
@@ -39,17 +39,21 @@
                             grow_factor = 1
                             border = "all"
                             border_size = 5
+                            # wmllint: markcheck off
```
stevecotton, line 42: Seems a bug in wmllint, if it's not recognising these as wanting image filenames.

Dou1622: Does someone need to create an issue for that?

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -430,7 +429,9 @@
                                                                                     [/column]
                                                                                     [column]
                                                                                         [label]
-                                                                                            label="↔"
+                                                                                            # wmllint: markcheck off
+                                                                                            label= "↔"
```
CelticMinstrel, line 433: I'm just not sure what the correct answer is for this one. Making it translatable doesn't seem like a terrible idea, but I don't know if it's a good idea.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -583,7 +584,7 @@
                                                                     border_size = 5
                                                                     [label]
                                                                         definition = "gold_small"
-                                                                        label = "Use WC2 Unitmarkers"
+                                                                        label = _ "Use WC2 Unitmarkers"
```
CelticMinstrel, line 587: Since it's going translatable, let's also remove the 2 while we're at it.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -603,7 +604,7 @@
                                                                     border_size = 5
                                                                     [label]
                                                                         definition = "gold_small"
-                                                                        label = "Use WC2 experimental pickup"
+                                                                        label = _ "Use WC2 experimental pickup"
```
CelticMinstrel, line 607: Since it's going translatable, let's also remove the 2 while we're at it.

data/campaigns/World_Conquest/gui/help_dialog.cfg
```diff
@@ -664,7 +665,7 @@
                                                                     border_size = 5
                                                                     [label]
                                                                         definition = "gold_small"
-                                                                        label = "WC2 version"
+                                                                        label = _ "WC2 version"
```
CelticMinstrel, line 668: Since it's going translatable, let's also remove the 2 while we're at it.

Maybe we can also spell it out in full – "World Conquest".

There may also be an argument that an independent version isn't needed, but that's probably beyond the scope of this PR.

data/core/terrain.cfg
```diff
@@ -1377,7 +1377,7 @@ Occasionally caves are <italic>text='illuminated'</italic>."
     id=cavewall_earthy_hewn
     name= _ "Cave Wall"
 #ifndef SCHEMA_VALIDATION
-    editor_name="Reinforced Earthy Cave Wall"
+    editor_name= _ "Reinforced Earthy Cave Wall"
```
CelticMinstrel, line 1380: This one was intentionally left non-translatable, as shown by the `#ifndef` above it.

CelticMinstrel:
You also need to run wmlindent on this.

data/campaigns/Under_the_Burning_Suns/units/units.cfg
Dou1622: Revert.

data/campaigns/World_Conquest/gui/help_dialog.cfg
Dou1622: Replaced it with empty strings instead.

data/campaigns/World_Conquest/gui/help_dialog.cfg
Dou1622: Replaced it with empty strings.

data/campaigns/World_Conquest/gui/help_dialog.cfg
Dou1622: Turn off markcheck instead.

data/campaigns/World_Conquest/gui/invest_dialog.cfg
Dou1622: Replaced it with empty strings.

data/campaigns/World_Conquest/gui/invest_dialog.cfg
Dou1622: Revert.

data/core/terrain-graphics/deprecated-internal-corners.cfg
Dou1622: Revert.

data/campaigns/Descent_Into_Darkness/scenarios/07a_A_Small_Favor.cfg
```diff
@@ -752,7 +752,9 @@
         [floating_text]
             x,y=$x1,$y1
             color=255,0,0
+            # wmllint: markcheck off
             text=-1
```
Elvish-Hunter, line 756: wmllint considers `text=` a translatable key. However, here we've got a numerical value, so is there any language that needs to have numbers translated? If so, the correct fix would be marking this string as translatable, otherwise I'd probably need to make wmllint skip the translation mark check for strings made only of numerical characters.

CelticMinstrel: > is there any language that needs to have numbers translated?

Yes, I think so. Any language using a non-Latin script will probably need to translate numbers. That said, putting them in the po file is probably the wrong answer even then – the correct answer would be to use localization routines to format the number (which will also insert commas or the equivalent as needed).

I was actually thinking of adding an `amount=` key to `[floating_text]` to enable that treatment.

data/campaigns/World_Conquest/gui/help_dialog.cfg
Dou1622: I search this symbol ("↔")in Wesnoth. It is only translated in zh_CN  po file(wesnoth/po/wesnoth-sota/zh_CH.po.) Since it has been translated should we use translation mark here? or should it be markcheck off?

Dou1622: > I'm just not sure what the correct answer is for this one. Making it translatable doesn't seem like a terrible idea, but I don't know if it's a good idea.

I search this symbol ("left_right_arrow")in Wesnoth. It is only translated in zh_CN po file(wesnoth/po/wesnoth-sota/zh_CH.po.) Since it has been translated should we use translation mark here? or should it be markcheck off? Or  should I remove it from zh_ch po file.

data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
```diff
@@ -367,13 +367,15 @@
             animate=no
             kill=no
         [/harm_unit]
-
+
```
stevecotton, line 370: These whitespace changes should make the CI check with `wmlindent` fail.

data/campaigns/World_Conquest/gui/settings_dialog.cfg
```diff
@@ -173,7 +173,7 @@
                         [column]
                             grow_factor = 1
                             [label]
-                                label = "hill_size"
+                                label = _ "hill_size"
```
stevecotton, line 176: The labels in this dialog might need renaming, rather than just marking as translatable. With its underscore, `hill_size` is the one that I noticed, but many of them seem to need answers to the question "what does this mean?"

data/campaigns/World_Conquest/gui/settings_dialog.cfg
```diff
@@ -109,7 +109,7 @@
                         [column]
                             grow_factor = 1
                             [label]
-                                label = "length"
+                                label = _ "length"
```
stevecotton, line 112: Examining the map-generator code, the meaning of this appears to be "approximate width and height of the map (which will be roughly square)". It's not ready for translation, because the English doesn't match the meaning.

data/campaigns/Secrets_of_the_Ancients/gui/zombie_recruit_dialog.cfg
```diff
@@ -179,7 +179,7 @@
                 horizontal_alignment = "left"
                 [label]
                     definition = "default_small"
-                    label = "   undead"
```
stevecotton, line 182: These should probably use the translations from the `wesnoth-help` domain, (https://wiki.wesnoth.org/GettextForWesnothDevelopers#Reusing_mainline_translations).

data/campaigns/Descent_Into_Darkness/scenarios/07a_A_Small_Favor.cfg
stevecotton: I'd guess `+1`, will work in most languages, including the non-Latin ones, based on looking at the translations of ANL's `Farmers produce +1 gold`. A few are converting the number to text, but that looks like a translator's choice.

However, for the ones that translate it to text, they probably need to know the type of object that it refers to, so that they can use the right gender (etc).

Arabic would likely add a right-to-left text marker.

data/campaigns/World_Conquest/gui/help_dialog.cfg
CelticMinstrel: That's an entirely different campaign, so it has nothing to do with this, but… what is it translated _to_? If it's just translated to the same thing, that may be an indication that it doesn't need to be translated (and there should be a translation mark removed in SotA wherever it's used). If it's translated to something different, that _may_ be an indication that it doesn't need to be translated, but I'd probably want to ask the translator about that.

data/campaigns/Descent_Into_Darkness/scenarios/07a_A_Small_Favor.cfg
CelticMinstrel: I think converting the number to text is simply incorrect – translators shouldn't be doing that.

Anything that doesn't involve rendering the numbers as text can be handled by numeric localization instead of gettext. Hence, my proposal to add `amount=` which formats the number according to the active locale (using the correct digit set, potentially adding commas or the like, etc) and uses that as the text.

data/campaigns/World_Conquest/gui/settings_dialog.cfg
CelticMinstrel: Yes, the underscore here should be replaced by a space, and some of the other labels may need improvement, perhaps.

data/campaigns/World_Conquest/gui/settings_dialog.cfg
CelticMinstrel: Maybe it should be "size"?

data/campaigns/Secrets_of_the_Ancients/gui/zombie_recruit_dialog.cfg
CelticMinstrel: And if the extra leading whitespace is required, then it should be nontranslatable ie `"  " + _"undead"`.

data/campaigns/Descent_Into_Darkness/scenarios/07a_A_Small_Favor.cfg
stevecotton: I'm not seeing the use-case for `amount=`. For all the languages that Wesnoth supports,  I think the "correct digit set" is Arabic Numerals, and I'd be surprised if any language uses commas for the range of numbers that are likely to appear in floating texts. Assuming it will be hitpoints or gold, it's unlikely to be large enough for a separator.

The edge case would be right-to-left languages, where we need to know which direction to display a two-digit number in. @hermestrismi, do you have any comments on that please?

data/campaigns/World_Conquest/gui/help_dialog.cfg
stevecotton: It's not translated in the zh_CN.po file, it's just used in the translators' notes that are also stored in that file.

As for whether this one should be translatable, I'd say "definitely not". If it's to be a translatable string, it needs a `# po:` hint to tell translators what it means in English, and in this case that could be "decoration separating two columns of a table, showing a relationship between the two items that's defined by WC's recruitment mechanics".

data/campaigns/tutorial/gui/character_selection.cfg
stevecotton: Good idea, logged it as #6995.

data/campaigns/Descent_Into_Darkness/scenarios/07a_A_Small_Favor.cfg
CelticMinstrel: But the thing is that using numeric localization handles all of those edge-cases for us. Sure, it's unlikely that the amount will be more than 999, but by using numeric localization, we ensure it's correct even if it _does_ get that large (which could happen in something like Legend of the Invincibles, for example). By using numeric localization, I _think_ we'd ensure that right-to-left languages are also handled correctly. Sure, all current languages use arabic numerals, but if we use the numeric localization system, it will just automatically use the correct alternate digit set if some other new language is added.

Basically, all your arguments in that post boil down to "we don't need it because it makes no difference in our limited usage case". But it's still correct even in our limited usage case, and if we _don't_ do it, we're potentially going to be in a situation someday where it's incorrect because we didn't do it. I'd even go so far to argue that _all_ numbers ever displayed by the engine should pass through numeric localization, which might be kind of a chore since they're probably scattered throughout the code… searching for `std::to_string` would probably find most of them, though.

data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
Dou1622: Revert whitespace change

Dou1622: Revert whitespace change

data/core/terrain.cfg
```diff
@@ -1377,7 +1377,7 @@ Occasionally caves are <italic>text='illuminated'</italic>."
     id=cavewall_earthy_hewn
     name= _ "Cave Wall"
 #ifndef SCHEMA_VALIDATION
-    editor_name= _ "Reinforced Earthy Cave Wall"
+    editor_name="Reinforced Earthy Cave Wall"
```
CelticMinstrel, line 1380: Don't you need a wmllint ignore comment here?

data/campaigns/World_Conquest/gui/help_dialog.cfg
Dou1622: @stevecotton Thank you, then should I keep the markcheck off here?

Pentarctagon: I'd say to keep the markcheck off here.

data/campaigns/Descent_Into_Darkness/scenarios/07a_A_Small_Favor.cfg
Pentarctagon: Either way, it sounds like this is "correct enough" as it is, and the other solutions mentioned should be their own issue/PR.

Pentarctagon:
It looks like all that remains for this PR is the last to review comments about the WC settings dialog?

Wedge009:
Please, can someone explain to me how to display the WC settings dialogue?

Other than that, I think I've fixed the white-space and indentation issues - not sure why the macOS release build hit the time-out limit, though (in previous checks that particular build ran fine).

CelticMinstrel:
> Please, can someone explain to me how to display the WC settings dialogue?

It's shown when generating a random map in the map editor. I'm not sure if you can see it by clicking on Settings when starting a game.

Wedge009:
Thanks, I'll try - the code did suggest it was only active in editor mode. When in MP lobby (at least for local MP) the map randomisation settings was greyed out.

Wedge009:
I only see it for WC 3p, but most certainly these strings should be marked for translation. I'll update soon.

Wedge009:
Okay, so 3p looks like a misnomer, but the various labels are also confusing. One might argue it's a separate issue, but I suppose if we're marking these for translation we should try to make them more explanatory as well.

* length: As steve pointed out, this is the approximate width of the map which is square by design.
* villages: It's not a direct count of villages, but higher numbers give more villages.
* castle: Larger numbers provide more castle tiles for recruiting at each keep.
* iterations: I don't see any pattern in changing this, but I assume iterations is an adequate label.
* hill_size: At low numbers the overall terrain is mainly flats, but has more variations of hills, etc, at higher numbers. I really can't discern a pattern here.
* island: The larger this factor is, the smaller the island generated. At factor 1, the whole map is land-based.
* castles: Number of keeps/players. Too large a number (including if the island is too small) and this results in an error, but that's a separate issue.

I think these labels should be capitalised, but the other labels are lower-case so I've followed suit.

If there are no further suggestions for the WC settings, then I think we're done here?

CelticMinstrel:
* length -> size?
* villages -> maybe it's "village density"? I'm not sure if that fits.

Wedge009:
Density works for me.

Wedge009:
Ugh, so the most recent check failed due to time-out but the prior commit succeeded. The only change was text.

Assuming the checks are fine, are there any other suggestions/objections before merging?

PR #6967, opened 2022-08-15
Replace RC4 credentials encryption with AES.

Pentarctagon:
Starting with OpenSSL 3, RC4 is not enabled/accessible by default.

---

The remaining issue is what to do for Apple systems. I could try to update it to also use AES though I have no way of testing to know if it works or not, or I could leave Apple systems as using RC4 which would break credentials file compatibility when synced with other OSes.


src/preferences/credentials.cpp
```diff
@@ -245,15 +246,24 @@ namespace preferences
...
 		} catch(const filesystem::io_exception&) {
 			ERR_CFG << "error writing to credentials file '" << filesystem::get_credentials_file() << "'";
 		}
 	}
 }
 
-// TODO: Key-stretching (bcrypt was recommended)
+/**
+ * Fills a secure_buffer with 32 bytes of junk, then overwrites it with the system login name, server login name, and server name.
+ * If this is more than 32 bytes, then it's truncated. If it's less than 32 bytes, then the pre-generated junk is used to pad it.
```
CelticMinstrel, line 259: The wording here makes it sound broken. The key needs to be deterministic, but the wording makes it sound like the junk is random (which it isn't).

CelticMinstrel:
I don't really get why you removed abstracting away the algorithm, but whatever. I mostly have no opinion on this. I recall choosing RC4 solely because it was symmetric and thus relatively simple to work with. (In fact, I think an earlier revision used straight XOR encryption.)

src/preferences/credentials.cpp
Pentarctagon: Updated.

Pentarctagon:
Apple code re-added, which should probably work.

src/preferences/credentials.cpp
```diff
@@ -245,15 +246,24 @@ namespace preferences
...
 		}
 	}
 }
 
-// TODO: Key-stretching (bcrypt was recommended)
+/**
+ * Fills a secure_buffer with 32 bytes of deterministically generated bytes, then overwrites it with the system login name, server login name, and server name.
+ * If this is more than 32 bytes, then it's truncated. If it's less than 32 bytes, then the pre-generated bytes are used to pad it.
+ *
+ * TODO: Key-stretching (bcrypt was recommended)
```
stevecotton, line 261: Does this TODO still make sense in the code, now we've already accepted that it's a trivial (hardcoded key) encryption?

Pentarctagon: Not really, IMO.

Pentarctagon:
Barring any other comments, I think this is ready to merge. The one question being whether someone on an Apple system wants to test this out before merging or after?

PR #6968, opened 2022-08-16
Add "attack_alignment" weapon special/ability for specific attacks

newfrenchy83:
Although it is possible to mimic the operation of an alignment of the unit according to the time of day in standard conditions with [damage], the use of a fixed value limits the use of this substitutes for pre-determined conditions, but in the case of deep caves or large illuminations, the absolute value is no longer 25 and the special damage continues to mimic the standard values. That's why I propose a [attack_alignment] which actually replicates the alignment for a single attack.
The objective being that a unit can have attacks of alignment different from the standard alignment.

for [unit_alignment] the changes affect all attacks but appear in UI with name of alignment modified

### DIFF
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_alignment.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_alignment.cfg
@@ -1,0 +1,58 @@
+#textdomain wesnoth-test
+
+{COMMON_KEEP_A_B_UNIT_TEST "special_alignment_test" (
+    [event]
+        name=start
+        [modify_unit]
+            [filter]
+            [/filter]
+            [effect]
+                apply_to = new_ability
+                [abilities]
+                    {ABILITY_ILLUMINATES}
+                [/abilities]
+            [/effect]
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [attack_alignment]
+                        alignment=lawful
+                    [/attack_alignment]
+                [/set_specials]
+            [/effect]
+        [/modify_unit]
+
+        {ATTACK_AND_VALIDATE 125}
+        {SUCCEED}
+    [/event]
+)}
+
+{COMMON_KEEP_A_B_UNIT_TEST "special_custom_alignment_test" (
+    [event]
+        name=start
+        [modify_unit]
+            [filter]
+            [/filter]
+            [effect]
+                apply_to = new_ability
+                [abilities]
+                    {ABILITY_ILLUMINATES}
+                [/abilities]
+            [/effect]
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [attack_alignment]
+                        alignment=lawful
+                        multiply=2
+                    [/attack_alignment]
+                [/set_specials]
+            [/effect]
+        [/modify_unit]
+
+        {ATTACK_AND_VALIDATE 150}
+        {SUCCEED}
+    [/event]
+)}
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -187,7 +187,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 
 	// Time of day bonus.
 	damage_multiplier += combat_modifier(
-			resources::gameboard->units(), resources::gameboard->map(), u_loc, u.alignment(), u.is_fearless());
+			resources::gameboard->units(), resources::gameboard->map(), u_loc, weapon->alignment_in_attack(), u.is_fearless());
 
 	// Leadership bonus.
 	int leader_bonus = under_leadership(u, u_loc, weapon, opp_weapon);
@@ -316,7 +316,7 @@ battle_context_unit_stats::battle_context_unit_stats(const unit_type* u_type,
 	int base_damage = weapon->modified_damage();
 	int damage_multiplier = 100;
 	damage_multiplier
-			+= generic_combat_modifier(lawful_bonus, u_type->alignment(), u_type->musthave_status("fearless"), 0);
+			+= generic_combat_modifier(lawful_bonus, weapon->alignment_in_attack(), u_type->musthave_status("fearless"), 0);
 	damage_multiplier *= opp_type->resistance_against(weapon->type(), !attacking);
 
 	damage = round_damage(base_damage, damage_multiplier, 10000);
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -186,8 +186,9 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	int damage_multiplier = 100;
 
 	// Time of day bonus.
+	std::pair<unit_alignments::type, int> alignment = weapon->specials_alignment();
 	damage_multiplier += combat_modifier(
-			resources::gameboard->units(), resources::gameboard->map(), u_loc, weapon->alignment_in_attack(), u.is_fearless());
+			resources::gameboard->units(), resources::gameboard->map(), u_loc, alignment.first, u.is_fearless(), alignment.second);
 
 	// Leadership bonus.
 	int leader_bonus = under_leadership(u, u_loc, weapon, opp_weapon);
@@ -1583,23 +1584,23 @@ int combat_modifier(const unit_map& units,
 		const gamemap& map,
 		const map_location& loc,
 		unit_alignments::type alignment,
-		bool is_fearless)
+		bool is_fearless, int special_bonus)
 {
 	const tod_manager& tod_m = *resources::tod_manager;
 	const time_of_day& effective_tod = tod_m.get_illuminated_time_of_day(units, map, loc);
-	return combat_modifier(effective_tod, alignment, is_fearless);
+	return combat_modifier(effective_tod, alignment, is_fearless, special_bonus);
 }
 
 int combat_modifier(const time_of_day& effective_tod,
 		unit_alignments::type alignment,
-		bool is_fearless)
+		bool is_fearless, int special_bonus)
 {
 	const tod_manager& tod_m = *resources::tod_manager;
 	const int lawful_bonus = effective_tod.lawful_bonus;
-	return generic_combat_modifier(lawful_bonus, alignment, is_fearless, tod_m.get_max_liminal_bonus());
+	return generic_combat_modifier(lawful_bonus, alignment, is_fearless, tod_m.get_max_liminal_bonus(), special_bonus);
 }
 
-int generic_combat_modifier(int lawful_bonus, unit_alignments::type alignment, bool is_fearless, int max_liminal_bonus)
+int generic_combat_modifier(int lawful_bonus, unit_alignments::type alignment, bool is_fearless, int max_liminal_bonus, int special_bonus)
 {
 	int bonus;
 
@@ -1620,6 +1621,7 @@ int generic_combat_modifier(int lawful_bonus, unit_alignments::type alignment, b
 		bonus = 0;
 	}
 
+	bonus *= special_bonus;
 	if(is_fearless) {
 		bonus = std::max<int>(bonus, 0);
 	}
--- a/src/actions/attack.hpp
+++ b/src/actions/attack.hpp
@@ -279,7 +279,7 @@ int combat_modifier(const unit_map& units,
 		const gamemap& map,
 		const map_location& loc,
 		unit_alignments::type alignment,
-		bool is_fearless);
+		bool is_fearless, int special_bonus = 1);
 
 /**
  * Returns the amount that a unit's damage should be multiplied by
@@ -287,13 +287,13 @@ int combat_modifier(const unit_map& units,
  */
 int combat_modifier(const time_of_day& effective_tod,
 		unit_alignments::type alignment,
-		bool is_fearless);
+		bool is_fearless, int special_bonus = 1);
 
 /**
  * Returns the amount that a unit's damage should be multiplied by
  * due to a given lawful_bonus.
  */
-int generic_combat_modifier(int lawful_bonus, unit_alignments::type alignment, bool is_fearless, int max_liminal_bonus);
+int generic_combat_modifier(int lawful_bonus, unit_alignments::type alignment, bool is_fearless, int max_liminal_bonus, int special_bonus = 1);
 /**
  * Function to check if an attack will satisfy the requirements for backstab.
  * Input:
--- a/src/gui/dialogs/attack_predictions.cpp
+++ b/src/gui/dialogs/attack_predictions.cpp
@@ -225,7 +225,7 @@ void attack_predictions::set_data(window& window, const combatant_data& attacker
 	const unit& u = *attacker.unit_;
 
 	const int tod_modifier = combat_modifier(resources::gameboard->units(), resources::gameboard->map(),
-		u.get_location(), u.alignment(), u.is_fearless());
+		u.get_location(), weapon->alignment_in_attack(), u.is_fearless());
 
 	if(tod_modifier != 0) {
 		set_label_helper("tod_modifier", utils::signed_percent(tod_modifier));
--- a/src/gui/dialogs/attack_predictions.cpp
+++ b/src/gui/dialogs/attack_predictions.cpp
@@ -224,8 +224,9 @@ void attack_predictions::set_data(window& window, const combatant_data& attacker
 	// Time of day modifier.
 	const unit& u = *attacker.unit_;
 
+	std::pair<unit_alignments::type, int> alignment = weapon->specials_alignment();
 	const int tod_modifier = combat_modifier(resources::gameboard->units(), resources::gameboard->map(),
-		u.get_location(), weapon->alignment_in_attack(), u.is_fearless());
+		u.get_location(), alignment.first, u.is_fearless(), alignment.second);
 
 	if(tod_modifier != 0) {
 		set_label_helper("tod_modifier", utils::signed_percent(tod_modifier));
--- a/src/reports.cpp
+++ b/src/reports.cpp
@@ -786,8 +786,8 @@ static int attack_info(const reports::context& rc, const attack_type &at, config
 		int specials_damage = at.modified_damage();
 		int damage_multiplier = 100;
 		const_attack_ptr weapon  = at.shared_from_this();
-		unit_alignments::type alignment = weapon->alignment_in_attack();
-		int tod_bonus = combat_modifier(get_visible_time_of_day_at(rc, hex), alignment, u.is_fearless());
+		std::pair<unit_alignments::type, int> alignment = weapon->specials_alignment();
+		int tod_bonus = combat_modifier(get_visible_time_of_day_at(rc, hex), alignment.first, u.is_fearless(), alignment.second);
 		damage_multiplier += tod_bonus;
 		int leader_bonus = under_leadership(u, hex, weapon);
 		if (leader_bonus != 0)
@@ -957,9 +957,9 @@ static int attack_info(const reports::context& rc, const attack_type &at, config
 		add_text(res, damage_and_num_attacks.str, damage_and_num_attacks.tooltip);
 		add_text(res, damage_versus.str, damage_versus.tooltip); // This string is usually empty
 
-		if(alignment != u.alignment()){
-			const std::string align = unit_type::alignment_description(alignment, u.gender());
-			const std::string align_id = unit_alignments::get_string(alignment);
+		if(alignment.first != u.alignment() || alignment.second != 1){
+			const std::string align = unit_type::alignment_description(alignment.first, u.gender());
+			const std::string align_id = unit_alignments::get_string(alignment.first);
 
 			color_t color = font::weapon_color;
 			if (tod_bonus != 0)
--- a/src/reports.cpp
+++ b/src/reports.cpp
@@ -786,7 +786,8 @@ static int attack_info(const reports::context& rc, const attack_type &at, config
 		int specials_damage = at.modified_damage();
 		int damage_multiplier = 100;
 		const_attack_ptr weapon  = at.shared_from_this();
-		int tod_bonus = combat_modifier(get_visible_time_of_day_at(rc, hex), u.alignment(), u.is_fearless());
+		unit_alignments::type alignment = weapon->alignment_in_attack();
+		int tod_bonus = combat_modifier(get_visible_time_of_day_at(rc, hex), alignment, u.is_fearless());
 		damage_multiplier += tod_bonus;
 		int leader_bonus = under_leadership(u, hex, weapon);
 		if (leader_bonus != 0)
@@ -955,6 +956,23 @@ static int attack_info(const reports::context& rc, const attack_type &at, config
 		}
 		add_text(res, damage_and_num_attacks.str, damage_and_num_attacks.tooltip);
 		add_text(res, damage_versus.str, damage_versus.tooltip); // This string is usually empty
+
+		if(alignment != u.alignment()){
+			const std::string align = unit_type::alignment_description(alignment, u.gender());
+			const std::string align_id = unit_alignments::get_string(alignment);
+
+			color_t color = font::weapon_color;
+			if (tod_bonus != 0)
+				color = (tod_bonus > 0) ? font::good_dmg_color : font::bad_dmg_color;
+
+			str << align << " (" << span_color(color) << utils::signed_percent(tod_bonus)
+				<< naps << ")";
+
+			tooltip << _("Alignment: ") << "<b>" << align << "</b>\n"
+				<< string_table[align_id + "_description"];
+
+			add_text(res, flush(str), flush(tooltip));
+		}
 
 		const std::string &accuracy_parry = at.accuracy_parry_description();
 		if (!accuracy_parry.empty())
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1297,6 +1297,56 @@ unit_alignments::type attack_type::alignment_in_attack() const
 	return *align;
 }
 
+std::pair<unit_alignments::type, int> attack_type::specials_alignment() const
+{
+	unit_alignments::type alignment = alignment_in_attack();
+	unit_ability_list align_list = get_specials_and_abilities("attack_alignment");
+	if(align_list.empty()){
+		return {alignment, 1};
+	}
+	std::map<std::string, unit_ability_list> lists;
+	std::map<std::string, bool> is_customised;
+	for(const auto& i : align_list) {
+		if(checking_alignment().count((*i.ability_cfg)["alignment"].str()) != 0){
+			const auto& alignment = (*i.ability_cfg)["alignment"].str();
+			lists[alignment].emplace_back(i);
+			if(!is_customised[alignment]){
+				if((*i.ability_cfg).has_attribute("value") || (*i.ability_cfg).has_attribute("add") || (*i.ability_cfg).has_attribute("sub") || (*i.ability_cfg).has_attribute("multiply") || (*i.ability_cfg).has_attribute("divide")){
+					is_customised[alignment] = true;
+				}
+			}
+		}
+	}
+	if(lists.empty()){
+		return {alignment, 1};
+	}
+	int count_liminal = lists["liminal"].size();
+	int count_chaotic = lists["chaotic"].size();
+	int count_lawful = lists["lawful"].size();
+	int count_neutral = lists["neutral"].size();
+	std::string str_alignment;
+	if(count_liminal > 0 && count_liminal >= count_chaotic && count_liminal >= count_lawful && count_liminal >= count_neutral){
+		str_alignment = "liminal";
+	}
+	else if((count_neutral > 0 && count_neutral > count_chaotic && count_neutral > count_lawful) || (count_lawful > 0 && count_lawful == count_chaotic)){
+		str_alignment = "neutral";
+	}
+	else if( count_lawful > 0 && count_lawful > count_chaotic){
+		str_alignment = "lawful";
+	}
+	else if( count_chaotic > 0 && count_chaotic > count_lawful){
+		str_alignment = "chaotic";
+	}
+
+	auto new_align = unit_alignments::get_enum(str_alignment);
+	if(new_align) {
+		alignment = *new_align;
+	}
+	int align_value = is_customised[str_alignment] ? (std::max(0, composite_value(lists[str_alignment], 100))/100) : 1;
+
+	return {alignment, align_value};
+}
+
 
 namespace { // Helpers for attack_type::special_active()
 
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1284,6 +1284,19 @@ int attack_type::modified_damage() const
 	return damage_value;
 }
 
+unit_alignments::type attack_type::alignment_in_attack() const
+{
+	std::optional<unit_alignments::type> align = unit_alignments::get_enum(attack_alignment());
+	if(self_){
+		//here the value returned can't be null, if self_ exist like in attack.cpp and gui and report functions, unit alignment returned if attack_alignment empty or don't match regular alignment.
+		//if this function should be called without unit specified then "neutral" returned instead of empty or incorrect value(in case or function called in other file in future).
+		align = (!attack_alignment().empty() && (checking_alignment().count(attack_alignment()) != 0)) ? unit_alignments::get_enum(attack_alignment()) : (*self_).alignment();
+	} else {
+		align = (!attack_alignment().empty() && (checking_alignment().count(attack_alignment()) != 0)) ? unit_alignments::get_enum(attack_alignment()) : unit_alignments::type::neutral;
+	}
+	return *align;
+}
+
 
 namespace { // Helpers for attack_type::special_active()
 
--- a/src/units/attack_type.hpp
+++ b/src/units/attack_type.hpp
@@ -91,6 +91,9 @@ public:
 	/** Returns alignment specified by 'attack_alignment' If empty and have unit's alignment returns the unit's alignment.
 	 */
 	unit_alignments::type alignment_in_attack() const;
+	/** Returns alignment used in [attack_alignment] and percentage of alignment bonus. If no such specials are active, it returns the unit's alignment.
+	 */
+	std::pair<unit_alignments::type, int> specials_alignment() const;
 
 	/** Calculates the number of attacks this weapon has, considering specials. */
 	void modified_attacks(unsigned & min_attacks,
--- a/src/units/attack_type.hpp
+++ b/src/units/attack_type.hpp
@@ -25,6 +25,7 @@
 #include <boost/dynamic_bitset_fwd.hpp>
 
 #include "units/ptr.hpp" // for attack_ptr
+#include "units/unit_alignments.hpp"
 
 class unit_ability_list;
 class unit_type;
@@ -59,6 +60,7 @@ public:
 	void set_type(const std::string& value) { type_ = value; set_changed(true); }
 	void set_icon(const std::string& value) { icon_ = value; set_changed(true); }
 	void set_range(const std::string& value) { range_ = value; set_changed(true); }
+	void set_attack_alignment(const std::string& value) { attack_alignment_ = value; set_changed(true); }
 	void set_accuracy(int value) { accuracy_ = value; set_changed(true); }
 	void set_parry(int value) { parry_ = value; set_changed(true); }
 	void set_damage(int value) { damage_ = value; set_changed(true); }
@@ -82,6 +84,13 @@ public:
 	std::vector<std::pair<t_string, t_string>> special_tooltips(boost::dynamic_bitset<>* active_list = nullptr) const;
 	std::string weapon_specials() const;
 	std::string weapon_specials_value(const std::set<std::string> checking_tags) const;
+
+	const std::set<std::string>& checking_alignment() const { return checking_alignment_; }
+
+	const std::string& attack_alignment() const { return attack_alignment_; }
+	/** Returns alignment specified by 'attack_alignment' If empty and have unit's alignment returns the unit's alignment.
+	 */
+	unit_alignments::type alignment_in_attack() const;
 
 	/** Calculates the number of attacks this weapon has, considering specials. */
 	void modified_attacks(unsigned & min_attacks,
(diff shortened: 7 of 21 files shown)


stevecotton:
Please add unit tests.

src/actions/attack.cpp
```diff
@@ -1604,6 +1604,63 @@ int under_leadership(const unit &u, const map_location& loc, const_attack_ptr we
...
+	if(!has_special_or_ability("special_alignment")){
+		return combat_modifier;
+	}
+	// if "alignment" attribute not eqauls to string in checking_alignment chain,
+	// or not "special_alignment special/abilities then alignment_type returned by default.
+	const std::set<std::string> checking_alignment = {"neutral", "lawful", "chaotic", "liminal"};
+	unit_ability_list abil_list = get_specials_and_abilities("special_alignment");
+	for(unit_ability_list::iterator i = abil_list.begin(); i != abil_list.end();) {
+		if(checking_alignment.count((*i->ability_cfg)["alignment"].str()) == 0){
+			i = abil_list.erase(i);
```
stevecotton, line 1618: IIUC, this code would allow a weapon to have a special that made the attack `lawful`, and also a special that made the same attack `chaotic`. This code would allow both of those through, and later would calculate the `lawful_value` and `chaotic_value`, and finally would add them together at which time they would cancel each other out.

For specials that change the damage type, for example the Holy Water item, I think it just uses "last one added wins". Following that pattern, if a unit picked up "make your melee attacks lawful" followed by "make your melee attacks chaotic", I'd expect the melee attacks to get the chaotic bonus.

I think that could simplify the code too - it could use the `for` loop's ranged syntax because it wouldn't need to erase items from the list, just store the last-seen alignment. Some of the later code might then be replaced with a call to `generic_combat_modifier()`.

newfrenchy83: If many specials presnet how choose alignment used (i think in 'leadership' cases)?

stevecotton: Good question. I guess it could happen for leadership-style changes of the damage type too, if there was a Mage of Light with "bless all adjacent units' weapons (change to `arcane` damage)" and another with "all adjacent units' weapons ignore armour (change to Era of Magic's `secret` damage type)".

data/test/scenarios/special_alignment.cfg
```diff
@@ -0,0 +1,123 @@
...
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        [unstore_unit]
+            variable=a
+            find_vacant=yes
+            x,y=$b.x,$b.y
```
stevecotton, line 84: Using `COMMON_KEEP_A_B_UNIT_TEST` or `COMMON_KEEP_A_B_C_D_UNIT_TEST` instead of `GENERIC_UNIT_TEST` means that they're already in place to attack each other.

Thanks for adding the test.

CelticMinstrel:
Why `[special_alignment]` and not just `[alignment]`?

newfrenchy83:
> Why `[special_alignment]` and not just `[alignment]`?

nemr changed to [alignment]

src/actions/attack.cpp
```diff
@@ -1604,6 +1604,63 @@ int under_leadership(const unit &u, const map_location& loc, const_attack_ptr we
...
+		} else {
+			++i;
+		}
+	}
+	if(abil_list.empty()){
+		return combat_modifier;
+	}
```
zishiwu123, line 1625: I think this check should occur before the for loop. It can be combined with the `!has_special_or_ability("alignment")` check.

src/actions/attack.cpp
```diff
@@ -1604,6 +1604,63 @@ int under_leadership(const unit &u, const map_location& loc, const_attack_ptr we
...
+
+	//if list no empty then lawful_bonus and max_liminal_bonus calculated and aplied to alignment specified
+	assert(display::get_singleton());
```
zishiwu123, line 1628: I think the calculations from lines 1629 through 1655 can be put into a separate function `get_lawful_and_max_liminal_bonus(unit_ability_list abil_list)` to make the code more modular.

src/actions/attack.cpp
zishiwu123, line 1625: You can put this check inside the `get_lawful_and_max_liminal_bonus` function and have it return 0 if `abil_list.empty()` is true.

zishiwu123 (commented):
I think we can make this code more modular and easier to maintain by separating the calculation of lawful/chaotic and max liminal bonus into a separate function
```cpp
int lawful_and_max_liminal_bonus(unit_ability_list abil_list) const
```
that takes in a unit_ability_list and returns the sum of those bonuses. What do you think?

src/actions/attack.cpp
```diff
@@ -1604,6 +1604,63 @@ int under_leadership(const unit &u, const map_location& loc, const_attack_ptr we
...
+		}
+		else if(chaotic_value == 0 && (*i.ability_cfg)["alignment"].str() == "chaotic") {
+			chaotic_value = -lawful_bonus;
+		}
+		else if(liminal_value == 0 && (*i.ability_cfg)["alignment"].str() == "liminal") {
+			liminal_value = max_liminal_bonus - std::abs(lawful_bonus);
+		}
+	}
+	//final calculation of bonus_value equals to sum of different alignment present in the list.
+	bonus_value = combat_modifier + lawful_value + chaotic_value + liminal_value;
```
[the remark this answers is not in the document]
newfrenchy83: If lawful bonus and chaotic bonus both calculated inside function with liminal bonus, pehaps.

src/actions/attack.cpp
```diff
@@ -1604,57 +1605,54 @@ int under_leadership(const unit &u, const map_location& loc, const_attack_ptr we
...
+	//check if alignment of unit value can be or not used
+	for(const auto& i : abil_list) {
+		if(combat_modifier != 0 && (*i.ability_cfg)["no_alignment_type"].to_bool()) {
+			combat_modifier = 0;
+			break;
+		}
+	}
+}
+
+static int lawful_and_max_liminal_bonus(unit_ability_list& abil_list, const std::string& alignment, int bonus_value)
```
zishiwu123, line 1627: Awesome work! But my bad, I think the name I originally proposed, `lawful_and_max_liminal_bonus`, is not a good name. It would be better to call it `get_combat_bonus`.

zishiwu123:
Good work! Unfortunately after thinking some more, I realize lawful_and_max_liminal_bonus is not a good function name. How about alignment_bonus instead?

Pentarctagon:
@zishiwu123 Who are you?

zishiwu123:
> @zishiwu123 Who are you?

I'm a fan of the Wesnoth game who found out about it a month ago and has been hooked on playing it. Now I'd like to help out the developers. I figured a good way to learn about the code would be to review the PRs.

Pentarctagon:
It's generally best to learn the code either by reading through parts you're interested in first, or taking a look through the Good First Issue tag and working on one of those. It's a bit odd to add code reviews about code you're not yet familiar with.

We also have a [forums](https://forums.wesnoth.org/), [Discord server](https://discord.gg/battleforwesnoth), and [IRC channels](https://wiki.wesnoth.org/Support#IRC) if you have questions about anything.

zishiwu123:
> It's generally best to learn the code either by reading through parts you're interested in first, or taking a look through the Good First Issue tag and working on one of those. It's a bit odd to add code reviews about code you're not yet familiar with.
> 
> 
> 
> We also have a [forums](https://forums.wesnoth.org/), [Discord server](https://discord.gg/battleforwesnoth), and [IRC channels](https://wiki.wesnoth.org/Support#IRC) if you have questions about anything.

Thanks for the pointers. I'll look at the forums and good first issues. Also I agree with your point that reviewing code that I'm not familiar with is a bit odd. My bad if I caused any confusion.

stevecotton:
Adding my current thoughts, but please don't read this as "changes required", it's just an update because I've not got review feedback ready yet. Definitely wait for @CelticMinstrel to comment on this.

The previously-discussed leadership-like ability to change other units' weapons' alignments works (code below). Now that I've seen it in code, I'm wondering if the tag name should be `[weapon_alignment]` or `[attack_alignment]`, to leave open the possibility of later adding a similar `[unit_alignment]`.

If multiple leadership-like abilities of this type have conflicting results, such as one changing to `liminal` and the other to `lawful`, there's an open question of what happens. The current code will handle it without OOS, but with surprising implementation detail that `lawful` and `chaotic` might cancel each other out, while `liminal` still works.

I feel that alignment and damage-type should be handled similarly, but at the moment I think it's not possible to create a leadership-like ability for a Mage of Light to bless adjacent units' weapons with `arcane` damage. If we add this PR then there should be a follow-up PR to make that possible. I find the damage types easier to think about from a lore perspective than the alignments, I can easily see a loyalists vs undead scenario with the Mage of Light and a Sargent with a similar ability that gives fire damage - in that case, it should be up to the scenario author to ensure that the abilities don't conflict (so don't apply the Sargent's abilities to arcane weapons), and I think we'd need to accept OOS if the scenario author didn't ensure that.

With the damage-type abilities, I would expect the engine's code to ensure that ability-based and non-ability-based changes couldn't conflict with each other. In other words, that picking up a Holy Water item and then being affected by the Sargent's power couldn't cause OOS, even if the scenario author forgot to add a filter to prevent that problem.

```
        [object]
            name=_ "liminal_weapons"
            duration=forever
            [filter]
                id=charlie
            [/filter]
            [effect]
                apply_to=new_ability
                [abilities]
                    [alignment]
                        id="liminal_by_charlie"
                        alignment=liminal
                        name="liminal_by_charlie"
                        affect_self=yes
                        affect_allies=yes
                        affect_enemies=yes
                        [affect_adjacent]
                        [/affect_adjacent]
                    [/alignment]
                [/abilities]
            [/effect]
        [/object]
```

Pentarctagon:
At a minimum, there should be an error logged if it's known that something will cause an OOS. It's already hard enough to figure out what is causing an OOS error, so we shouldn't knowingly be adding more ways to cause them without very compelling reasons.

newfrenchy83:
> Adding my current thoughts, but please don't read this as "changes required", it's just an update because I've not got review feedback ready yet. Definitely wait for @CelticMinstrel to comment on this.
> 
> The previously-discussed leadership-like ability to change other units' weapons' alignments works (code below). Now that I've seen it in code, I'm wondering if the tag name should be `[weapon_alignment]` or `[attack_alignment]`, to leave open the possibility of later adding a similar `[unit_alignment]`.
> 
> If multiple leadership-like abilities of this type have conflicting results, such as one changing to `liminal` and the other to `lawful`, there's an open question of what happens. The current code will handle it without OOS, but with surprising implementation detail that `lawful` and `chaotic` might cancel each other out, while `liminal` still works.
> 
> I feel that alignment and damage-type should be handled similarly, but at the moment I think it's not possible to create a leadership-like ability for a Mage of Light to bless adjacent units' weapons with `arcane` damage. If we add this PR then there should be a follow-up PR to make that possible. I find the damage types easier to think about from a lore perspective than the alignments, I can easily see a loyalists vs undead scenario with the Mage of Light and a Sargent with a similar ability that gives fire damage - in that case, it should be up to the scenario author to ensure that the abilities don't conflict (so don't apply the Sargent's abilities to arcane weapons), and I think we'd need to accept OOS if the scenario author didn't ensure that.
> 
> With the damage-type abilities, I would expect the engine's code to ensure that ability-based and non-ability-based changes couldn't conflict with each other. In other words, that picking up a Holy Water item and then being affected by the Sargent's power couldn't cause OOS, even if the scenario author forgot to add a filter to prevent that problem.
> 
> ```
>         [object]
>             name=_ "liminal_weapons"
>             duration=forever
>             [filter]
>                 id=charlie
>             [/filter]
>             [effect]
>                 apply_to=new_ability
>                 [abilities]
>                     [alignment]
>                         id="liminal_by_charlie"
>                         alignment=liminal
>                         name="liminal_by_charlie"
>                         affect_self=yes
>                         affect_allies=yes
>                         affect_enemies=yes
>                         [affect_adjacent]
>                         [/affect_adjacent]
>                     [/alignment]
>                 [/abilities]
>             [/effect]
>         [/object]
> ```

separate [alignment] in [attack_alignment] in place of alignment but without attribute and [unit_alignment] for impose the calculation of unit_alignment when [attack_alignment is present? Ok,i will try that. and i will modify the code for what prioritize liminal  on lawful or chaotic.

CelticMinstrel:
> I'm wondering if the tag name should be `[weapon_alignment]` or `[attack_alignment]`, to leave open the possibility of later adding a similar `[unit_alignment]`.

Not saying whether this is a good or bad idea, but the current tag name definitely doesn't rule out the possibility of a later unit alignment ability.

zishiwu123:
@newfrenchy83 - sorry if I caused you any confusion. @Pentarctagon I agree it was strange of me to submit a code review without having contributed before to Wesnoth. I tried to remove the review I submitted but can't find a way to do it from my account. I see it's possible for the person who submitted the PR to dismiss a review. @newfrenchy83 can you please dismiss my review? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review

newfrenchy83:
> @newfrenchy83 - sorry if I caused you and anyone else confusion. Thanks for removing me as a reviewer. @Pentarctagon I agree it was strange of me to submit a code review without having contributed before to Wesnoth. I'll be more mindful in the future.

this PR is in master repositoy, and i don't full control because i'm not administrator

newfrenchy83:
@stevecotton i create a [unit_alignment] here: https://github.com/newfrenchy83/wesnoth/pull/79

it is that what you want?

newfrenchy83:
i added a unit_alignment abilities here.

src/gui/dialogs/attack_predictions.cpp
```diff
@@ -220,8 +220,9 @@ void attack_predictions::set_data(window& window, const combatant_data& attacker
 	// Time of day modifier.
 	const unit& u = *attacker.unit_;
 
-	const int tod_modifier = combat_modifier(resources::gameboard->units(), resources::gameboard->map(),
+	int unit_combat_modifier = combat_modifier(resources::gameboard->units(), resources::gameboard->map(),
 		u.get_location(), u.alignment(), u.is_fearless());
+	const int tod_modifier = weapon->specials_alignment(unit_combat_modifier, u.is_fearless());
```
stevecotton, line 225: A lot of the logic of `specials_alignment` and `under_alignment_abilities` seems to duplicate the logic that's already in `combat_modifier` or `generic_combat_modifier`. Instead of passing `unit_combat_modifier` to `specials_alignment`, could these be done as:

```
const auto effective_alignment = weapon->specials_alignment();
const auto tod_modifier = combat_modifier(resources::gameboard->units(), resources::gameboard-->map(),
    u.get_location(), effective_alignment, u.is_fearless());
```
?

src/actions/attack.hpp
```diff
@@ -281,6 +281,8 @@ void attack_unit_and_advance(const map_location& attacker,
  */
 int under_leadership(const unit &u, const map_location& loc, const_attack_ptr weapon = nullptr, const_attack_ptr opp_weapon = nullptr);
 
+std::pair<int, std::string> under_alignment_abilities(const unit &u, const map_location& loc, int combat_modifier, bool is_fearless = false, bool use_gender = false);
```
stevecotton, line 284: Documentation would help for reviewing - would drafting it in French and then getting @CelticMinstrel or myself to translate work for you?

There's another comment about whether this should just calculate the alignment, leaving it to the caller to calculate the combat bonus for that alignment.

Instead of returning the alignment as a string, please use the `unit_alignments` templated enum.

src/reports.cpp
```diff
@@ -370,10 +370,11 @@ static config unit_alignment(reports::context & rc, const unit* u, const map_loc
 {
 	if (!u) return config();
 	std::ostringstream str, tooltip;
-	const std::string align = unit_type::alignment_description(u->alignment(), u->gender());
-	const std::string align_id = unit_alignments::get_string(u->alignment());
+	bool alignment_abilitie = ((under_alignment_abilities(*u, hex, 0).second).empty()) ? false : true;
+	const std::string align = alignment_abilitie ? translation::dsgettext("wesnoth", (under_alignment_abilities(*u, hex, 0, false, true).second).c_str()) : unit_type::alignment_description(u->alignment(), u->gender());
+	const std::string align_id = alignment_abilitie ? under_alignment_abilities(*u, hex, 0).second : unit_alignments::get_string(u->alignment());
```
stevecotton, line 375: Using the ternary operator here makes the lines long and hard to read. As a general rule, I'd prefer removing the `const` from the existing lines, and then adding a conditional block for the rare case that an ability is in effect:

```
auto align = unit_type::alignment_description(u->alignment(), u->gender());
auto align_id = unit_type::get_string(u->alignment());
if(alignment_abilitie) {
    align = ... ;
    align_id = ...;
}
```

However, in this case I think `under_alignment_abilities` should return the alignment using the `unit_alignments` enum, and the existing lines then use that instead of `u->alignment()`:
```
const auto effective_alignment = under_alignment_abilities(...);
const auto align = unit_type::alignment_description(effective_alignment, u->gender());
const auto align_id = unit_type::get_string(effective_alignment);
```

Returning the enumerated type would also mean that `under_alignment_abilities` wouldn't need to handle the gender at all - this is the only caller that uses it, and it can do it here via `alignment_description`.

newfrenchy83: i think removeunit_alignment abilities for now and wait another PR.

src/actions/attack.hpp
newfrenchy83: have you see sceenshot here:: https://github.com/newfrenchy83/wesnoth/pull/79 ?

stevecotton: I meant documentation for the function itself, describing what the arguments are and what it returns, just the part of the calculation that's done in this function. I know the functional currently does most of the calculation, but this fits with my other comments about trying to make this smaller by only calculating the alignment in this function, and leaving it to the caller to use another function to calculate the bonus.

stevecotton:
> It's already hard enough to figure out what is causing an OOS error, so we shouldn't knowingly be adding more ways to cause them without very compelling reasons.

Good point, I should have said "edge cases that the player can't anticipate" rather than "OOS". Take the example of a Mage that gives arcane damage to adjacent allies and a Sargent with a similar ability that gives fire damage. For balance and user experience, the end result should be predictable and deterministic even when both the Mage and Sargent are adjacent to a unit, and even when:

* the Mage and Sargent swap hexes
* the Mage and Sargent were recruited or recalled in the other order
* anything that's not player-visible is swapped (for example, it can't rely on sorting by unit id)

While the engine would likely avoid OOS by internally using some deterministic ordering of the units, the players aren't going to be happy if rotating the game board by 180 degrees changes the engine's sort order. Newfrenchy's code when I checked this would have lawful and chaotic alignment changes cancel each other out, which is a reasonable option.

With the damage-type abilities, I would expect the engine's code to ensure that ability-based and non-ability-based changes couldn't conflict with each other. In other words, that picking up a Holy Water item and then being affected by the Sargent's power would either always give fire damage or always give arcane damage, and be predictable about which. For this particular pairing, I'd expect the scenario author to want arcane damage to apply and to code the Sargent's abilities to not affect arcane weapons, however in general I'd expect the engine to either choose (ability-based overrides non-ability-based) or (non-ability-based overrides ability-based).

stevecotton:
Still wondering whether `specials_alignment` could return a `unit_alignments::type` or `std::optional<unit_alignments::type>`, and therefore avoid duplicating code from `combat_modifier(effective_tod, alignment, is_fearless)` or `generic_combat_modifier`. My feeling is that it feels like too much duplicated code.

newfrenchy83:
> Still wondering whether `specials_alignment` could return a `unit_alignments::type` or `std::optional<unit_alignments::type>`, and therefore avoid duplicating code from `combat_modifier(effective_tod, alignment, is_fearless)` or `generic_combat_modifier`. My feeling is that it feels like too much duplicated code.

No, else unit_alignment changed for all attack of unit, and attack-alignment must change alignment for attack choose no for all

newfrenchy83:
> Still wondering whether `specials_alignment` could return a `unit_alignments::type` or `std::optional<unit_alignments::type>`, and therefore avoid duplicating code from `combat_modifier(effective_tod, alignment, is_fearless)` or `generic_combat_modifier`. My feeling is that it feels like too much duplicated code.

it's done.

stevecotton:
> it's done.

Thanks. It needs a minor fixup, it seems the old version of alignment_bonus crept back in during your rebase, but that's the direction that I wanted it to change in.

src/actions/attack.cpp
```diff
@@ -1604,6 +1604,57 @@ int under_leadership(const unit &u, const map_location& loc, const_attack_ptr we
...
+	bool is_liminal = alignment_bonus(abil_list, "liminal");
+	bool is_neutral = alignment_bonus(abil_list, "neutral");
+	bool is_lawful = alignment_bonus(abil_list, "lawful");
+	bool is_chaotic = alignment_bonus(abil_list, "chaotic");
+	std::string str_alignment;
+	if(is_liminal){str_alignment = "liminal";}
+	else if((is_neutral && !is_lawful && !is_chaotic) || (is_lawful && is_chaotic)){str_alignment = "neutral";}
+	else if( is_lawful){str_alignment = "lawful";}
+	else if( is_chaotic){str_alignment = "chaotic";}
+	auto new_align = unit_alignments::get_enum(str_alignment);
```
stevecotton, line 1647: As it's an enum, please use the enum in the lines above instead of using strings and then converting them.

I've been doing some local experiments, will push a branch with my work-in-progress today.

src/actions/attack.cpp
```diff
@@ -1604,6 +1604,57 @@ int under_leadership(const unit &u, const map_location& loc, const_attack_ptr we
...
+	}
+	bool is_liminal = alignment_bonus(abil_list, "liminal");
+	bool is_neutral = alignment_bonus(abil_list, "neutral");
+	bool is_lawful = alignment_bonus(abil_list, "lawful");
+	bool is_chaotic = alignment_bonus(abil_list, "chaotic");
+	std::string str_alignment;
+	if(is_liminal){str_alignment = "liminal";}
+	else if((is_neutral && !is_lawful && !is_chaotic) || (is_lawful && is_chaotic)){str_alignment = "neutral";}
+	else if( is_lawful){str_alignment = "lawful";}
+	else if( is_chaotic){str_alignment = "chaotic";}
```
stevecotton, line 1646: With this block of `if` ... `else if` ... statements, there should at least one of

* an `else` at the end
* a comment saying what should happen if none are matched
* obvious handling of the default value afterwards, for example an `if(new_align) // it's not std::nullopt` at the start of the next block

stevecotton: Sorry, having added the third point during editing, I've now realised that that conditional block is already included, so you've already done what I was asking for.

src/actions/attack.cpp
```diff
@@ -1604,6 +1604,57 @@ int under_leadership(const unit &u, const map_location& loc, const_attack_ptr we
...
+	auto new_align = unit_alignments::get_enum(str_alignment);
+	const unit& up = (*self_);
+	unit &ur = const_cast<unit&>(up);
+	if(new_align) {
+		(ur).set_alignment(*new_align);
```
stevecotton, line 1651: Why does this need to change the unit's alignment, instead of just a `return *new_align;` ?

src/actions/attack.cpp
```diff
@@ -1604,6 +1604,57 @@ int under_leadership(const unit &u, const map_location& loc, const_attack_ptr we
...
+{
+	for(const auto& i : abil_list) {
+		if((*i.ability_cfg)["alignment"].str() == alignment) {
+			return {bonus_value, true};
+		}
+	}
+	return {0, false};
+}
+
+unit_alignments::type attack_type::specials_alignment() const
```
stevecotton, line 1629: This is implementing a method declared in src/units/attack_type.hpp, so should be in the src/units directory.

At the moment, most of attack_type's implementation is in src/units/abilities.cpp instead of src/units/attack_type.cpp. That seems a mess, but a mess to clean up in a different PR, so for now I'd suggest putting this in src/units/abilities.cpp too.

newfrenchy83: it's because [unit_alignment] abilitie what i want createin attack.cpp and whowill use same static functions, what i must let other in attack.cpp

src/actions/attack.cpp
newfrenchy83: removed

newfrenchy83:
@stevecotton add a separate commit here: https://github.com/newfrenchy83/wesnoth/pull/79/commits/2fe7b1f63217ee1a68cc3600c57760bbcc6dd3fb for [unit_alignment] abilitie

src/actions/attack.cpp
CelticMinstrel: In that case the correct approach is probably to make the functions not static and declare them in the associated header.

newfrenchy83: > In that case the correct approach is probably to make the functions not static and declare them in the associated header.

Functions moved to abilitie.cpp

CelticMinstrel:
If adding `[unit_alignment]` in this PR is intentional, please change the title to reflect that it's adding two new tags instead of just one.

newfrenchy83:
> If adding `[unit_alignment]` in this PR is intentional, please change the title to reflect that it's adding two new tags instead of just one.

done

newfrenchy83:
@stevecotton , @CelticMinstrel I modified the rules of priorities to take into account the unequal number unit/attack_alignment of different or identical alignment that there can be for a unit.

newfrenchy83:
@stevecotton , @CelticMinstrel I don't want to rush you but this PR has been open for 4 months and I still don't know what your intentions are. Could you take a look at it again, please? Same also for https://github.com/wesnoth/wesnoth/pull/7003. Thanks.

CelticMinstrel:
I would say I am fairly ambivalent about this feature. I don't mind it being in the game, but I'm not eager to see it added either.

Code-wise, it looks okay but probably has room for improvement.

newfrenchy83:
> I would say I am fairly ambivalent about this feature. I don't mind it being in the game, but I'm not eager to see it added either.
> 
> Code-wise, it looks okay but probably has room for improvement.

@CelticMinstrel what improvements do you suggest?

newfrenchy83:
@CelticMinstrel i added cumulative attribute to [attack_alignment] for what tod_bonus of weapon could be added to tod_bonus unit_alignment or unit type alignment, and not systematicelly replace them.

CelticMinstrel:
> what improvements do you suggest?

The implementation of `choose_alignment` is really messy. I think something more like this would be better:

```cpp
static void choose_alignment(unit_ability_list& abil_list, unit_alignments::type& alignment, bool& cumulative)
{
	std::map<std::string, int> values;
	std::map<std::string, bool> cumulatives;
	for(const auto& i : abil_list) {
		const auto& alignment = (*i.ability_cfg)["alignment"].str()
		values[alignment] += 1
		cumulatives[alignment] = true;
	}
	int is_liminal = values["liminal"], is_chaotic = values["chaotic"], is_lawful = values["lawful"], is_neutral = values["neutral"];
	std::string str_alignment;
	if(is_liminal > 0 && is_liminal > is_chaotic && is_liminal > is_lawful && is_liminal > is_neutral){
		str_alignment = "liminal";
	}
	else if((is_neutral > 0 && is_neutral > is_chaotic && is_neutral > is_lawful) || (is_lawful > 0 && is_chaotic >0 && is_lawful == is_chaotic)){
		str_alignment = "neutral";
	}
	else if( is_lawful > 0 && is_lawful > is_chaotic){
		str_alignment = "lawful";
	}
	else if( is_chaotic > 0 && is_chaotic > is_lawful){
		str_alignment = "chaotic";
	}
	auto new_align = unit_alignments::get_enum(str_alignment);
	if(new_align) {
		alignment = *new_align;
		cumulative = cumulatives[str_alignment];
	}
}
```

I haven't tested it, so there may be some errors in that code. Also, I think your cumulative implementation in the last commit is broken – it looks like `cumulative=yes` in a lawful alignment ability would take effect even if there was a chaotic alignment ability that supercedes it. That doesn't seem right to me? The above implementation theoretically fixes that but, as said, it's not tested. It could still be improved further, but I don't feel like it's worthwhile to do so unless we were to implement support for arbitrary custom alignments.

newfrenchy83:
When I read the code again, I understand that cumulative=yes only takes effect in an alignment if it is itlised, if a chaotic alignment is prioritized over lawful but has no cumulative=yes then then the cumulative variable will take the false value since no chaotic ability/weapon in the list has cumulative=yes.

As for the extension to other arbitrary alignments, I don't think it's possible unlike damage_type which only returns an attack type name and that's the target resistance in the wml code to this custom type or cannon which conditions the damage caused.

newfrenchy83:
it don't work with cumulatives[alignment] = true; because presence of cumulative=yes not checked

CelticMinstrel:
> As for the extension to other arbitrary alignments, I don't think it's possible …

Well, that's something to decide in the future when someone (possibly me) finds time to implement said custom alignments. Right now, it only has to work for the built-in alignments.

I can't think of any reason not to merge it at this point.

newfrenchy83:
> > As for the extension to other arbitrary alignments, I don't think it's possible …
> 
> Well, that's something to decide in the future when someone (possibly me) finds time to implement said custom alignments. Right now, it only has to work for the built-in alignments.
> 
> I can't think of any reason not to merge it at this point.

Ok

src/units/attack_type.hpp
```diff
@@ -84,6 +85,7 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 	std::vector<std::pair<t_string, t_string>> special_tooltips(boost::dynamic_bitset<>* active_list = nullptr) const;
 	std::string weapon_specials() const;
 	std::string weapon_specials_value(const std::set<std::string> checking_tags) const;
+	std::pair<unit_alignments::type, bool> specials_alignment() const;
```
stevecotton, line 88: /**
 * Returns the ToD to use when calculating damage, including effects from [attack_alignment] specials.
 * If no such specials are active, it returns the unit's alignment.
 * 
 * The boolean returned indicates whether the ToD bonus for the unit's alignment should be applied (by
 * the caller) in addition to the ToD from the attack; this is based on [attack_alignment]cumulative=.
 */

src/units/abilities.cpp
```diff
@@ -1126,6 +1126,76 @@ int attack_type::modified_damage() const
...
+	std::map<std::string, bool> cumulatives;
+	for(const auto& i : abil_list) {
+		const auto& alignment = (*i.ability_cfg)["alignment"].str();
+		values[alignment] += 1;
+		if((*i.ability_cfg)["cumulative"].to_bool()){
+			cumulatives[alignment] = true;
+		}
+	}
+	int is_liminal = values["liminal"], is_chaotic = values["chaotic"], is_lawful = values["lawful"], is_neutral = values["neutral"];
+	std::string str_alignment;
```
stevecotton, line 1153: It's better to use enum values, and we already have the `unit_alignments::type` enum for this:

```
	std::optional<unit_alignments::type> new_align;
	if(is_liminal > 0 && is_liminal >= is_chaotic && is_liminal >= is_lawful && is_liminal >= is_neutral){
		new_align = unit_alignments::type::liminal;
	}
	else if((is_neutral > 0 && is_neutral > is_chaotic && is_neutral > is_lawful) || (is_lawful > 0 && is_lawful == is_chaotic)){
		new_align = unit_alignments::type::neutral;
	}
	else if( is_lawful > 0 && is_lawful > is_chaotic){
		new_align = unit_alignments::type::lawful;
	}
	else if( is_chaotic > 0 && is_chaotic > is_lawful){
		new_align = unit_alignments::type::chaotic;
	}
	if(new_align) {
		alignment = *new_align;
		cumulative = cumulatives[unit_alignments::get_string(alignment)];
	}
```

newfrenchy83: except what if i could use unit::alignments::type in if loop before use if(new_align) i would done it in begun, it why a use sstring variable in first time before to define alignment variable with.

src/units/abilities.cpp
```diff
@@ -1126,6 +1126,76 @@ int attack_type::modified_damage() const
...
+	std::map<std::string, int> values;
+	std::map<std::string, bool> cumulatives;
+	for(const auto& i : abil_list) {
+		const auto& alignment = (*i.ability_cfg)["alignment"].str();
+		values[alignment] += 1;
+		if((*i.ability_cfg)["cumulative"].to_bool()){
+			cumulatives[alignment] = true;
+		}
+	}
+	int is_liminal = values["liminal"], is_chaotic = values["chaotic"], is_lawful = values["lawful"], is_neutral = values["neutral"];
```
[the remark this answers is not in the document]
CelticMinstrel: Personally, I don't like having them on individual lines, but I guess it's okay.

stevecotton:
The cumulative option makes the code more complex, and I'm wondering if we could get almost the same effect in a different way, using the existing C++ code.

If I understand correctly, a `lawful` unit using a `cumulative lawful` attack gets double the normal bonus/penalty for time of day, and for a `chaotic` unit using a `cumulative lawful` attack the effects would almost cancel each other out. Nearly the same effect ought to be possible with
```
[damage]
    id=something_unique
    multiply=1.25
    [filter_...]
        some time-of day check
    [/filter_...]
    description=_"Regardless of the alignment of the unit using this weapon, the weapon itself gets an extra +25% during daytime and -25% during nighttime."
[/damage]
[damage]
    # equivalent that does the -25% at night
[/damage]
```

newfrenchy83:
> The cumulative option makes the code more complex, and I'm wondering if we could get almost the same effect in a different way, using the existing C++ code.
> 
> If I understand correctly, a `lawful` unit using a `cumulative lawful` attack gets double the normal bonus/penalty for time of day, and for a `chaotic` unit using a `cumulative lawful` attack the effects would almost cancel each other out. Nearly the same effect ought to be possible with
> 
> ```
> [damage]
>     id=something_unique
>     multiply=1.25
>     [filter_...]
>         some time-of day check
>     [/filter_...]
>     description=_"Regardless of the alignment of the unit using this weapon, the weapon itself gets an extra +25% during daytime and -25% during nighttime."
> [/damage]
> [damage]
>     # equivalent that does the -25% at night
> [/damage]
> ```

i will remove cumulative option and work instead on customize alignment based on limonal, chaotic or lawful but with percentage or multiple of bonus_value

newfrenchy83:
@stevecotton this should be like that

[attack_alignment]
    id=extra_chaotic
    alignment =chaotic
    multiply=2 (double or normal chaotic)
    [filter_self/student]
        alignment=chaotic
    [/filter_...]
    description=_"this weapon get  -25% during day and 25% during the night, except if unit already chaotic, where value become -50/50%"
[/attack_alignment]
[attack_alignment]
id=chaotic_attack
    alignment=chaotic
[/attack_alignment]

Pentarctagon:
Is there a reason that `tod_bonus()`/`base_tod_bonus()` can't be used in the [damage]?

newfrenchy83:
> Is there a reason that `tod_bonus()`/`base_tod_bonus()` can't be used in the [damage]?

damage called before tod calculation, impossible to use tod_bonus in special int these condition, because [damage] should modify base_damage before leadership or tod_bonus.

newfrenchy83:
@stevecotton https://github.com/wesnoth/wesnoth/pull/6968/commits/2130f9b613b5511c560830209b05fec7a5ed8664 added to this PR

newfrenchy83:
@stevecotton @CelticMinstrel i added filtering for unit_alignment, because before [filter]alignment matche with unit wml alignment only.

newfrenchy83:
@CelticMinstrel @stevecotton i removed unit_alignment for risk of infinite loop. Can let core modifications or should iremove to this PR?

CelticMinstrel:
I don't really have an objection to the concept. However:

1. Why are you changing HttT? (That probably requires Nemaara's approval by the way.)
2. We shouldn't be adding weapon special macros in core that aren't used in core. Even if you're modifying HttT, the macros should be in HttT, and only the ones that are actually used.

newfrenchy83:
> I don't really have an objection to the concept. However:
> 
>     1. Why are you changing HttT and other campaigns? (That probably requires Nemaara's approval by the way.)
> 
>     2. We shouldn't be adding weapon special macros in core that aren't used in core. Even if you're modifying HttT, the macros should be in HttT, and only the ones that are actually used.

i removed that and create a PR after merging of code. 

Why changing Httt? When Konrad has Scpeter of fire,damge implemented must be for what value in chaotic tod equals to Li'sar wh is neutral, for what power of Konrad don't exceed Li'sar power when she has Scepter, i propose to resolve that in applying neutral alignment to Scepter attacks.


2. like i modify Httt and Nr with same neutral alignment, i prefer put in core for don't duplicate wmlm code.

CelticMinstrel:
Basically: yes, any changes to campaigns are better in a separate PR, as the people who will be expected to review them are different.

I think I'm fine with this as it stands, but it would be nice to get input from @gfgtdf.

gfgtdf:
Firstof,  i didn't read the whole conversation so far, so i probably missed soemthing.

I'm a bit unsure, on the one hand i see that for example when you want a weapon to behave "lawful" independent on the alignment of the unit Its nice to have this.

On the other hand i think [damage]multiply=(formula) , where formula can access the tod bonus is strictly more powerful and flexible.

Lastly i'm unsure about the name, iiuc the reason for [attack_alignment] over [alignment] to allow a seperate [unit_alignment] in [abilities], but then what would be exactly the  difference between  [attack_alignment]  and [unit_alignment] , like what would be usecases for one or the other ?

newfrenchy83:
> On the other hand i think [damage]multiply=(formula) , where formula can access the tod bonus is strictly more powerful and flexible.

I already sais that but like alignment tod_bonus add/sub a percentage to total damage(base +weapon:damage) and add alignment and modify damage canbe more complex and less readable for wml develloper. And attack_alignment can also change value of alignment.

newfrenchy83:
@stevecotton @CelticMinstrel i added https://github.com/wesnoth/wesnoth/pull/6968/commits/ed4e57fb4c5604d4d1aed46e34ee33a38a76b76e commit for filtered alignment attack when diffrent of unit_alignment

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_alignment.cfg
```diff
@@ -0,0 +1,246 @@
+#textdomain wesnoth-test
+
+{GENERIC_UNIT_TEST "special_alignment_test" (
```
stevecotton, line 3: Please update the tests to use the new macros that have been used in more recent PRs.

newfrenchy83:
I gave up on the idea of adding an attack alignment equivalent to accuracy for hit chance

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_alignment.cfg
newfrenchy83: done

knyghtmare:
Adding alignment attributes, this might potentially change the designing concepts of existing and coming neutral/liminal factions or the usual ones. The UMC implications of this is truly interesting and high potential.

PR #6973, opened 2022-08-21
Open the forum profile of the currently selected user.

Pentarctagon:
Rather than always opening your own.


Vultraz (changes requested):
Please use the method added in 72bfb7e391ed96856dd054806973b4cc5f521660. It gives access to the whole `user_info` object instead of just the name, meaning you don't need to manually search after.

Pentarctagon:
Updated.

Vultraz (approved):

PR #6975, opened 2022-08-21
1.17 raised terrain

doofus-01:
This is companion to #6927, it allows a cosmetic raised terrain.  It also allows larger raised/lowered regions to be marked as raised or lowered, so that the overlay slot (`^Fp` for example) isn't taken up by the gulch/bluff designation.  The marking is based on setting terrain-graphics flags to disable the little cliff transitions, since we can't rely on terrain codes to know what the map is supposed to look like. 

To use:

1. Draw the outline of the raised/lowered blob with the bluff/gulch terrain overlay.
2. Put the appropriate marker ("special" terrain group) on one of the bounded hexes.
3. The rest of the bounded hexes will get flood-filled to be on the same raised/lowered level.
4. Don't worry about the banded overlay, it is just a guide for the map-maker.  It should not show up in game.

There are still some issues: 

- [ ] Most major is that the terrain-graphics flag setting seems to only propagate up & left two hexes per terrain-graphics tag.  It's possible I'm doing something wrong, but it may be an engine limitation.  
- [ ] Another limitation is that we only have three elevation levels; I think that is fixable, but I've run out of time for this week.
- [ ] Related to the number of levels, is that high-low (vs high-normal) transitions don't have special tall cliff images.  I'm not sure how to deal with this, since the elevation change doesn't actually exist.  Special images for this will (I fear) cause the number of needed transitions to explode (same reason I've not touched the "cliffs" terrain idea that's been floating around for years).
- [x] A little more fun, and something I can actually deal with in this PR, is that we need steps or some sort of man-made transition, in addition to the little cliffs.
Screenshot_20220821_a
EDIT: Got the stairs done.  I don't know how to deal with the first and third issues, but I do plan to work on the second one.


doofus-01:
Screenshot_20220903_a
A draft of the built transition.  We still need one for wooden terrain.  Also, I'm realizing this does not play well with the bridges.

I thought I was clever and was avoiding the cliffs transition explosion, but I'm less sure that now...

doofus-01:
There are now wooden and stone stairs/clifflets.  This does need more work, but I think it's a usable improvement over the current state of affairs.  When this passes CI, I will merge, because I want to get past this and on to 

- a scheme where there can be gulch-within-gulch or bluff-upon-bluff  
- a point where there is a user test-case, as this is elevation stuff is not standard
Screenshot_20220905_a

Issue #6641, opened 2022-04-27
[Feature Request] [Multiplayer] Add a timestamp into server chat messages


Discontinuum:
### Describe the desired feature

At the moment server messages have no timestamps, so when you view a replay, chat messages have timestamps of when they were sent, and server messages, like those when player disconnects, have no real timestamp, in chat they are shown with real time on the moment of viewing. If it's possible, I suggest to add timestamps to the multiplayer server's messages

PR #6979, opened 2022-08-25
Timestamping server messages

Discontinuum:
Implements #6641

Just add "time" in the "speak" command that the server sends, I guess it should be fine?

I push to master, but I wonder, can it be backported into 1.16? Would be nice if possible

### DIFF
--- a/src/server/wesnothd/game.cpp
+++ b/src/server/wesnothd/game.cpp
@@ -1955,6 +1955,9 @@ void game::send_server_message(const char* message, std::optional<player_iterato
 
 		msg.set_attr("id", "server");
 		msg.set_attr_dup("message", message);
+		std::stringstream ss;
+		ss << ::std::time(nullptr);
+		msg.set_attr_dup("time", ss.str().c_str());
 	} else {
 		simple_wml::node& msg = doc.root().add_child("message");
 


Pentarctagon:
I'm assuming this doesn't show up on UI anywhere?

Discontinuum:
Well, the purpose is to have timestamps in replays so when you view a replay, chat messages from the server had correct timestamps. Right now server messages have no timestamps so they are shown with current time when you watch a replay

Pentarctagon:
Ah, alright. So it's populating an existing attribute - otherwise I wasn't sure where this would show up seeing as the only change is on the server side.

src/server/wesnothd/game.cpp
```diff
@@ -1955,6 +1955,12 @@ void game::send_server_message(const char* message, std::optional<player_iterato
 
 		msg.set_attr("id", "server");
 		msg.set_attr_dup("message", message);
+		const std::time_t time = ::std::time(nullptr);
+		std::stringstream ss;
+		ss << time;
+		const std::string timestr = ss.str();
+		const char* timecstr = timestr.c_str();
+		msg.set_attr_dup("time", timecstr);
```
soliton-, line 1963: I think it's ok if we write this a little shorter:
```suggestion
		std::stringstream ss;
		ss << ::std::time(nullptr);
		msg.set_attr_dup("time", ss.str().c_str());
```

soliton-:
Yes, this can also be backported to 1.16.

src/server/wesnothd/game.cpp
Discontinuum: I'm unsure where the string object returned by ss.str() in the middle of the expression will be destroyed. If it's destroyed before set_attr_dup copying the char array, the array will be already freed memory. But if my worry is unfounded, I'll shorten it like you suggest ofc

soliton-: If it would get destroyed in the middle of the expression that'd be pretty crazy. You can find an example of similar code in this same source file, btw.

https://en.cppreference.com/w/cpp/language/lifetime#Temporary_object_lifetime

Discontinuum: Alright, I've applied the change

soliton-:
Thanks!

PR #6980, opened 2022-08-25
Log exception type when doing catch(...).

Pentarctagon:


CelticMinstrel:
…what on earth is it doing in the Xcode project file…? It looks like it just moved files around without actually doing anything? Or am I missing something here?

Pentarctagon:
It also added utils/general.cpp.

PR #6984, opened 2022-08-26
[Units] Fix XP conventions oversight in some monster/horse Units

knyghtmare:
## Summary:

Certain units with no level up/advancements were not adhering to the max XP conventions which had been set for core units. This rectifies such oversights.

## Reference Table:

| Level | Maximum XP |
| ----- |  -------------- |
| 0 | 25 |
| 1 | 50 |
| 2 | 100 |
| 3 | 150 |
| 4 | 200 |
| 5 | 250 |
| 6 | 300 |

Note: Only Applicable for Units with no further advancements being available apart from `{AMLA_DEFAULT}`

Note 2: Some commits here should/can be cherry-picked since these units exist in 1.16 as well (all except the Hunter Caribe)

### DIFF
--- a/data/core/units/monsters/Caribe_Hunter.cfg
+++ b/data/core/units/monsters/Caribe_Hunter.cfg
@@ -18,7 +18,7 @@ units/monsters/caribe#enddef
         impact=80
     [/resistance]
     movement=8
-    experience=40
+    experience=100
     level=2
     alignment=neutral
     advances_to=null
--- a/data/core/units/monsters/Fire_Wraith.cfg
+++ b/data/core/units/monsters/Fire_Wraith.cfg
@@ -36,7 +36,7 @@ units/monsters/firewraith#enddef
         impact=90
     [/resistance]
     movement=6
-    experience=80
+    experience=100
     level=2
     alignment=neutral
     advances_to=null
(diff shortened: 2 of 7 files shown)


knyghtmare:
All checks have passed. Merging in 2 days if no arguments/comments

doofus-01:
If these are monster units, they should have XP that makes sense for "balance" or "flavor" (maybe a monster is supposed to be different than infantry).  Thumbs down on the reference table.

knyghtmare:
The XP conventions have been in effect for almost all mainline units since the 1.10 days. So, are you suggesting that this convention not be applicable to monster units? If so, then almost all monster units have to change. Not just these new ones.

I think referring the SP-content dev and MP-content dev here for their opinions is a requirement.

nemaara (approved):

doofus-01:
Sorry for the slow response.  No, I was asking that some thought be put into this on an individual unit basis.  A guideline is cool, but deviations aren't necessarily a bug. any more than any other stat.  If you've already got that covered, then no problem.  I had to object because you were about to merge, and it wasn't clear to me what was changing (without actually reviewing the files, which I didn't have time for).

knyghtmare:
Since you're asking the only thing that changed was the value in `experience=` attribute of the unit files. Nothing else has been altered.

knyghtmare:
> Okay to merge? Or, is there going to be a debate?

Two weeks...? Merging in 24 hours if no response.

PR #6985, opened 2022-08-26
CMake 2.24.0: Workaround for missing SDL2::SDL2main

stevecotton:
@bencsikandrei please would you review this? I can only add project members as reviewers, but you're probably the best reviewer.

In SDL 2.24.0, the CONFIG method doesn't add a target for libSDL2main. Seen on Debian Unstable, and seems to be https://github.com/libsdl-org/SDL/issues/6119 which is reported upstream on macOS. For consistency's sake, let's add a phony library we can link with, which is also the workaround that upstream suggests in the bug report.

### DIFF
--- a/cmake/FindSDL2.cmake
+++ b/cmake/FindSDL2.cmake
@@ -58,6 +58,15 @@ if(SDL2_FOUND)
 
 	mark_as_advanced(SDL2_INCLUDE_DIRS SDL2_LIBRARIES)
 
+	if(NOT TARGET SDL2::SDL2main)
+		# In SDL 2.24.0, the CONFIG method doesn't add a target for libSDL2main. Seen on Debian
+		# Unstable, and seems to be https://github.com/libsdl-org/SDL/issues/6119 which is reported
+		# upstream on macOS. For consistency's sake, let's add a phony library we can link with,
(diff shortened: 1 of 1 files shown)


bencsikandrei:
hello!

thanks for adding me to this!

dropped by just to let you know I'll definitely take a look ASAP! caught up right now, but will be back in a couple hours.

have time for a first glance

judging from your description, the adding the phony target idea is sound (it's probably also what I would have done) but I need to make sure it doesn't have some weird implications (it shouldn't tho)

I also like their other idea, since we're consuming `SDL` and don't need some backward compat thing for the build system - as `SDL` itself does:
```
target_link_libraries(my_executable PRIVATE $<$<TARGET_EXISTS:SDL2::SDL2main>:SDL2::SDL2main> SDL2::SDL2)
```
what do you think?

bencsikandrei:
alright, I stand by what I said earlier, looks good to me.
if it somehow poses some issues in the future, maybe the other idea, of conditionally linking with SDL2Main, can take its place, but for now I like this approach as it is simpler.

Pentarctagon (approved):

stevecotton:
Thanks for the feedback.

PR #6990, opened 2022-08-30
Fix inserting image=wesnoth-icon.png errors

Dou1622:
Fixes all wmllint "inserting 'image=wesnoth-icon.png'" errors.
Wmllint will insert a image for the narrator when there isn't one, and print a warning (see comment starting on line 2765).
I added an image for the narrator when there wasn't one.
See #4494

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
@@ -337,6 +337,7 @@
         [message]
             speaker=narrator
             message= _ "Ignorance is the parent of fear."
+            image=wesnoth-icon.png
         [/message]
 
         {GIVE_MALIN_EXPERIENCE 4}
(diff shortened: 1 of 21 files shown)


Pentarctagon:
Based on a previous time this had come up, I think we'd rather just have wmllint not insert that image/warn about it being missing. So instead the lines https://github.com/wesnoth/wesnoth/blob/9d5266d624246a0d4e81537f461ab98de0a94d44/data/tools/wmllint#L2765-L2784 should be deleted.

cooljeanius:
Isn't there supposed to be a magic comment that `wmllint` recognizes to suppress these? `# wmllint: no-icon` or something?

Pentarctagon:
The larger point though is that this simply isn't the expectation for what the default should be anymore. It doesn't make sense to have to add magic comments to suppress something that shouldn't be the default anyway.

stevecotton:
> It doesn't make sense to have to add magic comments to suppress something that shouldn't be the default anyway.

Opened PR #6991 on wmllint.

Pentarctagon:
This can be closed due to #6991 I assume?

stevecotton:
@Dou1622 thanks for your effort, and apologies that we hadn't fixed wmllint before now.

Dou1622:
> @Dou1622 thanks for your effort, and apologies that we hadn't fixed wmllint before now.

That is ok.

PR #6991, opened 2022-08-31
wmllint: stop adding wesnoth-icon.png to messages

stevecotton:
We're had at least two people look at the "good first issue" list, pick up
the wmllint warnings, open a PR, and then get a response that these .pngs
shouldn't be added by default. That's not a welcoming introduction.

IIRC we didn't get full consensus that these should never be added, but I
think it's time to open a "just drop that code, no options" PR. If someone
wants it back in a subset of circumstances, they can always create a PR to
readd it along with adding an option to enable it.

Inspired by #6990, previously discussed in #5164.

### DIFF
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -2762,26 +2762,6 @@ def hack_syntax(filename, lines):
                             lines.insert(j+1, new_line)
                             j += 1
                     j += 1
-    # Ensure that every speaker=narrator block without an image uses
-    # wesnoth-icon.png as an image.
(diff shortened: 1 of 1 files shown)


Pentarctagon (approved):

stevecotton:
I'm assuming that we don't backport this to 1.16 - on the one hand the default behavior was a bug, but on the other hand it would be an unnecessary change to the tool's behavior in a stable branch.

Pentarctagon:
Yeah, no need to backport it.

PR #6994, opened 2022-09-01
Give Shyde and Sylph unwalkable statistics

Jonathan-Kelly:
With the same unwalkable terrain movement and defense as the Quenoth versions of these units. Makes them behave as true flying/floating units as their wings, movetype name and other statistics suggest, while opening up interesting game play possibilities.

### DIFF
--- a/data/campaigns/Heir_To_The_Throne/utils/httt_utils.cfg
+++ b/data/campaigns/Heir_To_The_Throne/utils/httt_utils.cfg
@@ -642,7 +642,7 @@ fire:  +10%"
     # everyone on lava dies
     [kill]
         [not]
-            type=Gryphon Rider,Gryphon Master
+            type=Elvish Shyde,Elvish Sylph,Gryphon Rider,Gryphon Master
         [/not]
         [filter_location]
             terrain=Ql
--- a/data/campaigns/The_South_Guard/maps/06a_Into_the_Depths.map
+++ b/data/campaigns/The_South_Guard/maps/06a_Into_the_Depths.map
@@ -14,12 +14,12 @@ Xu, Xu, Xu, Xu, Xu, Uu^Vud, Uu^Emf, Qxu, Ur^Es, Qxu, Xuc, Xuc, Xuc, Qxu, Xu, Xu,
 Xu, Xu, Xu, Xu, Xuc, Xom, Uu^Em, Uu^Br\, Ur^Br\, Ur^Ebn, Xuc, Xuc, Qxu, Qxu, Xu, Xu, Uu^Vu, Uu^Em, Rd^Es, Rd^Es, Wwg, Ss, Sm, Sm^Em, Sm^Emf, Sm^Em, Uue^Em, Rd^Es, Uue, Ss^Em, Sm^Tf, Ss^Emf, Xue, Xue, Xu, Xu
 Xu, Xu, Xuc, Xuc, Xuc, Xom, Xom, Ur^Es, Ur^Br/, Ur^Br/, Uu^Em, Xuc, Qxu, Qxu, Xu, Xu, Uu^Em, Uu^Es, Xu, Ds^Esd, Sm, Ww^Ewl, Ss^Em, Ss^Vhs, Ss^Tf, Uue^Em, Uue^Es, Rd^Es, Xu, Sm^Emf, Rb^Em, Ss^Em, Xue, Xu, Xu, Xu
 Xu, Xuc, Xuc, Uu^Vu, Xom, Xom, Uu^Em, Ur^Br\, Ur^Es, Ur^Br|, Xu, Uh^Vud, Uh^Emf, Qxu, Qxu, Qxu, Uu, Uu^Es, Xu, Xu, Xu, Sm^Em, Sm^Em, Wwg, Ss^Em, Xu, Uue^Es, Ur^Es, Uue^Em, Xu, Sm^Em, Xue, Xue, Xu, Xu, Xu
-Xu, Xom, Uh^Emf, Uu^Em, Uu^Es, Uu, Ur^Es, Ur^Es, Cud, Ur^Br|, Xu, Uh^Em, Cud, Qxu, Qxu^Bs/, Qxu^Bs/, Qxu, Qxu, Qxu, Xu, Qxu, Sm^Emf, Ss^Em, Ww^Ewl, Ww^Ewl, Sm^Em, Uue^Em, Ur, Uue^Em, Ss^Em, Sm^Em, Sm^Emf, Ss^Edb, Xue, Xu, Xu
-Xu, Xu, Uh^Tf, Uh, Uu^Ebn, Cud, Xuc, Cud, 5 Kud, Cud, Ur^Es, Ur, Uu, Ur^Es, Qxu, Qxu, Qxu, Qxu, Qxu, Qxu, Qxu, Ss^Em, Ww^Ewl, Wwg, Wwg, Wwg, Rd^Es, Rd^Es, Uh^Emf, Rb^Em, Xue, Xue, Ss^Edb, Xue, Xu, Xu
-Xu, Xu, Xu, Uh^Em, Uu^Em, Xuc, Xuc, Xuc, Xuc, Xuc, Xuc, Uh^Em, Xu, Xu, Xu, Qxu, Qxu, Qxu, Qxu, Qxu, Qxu, Qxu, Xu, Wo, Wo, Wwg, Xu, Rd^Es, Rd^Es, Uh^Tf, Xue, Ss^Em, Uue^Emf, Xue, Xu, Xu
-Xu, Xu, Xu, Uh^Em, Uh^Emf, Uu^Vu, Xuc, Xu, Xu, Xuc, Xu, Xom, Xu, Xu, Qxu, Qxu, Qxu, Qxu, Qxu, Qxu, Qxu, Xu, Xu, Wo, Wog^Edb, Wwg^Vm, Ww^Ewl, Xu, Uu^Emf, Ur, Uue, Uue^Vud, Uue^Em, Uu^Em, Xu, Xu
-Xu, Xu, Xu, Xu, Xu, Uh^Tf, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Qxu, Qxu, Qxu, Qxu, Qxu, Qxu, Xu, Xu, Wo, Wwg, Wwg, Ww^Ewl, Xu, Uu^Tf, Ur, Uue^Emf, Uue^Tf, Uu^Edb, Uu^Edb, Xu, Xu
-Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Xu, Qxu, Qxu, Qxu, Qxu, Qxu, Wwg, Wog^Edb, Wog^Edb, Wog^Edb, Wwg^Edb, Wwg, Wwg^Edb, Cud, Cud, Ur^Edb, Ur, Cud, Cud, Xu, Xu, Xu
+Xu, Xom, Uh^Emf, Uu^Em, Uu^Es, Uu, Ur^Es, Ur^Es, Cud, Ur^Br|, Xu, Uh^Em, Cud, Qxu, Qxu^Bs/, Qxu^Bs/, Qxu, Qxu, Qxu, Xu, Xu, Sm^Emf, Ss^Em, Ww^Ewl, Ww^Ewl, Sm^Em, Uue^Em, Ur, Uue^Em, Ss^Em, Sm^Em, Sm^Emf, Ss^Edb, Xue, Xu, Xu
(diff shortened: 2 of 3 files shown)


nemaara (approved):

irydacea:
After years of designing maps under the assumption that Shydes will never cross unwalkable terrain I'm not sure how I feel about this.

CelticMinstrel:
Personally, I'd be more inclined to change the Quenoth units to conform _not_ cross unwalkable, rather than this.

knyghtmare:
> Personally, I'd be more inclined to change the Quenoth units to _not_ cross unwalkable, rather than this.

They have winged units. So, instead adapt scenarios to not make players ask "Why cannot a Shyde/Sylph fly over deep water but can fly over chasm?"

CelticMinstrel:
> "Why cannot a Shyde/Sylph fly over deep water but can fly over chasm?"

I think there's a pretty obvious candidate reason for being able to cross water but not chasms – their flight uses wings for steering but air magic for lift and locomotion. They can cross water because it has a certain surface tension, but with chasm their air cushion wouldn't provide any lift because it just disappears into the depths.

That doesn't explain lava (which is also unwalkable) though… also, can they cross deep water? I know they can cross shallow water, but I'm not sure if that explanation works for being able to cross shallow water but not deep water.

irydacea:
In the past it's been explained as them hovering not too far above ground
using magic rather than being able to fly into the skies like drakes or
bats.

Jonathan-Kelly:
The concept of "ground effect" faeries feels quite strange and counterintuitive. It also subtracts from the mystique of their having achieved true flight - an ability beyond the reach of even the wisest humans.

Since half the MP factions have flyers, the maps are already balanced to accommodate them. So the cases where some map tweaks might be needed should be quite rare.

Whereas a number of map changes would be needed for UtBS if we went the other way and removed flight. I remember a number of chasm crossings that were already quite painstaking _with_ proper Shydes and Sylphs.

Discontinuum:
> Since half the MP factions have flyers, the maps are already balanced to accommodate them. So the cases where some map tweaks might be needed should be quite rare.

It's more about UMC SP content, than MP faction. In MP high levels are relatively rare (if we talk about classical competitive). On the other hand, in UMC you are likely to have Shydes/Sylphs, and maps might be adapted to it. Overall, these kind of changes in very old units always pose a decent risk of turmoil for UMC

irydacea:
> The concept of "ground effect" faeries feels quite strange and counterintuitive. It also subtracts from the mystique of their having achieved true flight - an ability beyond the reach of even the wisest humans.

That's entirely a matter of opinion and I don't really see it that way. It's perfectly fair for magic of any kind to be bounded and limited, especially if it helps creators come up with more interesting challenges. 😉

> Whereas a number of map changes would be needed for UtBS if we went the other way and removed flight.

This is a good unstated point, because I remember being involved in discussions about the rework in the past where I suggested that Quenoth magic is mostly completely different to the original elves' because of the untold amount of time they spent unable to use any forests for channeling. So it's perfectly reasonable — expected, even — for them to have abilities that their forest ancestors did not possess, and viceversa.

Also, perhaps I forgot something – in which case I welcome your correction — but I don't think the true nature of the forest elves' flight ability is actually stated in mainline content anyway, since a lot of this stuff so far has been left to people's imagination until the day it becomes plot-relevant.

Hejnewar (approved):
It shouldn't affect MP balance at all.

nemaara:
This may need more discussion but I'd still prefer to have these units have unwalkable stats. Recognizing that they've been unable to cross chasms for a long time means that changing this causes trouble for UMC but I'm not sure whether this precedent is a good reason to not change it ever. So I would still like to merge this if there are not too many who disagree.

CelticMinstrel:
I'm only mildly opposed to this change, but… I don't think every flying unit needs to have exactly the same movement stats – some creatures are better at flying than others.

nemaara:
> I'm only mildly opposed to this change, but… I don't think every flying unit needs to have exactly the same movement stats – some creatures are better at flying than others.

I suppose mildly opposed isn't that bad. You bring up a good point that we don't need to have them have great stats over unwalkable, but it has bothered me from the very beginning that they weren't able to fly over it at all.

knyghtmare:
They have wings and should fly. Better question: why do Quenoth flyers still no fly over deep water ?
I believe if that if this is finally being allowed...Quenoth flying units should not be restricted from Deep water. 

The current statement:
"So, Shydes and Sylphs can fly over chasms now, but the Quenoth ones already did that but still cannot fly over deep water? Absurd"

Thus, Quenoth units should be allowed deep water mobility as well.

One further comment is that the established campaign IftU might have an exploit now due to this change. Player gets Shydes so they can now bulldoze scenarios with Shyde spam

max-torch:
> They have wings and should fly. Better question: why do Quenoth flyers still no fly over deep water ? I believe if that if this is finally being allowed...Quenoth flying units should not be restricted from Deep water.
> 
> The current statement: "So, Shydes and Sylphs can fly over chasms now, but the Quenoth ones already did that but still cannot fly over deep water? Absurd"
> 
> Thus, Quenoth units should be allowed deep water mobility as well.
> 
> One further comment is that the established campaign IftU might have an exploit now due to this change. Player gets Shydes so they can now bulldoze scenarios with Shyde spam

Balance of Blood is Thicker than Water and Zochtanol Isle in UtBS will be affected.

knyghtmare:
If you're concerned with balance so much, then I would point out that the _The Lost General_, _Sceptre of Fire_ and _Hasty Alliances_  scenarios of _Heir of the Throne_ will be affected severely as well. Sylphs and Shydes can navigate over chasm now and these scenarios rely on disallowing unit combat positioning via chasm usage.
Now, they are redundantly jokes...I mean the Sylphs Blitz squad can clear these scenarios with little to no effort. Another scenario is the one in TSG where they fight the lich and that is affected as well. Since I am on the topic, the scenario _Overlook_ of _Winds of Fate_ is affected as well. Thus, the list of changes that need to be followed up are many...

Note: _two scenarios of the upcoming NR-rework are listed as affected, But I sort of fixed it beforehand._

This change is good, but I think before merging, all possible after effects be resolved soon after or with this PR.

> IMO, it is worth finding a justification for the difference in movement stats in order to preserve the unique strategic/tactical flavor that makes UtBS unique and feel like a different, fresh experience.

I suggest altering their graphics to make their wings look underdeveloped and then restricting movement in both chasm and deep water.  It's believable in that sense since these malnourished wings would not be able to carry them over such terrain. Because other than that, your arguments are not that strong to me...

knyghtmare (approved):
Looks good to go. Just slightly rebalance/rework the following afterwards:
1. _The Lost General_ (HTTT)
2. _Hasty Alliances_ (HTTT)
3. _Sceptre of Fire_ (HTTT)
4. _Overlook_ (WoF)
5. _Into the Depths_ (TSG)

Will elaborate reasonings on request.

nemaara:
> Looks good to go. Just slightly rebalance/rework the following afterwards:
> 
> 1. _The Lost General_ (HTTT)
> 2. _Hasty Alliances_ (HTTT)
> 3. _Sceptre of Fire_ (HTTT)
> 4. _Overlook_ (WoF)
> 5. _Into the Depths_ (TSG)
> 
> Will elaborate reasonings on request.

Ah, you're right. If we merge these let's tackle the balance changes as well in the same PR.

nemaara (changes requested):
Need to redo mainline campaign maps where this would affect gameplay.

Jonathan-Kelly:
I reviewed the affected scenarios and have a few questions and proposals regarding what adjustments I should make:

> 1. _The Lost General_ (HTTT)

There is only one chasm tile at (15,13) which is fully unreachable by the antagonists since they lack flyers in this scenario. The player is expected to have Gryphon Riders and Gryphon Masters at this point, which versus Sylphs/Shydes have better melee and defense, but no ranged and take slightly longer to reach this spot.

The effect of this change here is negligible, while giving the player some minor tactical variety (choosing which type of flying unit to place there). So I recommend no change to this map. Otherwise, I can put a cave chasm bridge across it, which fits the lore of this place being "_once the center of a bustling dwarvish empire._"

> 2. _Hasty Alliances_ (HTTT)

This scenario has two phases. The first is to survive an initial assault by the Level 2 Trolls and Goblin Knights that have surrounded the protagonists and antagonists (forcing them to join forces). This first phase is unaffected by the changes.

The second phase is a generic kill the leader objective, which before and after this change is best done using dwarves for their high defense and mobility on cave terrain and high resistances. The only effect this change would have is allowing a few Shydes/Sylphs to push from (10,9) , (11,9) and (11,10) to assault the Troll Warrior leader in his castle, near the end of the scenario when its castle is empty.

But this would be a mediocre strategy for a number of reasons:
* Enemy Blood Bats can reach this area.
* Shydes/Sylphs have only 30% defense over unwalkable (and cave).
* Shydes/Sylphs have poor mobility over cave and unwalkable, so this ploy is slow to set up.
* High risk of next turn melee retaliation from chaotic units under cave lighting.
* High experience value of Shydes and extreme experience value of Sylphs is not worth this risk.
* Shydes/Sylphs are AMLA so the experience of killing a Trolll Warrior is mostly wasted (better given to a new dwarf).
* Dwarf recruits are generally much more efficient in this environment.

I recommend no change for this map. Otherwise, I can fill this one big chasm/lava area with water.

> 3. _Sceptre of Fire_ (HTTT)

So the only change I should make here is to add the Shyde and Sylph to the lava mechanic's do-not-kill list?

https://github.com/wesnoth/wesnoth/blob/08e79f4e36211fd0c48b619e769d736e8b650604/data/campaigns/Heir_To_The_Throne/utils/httt_utils.cfg#L645

Or was there something more? The map is procedurally generated, with no unwalkable tiles beyond those generated by the volcanic eruption (which takes place while the scenario progresses).

> 4. _Overlook_ (WoF)

No change needed here - more hostile flyers only improves balance, since at least half the protagonists can fly, and there is no turn limit, so all the pressure must come from the antagonists,

> 5. _Into the Depths_ (TSG)

There is a wide chasm area surrounding (18,18) which currently allows Blood Bats to pass through uncontested. Flying Shydes can do okay against them by combining they healing and slows to compensate for their poor defense over unwalkable and the bats' strong drains attacks in the dark. But deploying Shydes here deprives the player's main army of healers, and Shydes/Sylphs would have a very difficult time assaulting the undead stronghold in the southeast through this corridor, as they would have no support from more durable, expendable and/or melee focused units of the player's army.

I recommend no map changes. Otherwise, I can close off this circular chasm area from the surrounding lake/swamp by placing a few more cave wall tiles (which might help visually explain why the water does not all drain into the chasm). Alternatively, I could replace all the chasm areas with water and swamp, expanding on the theme of the map being a subterranean wetland,

So what do we think?

Also, should I post screenshots of these possible map changes here or is that costly in some way?

CelticMinstrel:
> Also, should I post screenshots of these possible map changes here or is that costly in some way?

I'm not sure if it's costly, but it _would_ help to understand what you're describing.

Jonathan-Kelly:
Alright then, I updated my previous comment with pictures of the possible changes.

knyghtmare:
> The player is expected to have Gryphon Riders and Gryphon Masters at this point, which versus Sylphs/Shydes have better melee and defense, but no ranged and take slightly longer to reach this spot.

It's not mandatory to have those units in the recruit-list as they are optionals. You underestimate how fast Shydes and Sylphs can reach there. Around the same time as your first castle of dwarves. Gryphons are heavily restricted by cave hexes and take longer.

I think the map should stay unchanged since it has a dialogue relating to that 7-hex chasm.

Jonathan-Kelly:
> Gryphons are heavily restricted by cave hexes and take longer.

True, the pathfinder says it takes them one turn longer to reach that spot, unless they both have the quick trait, in which case they arrive at the same time.

> I think the map should stay unchanged since it has a dialogue relating to that 7-hex chasm.

Alright, do you want any changes to the remaining three maps - _Hasty Alliance_, _Scepter of Fire_, or _Into The Depths_?

For _Into The Depths_ it looks like there is an event handler that gives Mebrin extra gold when he is sighted, for the final stage of the fight. Since a Shyde or Sylph could catch a somewhat early glimpse of him from the chasm, I should probably go with walling this area off with cave tiles as depicted above (in the second to last image).

For _Scepter of Fire_ I will add the Shyde and Sylph to the exempt list for the lava kill mechanic, as mentioned earlier.

Jonathan-Kelly:
@nemaara
> Need to redo mainline campaign maps where this would affect gameplay.

Alright, the maps and scenarios have been updated to only be positively affected by this.

nemaara (approved):
Thanks, will still leave this open for let's say 2-4 more weeks so people will have time to comment since this change can affect a lot of things.

irydacea:
I continue not to know how I feel about this change.

CelticMinstrel:
I continue to be mildly opposed on the grounds that not all flying units need to have the exact same stats.

nemaara:
Planning to merge on Saturday if there's no additional feedback.

Issue #6997, opened 2022-09-02
[1.17] Follow unit actions responds to multiturn moves by units currently hidden by fog


stevecotton:
### Game and System Information

- **Version:** 1.17.7+dev built from e2b1295ac6c520e1ad68e80a90390870a10022ec
- **Downloaded from:** built from Git
- **Build info:** GCC 12, e2b1295ac6c520e1ad68e80a90390870a10022ec
- **OS:** Linux (Debian Unstable)

Also reproducible with:
- e2b1295ac6c520e1ad68e80a90390870a10022ec + revert of 978229c24732f45975d80e158b7dbc0233b80e69
- e2b1295ac6c520e1ad68e80a90390870a10022ec + fix to #6716 that I'm about to open a PR for

### Description of the bug

Setting up a multi-turn move where the final hex is visible to the opponent triggers "follow unit actions" to scroll to the final hex of the multi-turn move, even if all movement that happens during the current turn is hidden by fog of war.

An explanation of what multi-turn moves are is in #6292. Players are IMO unlikely to trigger this bug, because it means setting up a multi-turn move to a place where the opponent's forces may move to before the multi-turn move finishes; however it's a bug and it shows up when testing the fix to #6716.

### Steps to reproduce the behavior

1. Ensure that the advanced preference "follow unit actions" is enabled
2. Run two instances of Wesnoth, arrange them on screen so that you can see both
3. Create a network multiplayer game between them, with fog enabled, on a map where the keeps are at least 2 turns' moves away from each other. Caves of the Basilisk or Isar's Cross (with 2 sides per player) both show the issue.
4. As player 1, make a multi-turn move that will end near the player 2's keep, but don't end turn
5. Observe that player 2's screen scrolls to where player 1's multiturn move will end

### Expected behavior

Don't leak this information.

PR #6998, opened 2022-09-02
Show units' max move ranges, unless it's the owner's turn

### LINKED ISSUE #6716: Movement range of units is not properly shown after the turn of the side owner ends

- **Version: 1.16.2**
- **Downloaded from: Steam**
- **OS: Windows**

### Description of the bug

After the side owner ends the turn, instead of showing the movement range that unit would have next turn, what is shown is still the current movement range. In additional content, I showcase the issue, I mouse over the unit, then press 2, then release and then I can see movement range of the unit. In addition to that 1 no longer works probably because of this issue. 

1. Create 4 player game in local multiplayer.
2. Use all of the movement on one of the leaders.
3. Check thier movement range from alied side perspective.
Or
1. Create 4 player game on multiplayer server.
2. Play and end the turn.
3. Check movement ranges of units from perspective of side owner.
Or
1. Observe any 4 player game and check units of a side that just ended the turn. 

### Expected behavior

After the owner ends the turn, all movement ranges of units are instantly refreshed and show where units can move next turn. 

### Additional context

https://youtu.be/MO54GIzembY 

Usually I would upload directly but for some reason it didn't let me do that.

stevecotton:
If it's another player's turn, then movement points will be reset in a refresh
phase and it's not useful to show current movement.

Partially reverts 978229c247 (af0968d9c3 on 1.16), using the old logic for
the movement range but the allies-based logic for goto_x,goto_y information.

Fixes #6716.

Note: bug #6997 is reproducible in all three of these combinations, so I'm treating it as an orthogonal issue:
* without this PR
* with this PR
* with a full revert of 978229c247

### DIFF
--- a/src/mouse_events.cpp
+++ b/src/mouse_events.cpp
@@ -361,19 +361,23 @@ void mouse_handler::show_reach_for_unit(const unit_ptr& un)
 				gui().set_route(&route);
 			}
 			over_route_ = true;
+		}
 
-			wb::future_map_if_active raii;
-			current_paths_ = pathfind::paths(*un, false, true,
-											 viewing_team(), path_turns_);
-		} else {
-			//unit under cursor is not on our team
-			//Note: planned unit map must be activated after this is done,
-			//since the future state includes changes to units' movement.
-			unit_movement_resetter move_reset(*un);
+		// Scope for the unit_movement_resetter and future_map_if_active.
+		{
+			// Making this non-null will show the unit's max moves instead of current moves.
+			// Because movement is reset to max in the side's refresh phase, use the max if
+			// that refresh will happen before the unit's side can move again.
+			std::unique_ptr<unit_movement_resetter> move_reset;
+			if(un->side() != side_num_) {
+				move_reset = std::make_unique<unit_movement_resetter>(*un);
+			}
 
+			// Handle whiteboard. Any move_reset must be done before this, since the future
+			// state includes changes to units' movement.
 			wb::future_map_if_active raii;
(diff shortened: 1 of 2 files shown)


Wedge009 (approved):
I'm not sure what I can offer here specifically, since I seldom play/test MP, unless you're asking about the Whiteboard which I also haven't done much with other than dabbling with it when I first got involved years ago. If I understand correctly, the core of the change is only resetting the unit movement when the unit's side is not the owner's - which matches your description in the PR.

It seems to make sense to me, so approving.

Discontinuum:
Maybe this wasn't mentioned in the original bug, or I misread the descriptions, but there is a second thing from which MP players suffer the most: suppose it's a turn of side N, your enemy. N moves units, and while he's doing that, the game during N's turn shows for others their full range instead of partial ranges. Is this fixed here?

Discontinuum:
Anyway, I think I should better build myself and check how it works from MP perspective

Discontinuum:
Good, I checked it, it seems it now works as it did and as it was comfortable. Although fonts are quite dizzying (in a bad way) tbh, is it how 1.17 is supposed to be rendered now (offtopic, I know, maybe that's a scope for another bug)?

изображение

изображение

изображение

Wedge009:
It looks like there's a black outline around the lettering in current master (135e16e8536), it's not present in 1.16.5 release.

The messages are also too close together, compared with 1.16.5 release. I suggest submitting a new report for this.

stevecotton:
Thanks for testing - sorry that I opened the PR and immediately went to bed instead of being around to answer questions.

PR #7002, opened 2022-09-06
Separate player-hosted games and bot-hosted games into tabs

### LINKED ISSUE #6939: Bots and How to Handle Them

In their current form bots are not at all problematic but as time goes on many more bots are planned to be added for longer events and it is my opinion that it does not create the best picture, having a lot of the lobby occupied with games that rarely start (for most of the day there is simply not enough players to fill them) might discourage others from playing and makes wesnoth seem like even more of a dead game, as well as possibly creating confusion among new players as some of them are planned to be accessible only after fulfilling certain conditions. 

In the past I know that there was an idea to create a separate tab in the multiplayer lobby for bots, and if that is possible I would support this idea. Another idea idea is to have a filter that filters out bots, but maybe there is a better idea. Also I don't know if these ideas are even possible to be done so that is also something to talk about.

Pentarctagon:
Bot-hosted games are now shown in a separate tab from player-hosted games, based on the new `auto_hosted` attribute on the game creation WML. This is not something that's available to be selected during game creation since it's assumed it will instead be set by the bot implementation.

Resolves #6939

### DIFF
--- a/data/gui/window/lobby_main.cfg
+++ b/data/gui/window/lobby_main.cfg
@@ -3,6 +3,117 @@
 ### Definition of the lobby screen
 ###
 
+#define _GUI_LOBBY_TABS
+[grid]
+
+	[row]
+
+		[column]
+			border = all
+			border_size = 5
+
+			[horizontal_listbox]
+				id = "games_list_tab_bar"
+				horizontal_scrollbar_mode = "never"
+				vertical_scrollbar_mode = "never"
+
+				[list_definition]
+					[row]
+						[column]
+                            [toggle_panel]
+                                linked_group = "tabs"
+
+                                [grid]
+
+                                    [row]
+
+                                        [column]
+                                            border = all
+                                            border_size = 5
+
+                                            [spacer][/spacer]
+
+                                        [/column]
+
+                                        [column]
+                                            grow_factor = 1
+                                            border = all
+                                            border_size = 5
+
+                                            [label]
+                                                id = "tab_label"
+                                                wrap = true
+                                            [/label]
+
+                                        [/column]
+
+                                        [column]
+                                            border = all
+                                            border_size = 5
+
+                                            [spacer][/spacer]
+
+                                        [/column]
+
+                                    [/row]
+
+                                [/grid]
+
+                            [/toggle_panel]
+						[/column]
+					[/row]
+				[/list_definition]
+
+				[list_data]
+
+					[row]
+
+						[column]
+
+							[widget]
+								id = "tab_label"
+								label = _ "Player Hosted"
+							[/widget]
+
+						[/column]
+					[/row]
+
+					[row]
+						[column]
+
+							[widget]
+								id = "tab_label"
+								label = _ "Bot Hosted"
+							[/widget]
+
+						[/column]
+
+					[/row]
+
+				[/list_data]
+			[/horizontal_listbox]
+
+		[/column]
+
+	[/row]
+
+	[row]
+		grow_factor = 1
+
+		[column]
+			grow_factor = 1
+			horizontal_grow = true
+			vertical_grow = true
+
+			{_GUI_GAME_LIST}
+
+		[/column]
+
+	[/row]
+
+[/grid]
+#enddef
+
 #define _GUI_GAME_LIST
 	[listbox]
 		id = "game_list"
@@ -593,6 +704,12 @@
 			id = "tooltip"
 		[/helptip]
 
+		[linked_group]
+			id = "tabs"
+			fixed_width = true
+			fixed_height = true
+		[/linked_group]
+
 		[grid]
 
 			[row]
@@ -741,7 +858,7 @@
 					border_size = 5
 
 					{GUI_FORCE_WIDGET_MINIMUM_SIZE 0 "((screen_height * 35) / 100)" (
-						{_GUI_GAME_LIST}
+						{_GUI_LOBBY_TABS}
 					)}
 				[/column]
 			[/row]
@@ -816,6 +933,12 @@
 			id = "tooltip"
 		[/helptip]
 
+		[linked_group]
+			id = "tabs"
+			fixed_width = true
+			fixed_height = true
+		[/linked_group]
+
 		[grid]
 
 			[row]
@@ -861,8 +984,6 @@
 										[/column]
 									[/row]
 
-									#{GUI_HORIZONTAL_SPACER_LINE}
-
 									[row]
 										grow_factor = 1
 										[column]
@@ -871,11 +992,9 @@
 											horizontal_grow = true
 											vertical_grow = true
 
-											{_GUI_GAME_LIST}
+											{_GUI_LOBBY_TABS}
 										[/column]
 									[/row]
-
-									#{GUI_HORIZONTAL_SPACER_LINE}
 
 									[row]
 										grow_factor = 0
@@ -914,3 +1033,4 @@
 #undef _GUI_FILTER_AREA
 #undef _GUI_GAME_LIST
 #undef _GUI_PLAYER_TREE_AREA
+#undef _GUI_LOBBY_TABS
--- a/src/gui/dialogs/multiplayer/lobby.cpp
+++ b/src/gui/dialogs/multiplayer/lobby.cpp
@@ -36,6 +36,7 @@
 #include "gui/widgets/settings.hpp"
 #include "gui/widgets/text_box.hpp"
 #include "gui/widgets/toggle_panel.hpp"
+#include "gui/widgets/stacked_widget.hpp"
 #include "gui/dialogs/server_info_dialog.hpp"
 
 #include "addon/client.hpp"
@@ -657,6 +658,9 @@ void mp_lobby::pre_show(window& window)
 		}
 	}
 
+	listbox& tab_bar = find_widget<listbox>(&window, "games_list_tab_bar", false);
+	connect_signal_notify_modified(tab_bar, std::bind(&mp_lobby::tab_switch_callback, this));
+
 	// Set up Lua plugin context
 	plugins_context_.reset(new plugins_context("Multiplayer Lobby"));
 
@@ -677,6 +681,12 @@ void mp_lobby::pre_show(window& window)
 	}, true);
 
 	plugins_context_->set_accessor("game_list",   [this](const config&) { return lobby_info_.gamelist(); });
+}
+
+void mp_lobby::tab_switch_callback()
+{
+	filter_auto_hosted_ = !filter_auto_hosted_;
+	update_gamelist_filter();
 }
 
 void mp_lobby::open_profile_url()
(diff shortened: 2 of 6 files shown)


data/gui/window/lobby_main.cfg
```diff
@@ -375,6 +375,18 @@
...
 
+			[column]
+				grow_factor = 0
+				border = "all"
+				border_size = 5
+
+				[toggle_button]
+					definition = "default"
+					id = "filter_auto_hosted"
+					label = _ "Auto-hosted games"
```
stevecotton, line 386: Needs a `# po:` comment telling the translators what this means.

I don't have a suggestion for a clearer English wording, but it doesn't seem clear in English either.

Vultraz: I suggest "Bot-hosted".

Pentarctagon: My concern would just be any sort of negative connotation associated with calling them bot-hosted. Maybe I'm concerned over nothing though.

Discontinuum:
Are you sure it should default to not showing? In that case people most likely never see them

igorbat:
Hi here! As a bot Product Manager I am fully against this change with default is show_auto_hosted=False. The power of bots is only in audience: from our perspective audience will drop significantly(at least 2 times by "Users per week") if we support this param on our side.

If you don't believe i would suggest to make an experiment(push this change in 1.16, see what happens: if metric drop significantly, then roll back the change)

If you want to fight any problem with bots in MP, then can you, please, describe it, may be we(bot managers) can figure out, how to deal with it on our side, instead of just killing bots as system(no jokes, this is how it looks from our perspective -- if you're against the bots we could just shut down them without any commits on your side)
For example, if problem is "we see 3 opened games by bots in lobby", then i am going to add "named hosting"[ when bot's game has prefix "hosted for USERNAME" for TIMEOUT, then game is cancelled ] , in about month, so everyone would know, who asked bot to host).
If problem is: "there are 8 games in play in lobby in 7 of them are "autohost rated", then this solution will lead to "we have 1 game in lobby and 7 hidden" [for players with show_auto_hosted=False] -- are you sure, that this is what you expected?

From our perspective, if default is show_auto_hosted=True then this change is not that useful for any problem which comes to our mind, but we are not against it at all.

Discontinuum:
Overall I think that this is the wrong problem being solved. What problem does this patch try to solve, too many visible bot games? But it's not even what is now, it was only a project.

The principal flaw here is that the core idea of bots is to make it easier for players, including new players and players who don't read forums/groups/discords/etc to play MP games because there is always an open game hosted in the lobby. At the same time it should satisfy the demand for competitive. I wrote about it in detail here https://forums.wesnoth.org/viewtopic.php?t=54861

Hiding bot games by default will make it all meaningless. New people stop coming into the system. And no, they won't read on the forums that to play competitive you need to click on "Allow bot games" filters. They don't read forums. Nor they read the message of the day in the lobby.  They will just see no open rated games.

Also, why has that plan with a lot of open games even come into being? The reason for it is to avoid selective pairing in rated games and to make people play random opponents (without being able to look at the rating of a guy in the queue and not play him if he's too strong). We didn't find any better solution with the current lobby. That is the problem to be solved, I believe

Pentarctagon:
I'm confused why there's confusion over what this is intended to solve - the issue it's resolving is linked in the first post.

Discontinuum:
Because the linked issue is a symptom of the real issue, the lack of UI/UX for automamted competitive events. Solving the symptom by hiding all bots games will just make bots events useless. If people don't see bots games (and they won't see, because they almost never change the default settings/filters/etc), they won't play them (and rn Afterlife Bot generates around a quarter of all MP activity). The whole point of bots was to allow playing in one click, without extra actions such as arranging on the forums/registering on some ladder site/etc

Pentarctagon:
It's also undesirable that there end up being so many bot-hosted games that they end up taking most of the UI space that players see when logging in. Having lots of bots can't result in human-hosted games being sidelined. My understanding is that this isn't the case now, but that there are definitely plans for using more bots to do more things.

igorbat:
Sure, let me describe the situation
- default is show_auto_hosted=False, every bot's game is marked: is the same for us as no bots at all
- default is show_auto_hosted=False, bot's games are marked by our choice(for example, only tournament games are marked, current games are not marked - so for current bot's users nothing is changed): kinda fine for us
- default is show_auto_hosted=True: doesn't solve possible problem
- make an agreement about max always-opened games by bots with bot makers: solve possible future problem
- additional tab in UI for bot games/for competitive OR something else similar to this: solve the problem bots were created for

Pentarctagon:
> additional tab in UI for bot games/for competitive OR something else similar to this: solve the problem bots were created for

I'm not sure that this is equivalent to what bots are doing? There'd need to be a lot more work done for that to be the case than just adding a new tab in the lobby UI.

Discontinuum:
Some bots make sense only when visible by default, like the ones that currently on. They have no conditions to join and are supposed to be joined by novices among others. If they will be hidden among others, it will cut them from any new audience (new people won't notice a small new checkbox)

> I'm not sure that this is equivalent to what bots are doing? There'd need to be a lot more work done for that to be the case than just adding a new tab in the lobby UI.

The tab is more visible than just a filter checkbox, it would be a step forward

Pentarctagon:
The default tab would still be for human-hosted games, to be clear.

Discontinuum:
In the case of the a tab, it's visible that there is another tab (at least I imagine that it would be a big tab-switcher over the main space). A checkbox is significantly harder to notice.

But I agree that the tab would not be quite complete solution, it would allow that the bot corner would not be completely unnoticed and forgotten. The real solution would be some interface that would allow a player to click "Find game" and then join a game, with a random opponent (so that players could not choose the opp, but the system could, considering ratings, etc). But that is probably even harder to make than tabs

Discontinuum:
After some pondering, I think there is another problem with this patch in the context of https://github.com/wesnoth/wesnoth/issues/6939 issue. The problem described in the issue is that a lot of open games that rarely start make not a nice picture, and I agree with that in essence.

But as far as the game started, it doesn't make sense to hide it, it's the running game with living people, which others may want to observe. If bots are popular, hiding already on-going games with living people will hide many games and "makes wesnoth seem like even more of a dead game", exactly what this issue wants to prevent.

So, the summary of what I think so far about this:

* Not all open bot games should be hidden by default, several bot games that allows everyone to join with almost no preconditions should let be shown, they don't clog the space. So the bot should have the possibility to mark its game or not to mark

* When the game started, with living people playing, it doesn't make sense to hide it after that. Abundance of open empty bot games may be not-good looking, but games with living people should be shown independently of how they started. The filter should hide only not yet started games

Pentarctagon:
Entirely non-functional at the moment, but this is what it would look like as tabs:
tabbed

CelticMinstrel:
> * default is show_auto_hosted=False, bot's games are marked by our choice(for example, only tournament games are marked, current games are not marked - so for current bot's users nothing is changed): kinda fine for us

This appears to be what the pull request actually does – by default you don't see bot games, but the bot has to opt in on marking the game as a bot game.

BTW, @Pentarctagon, we should probably also expose this to the plugin game creation API. I don't know if any bots are actually written as a plugin (the API has basically no documentation at all), but it's a possible way to make one. I'm not 100% sure, but I _think_ it'd be as simple as adding it here:

https://github.com/wesnoth/wesnoth/blob/caee3a6dfcb635142518ab456d89fac2a6956a3e/src/gui/dialogs/multiplayer/mp_create_game.cpp#L342-L359

However, since it doesn't have a corresponding user-visible control, you'd probably need custom code rather than the `UPDATE_ATTRIBUTE` macro.

Pentarctagon:
To be honest I have basically no idea what plugins actually are. I think they might be used with the MP tests on CI? If documentation is added then I could take a look at it, but this sounds like something outside of this PR which is meant to mostly be just a UI change.

My understanding though is that the current bots are written in python/go and then connect as their own client.

Discontinuum:
> This appears to be what the pull request actually does – by default you don't see bot games, but the bot has to opt in on marking the game as a bot game.
@CelticMinstrel yeah, seems so, that's fine, why I highlighted it so much is that in the comments of the original issue it was proposed to mark all games of a particular account instead of letting the bot itself decide.

> To be honest I have basically no idea what plugins actually are. I think they might be used with the MP tests on CI?
Although they are not documented at all, I looked at them some time before to investigate whether I can use them. Plugins are lua scripts that are called periodically (as I understand, it's roughly "every tick") and can manipulate the UI (if Wesnoth run with UI disabled, you must specify a plugin, obviously, that would control the game). Several UI elements are exposed as lua functions that a plugin can call. Right now plugins are used only for testing the mp lobby and their API have only functions needed for these test: enter the lobby, create/join game, leave game.

Actually I would love to see plugins API extended, especially if they could at some point access the game state. So that, for example, one could check from the plugin if one of the leaders dead. That would allow a step forward, a bot that is possibly autodetermines the winner (if they play till the leader death). Although my C++ experience is not very large, maybe I can add extra bindings for some other UI-controls, but whether I could expose the game API to it I doubt.

> My understanding though is that the current bots are written in python/go and then connect as their own client.
Current bots are written in Golang and connect with the help of their own client implementation library, yes

CelticMinstrel:
> To be honest I have basically no idea what plugins actually are. I think they might be used with the MP tests on CI? If documentation is added then I could take a look at it, but this sounds like something outside the scope of this PR which is meant to mostly be just a UI change.

They're used in the MP tests, yes.

> Actually I would love to see plugins API extended, especially if they could at some point access the game state. So that, for example, one could check from the plugin if one of the leaders dead. That would allow a step forward, a bot that is possibly autodetermines the winner (if they play till the leader death). Although my C++ experience is not very large, maybe I can add extra bindings for some other UI-controls, but whether I could expose the game API to it I doubt.

Yes, this is a good idea. I'd also like plugins to be able to start a single-player game, or maybe enter the map editor and manipulate it.

Exposing the game state… I wonder if we can directly expose a variable in one Lua kernel to another kernel… I _think_ from the Lua C API's perspective each kernel is simply another thread, so it may be possible…

...anyway, that's off-topic for this pull request. Maybe we should open a new issue about this?

Pentarctagon:
A new issue would be good.

Pentarctagon:
Alright, updated to use tabs instead now. From my local testing it *seems* to be working, but this also definitely needs some testing done with actual bots and players interacting at some point.

Vultraz (changes requested):
Please apply changes discussed on Discord.

Pentarctagon:
@Vultraz this should be addressed now.

src/gui/dialogs/multiplayer/lobby.cpp
```diff
@@ -657,6 +658,15 @@ void mp_lobby::pre_show(window& window)
 		}
 	}
 
+	stacked_widget& pager = find_widget<stacked_widget>(&window, "tabs_container", false);
```
Vultraz, line 661: What is this for still? Should be able to go.

Pentarctagon: The default layer for a stacked widget is set to -1, which would prevent most interactions with the game list if not set to 0 instead.

The `VALIDATE` is there since that appeared to be the standard thing to do when using a stacked widget.

Vultraz (approved):

src/gui/dialogs/multiplayer/lobby.cpp
CelticMinstrel: A stacked widget's "layer" should really be a bit field, as its _intended_ purpose is to display multiple things on top of each other…

More to the point, and this isn't really relevant to the PR anymore, but… even when all the layers are visible, are you saying the stacked widget doesn't allow interactions even with the topmost layer?

PR #7003, opened 2022-09-07
add a [damage_type] special or ability who modify the type of attack used

newfrenchy83:
At the suggestion of @stevecotton, I propose a special [damage_type] capable of modifying the type of attack used when the conditions are met. Its operation is similar to [attack_alignment]: https://github.com/wesnoth/wesnoth/pull/6968. .

the priority rules defined by value of priority_type attribute. lowestest value= highest priority

rules applied alos to "plague" special


data/schema/game_config.cfg
```diff
@@ -92,6 +92,10 @@
         name=alignment
         value="lawful|neutral|chaotic|liminal"
     [/type]
+    [type]
+        name=damage_type
+        value="arcane|blade|cold|fire|impact|pierce"
```
CelticMinstrel, line 97: This is not required. The correct schema type for a damage type is `string`.

src/units/abilities.cpp
```diff
@@ -1120,6 +1120,56 @@ void attack_type::modified_attacks(bool is_backstab, unsigned & min_attacks,
 	}
 }
 
+static void damage_type_list(unit_ability_list& abil_list)
+{	//check if have abilities or specials alignment and remove damage_type bad specified
+	const std::set<std::string> checking_damage = {"arcane", "blade", "cold", "fire", "impact", "pierce"};
```
CelticMinstrel, line 1125: This does not seem sustainable. There shouldn't be anything that requires enumerating all the damage types, because anyone can define a custom damage type.

CelticMinstrel:
> The default priority rules are; fire, arcane, cold, pierce, blade and impact

This seems both arbitrary and nonsensical. Shouldn't it be more of a "last one wins"? Or is there some reason not to do that? If so, we probably need to come up with something else, because anyone can add a custom damage type, and those need to work too.

newfrenchy83:
> > The default priority rules are; fire, arcane, cold, pierce, blade and impact
> 
> This seems both arbitrary and nonsensical. Shouldn't it be more of a "last one wins"? Or is there some reason not to do that? If so, we probably need to come up with something else, because anyone can add a custom damage type, and those need to work too.

and in case of leadership, the last ones will change if location of leaders interwined. Create a default priority is arbitrary, yes, but for what type choose don't change with location of leaders.

newfrenchy83:
for priority, i will justify my choice:
fire, can be used via magic or more clasiical: blade enchanted or fire arrows, torch etc.., it is fore me the more probability.
arcane: more used by magical lawful or chaotic, enchantement with arcane probable.
cold: less what two first.
pierce, blade and impact: physical damage more difficult for me(change a sword to pierce or impact damage?) and low priority.

Like this special must be used for change a damage type to another, the change more frequent will be prioritized except if wml coder use filter for regulate activity of special if na other damage-type with id specified is used in same time.

For custom damage type, i will try to resolve that without risk bug.

CelticMinstrel:
Hmm, I do see the issue with "last one wins" now… but I wonder if there's a better way to handle this…

newfrenchy83:
> Hmm, I do see the issue with "last one wins" now… but I wonder if there's a better way to handle this…
> 
> If/when I create the `[damage_type]` tag (under `[units]`) to _define_ damage types, the priority of each type can be specified there, so (for now) fixing custom damage types to last priority seems okay (even if it gives inconsistent results when you have multiple custom types).

it's why i wanted limit to damage type no custom, because i cannot define rules for two custom damage type in source code andwhat rules must be defined in wml code of special.

CelticMinstrel:
That's not what I'm worried about… I'm wondering if there's some better way to handle priority than a fixed sequence. I suppose… it may be good enough for now, but…

newfrenchy83:
> That's not what I'm worried about… I'm wondering if there's some better way to handle priority than a fixed sequence. I suppose… it may be good enough for now, but…

i pull a commit for change that with use of priority_type attribute. a special/abilities damage_type who use a  priority_type to highest value will be prioritized.

CelticMinstrel:
I did consider this possibility… it does seem to work, yeah. But what's the default if you don't specify `priority_type`? (Also, I think `type_priority` may be a better name.)

newfrenchy83:
> I did consider this possibility… it does seem to work, yeah. But what's the default if you don't specify `priority_type`? (Also, I think `type_priority` may be a better name.)

now, loewest vlaue is highest priority and valeu must be between 1 and 100. If no specified type_priority equals 100.

CelticMinstrel:
Okay, so:

* Lowest priority wins
* No priority specified is sorted last, after any explicit priorities

newfrenchy83:
> I think that seems fine. It's still possible in the future (when damage types are defined with a tag) to make the default priority depend on the damage type, but I think this should be fine for now.

could you give an example?

CelticMinstrel:
> could you give an example?

No. It's something that doesn't exist yet. I have plans to add it "someday".

newfrenchy83:
@CelticMinstrel  change codefor includetype_priority in "plague"

src/units/attack_type.hpp
```diff
@@ -97,6 +97,8 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 	unit_ability_list get_weapon_ability(const std::string& ability) const;
 	/** Returns list who contains get_weapon_ability and get_specials list for each ability type */
 	unit_ability_list get_specials_and_abilities(const std::string& special) const;
+	/** Returns list who contains get_specials_and_abilities with highest priority specified by value of attribute 'type_priority' for each ability plague or damage_type*/
+	unit_ability_list get_specials_and_abilities_with_string(const std::string& special) const;
```
CelticMinstrel, line 101: This is a terrible name for this function. "with_string" gives no indication whatsoever about how it differs from `get_specials_and_abilities`, since both take a single string argument with the same name.

Doesn't "with_priority" make more sense?

CelticMinstrel:
Okay, don't make any changes yet based on this comment.

Over the course of this pull request, we considered two distinct ways of deciding on priority.

1. Based entirely on the damage type.
2. Based on a priority integer specified in the ability.

Note that I never said we shouldn't do the first version, but I do consider the second version acceptable. The questions then is… is there merit in doing _both_?

We _could_ do a combined strategy whereby the `type_priority` key is used if present, and a priority based on the damage type is used if there is no `type_priority` key. That does in this case require hard-coding the priorities of built-in damage types, which I don't like, but that could be removed later if damage types are made more versatile. (Of course, we could also add this secondary level of priorities later, as well.)

Does anyone think this is a good idea?

data/schema/units/specials.cfg
```diff
@@ -93,6 +93,13 @@
 	super="units/unit_type/attack/specials/~generic~"
 	{SIMPLE_KEY type string}
 [/tag]
+[tag]
+	name="damage_type"
+	max=infinite
+	super="units/unit_type/attack/specials/~generic~"
+	{SIMPLE_KEY type string}
+	{SIMPLE_KEY type_priority f_int}
```
CelticMinstrel, line 101: I do want to request that this be changed to `f_real`.

src/units/abilities.cpp
```diff
@@ -1120,6 +1120,42 @@ void attack_type::modified_attacks(bool is_backstab, unsigned & min_attacks,
...
+	//the custom_priority attribute is used for define rules of priority of damage type
+	unit_ability_list abil_list = get_specials_and_abilities(special);
+	unit_ability_list temp_list;
+	int priority = 1;
+	if(abil_list.empty()){
+		return abil_list;
+	}
+	while(priority>=1 && priority<=100){
+		for(const auto& i : abil_list) {
+			int temp_priority = std::clamp((*i.ability_cfg)["type_priority"].to_int(1), 1, 100);
```
CelticMinstrel, line 1134: And this will need to be changed to… I think it's called `as_double`?

CelticMinstrel (changes requested):
I think the existing priority system does have one issue – you can get into a situation where it's not possible to insert an ability between two other abilities in the priority order. That's easy to solve – just don't use an integer for the priority. We can still clamp the priority range to [1,100], that's not a problem.

CelticMinstrel:
> change codefor includetype_priority in "plague"

Okay… this is a good idea but doesn't really belong in this pull request…

I guess we can leave it in here for the time being, though.

src/units/abilities.cpp
newfrenchy83: double, it's done

src/units/attack_type.hpp
newfrenchy83: corrected

newfrenchy83:
> Okay, don't make any changes yet based on this comment.
> 
> Over the course of this pull request, we considered two distinct ways of deciding on priority.
> 
>     1. Based entirely on the damage type.
> 
>     2. Based on a priority integer specified in the ability.
> 
> 
> Note that I never said we shouldn't do the first version, but I do consider the second version acceptable. The questions then is… is there merit in doing _both_?
> 
> We _could_ do a combined strategy whereby the `type_priority` key is used if present, and a priority based on the damage type is used if there is no `type_priority` key. That does in this case require hard-coding the priorities of built-in damage types, which I don't like, but that could be removed later if damage types are made more versatile. (Of course, we could also add this secondary level of priorities later, as well.)
> 
> Does anyone think this is a good idea?

yes, i donethat and i test first step, and it work. Now for damage_type, what no custom damage_type is prioritised:

i choose "fire", "arcane", "cold", "pierce", "blade", "impact", someone have best idea.

data/schema/units/specials.cfg
newfrenchy83: this is done

CelticMinstrel: Actually, I just realized that `f_real` is also wrong. Your code doesn't use `typed_formula` to process the value in here, so it should probably be just `real`.

newfrenchy83: > Actually, I just realized that `f_real` is also wrong. Your code doesn't use `typed_formula` to process the value in here, so it should probably be just `real`. (Maybe `s_real`, if other abilities have a habit of using `s_` prefixed types.)

`s_real`, or `real` then?

CelticMinstrel: I don't see any other `s_` prefixed types in the abilities. so let's go with `real`.

newfrenchy83: done

stevecotton:
The priority and `overwrite_specials` systems seem to be two different concepts that lead to similar goals. Could the priority system be used to replace the `overwrite_specials` system?

I haven't looked at the details yet, but it's a question that would help me understand both this and #7043.

newfrenchy83:
> The priority and `overwrite_specials` systems seem to be two different concepts that lead to similar goals. Could the priority system be used to replace the `overwrite_specials` system?
> 
> I haven't looked at the details yet, but it's a question that would help me understand both this and #7043.

i will think to that but overwrite_specials can be used to overwrite special used by same unit only, or all specials of sametype who affect unit(apply_to=opponent used by opponent)then what priority slect the highpriority specials and return only this one without dicriminate specials usedby unitor used by opponent and with apply_to=opponent.

src/actions/attack.cpp
```diff
@@ -158,7 +158,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	}
 
 	// Handle plague.
-	unit_ability_list plague_specials = weapon->get_specials_and_abilities("plague");
+	unit_ability_list plague_specials = weapon->get_specials_and_abilities_with_priority("plague", {"Soulless", "Walking Corpse"});
```
stevecotton, line 161: I haven't understood what's happening here, but hardcoding unit types into the engine is a no-no. UMC authors have to be able to make custom units by copying WML and Lua from mainline units, without needing to change the C++.

newfrenchy83: changes cancelled

newfrenchy83:
> I think the existing priority system does have one issue – you can get into a situation where it's not possible to insert an ability between two other abilities in the priority order. That's easy to solve – just don't use an integer for the priority. We can still clamp the priority range to [1,100], that's not a problem.

resolved.

newfrenchy83:
> I think the existing priority system does have one issue – you can get into a situation where it's not possible to insert an ability between two other abilities in the priority order. That's easy to solve – just don't use an integer for the priority. We can still clamp the priority range to [1,100], that's not a problem.

it's done.

gfgtdf:
I think this is  a good idea. But

> it's done.

Now you have the same issue but on a lower scale (0.1 instead of of 1), the proper solution is more to just take the lowest/highest number independent of scale.  You can for example use a std algorithm like std::max_element for that.


Also I really don't think the priority by type is needed, in particular since there is no naural order to attacktypes.


Lastly i don't see the need for the "if type is empty use name" logic at the end.

src/units/attack_type.hpp
```diff
@@ -88,6 +88,9 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 	/** Calculates the number of attacks this weapon has, considering specials. */
 	void modified_attacks(unsigned & min_attacks,
 	                      unsigned & max_attacks) const;
+
+	std::string damage_type(const std::string& damage_name) const;
```
CelticMinstrel, line 92: Please add a documentation comment.

CelticMinstrel:
> Also I really don't think the priority by type is needed, in particular since there is no naural order to attacktypes.

Although it's true that there's no natural ordering to damage types, I don't think it's unreasonable to set a default.

> Lastly i don't see the need for the "if type is empty use name" logic at the end.

I don't understand what that's even doing. It makes me question what the purpose of that function even is. I'm kinda confused now.

gfgtdf:
> Although it's true that there's no natural ordering to damage types, I don't think it's unreasonable to set a default

Well it also applies only to the mainline damange types, and the current implementation seems to give abilties with no "type priority" priority over types with that attribute. Also not doing it is also just less code and thus simpler code (in particular to maintain on to keep compatibility with), and i cannot really think of an advantage of this sorting.


Lastly i'm currently wondering whether we should  rename "type priority" to just "priority", i think this could make it easier in the future to stay consistent should we ever add other "nonstacking" abilities that are unrelated to types. But i admit i currently cannot think of such an ability.

CelticMinstrel:
That explanation is pretty incoherent but I think the idea is that _not_ using a priority makes the ability give a different effect depending on how the units are arranged on the map.

newfrenchy83:
> That explanation is pretty incoherent but I think the idea is that _not_ using a priority makes the ability give a different effect depending on how the units are arranged on the map.

yes is that. when [damage_type] used like leadership

newfrenchy83:
> > Although it's true that there's no natural ordering to damage types, I don't think it's unreasonable to set a default
> 
> Well it also applies only to the mainline damange types, and the current implementation seems to give abilties with no "type priority" priority over types with that attribute. Also not doing it is also just less code and thus simpler code (in particular to maintain on to keep compatibility with), and i cannot really think of an advantage of this sorting.
> 
> Lastly i'm currently wondering whether we should rename "type priority" to just "priority", i think this could make it easier in the future to stay consistent should we ever add other "nonstacking" abilities that are unrelated to types. But i admit i currently cannot think of such an ability.

mainline typepriority removed

newfrenchy83:
> > Also I really don't think the priority by type is needed, in particular since there is no naural order to attacktypes.
> 
> Although it's true that there's no natural ordering to damage types, I don't think it's unreasonable to set a default.
> 
> > Lastly i don't see the need for the "if type is empty use name" logic at the end.
> 
> I don't understand what that's even doing. It makes me question what the purpose of that function even is. I'm kinda confused now.

removed also

newfrenchy83:
> I think this is a good idea. But
> 
> > it's done.
> 
> Now you have the same issue but on a lower scale (0.1 instead of of 1), the proper solution is more to just take the lowest/highest number independent of scale. You can for example use a std algorithm like std::max_element for that.
> 
> Also I really don't think the priority by type is needed, in particular since there is no naural order to attacktypes.
> 
> Lastly i don't see the need for the "if type is empty use name" logic at the end.

code with std::max_element added instead of interval

src/units/attack_type.hpp
CelticMinstrel: There are some typos here, two of the `is` look like they should be `if`?

CelticMinstrel:
I prefer having the default priority based on damage type, but if we're not doing that, I think maybe priority should be a required key (`REQUIRED_KEY` in the schema, code throws an error or something if missing) instead of an optional key. It can stay optional in `[plague]` if that's easy.

gfgtdf:
>  I think maybe priority should be a required key (REQUIRED_KEY in the schema, code throws an error or something if missing) 

I don't think that'd solve things, people would just  explicitly put the default 0, and we'd still have  the same problem. you mentioned  before with leadership like abilities.

> I prefer having the default priority based on damage type, but if we're not doing that,instead of an optional key. It can stay optional in [plague] if that's easy.

I'm not strictly against the type priority put it should  also apply when the priorites  are the same (not only when they are not present.), and in order to also apply to umc damagetypes   it should  probably just sort them be alphabeticially.

Like a code  that orders by "priority" as first criteria and "type" by second criteria.

gfgtdf:
Like noted before i think, the type of priority should be double not int so that people can insert values between, (for that you dont need to change the code much just change the type from double to int and use to_double instead of to_int)

gfgtdf:
Oh and i wonder should the type key support formula , something like `type="if(other.level > 1, 'corpse', ''zombie)"`, i mean since this can be easily emulated with filters its probably not that important, but if it did support that before it shouldn't break it.

newfrenchy83:
> Oh and i wonder should the type key support formula , something like `type="if(other.level > 1, 'corpse', ''zombie)"`, i mean since this can be easily emulated with filters its probably not that important, but if it did support that before it shouldn't break it.

i don't want risk that

gfgtdf:
Hmm okay so behavior-wise i think looks is okay and mergable now. I personally would probably would write it a little differently but we can always do this after it is merged.

CelticMinstrel:
> just sort them be alphabeticially

I think that's kinda a ridiculous suggestion.

> probably to put everything except the the "flamin sword in water" commit into one i'd say.

Oh right, I was meaning to comment on that. I don't think it should be in this pull request at all. If you want to open a separate pull request to add it later, fine, but I don't think it's reasonable alter campaigns just to add an example of a new feature. (And if the ability isn't used anywhere then it shouldn't be defined.)

gfgtdf:
> I think that's kinda a ridiculous suggestion

Well its just as arbitiary as the current order, and unlike the current order, it solves the "two damage-type leadership different, if position of leader is inverted" problem not only for mainline damage types but also for custom umc damage types.

gfgtdf:
Oh i see you changed the order of the list of  6 to be in aphabetial order, that was not the point of my suggestion, the suggestion was to get rid of the list, and just to compare the type strings against each other, for example like more or less (untested code)
```
auto res = std::max_element(ab.begin(), ab.end(); [](const auto& first, const auto& second) {
    auto pri = [](const auto& ab) { return (*ab.ability_cfg)["priority"].to_double(0); }
    auto tye = [](const auto& ab) { return (*ab.ability_cfg)["type"].to_string(); }
    return pri(first) < pri(second) || (pri(first) == pri(second) && tpe(first) < tpe(second))
})

```

CelticMinstrel:
> Well its just as arbitiary as the current order

I don't think I agree on this? Compare the two orderings:

`{"fire", "arcane", "cold", "pierce", "blade", "impact"}`

`{"arcane", "blade", "cold", "fire", "impact", "pierce"}`

The second makes no sense whatsoever. Physical and magical damage types are freely intermixed with no logic at all. The first at least groups the magical and physical types in the ordering. Both are arbitrary, yes, but I think the first makes more sense than the second. I could still see reordering the damage types within the magical/physical grouping, perhaps.

> it solves the "two damage-type leadership different, if position of leader is inverted" problem not only for mainline damage types but also for custom umc damage types.

Well yes, _this_ part is true, I just don't think it's sufficiently compelling as an argument for making them be ordered alphabetically.

Actually, some day I'd probably want to define everything about a damage type in a global tag, and then that tag would specify the default priority for this ability, but we aren't there yet.

gfgtdf:
> Well yes, this part is true, I just don't think it's sufficiently compelling as an argument for making them be ordered alphabetically

Imo this is just a fallback sorting anyways to get rid of possible weird behavior, also to be more robust against things like oos (it's from a wml developers point sometimes hard to see whats allowed/dangerous, for example a code that removes and readds an ability could be intuitively considered "non-game-changing" but it might actually change the order of the abilities), no addon author should rely on the concrete ordering, if they want a reliable ordering, just use `priority`.

So i consider applying to all damage types more important, than using an ordering that is a little less arbitiary. Also the aphabetial ordering can be applied to plague aswell for the same reasons.

newfrenchy83:
> 

what utility of 'res' in code?

src/units/abilities.cpp
```diff
@@ -1145,49 +1145,68 @@ unit_ability_list attack_type::get_specials_and_abilities_with_priority(const st
...
-	type = (*abil_list.front().ability_cfg)["type"].str();
-	if(type.empty()) {
-		type = (*abil_list.front().ability_cfg)["name"].str();
+	return ((*abil_list.front().ability_cfg)["type"].str());
+}
+
+std::string attack_type::damage_type(const std::string& damage_name) const
+{
+	//order of mainline type when 'priority' attribute not used.
+	std::vector<std::string> checking_damage_type = {"arcane", "blade", "cold", "fire", "impact", "pierce"};
```
CelticMinstrel, line 1204: What. No.

gfgtdf:
> @CelticMinstrel ,@gfgtdf, i removed Htt and NR modicication but i must point a problem, when damag_type used, filter_weapon recognize native type of attack used and not modification by damage_type.

How is it working for the other specials ([damage], [attacks], [chance_To_hit]),? If those don't effect [filter_weapon] i'd expect [damage_type] doesn't effect [filter_weapon] either.

newfrenchy83:
> > @CelticMinstrel ,@gfgtdf, i removed Htt and NR modicication but i must point a problem, when damag_type used, filter_weapon recognize native type of attack used and not modification by damage_type.
> 
> How is it working for the other specials ([damage], [attacks], [chance_To_hit]),? If those don't effect [filter_weapon] i'd expect [damage_type] doesn't effect [filter_weapon] either.

For [damage] and [attacks], it seems that it is the damage and number of attacks not modified by the specials which are filtered, probably to avoid problems with infinite loops. What bothers me is that as it stands, damage_type is not the perfect equivalent of a set_type object type modification because in case of filtering, the new damage type is taken into account in the latter case but not with damage_type. Unless possibly being able to filter damage_type only for [filter_opponent for example, I don't see how to solve this problem without introducing a risk of a loop which would prevent using a programmed leadership damage_type to change a damage type by the one implemented in his code

newfrenchy83:
the situation is worse than I thought, because if two swordsman who has a special damage_type that changes blade to active fire only if the opposing attack is blade clashes, the infinite loop problem also manifests, for this reason and until further notice this request should not be merge into master

Pentarctagon:
Is the feature worth the added complexity?

newfrenchy83:
> the situation is worse than I thought, because if two swordsman who has a special damage_type that changes blade to active fire only if the opposing attack is blade clashes, the infinite loop problem also manifests, for this reason and until further notice this request should not be merge into master

i done an error, the infinite loop what i see was when i partially implemented an code for resolve loop in [filter_sel/student][filter_weapon] but in fact in filter_opponent, notinfinite_loop.

newfrenchy83:
> Is the feature worth the added complexity?

i fear what not

CelticMinstrel:
So have we decided not to do this after all?

Even if we're not adding `[damage_type]`, I think adding priority to `[plague]` could be worthwhile.

newfrenchy83:
I will reopen PR for propose an fix for resolve type filtering without create infinite loop, if damage type original filtered in damage_type only but what other filtering(other specials, event, animation or lua) matche with type changed thenno problems.

CelticMinstrel:
> @stevecotton now what [filter_attack] fixed in [88439d6](https://github.com/wesnoth/wesnoth/commit/88439d6427a81a6b131acd8b0f7bf04b2d981d71), this PR has chance to be merged anew.

There's an issue with that commit that should probably be fixed first.

newfrenchy83:
> > @stevecotton now what [filter_attack] fixed in [88439d6](https://github.com/wesnoth/wesnoth/commit/88439d6427a81a6b131acd8b0f7bf04b2d981d71), this PR has chance to be merged anew.
> 
> There's an issue with that commit that should probably be fixed first.

issue fixed and added https://github.com/wesnoth/wesnoth/pull/7003/commits/6ad9c51fbb18f1322803563216dc0df62c283bec to this PR

src/units/abilities.cpp
```diff
@@ -1113,6 +1113,64 @@ void attack_type::modified_attacks(unsigned & min_attacks,
...
+
+	std::vector<double> priorityvec;
+	std::vector<double>::iterator priority;
+	for(const auto& i : abil_list) {
+		priorityvec.push_back((*i.ability_cfg)["priority"].to_double(0));
+	}
+
+
+	priority = std::max_element(priorityvec.begin(), priorityvec.end());
+	int prc = std::count(priorityvec.begin(), priorityvec.end(), *priority);
```
stevecotton, line 1133: A couple of lines here are using exact equality with floating-point values. I haven't yet worked out what it should do instead, just making a note for myself.

CelticMinstrel: It's not _necessarily_ incorrect to use exact equality with floating-point values. It might be fine in this context – we're not doing math on the values, just parsing them from text and comparing them to each other.

newfrenchy83:
@stevecotton another comments in https://forums.wesnoth.org/viewtopic.php?p=681079#p681079

newfrenchy83:
@stevecotton @CelticMinstrel could you review https://github.com/wesnoth/wesnoth/pull/7003/commits/35ac9eb60820376c4204873640380aebad233509 , please? I hope what tags like modify_unit_type work in tests.

newfrenchy83:
@stevecotton @CelticMinstrel https://github.com/wesnoth/wesnoth/pull/7003/commits/3002e93dd9daea931efbc303d110ebe9cb257ea1 checking succefull(in progress for macos), and not without pain.

newfrenchy83:
@nemaara renamed in https://github.com/wesnoth/wesnoth/pull/7003/commits/a7f9075ace1ba3900a14f57fdeb351ab6be55f03

newfrenchy83:
@stevecotton imodified cource code for what damage type coded in special added to native damge type(it is the type who will make more damage who willbe selected). It is perhaps the only things who can diddrentiate [damage_type] to using to set_type in [effect]

newfrenchy83:
@stevecotton @CelticMinstrel added modification in https://github.com/wesnoth/wesnoth/pull/7003/commits/978ab177abecac5a4cc3904a29499b99614d4e10 with change of Ui in 
Capture d’écran 2023-06-13 105555 and

Capture d’écran 2023-06-13 105719

newfrenchy83:
> @stevecotton @CelticMinstrel added modification in [978ab17](https://github.com/wesnoth/wesnoth/commit/978ab177abecac5a4cc3904a29499b99614d4e10) with change of Ui in Capture d’écran 2023-06-13 105555 and
> 
> Capture d’écran 2023-06-13 105719

i alos modified codefor what ai and ai formula integrate damage_type()

newfrenchy83:
@stevecotton because add of https://github.com/wesnoth/wesnoth/commit/1ecc30972e00ccfb6911427b28cf1186d6648ddc, i must rebase this PR with https://github.com/wesnoth/wesnoth/pull/7746 for use this code instead of actual priority or open another PR with rebasement

newfrenchy83:
> @stevecotton because add of [1ecc309](https://github.com/wesnoth/wesnoth/commit/1ecc30972e00ccfb6911427b28cf1186d6648ddc), i must rebase this PR with #7746 for use this code instead of actual priority or open another PR with rebasement

rebased with https://github.com/wesnoth/wesnoth/pull/7746 for resolve issue

data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg
```diff
@@ -1553,7 +1553,14 @@
...
                                 range=melee
-                                set_type=arcane
+                                [set_specials]
+                                    mode=append
+                                    [damage_type]
+                                        id=arcane_damage
+                                        type=arcane
+                                        cumulative=yes
+                                    [/damage_type]
+                                [/set_specials]
```
stevecotton, line 1563: Although this is a method of fixing bug #7774, the code now in #7775 is a simpler way to do it. Either this code in UtBS S02 needs to know about the event in S03, or that event needs to know about this event; however doing fixing it in S03 just needs a
```
[not]
    type=arcane
[/not]
```

newfrenchy83: excuse me, i added https://github.com/wesnoth/wesnoth/pull/7003/commits/5e176e58da98aeda4d0b45a9c5019aab03e0dfea for fixing that. I don't think of https://github.com/wesnoth/wesnoth/issues/7774 issue when i postted https://github.com/wesnoth/wesnoth/pull/7003/commits/fd270c6cf0fd3afd7cbf3b8650ea6d8962292416

stevecotton: Would this special still be active even if the primary type of the attack was changed? I expected that it would, therefore the code in #7775 would change a (blade + arcane) weapon into a (fire + arcane) weapon, even without fd270c6.

newfrenchy83: Yes, it would remain active since I didn't implement a type-based filter, so even if I hadn't added a pure S3 secondary_type arcane filter, we would have fire+arcane.

newfrenchy83: I added the filter in S3 with https://github.com/wesnoth/wesnoth/commit/5e176e58da98aeda4d0b45a9c5019aab03e0dfea because it is the blessing of an ordinary weapon that competes with holy water, and one should have either fire or arcane but not fire+arcane, and therefore incase already blessed weapons, the fire blessing should not apply.

stevecotton:
I got a stack overflow running this test with `-t` and then hovering the mouse over Bob. It caused infinite recursion inside attack_type::special_active_impl. This was added to top of commit ddd9ffb135a5aac4767e5bbeafca12aed36585ce.

The test isn't ready yet, was just WIP taking your suggestions from #7746.

```
#####
# API(s) being tested: [overwrite]priority=,[overwrite][filter_specials]
##
# Actions:
# (Unnecessary?) Give alice an ability that blunts his weapons, changing blade and pierce to impact.
# Give Bob a (unnecessary? medium priority) teaching ability that changes blade, pierce and impact weapons to fire.
# Give Bob a high priority teaching ability that changes melee weapons to arcane.
##
# Expected end state:
# Alice's sword has damage type arcane, and her bow has damage type fire.
#####
{COMMON_KEEP_A_B_UNIT_TEST "override_priority" (
    [event]
        name=start

        [modify_unit]
            [filter]
            [/filter]
            max_hitpoints=100
            hitpoints=100
            attacks_left=1
        [/modify_unit]

        [object]
            silent=yes
            [effect]
                apply_to=new_ability
                [abilities]
                    [attacks]
                        value=10
                        [filter_student]
                            [filter_weapon]
                                type=blade
                            [/filter_weapon]
                        [/filter_student]
                    [/attacks]
                    [damage]
                        value=1
                    [/damage]
                    [chance_to_hit]
                        value=100
                    [/chance_to_hit]
                [/abilities]
            [/effect]
            [filter]
                id=alice
            [/filter]
        [/object]

        [object]
            silent=yes
            [effect]
                apply_to=new_ability
                [abilities]
                    [damage_type]
                        id=fire_damage
                        type=fire
                        [filter_student]
                            [filter_weapon]
                                type=blade
                                [or]
                                    type=impact
                                [/or]
                                [or]
                                    type=pierce
                                [/or]
                            [/filter_weapon]
                        [/filter_student]
                        overwrite_specials=one_side
                        [overwrite]
                            [filter_specials]
                                id=impact_damage
                            [/filter_specials]
                        [/overwrite]
                        [affect_adjacent]
                        [/affect_adjacent]
                    [/damage_type]

                    [damage_type]
                        id=melee_arcane
                        type=arcane
                        [filter_student]
                            [filter_weapon]
                                range=melee
                            [/filter_weapon]
                        [/filter_student]
                        overwrite_specials=one_side
                        [overwrite]
                            priority=3
                        [/overwrite]
                        [affect_adjacent]
                        [/affect_adjacent]
                    [/damage_type]

                [/abilities]
            [/effect]
            [filter]
                id=bob
            [/filter]
        [/object]

        # Currently crashes (stack overflow, infinite recursion) inside attack_type::special_active_impl when running with -t and hovering over Bob

        {SUCCEED}
    [/event]
)}
```

newfrenchy83:
> I got a stack overflow running this test with `-t` and then hovering the mouse over Bob. It caused infinite recursion inside attack_type::special_active_impl. This was added to top of commit [ddd9ffb](https://github.com/wesnoth/wesnoth/commit/ddd9ffb135a5aac4767e5bbeafca12aed36585ce).
> 
> The test isn't ready yet, was just WIP taking your suggestions from #7746.
> 
> ```
> #####
> # API(s) being tested: [overwrite]priority=,[overwrite][filter_specials]
> ##
> # Actions:
> # (Unnecessary?) Give alice an ability that blunts his weapons, changing blade and pierce to impact.
> # Give Bob a (unnecessary? medium priority) teaching ability that changes blade, pierce and impact weapons to fire.
> # Give Bob a high priority teaching ability that changes melee weapons to arcane.
> ##
> # Expected end state:
> # Alice's sword has damage type arcane, and her bow has damage type fire.
> #####
> {COMMON_KEEP_A_B_UNIT_TEST "override_priority" (
>     [event]
>         name=start
> 
>         [modify_unit]
>             [filter]
>             [/filter]
>             max_hitpoints=100
>             hitpoints=100
>             attacks_left=1
>         [/modify_unit]
> 
>         [object]
>             silent=yes
>             [effect]
>                 apply_to=new_ability
>                 [abilities]
>                     [attacks]
>                         value=10
>                         [filter_student]
>                             [filter_weapon]
>                                 type=blade
>                             [/filter_weapon]
>                         [/filter_student]
>                     [/attacks]
>                     [damage]
>                         value=1
>                     [/damage]
>                     [chance_to_hit]
>                         value=100
>                     [/chance_to_hit]
>                 [/abilities]
>             [/effect]
>             [filter]
>                 id=alice
>             [/filter]
>         [/object]
> 
>         [object]
>             silent=yes
>             [effect]
>                 apply_to=new_ability
>                 [abilities]
>                     [damage_type]
>                         id=fire_damage
>                         type=fire
>                         [filter_student]
>                             [filter_weapon]
>                                 type=blade
>                                 [or]
>                                     type=impact
>                                 [/or]
>                                 [or]
>                                     type=pierce
>                                 [/or]
>                             [/filter_weapon]
>                         [/filter_student]
>                         overwrite_specials=one_side
>                         [overwrite]
>                             [filter_specials]
>                                 id=impact_damage
>                             [/filter_specials]
>                         [/overwrite]
>                         [affect_adjacent]
>                         [/affect_adjacent]
>                     [/damage_type]
> 
>                     [damage_type]
>                         id=melee_arcane
>                         type=arcane
>                         [filter_student]
>                             [filter_weapon]
>                                 range=melee
>                             [/filter_weapon]
>                         [/filter_student]
>                         overwrite_specials=one_side
>                         [overwrite]
>                             priority=3
>                         [/overwrite]
>                         [affect_adjacent]
>                         [/affect_adjacent]
>                     [/damage_type]
> 
>                 [/abilities]
>             [/effect]
>             [filter]
>                 id=bob
>             [/filter]
>         [/object]
> 
>         # Currently crashes (stack overflow, infinite recursion) inside attack_type::special_active_impl when running with -t and hovering over Bob
> 
>         {SUCCEED}
>     [/event]
> )}
> ```

fixed with https://github.com/wesnoth/wesnoth/pull/7003/commits/86309693c0a97e7dbe05dbc74a67965e43c9307a

PR #7006, opened 2022-09-08
DiD S8: Handle situations where factions have not yet appeared

### LINKED ISSUE #7005: Malin talks about people he doesn't know about / don't exist

- **Version:** 1.17.4
- **Downloaded from:** wesnoth.org
- **Build info:** N/A
- **OS:** N/A

### Description of the bug

In DiD scenario "All alone", Malin still says, when leaving:

"The Terrain is hazardous in these mountains. Neither the paladins nor Dela will be able to pursue me."

even if Dela hasn't arrived and Cadeus' forces are gone

In "All alone", get the book and kill all of Cadeus' troops, bringing it to your camp by, say, turn 13.

### Expected behavior

Malin should not mention Dela at this point, and even mention of any Tath forces should refer to the possibility that _other_ parties might be searching for him.

### Additional context

_No response_

Wedge009:
Resolves #7005.

I think this is perhaps overkill, but clearly at least some people notice the discrepancy in the dialogue if the scenario is completed before certain factions have appeared.

I consider 'paladins' as a generalisation for Cadaeus' forces, so there doesn't need to be yet another piece of dialogue differentiating whether or not Cadaeus is alive. Similarly, I generalise Dela's forces as 'Dela's men' so it shouldn't matter if Dela is alive or not. Regardless, the underlying bug report requires that there be work for the translators.

I think this scenario file is the same in 1.16 branch, so could be back-ported. I have tested the three phases as stated in the commentary: before Cadaeus appears, after Cadaeus appears but before Dela appears, and after Dela appears. It makes sense to me, given the context of the bug report.

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
@@ -405,7 +405,6 @@
         {MODIFY_UNIT (id=Malin Keshar) facing ne}
 
         {VARIABLE dela_arrived no}
-        {VARIABLE cadaeus_arrived no}
     [/event]
 
     # This helps prevent Darken from getting himself killed easily by preventing
@@ -689,8 +688,6 @@
             speaker=Darken Volk
             message= _ "Pompous words from an arrogant fool. We shall see how well you fare once I have finished this upstart."
         [/message]
-
-        {VARIABLE cadaeus_arrived yes}
     [/event]
 
     # Dela joins the party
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
I guess this isn't your fault, but why on earth is it using integer variables for boolean values…?

Wedge009:
Maybe the original author was, like me, not an expert in WML? At any rate, I've changed the variables to boolean, but have not tested this version.

Descacharrado (approved):
I tested it, looks good to me. The fact that Malin comments on the paladins even after their leader is dead seems to fit, as paladins are a force to fear in this scenario and Malin would most likely be worried of the possibility of more coming.

data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
```diff
@@ -1080,10 +1083,52 @@
...
+                [/then]
+            [/elseif]
+
+            # Phase three: Cadaeus and Dela have arrived
+            # Generalise as 'Dela's men' in case Dela herself is killed.
+            # The scenario of both Cadaeus and Dela being killed is handled in the 'alternate victory' event.
+            [else]
+                [message]
+                    speaker=Malin Keshar
+                    message= _ "The terrain is hazardous in these mountains. Neither the paladins nor Dela’s men will be able to pursue me."
```
stevecotton, line 1128: Change, or add a `# po:` comment to say that `men` should be translated as gender-neutral?

Wedge009:
Thanks for testing. Yes, that was the intention (regarding the possibility of more paladins 'off screen'). Alternatively one would have to account for every possible combination of units being active or dead, leaders being active or dead, etc. As I implied at the start, I think that would be getting too nit-picky.

data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
```diff
@@ -1080,10 +1083,53 @@
...
+            [and]
+                [variable]
+                    name=dela_arrived
+                    boolean_equals=no
+                [/variable]
+            [/and]
+            [then]
+                [message]
+                    speaker=Malin Keshar
+                    message= _ "The terrain is hazardous in these mountains. No one should be able to pursue me."
```
nemaara, line 1103: Can we just change the line to this for all cases?

```suggestion
                    message= _ "The terrain is hazardous in these mountains. No one will be able to pursue me."
```

Wedge009: So just get rid of the references to paladins and Dela altogether? Okay...

nemaara: I think it might be easier this way? It sounds more like text he'd actually say than something more verbose anyway.

nemaara (approved):

PR #7007, opened 2022-09-10
1.17 alternate high/low terrain levels

doofus-01:
What this PR does:
1.  Provides an additional raised level and lowered level
2.  Improves the appearance in editor and makes use more intuitive (though I still won't claim it is intuitive)

What this PR won't do:
1.  Extra tall cliff transitions (i.e. it is always one step up or down)

I'm trying to figure out if there is some way to get a Möbius strip rule-set that makes it possible to build shallow staircases by alternating available levels.  At the moment, prospects are not looking good.
Screenshot_20220910_a


doofus-01:
This is an example of the five levels available.  It makes me realize the wooden/stone built transitions aren't so well suited to likely use-cases - transitions should probably be less jagged.
Screenshot_20220910_b

doofus-01:
This is not completed, but I'm inclined to merge anyway (assuming it passes CI) because it fixes some bugs I introduced in previous commits.

The more I look at this, the more I wish for some engine changes that I think would be optional/minor, but would obviate my terrain graphics hacks and allow interesting combat options in game.  I'll open an issue for that.

PR #6538, opened 2022-02-21
Fix some minor bugs in WC campaign

wuxiangjinxing:

PR #7012, opened 2022-09-17
Re pr 6538

Pentarctagon:
Re-PR of #6538 without the problematic second commit.

### DIFF
--- a/data/campaigns/World_Conquest/lua/map/scenario_utils/plot.lua
+++ b/data/campaigns/World_Conquest/lua/map/scenario_utils/plot.lua
@@ -49,7 +49,11 @@ function add_plot(scenario, scenario_num, nplayers)
 		end
 		start_message(ai1, true, _ "Foolish upstarts. You will never defeat us! Come, my allies. Let us crush them!")
 		start_message(ai1, false, _ "Yes! We will drive them into the sea!")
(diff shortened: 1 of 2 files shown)


data/campaigns/World_Conquest/lua/map/scenario_utils/plot.lua
```diff
@@ -49,7 +49,11 @@ function add_plot(scenario, scenario_num, nplayers)
 		end
 		start_message(ai1, true, _ "Foolish upstarts. You will never defeat us! Come, my allies. Let us crush them!")
 		start_message(ai1, false, _ "Yes! We will drive them into the sea!")
-		start_message("1,2,3", false, _ "Never fear! Your trusted friends stand beside you!")
+		if nplayers == 4 then
+			start_message("4", true, _ "Never fear! Your trusted friends stand beside you!")
```
soliton-, line 53: This seems weird. I would expect you'd just add ",4" to the first parameter. Now you either get a message from side 4 or from side 1, 2 and 3.

Also this changes a single `start_message` call when there are plenty more similar ones as well as `end_message`.

knyghtmare (approved):
it's good to do, but I think at some point adding properly tuned plot messages for WC might be worth doing instead of re-assigning dialogue to another new player instead of a supporting unit.

PR #7013, opened 2022-09-17
MP tooltips: allow line breaks between hexagons by using Unicode zero width space

walodar:
To allow potential line breaks between the hexagons in MP tooltips, this pull request inserts (invisible) Unicode zero width space characters between them.

Previously, I have already used this little trick for the hexagons in the help browser.

With this pull request, the same thing is done for the MP tooltips in the sidebar and the recruitment dialog, which covers all places where MP hexagons appear.

This only makes a visible difference if a unit's movepoints are extremely large, quite beyond any mainline unit, but I see no drawbacks.

### DIFF
--- a/src/gui/widgets/unit_preview_pane.cpp
+++ b/src/gui/widgets/unit_preview_pane.cpp
@@ -196,7 +196,8 @@ static inline std::string get_mp_tooltip(int total_movement, std::function<int (
 			const int movement_hexes_per_turn = total_movement / tm.moves;
 			tooltip << " ";
 			for(int i = 0; i < movement_hexes_per_turn; ++i) {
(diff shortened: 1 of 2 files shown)


stevecotton:
The build failure looks like a one-off incident (one of the unrelated tests timing out), so I've triggered a rebuild.

walodar:
Thank you! I am happy that all tests passed now.

stevecotton:
The UI looks like this, which seems a good way to handle units that should almost never appear in-game:
wrapped_movement

It's a tiny change, which is why I've been quicker about reviewing it than the more complex PRs.

walodar:
Great, thank you!

PR #7014, opened 2022-09-19
1.17 elevated terrain work

doofus-01:
Revised the stone stairs to be less jagged, and worked on some corner transitions.  The job isn't finished, as can be seen in the image below, and I haven't touched the wooden transitions.
Screenshot_20220918_a


Jonathan-Kelly:
For some reason, I find the stairs in that example image appear to be in an overhead perspective rather than an isometric one. Maybe because the southern staircase would be edge-on with our isometric viewpoint and so mostly invisible?

doofus-01:
Wesnoth graphics aren't really isometric, but I see what you mean.  I think the top-down perspective is from multiple things, some I can fix more easily than others, I just made some revisions that should help.

knyghtmare:
If you're adding a new map, the map-diff shouldn't be used as there is no comparison to be made.

I suggest ignoring that map-diff check as it's faulty/irrelevant here.

doofus-01:
Yeah, I'm not going to let Map Diff stop me.  I'll merge if that's the only failure.

PR #7016, opened 2022-09-20
Fix unknown speaker errors

Dou1622:
Fix unknown speaker or id errors
See #4494

### DIFF
--- a/data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
@@ -133,6 +133,7 @@
             [filter]
                 type="Fire Ant"
             [/filter]
+            # wmllint: recognize guarding_nest
             id=guarding_nest
             distance=3
             station_x,station_y=48,10
@@ -392,6 +393,7 @@ My sanction is given."
             x,y=28,16
         [/unit]
         [message]
+            # wmllint: recognize Karron
             speaker=Karron
             message= _ "So it is, that timid Gorlack attempts the Trial of Gaall."
         [/message]
--- a/data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
@@ -133,7 +133,6 @@
             [filter]
                 type="Fire Ant"
             [/filter]
-            # wmllint: recognize guarding_nest
             id=guarding_nest
             distance=3
             station_x,station_y=48,10
--- a/data/campaigns/Winds_of_Fate/scenarios/11_Crosswind.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/11_Crosswind.cfg
@@ -159,7 +159,7 @@
             [leader_goal]
                 x,y=4,36
                 auto_remove=yes
-                    #wmllint: recognize land_on_keep
+                #wmllint: recognize land_on_keep
                 id=land_on_keep
                 max_risk=1
             [/leader_goal]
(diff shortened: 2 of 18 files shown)


Pentarctagon:
As a side question I'm somewhat curious how wmllint determines what a valid speaker is vs isn't, but I'm fine this change as it is regardless of the answer to that.

data/campaigns/Winds_of_Fate/scenarios/11_Crosswind.cfg
```diff
@@ -159,6 +159,7 @@
             [leader_goal]
                 x,y=4,36
                 auto_remove=yes
+                #wmllint: recognize land_on_keep
```
CelticMinstrel, line 162: This isn't a unit so it's incorrect to use `recognize` on it.

data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
```diff
@@ -133,6 +133,7 @@
             [filter]
                 type="Fire Ant"
             [/filter]
+            # wmllint: recognize guarding_nest
```
CelticMinstrel, line 136: This also isn't a unit, so `recognize` is incorrect.

CelticMinstrel:
> As a side question I'm somewhat curious how wmllint determines what a valid speaker is vs isn't, but I'm fine this change as it is regardless of the answer to that.

I believe it scans for explicitly-placed units (in `[side]` and `[unit]` tags) and considers those units to be valid. I'm not sure if it also considers `[recall]` tags.

I've noted two IDs that you added `recognize` to that have nothing to do with units. If wmllint is reporting them as invalid speakers, then that is a bug in wmllint and should be fixed in wmllint – it should not analyze IDs nested under AI tags (`[ai]`, `[micro_ai]`, or `[modify_ai]`) unless they are in a tag whose name starts with "filter" (generally `[filter]`, `[filter_second]`, `[filter_own]`, or `[filter_enemy]`).

Dou1622:
> > As a side question I'm somewhat curious how wmllint determines what a valid speaker is vs isn't, but I'm fine this change as it is regardless of the answer to that.
> 
> I believe it scans for explicitly-placed units (in `[side]` and `[unit]` tags) and considers those units to be valid. I'm not sure if it also considers `[recall]` tags.
> 
> I've noted two IDs that you added `recognize` to that have nothing to do with units. If wmllint is reporting them as invalid speakers, then that is a bug in wmllint and should be fixed in wmllint – it should not analyze IDs nested under AI tags (`[ai]`, `[micro_ai]`, or `[modify_ai]`) unless they are in a tag whose name starts with "filter" (generally `[filter]`, `[filter_second]`, `[filter_own]`, or `[filter_enemy]`).

Do IDs should be analyzed in any "filter" tag? Is there a "filter" tag list?

Pentarctagon:
Every filter tag currently available is (I believe) present on the WML tags sidebar on the wiki (ie: on the right side of https://wiki.wesnoth.org/ReferenceWML) under the F section.

CelticMinstrel:
I don't think there's any tag that ends with `_filter`. As far as I know, nearly all filter tags begin with `filter_`. I _can_ think of one exception (other than the obvious `[filter]`): `[goal]` tags (and the other shorthand tags for defining goals) contain a `[criteria]` tag which is a filter (can be either a unit or location filter). Note that only _unit_ filters can contain an ID key, so `[filter_location]` for example isn't a concern.

Wedge009:
I reverted the two instances of non-units, but the rest makes sense to me. It's been long enough, I think this can be merged assuming the checks pass, right?

Pentarctagon:
That sounds fine to me.

Issue #7017, opened 2022-09-22
Reconsider movement cost colouring


eyalroz:
### Game and System Information

- **Version:** 1.17.7
- **Downloaded from:** Sources from Wesnoth website
- **Build info:** N/A
- **OS:** Devuan GNU/Linux, but it doesn't matter
-

### Description of the bug

Between v1.17.4 and v1.17.7 the unit info dialog/pane changed. Among other changes, the "Movement Cost" column now also shows the number of hexes, using brightly-colored full-hex-shaped icons, in addition to the movement points cost of passing one hex of the terrain type.

This is a repetition of information that is already available - the movement point cost. But - that is not a bug; the question is - is it worthwhile to add this different graphical representation of the same information.

I believe that it isn't. Why?

1. It looks garish - the screen is full of brightly-colored spots all over; that is distracting and annoying.
2. The number is irrelevant in multi-terrain movement - and movement is usually multi-terrain.
3. The player knows they need to count movement points, and this multiplicative-reciprocal information does not add much.
4. One can easily get the hex count wrong, when there are more than four or five hexes.

Please just drop the colored hexes and bring back the old layout of this column.

_Maybe_ keep the coloring of the number, although I don't like that much either.

Pentarctagon:
This was discussed pretty extensively before merging #5264

Wedge009:
> This is obviously redundant. But - you don't need me to tell you that.

The tone of many of your reports - this included - aren't exactly conducive to co-operative and constructive discussion. While in once sense it could be considered redundant, it does provide a visual or 'graphical' indication of the relative movement points on different terrain. While experienced players may not find this useful I'm sure relative newcomers may benefit from this representation.

I admit I haven't looked at this screen in detail since the change was merged. I'll grant the saturated colours for the hex icons probably could do with some toning down. As for the rest of your objections, well, I've had my own remarks both here and in other projects be dismissed out of hand despite being less subjective and more respectfully phrased.

I don't think your points are strong enough to warrant a removal of this change, however 'redundant' it might seem to you. On the point of the saturated colouring, I think this is perhaps the only thing that warrants this issue remaining open. However, I am similarly just expressing my opinion and I am willing to be overruled.

Wedge009:
For reference.

1.16.6 release:
![image](https://user-images.githubusercontent.com/13679322/191890798-4fb01f6a-b7f7-4f34-b344-5690f2696cb3.png)

1.17.8+dev (d01aabc81fb):
[...]

PR #7018, opened 2022-09-23
Reduce the saturation of hex colour for movement points

Wedge009:
Attempting to mitigate #7017.

It's a subtle difference, but I think it may help ease the complaints of 'garishness' regarding the recent addition to the help page. (Also applied to the side bar reports, etc.)

Before:
191890877-b944fe4f-0447-4323-9490-87b1624ed97d

After:
image


stevecotton (approved):
Visually it looks slightly better to me, and the code looks good.

Vultraz (approved):

PR #7021, opened 2022-09-24
change the deprecation message content for backstab option

newfrenchy83:


src/units/abilities.cpp
```diff
@@ -1784,7 +1784,7 @@ effect::effect(const unit_ability_list& list, int def, bool backstab, const_atta
 		const std::string& effect_id = cfg[cfg["id"].empty() ? "name" : "id"];
 
 		if (!cfg["backstab"].blank()) {
-			deprecated_message("backstab= in weapon specials", DEP_LEVEL::PREEMPTIVE, {1, 15, 0}, "Use [filter_adjacent] instead.");
+			deprecated_message("backstab= in weapon specials", DEP_LEVEL::PREEMPTIVE, {1, 15, 0}, "Use [filter_opponent] with formula instead. The code can be find in data\core\macros\abilities.cfg in WEAPON_SPECIAL_BACKSTAB macro");
```
CelticMinstrel, line 1787: "can be found"

Pentarctagon:
You'll want to use forward slashes in the path name rather than back slashes.

PR #7022, opened 2022-09-25
Timestamping server messages [1.16 backport]

Discontinuum:
Backport #6979

### DIFF
--- a/src/server/wesnothd/game.cpp
+++ b/src/server/wesnothd/game.cpp
@@ -1950,6 +1950,10 @@ void game::send_server_message(const char* message, std::optional<player_iterato
 
 		msg.set_attr("id", "server");
 		msg.set_attr_dup("message", message);
+		std::stringstream ss;
+		ss << ::std::time(nullptr);
+		msg.set_attr_dup("time", ss.str().c_str());
+
 	} else {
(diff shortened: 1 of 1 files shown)


Wedge009:
Seems reasonable to me. Just wondering why you didn't cherry-pick the original commit, though - looks like you've added an extra line compared with what was added to master.

Discontinuum:
Yeah, I didn't notice the extra line, my bad. I can fix it if that matters. 

> why you didn't cherry-pick the original commit
I tried to rebase the original commit onto 1.16 branch but it failed, maybe because I fetch branches from the Wesnoth repo with 1 commit depth, I'm too lazy to download the full 3GB repo history

Wedge009:
I don't think it matters too much, but it's a divergence from master that is probably best avoided. I normally cherry-pick in GUI, but I think it should be a rather quick operation, shouldn't need to download branches for it. If there are no objections to this PR, I could just cherry-pick it for you.

Wedge009:
Ah, I just noticed soliton's affirmation regarding back-porting, plus Pentarctagon assented in #wesnoth-dev too. Have back-ported in b5ce5010881d87e44701f69e86cbc6f3531d456e.

PR #7025, opened 2022-09-26
Remove "Apply to all units" option from the pick_advance mod

Discontinuum:
So, for starters I propose this change, completely remove "Apply to all units" option from the Plan Unit Advancements mod. It works in a buggy and confusing way: it enables the chosen advancement for all subsequent recruits and doesn't apply to the unit by clicking on which it was requested. There were no suggestions how to improve it and the more I thought about it, I came to a conclusion that it will be confusing either way. Some people think it sets for all units on the board, others that on all subsequent units on the board including the one you clicked on. So I propose for now just delete this option, people may set up advancements for each unit individually (although it may be tedious for very large scenario prob?) I'm not committed to this resolution and if any other good ideas on how to improve it appear, I will gladly double down on them.

As this mod is inlined in each game that uses it, afaik this change shouldn't cause any OOS and so could be merged right in 1.16

### DIFF
--- a/data/modifications/pick_advance/dialog.lua
+++ b/data/modifications/pick_advance/dialog.lua
@@ -17,8 +17,6 @@ function pickadvance.show_dialog_unsynchronized(advance_info, unit)
 
 	local unit_override_one = (advance_info.unit_override or {})[2] == nil
 		and (advance_info.unit_override or {})[1] or nil
-	local game_override_one = (advance_info.game_override or {})[2] == nil
-		and (advance_info.game_override or {})[1] or nil
 
 	local description_row = T.row {
 		T.column {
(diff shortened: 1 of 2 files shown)


stevecotton:
> As this mod is inlined in each game that uses it, afaik this change shouldn't cause any OOS and so could be merged right in 1.16

AFAIK, at the start of each scenario the game takes the list of names of enabled mods, and inlines the current version of that mod. So it's necessary to check whether the new code works with leftover variables from a previous scenario (and enables resetting values that are no longer available in the UI).

Note - I haven't tested this PR, and I haven't looked into the issue it's resolving, so it may already work correctly.

Discontinuum:
I think I've got an alternative idea about this, without deleting the feature, I'll try to prepare another pull request with the new idea and then we can discuss what to do

Pentarctagon:
Should this be closed then?

Discontinuum:
okay, I will reopen if I decide to return to this idea

PR #7026, opened 2022-09-27
Units: Wyvern Rider XP set to 200

knyghtmare:
This is a level 4 unit and so, by convention, `experience=` should be `200`.


knyghtmare:
Merging in 48 hours unless any arguments

knyghtmare:
and it's 48 hours

PR #7031, opened 2022-10-01
Remove deprecated backstab code and using wml filters instead

newfrenchy83:
the old code of deprecated backstab is removedin this PR and wml code used intead pushed in backstab attribute for it's work

### DIFF
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -82,8 +82,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 		bool attacking,
 		nonempty_unit_const_ptr oppp,
 		const map_location& opp_loc,
-		const_attack_ptr opp_weapon,
-		const unit_map& units)
+		const_attack_ptr opp_weapon)
 	: weapon(nullptr)
 	, attack_num(u_attack_num)
 	, is_attacker(attacking)
@@ -94,7 +93,6 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	, petrifies(false)
 	, plagues(false)
 	, poisons(false)
-	, backstab_pos(false)
 	, swarm(false)
 	, firststrike(false)
 	, disable(false)
@@ -149,7 +147,6 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	drains = !opp.get_state("undrainable") && weapon->has_special_or_ability("drains");
 	petrifies = !opp.get_state("unpetrifiable") && weapon->has_special_or_ability("petrifies");
 	poisons = !opp.get_state("unpoisonable") && weapon->has_special_or_ability("poison") && !opp.get_state(unit::STATE_POISONED);
-	backstab_pos = is_attacker && backstab_check(u_loc, opp_loc, units, resources::gameboard->teams());
 	rounds = weapon->get_specials_and_abilities("berserk").highest("value", 1).first;
 
 	firststrike = weapon->has_special_or_ability("firststrike");
@@ -180,7 +177,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	cth = std::clamp(cth, 0, 100);
 
 	unit_ability_list cth_specials = weapon->get_specials_and_abilities("chance_to_hit");
-	unit_abilities::effect cth_effects(cth_specials, cth, backstab_pos, weapon);
+	unit_abilities::effect cth_effects(cth_specials, cth, weapon);
 	cth = cth_effects.get_composite_value();
 
 
@@ -191,7 +188,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	chance_to_hit = std::clamp(cth, 0, 100);
 
 	// Compute base damage done with the weapon.
-	int base_damage = weapon->modified_damage(backstab_pos);
+	int base_damage = weapon->modified_damage();
 
 	// Get the damage multiplier applied to the base damage of the weapon.
 	int damage_multiplier = 100;
@@ -221,19 +218,19 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	if(drains) {
 		unit_ability_list drain_specials = weapon->get_specials_and_abilities("drains");
 		// Compute the drain percent (with 50% as the base for backward compatibility)
-		unit_abilities::effect drain_percent_effects(drain_specials, 50, backstab_pos, weapon);
+		unit_abilities::effect drain_percent_effects(drain_specials, 50, weapon);
 		drain_percent = drain_percent_effects.get_composite_value();
 	}
 
 	// Add heal_on_hit (the drain constant)
 	unit_ability_list heal_on_hit_specials = weapon->get_specials_and_abilities("heal_on_hit");
-	unit_abilities::effect heal_on_hit_effects(heal_on_hit_specials, 0, backstab_pos, weapon);
+	unit_abilities::effect heal_on_hit_effects(heal_on_hit_specials, 0, weapon);
 	drain_constant += heal_on_hit_effects.get_composite_value();
 
 	drains = drain_constant || drain_percent;
 
 	// Compute the number of blows and handle swarm.
-	weapon->modified_attacks(backstab_pos, swarm_min, swarm_max);
+	weapon->modified_attacks(swarm_min, swarm_max);
 	swarm = swarm_min != swarm_max;
 	num_blows = calc_blows(hp);
 }
@@ -255,7 +252,6 @@ battle_context_unit_stats::battle_context_unit_stats(const unit_type* u_type,
 	, petrifies(false)
 	, plagues(false)
 	, poisons(false)
-	, backstab_pos(false)
 	, swarm(false)
 	, firststrike(false)
 	, disable(false)
@@ -326,12 +322,12 @@ battle_context_unit_stats::battle_context_unit_stats(const unit_type* u_type,
 	cth = std::clamp(cth, 0, 100);
 
 	unit_ability_list cth_specials = weapon->get_specials("chance_to_hit");
-	unit_abilities::effect cth_effects(cth_specials, cth, backstab_pos, weapon);
+	unit_abilities::effect cth_effects(cth_specials, cth, weapon);
 	cth = cth_effects.get_composite_value();
 
 	chance_to_hit = std::clamp(cth, 0, 100);
 
-	int base_damage = weapon->modified_damage(backstab_pos);
+	int base_damage = weapon->modified_damage();
 	int damage_multiplier = 100;
 	damage_multiplier
 			+= generic_combat_modifier(lawful_bonus, u_type->alignment(), u_type->musthave_status("fearless"), 0);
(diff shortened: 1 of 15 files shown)


CelticMinstrel:
The third commit is wrong. Making the deprecation level INDEFINITE is only reasonable if the feature still functions. The only way to do this would be to move the deprecation message to a place where the ability is not const and modify the ability on the fly. If you're not going to do that, then please just delete the third commit.

newfrenchy83:
commit deleted, but i don't understand where i can move deprecation message

CelticMinstrel:
I'm not sure either.

newfrenchy83:
> The third commit is wrong. Making the deprecation level INDEFINITE is only reasonable if the feature still functions. The only way to do this would be to move the deprecation message to a place where the ability is not const and modify the ability on the fly. If you're not going to do that, then please just delete the third commit.

it's almost done.

newfrenchy83:
> The third commit is wrong. Making the deprecation level INDEFINITE is only reasonable if the feature still functions. The only way to do this would be to move the deprecation message to a place where the ability is not const and modify the ability on the fly. If you're not going to do that, then please just delete the third commit.

wml added, backstab continue to work after removing of old code.

src/actions/attack.cpp
```diff
@@ -221,19 +218,19 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	if(drains) {
 		unit_ability_list drain_specials = weapon->get_specials_and_abilities("drains");
 		// Compute the drain percent (with 50% as the base for backward compatibility)
-		unit_abilities::effect drain_percent_effects(drain_specials, 50, backstab_pos, weapon);
+		unit_abilities::effect drain_percent_effects(drain_specials, 50, false, weapon);
```
CelticMinstrel, line 221: This parameter could also be removed.

CelticMinstrel:
I see a call to a `backstab_check` function that has been removed, but the function itself was not removed.

src/actions/attack.cpp
newfrenchy83: done

newfrenchy83:
> backstab_check

backstab_check called in ai/default/aspect_attacks.cpp line 272. and i don't know if it can be removed.

CelticMinstrel:
> backstab_check called in ai/default/aspect_attacks.cpp line 272. and i don't know if it can be removed.

I took a quick look at it. It appears that that usage is still valid and will still function even though the `backstab` key is gone – the AI detects the backstab ability by ID.

newfrenchy83:
> > backstab_check called in ai/default/aspect_attacks.cpp line 272. and i don't know if it can be removed.
> 
> I took a quick look at it. It appears that that usage is still valid and will still function even though the `backstab` key is gone – the AI detects the backstab ability by ID.

i removed the last commit and let backstab_check

CelticMinstrel:
I think this looks good to merge, as long as it's been tested and works.

src/units/abilities.cpp
```diff
@@ -1784,11 +1784,9 @@ effect::effect(const unit_ability_list& list, int def, bool backstab, const_atta
 		const std::string& effect_id = cfg[cfg["id"].empty() ? "name" : "id"];
 
 		if (!cfg["backstab"].blank()) {
-			deprecated_message("backstab= in weapon specials", DEP_LEVEL::FOR_REMOVAL, {1, 19, 0}, "Use [filter_opponent] with a formula instead; the code can be found in data/core/macros/abilities.cfg in the WEAPON_SPECIAL_BACKSTAB macro.");
+			deprecated_message("backstab= in weapon specials", DEP_LEVEL::INDEFINITE, "", "Use [filter_opponent] with a formula instead; the code can be found in data/core/macros/abilities.cfg in the WEAPON_SPECIAL_BACKSTAB macro.");
```
stevecotton, line 1787: This line is only reached when the ability is active, if it's inactive then it's already been filtered out of the list before this function is called.

That's neither a review-failure nor a regression from the previous code, but it was a quirk when testing.

stevecotton (approved):
Thanks, this makes complex code much simpler.

I've run the WML unit tests, and confirm that they pass even when replacing the filter in the definition of `WEAPON_SPECIAL_BACKSTAB` with `backstab=yes`. It also correctly treats `backstab=no` to be the same as not having a filter.

newfrenchy83:
@CelticMinstrel the problem is when a [filter_opponent] is already present, i work for resolve that but else the PR can't be merged yet.

src/units/abilities.cpp
newfrenchy83: for me, the message must be showed when player whant fight with unit owner of backstab, and not each time what unit select. thiswhy i not moved message.

newfrenchy83:
> @CelticMinstrel the problem is when a [filter_opponent] is already present, i work for resolve that but else the PR can't be merged yet.

bug fixed bit i await what checking complete before to declare ready for review

src/units/abilities.cpp
stevecotton: The target audience should be UMC authors, rather than players. When WML needs to be updated, only the UMC author can do that update, which is why they're the target audience. Telling the players that it doesn't work is helpful to explain why things don't work as they expect, but the target should be to alert UMC authors when they're porting to new versions of Wesnoth. Ideally it would be shown to a UMC author who's doing a sanity-check of their era or their campaign, even if they're using debug to skip a lot of the combat.

The `deprecated_message` framework should already ensure that a player only sees it once per game.

newfrenchy83: ok, i will moves message for what it was showed immediatly

soliton-:
Excellent, this is exactly what I was hoping for. Even better if the deprecation message spells the replacement WML out in case it's moved around in mainline but this is great either way.

stevecotton:
> Even better if the deprecation message spells the replacement WML

I disagree on that - while UMC authors are the target audience, this message will sometimes get shown to players too. Having something that's relatively easy to write in a forum post makes it more likely to be reported, and more likely that players will be sure whether the most recent forum post for that add-on is talking about the same issue. So an error message that makes it clear that `WEAPON_SPECIAL_BACKSTAB` is the magic phrase to post is better than a longer message with WML code, IMO.

With the current verbosity, a UMC author should be able to find the macro, or at find least Wiki and forum posts about the macro.

stevecotton (approved):

soliton-:
The problem is that if the backstab code changes/moves there is a very small chance someone will update the deprecation message as well.

stevecotton:
The macro's name will never change, we'd break a lot of compatibility if it was renamed from `WEAPEN_SPECIAL_BACKSTAB`.

If the implementation of the macro changed, why would we want UMC authors to use a copy of the outdated code, instead of finding the current implementation and getting the latest code? I don't think it's going to change, but this doesn't seem a reason for a code copy.

The backstab code could move between files, so maybe the filename should be omitted?

newfrenchy83:
> The macro's name will never change, we'd break a lot of compatibility if it was renamed from `WEAPON_SPECIAL_BACKSTAB`.
> 
> If the implementation of the macro changed, why would we want UMC authors to use a copy of the outdated code, instead of finding the current implementation and getting the latest code? I don't think it's going to change, but this doesn't seem a reason for a code copy.
> 
> The backstab code could move between files, so maybe the filename should be omitted?

filename removed but not folder

soliton-:
This is about the backstab attribute not the macro. If some UMC uses the macro it won't get the deprecation message at all. So since the macro is not used there is a good chance the wanted behavior is not the same as mainline either.

The other thing is that the deprecation message should say what the compatibility code does. No reason to potentially let that diverge IMO.

newfrenchy83:
> This is about the backstab attribute not the macro. If some UMC uses the macro it won't get the deprecation message at all. So since the macro is not used there is a good chance the wanted behavior is not the same as mainline either.
> 
> The other thing is that the deprecation message should say what the compatibility code does. No reason to potentially let that diverge IMO.

No, no and no! the message is for explain what attribute is deprecated and wherre find the code in replacement. I already removed the filename and now i must explain how work the compatibility code? Sorry if i feel angry, but i would what this PR was merged as soon as possible, and i had 3 oher PR who await.

src/units/abilities.cpp
```diff
@@ -1667,10 +1658,29 @@ bool attack_type::special_active_impl(
...
+	if(special["backstab"].to_bool()){
+		config filter;
+		const std::string& backstab_formula = "	enemy_of(self, flanker) and not flanker.petrified	where	flanker = unit_at(direction_from(loc, other.facing))	";
+		filter["formula"] = backstab_formula;
+		if(!special.child("filter_opponent")){
+			cfg.add_child("filter_opponent", filter);
+		} else {
+			config filter_child = cfg.child("filter_opponent");
+			filter_child.add_child("and", filter);
+			cfg.child("filter_opponent") = filter_child;
```
CelticMinstrel, line 1675: I think instead of having this line assigning the config back, you can just add an & a few lines up:

```suggestion
			config& filter_child = cfg.child("filter_opponent");
			filter_child.add_child("and", filter);
```

Pentarctagon:
A deprecation message isn't the place to explain much of *anything* - it's to say what is deprecated, and if it's really so complex that it needs a detailed explanation then that explanation should be on the wiki.

src/units/abilities.cpp
stevecotton: Should this (both @newfrenchy83's code and the suggested replacement) use `child_or_add()` instead of `child()`?

soliton-:
> > This is about the backstab attribute not the macro. If some UMC uses the macro it won't get the deprecation message at all. So since the macro is not used there is a good chance the wanted behavior is not the same as mainline either.
> > The other thing is that the deprecation message should say what the compatibility code does. No reason to potentially let that diverge IMO.
> 
> No, no and no! the message is for explain what attribute is deprecated and wherre find the code in replacement. I already removed the filename and now i must explain how work the compatibility code? Sorry if i feel angry, but i would what this PR was merged as soon as possible, and i had 3 oher PR who await.

I already mentioned that this is great either way. I'm perfectly fine with this being merged as is. If people continue to want to bikeshed the deprecation message that is fine with me as well though. Even if it is a bit mind boggling to be honest. The replacement code is 3 lines of WML...

newfrenchy83:
@stevecotton  code simplified with reference and child_or_add()

CelticMinstrel:
> If people continue to want to bikeshed the deprecation message that is fine with me as well though.

I think the deprecation message can always be updated again if someone wants to. It doesn't have to be perfect in order to merge this.

src/units/abilities.cpp
```diff
@@ -1667,10 +1658,28 @@ bool attack_type::special_active_impl(
...
 
+	//Add wml filter if "backstab" attribute used.
+	if (!special["backstab"].blank()) {
+		deprecated_message("backstab= in weapon specials", DEP_LEVEL::INDEFINITE, "", "Use [filter_opponent] with a formula instead; the code can be found in data/core/macros/ in the WEAPON_SPECIAL_BACKSTAB macro.");
+	}
+	config cfg = special;
+	if(special["backstab"].to_bool()){
+		const std::string& backstab_formula = "	enemy_of(self, flanker) and not flanker.petrified	where	flanker = unit_at(direction_from(loc, other.facing))	";
+		config& filter_child = cfg.child_or_add("filter_opponent");
+		if(!special.child("filter_opponent")){
```
CelticMinstrel, line 1669:
```suggestion
		if(!special.has_child("filter_opponent")){
```

CelticMinstrel:
Is there good reason to keep this PR as separate commits, or should we squash the whole thing when merging?

stevecotton (approved):
Looks good to me. I'm assuming the PR should be squashed when merging.

src/units/abilities.cpp
```diff
@@ -1667,10 +1658,28 @@ bool attack_type::special_active_impl(
 			return false;
 	}
 
+	//Add wml filter if "backstab" attribute used.
+	if (!special["backstab"].blank()) {
+		deprecated_message("backstab= in weapon specials", DEP_LEVEL::INDEFINITE, "", "Use [filter_opponent] with a formula instead; the code can be found in data/core/macros/ in the WEAPON_SPECIAL_BACKSTAB macro.");
+	}
+	config cfg = special;
+	if(special["backstab"].to_bool()){
+		const std::string& backstab_formula = "	enemy_of(self, flanker) and not flanker.petrified	where	flanker = unit_at(direction_from(loc, other.facing))	";
```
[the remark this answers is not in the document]

PR #7033, opened 2022-10-02
Unit tests for changing filters in [event]name=attack

stevecotton:
I'm using the existing backstab test as a framework for the new tests, but this means they have to be in the same file as the backstab test.

@Pentarctagon, this is probably a good time to discuss your plans for reorganising the tests - where would the common code of the `TEST_BACKSTAB` macro go?

### DIFF
--- a/data/test/macros/generic_backstab_test.cfg
+++ b/data/test/macros/generic_backstab_test.cfg
@@ -1,0 +1,102 @@
+# wmllint: no translatables
+
+#define GENERIC_BACKSTAB_TEST NAME ANTIDIRECTIONS ABILITY_SHOULD_ACTIVATE EXTRA_SETUP
+    # Test that the backstab ability does double-damage if and only if it should
+    # be active in the circumstances created by EXTRA_SETUP.
+    #
+    # Given a set of directions, create units around Bob.  One of those units
+    # makes an attack, and the damage dealt is compared to the expectation that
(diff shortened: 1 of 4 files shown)


Pentarctagon:
Oh yeah, I really do need to get back to that don't I...

The idea was to put every test in its own file which would be named the same as the test it contained, and common code would be put in either a `data/test/macros/` or `data/test/utils/` folder.

stevecotton:
I think there's no rush, it's just something to include while doing related changes.

stevecotton:
I've moved all the backstab tests to separate files, so the code in the Git commits is arranged as I understand your plans to be. 

However, I feel it would be better to structure it this way instead:

* `GENERIC_BACKSTAB_TEST` in a separate file
* All four of the existing tests together in a single file, because when they're only 1-10 lines of code it's simpler to scroll up and down the complete file rather than swap between files.
* The two new tests along with `WEAPON_SPECIAL_BACKSTAB_BY_STATUE` together in a single file. Again they're tiny amounts of code, so it's easier to scroll up and down. The weapon itself is unlikely to be used in other tests.

For tests that need a lot of code, I'd agree with putting them in individual files. The borderline being the subjective question of whether I'd prefer to compare two tests side-by-side or not.

Pentarctagon:
That sounds fine to me.

stevecotton:
The build failure was #7037 again.

PR #7036, opened 2022-10-03
run_wml_tests: add an option to filter tests by name

stevecotton:
For example, `run_wml_tests --filter backstab` runs only the tests listed in `wml_test_schedule` that have `backstab` somewhere in their name.

I haven't tried to match the Boost test framework's [syntax for filters](https://www.boost.org/doc/libs/1_80_0/libs/test/doc/html/boost_test/utf_reference/rt_param_reference/run_test.html), because they expect tests to be organised into suites.

GTest's `--gtest_filter` agrees on the name and using regexps. That framework also allows negative patterns and multiple regexps separated by colons, but I'm not spending time adding that, nor does it seem necessary for our purposes.

### DIFF
--- a/run_wml_tests
+++ b/run_wml_tests
@@ -100,6 +100,10 @@ class TestListParser:
     def __init__(self, options):
         self.verbose = options.verbose
         self.filename = options.list
+        if options.filter is not None:
+            self.filter = re.compile(options.filter)
+        else:
+            self.filter = None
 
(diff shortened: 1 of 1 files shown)


stevecotton:
The build failure is unrelated - it's a timeout during the Boost tests, logged as #7037. I'll rerun that build to confirm that it's a timing issue.

PR #7039, opened 2022-10-04
Add [filter_weapon][not]special_id_active=forced_cth in chance_to_hit abilities using overwrite_specials attribute

newfrenchy83:
The FORCE_CHANCE_TO_HIT macrodon't overwrite a cth who using overwrite_specials attribute if [filter_weapon][not]special_id_active=forced_cth not addedin [filter_student/opponent]. i propose to add that in hard code for resolve this issue.

### DIFF
--- a/data/test/scenarios/test_force_chance_to_hit_macro.cfg
+++ b/data/test/scenarios/test_force_chance_to_hit_macro.cfg
@@ -37,6 +37,15 @@
                     [/chance_to_hit]
                 [/set_specials]
             [/effect]
+            [effect]
+                apply_to=new_ability
+                [abilities]
+                    [chance_to_hit]
+                        value=100
+                        overwrite_specials=one_side
+                    [/chance_to_hit]
+                [/abilities]
+            [/effect]
             [filter]
                 id=bob
             [/filter]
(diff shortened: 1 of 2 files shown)


stevecotton:
I not sure if I'm understanding the purpose correctly. Is this a correct understanding?
* problem: per-weapon abilities are always overriding per-unit abilities
* solution: add a way for one specific per-unit ability to override one specific per-weapon ability

newfrenchy83:
> I not sure if I'm understanding the purpose correctly. Is this a correct understanding?
> 
>     * problem: per-weapon abilities are always overriding per-unit abilities
> 
>     * solution: add a way for one specific per-unit ability to override one specific per-weapon ability

an weapon abilitie with overwrite_specials can overwrite ability/special of same type. it's ever case for specials butnot for abilities if the last have alos overwrite_specials attribute. And this by using of this attribute what FORCE_CHANCE_TO_HIT macro work. if a wml develloperusing a [chance_to_hit] ability with overwrite_specials(magical leadership by exemple), then the macro don't work to impose value<ability value, except if devellopper think to filter the ability used in FORCE_CHANCE_TO_HIT.

Like this macro must can be work in all situation, i proposethis PR for add filter in hard coding.

CelticMinstrel:
So this is a hard-coded special case in C++ to ensure that the FORCE_CHANCE_TO_HIT macro is never overshadowed by some other ability…

PR #7041, opened 2022-10-07
Add "all_others" option to overwrite_specials attribute for overwrite all specials of same type

newfrenchy83:
when overwrite_specials used with "all_others" all specials/abilities of same type wh don't have overwrite_specials=all_others will be overwrited.

Can resolve force_chance_to_hit macro can't overwrite chance_to_hit ability using overwrite_specials.

### DIFF
--- a/data/test/scenarios/test_force_chance_to_hit_macro.cfg
+++ b/data/test/scenarios/test_force_chance_to_hit_macro.cfg
@@ -37,6 +37,15 @@
                     [/chance_to_hit]
                 [/set_specials]
             [/effect]
+            [effect]
+                apply_to=new_ability
+                [abilities]
+                    [chance_to_hit]
+                        value=100
+                        overwrite_specials=one_side
+                    [/chance_to_hit]
+                [/abilities]
+            [/effect]
             [filter]
                 id=bob
             [/filter]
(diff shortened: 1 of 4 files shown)


newfrenchy83:
@CelticMinstrel can you review this PR instead  of https://github.com/wesnoth/wesnoth/pull/7039?

data/test/scenarios/test_force_chance_to_hit_macro.cfg
```diff
@@ -37,6 +37,15 @@
                     [/chance_to_hit]
```
stevecotton, line 37: This `[object]` now has two different `[chance_to_hit]` blocks. IIUC, a comment saying "check that the macro overrides both abilities and weapon-specials" would help.

The old documentation at the top of the file talks about the `unupgradable` trick, but that doesn't seem to match anything in this test.

stevecotton:
I'm worried that it's adding additional complexity to handle an issue caused by previously-added complexity. It might be the only way to handle this, but it doesn't feel good.

One thing that definitely needs to change - if `FORCE_CHANCE_TO_HIT` is used multiple times, then the last one added needs to be the one that's used. That's probably used in UMC, for a boss who's temporarily invulnerable until the hero gets something that triggers the boss' weakness.

For example, adding the lines with a `+` breaks the test, and it mustn't.
```
 {GENERIC_UNIT_TEST "test_force_chance_to_hit_macro" (
+    # Looks like the overwrite_ultimate might be a first-one-wins, rather than last-one-wins
+    {FORCE_CHANCE_TO_HIT (id=bob) (id=alice) 100 ()}
+    {FORCE_CHANCE_TO_HIT (id=alice) (id=bob) 100 ()}
     {FORCE_CHANCE_TO_HIT (id=bob) (id=alice) 0 ()}
     {FORCE_CHANCE_TO_HIT (id=alice) (id=bob) 0 ()}
```

PR #7043, opened 2022-10-08
Fix FORCE_CHANCE_TO_HIT macro

newfrenchy83:
with this fix, FORCE_CHANCE_TO_HIT macro work when unit affected using a cth ability with overwrite_specials.(But not if opponent using a cth ability with overwrite_specials and apply_to=opponent).

### DIFF
--- a/data/core/macros/utils.cfg
+++ b/data/core/macros/utils.cfg
@@ -536,7 +536,6 @@
         [/filter_condition]
 
         [if]
-            #This is to mute a warning message about retrieving a member of non-existant wml container.
             [variable]
                 name=unit.abilities.length
                 greater_than=0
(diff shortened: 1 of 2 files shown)


newfrenchy83:
@stevecotton can you review this PR please?

CelticMinstrel:
I think this method of "solving" it creates as many problems as it solves. Why not just stick to the original `[object]`, but add an extra `[effect]` that removes all `[chance_to_hit]` abilities? I'm pretty sure effects are applied in order, so that _shouldn't_ be contradictory.

That does mean extending `apply_to=remove_ability` to allow more sophisticated ability removal – currently it only requires the ID to match. I'm thinking using `config::matches` instead would be almost sufficient, though we'd probably also want a way to match an ability without knowing its tag.

PR #7046, opened 2022-10-10
Simplify code of overwrite_specials attribute for more readibility.

newfrenchy83:


src/units/abilities.cpp
```diff
@@ -1297,46 +1297,36 @@ static bool overwrite_special_affects(const config& special)
 	return (apply_to == "one_side" || apply_to == "both_sides");
 }
 
-unit_ability_list attack_type::overwrite_special_checking(const std::string& ability, unit_ability_list temp_list, const unit_ability_list& abil_list, const std::string& filter_self) const
+unit_ability_list attack_type::overwrite_special_checking(const std::string& ability, const unit_ability_list& temp_list, const unit_ability_list& abil_list, const std::string& filter_self, bool is_special) const
 {
 	bool overwrite_self = false;
 	bool overwrite_either = false;
-
+	unit_ability_list overwrite_list;
```
stevecotton, line 1304: As this is only used in the `else` block, it can be declared later in the function.

src/units/abilities.cpp
```diff
@@ -1297,46 +1297,36 @@ static bool overwrite_special_affects(const config& special)
...
-				i = temp_list.erase(i);
-			} else {
-				++i;
+			if(overwrite) {
+				overwrite_list.emplace_back(i);
 			}
 		}
+		return overwrite_list;
 	}
 	return temp_list;
```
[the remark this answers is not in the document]

src/units/abilities.cpp
newfrenchy83: Done

src/units/attack_type.hpp
```diff
@@ -132,8 +132,9 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 	 * @param temp_list the list checked and returned.
 	 * @param abil_list list checked for verify presence of overwrite_specials .
 	 * @param filter_self name of [filter_"self/student"] if is abilities or specials who are checked
+	 * @param is_special for determine if list is a special or a ability.
 	 */
-	unit_ability_list overwrite_special_checking(const std::string& ability, unit_ability_list temp_list, const unit_ability_list& abil_list, const std::string& filter_self = "filter_self") const;
+	unit_ability_list overwrite_special_checking(const std::string& ability, const unit_ability_list& temp_list, const unit_ability_list& abil_list, const std::string& filter_self = "filter_student", bool is_special = false) const;
```
stevecotton, line 137: Instead of changing a default value from `filter_self` to `filter_student`, please just remove the default so that all callers have to provide the value. Without this, someone comparing code between 1.16 and 1.17 may incorrectly assume that `overwrite_special_checking(special, spe_list, abil_list);` has the same behavior in both branches; that's a gotcha that's easy to avoid.

src/units/abilities.cpp
```diff
@@ -1307,38 +1307,29 @@ unit_ability_list attack_type::overwrite_special_checking(const std::string& abi
...
+	}
+
+	// At this point we need to return a changed list, so create a non-const one to return
+	unit_ability_list overwrite_list;
+	for(const auto& i : temp_list) {
+		bool overwrite = false;
+		if(!is_special && overwrite_either){
+			overwrite = overwrite_special_affects(*i.ability_cfg);
+		} else if(overwrite_self){
+			bool temp_is_ability = !is_special ? overwrite_special_affects(*i.ability_cfg) : false;
```
stevecotton, line 1325: I think that ternary simplifies to just `(!is_special) && overwrite_specials_affects(*i.ability_cfg)`.

src/units/attack_type.hpp
newfrenchy83: it's done

src/units/abilities.cpp
newfrenchy83: done here also

src/units/abilities.cpp
```diff
@@ -1271,18 +1271,18 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
 	}
 
 	abil_list.append(abil_other_list);
-	if(!abil_list.empty()){
-		abil_list = overwrite_special_checking(ability, abil_list, abil_list, "filter_student");
-	}
```
stevecotton, line 1276: Just wondering if moving this line could add a bug in udisplay.cpp's `unit_attack` function.

newfrenchy83: if a leadershiptype abilitiy is overwrited, the animation willbe played then what not actually but i will cancel this change

stevecotton: Should I review this PR now, or wait for another commit to be pushed to it?

newfrenchy83: Now

stevecotton:
I'm confused about the force-push removing the latest commit - is there still something left to fix?

src/units/abilities.cpp
```diff
@@ -1297,48 +1297,48 @@ static bool overwrite_special_affects(const config& special)
 	return (apply_to == "one_side" || apply_to == "both_sides");
 }
 
-unit_ability_list attack_type::overwrite_special_checking(const std::string& ability, unit_ability_list temp_list, const unit_ability_list& abil_list, const std::string& filter_self) const
+unit_ability_list attack_type::overwrite_special_checking(const std::string& ability, const unit_ability_list& temp_list, const unit_ability_list& abil_list, const std::string& filter_self, bool is_special) const
 {
 	bool overwrite_self = false;
+	bool overwrite_oppponent = false;
```
stevecotton, line 1303: Just noticed: there should only be two `p`s in `opponent`.

stevecotton:
Thanks for the simplification, my feeling is that it's ready to merge (just waiting so that you have time to correct the `oppponent`).

At the moment I'm haven't got a full understanding of the edge-cases, but it will be easier to understand them with the new code.

stevecotton (approved):

src/units/abilities.cpp

stevecotton: Sorry, you missed a couple. `src/units/abilities.cpp:1314:187: error: ‘overwrite_oppponent’`, and another on line 1315.

newfrenchy83: fixed

Issue #1063, opened 2017-05-08
Sorting of unofficial campaigns by language (GNA #5327)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?5327)** by gwain on 2006-02-17

it should be allowed to sort the campaigns by language, checking if there is a translation for the language selected in the options menu, or by the default menu

(Reproduced on All)
Release: 1.0.2+
Priority: 5 - Normal
Severity: 1 - Wish

gfgtdf:
I assume this is a request about the addon manager and not the campaign selection screen?

jostephd:
This should be pretty straightforward. The addon manager is implemented in src/gui/dialogs/addon/manager.cpp and src/gui/dialogs/addon/manager.cpp; it already shows the list of translations, all that's needed is to add some UI to match against them.

CelticMinstrel:
Probably in the filter UI?

jostephd:
Yeah, the value of the search textbox could be matched against the text in the "Translations" label. Or maybe a dropdown could be added. A dropdown would have the advantage that a player could filter for several languages at once ("Campaigns with either Danish or Swedish translation").

CelticMinstrel:
Yeah, filtering for multiple languages is probably a good idea, and even if you didn't want to support that, I think a dropdown is better. Actually... maybe this is a place where a combobox would be the best choice - basically a cross between a text field and a combobox, so they can either pick a language from the menu or type in a language code in the text field.

Kingofd:
Hello,
I am interested in adding this feature. 
I am thinking of adding a dropdown, similar to the "Type"-dropdown. 

"None"-translations will default to English translation.

As I am new could you give me some pointers as to where stuff is added to the UI?

soliton-:
Did you take a look at `src/gui/dialogs/addon/manager.cpp` as mentioned by jostephd above? Anything specific you're missing? The related config is in `data/gui/window/addon_manager.cfg` in case you haven't found it.

Kingofd:
@soliton- Thx, addon_manager.cfg was what I was missing

So the feature is done, however I need to iron out some bugs and unwanted behaviour:

Firstly: When I select multiple languages, it can happen, that the list, where the addons are shown does not expand to fit the newly selected addons that fit the language filter and it throws the following error:
[...]

PR #7048, opened 2022-10-12
Adding language filter option in addon menu

Kingofd:
My shot at [https://github.com/wesnoth/wesnoth/issues/1063](https://github.com/wesnoth/wesnoth/issues/1063)

Also merged some changes from master since I was working on it first a while ago

Looks like this now: 

FeaturePicture


Pentarctagon:
From the CI: looks like a windows build issue, an indentation issue, and and existing wmlindent error that for some reason doesn't cause a failure. Not sure what to make of that last one.

Kingofd:
I can't seem to see which unit test failed for the MacOS debug test. Can someone tell me where to find that info?

src/gui/dialogs/addon/manager.cpp
```diff
@@ -183,6 +184,10 @@ const std::vector<std::pair<ADDON_TYPE, std::string>> addon_manager::type_filter
 	{ADDON_UNKNOWN,        N_("addons_of_type^Unknown")},
 };
 
+std::vector<std::pair<int, std::string>> addon_manager::language_filter_types_{
+	{0,	"en_GB"},
```
Vultraz, line 188: Remove this. This should not be a static member. See comment with declaration.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +396,50 @@ void addon_manager::pre_show(window& window)
 	connect_signal_notify_modified(type_filter,
 		std::bind(&addon_manager::apply_filters, this));
 
+	// Language filter
+	// Clear current language filter vector, lest all entries are shown twice
+	addon_manager::language_filter_types_.clear();
```
Vultraz, line 401: If this is a regular class member, no need to clear it in pre-show.

src/gui/dialogs/addon/manager.hpp
```diff
@@ -103,6 +103,7 @@ class addon_manager : public modal_dialog
 
 	static const std::vector<std::pair<ADDON_STATUS_FILTER, std::string>> status_filter_types_;
 	static const std::vector<std::pair<ADDON_TYPE, std::string>> type_filter_types_;
+	static std::vector<std::pair<int, std::string>> language_filter_types_;
```
Vultraz, line 106: Since this is filled based on the languages of available addons, this should just be a regular class member and not static.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +396,50 @@ void addon_manager::pre_show(window& window)
...
+			if (!(b.rfind("en_GB", 0) == 0)) {
+				languages_available.push_back(b);
+			}
+		}
+	}
+	std::sort(languages_available.begin(), languages_available.end());
+	auto last = std::unique(languages_available.begin(),languages_available.end());
+	languages_available.erase(last, languages_available.end());
+	// Erase pt value, since otherwise Portugues do Brasil will be shown twice in the language list
+	(void)std::remove(languages_available.begin(), languages_available.end(), "pt");
```
Vultraz, line 417: We don't need `(void)` here. `std::remove` is not marked `[[nodiscard]]`.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +396,50 @@ void addon_manager::pre_show(window& window)
...
+	// Language filter
+	// Clear current language filter vector, lest all entries are shown twice
+	addon_manager::language_filter_types_.clear();
+	addon_manager::language_filter_types_.push_back(std::pair<int, std::string>(0, "en_GB"));
+
+	// Prepare shown languages, source all available languages from the addons themselves
+	std::vector<std::string> languages_available;
+	for(const auto& a : addons_) {
+		for (const auto& b : a.second.locales) {
+			if (!(b.rfind("en_GB", 0) == 0)) {
```
Vultraz, line 408: `en_GB` should not be hardcoded. If anything, it should be your current game language selection.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +396,50 @@ void addon_manager::pre_show(window& window)
...
+			}
+		}
+	}
+	std::sort(languages_available.begin(), languages_available.end());
+	auto last = std::unique(languages_available.begin(),languages_available.end());
+	languages_available.erase(last, languages_available.end());
+	// Erase pt value, since otherwise Portugues do Brasil will be shown twice in the language list
+	(void)std::remove(languages_available.begin(), languages_available.end(), "pt");
+
+	language_filter_types_[0].second = langcode_to_string(language_filter_types_[0].second) != "" ? langcode_to_string(language_filter_types_[0].second) : "English (GB)";
```
Vultraz, line 419: Again, `en_GB` should not be hardcoded. Remove this bit.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +396,50 @@ void addon_manager::pre_show(window& window)
...
+	for (long unsigned int i = 0; i < languages_available.size(); i++) {
+		std::string myLangCodeString = langcode_to_string(languages_available[i]);
+		// Only show languages, which have a translation as per langcode_to_string() method
+		// Do not show tranlations with their langcode e.g. "sv_SV"
+		if (myLangCodeString != "") {
+			language_filter_types_.push_back(std::pair<int, std::string>(i, langcode_to_string(languages_available[i])));
```
Vultraz, line 425: Please use `emplace_back`.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -677,7 +726,44 @@ boost::dynamic_bitset<> addon_manager::get_type_filter_visibility() const
...
+			// langcode -> string conversion vector, to be able to detect either
+			// langcodes or langstring entries
+			std::vector<std::string> lang_string_vector;
+			for (long unsigned int i = 0; i < a.second.locales.size(); i++) {
+				lang_string_vector.push_back(langcode_to_string(a.second.locales[i]));
+			}
+			// Find all toggle states, where toggle = true and lang = lang
+			for (long unsigned int i = 0; i < toggle_states.size(); i++) {
+				if (toggle_states[i] == true) {
+					if ((std::find(a.second.locales.begin(), a.second.locales.end(), language_filter_types_[i].second) != a.second.locales.end() ? true : false) || (std::find(lang_string_vector.begin(), lang_string_vector.end(), language_filter_types_[i].second) != lang_string_vector.end() ? true : false)) {
```
Vultraz, line 760: Format this on more than one line.

Vultraz (changes requested):
This is a good start, but needs a couple changes. Most notably, we shouldn't have a hardcoded language filter option.

src/gui/dialogs/addon/manager.cpp
stevecotton: Looks like it could be simplified with `utils::contains`, which just needs src/utils/general.hpp to be `#included`.

src/gui/dialogs/addon/manager.cpp
Kingofd: :+1: 
I initially hardcoded it, so in the language filter method it would be easier to default to English, if there are no translations listed in the addon itself. In the new commit I now remember the position of en_GB with an integer

src/gui/dialogs/addon/manager.cpp
Kingofd: :ok_hand:

src/gui/dialogs/addon/manager.hpp
Kingofd: :ok_hand:

src/gui/dialogs/addon/manager.cpp
Kingofd: The two windows CI tests on the very first iteration I submitted failed because of that error, thats why I added it

src/gui/dialogs/addon/manager.cpp
Kingofd: :ok_hand: 
I now changed it to no particular ordering now en_GB hardcode was for programming convenience in get_lang_filter_visibility(), see other discussion

src/gui/dialogs/addon/manager.cpp
Kingofd: :ok_hand:

src/gui/dialogs/addon/manager.cpp
Kingofd: :ok_hand:

src/gui/dialogs/addon/manager.cpp
Kingofd: utils::contains was a great tip, reduced reading complexity of the code greatly :ok_hand:

src/gui/dialogs/addon/manager.cpp
CelticMinstrel: I think the issue here is that `remove` doesn't resize the list, so even if it's not marked `[[nodiscard]]` it probably should be. The idiomatic usage of `remove` is like this:

```cpp
container.erase(std::remove(container.begin(), container.end(), value), container.end())
```

So, following that is probably the correct fix for the CI error. (You could also split it into two lines by assigning the result of `remove` to a variable, if you wanted to.)

Kingofd: Thx for the input, it should now be properly handled :ok_hand:

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +391,46 @@ void addon_manager::pre_show(window& window)
...
+		for (const auto& b : a.second.locales) {
+			languages_available.push_back(b);
+		}
+	}
+	std::sort(languages_available.begin(), languages_available.end());
+	auto last = std::unique(languages_available.begin(),languages_available.end());
+	languages_available.erase(last, languages_available.end());
+
+	// Erase pt value, since otherwise Portugues do Brasil will be shown twice in the language list
+	languages_available.erase(std::remove(languages_available.begin(), languages_available.end(), "pt"), languages_available.end());
```
Pentarctagon, line 407: Is removing this correct? There is both a "pt" and a "pt_BR" language code.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +391,46 @@ void addon_manager::pre_show(window& window)
...
+	// Prepare shown languages, source all available languages from the addons themselves
+	std::vector<std::string> languages_available;
+	for(const auto& a : addons_) {
+		for (const auto& b : a.second.locales) {
+			languages_available.push_back(b);
+		}
+	}
+	std::sort(languages_available.begin(), languages_available.end());
+	auto last = std::unique(languages_available.begin(),languages_available.end());
+	languages_available.erase(last, languages_available.end());
```
Pentarctagon, line 404: Why are elements being erased here?

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +391,46 @@ void addon_manager::pre_show(window& window)
...
+
+	for (long unsigned int i = 0; i < languages_available.size(); i++) {
+		std::string myLangCodeString = langcode_to_string(languages_available[i]);
+		// Only show languages, which have a translation as per langcode_to_string() method
+		// Do not show tranlations with their langcode e.g. "sv_SV"
+		if (myLangCodeString != "")
+			language_filter_types_.emplace_back(std::pair<int, std::string>(i, langcode_to_string(languages_available[i])));
+		// Remember en_GB pos in toggle vector, so packages with Translations: None can be defaulted to English (GB) lang_string
+		if (languages_available[i] == "en_GB")
+			en_GB_toggle_position = language_filter_types_.size() - 1;
```
Pentarctagon, line 417: Why is en_GB being treated as a special case?

src/gui/dialogs/addon/manager.cpp
Kingofd: Yes, this is intentional, since the already built in function `langcode_to_string` will only give out a valid translation for Portugues do Brasil. For the current state of localisation concerning language codes (pt, de, etc.) to language strings (Portugues, Deutsch etc.) available over that interface it should be correct. 

If one would refactor the `langcode_to_string` function that line would have to go, if it would provide a correct mapping for language codes to language strings

src/gui/dialogs/addon/manager.cpp
Kingofd: It is for removing [duplicates](https://stackoverflow.com/questions/16476099/remove-duplicate-entries-in-a-c-vector#)

src/gui/dialogs/addon/manager.cpp
Kingofd: All of the addons, where no translations are specified are assumed to have at least an English translation, so defaulting to English, that's why I need to know en_GBs position in the scraped lang code vector. 

As for why it is en_GB and not en_US, it's personal preference, I reckon

src/gui/dialogs/addon/manager.cpp
Pentarctagon: Is there a reason to not use an std::set?

src/gui/dialogs/addon/manager.cpp
Pentarctagon: Right, but why is this needed in the first place? Showing everything when no language filter options are chosen should be enough by itself, I would think.

Kingofd: Ah, now I see what you mean. Indeed, a person knowing that every addon has an English translation would then simply not filter the list. But then the English filter option would be effectively filtering nothing, if every addon really has an English translation and we would assume that every person using the addon store knows that. 

So how I see it we should either remove en_GB/en_US language filter option completely, as every addon is English by default anyways or we should keep it for UX purposes and show everything when it is selected, making it so that it filters nothing.

src/gui/dialogs/addon/manager.cpp
Kingofd: No, I will refactor it to use set. Can I use the C++20 standard options of std::set?

Pentarctagon: No - Wesnoth currently only allows up to C++17.

src/gui/dialogs/addon/manager.cpp
Pentarctagon: I'm not really following. There should be no assumption that every add-on is English or any particular language. Leaving the language filter empty simply means that the player doesn't care what translations are available.

If there exists an add-on with an en_US or en_GB translation added, then it should end up in the language filter drop down the same as any other language.

Pentarctagon: To add to that - the default language for any particular add-on is simply unknown. All we have available to be able to possibly filter on is what additional translations the add-on author makes available:
* If the default text for an add-on is in **English** and they add a **French** translation, then **French** is what it is filterable on.
* If the default text for an add-on is in **French** and they add an **English** translation, then **English** is what it is filterable on.
* When the language filter is empty, both add-ons from those two above examples should be displayed, not because we're assuming all add-ons are English by default but because the available translations aren't even looked at in the first place when deciding what to display on the UI.

CelticMinstrel: Shouldn't the default language of an add-on be known and included? That would make a lot more sense to me. How are people supposed to find add-ons in their language if some of them might have that language as the default and others might have it as a translation?

CelticMinstrel:
Also, as an aside, I think it would be nicer to filter only on language, not on locale. So en_US and en_GB are treated as equivalent. Someone who speaks pt_BR should still be able to more-or-less understand a pt_PT, so filtering on "Portuguese" should show add-ons with either language in my opinion.

src/gui/dialogs/addon/manager.cpp
Pentarctagon: Perhaps, but I'd say that's something for a separate issue/PR.

Pentarctagon:
The current implementation is multi-select so that'd already be possible to do if the player wants to.

CelticMinstrel:
I'm suggesting that we can simplify it for the player by filtering on language only. There _could_ be an option to secondarily filter by locale if there's demand, I suppose, but I think filtering on language alone is more broadly useful.

Kingofd:
It could be an issue building on this work, I think. That would need to tackle an improvement of the `langcode_to_string()` function, because that currently would not allow such behaviour

Pentarctagon (approved):

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +391,37 @@ void addon_manager::pre_show(window& window)
...
+	// Prepare shown languages, source all available languages from the addons themselves
+	std::set<std::string> languages_available;
+	for(const auto& a : addons_) {
+		for (const auto& b : a.second.locales) {
+			languages_available.insert(b);
+		}
+	}
+
+	// Erase pt value, since otherwise Portugues do Brasil will be shown twice in the language list
+	languages_available.erase(languages_available.find("pt"));
```
stevecotton, line 404: I'd prefer this to be some logic instead of a hardcoded list, because there will eventually be two `es` languages too.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +391,37 @@ void addon_manager::pre_show(window& window)
...
+	for (auto i: languages_available) {
```
Pentarctagon, line 405: Should be `const auto&`.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +391,37 @@ void addon_manager::pre_show(window& window)
...
+		std::string myLangCodeString = langcode_to_string(i);
```
Pentarctagon, line 406: Should be snake case.

src/gui/dialogs/addon/manager.cpp
Pentarctagon: As discussed on Discord: it would be better to have the language display name inserted into the set rather than hard coding the removal of "pt".

So change
`languages_available.insert(b);`
to
`languages_available.insert(langcode_to_string(b));`
and then erase with:
`languages_available.erase(languages_available.find(""));`

CelticMinstrel: I definitely don't think hard-coding a list is appropriate here, as there are any number of language codes that have variants. For example, fr_FR vs fr_CA, to name a common one. (Wesnoth may not have both _now_, but there's no reason to assume it will never have both.)

That said, you do have a point that Simplified vs Traditional Chinese are different enough that people might want to distinguish. This is kind of a difficult problem. I definitely think that filtering by language alone is a more sensible default, but it's also not terrible to support filtering on language plus locale as an option. That is quite a huge complication to this system, though – I think to really handle it well you'd need a completely new menu type (where the menu contents is a tree view similar to what you see sometimes in an app installer, with tri-state checkboxes on the non-lead nodes and regular dual-state checkboxes on the leaves).

So doing both is definitely out of scope for this pull request. This can be thought of as a first iteration of the language filtering system, so I'd say it should either:

1. Just filter on full locale – don't automatically merge pt_BR and pt_PT or anything like that.
2. Filter only on language, but add a hard-coded exception for (at least) Chinese so they can be filtered separately.

Just having the ability to filter by language is already great, and the system can be further refined later. I actually think being able to set an add-on's default locale is more important than limiting the filter to language alone without considering the full locale.

Kingofd: Guys, I found something I should have done a long time ago that fixes all these problems: getting ALL languages regardless of translation percentage. Normally only languages are shown that only exceed a certain translation threshold, but for the language filter case that should not matter. Therefore I show now all possible localisations (see pictures)
LangFilterNow

@Pentarctagon that turned out to be inviting some faulty behaviour in the GUI (see picture), so I kept the old implementation, which somehow eliminates that behaviour
LangFilterBroken

Pentarctagon: I'm not sure what issue the image is meant to show?

Kingofd: The language button dropdown menu fills the entire screen and is empty

Pentarctagon: Huh, well that's unexpected. Anyway, if the current code in the PR works, that's good then.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -391,6 +391,35 @@ void addon_manager::pre_show(window& window)
...
+			languages_available.insert(b);
+		}
+	}
+
+	for (const auto& i: languages_available) {
+		std::string lang_code_string = langcode_to_string(i);
+		// Only show languages, which have a translation as per langcode_to_string() method
+		// Do not show tranlations with their langcode e.g. "sv_SV"
+		if (lang_code_string != "")
+			language_filter_types_.emplace_back(std::pair<int, std::string>	(language_filter_types_.size(), lang_code_string));
```
Vultraz, line 408: Everything else looks good, though I would suggest one small style change here:
```cpp
if(std::string lang_code_string = langcode_to_string(i); !lang_code_string.empty()) {
	language_filter_types_.emplace_back(language_filter_types_.size(), std::move(lang_code_string));
}
```

Vultraz (approved):

Pentarctagon (approved):

src/gui/dialogs/addon/manager.cpp
Pentarctagon: I'll go ahead and merge once this is done then.

Pentarctagon:
Thanks!

PR #7050, opened 2022-10-15
Extend ability removal with remove_ability by attributes and no longer just by id

newfrenchy83:
using the id to designate the ability to remove can be quite tedious and especially incomplete when you want to use remove_ability to remove a whole type of ability with common characteristics. For this I suggest 1, to be able to filter the abilities by type. (we still need to place the id in the tag but this one cannot be differentiated from another) 2) filter by attribute in addition to id (name, affect_self/allies/enemies, active_on or apply_to or overwrite_specials for combat ability equivalent to a special 3) be able to place attributes outside ability tags to filter multiple types (works when no tag used in filtering)

the type of ability used is also filtered.

i apply hese modification in FORCE_CHANCE_TO_HIT macro for test.

### DIFF
--- a/data/core/macros/utils.cfg
+++ b/data/core/macros/utils.cfg
@@ -535,6 +535,54 @@
             [/and]
         [/filter_condition]
 
+        [store_unit]
+            [filter]
+                [filter_adjacent]
+                    x,y=$x1,$y1
+                [/filter_adjacent]
+                [not]
+                    x,y=$x2,$y2
+                [/not]
+                [not]
+                    x,y=$x1,$y1
+                [/not]
+            [/filter]
+            variable=force_remove
+        [/store_unit]
+
+        [foreach]
+            array=force_remove
+            [do]
+                [object]
+                    id=forced_re
+                    [filter]
+                        id=$force_remove[$i].id
+                    [/filter]
+                    silent=yes
+                    duration=turn end
+                    take_only_once=no
+                    [effect]
+                        apply_to=remove_ability
+                        [filter_ability]
+                            tag_name=chance_to_hit
+                            [and]
+                                overwrite_specials=both_sides,one_side
+                            [/and]
+                            [and]
+                                affect_allies=yes
+                                [or]
+                                    affect_enemies=yes
+                                [/or]
+                            [/and]
+                            [not]
+                                apply_to=opponent,defender
+                            [/not]
+                        [/filter_ability]
+                    [/effect]
+                [/object]
+            [/do]
+        [/foreach]
+
         [object]
             id=forced_cth
             [filter]
@@ -543,6 +591,18 @@
             silent=yes
             duration=turn end
             take_only_once=no
+            [effect]
+                apply_to=remove_ability
+                [filter_ability]
+                    tag_name=chance_to_hit
+                    [and]
+                        overwrite_specials=both_sides,one_side
+                    [/and]
+                    [not]
+                        apply_to=opponent,defender
+                    [/not]
+                [/filter_ability]
+            [/effect]
             [effect]
                 apply_to=new_ability
                 [abilities]
@@ -564,6 +624,10 @@
             [remove_object]
                 object_id=forced_cth
             [/remove_object]
+            [remove_object]
+                object_id=forced_re
+            [/remove_object]
+            {CLEAR_VARIABLE force_remove}
         [/event]
     [/event]
 
@@ -587,6 +651,53 @@
                 {EXTRA_CONDITIONS_WML}
             [/and]
         [/filter_condition]
+        [store_unit]
+            [filter]
+                [filter_adjacent]
+                    x,y=$x2,$y2
+                [/filter_adjacent]
+                [not]
+                    x,y=$x2,$y2
+                [/not]
+                [not]
+                    x,y=$x1,$y1
+                [/not]
+            [/filter]
+            variable=force_remove
+        [/store_unit]
+
+        [foreach]
+            array=force_remove
+            [do]
+                [object]
+                    id=forced_re
+                    [filter]
+                        id=$force_remove[$i].id
+                    [/filter]
+                    silent=yes
+                    duration=turn end
+                    take_only_once=no
+                    [effect]
+                        apply_to=remove_ability
+                        [filter_ability]
+                            tag_name=chance_to_hit
+                            [and]
+                                overwrite_specials=both_sides,one_side
+                            [/and]
+                            [and]
+                                affect_allies=yes
+                                [or]
+                                    affect_enemies=yes
+                                [/or]
+                            [/and]
+                            [not]
+                                apply_to=opponent,attacker
+                            [/not]
+                        [/filter_ability]
+                    [/effect]
+                [/object]
+            [/do]
+        [/foreach]
 
         [object]
             id=forced_cth
@@ -596,6 +707,18 @@
             silent=yes
             duration=turn end
             take_only_once=no
+            [effect]
+                apply_to=remove_ability
+                [filter_ability]
+                    tag_name=chance_to_hit
+                    [and]
+                        overwrite_specials=both_sides,one_side
+                    [/and]
+                    [not]
+                        apply_to=opponent,attacker
+                    [/not]
+                [/filter_ability]
+            [/effect]
             [effect]
                 apply_to=new_ability
                 [abilities]
@@ -617,6 +740,10 @@
             [remove_object]
                 object_id=forced_cth
             [/remove_object]
+            [remove_object]
+                object_id=forced_re
+            [/remove_object]
+            {CLEAR_VARIABLE force_remove}
         [/event]
     [/event]
     [event]
@@ -633,12 +760,10 @@
     [/event]
     [event]
         name=prestart,start,victory
-        [filter_condition]
-            [have_unit]
-                ability=forced_cth
-            [/have_unit]
-        [/filter_condition]
 
+        [remove_object]
+            object_id=forced_re
+        [/remove_object]
         [remove_object]
             object_id=forced_cth
         [/remove_object]
--- a/data/schema/filters/abilities.cfg
+++ b/data/schema/filters/abilities.cfg
@@ -1,0 +1,20 @@
+
+[tag]
+	name="$filter_abilities"
+	max=0
+	{SIMPLE_KEY id string_list}
+	{SIMPLE_KEY tag_name string_list}
+	{SIMPLE_KEY overwrite_specials string_list}
+	{SIMPLE_KEY apply_to string_list}
+	{SIMPLE_KEY active_on string_list}
+	{SIMPLE_KEY value string_list}
+	{SIMPLE_KEY add string_list}
+	{SIMPLE_KEY sub string_list}
+	{SIMPLE_KEY multiply string_list}
+	{SIMPLE_KEY divide string_list}
+	{SIMPLE_KEY affect_self s_bool}
+	{SIMPLE_KEY affect_allies s_bool}
+	{SIMPLE_KEY affect_enemies s_bool}
+	{DEFAULT_KEY type_value value_type empty}
+	{FILTER_BOOLEAN_OPS abilities}
+[/tag]
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -1408,6 +1408,136 @@ void unit::remove_ability_by_id(const std::string& ability)
 	}
 }
 
+static bool bool_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, bool def)
+{
+	if(filter[attribute].empty()) {
+		return true;
+	}
+
+	return filter[attribute].to_bool() == cfg[attribute].to_bool(def);
+}
+
+static bool string_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, const std::string& def)
+{
+	if(filter[attribute].empty()) {
+		return true;
+	}
+
+	const std::vector<std::string> filter_attribute = utils::split(filter[attribute]);
+	return ( std::find(filter_attribute.begin(), filter_attribute.end(), cfg[attribute].str(def)) != filter_attribute.end() );
+}
+
+static bool type_value_if_present(const config& filter, const config& cfg)
+{
+	if(filter["type_value"].empty()) {
+		return true;
+	}
+
+	std::string cfg_type_value;
+	const std::vector<std::string> filter_attribute = utils::split(filter["type_value"]);
+	if(!cfg["value"].empty()){
+		cfg_type_value ="value";
+	} else if(!cfg["add"].empty()){
+		cfg_type_value ="add";
+	} else if(!cfg["sub"].empty()){
+		cfg_type_value ="sub";
+	} else if(!cfg["multiply"].empty()){
+		cfg_type_value ="multiply";
+	} else if(!cfg["divide"].empty()){
+		cfg_type_value ="divide";
+	}
+	return ( std::find(filter_attribute.begin(), filter_attribute.end(), cfg_type_value) != filter_attribute.end() );
+}
+
+static bool matches_ability_filter(const config & cfg, const std::string& tag_name, const config & filter)
+{
+	if(!bool_matches_if_present(filter, cfg, "affect_self", true))
+		return false;
+
+	if(!bool_matches_if_present(filter, cfg, "affect_allies", true))
+		return false;
+
+	if(!bool_matches_if_present(filter, cfg, "affect_enemies", false))
+		return false;
+
+	if(!bool_matches_if_present(filter, cfg, "cumulative", false))
+		return false;
+
+	const std::vector<std::string> filter_type = utils::split(filter["tag_name"]);
+	if ( !filter_type.empty() && std::find(filter_type.begin(), filter_type.end(), tag_name) == filter_type.end() )
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "id", ""))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "apply_to", "self"))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "overwrite_specials", "none"))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "active_on", "both"))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "value", ""))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "add", ""))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "sub", ""))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "multiply", ""))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "divide", ""))
+		return false;
+
+	if(!type_value_if_present(filter, cfg))
+		return false;
+
+	// Passed all tests.
+	return true;
+}
+
+bool unit::ability_matches_filter(const config & cfg, const std::string& tag_name, const config & filter) const
+{
+	// Handle the basic filter.
+	bool matches = matches_ability_filter(cfg, tag_name, filter);
+
+	// Handle [and], [or], and [not] with in-order precedence
+	for (const config::any_child condition : filter.all_children_range() )
+	{
+		// Handle [and]
+		if ( condition.key == "and" )
+			matches = matches && ability_matches_filter(cfg, tag_name, condition.cfg);
+
+		// Handle [or]
+		else if ( condition.key == "or" )
+			matches = matches || ability_matches_filter(cfg, tag_name, condition.cfg);
+
+		// Handle [not]
+		else if ( condition.key == "not" )
+			matches = matches && !ability_matches_filter(cfg, tag_name, condition.cfg);
+	}
+
+	return matches;
+}
+
+void unit::remove_ability_by_attribute(const config& filter)
+{
+	set_attr_changed(UA_ABILITIES);
+	config::all_children_iterator i = abilities_.ordered_begin();
+	while (i != abilities_.ordered_end()) {
+		if(ability_matches_filter(i->cfg, i->key, filter)) {
+			i = abilities_.erase(i);
+		} else {
+			++i;
+		}
+	}
+}
+
 bool unit::get_attacks_changed() const
 {
 	for(const auto& a_ptr : attacks_) {
@@ -2115,9 +2245,13 @@ void unit::apply_builtin_effect(std::string apply_to, const config& effect)
 		}
 	} else if(apply_to == "remove_ability") {
 		if(const config& ab_effect = effect.child("abilities")) {
+			deprecated_message("[effect]apply_to=remove_ability [abilities]", DEP_LEVEL::INDEFINITE, "", "Use [filter_ability] instead in [effect]apply_to=remove_ability");
 			for(const config::any_child ab : ab_effect.all_children_range()) {
 				remove_ability_by_id(ab.cfg["id"]);
 			}
+		}
+		if(const config& fab_effect = effect.child("filter_ability")) {
+			remove_ability_by_attribute(fab_effect);
 		}
 	} else if(apply_to == "image_mod") {
 		LOG_UT << "applying image_mod";
--- a/src/units/unit.hpp
+++ b/src/units/unit.hpp
@@ -1772,6 +1772,20 @@ public:
 	 */
 	void remove_ability_by_id(const std::string& ability);
 
+	/**
+	 * Removes a unit's abilities with a specific ID or other attribute.
+	 * @param filter the config of ability to remove.
+	 */
+	void remove_ability_by_attribute(const config& filter);
+
+	/**
+	 * Verify what abilities attributes match with filter.
+	 * @param cfg the config of ability to check.
+	 * @param tag_name the tag name of ability to check.
+	 * @param filter the filter used for checking.
+	 */
+	bool ability_matches_filter(const config & cfg, const std::string& tag_name, const config & filter) const;
+
 
 private:
 
--- a/changelog_entries/extent_remove_ability_efect_to_attribute_changelog.md
+++ b/changelog_entries/extent_remove_ability_efect_to_attribute_changelog.md
@@ -1,0 +1,2 @@
+### WML Engine
+   * Add [filter_ability] usable instead of [abilities][tag name] to filter attributes including the type of ability used.
--- a/data/schema/game_config.cfg
+++ b/data/schema/game_config.cfg
@@ -41,6 +41,10 @@
         value="none|one_side|both_sides"
     [/type]
     [type]
+        name="value_type"
+        value="empty|value|add|sub|multiply|divide"
+    [/type]
+    [type]
         name="addon_type"
         value="sp|mp|hybrid"
     [/type]
--- a/data/schema/units/modifications.cfg
+++ b/data/schema/units/modifications.cfg
@@ -138,8 +138,13 @@
 				[/tag]
 			[/case]
 			[case]
-				value=new_ability,remove_ability
+				value=new_ability
 				{LINK_TAG "units/unit_type/abilities"}
+			[/case]
+			[case]
+				value=remove_ability
+				{LINK_TAG "units/unit_type/abilities"}
+				{FILTER_TAG "filter_ability" abilities ()}
 			[/case]
 			[case]
 				value=new_animation
--- a/data/test/scenarios/test_force_chance_to_hit_macro.cfg
+++ b/data/test/scenarios/test_force_chance_to_hit_macro.cfg
@@ -1,9 +1,4 @@
-# This unit test defines a WML object based implementation of the "unupgradable" ability
-# https://github.com/ProditorMagnus/Ageless-for-1-14/blob/52c1eaf31722bb58046a1b459d4f29daa8d88487/data/general_data/weapon_specials/unupgradable.cfg
-# and checks that it works. What is being tested here is that
 # FORCE_CHANCE_TO_HIT macro must doing misses all attacks then what [chance_to_hit] value equals 100%
-# - through [attacks]
-# - through [effect] increase_attacks
 
 {GENERIC_UNIT_TEST "test_force_chance_to_hit_macro" (
     {FORCE_CHANCE_TO_HIT (id=bob) (id=alice) 0 ()}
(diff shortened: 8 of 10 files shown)


CelticMinstrel:
Okay, so if I understand correctly, this is adding a new type of filter, rather than using FilterWML (`config::matches`)? Your schema doesn't match your sample code, though – you'll have to replace the LINK_TAG with an explicit tag definition using `super=` to connect it to the base abilities tag… although, I'm not quite sure _that_ is right either. Maybe it's the subtags of `[abilities]` that needs to use `super=`.

newfrenchy83:
> Okay, so if I understand correctly, this is adding a new type of filter, rather than using FilterWML (`config::matches`)?
> 
> Your schema doesn't match your sample code, though – you'll have to replace the LINK_TAG with an explicit tag definition using `super=` to connect it to the base `[abilities]` tag… although, I'm not quite sure _that_ is right either. Maybe it's the subtags of `[abilities]` that needs to use `super=`.

i simplify code for remove subtags and split to shema validation.

newfrenchy83:
@CelticMinstrel i think what this PR can be reviewed now.

src/units/unit.cpp
```diff
@@ -1395,12 +1395,132 @@ bool unit::has_ability_by_id(const std::string& ability) const
...
-void unit::remove_ability_by_id(const std::string& ability)
+static bool matches_ability_filter(config::all_children_iterator i, const config & filter, const std::string& ability, const std::set<std::string> checking_tags, bool in_tag)
+{
+	const std::string& ab_active_on = !i->cfg["active_on"].empty() ? i->cfg["active_on"].str() : "both";
+	const std::string& ab_overwrite_specials = !i->cfg["overwrite_specials"].empty() ? i->cfg["overwrite_specials"].str() : "none";
+	const std::string& ab_apply_to = !i->cfg["apply_to"].empty() ? i->cfg["apply_to"].str() : "self";
+	bool ab_affect_self = i->cfg["affect_self"].to_bool(true);
+	bool ab_affect_allies = i->cfg["affect_allies"].to_bool(true);
+	bool ab_affect_enemies = i->cfg["affect_enemies"].to_bool();
+	bool ab_cumulative = i->cfg["cumulative"].to_bool();
```
stevecotton, line 1406: I'm not sure which existing C++ functions would be the right ones to use, but a lot of this feels like code that should already have convenience functions in the existing filter implementations.

newfrenchy83: true for affect_self who have an function callable from unit::: function but not other in namespace in abilities.cpp

newfrenchy83: it not for verify what affect_allies/enemies were apppied to unit adjacent but only what attribute present in abilities with [affect_adjacent] tags

data/schema/filters/abilities.cfg
```diff
@@ -0,0 +1,15 @@
+
+[tag]
+	name="$filter_abilities"
+	max=0
+	{SIMPLE_KEY id string_list}
+	{SIMPLE_KEY name string_list}
```
CelticMinstrel, line 6: Ability name is a translatable string, right? It's probably not a good idea to support filtering in a translatable string.

newfrenchy83: Ok, i will remove name then, id sufficient like attribute without use name

CelticMinstrel:
I don't really understand the purpose of `checking_tags` (and `checking_weapon_*`) and `in_tag`. It seems like you may be overcomplicating the system a little bit, but I could be wrong.

src/units/unit.cpp
```diff
@@ -2120,9 +2236,13 @@ void unit::apply_builtin_effect(std::string apply_to, const config& effect)
 	} else if(apply_to == "remove_ability") {
 		if(const config& ab_effect = effect.child("abilities")) {
 			for(const config::any_child ab : ab_effect.all_children_range()) {
-				remove_ability_by_id(ab.cfg["id"]);
+				const config& filter = ab_effect.child(ab.key);
```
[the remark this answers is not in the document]
newfrenchy83: in this part tag of abilities used for filttered individualyy each abilities to remove(with id) or other attrbute:
example `[chance_to_hit]
overwrite_specials=one_side
apply_to=opponent
[/chance_to_hit]
[chance_to_hit]
overwrite_specials=both_sides
apply_to=opponent
[/chance_to_hit]
[chance_to_hit]
overwrite_specials=one_side
apply_to=defender
[/chance_to_hit]
[chance_to_hit]
overwrite_specials=both_sides
apply_to=defender
[/chance_to_hit]`

the wtriting is heavy, this why i added afilter_abilities below. 

and it is not ab.child because not compilabe, this must be here ab_effect.chil(ab.key)

newfrenchy83:
> I don't really understand the purpose of `checking_tags` (and `checking_weapon_*`) and `in_tag`. It seems like you may be overcomplicating the system a little bit, but I could be wrong.

overwrite_specials is used only for weapon like abilities with numerical value, check this attribute in other abilities don't make sensebe cause don't function in hides abilities orother. the attributes checked must be influence work of abilities and i don't want what an ability removed because an attribute added by error is present here it musn't be.

src/units/unit.cpp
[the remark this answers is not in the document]
CelticMinstrel: What if you removed the `const` on the line above? Would `ab.child` work then?

CelticMinstrel:
Okay, so if I parsed your words correctly, you're trying to say that `checking_tags` and `checking_weapon_*` are used to ensure that a filter on `value` is considered not to match when matching against an ability that does not have a value? So for example, the following filter:

```
[filter_abilities]
  value=5
[/filter_abilities]
```

will not match the following ability:

```
[hides]
  // whatever
[/hides]
```

?

I still don't understand the purpose of `in_tag`.

newfrenchy83:
> Okay, so if I parsed your words correctly, you're trying to say that `checking_tags` and `checking_weapon_*` are used to ensure that a filter on `value` is considered not to match when matching against an ability that does not have a value? So for example, the following filter:
> 
> ```
> [filter_abilities]
>   value=5
> [/filter_abilities]
> ```
> 
> will not match the following ability:
> 
> ```
> [hides]
>   // whatever
> [/hides]
> ```
> 
> ?
> 
> I still don't understand the purpose of `in_tag`.

if tag is used then id and other attributes musn't be writed with separation but one id by tag, i is already the case in master for id and i don't want change that, this why using of split is exclusive to filter_abilities tags and in_tag used for doing diffrence between two case(ability tag and [filter_abilities]. the function what i want add don't penalise work of wml develloper

src/units/unit.cpp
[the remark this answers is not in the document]
newfrenchy83: no, i  have 'src\units\unit.cpp|2240|error: 'struct config::any_child' has no member named 'child'|

data/schema/filters/abilities.cfg
newfrenchy83: removed

CelticMinstrel:
> if tag is used then id and other attributes musn't be writed with separation but one id by tag, i is already the case in master for id and i don't want change that, this why using of split is exclusive to filter_abilities tags and in_tag used for doing diffrence between two case(ability tag and [filter_abilities]. the function what i want add don't penalise work of wml develloper

So if I'm getting this right, the purpose of `in_tag` is to preserve backwards compatibility?

If that's the case, you should have a unit test to verify that behaviour hasn't changed. From looking at the code, I'm fairly sure it has.

newfrenchy83:
> So if I'm getting this right, the purpose of `in_tag` is to preserve backwards compatibility?
> 
> If that's the case, you should have a unit test to verify that behaviour hasn't changed. From looking at the code, I'm fairly sure it has.

test added

data/test/scenarios/test_remove_object_compatibility.cfg
```diff
@@ -0,0 +1,118 @@
+# This unit test is used for verify backward compatibility of remove_ability
+
+{GENERIC_UNIT_TEST "test_remove_object_compatibility" (
```
CelticMinstrel, line 3: This (and the filename) should probably say "remove_ability" rather than "remove_object".

data/test/scenarios/test_remove_object_compatibility.cfg
```diff
@@ -0,0 +1,118 @@
...
+            [filter]
+                id=bob
+            [/filter]
+            silent=yes
+            [effect]
+                apply_to=remove_ability
+                [abilities]
+                    [chance_to_hit]
+                        id=test_cth
+                    [/chance_to_hit]
```
CelticMinstrel, line 61: The thing that this is supposed to test is that the tag name here is ignored. So, change it to something other than `chance_to_hit` (but the ability itself still uses `chance_to_hit`). Then try running the unit test on unmodified master to make sure it passes there. If it passes on master, then check if it passes on this branch.

src/units/unit.cpp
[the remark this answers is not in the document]
CelticMinstrel: My mistake. Looks like it's supposed to be `ab.cfg`, not `ab.child`. I should double-check the correct name before I make suggestions…

data/test/scenarios/test_remove_object_compatibility.cfg
newfrenchy83: In fact tag name no longer ignored, i will replace (return false) in checking by deprecated message.

src/units/unit.cpp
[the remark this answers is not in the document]
newfrenchy83: changed

data/test/scenarios/test_remove_object_compatibility.cfg
CelticMinstrel: But tag name _is_ ignored on master, correct?

CelticMinstrel:
Okay, now we need to discuss how to handle this deprecation…

data/test/scenarios/test_remove_object_compatibility.cfg
newfrenchy83: correct, a tag is necessary, but tag name can be diffrent of ability tag name, and it is a problem in my point of view. the logic would what wml devellopper use the right tag name but if this particularity is know, impose the checking without advertissement will cause problem

src/units/unit.cpp
```diff
@@ -1395,12 +1395,121 @@ bool unit::has_ability_by_id(const std::string& ability) const
...
+
+	if(i->cfg.has_child("affect_adjacent")){
+		if ( !filter["affect_allies"].empty() && filter_affect_allies != i->cfg["affect_allies"].to_bool(true) )
+			return false;
+
+		if ( !filter["affect_enemies"].empty() && filter_affect_enemies != i->cfg["affect_enemies"].to_bool() )
+			return false;
+	}
+
+	const std::set<std::string> checking_weapon_abilities_with_value = {"leadership", "resistance", "attacks", "berserk", "damage", "chance_to_hit", "swarm", "drains", "heal_on_hit"};
```
stevecotton, line 1473: I think this list could be omitted - if the filter includes a `cumulative`, then why not check whether `i->cfg` is cumulative regardless of the tag name?

src/units/unit.hpp
```diff
@@ -1754,10 +1754,12 @@ class unit : public std::enable_shared_from_this<unit>
...
 	/**
-	 * Removes a unit's abilities with a specific ID.
-	 * @param ability The type of ability (tag name) to remove.
+	 * Removes a unit's abilities with a specific ID or other attribute.
+	 * @param filter the config of ability to remove.
+	 * @param ability The type of ability (tag) to remove.
+	 * @param in_tag if attribute checked was in tag abilities or in filter tag
 	 */
-	void remove_ability_by_id(const std::string& ability);
+	void remove_ability_by_attribute(const config& filter, const std::string& ability = "", bool in_tag = false);
```
stevecotton, line 1762: If a function is only called from two places, then I think it's clearer have no default values for arguments. With only two calling locations, just put the argument values in the calling location.

src/units/unit.cpp
```diff
@@ -1395,12 +1395,121 @@ bool unit::has_ability_by_id(const std::string& ability) const
...
+					return false;
+			}
+		}
+		if((checking_tags.count(i->key) != 0) || i->key == "resistance"){
+			if ( !filter_active_on.empty() && std::find(filter_active_on.begin(), filter_active_on.end(), ab_active_on) == filter_active_on.end() )
+				return false;
+		}
+	}
+
+	if ( !filter["affect_self"].empty() && filter_affect_self != i->cfg["affect_self"].to_bool(true) )
```
stevecotton, line 1462: It would need several helper functions, but I feel this would be easier to read given a helper
```
/**
 * If @a filter doesn't contain @a attribute, returns true (corresponding to a filter that isn't checking for the given attribute).
 * Otherwise returns true if @a cfg[attribute] matches @filter[attribute], comparing as a WML boolean with default value @def.
 *
 * TODO: I'm sure where, but surely this function already exists somewhere in the config classes or utility library?
 */
static bool matches_if_present_in_filter(const config& filter, const config& cfg, const std::string& attribute, bool def)
{
	if(filter[attribute].empty()) {
		return true;
	}

	return filter[attribute].to_bool() == cfg[attribute].to_bool(def);
}
```
so that the earlier definition of `filter_affect_self` could be removed, and the two lines here replaced with
```
	if(!matches_if_present_in_filter(filter, i->cfg, "affect_self", true))
		return false;
```

@CelticMinstrel does that helper function already exist somewhere?

src/units/unit.cpp
newfrenchy83: a yes, in shema, cumulative is in generic attribute, then what somes abilities doen't use numerical value and cumulative doesn't have effect. ok i remove the list then.

src/units/unit.hpp
newfrenchy83: done

newfrenchy83:
i fear what fail orchecking is because of https://github.com/wesnoth/wesnoth/commit/dd52baf2472db20d3cf4f28e48a55b3a9ab37253#diff-762206158d5fa674710fbb3ace9e31b12651fa53ceb38754aa67977b85bc2115 who must be removed

I TEST in my local belore and afeter commit, and it is only after commit what test fail

stevecotton:
Do you have more detail about what's broken?

I tested dd52baf2472db + 4831160feab, and the tests passed locally on that build.

newfrenchy83:
i removed last commit in wait to find error

newfrenchy83:
> Do you have more detail about what's broken?
> 
> I tested [dd52baf](https://github.com/wesnoth/wesnoth/commit/dd52baf2472db20d3cf4f28e48a55b3a9ab37253) + [4831160](https://github.com/wesnoth/wesnoth/commit/4831160feab6ca754dbb44681a655b835aee553c), and the tests passed locally on that build.

i done an error here, but i not find yet.

stevecotton:
In the reverted version, these two lines had broken logic and would `return false` on any cfg that lacks `affect_adjacent`, regardless of whether there's an `affect_allies`, an `affect_enemies`, or neither.

```
-       if(!matches_if_present_in_filter(filter, i->cfg, "affect_allies", true)|| !(i->cfg.has_child("affect_adjacent")))
-               return false;
-
-       if(!matches_if_present_in_filter(filter, i->cfg, "affect_enemies", false) || !(i->cfg.has_child("affect_adjacent")))
-               return false;
```


I'll be AFK for the rest of the (European) night, so won't check any changes until about 12 hours from now. Even then, take your time, there's no need to rush.

src/units/unit.cpp
```diff
@@ -1395,12 +1395,98 @@ bool unit::has_ability_by_id(const std::string& ability) const
...
+static bool matches_if_present_in_filter(const config& filter, const config& cfg, const std::string& attribute, bool def)
+{
+	if(filter[attribute].empty()) {
+		return true;
+	}
+
+	return filter[attribute].to_bool() == cfg[attribute].to_bool(def);
+}
+
+static bool string_matches_if_present_in_filter(const config& filter, const config& cfg, const std::string& attribute, const std::string& def, bool in_tag=false)
```
stevecotton, line 1407: I still haven't looked for existing implementations of this function, but I'd prefer to split this into two functions that would be generic enough to be moved to a general "filter helpers" library. Although that would mean putting an `if(in_tag) {...} else {}` block in `matches_ability_filter`, I think that be worthwhile if we can reuse the generic functions elsewhere.

* One with the current name, that's equivalent to this with `in_tag=false`.
* A `string_list_matches_if_present_in_filter`, which is equivalent to this with `in_tag=true`.

Alternatively, rename `in_tag` to something that would suit a generic function, such as `split_filter_as_list`.

newfrenchy83:
@stevecotton , @CelticMinstrel i hpe what you don't forget this PR.

CelticMinstrel:
We're currently waiting on a discussion of how to handle the deprecation.

newfrenchy83:
> I'm currently waiting on a discussion of how to handle the deprecation.

I removed the deprecation message and the ability type check is confined to [filter_abilities] so as not to generate compatibility issues since I don't know how to implement the message since the old method is not deprecated in itself and is even enriched by several attributes. I prefer to postpone this debate since I have two other PRs waiting plus one that I am working on locally but that requires some code in it before being proposed.

src/units/unit.cpp
```diff
@@ -1423,6 +1423,34 @@ static bool string_matches_if_present_in_filter_split(const config& filter, cons
...
+	return filter[attribute].to_int() == cfg[attribute].to_int();
+}
+
+static bool double_matches_if_present_in_filter(const config& filter, const config& cfg, const std::string& attribute)
+{
+	if(filter[attribute].empty()) {
+		return true;
+	}
+
+	return filter[attribute].to_double() == cfg[attribute].to_double();
```
stevecotton, line 1441: Comparing floating point numbers using `==` isn't right.

data/schema/filters/abilities.cfg
```diff
@@ -10,5 +10,10 @@
 	{SIMPLE_KEY affect_self s_bool}
 	{SIMPLE_KEY affect_allies s_bool}
 	{SIMPLE_KEY affect_enemies s_bool}
+	{SIMPLE_KEY value s_range_list}
+	{SIMPLE_KEY add s_range_list}
+	{SIMPLE_KEY sub s_range_list}
+	{SIMPLE_KEY multiply f_real}
+	{SIMPLE_KEY divide f_real}
```
stevecotton, line 17: Adding extra stuff means more to review, and should mean more unit tests too.

Seems odd that `add` and `sub` support ranges, but `multiply` and `divide` don't.

newfrenchy83: i reverted that

src/units/unit.cpp
```diff
@@ -2120,9 +2222,12 @@ void unit::apply_builtin_effect(std::string apply_to, const config& effect)
 	} else if(apply_to == "remove_ability") {
 		if(const config& ab_effect = effect.child("abilities")) {
 			for(const config::any_child ab : ab_effect.all_children_range()) {
-				remove_ability_by_id(ab.cfg["id"]);
+				remove_ability_by_attribute(ab.cfg, false);
 			}
 		}
+		if(const config& fab_effect = effect.child("filter_abilities")) {
```
stevecotton, line 2228: I'm wondering if the tag should be `[filter_ability]` rather than `[filter_abilities]`, to match the singular/plural of `[filter_attack]`, `[filter_location]`, `[filter_side]`, etc. However, that wouldn't agree with the singular/plural of the existing `[effect]apply_to=remove_ability[abilities]` tag.

stevecotton (changes requested):
@CelticMinstrel there's no deprecation here anymore, the old tag isn't being removed. I think that's good, as there's probably a lot of UMC using it, but is that OK for you?

Draft of the revised EffectWML for the Wiki:
```
* '''remove_ability''': Removes one or more abilities from a unit. Abilities are not reference counted: added, added, removed = gone. Exactly one of the two filter subtags should be present - if both are present then they'll be treated as a '''[or]''' rather than an '''[and]'''.
** '''[abilities]''': A subtag that takes an ability definition, for example a macro such as '''{ABILITY_ILLUMINATES}'''. Strictly speaking, it needs a subtag with an '''id''' attribute, and it matches any ability with that ID, regardless of the tag name.
** '''[filter_ability]''': {{DevFeature1.17|10}} A subtag that takes a list of attributes, and matches if those attributes are present in the ability; see [[FilterWML#Filtering_Abilities|FilterWML]].
```

And in FilterWML:

```
=== Filtering Abilities ===

These keys are used as filter input:

* '''tag_name''': matches the [[AbilitiesWML#The_.5Babilities.5D_tag|subtag's name]], such as '''[heals]''' or '''[regenerates]'''. Can be a comma-separated list.
* '''id''': matches the attribute of the same name, can be a comma-separated list
* '''overwrite_specials''': matches the attribute of the same name, can be a comma-separated list
* '''apply_to''': matches the attribute of the same name, can be a comma-separated list
* '''active_on''': matches the attribute of the same name, can be a comma-separated list
* '''affect_self''': matches the attribute of the same name
* '''affect_allies''': matches the attribute of the same name
* '''affect_enemies''': matches the attribute of the same name

'''[and]''', '''[or]''', and '''[not]''' subfilters are also supported.
```

CelticMinstrel:
> I think that's good, as there's probably a lot of UMC using it, but is that OK for you?

I'm explaining in more detail on IRC, but the short answer is: no, it's not really okay for me.

data/core/macros/utils.cfg
```diff
@@ -535,6 +535,54 @@
...
+                                [/or]
+                            [/and]
+                            [not]
+                                apply_to=opponent,defender
+                            [/not]
+                        [/filter_abilities]
+                    [/effect]
+                [/object]
+            [/do]
+        [/foreach]
```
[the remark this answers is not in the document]
newfrenchy83: squah commits in two only, one for harcoding and testing and one for fix FORCE_CHANCE_TO_HIT

src/units/unit.cpp
newfrenchy83: [filter_abilities]could filter more what one ability, i willl let the actual name.

newfrenchy83:
> > I think that's good, as there's probably a lot of UMC using it, but is that OK for you?
> 
> I'm explaining in more detail on IRC, but the short answer is: no, it's not really okay for me.

Me too, i don't like the problem of subtags not resolved but i don't know how handle deprecation and i don't want PR blocked more. i prefer report this discussion for another PR.

src/units/unit.cpp
newfrenchy83: changed to [filter_ability]

newfrenchy83:
> > I think that's good, as there's probably a lot of UMC using it, but is that OK for you?
> 
> I'm explaining in more detail on IRC, but the short answer is: no, it's not really okay for me.

@CelticMinstrel i see  2 option:

1) let filter tag like it is actually.

2) filter tags in [abilities] tag without pull message deprecation. if message show each time what tags in remove_ability[abilities] don't match with one of abilities checked, the wml devellopperwill believe what using [abilities] tag is deprecated.

CelticMinstrel:
Unfortunately, I don't understand what you're saying… so I can't figure out what the two options you're proposing are.

newfrenchy83:
> Unfortunately, I don't understand what you're saying… so I can't figure out what the two options you're proposing are.

I will try to rephrase,

either the ability tag is only checked when [filter_ability] is used as is the case in this PR, but the correspondence problem between the ability tag in remove_ability[abilities] is not solved, this which we don't like.

Either we solve the problem but without introducing a depreciation message beforehand because implemented as I had tried to do, this led to the appearance of the message even when the verified ability has nothing to do with the one sought, which implies that wml developers will get bugs if they use tags that don't match those of the abilities to be removed.

newfrenchy83:
this PR could be merged before 1.18 version, please?

newfrenchy83:
@Vultraz could you review this PR please?

Pentarctagon:
I'm not sure what vultraz has to do with this PR?

stevecotton:
Sorry that I haven't looked at this recently, but given @CelticMinstrel's "no" then the next input needs to come from him.

newfrenchy83:
@CelticMinstrel i fixed tag name not checked in [abilities] but without message.

src/units/unit.cpp
```diff
@@ -2116,9 +2269,12 @@ void unit::apply_builtin_effect(std::string apply_to, const config& effect)
 	} else if(apply_to == "remove_ability") {
 		if(const config& ab_effect = effect.child("abilities")) {
 			for(const config::any_child ab : ab_effect.all_children_range()) {
-				remove_ability_by_id(ab.cfg["id"]);
+				remove_ability_by_attribute(ab.cfg, false, ab.key);
```
CelticMinstrel, line 2272: Ultimately, if we're adding a `[filter_ability]` tag that is intended to replace the old method, I think the best approach is to keep this line unchanged and retain the old `remove_ability_by_id` function, and add a deprecation message if the `[abilities]` tag exists. If I'm understanding the code correctly, this would allow you to remove the `split_filter_as_list` parameter, right? That would also make the code easier to read.

I also have one other concern about the code readability, but I'll hold off on that for now. I'll mention it later.

newfrenchy83: it's done except for deprecation message

newfrenchy83: message added

data/core/macros/utils.cfg
```diff
@@ -535,6 +535,54 @@
...
+        [foreach]
+            array=force_remove
+            [do]
+                [object]
+                    id=forced_re
+                    [filter]
+                        id=$force_remove[$i].id
+                    [/filter]
+                    silent=yes
+                    duration=turn end
```
stevecotton, line 562: What if another unit attacks during the same turn? I haven't tested this, just wondering if this is a possible bug:

* Alice attacks Bob
* Charlie attacks Bob, but lingering effects from Alice's attack change Charlie's chance to hit.

newfrenchy83: But objectif rénover in attack end event

data/core/macros/utils.cfg
```diff
@@ -535,6 +535,54 @@
...
+                    take_only_once=no
+                    [effect]
+                        apply_to=remove_ability
+                        [filter_ability]
+                            tag_name=chance_to_hit
```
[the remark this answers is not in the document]
newfrenchy83: You speak of tag_name, right?

src/units/unit.cpp
```diff
@@ -2115,10 +2220,14 @@ void unit::apply_builtin_effect(std::string apply_to, const config& effect)
 		}
 	} else if(apply_to == "remove_ability") {
 		if(const config& ab_effect = effect.child("abilities")) {
+			deprecated_message("remove_ability[abilities][tag name]", DEP_LEVEL::INDEFINITE, "", "Use the remove_ability[filter_ability] instead");
```
CelticMinstrel, line 2223:
```suggestion
			deprecated_message("[effect]apply_to=remove_ability [abilities]", DEP_LEVEL::INDEFINITE, "", "Use [filter_ability] instead");
```

CelticMinstrel:
Hmm… my other concern was having several versions of `matches_if_present_in_filter`, but it looks like there are only two now…? But that said, please rename:

* `matches_if_present_in_filter` → `bool_matches_if_present`
* `string_matches_if_present_in_filter_split` → `string_matches_if_present`

I was contemplating the idea of just using a single template function for this purpose, but that doesn't seem to fit now.

data/core/macros/utils.cfg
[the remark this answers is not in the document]
newfrenchy83: @stevecotton except what i want use code for add logic in overwrite_specials attribute and what tag_name can be applied to both abilities and specials, i don't rename this attribute

src/units/unit.cpp
newfrenchy83: you don't precise what [filter_ability] must be used in [effect]apply_to=remove_ability

newfrenchy83: change done, but i added " in [effect]apply_to=remove_ability" for what develloper don'timagine what [filter_abilities] used instead of effect]apply_to=remove_ability (someone could unterstand that, but it have few probabilities)

data/core/macros/utils.cfg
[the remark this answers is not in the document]
CelticMinstrel: If you want it to be able to apply both to abilities and specials while also being as consistent as possible with unit and weapon filters, I'd recommend calling it `type`.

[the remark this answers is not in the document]
stevecotton: I'd prefer a name that's distinctive enough to easily find with `git grep`, but I'm not sure how to unify that with specials that aren't also abilities.

data/core/macros/utils.cfg
stevecotton: Thanks, I missed that detail.

data/core/macros/utils.cfg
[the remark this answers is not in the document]
newfrenchy83: 'type' is also a attribute in plague, and alos in [damag_type] who don't merged yet. tag_name is the only name who seem conveniant.

CelticMinstrel:
This is superceded by #7223, right?

newfrenchy83:
> This is superceded by #7223, right? Or #7168?

Both, but by https://github.com/wesnoth/wesnoth/pull/7223 in priority.

stevecotton:
Commenting here because I'm not sure who's subscribed to the new PRs yet, and this though isn't quite developed enough to make an issue or PR-with-unit-test about it yet: the math in abilities.cpp's `effect::effect` has interesting properties, particularly if more than one of `add`, `sub`, `multiply` and `divide` are in a single ability. I started testing because I spotted code in the current PRs that only expected one operation, so wanted to check if the abilities themselves supported more.

Testing in current master at 2fa006f7167555224ca9d5ab0f8da468ad29b481 + #6968 at 33c2e4d85e9818150c9d41e8a88913f4c9728993:

2 + 2 - 1 = 1. Happens because `sub` is a shorthand for `add` with a negative amount. If you have a weapons special with both `add` and `sub`, the `sub` removes the `add`. So for an Orcish Grunt with 2 attacks, `[attacks]add,sub=2,1` ends up with a total of 1.

3 * 2 / 3 = 2. The `multiply` and `divide` attributes happily coexist, even in a single ability. This could be useful, it would be a way to avoid floating point.

2 + 2 * 3 = 12. Addition before multiplication.

3 * 3.34 = 10. There's some support for decimal places.

3 * 3.334 = 9. There's decimal fixed point math in here, even though it passes through floating point during the implementation.

A lot of Wesnoth simply avoids floating point by using percentages and is therefore built on fixed-point math, even showing the fixed-point to the user for abilities such as `drains`; my feeling is that that's a good thing and likely to avoid OOS bugs.

PR #7051, opened 2022-10-15
1.17 - make some overlay terrain available on bluff/gulch

doofus-01:
For now, it's forests on bluffs.  Put *^Qhhf next to some forest, the sparse version of that forest grows on that bluff tile.  There should be some default version that's there when no neighboring forest exists, but currently it's just nothing.

I'm not sure what else is really needed after forest.  Fences, maybe. Rough terrain, walls, and castles aren't overlays so are a different case.


doofus-01:
If this passes CI, I'm done.  Small, realistic goals.

PR #7055, opened 2022-10-17
Fix 'attacks' abilities don't work(1.16)

newfrenchy83:


stevecotton:
IIUC, this fixes a bug that was introduced in 05b2ea2262885af85800361423003aa241df56fe, or 1.15.10; therefore the bug is already in 1.16.0 through 1.16.6.

The change would cause out-of-sync errors if someone was playing 1.16.6 and another player had a fixed 1.16.x. That means it shouldn't be backported to the stable branch - causing OOS is a bigger problem than the bug that this would fix.

PR #7064, opened 2022-10-25
Show the log directory in the build info on non-windows platforms.

Pentarctagon:
Also moves the last bit of potentially common code from log_windows into the base logger.

---

To be honest I'm tempted to tinker with it more to see if any more of the windows-specific code needs to be windows-specific, but I also don't have a Windows machine to actually test it out. Also a lot of it is around that weird console behavior windows has.

@Vultraz Can you build this PR and just make sure that the Windows logging to file still works?

### DIFF
--- a/src/gui/dialogs/game_version_dialog.hpp
+++ b/src/gui/dialogs/game_version_dialog.hpp
@@ -65,9 +65,8 @@ private:
 
 	std::map<std::string, std::string> path_map_;
 
-#ifdef _WIN32
+	/** path to current log file */
 	const std::string log_path_;
-#endif
 
 	typedef std::array<std::string, 4> deplist_entry;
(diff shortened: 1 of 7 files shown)


src/gui/dialogs/game_version_dialog.hpp
```diff
@@ -65,9 +65,7 @@ class game_version : public modal_dialog
 
 	std::map<std::string, std::string> path_map_;
 
-#ifdef _WIN32
 	const std::string log_path_;
```
stevecotton, line 68: Please either merge `log_path_` with `path_map_["logs"]`, or add some documentation about there being two similarly-named things.

Wedge009:
I checked the Windows build and it looks like file logging is still in the same location.

PR #7075, opened 2022-11-01
Fix male Dark Horse unit name

demario-w:
The male variation of the dark horse unit is using the female translatable string.
Fix by making a male specific string to allow accurate translation.


Wedge009:
Seems obvious to me.

PR #7079, opened 2022-11-01
wmllint: Ignore labels for translation if a PNG extension is found

### LINKED ISSUE #6995: wmllint: complaints about `[image]label="some_filename.png"` not being translatable
For example, the `data/campaigns/tutorial/gui/character_selection.cfg` changes proposed in #6966.

Seems a bug in wmllint, if it's not recognising these as wanting image filenames.

_Originally posted by @stevecotton in https://github.com/wesnoth/wesnoth/pull/6966#discussion_r946842999_

Wedge009:
Resolves #6995.

I'd prefer to use `pathlib.PurePath.suffix` but `os.path` is already used extensively here. Also, `os.path.isfile()` doesn't work because that actually checks for the presence of the file, not just if the string is a valid file expression.

I don't really know all the details from the discussion in #6995, I just noted that this dropped remaining 'label needs translation mark' (for `[image]`s) in master and 1.16 for me (including `units/human-princess.png~TC(1,magenta)`). This may not necessarily be the 'right' solution.

### DIFF
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -2480,8 +2480,9 @@ to be called on their own".format(filename, num))
                           % (filename, i+1, key))
                     lines[i] = lines[i].replace("=_","=")
                 if markcheck and value and not value.startswith("$") and not value.startswith("{") and not has_tr_mark and not ("wmllint: ignore" in comment or "wmllint: noconvert" in comment):
-                    # [image] uses label to specify the image path so do not mark for translation in those cases
-                    if not (key == 'label' and os.path.splitext(value)[1]):
+                    # [image] sometimes uses label to specify a reference to a PNG file
+                    # so do not mark for translation in those cases
+                    if not (key == 'label' and '.png' in value):
                         print('"%s", line %d: %s needs translation mark' \
                               % (filename, i+1, key))
                         lines[i] = lines[i].replace('=', "=_ ", 1)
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
The idea is good but I think this will cause a lot of false positives. In fact, I think nearly any string that contains a full sentence would be treated as an image with this test.

Wedge009:
I'm relying on `os.path.splitext`. Unfortunately it does seem rather simplistic, it just looks for the last occurring full-stop and assumes anything after that is the extension. So yeah, if a `label` has a full-stop present then it'll be treated as having a non-empty extension. I did a quick check against `pathlib.PurePath.suffix` and it has similar behaviour. (Although labels ending in a full-stop has `'.'` returned as the extension for `splitext` while `suffix` returns `''`.)

But then, what is the alternative? I suppose we could check the fourth-last character being a full-stop - assuming that the label is long enough - but then `units/human-princess.png~TC(1,magenta)` would be missed.

Since there are only a few instances of these false translation mark warnings, maybe we should just explicitly tell `wmllint` to ignore them?

As an aside, I noticed `label`s specifying file paths for `[widget]`s as well as `[image]`s.

Wedge009:
I did a search for all instances of `label` containing a reference to an image and as best as I could tell, only `.png` files are used. So I changed the condition to check against that instead. I think this will do the job - comparing results of `wmllint` against 1.16, false positives against image references in `data/campaigns/Secrets_of_the_Ancients/gui/zombie_recruit_dialog.cfg` and `data/campaigns/tutorial/gui/character_selection.cfg` are dropped.

stevecotton:
I think it should support ".webp" too; but it might be better to merge as-is, backport to 1.16 and then add ".webp" for 1.17 only.

Wedge009:
I did consider `webp`, as well, but didn't find any instances of that - to date.

PR #7083, opened 2022-11-01
DM S21: Show Delfador crossing at the Ford of Abez

### LINKED ISSUE #7081: DM: Clash at the Manor - Delfador takes indirect route

- **Version:** 1.17.9
- **Downloaded from:** N/A
- **Build info:** GCC 11
- **OS:** Devuan GNU/Linux Daedalus

### Description of the bug

In "DM: Clash at the Manor", Delfador is supposed to be riding back to Weldyn with all haste. Yet - he circles Gryphon mountain from the _west_ rather than taking East and going south via Tath.

This is a similar issue to #7067 .

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Wedge009:
Before:
before

After:
after

For me, the problem isn't so much the indirectness of the existing route. I can only imagine that the path through Tath would be better because the marked route to Dan'Tonk implies an established road that would be more expedient to traverse. If we're going nit-pick, however, I'd argue the bigger problem is how Delfador and company magically (perhaps literally) cross the Great River directly from Wesmere Forest. Instead, they ought to cross at the Ford of Abez.

Resolves #7081


knyghtmare (approved):
Looks good to me

Issue #7088, opened 2022-11-05
"Chance to hit: " not translated on any language


Toranks:
### Language

All

### Source

Mouse over weapon name on attack dialog

### Original Text

\src\gui\dialogs\unit_attack.cpp lines 174 and 185 has "Chance to hit: " but this isn't translated on any language

### Proposed Change

Add "Chance to hit: " to .po files on all languages.

Must be same translation as 
data/gui/window/attack_predictions.cfg:94
msgid "Chance to hit"

Pentarctagon:
As a side note, you can link directly to the lines:
https://github.com/wesnoth/wesnoth/blob/master/src/gui/dialogs/unit_attack.cpp#L185
https://github.com/wesnoth/wesnoth/blob/master/data/gui/window/attack_predictions.cfg#L94

stevecotton:
Oops, sorry that I missed this when merging #6462. The fix IIUC is to use the `N_` macro, but that needs some checks about which textdomain the strings end up in. Should probably also move the trailing whitespace out of the string.

This is only in 1.17, in a file which is currently being actively worked on (@newfrenchy83 has 3 or 4 PRs in progress), but this particular part of that file is probably stable.

newfrenchy83:
> Oops, sorry that I missed this when merging #6462. The fix IIUC is to use the `N_` macro, but that needs some checks about which textdomain the strings end up in. Should probably also move the trailing whitespace out of the string.
> 
> This is only in 1.17, in a file which is currently being actively worked on (@newfrenchy83 has 3 or 4 PRs in progress), but this particular part of that file is probably stable.

resolved in https://github.com/wesnoth/wesnoth/pull/7091

PR #7090, opened 2022-11-05
add "Chance to hit: " in wesnoth.pot and po files

newfrenchy83:
resolve https://github.com/wesnoth/wesnoth/issues/7088 issue


Pentarctagon:
See @stevecotton's comment please. Directly adding it to these files is not the solution.

PR #7091, opened 2022-11-05
Fix "Chance to hit" not translated in tooltip

newfrenchy83:
Fix https://github.com/wesnoth/wesnoth/issues/7088 issue


Wedge009:
Code change makes sense to me and it seems to resolve the issue. Thank you.

PR #7101, opened 2022-11-07
[Core/Macros] [1.17] [FR] add new macro for unit advancement sound event

### LINKED ISSUE #7085: Add a default sound effect for when units advance

One of the really satisfying little details about Fire Emblem games is the little jingle that plays when one of your units levels up; Wesnoth should have something similar for advancements. It doesn't even have to be a new one; one of the existing sound effects (e.g. "`fanfare-short.wav`") could just be used. I included the word "default" in the title because I was imagining that UMC devs could customize it by specifying a different sound effect in UnitTypeWML. Anyways, this is pretty minor; feel free to give it the "Low Priority" label.

knyghtmare:
Closes #7085

### DIFF
--- a/data/core/macros/scenario-utils.cfg
+++ b/data/core/macros/scenario-utils.cfg
@@ -82,3 +82,28 @@ Gg, Gg, Gg, Gg
         default=yes
     [/difficulty]
 #enddef
+
+#define UNIT_ADVANCEMENT_AUDIO_EVENT FILTER_WML SOUND_FILE
+    # Plays a sound {SOUND_FILE} whenever a unit advancements
+    # {FILTER} determines which side/unit-type
+    # 
+    # Example:
+    # {UNIT_ADVANCEMENT_AUDIO_EVENT (
+    #    [filter]
+    #        side=1
+    #    [/filter]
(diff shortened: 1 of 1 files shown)


knyghtmare:
lemme guess...the 2 failed checks are wmlindent isn't it?

soliton-:
Adding unused macros to mainline does not sound like a good idea.

knyghtmare:
I simply added the desired FR. If it will be used in UMC, it's not a bad idea. Also, some mainline campaigns can use it perhaps as well.

soliton-:
Unused macros will just break without anyone noticing until after a release.

Pentarctagon:
IMO this should be applied to mainline units, otherwise it's simple enough that UMC authors should be able to implement it themselves if they want it.

CelticMinstrel:
This seems like the sort of thing that should be applied at the engine level (with a preference to disable it), rather than doing something in WML.

knyghtmare:
> engine level (with a preference to disable it)

I think that's the ideal solution. Shall I close this? or delegate the event macro to a modification addon for the time being?

PR #7102, opened 2022-11-07
1.17 - standing animations for zombies

doofus-01:
For now, just the default WC.  I'm not sure which variations are most important, maybe `mounted`, `drake`, and `troll`?


doofus-01:
Just the default and mounted variation for now to see how it goes, other variations can be added later.

PR #7109, opened 2022-11-11
Unit traits exclusive

Kingofd:
Please, for your convenience, test it with a build, trait config file included. Do not include trait config in merge, tho, only for testing


CelticMinstrel:
You will need to update the schema to add the new keys.

data/core/macros/traits.cfg
```diff
@@ -143,6 +143,7 @@ While useful for any close-combat unit, strong is most effective for units who h
             apply_to=hitpoints
             increase_total=1
         [/effect]
+        require_traits=quick
```
knyghtmare, line 146: Why is trait strong requiring this?

data/core/macros/traits.cfg
```diff
@@ -180,6 +181,7 @@ Quick is the most noticeable trait, particularly in slower moving units such as
             apply_to=hitpoints
             increase_total=-5%
         [/effect]
+        exclude_traits=intelligent
```
knyghtmare, line 184: I am not exactly sure why the core traits file is modified to ensure the combination `intelligent,quick` is never picked.

data/core/macros/traits.cfg
```diff
@@ -217,6 +219,7 @@ Resilient units can be useful at all stages of a campaign, and this is a useful
             times=per level
             increase_total=1
         [/effect]
+        exclude_traits=intelligent
```
knyghtmare, line 222: Is there anything negative about the combo `resilient, intelligent`?

knyghtmare (changes requested):
I think there's some things which need to be clarified. Commented them separately.

knyghtmare:
> Do not include trait config in merge, tho, only for testing

Oh I see. This is just testing to see if the following happen/not-happen.

I can build over this and test it in the weekend. So, can you name the expected observations to compare with?

Kingofd:
It might be helpful to also look at the case of `strong` and `quick`
Since `strong` needs `quick`, it can only ever occur, that strong has been picked as the second trait, so always `quick, strong` never the other way round

src/units/unit.cpp
```diff
@@ -806,53 +806,114 @@ void unit::generate_traits(bool must_have_only)
...
+			bool already = false;
+			for(const config& mod : current_traits) {
+				if(mod["id"] == tid) {
+					already = true;
+					break;
+				}
+			}
+
+			if(already)
+				continue;
```
soliton-, line 848: Doesn't look like we have an explicit rule about it but I would either write this on one line or use curly braces. The way it is now you might just add an indented statement and expect it to be behind the if...

Pentarctagon:
Is this something that could have one or more WML unit tests added for it?

src/units/unit.cpp
```diff
@@ -806,53 +806,114 @@ void unit::generate_traits(bool must_have_only)
...
 	const unit_type& u_type = type();
 
-	// Calculate the unit's traits
 	config::const_child_itors current_traits = modifications_.child_range("trait");
-	std::vector<const config*> candidate_traits;
 
+	// Handle must-have only at the beginning
 	for(const config& t : u_type.possible_traits()) {
-		// Skip the trait if the unit already has it.
-		const std::string& tid = t["id"];
```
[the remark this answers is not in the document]
Kingofd: You are right, there needs to be a filter for that.

src/units/unit.cpp
```diff
@@ -806,53 +806,114 @@ void unit::generate_traits(bool must_have_only)
...
+			for(const config& mod : current_traits) {
+				if (mod["exclude_traits"] != "") {
+					for (const auto& c: utils::split(mod["exclude_traits"])) {
+						temp_exclude_traits.push_back(c);
+					}
+				}
+			}
+
+			std::sort(temp_require_traits.begin(), temp_require_traits.end());
+			std::sort(temp_exclude_traits.begin(), temp_exclude_traits.end());
```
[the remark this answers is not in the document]
Kingofd: It was for an early prototype that would allow for duplicate traits. Needless to say it now is not required, as every trait will only ever be present once on a unit

src/units/unit.cpp
Kingofd: I will make sure to include braces, whenever I use if's. I personally find it rather appealing + less cluttering -> easier to read, so changing the code-style might be a nice idea

Kingofd:
I was thinking the same thing. Where should I start with exploring the current test infrastructure?

I guess these test cases should work:

1. Test trait for exclusion ("not with")
2. Test trait for inclusion ("not without")

Anything else that should be included in the new tests?

Pentarctagon:
The current tests are in the `data/test/scenarios/` folder. They are executed by the `run_wml_tests` python script, with the list of tests and expected results in the `wml_test_schedule`. They can be run individually with either `wesnoth -u <test_name>` or `wesnoth -u <test_name> --showgui`.

Pentarctagon:
Tests aren't actually run unless they're added to `wml_test_schedule`.

data/core/macros/traits.cfg
```diff
@@ -374,3 +374,61 @@ Dim is a trait all too common in goblins and other lesser species. There are rea
...
+        id=test_excluded
+        male_name= _ "test_excluded"
+        female_name= _ "female^test_excluded"
+        help_text= _ "Test excluded"
+        [effect]
+            apply_to=hitpoints
+            increase_total=1
+        [/effect]
+    [/trait]
+#enddef
```
Pentarctagon, line 434: All of these `#define`s should be part of the unit test cfg file.

data/core/units/fake/Test_Unit_Require.cfg
```diff
@@ -0,0 +1,24 @@
...
+	movement=1
+	experience=1
+	level=1
+	alignment=neutral
+	advances_to=null
+	cost=1
+	usage=scout
+	hide_help=yes
+	do_not_list=yes
+[/unit_type]
```
Pentarctagon, line 24: Both of these unit types should also be located in the `data/test/` directory somewhere.

Kingofd: I have made a new folder for test units in general. Where do I add them to the environment, the tests can't seem to find them now.

data/schema/units/modifications.cfg
```diff
@@ -217,6 +217,8 @@
 	{SIMPLE_KEY male_name t_string}
 	{SIMPLE_KEY female_name t_string}
 	{SIMPLE_KEY description t_string}
+	{SIMPLE_KEY exclude_traits string_list}
+	{SIMPLE_KEY require_traits string_list}
```
Kingofd, line 221: My bad, the units should not have that keyword, it was wrongfully still in the unit file, I'll revert the change. The feature is strictly trait-macro bound, the units should not notice anything.

data/core/units/fake/Test_Unit_Require.cfg
Pentarctagon: For regular units, you'd need to add an include for them in https://github.com/wesnoth/wesnoth/blob/master/data/core/units.cfg

For test-only units, I believe you should be able to include them similarly in https://github.com/wesnoth/wesnoth/blob/master/data/test/_main.cfg

Kingofd: Added test unit capability to test textdomain :+1:

CelticMinstrel:
Is there any remaining reason to hold off on merging this? It seems okay to me.

Pentarctagon:
No objections from me.

data/test/_main.cfg
```diff
@@ -74,6 +74,7 @@
 {test/scenarios/wml_tests/WesnothFormulaLanguage}
 
 # Load test unit wml
+{test/scenarios/traits_exclude_include.cfg}
```
CelticMinstrel, line 77: I think the correct fix is instead to move this file into a different folder. @Pentarctagon should be able to help you figure out where it goes.

wml_test_schedule
```diff
@@ -345,6 +345,11 @@
 0 special_calculation_sub3_sub2_cumulative
 0 test_add_in_leadership_abilities
 0 test_sub_in_leadership_abilities
+<<<<<<< HEAD
+0 trait_exclusion_test
+0 trait_requirement_test
+=======
```
CelticMinstrel, line 351: You need to remove the merge markers here.

Pentarctagon:
`TRAIT_TEST_EXCLUDE`, `TRAIT_TEST_EXCLUDED`, `TRAIT_TEST_REQUIRE`, `TRAIT_TEST_REQUIRED` don't seem to be defined anywhere, so that's why the tests are failing.

CelticMinstrel:
I guess they were accidentally deleted in 584d53bd963c5c33033f979c3c7b900fcbd829bf?

data/test/_main.cfg
Kingofd: I was thinking the same thing, where should I put it @Pentarctagon ?

Kingofd:
Hmm, that's weird. I initially put them into the test file, however the test failed with the same error. Now I put it in its own file in `wesnoth/data/test/macros/traits.cfg`, where it should be loaded once the TEST environment is loaded, akin to the main `traits.cfg`.

I guess back to the test file it is (but then it needs to also be available for the `units.cfg`)?
Its also weird because it is a newer error, it did not appear in earlier iterations of this feature.

CelticMinstrel:
I'd also like to request that the new files be renamed to all lower-case.

data/test/_main.cfg
Pentarctagon: [trait] is under UnitsWML on the wiki, so it'd be https://github.com/wesnoth/wesnoth/tree/master/data/test/scenarios/wml_tests/UnitsWML

data/test/scenarios/traits_exclude_include.cfg
```diff
@@ -0,0 +1,77 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: exclude [trait] functionality
```
Pentarctagon, line 4: `exclude [trait] functionality` should be `[trait]exclude_traits=`

data/test/scenarios/traits_exclude_include.cfg
```diff
@@ -0,0 +1,77 @@
...
+				{ASSERT ({VARIABLE_CONDITIONAL excluded.modifications.trait[0].id equals "test_exclude"})}
+				{ASSERT ({VARIABLE_CONDITIONAL excluded.modifications.trait[1].id equals $nil})}
+			[/else]
+		[/if]
+		{SUCCEED}
+	[/event]
+)}
+
+#####
+# API(s) being tested: require [trait] functionality
```
Pentarctagon, line 47: `require [trait] functionality` should be `[trait]require_traits=`

Pentarctagon:
Hmm... WML validation failed again, but logged nothing about why.

Kingofd:
I ran the CI locally, there were some errors with the t_strings in the new test race, they should be fixed now. 

Just FYI, incidentally another error was shown specifically on my local installation: 
`Game Config failed validation! 
20230114 15:01:17 error preprocessor: Found filename containing whitespace: '_initial.cfg' in included directory 'data/schema/./filters'.`

It is not shown in the CI tests so it probably is something on my own copy.

data/test/scenarios/wml_tests/UnitsWML/traits_exclude_include.cfg
```diff
@@ -1,7 +1,7 @@
 # wmllint: no translatables
 
 #####
-# API(s) being tested: exclude [trait] functionality
+# API(s) being tested: [trait]exclude_traits= functionality
```
Pentarctagon, line 4: Remove `functionality`.

data/test/scenarios/wml_tests/UnitsWML/traits_exclude_include.cfg
```diff
@@ -44,7 +44,7 @@
 )}
 
 #####
-# API(s) being tested: require [trait] functionality
+# API(s) being tested: [trait]require_traits= functionality
```
Pentarctagon, line 47: Remove `functionality`.

data/test/units.cfg
```diff
@@ -5,12 +5,32 @@
...
 		[race]
 		id=test
-		name= "race^Test"
-		female_name= "race+female^Test"
-		plural_name= "race^Tests"
-		description= "This is a test unit race"
+		name= _ "Language"
+		female_name= _ "Language"
+		plural_name= _ "Language"
+		description= _ "Language"
```
CelticMinstrel, line 11: It's fine to use the original strings but just make them translatable. Since this file declares its textdomain to be `wesnoth-test`, nothing in here will be added to the gettext files, which don't scrape that textdomain. (This also means there's no need to use the `race^` prefix.)

Kingofd: Alright I deleted the prefixes so it should be fine now

Kingofd:
So after this is merged, where would I (or who) update the WML documentation?

Pentarctagon:
It's usually preferred that the person making the change also update the wiki documentation to reflect that change. So you'd update https://wiki.wesnoth.org/UnitsWML#.5Btrait.5D to add the two new attribute along with the note "1.17.13 or later" (similar to the `help_text` attribute).

Kingofd:
So, there are some HTTP error 5xx in the CI builds?

stevecotton:
Sadly we do sometimes get network errors on the CI, I'll rerun them.

CelticMinstrel:
By the way, the email on your commits does not seem to match the email that GitHub has for you, so you may want to either update your git config for the Wesnoth repository or let GitHub know that that's your email.

stevecotton:
Please document the new WML on the wiki.

PR #7112, opened 2022-11-11
Make Nightblade the default advancement for Slayer

Jonathan-Kelly:
Similarly to #6911 this gives a level 3 advancement to a default unit that is still missing one. Except in this case the advancement was already in core, just not fully utilized.

Also adds the Orcish Nightblade as a leader for Age of Heroes era and to campaign side recruit lists wherever the other level 3 orcish units are found.

### DIFF
--- a/data/core/macros/deprecated-utils.cfg
+++ b/data/core/macros/deprecated-utils.cfg
@@ -386,3 +386,7 @@ _"No gold carried over to the next scenario."#enddef
     # no interruption of movement, yet prevent the move from being undone.
     {DEPRECATE_ACTION_MACRO "NO_INTERRUPT_NO_UNDO" 3 "1.16" ""}
 #enddef
+
+#define ENABLE_NIGHTBLADE
+#deprecated 2 1.17 This advancement is now enabled by default.
+#enddef
--- a/data/core/macros/optional_unit_advancements.cfg
+++ b/data/core/macros/optional_unit_advancements.cfg
@@ -57,11 +57,6 @@
     {ENABLE_ADVANCEMENT "Great Wolf" "Direwolf" (set_experience=65)}
 #enddef
 
-#define ENABLE_NIGHTBLADE
-    # Place in a campaign or scenario definition to allow Orcish Slayer to advance to Orcish Nightblade.
-    {ENABLE_ADVANCEMENT "Orcish Slayer" "Orcish Nightblade" (set_experience=100)}
-#enddef
-
 #define ENABLE_TROLL_SHAMAN
     # Place in a campaign or scenario definition to allow Troll Whelp to advance to Troll Shaman.
     {ENABLE_ADVANCEMENT "Troll Whelp" "Troll Shaman" ()}
--- a/data/campaigns/Dead_Water/scenarios/05_Tirigaz.cfg
+++ b/data/campaigns/Dead_Water/scenarios/05_Tirigaz.cfg
@@ -115,7 +115,7 @@
         canrecruit=yes
         moves=0
         type=Orcish Sovereign
-        recruit=Wolf Rider, Goblin Knight, Goblin Pillager, Orcish Archer, Orcish Grunt, Orcish Crossbowman, Orcish Assassin, Orcish Warrior, Orcish Slurbow, Orcish Warlord, Orcish Slayer
+        recruit=Wolf Rider, Goblin Knight, Goblin Pillager, Orcish Archer, Orcish Crossbowman, Orcish Slurbow, Orcish Assassin, Orcish Slayer, Orcish Nightblade, Orcish Grunt, Orcish Warrior, Orcish Warlord
 
         [ai]
             [goal]
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter2/07_Elves_Last_Stand.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter2/07_Elves_Last_Stand.cfg
@@ -235,7 +235,7 @@
             x=6
             y=23
 #ifdef HARD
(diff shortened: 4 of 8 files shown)


data/core/macros/optional_unit_advancements.cfg
```diff
@@ -57,11 +57,6 @@
     {ENABLE_ADVANCEMENT "Great Wolf" "Direwolf" (set_experience=65)}
 #enddef
 
-#define ENABLE_NIGHTBLADE
```
knyghtmare, line 60: I am not sure deleting a core macro just out of the blue is a good approach. There are UMC authors using it still. So their campaigns will break instantly. Add a dummy macro here so their works are not hampered.

knyghtmare (changes requested):

data/core/macros/optional_unit_advancements.cfg
knyghtmare: Or, rather move it to the deprecated section.

max-torch:
After looking at the "Files Changed" section, the only major balance change seems to be for the campaign 'Son of the Black Eye'.
The effect is not small - you get two loyal orc assassins in the campaign, one a quarter of the way through and one halfway through, and there are no other times where you can recruit orc assassins. However, I am in favor of the change because there is a lot of feedback in the SotBE threads complaining about these orc assassins (too easy for them to end up dead, not useful enough) and, well, I tend to agree with it and also the Orc Nightblade just seems cool I guess.
**Focus on the fact that the two units are loyal throughout the entire campaign (imagine bringing two Nightblades around every scenario for half of a campaign) to really grasp the weight of this balance change.** You are more strongly rewarded for taking care of two loyal units.

The effect on balance is a lot smaller for all the other campaigns because you don't really get to have an orc assassin that you recall throughout the entire campaign

knyghtmare:
> one a quarter of the way through and one halfway through, and there are no other times where you can recruit orc assassins

https://github.com/wesnoth/wesnoth/blob/4edc8ef73eca3f468eee03e72a59c6d31e01435f/data/campaigns/Son_Of_The_Black_Eye/scenarios/04_The_Siege_of_Barag_Gor.cfg#L481-L484

You seem outdated...By a considerable margin. Replay SotBE.

max-torch:
> > one a quarter of the way through and one halfway through, and there are no other times where you can recruit orc assassins
> 
> https://github.com/wesnoth/wesnoth/blob/4edc8ef73eca3f468eee03e72a59c6d31e01435f/data/campaigns/Son_Of_The_Black_Eye/scenarios/04_The_Siege_of_Barag_Gor.cfg#L481-L484
> 
> You seem outdated...By a considerable margin. Replay SotBE.

Thanks. Yes, I was outdated. Are they recruitable in any of the future scenarios. Okay then I guess the only balance consideration is the fact that you get to have two loyal assassins.

knyghtmare:
> Are they recruitable in any of the future scenarios.

Well, yes. Once they are added to your recruitlist, they will always be available for recruitment.

I would also like to add that I am in favour as well. Just the `ENABLENIGHTBLADE` should be deprecated and not be deleted entirely.

Pentarctagon:
@max-torch This would also enable them for MP, right? Or did I misread the change?

knyghtmare:
The PR makes Nightblade a leader of the AoH Northerners faction, and allows Slayers to advance to them in MP and SP.

data/core/macros/optional_unit_advancements.cfg
Jonathan-Kelly: Alright, moved it to deprecated_utils.

Put it at the end of the file; let me know if you think there is a better spot for it within deprecated-utils.

knyghtmare (approved):

data/core/macros/optional_unit_advancements.cfg
knyghtmare: That's okay. Are you sure it wont throw any error? like the deprecated-utils.cfg is read before the advancements one so the two generalised advancement/disadvancement macros aren't read before.

max-torch:
> @max-torch This would also enable them for MP, right? Or did I misread the change?
kyght in the above comment answers your question. It was the right call that you asked hejne to weigh in on this because MP will be affected.

data/core/macros/deprecated-utils.cfg
```diff
@@ -386,3 +386,8 @@ _"No gold carried over to the next scenario."#enddef
     # no interruption of movement, yet prevent the move from being undone.
     {DEPRECATE_ACTION_MACRO "NO_INTERRUPT_NO_UNDO" 3 "1.16" ""}
 #enddef
+
+#define ENABLE_NIGHTBLADE
+    # Place in a campaign or scenario definition to allow Orcish Slayer to advance to Orcish Nightblade.
+    {ENABLE_ADVANCEMENT "Orcish Slayer" "Orcish Nightblade" (set_experience=100)}
```
CelticMinstrel, line 392: This is wrong. If the advancement is enabled by default, then this macro should just be empty (but with a `#deprecated` statement so there's a warning when someone uses it).

Jonathan-Kelly: Alright, how does it look now? Deprecation level 2 is good?

data/core/macros/optional_unit_advancements.cfg
Jonathan-Kelly: I did not see any errors in the log files. But now that I think about it, there are no calls to ENABLE_NIGHTBLADE anywhere in current mainline so any could-be errors may not have been thrown. I applied the fix suggested by CelticMinstrel below.

data/core/macros/deprecated-utils.cfg
CelticMinstrel: Level 2 is typically fine, yeah. Looks good.

Jonathan-Kelly: Cool, thanks for your help.

Hejnewar:
100 xp is way too high, get it down to 64 and it will be fine.

Jonathan-Kelly:
@Hejnewar  Okay, reduced Slayer experience to 64.

Let me know if you would also like a reduction to the Nightblade (AMLA) experience, which is currently set to 150.

nemaara (approved):

Hejnewar (approved):

knyghtmare:
3 approvals? will merge in the morning then.

PR #7119, opened 2022-11-13
Slightly reduce merfolk forest and hill move cost

Jonathan-Kelly:
From 5 to 4. Makes moving them from one water body to another somewhat less tedious. Likewise makes merfolk units a bit more viable on maps with separated waterways.

Just as it was before this change, the merfolk remain twice as slow as naga at moving through these two terrains.


nemaara (approved):
Fine with me, needs Hejenwar's approval.

Hejnewar (approved):

PR #7124, opened 2022-11-14
Disable saving of replays for story-only scenarios

### LINKED ISSUE #6180: Don't allow story scenario replays to take you to the next scenario automatically at the end
 - 1.15.18
 - Steam
 - Windows 10

Currently, when you load a _story scenario replay_, at the end of the scenario it loads the next scenario in the campaign. This is the exact same thing that occurs when you load a _start-of-scenario save_ of a _story scenario_.

The behavior that I was expecting, however, was that, at the end of a _story scenario replay_, the next scenario would not load.

### Screenshots
A start-of-scenario save and a replay of a story scenario has the exact same behavior.
image

### Rationale
Changing story scenario replays to behave like this would make it consistent with any other replay files that you would load, which do not automatically begin loading the next scenario once the end of the replay has been reached. It also gives more of a reason for the two different story scenario files, the replay file and start-of-scenario save, to exist, because there would now be an actual difference between them.

### LINKED ISSUE #7123: [Campaign][TRoW][UX] Patch Cutscenes to prevent generation of replay saves
The following Cutscene scenarios of TRoW have to be patched with `replay_save=no` inside `[endlevel]`.

- [ ] Return to Oldwood
- [ ] Southbay in Winter
- [ ] The Kalian
- [ ] Spy in the Woods
- [ ] The Plan

### Context:

Replay Saves for cutscene are not required as the replay functionality is not usable in it, as cutscenes start and end with a start event.

Wedge009:
Resolves #6180.
Resolves #7123.

### DIFF
--- a/data/campaigns/Dead_Water/scenarios/11_Getting_Help.cfg
+++ b/data/campaigns/Dead_Water/scenarios/11_Getting_Help.cfg
@@ -189,6 +189,7 @@
             result=victory
             linger_mode=no
             carryover_report=no
+            replay_save=no
             {NEW_GOLD_CARRYOVER 100}
         [/endlevel]
     [/event]
--- a/data/campaigns/Delfadors_Memoirs/scenarios/01_Overture.cfg
+++ b/data/campaigns/Delfadors_Memoirs/scenarios/01_Overture.cfg
@@ -219,6 +219,7 @@
             result=victory
             linger_mode=no
             carryover_report=no
+            replay_save=no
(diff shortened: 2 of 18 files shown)


knyghtmare (approved):

knyghtmare:
Needs one final check by the SP-content Dev to confirm the merge.

nemaara (approved):

soliton-:
How does this fix #6180?

PR #3414, opened 2018-07-30
Move generally-useful units to core so addons can use them

CelticMinstrel:
This moves many units from campaigns to core, allowing addons to safely use them without having to copy them into their own addon directory.

The following units have been moved (or are planned):

* [x] ~Quenoth faction (all units except Youth, Corrupted Elf, and Incarnation) - note that this is a total of 21 units~
* [x] Naga Hunter and ~Naga Guardian line~ (4 units total)
* [x] Merman Citizen and Brawler
* [x] Great Ogre (with optional define to enable it advancing from ogre; NOTE this define still needs to be added to LoW's extra_defines)
* [x] ~Wose Sapling and Shaman~ (with optional define to enable the shaman advancing from wose)
* [x] ~Skeleton Rider, Bone Knight, and Death Squire (with optional define to enable the death squire advancing from skeleton)~
* [x] Dwarvish Miner
* [x] Orcish Shaman
* [x] ~(new) Riderless tauroch, if I can find a sprite for it.~

This hasn't been tested yet, but for now it's up here for general review. I'll test it at some point once I get master built (which I think just means switching out libraries).

Issue #4753, opened 2020-02-20
Unit database doesn't show variations (website)


IskarJarak:
If I understood it correctly, variations were only used for visual diversity. With the Sand Scuttler variation to the Giant Scorpion we now have a variation in core that is relevantly different (massively different resistances) from the base unit. The unit database (https://units.wesnoth.org/1.15/mainline/en_US/mainline.html) does not display this variation at all though. Adding support for variations could be helpful.

IskarJarak:
Just remembered that this would apply to walking corpses as well - I know they can all be seen as SotA-specific units, but seeing the various different movement costs and defenses by switching through the variations for the main unit could be another interesting use-case.

cooljeanius:
> Just remembered that this would apply to walking corpses as well - I know they can all be seen as SotA-specific units, but seeing the various different movement costs and defenses by switching through the variations for the main unit could be another interesting use-case.

Yeah, I'm trying to set up a Walking Corpse spawner that randomizes its variations for one of my campaigns currently, and having a good viewer for all their variations on the website would be really helpful.

PR #7125, opened 2022-11-15
Give Kraken submerge ability

Jonathan-Kelly:
Similar to #3414 this moves the Kraken from DW and WoF into core as the level 3 advancement for the Cuttle Fish.

The cuttlefish has perhaps the most sophisticated camouflage ability in the animal kingdom and no need to breathe air, so there may be no better fit for the submerge stealth ability than this unit.

Slightly increases the gold costs for both units to compensate (just let me know if this or experience needs further adjustment). Tweaks campaign scenarios sufficiently affected due to having large areas of deep water terrain and multiple individuals of these unit types, Note that DW S6 is partly counter balanced by the increased Cuttle Fish cost.

### DIFF
--- a/data/campaigns/Winds_of_Fate/scenarios/04_Journey.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/04_Journey.cfg
@@ -404,9 +404,18 @@ I must advise against it."
             name=water-blast.wav
         [/sound]
         {LOYAL_UNIT 4 Kraken        3  3} {FACING se}
+        [+unit]
+            animate=yes
+            [status]
+                uncovered=yes
+            [/status]
+        [/unit]
         {NAMED_LOYAL_UNIT 4 Kraken  1 19 $kraken_id $kraken_name} {FACING se}
         [+unit]
             animate=yes
+            [status]
+                uncovered=yes
+            [/status]
         [/unit]
         [message]
             speaker=$kraken_id
@@ -463,6 +472,9 @@ I must advise against it."
             name=water-blast.wav
         [/sound]
         {LOYAL_UNIT 3 (Sea Serpent)       21  2} {FACING sw}
+        [+unit]
+            animate=yes
+        [/unit]
         {NAMED_LOYAL_UNIT 3 (Sea Serpent) 25 20 $serpent_id $serpent_name} {FACING sw}
         [+unit]
             animate=yes
--- a/data/campaigns/Winds_of_Fate/scenarios/04_Journey.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/04_Journey.cfg
@@ -77,6 +77,9 @@
             type=Merman Warrior
             facing=sw
         [/leader]
+        [ai]
+            aggression=0.7
+        [/ai]
     [/side]
 
     [side]
@@ -95,7 +98,7 @@
             facing=sw
         [/leader]
         [ai]
-            aggression=0.6
+            aggression=0.7
             passive_leader=yes # So he does not get himself killed.
         [/ai]
     [/side]
@@ -116,7 +119,6 @@
             facing=se
         [/leader]
         [ai]
-            aggression=0.6
             passive_leader=yes # So he does not get himself killed.
         [/ai]
     [/side]
--- a/data/core/units/monsters/Kraken.cfg
+++ b/data/core/units/monsters/Kraken.cfg
@@ -15,11 +15,14 @@
     advances_to=null
     attacks=1
     {AMLA_DEFAULT}
-    cost=62
+    cost=64
     undead_variation=swimmer
     usage=fighter
     description= _ "Krakens are gigantic creatures of the seas. They can grab their opponents with strong tentacles, or spit a poisonous black ink from a distance. The best way to survive an encounter with these monsters is to remain ashore."
     die_sound=water-blast.wav
+    [abilities]
+        {ABILITY_SUBMERGE}
+    [/abilities]
     {DEFENSE_ANIM "units/monsters/kraken/kraken-defend2.png" "units/monsters/kraken/kraken-defend1.png" squishy-hit.wav}
     [attack]
         name=tentacle


knyghtmare:
Requested Review of both content devs:
1. Several MP Survivals (Dark Forecast, Team Survival, Wilderlands) are affected
2. Several campaigns (mainline + UMC) are affected by this change.

CelticMinstrel:
I don't think the cuttlefish should advance to a kraken by default. The two are distinct species.

nemaara (commented):
I kind of agree with celmin, I don't think this being the default advancement is preferable. Additionally, I would like doofus's input on whether the kraken sprite is core worthy at this point in time. To me, it seems a bit lower quality than the cuttlefish but let's see what he says.

Hejnewar:
Kraken leveling from Cuttlefish is a no for me.
Submegre for kraken is a no for me. (Submerge is weird special technically a lot of units should have it but instead undead benefit from it the most so I guess it is more of a hiding under water than just submerging.)
Submegre for Cuttlefish is a maybe for me. Kinda makes sense but idk. 
Moving kraken core is a yes for me.

doofus-01:
I'm not thrilled with the kraken sprite as it is, especially if it is an advancement for cuttlefish (which I'm also not a fan of, but won't try to block).  I'm not saying the sprite sucks, but it doesn't look like a level 3 cuttlefish.

CelticMinstrel:
Yeah, the kraken sprite looks like a giant squid, which is an entirely different species from a cuttlefish. Even though they're both cephalopods, they're still quite different.

Jonathan-Kelly:
The real life cuttlefish looks similar to a squid hatchling (aside from the fins) so some clever writing in the unit descriptions could imply that there is scholarly disagreement between mages on whether this creature is simply a (huge) cuttlefish or actually a hatchling kraken. (This question eventually gets answered for the player when they play DW and see Inky level into a kraken).

However, there is no denying the two sprites have very different visual styles and it sounds like that is generally too undesirable for a default advancement. So I will remove the cuttlefish-to-kraken default advancement from this PR.

With that out of the way, the next question is whether the current sprite is good enough for core by itself, and it sounds like we may be split on that one.

@doofus-01 you have done an immense amount of work adding to the monster sprite set as a whole. I don't want this to be a fly in the ointment for you and all the good you have done with it. So how do you feel about the kraken sprite going into core now that it would be its own unit unrelated to the cuttle fish? Would it be adding more good than harm? Would the sprite need to be redone from scratch or just some touching up for you to be happy with it?

doofus-01:
> So how do you feel about the kraken sprite going into core now that it would be its own unit unrelated to the cuttle fish? Would it be adding more good than harm? Would the sprite need to be redone from scratch or just some touching up for you to be happy with it?

My current thinking is that it is good for a special campaign unit, however I'm reluctant to move it to mainline as-is.  It wouldn't be the worst sprite in core if it were moved, but it does need some work.  If it can add something useful to core, we should add it, so let's see where it goes in #7152?

Jonathan-Kelly:
@doofus-01

That is most helpful, thank you!

I will remove all kraken related changes from this PR then.

---

@Hejnewar
> (Submerge is weird special technically a lot of units should have it but instead undead benefit from it the most so I guess it is more of a hiding under water than just submerging.) Submerge for Cuttlefish is a maybe for me. Kinda makes sense but idk.

The thinking of the original developers was that submerge was only for non-air-breathers. So naga. merfolk, sea serpents and water serpents could not have this ability, as they all need to surface for air, which reveals where they are.

The caribe, cuttlefish and sea horse are the only water breathers in core presently, so they (along with undead which do not breathe at all) would be the only candidates for having the submerge ability. Of those three, the best at hiding would have to be the cuttlefish, due to its remarkably dynamic chromatophore camouflage.

cooljeanius:
Things I worry about with this PR:
- How it would affect campaigns where cuttlefish are enemies... regardless of flavor justifications, I can imagine that having to fight against cuttlefish with submerge would be pretty annoying from a gameplay perspective
- How it would affect the balance of campaigns where you play as cuttlefish (i.e. Inky's Quest)

CelticMinstrel:
> So naga. merfolk, sea serpents and water serpents could not have this ability, as they all need to surface for air, which reveals where they are.

Your list of air breathers seems wrong. I don't think there's any evidence that the merfolk need to surface for air, and they are typically depicted as not needing to. As for the sea serpent, it is based on deep sea fish, so I'd expect it to have gills.

> I don't think it makes sense for the cuttlefish to have it if the kraken does not.

The reason I said this is that a kraken is very much a deep sea creature, while the cuttlefish seems more of a creature of the shallow seas.

Jonathan-Kelly:
> Before this PR is merged, the survival _Dark Forecast_ might have to rechecked as there is an enemy wave in it which has cuttlefish as its bulk. So, if water map is placed and this submerge in play, then it'd a recipe for disaster.

I did. You just stay away from the river, which is the same thing you do whether or not they can submerge. _Dark Forecast_ is not a tightly balanced scenario as it instead focuses on being massively replayable. Each time is unique and you can never be sure what to expect from it next. So you just have to learn to roll with its oddball punches; if an enemy type appears while its favored terrain is widespread, then you probably should make a tactical retreat away from such terrain.

That said, _Dark Forecast_ factors in gold cost for its enemy spawns, so the amount we increase the cost of the cuttlefish (to account for its submerge ability) will likewise decrease the amount of them which appear in DF.

> I don't think there's any evidence that the merfolk need to surface for air, and they are typically depicted as not needing to.

I do not remember the merfolk ever pulling off any underwater feat which a cetacean or pinniped could not do as well. However, taking a closer look at their depictions, it appears the portrait artists gave the merfolk gill slits at the sides of the neck (about where you would expect to find them). So I think you are correct, they do not need to breathe air.

Which I guess brings us back to what Hejnewar said earier - submerge is for creatures that _hide_ well underwater.

> The reason I said this is that a kraken is very much a deep sea creature, while the cuttlefish seems more of a creature of the shallow seas.

Wesnoth's cuttlefish appears to be more of a sea monster, given its large sprite size, level 2 status, and `deepsea` move type which gives it half movement and reduced defense in shallow water (versus deep water).

And the thing with shallow water in wesnoth is it needs to be specifically somewhere between about 0.5 and 1.5 meters. Just deeper than fordable water but shallow enough that a steel clad dwarf or full plate heavy infantryman can stand with their head above water, otherwise they would drown. Wading depth, basically.

cooljeanius:
I just played through the scenario in Flight to Freedom where you have a fleet of ships and have to get past some Cuttlefish, and I'm pretty sure that would have been nightmarish if they'd had submerge; the map is almost entirely deep water.

CelticMinstrel:
> And the thing with shallow water in wesnoth is it needs to be specifically somewhere between about 0.5 and 1.5 meters.

I understand your reasoning for this, but I'm pretty sure there are also things that require it to be deeper. For example, ships can enter shallow water.

Jonathan-Kelly:
> For example, ships can enter shallow water.

I am presently working on a [replacement](https://github.com/wesnoth/wesnoth/pull/6781#issuecomment-1166839830) for the huge anachronistic galleons in the form of a much smaller, single sailed, shallow draft, medieval Cog (Kogge). It was the quintessential boat/ship of the middle ages, and really iconic looking with its wood battlements giving it the appearance and function of a floating castle. It only had a draft of 2-3 meters, and with its flat bottom, could land (beach itself) at low tide so its cargo could be transferred on foot without any dock infrastructure.

CelticMinstrel:
I think your approach is excessively "realistic". Wesnoth is a fantasy world. There's no reason why their maritime technology couldn't be ahead of their other technologies, especially considering that they canonically arrived on the continent by sea. That doesn't necessarily mean that the existing ships are the best representation of that – perhaps they're a little too advanced for the setting – but I think your approach goes too far in the opposite direction.

Jonathan-Kelly:
It may be best to save our discussion of maritime technology for the more relevant future topic. For this PR, what remains is the details of giving the cuttlefish its stealth ability...

@Hejnewar
How much do you want me to increase the gold cost for the cuttlefish to balance the submerge ability?
Or do you want something else like a reduction of its hit points?

(before adding submerge, these were set to `gold=36` and `hitpoints=67`)

cooljeanius:
> @Hejnewar How much do you want me to increase the gold cost for the cuttlefish to balance the submerge ability? Or do you want something else like a reduction of its hit points?
> 
> (before adding submerge, these were set to `gold=36` and `hitpoints=67`)

Reminder that the Cuttlefish has the "swarm" ability, so any change to its HP is also going have an affect on its melee attack

max-torch:
The whole point of this change (making Cuttlefish have submerge) seems to be to make the game more realistic. The game already makes a lot of concessions against realism (ranged attacks, villages, etc). Why is this particular change worth doing when compared to the many other unrealistic things that already exist in wesnoth?

We are weighing on one hand:
Potentially breaking the balance of multiple UMC and changing the experience of some mainline scenarios. I say changing the experience, presuming this to mean that the affected scenarios were rebalanced to maintain the same level of difficulty; this doesn't mean playing them won't feel different.
vs.
Fixing only one of the things, among the very many things, that are unrealistic in Wesnoth.

knyghtmare:
> Potentially breaking the balance of multiple UMC and changing the experience of some mainline scenarios.

I believe the most extreme case would be _Uncharted Island_ of _**Dead Water**_. The deep water fighting vs monsters is bad enough as it is, but now hidden lurking cuttlefish? I would wish the scenario to be rebalanced in this PR before merger.

cooljeanius:
> > Potentially breaking the balance of multiple UMC and changing the experience of some mainline scenarios.
> 
> I believe the most extreme case would be _Uncharted Island_ of _**Dead Water**_.

(most extreme case *in mainline*, that is)

Jonathan-Kelly:
> The whole point of this change (making Cuttlefish have submerge) seems to be to make the game more realistic.

That is not the case. The main reason is to help create an interesting spread of wild monster units which are largely unconstrained by the needs of competitive multiplayer balance (unlike factional units) and [unique to environments](https://forums.wesnoth.org/viewtopic.php?t=52894) rather than particular story lines or historical events.

To expand on that last part, consider all the times in mainline campaigns where orcs or undead show up out of nowhere as filler enemies. From a narrative standpoint it is silly and from a game play standpoint it is boring always fighting these same enemy types. This is a big problem which a lot of people complain about often. Part of its solution is having a diverse enough set of dangerous wildlife that we can use them as replacements for the typical orcish and undead filler.

Overall we have a good spread of monsters thanks to the massive contributions made by doofus-01 over the last development cycle, but there are some gaps left in places. For one example, there are only two monster units above level 3, which limits are ability to center challenges around singularly dangerous units, such as "boss fights". For another example, there are presently no diurnal creatures so chaotic factions never have a ToD in which to deliver a strong counter attack against them.

Most of these gaps can be filled by thoughtfully adjusting the statistics and abilities of the existing monster units. In this particular case, the cuttlefish solves the lack of any sea creature having the submerge ability. It gives us a stealthy sea creature to work with, by adjusting a unit which, besides its ranged poison attack, is a bit underwhelming and could use a buff anyway. And yes, realistically speaking, there is no sea creature better at camouflage than the cuttlefish (with the possible exception of its octopus cousin).

> I believe the most extreme case would be _Uncharted Island_ of _**Dead Water**_. The deep water fighting vs monsters is bad enough as it is, but now hidden lurking cuttlefish? I would wish the scenario to be rebalanced in this PR before merger.

I will give it a play through as soon as we have the final cuttlefish statistics decided.

max-torch:
@Jonathan-Kelly Thanks for the further discussion about the rationale for this. I wasn't sure what the other considerations behind this were before but your response explains it well.

CelticMinstrel:
But even given that rationale, there's the option of adding a new unit rather than modifying the cuttlefish, which would dodge the balance issue entirely.

doofus-01:
> But even given that rationale, there's the option of adding a new unit rather than modifying the cuttlefish, which would dodge the balance issue entirely.

I agree that we shouldn't mess with the cuttlefish too much.  The kraken could probably have submerge ability, or an IPF-based variation or advancement for the regular cuttlefish could have submerge.

cooljeanius:
> > But even given that rationale, there's the option of adding a new unit rather than modifying the cuttlefish, which would dodge the balance issue entirely.
> 
> I agree that we shouldn't mess with the cuttlefish too much. The kraken could probably have submerge ability, or an IPF-based variation or advancement for the regular cuttlefish could have submerge.

IPF-based variation would be good, I'd just want to see #4753 fixed first so it'd show up properly in the units database.

Jonathan-Kelly:
> The kraken could probably have submerge ability,

Alright, moved the ability to the kraken.

knyghtmare:
> > The kraken could probably have submerge ability,
> 
> Alright, moved the ability to the kraken.

Given the player faces Krakens in WoF S4, I do hope you have adjusted the scenario balance to keep the fun:Frustration ratio intact/similar... :)

Jonathan-Kelly:
I tested S4 on nightmare difficulty with the change and it seemed about the same.

But in the process I discovered that the default AI may not be accounting for the swarm anti-ability (since this is not used by any of the default factions which it was trained on). As a result, the allied sea serpents were not biting into the enemy krakens even when they had the ToD advantage and no other targets to attack. Increasing their side aggression level seems to have overcome the issue and thus made the scenario slightly easier. Which seems to counterbalance the stealthier kraken.

nemaara (commented):
I'm okay with kraken having submerge although some cutscenes might need some fixing if it spawns in deep water and you can't actually see it. @Hejnewar you fine with it too?

CelticMinstrel:
> some cutscenes might need some fixing if it spawns in deep water and you can't actually see it.

Easily done if this is needed – just spawn it with (if I recall correctly) "uncovered" status.

nemaara:
Oh perfect, that makes life easy.

Jonathan-Kelly:
To be clear, I only hit the re-request review icon and then github or a bot local to the wesnoth project removed @Hejnewar and @doofus-01 from the reviewers list. As far as I know, I do not even have the necessary permissions to add or remove reviewers since I am not a project member (I can only re-request reviews from those others have assigned to a PR I authored).

> I'm okay with kraken having submerge although some cutscenes might need some fixing if it spawns in deep water and you can't actually see it.

> Easily done if this is needed – just spawn it with (if I recall correctly) "uncovered" status.

Good points. I have made this change to WoF S4. No change needed for DW since "Inky" (who starts out as a cuttlefish but can be advanced to a kraken) is a player unit and so always visible.

cooljeanius:
> To be clear, I only hit the re-request review icon and then github or a bot local to the wesnoth project removed @Hejnewar and @doofus-01 from the reviewers list. As far as I know, I do not even have the necessary permissions to add or remove reviewers since I am not a project member (I can only re-request reviews from those others have assigned to a PR I authored).

There, I just put them back

doofus-01 (approved):
Adding submerge to kraken and adjusting the WoF scenario look OK to me.

Jonathan-Kelly:
@Hejnewar so what do you say?

To summarize, the (now core) kraken is based on the [giant squid](https://en.wikipedia.org/wiki/Giant_squid) and [colossal squid](https://en.wikipedia.org/wiki/Colossal_squid) which live in the deepest zones of the ocean (they are products of [deep sea gigantism](https://en.wikipedia.org/wiki/Deep-sea_gigantism)). And like their octopus and cuttlefish relatives, they have chromatophores for some additional camouflage ability. So presumably the Kraken is more at home hiding *deep* underwater where nearly no sunlight reaches than many other sea creatures, including most of those in the game.

Also its sprite lurks low in the water compared with its rival, the sea serpent, giving it a perhaps more stealthy look:
kraken_v_serpent

spixi:
Descent into Darkness has an endless scenario featuring Inky the Cuttle Fish. I dont think it would be a good idea when Inky could advance into a Kraken with the submerge ability.

Jonathan-Kelly:
In its current form, this PR would have no effect on DiD, unless DiD explicitly makes the kraken an advancement for the cuttlefish. 

This old PR has been significantly pared down from its original form, leaving most of its comments no longer relevant to it and confusing for new viewers. So I will close this and open a new PR focused only on adding the submerge ability to the new kraken monster unit.

cooljeanius:
> So I will close this and open a new PR focused only on adding the submerge ability to the new kraken monster unit.

Follow-up: the "new PR" mentioned here ended up being #7909

PR #7134, opened 2022-11-16
Simple refactor in read_argv

bencsikandrei:
use return of `emplace_back` to swap buffers
use argc to reserve space in the args vector
remove explicit constructor used in `push_back` vs `emplace_back`

### DIFF
--- a/src/commandline_argv.cpp
+++ b/src/commandline_argv.cpp
@@ -86,8 +86,9 @@ std::vector<std::string> read_argv([[maybe_unused]] int argc, [[maybe_unused]] c
 	return win32_read_argv(flat_cmdline);
 #else
 	std::vector<std::string> args;
+	args.reserve(argc);
 	for(int i = 0; i < argc; ++i) {
-		args.push_back(std::string(argv[i]));
+		args.emplace_back(argv[i]);
 	}
 	return args;
 #endif
--- a/src/commandline_argv.cpp
+++ b/src/commandline_argv.cpp
@@ -65,8 +65,7 @@ std::vector<std::string> win32_read_argv(const std::string& input)
 	std::vector<std::string> res;
 
 	while(win32_parse_single_arg(start, end, buffer)) {
-		res.emplace_back();
-		res.back().swap(buffer);
+		res.emplace_back().swap(buffer);
 	}
 
 	return res;


src/commandline_argv.cpp
```diff
@@ -86,8 +86,9 @@ std::vector<std::string> read_argv([[maybe_unused]] int argc, [[maybe_unused]] c
 	return win32_read_argv(flat_cmdline);
```
bencsikandrei, line 86: this could technically be applicable to win32 as well, the function wasn't using argc tho so I was a bit skeptical about it

src/commandline_argv.cpp
```diff
@@ -65,8 +65,8 @@ std::vector<std::string> win32_read_argv(const std::string& input)
 	std::vector<std::string> res;
 
 	while(win32_parse_single_arg(start, end, buffer)) {
-		res.emplace_back();
-		res.back().swap(buffer);
+		auto& back = res.emplace_back();
+		back.swap(buffer);
```
stevecotton, line 69: I feel this is one to leave to the compiler's optimiser.

src/commandline_argv.cpp
```diff
@@ -86,8 +86,9 @@ std::vector<std::string> read_argv([[maybe_unused]] int argc, [[maybe_unused]] c
 	return win32_read_argv(flat_cmdline);
 #else
 	std::vector<std::string> args;
+	args.reserve(argc);
 	for(int i = 0; i < argc; ++i) {
-		args.push_back(std::string(argv[i]));
+		args.emplace_back(argv[i]);
```
stevecotton, line 91: Nice.

src/commandline_argv.cpp
bencsikandrei: to me it feels more natural to use that returned value, instead of obtaining it again via another function call (inlined for sure, but why drop the return?)

bencsikandrei: reverted this one

Vultraz: As of C++17 (min we support) you can do `res.emplace_back().swap(buffer)`

bencsikandrei: > As of C++17 (min we support) you can do `res.emplace_back().swap(buffer)`

I wanted to do that, should I?

stevecotton: Yes please. Apologies for saying no originally, seeing the reviews from Soliton and Vultraz has made me realise that I was wrong.

bencsikandrei: it's alright, on it

PR #7135, opened 2022-11-16
Remove some un-needed references and optimize some string operations

bencsikandrei:
when returning by value, the ideal way to store the returned value is another value - using a const& works only because of how c++ prolongs the lifetime of temporaries with const&'s

use move on strings that get destroyed at scope end

fix some formatting differences between a couple functions (parenthesis spaces)

### DIFF
--- a/src/commandline_options.cpp
+++ b/src/commandline_options.cpp
@@ -42,8 +42,8 @@ static void validate(boost::any& v, const std::vector<std::string>& values,
 	if (values.size() != 2) {
 		throw po::validation_error(po::validation_error::invalid_option_value);
 	}
-	ret_val.first = values.at(0);
-	ret_val.second = values.at(1);
+	ret_val.first = values[0];
+	ret_val.second = values[1];
 	v = ret_val;
 }
 
@@ -558,14 +558,11 @@ commandline_options::commandline_options(const std::vector<std::string>& args)
 
 void commandline_options::parse_log_domains_(const std::string &domains_string, const int severity)
 {
-	if(std::vector<std::string> domains = utils::split(domains_string, ','); !domains.empty())
-	{
-		if(!log)
-		{
+	if(std::vector<std::string> domains = utils::split(domains_string, ','); !domains.empty()) {
+		if(!log) {
 			log.emplace();
 		}
-		for (auto&& domain : domains)
-		{
+		for (auto&& domain : domains) {
 			log->emplace_back(severity, std::move(domain));
 		}
 	}
@@ -610,8 +607,7 @@ std::vector<std::pair<unsigned int,std::string>> commandline_options::parse_to_u
 	const std::string expected_format
 			= "UINT"s + separator + "STRING";
 
-	for (const std::string &s : strings)
-	{
+	for (const std::string &s : strings) {
 		std::vector<std::string> tokens = utils::split(s, separator);
 		if(tokens.size() != 2) {
 			throw bad_commandline_tuple(s, expected_format);
@@ -637,8 +633,7 @@ std::vector<std::tuple<unsigned int,std::string,std::string>> commandline_option
 	const std::string expected_format
 			= "UINT"s + separator + "STRING" + separator + "STRING";
 
-	for (const std::string &s : strings)
-	{
+	for (const std::string &s : strings) {
 		const std::vector<std::string> tokens = utils::split(s, separator);
 		if(tokens.size() != 3) {
 			throw bad_commandline_tuple(s, expected_format);
--- a/src/commandline_options.cpp
+++ b/src/commandline_options.cpp
@@ -29,6 +29,8 @@
 #include <boost/program_options/value_semantic.hpp>  // for value, etc
 #include <boost/program_options/variables_map.hpp>  // for variables_map, etc
 
+#include <array>
+
 namespace po = boost::program_options;
 
 class two_strings : public std::pair<std::string,std::string> {};
@@ -40,8 +42,8 @@ static void validate(boost::any& v, const std::vector<std::string>& values,
 	if (values.size() != 2) {
 		throw po::validation_error(po::validation_error::invalid_option_value);
 	}
-	ret_val.first = values[0];
-	ret_val.second = values[1];
+	ret_val.first = values.at(0);
+	ret_val.second = values.at(1);
 	v = ret_val;
 }
 
@@ -564,14 +566,13 @@ void commandline_options::parse_log_domains_(const std::string &domains_string, 
 		}
 		for (auto&& domain : domains)
 		{
-			assert(log && "`log` must have been initialized before this loop");
 			log->emplace_back(severity, std::move(domain));
 		}
 	}
 }
 
 void commandline_options::parse_log_strictness (const std::string & severity) {
-	static const lg::logger* const loggers[4] {&lg::err(), &lg::warn(), &lg::info(), &lg::debug()};
+	static const std::array<const lg::logger*, 4> loggers {{&lg::err(), &lg::warn(), &lg::info(), &lg::debug()}};
 	for (const lg::logger * l : loggers ) {
 		if (severity == l->get_name()) {
 			lg::set_strict_severity(*l);
@@ -598,14 +599,16 @@ void commandline_options::parse_resolution_ (const std::string& resolution_strin
 		throw bad_commandline_resolution(resolution_string);
 	}
 
-	resolution.emplace(xres, yres);
+	resolution = std::pair(xres, yres);
 }
 
 std::vector<std::pair<unsigned int,std::string>> commandline_options::parse_to_uint_string_tuples_(const std::vector<std::string> &strings, char separator)
 {
 	std::vector<std::pair<unsigned int,std::string>> vec;
+
+	using namespace std::literals;
 	const std::string expected_format
-			= std::string() + "UINT" + separator + "STRING";
+			= "UINT"s + separator + "STRING";
 
 	for (const std::string &s : strings)
 	{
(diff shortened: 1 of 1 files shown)


src/commandline_options.cpp
```diff
@@ -558,28 +556,33 @@ commandline_options::commandline_options(const std::vector<std::string>& args)
...
-		if (!log)
-			log = std::vector<std::pair<int, std::string>>();
-		log->emplace_back(severity, domain);
+		if(!log)
+		{
+			log.emplace();
+		}
+		for (auto&& domain : domains)
+		{
+			assert(log && "`log` must have been initialized before this loop");
```
stevecotton, line 567: I think that assert can never trigger.

In general, things should handle the situation rather than assert.

src/commandline_options.cpp
```diff
@@ -595,18 +598,18 @@ void commandline_options::parse_resolution_ ( const std::string& resolution_stri
 		throw bad_commandline_resolution(resolution_string);
 	}
 
-	resolution = std::pair(xres, yres);
+	resolution.emplace(xres, yres);
```
stevecotton, line 601: While the new code is nice, the old code seems more readable to me - it's clear that it's replacing the old value (if any) rather than appending to a list. For code that's only called when parsing command line options, my preference is the old code.

stevecotton:
I like the changes that I haven't commented on.

src/commandline_options.cpp
```diff
@@ -558,28 +556,33 @@ commandline_options::commandline_options(const std::vector<std::string>& args)
...
+			assert(log && "`log` must have been initialized before this loop");
+			log->emplace_back(severity, std::move(domain));
+		}
 	}
 }
 
-void commandline_options::parse_log_strictness (const std::string & severity ) {
-	static const std::array<lg::logger const*, 4> loggers {{&lg::err(), &lg::warn(), &lg::info(), &lg::debug()}};
+void commandline_options::parse_log_strictness (const std::string & severity) {
+	static const lg::logger* const loggers[4] {&lg::err(), &lg::warn(), &lg::info(), &lg::debug()};
```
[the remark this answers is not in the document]
bencsikandrei: want to point my reasons out

We're not really using anything from what std array offers over a C style array, in this case

The code using array is more complex (as array is an aggregate and contains an aggregate so the double {})

Also, there's no need for its copy semantics or any other of its advantages (like the member side or others). Don't get me wrong, I'm all for std array but only when it brings something to the table.

Otherwise it's a cost we pay in compilation time for little to no gain.

All that being said I respect your opinion and will change it if you desire

src/commandline_options.cpp
bencsikandrei: Asserts are useful to verify some preconditions

Maybe it's tio much here, but I wanted to still offer the check .at was offering but without a cost at -O3

src/commandline_options.cpp
```diff
@@ -42,8 +40,8 @@ static void validate(boost::any& v, const std::vector<std::string>& values,
 	if (values.size() != 2) {
 		throw po::validation_error(po::validation_error::invalid_option_value);
 	}
-	ret_val.first = values.at(0);
-	ret_val.second = values.at(1);
+	ret_val.first = values[0];
+	ret_val.second = values[1];
```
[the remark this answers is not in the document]
bencsikandrei: This file has other places where it uses [1] and [0]. I'll try to see and post a compiler explorer link. But we're double checking something that can't happen. Would it not be nicer to show the intent better and not re-check?

Maybe the compile could optimize it away, it may not do so in debug.

src/commandline_options.cpp
bencsikandrei: You make a point, and I respect your opinion, but I have to disagree, here:s my reasoning

To me emplace is a much better api than operator=, in this case, as it show the intent of constructing in place rather than creating a temp pair and then putting it inside the optional.

Maybe emplace is a bad choice for a name...

Again, will revert the change

src/commandline_options.cpp
```diff
@@ -595,18 +598,18 @@ void commandline_options::parse_resolution_ ( const std::string& resolution_stri
...
 
-	resolution = std::pair(xres, yres);
+	resolution.emplace(xres, yres);
 }
 
 std::vector<std::pair<unsigned int,std::string>> commandline_options::parse_to_uint_string_tuples_(const std::vector<std::string> &strings, char separator)
 {
 	std::vector<std::pair<unsigned int,std::string>> vec;
-	const std::string& expected_format
+	const std::string expected_format
```
CelticMinstrel, line 607: This would also be a good place to use string literals.

```diff
-			= std::string() + "UINT" + separator + "STRING";
+			= "UINT"s + separator + "STRING"s;
```

(Requires a `using namespace std::literals` to be in scope.)

bencsikandrei: on it

src/commandline_options.cpp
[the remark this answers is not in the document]
bencsikandrei: you were 100% right! the compiler does manage to optimize the second if/exception throwing on all major platforms that wesnoth supports.

I'm a little surprised, even though I knew optimizers are insanely good

I expected that, at least in debug, it would emit more code for the .at(..)

here's the link: https://godbolt.org/z/r1x1P19vT

I'll revert it, but what happens with lines: 593, 616 (as of now) where the same pattern is used, i.e. check size != 2, throw, then used operator [] (not .at). do we want to change those to .at as well? do we change everything to []?

src/commandline_options.cpp
bencsikandrei: just realized I replied thinking this comment was on something else

I'll remove the assert

src/commandline_options.cpp
bencsikandrei: done

src/commandline_options.cpp
bencsikandrei: done

src/commandline_options.cpp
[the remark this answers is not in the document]
bencsikandrei: done

src/commandline_options.cpp
stevecotton: @CelticMinstrel does it subtly matter whether the `"STRING"` is marked as a literal instead of a C-style string?

Looks okay as-is to me, I assume the compiler will optimise anyway.

stevecotton (approved):

src/commandline_options.cpp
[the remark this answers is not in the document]
stevecotton: Thanks - I'll admit I'm flip-flopping on the decision considering that it's inconsistent with the other lines, and that all the code in this area is obviously doing input validation. Reading it as a file rather than just a diff, I'm currently thinking your change to using `[0]` and `[1]` here was the right one.

@CelticMinstrel do you have an opinion please?

src/commandline_options.cpp
```diff
@@ -558,28 +558,32 @@ commandline_options::commandline_options(const std::vector<std::string>& args)
...
+	if(std::vector<std::string> domains = utils::split(domains_string, ','); !domains.empty())
 	{
-		if (!log)
-			log = std::vector<std::pair<int, std::string>>();
-		log->emplace_back(severity, domain);
+		if(!log)
+		{
+			log.emplace();
+		}
+		for (auto&& domain : domains)
```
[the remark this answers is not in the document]
bencsikandrei: on it

src/commandline_options.cpp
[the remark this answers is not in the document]
CelticMinstrel: I generally default to using `[]` rather than `.at()` unless:

* You actually want the exception that `.at()` throws (perhaps you're planning to catch it) or
* It's a const map, so `[]` is not permitted (this point doesn't apply if it's a vector).

src/commandline_options.cpp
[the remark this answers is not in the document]
CelticMinstrel: Additionally, doesn't our style guide proscribe no space between control flow keywords and the opening parenthesis?

src/commandline_options.cpp
CelticMinstrel: I don't think it makes any actual difference to the resulting code. If you omit the second `s`, the compiler will still generate an implicit conversion to string there, I think, so the `s` just makes the conversion explicit. In any case, even if I'm wrong, I'm pretty sure it would come out the same after optimization.

src/commandline_options.cpp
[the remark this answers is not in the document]
bencsikandrei: done

src/wesnoth.cpp
```diff
@@ -1107,7 +1107,7 @@ int main(int argc, char** argv)
 			PLAIN_LOG << "Started on " << ctime(&t);
 		}
 
-		const std::string& exe_dir = filesystem::get_exe_dir();
+		const std::string exe_dir = filesystem::get_exe_dir();
```
stevecotton, line 1110: This line was rewritten in #7145, so it's a merge conflict to change it here.

Please rebase your changes, and force-push the new version of the branch.

bencsikandrei: dang, missed it

bencsikandrei: if you get to see my very weird rebase... how I managed to rebase like that, not sure, fixing it now

bencsikandrei: done

bencsikandrei:
got a bunch of labels because of my failed rebase :( sorry for that

stevecotton:
Thanks. No problem about the extra labels, it was just a few mouseclicks to undo them.

PR #7136, opened 2022-11-16
Make undo work again, undeprecate on_event()

### LINKED ISSUE #6898: Undo command not working

- **Version:** 1.17.6, master @ 720094361d446c9f05495906117dd2c773ce746f
- **Downloaded from:** git
- **OS:** Linux

### Description of the bug

The undo command no longer works. The "u" button has no apparent effect, and there's no "undo" option in the right click menu.

1. Start the tutorial
2. Move to the wrong place
3. Press "u"

### Expected behavior

The move undoes. I assume there should be a menu item in the right click popup menu as well, as that's what the tutorial message on the screen is saying.

### Additional context

I bisected this and it was introduced in the commit 7e234f8833282424b3535b9c334c751748f7222b

stevecotton:
Fixes #6898. The issue is that non-WML events added through the new events API always disable undo with no equivalent of WML's `[allow_undo]`. The long-term fix is to add a way to do that; however until that's available then listeners for `moveto` need to use the old `on_event` API. The old `on_event` API can't be deprecated yet, and this is enforced by our unit tests (the build fails if there are unexpected deprecation warnings during the tests).

Reverts most of 7e234f8833282424b3535b9c334c751748f7222b. Does not revert files that only listen for non-undoable events such as `die` or `new turn`.

Reverts the deprecation part of #5663's 8cd133263058a5df85f64988e348d2cf54d13a48.

### DIFF
--- a/changelog_entries/fix_undo.md
+++ b/changelog_entries/fix_undo.md
@@ -1,0 +1,2 @@
+ ### Miscellaneous and bug fixes
+   * Fixed not being able to undo moves (bug #6898)
--- a/data/lua/diversion.lua
+++ b/data/lua/diversion.lua
@@ -1,6 +1,7 @@
 
 local _ = wesnoth.textdomain 'wesnoth-help'
 local T = wml.tag
+local on_event = wesnoth.require("on_event")
 
 local u_pos_filter = function(u_id)
 
@@ -99,7 +100,7 @@ function wesnoth.wml_actions.on_undo_diversion(cfg)
 	status_anim_update(true)
 end
 
-wesnoth.game_events.add_repeating("moveto, die, recruit, recall", function()
+on_event("moveto, die, recruit, recall", function()
         status_anim_update()
 
 end)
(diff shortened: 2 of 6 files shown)


CelticMinstrel:
Reverting (most of) 7e234f8 I get, as that was my recommended quick fix for this, but why revert the deprecation too? There shouldn't be anything wrong with continuing to use things that are deprecated.

Pentarctagon:
There should be an issue opened to keep track of the missing undo functionality from the new API, even if #6898 is closed.

soliton-:
Would be great to get this merged ASAP. Having a major game feature broken for this long is pretty bad even if it's a dev version.

stevecotton:
@CelticMinstrel I'm assuming your "it's not the end of the world" is an approval, will merge this evening unless you say otherwise.

CelticMinstrel:
It's not an approval so much as a lack of disapproval? But close enough, I guess.

Issue #7129, opened 2022-11-15
Hero Units Should Have Traits (if They Are Not a Unique Unit?)


Ashmyr:
### Describe the desired feature

In general, I think this should apply to all campaigns. If your leader is a regular unit and he doesn't have traits, his own recruits are better than him. Leaders that are unique units have their own stats, but I still think they should have traits. Traits add character to the characters (lol), as well.
I'm sure there might be more (such as allies or loyal units you gain along the way), but just looking at the code for the leaders for each campaign...
- Kai Krellis in Dead Water has no traits.
- Malin Keshar in Descent into Darkness has no traits, but is a unique unit.
- Gweddry in Eastern Invasion has no traits, but is a unit that he cannot recruit. However, I still think he should have them.
- Konrad and Li'sar in Heir to the Throne have no unique traits, but are both unique units.
- Baldras in Liberty has no traits.
- Rugnur in Sceptre of Fire has no traits.
- Kapou'e in Son of the Black-Eye has no traits.
- Haldric in The Rise of Wesnoth has no traits, but is a unique unit.
- Deoran in The South Guard has no traits, but is a unique unit.
- Arvith in A Tale of Two Brothers has no traits.
- Kaleh in Under the Burning Suns has no traits, but is a unique unit.

Ashmyr:
I am also a fan of giving "illegal traits", ie, Jessene having _Dextrous_, and I feel that can add characterization as well. I think allied leaders (except for cutscenes) and Loyal Units should all have traits, but enemies don't have to.

knyghtmare:
List of proposed traits for each and accompanying logic?

Ashmyr:
I'm not quite sure for some. I think it needs to fit the personality and the campaign. If the leader has a lot of go to missions, maybe Quick? If the leader is expected to be a frontline unit, maybe Strong, Fearless or Resilient? Dextrous could help out leaders with a weak secondary attack and give them a small boost...

CelticMinstrel:
Yes, if it's a unique unit you can make the trait be nothing but "sugar" by adjusting the base stats so that the trait brings them back to the original level.

If it's not a unique unit, you'd need to consider balance in choosing traits.

Ashmyr:
Thoughts on leader traits:
Arvith: Strong, Intelligent
Deoran: Strong, Intelligent
Jessene: Dextrous, Strong
Gweddry: Resilient, Fearless
Konrad: Resilient, Dextrous
[...]

PR #7138, opened 2022-11-17
[TRoW][Units] Give Prince Haldric his resilient and healthy traits

knyghtmare:
Part of #7129

### DIFF
--- a/data/campaigns/The_Rise_Of_Wesnoth/scenarios/01_A_Summer_of_Storms.cfg
+++ b/data/campaigns/The_Rise_Of_Wesnoth/scenarios/01_A_Summer_of_Storms.cfg
@@ -76,6 +76,10 @@
         id=Prince Haldric
         name= _ "Prince Haldric"
         profile=portraits/haldric.webp
+        [modifications]
+            {TRAIT_RESILIENT}
+            {TRAIT_HEALTHY}
+        [/modifications]
         {GOLD 140 120 100}
         unrenamable=yes
         side=1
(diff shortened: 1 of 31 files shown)


nemaara (approved):
Or hear me out, don't nerf his HP!! (Or nerf it a little less ha, up to you).

knyghtmare:
> Or hear me out, don't nerf his HP!! (Or nerf it a little less ha, up to you).

Currently, the HP nerf is such that he is still

|Condition |  0 | 1 | 2 | 3 |
| -------- | ------| - | - | - |
| no traits | 24 | 36 | 48 | 64 |
| HP nerf | 19 | 29 | 39 | 53 | 
| with traits | 24 | 36 | 48 | 64 |

stevecotton:
Although he's a unique unit, when I right-click on him and select "Unit Description", the stats in the encyclopedia no longer match the stats of the unit.

For a normal unit, we'd accept that because the trait system's mechanics are a necessary part of making the combat mechanics more interesting. Doing it for unique units is unnecessary, and I think this idea shouldn't be implemented.

knyghtmare:
> Although he's a unique unit, when I right-click on him and select "Unit Description", the stats in the encyclopedia no longer match the stats of the unit.

Is this not the default case for every recruited unit with traits?

stevecotton:
I already answered that in my second paragraph - we have to do it for recruited units, because it's a necessary cost of adding that diversity for non-unique units. That doesn't mean it's good to do it when we don't have to.

nemaara (changes requested):
Okay stevecotton raised a good point. Haldric isn't a regular unit, he might not have AMLAs but since nobody else uses his unit we can treat him as a hero unit. Leaving this open because if you wanted to give him a dummy trait that adds some flavor you can do that.

PR #7144, opened 2022-11-17
[1.17]further simplification of overwrite_specials attribute code

newfrenchy83:
I know that I have already proposed a simplification of the code of this attribute, but several things lead me to propose this second rewrite.

First, using the first loop to define three boolean variables is more complicated than simply eliminating from the 'overwrite_specials' list the abilities not carrying the attribute then; if an ability with the attribute is present define a double for loop inside each other, where for each ability with overwrite_specials the type of ability removal from the returned list is defined at that time (one_side or both_sides). The code is even more readable than with the previous work in my opinion.

The second reason is that I eventually want to be able to use the request code https://github.com/wesnoth/wesnoth/pull/7050 to be even more selective about the abilities that can be removed or not and extend the use of the attribute beyond chance_to_hit (when (and if) these two PRs will be merged in master, I will have to add 3-4 lines of code in hadcoding to implement the filtering). Building a list instead of splicing the existing one is incompatible with this project.

### DIFF
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1297,40 +1297,41 @@ static bool overwrite_special_affects(const config& special)
 	return (apply_to == "one_side" || apply_to == "both_sides");
 }
 
-unit_ability_list attack_type::overwrite_special_checking(const std::string& ability, unit_ability_list temp_list, unit_ability_list abil_list, const std::string& filter_self, bool is_special) const
+unit_ability_list attack_type::overwrite_special_checking(const std::string& ability, unit_ability_list input, unit_ability_list overwriters, const std::string& filter_self, bool is_special) const
 {
-	for(unit_ability_list::iterator i = abil_list.begin(); i != abil_list.end();) {
+	for(unit_ability_list::iterator i = overwriters.begin(); i != overwriters.end();) {
 		if(!overwrite_special_affects(*i->ability_cfg)) {
-			i = abil_list.erase(i);
+			i = overwriters.erase(i);
 		} else {
 			++i;
 		}
 	}
-	if(abil_list.empty()){
-		return temp_list;
+	if(overwriters.empty()){
+		return input;
 	}
 
-	for(const auto& i : abil_list) {
+	for(const auto& i : overwriters) {
 		bool affect_side = ((*i.ability_cfg)["overwrite_specials"] == "one_side");
-		for(unit_ability_list::iterator j = temp_list.begin(); j != temp_list.end();) {
-			bool temp_overwrite = (is_special || !overwrite_special_affects(*j->ability_cfg));
-			bool is_overwritable = temp_overwrite;
-			if(affect_side){
+		for(unit_ability_list::iterator j = input.begin(); j != input.end();) {
+			bool is_overwritable = (is_special || !overwrite_special_affects(*j->ability_cfg));
+			bool one_side_overwritable = true;
+			if(affect_side && is_overwritable){
 				if(special_active_impl(shared_from_this(), other_attack_, *i.ability_cfg, AFFECT_SELF, ability, filter_self)){
-					is_overwritable = temp_overwrite && special_active_impl(shared_from_this(), other_attack_, *j->ability_cfg, AFFECT_SELF, ability, filter_self);
+					one_side_overwritable = special_active_impl(shared_from_this(), other_attack_, *j->ability_cfg, AFFECT_SELF, ability, filter_self);
 				}
 				else if(special_active_impl(other_attack_, shared_from_this(), *i.ability_cfg, AFFECT_OTHER, ability, filter_self)){
-					is_overwritable = temp_overwrite && special_active_impl(other_attack_, shared_from_this(), *j->ability_cfg, AFFECT_OTHER, ability, filter_self);
+					one_side_overwritable = special_active_impl(other_attack_, shared_from_this(), *j->ability_cfg, AFFECT_OTHER, ability, filter_self);
 				}
 			}
+			is_overwritable = is_overwritable && one_side_overwritable;
 			if(is_overwritable) {
-				j = temp_list.erase(j);
+				j = input.erase(j);
 			} else {
 				++j;
 			}
 		}
 	}
-	return temp_list;
+	return input;
 }
 
 	/**
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1297,48 +1297,40 @@ static bool overwrite_special_affects(const config& special)
 	return (apply_to == "one_side" || apply_to == "both_sides");
 }
 
-unit_ability_list attack_type::overwrite_special_checking(const std::string& ability, const unit_ability_list& temp_list, const unit_ability_list& abil_list, const std::string& filter_self, bool is_special) const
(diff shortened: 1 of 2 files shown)


src/units/attack_type.hpp
```diff
@@ -134,7 +134,7 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 	 * @param filter_self name of [filter_"self/student"] if is abilities or specials who are checked
 	 * @param is_special for determine if list is a special or a ability.
 	 */
-	unit_ability_list overwrite_special_checking(const std::string& ability, const unit_ability_list& temp_list, const unit_ability_list& abil_list, const std::string& filter_self, bool is_special) const;
+	unit_ability_list overwrite_special_checking(const std::string& ability, unit_ability_list temp_list, unit_ability_list abil_list, const std::string& filter_self, bool is_special) const;
```
stevecotton, line 137: I found the documentation a bit confusing, and think this would be clearer (including renaming some of the arguments). Please would you check if it's correct?

Assuming it is correct, you can either cherry-pick it or I can create a PR for it afterwards.

```
	/**
	 * Filter a list of abilities or weapon specials, removing any entries that are overridden by
	 * the overwrite_specials attributes of a second list.
	 *
	 * @param ability The special ability type who is being checked.
	 * @param input list to check, a filtered copy of this list is returned by the function.
	 * @param overwriters list that may have overwrite_specials attributes.
	 * @param filter_self name of [filter_"self/student"] if is abilities or specials who are checked.
	 * @param is_special if true, input contains weapon specials; if false, it contains abilities.
	 */
	unit_ability_list overwrite_special_checking(const std::string& ability, unit_ability_list input, unit_ability_list overwriters, const std::string& filter_self, bool is_special) const;
```

src/units/abilities.cpp
```diff
@@ -1297,48 +1297,40 @@ static bool overwrite_special_affects(const config& special)
...
-		} else if(overwrite_opponent){
-			overwrite = (!is_special && overwrite_special_affects(*i.ability_cfg)) || special_active_impl(shared_from_this(), other_attack_, *i.ability_cfg, AFFECT_SELF, ability, filter_self);
-		}
-		if(overwrite) {
-			overwrite_list.emplace_back(i);
+	for(const auto& i : abil_list) {
+		bool affect_side = ((*i.ability_cfg)["overwrite_specials"] == "one_side");
+		for(unit_ability_list::iterator j = temp_list.begin(); j != temp_list.end();) {
+			bool temp_overwrite = (is_special || !overwrite_special_affects(*j->ability_cfg));
+			bool is_overwritable = temp_overwrite;
```
stevecotton, line 1317: If `temp_overwrite` is false now, then all assignments to `is_overwritable` will also be false, because the ones below start with `temp_overwrite &&`.

stevecotton (changes requested):
Thanks - it's much easier to review a small refactor. I have a few comments, but my general feeling is that this will merge soon.

src/units/abilities.cpp
newfrenchy83: is_overwritable can't be true if is an ability with overwrite_specials attribute is_checked, then yes i will rephrase the code of if loop.

src/units/abilities.cpp
```diff
@@ -1297,48 +1297,40 @@ static bool overwrite_special_affects(const config& special)
...
-		} else if(overwrite_self){
-			overwrite = (!is_special && overwrite_special_affects(*i.ability_cfg)) || special_active_impl(other_attack_, shared_from_this(), *i.ability_cfg, AFFECT_OTHER, ability, filter_self);
-		} else if(overwrite_opponent){
-			overwrite = (!is_special && overwrite_special_affects(*i.ability_cfg)) || special_active_impl(shared_from_this(), other_attack_, *i.ability_cfg, AFFECT_SELF, ability, filter_self);
-		}
-		if(overwrite) {
-			overwrite_list.emplace_back(i);
+	for(const auto& i : abil_list) {
+		bool affect_side = ((*i.ability_cfg)["overwrite_specials"] == "one_side");
+		for(unit_ability_list::iterator j = temp_list.begin(); j != temp_list.end();) {
```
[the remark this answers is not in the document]
newfrenchy83: for one function only, idon't know.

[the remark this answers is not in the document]
stevecotton: I'm looking at doing that refactor elsewhere in the codebase, will open a followup PR after this one merges.

src/units/attack_type.hpp
```diff
@@ -127,14 +127,17 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 
 	// Configured as a bit field, in case that is useful.
 	enum AFFECTS { AFFECT_SELF=1, AFFECT_OTHER=2, AFFECT_EITHER=3 };
-	/** overwrite_special_checking : return an unit_ability_list list after checking presence or not of overwrite_specials
+		/**
```
stevecotton, line 130: Trivial issue - the whitespace is wrong on this line.

src/units/abilities.cpp
```diff
@@ -1297,48 +1297,41 @@ static bool overwrite_special_affects(const config& special)
...
+			if(affect_side && is_overwritable){
+				if(special_active_impl(shared_from_this(), other_attack_, *i.ability_cfg, AFFECT_SELF, ability, filter_self)){
+					one_side_overwritable = special_active_impl(shared_from_this(), other_attack_, *j->ability_cfg, AFFECT_SELF, ability, filter_self);
+				}
+				else if(special_active_impl(other_attack_, shared_from_this(), *i.ability_cfg, AFFECT_OTHER, ability, filter_self)){
+					one_side_overwritable = special_active_impl(other_attack_, shared_from_this(), *j->ability_cfg, AFFECT_OTHER, ability, filter_self);
+				}
+			}
+			is_overwritable = is_overwritable && one_side_overwritable;
+			if(is_overwritable) {
```
stevecotton, line 1327: The line above could be combined with this one, so it's just `if(is_overwritable && one_side_overwritable)`. However, no need to do that if your next PR is going to add additional logic that's easier with them kept separate.

stevecotton (approved):
@CelticMinstrel this looks good to me, but will wait until you've had time to look too.

src/units/attack_type.hpp
newfrenchy83: corrected

CelticMinstrel:
I'm afraid I have no idea what this is doing, so I don't think my review would be helpful…

newfrenchy83:
> I'm afraid I have no idea what this is doing, so I don't think my review would be helpful…

This is another code simplification of the overwrite_specials attribute to pave the way for adding a filter that will use code from https://github.com/wesnoth/wesnoth/pull/7050.
Otherwise, the operation of the attribute will not be affected and the code will be even easier to understand than before.

stevecotton:
My plan is to merge soon, just waiting for the CI to confirm that the build error introduced in a7b39bd8996a847fd4c2160bdb6d12b3efdc67f5 has been fixed.

newfrenchy83:
> My plan is to merge soon, just waiting for the CI to confirm that the build error introduced in [a7b39bd](https://github.com/wesnoth/wesnoth/commit/a7b39bd8996a847fd4c2160bdb6d12b3efdc67f5) has been fixed.

it seem what yes.

PR #7147, opened 2022-11-18
Refactor ability lists using utils::erase_if and std::remove_if

stevecotton:
C++20 will add std::erase_if as a convenience wrapper around std::remove_if, handling the subsequent call to std::vector::erase. This simplifies readability, avoiding the need to either store the result of std::remove_if in a temporary or to put the second argument of vector::erase after the body of a lambda.

This adds utils::erase_if, which does the same thing. The C++20 function returns the number of elements removed, but as none of the callers need that I've not implemented it in utils::erase_if.

This commit omits the refactor of attack_type::overwrite_special_checking, because that function is being worked on in parallel by @newfrenchy83. It'll be simpler for either of us to update it later instead of creating merge conflicts.

@newfrenchy83, please would you review this?

### DIFF
--- a/src/actions/heal.cpp
+++ b/src/actions/heal.cpp
@@ -30,6 +30,7 @@
 #include "units/abilities.hpp"
 #include "units/udisplay.hpp"
 #include "units/map.hpp"
+#include "utils/general.hpp"
 
 #include <list>
 #include <vector>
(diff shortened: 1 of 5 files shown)


newfrenchy83:
it's fine for me, itested and it's work.

PR #7152, opened 2022-11-21
1.17 - a couple more monsters

doofus-01:
Initial commit is a dragonfly, inspired by forum member Helmet in https://forums.wesnoth.org/viewtopic.php?t=56341.
These dragonflies are not lifted from that, but that's where the inspiration came from and I used his sprite as a starting point, so he should be in the credits.

Other things to add here:
- level 0 dragonfly naiad
- The kraken, following up on #7125 
- ~~possibly reorganize the "monsters" race further to monsters/beasts/bugs/etc.~~ Actually, this is better left to a different PR


CelticMinstrel:
I feel like a dragonfly is not complete without a mosquito.

https://units.wesnoth.org/1.14/Black_Cross_of_Aleron/en_US/Giant%20Mosquito.html

knyghtmare:
I approve the Dragonflies. Very high UMC usable value.

Mosquito looks like it could use a overhaul. Too big eyeballs.

Hejnewar:
Im disappointed in the stats of these things, they not at all reflect how strong they actually are irl. Dragonflies have 95% hunt success which is the highest in animal kingdom (in comparison lions have like 25%, african wild dogs have 85% but the again these are pack hunters and not solitary hunters like dragonfly), dragonflies can kill and eat the spider after geting caught in its web, adittionally dragonflies never sting.

Zireael07:
95% hunt success is INSANE.

CelticMinstrel:
…wait a second. Why is the kraken green? Was it always green?

doofus-01:
> I feel like a dragonfly is not complete without a mosquito.
https://units.wesnoth.org/1.14/Black_Cross_of_Aleron/en_US/Giant%20Mosquito.html

To be honest, I'm not really feeling the mosquito.  Does it fill some sort of niche that's missing?

> Im disappointed in the stats of these things, they not at all reflect how strong they actually are irl. [...]

Nice factoids, I'm not sure how that translates to anything here.  All I can say is the stats were in the context of the ants and since it had better vision and movement.  You are certainly free to adjust the stats later, especially if you update the fluff to support it.

> …wait a second. Why is the kraken green? Was it always green?

What color is it supposed to be?

CelticMinstrel:
> To be honest, I'm not really feeling the mosquito. Does it fill some sort of niche that's missing?

Prey for the dragonfly.

> What color is it supposed to be?

I don't really know. I checked and I guess it really is green in WoF and DW, but that just seems weird to me somehow. I guess there isn't really a problem if it's always been green though.

doofus-01:
> I suspect the kraken's green color is left over from way back when [the original cuttlefish sprite] was also green. The green does gives it a foreboding look, though on the other hand large squid species typically have a reddish orange appearance.

The old kraken sprite surely comes from the old cuttlefish.  Squid species are really weird, we could have an entire faction based on them, from vampire squids to piglet squids to giant squids, and there are different types of ink spray such as squid-shaped decoys, "smoke" screens, or poisons.

> What do you think about giving the dragonflies the marksman ability on their bite attack?

Makes sense to me, let's do it.

doofus-01:
I'm going to merge this now, can follow up next weekend as needed.

PR #7154, opened 2022-11-21
Escape < and > even if only one of them exists

ProditorMagnus:
In python negative index counts from the end, so without this change or another solution 
`example with only < one side` would be converted to  
`example with only < one sid&gt;example with only < one side`

### DIFF
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -1595,10 +1595,10 @@ def pangoize(message, filename, line):
         if "<" in reduced or ">" in reduced:
             if message == reduced: # No pango markup
                 here = message.find('<')
-                if message[here:here+4] != "&lt;":
+                if here != -1 and message[here:here+4] != "&lt;":
                     message = message[:here] + "&lt;" + message[here+1:]
                 here = message.find('>')
-                if message[here:here+4] != "&gt;":
+                if here != -1 and message[here:here+4] != "&gt;":
                     message = message[:here] + "&gt;" + message[here+1:]
             else:
                 print('"%s", line %d: < or > in pango string requires manual fix.' % (filename, line))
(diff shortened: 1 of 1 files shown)


ProditorMagnus:
I wonder why just string replace function isnt used.

data/tools/wmllint
```diff
@@ -1595,10 +1595,10 @@ def pangoize(message, filename, line):
         if "<" in reduced or ">" in reduced:
             if message == reduced: # No pango markup
                 here = message.find('<')
-                if message[here:here+4] != "&lt;":
+                if here != -1 and message[here:here+4] != "&lt;":
```
soliton-, line 1598: If you find `>` how can this ever be equal to `&lt;` anyway?

Elvish-Hunter:
> I wonder why just string replace function isnt used.

@ProditorMagnus I've been looking at the code, and indeed I couldn't find a good reason to explain why it was written like that.
One of the key points are these three lines:

        reduced = pangostrip(message)
        if "<" in reduced or ">" in reduced:
            if message == reduced: # No pango markup

which make sure that the original line didn't contain any Pango markup, so any unescaped `<` and `>` characters are certainly supposed to be used on their own (with an exception explained later). As @soliton- said, if the `.find()` method finds these characters, the string slice obtained with this:

    if here != -1 and message[here:here+4] != "&lt;":

can never be equal to `&lt;` or `&gt;`.
So, the best solution would probably be replacing the whole block with this:

    if "<" in message or ">" in message:
        reduced = pangostrip(message)
        if ("<" in reduced or ">" in reduced) and not rgb :
            if message == reduced: # No pango markup
                message = message.replace("<", "&lt;").replace(">", "&gt;")

Why have I added `and not rgb`? Because at line 1570 there's a check for old color markup tags, so in this case the characters shouldn't be replaced.

PR #7157, opened 2022-11-22
Fix [special_note] duplication when unstoring units

### LINKED ISSUE #7153: [unstore_unit] causes duplicate [movetype][special_note]s

- **Version:** 1.16.6+dev, 1.17.10+dev
- **Downloaded from:** Git
- **Build info:** 94d9cdb2db05ff99ed7ee7d579ec810a0d2c71b2
- **OS:** Linux

### Description of the bug

For movetypes that include `[special_note]`s, storing and unstoring a unit causes duplicates of the special note.

Originally reported by @Toranks in https://r.wesnoth.org/t56418

```
# wmllint: no translatables

#####
# API(s) being tested: [special_note], [unstore_unit]
##
# Actions:
# Alice is a Spectre, a unit which should have 1 special note from the 'undeadspirit' movetype.
# Store Alice
# Unstore Alice
# Store Alice
##
# Expected end state:
# Both stored copies of Alice have 1 special note.
#####
{GENERIC_UNIT_TEST "special_note_count" (
    [event]
        name=start

        # The copies are called alice1 and alice2 so it's easy to work out which assert failed.
        [store_unit]
            [filter]
                id=alice
            [/filter]
            variable=alice1
        [/store_unit]
        {ASSERT ({VARIABLE_CONDITIONAL alice1.special_note.length equals 1})}

        [unstore_unit]
            variable=alice1
        [/unstore_unit]

        [store_unit]
            [filter]
                id=alice
            [/filter]
            variable=alice2
        [/store_unit]
        {ASSERT ({VARIABLE_CONDITIONAL alice2.special_note.length equals 1})}
    [/event]
) SIDE1_LEADER="Spectre"}
```

### Expected behavior

_No response_

### Additional context

_No response_

stevecotton:
Fixes #7153

The movetype's special notes were becoming a new note for the individual unit.

Clean up the documentation on some of movetype's functions, as they had documentation in both the .hpp and the .cpp.

### DIFF
--- a/data/test/scenarios/special_note_from_movetype.cfg
+++ b/data/test/scenarios/special_note_from_movetype.cfg
@@ -1,0 +1,43 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: [movetype][special_note],[store_unit],[unstore_unit]
+##
+# Actions:
+# Alice is a Spectre, a unit whose 'undeadspirit' movetype includes a special note.
+# Store Alice
+# Unstore Alice
+# Store Alice
+##
+# Expected end state:
+# Both stored copies of Alice have no special notes, because they are a part of the movetype
+# rather than a part of the unit.
+#####
+{GENERIC_UNIT_TEST "special_note_from_movetype" (
+    [event]
+        name=start
+
+        # The copies are called alice1 and alice2 so it's easy to work out which assert failed.
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=alice1
+        [/store_unit]
+        {ASSERT ({VARIABLE_CONDITIONAL alice1.special_note.length equals 0})}
+
+        [unstore_unit]
+            variable=alice1
+        [/unstore_unit]
+
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=alice2
+        [/store_unit]
+        {ASSERT ({VARIABLE_CONDITIONAL alice2.special_note.length equals 0})}
+
+        {SUCCEED}
+    [/event]
+) SIDE1_LEADER="Spectre"}
--- a/data/test/scenarios/special_note_individual_unit.cfg
+++ b/data/test/scenarios/special_note_individual_unit.cfg
@@ -1,0 +1,60 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: [effect][special_note],[store_unit],unstore_unit]
+##
+# Actions:
+# Alice is a Spectre, a unit whose 'undeadspirit' movetype includes a special note.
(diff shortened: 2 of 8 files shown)


data/test/scenarios/special_note_individual_unit.cfg
```diff
@@ -0,0 +1,58 @@
...
+
+        [object]
+            [filter]
+                id=alice
+            [/filter]
+            [effect]
+                apply_to=profile
+                [special_note]
+                    # This exact string is checked later, so don't translate it
+                    note="Haunts the Grey Woods." #wmllint: ignore
```
stevecotton, line 29: Remembered to flag this for wmllint, forgot about the schema validation.

newfrenchy83:
What,s the problem with shema validation?

stevecotton:
I deliberately put an untranslated string in, and forgot to tell the schema validation that it's deliberate (because I assert what the string is later). Will push once the local tests finish.

stevecotton:
> Hmm, I'm not sure about this. It seems somehow wrong? The function you changed is only called in two places (patch_movetype and unit::write), and I can't decide if your change is correct in both cases…

Any questions that I can answer to help you decide? Some of the documentation already added to the .hpp explains my reasons.

> I'm not quite sure what you're asking, but that code does look like it'll fail schema validation, because there is no `note` key. I guess that's not related to your question though?

Thanks, that hinted that I needed to exclude a much bigger block from the validation. I ended up putting the entire test in the `#ifndef` (will upload that once it finishes running locally).

CelticMinstrel:
Well… it's good that it's documented but it just means that movetype::write doesn't do what it says it does? I see you did at least take the time to verify that the change is correct in both places that call it…

CelticMinstrel:
I guess that's what I'm saying, yeah.

I think there's precedent for the boolean argument in other classes…?

stevecotton:
`unit::write` also had a `write_all` argument, which is what I'd based my suggestions on. I've looked at it again, and it's called `write_all` because `unit::write` would otherwise omit data that hasn't changed; this still means that `unit::write` supports round-tripping even with `write_all == false`.

I've added a boolean, but called it `include_notes`.

data/test/scenarios/special_note_from_movetype.cfg
```diff
@@ -0,0 +1,43 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: [movetype][special_note], [store_unit], [unstore_unit]
```
stevecotton, line 4: There's another push queued up to remove extra whitespace here, as noted in https://github.com/wesnoth/wesnoth/pull/7160#discussion_r1030584772

PR #7158, opened 2022-11-22
Add a unit test for the preprocessor's #ifdef handling

stevecotton:
This one's surprising, but it's the behavior of 1.16, so add a test for the current behavior even though I wonder if it should change in 1.17.

@CelticMinstrel @newfrenchy83 this explains my current confusion in #7157.

### DIFF
--- a/data/test/scenarios/test_preprocessor_ifdef_in_arg.cfg
+++ b/data/test/scenarios/test_preprocessor_ifdef_in_arg.cfg
@@ -1,0 +1,49 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: the preprocessor
+##
+# Expected end state:
+# This test passes. None of the FAILs happen.
+#####
+{GENERIC_UNIT_TEST "test_preprocessor_ifdef_in_arg" (
+    [event]
+        name=start
+
+        # Every {FAIL} block is removed during the preprocessing stage. Although the preprocessor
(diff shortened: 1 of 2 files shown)


stevecotton:
Failure this time is the Boost macos one.

ProditorMagnus:
There is similar case for defining macros.
```
#define __A
#define __B
#enddef
```
is valid syntax, but if you try to expand {__A} then it fails.

CelticMinstrel:
It looks okay. It's supposed to be testing that preprocessor statements don't work within a multi-line macro expansion, right?

stevecotton:
> It's supposed to be testing that preprocessor statements don't work within a multi-line macro expansion, right?

They do something, the test is that they still do the same as they did before, even if that's not a useful something.

stevecotton:
> {FAIL} is evaluated, but its contents are not written to output file.

I've rewritten a paragraph of the docs - is it good now?

ProditorMagnus:
Looks like it correctly explains what happens now. I think we should create issue to change this behavior if one does not already exist.

stevecotton:
Logged as #7166.

> There is similar case for defining macros.
> 
> ```
> #define __A
> #define __B
> #enddef
> ```
> 
> is valid syntax, but if you try to expand {__A} then it fails.

This one's #1352, and apparently is expected behavior.

PR #7156, opened 2022-11-22
More unit test documentation.

Pentarctagon:
---

Skipped over `test_force_chance_to_hit_macro` for now since it's changing in #7050.

PR #7160, opened 2022-11-23
Add unit tests for interpolated event names and end_turn

stevecotton:
Move `event_name_variable_substitution` to the top of the file, as it's testing that the event triggers at all, before the tests that check which order the events are triggered in.

The old `order_of_variable_events1` seemed to be a combination of two tests that should exist, but the code didn't quite test either of them. This replaces it with a new `order_of_variable_events1` and `order_of_variable_events3`.

Docs partially written by Pentarctagon, this started as a review comment on the documentation PR.

### DIFF
--- a/data/test/scenarios/premature_end_turn.cfg
+++ b/data/test/scenarios/premature_end_turn.cfg
@@ -1,0 +1,46 @@
+# wmllint: no translatables
+
+# If we were designing the API anew, then [end_turn] before side 1 has even started its turn would
+# probably be treated as an error. However, it's worked this way for ages and has been tested to
+# work this way since 1.13.11.
+
+#####
+# API(s) being tested: [event],[end_turn]
+##
+# Actions:
+# In a "turn X" event, before any side has started their "side turn", end the current side's turn.
+##
+# Expected end state:
+# It ends side 1's turn.
+#####
+{GENERIC_UNIT_TEST "premature_end_turn1" (
+    [event]
+        name="turn 1"
+        [end_turn][/end_turn]
+    [/event]
+    [event]
+        name="side 2 turn 1"
+        {SUCCEED}
+    [/event]
+)}
+
+#####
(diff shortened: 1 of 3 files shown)


data/test/scenarios/premature_end_turn.cfg
```diff
@@ -0,0 +1,46 @@
+# wmllint: no translatables
+
+# If we were designing the API anew, then [end_turn] before side 1 has even started its turn would
+# probably be treated as an error. However, it's worked this way for ages and has been tested to
+# work this way since 1.13.11.
+
+#####
+# API(s) being tested: [event], [end_turn]
```
Pentarctagon, line 8: I've not been adding a space after the comma, so at least for consistency it should be removed here as well. Nothing is actually trying to parse the values listed here though so in practice it doesn't really matter.

Pentarctagon:
I'd prefer if I can pull the changes to `test_event_names_and_order.cfg` over to #7156 to avoid merge conflicts, since I'm changing that file there as well.

data/test/scenarios/premature_end_turn.cfg
stevecotton: Ah, I'd missed that.

Without the comma, how do we separate tags which depend on their parent? For example, #7157 has tests for two different `[special_note]` tags:
* `[effect][special_note], [store_unit], [unstore_unit]`
* `[movetype][special_note], [store_unit], [unstore_unit]`

stevecotton:
Happy to do it in either direction, but I thought the changes were too big for a documentation PR.

data/test/scenarios/premature_end_turn.cfg
stevecotton: Oh, I misread your comment as "I've not been adding the comma". It makes sense now I've realised what it actually says; new version will be pushed in a half-hour or so.

PR #7162, opened 2022-11-24
Don't call back() if open_window_stack is empty since that's UB.

### LINKED ISSUE #7104: Wesnoth 1.16.6 Flatpak crashes after recruiting unit when running under Wayland
Noticed this with both the Flathub and Fedora registries although the bug appears slightly differently on both.

With Flathub on 1.16.6, the recruited unit appears, and then the gui disappears but the process doesn't end, and I get an error message that says `Error sending request: Broken pipe`.

The Fedora registry has 1.17.7, and on that the crash occurs before the unit appears with an error message is `/usr/include/c++/12/bits/stl_vector.h:1230: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::back() [with _Tp = gui2::window*; _Alloc = std::allocator<gui2::window*>; reference = gui2::window*&]: Assertion '!this->empty()' failed.`

Now that I'm writing this there's a chance that it should be two separate issues but they seem both isolated to the Flatpak installs since I don't have the issue when installed natively.

Pentarctagon:
Resolves #7104

### DIFF
--- a/src/gui/widgets/text_box_base.cpp
+++ b/src/gui/widgets/text_box_base.cpp
@@ -316,7 +316,8 @@ void text_box_base::cursor_timer_callback()
 			cursor_alpha_ = 255;
 			return;
 		default:
-			if(get_window() != open_window_stack.back()) {
+			// back() on an empty vector is UB and was causing a crash when run on Wayland (see #7104 on github)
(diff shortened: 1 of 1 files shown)


Vultraz:
Better `!open_window_stack.empty()`

Issue #5498, opened 2021-01-27
Add a capability for SUF's variation= key to filter units with no variation


MrTitainin:
Currently, there's no way to filter units with no variation using SUF's variation key. Things like `variation=""` don't work.

Current workarounds are:
`[filter_wml] variation = "" [/filter_wml]`
`[not] variation = explicit list of all possible variations [/not]`

Desired feature:
Make a way to include empty variation on the list, either as only element or as one of many, so it would include in the filter units without a variation for the code like:
`variation=""`
`variation=",some"` or `variation="",some`

It would also be fine to make some special value, for example "none", to be treated as no variation in this case, but for now "none" is a valid variation id so it might not be the best solution.

Soliton suggests that treating empty string as special case and changing split function's behaviour in this case to include empty elements on the list (there's a flag for it) could solve the problem.

CelticMinstrel:
I like using the empty string for "no variation" when trying to match just a single variation, but I'm not as happy with using that when trying to match a list of variations… so maybe "none" is better? Does any unit in any add-on actually use "none" as a variation name? I wouldn't expect it to be very common, at least…

soliton-:
I'm not a fan of using "none" when the empty string is quite clear, avoids surprises and special handling. Especially when already using the empty string in another place.

CelticMinstrel:
But `variation=,thing` is _not_ clear in my opinion. It would match the base variation as well as the "thing" variation. I do like matching a single variation as the empty string, but when it's a list of variations, that's not so great.

AI0867:
Would a reserved/magic `varation=base` be a better option? That would allow `variation=base,thing` and the like.

CelticMinstrel:
What makes "base" better than "none"? I think a reserved variation name is fine in any case.

MrTitainin:
I want to note, that unlike `base` and other possibilities, `none` is already kind of reserved, as it has special behaviour in {RECRUIT_UNIT_VARIATIONS} macro. I still prefer empty string though.

CelticMinstrel:
Another possibility is choosing a name that's not even a valid variation name, for example `"-"`.

AI0867:
> What makes "base" better than "none"? I think a reserved variation name is fine in any case.

[...]

PR #7167, opened 2022-11-26
Fix WML for standing animations for Walking Corpse and Soulless intro…

walodar:
…duced in b0d327ea3c78209397571f11773f056b272f496f (#7102).

### DIFF
--- a/data/core/units/undead/Corpse_Walking.cfg
+++ b/data/core/units/undead/Corpse_Walking.cfg
@@ -29,11 +29,12 @@
 #define UNIT_BODY_WALKING_CORPSE_GRAPHICS BASE_NAME
     image="units/undead/{BASE_NAME}.png"
     {DEFENSE_ANIM "units/undead/{BASE_NAME}-defend.png" "units/undead/{BASE_NAME}.png" {SOUND_LIST:ZOMBIE_WEAK_HIT} }
-    # This standing animation only exists for default zombie for now. As variations are added, the filter should be updated
+    # This standing animation only exists for the base unit and variation="mounted" for now. As variations are added, the two filters should be updated.
     [standing_anim]
         [filter]
+            #included are the base unit and variation="mounted"
             [not]
-                variation="drake,dwarf,falcon,goblin,gryphon,horse,rat,sand_scorpion,saurian,scorpion,serpent,spider,swimmer,troll,wolf,wose"
+                variation="bat,beast_rider,boar,drake,dwarf,falcon,goblin,gryphon,horse,rat,sand_scorpion,saurian,scorpion,serpent,spider,swimmer,troll,wolf,wose"
             [/not]
         [/filter]
         start_time=0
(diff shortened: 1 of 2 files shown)


Pentarctagon:
What's the issue being fixed?

CelticMinstrel:
I surmise that the standing animations of the touched variations are inheriting the filter from the standing animations of the base variation, which somehow breaks things.

walodar:
Hi, yes, there are variations missing in the filters, so the standing animations are inherited by variations where there are no corresponding image files.

For example, this screenshot shows an invisible beast_rider variation of the Walking Corpse.

invisibleWalkingCorpse

And since the bat variations have directional standing animations, the inherited standing animations have to be removed there (like the inherited defense animations).

data/core/units/undead/Corpse_Walking.cfg
```diff
@@ -29,11 +29,12 @@
 #define UNIT_BODY_WALKING_CORPSE_GRAPHICS BASE_NAME
     image="units/undead/{BASE_NAME}.png"
     {DEFENSE_ANIM "units/undead/{BASE_NAME}-defend.png" "units/undead/{BASE_NAME}.png" {SOUND_LIST:ZOMBIE_WEAK_HIT} }
-    # This standing animation only exists for default zombie for now. As variations are added, the filter should be updated
+    # This standing animation only exists for the base unit and variation="mounted" for now. As variations are added, the two filters should be updated.
     [standing_anim]
         [filter]
+            #included are the base unit and variation="mounted"
```
soliton-, line 35: Why is this done via `[not]`? No way to refer to the base unit?

walodar: Hi, yes, exactly. The logic with [not] was already there, and I have also wondered if there is a better way. But at least as far as I know, specifying the base unit doesn't seem to be possible in WML. Fortunately, this way works (if the filter is correct).

CelticMinstrel: I guess `variation=` without a `[not]` does not work? Or maybe `variation=""`?

walodar: Hi, I have now tried your ideas. Unfortunately, both don't seem to work. But I don't think that's a big deal, the version with [not] is perhaps not extremely elegant, but it works.

soliton-: Well, the issue is maintainability. As this PR shows it's likely that this filter will fall out of date again.

There is already issue #5498 though. So the filter can be improved once that is resolved.

walodar: Thanks, this issue / feature request fits very well.

The current errors in the filters all date back to the filters' original creation 16 days ago in b0d327e (#7102). I hope that simply updating the filters when a new animation is added is easier, since a new animation will probably also be tested.

But I see your point that this could be a potential problem.

walodar:
Hi, may I ask if there is a chance this could be merged for the next release? This pull request fixes #7190, which is labeled as a high-priority bug.

Pentarctagon:
Whoops, looks like this was lost track of. I'll merge it tomorrow sometime prior to tagging 1.17.11 assuming there are no other comments.

walodar:
Great, thank you!

PR #7168, opened 2022-11-27
[1.17]Add ability filter and 'priority' for overwrite_specials

newfrenchy83:
This request groups 3 commits which, once approved, will have to be merged separately into the master branch.

The first two commits are identical to https://github.com/wesnoth/wesnoth/pull/7050

The third concerns the addition of a filter for the overwrite_specials attribute which exploits the code of the first commit.

### DIFF
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
@@ -1,0 +1,110 @@
+#####
+# API(s) being tested: [damage]overwrite_specials=both_sides,[overwrite_filter]
+##
+# Actions:
+# Three [damage] abilities are added, one with value=5, other with add=11 and the last with multiply=2.
+# The one with multiply=2 have overwrite_specials=both_sides and [overwrite_filter] for overwrite [damage] with add=11 but not value=5.
+# Have alice attack bob.
+##
+# Expected end state:
+# Alice has 100 - (5*2)=90 hitpoints because add=11 overwrited but not value=5 who replace 9 pt of damage of bob.
+#####
+{GENERIC_UNIT_TEST "test_overwrite_specials_filter" (
+    [event]
+        name=start
+        [modify_unit]
+            [filter]
+            [/filter]
+            max_hitpoints=100
+            hitpoints=100
+            attacks_left=1
+        [/modify_unit]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=new_ability
+                [abilities]
+                    [damage]
+                        id=test_dmg
+                        value=5
+                    [/damage]
+                    [attacks]
+                        id=test_att
+                        value=1
+                    [/attacks]
+                    [damage]
+                        id=test_overwrite
+                        add=11
+                    [/damage]
+                    [damage]
+                        id=test_overwrite_specials
+                        multiply=2
+                        overwrite_specials=both_sides
+                        [overwrite_filter]
+                            [not]
+                                type_value=value
+                            [/not]
+                        [/overwrite_filter]
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                    [/chance_to_hit]
+                [/abilities]
+            [/effect]
+            [filter]
+                id=bob
+            [/filter]
+        [/object]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+            kill=yes
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        [unstore_unit]
+            variable=a
+            find_vacant=yes
+            x,y=$b.x,$b.y
+        [/unstore_unit]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+
+        [do_command]
+            [attack]
+                weapon=0
+                defender_weapon=0
+                [source]
+                    x,y=$a.x,$a.y
+                [/source]
+                [destination]
+                    x,y=$b.x,$b.y
+                [/destination]
+            [/attack]
+        [/do_command]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        {ASSERT ({VARIABLE_CONDITIONAL a.hitpoints equals 90})}
+        {SUCCEED}
+    [/event]
+)}
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
@@ -1,9 +1,9 @@
 #####
-# API(s) being tested: [damage]overwrite_specials=both_sides,[overwrite_filter]
+# API(s) being tested: [damage]overwrite_specials=both_sides,[overwrite][filter_specials]
 ##
 # Actions:
 # Three [damage] abilities are added, one with value=5, other with add=11 and the last with multiply=2.
-# The one with multiply=2 have overwrite_specials=both_sides and [overwrite_filter] for overwrite [damage] with add=11 but not value=5.
+# The one with multiply=2 have overwrite_specials=both_sides and [overwrite][filter_specials] for overwrite [damage] with add=11 but not value=5.
 # Have alice attack bob.
 ##
 # Expected end state:
@@ -40,11 +40,13 @@
                         id=test_overwrite_specials
                         multiply=2
                         overwrite_specials=both_sides
-                        [overwrite_filter]
-                            [not]
-                                type_value=value
-                            [/not]
-                        [/overwrite_filter]
+                        [overwrite]
+                            [filter_specials]
+                                [not]
+                                    type_value=value
+                                [/not]
+                            [/filter_specials]
+                        [/overwrite]
                     [/damage]
                     [chance_to_hit]
                         value=100
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -1488,7 +1488,7 @@ static bool type_value_if_present(const config& filter, const config& cfg)
 	return ( std::find(filter_attribute.begin(), filter_attribute.end(), cfg_type_value) != filter_attribute.end() );
 }
 
-static bool matches_ability_filter(const config & cfg, const std::string& tag_name, const config & filter)
+static bool matches_ability_filter(const config & cfg, const std::string& tag_name, const config & filter, bool ignore_tag_name)
 {
 	if(!filter["affect_adjacent"].empty()){
 		bool adjacent = cfg.has_child("affect_adjacent");
@@ -1509,17 +1509,20 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 	if(!bool_matches_if_present(filter, cfg, "cumulative", false))
 		return false;
 
-	const std::vector<std::string> filter_type = utils::split(filter["tag_name"]);
-	if ( !filter_type.empty() && std::find(filter_type.begin(), filter_type.end(), tag_name) == filter_type.end() )
-		return false;
+	if(!ignore_tag_name){
+		const std::vector<std::string> filter_type = utils::split(filter["tag_name"]);
+		if ( !filter_type.empty() && std::find(filter_type.begin(), filter_type.end(), tag_name) == filter_type.end() ){
+			return false;
+		}
+		if(!string_matches_if_present(filter, cfg, "overwrite_specials", "none")){
+			return false;
+		}
 
+	}
 	if(!string_matches_if_present(filter, cfg, "id", ""))
 		return false;
 
 	if(!string_matches_if_present(filter, cfg, "apply_to", "self"))
-		return false;
-
-	if(!string_matches_if_present(filter, cfg, "overwrite_specials", "none"))
 		return false;
 
 	if(!string_matches_if_present(filter, cfg, "active_on", "both"))
@@ -1547,25 +1550,25 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 	return true;
 }
 
-bool unit::ability_matches_filter(const config & cfg, const std::string& tag_name, const config & filter) const
+bool unit::ability_matches_filter(const config & cfg, const std::string& tag_name, const config & filter, bool ignore_tag_name) const
 {
 	// Handle the basic filter.
-	bool matches = matches_ability_filter(cfg, tag_name, filter);
+	bool matches = matches_ability_filter(cfg, tag_name, filter, ignore_tag_name);
 
 	// Handle [and], [or], and [not] with in-order precedence
 	for (const config::any_child condition : filter.all_children_range() )
 	{
 		// Handle [and]
 		if ( condition.key == "and" )
-			matches = matches && ability_matches_filter(cfg, tag_name, condition.cfg);
+			matches = matches && ability_matches_filter(cfg, tag_name, condition.cfg, ignore_tag_name);
 
 		// Handle [or]
 		else if ( condition.key == "or" )
-			matches = matches || ability_matches_filter(cfg, tag_name, condition.cfg);
+			matches = matches || ability_matches_filter(cfg, tag_name, condition.cfg, ignore_tag_name);
 
 		// Handle [not]
 		else if ( condition.key == "not" )
-			matches = matches && !ability_matches_filter(cfg, tag_name, condition.cfg);
+			matches = matches && !ability_matches_filter(cfg, tag_name, condition.cfg, ignore_tag_name);
 	}
 
 	return matches;
--- a/changelog_entries/extent_remove_ability_efect_to_attribute_changelog.md
+++ b/changelog_entries/extent_remove_ability_efect_to_attribute_changelog.md
@@ -1,0 +1,2 @@
+### WML Engine
+   * Add [overwrite_filter] to filter [abilities/specials] which may or may not be overwritten by the overwrite_specials attribute.
--- a/changelog_entries/extent_remove_ability_efect_to_attribute_changelog.md
+++ b/changelog_entries/extent_remove_ability_efect_to_attribute_changelog.md
@@ -1,2 +1,2 @@
 ### WML Engine
-   * Add [overwrite_filter] to filter [abilities/specials] which may or may not be overwritten by the overwrite_specials attribute.
+   * Add [overwrite] tags who contain 'priority' attribute for define priority between two abilities with overwrite_specials, and [overwrite][filter_specials] to filter [abilities/specials] which may or may not be overwritten by the overwrite_specials attribute.
--- a/data/core/macros/utils.cfg
+++ b/data/core/macros/utils.cfg
@@ -494,6 +494,31 @@
     {CLEAR_VARIABLE unit_type_table,unit_type_table_i,random_placement_location}
 #enddef
 
+#define FILTER_ABILITY_FORCE_CHANCE_TO_HIT AFFECT_WHO ATTACKER
+    # filter used in FORCE_CHANCE_TO_HIT macro below
+    tag_name=chance_to_hit
+    [and]
+        overwrite_specials=both_sides,one_side
+    [/and]
+    [and]
+        {AFFECT_WHO}
+    [/and]
+    [and]
+        apply_to=self,{ATTACKER}
+    [/and]
+#enddef
+
+#define FILTER_ABILITY_FORCE_CHANCE_TO_HIT_OPPONENT ATTACKER
+    # filter used in FORCE_CHANCE_TO_HIT macro below
+    tag_name=chance_to_hit
+    [and]
+        overwrite_specials=both_sides,one_side
+    [/and]
+    [and]
+        apply_to=opponent,{ATTACKER}
+    [/and]
+#enddef
+
 #define FORCE_CHANCE_TO_HIT FILTER SECOND_FILTER CTH_NUMBER EXTRA_CONDITIONS_WML
     # Invisibly forces certain units to always have a specific chance to hit
     # when fighting against certain other units.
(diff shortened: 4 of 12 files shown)


newfrenchy83:
@stevecotton @CelticMinstrel , i knowwhat i shouldn't doing that but i duplicated https://github.com/wesnoth/wesnoth/pull/7050 here and adde a commit for add a filter for overwrite_specials attribute

CelticMinstrel:
I'm of two minds on this change. On the one hand, it's certainly a natural outgrowth of the `[filter_ability]` capability, so it makes sense in that regard. However, it also undermines that very same capability, as you won't be able to filter on abilities that use `[overwrite_filter]`. So I'm not sure if this is a good idea.

newfrenchy83:
> I'm of two minds on this change. On the one hand, it's certainly a natural outgrowth of the `[filter_ability]` capability, so it makes sense in that regard. However, it also undermines that very same capability, as you won't be able to filter on abilities that use `[overwrite_filter]`. So I'm not sure if this is a good idea.

i added ignore_tag_name variable for ignore type of abilities in overwrite_filter

newfrenchy83:
@CelticMinstrel , @stevecotton , sorry toinsist then what you both on your own PR, but don't forget this, please.

src/units/unit.cpp
```diff
@@ -1408,6 +1408,138 @@ void unit::remove_ability_by_id(const std::string& ability)
...
+	if(!string_matches_if_present(filter, cfg, "add", ""))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "sub", ""))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "multiply", ""))
+		return false;
+
+	if(!string_matches_if_present(filter, cfg, "divide", ""))
```
CelticMinstrel, line 1496: Values should be matched as a number, not a string. Matching as `double` should be sufficient.

Also, I feel that a filter on `add=-2` should match an ability with `sub=2` and vice versa. Similarly, a filter with `divide=2` should match an ability with `multiply=0.5` and vice versa.

newfrenchy83: Except that the use of string allows to filter a list of values like value=10,20, etc.. for each attribute instead of having to write
value=10
[or]
value=20
[or]
etc..
that must be why neither int nor double are used in the encoding of [filter_weapon] in the units/attack_type.cpp file.
As for the equivalence add=2 sub=-2, I don't believe it should be coded in the source code whereas a wml developer can code
add=2
[gold]
sub=-2

if he deems it necessary, but this should not be systematic, given that I consider, for my part, that the filtering of the writing of the operation is as important as that of the operation itself for a minimum clarity for the wml developer and thus avoid confusing it and less risk of bugs than if I try to encode an equivalence that I consider useless.

CelticMinstrel: > Except that the use of string allows to filter a list of values like value=10,20, etc.. for each attribute

No, that's still wrong. If you write `value=10.0,20.0` your code won't work. There's also no way to encode "match a number between 1 and 12". Supporting a comma-separated list has nothing to do with the fact that you're interpreting it as a string. You can still support a comma-separated list even while interpreting it as a number, using the function `parse_ranges`. This also adds support for syntax like `1-12`.



> As for the equivalence add=2 sub=-2, I don't believe it should be coded in the source code whereas a wml developer can code add=2 [or] sub=-2
> 
> if he deems it necessary, but this should not be systematic, given that I consider, for my part, that the filtering of the writing of the operation is as important as that of the operation itself for a minimum clarity for the wml developer and thus avoid confusing it and less risk of bugs than if I try to encode an equivalence that I consider useless.

I don't really understand why you'd want to not consider equivalent things to be equivalent. An ability that applies `add=-2` does the exact same thing as an ability that applies `sub=2`. An ability that applies `sub=-2` does the exact same thing as an ability that applies `add=2`. An ability that applies `multiply=0.5` does the exact same thing as an ability that applies `divide=2`. Why would the way the developer chooses to express it affect whether the filter matches?

You say it would be confusing to code this equivalence. Is it not more confusing if your filter with `divide=2` _doesn't always_ match an ability that halves the value? Is it not more confusing if your filter with `sub=2` _doesn't always_ match an ability that subtracts 2 from the value?

Besides which, you've already coded a way to match `add` but not `sub` if you really want to, using `type_value=add`. So, if someone really wants a filter that matches `add=-2` but does not match `sub=2`, they can just write:

```
add=-2
type_value=add
```

…right?

---

I think there is also another issue lurking here, which is the fact that `value` and related keys can also be a formula instead of a number. I think this is a case that needs to be handled somehow… but I'm not sure what is the best way to do so.

newfrenchy83: it's done in https://github.com/wesnoth/wesnoth/pull/7168/commits/2f36bd118dda5e27ebbc5e6888bb71ab0c315196 commit

src/units/unit.cpp
```diff
@@ -1408,6 +1408,181 @@ void unit::remove_ability_by_id(const std::string& ability)
...
+	return ( std::find(filter_attribute.begin(), filter_attribute.end(), cfg[attribute].str(def)) != filter_attribute.end() );
+}
+
+static bool double_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, const double& def)
+{
+	if(filter[attribute].empty()) {
+		return true;
+	}
+
+	return filter[attribute].to_double() == cfg[attribute].to_double(def);
```
stevecotton, line 1436: Floating-point numbers are a little inaccurate, so shouldn't be compared with the `==` operator.

Oddly enough, we don't seem to have a utility function for this in src/utils/math.h. However, that file does already have `in_ranges`, which might be the right function to use combined with the range-parsing functions.

stevecotton:
@CelticMinstrel I haven't looked yet at whether this concept should be added to the API or not. Before @newfrenchy83 spends too much time fixing small issues, @CelticMinstrel please would you confirm that you want to add this to the API? As in, if the small fixes are done, will the PR be ready to merge?

src/units/unit.cpp
newfrenchy83: for int, yes but not for double

newfrenchy83:
> 

Like i said before filter must be for matches this what be writed in wml code,, not the real value, and i prefer what convert 'value' attribute filter from string to double reported for another PR, because i already wait for this PR will merged.

src/units/unit.cpp
newfrenchy83: i removed commit because i feels what i causes more issue, i i would what this PR finally merged before to resolve that

newfrenchy83:
> @CelticMinstrel I haven't looked yet at whether this concept should be added to the API or not. Before @newfrenchy83 spends too much time fixing small issues, @CelticMinstrel please would you confirm that you want to add this to the API? As in, if the small fixes are done, will the PR be ready to merge?

i removed latest commit because small issues.

CelticMinstrel:
> @CelticMinstrel I haven't looked yet at whether this concept should be added to the API or not. Before @newfrenchy83 spends too much time fixing small issues, @CelticMinstrel please would you confirm that you want to add this to the API? As in, if the small fixes are done, will the PR be ready to merge?

I do think there should be a more fully-featured way of removing abilities in an effect. This is quite different from my original vision of it (which was just to use a WML filter, ie `config::matches`), but I don't think that's a bad thing. I do intend to merge this at some point once we've finished working out all the kinks.

src/units/unit.cpp
newfrenchy83: https://github.com/wesnoth/wesnoth/pull/7168/commits/15907b7f2141b16b529559d8fa17413d214cfb7a must be reponseto issue

src/units/unit.cpp
newfrenchy83: @CelticMinstrel i re pushed new version of conversion in https://github.com/wesnoth/wesnoth/pull/7168/commits/beee74e564f5cac3be3c571419a7f755cf2d9ca8 and re-ordered commit (titles with [1.17] must be showed inmaster after merging)

src/units/unit.cpp
CelticMinstrel: It's unfortunate that `in_ranges` supports double but `parse_ranges` does not, so I added 73a1b46bbfeecf24e5fb3c4b95438ade1f7a7064. Does that help?

newfrenchy83: @CelticMinstrel Yes, the compilation works fine, and frankly, it bothers me to have to specify every value even in a list instead of defining a range within which any included value matches. Thanks @CelticMinstrel for your help.

newfrenchy83: i pushed changes in PR

CelticMinstrel:
About considering `multiply=0.5` the same as `divide=2`:

I still think this is desirable behaviour, but there are some challenges to implementing it reasonably, so let's leave it for later.

The eventual implementation would need to both avoid floating-point inaccuracy issues (probably by working with rational numbers) and deal with decimal inaccuracy as well (to detect that `divide=3` is the same as `multiply=0.33333`). So it'd be quite a lot of work to implement.

The equivalence of `add=-2` and `sub=2` has no such problems, so we could either keep that in here or leave it for later.

CelticMinstrel:
Given that you are also adding an `[overwrite_filter]`, how useful is it to still be able to filter on `overwrite_specials`? I don't mind having the facility if you think it could be useful sometimes, but filtering on abilities that use an `[overwrite_filter]` clearly isn't something we can do very easily, and my impression is that `[overwrite_filter]` is intended to replace `overwrite_specials`.

But I might be wrong on some detail and ending at a wrong conclusion?

newfrenchy83:
> Given that you are also adding an `[overwrite_filter]`, how useful is it to still be able to filter on `overwrite_specials`? I don't mind having the facility if you think it could be useful sometimes, but filtering on abilities that use an `[overwrite_filter]` clearly isn't something we can do very easily, and my impression is that `[overwrite_filter]` is intended to replace `overwrite_specials`.
> 
> But I might be wrong on some detail and ending at a wrong conclusion?

Replace or nt overwrite_specials is not intended yet, and same with filter an ability with attribute can't overwrite  other who have also 'overwrite_special'(it could be overwrite itself). I prefer use overwrite_filter  for complete attribut insread of replace that because of deprecation message.

newfrenchy83:
> Given that you are also adding an `[overwrite_filter]`, how useful is it to still be able to filter on `overwrite_specials`? I don't mind having the facility if you think it could be useful sometimes, but filtering on abilities that use an `[overwrite_filter]` clearly isn't something we can do very easily, and my impression is that `[overwrite_filter]` is intended to replace `overwrite_specials`.
> 
> But I might be wrong on some detail and ending at a wrong conclusion?

@CelticMinstrel If I have to replace the overwrite_specials attribute by overwrite_filter it will not be in this PR because it will require more modification than simply implementing the filter and anyway we have [affect_adjacent] which works associated with the affect_allies/enemies attribute encoded at the exterior of the filter. The question is rather to know if we can consider providing overwrite_specials with an option to be able to overwrite abilities which also have this attribute while being more selective thanks to [overwrite_filter]

stevecotton:
> it will not be in this PR because it will require more modification than simply implementing the filter

What will use this, and how much more will the complexity grow? I'm worried that we'll get to the end, and then find that we have a lot more code to maintain, when it would be cleaner to just implement some new abilities in C++ alongside the 1.16 WML API.

I'll admit, without the investigation into how much the WML API would have to expand then I would probably have rejected the simple implementation in C++.

newfrenchy83:
> > it will not be in this PR because it will require more modification than simply implementing the filter
> 
> What will use this, and how much more will the complexity grow? I'm worried that we'll get to the end, and then find that we have a lot more code to maintain, when it would be cleaner to just implement some new abilities in C++ alongside the 1.16 WML API.
> 
> I'll admit, without the investigation into how much the WML API would have to expand then I would probably have rejected the simple implementation in C++.

@stevecotton for me, ad filter to overwrite_specials is sufficient i do not entend to modify the code after that.

newfrenchy83:
@stevecotton As I said before, adding [filter_overwrite] in addition to the overwrite_specials attribute is enough for me, so as not to complicate the task of wml developers, this filter can be used in addition to the attribute but is not essential for its operation, and despite questions from @CelticMinstrel, I do not intend to evolve [filter_overwrite] so that it replaces overwrite_specials.

newfrenchy83:
@CelticMinstrel , @stevecotton could you give your opinion on https://github.com/newfrenchy83/wesnoth/commit/0624bd297227c359d27b4dea6fe76750302e6ced please? This method is more simple or not?

src/units/unit.cpp
stevecotton: Having looked into [the implementation of `effect::effect`](https://github.com/wesnoth/wesnoth/pull/7050#issuecomment-1365145483), I'm thinking there should only be `[filter_ability]add=` which matches both `[ability]add=` and `[ability]sub=`, with no corresponding `[filter_ability]sub=`.

Also based on the implementation, `add` should only support integers.

CelticMinstrel: It doesn't seem terrible to support both `add` and `sub` though as near-synonyms.

Making them be integers only seems fine.

CelticMinstrel:
I almost forgot, but a new schema type is needed for (at least) `multiply` and `divide`. (Can `value` have real numbers, or only integers?)

It would be almost the same as the existing `s_range_list` but allows decimal points. This code goes in schema/game_config.cfg, presumably just after the definition of `s_range_list`.

```
    # definition of real_range_list and s_real_range_list, before macros are expanded
    {LIST_TYPE_COMPLEX range (
        [element]
            value="\d+(\.\d+)?"
        [/element]
        [element]
            value="\d+(\.\d+)?-\d+(\.\d+)?"
        [/element]
        [element]
            value="\d+(\.\d+)?-infinity"
        [/element]
    )}
```

That won't support scientific notation or some other edge cases, but I don't really think it needs to even if the engine technically allows it. (Who's going to use scientific notation in an ability definition?)

newfrenchy83:
> I almost forgot, but a new schema type is needed for (at least) `multiply` and `divide`. (Can `value` have real numbers, or only integers?)
> 
> It would be almost the same as the existing `s_range_list` but allows decimal points. This code goes in schema/game_config.cfg, presumably just after the definition of `s_range_list`.
> 
> ```
>     # definition of real_range_list and s_real_range_list, before macros are expanded
>     {LIST_TYPE_COMPLEX real_range (
>         [element]
>             value="\d+(\.\d+)?"
>         [/element]
>         [element]
>             value="\d+(\.\d+)?-\d+(\.\d+)?"
>         [/element]
>         [element]
>             value="\d+(\.\d+)?-infinity"
>         [/element]
>     )}
> ```
> 
> That won't support scientific notation or some other edge cases, but I don't really think it needs to even if the engine technically allows it. (Who's going to use scientific notation in an ability definition?)

added to shema and sorry for late but i was sick.

stevecotton:
> sorry for late but i was sick

Please take care of yourself. Not just about when you're ill, but also that you don't overprioritise working on Wesnoth, particularly when CelticMinstrel and I take so much time to respond. Sorry about how much time this has taken.

data/test/scenarios/wml_tests/ScenarioWML/EffectWML/test_remove_ability_by_filter.cfg
```diff
@@ -0,0 +1,132 @@
...
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        [unstore_unit]
+            variable=a
+            find_vacant=yes
```
stevecotton, line 101: These unit tests can use a `COMMON_KEEP` macro and `simulate_combat` to reduce the amount of boilerplate code, and so make it easier to see what's being tested. `simulate_combat` also means it can test the damage and number of blows rather than having to check Alice's remaining HP.

I think I posted code before without the other parts needed to make it work (probably missing the `COMMON_KEEP`), so here's a full working example:

```
#####
# API(s) being tested: [abilities][overwrite_specials][overwrite_filter] 
##
# Actions:
# Bob is given three [damage] abilities, one with value=5, other with add=11 and the last with multiply=2.
# The one with multiply=2 has overwrite_specials=both_sides and [overwrite_filter] to overwrite [damage]add=11 but not to overwrite [damage]value=5.
# Have bob attack alice.
##
# Expected end state:
# Bob's attack does 10 damage per hit, because the add=11 was overwritten but the value=5 and multiply=2 are still active.
#####
{COMMON_KEEP_A_B_UNIT_TEST "test_overwrite_specials_ability" (
    [event]
        name=start

        [object]
            silent=yes
            [effect]
                apply_to=new_ability
                [abilities]
                    [damage]
                        id=test_dmg
                        value=5
                    [/damage]
                    [attacks]
                        id=test_att
                        value=1
                    [/attacks]
                    [damage]
                        id=test_overwrite
                        add=11
                    [/damage]
                    [damage]
                        id=test_overwrite_specials
                        multiply=2
                    [/damage]
                    [overwrite_specials]
                        id=test_ability
                        [overwrite_filter]
                            tag_name=damage
                            type_value=add
                        [/overwrite_filter]
                    [/overwrite_specials]
                    [chance_to_hit]
                        value=100
                    [/chance_to_hit]
                [/abilities]
            [/effect]
            [filter]
                id=bob
            [/filter]
        [/object]

        [lua]
            code=<<
                local alice = wesnoth.units.find({id="alice"})[1]
                local bob = wesnoth.units.find({id="bob"})[1]

                -- Bob attacks Alice, doing it this way round ensures that it's melee rather than ranged combat.
                _, _, att_weapon, _ = wesnoth.simulate_combat(bob, alice)
                unit_test.assert_equal(att_weapon.num_blows, 1, "Bob’s weapon should only have 1 attack")
                unit_test.assert_equal(att_weapon.damage, 10, "Bob’s weapon should do 10 damage per hit")
            >>
        [/lua]

        {SUCCEED}
    [/event]
)}
```

newfrenchy83:
@stevecotton , @CelticMinstrel if you judge that the hardcoding is satisfactory and that only the writing of the tests can be improved with your new codes, I suggest that you merge this PR and propose the test modifications yourself instead of going through me and dragging the merging that has been pending for this PR since November 27, not to mention two others that have been waiting for much longer. Sorry if I'm starting to show signs of impatience but given the time you take to answer sometimes, and there, it's again to offer me to rewrite a test code for the umpteenth time (if it was hardcoding, I will understand but there!).

Please excuse me for my outburst but understand that my patience also has its limits.

newfrenchy83:
@stevecotton could i expect what this PR commited before next release, please?

stevecotton:
Sorry, but I'm not even sure whether we're going to decide #7271 before the next release.

newfrenchy83:
> Sorry, but I'm not even sure whether we're going to decide #7271 before the next release.

i don't speak of https://github.com/wesnoth/wesnoth/pull/7271, but i expect what tis PR was merged before prestring 1.18

stevecotton:
Ah, release as in 1.18, rather than 1.17.13? Okay, yes that should happen.

Are you writing add-ons using these features? Seeing what the intended uses will be helpful, although it might also mean we discuss other ways to implement the same abilities.

newfrenchy83:
> Ah, release as in 1.18, rather than 1.17.13? Okay, yes that should happen.
> 
> Are you writing add-ons using these features? Seeing what the intended uses will be helpful, although it might also mean we discuss other ways to implement the same abilities.

No, i don't write addon. I have think to write my personnal sequel to Iftu/Ats but i never want to publish that by respect to @irydacea

newfrenchy83:
I prefer what this PR merged instead of https://github.com/wesnoth/wesnoth/pull/7223. Pehaps what i remove later.

newfrenchy83:
@stevecotton , @CelticMinstrel new commit added here.

CelticMinstrel:
It feels like `[filter_ability]` should have a matching `[filter_ability_active]`.

newfrenchy83:
> It feels like `[filter_ability]` should have a matching `[filter_ability_active]`.

except what like [special_active it should have what id and type for filtering an ability active, and i don't see how doing that.

newfrenchy83:
> It feels like `[filter_ability]` should have a matching `[filter_ability_active]`.

@CelticMinstrel  added by https://github.com/wesnoth/wesnoth/pull/7168/commits/b1cd22fdc2411ab5443fedf5ffb47d5c6ba3b638

newfrenchy83:
@stevecotton , @CelticMinstrel this request is as complete as possible and I do not see what we could add. Could you take a look at it before phase near 1.18?

CelticMinstrel:
It looks fine to me.

newfrenchy83:
> It looks fine to me.

@CelticMinstrel , @stevecotton could you merge this PR then?

stevecotton:
> Are you writing add-ons using these features? Seeing what the intended uses will be helpful, although it might also mean we discuss other ways to implement the same abilities.

This is still the big question for me - who is planning to create units that use this feature?

Apologies if I've misunderstood, but I feel it's adding complexity to the engine without assessing whether it could be done with scenario-specific coding, in the few places where that is needed. Even if the scenario-specific coding would be ugly, there's value to keeping the engine smaller.

newfrenchy83:
> > Are you writing add-ons using these features? Seeing what the intended uses will be helpful, although it might also mean we discuss other ways to implement the same abilities.
> 
> This is still the big question for me - who is planning to create units that use this feature?
> 
> Apologies if I've misunderstood, but I feel it's adding complexity to the engine without assessing whether it could be done with scenario-specific coding, in the few places where that is needed. Even if the scenario-specific coding would be ugly, there's value to keeping the engine smaller.

i open poqt in forum: https://forums.wesnoth.org/viewtopic.php?p=681412#p681412 but i one reply only.

src/units/filter.cpp
```diff
@@ -806,6 +806,47 @@ void unit_filter_compound::fill(vconfig cfg)
...
+						if(args.u.ability_matches_filter(ab.cfg, ab.key, c.get_parsed_config())) {
+							return true;
+						}
+					}
+					return false;
+				});
+			}
+			else if (child.first == "filter_ability_active") {
+				create_child(child.second, [](const vconfig& c, const unit_filter_args& args) {
+				assert(display::get_singleton());
```
stevecotton, line 821: Either return or just skip this test instead of asserting.

newfrenchy83: added return and fix inentation

stevecotton:
I feel that `[filter_ability]` and `[filter_ability_active]` are good additions to the API, and result in easy-to-understand filter code in UMC. The `[remove_abilities]` change also seems good, and that commit contains a lot of the filter implementation. I think taking just those also covers what @Toranks is asking for.

I haven't looked into the `overwrite_filter` change yet, but AFAICS the other commits can be taken without it.

I haven't looked into the `FORCE_CHANCE_TO_HIT` macro change commit yet.

The question of how `add` and `sub` should interact with each other could become part of #7271's discussion, I think it probably fits better there. However, we can merge this and change it depending on that discussion.

With the unit tests, I added comments, moved them to a separate file and renamed (avoiding `_fail` as per #7597). Those can be found in newfrenchy83/wesnoth#81 .

How would everyone feel about:
* merging the three commits needed for `[filter_ability]`
* leaving the `overwrite_filter` and `FORCE_CHANCE_TO_HIT` commits unmerged but open for further discussion
* merging in newfrenchy83/wesnoth#81

The whole lot would get squashed down to either 1 commit or 2 (with the "extend ability removal with remove_ability" commit staying separate, and all the test changes getting merged into the "add in `[filter]` events" commit).

newfrenchy83:
> I feel that `[filter_ability]` and `[filter_ability_active]` are good additions to the API, and result in easy-to-understand filter code in UMC. The `[remove_abilities]` change also seems good, and that commit contains a lot of the filter implementation. I think taking just those also covers what @Toranks is asking for.
> 
> I haven't looked into the `overwrite_filter` change yet, but AFAICS the other commits can be taken without it.
> 
> I haven't looked into the `FORCE_CHANCE_TO_HIT` macro change commit yet.
> 
> The question of how `add` and `sub` should interact with each other could become part of #7271's discussion, I think it probably fits better there. However, we can merge this and change it depending on that discussion.
> 
> With the unit tests, I added comments, moved them to a separate file and renamed (avoiding `_fail` as per #7597). Those can be found in [newfrenchy83#81](https://github.com/newfrenchy83/wesnoth/pull/81) .
> 
> How would everyone feel about:
> 
>     * merging the three commits needed for `[filter_ability]`
> 
>     * leaving the `overwrite_filter` and `FORCE_CHANCE_TO_HIT` commits unmerged but open for further discussion
> 
>     * merging in [Add test docs for filter_ability newfrenchy83/wesnoth#81](https://github.com/newfrenchy83/wesnoth/pull/81)
> 
> 
> The whole lot would get squashed down to either 1 commit or 2 (with the "extend ability removal with remove_ability" commit staying separate, and all the test changes getting merged into the "add in `[filter]` events" commit).

i wait to see checking result before to push rebased commits

newfrenchy83:
> How would everyone feel about:
> 
>     * merging the three commits needed for `[filter_ability]`
> 
>     * leaving the `overwrite_filter` and `FORCE_CHANCE_TO_HIT` commits unmerged but open for further discussion
> 
>     * merging in [Add test docs for filter_ability newfrenchy83/wesnoth#81](https://github.com/newfrenchy83/wesnoth/pull/81)

done

stevecotton:
Thanks for rebasing and squashing commits.

@CelticMinstrel any objection to me merging the first of the 3 commits tonight, but leaving the other two open?

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
```diff
@@ -0,0 +1,110 @@
+#####
+# API(s) being tested: [abilities]['weapon']overwrite_specials filter [overwrite_filter]
```
Pentarctagon, line 2: It looks like this would more accurately be:
`[damage]overwrite_specials=both_sides,[overwrite_filter]`

newfrenchy83: corrected

CelticMinstrel:
> @CelticMinstrel any objection to me merging the first of the 3 commits tonight, but leaving the other two open?

No objection.

newfrenchy83:
@stevecotton , @CelticMinstrel I just added the commit https://github.com/wesnoth/wesnoth/pull/7168/commits/5dc64b9582bc7f186b3dc770ed47773378679dd9 in the request to give a priority system within overwrite_specials itself.
I then realize that if the 'traditional' specials (inside the [specials] tag) could also use this feature (so far only abilities can do it) then the priority system that I propose for plagues and [damage_type] would be obsolete because the overwrite_specials system would be more than enough. So I'm going to set up this reciprocity system since with [overwrite] priority, the refusal to give this function to [specials] loses its justification.

If this PR could then be merged into the main branch quickly enough, then I could rebase https://github.com/wesnoth/wesnoth/pull/7003 and close https://github.com/wesnoth/wesnoth/pull/7484

Issue #7037, opened 2022-10-03
Boost tests randomly failing on macos


stevecotton:
### Game and System Information

- **Version:** 1.17.8+dev , 880183774c0974b4d2ebdde8a09eab7dd3049516 + unrelated change to `run_wml_tests`
- **Build info:** This happened on the CI build of PR #7036
- **OS:** macos-intel (macos-11.0, Debug)

### Retitled bug

This issue has become a general "boost tests intermittently failing on macos debug and macos release builds". It's probably only one root cause, so the same issue.

### Description of the original bug / probable root cause of the retitled bug

test_gui2/test_last reports that dialogs weren't tested if they fail during construction.

This is an error message in a situation where the CI is going to fail anyway, so the bug is just that the message is confusing.

A run of the Boost tests started off with the first few tests timing out. As it's the first few, I'm guessing that it may have been sensitivity to the amount of time it takes to build the cache rather than anything about the code under test:

```
Run ./run_boost_tests --path=projectfiles/Xcode/build/"$CFG" --executable=unit_tests
  ./run_boost_tests --path=projectfiles/Xcode/build/"$CFG" --executable=unit_tests
  shell: /bin/bash -e {0}
  env:
    CFG: Debug
Test test_fire_event timed out: None
Test test_gui2_iterator timed out: None
Test test_gui2/modal_dialog_test_addon_auth timed out: None
Test test_gui2/modal_dialog_test_addon_connect timed out: b"20221003 09:52:38 error display: Theme 'Default' not found. Falling back to default theme.\n20221003 09:52:38 error display: Default theme not found.\n"
Test test_gui2/modal_dialog_test_addon_license_prompt passed
```
...
```
/Users/runner/work/wesnoth/wesnoth/src/tests/gui/test_gui2.cpp:671: error: in "test_gui2/test_last": check widget_list.size() == 0 has failed [2 != 0]
Window 'addon_auth' registered but not tested.
Window 'addon_connect' registered but not tested.
```

### Expected behavior

Either log that the dialog's test started (so that `test_gui2/test_last` passes), or improve the error message.

Improving the error message is definitely the easier option. Logging that the dialog's test started needs its id, which is obtained via a virtual function, so gets skipped if the constructor times out.

Pentarctagon:
I'm also somewhat curious exactly when and why this started being an issue. It definitely wasn't happening, then now seems like it happens semi-frequently, though randomly.

stevecotton:
[...]

PR #7175, opened 2022-11-29
Change Twitter link to Mastodon link in README.md

cooljeanius:
Pretty simple; since the project has announced a migration from Twitter to Mastodon, this updates the README to reflect it.


stevecotton:
The build failure is unrelated to this change, but is still somewhat interesting. It's on macos, but it's not in the Boost tests (#7037). One of the WML unit tests ran successfully, but then failed to load the replay when checking that the replay gets the same results.

Issue #6974, opened 2022-08-21
Pin comments in "this game" chat


xSkyEndx:
### description:
When you host a game it can sometime take a while till someone joins. You might go aft in the meanwhile, but want to let the player know, who joins your game, **that you're aft** for 10 min or something. Also it can be used to **share game informations** that you can't or don't want to put in the title. (explain what ladder is or what maps you have in your picker etc.)

### feature:
To pin comments in "this game" chat (that's the chat when you hosted a game and wait for the start)

CelticMinstrel:
So basically, "pin" here means the comment is visible to players who weren't there when it was originally sent, right?

I think the second use-case (custom game information) is better served by a game description set in the game creation screen, but "pinned" comments also seem useful.

xSkyEndx:
Yeah that what I mean. And your right that also can be an option.

Kingofd:
I'll see to it

stevecotton:
Instead of pinning, an alternative would be to for newly-joined clients to see all the chat that had happened before they joined. I'm wondering if that would do everything that pinning would, and be simpler than adding the pinning.

soliton-:
Chat message replay has now been implemented.

PR #7177, opened 2022-11-29
Pin comments chat

Kingofd:
This is my first shot at the "pinning comments" functionality:
https://github.com/wesnoth/wesnoth/issues/6974

Open questions:

1. I need to overload the two functions I implement in `chat_events.hpp`, which I need for `chatbox.cpp` also in `menu_events.hpp` where they effectively are not doing anything. Maybe there is a better way of doing this
2. Still searching for a good place for a client after leaving a game to tell it to delete its logs. Otherwise the players will always have access to the last game's chat logs

### DIFF
--- a/src/chat_command_handler.cpp
+++ b/src/chat_command_handler.cpp
@@ -154,4 +154,13 @@ void chat_command_handler::do_mp_report() {
 	}
 }
 
+void chat_command_handler::do_pin_message() {
+	chat_handler_.pin_message(get_data(), "", allies_only_);
 }
+
+void chat_command_handler::do_unpin_messages() {
+	//chat_handler_.unpin_messages();
+}
+
+}
+
--- a/src/chat_command_handler.cpp
+++ b/src/chat_command_handler.cpp
@@ -155,11 +155,11 @@ void chat_command_handler::do_mp_report() {
 }
 
 void chat_command_handler::do_pin_message() {
-	chat_handler_.pin_message(get_data(), "", allies_only_);
+	chat_handler_.pin_message(get_data(), "");
 }
 
 void chat_command_handler::do_unpin_messages() {
-	//chat_handler_.unpin_messages();
+	chat_handler_.unpin_messages();
 }
 
 }
--- a/src/chat_command_handler.hpp
+++ b/src/chat_command_handler.hpp
@@ -43,7 +43,9 @@ protected:
 	void do_version();
 	void do_clear_messages();
 	void do_mp_report();
-
+	void do_pin_message();
+	void do_unpin_messages();
+	
 	/** Request information about a user from the server. */
 	void do_info();
 
@@ -119,6 +121,11 @@ protected:
 			_("Request information about a nickname."), _("<nickname>"));
 		register_command("clear", &chat_command_handler::do_clear_messages,
 			_("Clear chat history."));
+		register_command("pin", &chat_command_handler::do_pin_message,
+			_("Pin a message."), _("<message>"));
+		register_command("unpin", &chat_command_handler::do_unpin_messages,
+			_("Unpin a message."), _("<message>"));
+
 	}
 private:
 	chat_handler& chat_handler_;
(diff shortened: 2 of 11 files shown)


knyghtmare:
Question: since this is a user-requested feature, how are you allowing it? You have not mentioned any procedure of test or screenshots showing the working pin and unpin.

Kingofd:
So I add 2 possible commands to the MP-chat:

1. /pin <message>
2. /unpin

With pin, the message is send to everyone present in the lobby, but - crucially - also to the game-object, which stores the message.
If a new player now should join the game it will be sent all the pinned messages.

If a player, that has pinned any messages will type `/unpin` all of the pinned messages will be cleared from the game-object, as will all pinned messages if the game object is destroyed or someone leaves the game.

The server accepts now messages with the "pin" subtype and will send it to the relevant game object.

No whisper-pins are allowed, no pins in the "lobby"-chat window are allowed, only in the standard "this game"-chat window.
pin_game_func
pin_game_func1
unpin

Kingofd:
One small "feature" / "issue" that still persists but I am unsure how to handle is, that since I change the way the "this game" standard log is allocated, something like this will happen, that you have the previous game chat in a little window-tab in the lobby screen:
Log_functionality

knyghtmare:
Hmm...suppose I want to add this:
```ini
[chat]
    speaker="Game"
    message= _ "/pin Welcome to WesLeague: League of Legends style scenario in Wesnoth"
[/chat]
```
Will it work with this feature?

Kingofd:
I am not quite sure how the speaker pipeline with commands works yet, I would need to test it. But as long as `preferences::login()` returns "Game" it should work.

In what context would that command invoke `chat_command_handler();`?

soliton-:
This does not seem very sensible to me I'm afraid. From a pin command I would expect that it marks an existing message as pinned and that there is some way to recognize that it is a pinned message.

soliton-:
> Hmm...suppose I want to add this:
> 
> ```ini
> [chat]
>     speaker="Game"
>     message= _ "/pin Welcome to WesLeague: League of Legends style scenario in Wesnoth"
> [/chat]
> ```
> 
> Will it work with this feature?

This feature is for the game lobby before a game starts. Not related to any scenario WML.

Kingofd:
> This does not seem very sensible to me I'm afraid. From a pin command I would expect that it marks an existing message as pinned and that there is some way to recognize that it is a pinned message.

Yes the name implies some other uses than it really provides. It is basically "on-join-send-chat-message". That however is a way clunkier name than "pinning". But I am always open for renaming the feature if there is a better name for it.

Otherwise if it is about recognising pinned comments one could accommodate that by prefacing "pinned" comments with a string

Pentarctagon:
Is this feature still be pursued?

Kingofd:
It needs either a code review, or closing, with a new issue opened for the solution suggestion for the original issue #6974

Pentarctagon:
stevecotton already mentioned the alternative in #6974 I think? I'm not sure why a new issue would need to be opened.

soliton-:
I would indeed prefer if the issue was solved rather by recording commands for the pre-game lobby.

Kingofd:
Could you elaborate what you mean by "recording commands"?

soliton-:
I mean similar to [replay data](https://wiki.wesnoth.org/ReplayWML) which is just a recording of the `[command]`s that get sent around. I guess in the pre-game lobby it's actually `[message]`s though. (When the game started it's `[command] [speak]`)

Basically record each `[message]` sent and send that recording to each new client that joins (similar to sending them the replay for started games) so it replays those messages. I think the messages currently have no timestamp so it would make sense to add that to record the original time they were sent.

Kingofd:
Ah, now I get it, I will get to it.

Replay already sent messages in pre game lobby and add timestamps.
I will assume that timestamps are not to be shown in the chat window only recorded in the "backend".

soliton-:
Timestamps are shown when you enable them under Preferences -> Multiplayer.

PR #7181, opened 2022-11-30
Just look for the toplevel dir tag

soliton-:
"campaign_name" is not an existing attribute in the given config anymore. An addon should only contain a single toplevel dir tag with the name of the addon anyway.

### DIFF
--- a/src/server/campaignd/addon_utils.cpp
+++ b/src/server/campaignd/addon_utils.cpp
@@ -125,10 +125,11 @@ void find_translations(const config& base_dir, config& addon)
 
 void add_license(config& cfg)
 {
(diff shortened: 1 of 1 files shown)


soliton-:
The CI failure is some unrelated unit test (test_basic_lua_aspect) producing a corrupted replay.

stevecotton:
Not sure who should review. It looks good to me, but I'm unfamiliar with the campaignd code.

stevecotton:
Just triggered a rerun of the CI failure, because while it was unrelated, having the red X in the PR list makes it less likely to be reviewed (IMO).

PR #7182, opened 2022-11-30
Document last batch of unit tests.

Pentarctagon:


data/test/scenarios/test_assert.cfg
```diff
@@ -79,7 +65,7 @@
 )}
 
 #####
-# API(s) being tested: unit_test.fail(),[fire_event]
+# API(s) being tested: [fire_event]
```
stevecotton, line 68: The old version of this was right. It is testing `unit_test.fail()`, and checking that mechanism persists the failure when returning from the fired event.

data/test/scenarios/test_lua_wml.cfg
```diff
@@ -36,6 +45,15 @@
     [/event]
 )}
 
+#####
+# API(s) being tested: wesnoth.wml_conditionals
+##
+# Actions:
+# Register a new WML cconditional tag [foo] with the optional attribute "bar".
```
stevecotton, line 52: Typo on "cconditional".

data/test/scenarios/test_move_unit.cfg
```diff
@@ -258,6 +267,15 @@
     [/event]
 )}
 
+#####
+# API(s) being tested: [move_unit]
+##
+# Actions:
+# Move alice in a circle.
```
stevecotton, line 274: Move alice in a circle, returning to his starting hex in a single move.

Explanation that could be added before the last ASSERT in this test:
        # The assert can't check that Alice moved, but it does check that the implementation of
        # [move_unit] didn't choose an unoccupied hex next to the (initially-occupied) starting hex.

data/test/scenarios/test_movetype.cfg
```diff
@@ -332,29 +332,164 @@
     [/modify_unit]
 #enddef
 
+#####
+# API(s) being tested: N/A
```
stevecotton, line 336: All of the `MOVETYPE_MOVE_SCEN` tests in this file are also testing `[find_path]`.

data/test/scenarios/test_movetype.cfg
```diff
@@ -332,29 +332,164 @@
...
+# Changing the vision costs doesn't affect the movement costs.
+# Changing the sequence of applying changes to movement cost and vision cost does not affect the outcome.
+#####
+{MOVETYPE_MOVE_SCEN test_elf_fast_cave_longsighted_movement (Elvish Archer) (1,1,2,2,2,2,2,2,1,1,1,3,1) {ORC_COSTS} (
+    {SET_MOVE_COSTS cave=1}
+    {SET_VISION_COSTS hills=1}
+)}
 
+#####
+# API(s) being tested: N/A
```
stevecotton, line 445: All of the `MOVETYPE_VISION_SCEN` tests in this file are also testing `[store_locations][filter_vision]`.

data/test/scenarios/test_role_types.cfg
```diff
@@ -1,3 +1,14 @@
+#####
+# API(s) being tested: [role],[recall]role=
+##
+# Actions:
+# Spawn some units into the recall list.
+# Assign those unit types to a role.
+# Recall using the role.
+##
+# Expected end state:
+# The first unit listed in the role is recalled.
```
stevecotton, line 10: This isn't [how roles work](https://wiki.wesnoth.org/InternalActionsWML#.5Brole.5D), all of the docs in this file need a rewrite. The `[role]` tag chooses one specific unit at the time the `[role]` tag is read.

Suggestion for this test:
```
#####
# API(s) being tested: [role],[recall]role=
##
# Actions:
# Spawn some units into the recall list.
# Assign a role, choosing a unit based on the unit type.
# Recall using the role.
##
# Expected end state:
# The role has been assigned to the unit whose type appeared first on the list.
# That unit is recalled.
#####
```

data/test/scenarios/two_plus_two.cfg
```diff
@@ -1,5 +1,9 @@
-# This test asserts that 2 + 2 = 4
-
+#####
+# API(s) being tested: N/A
```
stevecotton, line 2: Both of these tests are testing `[set_variable]add=`

data/test/scenarios/unknown_scenario_warning.cfg
```diff
@@ -95,12 +108,42 @@
 #enddef
 
 # The numbers in the names are (number of broken branches) (number of ok branches) (last scenario)
+#####
+# API(s) being tested: [endlevel]next_scenario
+##
+# Expected end state:
+# The next_scenario values are all identical, so log strict is broken with a warning referencing the singular next scenario.
```
stevecotton, line 115: One of the next_scenario values is valid (empty string indicates the end of the campaign), so log strict is broken with a warning saying only some of the next scenarios weren't found.

data/test/scenarios/unknown_scenario_warning.cfg
```diff
@@ -95,12 +108,42 @@
...
+#####
 {TEST_UNKNOWN_SCENARIO "unknown_scenario_1_0"}
+#####
+# API(s) being tested: [endlevel]next_scenario
+##
+# Expected end state:
+# The next_scenario values are not all identical, so log strict is broken with a warning saying only some of the next scenarios weren't found.
```
stevecotton, line 122: One of the next_scenario values is valid (empty string indicates the end of the campaign), so log strict is broken with a warning saying only some of the next scenarios weren't found.

data/test/scenarios/unknown_scenario_warning.cfg
```diff
@@ -95,12 +108,42 @@
...
+#####
 {TEST_UNKNOWN_SCENARIO "unknown_scenario_1_0_last" SCEN2=""}
+#####
+# API(s) being tested: [endlevel]next_scenario
+##
+# Expected end state:
+# The next_scenario values are not all identical, so log strict is broken with a warning saying only some of the next scenarios weren't found.
```
stevecotton, line 129: One of the next_scenario values is valid, so log strict is broken with a warning saying only some of the next scenarios weren't found.

data/test/scenarios/unknown_scenario_warning.cfg
```diff
@@ -95,12 +108,42 @@
...
+#####
 {TEST_UNKNOWN_SCENARIO "unknown_scenario_1_1" SCEN2="test_return"}
+#####
+# API(s) being tested: [endlevel]next_scenario
+##
+# Expected end state:
+# The next_scenario values are all identical, so log strict is broken with a warning saying only some of the next scenarios weren't found and lists both scenarios which weren't found.
```
stevecotton, line 136: There is more than one next_scenario value, but all of them are invalid, so log strict is broken with a warning saying all of the next scenarios weren't found and lists both scenarios which weren't found.

(The comment is wrong about all of the values being identical, one of them has an `_2` on the end.)

wml_test_schedule
```diff
@@ -397,6 +396,10 @@
 0 unknown_scenario_false_positives
 0 unknown_scenario_interpolated
 9 unknown_scenario_1_0
+9 unknown_scenario_1_0_last
+9 unknown_scenario_1_1
+9 unknown_scenario_2_0
+9 unknown_scenario_1_1_last_null
```
stevecotton, line 402: While these can be added, they're slow (as broke-strict tests have to be run individually), and the automatic testing mechanism won't see the difference between them. They show error dialogs that are intended to be read by end-users (players, not just WML authors), and the different tests show slightly different error messages for manual testing when the tester is reading what's displayed on the UI.

However, if the aim is to have a scripted check that all tests defined in data/test/scenarios are run then I understand adding them.

data/test/scenarios/unknown_scenario_warning.cfg
```diff
@@ -1,6 +1,14 @@
...
 
-# Check that the "this will lead to an unknown scenario" warning doesn't get triggered.
+#####
+# API(s) being tested: [endlevel]next_scenario=
+##
+# Expected end state:
+# The endlevel for empty_test doesn't print a warning (it exists as another unit test scenario).
+# The endlevel for $chosen_branch doesn't print a warning (it not existing is accepted).
+# The endlevel for victory doesn't print a warning.
+# The dummy containing a next_scenario is ignored.
```
stevecotton, line 10: The wording feels wrong here, not because the wording itself is wrong but because these tests are inconsistent with the general unit tests. If I was reading this along with the docs of other tests, I would assume this was executing the `[endlevel]` tags.

Might be clearer if it was `API(s) being tested: [scenario]`, because in the nearest ActionWML analogy to what's being tested here, it's checking an action while "executing" the `[test]` tag.

data/test/scenarios/unknown_scenario_warning.cfg
```diff
@@ -38,7 +46,12 @@
     [/event]
 )}
 
-# Not a branching scenario, but the only route uses a variable
+#####
+# API(s) being tested: [scenario]next_scenario
+##
+# Expected end state:
+# The endlevel for $chosen_branch doesn't print a warning (it not existing is accepted).
```
stevecotton, line 53: The variable being undefined is accepted. This is to avoid a false positive triggering the warning, with the assumption being that something will set that value while the scenario is running.

data/test/scenarios/test_terrain_mask.cfg
```diff
@@ -141,6 +274,15 @@ Hh, Hh"#enddef
 )}
 
 # mask_file= works (this is the same as the above test in all other respects)
+#####
+# API(s) being tested: [terrain_mask][rule]
+##
+# Actions:
+# Use a terrain mask rule to change existing terrain to either Ko or Co.
```
stevecotton, line 281: This one's missing from `wml_test_schedule`, and its name expands to `test_terrain_mask_file`.

These docs are wrong (based on running the test in interactive mode), it's also adding hills.

data/test/scenarios/test_assert.cfg
Pentarctagon: My thought here was that `unit_test.fail()` is already being tested in test_return.cfg, so removing it here narrows the number of APIs being tested, but that's true.

data/test/scenarios/test_role_types.cfg
Pentarctagon: Alright, will update. Given the text on the wiki in part says:
> However, there is no guarantee that roles will ever be assigned.

I interpreted that to mean the role was evaluated at time of use rather than when it's defined.

wml_test_schedule
```diff
@@ -5,7 +5,6 @@
 1 test_return_fail
 0 test_assert
 1 test_assert_fail
-1 test_assert_fail_two
```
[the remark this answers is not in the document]
Pentarctagon: This test did exist, I just removed it from test_assert.cfg since it was duplicated between `test_return_fail` (tests `{RETURN ([not][true][/true][/not])}`) and `test_assert` (tests `{ASSERT ([true][/true])}`).

wml_test_schedule
Pentarctagon: My thought was "they're there and have a valid end state they're testing, so why not?". There are a couple test scenarios that don't (they're solely meant for someone to run manually) which I'm not sure whether they should be in this directory and where to put them otherwise, but that's a separate thing.

data/test/scenarios/test_scoped_vars.cfg
```diff
@@ -1,3 +1,15 @@
+#####
+# API(s) being tested: wml_utils.scoped_var
+##
+# Actions:
+# Set test_var to 1.
+# Declare a scoped variable of test_var.
+# Set test_var to 5.
+##
+# Expected end state:
+# After the scoped variable is declared, test_var has no assigned value.
```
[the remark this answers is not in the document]
Pentarctagon: IIRC it's that a non-existent value can be represented as an empty string in certain circumstances, but then reverse isn't true.

@CelticMinstrel do you know?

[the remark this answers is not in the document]
stevecotton: Just realised that the answer is easy to get - just uncomment the `[inspect]`s in the test itself.

Your documentation is correct - at the first `ASSERT` within the `[args]` block, there's no WML variable called `test_var`.

Pentarctagon:
All comments have been addressed I believe, so barring anything else I'll go ahead and merge this later today.

data/test/scenarios/test_movetype.cfg
stevecotton: This one is unresolved. It might have got skipped over because it looked like the `MOVETYPE_MOVE_SCEN` comment, which was resolved.

data/test/scenarios/test_role_types.cfg
stevecotton: There's 5 tests in this file, the other 4 still need a similar update.

wml_test_schedule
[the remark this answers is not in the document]
stevecotton: Ah, ok. I suggest this addition, but will approved the PR without:
```
# Check that a typo in a test name in wml_test_schedule would be reported as a fail
9 there_is_no_test_with_this_name
```

data/test/scenarios/test_role_types.cfg
Pentarctagon: I believe they've been updated?

data/test/scenarios/test_move_unit.cfg
```diff
@@ -280,6 +298,8 @@
             dir=sw,s,se,ne,n,nw
         [/move_unit]
 
+# The assert can't check that Alice moved, but it does check that the implementation of
+# [move_unit] didn't choose an unoccupied hex next to the (initially-occupied) starting hex.
```
stevecotton, line 302: Please run wmlindent on this file (it isn't run automatically on the test dir, IIRC because it doesn't handle some of the Lua code correctly).

data/test/scenarios/test_role_types.cfg
```diff
@@ -41,6 +53,17 @@
     [/event]
 )}
 
+#####
+# API(s) being tested: [role],[have_unit]role=
+##
+# Actions:
+# Spawn some units on the map.
+# Assign those unit types to a role, choosing a unit based on the unit type.
```
stevecotton, line 61: "Assign a role to a unit" rather than "assign unit types to a role". A role is a just a property of a unit, roles aren't objects that can have lists of unit types assigned to them.

data/test/scenarios/test_role_types.cfg
```diff
@@ -75,6 +98,16 @@
     [/event]
 )}
 
+#####
+# API(s) being tested: [role],[have_unit]role=
+##
+# Actions:
+# Create a role with no types assigned.
```
stevecotton, line 105: A role isn't an object, it's not something that can be created.

Try to assign a role, but using a filter that doesn't match any unit (there are only 2 sides in this test).

data/test/scenarios/test_role_types.cfg
```diff
@@ -75,6 +98,16 @@
...
+##
+# Expected end state:
+# The role fails to choose a unit based on the unit type.
```
stevecotton, line 108: It's not the unit type, it's the `side=`.

data/test/scenarios/test_role_types.cfg
```diff
@@ -96,6 +129,18 @@
     [/event]
 )}
 
+#####
+# API(s) being tested: wesnoth.wml_actions.role
+##
+# Actions:
+# Spawn some units into the recall list.
+# Assign those unit types to a role, choosing a unit based on the unit type.
```
stevecotton, line 137: Again, this should be "assign a role to a unit".

data/test/scenarios/test_role_types.cfg
stevecotton: I've added comments to the individual tests, so marking this thread as resolved.

Pentarctagon:
Alright, *now* everything should be addressed.

data/test/scenarios/test_terrain_mask.cfg
```diff
@@ -13,139 +13,281 @@ raw#endarg
...
+    ignore_special_locations=yes
+[/terrain_mask]
+{ASSERT (
+    [not]
+        [have_location]
+            x,y=9,1
+            location_id=foo
+        [/have_location]
+    [/not]
+)}
```
stevecotton, line 158: This file (in other parts too) is looking worse for having wmlindent run on it.

stevecotton (approved):

data/test/scenarios/test_terrain_mask.cfg
Pentarctagon: Alright, I'm going to merge then. I went through and reverted a couple changed files since wmlindent definitely wasn't interpreting some of the lua parts correctly, but otherwise this is how it's handled in the rest of the codebase.

Issue #7169, opened 2022-11-27
Add-on language filter duplicates


Pentarctagon:
Add-ons are not consistent in the language codes they use, which results in there being both `it` and `it_IT` for example, which results in the language filter list containing duplicates for Italian. The language filter list shouldn't display these duplicates.

soliton-:
It should remember a list of language codes per display string then so that once you choose an entry it picks the appropriate language codes.

Another option would be to add the language code to the display string. This way the user can have a little better understanding of what they pick. For example if they pick an entry with language code `pt` they may be able to guess that it's actually not clear what version of portuguese the translation will be. And it will be possible to filter for a specific version of portuguese (i.e. `pt_BR` or `pt_PT`) or any (`pt`).

Kingofd:
The solution I thought of is pushing the `lang_code_to_string()` output trough yet another set.

stevecotton:
I'm in two minds about the options that @soliton- identified. My feeling is to merge #7185 (which implements the first option) and let Github automatically close this issue; however someone can reopen it if they want the second option implemented.

PR #7185, opened 2022-12-01
Language string duplicate removal

Kingofd:
Attempts to fix this issue here: https://github.com/wesnoth/wesnoth/issues/7169
I just push the output of `langcode_to_string()` through another set, which removes duplicates of the display string.

### DIFF
--- a/src/gui/dialogs/addon/manager.cpp
+++ b/src/gui/dialogs/addon/manager.cpp
@@ -408,7 +408,7 @@ void addon_manager::pre_show(window& window)
 			language_strings_available.insert(lang_code_string);
 		}
 	}
-	for (const auto& i: language_strings_available) {
+	for (auto& i: language_strings_available) {
 		language_filter_types_.emplace_back(language_filter_types_.size(), std::move(i));
 	}
 	// The language filter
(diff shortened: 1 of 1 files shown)


src/gui/dialogs/addon/manager.cpp
```diff
@@ -399,15 +399,18 @@ void addon_manager::pre_show(window& window)
...
 		// Only show languages, which have a translation as per langcode_to_string() method
 		// Do not show tranlations with their langcode e.g. "sv_SV"
+		// Also put them into a set, so same lang strings are not producing doublettes
 		if (std::string lang_code_string = langcode_to_string(i); !lang_code_string.empty()) {
-			language_filter_types_.emplace_back(language_filter_types_.size(), std::move(lang_code_string));
+			language_strings_available.insert(lang_code_string);
 		}
 	}
-
+	for (const auto& i: language_strings_available) {
```
stevecotton, line 411: The next line does a `std::move` on a `const`.

stevecotton (approved):

soliton-:
I don't really understand how this can be correct. Is this code duplicated somewhere to recalculate the correct mapping from display string to language code?

stevecotton:
It's duplicated in get_lang_filter_visibility(), which allows the add-on to be visible if the translated name of the language matches. Not ideal, but it works.

Issue #7029, opened 2022-09-29
Support use of savefiles as test scenario to check engine compatibility


demario-w:
### Describe the desired feature

As test procedures are hard to write, we are sitting on tons of savefiles that contains numerous combinations of units/traits/capabilities/terrains/ToD/... where the expected behavior is known: the result stored in the savefile.

As an illustration, the way to identify the incompatible way of rounding xp according traits was by hitting OOS during replay of 1.14 saves on BFW 1.16. If that test had been run during development, the problem would have been caught before release.

If a savefile can be replayed in background mode, the run terminates and the output shows if an OOS has occurred or not, all these saves could be used as test scenarios to verify the behavior of the game engine in various situations. All problems in engine do not lead to a straightforward OOS (as the complete game state is not available in the save) but most of them will ultimately ends up in a OOS (missing attack target, difference in damage, unmatched death, target hex cannot be reached, gold not enough...). If the number of saves used as tests is large enough, the likeliness that a given problem in the engine will end up in creating an OOS in one of them is high.

Most of the available savefiles are not generated on the version of the engine where the development is done. So if this trove of saves need to be exploited to generate test scenario, there is a need that the engine is supporting replay of past version saves.

Many known saves have been identified as catching backward-incompatibilities in the engine and they would be the first targets to be turned into scenarios to test against regression. 

Being able to replay old saves also allows to exercise the engine in ways that is not possible with saves from the current version by picking saves from game scenarios that haven't been ported to the current version (which is the master branch).

demario-w:
For the record, here is how I start a replay from a save right now:

$ wesnoth --core=core_1_14 --with-replay --validcache --load $i

where $i hold the name of the save file.
I need 4 more "clicks" (or hotkey press) to switch to the next file: "skip animation", "continuous replay" and double click on the button to close the window at the end. Can this number be reduced further?

The result of the test (not OOS) is not passed back to the calling shell at this time.

stevecotton:
It sounds like a good feature.

The current `--unit` test mechanism in `game_launcher::single_unit_test()` has something similar, but which is only used immediately after running the unit-test scenarios. Each test is recorded with the hardcoded filename of "unit_test_replay", and then it checks that it can be replayed without OOS.

CelticMinstrel:
Yeah, so the functionality already exists in the engine, which means it just needs to be exposed.

demario-w:
[...]

PR #7187, opened 2022-12-02
Add option to run a test based on replaying a specified save (feature #7029)

demario-w:
There is no description for the new command line option to avoid breaking translation as the change was implemented for branch 1.16. But a recent change in src/game_launcher.cpp makes the code incompatible for that branch now (feature #7029).

--------
The path for the replay is relative to wesnoth save directory.

The solution is based on:

    - adding a command line option 'replay_test'
    - putting the code to replay a save from game_launcher::single_unit_test()
      in a standalone function game_launcher::replay_test()
    - calling that function if the option --replay_test is passed
    - defaulting to skip animation ('skip_replay' in the code) when running
      a test based on save ('is_unit_test', by using replay_test or matching
      the replay part of unit_test)
    - adding a parameter to not linger to replay_controller::replay_controller
      after the end of of the game
    - do not linger at the end when running a test based on save
      (no_linger=is_unit_test_)
    - if any OOS is encountered during replay, the replay stops and the game
      lingers

The command line option --replay_test has no description.

The compatibility with the various use cases for --unit_test should be checked.


gfgtdf:
I want to point out that replay compatibility is not something that we guarantee during dev releases. So this might catch quite a few false positives when added to the ci.

Pentarctagon:
Even that wouldn't be terrible to be honest, as long as people update it and note what broke. A basic replay test of something really simple (ie: Freelands) would be a good baseline too.

Issue #7108, opened 2022-11-09
Players should be given the option to disable allied movement orbs.


Ashmyr:
### Describe the desired feature

I'm now playing on 1.16 on Windows.

I've talked about this on Discord.

I personally don't want to see the movements left on the allied units. It makes me think that I'm controlling the side before they start moving every single time. I can intuit how much movement they have left myself. I personally find the new feature rather confusing.

I tried going through preferences, but if I disable or change the allied orb color, no matter what, when it's the ally's turn, their units are shown with the green and red movement orbs, which I do not want to see because it's confusing to me. I also tried going through the code, but Knyghtmare has informed me this isn't in the game's code so I can't edit it.

Although I'm sure most people like the new feature, it's considerably confusing to me and I would honestly just prefer if there was an option to choose how it was before -- allied orbs have one color, and stay that way. I think that if such a big feature is introduced, players should generally have the option to go to the old version.

cooljeanius:
Personally I think there should be a compromise option, too: something like a blue outline (for being allied) around a green/yellow/red inner core, or maybe reuse the new half-orbs added for #5155 where one half is blue and the other is green/yellow/red. Or something. I dunno.

stevecotton:
Reusing the half-orbs is a lot easier than adding a new preference.

stevecotton:
Have you tried out the new orbs on 1.17, or are you waiting until they're backported to 1.16? I'd like to check what the feedback is before backporting.

cooljeanius:
> Have you tried out the new orbs on 1.17, or are you waiting until they're backported to 1.16? I'd like to check what the feedback is before backporting.

Uh, which of us are you asking, @Ashmyr or me?

stevecotton:
Both :D

knyghtmare:
Tested it on 1.17.12+dev. Seems enabling the option makes the allied side orbs behave the same as they did in 1.10.x-1.12.x

stevecotton:
Erm, you do mean *disabling* the option, right?

Are you happy with the UI once you've configured it to your preferenceL

knyghtmare:
> Erm, you do mean disabling the option, right?
> Are you happy with the UI once you've configured it to your preference?

Yes, and yes. Seems quite clear and understandable.

cooljeanius:
So wait how exactly do I turn this on again? I tried switching to the 1.17 branch, but it doesn't look like the allied orbs have changed at all since 1.16...

[...]

PR #7189, opened 2022-12-03
Use two-color orbs for allies' units

stevecotton:
They no longer look the same as the player's own units. Possible fix for #7108, and the plan is to backport this to 1.16 (it's improving the UI, not revealing more information that would be shown to players of earlier 1.16 versions).

This probably doesn't need a change to the in-game help, at least for the en_US version of the documentation. The current text in data/core/help.cfg is:

* Blue for allied units, except during that ally's own turn.
* During the ally's own turn, their units will be shown with the colors showing whether the units can still move and attack; however their moves, and the corresponding orb changes, are delayed as explained in "Shroud and Fog of War'.

The left castle shows what this PR does, the right castle shows the colors the other way round.
allied_two_color_orbs


newfrenchy83:
i have a problem, i have two color orb for unit of my side when i tap attacks_left=0 in debug mode, but what unit has yet movement point.

stevecotton:
> i have two color orb for unit of my side when i tap attacks_left=0 in debug mode, but what unit has yet movement point.

That's the disengage skill (#5155, #5523), at least I'm assuming you're seeing a red/yellow orb.

newfrenchy83:
> > i have two color orb for unit of my side when i tap attacks_left=0 in debug mode, but what unit has yet movement point.
> 
> That's the disengage skill (#5155, #5523), at least I'm assuming you're seeing a red/yellow orb.

ah, yes, it's smae without this PR.

Vultraz:
I am finding this quite hard to see... especially green+blue.

newfrenchy83:
> I am finding this quite hard to see... especially green+blue.

it's true what except highest resolution two color in same orb is hard to see, use palette of color for allies instead of two color will be perhaps best, no?

stevecotton:
> I am finding this quite hard to see... especially green+blue.

Any better on units that don't have a gold crown? Sorry for forgetting to include one of those in the sample image.
image

I don't want to add more colors, particularly if they only appear fleetingly.

Using the current PR's orange+blue "ally with partial movement" orb instead of green+blue for "ally with full movement" would be possible. Thanks to @CelticMinstrel for a comment that I'm [taking somewhat out of context](https://github.com/wesnoth/wesnoth/issues/5155#issuecomment-779540580):
> The distinction between "full" and "partial" just doesn't matter very much. In fact, you could get rid of green orbs entirely and I don't think much would be lost (other than what you mentioned about rest healing).

@doofus-01 pointed this out about icons, in [a discussion about the disengage orb](https://github.com/wesnoth/wesnoth/issues/5155#issuecomment-774546254):
> I see the logic behind the boots, but I wonder if there isn't reason to keep the orb shape, at least to some extent? Changing to a small item/icon makes it look like another decoration, like a crown, indicating something new. It also might be hard to read at a glance, especially if there are more non-orb movement-icons (and if there aren't, it will probably look a bit odd).

doofus-01:
I agree the effect is hard to see, but the problem it's trying to solve seems really subtle too, so maybe it's appropriate?

If #7108 is about getting confused whether it's your turn or not, is there some other indicator that can be used - something in the UI instead of the unit orbs?  There's already the little flag icons, but something more obvious could help.

Ashmyr:
I started #7108. It's moreso that I was used to my allies always having blue orbs, and me having green, yellow and red orbs. I prefer the old version, where my allies' orbs never changed and I could easily tell my units apart. I don't like seeing the movement left on allied units, I can pretty easily intuitively tell which ones have moved. I find the current system, where allies have the same movement orbs as me, rather confusing. A preference to the old option would be something I appreciate, although I think this half-orb thing is a good compromise.

stevecotton:
Converted back to draft because I'm prototyping an advanced preference that allows selecting any of these:

* Don't show ally orbs at all
* Use the two-color when that ally is playing **(this is the default)**
* Always use the one-color (with ally color, not status)

That's now configured as a pair of toggles, not a tri-state. If ally orbs aren't shown, then both the new setting and the existing orb-color setting will have no effect. If somebody else sent me the "prototype" commit for review then my response would be "nice, but add comments and see if adding comments gives you better ideas for the function names"; it's not up to standard yet.

I feel the two-color code is nice and I'd be happy to backport it, while the code for a new preference feels messy, and will still feel messy even after tidying it up.

doofus-01:
Alright, I don't really understand the 7108 issue then.  I don't have any objections or anything to offer, so I approve of whatever stevecotton comes up with.  If it turns out to be a mistake, there's still time to fix it.

stevecotton:
New preference, and made the white bit of text at the bottom smaller to match all the other text.

image

CelticMinstrel:
The fact that the ally orb colour selection doesn't disable when the "Allied units' orbs show" box is ticked looks like a bug because the label's description makes it sound like ticking that box won't use the selected ally orb colour. However, from what you said earlier, with that box ticked it still gets the ally colour when it's not its turn, right? Is there some way we can explain that more clearly?

Perhaps the correct solution is actually to shorten the label and put the detailed description in the tooltip? So the button would be labelled something like this:

* [x] Allied units' orbs show movement status

And the tooltip would say something like this:

> If ticked, the orb on allied units will display their unmoved/partial/moved status when it's that ally's turn. They will display an orb in the ally orb colour when it's not their turn.

stevecotton:
Saying "movement status" seems less clear than "moved/partial/unmoved" in the context of this specific dialog, because the user already needs to have understood what those terms mean to understand the first three settings on this dialog.

The ally color is always used; even during the ally's turn it's still used for one of the two colors on the orb.

None of the other color selections disable when their "Show ... orb" is unticked either.

CelticMinstrel:
> The ally color is always used; even during the ally's turn it's still used for one of the two colors on the orb.

I see.

I'm sure the wording could be improved, but I would definitely prefer it to explain very clearly exactly what the preference does.

src/units/drawer.cpp
```diff
@@ -62,6 +62,31 @@ std::unique_ptr<image::locator> get_orb_image(orb_status os)
...
+std::unique_ptr<image::locator> get_playing_ally_orb_image(orb_status os)
+{
+	if(!orb_status_helper::prefs_show_orb(orb_status::allied))
+		return nullptr;
+
+	if(!preferences::show_status_on_ally_orb())
+		return get_orb_image(orb_status::allied);
+
+	if(os == orb_status::disengaged)
+		os = orb_status::partial;
```
stevecotton, line 79: Going to rework this conditional, which might push this PR back until after #7261. The conditional should be replaceable with a comment, because `orb_status_helper::get_orb_color` should have logic to provide a single color for the disengaged state. As noticed in #7261, there's currently an option in game_config.cfg that's a leftover from the idea of using a new single-color orb for disengaged.

Backporting this PR is in any case postponed until after 1.16.7, because of the new translatable string in the preferences dialog.

stevecotton:
The two-color ally orbs don't change the minimap.

The disengaged orb does, but uses a single color that's set (in both `game_config.cfg` and in the preferences dialog's `select_orb_colors::post_show`) to the same color as the partial orb.

stevecotton:
Part of the problem with the preferences dialog is that it's using checkbox labels as section headers to indicate what the rows of colors refer to. However, adding actual section headers would take space and probably not look any better.

@Vultraz, what do you think of the below use of spacers to convey that some checkboxes are headings and some are just refinements of the previous setting? The implementation is currently on top of this PR, but if it seems reasonable then I'll split out the spacing and disengage change and make it a new PR that goes before this one. The intention would be to backport it, so that the current PR can be backported too.
image

stevecotton:
Will rebase on top of PR #7266 if the feedback looks positive for that PR.

stevecotton:
Rebased on top of the other PR.

As of the latest push, the possible combinations of those two checkboxes have these four results:

* Two-color when that ally is playing, single color (the ally color) at other times **(default)**
* Two-color when that ally is playing, no orb at other times
* Always use the one-color orb (the blue ally color, never the green/orange/red status colors)
* Don't show ally orbs at all (this has been selectable in the advanced preferences since 1.13.0)

stevecotton:
Rebased on the now-merged other PR.

src/units/drawer.cpp
stevecotton: Outdated, although it's not be automatically marked as such because this line didn't change. It's now covering the edge-case that the user has disabled the normal partial orb, but still wants to show the disengage orb as a reminder for the few units that have that skill.

stevecotton:
I think all of the comments have now been addressed.

> I'm sure the wording could be improved, but I would definitely prefer it to explain very clearly exactly what the preference does.
> 
> Perhaps some of the other toggles, or the colour selections, could also use tooltips.

In addition to the revised layout, the wording is now "During ally’s turn, use a two-color orb to show movement" with tooltip "During your allies’ turns, their units have a two-color orb. One half is the ally orb color, and the other half shows whether the unit is unmoved, partially moved or fully moved."

The new checkbox still has a purpose even when the main "show ally orb" is disabled. Enabling that combination means that the ally orbs are hidden when it's not that ally's turn, but the two-color orbs are shown during the turn of that ally.

PR #7192, opened 2022-12-04
Avoid throwing exceptions internally to lexical_cast when parsing an empty string

CelticMinstrel:
The motivation for this is that `lexical_cast` is called very frequently, even to the point of being called repeatedly in the main loop. By relying on exceptions internally, it becomes a hindrance to debugging.

This PR is mainly opened solely for CI purposes. Unless there are objections before the CI successfully completes, I will merge it immediately.

### DIFF
--- a/src/lexical_cast.hpp
+++ b/src/lexical_cast.hpp
@@ -220,6 +220,14 @@ struct lexical_caster<
 	{
 		DEBUG_THROW("specialized - To long long - From std::string");
 
+		if(value.empty()) {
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Using exceptions for non-exceptional cases is also not good though. That said, you do have a point that special-casing the empty string is a simpler option that should still solve the specific issue I observed, and I'm also not a fan of switching to the C API… plus there are still CI failures, so I'll give that approach a try.

PR #7193, opened 2022-12-04
1.17 elevated terrain Help Topic

doofus-01:
The terrain elevation feature is different than the usual terrain graphics.  So this is for a help topic in the "Map and Scenario Editor->Terrain Editor" tree.

### DIFF
--- a/data/core/editor/help.cfg
+++ b/data/core/editor/help.cfg
@@ -11,7 +11,7 @@
 [section]
     id=editor_mode_terrain
     title= _ "Terrain Editor"
-    topics=..editor_mode_terrain, editor_palette, editor_masks, editor_tool_paint, editor_tool_fill, editor_tool_select, editor_tool_paste, editor_tool_starting, editor_terrain_overlays, editor_deprecated_overlay
+    topics=..editor_mode_terrain, editor_palette, editor_masks, editor_tool_paint, editor_tool_fill, editor_tool_select, editor_tool_paste, editor_tool_starting, editor_terrain_overlays, editor_terrain_elevation, editor_deprecated_overlay
     sort_topics=no
 [/section]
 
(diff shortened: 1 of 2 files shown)


Pentarctagon:
Nice, thanks for adding this.

data/core/editor/help.cfg
```diff
@@ -165,6 +165,44 @@ This turns any base terrain into a village, providing income and healing." + "
 Fake Shroud looks like an unexplored area, even in scenarios that have shroud disabled and even when the player’s units can see the hex."
 [/topic]
 
+[topic]
+    id=editor_terrain_elevation
+    title= _ "Terrain Elevation Graphics"
+    text= _ "editor^<img>src='icons/terrain/terrain_group_elevation_30.png~SCALE(60,60)' align=left box=yes</img> - A cosmetic elevation effect that looks a little different in the editor than in the game due to visual aids in the editor; all the overlays described here are found in the terrain palette’s “special” group." + "
```
stevecotton, line 171: I think it's better to put the images in a non-translatable string, for example as in the `editor_tool_village` section.

stevecotton, line 171: They're all in the "elevation" group, only some are in the "specials" group.

data/core/editor/help.cfg
```diff
@@ -165,6 +165,43 @@ This turns any base terrain into a village, providing income and healing." + "
...
+
+The bluffs, gulches, and similar variations are much like a standard terrain overlay, and can simply be used that way. However, for larger areas of the map, where you will want to allow the use of other overlay terrains, you will want to combine these with the “markers” described in the next section." + "
+
+" +
+        _ "editor^<bold>text='Elevation Floodfill'</bold>" + "
+
+<img>src='terrain/grass/green.png~BLIT(terrain/floodfill/flood_high.png)' align=here box=yes</img> " + _ "Raised elevation on-map marker" + " 
+<img>src='terrain/floodfill/marker-high-tile.png' align=here box=yes</img>" + _ "Raised elevation editor palette icon
+
+To make a patch of terrain higher or lower, and still be able to use other overlay terrains such as trees, you will need to use these markers. Placing one of these arrows on the map will flood-fill the map with a color-coded haze; neither the haze nor the arrow will be visible in game. The hazed area will have a border ledge, making it look higher or lower than the adjacent tiles.
```
stevecotton, line 186: I'm not understanding the "making it look higher or lower", given that these flood-fill to an area contained by a bluff or gulch. IIUC, they make it appear the same height as the surrounding bluff or gulch.

doofus-01: I don't know how to respond, honestly.  It says what I meant and I don't see how you could be confused.

Pentarctagon: I read it as "the hazed area" meaning "the flood filled area surrounded by the border ledge". The adjacent higher or lower tiles are the tiles immediately outside the bluff/gulch.

doofus-01:
@stevecotton , @Pentarctagon 
I feel this PR has been aged to perfection, should we serve it or toss it?

Pentarctagon:
I have no objections, at least.

data/core/editor/help.cfg
stevecotton: I don't have a suggestion for improving it - it makes sense after reading the "how to use" and then rereading this paragraph. So my issue with it is that it's relying on info from later, that the area will be bordered by the bluff/gulch tiles, but adding that info here makes it too long.

stevecotton:
Sorry for taking so long, it looks ready.

PR #7196, opened 2022-12-06
WML: Adjust the weapon selection logic for [kill]

CelticMinstrel:
The previous logic simply did not make sense, especially for the secondary weapon. For example, it would ignore the secondary weapon specification if the primary weapon was not found on the secondary unit, which does not make any sense.

The cases that this changes are:
- There is a secondary unit and a primary attack, but no matching attack was found. Previously it would use the default, now it pretends that attack exists.
- There is a secondary attack specified, but no primary attack, either because it was unspecified or because it didn't match any attack on the secondary unit

In both these cases, the animation runs as if a matching attack exists, rather than using no attack and choosing the default animation.

### DIFF
--- a/data/lua/wml/kill.lua
+++ b/data/lua/wml/kill.lua
@@ -54,10 +54,10 @@ function wesnoth.wml_actions.kill(cfg)
 				if secondary_unit then
 					found_attack = secondary_unit:find_attack(primary)
 				end
-				if not found_attack then
-					primary = wesnoth.units.create_weapon(primary)
-				else
+				if found_attack then
 					primary = found_attack
+				else
+					primary = wesnoth.units.create_weapon(primary)
 				end
 				wesnoth.log('err', "Primary weapon:\n" .. wml.tostring(primary.__cfg))
 			end
@@ -66,10 +66,10 @@ function wesnoth.wml_actions.kill(cfg)
 				if primary then
 					found_weapon = unit:find_attack(secondary)
 				end
-				if not found_weapon then
-					found_weapon = wesnoth.units.create_weapon(secondary)
-				else
+				if found_weapon then
 					secondary = found_weapon
+				else
+					found_weapon = wesnoth.units.create_weapon(secondary)
 				end
 				wesnoth.log('err', "Secondary weapon:\n" .. wml.tostring(secondary.__cfg))
 			end
--- a/data/lua/wml/kill.lua
+++ b/data/lua/wml/kill.lua
@@ -59,7 +59,7 @@ function wesnoth.wml_actions.kill(cfg)
 				else
 					primary = wesnoth.units.create_weapon(primary)
 				end
-				wesnoth.log('err', "Primary weapon:\n" .. wml.tostring(primary.__cfg))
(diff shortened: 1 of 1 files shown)


data/lua/wml/kill.lua
```diff
@@ -50,19 +50,25 @@ function wesnoth.wml_actions.kill(cfg)
...
 			-- In the victory animation, this is simply swapped.
 			if primary then
+				local found_attack = nil
 				if secondary_unit then
-					primary = secondary_unit:find_attack(primary)
-				else
+					found_attack = secondary_unit:find_attack(primary)
+				end
+				if not found_attack then
 					primary = wesnoth.units.create_weapon(primary)
```
stevecotton, line 58: `primary` gets assigned on this line, but that's unconditionally overwritten by the following `primary = found_attack`.

data/lua/wml/kill.lua
```diff
@@ -50,19 +50,25 @@ function wesnoth.wml_actions.kill(cfg)
 			-- In other words, the attacker's weapon. The attacker is the secondary unit.
 			-- In the victory animation, this is simply swapped.
```
stevecotton, line 51: This line about "the victory animation" appears to be a left-over from copying this from `[animate_unit]`'s implementation. Should it be removed?

stevecotton:
I'm not familiar with this part of the codebase, but on first glance: `[animate_unit]`'s implementation also ensures that the units face each other, which seems to be missing here. On the one hand, the plot may call for a backstab to be animated; on the other hand, a defend animation is going to look bad if played while the unit faces the wrong way.

data/lua/wml/kill.lua
CelticMinstrel: What makes you think it's a leftover? On line 76, the death animation is added, and on line 78, the victory animation is added with the animations swapped.

(Those are the line numbers as of 6ec5dcb8c4c9e904c8459e941f8a9509beba76db.)

stevecotton: Somehow I missed line 78.

data/lua/wml/kill.lua
stevecotton: While the code now works, I think the conditional blocks should be swapped so it's `if found_attack ... else` rather than `if not found_attack ... else`.

stevecotton (approved):

data/lua/wml/kill.lua
```diff
@@ -50,20 +50,28 @@ function wesnoth.wml_actions.kill(cfg)
...
+				local found_weapon = nil
 				if primary then
-					secondary = unit:find_attack(secondary)
+					found_weapon = unit:find_attack(secondary)
+				end
+				if found_weapon then
+					secondary = found_weapon
 				else
-					secondary = wesnoth.units.create_weapon(secondary)
+					found_weapon = wesnoth.units.create_weapon(secondary)
```
stevecotton, line 72: Should this be `secondary = ...` ?

stevecotton (changes requested):

stevecotton (approved):

CelticMinstrel:
So, just double-checking, but that approval means this could now be merged, right?

stevecotton:
Yup :D

PR #7197, opened 2022-12-06
Handle out-of-bounds side numbers in Lua's sync.evaluate_multiple

stevecotton:
Change the C++ to show a Lua error instead of crashing with an assert.

There's no issue logged for this, but this adds a unit-test for reproducing the crash.

The code formatting is consistent with the rest of the file; the entire file should be run through clang-format, but I don't want to do that on a change that should be backported.

I haven't included a changelog entry, I'm on the borderline of whether it should have one or not.

### DIFF
--- a/data/test/scenarios/test_synced_side_number.cfg
+++ b/data/test/scenarios/test_synced_side_number.cfg
@@ -1,0 +1,37 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: wesnoth.sync.evaluate_multiple
+##
+# Actions:
+# Request a synchronised choice from 3 sides in a game with only 2 sides.
+##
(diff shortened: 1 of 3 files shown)


soliton- (approved):
Looks pretty straight forward to me.  An assert for something that can normally happen is just wrong.

stevecotton:
Thanks for the review.

@Pentarctagon, for #7204 the newly-added `data/test/scenarios/test_synced_side_number.cfg` should move into `data/test/scenarios/lua_tests/wesnoth/sync/`. My feeling is to merge this PR as-is for easy backporting, and then leave it to you to rebase #7204.

Pentarctagon:
That's fine.

Issue #7188, opened 2022-12-03
Make units that have the same stats to have the same checksum as in 1.16


demario-w:
### Describe the desired feature

To replay saves from 1.16 smoothly, it is required that units have the same checksum if the stats and traits have not changed.

The units with a weapon special in 1.17 (horseman, thief, shaman, mage...) have different checksums than what they did in 1.16. Example for an intelligent, quick horseman:
```
(1.16) checksum="B0xNnBFOKh>Lw-f-d9Jcec.F<6D.>568p3fcgwbBc5kj.b25Y7y2Ohe5l+84Alnp.gm<39sh139,+n<pzlu8dYUjp0kF4mcn9V7fD>4I3bs7l3jv>0IuvQDBp8L+M62,"
(1.17) checksum="BllDfGwgr>>yfbf,f4Jk1naz.a.Xhlc8,3cebP4dp4>n2h><e0F-Wcg+8jlbZ+jk.oh<,mB6IZ++llbW,qhiZYwaIX<7iuDVF<.yfmF>5F3M3u59s-sujKDdp5+ijg1K"
```

If the horseman loses charge special, the checksums are the same in both version. This is most likely due to the deprecation of SPECIAL_NOTES. When loading the save, the game is throwing the warning:

_error deprecation: SPECIAL_NOTES has been deprecated and removed.; Use [special_note]note= tags instead to apply special notes, or use the {NOTE_*} macros (note the singular)._

Pentarctagon:
Sounds like a similar sort of issue as #5743

demario-w:
When loading the 1.16 save in both version with --log-debug=unit, the logging in src/units/unit.cpp[get_checksum] gives different results:
```diff
-- int.qui.horseman_checksum-1.16.out  2022-12-03 22:41:17.326940154 +0800
+++ int.qui.horseman_checksum-1.17.out  2022-12-03 22:42:32.711605434 +0800
@@ -28,7 +28,6 @@
            apply_to = both
            id = charge
            multiply = 2
-           name = charge
            special_note = Using a charging attack doubles both damage dealt and received; this does not affect defensive retaliation.
        [/damage]
    [/specials]
```
The absence of the 'name' attribute (inside attack/specials/damage) in 1.17 is what makes the checksum to fail. If the following change is made to the version 1.17:
```diff
diff -r 4133771678ff src/units/unit.cpp
--- a/src/units/unit.cpp	Fri Dec 02 10:04:35 2022 +0800
[...]

PR #7200, opened 2022-12-07
Checksums: accept 1.16's checksums during replays

stevecotton:
Looking at the discussion in #7188, I thought it wouldn't be hard to have a fallback which only triggers when the normal checksum fails, and sees if the unit passes a 1.16 style checksum instead. Took a bit longer than expected, but looks like a reasonable solution to that problem without adding a maintenance burden.

This will also accept them during multiplayer matches, but it will still warn if the unit's stats have changed. When using the new option, the checksum includes data that has been ignored since 93fe5607a658c74b987.

### DIFF
--- a/src/units/unit.hpp
+++ b/src/units/unit.hpp
@@ -2014,6 +2014,19 @@ private:
 	int moves_;
 };
 
+namespace backwards_compatibility
+{
+/**
+ * Optional parameter for get_checksum to use the algorithm of an older version of Wesnoth,
+ * thus preventing spurious OOS warnings while watching old replays.
+ */
+enum class unit_checksum_version {
+	current,
+	version_1_16_or_older /**< Included some of the flavortext from weapon specials. */
+};
+
+} // namespace backwards_compatibility
+
 /**
  * Gets a checksum for a unit.
  *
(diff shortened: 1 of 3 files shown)


src/units/unit.hpp
```diff
@@ -2014,6 +2014,17 @@ struct unit_movement_resetter
...
+/**
+ * Optional parameter for get_checksum to use the algorithm of an older version of Wesnoth,
+ * thus preventing spurious OOS warnings while watching old replays.
+ *
+ * TODO: this needs a better name, and it ought to be in a namespace.
+ */
+enum class checksum_compatibility {
+	current_version,
+	version_1_16	/**< Included some of the flavortext from weapon specials. */
+};
```
stevecotton, line 2026: As the comment says, this needs a better name and should be in a namespace. Any suggestions?

Pentarctagon: `checksum_version` maybe? Then the two values would be `current` and `1_16_or_older` (since all versions older than 1.17 didn't skip those attributes.

soliton-: Namespace could be just `compatibility` or `backwards_compatibility`.

soliton- (approved):

PR #7203, opened 2022-12-07
Exclude nonvillages from outer_villages array

### LINKED ISSUE #7202: UtBS S01: has 1841 `outer_villages`

- **Version:** 1.17.10+dev
- **Downloaded from:** Built from Git
- **Build info:** GCC and b5ff5c3ddc706f1281bbbe2f235b16c4b00f92db
- **OS:** Linux

### Description of the bug

Low priority, because it's only visible when looking at WML variables. However, as it's a mainline campaign it should be reference code that can be copied by UMC authors.

The code that uses `[store_locations]` to create the `outer_villages` array ends up with an array that has 1841 entries. That doesn't affect gameplay, because the array is only accessed in `capture` events, so the non-village entries are ignored.

1. Start Under the Burning Suns
2. :debug
3. :inspect
4. Start to scroll down to the "units" section, and notice that the scrolling is slow and there are over 1800 entries in the `outer villages_array`.

### Expected behavior

About 6 entries

### Additional context

_No response_

ProditorMagnus:
Fixes #7202 

@nemaara could confirm that Gd^Vdt, Rd^Vdt, Re^Vda are intended to be included.

V terrains in that map:
Dd^Vda
Dd^Vdr
Dd^Vdt
Ds^Vda
Ds^Vdr
Ds^Vdt
Gd^Vdt
Rd^Vdt
Re^Vda

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/01_The_Morning_After.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/01_The_Morning_After.cfg
@@ -926,7 +926,7 @@
         # There is a number of houses inside the oasis and even more tents
         # by the necromancer's fort, we need to make sure that these are
         # ignored. To do this we store coordinates ([store_locations]) of
-        # all desert villages (Dd^Vt) that are not ([not]) within
+        # all desert villages (*^Vda, *^Vdr, *^Vdt) that are not ([not]) within
         # rectangle defined by points 34,1 50,12 (x=34-50, y=1-12) or
         # ([or]) rectangle defined by 13,37 20,34 . First rectangle
         # includes all villages near necromancer, the second covers whole
@@ -934,7 +934,7 @@
         # outer_villages (variable=outer_villages).
         [store_locations]
             variable=outer_villages
-            terrain=*^Vdt,*^Vda,*^
+            terrain=*^Vdt,*^Vda,*^Vdr
             [not]
                 x,y=32-50,1-12
                 [or]


soliton-:
Wouldn't it be more robust to filter for gives_income?

ProditorMagnus:
Comment explicitly mentioned desert villages.

soliton-:
If the intention is truly to consider just desert villages then perhaps the variable should have a better name.

Looks like we're just waiting for the filter to become outdated again then though.

ProditorMagnus:
I have never opened the campaign, not even for this change. I have no idea what the map looks like or how it is played, I just analyzed map file and identified the overlay to add to filter so that functional behaviour would not change.

ProditorMagnus:
If it was not about desert villages then filter would have been \*^V*.

stevecotton:
> I have never opened the campaign, not even for this change. I have no idea what the map looks like or how it is played, I just analyzed map file and identified the overlay to add to filter so that functional behaviour would not change.

While I think this is the right change, the reason that I logged a bug for it instead of doing a quick fix was that I didn't want to spend time playtesting the fix. I feel that PR reviewers should assume that the author playtested it, unless the PR mentions otherwise.

knyghtmare:
> I have never opened the campaign, not even for this change. I have no idea what the map looks like or how it is played, I just analyzed map file and identified the overlay to add to filter so that functional behaviour would not change.

I think this PR needs Playtesting before merging. Because applying a patch without "knowing you are aiming to fix" is an invitation to more follow-up bugs

ProditorMagnus:
Correct. That is also why asked for approval from nemaara.

soliton-:
> If it was not about desert villages then filter would have been *^V*.

If it was about desert villages the variable name would have desert in it... I don't think this is a particularly useful line of thought. Clearly there is no difference anyway since there are no other villages.

My point is simply that gives_income is a more robust filter by not leaning on convention for terrain strings and instead using the property that actually makes a terrain a village.

ProditorMagnus:
True. I was just not aware of the key or even most changes after 1.14.

Would be good to solve the need of relying on terrain code for other terrain concepts too, not just village.

> When adding terrains make sure the following files are also checked:
> data/multiplayer/factions/* contains favorite positions for the different factions, this is only used for the random map generator at the moment so it is not very important.
> data/core/macros/abilities.cfg contains the definition of submerge and ambush so depending on the change these need to be updated.

nemaara (approved):
Good to go.

cooljeanius (approved):

cooljeanius:
> Good to go.

OK, so... can some merge please?

CelticMinstrel:
Looks like I showed up late, but… is there a reason not to use `[show_villages]`? Is it planned to be deprecated, maybe? It's basically the same as `[store_locations]` but with an implicit `gives_income=yes` in the filter.

PR #7204, opened 2022-12-07
Reorganize unit tests.

Pentarctagon:
---

Reorganizes the tests to directories following the layout of https://wiki.wesnoth.org/ReferenceWML and https://wiki.wesnoth.org/LuaAPI.

Draft for now since some cleanup is still needed for the actual tests themselves.


CelticMinstrel:
So the Lua tests are structured by module, and WML tests are structured by the wiki page name… that seems like a decent structure.

I wonder if the `data/scenario-*.cfg` files should also be moved into manual_tests? That includes the default test scenario and some other random scenarios that I'm not sure the purpose of.

Pentarctagon:
That makes sense to me.

stevecotton:
I'd consider AbilitiesWML to be part of UnitsWML rather than ScenarioWML.

We're missing AnimationWML, which would also be part of UnitsWML.

I'm unsure about the correct location for FilterWML, and I'm not even sure that it fits in this structure as it also covers animations and abilities. Seems as if it would need to split into several parts, some in UnitsWML and some in ScenarioWML.

MapGeneratorWML generates a scenario, but IIUC it runs in the Mapgen Lua Kernel; it seems wrong to place it in ScenarioWML which runs in the Game Lua Kernel.

test_movetype.cfg probably belongs in or under UnitsWML. While we were documenting stuff I thought it was testing `[effect]apply_to=movetype`, but seeing it in this reorganisation I now recognise that it's using that as a means to test the implementation of `[movetype]`.

Pentarctagon:
> We're missing AnimationWML, which would also be part of UnitsWML.

Would there ever actually be able to be any AnimationWML tests though?

Pentarctagon:
I've moved AbilitiesWML and test_movetype.cfg. I've removed MapGeneratorWML (can be re-added wherever is appropriate if there's ever any such tests added), and also removed FilterWML (its sole test moved to ConditionalWML since the filter was used inside a [have_unit]).

stevecotton:
The vision ones too. Vision and Jamming are obscure parts of the movetype data.

CelticMinstrel:
> I'd consider AbilitiesWML to be part of UnitsWML rather than ScenarioWML.

Agreed.

> I'm unsure about the correct location for FilterWML, and I'm not even sure that it fits in this structure as it also covers animations and abilities. Seems as if it would need to split into several parts, some in UnitsWML and some in ScenarioWML.

FilterWML doesn't really belong under anything, as it's used all over the place for all sorts of purposes. I'd make it a toplevel category.

> MapGeneratorWML generates a scenario, but IIUC it runs in the Mapgen Lua Kernel; it seems wrong to place it in ScenarioWML which runs in the Game Lua Kernel.

Uh,  what are you talking about? We're talking about WML here. It doesn't "run". It's data. Okay, it's true that ScenarioWML contains code (events) that run in the Game Lua Kernel, but it also contains other things that are not code. As for MapGeneratorWML, there are two completely separate map generators, one of which creates a Mapgen Lua Kernel and runs some code, the other of which just uses some data to define parameters of the map generator.

Also, if unsure about any others, it might make sense to compare to how the schema files are structured.

Pentarctagon:
> FilterWML doesn't really belong under anything, as it's used all over the place for all sorts of purposes. I'd make it a toplevel category.

Since it's used all over the place, I sort of feel like it should have its own tests in each place it's used rather than a separate category.

CelticMinstrel:
Hmm… that does sort of make sense, but… how filters work mostly doesn't depend on where they're used.

Pentarctagon:
To put it another way, I wouldn't really want to use a test checking a filter in [have_unit] to also ensure that filters work as expected in events and abilities. Even if how it works should be consistent, it would still be best to have separate tests for all those cases.

CelticMinstrel:
If it's testing that filters work as expected in a given situation, then that's part of testing that given situation, yes. However, if we're talking about a unit test suite to prove that each option in a filter has the expected effect, then that probably belongs in a separate FilterWML category.

Pentarctagon:
So there needs to be a separate FilterWML folder in each category where filters can be used?

CelticMinstrel:
> So there needs to be a separate FilterWML folder in each category where filters can be used?

That's an interesting idea… but I'm not really sure if that's needed. I guess it depends on whether there are multiple files that test filters in each of those cases?

I was thinking there should also be a toplevel FilterWML category which just tests filters generally, not linked to a specific use-case. The use of filters in a specific category would be testing that you can, in fact, filter correctly in that category and probably wouldn't test every possible thing you can do with a filter, whereas the general FilterWML category would test every possible thing you can do with a filter in a specific, simple context.

Pentarctagon:
What would be an example of toplevel FilterWML test cases? I'm fine with that too, I'm just struggling to come up with an example of a filter that isn't tied to some other functionality.

CelticMinstrel:
I think if I were writing toplevel FilterWML test cases they'd all have the same basic form:

* Create a unit that should match the filter and a unit that shouldn't match the filter.
* Find all units matching the filter and verify that the unit that should match is in the list and the unit that shouldn't match isn't in the list (using `wesnoth.units.find`).
* Verify that the unit that should match the filter matches the filter (using `wesnoth.units.match`).
* Verify that the unit that shouldn't match the filter doesn't match the filter (using `wesnoth.units.match`).

Of course, that's for unit filters, but the same structure should work for other types of filters too. You could argue that this is still tied to some other functionality (it uses the Lua API), but I think this has the lowest connection with other functionality.

Pentarctagon:
Alright, makes sense. Barring any further comments then, I plan to merge this tomorrow.

data/test/scenarios/manual_tests/overlay_text_demo.cfg
```diff
@@ -1,6 +1,6 @@
 #textdomain wesnoth-test
 
-[test]
+[multiplayer]
```
stevecotton, line 3: Why do some `[test]`s get changed to be `[multiplayer]`?

data/test/scenarios/wml_tests/ScenarioWML/EffectWML/special_note_individual_unit.cfg
```diff
@@ -1,7 +1,7 @@
 # wmllint: no translatables
 
 #####
-# API(s) being tested: [effect][special_note],[store_unit],unstore_unit]
+# API(s) being tested: [effect][special_note]
```
stevecotton, line 4: The original list was correct - this test was added as part of fixing #7153, a bug in `[store_unit]`.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/put_to_recall_and_modify.cfg
```diff
@@ -5,7 +5,7 @@
 # which needs to be handled within the WML API.
 
 #####
-# API(s) being tested: [put_to_recall_list], [modify_unit], [role][auto_recall]
+# API(s) being tested: [role][auto_recall]
```
stevecotton, line 8: `[modify_unit]` should be in this list - the bug that was fixed alongside adding this test is that the `[modify_unit]` put Charlie back on the map instead of the recall list.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/test_modify_unit.cfg
```diff
@@ -15,7 +15,7 @@
 # have the side-effect of moving it to the correct side's recall list.
 
 #####
-# API(s) being tested: [modify_unit],[role][auto_recall]
+# API(s) being tested: [role][auto_recall]
```
stevecotton, line 18: This test was added as part of fixing a bug in `[modify_unit]`, it should definitely be in this list.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/InternalActionsWML/test_role_types.cfg
```diff
@@ -1,5 +1,5 @@
 #####
-# API(s) being tested: [role],[recall]role=
+# API(s) being tested: [recall]role=
```
stevecotton, line 2: Original was right.

data/test/scenarios/wml_tests/UnitsWML/special_note_from_movetype.cfg
```diff
@@ -1,7 +1,7 @@
 # wmllint: no translatables
 
 #####
-# API(s) being tested: [movetype][special_note],[store_unit],[unstore_unit]
+# API(s) being tested: [movetype][special_note]
```
stevecotton, line 4: This was part of a bugfix of `[store_unit]`.

data/test/scenarios/wml_tests/UnitsWML/test_movetype.cfg
```diff
@@ -442,36 +442,36 @@
 )}
 
 #####
-# API(s) being tested: [store_locations][filter_vision]
+# API(s) being tested: [movetype]
```
stevecotton, line 445: For this one (the first of the `MOVETYPE_VISION_SCEN` ones), I'd suggest `[movetype],[store_locations][filter_vision]`. This one is partly here to sanity-check the framework that it's using, and it's definitely checking `[store_locations][filter_vision]`'s main function - that the API works with fog.

data/test/scenarios/manual_tests/overlay_text_demo.cfg
Pentarctagon: For consistency with all scenarios in manual_tests.

data/test/scenarios/wml_tests/ScenarioWML/EffectWML/special_note_individual_unit.cfg
Pentarctagon: Then this is a test of [store_unit]. My intention with this change as well as the others you commented on below was to be more clear on what is meant to be tested vs other functionality the test is relying on in order to work.

Here for example, if the bug occurs in [store_unit], then it's relying on [effect][special_note] to give a special note to the unit and [unstore_unit] to put the unit back on the map, but the specific bug it's testing isn't part of either of those APIs.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/put_to_recall_and_modify.cfg
Pentarctagon: I'd say this should just be `[modify_unit]` then.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/test_modify_unit.cfg
Pentarctagon: I'd say this should just be `[modify_unit]` then.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/InternalActionsWML/test_role_types.cfg
Pentarctagon: I'd describe this one as "requires [role] to work, but is testing that you can recall by the role".

data/test/scenarios/wml_tests/UnitsWML/special_note_from_movetype.cfg
Pentarctagon: Same comment as previously where this should be just `[store_unit]` instead.

data/test/scenarios/wml_tests/UnitsWML/test_movetype.cfg
Pentarctagon: I'd call this one (similar to my other replies) as relying on `[store_locations][filter_vision]`, but primarily testing `[movetype]`. Or since this is just meant as a baseline test without having changed the movetype at all, it could just be N/A.

data/test/scenarios/wml_tests/ScenarioWML/EffectWML/special_note_individual_unit.cfg
stevecotton: If these really were unit tests by the strict definition of that term then I'd agree, but using WML to test C++ tends to turn tests that touch more than one API. If `[effect][special_note]` wasn't adding a note then this test would notice, so I'm understanding the purpose of choosing only one of the APIs.

It tests the small part of `[store_unit]` that's influenced by `[effect][special_note]`. It doesn't cover the full features of either API.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/InternalActionsWML/test_role_types.cfg
stevecotton: Because of the multiple units, it seems more a test of `[role]`s filter to me.

data/test/scenarios/wml_tests/ScenarioWML/EffectWML/special_note_individual_unit.cfg
Pentarctagon: Is the bug related to the `[effect][special_note]` API specifically, or simply that a unit has the special note? The bug, as I understand it, was the latter.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/InternalActionsWML/test_role_types.cfg
Pentarctagon: That's what the second test in this file (test_role_2) is testing though, isn't it?

stevecotton:
test_role_1 is checking that it can assign a role to the right unit on the recall list. test_role_2 is checking that it can assign a role to the right unit on the map.

Pentarctagon:
Should the `[recall]` even be used in test_role_1 in that case? The test could instead check:
```
{ASSERT (
    [not]
        [have_unit]
            side=1
            role=advisor
            type=Spectre
            id=correct
            search_recall_list=no
        [/have_unit]
    [/not]
)}
{ASSERT (
    [have_unit]
        side=1
        role=advisor
        type=Spectre
        id=correct
        search_recall_list=yes
    [/have_unit]
)}
```

data/test/scenarios/manual_tests/overlay_text_demo.cfg
CelticMinstrel: That's not an answer. Why are they changed? The point of a `[test]` scenario is that you can launch it via the `-t` command-line option or the "test list" from the main menu. If you change it to a `[multiplayer]` scenario, then that won't work, right?

`[test]` does not mean it's a unit test, after all.

Unless you have some strong reasoning, I think it's incorrect to change these.

Pentarctagon: It makes them all available via the in-game UI when launching Wesnoth with the -d argument.

Pentarctagon: To be honest I'm not really sure why we currently have some hidden behind the -d flag and some hidden behind the -t flag. If nothing else, test scenarios that require manual playing to use should be handled consistently.

stevecotton: It makes them available via the "create a multiplayer game" UI, allowing the user to select an era and requiring the user to select side controllers and leader types in the GUI. This doesn't seem good.

stevecotton:
Who is the target audience for these "API(s) being tested" lines, and what is the target purpose?

Some possible examples, which would get us to different destinations:

* They're supposed to be written before writing the test itself, to focus the test author's mind while writing the test code.
* They're useful for `git grep "being tested.*role"` to find which of the tests using that API are testing the API itself.

Taking `test_role_1`, I'd agree that it could be two separate tests. However, perfect is the enemy of good. If you're enforcing that now on the "API(s) being tested" line then the PRs splitting such tests into two should be done first - and for the sake of reviewing what went where, splitting tests shouldn't happen inside a giant PR that's moving other files around.

data/test/scenarios/manual_tests/overlay_text_demo.cfg
stevecotton: It breaks the (useful when working on the command line) ability to launch those tests directly with `wesnoth -t testname`.

It's made `wesnoth -t` (without a testname) show an `Error: unknown scenario` instead of launching the dumping-ground test scenario. (That I'm calling it the "dumping ground" is something that would need a separate PR to fix.)

Pentarctagon:
It'd be grepping for what APIs currently have tests. However given the reorganization, it's not ideal to have a single test that's testing APIs from multiple categories.

data/test/scenarios/manual_tests/overlay_text_demo.cfg
Pentarctagon: Changed back.

data/test/scenarios/wml_tests/ScenarioWML/EffectWML/special_note_individual_unit.cfg
Pentarctagon: Changed to `[store_unit]`.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/put_to_recall_and_modify.cfg
Pentarctagon: Changed to `[modify_unit]`.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/test_modify_unit.cfg
Pentarctagon: Changed to `[modify_unit]`.

data/test/scenarios/wml_tests/UnitsWML/special_note_from_movetype.cfg
Pentarctagon: Changed to `[store_unit]`.

data/test/scenarios/wml_tests/UnitsWML/test_movetype.cfg
Pentarctagon: Changed to N/A.

stevecotton:
> It'd be grepping for what APIs currently have tests. However given the reorganization, it's not ideal to have a single test that's testing APIs from multiple categories.

I'm wondering if the reorganisation is too fined-grained, or if it should be postponed until the tests fit the design. Removing stuff from the "API(s) tested" seems to be making grepping harder.

Pentarctagon:
It makes the grepping more correct, since the primary purpose of the tests is for other things.

Pentarctagon:
If you want to know what tests happen to use a particular API, well you can grep for that easily enough both before and after the reorganization.

Pentarctagon:
@stevecotton The remaining comments are, in my mind, fairly minor things, so if you feel strongly about them I'll just go and change them back. This is the last unit test related thing I plan to do for a while, and I'd much prefer to get this merged and move on to other non-unit test related things than continuing to debate them.

data/test/scenarios/manual_tests/test_max_menu_items.cfg
```diff
@@ -1,5 +1,36 @@
 #textdomain wesnoth-test
-{GENERIC_UNIT_TEST "test_max_menu_items" (
+
+[test]
+    name = _ "Unit Test test_max_menu_items"
+    map_file=test/maps/generic_unit_test.map
+    turns = -1
+    id = test_max_menu_items
+
```
stevecotton, line 8: Trailing whitespace on a couple of lines in this file.

I think it could go back to using `GENERIC_UNIT_TEST`, if the change was just to make it use `[multiplayer]`.

data/test/scenarios/macro_tests/test_force_chance_to_hit_macro.cfg
```diff
@@ -99,20 +109,13 @@
...
-        {ASSERT ({VARIABLE_CONDITIONAL a.hitpoints equals 100})}
-        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 100})}
+
         {SUCCEED}
     [/event]
+
+    [event]
+        name=attacker hits,defender hits
+        first_time_only=no
+        {FAIL}
```
stevecotton, line 119: Nice redesign of the test.

This file also has DOS-style line endings, might be a good time to convert it.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/put_to_recall_and_modify.cfg
stevecotton: I've marked the other remaining "API(s) being tested" lines as resolved, but this one should be reverted. I looked at whether it's possible to split it into two separate tests to fit the "it's not ideal to have a single test that's testing APIs from multiple categories" comment, however the only difference between the two tests would be that one ends before the `[role]` assignment.

The bug that prompted this test to be written requires a unit to be on the recall list with `x,y` values which are valid to be on-map; which requires using `[put_to_recall_list]` instead of creating it directly with `[unit]x,y=recall,recall`. Because of this, if it were split into two tests without using Lua, then both of them have to start at the "Create a unit on the map, belonging to side 1" step.

stevecotton:
Reminder that #7197's newly-added `data/test/scenarios/test_synced_side_number.cfg` should move into `data/test/scenarios/lua_tests/wesnoth/sync/`.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/put_to_recall_and_modify.cfg
Pentarctagon: Updated.

data/test/scenarios/manual_tests/test_max_menu_items.cfg
Pentarctagon: Updated.

data/test/scenarios/macro_tests/test_force_chance_to_hit_macro.cfg
Pentarctagon: Updated.

Pentarctagon:
The test from https://github.com/wesnoth/wesnoth/pull/7197 has been moved.

stevecotton (approved):

data/_main.cfg
```diff
@@ -49,22 +49,24 @@
 #endif
 [/ais]
 
+#ifdef DEBUG_MODE
+# load tests that need manual user input to run and validate
+{test/scenarios/manual_tests}
```
stevecotton, line 54: I'd like to change the controlling `#ifdef` back to `#ifdef TEST` so that the `GENERIC_UNIT_TEST` and `COMMON_KEEP` macros can be used in manual tests. Also, these tests current need `-d` in addition to `-t testname` on the command line.

Would changing this back break something else?

Pentarctagon: It shouldn't.

CelticMinstrel: Can't you just delete this block? Isn't it included in the `{test}` inclusion further down under `#ifdef TEST`?

Pentarctagon: It isn't, since it was done as part of the "make all manual test scenarios be accessible in the same way" change, which then ended up getting partially reverted because I guess people like using `-t` rather than `-u` for certain tests.

CelticMinstrel: That's the entire purpose of `-t` though – to run a test scenario that requires interaction. The manual test scenarios under `test/` aren't even all of them; there are more such scenarios in the `ai/` directory. It keeps them out of the create game screen while also providing a nice convenient list if you bind a key to the "Run Test Scenario" action.

Pentarctagon: The point being that we have some manual tests that require using `-t`, some that require using `-u`, and then the elevation test scenario that's available by default along with the rest of the normal MP scenarios. It doesn't make any sense.

CelticMinstrel: Are there really manual tests that require using `-u`? To begin with, any test that runs with `-u` will also run with `-it`, but if there are any manual tests that include `is_unit_test=yes` then that should be removed, and then they'll also show up in the list in the UI.

From the sounds of it, the elevation test scenario should probably be using `[test]` instead of `[multiplayer]`. If there are any other test scenarios that show up like that, the same argument applies.

Pentarctagon: `test_max_menu_items` has `is_unit_test = yes`. I could've sworn there were more, but apparently not.

PR #7205, opened 2022-12-08
[MP] Change Elevation Test Scenario Name and Desc

knyghtmare:
Currently, the name of this scenario is too long, and it causes unattractive-looking width expansion of the MP scenario lists. Thus, it should be enclosed in `DEBUG_MODE` like the other three test scenarios in the MP scenarios folder.

### DIFF
--- a/data/multiplayer/scenarios/Test_Elevation_Test.cfg
+++ b/data/multiplayer/scenarios/Test_Elevation_Test.cfg
@@ -1,10 +1,9 @@
 #textdomain wesnoth-multiplayer
 
-#ifdef DEBUG_MODE
 [multiplayer]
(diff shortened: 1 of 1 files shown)


cooljeanius:
LGTM

Pentarctagon:
I'm wondering if it would make more sense to move it with the other test scenarios under `data/`?

@doofus-01 as well.

knyghtmare:
Adjust name and description to better reflect on what it is? Okay, can be done.

Pentarctagon:
It sounds like this can be merged then?

cooljeanius:
> It sounds like this can be merged then?

Yeah I think so (does this count as a review?)

cooljeanius (approved):

PR #7206, opened 2022-12-08
wmllint: convert `^Uf` to `^Tf`

knyghtmare:
## Summary

`^Uf` has been deprecated since 1.15.x
`^Tf` is the new and current one.

### DIFF
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -195,7 +195,7 @@ mapchanges = (
     ("Qv",  "Mv"),
 
     # mushroom
-    ("Uu^Uf",  "Tb^Tf"),
+    ("^Uf",  "^Tf"),
 
     # new to 1.17.8
     ("Ior", "Iwo"),
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -195,7 +195,7 @@ mapchanges = (
     ("Qv",  "Mv"),
 
     # mushroom
-    ("Qv",  "Mv"),
+    ("Uu^Uf",  "Tb^Tf"),
 
     # new to 1.17.8
     ("Ior", "Iwo"),
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -194,6 +194,9 @@ mapchanges = (
     ("^Vcm",  "^Vc"),
     ("Qv",  "Mv"),
 
+    # mushroom
+    ("Qv",  "Mv"),
+
     # new to 1.17.8
     ("Ior", "Iwo"),
     ("Icn", "Iwc"),


cooljeanius:
this would let us check off one of the boxes in issue #6403 btw

Elvish-Hunter (changes requested):
From a technical point of view, this is a simple change (just one line of code), however deciding what's the correct action isn't so simple.
First of all: are we sure that the `Uu^Uf` terrain needs to be phased out and always replaced with `Tb^Tf`? I remember that there were some balancing issues that made this change problematic, have these been solved?
Second: since it's an overlay, it's possible to place `^Uf` mushrooms *on any terrain*, not just on `Uu`; all these instances wouldn't be converted by wmllint, unless another conversion rule (`("^Uf", "^Tf")`) is added after the first one - which I suggest doing.

knyghtmare:
while porting UMC and I have a lot of them, I will say the main ones are:
1. `Uu^Uf` -> `Uu^Tf`
2. `Uh^Uf` -> `Uh^Tf`
3. `Uue^Uf` -> `Uue^Tf`
I think just adding a single generalised `^Uf` -> `^Tf` can be the optimal porting solution.

Sincerely, 
UMC author with 50 add-ons

stevecotton:
@Elvish-Hunter IIRC (and I just done a quick re-check, but not a proper refresh of my memory), changing `*^Uf` to either of `Tb` or `Tb^Tf` has exactly the same stats. So it can look bad, but doesn't change gameplay.

@knyghtmare those would look the same, but not necessarily play the same.

knyghtmare:
> @Elvish-Hunter IIRC (and I just done a quick re-check, but not a proper refresh of my memory), changing `*^Uf` to either of `Tb` or `Tb^Tf` has exactly the same stats. So it can look bad, but doesn't change gameplay.
> 
> @knyghtmare those would look the same, but not necessarily play the same.

I have interviewed several UMC authors and they have all stated that `Tb` is visually unappealing and wish the change to be from `^Uf` to `^Tf`. Thus, I adjusted the PR

stevecotton:
Why not just leave them as `^Uf`? It's a deprecated terrain code that we're trying to stop being added to new maps (by hiding it in the editor), and we're actively encouraging authors to change if they're editing the map (by the pink `D`). However, what problems does it cause that are serious enough to outweigh the ones that an automated change could bring?

stevecotton:
> ..., if the "D" is not shown while **playing the scenario**, it's never noticed.

So, why not just leave them as `^Uf`?

Your image is not convincing, because it looks like a map that either wouldn't be fun to play, or in which the accompanying .cfg file has some scripting that changes terrain.

Pentarctagon:
Would it make sense to only do this conversion if a specific flag is provided when wmllint is run?

CelticMinstrel:
> they have all stated that `Tb` is visually unappealing

Then maybe the correct solution is to make it less visually unappealing?

> If you need further convincing, here is the extreme case map (99% D)

Can't you just use the fill tool on that (and maybe manually fix up a few with a different base)? Unless the base terrains are totally patchworked (which seems not to be the case), it doesn't seem like something that requires wmllint to convert.

knyghtmare:
Fine.

Issue #3868, opened 2019-01-12
Document gettext/boost::locale mutex


jostephd:
src/gettext.cpp uses a mutex to protect various function calls:

https://github.com/wesnoth/wesnoth/blob/8cbc3a4477d34fbb414877d84d59e805055d832c/src/gettext.cpp#L47

We should document in source code comments why the mutex is needed and which functions should call it. Some info is here:
https://github.com/wesnoth/wesnoth/pull/3847#discussion_r247317865

cooljeanius:
`git blame` says that @gfgtdf added it in b780da4 in order to fix #2013

cooljeanius:
PR #7207 will close this

PR #7207, opened 2022-12-09
Address some minor nits in gettext.cpp

cooljeanius:
This PR is pretty simple; it fixes issue #3868 and addresses a review comment that @CelticMinstrel had left in b780da4.

### DIFF
--- a/src/gettext.cpp
+++ b/src/gettext.cpp
@@ -44,6 +44,7 @@
 namespace bl = boost::locale;
 namespace
 {
+	//This mutex avoids the crash reported in bug #2013:
 	std::mutex& get_mutex() { static std::mutex* m = new std::mutex(); return *m; }
 
 	class default_utf8_locale_name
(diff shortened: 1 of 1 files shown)


src/gettext.cpp
```diff
@@ -44,6 +44,7 @@
 namespace bl = boost::locale;
 namespace
 {
+	//This mutex avoids the crash reported in bug #2013:
```
Pentarctagon, line 47: Ideally this should describe the crash directly here as well. Otherwise if we move off github at some point in the future, there's no guarantee that this will be useful (similar to what happened with GNA - comments referenced GNA issues that were no longer accessible).

cooljeanius: is abcded4 better?

soliton-:
Are the locks still needed? We got rid of the loading screen threading.

soliton-:
"Avoids random crash" is not a particularly useful description anyway.

PR #7209, opened 2022-12-12
Update deprecation but postpone removal of the SPECIAL_NOTES macro

stevecotton:
Units still using this in their descriptions will have the player-visible header change from "Special Notes:" to "Special Notes (1.14-style, please update to the new list format)".

The SPECIAL_NOTES macro was originally removed early in the 1.17 dev cycle. That removal was reverted and postponed in the roadmap until Jan 2023, on the grounds that it's a lot easier to test 1.17 when the big add-ons from 1.16 can run on it.

In 1.16, UMC that hasn't upgraded yet already has a cosmetic bug - the help pages of units still using the {SPECIAL_NOTES} macro will include duplicate notes (assuming the expected usage of {SPECIAL_NOTES} as a heading in [unit_type]description=, which is followed by old-style notes). These are minor cosmetic bugs, which are expected to be removed as UMC gets updated.

That leaves the issue of what to do with the deprecated macro in 1.18. My feeling is that we can easily continue to support the macro, albeit with the cosmetic bug, so we should keep it for 1.18. However we could make it clearer that the duplicated notes should be removed from the UMC.

This also removes some docs about NOTE_*s, those macros have already been removed after being deprecated in 3568b5ff66ece00ec09f40059e552123f356d962.

Example help page of a unit still using this macro, but using mainline abilities that already include the special note. We can't simply remove all old-style special notes, because UMC abilities of affected units probably haven't been updated to include the notes in the ability.
image

### DIFF
--- a/data/core/macros/special-notes.cfg
+++ b/data/core/macros/special-notes.cfg
@@ -1,10 +1,9 @@
 #textdomain wesnoth-help
 
-# SPECIAL_NOTES_* are the deprecated versions that provide the strings.
-# INTERNAL:SPECIAL_NOTES_* contain the strings shared by deprecated and non-deprecated macros. Do not use them directly.
-# NOTE_* are the new versions that provide an entire tag.
-# For now, the new are defined in terms of the old, but when adding new special notes,
-# only a NOTE_* form should be added.
+# The SPECIAL_NOTES macro and ability-specific SPECIAL_NOTES_* macros are deprecated, they are strings that were historically included directly in [unit_type]description.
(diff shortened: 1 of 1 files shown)


stevecotton:
@ProditorMagnus either in January (if this change is rejected) or in 1.19 (if this change is accepted), it will need around 1300 changes to Ageless Era.

cooljeanius (approved):

stevecotton:
@CelticMinstrel do you want more time to review, or can I merge this?

CelticMinstrel:
I think I looked at it and had nothing to say, and I still have nothing to say, so I guess you can go ahead.

PR #7047, opened 2022-10-10
Wmllint: add new terrain code updaters

knyghtmare:
Allows updating terrain codes for stone floors and wooden floors by Wmllint tool. Accommodates #7042

Issue #7034, opened 2022-10-02
Map Diff causes CI failure if new map is added


doofus-01:
### Game and System Information

- **Version:**
1.17.7
- **Downloaded from:**
here

### Description of the bug

Example is in PR #7014 https://github.com/wesnoth/wesnoth/pull/7014/commits/0b19ba6beb96425b9e40118c54e9bfc8a36e2f8b

### Steps to reproduce the behavior

Open a PR that has a new map

knyghtmare:
@macabeus I suppose this is your field of expertise?

doofus-01:
#7042 also failed Map Diff CI, because there were new terrain codes

macabeus:
Hey, I'm following this issue.
But I'll probably have time and mind to fix both issues only in 2-3 weeks

doofus-01:
No worries, I just had to document the issues as they come up.  Thanks for keeping an eye on it.

soliton-:
Another related issue is when terrain strings get added/changed as in #7042.

cooljeanius:
> Another related issue is when terrain strings get added/changed as in #7042.

Is this why #6057 is failing too, then?

soliton-:
Xo is not getting added or changed there is it?

cooljeanius:
> Xo is not getting added or changed there is it?

Not in the PR itself, but maybe it's something that changed in base Wesnoth between when I first opened the PR and now?

cooljeanius:
Well 870bd71 ignores the CI errors, but I guess the underlying issue of having the map-diff tool handle additions of new maps might still be worth keeping this open for?

cooljeanius:
see also cooljeanius/wesnoth-map-diff#79 for the separate-repo version of wesnoth-map-diff

Issue #6862, opened 2022-07-12
map diff tool fails on certain terrain and images are swapped


soliton-:
The map diff tool fails on maps with `_off^_usr` terrain as visible in this run:
https://github.com/wesnoth/wesnoth/runs/7301694657?check_suite_focus=true

It would be good if there would be an error message if it doesn't find an image for a terrain.

Currently the parser fails to remove comments from attribute values and the terrain.cfg entry in this case looks like `string=_off^_usr    # wmllint: ignore`. Since there are plenty moe entries with those wmllint comments it fails on all those terrains as well.

Easy test cases are lots of the mainline multiplayer maps like: `data/multiplayer/maps/2p_Aethermaw.map`

For maps where it works the images are swapped. I would expect the old map on the left and the new map on the right but it displays the other way around.
Example with the top left tile changed to a mountain: https://github.com/soliton-/wesnoth/pull/5

soliton-:
@macabeus ^

cooljeanius:
Another example of terrain where it fails: `^Xo`, as seen in #6057. I tried taking a look at it, but don't really grok the source yet, as I don't speak Typescript: https://github.com/cooljeanius/wesnoth/tree/egallager_wesnoth_map_diff_tinkering

soliton-:
I think everything reported here is actually fixed.

For Xo there actually is no image. It's not a parsing issue. For that I would suggest to just use a fallback image as wesnoth does in debug mode IIRC.

PR #7211, opened 2022-12-14
SoF S9: Map `Isay` to `Iasy` code swap

### LINKED ISSUE #7210: SoF last scenario can't load

1.17.10

### Description of the bug

imagen

- Play Sceptre of Fire to the last scenario or load 9_Caverns_of_Flame.map on map editor.

### Expected behavior

_No response_

### Additional context

_No response_

knyghtmare:
Closes #7210


knyghtmare:
**Note:** I will _ignore_ the map-diff failed check.
The logic is that it registers a change from a non-existent terrain to an existing one (which patches the bug). Thus, it should be failing as expected.

Wedge009:
See my comment in #7210. I think @doofus-01 would prefer the code to be `Isay`.

knyghtmare:
we shall await for doofus's remarks then.

cooljeanius:
> **Note:** I will _ignore_ the map-diff failed check. The logic is that it registers a change from a non-existent terrain to an existing one (which patches the bug). Thus, it should be failing as expected.

Is that #6862, #7034, or its own separate issue? cc @macabeus

knyghtmare:
> > **Note:** I will _ignore_ the map-diff failed check. The logic is that it registers a change from a non-existent terrain to an existing one (which patches the bug). Thus, it should be failing as expected.
> 
> Is that #6862, #7034, or its own separate issue? cc @macabeus

I simply do not like this..."feature". It's been annoying in my case so I choose consider it to be irrelevant with logical reasoning.

stevecotton:
Running wmllint (since #7047) will revert this change. I haven't looked at which terrain code should be used, but I think this PR should be rejected until wmllint is changed (or it should be changed in this PR).

Wedge009:
Just got confirmation in #7210 that the definition should be changed to `Isay` to match.

PR #7214, opened 2022-12-17
HttT S03: Rewrite intro text, hint about training troops for S06

stevecotton:
The Isle of Alduin doesn't have much dialogue at the start, it's a map where the island gives the player two fronts to progress on, and it's early enough to give the player time to prepare for the SoE.

### DIFF
--- a/data/campaigns/Heir_To_The_Throne/scenarios/03_The_Isle_of_Alduin.cfg
+++ b/data/campaigns/Heir_To_The_Throne/scenarios/03_The_Isle_of_Alduin.cfg
@@ -39,6 +39,11 @@
                 bonus=yes
                 carryover_percentage=40
             [/gold_carryover]
+
+            [note]
+                # po: a hint that the player has 3 scenarios before the first main test of their XP management
+                description= _ "You’ll need experienced troops at Elensefar."
+            [/note]
         [/objectives]
     [/event]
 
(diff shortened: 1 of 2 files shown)


knyghtmare (changes requested):
image

The message/note on the objectives...seems a bit bad as it expands the window width extensively to double it's initial width and the result is un-appealing.

I suggest that the message be shown once, in either a narrator `[message]` or `[object]` `description=`

nemaara (changes requested):
I prefer knyght's solution as well, plus it should mostly be a hint for easy/normal.

Edit: I mean the solution of Delfador saying the tactic thingy.

stevecotton:
When trying to add that hint as a single-line change to the dialogue, I reviewed the existing text and felt it didn't flow well. Reading [the feedback thread](https://r.wesnoth.org/t2959), Delfador should have more emotion about finding his childhood home overrun; however I've gone for practical thoughts first. All the current feedback is looking at the same intro - these lines date back to the third commit in the Git history, when Alduin was `data/scenarios/scenario02.cfg`.

The hint isn't `#ifdef` to only appear on easy, because the text wouldn't flow without it - take that line out and it goes from "there are more orcs than I thought" to "we must fight these orcs, who aren't the ones at the SoE". Also, players who are already doing XP management can feel knowledgeable when they see it.

I'm still hoping to backport, but worrying that it might be both too big a change, and an incomplete change (expecting that someone else will give this text a further cleanup). If anyone wants to take this and make their own PR with better prose, please feel free - it's become a bigger change than the time I allocated for working on it.

stevecotton:
Forgot to add a changelog entry, will push again.

knyghtmare (approved):

data/campaigns/Heir_To_The_Throne/scenarios/03_The_Isle_of_Alduin.cfg
```diff
@@ -193,17 +198,22 @@
...
+        [/message]
         [message]
             speaker=Delfador
-            message= _ "I fear so, Konrad. It seems that the orcs have come even here. Here to the place where I was born, where I was trained."
+            message= _ "If the orcs have come here, their forces at Elensefar must be even more numerous than I feared."
         [/message]
         [message]
-            speaker="Usadar Q'kai"
-            message= _ "Who is that? Oh, a party of elves has landed. We shall drive them back into the sea!"
+            speaker=Delfador
```
cooljeanius, line 210: So, Delfador has multiple "mood" portraits available; I wonder if it'd make sense to have him use his "mentoring" one here?

cooljeanius (changes requested):
A general review of "mood" portraits might be good to have here, I think...

data/campaigns/Heir_To_The_Throne/scenarios/03_The_Isle_of_Alduin.cfg
stevecotton: His default `[unit]profile=` picture has a `~RIGHT()` modifier on it, the one in `DELFADOR_MENTORING_ELF` doesn't. IMO the switch of side breaks the flow of dialogue, so my feeling is against using it.

cooljeanius: oh ok nvm then

cooljeanius (approved):

data/campaigns/Heir_To_The_Throne/scenarios/03_The_Isle_of_Alduin.cfg
Wedge009: It's still there. Assuming you meant to do so, I've marked it as resolved.

nemaara (approved):

PR #7216, opened 2022-12-19
1.17 - dunefolk rover portrait

doofus-01:
This was meant for the rover, extending it to the explorer might be a stretch.


doofus-01:
This would be better with review by someone like Lordbob, but I don't think that will happen anytime soon.
So I'll merge this, and if something needs to be fixed, you know where to find me.

PR #7220, opened 2022-12-21
[Campaign] [WoF] Enable Saurian Spearthrower

knyghtmare:
Since the unit has been (finally) added to core, I believe it should be made be available for players in _Winds of Fate_.

### DIFF
--- a/data/campaigns/Winds_of_Fate/_main.cfg
+++ b/data/campaigns/Winds_of_Fate/_main.cfg
@@ -27,6 +27,8 @@
     {CAMPAIGN_DIFFICULTY NIGHTMARE "units/drakes/armageddon-melee-6.png~CROP(0,19,62,64)~RC(magenta>red)" ( _ "Ancestor")  ( _ "Nightmare")}
 
     {ENABLE_ADVANCEMENT "Cuttle Fish" "Kraken" (set_experience=80)}
+    # enable saurian spearthrower
+    {ENABLE_SAURIAN_SPEARTHROWER}
 
     {./about.cfg}
 [/campaign]
--- a/data/campaigns/Winds_of_Fate/scenarios/10_Fire_Meets_Steel.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/10_Fire_Meets_Steel.cfg
@@ -34,7 +34,7 @@
     [side]
         side=1
         controller=human
-        recruit=Drake Burner, Drake Clasher, Drake Fighter, Drake Glider, Saurian Augur, Saurian Skirmisher, Fire Drake, Drake Flare, Drake Arbiter, Drake Thrasher, Drake Warrior, Sky Drake, Saurian Ambusher, Saurian Oracle, Saurian Soothsayer
+        recruit=Drake Burner, Drake Clasher, Drake Fighter, Drake Glider, Saurian Augur, Saurian Skirmisher, Fire Drake, Drake Flare, Drake Arbiter, Drake Thrasher, Drake Warrior, Sky Drake, Saurian Ambusher, Saurian Spearthrower, Saurian Oracle, Saurian Soothsayer
         {GOLD4 160 120 80 40}
         village_gold=4 # More income since the eyrie has been established.
         save_id=Player
--- a/data/campaigns/Winds_of_Fate/scenarios/11_Crosswind.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/11_Crosswind.cfg
@@ -33,7 +33,7 @@
     [side]
         side=1
         controller=human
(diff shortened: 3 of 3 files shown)


knyghtmare:
@Jonathan-Kelly Any comments/remarks?

Jonathan-Kelly:
Well, when I tested this unit during WoF's most recent balance update I found its utility to be a bit too much between the ambusher and clasher line. Also its lack of a shield makes it look like a more (fr)agile alternative to the ambusher, yet its statistics lean the opposite way.

So my plan was to give it the `skirmisher` ability, reduce its hit points and tweak its damage output to position it as a ranged attack focused alternative to the ambusher. Also rename it to the less awkward sounding "Saurian Peltast" perhaps. Then test and tweak it with my next balance update for WoF (which should partly focus on making saurians more useful, via map changes and the level 3 augurs, as well as this peltast idea).

CelticMinstrel:
> less awkward sounding "Saurian Peltast"

That's clearly a matter of opinion. To me, that's _more_ awkward sounding.

knyghtmare:
> So my plan was to give it the skirmisher ability, reduce its hit points and tweak its damage output to position it as a ranged attack focused alternative to the ambusher. Also rename it to the less awkward sounding "Saurian Peltast" perhaps. Then test and tweak it with my next balance update for WoF (which should partly focus on making saurians more useful, via map changes and the level 3 augurs, as well as this peltast idea).

I think this was already discussed and it was agreed that this unit-line (spearthrower) would not be some mirror of Ambusher. So, no `skirmisher`. It's a ranged focused archer already, and the level 3 is a formidable opponent indeed. So, any changes to it...will be denied by me.

Jonathan-Kelly:
> I think this was already discussed and it was agreed that this unit-line (spearthrower) would not be some mirror of Ambusher. So, no skirmisher. It's a ranged focused archer already, and the level 3 is a formidable opponent indeed.

There was much discussion but little agreement regarding this unit line (which was split into two unit lines at your request). I do not think anyone questioned if its level 3 can be formidable as an opponent, and no such opponent appears in this campaign anyway. To reiterate, the trouble with this unit is the lack of utility it adds to the faction, plus the mismatch of its statistics and appearance. (Leaving aside whether an archer should come from a melee skirmisher scout).

> So, any changes to it...will be denied by me.

If there is zero flexibility on possible improvements then I definitely have to vote 'no' on this PR. It would be unfair to tempt first time players to pour their gold and experience into a unit line that will have mediocre utility for them in later scenarios.

doofus-01:
> So my plan was to give it the skirmisher ability, reduce its hit points and tweak its damage output to position it as a ranged attack focused alternative to the ambusher. Also rename it to the less awkward sounding "Saurian Peltast" perhaps. Then test and tweak it with my next balance update for WoF (which should partly focus on making saurians more useful, via map changes and the level 3 augurs, as well as this peltast idea).

I'd say that the unit shouldn't be modified for this campaign, that would be confusing.  (If that was even the suggestion, maybe I'm misreading.)

> If there is zero flexibility on possible improvements then I definitely have to vote 'no' on this PR.

This reaction is understandable.  Getting the new units in front of players would be nice, but maybe a quick little 2p scenario is better.  The new units might get adjusted, no need to complicate this campaign with that.

Jonathan-Kelly:
I agree that it would not make sense to modify a core unit just to fit this one campaign specifically. Instead, the suggestion was to modify it to fit the drakes faction generally.

Drakes have strong ranged attacks of both fire and cold types - there is no unit in the game resistant to both these 'opposite' types of damage while being vulnerable to pierce. Drakes also have plenty of piercing melee weapons. So for this faction, a _"ranged focused archer"_ is a fifth wheel.

If it kept the `skirmisher` ability (inherited from the unit it levels up from) that might give it a useful niche. Which would also much better fit its lightweight visual appearance and help distinguish it more from the (non-skirmisher) saurian Guard.

> This reaction is understandable. Getting the new units in front of players would be nice, but maybe a quick little 2p scenario is better. The new units might get adjusted, no need to complicate this campaign with that.

I am actually quite enthusiastic about adding the saurian Prophet and Seer to the campaign as soon as they are added to core, along with your newest monsters like the Caribe and Dragonfly. And it is no big deal for WoF if any of these get adjusted afterwards.

The only problem is with adding a unit which is both poorly thought out in its present form _and_ arbitrarily blocked from improvement.

doofus-01:
> The only problem is with adding a unit which is both poorly thought out in its present form and arbitrarily blocked from improvement.

In that case, I'd also vote for closing this PR.  The major points here were already discussed in #6912, the "skirmisher" conclusion as I understood it was to just see how things shake out, especially once the other saurians were merged.  We're not there yet, and this particular flavor of saurian doesn't seem like the best fit anyway, as already noted.

PR #7221, opened 2022-12-21
[LoW] S3: fix mismatched coords of Urudin retreat

knyghtmare:
Urudin the Slayer is supposed to move to 12,4 (next to the castle of one of the warlords), instead he is set to move to 20,6 which makes little sense. The event of Urudin moving at 12,4 was supposedly never triggered.

Reported on Discord

> Hi
In the file 03_Kalian_under_Attack.cfg, the coordinates of line 200 and 454/455 are not the same. Is that just a typo or intentional?
Steam 1.16.6

### DIFF
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter1/03_Kalian_under_Attack.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter1/03_Kalian_under_Attack.cfg
@@ -188,6 +188,7 @@
         [ai]
             [engine]
                 name=lua
+                # Note that the move_full co-ordinates should match the moveto event for Urudin below
                 code=<<
                 local my_ai = { }
 
@@ -453,6 +454,7 @@
         [filter]
             id=Urudin
             [filter_location]
+                # Note that these co-ordinates should match the AI code for Urudin's side declared above
                 x=12
                 y=4
                 radius=3
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter1/03_Kalian_under_Attack.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter1/03_Kalian_under_Attack.cfg
@@ -197,7 +197,7 @@
                         local mhp, hp = urudin.max_hitpoints, urudin.hitpoints
                         local turn = wesnoth.current.turn
                         if turn >= 3 or hp < mhp / 2 then
-                            ai.move_full(urudin, 20, 6)
+                            ai.move_full(urudin, 12, 4)
                         end
                     end
                 end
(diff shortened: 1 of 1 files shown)


stevecotton:
I confirm that it doesn't trigger in 1.16.6+dev (1b24365dae125).

For the `[moveto]`'s trigger coordinates, I suggest making them a range in case the target hex is occupied (unlikely, but might as well).

Changing the AI coordinates from 20,6 to 12,4 probably means 2 extra turns of travel before reaching that point.

On turn 7, the map grows moving all hexes right and down by 2 hexes in each direction. 12,4 makes sense before the move, and looks reasonable after the move too.

Wedge009:
I was going to ask where (12,4) comes from but that is indeed well-spotted in the code, it makes sense to me that they should match. I don't recall seeing the dialogue at the event triggered by (12,4) before. Is there some way of mapping the co-ordinates to a variable somehow so that it doesn't have to be repeated redundantly? Failing that, perhaps add a comment explaining why these co-ordinates should be so.

This change should be back-ported too, I think.

data/campaigns/Legend_of_Wesmere/scenarios/chapter1/03_Kalian_under_Attack.cfg
```diff
@@ -451,9 +451,12 @@
...
         name=moveto
         [filter]
-            x=12
-            y=4
             id=Urudin
+            [filter_location]
+                x=12
+                y=4
+                radius=3
+            [filter_location]
```
stevecotton, line 459: Missing a `/` here, but that means it needs testing.

doofus-01:
> Is there some way of mapping the co-ordinates to a variable somehow so that it doesn't have to be repeated redundantly? Failing that, perhaps add a comment explaining why these co-ordinates should be so.

There's `[store_locations]` or just putting the coordinates in a macro.  There also might be some way with `location_id` that is mentioned in location filtering, but I'm not familiar with how one defines that.

CelticMinstrel:
I would recommend the special location approach. You can place a special location on the map in the editor and give it a name. Then it can be filtered on with `location_id`, and you can get its coordinates in Lua with `wesnoth.current.map.special_locations.name`.

nemaara (approved):

Wedge009:
At the very least, a comment against one or both of the matching co-ordinates would be good (to guard against the co-ordinates becoming mis-matched again later), but otherwise I vote to merge and back-port.

CelticMinstrel:
> I hope we move to that, but feel it's something to do when someone makes a major revision of a map or adds a new campaign, rather than adding it in small bugfixes like this one.

Much as I recommend it, I don't think that's a reason to refuse this PR unless someone comes along and actually does it before this is merged.

PR #7226, opened 2022-12-23
Ignore duplicate primary key errors for game_content_info table.

Pentarctagon:
The cause is that somehow, during regular play, MP campaigns can end up with duplicate values in the [content] data for each [addon]. A very similar issue was fixed in 6c980d12dcd56ccd16ff8b7150cd15ef19f0a821, but there's apparently some other way to trigger it that I haven't been able to reproduce. Since the error itself is harmless, it'll just be ignored server-side to avoid the spurious errors in the server log file.

### DIFF
--- a/src/server/wesnothd/server.cpp
+++ b/src/server/wesnothd/server.cpp
@@ -1676,11 +1676,16 @@ void server::handle_player_in_game(player_iterator p, simple_wml::document& data
 			const simple_wml::node& m = *g.level().root().child("multiplayer");
 			DBG_SERVER << simple_wml::node_to_string(m);
 			// [addon] info handling
+			std::set<std::string> primary_keys;
 			for(const auto& addon : m.children("addon")) {
 				for(const auto& content : addon->children("content")) {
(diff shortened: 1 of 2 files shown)


soliton-:
I'm not a big fan of ignoring the error in the low level DB function instead of the specific caller (I know there is just a single caller right now). Also this ignores the error even if it happens because an entry already exists in the DB from before not just when the config has duplicate entries.

soliton-:
A somewhat related issue is that it seems to me like `SOURCE` should be part of the primary key of `game_content_info` as well. SOURCE is the addon id which is also the dir name on the filesystem and enforced to be unique on the addon server. While the content ID of a modification for example could easily not be unique over all addons by accident.

Pentarctagon:
> I'm not a big fan of ignoring the error in the low level DB function instead of the specific caller (I know there is just a single caller right now). Also this ignores the error even if it happens because an entry already exists in the DB from before not just when the config has duplicate entries.

So it should instead be checked for and skipped in server.cpp instead?

soliton-:
> So it should be checked for and skipped in server.cpp instead?

Yes and preferably by checking what is inserted not blindly ignoring the DB error. If no duplicates are inserted in one go even a duplicate key error is valid.

Pentarctagon:
Alright, updated to do that. I'll update the primary key on the actual database after merging.

soliton- (approved):

soliton-:
Please also backport to 1.16.

Issue #6011, opened 2021-08-07
Need better Help dialog for unit variations


doofus-01:
### Game and System Information
Battle for Wesnoth v1.15.15+dev (f8a856bad7a-Modified) x86_64
1920x1015
### Describe the bug
Unit variations display in Help dialog look bad
 - Unit name (walking corpse, for example) is listed multiple times in tree view
 - Tree view line spacing jumps when a variation node is opened
 - There doesn't seem to be a convention on how to use `name` vs `variation_name`.  Comparing walking corpse to sand scuttler, probably either convention would work, but the GUI would have to deal with them differently.
 - Possibly the list of variations could be removed from the individual variation pages, since the base unit is linked to and the list is provided in the tree view.
 - Units with variations are bumped to the top of the list with a race in the tree view

### To Reproduce
Look at the unit descriptions in Help.

### Screenshots
![soulless-variation-help-pre](https://user-images.githubusercontent.com/13510196/128616062-f783803a-37e7-47d1-844d-2c165737fb0f.jpg)
![WC-variation-help-post](https://user-images.githubusercontent.com/13510196/128616064-12461102-2892-4564-a570-904d3a5c1573.jpg)

stevecotton:
> Unit name (walking corpse, for example) is listed multiple times in tree view

FWIW, the code controlling that is here, but this patch also removes it from the title of the right-hand pane:
```
diff --git a/src/help/help_impl.cpp b/src/help/help_impl.cpp
index 8f13d32db7b..144002a53de 100644
--- a/src/help/help_impl.cpp
+++ b/src/help/help_impl.cpp
@@ -956,7 +956,7 @@ void generate_unit_sections(const config* /*help_cfg*/, section& sec, int /*leve
                for (const std::string &variation_id : type.variations()) {
                        // TODO: Do we apply encountered stuff to variations?
                        const unit_type &var_type = type.get_variation(variation_id);
-                       const std::string topic_name = var_type.type_name() + "\n" + var_type.variation_name();
+                       const std::string topic_name = var_type.variation_name();
                        const std::string var_ref = hidden_symbol(var_type.hide_help()) + variation_prefix + var_type.id() + "_" + variation_id;
 
                        topic var_topic(topic_name, var_ref, "");
```

stevecotton:
Related issue:

* On the help page for abilities, the variation names are not shown. So when playing SotA, the page for the Plague ability is a list that says just "Walking Corpse" once for each variant.

[...]

PR #7236, opened 2022-12-28
1.17 move Sand Scuttler out of scorpion

doofus-01:
The potential unit race help menu tree rearrangements in #7228 reminded me how terrible the "variations" look there (#6011), and it's really only the zombies and the sand scuttler that use the feature, at least for core.  It's not super clear how to deal with the zombies, but the sand scuttler really stands out and can be more easily fixed/removed.

I used `[base_unit]` in case there was some reason to have the Giant Scorpion and Sand Scuttler remain linked.


CelticMinstrel:
Hmm, this could be a problem. The sand scuttler exists in 1.16 too, doesn't it?

doofus-01:
Yes, unfortunately, so it will be a pain for UMC migration.  Maybe this can be addressed with wmllint?  But I wonder how much "scuttler" is actually used.

CelticMinstrel:
Maybe it can be addressed in wmllint, but I think we could just leave the scuttler variation in place for 1.18 and deprecate it to be removed in 1.19. There is a DEPRECATED_UNIT macro which is meant for deprecating units, but I think it should also work to deprecate a variation if you placed it in the `[variation]` tag. I haven't tested this on a variation, however…

doofus-01:
I disagree with leaving it.  This isn't some "under-the-hood" API change, this is pretty visible and it is bad enough to need fixing, not fading out six years from now.  The scuttler was a band-aid in the first place, and there is a simple fix.  Well, I'm not sure how simple for wmllint, but generating a sensible error message should be possible.

knyghtmare (approved):

CelticMinstrel:
> This isn't some "under-the-hood" API change

It is a change that will hard break add-ons, which is not good no matter how important the fix is. I don't recall exactly what happens if you try to place a unit with a non-existent variation, but I can't imagine it would be good. It's probably not as bad as a missing unit, I guess. Probably the best thing that could happen is that it reverts to the default variation (and possibly alters the balance of the scenario).

CelticMinstrel:
> It's not super clear how to deal with the zombies

They have a strong reason to stay as variations, because they're the output of a plague special. As long as plague works the way it does, we can't go and change how the zombie unit is structured.

knyghtmare:
> Unless you can prove that "scuttler" is not used anywhere in 1.16 right now, I don't think it's a good idea to skip the deprecation process. Even then it's not really a good idea, but could be justified.

The three use cases of scuttler I have seen is probably Mercenaries, EI-rework and HoD. The first 2 are in 1.17.x and will be able to adjust, I suppose. HoD is still very early pre-alpha so I can easily make the change with little to no conflict.

CelticMinstrel:
Uses in 1.17.x don't matter. If the variation is completely unused (or doesn't exist) in 1.16, then there is basically no problem.

I don't know for certain whether it exists in 1.16, does someone want to confirm that first?

doofus-01:
If we truly couldn't touch this thing because some UMC used it in 1.16, that would mean our deprecation process needed revision.  Fortunately, we don't need to be that sclerotic, the variation can be hidden.

data/core/units/monsters/Giant_Scorpion.cfg
```diff
@@ -106,4 +106,101 @@
         [/frame]
         {SOUND:HIT_AND_MISS spear.ogg {SOUND_LIST:MISS} -100}
     [/attack_anim]
+    [variation]
+        variation_id=scuttler
+        variation_name= _ "scuttler"
+        name= _ "Sand Scuttler"
+        inherit=yes
+        hide_help=yes
+        inherit=yes
```
knyghtmare, line 115: duplicated `inherits=yes`

knyghtmare (changes requested):

data/core/units/monsters/Giant_Scorpion.cfg
doofus-01: removed

knyghtmare (approved):

PR #7237, opened 2022-12-28
Add basic achievements functionality.

Pentarctagon:
This reads the mainline achievements.cfg and then all the achievements of each installed add-on.

This is intentionally handled separately from other WML loading so that: a) All achievements and their status are able to be displayed on the main menu right after Wesnoth starts and regardless of which add-ons are active. b) Add-ons can add additional achievements to other content, whether UMC or mainline. For example, a modification that adds more achievements for mainline campaigns.

Marking something as achieved is handled by the new [set_achieved] tag and whether an achievement has been completed can be checked via [has_achievement].

There is no attempt to prevent people from manually editing which achievements they've accomplished.

NOTE: These are *not* in any way related to Steam achievements!

---

In-game popup example:
completed-ingame

Completed in the dialog:
completed-inlist

Not completed:
not-completed

Hidden:
hidden


data/gui/window/mp_report.cfg
```diff
@@ -1,6 +1,6 @@
 #textdomain wesnoth-lib
 ###
-### Definition of the window to provide authentication information when uploading an add-on
+### Definition of the window for reporting a player on the online multiplayer server
```
stevecotton, line 3: Seems unrelated to the current PR

knyghtmare:
I think this would be quite desired by both mainline and UMC (mostly UMC).
Anyways, one request: 
Allow `[achievement]` to have a field for icon. Just textual entries feels too lacking
```ini
[achievement]
    content_for={CONTENT_FOR}
    id={ID}
    icon={ICON_PATH}
    name={NAME}
    description={DESC}
    hidden={HIDDEN}
[/achievement]
```

src/achievements.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "game_config.hpp"
+#include "log.hpp"
+#include "preferences/general.hpp"
+#include "serialization/parser.hpp"
+#include "serialization/preprocessor.hpp"
+
+static lg::log_domain log_config("config");
+#define ERR_CONFIG LOG_STREAM(err, log_config)
+
+// TODO: horizontal scrollbar is present despite `horizontal_scrollbar_mode = "never"`
```
CelticMinstrel, line 28: The horizontal scrollbar you're seeing (probably) isn't something defined in the dialog file. There's a scrollbar in the window's widget definition which ensures that a dialog that's too wide can still be fully seen. If the scrollbar is scrolling the listbox's position and not just its content, then it's the scrollbar from the window definition.

src/gui/dialogs/achievements_dialog.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+void achievements_dialog::pre_show(window& win)
+{
+	std::vector<config> content_list;
+	content_names_ = &find_widget<menu_button>(&win, "selected_achievements_list", false);
+	connect_signal_notify_modified(*content_names_, std::bind(&achievements_dialog::set_achievements_content, this));
+
+	achievements_box_ = find_widget<listbox>(&win, "achievements_list", false, true);
+
+	for(const auto& [content_for, list] : achieve_.get_list())
```
stevecotton, line 47: As it's a new file, run it through clang-format unless it's so similar to another file that we should be doing a side-by-side comparison.

src/achievements.cpp
```diff
@@ -0,0 +1,116 @@
...
+// TODO: how to make the initial height more than a single achievement? hard code the height? add more achievements?
```
CelticMinstrel, line 29: I think there's a macro to make a widget take up a minimum size? It essentially stacks it on top of a spacer.

src/achievements.cpp
```diff
@@ -0,0 +1,116 @@
...
+// TODO: make checkbox to indicate which achievements have been achieved disabled/untoggleable
+// TODO: GUIWindowDefinitionWML in achievements_dialog.hpp
+// TODO: some way to display in-game when something is achieved?
```
CelticMinstrel, line 32: I want to say a modeless dialog, but as far as I know those still don't work over the map…

src/achievements.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+// TODO: handling translatable strings? needs tstring?
+// TODO: can content_for be translatable? probably not since it's used internally for things - need two separate attributes, one translatable and one not translatable?
```
CelticMinstrel, line 35: Hmm. It's kinda unclear what it actually refers to. Is it "just" a "unique identifier"? Then it wouldn't be translatable, but you'd still need to somehow map it to a translatable value.

It would be ideal if `content_for` could be an "addon ID", which is basically a directory name, and the game could get the translatable version by looking for that folder and… somehow obtaining it.

Since every `achievements.cfg` is loaded anyway, perhaps this could be achieved by just adding a key in `achievements.cfg` so it looks like this:

```
display_name=_"Tutorial"

[achievement]
    content_for=tutorial
    id=finished
    name=_"Complete the Tutorial"
    description=_"Complete both scenarios of the Tutorial campaign"
[/achievement]
```

The only missing piece here is how to handle elements of mainline. I think namespacing with `/` would do the trick for referring to them, so the above would become something like `content_for=mainline/campaigns/tutorial`, while a `content_for` without any slashes would look in the addons directory.

As for defining the display name for mainline campaigns, that's less clear; your model has a single `achievements.cfg` for all of mainline, but the above approach implies one display name per `achievements.cfg`. Perhaps an `[achievement_group]` tag that contains the `display_name`? An addon could omit it (and use the example above verbatim), but mainline would include several such tags, and addons could use the tag too if they wish. Perhaps they can be nested within each other, so that `/` syntax is actually referring to the hierarchy of `[achievement_group]`s…

src/achievements.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+// TODO: add to schema
```
CelticMinstrel, line 37: Although this sounds important, from your description it doesn't seem to be crucial. You essentially just need to define an entirely new schema, right? These `achievements.cfg` are not following the game config schema like most of our other WML.

src/game_events/action_wml.cpp
```diff
@@ -239,20 +240,40 @@ wml_action::wml_action(const std::string & tag, handler function)
...
  * \endcode
  */
 #define WML_HANDLER_FUNCTION(pname, pei, pcfg) \
-	static void wml_func_##pname(const queued_event &pei, const vconfig &pcfg); \
+	static void wml_func_##pname(const queued_event& pei, const vconfig& pcfg); \
 	static wml_action wml_action_##pname(#pname, &wml_func_##pname);  \
 	static void wml_func_##pname(const queued_event& pei, const vconfig& pcfg)
 
 
+WML_HANDLER_FUNCTION(set_achieved,,pcfg)
```
CelticMinstrel, line 257: Ohhh? A new C++-implemented WML tag? I believe we ultimately would prefer to phase out this entire file, so I'm not sure if it's a good idea to add new stuff here. It might be better to add a Lua API function and then implement the tag in Lua.

src/preferences/general.hpp
```diff
@@ -274,4 +274,7 @@ namespace preferences {
 	sort_order::type addon_manager_saved_order_direction();
 	void set_addon_manager_saved_order_direction(sort_order::type value);
 
+	bool achieved(const std::string& content_for, const std::string& id);
+	void set_achieved(const std::string& content_for, const std::string& id);
```
stevecotton, line 278: The names are a little ambiguous - `achieved` is a verb in the past tense, so it could be the thing that means "set achieved".

Just because there's no documentation in the file so far doesn't mean that's a good thing. We should set an image for other developers to follow :grin:

CelticMinstrel: `has_achieved` would be better, yeah. Maybe `is_achieved`.

src/game_events/conditional_wml.cpp
```diff
@@ -48,6 +49,14 @@ namespace game_events {
 namespace builtin_conditions {
 	std::vector<std::pair<int,int>> default_counts = utils::parse_ranges("1-infinity");
 
+	bool has_achievement(const vconfig& cfg)
+	{
+		const config& acfg = cfg.get_config();
+		std::string content_for = acfg["content_for"].str();
+		std::string id = acfg["id"].str();
+		return preferences::achieved(content_for, id);
```
stevecotton, line 57: This should probably have anti-OOS protection in network multiplayer. However, maybe it should still be active in local multiplayer.

Elvish-Hunter:
> Allow [achievement] to have a field for icon. Just textual entries feels too lacking

In addition to what @knyghtmare proposed, I also suggest adding two more keys for completed achievements: `icon_completed=` and `description_completed=`, which will be shown once the achievement has been completed; if they're missing, the regular `icon=` and `description=` will be used instead.

> Ohhh? A new C++-implemented WML tag? I believe we ultimately would prefer to phase out this entire file, so I'm not sure if it's a good idea to add new stuff here. It might be better to add a Lua API function and then implement the tag in Lua.

Which reminds me that I should restart working on porting `[set_variables]` to Lua...

src/achievements.cpp
Pentarctagon: Huh, alright. It seems to be present whenever the vertical scrollbar is present and can only scroll right or left by a tiny amount (less than the width of the vertical scrollbar.

src/achievements.cpp
Pentarctagon: My thought is that ideally it'd be something that doesn't interrupt what the player is currently doing by forcing them to click to dismiss it, but I don't think (?) Wesnoth current has something like that.

src/achievements.cpp
Pentarctagon: At the moment it's part of the identifier as well as displayed on the UI, which is the issue. So yeah something like `display_name` would be needed, at minimum.

src/achievements.cpp
Pentarctagon: Alright, that's good. I was wondering if this was even something it could do, given the different way it loads the WML.

src/game_events/action_wml.cpp
Pentarctagon: What I wasn't sure about was where to put it - there didn't seem to be any good module to add the C++ part of the implementation to other than maybe `wml-utils`.

src/preferences/general.hpp
Pentarctagon: I followed the naming convention used in the rest of the file though.

src/game_events/conditional_wml.cpp
Pentarctagon: What do you mean by anti-OOS protection?

Pentarctagon:
@Elvish-Hunter @knyghtmare the issue with icons is that the achievements are loaded separately from anything else - it doesn't load the add-on to get the achievements, it loads *just* the achievements.cfg - so it wouldn't be able to find the actual image files. A DataURI might work though since I think (?) those are supported for free everywhere an image is supported.

src/preferences/general.hpp
stevecotton: The rest of the file is using nouns as the getters' names, so maybe `achievement` and `set_achievement`.

src/game_events/conditional_wml.cpp
stevecotton: Consider how these would work in co-op multiplayer mode:

* a bonus "make it a bit more difficult" mini-boss that only appears if the player has already completed the campaign
* a multiplayer co-op campaign where there's a shortcut in an early map that's unlocked if you've already completed the campaign
* a whose recruit list depends on which units they've killed in previous scenarios (like the leaders of SotA), but which persistently unlocks those recruits even when loaded in an unrelated match
* a unit with a complex special ability which pops up several dialogs to control what it actually does; and after complaints about players trying it and then quitting, gets changed to only unlock if the player has completed the scenario which explains it
* a unit with a special ability which, if used wrongly, can block progress for several turns, which unlocks when you've played a scenario that explains it

These can all already be done with persistent storage; however persistent storage already has warnings about multiplayer OOS and how to avoid it. If we're making an tag to read these values easily via ConditionalWML, what are you expecting it to be used for?

Pentarctagon: I understand that different clients will likely have different achievements completed, and as such isn't safe to use in networked multiplayer. If that's something we don't want to let people use due to it being incorrectly used in networked multiplayer, then I can of course remove the tag.

src/achievements.cpp
CelticMinstrel: Right, which is why I suggested a modeless dialog, which could pop up in the corner, then disappear a few seconds later. I think that would be the ideal form of the notification. However, as far as I know, event handling on the map is still incompatible with GUI2 event handling – the mere existence of a modeless window onscreen will prevent the map from receiving any events at all. So, unless that's solved, we'll have to use a modal dialog, and the object popup seems like the best choice for that.

src/achievements.cpp
CelticMinstrel: Yeah, to be clear, I'm not saying we don't need to fix it, rather that `horizontal_scrollbar_mode=never` on the listbox probably isn't going to do something because that's not the scroll box that's producing that scrollbar.

I'm not quite sure what the actual way to fix it is. Maybe @Vultraz has some ideas.

src/achievements.cpp
CelticMinstrel: Note that the first form of my proposal would place `display_name` as a toplevel attribute, in case that was unclear.

So the structure I thought up would be something like:

* Toplevel: contains `display_name=`, `[achievement]`, and `[achievement_group]`. The mainline file probably only has `[achievement_group]`.
* `[achievement_group]`: Contains `id=`, `display_name=`, `[achievement]` and `[achievement_group]`.
* `[achievement]`: Contains `content_for=`, `id=`, `icon=`, `name=`, `name_completed=`, `description=`, * `description_completed=`, `hidden=`. Maybe `content_for=` is optional and defaults to a path for the containing achievement group? Or even, it's only allowed in toplevel achievements?

…and that's basically your schema. Does that sound like it works?

The ID used for `[set_achieved]`, `[has_achievement]`, and `content_for=` could be derived from the following format: `[~Addon_Folder_Name / ] (achievement_group_id / )+ achievement_id`, with the achievement ID omitted in `content_for=`.

data/gui/window/mp_report.cfg
Pentarctagon: Reverted.

src/gui/dialogs/achievements_dialog.cpp
Pentarctagon: Updated.

src/achievements.cpp
CelticMinstrel: Well, any WML can be validated, certainly. But if it's not included from the core _main.cfg, then it would be a separate schema. We have a number of additional schemas already, so adding one for achievements would be quite doable (it doesn't seem like it'll be a very complex schema anyway).

src/preferences/general.hpp
Pentarctagon: Updated.

src/game_events/action_wml.cpp
CelticMinstrel: You said C++ but I think you meant Lua?

There's certainly no module for achievements. I think something like `wesnoth.achievements` would be the best place for this API. I think it might be nice if it can be used like this:

```lua
function wesnoth.wml_actions.set_achieved(cfg)
  -- Maybe like this?
  wesnoth.achievements[cfg.content_for][cfg.id] = true
  -- Or like this?
  wesnoth.achievements[cfg.content_for].achieve(cfg.id)
end

function wesnoth.wml_actions.has_achievement(cfg)
  -- Maybe like this?
  return wesnoth.achievements[cfg.content_for][cfg.id]
  -- Or like this?
  wesnoth.achievements[cfg.content_for].has(cfg.id)
end
```

As for what files those functions would go in, normally wml-tags.lua and wml-conditionals.lua would be the correct place, unless they were long, complicated functions, in which case they'd go in wml/tag-name.lua.

src/game_events/conditional_wml.cpp
CelticMinstrel: I think stevecotton's point is more that we should warn the addon developer if they're using it in a way that might be unsafe?

CelticMinstrel:
Or we could just document that you must use the full path to the image, rather than a path relative to the binary path. I don't recall the exact details, and I think it's undocumented, but there _is_ a way to specify images without a binary path entry. I believe it's used in the save cache? So even if it's undocumented, it's something we rely on, and I don't think there'd be an issue with documenting it and making it official.

src/achievements.cpp
Pentarctagon: I'm not sure why `display_name` would be a toplevel attribute? If there's an `[achievement_group]`, then it seems like it would make more sense to have `display_name` and `content_for` as required attributes of that tag? So for that (and based on other comments):
```
[achievement_group]
    display_name=
    content_for=
    [achievement]
        name=
        name_completed=
        description=
        description_completed=
        data_uri_icon=
        data_uri_icon_completed=
        hidden=
    [/achievement]
    [achievement]
        name=
        name_completed=
        description=
        description_completed=
        data_uri_icon=
        data_uri_icon_completed=
        hidden=
    [/achievement]
    # etc...
[/achievement_group]
```

src/game_events/action_wml.cpp
Pentarctagon: No, I meant C++. This can't be implemented in pure lua due to needing to write newly completed achievements to the preferences file, unless there's already an API for that. `has_achievement` also wouldn't go under wml_actions, would it?

src/achievements.cpp
CelticMinstrel: Well, we could do without it, but it also doesn't seem like we lose anything from making `[achievement_group]` essentially be the root tag, so that the contents of toplevel is the same as the contents of that tag. Then addons that only have a few achievements don't even need to define a group; there's already an implicit group for their addon.

From your code just there, it feels like you have a slightly different concept of `[achievement_group]` than I do. How does `content_for=` work in your model? Is it just an arbitrary ID referenced verbatim directly from `[set_achieved]` or `[has_achievement]`

src/game_events/action_wml.cpp
CelticMinstrel: > there didn't seem to be any good module to add the C++ part of the implementation to

Then, I have no idea what you mean by this. There's basically no code in the WML tag implementation.

> `has_achievement` also wouldn't go under wml_actions, would it?

Yes, that was my mistake, sorry. I corrected it.

CelticMinstrel:
Ah, and there's one other possibility – we could just support `[binary_path]` in the achievements file. That would create a small amount of code duplication but would probably be worth it for an addon that has a large number of achievements.

knyghtmare:
What I meant was icons sort of flesh out the UI and make it look visually appealing. See attached.
One other thing I noticed is that the Achievement title font could use some changing...like it make it bold and golden? The description can stay as is.

src/game_events/action_wml.cpp
Pentarctagon: > Then, I have no idea what you mean by this. There's basically no code in the WML tag implementation.

I mean that I need to write new C++ code *somewhere* to do the preferences reading and writing for these tags, don't I?

src/achievements.cpp
Pentarctagon: It's meant to be the group ID used by whatever content the achievements are associated to. However, each add-on can have multiple achievement groups (ie: achievements per scenario in a scenario pack) as well as add achievements to entirely unrelated content (ie: a modification that adds more achievements to mainline campaigns).

CelticMinstrel: Both my model and yours seem to achieve that… yours looks a little simpler but means we need to be more careful of ID clashes…

src/game_events/action_wml.cpp
CelticMinstrel: Well, there is basically no code in the existing functions for those tags. It's literally taking the input from the tag and passing it on to another function. So yeah, moving that into game_lua_kernel.cpp would be perfectly fine.

CelticMinstrel:
For what it's worth, I don't think it's good to use `data_uri_icon=` as an attribute. It's kinda long, and even if initially only data URIs are supported, it would mean the tag name needs to change if more URIs are supported later on.

I think there _probably_ aren't any issues in supporting "absolute" image paths, even without any extra work. I'm not sure if there would be issues supporting `[binary_path]` in the achievements file. (It might require converting all those paths to absolute paths on load?)

Speaking of "absolute" paths, I remembered one place where they're [at least sorta documented](https://wiki.wesnoth.org/LuaAPI/filesystem#filesystem.resolve_asset):

src/achievements.cpp
stevecotton: The translations for all installed addons are already loaded on the title screen, because they're needed for the campaign selection menu. I haven't checked, but it's probable that the WML loading and saving functions are already storing the textdomain in the preferences for you.

src/game_events/conditional_wml.cpp
stevecotton: Yes, warn the addon developer, assuming that if it's accessible as ConditionalWML then addon developers will cut & paste without noticing surrounding warnings. However, we can assume that MP developers will at least test their MP code once...

So an OOS error `if(resources::controller->is_networked_mp() && ! in mp_sync::get_user_choice() or similar)`.

Pentarctagon: Alright, that makes more sense to me. It's not protection as in preventing it from happening, it's warning that it happened.

src/achievements.cpp
CelticMinstrel: With the approach Pentarctagon built here, we can't use the campaign title as the display name. An addon is a container that contains one or more root tags such as a campaign, an era, etc. There might not even _be_ a campaign.

I suppose it would be possible to use the addon name though, which would be loaded from _server.pbl or info.cfg.

Oh, but are you trying to say that translatable strings are still serializable? I think that's probably true, but even so, it doesn't seem right to use a translatable string as an identifier.

Pentarctagon: In this case, I believe it would result in (for example) achievements being "lost" if the player changes the language.

CelticMinstrel: I'm not confident that that's true, but even if it's not true, it doesn't seem like a good idea.

stevecotton: I just meant the display_name would probably be serializable without breaking the translation.

Pentarctagon: @CelticMinstrel What I mean is that if a translatable value were used as an identifier as well then if:
* Someone completed an achievement while using the French translation.
* Closed Wesnoth.
* Started Wesnoth and changed the language to German.
Then they wouldn't have that achievement marked as complete anymore since it would be comparing the French string (from the preferences file) to the German string (current language).

@stevecotton Thanks, good to know.

Pentarctagon:
@CelticMinstrel would you be able to update the xcode projectfiles for the new source files added? I've been trying to get it to work with the `add_source_file` script but the xcode builds on CI keep not seeing them for some reason.

src/achievements.cpp
CelticMinstrel: I got what you meant, and I suspect that wouldn't happen, but even then, I don't think it's a good idea to do things that way.

CelticMinstrel:
Could it be that `add_source_file` adds it to one target but it needs to be added to two targets? I don't really know, just speculation…

Pentarctagon:
I tried a few variations of adding the new files to the two targets (one for the wesnoth executable and one for the tests), and the script reports that the files exist there, so it shouldn't? But clearly it's not adding something that it needs to.

src/achievements.cpp
Pentarctagon: This has been addressed/completed.

src/achievements.cpp
Pentarctagon: Random thought: could floating text potentially be used for this? There'd need to be a way to add a "background" color behind the text to make it more visible and readable, but that wouldn't block input to the rest of the map while it's displayed.

CelticMinstrel:
I do wonder if it's actually even worthwhile having a macro to define an achievement especially when most of the arguments are optional. I guess it does make things _slightly_ less verbose-looking…

src/achievements.cpp
CelticMinstrel: Yes, I think that would be doable. You could make the text relatively large and have it appear in the corner of the screen. The only downside of this approach is that floating text is just text. I don't think there's any way to throw in the icon.

src/achievements.cpp
```diff
@@ -89,20 +84,27 @@ config achievements::read_achievements_file(const std::string& path)
  */
 void achievements::process_achievements_file(const config& cfg, const std::string& content_source)
 {
-	for(const config& c : cfg.child_range("achievement")) {
-		const std::string content_for = c["content_for"].str();
-		const std::string id = c["id"].str();
+	for(const config& achgrp : cfg.child_range("achievement_group")) {
```
CelticMinstrel, line 87: So this implementation implies:

* `[achievement_group]`s are not nestable
* There is no "implicit" `[achievement_group]` for an add-on. That is, an add-on with only a few achievements can't simply omit the tag and just drop `[achievement]` tags at toplevel.

I'm mentioning this mainly because it differs from the vision I had. Whether these are actually an issue, I'm not sure.

data/achievements.cfg
```diff
@@ -1,18 +1,43 @@
...
+
     [achievement]
-        content_for={CONTENT_FOR}
         id={ID}
         name={NAME}
+        name_completed={NAME_COMPLETE}
         description={DESC}
+        description_completed={DESC_COMPLETE}
+        data_uri_icon={DATA_ICON}
+        data_uri_icon_completed={DATA_ICON_COMPLETE}
```
CelticMinstrel, line 26: As mentioned before, I don't like encoding the format into the key, as that limits us from expanding it later. I'd prefer just `icon` and `icon_completed` even if they truly are limited to DataURIs only.

Pentarctagon: Testing it out, it looks like it's plain incorrect anyway since full paths (ie `data/add-ons/ANLEra/images/halo/shadow-mage-halo6.png` or `data/core/images/icons/bag.png`) do also work as-is.

src/achievements.cpp
Pentarctagon: > [achievement_group]s are not nestable

I'm not really clear what the benefit of allowing this would be.

> There is no "implicit" [achievement_group] for an add-on. That is, an add-on with only a few achievements can't simply omit the tag and just drop [achievement] tags at toplevel.

Needing to add an outer [achievement_group] even if there's only a few achievements doesn't seem like an issue to me.

data/achievements.cfg
CelticMinstrel: Yeah, I suspected that would already work.

src/achievements.cpp
CelticMinstrel: > I'm not really clear what the benefit of allowing this would be.

I guess the benefit would be that the group IDs have a lower chance of conflicting across add-ons? That's not that huge of a benefit tho…

Pentarctagon:
> I was also thinking just now that achievements with a progress counter are a common feature that this system doesn't support. For example something like "Kill 10 gryphons" would show how many you've killed so far, and be marked as complete when it reaches 10. Is that difficult to add on here?

I'd prefer to get this merged then separately add more features like this to the achievements. At a minimum it'd require a new tag (`[update_achievement]`?), a separate way to store in-progress achievements (right now completed achievements are just a comma delimited list), and there'd need to be a way to mark them as completed when killing the 10 gryphons (or whatever it happens to be) has been done.

src/achievements.cpp
Pentarctagon: This has been completed - the attribute is `characters_per_line` for the label.

src/achievements.cpp
Pentarctagon: This has been completed - the macro is `GUI_FORCE_WIDGET_MINIMUM_SIZE`.

data/achievements.cfg
Pentarctagon: Updated to just `icon` and `icon_completed`.

src/achievements.cpp
Pentarctagon: My guess/assumption would be that if there's an unintended conflict (as opposed to an intentional one where one add-on is adding more achievements to other content), UMC authors would add a prefix or a suffix to their IDs rather than nesting their `[achievement_group]`s.

stevecotton:
> I'd prefer to get this merged then separately add more features like this to the achievements. At a minimum ... there'd need to be a way to mark them as completed when killing the 10 gryphons (or whatever it happens to be) has been done.

I'd be tempted to say "leave that example to a campaign's WML to do". It would restrict it to the goal of "kill 10 gryphons in a single play-through"`*`, but the UMC author has a better understanding of what the goal is, so can fit the goal-in-progress UI to their goal. For the gryphons, maybe a floating-text that rises from each kill.

`*` unless the UMC author added a lot more complexity

Pentarctagon:
The limitation is that there wouldn't be a way for UMC authors to display that progress on the achievement itself, which would then make it more difficult for players to know how close they are to completing an achievement.

boost_test_schedule
```diff
@@ -53,6 +53,7 @@ test_gui2/modal_dialog_test_wml_error
 test_gui2/modal_dialog_test_wml_message_left
 test_gui2/modal_dialog_test_wml_message_right
 test_gui2/modal_dialog_test_wml_message_double
+test_gui2/modal_dialog_test_achievements_dialog
```
stevecotton, line 56: The lines around this are in alphabetical order, please move this one.

data/campaigns/Descent_Into_Darkness/scenarios/10_Endless_Night.cfg
```diff
@@ -1906,6 +1906,11 @@ I can still hear his infernal cackling ringing in my head. Sure thing, Zephrin,
     [event]
         name=enemies defeated
 
+        [set_achieved]
+            content_for="Descent into Darkness"
+            id="did_finished"
```
stevecotton, line 1911: DiD's final scenario is weird - this isn't the end of the campaign. The only things that are the end of the campaign are the routes that lead to `next_scenario=null`.

src/achievements.cpp
```diff
@@ -0,0 +1,107 @@
...
+// TODO: some way to display in-game when something is achieved?
+
+// TODO: add to schema
+// TODO: unit tests?
+
+/**
+ * Reads the mainline achievements.cfg and then all the achievements of each installed add-on.
+ *
+ * This is intentionally handled separately from other WML loading so that:
+ * a) All achievements and their status are able to be displayed on the main menu right after Wesnoth starts and regardless of which add-ons are active.
```
stevecotton, line 37: This seems overly complex, given that all add-on's `[campaign]`, `[modification]` or similar tags already have to be available at the title screen. I think this should be able to fit in with that mechanism.

src/achievements.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include "config.hpp"
+
+struct achievement
+{
+	std::string id_;
+	std::string name_;
+	std::string name_completed_;
+	std::string description_;
+	std::string description_completed_;
```
stevecotton, line 30: Several of these look like they should be `tstring`s.

src/achievements.hpp
```diff
@@ -0,0 +1,73 @@
...
+struct achievement_group
+{
+	std::string display_name_;
+	std::string content_for_;
+	std::vector<achievement> achievements_;
+
+	achievement_group(const config& cfg);
+};
+
+class achievements
```
stevecotton, line 59: A bit of documentation would be good. From the names, I'd assume `achievement_group` and `achievements` would be the same thing.

src/game_events/conditional_wml.hpp
```diff
@@ -34,5 +34,6 @@ namespace game_events
 		bool have_unit(const vconfig& cfg);
 		bool have_location(const vconfig& cfg);
 		bool variable_matches(const vconfig& cfg);
+		bool has_achievement(const vconfig& cfg);
```
stevecotton, line 37: Now would be a good time to sort these into alphabetic order.

src/achievements.cpp
CelticMinstrel: That's not true at all. Only add-ons intended for use in single-player need to be visible at the title screen. Anything written as multiplayer-only does not have to be (and usually isn't).

src/achievements.hpp
CelticMinstrel: Yes, all of these four should certainly be `t_string`.

The icon should not be; it can be localized in a different manner if really needed, though I don't think it should need localization at all.

src/achievements.cpp
stevecotton: Ah true, anything inside `#ifdef MULTIPLAYER` will be missing. However, instead of having to have an extra WML-loading routine in C++, would it be better for any multiplayer add-ons that include achievements to have some achievement-specific WML outside their `#ifdef MULTIPLAYER`block?

data/campaigns/Descent_Into_Darkness/scenarios/10_Endless_Night.cfg
Pentarctagon: You have nevertheless completed all scenarios in the campaign. It would be strange to not have any achievement like that for DiD.

src/achievements.cpp
Pentarctagon: That would still limit the viewable achievements when playing a game, since add-ons that aren't used are inactive, wouldn't it?

src/game_events/conditional_wml.hpp
Pentarctagon: It's been moved to `wesnoth.achievements` in `game_lua_kernel.cpp` at celticminstrel's request.

boost_test_schedule
Pentarctagon: Updated.

src/achievements.hpp
Pentarctagon: Updated.

src/game_events/action_wml.cpp
Pentarctagon: I've added it instead as `wesnoth.achievements.get()` and `wesnoth.achievements.has()`. I'm not really sure how to have it use the square brackets for access - is that what `__index` and `__newindex` are for?

src/game_events/conditional_wml.cpp
Pentarctagon: This has been completed.

src/achievements.hpp
Pentarctagon: Added.

src/achievements.hpp
```diff
@@ -51,18 +63,30 @@ struct achievement
 		if(description_completed_ == "") {
 			description_completed_ = description_;
 		}
+		if(icon_completed_ == "") {
```
CelticMinstrel, line 66: I would usually use `.empty()` instead of `== ""`…

src/scripting/game_lua_kernel.cpp
```diff
@@ -3089,7 +3089,13 @@ int game_lua_kernel::intf_has_achievement(lua_State *L)
 {
 	const char *content_for = luaL_checkstring(L, 1);
 	const char *id = luaL_checkstring(L, 2);
-	lua_pushboolean(L, preferences::achievement(content_for, id));
+
+	if(resources::controller->is_networked_mp() && synced_context::is_synced()) {
+		ERR_LUA << "Returning true for whether a player has completed an achievement due to being networked multiplayer.";
+		lua_pushboolean(L, true);
```
CelticMinstrel, line 3095: Wouldn't false be a more natural fallback here?

data/campaigns/Descent_Into_Darkness/scenarios/10_Endless_Night.cfg
CelticMinstrel: I think you won't achieve the wreath in this circumstance, right? While it's not strictly required that the wreath and the "completion" achievement be equivalent, I think it makes sense for them to be.

src/game_events/action_wml.cpp
CelticMinstrel: Yes. There are examples of it being done for other APIs. Using `get` and `has` isn't a terrible API though, and if you think there's some reason to possibly add other functions in the future, there might be some argument that it's better. Personally, barring someone thinking of such a potential future API, I still prefer square-bracket access.

The only other function I can think of would be a way to iterate through the available achievements (or is there some reason that's a bad idea?). Of course, it _would_ be nice if that could just be done with `pairs`…

data/campaigns/Descent_Into_Darkness/scenarios/10_Endless_Night.cfg
stevecotton: It won't get the wreath.

For the purposes of the achievements PR, I'd suggest using a different campaign. Using TSG would allow giving separate achievements for the two campaign branches.

src/achievements.cpp
stevecotton: True. I'll mark this resolved.

data/campaigns/Descent_Into_Darkness/scenarios/10_Endless_Night.cfg
knyghtmare: Another option would HttT, or achievements can be offered for the branches, 5A, 5B, 19A, 19B, 19C, 20A, 20B, and even 2 special achievements if Konrad/Lisar take specter of Fire

knyghtmare: > DiD's final scenario is weird - this isn't the end of the campaign. The only things that are the end of the campaign are the routes that lead to next_scenario=null.

I would like to point that you can complete DiD's Endless Night by 3-ways: (1) Do it 3x times or so, and let some enemy take the book, (2) get demolished by White Fang Orcs, (3) Get demolished by enemy after 3 wins

src/scripting/game_lua_kernel.cpp
```diff
@@ -3064,6 +3065,41 @@ int game_lua_kernel::intf_play_sound(lua_State *L)
...
+ * - Arg 1: string - content_for.
+ * - Arg 2: string - id.
+ * - Ret 1: boolean.
+ */
+int game_lua_kernel::intf_has_achievement(lua_State *L)
+{
+	const char *content_for = luaL_checkstring(L, 1);
+	const char *id = luaL_checkstring(L, 2);
+
+	if(resources::controller->is_networked_mp() && synced_context::is_synced()) {
```
stevecotton, line 3093: Commenting to test later - is `is_synced` the right condition? I'd assume we'd want this fallback anytime that we're not in a synchronised choice, which IIUC is different to being in a synchronised event.

knyghtmare:
I have listed like 41 possible achievements for HttT in a draft file (will open new tracker issue later oin) so I think they can added AFTER this PR is merged.

Why HttT? Because it's the flagship campaign and it's achievement set should come first before all others.

Why after merge? well, the PR implements the system, but I think shoving an achievement set of a whole campaign into it is out of scope.

CelticMinstrel:
Does that list include an achievement for… what was his name again… Moremirnu or something? The priest from the island.

knyghtmare:
> Does that list include an achievement for… what was his name again… Moremirnu or something? The priest from the island.

Well, yes. 2 actually. (those familiar with HttT's plot design will understand why)

src/game_events/action_wml.cpp
Pentarctagon: I think I'll leave it as `get()` and `has()`. There's the additional functionality of having "in progress" achievements that's been mentioned and it wouldn't surprise me at all if people came up with more things to do with achievements.

src/scripting/game_lua_kernel.cpp
Pentarctagon: If not, then https://github.com/wesnoth/wesnoth/blob/0eb6ecb42e3c9a3c1a283745441cf58c4390a255/src/side_filter.cpp#L220 also needs to be fixed.

src/scripting/game_lua_kernel.cpp
Pentarctagon: Updated.

src/game_events/action_wml.cpp
CelticMinstrel: I could imagine "in progress" achievements to be distinguished as being int-valued instead of bool-valued if using the bracket syntax, but I don't think I mind leaving it as it is.

But, is there any reason not to add a third function that lists all available achievements? (Not sure how that function would handle hidden achievements. It could exclude them, include them but label them as hidden, include them only if the user has them unlocked, or take an input to specify whether to include them, which could be either a boolean true/false or an enum string selecting one of the first three options I listed here.)

utils/CI/schema_validation.sh
```diff
@@ -33,6 +33,7 @@ validate() {
 
 validate_core() { validate "$1" ./wesnoth --validate data/_main.cfg; }
 validate_misc() { validate "$1" ./wesnoth --data-dir=. --validate=data/_main.cfg --preprocess-defines="$2"; }
+validate_achievements() { validate "Achievements schema" ./wesnoth --data-dir=. --validate=data/achievements.cfg; }
```
CelticMinstrel, line 36: This is validating the achievements, not the achievements schema, right?

data/schema/achievements.cfg
```diff
@@ -0,0 +1,23 @@
+[wml_schema]
+	{./macros.cfg}
+	{./types/basic.cfg}
+	[tag]
+		name="root"
+		[tag]
+			name="achievement_group"
```
CelticMinstrel, line 7: min and max default to 0 and 1, respectively. At least the latter is clearly wrong.

src/game_events/action_wml.cpp
Pentarctagon: Mainly just that I couldn't think of a use case for where it would be useful.

src/achievements.cpp
Pentarctagon: This has been added.

data/campaigns/Descent_Into_Darkness/scenarios/10_Endless_Night.cfg
Pentarctagon: Would it make more sense to just remove the DiD completion achievement, at least for now?

utils/CI/schema_validation.sh
Pentarctagon: Yep, fixed.

data/schema/achievements.cfg
Pentarctagon: Fixed as well.

data/campaigns/Descent_Into_Darkness/scenarios/10_Endless_Night.cfg
stevecotton: Yes, please remove it.

Pentarctagon: It has been removed.

Pentarctagon:
@Vultraz updated the screenshot in the first post with what you've mentioned on Discord so far.

Pentarctagon:
@CelticMinstrel @knyghtmare 

So to summarize the Discord conversation:
* Grey out the icon for uncompleted achievements (basically just append `~GS()` to the icon).
* Add some sort of way to indicate that an achievement is completed that's independent of anything the person who wrote the achievement does - ie: make the text green, give it a gold frame, etc.

Also, I think this was mentioned elsewhere, but instead of entirely hiding achievements with `hidden=yes` to instead have them be present but have the text displayed be something generic like "Hidden".

Also also, it'd probably be good to how the number of achievements vs the number achieved somewhere.

CelticMinstrel:
> * Grey out the icon for uncompleted achievements (basically just append `~GS()` to the icon).

The only question is whether we do this in the system, or in the mainline achievement definitions.

> Also also, it'd probably be good to how the number of achievements vs the number achieved somewhere.

Speaking of which, I wonder if there should be an "all" category in the dropdown menu that mixes together the achievements from all content into one gigantic list…

Pentarctagon:
> The only question is whether we do this in the system, or in the mainline achievement definitions.

In the system would make more sense, I think.

> Speaking of which, I wonder if there should be an "all" category in the dropdown menu that mixes together the achievements from all content into one gigantic list…

I feel like that'd end up really overcrowded, even with just mainline achievements once those are adde.

CelticMinstrel:
> In the system would make more sense, I think.

I kinda don't feel that this is a case where one way makes more sense than the other. The question is, do we want to force all achievements to be greyed out when they're not achieved? Or is that simply a convention we're choosing to apply to the mainline achievements? It's not "self-evident" that achievement icons should be greyed out when not achieved. It could make sense to leave it up to the add-on author. But there's also no reason we have to allow an add-on author to use a colour icon when the achievement isn't yet achieved.

So, if you want to do it in the system, that's fine.

knyghtmare:
> Add some sort of way to indicate that an achievement is completed that's independent of anything the person who wrote the achievement does - ie: make the text green, give it a gold frame, etc.

image

This needs tweaking but I can use these pop-up cards for now.

knyghtmare:
> Also, I think this was mentioned elsewhere, but instead of entirely hiding achievements with hidden=yes to instead have them be present but have the text displayed be something generic like "Hidden".

I think it's fine to show the hidden ones obscurely. Perhaps maybe add an new `hint_text=` to give a clue where this hidden achievement can/may be acquired. Since just showing a greyed icon with "HIDDEN" doesn't clarify much.

> Also also, it'd probably be good to how the number of achievements vs the number achieved somewhere.

I guess it can be done but I am skeptical if it should be done as it might promote elitism.
This is actually maybe beyond the scope of the PR and in consideration of a follow-up PR.

CelticMinstrel:
> Perhaps maybe add an new `hint_text=` to give a clue where this hidden achievement can/may be acquired.

Normally, `description` is used when the achievement is not achieved, and `description_completed` is used when it is achieved, right? Could that not be exactly he same for hidden achievements? So `description` is the hint text, in effect.

The only difference with hidden achievements would be what happens if `description_completed` is omitted. For a non-hidden achievement, that means the same description shows when it's achieved or not achieved, but for a hidden achievement, it would only show that description when achieved and show a generic description when not achieved.

Pentarctagon:
With this last commit, I believe everything has been addressed. Updated images for the UI have been added to the first post.

CelticMinstrel:
Looks like your manual edit attempt worked this time.

projectfiles/Xcode/The Battle for Wesnoth.xcodeproj/project.pbxproj
```diff
@@ -1502,8 +1502,8 @@
...
 		1234567890ABCDEF12345681 /* file_progress.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = file_progress.hpp; sourceTree = "<group>"; };
-		000000000000000000000001 /* achievements.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = achievements.cpp; sourceTree = "<group>"; };
-		000000000000000000000002 /* achievements.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = achievements.hpp; sourceTree = "<group>"; };
+		000000000000000000000001 /* achievements.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; name = achievements.cpp; path = ../../src/achievements.cpp; sourceTree = SOURCE_ROOT; };
+		000000000000000000000002 /* achievements.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; name = achievements.hpp; path = ../../src/achievements.hpp; sourceTree = SOURCE_ROOT;  };
```
CelticMinstrel, line 1506: ?????

What? Why would you do that? Everything's relative to its group.

…wait, did you maybe forget to add it to the group?

CelticMinstrel:
As I understand it, adding a file to Xcode has three basic requirements:

1. Declare the file.
2. Reference it in a group.
3. Reference it in one or more targets.

projectfiles/Xcode/The Battle for Wesnoth.xcodeproj/project.pbxproj
Pentarctagon: I don't even know what a group in xcode is.

CelticMinstrel: It's what Visual Studio calls a filter. Basically a container for source files within the project, to just group them together independently of how they're stored on the filesystem.

Using `sourceTree = "<group>"` is the preferred way of referencing files that exist within the source root. Our project links each group to a physical folder, so if the file is relative to the group and placed in the correct group, then it should have no issues finding it.

PR #7239, opened 2022-12-28
Validate _server.pbl files on load against the schema

CelticMinstrel:


cooljeanius:
Seems like the CI failures might actually be relevant, since they're validator-related... but then why do they only happen on the 2 ubuntu ones, I wonder?

CelticMinstrel:
Ah, I see. It's the middle commit. It's because that fixed a bug that hid a large number of errors in the game config schema, and now those errors are showing up like they should have a long time ago.

I'll fix it.

Pentarctagon:
@cooljeanius because the schema validation is only done during those two jobs.

cooljeanius:
> @cooljeanius because the schema validation is only done during those two jobs.

ah ok gotcha

CelticMinstrel:
So, now that this is functional and passing tests, I have a question.

This change means that if we change the format of the _server.pbl again in the future but forget to update the schema, addons using the new features would fail to load. Unlike with the schema validation of the main game config and GUI2 config, there's no way to verify in CI that the pbl schema was correctly updated.

Does this seem like an issue at all? I'm not really sure, it's just something I thought of when I made this change.

Pentarctagon:
Would it be possible to add a dummy or example pbl file that could be used for testing validation? Not sure what else could be done to test it otherwise.

CelticMinstrel:
It would be possible, but I don't think it actually ends up changing anything. It would still be a problem if the example pbl file and the schema were both not updated, right?

Pentarctagon:
That'd be true, yeah.

stevecotton:
There's no `changelog_entries/something.md` file, which is (now) the first place I'd look for the answer to this question:

> This change means that if we change the format of the _server.pbl again in the future but forget to update the schema, addons using the new features would fail to load.

When you say "on load", do you mean
* "anytime it reads the file"
* "only when following https://wiki.wesnoth.org/ValidationFAQ "
* "only in preparation before uploading to the server"
?

CelticMinstrel:
The first one. Basically that means that if you have an error you'll see it on first launch or on F5. If you ignored the error I suppose you'd also see it when you try to start a game or the map editor, but since the expected state is that you fix the error and retry, that doesn't seem like a major issue to me.

data/schema/pbl.cfg
```diff
@@ -34,9 +33,25 @@
...
+			[not]
+				forum_auth=yes
+			[/not]
+			[then]
+				{SIMPLE_KEY passphrase string}
+				{SIMPLE_KEY email email}
+			[/then]
+		[/if]
 		[tag]
 			name="feedback"
```
stevecotton, line 46: Although there's a `feedback` tag identified here, when passed the current version of Non-Fatal Wounds, https://github.com/stevecotton/Non_Fatal_Wounds/blob/d03b4e3/_server.pbl , it gives the error
`Tag [feedback] may not be used in []`.

stevecotton:
> Basically that means that if you have an error you'll see it on first launch or on F5.

That assumes the UMC author is starting Wesnoth specifically to work on their add-on, rather than to play a game. It also assumes that `_server.pbl` always indicates that the user is the UMC author, which while likely isn't guaranteed - they might have installed a version from the add-on's Git repo for testing.

This will make it harder to deprecate anything in PblWML, because any deprecations there will cause this dialog to appear in addition to standard deprecation messages. Note that other developers have decided to make deprecations less visible than I'd have chosen to make them, and I've had to argue to keep a dialog box like this in for "your save file is broken, in an hour or so you'll hit an Unknown Scenario error", which is a lot less trivial that a non-validating `_server.pbl`.

Instead of forcing the UMC author to urgently worry about a trivial issue, it would be much better to improve the https://wiki.wesnoth.org/ValidationFAQ so that the UMC authors can easily find the error when they're ready to deal with it.

Pentarctagon:
> Note that other developers have decided to make deprecations less visible than I'd have chosen to make them, and I've had to argue to keep a dialog box like this in for "your save file is broken, in an hour or so you'll hit an Unknown Scenario error", which is a lot less trivial that a non-validating _server.pbl.

I might be misremembering at this point, but IIRC most of the debate around that was more related to it being too wordy rather than it shouldn't be done at all.

For this PR though, it sounds like it would be better to show the error on upload rather than on starting wesnoth or using F5.

data/schema/pbl.cfg
CelticMinstrel: Looks like I messed up the last-second change to make it use `get_wml_location`. Pushing a fix asap.

CelticMinstrel:
> That assumes the UMC author is starting Wesnoth specifically to work on their add-on, rather than to play a game.

But the only side-effect is that that specific add-on isn't considered ready for upload by the add-ons manager. You can even still play the add-on if you want to.

Admittedly, it could get a little annoying having to dismiss it each time the game reloads the game config with different defines.

> This will make it harder to deprecate anything in PblWML, because any deprecations there will cause this dialog to appear in addition to standard deprecation messages.

That's only if we delete the definitions of the deprecated keys or tags from the pbl schema. We wouldn't do that as long as the game still knows how to handle them.

> Instead of forcing the UMC author to urgently worry about a trivial issue

I'd hardly call "your pbl file is wrong" a trivial issue… it's not a major one, I suppose, but I wouldn't call it trivial.

> For this PR though, it sounds like it would be better to show the error on upload rather than on starting wesnoth or using F5.

This seems like it could be a little tricky, as right now I literally just modified the place the pbl file is read to enable validation. But I suppose I could look into how hard it is to pass a boolean specifying whether to validate.

I personally think showing it at the title screen is a good idea. Showing it when you're starting a game (even one that doesn't even use the add-on) is annoying though. It should definitely be shown when you start up the add-ons manager.

Pentarctagon:
My main concern for showing it on wesnoth starting or using F5 is that with `forum_auth` support it's no longer necessary to have sensitive information (email and passphrase) stored in the pbl file, which means it's safe to include the pbl file as-is in a git repository, which means there may be more people who would see this error message despite not being the author.

That said I suppose it's still probably a fairly unlikely thing to happen.

CelticMinstrel:
I think it could become quite annoying if you have several add-ons with invalid pbls files, so I could see some merit in only validating if you're running the add-ons manager. (Perhaps also at the title screen if running with `--debug`?)

That said, it's _only_ annoying. It doesn't make you unable to play. Even the add-on that's broken will still work in all other respects.

CelticMinstrel:
Turned out to be easier than I expected. As of that last commit, it only shows the error when you try to launch the addons manager. The only downside is that if there's an error in any addon, the manager won't launch at all; it should be easy to fix that though if this is the direction we want to take.

I'd still like to also show the error at the title screen if you launch with `--debug`, but I'm not sure how to do that. Let me know if you have any suggestions, otherwise I'll poke at it a bit later.

Pentarctagon:
I was thinking you'd just check `game_config::debug`, but that covers more than just then `--debug` is used. Maybe that's close enough though, given this would be checked on the title screen? Though it would need to be set on `game_config` earlier than it is now (in `game_launcher.cpp`).

CelticMinstrel:
> I was thinking you'd just check `game_config::debug`

That should be half of it; the other half is, how do I determine from the game_config_manager that I'm at the title screen?

stevecotton:
> For this PR though, it sounds like it would be better to show the error on upload rather than on starting wesnoth or using F5.

> when you try to launch the addons manager

I feel it's good to show it when the user attempts to upload, because at that point it relates to what they're doing. For any other time, including opening the Add-ons Manager, why is it urgent enough to justify a modal dialog?

> I might be misremembering at this point, but IIRC most of the debate around [the you'll hit an Unknown Scenario dialog] was more related to it being too wordy rather than it shouldn't be done at all.

You're remembering the recent debate correctly, but that's just the recent one. There was [some in the PR itself](https://github.com/wesnoth/wesnoth/pull/5841#issuecomment-877906350); the [IRC at the time](https://irclogs.wesnoth.org/2021/06/%23wesnoth-dev.2021-06-08.log) covered both wordyness and "definitely not worth smacking the player in the face with a modal dialog" until I added a unit-test to ensure it wouldn't pop up for false positives.

CelticMinstrel:
> I feel it's good to show it when the user attempts to upload, because at that point it relates to what they're doing.

Hmm, to some degree that makes sense, but I also feel that it makes sense for an add-on with an invalid _server.pbl to not be listed as ready to publish, which requires validation to be done when the add-ons manager is open.

It also doesn't seem right to build the system with the assumption that users will have an invalid _server.pbl. Even if they got it from some add-on author's GitHub, it should still be a valid file. I think the assumption should be that normally, all _server.pbl files are valid, but there may occasionally be an invalid one, which 99% of the time will be the one the user is planning to upload anyway.

stevecotton:
Why would an error in a `_server.pbl` be so urgent that it's worth showing two error dialogs and then refusing to open the add-on manager at all?

Ask @Vultraz if you want to overrule me on that (as it's a UI decision), but for me it's a hard "this should never be the UX".

Vultraz:
I'd say preferable to disable the upload button and have its status show as "error in pbl file" or something like that. But we shouldn't have error popups, either when starting wesnoth or opening the addon manager.

Vultraz:
Why only in debug mode? Doesn't that somewhat defeat the purpose?

CelticMinstrel:
IMO the best place is:

1. At title screen in debug mode
2. On addons manager start (debug or no)

IOW if a dev runs with `--debug` they will see the issue immediate. If they don't, they'll see the issue when they open the addons manager to upload it. (The addons manager failing to start in this case is a bug, which I can easily fix.)

stevecotton:
> One error dialog. No refusing to open the addons manager. Where'd you get the idea that there's two?

If there's a translation mark on the `description=`, then two dialogs appear. `An error due to possibly invalid WML occurred` and `A local file with add-on publishing information could not be read.`

CelticMinstrel:
Okay, the "two dialogs" was a direct result of not catching the exception, which I've now fixed.

It now has two dialogs on startup if there's errors and you run with `--debug`, but that seems more minor since I imagine _most_ people won't run with `--debug`?

stevecotton:
Why is it so urgent for the user to fix it that it requires any error pop-ups, except at the time that the user is actually trying to upload it?

Why connect it to debug mode at all? Why would someone run in debug mode if they're thinking that their add-on is ready to upload and that's the task that they're starting Wesnoth to do? If someone's started debug mode for a different task, why is it useful to bug them about fixing an add-on first?

CelticMinstrel:
I don't understand why you hate the error popup so much in the first place. It's an exceptional circumstance that normally shouldn't happen at all, isn't it? Isn't it fine for it to be just a little bit annoying? Obviously it's not good for it to be super annoying and keep popping up all over the place, but it no longer does that.

I don't really have any reason to connect it to debug mode other than the fact that seeing the error on launch seems quite useful, yet you seem to disagree. The thought wasn't that someone would run in debug mode because they think their add-on is ready to upload but rather that someone working on an add-on might run habitually in debug mode.

cooljeanius:
yeah I run Wesnoth habitually in debug mode... but I also don't like error pop-ups much, so...

CelticMinstrel:
There won't be an error popup if there's no error, and if there's an error, wouldn't you want to fix it?

stevecotton:
> There won't be an error popup if there's no error, and if there's an error, wouldn't you want to fix it?

That's a question that indicates you haven't understood the difference between "urgent" and "important". None of the answers that you've given seem to answer the question of why it's sufficiently urgent to justify an error dialog.

CelticMinstrel:
That's a response that appears to indicate you have no interest in making me understand.

stevecotton:
I'm in complete agreement with what Vultraz said:

> I'd say preferable to disable the upload button and have its status show as "error in pbl file" or something like that. But we shouldn't have error popups, either when starting wesnoth or opening the addon manager.


For me, showing an error pop-up if the user clicks on the upload button is reasonable (and it would hopefully be greyed-out too). However, showing it at other times is interrupting a person who is doing a different task (or maybe just wanting to play).

CelticMinstrel:
> For me, showing an error pop-up if the user clicks on the upload button is reasonable (and it would hopefully be greyed-out too).

How are those two things compatible? You can't click on a greyed-out button.

stevecotton:
You could make a UI widget that looks the same as a greyed-out widget, but reacts to bring clicked on by showing a message about why the normal functionality is disabled.

Pentarctagon:
There are a variety of other error popups that can happen when uploading an add-on and there's some issue, so I don't see any reason why this couldn't be another one of those.

CelticMinstrel:
> You could make a UI widget that looks the same as a greyed-out widget, but reacts to bring clicked on by showing a message about why the normal functionality is disabled.

I'm not sure why you would think that could _possibly_ be a good idea…

> There are a variety of other error popups that can happen when uploading an add-on and there's some issue, so I don't see any reason why this couldn't be another one of those.

So your position is that it should be listed as a publishable addon even though it doesn't have a valid pbl file?

Pentarctagon:
> So your position is that it should be listed as a publishable addon even though it doesn't have a valid pbl file?

I'd say my position is essentially "it already shows an error popup when you attempt to upload and it can't find the icon image, so showing a different popup when schema validation fails is very similar".

CelticMinstrel:
Presumably this can now be squash-merged, if this latest commit is acceptable.

Pentarctagon:
It looks fine to me, at least.

stevecotton:
It seems okay. Minor point that doesn't block approval: the error dialog includes the very user-facing rather than UMC-author facing "When reporting the bug please include the following error message :".

stevecotton (approved):

PR #7244, opened 2022-12-30
1.17 more fauna work

doofus-01:
I've been slowly making my way through WoF, which makes use of random fauna.  I've also got the upgrade in #7228 in mind.  Also, also, also, #7236 made me realize there is a carbon-copy zombie scorpion variation that should not be a model for the future.

So, with all that in mind, I think we could use a little more variety and reorganization within existing monster/animal races (which could be renamed, Falcons -> Birds/Avians for example).

In this PR, I plan to add some ants and a raven, and get it over with in a couple days.  I know, the elephant in the room is actually a firedragon, and I hope to address that later this weekend, but it's not here.


knyghtmare:
1. what is "Test Ant"?
2. Is it possible to make the Firebomb Ant into a bomb? Like maybe some special/ability that makes the ant explode if critically wounded (HP lowered to 30%)?

stevecotton:
The fangs on the L1 Fire Ant's fire-ant.png look a bit too hefty for an L1 when looking at it with the help browser's upscaling algorithm, but seem fine when looking at the image itself. :/

data/core/units/monsters/Ant_Firebomb.cfg
```diff
@@ -0,0 +1,109 @@
...
+    name= _ "Firebomb Ant"
+    race=monster
+#ifndef __WMLUNITS__
+    image="{STATIC_IMAGE_TEMP}"
+#else
+    # Shown on units.wesnoth.org instead,
+    # because Image Path Functions aren't available there.
+    image="{PATH_TEMP}firebag-ant.png"
+#endif
+    profile="portraits/monsters/firebag_ant.webp"
```
stevecotton, line 20: There's a firebomb_ant.webp, but not this.

doofus-01: That was the first name, seemed too colloquial.

doofus-01:
> Is it possible to make the Firebomb Ant into a bomb? Like maybe some special/ability that makes the ant explode if critically wounded (HP lowered to 30%)?

Sounds good to me.  It's either an attack that's only active when critical, or it just blows up when killed.  I think the second option might be more interesting.  Or annoying... it will probably need some testing.

CelticMinstrel:
I think the second option would actually be really simple to implement (just an event with a `[harm_unit]` tag; making it AoE would even work if you wanted to). For the first option… I'm not quite sure how it would be done, but it would mean a player in control of the ant can choose not to use the ability, which would probably be nice in the niche case where a player is in control of the ant.

doofus-01:
The two things left to do here probably shouldn't be squashed with the rest, so I'll merge this and open new PRs for

- Generalize "falcon" to "avians" and add raven
- Explode ability for ant

Jonathan-Kelly:
@doofus-01
> Sounds good to me. It's either an attack that's only active when critical, or it just blows up when killed.

If you want the blast to harm all nearby units you could use [this weapon ability as a starting point](https://github.com/wesnoth/wesnoth/blob/master/data/campaigns/Winds_of_Fate/utils/abilities.cfg). It centers the blast around the attack target's location; if you want to center the blast on the ant's location then swap the `x,y=$x1,$y1` and `x,y=$x2,$y2` inside the `[filter_location]` blocks.

---

@CelticMinstrel
> I think the second option would actually be really simple to implement (just an event with a `[harm_unit]` tag; making it AoE would even work if you wanted to).

For the AoE, how would you go about taking into account the terrain defense of each unit inside the blast area? So a dwarf in the hills would have only a 40% chance of getting hit by shrapnel while a wose in an open field would have an 80% change of taking damage?

It seems like this is a missing feature of `[harm_unit]` - the option to roll a chance to hit for each unit it is applied to.

CelticMinstrel:
That's a good question. Why not open a feature request?

doofus-01:
I just went with affecting adjacent units in #7249.  There could probably be radial fading or other adjustments, but better to keep it simple.

spixi:
Maybe the Giant Ant line could be worked into a tree (advance to Fire or Soldier Ant at 20 XP).
9 gold is too cheap for the Fire Ant. 13 or 14 would be better.
I wonder if it would be useful to have a Sand Scuttling and Rock Scorpling (analogous to the Giant Scorpling).

CelticMinstrel:
Or Scorpling could advance to both types of scorpion?

spixi:
> Or Scorpling could advance to both types of scorpion?

The different types of scorpions have completely different habitats and movetypes. So I am unsure about this.

I also noted that the Sand Scuttler sprites are just recolored versions of the Giant Scorpling and some of them are also unused. Is that something we could also solve with a IPF or is the current Sand Scuttler artwork supposed to be a placeholder until someone out of the community comes up with alternative sprites?

knyghtmare:
> I also noted that the Sand Scuttler sprites are just recolored versions of the Giant Scorpion and some of them are also unused. Is that something we could also solve with an IPF or is the current Sand Scuttler artwork supposed to be a placeholder until someone out of the community comes up with alternative sprites?

Art contributors are currently in decline AFAIC.
So, making requests for new alternates to units is not exactly doable in the short run.
I mean making Giant Scorpling to advance to each of the other level 1 scorpions can fit.
Like adding geographical adaptation for survivability.

doofus-01:
The sand scuttler was a cheapo unit created to preserve campaign/scenario balance after someone realized the scorpion movetype (and resistances) made no sense and should be changed.  I noticed it has a cheapo zombie variation now, so why not a scorpling as well?  IPF would probably be the best way to handle it in that spirit.

Maybe there is some sort of data container like a trait that could handle it, in the cases of the scorpling & zombie.

spixi:
> The sand scuttler was a cheapo unit created to preserve campaign/scenario balance after someone realized the scorpion movetype (and resistances) made no sense and should be changed. I noticed it has a cheapo zombie variation now, so why not a scorpling as well? IPF would probably be the best way to handle it in that spirit.

That is exactly, what I did with #7332. I could have used a macro for the IPF function (like in [data/campaigns/Sceptre_of_Fire/units/Red_Wyrmlet.cfg](https://github.com/wesnoth/wesnoth/blob/9bfd432abef371c6c20d526fe05755fe310ba50d/data/campaigns/Sceptre_of_Fire/units/Red_Wyrmlet.cfg)). The Sand Scuttler uses separate sprites, which are just recolors of the Giant Scorpion.

I also noted that the portraits of the scorpions are shifted by different positions, which is most notable at u.w.o or in the Help view. The Level 0 variants have an unneccessary large canvas.

I can try to fix these minor issues tonight after work.

doofus-01:
The IPF work you did makes sense.  The portraits were shifted intentionally, so browsing the help didn't makes it quite as clear how cheap the portrait variations were.  Cheap is cheap though, so maybe the shifts should just be removed after all.

PR #7248, opened 2022-12-31
LoW: Make gold carry-over consistent with objectives

### LINKED ISSUE #7230: Legend of Wesmere S11 and S14 gold carry-over inconsistency
### Language

All

### Source

Battle of the Book, Campaign Legends of Wesmere

### Original Text

40% carried gold

### Proposed Change

80% carried gold

Note: If it's not a typo, and it's an error in the mission objectives and it really should be 40%, then change the objectives

Wedge009:
Resolves #7230

### DIFF
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter3/11_Battle_of_the_Book.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter3/11_Battle_of_the_Book.cfg
@@ -284,6 +284,7 @@
 #else
             bonus=yes
 #endif
+            carryover_percentage=40
         [/endlevel]
     [/event]
 
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter4/14_Human_Alliance.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter4/14_Human_Alliance.cfg
@@ -56,10 +56,10 @@ Chapter Four"
     [/story]
 
 #define PLAYER_GOLD
-    {GOLD 550 470 400}
+    {GOLD 600 500 400}
 #enddef
 #define MULTIPLAYER_GOLD
-    {GOLD 160 140 120}
+    {GOLD 200 175 150}
 #enddef
 
     [side]
--- a/data/campaigns/Legend_of_Wesmere/scenarios/chapter4/14_Human_Alliance.cfg
+++ b/data/campaigns/Legend_of_Wesmere/scenarios/chapter4/14_Human_Alliance.cfg
@@ -586,6 +586,7 @@ Chapter Four"
             result=victory
             bonus=no
             save=no
+            carryover_percentage=40
         [/endlevel]
     [/event]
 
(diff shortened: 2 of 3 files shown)


data/campaigns/Legend_of_Wesmere/scenarios/chapter3/11_Battle_of_the_Book.cfg
```diff
@@ -284,6 +284,7 @@
 #else
             bonus=yes
 #endif
+            carryover_percentage=40
```
knyghtmare, line 287: I think the carryover here should be 80 and not 40. Do you even know what the next scenario is?
2 cutscenes and then the large-scale orc vs elf & human battle. I dont think players would even approve of this...

data/campaigns/Legend_of_Wesmere/scenarios/chapter4/14_Human_Alliance.cfg
```diff
@@ -586,6 +586,7 @@ Chapter Four"
             result=victory
             bonus=no
             save=no
+            carryover_percentage=40
```
knyghtmare, line 589: I really think the balance adjustments should be playtested to ensure it's at least passable.

knyghtmare (changes requested):

data/campaigns/Legend_of_Wesmere/scenarios/chapter3/11_Battle_of_the_Book.cfg
Wedge009: 80 is the default. Yes, I saw that the next scenario is Human Alliance. The question is what the intention is for converting all the campaigns - including this one - to use 40% by default. Why make it 80 if all the others are 40?

data/campaigns/Legend_of_Wesmere/scenarios/chapter4/14_Human_Alliance.cfg
Wedge009: Shouldn't that sort of thing have been done when everything was converted to 40% in the first place?

Wedge009:
A reminder: this change isn't intended to be a 'balance' change. It's intended to make things consistent for what is already there, as highlighted by #7230.

The conversion to 40% was done by @nemaara in dbaeeea57e49d8f1fe07d3e9e149dca4e119141e, present since 1.15.0. It appears starting gold was already adjusted to accommodate the change in carry-over. So, to me, it looks like where carry-over remained at the default 80%, it's actually an oversight.

stevecotton:
If the bug was present when it was playtested (which I assume it was), then it seems necessary to assume that @nemaara tested it with the bug present, and balanced based on getting there with 80% gold.

Wedge009:
Hence why mentioned @nemaara, for input.

nemaara:
I can't remember why I missed this one but it'll need playtesting to fix the values. We can't blindly make it 40% without changing the gold values in the next scenario.

Wedge009:
Who is qualified for play-testing, then? I don't consider myself a suitable candidate. In the case that 80%->40% proves to be too difficult, is it just a matter of increasing the starting gold for the subsequent scenarios? I observe this was already done in the original conversion to 40%.

Wedge009:
I just came back from a short break - I'll update with your suggestions but will leave it open for now.

nemaara:
Can we leave this open for a week to see if more comments come in? After a week or so we can consider merging.

Wedge009:
I didn't set a specific time-frame, but yes, I left it open for further input.

Wedge009:
Has been a bit more than a week now and 1.17.12 has been tagged already. Are there any outstanding objections to this?

nemaara:
Let's say give 2 more days then merge?

stevecotton:
Should this be backported?

Wedge009:
See above, since this became a balance change I suggested not to back-port.

PR #7249, opened 2022-12-31
1.17 - new explode ability, given to firebomb ant

doofus-01:
The ant's animation is a bit half-assed, and the value probably needs adjusting, but the general idea is there.


data/core/_main.cfg
```diff
@@ -7,6 +7,7 @@
     code=<<
 wesnoth.dofile 'lua/wml-tags.lua'
 wesnoth.dofile 'lua/feeding.lua'
+wesnoth.dofile 'lua/explode_ability.lua'
```
CelticMinstrel, line 10: This file seems to be missing?

data/core/macros/abilities.cfg
```diff
@@ -388,6 +388,21 @@ Enemy units cannot see this unit while it is in deep water, except if they have
     [/dummy]
 #enddef
 
+#define ABILITY_EXPLODE TYPE VALUE
+    # Canned definition of the Explode ability to be included in an
+    # [abilities] clause.
+    # This works because of data/lua/explode_ability.lua
+    [dummy]
```
CelticMinstrel, line 395: I think it might be a good idea to use a different tag name, especially since it has some custom contents (`value` and `apply_to`).

data/core/macros/abilities.cfg
```diff
@@ -388,6 +388,21 @@ Enemy units cannot see this unit while it is in deep water, except if they have
...
+        id=explode
+        name= _ "explode"
+        female_name= _ "female^explode"
+        description=_ "This unit explodes upon death, causing damage to any adjacent units."
+        value={VALUE}
+        apply_to={TYPE}
+        special_note={INTERNAL:SPECIAL_NOTES_EXPLODE}
```
CelticMinstrel, line 402: As I recall, the `INTERNAL:SPECIAL_NOTES_*` macros only exist for backwards compatibility so that the strings can be used in two different places, so I think it would be fine to include the string directly right here.

CelticMinstrel (changes requested):
Looks good as far as I can see, but I'll still need to review the Lua portion of this.

I do have a few suggestions, but they're up for discussion. This is only marked as "request changes" because it's incomplete.

data/core/_main.cfg
doofus-01: Yeah...  it was.

data/core/macros/abilities.cfg
doofus-01: That's part of how to fix the schema complaints, I guess?

Pentarctagon:
Does the explosion damage the unit that attacked for both ranged and melee?

doofus-01:
It hurts everyone standing next to it.

data/lua/explode_ability.lua
```diff
@@ -0,0 +1,38 @@
...
+
+	if not u_exploder or not u_exploder:matches { ability = "explode" } then
+		return
+	end
+
+	local u_exploder_cfg = u_exploder.__cfg
+        local adjacent_units = wesnoth.units.find_on_map{T.filter_adjacent { id = u_exploder_cfg.id}}
+
+	for i,v in ipairs(wml.get_child(u_exploder_cfg, "abilities")) do
+		if v[1] == "dummy" and v[2].id == "explode" then
```
CelticMinstrel, line 23: It looks like you forgot to update this check. Both strings should read "explode".

Also, referring to the tag name as `v[1]` and its content as `v[2]` should be considered deprecated at least for mainline code. Prefer referring to `v.tag` and `v.contents` instead.

data/lua/explode_ability.lua
```diff
@@ -0,0 +1,38 @@
...
+			local dmg = v[2].value
```
CelticMinstrel, line 24: In the schema, you declared this value as an `f_int`. This means that the user can specify an integer, _or_ they can specify a formula enclosed in parentheses, for example `(unit.hitpoints / 2 + other_unit.hitpoints / 3)`.

If this is indeed the intended functionality, then you need to handle it as such. It would look something like this (adapted from similar code in random_placement.lua):

```suggestion
			local dmg = v.contents.value
			if type(dmg) == 'string' then
				if dmg:match('^%s*%(.*%)%s*$') then
					local params = {
						unit = u_exploder,
						other_unit = wesnoth.units.get(ec.x2, ec.y2)
					}
					local result = wesnoth.eval_formula(dmg, params)
					-- Now to convert it to an integer
					-- Pick only one of the following 2 lines
					dmg = math.floor(tonumber(result))
					--dmg = math.round(tonumber(result))
			else
				wml.error('invalid formula')
			end
		end
```

Alternatively, if you don't want to allow that, you can change the schema type to just `int`. This makes it inconsistent with built-in abilities, though.

data/schema/units/abilities.cfg
```diff
@@ -126,6 +126,13 @@
 		{DEFAULT_KEY pass_allied_units bool yes}
 	[/tag]
 [/tag]
+[tag]
+	name="explode"
+	max=infinite
+	super="units/unit_type/abilities/heals"
```
CelticMinstrel, line 132: Should be:

```suggestion
	super="units/unit_type/abilities/~generic~"
```

data/lua/explode_ability.lua
```diff
@@ -0,0 +1,38 @@
...
+	local u_exploder = wesnoth.units.get(ec.x1, ec.y1)
+
+	if not u_exploder or not u_exploder:matches { ability = "explode" } then
+		return
+	end
+
+	local u_exploder_cfg = u_exploder.__cfg
+        local adjacent_units = wesnoth.units.find_on_map{T.filter_adjacent { id = u_exploder_cfg.id}}
+
+	for i,v in ipairs(wml.get_child(u_exploder_cfg, "abilities")) do
```
CelticMinstrel, line 22: I think it's better to get rid of the extra variable here:
```suggestion
        local adjacent_units = wesnoth.units.find_on_map{T.filter_adjacent { id = u_exploder.id}}

	for i,v in ipairs(wml.get_child(u_exploder.__cfg, "abilities")) do
```

Also, it looks like you need to run the whitespace fixing script. If you're not on Windows, you can just run the script `./utils/CI/fix_whitespace.sh`. I'm not sure what options there are if you _are_ on Windows, other than running that in `git bash` or something.

CelticMinstrel (commented):
The implementation mostly looks good. I've made a few more suggestions.

doofus-01:
The schema tripped me up and I was just trying to get through the sideshow, but didn't test it enough afterwards.  I'll fix this next year.  Thanks.

data/lua/explode_ability.lua
```diff
@@ -0,0 +1,38 @@
...
+wesnoth.game_events.add_repeating("die", function()
+	local ec = wesnoth.current.event_context
+
+	if not ec.x1 or not ec.y1 then
+		return
+	end
+
+	local u_exploder = wesnoth.units.get(ec.x1, ec.y1)
+
+	if not u_exploder or not u_exploder:matches { ability = "explode" } then
```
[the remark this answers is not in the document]
doofus-01: I'm trying to think of a use case where you'd want to trigger this ability in a way that uses different keys and can stack with the existing ability, but can't come up with anything.  Am I missing something?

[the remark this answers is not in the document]
CelticMinstrel: I guess my main reasoning is that the way you've defined it makes it more like a class of abilities rather than a single ability, and for built-in abilities, that's what the tag name indicates.

data/core/macros/abilities.cfg
CelticMinstrel: Indeed, having a unique tag name certainly helps with making it validate in the schema.

CelticMinstrel:
Why does this pass CI when the whitespace in explode_ability.lua is all over the place?

Besides that, it looks like there's only one comment unaddressed.

data/core/macros/abilities.cfg
doofus-01: Or keep the macros and use it in both description and special_note, as they seem to mostly say the same thing.

doofus-01:
My whitespacing is coming from joe's lua formatting, which must be different from whatever formatting your script is enforcing, and the differences are invisible to me.  Must be related to tabs.

data/core/macros/abilities.cfg
CelticMinstrel: I will note that the special not is optional, so you could omit it if you think it isn't something that needs to be mentioned in the unit description (although, exploding _does_ seem notable to me…).

If it does make sense to use the exact same string for description and special note, then yeah, keeping the macro makes sense, but I'm not sure the current name of the macro suits that use…

CelticMinstrel:
> Must be related to tabs.

I think some lines are indented with tabs and some are indented with spaces…

CelticMinstrel:
> > Must be related to tabs.
> 
> I think some lines are indented with tabs and some are indented with spaces… maybe try either "convert tabs to spaces" or "convert spaces to tabs" in your editor?

stevecotton:
> Why does this pass CI when the whitespace in explode_ability.lua is all over the place?

Because the indentation check only runs on WML files, not .lua ones.

CelticMinstrel:
> Because the indentation check only runs on WML files, not .lua ones.

That doesn't sound accurate. It runs _utils/CI/fix_whitespace.sh_, which in addition to calling wmlindent does something with Lua and C++ files.

Pentarctagon:
It just removes trailing whitespace.

CelticMinstrel:
Ah, somehow I thought it also converted indentation. Okay.

Well, in that case, the Lua file will need to be manually re-indented.

data/lua/explode_ability.lua
```diff
@@ -0,0 +1,56 @@
+
+local _ = wesnoth.textdomain 'wesnoth-help'
+local T = wml.tag
+
+-- The explode event code
+wesnoth.game_events.add_repeating("die", function()
```
CelticMinstrel, line 6: I think this should use `game_events.add` instead of `add_repeating`. Then there will be no need for those early returns, because the function will only be called if the filters pass.

doofus-01: I don't understand the advantage, can you explain?

CelticMinstrel: `add_repeating` adds an event with no filters – your code will be called every time any unit dies, even if it's not an ant. With `add` you can also specify filters – the code will only be called when an ant dies (or other unit with the ability).

Functionally, I don't _think_ there's much difference, since it _is_ a repeating event. I think it's stylistically better, though.

data/lua/explode_ability.lua
```diff
@@ -0,0 +1,56 @@
...
+    local ec = wesnoth.current.event_context
+
+    if not ec.x1 or not ec.y1 then
+        return
+    end
+
+    local u_exploder = wesnoth.units.get(ec.x1, ec.y1)
+
+    if not u_exploder or not u_exploder:matches { ability_type = "explode" } then
```
gfgtdf, line 15: I think this should be abulity_type_active,

data/core/macros/abilities.cfg
```diff
@@ -387,3 +387,18 @@ Enemy units cannot see this unit while it is in deep water, except if they have
...
+    # Canned definition of the Explode ability to be included in an
+    # [abilities] clause.
+    # This works because of data/lua/explode_ability.lua
+    [explode]
+        id=explode
+        name= _ "explode"
+        female_name= _ "female^explode"
+        description=_ "This unit explodes upon death, causing damage to any adjacent units."
+        value={VALUE}
+        apply_to={TYPE}
```
gfgtdf, line 401: since apply_to already has a meaning in weapon specials (self/opponent) i this k it might be better if this key had a different name.

data/lua/explode_ability.lua
CelticMinstrel: I think it _should_ be `ability_type_active` but actually _can't_ be because we have no way in Lua to open a weapon specials context.

We probably need something that returns a closable table so you can use it like this:

```lua
local ctx <close> = wesnoth.units.weapons_context(whatever.parameters)
if not u_exploder or not u_exploder:matches { ability_type_active = "explode" } then
```

I'm not sure if that's the best name for it, but hopefully you get the idea.

data/core/macros/abilities.cfg
CelticMinstrel: That's a good point… why not just call it `damage_type`?

data/lua/explode_ability.lua
gfgtdf: I think the context should just be set during die events just like it's set during attack events.


But that wasn't my point, my point is that if we want to do that, the code below which reads damage type and damage amount should also ignore inactive abilities, and there is currently no way to do that.

CelticMinstrel: I think we need the context exposed to Lua regardless – even if it's not useful here, then it's useful in AI. But from what you're saying it seems like it should be useful here as well.

data/core/macros/abilities.cfg
doofus-01: OK, so the engine will change.  Master has developed since this was last touched and I hope to make the animation better anyway, so I'll close this PR.  I'll try again when the requirements are more firm.

doofus-01:
I will re-submit with something more up-to-date when the time is right.

PR #7252, opened 2023-01-01
1.17 generalize the birds race

doofus-01:
Changing the "falcon" race to "bird" is the biggest change, might need some sort of compatibility code.

As far as new units, the Raven is WIP.  Maybe there could be an ostrich/emu for something related to the Chocobone, or maybe that's just best left to fables and history.


CelticMinstrel:
Hmm, I wonder if maybe we should keep the ID as `falcon` instead of `bird` for now, to preserve backwards compatibility? After all, we don't really have a deprecation path for races right now…

It seems like it wouldn't be the first time a race name was changed without changing the ID – look at the saurians, whose ID is `lizard`. That said, the chance of someone adding a new falcon is comparatively slim, so if we can prove no add-ons on the 1.16 server did that… but even then it could break filters…

Yeah, I think it's probably best to leave the ID unchanged.

knyghtmare:
[quote]
, so if we can prove no add-ons on the 1.16 server did that… but even then it could break filters…
[/quote]

Hmm, I suspect one particular addon might break "Nameless Fellowship Saga" since it has a scenario where a hero unit has to kill a Falcon or Elder Falcon. So, that might need to be checked. 

If cleared, I think swapping race to generalised "birds" is okay. Also, maybe just leave in Falcon as a sub-race of birds with `help_taxonomy=`? I think that'd work

CelticMinstrel:
I'm simply saying that it's best _not_ to change the part that a) no player ever sees and b) will probably break add-ons.

Not breaking add-ons is generally the preferred state.

doofus-01:
> Not breaking add-ons is generally the preferred state.

Sure, but if the preferred method of doing that is "Don't touch that if it has been used ever", we end up with sludge.  Much better to change what needs to be changed and have some compatibility method.

Now, is falcons -> birds a needed change?  I thought it was, but it's also not that important.

CelticMinstrel:
Yeah, between "leave the ID unchanged forever" and "change the ID but have some compatibility method", I prefer the latter. But we'd have to build that method first – we don't have anything that works for races. Although some conversion function could probably be added to wmllint, that doesn't seem sufficient in this case.

doofus-01:
What exactly is being broken?  I don't see a race=falcon filter anywhere in that addon.  Even if there is something somewhere, we can certainly have a compatibility method.  Aren't you the guy who made 1,001 deprecations last cycle?

Leaving weird IDs scattered around makes it harder to do anything - each instance isn't all that big a barrier but if the IDs are a magic password instead of something that makes sense, it is sludge.

CelticMinstrel:
> Aren't you the guy who made 1,001 deprecations last cycle?

Every one of those deprecations does not break compatibility with older add-ons (unless I messed up on one or two of them, which would be a bug).

The on the other hand _does_ potentially break compatibility. An add-on that filtered for falcons would cease to function.

The compatibility path you added seems like it should work, though it might be nice to have a deprecation message on the old units. I'm not sure if it's the best approach, as it renames the _real_ units, but I guess I can accept it.

stevecotton:
I'm wondering why the "falcons" > "birds" change is needed. I'm not arguing against it, just wanting a pointer to the background info that should be read before joining the discussion.

Possibly compatibility paths to eventually (as in circa Wesnoth 1.22) end up without the sludge:
* Ask add-on developers to use `race=bird,falcon` in their SUFs (as done to UtBS's scenario files in 3bf3717e64460881a3f4c9516791391b93a33543)
* Add `[SUF]taxonomy=` or `[SUF]race_tree=` in 1.17, and ask add-on developers to use `=bird` instead of `race=falcon`.

data/core/units/monsters/Raven.cfg
```diff
@@ -0,0 +1,118 @@
...
+    {TRAIT_FERAL_MUSTHAVE}
+    hitpoints=18
+    movement_type=lightfly
+    movement=7
+    experience=24
+    level=0
+    alignment=neutral
+    advances_to=null
+    {AMLA_DEFAULT}
+    undead_variation=falcon
```
stevecotton, line 17: The Walking Corpse and Soulless variations should become birds too.

CelticMinstrel: Well, if your point is about consistency of IDs, yes. However, note that this does _not_ reference the race ID, so it's not strictly necessary to change it. At the same time, it quite likely has similar compatibility issues to change it.

doofus-01:
> Every one of those deprecations does not break compatibility with older add-ons (unless I messed up on one or two of them, which would be a bug).
The on the other hand does potentially break compatibility. An add-on that filtered for falcons would cease to function.

Weak.  You get to make significant deprecations when you want, but then suddenly it's something to be nipped in the bud if someone else does it and you notice.  This is hardly the first time an ID has been changed, it can be figured out.  If it's not executed well enough, that "would be a bug".

> In either case, the change from "falcon" to "bird" would get postponed until all the UMCs were ready.

Of course, with a deprecation warning somewhere, and moving mainline/core to the end-state as an example and influence on the future.

> The Walking Corpse and Soulless variations should become birds too.

It's a good point, but that's a whole other can of worms.  It will be simpler to solve once the living birds are finished.

-------
My frustration is that I don't want to waste precious time fighting over taxonomy of falcons, but I also don't want to say ravens or terror birds or owls or whatever are "falcons".  I'd rather work on useful content, and then if there are objections with how it's ultimately implemented, we can bicker over it at that time, instead of pretending nothing has ever been changed during a development cycle before.

Pentarctagon:
> Weak. You get to make significant deprecations when you want, but then suddenly it's something to be nipped in the bud if someone else does it and you notice. This is hardly the first time an ID has been changed, it can be figured out. If it's not executed well enough, that "would be a bug".

For renaming the Dunefolk, that was handled via https://github.com/wesnoth/wesnoth/blob/1.14/data/core/units/dunefolk/deprecated.cfg for a time, so I that approach could be used here as well.

If there's no way to effectively deprecate a race id, then another approach would be to have each "bird" given its own race. So have a separate Raven race, Owl race, etc.

CelticMinstrel:
It does look like doofus came up with something that should work for deprecating a race. We should consider distilling it down into a macro or two and shoving it into deprecation-utils.cfg.

sevu:
One can leave the old `[race]` tag, but not use for a unit type.
It will be hidden from the help unless an add-on defines a unit of this race.

doofus-01:
I don't have the time/energy/focus necessary to push this to completion, so I'll close it.  If anyone thinks there is anything useful here, it can always be linked to in an Issue.

PR #7265, opened 2023-01-06
Rename database column SOURCE to ADDON_ID.

Pentarctagon:


soliton-:
Should it also be `ADDON_VERSION`? Since the table is otherwise about content not the addon itself so it might help to make it clear that those 2 columns are related and about the addon instead.

soliton- (approved):

Pentarctagon:
Updated.

The SQL to update the database then being:
`alter table wesnothd_game_content_info change column SOURCE ADDON_ID VARCHAR(100) NOT NULL;`
`alter table wesnothd_game_content_info change column VERSION ADDON_VERSION VARCHAR(100) NOT NULL;`

PR #7266, opened 2023-01-08
Orb color preferences dialog layout and tooltip

stevecotton:
Add some spacing to make it clearer what is grouped with what, move the disengage orb's setting to be part of the partial orb's group, and add tooltip.

There's only one tooltip so far, does anyone feel the "Show ... orb" checkboxes should have them too?

This dialog uses checkbox labels as section headers to indicate what the rows of colors refer to. This commit doesn't change that, but it makes the grouping clearer. Actual section headers would take additional space and probably not look any better.

The intention would be to backport it, and then backport a rebased PR #7189 on top of it. The image below shows the intended final destination, the current PR doesn't include the "During ally's turn, use a two-colored orb with status" line.
image

### DIFF
--- a/changelog_entries/orb_prefs_layout.md
+++ b/changelog_entries/orb_prefs_layout.md
@@ -1,0 +1,2 @@
+ ### User interface
+   * Improved the layout of the orb colors dialog, and added tooltips.
--- a/data/gui/window/select_orb_colors.cfg
+++ b/data/gui/window/select_orb_colors.cfg
@@ -1,8 +1,19 @@
 #textdomain wesnoth-lib
 
+#define _GUI_ORB_SPACER_ROW
+	[row]
+		grow_factor = 0
+		[column]
+			[spacer]
+				height = 15
+			[/spacer]
+		[/column]
+	[/row]
+#enddef
+
 #define _GUI_ORB_CELL PURPOSE COLOR
 	[column]
-		border = "all"
+		border = "left,right,bottom"
 		border_size = 5
 		[toggle_button]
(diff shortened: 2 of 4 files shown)


Pentarctagon:
In general I'd say that tooltips are good to have, unless it's already extremely obvious what it means (subjective, I know).

A "before" screenshot would also be good to have here as a comparison.

stevecotton:
Before image (ignore the indented one checkbox, that's not in current master either): https://user-images.githubusercontent.com/101462/210826131-f73ffa7c-3def-42a6-a49b-62c7b5c944c0.png

Agreed about the "in general", but I'm not seeing a way for tooltips to add readability to "show unmoved orb". "When one of you units hasn't been moved, this orb is shown on them".  "You've seen these orbs in game, well this turns them on and off in the places that you've seen them".

Pentarctagon:
Definitely looks better with the grouping.

What does
> The image below shows the intended final destination, the current PR doesn't include the "During ally's turn, use a two-colored orb with status" line.

mean?

stevecotton:
> What does
> 
> > The image below shows the intended final destination, the current PR doesn't include the "During ally's turn, use a two-colored orb with status" line.
> 
> mean?

The screenshot was taken before splitting this out from the other PR. Please imagine the "Show ally orb" group has only the twe lines:
* "Show ally orb" 
* row of color buttons

stevecotton:
From Vultraz on Discord: "I don’t really like it but I can’t think of how to make it better :grimacing:".

Going to merge as-is to unblock the next PR.

PR #7268, opened 2023-01-09
Add ability to allow secondary authors to upload an add-on.

Pentarctagon:
The secondary_authors attribute is a comma-delimited list of forum accounts that can also upload updates to an add-on. Secondary authors can't change the primary/secondary authors and can't delete the add-on.

The primary author can also make someone else the primary author by:
* Putting their username in the secondary_authors attribute
* Putting someone else's username in the author attribute
* Uploading the add-on while selecting themselves

This works since they are still (until the upload is done) the primary author in the database, so they are allowed to change the authors list, including in this case.

---

Scenarios tested locally:
* upload as primary author, no secondary authors - **succeeds**, only primary author row exists
* upload as primary author, has secondary authors - **succeeds**, primary author row and secondary author rows exists
* upload as primary author, adds a new secondary author and removes another - **succeeds**, table reflects this change
* upload as primary author with changing primary author - **succeeds**, table reflects primary author and secondary author swapped
* upload as secondary author with changing primary author - **succeeds**, addon_authors is unchanged
* upload with non-existing primary author - **fails**
* upload with non-existing secondary author - **fails**
* upload with author as an existing user that is not the primary author - **fails**
* upload with author as an existing user that is not a secondary author - **fails**
* delete addon as secondary author - **fails**
* delete addon as primary author - **succeeds**, author rows deleted
* delete addon as non-existing primary author - **fails**
* delete addon as non-existing secondary author - **fails**
* delete by author as an existing user that is not the primary - **fails**
* delete by author as an existing user that is not a secondary author - **fails**
* delete by former secondary author after changing them to primary author - **succeeds**

### DIFF
--- a/src/server/campaignd/server.cpp
+++ b/src/server/campaignd/server.cpp
@@ -1317,12 +1317,19 @@ ADDON_CHECK_STATUS server::validate_addon(const server::request& req, config*& e
 			LOG_CS << "Validation error: client requested forum authentication but server does not support it";
 			return ADDON_CHECK_STATUS::SERVER_FORUM_AUTH_DISABLED;
 		} else {
-			if(!user_handler_->user_exists(upload["author"].str())) {
+			if(!user_handler_->user_exists(upload["uploader"].str())) {
 				LOG_CS << "Validation error: forum auth requested for an author who doesn't exist";
 				return ADDON_CHECK_STATUS::USER_DOES_NOT_EXIST;
 			}
 
-			if(!authenticate_forum(upload, upload["passphrase"].str())) {
+			for(const std::string& secondary_author : utils::split(upload["secondary_authors"].str(), ',')) {
+				if(!user_handler_->user_exists(secondary_author)) {
+					LOG_CS << "Validation error: forum auth requested for an author who doesn't exist";
+					return ADDON_CHECK_STATUS::USER_DOES_NOT_EXIST;
+				}
+			}
+
+			if(!authenticate_forum(upload, upload["passphrase"].str(), false)) {
 				LOG_CS << "Validation error: forum passphrase does not match";
 				return ADDON_CHECK_STATUS::UNAUTHORIZED;
 			}
@@ -1499,7 +1506,7 @@ void server::handle_upload(const server::request& req)
 	// Write general metadata attributes
 
 	addon.copy_or_remove_attributes(upload,
-		"title", "name", "author", "description", "version", "icon",
+		"title", "name", "uploader", "author", "secondary_authors", "description", "version", "icon",
 		"translate", "dependencies", "core", "type", "tags", "email", "forum_auth"
 	);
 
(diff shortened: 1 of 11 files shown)


CelticMinstrel:
I can't really tell from the code, but this seems to be adding a key to the pbl file? So that means if it merges before #7239, I'll have to update the schema in that PR, and if it merges after, you'll have to do that in this PR…

Pentarctagon:
It adds `secondary_authors`, so that's correct.

src/server/campaignd/server.cpp
```diff
@@ -1891,14 +1916,28 @@ void server::handle_change_passphrase(const server::request& req)
...
+	std::string uploader = addon["uploader"].str();
+	std::string id = addon["name"].str();
+	bool do_authors_exist = user_handler_->db_do_any_authors_exist(server_id_, id);
+	bool is_primary = user_handler_->db_is_user_primary_author(server_id_, id, uploader);
+	bool is_secondary = user_handler_->db_is_user_secondary_author(server_id_, id, uploader);
+
+	// allow if there is no author information - this is a new upload
+	// don't allow other people to upload if author information does exist
+	// don't allow secondary authors to remove the add-on from the server
+	if((do_authors_exist && !is_primary && !is_secondary) || (is_secondary && is_delete)) {
```
soliton-, line 1933: Is it guaranteed that if there is no author info that this must be a new upload?

src/server/common/dbconn.cpp
```diff
@@ -459,6 +460,56 @@ void dbconn::update_addon_download_count(const std::string& instance_version, co
...
+bool dbconn::is_user_author(const std::string& instance_version, const std::string& id, const std::string& username, int is_primary) {
+	try
+	{
+		return exists(connection_, "SELECT 1 FROM `"+db_addon_authors_table_+"` WHERE INSTANCE_VERSION = ? AND ADDON_ID = ? AND AUTHOR = ? AND IS_PRIMARY = ?",
+			instance_version, id, username, is_primary);
+	}
+	catch(const mariadb::exception::base& e)
+	{
+		log_sql_exception("Unable to check whether `"+username+"` is the primary author of "+id+" for version "+instance_version+".", e);
+		return true;
```
soliton-, line 472: Why return true here?

soliton- (changes requested):

src/server/common/dbconn.cpp
Pentarctagon: It shouldn't, good catch.

src/server/campaignd/server.cpp
Pentarctagon: Technically no, since it would also fail to find author info for existing add-ons using `forum_auth=yes` prior to this PR being merged. However, the only add-ons currently using that are test add-ons I uploaded previously, so it would not affect any real add-ons created by anyone else.

src/server/campaignd/server.cpp
```diff
@@ -1317,12 +1317,19 @@ ADDON_CHECK_STATUS server::validate_addon(const server::request& req, config*& e
...
-			if(!user_handler_->user_exists(upload["author"].str())) {
+			if(!user_handler_->user_exists(upload["uploader"].str())) {
 				LOG_CS << "Validation error: forum auth requested for an author who doesn't exist";
 				return ADDON_CHECK_STATUS::USER_DOES_NOT_EXIST;
 			}
 
-			if(!authenticate_forum(upload, upload["passphrase"].str())) {
+			for(const std::string& secondary_author : utils::split(upload["secondary_authors"].str(), ',')) {
+				if(!user_handler_->user_exists(secondary_author)) {
+					LOG_CS << "Validation error: forum auth requested for an author who doesn't exist";
```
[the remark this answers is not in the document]
Pentarctagon: Updated.

src/server/campaignd/server.cpp
```diff
@@ -1855,6 +1877,9 @@ void server::handle_delete(const server::request& req)
 		return;
 	}
 
+	if(addon["forum_auth"].to_bool()) {
+		user_handler_->db_delete_addon_authors(server_id_, addon["name"].str());
+	}
```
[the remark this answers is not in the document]
Pentarctagon: Updated.

src/server/common/dbconn.cpp
Pentarctagon: Updated.

src/server/campaignd/server.cpp
soliton-: I was rather wondering about uploads that did not use forum_auth so far but I see now that that is rejected earlier.

soliton- (approved):
Didn't check the GUI part but rest looks good to me.

PR #7271, opened 2023-01-10
change calculation rules so that add and sub are always used independently

newfrenchy83:
when two special weapons use multiply and divide with the same id, both operations are used, isn't that so why should it be different with 'add' and 'sub' where it's the larger value that is used (if asub=value_sub and add=value_add are used and value_sub>value_add then value_sub is used). This logic is counter-intuitive. that multiplication/division is applied to (base_value +- add/sub) is understandable but not this discrimination. For me add and sub should still be usable; even if it means changing the rules, but I think we will gain clarity in the end.

### DIFF
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_calculations.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_calculations.cfg
@@ -95,9 +95,9 @@
 # API(s) being tested: [attacks]sub=
 ##
 # Expected end state:
-# Bob's attack has 7 strikes. The difference to add2_add3_cumulative is that sub chooses the lower value, and the order of the abilities determines the order of the calculations.
+# Bob's attack has 10 strikes. Both the add and the negative sub are used.
 #####
-{TWO_CALCULATION_UNIT_TEST add2_sub_minus3_cumulative 5 (cumulative,add=yes,2) (cumulative,sub=yes,-3) 7}
+{TWO_CALCULATION_UNIT_TEST add2_sub_minus3_cumulative 5 (cumulative,add=yes,2) (cumulative,sub=yes,-3) 10}
 #####
 # API(s) being tested: [attacks]sub=,[attacks]cumulative=
 ##
@@ -130,25 +130,36 @@
 # API(s) being tested: [attacks]add=,[attacks]sub=
 ##
 # Expected end state:
-# Bob's attack has 2 strikes. When both add and sub are used in the same [attacks] tag, only the sub is used.
+# Bob's attack has 4 strikes. When both add and sub are used in the same [attacks] tag, both add and sub are used.
 #####
-{ONE_CALCULATION_UNIT_TEST add_sub_combined 5 (add,sub=2,3) 2}
+{ONE_CALCULATION_UNIT_TEST add_sub_combined 5 (add,sub=2,3) 4}
 #####
 # API(s) being tested: [attacks]add=,[attacks]sub=
 ##
 # Expected end state:
-# Bob's attack has 2 strikes. In separate [attacks] without separate ids, the order of the abilities determines which one is used.
+# Bob's attack has 4 strikes.
 #####
-{TWO_CALCULATION_UNIT_TEST add_sub_separated 5 (add=2) (sub=3) 2}
-{TWO_CALCULATION_UNIT_TEST sub_add_separated 5 (sub=2) (add=3) 8}
+{TWO_CALCULATION_UNIT_TEST add_sub_separated 5 (add=2) (sub=3) 4}
+# API(s) being tested: [attacks]add=,[attacks]sub=
+##
+# Expected end state:
+# Bob's attack has 6 strikes.
+#####
+{TWO_CALCULATION_UNIT_TEST sub_add_separated 5 (sub=2) (add=3) 6}
 #####
 # API(s) being tested: [attacks]cumulative=
 ##
 # Expected end state:
-# Bob's attack has 2 strikes. These tags don't specify ids, and cumulative doesn't affect add or sub, so the order of the abilities determines which one is used.
+# Bob's attack has 4 strikes. These tags don't specify ids, and cumulative doesn't affect add or sub.
 #####
-{TWO_CALCULATION_UNIT_TEST add_sub_cumulative 5 (cumulative,add=yes,2) (cumulative,sub=yes,3) 2}
-{TWO_CALCULATION_UNIT_TEST sub_add_cumulative 5 (cumulative,sub=yes,2) (cumulative,add=yes,3) 8}
+{TWO_CALCULATION_UNIT_TEST add_sub_cumulative 5 (cumulative,add=yes,2) (cumulative,sub=yes,3) 4}
+#####
+# API(s) being tested: [attacks]cumulative=
+##
+# Expected end state:
+# Bob's attack has 6 strikes. These tags don't specify ids, and cumulative doesn't affect add or sub.
+#####
+{TWO_CALCULATION_UNIT_TEST sub_add_cumulative 5 (cumulative,sub=yes,2) (cumulative,add=yes,3) 6}
 #####
 # API(s) being tested: [attacks]id=
 ##
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -2248,6 +2248,7 @@ effect::effect(const unit_ability_list& list, int def, const_attack_ptr att, EFF
 
 	int value_set = (wham == EFFECT_CUMULABLE) ? std::max(list.highest("value").first, 0) + std::min(list.lowest("value").first, 0) : def;
 	std::map<std::string,individual_effect> values_add;
+	std::map<std::string,individual_effect> values_sub;
 	std::map<std::string,individual_effect> values_mul;
 	std::map<std::string,individual_effect> values_div;
 
@@ -2311,9 +2312,9 @@ effect::effect(const unit_ability_list& list, int def, const_attack_ptr att, EFF
 				callable.add("base_value", wfl::variant(def));
 				return formula.evaluate(callable).as_int();
 			});
-			std::map<std::string,individual_effect>::iterator sub_effect = values_add.find(effect_id);
-			if(sub_effect == values_add.end() || sub < sub_effect->second.value) {
-				values_add[effect_id].set(ADD, sub, ability.ability_cfg, ability.teacher_loc);
+			std::map<std::string,individual_effect>::iterator sub_effect = values_sub.find(effect_id);
+			if(sub_effect == values_sub.end() || sub < sub_effect->second.value) {
+				values_sub[effect_id].set(ADD, sub, ability.ability_cfg, ability.teacher_loc);
 			}
 		}
 		if (const config::attribute_value *v = cfg.get("multiply")) {
@@ -2381,7 +2382,15 @@ effect::effect(const unit_ability_list& list, int def, const_attack_ptr att, EFF
 		effect_list_.push_back(val.second);
 	}
 
-	composite_value_ = static_cast<int>((value_set + addition) * multiplier / divisor);
+	/* Additional and subtraction are independent since Wesnoth 1.19.4. Prior to that, they affected each other.
+	 */
+	int substraction = 0;
+	for(const auto& val : values_sub) {
+		substraction += val.second.value;
+		effect_list_.push_back(val.second);
+	}
+
+	composite_value_ = static_cast<int>((value_set + addition + substraction) * multiplier / divisor);
 	//clamp what if min_value < max_value or one attribute only used.
 	if(max_value && min_value && *min_value < *max_value) {
 		composite_value_ = std::clamp(*min_value, *max_value, composite_value_);
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/attacks/attacks_add_sub.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/attacks/attacks_add_sub.cfg
@@ -8,7 +8,7 @@
 # Have the side 1 leader attack the side 2 leader with both its weapons
 ##
 # Expected end state:
-# The side 1 leader has 2 weapons each of which now has 3 strikes (5-2), add is ignored
+# The side 1 leader has 2 weapons each of which now has 4 strikes (5+1-2)
 #####
 {COMMON_KEEP_A_B_UNIT_TEST "attacks_add_sub" (
     [event]
@@ -27,5 +27,5 @@
         [/modify_unit]
     [/event]
 
-    {CHECK_STRIKES "3,3" ({SUCCEED})}
+    {CHECK_STRIKES "4,4" ({SUCCEED})}
 )}
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/chance_to_hit/chance_to_hit_add_sub.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/chance_to_hit/chance_to_hit_add_sub.cfg
@@ -4,12 +4,12 @@
 # API(s) being tested: [chance_to_hit]value=
 ##
 # Actions:
-# Give the leaders a chance_to_hit ability with both the value, add, and sub attributes, adding up to 10 but effectively zero because sub completely overrides the add attribute.
+# Give the leaders a chance_to_hit ability with both the value, add, and sub attributes.
 # Give each unit 1000 hp and 100 strikes
 # Have the side 1 leader attack the side 2 leader with both its weapons
 ##
 # Expected end state:
-# the add attributes is ignored
+# chance to hit is to 0%
 # The side 1 leader's first weapon strikes 0 times
 # The side 1 leader's second weapon strikes 0 times
 #####
@@ -27,7 +27,7 @@
                 [abilities]
                     {TEST_ABILITY attacks 100 () SELF=yes}
                     {TEST_ABILITY chance_to_hit 20 (add=10
-                    sub=20) SELF=yes}
+                    sub=30) SELF=yes}
                 [/abilities]
             [/effect]
         [/modify_unit]
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/damage/damage_add_sub.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/damage/damage_add_sub.cfg
@@ -8,7 +8,7 @@
 # Have the side 1 leader attack the side 2 leader with both its weapons
 ##
 # Expected end state:
-# The side 1 leader has 2 weapons each of which now has 3 damage (5-2), add is ignored
+# The side 1 leader has 2 weapons each of which now has 4 damage (5+1-2)
 #####
 {COMMON_KEEP_A_B_UNIT_TEST "damage_add_sub" (
     [event]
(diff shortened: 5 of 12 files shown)


stevecotton:
I'm not sure what we want to do with the 1.17 API, but I feel the current behavior in 1.16 is probably unexpected, and that we should add a warning (probably using the deprecation mechanism) to 1.16.

newfrenchy83:
> I'm not sure what we want to do with the 1.17 API, but I feel the current behavior in 1.16 is probably unexpected, and that we should add a warning (probably using the deprecation mechanism) to 1.16.

i don't think what deprecation mechanismwas relevant because it must show message what if add and sub used in same list, and with same id for don't have message no contextual.

CelticMinstrel:
Can someone explain what exactly this PR is doing? It has to do with add and sub are used in the same ability? Or only if it's two different abilities with the same ID? What exactly does it change?

newfrenchy83:
> Can someone explain what exactly this PR is doing? It has to do with add and sub are used in the same ability? Or only if it's two different abilities with the same ID? What exactly does it change?

Both, like show test modification. The change is what contrary multiply and divide what was usable with two ability with same id, it's not case with add/sub, this PR correct this bug(for me).

CelticMinstrel:
I think I vaguely get your goal, but I'm still not quite sure that the new behaviour is the correct behaviour. I don't think the old behaviour is correct either, though.

Pentarctagon:
Deprecate `sub` and tell people to multiply by -1.

(joking)

stevecotton:
My feeling is that we shouldn't do this, unless we find a bug where the old behavior can cause OOS and the new behavior fixes it. Both the old and new behaviors would need explanations, and so it's simpler to say "1.17 behaves the same as 1.16". Even though the new behavior is more logical than the old, there's more value in saying "UMC that was tested in 1.16 still works the same way".

Quick draft of some docs that should be on the Wiki, showing where changing the behavior would mean an extra thing to be aware of (note: just draft, I haven't checked that this is correct) :

This only applies to abilities that don't use `is_cumulable=true` in the C++ calculations.

The `cumulative=` key only affects `value=`. For `add=`, `sub=`, `multiply=` and `divide=`, having multiple instances of an ability with exactly the values for `id,add,sub,multiply,divide=` has the same effect as having one instance of that ability active. Having multiple instances with the same `id=` but different values for the other attributes can have surprising effects. It will use the largest values for `multiply=` and `divide=`, and either the largest `add=` or the most negative `sub=`. The last encountered `add=` or `sub=` determines which of `add=` or `sub=` is used, the other is ignored; (Since 1.17.11) both the largest `add=` and the most negative `sub=` are used.

Pentarctagon:
Given stevecotton's last comment, should this PR remain open?

newfrenchy83:
@Pentarctagon Given that when using both add and sub in an ability, only sub is considered, and since this PR allows their simultaneous use as one should intuitively expect, would it be possible to commit this PR at the last minute for 1.18 or is it already too late?

stevecotton:
The simple part of the answer is, it's not getting merged for 1.18.

However, the main answer, about when it became too late, is more complex. Although this PR would make the logic seem clearer in isolation, there's questions about how it interacts with filters, and how multiple abilities interact if `cumulative` says to only apply the largest value. Filters for `add=` have the feature of automatically checking for a `sub=` attribute if the ability doesn't have an `add=`. If abilities can usefully have both `add=` and `sub=`, should those filters try to match `add=`, `sub=`, or should they calculate the combined effect and try to match that?

To get a good API, we need answers that seem logical across the majority of such interactions, instead of looking at an issue in isolation, fixing it, and then looking at the new issues caused by the fix.

newfrenchy83:
> The simple part of the answer is, it's not getting merged for 1.18.
> 
> However, the main answer, about when it became too late, is more complex. Although this PR would make the logic seem clearer in isolation, there's questions about how it interacts with filters, and how multiple abilities interact if `cumulative` says to only apply the largest value. Filters for `add=` have the feature of automatically checking for a `sub=` attribute if the ability doesn't have an `add=`. If abilities can usefully have both `add=` and `sub=`, should those filters try to match `add=`, `sub=`, or should they calculate the combined effect and try to match that?
> 
> To get a good API, we need answers that seem logical across the majority of such interactions, instead of looking at an issue in isolation, fixing it, and then looking at the new issues caused by the fix.

For me f abilitie have both add and sub then filter must could check add and sub and not combined effect(filter check if abilitie contain add= with value checked or sub with value writed, and che sub=-add is already when abilitie not have add attribute confusing for me), and then what add=value matched was usable same if sub is present arre more coherent and more readable what combined value.

Pentarctagon:
So from the last comments it sounds like at a minimum, further changes are needed to this PR to have filters handle add and sub independently?

As for cumulative, my understanding is that essentially only affects leadership, and for that I'd say that it should use the largest value after taking add and sub into account.

newfrenchy83:
> As for cumulative, my understanding is that essentially only affects leadership, and for that I'd say that it should use the largest value after taking add and sub into account.

No else it is not possible to modify value with add and sub: if i sue marrsman and another chance_to_hit who add 10% regarless of chance_to_hit already used then cumulative=yes concern marksman only cumulative=yes use largest value between 'value' and base_value, and other attribute like add and sub can change that after

newfrenchy83:
> So if an ability had `value=15` with `add=15` (sums to 30) and another ability of the same type had `value=20` with `add=5` (sums to 25), then it would use the second ability and give you 25 since the value of 20 is higher than the value of 15?

That will be 'value=20' with 'add=15' the highest value of each attribute will be used and sum=35

newfrenchy83:
The calculation is done after the 'for' loop where highest value of each attribute are selected and after check of all list the calulation comp_val = value + add etc is done

stevecotton:
> further changes are needed to this PR to have filters handle add and sub independently?

It's not just filters, IMO, it's having an overall reference document (plus the core macros) where a UMC author thinking "how do I do X?" ends up using the common way to do it, and thus we get less edge-cases that need to be supported. Even if the current behavior isn't logical, I want to avoid having it change between 1.18 to 1.20, and then having it change yet again in 1.22.

newfrenchy83:
@Pentarctagon could you merge this PR, because i yet update test here.

newfrenchy83:
@Pentarctagon Test fixed

Pentarctagon:
For my previous comment:

> So for this change, the two general areas are the way the ability's effect is applied and the way the attributes on the ability are filtered, right?

I'm assuming the thumbs up from stevecotton means that's correct, so the way that add and sub are applied when used together and the way their values are filtered should match how multiply and divide work. It's been a while, but does this PR do that?

newfrenchy83:
> For my previous comment:
> 
> > So for this change, the two general areas are the way the ability's effect is applied and the way the attributes on the ability are filtered, right?
> 
> I'm assuming the thumbs up from stevecotton means that's correct, so the way that add and sub are applied when used together and the way their values are filtered should match how multiply and divide work. It's been a while, but does this PR do that?

Yes, if multiply and divide work in same special then add and sub should doing the same.

Pentarctagon:
@stevecotton does add and sub working the same as multiple and divide provide a good enough answer for

> It's not just filters, IMO, it's having an overall reference document (plus the core macros) where a UMC author thinking "how do I do X?" ends up using the common way to do it, and thus we get less edge-cases that need to be supported. Even if the current behavior isn't logical, I want to avoid having it change between 1.18 to 1.20, and then having it change yet again in 1.22.

?

soliton-:
I think this PR is fine to make the existing behavior more consistent.

In the long run I think it makes more sense to leave the whole add/sub/div/mul/value thing behind (deprecate indefinitely) and move to a single attribute (maybe `formula`) that supports WFL. Having all those separate attributes comes from a time before you could just do math with WFL. Now with WFL it's making everything more complicated for no benefit.

stevecotton:
If the long-term goal is to deprecate all four attributes, why not deprecate `sub` now instead of changing it?

I don't see an audience that would benefit from this change, over the alternative of deprecating `sub`. Anyone who is already using it in UMC will find the behavior subtly change, and possibly have to have separate code for 1.18 vs 1.20. Anyone who isn't already using it might as well be warned to use `add` instead.

This also affects the filters, but they aren't changed in this PR, so a filter for `add=-2` will still match an ability with `sub=2`, even though `add` and `sub` themselves are now independent. I think filters would be an argument against WFL, because how would filters handle WFL?

newfrenchy83:
@stevecotton @soliton- [experimental_filter_ability] takes into account the value of the attributes as they are written in the code: for example if [filter_ability]add=1-15 and we have a [damage]add=3 and another [damage]sub=10 but neither has its own id then the filter will match add=3 even if the latter is not taken into account in the final calculation because of sub=10, which creates an inconsistency that this PR also aims to fix.

Pentarctagon:
I can't imagine deprecating add/sub/multiply/divide would be worth the work for UMC authors.

soliton-:
> If the long-term goal is to deprecate all four attributes, why not deprecate `sub` now instead of changing it?
> 
> I don't see an audience that would benefit from this change, over the alternative of deprecating `sub`. Anyone who is already using it in UMC will find the behavior subtly change, and possibly have to have separate code for 1.18 vs 1.20. Anyone who isn't already using it might as well be warned to use `add` instead.

For a deprecation to happen the new way would need to get implemented first. I have no clue when that is going to happen. If you want to just deprecate `sub` because it does not change expressibility I suppose that is a possibility. Might as well deprecate add/sub/multiply/divide and we're probably already where I was suggesting...

> This also affects the filters, but they aren't changed in this PR, so a filter for `add=-2` will still match an ability with `sub=2`, even though `add` and `sub` themselves are now independent. I think filters would be an argument against WFL, because how would filters handle WFL?

To me exactly those weird filters are a good reason to move away from this. Who tries to filter on some formula parts and somehow make that work no matter how the formula is written? That sounds like a losing proposition.

You would filter on the base or composite value only.

---

> I can't imagine deprecating add/sub/multiply/divide would be worth the work for UMC authors.

The point of indefinite deprecation would be so they don't need to do anything.

newfrenchy83:
i don't see how using formula would be simpler what attribute for UMC author

Pentarctagon:
> The point of indefinite deprecation would be so they don't need to do anything.

That sounds like the worst of both worlds. It won't simplify the code since we'll never actually be able to remove support for those attributes while also annoying UMC authors with the deprecation message.

gfgtdf:
> Having all those separate attributes comes from a time before you could just do math with WFL...

The main reason for the current way of doing things is that it's independent of the order in which the effect is applied, if for example a unit is.undee the effect of multiple [damage] special then the current system apples all the adds first then the multiplications. If you replace them with a singular formula attribute we have to come up with a new way how multiple effects interact consistently

gfgtdf:
> That sounds like the worst of both worlds. It won't simplify the code since we'll never actually be able to remove support for those attributes while also annoying UMC authors with the deprecation message.


I fully agree (on the general statement, unrelated to the outcome of this particular feature)

soliton-:
> i don't see how using formula would be simpler what attribute for UMC author

Currently there are 5 attributes with formulas.

---

> > The point of indefinite deprecation would be so they don't need to do anything.
> 
> That sounds like the worst of both worlds. It won't simplify the code since we'll never actually be able to remove support for those attributes while also annoying UMC authors with the deprecation message.

If we show annoying messages for indefinite deprecation then that sounds like something to fix. Do you have an example?

Not much to say otherwise. That is the price for compatibility. If the code is so bad it needs to be removed ASAP then I guess we can't have compatibility.

---

> If you replace them with a singular formula attribute we have to come up with a new way how multiple effects interact consistently

Hmm, the weird cumulative handling does make it difficult...

Pentarctagon:
If we don't show a deprecation message to UMC authors for something that's deprecated indefinitely, what does deprecating it accomplish?

soliton-:
It shows those that check the documentation (because they write new code for example) what the preferred way is.

Indefinite deprecation means you should move away from that code but it will continue to work so you don't _have_ to rewrite existing working code. Pretty much like in documentation of any other project were certain use is discouraged even if it does work and will continue to work in the foreseeable future.

There is also no reason to show no message at all. It can be shown in debug mode or so but there is no reason for it to be annoying and I'm fairly sure that is what we already do. (not show a message, not sure if we show some in debug mode)

In the long run it hopefully accomplishes the use going away so if in the future there is a reason to change the deprecation level it might not even be needed since there are no more uses.

newfrenchy83:
could you stop of speak formulas here, please?

stevecotton:
Is anyone, other than the unit tests, using `sub`? Could all of those use `add` with a negative value and get the same effect?

I don't know the answers to those, but my guess is that switching to `add` would be the same amount of effort as switching to a new behavior of `sub`.

Pentarctagon:
Given this only affects cases where both add and sub are used in the same ability, I'm not sure why such lengthy discussion is needed. If anyone is serious in wanting to deprecate or remove attributes, that should be discussed in a separate issue rather than blocking this PR from getting merged.

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_calculations.cfg
```diff
@@ -130,23 +130,23 @@
...
+{ONE_CALCULATION_UNIT_TEST add_sub_combined 5 (add,sub=2,3) 4}
 #####
 # API(s) being tested: [attacks]add=,[attacks]sub=
 ##
 # Expected end state:
-# Bob's attack has 2 strikes. Even in separate [attacks] tags, without separate ids the sub overwrites the add.
+# Bob's attack has 4 strikes. In separate [attacks] tags, without separate ids  both add and sub are used.
 #####
-{TWO_CALCULATION_UNIT_TEST add_sub_separated 5 (cumulative,add=yes,2) (cumulative,sub=yes,3) 2}
+{TWO_CALCULATION_UNIT_TEST add_sub_separated 5 (cumulative,add=yes,2) (cumulative,sub=yes,3) 4}
```
stevecotton, line 142: There's a typo in the already-merged version of this test, and it's testing exactly the same code as `add_sub_cumulative`. This one shouldn't have had the `cumulative=yes` parts, it should have just been (and I've retested and confirm this passes too)
```
{TWO_CALCULATION_UNIT_TEST add_sub_separated 5 (add=2) (sub=3) 2} # result will become 4 with this PR
```

stevecotton:
> Given this only affects cases where both add and sub are used in the same ability

That's incorrect, `sub` disables `add` in other abilities, as checked in the `add_sub_separated` and `add_sub_cumulative` tests. This is why I'm wondering if any UMC is using that edge case.

If a UMC author has put add and sub in the same ability then they've a clear logic error in their code.

> For a deprecation to happen the new way would need to get implemented first. I have no clue when that is going to happen.
> 
> If you want to just deprecate `sub` because it does not change expressibility I suppose that is a possibility. Might as well deprecate add/sub/multiply/divide and we're probably already where I was suggesting...

The "new way" would be to use `add` with a negative value. If a UMC author wants to subtract from a number, then the answer should be "use add"; and if they want to modify it by a variable which might be either positive or negative, the answer should be "use add".

That doesn't apply to `divide`, because the multiply/divide pair has cases where only one of the two is the correct choice, because trying to use multiply to divide by exactly 3 would need knowledge of exactly how many decimal places the multiplier is represented with (I think `multiply=0.34` would be the closest to `divide=3`).

> The point of indefinite deprecation would be so they don't need to do anything.

The point of deprecating `sub` would be to warn that it behaves in a way that's probably not what they want. This PR would fix that, at the cost of breaking any UMC (if any) that has made use of the weird behavior - and the reason that I don't want this PR to merge is that I think that breakage outweighs the benefits of any useful use of `sub`.

gfgtdf:
> The "new way" would be to use add with a negative value. If a UMC author wants to subtract from a number, then the answer should be "use add"; 

When using "sub" the engine will choose the best (=The one that substracts the most, within the same id) ability avilable, just like the engine chooses for example the best leadership avilable when a unit is under multiple leaders, using "add" doesnt give that

> sub disables add in other abilities

From a brief look at the c++ code this statement seems incorrect on multiple levels

- The test leave the id empty which means they are both in the ""-id group, which is by itself a corner case, since most abilites have ids. Most abilitie have ids.
- also even within the same id group, sub doesnt always disable add, it just takes what came last (in the test the sub ability is after the add ability), during the actualy game this means that which ability is actually taken (in the case of multiple leadership-like abilities) depens on the position if the units, which is imo obviously undesired, and i assumed the main reason for this PR.

stevecotton:
>  also even within the same id group, sub doesnt always disable add, it just takes what came last (in the test the sub ability is after the add ability), during the actualy game this means that which ability is actually taken (in the case of multiple leadership-like abilities) depens on the position if the units, which is imo obviously undesired, and i assumed the main reason for this PR.

Okay, that potential OOS convinces me that this should merge, although I'd like to get an update to the tests merged first so that we know what it used to do, as the tests are useful documentation for what changed when.

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_calculations.cfg
newfrenchy83: fixed

stevecotton: Test's going to fail, because the result will become 4 with this PR.

However, I'd like to merge #9313 first, please.

newfrenchy83: ok, i fix and rebase after you.

newfrenchy83: i lready done a fix of test, sorry. when your commit willbe merged, i done another rebase.

gfgtdf:
I mean, I dont think it's technically an oos since th oder should be the same on all clients, it's just the it's kinda undefined in. bad way from an wml authors point of view.

newfrenchy83:
@stevecotton tests rebased, i wait check result.

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_calculations.cfg
```diff
@@ -95,9 +95,9 @@
 # API(s) being tested: [attacks]sub=
 ##
 # Expected end state:
-# Bob's attack has 7 strikes. The difference to add2_add3_cumulative is that sub chooses the lower value, and the order of the abilities determines the order of the calculations.
+# Bob's attack has 10 strikes. The difference to add2_add3_cumulative is that add and sub are both used.
```
stevecotton, line 98: Technically, that's not a difference.

Bob's attack has 10 strikes. Both the add and the negative sub are used.

src/units/abilities.cpp
```diff
@@ -2381,7 +2382,17 @@ effect::effect(const unit_ability_list& list, int def, const_attack_ptr att, EFF
 		effect_list_.push_back(val.second);
 	}
 
-	composite_value_ = static_cast<int>((value_set + addition) * multiplier / divisor);
+	/* Using 'substraction' independant of 'addition' for abilities with same id and 'add' and 'sub'
+	 * for what rule of calculation was base+ add_value - sub_value
+	 * instead of base + (add_value or -sub_value).
```
stevecotton, line 2387: The "instead of" makes sense in the context of a PR (because we know what's changing), but won't make sense when looking at this code in future.

"Additional and subtraction are independent since Wesnoth 1.19.4. Prior to that, they affected each other."

stevecotton (approved):

src/units/abilities.cpp
newfrenchy83: Fixed

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_calculations.cfg
newfrenchy83: Fixed

PR #7272, opened 2023-01-12
Test fire_event of kill unit on recall list

Toranks:
This test checks if the last_breath and die events are triggered when a unit on the recall list is killed.
As you can see, Alice does not trigger the last_breath and die events when she is on the recall list. I don't know if it's an expected behavior or not.

Edit: Changed conditions to make the test succeed instead of fail

imagen
imagen

### DIFF
--- a/doc/man/wesnoth.6
+++ b/doc/man/wesnoth.6
@@ -190,8 +190,8 @@ lists defined log domains (only the ones containing
 .I filter
 if used) and exits
 .TP
-.B --log-to-file
-redirects logged output to a file. Log files are created in the logs directory under the userdata folder.
+.B --no-log-to-file
+prevents redirecting logged output to a file. Log files are created in the logs directory under the userdata folder.
 .TP
 .BI --max-fps \ fps
 the number of frames per second the game can show, the value should be between
--- a/src/commandline_options.cpp
+++ b/src/commandline_options.cpp
@@ -261,7 +261,7 @@ commandline_options::commandline_options(const std::vector<std::string>& args)
 		("log-debug", po::value<std::string>(), "sets the severity level of the specified log domain(s) to 'debug'. Similar to --log-error.")
 		("log-none", po::value<std::string>(), "sets the severity level of the specified log domain(s) to 'none'. Similar to --log-error.")
 		("log-precise", "shows the timestamps in log output with more precision.")
-		("log-to-file", "log output is written to a file rather than to standard error.")
+		("no-log-to-file", "log output is written to standard error rather than to a file.")
 		;
 
 	po::options_description multiplayer_opts("Multiplayer options");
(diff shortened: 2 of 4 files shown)


Pentarctagon:
You'll also need to add `0 kill_fires_events` to https://github.com/wesnoth/wesnoth/blob/master/wml_test_schedule in order for new unit tests to actually be executed during the CI run. `wml_test_schedule` is of the format `<return code> <test scenario id>`, where the list of valid return codes can be found at https://github.com/wesnoth/wesnoth/blob/master/run_wml_tests#L22-L33

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/kill_fires_events.cfg
```diff
@@ -0,0 +1,34 @@
+# This test checks if the last_breath and die events are triggered when a unit on the recall list is killed with [kill] tag
+# The default behavior is that only units on the map trigger the events, so this test will fail with alice_breath and alice_die
```
Pentarctagon, line 2: There's a standard format used in all the other tests for documenting what the test does. For this one for example, it'd look like:
```
#####
# API(s) being tested: [kill]fire_event=yes
##
# Actions:
# Kill bob and alice.
##
# Expected end state:
# A last breath event and a die event is triggered for both bob and alice.
#####
```

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/kill_fires_events.cfg
```diff
@@ -0,0 +1,34 @@
...
+			id=alice
+		[/put_to_recall_list]
+		[kill]
+			id=alice,bob
+			fire_event=yes
+		[/kill]
+        {ASSERT ({VARIABLE_CONDITIONAL bob_die equals yes})}
+        {ASSERT ({VARIABLE_CONDITIONAL alice_die equals yes})}
+        {ASSERT ({VARIABLE_CONDITIONAL bob_breath equals yes})}
+        {ASSERT ({VARIABLE_CONDITIONAL alice_breath equals yes})}
```
Pentarctagon, line 30: When possible, it's best to try and have the test reach the SUCCEED macro, so if alice isn't actually going to trigger the events, changing those from checking `equals` to `not_equals` would be better.

stevecotton: For `yes` and `no`, there's `boolean_equals` and `boolean_not_equals`. The difference from`equals` is that `boolean_equals no` will also match the variable being unitialised or `false` as well as `no` - and in this case, matching when it's uninitialised matches if the event didn't fire.

Toranks: That is quite useful. It would allow me to abbreviate some code. I hadn't realized that, thanks, I thought they were equivalent.

stevecotton:
> I don't know if it's an expected behavior or not.

Thanks for the test. In situations like this, the "expected" behavior is whatever the current stable branch does - I've just tried cherry-picking the test to 1.16, and confirmed that 1.17 and 1.16 have matching behavior.

PR #7273, opened 2023-01-12
Add start script to force the correct arch.

Pentarctagon:
Since Steam itself is not a native ARM app, for macOS it launches Wesnoth using the x86_64 version instead of the ARM version. This script forces the correct arch to be used which leads to improved performance (especially for 1.16) when using ARM-based Macs.

---

@hrubymar10 - The main question I have is where this file needs to be in order for it to work with how Wesnoth is packaged. I assume putting it in the repository root won't actually work.
@irydacea Do you know if additional arguments can be provided to `"The Battle for Wesnoth"` when used with the `arch` command?

### DIFF
--- a/data/campaigns/tutorial/achievements.cfg
+++ b/data/campaigns/tutorial/achievements.cfg
@@ -2,5 +2,12 @@
 [achievement_group]
     display_name=_"Tutorial"
     content_for=tutorial
-    {ACHIEVEMENT "completed" _"Complete the Tutorial" _"Complete all scenarios of the Tutorial campaign."}
+    [achievement]
+        id="completed"
+        name=_"Complete the Tutorial"
+        description=_"Complete all scenarios of the Tutorial campaign."
+        icon="data/core/images/icons/potion_red_small.png"
+        icon_completed="data/core/images/icons/potion_green_small.png"
+        max_progress=2
+    [/achievement]
 [/achievement_group]
--- a/data/campaigns/tutorial/scenarios/01_Tutorial_part_1.cfg
+++ b/data/campaigns/tutorial/scenarios/01_Tutorial_part_1.cfg
@@ -1183,6 +1183,13 @@
             caption= _ "Victory"
             message= _ "After your victory notice, the map will be grayed out to indicate that the scenario is over; this is called <i>linger mode</i>. You will still be able to examine the final positions and state of your troops and any surviving enemies. When you’re finished, click the <b>End Scenario</b> button to go on to the next scenario in the campaign."
         [/hint_message]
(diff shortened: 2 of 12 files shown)


CelticMinstrel:
According to the manpage, trailing arguments are forwarded to the application.

It also sounds like `-64` potentially solves the issue I mentioned in chat, automatically choosing arm or intel based on the host machine.

Pentarctagon:
Alright, I'm going to enable logging to file then as well.

irydacea:
Like celmin said, arguments after the executable name will be passed onto it.

One thing worth considering is making sure that this usage of `arch` is accepted in the oldest macOS version we support atm, which appears to be 10.12 for both 1.16 and dev.

hrubymar10:
@Pentarctagon 

I would rather detect if we are running on apple silicon or not. if yes, then we can use `arch -arm64 /path/to/wesnoth` and if not, I would execute just `/path/to/wesnoth` which will run what it think is best for the system. As it would be x86_64 system, then it would run x86_64 slice.

Also every Apple Silicon macOS version could use `arch` command so that effectively solve the problem which @irydacea mentioned.

hrubymar10 (changes requested):
This script must be included inside MacOS or Resources folder withing .app bundle. It would be used only by Steam so more describing name would be something like `start_from_steam.sh`. Then we will instruct steam to not run the wesnoth executable in .app bundle, but that script.

Also, maybe we don't have to include that script inside .app bundle at all, but rather apply it to the steam launch cmd instead. The strong advantage of that solution would be, that this change will work right away, without need for new release.

One more thing: there is also probably something like `1.14` branch on steam where old version of wesnoth lives. That version is not arm64 compatible so steam launch cmd should take that in mind.

Pentarctagon:
Do you have a concern with the `arch`-based approach? It was tested and confirmed to work on both an ARM Mac and an Intel Mac by Iris and Knyghtmare respectively.

The `arch` approach could (I think) be done without needing a script. Currently it has:
```
Executable: The Battle for Wesnoth.app/Contents/MacOS/The Battle for Wesnoth
Arguments: 
```
which I think should be able to be changed to:
```
Executable: /usr/bin/arch
Arguments: -64 "The Battle for Wesnoth.app/Contents/MacOS/The Battle for Wesnoth"
```

The approach you brought up would definitely need a separate script though.

Pentarctagon:
@hrubymar10 any thoughts about the `arch`-based approach without using a script?

Pentarctagon:
Closing since I forgot this was already on my fork's master branch - the question still stands though.

PR #7274, opened 2023-01-12
Units: Giant Ant advances to Soldier Ant

knyghtmare:
Adds the missing advancement, unless there is argument that soldier ants and worker (Giant) ants should be different unit-lines as Darwinism goes.

### DIFF
--- a/data/core/units/monsters/Ant_Soldier.cfg
+++ b/data/core/units/monsters/Ant_Soldier.cfg
@@ -6,15 +6,15 @@
     image="units/monsters/ant/soldier.png"
     profile="portraits/monsters/ant-soldier.webp"
     {DEFENSE_ANIM "units/monsters/ant/soldier-defend.png" "units/monsters/ant/soldier.png" hiss.wav }
-    hitpoints=22
+    hitpoints=35
     movement_type=orcishfoot
     movement=6
-    experience=80
     level=1
     alignment=neutral
+    experience=50
     advances_to=null
     {AMLA_DEFAULT}
-    cost=14
+    cost=16
     usage=fighter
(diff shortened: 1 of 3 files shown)


data/core/units/monsters/Ant_Soldier.cfg
```diff
@@ -6,15 +6,15 @@
...
-    experience=80
     level=1
     alignment=neutral
+    experience=50
     advances_to=null
     {AMLA_DEFAULT}
-    cost=14
+    cost=16
     usage=fighter
     description= _ "Giant Ants are common in almost any environment, from caverns deep under the earth to the tops of tall mountains. Most common are the nondescript workers and scouts, but in times of conflict, a more dangerous Soldier caste enters the fray."
```
stevecotton, line 19: If L0 workers can turn into soldiers then this description would need changes.

Personally, having the workers and soldiers as different castes sounds more interesting, although it wouldn't fit the game mechanics so well.

knyghtmare: > Personally, having the workers and soldiers as different castes sounds more interesting, although it wouldn't fit the game mechanics so well.

Given the unit visual similarity (sprite+portraits), players/umc-authors expect an advancement to be present. This is game mechanics.

However, I think to resolve this, maybe we can add a secondary advancement to Giant Ant titled "Worker Ant".
Worker Ants would be weaker but tankier since they dont fight but rather work for the development of the ant colony.

CelticMinstrel: Why does that make them tankier? It's the soldier ants who should be tanky. Maybe make the worker ants faster instead? So, they do less damage but have more MP. They're not built for fighting, but they're better at running away.

spixi:
When we have a Soldier and a Worker caste, maybe we can also have a winged Drone class (fragile scout unit).

knyghtmare:
Suggesting new units is fine, but I think there is zero effort in attracting new pixel artists which is the underlying real problem which nobody seems to notice...

PR #7278, opened 2023-01-12
[1.16] Reduce log spam about the postponed removal of the SPECIAL_NOTES macro

stevecotton:
This is a partial cherry-pick of b9e4fdffb46b9933a602ed2601b051c642e7ba07, merely reflecting that the removal has been postponed from 1.17 to 1.19 along with updating some outdated comments.

The motivation for backporting this is to make other warnings that are printed to stdout/stderr more visible, instead of being scrolled offscreen by the SPECIAL_NOTES messages.


stevecotton:
Will run the CI when the vcpkg cache has been generated - running it now would duplicate the builds (and it's a build with an intermittent failure during the vcpkg/glib part).

PR #7279, opened 2023-01-12
Enable logging to file by default.

Pentarctagon:


stevecotton (approved):

CelticMinstrel:
Does that mean nothing will be written to stdout unless you use `--no-log-to-file`?

Pentarctagon:
Correct.

CelticMinstrel:
Okay, guess that should go in the Xcode scheme so devs don't get confused about why the game suddenly isn't outputting…

PR #7282, opened 2023-01-13
SotBE S14 Grüü apparition fixed

Toranks:
This, instead of killing Grüü, moves him along with the others to participate in the conversation, and also switches him to side 1 (this side is temporary and only for narrative purposes, later on turn 7 he switches to side 4 again).
Grüü is correctly stored on the variable when the conversation ends, and restored later on turn 7.

### DIFF
--- a/data/campaigns/Son_Of_The_Black_Eye/scenarios/14_Back_Home.cfg
+++ b/data/campaigns/Son_Of_The_Black_Eye/scenarios/14_Back_Home.cfg
@@ -202,7 +202,7 @@
 			[/filter]
 			x=5
 			y=34
-			side=4
+			side=1
 		[/modify_unit]
     [/event]
 
--- a/data/campaigns/Son_Of_The_Black_Eye/scenarios/14_Back_Home.cfg
+++ b/data/campaigns/Son_Of_The_Black_Eye/scenarios/14_Back_Home.cfg
@@ -196,14 +196,14 @@
         name=prestart
 
         {RENAME_IF_DEAD thelarion_dead Thelarion (_"Valan")}
-		[modify_unit]
-			[filter]
-				id=Grüü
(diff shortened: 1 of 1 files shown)


Toranks:
I have another tested alternative based on what @soliton- told me was the best alternative for both mod compatibility and debugging.

First, Side 4 have a fake Grüü unit that not interferes with original Grüü:
```
    [side]
        side=4
        id=Grüü2
        name=_"Grüü"
        type=Great Troll
	canrecruit=yes
        team_name=Kapoue
        user_team_name=_ "Kapou’e"
        controller=human
    [/side]
```
This unit is killed on prestart
```
	[kill]
            id=Grüü2
        [/kill]
```

And the rest of the code is exactly the same as how I have changed it so far. The real Grüü instead of moving from the initial position of Side 4, with a new Grüü unit created from scratch, will come directly from the recall list, without any alteration.
With this we have two advantages:
- Grüü will keep all its objects and variables, in case they are being played with mods. Or in the possibility that in the future items that Grüü can carry are added to the campaign (and they will not be deleted).
- Side 4 starts the stage with a real leader, for debugging purposes.

If it seems good to you, I add these two changes to the PR

Pentarctagon:
Which is the other PR? This is the only PR you have open currently.

stevecotton:
> other PR

#7281

nemaara (approved):

stevecotton:
Is this waiting for anyone else's approval before merging?

@soliton- you're mentioned in the comments, I assume you're happy with the changes?

stevecotton:
Oops, meant to squash-merge that.

soliton-:
I looked at this and didn't understand the diff. Didn't have the time to look more closely though. If it does what we discussed then I'm happy with it.

PR #7284, opened 2023-01-16
WC: fix abnormal commander mechanic

### LINKED ISSUE #7264: Commanders in World Conquest work improperly

- **Version:** tested with 1.16.\*, I think 1.17.\* also affected

### Description of the bug

Commanders are supposed to become recruiting leaders if the current leader of the player's side dies. They become, but in the next scenario the player has Peasant as the leader of his side, which is clearly not supposed to be

1. Take a commander as a bonus in the start of a scenario
2. Get your current leader killed. The commander becomes a new leader
3. Proceed to the next scenario. The commander can still recruit but your leader is a peasant. Death of the peasant means defeat

### Expected behavior

The commander is your only leader

### Additional context

_No response_

knyghtmare:
Fixes #7264

### DIFF
--- a/data/campaigns/World_Conquest/lua/game_mechanics/promote_commander.lua
+++ b/data/campaigns/World_Conquest/lua/game_mechanics/promote_commander.lua
@@ -19,8 +19,11 @@ on_event("die", function(cx)
 	}
 	commander = commander[1]
 	if commander then
+		commander:extract()
+		commander.id = "wc_leader"..u.side
 		commander.canrecruit = true
 		commander:remove_modifications({ id = "wc2_commander_overlay" })
+		commander:to_map()
 		wesnoth.wml_actions.message {
 			id = commander.id,
 			message = strings.promotion
--- a/data/campaigns/World_Conquest/lua/game_mechanics/promote_commander.lua
+++ b/data/campaigns/World_Conquest/lua/game_mechanics/promote_commander.lua
@@ -20,7 +20,7 @@ on_event("die", function(cx)
 	commander = commander[1]
 	if commander then
 		commander:extract()
-		commander.id = "wc2_leader"..u.side
+		commander.id = u.id
 		commander.canrecruit = true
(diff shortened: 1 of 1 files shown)


stevecotton:
Looks reasonable to me

knyghtmare:
> but IIUC not backportable.

I received requests that backport-compatibility be overlooked for World Conquest, as the 1.16 version is a "half-broken disaster" due to UX-bugs in it. (several artifacts not working as intended).
I know the 1.15.x cycle disabled WC for the most part so playtesting was not properly done but should getting stacked with so many negative feedback warrant some way to get updates into the 1.16.x WC?

stevecotton:
I can see the logic of accepting incompatible changes in WC for 1.16. Maybe open an issue or have a discussion in the Discord dev channel first?

@Hejnewar, how do you feel about it.

data/campaigns/World_Conquest/lua/game_mechanics/promote_commander.lua
```diff
@@ -19,8 +19,11 @@ on_event("die", function(cx)
 	}
 	commander = commander[1]
 	if commander then
+		commander:extract()
+		commander.id = "wc_leader"..u.side
```
soliton-, line 23: Wouldn't it need to be `wc2_leader` to match the save_id? If you use the id of the previous leader (u.id) instead it doesn't matter anyway.

stevecotton:
I wonder how much work it would be to turn WC back into an add-on, so players who get their 1.16 from a Linux distribution can download a fully-working version.

data/campaigns/World_Conquest/lua/game_mechanics/promote_commander.lua
```diff
@@ -19,8 +19,11 @@ on_event("die", function(cx)
 	}
 	commander = commander[1]
 	if commander then
+		commander:extract()
+		commander.id = "wc2_leader"..u.side
```
soliton-, line 23:
```suggestion
		commander.id = u.id
```

PR #7285, opened 2023-01-16
[LoW] fix scenario count inconsistency

### LINKED ISSUE #7218: [LoW] Inconsistency in Stated Scenario length between SP/MP
Scenario description says that the campaign is 18 scenarios. This is true...for single-player mode.

HOWEVER, yes, you all didn't notice this one...in MP mode (as done in c2c4d819fdde8ef6cc9e0217592b5a1052dc2ef7), one particular scenario is not accessible (S16). Thus, the campaign length in MP mode is actually 17.

knyghtmare:
Closes #7218


stevecotton:
Needs `# po:` comments on both descriptions.

nemaara (approved):

Issue #7217, opened 2022-12-19
pango layout problem with 1.16 branch on MSVC+cmake builds


Wedge009:
### Game and System Information

- **Version:** 1.16
- **Downloaded from:** GitHub
- **Build info:** MSVC + cmake
- **OS:** Windows 7 + 10

### Description of the bug

Been having a run of bad luck with building/running 1.16 on Windows. I was still using manual VC project files with MSVC for 1.16 branch until very recently. In my set-up some dependencies I built myself (eg OpenSSL), others I downloaded pre-built (eg SDL), others still I just used via vcpkg. This might sound like a horrible mix-up but it was working for me till recently.

pango is one of the dependencies I used from vcpkg. But I got to a point where the built binaries just wouldn't reach the main menu screen. Eventually I found it was a pango layout issue, resulting in the log file message:
```
User message: Failed to show a dialog, which doesn't fit on the screen.
Dev message: ...\src\gui\widgets\window.cpp:945 - in function 'gui2::window::layout' found the following problem: Failed to size window, wanted size 940x1916560 available size 1600,900 screen size 1600,900.
```

Giving up on it once I found the issue was in pango, I switched to full vcpkg build via cmake which resulted in missing images due to SDL being unable to load images for some reason (discussed in #wesnoth-dev). Eventually the vcpkg baseline was updated and I am now at the point where I was to begin with - ie pango layout failing with vcpkg builds.

### Steps to reproduce the behavior

1. Switch to 1.16 branch
2. Compile with MSVC + cmake
3. Attempt to run Wesnoth
4. Observe start-up screen shows squares instead of normal printable characters and subsequently crashes before reaching the main menu screen.

### Expected behavior

Normal execution through to the main menu screen.

### Additional context

master branch builds fine, as do all builds on Linux.

Wedge009:
![1 16](https://user-images.githubusercontent.com/13679322/208350564-80e1d7df-4830-4276-8049-79199559d9ac.png)

stevecotton:
What language is it using? Does it still happen in `en_US`?

Please check in your preferences file first, because if you just override it with `--language` then we may lose one of the reproduction steps for the bug.

Wedge009:
Removing preferences doesn't appear to make a difference.

From my previous debugging efforts, the call in question is `pango_layout_get_pixel_extents`:
[...]

PR #7287, opened 2023-01-16
Bump vcpkg baseline to use a non-obsolete version of libtool

stevecotton:
The vital commit is fd766eba2b4cf59c7123d46189be373e2cee959d, but I'm taking the most recent version of vcpkg as the new baseline because it might as well use the latest.


Wedge009:
Hmm, despite this update I'm still getting #7217.

PR #7289, opened 2023-01-18
Fix indent on unit tests, and make the CI check it

stevecotton:
The whole data/test directory was excluded from the CI's indentation check.
Two of the commits in this PR fix the indentation in some of the tests, and
then the final commit reduces the exclusion list to only

* data/test/scenarios/cve_tests/
* data/test/scenarios/lua_tests/
* data/test/scenarios/wml_tests/TerrainWML/

Exclude cve_tests because of the binary file that's masquerading as a .cfg;
I think it's correct to exclude the whole directory because any new tests
added there might also use weird edge cases in the parsing.

Exclude lua_tests as a "keep this PR small" thing. We should fix the bugs in
wmlindent, but I don't want to take on that work.

The exclusion of TerrainWML is again because I don't want to take on the work
of working out whether it's problems with terrain masks or bugs in wmlindent.

Most of the files in those directories wouldn't be changed, the ones that would
(or would trigger an error about not being valid UTF-8 ) are:

* data/test/scenarios/cve_tests/test_cve_2018_1999023_2.cfg
* data/test/scenarios/lua_tests/functional/lua_functional.cfg
* data/test/scenarios/lua_tests/wesnoth/map/mapgen_filter.cfg
* data/test/scenarios/wml_tests/TerrainWML/test_terrain_mask.cfg

Notes on specific files:
---

For the unit tests using sync.evaluate_single or sync.evaluate_multiple, I'm
using two wmlindent bugs against each other, putting two opening statements on
the same line to balance two closing statements on the same line. However, the
unit tests ought to be code that can be cut & pasted elsewhere without causing
indentation problems, and this is simpler than fixing wmlindent's Lua parsing.

There are two cve_2018_1999023 tests. The other one needs to be treated as a
binary file, but the one changed here is plain text that loads its attack from
a separate file.

The change in start_position_generic.cfg doesn't seem good, but it's not
particularly bad and it's what wmlindent currently does.

### DIFF
--- a/data/test/macros/shroud.cfg
+++ b/data/test/macros/shroud.cfg
@@ -48,21 +48,21 @@
 #enddef
 
 #define ASSERT_SHROUDED FILTER STATE
-[lua]
-    code=<<
-        local locs = wesnoth.map.find(wml.get_child(..., 'filter'))
-        local state = (...).state
-        for _,loc in ipairs(locs) do
-            local x, y = loc[1], loc[2]
-            unit_test.assert_equal(wesnoth.sides.is_shrouded(1, loc), state, string.format('shroud check at (%d,%d)', loc[1], loc[2]))
-        end
-    >>
-    [args]
-        [filter]
-            include_borders=yes
-            {FILTER}
-        [/filter]
-        state={STATE}
-    [/args]
-[/lua]
+    [lua]
+        code=<<
+            local locs = wesnoth.map.find(wml.get_child(..., 'filter'))
+            local state = (...).state
+            for _,loc in ipairs(locs) do
+                local x, y = loc[1], loc[2]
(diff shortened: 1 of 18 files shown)


stevecotton:
The recent push added reindentation for files included in PR #7109.

CelticMinstrel:
I don't know why you need my input on this, but I don't really have anything to say about it, which includes nothing bad.

stevecotton:
Thanks - I thought you'd be the one to review because the files that needed hand-editing were mainly Lua, and it's making a de-facto standard of putting `sync.evaluate_single(function()` on a single line.

CelticMinstrel:
I don't really have an issue with that standard. In fact, I guess I'd say I mildly prefer it to putting `function()` on the next line.

PR #7291, opened 2023-01-19
[1.16] HttT S03: Rewrite intro, hint about training troops for S06

stevecotton:
Backport of #7214, simple cherry-pick with nothing to fix up afterwards.

The island gives the player two fronts to progress on, and it's early enough to give the player time to prepare for the SoE. Hint to them about improving their recall list.

When trying to add that hint as a single-line change to the dialogue, I reviewed the existing text and felt it didn't flow well. Reading the feedback thread, Delfador should have more emotion about finding his childhood home overrun; however I've gone for practical thoughts first.

The hint isn't #ifdef to only appear on easy, because the text wouldn't flow without it - take that line out and it goes from "there are more orcs than I thought" to "we must fight these orcs, who aren't the ones at the SoE". Also, players who are already doing XP management can feel knowledgeable when they see it.

(cherry picked from commit 9f780903abfa592752cb02c767a4ef5992f6f9c7)

### DIFF
--- a/changelog_entries/httt_s03_intro.md
+++ b/changelog_entries/httt_s03_intro.md
@@ -1,0 +1,3 @@
+ ### Campaigns
+   * Heir to the Throne
+     * S03: Rewrite intro text, including a hint about training troops (PR #7291)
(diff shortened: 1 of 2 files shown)


changelog_entries/httt_s03_intro.md
```diff
@@ -0,0 +1,3 @@
+ ### Campaigns
+   * Heir to the Throne
+     * S03: Rewrite intro text, including a hint about training troops (PR #7214)
```
stevecotton, line 3: I'll push an update to use the 1.16 PR number - it's easier to do that than to wonder whether it's necessary.

nemaara (approved):

Issue #2913, opened 2018-04-17
hp/xp_bar_scaling attributes of a [unit] do not work


Psimage:
[Wiki](https://wiki.wesnoth.org/SingleUnitWML#Unit_Data) states that **hp_bar_scaling** and **xp_bar_scaling** attributes should override [unit_type] and [game_config] values for that particular unit but actually have no effect. They are also not persistent as stated in the wiki:
https://github.com/wesnoth/wesnoth/blob/6d5f5d4299f41f54c979d2a5b785b4b26b9d2be2/src/units/unit.cpp#L1392-L1523

xp/hp_bar_scaling assignment happens in unit's **init()** method but later is overwritten in the call to **advance_to()**: https://github.com/wesnoth/wesnoth/blob/6d5f5d4299f41f54c979d2a5b785b4b26b9d2be2/src/units/unit.cpp#L466-L517
https://github.com/wesnoth/wesnoth/blob/6d5f5d4299f41f54c979d2a5b785b4b26b9d2be2/src/units/unit.cpp#L888-L957

jyrkive:
Thanks for debugging this! 👍 

Looking at source code history, I didn't find any commits which would have caused this behavior. It looks like this bug has existed ever since the attributes were introduced in 1.11.8 (commit e8e9412d4d9c6381c3452b43dec8bd84dc5095c3).

Psimage:
@fendrin 😠

PR #7292, opened 2023-01-19
Remove hp_bar_scaling and xp_bar_scaling from [unit] schema

ProditorMagnus:
#2913

### DIFF
--- a/data/schema/units/single.cfg
+++ b/data/schema/units/single.cfg
@@ -23,8 +23,6 @@
 	{SIMPLE_KEY small_profile string}
 	{SIMPLE_KEY role string}
 	{SIMPLE_KEY unit_description t_string}
-	{SIMPLE_KEY hp_bar_scaling s_real}
-	{SIMPLE_KEY xp_bar_scaling s_real}
 	{SIMPLE_KEY traits_description t_string}
 	{SIMPLE_KEY alignment string}
 	{SIMPLE_KEY advances_to string_list}


Pentarctagon:
From the linked issue, it sounds like this has never worked then?

CelticMinstrel:
This is basically fine, but before merging, we should answer the question of whether we _want_ this to work.

Is there any benefit to overriding these values on a unit instance? Maybe a unit with an AMLA tree that allows drastically improving its health?

Pentarctagon:
It doesn't seem especially useful - I'd guess most boss/hero units would likely already have their own unit types.

stevecotton:
I'm wondering if the `[unit_type]` variants of this are useful either, as most boss/hero units that would use it are likely to have enough hitpoints to reach the maximum limits anyway.

I don't have every add-on installed, but for searching the ones that I have on hand: in A Vision Blinded S10, a `last_breath` animation of a necromancer who is obviously returning in a later scenario sets the values to negative numbers via `[modify_unit]`. Haven't watched it to see what it does.

stevecotton:
> I'm wondering if the `[unit_type]` variants of this are useful either,

Implementing support for hiding the XP bar with `[unit_type]xp_bar_scaling=0` is being discussed for #7339, however that wouldn't need `[unit]xp_bar_scaling` to be in the schema. Even if it ends up in the `[unit]` in the savegame, savegames aren't supported by the schema (`[replay_start]` isn't in the schema for starters).

stevecotton (approved):

PR #7298, opened 2023-01-21
bi: Fix --report SIGSEGV caused by uninitialised video

irydacea:
There is currently a bug in master causing `--report` to raise a segmentation fault because of an attempt to dereference a null pointer to the game window to check the game's display configuration.

At some point in 1.17.x, the check for whether the game window was null disappeared. This brings it back by exposing a check through the `video` namespace so as to avoid piggybacking further on the `video::get_window()` function, which we are meant to avoid using.

### DIFF
--- a/src/build_info.cpp
+++ b/src/build_info.cpp
@@ -550,6 +550,10 @@ list_formatter video_settings_report_internal(const std::string& heading = "")
 		placeholder = "Running in non-interactive mode.";
 	}
 
+	if(!video::has_window()) {
+		placeholder = "Video not initialized yet.";
+	}
+
 	if(!placeholder.empty()) {
 		fmt.set_placeholder(placeholder);
 		return fmt;
--- a/src/video.cpp
+++ b/src/video.cpp
@@ -397,6 +397,11 @@ void init_window()
 	update_framebuffer();
 }
 
+bool has_window()
+{
+	return bool(window);
+}
+
 point output_size()
 {
 	if (testing_) {
--- a/src/video.hpp
+++ b/src/video.hpp
@@ -85,6 +85,9 @@ bool testing();
 /* Windowing functions */
 /***********************/
 
+/** Whether the game has set up a window to render into */
+bool has_window();
+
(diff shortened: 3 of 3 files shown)


src/build_info.cpp
```diff
@@ -550,6 +550,10 @@ list_formatter video_settings_report_internal(const std::string& heading = "")
 		placeholder = "Running in non-interactive mode.";
 	}
 
+	if(!video::has_window()) {
+		placeholder = "Video not initialized yet.";
```
stevecotton, line 554: "yet" implies that it will be done later, seems incorrect for a code path that's only for `--report`.

src/build_info.cpp
```diff
@@ -550,6 +550,10 @@ list_formatter video_settings_report_internal(const std::string& heading = "")
 		placeholder = "Running in non-interactive mode.";
 	}
 
+	if(!video::has_window()) {
```
stevecotton, line 553: Should this be an `else if` of the `if(video::headless())`?

irydacea: It doesn't really matter.

irydacea: I suppose you could argue that the latter message will clobber the previous one if Wesnoth is really running in headless mode. On the other hand the odds that `--report` will be used and relied upon while explicitly in headless mode seem quite slim, which kinda makes the headless case redundant.

src/build_info.cpp
irydacea: Is this really that important? :\

stevecotton: I guess not.

src/build_info.cpp
stevecotton: No. It's not important.

My opinion on this has changed in the previous 30 minutes, after investigating why `--report` doesn't imply headless mode anyway. Using `--report` gets processed in `process_command_args`, and then there's a mandatory quit afterwards. So I think there's no codepath that actually gets to here with `video::headless()==true`.

stevecotton (approved):

PR #7299, opened 2023-01-21
bi: Re-add screen dpi report line, add information on the renderer setup

irydacea:
This patchset re-adds a clipboard report line for the Game Version dialog showing the game screen's DPI like in 1.16. This particular bit was dropped by mesilliac in the middle of an architectural change to how DPI is handled in the game, however it really shouldn't have been part of the removal given that it's meant to be diagnostic information for developer use when processing bug reports from players.

Additionally, it adds new information on the renderer in use including its name (e.g. "metal" on macOS, "direct3d" on Windows), whether it's a software rasteriser (indicated by "sw"), whether it's hardware accelerated (indicated by "hw"), whether VSync is enabled, and the maximum renderer texture size.

### DIFF
--- a/src/build_info.cpp
+++ b/src/build_info.cpp
@@ -564,6 +564,12 @@ list_formatter video_settings_report_internal(const std::string& heading = "")
 	fmt.insert("Final render target size", geometry_to_string(video::output_size()));
 	fmt.insert("Screen refresh rate", std::to_string(video::current_refresh_rate()));
 
+	const auto& renderer_report = video::renderer_report();
+
+	for(const auto& info : renderer_report) {
+		fmt.insert(info.first, info.second);
+	}
+
 	return fmt;
 }
 
--- a/src/build_info.cpp
+++ b/src/build_info.cpp
@@ -521,6 +521,13 @@ inline std::string geometry_to_string(point p)
 	return std::to_string(p.x) + 'x' + std::to_string(p.y);
 }
 
+template<typename coordinateType>
+inline std::string geometry_to_string(coordinateType horizontal, coordinateType vertical)
+{
+	// Use a stream in order to control significant digits in non-integers
+	return formatter() << std::fixed << std::setprecision(2) << horizontal << 'x' << vertical;
+}
+
 std::string format_sdl_driver_list(std::vector<std::string> drivers, const std::string& current_driver)
 {
 	bool found_current_driver = false;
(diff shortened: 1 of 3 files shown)


src/build_info.cpp
```diff
@@ -558,11 +567,25 @@ list_formatter video_settings_report_internal(const std::string& heading = "")
 	const auto& current_driver = video::current_driver();
 	auto drivers = video::enumerate_drivers();
 
+	const auto& dpi = video::get_dpi();
+	std::string dpi_report;
+
+	dpi_report = dpi.first == 0.0f || dpi.second == 0.0f ?
+				 "<unknown>" :
```
stevecotton, line 574: Is this specifically because the string `0x0` looks more like hexadecimal than like a pair of numbers?

src/build_info.cpp
```diff
@@ -521,6 +521,15 @@ inline std::string geometry_to_string(point p)
 	return std::to_string(p.x) + 'x' + std::to_string(p.y);
 }
 
+template<typename coordinateType>
+inline std::string geometry_to_string(coordinateType horizontal, coordinateType vertical)
+{
+	// Use a stream in order to control significant digits in non-integers
```
stevecotton, line 527: The default according to cppreference is 6 digits of precision, which seems inappropriate. I think it would be better to set the desired precision in code, rather than commenting while leaving it at 6.

src/video.hpp
```diff
@@ -319,6 +319,26 @@ class quit : public lua_jailbreak_exception
...
+
+/**
+ * Retrieves the current game screen DPI for the @a build_info API.
+ */
+std::pair<float, float> get_dpi();
+
+
+/**************************/
+/* Implementation details */
+/**************************/
```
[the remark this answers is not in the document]
irydacea: I just followed the same formatting already used in the file. If that's an issue shouldn't it be fixed already?

src/build_info.cpp
irydacea: No, it's simply a display convention I already used in previous versions of this code (cf 1.16) for nonsense values like this.

src/video.hpp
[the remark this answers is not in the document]
stevecotton: Urgh. I don't like it, but if it's matching what's already there then it's correct.

src/build_info.cpp
irydacea: Oops. Set the precision in the new version and also switched to using `formatter` since it's already in use in the file.

stevecotton (approved):

PR #7307, opened 2023-01-21
Fix HP lost with dehydrated on tooltip

### LINKED ISSUE #7295: UTBS: Tooltip of dehydration shows incorrect number of HP lost due to dehydration
### Language

All

### Source

https://github.com/wesnoth/wesnoth/blob/3e3cd0612148e3e7ad5ecc3c4cf8e1c29b26116f/data/campaigns/Under_the_Burning_Suns/lua/theme.lua#L14

### Original Text

tooltip = _ "dehydrated: This unit is dehydrated. It will lose 4 HP and have its damage reduced by 1 each turn during the day unless prevented by healers or cured by water at an oasis.\n\nUnits cannot be killed or deal no damage as a result of dehydration."

### Proposed Change
Dehydration loss has different values on different difficulties. Put **dehydration_loss** variable, similar to this line on https://github.com/wesnoth/wesnoth/blob/3e3cd0612148e3e7ad5ecc3c4cf8e1c29b26116f/data/campaigns/Under_the_Burning_Suns/scenarios/02_Across_the_Harsh_Sands.cfg#L341

  message= _ "During the daytime (Dawn, Morning, Midday, Afternoon, and Dusk) at the beginning of each your turns, every unit in a sand, road, rubble or sand dune hex will suffer from thirst. Each turn of thirst reduces a unit’s attack damage and causes $dehydration_loss hitpoints of damage. Healers can prevent dehydration from becoming worse, but cannot heal it. Only by refreshing at an oasis (any water hex) at the start of your turn will your units regain full attack strength."

Toranks:
It's simple, only a variable substitution, and I just tested it. The only problem is that the string will have to be translated again in all languages.
imagen
To be more sure, I've tried the spanish translation, and it works.
imagen

This should resolve #7295

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/lua/theme.lua
+++ b/data/campaigns/Under_the_Burning_Suns/lua/theme.lua
@@ -7,11 +7,12 @@ function wesnoth.interface.game_display.unit_status()
 	local u = wesnoth.interface.get_displayed_unit()
 	if not u then return {} end
 	local s = old_unit_status()
+	local dehydration_loss = wml.variables["dehydration_loss"]
 
 	if u.status.dehydrated then
 		table.insert(s, { "element", {
 			image = "misc/dehydration-status.png",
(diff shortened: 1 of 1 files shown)


Wedge009:
Assuming this is what lua code requires, this change makes sense to me. The re-translation remark shouldn't be an issue if the original string is wrong.

stevecotton:
Tested and works.

PR #7308, opened 2023-01-21
Fix game data dir search for Visual Studio builds on Windows

irydacea:
The logic we used here hasn't had the desired results since the move from using custom Visual Studio project files to using the CMake integration system, so this patch updates it to match the layout we get with the CMake integration on Visual Studio 2019 and 2022.

Note that without this patch, the game is unable to detect a valid game directory by default when launched by MSVC++ without an explicit path, which results in an assertion failure somewhere in the filesystem API.

### DIFF
--- a/src/wesnoth.cpp
+++ b/src/wesnoth.cpp
@@ -986,12 +986,19 @@ static std::string autodetect_game_data_dir(std::string exe_dir)
 	else if(filesystem::file_exists(exe_dir + "/../data/_main.cfg")) {
 		auto_dir = filesystem::normalize_path(exe_dir + "/..");
 	}
-	// In Windows debug builds, the EXE is placed away from the game data dir
-	// (in projectfiles\VCx\Debug), but the working directory is set to the
-	// game data dir. Thus, check if the working dir is the game data dir.
+	// Allow using the current working directory as the game data dir
 	else if(filesystem::file_exists(filesystem::get_cwd() + "/data/_main.cfg")) {
 		auto_dir = filesystem::get_cwd();
 	}
+#ifdef _WIN32
+	// In Windows builds made using Visual Studio and its CMake
+	// integration, the EXE is placed a few levels below the game data
+	// dir (e.g. .\out\build\x64-Debug).
+	else if(filesystem::file_exists(exe_dir + "/../../build") && filesystem::file_exists(exe_dir + "/../../../out")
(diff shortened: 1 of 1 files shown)


src/wesnoth.cpp
```diff
@@ -986,12 +986,16 @@ static std::string autodetect_game_data_dir(std::string exe_dir)
 	else if(filesystem::file_exists(exe_dir + "/../data/_main.cfg")) {
 		auto_dir = filesystem::normalize_path(exe_dir + "/..");
 	}
-	// In Windows debug builds, the EXE is placed away from the game data dir
-	// (in projectfiles\VCx\Debug), but the working directory is set to the
-	// game data dir. Thus, check if the working dir is the game data dir.
-	else if(filesystem::file_exists(filesystem::get_cwd() + "/data/_main.cfg")) {
-		auto_dir = filesystem::get_cwd();
```
stevecotton, line 993: Looking for a `data` dir in the current working directory sounds like it could be useful on other platforms, not just Windows. So I think this should be left in, even if the newly-added `_WIN32`-only search is tried first.

irydacea: Hm this is true. I completely forgot that this had relevance outside of Windows (but also for non-standard Windows builds or installs) until much later.

Wedge009:
As per #wesnoth-dev discussion when I eventually switched over to cmake builds, I worked around this issue by manually adjusting the launch json settings to look for the data directory two levels up. But if it's possible to have this automated, that's preferable.

irydacea:
Updated patch to preserve the existing behaviour while adding the new one.

stevecotton (approved):

stevecotton:
> As per #wesnoth-dev discussion when I eventually switched over to cmake builds, I worked around this issue by manually adjusting the launch json settings to look for the data directory two levels up. But if it's possible to have this automated, that's preferable.

I'm beginning to wonder about two directories up; doing it automatically is probably correct as `data/_main.cfg` is a pretty strong indication that it's the right directory, but I'm not sure if it's certain. One directory up seems safe, and this PR has extra sanity checks about choosing three directories up.

Wedge009:
Sorry, I was going off memory. It's three directories up. The default build directory I get with MSVC/cmake is `out/build/x64-Release/`, so have to go up three directories before being able to access `data/`.

irydacea:
3 directories up is precisely what this patch uses. 😀

PR #7323, opened 2023-01-25
Move the Merman Citizen and Brawler units to core

CelticMinstrel:
This allows addons to safely use them without having to copy them into their own addon directory.

Opened as a draft because it's not quite done – the stun event needs to be handled somewhere other than in the brawler unit definition.


knyghtmare:
>  the stun event needs to be handled somewhere other than in the brawler unit definition.

Rewrite it in Lua and add it in as a always present global event? Like how Feeding, diversion and explode works?
Or, add it inside the weapon special code define?

CelticMinstrel:
> Rewrite it in Lua and add it in as a always present global event? Like how Feeding, diversion and explode works?

I was thinking of doing this, yeah.

Pentarctagon:
It would however still need to use the old events API.

data/lua/stun.lua
```diff
@@ -0,0 +1,72 @@
+
```
stevecotton, line 1: Looks like we need to fix a bug in the fix-whitespace script, because the CI should have caught this.

CelticMinstrel: ???

data/lua/stun.lua
```diff
@@ -0,0 +1,72 @@
...
+            image = "misc/stunned-status-icon.png",
+            tooltip = _"stunned: This unit is stunned. It cannot enforce its Zone of Control."
+        })
+    end
+
+    return s
+end
+
+local function on_hit(weapon, opponent)
+    if not weapon:matches{special_id = 'stun'} then return end
```
stevecotton, line 22: Has this already been filtered for whether the ability is active or not? Although the Brawler's ability is always active, if `stun` becomes a general ability like `slow` then it should support abilities with filters.

Should it match by `id`, or should it use a `[stun]` tag and match by tagname?

data/lua/stun.lua
```diff
@@ -0,0 +1,72 @@
...
+    if not opponent.zoc then return end
+    local text
+    if opponent.gender == "female" then
+        text = _ "stunned"
```
stevecotton, line 26: Genders are swapped compared to the `if` condition.

data/lua/stun.lua
```diff
@@ -0,0 +1,72 @@
...
+    else
+        text = _ "female^stunned"
+    end
+	local color = stringx.join(',', {'196', '196', '128'})
```
stevecotton, line 30: Indentation

data/lua/stun.lua
CelticMinstrel: Yeah you're right, it should probably be `special_id_active`.

data/lua/stun.lua
CelticMinstrel: Oh wow, whoops.

data/lua/stun.lua
CelticMinstrel: I have no idea what you're talking about, but if the fix-whitespace script doesn't deal with it I can't be bothered.

data/lua/stun.lua
```diff
@@ -0,0 +1,72 @@
...
+local function defender_hits(ctx)
+    local weapon = wesnoth.units.create_weapon(wml.get_child(ctx, 'second_weapon'))
+    local opponent = wesnoth.units.get(ctx.x1, ctx.y1)
+    on_hit(weapon, opponent)
+end
+
+on_event("attacker_hits", attacker_hits)
+on_event("defender_hits", defender_hits)
+
+-- Removing a status via object doesn't work apparently?
```
stevecotton, line 66: Does the ZoC get reset at the right time?

CelticMinstrel: It certainly looks like it. I saw the tint disappear and the ellipse return to normal at the end of that unit's side's turn.

data/lua/stun.lua
stevecotton: One of our tools fixes blank lines at the end of the file, and two consecutive blank lines in the middle of files. Maybe it only runs on .cfg files.

data/lua/stun.lua
stevecotton: You should be bothered, no point in having low-quality code if it's little effort to fix.

data/lua/stun.lua
CelticMinstrel: But this is neither of those?

data/lua/stun.lua
CelticMinstrel: If there was a visible indentation issue I'd be bothered, but it looks fine to me.

data/lua/stun.lua
CelticMinstrel: Or actually I guess it should use a `[stun]` tag and `special_type_active`.

knyghtmare (approved):
PR seems okay

data/lua/stun.lua
stevecotton: It looks right in your editor, with the tab-stops set to 4 spaces. It looks wrong in Github's diff, and possibly other editors.

data/lua/stun.lua
stevecotton: Having checked the old discussion, okay blank first lines are accepted.

data/lua/stun.lua
stevecotton: Currently, there's only one .lua file that has any tabs in at all - WC's unittypedata.lua. I thought it would be complex to check for files that mix tabs and spaces, but a blanket "no tabs in .lua files" check would be easy. Will open a PR later.

CelticMinstrel: I don't see any issue on GitHub either.

Pentarctagon:
Were the Brawler's stats changed? It's not obvious from the diff.

Pentarctagon:
Ah, alright - I was looking at the overall changes in the "Files changed" tab. So the stats haven't changed, the event was just taken out of the unit type, correct?

CelticMinstrel:
Correct, the event is now registered in `stun.lua` and will also handle a custom `[stun]` ability.

I know you mentioned that it needs to use the old events API, but @stevecotton noted that feeding also still uses the new API because it doesn't matter if it breaks undo. Since these are attack and side turn end events, it also doesn't matter if they break undo (attacks and side turn are already not undoable), so I used the new event API.

PR #7324, opened 2023-01-25
Move the Dwarvish Miner unit to core

CelticMinstrel:
This allows addons to safely use it without having to copy the files into their own addon directory.

Also changes the tracking of what they're carrying to be a status instead of a role, which should be more portable to other use-cases.


knyghtmare (approved):
Arigatou Gosaimasu. 
I was planning on using them for a UMC, so I dont have to copy them over anymore.

Pentarctagon:
Was the Miner's stats changed? It's not obvious from the diff.

knyghtmare:
> Was the Miner's stats changed?

Yes, -1 MP when checking the file.

CelticMinstrel:
The diff shows -1 MP too.

Pentarctagon:
It looks like the SoF miner has 4 movement while the SotA miner has 5 movement.

nemaara:
Need to replay SotA to see, it's one of the campaigns I'm not quite as familiar with so I kinda forgor.

spixi:
> Need to replay SotA to see, it's one of the campaigns I'm not quite as familiar with so I kinda forgor.

SotA could just add the quick trait or inherit the unit like it is done with the bat line. Btw: Does anyone know why there is an inherited bat line? The only difference seems to be the standing animation, which is a little odd. Or is it only for compatibility to avoid that older save files break?

knyghtmare:
In case of SotA, I dont think 5 MP Dwarf Miner is required. The PR is just fine as its own.

CelticMinstrel:
Yeah, even if it _is_ required they can just be given the quick trait… but we still need to confirm if that's the case and give the trait if so.

data/core/units/dwarves/Miner.cfg
```diff
@@ -27,21 +27,20 @@
...
         [if]
             [filter]
-                role=loaded
-                [or]
-                    role=has_coal
-                [/or]
-                [or]
-                    role=has_gold
-                [/or]
+                status=has_coal,has_gold
```
stevecotton, line 33: This seems to only support two types of resource, and will need some workaround if a scenario has miners picking up crystal (#7378). That workaround might be as simple as "we use has_coal to mean crystal", but is there a nicer way to incorporate it into the unit?

CelticMinstrel: I was thinking about this too a little while ago. As it is right now I don't think there's a reason to call the status "has_xxx" instead of "loaded" or similar; however, I believe I chose this with the thought that the two statuses might have different graphics at some point in the future. But I'm not sure if there's a nice way to handle this generically in a way that would support other resources…

knyghtmare: maybe add a generalized "has_fullbag"?

CelticMinstrel: That falls under this:

> "loaded" or similar

On the other hand, adding it as an exclusive to the existing two seems kind of pointless, as it doesn't actually help someone who wants them to be loaded with a different resource if the sprites actually reflect the resource being carried.

nemaara (approved):

knyghtmare:
Can this be merged?

nemaara:
Yeehaw go ahead.

PR #7328, opened 2023-01-25
CI: Make the Windows CI builds use separate caches

stevecotton:
Looking at recent build failures, one of Debug or Release is using the prebuilt vcpkg from the cache, and the other one is detecting it but then deciding to rebuild many of the packages.

The build logs circa 8bd8c53f9887ae2440b9a4d29ec55655de08b519 are no longer accessible, only the success/failure and time taken are still accessible. Looking at build timings around those changes is inconclusive, however my guess is that the CI is noticing that CFG is different between the two builds, and that's forcing a rebuild for one of them.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -267,7 +267,7 @@ jobs:
           path: |
             D:/a/wesnoth/vcpkg
             D:/a/wesnoth/wesnoth/vcpkg_installed
-          key: win-cache-master-0011
+          key: win-cache-master-${{ matrix.cfg }}-0012
 
       - name: Add msbuild to PATH
         uses: microsoft/setup-msbuild@v1.1.3


stevecotton:
The line that I'm trying to avoid appearing in the logs is "The following packages will be rebuilt:" as line 9 of the "Run cmake" section.

Looking at the recent successes, it seems that the caching eventually starts working. I'm wondering if it's a race condition between the Debug and Release builds, where we only get a full cache when this happens:
1. Cache name changes
2. Both builds start at the same time
3. They find there's no cache
4. At least one completes successfully and uploads
5. Both builds start at the same time
6. Both find there's a cache containing a Debug build (or the other way round, but for this example assume Debug succeeded at step 4)
7. The Release build completes successfully and uploads a cache containing both builds

If that's what's happening, doing a solo rebuild of a failed Release build should get an image with both builds, assuming that there isn't a race condition with another build. This PR still seems worthwhile, to fix the race condition at the expense of using about 1.4GB of cache space.

stevecotton:
Checking the logs, there are two variations on this line depending on whether a cache already existed:
* "The following packages will be built and installed:"
* "The following packages will be rebuilt:"

Pentarctagon:
I was going to object, however it turns out that the maximum disk space available to the cache has been [increased from 5 GBs to 10 GBs](https://github.blog/changelog/2021-11-23-github-actions-cache-size-is-now-increased-to-10gb-per-repository/) somewhat recently. If it were still 5 GBs, 4 * ~1.2 would be very close to the old 5 GB limit.

stevecotton:
Ah, I saw it was 10GB. It being 5GB before would explain why it was converted to sharing the cache before.

soliton- (approved):
I don't know much about the cache behaviour but this makes sense to me.

PR #7329, opened 2023-01-25
Add the Elvish Hunter line to mainline

CelticMinstrel:
These are sort of a relic from the pre-Quenoth desert elves, only converted back to a wood elf. I think it's a nice flavour unit to have for some campaigns.


CelticMinstrel:
I converted it to draft because I noticed the unit files still need updating.

data/core/images/units/elves-wood/prowler.cfg
```diff
@@ -0,0 +1,131 @@
+#textdomain wesnoth-celmin-aoa
+
+[unit_type]
```
knyghtmare, line 3: Why the heck is CFG file inside images???

src/achievements.hpp
```diff
@@ -107,4 +107,4 @@ class achievements
 	void process_achievements_file(const config& cfg, const std::string& content_source);
 
 	std::vector<achievement_group> achievement_list_;
-};
```
knyghtmare, line 110: this seems very unrelated to the PR

src/gui/dialogs/achievements_dialog.cpp
```diff
@@ -122,4 +122,4 @@ void achievements_dialog::set_achievements_content()
 	}
 }
 
-} // namespace gui2::dialogs
```
knyghtmare, line 125: how is this related to the Elvish Hunter-line PR?

knyghtmare (commented):
Seems okay, apart from that cfg file making into the images/units/elves-wood directory

data/core/images/units/elves-wood/prowler.cfg
CelticMinstrel: Huh? I did accidentally do this but I thought I'd already deleted it. Oops!

src/gui/dialogs/achievements_dialog.cpp
CelticMinstrel: Looks like it got picked up from the fix-whitespace script.

src/achievements.hpp
CelticMinstrel: Guessing it got picked up from the fix-whitespace script.

knyghtmare (approved):

doofus-01:
I've got no issue with the concept and do understand the motivation, but these are old frankensprites.  I'd need to see a little more effort before I can approve.

Whatever way this gets resolved will be tied to all the other things currently being proposed.

cooljeanius (approved):

cooljeanius:
I hope Elvish-Hunter doesn't mind my requesting a review from him just based on his name... 😃

stevecotton (dismissed):
> I've got no issue with the concept and do understand the motivation, but these are old frankensprites. I'd need to see a little more effort before I can approve.

Marking this as "changes requested" because otherwise it shows up in the PR list as "approved". When @doofus-01's comment is addressed, please feel free to remove me from the reviewers list.

data/core/units/elves/Hunter.cfg
```diff
@@ -0,0 +1,212 @@
...
+    [/movement_costs]
+    # These guys have a better defense in swamp water and sands
+    [defense]
+        swamp_water=60
+        cave=50
+        mountains=30
+        hills=40
+        sand=60
+    [/defense]
+    description=_"Elvish hunters are specialized in trapping unwary opponents from a safe distance to ease the task of the swordsmen."
```
stevecotton, line 30: For a mainline unit, I'd expect a longer text. Maybe something about not all groups having shamans, or about the trappers having more movement than the shamans, maybe that the hunters aren't too bad in melee.

Probably should be something more gender-neutral than "swordsmen".

CelticMinstrel: I'm not opposed to making it gender-neutral, but it's probably mainly referring to the fighter line which is mechanically all-male.

Jonathan-Kelly:
Honestly, this old unit seems quite redundant with the elvish fighter and elvish shaman. So the effort it would take to update its mediocre looking sprite would probably be better applied to more useful unit designs.

doofus-01 (changes requested):
Needs a portrait and sprites would need to be updated/animated.

stevecotton:
To make the unit more interesting, maybe "their skill at spotting routes to follow prey lets them give hints to adjacent units about the fastest path", and give them tailwind+1.

cooljeanius:
this PR should be updated so that it passes the new copyright checker CI task

cooljeanius:
> this PR should be updated so that it passes the new copyright checker CI task

@CelticMinstrel could you please look into this?

CelticMinstrel:
I don't know where the Elvish Hunter unit originated. As I recall, I got these sprites from either IftU or AtS, but I do not know where Iris obtained them before that. There used to be a hunter in UtBS, but that was a "brown" version of this one. I'm not sure exactly what relationship the UtBS sprites have with these ones; either one could've been based on the other.

Given that it's pretty close to the archer, I'd guess it's probably GPL, but I don't know for sure.

doofus-01:
The images would need some changes/additions anyways, so no need to worry about the CSV for now.

data/core/units/elves/Hunter.cfg
TheShadowOfHassen: If this is still a problem, I can write one. You'd just have to tell me what it needs to say.  (If the unit is even going to be added)

cooljeanius:
> I don't know where the Elvish Hunter unit originated. As I recall, I got these sprites from either IftU or AtS, but I do not know where Iris obtained them before that.

I think she drew them herself:
https://forums.wesnoth.org/viewtopic.php?p=255787&hilit=Elvish+Hunter#p255787
https://forums.wesnoth.org/viewtopic.php?p=288729&hilit=Elvish+Hunter#p288729
https://forums.wesnoth.org/viewtopic.php?p=290636&hilit=Elvish+Hunter#p290636
@irydacea can you confirm?

knyghtmare:
One thing I agree with doofus here is that the level 2 and level 3 need the minimal animations done. They are just the base image for now. However, given that the current Sylph is also just a base image, I am not sure if this minimal requirement is still in effect.

irydacea:
@cooljeanius Not entirely true. Quoting myself from the first post you linked:

> These are basically rusty frankensteinings of existing art, more spec. Elvish Fighter and Archer.

Thus, technically they are just edits of Jetrel's (IIRC) version of both units.

data/core/units/elves/Hunter.cfg
Gothyoba: Hunters in “Elvish Hunters” needs to be capitalized according to the typography style guide.

Pentarctagon:
What's the status of this PR?

PR #7331, opened 2023-01-26
Make fix_whitespace complain about mixing tabs and spaces

stevecotton:
Ensures that files either use tabs for indentation or spaces for indentation, but don't switch between the two within the same file.

This doesn't fix the whitespace, it's a simple check to flag it up on the assumption that it's better to use an editor or code formatter to clean up the file.

Elsewhere in the CI we use the luacheck tool - while that can detect mixing tabs and spaces in a single line's indent, it doesn't check for inconsistent indentation within a file.

### DIFF
--- a/data/tools/check_mixed_indent
+++ b/data/tools/check_mixed_indent
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+# encoding: utf-8
+"""
+Flag an error if a single file uses both tabs and spaces for indenting.
+
+This does not change files, on the assumption that the mess is better
+handled with a text editor or a code reformatting tool.
+"""
+
+import  argparse, re, sys
+
+def check_indent(file):
+    has_tabs = False
+    with open(file, encoding="utf8") as lines:
+        for line in lines:
+            if re.match("\t", line):
+                has_tabs = True
+                break
+
+        if has_tabs:
+            lines.seek(0)
+            for line in lines:
+                if re.match("    ", line):
+                    print("{file} uses both tabs and spaces for indents".format(file=file))
+                    return False
+
+        # Only one type of indent appears at the start of each line, so this only needs to
+        # check one way round for mixtures on the same line.
+        if has_tabs:
+            mixed_line_re = re.compile("\t+ ")
+        else:
+            mixed_line_re = re.compile(" +\t")
+        lines.seek(0)
+        for line in lines:
+            if mixed_line_re.match(line):
+                print("{file} indents with tabs and spaces on the same line".format(file=file))
+                return False
+
+    return True
+
+if __name__ == "__main__":
+    ap = argparse.ArgumentParser(usage=__doc__)
+    ap.add_argument("files", metavar="string", nargs="*", help="Read input from these files")
+    options = ap.parse_args()
+
+    all_files_ok = True
+
+    for file in options.files:
+        try:
+            if not check_indent(file):
+                all_files_ok = False
+        except UnicodeDecodeError:
+            all_files_ok = False
+            print("{file}: invalid utf-8".format(file=file))
+        except:
+            all_files_ok = False
+            print("{file}: raised exception {error}".format(file=file, error=sys.exc_info()[0]))
+
+    if not all_files_ok:
+        sys.exit(1)
+
+    sys.exit(0)
--- a/utils/CI/fix_whitespace.sh
+++ b/utils/CI/fix_whitespace.sh
@@ -11,3 +11,5 @@ else
   find src/ -name \*.\[ch\]pp -print0 | xargs -0 sed -i '' 's/[[:blank:]]*$//'
   find data/ -name \*.lua -print0 | xargs -0 sed -i '' 's/[[:blank:]]*$//'
 fi
+
+find data/ -name \*.lua -print0 | xargs -0 data/tools/check_mixed_indent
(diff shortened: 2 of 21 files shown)


stevecotton:
I'm not fixing up the .cpp or .hpp files in this commit (and this doesn't scan them).

I'm also not scanning the Lua files in src/modules/lua/testes, and one of them is in latin-1 encoding (which makes this script error out).

utils/CI/fix_whitespace.sh
```diff
@@ -7,7 +7,9 @@ make -C data/tools reindent
 if [[ "$OS" == "Linux" ]]; then
   find src/ -name \*.\[ch\]pp -print0 | xargs -0 sed -i 's/[[:blank:]]*$//'
   find data/ -name \*.lua -print0 | xargs -0 sed -i 's/[[:blank:]]*$//'
+  find data/ -name \*.lua -print0 | xargs -0 data/tools/check_mixed_indent
 else
   find src/ -name \*.\[ch\]pp -print0 | xargs -0 sed -i '' 's/[[:blank:]]*$//'
   find data/ -name \*.lua -print0 | xargs -0 sed -i '' 's/[[:blank:]]*$//'
+  find data/ -name \*.lua -print0 | xargs -0 data/tools/check_mixed_indent
```
Pentarctagon, line 14: This can be present once outside of the if/else I think? The if/else exists because there's a difference in the `sed` syntax between Linux and macOS, but this is a python script.

Pentarctagon:
Adding Elvish_Hunter for the new python script.

stevecotton:
In the latest push:
* Moved outside of the if/else in fix_whitespace.sh
* Added a try/catch so that files with invalid utf-8 causes the script to print an error but continue on the next file

stevecotton:
For anyone wondering about C++ files, none mix spaces and tabs on a single line, but 48 of them use both spaces and tabs for indentation on separate lines.

<details>
<summary>List of files</summary>

* src/units/udisplay.hpp
* src/units/abilities.cpp
* src/units/udisplay.cpp
* src/game_events/entity_location.cpp
* src/teambuilder.cpp
* src/formula/function.hpp
* src/game_version.cpp
* src/pathutils.cpp
* src/server/common/resultsets/tournaments.cpp
* src/server/common/resultsets/tournaments.hpp
* src/server/common/resultsets/game_history.cpp
* src/server/common/resultsets/ban_check.hpp
* src/server/common/resultsets/ban_check.cpp
* src/server/common/resultsets/game_history.hpp
* src/server/wesnothd/player.cpp
* src/wml_separators.hpp
* src/editor/action/action_base.hpp
* src/editor/palette/location_palette.cpp
* src/editor/palette/palette_manager.cpp
* src/editor/palette/unit_palette.cpp
* src/editor/palette/item_palette.cpp
* src/image_modifications.cpp
* src/commandline_options.cpp
* src/display.cpp
* src/ai/composite/component.cpp
* src/ai/default/attack.cpp
* src/desktop/windows_tray_notification.hpp
* src/desktop/apple_notification.hpp
* src/xBRZ/xbrz.cpp
* src/pathfind/pathfind.cpp
* src/pathfind/pathfind.hpp
* src/pathfind/astarsearch.cpp
* src/whiteboard/utility.hpp
* src/gui/widgets/generator_private.hpp
* src/gui/widgets/scrollbar_container.hpp
* src/gui/dialogs/modal_dialog.hpp
* src/gui/core/register_widget.hpp
* src/filesystem.hpp
* src/movetype.cpp
* src/tests/utils/auto_parameterized.hpp
* src/tests/test_lexical_cast.cpp
* src/widgets/button.cpp
* src/actions/create.hpp
* src/actions/attack.cpp
* src/actions/undo.cpp
* src/actions/create.cpp
* src/actions/vision.cpp
* src/pathutils.hpp

</details>

CelticMinstrel:
Looking at the C++ files, the indentation convention for wrapped function definitions would require a run of tabs followed by a run of spaces.

I actually don't like this indentation convention, mind you… I think I'd prefer not to wrap functions at all, but even if they're wrapped I don't like aligning the indentation to the opening parenthesis.

data/tools/check_mixed_indent
```diff
@@ -0,0 +1,59 @@
...
+if __name__ == "__main__":
+    ap = argparse.ArgumentParser(usage=__doc__)
+    ap.add_argument("files", metavar="string", nargs="*", help="Read input from these files")
+    options = ap.parse_args()
+
+    all_files_ok = True
+
+    for file in options.files:
+        try:
+            all_files_ok &= check_indent(file)
```
Elvish-Hunter, line 48: In this line there's `&=`, which is a bitwise operator.
It'd be better to use this:

    if not check_indent(file):
        all_files_ok = False

data/tools/check_mixed_indent
```diff
@@ -0,0 +1,59 @@
...
+
+This does not change files, on the assumption that the mess is better
+handled with a text editor or a code reformatting tool.
+"""
+
+import  argparse, codecs, re, sys
+
+def check_indent(file):
+    has_tabs = False
+    with codecs.open(file, "r", "utf8") as lines:
```
Elvish-Hunter, line 14: The regular `open` function allows to specify an encoding. In this case, it should be enough and the `codecs` module could be removed from the script.

Elvish-Hunter (changes requested):

data/tools/check_mixed_indent
stevecotton: Will change to regular `open`.

I found some new behavior when testing after changing to the regular `open`, but it must have been there when using `codecs` too. Each subsequent `for line in lines:` needs a `lines.seek(0)` first, otherwise it continues from the place that the previous `for` had left it in. With that in, found 1 file that had tabs and spaces on the same line.

Elvish-Hunter (approved):

PR #7333, opened 2023-01-27
Add functionality for achievements to be partially complete.

Pentarctagon:
Instead of being either complete or incomplete, achievements can now specify a value at which they will be considered complete. For such achievements that are not yet complete, a progress bar is added to the achievements dialog showing how close to completion the achievement is.
img

### DIFF
--- a/data/campaigns/tutorial/achievements.cfg
+++ b/data/campaigns/tutorial/achievements.cfg
@@ -2,5 +2,12 @@
 [achievement_group]
     display_name=_"Tutorial"
     content_for=tutorial
-    {ACHIEVEMENT "completed" _"Complete the Tutorial" _"Complete all scenarios of the Tutorial campaign."}
+    [achievement]
+        id="completed"
+        name=_"Completed the Tutorial"
+        description=_"Completed all scenarios of the Tutorial campaign."
+        icon="data/core/images/icons/potion_red_small.png"
+        icon_completed="data/core/images/icons/potion_green_small.png"
+        max_progress=2
+    [/achievement]
 [/achievement_group]
--- a/src/achievements.cpp
+++ b/src/achievements.cpp
@@ -107,8 +107,11 @@ achievement_group::achievement_group(const config& cfg)
 
 		if(id.empty()) {
 			ERR_CONFIG << content_for_ + " achievement missing id attribute:\n" << ach.debug();
+		} else if(id.find(',') != std::string::npos) {
+			ERR_CONFIG << content_for_ + " achievement missing id " << id << " contains a comma, skipping.";
+			continue;
 		} else {
-			achievements_.emplace_back(ach, preferences::achievement(content_for_, id));
+			achievements_.emplace_back(ach, preferences::achievement(content_for_, id), preferences::progress_achievement(content_for_, id));
 		}
 	}
 }
--- a/src/preferences/general.cpp
+++ b/src/preferences/general.cpp
@@ -1047,10 +1047,16 @@ void set_achievement(const std::string& content_for, const std::string& id)
 		if(ach["content_for"].str() == content_for)
 		{
 			std::vector<std::string> ids = utils::split(ach["ids"]);
-			if(std::find(ids.begin(), ids.end(), id) == ids.end())
+
+			if(ids.empty())
+			{
+				ach["ids"] = id;
+			}
+			else if(std::find(ids.begin(), ids.end(), id) == ids.end())
 			{
 				ach["ids"] = ach["ids"].str() + "," + id;
 			}
+			ach.remove_children("in_progress", [&id](config cfg){return cfg["id"].str() == id;});
 			return;
 		}
 	}
@@ -1062,5 +1068,59 @@ void set_achievement(const std::string& content_for, const std::string& id)
 	prefs.add_child("achievements", ach);
 }
 
+int progress_achievement(const std::string& content_for, const std::string& id, int limit, int max_progress, int amount)
+{
+	if(achievement(content_for, id))
+	{
+		return -1;
+	}
+
+	for(config& ach : prefs.child_range("achievements"))
+	{
+		// if achievements already exist for this content and the achievement has not already been set, add it
+		if(ach["content_for"].str() == content_for)
+		{
+			// check if this achievement has progressed before - if so then increment it
+			for(config& in_progress : ach.child_range("in_progress"))
+			{
+				if(in_progress["id"].str() == id)
+				{
+					in_progress["progress_at"] = std::clamp(in_progress["progress_at"].to_int() + amount, 0, std::min(limit, max_progress));
+					return in_progress["progress_at"].to_int();
+				}
+			}
+
+			// else this is the first time this achievement is progressing
+			if(amount != 0)
+			{
+				config set_progress;
+				set_progress["id"] = id;
+				set_progress["progress_at"] = std::clamp(amount, 0, std::min(limit, max_progress));
+
+				config& child = ach.add_child("in_progress", set_progress);
+				return child["progress_at"].to_int();
+			}
+			return 0;
+		}
+	}
+
+	// else not only has this achievement not progressed before, this is the first achievement for this achievement group to be added
+	if(amount != 0)
+	{
+		config ach;
+		config set_progress;
+
+		set_progress["id"] = id;
+		set_progress["progress_at"] = std::clamp(amount, 0, std::min(limit, max_progress));
+
+		ach["content_for"] = content_for;
+		ach["ids"] = "";
+
+		config& child = ach.add_child("in_progress", set_progress);
+		prefs.add_child("achievements", ach);
+		return child["progress_at"].to_int();
+	}
+	return 0;
+}
 
 } // end namespace preferences
--- a/src/preferences/general.hpp
+++ b/src/preferences/general.hpp
@@ -277,7 +277,32 @@ namespace preferences {
 	sort_order::type addon_manager_saved_order_direction();
 	void set_addon_manager_saved_order_direction(sort_order::type value);
 
+	/**
+	 * @param content_for The achievement group the achievement is part of.
+	 * @param id The ID of the achievement within the achievement group.
+	 * @return True if the achievement exists and is completed, false otherwise.
+	 */
 	bool achievement(const std::string& content_for, const std::string& id);
+	/**
+	 * Marks the specified achievement as completed.
+	 *
+	 * @param content_for The achievement group the achievement is part of.
+	 * @param id The ID of the achievement within the achievement group.
+	 */
 	void set_achievement(const std::string& content_for, const std::string& id);
+
+	/**
+	 * Increments the achievement's current progress by @a amount if it hasn't already been completed.
+	 * If you only want to check the achievement's current progress, then omit the last three arguments.
+	 * @a amount defaults to 0, which will result in the current progress value being returned without being changed (x + 0 == x).
+	 *
+	 * @param content_for The id of the achievement group this achievement is in.
+	 * @param id The id for the specific achievement in the achievement group.
+	 * @param limit The maximum value that a specific call to this function can increase the achievement progress value.
+	 * @param max_progress The value when the achievement is considered completed.
+	 * @param amount The amount to progress the achievement.
+	 * @return The achievement's current progress, or -1 if it has already been completed.
+	 */
+	int progress_achievement(const std::string& content_for, const std::string& id, int limit = 999999, int max_progress = 999999, int amount = 0);
 
 } // end namespace preferences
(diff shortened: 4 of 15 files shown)


Vultraz:
Screenshot?

Pentarctagon:
Oh, right, added. I'm not sure about the progress bar, but I couldn't find a way to adjust its side either.

CelticMinstrel:
Hmmm… this is perfect for an achievement consisting of a series of steps that must be completed in order (your tutorial example), or a fungible counter like "kill 20 wolves" (where you wouldn't use a limit key in `[progress_achievement]`). However, it's not so great for a "collect them all" type of achievement. It can certainly be used for that, but the scenario would still need to track the progress separately.

How much work do you think it would be to support a "bit field" progress mode, in addition to a straight numeric mode?

…although if the goal is to require collecting them all in a single playthrough, would even that be sufficient? Hmm…

---

Also, just double-checking, but… achievements with no progress configured will still work just like before and not show the progress bar, right?

Pentarctagon:
> Hmmm… this is perfect for an achievement consisting of a series of steps that must be completed in order (your tutorial example), or a fungible counter like "kill 20 wolves" (where you wouldn't use a limit key in [progress_achievement]). However, it's not so great for a "collect them all" type of achievement. It can certainly be used for that, but the scenario would still need to track the progress separately.
> 
> How much work do you think it would be to support a "bit field" progress mode, in addition to a straight numeric mode?
> 
> …although if the goal is to require collecting them all in a single playthrough, would even that be sufficient? Hmm…

That sounds fairly complicated since that'd essentially be a sub-group of achievements within a single achievement - you'd need a way for the author to specify which piece of the achievement to complete and then presumably a way to indicate on the UI which pieces are complete vs not complete.

I guess it'd basically be like trying to nest an achievement group within a single achievement?

> Also, just double-checking, but… achievements with no progress configured will still work just like before and not show the progress bar, right?

Correct - any achievement that doesn't set a `max_progress` attribute defaults to 0, and any achievement where the max progress is 0 doesn't show a progress bar.

CelticMinstrel:
> then presumably a way to indicate on the UI which pieces are complete vs not complete.

Well, this part could just be a counter.

Vultraz (changes requested):
i would merge the counter into either the description or title and center the icon vertically in the left space

CelticMinstrel:
> i would merge the counter into either the description or title and center the icon vertically in the left space

My suggestion would be to right-align it in the same row as the title.

Pentarctagon:
> Well, this part could just be a counter+progress bar, same as what you've done here. It's just representing "number of bits set" rather than "current value".

How would counting the number of bits set be any different than a plain numeric counter?

CelticMinstrel:
> How would counting the number of bits set be any different than a plain numeric counter?

With a plain numeric counter, it looks kinda like this:

* max_progress is 3
* When collecting the each item, increment progress
* Display as 2/3 when progress value is 2

With a bit set, it looks like this:

* max_progress is 0x111 (ie 7)
* When collecting the first item, increment progress by 1 without carry (ie bitwise or)
* When collecting the second item, increment progress by 2 without carry
* When collection the second item, increment progress by 4 without carry
* Display as 2/3 when progress value is 0x011 (ie 3)

I think it wouldn't need to change the underlying data at all, only how it's modified and displayed.

That said, I'm also not sure this actually solves the problem I want… even with this I would probably have to track achievements in variables, since I'm looking for "collect them all in a single playthrough" rather than "collect them all across several playthroughs". Still… I guess this _could_ still be useful for a "obtain all endings" type of achievement.

Pentarctagon:
I guess I can kind of see the use? Though it sounds like it could be more easily accomplished by adding an achievement per thing and then have another achievement that unlocks when all the other achievements are completed, unless there'd be a ton of those as individual achievements.

src/achievements.cpp
```diff
@@ -91,6 +64,39 @@ void achievements::process_achievements_file(const config& cfg, const std::strin
...
 }
 
+/**
+ * Reads the mainline achievements.cfg and then all the achievements of each installed add-on.
+ *
+ * This is intentionally handled separately from other WML loading so that:
+ * a) All achievements and their status are able to be displayed on the main menu right after Wesnoth starts and regardless of which add-ons are active.
+ * b) Add-ons can add additional achievements to other content, whether UMC or mainline. For example, a modification that adds more achievements for mainline campaigns.
+ *
+ * NOTE: These are *not* in any way related to Steam achievements!
```
stevecotton, line 74: This line should probably still be in the class / constructor doc instead of moving to reload().

src/game_config_manager.cpp
```diff
@@ -211,8 +211,7 @@ void game_config_manager::load_game_config(bool reload_everything, const game_cl
 			game_config::config_cache_transaction main_transaction;
```
stevecotton, line 212: Remember that there's still whitespace to fix on line 212.

src/achievements.cpp
Pentarctagon: Updated.

src/game_config_manager.cpp
Pentarctagon: Updated.

CelticMinstrel:
> Though it sounds like it could be more easily accomplished by adding an achievement per thing and then have another achievement that unlocks when all the other achievements are completed, unless there'd be a ton of those as individual achievements.

That actually does sound like a good way to do it, provided it makes sense to have a separate achievement for each subtask.

In at least one of my ideas that doesn't make sense (it's collecting several items that are not distinct from one another, but that are scattered around in various places), but I do want it to be "collect them all in one session".

Pentarctagon:
Alright, I believe everything that's been brought up has been addressed now.

> That actually does sound like a good way to do it, provided it makes sense to have a separate achievement for each subtask.
> 
> In at least one of my ideas that doesn't make sense (it's collecting several items that are not distinct from one another, but that are scattered around in various places), but I do want it to be "collect them all in one session".

That sounds like it can be done with the current PR? You could track the items that are picked up and then have something like `limit=$found_items_count`.

CelticMinstrel:
> That sounds like it can be done with the current PR? You could track the items that are picked up and then have something like `limit=$found_items_count`.

Hmm… given the "in one playthrough" requirement and the "collect them all" requirement, that sounds like it works with respect to getting the achievement. However, the progress bar might be a little odd.

If on one playthrough you find the first and fourth item, and on another playthrough you found the second and third item, then progress will be at 2 out of 4.

Pentarctagon:
Given achievements have no concept of a playthrough, that'd be expected.

knyghtmare:
Will the counter reset for MP scenarios for each playthrough?

CelticMinstrel:
Thoughts on the latest screenshot, mirrored from IRC/Discord:

* The progress counter is fine as-is, but I think it might be even better if it was right-aligned in the same row as the title.
* I'd like a bit more space between the progress bar and the bottom edge of the selection.
* Technically unrelated to the current change, but it would be nice to have a counter in the bottom left that shows a counter of completed and total achievements in the current group.
* Also it would be nice if the UI would remember the last group selected.

One other thing I've noticed is that, with the current implementation, hidden achievements appear to add nothing at all to the system. They could be removed and nothing would change – an achievement that wants to be hidden would just have to set name and description to ??? and a hint respectively, and put the real name and description in the _completed fields.

knyghtmare:
Two more points which were discussed:
(actually might out of PR scope and can be moved to another issue maybe?)
- When the notification pop-up is displayed, there can be a built-in sound (fanfare-short.wav) playing
- Preferably this sound-playing feature can be disabled in preferences

stevecotton:
> When the notification pop-up is displayed, there can be a built-in sound (fanfare-short.wav) playing

That assumes that all achievements happen at happy times - what about one that triggers at the end of a sad chapter in the story?

Pentarctagon:
@CelticMinstrel 
> The progress counter is fine as-is, but I think it might be even better if it was right-aligned in the same row as the title.

The title is left aligned though?

> I'd like a bit more space between the progress bar and the bottom edge of the selection.

Alright, will update that.

> Technically unrelated to the current change, but it would be nice to have a counter in the bottom left that shows a counter of completed and total achievements in the current group.

That's been added, I just haven't updated the screenshot here yet.

> Also it would be nice if the UI would remember the last group selected.

Technically this has been added, but doesn't work yet for reasons I need to investigate.

> One other thing I've noticed is that, with the current implementation, hidden achievements appear to add nothing at all to the system. They could be removed and nothing would change – an achievement that wants to be hidden would just have to set name and description to ??? and a hint respectively, and put the real name and description in the _completed fields.

I plan to remove these in a separate PR.

@knyghtmare 
> When the notification pop-up is displayed, there can be a built-in sound (fanfare-short.wav) playing

I'm planning to introduce a `sound=` attribute in a separate PR.

knyghtmare:
> I'm planning to introduce a sound= attribute in a separate PR.

That works

CelticMinstrel:
> That assumes that all achievements happen at happy times - what about one that triggers at the end of a sad chapter in the story?

The sound to play could be an option in `[set_achievement]` and `[progress_achievement]`?  That does make the preferences setting difficult, but we could reopen the question of exposing preferences to Lua…

> The title is left aligned though?

I'm saying put the title and counter in separate blocks with different alignment.

> I plan to remove these in a separate PR.

I guess I'm okay with removing them, but I want to note that the other option is to make them actually be completely hidden as you originally planned, but note how many hidden achievements there are in the bottom left counter. Of course, that would render the hint text irrelevant, so I dunno what the best course of action is here. (Plus someone else specifically requested they not be truly hidden.)

data/gui/widget/progress_bar_default.cfg
```diff
@@ -56,3 +56,56 @@
 	[/resolution]
 
 [/progress_bar_definition]
+
+[progress_bar_definition]
```
CelticMinstrel, line 60: Should this go in a separate file? I think the usual thing is one definition per file, though there are probably other exceptions to that too…

Pentarctagon: Updated.

Pentarctagon:
Updated the screenshot in the first post.

> I'm saying put the title and counter in separate blocks with different alignment.

I prefer it where it is, personally. If vultraz says it needs to be aligned differently though then I'll move it.

src/preferences/general.hpp
```diff
@@ -277,7 +277,32 @@ namespace preferences {
...
+	 * Increments the achievement's current progress by @a amount if it hasn't already been completed.
+	 * If you only want to check the achievement's current progress, then omit the last three arguments.
+	 * @a amount defaults to 0, which will result in the current progress value being returned without being changed (x + 0 == x).
+	 *
+	 * @param content_for The id of the achievement group this achievement is in.
+	 * @param id The id for the specific achievement in the achievement group.
+	 * @param limit The maximum value that a specific call to this function can increase the achievement progress value.
+	 * @param max_progress The value when the achievement is considered completed.
+	 * @param amount The amount to progress the achievement.
+	 * @return The achievement's current progress, or -1 if it has already been completed.
```
stevecotton, line 304: What if it doesn't exist?

Pentarctagon: What if what doesn't exist?

data/campaigns/tutorial/achievements.cfg
```diff
@@ -2,5 +2,12 @@
 [achievement_group]
     display_name=_"Tutorial"
     content_for=tutorial
-    {ACHIEVEMENT "completed" _"Complete the Tutorial" _"Complete all scenarios of the Tutorial campaign."}
+    [achievement]
+        id="completed"
+        name=_"Complete the Tutorial"
```
stevecotton, line 7: Wouldn't block this PR about it, but the UI when the player gets this achievement feels wrong. They've played through two scenarios of being told what to do, and at the end a dialog puts up with "Complete the Tutorial" in the imperative. Either that dialog should have an "Achievement completed" header on it, or this text should be past tense such as "Tutorial completed".

stevecotton (approved):

src/preferences/general.hpp
stevecotton: If there's no achievement registered for that `content_for + id` pair.

Pentarctagon: A lua error would be logged.

data/campaigns/tutorial/achievements.cfg
Pentarctagon: Updated.

src/preferences/general.cpp
```diff
@@ -1062,5 +1068,61 @@ void set_achievement(const std::string& content_for, const std::string& id)
...
+			ach["ids"] = "";
+
+			config& child = ach.add_child("in_progress", set_progress);
+			prefs.add_child("achievements", ach);
+			return child["progress_at"].to_int();
+		}
+		return 0;
+	}
+
+	return -1;
```
stevecotton, line 1125: Most of this function could have one less indent, by changing the block around line 1075 to be
```
if](completed) {
    return -1;
}
````

This function has other early returns, so having an early return right at the top seems reasonable.

stevecotton (approved):

src/preferences/general.cpp
Pentarctagon: Updated.

Issue #7159, opened 2022-11-22
modify_unit recall_cost vector::_M_range_check


ProditorMagnus:
### Game and System Information

1.16.0

### Description of the bug

When running this action with side=2 or empty filter, chat area is filled with errors. When running with side=1, there are no visible errors. However, action of increasing recall cost works even despite errors.

### Steps to reproduce the behavior

Reproduced on 2p-clearing gushes with wesnoth.require("wml-utils").handle_event_commands(wml.load("~add-ons/EventLoader/action.cfg"))
```
[modify_unit]
    [filter]
		side=2
    [/filter]
    [object]
        [effect]
            apply_to="recall_cost"
            increase=10
            times="per level"
        [/effect]
    [/object]
[/modify_unit]
```

### Expected behavior

No visible errors.

### Additional context

https://discord.com/channels/231976805987385345/442775044590927873/1044512035431530506 has screenshot

Wedge009:
Possible to attach the screen directly?

stevecotton:
I'm not debugging this (other things to do), but my intuition is that there's a 1-based `side_` being used as the index of a 0-based array on line 2222 of unit.cpp:

https://github.com/wesnoth/wesnoth/blob/5ffa031e8e79b0e1e4b1cfb02302030ff6bb02b6/src/units/unit.cpp#L2222

ProditorMagnus:
![image](https://user-images.githubusercontent.com/7848963/203490916-49843596-1ea1-4d4d-b6b3-2f1c6065c5e2.png)

redone9211:
i am intrested in contributing in this issue, can you let me know how i can solve this issue.

CelticMinstrel:
Should be relatively easy, just add a range check, probably in `intf_add_modification`, and raise a proper Lua error that correctly describes the issue.

stevecotton:
@CelticMinstrel the bug here is that `[effect]apply_to="recall_cost"` is triggering that error, not the range check itself.

redone9211:
if possible can this issue be assigned to me, then we can figure out how exactly it can be resolved.

CelticMinstrel:
I suppose you could say there are multiple bugs, but an internal C++ error should not be exposed to Lua. Even if `[effect]apply_to="recall_cost"` throws an exception, it should be turned into something readable as a Lua error.

stevecotton:
[...]

PR #7339, opened 2023-01-29
UTBS Minion and Spire XP reduced

Toranks:
imagen

All the alien units have the same experience, resulting in this disproportion, and it looks pretty ugly.

I can change the ratio of the experience bar with xp_bar_scaling, but reducing the experience is the best option, it does not affect the balance since they are low health units, it is very almost impossible for them to level up.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Minion.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Minion.cfg
@@ -15,7 +15,7 @@
         cold=90
         arcane=120
     [/resistance]
-    experience=150
+    experience=50
     {AMLA_DEFAULT}
     level=0
     alignment=chaotic
--- a/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Minion.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Minion.cfg
@@ -15,7 +15,7 @@
         cold=90
         arcane=120
     [/resistance]
-    experience=0
+    experience=150
     level=0
     alignment=chaotic
     advances_to=null
--- a/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Minion.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Minion.cfg
@@ -15,8 +15,7 @@
         cold=90
         arcane=120
     [/resistance]
-    experience=50
-    {AMLA_DEFAULT}
+    experience=0
     level=0
     alignment=chaotic
     advances_to=null
--- a/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Spire.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Spire.cfg
@@ -22,7 +22,7 @@
         cave=50
     [/defense]
     movement=1
-    experience=0
+    experience=150
     level=0
     alignment=chaotic
     advances_to=null
--- a/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Spire.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/monsters/Alien_Spire.cfg
@@ -22,8 +22,7 @@
         cave=50
     [/defense]
     movement=1
-    experience=100
-    {AMLA_DEFAULT}
+    experience=0
     level=0
     alignment=chaotic
     advances_to=null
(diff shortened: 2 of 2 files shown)


Toranks:
With the changes
imagen

Pentarctagon:
I think that should've been fixed more generally already with #7293 ?

Toranks:
Partially, that only limits the maximum length. An experience bar many times longer than HP still looks ugly for a "minion" unit that will never advance.
If you really need an unit with a lot of experience, xp_bar_scaling would be another option.

stevecotton:
Should we rename this issue to avoid spoilers? Maybe "UTBS S12: Minion and Spire XP reduced"?

stevecotton:
50 XP seems a bit low, given that in this scenario they might be killing L3 units after the main attacks have damaged them.

I'm in two minds about the long bars - they do communicate clearly that these units aren't intended to level.

CelticMinstrel:
I wonder what happens if you set `experience=0` in the unit type config, along with no advancements at all (including no AMLA)…

Wedge009:
Out of curiosity, I set Grand Knight (final advancement) to `experience=0` and removed `{AMLA_DEFAULT}` and debug-created one in a game. It shows XP as - in the side-bar, and the unit has no XP bar, only HP. Killing a unit appears to have no effect and I didn't see any error messages in the log. So it appears the engine handles this situation fine, no division-by-zero or anything like that.

nemaara (approved):

CelticMinstrel:
The reason I mentioned it is because I thought that, if correctly handled, this might be a good place to do that.

Toranks:
I've tested and what @Wedge009 says works. Changing experience to 0 doesn't work by itself, it only changes experience to 1 and can levelup multiple times with just 1 attack. But if I remove {AMLA_DEFAULT} too, the experience changes to "-" and never levelup.
imagen

stevecotton (changes requested):
If you're relying on edge-case behavior then it's good to have a unit test, and I think it's mandatory in this case. If the behavior changes then the minions might get a levelup and full-heal after each combat, but that wouldn't be noticed at the time of the engine change, and players might assume it's an intended quirk of this scenario.

The current HOWTO for writing unit tests is here: https://github.com/wesnoth/wesnoth/issues/7159#issuecomment-1399414492

Toranks:
I already sent a PR with a test before, I'll use the template.
I imagine changing Alice's and Bob's experience to 0, and deleting Bob's AMLA DEFAULT but not Alice's.
They then fight each other.
The result should be that Alice levels up as many times as Bob's levels, and Bob stays the same.
Do you think a test like this would do the work to cover both cases?

stevecotton:
Harm Bob first, and check that he doesn't heal.

I guess Alice will only go up once, and then have some xp left over afterwards (because the level-up will probably give Alice a higher max xp before the next level).

stevecotton:
> I've tested and what @Wedge009 says works. Changing experience to 0 doesn't work by itself, it only changes experience to 1 and can levelup multiple times with just 1 attack. But if I remove {AMLA_DEFAULT} too, the experience changes to "-" and never levelup.

Thanks for writing the unit test. Based on experimenting with it, it seems that changing `max_experience` isn't necessary - just removing the advancement hides the xp bar and makes the sidebar show "-", even when `:inspect` shows that the unit has 1/42 xp.

If we don't need `max_experience=0`, then I'd prefer to avoid it. Looking in the C++, it looks like would be a division-by-zero error if drawer.cpp circa line 414's `double filled = static_cast<double>(experience) / static_cast<double>(max_experience);` was reachable with `max_experience==0`. There also seems to be a sanity check ensuring that doesn't happen, with `unit::init` enforcing a minimum of 1.

Toranks:
I have a change request here by @stevecotton , which was to do the unit test before relying on the experience behaviour that changes the PR. The unit test was already merged on master, so I assume I don't need to do anything else here, right?

stevecotton:
Based on the tests, removing the `{AMLA_DEFAULT}` hides the xp bars, without needing to reduce the experience. Just in case someone changes the behavior, I'd prefer to have either:

* high experience numbers in UtBS S12, even though they're hidden, or
* a comment in the zero_experience test noting that the behavior is used in UtBS S12.

My reasoning is that someone may change the behavior (and the test) without someone else spotting the bug in S12 afterwards. In either case, it would be good to have a comment in the UtBS files about hiding the XP.

Toranks:
I chose the first option. In case someone used the experience for something with mods or addons. I use the original exp number, which is already quite high.

PR #7341, opened 2023-01-30
Orb related cleanup for preferences and game_config

### LINKED ISSUE #7261: game_config schema: the schema has keys for orb colors, but not the toggles

- **Version:** 1.17.11+dev
- **Downloaded from:** built from source
- **Build info:** current master, with or without PR #7189

### Description of the bug

The validation in data/schema/core/config.cfg is aware of the `*_orb_color` preferences, and the `orb` and `orb_two_color` filenames, but it doesn't have entries for the 6 (+1 in PR #7189) settings below. Should it have?

```
//
// Orb display flags
//
bool show_ally_orb;
bool show_disengaged_orb;
bool show_enemy_orb;
bool show_moved_orb;
bool show_partial_orb;
bool show_status_on_ally_orb;
bool show_unmoved_orb;
```

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

stevecotton:
Assorted cleanup, fixes #7261.

* Add `game_config.cfg`'s `show_*_orb` to the schema
* Rename `show_partly_orb`
* Remove the unused `disengaged_orb_color` preference

The `show_*_orb` keys are supported in game_config.cfg, but currently they're not used in that file, instead falling back to the defaults in the C++. Add them to the schema, but not to `game_config.cfg` itself.

The name `show_partly_orb` was inconsistent - in other places, including the WML that's stored in the user's preferences file, it's called `show_partial_orb`.

Remove leftovers from the original implementation of the disengage orb, when it was a new color for the single-color orb. During the review process it became the two-color design, but these parts were left in.

The value of orb_status_helper::get_orb_color(orb_status::disengaged) is used (both before and after this commit) by the minimap; this commit changes get_orb_color to return the partial_color() directly. The result will be the same as before, because disengage_color() returned the same as partial_color(). Prior to this commit that was done by giving both orbs the same value in game_config.cfg. If the advanced preference to change orbs colors was used, the value chosen for the partial orb was passed to set_disengage_color() too.

### DIFF
--- a/data/game_config.cfg
+++ b/data/game_config.cfg
@@ -49,8 +49,6 @@
 
     [colors]
         ally_orb_color="lightblue"
-        # on the minimap, disengaged uses the same color as partial
-        disengaged_orb_color="brightorange"
         enemy_orb_color="black"
         moved_orb_color="red"
         partial_orb_color="brightorange"
--- a/data/schema/core/config.cfg
+++ b/data/schema/core/config.cfg
@@ -52,7 +52,6 @@
 		name="colors"
 		min=1
 		{SIMPLE_KEY ally_orb_color string}
-		{SIMPLE_KEY disengaged_orb_color string}
 		{SIMPLE_KEY enemy_orb_color string}
 		{SIMPLE_KEY moved_orb_color string}
 		{SIMPLE_KEY partial_orb_color string}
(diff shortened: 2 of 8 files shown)


Vultraz (approved):

CelticMinstrel (approved):
Seems fine? I assume you're asking my input because it touches the schema, but that part definitely has no issues.

PR #7343, opened 2023-01-31
Fix the missing logs issue (closes #7310)

### LINKED ISSUE #7310: `--help` should imply `--no-log-to-file`

- **Version:** 1.17.12+dev
- **Downloaded from:** Git
- **Build info:** a few local changes on top of 99bb65d1b3
- **OS:** Linux

### Description of the bug

Running `wesnoth --help` to remember the exact name of an option prints nothing to stdout or stderr, instead it dumps the help text to the log file. It doesn't give a hint about what it's done, where the log file is, or the name of the `--no-log-to-file` option.

_No response_

### Expected behavior

`--help` should output the text to stdout/stderr.

### Additional context

_No response_

CelticMinstrel:


src/wesnoth.cpp
```diff
@@ -1043,32 +1043,24 @@ int main(int argc, char** argv)
 	auto args = read_argv(argc, argv);
 	assert(!args.empty());
 
+	bool log_redirect = true, terminal_implied = false;
+	// This is optional<bool> instead of tribool because value_or() is exactly the required semantic
+	std::optional<bool> terminal_force;
```
stevecotton, line 1048: These things are complex enough that having documentation for them would be good, even though they're local variables.

src/wesnoth.cpp
```diff
@@ -1099,24 +1091,37 @@ int main(int argc, char** argv)
...
-			native_console_force = false;
+			terminal_force = false;
 		} else if(arg == "--wconsole") {
-			native_console_force = true;
+			terminal_force = true;
 		} else if(arg == "--wnoredirect") {
 			log_redirect = false;
 		}
+#endif
+		if(arg == "--wno-log-to-file") {
```
stevecotton, line 1108: Spurious `w` here?

src/commandline_options.cpp
```diff
@@ -262,6 +262,7 @@ commandline_options::commandline_options(const std::vector<std::string>& args)
 		("log-none", po::value<std::string>(), "sets the severity level of the specified log domain(s) to 'none'. Similar to --log-error.")
 		("log-precise", "shows the timestamps in log output with more precision.")
 		("no-log-to-file", "log output is written to standard error rather than to a file.")
+		("log-to-file", "log output is written to a file even if it would normally go to standard input.")
```
stevecotton, line 265: Maybe "cancels the effect of --no-log-to-file earlier on the command line"?

src/wesnoth.cpp
CelticMinstrel: Yup

src/wesnoth.cpp
CelticMinstrel: I started writing some and realized that I'm not even sure what `log_redirect` is for. It's only used in Windows.

stevecotton: I see `log_redirect` still exists, but it looks as if `terminal_implied`, `file_force` and `terminal_force` ought to cover all the possibilities. What is the remaining purpose of `log_redirect`?

CelticMinstrel: Mostly I was just trying to be careful about touching the Windows-specific side of this because I wasn't quite sure what it was all for. You might be right – perhaps `log_redirect` (and `--wnoredirect`) is now obsolete. I'm not quite sure, so I'd prefer to get confirmation from someone who actually uses Windows.

src/wesnoth.cpp
```diff
@@ -1099,24 +1096,37 @@ int main(int argc, char** argv)
...
+#endif
+		if(arg == "--no-log-to-file") {
+			terminal_force = true;
+		} else if(arg == "--log-to-file") {
+			file_force = true;
+		}
 	}
 
-	if(native_console_force.value_or(native_console_implied)) {
+	const bool log_to_terminal = terminal_force.value_or(terminal_implied);
```
stevecotton, line 1120: The variable is misnamed on Windows - on there it's whether to attach a terminal, but IIUC `--wconsole` without `--wnoredirect` should enable a console without cancelling the log redirection.

stevecotton:
While I'd like to get this merged soon, I feel there's something weird about the logic - it's now right for non-Windows, but the comments and code no longer seem to match what Windows was doing. So if we merge this now, I'm worried that we'll have another PR to fix it for Windows, and maybe ping-pong to fix it again for non-Windows.

@irydacea would you have time to review this, please?

CelticMinstrel:
Yeah, I'm not confident that it's correct for Windows, but I'd also like to get this merged asap so CI becomes useful again.

Pentarctagon:
The CI can also be fixed just by adding `--no-log-to-file` where needed.

CelticMinstrel:
> The CI can also be fixed just by adding `--no-log-to-file` where needed.

It can, but since we need to do this anyway, I think it's better to just do it.

CelticMinstrel:
I think the need to press Enter stems from it not being a console app yet attaching to the invoking console. It's as if on Linux you ended the command with `&` and then followed up with `bg %1`.

Pentarctagon:
> I think the need to press Enter stems from it not being a console app yet attaching to the invoking console. It's as if on Linux you ended the command with & and then followed up with bg %1.

That's done in wesnoth.cpp with `std::cin.get();`.

CelticMinstrel:
> That's done in wesnoth.cpp with `std::cin.get();`.

Ah yes, I guess that would do it too.

Wedge009:
I re-ran the above with `master` on Windows. The only difference I noticed was that `--no-log-to-file` works as expected for `fix_logs` but not `master`.

PR #7344, opened 2023-01-31
Unit tests to evaluate behavior with max_experience=0 and with no advancements

Toranks:
Check the behavior of units when their maximum experience is 0, with or without advancement available. The default engine behavior is that if a unit has any advancement, it can't have `max_experience=0`, so it always changes to `0/1`. If the unit have no advancements, then the experience bar will disappear and show `-` instead of `0/1`

This test is equivalent to `experience=0` on `[unit_type]` too. `[unit] max_experience=` takes by default the value of `[unit_type] experience=`

* API(s) being tested: `[unit] max_experience=0`

* Actions:
Change Alice's and Bob's max_experience to 0. Delete Bob's 'Orcish Warrior' advancement but not Alice's. Harm bob with 1 hitpoint to be sure that he have some damage. They then fight each other.

* Expected end state:
Alice levels up and changes her max_experience to the advancement. Bob don't advances and have 1 point of max experience. The default behaviour is for a unit with max_experience 0 always changes to 1, no matter if they have advancements or not (seen with inspect console).

Note: For some reason I've been forced to use `[object] [effect] apply_to=remove_advancement`, because without `[object]` it doesn't work.

imagen

This PR is complemented with #7339, for which it serves as a test.

### DIFF
--- a/data/test/scenarios/wml_tests/UnitsWML/unit_no_advancements.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/unit_no_advancements.cfg
@@ -7,7 +7,7 @@
 # Harm Alice and Bob with 1 hitpoint to be sure that they have some damage. Delete advancements of both units and give them 999 experience. They then fight each other.
 ##
 # Expected end state:
-# Bob and Alice doesn't level up, keeps their HP, and gains experience. The default behavior should be that a unit will never advances if it has no advancements, but still can gain experience internally (seen with inspect console).
+# Bob and Alice doesn't level up, keeps their HP, and gains experience. The default behaviour should be that a unit will never advances if it has no advancements, but still can gain experience internally (seen with inspect console).
 #####
 {COMMON_KEEP_A_B_UNIT_TEST "unit_no_advancements" (
     [event]
--- a/data/test/scenarios/wml_tests/UnitsWML/unit_no_advancements.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/unit_no_advancements.cfg
@@ -1,0 +1,69 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: [unit] [advancement]
+##
+# Actions:
+# Harm Alice and Bob with 1 hitpoint to be sure that they have some damage. Delete advancements of both units and give them 999 experience. They then fight each other.
+##
+# Expected end state:
+# Bob and Alice doesn't level up, keeps their HP, and gains experience. The default behavior should be that a unit will never advances if it has no advancements, but still can gain experience internally (seen with inspect console).
+#####
+{COMMON_KEEP_A_B_UNIT_TEST "unit_no_advancements" (
+    [event]
+        name=prestart
+        [harm_unit]
+            [filter]
+                id=bob,alice
+            [/filter]
+            amount=1
+        [/harm_unit]
+        [modify_unit]
+            [filter]
+                id=alice,bob
+            [/filter]
+            [object]
+                [effect]
+                    apply_to=remove_advancement
+                    types=Orcish Warrior,Elvish Ranger,Elvish Marksman
+                [/effect]
+            [/object]
+            experience=999
+        [/modify_unit]
+    [/event]
+    [event]
+        name=start
+        [do_command]
+            [attack]
+                weapon=0
+                defender_weapon=0
+                [source]
+                    x,y=5,3
+                [/source]
+                [destination]
+                    x,y=4,3
+                [/destination]
+            [/attack]
+        [/do_command]
+    [/event]
+    [event]
+        name=start
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=alice
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=bob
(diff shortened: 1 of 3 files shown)


data/test/scenarios/wml_tests/UnitsWML/zero_experience.cfg
```diff
@@ -0,0 +1,80 @@
...
+            [/filter]
+            variable=alice
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=bob
+        [/store_unit]
+        {ASSERT ({VARIABLE_CONDITIONAL bob.hitpoints numerical_not_equals max_hitpoints})}
```
Pentarctagon, line 75: Should be `$max_hitpoints`.

Toranks: $bob.max_hitpoints, yes
An oversight. Since it didn't give an error, I didn't realize

data/test/scenarios/wml_tests/UnitsWML/zero_experience.cfg
```diff
@@ -0,0 +1,80 @@
...
+#####
+# API(s) being tested: [unit] max_experience=0
+##
+# Actions:
+# Change Alice's and Bob's max_experience to 0. Delete Bob's 'Orcish Warrior' advancement but not Alice's. Harm bob with 1 hitpoint to be sure that he have some damage. They then fight each other.
+##
+# Expected end state:
+# Alice levels up. Bob doesn't, and keeps his previous health.
+#####
+{GENERIC_UNIT_TEST "zero_experience" (
```
stevecotton, line 12: There's a `COMMON_KEEP_A_B_UNIT_TEST` macro which means Alice and Bob already start adjacent to each other.

Toranks: I didn't know there were different maps for each test. I'll keep it in mind.

data/test/scenarios/wml_tests/UnitsWML/zero_experience.cfg
```diff
@@ -0,0 +1,76 @@
...
+        [modify_unit]
+            [filter]
+                id=alice,bob
+            [/filter]
+            [effect]
+                apply_to=max_experience
+                set=0
+            [/effect]
+        [/modify_unit]
+        [modify_unit]
```
stevecotton, line 24: At this point, `[store_unit]` and `{ASSERT ({VARIABLE_CONDITIONAL bob.max_experience numerical_equals 0})}`. That assert fails, in the stored WML Bob has `max_experience==1`.

That seems to be a sanity-check in `unit::init`, `std::max(1, v->to_int(max_experience_))`. I think it's good to have a test that documents that that sanity-check works, but the rest of this unit test isn't testing what you expect.

Toranks:
New test to complement:

API(s) being tested: [unit] [advancement]

* Actions:
Harm Alice and Bob with 1 hitpoint to be sure that they have some damage. Delete advancements of both units and give them 999 experience. They then fight each other.

* Expected end state:
Bob and Alice doesn't level up, keeps their HP, and gains experience. The default behaviour should be that a unit will never advances if it has no advancements, but still can gain experience internally (seen with inspect console).

data/test/scenarios/wml_tests/UnitsWML/unit_no_advancements.cfg
```diff
@@ -0,0 +1,69 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: [unit] [advancement]
```
stevecotton, line 4: The current standard is to have less spacing on the "API(s) being tested:" line - please do a `git grep` and see how the others look.

data/test/scenarios/wml_tests/UnitsWML/zero_experience.cfg
```diff
@@ -0,0 +1,77 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: [unit] max_experience=0
+##
+# Actions:
+# Change Alice's and Bob's max_experience to 0. Delete Bob's 'Orcish Warrior' advancement but not Alice's. Harm bob with 1 hitpoint to be sure that he have some damage. They then fight each other.
```
stevecotton, line 7: For the first sentence, I think "Try to set Alice's and Bob's max_experience to 0, check that the engine applies a minimum of 1." would be clearer. It's also good to add asserts in the middle of the test, to check that's true both before and after the combat.

BTW, Alice is male (just because Elvish Archers are by default).

data/test/scenarios/wml_tests/UnitsWML/zero_experience.cfg
```diff
@@ -0,0 +1,77 @@
...
+                [source]
+                    x,y=5,3
+                [/source]
+                [destination]
+                    x,y=4,3
+                [/destination]
+            [/attack]
+        [/do_command]
+    [/event]
+    [event]
```
stevecotton, line 57: This could be in the same event.

Toranks: Same with the other test, I guess

PR #7345, opened 2023-01-31
Remove unused deprecation functions from wml_exception.cpp/hpp

stevecotton:
Remove some unnecessary includes of wml_exception.hpp.

The last callers of these functions were removed in f545fbcbe8 and 4437691bdd, and any potential future callers can use config::get_old_attribute instead.

Removes some translatable strings.


stevecotton:
I can remove the include of config.hpp too, so pushing a new revision.

PR #7346, opened 2023-01-31
Add Sand Scamperer portrait and fix Giant Scorpion's 

spixi:
The portraits of Giant Scorpion and Sand Scuttler where not aligned properly. I fixed that, so that their position now matches. I also added a portrait for the Sand Scamperer (based on the Giant Scorpling and the Sand Scuttler with some color shifting and added sand). Finally I added a macro to increase maintainability of the Sand Scamperer's unit cfg.

This belongs to #7244


spixi:
I also added a recolored undead sprite for the undead sand scuttler.

knyghtmare:
Do remember to include screenshots for faster evaluations

spixi:
c
b
a

spixi:
d
Interestingly, the ~RC IPF is not applied to the debug create unit menu and to the unit info bar on the right. Is that a bug because it is a variation?

image

knyghtmare:
> Is that a bug because it is a variation?

Nyet. It's a limitation.

doofus-01:
Hrm, I can't say I like the recolor-as-a-new-unit concept in general, unless it's a place-holder (like the horses), but sand scamperer is already merged to core, apparently, so that ship has sailed and this particular update does make sense.

I'm not sure about the IPF not showing up in the zombie - it rings a bell, I'll try to look into it if no one else gets to it first.

doofus-01 (approved):
No comment on the scorpion-elder edit, will have to check that out later, but the rest seems fine.

data/core/units/undead/Corpse_Soulless.cfg
```diff
@@ -27,8 +27,12 @@
 
 # Variant animations for the Soulless
 #define UNIT_BODY_SOULLESS_GRAPHICS BASE_NAME
+#arg IPF
+NOP#endarg
+#arg ARG
+#endarg
```
soliton-, line 33: Seems unnecessarily complicated to split this into two arguments with a NOP default instead of just one argument.

spixi: Sure. I am unhappy with these NOP calls as well. WML macros cannot handle parentheses correctly, that is the reason for that mess. I will revise that and create new sprites with Imagemagick. That would also fix the issue with the IPF filter not showing up in the info bar. Please leave this PR open!

CelticMinstrel: > WML macros cannot handle parentheses correctly

Uhh… I'm pretty sure this is not true? If you quote it, there should be no issue.

soliton-: Simplified the macro in 1dd72e7.

PR #7349, opened 2023-02-01
Quenoth Rework

Hejnewar:
The faction design of the current Quenoth is not satisfactory with some faults here and there, and I think the faction deserves more effort and scope of originality. This PR will hopefully make the whole faction more interesting and enjoyable to play. Also sorry if this PR is too big. 

Credits: 
Jarom - for big help with the code. 
Everyone who helped with idea for the Quenoth Sun Singer and Sun Sylph.


CelticMinstrel (changes requested):
Whoa, there are a lot of strange things in here. It seems Formation has been removed?

I could read through the code and more or less work out what everything is doing, but it would be easier if you could describe all the changes in words.

More importantly, this is substituting macros into translatable strings, which won't work for translation. (Concatenating the macro as a non-translatable component doesn't help much with that issue; the string can be translated, but not correctly.)

Hejnewar:
As per request:
Fighter lines no longer have formation and it has been replaced with "uses X attacks" weapon special and more than 1 attack.
Scout lines are mostly unchanged except for balance changes.
Tauroch lines now have 2 attacks that are split between the rider and the tauroch with "once per turn" weapon special.
Mystic -> Shyde line now have Tailwind X ability that allows them to make units around them faster by X.
Sun Singer and Sylph now have a "song"  special, after "singing" (aka. attacking with) first verse, you can sing the second and for Sun Sylph you also get the third verse. Each verse is progressively stronger and gains additional attack special ray for the second verse and nova for the third verse, also singing third verse grants illumination for 3 turns.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1,5 +1,406 @@
...
+# wmlxgettext: [/attack]
+#enddef
+
+#define WEAPON_SPECIAL_SONG_VERSE ACTIVE_ON X DISTINGUISHER
+# wmlxgettext: [attack]
+# wmlxgettext: [specials]
+    [disable]
+        id = verse_{DISTINGUISHER}
+        name = _ "song verse"
+        description = _ "This attack can only be used as " + "{X}" + _ " verse of the song."
```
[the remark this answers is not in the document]
Hejnewar: Honestly, Im bad at describing stuff and I will gladly change the descriptions as long as I know how to change them.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1,5 +1,406 @@
 #textdomain wesnoth-help
 
+#define WEAPON_SPECIAL_USES_ATTACKS X
+# wmlxgettext: [attack]
+# wmlxgettext: [specials]
+    [disable]
+        id = uses_attacks_{X}
+        name = _ "uses " + "{X}" + _ " attacks"
```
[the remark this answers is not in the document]
CelticMinstrel: > Also, this seems like something that shouldn't need an event… I thought there was already a key to do this but I guess I was wrong…

This is now #7351.

CelticMinstrel:
Okay, so if I understand correctly:

* The tauroch riders have two attacks, each of which can be used once in a turn
* The sun line has one attack which can be used up to 3 times but has a different effect on each use

Does that sound about right?

As for the translation issues, I'll probably have suggestions for you a bit later.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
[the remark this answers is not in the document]
CelticMinstrel: I think the approach to making this translatable will have to follow this template:

```
#define STR_USES_1_ATTACKS
_"uses 1 attack"#enddef

#define STR_USES_2_ATTACKS
_"uses 2 attacks"#enddef

# etc up to however many attacks you need

#define SOME_MACRO X
    [disable]
        name = {USES_{X}_ATTACKS}
    [/disable]
#enddef
```

It's kind of unfortunate that we have no way to express it like this though:

`    name = _ ("uses $n attacks", {X})`

That would be the ideal way to handle this. Unfortunately, it does not work. We could maybe theoretically make it work (and #1135 tracks that request), but there are some issues that haven't been worked out, so it hasn't been done.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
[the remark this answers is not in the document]
CelticMinstrel: I think the same general template that I mentioned for "uses N attacks" can be applied here. The specific strings should look something like this:

```
#define STR_VERSE_RESTRICT_1
_ "This attack can only be used as the first verse of the song."#enddef

#define STR_VERSE_RESTRICT_2
_ "This attack can only be used as the second verse of the song."#enddef

#define STR_VERSE_RESTRICT_3
_ "This attack can only be used as the third verse of the song."#enddef
```

Potentially replace first/second/third with 1st/2nd/3rd if you prefer that.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Flanker.cfg
```diff
@@ -11,41 +11,60 @@
...
     advances_to=Quenoth Ranger
-    cost=28
+    cost=30
     usage=fighter
     description= _ "Some fighters find that their true talents lie not within rigid spear ranks, but out in the sands, striking at the flanks and rear of enemy forces. Fast and nimble, these elves make use of the weaknesses in enemy formations to disrupt their lines and sow chaos on the battlefield. In more cutthroat conflicts, Flankers even rely on the use of poison — a tool whose use was frowned on by their ancestors — to cripple their foes before charging in for the finish."
     die_sound={SOUND_LIST:ELF_HIT}
     [abilities]
         {ABILITY_SKIRMISHER}
     [/abilities]
+    attacks = 4
```
CelticMinstrel, line 28: Stylistically, it seems we prefer not to have spaces around the `=`.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1,5 +1,406 @@
...
+[+specials]
+# wmlxgettext: [/specials]
+# wmlxgettext: [/attack]
+#enddef
+
+#define ABILITY_TAILWIND X Y
+    [dummy]
+        id = tailwind_{X}
+        name = _ "tailwind"+{X}
+        description = _ "At the start of the turn this unit increases movement points of surrounding units by " + "{X}"
```
CelticMinstrel, line 376: This also seems like it would have translation problems of the same sort as the other two strings I've commented on. The solution is probably the same.

CelticMinstrel:
I'd like to request that all formulas be quoted even if they do not contain a plus sign.

Aside from that, there appear to be some schema validation issues, but due to #7310 I can't tell what they are. I didn't see anything obvious from scanning the code.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Fighter.cfg
```diff
@@ -11,20 +11,18 @@
...
     movement_type=quenoth_foot
     movement=5
     experience=40
     level=1
     alignment=lawful
     advances_to=Quenoth Warrior,Quenoth Flanker
-    cost=15
+    cost=12
     usage=fighter
     description= _ "Long ago, during more prosperous times, elven warriors favored the use of swords as more elegant, versatile weapons compared to other melee armaments. However, in recent times, the dearth of supplies for smithing has reduced the availability of blade-crafting, necessitating the fashioning of cheaper, more easily repaired weaponry. Quenoth Fighters are remarkably skilled in coordinated groups, where they can counteract the lack of more sophisticated weaponry by forming robust spear ranks capable of repelling nearly any frontal assault."
```
stevecotton, line 23: This text describes the Formation ability, and needs to be updated too.

With a redesign that changes the signature ability, I think it would be better to make this a new unit line; as working title maybe "Quenoth Duelist". While the Quenoth Fighter wouldn't be in the campaign any more, it could resurface in Ageless or other campaigns.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Sun_Singer.cfg
```diff
@@ -32,8 +33,22 @@
         icon=attacks/touch-faerie.png
     [/attack]
     [attack]
-        name=faerie fire
-        description= _"faerie fire"
+        name=sun invocation
+        description= _"sun invocation"
+        #textdomain wesnoth-utbs
```
stevecotton, line 38: Should probably move to include "sun invocation" in UtBS.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Flanker.cfg
Hejnewar: Yeah, in case of unit files sure my mistake, I use it mostly because it is way easier to read the code for me with this formating but that mostly applies to abilities and not units that already have different formating.

Hejnewar:
> The sun line has one attack which can be used up to 3 times but has a different effect on each use

Yes or at least I think that this describes it pretty well, you can try all the changes form this PR in either Under the Burning Suns But Better on 1.17 or Ultimate Recruits in 1.16 just be careful with debuging Sun Singers / Syplhs it is best to wait with them for 2 turns after debuging so they reset to the working state. 

Actually I should check leveling case again just in case...

Hejnewar:
> The sun line has one attack which can be used up to 3 times but has a different effect on each use

Yes or at least I think that this describes it pretty well, you can try all the changes form this PR in either Under the Burning Suns But Better on 1.17 or Ultimate Recruits in 1.16 just be careful with debuging Sun Singers / Syplhs it is best to wait with them for 2 turns after debuging so they reset to the working state. 

Actually I should check leveling case again just in case...

data/campaigns/Under_the_Burning_Suns/units/quenoth/Fighter.cfg
stevecotton: Even if the ability isn't around, it would be helpful to rename the units. That would make it easier to refer to in walkthroughs that will still need to cater to 1.16 players even after 1.18 is out.

The ability itself is likely to live on in the unit tests - while it isn't found there yet, there's parts of the C++ code that I want to be tested by something in mainline. If I can no longer find coverage for that C++ in mainline then it'll be added to the tests.

Hejnewar: Honestly I like the current simple name and I don't think that the change is necessary. Also this with this reasoning I would have to rename most of the units because well, most of them changed a lot. 

Well it being in unit tests is outside of reach of your usual gamer. :P

CelticMinstrel:
> I will try, but I strangely remember a bug with quouted formula, given that this was worked on during like 7 month period bit by bit I might be missremembering but I would rather have unquoted formula than a bug.

I think if quoting the formula breaks something, you should probably at least add a comment to note that quoting the formula breaks it.

Hejnewar:
I know there are still some things to be decided. (Quenoth Fighter name for example, and = in abilites but that one I will most likely remove in the end.) But I belive that I managed to fix most of the stuff requested. With attacks_used I would rather have this be shown in UI than as an ability, if this is not going to happen I will just add it as dummy descriptions.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -372,8 +351,8 @@
 #define ABILITY_TAILWIND X Y
     [dummy]
         id = tailwind_{X}
-        name = _ "tailwind"+{X}
-        description = _ "At the start of the turn this unit increases movement points of surrounding units by " + "{X}"
```
CelticMinstrel: We'll need to do something about the description, although tacking it on the end is slightly less bad than embedding it in the middle. Worst-case, we need a `po` comment letting translators know that "number of movement points" will be appended. It would be better though to do something like what was done with the song verse.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -372,8 +351,8 @@
...
+        name = _ "tailwind +"+{X}
+        description = _ "At the start of the turn this unit increases movement points of surrounding units by +"+{X}
```
CelticMinstrel, line 355: We'll need to do something about the description, although tacking it on the end is slightly less bad than embedding it in the middle. Worst-case, we need a po comment letting translators know that "number of movement points" will be appended. It would be better though to do something like what was done with the song verse.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1,5 +1,385 @@
...
+        [modify_unit]
+            [filter]
+                side = $side_number
+                [filter_adjacent]
+                    ability = tailwind_{X}
+                    is_enemy = no
+                [/filter_adjacent]
+                [not]
+                    [filter_adjacent]
+                        ability = tailwind_{Y}
```
stevecotton, line 372: This code is overly complex for something that doesn't generalise.

Because of passing `Y` as a macro argument, all the `[unit_type]`s with `tailwind_1`have to declare that it doesn't stack with `tailwind_2`. The `[unit_type]` with `tailwind_2` has to declare some value for `Y`, and it declares that it doesn't stack with `tailwind_3`.

However, if any unit did have `tailwind_3` then this code would let `tailwind_1` stack with it. So the code might as well be simpler and hardcode that there's only 2 levels of tailwind.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1,5 +1,385 @@
...
+# wmlxgettext: [attack]
+# wmlxgettext: [specials]
+    [dummy]
+        id = ray
+        name = _"ray"
+        description= _ "This weapon, when used offensively, also deals damage to a unit behind the target."
+        active_on = offense
+    [/dummy]
+[/specials]
+[/attack]
```
stevecotton, line 318: We shouldn't be adding unbalanced macros for abilities in mainline - compare the `feeding` ability for example.

CelticMinstrel: A better example for comparison is probably the Merman brawler's stun, since it's also an attack-triggered ability.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
CelticMinstrel: Come to think of it, if the goal is just to have the three tailwinds not stack, is there any reason they need a number encoded into the ID at all?

Hejnewar: It was done this way because without ID as soon as there is unit with tailwind +2 all tailwind will be +2, so mystics and shamans despite having it set to +1 will also give +2 and it doesn't matter if unit with tailwind +2 dies latter because there event is already there. This is also why we had to prevent the stacking.

CelticMinstrel: > as soon as there is unit with tailwind +2 all tailwind will be +2, so mystics and shamans despite having it set to +1 will also give +2 and it doesn't matter if unit with tailwind +2 dies latter because there event is already there.

??????

Why would that be the case? Can't the tailwind event somehow read the amount to increase from the unit providing the ability?

The logic would be something like:

* Find all units adjacent to a unit with tailwind.
* For each of those units, collect all the adjacent units with tailwind.
* Select the adjacent unit with the highest tailwind value.
* Apply that tailwind value; then move on to the next unit.

Hejnewar: Dont ask me, I dont know. Im just saying what is happening in game and explaining why this was done the way it is. I had a memory of this happening but I tested in like hour ago and indeed it works this way.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
Hejnewar: Just for me to understand this better. What is happening here is the tool reading ``[/specials]
[/attack]`` but not reading ``wmlxgettext: [attack] wmlxgettext: [specials]``, shich causes error for the tool and the way to prevent that is spliting the event and the ability / weapon special into two macros?

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
CelticMinstrel: Right, that happens if the event hard-codes the strength but the IDs of the event and ability do not. I'm saying it shouldn't need to hard-code the strength into the event logic _or_ the IDs. See the "explode" ability in #7249 for an example of how this can be done.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
stevecotton: The main reason isn't tooling, it's that we find the `[/specials][/attack]...events...[+attack][+specials]` style to be unpleasant to read. But also, it needs those comments for wmlxgettext.

In terms of how to do it, it might be time for @CelticMinstrel or I to write our preferred implementation, rather than to say that we can see a better way to do it.

@CelticMinstrel data/lua/stun.lua should have more comments, if it's going to be a reference document for how to implement abilities.

CelticMinstrel: > if it's going to be a reference document for how to implement abilities.

…why would it be a reference document? There's already a reference on the wiki for the game_events API, and there are several examples of abilities to look at including stun.lua.

stevecotton: I mean, if it's the example that we refer people to then it should have more comments.

stevecotton: @CelticMinstrel would it make sense for the schema to accept `[dummy]value=`?

@Hejnewar the commit that's auto-linked itself to this PR implements this:
```
#define UTBS_ABILITY_TAILWIND X
    [dummy]
        id = tailwind
        value = {X}
        name = _ "tailwind +"+{X}
        description = _ "At the start of the turn this unit increases movement points of surrounding units by +"+{X}
    [/dummy]
#enddef
```

CelticMinstrel: > would it make sense for the schema to accept `[dummy]value=`?

I don't think so; if you need a value for a "dummy" ability I think it makes more sense to take the approach in #7249 and use a custom tag which is then added to the schema as an ability with a value.

stevecotton:
@CelticMinstrel and @nemaara what should be our current best practice for enabling the `[event]` tags for the abilities?

I'm considering a line in each scenario, as SoF does with `SOF_RUNIC_EVENTS`, WoF does with `SECOND_INTENDANT_REPLACEMENT_EVENTS` and UtBS S02 currently does with `DEHYDRATION_EVENTS`.

It might use `[load_resource]` internally to the macro, but that's a detail that wouldn't need changes to the scenario files to do.
```
#define UTBS_ABILITY_EVENTS
    [load_resource]
        id=utbs_ability_events
    [/load_resource]
#enddef
```

CelticMinstrel:
> I'm considering a line in each scenario, as SoF does with `SOF_RUNIC_EVENTS`, WoF does with `SECOND_INTENDANT_REPLACEMENT_EVENTS` and UtBS S02 currently does with `DEHYDRATION_EVENTS`.

I'd put it in the `[campaign]` tag rather than in every scenario, unless you have a good reason to not want it in one scenario.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
stevecotton: @Hejnewar I've prototyped the implement-tailwind-as-a-single-event and event-via-resource in the couple of commits that have auto-linked to this PR, although the renaming from `[dummy]` to `[tailwind]` is still a todo. However I'm leaving it to you to work out what to do with the other abilities.

Hejnewar:
At this point Im just waiting to be roasted alive. But it works!

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1,5 +1,433 @@
...
+    [/event]
+#enddef
+
+#define STR_TAILWIND_RESTRICT_1
+_ "At the start of the turn this unit increases movement points of surrounding units by +1"#enddef
+
+#define STR_TAILWIND_RESTRICT_2
+_ "At the start of the turn this unit increases movement points of surrounding units by +2"#enddef
+
+#define UTBS_TAILWIND_DESCRIPTION X
```
stevecotton, line 424: UTBS_ABILITY_TAILWIND is now so small that the description could go in it, I don't see why this needs a separate macro.

However, the ability probably still needs to be changed from `[dummy]` to `[tailwind]`.

Hejnewar: How can I do that?

CelticMinstrel: It's literally what he said. Change `[dummy]` to `[tailwind]` and that's it.

Although I guess it also needs to be added to the schema as an ability that takes a value.

Hejnewar: Done, also the macro was created to make translation easier, requested by CelticMinstrel in one of the earlier reviews, I dont know if it is still necessary but I still kept it in the same way just to be safe.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1,5 +1,433 @@
...
+                [filter_side]
+                    [allied_with]
+                        side=$side_number
+                    [/allied_with]
+                [/filter_side]
+                [and]
+                    # filter_wml is slow, which is why there's already an ability=tailwind filter before this
+                    [filter_wml]
+                        [abilities]
+                            [dummy]
```
stevecotton, line 357: The `dummy` tags in the `filter_wml` need to change too (and there's another on line 379).

Hejnewar: Sorry, thanks and done.

CelticMinstrel (changes requested):
Okay, so there's still some remaining issues.

* Needs wmlindent run
* `[harm_unit]alignment=` isn't set to permit variables (this is an error in the schema in master though, not an issue with the PR)
* It appears the text "tailwind +x" is not marked translatable.
* Apparently there are some `active_on=` keys without a value, which is not allowed. I think the fix for this is to make `{WEAPON_SPECIAL_SONG_VERSE}` take a key-value instead of just the value, so it looks like `{WEAPON_SPECIAL_SONG_VERSE active_on=offense ...}`.
* You seem to have an `apply_to` in your `[missile_frame]`. This is not supported. (`apply_to` is a key of an animation, not a frame.)

Hejnewar:
Thanks for consideration and telling me this before I even asked. I hope this resolves all the problems.

CelticMinstrel (approved):
Looks good now. As far as I'm concerned this can be merged as soon as the remaining schema issue is fixed on master, which should be pretty soon. I'm not sure if anyone else has other changes to request though.

Hejnewar:
Great, I do plan to update them in the future if necessary after they have been played a bit more. Also it would be great if this PR made it to 1.17.13 but I wont be around to merge it (work), so if everything is fine before then I would like to request someone else to do so. Still I would like to have approval from Yumi.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Druid.cfg
```diff
@@ -9,15 +9,17 @@
     profile="portraits/quenoth/druid.webp"
     hitpoints=45
     movement_type=quenoth_foot
-    movement=5
-    experience=180
+    movement=6
+    experience=150
```
nemaara, line 13: Same here, why lower exp?

data/campaigns/Under_the_Burning_Suns/units/quenoth/Fighter.cfg
nemaara: I also prefer leaving the name fighter, but if it's really that big of a problem for code then I guess we can change it.

nemaara, line 23:
```suggestion
    description= _ "Long ago, during more prosperous times, elven warriors favored the use of swords as more elegant, versatile weapons compared to other melee armaments. However, in recent times, the dearth of supplies for smithing has reduced the availability of blade-crafting, necessitating the fashioning of cheaper, more easily repaired weaponry. To compensate for this diminishing in armament quality, the Quenoth have adopted a greater flexibility in their use. In the open sands, a fighter is trained to develop the acumen to split his attention between multiple enemies, be they brigand, wild creature, or undead."
```

data/campaigns/Under_the_Burning_Suns/units/quenoth/Flanker.cfg
```diff
@@ -11,43 +11,58 @@
     {TRAIT_RESILIENT}
     image=units/quenoth/flanker.png
     profile="portraits/quenoth/flanker.webp"
-    hitpoints=43
+    hitpoints=45
     movement_type=quenoth_foot
-    movement=5
-    experience=90
+    movement=6
+    experience=73
```
nemaara, line 17: exp y

data/campaigns/Under_the_Burning_Suns/units/quenoth/Mystic.cfg
```diff
@@ -28,17 +30,17 @@
...
-        damage=4
+        damage=3
         number=2
         range=melee
         icon=attacks/druidstaff.png
     [/attack]
     [attack]
         name=sand
-        description= _"sand"
+        description= _"sandstorm"
```
nemaara, line 40: I don't mind changing this to sandstorm for higher levels but it seems like it'd be neat to keep it as just plain sand for the lvl 1 unit.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Shaman.cfg
```diff
@@ -7,17 +7,19 @@
     gender=female
     image=units/quenoth/shaman.png
     profile="portraits/quenoth/shaman.webp"
-    hitpoints=35
+    hitpoints=37
     movement_type=quenoth_foot
-    movement=5
-    experience=65
+    movement=6
+    experience=60
```
nemaara, line 13: exp

data/campaigns/Under_the_Burning_Suns/units/quenoth/Mystic.cfg
```diff
@@ -7,17 +7,19 @@
...
     profile="portraits/quenoth/mystic.webp"
-    hitpoints=25
+    hitpoints=22
     movement_type=quenoth_foot
     movement=5
-    experience=28
+    experience=31
     level=1
-    alignment=neutral
+    alignment=lawful
```
nemaara, line 15: I forgot but why are we changing the alignment?

data/campaigns/Under_the_Burning_Suns/units/quenoth/Pathfinder.cfg
```diff
@@ -6,17 +6,17 @@
...
     race=quenoth
-    hitpoints=40
+    hitpoints=47
     movement_type=quenoth_horse
     [resistance]
         pierce=120
     [/resistance]
     movement=8
-    experience=80
+    experience=76
```
nemaara, line 15: xp

data/campaigns/Under_the_Burning_Suns/units/quenoth/Sun_Singer.cfg
```diff
@@ -9,16 +9,17 @@
     profile="portraits/quenoth/sun_singer.webp"
     hitpoints=36
     movement_type=quenoth_foot
-    movement=5
-    experience=105
+    movement=6
+    experience=80
```
nemaara, line 13: exp

data/campaigns/Under_the_Burning_Suns/units/quenoth/Tauroch_Protector.cfg
```diff
@@ -15,11 +15,11 @@
     {TRAIT_RESILIENT}
     {TRAIT_DEXTROUS}
     {TRAIT_HEALTHY}
-    hitpoints=71
+    hitpoints=59
```
nemaara, line 18: That's a huuuge nerf to HP, are you sure this is right?

data/campaigns/Under_the_Burning_Suns/units/quenoth/Tauroch_Stalwart.cfg
```diff
@@ -15,38 +15,42 @@
     {TRAIT_RESILIENT}
     {TRAIT_DEXTROUS}
     {TRAIT_HEALTHY}
-    hitpoints=56
+    hitpoints=48
```
nemaara, line 18: Again big gigantic nerf, are you sure about this?

data/campaigns/Under_the_Burning_Suns/units/quenoth/Warrior.cfg
```diff
@@ -11,28 +11,27 @@
     {TRAIT_RESILIENT}
     image=units/quenoth/warrior.png
     profile="portraits/quenoth/warrior.webp"
-    hitpoints=44
+    hitpoints=52
     movement_type=quenoth_foot
     movement=5
-    experience=80
+    experience=77
```
nemaara, line 17: exp

data/campaigns/Under_the_Burning_Suns/units/quenoth/Tauroch_Stalwart.cfg
Hejnewar: You are the one that asked me to do this because they felt too strong when you played them.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Tauroch_Protector.cfg
Hejnewar: You are the one that asked me to do this because they felt too strong when you played them. x2

data/campaigns/Under_the_Burning_Suns/units/quenoth/Mystic.cfg
Hejnewar: Because every single unit in this line / race other than them is lawful.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Mystic.cfg
Hejnewar: Maybe sandwind as moderated name? :P I honestly just wanted to be consistent and I liked the name way more.

nemaara: Nah then keep it.

data/campaigns/Under_the_Burning_Suns/units/quenoth/Tauroch_Protector.cfg
nemaara: Yeah but I seem to remember suggesting nerfing dmg more than HP. Unless I remembered wrong. Maybe we can make this at least 60 something? And if further dmg nerf would be needed to compensate then that's fine.

nemaara (approved):
Talked on discord, at a later date we can add an exp modifier patch to UtBS as well as further rebalance the taurochs. That doesn't need to be in this PR, a separate patch will be needed.

PR #7350, opened 2023-02-01
modify calling of unit_abilities::effect and get_composite_value()

newfrenchy83:
in addition to simplify the calling of calculation function for int weapon specials, it also because an commit what i want push in https://github.com/wesnoth/wesnoth/pull/6968 a attribute  require this function.

### DIFF
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1123,9 +1123,7 @@ void attack_type::modified_attacks(unsigned & min_attacks,
                                    unsigned & max_attacks) const
 {
 	// Apply [attacks].
-	unit_abilities::effect attacks_effect(get_specials_and_abilities("attacks"),
-	                                      num_attacks(), shared_from_this());
-	int attacks_value = attacks_effect.get_composite_value();
+	int attacks_value = composite_value(get_specials_and_abilities("attacks"), num_attacks());
 
 	if ( attacks_value < 0 ) {
 		attacks_value = num_attacks();
@@ -1148,8 +1146,7 @@ void attack_type::modified_attacks(unsigned & min_attacks,
  */
 int attack_type::modified_damage() const
 {
-	unit_abilities::effect dmg_effect(get_specials_and_abilities("damage"), damage(), shared_from_this());
-	int damage_value = dmg_effect.get_composite_value();
+	int damage_value = composite_value(get_specials_and_abilities("damage"), damage());
 	return damage_value;
 }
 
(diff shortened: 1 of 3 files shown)


src/units/abilities.cpp
```diff
@@ -1302,6 +1299,13 @@ unit_ability_list attack_type::get_specials_and_abilities(const std::string& spe
 	return abil_list;
 }
 
+int attack_type::composite_value(const unit_ability_list& abil_list, int base_value) const
+{
+	unit_abilities::effect align_effect(abil_list, base_value, shared_from_this());
+	int composite_value = align_effect.get_composite_value();
```
stevecotton, line 1305: The name `align_effect` is a little confusing, because this isn't always about alignment.

src/units/attack_type.hpp
```diff
@@ -90,6 +90,12 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 	                      unsigned & max_attacks) const;
 	/** Returns the damage per attack of this weapon, considering specials. */
 	int modified_damage() const;
+	/** 
+	 * Return the special weapon value, considering specials.
+	 * @param abil_list the list of special checked.
+	 * @param base_value the value modified ornot by function.
```
stevecotton, line 96: Needs a space in "or not".

stevecotton:
Nice, I like it.

I'm wondering if `int composite_value(std::string, bool include_abilities, int) const;` would allow the `get_specials` or `get_specials_and_abilities` call to move too; however I'm happy without that change. Do whatever fits best with the other PRs that you have planned.

src/units/abilities.cpp
CelticMinstrel: Is there any reason this whole function can't just be reduced to one line?

```cpp
return unit_abilities::effect(abil_list, base_value, shared_from_this()).get_composite_value();
```

newfrenchy83: No, thank you, itis corrected and rebased

newfrenchy83:
> Nice, I like it.
> 
> I'm wondering if `int composite_value(std::string, bool include_abilities, int) const;` would allow the `get_specials` or `get_specials_and_abilities` call to move too; however I'm happy without that change. Do whatever fits best with the other PRs that you have planned.

i must use directly list for change in other PR, because i need of list modified and not recreate list complete.

src/units/attack_type.hpp
newfrenchy83: done

PR #7354, opened 2023-02-04
Have hidden achievements actually be hidden.

Pentarctagon:
Also hidden_name and hidden_hint are no longer attributes of [achievement].


CelticMinstrel:
I guess with this we can actually have two layers of "hidden" achievements.

Layer 1: set non-completed icon to something ambiguous, non-completed name to ???, and non-completed description to a hint, but hidden to no.

Layer 2: Omit completed icon/name/description. Set hidden to yes.

CelticMinstrel (approved):

PR #7355, opened 2023-02-05
Save last selected achievement to the preferences

Pentarctagon:
Also give a better warning when amount isn't a number.

### DIFF
--- a/src/preferences/general.hpp
+++ b/src/preferences/general.hpp
@@ -274,6 +274,9 @@ namespace preferences {
 	sort_order::type addon_manager_saved_order_direction();
 	void set_addon_manager_saved_order_direction(sort_order::type value);
 
+	std::string selected_achievement_group();
+	void set_selected_achievement_group(const std::string& content_for);
+
 	/**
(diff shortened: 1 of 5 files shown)


CelticMinstrel (approved):

src/preferences/general.hpp
```diff
@@ -274,6 +274,9 @@ namespace preferences {
 	sort_order::type addon_manager_saved_order_direction();
 	void set_addon_manager_saved_order_direction(sort_order::type value);
 
+	std::string selected_achievement_group();
+	void set_selected_achievement_group(const std::string& id);
```
stevecotton, line 278: The naming of `id` feels a bit weird, because IIUC other functions call this `content_for` and call something else `id`.

stevecotton (approved):
Approving to clarify that I'm happy with it merging either with or without a rename of that `id`.

src/preferences/general.hpp
Pentarctagon: Yeah, true, updated.

PR #7359, opened 2023-02-05
Lua warning cleanup

CelticMinstrel:


CelticMinstrel:
I'd like to ask @knyghtmare to look over the changes in WC, particularly that one change to 3B_Wreck.lua that dedents several lines and moves a parenthesis.

Pentarctagon:
After fetching this into my local workspace, it appears this has reduced the number of things being warned about from ~1300 to ~300, so that's definitely an improvement.

The remaining seem to generally be a mix of:
* Method expects a maximum of X arguments but instead is receiving Y.
* Cannot assign X|Y|Z to parameter X.

The second one seems like it's just an issue with how it's parsing the allowed argument types, but the first seems like it may be warning about an actual issue?

CelticMinstrel:
> Cannot assign X|Y|Z to parameter X.

I think this is essentially that we have a variable that's known to be one of X, Y, or Z, and we're passing it to a function that only takes X without checking the type first. Since it's literally everywhere, I disabled the warning in the VSCode settings (for now at least).

> Method expects a maximum of X arguments but instead is receiving Y.

From what I can see there are basically 4 remaining cases of this (some of the cases have multiple instances).

* ca_recruit_rushers.lua does something strange and declares a stub of a function that might be loaded from another file, or something? The stub takes no arguments. I didn't look closely so I'm not quite sure where the real function is declared.
* cave_map_generator.lua is map generator code, so it requires two extra arguments that aren't declared in the LuaDoc. I just wasn't sure how to handle this case where the argument list depends on context.
* wml_actions.set_variable and wml_actions.clear_variable take an optional argument to enable code reuse in modify_side and modify_unit
* animate_unit passes an extra, unimplemented argument to scroll_to_hex. It's something we probably need to add at some point so I didn't want to remove it.

Pentarctagon (approved):

knyghtmare (changes requested):
You broke World Conquest...again...
image

Do not merge until fix

stevecotton:
This needs a rebase that reorganises stuff between commits. This one's too large to squash into a single commit, yet if it stays as the current set of commits then the feeding changes will be split across 3 commits with the latest one being a bugfix to a previous one.

My preference would be to split the feeding changes into a separate PR, so that can merge and this one can shrink.

stevecotton (changes requested):
Feeding is still broken, as detected by the failing unit test.

CelticMinstrel:
???

I _definitely_ remember testing it and it worked, so…

Well, I guess I'll just test it again.

knyghtmare:
> You broke World Conquest...again...

It loaded now.

> Feeding is still broken, as detected by the failing unit test.

I confirm this. Found while testing World Conquest

knyghtmare:
Also, found this when I loaded a campaign on the PR branch:
```ini
20230227 14:46:58 error scripting/lua: ai/micro_ais/micro_ai_helper.lua:368: attempt to concatenate a nil value (field 'tag')
stack traceback:
	ai/micro_ais/micro_ai_helper.lua:368: in field 'micro_ai_setup'
	lua/wml/micro_ai.lua:54: in local 'cmd'
	lua/wml-utils.lua:145: in field 'handle_event_commands'
	[C]: in ?
```

gfgtdf:
> I'd like to ask @knyghtmare to look over the changes in WC, particularly that one change to 3B_Wreck.lua that dedents several lines and moves a parenthesis.

This looks like it fixes a (small) bug in the WC code, i'd probably merge this one independently to put it into 1.16

CelticMinstrel:
Now that this has been rebased and reworked, I'm considering reducing the size by cherry-picking parts of it.

For example, any of the commits that _only_ change LuaDoc comments should be safe to merge to master without worry. I did some work to ensure no commits contained both doc and logic changes.

Extracting the WC and feeding changes into a separate PR may be worthwhile, but I'm not sure. If someone else wishes to cherry-pick either set of commits and work on that separately, that'd be fine with me.

There's also large swathes of changes to AI code, slowly weeding out places that still refer to locations as `[1]` and `[2]` instead of `x` and `y`. Though it _feels_ fairly safe to merge that, the fact remains that it's changing large chunks of code and the chance to introduce issues is definitely not zero.

Also in here is a fix for #8297 which should probably be cherry-picked separately.

CelticMinstrel:
Oh, and for the record: on this branch, my VSCode currently has ~300 Lua warnings, which if nothing else is definitely a big improvement for 1000+.

Pentarctagon:
Looks like it broke the WML tests somehow?

CelticMinstrel:
Lovely! It's not that big of a deal right now… but I'll look into it soon. I think I'd like to work to merge the ones tagged `[LuaDoc]` first though. I could gather them into a separate pull request but… is there really any point in doing that? They literally touch comments only, so that couldn't possibly break anything, right?

Pentarctagon:
I'd be really confused if a comment managed to break something.

stevecotton:
While I wouldn't expect docs to break anything, it seems trivial to create a branch, push it, and run the CI on it. It doesn't even need to be checked out locally, it can just be created on a commit near the start of your working branch.

CelticMinstrel:
Okay, now I have CI running on _only_ the `[LuaDoc]` commits, plus I've fixed the mistake with respect to #8297.

CelticMinstrel:
Now almost all the `[LuaDoc]` is gone, other than the ai_helper stuff which is still WIP. I split out the WC stuff into #8366, are there any other bits that look like they can easily be split out?

Pentarctagon:
Is this something you're looking to get into 1.18?

CelticMinstrel:
It would be nice, but I could definitely understand if it won't work out. I do want to have at least the restructuring of ai_helper's documentation ready for 1.18 though – that's just comments so there's no chance it can break anything. Going through the documentation has been helpful to point to some of the issues that make it not exactly a generally-usable API, so for 1.19 maybe I can look into tidying it up.

CelticMinstrel:
It would probably be a good idea to cherry-pick at least the most obvious and straightforward fixes for 1.18, like the first and fourth commits. The ones that adjust "obsolete" things on the other hand aren't really that important to be done by then.

data/campaigns/tutorial/lua/character_selection.lua
```diff
@@ -20,6 +20,7 @@ function wml_actions.select_character()
 	if character == 2 then
 		wesnoth.units.to_map({
 			type = "Fighteress",
+			side = 1,
```
stevecotton, line 23: The commit says that this fixes a warning, but I couldn't reproduce a warning in the engine. Is it a warning shown only in an IDE using the EmmyLua extension?

I feel this one's worth cherry-picking, because it's good for the tutorial to be example code that UMC authors can use as a reference, so it's good for it to show how to put the character on other sides too. I'll happily do that cherry-pick, if you tell me enough to clarify the commit message's comment about warnings.

CelticMinstrel: Okay, I'll explain what happened.

The `to_map` function was documented as taking only a `unit` as a parameter, which meant that any time someone passed a table, the Lua extension decided that _every field of a unit was a mandatory key_. That seemed wrong, so I made most of the keys optional in commit 3bf903e, but left the core fields of `type` and `side` mandatory.

However, I later realized that that produces _different_ warnings, because it now considers all those fields to be possibly `nil` when _reading_ from a unit. So I'm planning to revert that change and instead annotate that `to_map` can take either a `unit` or a `WMLTable`. This means it just won't check the contents of the table, which is maybe not the best outcome, but… we'd basically have to define a separate `unitWML` type to make it actually check correctly, and even then it wouldn't be perfect, as it can't validate WML children. I _might_ do that as well, not sure.

stevecotton: Cherry-picked as #8405

CelticMinstrel:
I've split parts of this out into separate PRs:

* #8461 
* #8462 
* #8463 
* #8464 
* #8465 
* #8466 

Once those are all merged to master, I'll also split out the `ai_helper` documentation work and then rebase this, which should make it much more manageable for review.

CelticMinstrel:
I'll probably merge this tomorrow, as long as the CI doesn't show any failures.

PR #7360, opened 2023-02-06
NR S07 - Arthian > Ro'Arthian typo

Toranks:
In only one place Ro'Arthian has an error in his name:

#. [part]
#: data/campaigns/Northern_Rebirth/utils/utils.cfg:160
msgid "Arthian"
msgstr "Arthian"

I realized it by checking the translations.

### DIFF
--- a/data/campaigns/Northern_Rebirth/utils/utils.cfg
+++ b/data/campaigns/Northern_Rebirth/utils/utils.cfg
@@ -157,7 +157,7 @@
 
 #define STORY_PART_ARTHIAN SPEECH_STRING
     [part]
-        story={CAPTION _"Ro'Arthian"} + {SPEECH_STRING}
+        story={CAPTION _"Ro’Arthian"} + {SPEECH_STRING}
         background=portraits/Arthian.webp
         scale_background=no
     [/part]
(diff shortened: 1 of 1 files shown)


Wedge009:
As a translated string, shouldn't it be `Ro’Arthian` ('curly' apostrophe) rather than `Ro'Arthian`?

For example:
https://github.com/wesnoth/wesnoth/blob/d3b6cf7b540e666591b92b304a6fdca1aa96a557/data/campaigns/Northern_Rebirth/scenarios/07a_Settling_Disputes.cfg#L222

Toranks:
> As a translated string, shouldn't it be `Ro’Arthian` ('curly' apostrophe) rather than `Ro'Arthian`?
> 
> For example:
> 
> https://github.com/wesnoth/wesnoth/blob/d3b6cf7b540e666591b92b304a6fdca1aa96a557/data/campaigns/Northern_Rebirth/scenarios/07a_Settling_Disputes.cfg#L222

LOL True. I don't even know where that apostrophe is on my spanish keyboard. In the spanish translation (not mine, but from the previous translator) use Ro'Arthian, the same apostrophe as in the id

Wedge009:
The 'curly' apostrophe isn't on a standard US keyboard either. `id` is different from `name`, the former doesn't get translated.

Issue #6422, opened 2022-01-05
Editor: add a way to hide or move the "Shift + left-click:"... text at the bottom of the screen


stevecotton:
1.16.x and 1.17, on all platforms.

Commit 8b236620dd39123501403322f52870373536edcb changed the text at the bottom of the map-editor screen from
```
    	FG: Savannah/Forest (Forest,) | BG: Green Grass/Grassland (Flat)
```
to the much longer
```
    	Left-click: Savannah/Forest (Forest, ) | Right-click: Green Grass/Grassland (Flat)
    	Shift + left-click: paint overlay layer only | Ctrl + click: copy terrain
```
which can get in the way of editing. This issue is a reminder to myself that I've promised @irydacea that I'll fix it. I'm currently thinking of adding a new button next to the "draw terrain codes", "draw hex coordinates" and "draw grid" toggle buttons.

FrostBlade5:
The text indeed can get in the way of editing, an extra button for terrain description sounds good, however if this is to much work i would aprecciate if the annoying text could simply be removed, anyways take your time it is not a game breaking thing, thank you in advance.

CelticMinstrel:
Removing the text entirely is also game breaking, as it hides important information (the identity of the currently-selected terrain).

cooljeanius:
yeah it's useful information that is only slightly annoying; adding a way to move it or hide it would be highly preferable to removing it entirely

CelticMinstrel:
Of course, there's also the possibility to move it to someplace else.

stevecotton:
Took a while looking through GUI layouts and themes trying to find out where the location of that widget is configured. Turns out that it's a special case that's hardcoded in the C++, and has over time been replaced with tooltips everywhere except in the editor.

Look at display.cpp/display.hpp in bb90f34fdd527aa67bfa93d231f2341c27ee3315 for what it is.
In 1.16 it's handled by the CVideo singleton.
In 1.17 it's moved to a global static in floating_label.cpp / floating_label.hpp.

It looks like the global variable might have been redundant in 1.16, as the value is also stored in editor_palettes.hpp's `help_handle_`. However, that has been removed from editor_palettes.hpp in 1.17.

stevecotton:
For everyone subscribed to this, please comment on the pictures in #7362. Not a complete fix, but is it enough for the stable branch?

cooljeanius:
[...]

PR #7362, opened 2023-02-06
[1.16] Editor: Move the text at the bottom of the screen

stevecotton:
Having it at the very bottom means that it will always obscure the southmost hexes
of the map, and the only way to avoid that is to scroll east/west, or to add an extra
row on the south. Moving it upwards makes it more annoying when working elsewhere on
the map, but it's a trade-off for being able to edit those southmost hexes.

Also, make the text smaller if it's very wide.

This is the 1.16 fix for #6422, with the idea that the second commit will be cherry-picked to 1.17 and then possibly expanded upon. I've also looked at moving the label between the top and bottom of the screen, but that would feature-creep into adding additional functionality to the floating_label support.

help_string_800x600
help_string_1280x1024
help_string_2560x1440

The first commit of the two in this PR
--

Use an editor-only copy of font::set_help_string

This function shows a tooltip-like text at the bottom of the editor's screen,
for example when the paint tool is active it shows which terrains are assigned
to the left and right mouse buttons. There is similar functionality in GUI2,
for example when hovering over the title screen's buttons, which is handled via
event::MESSAGE_SHOW_TOOLTIP.

In the 1.17 branch's 3900419f9230189c699cf2cddb06b31f5025a5f9, I checked that
nothing outside the editor was calling the function. However, that analysis was
done after 1.17's refactoring of the graphics system. For 1.16 I'm not going to
repeat that analysis; the copy that's left in video.cpp is probably dead code,
but I see little reason to remove it from the stable branch.

### DIFF
--- a/changelog_entries/editor_help_string.md
+++ b/changelog_entries/editor_help_string.md
@@ -1,0 +1,2 @@
+ ### Editor
+   * Scrolling the map north or south can move the text that overlays the map, so it doesn't always obscure the southmost hexes of the map. (issue #6422}
--- a/src/editor/controller/editor_controller.cpp
+++ b/src/editor/controller/editor_controller.cpp
@@ -1099,7 +1099,7 @@ void editor_controller::show_menu(const std::vector<config>& items_arg, int xloc
 
 void editor_controller::preferences()
 {
-	gui_->video().clear_all_help_strings();
+	gui_->clear_help_string();
 	gui2::dialogs::preferences_dialog::display();
 
 	gui_->redraw_everything();
--- a/src/editor/editor_display.cpp
+++ b/src/editor/editor_display.cpp
@@ -71,6 +71,22 @@ void editor_display::rebuild_terrain(const map_location &loc) {
 
 void editor_display::pre_draw()
 {
+	// If we're showing hexes near the north of the map, put the help string at the bottom of the screen.
+	// Otherwise, put it at the top.
+	if(help_handle_ != 0) {
+		const bool place_at_top = get_visible_hexes().top[0] > 2;
+
+		if(place_at_top != help_string_at_top_) {
+			const auto& r = font::get_floating_label_rect(help_handle_);
+			double delta = map_outside_area().h - r.h;
+			if(place_at_top) {
+				font::move_floating_label(help_handle_, 0.0, -delta);
+			} else {
+				font::move_floating_label(help_handle_, 0.0, delta);
+			}
+			help_string_at_top_ = place_at_top;
+		}
+	}
 }
 
 image::TYPE editor_display::get_image_type(const map_location& loc)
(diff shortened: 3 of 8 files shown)


stevecotton:
> but that would feature-creep into adding additional functionality to the floating_label support.

Explaining that a bit: Of the accessible functions in floating_label.hpp, the ones that set the position only take relative coordinates, and none return the current position of the label. `get_floating_label_rect` currently returns `{0, 0, size.x, size.y}`. To move the label between top and bottom either means tracking more state in the editor, or expanding the interface of the label support.

Pentarctagon:
I feel like it'd make more sense to add an option to disable it. Having it offset from the bottom like this looks rather odd.

stevecotton:
> I feel like it'd make more sense to add an option to disable it. Having it offset from the bottom like this looks rather odd.

I'd prefer to explore if there's a way to make it "just work" before needing to add buttons and options.

> I've also looked at moving the label between the top and bottom of the screen, but that would feature-creep into adding additional functionality to the floating_label support.

I had another look at doing that, and that's what the latest push does. In the image below, the text location is controlled by the north/south scroll position. If any pixels of the snow or water hexes that you can see on the minimap are on-screen, then the text is at the bottom (same position as in the current 1.16 and 1.17 releases, not offset like the earlier screenshots in this PR), otherwise it's at the top.

For very small maps, it will be at the bottom of the screen in the border outside the map area. For any screen resolution there will be a specific map-height that still has the old issue, but that's an improvement on it happening for every map size.

help_string_mobile_1280x1024

Pentarctagon:
I think that definitely looks better. Would it be reasonable to have it change positions based on whether the mouse cursor is hovering over the bottom text area, or is that not something that'd be easy to figure out?

CelticMinstrel:
Is the "specific map height" one that just barely fits into the viewport?

stevecotton:
> Is the "specific map height" one that just barely fits into the viewport?

Yes, that or a couple of hexes higher. Just realised that zoom level affects this, so zooming in or out a bit can be a workaround for getting at those hexes. Both directions of zoom work for this.

> Would it be reasonable to have it change positions based on whether the mouse cursor is hovering over the bottom text area, or is that not something that'd be easy to figure out?

Now I've realised that the zoom workaround exists, I'm not sure it would be useful to add logic based on the mouse position.

None of it's hard to do, but it's enough work that I'd prefer not to do it. It shouldn't be hard to work out that the mouse is over the bottom couple of hex rows; however I'd need to check if 1.17 does as many pre_redraw() calls as 1.16, or change the logic of when the text box is moved. Also it might cause the box to flicker annoyingly between top and bottom when painting a row near the bottom of the screen, which would need some hysteresis logic that isn't necessary when it's controlled by the north/south scroll position.

Pentarctagon:
Alright, this implementation sounds reasonable to me then.

stevecotton:
Thanks. I'll push a small documentation-only update to the .hpp file.

PR #7363, opened 2023-02-06
Status icons

Elvish-Hunter:
As it was discussed in the development channel, this PR adds three icons (artwork by me; keep in mind that I'm not an artist) for the guardian, unhealable and invulnerable units states; it also makes the invulnerable status formally known by the engine (currently it's checked by string comparison).


Pentarctagon:
The engine change looks fine to me.

CelticMinstrel:
I don't understand why anyone would want an icon for guardian, and I'm also unsure if unhealable is worth having an icon. But I'm fine with the invulnerable icon. Although I'm a little bit surprised you decided to do this in C++ rather than Lua, but I don't think it matters.

knyghtmare (commented):
I still think guardian is misrepresented by a shield, since the behavior is more analogous to "enemy within vision, attack enemy and end up in bad terrain"

CelticMinstrel:
> the behavior is more analogous to "enemy within vision, attack enemy and end up in bad terrain"

Even if that's the _actual_ behaviour, it's certainly not the intended behaviour and should probably be treated like a bug. Either that, or we should deprecate the guardian status.

knyghtmare:
> Even if that's the actual behaviour, it's certainly not the intended behaviour

The current behavior is exploited like this. If you have an Elven Hero guardian on a castle+village hex surrounded by shallow water, and you place any unit on the edge of the shallow water, the guardian will stupidly abandon the 60% defence and heals+8/perturn (village) and attack from shallow water. 
Now ask this: is this a guardian? or a fool? The answer is fool. So, yes...it's a bug that's been retreated as an "expected behavior". So, a status icon for guardian ought to be the clown emoji 🤡 

>  should probably be treated like a bug. Either that, or we should deprecate the guardian status.

You do realize that this has been around since the 1.4.6 days or earlier... right?
Deprecated such a thing is... quite extreme... I'd rather have it fixed. But we would be opening a big can of worms...since some mainline campaigns and UMC (hundreds) rely on this behavior of the guardian status in some of the scenarios...

How I fixed it for my UMC:
1. Added trait "guardian" to the guardian units that keeps unit restricted to castle+village
2. Vision costs are kept the same

CelticMinstrel:
> Deprecated such a thing is... quite extreme...

What are you talking about? Something being around forever doesn't mean it can't be deprecated. Still, fixing it would also be a good idea.

> rely on this behavior of the guardian status

You're telling me that people (even in mainline?) give units the "guardian" status because they want them to ignore terrain defense? I don't really think that's an argument against fixing it.

knyghtmare:
> You're telling me that people (even in mainline?) give units the "guardian" status because they want them to ignore terrain defense?

Yes. we got NR, Liberty, THoT, TRoW, SotA, DW and many more in that list...and almost infinite. The scenarios were tuned with this bugged behaviour as "expected" so fixing or deprecating it would spell doom. So, yeah..a can of nasty worms...

> I don't really think that's an argument against fixing it.
maybe we can simply add a guardian trait for guardian units which keeps them on castles? actually...might not do well in cases where guardian elven rangers are present in forests...

CelticMinstrel:
> The scenarios were tuned with this bugged behaviour as "expected" so fixing or deprecating it would spell doom.

Well no, given this, I think deprecating it is the correct answer. It doesn't do what you'd expect, yet tons of people have designed around expecting the unexpected behaviour, so fixing it is essentially impossible. Therefore we should aim to eventually get rid of the guardian status altogether and recommend people use the guardian MicroAIs instead.

And if it's probably going to be deprecated, that now makes two reasons not to give it an icon.

> Actually, we're debating out of the PR scope now...or are we?

Possibly a little, though it's not entirely unrelated, because it links back to the question of whether we should add an icon for it.

knyghtmare:
> guardian: if 'yes', the unit will not move, except to attack something in the turn it moves (so, it only can move if an enemy unit gets within range of it). Does the same as ai_special = "guardian".

This is from the Wiki. This is the actual behavior but is inconsistent with ideology associated with the term "guardian". So, it's more of a term change issue.

CelticMinstrel:
> is inconsistent with ideology associated with the term "guardian"

It may be overly simplistic but I don't think it's inconsistent.

Elvish-Hunter:
> I'm also unsure if unhealable is worth having an icon

My reasoning here was this: imagine that one of your enemy units has an ability, or weapon special, that can make your units unhealable (and hopefully, one of your units has an ability that cam make them healable again); wouldn't you want to have a way to check which ones of your units are affected without having to go into debug mode and recall the inspection window?
In this regard, the unhealable status is different from the other un* states (unpoisonable, undrainable, unplaguable), because those are usually associated with another trait (undead, mechanical, elemental); it's more like the petrified status.
Fun fact: while making the un* list, I almost wrote **unslowable** and **unpetrifiable** too, before I remembered that *those two don't exist at all*. :-P

> I still think guardian is misrepresented by a shield, since the behavior is more analogous to "enemy within vision, attack enemy and end up in bad terrain"

What icon should represent it (excluding for a moment the "no icon" option)?

stevecotton:
I'd agree with having an icon that reflects what "guardian" does, rather than what it sounds like. A bull's head would probably be hard to read in a small icon, maybe a set of arrows under the ZoC?

CelticMinstrel:
> I'd agree with having an icon that reflects what "guardian" does, rather than what it sounds like.

I think this is just doubling down on a bad idea. If it doesn't do what it sounds like, that's going to cause all kinds of confusion. If an icon then reflects what it actually does, I don't think it's going to help the confusion at all.

doofus-01:
These deprecation arguments remind me of something, but sticking to the "guardian" status icon - isn't "guardian" just an AI shortcut?  Is it really a status that needs an icon?  I guess it could be like what some wasps do to cockroaches, but how does it normally show up in Wesnoth?

Elvish-Hunter:
Since the general opinion is that the guardian status shouldn't have an icon, or at least nobody is sure that it should have an icon, I'm going to remove it from this PR.

Elvish-Hunter:
@doofus-01 Are the icons for the invulnerable and unhealable states OK?

src/units/unit.hpp
```diff
@@ -860,14 +860,15 @@ class unit : public std::enable_shared_from_this<unit>
...
-		STATE_GUARDIAN,   /** The unit is a guardian - it won't move unless a target is sighted */
-		STATE_UNKNOWN = -1/** A status effect not known to the engine */
+		STATE_SLOWED = 0,   /** The unit is slowed - it moves slower and does less damage */
+		STATE_POISONED,     /** The unit is poisoned - it loses health each turn */
+		STATE_PETRIFIED,    /** The unit is petrified - it cannot move or be attacked */
+		STATE_UNCOVERED,    /** The unit is uncovered - it was hiding but has been spotted */
+		STATE_NOT_MOVED,    /** The unit has not moved @todo Explain better */
+		STATE_UNHEALABLE,   /** The unit cannot be healed */
+		STATE_GUARDIAN,     /** The unit is a guardian - it won't move unless a target is sighted */
+		STATE_INVULNERABLE, /** The unit is invulnerable */
```
stevecotton, line 870: This is a bad comment, it's just repeating the words in the thing being documented.Compare that with the other items in this list.

Does it never get hit, or does it get hit but take 0 damage? Can it suffer other status effects from being hit?

CelticMinstrel: It never gets hit, so no, it cannot suffer other status effects from being hit.

doofus-01:
> @doofus-01 Are the icons for the invulnerable and unhealable states OK?

To be honest, I'd have a hard time knowing what those were supposed to depict.  That said, they do look distinct, so maybe merging them and waiting for player feedback would work?  The concept seems solid, so if needed we can change the icons before 1.18.

doofus-01 (approved):

Elvish-Hunter:
Since all the issues with this PR seems to have been fixed, I'm merging it.

src/reports.cpp
```diff
@@ -353,6 +353,14 @@ static config unit_status(reports::context & rc, const unit* u)
 		add_status(res, "misc/petrified.png", N_("petrified: "),
 			N_("This unit has been petrified. It may not move or attack."));
 	}
+	if (u->get_state(unit::STATE_UNHEALABLE)) {
+		add_status(res, "misc/unhealable.png", N_("unhealable: "),
+			N_("This unit is unhealable. It cannot be healed by healers or villages and doesn't benefit from resting."));
```
Wedge009, line 358: `doesn't` needs to use the typographic apostrophe: @Pentarctagon is there a string freeze current?

Pentarctagon: Nope.

Wedge009: Thanks. 90ef6bcc23606ac03b411a90436dec505a6a2f13

PR #7368, opened 2023-02-08
Small achievements cleanup

Pentarctagon:
---

@CelticMinstrel There isn't really much going on here, but I just wanted to have my use of `luaW_pcall` double checked before merging.

### DIFF
--- a/src/scripting/game_lua_kernel.cpp
+++ b/src/scripting/game_lua_kernel.cpp
@@ -3184,14 +3184,15 @@ int game_lua_kernel::intf_get_achievement(lua_State *L)
  * - Arg 2: string - id.
  * - Arg 3: int - the amount to progress the achievement.
  * - Arg 4: int - the limit the achievement can progress by
- * - Ret 1: int - the achievement's current progress after adding amount, -1 if completed, or -2 if not found or not a progressable achievement
+ * - Ret 1: int - the achievement's current progress after adding amount, -2 if not found, -1 not a progressable achievement (including if it's already achieved)
+ * - Ret 2: int - the achievement's max progress, -2 if not found, or -1 not a progressable achievement
  */
 int game_lua_kernel::intf_progress_achievement(lua_State *L)
 {
 	const char *content_for = luaL_checkstring(L, 1);
 	const char *id = luaL_checkstring(L, 2);
 	int amount = luaL_checkinteger(L, 3);
-	int limit = luaL_checkinteger(L, 4);
+	int limit = luaL_optinteger(L, 4, 999999999);
 
 	for(achievement_group& group : game_config_manager::get()->get_achievements()) {
 		if(group.content_for_ == content_for) {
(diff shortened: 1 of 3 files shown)


src/scripting/game_lua_kernel.cpp
```diff
@@ -3183,37 +3193,42 @@ int game_lua_kernel::intf_progress_achievement(lua_State *L)
...
-
 	for(achievement_group& group : game_config_manager::get()->get_achievements()) {
 		if(group.content_for_ == content_for) {
 			for(achievement& achieve : group.achievements_) {
 				if(achieve.id_ == id) {
-					// found the achievement
+					// check that this is a progressable achievement
+					if(achieve.max_progress_ == 0) {
+						ERR_LUA << "Attempted to progress achievement " << id << " for achievement group " << content_for << ", which does not have max_progress set.";
+						lua_pushinteger(L, -2);
```
CelticMinstrel, line 3203: That's not mentioned in the doc comment.

src/scripting/game_lua_kernel.cpp
```diff
@@ -3174,7 +3184,7 @@ int game_lua_kernel::intf_get_achievement(lua_State *L)
  * - Arg 2: string - id.
  * - Arg 3: int - the amount to progress the achievement.
  * - Arg 4: int - the limit the achievement can progress by
- * - Ret 1: WML table returned by the function.
+ * - Ret 1: int - the achievement's current progress after adding amount, -1 if completed, or 0 if not found
```
CelticMinstrel, line 3187: Doesn't it make more sense to return max_progress if it's completed?

src/scripting/game_lua_kernel.cpp
```diff
@@ -3183,37 +3193,42 @@ int game_lua_kernel::intf_progress_achievement(lua_State *L)
...
-			luaW_pushconfig(L, cfg);
+			lua_pushinteger(L, -2);
 			return 1;
 		}
 	}
 
 	// achievement group not found
 	ERR_LUA << "Achievement group " << content_for << " not found";
-	luaW_pushconfig(L, cfg);
+	lua_pushinteger(L, -2);
```
CelticMinstrel, line 3231: These two appear to contradict the doc comment.

src/scripting/game_lua_kernel.cpp
Pentarctagon: Updated.

src/scripting/game_lua_kernel.cpp
Pentarctagon: Fixed.

src/scripting/game_lua_kernel.cpp
Pentarctagon: That wouldn't be especially useful by itself, since the caller wouldn't know whether the value returned is the max_progress or not.

CelticMinstrel: They can obtain the max_progress separately, or you could return it as a second return value.

CelticMinstrel:
> I just wanted to have my use of `luaW_pcall` double checked before merging.

This part looks correct to me. I assume you've tested it too.

Pentarctagon:
Yep, I tried it with the tutorial completion achievement and it worked as expected.

src/scripting/game_lua_kernel.cpp
Pentarctagon: So that'd be
```
 * - Ret 1: int - the achievement's current progress after adding amount or -1 if not found
 * - Ret 2: int - the achievement's max progress or 0 if not a progressable achievement
```
?

CelticMinstrel (approved):
That works. The Ret 2 could be omitted entirely when negative, but it doesn't hurt anything to just double the first return value in that case.

PR #7372, opened 2023-02-10
Add achievements button to titlescreen.

Pentarctagon:


soliton-:
I think this could also be done as for example a trophy 🏆 icon and placed next to language selection or so. No need to add more to the menu and no need to translate anything.

knyghtmare:
> I think this could also be done as for example a trophy 🏆 icon and placed next to language selection or so. No need to add more to the menu and no need to translate anything.

Plus One to vote

stevecotton:
I like the idea of a trophy icon, but it would still need one string translated because it appears in the preferences menu's Hotkeys settings. Doesn't seem like a saving to avoid translating "View Achievements" either.

CelticMinstrel:
Yeah, a trophy icon is a nice idea, but it has no effect on needing a translation. It would also appear in the menu in-game if we added it there.

Pentarctagon:
Do we have a trophy icon that can be used?

Pentarctagon:
So, unless there's a trophy icon available to use, I'm going to go ahead and merge this tomorrow. If a trophy icon is added later, then the suggested UI changes here can of course be made.

PR #7374, opened 2023-02-10
Editor: move the text at the bottom of the screen

stevecotton:
This is the forwardport of #7362. I'm now seeing that more cleanup is possible, but I'd prefer to merge this close to as-is, so that any cleanup that happens in later commits is separate and easily reviewable for (if necessary) backporting to 1.16 too.

Fixes: #6422

Second commit (cherry-pick of 1.16's 29d86d2c369483c45aa599076eb076c5792b7ef3)
---

Editor: let the help text switch between the top and bottom of the screen

If the northmost hexes of the map are visible, put the text on the bottom; if
the southmost are visible, put it on the top. Also, make the text smaller if
it's very wide.

It used to be fixed to the bottom of the screen, but this meant that it always
obscured the southmost hexes of the map, and the only way to avoid that is to
scroll east/west, or to add an extra row on the south. Similarly, having it at
the very top obscures the northmost hexes.

Moving it to any fixed location that isn't the edge of the screen makes it
possible to edit all the hexes, but also makes it more annoying when working
elsewhere on the map; that trade-off was rejected in review.

For very small maps, it will be at the bottom of the screen in the border
outside the map area. For any screen resolution there will be a specific
map-height that still has the old issue, but that's an improvement on it
happening for every map size.

First commit (corresponds to 1.16's 76306d3fb61da9a046087fc862d960a6ac5d4cad)
---

Move set_help_string to editor-only code

No behavior changes, this is just separating it into the editor directory.

This corresponds to 1.16's 76306d3fb61da9a046087fc862d960a6ac5d4cad, but has
additional cleanup outside the editor directory, including removing the unused
code from floating_label.cpp. Even within the editor, there are differences
because of the graphics refactor.

### DIFF
--- a/src/editor/editor_display.cpp
+++ b/src/editor/editor_display.cpp
@@ -18,12 +18,15 @@
 #include "draw.hpp"
 #include "editor/controller/editor_controller.hpp"
 #include "editor/editor_display.hpp"
+#include "floating_label.hpp"
+#include "font/sdl_ttf_compat.hpp" // for pango_line_width
 #include "lexical_cast.hpp"
 #include "overlay.hpp"
 #include "reports.hpp"
 #include "team.hpp"
 #include "terrain/builder.hpp"
 #include "units/map.hpp"
+#include "video.hpp"
 
 namespace wb {
 	class manager;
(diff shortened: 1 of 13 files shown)


src/editor/editor_display.cpp
```diff
@@ -139,4 +159,44 @@ display::overlay_map& editor_display::get_overlays()
...
+		} else {
+			break;
+		}
+	}
+
+	const int border = 5;
+
+	font::floating_label flabel(str);
+	flabel.set_font_size(size);
+	flabel.set_position(canvas_size.x / 2, canvas_size.y);
```
stevecotton, line 189: The comment about possible cleanup was prompted by this line, which could avoid the need to call `font::move_floating_label` later in this function.

PR #7384, opened 2023-02-15
Further consolidate logging functionality between Windows and non-Windows platforms.

Pentarctagon:
For non-Windows platforms this adds:
* a new check for whether the logs directory exists and is writable by attempting to write a single whitespace to a dummy.log file.
* a popup shown immediately after the GUI system is initialized letting the player know if there was an issue creating the log file.

For Windows, this changes:
* instead of writing to the OS temp directory and then attempting to move that file to the logs directory, this uses the same check that was described above for non-Windows platforms.
* the alert shown when writing the dummy log file fails is now a Wesnoth alert message rather than the Windows-specific MessageBox.
* a failure to create a log file does not immediately exit wesnoth anymore.

---

@knyghtmare or @Wedge009 - Can one of you test the following on Windows?:
* By default, logs are still correctly written to the log file.
* Running with `--wconsole` still writes the output to the terminal.


CelticMinstrel:
I guess this method works, but it does leave me wondering why you didn't just use `filesystem::status` to determine if the directory is writable.

Pentarctagon:
There is no such method though? Also I figured that given the issues around OneDrive managed directories not being handled transparently, the best way to check if the log directory is writable is to try to actually write a file there.

CelticMinstrel:
Sorry, I misremembered the namespace alias. It's `std::filesystem::status` which I'm guessing translates to `bfs::status` or something in Wesnoth.

Wedge009:
Testing `Pentarctagon/master` on Windows. `--wconsole` appears to run as expected, with output to the terminal. No arguments fails to run whatsoever, dumps _two_ empty logs plus a single character `dummy.log`.

Running a debug build, I get an unhandled exception on the `<<` operator call in `log_in_progress::operator|()` in `log.cpp`:
https://github.com/wesnoth/wesnoth/blob/7477aab1bb2b3473b176d02ce2fa8a8be2368058/src/log.cpp#L404
(called from `wesnoth.cpp`, `SDL_main()` at line 1162).

`stream_` appears to be a valid output stream, `prefix_` is an empty string, `message` includes the string `Battle for Wesnoth v1.17.2+dev x86_64`.

I'm not sure what all this means, but suffice to say I think this needs more work for Windows.

Pentarctagon:
@Wedge009 Once the CI passes, can you try again with the new commit? This should (I think) work since now Windows and non-Windows is logging to file with exactly the same code path.

Pentarctagon:
The CI is fine - the failure is the glib issue again.

Wedge009:
`--wconsole` works the same.

No arguments works again, but with the addition of a `dummy.log` output. I assume that's due to `check_log_dir_writable()`. Is it desirable to leave the dummy file present after the checks are done? I'd prefer to get rid of it.

Pentarctagon:
Nice, I'm glad that worked.

Updated it to delete dummy.log at the end of checking whether the logs directory exists and is writable.

Wedge009:
That looks good to me now.

Pentarctagon:
Alright, I'll merge it tomorrow then unless any other issues are brought up.

PR #7387, opened 2023-02-16
[Terrains] Add treehouse villages

### LINKED ISSUE #7385: Merge Treehouse villages from UMC Terrain Pack into mainline

Was scrolling through the UMC Terrain Pack, and noticed these extremely beautiful treehouse villages, in my opinion the best terrain in the entire add-on. I believe these villages deserve to be mainlined, as their quality and amount of variations is more than enough to be on par with existing village terrain. Plus it would make designing elf maps a lot more interesting, instead of having to use the same old elf villages that veteran players already seen hundreds of times.

image

knyghtmare:
From the UMC Terrain Pack. Closes #7385


CelticMinstrel:
Hmm… feels like something is missing here? The version of these on my hard drive has "spring" and "fall" variations as well…

knyghtmare:
> Hmm… feels like something is missing here? The version of these on my hard drive has "spring" and "fall" variations as well…

image

All I got here

knyghtmare:
I found these from another UMC

image

I can see what you meant by the spring and fall variations. Well I guess I can these in as extra commits

Pentarctagon:
The style seems a bit different compared to other villages.

stevecotton:
Do we know who the original authors are, and the license that these are released under?

The UMC Terrains Pack lacks authorship data for the images, and contains a copy of the GPLv3 alongside the server-added GPLv2, without any comment about whether the intent was to license under GPLv2+ rather than exclusively under GPLv3.

I'm a bit wary because they don't look like they were drawn to fit in with Wesnoth's style, nor has the artist slipped in some hexagons to subtly indicate that they expect to appear in a hexagon based game.

CelticMinstrel:
> Do we know who the original authors are, and the license that these are released under?

They were originally posted [on the forums](https://forums.wesnoth.org/viewtopic.php?t=33573).

doofus-01 (commented):
Alternatives to the old elf villages certainly would be a good addition, and tree houses make sense, but I can't green-light this as is.  As Pentarctagon says, it's a different style.  Wesnoth terrain scale does have a lot of inconsistency, but the villages/houses do generally fit a certain pattern.  (The dark merman villages are more like these tree houses, but that probably means they should be updated, rather than added to.)

The base tree house is actually pretty close, wouldn't need much changed, if anything.  The others look more like lanterns, which is not necessarily a bad thing.  Some sort of elf-theme scenery could be useful.

knyghtmare:
Closed due to no positive response.

PR #7388, opened 2023-02-16
Display "any" availability instead of "none"

soliton-:
Feel free to suggest better wording.

### DIFF
--- a/src/help/help_impl.cpp
+++ b/src/help/help_impl.cpp
@@ -769,8 +769,8 @@ std::vector<topic> generate_trait_topics(const bool sort_generated)
 		text << "\n\n";
 		if (trait["availability"] == "musthave") {
 			text << _("Availability: ") << _("Must-have") << "\n";
-		} else if (trait["availability"] == "none") {
-			text << _("Availability: ") << _("Unavailable") << "\n";
+		} else if (trait["availability"] == "any") {
+			text << _("Availability: ") << _("Available also for leaders") << "\n";
 		}
 		topics.emplace_back(name, id, text.str());
 	}


Pentarctagon:
What effect does this change have, exactly?

CelticMinstrel:
The "none" availability doesn't do much – as far as I know, it just ensures that the trait is not included in the global trait pool (which is only relevant if it _is_ a global trait, as opposed to a race or unit trait). In other words, it means it won't be chosen randomly, but I suppose Soliton feels it's misleading to call that "unavailable". After all, it can still be added explicitly without issue.

The fact that "any" wasn't shown at all seems kind of strange to me, so I think making it show is a good thing. That's not really related to making "none" not show.

soliton-:
The "none" availability does nothing at all. Only "any" and "musthave" are checked and acted upon.

The text that is changed here is displayed in the help under `Traits` -> `undead` for example where it says `Availability: Must-have`.

CelticMinstrel:
> The "none" availability does nothing at all.

It does do something simply by virtue of being a value that's not checked, though. It means anything that checks the trait availability won't use this one.

soliton-:
I don't really get what you're trying to say. Setting availability to "none" is the same thing as not setting it at all. Where do you see that it will prevent the trait from being chosen?

CelticMinstrel:
Hmm. It does appear that I was mistaken… I definitely had a memory that "none" was suppose to make the random trait picker skip over it, but the code does not appear to do anything like that.

stevecotton:
Despite being player-visible text, neither the new nor the old nor the existing "Must-have" are player-readable wordings.

Better would be:

* All Undead units have this trait.
    - With add-ons installed, that might turn into "All Undead, Undead, Undead, Ancient Phantom, and Yet Another Race That's Called Undead units have this trait." :(
* This trait may be randomly assigned to Elves and Quenoth Elves, although it will not be randomly assigned to leaders.
* This trait may be randomly assigned to units of most races; it may also be randomly assigned to leaders.

soliton-:
Are you saying the availability note should be removed? Changing the help text of the traits is not in scope of this PR. I can't really tell what you're suggesting.

stevecotton:
I mean changing the text of the "availability: must-have" and "availability: none" notes.

soliton-:
To something about undead and elves? What does that have to do with the availability attribute?

stevecotton:
Some are only available to specific races.

soliton-:
But that has nothing to do with the availability attribute...

I suppose we can remove the availability note and change the help text for each trait in the way you hinted at.

CelticMinstrel:
I don't think the current architecture is supportive of stevecotton's suggestion. When there are traits only available to certain races, each race has its own copy of the trait. I'd suggest instead opening that as an issue and merging this as-is.

soliton-:
I mean, while generating help pages for each unit type (where we list the possible traits) we could also build a list of unit types per trait... But this is certainly not in scope for this PR.

In scope would be changing the word availability though since that seems to be confusing? Really what the attribute does is influence what traits can be picked from a pool of traits that is determined by global traits (under [units]), the race and/or unit type config. So for a given unit type it does not decide what that trait pool looks like but what can then be picked from that pool.

Then `musthave` means any trait with that availability attribute is chosen no matter what. Even if there are more than num_traits of them. And `any` means the trait can be (randomly) chosen for leaders. Without traits with availability `any` leaders get no traits (apart from `musthave` ones) as is the case in mainline. So not because they're leaders per se but because there are no suitable traits available.

It's a bit of an odd design. There is the occasional complaint that it's lame that leader units are weaker than their recruitable versions since they get no traits (in mainline). If we wanted to change that it'd be more useful to have traits exclusively available to leaders though. For example we could add a "heroic" trait that is better than the usual traits with such an availability to the global trait pool and then all leaders would get it.

stevecotton:
> I mean, while generating help pages for each unit type (where we list the possible traits) we could also build a list of unit types per trait... But this is certainly not in scope for this PR.

I would agree, but the trait list already checks which units have been discovered. About 40 lines above the lines already being changed in this PR, it goes through the list of all units to work that out.

I've got a prototype going, will open a draft PR tonight.

stevecotton:
* data/core/macros/traits.cfg:#define TRAIT_FEARLESS
* data/core/macros/traits.cfg:#define TRAIT_FEARLESS_MUSTHAVE

Both of those have `id=fearless`, but the one used by Ghouls (and a few others) also has `availability=musthave`.

stevecotton:
A simple availability note will always be misleading, because traits such as Fearless are must-have for some units (Ghouls and Walking Corpses), and randomly available for other units (Heavy Infantry).

I think we should either remove the note, or add the complexity of #7407.

soliton-:
Well, that is our fault for using the same ID for different traits. I'm fine with removing anyway. It's not particularly useful info.

CelticMinstrel:
???

What on earth are you talking about? They're not different traits. They're the same trait. That's _why_ they have the same ID.

soliton-:
It's two very similar traits with a clear difference in the availability attribute. If it were the same trait we wouldn't need two definitions.

You can argue that the availability attribute shouldn't be part of the trait definition since it just affects when the trait is chosen not how it works but that's not the way it is implemented currently. Certainly once we remove the availability note from the help the difference matters even less.

CelticMinstrel:
I don't see it that way. The availability attribute is clearly not part of the "identity" of the trait. Yes, you're right that it probably doesn't belong in the trait tag, technically, but regardless of implementation, it's still not part of the trait's "identity".

Every other detail between `{TRAIT_FEARLESS}` and `{TRAIT_FEARLESS_MUSTHAVE}` is identical. Therefore, they are the same trait.

PR #7399, opened 2023-02-20
Resolve Issue 7398: Crash on Attack with Planning Mode

Dyrcona:
The game crashes when the player clicks the Attack button with planning mode with a console error about a failed assert on line 164 of src/whiteboard/attack.cpp.  The assert uses > but it looks like it should be >=.

This commit switches from > to >= and appears to resolve the issue.


CelticMinstrel:
Hmm, yeah, that makes sense.

Wedge009:
Any reason not to merge this?

PR #7407, opened 2023-02-21
Show availability on traits' help pages

stevecotton:
This is very much just prototyping for the moment. It started as a "quick test of a review comment for #7388", and grew....

This logic was seeming too complex, until the TRAIT_FEARLESS_MUSTHAVE macro cropped up. As this commit notes on the Fearless trait's page:

Must-have for units: Ghast, Ghoul, ...
Available for non-leaders of race: Troll
Available for non-leaders of type: Dune Paragon, HI, ...


stevecotton:
Rejected in favour of removing the text entirely, as done in #7422. I'm doing that in a separate PR, as it's the opposite of this.

PR #7411, opened 2023-02-22
WoF make recall of some unit type trees undoable

Toranks:
The way it's written prevents undoing the recall action of the filtered unit trees (Drake Arbiter, Drake Thrasher and Drake Warrior). It is better to activate the event only if the other conditional is also fulfilled, it is unnecessary in the rest of the cases.
What I have done is simply moved the conditional from [if] to the [filter_condition] of the event itself

Just to clarify for those who don't know, the code makes the unit loyal if the original loyal of the same type tree died or was removed from the recall list. I've tested all cases with the new filters and works fine.

### DIFF
--- a/data/campaigns/Winds_of_Fate/utils/macros.cfg
+++ b/data/campaigns/Winds_of_Fate/utils/macros.cfg
@@ -8,23 +8,21 @@
         [filter]
             type_adv_tree=Drake Warrior
         [/filter]
-        [if]
+        [filter_condition]
             [not]
                 [have_unit]
                     role=fighter_intendant
                     search_recall_list=yes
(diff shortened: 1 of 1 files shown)


Jonathan-Kelly:
Huh, after all these years I had no idea you could undo a recall (though it makes sense in retrospect).

The change looks good and tests well. Thanks!

stevecotton:
Seems like it rewards getting the original loyals killed though - they get replaced with a 3-trait loyal.

However, removing one of the existing traits would be even worse.

stevecotton:
Looks good to me then.

nemaara (approved):

PR #7416, opened 2023-02-24
Fixup, complete, and enable the lobby's game history viewer.

Pentarctagon:
At some point in the future I'd like to come back and add some search parameters - right now it shows the game history of the currently selected player in the lobby, but ideally it would instead have options for player name, era, scenario, etc.

---

The replay link also only opens the player's browser which downloads the replay file. Ideally it would be downloaded by the client directly to the saves folder, but as far as I know there's currently no way to do that and this seemed like enough for a PR.

The UI:
image


data/gui/window/mp_match_history.cfg
```diff
@@ -1,5 +1,40 @@
...
+                        horizontal_alignment = "left"
+
+                        [label]
+							id = "{ID}_game_name"
+                            definition = "default"
+                        [/label]
+
+                    [/column]
+
+                [/row]
```
stevecotton, line 29: There's a mix of tabs and spaces in this file.

src/gui/dialogs/multiplayer/lobby.cpp
```diff
@@ -705,6 +706,16 @@ void mp_lobby::post_show(window& /*window*/)
 	plugins_context_.reset();
 }
 
+void mp_lobby::open_match_history()
+{
+	const mp::user_info* info = player_list_.get_selected_info();
+	if(info)
+	{
```
stevecotton, line 713: Minor, but this brace should be on the previous line.

data/gui/window/mp_match_history.cfg
Pentarctagon: Wasn't the whitespace fixup script changed to be able to catch this?

stevecotton: It was for Lua files, I expected wmlindent to cover .cfg files. Forgot that wmlindent isn't run on the GUI files...

stevecotton: I'm looking at how many changes it would need to get the GUI files consistent (just on a per-file basis, using either one or the other). You'll note that those files aren't part of #7424, I'm still looking at them...

src/server/wesnothd/server.cpp
```diff
@@ -1867,34 +1895,9 @@ void server::handle_player_in_game(player_iterator p, simple_wml::document& data
...
-
-			if(player_id != 0) {
-				LOG_SERVER << "Querying game history requested by player `" << player.name() << "` for player id `" << player_id << "`.";
-				user_handler_->async_get_and_send_game_history(io_service_, *this, p, player_id, offset);
-			}
-		}
-		return;
+	}
 	// Data to ignore.
-	} else if(
```
soliton-, line 1897: Is there a reason for this formatting change? (While it's just formatting I'd like to keep non-essential changes to a minimum in general. Makes for cleaner history.)

Regarding removing the function from in game handling, would it make sense to handle the request from either lobby or in game? I haven't looked at the client side yet. Perhaps more difficult to support it there but I wouldn't mind if the server supports either.

soliton- (approved):

src/server/wesnothd/server.cpp
Pentarctagon: It wasn't intentional, it's just how I happened to cut/paste it to the lobby handler.

Having it available in-game wouldn't be as useful since it would only ever show your own game history since there's not yet a way on the dialog itself to provide a player name. There would also need to be a hotkey and/or menu item to make the request (which doesn't exist yet either) as well as a check to not try to make the request in single player.

Vultraz (approved):

PR #7418, opened 2023-02-25
WoF S03,S08 - Added animation when units go to recall list

### LINKED ISSUE #7410: WoF S03 lack of animation when any unit returns to recall list

All

### Description of the problem

When a unit dies, it is sent to the recall list, but in an abrupt and ugly way. It suddenly disappears right on the last hit.

- WoF Scenario 3: The Contention
- Kill any unit

### Expected behavior

Some animation would be nice, I suggest this change just before this line:
https://github.com/wesnoth/wesnoth/blob/7b6027cdec232624a24a04895853684527278e54/data/campaigns/Winds_of_Fate/scenarios/03_The_Contention.cfg#L332
```
		[animate_unit]
			[filter]
				id=$unit.id
			[/filter]
			flag=pre_teleport
			with_bars=no
		[/animate_unit]
```
pre_teleport is the animation that best fits what is happening. Well configured, the movement animation could also work, but it would be a little more complicated to make it look good

### Additional context

_No response_

Toranks:
This resolves #7410  and the same situation on S08 Overlook, when the units escapes from the cave on 31,1
Uses the pre_teleport animation (in the absence of another specific animation for drakes), which does a smooth fade effect and is automatically generated on any unit.

### DIFF
--- a/data/campaigns/Winds_of_Fate/scenarios/03_The_Contention.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/03_The_Contention.cfg
@@ -329,6 +329,13 @@ You shall learn a pivotal lesson in humility!"
                 id=Gorlack,Karron
             [/not]
         [/filter]
+        [animate_unit]
+            [filter]
+                id=$unit.id
+            [/filter]
+            flag=pre_teleport
+            with_bars=no
+        [/animate_unit]
         [put_to_recall_list]
             id=$unit.id
             heal=yes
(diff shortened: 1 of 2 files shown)


Jonathan-Kelly:
This is a nice enhancement. Definitely adds polish.

nemaara (commented):
It could look nice but if you have a lot of units it gets a bit annoying to wait through? Should add the option to skip the animation with esc if possible.

Toranks:
Wait? It is a half second animation and can also be sped up with shift key like any other default animation.

nemaara:
I guess you theoretically shouldn't be losing that many units anyway and it would replace the death anim so alright.

nemaara (approved):

Issue #5709, opened 2021-04-15
The schema isn't formatted with wmlindent


stevecotton:
The files in data/schema/... are indented with tabs instead of spaces, running wmlindent on them generates around 5000 lines of purely whitespace changes. I suggest we run it and then enable checking the formatting during CI runs.

However, let's not do that right now, as it will generate merge conflicts with @CelticMinstrel's #5663. Reviewing that PR is the only reason I noticed this.

CelticMinstrel:
Just for the record, it won't generate merge conflicts with #5663 - the schema commits in that PR will disappear once I rebase the PR onto master, as they've already been committed to master.

I'm not exactly in favour of this, given that I prefer tabs over spaces everywhere. I do understand that having _consistent_ formatting is helpful, though, and if that's the goal, then schema isn't the only culprit - all the GUI WML also uses tabs.

shikadiqueen:
> all the GUI WML also uses tabs.

That would be because mordante also favoured tabs, plus the C++ and WML are tightly coupled and the C++ uses tabs.

Also, please **DO NOT** this as an invitation to change the GUI2 WML indentation style. I've got plenty of local branches and stashes that rely on it not changing.

CelticMinstrel:
If it were up to me, I'd change the wmlindent style to tabs instead. It's not really up to me, though.

shikadiqueen:
Yeah, that's just not gonna happen simply because it would be bad form for us to go back on the promise we made back then. That's the whole reason why wmlindent doesn't allow you to choose tabs instead of spaces.

CelticMinstrel:
That promise was already broken when mordante decided to use tabs though.

shikadiqueen:
No, it wasn't. GUI2 WML is an implementation detail of the game and people can and should use spaces in their own add-ons.

CelticMinstrel:
Um… add-ons are out-of-scope for wmlindent…

shikadiqueen:
No, they are not. It is designed for both mainline content and add-ons.

CelticMinstrel:
[...]

PR #7424, opened 2023-02-27
Fix mixed indents in the schema, and verify new additions

stevecotton:
This uses tabs instead of spaces, because although that's not our coding standard, it is only a few lines of changes.

The alternative of running wmlindent on these files changes 4700 lines; while I'm in favor of doing it, it's discussed in #5709.


CelticMinstrel:
I don't quite understand why you need my review when you only changed whitespace, but go right ahead.

PR #7427, opened 2023-02-28
UtBS S05 add logic to the main cavern event so scouting is taken into account

### LINKED ISSUE #7394: UtBS S05: Difficulty is highly dependent on where Kaleh is when the main cavern event triggers

- **Version:** 1.15.1+dev (but I've checked the scenario and this would still apply to 1.17.13+dev)

### Description of the bug

The difficulty in S05 is heavily influenced by how long it takes Kaleh to reach the keep after the main cavern is opened. That timing depends on whether Kaleh was (or was near) the explorer who triggered the moveto event, and how many movement points Kaleh has remaining at the time. Having someone scout forwards makes the scenario much harder; not impossibly harder, but "ohh, this must be intended to be an unusually hard scenario" harder.

A possible solution (particularly for players who are used to the tailwind ability) would be a narrator dialogue, "We assume all the Quenoth elves moved to the cavern mouth during that discussion. For this turn only, they have extra movement points so that you can move them into the cavern." along with extra MP based on how far away they were.

Originally reported, with replays, at https://r.wesnoth.org/p647258 . Just realised that I never logged an issue for it, so I'm logging it now as the Quenoth rework will mean a major retest of the balance.

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Descacharrado:
Closes #7394 

Line 846 (changing the position of the troll) is meant as a fix to a borderline issue, where Kaleh is in a position where he can't reach the keep if the player allies the dwarves, but can if he allies the trolls (because Zone of Control) - a different approach could be setting the troll in 30, 28. Not a major change anyways.

As seen in the picture, Kaleh can't get to the keep via the 6-moves that is represented with the blue line, as it crosses two tiles with ZoC and he can't reach either via the 7-moves represented with the green line, as it is too far away.
ZoCThen
Kaleh would be able to reach the keep with the new position of the troll.
ZoCNow

Line 933 is a debug message, useful for debugging or playtesting later.

Lines 934 - 973 
- If kaleh is the explorer, he just says "Whoa." and everything else in these lines is skipped
- If other unit is the explorer and
    1. the explorer does not reach too far out (x=26 or beyond). Then Kaleh moves up to the explorer
    2. the explorer reaches too far out. Then Kaleh moves to 24,28 (the first tile that triggers the moveto event)
  
Lines 974-1005 Kaleh's movement points are fully restored in EASY and MEDIUM, partially restored (only if necessary) in HARD

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/05_A_Subterranean_Struggle.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/05_A_Subterranean_Struggle.cfg
@@ -946,7 +946,7 @@
                 [message]
                     speaker=$explorer.id
                     #po: "this" refers to the trolls and dwarves fighting
-                    message= _ "Whoa. Kaleh, you have to come watch this."
+                    message= _ "Whoa. Kaleh, you have to come see this."
                 [/message]
                 # if the explorer is too far ahead (x>26), put kaleh safely behind
                 [if]
--- a/data/campaigns/Under_the_Burning_Suns/scenarios/05_A_Subterranean_Struggle.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/scenarios/05_A_Subterranean_Struggle.cfg
@@ -843,7 +843,7 @@
 
         #2 troll whelps
         {DEFENDER 3 (Troll Whelp) 31 30 (Troll Defender) ( _ "Troll Defender")}
-        {DEFENDER 2 (Troll Whelp) 29 28 (Troll Defender) ( _ "Troll Defender")}
+        {DEFENDER 2 (Troll Whelp) 29 29 (Troll Defender) ( _ "Troll Defender")}
 
         #western reinforcements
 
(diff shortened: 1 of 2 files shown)


data/campaigns/Under_the_Burning_Suns/scenarios/05_A_Subterranean_Struggle.cfg
```diff
@@ -930,10 +930,79 @@
...
+                [message]
+                    speaker=$explorer.id
+                    message= _ "Whoa."
+                [/message]
+            [/then]
+            [else]
+                [message]
+                    speaker=$explorer.id
+                    #po: "this" refers to the trolls and dwarves fighting
+                    message= _ "Whoa. Kaleh, you have to come watch this."
```
nemaara, line 949:
```suggestion
                    message= _ "Whoa. Kaleh, you have to come see this."
```

nemaara (approved):

Issue #7371, opened 2023-02-09
Revert b780da4


soliton-:
This [commit](https://github.com/wesnoth/wesnoth/commit/b780da4) was based on the loadscreen threading which has been removed.

CelticMinstrel:
Ah yes, I was wondering about that too awhile ago.

SyedSaifuddin045:
Is anyone working on this issue?
if not please assign it to me.

PR #7428, opened 2023-03-02
removed mutex according to issue #7371 Revert b780da4

SyedSaifuddin045:


stevecotton:
On the one hand, it's good to reformat the source code (and this does exactly match our clang-format style).

On the other hand, the reformatting changes `1 file changed, 6 insertions(+), 21 deletions(-)` into `1 file changed, 332 insertions(+), 370 deletions(-)`.

CelticMinstrel:
I wouldn't be surprised if he just had the IDE set to format on save and didn't realize, or something.

CelticMinstrel:
Also, maybe there's an error in the clang-format, because I thought our style guide was to not indent within namespaces? Although maybe anonymous namespaces are an exception…

SyedSaifuddin045:
Ohh Really sorry,Yes I used the Linter for formatting the code.I will make sure it's off.

CelticMinstrel:
If you could redo this without the linter that would be helpful.

Note that we don't have an issue with reformatting files (as long as it's to follow the style guide we use; there are probably a few clang-format changes required if using that). In future, if you want to use the linter, you can check in the reformatting in a separate commit before doing any actual work. The reason is that it makes it easier for us to review the change you made.

SyedSaifuddin045:
I force pushed please check, unfortunately I couldn't get the previous formatting of the document which would have helped in easy review.I just used the .clang-format file which was available in the src folder.

CelticMinstrel:
> Now I guess it's done properly it shows 16 deletion and 0 insertions. Sorry for the trouble I caused it is my first PR.

It's entirely possible that what you're looking at is done properly, but if that's the case, then you're not looking at the same thing I'm looking at here in this PR.

SyedSaifuddin045:
> > Now I guess it's done properly it shows 16 deletion and 0 insertions. Sorry for the trouble I caused it is my first PR.
> 
> It's entirely possible that what you're looking at is done properly, but if that's the case, then you're not looking at the same thing I'm looking at here in this PR.

Take a look at my screenshot, It shows me just 16 deletions and 0 insertions.
I wonder what went wrong now?
Screenshot (36)

CelticMinstrel:
Okay, now I'm looking at the same thing as you.

You deleted two blank lines in the comment at the top of the file. Could you put those back? Other than that I think it looks good now.

SyedSaifuddin045:
> Okay, now I'm looking at the same thing as you.
> 
> You deleted two blank lines in the comment at the top of the file. Could you put those back? Other than that I think it looks good now.

My bad,fixed it now please take a look.

stevecotton:
Looks good. Squash-merge once @soliton- reviews it?

src/gettext.cpp
```diff
@@ -469,7 +464,6 @@ std::string dsngettext (const char * domainname, const char *singular, const cha
 	std::string msgval;
 	//Braces so that only the line that needs to be in the lock is in it:
 	{
-		std::scoped_lock lock(get_mutex());
 		msgval = bl::dngettext(domainname, singular, plural, n, get_manager().get_locale());
 	}
```
soliton-, line 468: This code should go back to `std::string msgval = bl::dngettext(domainname, singular, plural, n, get_manager().get_locale());`

SyedSaifuddin045: Done Sir.🫡

src/gettext.cpp
```diff
@@ -466,12 +461,8 @@ inline const char* is_unlocalized_string2(const std::string& str, const char* si
...
 std::string dsngettext (const char * domainname, const char *singular, const char *plural, int n)
 {
-	std::string msgval;
-	//Braces so that only the line that needs to be in the lock is in it:
-	{
-		std::scoped_lock lock(get_mutex());
-		msgval = bl::dngettext(domainname, singular, plural, n, get_manager().get_locale());
-	}
+	std::string msgval = bl::dngettext(domainname, singular, plural, n, get_manager().get_locale());
+
```
Pentarctagon, line 465: One of the things that the CI checks is that there are no lines containing only whitespace or has trailing whitespace, so you need to delete the tab here.

soliton-:
Thanks for your contribution!

SyedSaifuddin045:
Thanks to all of you for providing help and guiding me through this.I will never forget my first PR.😄

PR #7430, opened 2023-03-03
campaignd: Fix blank license terms if [server_info] is missing

irydacea:
If interacting with a campaignd instance that has a fresh configuration or is otherwise missing the `[server_info]` tag, the client will present the player with completely blank licensing terms. This was caused by a logic issue with the processing of `[server_info]` wherein no default attribute values would ever be loaded if the tag was missing.

(This has never been relevant or found on the production campaignd instances because all of them have `[server_info]` tags explicitly provided for the purpose of formatting add-on forum thread URLs.)

### DIFF
--- a/src/server/campaignd/server.cpp
+++ b/src/server/campaignd/server.cpp
@@ -2051,8 +2051,8 @@ int main(int argc, char** argv)
 	} catch(const boost::program_options::error& e) {
 		PLAIN_LOG << "Error in command line: " << e.what();
 		return 10;
-	} catch(const config::error& /*e*/) {
-		PLAIN_LOG << "Could not parse config file";
+	} catch(const config::error& e) {
+		PLAIN_LOG << "Could not parse config file: " << e.message;
 		return 1;
 	} catch(const filesystem::io_exception& e) {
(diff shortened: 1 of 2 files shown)


src/server/campaignd/server.cpp
```diff
@@ -347,13 +347,17 @@ void server::load_config()
...
 	update_pack_lifespan_ = cfg_["update_pack_lifespan"].to_time_t(30 * 24 * 60 * 60);
 
-	if(const auto& svinfo_cfg = server_info()) {
-		server_id_ = svinfo_cfg["id"].str();
-		feedback_url_format_ = svinfo_cfg["feedback_url_format"].str();
-		web_url_ = svinfo_cfg["web_url"].str(default_web_url);
-		license_notice_ = svinfo_cfg["license_notice"].str(default_license_notice);
+	config svinfo_cfg;
+	// If the config file isn't empty, copy the server info tag instead of using all defaults
+	if(const auto& cfg = server_info()) {
```
soliton-, line 352: Could also change `server_info` to use `child_or_empty` and assign unconditionally.

irydacea: Just pushed a version that does just this

PR #7432, opened 2023-03-04
manual_tests: change macro loader to TEST

### LINKED ISSUE #7426: [Tests] `scenarios/manual_tests/scenario-test.cfg` unconsciously loads code into core
and triggers false-positives in log.

```ini
previously defined at campaigns/Under_the_Burning_Suns/utils/abilities.cfg:595
    included from test/scenarios/manual_tests/scenario-test.cfg:10
    included from _main.cfg:54
20230228 12:59:49 warning preprocessor: Redefining macro ABILITY_DISENGAGE without explicit #undef at ~add-ons/War_of_Legends/macros/abilities-new.cfg:179
    included from ~add-ons/War_of_Legends/_main.cfg:25
previously defined at campaigns/Under_the_Burning_Suns/utils/abilities.cfg:621
    included from test/scenarios/manual_tests/scenario-test.cfg:10
    included from _main.cfg:54
20230228 12:59:49 warning preprocessor: Redefining macro ABILITY_DISTRACT without explicit #undef at ~add-ons/War_of_Legends/macros/abilities.cfg:164
    included from ~add-ons/War_of_Legends/_main.cfg:25
previously defined at campaigns/Under_the_Burning_Suns/utils/abilities.cfg:643
    included from test/scenarios/manual_tests/scenario-test.cfg:10
    included from _main.cfg:54
20230228 12:59:49 warning preprocessor: Redefining macro WEAPON_SPECIAL_DAZE without explicit #undef at ~add-ons/War_of_Legends/macros/nightmare-utils.cfg:96
    included from ~add-ons/War_of_Legends/_main.cfg:25
previously defined at campaigns/Under_the_Burning_Suns/utils/abilities.cfg:631
    included from test/scenarios/manual_tests/scenario-test.cfg:10
    included from _main.cfg:54
20230228 12:59:49 warning preprocessor: Redefining macro WEAPON_SPECIAL_DAZE_EVENTS without explicit #undef at ~add-ons/War_of_Legends/macros/nightmare-utils.cfg:104
    included from ~add-ons/War_of_Legends/_main.cfg:25
previously defined at campaigns/Under_the_Burning_Suns/utils/abilities.cfg:686
    included from test/scenarios/manual_tests/scenario-test.cfg:10
    included from _main.cfg:54
```

These are all inside SP-campaigns and should NOT be loading them in MP.

knyghtmare:
Closes #7426

### DIFF
--- a/data/_main.cfg
+++ b/data/_main.cfg
@@ -49,7 +49,7 @@
 #endif
 [/ais]
 
-#ifdef DEBUG_MODE
+#ifdef TEST
 # load tests that need manual user input to run and validate
 {test/scenarios/manual_tests}
 #endif
--- a/data/_main.cfg
+++ b/data/_main.cfg
@@ -55,11 +55,15 @@
 [/textdomain]
 #endif
 
+#ifdef DEV_TEST
+# Please stop assaulting my log files when I load my addons
+# load tests that need manual user input to run and validate
+{test/scenarios/manual_tests}
+#endif
+
 #ifdef TEST
 {ai/scenarios/}
 {ai/micro_ais/scenarios/}
-# load tests that need manual user input to run and validate
-{test/scenarios/manual_tests}
 #define DONT_RELOAD_CORE
 #enddef
 
--- a/data/_main.cfg
+++ b/data/_main.cfg
@@ -49,11 +49,6 @@
 #endif
 [/ais]
 
-#ifdef TEST
-# load tests that need manual user input to run and validate
-{test/scenarios/manual_tests}
-#endif
-
 #ifdef EDITOR
 [textdomain]
     name="wesnoth-editor"
@@ -63,6 +58,8 @@
 #ifdef TEST
 {ai/scenarios/}
 {ai/micro_ais/scenarios/}
+# load tests that need manual user input to run and validate
+{test/scenarios/manual_tests}
 #define DONT_RELOAD_CORE
 #enddef
 


Pentarctagon:
Looks like there's a WML validation issue in the code behind that #ifdef. I assume [resource] isn't valid to be inside [units] since it's a top level tag?

knyghtmare:
> I assume [resource] isn't valid to be inside [units] since it's a top level tag?

`[units]` is a top-level tag. `[resource]` is normally defined inside `#ifdef` of content. Perhaps that is a reason why

CelticMinstrel:
Isn't that block of code completely redundant? Aren't those tests loaded further down in the other TEST block?

Pentarctagon:
No.

CelticMinstrel:
Okay, it would appear you're correct, but in that case… wouldn't it make sense to move this to data/test/_main.cfg?

Pentarctagon:
Probably this is another result of me not understanding why we have two entirely different sets of tests that can only be run manually.

CelticMinstrel:
What are you talking about? There's only one set of tests that can be run manually, isn't there? There's also the AI demo scenarios, which are also defined as test scenarios, but I'd consider them demos rather than tests. (But they can still somewhat function as manual test scenarios as well.)

Pentarctagon:
The `-t` tests vs the manual `-u` tests such as test_max_menu_items.

stevecotton:
All the `-t` tests can be run manually with `-u` too, although victory/defeat isn't how those tests report results, so a dialog saying "You have been defeated" doesn't necessarily mean the test failed. Running them with `-u` helps with debugging why a test is failing.

The difference between them is that
* the `-t` ones can automatically do all the actions and then work out whether the results match the expectations
* some `-u` tests need someone to manually look at them and say whether they passed or failed
* other `-u` tests are really workbenches for developers to do some action and then work out if the result matched their expectations. They need someone to manually do some actions.

Pentarctagon:
> the -t ones can automatically do all the actions and then work out whether the results match the expectations

What? Even the default test scenario (just `wesnoth -t`) doesn't work like that - it's a big scenario where you can manually trigger all sorts of actions.

stevecotton:
Oops, I've got and swapped round `-t` and `-u` in my comment. I'll fix it.

Pentarctagon:
How is `-t` different from `-u --show-gui`?

CelticMinstrel:
> The `-t` tests vs the manual `-u` tests such as test_max_menu_items.

Um, "manual `-u`" tests are not a thing. Any test can technically be run with either `-u` or `-t`, but you're not supposed to run manual tests with `-u`.

> How is `-t` different from `-u --showgui`?

`-t` doesn't return a result code that specifies a pass or fail. `-u` does, even with `--showgui`. The manual test scenarios are not configured to have a result in the form of victory or defeat – they're set up so that you can do things and view the outcome. Therefore, calling them with `-u` would just make the game hang, and calling them with `-u --showgui` also isn't useful because the returned result will be meaningless.

Pentarctagon:
`test_max_menu_items` and `test_relative_dir` were in the same place as the rest of the automated WML unit tests thus some of the confusion, though admittedly that's no longer the case. Would it be correct to say that:
* Everything in `data/test/scenarios/manual_tests` is intended to be run with `-t`
* All other tests in sub-folders under `data/test/scenarios` are primarily meant to be run with `-u` (for the automated WML unit tests)

?

stevecotton:
> `test_max_menu_items` and `test_relative_dir` were in the same place as the rest of the automated WML unit tests thus some of the confusion, though admittedly that's no longer the case.

I'd look at it the other way round - scenario-leaders.cfg, scenario-movethrough.cfg, scenario-story.cfg and scenario-test.cfg should all have been created in data/test, (or moved there when data/test was created, whichever came afterwards).

>     * Everything in `data/test/scenarios/manual_tests` is intended to be run with `-t`
>     * All other tests in sub-folders under `data/test/scenarios` are primarily meant to be run with `-u` (for the automated WML unit tests)

I thought that was the design of your reorg, and I agree with it. I haven't double-checked if it's currently true though, there might be a stray manual test somewhere.

CelticMinstrel:
I do think leaving the AI demo scenarios where they are is fine, though. They are meant to be run with `-t`, but they're also slightly more "user-facing" than all the other test scenarios.

data/_main.cfg
```diff
@@ -49,7 +49,7 @@
 #endif
 [/ais]
 
-#ifdef DEBUG_MODE
+#ifdef TEST
```
Pentarctagon, line 52: The contents of this #ifdef can be added to the already existing one starting on line 63.

CelticMinstrel: Or moved into data/test/_main.cfg

Pentarctagon:
So for the WML schema error, it looks like the [resource] defined at the end of `Under_the_Burning_Suns/utils/abilities.cfg` needs to be moved to a separate file?

CelticMinstrel:
I guess it could go in _main.cfg?

stevecotton:
I think we should just take Kaleh out of scenario-test.cfg. He was added in 34d48bd251d7d98ee0579dd98380705bceb596d7, which seems to just be a way to test his AMLAs without playing the campaign itself. Why not just test in an UtBS scenario?

I'm against changing UtBS' code for this - if the scenario-test wants Quenoth Elves, then it should have the code to cope with them. If scenario-test doesn't load that resource then it means you can create elves with debug mode, but their abilities won't work.

stevecotton:
> moved

I meant "completely removed" :D

CelticMinstrel:
I'm saying that the test scenario is including the abilities.cfg file inside `[units]` but should be including it at toplevel.

CelticMinstrel:
Doesn't the latest change mean that `-t` ceases to function entirely?

Why not just address the actual issue, that the test scenario includes abilities.cfg inside `[units]`?

PR #7433, opened 2023-03-04
[achievements] add macros for the icons

### LINKED ISSUE #7415: [Achievements] Macro defines for layering Achievement Icons
## Context:

This:
```ini
icon="attacks/blank-attack.png~SCALE(72,72)~BLIT(units/drakes/blademaster.png~RC(magenta>red), 0, 0)~BLIT(misc/achievement-frame-mp.png, 0, 0)"
```
Looks very packed, congested and hard to read for some.

So, what I will do is:
```ini
{ACHIEVEMENT_ICON_PLUS_BASE  (units/drakes/blademaster.png~RC(magenta>red)) (misc/achievement-frame-mp.png)}
{ACHIEVEMENT_ICON_NO_BASE  (units/drakes/blademaster.png~RC(magenta>red)) (misc/achievement-frame-mp.png)}
```

and also a similar one for icons made from images of the attacks/icons folder. Special case since they no require a "background" base.

knyghtmare:
Closes #7415

### DIFF
--- a/data/achievements.cfg
+++ b/data/achievements.cfg
@@ -1,13 +1,15 @@
 #textdomain wesnoth
 
 # macros for adding achievement icons
-#define ACHIEVEMENT_ICON_WITH_BASE BASE_IMG ICON_IMG FRAME_IMG
-data/core/images/{BASE_IMG}~SCALE(72,72)~BLIT({ICON_IMG}, 0, 0)~BLIT({FRAME_IMG}, 0, 0)
-#enddef
+#define ACHIEVEMENT_ICON_WBASE BASE_IMG ICON_IMG FRAME_IMG
+#arg FRAME_IMG
+misc/achievement-frame-mp.png#endarg
+{BASE_IMG}~SCALE(72,72)~BLIT({ICON_IMG}, 0, 0)~BLIT({FRAME_IMG}, 0, 0)#enddef
 
-#define ACHIEVEMENT_ICON_NO_BASE ICON_IMG FRAME_IMG
-data/core/images/{ICON_IMG}~BLIT({FRAME_IMG}, 0, 0)
-#enddef
+#define ACHIEVEMENT_ICON ICON_IMG FRAME_IMG
+#arg FRAME_IMG
+misc/achievement-frame-mp.png#endarg
+{ICON_IMG}~BLIT({FRAME_IMG}, 0, 0)#enddef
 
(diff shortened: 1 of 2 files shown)


soliton-:
Why not use the macro in mainline?

Would seem to make sense to at least make `FRAME_IMG` optional with `misc/achievement-frame-mp.png` as the default.

Hard coding `data/core/images/` seems dubious unless the base would be completely hard coded to blank-attack.png.

The names are pretty unwieldy. I would name one of those macros just `ACHIEVEMENT_ICON`. Maybe even remove the `ACHIEVEMENT_` prefix since this is only in the scope of achievements anyway.

data/achievements.cfg
```diff
@@ -1,5 +1,14 @@
 #textdomain wesnoth
 
+# macros for adding achievement icons
+#define ACHIEVEMENT_ICON_WITH_BASE BASE_IMG ICON_IMG FRAME_IMG
+data/core/images/{BASE_IMG}~SCALE(72,72)~BLIT({ICON_IMG}, 0, 0)~BLIT({FRAME_IMG}, 0, 0)
+#enddef
```
Pentarctagon, line 6: The #enddef here and below needs to be on the same line as the contents of the macro.

CelticMinstrel:
I'm also thinking this would ideally be in a separate file (which is then transcluded into this file). That allows add-ons to also use them.

Maybe make a file called achievement-macros.cfg at toplevel?

data/achievements.cfg
```diff
@@ -1,5 +1,16 @@
 #textdomain wesnoth
 
+# macros for adding achievement icons
+#define ACHIEVEMENT_ICON_WBASE BASE_IMG ICON_IMG FRAME_IMG
+#arg FRAME_IMG
+misc/achievement-frame-mp.png#endarg
+{BASE_IMG}~SCALE(72,72)~BLIT({ICON_IMG}, 0, 0)~BLIT({FRAME_IMG}, 0, 0)#enddef
+
+#define ACHIEVEMENT_ICON ICON_IMG FRAME_IMG
+#arg FRAME_IMG
```
CelticMinstrel, line 10: If this is supposed to be an optional argument, it shouldn't be included in the `#define` line.

CelticMinstrel:
Shouldn't that new file be included into the existing achievements file?

data/achievements.cfg
CelticMinstrel: This is still not resolved, and it also affects both macros.

PR #7443, opened 2023-03-08
Add curl-based functionality to download files from a URL.

Pentarctagon:
This is a complement to #7416 where it would be used to replace opening the replay's download URL in a browser.

### DIFF
--- a/src/network_download_file.cpp
+++ b/src/network_download_file.cpp
@@ -1,0 +1,75 @@
+/*
+	Copyright (C) 2003 - 2022
+	Part of the Battle for Wesnoth Project https://www.wesnoth.org/
+
+	This program is free software; you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation; either version 2 of the License, or
+	(at your option) any later version.
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY.
+
+	See the COPYING file for more details.
+*/
+
+#include "network_download_file.hpp"
+
+#include "filesystem.hpp"
+#include "log.hpp"
+
+#include <curl/curl.h>
+
+static lg::log_domain log_network("network");
+#define ERR_NW LOG_STREAM(err, log_network)
+#define DBG_NW LOG_STREAM(debug, log_network)
+
+namespace network
+{
+    static size_t write_callback(char* contents, size_t size, size_t nmemb, void* buffer)
+    {
+        size_t amount = size * nmemb;
+        static_cast<std::string*>(buffer)->append(contents, amount);
+        DBG_NW << "Downloaded " << amount << " bytes.";
+        return amount;
+    }
+
+    void download(const std::string& url, const std::string& local_path)
+    {
+        CURL* curl = curl_easy_init();
+        std::string buffer;
+        // curl doesn't initialize the error buffer until version 7.60.0, which isn't currently available on all supported macOS versions
+        char error[CURL_ERROR_SIZE];
+        std::fill_n(error, CURL_ERROR_SIZE-1, ' ');
+        error[CURL_ERROR_SIZE-1] = '\0';
+
+        if(curl) {
+            curl_easy_setopt(curl, CURLOPT_URL, url.c_str());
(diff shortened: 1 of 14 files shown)


CelticMinstrel:
Hmm… seems fine, but I think I'd prefer to change all instances of "curl" to "http" (excepting actual Curl API stuff obviously). That means the namespace, the file, and the logdomain, maybe one or two other things.

Pentarctagon:
It's not really a set of general http functionality though?

CelticMinstrel:
It's not _generic_ HTTP functionality, but that is basically what it's doing, isn't it?

I'm not specifically attached to HTTP if you have something better; I would just prefer not encoding the library used into the names of functions that theoretically could later be rewritten to use something else.

Also, I have a feeling making it required is going to be really annoying for me, as the Mac compile stuff will need updating. Could it perhaps be made optional so I can at least continue to build while waiting for that?

Pentarctagon:
I don't like how broad just "http" is, is the thing. Calling it "curl" may tie it to that specific library, but the name doesn't suggest a broader set of functionality than actually exists.

I don't plan to merge this until the MacCompileStuff is updated anyway, so I don't think it needs to be made optional.

CelticMinstrel:
I'm not quite sure why HTTP implies such a broader set of functionality than what already exists (this is literally "using HTTP to download a file) and why "curl" would fail to imply the same broader set of functionality (handling HTTP is literally what curl is for, so shouldn't both have the same implication), but there are surely other options if you dislike it that much.

For example you could just call the logdomain "download" and drop any references to curl outside of the actual API usage.

Pentarctagon:
Changed it to `network`.

src/curl_https_download_file.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <curl/curl.h>
+
+static lg::log_domain log_network("network");
+#define ERR_NW LOG_STREAM(err, log_network)
+#define DBG_NW LOG_STREAM(debug, log_network)
+
+namespace network
+{
+    static size_t write_callback(char* contents, size_t size, size_t nmemb, void* buffer)
```
Vultraz, line 28: Does this need  to be `static`?

Pentarctagon: Looks like it doesn't.

Pentarctagon: Actually, apparently macOS requires it to be static.

CelticMinstrel:
I said this in IRC too but I think the problem is in the Xcode file, not the environment. It looks like libcurl has not been added to a target definition.

Pentarctagon:
I'll be merging this in a couple days barring any other comments, since the CI is passing now.

src/network_download_file.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+            CURLcode res = curl_easy_perform(curl);
+
+            if(res != CURLE_OK) {
+                ERR_NW << "Error downloading file from url `" << url << "`.\n"
+                    << "Short error: " << curl_easy_strerror(res) << "\n"
+                    << "Long error: " << std::string(error);
+            } else {
+                try {
+                    filesystem::write_file(local_path, buffer);
```
soliton-, line 64: Does this overwrite files without asking?

Pentarctagon: Yes.

src/network_download_file.cpp
```diff
@@ -0,0 +1,75 @@
...
+    void download(const std::string& url, const std::string& local_path)
+    {
+        CURL* curl = curl_easy_init();
+        std::string buffer;
+        // curl doesn't initialize the error buffer until version 7.60.0, which isn't currently available on all supported macOS versions
+        char error[CURL_ERROR_SIZE];
+        std::fill_n(error, CURL_ERROR_SIZE-1, ' ');
+        error[CURL_ERROR_SIZE-1] = '\0';
+
+        if(curl) {
```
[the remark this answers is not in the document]
Pentarctagon: Added in ccb7001

src/network_download_file.cpp
Pentarctagon: Resolved in d280485

Issue #6952, opened 2022-08-07
wesnoth 1.17 crashes in Credits


grz0:
### Game and System Information

The Battle for Wesnoth version 1.17.6+dev (54fedb4d6a8-Clean) x86_64
Running on Fedora Linux 36 (Workstation Edition) x86_64
Distribution channel: Default

reproduced also on latest nightly flatpak

### Description of the bug

~~~
20220807 16:03:21 error draw/manager: Attempted to invalidate region [0,0|1920,1017] during draw
terminate called after throwing an instance of 'game::error'
  what():  invalidate during draw
Aborted (core dumped)
~~~

~~~
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0)
    at pthread_kill.c:44
#1  0x00007f0504e3ccb3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007f0504dec9c6 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f0504dd67f4 in __GI_abort () at abort.c:79
#4  0x00007f0505152b57 in __gnu_cxx::__verbose_terminate_handler() ()
    at ../../../../libstdc++-v3/libsupc++/vterminate.cc:95
#5  0x00007f050515e43c in __cxxabiv1::__terminate(void (*)()) (handler=<optimized out>)
    at ../../../../libstdc++-v3/libsupc++/eh_terminate.cc:48
#6  0x00007f050515d4a9 in __cxa_call_terminate(_Unwind_Exception*) (ue_header=0x56248b7f9a50)
    at ../../../../libstdc++-v3/libsupc++/eh_call.cc:54
#7  0x00007f050515dbc6 in __cxxabiv1::__gxx_personality_v0(int, _Unwind_Action, _Unwind_Exception_Class, _Unwind_Exception*, _Unwind_Context*)
    (version=<optimized out>, actions=6, exception_class=5138137972254386944, ue_header=<optimized out>, context=0x7ffe883850c0) at ../../../../libstdc++-v3/libsupc++/eh_personality.cc:688
#8  0x00007f0504fc6c74 in _Unwind_RaiseException_Phase2
    (exc=0x56248b7f9a50, context=0x7ffe883850c0, frames_p=0x7ffe883851b0) at ../../../libgcc/unwind.inc:64
[...]

PR #7445, opened 2023-03-10
Reimplement the end_credits with a [listbox]

stevecotton:
The UX here is terrible, but at least it doesn't crash. My intention here is to provide a stepping stone which I hope someone familiar with the GUI2 widgets will take and then improve the UX to solve the bug which this is just a part of, #6952.

UX issues include:
* It sets the background of an item to blue.
* Mousing over the list can make the background shadowed.
* Pressing "up" at the start jumps to the bottom of the list.

I'd be happy if someone takes this and reworks it. No need to keep this as a separate commit, cherry-pick bits from it as you wish. Alternatively, maybe there are already better widgets to use, or it could be merged with the expectation that it'll be improved before 1.18 releases.

### DIFF
--- a/data/gui/window/end_credits.cfg
+++ b/data/gui/window/end_credits.cfg
@@ -3,6 +3,93 @@
 ### Definition of the credits screen
 ###
 
+#define GUI_CREDITS_LIST_DEFINITION
+	# Contents of the [listbox], split with a macro to reduce the levels of indentation
+	[list_definition]
+
+		[row]
+
+			[column]
+				vertical_grow = true
+				horizontal_grow = true
+
+				[toggle_panel]
+					definition = "default"
+					return_value_id = "ok"
+					[grid]
+
+						[row]
+
+							[column]
+								vertical_grow = true
+								horizontal_alignment = "center"
+
+								grow_factor = 1
+								border = "all"
+								border_size = 5
+
+								[label]
+									id = "group_header"
+									definition = "default_huge"
+									text_alignment = "center"
+									use_markup = true
+								[/label]
+
+							[/column]
+
+						[/row]
+
+						[row]
+
+							[column]
+								vertical_grow = true
+								horizontal_alignment = "center"
+
+								grow_factor = 1
+								border = "all"
+								border_size = 5
+
+								[label]
+									id = "section_title"
+									definition = "default_large"
+									text_alignment = "center"
+									use_markup = true
+								[/label]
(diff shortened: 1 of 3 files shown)


Pentarctagon:
Could you post a screenshot of it?

stevecotton:
> does a listbox allow you to set the maximum selections to 0 so that items can't be selected at all?

It appears not - there's a `has_minimum_` and `has_maximum_`, but they're booleans that work on the "zero, one, infinity" rule. Having exactly one selected is always supported, the booleans allow or disallow "zero" and "two or more" respectively.

> Could you post a screenshot of it?

Immediately after pressing the credits button:
opening

Opening the credits and pressing "up" jumps to the bottom of the list, as well as increasing the scrolling speed (once you scroll back up). These are for "Eastern Invasion, Revised":
after_pressing_up

Hovering the mouse highlights in blue:
hovering

CelticMinstrel:
Definitely seems glitchy, but glitch is also better than crashing…

Pentarctagon:
The CI error is that boost test timeout issue, which was fixed in master.

stevecotton:
(Mainly to @Vultraz) what do we want to do with this PR? I could rebase it to include the CI workaround that Pentarctagon mentioned, but the main question is whether someone else is going to provide a better fix to the credits crash itself.

CelticMinstrel:
Personally I think this should be kept as a last resort, as I'm pretty sure the gamestate inspector crash is the same underlying cause.

Pentarctagon:
The gamestate inspector crash could be fixed by dividing the limit it current has before it paginates by the pixel scale multiplier, I would think?

CelticMinstrel:
If that's the case and someone tries it and it works, then great?

Pentarctagon:
The credits aren't currently paginated is the problem. I think paginating the credits by section/add-on also makes a lot of sense though.

stevecotton:
> The CI error is that boost test timeout issue, which was fixed in master.

Thanks, rebased.

gfgtdf:
How about using a Treeview instead? A treeview can be used as  general purpose "list generator" that, doesn't require the elements to be clickable (or to be of the same type), (for this reason most of the listings for example in WC2 are actually using treeviews)

CelticMinstrel:
A treeview obviously works for this purpose, though it feels kind of overengineered to use here. But sure, it is another option.

Issue #5374, opened 2020-12-19
NR S9, Eryssa is created with a loyal marker instead of a silver crown


Konrad22:
### Game and System Information
 - 1.15.7
 - Steam
 - Windows 10

### Describe the bug
Literally the title.

### To Reproduce

1. Skip to S9.
2. Free Eryssa.
3. Observe that she only has a loyal marker (but her hero death is always with 'lose the scenario' associated, which is also reflected in the scenario objectives).

### Expected behavior
She get's a silver crown so you can just look at her and go 'yeah, this one has to survive'.
(Which is especially necessary for players who know older versions NR, when she was allowed to die.)

Wedge009:
I don't know too much about WML, but at a guess, it's because she's introduced in S8 and it's a bit of an unusual situation (compared with other campaigns) that rescuing her is not a compulsory object. Moving so S9, she's added to the recall list as-is - perhaps that is the point at which she needs to be 'upgraded' to 'hero' status to be consistent with subsequent scenarios making her death an instant loss condition?

Is it as simple as assigning her `TRAIT_LOYAL_HERO` instead of `TRAIT_LOYAL`? Not sure if @sevu is still around and working on this...

Issue #7448, opened 2023-03-11
Nr: Some Problems with Northern Rebirth


TheShadowOfHassen:
So I was playing Northern rebirth and I have found a few bugs and have a few suggestions.

First I'm assuming you know this but the art for the portraits are drastically different from the rest of the game. They're not bad but the art style is different.

Another thing is it may be in other places but In scenario 10 Stolen Gold  Hamel refers the dwarves as men more than once.

The last problem is also mentioned in issue # 5374 If Eryssa dies after you rescue her you loose the game, I don't remember either the objectives saying that or there being a crown over her head.

Their not breaking issues or anything but they'd be nice to have fixes. Thanks for the great game.

Wedge009:
Thanks for the feedback but I'm not sure I see anything new here. Art development is an ongoing effort and you're right, the art for NR is a bit inconsistent but I think it's one of the oldest in terms of introduction.

There is some discussion of 'men' to refer to non-human people in #4880.

If #5374 is already logged, then that covers the situation with Eryssa, doesn't it?

TheShadowOfHassen:
#5374 does sort of log it except it's been assigned to someone two years ago and It wasn't done. I brought it up again because I wasn't sure if anyone was still doing it. It has not been a very fun experience to forget I have to keep her alive and when she dies I loose

Also I noticed looking through the campaign files that there is a hole bunch of dialog and stuff for a plot arc that needs edited out. You probably know that.

Another comment I have for the campaign is that the battles are so much bigger and it takes me days of playing to get through them. It gets kind of boring but that's just me.

I just wanted to bring these problems to attention. I'll close the issue when the issue for Eryssa is fixed. I could figure it out if everyone else is busy. I don't think it'd be too hard.

Wedge009:
Keep in mind this is a volunteer, community project and a lot of issues are quite old but by the same token eventually many issues to get resolved at some point. You're welcome to have a go if you'd like to see something improved.

I understand there's an ongoing effort to thoroughly revise a bunch of official campaigns and I think NR is one of them. @nemaara may update you on that.

I believe NR is supposed to be at the harder end of the spectrum so larger battles are part of that. At the same time, a large battle that doesn't really have anything going for it beyond its scale is probably a candidate for improvement - I know there are plenty of other contributors far more critical in this aspect than I am who are actively working on game-play improvements.

knyghtmare:
[...]

PR #7449, opened 2023-03-12
NR: Give Eryssa hero trait since her death is a loss condition

Wedge009:
I never got an answer to my question in #5374, but this change seems to do what was requested.


knyghtmare:
Eryssa is actually an optional hero so having the loyal icon was maybe intentional.

Wedge009:
Yes, but once she is in the party, she becomes a loss condition on death, doesn't she? Personally I don't pay too much attention to the 'crowns' but there's requests for this in #5374 and #7448.

cooljeanius (approved):

nemaara (approved):

Issue #7447, opened 2023-03-11
Quenoth Fighter unit line multi-attack implementation needs improvement


abepressman:
### Game and System Information

- **Version:** 1.17.13
- **Downloaded from:** Steam
- **Build info:**
- **OS:** Mac OS

### Description of the bug

For some reason, the Quenoth Fighter line of units appear to have gained extra attacks that don't do anything. (I don't know what was _intended_, but at least on the version I'm running, they merely have copies of their attacks that deal less damage)
<img width="500" alt="Screen Shot 2023-03-11 at 6 49 41 AM" src="https://user-images.githubusercontent.com/26127550/224482758-16fd069e-f832-4344-98ea-0ac3bc2c5138.png">
<img width="534" alt="Screen Shot 2023-03-11 at 6 49 36 AM" src="https://user-images.githubusercontent.com/26127550/224482759-f9c6e47c-c41b-4670-9439-8576423abb00.png">

knyghtmare:
Apparently, they use different number of "attacks" for each. Meaning, that 9-1 there can hit more than one opponent or the same opponent twice.
However, I see a very bad UX here since nothing of this note has been shown to the player (you).
I explained this before..nobody listened and now you have an open issue here. LOL

Hejnewar:
Now Im waiting for bug report that states that Taurochs are useless and that Fighters are OP because both a bugged and it is really easily noticable after a while. Also the bug for UI and gameplay are already opened when it comes to attacks used.

abepressman:
> Apparently, they use different number of "attacks" for each. Meaning, that 9-1 there can hit more than one opponent or the same opponent twice.

SO, this is not at all intuitive or explained. If they make the 9-2 attack, their turn ends. If they make the 9-1 attack first, they can follow it up with the 9-2 attack.

> Actually, it's here but not many would take time to read a unit text:

The unit text doesn't explain at all how the ability actually works, and if you're a normal player who always uses the better attack first, you'll likely never figure it out.

Wedge009:
Again, the heading was misleading. As best as I can see there are two issues:
* The reason for the seemingly duplicated attack entries is not clear at first - I take it that either the level 1 fighter can expend both of its attacks on one target, as conventional for most other main-line units, or only attack once and then possibly use its other attack on another target.
* This brings up the other problem in that the fighter can target one opponent with the single attack and then on the subsequent attack choose the double attack option.

abepressman:
[...]

Issue #7446, opened 2023-03-11
Tauroch units can only use each attack once per scenario instead of once per turn


abepressman:
### Game and System Information

- **Version:** 1.17.13
- **Downloaded from:** Steam
- **Build info:** ?
- **OS:** Mac OSX

### Description of the bug

Decided to try out Under the Burning Suns to see how the new units work. Unfortunately, once my taurochs attack, they can't attack anymore. At all. This is not how the ability "once per turn" is described.

abepressman:
Just updating to add: However the Quenoth fighter line was *supposed* to change, I don't think it quite worked right. The units now have extra attacks, but those attacks are completely useless.

Toranks:
![imagen](https://user-images.githubusercontent.com/40789364/224479711-64166b39-746a-4613-b302-2c09f56443ab.png)
I just checked all the variations of the Tauroch and they work fine. Aren't you using a campaign already started when the changes were made? That can mess things.

knyghtmare:
Works for me.

abepressman:
<img width="1439" alt="Screen Shot 2023-03-11 at 6 46 35 AM" src="https://user-images.githubusercontent.com/26127550/224482511-f80bf038-dfb5-4f48-abdf-15a851f1a238.png">

Screenshot attached. Maybe it's a bug related to mac, or steam, or something to do with how they spawn in the first level?

I just started playing a new game from the beginning, yeah. None of my taurochs can attack more than once, period.

The Quenoth Fighter line also appears to be broken with respect to its attacks. (At higher levels, it gains multiple basically identical attacks, that don't seem to be usable in any circumstance.)

knyghtmare:
WAIT

![image](https://user-images.githubusercontent.com/5283677/224482609-15f6a75f-c00c-4331-8a89-2b94e72a44d4.png)

abepressman:
<img width="500" alt="Screen Shot 2023-03-11 at 6 49 41 AM" src="https://user-images.githubusercontent.com/26127550/224482635-4bf0c8aa-76db-47d8-bb7e-edb9908a14d6.png">
<img width="534" alt="Screen Shot 2023-03-11 at 6 49 36 AM" src="https://user-images.githubusercontent.com/26127550/224482636-8e8fbbed-c22d-4dba-9604-27b378315f8c.png">

Also (and I could put it in a different thread if that's better), the changes to the Quenoth fighters aren't working as intended either.

[...]

Issue #7367, opened 2023-02-08
Movement_used and attacks_used keys should be represented in game


Hejnewar:
### Describe the desired feature

Currently two attack attributes in mainline exist without being represented anywhere. Movement_used and attacks_used keys. For better player experience as well as ease of use in mainline and UMC I believe that they should be shown somewhere within the UI.

CelticMinstrel:
Suggestion for attacks used: "(3)" after the attack name (if attacks left and max attacks are shown elsewhere).

For movement used, maybe "(2 MP)". With both non-default, "(3; 2 MP)".

We definitely need to show attacks left and max attacks somewhere too.

CelticMinstrel:
Ah, perfect. I missed that (it's probably only shown when actually relevant too).

Wedge009:
If it's already present, does that mean the issue can be closed?

On the other hand, how did you 'activate' the extra information? I added attacks via debug mode and didn't see any difference. Presumably it's the unit definition that needs to be changed.

Hejnewar:
> If it's already present, does that mean the issue can be closed?

No because the uses X attacks part is only in my mod and this was one of the two original problems. The attacks just show up if you put more than one max attack in a unit file but if the unit is reduced back to 1 attack they disappear so technically it should work with adding as well. 

 ``attacks = 2 ``

CelticMinstrel:
> If it's already present, does that mean the issue can be closed?

No, this issue is about information on a specific attack – the number of movement points consumed when you use the attack. (And the amount it deducts from attacks_left.)

> On the other hand, how did you 'activate' the extra information?

For attacks_left and max_attacks I assume it's "activated" by having a max_attacks greater than 1.

For attacks_used I would expect it to be shown under the same condition, I think, but it could maybe be omitted if it's either equal to 1 or equal to max_attacks (pick one, not both).

For movement_used I would expect it to be shown whenever it's less than max_moves.

PR #7450, opened 2023-03-12
UtBS - Update and Fixes

Hejnewar:
Fixing #7447 #7446 

As well as, various unit stat changes. 

Finally XP and Recall Cost changes, honestly I really like them, yes it will probably require more work but its really fun to play for me. It does make the game overall harder, maybe with the exception of Normal difficulty where recalling for leveling is now cheaper and L3 units still cost around 20g. 

Hard is now like previous Nightmare and I can roll thru it  just as easily and Nightmare is finally a challenge (thus far I managed to beat 4 scenarios and I will continue this run when I have time, no idea if I will win tho). 

The problem currently might be a bigger difference in difficulty between well difficulties but maybe thats also for the better.


data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1,5 +1,78 @@
...
+#define STR_USES_ATTACKS_RESTRICT_6
+_ "This attack uses 6 attacks"#enddef
+
+#define WEAPON_SPECIAL_USES_ATTACKS X
+    [disable]
+        id = uses_attacks
+        name = {STR_USES_ATTACKS_NAME_{X}}
+        description = {STR_USES_ATTACKS_RESTRICT_{X}}
+        [filter_self]
+            formula = "self.max_attacks < {X}"
```
CelticMinstrel, line 45: Shouldn't this be `self.attacks_left`?

data/campaigns/Under_the_Burning_Suns/utils/macros.cfg
```diff
@@ -134,3 +134,36 @@ if(red <= green and green <= blue + 3, ({NEWCHAN} / 255.0)^2 * 255.0, {ORIGCHAN}
...
     [/event]
 #enddef
+
+#define UTBS_XP_MODIFIER
+    experience_modifier={ON_DIFFICULTY 100 150 200}
+#enddef
+
+#define UTBS_CORRECT_RECALL_COST EVENT
+    [event]
+        name = {EVENT}
```
CelticMinstrel, line 144: It doesn't look like there's any reason for this to be a parameter? Every instance passes `start`.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
Hejnewar: Oh I cant filter it by ``self.attacks_left`` because then attack will stop before it even happens because attacks get consumed and attack is instantly disabled because condition for disabling is fullfilled thus using ``self.max_attacks`` is actually a working option.

CelticMinstrel: Your explanation is pretty incoherent, but let me see if I've understood it correctly.

If you use `attacks_left`, the game will allow you to use the attack, but then the attack will fail because the state of the filter changes between the time you choose the attack and the time the attack is executed.

Is that correct?

data/campaigns/Under_the_Burning_Suns/utils/macros.cfg
Hejnewar: Oversight

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
Hejnewar: I guess yes.
Like you try to attack -> attacks are consumed -> disable turns on because you no longer meet the required amount of attacks -> attack is disabled.

CelticMinstrel: Hmm… that seems really unfortunate… I'll see if I can think of a better solution but I guess this works for now.

stevecotton: Would this `[disable]` and the related strings be needed at all if the engine's handling of `attacks_left` was fixed?

If it needs to be fixed in the engine, it would be better to have the campaign testing the engine code all the time instead of adding a workaround in the campaign.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -236,18 +309,52 @@ _ "third verse"#enddef
...
-        [/modify_unit]
+            variable = opts_units
+        [/store_unit]
+        [foreach]
+            array = opts_units
+            variable = current_opts_unit
+            [do]
+                [foreach]
+                    array = opts_units
+                    variable = current_opts_unit
```
stevecotton, line 327: This shadows the outer loop`s `current_opts_unit` and it looks like the outer loop is redundant.

data/campaigns/Under_the_Burning_Suns/utils/macros.cfg
```diff
@@ -134,3 +134,36 @@ if(red <= green and green <= blue + 3, ({NEWCHAN} / 255.0)^2 * 255.0, {ORIGCHAN}
...
+            [do]
+                [store_unit_type]
+                    type = $this_item.type
+                    variable = utbs_unit_type
+                [/store_unit_type]
+                [modify_unit]
+                    [filter]
+                        find_in = this_item
+                    [/filter]
+                    recall_cost = "$($utbs_unit_type.cost*{ON_DIFFICULTY 0.33 0.66 1})"
```
stevecotton, line 163: Is this explained to the player?

There's an old warning in storytxt.cfg that's shown at the start of the campaign, but it's about the day/night cycle. The rules of the post-1.16 changes to UtBS don't seem to be explained anywhere.

nemaara (approved):
I'm both scared and excited, but let's try it out and see how it plays.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
CelticMinstrel: I think something should be done in the engine but I haven't worked out what that would look like. The most obvious option (deducting from `moves` and `attacks_left` only after the attack completes) is probably a significant compatibility issue and can't easily be considered. Other options I've considered are adding a formula variable `is_attacking` which is only true literally in the middle of an attack (so the filter becomes `is_attacking or attacks_left < attacks_used`) or adding a new key in the weapon to toggle between the two behaviours (something like `needs_exact_attacks=true` to select the UtBS-required behaviour).

Regarding the strings, those are presumably a way to deal with #7367, so I assume they wouldn't be needed if that was addressed.

stevecotton: Do we need it to affect anything other than the UI when the user is selecting an attack, and the list when the AI is selecting an attack?

Adding a new key/attribute and then just affecting those two points sounds simpler (without looking into the details) than anything which needs to handle special cases once the attack has already been chosen.

CelticMinstrel: I'm not quite sure what you're trying to say but it feels like you've missed a point somewhere in here.

The UI for selecting an attack is fine. The filter states that the attack is enabled, so it lets you pick it.

Then the attack system deducts `attacks_used` from the current `attacks_left`.

Now the filter states that the attack is disabled, so the attack system aborts the attack and nothing happens.

stevecotton: When is this `[disable]` needed? Could the entire `[disable]` be removed?

If the UI only lets the user pick attacks that have enough `attack_left`, then why does the attack need a `[disable]` at all?

CelticMinstrel: That does solve this specific use-case. I do think that using `[disable]` (or any ability) based on `attacks_left` or `moves` should ideally function rather than causing an abort of the attack, but that wouldn't be needed in this specific case if there were a key that specifies the attack cannot be executed if there are insufficient `attacks_left`. That's the third possible solution I noted in my earlier post (`needs_exact_attacks=true`).

stevecotton: Once the attack has been selected (and possibly after any `pre_attack` event has fired), I think any logic based on `attacks_left` or `moves` should see the value after the cost of attacking has been deducted. That would simplify scripting dialogue such as the unit saying "I can attack again", or even animation where the unit is breathing more heavily when they've already attacked.

However, for this PR my main point is that I'd prefer the mainline campaigns to use the engine's features rather than coding workarounds for them. Otherwise we'll release 1.18 and then find that part of the engine wasn't tested and all UMC ends up copying the workaround too.

data/campaigns/Under_the_Burning_Suns/utils/macros.cfg
Hejnewar: "It changes certain Wesnoth standards, such as the elves’ stats" 

Well it seems to be really old because it still refers to the Desert elves. 

I will add it there.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
CelticMinstrel: I think that makes sense, yeah, but in this one case it doesn't make sense, and I imagine you could come up with other cases where it doesn't make sense. Perhaps the second solution's better? Then the formula could be forced to be true while the attack is in progress, and while it's not, it's correctly based on `moves` or `attacks_left`.

CelticMinstrel: > I think that makes sense, yeah, but in this one case it doesn't make sense, and I imagine you could come up with other cases where it doesn't make sense. Perhaps the second solution's better? Then the formula could be forced to be true while the attack is in progress, and while it's not, it's correctly based on `moves` or `attacks_left`.

Another option is to not evaluate the filter at all in the middle of the attack process. I don't know if that's viable though. Maybe some things rely on the possibility of it changing to abort the attack.

Hejnewar:
Is there anything else to do or is it mergable?

stevecotton:
> Is there anything else to do or is it mergable?

Please rebase and check that the CI passes. (I haven't looked at updating other comments yet, but not breaking `git bisect` is my biggest concern).

data/campaigns/Under_the_Burning_Suns/utils/macros.cfg
stevecotton: Just saying "it changes certain Wesnoth standards, such as ... recall costs" doesn't tell the player anything that they can plan for, and without reading the WML the numbers' origins aren't going to be obvious.

Pentarctagon:
There are a couple unresolved review comments as well.

data/campaigns/Under_the_Burning_Suns/utils/macros.cfg
Hejnewar: Well it says thet the campaign changes stats of elves without any specifics (and players play with completely different faction anyway). Im not planning to update this myself any further but I can add a comment stating that it would be good to actually update it.

Hejnewar:
Since all checks pass Im going to merge this.

PR #7451, opened 2023-03-14
Fix -Wshorten-64-to-32

Pentarctagon:
---

I'm going to try and go through the rest of them as well (hence the draft PR status), but this change by itself reduces the number of warnings from 2638 to 1721.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -462,6 +462,19 @@ else()
 	RemoveFlag(CMAKE_C_FLAGS_MINSIZEREL CACHE "/DNDEBUG" "Default C flags for MinSizeRel")
 endif()
 
+if(ENABLE_MYSQL)
+	execute_process(COMMAND mysql_config --cflags OUTPUT_VARIABLE MYSQL_CFLAGS OUTPUT_STRIP_TRAILING_WHITESPACE)
+	string(REGEX REPLACE "-I" "" MYSQL_CFLAGS "${MYSQL_CFLAGS}")
+	string(REGEX REPLACE "-DNDEBUG" "" MYSQL_CFLAGS "${MYSQL_CFLAGS}")
+	execute_process(COMMAND mysql_config --libs OUTPUT_VARIABLE MYSQL_LIBS OUTPUT_STRIP_TRAILING_WHITESPACE)
+	add_subdirectory("src/modules/mariadbpp/")
+endif()
+
+# ignore this warning for mariadbpp submodule by adding it after adding the mariadbpp subdirectory
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+	add_compile_options(-Wshorten-64-to-32)
+endif()
+
 # When the path starts with a / on a Unix system it's an absolute path.
 # This means that on Windows the path used is always relative.
 if(IS_ABSOLUTE "${LOCALEDIR}")
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -584,7 +584,7 @@ battle_context battle_context::choose_defender_weapon(nonempty_unit_const_ptr at
 	VALIDATE(attacker_weapon < attacker->attacks().size(), _("An invalid attacker weapon got selected."));
 
 	const attack_type& att = attacker->attacks()[attacker_weapon];
-	auto no_weapon = [&]() { return battle_context(attacker, attacker_loc, attacker_weapon, defender, defender_loc, -1); };
+	auto no_weapon = [&]() { return battle_context(attacker, attacker_loc, static_cast<int>(attacker_weapon), defender, defender_loc, -1); };
 	std::vector<battle_context> choices;
 
 	// What options does defender have?
@@ -594,7 +594,7 @@ battle_context battle_context::choose_defender_weapon(nonempty_unit_const_ptr at
 			//no need to calculate the battle_context here.
 			continue;
 		}
-		battle_context bc(attacker, attacker_loc, attacker_weapon, defender, defender_loc, i);
+		battle_context bc(attacker, attacker_loc, static_cast<int>(attacker_weapon), defender, defender_loc, static_cast<int>(i));
 
 		if(bc.defender_stats_->disable) {
 			continue;
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -1666,7 +1666,7 @@ bool backstab_check(const map_location& attacker_loc,
 	}
 
 	// If sides aren't valid teams, then they are enemies.
-	if(std::size_t(defender->side() - 1) >= teams.size() || std::size_t(opp->side() - 1) >= teams.size()) {
+	if(static_cast<std::size_t>(defender->side() - 1) >= teams.size() || static_cast<std::size_t>(opp->side() - 1) >= teams.size()) {
 		return true;
 	}
 
--- a/src/actions/attack.cpp
+++ b/src/actions/attack.cpp
@@ -152,7 +152,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	firststrike = weapon->has_special_or_ability("firststrike");
 
 	{
-		const int distance = distance_between(u_loc, opp_loc);
+		const int distance = static_cast<int>(distance_between(u_loc, opp_loc));
 		const bool out_of_range = distance > weapon->max_range() || distance < weapon->min_range();
 		disable = weapon->has_special("disable") || out_of_range;
 	}
@@ -575,7 +575,7 @@ battle_context battle_context::choose_attacker_weapon(nonempty_unit_const_ptr at
 /** @todo FIXME: Hand previous defender unit in here. */
 battle_context battle_context::choose_defender_weapon(nonempty_unit_const_ptr attacker,
 		nonempty_unit_const_ptr defender,
-		unsigned attacker_weapon,
+		std::size_t attacker_weapon,
 		const map_location& attacker_loc,
 		const map_location& defender_loc,
 		const combatant* prev_def)
--- a/src/actions/undo.cpp
+++ b/src/actions/undo.cpp
@@ -360,7 +360,7 @@ void undo_list::undo()
 	undos_.pop_back();
 	if (undo_action* undoable_action = dynamic_cast<undo_action*>(action.get()))
 	{
-		int last_unit_id = resources::gameboard->unit_id_manager().get_save_id();
+		int last_unit_id = static_cast<int>(resources::gameboard->unit_id_manager().get_save_id());
 		if ( !undoable_action->undo(side_) ) {
 			return;
 		}
--- a/src/actions/undo.cpp
+++ b/src/actions/undo.cpp
@@ -360,11 +360,11 @@ void undo_list::undo()
 	undos_.pop_back();
 	if (undo_action* undoable_action = dynamic_cast<undo_action*>(action.get()))
 	{
-		int last_unit_id = static_cast<int>(resources::gameboard->unit_id_manager().get_save_id());
+		std::size_t last_unit_id = resources::gameboard->unit_id_manager().get_save_id();
 		if ( !undoable_action->undo(side_) ) {
 			return;
 		}
-		if(last_unit_id - undoable_action->unit_id_diff < 0) {
+		if(undoable_action->unit_id_diff > last_unit_id) {
 			ERR_NG << "Next unit id is below 0 after undoing";
 		}
 		resources::gameboard->unit_id_manager().set_save_id(last_unit_id - undoable_action->unit_id_diff);
--- a/src/actions/vision.cpp
+++ b/src/actions/vision.cpp
@@ -594,8 +594,8 @@ std::vector<int> get_sides_not_seeing(const unit & target)
 	const std::vector<team> & teams = resources::gameboard->teams();
 	std::vector<int> not_seeing;
 
-	std::size_t team_size = teams.size();
-	for ( std::size_t i = 0; i != team_size; ++i)
+	int team_size = static_cast<int>(teams.size());
+	for ( int i = 0; i != team_size; ++i)
 		if ( !target.is_visible_to_team(teams[i], false) )
 			// not_see contains side numbers; i is a team index, so add 1.
 			not_seeing.push_back(i+1);
--- a/src/ai/composite/property_handler.hpp
+++ b/src/ai/composite/property_handler.hpp
@@ -109,7 +109,7 @@ public:
 			handle_delete(with_same_id);
 		}
 
-			typename ptr_vector::iterator i = std::find_if(values_.begin(),values_.end(),path_element_matches<ptr>(child));
+		typename ptr_vector::iterator i = std::find_if(values_.begin(),values_.end(),path_element_matches<ptr>(child));
 		return do_add(i-values_.begin(),cfg);
 	}
 
@@ -144,7 +144,7 @@ protected:
 		factory_(vec, cfg);
 	}
 private:
-	bool do_add(int pos, const config &cfg)
+	bool do_add(long pos, const config &cfg)
 	{
 		if (pos<0) {
 			pos = values_.size();
--- a/src/ai/contexts.cpp
+++ b/src/ai/contexts.cpp
@@ -402,7 +402,7 @@ void readonly_context_impl::calculate_moves(const unit_map& units, std::map<map_
 			if(!enemy && resources::gameboard->map().is_village(dst)) {
 				for(std::size_t n = 0; n != resources::gameboard->teams().size(); ++n) {
 					if(resources::gameboard->teams()[n].owns_village(dst)) {
-						int side = n + 1;
+						int side = static_cast<int>(n) + 1;
 						if (get_side() != side && !current_team().is_enemy(side)) {
 							friend_owns = true;
 						}
@@ -971,7 +971,7 @@ const map_location& readonly_context_impl::nearest_keep(const map_location& loc)
 		if (avoided_locations.find(*i)!=avoided_locations.end()) {
 			continue;
 		}
-		const int distance = distance_between(*i,loc);
+		const int distance = static_cast<int>(distance_between(*i,loc));
 		if(res == nullptr || distance < closest) {
 			closest = distance;
 			res = &*i;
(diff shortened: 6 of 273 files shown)


stevecotton:
`std::ptrdiff_t` seems more appropriate than `long`, however I think the best thing would be to move it under the line than typdefs `Itor`:

```
 		typedef child_list::iterator Itor;
+		typedef Itor::difference_type difference_type;
```

src/ai/composite/property_handler.hpp
```diff
@@ -144,7 +144,7 @@ class vector_property_handler : public base_property_handler {
 		factory_(vec, cfg);
 	}
 private:
-	bool do_add(int pos, const config &cfg)
+	bool do_add(std::size_t pos, const config &cfg)
```
gfgtdf, line 147: This probably changes  behaviour  since the line below checks for `pos<0`

Pentarctagon: Yeah, it looks like the CI caught it as well. I'm a bit surprised that works like that, but I suppose it does make sense.

src/editor/map/map_context.hpp
```diff
@@ -107,7 +107,7 @@ class map_context : public display_context
 
 	void new_area(const std::set<map_location>& area) {
 		tod_manager_->add_time_area("", area, config());
-		active_area_ = tod_manager_->get_area_ids().size() -1;
+		active_area_ = static_cast<int>(tod_manager_->get_area_ids().size()) - 1;
```
Vultraz, line 110: Might be better to make `active_area_` size_t?

Pentarctagon: If `tod_manager_->get_area_ids().size()` returns 0, then `active_area_` would end up as -1.

Vultraz: Agh. Bad design but out of scope to fix here. >_<

src/server/common/simple_wml.hpp
```diff
@@ -91,7 +91,7 @@ class string_span
 	const char* begin() const { return str_; }
 	const char* end() const { return str_ + size_; }
 
-	int size() const { return size_; }
+	std::size_t size() const { return size_; }
```
gfgtdf, line 94: Did you check that this is safe?, when a code that calls these does something like 
```
int a = somefunction();
if (a < cfg.size()) {

}
```
Your patch changes behavior in a way that when somefunction() returned  -1 it was false before and not it might be true. So you probably have to check all codes that call size()

Pentarctagon: That'd be `-Wsign-compare`, which doesn't occur as a result of that change.

Pentarctagon:
There was #7103.

Pentarctagon:
I'll be going back through and double checking these commits over the next few days, but if there are no other comments by then I plan to go ahead and merge this.

CMakeLists.txt
```diff
@@ -462,6 +462,19 @@ else()
...
+if(ENABLE_MYSQL)
+	execute_process(COMMAND mysql_config --cflags OUTPUT_VARIABLE MYSQL_CFLAGS OUTPUT_STRIP_TRAILING_WHITESPACE)
+	string(REGEX REPLACE "-I" "" MYSQL_CFLAGS "${MYSQL_CFLAGS}")
+	string(REGEX REPLACE "-DNDEBUG" "" MYSQL_CFLAGS "${MYSQL_CFLAGS}")
+	execute_process(COMMAND mysql_config --libs OUTPUT_VARIABLE MYSQL_LIBS OUTPUT_STRIP_TRAILING_WHITESPACE)
+	add_subdirectory("src/modules/mariadbpp/")
+
+	# ignore this warning for mariadbpp submodule
+	if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+		add_compile_options(-Wshorten-64-to-32)
```
stevecotton, line 474: The comments says "ignore", should there be a `no-` somewhere in the added option?

src/actions/undo.cpp
```diff
@@ -360,7 +360,7 @@ void undo_list::undo()
 	undos_.pop_back();
 	if (undo_action* undoable_action = dynamic_cast<undo_action*>(action.get()))
 	{
-		int last_unit_id = resources::gameboard->unit_id_manager().get_save_id();
+		int last_unit_id = static_cast<int>(resources::gameboard->unit_id_manager().get_save_id());
```
stevecotton, line 363: If `get_save_id` and `set_save_id` use std::size_t, then why is this `int`?

Making it `std::size_t` would require fixing the sanity check comparing it to `undoable_action->unit_id_diff`, and that descends into a rabbit warren of the TODO comment that's on `undo_action::unit_id_diff`, however the cast to `int` doesn't look right.

src/actions/vision.cpp
```diff
@@ -594,8 +594,8 @@ std::vector<int> get_sides_not_seeing(const unit & target)
 	const std::vector<team> & teams = resources::gameboard->teams();
 	std::vector<int> not_seeing;
 
-	std::size_t team_size = teams.size();
-	for ( std::size_t i = 0; i != team_size; ++i)
+	int team_size = static_cast<int>(teams.size());
+	for ( int i = 0; i != team_size; ++i)
 		if ( !target.is_visible_to_team(teams[i], false) )
 			// not_see contains side numbers; i is a team index, so add 1.
 			not_seeing.push_back(i+1);
```
stevecotton, line 601: Not for this PR, but if we don't already have one, would a function for converting team numbers to side numbers and back be useful, instead of having obscure `+1` and `-1` (sometimes with a comment) around the code?

src/ai/contexts.cpp
```diff
@@ -1063,24 +1063,23 @@ double readonly_context_impl::power_projection(const map_location& loc, const mo
...
+					best_rating = static_cast<int>(rating);
 					best_unit = it->second;
 				}
 			}
 		}
 
 		if (!best_unit.valid()) continue;
 		map_location *pos = std::find(beg_used, end_used, best_unit);
-		int index = pos - beg_used;
+		long index = pos - beg_used;
```
stevecotton, line 1082: Using `long` seems incorrect. If this is looking for a `signed size_t`, why not `std::ptrdiff_t`, the appropriate `different_type` or just `auto index = std::distance(...)`?

src/ai/default/aspect_attacks.cpp
```diff
@@ -141,7 +141,7 @@ void aspect_attacks_base::do_attack_analysis(const map_location& loc,
 
 	ai::manager::get_singleton().raise_user_interact();
 	const int max_attack_depth = 5;
-	if(cur_analysis.movements.size() >= std::size_t(max_attack_depth)) {
+	if(cur_analysis.movements.size() >= static_cast<std::size_t>(max_attack_depth)) {
```
stevecotton, line 144: Fix the previous line instead, these two lines are the only place this constant is used.

src/display.hpp
```diff
@@ -110,7 +110,7 @@ class display : public gui2::top_level_drawable
 
 	/** The viewing team is the team currently viewing the game. */
 	std::size_t viewing_team() const { return currentTeam_; }
-	int viewing_side() const { return currentTeam_ + 1; }
+	int viewing_side() const { return static_cast<int>(currentTeam_) + 1; }
```
stevecotton, line 113: Seems like fixing the wrong problem, `viewing_team` and `viewing_side` should return the same type.

This one's in the third commit of the branch, so sorry for not catching it earlier.

src/gui/dialogs/unit_advance.hpp
```diff
@@ -25,7 +25,7 @@ class unit_advance : public modal_dialog
 public:
 	unit_advance(const std::vector<unit_const_ptr>& samples, std::size_t real);
 
-	int get_selected_index() const
+	std::size_t get_selected_index() const
```
stevecotton, line 28: The only caller of this casts it back to `int`, so this seems a bad change. Both the caller and the old method signature knew something that `std::size_t` doesn't match, so the question should be, is `selected_index_` the wrong type?

I think it is the wrong type, it's being set to the value of `int listbox::get_selected_row()`.

src/map_command_handler.hpp
```diff
@@ -201,7 +201,7 @@ class map_command_handler
 			std::string string_user = get_cmd();
 			int distance = 0;
 			// Minimum length of the two compared strings.
-			int len_min = 0;
+			std::size_t len_min = 0;
```
stevecotton, line 204: `distance` and `len_min` are never accessed outside the `if` block below, so they can just be declared as `const auto` at the point that they're first assigned a meaningful value.

stevecotton (changes requested):
> I'll be going back through and double checking these commits over the next few days, but if there are no other comments by then I plan to go ahead and merge this.

Sorry for not looking at this before (I'd been assuming it was all trivial changes), but this is too large to go through in a single PR. Many of the warnings seem to be signs of bad code that should be improved, but not in the simple way of adding a cast that the compiler would do automatically.

Pentarctagon:
Should I should just close the PR, or something else?

Pentarctagon:
To add: I can split this up into a whole bunch of smaller PRs too, but please define which commits you want in each PR if that's the case.

stevecotton:
Ok, will take me a while to look through it and give a full response.

Pentarctagon:
To be clear: I'm thinking it wouldn't be a huge amount of effort to essentially move this entire set of commits to a different branch and then move one or a few of those commits at a time back to my fork's master branch and open a PR for just those. I don't intend to split commits or move changes between commits.

CMakeLists.txt
Pentarctagon: No, the comment is actually correct - the warning is not applied to the mariadbpp submodule since it comes before the `add_subdirectory("src/modules/mariadbpp/")` command.

stevecotton: Please would you change the comment to make that more obvious?

src/display.hpp
stevecotton: Thanks for the explanation. Not marking as "resolved" yet, because it's a reminder to work out where to improve the documentation.

stevecotton:
I approve all of these, still looking through the others.

* float() -> static_cast<float>
* uint32_t() -> static_cast<uint32_t>
* int32_t() -> static_cast<int32_t>
* Fix -Wshorten-64-to-32: catalog.hpp.
* Fix -Wshorten-64-to-32: frame_private.hpp.
* Fix -Wshorten-64-to-32: listbox.hpp.
* Fix -Wshorten-64-to-32: multimenu_button.hpp.
* Fix -Wshorten-64-to-32: selectable_item.hpp.

For several of the other .hpp files, I'd like changes which then change individual files in the "misc .cpp files" commits.

src/game_config_manager.cpp
```diff
@@ -635,11 +635,11 @@ void game_config_manager::load_addons_cfg()
 		const std::string& msg1 =
 			_n("The following add-on had errors and could not be loaded:",
 			   "The following add-ons had errors and could not be loaded:",
-			   n);
+			   static_cast<int>(n));
```
stevecotton, line 638: Passing the result of `<container>::size()` to `_n` seems a logical thing to do, I think we  should add an overload to src/gettext.hpp that accepts `std::size_t` instead of casting it in multiple places.

Pentarctagon:
> For several of the other .hpp files, I'd like changes which then change individual files in the "misc .cpp files" commits.

What does this mean?

stevecotton:
> > For several of the other .hpp files, I'd like changes which then change individual files in the "misc .cpp files" commits.

For example, #7495 is an alternative to 1c655a6b06f9ab0f2e9a5add78923a7d67805b20's change to random.hpp. However, it then needs a change in attack_prediction.cpp, and that .cpp file is bundled together with many other changes in f3521a48b504fd04bcc3a557ecf71e9ed247f321 "Fix -Wshorten-64-to-32: misc cpp, part 1.". In this case, the change it needs is additional to the ones already in that commit.

However, some of the changes that I want will conflict with changes in the "misc cpp" bundles. For example, #7496 compared to 838c4c1d7df45108b5669d5525fbb73b62600701, one just documents, the other converts some ints in editor_palettes.hpp to std::size_t. However, they're measurements in pixels which are used in SDL_Rrct structures, therefore it makes sense for them to be ints. In the "misc cpp part 1" commit, several `static_cast<int>` statements are added, which wouldn't be necessary if they were ints anyway, so those changes should be removed from the bundle.

Pentarctagon:
I kind of feel like it'd make sense to just squash all these together then. It's just not worth the effort to move things around through these 41 commits so they're organized in the most ideal way - I'll probably end up spending more time dealing with rebase issues and figuring out how to use amend than it took to make the actual code changes.

src/actions/attack.cpp
```diff
@@ -152,7 +152,7 @@ battle_context_unit_stats::battle_context_unit_stats(nonempty_unit_const_ptr up,
 	firststrike = weapon->has_special_or_ability("firststrike");
 
 	{
-		const int distance = distance_between(u_loc, opp_loc);
+		const int distance = static_cast<int>(distance_between(u_loc, opp_loc));
```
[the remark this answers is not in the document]
CelticMinstrel: Barring the invalid location, I think x and y have a minimum of -1. The invalid location is (-999,-999) or something like that.

src/actions/vision.cpp
CelticMinstrel: I thought we already had that…? Should probably use it here. I think it'd look something like:

```cpp
std::size_t team_size = teams.size();
for ( std::size_t i = 1; i != team_size; ++i)
	if ( !target.is_visible_to_team(get_team(i), false) )
```

As a side note, I wouldn't mind if the formatting was fixed here while you're changing other things, though I also wouldn't mind if it's not.

src/ai/contexts.cpp
CelticMinstrel: This is precisely `difference_type`, which almost always expands to `ptrdiff_t`. Probably `auto` is good enough.

src/display.hpp
CelticMinstrel: Then we should remove `viewing_team()` entirely.

src/actions/undo.cpp
Pentarctagon: I'm not trying to make a judgement on whether it's correct or not - the PR is mainly just aimed at making instances like this more explicit in what they're doing.

src/gui/dialogs/unit_advance.hpp
Pentarctagon: If the caller wants it as int for its own reasons, then it's free to do that. It seems kind of weird to have it return a signed value when it can actually only ever return unsigned values though.

src/actions/undo.cpp
stevecotton: Making them explicit will suggest to anyone reviewing the code later that someone made the judgement, so either they'll skip the cleanup, or they'll spend time working with that apparent judgement.

These are just lint warnings - it's good to fix them and I welcome this PR opening the debate, but if we merge as-is then months down the line we'll be used to looking at code, wondering why it does that, running `git blame` and then mumbling "oh, it's that set of quick lint fixes".

src/actions/vision.cpp
gfgtdf: in this case it'd probably be simpler to do (not the scope of this pr know)
```
for ( const team& t: teams())
	if ( !target.is_visible_to_team(t, false) )
			not_seeing.push_back(t.side());
```

src/game_config_manager.cpp
Pentarctagon: So essentially:
```
[[maybe_unused]] inline static std::string _n(const char* str1, const char* str2, std::size_t n)
{ return translation::dsngettext(GETTEXT_DOMAIN, str1, str2, static_cast<int>(n)); }
```
?

CMakeLists.txt
```diff
@@ -462,6 +462,19 @@ else()
...
 
+if(ENABLE_MYSQL)
+	execute_process(COMMAND mysql_config --cflags OUTPUT_VARIABLE MYSQL_CFLAGS OUTPUT_STRIP_TRAILING_WHITESPACE)
+	string(REGEX REPLACE "-I" "" MYSQL_CFLAGS "${MYSQL_CFLAGS}")
+	string(REGEX REPLACE "-DNDEBUG" "" MYSQL_CFLAGS "${MYSQL_CFLAGS}")
+	execute_process(COMMAND mysql_config --libs OUTPUT_VARIABLE MYSQL_LIBS OUTPUT_STRIP_TRAILING_WHITESPACE)
+	add_subdirectory("src/modules/mariadbpp/")
+
+	# ignore this warning for mariadbpp submodule
+	if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
stevecotton, line 473: Needs to be outside the `if(ENABLE_MYSQL)` conditional.

Should probably be `if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")`, but it's getting late so I'm not testing that tonight.

CMakeLists.txt
Pentarctagon: Updated.

src/ai/contexts.cpp
Pentarctagon: Updated.

src/ai/default/aspect_attacks.cpp
Pentarctagon: Updated.

src/gui/dialogs/unit_advance.hpp
Pentarctagon: Updated.

src/map_command_handler.hpp
Pentarctagon: Updated.

CMakeLists.txt
Pentarctagon: Fixed the first part.

This is consistent with how clang is checked for elsewhere in the file though.

src/actions/vision.cpp
CelticMinstrel: Yeah, that looks good to me.

src/actions/undo.cpp
Pentarctagon: Added a new commit with an alternate change for this.

stevecotton:
Chipping away at merging these - would everyone be happy merging the first two commits (config.hpp and the income parts of team.hpp)?

Pentarctagon:
I think it'd make more sense for me to merge this and then it can be a reference for your changes that you can add whenever you want on top of it, rather than having this sit here in the mean time and me needing to keep rebasing it and dealing with potential conflicts.

Vultraz:
Seconded on the LGTM.

stevecotton:
> I think it'd make more sense for me to merge this and then it can be a reference for your changes that you can add whenever you want on top of it, rather than having this sit here in the mean time and me needing to keep rebasing it and dealing with potential conflicts.

If the options are a binary "close or merge", then my preference is "close". The commit that overloads `_n` reverted several chunks from earlier in this PR, and I think many of the other changes that would be done on top would similarly revert lines. So it's easier to just start without this PR, and then have the compiler warnings pointing to things that still need to be investigated.

I've got a local branch to track the commits in this PR, so you can move your `master` branch without me loosing them.

src/display.hpp
Vultraz: Seconded.

PR #6057, opened 2021-08-31
update NR difficulty

cooljeanius:
NR is notoriously difficult and annoying, so this PR attempts to soften it a little bit on the easier difficulties (not on HARD, though). These changes were originally made during my 1.14 playthru of NR, and haven't been tested on 1.15/1.16/1.17 yet, although I'm really dreading having to test them again... I'm leaving this marked as a draft because there are rejects files for map files that I haven't managed to wiggle back in yet, because this should really wait for until after ~~1.16~~ ***1.18*** is released, and because I still need to do another pass at editing/testing.

PR #7452, opened 2023-03-14
Try to fix map diff failure when terrain doesn't have an image associated with it

cooljeanius:
I'm just trying some suggestions from @macabeus to see if we can fix the map diff failure in #6057
(idk if this is even actually valid TypeScript, much less whether it will actually work or not)


utils/wesnoth-map-diff/package.json
```diff
@@ -12,6 +12,7 @@
   "license": "GNU GPL v2+",
   "dependencies": {
     "jimp": "0.16.1",
+    "npm": "^9.2.0",
```
macabeus, line 15: We don't need to include `npm` as a dependency here
```suggestion
```

utils/wesnoth-map-diff/src/images.ts
```diff
@@ -22,6 +22,9 @@ const getDictTerrainType2ImagesPath = async () => {
       }
 
       acc[terrainType.string.value] = terrainType.symbol_image.value
+      if (!acc[terrainType.string.value]) {
```
macabeus, line 25: It isn't enough. On line 20, we have an early return when `symbol_image` is `undefined`.

We should change the logic to something like this:
```ts
.reduce((acc, terrainType) => {
  if (terrainType.symbol_image) {
    acc[terrainType.string.value] = terrainType.symbol_image.value
  }

  if (terrainType.string) {
    acc[terrainType.string.value] = terrainType.string.value
  }

  return acc
```

utils/wesnoth-map-diff/src/images.ts
```diff
@@ -36,7 +39,10 @@ const readTerrainImages = async () => {
   const promises = Object
     .entries(dictTerrainType2ImagesPath)
     .map(async ([terrainType, imageName]) => {
-      const image = await Jimp.read(`${imageBasepath}/${imageName}.png`)
+      image = await Jimp.read(`${imageBasepath}/${imageName}.png`)
+      if (!image) {
+        image = await Jimp.read(`../../../images/${imageName}.png`)
```
macabeus, line 44: Is it really work?
On the folder `/images` we have no `png` file

utils/wesnoth-map-diff/src/images.ts
```diff
@@ -71,10 +77,12 @@ const produceImagesGetter = async (): Promise<ImagesGetter> => {
         throw new Error(`Missing image for "${baseCode}"`)
       }
 
-      // todo: we should use the defaultBase correctly
+      // TODO: we should use the defaultBase correctly
       if (miscCode) {
         const miscImage = images[`^${miscCode}`]
         if (!miscImage) {
+          // FIXME: this is reached by certain invisible overlays,
+          // such as "^Xo":
```
macabeus, line 85: Since this PR is already an attempt to fix this issue, we don't need to include this comment
```suggestion
```

utils/wesnoth-map-diff/src/images.ts
```diff
@@ -22,6 +22,9 @@ const getDictTerrainType2ImagesPath = async () => {
       }
```
macabeus, line 22: > idk if this is even actually valid TypeScript, much less whether it will actually work or not

Can you follow the package readme? It explains how to run it.
Let me know if you need some help on it

utils/wesnoth-map-diff/src/images.ts
cooljeanius: I dunno, I was just making something up

utils/wesnoth-map-diff/src/images.ts
cooljeanius: so, like in [8961e19](https://github.com/wesnoth/wesnoth/pull/7452/commits/8961e19845a7c863f6cce77157df2a39cf07ff77)?

utils/wesnoth-map-diff/src/images.ts
cooljeanius: I did when I first created this branch, but that was a few months ago and I forgot about it... (it's also how I ended up with the npm dependency that you told me to remove in [a24458d](https://github.com/wesnoth/wesnoth/pull/7452/commits/a24458d5d204f6effabffbe3d4fb0958e2123af0))

cooljeanius: ok I just tried doing `npm run build:dev` and this was the output it printed:
```
npm run build:dev
npm WARN cli npm v9.6.0 does not support Node.js v15.14.0. You should probably upgrade to a
npm WARN cli newer version of node as we can't make any promises that npm will work with this
npm WARN cli version. This version of npm supports the following node versions: `^14.17.0 ||
npm WARN cli ^16.13.0 || >=18.0.0`. You can find the latest version at https://nodejs.org/.

> wesnoth-map-diff@0.0.1 build:dev
> tsc -p tsconfig-dev.json

src/images.ts:40:7 - error TS2552: Cannot find name 'image'. Did you mean 'Image'?

40       image = await Jimp.read(`${imageBasepath}/${imageName}.png`)
         ~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:17148:13
    17148 declare var Image: {
                      ~~~~~
    'Image' is declared here.

src/images.ts:41:12 - error TS2552: Cannot find name 'image'. Did you mean 'Image'?

41       if (!image) {
              ~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:17148:13
    17148 declare var Image: {
                      ~~~~~
    'Image' is declared here.

src/images.ts:42:9 - error TS2552: Cannot find name 'image'. Did you mean 'Image'?

42         image = await Jimp.read(`../../../images/${imageName}.png`)
           ~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:17148:13
    17148 declare var Image: {
                      ~~~~~
    'Image' is declared here.

src/images.ts:44:28 - error TS2552: Cannot find name 'image'. Did you mean 'Image'?

44       return [terrainType, image]
                              ~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:17148:13
    17148 declare var Image: {
                      ~~~~~
    'Image' is declared here.


Found 4 errors.

```
...is that because I removed the `const` from `image`? I thought that would be necessary to make a second attempt at assigning to it...

cooljeanius: yeah, if I put the `const` back, I get:
```
> tsc -p tsconfig-dev.json

src/images.ts:42:9 - error TS2588: Cannot assign to 'image' because it is a constant.

42         image = await Jimp.read(`../../../images/${imageName}.png`)
           ~~~~~


Found 1 error.
```

utils/wesnoth-map-diff/src/images.ts
macabeus: Yeah, but with a minor change:
```diff
- if (!acc[terrainType.string.value] && terrainType.string.value) {
+ if (!acc[terrainType.string.value] && terrainType.string) {
```

utils/wesnoth-map-diff/src/images.ts
macabeus: > .is that because I removed the const from image? I thought that would be necessary to make a second attempt at assigning to it...

You always need to declare a variable before using it.
If you need to have a mutable variable, use `let`.
```diff
- const image = await Jimp.read(`${imageBasepath}/${imageName}.png`)
+ let image = await Jimp.read(`${imageBasepath}/${imageName}.png`)
```

utils/wesnoth-map-diff/src/images.ts
cooljeanius: Use the new directory created in [3a92910](https://github.com/wesnoth/wesnoth/pull/7452/commits/3a929108507424b48507b5170727846450e62978) instead? (N.B. I haven't actually put anything there yet...)

macabeus: Yeah. It's a good idea.
Then we can have an image in `utils/wesnoth-map-diff/images/` mapping the terrain string value into an image.
For example, given the terrain:
```
[terrain_type]
    editor_image=impassable-editor
    id=impassable_overlay
    editor_name= _ "Impassable Overlay"
    string=^Xo
    aliasof=_bas,Xt
    mvt_alias=Xt
    editor_group=obstacle, special
    hide_help=yes
[/terrain_type]
```
We should have an image at `utils/wesnoth-map-diff/images/Xo.png`

cooljeanius: > We should have an image at `utils/wesnoth-map-diff/images/Xo.png`

What should it look like? Just copy an existing one?

macabeus: Yeah. Copying the existing one should work well

cooljeanius: > Yeah. Copying the existing one should work well

...actually, if we're just going to be using existing images, why not just fix the path so that it finds the original properly, rather than duplicating things?

macabeus: We can do it as well.
There is nothing within the `Xo` `terrain type` object that makes a link from it to an image named `impassable-editor.png`.

We would need to create some mapper into `map-diff` project to map `Xo` into `impassable-editor.png` (and the same for the other tiles that use a different image, if there is one).

It's a good solution, and we can do it, of course. Its maintenance is probably better as well.
But I was thinking about the most simple solution for now.

If you prefer to create the mapper, it's great.

cooljeanius: I mean... isn't that what the `editor_image=impassable-editor` key is for?

cooljeanius: I think [b116845](https://github.com/wesnoth/wesnoth/pull/7452/commits/b11684560fd3a5026124be0bad91480874dc4fb3) ought to help?

macabeus: Or, great! I overlooked this property 😅
Maybe we can simplify it a bit by removing the last `if`, since it isn't adding an image path into `acc` (which will be an issue later when calling `Jimp.open`)
```diff
-      if (!acc[terrainType.string.value] && terrainType.string) {
-       acc[terrainType.string.value] = terrainType.string.value
-      }
```

cooljeanius: well I mean that's only reached if the previous attempts fail; I don't think it'd hurt to keep it as a backup?

macabeus: Well... if you prefer, we can keep it, of course. But I'm unsure if this attempt will be used.
Is there an image filename that matches `terrainType` is the previous ones failed?

cooljeanius: I'm not sure, but if there isn't, maybe I (or someone) will create one?

macabeus: If there isn't, I think it isn't necessary.
You can instead use a hardcoded filename for a missing image.
For example, [`grid.png`](https://github.com/wesnoth/wesnoth/blob/master/data/core/images/terrain/mask.png) or [`mask.png`](https://github.com/wesnoth/wesnoth/blob/master/data/core/images/terrain/mask.png).
It'll avoid crashing and make us know if there is an image missing.

cooljeanius: ok, I went with `grid` in [ef9ed61](https://github.com/wesnoth/wesnoth/pull/7452/commits/ef9ed61c5b9d37bc942b2d758acb7bca3b909dd5)

macabeus: Great! Can you run the new `map-diff` with the same diff [from this PR](https://github.com/wesnoth/wesnoth/pull/6057), and check if it's working now?

Also, can you run the tests?
`npm run test`

cooljeanius: > Great! Can you run the new `map-diff` with the same diff [from this PR](https://github.com/wesnoth/wesnoth/pull/6057), and check if it's working now?
> 

Eh,  I think it'd be easier to just merge this and then see if it passes

> Also, can you run the tests? `npm run test`

Yeah I did; they passed. Marking this as resolved now.

utils/wesnoth-map-diff/images/.gitignore
```diff
@@ -0,0 +1 @@
+!.
```
macabeus, line 1: We don't need this folder
```suggestion
```

Pentarctagon:
I'm not sure why I was requested as a reviewer?

cooljeanius:
> I'm not sure why I was requested as a reviewer?

You were one of the people in the git history for this directory who had committed to it

utils/wesnoth-map-diff/images/.gitignore
cooljeanius: it's still referenced at line 49 of `images.ts`: https://github.com/wesnoth/wesnoth/pull/7452/files#diff-e229e38d955fc255563574ebed9dc70546a7fba2852ebf7ef1fe3a901cffcc01R49

cooljeanius: We might still put something in it later, though

Pentarctagon:
I'm pretty sure all I did was an auto-update of dependencies when a github bot pointed out a vulnerability in one of them.

cooljeanius:
> I'm pretty sure all I did was an auto-update of dependencies when a github bot pointed out a vulnerability in one of them.

Well, in any case, I'm going to need *someone* to merge this, once it's agreed upon that we want to do so, as I don't think either Macabeus or I can do so ourselves...

utils/wesnoth-map-diff/images/.gitignore
cooljeanius: ...and there. Now, as of [0bbab66](https://github.com/wesnoth/wesnoth/pull/7452/commits/0bbab662312bb53941d1df2f0e35f487000535a9), there is something in it.

Pentarctagon:
From just skimming it, it doesn't look like it's doing anything odd, so as long as it works then I'm good with it.

cooljeanius:
> From just skimming it, it doesn't look like it's doing anything odd, so as long as it works then I'm good with it.

So, I tried merging this PR into #6057 to see if that would fix the map-diff failure there, but it didn't seem to work... which branch controls the version of GitHub Actions that actually get run? Would this have to be merged to master first for any fixing of the map-diff failure in #6057 to be reflected there?

Pentarctagon:
Changes in the PR get added on top of master and then CI is run. If it doesn't work in the PR then it wouldn't work on master either.

utils/wesnoth-map-diff/src/images.ts
```diff
@@ -36,7 +40,21 @@ const readTerrainImages = async () => {
...
+      let image = await Jimp.read(`${imageBasepath}/${imageName}.png`)
+      // Try a number of other places, as well:
+      if (!image) {
+        image = await Jimp.read(`../../../images/${imageName}.png`)
+      }
+      if (!image) {
+        image = await Jimp.read(`./images/${imageName}.png`)
+      }
+      if (!image) {
+        image = await Jimp.read(`../../../attic/${imageName}.png`)
```
soliton-, line 52: What's the point of this?

utils/wesnoth-map-diff/src/images.ts
```diff
@@ -17,11 +17,15 @@ const getDictTerrainType2ImagesPath = async () => {
...
     .reduce((acc, terrainType) => {
-      if (terrainType.symbol_image === undefined) {
-        return acc
+      if (terrainType.symbol_image) {
+        acc[terrainType.string.value] = terrainType.symbol_image.value
+      }
+      if (!acc[terrainType.string.value] && terrainType.editor_image) {
+        acc[terrainType.string.value] = terrainType.editor_image.value 
+      }
+      if (!acc[terrainType.string.value]) {
```
soliton-, line 26: There should probably be a message here so we can see what is missing.

macabeus:
@Pentarctagon 
I'm not sure about that. From the doc: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
> Changes in the PR get added on top of master and then CI is run. If it doesn't work in the PR then it wouldn't work on master either.

macabeus:
@Pentarctagon 
I'm not sure about that. From the doc: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does.

Pentarctagon:
Well, either way, the second part remains true:
> If it doesn't work in the PR then it wouldn't work on master either.

macabeus:
I ran it locally, and it worked. It should work in GitHub Actions as well as when using this new branch.
Note that now it's displaying the `IO` tiles 

image

utils/wesnoth-map-diff/src/images.ts
cooljeanius: how would I do that? @macabeus help?

Pentarctagon:
I'd suggest rebasing this PR against master and then in a separate commit adding something that would have failed before but now won't.

utils/wesnoth-map-diff/src/images.ts
macabeus: You can use `console.log` to output a message in the console - that will be displayed in GitHub Actions.

macabeus: For instance:
```
console.log(`No image defined for the tile "${terrainType.string.value}"`)
```

cooljeanius: added in [32be9d8](https://github.com/wesnoth/wesnoth/pull/7452/commits/32be9d85576193101aa6cf50b095b4125b615429)

utils/wesnoth-map-diff/src/images.ts
```diff
@@ -36,7 +40,21 @@ const readTerrainImages = async () => {
...
+      if (!image) {
+        image = await Jimp.read(`../../../images/${imageName}.png`)
+      }
+      if (!image) {
+        image = await Jimp.read(`./images/${imageName}.png`)
+      }
+      if (!image) {
+        image = await Jimp.read(`../../../attic/${imageName}.png`)
+      }
+      if (!image) {
```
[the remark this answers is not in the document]
cooljeanius: > There should probably be a message here so we can see what is missing.

ok, added in [32be9d8](https://github.com/wesnoth/wesnoth/pull/7452/commits/32be9d85576193101aa6cf50b095b4125b615429)

cooljeanius:
> I'd suggest rebasing this PR against master and then in a separate commit adding something that would have failed before but now won't.

...you sure rebasing was the right thing to do? I did that, but it seems to have added a bunch of unrelated commits... I guess this might need squashing?

cooljeanius:
Superseded by #7499

PR #7454, opened 2023-03-15
NR: Dialog/ Story Rewrite: Scenario One.

TheShadowOfHassen:
So I was talking in issue #7448 and someone said that I could make some dialog changes but thought I remembered some other editing problems in the story. I looked at the first scenario and a lot of things in the story that didn't make sense on reading through. I'd like to, if you don't mind, go through each scenario and do some editing.

I don't want my editing to hurt anyone's feelings, and I tried to separate each change in a different commit. So you can pick and choose the changes and squash the commits you like. If you don't like any of the changes, you can reject them all.

I edited it as well as I could, and I ran it through LanguageTool. A lot of the problems I found were that the sentences were too long and made my eyes cross. I tried to shorten them.

PR #7461, opened 2023-03-18
SotBE: Replace unfortunate wording

Wedge009:
Kapou'e does not appear to be making a comical statement, 'our rear' could be interpreted (admittedly, in a juvenile fashion) as 'rear end', or posterior.

Since #7454 was accepted I think this is a good time to make this change as well (noted this a long time ago but refrained from changing due to the complaints about re-translation overhead).


nemaara (approved):

Wedge009:
@Pentarctagon Looks like I missed the tagging for 1.17.14. But there's no string freeze for master, is there?

Pentarctagon:
Correct.

PR #7470, opened 2023-03-22
DiD: Remove invalid syntax

Wedge009:
Not really sure what's correct here (hence the PR for the CI check), but https://wiki.wesnoth.org/AbilitiesWML#Common_keys_and_tags_for_every_ability seems to suggest that no `[affect_adjacent]` tags is sufficient to disable the affect-adjacent functionality.

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
+++ b/data/campaigns/Descent_Into_Darkness/utils/abilities.cfg
@@ -1095,7 +1095,6 @@
         female_name= _ "female^masochism"
         description= _ "This unit gains up to 50% damage as its hitpoints are reduced."
         affect_self=yes
-        affect_adjacent=no
     [/leadership]
 #enddef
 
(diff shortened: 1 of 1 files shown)


nemaara (approved):
Does it need octalots approval too or just merge?

stevecotton:
I'm AFK and can't see why the CI failed, but it looks like the "enlightened" string should be translatable. A po comment would be good too, to say whether it's an adjective or a past-tense verb; or to say it shouldn't be translated.

The wmllint directive on the next line has a typo too.

CelticMinstrel:
So yeah, it needs to be one of:

1. A translatable string.
2. A string with variable substitutions (not necessarily translatable).
3. An integer.

Obviously 1 is the correct answer in this case.

Wedge009:
> Obviously 1 is the correct answer in this case.

I was going to do that but then I saw the explicit mark check disabling so I left as a non-translation string. I'll change it over then.

PR #7483, opened 2023-03-25
EBO for unique_ptr with SDL_RWops

bencsikandrei:
unique_ptr is designed to allow EBO (empty base class optimization)
for the use case of SDL_RWops this fits well and allows the UP to be half the size it is now (8 bytes vs 16 bytes on x86-64)

see this compiler explorer link:
https://godbolt.org/z/KjnTeqhbv

it also allows easier construction of the unique_ptrs as the ctor no longer requires passing the address of the function to call
since runtime setting of the deleter is not required, that's an added bonus

### DIFF
--- a/src/filesystem.hpp
+++ b/src/filesystem.hpp
@@ -39,7 +39,13 @@ namespace filesystem {
 using scoped_istream = std::unique_ptr<std::istream>;
 using scoped_ostream = std::unique_ptr<std::ostream>;
 
-typedef std::unique_ptr<SDL_RWops, void(*)(SDL_RWops*)> rwops_ptr;
+struct sdl_rwops_deleter
+{
+	void operator()(SDL_RWops*) const noexcept;
+};
+
+using rwops_ptr = std::unique_ptr<SDL_RWops, sdl_rwops_deleter>;
+static_assert(sizeof(rwops_ptr) == 8);
 
 rwops_ptr make_read_RWops(const std::string &path);
 rwops_ptr make_write_RWops(const std::string &path);
--- a/src/filesystem_sdl.cpp
+++ b/src/filesystem_sdl.cpp
@@ -41,8 +41,13 @@ static std::size_t SDLCALL ofs_write(struct SDL_RWops *context, const void *ptr,
 static int SDLCALL ifs_close(struct SDL_RWops *context);
 static int SDLCALL ofs_close(struct SDL_RWops *context);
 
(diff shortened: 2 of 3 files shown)


gfgtdf:
Looks good to me, I'm not sure whether it is performance wise relevant, but i like the new code more.

bencsikandrei:
There's probably little to no perf improvement especially since this is used rarely. But I feel it's nice that it's smaller and easier to construct

bencsikandrei:
Is there a way to re-run only the failing step of the CI?
I can't find any way of doing that myself
I could also rebase/amend and that would re-trigger the whole thing but that seems like a waste

Pentarctagon:
The failure is just the github PR labeler, so it's fine.

src/filesystem.hpp
```diff
@@ -39,7 +39,13 @@ namespace filesystem {
...
 using scoped_ostream = std::unique_ptr<std::ostream>;
 
-typedef std::unique_ptr<SDL_RWops, void(*)(SDL_RWops*)> rwops_ptr;
+struct sdl_rwops_deleter
+{
+	void operator()(SDL_RWops*) const noexcept;
+};
+
+using rwops_ptr = std::unique_ptr<SDL_RWops, sdl_rwops_deleter>;
+static_assert(sizeof(rwops_ptr) == 8);
```
soliton-, line 48: Either this assert should have a useful message or it should be removed.

Pentarctagon: The point being that it's not necessarily self-documenting - what's plainly obvious to you isn't necessarily so obvious to others.

bencsikandrei: I see, you're right, I'm on it, will remove the assert ASAP (I feel it's better to remove it)

bencsikandrei: the code has now changed and it's much better than what I did, should I still add a comment? should I remove the static_assert?

[here's a link to the latest version of it](https://github.com/wesnoth/wesnoth/commit/0e312fb74de309b2c2516e38b95b36ff2376222c)

soliton-: See https://github.com/wesnoth/wesnoth/commit/7bff35f991ee6bb4db4918a9165ad921a69086a7#r106243393
There is still no hint on why it's there so if there is no reason other than testing EBO then it should be removed.

bencsikandrei: on it

PR #7484, opened 2023-03-25
Create 'priority' attribute  for plague used like leadership

newfrenchy83:


Pentarctagon:
What does this do? I lost track of this part from the other PR.

newfrenchy83:
> What does this do? I lost track of this part from the other PR.

if plague is used like leadership and two abilities plague influence same unit, priority=highest_value impose lague who has that regarlessof position of leaders.

CelticMinstrel:
Basically: in a situation where a unit is affected by two leader-like units that impose a plague special on their weapon attack, this controls which of those two plague specials takes effect (since obviously the unit can't be transformed into two different things).

I believe without this the selection would be arbitrary and depend on the positioning of the units.

gfgtdf:
It would of course also work when the unit has different [plague] abilities, i'm not against this feature.

CelticMinstrel:
> It would of course also work when the unit has different [plague] abilities

Yeah, it could for example be useful if a unit has two different plague abilities that activate under different conditions, but there's some overlap between the conditions.

newfrenchy83:
> > It would of course also work when the unit has different [plague] abilities
> 
> Yeah, it could for example be useful if a unit has two different plague abilities that activate under different conditions, but there's some overlap between the conditions.

actually, the plague slected is the first of the list, but tit is on leadership using what it is a problem, but i don't think at activation condition who overlap

src/units/abilities.cpp
```diff
@@ -1256,6 +1256,38 @@ namespace { // Helpers for attack_type::special_active()
 
 }//anonymous namespace
 
+unit_ability_list attack_type::get_specials_and_abilities_with_priority(const std::string& special) const
+{
+	//the custom_priority attribute is used for define rules of priority of damage type
```
CelticMinstrel, line 1261: This comment mentions damage type, but the damage type case no longer exists.

src/units/abilities.cpp
```diff
@@ -1229,6 +1229,40 @@ namespace { // Helpers for attack_type::special_active()
...
+		}
+	}
+
+	if(!priorityvec.empty()){
+		double priority = *(std::min_element(priorityvec.begin(), priorityvec.end()));
+		int prc = std::count(priorityvec.begin(), priorityvec.end(), priority);
+		for(const auto& i : abil_list) {
+			if((*i.ability_cfg)["priority"].to_double(0) == priority) {
+				temp_list.emplace_back(i);
+				if(prc == 1){
```
stevecotton, line 1254: I don't understand what this is doing. I think `prc` is meant to optimise the loop when there's only one ability with the top priority, and the `return temp_list` returns the same result as the later `if(!temp_list.empty())` conditional would do without this optimisation.

However, calculating `prc` requires an extra scan of the list, so it isn't an optimisation.

src/units/abilities.cpp
```diff
@@ -1229,6 +1229,40 @@ namespace { // Helpers for attack_type::special_active()
 
 }//anonymous namespace
 
+unit_ability_list attack_type::get_specials_and_abilities_with_priority(const std::string& special) const
```
stevecotton, line 1232: I suggest putting this directly beneath `get_specials_and_abilities`, as it continues on from that code.

newfrenchy83: then all weapon specials could use that.

newfrenchy83: i refuse to doing that because mone specials whreturn string value must be concerned by that.

newfrenchy83: excuse me, i don't understand that what uoou said. I will  relocate function

src/units/abilities.cpp
newfrenchy83: removed

src/units/abilities.cpp
newfrenchy83: function relocated

src/units/abilities.cpp
```diff
@@ -1264,14 +1265,66 @@ unit_ability_list attack_type::get_specials_and_abilities(const std::string& spe
...
+		}
+		if(!checking_string.empty()){
+			unit_ability_list priority_list;
+			int i = 0;
+			int n = checking_string.size();
+			while (i>=0 && i<n){
+				for(const auto& j : temp_list) {
+					if((*j.ability_cfg)["type"].str() == checking_string.at(i)){
+						priority_list.emplace_back(j);
+						return priority_list;
```
stevecotton, line 1317: For both `plague` and PR #7003's `damage_type`, the caller only uses the `front()` item of the returned list. If all of the callers will only want one item, it would be clearer to me and less code to return either a `std::optional` or a list that's documented to contain at most one item.

```
If `get_specials_and_abilities(special)` would return an empty list, returns `std::nullopt`.
Otherwise, returns the highest priority item in the list.
If two or more items are tied for highest priority, returns the one that was closer to the start of the list.
```

newfrenchy83: i will document

newfrenchy83: if use of  checking_string approved, can i remove use of priority attribute?

src/units/abilities.cpp
```diff
@@ -1264,14 +1265,49 @@ unit_ability_list attack_type::get_specials_and_abilities(const std::string& spe
...
+	//Otherwise, returns the highest priority item in the list.
+	//If two or more items are tied for highest priority, returns the one that was closer to the start of the list.
+	unit_ability_list abil_list = get_specials_and_abilities(special);
+	if(abil_list.empty()){
+		return abil_list;
+	}
+
+
+	std::vector<std::string> checking_string;
+	std::vector<config> lists = resources::gamedata->get_special_string_list();
```
stevecotton, line 1285: If there's more than one list for a given type, which list should win the next loop?

My worry here is, what if a multiplayer scenario and an era both provided such a list? What should be the rules for who provides the list?

gfgtdf:
Was there any discussion that i missed on why this this or has changed or what the advantages of this over the original suggestion are? So far having extra keys (with a very generic "special_string_list" name) in scenario seems harder to use and harder to understand that the priority attribute.

src/units/abilities.cpp
newfrenchy83: i will said the first to be loaded, but i understand and remove that.

src/units/abilities.cpp
newfrenchy83: i willremain on priority attribute

src/units/abilities.cpp
newfrenchy83: if i judge by 389:saved_game::expand_mp_events() in saved_game.cpp theorder should be: era, campaign and resource. scenario lodaded before all other(expand_scenario();).

stevecotton:
This changes the behavior when a unit has a plague weapon special, and is adjacent to a teacher, and neither ability has a priority attribute. Previously the units own weapon special would be used, with this PR the taught ability would be used; I think using the weapon special is more logical, although which option is better would depend on the units. PR #7738 has a draft `plague_without_priority` test that shows what I mean.

> So far having extra keys (with a very generic "special_string_list" name) in scenario seems harder to use and harder to understand that the priority attribute.

That got removed again in later pushes.

newfrenchy83:
> This changes the behavior when a unit has a plague weapon special, and is adjacent to a teacher, and neither ability has a priority attribute. Previously the units own weapon special would be used, with this PR the taught ability would be used; I think using the weapon special is more logical, although which option is better would depend on the units. PR #7738 has a draft `plague_without_priority` test that shows what I mean.
> 
> > So far having extra keys (with a very generic "special_string_list" name) in scenario seems harder to use and harder to understand that the priority attribute.
> 
> That got removed again in later pushes.

i repull [plague] in [specials] prior

stevecotton:
With the repull commit, the priority doesn't allow overriding weapon specials, even with priority.

In the test that I provided, give Charlie's ability `priority=1`. I'd now expect Charlie's ability to override Bob's, and a Chocobone to spawn, but it still spawns a Blood Bat.

newfrenchy83:
> With the repull commit, the priority doesn't allow overriding weapon specials, even with priority.
> 
> In the test that I provided, give Charlie's ability `priority=1`. I'd now expect Charlie's ability to override Bob's, and a Chocobone to spawn, but it still spawns a Blood Bat.

Ok, i will remove that then?

newfrenchy83:
> With the repull commit, the priority doesn't allow overriding weapon specials, even with priority.
> 
> In the test that I provided, give Charlie's ability `priority=1`. I'd now expect Charlie's ability to override Bob's, and a Chocobone to spawn, but it still spawns a Blood Bat.

commit removed

newfrenchy83:
i will close this PR and concnetrate on https://github.com/wesnoth/wesnoth/pull/7746

PR #7488, opened 2023-03-26
Minor fixes filesystem

bencsikandrei:
Remove a couple cases where the same header was included in both .cpp and .hpp of some modules. No harm really, but it looks like it was either the result of a merge conflict resolution or simply an omission.
Since I was there, remove copying of `default_blacklist` in multiple TUs, which include `filesystem.hpp`, by using extern (instead of a static const) inside the header. TBH `inline` would have worked just the same, but since the filesystem module already has a .cpp might as well just put it to use.

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -18,14 +18,14 @@
  * Routines to set up the display, scroll and zoom the map.
  */
 
+#include "display.hpp"
+
 #include "arrow.hpp"
 #include "color.hpp"
 #include "cursor.hpp"
-#include "display.hpp"
 #include "draw.hpp"
 #include "draw_manager.hpp"
 #include "fake_unit_manager.hpp"
-#include "filesystem.hpp"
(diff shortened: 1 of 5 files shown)


src/display.cpp
```diff
@@ -18,14 +18,14 @@
  * Routines to set up the display, scroll and zoom the map.
  */
 
+#include "display.hpp"
```
bencsikandrei, line 21: display.hpp already needs to include filesystem.hpp

src/serialization/preprocessor.cpp
```diff
@@ -23,7 +23,6 @@
 
 #include "buffered_istream.hpp"
 #include "config.hpp"
-#include "filesystem.hpp"
```
bencsikandrei, line 26: we can see below that the .hpp of this module already needs filesystem.hpp

PR #7490, opened 2023-03-27
DiD S8: Disallow undead bat from retrieving the Book

### LINKED ISSUE #7485: [BUG][Descent into Darkness][S8 - Alone at Last] Walking Corpse (Bat) can take book item

- **Version:** 1.17.14+dev
- **Downloaded from:** Github
- **Build info:** Msys2 Compiler with Scons cmd
- **OS:** windows 10

### Description of the bug

Scenario instructions state that the Bats and Ghosts cannot carry the book. This is true..
BUT..a walking corpse of a bat can carry the book.

1. Take the book with a Walking Corpse of Bat variant

### Expected behavior

Bat Corpses cannot carry the book of the Dead.

### Additional context

image

Wedge009:
Resolves #7485.

I tested with 'Walking' Corpse Bat, Soulless Bat, and regular (humanoid) Walking Corpse and regular Soulless. First two behave the same as live bats (aside from not squeaking) while the latter two behave as a regular unit picking up the book normally. Re-using Malin's line about the bat avoids translation overhead.

I think this should be back-ported too.

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/08_Alone_at_Last.cfg
@@ -950,6 +950,10 @@
                 type_adv_tree=Ghost,Vampire Bat
             [/not]
             [not]
+                type_adv_tree=Walking Corpse
+                variation=bat
+            [/not]
+            [not]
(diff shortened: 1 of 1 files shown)


Wedge009:
Hmm, looks like vcpkg failure again.

nemaara (approved):

PR #7494, opened 2023-03-29
Overhaul image cache to be hash-based rather than index-based

Vultraz:
Gets rid of that dreadful last_index stuff. It basically meant a slot was reserved for every image locator in every cache (even if never populated) rather than only the needed cache. The index was also never reset for the lifetime of the program. This new design uses a hashmap based on the locators themselves.

### DIFF
--- a/src/picture.hpp
+++ b/src/picture.hpp
@@ -18,6 +18,7 @@
 #include "map/location.hpp"
 #include "terrain/translation.hpp"
 
+#include <optional>
 #include <unordered_map>
 
 class surface;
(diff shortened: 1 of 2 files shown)


src/picture.hpp
```diff
@@ -74,9 +75,9 @@ class locator
 
 	locator& operator=(const locator& a);
 
-	bool operator==(const locator& a) const { return index_ == a.index_; }
-	bool operator!=(const locator& a) const { return index_ != a.index_; }
-	bool operator<(const locator& a) const { return index_ < a.index_; }
+	bool operator==(const locator& a) const { return hash() == a.hash(); }
+	bool operator!=(const locator& a) const { return !operator==(a); }
+	bool operator<(const locator& a) const { return hash() < a.hash(); }
```
Pentarctagon, line 80: What's the purpose of the overloaded less than operator? Comparing whether one hash is less than another hash seems like it wouldn't be especially useful?

CelticMinstrel: Hold on, this is just completely wrong. You can't implement operator== by comparing the hash, that completely defeats the point of using the locator instead of the hash as a key.

Vultraz: Resolved both.

PR #7496, opened 2023-03-29
Editor palettes: fix -Wshorten-64-to-32 and improve docs

stevecotton:
Use int for measurements of pixels, vs std::size_t for container sizes.

The scroll_bottom() methods were never called, although scroll_top() is used when switching between subgroups. The logic is scroll_bottom() was broken; items_start wasn't changed inside the loop, so the loop would run at most once.

### DIFF
--- a/src/editor/palette/editor_palettes.cpp
+++ b/src/editor/palette/editor_palettes.cpp
@@ -162,7 +162,10 @@ std::size_t editor_palette<Item>::active_group_index()
 template<class Item>
 void editor_palette<Item>::adjust_size(const SDL_Rect& target)
 {
-	const int items_fitting = (target.h / item_space_) * columns_;
+	// The number of columns is passed to the constructor, and isn't changed afterwards. It's likely to be
+	// exactly 4, but will always be a small number which makes the next cast reasonable.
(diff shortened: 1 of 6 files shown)


stevecotton:
Switching to draft because it probably still needs some (different) `static_casts` in editor_palettes.cpp.

Pentarctagon (approved):

CelticMinstrel:
On the one hand, removing `scroll_bottom()` seems wrong – it should probably be used somewhere! But on the other hand, the code in there looks like it never actually worked. I think it's an infinite loop?

stevecotton:
It's not infinite because `items_start` doesn't change within the loop, so the loop ends after at most one iteration.

Although having a `scroll_top()` implies there should be a `scroll_bottom()`, I think the only thing we need is an implementation of `scroll_to_somewhere_logical_after_changing_pages()`. I'm not suggesting renaming `scroll_top()`, just pointing out that naming it for its duty wouldn't imply the existence of a `scroll_bottom()`.

CelticMinstrel:
Well, my thought was that pressing Home would scroll to the top and pressing End would scroll to the bottom, but that's certainly not a very important feature, and also Home _doesn't_ do that right now so I guess it would be a feature request.

PR #7497, opened 2023-03-30
deploy and refactor config::optional_child

gfgtdf:


gfgtdf:
okay i... t seems like git add -A also adds unstaged files .. ignore that code in src/play_flow/

data/campaigns/World_Conquest/lua/map/main.lua
```diff
@@ -31,14 +31,19 @@ local function get_map_generator(scenario_data)
 	end
 end
 
+local function get_scenario_data(nplayers, scenario_num)
+	wesnoth.dofile(string.format("./scenarios/WC_II_%dp_scenario%d.lua", nplayers, scenario_num))
```
CelticMinstrel, line 35: …isn't this missing a return…?

data/campaigns/World_Conquest/lua/map/postgeneration_utils/engine.lua
```diff
@@ -41,6 +41,10 @@ function set_terrain_impl(data)
 	end
 end
 
+
+-- Similar to multiple set_terrain calls (see below), but it first runs all the
```
CelticMinstrel, line 45: Could I ask you to use the same format as the core LuaDoc for these comments? That is, three `---` and the various `@metadata` commands.

data/campaigns/World_Conquest/lua/map/main.lua
gfgtdf: oh indeed thx

data/campaigns/World_Conquest/lua/map/postgeneration_utils/engine.lua
gfgtdf: hmm i could, do you have a link to a sample documentation ?

data/campaigns/Under_the_Burning_Suns/scenarios/01_The_Morning_After.cfg
```diff
@@ -647,6 +647,7 @@
         [/filter]
 
         # Allow to undo the move before the event is triggered
+        # TODO: remove this its probably unsafe.
```
CelticMinstrel, line 650: Maybe it could be made safe by unregistering the moveto event handler added on line 668 in `[on_undo]`?

src/config.hpp
```diff
@@ -341,6 +443,8 @@ class config
 	 *       For instance, -1 is the index of the last child.
 	 */
 	config& child(config_key_type key, int n = 0);
+	config& manditory_child(config_key_type key, int n = 0);
+	const config& manditory_child(config_key_type key, int n = 0) const;
```
CelticMinstrel, line 447: This is misspelled, should be `mandatory_child`.

src/game_config_view.cpp
```diff
@@ -49,27 +49,36 @@ config_array_view game_config_view::child_range(config_key_type key) const
...
+const config& game_config_view::manditory_child(config_key_type key) const
 {
 	for(const config& cfg : cfgs_) {
 		if(const auto res = cfg.optional_child(key)) {
 			return res.value();
 		}
 	}
-	const config cfg;
-	return cfg.child("invalid");
+	throw "missing anditory child";
```
CelticMinstrel, line 71: Missing M. Also should probably be a custom exception type instead of a `const char*`. If you can't be bothered, at least make it `std::string`.

data/campaigns/World_Conquest/lua/map/postgeneration_utils/engine.lua
CelticMinstrel: Any file in data/lua/core/ for example.

src/ai/configuration.cpp
```diff
@@ -217,7 +217,7 @@ bool configuration::parse_side_config(side_number side, const config& original_c
 	DBG_AI_CONFIGURATION << "side " << side << ": config contains:"<< std::endl << cfg;
 
 	//insert default config at the beginning
-	if (default_config_) {
+	if (true) {
```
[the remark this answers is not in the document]
gfgtdf: `default_config_` (defined in the header file) is not a reference.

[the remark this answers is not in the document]
CelticMinstrel: No, I mean… if it's always true, why is there an if at all?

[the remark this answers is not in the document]
gfgtdf: probably because the person who write that code assumed that ` default_config_ = ais.child("default_config");` a few lines above would rebind the  default_config_ object instead of invoking `config::operator=`. One could fix that code, but i explicitly decided not to change behviour in this commit too much since its already complicated enough.

data/campaigns/Under_the_Burning_Suns/scenarios/01_The_Morning_After.cfg
gfgtdf: Hmm yes we could,  although it would probably be simpler to move the line 668 handler out of there and control it via a variable which is then reset in [on_undo]. Also the outer eents need first_time_only=no for that to work. (so that it can fire egain of redoing)

src/config.cpp
```diff
@@ -485,9 +417,8 @@ config& config::child_or_add(config_key_type key)
 	return add_child(key);
 }
 
-utils::optional_reference<const config> config::get_deprecated_child(config_key_type old_key, const std::string& in_tag, DEP_LEVEL level, const std::string& message) const {
-	check_valid();
-
+utils::optional_reference<const config> config::get_deprecated_child(config_key_type old_key, const std::string& in_tag, DEP_LEVEL level, const std::string& message) const
```
CelticMinstrel, line 420: Should this be changed to an `optional_config` too?

src/persist_var.cpp
```diff
@@ -68,17 +68,17 @@ static void get_global_variable(persist_context &ctx, const vconfig &pcfg)
 	config::attribute_value pcfg_side = pcfg["side"];
 	const int side = pcfg_side.to_int(resources::controller->current_side());
 	persist_choice choice(ctx, global, side);
-	config cfg = mp_sync::get_user_choice("global_variable",choice,side).child("variables");
+	config cfg = mp_sync::get_user_choice("global_variable",choice,side).mandatory_child("variables");
```
CelticMinstrel, line 71: Doesn't it look like this should be an `optional_child`? Given that things were trying to test its validity further down.

src/gui/dialogs/addon/manager.cpp
```diff
@@ -545,7 +545,7 @@ void addon_manager::toggle_details(button& btn, stacked_widget& stk)
 void addon_manager::fetch_addons_list()
 {
 	client_.request_addons_list(cfg_);
-	if(!cfg_) {
+	if(false) {
```
CelticMinstrel, line 548: This also seems like a config that should maybe be optional.

src/ai/configuration.cpp
```diff
@@ -48,9 +48,9 @@ void configuration::init(const game_config_view &game_config)
 	era_ai_configurations_.clear();
 	mod_ai_configurations_.clear();
 
-	const config &ais = game_config.child("ais");
-	default_config_ = ais.child("default_config");
-	if (!default_config_) {
+	const config &ais = game_config.mandatory_child("ais");
+	default_config_ = ais.mandatory_child("default_config");
+	if ((false)) {
```
CelticMinstrel, line 53: This also seems like a place that should use `optional_child`. In this case I guess It'd look something like "if(optional_child) { default_config_ = … }`

src/ai/composite/aspect.hpp
```diff
@@ -250,8 +250,8 @@ class composite_aspect : public typesafe_aspect<T> {
 			add_facet(-1,cfg_element);
 		}
 
-		config _default = this->cfg_.child("default");
-		if (_default) {
+		config _default = this->cfg_.mandatory_child("default");
+		if ((true)) {
```
CelticMinstrel, line 254: Again, doesn't this look like it should be an `optional_child`?

src/ai/configuration.cpp
```diff
@@ -217,7 +217,7 @@ bool configuration::parse_side_config(side_number side, const config& original_c
 	DBG_AI_CONFIGURATION << "side " << side << ": config contains:"<< std::endl << cfg;
 
 	//insert default config at the beginning
-	if (default_config_) {
+	if ((true)) {
```
CelticMinstrel, line 220: I guess this should just test `if(!default_config_.empty()`?

src/ai/default/ca.cpp
```diff
@@ -277,9 +277,9 @@ move_leader_to_goals_phase::~move_leader_to_goals_phase()
 double move_leader_to_goals_phase::evaluate()
 {
 
-	const config &goal = get_leader_goal();
+	config goal = get_leader_goal();
 	//passive leader can reach a goal
-	if (!goal) {
+	if ((false)) {
```
CelticMinstrel, line 282: Any reason not to use an `optional_child` in `get_leader_goal()`? I guess that's worked into the aspects system so it might be easier to test here `if(goal.empty())`.

src/server/campaignd/server.cpp
```diff
@@ -433,7 +433,7 @@ void server::load_config()
 			config data;
 			in = filesystem::istream_file(filesystem::normalize_path(addon_file));
 			read_gz(data, *in);
-			if(!data) {
+			if ((false)) {
```
CelticMinstrel, line 436: I think this should probably look something like `if(!read_gz(data, *in))`. I don't know exactly what `read_gz` looks like so it might be a bit different, but something in that spirit.

data/campaigns/World_Conquest/lua/map/postgeneration_utils/engine.lua
CelticMinstrel: I guess this isn't in the PR anymore, will it be committed separately?

src/ai/configuration.cpp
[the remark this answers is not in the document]
CelticMinstrel: Continued at https://github.com/wesnoth/wesnoth/pull/7497#discussion_r1155214338

gfgtdf:
@CelticMinstrel I tried to avoid changing behaviour in this pr, I'll probably look at those if (false)/if (true) later

CelticMinstrel:
> @CelticMinstrel I tried to avoid changing behaviour in this pr, I'll probably look at those if (false)/if (true) later

???

What are you talking about? Doesn't _introducing_ those if(true/false) change behaviour?

gfgtdf:
> What are you talking about? Doesn't introducing those if(true/false) change behaviour?

No, before they were check that always returns true/false i just replaced them with true/false directly

CelticMinstrel:
> No, before they were checks that always returns true/false i just replaced them with true/false directly (see the commits description for more information)

Um… I don't think that's true? At least, not for all of them. Some of those are places where it looks like you wrongly used `mandatory_child` instead of `optional_child`. Some might be places where the code was wrongly checking config validity, sure. But at least these three look like they change behaviour in this PR:

https://github.com/wesnoth/wesnoth/pull/7497#discussion_r1155213853
https://github.com/wesnoth/wesnoth/pull/7497#discussion_r1155214011
https://github.com/wesnoth/wesnoth/pull/7497#discussion_r1155214206

gfgtdf:
The old code literally tested for the _local variable_ `_default` ... .The local variable can never be a reference to the invalid config (which is what operator bool tests) since its not a reference.

CelticMinstrel:
I think I basically understand your point – there was a huge hole in the old config validity test and these few cases fell through that hole and never worked. Still, leaving in extremely obvious dead code doesn't seem like a good idea. I'd still prefer to see them all fixed in this PR one way or another, though keeping it in a separate commit would be fine.

gfgtdf:
> . Personally I'd still prefer to see them all fixed in this PR

No, i really want to merge this pr asap, in particular since it is very susceptible to merging conflics. (or even problems that won't even be visible as merging conflics when a new code uses .child() ), So ill probably merge if after this test passes.

CelticMinstrel:
> (or even problems that won't even be visible as merging conflics when a new code uses .child() )

But these at least would be a build failure since .child() no longer exists.

gfgtdf:
Its not half finished, it does exactly what i intended to do when i started this.

CelticMinstrel:
It _is_ half finished – it leaves in "if(true)" and "if(false)" statements that are intended to be removed at a later time. What is that if not half finished?

gfgtdf:
These if true/false were already there, just disguised, Fixing them is just not the scope of this pr.

PR #7502, opened 2023-03-31
Fix -Wshorten-64-to-32: config.hpp and config.cpp

stevecotton:
Use std::size_t instead of unsigned int for container sizes.

There's one place where it still uses signed int, for using a negative index to access children from the end of the array. Overflow shouldn't be a worry here, as it would require 2**31 objects to be in the container, implying the container and children would use at least 32GB even if every child object was empty.

This expanded from review comments on Pentarctagon's #7451, seemed sensible to just make a PR. It goes in a different direction from that PR, trying to clean up the code's behavior instead of documenting what the compiler was already doing to the old code.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -369,7 +369,7 @@ config::const_child_itors config::child_range(config_key_type key) const
 	return const_child_itors(const_child_iterator(p->begin()), const_child_iterator(p->end()));
 }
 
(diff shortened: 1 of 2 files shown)


stevecotton:
@Pentarctagon any objection to me merging this? I'm planning to make small PRs rather than add more files to this one.

Pentarctagon (approved):

PR #7509, opened 2023-04-02
 Fix a bug in [event][filter_attack] when trying to use special_id/type_active to detect a special whose activity is conditional.

newfrenchy83:
Fix a bug in [event][filter_attack] when trying to use special_id/type_active to detect a special whose activity is conditional.

When we try to trigger an event conditioned by the use of a special whose activity is subject to a condition, we cannot use special_active because the specials are always considered inactive. Additionally, abilities used as weapons possessed or taught by a leadership are also not detectable in [event][filter_attack]. This PR is there to remedy this double problem.


src/game_events/handlers.cpp
```diff
@@ -241,10 +241,31 @@ struct filter_attack : public event_filter {
 	{
 		const unit_map& units = resources::gameboard->units();
 		const auto& loc = first_ ? event_info.loc1 : event_info.loc2;
-		auto unit = units.find(loc);
-		if(unit != units.end() && loc.matches_unit(unit)) {
-			const config& attack = event_info.data.child(first_ ? "first" : "second");
-			return swf_.empty() || matches_special_filter(attack, swf_);
+		const auto& loc_d = first_ ? event_info.loc2 : event_info.loc1;
+		auto unit_a = units.find(loc);
+		assert(unit_a.valid());
```
stevecotton, line 246: Unless you can explain why it will always find a unit, replacing `if(unit != units.end ...)` with an `assert` isn't going to be approved.

Even with an explanation, the defensive code is preferable.

newfrenchy83: line 251 if(unit_a != units.end() && loc.matches_unit(unit_a)) { is here, i not replace but added assert but i can remove that.

newfrenchy83: assert removed

src/game_events/handlers.cpp
```diff
@@ -241,10 +241,29 @@ struct filter_attack : public event_filter {
...
+				auto opp_weapon = std::make_shared<const attack_type>(temp_other_weapon);
+				const_attack_ptr second_attack = opp_weapon->shared_from_this();
+				auto ctx = attack->specials_context(u.shared_from_this(), opp.shared_from_this(), loc, loc_d, first_, second_attack);
+				std::optional<decltype(ctx)> opp_ctx;
+				if(second_attack) {
+					opp_ctx.emplace(second_attack->specials_context(opp.shared_from_this(), u.shared_from_this(), loc_d, loc, !first_, attack));
+				}
+				return swf_.empty() || matches_special_filter(attack, swf_ , loc);
+			} else {
+			auto ctx = attack->specials_context(u.shared_from_this(), loc, first_);
```
stevecotton, line 264: What's this for? It seems to go out of scope before being used.

Also, the indentation on this line is wrong.

src/game_events/conditional_wml.cpp
```diff
@@ -212,16 +212,38 @@ bool conditional_passed(const vconfig& cond)
...
 {
-	if (!cfg) {
+	if (!attack) {
 		WRN_NG << "attempt to filter attack for an event with no attack data.";
 		// better to not execute the event (so the problem is more obvious)
 		return false;
 	}
+	const unit_map& units = resources::gameboard->units();
+	const unit_map::const_iterator att = units.find(a);
+	assert(att.valid());
```
stevecotton, line 224: How do we know this assert won't trigger?

Given that we already found two units in the calling function, why is there an additional units.find() call in this function?

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filters.cfg
```diff
@@ -198,6 +198,25 @@
         # Make sure the attacks hit
         {FORCE_CHANCE_TO_HIT (id=bob) (id=alice) 100 ()}
         {FORCE_CHANCE_TO_HIT (id=alice) (id=bob) 100 ()}
+        [object]
```
stevecotton, line 201: This should be a new unit test, because there are other ways that `event_test_filter_attack` could break. While those would likely make this test fail too, it's worth testing the general functionality first.

One possible bug would be if `[filter_attack]` only matched for attacks with specials. Having two separate tests would catch that, but this single combined test wouldn't.

src/game_events/handlers.cpp
newfrenchy83: i fix that(i think)

src/game_events/conditional_wml.cpp
newfrenchy83: remove that and put const unit in declaration

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filters.cfg
newfrenchy83: added separate test

src/game_events/handlers.cpp
stevecotton: Ok, indentation is fixed and the "goes out of scope" was my misunderstanding (the context is accessible via the weapon, and removes itself when it goes out of scope).

src/game_events/handlers.cpp
```diff
@@ -241,10 +241,29 @@ struct filter_attack : public event_filter {
...
+		if(unit_a != units.end() && loc.matches_unit(unit_a)) {
+			const unit& u = *unit_a;
+			const config& temp_weapon = event_info.data.child(first_ ? "first" : "second");
+			auto weapon = std::make_shared<const attack_type>(temp_weapon);
+			const_attack_ptr attack = weapon->shared_from_this();
+			if(unit_d != units.end() && loc_d.matches_unit(unit_d)) {
+				const unit& opp = *unit_d;
+				const config& temp_other_weapon = event_info.data.child(!first_ ? "first" : "second");
+				auto opp_weapon = std::make_shared<const attack_type>(temp_other_weapon);
+				const_attack_ptr second_attack = opp_weapon->shared_from_this();
```
stevecotton, line 256: The `shared_from_this()` calls are unnecessary - `weapon` and `opp_weapon` are already `shared_ptr` instances, so there's no need for `attack` and `second_attack`.

src/game_events/handlers.cpp
```diff
@@ -241,10 +241,29 @@ struct filter_attack : public event_filter {
...
+				auto ctx = attack->specials_context(u.shared_from_this(), opp.shared_from_this(), loc, loc_d, first_, second_attack);
+				std::optional<decltype(ctx)> opp_ctx;
+				if(second_attack) {
```
stevecotton, line 259: IIUC, second_attack is never false here, so 3 lines can be reduced to just
`auto opp_ctx = opp_weapon->specials_context...`.

newfrenchy83: and if second_unit is Dark Adept and attackis melee, second_attack will be nulll, isn't it?

stevecotton: Ah, I added the edited paragraph just at the same time as you replied. It's not `null`: https://stackoverflow.com/questions/61095800/is-it-possible-that-make-shared-has-no-any-exception-but-returns-a-nullptr

stevecotton: No need for std::optional.

gfgtdf:
Seeing this, i really wonder why the event info stores the attack info as a config (i know that that's not your "fault" and shoudn't stop you from doing this). I mean, why can't it store the index or the id of the attack and then use the units attack object directly (which i think already has the correct battle context set)? I think using the units attack object directly would probably make the handling attack contexts like this easier. Note that the attack.cpp code also uses a mix of the index and the weapons id (id in code = the name attribute in wml) to identify the weapons not some other config object

src/game_events/handlers.cpp
newfrenchy83: i prefer what it remain foruse emplace( instead of auto opp_ctx = opp_weapon->specials_context....

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filters.cfg
```diff
@@ -275,6 +275,84 @@
...
+        name=attack
+        [filter_attack]
+            special_id_active=ability_poison_blade
+        [/filter_attack]
+        {ASSERT ({VARIABLE_CONDITIONAL side_number equals 2})}
+        {VARIABLE_OP triggers add 1}
+    [/event]
+    [event]
+        name=turn 2
+        {RETURN ({VARIABLE_CONDITIONAL triggers equals 1})}
```
stevecotton, line 352: I suggest checking the filter both when attacking and when defending:

```
@@ -339,17 +339,29 @@
         [/do_command]
         [end_turn][/end_turn]
     [/event]
+
+    [event]
+        name=attack
+        [filter_second_attack]
+            special_id_active=ability_poison_blade
+        [/filter_second_attack]
+        {ASSERT ({VARIABLE_CONDITIONAL side_number equals 1})}
+        {VARIABLE_OP triggers_on_defense add 1}
+    [/event]
     [event]
         name=attack
         [filter_attack]
             special_id_active=ability_poison_blade
         [/filter_attack]
         {ASSERT ({VARIABLE_CONDITIONAL side_number equals 2})}
-        {VARIABLE_OP triggers add 1}
+        {VARIABLE_OP triggers_on_attack add 1}
     [/event]
+
     [event]
         name=turn 2
-        {RETURN ({VARIABLE_CONDITIONAL triggers equals 1})}
+        {ASSERT ({VARIABLE_CONDITIONAL triggers_on_attack equals 1})}
+        {ASSERT ({VARIABLE_CONDITIONAL triggers_on_defense equals 1})}
+        {SUCCEED}
     [/event]
```

src/game_events/handlers.cpp
stevecotton: Did some testing, and it seems the `opp_ctx` isn't needed at all. Same results with and without the extra bit of test that I've added in a separate comment.

* The test passes when the `auto ctx =` line is there, even without `opp_ctx`
* It fails when `ctx` is removed, even if the `opp_ctx` line is still there.

I tested with these two ways of declaring `opp_ctx` with `decltype(ctx)`:
```
std::optional<attack_type::specials_context_t> opp_ctx;
// followed by the unchanged opp_ctx.emplace(...) line
```

```
auto opp_ctx = second_attack->specials_context(opp.shared_from_this(), u.shared_from_this(), loc_d, loc, !first_, attack);
```

newfrenchy83: There is one case where opp_ctx is still necessary, and that is when the activity of the special weapon being tested depends on the presence and activity of another special weapon held by the opponent. By testing the absence of opp_ctx in this configuration, nothing happens, the declaration of opp_ctx is changed, okay, I detect no problem, but this variable remains necessary and should not be removed.

newfrenchy83: declaration changed, but opp_ctx remain.

stevecotton: I'm wondering if that's handled by the `second_attack` argument in
```
auto ctx = attack->specials_context(u.shared_from_this(), opp.shared_from_this(), loc, loc_d, first_, second_attack);
```
instead of relying on `opp_ctx`.

newfrenchy83: if opp_ctx unused, then second_attack has const config data only, for what fix will be perfect, both weapon must use ctx and opp_ctx.

newfrenchy83:
@stevecotton i added another test where special tested active only if opponent has specific special active. If test check without op_ctx in hardcoding, test fail.

newfrenchy83:
@stevecotton don't forget this PR

src/game_events/conditional_wml.hpp
```diff
@@ -24,11 +24,12 @@
 class vconfig;
 
 #include "config.hpp"
+#include "units/unit.hpp"
 
 namespace game_events
 {
 	bool conditional_passed(const vconfig& cond);
-	bool matches_special_filter(optional_const_config cfg, const vconfig& filter);
+	bool matches_special_filter(const_attack_ptr attack, const vconfig& filter);
```
stevecotton, line 32: I think this function could be removed, with the lines in handlers.cpp changed to directly `return swf_.empty() || attack->matches_filter(swf_.get_parsed_config());`

From the implementation of the function:
* The `if (!attack)` block is no longer necessary, as we know that `std::make_shared` doesn't return `nullptr`.
* The comment about putting `auto attack` on the heap might still be useful, but belongs on the line before `attack = std::make_shared`, because that's what that comment refers to.
* After removing those, there's only one line left in `matches_special_filter`.

src/game_events/handlers.cpp
```diff
@@ -241,10 +241,24 @@ struct filter_attack : public event_filter {
...
+		if(unit_a != units.end() && loc.matches_unit(unit_a)) {
+			const unit& u = *unit_a;
+			auto temp_weapon = event_info.data.optional_child(first_ ? "first" : "second");
+			const_attack_ptr attack = std::make_shared<const attack_type>(*temp_weapon);
+			if(unit_d != units.end() && loc_d.matches_unit(unit_d)) {
+				const unit& opp = *unit_d;
+				auto temp_other_weapon = event_info.data.optional_child(!first_ ? "first" : "second");
+				const_attack_ptr second_attack = std::make_shared<const attack_type>(*temp_other_weapon);
+				auto ctx = attack->specials_context(u.shared_from_this(), opp.shared_from_this(), loc, loc_d, first_, second_attack);
+				auto opp_ctx = second_attack->specials_context(opp.shared_from_this(), u.shared_from_this(), loc_d, loc, !first_, attack);
```
stevecotton, line 256: By putting the `shared_from_this()` on the lines declaring `u` and `opp`
```
				const auto opp = unit_d->shared_from_this();
```
then this line becomes the more readable
```
				auto opp_ctx = second_attack->specials_context(opp, u, loc_d, loc, !first_, attack);
```

This one's an optional change - I'd be happy to merge without it.

newfrenchy83: doone, but wait if check complete without fail

stevecotton:
Thanks for the fix.

src/game_events/handlers.cpp
```diff
@@ -241,10 +242,24 @@ struct filter_attack : public event_filter {
...
-		if(unit != units.end() && loc.matches_unit(unit)) {
-			auto attack = event_info.data.optional_child(first_ ? "first" : "second");
-			return swf_.empty() || matches_special_filter(attack, swf_);
+		const auto& loc_d = first_ ? event_info.loc2 : event_info.loc1;
+		auto unit_a = units.find(loc);
+		auto unit_d = units.find(loc_d);
+		if(unit_a != units.end() && loc.matches_unit(unit_a)) {
+			const auto u = unit_a->shared_from_this();
+			auto temp_weapon = event_info.data.optional_child(first_ ? "first" : "second");
+			const_attack_ptr attack = std::make_shared<const attack_type>(*temp_weapon);
```
CelticMinstrel, line 251: This is wrong. Either use `mandatory_child` or check if the child actually exists. I'm not sure which is correct here, but using `optional_child` and then assuming it exists is definitely not correct.

src/game_events/handlers.cpp
```diff
@@ -241,10 +242,24 @@ struct filter_attack : public event_filter {
...
+			if(unit_d != units.end() && loc_d.matches_unit(unit_d)) {
+				const auto opp = unit_d->shared_from_this();
+				auto temp_other_weapon = event_info.data.optional_child(!first_ ? "first" : "second");
+				const_attack_ptr second_attack = std::make_shared<const attack_type>(*temp_other_weapon);
```
CelticMinstrel, line 255: Same as above.

CelticMinstrel:
> Thanks for the fix.

The fix is nice, I just wish it wasn't wrong.

src/game_events/handlers.cpp
stevecotton: Oh, didn't notice the conflict with #7497.

PR #7510, opened 2023-04-02
Correctly handle no game history data existing.

### LINKED ISSUE #7489: MP Lobby: Assertion error in Match History tabs

I understand Match History was only implemented in 1.17 so this isn't applicable for 1.16. Since I don't play MP, I assume this is why have no history. However, attempting to use the history function results in an unresponsive UI while the client attempts to find data that isn't there. This is confirmed by log messages:

> error network: Timed out waiting for history data, returning nothing

Time-out is supposed to be only 5 seconds, but it still feels strange to have the UI do nothing (and block redraws). Perhaps attempt the network transfer asynchronously - maybe 5 seconds is too short an interval to bother with showing a progress bar or any such feedback.

Additionally, the history dialogue should probably not be shown at all if no history data is found - display some text to the user instead, something simple like 'no history found'.

Pentarctagon:
Fixes #7489

---

@Wedge009 I'm not able to reproduce the crash you mentioned in that issue. Could you get a stack trace for it?

### DIFF
--- a/src/server/common/resultsets/game_history.cpp
+++ b/src/server/common/resultsets/game_history.cpp
@@ -52,6 +52,13 @@ void game_history::read(mariadb::result_set_ref rslt)
 std::unique_ptr<simple_wml::document> game_history::to_doc()
 {
     auto doc = std::make_unique<simple_wml::document>();
+
+    if(results.size() == 0)
+    {
+        doc->set_attr_dup("no_data", "yes");
+        return doc;
+    }
+
     for(const auto& result : results)
     {
         simple_wml::node& ghr = doc->root().add_child("game_history_result");
--- a/src/gui/dialogs/multiplayer/match_history.cpp
+++ b/src/gui/dialogs/multiplayer/match_history.cpp
@@ -20,6 +20,7 @@
 #include "gettext.hpp"
 #include "filesystem.hpp"
 #include "gui/auxiliary/find_widget.hpp"
+#include "gui/dialogs/message.hpp"
 #include "gui/widgets/button.hpp"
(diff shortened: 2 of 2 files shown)


Pentarctagon:
Alright, the crash has been fixed as well now. Barring anything else being said, I plan to merge this in the next day or two.

Wedge009:
Thanks, sounds good to me.

src/server/common/resultsets/game_history.cpp
```diff
@@ -52,6 +52,13 @@ void game_history::read(mariadb::result_set_ref rslt)
 std::unique_ptr<simple_wml::document> game_history::to_doc()
 {
     auto doc = std::make_unique<simple_wml::document>();
+
+    if(results.size() == 0)
+    {
+        doc->set_attr_dup("no_data", "yes");
```
soliton-, line 58: Why is this necessary? Just no `[game_history_result]` means the same thing, no? If more structure is needed then the existing WML should be wrapped in `[game_history_results]` or so. (Certainly the way just an error attribute is sent right now is not so great.)

Pentarctagon: If an unrelated response is sent from the server to the client, then it would also have no [game_history_result].

soliton-: If an unrelated response contains a `no_data` attribute then you have a similar issue. Thus the suggestion to always send a tag like `[game_history_results]`.

Pentarctagon: I will update this then once the PR for storing a leader's information in the database has been resolved.

PR #7512, opened 2023-04-04
Add support for [effect] apply_to=level

Elvish-Hunter:
The title here is pretty self-explanatory.

Regarding use cases, this can be useful every time a unit needs to increase its level in an AMLA: for example, I have two units like this in TSoG and until now I had to use variations as a workaround for the lack of this feature. In an RPG-style add-on, this can allow a unit to level up (or down, if you wish so) as much as you want.

Implementation notes:

* there are two keys to change the unit's level, which are `increase=` and `set=`;
* negative levels are correctly supported;
* percentages aren't supported. Usually levels are small numbers so it doesn't make much sense to use percentages, but I can add them if you want;
* using `apply_to=level` and `times=per level` at the same time is disallowed to avoid confusion (would that apply to the levels before this effect is applied, or would we have a recursion?); if such combination is used, the engine emits a warning (not sure if `ERR_UT` would be better than `WRN_UT`) and uses 1 as default value.

### DIFF
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -1907,7 +1907,7 @@ std::size_t unit::modification_count(const std::string& mod_type, const std::str
 
 const std::set<std::string> unit::builtin_effects {
 	"alignment", "attack", "defense", "ellipse", "experience", "fearless",
-	"halo", "healthy", "hitpoints", "image_mod", "jamming", "jamming_costs",
+	"halo", "healthy", "hitpoints", "image_mod", "jamming", "jamming_costs", "level",
 	"loyal", "max_attacks", "max_experience", "movement", "movement_costs",
 	"new_ability", "new_advancement", "new_animation", "new_attack", "overlay", "profile",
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
I'd've supported writing this in Lua, except for this caveat:

> * using `apply_to=level` and `times=per level` at the same time is disallowed to avoid confusion (would that apply to the levels before this effect is applied, or would we have a recursion?); if such combination is used, the engine emits a warning (not sure if `ERR_UT` would be better than `WRN_UT`) and uses 1 as default value.

So I think this implementation looks good. Due to the above caveat I think it's reasonable to implement this in the engine.

gfgtdf:
Looks good to me. Note that (especially if you want to use this in 1.16) you can also implemented this in lua in your add-on. But for 1.17 it sure makes sense to have this in the engine.

knyghtmare (approved):
I already use this for my add-on so having it in engine is a good addition.

PR #7514, opened 2023-04-08
Add comments to display and display_context, silence 64-to-32 warnings

stevecotton:
Silence the warnings in `viewing_side()` and `playing_side()` with a `static_cast`.

The static casts are for -Wshorten-64-to-32 warnings, it would take a lot of work to fix this in a better way.

### DIFF
--- a/src/display.hpp
+++ b/src/display.hpp
@@ -80,6 +80,9 @@ namespace wb {
 
 class gamemap;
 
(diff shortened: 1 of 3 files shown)


Vultraz (approved):
In terms of comments this is fine, though I agree the API desperately needs cleaning up.

PR #7515, opened 2023-04-09
Add achievements to DiD

nemaara:
Like the title says. I playtested some but not others (will do that soon). It will help if others can confirm if they work as well.

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/achievements.cfg
+++ b/data/campaigns/Descent_Into_Darkness/achievements.cfg
@@ -1,12 +1,60 @@
 #textdomain wesnoth-did
+
+# achievements in DiD generally should be based around exploration and trying funny things in the campaign
+# these achievements shouldn't be so much about battle challenges, I want to give players another incentive to explore the little details
 [achievement_group]
     display_name=_"Descent into Darkness"
     content_for=descent_into_darkness
     [achievement]
         id = "did_rat_eater"
-        name = _"Rat Eater"
-        description = _"Have a Ghoul eat way too many rats during Descent into Darkness's A Haunting in Winter."
-        icon="data/core/images/icons/potion_red_small.png"
+        name = _"A Big Hunk of Flesh"
+        #po: the achievements are somewhat written as if from Malin's point of view which is why there's some weird and sometimes quirky flavor in them
+        description = _"Feed a Ghoul way too many rats."
+        icon="data/core/images/attacks/fangs-animal.png"
+        icon_completed="data/core/images/attacks/fangs-animal.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-8-acid.png",0,0)"
+    [/achievement]
+    [achievement]
+        id = "did_drogan"
+        #po: you can interpret this as a Harry Potter reference or you can think about it as Malin making fun of Drogan's country accent, or you can take artistic liberty with the translation
+        name = _"Yer a Ghoul, Drogan"
+        description = _"Make that arrogant jerk Drogan into just another one of your minions."
+        icon="data/core/images/attacks/staff-plague.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-8-acid.png",0,0)"
+        hidden = yes
+    [/achievement]
+    [achievement]
+        id = "did_sushi"
+        name = _"High Grade Sushi"
+        description = _"Upgrade a normal swimmer zombie into a super zombie."
+        icon="data/core/images/attacks/tail-merman.png"
+        icon_completed="data/core/images/attacks/tail-merman.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-8-grey.png",0,0)"
+    [/achievement]
+    [achievement]
+        id = "did_drunk"
+        name = _"Drunken Fist Style"
+        description = _"Make the drunken sailor zombie drink some poorly aged rum."
+        icon="data/core/images/attacks/fist.png"
+        icon_completed="data/core/images/attacks/fist.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-8-grey.png",0,0)"
+    [/achievement]
+    [achievement]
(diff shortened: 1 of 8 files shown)


knyghtmare (approved):
Mostly okay, but two inquiries:

1. Should they all be hidden? 
2. Sholuldn't the Monolith one be a progress counter (0/8)?

nemaara:
Okay good point let me unhide a couple and figure out how to make it a progress bar.

knyghtmare:
> Okay good point let me unhide a couple and figure out how to make it a progress bar.

while you're at it, add better looking icon frames. We have added them in earlier. :p

nemaara:
All set and I playtested them all. Waiting a few days for feedback before merging.

CelticMinstrel:
I think your way of doing the monolith achievement is a bit strange. Wouldn't it be simpler to just set `limit=$monolith` and do away with the tests on the monolith variable in scenario 10, just unconditionally calling `[progress_achievement]`? I think that would even give the same effect from the player perspective, with less complicated code.

It might mean the `VARIABLE_OP` needs to go before `[progress_achievement]` instead of after.

data/campaigns/Descent_Into_Darkness/achievements.cfg
```diff
@@ -1,12 +1,62 @@
...
+        description = _"Feed a Ghoul way too many rats."
+        icon="data/core/images/attacks/fangs-animal.png"
+        icon_completed="data/core/images/attacks/fangs-animal.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-8-acid.png",0,0)"
+    [/achievement]
+    [achievement]
+        id = "did_drogan"
+        name = _"Yer a Ghoul, Drogan"
+        description = _"Make that arrogant jerk Drogan into just another one of your minions."
+        icon="data/core/images/attacks/staff-plague.png"
+        icon_completed="data/core/images/attacks/staff-plague.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-8-acid.png",0,0)"
```
CelticMinstrel, line 21: I believe `icon=` is unused for hidden achievements so you might as well just remove it. (Or you can set `icon=` to the current value of `icon_completed=` and remove `icon_completed=` – that should give the same effect, but I think it's up to you which seems better.)

This applies to all the hidden achievements.

nemaara:
It might be but this is just the way I thought of doing it to be safe. I could blindly put in your code suggestion though if you're sure it'll work.

data/campaigns/Descent_Into_Darkness/achievements.cfg
nemaara: Okay.

CelticMinstrel:
I did do it that way for an achievement in my campaign, and I remember the achievement working. Unfortunately, I don't remember if I remember it working after I switched it to that method or if I only remember it working before `[progress_achievement]` existed.

So, I'm pretty sure it'll work, but not 100% sure.

nemaara:
Huh actually progress_achievement's behavior is kind of bad because you can just reload to get the same event and get however much progress you need, so it needs variable conditionals anyway.

data/campaigns/Descent_Into_Darkness/achievements.cfg
```diff
@@ -1,12 +1,59 @@
 #textdomain wesnoth-did
+
+# achievements in DiD generally should be based around exploration and trying funny things in the campaign
+# these achievements shouldn't be so much about battle challenges, I want to give players another incentive to explore the little details
+# also the achievements are somewhat written as if from Malin's point of view which is why there's some weird and sometimes quirky flavor in them
```
stevecotton, line 5: Some of this should be a `# po:` hint, otherwise the translators won't see it. Maybe put in to the first achievement.

data/campaigns/Descent_Into_Darkness/achievements.cfg
```diff
@@ -1,12 +1,59 @@
...
-        description = _"Have a Ghoul eat way too many rats during Descent into Darkness's A Haunting in Winter."
-        icon="data/core/images/icons/potion_red_small.png"
+        name = _"A Big Hunk of Flesh"
+        description = _"Feed a Ghoul way too many rats."
+        icon="data/core/images/attacks/fangs-animal.png"
+        icon_completed="data/core/images/attacks/fangs-animal.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-8-acid.png",0,0)"
+    [/achievement]
+    [achievement]
+        id = "did_drogan"
+        name = _"Yer a Ghoul, Drogan"
```
stevecotton, line 18: Should have a `# po:` comment, you can't expect all the translators to spot the reference.

However, I'd prefer not to have pop-culture references in the official campaigns.

data/campaigns/Descent_Into_Darkness/achievements.cfg
nemaara: Okay done.

data/campaigns/Descent_Into_Darkness/achievements.cfg
nemaara: They don't have to translate it as the reference, it's just something for fun as the name of the achievement. I wouldn't mind if someone translating it construing it as Malin making fun of Drogan's uneducated country accent either.

stevecotton: Fidelity is part of the translators' task:
* if a line is obscure, they will spend time understanding it, time that you should save them by adding the `# po:` comment. They aren't normal readers who can just skip past an odd line, they need to stop and work out what it says.
* if you want them to have additional artistic license for translating a specific line, that's something to put in the `# po:` comment.

nemaara: Okay good point I will add a po comment for them.

Elvish-Hunter:
Just to err on the side of caution, I suggest renaming the "Michelin Starred Sushi" achievement to a more trademark-friendly "5 Stars Sushi".

PR #7518, opened 2023-04-09
Add achievements to Liberty

nemaara:
Like the title says. Giving a few days for feedback on this one too.

### DIFF
--- a/data/campaigns/Liberty/achievements.cfg
+++ b/data/campaigns/Liberty/achievements.cfg
@@ -1,1 +1,41 @@
 #textdomain wesnoth-l
+[achievement_group]
+    display_name=_"Liberty"
+    content_for=liberty
+    [achievement]
+        id="liberty_or_death"
+        name=_"Give Me Liberty or Give Me Death!"
+        description=_"Complete Civil Disobedience with at least 3 peasants remaining."
+        icon="data/core/images/attacks/pitchfork.png"
+        icon_completed="data/core/images/attacks/pitchfork.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-3-red.png",0,0)"
+    [/achievement]
+    [achievement]
+        id="liberty_nameless"
+        name=_"The Nameless General"
+        description=_"Find out the name of the general in Unlawful Orders."
+        icon="data/core/images/icons/helmet_chain-coif.png"
+        icon_completed="data/core/images/icons/helmet_chain-coif.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-8-grey.png",0,0)"
+    [/achievement]
+    [achievement]
+        id="liberty_iron"
+        name=_"Harper vs. Goliath"
+        description=_"Defeat an Iron Mauler with Harper in Hide and Seek."
+        icon="data/core/images/attacks/slingshot.png"
+        icon_completed="data/core/images/attacks/slingshot.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-3-red.png",0,0)"
+    [/achievement]
+    [achievement]
+        id="liberty_hunter"
+        #po: this is the achievement for the scenario The Hunters
+        name=_"Perfect Hunter"
+        description=_"Complete The Hunters with no losses"
+        icon="data/core/images/attacks/blowgun.png"
+        icon_completed="data/core/images/attacks/blowgun.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-8-grey.png",0,0)"
+    [/achievement]
+    [achievement]
+        id="liberty_army"
+        name=_"Force of Arms is Pointless Without Strategy"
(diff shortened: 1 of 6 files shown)


data/campaigns/Liberty/scenarios/04_Unlawful_Orders.cfg
```diff
@@ -823,6 +824,24 @@
...
+        [if]
+            [have_location]
+                time_of_day=lawful
+            [/have_location]
+            [then]
+                [modify_unit]
+                    [filter]
+                        id=Kestrel
+                    [/filter]
+                    name= _ "General Kestrel"
```
CelticMinstrel, line 836: Does this actually do anything? Is the name even visible at all? The unit is killed right afterwards, and he doesn't say anything else after this.

CelticMinstrel:
I think there's something wrong with your way of making units nameless.

I'm pretty sure you don't have to set the name to a space to make it have no name? You can instead leave name empty (or absent) and set `generate_name=no`.

With the `NAMED_LOYAL_UNIT` you have a tag amendment, so you can add `generate_name=no` there as well. And in the recruit event I'm pretty sure `name=` or `name=""` will do what you want without causing the schema validator to complain.

Also, while we're on the subject of the nameless units… you added an event to reveal Kestrel's name, but not Rothel's? Also, regardless of that, Rothel is no longer an `NAMED_LOYAL_UNIT` but just a `LOYAL_UNIT`.

data/campaigns/Liberty/scenarios/04_Unlawful_Orders.cfg
nemaara: ??? How did this block of code end up there, it was supposed to be split in half before when he talked and then after for the achievement. Awkward.

nemaara:
I tried the name as both an empty key and with no space and both ended up generating a name so unless you know of a way to make the AI recruit nameless units I can't figure out how else to do it.

CelticMinstrel:
Um, so there are two separate issues here, right?

The first is Kestrel and Rothel. I'm pretty sure they should just work with `generate_name=no`?

The second is the recruit event. I think setting the name to blank _should_ work, but… there's the implementation detail that `[modify_unit]` is (unfortunately) a store+unstore internally, so maybe also adding `generate_name=no` there would make it work?

data/campaigns/Liberty/scenarios/05_Hide_and_Seek.cfg
```diff
@@ -586,6 +586,30 @@ Been a tough few days since we left Elensefar with all them patrols running arou
...
+        [filter]
+            type=Iron Mauler
+        [/filter]
+        [filter_second]
+            id=Harper
+        [/filter_second]
+
+        [message]
+            speaker=Harper
+            message= _ "Uncle, we did it! We beat one of them iron monsters!"
```
stevecotton, line 600: This should have a `# po:` comment, because the translators won't have the benefit of seeing that it's in an iron mauler's die event.

data/campaigns/Liberty/scenarios/06_The_Hunters.cfg
```diff
@@ -222,6 +222,8 @@ The sun don’t really shine at all in these Grey Woods, but truth be told, I ha
         {GENERIC_UNIT 2 (Shadow Mage) 18 22}
         {GENERIC_UNIT 4 (Heavy Infantryman) 7 2} {GUARDIAN}
         {GENERIC_UNIT 4 (Heavy Infantryman) 7 6} {GUARDIAN}
+
+        {VARIABLE losses false}
```
stevecotton, line 226: Should have a corresponding `CLEAR_VARIABLE`.

nemaara:
Fixed the method for making the units nameless and added suggestions from the other feedback. Anything else?

data/campaigns/Liberty/achievements.cfg
```diff
@@ -1 +1,40 @@
...
+    [achievement]
+        id="liberty_iron"
+        name=_"Harper vs. Goliath"
+        description=_"Defeat an Iron Mauler with Harper in Hide and Seek."
+        icon="data/core/images/attacks/slingshot.png"
+        icon_completed="data/core/images/attacks/slingshot.png~SCALE(72,72)~BLIT("data/core/images/misc/achievement-frames/frame-3-red.png",0,0)"
+    [/achievement]
+    [achievement]
+        id="liberty_hunter"
+        name=_"Perfect Hunter"
```
stevecotton, line 28: A `# po:` hint that this corresponds to the level name would be good, in the translation tool it'll be detached from the `description=` line.

nemaara: Uhm, they're just translating the word hunter? It's not a special word.

stevecotton: They, and that might mean someone else on the same translation team, are also translating the scenario name "The Hunters", the `# po:` comment would be a hint to use similar translations for both.

PR #7521, opened 2023-04-10
Temporarily disable Windows CI builds

stevecotton:
It's better to disable them that to get in the habit of ignoring the CI fails from the following two issues; this can be reverted once either is fixed.

glib has a bug where it builds things out of order which results in it failing to build the dependencies, which fails the build.
https://github.com/microsoft/vcpkg/issues/28722

vcpkg also has a bug where it very often fails to use the cache of the previously built dependencies even if none of them changed, which then makes it rebuild all the dependencies again.
https://github.com/microsoft/vcpkg/issues/26601

If the glib bug was fixed, then the errors wouldn't happen since rebuilding the dependencies wouldn't fail. If the caching bug was fixed, then the errors wouldn't happen since it'd just reuse the cached dependencies.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -243,78 +243,13 @@ jobs:
       - name: Run unit tests
         run: ./run_boost_tests --path=projectfiles/Xcode/build/"$CFG" --executable=unit_tests
 
-  windows:
-    strategy:
-      fail-fast: false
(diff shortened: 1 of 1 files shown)


stevecotton:
Hmm - that's disabled all the builds, not just the Windows ones. Working out what's wrong...

PR #7532, opened 2023-04-13
fix test of presence of attack/second_attack in filter_attack function

newfrenchy83:
resolve problem in https://github.com/wesnoth/wesnoth/pull/7509#discussion_r1165220995discussion

### DIFF
--- a/src/game_events/handlers.cpp
+++ b/src/game_events/handlers.cpp
@@ -248,17 +248,22 @@ struct filter_attack : public event_filter {
 		if(unit_a != units.end() && loc.matches_unit(unit_a)) {
 			const auto u = unit_a->shared_from_this();
 			auto temp_weapon = event_info.data.optional_child(first_ ? "first" : "second");
-			const_attack_ptr attack = std::make_shared<const attack_type>(*temp_weapon);
-			if(unit_d != units.end() && loc_d.matches_unit(unit_d)) {
-				const auto opp = unit_d->shared_from_this();
-				auto temp_other_weapon = event_info.data.optional_child(!first_ ? "first" : "second");
-				const_attack_ptr second_attack = std::make_shared<const attack_type>(*temp_other_weapon);
-				auto ctx = attack->specials_context(u, opp, loc, loc_d, first_, second_attack);
-				auto opp_ctx = second_attack->specials_context(opp, u, loc_d, loc, !first_, attack);
-				return swf_.empty() || attack->matches_filter(swf_.get_parsed_config());
-			} else {
-				auto ctx = attack->specials_context(u, loc, first_);
-				return swf_.empty() || attack->matches_filter(swf_.get_parsed_config());
+			if(temp_weapon){
+				const_attack_ptr attack = std::make_shared<const attack_type>(*temp_weapon);
+				if(unit_d != units.end() && loc_d.matches_unit(unit_d)) {
+					const auto opp = unit_d->shared_from_this();
+					auto temp_other_weapon = event_info.data.optional_child(!first_ ? "first" : "second");
+					const_attack_ptr second_attack = temp_other_weapon ? std::make_shared<const attack_type>(*temp_other_weapon) : nullptr;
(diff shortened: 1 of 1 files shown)


newfrenchy83:
@stevecotton @CelticMinstrel i hope what this fix is conveniant for bothof you, ah, i also repulll old declaration for op_ctx because abuse of auto declaration is not readable.

src/game_events/handlers.cpp
```diff
@@ -247,18 +247,23 @@ struct filter_attack : public event_filter {
...
-				const_attack_ptr second_attack = std::make_shared<const attack_type>(*temp_other_weapon);
-				auto ctx = attack->specials_context(u, opp, loc, loc_d, first_, second_attack);
-				auto opp_ctx = second_attack->specials_context(opp, u, loc_d, loc, !first_, attack);
-				return swf_.empty() || attack->matches_filter(swf_.get_parsed_config());
-			} else {
-				auto ctx = attack->specials_context(u, loc, first_);
-				return swf_.empty() || attack->matches_filter(swf_.get_parsed_config());
+			auto temp_weapon = event_info.data.mandatory_child(first_ ? "first" : "second");
+			const_attack_ptr attack = std::make_shared<const attack_type>(temp_weapon);
+			if(attack){
```
stevecotton, line 252: `attack` can never be `nullptr` here, so this seems to be testing the wrong condition.

src/game_events/handlers.cpp
```diff
@@ -247,18 +247,23 @@ struct filter_attack : public event_filter {
...
+				if(unit_d != units.end() && loc_d.matches_unit(unit_d)) {
+					const auto opp = unit_d->shared_from_this();
+					auto temp_other_weapon = event_info.data.mandatory_child(!first_ ? "first" : "second");
```
stevecotton, line 255: Are "first" and "second" both mandatory, even if the defender doesn't have a weapon? I'm testing that at the moment.

src/game_events/handlers.cpp
```diff
@@ -247,18 +247,23 @@ struct filter_attack : public event_filter {
...
+					const_attack_ptr second_attack = std::make_shared<const attack_type>(temp_other_weapon);
+					auto ctx = attack->specials_context(u, opp, loc, loc_d, first_, second_attack);
+					std::optional<decltype(ctx)> opp_ctx;
```
stevecotton, line 258: Seems less readable than `auto opp_ctx =` to me.

src/game_events/handlers.cpp
CelticMinstrel: And you're using `mandatory_child` which means `temp_weapon` will never be null either. (If the child doesn't exist, an exception is thrown.)

CelticMinstrel: By the way, does this crash if you put `[filter_attack]` in a moveto event (for example)?

src/game_events/handlers.cpp
newfrenchy83: what you suggest, in the end?

src/game_events/handlers.cpp
newfrenchy83: must i remove if(attack) loop or could i let like it?

src/game_events/handlers.cpp
newfrenchy83: reuse optionnal_child and added check attack/second_atack

src/game_events/handlers.cpp
CelticMinstrel: I don't think `auto opp_ctx` works? That requires you to initialize it immediately, but the whole point is that we may or may not want to initialize it.

CelticMinstrel:
I think this fix looks fine.

newfrenchy83:
> I think this fix looks fine.

Ok

src/game_events/handlers.cpp
stevecotton: Tested, and it now works okay even in a moveto event.

PR #7534, opened 2023-04-14
More unit tests for attack_filter

stevecotton:
Add `first_time_only=no` to some existing tests, and verify that events that should only trigger once do trigger exactly once.

A new test of what happens when a unit with only a ranged weapon fights a unit with only a melee weapon. That's the "no defense" test.

A new test of `[filter_attack]` when an event lacks attack data. In testing, I found that it used to trigger a warning that I thought was useless code, and had removed during 88439d6427a81a6b131acd8b0f7bf04b2d981d71's review. I think removing it is still good, at the time that the filter was checked it would have warned "attempt to filter attack for an event with no attack data."

@newfrenchy83 please would you review my new tests?


data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filters.cfg
```diff
@@ -239,42 +238,62 @@
...
     [/event]
     [event]
         name=attack
+        first_time_only=no
         [filter_attack]
             damage=9
         [/filter_attack]
         {ASSERT ({VARIABLE_CONDITIONAL side_number equals 2})}
-        {VARIABLE_OP triggers add 1}
+        {VARIABLE_OP triggerson_on_bob_attack add 1}
```
CelticMinstrel, line 255: Typo

newfrenchy83:
@stevecotton , i understand for moveto test, but for "no defense" test, the test shoudn't fail because alice don't have melee attack?

stevecotton:
For the "no defense" test, I'd be happy if the behavior changed to what you expect. I also expected Alice to have no melee attack, and had to debug what the engine is doing.

My intent for this test is to document what the engine does at the moment. It can change until the branch becomes stable (circa 1.18.0's release), but once the branch becomes stable then the test will check that it doesn't change in 1.18.x.

newfrenchy83:
> For the "no defense" test, I'd be happy if the behavior changed to what you expect. I also expected Alice to have no melee attack, and had to debug what the engine is doing. My intent for this test is to document what the engine does at the moment. It can change until the branch becomes stable (circa 1.18.0's release), but once the branch becomes stable then the test will check that it doesn't change in 1.18.x.

fixed with https://github.com/wesnoth/wesnoth/pull/7535

stevecotton:
Removed the "no defense" test, as it'll be fixed in either #7535 or #7537.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filters.cfg
stevecotton: Good catch, and it pointed to some `RETURN` statements that needed to be changed to `ASSERT` because it was only checking the first one.

stevecotton:
Rebasing to run the tests on the latest version of master.

PR #7535, opened 2023-04-14
fix filter_second_attack matche when_no defense.

newfrenchy83:
if a unit in defense has no attack of range of attacker and under influence of abilitie used like weapon a filter_second_attack can matche it. This PR is a fix for this bug encountered with https://github.com/wesnoth/wesnoth/pull/7534/files#diff-8d03fe35ba9cd63a235a745ba5739e4f37f175915e73324125d3881953968b3c who succed then it must fail

### DIFF
--- a/src/game_events/handlers.cpp
+++ b/src/game_events/handlers.cpp
@@ -249,6 +249,9 @@ struct filter_attack : public event_filter {
 			const auto u = unit_a->shared_from_this();
 			auto temp_weapon = event_info.data.optional_child(first_ ? "first" : "second");
 			if(temp_weapon){
+				if(!first_ && temp_weapon["range"].empty()){
+					return false;
+				}
 				const_attack_ptr attack = std::make_shared<const attack_type>(*temp_weapon);
 				if(unit_d != units.end() && loc_d.matches_unit(unit_d)) {
 					const auto opp = unit_d->shared_from_this();


newfrenchy83:
@CelticMinstrel @stevecotton , could you verify this fix,please?

src/game_events/handlers.cpp
```diff
@@ -249,6 +249,9 @@ struct filter_attack : public event_filter {
 			const auto u = unit_a->shared_from_this();
 			auto temp_weapon = event_info.data.optional_child(first_ ? "first" : "second");
 			if(temp_weapon){
+				if(!first_ && temp_weapon["range"].empty()){
```
CelticMinstrel, line 252: I don't understand. Why checking for empty range?

newfrenchy83: it seem what in second_attack, a weapon empty created when second unit has no attack of same range what attacker. in that case check if range is empty verify what weapon is real or not.

stevecotton: It works, you and I know what it does, but if we had to read this code in 1 month's or 6 months' time, would we still understand what it did? I certainly wouldn't, I'm only finding it readable now because of the context that I know exactly which bug this line is fixing.

I think @gfgtdf's suggestion of reworking what gets passed to this function is good, it makes me more likely to approve a temporary fix, but the golden rule of writing maintainable code is that "temporary fixes" often end up left in the code until another bug comes along, which makes the next debugging task more complex.

I'm sleepy and I'm rambling. I've got an idea for a better fix using a factory method, but will look at that tomorrow. The main part of writing it will be the documentation rather than then code, so please leave it for me to do. Something along the lines of:

```
/** Big detailed doc here, pointing back to the "none" added in b83ccd632e1898f937ea084fd27c62d31d1415b9 */
std::optional<attack_type> get_weapon_from_event(const config& event_data, bool first_or_secord)
{
    ...
    if(it's one of these "none" weapons) {
        return nullopt;
    }
}
```

PR #7537, opened 2023-04-15
Test attack_filter with melee-only vs ranged-only units

stevecotton:
A new test of what happens when a unit with only a ranged weapon fights a unit with only a melee weapon, along with a code change to ignore the placeholder attack that caused the new test to fail.

(The next paragraph is a suggestion for future work, I'm not intending to add it to this PR.)
This only affects the filtering so far, there should probably be another additional test to check what the combat results are. Not added in this commit, as the test here is specific to the code path being fixed; also testing the actual results would need an ability such as the Deep Elves' Stardust that passively affects the amount of damage taken, and doesn't need the defender to hit with a slows ability.

@newfrenchy83 I'm suggesting this instead of #7535, please would you review it?

### DIFF
--- a/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_attack_no_defense.cfg
+++ b/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_attack_no_defense.cfg
@@ -1,0 +1,108 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: [event][filter_attack]
+##
+# Actions:
+# Remove Alice's melee attack.
+# Give Alice an ability that adds a drain special to all of his weapons.
+# Define events that use filter_attack matching Alice's drain special.
+# Have Alice use a ranged attack against Bob, who only has a melee attack.
+# Have Bob use a melee attack against Alice.
+##
+# Expected end state:
+# An event triggers when Alice attacks during side 1's turn.
+# No event triggers when Bob attacks Alice.
+#####
+{GENERIC_UNIT_TEST event_test_filter_attack_no_defense (
+    [event]
+        name=turn 1
+        [object]
+            silent=yes
+            [effect]
+                apply_to=remove_attacks
+                range=melee
+            [/effect]
+            [effect]
+                apply_to=new_ability
+                [abilities]
+                    [drains]
+                        id=test_ability_drain
+                    [/drains]
+                [/abilities]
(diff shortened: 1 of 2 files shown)


newfrenchy83:
it seem fine for me.

newfrenchy83:
I wonder, if we use [filter_second_attack][not]special_id_active but the defender does not have an attack with the same range as the attacker, then must we say that the filter still matches (if we leave placeholder) or not (placeholder removed), if the current configuration says "event triggered because special absent since there is no attack" then try to fix it, it may bring a new bug. it will test that for verify, but i fear what don't fix bug is less bad actually

newfrenchy83:
the problem is for abilitie likeweapon only, if fix, it must be restrains to that.

src/game_events/handlers.cpp
```diff
@@ -236,6 +236,28 @@ struct filter_unit : public event_filter {
...
+ *
+ * This also handles choosing the first or second weapon.
+ *
+ * Though it may seem wasteful to use std::shared_ptr instead of std::optional, it's necessary.
+ * matches_filter() could potentially call a WFL formula, which would call shared_from_this().
+ */
+std::shared_ptr<const attack_type> get_attack_event_weapon(const config& event_data, bool first) {
+	// The placeholders are constructed in attack.cpp's attack::fire_event, and contain only the
+	// name="none", so can be recognised by checking for a range attribute.
+	auto attack_cfg = event_data.optional_child(first ? "first" : "second");
```
newfrenchy83, line 253: best to let placeholder and use that https://github.com/wesnoth/wesnoth/pull/7539 for resolve problem of special_id_active matches in placeholder weapon

src/game_events/handlers.cpp
```diff
@@ -236,6 +236,28 @@ struct filter_unit : public event_filter {
...
+	if(attack_cfg && attack_cfg->has_attribute("range")) {
+		return std::make_shared<const attack_type>(*attack_cfg);
+	}
+	return {};
```
CelticMinstrel, line 257: I think I'd prefer this to be `nullptr` rather than `{}`.

CelticMinstrel:
I think this looks good.

newfrenchy83:
> I think this looks good.

Could you reviewhttps://github.com/wesnoth/wesnoth/pull/7539before to commit this PR?

newfrenchy83:
it seem fine for me

PR #7538, opened 2023-04-15
Stop removing manually-added labels on Github PRs

stevecotton:
Setting sync-labels to "true" means that it removes labels which people have added, but which aren't implied by the filelist.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -9,4 +9,4 @@ jobs:
     - uses: actions/labeler@main
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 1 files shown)


gfgtdf:
Does it also apply to manually removed labels?(when i remove a label that was assigned by the auto labeler)

Pentarctagon (approved):

stevecotton:
> Does it also apply to manually removed labels?(when i remove a label that was assigned by the auto labeler)

Doesn't look like it, the implementation seems to just be "skip over the delete step":
https://github.com/actions/labeler/blob/9471598e3b7ff22b2fa181bd79addf94cb3e0847/src/labeler.ts#L52-L58

PR #7539, opened 2023-04-15
fix ability used like weapon affect placeholder weapon used in defense when not attack of same range what attacker

newfrenchy83:
best fix what https://github.com/wesnoth/wesnoth/pull/7537 or https://github.com/wesnoth/wesnoth/pull/7535 because don't delete placeholder weapon of filter_attack and applied to weapon like ability only.

### DIFF
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1538,8 +1538,10 @@ bool attack_type::has_special_or_ability(const std::string& special, bool specia
 	//Now that filter_(second)attack in event supports special_id/type_active, including abilities used as weapons,
 	//these can be detected even in placeholder attacks generated to compensate for the lack of attack in defense against an attacker using a range attack not possessed by the defender.
 	//It is therefore necessary to check if the range is not empty (proof that the weapon is not a placeholder) to decide if has_weapon_ability can be returned or not.
-	bool real_weapon_ability = (range() != "") ? has_weapon_ability(special, special_id, special_tags) : false;
-	return (has_special(special, false, special_id, special_tags) || real_weapon_ability);
+	if(range().empty()){
+		return false;
+	}
+	return (has_special(special, false, special_id, special_tags) || has_weapon_ability(special, special_id, special_tags));
 }
 //end of emulate weapon special functions.
 
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1535,7 +1535,9 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 
 bool attack_type::has_special_or_ability(const std::string& special, bool special_id, bool special_tags) const
 {
-	return (has_special(special, false, special_id, special_tags) || has_weapon_ability(special, special_id, special_tags));
+	//check if shared_from_this() is real attack or placeholder before to use or not has_weapon_ability function.
+	bool real_weapon_ability = (range() != "") ? has_weapon_ability(special, special_id, special_tags) : false;
+	return (has_special(special, false, special_id, special_tags) || real_weapon_ability);
 }
 //end of emulate weapon special functions.
 
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1535,7 +1535,9 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
 
 bool attack_type::has_special_or_ability(const std::string& special, bool special_id, bool special_tags) const
 {
-	//check if shared_from_this() is real attack or placeholder before to use or not has_weapon_ability function.
+	//Now that filter_(second)attack in event supports special_id/type_active, including abilities used as weapons,
+	//these can be detected even in placeholder attacks generated to compensate for the lack of attack in defense against an attacker using a range attack not possessed by the defender.
+	//It is therefore necessary to check if the range is not empty (proof that the weapon is not a placeholder) to decide if has_weapon_ability can be returned or not.
 	bool real_weapon_ability = (range() != "") ? has_weapon_ability(special, special_id, special_tags) : false;
 	return (has_special(special, false, special_id, special_tags) || real_weapon_ability);
 }


newfrenchy83:
@CelticMinstrel @stevecotton could you review th PR? Thanks

stevecotton:
This is definitely a better fix than my #7537. However, I wouldn't understand the code+comment if I ran into it again in a couple of months time `:-/` . I'll have another go at #7537 based on this.

newfrenchy83:
@stevecotton i try to said What ability like weapon must be checked only if attack is real, without remove placeholder (remove that can source od bugs)

newfrenchy83:
> This is definitely a better fix than my #7537. However, I wouldn't understand the code+comment if I ran into it again in a couple of months time `:-/` . I'll have another go at #7537 based on this.

https://github.com/wesnoth/wesnoth/pull/7537 , not https://github.com/wesnoth/wesnoth/pull/7534/commits ?

newfrenchy83:
@stevecotton , add https://github.com/wesnoth/wesnoth/pull/7539/commits/a157f8546938bcdf759453050f357ef952db2f36 for best explanation

stevecotton:
Thanks, this looks good. Will give @CelticMinstrel a bit of time to comment, but ready to merge.

There's one part of this that I think is a good thing for the development period before 1.18.0, and so I think we should merge as-is, but we might want to optimise once most UMC has been ported to 1.18. This will currently always evaluate both `has_weapon_ability` and `has_special`, which could detect UMC abilities containing formulas that throw WML exceptions; reordering the code could skip one of those calls.

newfrenchy83:
> 

what should understand? What it should create a special_id/type_active for abilities only and special_id/type_active for specials only?

stevecotton:
> what should understand?

C++ has "short-circuit optimisation". With a line such as
```
return (has_special(...) || has_weapon_ability(...));
```
it works out the value of the left-hand side of the `||` first, and if `has_special` returns `true` then it doesn't call `has_weapon_ability` at all. The fixed code always calls both functions, so it will be a little bit slower, but it has the good side-effect of catching edge-cases in UMC code.

newfrenchy83:
> > what should understand?
> 
> C++ has "short-circuit optimisation". With a line such as
> 
> ```
> return (has_special(...) || has_weapon_ability(...));
> ```
> 
> it works out the value of the left-hand side of the `||` first, and if `has_special` returns `true` then it doesn't call `has_weapon_ability` at all. The fixed code always calls both functions, so it will be a little bit slower, but it has the good side-effect of catching edge-cases in UMC code.

this is boolean function, not list function, what one or both function in same time don't important because value true will be returned, for call poison or slow specials, it is not important because bool function, for special_id_active checking, i can speparate filter in two: special_id/type_active who call has_special and weapon_ability_id/type_active who call has_weapon_ability.

stevecotton:
I prefer e5bbce57042 to 3f67ce0d998 at this stage of the 1.18 development cycle.

> this is boolean function, not list function, what one or both function in same time don't important because value true will be returned

It's not about the return type, it's about what happens internally in those functions. They may loop over adjacent units and check which specials are active; it makes them slow but it also means that edge-cases in UMC formulas will be checked. That's a trade-off which is good to have the extra checking when UMC authors port their add-ons to 1.18, but then optimise after the main UMC porting phase is done.

src/units/abilities.cpp
```diff
@@ -1535,7 +1535,17 @@ bool attack_type::has_weapon_ability(const std::string& special, bool special_id
...
+		return true;
+	}
+	//Now that filter_(second)attack in event supports special_id/type_active, including abilities used as weapons,
+	//these can be detected even in placeholder attacks generated to compensate for the lack of attack in defense against an attacker using a range attack not possessed by the defender.
+	//It is therefore necessary to check if the range is not empty (proof that the weapon is not a placeholder) to decide if has_weapon_ability can be returned or not.
+	bool real_weapon_ability = (range() != "") ? has_weapon_ability(special, special_id, special_tags) : false;
+	if(real_weapon_ability){
+		return true;
+	}
+	return false;
```
stevecotton, line 1548: No need for a `real_weapon_ability` variable. This can be written as:

```
	if(range().empty()){
		return false;
	}
	return has_weapon_ability(special, special_id, special_tags);
```

newfrenchy83: done in https://github.com/wesnoth/wesnoth/pull/7539/commits/5126b88e5d8a4c781ccaa1f7b4ff51aaf3348738

stevecotton: I'm confused - I thought that the `has_special(...)` would need to be before the `if(range().empty())`, because abilities might give extra defense even when the unit doesn't have a weapon to counterattack with.

newfrenchy83: Fixed

newfrenchy83:
@stevecotton special_id_active is called by the weapon the special is assigned to, not the weapon the special is encoded in. If for example I use 'charge' against a Dark Adept with no melee attack, and I encode [filter_second_attack]special_id_active=charge for an event, it will be triggered if has_special called even if range is empty. That's why I wanted checkrange before everything else. After all a placeholder can't have an encoder special, and an enemy's special with apply_to=opponent only applies to real weapons (what's the point of using formation to lower the chance_to_hit of a non-existent weapon?)

stevecotton:
I'm confused - I think this is ready to merge, should I merge it now?

What happens if:
* attacker has a normal ranged weapon
* defender has only melee attacks, but has formation and has another unit with formation next to it?

I think that should trigger the ability, even though the defender doesn't have a weapon (assume it has a shield that it can't attack with).

newfrenchy83:
@stevecotton range.empty anew placed before reamain, else special applied to opponent can call [filter_second_attack] same if defender no attack of same range. Now PR can be merged.

newfrenchy83:
@stevecotton if you want trigger a event with formation by owner, using [filter(_second)]ability_id_active=formation

PR #7540, opened 2023-04-16
Add macros support to wmlxgettext

Slayer95:
e.g.

```c++
#textdomain wesnoth-test

#define GREET MODE WHOM
[{MODE}]
    {MODE} = _ "Hello, {WHOM}."
[/{MODE}]
#enddef

{GREET message world}

#define MOODY_GREET MODE WHOM
#arg QUALIFIER
meh#endarg
   {GREET {MODE} (very {QUALIFIER} {WHOM})}
#enddef

{MOODY_GREET message world QUALIFIER=good}
{MOODY_GREET message world QUALIFIER=bad}
{MOODY_GREET message world}
```

Outputs:
```
#: test/test.cfg:5
msgid "Hello, world."
msgstr ""

#: test/test.cfg:5
msgid "Hello, very good world."
msgstr ""

#: test/test.cfg:5
msgid "Hello, very bad world."
msgstr ""

#: test/test.cfg:5
msgid "Hello, very meh world."
msgstr ""
```

### DIFF
--- a/data/tools/pywmlx/nodemanip.py
+++ b/data/tools/pywmlx/nodemanip.py
@@ -23,14 +23,18 @@ def _closenode_update_dict(podict):
                 # be ignored, then wrong context info would be attached to these sentences,
                 # or to an undetermined future sentence.
                 wmlerr(*unbalanced_wml)
+
+            # Get the relevant catalog d for this sentence. Initialize it if needed.
             if i.domain in podict:
                 d = podict[i.domain]
             else:
                 podict[i.domain] = dict()
                 d = podict[i.domain]
-            posentence = d.get(i.sentence)
+
+            dict_key = i.sentence_id
+            posentence = d.get(dict_key)
             if posentence is None:
-                d[i.sentence] = (
+                d[dict_key] = (
                        nodes[-1].nodesentence_to_posentence(i) )
             else:
                 posentence.update_with_commented_string(
--- a/data/tools/pywmlx/nodemanip.py
+++ b/data/tools/pywmlx/nodemanip.py
@@ -119,13 +119,13 @@ def closenode(closetag, mydict, lineno):
             nodes = None
 
 
-def addNodeSentence(sentence, *, domain, ismultiline, lineno, lineno_sub,
-                    override, addition, plural=None):
+def addNodeSentence(sentence, *, domain, macro=None, ismultiline,
+                    lineno, lineno_sub, override, addition, plural=None):
     global nodes
     if nodes is None:
         nodes = [pos.WmlNode(fileref=fileref, fileno=fileno,
                               tagname="", autowml=False)]
-    nodes[-1].add_sentence(sentence, domain=domain, ismultiline=ismultiline,
+    nodes[-1].add_sentence(sentence, domain=domain, macro=macro, ismultiline=ismultiline,
                            lineno=lineno, lineno_sub=lineno_sub,
                            override=override, addition=addition,
                            plural=plural)
--- a/data/tools/pywmlx/postring.py
+++ b/data/tools/pywmlx/postring.py
@@ -84,8 +84,13 @@ class WmlNodeSentence:
     def __init__(self, sentence, *, domain, macro=None, ismultiline,
                  lineno, lineno_sub=0, plural=None,
                  override=None, addition=None):
-        self.sentence = sentence
         self.domain = domain
+        self.sentence = sentence
+        self.sentence_id = sentence
+        if macro is None:
+            self.sentence_id = '\x01' + sentence
+        else:
+            self.sentence_id = '\x00' + '\x00'.join(map(str, macro)) + '\x00\x00' + sentence
         # Say if it is multiline or not.
         self.ismultiline = ismultiline
         self.macro = macro
--- a/data/tools/pywmlx/postring.py
+++ b/data/tools/pywmlx/postring.py
@@ -15,12 +15,13 @@ class PoCommentedStringPL:
 
 class PoCommentedString:
     def __init__(self, sentence, domain, *, orderid, ismultiline,
-                 wmlinfos, finfos, addedinfos, plural=None):
+                 wmlinfos, finfos, macro = None, addedinfos, plural=None):
         self.sentence = sentence
         self.domain = domain
         self.wmlinfos = wmlinfos
         self.addedinfos = addedinfos
         self.finfos = finfos
+        self.macro = macro
         self.orderid = orderid
         self.ismultiline = ismultiline
         self.plural = None
@@ -80,12 +81,14 @@ class PoCommentedString:
 
 # WmlNodeSentence use PoCommentedStringPL for 'plural' parameter
 class WmlNodeSentence:
-    def __init__(self, sentence, *, domain, ismultiline, lineno, lineno_sub=0,
-                 plural=None, override=None, addition=None):
+    def __init__(self, sentence, *, domain, macro=None, ismultiline,
+                 lineno, lineno_sub=0, plural=None,
+                 override=None, addition=None):
         self.sentence = sentence
         self.domain = domain
         # Say if it is multiline or not.
         self.ismultiline = ismultiline
+        self.macro = macro
         self.lineno = lineno
         # lineno_sub:
         # used only in WmlNodeSentence. This parameter is actually used
@@ -128,8 +131,8 @@ class WmlNode:
         self.wmlinfos = None
         self.autowml = autowml
 
-    def add_sentence(self, sentence, *, domain, ismultiline, lineno,
-                     lineno_sub=0, plural=None, override=None, addition=None):
+    def add_sentence(self, sentence, *, domain, macro=None, ismultiline, lineno, lineno_sub=0,
+                     plural=None, override=None, addition=None):
         if self.sentences is None:
             self.sentences = []
         # 'plural' parameter accepted by WmlNode.add_sentence can be:
@@ -157,6 +160,7 @@ class WmlNode:
                 plural = None
         self.sentences.append( WmlNodeSentence(sentence,
                                           domain=domain,
+                                          macro=macro,
                                           ismultiline=ismultiline,
                                           lineno=lineno,
                                           lineno_sub=lineno_sub,
@@ -208,6 +212,7 @@ class WmlNode:
                                wmlinfos=[],
                                finfos=[self.fileref +
                                         str(nodesentence.lineno)],
+                               macro=nodesentence.macro,
                                addedinfos=nodesentence.addedinfo,
                                plural=nodesentence.plural )
                 else:
@@ -218,6 +223,7 @@ class WmlNode:
                                wmlinfos=[],
                                finfos=[self.fileref +
                                         str(nodesentence.lineno)],
+                               macro=nodesentence.macro,
                                addedinfos=[],
                                plural=nodesentence.plural )
             else: # having a non-empty override
@@ -230,6 +236,7 @@ class WmlNode:
                                wmlinfos=[nodesentence.overrideinfo],
                                finfos=[self.fileref +
                                         str(nodesentence.lineno)],
+                               macro=nodesentence.macro,
                                addedinfos=nodesentence.addedinfo,
                                plural=nodesentence.plural )
                 else:
@@ -240,6 +247,7 @@ class WmlNode:
                                wmlinfos=[nodesentence.overrideinfo],
                                finfos=[self.fileref +
                                         str(nodesentence.lineno)],
+                               macro=nodesentence.macro,
                                addedinfos=[],
                                plural=nodesentence.plural )
         # if you don't have override and autowml is true
@@ -254,6 +262,7 @@ class WmlNode:
                                wmlinfos=[self.assemble_wmlinfo()],
                                finfos=[self.fileref +
                                         str(nodesentence.lineno)],
+                               macro=nodesentence.macro,
                                addedinfos=nodesentence.addedinfo,
                                plural=nodesentence.plural )
             else:
@@ -264,6 +273,7 @@ class WmlNode:
                                wmlinfos=[self.assemble_wmlinfo()],
                                finfos=[self.fileref +
                                         str(nodesentence.lineno)],
+                               macro=nodesentence.macro,
                                addedinfos=[],
                                plural=nodesentence.plural )
         # if you don't have override and autowml is false
@@ -278,6 +288,7 @@ class WmlNode:
                                wmlinfos=[],
                                finfos=[self.fileref +
                                         str(nodesentence.lineno)],
+                               macro=nodesentence.macro,
                                addedinfos=nodesentence.addedinfo,
                                plural=nodesentence.plural )
             else:
@@ -288,5 +299,6 @@ class WmlNode:
                                wml_infos=[],
                                finfos=[self.fileref +
                                         str(nodesentence.lineno)],
+                               macro=nodesentence.macro,
                                addedinfos=[],
                                plural=nodesentence.plural )
--- a/data/tools/pywmlx/state/machine.py
+++ b/data/tools/pywmlx/state/machine.py
@@ -249,7 +249,7 @@ class PendingLuaString:
                     _dictionary[_currentdomain] = dict()
                 loc_posentence = _dictionary[_currentdomain].get(self.luastring)
                 if loc_posentence is None:
-                    _dictionary[_currentdomain][self.luastring] = PoCommentedString(
+                    _dictionary[_currentdomain]['\x01' + self.luastring] = PoCommentedString(
                                 self.luastring,
                                 _currentdomain,
                                 orderid=(fileno, self.lineno, _linenosub),
--- a/data/tools/pywmlx/state/machine.py
+++ b/data/tools/pywmlx/state/machine.py
@@ -127,19 +127,13 @@ def switchdomain(lineno, domain):
         _currentdomain = domain
 
 
-def checksentence(mystring, finfo, *, islua=False):
+def checksentence(mystring, finfo, *, israw=False):
     m = re.match(r'\s*$', mystring)
     if m:
         wmlwarn(finfo, "found an empty translatable message")
         return 1
-    elif warnall() and not islua:
-        if "}" in mystring:
-            wmsg = ("found a translatable string containing a WML macro. "
-                    " Translation for this string will NEVER work")
-            wmlwarn(finfo, wmsg)
-            return 2
-        else:
-            return 0
+    elif not israw and "{" in mystring:
+        return 2
     else:
         return 0
 
@@ -228,7 +222,7 @@ class PendingLuaString:
             _linenosub += 1
             finfo = pywmlx.nodemanip.fileref + ":" + str(self.lineno)
             fileno = pywmlx.nodemanip.fileno
-            errcode = checksentence(self.luastring, finfo, islua=True)
+            errcode = checksentence(self.luastring, finfo, israw=True)
             if errcode != 1:
                 # when errcode is equal to 1, the translatable string is empty
                 # so, using "if errcode != 1"
@@ -297,6 +291,8 @@ class PendingWmlString:
         global _pending_overrideinfo
         global _linenosub
         global _pending_winfotype
+        global _pending_wmacroname
+        global _pending_wmacroline
         if _pending_winfotype is not None:
             if self.ismultiline is False and self.istranslatable is False:
                 winf = _pending_winfotype + '=' + self.wmlstring
@@ -306,7 +302,10 @@ class PendingWmlString:
             return
         if self.istranslatable:
             finfo = pywmlx.nodemanip.fileref + ":" + str(self.lineno)
-            errcode = checksentence(self.wmlstring, finfo, islua=False)
+            errcode = checksentence(self.wmlstring, finfo, israw=self.israw)
+            macro = None
+            if errcode == 2:
+                macro = (_pending_wmacroname, pywmlx.nodemanip.fileref, _pending_wmacroline)
             if errcode != 1:
                 # when errcode is equal to 1, the translatable string is empty
                 # so, using "if errcode != 1"
@@ -318,6 +317,7 @@ class PendingWmlString:
                     self.wmlstring = re.sub('""', r'\"', self.wmlstring)
                 pywmlx.nodemanip.addNodeSentence(self.wmlstring,
                                              domain=_currentdomain,
+                                             macro=macro,
                                              ismultiline=self.ismultiline,
                                              lineno=self.lineno,
                                              lineno_sub=_linenosub,
--- a/data/tools/pywmlx/state/machine.py
+++ b/data/tools/pywmlx/state/machine.py
@@ -70,6 +70,11 @@ _pending_winfotype = None
 # If no lua functions already encountered, this var will be None
 _pending_luafuncname = None
 
+# the macro definition currently analyzed in a wml code (if any).
+# If outside of any macros, this var will be None
+_pending_wmacroname = None
+_pending_wmacroline = None
+
 # ----------
 
 # pending lua/wml string (they will be evaluated, and if translatable it will
--- a/data/tools/pywmlx/state/wml_states.py
+++ b/data/tools/pywmlx/state/wml_states.py
@@ -12,7 +12,7 @@ class WmlIdleState:
         self.iffail = None
 
     def run(self, xline, lineno, match):
-        _nextstate = 'wml_checkdom'
+        _nextstate = 'wml_enddef'
         if pywmlx.state.machine._pending_wmlstring is not None:
             pywmlx.state.machine._pending_wmlstring.store()
             pywmlx.state.machine._pending_wmlstring = None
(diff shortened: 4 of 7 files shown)


cooljeanius:
how do existing potfiles look when regenerated with a version of `wmlxgettext` that contains these changes?

Slayer95:
> 



> how do existing potfiles look when regenerated with a version of `wmlxgettext` that contains these changes?

(Invalid) sentences with ``{PARAMETERS}`` are removed, processed, and appended. They are otherwise the same.

```diff
diff --git a/test-before.pot b/test.pot
index b0ca097..f5a1949 100644
--- a/test-before.pot
+++ b/test.pot
@@ -1631,10 +1631,6 @@ msgstr ""
 msgid "This unit cannot be killed."
 msgstr ""
 
-#: TLU/main-macros.cfg:5
-msgid "Hello, {WHOM}."
-msgstr ""
-
 #. [message]: speaker=Mehir
 #: TLU/scenarios/00_Prologue.cfg:28
 msgid "Ridiculous price aside, I can't quite afford the elemental."
@@ -13215,3 +13211,19 @@ msgstr ""
 msgid "female^Citizen"
 msgstr ""
 
+#: TLU/main-macros.cfg:5
+msgid "Hello, world."
+msgstr ""
+
+#: TLU/main-macros.cfg:5
+msgid "Hello, very good world."
+msgstr ""
+
+#: TLU/main-macros.cfg:5
+msgid "Hello, very bad world."
+msgstr ""
+
+#: TLU/main-macros.cfg:5
+msgid "Hello, very meh world."
+msgstr ""
+
```

cooljeanius:
> > 
> 
> > how do existing potfiles look when regenerated with a version of `wmlxgettext` that contains these changes?
> 
> (Invalid) sentences with `{PARAMETERS}` are removed, processed, and appended. They are otherwise the same.

So, there's a current `wmllint` warning about situations like this ("macro reference in translatable string"); does that need to be updated at all?

Slayer95:
> > > 
> > 
> > 
> > > how do existing potfiles look when regenerated with a version of `wmlxgettext` that contains these changes?
> > 
> > 
> > (Invalid) sentences with `{PARAMETERS}` are removed, processed, and appended. They are otherwise the same.
> 
> So, there's a current `wmllint` warning about situations like this ("macro reference in translatable string"); does that need to be updated at all?

Yes. Updated!

CelticMinstrel:
So if I understand correctly, what this does is detect every possible combination that's substituted into the string, and add all of those separately to the gettext catalogue?

Slayer95:
> So if I understand correctly, what this does is detect every possible combination that's substituted into the string, and add all of those separately to the gettext catalogue?

Exactly

CelticMinstrel:
I can't really comment on the code since I have little idea what any given piece is doing, but that's the only possible approach that could work to support this, so I tentatively approve.

Elvish-Hunter:
A thing I noticed with this PR is that its scope is too wide. While the title says "Add macro parameters support to wmlxgettext", it also covers wmllint and wmlscope: unless it's strictly needed to make wmlxgettext work, anything regarding these two tools should be moved into another PR (merely editing the title wouldn't be a good idea).
Another issue is that the `wmlmacro` suffers from lack of comments, making it hard to understand what's going on at a first glance and why something has been done in a certain way.
With that said, I'm sending a forum PM to Nobun/@AncientLich, so he can check this PR too (he's the author of the Python version of wmlxgettext, after all).

Slayer95:
I hopefully addressed your concerns, @Elvish-Hunter

AncientLich:
@Slayer95 Adding a support for WML Macros would be too complex, error-prone and would be a very huge work for very little benefit: this is why there is no direct support for WML Macros.
This is also the reason why I implemented the "special comments" "# po: " and "# po-ovverride: " in order to allow developer to customize informations when default ones are not meaningful or or requires extra informations. 
See: https://wmlxgettext-unoff.readthedocs.io/en/latest/enduser/wml.html (paragraphs 1.3 , 1.4 and 1.6)

In case of macros (and in 99,9999% of cases) you require to use "#po: " special comment. You can add how many "#po: " lines as you lilke.

In your example, however, I don't understand how wmlxgettext could obtain the sentences. Probably there is something added by other developers, but in my original code wmlxgettext is not able to construct translatable sentences interpreting WML macros, for the reason in preface (too complex for me, and too error-prone).

Note:
"po-ovverride" is very-very case-specific, and you will actually need it only when the default informations must bereplaced completely for that sentence with a custom one (because completely wrong in that situation, for example). Only in that case you could need to use "# po-override". In that case, if you require to add multiple line comment that must replace the default one, then you must input "#po-override" for the first line followed by "#po" for following lines for example

`# po-override: [message]: speaker=FinalBoss
# po: other informations (line 2)
# po: other informations (line 3)
[message]
   speaker=unknown_speaker
   message= _ "translatable message"
[/message]`


PS. Thank @Elvish-Hunter for pointing me this PR

CelticMinstrel:
> In your example, however, I don't understand how wmlxgettext could obtain the sentences. Probably there is something added by other developers, but in my original code wmlxgettext is not able to construct translatable sentences interpreting WML macros, for the reason in preface (too complex for me, and too error-prone).

I guess this suggests that some more comments on how it works would be good…?

Slayer95:
> This is also the reason why I implemented the "special comments" "# po: " and "# po-ovverride: " in order to allow developer to customize informations when default ones are not meaningful or or requires extra informations. In case of macros (and in 99,9999% of cases) you require to use "#po: " special comment.

Your po and po-override directives are only designed to add comments, and in no way, shape, or form do they allow to remove invalid sentences and add valid ones. Furthermore, ``# po-override:`` doesn't work as documented at all. Which is why in case of macros (and in 100% of cases) it doesn't solve any problem. In fact, in both of its featurings in the entire Wesnoth repository, it's used exactly like a ``# po:`` comment, as opposed to how it's documented.

> I guess this suggests that some more comments on how it works would be good…?

I have already added plenty of code comments, so I invite you to have another look to the code. I think Nobu's comment suggest that the fact that this is a pull request and not a feature request was somehow missed.

CelticMinstrel:
> Furthermore, `# po-override:` doesn't work as documented at all. Which is why in case of macros (and in 100% of cases) it doesn't solve any problem. In fact, in both of its featurings in the entire Wesnoth repository, it's used exactly like a `# po:` comment, as opposed to how it's documented.

Hmm? I think it's purpose was to override the automatically generated context comments that wmlxgettext generates. Are you saying it doesn't do that?

Slayer95:
> Hmm? I think its purpose was to override the automatically generated context comments that wmlxgettext adds. (If I recall correctly those are #: instead of #. but I might be remembering wrong.) Are you saying it doesn't do that?

It does override them, but it prints a comment using Python's ``['list', 'of', 'strings']`` representation.

CelticMinstrel:
> It does override them, but it prints a comment using Python's `['list', 'of', 'strings']` representation.

So it does what it's supposed to do but messes up on the final step. Okay.

Slayer95:
> So this reimplements the preprocessor of wesnoth. This was attempted before in wmlparser.py for units.wesnoth.org and didn't work out that well. (This was quite a while ago with a simpler preprocessor but there were perhaps also more changes to it at the time.) 

Not exactly. Its overlap with the preprocessor is just macro expansion. No support for conditional directives nor other types of directives.

> To solve that wmlparser3.py uses wesnoth to preprocess and then just parses WML.

The problem with using the wesnoth preprocessor is that each run supports a single set of defines. So, for example, if I want to get the possible outputs for _``You are playing in {ON_DIFFICULTY easy normal hard} difficulty.``, then I need to run the preprocessor with ``--preprocess=EASY``, ``--preprocess=NORMAL``, and ```preprocess=HARD``, and that's only after the knowledge that these defines are mutually exclusive. In general, the amount of preprocessor runs needed is exponential on the amount of variables (2^n) defined for a campaign/era.

Originally I tried to use it, but that attempt didn't end well. So, I found ``wmltools3``, which is already used by ``wmlscope`` and ``wmllint``, and so it seems it has been deemed good enough for the project, and also use it (therefore inheriting its faults such as deficient parse of parens, as seen in the test output).

> In general I think it'd be nice if this can be made to work. While I don't think it's good practice (even when it works) UMC authors probably construct translatable strings with macros right now and will continue to. And if the translation data is there wesnoth doesn't mind.

Historically, it was a bad practice because there were no tools to make those translations work at all. However, as long as tools exist, there is nothing wrong about it. On the other hand, ``_"Strings with $variable substitutions."`` will forever be a bad practice because of declensions that depend on a ``variable`` that is unknown to translators and to ``gettext``.

That's because even if #1135 were to be fixed, that would only solve the case for strings with one numerical variable, and both multivariate strings or strings with non-numerical variables would remain unsupported. In contrast, macros do solve those cases by exhaustively generating new PO sentences for translators to work with. If you insist that macros are a bad practice, then let me ask you a question: Is this a good practice?

https://github.com/wesnoth/wesnoth/blob/2b01dd8d7428720ea2e018c3deee555b308ff6dc/data/campaigns/Under_the_Burning_Suns/scenarios/09_Blood_is_Thicker_Than_Water.cfg#L2512-L2535

See also #6158

soliton-:
I don't think the comparison to `_"Strings with $variable substitutions."` is that useful since the main use case is for variable content that is determined at runtime. If there are better solutions for that then those should be explored but I don't see how that can possibly be solved with macros at all.

If the content is not runtime dependent or just dependent on a small number of possibilities then something like the `MESSAGE_DEPEND_ON_ALLY` solution works fine. To clarify the bad practice I mean is `_"String with {MACRO} argument injection."`. I don't mean any macro use is bad. Though if we actually could do away with macros without losing features...

Given this is based on the same code as wmllint there is presumably no chance that it is going to work correctly in all cases. Looking at the last of your examples you already know that. So now the question is what benefit does this bring. Some cases will work and some won't and UMC authors should work out what is what? Do you think you can improve this to get more cases to work?

Btw, you have `msgid "Hello, very test opt whitespace and quotes world."` in your pot output while `wesnoth --preprocess` outputs
```
[message]
        message=_"Hello, very " +
                "test opt whitespace and quotes world."
[/message]
```
That does not look like it's going to work. You'd need `{MOODY_GREET message world (QUALIFIER=test opt whitespace and quotes)}`. Similar issue with the other cases that inject double quoted arguments.

I think in its current form this can only be beneficial if someone works out exactly what cases work and we somehow get UMC authors to make sure to just use those. Or rather reject all other cases with clear error messages...

CelticMinstrel:
> I think in its current form this can only be beneficial if someone works out exactly what cases work and we somehow get UMC authors to make sure to just use those. Or rather reject all other cases with clear error messages...

I agree that this would need to tell the user if they've done something that isn't going to work.

Slayer95:
> but I don't see how that can possibly be solved with macros at all.

When ``variable`` has too many alternatives (such as randomly generated unit names), it's not viable (but still possible). Still, for problems such as UtBS dialogues, my recommendation would be:

```
#define CUSTOMIZED_MSG_GN NAME
[case]
value={NAME}
[message]
speaker={NAME}
message=_"foo {NAME} bar"
[/message]
[/case]
#enddef

[switch]
variable=ally.name
{CUSTOMIZED_MSG_GN Grog}
{CUSTOMIZED_MSG_GN Nog}
{CUSTOMIZED_MSG_RJ Rogrimir}
{CUSTOMIZED_MSG_RJ Jarl}
[/switch]
```

> Some cases will work and some won't and UMC authors should work out what is what? Do you think you can improve this to get more cases to work?

UMC authors are already used to wmllint's parens restrictions. But I can probably improve that, yes. Is there a preprocessor test suite I can look at for that purpose?

> That does not look like it's going to work. You'd need {MOODY_GREET message world (QUALIFIER=test opt whitespace and quotes)}. Similar issue with the other cases that inject double quoted arguments.

Oh! I didn't realize those quoted arguments are untranslatable. I'll have to check what happens if they have ``_`` preffix. I have a fix in mind so long as I get the correct tokens from ``wmliterator``

> I agree that this would need to tell the user if they've done something that isn't going to work.

While the logged info for failures is not very useful yet, that's the intention. I'll polish it. And also ensure that the quotes issue is caught. As for parens, that's a parser bug and out of scope imo, but I can probably investigate it if that's all is missing.

CelticMinstrel:
> When `variable` has too many alternatives (such as randomly generated unit names), it's not viable (but still possible). Still, for problems such as UtBS dialogues, my recommendation would be:

That's overkill in most cases. When it's a unit's name that needs to be substituted, we already have a solution for it:

```
[message]
    speaker=unit
    message=_"foo $name bar"
    female_message=_"female^foo $name bar"
[/message]
```

Of course, if the message needs to depend on more than just the gender of the unit (for example, to use different dialects for a dwarf than for a troll), then your solution seems perfectly reasonable to me.

Slayer95:
> We'd definitely love to have a preprocessor test suite, as the lack of tests makes us reluctant to touch the preprocessor. It should probably be done as a Boost unit test rather than using the WML unit test system.

Snapshot testing should do the job.

> Is it possible to at least print a warning when code that triggers the bug is detected? Or do you not even have that level of information?

Yes. I've just added warnings for all known failure modes. Fixed the quotes issue, too.

The following sentences are extracted from ``wmlxgettext`` and ``--preprocess`` output, which perfectly match for non-commented lines.

<details><summary>Sentences</summary>

```
"Hello, AB."
"Hello, Albus."
"Hello, Harry."
"Hello, ladies, and gentlemen."
"Hello, Nicholas."
"Hello, very easy world."
"Hello, very good world."
"Hello, very great world."
"Hello, very hard world."
"Hello, very meh argument in parens."
"Hello, very meh land."
"Hello, very meh planet."
"Hello, very meh universe."
"Hello, very meh world."
"Hello, very nested land."
"Hello, very nested planet."
"Hello, very nested universe."
"Hello, very normal world."
"Hello, world."
"Hello, wow."
```
</details>

<details><summary>test.cfg</summary>

```rust
#textdomain wesnoth-test

#define GREET MODE WHOM
[{MODE}]
    {MODE} = _ "Hello, {WHOM}."
[/{MODE}]
[{MODE}]
    {MODE} = _ "Hello, {ON_DIFFICULTY Nicholas Harry Albus}."
[/{MODE}]
#enddef

{GREET message world}

#define MOODY_GREET MODE WHOM
#arg QUALIFIER
meh#endarg
   {GREET {MODE} (very {QUALIFIER} {WHOM})}
#enddef

#define DOGE
wow#enddef

#define A
A#enddef

#define B
B#enddef

#define AB
{A}{B}#enddef

#define CONCAT X Y
{X}, and {Y}#enddef

{GREET message {DOGE}}
{GREET message {AB}}
{GREET message ({CONCAT ladies gentlemen})}

{MOODY_GREET message world}
{MOODY_GREET message world QUALIFIER=good}
{MOODY_GREET message world (QUALIFIER=great)}

{MOODY_GREET message (argument in parens)}

{MOODY_GREET message ({ON_DIFFICULTY universe planet land})}
{MOODY_GREET message ({ON_DIFFICULTY universe planet land}) QUALIFIER=nested}
{MOODY_GREET message world (QUALIFIER={ON_DIFFICULTY easy normal hard})}

# quotes: untranslatable
# OK - Untranslatable quoted argument " argument with outer quotes and inner padding " at D:\[redacted]\test.cfg:51
{MOODY_GREET message " argument with outer quotes and inner padding "}
# OK - Untranslatable quoted argument " argument in parens with outer quotes and inner padding " at D:\[redacted]\test.cfg:53
{MOODY_GREET message (" argument in parens with outer quotes and inner padding ")}
# OK - Untranslatable quoted argument "optional argument in parens with outer quotes" at D:\[redacted]\test.cfg:55
{MOODY_GREET message world (QUALIFIER="optional argument in parens with outer quotes")}
# OK - Untranslatable quoted argument argument "with inner quotes" at D:\[redacted]\test.cfg:57
{MOODY_GREET message (argument "with inner quotes")}

# parens abuse (shouldn't work but do)
# preprocessor expects 2 arguments, but has 3 arguments
# {MOODY_GREET message world QUALIFIER=(misplaced parens)}
# {MOODY_GREET message world (QUALIFIER=(too many parens))}
# Untranslatable quoted argument "misplaced parens and quotes" at D:\[redacted]\test.cfg:64
# {MOODY_GREET message world QUALIFIER=("misplaced parens and quotes")}

# should work but don't

# Macro {ON_DIFFICULTY easy   universe   normal   planet   hard   land} expects 3 arguments, but has 6 arguments.
# {MOODY_GREET message ({ON_DIFFICULTY (easy universe) (normal planet) (hard land)}) QUALIFIER=nested}

# Macro {ON_DIFFICULTY4 test   prevOpt   whitespacetest   nextOpt   whitespace   test   noneOpt   whitespace   test   bothOpt   whitespace} expects 4 arguments, but has 11 arguments.
# {MOODY_GREET message world QUALIFIER={ON_DIFFICULTY4 ( test prevOpt whitespace) (test nextOpt whitespace ) (test noneOpt whitespace) ( test bothOpt whitespace )}}

# Macro {ON_DIFFICULTYeasy} not defined. Strings containing it will NOT be processed in the generated .pot file.
# This error might also be fixed by adding extra parens around any macro arguments.
# {MOODY_GREET message {ON_DIFFICULTY (easy universe) (normal planet) (hard land)} QUALIFIER=nested}

# Macro {ON_DIFFICULTYuniverseplanetland} not defined. Strings containing it will NOT be processed in the generated .pot file.
# This error might also be fixed by adding extra parens around any macro arguments.
# {MOODY_GREET message {ON_DIFFICULTY universe planet land}}

# Macro {ON_DIFFICULTYuniverseplanetland} not defined. Strings containing it will NOT be processed in the generated .pot file.
# This error might also be fixed by adding extra parens around any macro arguments.
# {MOODY_GREET message {ON_DIFFICULTY universe planet land} QUALIFIER=good}

# Macro {ON_DIFFICULTYeasynormalhard} not defined. Strings containing it will NOT be processed in the generated .pot file.
# This error might also be fixed by adding extra parens around any macro arguments.
# {MOODY_GREET message world QUALIFIER={ON_DIFFICULTY easy normal hard}}

# Macro {ON_DIFFICULTYeasynormalhard} not defined. Strings containing it will NOT be processed in the generated .pot file.
# This error might also be fixed by adding extra parens around any macro arguments.
# {MOODY_GREET message world QUALIFIER={ON_DIFFICULTY easy normal hard}}
```
</details>

CelticMinstrel:
> Snapshot testing should do the job.

?

Slayer95:
> > Snapshot testing should do the job.
> 
> ?

https://pypi.org/project/pytest-snapshot/

Step 1. Get a few ``.cfg`` files together.
Step 2. Tell the snapshot library that their preprocessor output should always match the snapshot.
Step 3. Profit!

BTW, fixed parens to the best of my knowledge. I guess it's time to split stuff into new PRs.

AncientLich:
> > Adding a support for WML Macros would be too complex, error-prone and would be a very huge work for very little benefit: this is why there is no direct support for WML Macros.

> I definitely understand this sentiment, but someone has supposedly done that huge work so (assuming it actually works) I think we might as well reap the little benefit.

Sorry if I didn't explain correctly what I meant. I can only admire anybody who was able to implement that feature I wasn't able to do myself, and I wasn't even sure that souch feature was indeed added by anyone else. Now that I have a confirmation of that, I can explicitly state that I have no enough knowledge to solve the issue, so I will explain how wmlxgettext was assumed to work when I wrote it, since I know only know how the code worked when I wrote it. Unluckly my contribution, cannot go further. 

> Hmm? I think its purpose was to override the automatically generated context comments that wmlxgettext adds. (If I recall correctly those are #: instead of #. but I might be remembering wrong.) Are you saying it doesn't do that?

Me neither. However the purpose was to replace a SINGLE **context comment** of a **single instance of that sentence**, not touching source comments.
When I used "comments" using plural form (and sorry for my bad english) I meant the sum of single informations (separated by commas) retrieved by context comment that represent, however a single line of `#. context comment` (assuming `#.` is for context comment and `#:` for source comment, but could be the opposite)

> Furthermore, `# po-override:` doesn't work as documented at all. Which is why in case of macros (and in 100% of cases) it doesn't solve any problem.

Like I said in the starting point of this reply, I am sorry to say that I am only able to answer only partially, hoping that my answer would allow other developers to understand where the problem is currently located in order to allow `# po-override:` to work again as originally intended, or to remove it at all if it is evalued as a no-benefit feature.
` # po-override:` was designed in order to erase all the default comments for the very next translatable sentence found in .cfg/.lua file immediately following `# po-override:` (and the same logic was also for `# po:`).
however, it is important to remember that .po format wants that every sentence is indexed as unique, and the same exact sentence cannot be placed more than once in a .po file.
For this reason, if a sentence (in particular, more easily, a short sentence) is used more times in your .cfg code, wmlxgettext take notes every time it occurs. So here is an example how `# po-override:` supposed to work when a sentence occurs multiple times

`[message]`
`speaker=Alfa`
`message=_ "Hi. "`
`[/message]`
`[message]`
`speaker=Beta`
`message=_ "Hi. "`
`[/message]`
`# po-override: [message] speaker=Gamma`
`[message]`
`speaker=DumbSpeakerName`
`message=_ "Hi. "`
`[/message]`
`[message]`
`speaker=Delta`
`message=_ "Hi. "`
`[/message]`



in that case, the expected result would be (probably in different order, and now I don't remember how line of codes are commented)
`#. [message] speaker=Alfa`
`#. [message] speaker=Beta`
`#. [message] speaker=Gamma`
`#. [message] speaker=Delta`
`#: (line of code 1)`
`#: (line of code 2)`
`#: (line of code 3)`
`#: (line of code 4)`
`msgid="Hi."`
`msgstr=""`

So `#po-override` was designed to prevent to use the automated comment that wmlxgettex would be apply for that specific instance of the sentence, replacing it with custom one (in the example it would write `"[message] speaker=Gamma"` instead of `"[message] speaker=DumbSpeakerName"`. The original intent of that "command" was originally to destroy a an automated line of comment applied to a translatable line when wrong in that context (expecially this could happen when using translatable lines in macro that, in the original tool, was not managed at all and wmlxgettext, at that time, could possibly generate a line of comment.

Probably, seeing that now wmlxgettext is able to manage also macro in a more sophisticated way than the original code I wrote, I could assume that `# po-ovverride` could be dropped, as long as wmlxgettext would hopefully never add an undersired / wrong automated context comment that would need a replacement / erase. I don't know if that option is actually useful somehow even in a corner-case situation...

Slayer95:
@AncientLich thanks for your detailed explanation. I can now see that the ``po-override`` feature provides legitimate benefits, although it's unfortunate it has seen little usage.

I also figured out how that feature is relevant to macros, which you briefly referred at in your first comment. That is, in fixing the parsed ``#.`` context for _macro calls where the argument is a translatable string_. However, that's not what the current PR deals with, since it actually deals with _translatable strings that include macro parameters_ (every other commit and even the macro expansion feature I recently added works toward that goal.)

soliton-:
> > but I don't see how that can possibly be solved with macros at all.
> 
> When `variable` has too many alternatives (such as randomly generated unit names), it's not viable (but still possible). 

You seem to have a ignored or misunderstood the part about "variable content that is determined at runtime".  If the content is the current date for example then how are you going to map that to a macro at all?

> > Some cases will work and some won't and UMC authors should work out what is what? Do you think you can improve this to get more cases to work?
> 
> UMC authors are already used to wmllint's parens restrictions.

Regarding this and something about wmllint's parsing being resilient. You have to be generous to even call it parsing. I've recently improved the parse_attribute function a little but it's still really just guessing. (For example it's easy to see it has no clue about `key1,key2=value1,value2` or multi-line assignments.)

The thing is that wmllint is not essential. If it does nonsense you revert that part or perhaps workaround it or just stop using wmllint. If the way translations work depends on that kind of quality that is much less acceptable in my opinion.

Slayer95:
> You seem to have a ignored or misunderstood the part about "variable content that is determined at runtime". If the content is the current date for example then how are you going to map that to a macro at all?

Well. Variable content that is determined at runtime simply cannot be inserted in translatable strings, be it using macros or variables. The function used to generate the date string should be locale-aware instead, and such a string should be context independent, i.e. place it in a header or a box. Let's remain focused in what this fixes, rather than in the unfixable.

> You have to be generous to even call it parsing. I've recently improved the parse_attribute function a little but it's still really just guessing. (For example it's easy to see it has no clue about key1,key2=value1,value2 or multi-line assignments.)

I am only using the preprocessor-related bits of those libraries. I don't see how your examples are relevant. Are you saying that ``{MACRO OPT1,OPT2=FOO,BAR}`` is a thing? If it is, sorry but I cannot grief over such syntax loss. As for any issues with multiline calls, someone more interested will have to deal with them in the relevant library.

> it does nonsense you revert that part or perhaps workaround it or just stop using wmllint. If the way translations work depends on that kind of quality that is much less acceptable in my opinion.

I don't think it does nonsense related to macros anymore, since your previous observations prompted me to fix glaringly wrong behavior. Even if something remains wrong, warnings will be emitted if wmlmacro is passed nonsense.

AncientLich:
There is another point I can add about _"variablecontent that is determined at runtime"_. There is an important thing to take mind when designing WML/lua code that contains sentences that will be translated in different languages.

Wmlxgettext doesn't (and can't) use code included in wesnoth binary (a.k.a. wesnoth.exe, in windows), for various reason, one of them is related to wesnoth(.exe) build process: it is required that wmlxgettext work even if wesnoth(.exe) does not exist, and for this reason wmlxgettext must be 100% indipendend of wesnoth.exe.

This is also the reason why wmlxgettext and wesnoth.exe read WML/lua code in a complete different way.
While wesnoth.exe interpretes WML/lua code in runtime (so it can read what actually code is doing at that point and knows actual data at that point of execution of the campaign / other addon).

This is not true for wmlxgettext.
Wmlxgettext does not execute WML/lua code, in any way, and it does not parse WML/lua code in any way. It _"simply"_ read the code as-is, trying to extract all the relevant informations for translations. In other worlds, wmlxgettext can't be as accurate as an actual WML/lua parser / interpreter like the one included in wesnoth.exe code.

This also means that a developer, while writing her/his code, if she/he wants to make it translatable, must consider not only how the code will actually work, but also how the code will be interpreted when the text is not interpreted, but read from the start to the end of the file, like a person would read a book (this is how wmlxgettext works).

This means, that having a variable content that is determinated a runtime, while it will managed correctly in an actual interpreter, like wesnoth.exe, it is not guaranteed to work when read by an application, like wmlxgettext, that does not parse the code, but only reads it (and, now, also trying to decode macros in some way).

cooljeanius:
> @AncientLich thanks for your detailed explanation. I can now see that the `po-override` feature provides legitimate benefits, although it's unfortunate it has seen little usage.
> 

I intend to use it further in ANO once we can get some guarantees as to what exactly the future of the directive is going to be, pending resolution of the PRs related to it (this and #7574)

Slayer95:
The status of this PR is that I need to

1) Rebase this PR once again
2) Rerun the tests. 
3) The test script is written in Node.js, and it covers a single WML+Macros file, so I am not clear how to go about incorporating them into the repo, or if I should do that at all.
4) The original version of this PR included changes to wmllint so that it no longer complains of trying to use this now-supported feature. The current version lacks this diff, so another PR shiuld ve opened for it.
5) @irydacea has raised transversal concerns about confusion arising from invoking macros from third-party libraries, and passing arguments that result in in-game sentences 
absent from the library's PO file. This should be dealt with by adding a --dependencies flag (and UI) to wmlxgettxt.
6) Complementary to (5), libraries should be able to declare strings to be included in the POT file, without actually using them. That's already possible with the #wmlxgettext directive, but this PR fails to add support for macros yet. This requires further changes to wmltools.

Pentarctagon:
@Slayer95 do you plan to get back to this PR at some point?

Slayer95:
I was off GitHub for quite some time because of its unwelcome mandatory 2FA. I finally figured it out today, so I'll eventually deal with https://github.com/wesnoth/wesnoth/pull/7540#issuecomment-1640213865

I'm glad past myself wrote up such a summary.

cooljeanius:
Any idea about the merge conflicts?

Slayer95:
The conflicting commits are all mine, and not overly complex. I apologize for keeping you all waiting, but I still need to schedule some time for Wesnoth.

Pentarctagon:

PR #7542, opened 2023-04-16
Achievements - use lua_error instead of returning -2

### LINKED ISSUE #7528: wesnoth.achievements.progress should probably throw an error instead of returning an error code
This means that people who don't care about the error (because they know it'll never happen) can ignore it, and people who do care can use `pcall` and check the error if it occurs.

Pentarctagon:
Fixes #7528

### DIFF
--- a/src/scripting/game_lua_kernel.cpp
+++ b/src/scripting/game_lua_kernel.cpp
@@ -3189,8 +3189,8 @@ int game_lua_kernel::intf_get_achievement(lua_State *L)
  * - Arg 2: string - id.
  * - Arg 3: int - the amount to progress the achievement.
  * - Arg 4: int - the limit the achievement can progress by
- * - Ret 1: int - the achievement's current progress after adding amount, -2 if not found, -1 not a progressable achievement (including if it's already achieved)
- * - Ret 2: int - the achievement's max progress, -2 if not found, or -1 not a progressable achievement
+ * - Ret 1: int - the achievement's current progress after adding amount or -1 if not a progressable achievement (including if it's already achieved)
+ * - Ret 2: int - the achievement's max progress or -1 if not a progressable achievement
  */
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Looks good.

I'm just checking, but none of the code using this API in mainline was checking for the -2 return value, right?

Pentarctagon:
Correct.

CelticMinstrel:
Okay, and… in all the places in mainline that use this API, it's correct for it to print that error message to chat, right? I think it would show as a Lua Error, which is slightly less than ideal if it came from the WML tag, but that's a minor detail.

Pentarctagon:
Yep.

src/scripting/game_lua_kernel.cpp
```diff
@@ -3189,8 +3189,8 @@ int game_lua_kernel::intf_get_achievement(lua_State *L)
  * - Arg 2: string - id.
  * - Arg 3: int - the amount to progress the achievement.
  * - Arg 4: int - the limit the achievement can progress by
- * - Ret 1: int - the achievement's current progress after adding amount, -2 if not found, -1 not a progressable achievement (including if it's already achieved)
- * - Ret 2: int - the achievement's max progress, -2 if not found, or -1 not a progressable achievement
+ * - Ret 1: int - the achievement's current progress after adding amount or -1 not a progressable achievement (including if it's already achieved)
+ * - Ret 2: int - the achievement's max progress or -1 not a progressable achievement
```
stevecotton, line 3193: Both lines would be easier to read if the "not" was changed to "if not".

stevecotton (approved):

src/scripting/game_lua_kernel.cpp
Pentarctagon: Updated.

CelticMinstrel (approved):

Issue #7543, opened 2023-04-16
Pango 1.50.12+ cannot load non-system fonts for Wesnoth 1.16 on Windows


Wedge009:
### Game and System Information

- **Version:** 1.16.9
- **Downloaded from:** Steam, but I expect other Windows releases to exhibit the same problem.
- **Build info:** Official release build
- **OS:** Windows 7 and 10

### Description of the bug

Pango reports being unable to load Lato font as per logs:
```
(process:11320): Pango-WARNING **: 09:21:34.812: couldn't load font "Lato Not-Rotated 20", falling back to "Sans Not-Rotated 20", expect ugly output.

(process:11320): Pango-WARNING **: 09:21:34.824: couldn't load font "Lato 14", falling back to "Sans 14", expect ugly output.

(process:11320): Pango-WARNING **: 09:21:34.840: couldn't load font "Lato Not-Rotated 17", falling back to "Sans Not-Rotated 17", expect ugly output.
Checking lua scripts... ok

(process:11320): Pango-WARNING **: 09:21:35.269: couldn't load font "Lato Italic Not-Rotated 17", falling back to "Sans Italic Not-Rotated 17", expect ugly output.

(process:11320): Pango-WARNING **: 09:21:35.275: couldn't load font "Lato Bold Not-Rotated 17", falling back to "Sans Bold Not-Rotated 17", expect ugly output.

(process:11320): Pango-WARNING **: 09:21:35.307: couldn't load font "Lato Not-Rotated 16", falling back to "Sans Not-Rotated 16", expect ugly output.

(process:11320): Pango-WARNING **: 09:21:35.311: couldn't load font "Lato Not-Rotated 15", falling back to "Sans Not-Rotated 15", expect ugly output.

(process:11320): Pango-WARNING **: 09:21:35.313: couldn't load font "Lato Not-Rotated 13", falling back to "Sans Not-Rotated 13", expect ugly output.

(process:11320): Pango-WARNING **: 09:21:35.317: couldn't load font "Lato Not-Rotated 14", falling back to "Sans Not-Rotated 14", expect ugly output.
```

So far, I only see this issue on 1.16 releases for Windows. 1.17 doesn't appear affected, neither are Linux releases.

1.16.9 on Windows 7:
![169](https://user-images.githubusercontent.com/13679322/232349974-097db2fb-c077-4fa5-a0db-5261b7f56a2c.jpg)

### Steps to reproduce the behavior

[...]

Issue #7462, opened 2023-03-19
Monospace font doesn't work on pango markup


Toranks:
### Game and System Information

The Battle for Wesnoth version 1.17.14 x86_64
Running on Microsoft Windows 7 Service Pack 1 (6.1.7601) x86_64
Distribution channel: Steam

Game paths
==========

Data dir:        D:\SteamLibrary\steamapps\common\wesnoth
User config dir: C:\Users\USER\Documents\My Games\Wesnoth1.17
User data dir:   C:\Users\USER\Documents\My Games\Wesnoth1.17
Saves dir:       C:\Users\USER\Documents\My Games\Wesnoth1.17\saves
Add-ons dir:     C:\Users\USER\Documents\My Games\Wesnoth1.17\data\add-ons
Cache dir:       C:\Users\USER\Documents\My Games\Wesnoth1.17\cache
Logs dir:        C:\Users\USER\Documents\My Games\Wesnoth1.17\logs

Libraries
=========

Boost:             1.81
Lua:               5.4.4
OpenSSL/libcrypto: 3.0.0h-dev (runtime 3.0.0h-dev)
Cairo:             1.17.8 (runtime 1.17.8)
Pango:             1.50.14 (runtime 1.50.14)
SDL:               2.26.3 (runtime 2.26.3)
SDL_image:         2.6.3 (runtime 2.6.3)
SDL_mixer:         2.6.3 (runtime 2.6.3)

Features
========

Lua console completion:       yes
Win32 notifications back end: yes

Current video settings
======================

SDL video drivers:        [windows] offscreen dummy
Window size:              1920x1080
Game canvas size:         1920x1080
Final render target size: 1920x1080
Screen refresh rate:      60
Screen dpi:               96.00x96.00
Renderer:                 direct3d (hw)
Maximum texture size:     16384x16384
VSync:                    on

Current audio settings
======================

SDL audio drivers:  [wasapi] directsound winmm
Number of channels: 2
Output rate:        44100 Hz
Sample format:      signed 16 bit little-endian
Sample size:        4096 bytes

Installed add-ons
=================

A_New_Order:                  1.5.0
Advance_Wesnoth_Wars_fork:    1.17.13.5
Akladian_Music:               1.0.1
An_Orcish_Incursion:          1.0.3
Dead_Bodies:                  1.3.3
Dead_Bodies_terrain:          1.0.2
Eastern_Invasion_Dalas:       2.2.9
Era_of_Magic:                 4.0.2
[...]

PR #7544, opened 2023-04-17
Backport of #6685

Pentarctagon:
Should hopefully address #7543

### DIFF
--- a/packaging/windows/SConscript
+++ b/packaging/windows/SConscript
@@ -9,6 +9,8 @@ for bin in binaries:
     try:
         env["RCCOM"] = '$RC $_CPPDEFFLAGS $RCINCFLAGS ${RCINCPREFIX} ${SOURCE.dir} $RCFLAGS -i $SOURCE -o $TARGET'
         res = env.RES(bin + ".rc", CPPPATH=["#/"])
+        if bin == "wesnoth":
+            Depends(res, File("wesnoth.exe.manifest"))
     except AttributeError:
         pass
 
--- a/packaging/windows/wesnoth.exe.manifest
+++ b/packaging/windows/wesnoth.exe.manifest
@@ -24,4 +24,9 @@
             <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
         </application>
     </compatibility>
+    <asmv3:application>
+        <asmv3:windowsSettings>
+            <activeCodePage xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">UTF-8</activeCodePage>
+        </asmv3:windowsSettings>
+    </asmv3:application>
 </assembly>
--- a/src/font/font_config.cpp
+++ b/src/font/font_config.cpp
@@ -38,18 +38,7 @@
 
 #include <cairo-features.h>
 
-#ifdef CAIRO_HAS_WIN32_FONT
-#include <windows.h>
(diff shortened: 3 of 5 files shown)


CelticMinstrel:
Issue numbers in the title don't create an auto-link, so: #6685

loonycyborg (approved):

Wedge009:
@Pentarctagon The Windows installer from the artefacts seemed to include a build that loads the Lato font correctly. But you also need to back-port #6923 because now #6908 is showing.

gfgtdf:
This looks like a risky commit to me,  did you test that it did nott break https://github.com/wesnoth/wesnoth/issues/7462 in 1.16

gfgtdf:
Accidentally clicked the wrong button

Pentarctagon:
> This looks like a risky commit to me, did you test that it did nott break https://github.com/wesnoth/wesnoth/issues/7462 in 1.16

It would of course result in that bug occurring in 1.16 as well.

> Furthermore iirc the " utf8" part is not supported by windows 7, so if it is needed for the fix, the fix might not work under windows 7

I think this was tested and worked fine? The minimum windows version for 1.17 is still listed as Windows 7 and I'm pretty sure I would've updated that if it changed because of #6685.

Pentarctagon:
@Wedge009 could you check whether or not having non-ascii characters in the path where you installed Wesnoth to causes any issues?

Wedge009:
I don't normally have any non-ASCII characters nor special characters (besides space) in my Windows directories. I'm travelling again, should be back home in around 9 days from time of this post.

gfgtdf:
> It would of course result in that bug occurring in 1.16 as well.

Knowingly breaking addons code in stable, even if it's just visual issues that some might consider minor, is obviously something that we shouldtry try to avoid.

Are there reasons against simply shipping 1.16.9 with the previous pango versions?

gfgtdf:
Since there is probably still some time until 1.16.10 gets tagged, my approach would be to wait and hope that either #7462 or #7543 is fixed upstream (especially since it seems that pango releases are kinda frequent), and postpone the decision of merging this for now.

Pentarctagon:
Sure, there's no rush. My guess though would be that since #7462 happened when switching to us using fontconfig on Windows rather than from upgrading to a particular version of pango, it's less likely to be quickly fixed.

Pentarctagon:
Seems like https://github.com/wesnoth/wesnoth/pull/7547 can't be backported since the version of fontconfig provided by the Scout steam runtime is too old.

Wedge009:
@Pentarctagon I took the installer from the most recent 'action' that I could find (https://github.com/wesnoth/wesnoth/actions/runs/4812991123) and used `é` in my installation path (this is on Windows 7). In this case I get a crash on launch, though the logs appear to show something closer to #7217 than #7543.

`Dev message: /wesnoth/src/gui/widgets/window.cpp:945 in function 'layout' found the following problem: Failed to size window; wanted size 980,426877 available size 1920,1200 screen size 1920,1200.`

Nonetheless, removing the `é` from the path name such that only ASCII characters and 'normal' symbols (space character, `+` sign from `1.16.9+dev`, etc, remain) and the game starts normally again.

For what it's worth, at least the Lato font is loading normally. Though #7089 is introduced to 1.16 (previously I only observed it in 1.17).

Pentarctagon:
Alright, so this can't be merged then.

PR #7545, opened 2023-04-17
HttT: Restore context information for Haralandum's name

Slayer95:
4eb7fccaa0 made this comment a ``po-override`` without given justification.

### DIFF
--- a/data/campaigns/Heir_To_The_Throne/scenarios/20b_Underground_Channels.cfg
+++ b/data/campaigns/Heir_To_The_Throne/scenarios/20b_Underground_Channels.cfg
@@ -124,7 +124,7 @@
         side=4
         type=Ancient Wose
         id=Haralamdum
-        # po-override: Haralamdum is ancient, and his mind does not work well any more. He takes things very slowly and carefully, even for a Wose. He halts and pauses, often searching for something in his memory, using meaningless syllables to fill the gap. The syllables of his name come from the phrase he uses to mean "that is all" or "I am done" which he says often, usually out of frustration with how hasty everyone seems. Nobody alive knows his real name, and he has likely forgotten as well. Feel free to add, move, or remove pauses to break up, or even scatter, his train of thought. For inspiration, think of Asperger Syndrome.
+        # po: Haralamdum is ancient, and his mind does not work well any more. He takes things very slowly and carefully, even for a Wose. He halts and pauses, often searching for something in his memory, using meaningless syllables to fill the gap. The syllables of his name come from the phrase he uses to mean "that is all" or "I am done" which he says often, usually out of frustration with how hasty everyone seems. Nobody alive knows his real name, and he has likely forgotten as well. Feel free to add, move, or remove pauses to break up, or even scatter, his train of thought. For inspiration, think of Asperger Syndrome.
         name= _ "Haralamdum"
         controller=ai
         canrecruit=yes


Pentarctagon:
What's the difference between the two?

Slayer95:
> What's the diff

```diff

-#. ['Haralamdum is ancient, and his mind does not work well any more. He takes things very slowly and carefully, even for a Wose. He halts and pauses, often searching for something in his memory, using meaningless syllables to fill the gap. The syllables of his name come from the phrase he uses to mean "that is all" or "I am done" which he says often, usually out of frustration with how hasty everyone seems. Nobody alive knows his real name, and he has likely forgotten as well. Feel free to add, move, or remove pauses to break up, or even scatter, his train of thought. For inspiration, think of Asperger Syndrome.']
+#. [side]: type=Ancient Wose, id=Haralamdum
+#. Haralamdum is ancient, and his mind does not work well any more. He takes things very slowly and carefully, even for a Wose. He halts and pauses, often searching for something in his memory, using meaningless syllables to fill the gap. The syllables of his name come from the phrase he uses to mean "that is all" or "I am done" which he says often, usually out of frustration with how hasty everyone seems. Nobody alive knows his real name, and he has likely forgotten as well. Feel free to add, move, or remove pauses to break up, or even scatter, his train of thought. For inspiration, think of Asperger Syndrome.
 #: Heir_To_The_Throne/scenarios/20b_Underground_Channels.cfg:128
 msgid "Haralamdum"
 msgstr ""
```

Slayer95:
BTW, despite the long tip to translators, only the following languages change his name in some sort:

- bg: Хараламдум
- lt: Haralamdumas
- ru: Хараламдум
- sr: Хараламдум
- sr@ijekavian: Хараламдум
- zh_CN: 哈拉兰杜姆
- zh_TW: 喝阿郞敦

But all of them are actually phonetic transcriptions.

sevu (approved):

soliton-:
I mean the tip is rather about how to translate their dialog text not their name.

CelticMinstrel:
Why is this tip suggesting autism for inspiration? That doesn't seem like a great suggestion?

Slayer95:
> Why is this tip suggesting autism for inspiration? That doesn't seem like a great suggestion?

It's certainly not nice. It also struck me as odd, and as it turns out, the exact reference is wrong. Quoting [hopebridge](https://www.hopebridge.com/blog/does-aspergers-differ-from-autism-spectrum-disorder/):

> Individuals with Asperger’s Disorder do not typically exhibit language delays.
This is one of the major distinctions between the two disorders. Unlike classic autism, it is common for those with Asperger’s to have strong speech skills and an absence of language delays, although they do exhibit other communication challenges.

CelticMinstrel:
I think that quote is a bit misleading. Asperger's Syndrome is basically "autism but they're pretty good at pretending they're not". So, they might have strong speech skills in certain contexts but still have difficulty articulating things in other contexts. However, I still think it's wrong to use that as a reference, because it's simply too general. Autism is a broad spectrum of states comprising a variety of mixtures of various disabilities. Speech disorders are potentially included, but not necessarily. That is, some autistic people may have speech disorders while others do not.

I think a better example would be Alzheimer's. It's even more thematically fitting, since Haralandum's condition is due to age, and Alzheimer's is common among the aged. In fact, the names are similar enough that I wonder if someone actually got the two of them confused when writing the comment.

Slayer95:
Claiming that Asperger's "pretend" they are not autistic strikes to me as very unsensitive towards people who still today identify as AS. However, the root of the confusion is that that the relatively new 5th version of DSM (2013) merged AS together with other autism forms into a single umbrella ASD concept. Prior to that, DSM IV did provide some diagnosis criteria that favored AS' language skills.

Anyway, Asperger Syndrome concept is dated by now, so it probably shouldn't be referred to at all. And you are right that autism spectrum is too broad. Therefore, neither of them should be used. As for the Alzheimer hypothesis, it's plausible.

CelticMinstrel:
> Claiming that Asperger's "pretend" they are not autistic strikes to me as very unsensitive towards people who still today identify as AS.

No, it's literally a thing with all autistic people, in varying degrees. It's normally called "masking" rather than "pretending", but that's basically what it is.

But yes the Asperger's identification is dated now. I recommend either changing it to Alzheimer's or finding a different appropriate reference.

PR #7546, opened 2023-04-17
Support and use FEMALE_NAME in macros for named units with random gender

Slayer95:


soliton-:
If it's basically a translation bug to not specify a female name then the argument should not be optional.

Even better if the need for this could be eliminated. Why doesn't it work without this?

soliton-:
I guess the issue is that translations simply cannot specify a separate female translation so far?

Slayer95:
> If it's basically a translation bug to not specify a female name then the argument should not be optional.

Agreed. That change could go to 1.17 because of compat breakage.

> Even better if the need for this could be eliminated. Why doesn't it work without this?

> I guess the issue is that translations simply cannot specify a separate female translation so far?

Exactly so.

Slayer95:
> If it's basically a translation bug to not specify a female name then the argument should not be optional.

However, some (most) units don't have gender variants, so forcing duplication then would be odd.

CelticMinstrel:
The reasons for it to be optional are:

1. Compatibility. Adding a new non-optional argument to an existing macro breaks compatibility.
2. It's not always an issue. If it's a unit type that has only one gender, then there's no problem. Also, if they tack on `[+unit]gender=(fe)male` then there's no problem either.

Slayer95:
It seems like the preprocessor didn't like spaces in-between/around ``FEMALE_NAME=_``. Fixed.

Slayer95:
> We should actually support female_name in [unit].

I'd better do this (build fails because ``{MODIFY_UNIT}`` (i.e. ``[store_unit]``) is not allowed in ``SideWML``. Where's that supposed to be defined?

Slayer95:
It should be something like this? LMK if I'm missing something.

CelticMinstrel:
I think that looks good probably?

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
```diff
@@ -546,21 +546,21 @@
...
 #endif
         #hack to make Scylla female
         [+unit]
             gender=female
         [/unit]
 
 #ifdef HARD
-        {NAMED_GENERIC_UNIT 2 (Naga Warrior) 16 34 () ( _ "Lake Naga")}
-        {NAMED_GENERIC_UNIT 2 (Naga Warrior) 16 35 () ( _ "Lake Naga")}
+        {NAMED_GENERIC_UNIT 2 (Naga Warrior) 16 34 () ( _ "Lake Naga") FEMALE_NAME=_"female^Lake Naga"}
```
stevecotton, line 559: Maybe `# po: name of an enemy unit in water, not a geographic feature`

src/config.cpp
```diff
@@ -784,6 +784,17 @@ const config::attribute_value* config::get(config_key_type key) const
 	return i != values_.end() ? &i->second : nullptr;
 }
 
+/**
+ * Chooses a value. If the value specified by @a key is
+ * blank, then @a default_key is chosen instead.
+ */
```
stevecotton, line 790: Duplicated documentation - for this class, please put it in the header file only.

src/config.hpp
```diff
@@ -474,6 +474,12 @@ class config
 	 */
 	const attribute_value *get(config_key_type key) const;
 
+    /**
+     * Chooses a value. If the value specified by @a key is
+     * blank, then @a default_key is chosen instead.
```
stevecotton, line 479: ... If both values are blank or not set, then an empty value is returned.

src/config.hpp
```diff
@@ -474,6 +474,12 @@ class config
...
+     */
+    const attribute_value &get_or(const config_key_type key, const config_key_type default_key)
```
stevecotton, line 481: Can probably be a `const` method, and is missing the semicolon.

CelticMinstrel: `const attribute_value& get_or`

src/config.cpp
```diff
@@ -784,6 +784,13 @@ const config::attribute_value* config::get(config_key_type key) const
 	return i != values_.end() ? &i->second : nullptr;
 }
 
+const config::attribute_value& config::get_or(config_key_type key, config_key_type default_key) const
+{
+    check_valid();
```
Slayer95, line 789: > error: use of undeclared identifier 'check_valid'

I don't get why it's deemed undefined. Is it because it's called by an ``inline`` function?

Slayer95: Well. That wasn't it :|

CelticMinstrel: Isn't `check_valid()` the function that @gfgtdf deleted recently?

Slayer95: Oh! That makes sense. I didn't have that commit pulled.

stevecotton (changes requested):

Slayer95:
Updated

stevecotton:
The flatpak build is a separate failure, unrelated to this PR.

stevecotton (approved):

data/core/macros/unit-utils.cfg
```diff
@@ -132,15 +141,20 @@
...
+    #!  FEMALE_NAME=_"female^Mochi"}
     #
+#arg FEMALE_NAME
+{NAME_STRING}#endarg
     [unit]
         side={SIDE}
         type={TYPE}
         id={ID_STRING}
-        name={NAME_STRING}
+        male_name={NAME_STRING}
```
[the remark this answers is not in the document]
CelticMinstrel: I'm pretty sure most of the places where we do this use 3 keys – name, male_name, and female_name.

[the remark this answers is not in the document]
stevecotton: Seems to be 2 keys in most places, with `name` + `female_name` being slightly more common than `male_name` + `female_name`. Can't find any instances of all 3 being used together, although there are places where `plural_name` appears with 2 of the 3.

There's also `message` + `female_message` or `male_message` + `female_message`. There are surprising few gendered messages of any kind, with the tutorial using `message` and HTTT using `male_message`.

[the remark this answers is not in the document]
CelticMinstrel: Sorry, I meant which keys are supported, rather than which keys are used. I think you're correct that `male_name` isn't used very often though.

data/schema/units/single.cfg
```diff
@@ -8,6 +8,8 @@
 	{SIMPLE_KEY id string}
 	{SIMPLE_KEY gender gender,subst}
 	{SIMPLE_KEY name t_string}
+	{SIMPLE_KEY male_name t_string}
+	{SIMPLE_KEY female_name t_string}
```
stevecotton, line 12: @CelticMinstrel Are we ending up with two slightly different `[unit]` tags here? As in, there's a list of attributes which the results of `[store_unit]` could create, including `name` but not these two; and there's a superset which is the keys recognised when using `[unit]` to create a unit?

CelticMinstrel: I'm not sure that the description "two slightly different `[unit]` tags" fits, but you're correct: the keys accepted in `[unit]` are a superset of the keys actually serialized in a unit. That's not new – the placement keys are never serialized either, for example.

stevecotton: In that case, it looks good to me and the Flatpak build-failure isn't a blocker.

Pentarctagon:
Is this good to be merged then?

Slayer95:
Yes. Is there any review missing?

data/campaigns/The_South_Guard/scenarios/06a_Into_the_Depths.cfg
```diff
@@ -275,40 +275,20 @@
 #ifdef EASY
         {GENERIC_UNIT 2 (Revenant) 19 26} {GUARDIAN}
         {GENERIC_UNIT 2 (Revenant) 19 29} {GUARDIAN}
-        {NAMED_GENERIC_UNIT 2 (Dark Sorcerer) 20 27 (Mal Maul) (_"Mal Maul")} {GUARDIAN}
+        # po: Dark Sorcerer name
+        {NAMED_GENERIC_UNIT 2 (Dark Sorcerer) 20 27 (Mal Maul) (_"Mal Maul") FEMALE_NAME=_"female^Mal Maul"} {GUARDIAN}
```
CelticMinstrel, line 279: Does this po note get applied to both the male and female versions of the string? If not, the first of these probably needs to be split into multiple lines, something like this:

```suggestion
        # po: Dark Sorcerer name
        {NAMED_GENERIC_UNIT 2 (Dark Sorcerer) 20 27 (Mal Maul) (_"Mal Maul")
        # po: Dark Sorceress name
        FEMALE_NAME=_"female^Mal Maul"} {GUARDIAN}
```

data/campaigns/Under_the_Burning_Suns/scenarios/04_Descending_into_Darkness.cfg
```diff
@@ -556,11 +556,12 @@
         [/unit]
 
 #ifdef HARD
-        {NAMED_GENERIC_UNIT 2 (Naga Warrior) 16 34 () ( _ "Lake Naga")}
-        {NAMED_GENERIC_UNIT 2 (Naga Warrior) 16 35 () ( _ "Lake Naga")}
+        # po: name of an enemy unit in water, not a geographic feature
+        {NAMED_GENERIC_UNIT 2 (Naga Warrior) 16 34 () ( _ "Lake Naga") FEMALE_NAME=_"female^Lake Naga"}
```
CelticMinstrel, line 560: Same concern as above. If it's an issue, something like this:

```suggestion
        # po: name of an enemy unit in water, not a geographic feature
        {NAMED_GENERIC_UNIT 2 (Naga Warrior) 16 34 () ( _ "Lake Naga")
        # po: name of an enemy unit in water, not a geographic feature
        FEMALE_NAME=_"female^Lake Naga"}
```

CelticMinstrel:
For what it's worth, even if the above two things _are_ an issue, there's also the option of addressing it in wmlxgettext somehow.

Slayer95:
Right. The po comment only applied to the first string. I fixed it for these campaigns, but it could still be improved later in ``wmlxgettext``. I am not sure whether making ``#po:`` distributive over the next line is the best path forward, though. That will need further considerationg.

CelticMinstrel:
Yeah, I'm not sure either.

stevecotton:
I think we can merge now, or?

CelticMinstrel:
I have nothing else to add, so I think so.

Slayer95:
Could someone please confirm whether these commits are enough so that this excerpt from [SideWML's wiki](https://wiki.wesnoth.org/SideWML) holds true?

> no_leader: if 'no' (default), then keys describing a unit which will begin on the side's keep will be the remainder of the [side] tag, See [SingleUnitWML](https://wiki.wesnoth.org/SingleUnitWML)

That is, given the following

```
[side]
side=2
controller=ai
name=_"Male Leader"
female_name=_"Female Leader"
type=Dark Adept
[/side]
``` 

Will it be automatically accepted? This comments suggests it will:

https://github.com/wesnoth/wesnoth/blob/f8bd32eb157928cf3c0e4e4828eeaf1e2ec9befd/src/team.cpp#L53-L55

Pentarctagon:
It looks like that's the case. The best way to know though is to test it.

PR #7549, opened 2023-04-19
Fix modify_unit recall_cost indexing issue 

mjmagno:
Fixed issue #7159. Bug had to do with indexing into 0 based team vector with 1 based side_ member variable as speculated in the thread. Added method to return side_ - 1 and then used in apply_to recall_cost code.

### DIFF
--- a/data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/modify_unit_recall_cost.cfg
+++ b/data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/modify_unit_recall_cost.cfg
@@ -33,7 +33,7 @@
             [object]
                 [effect]
                     apply_to="recall_cost"
-                    increase=10
+                    increase=15
                 [/effect]
             [/object]
         [/modify_unit]
@@ -58,7 +58,7 @@
         [/store_unit]
 
         {ASSERT (
-            {VARIABLE_CONDITIONAL temp.recall_cost equals 20}
+            {VARIABLE_CONDITIONAL temp.recall_cost equals 25}
         )}
 
         # Recall Charlie next to Side 2 leader
--- a/data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/modify_unit_recall_cost.cfg
+++ b/data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/modify_unit_recall_cost.cfg
@@ -1,12 +1,12 @@
 #textdomain wesnoth-test
 # Actions:
 # Create Charlie, belonging to side 2, on the recall list
-# Increase Charlie's recall cost 
+# Increase Charlie's recall cost
 # Recall Charlie next to his side's leader.
 ##
 # Expected end state:
 # Charlie is on the map next to side 2's leader.
-# Charlie's recall cost has increased by 10 
+# Charlie's recall cost has increased by 10
 #####
 {GENERIC_UNIT_TEST "modify_unit_recall_cost" (
     [event]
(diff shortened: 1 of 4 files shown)


src/units/unit.cpp
```diff
@@ -2296,7 +2296,7 @@ void unit::apply_builtin_effect(std::string apply_to, const config& effect)
 	} else if(apply_to == "recall_cost") {
 		const std::string& increase = effect["increase"];
 		const std::string& set = effect["set"];
-		const int recall_cost = recall_cost_ < 0 ? resources::gameboard->teams().at(side_).recall_cost() : recall_cost_;
+		const int recall_cost = recall_cost_ < 0 ? resources::gameboard->teams().at(side_to_index()).recall_cost() : recall_cost_;
```
Pentarctagon, line 2299: Should rather be `resources::gameboard->get_team(side_)`.

Vultraz: Yes. Use that instead of adding a `side_to_index` function.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/modify_unit_recall_cost.cfg
```diff
@@ -0,0 +1,83 @@
+#textdomain wesnoth-test
+# Actions:
+# Create Charlie, belonging to side 2, on the recall list
+# Increase Charlie's recall cost
```
stevecotton, line 4: The CI check is complaining about two things in this file: it has CRLF line endings, and some lines have trailing whitespace.

Running `data/tools/wmlindent data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/modify_unit_recall_cost.cfg` will fix both issues.

gfgtdf:
The code should also check whether resources::gameboard is nullptr before doing that, I'm not sure whether that code can also be run in the scenario editor but it better to be safe here.

Pentarctagon:
If it's null, what should the default be? The usual 20?

stevecotton:
Given that the existing code also assumed that `resources::gameboard` is safe to access at this point, it's not introducing a regression. While fixing it would be nice to have, I feel it's not something to ask a new contributor for.

mjmagno:
The change doesn't sound too difficult, I don't mind making it as long as I know what the default should be

gfgtdf:
> If it's null, what should the default be? The usual 20?

I'd say so, but i think it's only null during the editior so the exact value is probably not that important, it's only to be extra safe against crashes.

> it's not introducing a regression. While fixing it would be nice to have, I feel it's not something to ask a new contributor for.

I know it's not a regression, but it would be nice if this was fixed too. it's not a requirement for merging this though

Pentarctagon:
I think this is good to be merged?

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/modify_unit_recall_cost.cfg
```diff
@@ -0,0 +1,83 @@
...
+        # Check that Charlie's recall cost has increased
+        [store_unit]
+            [filter]
+                id=Charlie
+            [/filter]
+            variable=temp
+        [/store_unit]
+
+        {ASSERT (
+            {VARIABLE_CONDITIONAL temp.recall_cost equals 20}
```
stevecotton, line 61: Starting with an initial cost of 10 and then increasing by 10 means the expected value is the same as the default value. Please change one of those numbers, so that this is testing that the unit has an individual recall cost.

mjmagno: Got it, good catch

stevecotton:
The flatpak build is a separate failure, unrelated to this PR.

CelticMinstrel:
Shouldn't the default have been game_config::recall_cost?

PR #7552, opened 2023-04-19
WoF: Patch Assertion by Adding Dummy Player Side to StoryText-Cutscenes

### LINKED ISSUE #7550: [Campaigns][Bug] WoF unplayable after Scenario 2
Something causes the client to assert to desktop after loading the third scenario

image

Wesnoth 1.17.15+dev
Github
Windows 10 (msys 2)

### More Context:

https://forums.wesnoth.org/viewtopic.php?p=681121#p681121

### Extra Remarks:

before adding new features. playtest your code...
Sincerely, LK

knyghtmare:
Closes #7550

### DIFF
--- a/data/campaigns/Winds_of_Fate/scenarios/02x_Victory_Feast.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/02x_Victory_Feast.cfg
@@ -143,6 +143,7 @@ I shall win the new flight myself, so that with it the Domain of Vladnir will ho
         name=prestart
         [endlevel]
             result=victory
+            carryover_report=no
             linger_mode=no
             replay_save=no
         [/endlevel]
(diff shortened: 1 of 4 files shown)


CelticMinstrel:
Um… maybe we should fix the assert instead of working around it?

knyghtmare:
> Um… maybe we should fix the assert instead of working around it?

I think scenario files without SideWML should be discouraged so adding dummy player side to ensure campaign progression is the way to go. 

The other mainline example "LoW" also has dummy player side(s) to ensure campaign progression.

gfgtdf:
> I think scenario files without SideWML should be discouraged

The comment in the wml code suggests that it solely to fix an assertion though, in which case i think it's better to merge #7555

CelticMinstrel:
> I think scenario files without SideWML should be discouraged

Sure. But they also shouldn't crash the engine.

It's correct to merge this PR, but we need to fix the assert too.

knyghtmare:
> It's correct to merge this PR, but we need to fix the assert too.

I fixed 2 bugs in this PR so yeah

soliton-:
I disagree that adding one dummy side definition where none was needed before is desirable.

If we really wanted to require story only scenarios to include useless side definitions then 1) that needs to be done in all mainline campaigns, 2) tooling like wmllint needs to at a minimum detect this.

Really though that sounds like a change just to annoy UMC maintainers.

knyghtmare:
> Really though that sounds like a change just to annoy UMC maintainers.

You twist my beneficial intent into something the exact opposite to irk the UMC community which I represent. very well...

PR #7555, opened 2023-04-19
Return early if there are no sides

soliton-:
Avoids calling modulo with 0.

### DIFF
--- a/src/playsingle_controller.cpp
+++ b/src/playsingle_controller.cpp
@@ -195,7 +195,7 @@ void playsingle_controller::skip_empty_sides(int& side_num)
 	const int sides = static_cast<int>(get_teams().size());
 	if(sides == 0) return;
 	const int max = side_num + sides;
-	while (gamestate().board_.get_team(modulo(side_num, sides)).is_empty()) {
+	while (gamestate().board_.get_team(modulo(side_num, sides, 1)).is_empty()) {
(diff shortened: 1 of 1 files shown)


soliton-:
Should fix #7550.

gfgtdf:
With this it looks like this function does nothing when there are no sides and throws when there is for example only one side with `controller=null`, which makes it a bit inconsistent. Still, since it fixes #7550 it still good to go in, but i should probably improve it later, (i can't right now since I'm not in my main pc)

soliton-:
> Was this tested to make sure that the campaigns no longer get the assert (or some other issue)?

No, if 40fbf95 is completely wrong then this does not help. I won't be able to test this anytime soon.

gfgtdf:
Technically that commit is a bit wrong since it might result in an out of range side num like i said will fix later.
(That function is usually used  the end of the side where this is intentionally possible to indicate new turns)

Pentarctagon:
I'd like to tag 1.17.16 this weekend to push a fix for #7550 out. Is this PR good to be merged, or does it need further changes?

soliton-:
From my side it's good to go. @gfgtdf seems to approve? Please do it through the github UI.

gfgtdf (approved):
Looks good to me, but i can it test it here.

PR #7559, opened 2023-04-21
Made flatpak ci job upload bundle too

loonycyborg:
This adds new CI artifact containing file wesnoth.flatpak. It can be installed and run like this:
`flatpak --user install wesnoth.flatpak`
`flatpak run --branch=ci org.wesnoth.Wesnoth`
After done with with testing it can be removed to save disk space:
`flatpak uninstall org.wesnoth.Wesnoth` and choose install from branch `ci` if there are other installs.


soliton- (approved):
Looks good to me and works. Just a little weird that the version says "Modified" instead of "Clean".

soliton-:
jq outputs to a new file.

loonycyborg:
Probably that's something about the way checkout is made then.

soliton-:
Yeah, it looks like the check does `git status --untracked-files=normal --porcelain` and Clean means no output. So it might be the new json file.

soliton-:
So we could put the file elsewhere or pick one that is in .gitignore. Is the filename important?

PR #7561, opened 2023-04-22
Improvements to wmltools3

Slayer95:
Split out from #7540

### DIFF
--- a/data/tools/wesnoth/wmltools3.py
+++ b/data/tools/wesnoth/wmltools3.py
@@ -969,9 +969,10 @@ class CrossRef:
                 except UnicodeDecodeError as e:
                     pass # to not have the invalid UTF-8 file warning printed twice
         # Check whether each namespace has a defined export property
-        for namespace in self.dirpath:
-            if namespace not in self.properties or "export" not in self.properties[namespace]:
-                print("warning: %s has no export property" % namespace)
+        if self.warnlevel >= 1:
+            for namespace in self.dirpath:
+                if namespace not in self.properties or "export" not in self.properties[namespace]:
+                    print("warning: %s has no export property" % namespace)
     def exports(self, namespace):
         return namespace in self.properties and self.properties[namespace].get("export") == "yes"
     def subtract(self, filelist):
--- a/data/tools/wesnoth/wmltools3.py
+++ b/data/tools/wesnoth/wmltools3.py
@@ -300,8 +300,6 @@ def parse_macroref(start, line):
         # stop matching on the first one, because the argument value might contain one too
         if re.match(r"^([A-Z0-9_]+?)=", arg):
             opt_arg, arg = arg.split("=", 1)
-        if arg.startswith('"') and arg.endswith('"'):
-            arg = arg[1:-1].strip()
         if opt_arg:
             optional_args[opt_arg] = arg
             opt_arg = ""
--- a/data/tools/wesnoth/wmltools3.py
+++ b/data/tools/wesnoth/wmltools3.py
@@ -576,7 +576,7 @@ class CrossRef:
         if self.exports(defn.namespace):
             # Macros and resources in subtrees with export=yes are global
             return True
-        elif not self.filelist.neighbors(defn.filename, fn):
+        elif defn.filename != fn and not self.filelist.neighbors(defn.filename, fn):
             # Otherwise, must be in the same subtree.
             return False
         else:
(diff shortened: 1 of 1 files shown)


data/tools/wesnoth/wmltools3.py
```diff
@@ -286,66 +294,126 @@ def isresource(filename):
     return ext and ext[1:] in resource_extensions
 
 def parse_macroref(start, line):
-    def handle_argument():
-        nonlocal opt_arg
-        nonlocal arg
-        nonlocal optional_args
-        nonlocal args
+    def handle_argument(buffer):
```
Elvish-Hunter, line 297: Would it be possible to declare both `optional_args` and `args` as `nonlocal`? It won't have any practical effect in this specific case because Python handles mutable types by reference, but it'll make more evident that you're referring to variables defined outside this function.

data/tools/wesnoth/wmltools3.py
```diff
@@ -286,66 +294,126 @@ def isresource(filename):
...
+        opt_arg = ""
+
+        arg = "".join(buffer)
+        if arg[0:1].isspace():
```
Elvish-Hunter, line 301: With slicing, `[0:1]` is the same as writing `[0]` because the last index is not included.

data/tools/wesnoth/wmltools3.py
```diff
@@ -286,66 +294,126 @@ def isresource(filename):
...
-                arg += line[i]
-            brackdepth += 1
-        elif line[i] == "}":
-            brackdepth -= 1
-            if brackdepth == 0:
-                if not line[i-1].isspace():
-                    handle_argument()
+        added_arg = False
+        if depth[QUOTE] > 0:
+            if line[i] == QUOTE and line[i+1:i+2] != QUOTE:
```
Elvish-Hunter, line 355: Again slicing: `[i+1:i+2]` is the same as `[i+1]`.

data/tools/wesnoth/wmltools3.py
```diff
@@ -705,15 +778,22 @@ def scan_for_definitions(self, namespace, filename):
...
-                            here.optional_args.append(line.strip().split()[1])
+                            here._optional_args.append([line.strip().split()[1],""])
                         elif line.strip()[0] != "#":
                             state = States.MACRO_BODY
-                    elif state == States.MACRO_OPTIONAL_ARGUMENT and "#endarg" in line:
+                    elif state == States.MACRO_OPTIONAL_ARGUMENT and not "#endarg" in line:
+                        here._optional_args[-1][1] += line
+                    elif state == States.MACRO_OPTIONAL_ARGUMENT:
+                        end_arg_index = line.index("#endarg")
+                        here._optional_args[-1][1] += line[0:end_arg_index]
```
Elvish-Hunter, line 795: It can be really confusing to have two variables called `_optional_args` and `optional_args`, with the only difference being a leading underscore. Please consider making the two names a bit more different.

data/tools/wesnoth/wmltools3.py
```diff
@@ -286,66 +294,126 @@ def isresource(filename):
...
+    depth = {
+        EQUALS: 0,
+        OPEN_BRACE: 0,
+        OPEN_PARENS: 0,
+        QUOTE: 0,
+    }
+    wrapper_stack = []
+    prev = ""
+
+    def close_token(token):
```
Elvish-Hunter, line 331: Can you add a small comment to make clear at a quick glance when it's better to use `close_token()` or when `close_if_token()` should be preferred?

Elvish-Hunter (changes requested):
The patch looks very good; I'm marking it as "Request changes" just to allow fixing the little issues I found. Once they're fixed, this can be merged.

data/tools/wesnoth/wmltools3.py
Slayer95: Alright

data/tools/wesnoth/wmltools3.py
Slayer95: Unfortunately, ``[0]`` raises out of bounds error if the argument is empty, e.g. in ``{MACRO arg1()}``

data/tools/wesnoth/wmltools3.py
Slayer95: Similarly, out of bounds if this is the end of the input.

data/tools/wesnoth/wmltools3.py
Slayer95: True. Do you think ``_raw_optional_args`` would be appropriate?

data/tools/wesnoth/wmltools3.py
Slayer95: Fair enough

Slayer95:
Updated per your review, @Elvish-Hunter . I have also fixed L404-405 to properly handle the empty arg case I mentioned.

data/tools/wesnoth/wmltools3.py
Elvish-Hunter: You're right: slicing, unlike indexing, actually returns an empty string instead of raising an error. I guess I forgot because I don't rely on this specific behavior that often...

Elvish-Hunter:
Now that everything has been fixed, if you don't want to add anything else I can merge it.

Elvish-Hunter (approved):

Slayer95:
Please, go ahead, @Elvish-Hunter , and thanks!

Slayer95:
At last it built. It was timing out in a random WML test earlier...

data/tools/wesnoth/wmltools3.py
```diff
@@ -286,66 +294,138 @@ def isresource(filename):
...
+                close_if_token(EQUALS)
+                added_arg = handle_argument(buffer)
+            buffer.append(line[i])
+        else:
+            buffer.append(line[i])
+
+        prev_added_arg = added_arg
+
+    args.pop(0)
+    return (args, optional_args, depth[OPEN_BRACE] > 0)
```
stevecotton, line 428: We're missing the corresponding change to wmllint, and it doesn't seem to be in your open PRs / open branches. Please would you share it?

Slayer95: Sorry for the trouble, @stevecotton . Patch in #7662.

PR #7562, opened 2023-04-23
[1.17] - make engine-bobbing sprites use the correct primary frame

### LINKED ISSUE #7551: [BUG][Units] Great SeaHorse Standing Animation prevents Poison Status effect Green Hue from being Shown
image

Screenshot clarifies what I mean

doofus-01:
To keep the ellipses and status indicators from bobbing/drifting with the base frame, I'd put the body sprite in an alternate frame, but did not think to check for status effect color-shifts like poison or slows.  This is a simple fix for that, only question is if I missed some unit.

fixes #7551


doofus-01:
Is the flatpak failure significant?

Pentarctagon:
It's not.

PR #7563, opened 2023-04-23
filter true damage and attacks in [filter_attack/weapon] except in [damage] and [attacks]

newfrenchy83:
Now that [event][filter_attack] can manage the activity of special weapons, the question arises of the possibility or not of filtering the damage and the number of attacks taking into account the modifications made by [damage] and [attacks].

In this PR, the real damages and attacks are taken into account except for the special ones in order to avoid problems of infinite loops.

I admit, it was also while looking for this solution for damage_type that I came to propose it for two filterable parameters in special weapons, which had led to filtering only the unchanged parameters.

### DIFF
--- a/changelog_entries/change_filtering_filter_weapon.md
+++ b/changelog_entries/change_filtering_filter_weapon.md
@@ -1,2 +1,2 @@
  ### WML Engine
-   * filter "damage" and filter "number" in [filter_attack/weapon] matche with real damage or attack number values except respectively in [damage] and [attacks]
+   *  "Add weapon filter keys special_damage and special_number, which account for special [damage] or [attacks]."
--- a/changelog_entries/change_filtering_filter_weapon.md
+++ b/changelog_entries/change_filtering_filter_weapon.md
@@ -1,0 +1,2 @@
+ ### WML Engine
+   * filter "damage" and filter "number" in [filter_attack/weapon] matche with real damage or attack number values except respectively in [damage] and [attacks]
--- a/data/schema/filters/weapon.cfg
+++ b/data/schema/filters/weapon.cfg
@@ -18,8 +18,8 @@
 	{SIMPLE_KEY accuracy s_range_list}
 	{SIMPLE_KEY movement_used s_range_list}
 	{SIMPLE_KEY attacks_used s_range_list}
-	{SIMPLE_KEY effective_damage s_range_list}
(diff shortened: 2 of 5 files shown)


gfgtdf:
I only looked at the code briefly, but I think it could still be infinite recursions of one for example has a damage special that depends on the number of attacks and an attack special that depends on the amount of damage though

newfrenchy83:
> I only looked at the code briefly, but I think it could still be infinite recursions of one for example has a damage special that depends on the number of attacks and an attack special that depends on the amount of damage though

i choose to use new attribute instead for don't disorient wml devellopper because [damage] and [attacks] already exist

changelog_entries/change_filtering_filter_weapon.md
```diff
@@ -0,0 +1,2 @@
+ ### WML Engine
+   * filter "damage" and filter "number" in [filter_attack/weapon] matche with real damage or attack number values except respectively in [damage] and [attacks]
```
stevecotton, line 2: IIUC, this was the first implementation, but not the current state of the PR.

I think it should be "Add weapon filter keys `effective_damage` and `effective_number`, which account for time of day, leadership, being slowed, etc."

newfrenchy83: except what effective_damage don't account time of day or leadership because modified_damage calculation was before [leadership] or Tod, and slow, imut rename filter like special_damage or abandon this idea(this implementation is for integrate [damage] or [attacks] but that all)

stevecotton: Ah, I misunderstood what it would be used for then.

stevecotton:
The name `specials` gets used a lot, I feel it's too generic to use here. However, I'm don't understand what the purpose of this filter is, or what it would be used for.

stevecotton:
Even when `[swarm]` is applied directly as a weapon special, it does not affect `special_number`. For example, this assert fails when added to data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/swarm_disables_upgrades.cfg
```
        # At this point, without [swarm] bob would have 10 + 13 = 23 attacks.
        {ASSERT (
            [have_unit]
                id=bob
                [has_attack]
                    special_number=1
                [/has_attack]
            [/have_unit]
        )}
```

newfrenchy83:
i will abandon this idea and close this PR

Issue #7507, opened 2023-04-01
Add a "simple planning mode" to replace DSU


gfgtdf:
### Describe the desired feature

DSU has a few drawbacks:
1) Its complicates the mp sync mechanics quite a lot, since the commit vision command reevaluated the undo stack, the game now has to make sure the undo stack is the same on all clients.
2) We have to make sure dsu is disabled when making a ai handle a side since the ai cannot handle it.
3) It might lead break scenarios when a sighted even is now fired much later that expected.
4) While it guards the player from missclicks in some cases, its still very much possible that move cannot be undone.

Planning mode, which is supposed to replace DSU, is considered to be too complicated by some players, so the proposal is to add a "plannign mode light", it should:
1) Not draw many arrows
2) Not allow planning of multiple-turn moves.
3) Still show a clear indicator, that the current gamestate is only the "planned gamestate" is that's the case.
4) Probably autoamticially end itself when attacking a unit or recruiting a unit.

knyghtmare:
What is DSU? 😨

gfgtdf:
"Delayed Shroud Updates"

gfgtdf:
> I haven't looked at the DSU code, but would it simplify the code while still being a reasonable UI to force an immediate shroud redraw if a WML or Lua sighted event should be triggered?

I don't think this would make it simpler, Other events might still depend on which tile is visible, (filter_vision) and we still have the same problem. Im currently thinking making a special exception for sighted event would just make it it even more complicated.

Furthermore, this would allow players to "fish" for sighted events  (not a big problem, but should be noted)

> shroud redraw

I think you know but just to make it sure: This is not about the drawing code, DSU is about on purpose not uncovering tiles to allow undoing of moves.

PR #7566, opened 2023-04-23
Lua API to the undo system

CelticMinstrel:
This fixes the issue of the new events API not allowing a way to make events undoable, beyond calling into the corresponding WML actions.

The API additions are:

* `wesnoth.game_events.set_undoable` sets whether or not the current event can be undone.
* `wesnoth.game_events.add_undo_actions` adds actions to be performed when the current event is undone. Can be either a Lua function or a table of ActionWML.
* `wesnoth.game_events.add_repeating` takes an optional third argument which a function to call on undo, which automatically marks the event undoable. (Or it was supposed to, I think I actually missed that.)

There are a couple of things missing – the above parenthetical, and what to do when serializing an undo stack that contains a Lua function. Obviously it's not going to work, but we can't just ignore it, as I imagine it could cause strange things to happen when loading a saved game.

This was tested and working prior to rebasing onto the latest master. I don't have any reason to think the rebase would've broken anything though.

### DIFF
--- a/data/lua/diversion.lua
+++ b/data/lua/diversion.lua
@@ -1,7 +1,6 @@
 
 local _ = wesnoth.textdomain 'wesnoth-help'
 local T = wml.tag
-local on_event = wesnoth.require("on_event")
 
 local u_pos_filter = function(u_id)
 
@@ -37,8 +36,7 @@ local u_pos_filter = function(u_id)
         end
 end
 
-
-local status_anim_update = function(is_undo)
+local function status_anim_update(is_undo)
 
         local ec = wesnoth.current.event_context
         local changed_something  = false
@@ -88,20 +86,16 @@ local status_anim_update = function(is_undo)
                     }
                 end
         end
-        if changed_something and not is_undo then
-                wesnoth.wml_actions.on_undo {
-                        wml.tag.on_undo_diversion {
-                        }
-                }
+        if not is_undo then
+                wesnoth.game_events.set_undoable(true)
+                if changed_something then
+                        wesnoth.game_events.add_undo_actions(function(_)
+                                status_anim_update(true)
+                        end)
+                end
         end
 end
 
-function wesnoth.wml_actions.on_undo_diversion(cfg)
-        status_anim_update(true)
-end
-
-on_event("moveto, die, recruit, recall", function()
-        status_anim_update()
-
+wesnoth.game_events.add_repeating("moveto, die, recruit, recall", function(_)
+        status_anim_update(false)
 end)
-
--- a/data/lua/diversion.lua
+++ b/data/lua/diversion.lua
@@ -87,9 +87,9 @@ local function status_anim_update(is_undo)
                 end
         end
         if not is_undo then
-                wesnoth.game_events.set_undoable(true)
+                wesnoth.experimental.game_events.set_undoable(true)
                 if changed_something then
-                        wesnoth.game_events.add_undo_actions(function(_)
+                        wesnoth.experimental.game_events.add_undo_actions(function(_)
                                 status_anim_update(true)
                         end)
                 end
--- a/src/synced_context.cpp
+++ b/src/synced_context.cpp
@@ -358,6 +358,16 @@ void synced_context::add_undo_commands(const config& commands, const game_events
 	undo_commands_.emplace_front(commands, ctx);
 }
 
+void synced_context::add_undo_commands(int idx, const game_events::queued_event& ctx)
+{
+	undo_commands_.emplace_front(idx, ctx);
+}
+
+void synced_context::add_undo_commands(int idx, const config& args, const game_events::queued_event& ctx)
+{
+	undo_commands_.emplace_front(idx, args, ctx);
+}
+
 set_scontext_synced_base::set_scontext_synced_base()
 	: new_rng_(synced_context::get_rng_for_action())
 	, old_rng_(randomness::generator)
--- a/data/lua/core/game_events.lua
+++ b/data/lua/core/game_events.lua
@@ -1,0 +1,15 @@
+--[========[Game Event Functions]========]
+
+if wesnoth.kernel_type() == "Game Lua Kernel" then
+    -- The undo API is still experimental, so move those functions
+
+    -- Make sure wesnoth.experimental.game_events actually exists
+    -- It's done this way so it doesn't break if we later need to add things here from C++
+    wesnoth.experimental = wesnoth.experimental or {}
+    wesnoth.experimental.game_events = wesnoth.experimental.game_events or {}
+
+    wesnoth.experimental.game_events.set_undoable = wesnoth.game_events.set_undoable
+    wesnoth.experimental.game_events.add_undo_actions = wesnoth.game_events.add_undo_actions
+    wesnoth.game_events.set_undoable = nil
+    wesnoth.game_events.add_undo_actions = nil
+end
--- a/data/lua/on_event.lua
+++ b/data/lua/on_event.lua
@@ -51,5 +51,5 @@ local function on_event(eventname, priority, fcn)
 	end
 end
 
-core_on_event = on_event
+core_on_event = wesnoth.deprecate_api('on_event', 'wesnoth.game_events.add_repeating', 1, nil, on_event)
 return on_event
--- a/data/lua/on_event.lua
+++ b/data/lua/on_event.lua
@@ -51,5 +51,5 @@ local function on_event(eventname, priority, fcn)
 	end
 end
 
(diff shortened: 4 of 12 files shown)


gfgtdf:
> and what to do when serializing an undo stack that contains a Lua function

I don't see a good way to solve this actually (remember that the default Lua bytecode serializer is.not an option for security reasons). So i really think it would be better not to add a broken function (allowing Lua function handlers there) when we dont even have an idea on how it can be fixed

CelticMinstrel:
I wasn't suggesting using the Lua bytecode serializer. I just mean we need to do _something_. It might be just a warning message. It might be clipping the undo stack on serialization to exclude frames that are unserializable. It migh mean adding some way for the specific function to have a way to opt in for a way to make it serializable (not using bytecode). I'm not quite sure. But we can't just ignore it entirely.

gfgtdf:
But i cannot think of any cases where would want to add a Lua function undo handler if it always breaks when saving a game, so why would we have it? In event it kinda makes sense because of events that are added by Lua code in theloading code, but i see no reason to ever use it in undo handlers.


> It might be clipping the undo stack on serialization to exclude frames that are unserializable

This is also not that easy because of DSU.

CelticMinstrel:
> it doesn't make sense to add Lua handlers without having fixed this

I disagree. While the serializability is something that we definitely need to handle in **some** way, I don't see the undo stack as something that absolutely needs to be serialized. The most common use-case for undo is "I just did something and now I want to undo it", not "I returned to my saved game from 3 days ago and decided I actually want to undo my last move". Allowing the second case is "nice-to-have", but not "essential".

> This is also not that easy because of DSU.

How so? Personally I feel that clipping the undo stack is the best way to handle this.

gfgtdf:
> How so? Personally I feel that clipping the undo stack is the best way to handle this.

Because with DSU active, you cannot simply clip the undo stacks since it contains information about moves that we did not yet clear the shroud for. So clearing/clipping the u do stack must happen synced since it might clear the delayed shroud and also fire sighted events (which is why I really want to remove DSU)  see also #7507

src/synced_context.cpp
```diff
@@ -349,6 +349,16 @@ void synced_context::add_undo_commands(const config& commands, const game_events
 	undo_commands_.emplace_front(commands, ctx);
 }
 
+void synced_context::add_undo_commands(int idx, const game_events::queued_event& ctx)
```
Vultraz, line 352: I think this should be a variadic template w/ forwarding.

CelticMinstrel:
> Because with DSU active, you cannot simply clip the undo stacks since it contains information about moves that we did not yet clear the shroud for. So clearing/clipping the u do stack must happen synced since it might clear the delayed shroud and also fire sighted events

I suppose this would mean that adding anything to the undo stack that's not serializable would need to clear the delayed shroud right then and there. Then clipping the undo stack could still be done on save only (while maintaining the full stack in memory).

gfgtdf:
> I suppose this would mean that adding anything to the undo stack that's not serializable would need to clear the delayed shroud right then and there

Hmm yes this is probably the simlest way to handle it.

stevecotton:
Where are the tests?

CelticMinstrel:
I actually don't think this is testable. How can you undo in a test?

stevecotton:
The reason for asking for a test is that I'm more interested in reviewing "what code will UMC need in order to use this?" rather than "what does the implementation look like?".

Although it can't be automated, it could use a manual test.
* Have a `moveto` that does non-lethal damage, and on undo put the unit back to its old hp. Requires storing the info about how many HP the unit had, as there are multiple initial numbers that would result in 1 hp after non-lethal damage.
* Have a `moveto` that advances the unit, with an undo that returns it back to it pre-levelling state.

data/lua/diversion.lua
```diff
@@ -88,20 +86,16 @@ local status_anim_update = function(is_undo)
...
-        if changed_something and not is_undo then
-                wesnoth.wml_actions.on_undo {
-                        wml.tag.on_undo_diversion {
-                        }
-                }
+        if not is_undo then
+                wesnoth.game_events.set_undoable(true)
+                if changed_something then
+                        wesnoth.game_events.add_undo_actions(function(_)
+                                status_anim_update(true)
```
stevecotton, line 93: In what order does the undo happen? From this code, it seems the units are already in position before whatever triggered this, so step 5 comes before step 6:

1. player moves unit with `diversion`
2. position of units update on map
3. moveto event triggers `status_anim_update` to recalculate the effects
4. player presses "undo"
5. position of units update on map
6. undo event triggers `status_anim_update` to recalculate the effects

However, consider an event that removes a unit from the map, for example a scenario where the player must move units to an exit point, and then they're immediately moved to the player's recall list. How can that be undone, unless the code for the undo runs and puts the unit back on the map before the movement is reversed? So in that case, step 6 must come before step 5.

CelticMinstrel: I think the code itself would just need to be written to handle it somehow. Like, in your example, it stores the ID and uses it to look up the unit on the recall list and put it back on the map.

gfgtdf: I think in theory it makes sense if the order of execution when undoing is the reverse of the order of execution during the original move (which would mean the when u doing first the move to event is undone and then the actual move). However I'm unsure how feasible this is, especially when there are also events like enter_hex and exit_hex that happen during the move.

CelticMinstrel: That does kinda make sense. It also breaks this diversion implementation. How would diversion work if the undo actions were triggered before the units were moved back to their prior configuration?

gfgtdf: I don't really know much about the animation code, so i am not sure what the code does currently, but the general idea is is to remember the previous state/changes and pass it to the undo handler.

Pentarctagon:
Any update on this? The 1.18 beta releases will be starting in a few months (first beta will be tagged on Oct 15) which is usually when we tell UMC authors it's a good time to start porting add-ons, and it'd be good for this to be resolved by that point.

CelticMinstrel:
It might be hard with the summer heat, but I'll try to find time to work on this.

CelticMinstrel:
Any issues with merging the current state of this, as an experimental API? There are still three or four things left to do that have been discussed in this thread, but I think it's quite helpful for it to at least be in the game, as it was a significant missing feature in the new events API that was added.

Right now, the WML undo API is calling into an experimental API, but I think that's fine, as there isn't really any difference between what the previous C++ tag did and what the new Lua tag does; one of the reasons this is unfinished is because the Lua case _needs_ to do something extra because of the unserializability issue.

The remaining TODO list for this feature seems to be:

*A test case
*`wesnoth.game_events.add_repeating` needs to be able to specify whether to allow undo
*undo stack needs to be clipped(?) if it contains something unserializable (a Lua function reference)

For the 3rd one I think we haven't even worked out exactly what that means. In single-player I'd assume you clip it only when saving the game. I'm not sure how it'll work in multiplayer, but it'll probably be more complicated than that.

Pentarctagon:
What would happen if the undo stack does contain a lua function reference?

CelticMinstrel:
> What would happen if the undo stack does contain a lua function reference?
> 
> edit- Mainly I'm wondering whether it's possible for someone to put something with a lua function reference onto the undo stack accidentally. I'm assuming that the WML undo API wouldn't cause that to happen.

It would serialize as a "blank" entry in the undo stack. I'm not sure what the implications of that are, exactly; is the undo stack serialized into saved games? If so, undoing that particular action would not work after reloading.

And yeah, the WML API only puts WML undo actions onto the stack, so there isn't a way to get a Lua function reference onto the undo stack without directly calling the experimental Lua API.

Pentarctagon:
Sounds fine to me then.

gfgtdf:
1) The third optional undofunction parameter of add_repeating seems to clash with the third optional priority parameter (and i still think that the priority parameter  should by at position 2, see https://github.com/wesnoth/wesnoth/issues/7582#issuecomment-1542558866 , although this is not really the topic of the pr)


> If so, undoing that particular action would not work after reloading.

From a quick look at the code it seems, like undoing is still possible but that particular undo function will just not be executed, which can corrupt the gamestate or cause OOS in replays when a players attempts to undo such an action after the game was reloaded (because it will still be undone but only partly).

CelticMinstrel:
> From a quick look at the code it seems, like undoing is still possible but that particular undo function will just not be executed, which can corrupt the gamestate or cause OOS in replays when a players attempts to undo such an action after the game was reloaded (because it will still be undone but only partly).

Yeah, pretty much.

> 1. The third optional undofunction parameter of add_repeating seems to clash with the third optional priority parameter (and i still think that the priority parameter  should by at position 2, see

I'll look… it was supposed to be a 4th parameter in order to not clash, but I likely missed something in the merge.

PR #7570, opened 2023-04-25
wmlxgettext: Handle multiple strings in single line better

Slayer95:
This provides enough flexibility to document the different meanings of
    multiple strings in a single line, e.g. in macro calls, taking advantage of
    the design goal of #po-override (only one override allowed per sentence),
    as documented in
    https://wmlxgettext-unoff.readthedocs.io/en/latest/enduser/wml.html#po-override-override-info

- ``#po``, ``#po-override`` directives are no longer consumed by untranslatable strings.
- Each translatable string consumes a single ``#po-override`` directive.

Given the following example:
```php
#textdomain wesnoth-test

# po: Unit name
# po: This is a unit tasked with guarding a place
# po-override: Male unit name
# po-override: Female unit name
{NAMED_NOTRAIT_UNIT 2 ("Silver Mage") 13 7 ("") ( _ "Guardian") FEMALE_NAME=_"female^Guardian"}
```

Old output:
```pot
#: input/test.cfg:7
msgid "Guardian"
msgstr ""

#: input/test.cfg:7
msgid "female^Guardian"
msgstr ""
```

New output:
```pot
#. Male unit name
#. Unit name
#. This is a unit tasked with guarding a place
#: input/test.cfg:7
msgid "Guardian"
msgstr ""

#. Female unit name
#: input/test.cfg:7
msgid "female^Guardian"
msgstr ""
```

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
@@ -544,13 +544,13 @@
                     message= _ "Yes, yes it is. Nobody’ll support ya on yer crazy path to depravity."
                 [/message]
 
-                # po: Hint, that the loyal units stay behind
                 [message]
                     side=1
                     race=human
                     [not]
                         id=Malin Keshar,Dela Keshar,Drogan
                     [/not]
+                    # po: Hint, that the loyal units stay behind
                     message= _ "We should all be glad to see you gone, necromancer. I detest having fought by your side."
                 [/message]
 
@@ -615,13 +615,13 @@
                     message= _ "If ya go out on a quest to seek depravity, none o’ us will support ya."
                 [/message]
 
-                # po: Hint, that the loyal units stay behind
                 [message]
                     side=1
                     race=human
                     [not]
                         id=Malin Keshar,Dela Keshar,Drogan
                     [/not]
+                    # po: Hint, that the loyal units stay behind
                     message= _ "Better to die than to ally ourselves with a necromancer."
                 [/message]
 
--- a/data/campaigns/Descent_Into_Darkness/scenarios/03_A_Haunting_in_Winter.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/03_A_Haunting_in_Winter.cfg
@@ -1233,9 +1233,9 @@ All sense of time and being are entombed within darkness. How long has passed si
     [event]
         name=rat eating4
 
-        #po: Malin Keshar breaks the fourth wall here and addresses the player, whom the 'you' refers to.
         [message]
             speaker=Malin Keshar
+            #po: Malin Keshar breaks the fourth wall here and addresses the player, whom the 'you' refers to.
             message= _ "You. You there behind the screen. You are the one controlling me. Why are you doing this? What did I do to deserve this torment? Please, I beg of you. End me now. Release me from my misery!"
         [/message]
     [/event]
--- a/data/campaigns/Descent_Into_Darkness/scenarios/07b_A_Small_Favor2.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/07b_A_Small_Favor2.cfg
@@ -258,11 +258,9 @@
                 [/not]
             [/filter_location]
             [command]
-                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type"
-                # po: Unit name
-                $loc.x $loc.y () (_ "Guard")
-                # po: Unit name
-                FEMALE_NAME=_"female^Guard"} {GUARDIAN}
+                # po-override: Unit name
+                # po-override: Unit name
+                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type" $loc.x $loc.y () (_ "Guard") FEMALE_NAME=_"female^Guard"} {GUARDIAN}
             [/command]
         [/random_placement]
 
--- a/data/campaigns/Descent_Into_Darkness/scenarios/07b_A_Small_Favor2.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/07b_A_Small_Favor2.cfg
@@ -258,8 +258,9 @@
                 [/not]
             [/filter_location]
             [command]
+                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type"
                 # po: Unit name
-                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type" $loc.x $loc.y () (_ "Guard")
+                $loc.x $loc.y () (_ "Guard")
                 # po: Unit name
                 FEMALE_NAME=_"female^Guard"} {GUARDIAN}
             [/command]
--- a/data/campaigns/Descent_Into_Darkness/scenarios/07c_A_Small_Favor3.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/07c_A_Small_Favor3.cfg
@@ -45,8 +45,9 @@
         hidden=yes
         color=purple
 
+        {NAMED_NOTRAIT_UNIT 2 (Silver Mage) 11 9 ("")
         # po: Book guards
-        {NAMED_NOTRAIT_UNIT 2 (Silver Mage) 11 9 ("") ( _ "Guardian")
+        ( _ "Guardian")
         # po: Book guards
         FEMALE_NAME=_"female^Guardian"} {GUARDIAN}
         {NAMED_NOTRAIT_UNIT 2 (Silver Mage) 13 7 ("") ( _ "Guardian") FEMALE_NAME=_"female^Guardian"} {GUARDIAN}
@@ -339,8 +340,9 @@
                 [/not]
             [/filter_location]
             [command]
+                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type"
                 # po: Unit name
-                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type" $loc.x $loc.y () (_ "Guard")
+                $loc.x $loc.y () (_ "Guard")
                 # po: Unit name
                 FEMALE_NAME=_"female^Guard"} {GUARDIAN}
             [/command]
--- a/data/campaigns/Descent_Into_Darkness/scenarios/07c_A_Small_Favor3.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/07c_A_Small_Favor3.cfg
@@ -45,11 +45,9 @@
         hidden=yes
         color=purple
 
-        {NAMED_NOTRAIT_UNIT 2 (Silver Mage) 11 9 ("")
-        # po: Book guards
-        ( _ "Guardian")
-        # po: Book guards
-        FEMALE_NAME=_"female^Guardian"} {GUARDIAN}
+        # po-override: Book guards
+        # po-override: Book guards
+        {NAMED_NOTRAIT_UNIT 2 (Silver Mage) 11 9 ("") ( _ "Guardian") FEMALE_NAME=_"female^Guardian"} {GUARDIAN}
         {NAMED_NOTRAIT_UNIT 2 (Silver Mage) 13 7 ("") ( _ "Guardian") FEMALE_NAME=_"female^Guardian"} {GUARDIAN}
 
         [ai]
@@ -340,11 +338,7 @@
                 [/not]
             [/filter_location]
             [command]
-                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type"
-                # po: Unit name
-                $loc.x $loc.y () (_ "Guard")
-                # po: Unit name
-                FEMALE_NAME=_"female^Guard"} {GUARDIAN}
+                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type" $loc.x $loc.y () (_ "Guard") FEMALE_NAME=_"female^Guard"} {GUARDIAN}
             [/command]
         [/random_placement]
         [random_placement]
--- a/data/campaigns/Descent_Into_Darkness/scenarios/10_Endless_Night.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/10_Endless_Night.cfg
@@ -1420,9 +1420,9 @@
                 {VARIABLE metInky 1}
             [/then]
             [else]
-                # po: squiddy means "squid-like" in this sentence
                 [message]
                     speaker=Mal Keshar
+                    # po: squiddy means "squid-like" in this sentence
                     message= _ "We meet again, squiddy one."
                 [/message]
             [/else]
@@ -1464,9 +1464,9 @@
                             speaker=Mal Keshar
                             message= _ "Release me at once you oversized piece of calamari— eeeeaaaaaaaaa!"
                         [/message]
-                        # po: Mal Keshar is sinking...
                         [message]
                             speaker=narrator
+                            # po: Mal Keshar is sinking...
                             message= _ "<i>Blub... blub... blub...</i>"
                         [/message]
 
--- a/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
@@ -816,8 +816,9 @@
             [/remove_item]
             [floating_text]
                 x,y=29,31
-                # po: Treat this as if it was a sound-effect in a comic book.
-                text="<span color='#ff7722'>" + _ "Smash!" + "</span>"
+                text="<span color='#ff7722'>" +
+                    # po: Treat this as if it was a sound-effect in a comic book.
+                    _ "Smash!" + "</span>"
             [/floating_text]
             {QUAKE "rumble.ogg"}
             [scroll_to]
--- a/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
@@ -816,9 +816,8 @@
             [/remove_item]
             [floating_text]
                 x,y=29,31
-                text="<span color='#ff7722'>" +
-                    # po: Treat this as if it was a sound-effect in a comic book.
-                    _ "Smash!" + "</span>"
+                # po: Treat this as if it was a sound-effect in a comic book.
+                text="<span color='#ff7722'>" + _ "Smash!" + "</span>"
             [/floating_text]
             {QUAKE "rumble.ogg"}
             [scroll_to]
--- a/data/campaigns/The_South_Guard/scenarios/06a_Into_the_Depths.cfg
+++ b/data/campaigns/The_South_Guard/scenarios/06a_Into_the_Depths.cfg
@@ -275,10 +275,9 @@
 #ifdef EASY
         {GENERIC_UNIT 2 (Revenant) 19 26} {GUARDIAN}
         {GENERIC_UNIT 2 (Revenant) 19 29} {GUARDIAN}
-        # po: Dark Sorcerer name
-        {NAMED_GENERIC_UNIT 2 (Dark Sorcerer) 20 27 (Mal Maul) (_"Mal Maul")
-        # po: Dark Sorceress name
-        FEMALE_NAME=_"female^Mal Maul"} {GUARDIAN}
+        # po-override: Dark Sorcerer name
+        # po-override: Dark Sorceress name
+        {NAMED_GENERIC_UNIT 2 (Dark Sorcerer) 20 27 (Mal Maul) (_"Mal Maul") FEMALE_NAME=_"female^Mal Maul"} {GUARDIAN}
 #endif
 #ifdef NORMAL
         {GENERIC_UNIT 2 (Revenant) 19 26} {GUARDIAN}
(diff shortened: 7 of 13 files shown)


Slayer95:
cc @AncientLich

AncientLich:
The behaviour you reported does not surprise me.
In this context, only the second #po-override is expected to work, and only for the very next sentence (so only for "Guardian" and not for "female_Guardian"

While I will, in this context, recall what I said in #7540 about _"variable content"_, I can see, however, a possible work-around in this particular context, using `# wmlxgettext`, together with `#po-override`

`# wmlxgettext` is a particular instruction that allows to add a fake WML code in order to allow wmlxgettext to read a code like it would exist even if it is not an actual WML code.
Its original purpose was to allow to rebalance unbalanced tags (when unbalanced on-purpose) but nothing forbids you to use this command also in order to instruct wmlxgettext how to act in a specific context (however it works only in WML .cfg files and not in .lua files)

So... I could image that this work-around could work

```
# po-override: "Guardian": This is a unit tasked with guarding a place (male version)
# wmlxgettext: _ "Guardian"
# po-override: "female^Guardian": This is a unit tasked with guarding a place (female version)
# wmlxgettext: _"female^Guardian"

# test your code here, and see if there is a third, undesired comment generated or not 
# if wmlxgettext generates an undesired third comment, then try to add an empty po-override and see what happens
{NAMED_NOTRAIT_UNIT 2 ("Silver Mage") 13 7 ("") ( _ "Guardian") FEMALE_NAME=_"female^Guardian"}
```
The downside of this work-around is that you will see two additional line of codes mentioned in source references.
Probably you could use a more elegant `#wmlxgettext` fake code (also composed by multiple lines) but I don't remember well WML reference anymore, so I wrote a very dumb code that should do the work, hopefully.

Remember that, if you use that macro for guardian more then once, you actually need to use that additional lines of code with `# wmlxgettext` and `# po-override` only once, in this case. So in all the following instances of "Guardian" NAMED_NOTRAIT_UNIT you will not require to add any other comment.

However, personally, I would suggest you to write your code slightly differently. For example I would suggest to define a custom macro in order to place a Guardian, so you would use, for example {GUARDIAN} instead of the parametrized macro NAMED_NOTRAIT_UNIT: easier to read and quicker to write, expecially if you need it multiple times (unless you use GUARDIAN only once. In that case using NAMED_NOTRAIT_UNIT is more convenient).

Slayer95:
> The behaviour you reported does not surprise me.
In this context, only the second #po-override is expected to work, and only for the very next sentence (so only for "Guardian" and not for "female_Guardian"

As seen in the ``Old output:`` section, none of those ``#po`` nor ``#po-override`` comments currently work, since they are all swallowed by the ``"Silver Mage"`` non-translatable string (*).

> \# wmlxgettext is a particular instruction that allows to add a fake WML code in order to allow wmlxgettext to read a code like it would exist even if it is not an actual WML code.
Its original purpose was to allow to rebalance unbalanced tags (when unbalanced on-purpose) but nothing forbids you to use this command also in order to instruct wmlxgettext how to act in a specific context (however it works only in WML .cfg files and not in .lua files)
So... I could image that this work-around could work

That indeed works. However, it's quite untidy code. My proposal is making a sequence of ``#po-override`` be applied sequentially to encountered strings.

```php
# po-override: Male unit name
# po-override: Female unit name
{NAMED_NOTRAIT_UNIT 2 ("Silver Mage") 13 7 ("") ( _ "Guardian") FEMALE_NAME=_"female^Guardian"}
```

So that _"Male unit name"_ is seamlessly matched to **_"Guardian"**, and _"Female unit name"_ is matched to **_"female^Guardian"**.

However, that cannot currently happen since ``wmlxgettext`` forgets about comments on the first string found. That has always been done here without justification https://github.com/wesnoth/wesnoth/blob/c30c30acfc3b8ba85742a449e2e65beec3dedfcd/data/tools/pywmlx/state/machine.py#L290-L291 . The Lua analog at least says it's intended, but why so?

https://github.com/wesnoth/wesnoth/blob/c30c30acfc3b8ba85742a449e2e65beec3dedfcd/data/tools/pywmlx/state/machine.py#L243-L246

> However, personally, I would suggest you to write your code slightly differently. For example I would suggest to define a custom macro in order to place a Guardian, so you would use, for example {GUARDIAN} instead of the parametrized macro NAMED_NOTRAIT_UNIT: easier to read and quicker to write, expecially if you need it multiple times (unless you use GUARDIAN only once. In that case using NAMED_NOTRAIT_UNIT is more convenient).

That's also nice, but I am proposing a general solution that doesn't involve faking WML code at all.

CelticMinstrel:
The two things I draw from your explanation are:

1. We really need to do _something_ to handle multiple strings on the same line, whether it's what's done in this pull request, or warning about it, or even both. Completely ignoring it and just doing "whatever" is the one thing that isn't acceptable.
2. This change needs to make sure it's not misapplying a po comment to a different string like in the example shown.

Slayer95:
First of all, silent failures are bad. Specially so in the one task ``wmlxgettext`` is supposed to deal with: collecting translator comments and attaching them to translatable sentences. I address that in #7573 .

However, I think it shouldn't be the case that non-translatable strings affect the build ``.pot`` file. I am already addressing that in this PR (#7570). 

> This change needs to make sure it's not misapplying a po comment to a different string like in the example shown.
As for the usability concern regarding forgetting the underscore, it's not too hard to add a warning if the ``#po`` comment is too far from the associated sentence.
#po, #po-override directives are no longer consumed by untranslatable strings.

I will be rebasing this PR on #7573.

AncientLich:
> We really need to do something to handle multiple strings on the same line, whether it's what's done in this pull request, or warning about it, or even both. Completely ignoring it and just doing "whatever" is the one thing that isn't acceptable.
> This change needs to make sure it's not misapplying a po comment to a different string like in the example shown.

As long as I am not currently maintaining wmlxgettext I can only share my point-of-view and I can only explain how the code was designed and why I did the choices I did in the time I wrote it.  
Here, I am sharing my personal doubts about some counter-effect about features, and I am sharing my opionon what I would suggest to do, but any final decision is up to you active developers and, in general, wesnoth team.

> First of all, silent failures are bad.
I agree with you. I had a great hadaches while writing this part of code. The pending phase (and the trasnformation in po sentence) was a very nightmare to design.
The huge problem is, while trying to add those extra features (po and po-override that did not exist in previous perl wmlxgettext tool) is that you can't know how many lines separates the instruction and the translatable string itself, for example in case of a very long tag with only one translatable string, with the author that could place the command before the beginning of the tag itself.
So, programmaticly speaking, the concept of "near" is not the same of "near" in human world, while you can, as a person, analyze a context and understand what you are referring exactly, this is not true on designing wmlxgettext code. For this reasn, nothing could exclude that, "near" could be also 100 rows or even more, in theory.

> Indeed it shouldn't be the case that non-translatable strings affect the build .pot file. I am already addressing that in in https://github.com/wesnoth/wesnoth/commit/f06450e247acc13cf14af4056e6be3f49ecdacd8, as explained in the OP

I don't agree that it is a good idea to simply remove the cleaning, at the current state.
Probably, in order to remove properly that limitation and avoiding counter-effects it could be required to re-design the PendingWML and PendingLua in a stronger way (I do not have actual idea how, however), excpecially in order to try to add a support for custom comment to be split around multiple translatable sentences into a single line of WML code like a parametrized macro. And possibly the solution is add another command that instruct that specific situation, a command like:

```
# po-array-def: 2
# po-array[1]: sentence for first translatable string
# po-array[2]: sentence for second translatable string
```
but the real issue is the pending phase management in general.
Pending phase (lua or WML) exist even if no translatable string exists and must collect and manage a lot of different scenarios.
Pending phase collects informations line-by-line, so dropping wrong information is crucial (for example dropping informations collected from a closed tag that does not contain translatable strings). But Pending exist the same, regardless of WML point of code. So the same Pending exists during all the file reading process, until reach of files, clearing internal informations that needs to be dropped (like the example).

Slayer95:
Regarding
>  it shouldn't be the case that non-translatable strings affect the build .pot file

@AncientLich has raised some concern about side effects, but imo adding enough warnings/errors should do the job.

> you can't know how many lines separates the instruction and the translatable string itself [...] programmaticly speaking, the concept of "near" is not the same of "near" in human world [...] nothing could exclude that, "near" could be also 100 rows or even more, in theory.

It's true that it's not generally possible to know how many lines apart the instruction and the translatable string are. However, I have figured out an appropriate and practical definition of "near", catering specifically to the "forgetting ``_``" use case. That is, let "near" mean _"no further apart than the next non-translatable string, measured in line count"_. Therefore, the following would raise an error:

```php
[message]
#po: explanation on string I forgot to mark as translatable
message = "You are a wizard, Harry."
female_message = _"You are a witch, Harriet."
#Note: After #7573, even if there is no ``female_message`` an error will be raised when closing the tag.
[/message]
```

But the following would not:
```php
[message]
#po: explanation written at the start of ``[message]`` as stylistic choice.
speaker=$unit.id
message=_"You are a wizard, Harry."
female_message=_"You are a witch, Harriet."
[/message]
```
> for example in case of a very long tag with only one translatable string, with the author that could place the command **before** the beginning of the tag itself.

In fact, when a ``[message]`` tag is opened, all collected ``#po: `` directives are deleted, so such a comment must always be placed _inside_ the tag, and never before.

https://github.com/wesnoth/wesnoth/blob/c30c30acfc3b8ba85742a449e2e65beec3dedfcd/data/tools/pywmlx/state/wml_states.py#L199-L207

Slayer95:
@stevecotton and @CelticMinstrel have posted relevant comments in #7573 . Respectively:

> Would the WML be clearer if the whole thing could be put on one line (including the line that's currently split), and have a single comment that's suitable for both genders?
[...] Allowing two translatable strings on the same line might be out-of-scope for the current PR [sic]
[...] I think the old WML should be acceptable here, but should apply the comment to the translatable string. OTOH, I'd be okay with saying that's a feature that should be postponed to another PR [sic].

> I was thinking we could have # po: ... apply a comment to the first translatable string on the next line and # po-all: ... apply it to all translatable strings on the next line, but I'm not sure if that's a good design.

If you allow me, @CelticMinstrel , I'd rephrase both instances of _"on the next line"_ to _"on the next appliable line"_ (literally restricting ``#po`` to adjacent lines would remove the options of adding multiple ``#po`` comments, as well as of writing ``#po`` comments further apart.)

So, together with @AncientLich 's ``#po-array``, there are now three competing proposals to deal with having several translatable strings in a single line.

1. Queue: ``#po-override``
2. Distributive: ``#po-all``
3. Array: ``#po-array-def`` + ``#po-array``

The first thing to be noted is that, strictly speaking, my proposal (1) breaks backwards compat. Now, compat breakage isn't introduced just because. In fact, it keeps ``wmlxgettext``'s syntax minimal. Still, someone could say that doesn't outweight the cost of breakage. However, there is nuance in such "breakage":
- As of today, in wesnoth's ``master`` branch, there are zero (0) instances of ``#po-override`` comments.
- Since this is a change to a build tool, there is no effect on currently released content, including add-ons. Any disrupted add-on developers will be able to choose between retrieving an older ``wmlxgettext`` version, or updating their code.
- Breakage coming from this proposal involves authors using multiple ``#po-override`` comments for a single translatable sentence, which is [not allowed according to the docs](https://wmlxgettext-unoff.readthedocs.io/en/latest/enduser/wml.html#po-override-override-info) , but does actually accumulate them against the same sentence.
- Current output format of ``#po-override`` is broken (it outputs comments in Python's ``['list', 'of', 'strings']`` representation). Note that ``'of'``, and ``'strings'`` come from multiple ``#po-override`` directives, which shouldn't be possible, as stated in previous bulletpoint.

With that out of the way, let's focus on our use cases. Just how often do we meet a chance to put more than one translatable string in a single line?

I am not too familiar with Lua APIs, but I can tell there is support for ``gettext`` plurals. Only the singular variant is of course treated as an independent translatable string.

In WML it's the gendered variants that are involved: ``name`` (for units/messages), and ``text`` (for ``[unstore_unit]``), and more recently macros (#7546). IMO, in a _WML tag declaration_, while it might make stylistical sense to group male and female variants of a string in the same line, it would often cause readers to miss the rightmost variant, given [known text scanning patterns](https://www.nngroup.com/articles/text-scanning-patterns-eyetracking/), and the length translatable strings may reach. In contrast, _in macros_, brevity is the name of the game, so I think it's critical that macros are properly addressed by whatever is decided.

And as it turns out, macros that expand to WML tags are supposed to be tagged with ``#po-override``. Otherwise, automatically extracted data (``speaker``, etc.) would leak from parent WML tags into the macro arguments. So, whatever is decided, there needs to be an ``override`` variant to it.

AncientLich:
Moreover I think there is very high chances (if not 100%, very near to 100%) that no-one, until now, ever used `#po-override` instruction in his/her wesnoth addon, so I assume that changing its behaviour from scratch (for example allowing multiple instances of #po-override or removing completely that command) has very high chance to do not have any impact to any previously released third-party wesnoth addon.
So I think that any solution could be equally followed.
The only extra suggestion I can add is that the "command" must be understandable... the name `# po-override` used the concept was called in that way since its expected behaviuour was allowing developer to override the automatic informations assigned to a single instance of that sentence and replace that line of comment using a custom one instead (one custom line of comment which replaces one code-based line of comment). So, I think that every instruction should have a name that would help the developer to understand what exactly was doing, so using a name that could be as much self-explanatory as possible

Slayer95:
The concept of overriding has two dimensions. (a) _Overriding_ the automatically extracted context, (b) _Overriding_ other directives. I think it's not directly clear to an uninformed or misguided reader/author which of them or if both apply. (Only (a) does, contrary to the docs).

This proposal, however, would canonize only (a). So, should they see two overrides together or a ``#po`` and ``#po-override`` even, they would possibly get confused or attempt to "fix" them by removing the first one.

Now, if we were to look for a more apt name that allows two directives without confusion, I think we can draw inspiration on the definition of ``(a)`` and call it ``#po-ctx-*``.

Then, `#po-ctx-queue``` would appy to on sentence, and ``#po-ctx-all`` would apply to all sentences in the entire line.

But note here that ``queue`` and ``all`` also introduce a 2D meaning issue, and someone might think that ``queue`` appends to the list of extracted infos, while ``all`` replaces them _all_. Therefore, I believe that ``#po-ctx-each`` would be the superior option. And with the meaning of ``#po-ctx-each`` clear, inference of what ``#po-ctx-queue`` means is simpler.

Still, an advantage of the ``#po-array[n]`` approach would be the removal of the question of what "queue" means. However, I can't get behind the strict syntax proposed by @AncientLich . I think it'd be better to do without ``#po-array-def`` (it's possible if pending directives are cleared often enough). Also, supporting sparse arrays would be the cherry on top of this approach. Of course, for macros, it'd be ``#po-ctx-array``.

CelticMinstrel:
> Moreover I think there is very high chances (if not 100%, very near to 100%) that no-one, until now, ever used `#po-override` instruction in his/her wesnoth addon, so I assume that changing its behaviour from scratch (for example allowing multiple instances of #po-override or removing completely that command) has very high chance to do not have any impact to any previously released third-party wesnoth addon.
> So I think that any solution could be equally followed.

We looked into that:

> [11:18am] wdiscord: [Elvish_Hunter] I've checked the 1.14, 1.16 and 1.17 servers and the only add-ons which use `#po-override` are `Arena` (1.16/1.17), `Undead_Empire` (1.14, 1.16, 1.17), `Nightmare_To_The_Throne` (1.16) and `The_South_Guard_Nightmare` (1.16).

So there are 4 add-ons that use it. I don't know if they're using it as intended or if they're using it wrong.



> (b) _Overriding_ other directives.

I don't think this is useful, as it creates confusion over which comment actually applies to the string in cases where the string appears more than once.

Slayer95:
> add-ons which use #po-override are

> Arena (1.16/1.17)

Suboptimally used (``#po`` would be terser), but not wrong.

> Undead_Empire (1.14, 1.16, 1.17)

Full checkmarks in 5 instances. This add-on also exhibits usage of ``#po`` together with ``#po-override`` for a single sentence.

> Nightmare_To_The_Throne (1.16)

Suboptimally used (``#po`` would provide more information), but not wrong. Inherited from mainline's TSG, where it was recently removed in #7545.

> The_South_Guard_Nightmare (1.16).

Full checkmarks for this instance. Inherited from mainline's TSG, where it was recently removed in mainline in [#7546](https://github.com/wesnoth/wesnoth/pull/7546/files#diff-717070f87860af91ed820c680b1ea4f72aa3196e825e5d298b8c7b759952d94aL302). In retrospective, I have to admit that removal wasn't optimal, but we are on the road there.

### Compat:
Overall, none of these would be affected if ``#po-override`` were to be changed as described, but it would of course not be fair to remove that directive without cooperating with them.

Slayer95:
Something else about ``#po-array``: Both ``gettext`` and ``WML`` use zero-based array indices. However, Lua is one-based. That leaves room for confusion and off-by-one errors whatever convention is followed. In contrast, the concept of ``queue`` is unambiguous in that regard.

As for ``#po-each`` / ``po-all``, it has in its favor that it would result in more concise comments and it would be a clean addition on top of the current convention of prepending gender variants with ``female^``. In contrast, other alternatives would result in duplicate comments of the sort:

```
#directive a: Generic comment
#directive b: Generic comment
{MACRO _"string" _"female^string"
```
(Here, ``Generic comment`` is the same for both directive a and directive b.)

Or 

```
#directive-a: Male comment
#directive-b: Female comment
{MACRO _"string" _"female^string"
```
(Here, ``female`` is doubly-specified for the second sentence.)

Slayer95:
Re: The four addons that use ``#po-override``: @sevu mentioned they are willing to remove it from the first two. The other two are authored by @beetlenaut .

AncientLich:
I have a question: are there any scenario, other than WML macros, that it expected to have a specific support for multiple SINGLE (not concatenated) translatable string into a single line?
if the answer is NO, than a proposal could be this one: introduce a command `#po-wmacro` specific for macros, that will act partially like po-override (preventing automated informations that, in case of macros, it would be probably wrong), but that can be used only if the very next commands can be only: 
- another `#po-wmacro` instruction, intended one instruction exactly one translatable string 
- a WML macro (I assume that, as long as wmlxgettext can replace a macro meaning, it could understand it is parsing a macro)

in this way wmlxgettext could:
- return an error if `# po-wmacro` does not immediately follow a macro or another `# po-wmacro` instruction (returning an error if comments, other wmlxgettext instruction or even blank line separates `# po-wmacro` instructions from WML macro, in order to let it use strictly to WML MACRO
- return an error if the number of `# po-wmacro` consecutive instruction doesn't match exactly the number of translatable string that the macro is expected to have (and not counting untranslatable strings)... so if a macro has 1 non-translatable string and 2 translatable strings, than error will be raised if there aren't exactly two or zero `# po-wmacro` instructions (according that no developer should be never forced to customize a comment if not actually required)

This, however, could be difficult to code, and probably will require a lot of work on PendingWML class in order to activate this command, including changes into nodesentence to posentence conversion, and all checks in order to verify if `# po-wmacro` is used properly and that it will ensure, like `# po-override`, to avoid unwanted automation.

CelticMinstrel:
> multiple SINGLE (not concatenated) translatable string

Why the emphasis on "not concatenated"? I could imagine some cases where you'd have multiple strings on the same line because they are concatenated, for example:

```
description="<b>" + _"Some Header" + "</b>: " + _"Detailed Description"
```

Slayer95:
> I assume that, as long as wmlxgettext can replace a macro meaning, it could understand it is parsing a macro

For maximum reliability, wmlxgettext has no knowledge nor does it attempt to parse macro calls in #

AncientLich:
> Why the emphasis on "not concatenated"?

Simply because multiple concatenated strings is actually a single string. In that case, probably, the expected result could be also a single translatable string, so there isn't an actual need on introducing something that allows to split comment informations between different translatable strings placed in a single line.

In case of three (not concatenated) translatable strings we have three different strings that could need three different comments to be assigned.
So why I emphatized "not concateated". Because when we have multiple translatable strings in the same line, and they are not concatenated, than we have a situation were we actually need to separate comments and assign each comment to the correct translatable string.

Slayer95:
> Simply because multiple concatenated strings is actually a single string. In that case, probably, the expected result could be also a single translatable string, so there isn't an actual need on introducing something that allows to split comment informations between different translatable strings placed in a single line.

Theoretically, that's a possibility, and I guess concatenating them would be easier on translators. However, Wesnoth currently treats the substrings as mutually independent for localization purposes.

There's also this case
```css
        [message]
            speaker=unit
            message,female_message=_"I am male",_"I am female"
        [/message]
```

At this point, I am thinking it'd be for the best to implement more than one of the mentioned proposals, but which of them is still a question.

CelticMinstrel:
> Simply because multiple concatenated strings is actually a single string.

I don't understand the basis for this statement. Sure, they are a single `t_string` object in the resulting config, but they are still two separate entries in the po file.

AncientLich:
> I don't understand the basis for this statement. 

It is not important, as long as I have enough information to answer my original question: "Could be a solution to add a specific support for WML macro? (like `po-wmacro` purposal)?" "No, it is not a good idea, as long as there are other scenarios where it is possible to find multiple translatable string placed in a single line of code".

So we can easily say that the purpose of `po-wmacro` command and its mechanics must be dropped, and another solution must be used instead (changing `po-override` behaviour, introduce `po-array` or something similar, or introducing something even different we still didn't purpose)

Elvish-Hunter (approved):
The fact that only a few add-ons use the `#po-override` directive (incorrectly in some instances) and that the updated behavior just "makes sense" has convinced me to approve this PR (which still has conflicts that need to be fixed for merging).
Sorry for not having been able to review it sooner.

cooljeanius:
> only a few add-ons use the `#po-override` directive (incorrectly in some instances)

I mean, I was considering starting using it based on the outcome of this PR...

Pentarctagon:
Just to confirm - is this ready to be merged or are there concerns remaining?

stevecotton:
Took a quick look at rebasing this, but merging it with the changes of #7574 is non-trivial. It's probably not hard, but doing and testing it today before the slightly-delayed freeze deadline would need someone who's already familiar with the changes.

Labelling as postponed, but please don't let that stop you if you already have the knowledge.

cooljeanius:
> Is this still being pursued?

I hope so; some of my add-ons could use this... although I am kind of concerned about potential compatibility breaks...

Pentarctagon:
@Slayer95 do you intend to come back to this at some point?

Slayer95:
It appears to me that consensus has been favorable to this PR. Just needs a rebase.

Please speak up if you disagree.

cooljeanius:
> It appears to me that consensus has been favorable to this PR. Just needs a rebase.
> 
> Please speak up if you disagree.

Well, just make sure to test and document the changes... I'm still worried about compatibility breaks for my add-ons, but I'm also looking forward to the new potential it could bring.

Wedge009:
> Just needs a rebase.

Looks like there are conflicts with a re-base, though.

Pentarctagon:

PR #7571, opened 2023-04-25
Add some search terms to the match history dialog.

Pentarctagon:
This still defaults to searching by the selected player name, but now also allows searching by:
* player name
* game name
* one of scenario id, era id, or modification id

The game name, scenario id, era id, and modification id support a leading and/or trailing wildcard for partial matches by essentially replacing the leading and/or trailing asterisk with a percent sign.

Scenario, era, and modification parameters are the ID, not the name, since the server gets the translated value for the name. Therefore searching by name would only give partial results in nearly all situations. So while this is probably unintuitive to a player, it still seems like the less bad option.

---

The updated UI:
Screenshot from 2023-04-25 18-20-43

The close button and left/right arrows have been moved up into the same line as the new Search button to save vertical space.

### DIFF
--- a/src/gui/dialogs/multiplayer/match_history.cpp
+++ b/src/gui/dialogs/multiplayer/match_history.cpp
@@ -24,6 +24,8 @@
 #include "gui/widgets/button.hpp"
 #include "gui/widgets/label.hpp"
 #include "gui/widgets/listbox.hpp"
+#include "gui/widgets/menu_button.hpp"
+#include "gui/widgets/text_box.hpp"
 #include "gui/widgets/window.hpp"
 #include "network_download_file.hpp"
 #include "serialization/string_utils.hpp"
@@ -56,7 +58,37 @@ void mp_match_history::pre_show(window& win)
 	connect_signal_mouse_left_click(newer_history, std::bind(&mp_match_history::newer_history_offset, this));
 	connect_signal_mouse_left_click(older_history, std::bind(&mp_match_history::older_history_offset, this));
 
+	button& search = find_widget<button>(&win, "search", false);
+	connect_signal_mouse_left_click(search, std::bind(&mp_match_history::new_search, this));
+
+	text_box& search_player = find_widget<text_box>(&win, "search_player", false);
+	search_player.set_value(player_name_);
+
+	std::vector<config> content_types;
+	content_types.emplace_back("label", _("Scenario"));
+	content_types.emplace_back("label", _("Era"));
+	content_types.emplace_back("label", _("Modification"));
+
+	find_widget<menu_button>(&win, "search_content_type", false).set_values(content_types);
+
 	update_display();
+}
+
+void mp_match_history::new_search()
+{
+	int old_offset = offset_;
+	std::string old_player_name = player_name_;
+	text_box& search_player = find_widget<text_box>(get_window(), "search_player", false);
+	player_name_ = search_player.get_value();
+
+	// display update failed, set the offset back to what it was before
+	if(!update_display()) {
+		offset_ = old_offset;
+		player_name_ = old_player_name;
+	} else {
+		label& title = find_widget<label>(get_window(), "title", false);
+		title.set_label(VGETTEXT("Match History — $player", {{"player", player_name_}}));
+	}
 }
 
 void mp_match_history::newer_history_offset()
@@ -79,7 +111,7 @@ void mp_match_history::older_history_offset()
 
 bool mp_match_history::update_display()
 {
-	const config history = request_history(offset_);
+	const config history = request_history();
 
 	// request failed, nothing to do
 	if(history.child_count("game_history_results") == 0) {
@@ -163,12 +195,15 @@ bool mp_match_history::update_display()
 	return true;
 }
 
-const config mp_match_history::request_history(int offset)
+const config mp_match_history::request_history()
 {
 	config request;
 	config& child = request.add_child("game_history_request");
-	child["offset"] = offset;
-	child["search_for"] = player_name_;
+	child["offset"] = offset_;
+	child["search_player"] = player_name_;
+	child["search_game_name"] = find_widget<text_box>(get_window(), "search_game_name", false).get_value();
+	child["search_content_type"] = find_widget<menu_button>(get_window(), "search_content_type", false).get_value();
+	child["search_content"] = find_widget<text_box>(get_window(), "search_content", false).get_value();
 	DBG_NW << request.debug();
 	connection_.send_data(request);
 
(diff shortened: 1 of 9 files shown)


src/server/common/dbconn.cpp
```diff
@@ -147,31 +184,91 @@ std::unique_ptr<simple_wml::document> dbconn::get_game_history(int player_id, in
...
-"inner join "+db_game_content_info_table_+" era "
+"  and scenario.GAME_ID = game.GAME_ID ";
+	if(search_content_type == 0)
+	{
+		if(!search_content.empty())
+		{
+			game_history_query += " and scenario.ID like concat(";
+			game_history_query += search_content_leading ? "'%'," : "";
+			game_history_query += " ? ";
+			game_history_query += search_content_trailing ? ",'%')" : ")";
```
soliton-, line 218: Is concat needed? Why not just replace `*` with `%` and `?` with `_` in the search strings?

src/server/common/dbconn.cpp
```diff
@@ -147,31 +184,91 @@ std::unique_ptr<simple_wml::document> dbconn::get_game_history(int player_id, in
 "    select 1 "
 "    from "+db_game_player_info_table_+" player1 "
 "    where game.INSTANCE_UUID = player1.INSTANCE_UUID "
-"      and game.GAME_ID = player1.GAME_ID "
-"      and player1.USER_ID = ? "
-"  ) "
+"      and game.GAME_ID = player1.GAME_ID ";
+	game_history_query += player_id != 0 ? " and player1.USER_ID = ? " : " and player1.USER_ID != ? ";
```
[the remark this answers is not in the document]
Pentarctagon: Yep.

src/server/common/dbconn.cpp
Pentarctagon: I tried that initially, and what ends up happening is the `%` gets treated as a literal part of the string rather than as a wildcard.

src/server/common/dbconn.cpp
```diff
@@ -147,31 +184,91 @@ std::unique_ptr<simple_wml::document> dbconn::get_game_history(int player_id, in
...
+	if(!search_game_name.empty())
+	{
+		game_history_query += " and game.GAME_NAME like concat(";
+		game_history_query += search_game_name_leading ? "'%'," : "";
+		game_history_query += " ? ";
+		game_history_query += search_game_name_trailing ? ",'%')" : ")";
+	}
+	else
+	{
+		game_history_query += " and game.GAME_NAME != ? ";
```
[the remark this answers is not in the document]
Pentarctagon: It's not, but someone intentionally creating a game with no name is a rare thing and I wasn't sure of a better way to do it.

The brace formatting is consistent with the rest of the file.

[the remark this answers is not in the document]
Pentarctagon: I could perhaps change it to `and game.GAME_NAME != ? and game.GAME_NAME is null`? The null condition at least can never be true.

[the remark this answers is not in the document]
soliton-: I think it's better to just not try to shoehorn everything into one query. While you build up the query you can as well build a list of only relevant parameters.

src/server/common/dbconn.cpp
Pentarctagon: It's certainly possible - I'll take a closer look.

src/server/common/dbconn.cpp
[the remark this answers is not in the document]
Pentarctagon: The issue I wasn't sure how to work around was how to handle having a variable list of parameters depending on what was passed in. A `std::vector<std::variant<...>>` might be a more flexible approach than a parameter pack I suppose.

src/gui/dialogs/multiplayer/match_history.cpp
```diff
@@ -187,7 +222,7 @@ const config mp_match_history::request_history(int offset)
 				DBG_NW << "Received non-history data: " << response.debug();
 				if(!response["error"].str().empty()) {
 					ERR_NW << "Received error from server: " << response["error"].str();
-					gui2::show_error_message(_("The server responded with an error:")+response["error"].str());
+					gui2::show_error_message(_("The server responded with an error: ")+response["error"].str());
```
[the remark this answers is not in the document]
Pentarctagon: Fixed.

src/server/common/dbconn.cpp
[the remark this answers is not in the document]
Pentarctagon: Updated to use a `std::vector<std::variant<...>>`.

src/server/common/dbconn.cpp
Pentarctagon: Updated to use `utils::to_sql_wildcards`.

src/server/common/dbconn.cpp
```diff
@@ -117,10 +118,46 @@ std::string dbconn::get_tournaments()
 	}
 }
 
-std::unique_ptr<simple_wml::document> dbconn::get_game_history(int player_id, int offset)
+std::pair<bool, bool> substitute_wildcards(std::string& parameter)
```
soliton-, line 121: This is not used anymore now, I think.

src/server/common/dbconn.cpp
```diff
@@ -147,31 +184,82 @@ std::unique_ptr<simple_wml::document> dbconn::get_game_history(int player_id, in
...
 "  and player.USER_ID != -1 "
-"  and game.END_TIME is not NULL "
-"inner join "+db_game_content_info_table_+" scenario "
+"  and game.END_TIME is not NULL ";
+
+	if(!search_game_name.empty())
+	{
+		game_history_query += " and game.GAME_NAME like concat(?)";
+
+		utils::to_sql_wildcards(search_game_name, false);
```
soliton-, line 205: `to_sql_wildcards` escapes `_` and does not support `?`. Is that wanted here as well? Do we tell the user somewhere that they can use wildcards? (We might as well tell them to use SQL wildcards?)

Pentarctagon: `to_sql_wildcards` optionally escapes `_`, and here is set to not do that (the second argument to the method).

soliton-: Alright, what about the rest?

Pentarctagon: I wouldn't expect regular people to know what SQL wildcards are, and underscores in particular are already in a lot of IDs so treating them as a wildcard would be unexpected. Asterisks are at least somewhat more widely known.

soliton-: Do we tell the user somewhere that they can use wildcards?

Pentarctagon: Not currently, no.

src/server/common/dbconn.cpp
Pentarctagon: Removed.

src/server/common/dbconn.cpp
Pentarctagon: Updated.

soliton-: Ok, I guess that can be done later. I don't think it makes much of a difference whether we tell users that `*` matches any character and `?` a single one or the same for `%` and `_`. Since those characters match themselves it's not that important if you don't escape them.

I think the conversion makes it more confusing since `%` works anyway and if you write `\*` you get `\%` instead.

PR #7573, opened 2023-04-26
wmlxgettext: Error out on unused #po, #po-override directives

Slayer95:
```
# po: narrator is explaining the tale
[message]
speaker=narrator
message= "This is the story I am narrating, but I forgot to mark is a translatable"
[/message]

[message]
speaker=Delfador
message= _ "Hi. I am a mage, and my sentence will be translated"
[/message]
```

According to @AncientLich , current behavior is:
> in this situation, discarding all previous comments after the first sentence found avoids to place itself in a wrong place. If the # po wouldn't discarded, than it would be applied to the next translatable string, with the result of adding the comment "narrator is explaining the tale" in the sentence "Hi. I am a mage, and my sentence will be translated".

> This behaviour is thinked in order to prevent to misplace comments or overrides. The logic around this choice was "better dropping an information, than adding a completely wrong one that could add confusion to a translator".
The same reason is more strong when thinking about lua because wmlxgettext is even more limted on understand what lua is doinf than WML, so introducing a more strict control is even more necessary.

So this is a silent failure in the procedures ``wmlxgettext`` is expected to deal with. This PR makes them loud errors.

### DIFF
--- a/data/campaigns/Descent_Into_Darkness/scenarios/07b_A_Small_Favor2.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/07b_A_Small_Favor2.cfg
@@ -258,8 +258,9 @@
                 [/not]
             [/filter_location]
             [command]
+                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type"
                 # po: Unit name
-                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type" $loc.x $loc.y () (_ "Guard")
+                $loc.x $loc.y () (_ "Guard")
                 # po: Unit name
                 FEMALE_NAME=_"female^Guard"} {GUARDIAN}
             [/command]
--- a/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
@@ -816,8 +816,9 @@
             [/remove_item]
             [floating_text]
                 x,y=29,31
-                # po: Treat this as if it was a sound-effect in a comic book.
-                text="<span color='#ff7722'>" + _ "Smash!" + "</span>"
+                text="<span color='#ff7722'>" +
+                    # po: Treat this as if it was a sound-effect in a comic book.
+                    _ "Smash!" + "</span>"
             [/floating_text]
             {QUAKE "rumble.ogg"}
             [scroll_to]
--- a/data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/01_Saving_Parthyn.cfg
@@ -544,13 +544,13 @@
                     message= _ "Yes, yes it is. Nobody’ll support ya on yer crazy path to depravity."
                 [/message]
 
-                # po: Hint, that the loyal units stay behind
                 [message]
                     side=1
                     race=human
                     [not]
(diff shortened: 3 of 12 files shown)


Pentarctagon:
The CI failure will need to be fixed before merging.

CelticMinstrel:
It looks like your third commit accidentally got some other stuff mixed in? Why is it adding po notes where they didn't exist before? Why is it adding FEMALE_NAME? Regardless of whether that _should_ be done, it shouldn't be done _here_.

Slayer95:
> It looks like your third commit accidentally got some other stuff mixed in? Why is it adding po notes where they didn't exist before? Why is it adding FEMALE_NAME? Regardless of whether that should be done, it shouldn't be done here.

Right. Sorry for the mess. Updating and rebasing on ``master`` fixed it.

There are still errors in random files being revealed by CI...

Slayer95:
Note that, while 98aff6b may seem to go contrary to the goal of not silently failing, it _is_ required in order to run ``wmlxgettext`` on campaigns that contain files using foreign domains. For example, ``--domain wesnoth-wof`` would crash on ``Winds_of_Fate/lua/lua.cfg``, which uses ``wesnoth-multiplayer``.

data/campaigns/Descent_Into_Darkness/scenarios/07b_A_Small_Favor2.cfg
```diff
@@ -258,8 +258,9 @@
                 [/not]
             [/filter_location]
             [command]
+                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type"
                 # po: Unit name
-                {NAMED_NOTRAIT_UNIT 2 "$guardtypetable[$($loc.n % $guardtypetable.length)].type" $loc.x $loc.y () (_ "Guard")
+                $loc.x $loc.y () (_ "Guard")
```
stevecotton, line 263: Would the WML be clearer if the whole thing could be put on one line (including the line that's currently split), and have a single comment that's suitable for both genders?

Maybe `# po: A role-based name, will be used for soldiers of the spearman, fencer and heavy infantry lines`.

Allowing two translatable strings on the same line might be out-of-scope for the current PR.

data/campaigns/Sceptre_of_Fire/scenarios/9_Caverns_of_Flame.cfg
```diff
@@ -816,8 +816,9 @@
             [/remove_item]
             [floating_text]
                 x,y=29,31
-                # po: Treat this as if it was a sound-effect in a comic book.
-                text="<span color='#ff7722'>" + _ "Smash!" + "</span>"
+                text="<span color='#ff7722'>" +
+                    # po: Treat this as if it was a sound-effect in a comic book.
+                    _ "Smash!" + "</span>"
```
stevecotton, line 821: I think the old WML should be acceptable here, but should apply the comment to the translatable string. OTOH, I'd be okay with saying that's a feature that should be postponed to another PR.

stevecotton:
Sorry, commented before checking the other PR. Have marked both my comments as resolved.

data/campaigns/Descent_Into_Darkness/scenarios/07b_A_Small_Favor2.cfg
CelticMinstrel: I was thinking we could have `# po: ...` apply a comment to the first translatable string on the next line and `# po-all: ...` apply it to all translatable strings on the next line, but I'm not sure if that's a good design.

CelticMinstrel:
I added a reply to one of the resolved comments.

Slayer95:
> Sorry, commented before checking the other PR. Have marked both my comments as resolved.

> I added a reply to one of the resolved comments.

Copied them and replied in #7570

PR #7574, opened 2023-04-29
Refactor collection of #po, #po-override comments

Slayer95:
This is a pure refactor that simplifies the code and makes it easier to implement new TBD directives and/or modify behavior of ``#po-override``.

Related: #7570


Pentarctagon:
@Elvish-Hunter are you around?

Elvish-Hunter (approved):
I don't see anything blatantly wrong with this PR, so I think it can go in.

PR #7577, opened 2023-04-30
Schema-validate several things that were being missed

CelticMinstrel:
* WML dialogs loaded in Lua via `wml.load`
* The fonts.cfg file
* The language.cfg file

My one missing piece in here is that a change to po_stat.py _may_ be needed to ensure that en_US.cfg contains `percent=100`. Does anyone know if that's needed and if so, how to do it?


data/languages/en_US.cfg
```diff
@@ -2,4 +2,5 @@
     name="English (US)"
     locale=en_US
     alternates=C, en_PH
+    percent=100
```
stevecotton, line 5: Why not make this key optional?

CelticMinstrel: Because this is literally the _only_ file that does not include it. It doesn't make sense to me to make it optional just so one file can omit it, when it makes just as much sense for that one file to set it to 100.

Pentarctagon:
I think it's fine to leave `po_stat.py` as-is. `en_US.cfg` has apparently always been hand-edited.

PR #7580, opened 2023-05-01
wmlunits: Handle code paths for missing attributes

Slayer95:
This PR deals with several issues similar to the one reported in #7528. However, the specific reason behind #7528 should get clarified after someone pulls up the log files.

### DIFF
--- a/data/tools/unit_tree/wiki_output.py
+++ b/data/tools/unit_tree/wiki_output.py
@@ -16,29 +16,28 @@ def main():
 
     punits = {}
 
-    defines = "NORMAL,ENABLE_ARMAGEDDON_DRAKE,ENABLE_DWARVISH_ARCANISTER," +\
-        "ENABLE_DWARVISH_RUNESMITH,ENABLE_ANCIENT_LICH,ENABLE_DEATH_KNIGHT," +\
-        "ENABLE_TROLL_SHAMAN,ENABLE_WOLF_ADVANCEMENT,ENABLE_WOSE_SHAMAN," +\
-        "ENABLE_PARAGON,ENABLE_SAURIAN_SPEARTHROWER,DISABLE_GRAND_MARSHAL"
+    base_defines = ["NORMAL"]
 
     sys.stderr.write("Parsing core units...\n")
-    wesnoth.parser.parse_text("{core/units.cfg}", defines)
+    wesnoth.parser.parse_text("{core/units.cfg}", ",".join(base_defines))
     punits["mainline"] = wesnoth.parser.get_all(tag = "units")
     punits["mainline"] += wesnoth.parser.get_all(tag = "+units")
 
     all_campaigns = {}
     sys.stderr.write("Parsing campaigns...\n")
-    wesnoth.parser.parse_text("{campaigns}", defines)
+    wesnoth.parser.parse_text("{campaigns}", ",".join(base_defines))
     campaigns = wesnoth.parser.get_all(tag = "campaign")
     for campaign in campaigns:
+        campaign_defines = base_defines[:]
         define = campaign.get_text_val("define")
         ed = campaign.get_text_val("extra_defines")
-        if ed: define += "," + ed
+        if define is not None: campaign_defines.append(define)
+        if ed is not None: campaign_defines.extend(ed.split(","))
         name = campaign.get_text_val("name", translation = translated.translate)
         sys.stderr.write("Parsing " + name + "...\n")
         campaign.name = name
         all_campaigns[campaign.get_text_val("id")] = campaign
-        wesnoth.parser.parse_text("{campaigns}", defines + "," + define)
+        wesnoth.parser.parse_text("{campaigns}", ",".join(campaign_defines))
         punits[name] = wesnoth.parser.get_all(tag = "units")
         punits[name] += wesnoth.parser.get_all(tag = "+units")
 
@@ -47,7 +46,12 @@ def main():
     for campaign, unitslists in list(punits.items()):
         for unitlist in unitslists:
             for race in unitlist.get_all(tag = "race"):
-                races[race.get_text_val("id")] = race
+                race_id = race.get_text_val("id")
+                if race_id is None: race_id = race.get_text_val("name")
+                if race_id is None:
+                    continue
+                else:
+                    races[race_id] = race
 
     # Go through all units and put them into a dictionary.
     all_units = {}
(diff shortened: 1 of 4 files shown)


data/tools/unit_tree/wiki_output.py
```diff
@@ -16,29 +16,31 @@ def main():
 
     punits = {}
 
-    defines = "NORMAL,ENABLE_ARMAGEDDON_DRAKE,ENABLE_DWARVISH_ARCANISTER," +\
+    base_defines = "NORMAL,ENABLE_ARMAGEDDON_DRAKE,ENABLE_DWARVISH_ARCANISTER," +\
         "ENABLE_DWARVISH_RUNESMITH,ENABLE_ANCIENT_LICH,ENABLE_DEATH_KNIGHT," +\
         "ENABLE_TROLL_SHAMAN,ENABLE_WOLF_ADVANCEMENT,ENABLE_WOSE_SHAMAN," +\
-        "ENABLE_PARAGON,ENABLE_SAURIAN_SPEARTHROWER,DISABLE_GRAND_MARSHAL"
+        "ENABLE_PARAGON,ENABLE_SAURIAN_SPEARTHROWER,DISABLE_GRAND_MARSHAL".split(",")
```
CelticMinstrel, line 22: I think `base_defines` can just be set to `"NORMAL"` now. All the others on this list are related to the removed `[advancefrom]` tag.

data/tools/unit_tree/wiki_output.py
```diff
@@ -16,29 +16,31 @@ def main():
...
     sys.stderr.write("Parsing campaigns...\n")
-    wesnoth.parser.parse_text("{campaigns}", defines)
+    wesnoth.parser.parse_text("{campaigns}", ",".join(base_defines))
     campaigns = wesnoth.parser.get_all(tag = "campaign")
     for campaign in campaigns:
+        campaign_defines = [] + base_defines
         define = campaign.get_text_val("define")
         ed = campaign.get_text_val("extra_defines")
-        if ed: define += "," + ed
+        if define is not None: campaign_defines += define.split(",")
```
CelticMinstrel, line 37: I could be mistaken, but I don't believe `define=` accepts a comma-separated list?

data/tools/unit_tree/wiki_output.py
```diff
@@ -16,29 +16,31 @@ def main():
...
     punits["mainline"] = wesnoth.parser.get_all(tag = "units")
     punits["mainline"] += wesnoth.parser.get_all(tag = "+units")
 
     all_campaigns = {}
     sys.stderr.write("Parsing campaigns...\n")
-    wesnoth.parser.parse_text("{campaigns}", defines)
+    wesnoth.parser.parse_text("{campaigns}", ",".join(base_defines))
     campaigns = wesnoth.parser.get_all(tag = "campaign")
     for campaign in campaigns:
+        campaign_defines = [] + base_defines
```
CelticMinstrel, line 34: Is this a way of cloning the list? I wonder if a slice would be a better way of doing it…

data/tools/wmlunits
```diff
@@ -317,23 +317,23 @@ def list_contents():
             sys.stdout.write("ok\n")
         except wmlparser3.WMLError as e:
             ef = open(logname, "w")
-            ef.write("<PARSE ERROR>\n")
+            ef.write("<PARSE_ERROR>\n")
```
CelticMinstrel, line 320: I don't get the point of adding underscores here. It's just a log to be read by people, right? It's not like it's an actual XML file.

data/tools/unit_tree/wiki_output.py
Slayer95: Sorry. Noted.

data/tools/unit_tree/wiki_output.py
Slayer95: Neat

data/tools/unit_tree/wiki_output.py
Slayer95: Yeah. That's what I'd normally do. However, this file overloads ``+`` for arrays all over the place, so I figured I'd also follow that style.

data/tools/wmlunits
Slayer95: Well, it is an HTML file. ``<PARSE ERROR>`` is not illegal _per se_, but ``</PARSE ERROR>`` is.

CelticMinstrel: This isn't valid as HTML either, so I'd say changing it from one invalid thing to another invalid thing doesn't make much sense…

data/tools/unit_tree/wiki_output.py
CelticMinstrel: I'm not really invested in one or the other method, it just looked a little strange to me.

data/tools/wmlunits
Slayer95: That's a good question, actually. The procedure that generates the actual ``.html`` file elsewhere expected the closing tags to be their own line. Took care of that.

CelticMinstrel: So basically it's passing that fake XML-style tag through something that converts it to HTML?

CelticMinstrel: I feel it should be something like `<p class="...">` but I guess that's out of scope here.

Elvish-Hunter:
Once again, your patch looks good (and thanks to @CelticMinstrel for reviewing it). At the moment, the only improvement I can think of is replacing the `open() ... close()` syntax with context managers. For example, at line 332 this block:
```python
            ef = open(logname, "w")
            ef.write("<INTERNAL ERROR>\n")
            ef.write(repr(e))
            ef.write("</INTERNAL ERROR>\n")
            ef.write("\n</INTERNAL ERROR>\n")
            ef.close()
            sys.stdout.write("failed\n")
```
can be written as
```python
            with open(logname, "w") as ef:
                ef.write("<INTERNAL ERROR>\n")
                ef.write(repr(e))
                ef.write("</INTERNAL ERROR>\n")
                ef.write("\n</INTERNAL ERROR>\n")
            print("failed", file=sys.stdout)
```
The last line takes advantage of `print()` automatically adding a line ending unless the `end=` argument is specified.

Slayer95:
Updated

data/tools/wmlunits
Slayer95: Moved to #7581

Elvish-Hunter:
Thank you. Just a little thing: if no `file=` argument is given, `print()` prints by default to `sys.stdout`. So, `print("failed", file=sys.stdout)` is equivalent to `print("failed")`.
After that, I think this can be merged.

Slayer95:
258173

> Thank you. Just a little thing: if no `file=` argument is given, `print()` prints by default to `sys.stdout`. So, `print("failed", file=sys.stdout)` is equivalent to `print("failed")`.

Indeed.

> By the way, do you already have an account on our forums?

I go by IceSandslash (258173).

PR #7581, opened 2023-05-01
wmlunits: Specify error as class in html reports

Slayer95:
Suggested by @CelticMinstrel

### DIFF
--- a/data/tools/unit_tree/overview.py
+++ b/data/tools/unit_tree/overview.py
@@ -145,7 +145,7 @@ def main(folder):
             for line in text.splitlines():
                 line = line.strip()
                 if line in ["<INTERNAL ERROR>", "<WML ERROR>", "<PARSE ERROR>", "<TIMEOUT ERROR>"]:
-                    htmlerr.write('<p>')
+                    htmlerr.write('<p class="error %s">' % line[1:-1].replace(" ", "-").lower())
                 elif line in ["</INTERNAL ERROR>", "</WML ERROR>", "</PARSE ERROR>", "</TIMEOUT ERROR>"]:
                     htmlerr.write('</p>')
                 else:


ProditorMagnus:
What is the benefit of this change?

Slayer95:
> What is the benefit of this change?

It preserves semantics of the exception. There are a variety of sources (MDN [1](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/class), SO [2](https://stackoverflow.com/questions/56713447/when-is-it-necessary-to-add-class-for-semantics), [3](https://stackoverflow.com/questions/32209520/in-semantic-html-does-the-class-attribute-mean-anything-in-the-absence-of-css-or)) that explain how semantic classes are useful.

Elvish-Hunter:
The string interpolation is used incorrectly here and this is the output I got in an IDLE shell:
```python
>>> line="<INTERNAL ERROR>"
>>> print('<p class="error %s">', line[1:-1].replace(" ", "-").lower())
<p class="error %s"> internal-error
```
Besides fixing it, I suggest using either the `.format()` method or f-strings.

Slayer95:
> The string interpolation is used incorrectly here and this is the output I got in an IDLE shell:

Damn. That proves there's no patch small enough that cannot have bugs.

> f-strings

Does ``wmlunits`` run with Python 3.6? I added the missing ``%`` to stay safe.

CelticMinstrel:
I don't know about f-strings, but .format() should be safe regardless, right? I'm pretty sure that was already in Python 2.

Elvish-Hunter:
> Does `wmlunits` run with Python 3.6? I added the missing `%` to stay safe.

It does. The VM which runs wmlunits has Python 3.9 installed, but if you want to stay safe the `format()` method has been available since 2.6.

Slayer95:
I mean, #7580 was just merged and it introduces a bunch of ``%`` operators, so it doesn't make much sense to use a different approach here imo.

Slayer95:
Yeah, right.

PR #7583, opened 2023-05-02
Implement event priority

### LINKED ISSUE #7582: Support event priority in the WML event pump

ac717f70b8 added event priority support to Lua, which reached Stable in 1.14.

Fast-forward today, and customization of WML events is still very precarious (``[fire_event][data]`` was just introduced in 1.17.6). Support for ``priority`` attribute in ``[event]`` tags would be very helpful and open many possibilities:
- Addons that add high-priority event handlers that use ``[return]`` in order to cancel events.
- Support for reliably refactoring routines from monolithic blocks to modularized customizable events.
- Independence of ``[event]`` execution timing from file/line placements, in case there are several listeners to the same event name.

Slayer95:
For any given event name, events execute in order of decreasing priority.

Priority is a real number, and may be assigned via the `priority` attribute for
the WML ``[event]`` tag, or through the Lua APIs:

- ``wesnoth.game_events.add({priority = number})``
- ``wesnoth.game_events.add_repeating(name, action, [priority])``

Note that delayed variable substitution is not currently supported in the WML attribute.

Fixes #7582

### DIFF
--- a/data/schema/core/addons.cfg
+++ b/data/schema/core/addons.cfg
@@ -126,6 +126,7 @@
 	{SIMPLE_KEY filter_formula string}
 	{DEFAULT_KEY first_time_only bool yes}
 	{DEFAULT_KEY delayed_variable_substitution bool no}
+    {DEFAULT_KEY priority s_real 0.0}
 
 	{FILTER_TAG "filter" unit ()}
 	{FILTER_TAG "filter_second" unit ()}
--- a/src/scripting/game_lua_kernel.cpp
+++ b/src/scripting/game_lua_kernel.cpp
@@ -3906,9 +3906,13 @@ static std::string read_event_name(lua_State* L, int idx)
  * id: Event ID
  * menu_item: True if this is a menu item (an ID is required); this means removing the menu item will automatically remove this event. Default false.
  * first_time_only: Whether this event should fire again after the first time; default true.
+ * priority: Number that determines execution order. Events execute in order of decreasing priority, and secondarily in order of addition.
  * filter: Event filters as a config with filter tags, a table of the form {filter_type = filter_contents}, or a function
+ * filter_args: Arbitrary data that will be passed to the filter, if it is a function. Ignored if the filter is specified as WML or a table.
  * content: The content of the event. This is a WML table passed verbatim into the event when it fires. If no function is specified, it will be interpreted as ActionWML.
  * action: The function to call when the event triggers. Defaults to wesnoth.wml_actions.command.
+ *
+ * Lua API: wesnoth.game_events.add
  */
 int game_lua_kernel::intf_add_event(lua_State *L)
 {
@@ -3916,6 +3920,7 @@ int game_lua_kernel::intf_add_event(lua_State *L)
 	using namespace std::literals;
 	std::string name, id = luaW_table_get_def(L, 1, "id", ""s);
 	bool repeat = !luaW_table_get_def(L, 1, "first_time_only", true), is_menu_item = luaW_table_get_def(L, 1, "menu_item", false);
+	double priority = luaW_table_get_def(L, 1, "priority", 0.);
 	if(luaW_tableget(L, 1, "name")) {
 		name = read_event_name(L, -1);
 	} else if(is_menu_item) {
@@ -3927,7 +3932,7 @@ int game_lua_kernel::intf_add_event(lua_State *L)
 	if(id.empty() && name.empty()) {
 		return luaL_argerror(L, 1, "either a name or id is required");
 	}
-	auto new_handler = man.add_event_handler_from_lua(name, id, repeat, is_menu_item);
+	auto new_handler = man.add_event_handler_from_lua(name, id, repeat, priority, is_menu_item);
 	if(new_handler.valid()) {
 		bool has_lua_filter = false;
 		new_handler->set_arguments(luaW_table_get_def(L, 1, "content", config{"__empty_lua_event", true}));
@@ -3986,6 +3991,10 @@ int game_lua_kernel::intf_add_event(lua_State *L)
 /** Add a new event handler
  * Arg 1: Event to handle, as a string or list of strings; or menu item ID if this is a menu item
  * Arg 2: The function to call when the event triggers
+ *
+ * Lua API:
+ * - wesnoth.game_events.add_repeating
+ * - wesnoth.game_events.add_menu
  */
 template<bool is_menu_item>
 int game_lua_kernel::intf_add_event_simple(lua_State *L)
@@ -3993,6 +4002,7 @@ int game_lua_kernel::intf_add_event_simple(lua_State *L)
 	game_events::manager & man = *game_state_.events_manager_;
 	bool repeat = true;
 	std::string name = read_event_name(L, 1), id;
+	double priority = luaL_optnumber(L, 3, 0.);
 	if(name.empty()) {
 		return luaL_argerror(L, 1, "must not be empty");
 	}
@@ -4000,7 +4010,7 @@ int game_lua_kernel::intf_add_event_simple(lua_State *L)
 		id = name;
 		name = "menu item " + name;
 	}
-	auto new_handler = man.add_event_handler_from_lua(name, id, repeat, is_menu_item);
+	auto new_handler = man.add_event_handler_from_lua(name, id, repeat, priority, is_menu_item);
 	if(new_handler.valid()) {
 		// An event with empty arguments is not added, so set some dummy arguments
 		new_handler->set_arguments(config{"__quick_lua_event", true});
@@ -4011,6 +4021,8 @@ int game_lua_kernel::intf_add_event_simple(lua_State *L)
 
 /** Add a new event handler
  * Arg: A full event specification as a WML config
+ *
+ * WML API: [event]
  */
 int game_lua_kernel::intf_add_event_wml(lua_State *L)
 {
--- a/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_lua1.cfg
+++ b/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_lua1.cfg
@@ -1,0 +1,31 @@
+#####
+# API(s) being tested: wesnoth.game_events.add
+##
+# Actions:
+# Use lua's game_events.add to attach a low-priority "new turn" event.
+# Use lua's game_events.add to attach a high-priority "new turn" event.
+##
+# Expected end state:
+# The test passes when high-priority event is triggered.
+#####
+{GENERIC_UNIT_TEST "events_test_lua_events_add_priority_vs_origin" (
+    [event]
+        name=preload
+        [lua]
+            code=<<
+                wesnoth.game_events.add{
+                    id = 'low',
+                    name = 'start',
+                    priority = -1,
+                    action = unit_test.fail
+                }
+                wesnoth.game_events.add{
+                    id = 'high',
+                    name = 'start',
+                    priority = 1,
+                    action = unit_test.succeed
+                };
+            >>
+        [/lua]
+    [/event]
+)}
--- a/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_lua2.cfg
+++ b/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_lua2.cfg
@@ -1,0 +1,21 @@
+#####
+# API(s) being tested: wesnoth.game_events.add_repeating
+##
+# Actions:
+# Use lua's game_events.add_repeating to attach a low-priority "new turn" event.
+# Use lua's game_events.add_repeating to attach a high-priority "new turn" event.
+##
+# Expected end state:
+# The test passes when high-priority event is triggered.
+#####
+{GENERIC_UNIT_TEST "events_test_lua_events_add_repeating_priority_vs_origin" (
+    [event]
+        name=preload
+        [lua]
+            code=<<
+                wesnoth.game_events.add_repeating('start', unit_test.fail, -1)
+                wesnoth.game_events.add_repeating('start', unit_test.succeed, 1)
+            >>
+        [/lua]
+    [/event]
+)}
--- a/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_wml1.cfg
+++ b/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_wml1.cfg
@@ -1,0 +1,28 @@
+#####
+# API(s) being tested: [event]priority=
+##
+# Actions:
+# Set X to 1 in the prestart event.
+# Add a low-priority start event handler.
+# In a new high-priority start event: Check if X is 1, Set X to 2.
+# In the added low-priority start event handler: Check if X is 2.
+##
+# Expected end state:
+# The test passes when the low-priority event is executed last.
+#####
+{GENERIC_UNIT_TEST "events_test_priority_vs_origin" (
+    [event]
+        name = prestart
+        {VARIABLE X 1}
+        [event]
+            name = start
+            {RETURN ({VARIABLE_CONDITIONAL X equals 2})}
+        [/event]
+        [event]
+            name = start
+            priority = 1
+            {ASSERT ({VARIABLE_CONDITIONAL X equals 1})}
+            {VARIABLE X 2}
+        [/event]
+    [/event]
+)}
--- a/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_wml2.cfg
+++ b/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_wml2.cfg
@@ -1,0 +1,29 @@
+#####
+# API(s) being tested: [event]priority=
+##
+# Actions:
+# Set X to 1 in the prestart event.
+# In a start event handler with negative priority: Check if X is 1, Set X to 2.
+# In a start event handler with the same priority: Check if X is 2.
+##
+# Expected end state:
+# The test passes when the second event is executed last.
+#####
+{GENERIC_UNIT_TEST "events_test_same_priority" (
+    [event]
+        name = prestart
+        {VARIABLE X 1}
+        [event]
+            name = start
+            priority = -1
+            {ASSERT ({VARIABLE_CONDITIONAL X equals 1})}
+            {VARIABLE X 2}
+        [/event]
+        [event]
+            name = start
+            priority = -1
+            {ASSERT ({VARIABLE_CONDITIONAL X equals 2})}
+            {SUCCEED}
+        [/event]
+    [/event]
+)}
--- a/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_wml3.cfg
+++ b/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_priority_wml3.cfg
@@ -1,0 +1,169 @@
+#####
+# API(s) being tested: [event]priority=
+##
+# Actions:
+# Use event handlers to append alphabet letters to X.
+# Each event handler increases Y by 1.
+# Each event handler checks the current value of X, Y vs expectation.
+##
+# Expected end state:
+# The test passes when X is `abcdefghijklmnopq` and Y is 17.
+#####
+{GENERIC_UNIT_TEST "events_test_multi_int" (
+    [event]
+        name = prestart
+        {VARIABLE X "a"}
+        {VARIABLE Y 1}
+        [event]
+            id = final-check
+            name = start
+            priority = -999999
+            {ASSERT ({VARIABLE_CONDITIONAL Y equals 17})}
+            {RETURN ({VARIABLE_CONDITIONAL X equals abcdefghijklmnopq})}
+        [/event]
+        [event]
+            id = step-3
+            name = start
+            priority = 13
+            {ASSERT ({VARIABLE_CONDITIONAL Y equals 3})}
+            {ASSERT ({VARIABLE_CONDITIONAL X equals abc})}
+            {VARIABLE_OP X formula "'$X'..'d'"}
+            {VARIABLE_OP Y add 1}
+        [/event]
+        [event]
+            id = step-5
+            name = start
+            priority = 8
+            {ASSERT ({VARIABLE_CONDITIONAL Y equals 5})}
+            {ASSERT ({VARIABLE_CONDITIONAL X equals abcde})}
+            {VARIABLE_OP X formula "'$X'..'f'"}
+            {VARIABLE_OP Y add 1}
+        [/event]
+        [event]
+            id = step-2
+            name = start
+            priority = 14
+            {ASSERT ({VARIABLE_CONDITIONAL Y equals 2})}
+            {ASSERT ({VARIABLE_CONDITIONAL X equals ab})}
+            {VARIABLE_OP X formula "'$X'..'c'"}
+            {VARIABLE_OP Y add 1}
+        [/event]
+        [event]
+            id = step-1
+            name = start
+            priority = 15
+            {ASSERT ({VARIABLE_CONDITIONAL Y equals 1})}
+            {ASSERT ({VARIABLE_CONDITIONAL X equals a})}
+            {VARIABLE_OP X formula "'$X'..'b'"}
+            {VARIABLE_OP Y add 1}
+        [/event]
+        [event]
+            id = step-4
+            name = start
+            priority = 12
+            {ASSERT ({VARIABLE_CONDITIONAL Y equals 4})}
+            {ASSERT ({VARIABLE_CONDITIONAL X equals abcd})}
+            {VARIABLE_OP X formula "'$X'..'e'"}
+            {VARIABLE_OP Y add 1}
+        [/event]
+        [event]
+            id = step-8
+            name = start
+            priority = 1
+            {ASSERT ({VARIABLE_CONDITIONAL Y equals 8})}
+            {ASSERT ({VARIABLE_CONDITIONAL X equals abcdefgh})}
+            {VARIABLE_OP X formula "'$X'..'i'"}
+            {VARIABLE_OP Y add 1}
+        [/event]
+        [event]
+            id = step-6
+            name = start
+            priority = 7
+            {ASSERT ({VARIABLE_CONDITIONAL Y equals 6})}
+            {ASSERT ({VARIABLE_CONDITIONAL X equals abcdef})}
+            {VARIABLE_OP X formula "'$X'..'g'"}
+            {VARIABLE_OP Y add 1}
+        [/event]
+        [event]
+            id = step-9
+            name = start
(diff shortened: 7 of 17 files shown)


src/scripting/game_lua_kernel.cpp
```diff
@@ -3900,22 +3900,35 @@ static std::string read_event_name(lua_State* L, int idx)
 	}
 }
 
+static int read_opt_event_priority(lua_State* L, int idx)
```
CelticMinstrel, line 3903: I don't get why this needs to be a separate function…? It's only used once…

Slayer95: I didn't find that API. Sounds like just what is needed.

Slayer95: There was a reason why I couldn't find it. That API has been deprecated since Lua 5.3, and apparently Wesnoth's compilation flags don't include it.

Pentarctagon:
You need to add them to `wml_test_schedule` as well.

gfgtdf:
I currently wonder whether it would  be better to make it a float, for example in the case that addon A defines events with priority 1 and 0, then an addon B could still have an event that happens between then by giving it priority 0.5.

Otherwise looks good to me so far

gfgtdf:
I _think_ this misses code to write the priority when serializing the wml event, but its also possible that i'm not up to date with how the event code works these days.

CelticMinstrel:
You appear to be correct…

src/scripting/game_lua_kernel.cpp
CelticMinstrel: It was renamed to `luaL_optinteger`? Since that's what you ended up using (until changing it to float that is).

CelticMinstrel:
Should it maybe be double instead of float?

src/scripting/game_lua_kernel.cpp
Slayer95: It was removed in favor of ``luaL_optinteger``, which uses the largest available integer type.

Slayer95:
``config`` seems to support ``real`` and ``s_real`` types. I figure the former is ``double`` and the later ``float``? Adding this as ``double`` sounds kinda like a waste of space...

CelticMinstrel:
> `config` seems to support `real` and `s_real` types.

This is the schema, not the config. The schema judges types by text matching (regular expressions); `s_real` is "subst real", meaning "either a real number or a string containing but not ending with `$`". Real numbers are represented by a double in memory in the `config` clas.

Slayer95:
Talking of substitution... 
With the default value of ``delayed_variable_substitution=yes``, ``first_time_only`` doesn't seem to accept variables. Variable ``first_time_only`` with dvs can be worked around by wrapping the ``[event]`` tag in two alternate conditional paths.

However, there doesn't seem to be an obvious way to work around that for ``priority``, and it would be unfortunate not to support variable priority. Maybe ``priority`` should be a tag with attributes similar to ``[set_variable]`` ?

Pentarctagon:
Is there a need for it to support variables?

Victoriis (approved):

CelticMinstrel:
I'm pretty sure `first_time_only` supports the use of variables at the time of event registration. That means that variables won't be parsed if it's a toplevel event, but for nested events, variables will be parsed at the time the event is added.

I'm pretty sure the same will apply to `priority` without any extra work. If you wanted the ability to expand `priority` on the fly so that an event handler's priority can change after it's been registered, that would probably take extra effort. At this point I don't think that much extra effort is worthwhile, but if you disagree you're free to make the attempt.

If I'm wrong on the current behaviour of `first_time_only`, it's probably not that hard to get that behaviour somewhere in `wesnoth.wml_actions.event` or the APIs it calls, and the same process would then apply to `priority`. That's independent of the extra effort mentioned in the previous paragraph.

soliton-:
Is there any concept on how this should be used? The direction of making the priority a float sounds a little crazy. How would an UMC author decide what priority numbers to use?

gfgtdf:
> Is there a need for it to support variables?

Imo not, i really think we shouldn't add support for delayed_variable_substitution unless we ha e a concrete usecase that cannot be solved otherwise (in fact i think supporting delayed variable substitution in 'name' was already a bad idea)

> How would an UMC author decide what priority numbers to use?

He'd use the default, unless there is another event that he explicitly wants to be executed before/after 

> Is there any concept on how this should be used?

See the discussion in #7582

Slayer95:
> I'm pretty sure first_time_only supports the use of variables at the time of event registration. That means that variables won't be parsed if it's a toplevel event, but for nested events, variables will be parsed at the time the event is added.

Yeah, it works fine with ``delayed_priority_substitution=no``. But that's not the default value.

Delayed priority is a concept supported in all the Pokémon games released in the last five years. It's true that it involves more work. However, it's apparently more intuitive to players.

For example, if two competing events are associated to actors that will attack or otherwise influence each other, a third event may modify their speed on the fly and make one of them attack earlier.

But at this point it might (or might not, still gotta try) be more work than I can handle. So it would be nice to just have it as a tag so the feature can be incrementally improved. Similarly, it could be possible to have several ``priority`` tags each associated to a different event name.

Slayer95:
In fact, two features related to delayed variable substitution seem to have been mentioned:
- ``delayed_variable_substitution=yes``, and compute variable ``priority``/``first_time_only`` as such, delayed. This is what I talk about in the comment above re: Pokémon.
- ``delayed_variable_substitution=yes``, but compute variable ``priority``/``first_time_only`` at attach time. This would probably be the most common use case for variables: __do this at priority N__, __do that afterwards (N-1)__, __do something else earlier (N+1)__.

> The direction of making the priority a float sounds a little crazy. How would an UMC author decide what priority numbers to use?

It would be good practice for authors to use the smallest number set needed. By which I mean, only 0 if possible, else add positive ints, else negative ints, else floats. In particular, floats should be generally reserved for interactions between multiple addons.

CelticMinstrel:
I think you're misunderstanding something. `delayed_variable_substitution` has nothing to do with this case. It's not even a feature of the event system. All it does is tell `wesnoth.wml_actions.event` to not substitute variables when the event handler is registered, which means it has no effect whatsoever in a toplevel event.

Variables is always evaluated when an event handler is run, but at present that only applies to the `name` key and to the ActionWML tags. Variables are also evaluated when an event registers another event via WML, unless `delayed_variable_substitution` is set to `yes`.

So, substituting the priority at registration time is something that should already work (but requires `delayed_variable_substitution=no` in a nested event). Substituting the priority when the handler is triggered takes extra effort and is completely compatible with the other form, since it requires `delayed_variable_substitution=yes` _or_ a toplevel event.

In other words:

> `delayed_variable_substitution=yes`, but compute variable `priority`/`first_time_only` at attach time.

This would not be a thing, if I understand your meaning (maybe I misunderstood though?). Computing the priority from a variable at registration time would mean you're using `delayed_variable_substitution=no`.

If I _am_ misunderstanding, then I don't know what the difference is between your two cases.

Slayer95:
> In other words:
>> delayed_variable_substitution=yes, but compute variable priority/first_time_only at attach time.
This would not be a thing, if I understand your meaning.

I am talking about variables in a generic sense, not necessarily ``$substituted``. That's why I propose ``[priority]``. It could support:

```css
[event]
name="happening"
delayed_variable_substitution="yes"
[priority]
mode="early_variable"
name="my_priority"
[/priority]
[/event]
```

```css
[event]
name="happening"
delayed_variable_substitution="yes"
[priority]
# With attribute priority, equivalent to priority="$my_priority"
mode="delayed_variable"
name="my_priority"
[/priority]
[/event]
```
```css
[event]
name="happening"
delayed_variable_substitution="no"
[priority]
# With attribute priority, equivalent to priority="$my_priority"
mode="early_variable"
name="my_priority"
[/priority]
[/event]
```


```css
[event]
name="happening"
delayed_variable_substitution="no"
[priority]
mode="delayed_variable"
name="my_priority"
[/priority]
[/event]
```

```css
[event]
name="happening"
[priority]
mode="literal"
value="1"
[/priority]
[/event]
```

> delayed_variable_substitution has nothing to do with this case. It's not even a feature of the event system

> in fact i think supporting delayed variable substitution in 'name' was already a bad idea

AFAICT, the problem here is that event metadata is not properly isolated from the event actions. In principle, there is no reason why a single ``delayed_variable_substitution_`` parameter should determine the behavior of both the event actions and the event name.

A better approach would have been tag-based names with support for ``mode=literal``, and ``mode=early_variable``, and drop dvs in event names.

Pentarctagon:
While I don't feel *strongly* about it, I'm starting to feel like this is getting too complicated again:
* Making the priority a float is a mis-feature. If you're setting priority to make things work as you want within your own add-on(s), then you have full control over the priority values and can already set them whatever they need to be. If you're using a fractional priority to work around someone else's add-on with whom you're not cooperating to make the event things work, then that's incredibly fragile and IMO isn't something that should be encouraged.
* Going out of our way to add support for variable priority so that event handlers can be run not only in a user-defined ordering but also in a variable and even potentially an entirely random ordering seems like it would end up being more confusing than anything else.

Having a single `priority=<some integer value>` is very simple and easy to understand even to someone new to WML - it's practically self-documenting. Having to consider `delayed_variable_substitution` and a `mode` for how the priority is evaluated is essentially the opposite of that.

CelticMinstrel:
> That's why I propose `[priority]`.

This proposal seems extremely overcomplicated. Four of the five examples you give can already be expressed with just a priority key, so why complicate things by making it a tag?

1. is `priority=$|my_priority` (but you have to change `delayed_variable_substitution` to `no`)
2. is `priority=$my_priority`
3. is `priority=$my_priority`
4. is `priority=$|my_priority`
5. is `priority=1`

Only the first one doesn't work – if you need delayed variable substitution on the event actions but want the priority to be immediate, that's not supported. However, it could easily be supported on nested events with a small modification to `wesnoth.wml_actions.event`.

I don't think the real value is overcomplicating it, however – a use-case of priority might be a modification that wants an event to run relative to events in another add-on which you don't have control of. I'm not sure whether variable priority (in the sense of what's already done with `name`) is overcomplicating things; it does sort of make sense to support it for consistency with `name`, but it doesn't seem like a crucial feature.

Slayer95:
> If you're using a fractional priority to work around someone else's add-on with whom you're not cooperating to make the event things work, then that's incredibly fragile and IMO isn't something that should be encouraged.

In principle that's true, but the beauty of open source is that often people will figure out ways to do things. For instance, it would not be "incredibly fragile" to keep supporting an addon that gets yearly updates or rarer. As for the degree of cooperation that is optimally required, exposing APIs would be ideal. Do you think it would be too wild to think of ``$AtSHappeningPriority`` as a valid API? Of course, to use that, early variable substitution for in ``priority`` is a requirement.

> entirely random ordering

Not sure what you mean by that - WFL is not within the proposal. The idea would be just accessing the value of a variable (which could of course be altered in unpredictable ways, but independently of the Event System.)

> if you need delayed variable substitution on the event actions but want the priority to be immediate, that's not supported. However, it could easily be supported on nested events with a small modification to wesnoth.wml_actions.event.

It's good that it's easy to implement, because it should be supported out of the box imo. Remember that dvs is the default for all events so a newcomer to the event system may try to use a variable and then wonder why the hell it doesn't work.

> I'm not sure whether variable priority (in the sense of what's already done with name) is overcomplicating things [...] it does sort of make sense to support it for consistency with name, but it doesn't seem like a crucial feature.

Agreed. I just think we should allow API design space for it. I have also explained an specific use case for variable priority.

I can also imagine some way in which delayed variable substitution can be leveraged for ``name``, but I'd love to know whether it's actually being used with
a) Actual **variables**, or
b) **Constants** (i.e. never-changed scalar variables), which means that ``delayed_variable_substitution=yes`` is used either because it's the default or because of its effects in the ActionWML tags, and dvs support in ``name`` just happens to make those constant-variables work).

gfgtdf:
>  but I'd love to know whether it's actually being used with

Afaik _all_ mainlne cases that uses variables in the [event]name= key explicitly use delayed_variable_substitution=no, because usually you don't want delayed_variable_subsitution in name=. However it's very likely that there is at least one addon that uses name= with actual variables

Pentarctagon:
> In principle that's true, but the beauty of open source is that often people will figure out ways to do things and keep doing them. For instance, it would not be "incredibly fragile" to keep supporting an addon that gets yearly updates or rarer.

I think that Wesnoth itself has a new stable branch released every 2 years or so and there are **always**  problems and complaints about things that got deprecated, removed, or just don't work quite the same.

If the two add-on authors are already cooperating, then they can already sync up which events need which priorities. If they aren't cooperating, then author 1 changes the variable value or usage for their own reasons and it'll be weeks or months until someone figures out why some strange bug suddenly started happening in author 2's add-on.

> As for the degree of cooperation that is optimally required, exposing APIs would be ideal. Do you think it would be too wild to think of $AtSHappeningPriority as a valid API? Of course, to use that, early variable substitution in priority is a requirement.

IMO, that sounds absolutely horrible. A globally readable/writable variable is not an API, or if it does technically count as one then it's just about the worst possible way to create an API.

Slayer95:
> A globally readable/writable variable is not an API, or if it does technically count as one then it's just about the worst possible way to create an API.

If it were a container variable ``AtS_API``, then usage like ``AtS_API.AlignmentPriority`` would be virtually indistinguishable from what countless platforms and services provide.
- The fact that it's global seems ugly, but that's what WML offers, y'know, and scripting is traditionally heavily reliant on globals (Lua, Python, JS all use globals by default). 
- Of course an API needs to be readable?
- The variable may be writable, but that needn't affect the behavior of an event defined by author 1 with ``delayed_variable_substitution=no``. If the event code of author 1 is not secured, then author 2 may actually provide _fixes_ to bugs. On the other hand, if author 2 is rogue, then... Would their changing the priority of random events be the worst concern, really?
-

CelticMinstrel:
> WFL is not within the proposal

Actually, it is. WFL is supported automatically _everywhere variable substitution is supported_. So anything that supports variables supports WFL. That said, the WFL doesn't have access to anything external so there are limits on what you can do with it (most commonly it's used in conjunction with substituting additional variables as in your example, or with macros).

That said, allowing WFL more directly via `typed_formula<double>` is another option. That would allow adding whatever custom variables make sense to the formula inputs. The syntax would be `priority="(some formula)"` instead of the `priority=$(some formula)` that is automatically supported as part of variable expansion.

> But unless I am missing something, I don't think `is priority=$|my_priority` would work? It would fall within the `name-like-dvs` features.

It works for case 4 where you need variable substitution to _not_ be delayed for everything else, but need it to be delayed for just the priority. It doesn't fully work for case 1.

> I can also imagine some way in which delayed variable substitution can be leveraged for `name`, but I'd love to know whether it's actually being used with
> a) Actual **variables**, or
> b) **Constants** (i.e. never-changed scalar variables), which means that `delayed_variable_substitution=yes` is used either because it's the default or because of its effects in the ActionWML tags, and dvs support in `name` just happens to make those constant-variables work).

I think variables in event names are somewhat common, usually events such as `name="turn $($time_of_that_event + 3)"` which would guarantee the event fires whenever the current turn is 3 greater than whatever is stored in `$time_of_that_event`. I'm not aware of any cases that substitute things other than turn or side numbers into the event name.

Pentarctagon:
@Slayer95 The point is that these are edge cases, and it doesn't seem to be worth it to add significant complexity just for sake of addressing them. Even needing to use just the plain `priority` attribute would be a fairly rare edge case already, much less `[priority]`.

Slayer95:
> Is there some reason why you're putting quotes around your numbers in the unit tests? That seems strange.

Been trying to rule out whether the tests are wrong. So far it seems that I am mishandling floats somewhere.

> Actually, it is. WFL is supported automatically [...] more directly via typed_formula<double> is another option [...]

Good to know those are/can be seamlessly added.

>  rare edge case already, much less [priority]

OK, I've been convinced that ``[priority]`` is mostly redundant given the ``$|`` substitution syntax.

Slayer95:
From the commit message:

> Note that delayed variable substitution is not currently supported in the WML priority attribute.

That can be implemented incrementally later. So, I think, unless further concerns are raised, this should be ready to go.

stevecotton:
> As for the degree of cooperation that is optimally required, exposing APIs would be ideal. Do you think it would be too wild to think for author 1 to expose `$AtSHappeningPriority` as a valid API? Then author 2 would use it like:
> 
> ```
> 
> {VARIABLE injected_priority "$($AtSHappeningPriority + 0.5)"}
> [event]
> priority="$injected_priority"
> [/event]
> ```

Events can already fire other events by name, and the engine will handle the existence of any number (including zero) of events by that name before returning to the original event. So if the add-ons already have the level of cooperation needed for an injected priority, AtS's `[event]name=happening` could instead start by firing a `name=pre_happening` event.

That doesn't cover all the possibilities, but it covers some of them without needing the complexity of a priority feature.

Slayer95:
But what if there is no cooperation? Then author 2 himself (or a third author) could scrape or otherwise look up the priorities and publish them as a dependency. For author 2 it would simply be an abstraction layer.

It's true that if two addons cooperate or even just a single author is insightful enough, very beautiful APIs will be developed, but not every possibility can be covered. For instance, if there are two different addons that want to use ``pre_happening``, then which one gets its turn first? Alphabetically according to how addons are (likely) to be loaded? That's just bad.

gfgtdf:
Looks good to me i think, some minor issues: I'd probably returns by value in the `priority()` getter but i guess that counts as personal preference.

Pentarctagon:
@Slayer95 I think you need to more clearly define what your goals are for this feature.

gfgtdf:
It's pretty clear imo, sometimes you have handler for the same event for the same events, and want them to execute in a certain order. These can be within one add on or from different add ons. World Conquest II for example uses this already in its lua, events, the same reasons why WC2s lua code uses it probably also apply to other umc codes. Furthermore it allows easier coordination between different addons listeing to the same event.

Pentarctagon:
Right, and your WC example is a pretty straight forward use case. But the topic that seems to keep coming up is "but what if I want to mess with event priority in some arbitrary add-on that I'm not cooperating with", and I'm not sure why that's a relevant consideration here. If you try to do that, then fine, but you do it at your own risk.

Slayer95:
... @Pentarctagon , you seem to be favorable to the basic concept of priority.

Regarding so-called "messing with priority", that's just a possibility that is opened up and that cannot be covered by currently available methods (i.e. without this PR), and is therefore a counter-argument to @stevecotton 's objections .

Whether such a feature is wanted or not doesn't affect the current _minimal_ implementation up to review. It's minimal because of such concerns about alleged complexity added by this feature.

I'll also add that it's my view that priority is a basic building block for moddable event-based game engines, and there are thus a wide variety of purposes for which it can be used. When there are multiple ways in which such purposes can be achieved, that means the engine is powerful and expressive, and of course everyone can use the available tools to build content, be it semantic WML, macros, custom WML events, Lua, etc.

gfgtdf:
I still want to point out that wc2 uses it even in 1.17 (So its not only used to work around compatibilty issues), for example i have 3 recruit events:
1) On the first "recruit" of a scenario WC2 shows the invest dialog
2) I want to add the units training buffs on recruit.
3) The "Full movement when recruited/recalled" ability uses the recruit event.

Since (2) might want to add the ability implemented by (3) It'd be nice if there was a way make sure (2) is executed before (3). (this is currently implemented via on_events priority parameter)
Since (1) might add a training i want to execute (1) before (2) (I currently can't because (1) is a wml event and (2) is a lua on_event)

Also all of these events make sense on their own, none of these need the other to work, the ordering is only needed when all 3 are used together like in WC2. What i want to say here is that any implementation like "one of these fires a custom event that the others react to" isn't really what i want (also it'd be more complex than just having priority attribute from the umc authors point of view).

stevecotton:
The basic concept of priority does seem to make sense. I'm still not convinced that floating point is better than signed integer.

> On the first "recruit" of a scenario WC2 shows the invest dialog

I think that's a legacy of which events were available when WC was written. The UX feels as if it should be shown after the player has been able to look around the map, but before they move, recruit or recall; instead it's shoehorned in after showing the recruit dialog once and after choosing which unit to recruit.

The animation of the recruit spawning happens first, which is presumably a legacy of being written before the `prerecruit` and `prerecall` events were added. However, even `prerecruit,prerecall` seems too late for the UX; maybe it wants a new event for when the player first triggers an action (such as recruiting) that's more than just looking at the map.

> the beauty of open source is that often people will figure out ways to do things and *keep doing them*

However the ugliness of open source is that those tricks will be cut & pasted into new code, even if a better way has subsequently been added.

> I'll also add that it's my view that priority is a basic building block for moddable event-based game engines, and there are thus a wide variety of purposes for which it can be used. When there are multiple ways in which such purposes can be achieved, that means the engine is powerful and expressive, and of course everyone can use whatever available tools they prefer to build content, be it semantic WML, macros, custom WML events, Lua, etc.

Are the game engines that you're referring to here shipped with the games themselves? As in, whatever version the game developer used during development is the version that the player ends up using for that specific game?

Supporting multiple ways to do things can mean higher maintenance cost for the engine, and edge-cases that aren't tested until the behavior is already frozen for a stable branch.

Pentarctagon:
What are the remaining issues/questions with this PR?

stevecotton:
For me, it's just the `int` vs `float` question. I might be convinced that `float` is necessary, but at the moment it seems an edge-case.

Pentarctagon:
Does anyone have any comments on the int vs float question?

CelticMinstrel:
I think infinitesimalness is pretty important for a priority system. Sure, a float isn't _technically_ infinitesimal, but it's closer to that ideal than an integer.

Note that when I say "float" I don't necessarily mean the type that C calls `float`, but rather "a floating-point type". I'd probably say `double`.

stevecotton:
I meant to use #7631 for that question, but in the end it didn't get discussed there.

I feel that integers with a guideline such as "use a multiple of 100 or 1000 unless there's a reason not to" is going to cover almost any usecase that doubles would cover. That would match the design of the MicroAI's CA ranking system. It would also match the campaign ranking system, and while we are finding problems with the campaign ranking system, they seem to stem from choosing too small a step size for the "multiples of 5" guideline.

With integers, there's no doubt when two values are meant to be exactly equal. There's an argument that that property is true for doubles too, because they'll always have come from the same string-to-double function; but that feels like a fragile thing to rely on if anyone starts doing calculations to fit between other add-ons' published priorities. That argument feels to me like "doubles have that property, as long as you stick to usecases that would be satisfied with integers".

If we use integers for 1.18 and find we need the extra granularity, then it's easy to switch to doubles for 1.20. OTOH, once we use doubles in the API then we can't switch back to floats, as it will break any add-on that uses 0.5 or 1.5.

CelticMinstrel:
> I feel that integers with a guideline such as "use a multiple of 100 or 1000 unless there's a reason not to" is going to cover almost any usecase that doubles would cover.

Well, real numbers mean you can nominally squeeze something in no matter what multiple you choose. Obviously that's not strictly true for floating-point but it's probably close enough.

> With integers, there's no doubt when two values are meant to be exactly equal.

In this case the values are primarily meant to be put in order, so I don't think exact equality testing is even a required feature. That said…

> There's an argument that that property is true for doubles too, because they'll always have come from the same string-to-double function;

Yeah. Floating-point equality only breaks down once you start doing math on them.

> but that feels like a fragile thing to rely on if anyone starts doing calculations to fit between other add-ons' published priorities.

If they do a calculation in their head and come up with a number, that's the same as not doing a calculation, so I don't quite see the issue here. Are you expecting people to be using `$(...)` formulas in the priority? Although, even if you were, those don't even support floating-point calculations.

stevecotton:
> Are you expecting people to be using `$(...)` formulas in the priority?

Exactly that was suggested [earlier in this PR's discussion](#issuecomment-1533547735).

CelticMinstrel:
> > Are you expecting people to be using `$(...)` formulas in the priority?
> 
> Exactly that was suggested [earlier in this PR's discussion](#issuecomment-1533547735).

Okay, fair point. But the only floating-point math involved in that scenario is a division by 1000. Floating-point math is still deterministic, so if you divide the same integer by 1000 a million times, the results will all compare equal.

gfgtdf:
> that feels like a fragile thing to rely on if anyone starts doing calculations to fit between other add-ons' published priorities

We do already rely on floating point calculations though, in particular umc lua code that does any math computations does rely on matching double computations along different cpus (otherwise oos happens).

We also do some work to make sure that this is the case here:
https://github.com/wesnoth/wesnoth/blob/5cd12df0af7c3fe0ae4de94bdb2b2d30e89b30a2/src/wesnoth.cpp#L674

and here:
https://github.com/wesnoth/wesnoth/blob/64dcc3feb76a1252116d90c1139cdb6f37b8bfa2/src/random.cpp#L32

CelticMinstrel:
> We are already relying on matching floating point calculations though, in particular any umc lua code that does any math computations does rely on matching double computations along different cpus (otherwise oos happens).

That's not the issue with floating-point math at all. The issue is that _different_ calculations that _should_ give the same answer _don't_, due to rounding.

> although i have to admit that i don't really understand how the second one can determine at compile time what floatingmath mode the cpu supports

I'm pretty sure that's not querying the CPU's capabilities. It's asking the compiler whether it's using IEEE 754-compliant instructions to implement `double` arithmetic. The compiler knows what floating-point math is supported because it knows what its target CPU is.

gfgtdf:
> That's not the issue with floating-point math at all. The issue is that different calculations that should give the same answer don't, due to rounding.

Well, floats not being perfect maths is nothing new. And in particular in the case of event priorities same values would mean "i don't care what the order is since they are independent" so either one coming first should be fine (there is no such thing as events being executed at the same time).

stevecotton:
Although the technical discussion about floating-point math is interesting, it seems to have veered off the task of deciding whether floating-point should be used by this specific PR.

I think the discussion has solved the question for @newfrenchy83's PRs, because abilities with exactly the same priority could be an expected event when two units of the same type meet.

Pentarctagon:
If Wesnoth already relies on consistent floating point results for other things, then that aspect seems moot here. As long as 1.66 is never evaluated to be greater than 1.67, that's really all that's needed.

data/schema/core/addons.cfg
```diff
@@ -126,6 +126,7 @@
 	{SIMPLE_KEY filter_formula string}
 	{DEFAULT_KEY first_time_only bool yes}
 	{DEFAULT_KEY delayed_variable_substitution bool no}
+    {DEFAULT_KEY priority s_real 0.0}
```
stevecotton, line 129: For consistency with the rest of the file, this line should use a tab instead of spaces. Or alternatively, we fix #5709, but that causes merge conflicts with several open PRs (probably, I haven't checked how may of the current PRs touch the schema).

stevecotton:
I concede that I'm outvoted on the float-vs-int debate. Just a trivial whitespace change in the schema and then we merge?

Pentarctagon:
@Slayer95 are you able to fix the whitespace inconsistency, or should I go ahead and merge this and then fix it myself right after?

Pentarctagon:
Alright, so barring anything else being said, I'll go ahead and merge this in a few days.

PR #7585, opened 2023-05-02
Refactor Statistics

gfgtdf:
Previously statistics were stored in global variables, now it is a part of saved_game. With this saved_game now finally represent the contents of a safefile as it was intended to, without needing to fill the statistics part in some global variable.

In particular now no longer have to manually reset the statistics as random parts of the code, it gets reset along with the saved_game object. Also it is now in theroy possible for multiple saved_game objects to exist.

Statistics was split in two objects, the statistics_record which only contains the data, and statistics_t, which provides methods to modify statistics during a game (to get cleaner dependencies)


Pentarctagon:
And the statistics still work as expected in various situations - playing a single player game, reloading a game, online multiplayer, single/multiplayer campaigns, etc?

gfgtdf:
sure

PR #7588, opened 2023-05-04
Reorder WML includes so manual tests can use the unit-test macros

stevecotton:
Move the line that includes data/test/scenarios/manual_tests/ below the line that includes data/test/macros/. This allows a manual test to use the setup in COMMON_KEEP_A_B_C_D_UNIT_TEST, putting units in the right position to test combat and leadership abilities.

Also improve comments in data/test/_main.cfg.

### DIFF
--- a/data/_main.cfg
+++ b/data/_main.cfg
@@ -56,8 +56,10 @@
 #endif
 
 #ifdef TEST
-# load tests that need manual user input to run and validate
-{test/scenarios/manual_tests}
+# Manual demonstrations or tests of the AI code. Most of these are showcases of the available
+# functionality, explaning what a parameter or Micro AI does. They can be run by binding a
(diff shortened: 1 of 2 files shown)


CelticMinstrel:
Some additional notes that may or may not be worth adding in your comments:

* The manual tests can also be run from within the GUI using the binding to the Test hotkey and pressing it at the main menu.
* The AI scenarios (in ai/scenarios and ai/micro_ais/scenarios) are also intended to be run the same as the manual tests, with `-t` or the Test hotkey. They're less tests and more showcases, though.

stevecotton:
Thanks for the suggested extra comments, I've added them.

stevecotton:
There's a data/test/readme.txt file. @CelticMinstrel, which of the two options sounds better to you?
* duplicate the new docs in the readme
* move the new docs to the readme

The text is short enough that I'm leaning towards "duplicate".

PR #7589, opened 2023-05-04
Ability refactor

gfgtdf:
We create a class ability, and make units/unit_type hold a vector of those instead of just a config, this not only allows us to safely add/remove abilities, we could also add more members to abilities.


src/units/ability.cpp
```diff
@@ -0,0 +1,45 @@
...
+	(at your option) any later version.
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY.
+
+	See the COPYING file for more details.
+*/
+
+#include "units/ability.hpp"
+
+void unit_ability_t::parse_vector(const config& abilities_cfg, std::vector<ability_ptr>& res)
```
CelticMinstrel: Is there a reason to take this by reference rather than returning it?

I would prefer returning it unless there is a good reason to do otherwise (or rather, just merge this code into `cfg_to_vector`).

I see it's called in two places beyond this file, but one of those is in a constructor and the other one clears the list first, so neither of them rely on it being a reference parameter.

src/units/attack_type.hpp
```diff
@@ -54,7 +55,27 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 	int num_attacks() const { return num_attacks_; }
 	double attack_weight() const { return attack_weight_; }
 	double defense_weight() const { return defense_weight_; }
-	const config &specials() const { return specials_; }
+	const ability_vector &specials() const { return specials_; }
+
+	/**
+	 * Returns a vector of all specials of the given type, unlike get_specials it
+	 * doesn't check whether  the specials are active.
```
CelticMinstrel, line 62: I think all of these functions should be renamed to make it more clear which ones check whether the specials are active. Also, the function above this one needs a Doxygen comment.

* `specials()` &rarr; `get_all_specials()` – returns all specials, not checking whether they're active
* `specials(std::string)` &rarr; `get_all_specials_of_type(std::string)` – returns specials with a particular tag, not checking whether they're active
* `get_specials(std::string)` &rarr; `get_active_specials_of_type(std::string)` – returns active specials with a particular tag
* `weapon_specials()` &rarr; `get_active_specials()` – not exactly sure if this is one of them, but if it in fact returns all active specials as the name seems to imply, then it should be renamed as mentioned, and if it returns all specials without checking for whether they're active, then it's redundant with `specials()` and should be removed. If it's some third thing, then further discussion is probably needed.

src/units/ability.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include "config.hpp"
+#include "units/ptr.hpp"
+
+#include <string>
+#include <string_view>
+#include <vector>
+
+using ability_vector = std::vector<ability_ptr>;
+
+class unit_ability_t
```
CelticMinstrel, line 27: I'm not quite sure about this. Is the `_t` suffix consistent with our other classes? It feels inconsistent to me.

src/units/active_ability_list.cpp
```diff
@@ -0,0 +1,352 @@
+/*
+	Copyright (C) 2006 - 2022
+	by Dominic Bolin <dominic.bolin@exong.net>
```
CelticMinstrel, line 3: It's one thing to put Dave in by default, as he's the founder of the project, but filling in anyone else as the copyright seems wrong.

src/units/abilities.cpp
```diff
@@ -128,6 +128,14 @@ A poisoned unit cannot be cured of its poison by a healer, and must seek the car
  * ...
  * [/abilities]
  *
+ * In this file use the following terminology: during an evaulation of an ability there are at most 3 unit involved:
+ * 1)   The Teacher (the unit that own the ability)
+ * 2)   The Student (the unit that is affected by the ability,
+ *      usually the same as the teacher), but might also be a different
+ *      unit via [affect_adjacent]
+ * 2.5) The Opponent of a unit in a fight.
```
CelticMinstrel, line 136: Shouldn't this say "The Opponent of _the student_ in a fight"? That seems a little clearer, as if I understand correctly the student is always the one fighting when student and teacher are not the same unit.

src/units/ability.cpp
```diff
@@ -15,6 +15,28 @@
...
+{
+	//Add wml filter if "backstab" attribute used.
+	if (!cfg_["backstab"].blank()) {
+		deprecated_message("backstab= in weapon specials", DEP_LEVEL::INDEFINITE, "", "Use [filter_opponent] with a formula instead; the code can be found in data/core/macros/ in the WEAPON_SPECIAL_BACKSTAB macro.");
+	}
+	if(cfg_["backstab"].to_bool()){
+		const std::string& backstab_formula = "enemy_of(self, flanker) and not flanker.petrified where flanker = unit_at(direction_from(loc, other.facing))";
+		config& filter_opponent = cfg.child_or_add("filter_opponent");
+		if(!filter_opponent.empty()) {
+			filter_opponent = filter_opponent.add_child("and");
```
CelticMinstrel: This is incorrect. You're attempting to rebind the reference `filter_opponent`, but that's not supported, so instead you're wiping out the filter that you're trying to augment (by assigning a default-constructed config to it). In fact, even the `[and]` tag that you just added will be wiped out.

I suggest just reverting to the code that was originally here, which was correct.

src/units/attack_type.hpp
```diff
@@ -104,7 +104,8 @@ class attack_type : public std::enable_shared_from_this<attack_type>
 	 * @param special_id If true, match @a special against the @c id of special tags.
 	 * @param special_tags If true, match @a special against the tag name of special tags.
 	 */
-	bool has_special(const std::string& special, bool simple_check=false, bool special_id=true, bool special_tags=true) const;
+	bool has_special(const std::string& special, bool special_id=true, bool special_tags=true) const;
+	bool has_special_simple(const std::string& special, bool special_id=true, bool special_tags=true) const;
```
CelticMinstrel, line 108: Needs a Doxygen comment. What exactly makes this simpler from the other `has_special`? Maybe there's a better name for it.

data/test/units/test_attack_dummy.cfg
```diff
@@ -0,0 +1,30 @@
+#textdomain wesnoth-test
+
+
+# A Unit with simple stats for testing attack & ability releated functions..
+
+[unit_type]
+    id="test_attack_dummy"
+    name=_ "dummy_unit^Test Unit Attack"
+    image="misc/blank-hex.png"
```
CelticMinstrel, line 9: I suggest using something other than blank, in case someone wants to debug a test using this with `--showgui`. Perhaps something from the _scenery_ directory?

src/units/ability.cpp
CelticMinstrel: Just a note – it says this diff is outdated, but the same error is still present in the latest code.

src/units/active_ability_list.cpp
Slayer95: This file was split out from ``abilities.cpp`` (formerly ``unit_abilities.cpp``), which was originally attributed to Dominic. And even though he hasn't been active for over 15 years, it doesn't seem entirely wrong to attribute this file to him.

Frankly there have been lots of changes to that file since then, but no full rewrite (as was the case in 72b0cf9bdbf).

src/units/abilities.hpp
```diff
@@ -18,42 +18,3 @@
 #include "map/location.hpp"
 #include "units/ptr.hpp"
 
-class unit_ability_list;
```
stevecotton, line 21: Having both "ability.cpp" and "abilities.cpp" feels like it'll be an source of annoyance when forgetting which one is which. Is there a plan to have only one by the end of the refactor?

Leaving an empty file called "abilities.hpp" while adding "ability.hpp" also seems bad.

I understand the desire to split what's now in ability.cpp from the mish-mash of stuff in abilities.cpp, however I think it's the mish-mash that should move instead.

src/units/ability.cpp
gfgtdf: > one of those is in a constructor

It supports multiple [abilities] tag (idk why, but it was the previous behaviour) so using cfg_to_vector wouold be more complicated there

src/units/attack_type.hpp
gfgtdf: hmm probably, i'm hoping that i can get rid a at least one of them, and will probably do this last then.

src/units/active_ability_list.cpp
gfgtdf: I'll probably remove it , i just copied the file and didn't  think about this.

src/units/abilities.cpp
gfgtdf: yes

src/units/attack_type.hpp
gfgtdf: I the name comes form the simple_check parameter which was there before, but basicially has_special returns actve specials, and simple returns speciasl of theis attack

data/test/units/test_attack_dummy.cfg
gfgtdf: probably

src/units/abilities.hpp
gfgtdf: > Is there a plan to have only one by the end of the refactor?

I was hoping so, not sure yet whether ill actually get there.

src/units/ability.cpp
CelticMinstrel: If that's the case it looks like your refactor might have missed it, because I didn't see anywhere that called it more than once, eg on a loop.

src/units/attack_type.hpp
CelticMinstrel: So it's simple because it doesn't check if the special is active?

src/units/ability.cpp
gfgtdf: In unit::init which is i think the only place that uses it.

stevecotton:
If the refactor will stay as a draft for a long time then it would be nice to cherry-pick the new test.

gfgtdf:
>If the refactor will stay as a draft for a long time then it would be nice to cherry-pick the new test.


I'd say its unlikely that i'll  finish it this month.

data/test/units/test_attack_dummy.cfg
stevecotton: I've taken a copy of this in #7738, renamed it "Melee Test Quintain", and used a sword as the `image`.

gfgtdf:
since the main changes of this have been implemented in #10670 and related prs, im closing this

PR #7591, opened 2023-05-05
Fix insecure SSL/TLS connexion used in campaignserver_client.py

alberic89:
Add a minimum version of protocol for security.


Elvish-Hunter:
While it doesn't hurt to add it, merging this PR will bump the minimum required version of Python to 3.7; is this what we want?
I think we can safely bump the minimum version to 3.8, which is the default one on Ubuntu 20.04 LTS (18.04, which used 3.6, will go in End-of-Life at the end of May) and the latest one available on Windows 7; unless someone can think of a reason to not bump the minimum Python version, it can be merged.

Pentarctagon:
Sounds fine to me.

PR #7596, opened 2023-05-06
Don't allow markup in achievement titles.

CelticMinstrel:
This is done as a guarantee that there will always be one unambiguous way to identify that the achievement is completed.


Pentarctagon:
So this doesn't change anything visually, it just prevents using pango markup in the name?

CelticMinstrel:
> So this doesn't change anything visually

It does change the colour from "khaki" to the title colour.

Pentarctagon (approved):

PR #7600, opened 2023-05-06
Don't use optional_child for widget state definitions when it's mandatory

CelticMinstrel:


gfgtdf:
looks good to me i think

stevecotton:
These strings are based on the names of WML tags, and so it's helpful to use the tag name (even though that looks like English). Even German speakers aren't going to immediately associate "Benötigter Vordergrund für Tafeldefinition fehlt" with a `[panel]` missing its `[foreground]`.

Had I noticed the new translatable string that this PR added before the string-freeze started, I'd definitely want the same change that happened in #8433. @demario-w and @Hejnewar the two of you are generally good people to ask when I wonder "would it still be worth doing something involving translatable strings". So a question to you: it would definitely be good for the teams who haven't translated these yet to remove them completely, and use an configurable error string that already exists (exactly as #8433 did). Do you feel that's the right option, given that we can't go back in time and fix back then?

Another option is to push po hints onto all of these, as in
```
	// TRANSLATORS: an error in GUI WML, with the name of a WML tag written as if it was English text.
	// You could translate "Missing required state for [multimenu_button]" instead.
```

Hejnewar:
Tags should be in english and shouldnt be translated because the code is in english. No need for them to be translatable, probably would only add to confusion.

gfgtdf:
Since those error messages are clearly directed towards developers, it's questionable whether its worth having them translatable, since all other resources (wiki, forums, wml, comments  etc) are in english too.

Also if a user from another language wants to report a bug in an addon, the addon authors might not be able to understand it if its was trasnlated into the players language.

CelticMinstrel:
> Since those error messages are clearly directed towards developers, it's questionable whether its worth having them translatable, since all other resources (wiki, forums, wml, comments etc) are in english too.

I don't entirely agree. Even if they can speak and read English, I'm sure there must be developers who'd be more comfortable seeing errors (and documentation for that matter) in their native language.

> Also if a user from another language wants to report a bug in an addon, the addon authors might not be able to understand it if its was trasnlated into the players language.

This point is valid though… and it's also basically unsolvable. If you drop the assumption that all developers are native English speakers, then that means there's no scenario where the chance of the developer having difficulty reading the error message in a bug report is zero.

Maybe that's one of the reasons why a lot of projects assign numerical codes to each error? Since then you can easily look up the numerical code in the documentation in your language of choice and see it written out perfectly in said language. Could still be a little tricky for messages with more than one placeholder, but still…

PR #7602, opened 2023-05-07
1.17 - more overgrown walls

doofus-01:
More like kudzu than hedges, so name may need revision, but the general idea is there.  The real utility will be in the mixed/transition images; the odds of anyone wanting to make a whole map with this are low, but the basic images are needed first.

Screenshot_20230506_

EDIT: The most important transition images now exist.

Screenshot_20230507_


Pentarctagon:
Yeah, a hedge would have leaves on the top part as well rather than black like the walls. I can't really think of an alternative though - "extra overgrown walls" is just... bad.

CelticMinstrel:
> Yeah, a hedge would have leaves on the top part as well rather than black like the walls.

Though that kinda makes sense, I don't think it's a given. I think it makes sense to use this new wall as a hedge too.

PR #7604, opened 2023-05-07
Refactor TeamColorizer into a Python library

Slayer95:
Reduces subprocess spamming in ``wmlunits``.

### DIFF
--- a/data/tools/TeamColorizer
+++ b/data/tools/TeamColorizer
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+
+import sys, argparse
+from unit_tree.team_colorizer import colorize
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description="""Map the magenta team-color patches in the input image to red (or a custom
+color) in the output image, copy the result to output.""",
+        formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument(
+        "-d", "--dryrun",
+        action="store_true",
+        help="""Print the command to be executed, but don't actually
+generate the output image.""")
+    parser.add_argument(
+        "-v", "--verbose",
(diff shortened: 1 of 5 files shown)


data/tools/TeamColorizer
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
+
+import sys, argparse, subprocess
```
ProditorMagnus, line 3: why import subprocess?

data/tools/TeamColorizer
```diff
@@ -0,0 +1,93 @@
...
+    exit_code = colorize(
+        namespace.color,
+        namespace.input_file,
+        namespace.output_file,
+        magick=namespace.magick,
+        method=namespace.method,
+        hex=namespace.hex,
+        namespace=namespace
+    )
+    if not dry_run and exit_code != 0:
```
ProditorMagnus, line 90: previously dryrun

Slayer95: Good catch

data/tools/TeamColorizer
Slayer95: Leftover

CelticMinstrel:
Why does this use ImageMagick, anyway? It seems to me that it'd make more sense to use Pillow.

Slayer95:
Probably because this was written before Pillow was a thing.

Elvish-Hunter:
Although the script was written before Pillow was made, back then PIL already existed (and Pillow is a fork of it). I suppose that there are some good reasons to not get rid of ImageMagick support (perhaps better image compression?), but it'd be nice to add support for Pillow as well (in another pull request).
That said, I don't see anything really wrong with the code, so I think it can be merged.

PR #7605, opened 2023-05-07
Make wmlunits Windows-compatible

Slayer95:


data/tools/unit_tree/helpers.py
```diff
@@ -155,8 +174,8 @@ def copy_and_color_images(self, target_path):
                     p.wait()
             else:
                 sys.stderr.write(
-                    "Warning: Required image %s does not exist (referenced by %s).\n" % (
-                        image.id_name, ", ".join(image.addons)))
+                    "Warning: Required image %s does not exist t %s (referenced by %s).\n" % (
```
ProditorMagnus, line 177: t?

data/tools/unit_tree/html_output.py
```diff
@@ -12,7 +12,7 @@
 import unit_tree.helpers as helpers
 import wesnoth.wmlparser3 as wmlparser3
 
-PICS_LOCATION = "../../pics"
+PICS_LOCATION = os.path.join("..", "..", "pics")
```
ProditorMagnus, line 15: are you sure this creates valid html?

Slayer95: Yes, it later goes through ``path2url``

data/tools/unit_tree/helpers.py
Slayer95: Oops

data/tools/wmlunits
```diff
@@ -33,12 +33,17 @@ def copy_images():
...
-                         stderr=subprocess.PIPE)
-    out, err = p.communicate()
-    return out
+def shell_out(wesnoth_exe, com):
+    base_com = [wesnoth_exe, "--nobanner"] + com
+    cli_opt_matrix = [[], ["--no-log-to-file"], ["--wnoconsole", "--wnoredirect"]]
+    for cli_opts in cli_opt_matrix:
+        com = base_com + cli_opts
+        p = subprocess.run(com, capture_output=True, text=True)
+        # Wesnoth 1.16 writes to stdout, 1.17 to stderr
```
soliton-, line 42: This is a bug. I'd rather that gets fixed instead of adding workarounds. Also the output is mangled by sanitization so it's currently broken either way.

data/tools/unit_tree/helpers.py
```diff
@@ -40,11 +27,15 @@ def __init__(self, wesnoth_exe, userdir, datadir, magick_exe="convert"):
...
         self.userdir = userdir
-        if not self.userdir:
-            self.userdir = get_userdir(wesnoth_exe)
+        self.hide_paths = [
+            os.path.join(self.userdir, "data", "add-ons"),
+            os.path.join(self.userdir, "data"),
+            os.path.join(self.userdir),
+            os.path.join(self.datadir, "data", "add-ons"),
+            os.path.join(self.datadir, "data"),
+            os.path.join(self.datadir)
```
soliton-, line 37: Are these paths expected to exist somewhere? Why was this changed from the paths used before?

data/tools/unit_tree/helpers.py
```diff
@@ -95,28 +90,31 @@ def add_image_check(self, addon, name, no_tc=False, check_transparent=False):
...
+                    break
+            else:                
+                # root_dir is prefix after normalization
+                root_dir = vpath.split("/")[0].split("\\")[0]
+                if prefix_fallback is None or (root_dir != "general" and root_dir != prefix_fallback):
+                    print("Path \"%s\" outside valid folders." % vpath)
+                    vpath = "403.png"
+
+            encoded_dir_name = base64url.encode_str(os.path.dirname(vpath))
+            sanitized_file_name = re.sub(r'[^a-zA-Z0-9_\.-]' , "", ".".join(os.path.basename(vpath).split("(")[0].split(".")[-2:]))
```
soliton-, line 110: Does the backslash do anything there?

data/tools/unit_tree/helpers.py
```diff
@@ -40,11 +26,16 @@ def __init__(self, wesnoth_exe, userdir, datadir, magick_exe="convert"):
...
-            self.userdir = get_userdir(wesnoth_exe)
+        self.hide_paths = [
+            os.path.join(self.userdir, "data", "add-ons"),
+            os.path.join(self.userdir, "data"),
+            os.path.join(self.userdir),
+            os.path.join(self.datadir, "data", "add-ons"),
+            os.path.join(self.datadir, "data"),
+            os.path.join(self.datadir)
+        ]
+>>>>>>> 8c702a0eda1 (Fixes)
```
soliton-: Since the commits look fine it'd be good to fix this so we don't have to squash merge.

data/tools/wmlunits
Slayer95: There are already released versions that use stderr, so we might as well not pointlessly remove compat with them.

data/tools/unit_tree/helpers.py
Slayer95: Good point  Gotta double check.

data/tools/unit_tree/helpers.py
Slayer95: These are the root dirs that were trimmed in make_name. There shouldn't be functional change in that regard. (But make_name produces different output now anyway.)

Slayer95: Oh, I see what you mean, maybe. The ``get_datadir()`` and ``get_userdir()`` functions are just fallbacks that are not needed, since this constructor is always called with those parameters.

soliton-: As far as I can tell previously the paths were `options.config_dir`, `options.data_dir` and `"data"`.

`os.path.join(self.datadir, "data", "add-ons")` shouldn't exist at all.

data/tools/wesnoth/base64url.py
```diff
@@ -0,0 +1,23 @@
+import base64, re
+
+# Standard Python library provides
+# base64.urlsafe_b64encode, base64.urlsafe_b64encode
```
soliton-, line 4: I guess one should say decode here.

data/tools/wmlunits
Slayer95: You don't mean the sanitization at make_name, do you?

data/tools/unit_tree/helpers.py
Slayer95: Right... Might as well omit it. As for config_dir/data/add-ons, I am also trimming ``add-ons`` in an effort to have the shortest feasible paths (Windows has a nasty max path length of 260 chars.)

data/tools/wmlunits
soliton-: No, I'm just talking about wesnoth's output for `--data-path` and the like.

This is now fixed in wesnoth with 60184ff.

I think it'd be more future-proof to check the exit status for errors instead of "parsing" stderr and use stdout as the return value. (I should add something to CI to make sure --data-path output is not broken again...)

data/tools/unit_tree/helpers.py
soliton-: Alright, sounds good.

PR #7607, opened 2023-05-08
Add support for distinct sub-achievements.

Pentarctagon:
This adds support for having up to 28 distinct sub-achievements within a single achievement. This limit exists since wesnoth's layout isn't smart enough to tell a horizontal listbox to actually use its scrollbar instead of forcing a horizontal scrollbar on the whole window.

Additionally this adds the [set_sub_achievement] and [has_sub_achievement] WML tags and their respective lua functions. [has_sub_achievement] is unsafe for use in MP, for the same reasons that [has_achievement] is.

---

Screenshot from 2023-05-07 19-10-29

### DIFF
--- a/data/gui/window/achievements_dialog.cfg
+++ b/data/gui/window/achievements_dialog.cfg
@@ -3,6 +3,36 @@
 ### Definition of the window for showing achievements and their statuses
 ###
 
+#define SUB_ACHIEVEMENT ID
+	[column]
+		grow_factor = 1
+		border = "all"
+		border_size = 2
+
+		[drawing]
+			id = "sub_icon{ID}"
+			definition = "default"
+
+			width = 36
+			height = 36
+
+			[draw]
+
+				[image]
+					name = "(text)"
+					w = "(min(image_original_width, 36))"
+					h = "(min(image_original_height, 36))"
+
+					{GUI_CENTERED_IMAGE}
+				[/image]
+
+			[/draw]
+
+		[/drawing]
+
+	[/column]
+#enddef
+
 [window]
 	id = "achievements_dialog"
 	description = "Dialog for displaying achievements and their statuses"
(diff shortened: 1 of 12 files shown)


CelticMinstrel:
I realize it's a bit late to be saying this, but I don't think I like the achievements API at all…

Back on topic: this looks mostly fine, but I'm a bit worried about the clash with `[progress_achievement]`. What happens if you have 9 out of 10 sub-achievements and then something calls `[progress_achievement]`? Will it be marked as complete without filling in the final sub-achievement?

Pentarctagon:
> I realize it's a bit late to be saying this, but I don't think I like the achievements API at all…

It hasn't made it into a stable release yet, so the API can still change.

> Back on topic: this looks mostly fine, but I'm a bit worried about the clash with [progress_achievement]. What happens if you have 9 out of 10 sub-achievements and then something calls [progress_achievement]? Will it be marked as complete without filling in the final sub-achievement?

`intf_progress_achievement` in `game_lua_kernel` checks that no sub-achievements are defined.

CelticMinstrel:
Okay, ignoring for now the name "sub-achievement", which sounds awkward and I don't really like…

This basically models the concept of a "collect them all" achievement that doesn't require the collection to happen within a single playthrough, correct? It's equivalent to my concept of storing a bitfield to represent which ones have been collected, though you've used a set instead of a bitfield – same effect, but perhaps easier to understand.

Thinking about it… I wonder if sub-achievements should have a `multiplicity` parameter? Setting `multiplicity=5` would effectively be a shorthand for defining 5 sub-achievements with the exact same parameters (except maybe the ID). Of course, that's only a convenience, as nothing is stopping you from _actually doing that_…

Based on your screenshot, given you have an achievement with sub-achievements `[1], [2], [3], [4], [5]`, and you've achieved `[1]` and `[4]`, what you would see is:

`[1] [4]`

rather than

`[1] [?] [?] [4] [?]`

Is that correct? No placeholder is shown for unachieved sub-achievements? I guess I don't see **that** as a problem, but it does lead to a few more questions…

* Does the order they're shown in depend on the order they're achieved in?
* There's no text associated with the sub-achievements? I feel like they could have a tooltip, maybe.
* There's no way to tell which sub-achievements haven't been achieved, only how many, right? I'm not sure if this is an issue or not.

Pentarctagon:
No sub-achievements are hidden - if they're completed then the image is colored, if they aren't then they're greyed out.

Tooltips don't appear to be supported for [image]. I tried adding it, but nothing appeared when I hovered over a sub-achievement's image.

Pentarctagon:
It looks like it does actually support tooltips - I had just set it in the wrong place when I tried before.

data/multiplayer/achievements.cfg
```diff
@@ -12,6 +12,11 @@
         description_completed= _ "Completed the survival scenario <i>2p - Dark Forecast</i>."
         icon="attacks/blank-attack.png~SCALE(72,72)~BLIT(units/drakes/blademaster.png~RC(magenta>red), 0, 0)~BLIT(misc/achievement-frame-mp.png, 0, 0)"
         icon_completed="attacks/blank-attack.png~SCALE(72,72)~BLIT(units/drakes/blademaster.png~RC(magenta>red), 0, 0)~BLIT(misc/achievement-frame-mp.png, 0, 0)"
+        [sub_achievement]
+            id = hi
+            description = _"asdasds"
```
stevecotton, line 17: Needs a po comment ;D

Issue #3953, opened 2019-02-27
Support in-place run of wmlunits on single addon


ProditorMagnus:
Currently wmlunits works by moving related addon folders from input directory to add-ons folder, evaluating them, and then moving folders back.

When input folder already is add-ons folder, then work could be done without moving folders. It might not be safe or useful for production run, but would be good for running it on locally developed add-ons.

Additionally, running wmlunits on single addon could be made easier - instead of --list followed by --batch there could be argument of addon folder name.

ProditorMagnus:
Running wmlunits even on mainline isnt overly easy. Managed it with
```
#!/bin/bash
# script running on the wesnoth server to update units.wesnoth.org
set -e -u

# must run from the correct directory
dir="C:\Users\Ravana\games\battle-for-wesnoth-1.19.20"

EXTRA_ADDON_OPTIONS=${EXTRA_ADDON_OPTIONS:-}
EXTRA_WMLUNITS_OPTIONS=${EXTRA_WMLUNITS_OPTIONS:-}
TIMEOUT=${TIMEOUT:-200}
addons=$dir/tree_addons
exe=$dir/wesnoth.exe
config=$dir/tree_config
data=$dir
log=$dir/tree_logs
tmpout=$dir/tree_output


TOOLS="C:\Users\Ravana\games\battle-for-wesnoth-1.19.19/data/tools"
export TEMP=$dir/tree_temp

rm -r $log
mkdir $log


cd "$dir"

python3 -u "$TOOLS/wmlunits" -D "$data" -o "$tmpout" --language "es_ES" -w "$exe" \
    -C "$config" -a "$addons" -L "$dir/overview.txt" -B "$dir/overview.txt" \
    -T "$TIMEOUT" $EXTRA_WMLUNITS_OPTIONS > "$log/wmlunits.txt" 2>&1
```
All unit images are broken though. It looks like webp/png conflict, but even the folder that is supposed to contain images has mainly portraits, no units.

cooljeanius:
> Running wmlunits even on mainline isnt overly easy. Managed it with
> 
> ```
> #!/bin/bash
> # script running on the wesnoth server to update units.wesnoth.org
> set -e -u
> 
> # must run from the correct directory
> dir="C:\Users\Ravana\games\battle-for-wesnoth-1.19.20"
> 
> EXTRA_ADDON_OPTIONS=${EXTRA_ADDON_OPTIONS:-}
[...]

PR #7608, opened 2023-05-08
wmlunits: Support filtering add-ons

Slayer95:
In order to process only a relevant subset of the contents of the add-ons folder. Matching add-ons must be installed in the folder defined by --addons.

This works by modifying the output produced by --list, and feeding it seamlessly to --batch.

Add-ons are specified as positional arguments.

See also #3953

### DIFF
--- a/data/tools/wmlunits
+++ b/data/tools/wmlunits
@@ -7,6 +7,7 @@ Run without arguments to see usage.
 """
 
 import argparse
+import glob
 import multiprocessing
 import os
 import queue
@@ -46,6 +47,19 @@ def move(f, t, name):
 
     shutil.move(os.path.join(f, name), t)
 
+
+def get_filtered_addons(filter, addons_folder):
+    matches = set()
+    for input_path in filter:
+        dirname, basename = os.path.split(input_path)
+        if not dirname or dirname == addons_folder:
+            matches.add(basename)
+        else:
+            print("File \"%s\" not found inside the --addons folder)" % input_path)
+            return None
+
+    return list(matches)
+
 _info = {}
 def get_info(addon):
     global _info
@@ -61,12 +75,15 @@ def get_info(addon):
             parser.parse_file(path)
             _info[addon] = parser
         else:
-            print("Cannot find " + path)
+            print(("Cannot find %s - addon must be installed at %s, " +
+                "or copied from a valid installation.") % (path, options.addons))
     except wmlparser3.WMLError as e:
         print(e)
     return _info[addon]
 
 _deps = {}
+
+# global_addons - lists all the addons installed in the folder specified with --addons
 global_addons = set()
 def get_dependencies(addon):
     global _deps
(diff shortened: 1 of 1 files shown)


Elvish-Hunter:
I agree with @soliton- : `--filter` should accept only one argument, which can probably be a regexp. But if, instead of the usual `store`, you use the `append` action, users will be able to use the `--filter` option how many times they like, thus allowing multiple filters. Also, since the filtering is based on paths, you may consider globbing as an easier alternative to regexps (check the `glob` module in the standard library).

Slayer95:
> I agree with @soliton- : `--filter` should accept only one argument, which can probably be a regexp. But if, instead of the usual `store`, you use the `append` action, users will be able to use the `--filter` option how many times they like, thus allowing multiple filters. Also, since the filtering is based on paths, you may consider globbing as an easier alternative to regexps (check the `glob` module in the standard library).

A variadic ``--filter`` option would have allowed to run ``wmlunits`` over several unrelated add-ons, while also seamlessly supporting glob input with help from the shell. However, if the parser accepted only a single non-expanded glob, that would restrict ``wmlunits`` into only running over add-ons sharing a given namespace.

Furthermore, a glob-oriented API would invite users into thinking that they may input add-ons in any random folder, while they are expected to be in the ``--addons`` folder.

So I think the remaining options are either using regexp or making the filtered addons positional arguments.

CelticMinstrel:
> A variadic `--filter` option would have allowed to run `wmlunits` over several unrelated add-ons, while also seamlessly supporting glob input with help from the shell. However, if the parser accepted only a single non-expanded glob, that would restrict `wmlunits` into only running over add-ons sharing a given namespace.

Doesn't this still work if it's `action='append'`? It just means `wmlunits` needs to make sure to insert `--filter` in front of each individual filter pattern.

Slayer95:
> Doesn't this still work if it's action='append' instead of nargs='+'? Other than needing to quote your globs and expand them from Python instead of from the shell.

If I read a command that says ``wmlunits input --filter A --filter B``, I'd expect the program to process (input ∩ A ∩ B). If it said ``wmlunits input --process A --process B``, I'd expect it to work linearly, first A, then B. None of that matches what ``wmlunits`` offers. Then again some people might expect it to only process A, or to only process B. In fact, anything with repeated flags is a no-no from me.

On the other hand, ``wmlunits [redacted] Brave_Wings Tale_of_Vaniyera``... There can hardly be anything clearer to process both addons.

soliton-:
I agree that while multiple --filter options is a possibility it's not convenient. I'm fine with positional arguments as well.

Elvish-Hunter (approved):
The variadic `filter` argument seems the right approach, so I'm going to approve this PR. However, there are some conflicts with the current master branch that need to be fixed, otherwise it cannot be merged.

Slayer95:
Well? @Elvish-Hunter , while I would also originally like --filter, I still think it would be wiser to go with positional arguments.

I just need to clear my backlog of non-Wesnoth stuff, and implement the resolution.

Pentarctagon:

PR #7610, opened 2023-05-08
move the carryover gold handling and dialog to lua 

gfgtdf:
This moves the carryover gold handling to lua, so that it easier to customize  it.


data/test/scenarios/wml_tests/ScenarioWML/SideWML/test_no_sides.cfg
```diff
@@ -0,0 +1,17 @@
+#textdomain wesnoth-test
+# This series of tests checks that scenarios are ending as appropriate.
```
CelticMinstrel, line 2: The comment seems inaccurate

gfgtdf: fixed

src/scripting/game_lua_kernel.cpp
```diff
@@ -1704,6 +1704,7 @@ int game_lua_kernel::impl_current_get(lua_State *L)
 	return_int_attrib("turn", play_controller_.turn());
 	return_string_attrib("synced_state", synced_state());
 	return_bool_attrib("user_can_invoke_commands", !events::commands_disabled && gamedata().phase() == game_data::TURN_PLAYING);
+	return_bool_attrib("is_replaying", play_controller_.is_replay());
```
CelticMinstrel, line 1707: The convention for attributes of `wesnoth.current` is that reading it out makes sense as a phrase – "current map", "current schedule", "current event context", "current user can invoke commands", etc. This one doesn't fit that convention – "current is replaying" doesn't make sense.

Maybe `user_is_replaying`?

CelticMinstrel: Hmm… `wesnoth.scenario.is_replaying` doesn't really make sense… maybe you could put it in`wesnoth.interface` as a function…

Slayer95:
The default ``do_carryover_gold()`` function needs to be split in 4 functions:
1. Calculation of gold bonus. It should return (*) a table that, for each side, contains:
- A boolean signaling whether they get a bonus.
- Total gold bonus.
- Gold bonus per turn.
2. Gold increase. It should accept the table from (1) and make it effective.
3. Calculation of bonus message. It should accept the table from (1) (or a side subtable?), and return the customizable parts as a table.
4. Display of bonus message. It should accept the table from (3) and show it.

Using events for carryover looks good. However, putting the entire ``do_carryover_gold()`` main routine as an event handler may not be that good (**). Instead, I wonder if it would be a palatable approach to fire events as the first instruction in ``(1)``, ``(2)``, ``(3)``, ``(4)``. These events can then be intercepted by user code that mimicks their respective APIs, and that calls ``wesnoth.cancel_action()``. Then, if I am reading this correctly, the return value of ``fire()`` should be a tuple (***), and its second element should tell the main routine whether the event handlers called ``cancel_action``.

(*) By ``return`` I mean setting these values as attributes in a public table or push them to the Lua stack.
(**) If the ``cancel_action`` approach is followed, the main routine would seemingly share the same pump instance with its subevents, rendering the modularization useless.
(***) At least from the C++ side. That information doesn't seem to be pushed to the Lua stack at ``game_lua_kernel::intf_fire_event`` -but it should.

Slayer95:
Another event-based approach would rely on the other API mentioned in #7582 : aborting subsequent event handlers to the current event (maybe ``wesnoth.game_events.extinguish()``). Then, it could be used for all of ``1``, ``2``, ``3``, ``4``, and even the main ``do_carryover_gold()`` routine titself were it also an event handler (as is in the current state of the PR.) Either way, it seems the event system needs further improvements to be usable.

gfgtdf:
> The default do_carryover_gold() function needs to be split in 4 functions ... wonder if it would be a palatable approach to fire events ...

While i agree that parts of the code could be designed to be a bit more reusable, i also don't think its worth to write a event based  framework here, in particular if it makes it harder for the main usecase i had in mind when writing this, which are campaigns that want to completely replace the carryover gold mechanic, (no gold carried over at all, and probably reward quick finishes in another way (maybe items or free exp), or campaigns that have a score system)

Slayer95:
>  don't think its worth to write a event based framework here

Understandable.

Not sure what sort of API you have in mind. However, a baseline approach could be to make ``do_carryover_gold()`` call ``scenario.do_carryover_gold`` if defined, else 
``do_carryover_gold_default()``.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,146 @@
...
+	end
+end
+
+
+wesnoth.game_events.add {
+	name = "scenario_end",
+	id = "carryover_gold",
+	first_time_only = false,
+	--priority = -1000,
+	action = function()
```
CelticMinstrel, line 143: Just `action = do_carryover_gold` should suffice, right?

gfgtdf: i was considering making the variable `do_carryover_gold` non local, so that it could be replaced by umc. That's why i added this indirection

CelticMinstrel: Hmm… someone wanting to replace it entirely could just unregister this event. I guess what you say makes sense if they want to mildly alter the behaviour, maybe…

gfgtdf:
So i split the function into smaller parts, since the parts are then kinda public api, we now have to worry about making them pretty.

data/lua/carryover_gold.lua
gfgtdf: hmm i also wasn't sure how cleanly unregistering an event handler that was set in preload (and by preload i also mean toplevel lua) works, but i guess you just have to unregister them at preload aswell.

data/lua/scenario_end_events.lua
```diff
@@ -0,0 +1,11 @@
+carryover = wesnoth.require("carryover_gold.lua")
```
CelticMinstrel, line 1: This should be local.

data/lua/carryover_gold.lua
```diff
@@ -47,83 +52,108 @@ local function get_popup_text_basic()
 	return title, report
 end
 
-local function do_carryover_gold()
+--- Calculates the amount of carryover gold.
```
CelticMinstrel, line 55: Please make these documentation comments conform to the format of the documentation comments in other modules, for example `location_set` or `functional`. That means using `@param` and `@return` annotations in addition to 3 dashes for the brief description as you've done here.

data/lua/carryover_gold.lua
```diff
@@ -19,17 +22,19 @@ local function get_num_persistent_teams()
...
 end
 
-local function get_num_villages()
+--- Gets the number of villages on the map.
+function carryover.get_num_villages()
 	return #(wesnoth.map.find{ gives_income = true })
 end
 
-local function half_signed_value(num)
+--- Like tostring(num) but uses a prettier minus sign, to be used in the UI.
```
CelticMinstrel, line 30: This is clearly not true, it's just returning `tostring(num)` directly.

Also, this seems like something that probably belongs in either the `mathx` or `stringx` module, not in here.

gfgtdf: Tbh i don't even know why it's called that i just copied the name from the c++ function, but yeah the implementation is not there yet, idk whether it'd be really useful another places or whether other places usually want to use the normal minus sign.

CelticMinstrel: If it's a C++ function, we could just make this a thin wrapper around the C++ function. We should at least consider changing the name though, as it doesn't make sense to me.

I also think it should return a `t_string`, rather than a regular Lua string.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,167 @@
...
+	-- Note that both strings are the same in English, but some languages will
+	-- want to translate them differently.
+	if side.carryover_add then
+		if side.carryover_gold > 0 then
+			goldmsg = _("You will start the next scenario with $gold on top of the defined minimum starting gold.",
+				"You will start the next scenario with $gold on top of the defined minimum starting gold.", side.carryover_gold)
+
+		else
+			goldmsg = _("You will start the next scenario with the defined minimum starting gold.",
+				"You will start the next scenario with the defined minimum starting gold.", side.carryover_gold)
```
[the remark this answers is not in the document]
gfgtdf: Why do you think so?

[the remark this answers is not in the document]
gfgtdf: Oh right now i see it

gfgtdf:
Now that 1.18 is branched of i'll probably merge this the next days

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
+local _ = wesnoth.textdomain "wesnoth"
+local carryover = {}
+
+--- Returns true iff there is not a single side that is controlled by the local human player.
+---@return boolean
+function carryover.get_is_observer()
```
CelticMinstrel, line 6: This seems like something that should be moved into `wesnoth.interface`.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+	for _, side in ipairs(wesnoth.sides) do
+		if side.controller == "human" and side.is_local then
+			return false
+		end
+	end
+	return true
+end
+
+--- Show names when thre are multiple persistent teams.
+---@return integer
```
CelticMinstrel, line 16: Please add a comment explaining the return value, like this:

`---@return integer #the number of things`

data/lua/carryover_gold.lua
CelticMinstrel: This still applies now.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+	else
+		return tostring(num)
+	end
+	return tostring(num)
+end
+
+--- Calculates the amount of carryover gold.
+---@param side side
+---@param turns_left integer
+---@return number (finishing_bonus), number (finishing_bonus_per_turn)
```
CelticMinstrel, line 47:
```suggestion
---@return number #total finishing bonus
---@return number #finishing bonus per turn
```

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+function carryover.calculate_finishing_bonus(side, turns_left)
+	local num_villages = carryover.get_num_villages()
+
+	local finishing_bonus_per_turn = num_villages * side.village_gold + side.base_income
+	local finishing_bonus = math.ceil(side.carryover_bonus * finishing_bonus_per_turn * turns_left)
+	return finishing_bonus, finishing_bonus_per_turn
+end
+
+---@return boolean
```
CelticMinstrel, line 56: Please add a comment, either with `#` at the end of this line or with `---` on a separate line above, that explains what `get_report_visible` does.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+function carryover.get_report_visible()
+	local is_observer = carryover.get_is_observer()
+	local is_replay = wesnoth.current.is_replaying
+	--return true
```
CelticMinstrel, line 60: This probably needs to be deleted.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+	return wesnoth.scenario.end_level_data.carryover_report and (not is_observer) and (not is_replay)
+end
+
+--- Whether to show the sides names in the scenario end report.
```
CelticMinstrel, line 64:
```suggestion
--- Whether to show the sides' names in the scenario end report.
```

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+---@return boolean
+function carryover.get_show_side_names()
+	local num_persistent_teams = carryover.get_num_persistent_teams()
+	return num_persistent_teams > 1
+end
+
+
+--- returns a stub to show in the scenario end dialog.
+---@return string (title), string (report)
```
CelticMinstrel, line 73:
```suggestion
---@return string #title
---@return string #report
```

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+	else
+		title = _("Defeat")
+		report = report .. _("You have been defeated!")
+	end
+	return title, report
+end
+
+-- returns the first part of the carryover mesage, explain remaining gold
+---@param side side
+---@param info table
```
CelticMinstrel, line 95: Please specify the format of this table.

There are two ways to do this. You can do it inline like this:

`{field1: type, field2: type}`

Or on the lines above you can add a block of comments like this:

```lua
---@class some_descriptive_name_for_the_table
---@field field1 type explanation of this field
---@field field2 type explanation of this field
```

If there's no obvious name for the type, `remaining_gold_options` or `remaining_gold_info` should be fine.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+---@return string
+function carryover.remaining_gold_message(side, info)
+	return "<small>\n" .. _("Remaining gold: ") .. carryover.half_signed_value(side.gold) .. "</small>"
+end
+
+---@param side side
+---@param info table
```
CelticMinstrel, line 102: Please specify the format of this table too. It looks like it has 3 optional fields? See my previous comment for how.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+	if info.finishing_bonus_per_turn ~= nil then
+		res = res .. "<small>" .. _("Early finish bonus: ") .. info.finishing_bonus_per_turn .. _(" per turn") .. "</small>\n"
+	end
+	res = res .. "<small>" .. _("Total bonus: ") .. info.finishing_bonus .. "</small>"
+
+	return res
+end
+
+---@param side side
+---@param info table
```
CelticMinstrel, line 118: Please specify the format of this table. Also, add a comment explaining what the function does.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+		title = _("Victory")
+		report = report .. "<b>" .. _("You have emerged victorious!") .. "</b>"
+	else
+		title = _("Defeat")
+		report = report .. _("You have been defeated!")
+	end
+	return title, report
+end
+
+-- returns the first part of the carryover mesage, explain remaining gold
```
CelticMinstrel, line 93: This comment isn't descriptive enough. "explain remaining gold" isn't an explanation, it's just restating the function name.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+
+---@param side side
+---@param info table
+---@return string
+function carryover.total_gold_message(side, info)
+	return "<small>" .. _("Total gold: ") .. carryover.half_signed_value(side.gold + info.finishing_bonus) .. "</small>"
+end
+
+---@param side side
+---@param info table
```
CelticMinstrel, line 125: Please specify the format of this table. Also, add a comment explaining what the function does.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+---@return string
+function carryover.percentage_remaining_message(side, info)
+	return "<small>" .. _("Carryover percentage: ") .. side.carryover_percentage .. "</small>"
+end
+--- returns the last part of the gold report for a single side, describing how much gold is added in the next scenario.
+---@param side side
+---@param info table
```
CelticMinstrel, line 132: Please specify the format of this table. Also, add a line break before this comment block.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+			side.carryover_gold)
+	end
+
+	-- xgettext:no-c-format
+	return goldmsg:vformat{ gold = tostring(side.carryover_gold) }
+end
+
+--- returns the default gold report for a single side.
+---@param side side
+---@param info table
```
CelticMinstrel, line 165: Please specify the format of this table. Also, the comment could be improved.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+	if side.carryover_gold > 0 then
+		report = report .. "\n" .. carryover.percentage_remaining_message(side, info)
+	end
+
+
+	report = report .. "\n\n" .. carryover.carryover_mesage(side, info)
+	return report
+end
+
+---@return boolean
```
CelticMinstrel, line 186: Please add a comment indicating what this function does. I realize it may seem obvious from the name, but I'd like to make it explicit too.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+function carryover.has_next_scenario()
+	return wesnoth.scenario.next ~= "" and wesnoth.scenario.next ~= "null" and wesnoth.scenario.end_level_data.proceed_to_next_level
+end
+
+---@return table
+function carryover.turns_left()
```
CelticMinstrel, line 192: This says it returns a table but it actually returns an integer.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+	return math.max(0, wesnoth.scenario.turns - wesnoth.current.turn)
+end
+
+--- sets the sides carryover gold and returns information that is used in the report message
```
CelticMinstrel, line 196:
```suggestion
--- sets the side's carryover gold and returns information that is used in the report message
```

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+---@param side side
+---@return table
```
CelticMinstrel, line 198: Please specify the format of this table.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+	side.carryover_gold = math.ceil((side.gold + finishing_bonus) * side.carryover_percentage / 100)
+
+	return {
+		turns_left = turns_left,
+		finishing_bonus = finishing_bonus,
+		finishing_bonus_per_turn = finishing_bonus_per_turn,
+	}
+end
+
+function carryover.do_carryover_gold()
```
CelticMinstrel, line 212: Why is this function undocumented?

src/playsingle_controller.cpp
```diff
@@ -394,7 +394,7 @@ void playsingle_controller::do_end_level()
 	}
 
 	persist_.end_transaction();
-	carryover_show_gold(gamestate(), is_observer() || is_replay(), is_observer(), saved_game_.classification().is_test());
+	// carryover_show_gold(gamestate(), is_observer() || is_replay(), is_observer(), saved_game_.classification().is_test());
```
CelticMinstrel, line 397: Please delete.

src/scripting/game_lua_kernel.cpp
CelticMinstrel: This still needs to be addressed.

CelticMinstrel:
carryover_show_gold.?pp have nothing left in them now, so please delete them. I realize that means editing the Xcode project file, but it's honestly not that hard to remove a file from it – just delete all lines mentioning the file.

Please fix at _least_ everything I mentioned that's not related to documentation before merging. I'd prefer you to fix all the documentation issues too, though.

data/lua/carryover_gold.lua
```diff
@@ -0,0 +1,248 @@
...
+			goldmsg = goldmsg .. _("You will start the next scenario with the defined minimum starting gold.")
+		end
+	else
+		goldmsg = goldmsg .. _(
+			"You will start the next scenario with $gold or its defined minimum starting gold, whichever is higher.",
+			"You will start the next scenario with $gold or its defined minimum starting gold, whichever is higher.",
+			side.carryover_gold)
+	end
+
+	-- xgettext:no-c-format
```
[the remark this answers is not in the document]
gfgtdf: pretty sure i coped it from the c++ code so not sure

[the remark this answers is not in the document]
CelticMinstrel: If that's the case, remove it. xgetext is the tool we use to gather translatable strings from C++ files, but it's not used on Lua files.

gfgtdf:
hmm iirc this mostly waits for https://github.com/wesnoth/wesnoth/pull/9033 at this point.

PR #7620, opened 2023-05-10
Simplify logging options

Pentarctagon:


run_wml_tests
```diff
@@ -170,8 +170,7 @@ class WesnothRunner:
                 path += "-debug"
         self.common_args = [path, "--nobanner"]
         if os.name == 'nt':
-            self.common_args.append("--wnoconsole")
-            self.common_args.append("--wnoredirect")
+            self.common_args.append("--no-log-to-file")
```
CelticMinstrel, line 173: Is there any reason why this is Windows-only? Does it have a different effect on Windows? It should be implicit on Mac and Linux due to the `-u` flag also being passed, which would mean adding it explicitly doesn't change anything.

src/log_windows.cpp
```diff
@@ -86,7 +86,7 @@ log_file_manager::log_file_manager(bool native_console)
 		// with the console subsystem flag and that the standard streams are
 		// already pointing to the console.
 		LOG_LS << "Console already attached at startup (built with console subsystem flag?), log file disabled.";
-	} else if(native_console) {
+	} else {
```
CelticMinstrel, line 89: Does this run even when logging to a file? That seems wrong. Or am I misinterpreting something?

src/wesnoth.cpp
```diff
@@ -1103,33 +1102,23 @@ int main(int argc, char** argv)
...
-		} else if(arg == "--wconsole") {
-			terminal_force = true;
-		} else if(arg == "--wnoredirect") {
 			write_to_log_file = false;
 		}
-#endif
 
 		if(arg == "--no-log-to-file") {
-			terminal_force = true;
+			write_to_log_file = false;
```
CelticMinstrel, line 1109: I think the removal of `terminal_force` looks wrong. If one of the "terminal switches" is found, shouldn't `write_to_log_file` be set to `true` instead of `false` on line 1105?

CelticMinstrel:
It feels like there's something missing here… isn't there an NSIS script somewhere? Or is that kept outside the repository? It'll need updating with this PR.

src/log_windows.cpp
Pentarctagon: This function only gets called if logging to file has been disabled.

CelticMinstrel: That seems to imply that it is misnamed though.

Pentarctagon:
NSIS as in the thing used to create the standalone windows installer?

CelticMinstrel:
Yes, I think that was NSIS, wasn't it? There are a couple of other options out there, but as I recall, we were using NSIS.

src/wesnoth.cpp
Pentarctagon: If a terminal switch is found then that would mean it should write output to the terminal, in which case `write_to_log_file` should be false.

run_wml_tests
Pentarctagon: True, I forgot about that.

src/wesnoth.cpp
CelticMinstrel: Oh, right. Never mind then.

Pentarctagon:
There's the cwesnoth.cmd script, but I don't think these options are at all related to the NSIS installer. @loonycyborg correct me if I'm wrong though.

src/log_windows.cpp
Pentarctagon: The `log_file_manager` class?

CelticMinstrel: Yes, if it's meant for logging to a console, not a file, then why is it called `log_file_manager`?

run_wml_tests
Pentarctagon: Updated.

src/log_windows.cpp
Pentarctagon: Updated.

loonycyborg:
Yes. NSIS is used atm. cwesnoth.cmd is installed alongside wesnoth.exe and wesnothd.exe by it.

CelticMinstrel:
Does this PR necessitate changes to the NSIS script? Or does only cwesnoth.cmd need to be updated?

loonycyborg:
> It feels like there's something missing here… isn't there an NSIS script somewhere? Or is that kept outside the repository? It'll need updating with this PR.

https://github.com/wesnoth/wesnoth/blob/master/packaging/windows/Wesnoth.nsi.in

loonycyborg:
Seems it doesn't make any shortcuts using logging options, only installs cwesnoth.cmd

Pentarctagon:
Any objections if I go ahead and merge this then?

loonycyborg:
Seems it updates cwesnoth.cmd too so should be ok.

PR #7621, opened 2023-05-12
Let wesnothd and campaignd build without SDL.

Pentarctagon:


irydacea:
Works for me on Linux (both campaignd and wesnothd build and run).

Would it be possible to backport this to 1.16? It would help me with testing and fixing issues with the 1.16 campaignd in an isolated environment without having to install SDL 2 and all of its unnecessary dependencies.

soliton- (approved):

Pentarctagon:
Huh, it's bizarre that it's a macOS-specific failure.

stevecotton:
It's surprising, but looking at that test, believable. There's a couple of macos #ifdefs in config_cache.cpp:

https://github.com/wesnoth/wesnoth/blob/68a123df01a5013e275dddf6f941251954aba8b6/src/config_cache.cpp#L43-L56

Pentarctagon:
Right, what I'm confused about is how the changes made here would've caused those to give different results. Maybe the `game_config::wesnoth_version.str()` is giving a different value now?

stevecotton:
Looks like we might be pulling in definitions, I guess of `TARGET_OS_IPHONE`, via SDL's include of `TargetConditionals.h`:

https://github.com/libsdl-org/SDL/blob/237348c772b4ff0e758ace83f471dbf8570535e2/include/SDL_platform.h#L73-L96

Pentarctagon:
Yep, looks like that was it.

Pentarctagon:
@irydacea I tried build wesnothd and campaignd with the latest 1.16 and both already build for me without SDL2. So I don't think there's anything to backport?

stevecotton:
I think the `target["IPHONEOS"]` line should be removed or commented in config_cache.cpp too, as it'll no longer be used.

@singalen FYI, this PR has the side-effect of removing some indirect includes of `TargetConditionals.h`..

.github/workflows/ci-main.yml
```diff
@@ -45,23 +45,39 @@ jobs:
...
             cmake)
-              cmake -DENABLE_GAME=true -DENABLE_SERVER=true -DENABLE_CAMPAIGN_SERVER=true \
+              cmake -DENABLE_GAME=true -DENABLE_SERVER=false -DENABLE_CAMPAIGN_SERVER=false \
                 -DENABLE_TESTS=true -DCMAKE_BUILD_TYPE="$CFG" -DCXX_STD="$CXX_STD" \
                 -DEXTRA_FLAGS_CONFIG="-pipe" -DENABLE_STRICT_COMPILATION=true -DENABLE_MYSQL=true \
                 -DENABLE_NLS=false -DFORCE_COLOR_OUTPUT=true -DENABLE_LTO="$LTO" -DLTO_JOBS=2 .
               make conftests
               make VERBOSE=1 -j2
+
+              apt remove -y -qq libsdl2-dev libsdl2-dev libsdl2-mixer-dev
```
sevu, line 74: The 2nd `libsdl2-dev` should probably have been libsdl2-image-dev

CelticMinstrel: Just noting that this same error appears both in the `scons` and the `cmake` builds.

sevu: Actually, since libsdl2-image-dev and libsdl2-mixer-dev both depend on libsdl2-dev, it doesn't make a difference if they are listed here.

Pentarctagon: Fixed.

PR #7622, opened 2023-05-12
HttT: Improve Thoria branch rewards

Slayer95:
There are concerns about the rewards from this campaign branch being disappointing,
so this PR improves that.

- Sergeant gets a third resilient trait (which explains his survival.)
- Book of Wose Lore now also effectively grants intelligent trait. Tailored for a newly acquired low-level unit (e.g. the Sergeant, to ease its journey to Grand Marshall.)
-


data/campaigns/Heir_To_The_Throne/scenarios/20b_Underground_Channels.cfg
```diff
@@ -660,6 +660,10 @@
                                                 forest=1
                                             [/movement_costs]
                                         [/effect]
+                                        [effect]
```
knyghtmare, line 663: Looks okay but the bonus description should be edited to reflect the decrement in max XP required.

knyghtmare (changes requested):

data/campaigns/Heir_To_The_Throne/scenarios/20b_Underground_Channels.cfg
Slayer95: The current description is not too accurate:

> This massive book is filled with the Lore and Legends of the Wose. Any who read it learn deeply of the ways of the twig and leaf, the tree and the forest.

> "Upon finishing the book, $unit.name| has a much greater understanding of the forest, its nature and ways, and all the creatures within it. Unfortunately, as the last page was turned, the book crumbled to dust.

Keywords: understanding -> intelligence; forest -> ambush. But they are both long shots.

nemaara (approved):

sevu:
Having 3 traits is not good because the UI is bad at displaying more than 2.
Fearless is an offensive trait, it increases the damage. How about replacing that by resilient if the unit is too squishy?
The bonus of this unit is not its damage, but that it can support others with leadership. The best trait would actually be intelligent … though, we could say that is taken care of by giving it free experience points.
Resilient brings it on pair with the Spearman in terms of hitpoints.

Slayer95:
> if the unit is too squishy?

The problem isn't exactly that the unit is squishy. It's that the sergeant doesn't quite match up to the Sword of Fire / Void Armor given in the other quests. Any additional bonus works - I just chose resilient because this unit gotta fight _a lot_ in order to reach Grand Marshall within 3 scenarios. Fearless also helps him secure more kills by fighting both day and night.

Wasn't there an active proposal to make loyal units get up to two more traits? Is the barrier just layout?

sevu:
> Is the barrier just layout?

Pretty much. Translations also tend to be longer, so if 3 traits fit in one line the issue is still there.

> to make loyal units get up to two more traits

loyal is the best trait already, so … but the actual reason is the layout.

How about automatically recalling the Sergeant in every scenario.
That causes the unit to be there to collect experience, but regardless of whether the unit manages to reach level 4, it will be one additional unit which one can bring to the battle.

It looks also like that it will be easier to reach level 4 in the future, currently one needs 5+80+180 xp, ~33 level 1 kills
The discussion in the forum about balance changes for 1.18 currently aims to change that to ~1~5+85+105 xp, ~25 level 1 kills

Might also be that it is fearless because the next level takes place underground.

_Still, the two rewards of the C branch are not up to pair._
Both look like not so useful gimmicks.

Slayer95:
See also what @beetlenaut did for their Nightmare difficulty: https://forums.wesnoth.org/viewtopic.php?t=55034 (post from 2021).

> What I did do was change the experienced sergeant into an experienced lieutenant. That should make it just possible to level him up to a level-4 leader in time to be useful--I'm assuming you are using mostly level-3 units at this point. (If you are playing this still, feel free to use the debug commands to give the guy 70 more XP!).

Said 70 XP bonus is potentially (on the higher end) equivalent to the 1.18 balance change. But as mentioned in beetlenaut's post, that would make it "just possible", which doesn't make it a great reward. Furthermore, while you are getting a potential Level 4 unit, you truly get a Level 1 unit (and weaker than a Spearman).

While autorecall also helps to improve the situation on the lower end of what you truly get, that would double down on its loyal trait, with unpredictable effects on the last few scenarios' economy.  So, maybe it would be safer to make the Sergeant a Lieutenant (independently of the General XP change), so that the player gets a more noticeable reward. Because, frankly, anybody would be disappointed by their first meeting with Drakes climaxing in a Level 1 unit as reward.

sevu:
That is also an option.

Slayer95:
@nemaara , thoughts on this?

- ~Sergeant gets a third resilient trait (which explains his survival.)~
- 19c Sergeant is a Lieutenant instead.
- Book of Wose Lore now also effectively grants intelligent trait. Tailored for a newly acquired ~low-level~ unit (e.g. the ~Sergeant~ Lieutenant, to ease its journey to Grand Marshall.)

nemaara:
That's okay with me too.

data/campaigns/Heir_To_The_Throne/scenarios/19c_Cliffs_of_Thoria.cfg
```diff
@@ -716,7 +716,7 @@ Soooo... It is you who sent your subordinates to attack us. Now when we’ve des
         [message]
             speaker=unit
             scroll=no
-            message= _ "It is a young human sergeant, barely more than a boy. He appears badly hurt but still alive."
+            message= _ "It is a young human officer, barely more than a boy. He appears badly hurt but still alive."
```
Slayer95, line 719: There seems to be a wordplay related to the name of Warven being similar to Dwarven, and that dwarves _may_ be confused with human children. Is it worth adding a #po comment? Or might I be overthinking it?

nemaara: Well I don't think it's a problem for me reading it since it's clearly in a name field?

nemaara:
If no more comments I'm merging tomorrow.

sevu:
The `experience=27` meant 5 exp until level up.
This can, but doesn't need to be, increased. At least the reason for it to remain at 27 is no more given.

Maybe 43?

PR #7623, opened 2023-05-12
Rewrite macOS version report again

irydacea:
This version doesn't rely on custom parsing of the version number from the (notably **not meant to be parsed**) `operatingSystemVersionString` property, and explicitly appends the version number components to the reported string instead so we don't get weirdness like `Apple macOS 13.3.1 ((Build` from an `operatingSystemVersionString` value like `Version 13.3.1 (a) (Build 22E772610a)` (found after the early May 2023 rapid response patch to Ventura).

Additionally, this replaces the clunky version comparison to decide whether to use the macOS or OS X branding with the `@available` syntax.

<https://developer.apple.com/library/archive/releasenotes/AppKit/RN-AppKit/index.html> (Under "Runtime Version Check")

(**Note:** 1.16 also needs this patch)

### DIFF
--- a/src/desktop/apple_version.mm
+++ b/src/desktop/apple_version.mm
@@ -37,25 +37,18 @@ namespace apple {
 		// Standard Apple version
 		//
 
-		std::string version_string = "";
-
-		NSArray *version_array = [[[NSProcessInfo processInfo] operatingSystemVersionString] componentsSeparatedByString:@" "];
-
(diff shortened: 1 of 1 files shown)


Pentarctagon (approved):

hrubymar10 (approved):
Thank you! This will work better for sure

PR #7624, opened 2023-05-12
wmlxgettext: Support extracting all textdomains at once 

Slayer95:


data/tools/GUI.pyw
```diff
@@ -10,11 +10,13 @@
 
 # threading and subprocess are needed to run wmllint without freezing the window
 # codecs is used to save files as UTF8
+# gettext provides internationalization and localization (i18n, l10n)
```
Elvish-Hunter, line 13: Why is the localization stuff from PR #7627 in this one too?

data/tools/GUI.pyw
```diff
@@ -460,28 +463,42 @@ It has a context menu and a save file selection dialog."""
...
+        if os.path.exists(current_dir):
+            return askdirectory(initialdir=current_dir)
+        # otherwise attempt to detect the user's userdata folder
+        else:
+            return askdirectory(initialdir=get_addons_directory())
+    def on_browse(self):
+        if self.filetypes is None:
+            directory = self.on_browse_dir()
+        else:
+            directory = self.on_browe_file(self)
```
Elvish-Hunter, line 501: `on_browe_file`? Looks like a typo.

data/tools/pywmlx/state/machine.py
```diff
@@ -13,7 +13,14 @@
 import pywmlx.nodemanip
 import pdb
 
+# Universe - convenient singleton for which
```
Elvish-Hunter, line 16: It looks like some stuff from PR #7540 leaked into this one.

Elvish-Hunter (changes requested):
I found a typo that has to be corrected (`on_browe_file`); also, stuff from two other PRs leaked into this one and needs to be removed.

data/tools/GUI.pyw
Slayer95: Sorry, it seems I rebased against the wrong branch.

data/tools/pywmlx/state/machine.py
Slayer95: No, this is intended.

data/tools/GUI.pyw
Slayer95: Fixed

cooljeanius:
This caused #9131, apparently

Issue #1786, opened 2017-06-13
wmlxgettext aborts after first error


sevu:
The python wmlxgettext script doesn't tolerate errors.
I don't know if this is a good or a bad thing. I think one can argue that it is better.
It has the side-effect that it tells only about one error (the first encountered) each time.

gfgtdf:
I don't know about wmlxgettext. but i have the same problem with travis builds ( they stop after finding an error ina  file instead of trying to compile the other files). So i can tell how annoying this can be.

CelticMinstrel:
I'd prefer if it at least had an option to try and recover from errors when possible...

PR #7625, opened 2023-05-12
wmlxgettext: Improve parser resiliency for malformed files

Slayer95:
- WML files with no translatable strings will no longer
crash the script on unbalanced tags (but translatable will).
- .cfg files belonging to man(1) will no longer crash the script.


Slayer95:
This is related to #1786 .

After this, making ``wmlxgetext`` tolerate _all_ errors is a one-liner, but I'd recommend against that.

PR #7627, opened 2023-05-12
Make GUI.pyw translatable

Slayer95:
- POT file with textdomain ``wesnoth-tools`` is generated with the command:
```console
xgettext --language=Python --output=wesnoth-tools.py.pot --force-po --add-comments=TRANSLATORS --copyright-holder="Wesnoth development team" --msgid-bugs-address="https://bugs.wesnoth.org/" --from-code=UTF-8 --sort-by-file --keyword=_ --keyword=N_ --keyword=sgettext --keyword=vgettext --keyword=VGETTEXT --keyword=_n:1,2 --keyword=N_n:1,2 --keyword=sngettext:1,2 --keyword=vngettext:1,2 --keyword=VNGETTEXT:1,2 data/tools/*.pyw data/tools/*.py
```

### DIFF
--- a/data/tools/GUI.pyw
+++ b/data/tools/GUI.pyw
@@ -10,11 +10,14 @@
 
 # threading and subprocess are needed to run wmllint without freezing the window
 # codecs is used to save files as UTF8
+# locale and gettext provides internationalization and localization (i18n, l10n)
 # queue is needed to exchange informations between threads
 # if we use the run_tool thread to do GUI stuff we obtain weird crashes
 # This happens because Tk is a single-thread GUI
-import sys,os,threading,subprocess,codecs
+import argparse,sys,os,threading,subprocess,codecs
 
+import locale
+import gettext
 import queue
 
 from wesnoth import version
@@ -42,6 +45,60 @@ APP_DIR,APP_NAME=os.path.split(os.path.realpath(sys.argv[0]))
 WESNOTH_ROOT_DIR=os.sep.join(APP_DIR.split(os.sep)[:-2])  # pop out "data" and "tools"
 WESNOTH_DATA_DIR=os.path.join(WESNOTH_ROOT_DIR,"data")
 WESNOTH_CORE_DIR=os.path.normpath(os.path.join(WESNOTH_DATA_DIR,"core"))
+WESNOTH_TRAN_DIR=os.path.join(WESNOTH_ROOT_DIR, "translations")
+
+_=lambda x:x
+
+def set_default_locale():
+    global _
+
+    # TODO: Replace CLI args for a proper locale selection GUI.
+    # More importantly, code to dynamically update the text/layout is missing.
+    parser = argparse.ArgumentParser(
+        description=_("Open a Graphical User Interface (GUI) to WML Tools"),
+        usage=("GUI.pyw [--lang=" + _("LANGUAGE") + "]")
+    )
+    parser.add_argument(
+        '--lang',
+        # TRANSLATORS: ISO 15897 "Procedures for the registration of cultural elements" specifies use
+        # of underscores in the locale id. For example, use en_GB, not en-GB.
+        help=_("Launch GUI.pyw in the given language. Language code must follow ISO 15897 (e.g. en_GB).")
+    )
+    opts = parser.parse_args(sys.argv[1:])
+    if opts.lang is not None:
+        if not os.path.isdir(WESNOTH_TRAN_DIR):
+            showerror(_("Error"),
+                # TRANSLATORS: {0} is "translations", the folder where compiled translation files (.mo) are stored.
+                _("`{0}` folder not found. Please run the GUI.pyw executable packaged with the Wesnoth installation.").format("translations")
+            )
+        try:
+            _=gettext.translation("wesnoth-tools",WESNOTH_TRAN_DIR,languages=[opts.lang],fallback=False).gettext
+        except:
+            showerror(_("Error"),
+                # TRANSLATORS: {0} is the ISO 15897 code for the language selected by the user.
+                _("Locale {0} not recognized.").format(opts.lang)
+            )
+        return
+
+    try:
+        system_locale = locale.getlocale()[0]
+        _=gettext.translation("wesnoth-tools",WESNOTH_TRAN_DIR,languages=[system_locale],fallback=False).gettext
+    except Exception as ex:
+        # Needed for compat with Python <3.10, and/or Windows 7/8.
+        system_locale = locale.getdefaultlocale()[0]
+        _=gettext.translation("wesnoth-tools",WESNOTH_TRAN_DIR,languages=[system_locale],fallback=False).gettext
+
+def on_update_locale(value):
+    if value is None:
+        try:
+            set_default_locale()
+        except:
+            # _ defaults to identity lambda.
+            pass
+    else:
+        pass
+
+on_update_locale(None)
 
 def wrap_elem(line):
     """If the supplied line contains spaces, return it wrapped between double quotes"""
@@ -268,7 +325,8 @@ Self destroys when the tool thread is over"""
         frame=Frame(self)
         frame.pack(fill=BOTH,expand=YES)
         wait_label=Label(frame,
-                         text="{0} is running\nPlease wait...".format(tool),
+                         # TRANSLATORS: {0} is the name of command being executed.
+                         text=_("Running: {0}\nPlease wait...").format(tool),
                          justify=CENTER)
         wait_label.grid(row=0,
                         column=0,
@@ -282,7 +340,7 @@ Self destroys when the tool thread is over"""
                            padx=5,
                            pady=5)
         terminate_button=Button(frame,
-                                text="Terminate script",
+                                text=_("Terminate script"),
                                 image=ICONS["process-stop"],
                                 compound=LEFT,
                                 command=self.terminate)
@@ -330,24 +388,24 @@ If the widget isn't active, some options do not appear"""
         control_key = "Command" if self.tk.call('tk', 'windowingsystem') == "aqua" else "Ctrl"
         # str is necessary because in some instances a Tcl_Obj is returned instead of a string
         if str(widget.cget('state')) in (ACTIVE,NORMAL): # do not add if state is readonly or disabled
-            self.add_command(label="Cut",
+            self.add_command(label=_("Cut"),
                              image=ICONS['cut'],
                              compound=LEFT,
                              accelerator='%s+X' % (control_key),
                              command=lambda: self.widget.event_generate("<<Cut>>"))
-        self.add_command(label="Copy",
+        self.add_command(label=_("Copy"),
                          image=ICONS['copy'],
                          compound=LEFT,
                          accelerator='%s+C' % (control_key),
                          command=lambda: self.widget.event_generate("<<Copy>>"))
         if str(widget.cget('state')) in (ACTIVE,NORMAL):
-            self.add_command(label="Paste",
+            self.add_command(label=_("Paste"),
                              image=ICONS['paste'],
                              compound=LEFT,
                              accelerator='%s+V' % (control_key),
                              command=lambda: self.widget.event_generate("<<Paste>>"))
         self.add_separator()
-        self.add_command(label="Select all",
+        self.add_command(label=_("Select all"),
                          image=ICONS['select_all'],
                          compound=LEFT,
                          accelerator='%s+A' % (control_key),
@@ -409,7 +467,7 @@ class SelectDirectory(LabelFrame):
     def __init__(self,parent,textvariable=None,**kwargs):
         """A subclass of LabelFrame sporting a readonly Entry and a Button with a folder icon.
 It comes complete with a context menu and a directory selection screen"""
-        super().__init__(parent,text="Directory",**kwargs)
+        super().__init__(parent,text=_("Directory"),**kwargs)
         self.textvariable=textvariable
         self.dir_entry=EntryContext(self,
                                     width=40,
@@ -421,13 +479,13 @@ It comes complete with a context menu and a directory selection screen"""
         self.dir_button=Button(self,
                                image=ICONS['browse'],
                                compound=LEFT,
-                               text="Browse...",
+                               text=_("Browse..."),
                                command=self.on_browse)
         self.dir_button.pack(side=LEFT)
         self.clear_button=Button(self,
                                  image=ICONS['clear16'],
                                  compound=LEFT,
-                                 text="Clear",
+                                 text=_("Clear"),
                                  command=self.on_clear)
         self.clear_button.pack(side=LEFT)
     def on_browse(self):
@@ -460,13 +518,13 @@ It has a context menu and a save file selection dialog."""
         self.file_button=Button(self,
                                 image=ICONS['browse'],
                                 compound=LEFT,
-                                text="Browse...",
+                                text=_("Browse..."),
                                 command=self.on_browse)
         self.file_button.pack(side=LEFT)
         self.clear_button=Button(self,
                                  image=ICONS['clear16'],
                                  compound=LEFT,
-                                 text="Clear",
+                                 text=_("Clear"),
                                  command=self.on_clear)
         self.clear_button.pack(side=LEFT)
         self.filetypes = filetypes
@@ -509,12 +567,12 @@ class WmllintTab(Frame):
         super().__init__(parent)
         self.mode_variable=IntVar()
         self.mode_frame=LabelFrame(self,
-                                   text="wmllint mode")
+                                   text=_("wmllint mode"))
         self.mode_frame.grid(row=0,
                              column=0,
                              sticky=N+E+S+W)
         self.radio_normal=Radiobutton(self.mode_frame,
-                                      text="Normal",
+                                      text=_("Normal"),
                                       variable=self.mode_variable,
                                       value=0)
         self.radio_normal.grid(row=0,
@@ -522,9 +580,9 @@ class WmllintTab(Frame):
                                sticky=W,
                                padx=10)
         self.tooltip_normal=Tooltip(self.radio_normal,
-                                    "Perform file conversion")
+                                    _("Perform file conversion"))
         self.radio_dryrun=Radiobutton(self.mode_frame,
-                                      text="Dry run",
+                                      text=_("Dry run"),
                                       variable=self.mode_variable,
                                       value=1)
         self.radio_dryrun.grid(row=1,
@@ -532,9 +590,10 @@ class WmllintTab(Frame):
                                sticky=W,
                                padx=10)
         self.tooltip_dryrun=Tooltip(self.radio_dryrun,
-                                    "Do not perform changes")
+                                    # TRANSLATORS: Explanation for Dry run.
+                                    _("Do not perform changes"))
         self.radio_clean=Radiobutton(self.mode_frame,
-                                     text="Clean",
+                                     text=_("Clean"),
                                      variable=self.mode_variable,
                                      value=2)
         self.radio_clean.grid(row=2,
@@ -542,9 +601,10 @@ class WmllintTab(Frame):
                               sticky=W,
                               padx=10)
         self.tooltip_clean=Tooltip(self.radio_clean,
-                                   "Delete *.bak files")
+                                   # TRANSLATORS: Explanation for Clean.
+                                   _("Delete *.bak files"))
         self.radio_diff=Radiobutton(self.mode_frame,
-                                    text="Diff",
+                                    text=_("Diff"),
                                     variable=self.mode_variable,
                                     value=3)
         self.radio_diff.grid(row=3,
@@ -552,9 +612,10 @@ class WmllintTab(Frame):
                              sticky=W,
                              padx=10)
         self.tooltip_diff=Tooltip(self.radio_diff,
-                                  "Show differences in converted files")
+                                   # TRANSLATORS: Explanation for Diff.
+                                  _("Show differences in converted files"))
         self.radio_revert=Radiobutton(self.mode_frame,
-                                      text="Revert",
+                                      text=_("Revert"),
                                       variable=self.mode_variable,
                                       value=4)
         self.radio_revert.grid(row=4,
@@ -562,15 +623,16 @@ class WmllintTab(Frame):
                                sticky=W,
                                padx=10)
         self.tooltip_revert=Tooltip(self.radio_revert,
-                                    "Revert conversions using *.bak files")
+                                   # TRANSLATORS: Explanation for Revert.
+                                    _("Revert conversions using *.bak files"))
         self.verbosity_frame=LabelFrame(self,
-                                        text="Verbosity level")
+                                        text=_("Verbosity level"))
         self.verbosity_frame.grid(row=0,
                                   column=1,
                                   sticky=N+E+S+W)
         self.verbosity_variable=IntVar()
         self.radio_v0=Radiobutton(self.verbosity_frame,
-                                  text="Terse",
+                                  text=_("Terse"),
                                   variable=self.verbosity_variable,
                                   value=0)
         self.radio_v0.grid(row=0,
@@ -578,7 +640,7 @@ class WmllintTab(Frame):
                            sticky=W,
                            padx=10)
         self.radio_v1=Radiobutton(self.verbosity_frame,
-                                  text="List changes",
+                                  text=_("List changes"),
                                   variable=self.verbosity_variable,
                                   value=1)
         self.radio_v1.grid(row=1,
@@ -586,7 +648,7 @@ class WmllintTab(Frame):
                            sticky=W,
                            padx=10)
         self.radio_v2=Radiobutton(self.verbosity_frame,
-                                  text="Name files before processing",
+                                  text=_("Name files before processing"),
                                   variable=self.verbosity_variable,
                                   value=2)
         self.radio_v2.grid(row=2,
@@ -594,7 +656,7 @@ class WmllintTab(Frame):
                            sticky=W,
                            padx=10)
         self.radio_v3=Radiobutton(self.verbosity_frame,
-                                  text="Show parse details",
+                                  text=_("Show parse details"),
                                   variable=self.verbosity_variable,
                                   value=3)
         self.radio_v3.grid(row=3,
@@ -602,13 +664,14 @@ class WmllintTab(Frame):
                            sticky=W,
                            padx=10)
         self.options_frame=LabelFrame(self,
-                                      text="wmllint options")
+                                      text=_("wmllint options"))
         self.options_frame.grid(row=0,
                                 column=2,
                                 sticky=N+E+S+W)
         self.stripcr_variable=BooleanVar()
         self.stripcr_check=Checkbutton(self.options_frame,
-                                       text="Convert EOL characters to Unix style",
+                                       # TRANSLATORS: Special characters marking end of line.
+                                       text=_("Convert EOL characters to Unix style"),
                                        variable=self.stripcr_variable)
         self.stripcr_check.grid(row=0,
                                 column=0,
@@ -616,7 +679,7 @@ class WmllintTab(Frame):
                                 padx=10)
         self.missing_variable=BooleanVar()
         self.missing_check=Checkbutton(self.options_frame,
-                                        text="Don't warn about tags without side= keys",
+                                        text=_("Don't warn about tags without side= keys"),
                                         variable=self.missing_variable)
         self.missing_check.grid(row=1,
                                  column=0,
@@ -624,7 +687,7 @@ class WmllintTab(Frame):
                                  padx=10)
         self.known_variable=BooleanVar()
         self.known_check=Checkbutton(self.options_frame,
-                                     text="Disable checks for unknown units",
+                                     text=_("Disable checks for unknown units"),
                                      variable=self.known_variable)
         self.known_check.grid(row=2,
                               column=0,
@@ -632,7 +695,7 @@ class WmllintTab(Frame):
                               padx=10)
         self.spell_variable=BooleanVar()
         self.spell_check=Checkbutton(self.options_frame,
-                                     text="Disable spellchecking",
+                                     text=_("Disable spellchecking"),
                                      variable=self.spell_variable)
         self.spell_check.grid(row=3,
                               column=0,
@@ -640,7 +703,7 @@ class WmllintTab(Frame):
                               padx=10)
         self.skip_variable=BooleanVar()
         self.skip_core=Checkbutton(self.options_frame,
-                                   text="Skip core directory",
+                                   text=_("Skip core directory"),
                                    variable=self.skip_variable,
                                    command=self.skip_core_dir_callback)
         self.skip_core.grid(row=4,
@@ -664,7 +727,7 @@ class WmlscopeTab(Frame):
     def __init__(self,parent):
         super().__init__(parent)
         self.options_frame=LabelFrame(self,
-                                      text="wmlscope options")
+                                      text=_("wmlscope options"))
         self.options_frame.grid(row=0,
                                 column=0,
                                 sticky=N+E+S+W)
@@ -674,7 +737,7 @@ class WmlscopeTab(Frame):
                                  sticky=N+E+S+W)
         self.crossreference_variable=BooleanVar() # equivalent to warnlevel 1
         self.crossreference_check=Checkbutton(self.normal_options,
-                                              text="Check for duplicate macro definitions",
+                                              text=_("Check for duplicate macro definitions"),
                                               variable=self.crossreference_variable)
         self.crossreference_check.grid(row=0,
                                        column=0,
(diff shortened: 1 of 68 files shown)


Slayer95:
> Source `po/wesnoth-dev/af.po' not found, needed by target `translations/af/LC_MESSAGES/wesnoth-dev.mo'.

I suppose it's up to me to create all those po files? Is there a script?

Slayer95:
@ivanovic

Pentarctagon:
Translation files are generated as part of the cmake or scons commands at https://github.com/wesnoth/wesnoth/blob/master/.github/workflows/ci-main.yml#L207-L219

CelticMinstrel:
Is it possible to make it output relative paths in the po(t) files?

Slayer95:
> Is it possible to make it output relative paths in the po(t) files?

Good point. That means:

```console
 ./data/tools/pyxgettext -p po/wesnoth-dev -o wesnoth-dev.pot data/tools/*.pyw data/tools/*.py
```

CelticMinstrel:
I think regexes should be a raw string actually…

CelticMinstrel:
The second commit claims to revert a commit that doesn't seem to exist anymore.

sevu:
The `FINDCFG` lists command to fine these files, in which wmlxgettext will look for strings belonging to the textdomain `wesnoth-dev`. So, in this case rather none.

Slayer95:
> The `FINDCFG` lists commands to find those files, in which wmlxgettext will look for strings belonging to the textdomain `wesnoth-dev`. So, in this case it should be rather none.

Since that domain now exists, there's a chance someone uses that domain for CFG/Lua files. It's better to handle that case early than leaving a footgun.

> The second commit claims to revert a commit that doesn't seem to exist anymore.

Oops.

CelticMinstrel:
By the way, that `pygettext` wrapper looks weird, like it's not forwarding any of the parameters at all? `main()` is called without any arguments…

Slayer95:
> By the way, that `pygettext` wrapper looks weird, like it's not forwarding any of the parameters at all? `main()` is called without any arguments…

The parameters are parsed by the actual ``pygettext`` library, so there's no need to pass any parameters. Unless you mean that the defined ``main()`` function is redundant.

CelticMinstrel:
You mean that `pygettext.main` reads them straight from `sys.argv`? I guess that makes sense… still looks weird to be defining `main()` to forward all parameters and then not pass any though.

stevecotton:
Who will benefit from this translation? These tools are for WML authors, which means they already need to handle WML, understand the existing tag names, and read documentation which (except for the parts which are missing entirely) is in English. I feel it's going to add more effort for the teams that are maintaining 100% complete translations, without increasing the number of people who can use these tools.

I'm slightly in favor`*` of adding this as a domain similar to `wesnoth-test`, which allows our make-it-translatable coding style but doesn't have .pot or .po files generated.

Slayer95:
> Who will benefit from this translation? These tools are for WML authors, which means they already need to handle WML, understand the existing tag names, and read documentation which (except for the parts which are missing entirely) is in English.

Do you forget that this is an open source project where people implement what is useful for them? And are you so courageous to assume that everybody understands English better than their native tongue?

> I feel it's going to add more effort for the teams that are maintaining 100% complete translations

Then they'd simply no longer reach 100% completion, until they figure it out. I don't see how that's a problem.

> without increasing the number of people who can use these tools.

How can you assure it?

> I'm slightly in favor* of adding this as a domain similar to wesnoth-test, which allows our make-it-translatable coding style but doesn't have .pot or .po files generated.

This doesn't make sense.

Pentarctagon:
I'd suggest asking the translators to see if they think it's an issue or not - there's the translators mailing list as well as 5 members of the Translators group on Discord.

If it does stay though, I'd suggest changing the textdomain name to `wesnoth-tools`. `wesnoth-dev` is rather undescriptive.

Slayer95:
FTR, they are 77 sentences, with 2.99 words each on average.

Slayer95:
Anyway, I am having a really hard time figuring out how to please CMake, bash, and shell together.

The regexp for the Python definition of ``_`` is:

```console
^ *_ *= *gettext\.translation\( *(['"])${DOMAIN}\1/
```

In order to pass it to grep, gotta escape a quotemark of choice, and all backslashes:
```console
if grep '^ *_ *= *gettext\\.translation\\( *([\'"])${DOMAIN}\\1'
```

And to pass this to cmake, gotta escape all backslashes again, as well as the double quotes:
```console
if grep '^ *_ *= *gettext\\\\.translation\\\\( *([\\'\"])${DOMAIN}\\\\1'
```

Am I missing something?

CelticMinstrel:
> I'm slightly in favor`*` of adding this as a domain similar to `wesnoth-test`, which allows our make-it-translatable coding style but doesn't have .pot or .po files generated.

This doesn't make any sense. GUI.pyw isn't being passed through a schema validator that requires certain things to be translatable.

> Who will benefit from this translation?

Probably people like newfrenchy, I would think. Even if you can understand English, it's likely a lot more comfortable to see a UI in your native language… and there's barely any text here at all, so I doubt it would impact translation completeness that much.

> I'd suggest asking the translators to see if they think it's an issue or not - there's the translators mailing list as well as 5 members of the Translators group on Discord.

I don't understand what you're suggesting to ask them? There's no sense in asking them whether there's a point in translating this stuff. Did you have some other question in mind?

Pentarctagon:
> I don't understand what you're suggesting to ask them? There's no sense in asking them whether there's a point in translating this stuff. Did you have some other question in mind?

The issue seems (?) to be whether this is worth the additional work it would add for translators, so surely the easiest thing to do is directly ask them if it's a problem?

CelticMinstrel:
> > I don't understand what you're suggesting to ask them? There's no sense in asking them whether there's a point in translating this stuff. Did you have some other question in mind?
> 
> The issue seems (?) to be whether this is worth the additional work it would add for translators, so surely the easiest thing to do is directly ask them if it's a problem?

No, I don't think it's worth asking the translators whether we should translate something. That doesn't make any sense at all.

CelticMinstrel:
> Anyway, I am having a really hard time figuring out how to please CMake, bash, and shell together.
> 
> The regexp for the Python definition of `_` is:
> 
> ```
> ^ *_ *= *gettext\.translation\( *(['"])${DOMAIN}\1/
> ```
> 
> In order to pass it to grep, gotta escape a quotemark of choice, and all backslashes:
> 
> ```
> if grep '^ *_ *= *gettext\\.translation\\( *([\'"])${DOMAIN}\\1'
> ```
> 
> And to pass this to cmake, gotta escape all backslashes again, as well as the double quotes:
> 
> ```
> if grep '^ *_ *= *gettext\\\\.translation\\\\( *([\\'\"])${DOMAIN}\\\\1'
> ```
> 
> Am I missing something?

Does cmake not have any kind of raw-string syntax?

Even if it doesn't, I think you can cut out one level of escaping by rewriting your grep command. I think something along the lines of the following might work (with pipefail):

```bash
if cat << EOF | grep
^ *_ *= *gettext\.translation\( *(['"])${DOMAIN}\1/
EOF
```

I'm not sure, but maybe that kind of approach is easier to work out than figuring out what needs to be escaped and double-escaped and triple-escaped.

Pentarctagon:
Could also probably stick it in a shell script file and have cmake call that?

data/tools/GUI.pyw
```diff
@@ -268,7 +278,8 @@ Self destroys when the tool thread is over"""
         frame=Frame(self)
         frame.pack(fill=BOTH,expand=YES)
         wait_label=Label(frame,
-                         text="{0} is running\nPlease wait...".format(tool),
+                         # TRANSLATORS: {0} is the name of command being executed.
```
stevecotton, line 281: This translation comment isn't appearing the generated .pot file.

Slayer95: I haven't pushed the new .pot file after adding this comment yet. Did you test it?

Slayer95: 12d57c090b87d79bca3e0d5618fdf63ff3d6f635

CelticMinstrel:
Come to think of it, I guess this PR merits a new section on [https://wiki.wesnoth.org/GettextForWesnothDevelopers](this page)?

I also want to point out that searching for Python files based on the extension is possible not a great idea. It works for GUI.pyw but fails for basically every other tool in the repository. I'm not sure what kind of priority there is to translating stuff like the `--help` output of `wmllint` and other tools, but I feel like this system should at least support it should we decide to do it at some point.

sevu:
it is not

data/tools/GUI.pyw
stevecotton: I didn't test it, I assumed the PR would show what it was currently generating.

Slayer95:
> I also want to point out that searching for Python files based on the extension is possible not a great idea. It works for GUI.pyw but fails for basically every other tool in the repository.

Indeed. Currently, for scons it reads the shebang, but dealing with cmake is so annoying. I almost want to hardcode the list of files.

data/tools/GUI.pyw
```diff
@@ -42,6 +45,13 @@ APP_DIR,APP_NAME=os.path.split(os.path.realpath(sys.argv[0]))
 WESNOTH_ROOT_DIR=os.sep.join(APP_DIR.split(os.sep)[:-2])  # pop out "data" and "tools"
 WESNOTH_DATA_DIR=os.path.join(WESNOTH_ROOT_DIR,"data")
 WESNOTH_CORE_DIR=os.path.normpath(os.path.join(WESNOTH_DATA_DIR,"core"))
+WESNOTH_TRAN_DIR=os.path.join(WESNOTH_ROOT_DIR, "translations")
+
+try:
+    SYSTEM_LOCALE = locale.getdefaultlocale()[0]
```
stevecotton, line 51: Triggers a deprecation warning with Python 3.11, this is going to be removed in Python 3.13 (python/cpython#90817).

The warning says "Use setlocale(), getencoding() and getlocale() instead"

stevecotton:
> And are you so courageous to assume that everybody understands English better than their native tongue?
>
> ...
>
> FTR, they are 77 sentences, with 2.99 words each on average.

When a workflow requires a language to be used anyway, it's often easier to stay in one language than to switch between two, even when the other language is the native one. When you say it's only 77 sentences, you're not counting the other changes that would be needed to other tools and to documentation in order to make it a single-language workflow in any language other than English.

So, please, who are the target audience that will benefit from this?

> There's no sense in asking them whether there's a point in translating this stuff. Did you have some other question in mind?

For translations that have achieved 100%, each new point-release has a few new strings, so they do a bit of work to get back to 100%. Having new strings either means the extra work of translating them, or the extra work of remembering that 99.8% (or whatever) is the new 100%; this means that ignoring it is in the long term more effort than translating it.

Another question, that I think we should answer ourselves before asking the translators, is "is the current text clear enough for an English-speaking audience?". My answer to that it isn't, at least not to a wide enough audience to be ready as an input to translation. Reviewing the text for an English-speaking audience will result in a lot of these strings changing, so making them translatable before doing that is likely to reignite [the complaints about making busy-work for the translators](https://r.wesnoth.org/t55582).

For example,

* "Ignore refcount 0 on names matching regexp", should probably be described as unused functions and variables, however I haven't reviewed what exactly it does.
* "Ignore files matching the specified regular expression", should say filenames.

> GUI.pyw isn't being passed through a schema validator that requires certain things to be translatable.

Coding standards aren't just for the tools, it's also for the instinct of "I'm in this project, where every user-visible string should be translatable".

> Could also probably stick it in a shell script file and have cmake call that?

I'd suggest using a Python script instead, but yeah.

Slayer95:
Stop acting like you care about translators. The project has easily had a decade without noticeable improvements to the tools translators use. If you care so much about them, go ahead and approve and convince everyone about the benefits of my ``wmlxgettext`` PRs, or do your own contributions, such as implementing the much needed ``en_US`` locale, rather than objecting improvement initiatives. Incidentally, this PR also brings improvements to how wmlxgettext is used.

> So, please, who are the target audience that will benefit from this?

When a workflow requires a language to be used anyway, it's often easier to stay in one language than to switch between two, even when the other language is the native one. When you say it's only 77 sentences, you're not counting the other changes that would be needed to other tools and to documentation in order to make it a single-language workflow in any language other than English.

Swap native language for English and vice versa.

> So, please, who are the target audience that will benefit from this?

Target audience has already been named by @CelticMinstrel . I refuse to engage in further namecalling.

Pentarctagon:
> Stop acting like you care about translators. The project has easily had a decade without noticeable improvements to the tools translators use. If you care so much about them, go ahead and approve and convince everyone about the benefits of my wmlxgettext PRs, or do your own contributions, such as implementing the much needed en_US locale, rather than objecting against improvement initiatives. Incidentally, this PR also brings improvements to how wmlxgettext is used.

Bluntly, we have little idea what tools translators use or how they use them or what can be done to improve them because basically all translation is done siloed per language, with little to no communication between the translators and the rest of the project. There are occasional questions about how a particular word or sentence should be interpreted, but that's generally been about it.

For https://github.com/wesnoth/wesnoth/pull/7570 specifically, my reading of it is that there's still not been a final decision reached on which approach to use, so how can it be merged? More generally for the rest, the python tools maintainer is Elvish_Hunter, and you do need to wait for them to review your PRs.

> Another question, that I think we should answer ourselves before asking the translators, is "is the current text clear enough for an English-speaking audience?". My answer to that it isn't, at least not to a wide enough audience to be ready as an input to translation. Reviewing the text for an English-speaking audience will result in a lot of these strings changing, so making them translatable before doing that is likely to reignite [the complaints about making busy-work for the translators](https://r.wesnoth.org/t55582).

Wesnoth's policy has always been that nobody should be translating the development releases without the expectation that there may be significant amounts of rework, so this is not an immediate concern with this PR at least.

> I refuse to engage in further namecalling.

There hasn't been any name calling here, so I'm not sure what you're referring to.

CelticMinstrel:
> For example,
> 
>     * "Ignore refcount 0 on names matching regexp", should probably be described as unused functions and variables, however I haven't reviewed what exactly it does.
> 
>     * "Ignore files matching the specified regular expression", should say filenames.

You do make a good point that we should review the actual content of these strings to make sure they describe the functionality clearly enough.

Other than that I don't understand what you're trying to say at all.

> Coding standards aren't just for the tools, it's also for the instinct of "I'm in this project, where every user-visible string should be translatable".

That doesn't seem like it justifies your suggestion at all. The statement "every user-visible string should be translatable" means that every user-visible string should be translatable, _not_ that every user-visible string should be marked as translatable but potentially not actually translated.

Pentarctagon:
Also, just as another comment: including the motivation/reasoning for adding a feature is always helpful. From Discord, Toranks (a translator) said:

> Translating GUI.pyw was actually a joint idea of Slayer95 and me the other day

Knowing up front that this was an idea that someone who's currently also translating the game was involved in suggesting this would've pre-emptively addressed some of these questions.

Slayer95:
Wesnoth might not be the most inclusive project on Earth, but I'd think that I don't need to justify how making the project more open by allowing translation of the parts that contributors use is important.

data/tools/GUI.pyw
Slayer95: Unfortunately, ``getlocale()`` is unusable in Windows, and [it's also on the road of deprecation](https://bugs.python.org/issue43557). Until an actual viable alternative exists, this is probably the best option to get well, the _default_ locale.

Pentarctagon:
My point is that when stevecotton (aka octalot) raised a concern about adding the new strings to be translated, instead of replying with

> Stop acting like you care about translators. The project has easily had a decade without noticeable improvements to the tools translators use. If you care so much about them, go ahead and approve and convince everyone about the benefits of my wmlxgettext PRs, or do your own contributions, such as implementing the much needed en_US locale, rather than objecting against improvement initiatives. Incidentally, this PR also brings improvements to how wmlxgettext is used.

you could have simply said something along the lines of "This was an idea that I and one of Wesnoth's translators came up with".

Pentarctagon:
Additionally, as some background context I guess:
* As stevecotton linked, there have been complaints from some translators about there being too much churn in the string changes made. Our stance was, and is, that Wesnoth is a moving target and that we're not going to leave the base text as being worse in order to prevent this kind of churn. That thread was however the most substantive discussion of translations issues that I can recall happening for as long as I've been around though, so it's unfortunate that not a lot ended up coming of it - as I listed in my [second to last post](https://forums.wesnoth.org/viewtopic.php?p=673454#p673454), there were some ideas for improving things but before I spend time on them I'd like some kind of a definite yes or no answer to whether they'd be useful and actually used. From the little I'm aware of, it seems as though each translation team has their own workflow, their own set of tools, etc, and so there's no way for me to know if *anything* would end up actually be helpful to anyone beyond things like trying to add more po comments. That then being another major problem - I don't actually know who I need to ask to get that kind of an answer (and I'm the project lead!). Honestly the closest we've had is the recent Translators group on Discord, as sad as that probably sounds.
* There has also fairly recently been concerns raised in other PRs related to weapon specials and abilities that features (and complexity) are being added without knowing whether anyone actually wants or would use the functionality. This type of concern is (I assume) why stevecotton was asking who the target audience for this would be.

Slayer95:
Alright, I probably overreacted. Please don't mind my hasty words, @stevecotton .

Now, I think your criteria for "anyone" may be dangerously thin. It can and should be expected that someone who proposes and writes a feature will also use it. And if they do, then someone else will. If you are forever waiting for a thread to reach 10 pages of purely supporting comments and then _also_ one of the few devs getting interested in it enough to consider implementing it, actually doing it, and then getting it through review, then nothing will ever get done.

Slayer95:
Ready for review now. I have also added a way for people to launch the GUI in English (or alternative languages), rather than in their system language if so they wish.

data/tools/GUI.pyw
stevecotton: Thanks for improving the code.

Minor correction: `getlocale()` doesn't seem to be on the road to deprecation. That's proposed in the first message of that thread, however it gets voted against, and the same person who proposed it then adds the deprecation message for `getdefaultlocale()`.

Elvish-Hunter:
> More generally for the rest, the python tools maintainer is Elvish_Hunter, and you do need to wait for them to review your PRs.

"Him", please :-)

> I have also added a way for people to launch the GUI in English (or alternative languages), rather than in their system language if so they wish.

That looks like a potential issue. GUI.pyw was designed to be launched with a simple double-click, and adding a command line option to select the language kind of defeats this point (granted, you can still create a symlink/launcher to the script and edit its command line, but it seems less than ideal). Probably a better solution would be adding a combobox menu into the GUI and add support for some sort of preferences file (which might also store the last used folder, for example) - but that'd be for me to code.

More in general, there's another issue: unlike the main game, the Python scripts aren't intended for general use, because they're aimed to WML maintainers whom are supposed to know English *because WML and its documentation are in English*.
Besides, all the other maintenance tools called by GUI.pyw still output their messages in English: do we want to translate them as well? It'd look strange to have their output in English even if you selected another language in the GUI.
Let's not forget that, until now, these tools have never been placed under feature or string freeze, because features and bugfixes may need to be backported to the stable series at any moment (and indeed they have been). Allowing translations may require enforcing freezes.
There's also the chance that a translation may alter a message to the point that it makes bug reports harder to make, especially if we end up translating the other tools as well.
In conclusion, I'm with stevecotton when he says that
> When a workflow requires a language to be used anyway, it's often easier to stay in one language than to switch between two, even when the other language is the native one.

Moreover, this looks like placing an unnecessary burden both on the translators (because every little fix may need translation updates) and on the coders (because now they have to take care of an additional workflow and possibly of freezes). And, as far as I know, the only translators who actively use the Python tools are Toranks and I, so I'm not sure how many users would actually benefit from this change.

BTW, from a technical point of view the PR looks fine (but I'd like if someone else could review the Scons part, since I'm not familiar with it); we just need to make sure that it's actually useful for someone and doesn't make their job harder for someone else.

> Honestly the closest we've had is the recent Translators group on Discord, as sad as that probably sounds.

@Pentarctagon I wonder if the Translator forum group can be tried for that purpose.

Pentarctagon:
The Translators group could be used, but at this point the large majority of the people in it are actually inactive.

CelticMinstrel:
> do we want to translate them as well?

I was thinking we might as well do that, now that GUI.pyw is translated.

> they're aimed to WML maintainers whom are supposed to know English

Obviously they need to know English for that reason, yes. But they might not be completely comfortable working in English.

loonycyborg (approved):

Pentarctagon:
So, coming back to this:
* Does anyone have any concerns about this from a purely technical perspective?
* In terms of @stevecotton's concern about whether there is an audience for this change, a couple people have said they would find it useful and nobody has said they don't want it. Obviously that's an absolutely abysmal sample size, but I'm not sure of a better way to determine this, so I would say that's good enough to say that there would be at least some people who would use this.
* As far as whether the current UI strings are in a good enough state that they should be made available for translation, I would say that an issue should be opened for it but that reviewing them isn't necessarily a requirement for this PR. Wesnoth's long-standing policy is that the dev branch shouldn't be the target of any translation team and translations they made against it they do at their own risk. Though of course if @Slayer95 wants to review them with this PR, of course feel free.

Therefore, if there are no concerns with point 1, and nobody disagrees with points 2 and 3, then I would say this can be merged.

data/tools/FINDPY
```diff
@@ -0,0 +1,27 @@
...
+echo "data/tools/wmlflip" ;
+echo "data/tools/wmlindent" ;
+echo "data/tools/wmllint" ;
+echo "data/tools/wmllint-1.4" ;
+echo "data/tools/wmlscope" ;
+echo "data/tools/wmlunits" ;
+echo "data/tools/wmlxgettext" ;
+} | sort |
+while read file; do
+	if grep -m 1 -E " *_ *= *gettext\\.translation\\((['\"])$1\\1" $file > /dev/null 2>&1;
```
stevecotton, line 23: Is there a need to do this, instead of just an unconditional `echo $file` (or just not filter at all)?

po/CMakeLists.txt
```diff
@@ -56,25 +57,43 @@ if(ENABLE_POT_UPDATE_TARGET)
...
-			DEPENDS ${PROJECT_SOURCE_DIR}/po/${DOMAIN}/POTFILES.in
+			DEPENDS ${PROJECT_SOURCE_DIR}/po/${DOMAIN}/POTFILES_CPP.in
 			WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
 			COMMENT "pot-update [${DOMAIN}]: Generated source pot file."
 		)
 
-		# Generate pot file for wml data.
+		# Generate pot file for Python strings.
+		add_custom_command(
+			# misses bug address
```
stevecotton, line 80: What does this comment mean?

po/CMakeLists.txt
```diff
@@ -56,25 +57,43 @@ if(ENABLE_POT_UPDATE_TARGET)
...
+			# create the pot file make sure we always get output.
+			COMMAND ${GETTEXT_XGETTEXT_EXECUTABLE} ${GETTEXT_XGETTEXT_OPTIONS}
+					--language=Python
+					--output=${PROJECT_SOURCE_DIR}/po/${DOMAIN}/${DOMAIN}.py.pot
+					${PROJECT_SOURCE_DIR}/data/tools/GUI.pyw
+			# replace the chartype
+			COMMAND sed -i
+				s/charset=CHARSET/charset=UTF-8/
+				${PROJECT_SOURCE_DIR}/po/${DOMAIN}/${DOMAIN}.py.pot
+			# Remove some header info - Need to test whether needed.
```
stevecotton, line 91: Seems to be a leftover comment that no longer has code with it

data/tools/FINDPY
Slayer95: Unless we prefer refactoring the build scripts _and_ special casing the ``wesnoth-tools`` domain, I think so.

I was actually thinking of drying the ``FINDCFG`` scripts (there is one of them, exactly the same, for each text domain), using a similar approach to this ``FINDPY`` file.

po/CMakeLists.txt
Slayer95: Oh, this is a leftover. No idea what it means, tbh.

data/tools/FINDPY
stevecotton: Ah, I thought it was filtering for files that contained `gettext.translation` to filter out files that don't contain any translatable text, rather than filtering for a specific textdomain.

Please add comments about what the behavior of this script is. It's different to the `FINDCFG` scripts (which don't filter for textdomains), and reading a regex isn't documentation.

Also, instead of `-m 1` and a conditional print, grep's `--files-with-matches` option will print just the filename for you.

po/CMakeLists.txt
```diff
@@ -56,25 +57,40 @@ if(ENABLE_POT_UPDATE_TARGET)
...
 
-		# Generate pot file for wml data.
+		# Generate pot file for Python strings.
+		add_custom_command(
+			OUTPUT ${PROJECT_SOURCE_DIR}/po/${DOMAIN}/${DOMAIN}.py.pot
+			# create the pot file make sure we always get output.
+			COMMAND ${GETTEXT_XGETTEXT_EXECUTABLE} ${GETTEXT_XGETTEXT_OPTIONS}
+					--language=Python
+					--output=${PROJECT_SOURCE_DIR}/po/${DOMAIN}/${DOMAIN}.py.pot
+					${PROJECT_SOURCE_DIR}/data/tools/GUI.pyw
```
stevecotton, line 84: Should this line be using FINDPY instead of a specific filename?

Slayer95: Yeahh...... I forgot about this... But I can't tell what's going on with the build :disappointed:

data/tools/FINDPY
Slayer95: Great suggestion. Applied.

Elvish-Hunter:
> Therefore, if there are no concerns with point 1, and nobody disagrees with points 2 and 3, then I would say this can be merged.

Sorry for the delay (real life got in the way once again). As long as this PR (and making the other Python tools translatable) does not result in any future string or feature freeze, it can go in.

Pentarctagon:
String freezes apply to all translatable strings, which would now include this tool.

Pentarctagon:
I think there might be a misunderstanding? A string freeze only means you can't change text that's marked as available for translation. It doesn't mean you can't make any changes to the tool at all.

Elvish-Hunter:
In that case, we'll have to evaluate very carefully what kind of issues can arise from placing those tools under string freeze before taking the decision of translating them. But, as I said, GUI.pyw doesn't change very often, so I can finally approve this PR.

Elvish-Hunter (approved):

po/CMakeLists.txt
```diff
@@ -56,40 +57,56 @@ if(ENABLE_POT_UPDATE_TARGET)
...
+			COMMENT "pot-update [${DOMAIN}]: Generating python pot file."
+		)
+		# Generate pot file for WML/Lua strings.
 		add_custom_command(
 			OUTPUT ${PROJECT_SOURCE_DIR}/po/${DOMAIN}/${DOMAIN}.wml.pot
 
-			COMMAND ${WMLXGETTEXT}
-				--directory=${PROJECT_SOURCE_DIR}
+			COMMAND ${Python_EXECUTABLE} ${WMLXGETTEXT}
+				--force-po
```
Slayer95, line 98: This line gets eaten in CI...

:question:

Slayer95:
So, along the way to get this working 100%, I rewrote several of the cmake commands to Python. ``update_pot_source_dependencies.cmake`` was simplified a lot, while the cmake algorithm remains the same overall.

The Python routines followed by cmake are different than the ones followed by scons. AFAICT, scons does it better in that it supports only one domain per file for those files analyzed with ``xgettext``, and that, well, matches ``xgettext`` default behavior. An as aside, note that, in order to use more than 1 domain in a single cpp/python file, disambiguation through different ``--keyword`` values would be needed, which means that translatable strings would be marked e.g. as ``GETTEXT_WESNOTH_LIB()``, ``_wesnoth_lib()``.

So the ``cmake`` algorithm would have to be changed as a follow-up in order to match the more compliant algorithm ``scons`` uses. But I am not comfortable at all with ``cmake`` (somehow I still managed to improve it), so hopefully that can be left for later.

That refactoring would be a chance to also refactor the usage of ``wmlxgettext``, which is now able to extract strings from all text domains in a single run (#7624).

Finally, regarding the domain and the amount of entries added: The ``wesnoth-tools`` domain overall adds 99 entries (yes, more were added since the last report I made on its size.) I thought that there might be some overlap with ``wesnoth-editor``, or even ``wesnoth-lib``.

As it turns, out, there are 2 entries shared with ``wesnoth-editor`` (which today has 109 entries), and 7 shared with ``wesnoth-lib`` (which today has 1518 entries.) As complementary info, ``wesnoth-editor`` has over 90% translation rate in 25 languages, while ``wesnoth-lib`` achieves that rate only in 16 languages. Overall, those numbers don't justify merging ``wesnoth-tools`` into one of those preexisting domains, imo.

data/tools/FINDPY
stevecotton: I still feel this the FINDPY file needs more comments, but the issue raised here is resolved, so I'll mark the conversation as resolved.

Pentarctagon:
Given wesnoth-tools could expand if other tools become translatable, I think having its own textdomain is fine.

po/SConscript
```diff
@@ -48,22 +72,43 @@ if "pot-update" in COMMAND_LINE_TARGETS:
...
+                )
+
+        # py
+        with open(join(domain, "POTFILES_PY.in"), "w") as potfiles:
+            potfiles.writelines([line + "\n" for line in sorted(domain_tools_py[domain[7:]])])
+        tools_py = [Dir("#").File(x) for x in domain_tools_py[domain[7:]]]
+        if tools_py:
+            tools_py_pot = env.Command(
+                join(domain, domain + ".cpp.pot"),
+                sources_cpp,
```
stevecotton, line 100: Should the last two lines be `domain + ".py.pot"` and `sources_py` ?

Slayer95: > Should the last two lines be domain + ".py.pot" and sources_py ?

Yes, thanks for tracking this down. That explains your findings that ``wesnoth`` is emptied.

> renaming tools_py as sources_py

This also makes sense.

stevecotton:
It adds strings to the `wesnoth` textdomain's POT file, which is bug #7704. Dithering between the options:

* Block until #7704 is fixed (or at least, the strings that shouldn't be separate entries in the POT file are fixed)
* Merge and start the process of improving the English text in GUI.pyw

Slayer95:
> It adds strings to the wesnoth textdomain's POT file, which is bug https://github.com/wesnoth/wesnoth/issues/7704.

Updated the script to be more strict. Text domain will no longer default to ``wesnoth`` in Python files.

stevecotton:
SCons builds are now looking good.

CMake builds are missing the .cfg and .lua files from `wesnoth`, `wesnoth-editor`, `wesnoth-help` and `wesnoth-lib`.

Slayer95:
> CMake builds are missing the .cfg and .lua files from `wesnoth`, `wesnoth-editor`, `wesnoth-help` and `wesnoth-lib`.

Hmmm. That's not [what I see](https://github.com/wesnoth/wesnoth/compare/master...Slayer95:wesnoth:wmltools-gui-i18n-test?expand=1).

po/CMakeLists.txt
```diff
@@ -56,40 +57,56 @@ if(ENABLE_POT_UPDATE_TARGET)
...
 				--domain=${DOMAIN}
+				--directory=${PROJECT_SOURCE_DIR}/data
```
stevecotton, line 100: This path includes `data`, however the paths outputted by the `FINDCFG` scripts also include `data`. After removing this line completely (so it uses the existing `WORKING_DIRECTORY` line), the data seems good for CMake.

> Hmmm. That's not [what I see](https://github.com/wesnoth/wesnoth/compare/master...Slayer95:wesnoth:wmltools-gui-i18n-test?expand=1).

The "cmake in Windows" commit in that branch changes the next couple of lines to use `--recursive` instead of FINDCFG.

stevecotton:
SCons builds are losing the .cpp files from po/wesnoth/wesnoth.pot again.

po/CMakeLists.txt
Slayer95: Oh... this diff was a leftover from that branch, actually.

stevecotton (approved):
Looks good to me, although I'll leave it open for comments for a couple of days.

About merging it's a bit complex whether to squash-merge or not.
* The first commit ("i18n: Make GUI.pyw translatable") makes sense as an individual commit
* The second commit's script changes would be good as an individual commit, but the addition of the `wesnoth-tools` domain to the build scripts is also in this commit, which breaks the build (it can't generate the .mo files without the .po files).
* Putting the addition of the new domain, the POT and the PO files in one commit makes sense to me, and would avoid a build-breakage for the .mo files.
* Would like to squash the "fixes" commit.

@Slayer95, please would you be so kind as to do a rearrange-into-perfect-history rebase?

Slayer95:
I initially split just the few lines that add ``wesnoth-tools`` itself into its own commit, but I figured I'd squash it into the commit that adds the ``po`` files. Let me know if you think it would be better otherwise.

Issue #6510, opened 2022-02-11
Northern Rebirth Prologue: Don't use "reckon" twice in succeeding sentences


eyalroz:
English text, second "screen" of NR campaign prologue sequence:

> The dwarves of Knalga, cannily **reckoning** the gains from their trade ... The people... **reckoned** themselves well able to fight...

The use of "reckon" in succeeding sentences doesn't sound right. I suggest replacing one of them with something else - at the word or phrase level. The second use could be replaced with "considered". The first use might benefit from some expansion or clarification. It's not exactly clear why the "canny reckoning of gains" in itself leads to building fortifications, other than for the consideration in the second sentence.

Descacharrado:
Permalink to the code (master): https://github.com/wesnoth/wesnoth/blob/933c273b997b61ebe4cf435c4e535154ba975abd/data/campaigns/Northern_Rebirth/scenarios/01_Breaking_the_Chains.cfg#L23

Wedge009:
For a small change like this, I suggest making a PR. To me, the words are far enough apart that it doesn't sound too bad but if it's that important for you, you can submit your suggested change.

cooljeanius:
I think the repetition might actually be an intentional form of wordplay, since each usage of "reckon" is in a different sense: the first one is more in an archaic sense of the word meaning "realizing" (in the sense of "making real") or "capitalizing upon" or "taking advantage of", while the second one is in the more modern/colloquial sense of "thinking"

eyalroz:
@cooljeanius : That's a problem, because (1.) narration should not be jovial/humorous (2.) We shouldn't have sentences mixing modern and archaic language like that.

cooljeanius:
> @cooljeanius : That's a problem, because (1.) narration should not be jovial/humorous (2.) We shouldn't have sentences mixing modern and archaic language like that.

OK well I added a commit (b3a3d4b) to PR #6057 that changes the second "reckoned" to "considered" at least

cooljeanius:
So, #7454 recently updated some of the dialogue in this area, but after rechecking, it looks like it didn't address this particular nit, though...

TheShadowOfHassen:
Sorry I didn't see this sooner I've been busy and github didn't notify me of references to my pull request.

I'm the person who did the rewrite of the first NR script. I want to do the rest of the campaign but I haven't had time.

So I left both reckoning to have parallel construction. I honestly think it's better with it, but if a lot of people don't like it...

[...]

PR #7628, opened 2023-05-13
Nr: Dialog/ Story Rewrite : Scenario One (Fix) and herodeaths

TheShadowOfHassen:
So this is another part of the northern rebirth editing /rewrite I'm doing. The last one was #7454. This one isn't the next scenario but I should be able to get that soon. This one is fixing a couple editing problems I didn't catch in scenario one and I real quick checked the herodeaths.

The main thing is in scenario one I rewrote a paragraph in order to fix issue #6510. I also changed a sentence that seemed to really pop out to be confusing. The bit about raiding parties becoming companies becoming war-bands. Because while you can have a company of orc soldiers, it's different from a war band.

I also added an oxford comma. Are we using oxford comas in wesnoth? If not I'll change it back. 

The hero death changes are pretty self explanatory.

### DIFF
--- a/data/campaigns/Northern_Rebirth/scenarios/01_Breaking_the_Chains.cfg
+++ b/data/campaigns/Northern_Rebirth/scenarios/01_Breaking_the_Chains.cfg
@@ -20,7 +20,7 @@
             {PROLOGUE_STAGE1}
         [/part]
         [part]
-            story= _  "The dwarves of Knalga, cannily reckoning the gains from trade, built strong fortifications around Dwarven Doors. The people who called it home, dwarven and human alike, knew that the wealth of their city might arouse envy from afar — hence they kept their weapons sharp, and reckoned themselves well able to fight off any bandit gang or petty warlord that could arise in the thinly-settled Northlands."
+            story= _ "The dwarves of Knalga, cannily reckoning the gains from trade, built strong fortifications around Dwarven Doors. The dwarves and humans who called it home knew that the wealth of the city might arouse envy from afar --hence they kept their weapons sharp, and thought themselves able to fight off any bandit game or petty warlord."
 
             {NR_BIGMAP}
             {PROLOGUE_STAGE1_END}
@@ -31,12 +31,12 @@
             {PROLOGUE_STAGE1_END}
         [/part]
         [part]
-            story= _ "At first, the incursions were mere rabble that were easily felled. Complacent in their own prowess, the people of the Dwarven Doors ignored the orcs even as the raids grew more numerous and threatening. In time, the raiding parties became companies and full war-bands. Then at last, there arose a mighty warlord of the Bloody Sword tribe, Khazg Black-Tusk, who surrounded the entirety of their defenses with a great host of warriors."
+            story= _ "At first, the incursions were mere rabble that were easily felled. Complacent in their own prowess, the people of the Dwarven Doors ignored the orcs even as the raids grew more numerous and threatening. However in time, the raiding parties became larger until they were more akin to war bands. At last, there arose a mighty warlord of the Bloody Sword tribe, Khazg Black-Tusk, who surrounded the entirety of their defenses with a great host of warriors."
             {NR_BIGMAP}
             {PROLOGUE_STAGE1_END}
         [/part]
         [part]
-            story= _ "It was a time of doom and great deeds, of fire and blood and slaughter. The men and dwarves fought valiantly. But there were always, it seemed, more orcs to replace the slain, while the defenders could find no succor. Pleas for help went unheeded in other lands for men, elves and even dwarves, were wrapped up in their own affairs and struggles or deemed the orcs too far away to bring unto them harm."
+            story= _ "It was a time of doom and great deeds, of fire and blood and slaughter. The men and dwarves fought valiantly. But there were always, it seemed, more orcs to replace the slain, while the defenders could find no succor. Pleas for help went unheeded in other lands for men, elves, and even dwarves, were wrapped up in their own affairs and struggles or deemed the orcs too far away to bring unto them harm."
             {NR_BIGMAP}
             {PROLOGUE_STAGE1_END}
         [/part]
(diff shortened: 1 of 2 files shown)


eyalroz:
* "thought themselves" - perhaps "considered themselves"; it would be closer to "reckoned" without repeating the word.
* "bandit game" - are you sure that's what you meant to write? Do you mean "game" in the sense of hunting, like "game animals"?
* "became larger until they were more akin to war bands" - perhaps "grew larger"?
* "However in time," - perhaps "With time, however" or "In time, however" or "Years passing, however"?

nemaara:
I don't generally think this is an improvement for me. We do use Oxford commas so that one can stay but that's a very simple fix. If you want to make another attempt you can or go ahead and attempt S2 instead.

eyalroz:
> I don't generally think this is an improvement for me.

Can you elaborate?

nemaara:
It doesn't add any content to the sentences and I don't think the original style is improved by these changes. In general, I discourage endlessly changing style for the sake of nitpicky wording. Generally what we need are changes that add more content to the story.

eyalroz:
This is a bug fix (actually, two bug fixes and some tweaking). Text needs bug fixes just like code. Not all PRs are new features/content... it's not like this dialog gets rewritten every couple of years.

TheShadowOfHassen:
@eyalroz thanks for the read through. Those are valid suggestions that I fixed and will make it even better. I wouldn't mind at all if you looked over each pull request for this rewrite and have your opinion.

So I'm doing this rewrite/edit as if I was editing one of my stories.  This pull request was more nit-picky than before, however I agree with @eyalroz they are basically bug fixes.  If I was programming wesnoth I would try to make sure the code was easy to understand and efficient. It isn't as cut and dry in writing as it is programming. In programming if you have an error it will pop up or if you have a bug something won't work in the code. In editing all you have to know it is wrong is your head and what others say. However there is problems in writing, they are just harder to find.

@nemaara I understand your concern of endlessly nitpicking and that won't happen. (Mostly because I don't like doing it) I did this pull request mostly to fix #6510. The other bit in the first scenario I added because I saw i, and, in my opinion, the changes took the writing from a 89 to a 92 (Arbitrary numbers, I hope you understand what I mean.) I looked at the hero deaths because I thought I could get them done real quick so I wouldn't need to make a different pull request with them, and save everyone time. The changes really didn't add anything I know but I thought it made it a bit better.

The one point is I never set out to add new content to the scenario. When I reword things it might sound like I am adding more because I am taking what I thought of the story and expressing it clearly (at least, I hope) No offense to the people who wrote it, but while I understood the campaign when I played it, reading wasn't the easiest.

I understand your concern of nitpicking and it won't happen. That said, I'd appreciate it (if there isn't anything wrong with it) if you'd review this so it can be merged. Next scenario will be number 2.

data/campaigns/Northern_Rebirth/scenarios/01_Breaking_the_Chains.cfg
```diff
@@ -20,7 +20,7 @@
             {PROLOGUE_STAGE1}
         [/part]
         [part]
-            story= _  "The dwarves of Knalga, cannily reckoning the gains from trade, built strong fortifications around Dwarven Doors. The people who called it home, dwarven and human alike, knew that the wealth of their city might arouse envy from afar — hence they kept their weapons sharp, and reckoned themselves well able to fight off any bandit gang or petty warlord that could arise in the thinly-settled Northlands."
+            story= _ "The dwarves of Knalga, cannily reckoning the gains from trade, built strong fortifications around Dwarven Doors. The dwarves and humans who called it home knew that the wealth of the city might arouse envy from afar --hence they kept their weapons sharp, and considered themselves able to fight off any bandit gang or petty warlord."
```
nemaara, line 23: The em dash is intentionally used here instead of the double dash, please keep the original. Please keep "reckoned" but you can drop the well.

data/campaigns/Northern_Rebirth/scenarios/01_Breaking_the_Chains.cfg
```diff
@@ -31,12 +31,12 @@
             {PROLOGUE_STAGE1_END}
         [/part]
         [part]
-            story= _ "At first, the incursions were mere rabble that were easily felled. Complacent in their own prowess, the people of the Dwarven Doors ignored the orcs even as the raids grew more numerous and threatening. In time, the raiding parties became companies and full war-bands. Then at last, there arose a mighty warlord of the Bloody Sword tribe, Khazg Black-Tusk, who surrounded the entirety of their defenses with a great host of warriors."
+            story= _ "At first, the incursions were mere rabble that were easily felled. Complacent in their own prowess, the people of the Dwarven Doors ignored the orcs even as the raids grew more numerous and threatening. With time, however, the raiding parties grew larger until they were more akin to war bands. At last, there arose a mighty warlord of the Bloody Sword tribe, Khazg Black-Tusk, who surrounded the entirety of their defenses with a great host of warriors."
```
nemaara, line 34:
```suggestion
            story= _ "At first, the incursions were mere rabble that were easily felled. Complacent in their own prowess, the people of the Dwarven Doors ignored the orcs even as the raids grew more numerous and threatening. With time, the raiding parties grew into war bands; then at last, there arose a mighty warlord of the Bloody Sword tribe, Khazg Black-Tusk, who surrounded the entirety of their defenses with a great host of warriors."
```

nemaara (changes requested):

nemaara:
Again I want to stress this because it's important, but story text is not code. I don't want us to treat editing like bug fixing because writing has many reasons behind style beyond denotative meaning. This is something I've noticed before where a lot of grammar and word choice nitpicking ends up making dialogue and narration feel worse after the editing even though it might be more "technically correct". Clarity is good but please keep in mind that the original author's style is also important.

nemaara:
Small nit again, there needs to be a space after the em dash. Can merge after.

TheShadowOfHassen:
I think I fixed it but it still says there's a change requested... I'm not very good with github.

@nemaara I completely agree that editing shouldn't suck out the original author's writing style, and I don't want that to happen.

I'll fix the one last thing real quick

nemaara (approved):

PR #7629, opened 2023-05-14
1.17 new terrain - ruined walls

doofus-01:
There are at least three types of ruined walls I can think of:

- [x] 1.  Damaged stone walls (Xor)
- [x] 2.  Damaged cave walls (Xur)
- [x] 3.  Ruined walls, was originally an overlay, but is now alias of cave or flat (whichever is worse) - (Exos)

Then there are all the transitions:
- [x] 4.  Transitions (mostly done)

For now, just item 1 is done.
EDIT: some progress on item 3, though there is still some work to do:
Screenshot_20230514_


doofus-01:
I think all the important transitions are completed.
Screenshot_20230521_
The wooden wall to ruins transition probably needs its own image set, but it seems like an odd grouping, so I'm considering it low priority.

doofus-01:
Thinking about the use cases, it's probably more important to have the overlay terrain slot available than to have choice of base terrain.  So now it's three separate terrains Xor, Xur, and Exos.  The base graphics for Exos are the dark dirt.
Screenshot_20230521_b

PR #7638, opened 2023-05-19
Language selection UI overhaul

irydacea:
(Please ignore the first commit including a separate bugfix.)

This overhaul introduces a GUI option to show all languages regardless of their completion threshold, as well as a side panel including a general explanation of community translations and how players can potentially contribute to them, plus a link to the main translations page on the wiki.

Screenshot 2023-05-19 at 01 23 16
Screenshot 2023-05-19 at 01 23 38

A few points to make here:

* The URL points to the main page on purpose, since it's meant to be a catch-all for all languages regardless of which one the player has got selected or configured (those are two different things at the moment).

* There's an issue where when showing all translations, some of the non-Latin translation names cause the baseline of the language name labels to end up entirely misaligned with the completion labels. I am not entirely sure how to fix this right now.

* Showing all translations alters the dialog's layout permanently. I believe there's no way out of this with the current GUI API.

* We use the stock text colouring function used for unit defence values. This might not be entirely ideal because they are all very green for translation stats above the minimum threshold, so there's basically no variance by default until all translations are shown together.

* An idea could be to add a link to the gettext.wesnoth.org stats page in the dialog.

Additionally, I'd like to look into potentially reflecting the selection in the dialog or the entire Wesnoth UI without having to press OK, but I'm not entirely sure how feasible that is.

### DIFF
--- a/src/gui/dialogs/language_selection.cpp
+++ b/src/gui/dialogs/language_selection.cpp
@@ -40,12 +40,10 @@ REGISTER_DIALOG(language_selection)
 
 language_selection::language_selection()
 	: modal_dialog(window_id())
-	, show_all_(get_min_translation_percent() == 0)
 	, langs_(get_languages(true))
 	, complete_langs_(langs_.size())
 {
-	// Markup needs to be enabled for the link to be highlighted
-	register_label("contrib_url", true, translations_wiki_url, true);
+	bool show_all = get_min_translation_percent() == 0;
 
 	const language_def& current_language = get_language();
 
@@ -57,12 +55,16 @@ language_selection::language_selection()
 			// too, regardless of what the initial threshold setting is.
 			complete_langs_[i] = true;
 			if(langs_[i].percent < get_min_translation_percent()) {
-				show_all_ = true;
+				show_all = true;
 			}
 		} else {
 			complete_langs_[i] = langs_[i].percent >= get_min_translation_percent();
 		}
 	}
+
+	register_bool("show_all", true, show_all);
+	// Markup needs to be enabled for the link to be highlighted
+	register_label("contrib_url", true, translations_wiki_url, true);
 }
 
 void language_selection::shown_filter_callback()
@@ -70,11 +72,9 @@ void language_selection::shown_filter_callback()
 	window& window = *get_window();
 
 	toggle_button& show_all_toggle = find_widget<toggle_button>(&window, "show_all", false);
-	show_all_ = show_all_toggle.get_value_bool();
-
 	listbox& list = find_widget<listbox>(&window, "language_list", false);
 
-	if(show_all_) {
+	if(show_all_toggle.get_value_bool()) {
 		list.set_row_shown(boost::dynamic_bitset<>{langs_.size(), ~0UL});
 	} else {
 		list.set_row_shown(complete_langs_);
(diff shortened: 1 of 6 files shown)


src/gui/dialogs/language_selection.cpp
```diff
@@ -20,38 +20,102 @@
...
+		list.set_row_shown(boost::dynamic_bitset<>{langs_.size(), ~0UL});
+	} else {
+		list.set_row_shown(complete_langs_);
+	}
+}
+
 void language_selection::pre_show(window& window)
 {
+	styled_widget& contrib_url = find_widget<styled_widget>(&window, "contrib_url", false);
+	contrib_url.set_label(translations_wiki_url);
```
Vultraz, line 89: A case where `register_label` in the ctor is enough.

src/gui/dialogs/language_selection.cpp
```diff
@@ -20,38 +20,102 @@
...
 		data["language"]["label"] = lang.language;
+		data["translated_total"]["label"] = "<span color='" + game_config::red_to_green(lang.percent).to_hex_string() + "'>" + std::to_string(lang.percent) + "%</span>";
+		data["translated_total"]["use_markup"] = "true";
 
 		list.add_row(data);
+
 		if(lang == current_language) {
 			list.select_last_row();
-		}
+		};
```
Vultraz, line 113: ?

src/gui/dialogs/language_selection.cpp
```diff
@@ -62,9 +126,15 @@ void language_selection::post_show(window& window)
 
 		assert(res != -1);
 
-		const std::vector<language_def>& languages = get_languages();
-		::set_language(languages[res]);
-		preferences::set_language(languages[res].localename);
+		if(langs_[res] == get_language()) {
```
Vultraz, line 129: I'd invert this condition and put the actions in the block.

irydacea: Oops. This wasn't even meant to be in the PR yet. Noted though.

CelticMinstrel:
> Additionally, I'd like to look into potentially reflecting the selection in the dialog or the entire Wesnoth UI without having to press OK, but I'm not entirely sure how feasible that is.

I think reflecting the selection in the dialog is quite feasible. I'm not sure about reflecting the selection in the title screen _behind_ the dialog though.

irydacea:
I figured out most of the necessary code to refresh all language features — including loading the correct fonts — as soon as an option is clicked in a separate version of the same PR branch that includes a separate patch for it: https://github.com/irydacea/wesnoth/commit/6fb3346fbe45a893d6fd6f68133035d9e5c0cbc5

While that even fixes the issue with the titlescreen logo requiring a manual refresh (but not the Version and Language menu labels) it also highlights how horribly inefficient the whole process is, at least on a Debug build. The addition of the image cache purge increases the time Wesnoth spends unresponsive when selecting a language from about 0.33 seconds to about 3 seconds. So I'm not really sure if it's worth incorporating that patch into this PR.

irydacea:
Was there anything else that needed to be done here? I believe vultraz said he wasn't interested in the version that switches the translation as soon as the selection changes.

Pentarctagon:
Not related to the PR, but I should probably try contacting the people listed on the WesnothTranslation page to see how many are outdated.

Vultraz (approved):

Pentarctagon:
Given vultraz has since approved, it looks like this can go in?

irydacea:
I can rebase it. I wasn't sure if we're allowed to merge larger features into dev yet though.

Pentarctagon:
1.18 has been pretty stable, so I think it's fine.

PR #7640, opened 2023-05-20
gui2/title_screen: Fix Version label translation not updating

irydacea:
This unifies both the Version and the Language menu labels into a single method for both that is called both during construction and after switching languages. Both have this issue where they aren't `t_string` instances, so changing languages has no effect on them for that reason and they need to be manually re-translated.


Wedge009:
Oh, I thought this was already merged. Curiously, at a124571b2abbfc1bfb7427298d3649b84e36ba7e, language label is changing properly for me.

PR #7641, opened 2023-05-20
Add libcurl to library version reports in Game Version dialog

irydacea:


src/build_info.cpp
```diff
@@ -49,6 +49,8 @@
 #include <openssl/opensslv.h>
 #endif
 
+#include <curl/curl.h>
```
stevecotton, line 52: I'd prefer it in alphabetical order between `boost` and `openssl`, but that's a minor request.

irydacea: The library list has never been in alphabetical order, though: Boost, Lua, OpenSSL/libcrypto, libcurl, Cairo, Pango, SDL, SDL_image, SDL_mixer, libpng (when it was listed). The order has always meant to be more "topical" than anything.

PR #7648, opened 2023-05-21
Refresh po4a workflow

loonycyborg:
currently this fixes error with newest po4a but still po4a complains about po4a-gettextize/po4a-translate being deprecated. Could also go unified po4a approach but it seems less buildsystem-friendly.

Issue #7149, opened 2022-11-20
pot-update fails on Debian Unstable "po4a-gettextize ... provide the same amount of master files and localized files"


stevecotton:
### Game and System Information

- **Version**: 1.16.6, 1.17.10. The bug is a change in toolchain behavior, so previously-working versions have broken.
- **OS:** Debian Unstable, with po4a version 0.68-1
- **Bug logged with the OS vendor:** https://bugs.debian.org/1020821

### Description of the bug

Trying to build the translations (`scons pot-update`) on an up-to-date Debian Linux install fails with:
```
/usr/bin/po4a-gettextize -f man -m doc/man/wesnoth.6 -m doc/man/wesnothd.6 -p po/wesnoth-manpages/wesnoth-manpages.pot
You must provide the same amount of master files and localized files to synchronize them, as po4a-gettextize is intended to synchronize master files and previously existing translations. If just want to extract POT files of your master files, please use po4a-updatepo. Please note that the most convenient way of using po4a is to write a po4a.conf file and use the integrated po4a(1) program.
```

This is a toolchain change, which has also affected the debian-reference and maint-guide packages. The decision there seems to be that we're using po4a in an unsupported way, but that it's easy to change to using a supported way instead.

loonycyborg:
I'm getting that too now. Will look into fixing this. Hopefully our workflow isn't incompatible with new po4a way.

stevecotton:
Fixed by the combination of #7648 and #7649.

PR #7649, opened 2023-05-21
Support po4a version 0.68-1 (cmake part)

stevecotton:
This and @loonycyborg's #7648 are both working on #7149. Ended up with this being the CMake one, and the other being the SCons one.

The po4a tool's author considers running po4a-gettextize to be a once-and-done operation, however it seems to be moving towards acting as its own build system, expecting to be run with a config file telling it which files to build.

See https://bugs.debian.org/1020821

### DIFF
--- a/cmake/FindTranslationTools.cmake
+++ b/cmake/FindTranslationTools.cmake
@@ -17,7 +17,6 @@ This module defines the following variables:
 	DOS2UNIX_EXECUTABLE
 	PO4A-TRANSLATE_EXECUTABLE
 	PO4A-UPDATEPO_EXECUTABLE
-	PO4A-GETTEXTIZE_EXECUTABLE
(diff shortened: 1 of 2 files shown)


loonycyborg:
I think rename is a better way, po4a-gettextize only exists to adopt existing translations already done without po4a, it's not too useful to have a builder for it.

loonycyborg:
but switching to single invocation of po4a with appropriate config file would obsolete all this.

stevecotton:
I'm not sure if the "invoke po4a with an appropriate config" will do what we expect. Not tested yet, but the manpage for it suggests that it chooses not to generate documents at all instead of generating documents with less than a certain threshold translated, so I'm not sure it fits into a build system that expects documents to be generated.

PR #7651, opened 2023-05-22
1.17 Updated Elf Fighter sprite

doofus-01:
from here https://forums.wesnoth.org/viewtopic.php?p=681569

Still need to review a bit more, but I wanted to make sure it wasn't forgotten again.


Pentarctagon:
It looks like this also has more animations than the current sprite, correct?

doofus-01:
It has a running animation, in exchange for the idle animation.  
What needs review is:

- the running is SE-only, though it might still be OK for northern movement
- the other animations might be a little too faithful to the existing
- how to fit this in with #6417 and the advancements

Probably best to commit this and then follow up, but no need to decide today.

doofus-01:
Forgot what I was doing, let's save that for later

PR #7652, opened 2023-05-23
desktop: Fix enumerate_storage_devices() inaccuracy on Linux

irydacea:
Current Linux distributions tend to use udisks2 for storage device management, which uses `/run/media/USERNAME` instead of `/media` for gathering mount points. This commit makes this the first candidate for enumeration of mount point collections.

### DIFF
--- a/src/desktop/paths.cpp
+++ b/src/desktop/paths.cpp
@@ -106,7 +106,15 @@ void enumerate_storage_devices(std::vector<path_info>& res)
 	// reasoning here is that if any or all of them are non-empty, they are
 	// probably used for _something_ that might be of interest to the user (if not
 	// directly and actively controlled by the user themselves).
-	static const std::vector<std::string> candidates { "/media", "/mnt" };
+	// Note that the first candidate is actually /run/media/USERNAME -- we need
+	// to fetch the username at runtime from passwd to complete the path.
+	std::vector<std::string> candidates { "/media", "/mnt" };
+
+	// Fetch passwd entry for the effective user the current process runs as
+	if(const passwd* pw = getpwuid(geteuid()); pw && pw->pw_name && pw->pw_name[0]) {
+		candidates.emplace(candidates.begin(), "/run/media/");
+		candidates.front() += pw->pw_name;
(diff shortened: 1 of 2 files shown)


src/desktop/paths.cpp
```diff
@@ -106,7 +106,18 @@ void enumerate_storage_devices(std::vector<path_info>& res)
...
+	// to fetch the username at runtime from passwd to complete the path.
+	std::vector<std::string> candidates { "/run/media/", "/media", "/mnt" };
+
+	// Fetch passwd entry for the effective user the current process runs as
+	if(const passwd* pw = getpwuid(geteuid()); pw) {
+		candidates.front() += pw->pw_name;
+	} else {
+		// We probably don't want to list /run/media. If the user wants to use
+		// it anyway though, let them.
+		candidates.erase(candidates.begin());
```
stevecotton, line 119: I think it the code would be clearer if it did `candidates.insert` in the success case, or if it just started with an empty vector and `push_back`'d the strings.

stevecotton (approved):

src/desktop/paths.cpp
```diff
@@ -106,7 +106,18 @@ void enumerate_storage_devices(std::vector<path_info>& res)
 	// reasoning here is that if any or all of them are non-empty, they are
 	// probably used for _something_ that might be of interest to the user (if not
 	// directly and actively controlled by the user themselves).
-	static const std::vector<std::string> candidates { "/media", "/mnt" };
+	// Note that the first candidate is actually /run/media/USERNAME -- we need
+	// to fetch the username at runtime from passwd to complete the path.
+	std::vector<std::string> candidates { "/run/media/", "/media", "/mnt" };
+
+	// Fetch passwd entry for the effective user the current process runs as
+	if(const passwd* pw = getpwuid(geteuid()); pw) {
```
CelticMinstrel, line 114: If you're going to go to the trouble of using the semicolon here, I'd think I'd prefer to write out a proper expression (`pw != nullptr`). Or just drop the `; pw` altogether.

PR #7662, opened 2023-05-25
Update wmllint calls to parse_macroref

Slayer95:
Fixup for #7561.

### DIFF
--- a/data/tools/wmllint
+++ b/data/tools/wmllint
@@ -2394,38 +2394,39 @@ to be called on their own".format(filename, num))
                 # this here rather than hack_syntax so the character can be
                 # recognized.
                 if macname == 'LOYAL_UNIT':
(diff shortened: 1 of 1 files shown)


data/tools/wmllint
```diff
@@ -2503,7 +2504,7 @@ to be called on their own".format(filename, num))
                 if capitalization_error.search(lines[i]):
                     print('"%s", line %d: probable capitalization or punctuation error' \
                           % (filename, i+1))
-                if key == "message" and in_message and not in_option and not ("wmllint: ignore" in comment or "wmllint: noconvert" in comment):
+                if False:
```
stevecotton, line 2507: Is this something that should be here, or a left-over from debugging?

Slayer95: Oops

PR #7669, opened 2023-05-30
1.17 - Make map elevation available to [scenarios]

doofus-01:
This is to create a tag that determines the elevation data on a map that uses the feature correctly.  The five-level limit is pretty baked in, because that's how it is in the terrain graphics.  The use for this would be anything that could use a location set (group of `x,y` ) as a filter.
```
[elevation_score_map]
    high_border=<terrain code> (*^Qhh* default, key is optional)
    low_border=<terrain code> (*^Qhu* default, key is optional)
    test_labels = <boolean> ( default false )
[/elevation_score_map]
```
The effect of this tag is to create some WML variables `[elevation_area]`, one per elevation marker (the arrows in the editor) + the unassigned default elevation
It also creates a WML variable `[elevation_data]` that is basically a summary, in case that helps somehow.  Currently it is just a count of the different elevation levels.

This can be tested in the `Test_Elevation_Test.cfg` MP scenario, which has a simple example use-case, where higher-ground attacker (horseman) gets a damage boost and lower-ground attacker (wolf rider) gets a damage nerf.
Screenshot_20230611_

---- original post ------
I'm really not the right person to be doing this, but we've got to start somewhere.  This is a first draft, I need to reconcile it with the terrain-graphics rules better and then work on improving the efficiency (the failsafe cycle works, but it causes Wesnoth to choke, need to get the main flood-fill to work better).  It reminds me of Tron, I'm probably reinventing the wheel.

For this PR, I hope to get the elevation data available to the scenario author (for now it is a label, which is obviously just debug).  Goal is for each hex to have elevation data (low-low, low, default, high, high-high).  How this data should be used is out of scope.

This is a hack, just like the elevation terrain-graphics.  Best action would be to replace both of them with an engine update, but 1.18 isn't that far away, so this may be the best realistic option.
Screenshot_20230529_

### DIFF
--- a/data/lua/wml/elevation_scoring.lua
+++ b/data/lua/wml/elevation_scoring.lua
@@ -1,4 +1,3 @@
-local utils = wesnoth.require "wml-utils"
 local location_groups = {}
 
 --[[ usage
--- a/data/lua/wml/elevation_scoring.lua
+++ b/data/lua/wml/elevation_scoring.lua
@@ -152,11 +152,8 @@ wesnoth.wml_actions.elevation_score_map = function(cfg)
                                         table.insert(location_groups[j].hexes, {x= new_hexes[k][1], y= new_hexes[k][2]})
                                         table.remove(border_locations, m)                                        
                                    end
-                                else
-                                    -- Nothing to do for high-high/low-low
                                 end
                                 al_found = 'no'
-                        else
                         end
                     end
                     if #candidate_locations < 1 then 
@@ -170,9 +167,7 @@ wesnoth.wml_actions.elevation_score_map = function(cfg)
                         local on_border = true
                         while on_border do
                             local random_index = mathx.random(#candidate_locations)
-                            if wesnoth.current.map:on_border(anchor_hex.x, anchor_hex.y) then
-                                    --table.remove(candidate_locations, random_index)
-                            else
+                            if not (wesnoth.current.map:on_border(anchor_hex.x, anchor_hex.y)) then
                                     on_border = false
                                     anchor_hex.x, anchor_hex.y = candidate_locations[random_index].x, candidate_locations[random_index].y
                                     table.remove(candidate_locations, random_index)
--- a/data/lua/wml/elevation_scoring.lua
+++ b/data/lua/wml/elevation_scoring.lua
@@ -13,10 +13,10 @@ local location_groups = {}
 
 --[[ There are two cycles:
  - First Cycle that grows out from the marker terrains, placing each hex in an elevation group, stopping when it gets to a border or runs out of available hexes.
-   Most likely it gets itself stuck in a corner, rather than really running out of hexes to assign, so restart from some random candidate (non-repeating) hex 
+   Most likely it gets itself stuck in a corner, rather than really running out of hexes to assign, so restart from some random candidate (non-repeating) hex
    somewhere in that blob.
 
- - Second Cycle iterates over all remaining available hexes and checks if they are a border terrain type, and if next to a high-high/low-low border or not, 
+ - Second Cycle iterates over all remaining available hexes and checks if they are a border terrain type, and if next to a high-high/low-low border or not,
    to assign to one of the four non-default elevations.  If not a border terrain type, they stay default
  ]]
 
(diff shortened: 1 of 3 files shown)


CelticMinstrel:
This looks pretty complicated, and I'm not sure I understand the purpose, so I think it would be nice to see an actual use-case for it.

doofus-01:
https://forums.wesnoth.org/viewtopic.php?t=56456 has some discussion of use cases.

CelticMinstrel:
I meant something a little more concrete.

doofus-01:
The first cycle works much better now, it's almost usable.  I might be able to delete the second cycle.  The final cleanup cycle is still TBD.
Screenshot_20230604_

doofus-01:
It's working now, and has a possible use case.  The "implementation" (lua tables -> WML events/data/whatever) was a pretty small part of this, writing elevation data out to something useful in scenario WML is all I aimed to do.  So, pretty much done, aside from the whitespace etc.

Is there a way to see what exactly failed with CI?  I don't have luacheck.

CelticMinstrel:
> has a possible use case

Where?

Pentarctagon:
The CI shows the errors as well:
```
    data/lua/wml/elevation_scoring.lua:155:33: (W542) empty if branch
    data/lua/wml/elevation_scoring.lua:159:25: (W542) empty if branch
    data/lua/wml/elevation_scoring.lua:173:90: (W542) empty if branch
```

doofus-01:
> The CI shows the errors as well:

Thanks, I'd thought there was more to it than that, but that was with the `fix_whitespace.sh` script

doofus-01:
> Where?

I don't know why you've been trolling this, but I've tried to make the OP a summary of the situation.

CelticMinstrel:
> I don't know why you've been trolling this

That's really rude.

I think storing the elevation in the unit variables isn't a great idea. It means it will be out-of-date more often than not. I think it would be better to use `[time_area]` to define the elevation areas.

Pentarctagon:
If I'm understanding correctly, the use case is to make it easier to tell what a unit's elevation is relative to other hexes on the map, correct?

CelticMinstrel:
Right, but elevation is logically a property of a tile, not a unit. So, putting it on the unit means it's only valid if the unit hasn't moved, which means the elevation mapping (or at least the part of it on the units) would need to be recalculated every time a unit moves or is spawned.

Using a `[time_area]` better preserves the fact that it's a property of the tile. The filter would change from

```
[filter_condition]
    [variable]
        name=unit.variables.elevation
        greater_than=$second_unit.variables.elevation
    [/variable]
[/filter_condition]
```

to something similar to

```
[filter]
    [filter_location]
        area=high
    [/filter_location]
[/filter]
[filter_second]
    [filter_location]
        area=low
    [/filter_location]
[/filter_second]
```

Generalizing that two-step filter to three or more steps isn't trivial, admittedly, and off the top of my head I'm not quite sure what it would look like. I think using two separate filters like above doesn't work. Perhaps there could be a custom conditional_wml filter.

Pentarctagon:
Making a tile's height part of a [time_area] is also kind of strange though, since it has nothing to do with times. Is there a reason to not make it a property of the tile itself, so it would work in a regular location filter?

CelticMinstrel:
> Making a tile's height part of a [time_area] is also kind of strange though, since it has nothing to do with times.

Time areas have nothing to do with times except insofar as you can put a custom schedule inside them. They are a general mechanism of grouping tiles into a defined area. The reason they're called `[time_area]` is basically backwards compatibility.

> Is there a reason to not make it a property of the tile itself, so it would work in a regular location filter?

This certainly makes sense, but it doesn't work too well for something that's being built on top of the engine, rather than as a part of the engine. We don't have a way to add a custom filter to a location filter from Lua. Adding an elevation check to the hex metatable does make sense, I think, but that doesn't help for filtering. Time areas do, though are a bit awkward for this use-case.

doofus-01:
I give up, I'll just release this as an add-on.

CelticMinstrel:
That seems better, since it's clearly in an unfinished state.

Pentarctagon:
@CelticMinstrel it's not incomplete so much as this isn't intended as the final state in the first place. He even said in the opening post:
> This is a hack, just like the elevation terrain-graphics. Best action would be to replace both of them with an engine update, but 1.18 isn't that far away, so this may be the best realistic option.

and

> For this PR, I hope to get the elevation data available to the scenario author (for now it is a label, which is obviously just debug). Goal is for each hex to have elevation data (low-low, low, default, high, high-high). How this data should be used is out of scope.
> 
> This is a hack, just like the elevation terrain-graphics. Best action would be to replace both of them with an engine update, but 1.18 isn't that far away, so this may be the best realistic option.

PR #7672, opened 2023-05-30
Write log output both to stderr and logfile at the same time

loonycyborg:
this uses tee sink from boost.iostreams

### DIFF
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -181,7 +181,7 @@ void set_log_to_file()
 	if(is_log_dir_writable_.value_or(false)) {
 		// get the log file stream and assign cerr+cout to it
 		output_file_path_ = filesystem::get_logs_dir()+"/"+unique_log_filename();
-		static std::unique_ptr<std::ostream> logfile { std::move(filesystem::ostream_file(output_file_path_)) };
+		static std::unique_ptr<std::ostream> logfile { filesystem::ostream_file(output_file_path_) };
 		static std::ostream cerr_stream{std::cerr.rdbuf()};
 		//static std::ostream cout_stream{std::cout.rdbuf()};
 		auto cerr_tee { boost::iostreams::tee(*logfile, cerr_stream) };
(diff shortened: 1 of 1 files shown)


loonycyborg:
This would leak logfile stream object but I'm not sure where best to attach it.

src/log.cpp
```diff
@@ -179,7 +181,11 @@ void set_log_to_file()
 	if(is_log_dir_writable_.value_or(false)) {
 		// get the log file stream and assign cerr+cout to it
 		output_file_path_ = filesystem::get_logs_dir()+"/"+unique_log_filename();
-		output_file_.reset(filesystem::ostream_file(output_file_path_).release());
+		static std::unique_ptr<std::ostream> logfile { filesystem::ostream_file(output_file_path_) };
```
stevecotton, line 184: Could this cause it to be deleted before the last log statement when exiting? I'm worried that it might add a Heisenbug that only triggers when trying to debug a different Heisenbug.

stevecotton:
How does, and how should, this interact with the various command-line options for enabling or disabling logging?

Sorry to say, but I feel this should be a draft until those options' docs are updated.

loonycyborg:
This commit has no particular effect unless logging is enabled. In case of logging is enabled it ensures that everything is written to stderr in addition to the log file. So there is no need to update documentation unless it explicitly states that output on stdout/stderr is suppressed.

src/log.cpp
loonycyborg: Static local variables are destructed at the same time as other global variables, that is already after main() is exited. Thus this can cause trouble only if you're logging from a destructor of global variable which is already an UB because it could happen after logger is destructed.

stevecotton:
I'm not sure if the docs need changing. `--no-log-to-file`'s "log output is written to standard error rather than to a file." would probably be better as "log output is only written to standard error."

Pentarctagon (approved):

PR #7674, opened 2023-05-30
Remove an explicit GETTEXT_DOMAIN "wesnoth"

stevecotton:
C++ source files are considered to be in textdomain "wesnoth" by default.

The 1.16 CMake scripts that sort files into textdomains miss this file, because it treats files with no `GETTEXT_DOMAIN` as being in "wesnoth", but not ones that explicitly declare themselves in "wesnoth".

Doing a pot-update with SCons includes this file anyway, so the official "pot-update and regenerate doc files" commits already include it, even without this fix.

### DIFF
--- a/src/gui/dialogs/addon/install_dependencies.cpp
+++ b/src/gui/dialogs/addon/install_dependencies.cpp
@@ -15,8 +15,6 @@
 
 #include "install_dependencies.hpp"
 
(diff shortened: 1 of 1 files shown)


soliton-:
Sounds like something to fix in cmake. I guess it matters less if this issue does not exist in master.

PR #7675, opened 2023-05-30
[1.16] UtBS: Make the name of the dark assassin translatable again

stevecotton:
It needs to be translatable because it needs to match the other strings that call him by name. It was translatable until 1.15.4, and several of the .po files that are already in the 1.16 still include it as an obsolete message, so the next pot-update run will turn this back into a translated text in Bulgarian, Greek, Irish, Galician, Latin, Lithuanian, Russian, Serbian and Traditional Chinese.

Opening as a PR, but will be linking to this from the forum thread about translations.


stevecotton:
Although there's still discussion going on in the forum, this change would be allowed under the [Fedora freeze rules](https://fedoraproject.org/wiki/L10N_Freezes#What_changes_are_not_affected_by_the_string_freeze?).

PR #7679, opened 2023-05-31
Fixed old typo in definition of unit recruit window

Gothyoba:
Just a typo fix. Nothing else. This should be eligible for merging (I think).


Wedge009:
7 years old, well spotted, and thanks.

PR #7688, opened 2023-06-01
Fixed another old typo

Gothyoba:
This is a 16 year old typo. I don't think this should break anything, though I'm unsure what exactly this file is for. This should probably be eligible for merging.


stevecotton:
My feeling is to just delete this script. Having a checklist of things to do is useful, but the script:

* Still expects configure.ac and Makefile.am
* Only knows part of what to do for .lua files
* Could have the entire `overwrite` function replaced by sed's `--in-place` option

I think it's going to be easier to just use `git grep` to find things that need replacing and to replace them, than it will be to update the script.

Wedge009:
Those may be valid points but there's no harm in correcting the commentary is there?

stevecotton:
@Gothyoba, please would you push a commit to delete the file?

@Wedge009, that's true, but if we merge this then @soliton- or I would immediately open a new PR to delete the file afterwards. So I'd prefer to have the Git history give the credit to the person who prompted the discussion by looking at the file.

PR #7692, opened 2023-06-03
hopefully fixes #6883

### LINKED ISSUE #6883: Error out if data dir is not found automatically

Currently you get an assert that's not very clear about what the issue is: https://github.com/wesnoth/wesnoth/blob/bc94f37f767960f053ff5f77cc5cf3b5100042a1/src/filesystem.cpp#L1497

Better would be to error out if no data dir was found automatically and no `--data-dir` option was given.

tofilwiktor:
I am unsure if this is an acceptable way to do this, I'd be glad to hear any suggestions if it's not.

### DIFF
--- a/src/wesnoth.cpp
+++ b/src/wesnoth.cpp
@@ -1158,7 +1158,7 @@ int main(int argc, char** argv)
 					PLAIN_LOG << "Automatically found a possible data directory at: " << auto_dir;
 				}
 				game_config::path = std::move(auto_dir);
-			} else {
+			} else if(!game_config::path.empty()) {
 				bool data_dir_specified = false;
 				for(int i=0;i<argc;i++) {
 					if(std::string(argv[i]) == "--data-dir" || boost::algorithm::starts_with(argv[i], "--data-dir=")) {
--- a/src/wesnoth.cpp
+++ b/src/wesnoth.cpp
@@ -1159,16 +1159,16 @@ int main(int argc, char** argv)
 				game_config::path = std::move(auto_dir);
 			}
 			else {
-				bool dataDirSpecified = false;
+				bool data_dir_specified = false;
 				for(int i=0;i<argc;i++)
 				{
 					if(std::string(argv[i]) == "--data-dir")
 					{
-						dataDirSpecified = true;
+						data_dir_specified = true;
 						break;
 					}
 				}
-				if (!dataDirSpecified) {
+				if (!data_dir_specified) {
 					PLAIN_LOG << "Cannot find a data directory. Specify one with --data-dir";
 					return 1;
 				}
--- a/src/wesnoth.cpp
+++ b/src/wesnoth.cpp
@@ -89,6 +89,7 @@
 
 #include <boost/iostreams/filtering_stream.hpp> // for filtering_stream
 #include <boost/program_options/errors.hpp>     // for error
+#include <boost/algorithm/string/predicate.hpp> // for checking cmdline options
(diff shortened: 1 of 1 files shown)


gfgtdf:
hmm i think it might be easier just to check for `game_config::path.empty()` after the commandline arguments (in particular data-fir) are parsed.

Pentarctagon:
Either way, keep in mind that wesnoth's code base uses snake case, not camel case.

Also, welcome!

tofilwiktor:
@gfgtdf I have tried doing that but then the data dir defaults to `/usr/local/share/wesnoth`, so it is not empty and the game still errors out.

Wedge009:
> Either way, keep in mind that wesnoth's code base uses snake case, not camel case.

When in doubt, follow what the existing code does.

src/wesnoth.cpp
```diff
@@ -1158,6 +1158,21 @@ int main(int argc, char** argv)
 				}
 				game_config::path = std::move(auto_dir);
 			}
+			else {
+				bool data_dir_specified = false;
+				for(int i=0;i<argc;i++)
+				{
+					if(std::string(argv[i]) == "--data-dir")
```
stevecotton, line 1165: Needs to also recognise `boost::algorithm::starts_with(argv[i], "--data-dir=")`

src/wesnoth.cpp
```diff
@@ -1158,6 +1158,21 @@ int main(int argc, char** argv)
 				}
 				game_config::path = std::move(auto_dir);
 			}
+			else {
+				bool data_dir_specified = false;
+				for(int i=0;i<argc;i++)
+				{
```
stevecotton, line 1164: Coding standard and surrounding code both put the opening brace on the same line as the conditional.

gfgtdf:
> @gfgtdf I have tried doing that but then the data dir defaults to /usr/local/share/wesnoth, so it is not empty and the game still errors out.

But the issue in #6883 is about an assertion that happens exactly when `game_config::path.empty()`. If you get another erro that might be another bug. Also what is the error you get?

soliton-:
Once the empty check for `game_config::path` is added I think this is good to go. Removing all the argument pre-checking we currently have going on is a good idea but doesn't have to be done here.

CelticMinstrel:
Any reason not to show a message box? A typical user won't see the log message.

gfgtdf:
> Any reason not to show a message box? A typical user won't see the log message.

If the data directory is not found then we probably also cannot use our gui framework since it relies on the wml widget definitions, using a different library is probably rather complicated and not worth for this.

CelticMinstrel:
> It seems like SDL has a function `SDL_ShowSimpleMessageBox`

Yes, that's the one I was thinking of.

> if we do that, we should probably try to be consistent with other types of errors that prevent wesnoth form starting

Yes, we definitely should.

> consider not showing a messagebox when there is currently a console attached

Yeah, I was thinking that too, but I'd consider that a low priority. On Linux, it would be if the output stream is a terminal.

src/wesnoth.cpp
```diff
@@ -1157,6 +1158,18 @@ int main(int argc, char** argv)
 					PLAIN_LOG << "Automatically found a possible data directory at: " << auto_dir;
 				}
 				game_config::path = std::move(auto_dir);
+			} else {
```
soliton-, line 1161:
```suggestion
			} else if(!game_config::path.empty()) {
```

PR #7694, opened 2023-06-05
Display Preferences touch-ups

irydacea:
This PR started out as a reorganisation of the Theme selection in the Preferences -> Display section but now includes other layout/label changes in the same section.

Screenshot 2023-06-05 at 01 00 00

## General changes

* The Theme selection is now a dropdown menu in the Preferences panel rather than a sub-dialog (sub-dialogs are not ideal), and the labels clarify it's the Interface theme
* The Show grid/team colours/floating labels options are now grouped with the theme selection as part of a general UI options group
* The Automatic pixel scale multiplier checkbox has a shorter label to avoid redundancy
* The Resolution options now refer to the window size instead, matching the current implementation of Fullscreen which is really Borderless windowed and always uses the desktop resolution, meaning the game's "resolution" is nothing but the non-fullscreen window size
* The VSync and Limit FPS options are grouped together with the Window size options
* The Animated map/water and Show unit standing/idle animations are in two separate groups of their own now

## Other ideas

* Relabel Show team colors as Show team color ellipses?
* Relabel Show floating labels as Show damage indicators? (The game implementation uses "floating labels" to refer to map labels and the like, which are not affected by this option in any way. Even if the internal name were different, the fact is that static labels still "float" in a way)
* Drop "Show" from several options (It is rather redundant)

## Note

* The old Theme selection menu and the `show_theme_dialog()` function are **not** gone and can still be triggered using the in-game console via the `:theme` command
* I left in some commented-out fragments in the first commit I need to remove before merging

### DIFF
--- a/data/gui/window/preferences/03_display.cfg
+++ b/data/gui/window/preferences/03_display.cfg
@@ -375,11 +375,21 @@
 			border_size = 5
 			horizontal_alignment = "left"
 
-			[button]
-				id = choose_theme
-				label = _ "Theme"
+			[label]
+				label = _ "Theme:"
+				definition = "default_small"
+			[/label]
+		[/column]
+
+		[column]
+			border = "all"
+			border_size = 5
+			horizontal_alignment = "left"
+
+			[menu_button]
+				id = "choose_theme"
 				tooltip = _ "Change the in-game theme"
-			[/button]
+			[/menu_button]
 		[/column]
 	[/row]
 #enddef
--- a/data/gui/window/preferences/03_display.cfg
+++ b/data/gui/window/preferences/03_display.cfg
@@ -211,7 +211,7 @@
 
 						[toggle_button]
 							id = "show_ellipses"
-							label = _ "Show team colors"
+							label = _ "Team colors"
 							tooltip = _ "Show a colored circle around the base of each unit to show which side it is on"
 						[/toggle_button]
 					[/column]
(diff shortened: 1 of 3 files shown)


Vultraz (approved):

irydacea:
Opinions on my points under "other ideas"?

Vultraz:
- **Relabel Show team colors as Show team color ellipses?** - I don't think 'ellipses' is particularly intuitive here, I think I'd leave it as-is for now...
- **Relabel Show floating labels as Show damage indicators?** - Maybe 'Damage during combat'? 'Indicators' also seems rather vague.
- **Drop "Show" from several options (It is rather redundant)** - Good idea, go ahead

irydacea:
"Damage during combat" doesn't sound like an improvement to me to be honest. "Combat damage indicators" perhaps?

Vultraz:
Sounds good to me.

irydacea:
I changed the tooltip attached to that option as well, see diff.

**Latest revision**

Screenshot 2023-06-10 at 16 36 01

Vultraz:
Worth explicitly labeling it "Grid overlay"?

irydacea:
I changed the Grid and Team colors labels (as you suggested on Discord) in the latest commit.

Vultraz:
If there's nothing else on this, I say good to go.

PR #7698, opened 2023-06-06
Avoid using double-quote as a C++ raw string delimiter-char

### LINKED ISSUE #7667: xgettext warnings

- **Version:**
Master

### Description of the bug

``translations`` CI workflow produces several warnings when running ``xgettext`` over C++ files.

_No response_

### Expected behavior

_No response_

### Additional context

> src/ai/composite/component.cpp:137: warning: a double-quote in the delimiter of a raw string literal is unsupported
src/editor/map/map_context.cpp:159: warning: a double-quote in the delimiter of a raw string literal is unsupported
src/editor/map/map_context.cpp:175: warning: a double-quote in the delimiter of a raw string literal is unsupported
src/editor/map/map_context.cpp:181: warning: a double-quote in the delimiter of a raw string literal is unsupported
src/editor/map/map_context.cpp:645: warning: a double-quote in the delimiter of a raw string literal is unsupported
/usr/bin/xgettext: warning: msgid 'The following add-on appears to have publishing or version control information stored locally, and will not be removed:' is used without plural and with plural.
                            src/gui/dialogs/addon/manager.cpp:856: Here is the occurrence without plural.
                            src/addon/manager_ui.cpp:192: Here is the occurrence with plural.
                            Workaround: If the msgid is a sentence, change the wording of the sentence; otherwise, use contexts for disambiguation.
/usr/bin/xgettext: warning: msgid 'The following add-on could not be deleted properly:' is used without plural and with plural.
                            src/gui/dialogs/addon/manager.cpp:864: Here is the occurrence without plural.
                            src/addon/manager_ui.cpp:202: Here is the occurrence with plural.
                            Workaround: If the msgid is a sentence, change the wording of the sentence; otherwise, use contexts for disambiguation.
src/server/campaignd/server.cpp:104: warning: a double-quote in the delimiter of a raw string literal is unsupported
src/server/campaignd/server.cpp:104: warning: unterminated string literal
src/server/campaignd/server.cpp:120: warning: unterminated string literal
src/tests/gui/test_gui2.cpp:736: warning: a double-quote in the delimiter of a raw string literal is unsupported
src/tests/gui/test_gui2.cpp:736: warning: unterminated string literal
src/tests/gui/test_gui2.cpp:738: warning: unterminated string literal

stevecotton:
This avoids warnings from xgettext, which considers `R"""..."""` to be sufficiently unusual that it doesn't have support for it.

Sanity-checking with `strings wesnoth_binary | grep map_data` shows the expected strings in the binary.

Fixes #7667


CelticMinstrel:
Why does xgettext care about something that's not defined as an xgettext prefix?

stevecotton:
Because it needs to work out where the raw string ends.

CelticMinstrel:
So basically they're just too lazy to properly parse it.

Do those files even have any translatable strings at all?

stevecotton:
> So basically they're just too lazy to properly parse it.

You could write a patch and send it upstream, but it seems unproductive to complain here without giving a suggestion for what you'd do about the warning instead.

> Do those files even have any translatable strings at all?

Yes they do. Also I think adding exclusions for these files from the pot-update would be worse than just using these different delimiter characters.

We could use a different d-char-sequence, it's a string up to 16 characters long. I wouldn't be happy with stuffing in-jokes in there, but xgettext is happy with `necromancy` as a d-char-sequence.

stevecotton:
@CelticMinstrel is your intention to reject this PR, or are you just letting of steam about a third-party tool?

Vultraz:
Is there any reason not to just use `R"` without a prefix?

CelticMinstrel:
Yes. If you use `R"(` without a prefix, the string cannot contain the sequence `)"`. It looks like that would only break one of the regexes modified by this commit though.

stevecotton:
Two of the regexs include `")` and would break if an empty d-char-sequence was used. I'd prefer to use one d-char-sequence consistently across the source.

IMO, using a d-char-sequence is slightly more readable than using a empty one, it's a reminder that some of the text inside the double-quotes is a delimiter, and thus a reminder that the outermost pair of brackets isn't part of the regex.

CelticMinstrel:
> Two of the regexs include `")`

The `")` sequence is not an issue, it's only `)"` you would have to worry about.

stevecotton:
Ok, only one string is contains `)"`. What about consistency across the source and the way that a non-empty d-char-sequence acts as a reminder that the outermost brackets aren't part of the regex?

CelticMinstrel:
> the way that a non-empty d-char-sequence acts as a reminder that the outermost brackets aren't part of the regex?

This argument does sorta make sense.

stevecotton:
Switched to draft because most of these regexs may be removed soon anyway by #7719 and #7721.

stevecotton:
> The `")` sequence is not an issue, it's only `)"` you would have to worry about.

Which was removed in 10997c39a3d559d699143f67e9ede7947e2193b3, so replacing all of the remaining `R"""(text)"""` strings with just `R"(text)"` is now possible.

soliton- (approved):

PR #7705, opened 2023-06-10
wmlunits: Fix text domains used for translatable strings

### LINKED ISSUE #7704: `u.w.o` several headings and titles aren't in the .pot file
IIUC, the units database uses the translations from the main game, however some of the text that is translated in the in-game help isn't translated on `u.w.o`. Look at `zh_CN` because it's easy to see which headings are still in the Latin alphabet. The "Information", "Attacks (damage × count)", "Terrain" and most of the row-titles in the information box are still in English (although the text "Armageddon Drake" in that area is the unit ID, so isn't translatable).
https://units.wesnoth.org/1.17/mainline/zh_CN/Armageddon%20Drake.html

On the other side of this, we have the not-yet-merged PR #7627, which has the side effect of adding data/tools/unit_tree/html_output.py to the inputs of xgettext, and as you can see from the part of the .pot quoted below, many of the strings aren't used in any other file.

For most of these, it looks as if html_output.py should be changed to match the actual strings used in the game. A few of them are available in textdomain `wesnoth-lib` but not in `wesnoth`.

```
#: data/tools/unit_tree/html_output.py:570
#: src/addon/info.cpp:253
msgid "addon_type^Campaign"

#: data/tools/unit_tree/html_output.py:575
#: src/gui/dialogs/multiplayer/match_history.cpp:69
msgid "Era"

#: data/tools/unit_tree/html_output.py:586
msgid "Race"

#: data/tools/unit_tree/html_output.py:588
msgid "all"

#: data/tools/unit_tree/html_output.py:623
#: src/help/help_impl.cpp:632
msgid "Factions"

#: data/tools/unit_tree/html_output.py:679
msgid "Language"

#: data/tools/unit_tree/html_output.py:908
#: data/tools/unit_tree/html_output.py:1146
msgid "Cost: "

#: data/tools/unit_tree/html_output.py:909
#: data/tools/unit_tree/html_output.py:1147
msgid "HP: "

#: data/tools/unit_tree/html_output.py:910
#: data/tools/unit_tree/html_output.py:1151
msgid "XP: "

#: data/tools/unit_tree/html_output.py:911
msgid "MP: "

#: data/tools/unit_tree/html_output.py:1101
msgid "Advances from: "

#: data/tools/unit_tree/html_output.py:1120
msgid "Advances to: "

#: data/tools/unit_tree/html_output.py:1148
msgid "Moves: "

#: data/tools/unit_tree/html_output.py:1149
msgid "Vision: "

#: data/tools/unit_tree/html_output.py:1150
msgid "Jamming: "

#: data/tools/unit_tree/html_output.py:1152
#: src/units/helper.cpp:59
msgid "Level: "

#: data/tools/unit_tree/html_output.py:1153
#: src/help/help_impl.cpp:1130
#: src/reports.cpp:393
msgid "Alignment: "

#: data/tools/unit_tree/html_output.py:1169
msgid "Abilities: "

#: data/tools/unit_tree/html_output.py:1182
msgid "unit help^Attacks"

#: data/tools/unit_tree/html_output.py:1244
#: src/reports.cpp:510
msgid "Resistances: "

#: data/tools/unit_tree/html_output.py:1281
#: data/tools/unit_tree/html_output.py:1287
msgid "Terrain"

#: data/tools/unit_tree/html_output.py:1288
msgid "Movement Cost"

#: data/tools/unit_tree/html_output.py:1289
msgid "Defense"
```

Slayer95:
Fixes #7704.

This commit also reworks the _ function into several functions, since it would conflict with xgettext's supported keywords.

Test:

```sh
xgettext --language=Python --output=uwo-wesnoth.py.pot --add-comments=TRANSLATORS --copyright-holder="Wesnoth development team" --msgid-bugs-address="https://bugs.wesnoth.org/" --from-code=UTF-8 --sort-by-file --keyword=_w --exclude-file=po/wesnoth/wesnoth.pot data/tools/unit_tree/html_output.py
if test -f uwo-wesnoth.py.pot; then
	echo "Extraneous msgid in wesnoth"
	exit 1;
fi

for SUFFIX in editor help lib; do
	xgettext --language=Python --output=uwo-wesnoth-$SUFFIX.py.pot --add-comments=TRANSLATORS --copyright-holder="Wesnoth development team" --msgid-bugs-address="https://bugs.wesnoth.org/" --from-code=UTF-8 --sort-by-file --keyword=_w_$SUFFIX --exclude-file=po/wesnoth-$SUFFIX/wesnoth-$SUFFIX.pot data/tools/unit_tree/html_output.py
	if test -f uwo-wesnoth-$SUFFIX.py.pot; then
		echo "Extraneous msgid in wesnoth-$SUFFIX"
		exit 1;
	fi
done
```

### DIFF
--- a/data/tools/unit_tree/html_output.py
+++ b/data/tools/unit_tree/html_output.py
@@ -364,8 +364,12 @@ class HTMLOutput:
         self.isocode = isocode
         global_htmlout = self
 
-    def translate(self, string, domain):
-        return self.translation.translate(string, domain)
+    def make_gettext_single(self, domain):
+        # xgettext requires a different function per text domain.
+        return lambda string: self.translation.translate(string, domain)
+
+    def make_gettext(self, *args):
+        return tuple(map(self.make_gettext_single, args))
 
     def analyze_units(self, grouper, add_parents):
         """
@@ -477,18 +481,8 @@ class HTMLOutput:
     def write_navbar(self, report_type):
         def write(line):
             self.output.write(line)
-        # xgettext requires a different function per text domain.
-        def _w(msgid):
-            return self.translate(msgid, "wesnoth")
 
-        def _w_editor(msgid):
-            return self.translate(msgid, "wesnoth-editor")
-
-        def _w_help(msgid):
-            return self.translate(msgid, "wesnoth-help")
-
-        def _w_lib(msgid):
-            return self.translate(msgid, "wesnoth-lib")
+        _w, _w_editor, _w_help, _w_lib = self.make_gettext("wesnoth", "wesnoth-editor", "wesnoth-help", "wesnoth-lib")
 
         all_written_html_files.append((self.isocode, self.output.filename))
 
@@ -814,18 +808,8 @@ class HTMLOutput:
     def write_units(self):
         def write(line):
             self.output.write(line)
-        # xgettext requires a different function per text domain.
-        def _w(msgid):
-            return self.translate(msgid, "wesnoth")
 
-        def _w_editor(msgid):
-            return self.translate(msgid, "wesnoth-editor")
-
-        def _w_help(msgid):
-            return self.translate(msgid, "wesnoth-help")
-
-        def _w_lib(msgid):
-            return self.translate(msgid, "wesnoth-lib")
+        _w, _w_editor, _w_help, _w_lib = self.make_gettext("wesnoth", "wesnoth-editor", "wesnoth-help", "wesnoth-lib")
 
         rows = self.unitgrid
         write('<table class="units">\n<colgroup>')
--- a/data/tools/unit_tree/html_output.py
+++ b/data/tools/unit_tree/html_output.py
@@ -997,6 +997,8 @@ class HTMLOutput:
         return n
 
     def write_unit_report(self, output, unit):
+        _w, _w_editor, _w_help, _w_lib = self.make_gettext("wesnoth", "wesnoth-editor", "wesnoth-help", "wesnoth-lib")
+
         def write(line):
             self.output.write(line)
         def _(msgid, textdomain="wesnoth"):
--- a/data/tools/unit_tree/html_output.py
+++ b/data/tools/unit_tree/html_output.py
@@ -108,7 +108,7 @@ WESMERE_HEADER = '''\
 </div>
 </div>
 
-<div id="content" role="main">'''
+<div id="content" role="main">'''  # TODO?: Localize header
 
 WESMERE_FOOTER = '''\
 </div> <!-- end content -->
@@ -124,7 +124,7 @@ WESMERE_FOOTER = '''\
 </div></div></div>
 
 </body></html>
-'''
+'''  # TODO?: Localize footer
 
 HTML_CLEAR_FLOATS = '<div class="reset"></div>'
 
@@ -150,7 +150,7 @@ def website_footer():
 
 def build_timestamp():
     """Returns an element containing the current date and time."""
-    return '<div id="lastmod">Last updated on %s.</div>' % time.ctime()
+    return '<div id="lastmod">Last updated on %s.</div>' % time.ctime() # TODO?: Localize timestamp
 
 
 all_written_html_files = []
@@ -477,8 +477,18 @@ class HTMLOutput:
     def write_navbar(self, report_type):
         def write(line):
             self.output.write(line)
-        def _(msgid, textdomain="wesnoth"):
-            return self.translate(msgid, textdomain)
+        # xgettext requires a different function per text domain.
+        def _w(msgid):
+            return self.translate(msgid, "wesnoth")
+
+        def _w_editor(msgid):
+            return self.translate(msgid, "wesnoth-editor")
+
+        def _w_help(msgid):
+            return self.translate(msgid, "wesnoth-help")
+
+        def _w_lib(msgid):
+            return self.translate(msgid, "wesnoth-lib")
 
         all_written_html_files.append((self.isocode, self.output.filename))
 
@@ -567,12 +577,12 @@ class HTMLOutput:
         # campaigns/eras navigation.
 
         # Campaigns
-        add_menu("campaigns_menu", _("addon_type^Campaign"))
+        add_menu("campaigns_menu", _w("addon_type^Campaign"))
         write(PRE_PLACEHOLDER_CAMPAIGNS.decode('utf-8'))
         end_menu()
 
         # Eras
-        add_menu("eras_menu", _("Era"))
+        add_menu("eras_menu", _w_lib("Era"))
         write(PRE_PLACEHOLDER_ERAS.decode('utf-8'))
         end_menu()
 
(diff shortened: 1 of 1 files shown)


data/tools/unit_tree/html_output.py
```diff
@@ -1143,30 +1163,30 @@ def clean_uval(name):
...
-            ("cost",       _("Cost: ")),
-            ("hitpoints",  _("HP: ")),
-            ("movement",   _("Moves: ")),
-            ("vision",     _("Vision: ")),
-            ("jamming",    _("Jamming: ")),
-            ("experience", _("XP: ")),
-            ("level",      _("Level: ")),
-            ("alignment",  _("Alignment: ")),
-            ("id",         "Id: ")
+            ("cost",       _w_editor("Cost: ")),
```
stevecotton, line 1166: Removing the trailing space finds a lot of these strings in `wesnoth-help`, which makes sense to me as it's where the in-game help page should come from.

That may also snowball into working out which other strings in wesnoth-help shouldn't have trailing spaces, will try and find a simpler way to unblock #7627.

data/tools/unit_tree/html_output.py
```diff
@@ -150,7 +150,7 @@ def website_footer():
 
 def build_timestamp():
     """Returns an element containing the current date and time."""
-    return '<div id="lastmod">Last updated on %s.</div>' % time.ctime()
+    return '<div id="lastmod">Last updated on %s.</div>' % time.ctime() # TODO?: Localize timestamp
```
Wedge009, line 153: Whether UTC or the user's local time, it really ought to be specified. Plenty of times I've assumed some US time-zone (because the Anglophonic web seems to stuck with US-centric conventions more often than not), but it's actually been UTC - and vice-versa too.

https://docs.python.org/3/library/datetime.html#datetime.datetime.ctime
It appears `ctime()` explicitly states that it won't do this (specify the time-zone). From what I can see, the string is just used at the bottom of a page ('fine print') so I think `isoformat(sep=' ', timespec='seconds')` will be an improvement over `ctime()`.

data/tools/unit_tree/html_output.py
```diff
@@ -477,8 +481,8 @@ def grid_place(nodes, x):
     def write_navbar(self, report_type):
         def write(line):
             self.output.write(line)
-        def _(msgid, textdomain="wesnoth"):
-            return self.translate(msgid, textdomain)
+
+        _w, _w_editor, _w_help, _w_lib = self.make_gettext("wesnoth", "wesnoth-editor", "wesnoth-help", "wesnoth-lib")
```
stevecotton, line 485: Putting them on a single line is less readable, IMO. Also, it means the entire line changes if a domain is added or removed (as I hope _w_editor will be in a subsequent PR).

data/tools/unit_tree/html_output.py
```diff
@@ -583,9 +587,9 @@ def end_menu(is_table_container=False):
             target = "mainline.html"
 
         if not self.is_era:
-            add_menu("races_menu", _("Race", "wesnoth-lib"))
+            add_menu("races_menu", _w_lib("Race"))
 
-            add_menuitem('mainline.html', _("all", "wesnoth-editor"))
+            add_menuitem('mainline.html', _w_editor("all"))
```
stevecotton, line 592: Not an issue for this PR, but this is the only string coming from `_w_editor`, it's got no context to say what it is, and it seems out of place.

The meaning here is "show the page with the full list of units sorted by race, and jump to the top", in a menu where every other option is the name of a race, but just means "jump to the same unfiltered list, but scroll to the race's units". So if we weren't constrained to available strings, I'd suggest "top" or "top of page".

The meaning in `wesnoth-editor` is "show an unfiltered list of terrains/units/items". The same meaning in `wesnoth-lib` is found in the add-on manager's "None Selected", which wouldn't fit here.

data/tools/unit_tree/html_output.py
Slayer95: Is this really superior?

```python
_w = self.make_gettext("wesnoth")
_w_editor = self.make_gettext("wesnoth-editor")
_w_help = self.make_gettext("wesnoth-help")
_w_lib = self.make_gettext("wesnoth-lib)
```
Remember that snippet would have to be included more than once (in fact, 3 times -the current state of the PR only 2.)

CelticMinstrel: If the worry is about changing the entire line just for one domain, then what about this instead?

```python
_w,
_w_editor,
_w_help,
_w_lib =
    self.make_gettext(
        "wesnoth",
        "wesnoth-editor",
        "wesnoth-help",
        "wesnoth-lib"
    )
```

Not the most pretty formatting, but does avoid touching the whole statement just to add or remove a domain.

I'm not sure if a `\` is also needed at the end of the first 3 lines – I know `=` and `(` trigger statement continuation but I'm not sure if `,` does.

Pentarctagon:
Any reason not to merge this?

Pentarctagon:
Well, I'll go ahead and merge this in a few days if nothing else is said.

Wedge009:
Well, I was only concerned about the 'last updated' string, but that seems 'out of scope'.

data/tools/unit_tree/html_output.py
Wedge009: I just did a quick test and `\` was needed for the first four lines - including the `=`.

Pentarctagon:
If the change for that would just be changing `ctime()` to `isoformat(sep=' ', timespec='seconds')`, then that'd be trivial to do post-merge.

Wedge009:
Only if there was a consensus, but didn't seem like there was any interest.

soliton-:
Displaying the time zone would certainly be useful.

Wedge009:
Just in case it wasn't clear, I was waiting for this to be merged before attempting to do anything with the time-stamp.

Slayer95:
I have just pushed what I had in the staging area, as I mentioned in https://github.com/wesnoth/wesnoth/pull/7705#discussion_r1226337434. But the reviews haven't been addressed yet.

stevecotton:
This one isn't blocked during the string freeze - it's reusing strings from mainline, but changing these .py files doesn't result in changes to the .po files.

Pentarctagon:
It sounds like this is something that can be merged then? Or were there still issues with it?

soliton-:
I'd merge this if all the TODO stuff is removed and the `quote=False` use is fixed. As far as I can tell it is used a lot even when injecting into quotes. Admittedly there is wrong use already existing but at least this shouldn't add more.

Pentarctagon:
@Slayer95 do you plan to come back to this PR at some point?

Slayer95:
This needs a rebase, and some.... code style bikeshedding...?

Wedge009:
If it's just style issues outstanding then it should be fine merge, right? GitHub says there are no conflicts on re-base. Also the Windows check failure has expired so the log is no longer available - can always re-run the check, I suppose.

soliton-:
The quoting issue is not style. If you inject into an attribute using `quote=False` is simply wrong.

Pentarctagon:

PR #7710, opened 2023-06-12
WML Tools: Retain file owner, group, and permissions when writing new file

### LINKED ISSUE #7708: wmlindent changes file permissions
wmlindent replaces the original file with a new one. That way
- file permissions are set to the default, which might differ from the previous permissions
- if the file was previously linked by more than one hardlink, this would no longer be the case

Wedge009:
Resolves #7708

I only tested on Linux, not sure if there needs to be other things done to accommodate Windows platform (as with `os.rename`).

Would also appreciate review of the other changes too - some really weird code here, I'm guessing due to age, like use of equality operations where it looks like it's meant to be an assignment, plus a nonsensical try-except block that doesn't do anything with the exception.

### DIFF
--- a/data/tools/wmlindent
+++ b/data/tools/wmlindent
@@ -288,41 +288,51 @@ def convertor(linefilter, arglist, exclude):
                         print("wmlindent: %s excluded" % filename, file=sys.stderr)
                     continue
                 else:
+                    backup = f'{filename}-bak'
+                    if sys.platform == 'win32' and os.path.exists(backup):
+                        os.remove(backup)
+                    os.rename(filename, backup)
+
                     try:
-                        with codecs.open(filename,"r","utf8") as infp, codecs.open(filename + ".out", "w","utf8") as outfp:
+                        with codecs.open(backup, "r", "utf8") as infp,\
+                                codecs.open(filename, "w", "utf8") as outfp:
                             linefilter(filename, infp, outfp)
+
                     except bailout as e:
+                        os.remove(filename)
+                        os.rename(backup, filename)
                         print('wmlindent: "%s", %d: %s' % (e.filename, e.lineno, e.msg), file=sys.stderr)
-                        os.remove(filename + ".out")
+
                     except UnicodeDecodeError as e:
-                        os.remove(filename + ".out")
+                        os.remove(filename)
+                        os.rename(backup, filename)
                         print('wmlindent: "{}" is not a valid UTF-8 file'.format(filename), file=sys.stderr)
+
                     except KeyboardInterrupt:
-                        os.remove(filename + ".out")
+                        os.remove(filename)
+                        os.rename(backup, filename)
                         print("wmlindent: %s interrupted" % filename, file=sys.stderr)
+
                     else:
-                        if filecmp.cmp(filename, filename + ".out"):
+                        if filecmp.cmp(filename, backup):
                             if verbose >= 2:
                                 print("wmlindent: %s unchanged" % filename, file=sys.stderr)
-                            os.remove(filename + ".out")
+                            os.remove(filename)
+                            os.rename(backup, filename)
+
                         else:
                             if not quiet:
                                 print("wmlindent: %s changed" % filename, file=sys.stderr)
+
                             if dryrun:
-                                os.remove(filename + ".out")
+                                os.remove(filename)
+                                os.rename(backup, filename)
+
                             else:
-                                file_stat = os.stat(filename)
-
-                                os.remove(filename) # For Windows portability
-                                # There's a tiny window open if you keyboard-
-                                # interrupt here. It's unavoidable, because
-                                # there's no known way to do an atomic rename
-                                # under Windows when the target exists -- see
-                                # Python manual 14.1.4::rename()
-                                os.rename(filename + ".out", filename)
-
+                                file_stat = os.stat(backup)
                                 os.chmod(filename, file_stat.st_mode)
                                 os.chown(filename, file_stat.st_uid, file_stat.st_gid)
+                                os.remove(backup)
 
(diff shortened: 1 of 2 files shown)


irydacea:
While the intention is appreciated, this ultimately doesn't solve the root problem highlighted in #7708, which is the fact that the WML tools create a new file instead of overwriting the existing one. Sure, you can manually fix permissions and ownership, but you'll still break hard links simply because of the fact that the inode number changes; and you won't be able to manually fix links without scanning the entire filesystem looking for those I'm afraid.

A better solution would be to create a backup of the file first and then modify it in place.

Wedge009:
Looking more closely, that's the approach `wmllint` already takes. I've attempted to modify `wmlindent` to do the same.

Admittedly, there's a lot of repetition here, but I would only feel comfortable to refactor the repetition if I was making a more comprehensive rewrite.

Elvish-Hunter:
> Admittedly, there's a lot of repetition here, but I would only feel comfortable to refactor the repetition if I was making a more comprehensive rewrite.

@Wedge009 Indeed there's a lot of repetition, however I've seen that most of them are in wmlindent's `try ... except ... else` block. As it is structured now, it looks like the following two lines
```python
    os.remove(filename)
    os.rename(backup, filename)
```
are always executed, no matter if there's an error or not. In this case, perhaps moving them under a `finally:` statement (which is always executed) might be a solution?

Wedge009:
Yeah, I thought about the `finally` aspect, but at the time was just trying to make sure I got the logic right. I'll review it, soon, thanks for the review.

Wedge009:
I remember why I didn't use `finally` - there is still the one case where we don't want to revert to the original file.

At any rate, I need to revise this - I realised it's not the renaming that's dropping the meta-data, it's the writing of a new file.

Wedge009:
Okay, I think this makes more sense now, plus I found cross-platform renaming is supposed to be done with `os.replace` - that eliminates the repetition of `os.remove` followed by `os.rename`.

CelticMinstrel:
Is the comment about not modifying the timestamp still true?

Wedge009:
When there is no change made, yes.

stevecotton:
`os.replace` breaks hardlinks. Not sure if that's still an issue, because it seems to only be used if the operation errors out anyway.

Wedge009:
I don't know what to do then. I wonder if links - hard or soft - is a reasonable situation given I expect this to be used on Wesnoth data files. Also, the suggestion was to operate on the original file rather than write to a new file and then replace the original. What other alternatives might there be?

Does the way `wmlindent` currently stands work with hard links?

stevecotton:
`wmlindent` breaks hard links, both with and without this PR. Haven't thought about your other questions, but might reply again later (hopefully someone else will first).

ProditorMagnus:
I dont think its worth changing just to support links better, but creating backup and third output file, and in case of backup!=third emptying and writing to original file should work.

Wedge009:
It's not just links, it's all meta-data associated with a file. The original issue focused on permissions, and that's what I looked at. (@sevu Thoughts?)

Having three files doesn't sound like it will solve the issue. Would still need to make copies and then replace the original eventually - I assume that links will still break in that case. Can anyone describe to me a realistic situation where one has Wesnoth data files set-up with links?

soliton-:
I think the core question is if it's more important to have a consistent file in case of an error by writing to a new file and on success doing an atomic rename. Or if it's more important to keep the original file at the risk of getting an inconsistent file when an error occurs but keep meta data, not break links, maintain open files etc.

As far as I can tell wmlindent takes the first approach and wmllint does somewhat as well but renames the original first and the new file gets the old name, so no consistency guarantee.

Wedge009:
I'm not a regular user of these tools so I defer to others on that decision. Sounds like status quo is preferable and @sevu's request will be indefinitely on hold?

soliton-:
I mean the status quo on wmllint seems fairly useless. It neither keeps the original file (well, it does as the backup but that's not very useful) or gives a consistency guarantee. I think your change there is good. That would make wmllint work like the second case. Ideally the revert feature would then also keep the original file by reading in the backup file content and rewriting the original file with it instead of renaming the (new) backup file over the old name.

It's probably more useful to keep the original file than having a consistency guarantee. It's not like people are using these tools and concurrently starting wesnoth and expecting a working addon or so. And if an error happens (out of disk space or something, not an internal tool error) they should be able to fix that and potentially restore stuff from the backup files.

I think it's reasonable to switch wmlindent to that mode as well.

stevecotton:
> the status quo on wmllint seems fairly useless

Alternative suggestion: make wmllint be `--dry-run` by default, and suggest that users create backups or use version control themselves before they run it with `--no-dry-run`.

sevu:
I do not use hardlinks here, it is more like, if hardslinks are intact, everything else is. And for a tool which is told to optimize a file it would be an unexpected side effect.

I like the changes you did and would like to see that in. If I see that correctly, with using the three line 
If win: os.remove; os.move instead of the one-line os.replace it would cover the situation more broadly.

Wedge009:
Are you saying to revert to the original remove-target-for-win32-before-rename? According to the documentation, changing to replace is supposed to enable the same behaviour across Windows and Unix platforms.

I'm wondering if the issues with links can be left as a separate issue since it's already a problem in WML tools even without this change.

soliton-:
I don't think the current wmlindent changes are good. If there is no desire to switch it to keeping the original file then it should just be left as is except using copy2 or whatever to copy metadata. In the current PR state it's a surprising mix of new and original file.

soliton-:
I would definitely not abandon everything in this PR. At a minimum the wmllint changes are all good as far as I can tell.

PR #7711, opened 2023-06-12
Contributing: the Typography and Gettext docs are separate Wiki pages

stevecotton:
There's also <https://wiki.wesnoth.org/UI_Style_Guide>, which I haven't linked here.


Wedge009:
The original seems like an error. I think this should be merged as soon as reasonable.

PR #7713, opened 2023-06-13
Create test for special_id_active in the style of eoma_beam

ProditorMagnus:
Test created to verify that special_id_active does not work in 1.16.0.

https://github.com/inferno8/wesnoth-Era_of_Magic/issues/30 https://github.com/inferno8/wesnoth-Era_of_Magic/pull/54


gfgtdf:
I think this is specific to `[filter_attack]`in `[event]`. If you'd use `[filter][has_attack]` it would probably work

gfgtdf:
And here the commit that fixed your usecase in 1.17 : https://github.com/wesnoth/wesnoth/pull/7509/files

ProditorMagnus:
> (it uses a trick that a special with a filter_self is never considered active unless its the currently used attack)
From user interface it looked like it is always considered active.

gfgtdf:
https://github.com/wesnoth/wesnoth/pull/7509/files seems to already added some test for it, not sure whetehr it makes this redundant or not

gfgtdf:
> From user interface it looked like it is always considered active.

Its possible that the ui code handles thing a bit differently and handles filters by default as true when no unit  is found, while the actual attack code does evaluate it to false in this case.

ProditorMagnus:
I noticed the existing tests too when creating change for test schedule. With use of lua_function it is still separate enough case to test.

data/test/scenarios/wml_tests/FilterWML/special_id_active.cfg
```diff
@@ -0,0 +1,111 @@
+# This unit test reproduces setup of weapon special such as https://github.com/inferno8/wesnoth-Era_of_Magic/blob/adf227cba2e66602c4bb49dc203fa751865aa124/utils/beam.cfg, which is not supposed to work in rpg.
+# Focus of this test is whether [filter_attack]special_id_active work for [filter_self]lua_function controlled conditional weapon special
```
stevecotton, line 2: Please put comments in the "API(s) being tested:" template format, and add the extras that that template suggests.

Line 1 is good and should stay, but line 2 would fit in the standard template.

data/test/scenarios/wml_tests/FilterWML/special_id_active.cfg
```diff
@@ -0,0 +1,111 @@
...
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        [unstore_unit]
+            variable=a
+            find_vacant=yes
+            x,y=$b.x,$b.y
```
stevecotton, line 51: Instead of `GENERIC_UNIT_TEST`, using `COMMON_KEEP_A_B_UNIT_TEST` will put Alice in position to attack, and then you can remove this `[unstore_unit]` along with the `[store_unit]` on line 35.

data/test/scenarios/wml_tests/FilterWML/special_id_active.cfg
ProditorMagnus: updated and named file to match test id

data/test/scenarios/wml_tests/FilterWML/special_id_active.cfg
ProditorMagnus: done

data/test/scenarios/wml_tests/FilterWML/special_id_active_lua_function.cfg
```diff
@@ -0,0 +1,106 @@
...
+#####
+
+{COMMON_KEEP_A_B_UNIT_TEST "special_id_active_lua_function" (
+    [event]
+        name=preload
+        first_time_only=yes
+        [lua]
+            code=<<
+conditional_special_active = function()
+	return wml.variables.conditional_special_active
```
stevecotton, line 17: Having the same name for the Lua function and the WML variable makes the test harder to read. I suggest renaming the WML variable as `allow_extra_damage`.

Also, the indentation is off here (and wmlindent seems happy both as-is and when it's corrected).

ProditorMagnus: It is Lua, out of scope of wmlindent.

stevecotton: Yes, but the coding standard that's at least de-facto for the data/test directory is to have `>>` at the same level of indentation as `code=<<`, and the first line of the Lua being one indent more than that.

data/test/scenarios/wml_tests/FilterWML/special_id_active_lua_function.cfg
```diff
@@ -0,0 +1,106 @@
...
+        name=start
+        [object]
+            silent=yes
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [damage]
+                        id=conditional_special
+                        name="conditional_special"
```
stevecotton, line 77: The WML validation wants a translation mark.

data/test/scenarios/wml_tests/FilterWML/special_id_active_lua_function.cfg
ProditorMagnus: I wouldnt mix Lua and wml indents.

I found one example of my preferred style
```
data/campaigns/Son_Of_The_Black_Eye/utils/utils.cfg-    [lua]
data/campaigns/Son_Of_The_Black_Eye/utils/utils.cfg:        code = <<
data/campaigns/Son_Of_The_Black_Eye/utils/utils.cfg-function wesnoth.custom_synced_commands.ship_unload(cfg)
```

ProditorMagnus: I would accept compromise as code=<<conditional_special_active = function() return wml.variables.allow_extra_damage end>>

ProditorMagnus:
Is Lua code style the only concern left?

Pentarctagon:
It looks like it. IMO the indentation should be kept consistent with how its done elsewhere in mainline.

stevecotton:
The indentation was the main part. Thanks for starting to change it, but to match mainline it should be indented one level further, and there's also a tabs-vs-spaces issue to fix.

The other other issue is the comment block, which doesn't match the format of the other files.

ProditorMagnus:
Can you explain what you mean by tabs-vs-spaces issue?

stevecotton:
The reindented Lua blocks are indented with tabs, but they're in files that are indented with spaces.

ProditorMagnus:
All mainline Lua I checked recently uses tabs.

ProditorMagnus:
Updated comment format

ProditorMagnus:
Added extra levels.

stevecotton:
> All mainline Lua I checked recently uses tabs.

That only seems to be for .lua files, not for Lua blocks within .cfg files when the .cfg itself is already using spaces.

There are a few that use tabs within .cfg files, which can be found with `check_mixed_indent`, but it's a small minority. Those should be fixed for consistency, but please don't add them to this PR - I'd prefer the PR to just be about the test (and not to touch SotBE).
```
data/tools/check_mixed_indent `find data -name '*.cfg'`
```

ProditorMagnus:
That command gives `/usr/bin/env: 'python3': Argument list too long`

ProditorMagnus:
Works with ```grep -r -l '[lua]' `find data -name '*.cfg'` | xargs data/tools/check_mixed_indent```

ProditorMagnus:
82 files with `[lua] with only one type of indent`
5 files with `uses both tabs and spaces for indents has_lua: True`

But some of those 82 solve it by not using indent at all, for example `data/core/_main.cfg`

data/test/scenarios/wml_tests/FilterWML/special_id_active_lua_function.cfg
```diff
@@ -0,0 +1,115 @@
+#textdomain wesnoth-test
+
+#####
+# This unit test reproduces setup of weapon special such as https://github.com/inferno8/wesnoth-Era_of_Magic/blob/adf227cba2e66602c4bb49dc203fa751865aa124/utils/beam.cfg, which is not supposed to work in rpg.
+##
+# API(s) being tested: [event][filter_attack]special_id_active, weapon special [filter_self]lua_function
```
stevecotton, line 6: To fit with the other tests, this line should be, without a space after the comma:
`[event][filter_attack]special_id_active=,[damage][filter_self]lua_function=`

With @Pentarctagon's test reorg, his aim was to reduce the number of APIs listed in that line. However, I can see the logic of saying that this is specifically testing that the event filtering code has a working lua filter, whether or not that filtering code is shared with other weapon filters.

stevecotton:
The only blocker for this is getting that SotBE file out of the PR. Would everyone be happy if I did a manual squash for that?

ProditorMagnus:
I can try replacing sotbe with master version, then squash should hopefully drop those commits.

ProditorMagnus:
That doesnt seem useful actually, with https://github.com/wesnoth/wesnoth/commit/acb4fc30b02658150f3e7e801a92dca4cd4c58b8 it does not look safe.

data/test/scenarios/wml_tests/FilterWML/special_id_active_lua_function.cfg
Pentarctagon: More generally, the goal is for unit tests to be primarily testing only one thing (or as few things as possible) even if they depend on other things also working.

PR #7716, opened 2023-06-14
[1.16] WC: add po hints to the messages when units join

stevecotton:
These didn't have context in the .po files, so this makes clear which unit is joining.

The Ulf and Ghoul are special, a couple of others get a bit more detail about British slang.

### DIFF
--- a/data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
+++ b/data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
@@ -5,188 +5,188 @@ type_infos = {
 		founddialogue = _"You look like you could use some help. Mind if I join in? It’s been a while since I had a good fight!",
 		reply = _ "Excellent. We could always use more help.",
 	},
-	["Orcish Grunt"] = { 
+	["Orcish Grunt"] = {
 		founddialogue=_"’bout time. Been forever since I had a good fight, eh?",
 		image="units/orcs/grunt.png",
 		name="Orcish Grunt",
 	},
-	["Troll Whelp"] = { 
+	["Troll Whelp"] = {
 		founddialogue=_"Who you? I help you smash!",
 		image="units/trolls/whelp.png",
 		name="Troll Whelp",
 	},
-	["Orcish Archer"] = { 
+	["Orcish Archer"] = {
 		founddialogue=_"My clan was destroyed long ago, leaving me to fend for myself. Let me join your army and I will fight as though your clan was my own!",
 		image="units/orcs/archer.png",
 		name="Orcish Archer",
 	},
-	["Orcish Assassin"] = { 
+	["Orcish Assassin"] = {
 		founddialogue=_"Hey boss. Looks like you’ve got some prey that need killing.",
 		image="units/orcs/assassin.png",
 		name="Orcish Assassin",
 	},
-	["Wolf Rider"] = { 
+	["Wolf Rider"] = {
 		founddialogue=_"I hunts. I join you, I hunts good for you!",
 		image="units/goblins/wolf-rider.png",
 		name="Wolf Rider",
-		alt_reply = { { 
+		alt_reply = { {
 			race="wose",
 			reply=_"Very well. But take care where your dog does its stuff.",
 		} }
 	},
-	["Orcish Leader"] = { 
+	["Orcish Leader"] = {
 		founddialogue=_"Heh, looks like you whelps might be getting in over your heads. Good thing I’m here now to win this fight for ya, huh?",
 		image="units/orcs/leader.png",
 		name="Orcish Leader",
 	},
-	["Naga Fighter"] = { 
+	["Naga Fighter"] = {
 		founddialogue=_"I too have come a long way to this strange land. Perhaps we were destined to join blades here.",
 		image="units/nagas/fighter.png",
 		name="Naga Fighter",
-		alt_reply = { { 
+		alt_reply = { {
 			race="gryphon",
 			reply=_"Looks like we fished some tasty help.",
 		} }
 	},
-	["Elvish Fighter"] = { 
+	["Elvish Fighter"] = {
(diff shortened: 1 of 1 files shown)


data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
```diff
@@ -153,93 +180,113 @@ type_infos = {
...
 		image="units/dwarves/ulfserker.png",
 		name="Dwarvish Ulfserker",
 		alt_reply = { { 
 			race="dwarf",
+			-- po: the speaker is another dwarf, after an Ulf joins their side
 			reply=_"Brave words. Welcome to The Fight Club...",
 		}, {
 			race="human",
+			-- po: the speaker is a human, after an Ulf joins their side. TODO: how does this not result in an immediate fight?
 			reply=_"Brave words. Welcome little big man.",
```
stevecotton, line 230: Can't be changed for 1.16, but making a height joke to an Ulf who you've just met seems a foolish act.

data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
```diff
@@ -153,93 +180,113 @@ type_infos = {
...
 	},
 	["Dwarvish Scout"] = { 
+		-- po: a Dwarvish Scout joins a player's side
 		founddialogue=_"Having trouble, eh? Never worry, lads, we’ll sort ’em out soon enough!",
 		image="units/dwarves/scout.png",
 		name="Dwarvish Scout",
 	},
 	["Drake Fighter"] = { 
+		-- po: a Drake Fighter joins a player's side. TODO: They have war blades, not swords.
 		founddialogue=_"The ancient spirits tell me my destiny lies with yours. My sword is at your command.",
```
stevecotton, line 256: Should we suggest translating "sword" as the war blade?

data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
stevecotton: Sometimes, the role of a PR is to catch what-was-I-thinking commits like this. Thanks, I'll remove the TODOs and make an issue for them.

stevecotton: If it's meant to be a reference to [a film](https://en.wikipedia.org/wiki/Little_Big_Man_(film)) or [a real Native American Chief](https://en.wikipedia.org/wiki/Little_Big_Man), then I guess "Brave words" is also a reference that I missed.

However, a PO hint should either say something about the references, or suggest an alternative that could be translated without the references.

data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
```diff
@@ -350,51 +417,45 @@ type_infos = {
 		} }
 	},
 	["Woodsman"] = { 
+		-- po: a Woodsman joins a player's side
 		founddialogue=_"Ho there, friends! I am but a soldier of humble circumstances, yet long have I dreamed of joining great wars beyond our shores. Let me join your mission!",
 		name="Woodsman",
 	},
 	["Dune Herbalist"] = { 
+		-- po: a Dune Herbalist joins a player's side
```
soliton-, line 425: Same as Mage.

data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
```diff
@@ -350,51 +417,45 @@ type_infos = {
...
 		name="Dune Herbalist",
 	},
 	["Dune Soldier"] = { 
+		-- po: a Dune Soldier joins a player's side
 		founddialogue=_"I too have come a long way to this strange land. Perhaps we were destined to join blades here.",
 		image="units/dunefolk/soldier.png",
 		name="Dune Soldier",
 	},
 	["Dune Rover"] = { 
+		-- po: a Dune Rover joins a player's side
```
soliton-, line 437: Same as Spearman and Woodsman.

stevecotton: While it doesn't quite look right, I think this would be as helpful as it can be for translators, and less likely to break if there are revisions to the lines.

All the hints for a line accumulate, so in Poedit's hint box it will say
> a Spearman joins a player's side
> a Woodsman joins a player's side
> a Dune Rover joins a player's side

soliton-: I assumed you wanted to combine the hints since you did it for
> po: speaker is a Gryphon, just after a naga or merfolk joins their side

for example. I don't mind either way.

stevecotton:
I've added a whitespace cleanup as a separate commit, which makes cherry-picking to the master branch much easier.

Noteworthy units in the latest version:
* The Ulf's comment is expanded. It no longer says TODO but I'm still expecting it to change in 1.17.
* The Drake Fighter's comment now says the unit will have a war blade instead of a sword.
* The Dune Piercer no longer has a translation comment, just a normal comment saying the unit doesn't exist in 1.16. The string is still used by the Orcish Leader.

data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua
stevecotton: The comments have been updated, and the TODO removed (although I still think the line is likely to change in 1.17).

Github isn't marking this as `Outdated`, even though it is outdated, because this comment thread is on the unchanged line rather than the `po` line.

Hejnewar: I agree with soliton here. IMO its fine but I dont have any strong feelings about that. 

At least it sounds like a good way to make ulf follow you. :P

Hejnewar (approved):

Issue #4316, opened 2019-09-05
Accept [filter] in StandardUnitFilter


ProditorMagnus:
Goal is to solve the issue of some tags asking for [filter], and other tags asking for content of [filter].

I think it would make most sense to treat
```
[filter]
  filter_1
  [filter]
    filter_2
  [/filter]
[/filter]
```
as
```
[filter]
  filter_1
  [and]
    filter_2
  [/and]
[/filter]
```

This has been discussed before, I think there were technical issues in supporting this.

gfgtdf:
> Goal is to solve the issue of some tags asking for [filter], and other tags asking for content of [filter].

can you please explain this more, i dont really understand whats the issue here.

AI0867:
Some tags *are* a StandardUnitFilter, with a few extra keys. Some *contain* a `[filter]` that is a StandardUnitFilter.

Example of the former: https://wiki.wesnoth.org/DirectActionsWML#.5Bkill.5D
Example of the latter: https://wiki.wesnoth.org/DirectActionsWML#.5Bmodify_unit.5D

GregoryLundberg:
Would it be possible to maintain a state indicating we've seen a filter clause? In the OP, the first form, with the embedded [filter] should be a syntax error. However, in a [kill] enclosing the SUF in [filter] should only provoke a warning.

gfgtdf:
> Some tags are a StandardUnitFilter, with a few extra keys. Some contain a [filter] that is a StandardUnitFilter.

Yes i know, but whats the problem with that?

stevecotton:
I'm in two minds about the change itself, but if this isn't merged then I think we should print a warning when finding a `[filter]` inside an SUF.

GregoryLundberg:
@gfgtdf The issue as I see it is, unless you know it already, or have the WML manual pages on screen for each tag, it can be surprising that some tags need [filter] and others don't.

The question, as I see it, is how to apply the principle of least surprise ... allowing the WML through when the intent is clear and disallowing obviously wrong constructs.

jostephd:
I support the request, to treat `[filter][filter]` the same as `[filter][and]`. It will make it so it's _always_ valid to use an explicit `[filter]` tag. This will make it easier to write WML code.

I also agree that if we do _not_ make this change, then we should make `[kill] [filter]` a warning or an error, since what it will do (nothing, [filter] will be ignored) isn't what the WML author intended.

[...]

PR #7717, opened 2023-06-14
fail if [kill] contains [filter], re #4316

ProditorMagnus:


soliton-:
The question is how useful it is to add arbitrary amounts of such checks everywhere. This should be caught by validating the WML, no?

Additionally we could support [filter] everywhere where it is inlined now and avoid the confusion altogether? Inlining the filter is a bad design anyway since it leads to key name collisions and it'd be nice to move away from this.

ProditorMagnus:
I agree, but I am not able to provide full solution for #4316 and it has been long enough that I felt even partial solution would help.

gfgtdf:
Alternatively we could just allow filter there, then we'd have some line `local filter = wml.get_child(cfg, "filter) or cfg`and use the variable filter later.

CelticMinstrel:
> This should be caught by validating the WML, no?

There is no expectation that add-on WML will be validated.

stevecotton:
Given that an add-on might already have been tested and rebalanced without spotting the bug, would it be better to show a warning instead of a hard error? Possibly not, I'm just wondering aloud.

ProditorMagnus:
I followed the idea of [modify_unit] - there [filter] is mandatory, even if I want to change all units.

cooljeanius:
I can confirm that I've been bitten by the problem that this PR is addressing before, although I wonder if this is the correct place to address it... maybe a `wmllint` check instead?

ProditorMagnus:
I believe previous answer `There is no expectation that add-on WML will be validated` fits wmllint too.

This PR is not the greatest solution, but I believe it still brings value.

CelticMinstrel:
I think putting it here is definitely better than putting it in wmllint or the schema.

Perhaps an even better place would be in the filter parsers (in C++) – if they see a `[filter]` tag (or, perhaps, any tag whose name begins with "filter") they should complain that it's invalid. That would mean you automatically get the warning for _any_ tag that uses an embedded filter if you add any filter tag to it.

cooljeanius:
> I think putting it here is definitely better than putting it in wmllint or the schema.
> 

All right, how about if we say "in addition to" instead of "instead"?

data/lua/wml/kill.lua
```diff
@@ -3,6 +3,10 @@ local location_set = wesnoth.require "location_set"
 local kill_recursion_preventer = location_set.create()
 
 function wesnoth.wml_actions.kill(cfg)
+	if wml.get_child(cfg, "filter") then
+		wml.error "do not use [filter] in [kill]"
```
stevecotton, line 7: The writing style doesn't match similar error messages that are already in the code. Suggestions:

* "[kill] has unsupported [filter] subtag"
* "[filter] tag found inside [kill]"

ProditorMagnus: `"[kill] has unsupported [filter] subtag"` looks nicer.

CelticMinstrel:
> I think putting it here is definitely better than putting it in wmllint or the schema.
> 
> Perhaps an even better place would be in the filter parsers (in C++) – if they see a `[filter]` tag (or, perhaps, any [i]unknown[/i] tag whose name begins with "filter") they should complain that it's invalid. That would mean you automatically get the warning for _any_ tag that uses an embedded filter if you add any filter tag to it. The only potential downside to that is that I'm not sure if that'll give you the identity of the tag that add the incorrect filter in it… I guess it would probably show up in a stack trace if nothing else?

data/lua/wml/kill.lua
ProditorMagnus: schema uses 20230713 19:25:15 error validation: Tag [event] may not be used in [effect]

stevecotton: That sounds good to me too.

stevecotton (approved):

Issue #7718, opened 2023-06-14
WC: Improve dialogue when units join


stevecotton:
### Language

English

### Source

data/campaigns/World_Conquest/lua/game_mechanics/unittypedata.lua

### Original Text

When you find a unit in World Conquest, there's text based on the unit type. #7716 annotates these with po comments, and notes a few British slang phrases.

A few stood out:

### Drake Fighter

"The ancient spirits tell me my destiny lies with yours. My sword is at your command." is said by a Drake Fighter, with a war blade.

### Ulf

The text about finding an Ulf appears to be based on Native American phrases, possibly clear to `en_US` speakers. As a native `en_GB` speaker, I missed the references, and it came across as a jibe about the dwarf's height. If it's meant to be a reference, then the logic seems reversed - why would another dwarf use the generic form, but a human feel comfortable guessing a name?

* "Chin up, lads. Today is a good day to die!",
* If the speaker is another dwarf: "Brave words. Welcome to The Fight Club..."
* If the speaker is human: "Brave words. Welcome little big man."

### Horseman

This one's more a sanity check, but I just got a "is that a bit sexist?" feel when looking at it. If a female unit discovers a Cavalryman or Horseman, they respond to the rider with, "Of course. How could a girl say no to a man riding a horse?"

### Proposed Change

The Drake Fighter should say "blade". Not sure about the other ones.

Wedge009:
Who wrote all this originally? I also had no idea about the alternate meaning of 'brave words' - I took the phrase literally.

And I suppose the thief's cockney accent is deliberate, I think it's stereotypical enough that it would make sense to en_US readers, but I'm not sure (since some notable British works have had to be 'localised' for American audiences).

I don't care to get too hung up on identity debates. And the text seems to be rather casual in nature. But given the medieval context of Wesnoth's setting, perhaps the cavalryman could be rephrased as a knight submitting himself in service to a lady.

gfgtdf:
> Who wrote all this originally? I

I don't know but the wc2 credits say: `Prose, Grammatical and Text Assistance: TL, Natasiel, Rigor, tekelili` so so far our best  guess is one of those

stevecotton:
It was already present in "World Conquest tek" 0.5.3.2.2, on the 1.10 add-ons server. Couldn't find it elsewhere.

stevecotton:
[...]

PR #7722, opened 2023-06-17
WC: change some messages when units join, and add po hints

stevecotton:
Add to 1.17 the po hints that #7716 adds to 1.16, and fix the issues of #7718.

Drake Fighters now say "blade" instead of "sword".

Remove the response when a human finds an Ulf, because it didn't sound right as the first thing a human says to a dwarf. The Ulf’s "a good day to die" could be read as a Native American battlecry, and "brave words" as "words of an Indian Brave", leaving the "little big man" to sound like a name.

Without catching the culture reference, it came across as a jibe about the dwarf's height.

Catching the culture reference, the logic seems reversed. Why would another dwarf use the generic form, but a human feel comfortable saying that?


stevecotton:
@Hejnewar, are you happy for this to merge? I've added you as reviewed because WC is a multiplayer campaign.

Issue #7709, opened 2023-06-11
[1.17] A status change is not immediately visible in the sprite


sevu:
Changing a unit's status often comes with a graphical change. Currently on 1.17:
1) Change a unit's status with `:unit status=petrified` (or poisoned).
2) _Unit sprite is not yet gray_, standing animations still play (e.g. for the Jinn unit)
3) Click the unit. Then the graphics are updated.

ProditorMagnus:
Debug-only or normal status change too?

gfgtdf:
Is this new in 1.17?

sevu:
Debug-only.
New in 1.17 because there was the switch to hardware acceleration.

Wedge009:
I could not replicate this in current master, 9e61449562a. I also had no such issue with 1.16.9 release, 1.17.6 tag, 1.17.7 tag. Tested on Linux/X11. Tested units with standing animations and those without.

Was there a specific unit you encountered this with?

Wedge009:
Okay, here is what I can confirm: the unit shading changes from neutral to green for poisoned or grey for petrified immediately upon the debug-command. But for petrified only, the only issue I see is the unit animations still playing, which obviously doesn't make sense for a petrified unit. The issue subject implied that the unit colour shading was not showing, but it seems like that's not the actual issue - is that a correct assessment?

Wedge009:
Clarifying the animations, some units like the loyalist halberdier and orc archer, have standing animations that show intermittently. Introduced in 1.17 were 'breeze' animations, showing small things like the tail on the orcish grunt's helmet moving in the breeze. The latter, I observe still playing after debug-petrification. The former, I have not.

doofus-01:
Some of this might be related to #7551.  I'll look into it.

doofus-01:
Looked into it with recent master, and:
 - The Jinn was an oversight on my part, easy to fix, need to check if there is anything else where it matters more (i.e. can be poisoned)
 - The idle animations (belching ogre, for example) don't play when the units are petrified.
 - The standing animations (bobbing skeleton, for example) play for a petrified unit that was made that way via debug, then get no interaction to refresh, and it stops eventually.  That seems pretty niche and low priority, does it happen in a real use case, like a `[petrify]` direct action?

Wedge009:
@sevu Can you review master now that #7725 has been merged?

doofus-01:
[...]

PR #7725, opened 2023-06-17
1.17 - add primary key to some unit animations that need it

doofus-01:
Some units still didn't change color when they were supposed to (poisoned or petrified*, anything else?)
 - Jinn 
 - Wild Wyvern
 - WC & Soulless Falcon

I think this is it, but if anyone finds something, I can add it to the list.

partially deals with #7709 

[*] via debug, they do turn gray under most circumstances.


Wedge009:
Can confirm the Jinn's colour changes as expected for petrification and poisoning. As you explain in #7709 the animation is probably due to the debug method of changing status - I assume if the petrification is through normal game-play then there wouldn't be standing animation playing.

PR #7730, opened 2023-06-20
1.17 Elf fighter sprite updates

doofus-01:
#7651 updated the elvish fighter with what had been overlooked for years, and added a running animation.  There was some question about the shoulders and chest, which I agreed were a bit too much, so I am proposing a slight revision.  

The running animation is great, and can still be used after the update, but the basic combat animations are too similar to the old frames (as was intended at the time; the goalposts have moved).  So I am updating those, though for now it is just melee frames.

At the end of this, it will be subjected to #6417


MechanicalDragon963:
I don't think you should have changed Vyncyn's version, it looks better. Why are you changing the good sprites instead of working on the really problematic ones?

knyghtmare:
Screenshot 2023-06-20 at 4 07 19 PM

Comparing both, I do think Vyncyn's one does look better. This "revised" one, I do not like. It's overly generic.

> instead of working on the really problematic ones

True, The Troll Warrior and Great Troll have the same sprite for years now... and Javelineer has been left in the void.

Hejnewar:
I will defend it. It is actually more clear and not so blurry as the old one, it is also simpler which most likely improves animation prospects and I would rather have simple animated sprites than complex and impossible to animate ones, aslo simplicity decreases entry level so Im all for simple sprites. The shoulder pads also fit better.

Mirion147:
I do think there is merit to the point of this not needing to be improved, especially after getting an improved version already, but more specifically in reference to the sprite, i don't necessarily think it's bad. But i do think it's overly bright, if you went with the same dark green i could probably live with it.

Not sure what the priority list is looking like though if this is getting worked on twice vs other items that are just using placeholder sprites

doofus-01:
Those complaining about this getting "worked over twice" can just go away; you are framing this in the worst possible way and who are you that you get to dictate the workflow?  It's been a collaboration of volunteers to get here, it will continue to be a collaboration of volunteers.  That's not an invitation to debate.

Hejnewar is right that animations were a big part of the motivation (as I said in the OP).  Any color issues can be easily be addressed later, that's the point of #6417.

Mirion147:
> Those complaining about this getting "worked over twice" can just go away...

Glad to hear how accepting you are of community feedback! Thank you for leading this community with an iron fist 😅

PR #7731, opened 2023-06-20
LoW/ANL/WC update instances of `units/elves-wood/fighter` to account extra introduced DIR

knyghtmare:
Fixes this:
image


soliton-:
The image is also used in WC and ANL.

soliton-:
Yeah, it should be added to wmllint.

PR #7732, opened 2023-06-20
[Wmllint] add a new find and replace

knyghtmare:
for the Elven Fighter unit


data/tools/wmllint
```diff
@@ -1288,6 +1288,8 @@ linechanges = (
         ("portraits/monsters/sea-serpent.png", "portraits/monsters/sea-serpent.webp"),
         ("portraits/monsters/big-icemonax.png", "portraits/monsters/big-icemonax.webp"),
         ("portraits/monsters/ogre.png", "portraits/monsters/ogre.webp"),
+        # introduced in wesnoth 1.17.18
```
soliton-, line 1291: I think this is a little confusing since it's not in 1.17.18. Should either say `1.17.19` or `after`.

knyghtmare: image
Oops...I see that it is x.x.19

PR #7736, opened 2023-06-27
Fix a writing in matches_ability_filter function

newfrenchy83:
error don't affect functionality, but it is for writing quality.


stevecotton:
That's nice and simple.

PR #7738, opened 2023-06-27
Add [test_do_attack_by_id], a wrapper for [do_command][attack]

stevecotton:
Just a draft for the moment, as several different improvements have ended up intertwined with each other. If anyone wants to use one of these on its own, please shout, because I feel this PR is going to be in draft for about a month (other things in life are currently higher priority).

The new tag takes ids instead of needing coordinates. So instead of storing units or hardcoding values, an attack can be done like this:

```
[test_do_attack_by_id]
	attacker=alice
	defender=bob
	weapon=0
[/test_do_attack_by_id]
```

Add a simpler unit for unit tests than the Elvish Archer / Orcish Grunt combo, cherry-picked from gfgtdf's PR #7589. This is a unit with a 10x1 melee attack, and 0% defense on all terrains.

Edit on 29th June: this now includes a test for plague abilities, to see if a plague weapon special always takes priority over a taught plague ability.

### DIFF
--- a/data/test/lua/wml_tags.lua
+++ b/data/test/lua/wml_tags.lua
@@ -1,0 +1,48 @@
+--! #textdomain wesnoth-test
+
+local T = wml.tag
+
+---Wrapper for [do_command][attack] that takes unit ids instead of map coordinates,
+---as tests are likely to already know the id.
+---
+---cfg.attacker,cfg.defender are unit ids, and map to [attack]source,destination=
+---cfg.weapon,cfg.defender_weapon are optional, and map to [attack]weapon,defender_weapon=
+---
+---cfg.resupply_attacks_left is an optional int, if set then it will ensure the attacker has
+---at least that many attacks_left before calling [do_command][attack].
+function wesnoth.wml_actions.test_do_attack_by_id(cfg)
+	-- can't be called "source", otherwise wmllint wants to add translation marks
+	if not cfg.attacker then
+		wml.error("[test_do_attack] missing required attacker= attribute")
+	end
+	local attacker = wesnoth.units.find_on_map { id = cfg.attacker }[1]
+	if not attacker or not attacker.valid then
+		wml.error("[test_do_attack] attacker did not match a unit")
+	end
+
+	if not cfg.defender then
+		wml.error("[test_do_attack] missing required defender= attribute")
+	end
+	local defender = wesnoth.units.find_on_map { id = cfg.defender }[1]
+	if not defender or not defender.valid then
+		wml.error("[test_do_attack] defender did not match a unit")
+	end
+
+	local weapon = cfg.weapon or ""
+	local defender_weapon = cfg.defender_weapon or ""
+
+	-- Avoid needing to modify units to give them multiple attacks per round
+	-- This attribute is a number rather than a boolean, to support [attack]attacks_used=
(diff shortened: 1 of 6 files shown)


stevecotton:
@ProditorMagnus for tests it seemed likely that the author will always know the exact ID, so having `[test_do_attack[` take unit filters seemed longwinded. The alternative code for that was:

```
local T = wml.tag

function wesnoth.wml_actions.test_do_attack(cfg)
	local filter = wml.get_child(cfg, "source") or wml.error("[test_do_attack] missing required [source] tag")
	local attacker = wesnoth.units.find_on_map(filter)[1]
	if not attacker or not attacker.valid then
		wml.error("[test_do_attack] filter for [source] did not match a unit")
	end

	filter = wml.get_child(cfg, "destination") or wml.error("[test_do_attack] missing required [destination] tag")
	local defender = wesnoth.units.find_on_map(filter)[1]
	if not defender or not defender.valid then
		wml.error("[test_do_attack] filter for [destination] did not match a unit")
	end

	local weapon = cfg.weapon or ""
	local defender_weapon = cfg.defender_weapon or ""

	wesnoth.wml_actions.do_command {
		T.attack {
			T.source { x = attacker.x, y = attacker.y },
			T.destination { x = defender.x, y = defender.y },
			weapon = weapon,
			defender_weapon = defender_weapon,
		}
	}
end
```

data/test/lua/wml_tags.lua
```diff
@@ -0,0 +1,33 @@
+--! #textdoman wesnoth-test
+
+local T = wml.tag
+
+function wesnoth.wml_actions.test_do_attack_by_id(cfg)
+	if not cfg.source then
```
stevecotton, line 6: This attribute has to be renamed, because wmllint expects `source` to have a translation mark.

stevecotton:
The just-pushed commit with a new test probably determines how I'll split this into sensibly-sized PRs. One of the usecases for `[test_do_attack_by_id]` is to simplify writing a unit test for #7484's `[plague]priority=` attribute, and that unit test will be big enough that it wants its own PR. There's likely to be a without-priority version of that test, to ensure that plague doesn't cause OOS when no priorities are defined (so checking that it's deterministic based on the units' locations).

If anyone wants to grab the new unit into their own PRs then that would be welcome too.

soliton-:
I would avoid creating new WML specifically for tests. Rather create new WML that is generally available and useful and use that in tests.

stevecotton:
If it was more generally available, I think it would feature-creep to something that isn't useful.

Part of the simplicity is that this handles errors by failing the test, or by doing nothing and having the test fail later. Outside unit tests, there's a possibility of there being no unit with that id, or that the two units aren't adjacent to each other. UMC could safely use it for "spawn a unit next to an explorer, and attack immediately" and "an explorer moved adjacent to a guardian, attack immediately", but I think all other cases would need UMC to store units and do sanity checks first. Even for the UMC cases I've identified, it's likely that the event is already auto-storing `$unit`, and that the event either auto-stores `$second_unit` or already includes a hardcoded `x,y=` filter.

A general tag would likely take SUFs instead of ids, which adds the error-handling of multiple units matching the filter too.

Pentarctagon:
We also do already have test-specific code (ie: [test], lua's `unit_test` module).

PR #7742, opened 2023-06-29
Fix bug in overwrite_specials

newfrenchy83:
I forget what abilities is the only who can use 'overwrite_specials' and what'filter_student' must ever checked instead of 'filter_self' in reference list


stevecotton:
Please could you give an example of how to trigger the bug? As in, what would be a case where it would give the wrong result.

newfrenchy83:
in fact 'writing error' will be more pertinent what bug, but it can imagine what ability weapon with [filter_self] is checked with search to weapon filter_self instead of "filter_student", it could return an false check

stevecotton:
I feel that it should have better documentation, but it makes sense to merge as is, and then I can write docs and open a new PR.

Issue #7009, opened 2022-09-14
[modify_unit_type] check if the advancement is already possible for the [unit_type]


sevu:
### Describe the desired feature

`[modify_unit_type]add_advancement=` adds the advancement to the list.
If it was already in the list, it's added twice.
This happens for example when both a [modification] and an [era] include the macro to unlock an advancement.
It causes problems with the mod in mainline.

![double](https://user-images.githubusercontent.com/21158813/190189994-8242a342-aee6-4ec9-8d97-d538fca7c83f.png)

the-nostromo-1:
Hi, is this issue still available to be worked on?

gfgtdf:
I'd think so, at least I'm not aware of this being fixed.

the-nostromo-1:
Awesome, thank you for the reply. I'm new to contributing to projects, do you have any advice where/what file to start with for this issue?

gfgtdf:
This depends on your Preknowledge, if i were to to it it just do a text search for `modify_unit_type` or `add_advancement` in the .cpp files and then trying to understand these programmcodes and how they can be changed

CelticMinstrel:
Before you get started on the issue I recommend building Wesnoth from source, if you haven't already.

My recommendation for finding where in the code this is handled would be to search the project (using your IDE of choice, or grep) for the string `"modify_unit_type"` (including the quotes), which will immediately show you where it's read from the config. From there you should be able to figure out where to go next.

the-nostromo-1:
Ok, sounds good! Thank you guys for the advice, I really appreciate it!

gfgtdf:
Fixed in #7743

PR #7743, opened 2023-06-29
Prevent duplicate advancements from being added to unit type

nullmoose:
Fixes issue #7009 

When adding an advancement to a unit type, check that the advancement is already present in `advances_to_`. If the advancement does already exist, it will not be added.


Pentarctagon:
Adding a test to https://github.com/wesnoth/wesnoth/tree/master/data/test/scenarios/wml_tests/ScenarioWML would be good.

nullmoose:
@Pentarctagon Ah thank you, I was wondering where the most appropriate place to put a test would be. Looking at the file structure, would it make more sense to put the test under `UnitsWML` rather than `ScenarioWML`?

Pentarctagon:
At least from the wiki, it looks like it would go under [scenario].

Pentarctagon:
Also maybe `std::find` would be better suited here?

CelticMinstrel:
> @Pentarctagon Ah thank you, I was wondering where the most appropriate place to put a test would be. Looking at the file structure, would it make more sense to put the test under `UnitsWML` rather than `ScenarioWM`

I believe the idea was mostly to choose the file based on which wiki page the feature is documented on – is that correct, Pentarctagon? The `[modify_unit_type]` tag is documented on [AddonsWML](https://wiki.wesnoth.org/AddonsWML). `ScenarioWML` is pretty close, as a scenario is a "subclass" of an addon component.

Pentarctagon:
It was, but AddonsWML isn't an especially descriptive name for a set of unit tests - all WML can be used in add-ons after all. So I figured that ScenarioWML would make the most sense since that's where the tag would be used in the test.

src/units/types.cpp
```diff
@@ -1367,7 +1367,9 @@ void unit_type::apply_scenario_fix(const config& cfg)
 	}
 	if(auto attr = cfg.get("add_advancement")) {
 		for(const auto& str : utils::split(attr->str())) {
-			advances_to_.push_back(str);
+			if(std::none_of(advances_to_.begin(), advances_to_.end(), translation::compare(str))) {
```
ProditorMagnus, line 1370: unit type is not translatable, is `translation::compare` correct?

CelticMinstrel:
> It was, but AddonsWML isn't an especially descriptive name for a set of unit tests - all WML can be used in add-ons after all. So I figured that ScenarioWML would make the most sense since that's where the tag would be used in the test.

I think it would be used more in CampaignWML, but there's not CampaignWML folder either. ScenarioWML is suitable in any case.

data/test/scenarios/wml_tests/ScenarioWML/UnitTypeWML/add_advancements.cfg
```diff
@@ -0,0 +1,64 @@
+#####
+# API(s) being tested: [modify_unit_type][add_advancement]
+##
+# Actions:
+# Create an Advancing Unit unit type with an Advanced Unit advancement.
```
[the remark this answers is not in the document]
nullmoose: Ah dang it, my bad. I had updated the tests but had forgotten to update the descriptions afterward. I'll get that fixed.

CelticMinstrel:
1. It seems wrong for UnitTypeWML to be nested under ScenarioWML. You can't define a unit type in a scenario. UnitTypeWML, if it exists, should be a separate category beside ScenarioWML.
2. `[modify_unit_type]` is not really part of UnitTypeWML in the first place, so I don't think it makes sense to put it there. Sure, it may be an operation on unit types, but it doesn't go in the `[unit_type]` tag.

nullmoose:
@CelticMinstrel Got it. So just put add_advancements.cfg directly under ScenarioWML?

src/units/types.cpp
```diff
@@ -1367,7 +1367,9 @@ void unit_type::apply_scenario_fix(const config& cfg)
 	}
 	if(auto attr = cfg.get("add_advancement")) {
 		for(const auto& str : utils::split(attr->str())) {
-			advances_to_.push_back(str);
+			if(std::find(begin(advances_to_), end(advances_to_), str) == std::end(advances_to_)) {
```
gfgtdf, line 1370: i think we have an utility function `utils::contains` for this.

nullmoose: @gfgtdf Ah, sure enough. Updated to use `utils::contains`

gfgtdf (approved):
Looks good to me, but i'm not really an expert on the wml tests.

ProditorMagnus:
I disagree with creating unit to negative coordinates. It could be part of different test about how unit placements work, but for purpose of this test either create unit into variable, or to valid and nonoccupied x,y on map.

gfgtdf:
> I disagree with creating unit to negative coordinates

I agree (i didn't look at the tests), im not sure what the intended behavior is, i assume it goes to the recall list then? Or does it pick another location ?

nullmoose:
> I disagree with creating unit to negative coordinates. It could be part of different test about how unit placements work, but for purpose of this test either create unit into variable, or to valid and nonoccupied x,y on map.

Fair enough. That was definitely a little copy/paste from a different test. I can update that. 👍

Pentarctagon:
Any other objections to merging this?

ProditorMagnus (approved):
It is suitable test. While it also tests the order that unit types are listed in advances_to, that is how advancements currently work so can stay.

src/units/types.cpp
```diff
@@ -1367,7 +1367,9 @@ void unit_type::apply_scenario_fix(const config& cfg)
 	}
 	if(auto attr = cfg.get("add_advancement")) {
 		for(const auto& str : utils::split(attr->str())) {
-			advances_to_.push_back(str);
+			if(!utils::contains(advances_to_, str)) {
```
Vultraz, line 1370: Honestly would probably be better to make `advances_to_` a `set`.

ProditorMagnus: Probably, but that is not something to ask during this pr.

stevecotton:
@ProditorMagnus are the issues you raised all resolved?

ProditorMagnus:
Not fully but it is good enough to merge.

src/units/types.cpp
gfgtdf: Iirc the order is important though, since off-turn advancements always choose the first one. So changing it to set might change behavior

Pentarctagon: Having it always choose the first advancement when a unit advances off-turn is intentional.

CelticMinstrel: Right, but making it be a set means a designer can't control which is the default advancement in this case – it'll take the one that's alphabetically first. It's better to leave open that avenue of designer control, I think. So no set, just an array which is kept unique.

gfgtdf: > Having it always choose the first advancement when a unit advances off-turn is intentional.

Well yes, did something i said make it sound like that's not the case?

PR #7745, opened 2023-06-30
simplify checking for overwrite_specials when 'one_side' used

newfrenchy83:
Because complexity of special_active function, i was afraid to use directly special_affects_self/opponent, but like in this state, only apply to self/opponent/etc. is pertinent, i have finally try that.


stevecotton:
Looks good. Should I merge this, and then you rebase the other PRs on top?

newfrenchy83:
> Looks good. Should I merge this, and then you rebase the other PRs on top?

Yes, do it, please.

PR #7746, opened 2023-06-30
Add priority and filter to overwrite specials

newfrenchy83:
Since I add another feature to overwrite_specials in addition to filtering with the priority system, which obviates the need to limit its use to abilities used as weapons, I prefer to open this query to group the main commits that are related to it .

### DIFF
--- a/data/core/macros/utils.cfg
+++ b/data/core/macros/utils.cfg
@@ -552,6 +552,9 @@
                         cumulative=no
                         affect_self=yes
                         overwrite_specials=both_sides
+                        [overwrite]
+                            priority=1000
+                        [/overwrite]
                         [filter_opponent]
                             {SECOND_FILTER}
                         [/filter_opponent]
@@ -605,6 +608,9 @@
                         cumulative=no
                         affect_self=yes
                         overwrite_specials=both_sides
+                        [overwrite]
+                            priority=1000
+                        [/overwrite]
                         [filter_opponent]
                             {SECOND_FILTER}
                         [/filter_opponent]
--- a/data/schema/filters/abilities.cfg
+++ b/data/schema/filters/abilities.cfg
@@ -4,9 +4,9 @@
 	max=0
 	{SIMPLE_KEY id string_list}
 	{SIMPLE_KEY tag_name string_list}
-	{SIMPLE_KEY overwrite_specials string_list}
+	{SIMPLE_KEY overwrite_specials ability_overwrite}
 	{SIMPLE_KEY apply_to string_list}
-	{SIMPLE_KEY active_on string_list}
+	{SIMPLE_KEY active_on ability_context}
 	{SIMPLE_KEY value s_int_range_list}
 	{SIMPLE_KEY add s_int_range_list}
 	{SIMPLE_KEY sub s_int_range_list}
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
@@ -1,0 +1,106 @@
+#####
+# API(s) being tested: [damage]with overwrite_specials attibute and [overwrite][filter_specials]
+##
+# Actions:
+# Three [damage] abilities are added, one with value=5, other with add=11 and the last with multiply=2.
+# The one with multiply=2 have overwrite_specials=both_sides and [overwrite][filter_specials] for overwrite [damage] with add=11 but not value=5.
+# Have alice attack bob.
+##
+# Expected end state:
+# Alice has 100 - (5*2)=90 hitpoints because add=11 overwrited but not value=5 who replace 9 pt of damage of bob.
+#####
+{GENERIC_UNIT_TEST "test_overwrite_specials_filter" (
+    [event]
+        name=start
+        [modify_unit]
+            [filter]
+            [/filter]
+            max_hitpoints=100
+            hitpoints=100
+            attacks_left=1
+        [/modify_unit]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=new_ability
+                [abilities]
+                    [damage]
+                        id=test_dmg
+                        value=5
+                    [/damage]
+                    [attacks]
+                        id=test_att
+                        value=1
+                    [/attacks]
+                    [damage]
+                        id=test_overwrite
+                        add=11
+                    [/damage]
+                    [damage]
+                        id=test_overwrite_specials
+                        multiply=2
+                        overwrite_specials=both_sides
+                        [overwrite]
+                            [filter_specials]
+                                [not]
+                                    type_value=value
+                                [/not]
+                            [/filter_specials]
+                        [/overwrite]
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                    [/chance_to_hit]
+                [/abilities]
+            [/effect]
+            [filter]
+                id=bob
+            [/filter]
+        [/object]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+            kill=yes
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        [unstore_unit]
+            variable=a
+            find_vacant=yes
+            x,y=$b.x,$b.y
+        [/unstore_unit]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+
+        [do_command]
+            [attack]
+                weapon=0
+                defender_weapon=0
+                [source]
+                    x,y=$a.x,$a.y
+                [/source]
+                [destination]
+                    x,y=$b.x,$b.y
+                [/destination]
+            [/attack]
+        [/do_command]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+        {ASSERT ({VARIABLE_CONDITIONAL a.hitpoints equals 90})}
+        {SUCCEED}
+    [/event]
+)}
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1336,18 +1336,8 @@ static bool overwrite_special_affects(const config& special)
 
 unit_ability_list attack_type::overwrite_special_overwriter(unit_ability_list overwriters, const std::string& tag_name) const
 {
-	std::vector<double> priorityvec;
 	//remove element without overwrite_specials key, if list empty after check return empty list.
 	utils::erase_if(overwriters, [&](const unit_ability& i) {
-		if(overwrite_special_affects(*i.ability_cfg)){
-			auto overwrite_specials = (*i.ability_cfg).optional_child("overwrite");
-			if(overwrite_specials && !overwrite_specials["priority"].empty()){
-				auto ret = std::find(priorityvec.begin(), priorityvec.end(), overwrite_specials["priority"].to_double(0));
-				if(ret == priorityvec.end()){
-					priorityvec.push_back(overwrite_specials["priority"].to_double(0));
-				}
-			}
-		}
 		return (!overwrite_special_affects(*i.ability_cfg));
 	});
 
@@ -1355,41 +1345,23 @@ unit_ability_list attack_type::overwrite_special_overwriter(unit_ability_list ov
 		return overwriters;
 	}
 
-	//if vector no empty and list contain 2 or more elements, continue checking else return list.
-	if(!priorityvec.empty() && (overwriters.size())){
-		if(priorityvec.size() >= 2){
-			std::sort(priorityvec.begin(), priorityvec.end(),[](const double &l, const double &r)
-				{
-					return l > r;
-				}
-			);
-		}
-		for(double priority : priorityvec){
-			//erase elements with priority who matches conditions
-			utils::erase_if(overwriters, [&](const unit_ability& i) {
-				auto overwrite_specials = (*i.ability_cfg).optional_child("overwrite");
-				bool is_overwritable = false;
-				if(overwrite_specials && overwrite_specials["priority"].to_double(0) == priority){
-					is_overwritable = overwrite_special_checking(overwriters, *i.ability_cfg, tag_name);
-				}
-				return (is_overwritable);
-			});
-			if(overwriters.size() <= 1){
-				break;
+	if(overwriters.size() >= 2){
+		utils::sort_if(overwriters,[](const unit_ability& i, const unit_ability& j){
+			auto oi = (*i.ability_cfg).optional_child("overwrite");
+			double l = 0;
+			if(oi && !oi["priority"].empty()){
+				l = oi["priority"].to_double(0);
 			}
-		}
-		//remove elements without priority who matches conditions
-		//determined by priority elements remaining in list.
-		if(overwriters.size() > 1){
-			utils::erase_if(overwriters, [&](const unit_ability& i) {
-				auto overwrite_specials = (*i.ability_cfg).optional_child("overwrite");
-				bool is_overwritable = false;
-				if(!overwrite_specials || overwrite_specials["priority"].empty()){
-					is_overwritable = overwrite_special_checking(overwriters, *i.ability_cfg, tag_name);
-				}
-				return (is_overwritable);
-			});
-		}
+			auto oj = (*j.ability_cfg).optional_child("overwrite");
+			double r = 0;
+			if(oj && !oj["priority"].empty()){
+				r = oj["priority"].to_double(0);
+			}
+			return l > r;
+		});
+		utils::erase_if(overwriters, [&](const unit_ability& i) {
+			return (overwrite_special_checking(overwriters, *i.ability_cfg, tag_name));
+		});
 	}
 	return overwriters;
 }
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -1316,8 +1316,9 @@ unit_ability_list attack_type::get_specials_and_abilities(const std::string& spe
 	abil_list.append(get_weapon_ability(special));
 	unit_ability_list overwriters = overwrite_special_overwriter(abil_list, special);
 	if(!abil_list.empty() && !overwriters.empty()){
-		int list_count = 0;
-		overwrite_special_checking(abil_list, overwriters, special, list_count);
+		utils::erase_if(abil_list, [&](const unit_ability& j) {
+			return (overwrite_special_checking(overwriters, *j.ability_cfg, special));
+		});
 	}
 	return abil_list;
 }
@@ -1337,12 +1338,19 @@ unit_ability_list attack_type::overwrite_special_overwriter(unit_ability_list ov
 {
 	std::vector<double> priorityvec;
 	int list_count = 0;
+	//remove element without overwrite_specials key, if list empty after check
+	//return empty list.
+	//If element contains priority with overwrite_specials key valid,
+	//value added in vector.
 	utils::erase_if(overwriters, [&](const unit_ability& i) {
 		if(overwrite_special_affects(*i.ability_cfg)){
 			list_count += 1;
 			auto overwrite_specials = (*i.ability_cfg).optional_child("overwrite");
 			if(overwrite_specials && !overwrite_specials["priority"].empty()){
-				priorityvec.push_back(overwrite_specials["priority"].to_double(0));
+				auto ret = std::find(priorityvec.begin(), priorityvec.end(), overwrite_specials["priority"].to_double(0));
+				if(ret == priorityvec.end()){
+					priorityvec.push_back(overwrite_specials["priority"].to_double(0));
+				}
 			}
 		}
 		return (!overwrite_special_affects(*i.ability_cfg));
@@ -1352,6 +1360,7 @@ unit_ability_list attack_type::overwrite_special_overwriter(unit_ability_list ov
 		return overwriters;
 	}
 
+	//if vector no empty and list contain 2 or more elements, continue checking else return list.
 	if(!priorityvec.empty() && (list_count >= 2)){
 		if(priorityvec.size() >= 2){
 			std::sort(priorityvec.begin(), priorityvec.end(),[](const double &l, const double &r)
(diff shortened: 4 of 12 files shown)


newfrenchy83:
@stevecotton @CelticMinstrel could you review this PR, please?

src/units/abilities.cpp
```diff
@@ -1283,33 +1288,66 @@ static bool overwrite_special_affects(const config& special)
...
-					one_side_overwritable = special_active_impl(other_attack_, shared_from_this(), *j.ability_cfg, AFFECT_OTHER, ability, filter_self);
+				else if(special_affects_opponent(*i.ability_cfg, !is_attacker_)){
+					one_side_overwritable = special_affects_opponent(*j.ability_cfg, !is_attacker_);
+				}
+			}
+			bool special_matches = true;
+			if(overwrite_specials){
+				auto overwrite_filter = (*overwrite_specials).optional_child("filter_specials");
+				if(overwrite_filter && is_overwritable && one_side_overwritable){
+					special_matches = self_ ? (*self_).ability_matches_filter((*j.ability_cfg), ability, *overwrite_filter, true) : false;
```
stevecotton, line 1347: If #7745 merges, what happens to `ability` on this line?

newfrenchy83: i repull ability declaration in rebase because i need that, but must be removed in https://github.com/wesnoth/wesnoth/pull/7745 because don't used before what filter implemented

newfrenchy83: ability removed(in filter using was 'cosmetic' only)

data/schema/filters/abilities.cfg
```diff
@@ -19,3 +17,11 @@
 	{DEFAULT_KEY type_value value_type empty}
 	{FILTER_BOOLEAN_OPS abilities}
 [/tag]
+
+[tag]
+	name="$filter_abilities"
+	max=0
+	super="$filter_abilities_generic"
+	{SIMPLE_KEY tag_name string_list}
+	{SIMPLE_KEY overwrite_specials string_list}
```
stevecotton, line 26: Why does this take `string_list` instead of `ability_overwrite`?

src/units/attack_type.hpp
```diff
@@ -135,15 +135,23 @@ class attack_type : public std::enable_shared_from_this<attack_type>
...
 	// Configured as a bit field, in case that is useful.
 	enum AFFECTS { AFFECT_SELF=1, AFFECT_OTHER=2, AFFECT_EITHER=3 };
+	/**
+	 * Filter a list of abilities or weapon specials, removing any entries that don't own
+	 * the overwrite_specials attributes.
+	 *
+	 * @param overwriters list that may have overwrite_specials attributes.
+	 * @param priority value used below for checking of input list in overwrite_special_checking function.
+	 */
+	void overwrite_special_overwriter(unit_ability_list& overwriters, double& priority) const;
```
stevecotton, line 145: If this will only be called within abilities.cpp, it can be a static function in that file, instead of needing to add it to the header file.

src/units/abilities.cpp
```diff
@@ -1283,23 +1288,49 @@ static bool overwrite_special_affects(const config& special)
...
+			}
+		}
+		if(!priorityvec.empty()){
+			priority = *(std::max_element(priorityvec.begin(), priorityvec.end()));
+			for(unit_ability_list::iterator i = overwriters.begin(); i != overwriters.end();) {
+				auto overwrite_specials = (*i->ability_cfg).optional_child("overwrite");
+				if(overwrite_specials && overwrite_specials["priority"].to_double(0) == priority) {
+					++i;
+				} else {
+					i = overwriters.erase(i);
```
stevecotton, line 1318: How does `[overwrite]priority=` interact with `[overwrite][filter_specials]`?

IIUC, this will filter out based on priority first, and then `overwrite_special_checking` will check the filters afterwards. An overwriter with a high priority but whose filter doesn't match will still remove a lower-priority overwriter whose filter does match.

newfrenchy83: [filter_specials] is for remove abilities/specials who matches except the overwriter(don't remove itself), overwriter don't must be overwritable. When priority used, this the highest who is overwriter, else it is all ability who have overwrite_specials attributes

data/schema/filters/abilities.cfg
newfrenchy83: and why active_on take string_list then what only value must be"offense,defense,both"?

src/units/abilities.cpp
stevecotton: Sorry, I don't understand your comment. I'm not good at reading French, but I assume that's an easier language for you, and I'll attempt to read it if you write explanations in it.

I'm trying to work out whether the complexity is necessary, and the current example that I can come up with for having both `priority` and `[filter_specials]` would be:
1. The player's units have a debuff which has made their weapons blunt, changing any `pierce` or `blade` to `impact`.
2. A teacher has an ability to make adjacent units' `pierce`, `blade` and `impact` weapons do `fire` damage instead. This should overwrite the blunt debuff.
3. A teacher has an ability to make adjacent units' `melee` attacks do `arcane` damage instead. In this scenario, `arcane` is always better than `fire`, so this should overwrite both of the other changes.

A unit next to both teachers should therefore have arcane melee attacks and fire ranged attacks. How would that be implemented in this system?

It's a complex example, but the `[overwrite]` feature seems to have complexity that's only needed to support examples involving the interaction of three or more abilities.

newfrenchy83: > Sorry, I don't understand your comment. I'm not good at reading French, but I assume that's an easier language for you, and I'll attempt to read it if you write explanations in it.
> 
> I'm trying to work out whether the complexity is necessary, and the current example that I can come up with for having both `priority` and `[filter_specials]` would be:
> 
>     1. The player's units have a debuff which has made their weapons blunt, changing any `pierce` or `blade` to `impact`.
> 
>     2. A teacher has an ability to make adjacent units' `pierce`, `blade` and `impact` weapons do `fire` damage instead. This should overwrite the blunt debuff.
> 
>     3. A teacher has an ability to make adjacent units' `melee` attacks do `arcane` damage instead. In this scenario, `arcane` is always better than `fire`, so this should overwrite both of the other changes.
> 
> 
> A unit next to both teachers should therefore have arcane melee attacks and fire ranged attacks. How would that be implemented in this system?
> 
> It's a complex example, but the `[overwrite]` feature seems to have complexity that's only needed to support examples involving the interaction of three or more abilities.

First the blunting debuff, we're going to go with an 
```
[damage_type]
	id=impact_damage
	type=impact
	[filter_self]
		[filter_weapon]
			type=blade
			[or]
				type=pierce
			[/or]
		[/filter_weapon]
	[/filter_self]
[/damage_type]
```

2```
[damage_type]
	id=fire_damage
	type=fire
	[filter_student]
		[filter_weapon]
			type=blade
			[or]
				type=pierce
			[/or]
		[/filter_weapon]
	[/filter_student]
	overwrite_specials=one_side
	[overwrite]
		[filter_specials]
			id=impact_damage
		[/filter_specials]
	[overwrite]
[/damage_type]
```

3```
[damage_type]
	id=melee_arcane
	type=arcane
	[filter_student]
		[filter_weapon]
			range=melee
		[/filter_weapon]
	[/filter_student]
	overwrite_specials=one_side
	[overwrite]
		priority=3
	[overwrite]
[/damage_type]
```

data/schema/filters/abilities.cfg
stevecotton: `active_on` should probably only take an `ability_context` too

newfrenchy83: in filter? Ok.

src/units/abilities.cpp
stevecotton: I'm still wondering if it does the right thing, and what the interaction between different `[overwrite]` things should be. Based on a build of c02eca63f78501f9a62849f1dadda41613d19a4e, here's what I'm seeing:

```
#####
# API(s) being tested: [overwrite]priority=,[overwrite][filter_specials]
##
# Actions:
# Give Bob an ability minimal_damage, which reduces his weapon damage to 1.
# Give Bob an ability double_damage, which multiplies the damage by 2 and also overwrites minimal_damage.
# Give Bob an ability seven_times_damage, which multiplies the damage by 7 and with a high priority would overwrite another ability which isn't part of this test.
##
# Expected end state:
# TODO: expectation is currently under discussion
#####
{COMMON_KEEP_A_B_UNIT_TEST "override_priority" (
    [event]
        name=start

        [modify_unit]
            [filter]
            [/filter]
            max_hitpoints=100
            hitpoints=100
            attacks_left=1
        [/modify_unit]

        [object]
            silent=yes
            [effect]
                apply_to=new_ability
                [abilities]
                    [chance_to_hit]
                        value=100
                    [/chance_to_hit]
                [/abilities]
            [/effect]
            [filter]
                id=alice
            [/filter]
        [/object]

        [object]
            silent=yes
            [effect]
                apply_to=new_ability
                [abilities]
                    [damage]
                        id=minimal_damage
                        value=1
                        overwrite_specials=one_side
                    [/damage]
                    [damage]
                        id=double_damage
                        multiply=2
                        overwrite_specials=one_side
                        [overwrite]
                            priority=2
                            [filter_specials]
                                id=minimal_damage
                            [/filter_specials]
                        [/overwrite]
                        [affect_adjacent]
                        [/affect_adjacent]
                    [/damage]
                [/abilities]
            [/effect]
            [filter]
                id=bob
            [/filter]
        [/object]

        # Bob's attack now does 18 damage. It was originally 9, and the double_damage ability overwrites the minimal_damage one.

        # TODO: automate the test. Currently this is a manual test, and the event breaks here so that the dev can look at the situation, and then end Alice's turn to see the next part.
    [/event]

    [event]
        name=side 2 turn

        [object]
            silent=yes
            [effect]
                apply_to=new_ability
                [abilities]
                    [damage]
                        id=seven_times_damage
                        multiply=7
                        overwrite_specials=one_side
                        [overwrite]
                            priority=3
                            [filter_specials]
                                # As this doesn't match double_damage, this isn't expected to overwrite double_damage's overwrite, but it does because only the highest priority overwrite is used.
                                id=not_used_in_this
                            [/filter_specials]
                        [/overwrite]
                        [affect_adjacent]
                        [/affect_adjacent]
                    [/damage]
                [/abilities]
            [/effect]
            [filter]
                id=bob
            [/filter]
        [/object]

        # Bob's attack now does 14 damage. Even though double_damage is still active, it no longer overwrites minimal_damage because seven_times_damage has a higher priority overwrite.

        #{SUCCEED}
    [/event]
)}
```

newfrenchy83: Yes, i encoded like this, must be change?

stevecotton: It's a question of what the purpose is, what the design goal is. Which UMC abilities need it to work in which way, and which UMC abilities would need to work in a different way?

A review needs to be not just about the code, but about the design and requirements. Here I'm asking what the requirements are, what is it that this code needs to support. When asking myself "is the value that this adds worth the complexity that it adds", the "value" comes entirely from the parts that the players will see, the abilities in mainline and UMC that can be implemented using this.

newfrenchy83: add redesign and fix highest priority overwrite without discrimination in https://github.com/wesnoth/wesnoth/pull/7746/commits/6190925da10a8d590703675242daa9cbe611e6dd after rebased

stevecotton: I really need an answer to "what is the value that this gives to players", rather than updating code. At the moment, the three requirements for priority and overwrite that I understand are:
* Multiple abilities or weapon specials providing `plague`, but spawning different types of zombies
* Multiple abilities or weapon specials changing the chance-to-hit
* Multiple abilities or weapon specials changing the damage type (I know the other PR is needed too for that, but my question in this PR is "what needs the priority and filters". https://r.wesnoth.org/t56763 says why we need to change damage types via abilities, but to me it doesn't explain the need for the filters)

So I'm interested in working out what the simplest implementation is which can cover those three areas.

I'm not good at reading French, but I assume that's an easier language for you, and I'll attempt to read it if you write explanations in it.

newfrenchy83: and here is the last case of damage_type, this time it is a different case of plague because it exploits the fact that damage_type returns two character strings and not only one like for plague:
There are three damage_types that are exerted on a unit, a 'blunting' damage_type that affects blades and piercing attacks and coupled with a special damage that reduces the power of these worn weapons (the latter will not be represented in the example) , another adds the fire type to all the weapons of the blade or pierce type units and the last which adds the arcane type to the blade or piercing attacks of melee and which crushes the previous one in this case, but spares the first damage_type which turns blade or piercing into impact. The first is encoded as an ability to justify the presence of the filter (in specials it suffices to implement it only in the attacks concerned)

```
#define WEAPON_SPECIAL_IMPACT_TYPE
     [damage_type]
         id=impact
         name= _ "impact"
         type=impact
         [filter_student]
             [filter_weapon]
                 type=blade,pierce
             [/filter_weapon]
         [/filter_student]
     [/damage_type]
#enddef

#define WEAPON_SPECIAL_FIRE_SECOND_TYPE
     [damage_type]
         id=fire_sec
         name= _ "fire"
         type=fire
         cumulative=yes
         overwrite_specials=both_sides
         [overwrite]
             priority=1
             [filter_specials]
                 cumulative=yes
             [/filter_specials]
         [/overwrite]
         [filter_student]
             [filter_weapon]
                 type=blade,pierce
             [/filter_weapon]
         [/filter_student]
     [/damage_type]
#enddef

#define WEAPON_SPECIAL_ARCANE_SECOND_TYPE
     [damage_type]
         id=arcane_sec
         name= _ "arcane"
         type=arcane
         cumulative=yes
         overwrite_specials=both_sides
         [overwrite]
             priority=2
             [filter_specials]
                 cumulative=yes
             [/filter_specials]
         [/overwrite]
         [filter_student]
             [filter_weapon]
                 type=blade,pierce
                 range=melee
             [/filter_weapon]
         [/filter_student]
     [/damage_type]
#enddef

#define WEAPON_SPECIAL_COLD_SECOND_TYPE
     [damage_type]
         id=cold_sec
         name= _ "cold"
         type=cold
         cumulative=yes
         [filter_student]
             [filter_weapon]
                 type=blade
             [/filter_weapon]
         [/filter_student]
     [/damage_type]
#enddef
```

To these three types is added a damage_type which adds cold to sharp weapons but which is given priority by the two previous ones

src/units/abilities.cpp
```diff
@@ -1258,16 +1258,19 @@ unit_ability_list attack_type::get_weapon_ability(const std::string& ability) co
...
 	unit_ability_list spe_list = get_specials(special);
-	if(!spe_list.empty()){
-		spe_list = overwrite_special_checking(spe_list, abil_list, true);
-		if(special == "plague" && !spe_list.empty()){
+	abil_list.append(spe_list);
+	unit_ability_list overwriters = overwrite_special_overwriter(abil_list);
+	if(!abil_list.empty() && !overwriters.empty()){
+		abil_list = overwrite_special_checking(abil_list, overwriters);
+	}
+	if(special == "plague" && !spe_list.empty()){
```
stevecotton, line 1267: I've tried simplifying this function, and it passes the unit tests but I'm not sure if I'm missing something.

IIUC, some callers only look at the first item on the list, and this special-case for `plague` is to ensure that a weapon special will be returned before an ability used as a weapon special. I think the entire special case code can be removed by calling `get_specials()` before `get_weapon_ability()`:

```
unit_ability_list attack_type::get_specials_and_abilities(const std::string& special) const
{
	unit_ability_list abil_list = get_specials(special);
	abil_list.append(get_weapon_ability(special));
	unit_ability_list overwriters = overwrite_special_overwriter(abil_list);
	if(!abil_list.empty() && !overwriters.empty()){
		abil_list = overwrite_special_checking(abil_list, overwriters);
	}
	return abil_list;
}
```

newfrenchy83: Ok, i will try that

stevecotton:
Something that I'm confused about: could `attack_type` make `get_specials` and `get_weapon_ability` private, leaving only the `get_specials_and_abilities` function public?

At the moment, I'm looking at the only other caller of `get_specials`, which is attack.cpp's constructor of `battle_context_unit_stats`, and wondering why it wouldn't want the full list that comes from `get_specials_and_abilities` instead. So far I've only checked that the unit tests pass after making that change, I haven't worked out how to test whether the behavior is better or worse.

gfgtdf:
>> At the moment, I'm looking at the only other caller of get_specials, which is attack.cpp's constructor of battle_context_unit_stats, and wondering why it wouldn't want the full list that comes from get_specials_and_abilities instead. So far I've only checked that the unit tests pass after making that change, I haven't worked out how to test whether the behavior is better or worse.

i stumbled over this in my (ongoing) ability refactor aswell so: It doen't matter in that place, that's the code that checks weapon specials when simulating combats between different unit types (instead of actual units) and that codepath simply cannot handle combat abilities (and its probably quite difficult (and not worth) implementing since it would require some way to evaluate standard unit filters ) in the unit types (instead of actual units). so these two functions always return the same list.

newfrenchy83:
> Something that I'm confused about: could `attack_type` make `get_specials` and `get_weapon_ability` private, leaving only the `get_specials_and_abilities` function public?
> 
> At the moment, I'm looking at the only other caller of `get_specials`, which is attack.cpp's constructor of `battle_context_unit_stats`, and wondering why it wouldn't want the full list that comes from `get_specials_and_abilities` instead. So far I've only checked that the unit tests pass after making that change, I haven't worked out how to test whether the behavior is better or worse.

Me neither, I don't like the idea of merging the constitution of the two lists in a single function used for checking unit and unit_type because in this last case only the possession by unit_type can be checked (in the best case) possibly doing the difference between affect_self=yes(ability present) and affect_adjacent(absent if affect_self=no) but without considering that location information does not exist. Same for appy_to=self/opponent since the information of a hypothetical opponent is non-existent.

newfrenchy83:
> Something that I'm confused about: could `attack_type` make `get_specials` and `get_weapon_ability` private, leaving only the `get_specials_and_abilities` function public?
> 
> At the moment, I'm looking at the only other caller of `get_specials`, which is attack.cpp's constructor of `battle_context_unit_stats`, and wondering why it wouldn't want the full list that comes from `get_specials_and_abilities` instead. So far I've only checked that the unit tests pass after making that change, I haven't worked out how to test whether the behavior is better or worse.

Anyway in get_specials() if other_attack_ does not exist, the list is returned immediately after the check of self_attack which implies that the code for the list of [specials] and that of [abilities] must be constituted in two functions well distinct before merging them so even if I come to use get_specials_and_abilities for the unit_type check I should still keep get_specials and get_weapon_ability to constitute each list independently, and the fact that the presence of other_attack_ in get_weapon_ability is not checked is for me an extra argument to keep get_specials in this simulation. Add to this that it is the presence of self_ and other_ that is checked before each step; and the get_weapon_ability list will always be empty, so don't even bother to use get_specials_and_abilities instead of get_specials because it won't have any effect.

newfrenchy83:
@stevecotton @gfgtdf And I remind you that get_weapon_ability() is used independently in 'udisplay.cpp'

data/schema/filters/abilities.cfg
CelticMinstrel: Just define a list type based on `ability_context`: `{LIST_TYPE ability_context}`

That defines a new `ability_context_list` type.

If there's no need for the non-list version of the type, you can define a list that's not based on an existing type with `LIST_TYPE_COMPLEX`:

```
{LIST_TYPE_COMPLEX ability_context
[element]
# definition of the list element goes here
[/element]
}```

Comment (newfrenchy83) [reply data/schema/filters/abilities.cfg]: LIST_TYPE_COMPLEX seems to me more suitable for filtering ranges of numbers than for a list of characters (both, offense and defense for active_on)

**Comment by newfrenchy83** [comment]
@Pentarctagon here some explanation for this PR:

first of all, a quick reminder of what overwrite specials is.

this attribute is used so that an ability used as a special weapon can overwrite weapons/abilities of the same type (a macro like FORCE_CHANCE_TO_HIT in mainline thus continues to function despite the presence of an ability of higher value, or conversely, a a developer using a charge type [damage] with value multiply=2.5 can overwrite the charge used by a mainline unit without removing the special weapon unit code), but this approach has two defaults which are corrected here:

first the fact that these abilities with overwrite_specials can never be overwritten, for this I add in an [overwrite] tag the priority attribute, if one or more abilities from the list have overwrite_specials with priority the one which has the highest priority high will crush the others..

secondly, the introduction of a [filter_specials] in [overwrite] so that the overwriting is done in a targeted manner, if we take the second example, we can thus overwrite only the [damage] having the characteristics of 'charge '(active_on=offense, apply_to=both, value_type=multiply).

now let's move on to the code, first of all is done in the function overwrite_special_overwriter the elimination of any ability not having overwrite_specials, and at the same time, check that those having the attribute also have priority, if this is the case each priority will be added to a vector then once this is completed, classified in descending order.

if the vector is not empty then we go to step two for each value of the vector each ability having priority is added to a temporary list which will serve as a reference for the use of overwrite_special_checking in order to remove any overwrite from the list lower priority ability which corresponds to the reference filters. The process is repeated for each vector value until the end of the vector is reached or the list is reduced to one element.

once overwriter is created, this serves as a reference for a second use of overwrite_special_checking, each element of overwrite will be used as a reference for the filtering of abil_list which will finally be used.

with these modifications, overxrite_specials can be used in traditonal weapon specials and not in abiliy only anymore.

File: data/core/macros/utils.cfg [hunk 6]
```diff
@@ -552,6 +552,9 @@
                         cumulative=no
                         affect_self=yes
                         overwrite_specials=both_sides
+                        [overwrite]
+                            priority=1000
```
Pentarctagon, line 556: Is 1000 a special value, or is it just meant to be a really high value?

newfrenchy83: It is a really high value

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
```diff
@@ -0,0 +1,112 @@
+#####
+# API(s) being tested: [damage]overwrite_specials=both_sides,[overwrite][filter_specials]
```
Pentarctagon, line 2: There should be two separate tests - one for `[damage]overwrite_specials=both_sides` and one for `[overwrite][filter_specials]`.

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
```diff
@@ -0,0 +1,112 @@
...
+                        add=11
+                    [/damage]
+                    [damage]
+                        id=test_overwrite_specials
+                        multiply=2
+                        overwrite_specials=both_sides
+                        [overwrite]
+                            [filter_specials]
+                                [not]
+                                    type_value=value
```
Pentarctagon, line 46: What is `type_value`? It doesn't look like it's new for this PR, but I can't find it documented anywhere either.

newfrenchy83: > What is `type_value`? It doesn't look like it's new for this PR, but I can't find it documented anywhere either.

 `type_value` is an attribute filter used for determine if an ability or specialuse attribite like `value`, `add`, multiply` or `divide` without determine without having to specify a value range, here it is for verify what special using value attribute, but now value range accept negative value,  `type_value` will be prehaps be obsolete.

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
newfrenchy83: > There should be two separate tests - one for `[damage]overwrite_specials=both_sides` and one for `[overwrite][filter_specials]`.

[overwrite][filter_specials]. can't work without overwrite_specials attribute and like alice don't use [damage]apply_to=opponent, what overwtites_specials=both_sides or one_side is not pertinent in this test. Only  `[overwrite][filter_specials]` is realyy tested here.

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
newfrenchy83: changed in https://github.com/wesnoth/wesnoth/pull/7746/commits/bd6824c6ffdf21e92740606ee270d13d544d525f.

newfrenchy83: commit removed

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
Pentarctagon: Alright, then the `API(s) being tested` line should be updated to show that.

newfrenchy83: line changed

newfrenchy83:
@Pentarctagon do't forget this PR, please.

Pentarctagon:
Does #7924 have any effect on this PR?

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
Pentarctagon: Alright, so what this means is: overwrite any weapon special or "weapon special as an ability" that doesn't have a `value` attribute?

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
```diff
@@ -0,0 +1,112 @@
...
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
```
Pentarctagon, line 108: This [store_unit] can be removed, seeing as `b` isn't used for anything after this.

src/units/abilities.cpp
```diff
@@ -1573,6 +1617,133 @@ bool attack_type::has_special_or_ability(const std::string& special, bool specia
 }
 //end of emulate weapon special functions.
 
+//function for filter abilities or specials
+namespace {//helper for attack_type::special_matches_filter() and unit::ability_matches_filter()
```
Pentarctagon, line 1621: Is there a reason that all these were moved over from `unit.cpp`? If nothing else, this will cause a conflict once #7924 is merged.

newfrenchy83:
> Does #7924 have any effect on this PR?

Yes, because this PR uses the same filter as [event]ability, it was by testing a [damage]add=11 with a filter [not]value=0-20 that I realized that it does not 'was not overwritten as it should have been (value=0 but it uses another numeric attribute)

src/units/abilities.cpp
newfrenchy83: moved reverted

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/test_overwrite_specials_filter.cfg
newfrenchy83: done

src/units/abilities.cpp
```diff
@@ -1338,23 +1333,61 @@ static bool overwrite_special_affects(const config& special)
...
+			);
+		}
+		for(double priority : priorityvec){
+			if(priority > 0){
+				unit_ability_list temp_overwriters;
+				std::pair<unit_ability_list, int> overwrite_list;
+				temp_overwriters.append_if(overwriters, [&](const unit_ability& i) {
+					auto overwrite_specials = (*i.ability_cfg).optional_child("overwrite");
+					return (overwrite_specials && overwrite_specials["priority"].to_double(0) == priority);
+				});
```
Pentarctagon, line 1370: Why is this only appending to `temp_overwriters` if it matches a priority from `priorityvec`? Is there any case where there would be no matching priority?

src/units/abilities.cpp
```diff
@@ -1338,23 +1333,61 @@ static bool overwrite_special_affects(const config& special)
...
-unit_ability_list attack_type::overwrite_special_checking(unit_ability_list input, unit_ability_list overwriters, bool is_special) const
+unit_ability_list attack_type::overwrite_special_overwriter(unit_ability_list overwriters, const std::string& tag_name) const
 {
-	for(unit_ability_list::iterator i = overwriters.begin(); i != overwriters.end();) {
-		if(!overwrite_special_affects(*i->ability_cfg)) {
-			i = overwriters.erase(i);
-		} else {
-			++i;
+	std::vector<double> priorityvec;
+	int list_count = 0;
```
Pentarctagon, line 1339: This ends up being the number of elements in `overwriters`, correct?

src/units/abilities.cpp
newfrenchy83: For each value of the vector processed, only the elements having this priority value can be used to decide which elements of lower priority will be removed or not, once this processing has been carried out the list cleared of these elements serves again as a check to constitute the temporary list made up of following priority items where they still exist. let's take the example of a1,a2 and a3 which all have overwrite_specials but a1 has priority over a2 and a3 and a2 has priority over a3 but the filters of a1 do not correspond to a3 but on the other hand a2 can overwrite a3, if I use the list complete as a reference for its own splicing a1 overwrites a2 but a2overwrites a3 in the output list because the input list does not change. with a temporary list we first have a1 which overwrites a2 but not a3, a2 cannot overwrite a3 because it is already overwritten when there since the priority value which corresponds to it no longer corresponds to the list at the end of checking by a1 , so a1 and a3 are found in the final overwriter list

src/units/abilities.cpp
newfrenchy83: list_count count number of element in list because class ability_list don't have function for enumerate number of element.

Pentarctagon: Please instead add a `size()` method to `unit_ability_list`, which returns the size of `cfgs_`.

newfrenchy83:
@Pentarctagon it was a lot of commit but with https://github.com/wesnoth/wesnoth/pull/7746/commits/a2b40c512ac70125324eb4a5bf18d53c93c7ca5c, i maybe find the solution.

src/units/unit.hpp
```diff
@@ -94,6 +94,7 @@ class unit_ability_list
 	const unit_ability& front() const  { return cfgs_.front(); }
 	unit_ability&       back()         { return cfgs_.back();  }
 	const unit_ability& back()  const  { return cfgs_.back();  }
+	int                 size()        { return cfgs_.size(); }
```
Pentarctagon, line 97: A vector's size() returns an std::size_t, not an int.

newfrenchy83: fixed

newfrenchy83:
@Pentarctagon it should be good, now.

newfrenchy83:
> I definitely like this approach more. I'll leave this open for a couple days to see if anyone else says anything, otherwise I'll merge it.
> 
> I've also added some comments - essentially how I understood it while reading through the code.

Thancks for comments.

Issue #7750, opened 2023-07-01
Liberty, S5 Hide and Seek, death of guide seems to be no problem


Konrad22:
### Game and System Information

- **Version:** 1.17.18

### Description of the bug

In Hide and Seek you try to reach the Grey Woods, and the guide that will bring you to their leader.
But if the guide dies, it is not mentioned or taken notice of.
![grafik](https://github.com/wesnoth/wesnoth/assets/39015360/4493db42-a533-4f86-a758-1d2dd06fd3d1)

### Steps to reproduce the behavior

Load the provided save file.
Let a guard spot you.
Watch how the enemies flock to you and kill your guide upon sight.
Game doesn't react to it at all.
[Liberty-Hide and Seek-Auto-Save19.gz](https://github.com/wesnoth/wesnoth/files/11925993/Liberty-Hide.and.Seek-Auto-Save19.gz)

### Expected behavior

I guess, if he dies, you probably shoud lose.

I think it's an edge case that simply was never taken care off, as you are trying very had to avoid being spotted, something that's entirely doable. So the guide pretty much never is killed. I'd expect that it will happen more often in the future though, because of the achievement of killing an Iron Mauler in this scenario.

PR #7756, opened 2023-07-02
Liberty S5: Make the loss of Hans a lose-condition

Wedge009:
Hans only makes an appearance in this scenario and isn't even mentioned afterwards (that I could see). Still, Link explicitly states that Hans will guide Baldras and Harper to the Grey Woods so I suppose it does make sense to make a consequence of Hans' death.

I'm not a writer, happy for the dialogue to be updated directly without my input. But the change itself should address @Konrad22's concern in #7750.

### DIFF
--- a/data/campaigns/Liberty/scenarios/05_Hide_and_Seek.cfg
+++ b/data/campaigns/Liberty/scenarios/05_Hide_and_Seek.cfg
@@ -148,7 +148,7 @@ Been a tough few days since we left Elensefar with all them patrols running arou
         [objectives]
             side=1
             [objective]
-                description= _ "Enter the Grey Woods"
+                description= _ "Enter the Grey Woods with Hans"
                 condition=win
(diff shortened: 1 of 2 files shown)


data/campaigns/Liberty/scenarios/05_Hide_and_Seek.cfg
```diff
@@ -717,6 +721,27 @@ Been a tough few days since we left Elensefar with all them patrols running arou
...
+    [event]
+        name=last breath
+        [filter]
+            id=Hans
+        [/filter]
+        [message]
+            speaker=Hans
+            message= _ "You fools! You weren’t supposed to lead them to me..."
+        [/message]
+        [message]
```
nemaara, line 736: No need for an extra message from Baldras, Hans' alone is okay.

nemaara (approved):

PR #7763, opened 2023-07-03
DiD S3: Resolved Darken Volk speaking from the darkness

### LINKED ISSUE #7758: DiD, S3, map is shrouded anew after Malins relflecting

- **Version:** 1.17.18

### Description of the bug

After you reach the end of the map, Malin has to confront himself, and summons a ghost. After that, he returns to the 'normal' map, which is now mostly shrouded, including Darken Volk.

Load the save, get through the introspection, check how Darken Volk is no longer visible.
[DiD-A Haunting in Winter-Auto-Save141.gz](https://github.com/wesnoth/wesnoth/files/11929464/DiD-A.Haunting.in.Winter-Auto-Save141.gz)

### Expected behavior

Either the map not being shrouded anew, or shared vision with Darken Volk, so he isn't speaking from the dark.

### Additional context

_No response_

Wedge009:
Resolves #7758

New commit instead of just reverting 4a5f38c5755da9e4ab3511f1daf2e3987e33d9f7 so I could add an explanatory comment to hopefully avoid this kind of breakage again.

### DIFF
--- a/changelog_entries/did_s3.md
+++ b/changelog_entries/did_s3.md
@@ -1,0 +1,3 @@
+ ### Campaigns
+   * Descent into Darkness
+     * S3: Resolved Darken Volk speaking from the darkness at the end of the scenario (#7758).
--- a/data/campaigns/Descent_Into_Darkness/scenarios/03_A_Haunting_in_Winter.cfg
+++ b/data/campaigns/Descent_Into_Darkness/scenarios/03_A_Haunting_in_Winter.cfg
@@ -2097,6 +2097,11 @@ All sense of time and being are entombed within darkness. How long has passed si
 
         {CLEAR_VARIABLE units}
 
+        # Restore Darken Volk's MP so that he isn't stuck behind the shroud after the redraw
+        {MODIFY_UNIT (id=Darken Volk) max_moves 6}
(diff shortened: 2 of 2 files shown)


Zireael07:
Kinda was thematically cool but I see it was a bug :P

nemaara:
I moved that to the victory event so if you were skipping through with debug you would easily be able to. I wonder how we can still do that, maybe just keep both blocks of code and add in a comment?

Wedge009:
Sorry, 'easily be able to' what?

As best as I can tell, the victory event follows immediately after the defeating the 'rebel soul' event anyway.

Wedge009:
I haven't used those commands before. So you're saying leave the block in the victory event?

nemaara:
I wonder if we can just have both blocks? The victory one doesn't hurt anything and is only helpful upon debug, while the first one is necessary to make shroud work properly. And then add a comment explaining why there's two blocks. Unless there's a better way to implement?

nemaara (approved):

Wedge009:
Fudge, used tabs for indentation instead of spaces (was using diff tool to restore the block instead of my normal text editor).

PR #7765, opened 2023-07-04
[Wmllint] add two missing corrections for Naga Warrior and Myrmidon

knyghtmare:
Adds missing wmllint entries for image path change
- Naga Warrior
- Naga Myrmidon


knyghtmare:
@Toranks does this patch what you have encountered?

Toranks:
imagen
This is what wmllint changes me in the latest version 1.17.18
As you can see, it does not update the base routes, nor the animations.
It's strange since according to what I see on this PR, it should update at least all the routes of the fighter with the previous code

Elvish-Hunter:
> It's strange since according to what I see on this PR, it should update at least all the routes of the fighter with the previous code

Indeed it should, so perhaps the error is somewhere else. This PR, however, looks correct, so I'm merging it.

soliton-:
How exactly is this change useful? If you replace `units/nagas/warrior` then you already replace `units/nagas/warrior-` anyway.

Toranks:
I guess he did it because the naga fighter already had the same scheme before.

soliton-:
The naga fighter has a replacement rule for `units/nagas/fighter.png` -> `units/nagas/fighter/fighter.png`. I'm going to revert this later since it adds nothing.

Issue #5860, opened 2021-06-13
Explain parry and accuracy better in the future


Hejnewar:
### Game and System Information
 - 1.15.13

### Describe the desired feature
Parry and accuracy should be better explained in game. Currently just % is displayed and in tooltip name, there is no information about what they do. 

### Screenshots
![parry](https://user-images.githubusercontent.com/41817546/121820739-be029b00-cc94-11eb-9acd-bac94d29d462.JPG)
![parry1](https://user-images.githubusercontent.com/41817546/121820743-bfcc5e80-cc94-11eb-816b-e18478b7f9e2.JPG)

Wedge009:
I haven't seen parry/accuracy before - is it new in 1.15?

soliton-:
Already there since a couple of stable versions but not used in mainline.

Wedge009:
Yeah, I realised that afterwards, I think while looking at #4348.

stevecotton:
Adding "bonus to chance to hit" to the tooltip in reports.cpp would help. A small task, but it also requires cleaning up the code to make it better to translate, allowing the translator to move the number to the best place in the string, maybe having separate strings for bonus vs malus.

cooljeanius:
Related: #7895 for `u.w.o`, and #2517 for the in-game component

Issue #2517, opened 2018-02-19
Display effects of accuracy/parry in the attack dialogue


Kwandulin:
Since 1.13.2 we have the possibility to set accuracy and parry values for an attack through [effect]. Those values are displayed in the side bar next to the attacks, but they are not displayed in the attack dialogue when attacking another unit. As those values can have a great impact on the outcome of a battle, it might be reasonable to display those.

gfgtdf:
ok wait, maybe i minsuindersttod is the iccure that they displayed acciracy doesn't take the parry/accuracy  into effect or that it is not displayed that those changes come from a accuracy/parry effect?

CelticMinstrel:
Just for the record, I'm pretty sure `parry` and `accuracy` have been around for far, far longer than that (since like, 1.8 or something?). I'd be very surprised if they're not accounted for in the calculations shown in the attack dialog. I _wouldn't_ however be surprised if the attack dialog doesn't include them in the breakdown.

Kwandulin:
I think the parry/accuracy should be displayed separately in the attack dialogue, like 

"sword
6x4
40% (+20%)"

instead of

"sword
6x4
60%"

cooljeanius:
> I think the parry/accuracy should be displayed separately in the attack dialogue, like
> 
> "sword 6x4 40% (+20%)"
> 
> instead of
> 
> "sword 6x4 60%"

Advanced preference for toggling between the two? Or maybe do one as a tooltip of the other?

CelticMinstrel:
A tooltip with the breakdown makes sense to me.

cooljeanius:
Related: #7895 (for `wmlunits`)

cooljeanius:
> Related: #7895 (for `wmlunits`)

...and also #5860 (possible dup?)

Wedge009:
@cooljeanius I'm not familiar with the milestone convention but I hope it's okay to move this to 1.19 given 1.18 is already out.

Wedge009:
Also agreed that #5860 probably has some overlap but it's inaccessible for the moment.

doofus-01:
I can't access the referenced #5860, but I can say that up to 1.19.2+, the accuracy/parry UI is still an issue:
![Screenshot_20240804_parry](https://github.com/user-attachments/assets/ecdb3e10-7e85-4e8e-af6b-f8d8f44e4b9b)

[...]

PR #7766, opened 2023-07-04
Balance for Rebels 1.18

Hejnewar:
Level 1: 
Fighter - xp cost changed from 40 to 37.
Merman Hunter - xp changed from 35 to 33.
Scout - xp changed from 32 to 42.

Level 2: 
Hero - xp changed from 90 to 80.
Captain - xp changed from 90 to 85.
Marksman - cost changed from 31 to 34, xp changed from 80 to 80.
Ranger - xp changed from 90 to 85. 
Rider - ranged damage changed from 9 to 11, hp changed form 46 to 49, cost changed from 28 to 35, xp changed from 53 to 75.
Druid - cost changed from 27 to 25, xp changed from 80 to 85.
Sorceress - cost changed from 32 to 34, xp changed from 100 to 90. 
Elder Wose - cost changed from 27 to 28, xp changed from 100 to 80.
Merman Netcaster - cost changed from 27 to 26, xp changed from 85 to 54.
Merman Spearman - cost changed from 27 to 22, xp changed from 85 to 54.

Level 3:
Champion - melee damage changed from 9 to 8. added 10 melee accuracy, hp changed from 70 to 72, cost changed from 48 to 61.
Marshal - hp changed from 62 to 68, cost changed from 54 to 67.
Sharpshooter - ranged damage changed from 10 to 12. cost changed from 51 to 62.
Avenger - ranged damage changed from 10 to 11, hp changed from 55 to 59, cost changed from 53 to 66.
Outrider - ranged damage changed from 8 to 11, melee damage changed from 7 to 8,  hp changed form 57 to 60, mp from 10 to 11, cost changed from 43 to 58.
Shyde - hp changed from 46 to 51, melee damage changed from 6 to 7, ranged slow (entangle) changed from 6 to 8, ranged magical (thorns) changed from 8 to 14, cost changed from 52 to 69.
Enchantress - cost changed from 55 to 70, ranged slow damage changed from 5 to 7, ranged magical damage changed from 9 to 13, xp changed from 180 to 180.
Ancient Wose - cost changed from 48 to 50.
Entangler - cost changed from 46 to 42.
Javelineer - cost changed from 48 to 55.

Level 4:
Sylph - hp changed from 60 to 68, ranged slow damage changed from 6 to 7, ranged magical damage changed from 10 to 16, cost changed from 67 to 148 (this is wrong but as commited will fix latter).

### DIFF
--- a/data/core/units/elves/Champion.cfg
+++ b/data/core/units/elves/Champion.cfg
@@ -5,7 +5,7 @@
     race=elf
     image="units/elves-wood/champion.png"
     profile="portraits/elves/hero.webp"
-    hitpoints=70
+    hitpoints=72
     movement_type=woodland
     movement=5
     {LESS_NIMBLE_ELF}
@@ -14,7 +14,7 @@
     alignment=neutral
     advances_to=null
     {AMLA_DEFAULT}
-    cost=48
+    cost=61
     usage=fighter
     description= _ "Elves are typically peaceable by nature and most will attack only in retaliation for some wrongdoing wrought upon them. However, there are some elves who deliberately seek battle, purposeful or otherwise, for their own enjoyment. Through constant combat, these warriors either learn to hone their swordsmanship to a level far beyond most of their peers or perish in the volatile frenzy of melee. Those capable of persisting eventually grow skilled enough to wholly conquer the battlegrounds, becoming veritable champions of war as they mercilessly cut through their foes. Having spent their whole lives refining their prowess with the blade, these elves are exceptionally dangerous and should never be underestimated."
     die_sound={SOUND_LIST:ELF_HIT}
@@ -24,8 +24,9 @@
         icon=attacks/greatsword-elven.png
         type=blade
         range=melee
-        damage=9
+        damage=8
         number=5
+        accuracy=10
     [/attack]
     [attack]
         name=bow
--- a/data/core/units/elves/Avenger.cfg
+++ b/data/core/units/elves/Avenger.cfg
@@ -7,7 +7,7 @@
     image="units/elves-wood/avenger.png"
     small_profile="portraits/elves/ranger.webp~CROP(20,20,400,400)"
     profile="portraits/elves/ranger.webp"
-    hitpoints=55
+    hitpoints=59
     movement_type=woodland
     movement=6
     experience=150
@@ -15,7 +15,7 @@
     alignment=neutral
     advances_to=null
     {AMLA_DEFAULT}
-    cost=53
+    cost=66
     usage=mixed fighter
     description= _ "The curious name of the Elvish ‘Avengers’ comes from a tactic often employed by these master rangers. The enemy is allowed to break through a feint defense and when the vulnerable troops behind the front line follow, these archers break cover and attack, cutting supply lines and surrounding the enemy in one fell stroke. This has, at times, been interpreted as a form of vengeance for their brethren lost earlier in the battle. While not so base in design, it is not at all inaccurate."
     [special_note]
@@ -49,7 +49,7 @@
         icon=attacks/bow-elven.png
         type=pierce
         range=ranged
-        damage=10
+        damage=11
         number=4
     [/attack]
     {DEFENSE_ANIM_RANGE "units/elves-wood/avenger-sword-defend.png" "units/elves-wood/avenger-sword.png" {SOUND_LIST:ELF_HIT} melee}
--- a/data/core/units/elves/Captain.cfg
+++ b/data/core/units/elves/Captain.cfg
@@ -16,7 +16,7 @@
     movement_type=woodland
     movement=5
     {LESS_NIMBLE_ELF}
-    experience=90
+    experience=85
     level=2
     alignment=neutral
     advances_to=Elvish Marshal
--- a/data/core/units/elves/Druid.cfg
+++ b/data/core/units/elves/Druid.cfg
@@ -10,11 +10,11 @@
     hitpoints=36
     movement_type=woodland
     movement=5
-    experience=80
+    experience=85
     level=2
     alignment=neutral
     advances_to=Elvish Shyde
-    cost=27
+    cost=25
     usage=healer
     description= _ "Some elves choose to deepen their connection to the corporeal world, thereby enhancing their already potent abilities to foster and manipulate natural energies. Perhaps the most notable effect of these druids’ magic is the marked sense of motion in the deep woodlands that they preserve. Outsiders wandering through Elven woods often speak of entangling roots and vines, moving nearly as fast as a warrior’s marching pace, or the creaking shamble of ancient trees, lumbering across entire camps in just a single night. The forest and the earth become almost like a connected entity, something that lives and breathes like a single massive organism. As it flourishes under the druids’ careful tending, the forest kindles their magics in turn, symbiotically enhancing the already potent powers that these elves possess."
     die_sound={SOUND_LIST:ELF_FEMALE_HIT}
--- a/data/core/units/elves/Enchantress.cfg
+++ b/data/core/units/elves/Enchantress.cfg
@@ -14,7 +14,7 @@
     level=3
     alignment=neutral
     advances_to=Elvish Sylph
-    cost=55
+    cost=70
     usage=mixed fighter
     description= _ "In the context of Elven magic, the notion of enchanting is a little bit misleading since it almost always refers to manipulation of an object’s essence for the purpose of divination or insight. A favorite among the corresponding enchantresses is imbuing their natural surroundings with the breath of the arcane, yielding secret groves that exist at the juncture between the corporeal and ethereal worlds. These mysterious domains are kept by their equally enigmatic caretakers, who mostly keep to themselves away from prying eyes. Though rather reserved and unsociable, these elves are well-regarded by most of their kin and sometimes called upon for guidance especially in unusual or peculiar matters."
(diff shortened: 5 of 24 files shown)


knyghtmare (approved):

Pentarctagon:
Looks like this broke the pathfinding WML unit tests, probably because of the Outrider's movement change.

Hejnewar:
The real question is what can be done about that? Honestly tests breaking with unit mp changes doesnt sound great.

stevecotton:
Having the pathfinding tests depend on the mainline units seems a reasonable tradeoff to me, given that a rebalance is a very rare event. There's documentation about how to update these tests:
https://github.com/wesnoth/wesnoth/blob/b50cf83f7e358ee11b094d129861f65a51bcdfa3/data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/InternalActionsWML/characterize_pathfinding.cfg#L2-L14

Pentarctagon:
It might be simplest to have a separate PR specifically for the movement changes. Also any changes to the elvish archer's or orcish grunt's hp or weapon damage would likely break a bunch of tests, since those are the two units used in nearly all test scenarios.

Hejnewar:
Thats decent idea. If archer got changes it would be in this PR, and grunt doesnt get any changes, so at least thats something I guess.

stevecotton:
Movement changes are probably the changes most likely to need campaign changes. For example, SoF S07:
https://github.com/wesnoth/wesnoth/blob/1efa0c8c54eb9247b93bcc9dc1ba96cc477b183f/data/campaigns/Sceptre_of_Fire/scenarios/7_Outriding_the_Outriders.cfg#L324-L329

I don't know how many other examples there are, I've just remembered that scenario.

Hejnewar:
Well, honestly changes to rebels are probably the biggest in terms of numbers, because they were supposed to be more elite. I just did what I thought was in character for the unit line, but I can simply swap it for something else if needed.

data/core/units/elves/Champion.cfg
```diff
@@ -24,8 +24,9 @@
         icon=attacks/greatsword-elven.png
         type=blade
         range=melee
-        damage=9
+        damage=8
         number=5
+        accuracy=10
```
cooljeanius, line 29: Why modify accuracy directly in the attack like this, rather than adding a weapon special?

Hejnewar: What is the purpose of this stat existing then? It exists so im using it. Not gonna do weapon specials for stuff that exists already as a stat.

cooljeanius: well ok, does it show up anywhere in-game? How is the player supposed to know about it being there?

Hejnewar: Why is this question for me? If players dont see unit stats then thats UI probelm. 

It is visible, not very well explained, but visible.

stevecotton: There are no unit tests for `accuracy`, it's not used directly in mainline WML (there's some use of `increase_accuracy`, but not `accuracy`), and it's used exactly once in the entirety of Ageless Era. It has the C++ code to support it in the combat calculations and store_unit/unstore_unit, but that's about it.

If we're going to support it properly, it should have tests and UI, and there are already two issues about that (#2517 and #5860). However, as we're getting close to the release candidates for 1.18, it seems late in the development cycle to add those to the roadmap.

Using a weapon special instead would be a simple way to ensure it works well with the current UI, and at this point in the 1.18 cycle that seems the best choice to me.

knyghtmare: >  If players dont see unit stats then thats UI problem.

I agree with Hejnewar here. The UI should be enhanced to augment the UX of the game and I dont see this is a argumentative blocker for the balance patch. Also, the accuracy of +10% is shown in the sidebar under the unit's attack as a `+10%/0%` which means +10% to CTH on attack, and 0% to enemy CTH on defense.

Hejnewar: So, because no one uses it, it was never a feature that was made stable and its usage is discouraged because it is unstable? Since I wanted to use it in many places, there was an issue for few years. It wasnt made any less bad. Now I want to use it. 

And well I will use it anyway so maybe that issue is finally resolved.

stevecotton: If we were at 1.17.5 or 1.19.5, then I'd agree that using it is the best way to get the issues tested and resolved.

But we're about 3 months before string-freeze with other UI issues that need fixing, and still working on getting the event system and ability filter changes tested. I hope that the accuracy feature is stable, but if it isn't, then I doubt it will be fixed for the 1.18 branch.

Pentarctagon: What's the problem with the current UI for it?

Also, "you can't use features that have no unit tests" isn't a requirement that I'm aware of? It would certainly be good for a unit test to be added for it, but as long as Hejnewar has tested it himself and can confirm that it does work the way it's supposed to, then IMO that should be good enough.

Hejnewar: It just doesnt say what it does, if we threat it like any other stat that doesnt need explaining thats fine but right now acuraccy x / parry y is all it says. We can assume its fine but usually there is some confusion around it.

stevecotton: Changing the tooltip in #5860 to say `<b>Accuracy:</b> +10% bonus to chance to hit` would solve a lot of my concerns about the UI. Not difficult, but without @cooljeanius flagging this issue it wouldn't have happened before the string-freeze. I'm still not sure that it will happen in time for the string freeze.

> Also, "you can't use features that have no unit tests" isn't a requirement that I'm aware of?

It's more a general feel of how many of the edge cases have been tested, how it interacts with other abilities that change the chance-to-hit stats, for SP how it interacts with the PRNG options. The existence of unit tests is reassuring, the amount that it gets used in Mainline + Ageless is my main measure for guessing how many of the edge cases are known.

For PRs, my sad lesson from experience is to assume that testing was minimal unless someone say otherwise.

Pentarctagon: @Hejnewar have you tested using the Champion with the `accuracy` attribute added? Or @ProditorMagnus have there been any issues with it in Ageless Era?

ProditorMagnus: Rashy Era uses it out of actively developed source eras. With 4.30 it was also added to several other units for balancing. I would encourage tooltip change too, but values are visible from attack dialog so that should not be a blocker. @andrzejskutela uses it for boss units too.

stevecotton: I was looking at Ageless 4.28. I confirm, I'm a lot happier about this PR after looking at the current Ageless.

ProditorMagnus: Getting it to core unit should also help to speed up getting it to wmlunits.

cooljeanius: > Getting it to core unit should also help to speed up getting it to wmlunits.

Is there a separate issue for that?

andrzejskutela: for the record, since my name was mentioned
[accuracy + parry visual](https://ibb.co/QXYJR6y)
[hover](https://ibb.co/x5Tw6NV)

ProditorMagnus: Shouldnt be.

Hejnewar: Well for now I will merge it, if there is still some discussion needed it can be moved to an issue and if something is really wrong I can always change the change.

cooljeanius: for cross-referencing purposes, I opened #7895 for the `wmlunits` part

cooljeanius:
> Movement changes are probably the changes most likely to need campaign changes. For example, SoF S07:
> 
> https://github.com/wesnoth/wesnoth/blob/1efa0c8c54eb9247b93bcc9dc1ba96cc477b183f/data/campaigns/Sceptre_of_Fire/scenarios/7_Outriding_the_Outriders.cfg#L324-L329
> 
> I don't know how many other examples there are, I've just remembered that scenario.

This was never addressed, and really should have been. Since it wasn't, it's now caused #8589

spixi:
Regarding the Merman changes:

The difference between Netcaster to Entangler is 16 Gold for 54 XP
The difference between Spearman and Javelineer is 33 Gold for 54 XP

This is a hard mismatch. The Netcaster is more expensive than the Spearman, but it is the other way around for their Level 3 counterparts. Either, the XP requirement of the Spearmen has to be increased or the Javelineer needs a nerf and a cost reduction.

spixi:
Suggestions:
* Increase the Entangler's melee attack from 8-3 to 9-3, because of RIPLIB
* Set the Entangler's cost back to 46
* Increase the Spearman's HP from 43 to 44 and its cost from 22 to 23
* Add parry 10 % to Spearman's and Javelineer's melee attack
* Increase XP requirement of the Spearman from 54 to 72 (with the slightly higher bulk it is easier to level up)

spixi:
Javelineer at 55 gold is pretty expensive in comparision to Siren, which has the same MP and defenses, but a 15-3 magical ranged attack, compared to the 10-4 non-magical of the Javelineer AND only costs 42. 7 mote HP is not worth 13 gold.

knyghtmare:
@spixi Please open a new discussional Issue ticket here instead of commenting on a merged pull request.

PR #7779, opened 2023-07-10
Bump tough-cookie from 4.0.0 to 4.1.3 in /utils/wesnoth-map-diff

cooljeanius:
I have Dependabot PRs turned on in my fork, so it opened this against mine today, so I'm submitting it back upstream as well.

Bumps [tough-cookie](https://github.com/salesforce/tough-cookie) from 4.0.0 to 4.1.3.
- [Release notes](https://github.com/salesforce/tough-cookie/releases)
- [Changelog](https://github.com/salesforce/tough-cookie/blob/master/CHANGELOG.md)
- [Commits](https://github.com/salesforce/tough-cookie/compare/v4.0.0...v4.1.3)

---
updated-dependencies:
- dependency-name: tough-cookie
  dependency-type: indirect
...

Signed-off-by: dependabot[bot] <support@github.com>

### DIFF
--- a/utils/wesnoth-map-diff/package-lock.json
+++ b/utils/wesnoth-map-diff/package-lock.json
@@ -6288,6 +6288,12 @@
         "url": "https://opencollective.com/fast-check"
       }
     },
+    "node_modules/querystringify": {
+      "version": "2.2.0",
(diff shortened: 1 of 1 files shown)


soliton-:
This does not look like a security update. I suggest to stop doing this unless it fixes actual issues. If we really want this update spam we should configure dependabot for this repo.

PR #7790, opened 2023-07-12
1.16 wmlunits fix syntax error

ProditorMagnus:


ProditorMagnus:
After that still getting `AttributeError: Can't pickle local object 'list_contents.<locals>.parse.<locals>.f'` and `OSError: [WinError 87] The parameter is incorrect` so stopping further attempts for now.

Wedge009:
This makes sense to me, and checks have completed.

PR #7792, opened 2023-07-13
Add game logic to King of the Hill scenario

Jonathan-Kelly:
The premise of this venerable mainline map was always for players to fight over the central hilltop castle. Yet its scenario gave no perks to combat the downsides of taking the hill:

- No recruiting while leader is traveling there.
- Leader vulnerable while traveling.
- Hill surrounded by enemies on multiple sides.

This update adds a powerful perk to hopefully make the hill worth fighting over.

~~When a leader sits on the hill keep, all other castles collapse into ruins so that no other side can recruit. Only if this leader (the king) is assassinated or abdicates his hilltop throne will the other castles be restored. To make the situation more clear, there is no fog or shroud on the hill, so that all players can see some of what goes on there at all times.~~

~~[Forum Development Thread](https://forums.wesnoth.org/viewtopic.php?t=56209)~~

UPDATE: Based on extensive feedback from @Hejnewar on discord, rewrote the scenario to be as follows:

A side with a leader holding the hill keep gets bonus gold and a bonus leader (who spawns on their original keep), A treasure chest which only a leader can unlock further rewards the first player to take the hill.

Also there are two tweaks to the King of the Hill map itself. One switches a handful of tiles to improve water access for the northern castles so that they are less disadvantaged when deploying water units. The other is a minor cosmetic touch up to help distinguish the cold and wet northern regions from dry and hot southern ones by switching to a slightly greener shade of grass under the pine forest tiles.

### DIFF
--- a/data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
+++ b/data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
@@ -1,62 +1,275 @@
 #textdomain wesnoth-multiplayer
 
-# original map by telex4, recreated by Becephalus 1/06, revised bex 05/06
+#define SPAWN_HEIR SIDE PLACEMENT
+    [store_side]
+        side={SIDE}
+    [/store_side]
+    # Get the faction leader list of the side.
+    [lua]
+        code=<<
+            local t = ...
+            for faction in wml.child_range(wesnoth.scenario.era, "multiplayer_side") do
+                if faction.id == t.faction then
+                    wml.variables.random = faction.leader
+                    return
+                end
+            end
+            >>
+        [args]
+            faction=$side.faction
+        [/args]
+    [/lua]
+    # Pick a random leader from the list.
+    {RANDOM $random}
+    [store_unit_type]
+        type=$random
+    [/store_unit_type]
+    # Give this leader whichever role is unfilled.
+    [if]
+        [have_unit]
+            side=$side.side
+            role=heir
+        [/have_unit]
+        [then]
+            {VARIABLE role reclaimer}
+        [/then]
+        [else]
+            {VARIABLE role heir}
+        [/else]
+    [/if]
+    # Spawn it.
+    [unit]
+        side=$side.side
+        type=$random
+        canrecruit=yes
+        role=$role
+        recall_cost=$unit_type.cost
+        placement={PLACEMENT}
+    [/unit]
+#enddef
 
 [multiplayer]
     id=multiplayer_King_of_the_Hill
     name= _ "4p — King of the Hill"
     map_file=multiplayer/maps/4p_King_of_the_Hill.map
-    description= _ "Controlling the area around the central keep is very lucrative in this 4 player FFA map. Works fine 2 vs. 2 as well. There are 36 villages." + _ " Recommended setting of 2 gold per village and 150 starting gold."
+    description= _ "The King is dead. Without an heir, his quarrelsome lords now vie for the hilltop throne and all the power and wealth it commands. Are you mighty enough to take this seat of power before your rivals do? Can you hold out against their many assassins? Is your arse truly great enough to fill the throne of a king? This free for all map has 38 villages and bonus income for the leader on the center keep." + _ " Recommended setting of 2 gold per village and 150 starting gold."
 
     {DEFAULT_SCHEDULE}
     {DEFAULT_MUSIC_PLAYLIST}
 
     [side]
-        [ai]
-            villages_per_scout=16
-        [/ai]
         side=1
-        team_name=south-west
-        user_team_name= _ "teamname^Southwest"
-        canrecruit=yes
+        team_name=southwest
+        user_team_name= _ "Southwest"
         controller=human
         gold=150
-        fog=yes
-    [/side]
-    [side]
         [ai]
             villages_per_scout=16
         [/ai]
+    [/side]
+    [side]
         side=2
-        team_name=south-east
-        user_team_name= _ "teamname^Southeast"
-        canrecruit=yes
+        team_name=southeast
+        user_team_name= _ "Southeast"
         controller=human
         gold=150
-        fog=yes
-    [/side]
-    [side]
         [ai]
             villages_per_scout=16
         [/ai]
+    [/side]
+    [side]
         side=3
-        team_name=north-east
-        user_team_name= _ "teamname^Northeast"
-        canrecruit=yes
+        team_name=northeast
+        user_team_name= _ "Northeast"
         controller=human
         gold=150
-        fog=yes
-    [/side]
-    [side]
         [ai]
             villages_per_scout=16
         [/ai]
+    [/side]
+    [side]
         side=4
-        team_name=north-west
-        user_team_name= _ "teamname^Northwest"
-        canrecruit=yes
+        team_name=northwest
+        user_team_name= _ "Northwest"
         controller=human
         gold=150
-        fog=yes
+        [ai]
+            villages_per_scout=16
+        [/ai]
     [/side]
+    [side]
+        side=5
+        controller=ai
+        controller_lock=yes
+        team_name=guard
+        team_lock=yes
+        user_team_name= _ "High Guard"
+        color=gold
+        color_lock=yes
+        gold=150
+        [ai]
+            villages_per_scout=16
+        [/ai]
+    [/side]
+
+    [options]
+        [slider]
+            id="bonus_income"
+            default=15
+            min=5
+            max=40
+            name= _ "Hill Bonus Income"
+            description= _ "The amount of extra gold per turn granted to the leader who occupies the hill keep."
+        [/slider]
+        [slider]
+            id="chest_gold"
+            default=75
+            min=25
+            max=500
+            step=25
+            name= _ "Treasure Chest Gold"
+            description= _ "The amount of gold stashed inside the treasure chest on the hill."
+        [/slider]
+    [/options]
+
+    [event]
+        name=prestart
+        [objectives]
+            [objective]
+                description= _ "Defeat enemy leaders"
+                condition=win
+            [/objective]
+            [note]
+                description= _ "The treasure chest on the hill keep can only be unlocked by a leader, for a bonus of $chest_gold gold."
+            [/note]
+            [note]
+                description= _ "A side with a leader holding the hill keep gets +$bonus_income gold income each turn."
+            [/note]
+            [note]
+                description= _ "Each side gets two leaders. Whenever a leader is lost then another appears on the recall list at its full cost."
+            [/note]
+        [/objectives]
+        # The hill is always visible to all players.
+        [lift_fog]
+            location_id=5
+            radius=4
+            multiturn=yes
+        [/lift_fog]
+        [remove_shroud]
+            location_id=5
+            radius=4
+        [/remove_shroud]
+        # Heirs of computer players try to take the hill keep.
+        [micro_ai]
+            side=1,2,3,4
+            ai_type=goto
+            action=add
+            release_unit_at_goal=yes
+            [filter]
+                role=heir
+            [/filter]
+            [filter_location]
+                location_id=5
+            [/filter_location]
+        [/micro_ai]
+        # Reclaimers of computer players try to (re)take a home keep.
+        [micro_ai]
+            side=1,2,3,4
+            ai_type=goto
+            action=add
+            release_unit_at_goal=yes
+            [filter]
+                role=reclaimer
+            [/filter]
+            [filter_location]
+                location_id=1,2,3,4
+            [/filter_location]
+        [/micro_ai]
+    [/event]
+
+    # Give each side an heir (a second leader).
+    [event]
+        name=turn 1 refresh
+        [store_side]
+            variable=sides
+        [/store_side]
+        [foreach]
+            array=sides
+            [do]
+                [if]
+                    [have_unit]
+                        side=$this_item.side
+                    [/have_unit]
+                    [then]
+                        {SPAWN_HEIR $this_item.side leader}
+                    [/then]
+                [/if]
+            [/do]
+        [/foreach]
+    [/event]
+
+    # A leader on the hill keep gets a kingly tribute each turn.
+    [event]
+        name=turn refresh
+        first_time_only=no
+        [if]
+            [have_unit]
+                [filter_location]
+                    location_id=5
+                [/filter_location]
+                canrecruit=yes
+                side=$side_number
+            [/have_unit]
+            [then]
+                [gold]
+                    side=$side_number
+                    amount=$bonus_income
+                [/gold]
+                [scroll_to]
+                    location_id=5
+                [/scroll_to]
+                [floating_text]
+                    location_id=5
+                    text= _ "<span color='gold'>+$bonus_income gold</span>"
+                [/floating_text]
+            [/then]
+        [/if]
+    [/event]
+
+    # If a side loses a leader then put a new heir on its recall list.
+    [event]
+        name=die
+        first_time_only=no
+        [filter]
+            canrecruit=yes
+        [/filter]
+        {SPAWN_HEIR $unit.side map}
+    [/event]
+
+    # A leader unlocks the hill treasure chest and takes its gold.
+    {PLACE_IMAGE items/chest.png 18 18}
+    [event]
+        name=moveto
+        [filter]
+            [filter_location]
+                location_id=5
+            [/filter_location]
+            canrecruit=yes
+            side=1,2,3,4
+        [/filter]
+        [gold]
+            amount=$chest_gold
+            side=$unit.side
+        [/gold]
+        [remove_item]
+            x,y=$x1,$y
+            image=items/chest.png
+        [/remove_item]
+        [floating_text]
+            x,y=$x1,$y1
+            text= _ "<span color='gold'>+$chest_gold gold</span>"
+        [/floating_text]
+    [/event]
 [/multiplayer]
+
+#undef SPAWN_HEIR
--- a/data/multiplayer/maps/4p_King_of_the_Hill.map
+++ b/data/multiplayer/maps/4p_King_of_the_Hill.map
@@ -1,38 +1,38 @@
 Ms, Ms, Aa, Ha, Aa, Ai, Ai, Aa^Fpa, Ha, Ha, Ai, Ha, Aa^Fpa, Ha, Ms, Ha, Ms, Ms, Ha, Ha, Ha, Ha, Aa^Fpa, Ha, Ha, Ha, Ha, Ha, Ai, Ha, Ha, Ha, Ha, Ms, Ms, Ms, Ha, Ms
-Ms, Ms, Aa, Ha, Aa, Ai, Ai, Aa^Fpa, Ms, Ms^Xm, Ms, Ai, Ms, Aa^Fpa, Ha, Ai, Aa^Fpa, Ha, Ms, Ai, Ms, Ms, Aa^Fpa, Ha, Ha^Vhha, Ha, Ha, Ha, Ms, Ai, Ms, Ha, Ha, Ms, Ai, Ms, Ms, Ha
-Aa^Fpa, Aa^Fpa, Aa^Fpa, Aa, Ch, Aa, Aa, Ww, Aa, Ms, Rd, Ha, Aa, Ms, Aa^Fpa, Ms, Ms, Ha, Ms, Ms, Ha, Ms, Ms, Aa^Fpa, Rd, Ai, Ai, Ai, Ai, Ai, Ha, Ch, Ch, Aa^Fpa, Ha, Ha, Ms, Ha
-Gs^Fp, Gs^Fp, Gg, 4 Kh, Ch, Ch, Gg, Ww, Ww^Bw/, Ww^Bw/, Aa, Rd, Rd, Ha, Ha, Ha, Aa^Fpa, Ha, Ha, Ms, Aa^Fpa, Ms, Ha, Rd, Aa, Aa^Fpa, Ai, Ai, Ai, Ai, Ha, Ch, Rr, 3 Kh, Ch, Aa^Fpa, Aa^Fpa, Aa^Fpa
-Gs^Fp, Gs^Fp, Gg, Ch, Ch, Ch, Rr, Rr, Hh, Ww, Ww, Aa^Fpa, Aa^Fpa, Rd, Rd, Aa^Fpa, Rd, Aa, Rd, Aa, Rd, Aa^Fpa, Rd, Rd, Rd, Aa, Aa, Ww, Ww, Ww, Aa, Aa, Rr, Rr, Ch, Ch, Ha, Ha
-Gg, Gg, Gs^Fp, Gg, Gg, Gg, Rr, Mm, Gg, Ww, Ww, Ha, Ms, Aa, Aa^Vea, Rd, Aa, Rd, Aa^Fpa, Rd, Aa, Rd, Aa^Vha, Aa, Aa^Fpa, Rd, Ww^Bw\, Ww, Aa, Aa^Fpa, Aa^Vha, Rr, Ha, Aa, Aa^Fpa, Ha, Ha^Vhha, Ha
-Gs^Fp, Gs^Fp, Gg, Gg, Rd, Rd, Gs^Fp, Rr, Rr, Gs^Fp, Gs^Fp, Ww, Ww, Ww, Wwf, Rd, Wwf, Aa, Ww, Ha, Wwf, Wwf, Wwf, Ww, Ww, Aa, Ww, Ww^Bw\, Rd, Ha, Rr, Rr, Rd, Rd, Rd, Rd, Aa^Fpa, Aa^Fpa
-Gg^Tf, Gg, Gs^Fp, Rd, Gg, Gg, Hh, Gg, Gg, Rr, Rr, Hh, Hh, Ha, Aa, Wwf, Aa, Ww, Aa^Fpa, Ww, Aa, Rd, Aa, Ha, Ms, Ww, Ww, Aa, Ha, Rr, Aa^Fpa, Ms, Rd, Aa^Fpa, Aa, Aa, Ai, Ai
-Hh, Hh, Hh^Vhh, Rd, Gs^Fp, Hh, Mm, Mm, Gg, Gg, Rr, Gs^Fp, Gg^Vh, Gg, Rd, Rd, Rd, Ha, Rd, Aa^Fpa, Rd, Rd, Gg, Ha, Wwf, Ww, Ww, Aa^Fpa, Ms, Rr, Rd, Aa^Fpa, Rd, Aa, Ww, Ai, Ha, Ha
-Hh, Hh, Gs^Fp, Rd, Gg, Hh, Gg, Gg^Tf, Re, Re, Gg, Rr, Rr, Rd, Gg^Tf, Gg, Gs^Fp, Rd, Gg^Ve, Rd, Gg, Gg, Gg, Wwf, Wwf, Wwf, Aa, Aa, Rr, Rr, Aa, Rd, Aa, Ww^Bw\, Gg, Aa, Ha, Ha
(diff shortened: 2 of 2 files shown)


Hejnewar:
...but that isnt a big bonus for you.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -54,9 +48,135 @@
...
+    [/event]
+
+    # Leaders must not move on first turn ahead of recruits, to balance team games against an opening leader rush for the Hill.
+    [event]
+        name=turn 1
+        [modify_unit]
+            [filter]
+                canrecruit=yes
+            [/filter]
+            moves=0
```
ProditorMagnus, line 90: I dont think this does what you think it does

knyghtmare: it kind of does. I tested the addon version of this and it works as intended

ProditorMagnus: Right, first turn is special and doesnt have refresh.

Jonathan-Kelly:
Based on extensive feedback from @Hejnewar on discord, rewrote the scenario to be as follows:

A side with a leader holding the hill keep gets bonus gold and a bonus leader (who spawns on their original keep), A treasure chest which only a leader can unlock further rewards the first player to take the hill.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: This code was no longer useful after removing the castle destruction mechanic so it is gone now, anyway.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -54,9 +48,167 @@
...
     [/side]
+
+    [options]
+        [slider]
+            id="bonus_income"
+            default=15
+            min=5
+            max=40
+            name= _ "Hill Bonus Income"
+            description= _ "Gold Per Turn To Leader On Hill Keep"
```
CelticMinstrel, line 62:
```suggestion
            description= _ "The amount of extra gold per turn granted to the leader who occupies the hill keep."
```

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -54,9 +48,167 @@
...
+        [/slider]
+        [slider]
+            id="chest_gold"
+            default=100
+            min=25
+            max=500
+            step=25
+            name= _ "Treasure Chest Gold"
+            description= _ "Gold Inside Hill Treasure Chest"
```
CelticMinstrel, line 71:
```suggestion
            description= _ "The amount of gold stashed inside the treasure chest on the hill."
```

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -54,9 +48,167 @@
...
+
+    [event]
+        name=prestart
+        [objectives]
+            [objective]
+                description= _ "Defeat enemy leaders"
+                condition=win
+            [/objective]
+            [note]
+                description= _ "A side with a leader holding the hill keep gets +$bonus_income gold income each turn"
```
CelticMinstrel, line 83:
```suggestion
                description= _ "A side with a leader holding the hill keep gets +$bonus_income gold income each turn."
```

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -54,9 +48,167 @@
...
+            [/note]
+            [note]
+                description= _ "This side will also get a second leader on its original keep"
```
CelticMinstrel, line 86:
```suggestion
                description= _ "This side will also get a second leader on its original keep."
```

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -54,9 +48,167 @@
...
+            [/note]
+            [note]
+                description= _ "The treasure chest on the hill keep can only be unlocked by a leader for its $chest_gold gold"
```
CelticMinstrel, line 89:
```suggestion
                description= _ "The treasure chest on the hill keep can only be unlocked by a leader, for a bonus of $chest_gold gold."
```

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -6,7 +6,7 @@
     id=multiplayer_King_of_the_Hill
     name= _ "4p — King of the Hill"
     map_file=multiplayer/maps/4p_King_of_the_Hill.map
-    description= _ "Controlling the area around the central keep is very lucrative in this 4 player FFA map. Works fine 2 vs. 2 as well. There are 36 villages." + _ " Recommended setting of 2 gold per village and 150 starting gold."
+    description= _ "The King is dead. Without an heir. his quarrelsome lords now vie for the hilltop throne and all the power and wealth it commands. Are you mighty enough to take this seat of power before your rivals do? Can you hold out against their many assassins? Is your arse truly great enough to fill the throne of a king? This 4 player FFA map works fine 2 vs. 2 as well. There are 38 villages and bonus income for a leader on the center keep." + _ " Recommended setting of 2 gold per village and 150 starting gold."
```
CelticMinstrel, line 9: Basic fix:

```suggestion
    description= _ "The King is dead. Without an heir, his quarrelsome lords now vie for the hilltop throne and all the power and wealth it commands. Are you mighty enough to take this seat of power before your rivals do? Can you hold out against their many assassins? Is your arse truly great enough to fill the throne of a king? This 4 player FFA map works fine 2 vs. 2 as well. There are 38 villages and bonus income for the leader on the center keep." + _ " Recommended setting of 2 gold per village and 150 starting gold."
```

I also don't think the "arse" reference is appropriate (not to mention it's a Britishism).

CelticMinstrel (commented):
This mostly looks good, but I think option descriptions should be more verbose (they're shown as a tooltip, so there's no worry about space), and I spotted a couple of other minor things.

Hejnewar:
My two remaining concerns are:
1. Willingness of players to take the risk and fight fo the hill.
2. Map thats discouragining direct combat between players.
3. Some pathway issues (limited capacity).

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: Okay so tool tips should just use regular grammar (including a period), I will try to keep that in mind in the future.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: Looks good, suggestion applied.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: So the objectives win/lose conditions use capitalization to start the sentence but no period, and then objectives notes use standard capitalization with a period?

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: Looks good, applied.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: Suggestion applied, thanks.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -54,9 +48,167 @@
...
+        name=moveto
+        [filter]
+            x,y=18,18
+            canrecruit=yes
+        [/filter]
+        [gold]
+            amount=$chest_gold
+            side=$unit.side
+        [/gold]
+        {REMOVE_IMAGE $x1 $y1}
```
gfgtdf, line 208: This looks like it would remove all images on this tile, which could break add-ons/modifications that want to place images

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
CelticMinstrel: I think if it's a full sentence it should have a period. If it's a partial sentence (eg a phrase like "Death of the leader") then I think it doesn't need a period.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
CelticMinstrel: See `[remove_item]` for how to avoid that.

CelticMinstrel:
Someone, possibly @nemaara or @Pentarctagon, needs to either explicitly approve the new map description or request changes.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -54,9 +48,167 @@
...
+                    [then]
+                        [store_unit]
+                            [filter]
+                                x,y=18,18
+                            [/filter]
+                            variable=twin
+                        [/store_unit]
+                        [switch]
+                            variable=side_number
+                            [case]
```
gfgtdf, line 142: Using [store_starting_location] would probably be better here.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: Good suggestions. Applied.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: That is much cleaner, thanks!

nemaara (approved):

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -1,12 +1,10 @@
 #textdomain wesnoth-multiplayer
 
-# original map by telex4, recreated by Becephalus 1/06, revised bex 05/06
-
 [multiplayer]
     id=multiplayer_King_of_the_Hill
     name= _ "4p — King of the Hill"
     map_file=multiplayer/maps/4p_King_of_the_Hill.map
-    description= _ "Controlling the area around the central keep is very lucrative in this 4 player FFA map. Works fine 2 vs. 2 as well. There are 36 villages." + _ " Recommended setting of 2 gold per village and 150 starting gold."
+    description= _ "The King is dead. Without an heir, his quarrelsome lords now vie for the hilltop throne and all the power and wealth it commands. Are you mighty enough to take this seat of power before your rivals do? Can you hold out against their many assassins? Is your arse truly great enough to fill the throne of a king? This 4 player FFA map works fine 2 vs. 2 as well. There are 38 villages and bonus income for the leader on the center keep." + _ " Recommended setting of 2 gold per village and 150 starting gold."
```
Pentarctagon, line 7: Suggestion:

> This 4 player FFA map works fine 2 vs. 2 as well.

"This map can be played as a free for all or 2v2.

gfgtdf:
The description says that it can also be played as 2vs2, but from what i take from the discussion here, the newly added logic somehow relies on all players teaming up against  the person holding the middle, So is it really true that the map still works in 2vs2? (if not that part of the descriptions should probably just be removed)

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
CelticMinstrel: Does that mean you're okay with the use of "arse" in here?

Jonathan-Kelly: Suggestion applied, thanks.

CelticMinstrel: It _is_ more of a British term, I believe, but as long as you're okay with it…

Pentarctagon: It did seem out of character compared to the rest of the game, but as long as we don't keep adding more and more profanity laced descriptions to things I don't think it's an issue. It definitely gives it a more amusing/interesting feel.

Pentarctagon (approved):

Hejnewar:
> I believe it should be fine as a 2v2, especially after the recent rewrite based on Hejnewar's feedback

Actually Im hyperfocusing on FFA. I guess I should also say something more here since I have some time. 

In case of scenarios it is generally much better to focus on one thing than to achieve many things at once because players have a choice and we can provide muliple scenarios so players wanting to play 2v2 will choose scenario perfect for 2v2 and FFA players would rather play scenario focused on FFA rather than something compromised. 

That said it isnt a lot of work to get it to good standard in this case. 

1. Options, never give players sliders, sliders are great but players arent great with sliders. This kills lots of promissing scenarios cuz creators think that players will just balance them themselves but yeah thats not happening. 
2. There should be 3 checkboxes, 1) FFA, 2) 2v2, 3) custom (enables sliders), cuz each mode requires different balance and players should have defaults to fall back on after failing with sliders. 
3. 5th side, basically a roadbump that prevents players form getting to the keep too fast. (It can also exist in FFA it just has to be way weaker - actually thats good idea).

Mics:
Non-blocker but starting with different leaders would be cool.

Jonathan-Kelly:
>Actually Im hyperfocusing on FFA.

That seems perfectly fine to me; I can just cut "2v2" from the map description. After all, it is not called "King**S** of the Hill". :)

_"Highlanders: There can be only two."_

>2. There should be 3 checkboxes, 1) FFA, 2) 2v2, 3) custom (enables sliders), cuz each mode requires different balance and players should have defaults to fall back on after failing with sliders.

We could do something like this if `[options]` allowed more than just the one "default" preset or if the sliders became greyed-out and non-interactive or invisible until the "custom" checkbox was checked.

But the bigger issue might be that we do not know what percentage of players actually click on the "Custom Options" tab when selecting a scenario, especially since almost all scenarios have nothing on that tab. Most players might set up a team game on the second scenario creation screen without realizes they need to click a "2v2" checkbox on a (usually empty/useless) tab of the first scenario creation screen.

So like you said (regarding scenarios generally), it is probably better to just create a second scenario if we want a 2v2 KotH.

>5th side, basically a roadbump that prevents players form getting to the keep too fast. (It can also exist in FFA it just has to be way weaker - actually thats good idea).

Okay, so for FFA, you want me to put an exclusively computer controlled 5th side on the hill? Since I know it is an AI I can boost its `village_gold` a bit too.

>Non-blocker but starting with different leaders would be cool.

Hmm, there may not currently be an elegant way to query a faction's leader list from WML or Lua. Otherwise it would be a fun idea.

Hejnewar:
> Okay, so for FFA, you want me to put an exclusively computer controlled 5th side on the hill? Since I know it is an AI I can boost its village_gold a bit too.

There is no need to make this side overtuned or hard to beat. About as strong as player side should be fine, its purpose is mitigating P1 avdantage in the first place and starting the action sooner. The map kinda feels empty without it after testing with it. 

> Hmm, there may not currently be an elegant way to query a faction's leader list from WML or Lua. Otherwise it would be a fun idea.

This is fine, UR uses something similar so it is surely possible to do. Anyway I can improve upon it after this is merged.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -44,11 +72,25 @@
...
-        user_team_name= _ "teamname^Northwest"
+        team_name=northwest
+        user_team_name= _ "Northwest"
         controller=human
         gold=150
     [/side]
+    [side]
+        [ai]
+            villages_per_scout=16
+        [/ai]
```
CelticMinstrel, line 83: I only just noticed this, but why are the `[ai]` blocks at the top of each `[side]` instead of at the bottom? It does look like it was already like that before you started, so not really relevant to this pull request, but I thought we had a convention of always placing keys above tags (with a few exceptions in certain special circumstances, but this wasn't one of those for sure).

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -1,10 +1,38 @@
...
+    [if]
+        [variable]
+            name=unit_type.level
+            equals=0
+        [/variable]
+        [then]
+            {RANDOM $unit_type.advances_to}
+        [/then]
+    [/if]
+    [unit]
```
CelticMinstrel, line 21: Should this guy use the "disposable leader" overlay? Or do you get a game over when he dies?

CelticMinstrel:
> But there may be a better solution; randomly picking second leaders from the level 1 units on the recruit list, since the recruit list _can_ be easily read by WML, unlike the leader list. I just added an early version of this feature for you to test. If you like it, I can refine it a bit more so it works better for Age of Heroes and most conceivable UMC factions.

Choosing from the recruit list isn't a _terrible_ solution, but I think it would definitely be much better to choose from the leader list if possible. I believe you can use the currently-unsupported `wesnoth.get_era` function and pass it `wesnoth.scenario.mp_settings.mp_era`. (The function is unsupported because we haven't entirely settled on what it should be replaced with; even unsupported functions usually go through a deprecation cycle.) Of course, this does require you to write Lua, so I don't know to what degree it's worth it.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: Good catch, moved them to the bottom.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: You get a game over only when you have no leaders left.

Jonathan-Kelly:
> I believe you can use the currently-unsupported wesnoth.get_era function and pass it wesnoth.scenario.mp_settings.mp_era.

How do you access the era's `[multiplayer_side]` subtags from Lua? In other words, how do you access the era's factions?

CelticMinstrel:
If I recall correctly, the `get_era` function just returns the era's full WML, which would mean you can iterate over the `multiplayer_side` subtags. The reason the function is unsupported is that it's not guaranteed to return anything, because an era might not even be loaded even if you have it installed, but if you're only asking for the active era, it should be fine.

Jonathan-Kelly:
Picking from the leader list via lua seems to be working well. Thanks for the advice @CelticMinstrel

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -4,20 +4,44 @@
...
-        [/variable]
-        [then]
-            {RANDOM $unit_type.advances_to}
-        [/then]
-    [/if]
+    # Get the faction leader list of the side.
+    [lua]
+        code=<<
+            local t = ...
+            local era = wesnoth.get_era(wesnoth.scenario.era.id)
```
CelticMinstrel, line 11: I forgot this existed… I think maybe it will work to just use `wesnoth.scenario.era` and not call `wesnoth.get_era`?

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -4,20 +4,44 @@
...
+            for i,v in ipairs(era) do
+                if v["tag"] == "multiplayer_side" then
```
CelticMinstrel, line 13: I'd use `for faction in wml.child_range("multiplayer_side")` here.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -4,20 +4,44 @@
...
+                    for k,faction in pairs(v) do
+                        if faction["id"] == t.faction then
```
CelticMinstrel, line 15: Even if you wanted to use `ipairs` for some reason, this isn't correct. You would write `local faction = v.contents` (or `v["contents"]` if you prefer) and go from there.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -4,20 +4,44 @@
...
+                            wesnoth.set_variable("random", faction["leader"])
```
CelticMinstrel, line 16: This should be `wml.variables["random"] = faction["leader"]`, or `wml.variables.random = faction.leader`.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
```diff
@@ -4,20 +4,44 @@
...
+        [/args]
+    [/lua]
+    # Pick a random leader from the list and then switch to a level 1 version of it.
+    {RANDOM $random}
+    [lua]
+        code=<<
+            local t = ...
+            for id,unit_type in pairs(wesnoth.unit_types) do
+                for i,advancement in pairs(unit_type["advances_to"]) do
+                    if advancement == t.advancement and wesnoth.unit_types[id]["level"] == 1 then
```
CelticMinstrel, line 33: I think this logic isn't going to work on an Age of Heroes era – there will be no level 1 unit found. However, in Age of Heroes I think I would expect it to choose a level 2 version of the level 3 leader, so why not just check that the level is less than the level of the main unit? ie, `wesnoth.unit_types[id].level < wesnoth.unit_types[t.advancement].level` or so.

If you _really_ want it to be a level 1 even in Age of Heroes, you're going to have to make this more complicated, because it'll have to find the level 2 that advances to it, and then the level 1 that advances to that level 2.

There's also the question of whether to do something special to handle converging unit lines. It doesn't seem like a high priority though.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
CelticMinstrel: So in that case it kind makes sense to call _both_ your leaders disposable when you have two, but when you have one, they're not disposable…

Probably not worth doing anything special in that case, I think.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: You are correct, it works. Updated.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: That is a lot cleaner, thanks.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: Was attempting to loop over only the integer indexed subtags and skip over the string 'indexed' attributes. But your improvement of using `wml.child_range` makes this little attempted optimization obsolete.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: That is much more readable, thanks. Change applied in both instances.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: There was (and maybe still is) a rare bug where instead of being "demoted" one level, the heir would be demoted two or more levels somehow. So in Default some players would rarely get a level 0 heir or in Age of Heroes a level 1 heir. Which lead me to assume that `advances_to` was somehow walking all levels of the unit advancement line. So I tried to lock it to level 1 for consistency (plus for some other balance reasons).

But it looks like the level lock did break the demotion mechanic in Age of Heroes as you predicted, so I removed it.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: Makes sense; I will leave the default leadership overlay in place then.

Hejnewar (approved):

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: We decided to remove this whole "demotion" mechanic to support UMC eras that have leaders with no level 1 in their advancement line. Now second leaders are simply pulled from the faction leader list and used as is.

Jonathan-Kelly:
Okay, this is ready to merge.

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Wedge009: I only just noticed this while reviewing the translation files. I did wonder about its acceptability till I saw this conversation. I was also wondering about the British spelling - I suppose it must somehow look less vulgar than the American spelling? (Which I tend to read as 'female donkey' anyway...)

CelticMinstrel: …I thought it just meant "donkey", not specifically a female?

Wedge009: You're probably right.

Jonathan-Kelly: > I was also wondering about the British spelling

America and other overseas colonies did not really exist in the pre-renaissance time period which BfW emulates. Maybe it is just me, but I find reading or hearing such modern dialects in a dark age or medieval setting to be jarring and mood breaking. So paying at least some lip service to period and setting "authenticity" felt somehow appropriate.

Anyway though, that was the only reason for the spelling.

Wedge009: Ah, I forgot about the historical aspect. Same reasoning for why it's the Sceptre of Fire, not scepter as those following American conventions usually spell the word.

CelticMinstrel: > America and other overseas colonies did not really exist in the pre-renaissance time period which BfW emulates. Maybe it is just me, but I find reading or hearing such modern dialects in a dark age or medieval setting to be jarring and mood breaking. So paying at least some lip service to period and setting "authenticity" felt somehow appropriate.

That would be an argument in favour of using American or Canadian English though, as they're less diverged from the historical dialect that would've been spoken in 15th or 16th century Britain – in particular, unlike Britain they retained the R sound after most vowels. (Non-rhotic English was just starting to become a thing around the colonization era; it wasn't considered standard until around the 18th century.)

Wedge009: I don't think you can make a blanket statement one way or another - it would be dependent on the specifics of a given variation. For example, I stayed with -ize suffixes for `en_GB` because that's the convention the previous maintainers followed and I suppose it follows the rationale that -ise suffixes is a newer convention. But there are lot of aspects of American English I would be surprised if they weren't very modern constructs.

Anyway, I was only querying this because I was initially surprised at the vulgarity introduced here (though I appreciate prose for MP is often more light-hearted and less formal that for SP). And I did find it mildly amusing though I usually wouldn't use the word myself. I wasn't intending to start another prolonged discussion on the nuances of language - some people might complain about being 'spammed' again.

Pentarctagon: I'd say there's definitely a difference between debating something in a single comment thread with a few comments a day vs hitting 200 comments in under 2 days. If anyone thinks something like this comment thread is spam, well then that's their problem.

CelticMinstrel: > But there are lot of aspects of American English I would be surprised with if they weren't very modern constructs.

True enough. I was merely pointing out one way that _British_ English has changed since the colonization era, as a way to indicate that it's wrong to assume British English is "older". Neither American nor British English is the dialect spoken in the colonization era – they've both evolved in different directions from there.

soliton-:
This looks like it was squash merged for no reason. Please don't do that unless it is actually needed. (You actually want to combine all commits or when there is just a single commit like here you actually want to rewrite the commit message.)

data/multiplayer/scenarios/4p_King_of_the_Hill.cfg
Jonathan-Kelly: > in particular, unlike Britain they retained the R sound after most vowels.

That is true, but only of the spoken form. The "silent" R's still appear in the spelling, and since there is no voice acting in Wesnoth, it does not act as an anachronism.

Jonathan-Kelly:
Actually, I did a courtesy squash before the merge. The commits had become fairly messy and interdependent over the course of the PR's evolution, such that if one was reverted it could break others in surprising ways.

soliton-:
Right, so no point in squashing it again.

PR #7795, opened 2023-07-15
WC II - Typo and translatability of Difficulty and Version

Toranks:
imagen

It should be backported too

### DIFF
--- a/data/campaigns/World_Conquest/lua/campaign/objectives.lua
+++ b/data/campaigns/World_Conquest/lua/campaign/objectives.lua
@@ -1,12 +1,13 @@
 --creates the objectives of the wc2 scenarios.
 
 local _ = wesnoth.textdomain 'wesnoth-wc'
+local lib = wesnoth.textdomain 'wesnoth-lib'
 local strings = {
 	wct_victory_condition = _"Defeat all enemy leaders and commanders",
 	turns = _"Turns run out",
 	wct_defeat_condition = _ "Lose your leader and all your commanders",
-	difficulty = _"Difficulty: ",
-	version = _"Version: ",
+	difficulty = lib("Difficulty: "),
+	version = lib("Version: "),
 	help_available = _ "An in-game help is available: right-click on any empty hex.",
 }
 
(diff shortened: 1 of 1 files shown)


Wedge009:
Regarding back-port, I seem to recall a change in rules (or perhaps, an establishment of a standard set of rules) regarding string changes on stable branches recently. But 'Difficulty: ' and 'Version: ' are both existing translated strings (specifically, `wesnoth-lib` domain) so I think it should be fine to back-port.

Wedge009 (approved):

stevecotton:
To use the existing strings in 1.16, this ought to look up those strings in the `wesnoth-lib` textdomain. Just create a second local variable similar to the existing `local _ = wesnoth.textdomain 'wesnoth-wc'` a few lines above the one that's changed.

Toranks:
Take a look at the change, to see if I've done it correctly.

Pentarctagon:
That'd only be needed for 1.16 though, right?

Toranks:
> That'd only be needed for 1.16 though, right?

Why? It's for both

stevecotton:
I think we should merge this. While 1.17 doesn't need to avoid adding strings to `wesnoth-wc`, it's easier to keep track of backports if the commit is the same in both places.

gfgtdf:
I wonder whether there is still any reason to have the version there, it used to be the add-ons version but since it was Mainlined there is no such thing as "add-on version"

soliton-:
Yeah, the version part should just be removed.

Toranks:
Do you think it is acceptable to comment on it, so that it is known that it was there? Maybe for debugging purpose

Wedge009:
Heh, couldn't work out how to retract approvals, but you updated it already anyway.

gfgtdf:
It looks to me like you are just deleting the version string variable without changing the code that actually uses it

Toranks:
Difficulty still uses the code

stevecotton (changes requested):
> It looks to me like you are just deleting the version string variable without changing the code that actually uses it

This. Look at line 31 of the file (same line number in both the old and new versions).

stevecotton (commented):
I'm wondering if the other two instances of `wc2_host_version` can be removed too, but this PR now looks good to me. @Hejnewar would be the person who needs to approve it though.

PR #7796, opened 2023-07-16
Movement changes and test update

Hejnewar:
I put a lot of blood into this because cmd breaks my wesnoth and prevents me from really doing this in a smooth way so at least I hope it works. 

Saurain Soothsayer movement points changed from 6 to 7. 
Outrider movement points changed from 10 to 11.


ProditorMagnus:
I suggest creating dedicated test unit for movement test which has old movement parameters.

PR #7801, opened 2023-07-17
Arcane Resistances Rework

Hejnewar:
Smallfoot, armoredfoot, elusivefoot, mounted,  fly, smallfly, lightfly, deepsea, mountainfoot, gurefoot, rodentfoot, lizard, dunefoot, duneelusivefoot, dunearmoredfoot, dunehorse, dunearmoredhorse - arcane resistance changed from 20% to 10%. 
Treefolk - arcane resistance changed from -30% to -20%.
Undeadfoot - arcane resistance changed from -50% to -20%.
Undeadfly - arcane resistance changed from -40% to -20%.
Drakefly, drakeglide, drakeglide2, drakefoot - arcane resistance changed from -30% to -10%.

Float ???

Paladin - arcane resistance changed from 60% to 30%.
White mage - arcane resistance changed from 40% to 30%.
Mage of Light - arcane resistance changed from 60% to 50%.
Lich - arcane resistance changed from -40% to -20%.

@nemaara kinda lore changes, paladin no longer is on a level of MoL with arcane resistance and Lich doesnt have arcane resistane lower than ancient lich. I also dont know exactly know what to do with ghoul from lore perspective should they have higher arcane resistance or should it go down by 10%, Im leaning towards 10% but we will see.

### DIFF
--- a/changelog_entries/ArcaneResistancesRework.md
+++ b/changelog_entries/ArcaneResistancesRework.md
@@ -1,0 +1,13 @@
+ ### Units
+   * Movement type changes:
+     * Smallfoot, armoredfoot, elusivefoot, mounted, fly, smallfly, lightfly, deepsea, mountainfoot, gurefoot, rodentfoot, lizard, dunefoot, duneelusivefoot, dunearmoredfoot, dunehorse, dunearmoredhorse - arcane resistance changed from 20% to 10%.
+     * Treefolk - arcane resistance changed from -30% to -20%.
+     * Undeadfoot - arcane resistance changed from -50% to -20%.
+     * Undeadfly - arcane resistance changed from -40% to -20%.
+     * Drakefly, drakeglide, drakeglide2, drakefoot - arcane resistance changed from -30% to -10%.
+   * Loyalists:
+     * Paladin - arcane resistance changed from 60% to 30%.
+     * White mage - arcane resistance changed from 40% to 30%.
+     * Mage of Light - arcane resistance changed from 60% to 50%.
+   * Undead:
+     * Lich - arcane resistance changed from -40% to -20%.
--- a/data/core/macros/movetypes.cfg
+++ b/data/core/macros/movetypes.cfg
@@ -62,7 +62,7 @@
         impact=120
         fire=100
         cold=100
-        arcane=80
+        arcane=90
     [/resistance]
 #enddef
 
@@ -133,6 +133,6 @@
         impact=80
         fire=50
         cold=150
-        arcane=130
+        arcane=110
     [/resistance]
 #enddef
--- a/data/core/units.cfg
+++ b/data/core/units.cfg
@@ -458,7 +458,7 @@ The life span of the wose is unknown, although the most ancient members of this 
             impact=100
             fire=100
             cold=100
-            arcane=80
+            arcane=90
         [/resistance]
     [/movetype]
 
@@ -588,7 +588,7 @@ The life span of the wose is unknown, although the most ancient members of this 
             impact=90
             fire=110
             cold=100
-            arcane=80
+            arcane=90
         [/resistance]
     [/movetype]
 
@@ -632,7 +632,7 @@ The life span of the wose is unknown, although the most ancient members of this 
             impact=120
             fire=100
             cold=100
-            arcane=80
+            arcane=90
         [/resistance]
     [/movetype]
 
@@ -674,7 +674,7 @@ The life span of the wose is unknown, although the most ancient members of this 
             impact=70
             fire=100
             cold=100
-            arcane=80
+            arcane=90
         [/resistance]
         [special_note]
             note={INTERNAL:SPECIAL_NOTES_DEFENSE_CAP}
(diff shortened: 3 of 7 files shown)


soliton-:
What's the idea here?

Hejnewar:
I will copy what I wrote on forums: 

This agenda has been on the list for quiet some time. Actually changing all the (arcane) resistances to better fit the growing needs of PvE and PvPvE, as well as potentially help with introduction of arcane based units on lower levels.

Reason:
Arcane resistance is very varied one, not counting special units it has the range of 70, from 20 to -50. It could be said that cold is similarly problematic but there it is surely not great but in reality it only affects 1 faction vs 1 unit line. Here on the other hand we have variety of factions and units that can use arcane damage and pretty much every race has its own arcane resistance. So while on level 1 this issue isnt present on higher levels it becomes very apparent.

Slayer95:
The range of 70 is misrepresenting the issue (assuming there is an issue). Calculation-wise, the difference should be measured as a ratio: 1.2/0.5 = 2.4. With the same range of 70, a different ratio may be achieved by shifting all resistances. E.g. from (20, -50) to (60, -10), yielding a ratio of 1.6/0.9 = 1.78.

Also, shouldn't this "rework" be accompanied by changes to the base power levels of units with arcane attacks?

Hejnewar:
> Calculation-wise, the difference should be measured as a ratio: 1.2/0.5 = 2.4

No if anything it should be 1,5/0,8 = 1,875, and if we reverse them 0,8/1,5 = 0,5(3)

> (60, -10), yielding a ratio of 1.6/0.9 = 1.78

Again, it should be 1,1/0,4 = 2,75, and if we reverse them 0,4/1,1 = 0,(36)

> Also, shouldn't this "rework" be accompanied by changes to the base power levels of units with arcane attacks?

It could, but why? The goal of this PR is to make some disucssion happen around this. Once this is fine I can add the changes. Also I was occupied with other stuff yesterday. I didnt feel like balance, but I still wanted to start this.

Slayer95:
> No if anything it should be 1,5/0,8 = 1,875, and if we reverse them 0,8/1,5 = 0,5(3)

Alright. I had the damage formula backwards.

> Anyway, imo this doesnt help average human anyway and just creates more confusion.

The idea is to provide a way to measure the impact of the change.

> It could, but why? 

Without a significant buff after this, the Paladin loses a large part of its use. I expect the same to happen to other units.

Hejnewar:
> The idea is to provide a way to measure the impact of the change.

This actually tells no story here, unless we assume that every arcane damage dealer is supposed to be a specialist  made to kill undead and then we should only look at reduction of negative resistances. 

> Without a significant buff after this, the Paladin loses a large part of its use. I expect the same to happen to other units.

Thats false. Paly is fine without ud to fight. Elves are fine to kill as well. 

Also Im sure that overall arcane damage dealers were... buffed. Simply because way more units were nerfed than buffed. "But they lost their niche!" truth is that the niche never existed, they were simply as good usually as other units and what was happening when they met low res targets was simply slaughter. Actually exatly this pushed me to make this happen (with possibility of arcane mage). 

Thus, it is very likely that the only damage change will actually be adept arcane damage buff (cuz it was only used vs ud).

Slayer95:
Alright. It's not up to me to judge how specific units are affected. I am satisfied as long as their cases are properly studied.

ReynBolt:
**IPS** here.

My only concern about this is how this change will impair, particularly to lv1 ghost. Ghost been always a problematic unit in default balance. Ghost melee attack has to be increased to 5-3 which would help the unit to maintain constant its damage in certain interactions (Woses, Drakes, etc) while not being denied to drain only 1 HP from human (20% arcane res) in neutral ToD.

If lv2 Wraith and lv3 Spectre will keep same values, then they would require lower XP requeriments or more HP. Still changing Wraith from 6-4 to 7-4 makes an huge impact in ToD (7-4 vs 9-4 interaction...) againist units that have no arcane ressistance or as elves -10% arcane ressistance.

Ghosts , Wraith and Spectres needs a reconsideration considering reworks in arcane ressistances.

Hejnewar:
Dont kill dwarf pls. 

I actuallt think that most of this has positive impact overall, worse against drakes? Good. Better vs loy? Good. Worse vs trees? Bad, BUT Elv vs UD is actually mostly dependent on killing trees fast and ghosts arent killing trees fast anyway and at the same time UD got huge buff vs leaders that you previously couldnt really push into. 

Why would they require any buff? I dont see any reason for that. Why would they need to keep the damage as it was?

Hejnewar:
Im actually wondering if I should get drakes to -10% instead.

Hejnewar:
> Did you mean +10% (positive) ? If so, I can work on some small lore adjustments for them to match this. 

No -10% (110 in code)

> I have always wondered why should the Paladin have any arcane attacks or resistances at all? The Grand Knight loses a movement point, so the Paladin is the logical continuation of the 8 MP Knight, but then the Paladin loses the Knight's blade attack when it gets replaced by an arcane attack.

Quick we need new knight level up or RIPLB! :D

CelticMinstrel:
> I have always wondered why should the Paladin have any arcane attacks or resistances at all?

Obviously the paladin has an arcane attack because it's channelling the same "power of light" that the white mage taps into.

Arcane resistance could maybe be explained by special armour.

PR #7799, opened 2023-07-16
Fix song verse ability for UtBS

Hejnewar:
As in title, still no description improvement tho so with that help would be nice.

PR #7805, opened 2023-07-18
add a "halo" and'overlay' abilities for add overlays so that the unit  benefits from a second halo or overlay when conditions are matched

newfrenchy83:
One of the things that bothers me is the permanent character of the halos of the Mage of Light and other units with the "illuminates" ability, which forces them to program only a permanent illumination applied only to the possessor of the ability.

Adding the halo ability does not change anything about the behavior of the unit, but can be used in several cases:

1 allowing the use of ""illuminates" whose activity would be variable, in this case encoding the halo in halos ability and not in the unit_type allows to modulate the appearance of the halo under the same conditions

2. Applying illumination to adjacent units, I know it's pretty cheesy but a set developer might consider it easier if the hlo display follows the same logic.

3 The halo used to illustrate the possession in the unit of a special weapon used as leadership, the halo would be used to raise the possessor of the ability.

for overlay,same logic for illustrate possession of a special weapon used as leadership, or influence on student

### DIFF
--- a/data/core/macros/abilities.cfg
+++ b/data/core/macros/abilities.cfg
@@ -387,3 +387,47 @@ Enemy units cannot see this unit while it is in deep water, except if they have 
         special_note={INTERNAL:SPECIAL_NOTES_FEEDING}
     [/dummy]
 #enddef
+
+#define ADD_ABILITY_HALO ID HALO
+    # Show the image HALO above the unit when the ability with id ID is active
+    [halo]
+        id=halo_{ID}
+        halo_image={HALO}
+        [filter]
+            ability_id_active={ID}
+        [/filter]
+    [/halo]
+#enddef
+
+#define ADD_ABILITY_HALO_ADJACENT ID HALO
+    # Show the image HALO above the adjacent unit when the ability with id ID is active
+    [halo]
+        id=halo_{ID}
+        halo_image={HALO}
+        affect_self=no
+        affect_allies=yes
+        [filter]
+            ability={ID}
+        [/filter]
+        [affect_adjacent]
+            [filter]
+                ability_id_active={ID}
+            [/filter]
+        [/affect_adjacent]
+    [/halo]
+#enddef
+
+#define ICONS_ABILITIES ID ICON
+    [overlay]
+        id=icon_{ID}
+        overlay_image={ICON}
+        [filter]
+            ability_id_active={ID}
+        [/filter]
+    [/overlay]
+#enddef
+
+#define ABILITY_ILLUMINATES_AND_HALO HALO
+    {ABILITY_ILLUMINATES}
+    {ADD_ABILITY_HALO illumination {HALO}}
+#enddef
--- a/data/schema/units/abilities.cfg
+++ b/data/schema/units/abilities.cfg
@@ -42,6 +42,18 @@
 	{WML_MERGE_KEYS}
 [/tag]
 [tag]
+	name="halo"
+	max=infinite
+	super="units/unit_type/abilities/~generic~"
+	{SIMPLE_KEY halo_image string}
+[/tag]
+[tag]
+	name="overlay"
+	max=infinite
+	super="units/unit_type/abilities/~generic~"
+	{SIMPLE_KEY overlay_image string}
+[/tag]
+[tag]
 	name="heals"
 	max=infinite
 	super="units/unit_type/abilities/~generic~"
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -476,6 +476,50 @@ bool unit::has_ability_type(const std::string& ability) const
 	return !abilities_.child_range(ability).empty();
 }
 
+//this string abilitie is used for purpose of add a second halo if condition is matched.
+//It is put here header declaration commodity.
+static void create_vector(std::vector<std::string>& image_list, const config& cfg, const std::string& tag_name)
+{
+	auto ret = std::find(image_list.begin(), image_list.end(), cfg[tag_name].str());
+	if(ret == image_list.end()){
+		image_list.push_back(cfg[tag_name].str());
+	}
+}
+
+const std::vector<std::string> unit::halo_or_icon_abilities(const std::string& tag_name, const map_location& loc) const
+{
+	std::vector<std::string> image_list;
+	for (const config &i : this->abilities_.child_range(tag_name)) {
+		if(ability_active(tag_name, i, loc) && ability_affects_self(tag_name, i, loc) && !i[tag_name + "_image"].str().empty()){
+			create_vector(image_list, i, tag_name + "_image");
+		}
+	}
+
+	const unit_map& units = get_unit_map();
+
+	const auto adjacent = get_adjacent_tiles(loc);
+	for(unsigned i = 0; i < adjacent.size(); ++i) {
+		const unit_map::const_iterator it = units.find(adjacent[i]);
+		if (it == units.end() || it->incapacitated())
+			continue;
+		if ( &*it == this )
+			continue;
+		for (const config &j : it->abilities_.child_range(tag_name)) {
+			if(affects_side(j, side(), it->side())
+				&& it->ability_active(tag_name, j, adjacent[i])
+				&& ability_affects_adjacent(tag_name, j, i, loc, *it)
+				&& !j[tag_name + "_image"].str().empty())
+			{
+				create_vector(image_list, j, tag_name + "_image");
+			}
+		}
+	}
+	if(!image_list.empty()){
+		std::sort(image_list.begin(), image_list.end());
+	}
+	return image_list;
+}
+
 void attack_type::add_formula_context(wfl::map_formula_callable& callable) const
 {
 	if(unit_const_ptr & att = is_attacker_ ? self_ : other_) {
--- a/src/units/animation_component.hpp
+++ b/src/units/animation_component.hpp
@@ -44,7 +44,9 @@ public:
 		frame_begin_time_(0),
 		draw_bars_(false),
 		refreshing_(false),
-		unit_halo_() {}
+		unit_halo_(),
+		abil_halos_(),
+		abil_halos_ref_() {}
 
 	/** Copy construct a unit animation component, for use when copy constructing a unit. */
 	unit_animation_component(unit & my_unit, const unit_animation_component & o) :
@@ -56,7 +58,9 @@ public:
 		frame_begin_time_(o.frame_begin_time_),
 		draw_bars_(o.draw_bars_),
 		refreshing_(o.refreshing_),
-		unit_halo_() {}
+		unit_halo_(),
+		abil_halos_(),
+		abil_halos_ref_() {}
 
 	/** Chooses an appropriate animation from the list of known animations. */
 	const unit_animation* choose_animation(
@@ -134,4 +138,8 @@ private:
 
 	/** handle to the halo of this unit */
 	halo::handle unit_halo_;
+	/** handle to the abilities halos of this unit */
+	std::vector<halo::handle> abil_halos_;
+	/** vector used to check that halo_abilities vector isn't modified between each read */
+	std::vector<std::string> abil_halos_ref_;
 };
--- a/src/units/drawer.cpp
+++ b/src/units/drawer.cpp
@@ -389,6 +389,13 @@ void unit_drawer::redraw_unit(const unit& u) const
 			}
 		};
 
+		const std::vector<std::string> overlays_abilities = u.overlays_abilities();
+		for(const std::string& ov : overlays_abilities) {
+			if(texture tex = image::get_texture(ov)) {
+				textures.push_back(std::move(tex));
+			}
+		};
+
 		disp.drawing_buffer_add(display::LAYER_UNIT_BAR, loc, [=,
 			textures      = std::move(textures),
 			adj_y         = adjusted_params.y,
@@ -461,6 +468,40 @@ void unit_drawer::redraw_unit(const unit& u) const
 	} else if(has_halo) {
 		halo_man.set_location(ac.unit_halo_, halo_x, halo_y);
 	}
+
+	const std::vector<std::string> halos_abilities = u.halo_abilities();
+	bool has_abil_halo = !ac.abil_halos_.empty() && ac.abil_halos_.front() && ac.abil_halos_.front()->valid();
+	if(!has_abil_halo && !halos_abilities.empty()) {
+		for(const std::string& halo_ab : halos_abilities){
+			ac.abil_halos_.push_back(halo_man.add(
+				halo_x, halo_y,
+				halo_ab + u.TC_image_mods(),
+				map_location(-1, -1)
+			));
+		}
+	}
+	if(has_abil_halo && (ac.abil_halos_ref_ != halos_abilities || halos_abilities.empty())){
+		for(halo::handle& abil_halo : ac.abil_halos_){
+			halo_man.remove(abil_halo);
+		}
+		ac.abil_halos_.clear();
+		if(!halos_abilities.empty()){
+			for(const std::string& halo_ab : halos_abilities){
+				ac.abil_halos_.push_back(halo_man.add(
+					halo_x, halo_y,
+					halo_ab + u.TC_image_mods(),
+					map_location(-1, -1)
+				));
+			}
+		}
+	}
+	else if(has_abil_halo){
+		for(halo::handle& abil_halo : ac.abil_halos_){
+			halo_man.set_location(abil_halo, halo_x, halo_y);
+		}
+	}
+
+	ac.abil_halos_ref_ = halos_abilities;
 
 	ac.anim_->redraw(params, halo_man);
 	ac.refreshing_ = false;
--- a/src/units/unit.hpp
+++ b/src/units/unit.hpp
@@ -1559,6 +1559,18 @@ public:
 		return halo_.value_or("");
 	}
 
+	const std::vector<std::string> halo_or_icon_abilities(const std::string& tag_name, const map_location& loc) const;
+
+	const std::vector<std::string> halo_or_icon_abilities(const std::string& tag_name) const
+	{
+		return halo_or_icon_abilities(tag_name, loc_);
+	}
+	/** Get the [halo] abilities halo image(s). */
+	const std::vector<std::string> halo_abilities() const
+	{
+		return halo_or_icon_abilities("halo");
+	}
+
 	/** Set the unit's halo image. */
 	void set_image_halo(const std::string& halo);
 
@@ -1602,6 +1614,11 @@ public:
 		return overlays_;
 	}
 
+	/** Get the [overlay] ability overlay images. */
+	const std::vector<std::string> overlays_abilities() const
+	{
+		return halo_or_icon_abilities("overlay");
+	}
 	/**
 	 * Color for this unit's *current* hitpoints.
 	 *
--- a/changelog_entries/add_halo_overlay_abilities.md
+++ b/changelog_entries/add_halo_overlay_abilities.md
@@ -1,0 +1,2 @@
+### WML Engine
+   * add a [halo] and an [overlay] ability to support halos or overlays illustrate the activity of other abilities like [illuminates] or others
--- a/data/campaigns/Under_the_Burning_Suns/units/quenoth/Sun_Sylph.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/quenoth/Sun_Sylph.cfg
@@ -7,6 +7,7 @@
     gender=female
     image=units/quenoth/sun_sylph/sun-sylph.png
     profile="portraits/quenoth/sun_sylph.webp"
+    halo=halo/elven/shyde-stationary-halo[1~6].png:150
     hitpoints=42
     movement_type=quenoth_float
     movement=6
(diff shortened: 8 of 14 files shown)


newfrenchy83:
i open topic on forum in https://forums.wesnoth.org/viewtopic.php?t=57087

Toranks:
It would be useful if it also worked with overlays, to have active icons in certain conditions. For example, a while ago I made an icon that activated when a unit had an ability active. I always had a lot of trouble getting that icon right, and this solution would fix it instantly.

newfrenchy83:
@stevecotton could you review this PR, please?

newfrenchy83:
@CelticMinstrel @stevecotton someone could review thPR, please? Thanks.

gfgtdf:
I think the idea is good, some things i noticed: 
- `halos` supports `priority` in another or of yours you gave up on priority in favor `[overwrite] priority=`, im not sure which one i like better but we should stay consistent so i think its better to remove the priority feature from this pr until we came to a conclusion what the best way to implement priority is.
- The wml changes to utbs seem incomplete, in particular `ABILITY_SONG_VERSE_ILLUMINATES` seems to be defined but not used (or can i just not see it ?)
- Do we have some rules which ability tags have a trailing `s` and which not? From looking at the wiki it seems pretty inconsistent already so i don't really mind the current naming, but in case someone know more than me about this please let us know.

newfrenchy83:
> I think the idea is good, some things i noticed:
> 
>     * `halos` supports `priority` in another pr of yours you gave up on priority in favor `[overwrite] priority=`, im not sure which one i like better but we should stay consistent so i think its better to remove the priority feature from this pr until we came to a conclusion what the best way to implement priority is.
> 
>     * The wml changes to utbs seem incomplete, in particular `ABILITY_SONG_VERSE_ILLUMINATES` seems to be defined but not used (or can i just not see it ?)
> 
>     * Do we have some rules which ability tags have a trailing `s` and which not? From looking at the wiki it seems pretty inconsistent already so i don't really mind the current naming, but in case someone know more than me about this please let us know.

For change in Utbs, this because i want what player acces to deescription of illuminates same when ability is inactive(actually, it must give ability in same time what halo via [obect] in song verse)

For plural form, i will correct that and change description in Utbs.

For priority, overwrite_specials with or without priority work for abilities used like weapon only, this because for halo, i must create a priority attribute.

src/units/unit.hpp
```diff
@@ -1786,6 +1786,12 @@ class unit : public std::enable_shared_from_this<unit>
 	 */
 	bool ability_matches_filter(const config & cfg, const std::string& tag_name, const config & filter) const;
 
+	// Return a string used for create a second halo for unit under "halos" ability if active
+	const std::string halo_abilities() const;
+
+	// Return a vector string used for create a second set of overlays for unit under "overlays" ability if active
```
stevecotton, line 1792: I'd prefer to have this near the `overlays()` function, instead of being nested within the abilities. As for documentation, maybe

List of the active abilities' overlay= attributes.

stevecotton:
Could someone tell me what the difference between a halo and an overlay is, please? I'm wondering if the halos are just an older and less flexible overlay (although it seems the halo has team coloring, and I'm not sure if the overlays do).

newfrenchy83:
> The most important difference seems to be that halos can be animated? Al least that what I always thought

Yes it is correct

stevecotton:
The wml in UtBS is also getting changed in #7799, so I'm not worried about it being complete here (at least, not until the exact point that we merge). I'm treating that change as an example of what the ability can do.

A question for this PR is whether an ability halo should override image_halo. As the code currently is, a unit can have both, I'm wondering if it would be a feature to allow an ability to hide the normal halo by adding a blank image as the ability halo.

I assume that the priority feature from #7746 would be able to overwrite abilities that create halos, and it would simplify this PR to not handle priority here.

src/units/drawer.cpp
```diff
@@ -462,6 +469,23 @@ void unit_drawer::redraw_unit(const unit& u) const
 		halo_man.set_location(ac.unit_halo_, halo_x, halo_y);
 	}
 
+	bool has_abil_halo = ac.abil_halo_ && ac.abil_halo_->valid();
+	const std::string& halo_abilities = u.halo_abilities();
+
+	if(!has_abil_halo && !halo_abilities.empty()) {
```
stevecotton, line 475: There seems to be a missing case here. Assume the unit has two abilities that show different halos, and something triggers the change from one to the other - then `has_abil_halo` will be true, `halo_abilities.empty()` will be false, yet we'll still have the wrong image.

src/units/abilities.cpp
```diff
@@ -471,6 +471,55 @@ bool unit::has_ability_type(const std::string& ability) const
...
 }
 
+//this string abilitie is used for purpose of add a second halo if condition is matched.
+//It is put here header declaration commodity.
+const std::string unit::halo_abilities() const
+{
+	unit_ability_list halo_list = get_abilities("halo");
+	if(!halo_list.empty()){
+		utils::erase_if(halo_list, [&](const unit_ability& i) {
+			return ((*i.ability_cfg)["image"].empty() || (*i.ability_cfg)["image"].str().empty());
```
stevecotton, line 481: I think this was needed if any ability could add a halo, but if only tagname `[halo]` abilities can add a halo then it doesn't seem necessary to check for the empty string. A `[halo]image=""` could be a way to disable other `[halo]` abilities.

data/core/macros/abilities.cfg
```diff
@@ -387,3 +387,23 @@ Enemy units cannot see this unit while it is in deep water, except if they have
         special_note={INTERNAL:SPECIAL_NOTES_FEEDING}
     [/dummy]
 #enddef
+
+#define ILLUMINATES_HALOS ID HALO
```
stevecotton, line 391: Either the macro name should change, or it shouldn't take an ID.

```
#define ILLUMINATES_HALO HALO
  # Provide a halo image which is only shown when the unit's ABILITY_ILLUMINATES is active
  ...
```

```
#define ADD_ABILITY_HALO ID HALO
  # Show the image HALO above the unit when the ability with id ID is active
  ...
```

data/schema/units/abilities.cfg
```diff
@@ -41,6 +41,19 @@
 	{FILTER_TAG "filter_base_value" base_value ()}
 	{WML_MERGE_KEYS}
 [/tag]
+[tag]
+	name="halo"
+	max=infinite
+	super="units/unit_type/abilities/~generic~"
+	{SIMPLE_KEY image string}
```
stevecotton, line 48: I'd prefer to call it `halo_image`.

I'm wondering if it's better to have this need to be in a `[halo]` tag, or whether it should be support in any ability regardless of the tag name. Requiring the tag to be called `[halo]` seems good for getting it into 1.18 quickly. Calling it `halo_image` would allow us to use `[halo]halo_image=` in 1.18, and then easily decide to allow `[any_tagname]halo_image=` in 1.19.

newfrenchy83:
> The wml in UtBS is also getting changed in #7799, so I'm not worried about it being complete here (at least, not until the exact point that we merge). I'm treating that change as an example of what the ability can do.
> 
> A question for this PR is whether an ability halo should override image_halo. As the code currently is, a unit can have both, I'm wondering if it would be a feature to allow an ability to hide the normal halo by adding a blank image as the ability halo.
> 
> I assume that the priority feature from #7746 would be able to overwrite abilities that create halos, and it would simplify this PR to not handle priority here.

if a unit add [unit_type]halo= and [halo]image= then the two halos willbe displayed

src/units/drawer.cpp
newfrenchy83: missing case fixed

data/schema/units/abilities.cfg
newfrenchy83: changed to halo_image

gfgtdf:
would it make these for units to have multiple halos though (are there usecases?), if yes then it probably also makes sense to allow multiple halos to be added by the halo abilitiy#.

newfrenchy83:
> would it make these for units to have multiple halos though (are there usecases?), if yes then it probably also makes sense to allow multiple halos to be added by the halo abilitiy#.

I make new variable for halo_ability then an ability list one halo only in addition of unit halo

src/units/abilities.cpp
newfrenchy83: checking empty string removed

data/core/macros/abilities.cfg
newfrenchy83: > Either the macro name should change, or it shouldn't take an ID.
`#define ADD_ABILITY_HALO ID HALO
  # Show the image HALO above the unit when the ability with id ID is active
  ...`

second option choose

newfrenchy83: > Either the macro name should change, or it shouldn't take an ID.
> 
> ```
> 
> ```
> #define ADD_ABILITY_HALO ID HALO
>   # Show the image HALO above the unit when the ability with id ID is active
>   ...
> ```

second choice

src/units/abilities.cpp
stevecotton: Much more can be removed - the whole `utils::erase_if` block is unneeded, AFAICS.

data/schema/units/abilities.cfg
stevecotton: I meant rename the `image=` attribute, not the `[halo]` tag.

stevecotton:
What is the usecase for displaying two halos (from either possible source)?

newfrenchy83:
> What is the usecase for displaying two halos (from either possible source)?

let's take the example of Sun Sylph, currently, for her to have the halo when she lights up, we must deprive her of the characteristic halo of forest Syphe and Shyde, the idea here and that she can have both halos in at the same time (as a Sylph, she should have the ancestral halo in addition to that of enlightenment, Elynia in IftU has both even if it is a single halo which the images of the two have been combined into one alone)

data/schema/units/abilities.cfg
newfrenchy83: done

src/units/drawer.cpp
stevecotton: Not completely fixed, but possibly only a single-frame glitch.

At the moment, it'll remove the halo instead of replacing it. However, on the next draw it will detect that there's no halo and add the correct one. (Reviewed, but not tested)

src/units/abilities.cpp
```diff
@@ -471,6 +471,50 @@ bool unit::has_ability_type(const std::string& ability) const
...
+const std::string unit::halo_abilities() const
+{
+	unit_ability_list halo_list = get_abilities("halo");
+	if(halo_list.empty()){
+		return "";
+	}
+	std::vector<double> priorityvec;
+	for(auto& i : halo_list) {
+		if(!(*i.ability_cfg)["priority"].empty()){
+			priorityvec.push_back((*i.ability_cfg)["priority"].to_double(0));
```
stevecotton, line 485: I'm not happy with having a halo-specific priority, and won't approve this PR if it adds priorities. If both this and #7746 are merged then there would be two different priority mechanisms for halo abilities.

newfrenchy83: i removed halo and only overlay ability remain.

newfrenchy83: > I'm not happy with having a halo-specific priority, and won't approve this PR if it adds priorities. If both this and #7746 are merged then there would be two different priority mechanisms for halo abilities.

overwite_specials work for specials only(abilities or specials implemented) and i resolved issue with vector and map system could show 12 abilite halo maximum.

stevecotton: Why do halos need a priority, and why does it need to be a different mechanism to the `[overwrite_specials]` one?

The problem is complexity of the API, which becomes more confusing if there are two things that are priorities, but which work differently.

newfrenchy83:
> would it make these for units to have multiple halos though (are there usecases?), if yes then it probably also makes sense to allow multiple halos to be added by the halo abilitiy#.

multiple halo ability allowed now.

src/units/abilities.cpp
newfrenchy83: priority removed

src/units/unit.hpp
```diff
@@ -1790,6 +1802,8 @@ class unit : public std::enable_shared_from_this<unit>
 private:
 
 	const std::set<std::string> checking_tags_{"attacks", "damage", "chance_to_hit", "berserk", "swarm", "drains", "heal_on_hit", "plague", "slow", "petrifies", "firststrike", "poison"};
+
+	const std::set<std::string> checking_halo_tags_{"heals", "hides", "illuminates", "leadership", "regenerate", "resistance", "skirmisher", "teleport", "attacks", "damage", "chance_to_hit", "berserk", "swarm", "drains", "heal_on_hit", "plague", "slow", "petrifies", "firststrike", "poison"};
```
stevecotton, line 1806: Having it be an ability named `[halo]` (and `[overlay]`) seems clearer to me than relying on a hardcoded list of other tag names. Is this added because priority was removed, and if so, why wasn't the method of filtering based on an ability being active able to work without priority?

I know that I asked for the attribute to be named `halo_image=` with the idea that we could put it into other tags, but that's something that I'd like to look at in 1.19 rather than the run-up to releasing 1.18.

src/units/animation_component.hpp
```diff
@@ -134,4 +142,11 @@ class unit_animation_component
 
 	/** handle to the halo of this unit */
 	halo::handle unit_halo_;
+	/** handle to the abilities halos of this unit */
+	std::map<int, halo::handle> abil_halos_;
+	/** map used for check what each abil_halos_ don't modified */
```
stevecotton, line 147: Without some more context, "don't modified" doesn't explain what it means.

src/units/drawer.cpp
```diff
@@ -462,6 +469,29 @@ void unit_drawer::redraw_unit(const unit& u) const
...
+		}
+		if(has_abil_halo && (halo_ab.empty() || (ac.abil_halos_ref_[i] != halo_ab))) {
+			halo_man.remove(ac.abil_halos_[i]);
+			ac.abil_halos_[i].reset();
+		} else if(has_abil_halo) {
+			halo_man.set_location(ac.abil_halos_[i], halo_x, halo_y);
+		}
+		ac.abil_halos_ref_[i] = halo_ab;
+	}
+	ac.size_vec_ref_ = ac.size_vec_;
```
stevecotton, line 493: If I'm reading this correctly, this seems to be using `std::map` as a way of getting a `std::vector` that doesn't complain about using out-of-bounds indices, and it'll one-by-one remove and re-add halos if the size or order of the vector changes.

It feels like the wrong algorithm to do this, but am I misunderstanding what it's meant to do?

newfrenchy83: I tried to do without map variables by using vector instead and opting for more global checking, but in this case I have overprinting problems and even after this first problem was solved, the set_location function was not working properly and I had the halo which struggled to follow the unit in its movement. This algorithm may not be ideal, but it's the only one that works perfectly

I'll keep it here for now

src/units/unit.hpp
newfrenchy83: priority alreadyremoved

newfrenchy83: > I know that I asked for the attribute to be named `halo_image=` with the idea that we could put it into other tags, but that's something that I'd like to look at in 1.19 rather than the run-up to releasing 1.18.

i pull https://github.com/wesnoth/wesnoth/pull/7856 for show you that what could be the code for this idea for 1.19

src/units/animation_component.hpp
newfrenchy83: doxygen corrected

src/units/drawer.cpp
newfrenchy83: @stevecotton algorythm changed for vector

newfrenchy83:
@stevecotton i think what i have resolved all problems, could you review this PR, please?

src/units/animation_component.hpp
```diff
@@ -134,4 +138,8 @@ class unit_animation_component
 
 	/** handle to the halo of this unit */
 	halo::handle unit_halo_;
+	/** handle to the abilities halos of this unit */
+	std::vector<halo::handle> abil_halos_;
+	/** vector used for check what halo_abilities vector don't modified between each reading */
```
CelticMinstrel, line 143: This needs to be proper English. I'm sort of guessing here, but do you mean this?
```suggestion
	/** vector used to check that halo_abilities vector isn't modified between each read */
```

CelticMinstrel:
The idea seems good, but the design strikes me as a little strange. We're basically adding two new types of ability, `[halo]` and `[overlay]`, that do nothing at all? It seems to me that the exact same functionality could be achieved more simply and less confusingly by just supporting `halo_image` and `overlay_image` in _all_ abilities. If you want a halo or overlay not linked to an ability you could still use `[halo]` or `[overlay]` in that approach just like what's done here, but the halos and overlays that _are_ linked to an ability are much more strongly linked.

Also, is there any demand for something like this on weapon specials? It's not clear to me if it makes sense to have a halo or overlay weapon special. Maybe it does… or maybe it doesn't… but I just want to make sure that the idea hasn't simply been overlooked. Does it make sense?

newfrenchy83:
> The idea seems good, but the design strikes me as a little strange. We're basically adding two new types of ability, `[halo]` and `[overlay]`, that do nothing at all? It seems to me that the exact same functionality could be achieved more simply and less confusingly by just supporting `halo_image` and `overlay_image` in _all_ abilities. If you want a halo or overlay not linked to an ability you could still use `[halo]` or `[overlay]` in that approach just like what's done here, but the halos and overlays that _are_ linked to an ability are much more strongly linked.
> 
> Also, is there any demand for something like this on weapon specials? It's not clear to me if it makes sense to have a halo or overlay weapon special. Maybe it does… or maybe it doesn't… but I just want to make sure that the idea hasn't simply been overlooked. Does it make sense?

the idea of halo and overlay in all abilities is pull in https://github.com/wesnoth/wesnoth/pull/7856 request

newfrenchy83:
@Pentarctagon 

very good, so I will explain the different steps of the code used to create the halo and overlay capabilities.

First of all, in abilities.cpp, the string vector is created which will be used for adding overlay and halo. For this purpose and to save calculation, I reproduced the code used for get_abilities in order to to directly constitute the vector instead of using this function then analyzing the list to create the vector.

Each time a halo/overlay_image is detected, the create_vector function determines whether the image is already present in the vector before adding it or not, this to avoid overlays.

Finally the vector is reorganized in alphabetical order to avoid, for example, a change in the order of the images when two owners of [halo] with [affect_adjacent] of swap their position near the student unit.

Once the vector is created, we move on to the second step accomplished in the animation_component and especially drawer.cpp files.

The creation of an overlay list already using a string vector, I only had to copy this part of the code with the ability vector as source.

For the halos it was much more complicated,
I will first explain for unit halo: first check that the unit has a valid halo then, if this is not the case but e u.halo is not empty, then it must be must be added, conversely if u.halo is empty but the halo is still present in the list, it is removed, and finally if halo already added and present in the wml, then the position information is updated days.

In the case of halo ability, in addition to the use of for loops to add halos coming from a vector, I duplicated the first step of adding the halo to take into account modifications of the vector even when the The target unit already has ability type halos

Pentarctagon:
How does #7856 relate to this PR? Does #7856 effectively supercede this PR for 1.19+?

newfrenchy83:
> How does #7856 relate to this PR? Does #7856 effectively supercede this PR for 1.19+?

The 3th commit of https://github.com/wesnoth/wesnoth/pull/7856 are same what this PR then yes https://github.com/wesnoth/wesnoth/pull/7856 supercede this pr for 1
19(but i will prefer for 1.18

Pentarctagon:
Is there a reason not to have https://github.com/wesnoth/wesnoth/pull/7856 in 1.18?

newfrenchy83:
> Is there a reason not to have #7856 in 1.18?

@stevecotton shoud replace to this question because it's him who want postpone https://github.com/wesnoth/wesnoth/pull/7856

newfrenchy83:
> The idea seems good, but the design strikes me as a little strange. We're basically adding two new types of ability, `[halo]` and `[overlay]`, that do nothing at all? It seems to me that the exact same functionality could be achieved more simply and less confusingly by just supporting `halo_image` and `overlay_image` in _all_ abilities. If you want a halo or overlay not linked to an ability you could still use `[halo]` or `[overlay]` in that approach just like what's done here, but the halos and overlays that _are_ linked to an ability are much more strongly linked.
> 
> Also, is there any demand for something like this on weapon specials? It's not clear to me if it makes sense to have a halo or overlay weapon special. Maybe it does… or maybe it doesn't… but I just want to make sure that the idea hasn't simply been overlooked. Does it make sense?

@Pentarctagon as you can see @CelticMinstrelis also in favor of using attributes in all abilities to better link them so don't hesitate to look at https://github.com/wesnoth/wesnoth/pull/7856 since the first three commit are those of this PR

stevecotton:
> Is there a reason not to have #7856 in 1.18?

Reviewing it again, I think I'm now in favor of #7856's approach going into 1.18. The reasons for not doing that are outdated.

A month ago, the code for that approach was [using a hardcoded list of tags](https://github.com/wesnoth/wesnoth/pull/7805#pullrequestreview-1578126312). I think the code has been streamlined since [the other comment about dedicated tags](https://github.com/wesnoth/wesnoth/pull/7805#pullrequestreview-1573477532) was written.

Both of those links are to comments in this review that are likely hidden under the "45 hidden items Load more..." fold.

newfrenchy83:
> > Is there a reason not to have #7856 in 1.18?
> 
> Reviewing it again, I think I'm now in favor of #7856's approach going into 1.18. The reasons for not doing that are outdated.
> 
> A month ago, the code for that approach was [using a hardcoded list of tags](https://github.com/wesnoth/wesnoth/pull/7805#pullrequestreview-1578126312). I think the code has been streamlined since [the other comment about dedicated tags](https://github.com/wesnoth/wesnoth/pull/7805#pullrequestreview-1573477532) was written.
> 
> Both of those links are to comments in this review that are likely hidden under the "45 hidden items Load more..." fold.

Then i will close this PR for https://github.com/wesnoth/wesnoth/pull/7856 reviewed

PR #7808, opened 2023-07-20
Move {join,host}.lua

soliton-:
Creating a PR to see if anyone is actually using mp_tests. If so it can also be moved to utils instead of removing it.


gfgtdf:
I do think that the ability to run the mp tests is useful, especially when you try to debug a problem that was caught by the mp tests.

I'm not really sure what the difference between running `mp_tests` and `utils/CI/mp_test_executor.sh` is though. If I can just run the later instead I'm fine with removing it.

soliton-:
You can just run `utils/CI/mp_test_executor.sh`. That is what is run for each commit.

`mp_tests` specifically opens gnome-terminal with a couple of tabs open that run similar stuff to the CI script.

PR #7809, opened 2023-07-20
Re-enable Windows CI jobs

Pentarctagon:
Based on some feedback, doing a `git checkout` may help with the frequent cache misses of the vcpkg dependencies.


soliton-:
There should be a comment explaining what the checkout helps with and how to figure out what commit to check out.

Pentarctagon:
If it actually works, sure. The commit chosen is completely arbitrary though.

Issue #7810, opened 2023-07-22
New Editor (#7719) breaks quick reload (F5)


doofus-01:
### Game and System Information

- **Version:**
- Battle for Wesnoth v1.17.19+dev (c2ad332b21d-Modified) x86_64 (includes staged `src/modules/lua (new commits)` somehow?  I didn't touch those...  EDIT: I see it's a link to a different source tree)
- **Downloaded from:**
- here
- **Build info:**
- scons
- **OS:**
Linux

### Description of the bug

Multiple bugs, some are minor, at least one is not: 
 1.  When starting map editor, the new menu "Add-on Selection" menu pops up.  Click "show all" check-box, and any folder in the add-ons folder becomes an option, including temporary or "attic" folders that you could formerly use to disable or hide data.
 2.  It does not seem to matter what you choose, for the map editor - select "Mainline", and UMC terrain shows up just like before.  UMC terrain being available is actually good, so you can still load more than one dependency if you want, but then why did we have to deal with the new menu?
 3.  From the plain grass map, hit F5 to reload.  Or draw something before F5, it doesn't matter.  This leads to two issues:  (A) instead of quickly being brought back to the loaded map(s), you get the "Add-on Selection" menu again; (B) after the menu, the game exits with 
 ```
error editor: Editor exception in editor::start: File does not have .map or .cfg extension
Game error: File does not have .map or .cfg extension
```

### Steps to reproduce the behavior

See above

### Expected behavior

1.  Only show add-ons (folder contains `_main.cfg`?)
2.  Don't make the new menu pop up for the map editor?  Or make the menu do something, but if you do that, allow multiple selections and let it remember previous selections
3. Reloading a map with F5 shouldn't bring up the new menu and it shouldn't cause the game to exit

### Additional context

Rapidly reloading with F5 is a must-have for terrain development/tinkering, placing menus and speed-bumps in there should be avoided.  From what I see, the new feature is getting loaded too soon, it should only be for the scenario editor and should not be involved with the map editor.

doofus-01:
@Pentarctagon  - I'd say this is high priority, even if it doesn't break things for most players.

CelticMinstrel:
[...]

PR #7811, opened 2023-07-23
Fix parts of #7810

Pentarctagon:
F5 causing the editor to crash - the editor saves temporary files out to disk which don't have an extension, so give them the proper extension
The add-on selection dialog showing non-add-on folders in the add-ons folder - only show files with a _main.cfg

### DIFF
--- a/src/editor/map/context_manager.cpp
+++ b/src/editor/map/context_manager.cpp
@@ -815,6 +815,11 @@ void context_manager::save_all_maps(bool auto_save_windows)
 			if(name.empty() || filesystem::is_directory(name)) {
 				std::ostringstream s;
 				s << default_dir_ << "/" << "window_" << i + 1;
+				if(!get_map_context().is_embedded() && !get_map_context().is_pure_map()) {
+					s << ".cfg";
+				} else {
+					s << ".map";
+				}
 				name = s.str();
 				get_map_context().set_filename(name);
 			}
--- a/src/editor/map/map_context.cpp
+++ b/src/editor/map/map_context.cpp
@@ -221,7 +221,7 @@ map_context::map_context(const game_config_view& game_config, const std::string&
 
 			if(!map_data_loc.empty()) {
 				if(map_data_loc.find("\"{") == std::string::npos) {
-					// 2.0 Embedded map
+					// 2.0 Embedded pure map
 					LOG_ED << "Loading embedded map file";
 					embedded_ = true;
 					pure_map_ = true;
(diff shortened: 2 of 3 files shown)


CelticMinstrel:
Linking to #7810 since title issue references aren't detected.

CelticMinstrel:
> F5 causing the editor to crash - the editor saves temporary files out to disk which don't have an extension, so give them the proper extension

Well, I guess this is good enough as a quick-and-dirty solution, but we really shouldn't be implementing F5 as "save to disk, refresh, and load back in" in the first place…

> The add-on selection dialog showing non-add-on folders in the add-ons folder - only show files with a _main.cfg

I don't really like this solution… I think _initial.cfg should be a valid alternate way of defining an add-on. However, it looks like that also doesn't work right now so I guess it'd be more than just a change to the map editor…

Pentarctagon:
Aren't downloaded add-ons just a subset of non-publishable add-ons?

Pentarctagon:
Also why would an _initial.cfg be needed as an alternative to a _main.cfg?

CelticMinstrel:
> Aren't downloaded add-ons just a subset of non-publishable add-ons?

They are, yes. I just thought the three-way distinction of "has _server.pbl", "has info.cfg", and "has neither" might be useful.

> Also why would an _initial.cfg be needed as an alternative to a _main.cfg?

It's probably a niche case, but I was thinking an add-on that's sort of a compilation of a bunch of mini add-ons could be structured with each mini add-on in its own subfolder (potentially with a _main.cfg) and some common stuff in the main folder in _initial.cfg, _final.cfg, or other config files.

This doesn't appear to work right now. It might be easy to make it work though – just get the game to load `Addon_Folder` instead of `Addon_Folder/_main.cfg`.

Pentarctagon:
IMO "has _server.pbl" is enough here - either an add-on is one that they're publishing or it's not. Splitting it further would probably be more confusing than helpful, especially given how rare it would be for an add-on to have neither.

CelticMinstrel:
Maybe. It _would_ be unusual to have neither; probably that only applies to a manually-created add-on early in the creation process.

doofus-01:
Using recent BfW 1.17 a bit more, I think a (temporary?) fix for issue 2 from #7810 would be to have a "don't ask me again" checkbox on the new menu, and maybe in advanced preferences as well, to uncheck it.

CelticMinstrel:
I feel like this isn't a good place to put a "don't ask me again" checkbox. That's good for places where you probably always want the same answer, but in this case, you probably want the same answer for a few weeks and then want to change to a different answer, so "don't ask me again" is just a different kind of annoyance.

Issue #7089, opened 2022-11-05
Wesnoth creates D:\a\msys64\var\cache\fontconfig with two files inside


Toranks:
### Game and System Information

The Battle for Wesnoth version 1.17.9 x86_64
Running on Microsoft Windows 7 Service Pack 1 (6.1.7601) x86_64
Distribution channel: Steam

### Description of the bug

While the game is active it won't let me delete the files, but even if I delete the entire path, when I start the game it is created again.

### Steps to reproduce the behavior

1. Start the game
2. Go to the main folder of the drive (X:) when the game is installed
3. X:\a\msys64\var\cache\fontconfig appears with two cache files inside

### Expected behavior

Store cache files on
Cache dir:       C:\Users\USER\Documents\My Games\Wesnoth1.17\cache

### Additional context

Possibly only happens on Windows 7

Wedge009:
I confirm this using the 1.17.9 Windows installer package from SourceForge. But only on Windows 7, I couldn't find a similar directory for Windows 10 - and I have no idea why that would be the case. I'm not aware of the details of how Wesnoth is cross-compiled for Windows releases.

For my Windows 7 set-up I have C-drive for boot (including Windows installation), D-drive for data, X-drive as a RAM-disk and Z-drive for back-up. The fontconfig cache went to D-drive for me.

Wedge009:
Revisited this while investigating something else. Confirming this was introduced at some point between 1.17.4 and 1.17.5.

Pentarctagon:
It's possible I suppose?

Wedge009:
I'll check if 1.16 starts having the same problem if/when it gets back-ported then.

Pentarctagon:
@loonycyborg Could this potentially be solved by setting a `<cachedir>` in https://github.com/wesnoth/wesnoth/blob/master/fonts/fonts.conf?

knyghtmare:
Confirmed on Windows 10.

Wedge009:
What's your drive layout? Maybe it happens only if a D-drive is available.

knyghtmare:
I have a main C Drive and a second D Drive...and that's it.

![image](https://user-images.githubusercontent.com/5283677/232423361-91271743-0893-4592-939b-7fe26b5516a6.png)

Wedge009:
So presumably there's a hard-coded desire to write to a D-drive in fontconfig somewhere. If there isn't a D-drive, it just doesn't have a cache.

[...]

PR #7816, opened 2023-07-26
try enabling vcpkg's dependency graph

Pentarctagon:


Wedge009:
(Unrelated to PR) Is `D:\a\` path some sort of convention? It caught my eye because it's the same path as #7089.

Pentarctagon:
I don't think it's any sort of convention, other than perhaps D being next letter after C.

Wedge009:
I was thinking more about the `a` directory too, but never mind, probably just a coincidence then.

.github/workflows/ci-main.yml
```diff
@@ -7,6 +7,13 @@ on:
     branches: [ master ]
   workflow_dispatch:
 
+permissions:
+  contents: write
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  VCPKG_FEATURE_FLAGS: dependencygraph
```
soliton-, line 15: I would rather put the env and permissions just on the job that needs it. Also I'm pretty sure the GITHUB_TOKEN part doesn't do anything since that is the default.

Pentarctagon: It's how the documentation said to do it.

PR #7819, opened 2023-07-31
AToTB: Remove the instant-communication amulets

stevecotton:
Having these in canon felt like a plot hole in other campaigns. If an apprentice mage could make these amulets circa 350 YW then every village and border-fort should have them, but Wesnoth's storylines work better if a messenger on horseback is the usual method of long-distance communication. For example, TSG is about a commander sent to discipline a lazy outpost, only to find that they should have brought reinforcements. When it's the enemy trying to communicate, DM S21 and UtBS S08 both involve stopping the messengers.

This changes the intro to have a beacons for communicating with the trade caravans, I'm leaving it ambiguous whether it's a smoke signal or a fire signal. Fire beacons might make more sense if the village sees the caravan's campfires at night and, in normal usage, light the beacon to trade.

While this adds beacons to canon, they're an ancient technology with a limited range and very limited bandwidth. There are easy plot reasons not to set beacons up, such as the costs of having people watch for and maintain them, there are easy plot reasons why they don't work on a given day, such as weather, and there are easy plot reasons for attempted communication to be ignored by assuming that it's a fire for a different purpose.

Previously suggested in https://r.wesnoth.org/p669889 .

### DIFF
--- a/data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
+++ b/data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
@@ -47,32 +47,38 @@
     [story]
         [part]
             music=revelation.ogg
-            story=_ "The remote freehold of Maghre in the western reaches of the Kingdom of Wesnoth was once a peaceful place, its inhabitants largely unaware of the comings and goings of the wider world. Wars and the rumor of wars touched them not, until the day a dark mage settled in the region and began seeking sacrifices for his evil summonings."
-            background=story/Two_Brothers_M1P1.webp
+            # po: This text is shown with the journey map, showing the village in the foothills along the route between Aldril and Dan’Tonk.
+            # po: I'm assuming the beacons either use smoke signals or basic fire signals, with a few agreed ones to request goods or indicate that surplus is for sale.
+            story=_ "The remote freehold of Maghre in the western reaches of the Kingdom of Wesnoth was once a peaceful place, its inhabitants largely unaware of the comings and goings of the wider world. Wars and the rumor of wars touched them not. Even merchants were rarely seen, although caravans between Aldril and Dan’Tonk might send a wagon when called by the beacon tower."
+            {TB_BIGMAP}
+            {JOURNEY_REST_BEFORE_STAGE1}
         [/part]
         [part]
-            story=_ "Skeletons and zombies killed cattle and fired fields. <i>“Fear and obey Mordak the Mage!”</i> they cried in fell voices as they did their foul deeds. People vanished from isolated farmsteads. Men and women began to fear the night, and their children even the bright day. But the nearest lord was more than a day’s ride distant, and messengers sent to seek his help did not return."
+            story=_ "Then came the day that a dark mage settled in the region and began seeking sacrifices for his summonings.
+
+Skeletons and zombies killed cattle and fired fields. <i>“Fear and obey Mordak the Mage!”</i> they cried in fell voices as they did their foul deeds. People vanished from isolated farmsteads. Men and women began to fear the night, and their children even the bright day. But the nearest lord was more than a day’s ride distant, and messengers sent to seek his help did not return."
             background=story/Two_Brothers_M1P1.webp
         [/part]
         [part]
             story=_ "There was a man named Baran who had shown talent as a mage when he was young, gone to the great Academy on the Isle of Alduin, and returned to work his magic in the land where he was born. The people looked to him for help and leadership. He found weapons half-forgotten from the times of their sires and grandsires hanging in many houses, and bade the villagers to take them down and clean and oil them. He set the smiths of Maghre to making spearheads and ax-blades for the rest."
-            background=story/Two_Brothers_M1P1.webp
+            # Although I'd prefer to have less image changes during the story, this text fits neither the previous image of a necromancer, nor the next one which shows Arvith.
+            background=portraits/baran.webp
         [/part]
         [part]
-            story=_ "Now Baran had a brother named Arvith who had also left Maghre to seek his fortune, and had become the leader of a small band of horsemen who hired out as guards to merchant caravans. Fortunate it was for all that when Baran was but an apprentice mage, he had made a pair of amulets for himself and his brother, with which they might call to each other when in dire need. Baran sent out that call."
+            story=_ "Now Baran had a brother named Arvith who had also left Maghre to seek his fortune, and had become the leader of a small band of horsemen who hired out as guards to merchant caravans. Fortunate it was for all that, in better times, the brothers had talked of using trade beacons for other signals when in dire need. Baran sent out that call."
             background=story/Two_Brothers_M1P2.webp
         [/part]
         [part]
             story=_ "12 V, 363 YW
 Excerpt from the journal of Baran of Maghre
 
-If I could but face this ‘Mordak’! I think my magic might prove stronger than his. But he bides in the hills, well-guarded by his servants, and I muster frightened peasants to fight his minions with blades and sticks. I need my brother; he always had a better head for battle than I.
+If I could but face this ‘Mordak’! I think my magic might prove stronger than his. But he bides in the hills, well-guarded by his servants, and I muster frightened peasants to fight his minions with blades and sticks.
 
-Will he heed the call? I do not know if he has kept the amulet; we have not spoken since that evil day at Toen Caric. If he will not come for me, perhaps he will return to aid our village in its hour of desperate need."
+I need my brother; he always had a better head for battle than I. Yet we have not spoken since that evil day at Toen Caric. If he will not come for me, perhaps he will return to aid our village in its hour of desperate need."
             background=story/Two_Brothers_M1P2.webp
         [/part]
         [part]
-            story=_ "Heeding the call of his amulet, Arvith gathered such men as he could and hurried to Maghre to help Baran."
+            story=_ "Heeding the call, Arvith gathered such men as he could and hurried to Maghre to help Baran."
             show_title=yes
             {TB_BIGMAP}
             {JOURNEY_STAGE1}
--- a/data/campaigns/Two_Brothers/scenarios/04_Return_to_the_Village.cfg
+++ b/data/campaigns/Two_Brothers/scenarios/04_Return_to_the_Village.cfg
@@ -362,7 +362,8 @@ But I am still troubled. I wonder... is this sense of foreboding I feel merely a
 
         [message]
             speaker=Arvith
-            message= _ "I must go back to earning my living. But we have our amulets, little brother. If you are beset again, I will come."
+            # po: "you have my word" is a idiom for promising to do something
+            message= _ "I must go back to earning my living. But you have my word, little brother; if you are beset again, I will come."
         [/message]
 
         [endlevel]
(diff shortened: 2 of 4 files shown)


data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
```diff
@@ -47,32 +47,40 @@
     [story]
         [part]
             music=revelation.ogg
-            story=_ "The remote freehold of Maghre in the western reaches of the Kingdom of Wesnoth was once a peaceful place, its inhabitants largely unaware of the comings and goings of the wider world. Wars and the rumor of wars touched them not, until the day a dark mage settled in the region and began seeking sacrifices for his evil summonings."
-            background=story/Two_Brothers_M1P1.webp
+            # po: This text is shown with the journey map, showing the village in the foothills along the route between Aldril and Dan’Tonk.
+            # po: I'm assuming the beacons either use smoke signals or basic fire signals, with a few agreed ones to request goods or indicate that surplus is for sale.
+            story=_ "The remote freehold of Maghre in the western reaches of the Kingdom of Wesnoth was once a peaceful place, its inhabitants largely unaware of the comings and goings of the wider world. Wars and the rumor of wars touched them not.
+
+Even merchants were rarely seen. Caravans between Aldril and Dan’Tonk would sometimes send a wagon to the village when called by the beacon tower."
```
nemaara, line 54: Probably should just be one paragraph, no need to separate.

data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
```diff
@@ -47,32 +47,40 @@
...
+            {TB_BIGMAP}
+            {JOURNEY_REST_BEFORE_STAGE1}
         [/part]
         [part]
-            story=_ "Skeletons and zombies killed cattle and fired fields. <i>“Fear and obey Mordak the Mage!”</i> they cried in fell voices as they did their foul deeds. People vanished from isolated farmsteads. Men and women began to fear the night, and their children even the bright day. But the nearest lord was more than a day’s ride distant, and messengers sent to seek his help did not return."
+            story=_ "Then came the day that a dark mage settled in the region and began seeking sacrifices for his evil summonings.
```
nemaara, line 59:
```suggestion
            story=_ "Then came the day that a dark mage settled in the region and began seeking sacrifices for his summonings.
```

data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
```diff
@@ -47,32 +47,40 @@
...
+
+Skeletons and zombies killed cattle and fired fields. <i>“Fear and obey Mordak the Mage!”</i> they cried in fell voices as they did their foul deeds. People vanished from isolated farmsteads. Men and women began to fear the night, and their children even the bright day. But the nearest lord was more than a day’s ride distant, and messengers sent to seek his help did not return."
```
nemaara, line 61:
```suggestion
Skeletons and zombies killed cattle and fired fields. <i>“Fear and obey Mordak the Mage!”</i> they cried in fell voices as they did their deeds. People vanished from isolated farmsteads. Men and women began to fear the night, and their children even the bright day. But the nearest lord was more than a day’s ride distant, and messengers sent to seek his help did not return."
```

data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
```diff
@@ -47,32 +47,40 @@
...
-            background=story/Two_Brothers_M1P1.webp
+            # Although I'd prefer to have less image changes during the story, this text fits neither the previous image of a necromancer, nor the next one which shows Arvith.
+            background=portraits/baran.webp
         [/part]
         [part]
-            story=_ "Now Baran had a brother named Arvith who had also left Maghre to seek his fortune, and had become the leader of a small band of horsemen who hired out as guards to merchant caravans. Fortunate it was for all that when Baran was but an apprentice mage, he had made a pair of amulets for himself and his brother, with which they might call to each other when in dire need. Baran sent out that call."
+            story=_ "Now Baran had a brother named Arvith who had also left Maghre to seek his fortune, and had become the leader of a small band of horsemen who hired out as guards to merchant caravans. Fortunate it was for all that, in better times, the brothers had talked about trade beacons and signals to use when in dire need. Baran sent out that call."
```
nemaara, line 70:
```suggestion
            story=_ "Now Baran had a brother named Arvith who had also left Maghre to seek his fortune, and had become the leader of a small band of horsemen who hired out as guards to merchant caravans. Fortunate it was for all that, in better times, the brothers had talked about beacons and signals that might be used when in dire need. Baran sent out that call."
```

data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
CelticMinstrel: I just want to point out that @stevecotton did not change this line. Not that there's anything wrong with asking him to, but it's currently identical to what was previously there.

CelticMinstrel:
I have two general comments.

1. The amulets aren't necessarily any more high-tech than the beacons you've replaced them with, are they? From what I recall, there's no mention of them being usable for communication; it could be something very primitive, like stroking one amulet causes the other to grow warm.
2. I don't particularly have an issue with removing words like "evil" and "foul", but I think the intent was for the narration to be from Baran's point of view. Would he not consider those things to be "evil" and "foul"?

data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
stevecotton: Here I'm wanting to convey that the beacon was only expected to be used for trade, and it's something special to the brothers that they'd agreed an additional signal.

stevecotton:
> The amulets aren't necessarily any more high-tech than the beacons you've replaced them with, are they?

I think the amulets can only send a general "I've been activated", but if an apprentice can make these then the cost is trivial compared to building and garrisoning a fort. A fort could have a set of them for "enemy scouts seen", "enemy army approaching", and a handful of other messages.

data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
nemaara: How about the other version I gave you below?

TheShadowOfHassen:
Another thought is there are fairy tales that tell of magical objects that warn one brother about the other brother's need. I believe in the story I read it was a knife in a tree. 

I haven't re read AToTB recently but something like that could work. Something that they had from their godfather (super powerful mage?) It might be to much but its an idea...

nemaara:
Also my suggested revisions are just suggestions, you can take them or leave them as you like. It's okay to merge from my point of view.

TheShadowOfHassen:
@stevecotton I just had the idea and thought I'd put it out. It would be a way to remove the instant communication amulets. Re reading the scenario makes me think that bonfires would be easier, but it was just an idea. (You don't have to take it.)

I really have no reason I just thought it was a cool idea.

Jonathan-Kelly:
>  If an apprentice mage could make these amulets circa 350 YW then every village and border-fort should have them, but Wesnoth's storylines work better if a messenger on horseback is the usual method of long-distance communication.

I agree. Though fixing this might be as simple as changing the origin of the amulets to either:
- Creations of a Silver Mage. (Crafting like amulets is secret knowledge.)
- Creations of the now extinct Wesfolk civilization during its golden age. (Crafting like amulets is forgotten knowledge.)

> Is there a reason why the story would be better with such a device?

The amulets are more personal. The story indicates there was a "falling out" between the brothers, bad enough that one questions if the other would even help him in a matter of life and death, The amulets predate their falling out, and may likewise represent the familial bond which predates whatever event came between them. And which ultimately proves stronger.

TheShadowOfHassen:
ason why the story would be better with such a device?
> 
> The amulets are more personal. The story indicates there was a "falling out" between the brothers, bad enough that one questions if the other would even help him in a matter of life and death, The amulets predate their falling out, and may likewise represent the familial bond which predates whatever event came between them. And which ultimately proves stronger.

@Jonathan-Kelly that actually is a very good reason.

data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
stevecotton: I'll make the change, but when combining it into a single paragraph, I also had to make the two sentences into one.

data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
stevecotton: I'm happy either way about removing the "foul" and "evil", but would prefer to leave it as-is for the simple reason that I dislike single-word changes in text that's already translated.

Moving whole paragraphs in already-translated text sounds like more work that a single word, but it doesn't require the detailed "what has changed here?" search that a single word does.

data/campaigns/Two_Brothers/scenarios/01_Rooting_Out_a_Mage.cfg
stevecotton: Taking "evil" out of this one makes it closer to having an alliteration.

Jonathan-Kelly:
> I fall into the camp which would want the author to eventually show that the family bond didn't need the trinkets, for them to be a item that fuels character development, but which the characters outgrow.

Well they are not just trinkets in (the story's) reality; they serve a useful purpose by warning one brother that the other is facing serious trouble. The brothers would not outgrow a responsibility to help one another.

Either brother could have sold, given, or thrown away his amulet after their falling out. The fact they both kept them shows that even if they were no longer 'on speaking terms' they still had concern for each other's well being.

stevecotton:
Planning to merge before the release tonight, unless @nemaara says no.

> Either brother could have sold, given, or thrown away his amulet after their falling out.

If we put too much emphasis on that, I think it diminishes the story that's currently being told. At the start even Arvith isn't certain whether there's still enough bond between them, and the dialogue in S01's `start` event has Arvith twice saying that he's doing it for their hometown rather than for Baran.

Jonathan-Kelly:
> If we put too much emphasis on that, I think it diminishes the story that's currently being told.

I agree; what I meant was the old dialogue already covered this possibility with: _"Will he heed the call? I do not know if he has kept the amulet; we have not spoken since that evil day at Toen Caric."_

> the dialogue in S01's start event has Arvith twice saying that he's doing it for their hometown rather than for Baran.

He does say that directly to Baran in a brisk toned conversation, but there are a number of hints throughout the story that this is not entirely true.

Anyway, it looks like your changes do a good job of extracting the amulets without any ill effects. Sorry if I sounded dismissive of your work on this.

data/campaigns/Two_Brothers/scenarios/04_Return_to_the_Village.cfg
```diff
@@ -362,7 +362,8 @@ But I am still troubled. I wonder... is this sense of foreboding I feel merely a
 
         [message]
             speaker=Arvith
-            message= _ "I must go back to earning my living. But we have our amulets, little brother. If you are beset again, I will come."
+            # po: "you have my word" is a idiom for promising to do something
```
Wedge009, line 365: Granted, it's just the `po` hint rather than the actual prose, but I think this should read 'an idiom'.

PR #7825, opened 2023-08-02
[1.16] Fix #7820 "Enable planning mode on start" Setting

### LINKED ISSUE #7820: [Bug] Planning mode was not activated at MP game start with enabled corresponding setting

- **Version:** 1.16.9
- **Downloaded from:** Steam
- **OS:** Ubuntu 18

### Description of the bug

When every MP game starts, I have disabled planning mode and manually turn it on every time.
Setting "Enable planning mode on start" is checked.

This started at 1.16.9 probably.

1. Check "Enable planning mode on start".
2. Join to MP game Isar's Cross (hosted by bot IsarFoundation).
3. After game starts, planning mode would not be active.

### Expected behavior

Planning mode is active at game start.

### Additional context

_No response_

gfgtdf:
previously the code that tried to activate the whiteboard failed because it checks `can_modify_game_state` which checks for `!synced_context::is_unsynced`.

This moves init_side_end() further down to be called after check_victory and do_final_checkup for technical reasons. But this is safe to do since these functions only set unrelated variables and don't throw (except in case of  a gamebreaking error ofc)

### DIFF
--- a/src/play_controller.cpp
+++ b/src/play_controller.cpp
@@ -458,6 +458,7 @@ void play_controller::maybe_do_init_side()
 
 void play_controller::do_init_side()
 {
+	{
 	set_scontext_synced sync;
 	log_scope("player turn");
 	// In case we might end up calling sync:network during the side turn events,
@@ -525,9 +526,10 @@ void play_controller::do_init_side()
 	// Make sure vision is accurate.
 	actions::clear_shroud(current_side(), true);
 
-	init_side_end();
 	check_victory();
 	sync.do_final_checkup();
+	}
+	init_side_end();
 }
 
 void play_controller::init_side_end()


src/play_controller.cpp
```diff
@@ -458,6 +458,7 @@ void play_controller::maybe_do_init_side()
 
 void play_controller::do_init_side()
 {
+	{
```
stevecotton, line 461: Putting sync in a std::unique_ptr might be nicer than having the entire function's indentation off-by-one.

gfgtdf:
hmm okay it seems liek the only thing left to do here is to decie whetehr its better to have the indentaion or to use a unique_ptr

gfgtdf:
I went with the extra indentaton for now, mostly because `finish_side_turn` a few lines above does the same thing.

gfgtdf:
Done in https://github.com/wesnoth/wesnoth/commit/8537aad68f9c33700d17ede8419bba069d8e3d1a

PR #7828, opened 2023-08-03
Don't ask for the addon again when using F5

Pentarctagon:


src/editor/editor_main.cpp
```diff
@@ -83,27 +83,30 @@ void initialize_addon(const std::string& addon_id)
...
+				} else {
+					return EXIT_STATUS::EXIT_NORMAL;
+				}
 			}
-		}
 
-		if(addon_id == "newaddon") {
-			std::int64_t current_millis = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
-			addon_id = "MyAwesomeAddon-"+std::to_string(current_millis);
+			if(addon_id == "newaddon") {
```
CelticMinstrel, line 104: I didn't notice this in the original PR, but… I feel like this should be something that's not actually valid as an addon ID. Though the chance of someone wanting to make an addon called "newaddon" is admittedly fairly low, but still.

CelticMinstrel:
Looks good, barring the compiler warning.

src/editor/editor_main.cpp
stevecotton: Something valid to the client, but recognised and rejected (with a specific error message) by campaignd?

I feel it should be a translatable string, but that makes the campaignd part harder.

CelticMinstrel: It doesn't seem like something where translating it would be crucial, as the intent is that the user overrides it. But then again… when you create a new document in many programs, it's called "Untitled"… does that get translated? I don't actually know.

stevecotton: It does in LibreOffice, Gimp, and there's a default filename in Blender. I think all the Microsoft Office programs do too.

Pentarctagon: I mean I can change it to just something like `///newaddon///`.

Pentarctagon: The generated name also isn't something that should really be translated IMO, since it doesn't really have any particular meaning other than "add something to the name besides just the timestamp milliseconds".

PR #7814, opened 2023-07-25
Remove one or more special weapons based on common stats and add 'filter_special' in  [filter_weapon/attack/has_attack] filters

newfrenchy83:
In [effect] you can add a special weapon or replace the existing specials with others, or specifically remove a weapon by the identifier but not by the tag_name or other characteristics.

The use of a [filter_special] associated with a mode=remove should fill this gap

Fix https://github.com/wesnoth/wesnoth/issues/7812 issue

PR #7829, opened 2023-08-03
Support negative numbers in ranges, move to utils/config_filters.cpp

stevecotton:
This was an idea from PRs #7003 and #7746: move some functions that, although currently only used in the weapons and ability code, would be applicable for any config-based filter. @newfrenchy83, @CelticMinstrel and @gfgtdf, what do you think?
Edit: source of the idea was https://github.com/wesnoth/wesnoth/pull/7814#discussion_r1273856127

The task then expanded into how to handle negative values in filters.

Adds support for using these in the weapons and ability filters:
* "-1", which was previously treated as an parse error (no number before the separator).
* "-3--1"
* Using "-infinity" as the lower number in the range, provided a different upper number is given.

Using `..` won't be supported, at least not in this PR.

This treats "-infinity" (with no other number), "-infinity..-infinity",
"infinity" (with no other number) and "infinity..infinity" as errors. It seems
unlikely that someone would intend to use a filter that can't match any
reasonable number.

The range "-infinity..infinity" will be parsed successfully. I don't see a use case for that,
but nor do I see a reason to add a special case to reject it.

This started as a change to move common filter functions from unit.cpp to
somewhere that they could be reused for other config-based filters. In the
process a missing feature was found and added, the move is still included in a
single Git commit because the move was required in order to make these
functions accessible to the Boost unit tests.

Two CodeBlocks project files additionally get src/utils/any.hpp added,
which was in one of them but missing from the other two. I noticed because
these are alphabetically at the start of the src/utils file list.

### DIFF
--- a/data/schema/game_config.cfg
+++ b/data/schema/game_config.cfg
@@ -80,8 +80,8 @@
             [/type]
         [/union]
     [/type]
-    # definition of range_list and s_range_list, before macros are expanded
-    {LIST_TYPE_COMPLEX range (
+    # definition of unsigned_range_list and s_unsigned_range_list, before macros are expanded
+    {LIST_TYPE_COMPLEX unsigned_range (
         [element]
             value="\d+"
         [/element]
@@ -92,16 +92,42 @@
             value="\d+-infinity"
         [/element]
     )}
+    # definition of int_range_list and s_int_range_list, before macros are expanded
+    {LIST_TYPE_COMPLEX int_range (
+        [element]
+            value="(-?)\d+"
+        [/element]
+        [element]
+            value="-\d+-(-?)\d+"
+        [/element]
+        [element]
+            # If the first number is positive, the second one has to be too
+            value="\d+-\d+"
+        [/element]
+        [element]
+            value="-infinity-(-?)\d+"
+        [/element]
+        [element]
+            value="(-?)\d+-infinity"
+        [/element]
+    )}
     # definition of real_range_list and s_real_range_list, before macros are expanded
     {LIST_TYPE_COMPLEX real_range (
         [element]
-            value="\d+(\.\d+)?"
+            value="(-?)\d+(\.\d+)?"
         [/element]
         [element]
+            value="-\d+(\.\d+)?-(-?)\d+(\.\d+)?"
+        [/element]
+        [element]
+            # If the first number is positive, the second one has to be too
             value="\d+(\.\d+)?-\d+(\.\d+)?"
         [/element]
         [element]
-            value="\d+(\.\d+)?-infinity"
+            value="-infinity-(-?)\d+(\.\d+)?"
+        [/element]
+        [element]
+            value="(-)?\d+(\.\d+)?-infinity"
         [/element]
     )}
     [type]
@@ -369,7 +395,7 @@
         name="coordinates"
         [union]
             [type]
-                link=range_list
+                link=unsigned_range_list
             [/type]
             [type]
                 value="recall"
@@ -390,7 +416,8 @@
     [/type]
     {SUBST_TYPE coordinate}
     {SUBST_TYPE coordinates}
-    {SUBST_TYPE range_list}
+    {SUBST_TYPE unsigned_range_list}
+    {SUBST_TYPE int_range_list}
     {SUBST_TYPE real_range_list}
     {SUBST_TYPE terrain_code}
     {SUBST_TYPE terrain_list}
--- a/src/serialization/string_utils.cpp
+++ b/src/serialization/string_utils.cpp
@@ -28,6 +28,7 @@
 #include <cassert>
 #include <array>
 #include <limits>
+#include <optional>
 #include <stdexcept>
 
 #include <boost/algorithm/string.hpp>
@@ -822,24 +823,59 @@ std::vector<std::string> quoted_split(const std::string& val, char c, int flags,
 	return res;
 }
 
+namespace
+{
+/**
+ * Internal common code for parse_range and parse_range_real.
+ *
+ * If str contains two elements and a separator such as "a-b", returns a and b.
+ * Otherwise, returns the original string and std::nullopt.
+ */
+std::pair<std::string, std::optional<std::string>> parse_range_internal_separator(const std::string& str)
+{
+	// If turning this into a list with additional options, ensure that "-" (if present) is last. Otherwise a
+	// range such as "-2..-1" might be incorrectly split as "-2..", "1".
+	static const auto separator = std::string{"-"};
+
+	// Starting from the second character means that it won't interpret the minus
+	// sign on a negative number as the separator.
+	// No need to check the string length first, as str.find() already does that.
+	auto pos = str.find(separator, 1);
+	auto length = separator.size();
+
+	if(pos != std::string::npos && pos + length < str.size()) {
+		return {str.substr(0, pos), str.substr(pos + length)};
+	}
+
+	return {str, std::nullopt};
+}
+} // namespace
+
 std::pair<int, int> parse_range(const std::string& str)
 {
-	const std::string::const_iterator dash = std::find(str.begin(), str.end(), '-');
-	const std::string a(str.begin(), dash);
-	const std::string b = dash != str.end() ? std::string(dash + 1, str.end()) : a;
-	std::pair<int,int> res {0,0};
+	auto [a, b] = parse_range_internal_separator(str);
+	std::pair<int, int> res{0, 0};
 	try {
-		if (b == "infinity") {
-			res = std::pair(std::stoi(a), std::numeric_limits<int>::max());
+		if(a == "-infinity" && b) {
+			// The "&& b" is so that we treat parse_range("-infinity") the same as parse_range("infinity"),
+			// both of those will report an invalid range.
+			res.first = std::numeric_limits<int>::min();
 		} else {
-			res = std::pair(std::stoi(a), std::stoi(b));
+			res.first = std::stoi(a);
 		}
 
-		if (res.second < res.first) {
+		if(!b) {
 			res.second = res.first;
+		} else if(*b == "infinity") {
+			res.second = std::numeric_limits<int>::max();
+		} else {
+			res.second = std::stoi(*b);
+			if(res.second < res.first) {
+				res.second = res.first;
+			}
 		}
 	} catch(const std::invalid_argument&) {
-	    ERR_GENERAL << "Invalid range: "<< str;
+		ERR_GENERAL << "Invalid range: " << str;
 	}
 
 	return res;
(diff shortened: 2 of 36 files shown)


CelticMinstrel:
Seems fine. A cpp file makes sense to me as there's no template code here.

src/utils/config_filters.hpp
```diff
@@ -0,0 +1,96 @@
...
+	return in_ranges<int>(cfg[attribute].to_int(0), utils::parse_ranges(filter[attribute].str()));
+}
+
+/**
+ * Supports filters using "add" and "sub" attributes, for example a filter `add=1` matching a cfg containing either
+ * `add=1` or `sub=-1`; this assumes that code elsewhere has already checked that cfg contains at most one of those
+ * keys.
+ *
+ * This only checks for the presence of @a attribute in the filter, so the caller should call this function a second
+ * time, with @a attribute and @opposite reversed.
```
soliton-, line 68:
```suggestion
 * time, with @a attribute and @a opposite reversed.
```

gfgtdf:
I don't have a really strong opinion on this, 

On the one hand I'm not really convined that these functions are reusable (in particular since some filters like the unit filter convert the cfg into a (c++) filter object first and that ideally all filters should be converted to such a structure).

On the other hand I think that the unit.cpp file is too big and everything that make it easier to read is good.

stevecotton:
Found an issue during testing, and I'm wondering about the significance of it to all of the open PRs.

The filters that take ranges don't accept negative values. Trying to do this will general a lot of `error general: Invalid range: -1` logs:
```
[filter_specials]
    value=-1
[/filter_specials]
```

As well as affecting the `add`/`sub` pair, that affects looking for specials that damage the attacker on a successful hit by using `[heal_on_hit]` or `[drains]` with negative values.
https://github.com/ProditorMagnus/Ageless-for-1-14/blob/f0c159c6ab97d5b3f9e6d45b6e0477829954cb94/data/MiE_data/abilities.cfg#L3-L23
https://github.com/ProditorMagnus/Ageless-for-1-14/blob/f0c159c6ab97d5b3f9e6d45b6e0477829954cb94/units/EoMa_units/Tharis/Crimson_Blade.cfg#L34-L44

newfrenchy83:
> Found an issue during testing, and I'm wondering about the significance of it to all of the open PRs.
> 
> The filters that take ranges don't accept negative values. Trying to do this will general a lot of `error general: Invalid range: -1` logs:
> 
> ```
> [filter_specials]
>     value=-1
> [/filter_specials]
> ```
> 
> As well as affecting the `add`/`sub` pair, that affects looking for specials that damage the attacker on a successful hit by using `[heal_on_hit]` or `[drains]` with negative values. https://github.com/ProditorMagnus/Ageless-for-1-14/blob/f0c159c6ab97d5b3f9e6d45b6e0477829954cb94/data/MiE_data/abilities.cfg#L3-L23 https://github.com/ProditorMagnus/Ageless-for-1-14/blob/f0c159c6ab97d5b3f9e6d45b6e0477829954cb94/units/EoMa_units/Tharis/Crimson_Blade.cfg#L34-L44

this https://github.com/wesnoth/wesnoth/pull/7830 PR should fix that in using '_' instead of'-' like separator.

CelticMinstrel:
I was actually wondering if `parse_range` should be divided into `parse_range_signed` and `parse_range_unsigned`…

newfrenchy83:
it's is correct for me

stevecotton:
> I was actually wondering if `parse_range` should be divided into `parse_range_signed` and `parse_range_unsigned`…

What would be the use case for them, where in the API implementation would they be used?

CelticMinstrel:
> What would be the use case for them, where in the API implementation would they be used?

This is more of an argument for keeping them separate specifically in the schema, not in the underlying code, but… most of the places that use `range_list` in the schema expect positive numbers, usually coordinates. In fact, I think `value` might be the _only_ one where negative values make sense. (Where is that actually used, anyway? It's the new ability filter, isn't it? I can't seem to find it in the schema…)

But if the schema is going to reject something, it usually makes sense for the game to reject it too, which suggests having separate functions for signed and unsigned ranges.

src/serialization/string_utils.cpp
```diff
@@ -822,47 +823,100 @@ std::vector<std::string> quoted_split(const std::string& val, char c, int flags,
...
+ * Internal common code for parse_range and parse_range_real.
+ *
+ * If str contains two elements and a separator such as "a..b" or "a-b", returns a and b.
+ * Otherwise, returns the original string and std::nullopt.
+ */
+std::pair<std::string, std::optional<std::string>> parse_range_internal_separator(const std::string& str)
+{
+	// If adding additional options to this list, ensure that "-" (if present) is last. Otherwise a
+	// range such as "-2..-1" might be incorrectly split as "-2..", "1".
+	static const auto possible_separators = std::vector<std::string>{"..", "-"};
```
CelticMinstrel, line 838: Probably not a "now" thing but there _might_ be an argument for adding `~` to this list, since it's used in the animation syntax.

CelticMinstrel:
I notice the two parse_range functions have almost identical code and wonder if it's worth splitting out a template helper function… maybe not though since it's _only_ the two cases and an unsigned case would have different code.

stevecotton:
> I notice the two parse_range functions have almost identical code

Erm, yeah. Thanks for your comment, it made me realise that I'd cut-and-pasted `std::stoi` calls into the one that should use doubles. The latest push fixes that and adds another unit test, this time a Lua test as that seems to be the only API that accesses the double version.

CelticMinstrel:
> The latest push fixes that and adds another unit test, this time a Lua test as that seems to be the only API that accesses the double version.

Since this is a C++ API it probably could've been a Boost unit test, but I guess the Lua one will do.

Did you have actual thoughts on the possibility of an unsigned version? Or should we just use the signed version and afterwards separately validate that the range is positive?

stevecotton:
> This is more of an argument for keeping them separate specifically in the schema, not in the underlying code, but… most of the places that use `range_list` in the schema expect positive numbers, usually coordinates. In fact, I think `value` might be the _only_ one where negative values make sense. (Where is that actually used, anyway? It's the new ability filter, isn't it? I can't seem to find it in the schema…)

The new ones are in data/schema/filters/abilities.cfg's `[tag]name="$filter_abilities"`. From there, `value`, `add`, `sub` and maybe `multiply`. Another place where the filters should allow negative values is for `[attack]accuracy=` and `[attack]parry=`.

In terms of implementing a check in C++, I think it would be easier to have a wrapper function that calls the parse-anything one, and then validates the lower bound. Keeping the existing name `in_ranges` for the wrapper and having `in_signed_ranges` for the new behavior makes sense to me.

OTOH, there aren't many places that currently call it, so renaming it to `in_positive_ranges` would also work for me. I'd still suggest using `in_signed_ranges` as the name for the one that accepts negatives.

src/utils/config_filters.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * Checks whether the filter matches the value of @a cfg[@a attribute]. If @a cfg doesn't have that
+ * attribute, assume that an unset value is equivalent to @a def.
+ *
+ * Always returns true if the filter puts no restriction on the value of @a cfg[@a attribute].
+ */
+bool bool_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, bool def);
+bool double_matches_if_present(const config& filter, const config& cfg, const std::string& attribute);
+
+bool int_matches_if_present_positive(const config& filter, const config& cfg, const std::string& attribute);
+bool int_matches_if_present_signed(const config& filter, const config& cfg, const std::string& attribute);
```
stevecotton, line 36: The function names are getting a bit long, I'm open to suggestions.

Also, just noticed that this one still needs documentation.

data/schema/game_config.cfg
```diff
@@ -80,17 +80,54 @@
...
+        [/element]
+        [element]
+            value="\d+\.\.\d+"
+        [/element]
+        [element]
+            value="\d+-infinity"
+        [/element]
+        [element]
+            value="\d+\.\.infinity"
+        [/element]
```
CelticMinstrel, line 130: This could actually just reference `positive_range` directly, since it's identical. That does mean that there are technically two elements that can potentially match a range like "9" but that doesn't seem like much of an issue.
```suggestion
        # If the first number is positive, the second one has to be too
        [element]
            link="positive_range"
        [/element]
```

Also, I believe you said it accepts `-infinity..infinity`, but you're rejecting it at the schema level? Is that intended or an oversight? I could see either way, honestly.

data/schema/game_config.cfg
```diff
@@ -390,8 +427,9 @@
     [/type]
     {SUBST_TYPE coordinate}
     {SUBST_TYPE coordinates}
-    {SUBST_TYPE range_list}
+    {SUBST_TYPE positive_range_list}
     {SUBST_TYPE real_range_list}
+    {SUBST_TYPE signed_range_list}
```
CelticMinstrel, line 432: It's not like it matters, but I'd probably push this up one line to keep the subst types in the same order as the the list types.

data/schema/game_config.cfg
stevecotton: I feel that making a filter with "accept anything" is mistake, but one that isn't worth special casing in either location.

data/schema/game_config.cfg
stevecotton: Can do. Not sure if it's better or worse to reorder the list types into alphabetical order instead.

src/utils/config_filters.hpp
stevecotton: I've gone for a rename to this trio:

* int_matches_if_present
* int_matches_if_present_or_negative
* unsigned_matches_if_present

data/schema/game_config.cfg
stevecotton: Wondering about renaming "positive_range" to "unsigned_range" - it's more accurate, but it's also visually a lot closer to "signed_range".

stevecotton:
@hrubymar10 please would you supply the changes needed for the Xcode files?

I think this is ready to merge, except for the Xcode support.

data/schema/game_config.cfg
CelticMinstrel: I was thinking that too, but I'm not sure. The int types in schema are named `int` and `unsigned` so another possibility would be `int_range` and `unsigned_range`.

gfgtdf:
Okay this has changes quite a bit since my last comment, it changes the main function to parse ranges  (which is iirc afaik used in particular to parse all location ranges)  to accept `..` so i guess the new main question is do we really want that, usually supporting multiple syntaxes makes it rather more difficult than less, because it meas that people (and tools) who read the code now need to know both syntaxes so i am a bit scelpital, .

stevecotton:
@CelticMinstrel thanks for the Xcode file, but it seems to be missing config_filters.cpp from the build for the Boost unit tests. Please would you push again?

stevecotton:
> usually supporting multiple syntaxes makes it rather more difficult than less

It's a good point. `value="-50--1"` isn't that hard to read, it would provide the necessary functionality, and it would only be usable in places that support negative values, so dropping the `..` support from this PR would prevent a new syntax from creeping in elsewhere.

src/serialization/string_utils.cpp
stevecotton: Based on comments, I've removed `..` too, so it's just looking for `-`.

CelticMinstrel:
> > usually supporting multiple syntaxes makes it rather more difficult than less
> 
> It's a good point.

Is it though? _Why_ exactly is it more difficult to support two alternative syntaxes?

src/serialization/string_utils.cpp
```diff
@@ -822,57 +823,102 @@ std::vector<std::string> quoted_split(const std::string& val, char c, int flags,
...
 
+namespace
+{
+/**
+ * Internal common code for parse_range and parse_range_real.
+ *
+ * If str contains two elements and a separator such as "a-b", returns a and b.
+ * Otherwise, returns the original string and std::nullopt.
+ */
+std::pair<std::string, std::optional<std::string>> parse_range_internal_separator(const std::string& str)
```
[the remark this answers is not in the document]
stevecotton: While it could do, any string returned is expected to fit within the short-string optimisation length. I think it'll turn out to be about the same amount of data either way round, even before the compiler optimisations.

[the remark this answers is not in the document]
gfgtdf: > While it could do, any string returned is expected to fit within the short-string optimisation length. I think it'll turn out to be about the same amount of data either way round, even before the compiler optimisations.


Hmm, i really don't know how much the impact is then, i think this is probably something only an actual benchmark will tell us, (which we probably won't do).


>because it can be implemented based on a C function that expects the null-terminator at the end.

From how i understood it, it calls std::strtol which doesn't expect a terminating null character, 

https://en.cppreference.com/w/cpp/string/basic_string/stol

stevecotton:
> Is it though? _Why_ exactly is it more difficult to support two alternative syntaxes?

The point I was agreeing with is that it's more difficult for humans to read. Making the rules more permissive makes it easier for the person initially writing the code, but adds an overhead to the future maintenance, but this feels like it's leading into the same territory as the discussions about whitespace rules.

src/serialization/string_utils.cpp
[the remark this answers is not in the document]
stevecotton: (There's a bit of discussion on IRC after this)

@gfgtdf would you accept the PR as-is? I think any changes to use `from_chars` should be a separate PR.

stevecotton:
The force-push was just to put the Xcode updates in the same commit as the new files.

src/units/unit.cpp
```diff
@@ -1528,10 +1485,10 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 	if(!int_matches_if_present(filter, cfg, "value"))
 		return false;
```
newfrenchy83, line 1486: "value" "multiply" and divide" can also be negative, i don't undestand why you don't use matches_if_present_or_negative for them.

newfrenchy83: oh, excuse me, i trust what int_matches_if_present_or_negative was for support negative numberin range.

src/serialization/string_utils.cpp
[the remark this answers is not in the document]
gfgtdf: I didnt check the whole pr but the comment above is not something that i'd require to be solved before merging, especially since it seems like that creating a std::string was also the beahvior before

PR #7830, opened 2023-08-04
Use '_' instead of '-' as separator in filter_abilites filters

newfrenchy83:
Since '-' is usually used to define a range of numbers but the same sign is used to define a negative number, it is impossible to filter a negative number [filter_abiliti] and similar filter hence the idea of using ' _' instead in this specific case to remedy the problem.

### DIFF
--- a/data/schema/filters/abilities.cfg
+++ b/data/schema/filters/abilities.cfg
@@ -7,11 +7,11 @@
 	{SIMPLE_KEY overwrite_specials string_list}
 	{SIMPLE_KEY apply_to string_list}
 	{SIMPLE_KEY active_on string_list}
-	{SIMPLE_KEY value s_range_list}
-	{SIMPLE_KEY add s_range_list}
-	{SIMPLE_KEY sub s_range_list}
-	{SIMPLE_KEY multiply s_real_range_list}
-	{SIMPLE_KEY divide s_real_range_list}
+	{SIMPLE_KEY value s_range_spe_list}
+	{SIMPLE_KEY add s_range_spe_list}
+	{SIMPLE_KEY sub s_range_spe_list}
+	{SIMPLE_KEY multiply s_real_range_spe_list}
+	{SIMPLE_KEY divide s_real_range_spe_list}
 	{SIMPLE_KEY affect_adjacent s_bool}
 	{SIMPLE_KEY affect_self s_bool}
 	{SIMPLE_KEY affect_allies s_bool}
--- a/data/schema/filters/abilities.cfg
+++ b/data/schema/filters/abilities.cfg
@@ -7,11 +7,11 @@
 	{SIMPLE_KEY overwrite_specials string_list}
 	{SIMPLE_KEY apply_to string_list}
 	{SIMPLE_KEY active_on string_list}
-	{SIMPLE_KEY value s_range_spe_list}
-	{SIMPLE_KEY add s_range_spe_list}
-	{SIMPLE_KEY sub s_range_spe_list}
-	{SIMPLE_KEY multiply s_real_range_spe_list}
-	{SIMPLE_KEY divide s_real_range_spe_list}
+	{SIMPLE_KEY value s_range_tilde_list}
+	{SIMPLE_KEY add s_range_tilde_list}
+	{SIMPLE_KEY sub s_range_tilde_list}
+	{SIMPLE_KEY multiply s_real_range_tilde_list}
+	{SIMPLE_KEY divide s_real_range_tilde_list}
 	{SIMPLE_KEY affect_adjacent s_bool}
 	{SIMPLE_KEY affect_self s_bool}
 	{SIMPLE_KEY affect_allies s_bool}
--- a/data/schema/game_config.cfg
+++ b/data/schema/game_config.cfg
@@ -110,10 +110,10 @@
             value="\d+"
         [/element]
         [element]
-            value="\d+_\d+"
+            value="\d+~\d+"
         [/element]
         [element]
-            value="\d+_infinity"
+            value="\d+~infinity"
         [/element]
     )}
     # definition of real_range_spe_list and s_real_range_spe_list, before macros are expanded
@@ -122,10 +122,10 @@
             value="\d+(\.\d+)?"
         [/element]
         [element]
-            value="\d+(\.\d+)?_\d+(\.\d+)?"
+            value="\d+(\.\d+)?~\d+(\.\d+)?"
         [/element]
         [element]
-            value="\d+(\.\d+)?_infinity"
+            value="\d+(\.\d+)?~infinity"
         [/element]
     )}
     [type]
--- a/data/schema/game_config.cfg
+++ b/data/schema/game_config.cfg
@@ -104,28 +104,40 @@
             value="\d+(\.\d+)?-infinity"
         [/element]
     )}
-    # definition of range_spe_list and s_range_spe_list, before macros are expanded
-    {LIST_TYPE_COMPLEX range_spe (
+    # definition of range_tilde_list and s_range_tilde_list, before macros are expanded
+    {LIST_TYPE_COMPLEX range_tilde (
         [element]
-            value="\d+"
+            value="-?\d+"
         [/element]
         [element]
-            value="\d+~\d+"
+            value="-?\d+~-?\d+"
         [/element]
         [element]
-            value="\d+~infinity"
+            value="-?\d+~infinity"
+        [/element]
+        [element]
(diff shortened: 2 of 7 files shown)


newfrenchy83:
it should reolve https://github.com/wesnoth/wesnoth/pull/7829

gfgtdf:
How about using `..` which would match the syntax used for example by ` [set_variable]rand=`

newfrenchy83:
> How about using `..` which would match the syntax used for example by ` [set_variable]rand=`

using `..`? ok i will try that

> How about using `..` which would match the syntax used for example by ` [set_variable]rand=`

it's done

newfrenchy83:
> > How about using `..` which would match the syntax used for example by ` [set_variable]rand=`
> 
> using `..`? ok i will try that
> 
> > How about using `..` which would match the syntax used for example by ` [set_variable]rand=`
> 
> it's done

impossible in parse_range the separator must have one character only.you prefr what instead : using '_' or perhaps ':' instead?

newfrenchy83:
@gfgtdf i cannot use `..` would you what i use `:`instead?

gfgtdf:
@CelticMinstrel an opinion on what the best separator would be?

CelticMinstrel:
If it must be a single character, then we could use `~` to match up with the animation string expansion syntax.

data/schema/filters/abilities.cfg
```diff
@@ -7,11 +7,11 @@
 	{SIMPLE_KEY overwrite_specials string_list}
 	{SIMPLE_KEY apply_to string_list}
 	{SIMPLE_KEY active_on string_list}
-	{SIMPLE_KEY value s_range_list}
-	{SIMPLE_KEY add s_range_list}
-	{SIMPLE_KEY sub s_range_list}
-	{SIMPLE_KEY multiply s_real_range_list}
-	{SIMPLE_KEY divide s_real_range_list}
+	{SIMPLE_KEY value s_range_spe_list}
```
CelticMinstrel, line 10: What is spe? It looks like a spelling mistake… or is it an acronym?

src/serialization/string_utils.cpp
```diff
@@ -822,9 +822,9 @@ std::vector<std::string> quoted_split(const std::string& val, char c, int flags,
 	return res;
 }
 
-std::pair<int, int> parse_range(const std::string& str)
+std::pair<int, int> parse_range(const std::string& str, bool alter_sep)
 {
-	const std::string::const_iterator dash = std::find(str.begin(), str.end(), '-');
+	const std::string::const_iterator dash = alter_sep ? std::find(str.begin(), str.end(), '_') : std::find(str.begin(), str.end(), '-');
```
CelticMinstrel, line 827: I think rather than having a switch between the two syntaxes, it would make more sense to try both. Try the new one first, and if that fails, try the old one. If both fail, it's probably just a single number.

By the way, if this is the reason you said it has to be a single character, then you're just using the wrong function. You should use `str.find()` instead if you want to search for a substring. That said, making it be multiple characters would also complicate the logic later on, so I'm not specifically recommending you do this. I'll leave it up to you which is better.

CelticMinstrel (commented):
Mostly looks good. Assuming the new schema type is left in, it should probably be renamed, but I'm not sure what to at the moment.

CelticMinstrel:
Oh, also: couldn't the underlying issue be resolved by starting the search at the second character, instead of the first character? It's not valid syntax to start with a range specifier anyway.

data/schema/filters/abilities.cfg
newfrenchy83: spe for special because is a special case, it don't find best for now.

src/serialization/string_utils.cpp
newfrenchy83: except what '-' is also for negative value, i don't like idea to check that for both value and separator for intervalof value.

newfrenchy83:
> Oh, also: couldn't the underlying issue be resolved by starting the search at the second character, instead of the first character? It's not valid syntax to start with a range specifier anyway.

the search of first character is in function used in many function in wesnoth source code, i prefer not touch to that

soliton-:
Unless there is an actual logical/technical reason that neither `-` nor `..` can be used we should not invent another range syntax. I can't tell what the issue with either is.

CelticMinstrel:
`~` is also used already in the game, but not inventing another range syntax definitely sounds like a good idea.

newfrenchy83:
> Unless there is an actual logical/technical reason that neither `-` nor `..` can be used we should not invent another range syntax. I can't tell what the issue with either is.

and if i use "neg" prefix for negative value instead of `-`, coud we use same range syntax?

newfrenchy83:
> > Unless there is an actual logical/technical reason that neither `-` nor `..` can be used we should not invent another range syntax. I can't tell what the issue with either is.
> 
> and if i use "neg" prefix for negative value instead of `-`, coud we use same range syntax?

i pull https://github.com/wesnoth/wesnoth/pull/7831 like another solution to esolve that.

data/schema/filters/abilities.cfg
stevecotton: If we're using `~` as the separator character, I suggest `s_range_tilde_list`. I'd still prefer `..` or even the minus sign (with logic to resolve ambiguity).

CelticMinstrel: For what it's worth, in the regex it would look like this:
```
        [element]
            value="-?\d+"
        [/element]
        [element]
            value="-?\d+--?\d+"
        [/element]
        [element]
            value="-?\d+-infinity"
        [/element]
        [element]
            value="-infinity--?\d+"
        [/element]
        [element]
            value="-infinity-infinity"
        [/element]
```

Actually implementing that in the code is separate issue; and even if we did this, it might be worth making it a separate type, as we might want some range lists to allow negative values and others not to.

newfrenchy83: > If we're using `~` as the separator character, I suggest `s_range_tilde_list`. I'd still prefer `..` or even the minus sign (with logic to resolve ambiguity).

for `..`impossible because two or more character in separator will be reject in Macos. It's between `~` and `_` what the choice willl be done. personally i prefer `~` for esthetical reason and also arlredy used like separator in halo but is not decided yet.

data/schema/game_config.cfg
```diff
@@ -104,6 +104,42 @@
...
+            value="-?\d+~-?\d+"
+        [/element]
+        [element]
+            value="-?\d+~infinity"
+        [/element]
+        [element]
+            value="-infinity~-?\d+"
+        [/element]
+        [element]
+            value="-infinity~infinity"
```
stevecotton, line 122: While it fits the rest of the syntax, I feel "allow any number at all" isn't useful as a filter. I'd suggest avoiding any special cases for it, and do whatever is the least code.

data/schema/filters/abilities.cfg
stevecotton: I had a look at supporting `..`, will push to #7829 when it finishes the local test.

data/schema/game_config.cfg
newfrenchy83: last case removed and test added

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
```diff
@@ -110,6 +112,56 @@
...
+# Actions:
+# Use the common setup from FILTER_ABILITY_TEST.
+# Add an event with a filter matching Alice's drains ability.
+# Alice attacks Bob and then Bob attacks Alice, as defined in FILTER_ABILITY_TEST.
+##
+# Expected end state:
+# The filtered event is triggered exactly once.
+#####
+{GENERIC_UNIT_TEST event_test_filter_ability_neg_value (
+    {FILTER_ABILITY_TEST -25}
```
stevecotton, line 145: `{FILTER_ABILITY_TEST DRAINS_VALUE=-25}`, otherwise the preprocessor complains that this macro doesn't expect an argument.

src/serialization/string_utils.cpp
stevecotton: Another possibility would be "don't accept `-` as a separator if it appears elsewhere in the string", so the old syntax with `-` would only be allowed if both numbers were zero-or-more.

newfrenchy83: you'll already find solution in https://github.com/wesnoth/wesnoth/pull/7829, and i don't see how locate good '-' because each of two number can be have one or more characters in string

PR #7831, opened 2023-08-05
Use 'neg' like prefix for negative value in parse_range

newfrenchy83:
Since '-' is usually used to define a range of numbers but the same sign is used to define a negative number, it is impossible to filter a negative number [filter_abiliti] and similar filter hence the idea of using 'neg' instead for define negative value in parse_range


CelticMinstrel:
My general reaction is "no, why would you want this".

The other PR was better, though I'm not convinced it's the best – it is entirely possible to use `-` as the range separator _and also_ support negative numbers.

PR #7832, opened 2023-08-05
address codeql issue

Pentarctagon:


src/game_launcher.cpp
```diff
@@ -843,16 +846,20 @@ void game_launcher::start_wesnothd()
 	SetEnvironmentVariableA("SDL_STDIO_REDIRECT", "0");
 #endif
 	LOG_GENERAL << "Starting wesnothd: "<< command;
-	if (std::system(command.c_str()) == 0) {
-		// Give server a moment to start up
+	try
+	{
+		bp::spawn(command);
```
CelticMinstrel, line 851: Should be something like:
```cpp
bp::spawn(wesnothd_program, "-c", config, "-d", "-t", "2", "-T", "5")
```

I guess omit the `-d` on Windows? I don't really get why though. And it looks like Windows is using some magic feature to set the console titlebar, which would be lost with the switch to Boost.Process, so I'd suggest making wesnothd set its own title instead.

But basically, the change made here does _not_ address the CodeQL issue.

CelticMinstrel: For what it's worth, setting the title within wesnothd is about this simple:

```cpp
std::string title = "wesnoth server";
#ifdef WIN32
	SetConsoleTitleA(title.c_str());
#else
	cout << "\033]0;" << title << "\007";
#endif
```

Pentarctagon: Where would this go exactly? Just in main()?

soliton-: Yeah, after argument parsing.

src/game_launcher.cpp
```diff
@@ -834,21 +834,15 @@ void game_launcher::start_wesnothd()
 		filesystem::write_file(config, filesystem::read_file(filesystem::get_wml_location("lan_server.cfg")));
 	}
 
+	LOG_GENERAL << "Starting wesnothd";
+	try
+	{
 #ifndef _WIN32
-	std::string command = "\"" + wesnothd_program +"\" -c \"" + config + "\" -d -t 2 -T 5";
+		bp::spawn(wesnothd_program, "-c", "\""+config+"\"", "-d");
```
CelticMinstrel, line 841: Don't surround the config with quotes. That'll almost certainly fail.

src/game_launcher.cpp
```diff
@@ -834,21 +834,15 @@ void game_launcher::start_wesnothd()
...
 #else
-	// start wesnoth as background job
-	std::string command = "cmd /C start \"wesnoth server\" /B \"" + wesnothd_program + "\" -c \"" + config + "\" -t 2 -T 5";
-	// Make sure wesnothd's console output is visible on the console window by
-	// disabling SDL's stdio redirection code for this and future child
-	// processes. No need to bother cleaning this up because it's only
-	// meaningful to SDL applications during pre-main initialization.
-	SetEnvironmentVariableA("SDL_STDIO_REDIRECT", "0");
+		// start wesnoth as background job
+		bp::spawn("cmd", "/C", "start", "/B", wesnothd_program, "-c", "\""+config+"\"");
```
CelticMinstrel, line 844: This doesn't resolve the issue at all. You're still calling the shell to parse a command line. I think it should be fine to just delete the first two arguments and put `"start"` as the first argument, with the caveat that it probably needs to be the _actual path to `start.exe`_ rather than just the name.

src/server/wesnothd/server.cpp
```diff
@@ -3106,5 +3094,9 @@ int main(int argc, char** argv)
 		}
 	}
 
-	return wesnothd::server(port, keep_alive, config_file, min_threads, max_threads).run();
+#ifdef WIN32
+	SetConsoleTitleA("wesnoth server");
```
CelticMinstrel, line 3098: So you decided not to set the title on Linux…?

Pentarctagon: What would be the point of printing that to cout? That's not a title to anything.

Pentarctagon:
I'm going to just close this at this point. Realistically this is a non-issue and it doesn't seem like it's worth the effort to resolve whatever tiny risk it represents - there doesn't appear to be any cross-platform way to either launch a process without a shell or to validate the input, and I'm not going to custom create my own.

src/server/wesnothd/server.cpp
CelticMinstrel: I guess you've ultimately decided to scrap this, but… it's a terminal escape code that instructs the terminal to set its own title. That's also why you'd output it directly to cout instead of to one of the log streams.

Pentarctagon:
To add some details for anyone (including me) looking at this in the future:
* Apparently, for some reason, it seems that Boost.Process uses a shell when given one big command string to run, but doesn't use a shell when passed a command as separate arguments.
* The `start` part of the Windows command is used to put wesnothd's terminal window in the background so the player doesn't see it.
* The `/C` part of the Windows command runs all the remaining text on the line as a command in a shell, which is then (in theory) subject to the same command injection issue that it was when using `std::system` instead of `boost::process::spawn`. Thus, changing to use `boost::process::spawn` doesn't fully solve the issue anyway.
* The two parts of the command that use variables are (1) the wesnothd executable which doesn't represent a security risk because anyone/anything capable of changing the preference value for which executable to run could... just run that executable directly anyway, and (2) the server config path which is the hard coded `lan_server.cfg` file in the user config directory.

As such, unless anyone is opposed, I'm going to go ahead and close the [CodeQL issue](https://github.com/wesnoth/wesnoth/security/code-scanning/6) as not being an actual risk.

CelticMinstrel:
> the server config path which is the hard coded `lan_server.cfg` file in the user config directory

It's not hard-coded actually – it's using standard WML path resolution. I don't think that changes anything about your conclusion though… an add-on wouldn't be able to place a file at the root of the WML filesystem, and the only thing that _might_ possibly be able to override that config file with a different one is placing a `lan_server.cfg` next to your `add-ons` folder… and if you can do that, you can also directly edit the config file, right?

Pentarctagon:
To clarify, I meant only that the filename `lan_server.cfg` is hard coded. The directory that file is in obviously isn't.

More generally, my point is that anything you could get wesnoth to do by doing a command injection when it launches wesnothd is something that could already be done directly anyway since it would mean the person/program doing the injection already has full access to the account running the wesnoth executable - they wouldn't be able to make the changes necessary to do the command injection otherwise.

CelticMinstrel:
I think you're probably right, as long as custom cores definitely can't override it.

Pentarctagon:
What would be the risk of a custom core providing a different `lan_server.cfg`? And how is that related to the command injection risk?

soliton-:
If you can influence what file would be used you could influence what file path gets injected. That is not the case here though since besides the user config dir (which is influenced by an env variable though which is exactly what codeql found as the issue) the location is fixed.

While perhaps there is a way to influence what file is copied to the fixed location (which would be a different issue) the injected path only depends on the user config dir so unless an attacker could somehow set env variables it can't be exploited.

Anyway, there is no need to focus on security implications like a hacker getting access to your data or something. It's simply bad code that breaks in general so it should be fixed.

Pentarctagon:
How does it break in general? Generally, it seems like it's been working fine.

Pentarctagon:
Also it looks like (?) Windows doesn't use any environment variables for determining the path, so this PR's fix might actually be correct after all?

CelticMinstrel:
> What would be the risk of a custom core providing a different `lan_server.cfg`? And how is that related to the command injection risk?

Maybe it's not related. I think I'm just imagining the issue in any case – a custom core doesn't reroot the virtual filesystem around its custom core root, it just loads that directory _instead of_ the mainline core directory. The mainline core directory still exists in the virtual filesystem, and the core can even reference specific files in it if it wants to. So I believe this code will still find the same `lan_server.cfg` even when you have a custom core enabled.

It does kinda make sense for it to be moved to `data/hardwired` to make this explicitly clear, but I don't think it's necessary.

soliton-:
In general meaning under any circumstance. While i'm sure it works the majority of the time it does not work with every path.

soliton-:
The quoting for the posix version is still wrong. the config argument should not contain literal quotes. It might work on windows with cmd.

soliton-:
Also it sounds to me like spawn will run the command in the background so -d would not be necessary.

src/game_launcher.cpp
soliton-: Also need to call `search_path` since `wesnothd_program` isn't always the full path.

src/game_launcher.cpp
```diff
@@ -831,28 +834,26 @@ void game_launcher::start_wesnothd()
...
-	std::string command = "\"" + wesnothd_program +"\" -c \"" + config + "\" -d -t 2 -T 5";
+		bp::spawn(wesnothd_program, "-c", "\""+config+"\"", "-d");
 #else
-	// start wesnoth as background job
-	std::string command = "cmd /C start \"wesnoth server\" /B \"" + wesnothd_program + "\" -c \"" + config + "\" -t 2 -T 5";
-	// Make sure wesnothd's console output is visible on the console window by
-	// disabling SDL's stdio redirection code for this and future child
-	// processes. No need to bother cleaning this up because it's only
-	// meaningful to SDL applications during pre-main initialization.
-	SetEnvironmentVariableA("SDL_STDIO_REDIRECT", "0");
```
soliton-, line 843: Is this not needed anymore?

Pentarctagon:
@soliton- does switching to `boost::process::spawn` solve anything though? That's what I'd like confirmation on, given I've been wrong repeatedly on different things so far.

soliton-:
Since we just want to start wesnothd in the background and forget about it spawn fits from what I read about it. Documentation that I found was not that great to explain if it might also call a shell in some circumstance like I think you mentioned though. That should definitely be avoided.

CelticMinstrel:
It seems that Boost.Process isn't properly documented – nowhere does it say that `bp::spawn` can call a shell, and the example suggests that it's not supposed to (you should use `bp::system` if that's what you want), but from the sounds of it, it seems like it _can_?

There's also the possibility of _not_ using Boost.Process and just copying what open.cpp does – namely, directly calling the underlying POSIX or WinAPI function. I would guess that the Windows version offers some flag that lets you request it to launch in the background or whatever it is you need.

Pentarctagon:
Generally I'd lean towards letting a library like Boost handle cross-platform issues where possible.

CelticMinstrel:
That's assuming it actually offers the features you need though, right?

Pentarctagon:
The feature needed is just "run the process without using a shell", isn't it?

CelticMinstrel:
No, you needed some ability to start it minimized or in the background or something, right? Whatever the `/B` flag to `start` does. If you used CreateProcess directly I bet it has a flag to do that, whereas I don't know that Boost.Process exposes such a feature.

Pentarctagon:
Well, I'm not going to experiment with the underlying Windows API given I don't have a Windows PC to test with.

soliton-:
There is https://www.boost.org/doc/libs/1_82_0/doc/html/boost/process/windows/create_no_window.html

If someone can figure out how to use that to get the same behaviour without using cmd and its start command then that'd be great but barring that it's already a win to not call cmd via a shell.

Pentarctagon:
It looks like that can only be used with `boost::process::child`, which only accepts a single command string rather than the command and its options as separate strings.

CelticMinstrel:
> which only accepts a single command string rather than the command and its options as separate strings.

I don't know where you get this from? The `child` constructor is variadic.

Pentarctagon:
The examples I found didn't use the variadic version to have both strings and non-strings as arguments, but let's see if that compiles or not.

src/game_launcher.cpp
```diff
@@ -831,23 +834,26 @@ void game_launcher::start_wesnothd()
...
-	std::string command = "\"" + wesnothd_program +"\" -c \"" + config + "\" -d";
+		bp::child c(wesnothd_program, "-c", config);
 #else
-	// start wesnoth as background job
-	std::string command = "cmd /C start \"wesnoth server\" /B \"" + wesnothd_program + "\" -c \"" + config + "\"";
+		bp::child c(wesnothd_program, "-c", config, bp::windows::create_no_window);
 #endif
-	LOG_GENERAL << "Starting wesnothd: "<< command;
-	if (std::system(command.c_str()) == 0) {
-		// Give server a moment to start up
```
soliton-, line 842: Comment should be kept to explain the delay call.

src/server/wesnothd/server.cpp
```diff
@@ -3094,5 +3094,12 @@ int main(int argc, char** argv)
 		}
 	}
 
+const std::string title = "wesnoth server";
+#ifdef WIN32
```
soliton-, line 3098:
```suggestion
#ifdef _WIN32
```

src/server/wesnothd/server.cpp
```diff
@@ -3094,5 +3094,12 @@ int main(int argc, char** argv)
...
+	SetConsoleTitleA(title.c_str());
+#else
+	std::cout << "\033]0;" << title << "\007";
```
soliton-, line 3101: I think this was suggested because it looked like a window title was set on windows before but the whole cmd call was only there to not actually show a window so...

I don't mind if we want to do this either way but then we should check `isatty` before writing this out.

Pentarctagon: Honestly I'd prefer to just not set the title - if someone is starting wesnothd from the command line then they'd already know what it's open for.

CelticMinstrel: I personally think that this is nice to have, though not essential. I definitely don't mind calling `isatty` before writing terminal escape codes.

Certainly they'll know what it's open for if they launch it from the command-line… but that's assuming they launch it manually (for example, not some automatic script run on launch) and also that they launched it recently (for example, they didn't leave for a 3 month vacation and come back). Neither of those two examples are cases that absolutely need this, but it's easy to do and, in my opinion, just nice to have.

src/game_launcher.cpp
```diff
@@ -831,23 +835,26 @@ void game_launcher::start_wesnothd()
 		filesystem::write_file(config, filesystem::read_file(filesystem::get_wml_location("lan_server.cfg")));
 	}
 
+	LOG_GENERAL << "Starting wesnothd";
+	try
+	{
 #ifndef _WIN32
-	std::string command = "\"" + wesnothd_program +"\" -c \"" + config + "\" -d";
+		bp::child c(wesnothd_program, "-c", config);
```
CelticMinstrel, line 842: Wasn't the Windows version supposed to have some extra flag or something…?

Pentarctagon: Yeah, but it failed to build, so I'm seeing if it builds at all without the special flag.

CelticMinstrel:
> The examples I found didn't use the variadic version to have both strings and non-strings as arguments, but let's see if that compiles or not.

It looks like Boost.Process uses something like a Named Parameters idiom, so it can tell whether it's a "special parameter" or just a command line parameter by the type.

src/server/wesnothd/server.cpp
CelticMinstrel: Regardless of whether we want to do this, I wouldn't say it necessarily has to be here in this PR.

Pentarctagon:
Alright, it looks like the Windows build issue has been resolved - apparently it's just always been wrong to have cmake do `-D_WIN32_WINNT=_WIN32_WINNT_WIN7`.

src/game_launcher.cpp
CelticMinstrel: Is the `-d` no longer needed?

Pentarctagon: Detaching the child process is equivalent, I think?

src/game_launcher.cpp
Pentarctagon: Using `search_path` results in it not finding the wesnothd executable, whether it's using the full path or not.

src/game_launcher.cpp
soliton-: Yes.

src/game_launcher.cpp
soliton-: Ok, that's weird. That is pretty much the example from the [documentation](https://www.boost.org/doc/libs/1_82_0/doc/html/boost_process/tutorial.html#boost_process.tutorial.launch_mode).
Maybe `search_path` doesn't like it when you give it a path already.

Well, if it works without `search_path` even when it's just wesnothd then that's fine.

Pentarctagon: To clarify (with both `wesnoth` and `wesnothd` in my local repo's root folder):
* If I use `search_path`, then with either just `wesnothd` or the full path it fails to find the executable.
* If I don't use `search_path`, then with either just `wesnothd` or the full path it does find the executable.
* I didn't test it with `wesnothd` being somewhere else that would be found from searching the PATH locations.

CelticMinstrel: Apparently `search_path` only works if you pass it _just_ the program name. Glancing over the code, it appears that maybe `get_program_invocation` is doing the same thing that `search_path` would? I don't know for certain though.

Pentarctagon: I tried passing just the program name by doing `wesnothd_program = filesystem::base_name(wesnothd_program);` and it still failed if I used `search_path`.

CelticMinstrel: Weird… but still, am I correct about what `get_program_invocation` does? If so, then it's correct not to use `search_path` anyway, which I suppose is a good thing if `search_path` just doesn't work.

CelticMinstrel: The main point being that it returns an absolute path, regardless of the method.

So given that, using `search_path` would be wrong even if it functions.

Pentarctagon:
Any further issues before I merge this?

src/game_launcher.cpp
soliton-: You can definitely start wesnothd just by its name if that is the question.

soliton-: Which does not work anymore now.

CelticMinstrel: If that functionality is important, I suppose you could add something like:

```cpp
if(wesnothd_program.find('/') == std::string::npos) {
    wesnothd_program = bp::search_path(wesnothd_program);
}
```

soliton-: Would be good to not require people to pick the wesnothd binary if we can avoid it.

CelticMinstrel: I guess the only problem with that would be, if the binary is in PATH, what are the chances that it's the wrong binary (a different version)?

Pentarctagon: That would've already been a risk though, right?

soliton-: Yes.

Pentarctagon: So I installed the wesnoth-server package (installing the client doesn't also install the server executable), made a copy of `wesnothd-1.16` named as just `wesnothd` so that `which wesnothd` returns `/usr/games/wesnothd`. `bp::search_path` however still does not find anything - `bp::search_path(filesystem::base_name(wesnothd_program)).string()` returns an empty string.

Pentarctagon: Also I'm confused over how the previous implementation would've searched the path? `filesystem::get_program_invocation` certainly wouldn't have done that, or given a value that would've resulted in that being done by `std::system`.

CelticMinstrel: It's because `std::system` invokes the shell, which of _course_ searches the path when running a command by name.

soliton-: It seems fairly irrelevant what the preference file contains. If people put nonsense in there than they can also fix it and put the right thing in there.

The default is determined by argv[0] so it's up to the caller what it puts there. If you call wesnoth with the full path there is probably a good chance the caller also puts the full path in argv[0] and if you just call wesnoth then there is a good chance it'll also just contain wesnoth.

soliton-: Since using argv[0] is not particularly great and we already have `get_exe_dir` which does a better job at finding the wesnoth binary dir (at least on windows and anything with /proc like linux) and is used for the starting point for the wesnothd path picking dialog we could use that in `get_program_invocation` as well. Though it's not quite ideal that it falls back to `get_cwd` a fallback to PATH search (and then to cwd) would be better still. (And better handling for OSes without /proc like macos.)

Otherwise the code CelticMinstrel suggested is pretty much what a shell does.

Pentarctagon: > It's because std::system invokes the shell, which of course searches the path when running a command by name.

Right, but `filesystem::get_program_invocation` will never give you just the program name, it will always give you the full path to where it guesses wesnothd will be. In which case in practice that would never have resulted in a PATH search because it would always have used a full path to the wesnothd executable. So adding functionality to try and find wesnothd via the PATH would be new functionality.

Though I suppose it's not especially relevant either way until `search_path` returns  an actual path to use.

CelticMinstrel: But the other branch, that looks up the path from the preferences, theoretically could return just the program name… though only if the user manually edited the preferences file, I think.

soliton-: `get_program_invocation` will return just the program name as explained above. No need to edit preferences.

soliton-: https://github.com/wesnoth/wesnoth/blob/a6883aabe0845f219cbb813fda3c3450086a6415/src/wesnoth.cpp#L754
`args[0]` there is the same as `argv[0]` which is whatever the caller put there. That might be the full path to wesnoth (relative or absolute) or it might just be wesnoth (or really anything the caller wants but usually one of those options).

These are the log messages I get from the flatpak build:
> 20230823 10:13:54 warning general: Failed to start server wesnothd:
execve failed: No such file or directory

Pentarctagon: So the changes would be to:
* use `filesystem::get_exe_dir()` instead of `game_config::wesnoth_program_dir`.
* update `filesystem::get_exe_dir()` to check `bp::search_path()` results before falling back to `get_cwd()`?

soliton-: Yeah, that sounds good.

Probably still need to use `get_program_invocation` to get the right program name for `bp::search_path()` but I'd remove the `game_config::wesnoth_program_dir` part.

PR #7834, opened 2023-08-07
WC: fix to in 6A_Rural mapgen

gfgtdf:
A random thing i noticed while searching for something else, pr because  not tested yet.


stevecotton:
Adding a comment to that block would be nice. `-- Change cave walls around Dwarvish constructions to mine walls`.

Looks correct, but I haven't tested it either.

PR #7835, opened 2023-08-07
Cut arbitrary hardcoded restrictions of scenario settings

Jonathan-Kelly:
And widen the corresponding 'soft' limits for the multiplayer scenario creation screen.

~~Normally I thoroughly test everything before a pull request, but am not currently set up to build from source, So I cannot test the small, surgical, engine level changes in this one.~~ Edit: now tested and working.

### DIFF
--- a/data/gui/window/mp_create_game.cfg
+++ b/data/gui/window/mp_create_game.cfg
@@ -348,9 +348,9 @@
 					maximum_value_label = _ "unlimited"
 				)}
 
-				{_GUI_MP_CREATE_OPTION_SLIDER ( _ "Village Gold")        "village_gold"        1  5   ( _ "The amount of income each village yields per turn") ()}
+				{_GUI_MP_CREATE_OPTION_SLIDER ( _ "Village Gold")        "village_gold"        0  8   ( _ "The amount of income each village yields per turn") ()}
 
-				{_GUI_MP_CREATE_OPTION_SLIDER ( _ "Village Support")     "village_support"     0  4   ( _ "The number of unit levels each village can support") ()}
+				{_GUI_MP_CREATE_OPTION_SLIDER ( _ "Village Support")     "village_support"     0  8   ( _ "The number of unit levels each village can support") ()}
 
 				{_GUI_MP_CREATE_OPTION_SLIDER ( _ "Experience Modifier") "experience_modifier" 30 200 ( _ "The amount of experience a unit needs to advance") (step_size=10)}
 
--- a/src/map_settings.cpp
+++ b/src/map_settings.cpp
@@ -39,17 +39,17 @@ int get_turns(const std::string& value)
 
 int get_village_gold(const std::string& value, const game_classification* classification)
 {
-	return std::clamp<int>(lexical_cast_default<int>(value, ((classification && !classification->is_normal_mp_game()) ? 1 : 2)), 1, 5);
+	return lexical_cast_default<int>(value, ((classification && !classification->is_normal_mp_game()) ? 1 : 2));
 }
 
 int get_village_support(const std::string& value)
 {
-	return std::clamp<int>(lexical_cast_default<int>(value, 1), 0, 4);
+	return lexical_cast_default<int>(value, 1);
 }
 
 int get_xp_modifier(const std::string& value)
 {
-	return std::clamp<int>(lexical_cast_default<int>(value, 70), 30, 200);
+	return lexical_cast_default<int>(value, 70);
 }
 
 } // end namespace settings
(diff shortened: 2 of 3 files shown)


gfgtdf:
Did you test what happens when you set it to value out of range (like 15?) (with both `force_lock_settings = yes` and `force_lock_settings = no`) ?

gfgtdf:
> I did - when a value is set out of range in the scenario WML ..

I mean after your changes.

Jonathan-Kelly:
I tested the changes to `mp_create_game.cfg` and those work correctly. But I could not test the changes to `map_settings.cpp` as that requires building from source.

gfgtdf:
When it's so unclear whether the changes actually change anything Ina reliable way they obviously need to be tested before they can be merged.

For the actual raising of the settings I don't have a strong opinion, on the one hand more options could be better here, on the other hand these sliders could then always look a bit off when people actually only use the left water of it.

Jonathan-Kelly:
> When it's so unclear whether the changes actually change anything Ina reliable way they obviously need to be tested before they can be merged.

I agree... would you be so kind as to test them for me? : )

All you have to do is set `mp_village_gold=0` in a multiplayer scenario, go to the `Game Settings` tab of the scenario creation screen for that scenario, make sure the `Use Map Settings` checkbox is checked and then see if the `Village Gold` slider reads 0 as it should. If it instead reads 1, then the changes are not working somehow. But I am 80% sure they will.

> these sliders could then always look a bit off when people actually only use the left water of it

I can lower the upper limits of the sliders below 9 if you prefer, like 8 or 6 maybe?

soliton-:
Note that the CI (https://github.com/wesnoth/wesnoth/actions/runs/5787469239 scroll down to Artifacts) builds a windows installer and a flatpak bundle for linux to test.

gfgtdf:
> I agree... would you be so kind as to test them for me? : )

I'm rather lazy with testing myself, so no :(

> All you have to do is set mp_village_gold=0 in a multiplayer


No, this is not what the test as about, the point is that you remove the clamp in map_settings.cpp which clamps the value to the slider range, what should be tested is what happens if you write something like `mp_village_gold=20`

Jonathan-Kelly:
> Note that the CI (https://github.com/wesnoth/wesnoth/actions/runs/5787469239 scroll down to Artifacts) builds a windows installer and a flatpak bundle for linux to test.

Praise be to you, soliton! Wow, this is like getting your own personal nightly builds.

Alright, so now I have tested this PR and it works!

> No, this is not what the test as about, the point is that you remove the clamp in map_settings.cpp which clamps the value to the slider range, what should be tested is what happens if you write something like mp_village_gold=20

Tested this too, and the slider reads 9 while in game the effect is (correctly) 20. This is sub-optimal behavior from the interface but seems much better than the status quo of overriding both the scenario WML and the values specified in `mp_create_game.cfg` with overly tight, hard coded clamps (which have to be separately adjusted to match `mp_create_game.cfg`).

CelticMinstrel:
> Tested this too, and the slider reads 9 while in game the effect is (correctly) 20. This is sub-optimal behavior from the interface

I'd definitely flag this as an issue, but not a blocker. I think the way to handle it would be for the game settings screen to check if the default value is in range, and if not, adjust the slider's range to include the default value.

gfgtdf:
> Tested this too, and the slider reads 9 while in game the effect is (correctly) 20

Is this fir force_lock_settings=yes or no ?

Jonathan-Kelly:
> I'd definitely flag this as an issue, but not a blocker. I think the way to handle it would be for the game settings screen to check if the default value is in range, and if not, adjust the slider's range to include the default value.

That seems best.

> Is this fir force_lock_settings=yes or no ?

Both, the effect (on the interface) is the same with both `yes` and `no`.

Also, you should know that apparently `force_lock_settings=yes` is already broken on both 1.16 and 1.17. Any scenario with it set `yes` starts with no leaders and an immediate "you are defeated" message. At least on Windows 10.

gfgtdf:
> Also, I must report that apparently force_lock_settings=yes is already broken on both 1.16 and 1.17

Hmm i think Also, force_lock_settings implies leader/faction lock so when the side has no leader it gets no leader, not sure whether  i'd call it a bug

Jonathan-Kelly:
> so when the side has no leader defines in wml, it gets no leader, not sure whether i'd call it a bug

If the wml does not define a leader then the default should still be applied (random leader from faction leader list), as the intent of a lock is to override player choice, not to override default settings. Especially since this lock is an attribute of `[multiplayer]` and nearly all multiplayer scenarios (purposefully) do not set player leaders.

But in any case, this pull request is unrelated to force_lock_settings.

Jonathan-Kelly:
@stevecotton Do you see any issues with this small change or can it get your approval?

stevecotton:
Looks reasonable to me, as gfgtdf and celmin seem happy.

Jonathan-Kelly:
Okay, good, Sounds like this is ready to merge then, if anyone wants to to push that nice green merge button : )

Issue #7654, opened 2023-05-23
--preprocess ignores specified file if it does not end in .cfg


soliton-:
`wesnoth --preprocess file .` prints `preprocessing specified resource: file ...` but does not actually process the file (for directories it's a different matter).

Wesnoth should either print a warning/error or process the file. Not processing without any hint is confusing.

CelticMinstrel:
Given that we have the preferences file which is WML but doesn't have an extension, I'd say it should process the file.

ekjotsinghmakhija:
Is anyone working on this issue? I want to work on this issue as it would be great for me as it is a `Good First Issue`.

soliton-:
I don't think anyone is working on this. Feel free to.

PR #7838, opened 2023-08-09
deleted unnecessary check

Welrub:
The problem in #7654 was that the preprocessor ignored files with an extension other than .cfg. Now it will preprocess all files

### DIFF
--- a/src/serialization/preprocessor.cpp
+++ b/src/serialization/preprocessor.cpp
@@ -614,8 +614,7 @@ public:
 			const std::string& name = *(pos_++);
 			unsigned sz = name.size();
 
-			// Use reverse iterator to optimize testing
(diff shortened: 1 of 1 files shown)


soliton-:
The goal is not to change the preprocessor. We want to just change the preprocess command line option.

soliton-:
https://github.com/wesnoth/wesnoth/blob/4a9acd6e06d82fe070d0ffd23311c0ea40785ca1/src/serialization/preprocessor.cpp#L1774
Presumably this check needs to be disabled but only when called by the preprocess option logic so probably needs a new argument for that.

gfgtdf:
From how i read the code this should only be reached when iterting over a directory (`preprocessor_file::files_` is never filled unless `is_directory_` is true), so i don't understand yet how this can have an effect on handling a single file with the preprocess option.

stevecotton:
Closing, replaced by #8080.

PR #7839, opened 2023-08-10
persist last chosen add-on in the preferences

Pentarctagon:


CelticMinstrel:
I'm a little unsure about this. It seems like it might be surprising if you open the map editor to work on Addon C and find that it instead opened Addon A because that's the last one you worked on.

src/gui/dialogs/editor/choose_addon.cpp
```diff
@@ -69,28 +73,39 @@ void editor_choose_addon::populate_list(bool show_all)
...
 	for(const std::string& dir : dirs) {
 		if((show_all || filesystem::file_exists(filesystem::get_addons_dir() + "/" + dir + "/_server.pbl"))
 			&& filesystem::file_exists(filesystem::get_addons_dir() + "/" + dir + "/_main.cfg")) {
 			const widget_data& entry{
 				{"existing_addon_id", widget_item{{"label", dir}}},
 			};
 			existing_addons.add_row(entry);
+			i++;
+			if(dir == preferences::editor_chosen_addon()) {
+				selected_row = i;
```
stevecotton, line 103: Could you remove `i` completely by using `selected_row = existing_addons.get_item_count()` here?

src/preferences/general.hpp
```diff
@@ -326,4 +326,7 @@ namespace preferences {
 	 */
 	void set_sub_achievement(const std::string& content_for, const std::string& id, const std::string& sub_id);
 
+	void set_editor_chosen_addon(const std::string& addon_id);
```
stevecotton, line 329: Documentation would be good, because this is a general file where people will be reading it while thinking about other contexts than the new addon-aware editor.

src/gui/dialogs/editor/choose_addon.cpp
```diff
@@ -47,11 +48,14 @@ void editor_choose_addon::post_show(window& win)
 
 	if(selected_row == 0) {
 		addon_id_ = "///newaddon///";
+		preferences::set_editor_chosen_addon("");
```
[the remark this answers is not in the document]
Pentarctagon: Because it's the first option anyway.

src/gui/dialogs/editor/choose_addon.cpp
Pentarctagon: Almost; needed to subtract 1 from it.

src/preferences/general.hpp
Pentarctagon: Added.

stevecotton: Looks good, and I feel the docs for `editor_chosen_addon` cover what I wanted for both functions. I'm happy for the PR to merge as-is.

Suggestion for the setter: The add-on chosen when the editor was last opened (or changed to later within the editor). Will be left set after exiting the editor, so that it can be used as a default next time.

stevecotton (approved):

src/preferences/general.hpp
Pentarctagon: I'm not sure what you mean by
> or changed to later within the editor

stevecotton: I'm assuming the editor could have a way to select a different add-on without exiting and reopening the editor.

Pentarctagon: It intentionally does not.

PR #7840, opened 2023-08-11
fix self-detecting unit issue in abilities [affect_adjacent] and [fiter_adjacent]

newfrenchy83:
…

when selecting a unit with an active ability only when adjacent to another, it marks active when the cursor is placed next to it

and with https://github.com/wesnoth/wesnoth/pull/7805 it's the problem of displaying the halo with an ability that only affects adjacent units, in this case the owner briefly displays the halo when he moves to side of a unit meeting the conditions


stevecotton:
Looks good to me.

PR #7844, opened 2023-08-13
1.17 various troll sprite things

doofus-01:
Great troll still needs its attack animation, at the very least.


cooljeanius (approved):

Hejnewar:
Looks great but if possible I would like to see slightly darker shadows on level 3 so that it is more similar to level 2 and its muscles are EVEN MORE exposed and visible. 3:)

doofus-01:
> Looks great but if possible I would like to see slightly darker shadows on level 3 so that it is more similar to level 2 and its muscles are EVEN MORE exposed and visible. 3:)

I was actually thinking the trolls are going too far into boy-toy territory (meaning bare, gym-buff torso for sex appeal, but not real strength), and should go more toward rocks instead of muscles.  Higher contrast & shadows works with either path though.  The portraits would work with either as well.

doofus-01:
If this passes CI, I'll commit tomorrow, let me know if there is any glaring issue.

cooljeanius (approved):

PR #7855, opened 2023-08-17
Don't show the add-on selection dialog on editor start.

### LINKED ISSUE #7810: New Editor (#7719) breaks quick reload (F5)

Pentarctagon:
Fixes #7810

### DIFF
--- a/data/gui/window/editor_choose_addon.cfg
+++ b/data/gui/window/editor_choose_addon.cfg
@@ -3,6 +3,7 @@
 ### Definition of the window to let a player choose an add-on ID in the editor
 ###
 
+# TODO: add label explaining what this is for/what it does
 [window]
 	id = "editor_choose_addon"
 	description = "Add-on editor ID selection dialog"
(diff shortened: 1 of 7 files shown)


Pentarctagon:
This will be merged prior to the release on 8/20 barring any issues.

PR #7856, opened 2023-08-17
give to abilitie support of halo or overlay

newfrenchy83:
generalise halo_image and overlay_image to all abilities for 1.19 version and later.
i put that now for let you review my work.

PR #7865, opened 2023-08-21
add a 'replacement_type' and 'alternative_type' attribute in [damage] special or ability who modify the type of attack used

newfrenchy83:
At the suggestion of @stevecotton, I propose a special 'type' and 'added_type' attribute in [damage] capable of modifying the type of attack used when the conditions are met.

### DIFF
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_damage_type.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_damage_type.cfg
@@ -1,0 +1,316 @@
+#textdomain wesnoth-test
+
+#####
+# API(s) being tested: [damage]replacement_type=
+##
+# Actions:
+# Give both Alice and Bob 100% chance to hit.
+# Give Bob one [damage] with replacement_type=fire and two [damage] with replacement_type=cold
+# change resistance of Bob to arcane to 50% and fire to -100%.
+# Give Alice one [damage] with replacement_type=cold, two [damage] with replacement_type=arcane and three with replacement_type=fire
+# and change Alice resistance to cold to -100% and fire to 50%.
+# Move Alice next to Bob, and have Alice attack Bob.
+##
+# Expected end state:
+# Alice attack with fire and Bob use cold.
+#####
+{GENERIC_UNIT_TEST "damage_type_test" (
+    [event]
+        name=start
+        [modify_unit]
+            [filter]
+            [/filter]
+            max_hitpoints=100
+            hitpoints=100
+            attacks_left=1
+        [/modify_unit]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=resistance
+                replace=yes
+                [resistance]
+                    arcane=50
+                    fire=200
+                [/resistance]
+            [/effect]
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [attacks]
+                        value=1
+                    [/attacks]
+                    [damage]
+                        value=12
+                    [/damage]
+                    [damage]
+                        replacement_type=fire
+                    [/damage]
+                    [damage]
+                        replacement_type=cold
+                    [/damage]
+                    [damage]
+                        replacement_type=cold
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                    [/chance_to_hit]
+                [/set_specials]
+            [/effect]
+            [filter]
+                id=bob
+            [/filter]
+        [/object]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=resistance
+                replace=yes
+                [resistance]
+                    cold=200
+                    fire=50
+                [/resistance]
+            [/effect]
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [attacks]
+                        value=1
+                    [/attacks]
+                    [damage]
+                        value=12
+                    [/damage]
+                    [damage]
+                        replacement_type=cold
+                    [/damage]
+                    [damage]
+                        replacement_type=arcane
+                    [/damage]
+                    [damage]
+                        replacement_type=arcane
+                    [/damage]
+                    [damage]
+                        replacement_type=fire
+                    [/damage]
+                    [damage]
+                        replacement_type=fire
+                    [/damage]
+                    [damage]
+                        replacement_type=fire
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                    [/chance_to_hit]
+                [/set_specials]
+            [/effect]
+            [filter]
+                id=alice
+            [/filter]
+        [/object]
+
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+            kill=yes
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        [unstore_unit]
+            variable=a
+            find_vacant=yes
+            x,y=$b.x,$b.y
+        [/unstore_unit]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+
+        [do_command]
+            [attack]
+                weapon=0
+                defender_weapon=0
+                [source]
+                    x,y=$a.x,$a.y
+                [/source]
+                [destination]
+                    x,y=$b.x,$b.y
+                [/destination]
+            [/attack]
+        [/do_command]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        #damage without modification are 12, if test fail hitpoints !=76
+        #if succed then damage by alice 24(bob vulnerable to fire and fire prioritized)
+        #if succed then damage by bob 24(alice vulnerable to cold, cold priority is 1)
+        {ASSERT ({VARIABLE_CONDITIONAL a.hitpoints equals 76})}
+        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 76})}
+        {SUCCEED}
+    [/event]
+)}
+
+#####
+# API(s) being tested: [damage]alternative_type=
+##
+# Actions:
+# Give both Alice and Bob 100% chance to hit.
+# Give Bob one [damage] with alternative_type=cold
+# change resistance Bob to blade to -100% and fire to 0%.
+# Give Alice one [damage] with alternative_type=fire
+# and change resistance to cold to -100% and blade to 0%.
+# Move Alice next to Bob, and have Alice attack Bob.
+##
+# Expected end state:
+# Alice attack with blade and Bob use cold.
+#####
+{GENERIC_UNIT_TEST "damage_secondary_type_test" (
+    [event]
+        name=start
+        [modify_unit]
+            [filter]
+            [/filter]
+            max_hitpoints=100
+            hitpoints=100
+            attacks_left=1
+        [/modify_unit]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=resistance
+                replace=yes
+                [resistance]
+                    blade=200
+                    fire=100
+                [/resistance]
+            [/effect]
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [attacks]
+                        value=1
+                    [/attacks]
+                    [damage]
+                        value=12
+                    [/damage]
+                    [damage]
+                        alternative_type=cold
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                    [/chance_to_hit]
+                [/set_specials]
+            [/effect]
+            [filter]
+                id=bob
+            [/filter]
+        [/object]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=resistance
+                replace=yes
+                [resistance]
+                    cold=200
+                    blade=100
+                [/resistance]
+            [/effect]
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [attacks]
+                        value=1
+                    [/attacks]
+                    [damage]
+                        value=12
+                    [/damage]
+                    [damage]
+                        alternative_type=fire
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                    [/chance_to_hit]
+                [/set_specials]
+            [/effect]
+            [filter]
+                id=alice
+            [/filter]
+        [/object]
+
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
+            kill=yes
+        [/store_unit]
+        [store_unit]
+            [filter]
+                id=bob
+            [/filter]
+            variable=b
+        [/store_unit]
+        [unstore_unit]
+            variable=a
+            find_vacant=yes
+            x,y=$b.x,$b.y
+        [/unstore_unit]
+        [store_unit]
+            [filter]
+                id=alice
+            [/filter]
+            variable=a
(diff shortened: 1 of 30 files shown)


newfrenchy83:

newfrenchy83:
@CelticMinstrel @stevecotton In current implementation of [damage]type=
I modified the filter [filter_weapon]type= to only detect the original attack type in special weapons in order to avoid infinite loop problems and thus be able to use the original attack type as a filter in these case, and use 'changed_type' instead to filter the type used (adversary's attack) but for [event][filter_attack]type= it is the type used which is filtered and therefore be the type modified by [damage]type= when active, either the original type (unit_type or changed by [effect]) in others.

What seems most practical to you? Does the current operation described above, or limit 'type' to the original attack type in all filters and use 'changed_type' for the active type?

nemaara:
Can you remove the EI ones? I can't comment on the code so someone else will have to do that.

newfrenchy83:
> Can you remove the EI ones? I can't comment on the code so someone else will have to do that.

done

Pentarctagon:
@newfrenchy83 I'm going to start taking a look at your PRs hopefully starting tomorrow, however I'm not familiar with this area of code and there's only a month until the feature freeze, so it's unlikely that all of them will be able to be merged for 1.18. Given that, in what order would you prefer that I review your PRs?

newfrenchy83:
> @newfrenchy83 I'm going to start taking a look at your PRs hopefully starting tomorrow, however I'm not familiar with this area of code and there's only a month until the feature freeze, so it's unlikely that all of them will be able to be merged for 1.18. Given that, in what order would you prefer that I review your PRs?

https://github.com/wesnoth/wesnoth/pull/7856, this pr, https://github.com/wesnoth/wesnoth/pull/7746, https://github.com/wesnoth/wesnoth/pull/6968 and https://github.com/wesnoth/wesnoth/pull/7814.

stevecotton:
> #7856, this pr, #7746, #6968 and #7814.

Should the first one be #7805 instead of #7856?

Sorry that I've lacked the mental capacity to work on these, and that I didn't inform @Pentarctagon when I should have done.

newfrenchy83:
> > #7856, this pr, #7746, #6968 and #7814.
> 
> Should the first one be #7805 instead of #7856?
> 
> Sorry that I've lacked the mental capacity to work on these, and that I didn't inform @Pentarctagon when I should have done.

yes, but I would have preferred to link the halos to all types of abilities in addition to unit_type halo from version 1.18 instead of having to carry over to 1.19 (and therefore to the stable version 1.20), and I would like us to ' explains the problem, I thought about using a different macro for illumination in the mainline and ABILITY_ILLIMINATION doesn't call halo by default, and those who continue to use unit_type halo should not be penalized. and seeing that in case of acceptance before the freeze, then new code will be placed for the stable version and 1.19 from the first release.

and, @Pentarctagon  I prefer to reorganize the list as follows: https://github.com/wesnoth/wesnoth/pull/7805 (https://github.com/wesnoth/wesnoth/pull/7856 not accept for 1.18), https://github.com/wesnoth/wesnoth/pull/7746, this pr, https://github.com/wesnoth/wesnoth/pull/6968 and https://github.com/wesnoth/wesnoth/pull/7814.

Pentarctagon:
Alright, so I will (probably) sequentially review in the following order:
* #7805 
* #7746 
* #7865 
* #6968 
* #7814 

I may not get to all of these prior to 1.17.22 however.

Pentarctagon:
Why did 5 of your PRs get force pushed at about the same time?

newfrenchy83:
> Why did 5 of your PRs get force pushed at about the same time?

I regularly do local updates of these branches (some have been there for more than a year) to ensure that there is no conflict with the master branch and around once a month, I force push in requests, like today

Pentarctagon:
Is there a difference between how this would be used vs #7003?

newfrenchy83:
> Is there a difference between how this would be used vs #7003?

no difference in terms of operation but I hesitate between adding these attributes in [damage] or using [damage_type]

newfrenchy83:
@Pentarctagon I will now explain the addition of 'type' and 'added_type' in [damage]

the damage_type() function returns a pair of strings:
'type' is used to replace an attack type with the one specified when 'damage' is active, for example if a sword becomes dull or covered in ice due to fairly intense wet cold (snowy or frozen terrain) , we can give it the 'impact' type associated with a division by two of its attack power. We can also imagine an electric weapon that locks in water for safety.

'added_type' returns a type of attack which combines with the original type in a semi-conservative manner, when the enemy is more sensitive to the added type than to the original type, it is the first which is used, otherwise it 'is the original type, and using an 'added_type' rather than formula filtering allows both types to be displayed in the UI even when the original type is chosen.

Moreover, the UI is modified to display the type of attack in the combat preview window; while the unit bar can for each attack display the type added when it exists.

The two chains are chosen according to two criteria, the alphabetical order and the number of copies of each chain in the vector, if each type is present in a single copy, it is the alphabetical order which takes precedence, otherwise it is the type with the highest copy count is returned and the 'type' and 'added_type' are processed into two different vectors to be selected independently.

Finally I added an application of this second case in the Utbs campaign with the recipient of the vial of holy water who will now use a combined blade/pierce weapon with arcane

nemaara:
Yes the changes to UtBS are fine.

newfrenchy83:
> @nemaara are you good with the changes to UtBS?
> 
> @newfrenchy83 given what was said previously, should #7003 be closed in favor of this PR?

yes

Pentarctagon:
What is `modified_damage`?

newfrenchy83:
> What is `modified_damage`?

i removed this, it will be for later, in another PR.

newfrenchy83:
@nemaara do you accept this https://github.com/wesnoth/wesnoth/pull/7865/commits/cf0797bde712653be1883da280d2b8fc0cac0e2a in this PR?

Pentarctagon:
Can you add a screenshot of that the UI looks like?

---

Also, the plan for right now is:
* Have #7924 (hopefully) merged for 1.17.22.
* Get this PR merged for 1.17.23. Yes, this breaks the string/feature freeze a bit, but the [filter_ability]/[filter_ability_active] changes other devs want to do are already going to be breaking feature freeze in this area of the engine anyway.
* #7814 and #6968 will need to wait for 1.19. Given that's the case, please don't force push to either of those PRs until 1.18.0 is released - doing that takes up CI time.

newfrenchy83:
> Can you add a screenshot of that the UI looks like?
> 
here are the first 3 captures for the added_type=arcane case:
added_type_report
added_type_attack_prediction
added_type_damage_calculation

we see that here the two types are shown simultaneously to remind us that the weapon combines two types (blade since it is a sword and arcane because of the holy water)

The next case concerns the use of type to convert a sword into a fire sword but only when the opponent is undead.:


type_fire_undead_only

type_fire_undead_attack

Here the blade is replaced by fire when the special is active, which is only the case when facing an undead, it is therefore blade outside of combat and fire in attack prediction if the opponent is undead (fire is also specified in the name of the special but that might not be the case)

Pentarctagon:
I think I'd prefer the attributes be renamed:
* `type` to `replacement_type`
* `added_type` to `alternative_type`

Also, rather than determining the priority of a damage type by the number of times it appears in a [damage] tag, would it make more sense to add a `type_priority` attribute?

newfrenchy83:
> I think I'd prefer the attributes be renamed:
> 
> * `type` to `replacement_type`
> * `added_type` to `alternative_type`
> 
> Also, rather than determining the priority of a damage type by the number of times it appears in a [damage] tag, would it make more sense to add a `type_priority` attribute?

Overwrite_specials with priority can be used instead of create à priority attribute. But its okay for renaming

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_damage_type.cfg
```diff
@@ -0,0 +1,287 @@
+
+{GENERIC_UNIT_TEST "damage_type_test" (
```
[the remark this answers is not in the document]
newfrenchy83: done

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_damage_type.cfg
```diff
@@ -0,0 +1,314 @@
+#####
+# API(s) being tested: Test of replacement_type in [damage] and give priority of greater number of same type used.
```
Pentarctagon, line 2: Should be just `[damage]replacement_type`.

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_damage_type.cfg
```diff
@@ -0,0 +1,314 @@
...
+        #if succed then damage by alice 24(bob vulnerable to fire and fire prioritized)
+        #if succed then damage by bob 24(alice vulnerable to cold, cold priority is 1)
+        {ASSERT ({VARIABLE_CONDITIONAL a.hitpoints equals 76})}
+        {ASSERT ({VARIABLE_CONDITIONAL b.hitpoints equals 76})}
+        {SUCCEED}
+    [/event]
+)}
+
+#####
+# API(s) being tested: Test of alternative_type used when doing more damage what attack type
```
Pentarctagon, line 170: Should be just `[damage]alternative_type`.

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/special_damage_type.cfg
```diff
@@ -0,0 +1,314 @@
+#####
+# API(s) being tested: [damage]replacement_type.
```
Pentarctagon, line 2: No period either - this isn't a sentence.

newfrenchy83: i don't understand

src/units/abilities.cpp
```diff
@@ -1168,6 +1168,48 @@ void attack_type::modified_attacks(unsigned & min_attacks,
 	}
 }
 
+//Functions used for change damage_type list with damage
+static std::vector<std::string> vector_type(unit_ability_list& abil_list, const std::string& type)
```
Pentarctagon, line 1172: `abil_list` can be const here. Also, please change the function name to `damage_type_list`.

src/units/abilities.cpp
```diff
@@ -1168,6 +1168,48 @@ void attack_type::modified_attacks(unsigned & min_attacks,
...
+		if(type_list.size() >= 3){
+			std::unordered_map<std::string, unsigned int> type_count;
+			for( const std::string& character : type_list ){
+				type_count[character]++;
+			}
+			std::sort( std::begin( type_list ) , std::end( type_list ) , [&]( const std::string& rhs , const std::string& lhs ){
+				return type_count[lhs] < type_count[rhs];
+			});
+		}
+	}
```
Pentarctagon, line 1191: This should be:
```
	if(type_list.size() == 2){
		std::sort(type_list.begin(), type_list.end());
	} else if(type_list.size() >= 3){
		std::unordered_map<std::string, unsigned int> type_count;
		for( const std::string& character : type_list ){
			type_count[character]++;
		}
		std::sort( std::begin( type_list ) , std::end( type_list ) , [&]( const std::string& rhs , const std::string& lhs ){
			return type_count[lhs] < type_count[rhs];
		});
	}
```
otherwise you sort all lists with 3 or more elements twice.

src/gui/dialogs/unit_attack.cpp
```diff
@@ -115,6 +115,25 @@ void unit_attack::pre_show(window& window)
 			attacker_itor_->get_location(), false, attacker.weapon
 		);
 
+		std::pair<std::string, std::string> types = attacker_weapon.damage_type();
+		std::string attw_type_bis = types.second;
```
Pentarctagon, line 119: What does the `_bis` suffix mean?

src/movetype.cpp
```diff
@@ -743,7 +743,12 @@ utils::string_map_res movetype::resistances::damage_table() const
  */
 int movetype::resistances::resistance_against(const attack_type & attack) const
 {
-	return cfg_[attack.type()].to_int(100);
+	std::pair<std::string, std::string> types = attack.damage_type();
+	int res = cfg_[types.first].to_int(100);
```
Pentarctagon, line 747: `cfg_[types.first].to_int(100);` should be replaced with using the `resistance_against` method directly below this one.

src/movetype.cpp
```diff
@@ -743,7 +743,12 @@ utils::string_map_res movetype::resistances::damage_table() const
...
+	if(!(types.second).empty()){
+		res = std::max(res, cfg_[types.second].to_int(100));
```
Pentarctagon, line 749: Same here.

src/units/abilities.cpp
newfrenchy83: done.

src/gui/dialogs/unit_attack.cpp
newfrenchy83: relaced by _second suffix

src/movetype.cpp
newfrenchy83: replaced

src/movetype.cpp
newfrenchy83: here also

src/units/abilities.cpp
```diff
@@ -1168,6 +1168,47 @@ void attack_type::modified_attacks(unsigned & min_attacks,
...
+	if(type_list.size() == 2){
+		std::sort(type_list.begin(), type_list.end());
+	} else if(type_list.size() >= 3){
+		std::unordered_map<std::string, unsigned int> type_count;
+		for( const std::string& character : type_list ){
+			type_count[character]++;
+		}
+		std::sort( std::begin( type_list ) , std::end( type_list ) , [&]( const std::string& rhs , const std::string& lhs ){
+			return type_count[lhs] < type_count[rhs];
+		});
```
Pentarctagon, line 1189: Does this sort correctly if there are two damage types both with two instances? For example:
```
[damage]
type=cold
[/damage]
[damage]
type=cold
[/damage]
[damage]
type=arcane
[/damage]
[damage]
type=arcane
[/damage]
```
?

newfrenchy83: that's why I implemented the loops if one inside the other, moreover the alphabetical classification must be done when the vector contains two or more elements, not simply two elements. I need to revert the changes you requested on this to resolve this.

newfrenchy83: change reverted.
by first carrying out the alphabetical reorganization we ensure that if two cold and two fire exist in the vector cold are already placed before fire before the numerical sorting, and if these two elements have the greatest number of occurrences in the vector then logically all cold should be placed at the head of the vector before fire is classified in turn. The relative alphabetical order of the elements with the greatest number of copies should then be preserved.

Pentarctagon: Alright, my bad.

soliton-:
There appears to have been no documentation added to the wiki so far.

newfrenchy83:
> There appears to have been no documentation added to the wiki so far.

added here: https://wiki.wesnoth.org/AbilitiesWML#Extra_keys_used_by_the_.5Bdamage.5D_special_.28Version_1.17.23_and_later_only.29

CelticMinstrel:
The FilterWML documentation does not make any sense. I also don't understand why you'd need new keys in FilterWML. If the weapon has a damage type of fire, no matter how it got that type, shouldn't `type=fire` match that weapon? Why should it matter whether it's fire because it was defined as a fire weapon, or because it's under the influence of an ability that converts it to a fire weapon, or because it's under the influence of an ability that adds fire damage to it?

And if we've decided that weapons can have more than one damage type, why not three damage types? Or four damage types? If you can add an extra damage type with an ability, why not define the weapon with two damage types from the start?

I don't think the concept of this PR is bad. I like the idea of having mixed damage types, and conditional damage types are also a great idea for a special ability. But it seems to me that it wasn't very well thought out at all.

PR #7869, opened 2023-08-25
Replace idiom 'by hand' with its actual meaning

Wedge009:
While native Anglophones should know what 'by hand' means in these contexts, I think it's clearer for non-native readers and translators to use its meaning of non-automatic, non-machine -> manual processes.

### DIFF
--- a/data/core/editor/help.cfg
+++ b/data/core/editor/help.cfg
@@ -412,9 +412,9 @@ Loading a .cfg file has different results depending on the contents of the .cfg 
 [topic]
     id=editor_separate_events_file
     title= _ "Using a separate file for WML events"
-    text= _ <<When loading a .cfg file, the scenario editor understands files created by the scenario editor, but is likely to have difficulty with files that have been edited by hand.
+    text= _ <<When loading a .cfg file, the scenario editor understands files created by the scenario editor, but is likely to have difficulty with files that have been edited manually.
 
-One option is to create a separate WML file, also with the .cfg extension, which uses the WML preprocessor to include the editor-created file. This separate file contains both the [scenario] tag and any hand-edited WML such as events. With this workflow, the add-on’s file structure could look like this:
+One option is to create a separate WML file, also with the .cfg extension, which uses the WML preprocessor to include the editor-created file. This separate file contains both the [scenario] tag and any manually-edited WML such as events. With this workflow, the add-on’s file structure could look like this:
 
 <header>text='Example'</header>
 If your add-on will only be used on 1.18 and later, it is instead recommended to use the new include_file attribute to load a .cfg file containing additional WML into the scenario.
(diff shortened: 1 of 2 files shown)


CelticMinstrel:
It's kinda funny, because although I think you're probably right, the word "manually" is directly derived from the Latin for "hand".

doofus-01:
_mano_ as well.  I don't pretend to know what the world thinks, but in the western hemisphere at least, there shouldn't be that much confusion.

CelticMinstrel:
> in the western hemisphere at least

That's exactly Wedge's point. _Only_ in the Western hemisphere (or more precisely, in Romance languages, or at most Indo-European languages) is this an obvious connection.

doofus-01:
Yeah, well, given the alphabet and characters we're using, that's always going to be an issue.  Let's nit-pick everything that might be Euro-centric, why single this comment out?

(To be clear, I've got no objection to changing the strings, but I think this is silly, and I'd be frustrated if it was blocking something I'd been working on.)

Wedge009:
I'm just trying to follow the same mind-set that @stevecotton does with all the `#po` hints. Is it really silly to try to make Wesnoth accessible to non-native readers? Why bother with translation efforts at all, then? Let's just make everything US-centric like most things already are.

I only noticed this when reviewing the translation strings recently and it occurred to me that this 'by hand' idiom isn't necessarily clear to someone whose understanding and command of the English language might not be as strong. Sure, there are probably plenty of instances where changing things to accommodate non-native readers might be considered excessive, but I honestly don't think this is one of them.

I also feel like I've been receiving a lot of criticism for trying to improve things in terms of the nuances of prose and how Wesnoth communicates. I understand genuinely constructive criticism and discussion is a good thing. But to be dismissed off-hand as a 'lunatic' or 'silly' is rather discouraging.

And for the record, I also have a Euro-centric background - at least in terms of language and culture. So it's not as though this suggestion is coming from someone 'outside' of that.

doofus-01:
> I also feel like I've been receiving a lot of criticism for trying to improve things in terms of the nuances of prose and how Wesnoth communicates. I understand genuinely constructive criticism and discussion is a good thing. But to be dismissed off-hand as a 'lunatic' or 'silly' is rather discouraging.

Do what you think is right, maybe string changes such as this really do make translations easier.  It would be good to hear from a (non-western) translator.  If these corrections are not blocking anything, I'll bite my tongue in the future.

Zireael07:
I admit I'm confused. "Hand" is a very common English word, and the preposition "by" makes the whole phrase understandable whatever your native language.

"Manual" or "manually" however is not a Germanic word, it's of Latin origin and somewhat more obscure.

CelticMinstrel:
> whatever your native language

Are you _sure_ this is true for every language in the world? It is certainly a somewhat idiomatic usage. You're not literally using your hands to edit the file (okay, technically your hands are involved, but that would also be the case if you're using the map editor, so it's not a condition that distinguishes "by hand" from "not by hand"). It might be obvious in a lot of languages that have a similar idiomatic usage for this meaning. But there might be a lot of languages which use a very different idiom for this.

Although, that other languages might have a different idiom is not in and of itself a reason to not use the idiom. Mapping idioms is part of the work of a translator, after all – if someone translates "He's hungry" into French as "Il est affamé", then they're not doing their job properly. Still, in this context (documenting a tool) I think it does make some measure of sense to attempt to avoid idioms. This is a place where it would be great for it to be clear even for a non-native speaker whose language doesn't have a full translation in Wesnoth, so they're forced to read the English instead of the non-existent translation.

Pentarctagon:
This feels like it's not a useful thing to debate (if not outright bikeshedding) - rather than speculate here among ourselves, someone who this impacts should just be asked about it. @kabachuha is the Russian translator, hermestrismi (forums) is the Arabic translator, and there's been a very recent update to the Japanese translation so I assume there's someone you could ask there as well (though not sure if it's still RatArmy as listed on https://wiki.wesnoth.org/WesnothTranslations).

Pentarctagon:
To be clear, I don't think this would've been a silly or pointless change. Just that if we're going to make a change for others, then we should confirm with those other people that the change is correct.

PR #7871, opened 2023-08-25
Also search the path for executable

Pentarctagon:


src/filesystem.cpp
```diff
@@ -1002,18 +1004,34 @@ std::string get_exe_dir()
...
-		return get_cwd();
+	// check the PATH for wesnoth's location
+	// versionless
+	std::string exe = filesystem::base_name(filesystem::get_program_invocation("wesnoth"));
+	bfs::path search = bp::search_path(exe).string();
+	if(!search.string().empty()) {
+		return search.parent_path().string();
+	}
+
+	// with version
```
soliton-, line 1025: I'd say with version should be tried first then.

src/filesystem.cpp
```diff
@@ -1002,18 +1004,34 @@ std::string get_exe_dir()
...
+			return exe.parent_path().string();
 		}
+	}
 
-		return exe.parent_path().string();
-	} else {
-		return get_cwd();
+	// check the PATH for wesnoth's location
+	// versionless
+	std::string exe = filesystem::base_name(filesystem::get_program_invocation("wesnoth"));
```
soliton-, line 1019: It's never a path now, right? So no point in calling base_name.

src/filesystem.cpp
```diff
@@ -1698,7 +1716,7 @@ std::string get_program_invocation(const std::string& program_name)
 #endif
 	);
 
-	return (bfs::path(game_config::wesnoth_program_dir) / real_program_name).string();
```
soliton-, line 1701: `game_config::wesnoth_program_dir` can be completely removed now. It's not used anywhere else.

src/filesystem.cpp

src/filesystem.cpp

src/filesystem.cpp
Pentarctagon: Updated.

soliton- (approved):

PR #7859, opened 2023-08-17
Add Wyrm and Wyrmlet (amphiptere) to Core (from SoF)

Jonathan-Kelly:
Their statistics are (roughly) aimed at complementing the other small but fast flying creatures in core - bats, falcons, dragonflies. Whereas those other small flyers exclusively use melee attacks, this wyrm line focuses instead on its ranged attack (fire breath). But it is meant to be otherwise comparable to them in general power (and thus less power than the special variants in SoF).

The appearance of SoF's Cave Wyrm and Cave Wymlet are also color graded from green to blue to help distinguish them from these new core variants and to perhaps give them a more subterranean appearance. Other color options include yellow (as with the mythical [amphiptere](https://en.wikipedia.org/wiki/Amphiptere)) or a chalky cave albinism look. Also this experimentally gives the Cave Wyrm and Cave Wyrmlet frost breath to further distinguish them from the other variants (but this is just an idea).

PR #7827, opened 2023-08-02
Dragons! (Postponed 1.19)

Hejnewar:
As we all know dragons are weak. 

These changes are here to help with that a bit. 

Fire Dragon - hp changed form 101 to 500, cold resistance changed from -50% to 10%, 14-4 fire breath changed to 14-6, bite changed from 21-2 to 21-4, tail changed from 24-1 to 24-3, xp changed from 250 to 500, cost changed form 100 to 1297g, leadership removed. (Dont approach without at least 7 ancient liches. It bites.)
Skeletal dragon - hp changed from 86 to 171, jaw changed from 10-4 to 17-4, claws changed from 25-2 to 24-3, cost changed from 100 to 297. 

This is a separate issue from monsters balance pass because Draconic weakness is a long standing issue that sparks discussion (im knida counting on that) (also I didnt finish monsters and felt like this PR).

Issue #7146, opened 2022-11-18
Equivalent of [allow_undo] for events added through 1.17's new events API


stevecotton:
### Describe the desired feature

Non-WML events added through the new events API always disable undo with no way to opt out. The long-term fix is to add a way to do that.

Root cause of #6898, which had a sort-term fix in #7136 of undeprecating the old `on_event` API.

stevecotton:
PR #7566 has added this under wesnoth.experimental, so the 1.17 milestone has been reached.

The issue can stay open for moving it to non-experimental in 1.19.

Issue #2559, opened 2018-02-26
Allow marking units as spoilery


CelticMinstrel:
The units database on the Wesnoth site is a nice place to see a list of all units that exist in a given campaign or era. It also offers ways to hide specific units from the list, using `hide_help` or `do_not_list`. As I understand it, this is generally used for units that are only used for internal mechanical purposes, such as the fog clearer.

However, recently it has started to be used to hide spoiler-inducing units from the units database. In my opinion, this starts to defeat the point of the units database, which is to be, essentially, a database of all units. At the same time, I can approve of the goal of wanting to avoid giving away spoilers.

I think a better solution would be to offer _another_ key in `[unit_type]` which is used only by the units database parser. This would mark the unit as a "spoiler" unit, which should by default be hidden in the units database display. The units database website could then add a button or other mechanism to allow you to show spoiler units. I think this would be a good compromise.

In summary, my proposal is:
* Add a key to `[unit_type]` which marks the unit as a spoiler unit
* Add a button or other UI element on the website to show these spoiler units, which are hidden by default.

ProditorMagnus:
This is already possible, #ifdef \_\_WMLUNITS__

ProditorMagnus:
Some units already have different values for database only https://github.com/wesnoth/wesnoth/blob/9bfd432abef371c6c20d526fe05755fe310ba50d/data/core/units/monsters/Fire_Ant.cfg#L13

CelticMinstrel:
The goal is to avoid spoilers while still allowing the unit to be viewed on units.wesnoth.org, but that approach prevents it from ever showing up.

cooljeanius:
Yeah, I'd look to existing implementations of spoiler tags on various websites/apps (e.g. Wesnoth forums, Discord, Mastodon) for inspiration; usually being "spoilery" means that the content is still there, just obfuscated in some way, so that people have to click through to confirm that they actually want to see it

Slayer95:
I am thinking of "flipping" the spoiler unit cards in u.w.o. so that the back of the card shows a customizable message: <i>Spoiler: This unit will be revealed in Chapter 2 of the campaign.</i>, or <i>Spoiler: Travelers have long talked about horrors on the far north.</i>.

But that sort of thing can be useful for the in-game help too, right? At least for me, it has always been a pain point that advancements get hidden every time Wesnoth is updated (or who knows when that happens).

Pentarctagon:
[...]

PR #7874, opened 2023-08-26
Add "aliens" boss to core from UtBS

Jonathan-Kelly:
@Hejnewar and myself are working on putting together some core boss opponents for use in a new multiplayer scenario type (hunt) as well as traditional survival scenarios.

This adds a foxfire mushroom and slime mold inspired 'spawner' boss. It consists of an immobile fruiting body unit which produces mobile "spore" units. These spores can level into barely mobile "stalk" units which nourish the fruiting body to increase the rate at which it produces new spores. Whereas this sporulation rate _decreases_ with the number of living spores, acting like a kind of upkeep (and with stalks acting like villages). A stalk can eventually level into another fruiting body which acts as another spawn point for new spores and also multiplies the spore production rate.

Fungoid units generally have cold attacks with the 'drain' special, making them easier to take on for the undead faction, to compensate for how disadvantaged undead are against the new #7827 fire dragon boss and its #7859 wyrm minions.

Fungoid mobility is very poor, but this is partly counterbalanced by their stun special and a new 'miasma' ability which weakly but broadly attacks all units within a small radius (the damage and radius is determined by the level of the fungoid unit). Also all damage inflicted by spores and stalks heals the immobile fruiting body by the same amount, making it difficult to kill until enough of its minions are defeated.


data/core/units/monsters/Fungoid_Body.cfg
```diff
@@ -0,0 +1,275 @@
...
+    movement_type=smallfoot
+    movement=0
+    experience=150
+    {AMLA_DEFAULT}
+    level=3
+    alignment=chaotic
+    advances_to=null
+    cost=403
+    usage=mixed fighter
+    description= _ "This thing is impossible to describe, no one has seen anything like it before."
```
CelticMinstrel, line 34: This is like the worst possible description. I'd prefer a blank description to be filled in later over this monstrosity.

data/core/units/monsters/Fungoid_Body.cfg
```diff
@@ -0,0 +1,275 @@
...
+        name=attacker hits
+        first_time_only=no
+        [filter]
+            type=Fungoid Spore,Fungoid Stalk
+        [/filter]
+        [heal_unit]
+            [filter]
+                type=Fungoid Body
+            [/filter]
+            amount=$damage_inflicted
```
CelticMinstrel, line 96: Does this have to be a direct transfer? Perhaps you could add an inefficiency to the process, maybe something like `$($damage_inflicted * 0.8)`. Maybe the stalks could even heal the main body more than the spores do.

data/core/units/monsters/Fungoid_Body.cfg
```diff
@@ -0,0 +1,275 @@
...
+            animate=yes
+        [/heal_unit]
+    [/event]
+
+    # Every turn there is a chance a fruiting body will produce a new spore unit.
```
CelticMinstrel, line 115: I think there should be a `[dummy]` ability documenting this fact.

data/core/units/monsters/Fungoid_Spore.cfg
```diff
@@ -0,0 +1,126 @@
...
+        {ABILITY_REGENERATES}
+        [dummy]
+            id=miasma
+            name= _ "miasma"
+            description=_ "This unit passively harms those not of its side at a distance and rate equal to its level."
+        [/dummy]
+    [/abilities]
+    # Miasma ability.
+    [event]
+        id=Fungoid Miasma
```
CelticMinstrel, line 38: This appears to be duplicated in the three units. Couldn't it be made into a macro?

data/core/units/monsters/Fungoid_Body.cfg
```diff
@@ -0,0 +1,275 @@
...
+    description= _ "This thing is impossible to describe, no one has seen anything like it before."
+    {DEFENSE_ANIM_FILTERED ("units/monsters/core-compact.png") ("units/monsters/core-compact.png") ("wail.wav") (direction=n,ne,se,s)}
+    {DEFENSE_ANIM_FILTERED ("units/monsters/core-compact.png~FL(horiz)") ("units/monsters/core-compact.png~FL(horiz)") ("wail.wav") (direction=nw,sw)}
+
+    [abilities]
+        {ABILITY_REGENERATES}
+        [dummy]
+            id=miasma
+            name= _ "miasma"
+            description=_ "This unit passively harms those not of its side at a distance and rate equal to its level."
```
CelticMinstrel, line 43: I think this should probably have a special note as well.

I also think there should be no need to use an event for that. We can already easily make a "harms" ability… as far as I can tell the only issue is that it can't specify a damage type.

data/core/units/monsters/Fungoid_Body.cfg
```diff
@@ -0,0 +1,275 @@
...
+        type=cold
+        range=ranged
+        damage=10
+        number=3
+        [specials]
+            {WEAPON_SPECIAL_STUN}
+        [/specials]
+    [/attack]
+    [attack_anim]
+        # The following attack filter deliberately matches both ranged and melee.
```
CelticMinstrel, line 192: Using the same animation for two different attacks seems bad… as does having two identical attacks that differ only in range.

CelticMinstrel:
This broadly seems okay. Ultimately I think the abilities should be rewritten in Lua, but I wouldn't consider that a blocker to merging this – especially since #7146 isn't done yet.

doofus-01:
I'm not questioning the notion of a fungus monster boss for an MP scenario, but does it have to be the UtBS alien?

Hejnewar:
Funny thing, I wanted them cuz they look very much like something you would find in mad scientist laboratory. I didnt even think about fungus. 

> I'm not questioning the notion of a fungus monster boss for an MP scenario, but does it have to be the UtBS alien?

Nope, it can be whatever but art is hard and this exists. While on topic Ant Queen would be nice addition to ants.

doofus-01:
> Nope, it can be whatever but art is hard and this exists. While on topic Ant Queen would be nice addition to ants.

Fair enough, but recycling of assets (or this general drive to pull everything into core) cheapens the campaigns.  If there is no reason to use the UtBS alien, let us come up with something new.  A fungus monster doesn't require anatomy lessons, should be possible to come up with something entertaining.  Let me know what specific assets are needed.

Hejnewar:
After thinking about it for some time, if the limit is 1 unit, then the best choice is not parasite or predator but symbiont. That will work well with already existing units of woses, it can be made into centerpiece, it can be rather disposable for the woses, there is a lot of options with that and thtas exactly what I want to see, possibilities to use one unit in different scenarios.

CelticMinstrel:
I feel like I'm missing some context here, because what you just said makes no sense whatsoever… I especially have no idea how wose fit into this at all.

Hejnewar:
You know, trees and shrooms have pretty good relationship, they very much help eachother grow and protect from parasites and other dangers. Thats especially un places that are originaly not great for them to live in. If trees can manifest as sentient I see no reason shrooms cant do something similar and start walking.

CelticMinstrel:
So you're implying this fungal creature could have a symbiotic relationship with wose?

Hejnewar:
Well why not. Imo it can work pretty well in game.

CelticMinstrel:
Maybe the miasma ability is filtered so that units of the wose race are not affected? Or the miasma might even benefit wose instead of harming them?

data/core/units/monsters/Fungoid_Body.cfg
Jonathan-Kelly: This is taken directly from UtBS, which generally has the worst possible writing quality. My plan was and is to rewrite the description(s) once the unit(s) design has solidified.

data/core/units/monsters/Fungoid_Body.cfg
Jonathan-Kelly: Yeah, it could very well end up being half efficiency, like with the `drains` ability.

data/core/units/monsters/Fungoid_Body.cfg
Jonathan-Kelly: Yeah that is a good point. If this mechanic remains directly part of the `unit_type` then that will should be the way to handle it. In this present form it might need a very long description though, to cover all the mechanics involved.

data/core/units/monsters/Fungoid_Spore.cfg
Jonathan-Kelly: It probably could be a macro but is too immature to become a _core_ macro. This ability could end up being dropped or modified and renamed. For example, right now it is conceptualized as a near neutrally buoyant noxious gas, so it floats everywhere and ignores terrain defenses. But it could end up considering terrain defense and then also need a new name and description.

data/core/units/monsters/Fungoid_Body.cfg
Jonathan-Kelly: A special note makes sense, agreed.

The cold damage type is preferred. Though maybe at this low of a damage level precision issues make damage type modifiers either ineffectual or a bit too strong. I will have to test.

Hejnewar:
I wouldnt call that miasma under any circumestances but even if spores are really dense it wouldnt have much of an effect on something that doesnt breathe like humans, it would be pretty much as harmful as dust is to plants. 

If we talk about mycorrhiza it is not as instant but still can exist pretty much anywahere and we are talking magic here, I can imagine many ways in which this can work. Additionally given that we have only one wose for all trees is isnt much of a streach that shroom unit A is only worlking with tree unit A and doesnt give any bonuses to tree unit B, because that also happens a lot irl certain species prefer other certain species and shroom type is dependent on forest type. (It is more crucial for shrooms to have trees than for trees to have shrooms but they still benefit from having them a lot.)

data/core/units/monsters/Fungoid_Body.cfg
CelticMinstrel: That note on harms and damage types was sort of wishful thinking for a feature we don't have, more than a request for a change.

data/core/units/monsters/Fungoid_Body.cfg
CelticMinstrel: It's fine for the melee attack to not lunge, but I really don't think it should be the _same_ animation as the ranged attack. Some example possibilities:

* Ranged is a missile (thorn maybe?), melee is a tentacle attack
* Ranged is a really long tentacle, melee is a shorter tentacle
* Ranged is a tentacle, melee just sort of headbutts without moving at the base

data/core/units/monsters/Fungoid_Spore.cfg
CelticMinstrel: I think reimplementing it in Lua would be preferable to a macro, but a macro is preferable to repeating the exact same code in 3 places.

data/core/images/units/monsters/core-compact.png
doofus-01, line 1: As discussed in the general comments, I'm against moving these UtBS aliens into core, especially for something unrelated.  I think the conclusion was that a fungus wose monster can be used instead?

data/core/units/monsters/Fungoid_Body.cfg
```diff
@@ -0,0 +1,275 @@
...
+    # Miasma ability.
+    [event]
+        id=Fungoid Miasma
+        name=turn end
+        first_time_only=no
+        [store_unit]
+            [filter]
+                type="Fungoid Spore","Fungoid Stalk","Fungoid Body"
+            [/filter]
+            variable=fungoids
```
doofus-01, line 55: If this miasma ability becomes mainline, filtering on ability instead of type would make it simpler to use more generally for UMC.

For this ability, it would be nice (necessary?) to have a visual indicator in the standing animation, some sort of halo cloud or something.  

If the ability is just an intended mechanic for the MP scenario you and Hejnewar are working on, then the filtering and standing animation probably don't need to be changed.

doofus-01 (changes requested):
My requests:
1.  Confirm whether recycling the UtBS alien is still in the plans, or what the new plan is
2.  If the miasma ability is a core ability, a visual indication (standing animation) should be addressed.

Thanks.

data/core/images/units/monsters/core-compact.png
Jonathan-Kelly: > I think the conclusion was that a fungus wose monster can be used instead?

I mean honestly... I wouldn't mind having both.  : \

The forest/wose symbiotic, [fairy fire](https://en.wikipedia.org/wiki/Foxfire) looking, enchanted mushroom creature would be great for a forest biome boss fight (either as a minion for Ancient Wose bosses or as a boss itself). 

While creepy super-deep-dwelling subterranean creatures from UtBS would make an excellent cave biome boss fight.

> I'm against moving these UtBS aliens into core, especially for something unrelated.

I have been thinking about this quite a lot, both before launching this PR and since you raised this point. I agree that the sprites should not be reused for something unrelated to what we see in UtBS. But my question is this - what actually are the mysterious creatures in UtBS and how are they connected to the game's story world at all?

We may want to consider that there is nothing player-facing in UtBS which indicates they are supposed to be a space alien (besides maybe the term "energy ray" on one of the attacks). There is no spacecraft or other obvious technology visible above or inside the lair. No dialogue in UtBS or other campaigns hints that aliens from outer space exist or are going to show up in the end.

Instead, the story sets up the big twist that what had appeared to be divine guidance from the Quenoth's goddess, is actually a deception by a demonic imposter. That this demon may _also be a space alien_ is badly supported in-game and seems to distract from the central theme of the story.

That is why, to me at least, a big improvement would be to:
* Move these to core with some game mechanics improvements. (This PR)
* Write mysterious yet more detailed and more thematic flavor text for them. (This PR)
* Have them appear as a creepy deep cave boss in the upcoming MP scenario(s). (In Progress)
* Have them appear in at least one other mainline campaign with more clues to their nature and origin.

As for what their origin should be, I propose this - they are creatures from the deepest underground realm of the world which were either accidentally unearthed by the dwarves or conversely were only held back from reaching the surface by the dwarves (until the catastrophic events of  _The Fall)_ threw everything into chaos).

An _underworld_ origin better ties into their "demonic" role in UtBS and its biggest plot twist. And it gives our wider world building something strange, creepy and even a bit wondrous to find inhabiting the darkest reaches of the subterranean world.

Jonathan-Kelly:
> Fair enough, but recycling of assets (or this general drive to pull everything into core) cheapens the campaigns.

For my part, the only things I want to pull into core from mainline campaigns are the non-anthropomorphic creatures and then have their occurrence be mostly biome or region specific. The only thing left to request moving to core (after this) is perhaps the "Monster Crab" (also from UtBS) since its animation looks like it might be high quality (but I need to check).

data/core/units/monsters/Fungoid_Body.cfg
Jonathan-Kelly: > If this miasma ability becomes mainline, filtering on ability instead of type would make it simpler to use more generally for UMC.

Bah, I should have already thought of that, sorry.

> For this ability, it would be nice (necessary?) to have a visual indicator in the standing animation, some sort of halo cloud or something.

Yeah, that plus having the damage applied to all the victims at once would really speed up the presentation for the player while keeping it clear where the damage is coming from.

data/core/images/units/monsters/core-compact.png
doofus-01: I don't really disagree with any of that, but there is this whole IftU/AtS/Genesis lore that should probably be consulted, even if it isn't mainline.  I'm not qualified to say if there is a conflict here, so I'll ask @nemaara to review.

nemaara: I don't think we should have alien units in core either. Fungoids are fine but I'm very iffy on the cores. At the very least we would need to be extreeeemely careful about the descriptions. If you really want to write their descriptions, that's fine, but like doofus said I will ask that they are compatible with IftU/AtS given how major those UMC are in the wesnoth community. Not asking you to use their lore, but to maintain comptaibility with them.

nemaara (changes requested):
See comment under the core thing.

data/core/units/monsters/Fungoid_Body.cfg
Jonathan-Kelly: That makes sense.

data/core/units/monsters/Fungoid_Spore.cfg
Jonathan-Kelly: I am game to try rewriting this in Lua if that is your recommendation.

data/core/units/monsters/Fungoid_Body.cfg
Jonathan-Kelly: In terms of visual design, the headbutt could potentially work for the Stalk unit but maybe not the Body unit, while a tentacle attack could work for both (and it might also have the `slows` weapon special, which would be interestingly unique on a melee attack).

Have you seen the existing approach in game, though? Like while playing the final scenario of UtBS? It may just be me, but I find it suits these creatures, that they solemnly stand and cast this one attack from any distance. It kind of gives them an air of otherworldly power and indifference (which persists even when they are attacked up close).

data/core/images/units/monsters/core-compact.png
Jonathan-Kelly: Alright, that should be no problem. The style of mainline unit descriptions seems to be from the limited (and sometimes subjective) perspective of wesnothian mage(s) writing semi-scholarly articles from their academy on Alduin (or such). And these very deep dwelling creatures might be rarely encountered by even the dwarves and trolls, so knowledge of them would be scarce and second hand (from the perspective of the in-world descriptions author).

cooljeanius: I agree with @nemaara; this all feels reaaaalllly spoilery. At the very least we should get #2559 fixed before even thinking about merging this.

cooljeanius:
> It sounds like the consensus generally was that this shouldn't be done?

Well, at least, not until #2559 is resolved.

Jonathan-Kelly:
I think the consensus was that it needs a few key changes, all of which are not too difficult.

But since 1.18 has already gone by there is no longer much need to keep this open as a pull request at this time. So closing for now.

PR #7879, opened 2023-08-28
1.17 - Ant Queen

doofus-01:
This is a first draft to get the assets on the table; the details, fluff, and abilities should probably be changed.


CelticMinstrel:
I propose a leadership ability that affects all friendly ants within a 2 or 3 tile radius.

Or maybe a healing ability that affects only friendly ants.

Dalas121:
The Eastern Invasion PR has its own version of an Ant Queen; they should probably be unified at some point.  No attack animation.
queen-ant
queen-ant-defend

doofus-01:
Leadership seems reasonable.  I'll look into the EI-PR next week.  Thanks.

Hejnewar:
I see this unit in two ways, first one is very standard stay on the keep and recruit leader, it doesnt have to be any special. (Good for general use.)

Second one is similar in function but works completely differently, it either just spawn eggs every turn or spawns them conditionally (on ant from the same side kill for example), this one is way more lore related and could be more interesting gameplay-wise depending on exact way of implementation. (More interesting.) 

Also I will respond to the other PR later today.

Jonathan-Kelly:
Can this also be used as a queen for your new fire ant line, perhaps using the color shift `~CS` function? Or would it need to be bigger?

doofus-01:
>econd one is similar in function but works completely differently, it either just spawn eggs every turn or spawns them conditionally (on ant from the same side kill for example), this one is way more lore related and could be more interesting gameplay-wise depending on exact way of implementation. (More interesting.) 

For now I just made it have leadership.  An egg plague weapon special could be interesting yet still simple and intuitive enough.

> The Eastern Invasion PR has its own version of an Ant Queen; they should probably be unified at some point. It doesn't have an attack animation, just the base image and a defense frame.

Looking at the EIR unit, I'm not sure they should be unified, mostly because I don't know the context in the campaign and what changing it would do.  Could it just be its own thing?  Simple name & ID change, slight description string change -> no conflict.

> Can this also be used as a queen for your new fire ant line, perhaps using the color shift ~CS function? Or would it need to be bigger?

I would say they should be different/bigger, but I guess it will come down to where we are at last call for 1.18.

CelticMinstrel:
> Could it just be its own thing?

I think it makes sense to use the same sprites, if nothing else.

doofus-01:
If it uses the same sprite, what's the point?  Go with what makes sense for the EIR scenarios, @Dalas121, and if you need to retain the unit as you had it, keep the sprite, just change the name/ID.

In other news, I've updated the queen to have egg-plague special now.

data/core/units/monsters/Ant_Egg.cfg
```diff
@@ -0,0 +1,52 @@
...
+    [standing_anim]
+        start_time=-50
+        [frame]
+            image="misc/blank-hex.png:50"
+        [/frame]
+        [egg_frame]
+            image="units/monsters/ant/ant_egg.png:50"
+            auto_vflip=no
+            auto_hflip=no
+        [/egg_frame]
```
CelticMinstrel, line 19: Does adding `auto_flip=no` to the main frame not work?
```suggestion
        [frame]
            image="units/monsters/ant/ant_egg.png:50"
            auto_vflip=no
            auto_hflip=no
        [/frame]
```

CelticMinstrel:
So I guess there's reason not to in this case, but… with the egg being an output of plague, you _could_ make all the variations for it, like the zombie has.

Hejnewar:
> So I guess there's reason not to in this case, but… with the egg being an output of plague, you could make all the variations for it, like the zombie has.

No. Never. Plague is cool and all, but the slowdown in art in other areas I believe is quiet severe. One is enough. 

>  An egg plague weapon special could be interesting yet still simple and intuitive enough.

I dont really see it as ant-like, but I have the art so I can make variations of this unit with different abilities so good enough. 

> The Egg

It has more hp then the unit it advances into and it seems like the only way to level it up is by enemy hitting it right now. Maybe give it simple ability granting 1 xp every turn? (Generic name would be nice but I cant think of any.)

Dalas121:
In EI, eggs hatch when they start their turn adjacent to an enemy.  The player can avoid eggs, burst them down in 1 turn, or let them hatch and kill the spawns for more xp.

data/core/units/monsters/Ant_Egg.cfg
doofus-01: It works for most of the animation, but there is some sort of gap, whether in attack vs defense animation timing or something else, that ends up with default behaviour blinking in and out, so the egg does some quick flips back & forth.  `[egg_frame]` was the simplest way out.

doofus-01:
> So I guess there's reason not to in this case, but… with the egg being an output of plague, you could make all the variations for it, like the zombie has.

If it's just Giant Ant vs Fire Ant variants, that's no problem and is in recent commits.  If you mean custom sprites for each corpse type, then I agree with Hejnewar, it's too much, even if it could all be frankensprited.  At least until we can get the open-source AI to do it for us.

> It has more hp then the unit it advances into and it seems like the only way to level it up is by enemy hitting it right now. Maybe give it simple ability granting 1 xp every turn? (Generic name would be nice but I cant think of any.)

Yeah, it's still in flux.  Getting the mechanics ironed out and adding a zombie ant (since there are so many ants now) are what will make this PR complete.

> In EI, eggs hatch when they start their turn adjacent to an enemy. The player can avoid eggs, burst them down in 1 turn, or let them hatch and kill the spawns for more xp.

Currently, these hatch when they die or advance, but allowing the player to squash them makes sense.  Maybe hatching when starting turn next to enemy, or if an adjacent egg dies, or when XP-advancing is the way to go here?

CelticMinstrel:
> No. Never. [...] One is enough.

I don't think it's reasonable to say "never". It may not be appropriate here, specifically.

> Plague is cool and all, but the slowdown in art in other areas I believe is quiet severe.

I'll note that variations don't have to have custom art. They could just tweak the stats a little – for example an egg made from merfolk might have improved water defence.

> I dont really see it as ant-like, but I have the art so I can make variations of this unit with different abilities so good enough.

True, it's more wasp-like than ant-like… definitely a cool idea in any case.

> If it's just Giant Ant vs Fire Ant variants

I'm not talking about variants for different ant types. In fact we probably shouldn't do that, as it could in theory conflict with the plague mechanics. If we want a fire ant version of the egg I think it should be a separate unit, though it could inherit from the giant ant egg.

What I'm talking about is the same as what the zombie has. The mechanics of the plague special uses the `undead_variant` key from the victim's unit to determine which variant of the plague unit to spawn. This mechanic will be functioning in theory for the ant egg regardless of whether or not we leverage it (I think it uses the default variation if the plague unit doesn't have the required variation).

> If you mean custom sprites for each corpse type, then I agree with Hejnewar, it's too much, even if it could all be frankensprited. At least until we can get the open-source AI to do it for us.

I don't mean custom sprites, specifically. The variant will use the base unit's animations if you don't override them, I'm pretty sure. I mean adding the same set of variations that zombies have, possibly with some tweaked stats. If you wanted to change the sprite for some of them, fine, but that's not a required aspect of the suggestion.

doofus-01:
> I'm not talking about variants for different ant types. In fact we probably shouldn't do that, as it could in theory conflict with the plague mechanics. If we want a fire ant version of the egg I think it should be a separate unit, though it could inherit from the giant ant egg.

Ant types is the only variant that makes sense to me, it seems to work OK.

>What I'm talking about is the same as what the zombie has. The mechanics of the plague special uses the undead_variant key from the victim's unit to determine which variant of the plague unit to spawn. This mechanic will be functioning in theory for the ant egg regardless of whether or not we leverage it (I think it uses the default variation if the plague unit doesn't have the required variation).

I don't think I'm understanding what you are suggesting.  There should be an ant type for each unit/race?

> I don't mean custom sprites, specifically. The variant will use the base unit's animations if you don't override them, I'm pretty sure. I mean adding the same set of variations that zombies have, possibly with some tweaked stats. If you wanted to change the sprite for some of them, fine, but that's not a required aspect of the suggestion.

Recycling sprites, so that the same images represent different interactions, is a hard "no" from me, especially if variants can be present on the same map.

------------------------------------------
In any case, I consider this PR done, just need to make sure it's OK to merge.  Ants could probably use more work, such as a hatching animation and maybe making them their own race, but that's not for this PR.

Hejnewar:
IMO, to be good this PR needs better description of how the egg actually works because there is no word anywhere how it works that player can see in game unless I missed something.

Hejnewar:
Honestly I dont see any reason to do this, it adds unnecessary complexity and work, sometimes probably player confusion as there isnt any visual indicator and it isnt adding to realism. At best its like small quirky thing that exists but no one is paying any attention to. At least zombies can be effectively used in campaigns for immersion.

CelticMinstrel:
Mostly I just want to make sure that if it's not done, it's not done on purpose, and not because the possibility was simply overlooked.

doofus-01:
> IMO, to be good this PR needs better description of how the egg actually works because there is no word anywhere how it works that player can see in game unless I missed something.

Uh, yeah...  Now it has something at least.

data/core/units/monsters/Ant_Egg.cfg
```diff
@@ -38,7 +43,9 @@
     advances_to=Giant Ant
     cost=4
     usage=fighter
-    description= _ "Giant Ants are common in almost any environment, and the eggs from which they emerge can often be found in caves or any sheltered environment.  Most eggs come from one of the colony queens, but the workers transport them to any suitable location."
+    description= _ "Giant Ants are common in almost any environment, and the eggs from which they emerge can often be found in caves or any sheltered environment.  Most eggs come from one of the colony queens, but the workers transport them to any suitable location.
+
+" + {ANT_EGG_TEXT}
```
CelticMinstrel, line 48: I have a feeling that maybe the paragraph break should be separated out as a non-translatable string? Something like this:

```
...to any suitable location" + "

" + {ANT_EGG_TEXT}
```

doofus-01: Why is that?

CelticMinstrel: I'm not quite sure where I got this from, but I think there was a recommendation not to put whitespace at the beginning or end of translatable strings. I might be imagining it though.

doofus-01: Let's see if anyone complains.

CelticMinstrel: Or if the paragraphs get stuck together into a single paragraph in some languages.

data/core/units/monsters/Ant_Egg.cfg
CelticMinstrel: Hmm… feels like there must be something missing here… if we can figure out what it is, I'd definitely prefer not to use an `[egg_frame]`… and there's always the chance it's an actual bug…

doofus-01:
I don't think the CI/windows failure is related to this, I'll plan to merge soon.

PR #6677, opened 2022-05-03
UtBS S8 fixed wallstuck units on and moved PASSABLE_HEX macro to core

Descacharrado:
Half of #6439 Moves the macro first used in #6357 and uses it to fix the possibility of wall-stuck units in UtBS S8.

Test photos:
test_utbs_s8
test_trow_s19

PR #7883, opened 2023-08-29
Refact the PASSABLE_HEX implementation

Descacharrado:
In #6677 I moved the PASSABLE_HEX macro into core so I could use it to fix wallstuck spawned units. As commented on discord (@soliton-), it makes much more sense to move passable=yes to the unit generating macros so the default behaviour is more aligned to what one would expect (units appearing in places where they can walk). One can still opt-out of this by overwriting the default:

`{NAMED_UNIT 3 (Skeleton) 15 40 () ( _ "Restless Dead") (upkeep,animate,passable=free,yes,no)}` (Tested)

With this in mind, if we approve this PR maybe we should remove the PASSABLE_HEX macro from core, I dont know if we would need to deprecate it first.

Marking as draft for discussion on the matter.


soliton-:
PASSABLE_HEX is new in 1.17 so I don't see a need for deprecation.

CelticMinstrel:
Hmm. Related to this, I think we should add a new `fallback=yes|no` (pending finalized name) option to the `[unit]` tag, and set it to yes in the unit macro as well. What it would do is fall back to impassable placement if a passable hex cannot be found.

Relevant code:

https://github.com/wesnoth/wesnoth/blob/9d3b8241fa3887c0f03ba30278fa9ef28bf31de3/src/actions/unit_creator.cpp#L114-L165

soliton-:
@Descacharrado I think it’d be good to also remove the macro definition and mark this as ready. Then we can wait a bit longer and push if there are no more comments. Would be nice to implement what celmin suggests as well but even without I’m fairly confident this does more good than bad.

soliton- (approved):

soliton-:
@Descacharrado can you please sync your branch so the checks run again with current master?

soliton-:
Ok, looks like we need to update some tests that depend on the passable=no default.

data/core/macros/unit-utils.cfg
```diff
@@ -27,6 +27,7 @@
         type={TYPE}
         x={X}
         y={Y}
+        passable=yes
```
soliton-, line 30:
```suggestion
#ifdef TEST
        overwrite=yes # unit tests like to overwrite
#else   
        passable=yes
#endif  
```

data/core/macros/unit-utils.cfg
```diff
@@ -27,6 +27,11 @@
         type={TYPE}
         x={X}
         y={Y}
+#ifdef TEST
+        overwrite=yes # unit tests like to overwrite
+#else   
+        passable=yes
+#endif
```
soliton-, line 34:
```suggestion
#ifdef TEST
        overwrite=yes # unit tests like to overwrite
#else
        passable=yes
#endif
```

data/core/macros/unit-utils.cfg
```diff
@@ -27,6 +27,11 @@
         type={TYPE}
         x={X}
         y={Y}
+#ifdef TEST
+        overwrite=yes # unit tests like to overwrite
+#else
```
soliton-, line 32:
```suggestion
#ifndef TEST
```

stevecotton:
I'm opposed to using `#ifdef TEST` - code in the unit tests should be example code that can be cut&pasted into other places and work the same way. If lots of unit tests need to change because of a change to a macro, then the Git history ought to say that.

The PR title looked like it was only changing `PASSABLE_HEX`, but it's changing `UNIT` and `NAMED_UNIT`. I think that should be treated as an API change, and rejected from 1.18.

soliton-:
I'm still trying to work out what the issue with the unit tests is.

We can certainly postpone switching to sensible UNIT macros a couple more years.

Pentarctagon:
@soliton- were you still planning on working on this?

soliton-:
Well, kind of moot since I didn't get to it before 1.18.

I certainly would still like to deprecate all the `[+tag]` macros. I think it's a bad habit. We should either use a macro tailored for the use case or use a generic macro that allows adding whatever WML when calling the macro as in the case of the `UNIT` macro for example.

Pentarctagon:
Should this be closed then, or left open?

soliton-:
Closing this in favor of #8762.

Issue #4474, opened 2019-10-16
Refactor handling of factions and leaders


Wedge009:
Exposed by #2954 and #4472. Indeed, the commentary in `mp_staging::update_leader_display()` is telling:
https://github.com/wesnoth/wesnoth/blob/c4c4543fe4ef6a59e49e228201fcbe3058d195ac/src/gui/dialogs/multiplayer/mp_staging.cpp#L442-L444

In the same way, the `"random"` string value was cause for confusion as seen in #4472.

Discussion in #wesnoth-dev included:
> In principle, the way to design this cleanly would be to have three possibilities: none / random / specific, and if it's the latter, have another field that says which specific unit it is.

I would consider 'none' to correspond to the current `"null"`, 'random' of course corresponds to `"random"`, and then 'specific' would be the races/factions pulled in from Wesnoth data (`*.cfg`) files.

gfgtdf:
I don't think having special strings there is a real problem, as long as it's documented and while we could in theory replace "null" with "none" it could be a compatibility break which is not worth it.

Maybe it would be cleaner though if that code was a part of flg_managwe though, like adding a new function `flg_manager::get_leader_name()`

CelticMinstrel:
It's just in the "flg" manager, so why not make the leader and gender there be a variant?

```cpp
struct random_t {} random;

std::variant<std::monostate, random_t, std::string> current_leader;
std::variant<std::monostate, random_t, std::string> current_gender;
```

PR #7887, opened 2023-08-30
Fix #7886: leaders type set to null in campaigns.

### LINKED ISSUE #7886: controller=null sides in campaigns work inconsistently if you add canrecruit=yes unit

- **Version:** 1.16.5, also on 1.16.1
- **Downloaded from:** 1.16.5 Flatpak. 1.16.1 built from source
- **OS:** Ubuntu 20.04

### Description of the bug

Wiki says that "null: the side doesn't get a turn to move and doesn't have a leader generated from the contents of the [side] tag. (It still can get units from [unit] tags in the [side] tag.)" Let's suppose you have the following code

```
	[side]
		side=2
		controller=null
		user_team_name= _ "Good Guys"
		team_name=we
		[unit]
			type=Footpad
			x, y=6, 4
			id=alright
			name = "Non recruiting"
		[/unit]
		[unit]
			type=General
			id=bugged
			x, y=5, 9
			canrecruit=yes
			name = "Recruiting"
		[/unit]
	[/side]
```

You'd expect that side=2 receives two units, a General with canrecruit=yes and an ordinary Footpad. But in fact the outcome depends on whether you proceed to the scenario from the previous scenario of the campaign or you load it from a start-of-scenario save.

If you load it from a start-of-scenario save, it works

If you proceed from the previous scenario, logs say
> error engine/team_construction: game_error: unknown unit type: null

and General doesn't appear on the board.

I made a short demo campaign for this bug 
[Null_Controller_Bug.zip](https://github.com/wesnoth/wesnoth/files/12478624/Null_Controller_Bug.zip)

1. Download and install the demo campaign
2. Start it. Win the first scenario (it's enough to end turn and it will claim victory)
3. Go next scenario. You will see that you have only the leader of side 1 and the Footpad on the board, no General.
4. Now return to the main menu, click Load and choose the start-of-scenario save for the second scenario (for me it's called "controllernullBUG-Big Bug.gz")
5. Notice that now the General appears and speaks his line of text

### Expected behavior

General appears in both cases

### Additional context

One could argue that `controller=null` sides shouldn't have leaders anyway. I still think there are two reasons it's a bug
1. Wiki only says the side doesn't get a leader generated from `[side]` tag and says units with `[unit]` tags can be added, without any explicit restriction on them

And most importantly

2. There is inconsistent behaviour here, when proceeding from the previous scenario and loading from a start-of-scenario save produced different outcomes. 

I think it shouldn't be like that, and although this particular bug itself shoots only in a fairly limited amount of cases, it may highlight some deeper pitfalls with Wesnoth campaigns' code

gfgtdf:


Wedge009:
The comment in the line following this change reminds me of the same issue in #4474.

gfgtdf:
I think that code could also set it to an empty string and it would have the same effect.

CelticMinstrel:
Do campaigns always have "controller lock"?

gfgtdf:
Yes (unless explicitly set to no in the WML)

stevecotton:
The bug was already present in 1.16.1, so not a regression on the 1.18 branch. (I'm just checking which of these fixes need to be in the next stable branch.)

soliton-:
> Yes (unless explicitly set to no in the WML)

So in the cases `controller_lock` is set to no the issue remains?

soliton-:
Based on the documentation for controller=null
> the side doesn't get a turn to move and doesn't have a leader generated from the contents of the [side] tag. (It still can get units from [unit] tags in the [side] tag.) 

It rather sounds like for controller=null [unit] tags in [side] should not be considered the leader. (Or the documentation should be changed.)

gfgtdf:
> So in the cases controller_lock is set to no the issue remains?

Probably yes, but i don't see any anyone would ever do that (in a campaign). and a better fix would be more complicated (and i feel that'd risk introducing new bugs shortly before 1.18)


> It rather sounds like for controller=null [unit] tags in [side] should not be considered the leader. (Or the documentation should be changed.)

I wouldn't read it that way, for me it sounds like just the directly content of [side] is not used to generate a leader. [unit]s can still explicitly have canrecruit=yes in them set.

But i also think that part of the wiki is also no longer correct, i don't think that code makes a difference between controller=null and other sides.

gfgtdf:
I do however thing that the mp connect code should never attempt to modify the contents of a [unit] in [side] (note that this in particular implies the recall list). But the original code was explicitly written with this possibility so i'm hesitant to remove it.

soliton-:
Seeing 14c9ed4c5dc155c0a244278f34eec3730c840a68 I'd say the chance is small that was correct.

gfgtdf:
I'm not sure what you mean here.

soliton-:
That's the commit that introduced setting to "null".

gfgtdf:
so the "that" in your last message refered to the commit ?

gfgtdf:
I just tested that in sp controller=null does not change when leaders are generated, ill remove that section in the wiki i think.

gfgtdf:
But for mp the situation is a bit different since the user can disable sides (for example to play a 2vs2 map as a 1vs2), i think that's where the commit comes from.

gfgtdf:
So the idea of the new code is:

 CNTR_EMPTY && !controller lock:
->This was a side that was disable by the user, remove the leader
 CNTR_EMPTY && controller lock:
-> This was an uncontrollable side specified by the scenario designer, don't change it.

soliton-:
Given the error of "error engine/team_construction: game_error: unknown unit type: null" a fix seems accidental at best.

gfgtdf:
I _think_ you can also replace "null" with just an empty string, should have the same effect.

gfgtdf:
Didn't test that one though

gfgtdf:
https://github.com/wesnoth/wesnoth/blob/13353f821271873c62fff377e213396881efa384/src/game_initialization/connect_engine.cpp#L1060

This is the code that explicitly supports changing carried over leaders.

PR #7902, opened 2023-09-11
1.17 - Ravens

doofus-01:
This is something I'd started earlier in the year.  The window for new things is closing, and while lack of ravens isn't the most pressing issue, it is a project I think I can get completed in time.  Very WIP for now.

EDIT:  Roc isn't a Raven, but it is a bird, and something needs to be done about it before 1.18.


CelticMinstrel:
Ooooh, that's a much better roc than the previous one! 👍

Hejnewar:
3 things: 
1. Raven evolves from dark flying into dark electrict type!
2. The shadow under current roc is bigger than the shadow under PR roc. Just observation.
3. Stop making stealthy balance changes without a word in PR, looking from the past also no mention in changelog, no tag (not that I would see it but still), no anything. This isnt the first time and I really hate that, especially because Im kinda responsible for keeping everything in good state when it comes to units.

doofus-01:
@CelticMinstrel - Thanks!  Existing Roc may have been a placeholder for a project that stalled.  The melee attack is from the Sinbad stories.  Better animations can be added in stable 1.18 (as long as no images are removed).

@Hejnewar - I have read your comment.

CelticMinstrel:
I'm not sure if we have an actual policy on this, but personally I think it's better for all mechanical stuff (like attack, movement costs, defense, etc) to go above all the unit animations. I'm mentioning it because I happened to notice that the Black Herald's attacks and defenses are sandwiched between some of the animations.

Hejnewar:
> I'm not sure if we have an actual policy on this

I agree it would be good to have it somewhat standarised.

Jonathan-Kelly:
The Roc has a dive bomb attack, nice! Looking forward to seeing it in action. Bumping it up from a level 2 is also highly welcome. Given the stories of Rocs carrying away elephants, how about taking it higher (no pun intended) to a level 4... or even a level 5? It could make for a fun new "boss" unit (definitely good to have more of those).

The ravens with their slightly slower movement speed should be a fine alternative to the falcons in scenarios which use fog. Just one small anatomical suggestion coming from hands on experience with corvids - they actually do not have talons, just blunt nails with weak foot grip strength like a song bird, except bigger. Their beak is a good deal more formidable, but in a pecking kind of motion, as their bite force is also on the weak side. So maybe more of a "piercing" damage type fits that.

doofus-01:
> Just one small anatomical suggestion coming from hands on experience with corvids - they actually do not have talons, just blunt nails with weak foot grip strength like a song bird, except bigger.

Yeah, there were some hangovers from the falcons that were a starting point.  The pierce beak-only makes sense.  I'm not totally happy with the current sprites, but I think they will work for now.  The XP and costs are almost random, I'm sure whatever I put there will be wrong.  The lightning attack isn't random though.

>  Given the stories of Rocs carrying away elephants, how about taking it higher (no pun intended) to a level 4... or even a level 5? It could make for a fun new "boss" unit (definitely good to have more of those).

I had similar thoughts.  If the boss scenario required dealing with a Roc roost, several (slightly OP) L3 could make sense.  If it's a mamma Roc and helpers (falcons?) then boosting it further makes sense.

>I 'm not sure if we have an actual policy on this, but personally I think it's better for all mechanical stuff (like attack, movement costs, defense, etc) to go above all the unit animations.

I agree it would be nice if there was a convention about the order of unit config data.  Anyone who feels strongly enough should definitely open a PR.

Hejnewar:
> I agree it would be nice if there was a convention about the order of unit config data. Anyone who feels strongly enough should definitely open a PR.

I might but not anytime soon.

Jonathan-Kelly:
> The lightning attack isn't random though.

Gives it a fittingly folklore flavor, a bit like your dark horse has. Also good to have more wildlife and monsters with ranged attacks.

> I had similar thoughts. If the boss scenario required dealing with a Roc roost, several (slightly OP) L3 could make sense. If it's a mamma Roc and helpers (falcons?) then boosting it further makes sense.

Maybe a bit of each? Like a nesting pair with a bunch of low level helpers. Since Wesnoth's unit leveling system mostly scales linearly, even a level 4 or 5 is fairly easy for players to dispatch with a small force of level 1 recruits.

> I agree it would be nice if there was a convention about the order of unit config data.  Anyone who feels strongly enough should definitely open a PR.

What do you all think of the layout used on the [WoF unit configs](https://github.com/wesnoth/wesnoth/tree/master/data/campaigns/Winds_of_Fate/units)? I separated and labeled each section with a title comment, so it uses a bit more white space but maybe the improved human readability can justify that? I might do something like this for all units in 1.19, if we like the layout.

doofus-01:
> Maybe a bit of each? Like a nesting pair with a bunch of low level helpers. Since Wesnoth's unit leveling system mostly scales linearly, even a level 4 or 5 is fairly easy for players to dispatch with a small force of level 1 recruits.

Since it can fly, it would make sense if it could escape the ring of cannon-fodder technique.  Maybe just skirmisher is enough.  In any case, adjustments can be in a separate PR.

> What do you all think of the layout used on the [WoF unit configs](https://github.com/wesnoth/wesnoth/tree/master/data/campaigns/Winds_of_Fate/units)? I separated and labeled each section with a title comment, so it uses a bit more white space but maybe the improved human readability can justify that? I might do something like this for all units in 1.19, if we like the layout.

I got nothing against it, just not sure revising the layout of existing files is worth the effort.  The unit configs are much simpler than anything else, as each section is pretty self-contained.  But if you're willing to do it, it's your call.

CelticMinstrel:
> Maybe just skirmisher is enough.

Skirmisher isn't _quite_ enough to escape "ring-of-cannon-fodder" since it doesn't let it move through an actual unit. I think that might actually be a cool thing to add to the roc though.

data/core/units/monsters/Roc.cfg
```diff
@@ -6,44 +6,79 @@
...
     experience=100
-    level=2
+    level=3
     alignment=neutral
     advances_to=null
     {AMLA_DEFAULT}
     cost=36
     usage=scout
-    description= _ "Far from human cities in the great desert mountains, the large birds of prey known as Rocs can sometimes be found rapidly scouring the sands on their hunts. Despite their large size, Rocs tend to be reclusive and are only rarely encountered by other sapient races. Their elusive nature has raised them to near mythological status, particularly among the nomads of the deserts.
+    description= _ "In remote desert mountains, far from any city or settlement, enormous raptors are rumoured to nest in the lofty crags.  Travellers and nomads claim to have witnessed livestock and grown men snatched into the sky without warning by these monsters.  While one should not place much faith in the accounts of the desert roamers, they are less fantastical than the drunken lies of sailors who claim to have witnessed full-rigged ships, attempting to navigate the narrow straights between the sheer cliffs where nothing can grow, sunk by giant eagles dropping large rocks.
```
Wedge009, line 20: > narrow straights

I think this should be 'narrow straits'. While strait is, by definition, narrow, I read this as intended to convey a picture of a passage perhaps only a dozen or two metres wide (between the sheer cliffs), so I'm okay with leaving narrow there.

@Pentarctagon Another correction needed post-freeze.

Pentarctagon: Could you open an issue that lists anything you find?

Wedge009: Planning to do so.

PR #7905, opened 2023-09-13
fix #7821: all cases of the calculation should be cached

### LINKED ISSUE #7821: AI recruitment doesn't cache the combat simulations

- **Version:** 1.17.19+dev

### Description of the bug

I haven't profiled this to see if the optimisation is required, however the AI does combat analysis before recruiting, so it makes sense that a cache would be required for good performance. According to the documentation it's cached:

```
/**
 * For Combat Analysis.
 * Calculates how good unit-type a is against unit type b.
 * If the value is bigger then 0, a is better then b.
 * If the value is 2.0 then unit-type a is twice as good as unit-type b.
 * Since this function is called very often it uses a cache.
 */
double recruitment::compare_unit_types(const std::string& a, const std::string& b) {
```

However, looking further down the function, it only caches the result if one of the units involved has no attacks, such as the 1.14 (or older) version of the Elvish Lady. If both units do damage to each other, then we reach one of these statements and skip inserting the result into the cache.

```
		// Normal case
		double value_of_a = damage_to_b / (b_max_hp * a_cost);
		double value_of_b = damage_to_a / (a_max_hp * b_cost);

		if (value_of_a > value_of_b) {
			return value_of_a / value_of_b;
		} else if (value_of_a < value_of_b) {
			return -value_of_b / value_of_a;
		} else {
			return 0.;
		}
```

1. Put a breakpoint or debug statement on the `return *cache_value;` line
2. Make the AI play a match
3. Notice that the breakpoint is never hit

### Expected behavior

_No response_

### Additional context

It looks like the bug has been there since dc72d7b67e60, almost 10 years ago.

GomesGoncalo:
in the previous logic some cases return from the function and others would set a return variable that would be set later.

this commit, changes all the cases into all returns so that other cases added in the future should not avoid caching.

right now there are no ways to avoid the caching which should be the expected behaviour (unless explicitly added the case where cache is unwanted)

### DIFF
--- a/src/ai/default/recruitment.cpp
+++ b/src/ai/default/recruitment.cpp
@@ -933,11 +933,11 @@ double recruitment::compare_unit_types(const std::string& a, const std::string& 
 		double value_of_b = damage_to_a / (a_max_hp * b_cost);
 
 		if (value_of_a > value_of_b) {
-			return value_of_a / value_of_b;
+			retval = value_of_a / value_of_b;
 		} else if (value_of_a < value_of_b) {
-			return -value_of_b / value_of_a;
+			retval = -value_of_b / value_of_a;
 		} else {
-			return 0.;
+			retval = 0.;
 		}
 	}
 


soliton-:
Looks good. Is there a reason why you use a lambda and invoke?

GomesGoncalo:
> Looks good. Is there a reason why you use a lambda and invoke?

Nothing in particular, felt like the lambda was a nice way to force anything in that logic to be cached (so there's no way around it - well maybe throwing an exception)
And I think that lambda's `operator()`  will get inlined so there shouldn't be any overhead.

And the invoke also not a strong reason, I prefer invoke to calling `operator(), I think it makes it a bit clearer. (e.g. being f a lambda you do g(f()) and in some other place just do g(f) to pass the lambda inside, I know the `()` is there but sometimes people might not see it, so just feel invoke makes it clear)

(So, mostly my preference :smile:, it really makes no difference, if you prefer a separate function and just call the function that's totally fine)

soliton-:
> Nothing in particular, felt like the lambda was a nice way to force anything in that logic to be cached (so there's no way around it - well maybe throwing an exception)

That seems futile. The logic just has to be written correctly.

I'd prefer just normal inline logic as was already there. I don't think moving this to a lambda helps readability but if that where to be done then it should be a separate function.

GomesGoncalo:
Alright, on further thoughts I see how reading a return in the middle of the function that doesn't really return from the function could also at first glance not be clear; updated it!

GomesGoncalo:
(updated the commit message)

soliton-:
Thanks, I'll merge once the checks ran through if nobody beats me to it.

Wedge009:
Looks like the usual glib-on-vcpkg issue, so presumably this is okay to merge.

PR #7906, opened 2023-09-15
Send messages to newly joining players

Kingofd:
This is my second shot at "pinning" comments #6974
Now they are being replayed to each player newly joining

It is as of yet without sending the timestamp with the original message, as that would entail much more effort than just adding the base functionality


src/server/wesnothd/game.cpp
```diff
@@ -1425,6 +1428,21 @@ bool game::add_player(player_iterator player, bool observer)
 			user->info().name() + " has the same IP as: " + clones);
 	}
 
+	// If not observer, send the game chat log.
```
soliton-, line 1431: Why not send to observers?

src/server/wesnothd/game.cpp
```diff
@@ -882,7 +884,8 @@ void game::process_message(simple_wml::document& data, player_iterator user)
 
 	const simple_wml::string_span& msg = (*message)["message"];
 	chat_message::truncate_message(msg, *message);
-
+	// Buffer message for sending message history to newly joined players
+	msg_queue_.push_back(std::pair<std::string, std::string>(user->info().name(), msg.to_string()));
```
soliton-, line 888: Why not save the wml? You could just use the existing history data structure and functions.

gfgtdf:
So with this I assume they are shown twice? Once at game start and once again after the original position they are in?

Another thing we should think of is how we are going to handle messages that are sent while the newly joined player is replaying

soliton-:
This about messages before game start.

gfgtdf:
Oh I misread the original issue then.

src/server/wesnothd/game.cpp
Kingofd: Wasn't sure how chat to observers is handled, it now will be sent to all observers also

src/server/wesnothd/game.cpp
Kingofd: At the `simple_wml::document& data` could be some other data that's not meant for others, so I would need to extract the message part in either case.

Later in the resend logic I will need to repackage the data structures to a `simple_wml::document` for the `server.send_to_player()` function anyway, so saving `simple_wml::string_span` doesn't seem worth the overhead.

Kingofd: Or do you mean another existing data structure I have overlooked?

soliton-: You can use `record_data` and `send_history` before start for this similar to how they're used after start for the replay data.

There is no support for merging of unrelated documents. If a custom client does that then neither wesnoth nor wesnothd will understand it.

src/server/wesnothd/game.cpp
```diff
@@ -1748,6 +1755,24 @@ void game::send_history(player_iterator player) const
...
+{
+	if(chat_history_.empty()) {
+		return;
+	}
+	try {
+		for(auto& h : chat_history_) {
+			std::string buf = h->output();
+			auto doc = std::make_unique<simple_wml::document>(buf.c_str(), simple_wml::INIT_STATIC);
+			doc->compress();
+			server.send_to_player(player, *doc);
```
soliton-, line 1768: Should not be necessary to make new documents here.

soliton- (changes requested):

Pentarctagon:
Can this be merged?

Kingofd:
Ye, I incorporated both change requests from soliton-, apparently the "changes requested" flag did not update

src/server/wesnothd/game.cpp
```diff
@@ -1748,6 +1752,21 @@ void game::send_history(player_iterator player) const
...
+	if(chat_history_.empty()) {
+		return;
+	}
+	try {
+		for(auto& h : chat_history_) {
+			server.send_to_player(player, *h);
+		}
+	} catch(const simple_wml::error& e) {
+		WRN_CONFIG << __func__ << ": simple_wml error: " << e.message;
+	}
```
soliton-, line 1766:
```suggestion
	for(auto& h : chat_history_) {
		server.send_to_player(player, *h);
	}
```

soliton- (approved):

soliton-:
Can be merged once checks ran through again.

soliton-:
Thanks for sticking with it and implementing this!

Issue #7082, opened 2022-11-01
Add wmllint to Travis builds


Wedge009:
### Describe the desired feature

`wmllint` errors, genuine or otherwise, have been discussed on and off, most recently in #5799. Commentary in #4381 suggests adding it to the Travis build.

Pentarctagon:
We use Github Actions these days, not Travis :P

Wedge009:
I have no idea what the terminology is, I was just following what was written.

sevu:
Exists for add-ons:
https://github.com/czyzby/wesnoth-wml-linter

It is a bit different though, in that it runs wmllint on mainline (to get the unit ids etc) and the commit/pr in the add-on repo, and filters the output which belongs to mainline.

Pentarctagon:
> I have no idea what the terminology is, I was just following what was written.

For future reference then - travis and github actions are CI service providers. Travis is what we used before switching to github actions.

czyzby:
@Pentarctagon What is the expected behavior? Should the jobs fail if wmllint/wmlindent/wmlscope report any issues? Because right now there are quite a lot warnings if you run those.

Pentarctagon:
Yes, the expectation would be that it should fail. And yeah, it's known there's a bunch of things it warns on. The issue really is that nobody has gone through all those warnings and either fixed the issue or removed the warning for being bad/irrelevant.

czyzby:
What about spellchecking? I imagine it might increase the number of errors significantly, since you have to register all non-dictionary words with `wmllint: general spelling(s)`.

Pentarctagon:
What's the question about spellchecking?

czyzby:
Should it also fail the pipeline or be skipped in the automatic check?

Pentarctagon:
It should fail the pipeline.

czyzby:
OK, last question - could you give me a list of all tools that you want and any command line flags that you'd like to include? I'm assuming we'd like to run `wmllint`, `wmlindent` and `wmlscope` in dry-run mode wherever possible. Any extra flags for `wmlscope`?

Pentarctagon:
wmlindent is already run. I'm not familiar with wmlscope.

cooljeanius:
> wmlindent is already run. I'm not familiar with wmlscope.

I usually end up skipping that one, just because it prints *so much* output that I don't know where to start... might be worth asking @Elvish-Hunter about...

CelticMinstrel:
[...]

PR #7908, opened 2023-09-16
Update WoF

Jonathan-Kelly:
This will probably be its last big update of for this development cycle. The focus was on game mechanics and balance:

* Added an Easy difficulty setting for new players.
* Recall costs now vary per unit (being equal to its full recruit cost). Similar to #7450 for UtBS.
* Removed all remaining turn limits from scenarios that had no story justification for them.
* Wildlife spawners rewritten for easier maintenance and better balance (they now have populations caps).
* Added the excellent new core wildlife units by @doofus-01
* Tested with the latest unit balance improvements by @Hejnewar


data/campaigns/Winds_of_Fate/about.cfg
```diff
@@ -18,7 +18,7 @@
     [/entry]
     [entry]
         name="inferno8"
-        comment="storm daemon unit sprite and attack icons from Era of Magic"
+        comment="storm wisp unit sprite and attack icons from Era of Magic"
```
nemaara, line 21: I kinda forgot to check this when we first merged the campaign, but did you add licensing for these images? I think EoMa uses CC BY-SA but not entirely sure.

data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
```diff
@@ -211,27 +154,29 @@
...
             message= _ "Our trial for aspirancy has come at last.
 The hunt that awaits us below will prove a worthy test."
         [/message]
         [message]
-            speaker=Reshan
+            speaker=Resha
             message= _ "If only we could still count Karron among us...
 
-I fear without his aid this trial will outmatch us.
+I fear without her aid this trial will outmatch us.
```
nemaara, line 173: Are we absolutely sure we want to make the flying fire lizards exhibit human sexual dimorphism? I would have thought having them be hermaphroditic would be more interesting, but I'd like to hear some other peoples' reasoning too.

data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
```diff
@@ -57,106 +57,21 @@
...
-
-    [side]
-        side=7
-        controller=ai
-        hidden=yes
-        team_name=serpents
-        user_team_name= _ "Serpents"
-        no_leader=yes
-    [/side]
+    {SPAWNER_SIDE 2  "Stymphalian"    Mm      {ON_DIFFICULTY4 0.1  0.25  0.5  1}   9 2}
```
nemaara, line 60: How do these spawns vary on difficulty? Is it linear based on the numbers provided? Basically a 10x difference between easy and nightmare is still perhaps too much variance because as I understand it, this is a per turn spawn rate?

data/campaigns/Winds_of_Fate/scenarios/05_Threshold.cfg
```diff
@@ -865,7 +715,7 @@ We would be wise to secure all settlements before then."
 
 I have found anoþer Aquadyne. It is here, deep beneaþ Þe Þree Sisters. It has only been þree years since Haldric destroyed þe elf gosts þat were sapping it and already þere is a mælstrom formed off þe norþ shore. Just like þose mælstroms Seeker Perþ discovered west of Alduin before a Wisp chased away his ship.
```
nemaara, line 716: I somehow missed this the first time, but do we know if these characters can be displayed on all systems? Are there some that they will just be displayed as boxes on?

data/campaigns/Winds_of_Fate/scenarios/05_Threshold.cfg
```diff
@@ -865,7 +715,7 @@ We would be wise to secure all settlements before then."
...
 
-Indeed, þere must be an Aquadyne every twenty leagues around þe entire Great Continent. Everyþing is just as Seeker Oþila had predicted. We should never have left her behind on þe Green Isle. Wiþout her work it may take centuries before we of þe Silver Order know what mankind is dealing wiþ on þis continent. Þe Færie Realm is stronger here þan we ever imagined.
+Indeed, þere must be an Aquadyne every twenty leagues around þe entire Great Continent. Everyþing is just as Seeker Oþila had predicted. We should never have left her behind on þe Green Isle. Wiþout her work it may take centuries before we of þe Silver Order know what mankind is dealing wiþ on þis continent. Þe Færie World is stronger here þan we ever imagined.
```
nemaara, line 718: What's the reasoning for world vs. realm? We don't actually have fleshed out lore for either, but I want to get your thoughts.

data/campaigns/Winds_of_Fate/scenarios/07_Harvest.cfg
```diff
@@ -453,7 +463,7 @@ My diary is very personal, Vonel!"
             # The Lady Dionli referred to is from TRoW.
             message= _ "7th day of Greenleaf
 
-...the Sapphire of Ice has yet improved even my own talent for the art of scrying. I have discarded my Orb of Pondering in favor of it. Soon I shall peer into the far side of our Faerie Realm, which that pretender Lady Dionli claims she fears to descry. She simply has not the talent to do so, though she shall never admit it. How I despise that upstart Lady Dionli, that usurper of my rightful place in the Ka’lian, whose only real talent is the art of..."
+...the Sapphire of Ice has yet improved even my own talent for the art of scrying. I have discarded my Orb of Pondering in favor of it. Soon I shall peer into the far side of our Faerie World, which that pretender Lady Dionli claims she fears to descry. She simply has not the talent to do so, though she shall never admit it. How I despise that upstart Lady Dionli, that usurper of my rightful place in the Ka’lian, whose only real talent is the art of..."
```
nemaara, line 466: Same here.

data/campaigns/Winds_of_Fate/scenarios/08_Overlook.cfg
```diff
@@ -581,10 +583,10 @@ Now, I must resume my work and you must leave..."
 <i>Faerie</i> creatures have followed you...
 
 If they come much closer <b><u>It</u></b> will see what has been created here.
-The Faerie Realm sees and feels through the senses of all the simpler beings it has transmuted. And if it only glimpses my work in this early phase, we may lose our world’s best hope against it."
+The Faerie World sees and feels through the senses of all the simpler beings it has transmuted. And if it only glimpses my work in this early phase, we may lose our world’s best hope against it."
```
nemaara, line 586: And here.

data/campaigns/Winds_of_Fate/_main.cfg
```diff
@@ -21,9 +21,9 @@
 
 " + _ "(Hard level, 11 scenarios.)"
 
-    {CAMPAIGN_DIFFICULTY EASY      "units/drakes/burner-fly-1.png~CROP(0,0,72,72)~RC(magenta>red)"        ( _ "Aspirant")  ( _ "Normal")}
-    {CAMPAIGN_DIFFICULTY NORMAL    "units/drakes/fire-fire-se-1.png~RC(magenta>red)"                      ( _ "Intendant") ( _ "Challenging")} {DEFAULT_DIFFICULTY}
-    {CAMPAIGN_DIFFICULTY HARD      "units/drakes/inferno-fire-se-3.png~RC(magenta>red)"                   ( _ "Dominant")  ( _ "Difficult")}
+    {CAMPAIGN_DIFFICULTY EASY      "units/drakes/burner-fly-1.png~CROP(0,0,72,72)~RC(magenta>red)"        ( _ "Aspirant")  ( _ "Easy")}
```
nemaara, line 24: I think the easier difficulties might still need to be a bit easier, given that you added the recall costs in here. New players make some surprising choices regarding gold management and exp, so giving them extra gold might be helpful.

data/campaigns/Winds_of_Fate/about.cfg
Jonathan-Kelly: This sprite in particular is GNU GPLv2 from the first(?) release of Era of Magic back in the 1.10 days. I will have to check but I think the others are old enough to be from that same time period where all art was licensed GPL along with the code.

data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
Jonathan-Kelly: This is actually reptilian, avian and mammalian sexual dimorphism, not just human. The woses would be the best fit if you want a hermaphroditic species (as such is very common in the plant kingdom).

data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
Jonathan-Kelly: Correct, the spawn rate is linear and per turn. But there are three more important factors here - population cap, starting population and infighting.

The newly added population cap helps further regulate the highest difficulty. The preexisting infighting factor does the same; each species attacks the others, not just the player side. So as their populations grow, so too does the amount of cross species conflict.

On the other side of things, the starting population (which is the same across all difficulties) keeps the easiest difficulty from getting too easy. The thinking here is that beginner wesnoth players will need quite a lot more time to hunt down and kill the initial set of animals and then heal and regroup their own units. So for them, the re-spawn rate is just a trickle.

I tested and completed all scenarios on nightmare to make sure the top end is at least consistent, though I worry it may be a bit too easy. Beetlenaut seemed to think the last version's nightmare was a bit on the tough side, so hopefully it is close to where it should be even it got a bit easier this time.

data/campaigns/Winds_of_Fate/scenarios/05_Threshold.cfg
Jonathan-Kelly: These characters (upper and lower case "thorn" from the runic alphabet) are no longer part of English but are still used throughout modern day Icelandic. So my guess is any internationally compliant system must support it?

data/campaigns/Winds_of_Fate/scenarios/05_Threshold.cfg
Jonathan-Kelly: The wose unit description refers to it as "faerie world" but somehow I misremembered it as 'realm' when I first wrote this. So the aim here is just better consistency with the wider lore and fewer synonyms for the player to try to remember.

data/campaigns/Winds_of_Fate/_main.cfg
Jonathan-Kelly: I can bring it lower if you wish, but (for scenarios where recruiting is possible) it is currently giving beginner players twice as much gold and income as their opponent team.

In the scenarios where villages are important `village_support=0` for the opponent team essentially means half income and in the final "eyrie scenarios" S10 and S11, players also get double `village_gold` and `village_support`.

data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
nemaara: Good point I totally forgot my bio classes.

data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
nemaara: Yeah you've been doing a good job playlisting on the hard levels. I was more wondering if there were sufficient spawns on easy to look like there was an okay amount of fauna in the scenario?

data/campaigns/Winds_of_Fate/_main.cfg
nemaara: Okay let's let it be as you have it in the pr.

data/campaigns/Winds_of_Fate/about.cfg
nemaara: Can you please check? I would like to make sure we respect inferno8's wishes regarding licensing on his own work.

data/campaigns/Winds_of_Fate/scenarios/05_Threshold.cfg
nemaara: I guess if they display as boxes, someone will let us know. But I'm leaving this unresolved for now since there's a chance someone else may come review before we merge who knows one way or another.

data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
CelticMinstrel: I already assumed the woses were hermaphroditic.

Hermaphroditism is pretty common even among animals (about 5%, apparently). I'm not saying the drakes should be hermaphroditic, but it wouldn't be _that_ strange if they were.

Can someone explain exactly where sexual dimorphism is being exhibited here? Nearly all the text changes are just pronoun changes… I don't see anything anywhere to suggest sexual dimorphism.

(Note: In case someone is confused, "sexual dimorphism" is when the males and females look dramatically different. A good example is ducks. So maybe that wasn't actually what you meant?)

data/campaigns/Winds_of_Fate/about.cfg
Jonathan-Kelly: I did, it is GNU GPLv2. You can confirm by going to https://addons.wesnoth.org/1.10/ and downloading Era of Magic from there. The relevant files are named `COPYING` (the GPL license itself) and `CREDITS` (lists inferno8 as the sole artist). The sprite in question was replaced by another and removed from later versions of EoMa so we have to use the licensing provided in this older version found on the BfW 1.10 addons server archive.

Also I asked inferno8 for his explicit permission to use the sprite in WoF and [he kindly gave it back here.](https://github.com/wesnoth/wesnoth/pull/6875#issuecomment-1186109407)

data/campaigns/Winds_of_Fate/scenarios/01_The_Hunt.cfg
Jonathan-Kelly: My sense is yes but you can judge for yourself- here is a short replay of the default AI playing the first scenario on easy:
[easy_difficulty_test.gz](https://github.com/wesnoth/wesnoth/files/12673842/easy_difficulty_test.gz)
Since each of the 8 species has a 10% chance of spawning per turn (plus four more at 5%) the island tends to stay populated at easy difficulty.

data/campaigns/Winds_of_Fate/scenarios/05_Threshold.cfg
Jonathan-Kelly: It is to give the impression that the text has been incompletely cyphered by a mage of the Silver Order, and may also be an older dialect used for formal writing within the Order. Basically, it is a small word puzzle for players to decypher.

Sounds like the characters will work on Mac then, and I can confirm it works on Windows 10. So that just leaves Linux.

CelticMinstrel: Ultimately whether they work is mostly up to the font used, which if I'm not mistaken is controlled by Wesnoth. So I wouldn't expect this to differ by OS. The thorn character _might_ have an issue appearing in some translations that use completely different character sets which assign a different font, but those translations should've eliminated it in the translated text anyway.

I'm also think the po note for this string is wrong. Explaining the thorn isn't bad but isn't the most important detail. You should somehow incorporate what you said in the previous point into the note. That will allow translators to work out their own way of giving a similar impression.

Jonathan-Kelly: That is a good point. Updated the po note with this information.

data/campaigns/Winds_of_Fate/scenarios/10_Fire_Meets_Steel.cfg
```diff
@@ -319,7 +308,7 @@
                 carryover_percentage=0
             [/gold_carryover]
             [note]
-                description= _ "The new eyrie quadruples village income and trains veteran recruits"
+                description= _ "The new eyrie doubles village output and trains veteran recruits"
```
nemaara, line 311: How about using "income and support" to be clearer?

data/campaigns/Winds_of_Fate/scenarios/11_Crosswind.cfg
```diff
@@ -185,15 +192,15 @@
...
+                description= _ "Death of Resha"
                 condition=lose
             [/objective]
             [objective]
                 description= _ "Death of Zedrix"
                 condition=lose
             [/objective]
             [note]
-                description= _ "The new eyrie quadruples village income and trains veteran recruits"
+                description= _ "The new eyrie doubles village output and trains veteran recruits"
```
nemaara, line 203: Same here.

data/campaigns/Winds_of_Fate/scenarios/10_Fire_Meets_Steel.cfg
```diff
@@ -748,33 +737,33 @@ We shall find no better neighbors than each other."
...
-            speaker=Reshan
+            speaker=Resha
             message= _ "Gorlack, it seems no cave can keep us!"
         [/message]
         [message]
             speaker=Gorlack
             image_pos=right
             mirror=yes
-            message= _ "reshan...
+            message= _ "resha...
```
nemaara, line 757: Capitalize Resha?

data/campaigns/Winds_of_Fate/scenarios/10_Fire_Meets_Steel.cfg
Jonathan-Kelly: Good idea. Applied the change.

data/campaigns/Winds_of_Fate/scenarios/10_Fire_Meets_Steel.cfg
Jonathan-Kelly: Hmm, not sure... was trying to give it the feeling of being said very quietly and softly in disbelief. Wanted to make the tonality sound unique in some way since it is meant to be a turning point for this character. What is your sense on it?

nemaara: I figured that was your idea but to me it comes across as just a typo since everything else is capitalized. This is just current grammatical conventions of course, but just "..." suffices seeing that Gorlack is pretty decisive otherwise. You could even add two of them like "... Resha......" for emphasis if you really wanted. Again mostly just following current conventions.

CelticMinstrel: I don't think making it lowercase gives a sense of being said softly. The prefix `...` might do that, or maybe making it a smaller font size?

Jonathan-Kelly: Alright, capitalized it. The smaller font size is a good idea but was a bit too hard to read, so instead I gave the following exclaimed line a larger font size.

nemaara (approved):
Thanks for addressing feedback! I can merge whenever you're ready.

Jonathan-Kelly:
Thanks for providing feedback. Yeah, this should be ready to merge now.

data/campaigns/Winds_of_Fate/scenarios/05_Threshold.cfg
```diff
@@ -224,62 +213,24 @@
...
-            village_value=0
-            [avoid]
-                terrain=*^V*
-            [/avoid]
-        [/ai]
-    [/side]
-
-    [side]
-        side=6
+        side=13
```
Wedge009, line 228: What's the reason for side 13? Just asking because `wmllint` complains about a side number being out of sequence (I note there are sides 1-5, is this supposed to be side 4?)

Jonathan-Kelly: Sides 4 through 12 are created by the above `SPAWNER_SIDE` macro [calls](https://github.com/wesnoth/wesnoth/blob/8cebaf5b51838d82525835dc67085ec2c9c81433/data/campaigns/Winds_of_Fate/scenarios/05_Threshold.cfg#L216-L224) as defined over [here](https://github.com/wesnoth/wesnoth/blob/8cebaf5b51838d82525835dc67085ec2c9c81433/data/campaigns/Winds_of_Fate/utils/macros.cfg#L97-L111).

I guess since this must have passed the CI checks five months ago there maybe was some update to `wmllint` since then which is causing it to complain?

Wedge009: I thought it might be something like that. Maybe need to add a comment to make this clearer.

I don't think `wmllint` is actually used in CI yet. That's what #7082 is supposed to record.

CelticMinstrel: In that case you should append `# wmllint: skip-side` to every call of the `SPAWNER_SIDE` macro. (Or if the macro call is really long you could put that above each call instead.)

Jonathan-Kelly: Should I do that for the `SPAWNER_SIDE` calls in scenarios 1 and 9 also? They are not causing any complaints from `wmllint`, presumably because there is no explicit `[side]` definition following them with a higher side number. But maybe liberal use of `# wmllint: skip-side` is best practice in general?

---

Looks like `wmllint` has two other complaints:

`rank for mainline campaigns should be a multiple of 5 between 0 and 299 (included)`

`11_Crosswind.cfg", line 169: unknown 'land_on_keep' referred to by id`

Fixing the first involves changing the ranks for WoF, SoF and LoW. The second complaint message seems unclear to me.

CelticMinstrel: > Should I do that for the `SPAWNER_SIDE` calls in scenarios 1 and 9 also? They are not causing any complaints from `wmllint`, presumably because there is no explicit `[side]` definition following them with a higher side number. But maybe liberal use of `# wmllint: skip-side` is best practice in general?

It can't hurt, on the off-chance that someone later on decides to add sides after them.

> `11_Crosswind.cfg", line 169: unknown 'land_on_keep' referred to by id`
> 
> [...] message seems unclear to me.

Without seeing the line in context, my guess would be that it saw `id=land_on_keep` in some tag that has nothing to do with units and assumed it was supposed to refer to a unit.

PR #7909, opened 2023-09-16
Give kraken submerge ability

Jonathan-Kelly:
Scenario designers can finally have access to a sea creature unit with the submerge ability.

The kraken is based on the [giant squid](https://en.wikipedia.org/wiki/Giant_squid) and [colossal squid](https://en.wikipedia.org/wiki/Colossal_squid) which live in the deepest zones of the ocean (they are products of [deep sea gigantism](https://en.wikipedia.org/wiki/Deep-sea_gigantism)) and have chromatophores for some additional camouflage ability. Also the kraken's eerie green sprite lurks low in the water giving it a stealthy appearance:
sneaky_kraken

No balance changes will be needed to DW (where the kraken serves as an advancement for the player's single cuttlefish unit) or WoF (which has already been tested with this). Since it has only recently been added to core, the kraken does not yet appear anywhere else in the game (or UMC) so right now is the perfect time to for this change.

### DIFF
--- a/data/core/units/monsters/Kraken.cfg
+++ b/data/core/units/monsters/Kraken.cfg
@@ -15,11 +15,14 @@
     advances_to=null
     attacks=1
     {AMLA_DEFAULT}
-    cost=62
+    cost=64
     undead_variation=swimmer
     usage=fighter
     description= _ "Krakens are gigantic creatures of the seas. They can grab their opponents with strong tentacles, or spit a poisonous black ink from a distance. The best way to survive an encounter with these monsters is to remain ashore."
     die_sound=water-blast.wav
+    [abilities]
+        {ABILITY_SUBMERGE}
+    [/abilities]
     {DEFENSE_ANIM "units/monsters/kraken/kraken-defend2.png" "units/monsters/kraken/kraken-defend1.png" squishy-hit.wav}
     [attack]
         name=tentacle


Hejnewar:
Assuming that I not at all care about lore and only about making cool units:

This makes it not much more useful, not much more cooler or interesting, it is still stronger cuttlefish with submerge. It limits it's usage on some scenarios, makes it better in others but that's about it. That's problematic because there is nothing really to review and what's left are personal feelings really. "Imo it's good / imo it's not good." And imo creators can add the ability quite easily if necessary. 

If a unit is problematic then it should be reworked not just slightly changed. (But then again it is used in DW as cuttlefish advancement and it has like no relation visually with it.) I could propose something to make it unique but I'm afraid it would end up much different than it is.

Jonathan-Kelly:
> This makes it not much more useful, not much more cooler or interesting, it is still stronger cuttlefish with submerge. It limits it's usage on some scenarios, makes it better in others but that's about it.

The cuttlefish and sea serpent work for those scenarios where submerge might be limiting. And for scenarios with limited deep ocean tiles there is obviously no problem at all.

This is for those scenarios which benefit from having a deep ocean sea monster with submerge. Which is a pretty basic thing for at least one sea monster to be able to do - disappear under the water.

> If a unit is problematic then it should be reworked not just slightly changed. (But then again it is used in DW as cuttlefish advancement and it has like no relation visually with it.)

Can we please stop keep making _perfect the enemy of good?_ At least in this one case where we have this pseudo-advancement requirement from DW.

> I could propose something to make it unique but I'm afraid it would end up much different than it is.

Much different in what ways?

cooljeanius:
N.B. the previous version of this PR was #7125

Hejnewar:
> The cuttlefish and sea serpent work for those scenarios where submerge might be limiting. And for scenarios with limited deep ocean tiles there is obviously no problem at all.

> This is for those scenarios which benefit from having a deep ocean sea monster with submerge. Which is a pretty basic thing for at least one sea monster to be able to do - disappear under the water.

And the one that can do that is the biggest one of them all. 

> Can we please stop keep making perfect the enemy of good? At least in this one case where we have this pseudo-advancement requirement from DW. 

This isnt good. This is neutral. I dont see any advantages of having or not having this ability on this unit. People also are divided or maybe even on negative side. 

> Much different in what ways?

I would give it swamp, shallow and deep water "burrow", which would expand the amount of environoments where it can be encountered or at least that would be encouraged. 2 mp cost in these places also would limite the danger of it there. Or just for deep water.

Another one would probably be actually using ink as intended with disengage instead of poison. Kinda split level could even have some kind of invisibility on it for a turn. (But that is highly PvP nice stuff hard to use by ai, so probably not.) 

But honestly I have nothing against leaving it as is.

cooljeanius:
> I would give it swamp, shallow and deep water "burrow", which would expand the amount of environoments where it can be encountered or at least that would be encouraged. 2 mp cost in these places also would limite the danger of it there. Or just for deep water.
>

I thought making it like the "burrow" ability would also mean that it'd only apply when resting, which I think would be an improvement. I personally do not like getting ambushed much, and would support any measures to reduce the number of situations in which this would apply. 

> Another one would probably be actually using ink as intended with disengage instead of poison. Kinda split level could even have some kind of invisibility on it for a turn. (But that is highly PvP nice stuff hard to use by ai, so probably not.)

I would support giving it disengage.

Hejnewar:
> How is the kraken so different from the woses and swamp lizard? 

Well, wose is a tree in a forest, swapm lizard is green with only eyes visible in green. Kraken is green in blue water. 

> Anything else

Honestly I would just support new unit or unit line for that instead.

Jonathan-Kelly:
> Well, wose is a tree in a forest, swapm lizard is green with only eyes visible in green. Kraken is green in blue water.

The kraken is literally a sea-green sea monster. Sprites can also be recolored fairly easily since they use a small color palette, so if you want it to be blue, it can be blue.

Jonathan-Kelly:
Well it has been a year since this began and now at the string freeze it seems unlikely to go any further.

I just hope that during 1.19, monsters will be balanced towards having some actually formidable stats and diverse abilities. In a game where the player controls an army, it is not like a single level 2/3/4 unit can pose an interesting threat, let alone fit the term "monster".

PR #7913, opened 2023-09-19
add `"pkgconf"` to `vcpkg.json`

Chaojimengnan:
When I was compiling on Windows, I got an error:

```text
[cmake] CMake Error at D:/program/cmake/share/cmake-3.24/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
[cmake]   Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
```

PkgConfig installation is quite troublesome in Windows, but vcpkg provides it for us, so we only need to add the following line in `vcpkg.json` to automatically add Pkgconfig during Windows build:

```json
    {
      "name": "pkgconf",
      "platform": "windows"
    }
```


Pentarctagon:
The note in https://github.com/wesnoth/wesnoth/blob/master/INSTALL.md#windows should also be removed then.

Pentarctagon:
Thanks!

PR #7915, opened 2023-09-19
Add ambusher portrait to core from SotBE

Jonathan-Kelly:
Apparently we have had a saurian ambusher portrait for the past fifteen years. It was only being used with a minor character in one campaign.


doofus-01 (commented):
It's a character portrait and it's showing its age.  It wouldn't stand out compared to the other saurians and drakes, I guess.  If there's an overwhelming consensus that it should move, I won't stand in the way, but I'm not in favor of it.

Jonathan-Kelly:
Is it really better to keep using the skirmisher portrait as a placeholder for the ambusher (and flanker) when we have an actual ambusher portrait done in the same style and quality level?
skirmisher
ambusher
augur

doofus-01:
Updates to the saurians would be welcome, but the general "move all mainline campaign assets to core" mindset that seems to have taken hold here weakens the campaigns and adds little to the Wesnoth project overall.

Is it worth it?  I don't think it is, but at least a few others clearly think differently.  This isn't the hill I want to die on, so like I said earlier, I won't block it.

The good news here is that this doesn't have to be decided before the feature freeze.  New images can be added to stable releases, though existing images can't be deleted or moved.

CelticMinstrel:
I don't have an opinion on the topic of whether this should be moved or not, but I do want to note that that portrait is intended to refer to Inarix, not a saurian ambusher. The fact that Inarix happens to be a saurian ambusher is just coincidence; it doesn't necessarily mean the portrait is suitable as a generic saurian ambusher portrait.

I don't intend to block this move. I do think the reasoning for it is a little lacking.

Jonathan-Kelly:
> Updates to the saurians would be welcome, but the general "move all mainline campaign assets to core" mindset that seems to have taken hold here weakens the campaigns and adds little to the Wesnoth project overall.

Well, I do not really subscribe to that mindset either, and I have actually spoken against it in a number of cases, especially where it would lower quality standards or add redundancy.

In the specific case of this PR, the areas of mainline which would benefit are mostly the campaigns themselves:
* Winds of Fate, wherein a saurian ambusher is a primary character.
* All other campaigns involving saurians, especially where an ambusher or flanker has a speaking role.
* All MP eras.

> The good news here is that this doesn't have to be decided before the feature freeze. New images can be added to stable releases, though existing images can't be deleted or moved.

Deciding after that point means unnecessary data duplication, so we should really try to get it done before then.

> The fact that Inarix happens to be a saurian ambusher is just coincidence; it doesn't necessarily mean the portrait is suitable as a generic saurian ambusher portrait.

True, though my (subjective) proposal here it that this is indeed suitable as the generic ambusher portrait which the game is still lacking. Also, this portrait does not seem to portray anything specific to the character of Inarix _besides_ the fact she is a saurian ambusher. So I am curious what others feel.

Jonathan-Kelly:
@doofus-01 time is nearly up so the choice is yours to merge before the freeze or else I will close once it arrives.

For the record, it makes sense to me not to merge this if you feel the art is lower quality or stylistically mismatched from that of the skirmisher and augur. Likewise if you feel the art speaks to the nature of a particular saurian individual better than the saurian ambushers as a whole.

doofus-01:
>  time is nearly up so the choice is yours to merge before the freeze or else I will close once it arrives.

I'm not inclined to pull the trigger on this, but if you want to close it and file an issue so the thought stays alive, that might be a way to split the difference.  It's not that I think the style is so mismatched or piss-poor, it's that the style is also a good fit with the campaign portraits (same artist did them all), and I'm not in favor of chipping away at the campaign for this.  I also suspect the artist and the art director knew about this and they didn't make the move.

Similar to what I said earlier, we can add images at any time.  It's not completely unthinkable that the core saurian portraits could be replaced in the coming year.  It's also possible that the inarix.webp could be edited to make a portrait that's different enough.

PR #7916, opened 2023-09-24
Add game logic to Forest of Fear

Jonathan-Kelly:
Similarly to #7792 this adds game logic to Forest of Fear rewarding taking and holding the central castle. Each turn, dangerous forest wildlife and magical creatures enter from the edges of the map to attack players and each other (while ignoring villages). Yet these creatures will ally with whichever player occupies the faerie keep using their leader, and for as long as they do. Which means both relief from their relentless attacks and the ability to see through their eyes (`share_vision`) by the magic of the faerie throne.

As with King of the Hill, each side gets two leaders and whenever one dies another replacement leader is always available from the recall list at its full recruiting cost. And a treasure chest of gold (which only a leader can unlock) awaits the first player to claim the center.

### DIFF
--- a/data/multiplayer/maps/5p_Forest_of_Fear.map
+++ b/data/multiplayer/maps/5p_Forest_of_Fear.map
@@ -12,7 +12,7 @@ Gs^Fp, Gs^Fp, Re, Ww, Gs^Fms, Mm, Hh, Rp, Gs^Fp, Rd, Rd, Rd, Gg, Rd, Rp, Rp, Hh,
 Ss, Ss, Gg, Re, Gg, Re^Tf, Hh, Rp, Gg, Gg, Gs^Fds, Hh, Hh, Gg, Rp, Gg^Vh, Gs^Fms, Mm, Gg, Wwf, Ww, Mm, Gs^Fp, Re, Gg^Ve, Gs^Fp, Mm, Gg, Rp, Gs^Fp, Hh, Hh
 Ss, Ss, Mm, Re, Gg, Gg, Gs^Fp, Rp, Gg, Gs^Fms, Gs^Fp, Mm, Ss, Gs^Fp, Ww^Bw|, Hh, Ww, Hh, Wwf, Ww^Vm, Ww, Gs^Fms, Re, Re, Hh, Hh, Gs^Fms, Gs^Fms, Rp, Hh, Gs^Fp, Gs^Fp
 Ss, Ss, Hh, Re, Re, Gg, Rp, Rp, Re, Gg, Ss^Vhs, Hh, Ss, Ss, Chs, Ss, Chw, Ww, Ww, Wwf, Gg, Gs^Fds, Re, Gg, Gs^Fp, Re^Tf, Hh, Gs^Fds, Rp, Gg, Hh, Hh
-Gs^Fp, Gs^Fp, Gs^Fms, Gg^Vh, Gg, Rp, Gs^Fp, Hh, Hh, Re, Gs^Fms, Ss, Ww, Gg, Gg, Chr, Khw, Chw, Ww, Ww, Hh, Re, Mm, Mm, Mm, Hh, Rd, Rd, Gg^Vh, Rp, Gs^Fp, Gs^Fp
+Gs^Fp, Gs^Fp, Gs^Fms, Gg^Vh, Gg, Rp, Gs^Fp, Hh, Hh, Re, Gs^Fms, Ss, Ww, Gg, Gg, Chr, 6 Khw, Chw, Ww, Ww, Hh, Re, Mm, Mm, Mm, Hh, Rd, Rd, Gg^Vh, Rp, Gs^Fp, Gs^Fp
 Gs^Fp, Re^Tf, Hh, Gg, Gs^Fp, Rp, Rp, Gg, Mm, Re, Re, Hh, Mm, Wwf, Hh, Chr, Ww, Ww^Bw\, Gs^Fp, Gs^Fp, Hh^Vhh, Re, Gs^Fp, Mm, Gs^Fds, Rd, Gg, Gs^Fp, Hh, Rp, Gg, Gg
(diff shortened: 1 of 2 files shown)


Hejnewar:
I have just one question and that is, why?

Jonathan-Kelly:
In order to:

* Give an old and no longer popular map some more substance along the lines of its theme. It is called _Forest of Fear_, so what can be fearsome about a forest? The wild animals which inhabit it - wolves, bears, and huge prehistoric dragonflies!

* Expand the _King of the Hill/Castle_ multiplayer game type beyond just one scenario. Eventually we could even have one of these for each faction; each disputed castle granting a faction themed bonus (in this case it is an elven nature bonus).

* Help achieve the goal of adding PvPvE scenarios to the game as per the 1.17 roadmap:
`1.17.12 (01/15/2023)`
`New multiplayer maps focused on player vs player vs ai (Hejnewar)`

Hejnewar:
> Give an old and no longer popular map some more substance along the lines of its theme. It is called Forest of Fear, so what can be fearsome about a forest? The wild animals which inhabit it - wolves, bears, and huge prehistoric dragonflies!

There is a reason it is unpopular, its bad. Tho 5p maps are generally probably rarely played, 6p can be seen quiet regularily if there is enough people.

> Expand the King of the Hill/Castle multiplayer game type beyond just one scenario. Eventually we could even have one of these for each faction; each disputed castle granting a faction themed bonus (in this case it is an elven nature bonus).

No, I dont think that repeated content is good. Even with some twist it just doesnt provide something new for players and if you'r worried that super fans of this mode will only have one map to play then the quality should be at least on pair with the original. And is it? It isnt, the map is way worse despite being open (which is good). The gimick feels somewhat wierd since the animals fight against each other and have very random spawn points it seems so some player can be kinda left out. 

Perhaps making waves and making it more survivalish would be better but then the control (thats pretty op) shouldnt be there. 

> Help achieve the goal of adding PvPvE scenarios to the game as per the 1.17 roadmap:
1.17.12 (01/15/2023)
New multiplayer maps focused on player vs player vs ai (Hejnewar)

Reality verified.

Jonathan-Kelly:
> Tho 5p maps are generally probably rarely played, 6p can be seen quiet regularily if there is enough people.

That may be because this is the only actual 5p map in mainline? Since _Auction X_ and _The Wilderlands_ generally have the computer take side 5.

> No, I dont think that repeated content is good.

Well, I agree. Of the about sixty multiplayer scenarios, over fifty have no scenario mechanics at all. They are just blank PvP maps. So the goal here is to do something different from that status quo (as well as from the new KotH).

> It isnt, the map is way worse despite being open (which is good).

What would you like to see changed about the map?

> The gimick feels somewhat wierd since the animals fight against each other and have very random spawn points it seems so some player can be kinda left out.

The animal quarrels are just to make the world seem more alive and slightly weaken the faerie keep bonus, but it can be turned off by default, if it is a big deal?

The spawn points are all along the edges of the map, but each species spawns on one of its "native" terrain types. Do you want me to disable terrain based spawning so it is a bit more a bit more like Dark Forecast?

> Perhaps making waves and making it more survivalish would be better but then the control (thats pretty op) shouldnt be there.

Well those changes would make it more like existing content, which we agree is less desirable. Also note that the center bonus does not give a player control over the wildlife, it only makes them allies (they remain computer controlled). But if this bonus is still too strong then spawn rates or population caps for the wildlife could be lowered or their sharing vision could be removed.

Hejnewar:
> Well, I agree. Of the about sixty multiplayer scenarios, over fifty have no scenario mechanics at all. They are just blank PvP maps. So the goal here is to do something different from that status quo (as well as from the new KotH).

Yeah and thats good thing, you dont need mechanics if youjast want to enjoy pvp. 

> What would you like to see changed about the map?

For KotH this map is trully terrible, the distances are very uneven (2-4 turns), villages are really uneven (some players have many safe vills some literally 0), the bridge meta in certain points, to a lesser extent terrain. If I can see that in like just launching the scenario then Im quiet concerened that there wasnt enough thought put into this and there was no playtesting. 

> spawns

The problem with spawns is that they dont really create much pressure, are mostly random, often light themselves along the border, now that Im listing it... they dont quiet benefit anything. 

> Well those changes would make it more like existing content, which we agree is less desirable. 

Yes. It is making it more like DF, but its making it was less like KotH. 

TL;DR: 
Throwing a mechanic customised to certain map on top of an existing map and expecing it to just work well is bad idea.

Jonathan-Kelly:
> For KotH this map is trully terrible, the distances are very uneven (2-4 turns), villages are really uneven (some players have many safe vills some literally 0), the bridge meta in certain points, to a lesser extent terrain.

Hmm.... by themselves, those things may not be too hard to fix. If you want the player count changed (reduced) from 5, then that frees more room to reposition the castles and villages. Some reshaping of the waterways could fix the bridge issue, as it did with KotH's map.

> The problem with spawns is that they dont really create much pressure, are mostly random, often light themselves along the border,

So would increasing (doubling?) the spawn rate and more evenly spreading the spawn points address these particular problems?

Hejnewar:
At this point you could as well make new more adapted map instead of changing this cuz imo that would be similar amout of work. 

The spawn rate isnt the problem the behaviour is, randomness of it is also a problem kinda. Also imo forest animals arent that scary.

Jonathan-Kelly:
> At this point you could as well make new more adapted map instead of changing this cuz imo that would be similar amout of work.

Well, I do not mind the work of renovating an older map, but this amount of change would essentially overwrite it. Since _Forest of Fear_ is something which has been a part of the game in some form or another for [almost twenty years](https://github.com/wesnoth/wesnoth/commit/36f4fa7f6142cf8e31c21bd487ca25a1a5a1d644), it seems preferable to close this pull request instead.

A PvPvE survival scenario seems like it should be fun, even if it needs some refinements. Like your idea of clumping the spawns into waves or having as many faeries thrones as the number of players minus one, so that it becomes a game of _"musical chairs"_ rather than _"king of the hill"_. With the freeze now only two weeks away, it seems this will not be happening for 1.18, unfortunately,

Hejnewar:
Honestly it pretty much never had a chance, too little time and I personally dont have time to oversee it like last time and otherwise I didnt see it passing. 

Honestly x2, its good that you have the will to work but you dont have the feel for the game, take your time go play mp or something.

PR #7920, opened 2023-09-27
Mousing over a village show the owner side in the UI

AugustVesterbacka:
Solves  #4344

Shows the village owner side when a tile is hovered in the terrain info field.

Fits the field well when the terrain code is two or less (type, village), 
I have not find a scenario where these have been more, and players are less than 10. 
If these conditions are not met the number would go outside the box, but you could still see the flag color clearly.

VillageOwnerSide


Wedge009:
Thanks for that. The change seems to work as intended for me. I don't know if there's a better place to put the side icon, but the current set-up seems like a reasonable spot to me.

Not sure what's going on with the failing check.

stevecotton:
The information available should be limited if the hex is under fog of war, as the code about 50 lines between the new code does.

src/reports.cpp
```diff
@@ -1575,6 +1575,26 @@ REPORT_GENERATOR(terrain_info, rc)
...
+
+			std::string flag_icon = owner_team.flag_icon();
+			std::string old_rgb = game_config::flag_rgb;
+			std::string new_rgb = team::get_side_color_id(owner_team.side());
+			std::string mods = "~RC(" + old_rgb + ">" + new_rgb + ")";
+			if (flag_icon.empty())
+				flag_icon = game_config::images::flag_icon;
+			std::string tooltip = side_tooltip(owner_team);
+			std::string side = std::to_string(owner_team.side());
+
```
stevecotton, line 1592: The CI is failing because of trailing whitespace on this line.

src/reports.cpp
```diff
@@ -1575,6 +1575,26 @@ REPORT_GENERATOR(terrain_info, rc)
...
+		int owner = rc.dc().village_owner(mouseover_hex);
+		if (owner != 0) {
+			const team &owner_team = rc.dc().get_team(owner);
+
+			std::string flag_icon = owner_team.flag_icon();
+			std::string old_rgb = game_config::flag_rgb;
+			std::string new_rgb = team::get_side_color_id(owner_team.side());
+			std::string mods = "~RC(" + old_rgb + ">" + new_rgb + ")";
+			if (flag_icon.empty())
+				flag_icon = game_config::images::flag_icon;
```
soliton-, line 1589: Please use parens or write on a single line.

stevecotton: I'd prefer parens instead of the single-line style.

CelticMinstrel: Don't you mean braces, not "parens"?

src/reports.cpp
```diff
@@ -1575,6 +1575,26 @@ REPORT_GENERATOR(terrain_info, rc)
 			continue;
 		blit_tced_icon(cfg, terrain_id, terrain_icon, high_res, terrain_name);
 	}
+
+	if (map.is_village(mouseover_hex)) {
```
CelticMinstrel, line 1579: This file doesn't seem to have been updated, but I think our style is `if(` without the space.

Pentarctagon:
FYI adding `resolves #4344` to the commit message will link the issue to this PR, so the issue will be automatically closed when the PR merges.

CelticMinstrel:
That could also be done by us if we squash merge.

CelticMinstrel:
Can someone cherry-pick that reformat commit to master? (Or maybe re-run it on master if it doesn't apply cleanly enough.) I don't like having a reformat in the PR as it interferes with judging the PR itself.

soliton-:
Unless we want to run clang-format on everything I don't think that is useful. Better to revert the reformatting IMO.

soliton- (approved):
Looks good to me.

(The copyright job fails but I assume that is unrelated.)

AugustVesterbacka:
I think an issue should be opened for reformatting the reports.cpp file, for code consistency and ensuring the wesnoth standard.

Pentarctagon:
Was there anything preventing this from being merged?

AugustVesterbacka:
Accidentally closed it

soliton-:
Actually I get an out of range exception in the editor when hovering over a village. In scenario mode it works fine. I'm guessing the owner_team stuff needs checking.

CelticMinstrel:
Why was terrain_info.cpp (seemingly an unrelated file) reformatted in this PR?

stevecotton:
> Why was terrain_info.cpp (seemingly an unrelated file) reformatted in this PR?

That's already been fixed (at least, as long as we remember to squash-merge).

Pentarctagon:
Was the issue Soliton mentioned addressed?

stevecotton:
Oops - somehow I overlooked that. Testing the editor now.

CelticMinstrel:
I don't like how all those unrelated braces were added in reports.cpp, but I guess it's only a few so not _too_  bad…

Issue #7923, opened 2023-09-29
config_filter.cpp functions assume the default value is always zero


newfrenchy83:
### Game and System Information

- **Version:**1.17
- **Downloaded from:**source
- **Build info:**gcc
- **OS:**windows10

### Description of the bug

when I use the filter [filter_ability] to only detect abilities with a value between 0 and 20, the linked event is triggered by uints with abilities not having the 'value' attribute but corresponding to the other criteria because the filter assigns the value 0 by default to 'value', 'add' and 'sub' and the value 1 to 'multiply' and 'divide'.

### Steps to reproduce the behavior

implement an event with [filter][filter_ability]value=0-20
2 try to trigger the event with an ability that does not have the 'value' attribute

The event should not be triggered, which is not the case because the system understand what value=0 then what this attribute is absent.

### Expected behavior

theevent should be trigered what if unit own abilité wit attribute 'value' and value between 0 and 20.

newfrenchy83:
I just realized I made an interpretation error, in fact it is because the code attributes a default value to the value type attributes that the bug appears since the bu seems to be limited to the test of https://github .com/wesnoth/wesnoth/pull/7746 when I try to test using the filter[filter_specials][not]value=0-20 so that [damage]value=5 is not overwritten but then[damage] add=11 is not either but everything works again if I use an interval 1-20

soliton-:
It's very difficult to figure out what issue is described here. Please provide a complete example that reproduces the issue.

newfrenchy83:
> It's very difficult to figure out what issue is described here. Please provide a complete example that reproduces the issue.

i updated description and linked a PR for fix this issue.

soliton-:
So the issue is filtering for abilities with `value=0`. Your expectation is that this will not match any abilities that do not use the `value` key at all and it will not match abilities that do use the `value` key but where `value=0` was _not_ explicitely specified but is merely the default.

Is that understanding correct?

soliton-:
From https://wiki.wesnoth.org/AbilitiesWML it does not look like there are any abilities where `value` has a default though.

newfrenchy83:
> So the issue is filtering for abilities with `value=0`. Your expectation is that this will not match any abilities that do not use the `value` key at all and it will not match abilities that do use the `value` key but where `value=0` was _not_ explicitely specified but is merely the default.
> 
> Is that understanding correct?

yes, not key not value is not explicitely specified.

newfrenchy83:
[...]

PR #7924, opened 2023-09-29
repair detects non-existent default attribute value in numérical attribute

newfrenchy83:
Fix https://github.com/wesnoth/wesnoth/issues/7923 .when we want to detect an ability with value=0-20 and an ability with add=15 is present, this is detected because the system assigns a value of 0 when 'value' is not present, and so the system matches abilities that do not use 'value' or even no numerical value at all. it is therefore necessary to decide that the absence of an attribute is a non-correspondence to the criteria.

### DIFF
--- a/data/schema/game_config.cfg
+++ b/data/schema/game_config.cfg
@@ -41,6 +41,14 @@
         value="none|one_side|both_sides"
     [/type]
     [type]
+        name="bool_or_empty"
+        [list]
+            [element]
+                value="true|false|yes|no|none"
+            [/element]
+        [/list]
+    [/type]
+    [type]
         name="addon_type"
         value="sp|mp|hybrid"
     [/type]
@@ -423,6 +431,7 @@
     {SUBST_TYPE anim_hits}
     {SUBST_TYPE f_int}
     {SUBST_TYPE alignment}
+	{SUBST_TYPE bool_or_empty}
     [type]
         name="global"
         value="global"
--- a/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
+++ b/data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
@@ -307,3 +307,140 @@
 )}
 
 #undef FILTER_ABILITY_TEST
+
+##
+# Common code for the tests in this file.
+##
+# Actions:
+# Give both Alice and Bob 100% chance to hit.
+# Give Alice a drains ability to value 50(defined in hardcoding by default).
+# During Alice's turn, move Alice next to Bob, and have Alice attack Bob.
+# During Bob's turn, have Bob attack Alice.
+##
+#define FILTER_ABILITY_TEST_WITHOUT_VALUE
+    [event]
+        name=start
+        # Make sure the attacks hit
+        {FORCE_CHANCE_TO_HIT (id=bob) (id=alice) 100 ()}
+        {FORCE_CHANCE_TO_HIT (id=alice) (id=bob) 100 ()}
+        [modify_unit]
+            [filter]
+            [/filter]
+            # Make sure they don't die during the attacks
+            [status]
+                invulnerable=yes
+            [/status]
+        [/modify_unit]
+        [object]
+            silent=yes
+            [effect]
+                apply_to=new_ability
+                [abilities]
+                    [drains]
+                    [/drains]
+                [/abilities]
+            [/effect]
+            [filter]
+                id=alice
+            [/filter]
+        [/object]
+        {VARIABLE triggers 0}
+    [/event]
+
+    [event]
+        name=side 1 turn 1
+        [do_command]
+            [move]
+                x=7,13
+                y=3,4
+            [/move]
+            [attack]
+                [source]
+                    x,y=13,4
+                [/source]
+                [destination]
+                    x,y=13,3
+                [/destination]
+            [/attack]
+        [/do_command]
+        [end_turn][/end_turn]
+    [/event]
+
+    [event]
+        name=side 2 turn
+        [do_command]
+            [attack]
+                [source]
+                    x,y=13,3
+                [/source]
+                [destination]
+                    x,y=13,4
+                [/destination]
+            [/attack]
+        [/do_command]
+        [end_turn][/end_turn]
+    [/event]
+#enddef
+
+#####
+# API(s) being tested: [event][filter][filter_ability]
+##
+# Actions:
+# Use the common setup from FILTER_ABILITY_TEST_WITHOUT_VALUE.
+# Add an event with a filter matching Alice's drains ability who had no value specified.
+# Alice attacks Bob and then Bob attacks Alice, as defined in FILTER_ABILITY_TEST_WITHOUT_VALUE.
+##
+# Expected end state:
+# The filtered event is triggered exactly once.
+#####
+{GENERIC_UNIT_TEST event_test_filter_ability_with_value_by_default (
+    {FILTER_ABILITY_TEST_WITHOUT_VALUE}
+    [event]
+        name=attack
+        first_time_only=no
+        [filter]
+            [filter_ability]
+                tag_name=drains
+                value=50
+            [/filter_ability]
+        [/filter]
+        {ASSERT ({VARIABLE_CONDITIONAL side_number equals 1})}
+        {VARIABLE_OP triggers add 1}
+    [/event]
+    [event]
+        name=turn 2
+        {RETURN ({VARIABLE_CONDITIONAL triggers equals 1})}
+    [/event]
+)}
+
+#####
+# API(s) being tested: [event][filter][filter_ability]
+##
+# Actions:
+# Use the common setup from FILTER_ABILITY_TEST_WITHOUT_VALUE.
+# Add an event with a filter matching Alice's drains ability who had no value specified.
+# Alice attacks Bob and then Bob attacks Alice, as defined in FILTER_ABILITY_TEST_WITHOUT_VALUE.
+##
+# Expected end state:
+# The filtered event is never triggered.
+#####
+{GENERIC_UNIT_TEST event_test_filter_ability_no_match_by_default (
+    {FILTER_ABILITY_TEST_WITHOUT_VALUE}
+    [event]
+        name=attack
+        first_time_only=no
+        [filter]
+            [filter_ability]
+                tag_name=drains
+                value=35
+            [/filter_ability]
+        [/filter]
+        {FAIL}
+    [/event]
+    [event]
+        name=turn 2
+        {SUCCEED}
+    [/event]
+)}
+
+#undef FILTER_ABILITY_TEST_WITHOUT_VALUE
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -1435,7 +1435,7 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 	if(!bool_matches_if_present(filter, cfg, "affect_self", true))
 		return false;
 
-	if(!bool_matches_if_present(filter, cfg, "affect_allies", true))
+	if(!bool_or_empty(filter, cfg, "affect_allies"))
 		return false;
 
 	if(!bool_matches_if_present(filter, cfg, "affect_enemies", false))
@@ -1460,8 +1460,28 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 	if(!string_matches_if_present(filter, cfg, "active_on", "both"))
 		return false;
 
-	if(!int_matches_if_present(filter, cfg, "value"))
-		return false;
+	if(!filter["value"].empty()){
+		bool has_other_key = (!cfg["add"].empty() || !cfg["sub"].empty() || !cfg["multiply"].empty() || !cfg["divide"].empty());
+		if(!has_other_key){
+			if(tag_name == "drains"){
+				if(!int_matches_if_present(filter, cfg, "value", 50)){
+					return false;
+				}
+			} else if(tag_name == "berserk"){
+				if(!int_matches_if_present(filter, cfg, "value", 1)){
+					return false;
+				}
+			} else if(tag_name == "heal_on_hit" || tag_name == "heals" || tag_name == "regenerate" || tag_name == "leadership"){
+				if(!int_matches_if_present(filter, cfg, "value" , 0)){
+					return false;
+				}
+			}
+		} else {
+			if(!int_matches_if_present(filter, cfg, "value")){
+				return false;
+			}
+		}
+	}
 
 	if(!int_matches_if_present_or_negative(filter, cfg, "add", "sub"))
 		return false;
--- a/src/utils/config_filters.cpp
+++ b/src/utils/config_filters.cpp
@@ -32,7 +32,7 @@ bool utils::config_filters::bool_matches_if_present(const config& filter, const 
 bool utils::config_filters::string_matches_if_present(
 	const config& filter, const config& cfg, const std::string& attribute, const std::string& def)
 {
-	if(filter[attribute].empty()) {
+	if(!filter.has_attribute(attribute)) {
 		return true;
 	}
 
@@ -43,43 +43,81 @@ bool utils::config_filters::string_matches_if_present(
 
 bool utils::config_filters::unsigned_matches_if_present(const config& filter, const config& cfg, const std::string& attribute)
 {
-	if(filter[attribute].empty()) {
+	if(!filter.has_attribute(attribute)) {
 		return true;
+	}
+	if(!cfg.has_attribute(attribute)) {
+		return false;
 	}
 
 	return in_ranges<int>(cfg[attribute].to_int(0), utils::parse_ranges_unsigned(filter[attribute].str()));
 }
 
-bool utils::config_filters::int_matches_if_present(const config& filter, const config& cfg, const std::string& attribute)
+bool utils::config_filters::int_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, std::optional<int> def)
 {
-	if(filter[attribute].empty()) {
+	if(!filter.has_attribute(attribute)) {
 		return true;
 	}
+	if(!cfg.has_attribute(attribute) && !def) {
+		return false;
+	}
 
-	return in_ranges<int>(cfg[attribute].to_int(0), utils::parse_ranges_int(filter[attribute].str()));
+	int value_def = def ? (*def) : 0;
+	return in_ranges<int>(cfg[attribute].to_int(value_def), utils::parse_ranges_int(filter[attribute].str()));
 }
 
 bool utils::config_filters::int_matches_if_present_or_negative(
-	const config& filter, const config& cfg, const std::string& attribute, const std::string& opposite)
+	const config& filter, const config& cfg, const std::string& attribute, const std::string& opposite, std::optional<int> def)
 {
-	if(int_matches_if_present(filter, cfg, attribute)) {
+	if(int_matches_if_present(filter, cfg, attribute, def)) {
 		return true;
 	}
 
-	// don't check for !cfg[opposite].empty(), as this assumes a default value of zero (similarly to
-	// int_matches_if_present)
-	if(cfg[attribute].empty()) {
-		return in_ranges<int>(-cfg[opposite].to_int(0), utils::parse_ranges_int(filter[attribute].str()));
+	// Check if cfg[opposite].empty() and have optional def.
+	// If def don't exist return false.
+	if(!cfg.has_attribute(attribute)) {
+		if(!cfg.has_attribute(opposite) && !def) {
+			return false;
+		}
+		int value_def = def ? (*def) : 0;
+		return in_ranges<int>(-cfg[opposite].to_int(value_def), utils::parse_ranges_int(filter[attribute].str()));
 	}
 
 	return false;
 }
 
-bool utils::config_filters::double_matches_if_present(const config& filter, const config& cfg, const std::string& attribute)
+bool utils::config_filters::double_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, std::optional<double> def)
 {
-	if(filter[attribute].empty()) {
+	if(!filter.has_attribute(attribute)) {
+		return true;
+	}
+	if(!cfg.has_attribute(attribute) && !def) {
+		return false;
+	}
+
+	double value_def = def ? (*def) : 1;
+	return in_ranges<double>(cfg[attribute].to_double(value_def), utils::parse_ranges_real(filter[attribute].str()));
+}
+
+bool utils::config_filters::bool_or_empty(const config& filter, const config& cfg, const std::string& attribute)
+{
+	if(!filter.has_attribute(attribute)) {
 		return true;
 	}
 
-	return in_ranges<double>(cfg[attribute].to_double(1), utils::parse_ranges_real(filter[attribute].str()));
+	std::set<std::string> filter_attribute = utils::split_set(filter[attribute].str());
+	bool is_matches = false;
+	if(cfg.has_attribute(attribute)){
+		if(filter_attribute.count("yes") != 0 || filter_attribute.count("true") != 0){
+			is_matches = cfg[attribute].to_bool();
+		}
+		if(!is_matches && (filter_attribute.count("no") != 0 || filter_attribute.count("false") != 0)){
+			is_matches = !cfg[attribute].to_bool();
+		}
+	} else {
+		if(filter_attribute.count("none") != 0){
+			is_matches = true;
+		}
+	}
+	return is_matches;
 }
--- a/src/utils/config_filters.hpp
+++ b/src/utils/config_filters.hpp
@@ -34,8 +34,14 @@ namespace utils::config_filters
  */
 bool bool_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, bool def);
 
-bool double_matches_if_present(const config& filter, const config& cfg, const std::string& attribute);
-bool int_matches_if_present(const config& filter, const config& cfg, const std::string& attribute);
+/**
+ * Checks whether the filter matches the value of @a cfg[@a attribute]. If @a cfg doesn't have that
+ * attribute, assume that an unset value is equivalent to @a def if exist, else value false is returned.
+ *
+ * Always returns true if the filter puts no restriction on the value of @a cfg[@a attribute].
+ */
+bool double_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, std::optional<double> def = NULL);
+bool int_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, std::optional<int> def = NULL);
 
 /**
  * Restricts filters to only looking for values that are zero or more.
@@ -50,15 +56,17 @@ bool unsigned_matches_if_present(const config& filter, const config& cfg, const 
  * `add=1` or `sub=-1`; this assumes that code elsewhere has already checked that cfg contains at most one of those
  * keys.
  *
- * This only checks for the presence of @a attribute in the filter, so the caller should call this function a second
- * time, with @a attribute and @a opposite reversed.
+ * This only checks for the presence of @a attribute and @a opposite in the filter, so the caller should call this function a second
+ * time, with @a attribute and @a opposite reversed and if none of these attribute is here value false is returned.
  *
  * The function is named "negative" in case we later want to add a "reciprocal" for the "multiply"/"divide" pair.
  */
 bool int_matches_if_present_or_negative(
-	const config& filter, const config& cfg, const std::string& attribute, const std::string& opposite);
+	const config& filter, const config& cfg, const std::string& attribute, const std::string& opposite, std::optional<int> def = NULL);
 
 bool string_matches_if_present(
 	const config& filter, const config& cfg, const std::string& attribute, const std::string& def);
+
+bool bool_or_empty(const config& filter, const config& cfg, const std::string& attribute);
 
 } // namespace utils::config_filters
--- a/data/schema/filters/abilities.cfg
+++ b/data/schema/filters/abilities.cfg
@@ -14,7 +14,7 @@
 	{SIMPLE_KEY divide s_real_range_list}
 	{SIMPLE_KEY affect_adjacent s_bool}
 	{SIMPLE_KEY affect_self s_bool}
-	{SIMPLE_KEY affect_allies s_bool}
+	{SIMPLE_KEY affect_allies bool_or_empty}
 	{SIMPLE_KEY affect_enemies s_bool}
 	{FILTER_BOOLEAN_OPS abilities}
 [/tag]
--- a/src/tests/test_config_filters.cpp
+++ b/src/tests/test_config_filters.cpp
@@ -84,4 +84,14 @@ BOOST_AUTO_TEST_CASE(test_int_add_sub_filter)
 	BOOST_ASSERT(int_matches_if_present_or_negative(sub_2_to_4, add_minus_3, "sub", "add"));
 }
 
+BOOST_AUTO_TEST_CASE(test_without_attribute_filter)
+{
+	config add_3 {"add", 3};
+	config value_3 {"value", 3};
+
+	BOOST_ASSERT(!int_matches_if_present(value_3, add_3, "value"));
+	BOOST_ASSERT(int_matches_if_present(value_3, add_3, "value", 3));
+	BOOST_ASSERT(!int_matches_if_present_or_negative(add_3, value_3, "add", "sub"));
+}
+
 BOOST_AUTO_TEST_SUITE_END()
--- a/wml_test_schedule
+++ b/wml_test_schedule
@@ -149,6 +149,8 @@
 0 event_test_filter_ability_no_match_neg_prefix
 0 event_test_filter_ability_active
 0 event_test_filter_ability_active_inactive
+0 event_test_filter_ability_with_value_by_default
+0 event_test_filter_ability_no_match_by_default
 0 test_ability_id_active
 0 test_ability_id_not_active
 0 event_test_filter_attack


newfrenchy83:
@Pentarctagon please review thisPR in priority, it's important.

Pentarctagon:
Can you add a unit test for this as well?

soliton-:
I don't think this is the right thing to do. Instead of hard-coding default/no default the functions should be changed to take the default as an optional or variant or so so the caller can signal no default. And in the case of no default the config parameter in _all_ the "matches_if_present" functions should be checked for existence. Also I think you should use `blank()` not `empty()` to check.

newfrenchy83:
> I don't think this is the right thing to do. Instead of hard-coding default/no default the functions should be changed to take the default as an optional or variant or so so the caller can signal no default. And in the case of no default the config parameter in _all_ the "matches_if_present" functions should be checked for existence. Also I think you should use `blank()` not `empty()` to check.

option addedin filter with https://github.com/wesnoth/wesnoth/pull/7924/commits/414e713be988b9e00a22842861317500fccabfb3

soliton-:
> option addedin filter with [414e713](https://github.com/wesnoth/wesnoth/commit/414e713be988b9e00a22842861317500fccabfb3)

I don't know what that is supposed to help. Whether some ability attribute has a default or not has nothing at all to do with the filter.

newfrenchy83:
> > option addedin filter with [414e713](https://github.com/wesnoth/wesnoth/commit/414e713be988b9e00a22842861317500fccabfb3)
> 
> I don't know what that is supposed to help. Whether some ability attribute has a default or not has nothing at all to do with the filter.

changed to https://github.com/wesnoth/wesnoth/pull/7924/commits/43c5a6a9a6e83927e535193e833b9568e1debc99

newfrenchy83:
> Can you add a unit test for this as well?

test added in https://github.com/wesnoth/wesnoth/pull/7924/commits/07561efdb57afb1d7bc9fbbf24ad8609a3e16468

soliton-:
> changed to [43c5a6a](https://github.com/wesnoth/wesnoth/commit/43c5a6a9a6e83927e535193e833b9568e1debc99)

That is still looking in the filter which makes no sense.

There are specific ability types. They have attributes and these attributes have defaults or not. That is not a variable thing.

As I said the caller should decide if there is a default and what it is. With that I mean the `def` parameter that some of those functions already have but for some reason not all of them.

newfrenchy83:
> > changed to [43c5a6a](https://github.com/wesnoth/wesnoth/commit/43c5a6a9a6e83927e535193e833b9568e1debc99)
> 
> That is still looking in the filter which makes no sense.
> 
> There are specific ability types. They have attributes and these attributes have defaults or not. That is not a variable thing.
> 
> As I said the caller should decide if there is a default and what it is. With that I mean the `def` parameter that some of those functions already have but for some reason not all of them.

where it must be specified then!?

soliton-:
The caller of a function is the part of the code that calls that function. As far as I can see all the relevant functions are called here:

https://github.com/wesnoth/wesnoth/blob/e84397256e27e160d4d4ec09539afbac466a27d8/src/units/unit.cpp#L1446-L1505

This is where the default is already specified for the bool and string functions. Presumably wrongly in the case of "affect_allies" for example which the wiki says has no default.

So there all calls should specify a default which includes a default of "no default". I.e. nullopt with optional or so.

newfrenchy83:
> The caller of a function is the part of the code that calls that function. As far as I can see all the relevant functions are called here:
> 
> https://github.com/wesnoth/wesnoth/blob/e84397256e27e160d4d4ec09539afbac466a27d8/src/units/unit.cpp#L1446-L1505
> 
> This is where the default is already specified for the bool and string functions. Presumably wrongly in the case of "affect_allies" for example which the wiki says has no default.
> 
> So there all calls should specify a default which includes a default of "no default". I.e. nullopt with optional or so.

i added a by_default variable for "value" checking because it is one attribute numerial who can have default value in drains berserk or other, add sub multiply and divide don't have default value.

soliton-:
> 'value' is the only numeric attribute that can have a default value in certain weapons and abilities (drains, berserk, heals), so adding a 'default' option and a default value when checking is only useful in this case accurate. for add sub multiply and divide, their lack of checkability simply means that they are not used

Hard to understand what you're saying.

Either an attribute has a default value or it does not. I don't know what you mean with "can have a default value".

Drains and berserk are weapon specials. Why are they relevant here? If heals really has a default for the value attribute then that needs to be documented and used properly just for the heals ability. Would make the fix more complex since currently the code just cares for attribute names not in what tag name they are used.

Can't tell what the last sentence means at all.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
```diff
@@ -296,3 +296,134 @@
...
+#####
+# API(s) being tested: [event][filter][filter_ability]default_value
+##
+# Actions:
+# Use the common setup from FILTER_ABILITY_TEST_WITHOUT_VALUE.
+# Add an event with a filter matching Alice's drains ability.
+# Alice attacks Bob and then Bob attacks Alice, as defined in FILTER_ABILITY_TEST.
+##
+# Expected end state:
+# The filtered event is triggered exactly once.
```
Pentarctagon, line 379: Needs `first_time_only=no` then.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
```diff
@@ -296,3 +296,134 @@
...
+                    x,y=13,4
+                [/destination]
+            [/attack]
+        [/do_command]
+        [end_turn][/end_turn]
+    [/event]
+#enddef
+
+#####
+# API(s) being tested: [event][filter][filter_ability]default_value
```
Pentarctagon, line 371: `default_value` is not an attribute, so it should not be added here.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
```diff
@@ -296,3 +296,134 @@
...
+# The filtered event is triggered exactly once.
+#####
+{GENERIC_UNIT_TEST event_test_filter_ability_without_value (
+    {FILTER_ABILITY_TEST_WITHOUT_VALUE}
+    [event]
+        name=attack
+        [filter]
+            [filter_ability]
+                tag_name=drains
+                value=10-45
```
Pentarctagon, line 388: Is this triggered when the attack drains 10-45 hp?

Pentarctagon:
I think I'm confused about what the issue is. If the ability's `value` attribute is not specified and it defaults to 0, it should be handled exactly the same as if the `value` attribute was explicitly set to be 0. That behavior seems correct to me.

newfrenchy83:
@Pentarctagon in abilities used as weapons like chance_to_hit or damage 'value' cannot have a fixed value by default, in these cases the absence of the 'value' attribute should not be considered as a default value as it would be the case for drains and trying to detect a value between 0 and 20 at an ability[damage]multiply=2 should result in failure, [damage] should not match because value is not used and its default value is not 0 but base_value, the same for drains whose default value is 25.
For 'add' and others, when the developer uses add=0-5 in the filter this means (for me) that not only must the ability have an 'add' attribute between 0 and 5 but that the use of the addition operation which is tested which is tested because unlike the other attributes the numerical value function is carried by 4 interchangeable attributes (fixed value or mathematical operation).

the check of affect_self is carried by a single attribute, active_on also but the mathematical operations are not defined by default the absence of 'add' should not mean add=0 but the addition function is not used.

Pentarctagon:
It sounds like the attributes being filtered on are incomplete then. Instead you should be filtering on `value=0-20` and `multiply=1`.

I disagree that it should ignore default values.

Pentarctagon:
After discussing this on Discord, how it should function is:
* if `value` is present by itself and it matches a filter on `value`, then it matches.
* if `value` is present along with add/sub/etc and it matches a filter on `value`, then it matches.
* if `value` is not present but there are other add/sub/etc attributes and the filter matches the default `value`, then it shouldn't match.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
```diff
@@ -296,3 +296,143 @@
...
+            silent=yes
+            [effect]
+                apply_to=new_ability
+                [abilities]
+                    [drains]
+                        [filter]
+                            [filter_location]
+                                time_of_day=neutral
+                            [/filter_location]
+                        [/filter]
```
Pentarctagon, line 333: Is there a reason this filter needs to be part of these tests?

newfrenchy83:
> After discussing this on Discord, how it should function is:
> 
>     * if `value` is present by itself and it matches a filter on `value`, then it matches.
> 
>     * if `value` is present along with add/sub/etc and it matches a filter on `value`, then it matches.
> 
>     * if `value` is not present but there are other add/sub/etc attributes and the filter matches the default `value`, then it shouldn't match.

and if value is not present but weapon/abilitie return boolinstead of numerical?

newfrenchy83:
@Pentarctagon i would know where i must put checking of presnece of attribute, in unit.cpp or config_filter.cpp?

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
newfrenchy83: no i removed this

newfrenchy83:
@Pentarctagon i want add a default_value filter who matche when a ability with default value don't use 'value' 'add' or other numérical attribute, what think you of that?

Pentarctagon:
> and if value is not present but weapon/abilitie return boolinstead of numerical?

> @Pentarctagon i want add a default_value filter who matche when a ability with default value don't use 'value' 'add' or other numérical attribute, what think you of that?

This would be for the four things currently checked with `bool_matches_if_present`?

> @Pentarctagon i would know where i must put checking of presnece of attribute, in unit.cpp or config_filter.cpp?

I think it should go in `config_filter.cpp`, and some tests added to `tests/test_config_filters.cpp` for it.

newfrenchy83:
> I think it should go in `config_filter.cpp`, and some tests added to `tests/test_config_filters.cpp` for it.

what test by exemple?

newfrenchy83:
> > and if value is not present but weapon/abilitie return boolinstead of numerical?
> 
> > @Pentarctagon i want add a default_value filter who matche when a ability with default value don't use 'value' 'add' or other numérical attribute, what think you of that?
> 
> This would be for the four things currently checked with `bool_matches_if_present`?
> 
something like that
`if(!filter["using_default"].empty()){
			bool not_attribute = filter["value"].empty() && filter["add"].empty() && filter["sub"].empty() && filter["multiply"].empty() && filter["divide"].empty();
			bool using_default = filter["using_default"].to_bool(false);
			if((using_default && !not_attribute) || (using_default && !not_attribute))
				return false;
		}`

Pentarctagon:
> what test by exemple?

I mean that if you're going to add more functions to `config_filter.cpp`, then you should add tests for those functions to `tests/test_config_filters.cpp`.

> something like that

So when `default_value=yes` and there's none of `value`, `add`, `sub`, `multiply`, or `divide`, then it would match against the default value of the ability?

newfrenchy83:
> > what test by exemple?
> 
> I mean that if you're going to add more functions to `config_filter.cpp`, then you should add tests for those functions to `tests/test_config_filters.cpp`.
> 
> > something like that
> 
> So when `default_value=yes` and there's none of `value`, `add`, `sub`, `multiply`, or `divide`, then it would match against the default value of the ability?

forget this idea, it's an error review this PR like it is, please.

newfrenchy83:
> > what test by exemple?
> 
> I mean that if you're going to add more functions to `config_filter.cpp`, then you should add tests for those functions to `tests/test_config_filters.cpp`.
> 
added in https://github.com/wesnoth/wesnoth/pull/7924/commits/0f9c77bf6a50d7e310c25949248f53392b58ef1e commit

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
```diff
@@ -296,3 +296,137 @@
...
+                value=33
+            [/filter_ability]
+        [/filter]
+        {FAIL}
+    [/event]
+    [event]
+        name=turn 2
+        {SUCCEED}
+    [/event]
+)}
```
Pentarctagon, line 430: The difference between these two tests is just filtering on a range vs on a single value, correct?

src/utils/config_filters.cpp
```diff
@@ -55,6 +55,9 @@ bool utils::config_filters::int_matches_if_present(const config& filter, const c
 	if(filter[attribute].empty()) {
 		return true;
 	}
+	if(cfg[attribute].empty()) {
+		return false;
+	}
```
Pentarctagon, line 60: `unsigned_matches_if_present` needs to be either updated as well, or just removed since nothing uses it anyway.

Pentarctagon:
I think aside from those last two comments, I'm good with this.

src/utils/config_filters.cpp
newfrenchy83: > `unsigned_matches_if_present` needs to be either updated as well, or just removed since nothing uses it anyway.

function updated, but i prefer what it remain in case where someone would using this in future.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
newfrenchy83: second test removed.

soliton-:
> After discussing this on Discord, how it should function is:
> 
> * if `value` is present by itself and it matches a filter on `value`, then it matches.
> * if `value` is present along with add/sub/etc and it matches a filter on `value`, then it matches.
> * if `value` is not present but there are other add/sub/etc attributes and the filter matches the default `value`, then it shouldn't match.

As far as I can tell neither this nor anything of my first comment has been implemented.

newfrenchy83:
> > After discussing this on Discord, how it should function is:
> > 
> > * if `value` is present by itself and it matches a filter on `value`, then it matches.
> > * if `value` is present along with add/sub/etc and it matches a filter on `value`, then it matches.
> > * if `value` is not present but there are other add/sub/etc attributes and the filter matches the default `value`, then it shouldn't match.
> 
> As far as I can tell neither this nor anything of my first comment has been implemented.

when you said value present by itself, you speak of value by default like in drains?

newfrenchy83:
> > After discussing this on Discord, how it should function is:
> > 
> > * if `value` is present by itself and it matches a filter on `value`, then it matches.
> > * if `value` is present along with add/sub/etc and it matches a filter on `value`, then it matches.
> > * if `value` is not present but there are other add/sub/etc attributes and the filter matches the default `value`, then it shouldn't match.
> 
> As far as I can tell neither this nor anything of my first comment has been implemented.

does this https://github.com/wesnoth/wesnoth/pull/7924/commits/66563884d6f58650047651aab5484792c66d39e9 commit meet your expectations, @Pentarctagon ?

Pentarctagon:
@soliton- I'm not sure what you mean - there are tests passing which test that it doesn't match against the default anymore.

newfrenchy83:
> @soliton- I'm not sure what you mean - there are tests passing which test that it doesn't match against the default anymore. I'm also not sure what adding a default value as an argument to all the calls to the `*_matches_if_present` functions would accomplish since the tag name needs to be checked at some point regardless to know what the expected default is.

this comment is fore me or @soliton- ?

Pentarctagon:
The `@soliton-` means it's addressed to him :P

newfrenchy83:
> > After discussing this on Discord, how it should function is:
> > 
> > * if `value` is present by itself and it matches a filter on `value`, then it matches.
> > * if `value` is present along with add/sub/etc and it matches a filter on `value`, then it matches.
> > * if `value` is not present but there are other add/sub/etc attributes and the filter matches the default `value`, then it shouldn't match.

i think what this time all condition expected are matched.

soliton-:
> @soliton- I'm not sure what you mean - there are tests passing which test that it doesn't match against the default anymore. I'm also not sure what adding a default value as an argument to all the calls to the `*_matches_if_present` functions would accomplish since the tag name needs to be checked at some point regardless to know what the expected default is.

Sure, the caller would check the tag name.

Pentarctagon:
So you want `int_matches_if_present()` to use a single `std::optional<int>` argument instead of the two arguments currently being added(`bool no_default_value, int def`)?

soliton-:
> So you want `int_matches_if_present()` to use a single `std::optional<int>` argument instead of the two arguments currently being added(`bool no_default_value, int def`)?

Seems simpler but that's more of an implementation detail that's not that important.

soliton-:
On a quick look the current code looks much better.

src/units/unit.cpp
```diff
@@ -1482,8 +1482,26 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 	if(!string_matches_if_present(filter, cfg, "active_on", "both"))
 		return false;
 
-	if(!int_matches_if_present(filter, cfg, "value"))
-		return false;
+	if(!filter["value"].empty()){
+		bool has_other_key = (cfg["add"].empty() && cfg["sub"].empty() && cfg["multiply"].empty() && cfg["divide"].empty()) ? false : true;
```
soliton-, line 1486:
```suggestion
		bool has_other_key = !(cfg["add"].empty() && cfg["sub"].empty() && cfg["multiply"].empty() && cfg["divide"].empty());
```
IMO that is easier to read. Perhaps even clearer might be: `!cfg["add"].empty() || !cfg["sub"].empty() ||` ...

src/units/unit.cpp
```diff
@@ -1482,8 +1482,26 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
...
+		if(tag_name == "drains"){
+			if(!int_matches_if_present(filter, cfg, "value" , has_other_key, 25)){
```
soliton-, line 1488: The wiki says the default is 50.

newfrenchy83: fixed

src/units/unit.cpp

Pentarctagon:
@soliton- any other issues that should be addressed before merging?

Pentarctagon:
Well, I'll merge this tomorrow if nothing else is said.

soliton-:
As before all functions should take a default parameter. Further I see no change to "affect_allies" which is documented as using no default.

Also it still makes no sense to me why `empty()` is used instead of `blank()`.

newfrenchy83:
> As before all functions should take a default parameter. Further I see no change to "affect_allies" which is documented as using no default.
> 
> Also it still makes no sense to me why `empty()` is used instead of `blank()`.

except value in some case, all numérical function son't have default parameter, because it is diffrent case of boolean function(in abilities and weapon specials value,add,tec.. are facultative because only part of them return numerical value).

For empty used instead of blank we must checked if attribute exist or not and must be done with empty() then what blank() is diffrent, and in atta_typecpp, is empty() who is used for that.

Pentarctagon:
> Further I see no change to "affect_allies" which is documented as using no default.

I think the documentation should be updated, rather than changing on behavior that people might be relying on now.

> As before all functions should take a default parameter.

To be more specific - do you mean that `int_matches_if_present_or_negative()` and `double_matches_if_present()` should take 0 and 1 respectively as an argument to the method rather than hard coding the value?

newfrenchy83:
Further I see no change to "affect_allies" which is documented as using no default.
> 
 
fixed in https://github.com/wesnoth/wesnoth/pull/7924/commits/620db1590e14ed351f7882189dbc32d424ed912b

Pentarctagon:
> Since int_matches_if_present does not only take 0 as default I don't see why int_matches_if_present_or_negative would.
> 
> I have not checked if there are different defaults for double but even if not now then perhaps later. IMO only the caller can decide the default. The helper functions have no context to know one way or another.

So you want lines 1514-1524 changed to be:
```
	if(!int_matches_if_present_or_negative(filter, cfg, "add", "sub", true, 0))
		return false;
	if(!int_matches_if_present_or_negative(filter, cfg, "sub", "add", true, 0))
		return false;
	if(!double_matches_if_present(filter, cfg, "multiply", true, 1))
		return false;
	if(!double_matches_if_present(filter, cfg, "divide", true, 1))
		return false;
```
?

newfrenchy83:
> > Since int_matches_if_present does not only take 0 as default I don't see why int_matches_if_present_or_negative would.
> > I have not checked if there are different defaults for double but even if not now then perhaps later. IMO only the caller can decide the default. The helper functions have no context to know one way or another.
> 
> So you want lines 1514-1524 changed to be:
> 
> ```
> 	if(!int_matches_if_present_or_negative(filter, cfg, "add", "sub", true, 0))
> 		return false;
> 	if(!int_matches_if_present_or_negative(filter, cfg, "sub", "add", true, 0))
> 		return false;
> 	if(!double_matches_if_present(filter, cfg, "multiply", true, 1))
> 		return false;
> 	if(!double_matches_if_present(filter, cfg, "divide", true, 1))
> 		return false;
> ```
> 
> ?
> 
> It's also maybe worth noting that the functions in `config_filters` are only actually used here in the `matches_ability_filter()` method of `unit.cpp`. My assumption is that they were put into a separate file so that their boost tests could be implemented more easily/cleanly rather than them being intended as general use functions.

what must i doing, then?

Pentarctagon:
@newfrenchy83 I don't know. Please don't do anything yet.

Pentarctagon:
@newfrenchy83 please change the functions `int_matches_if_present_or_negative()` and `double_matches_if_present()` to match how the `int_matches_if_present()` function works in terms of the `bool no_default_value` and `int def` arguments that this PR is adding.

newfrenchy83:
>>
>@newfrenchy83 please change the functions int_matches_if_present_or_negative() and double_matches_if_present() to match >how the int_matches_if_present() function works in terms of the bool no_default_value and int def arguments that this PR is >adding.

changed with https://github.com/wesnoth/wesnoth/pull/7924/commits

data/schema/game_config.cfg
```diff
@@ -44,6 +44,10 @@
         name="value_type"
         value="empty|value|add|sub|multiply|divide"
     [/type]
+    [type]
+        name="affect_allies_filter"
+        value="empty|yes|no"
```
soliton-, line 49: This allows a literal "empty" value.
```suggestion
        value="|true|false|yes|no"
```

soliton-: Ugh, you actually added support for a literal empty value... Why did you do that?

src/units/unit.cpp
```diff
@@ -1457,8 +1457,13 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 	if(!bool_matches_if_present(filter, cfg, "affect_self", true))
 		return false;
 
-	if(!bool_matches_if_present(filter, cfg, "affect_allies", true))
+	if(!filter["affect_allies"].empty() && !cfg["affect_allies"].empty()){
+		if(!bool_matches_if_present(filter, cfg, "affect_allies", true)){
+			return false;
+		}
+	} else if(!filter["affect_allies"].empty() && filter["affect_allies"].str() != "empty"){
```
soliton-, line 1464: This is very confusing. This should check if affect_allies exists in the filter at all. Then it should check if it is empty and return the comparison to what's in cfg or if it's not empty do the check as before.

src/units/unit.cpp
```diff
@@ -1457,8 +1457,13 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
...
 		return false;
+	}
```
soliton-, line 1466: This is very confusing. This should check if affect_allies exists in the filter at all. Then it should check if it is empty and return the comparison to what's in cfg or if it's not empty do the check as before.

newfrenchy83: Because when affect_allies  no specified in abilites only unit adjacent of same side affected but nt unit of allied side and if affect_allies =no none unit allies, or same side are affected affect_allies has 3 vale,true,false and empty because absence of attributechas value différente what if affect_allies =no/yes

newfrenchy83: Wiki is correct for affect_allies  not have default_value

data/schema/game_config.cfg
newfrenchy83: Because empty is an value specific to affect_allies, by default when affect adjacent, abilite affect only same side

src/units/unit.cpp
soliton-: > Because when affect_allies no specified in abilites only unit adjacent of same side affected but nt unit of allied side and if affect_allies =no none unit allies, or same side are affected affect_allies has 3 vale,true,false and empty because absence of attributechas value différente what if affect_allies =no/yes

Ok. Does any of that change how to properly implement the check?

data/schema/game_config.cfg
soliton-: Sure but that is the empty string not the literal string `empty`...

newfrenchy83: what suggest instead?

src/units/unit.cpp
newfrenchy83: like affect_allies no have default value check if abilitie has this attribute is also important what value of attribute dand filter["filter_allies"]=empty is for check if abilitie using attribute or not.

data/schema/game_config.cfg
soliton-: I made a literal suggestion in the first comment. You can click on commit if you like.

src/units/unit.cpp
soliton-: This is getting surreal. Do you understand the difference between an empty string with no characters and the string `empty` with 5 characters?

newfrenchy83: then suggest another string instead of empty and i change.

newfrenchy83: "empty" changed to "no_used"

data/schema/game_config.cfg
newfrenchy83: i mademy own changes

src/units/unit.cpp
soliton-: wtf

newfrenchy83: > wtf

excuse me?

Pentarctagon: As per the [wiki](https://wiki.wesnoth.org/AbilitiesWML#Common_keys_and_tags_for_every_ability), there are three valid values: yes, no, and not present.

The literal string value of `"empty"` or `"no_used"` is not valid anywhere.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
CelticMinstrel: Huh? That doesn't seem correct if the intention is to assert that it's triggered exactly once…? Or am I misunderstanding something?

src/units/unit.cpp
Pentarctagon: Please change it back to "empty".

CelticMinstrel: I also don't understand what @soliton- and @Pentarctagon are complaining about here; using the string "empty" for this purpose seems perfectly natural to me, or perhaps "default" would work instead. Admittedly, using the empty string for that purpose isn't terrible, I suppose. It's not like this string is going to be matched directly against the value of `affect_allies` in the abilitiy; it's an arbitrary string that means the `affect_allies` attribute **must not be present** in the ability. So whether that string is `"empty"` or `"default"` or `"no_used"` or `""` doesn't make that much of a difference.

data/schema/game_config.cfg
CelticMinstrel: I think this should be more like:

```
[type]
    name="affect_allies_filter"
    [list]
        [element]
            value="true|false|yes|no|empty" # replace empty with whatever we end up settling on
        [/element]
    [/list]
[/type]
```

And then in the code, handle it as a comma-separated list. So you could ask for:

* `affect_allies=yes` – abilities that affect allies and own units
* `affect_allies=empty,no` – any abilities that don't affect allies (but may affect own units)
* `affect_allies=yes,empty` – any ability that affects own units (and possibly allies too)
* etc

Though since there are only a few combos, I suppose there's also the possibility of giving each combo a unique name… but that might just be confusing.

src/units/unit.cpp
soliton-: It makes no sense to use a different value then the ability uses. That is just confusing. The ability uses `affect_allies=` or `affect_allies=""` if that is more understandable. There is zero reason to not simply do the same in the filter.

data/schema/game_config.cfg
soliton-: The list part is correct. I have no idea what the issue is with using an empty value to represent the exact same in the ability.

src/units/unit.cpp
soliton-: Ok, I misunderstood that the ability code uses the empty value instead of not specifying the attribute but the wiki clearly says "not used". My bad. If someone can point out the exact code where this is handled I would appreciate it.
https://github.com/wesnoth/wesnoth/blob/000ba55eb1c5f0261132aa734aaf4b05bb0fdf21/src/units/abilities.cpp#L1037-L1038
This handles `affect_allies` but does not treat not set special as far as I can tell.

So we do need a special value in the filter since we do need to distinguish between the filter just not filtering for affect_allies and filtering for affect_allies to not be there in the ability. "empty" or "" would be confusing for that so "not set" exactly as the wiki puts it or "not used" would be options.

newfrenchy83: perhaps what imust added "smae_side_only" in code for could used like value by default?

newfrenchy83: > Please change it back to "empty".

done

data/schema/game_config.cfg
newfrenchy83: it's done, but i continue to use 'empty' value

src/utils/config_filters.cpp
```diff
@@ -46,40 +46,72 @@ bool utils::config_filters::unsigned_matches_if_present(const config& filter, co
...
 
-	return in_ranges<double>(cfg[attribute].to_double(1), utils::parse_ranges_real(filter[attribute].str()));
+	return in_ranges<double>(cfg[attribute].to_double(def), utils::parse_ranges_real(filter[attribute].str()));
+}
+
+bool utils::config_filters::bool_or_empty(const config& filter, const config& cfg, const std::string& attribute)
+{
+	if(filter[attribute].empty()) {
```
CelticMinstrel, line 101: Should this be `blank()`?

src/utils/config_filters.cpp
```diff
@@ -46,40 +46,72 @@ bool utils::config_filters::unsigned_matches_if_present(const config& filter, co
...
+	if(cfg.has_attribute(attribute)){
+		if(!cfg[attribute].to_bool()){
+			cfg_bool_value = "no";
+		} else if(cfg[attribute].to_bool()){
+			cfg_bool_value = "yes";
+		}
+	} else {
+		cfg_bool_value = "empty";
+	}
+	return ( std::find(filter_attribute.begin(), filter_attribute.end(), cfg_bool_value) != filter_attribute.end() );
```
CelticMinstrel, line 116: This doesn't seem to support the yes/true and no/false equivalency that is normal to WML booleans.

src/utils/config_filters.cpp
newfrenchy83: i don't know, what diffrence between empty and blank?

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
stevecotton: It's needed to check that the filter works. With `first_time_only=yes` then it'll be triggered at most once even if the filter passes multiple times.

newfrenchy83: @stevecotton first_time_only is never specified in the tests earlier in this file and that wasn't a problem, right? I copied them while only changing the interest parameters.

src/utils/config_filters.cpp
Pentarctagon: `blank()` checks if the attribute exists. `empty()` checks if the attribute exists and if it does then it also checks whether its value is an empty string.

newfrenchy83: i changed code for what if (filter[attribute.empty() and !cfg.has_attribute(attribute) then return true

newfrenchy83:
@Pentarctagon it should be good now.

src/utils/config_filters.cpp
Pentarctagon: I don't know why that would be necessary? The point was that `filter[attribute].empty()` should be `filter[attribute].blank()`.

newfrenchy83: it's done

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
Pentarctagon: I didn't look at any earlier tests, but I think that in general you should not use `first_time_only=yes` to ensure something in a test only happens once. It should use `first_time_only=no` and check that it occurs the expected number of times.

src/utils/config_filters.cpp
Pentarctagon: I suggest using `utils::split_set()` and then using its `count()` method to check whether it contains yes/no/true/false/empty.

src/utils/config_filters.cpp
Pentarctagon: @CelticMinstrel did you mean just here specifically, or throughout the config_filters.cpp file?

src/utils/config_filters.cpp
newfrenchy83: it's done.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
```diff
@@ -296,3 +303,140 @@
 )}
 
 #undef FILTER_ABILITY_TEST
+
+##
+# Common code for the tests in this file.
+##
+# Actions:
+# Give both Alice and Bob 100% chance to hit.
+# Give Alice a drains ability which is only active during liminal times of day and to value 25(defined in hardcoding by default).
```
stevecotton, line 312: This line's documentation about the value is wrong.

src/utils/config_filters.cpp
```diff
@@ -46,40 +46,75 @@ bool utils::config_filters::unsigned_matches_if_present(const config& filter, co
...
-	const config& filter, const config& cfg, const std::string& attribute, const std::string& opposite)
+	const config& filter, const config& cfg, const std::string& attribute, const std::string& opposite, bool no_default_value, int def)
 {
-	if(int_matches_if_present(filter, cfg, attribute)) {
+	if(int_matches_if_present(filter, cfg, attribute, no_default_value , def)) {
 		return true;
 	}
 
 	// don't check for !cfg[opposite].empty(), as this assumes a default value of zero (similarly to
 	// int_matches_if_present)
```
stevecotton, line 76: This documentation is outdated. However, it should be replaced by a description of the new logic, not just removed.

src/utils/config_filters.hpp
```diff
@@ -56,9 +56,11 @@ bool unsigned_matches_if_present(const config& filter, const config& cfg, const
  * The function is named "negative" in case we later want to add a "reciprocal" for the "multiply"/"divide" pair.
  */
 bool int_matches_if_present_or_negative(
-	const config& filter, const config& cfg, const std::string& attribute, const std::string& opposite);
+	const config& filter, const config& cfg, const std::string& attribute, const std::string& opposite, bool no_default_value = true, int def = 0);
```
stevecotton, line 59: `std::optional<int>` would encapsulate the logic better than `bool, int`.

stevecotton:
> It's also maybe worth noting that the functions in `config_filters` are only actually used here in the `matches_ability_filter()` method of `unit.cpp`. My assumption is that they were put into a separate file so that their boost tests could be implemented more easily/cleanly rather than them being intended as general use functions.

I was intending to use them more generally, to tidy up other filter code. However, I'm happy for the functions to become less general if it makes the 1.18 codebase easier to read.

src/utils/config_filters.hpp
stevecotton: Ah, that's already been mentioned. Resolving as it's just an internal detail.

src/utils/config_filters.cpp
newfrenchy83: updated documentation

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
newfrenchy83: fixed

stevecotton: It's also not locked to liminal times of day.

src/utils/config_filters.cpp
newfrenchy83: fixed

data/test/scenarios/wml_tests/ScenarioWML/EventWML/events-test_filter_ability.cfg
newfrenchy83: fixed

src/utils/config_filters.cpp
stevecotton: Looking further at this, I think the default for this function can only be zero. The purpose of the `..._or_negative` function is to handle two opposites that are symmetrical around zero, that `add=-2` is equivalent to `sub=2`.

newfrenchy83: but if i put a default value, then if i check with add=-1-3 an ability with value=15, then with add=0 by default it match ability then what it should'nt.

newfrenchy83:
> I think besides `int_matches_if_present_or_negative` that makes sense. That one is pretty specific and not sure if there are a lot of other places with this sub/add setup. It's also a bit questionable if this logic shouldn't be rather handled by the caller/user. Otherwise it might be surprising why multiply/divide for example are not handled similarly.
> 
> Those functions should all take a default though and use `blank()` (or `has_attribute()`) instead of `empty()`.

ok for attribute but not for default because each check of attribute who have default value match with ability who using other attribute.

newfrenchy83:
@Pentarctagon i think what it is good, it's time.

Pentarctagon:
There are a few review comments still open from celticminstrel and stevecotton - once those are marked resolved I'll click the merge button.

newfrenchy83:
> There are a few review comments still open from celticminstrel and stevecotton - once those are marked resolved I'll click the merge button.

I made the repairs and corrections requested in these comments (except the default value of @soliton which would amount to closing this PR) and closed the conversations myself, are you ready to merge this PR if no other comments is placed?

soliton-:
The new `bool_or_empty` function has no documentation.

After all this confusion "empty" is definitely not the term to use to mean the attribute is not set/used.

At the end of this it should be clear what the documentation of `[filter_ability]` is and the wiki should be updated with that.

newfrenchy83:
> The new `bool_or_empty` function has no documentation.
> 
> After all this confusion "empty" is definitely not the term to use to mean the attribute is not set/used.
> 
> At the end of this it should be clear what the documentation of `[filter_ability]` is and the wiki should be updated with that.

what term would you what i put then?

CelticMinstrel:
Maybe "none" would work? Because you're asking to filter for abilities with no value specified.

src/utils/config_filters.cpp
CelticMinstrel: I'm not sure. I only looked at this one place. If 'empty()` is being used to check if an attribute exists though, then that's wrong. It should either use `blank()` or `has_attribute()`.

newfrenchy83: ialready replace empty() by has_attribute except in unsigned_matches

newfrenchy83: string_matches_if_present and bool_matches_if_present are only function with empty(), should replace wit has_attribute in these functions or not, i think what empty() must be used for bool because empty attribute means what filter check nor true nor false, and for string, i don't know.

newfrenchy83: that string can take an empty value, ok, because we could check the attributes of this type which are empty, but for bool it is better to continue using empty() because I don't know if we should expect that an unspecified filter is false or true by default and to avoid confusion, the check can only be performed if the value is specified.

newfrenchy83: it's done, onlu bool_if_present continue to use empty() instead of bank() or has_attribute()

newfrenchy83:
> Maybe "none" would work? Because you're asking to filter for abilities with no value specified.

"empty"replaced by "none".

Pentarctagon:
Will merge before tagging tomorrow, barring anything else coming up.

newfrenchy83:
Thancks, can i expert same for https://github.com/wesnoth/wesnoth/pull/7865 or must i accept what was not commited for 1.18?

Pentarctagon:
https://github.com/wesnoth/wesnoth/pull/7865 will be merged for 1.17.23. Some other devs want to make some changes for [filter_ability]/[filter_ability_active] for that release anyway, so might as well try to fit that in too.

newfrenchy83:
> #7865 will be merged for 1.17.23. Some other devs want to make some changes for [filter_ability]/[filter_ability_active] for that release anyway, so might as well try to fit that in too.

If it's about using a single tag [filter_bility] for both [filter_ability] and [filter_ability_active] as suggested in https://github.com/wesnoth/wesnoth/issues/7926 I don't think it's desirable to do this because it involves creating an attribute to determine whether it is necessary to check the abilities encoded in the unit or whether it is necessary to check the influence of an ability on this unit. This will be complicated to implement and will do nothing functionally.

If there are other suggestions on this subject, I can possibly work on them.

Pentarctagon:
I haven't been keeping track of what's going on with that, so I don't know.

PR #7930, opened 2023-10-01
1.17 - Windows (Xo* embellishment)

doofus-01:
This is an embellishment overlay for `Xo*`-style walls.  Currently only:
 - `Xos` barred gap with time-of-day-variants (though day version needs work)
 - `Xoc` stained glass window (also not great, but works for now)

Despite the raw state, I think all the "API changes" needed are here, so it should be OK for feature freeze.  The images can be touched-up and other variations can be added later.
Screenshot_20230930_


doofus-01:
I'll plan to merge this tomorrow, with fixed copyright CSV, unless someone sees a reason this shouldn't be in BfW 1.18.

CelticMinstrel:
In the presence of `[time_area]`, I suppose this reflects the time of day on the wall hex? I suppose that's probably good enough since walls are impassable… if it wasn't too difficult it might make sense to read the time of day on the hex _behind_ the wall hex, but I bet that _is_ a bit tricky and maybe not worth it.

doofus-01:
I think it's up to the scenario author, using `[time_area]` to include the appropriate hexes.  There might be some cleanup needed in tints bleeding through the wall, but I don't think that will require an API change.

Issue #7201, opened 2022-12-07
Inspect window can crash on HDPI displays (1.17 only)


stevecotton:
### Game and System Information

- **Version:** 1.17.10+dev
- **Downloaded from:** Built from Git
- **Build info:** GCC and b5ff5c3ddc706f1281bbbe2f235b16c4b00f92db
- **OS:** Linux

### Description of the bug

The inspect window paginates data so a maximum of 20000 characters appear in the right-hand window. On an HDPI display, 1.17's large fonts can make the 20000 characters larger than Pango's limits. An example of circa 14000 characters is already enough to need 2740x33009 pixels on my 3840x2160 screen, which allows the surface to be created in  pango_text::create_surface but then gets CAIRO_STATUS_INVALID_SIZE in pango_text::render.

Similar to #6952, there's no longer viewport code in canvas.cpp limiting the area to the drawn.

In #7194, I reduced the character limit while testing to trigger pagination, but didn't test with the full 20000 character limit.`

### Steps to reproduce the behavior

1. Start Under the Burning Suns, because the two starting units have lots of AMLA data stored in their `[unit]`
2. `:debug`
3. `:inspect`
4. Scroll down to the "units" section (this takes some time, the scrolling is slow and there are 1800 entries in the `outer villages` array)
5. Click on either unit (there are only two at the moment, Kaleh and Nym)
6. The game exits, `Caught general 'St12length_error' exception: Text is too long to render`.

stevecotton:
That `outer_villages` array's size is #7202.

Wedge009:
Admittedly I'm not on what would be considered a 'hidpi' monitor (2560x1440) but I didn't get a crash. Instead, I sometimes get a display, sometimes not, but I see the following in my stderr:
`20221208 08:53:37 error SDL: When creating texture from surface: Texture dimensions are limited to 16384x16384`

stevecotton:
I think the same root cause has two different error messages, depending on the font size. If the height is 2^15 or more pixels then it hits the `Text is too long to render` sanity check, if it's between 2^14 and 2^15 then it passes that and runs into the `Texture dimensions are limited` check.

However, it's not worth making the `too long to render` check have the same behavior as the `texture dimensions are limited` check, IMO. Instead we should either fix canvas.cpp to not request such huge viewports, or we could work around it by limiting the amount of text per page to a smaller length.

stevecotton:
[...]

PR #7931, opened 2023-10-01
Credits Crash Bug fix

irregularBismuth:
## Fix Issue ##
#6952 
## Solution ##
Implements sliding window by parsing the markup of the string and splitting it into smaller chunks and combining three chunks at a time. By doing this the users will see the credits without it crashing. 

The current solution introduces some magic numbers for how long to move upward in the scroll bar this is not a perfect solution and will need more work as there are some visual glitches!

## New Delimitation ##
As users scroll further, content from the first chunk is replaced by that from the fourth chunk and so on. Consequently, once users move past the initial segment, they cannot revisit it without restarting the credits, introducing a limitation in the scrolling functionality.


CelticMinstrel:
Looks mostly okay, but you need to fix the indentation in `get_background_images` and the `credits_group` constructor, at minimum. There may be more places as well. Just look at the diff where the indentation was changed.

Pentarctagon:
Thanks for looking into this!

Does this work when using the scaling option in the preferences?

irregularBismuth:
Pixel scale multiplier (Preferences -> display -> pixel scale mult) doesn't seem to change anything if that's the scaling option you mean? 

src/about.cpp and src/font/text.cpp shouldn't differ anything from main branch so will be removing them from this branch

src/gui/dialogs/end_credits.cpp
```diff
@@ -90,14 +90,69 @@ void end_credits::pre_show(window& window)
...
+	
+	content_ = focus_ss.str().empty() ? ss.str() : focus_ss.str();
+
+	//TODO: Fix this better ? 
+	//Don't really know how to deal with this the best way but 
+	//lambda fucntion to parse the markup language ensure when storing the 
+	//substrings it doesn't cut in between spans and thus making markup invalid
+	//when combinding multiple substrings for the sliding window
+	//IrregularBismuth 2023-09-30
+	auto populateChunks = [this](const size_t& maxChunkSize) -> void {
```
Pentarctagon, line 99: Wesnoth uses snake case rather than camel case for variable names.

Pentarctagon:
As a side question - could something like this also be applicable to #7201, since that's a similar issue with the inspect dialog.

irregularBismuth:
The same error code of  'St12length_error' exception: Text is too long to render' was given for the credits #6952 so something similar would probably work for #7201 as long as there is something you can check when you need to swap the data for the sliding window to work  since in the credits you can utilize the cur_pos of the scroll bar.

Didn't get it to build when removing the about.cpp and font/text.cpp so added them again but copied them from the main branch

Pentarctagon:
Alright, cool. Also the indentation can be fixed with https://github.com/wesnoth/wesnoth/blob/master/utils/CI/fix_whitespace.sh (if you're using macOS or Linux), or I or another dev should be able to run it and push to this PR.

CelticMinstrel:
> As a side question - could something like this also be applicable to #7201, since that's a similar issue with the inspect dialog?

Though something similar would certainly work, I think for inspect it would be more important to make it reversible so you can also scroll back to the top.

> Also the indentation can be fixed with https://github.com/wesnoth/wesnoth/blob/master/utils/CI/fix_whitespace.sh (if you're using macOS or Linux)

Not entirely true. That would not have fixed the indentation issue I saw in here, which was _missing_ indentation, not the wrong type of indentation. Some lines had mysteriously been outdented a level.

src/gui/dialogs/end_credits.cpp
```diff
@@ -51,53 +52,97 @@ void end_credits::pre_show(window& window)
...
+				size_t tag_open_pos = content_.rfind("<span", end_pos);
+				if (tag_open_pos != std::string::npos && tag_open_pos > newline_pos) {
+					tag_close_pos = content_.find("</span>", tag_open_pos);
+					if (tag_close_pos != std::string::npos) {
+						end_pos = tag_close_pos + 7;  // +7 to include the closing tag
+					}
+				}
+			}
+		content_substrings_.push_back(content_.substr(start_pos, end_pos - start_pos));
+		//	std::cout << contentSubstrings_.size() << " size of vector" << std::endl;
```
Pentarctagon, line 128: I assume this can be deleted?

src/gui/dialogs/end_credits.hpp
```diff
@@ -47,16 +47,28 @@ class end_credits : public modal_dialog
...
 
 	// The speed of auto-scrolling, specified as px/s
 	int scroll_speed_;
 
 	uint32_t last_scroll_;
+	//Fix credits crash 
+	static constexpr size_t sliding_size_=2; // slidingSize alters how many of the sliding contents are to be run at once 
+						  // n-1 = 2 => 3 strings at once concatinated 
+	static constexpr size_t screen_space_ = 300; // magic number #1  
+	static constexpr size_t magic_number_ = 150; // magic number #2
```
Pentarctagon, line 64: This definitely needs a better name.

Also generally, it would be better for these variable comments to be in doxygen format.

src/gui/dialogs/end_credits.cpp
irregularBismuth: Yeah the print is not needed

Pentarctagon:
@irregularBismuth are you going to be coming back to this?

Pentarctagon:
Cleaned up the formatting and a bit of the code. It jerks periodically while scrolling, but otherwise it seems to work.

CelticMinstrel:
There are still a bunch of blank lines that were deleted for no apparent reason.

Pentarctagon:
Should be good now.

Pentarctagon:
I'm going to see if I can figure out what's causing the credits to jerk when switching to the next chunk of text, and I plan to merge this in a few days regardless - it's an improvement over the current "make wesnoth crash" behavior.

Wedge009:
I just gave this a try - definitely better than crashing. From what I (re-)read in #6952, this issue was a necessary result of the graphics overhaul where the text cache can't handle the huge credits texture.

With regards to the new limitation that one cannot scroll the credits once a segment has dropped out, I don't recall being able to reverse the direction of credits and I can't seem to do it now anyway. So it seems like the biggest issue is the transition between segments.

...although looking at the results more closely, I only seem to get through two segments before the credits just freeze. The bottommost line is consistently Jason Frailey (Valdroni).
https://github.com/wesnoth/wesnoth/blob/50a21a2c6c648b4ebcc2ab9fbfaed9b03f5d7f0f/data/core/about.cfg#L681

CelticMinstrel:
> With regards to the new limitation that one cannot scroll the credits once a segment has dropped out, I don't recall being able to reverse the direction of credits and I can't seem to do it now anyway. So it seems like the biggest issue is the transition between segments.

If I recall correctly, the up arrow was supposed to reverse the direction, and the down arrow was supposed to increase the speed. I think the scroll wheel was also supposed to work. These aren't crucial features, however.

Wedge009:
Up key only slowed down the scroll speed. I did notice some scroll up behaviour with mouse wheel in 1.16 but it wasn't consistent.

Anyway, I suppose even a frozen credits is better than a complete crash.

Pentarctagon:
@Wedge009 I simplified its handling a bit - let me know if it helps at all. It will still crash once you get to the end of the credits, but it sounds like that'd be an improvement from what you're seeing currently.

Wedge009:
Unfortunately I don't have time to dig deeper, but with the current state of the PR, it looks like it's freezing after the first segment now instead of the second. Last line seen is now Ben Wenzel (artisticdude). My window is 3840x2076 (UHD less window decorations and shell task bar).

I haven't ever got to the end of the credits to experience any crashes. I just escape out of the credits and the game continues to run normally.

Wedge009:
I noticed an oddity that I can't find in code, so I'm wondering if this PR is changing the text somewhat when splitting the textures into segments.

image

There's no line-break between the Estonian translation heading and the first translator. From the code, it looks like the syntax is correct.

https://github.com/wesnoth/wesnoth/blob/3f95df8fdc0310130b4aa61841f0e48ea83a32c3/data/core/about_i18n.cfg#L480-L517

Pentarctagon:
Pushed a fix for the text issue. I can't reproduce any other crashes besides the one at the very end though.

Wedge009:
I'm trying to work out what the end of the credits is supposed to be. In 1.16 I get to what looks like the end of Winds of Fate credits and then it stops. Doesn't go back to the main menu automatically as I thought it would - hence the Close button, I suppose.

So in master I get the same behaviour (ignoring the bumpy transitions between segments), albeit only at QHD window size. It's only when I run the credits a second time that I get the segmentation fault.

At UHD (maximised window not borderless full screen), I still have it freezing at what seems to be the first segment. I'm not sure if there's any cause with 'magic numbers' - as best as I can tell that's in relation to line length, not segment size.

Aside from the bumpy transitions between segments, overall the animation looks much smoother with the hardware rendering than it does in 1.16 with software.

Pentarctagon:
If you can provide a stack trace of the crash that'd be helpful. Otherwise I can't reproduce any crashes locally anymore.

Wedge009:
I didn't have any crashes after that last change. I was trying to work out why the animation stops on certain window sizes. I suspect for a related reason to the bumpy transition between segments: the positioning seems to be based on a timer and that timer may not necessarily be an accurate representation of the actual positioning - so when it comes time to switch to the next segment it may be slightly before or after the positioning of the previous segment.

Pentarctagon:
I had hoped that changing it to split the text into chunks based on the number of lines instead of the number of characters would help with the jerkiness since all chunks would be the same height, but it didn't seem to help.

Currently it uses `SDL_GetTicks()` which should be fine, but maybe using something from `std::chrono` with better precision would help.

Pentarctagon:
So actually if I have it set the scrollbar position to 75 (I don't know why 75 is the magic number) after updating the text as well as remove the `span`s from the headers, then it scrolls almost completely smoothly. So probably the issue is that the actual text height of each chunk isn't identical.

Also it doesn't work well at all with the pixel scale multiplier, though I think that might be at least in part because on my screen it can't fit all 3 chunks of text on the UI at the same time.

Wedge009:
I still have some slight jitter between segments but it is better than before. If it didn't freeze when my window size is UHD, I'd say it's almost ready to merge. (It also freezes at the penultimate segment - I think - in the middle of the campaign credits for Under the Burning Suns, before Winds of Fate.)

src/gui/dialogs/end_credits.cpp
```diff
@@ -44,13 +44,15 @@ end_credits::end_credits(const std::string& campaign)
...
+	, first_idx_(0)
+	, last_idx_(first_idx_ + sliding_size_)
 {
 }
 
 void end_credits::pre_show(window& window)
 {
 	// Delay a little before beginning the scrolling
-	last_scroll_ = SDL_GetTicks() + 3000;
+	last_scroll_ = SDL_GetTicks();
```
Wedge009, line 55: If the delay is removed now, should remove the comment explaining that delay as well.

Pentarctagon: True.

mesilliac:
> From what I (re-)read in https://github.com/wesnoth/wesnoth/issues/6952, this issue was a necessary result of the hardware rendering changes

It was absolutely not necessary, and frankly i'm still pissed about it.

Anyway it's good that this is finally getting worked on, even if it's not the original culprit who is fixing it.

PR #7933, opened 2023-10-01
update_copyrights - improve performance a lot

Pentarctagon:


Wedge009:
I haven't looked at this in detail, but wondering what the performance change was. Is it that you're comparing hashes now instead of meta-data?

ProditorMagnus:
What is the reason for looking file size?

Pentarctagon:
@Wedge009 yeah, apparently looking up the most recent commit information via `git log` is incredibly slow.

@ProditorMagnus checking file size is even faster than getting the hash.

cooljeanius (approved):

ProditorMagnus:
>             size = str(os.path.getsize(file))
>             hash = do_hash(file)
You get hash no matter what size says.

Pentarctagon:
True.

cooljeanius:
> > ```
> >         size = str(os.path.getsize(file))
> >         hash = do_hash(file)
> > ```
> 
> You get hash no matter what size says. But size collision feels so rare case that I did not feel need to point that out yesterday to separately check them.
> size=, hash= -> often
> size!=, hash!= -> happens
> size=, hash!= -> very rare, no need to optimize for it
> size!=, hash= -> extremely rare

So, I started using the `./update_copyrights` script from this PR with the "Size" field before this feedback led to that field being removed, so now my `copyrights.csv` file has problems with it leading to cooljeanius/wesnoth#104... how do I fix it?

Pentarctagon:
What are you using it for?

ProditorMagnus:
Replace it with current version.

cooljeanius:
> Replace it with current version.

I did, but that's what caused the failure. It was working fine when I was using the version that included the "Size" field.

Pentarctagon:
You need to recreate it from scratch.

PR #7934, opened 2023-10-02
1.17 - Jumping Spider

doofus-01:
This is a last minute addition to the Spider roster, I think it would be nice to have some variety there.  It would have been better to include this earlier, but we are where we are.

The images were thrown together somewhat quickly, will need updates, but they should be good enough for the next release?
Screenshot_20231001_


Pentarctagon:
The left and right look horrifying.

knyghtmare:
Shouldn't this unit get like better forest defense?

Hejnewar:
That triggers arachnophobia hard and I don't even have it. Kill on sight still.

Jonathan-Kelly:
It successfully captures that 'adorable alien' feel of the real life version. 👍

Maybe it could have the `skirmisher` ability since it can _jump over_ enemy zones of control?

CelticMinstrel:
This runs into the same problem the dunefolk had / have – marksman applied to a melee ability breaks translations. I guess it's not enough to be a blocker, but…

doofus-01:
I've updated the configuration, removing Marksman (which was originally chosen for their complicated eyes) in favor of Drain, which is a bit more stereotypical arthropod, but isn't poison or slow.  A slowing/paralyzing venom could make sense, but that's too similar to the existing Giant Spider anyways, so I didn't look into it much.  

> Maybe it could have the skirmisher ability since it can jump over enemy zones of control?

That's not a bad idea, but I'd want to work out the animations first.  Maybe there can be some variant with this in BfW 1.19?  This one spider is all I can realistically get done in time for BfW 1.18.

I'll plan to merge this once it passes CI and the copyright CSV is updated.  The fluff should probably be worked on a bit more, I don't remember if that is possible in beta.

doofus-01:
I'm not totally confident about the specs on this spider, but time's running out.  I'll merge this, and the Windows Terrain thing (#7930), then update the copyrights CSV in a separate commit.

Issue #7768, opened 2023-07-04
Missing updates for empty team labels in multiplayer 


Efrindil:
### Game and System Information

- **Version:** 1.16.9
- **Downloaded from:** Steam
- **Build info:** -
- **OS:** Windows 10

### Description of the bug

When you update an existing map label with an empty string. Only the local label is updated and not the labels of all team members.

### Steps to reproduce the behavior

1. Start a multiplay Game
2.  Create a Map label via Right click -> Set Label -> a non-empty arbitrary text
3. Set the Label to empty via Right click -> Set Label -> empty String

Now only the player who set the label has an empty label. All other players have the non-empty arbitrary text.

### Expected behavior

The expected behavior would be that all players have an empty string for that label. 

### Additional context

Probably the problem lies in this method https://github.com/wesnoth/wesnoth/blob/ade6e5803a6273cabd17e88628b6ad2cb469e0d6/src/map/label.cpp#L147
Since the updates are only carried out when the labels are not empty

soliton-:
The issue is rather that setting to empty removes the label and removing labels is not recorded at all.

https://github.com/wesnoth/wesnoth/blob/ade6e5803a6273cabd17e88628b6ad2cb469e0d6/src/menu_events.cpp#L832-L835

A problem there is that set_label also returns a null pointer if no label was removed. It might not be an issue to record a removal anyway but probably better to only record one if a label was actually removed.

sanchisingh01:
i would like to work on this issue

Pentarctagon:
Go ahead then.

sanchisingh01:
I wanted to ask, I am not able to build this project from my laptop, I'm having some trouble setting up the prerequisites . I've made some changes, so how do I proceed?  this is my first time contributing

Pentarctagon:
What OS are you using?

sanchisingh01:
windowa 11

sanchisingh01:
windows  11

Pentarctagon:
Alright, the Windows instructions are at https://github.com/wesnoth/wesnoth/blob/master/INSTALL.md#windows

ecomitz:
> windows 11

Per chance, are you still working on this one? Last we heard anything was August

ecomitz:
I created video documentation of this issue, which can be found here: 

https://www.youtube.com/watch?v=d5Ywn80WFQY 

[...]

PR #7942, opened 2023-10-07
Fix: global label update bug with empty string

raj007-star:
#7768 

# Bug-fix
**This commit implements a fix for a bug related to updating global labels with an empty string. The issue was that when a player updated an existing map label with an empty string, only the local label was updated, and not the labels of all team members.**

**To address this problem, the code was modified to correctly handle the update of global labels with empty strings. When a label is updated with an empty string, it now removes the old label for the current team and restores the global label in the same spot if there was one. This ensures that all players see an empty string for that label, as expected.**

**The changes made in this commit ensure that the behavior of updating map labels with empty strings is consistent for all players in a multiplayer game.**

```
const terrain_label* map_labels::set_label(const map_location& loc,
		const t_string& text,
		const int creator,
		const std::string& team_name,
		const color_t color,
		const bool visible_in_fog,
		const bool visible_in_shroud,
		const bool immutable,
		const std::string& category,
		const t_string& tooltip)
{
	terrain_label* res = nullptr;

	// See if there is already a label in this location for this team.
	// (We do not use get_label_private() here because we might need
	// the label_map as well as the terrain_label.)
	team_label_map::iterator current_label_map = labels_.find(team_name);
	label_map::iterator current_label;

	if(current_label_map != labels_.end() &&
		(current_label = current_label_map->second.find(loc)) != current_label_map->second.end())
	{
		// Found old checking if need to erase it
		if(text.str().empty()) {
			// Erase the old label.
			current_label_map->second.erase(current_label);

			// Remove the label from other team members
			for(auto& team : labels_) {
				if(team.first != team_name) {
					auto itor = team.second.find(loc);
					if(itor != team.second.end()) {
						team.second.erase(itor);
					}
				}
			}

			// Restore the global label in the same spot, if any.
			if(terrain_label* global_label = get_label_private(loc, "")) {
				global_label->recalculate();
			}
		} else {
			current_label->second.update_info(
				text, creator, tooltip, team_name, color, visible_in_fog, visible_in_shroud, immutable, category);

			res = &current_label->second;
		}
	} else if(!text.str().empty()) {
		// See if we will be replacing a global label.
		terrain_label* global_label = get_label_private(loc, "");

		// Add the new label.
		res = add_label(
			*this, text, creator, team_name, loc, color, visible_in_fog, visible_in_shroud, immutable, category, tooltip);

		// Hide the old label.
		if(global_label != nullptr) {
			global_label->recalculate();
		}
	}

	categories_dirty = true;
	return res;
}

```
*The key change made in the updated code is within the block that handles labels with empty strings. It ensures that when a label is updated with an empty string and removed, the global label (if any) is correctly restored in the same location. This change ensures that all players see an empty string for that label when it is updated, addressing the bug you described.*

### DIFF
--- a/src/map/label.cpp
+++ b/src/map/label.cpp
@@ -145,59 +145,60 @@ void map_labels::set_team(const team* team)
 }
 
 const terrain_label* map_labels::set_label(const map_location& loc,
-    const t_string& text,
-    const int creator,
-    const std::string& team_name,
-    const color_t color,
-    const bool visible_in_fog,
-    const bool visible_in_shroud,
-    const bool immutable,
-    const std::string& category,
-    const t_string& tooltip)
+		const t_string& text,
+		const int creator,
+		const std::string& team_name,
+		const color_t color,
+		const bool visible_in_fog,
+		const bool visible_in_shroud,
+		const bool immutable,
+		const std::string& category,
+		const t_string& tooltip)
 {
-    terrain_label* res = nullptr;
+	terrain_label* res = nullptr;
 
-    // See if there is already a label in this location for this team.
-    team_label_map::iterator current_label_map = labels_.find(team_name);
-    label_map::iterator current_label;
+	// See if there is already a label in this location for this team.
+	// (We do not use get_label_private() here because we might need
+	// the label_map as well as the terrain-label.)
+	team_label_map::iterator current_label_map = labels_.find(team_name);
+	label_map::iterator current_label;
 
-    if (current_label_map != labels_.end() &&
-        (current_label = current_label_map->second.find(loc)) != current_label_map->second.end())
-    {
-        // Found old label, checking if it needs to be erased
-        if (text.str().empty()) {
-            // Erase the old label.
-            current_label_map->second.erase(current_label);
-        }
-        else {
-            current_label->second.update_info(
-                text, creator, tooltip, team_name, color, visible_in_fog, visible_in_shroud, immutable, category);
-            res = &current_label->second;
-        }
-    }
-    else if (!text.str().empty()) {
-        // Add the new label.
-        res = add_label(
-            *this, text, creator, team_name, loc, color, visible_in_fog, visible_in_shroud, immutable, category, tooltip);
-    }
+	if(current_label_map != labels_.end() &&
+		(current_label = current_label_map->second.find(loc)) != current_label_map->second.end())
+	{
+		// Found old label, checking if it needs to be erased
+		if(text.str().empty()) {
+			// Erase the old label.
+			current_label_map->second.erase(current_label);
+		} else {
+			current_label->second.update_info(
+				text, creator, tooltip, team_name, color, visible_in_fog, visible_in_shroud, immutable, category);
+				
+			res = &current_label->second;
+		}
+	} else if(!text.str().empty()) {
+		// Add the new label.
+		// See if we will be replacing a global label.
+		res = add_label(
+			*this, text, creator, team_name, loc, color, visible_in_fog, visible_in_shroud, immutable, category, tooltip);
+	}
 
-    // Check if it's a global label (team_name is empty) and apply the update to all labels.
-    if (team_name.empty()) {
-        for (auto& team_labels : labels_) {
-            if (team_labels.first != team_name) {
-                terrain_label* global_label = team_labels.second[loc];
-                if (global_label) {
-                    global_label->update_info(
-                        text, creator, tooltip, team_labels.first, color, visible_in_fog, visible_in_shroud, immutable, category);
-                }
-            }
-        }
-    }
+	// Check if it's a global label (team_name is empty) and apply the update to all labels.
+	if(team_name.empty()) {
+		for(auto& team_labels : labels_) {
+			if(team_labels.first != team_name) {
+				terrain_label* global_label = team_labels.second[loc];
+				if (global_label) {
+					global_label->update_info(
+						text, creator, tooltip, team_labels.first, color, visible_in_fog, visible_in_shroud, immutable, category);
+				}
+			}
+		}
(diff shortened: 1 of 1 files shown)


Pentarctagon:
Please use tabs instead of spaces so that the changes you made can actually be seen in the diff.

raj007-star:
Sure, doing it now.

src/map/label.cpp
```diff
@@ -145,60 +145,60 @@ void map_labels::set_team(const team* team)
...
-		const bool visible_in_fog,
-		const bool visible_in_shroud,
-		const bool immutable,
-		const std::string& category,
-		const t_string& tooltip)
-{
-	terrain_label* res = nullptr;
-
-	// See if there is already a label in this location for this team.
-	// (We do not use get_label_private() here because we might need
```
CelticMinstrel, line 161: Why did you remove this comment? I don't see anything to suggest it's no longer relevant?

src/map/label.cpp
```diff
@@ -145,60 +145,60 @@ void map_labels::set_team(const team* team)
...
+
+    if (current_label_map != labels_.end() &&
+        (current_label = current_label_map->second.find(loc)) != current_label_map->second.end())
+    {
+        // Found old label, checking if it needs to be erased
+        if (text.str().empty()) {
+            // Erase the old label.
+            current_label_map->second.erase(current_label);
+        }
+        else {
```
CelticMinstrel, line 172: Please keep this on the same line as the opening brace, like it was previously.

src/map/label.cpp
```diff
@@ -145,60 +145,60 @@ void map_labels::set_team(const team* team)
...
+            current_label->second.update_info(
+                text, creator, tooltip, team_name, color, visible_in_fog, visible_in_shroud, immutable, category);
+            res = &current_label->second;
+        }
+    }
+    else if (!text.str().empty()) {
```
CelticMinstrel, line 178: Again, same line as closing brace.

src/map/label.cpp
```diff
@@ -145,60 +145,60 @@ void map_labels::set_team(const team* team)
...
+        res = add_label(
+            *this, text, creator, team_name, loc, color, visible_in_fog, visible_in_shroud, immutable, category, tooltip);
+    }
+
+    // Check if it's a global label (team_name is empty) and apply the update to all labels.
+    if (team_name.empty()) {
+        for (auto& team_labels : labels_) {
+            if (team_labels.first != team_name) {
+                terrain_label* global_label = team_labels.second[loc];
+                if (global_label) {
```
CelticMinstrel, line 189: I'm pretty sure our style guide dictates no space between if/for and the opening parenthesis.

CelticMinstrel:
It can still be reviewed by adding `?w=1` to the end of the URL, but the whitespace will still need to be fixed before it can be merged.

src/map/label.cpp
raj007-star: I am working on it. Give me some time.

src/map/label.cpp
```diff
@@ -167,10 +167,20 @@ const terrain_label* map_labels::set_label(const map_location& loc,
 		(current_label = current_label_map->second.find(loc)) != current_label_map->second.end())
 	{
 		// Found old checking if need to erase it
-		if(text.str().empty()) {
+		if (text.str().empty()) {
```
CelticMinstrel, line 170: Another space to delete.

raj007-star:
Done!

src/map/label.cpp
Pentarctagon, line 170: Remove the added space.

src/map/label.cpp
```diff
@@ -190,7 +200,7 @@ const terrain_label* map_labels::set_label(const map_location& loc,
 			*this, text, creator, team_name, loc, color, visible_in_fog, visible_in_shroud, immutable, category, tooltip);
 
 		// Hide the old label.
-		if(global_label != nullptr) {
+		if (global_label != nullptr) {
```
Pentarctagon, line 203: Remove the added space.

raj007-star: Done. Check once, where the spaces shouldn't be.

raj007-star: I checked and removed the spaces.

soliton-:
I don't really see how this fixes the bug. How did you test it?

soliton-:
I don't really see how this fixes the bug. How did you test it?

raj007-star:
> I don't really see how this fixes the bug. How did you test it?

I am new to this don't have much experience regarding testing.

soliton-:
If you cannot reproduce a bug it's pretty difficult to fix it. You must have done something to get to the code in this PR?

Pentarctagon:
If you didn't test it, how do you know this fixes the issue?

raj007-star:
> If you didn't test it, how do you know this fixes the issue?

Hi,
I appreciate your feedback and concern. I understand the importance of thoroughly testing changes before pushing them to the repository, and I apologize for not doing so in this instance. I am new to this and I just wanted to contribute.

The reason I believed the code would address the issue is that I followed best practices and reviewed the code carefully, and it appeared to target the specific problem described in the issue. However, I realize that my assumption doesn't replace proper testing.

I have learned from this experience, and I will make sure to rigorously test any future contributions to ensure they work as intended. If you'd like, I can perform testing now to confirm that the code fixes the issue properly.

Thank you for your understanding.

Pentarctagon:
As a general statement: please never submit untested code. Unless the code and the overall code base being changed are both trivial, carefully reading the code is not a substitute for actual testing.

gfgtdf:
> I don't really see how this fixes the bug.

I second this, the code doesnt relate to the observed problem (label updates not being sent to other players) at all.

PR #7943, opened 2023-10-09
update copyrights.csv

doofus-01:
for #7930 and #7934


Pentarctagon:
The results printed out by the CI look really odd. What was the `update_copyrights` command you used?

doofus-01:
I used just the defaults (no additional arguments), from the base directory.  First attempt, I overwrote the original `copyrights.csv` with the output.  Second attempt, I manually moved rows to where I thought they should be.  Just realizing now that the hashes are offset, I'll try fixing that.

But generally, yeah, this is a PITA, something must not be set up right.  I'm not going to solve this tonight, but it might just be some configuration on my end.

Pentarctagon:
How I do it is:
* run just the script (no arguments), which creates an `output.csv` in the same directory
* replace the existing `copyrights.csv` with `output.csv`
* all the rows that need updates should be at the top of the csv

Given some rows are getting two columns with a has in them, my guess would be that you might need to do a git pull to get the most recent version of the script and `copyrights.csv`.

Pentarctagon:
Pushed the update earlier today.

PR #7945, opened 2023-10-09
New units & stragglers balance

Hejnewar:
Im not even writing changelog for this, all (or 80%) was added / changed during this patch anyway, 10% is pretty much bugfix and what remains are maybe ant queens unless they were added during this patch as well. 

If no objections mergin withing like soon(tm).

Fail CI I dare u.


Hejnewar:
What art do you have here for review? I dont see any.

Jonathan-Kelly:
What about #7909 ?

knyghtmare:
> What art do you have here for review? I dont see any.

Yeah, it's not required so I un ticked him as reviewer. It's purely stat balancing.

Issue #7944, opened 2023-10-09
[filter_ability] oddity when type_value is empty


soliton-:
https://github.com/wesnoth/wesnoth/blob/627244b0fde9301ff88283185bba4e161bacb518/src/units/unit.cpp#L1424-L1444

This function returns true early if `type_value` is empty but if there is an empty value in a comma separated list it's used anyway. Also the schema seems to suggest an empty value is valid.

I think the goal is to support checking for an ability with neither `value`, `add`, `sub`, `multiply` nor `divide`? If so the initial empty check should use `blank` or `has_attribute` instead. Otherwise empty values need to be removed or cause an error or something for lists as well.

Also this should be documented on the wiki.

soliton-:
Looks like this was added in #7616 by @newfrenchy83.

newfrenchy83:
oddity fixed in https://github.com/wesnoth/wesnoth/pull/7924/commits/eed6edc9bdbeadbddb6180fd6d968537279818aa

soliton-:
@newfrenchy83 Write documentation for the attribute `type_value`.
I've explained how the function `type_value_if_present` is wrong. Your commit does not fix that. Just explain how it should work. Do not write any code. Use english words.

newfrenchy83:
The purpose of value_type is to check which numeric attribute is used in an ability, if the filter is not present or is empty, it returns true by default, otherwise, you must first check if one of these attributes is present before comparing with the filter, the problem you raise is that I did not impose on the filter to only be able to take value, add tec as valuer, right? So I'm going to change that immediately.

If https://github.com/wesnoth/wesnoth/pull/7924#issuecomment-1751679935 is merged, then this filter will be obsolete and may be deprecated.

CelticMinstrel:
My understanding was that `value_type` is meant to detect which of the add/sub/multiply/divide/value attributes are used. For example, if you want to match only multiplicative abilities, you can write `value_type=multiply`.

newfrenchy83:
> My understanding was that `type_value` is meant to detect which of the add/sub/multiply/divide/value attributes are used. For example, if you want to match only multiplicative abilities, you can write `type_value=multiply`.
> 
> I don't see how that's obsoleted by #7924.

after bug fixed use multiply=0-4 will be same effect and limit matche to value between 0 and 4 without need of value_type, but if you don't want use a value range then value_type is not obsoleted

newfrenchy83:
[...]

PR #7946, opened 2023-10-12
Remove [filter_ability]type_value=

stevecotton:
This addresses #7944 without addressing #7923. The code and schema changes are a subset of #7924, the test changes are a subset of those in #7924 with an extra assert added on top. I think #7924 should probably be merged, but there's a few changes still under discussion, and so it would be worth merging this and then rebasing to reduce the size of #7924.

There are questions about how `[filter_ability]type_value=` should treat empty values. The use cases that we can think of can be implemented in other ways, for example, if we want to look for big damage boosts, then we could filter for `add=5-infinity` `[or]` `multiply=2-infinity`. The simple answer seems to be to remove `type_value=` from the API before we freeze for 1.18.

Make the ability filter tests stricter. With the default first_time_only=yes then these events would be triggered at most once even if the filter would have passed multiple times.

### DIFF
--- a/data/schema/filters/abilities.cfg
+++ b/data/schema/filters/abilities.cfg
@@ -16,6 +16,5 @@
 	{SIMPLE_KEY affect_self s_bool}
 	{SIMPLE_KEY affect_allies s_bool}
 	{SIMPLE_KEY affect_enemies s_bool}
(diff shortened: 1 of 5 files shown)


soliton- (approved):
Didn't look too closely at the tests.

PR #7950, opened 2023-10-15
1.17 - Last-minute string cleanups

doofus-01:
This should make the fluff tied to the ravens less placeholder-y.  I'll see if there are any other strings I added that need a quick revision before the freeze, but if 1.17.22 can't wait any longer, this can be merged (assuming it doesn't fail CI for some reason).

EDIT: As much as I dislike the current name of the jumping spiders, I'm having trouble thinking of a clear improvement.
EDIT2:  This is probably all I'll get done, will merge in a couple hours

### DIFF
--- a/data/core/units.cfg
+++ b/data/core/units.cfg
@@ -355,7 +355,9 @@ Orcs who were not the strongest of their litter tend to specialize in other skil
         plural_name= _ "race^Ravens"
         description= _ "Ravens are general scavengers, often associating with hunters or predators in the hope of sharing in the kill.  They do not have the powerful beak or claws of the gryphons or raptors, but can still be a threat because of their intelligence and ability to recognize any advantage.
 
-No predators have the numbers to rival orcs or humans, so these races have constant interaction with ravens.  In the aftermath of a battle, the dead and wounded are a banquet for ravens, thus ravens are a nuisance for allies of the defeated but are seen by some victors as part of martial ceremonies.  As intelligent scavengers, ravens do not need war dead to survive.  Some farmers struggle to keep ravens away from freshly sown seeds, but others are happy to have them around as sentries."
+In the aftermath of a battle, the dead and wounded are a banquet for ravens, thus ravens are a nuisance for allies of the defeated but are seen by some victors as part of martial ceremonies.  The violence and large numbers that characterize both orcs and humans bring these races into constant interaction with ravens in this context.  For this reason, the birds have become characters in fables, parables, and superstitions related to war.
+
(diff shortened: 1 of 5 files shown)


stevecotton:
Several instances of double-spaces after a full stop.

data/core/units.cfg
```diff
@@ -355,7 +355,9 @@ Orcs who were not the strongest of their litter tend to specialize in other skil
...
 
-No predators have the numbers to rival orcs or humans, so these races have constant interaction with ravens.  In the aftermath of a battle, the dead and wounded are a banquet for ravens, thus ravens are a nuisance for allies of the defeated but are seen by some victors as part of martial ceremonies.  As intelligent scavengers, ravens do not need war dead to survive.  Some farmers struggle to keep ravens away from freshly sown seeds, but others are happy to have them around as sentries."
+In the aftermath of a battle, the dead and wounded are a banquet for ravens, thus ravens are a nuisance for allies of the defeated but are seen by some victors as part of martial ceremonies.  The violence and large numbers that characterize both orcs and humans bring these races into constant interaction with ravens in this context.  For this reason, the birds have become characters in fables, parables, and superstitions related to war.
```
stevecotton, line 358: "large numbers" seems more like it should be "large numbers of casualties". However, it feels like too much dwelling on the consequences of violence, ignoring the normal handwave of video games forgetting about the victims.

data/core/units.cfg
```diff
@@ -355,7 +355,9 @@ Orcs who were not the strongest of their litter tend to specialize in other skil
...
+
+As intelligent scavengers, ravens do not need war dead to survive.  Some farmers struggle to keep ravens away from freshly sown seeds, but others are happy to have them around as sentries."
```
stevecotton, line 360: Sentries against what?

doofus-01: Intruders, visitors, anything.

data/core/units.cfg
doofus-01: Large numbers could also be the populations in general.  Either interpretation seems OK to me.

doofus-01:
> Several instances of double-spaces after a full stop. They're already in the text, so this isn't a regression, but it would still be good to fix before the string freeze.

Maybe a bit of a tangent, but what are the chances we can ever drop that rule?  Does double-space break anything?

Pentarctagon:
At this point I think it's not that it'd break anything so much as it is for consistency with all the rest of the game's text.

doofus-01:
OK.  It will be an ongoing issue, since double-space was pounded into me as a student and I've never heard a convincing argument to break the habit (and make monospace harder to read).  But I'll try here.

PR #7951, opened 2023-10-15
Fix crash when mousing over a village in the terrain editor

stevecotton:
I merged #7920, missing @soliton- 's comment that - "Actually I get an out of range exception in the editor when hovering over a village. In scenario mode it works fine. I'm guessing the owner_team stuff needs checking."

This is the fix, and most of it is just a whitespace change for the extra level of indentation.


stevecotton:
@Pentarctagon this should merge before the tagging, I'm only leaving it open in case soliton has comments.

Issue #7949, opened 2023-10-14
Add hooks for more mouse events


daniel-santos:
### Describe the desired feature

Hello,

I would love to be able to hook the right-click event, prior to the pop-up menu displaying so I could modify a menu entry based upon context. However, it looks like I'll have to do that in mouse-over, which will be more CPU wasteful. All-in-all, it would be nice to just have access to a greater number of them and let the add-on designer just not write crappy code that would lock up the game.

CelticMinstrel:
Does a right-click trigger the existing mouse pressed event? I don't remember if I've checked that.

gfgtdf:
> Does a right-click trigger the existing mouse pressed event? I don't remember if I've checked that.

Pretty sure it doesnt, The mostclick event is iirc implemented via one of the "select/move/attack" ,  "Move/Attack" Or  "Select" hotkeys, and right-click menu bypass the hotkey system iirc.

doofus-01:
Is filtering the context for the various menu items to show up not enough?  Can you post an example?

CelticMinstrel:
It's not enough when the text needs to be different for every hex and can dynamically change from turn to turn.

I heard that variables in menu descriptions are also not substituted before the menu is shown, so perhaps that should be opened as a separate issue. If the menu item name contains `$turn_number` (which would be written in the WML as `$|turn_number` to defer substitution) then it should show the current turn number each time you open the menu.

doofus-01:
> It's not enough when the text needs to be different for every hex and can dynamically change from turn to turn.

The hexes can be certainly be filtered.  I've never tried for turns but it should be possible, especially if #7948 is fixed.

So, still need an example.

CelticMinstrel:
#7948 is what I referenced in the second paragraph of my previous comment.

I suppose it's not impossible to add one menu item for every hex, but that seems like a kinda crazy workaround for this and #7948…

doofus-01:
Crazy workaround for what?  Example, please.

gfgtdf:
> However, it looks like I'll have to do that in mouse-over, which will be more CPU wasteful

I want to point out that there are a lot this things happening on "mouse over" already, in particular all the calculations for the right side panel, in particular the attack stats, like checking the resistance of all attacks against all enemy unit types, and checking whether abilities are active.

[...]

PR #7952, opened 2023-10-15
Add lua mouse button event handler

daniel-santos:
This is a flexible event handler mechanism to allow for more advanced add-on features. It does not allow the add-on to consume the value however, so there's no mechanism for the add-on to requrest the default action not be executed.

We would normally deal separately with mouse down, up and click, with the later being an event where the mouse has gone both down and up on the same UI component. However, Wesnoth's code seems to allow a mouse up on many UI components to serve as a click, so this implementation follows that behavior.

This addresses issue #7949

I'm still in the process of testing, but would like to get this in before the freeze!

### DIFF
--- a/src/mouse_events.cpp
+++ b/src/mouse_events.cpp
@@ -603,13 +603,7 @@ void mouse_handler::mouse_motion(int x, int y, const bool browse, bool update, m
 bool mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_t button,
 									   map_location loc, bool click)
 {
-	if (gui().view_locked() || button < SDL_BUTTON_LEFT || button > SDL_BUTTON_X2) {
-		return false;
-	} else if (event.state > SDL_PRESSED || !gui().get_map().on_board(loc)) {
-		return false;
-	}
-
-	static const std::string buttons[6] = {
+	static const std::array<const std::string, 6> buttons = {
 		"",
 		"left",		// SDL_BUTTON_LEFT
 		"middle",	// SDL_BUTTON_MIDDLE
@@ -617,14 +611,15 @@ bool mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_
 		"mouse4",	// SDL_BUTTON_X1
 		"mouse5"	// SDL_BUTTON_X2
 	};
-	static const std::string events[3] = {
-		"up",		// SDL_RELEASED
-		"down",		// SDL_PRESSED
-		"click"
-	};
+
+	if (gui().view_locked() || button < SDL_BUTTON_LEFT || button > buttons.size()) {
+		return false;
+	} else if (event.state > SDL_PRESSED || !gui().get_map().on_board(loc)) {
+		return false;
+	}
 
 	if(game_lua_kernel* lk = pc_.gamestate().lua_kernel_.get()) {
-		lk->mouse_button_callback(loc, buttons[button], events[event.state]);
+		lk->mouse_button_callback(loc, buttons[button], (event.state == SDL_RELEASED ? "up" : "down"));
 
 		// Are we being asked to send a click event?
 		if (click) {
@@ -633,7 +628,7 @@ bool mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_
 				return false;
 			}
 			// We allow this event to be consumed, but not up/down
-			return lk->mouse_button_callback(loc, buttons[button], events[2]);
+			return lk->mouse_button_callback(loc, buttons[button], "click");
 		}
 	}
 	return false;
--- a/src/mouse_events.cpp
+++ b/src/mouse_events.cpp
@@ -599,7 +599,7 @@ void mouse_handler::mouse_motion(int x, int y, const bool browse, bool update, m
 
 // Hook for notifying lua game kernel of mouse button events. We pass button as
 // a serpaate argument than the original SDL event in order to manage touch
-// emulation (i.e., long touch = right click) and such.
+// emulation (e.g., long touch = right click) and such.
 bool mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_t button,
 									   map_location loc, bool click)
 {
@@ -610,17 +610,17 @@ bool mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_
 	}
 
 	static const std::string buttons[6] = {
-		[0]					= "",
-		[SDL_BUTTON_LEFT]	= "left",
-		[SDL_BUTTON_MIDDLE]	= "middle",
-		[SDL_BUTTON_RIGHT]	= "right",
-		[SDL_BUTTON_X1]		= "x1",
-		[SDL_BUTTON_X2]		= "x2",
+		"",
+		"left",		// SDL_BUTTON_LEFT
+		"middle",	// SDL_BUTTON_MIDDLE
+		"right",	// SDL_BUTTON_RIGHT
+		"mouse4",	// SDL_BUTTON_X1
+		"mouse5"	// SDL_BUTTON_X2
 	};
 	static const std::string events[3] = {
-		[SDL_RELEASED]		= "up",
-		[SDL_PRESSED]		= "down",
-		[2]					= "click",
+		"up",		// SDL_RELEASED
+		"down",		// SDL_PRESSED
+		"click"
 	};
 
 	if(game_lua_kernel* lk = pc_.gamestate().lua_kernel_.get()) {
--- a/src/mouse_events.cpp
+++ b/src/mouse_events.cpp
@@ -597,6 +597,21 @@ void mouse_handler::mouse_motion(int x, int y, const bool browse, bool update, m
 	}
 }
 
+// Hook for notifying lua game kernel of mouse button events. We pass button as
+// a serpaate argument than the original SDL event in order to manage touch
+// emulation (i.e., long touch = right click) and such.
+void mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_t button, map_location loc)
+{
+	if (loc == map_location::null_location())
+		return;
+
+	if(game_lua_kernel* lk = pc_.gamestate().lua_kernel_.get()) {
+		// Ever any reason to allow lua functions to consume an event? If so,
+		// would have to make sure state stays correctly updated.
+		lk->mouse_button_callback(loc, button, event.state == SDL_PRESSED);
+	}
+}
+
 unit_map::iterator mouse_handler::selected_unit()
 {
 	unit_map::iterator res = find_unit(selected_hex_);
--- a/src/mouse_events.cpp
+++ b/src/mouse_events.cpp
@@ -604,11 +604,22 @@ void mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_
 {
 	if (loc == map_location::null_location())
 		return;
+	if (button < SDL_BUTTON_LEFT || button > SDL_BUTTON_X2)
+		return;
+
+	static const std::string buttons[6] = {
+		[0]					= "",
+		[SDL_BUTTON_LEFT]	= "left",
+		[SDL_BUTTON_MIDDLE]	= "middle",
+		[SDL_BUTTON_RIGHT]	= "right",
+		[SDL_BUTTON_X1]		= "x1",
+		[SDL_BUTTON_X2]		= "x2",
+	};
 
 	if(game_lua_kernel* lk = pc_.gamestate().lua_kernel_.get()) {
 		// Ever any reason to allow lua functions to consume an event? If so,
 		// would have to make sure state stays correctly updated.
-		lk->mouse_button_callback(loc, button, event.state == SDL_PRESSED);
+		lk->mouse_button_callback(loc, buttons[button], event.state == SDL_PRESSED);
 	}
 }
 
--- a/src/mouse_events.cpp
+++ b/src/mouse_events.cpp
@@ -600,12 +600,14 @@ void mouse_handler::mouse_motion(int x, int y, const bool browse, bool update, m
 // Hook for notifying lua game kernel of mouse button events. We pass button as
 // a serpaate argument than the original SDL event in order to manage touch
 // emulation (i.e., long touch = right click) and such.
-void mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_t button, map_location loc)
+bool mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_t button,
+									   map_location loc, bool click)
 {
-	if (loc == map_location::null_location())
-		return;
-	if (button < SDL_BUTTON_LEFT || button > SDL_BUTTON_X2)
-		return;
+	if (gui().view_locked() || button < SDL_BUTTON_LEFT || button > SDL_BUTTON_X2) {
+		return false;
+	} else if (event.state > SDL_PRESSED || !gui().get_map().on_board(loc)) {
+		return false;
+	}
 
 	static const std::string buttons[6] = {
 		[0]					= "",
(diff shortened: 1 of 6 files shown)


CelticMinstrel:
Looking over it, my only complaint would be that the button should be passed to the Lua handler as a string, not a number. That's the normal way to handle an enumerated value in Lua (there's even a function, `luaL_checkoption`, to grab one from the stack. I don't see a push version of that unfortunately, so you'll probably need to use a switch-case here.)

src/mouse_events.cpp
```diff
@@ -597,6 +597,21 @@ void mouse_handler::mouse_motion(int x, int y, const bool browse, bool update, m
...
+// a serpaate argument than the original SDL event in order to manage touch
+// emulation (i.e., long touch = right click) and such.
+void mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_t button, map_location loc)
+{
+	if (loc == map_location::null_location())
+		return;
+
+	if(game_lua_kernel* lk = pc_.gamestate().lua_kernel_.get()) {
+		// Ever any reason to allow lua functions to consume an event? If so,
+		// would have to make sure state stays correctly updated.
```
gfgtdf, line 610: I think the answer here is yes, iirc one of the main complaints of the existing lua mouse clicked event was that there is no option to consume the event.

gfgtdf:
i don't really see a problem with this

but some question are in my mind right now, which imo should me mentioned in a possible documentation for it.:

1) does this trigger only when clicked on the map?
2) if not, what is the value of the loc parameter when the mouse was clicked ouf of map?
3) does this trigger when clicking on the minimap?
4) How does it  transfer to other input types (like the steamdeck and mobile ports) ?
5) Does it also trigger while other dialogs are open (like [message]?)
6) Does it also trigger while the map is blocked for another reason ? 


In the long run we should also decide whether it really makes sense to have two events that do similar things, and if yes maybe the existing one should be renamed (since its iirc tied to the "select" hotkey is could for example be rename into file_selected if we want to keep it )

src/mouse_events.cpp
daniel-santos: ah hah! Well then let's doing right. I'm not 100% versed in every mousey use-case in Wesnoth, as we can be in a drag event at the mouse up. I think I need to study the code a bit more to be clear on how far we drag before we start considering it a drag event and such.

Maybe it would be a good idea to separate it out into mouse up, down and click with the later only occurring if the up event is in the same hex as the down -- I can be a sloppy clicker sometimes, but this should be a big enough target! For the sake of the lua support, I'm only supporting mouse button events on hex tiles, so that would probably be a reasonable way to distinguish a mouse up that's part of a drag event and one that should be considered a real "click"

src/mouse_events.cpp
```diff
@@ -604,11 +604,22 @@ void mouse_handler::mouse_button_event(const SDL_MouseButtonEvent& event, uint8_
...
+		return;
+
+	static const std::string buttons[6] = {
+		[0]					= "",
+		[SDL_BUTTON_LEFT]	= "left",
+		[SDL_BUTTON_MIDDLE]	= "middle",
+		[SDL_BUTTON_RIGHT]	= "right",
+		[SDL_BUTTON_X1]		= "x1",
+		[SDL_BUTTON_X2]		= "x2",
+	};
```
CelticMinstrel, line 617: Okay, so I have two comments here.

1. Unfortunately, designated initializers are new in C++20, but Wesnoth builds as C++17. There are a couple of options here; I don't think compatibility wrappers can work for this feature, but I'm willing to be proven wrong; or you could use a mapping type (`std::map` or `std::unordered_map`).
2. `x1` and `x2` seem a bit too obtuse for my liking. I think I'd prefer using `ext1` and `ext2`, or maybe `extra` or `extended`.

daniel-santos:
Hello there! I guess I covered a tiny bit of this in the above comment.
>     1. does this trigger only when clicked on the map?
That was my thought on a reasonable implementation. Maybe this could go further? I would rather not ever give the lua later the low-level x,y pixel info, so I would rather translate any such events into a UI element and a logical target.

>     2. if not, what is the value of the loc parameter when the mouse was clicked ouf of map?
Always valid.

>     3. does this trigger when clicking on the minimap?
No, but should there be a separate hook for that?

>     4. How does it  transfer to other input types (like the steamdeck and mobile ports) ?
Yoikes, I have no idea! But I'm following the mechanism used to determine if it's a _logical_ right, middle of left click -- so the logic in [`mouse_handler_base::is_left_click()`](https://github.com/wesnoth/wesnoth/blob/9183ee6791a4154e97b37e66a616e7e4f7d4a6ac/src/mouse_handler_base.cpp#L250C26), et. al., along with transforming a long-touch into a right click (or mouse down) event.

>     5. Does it also trigger while other dialogs are open (like [message]?)
I'll have to test this case.

>     6. Does it also trigger while the map is blocked for another reason ?
I think that it may. `mouse_handler_base::is_left_click()` will return false if `command_active()`, which I guess is an apple keyboard thing, but it doesn't look like it's avoiding other states. I'll have to dig deeper into this one.

> In the long run we should also decide whether it really makes sense to have two events that do similar things, and if yes maybe the existing one should be renamed (since its iirc tied to the "select" hotkey is could for example be rename into tile_selected if we want to keep it )

I think I'll have to agree with you on that. I didn't read too deeply on `on_mouse_action`, but I would agree they could be merged. This will take a bit more consideration.

CelticMinstrel:
> I think I'll have to agree with you on that. I didn't read too deeply on `on_mouse_action`, but I would agree they could be merged. This will take a bit more consideration.

I think it does kinda make sense to have "low-level" mouse events (press, release, click, drag, etc) and "high-level" events (tile selected, unit selected, drag a unit, etc), but I haven't really given this a lot of thought…

daniel-santos:
> > ```
> > 2. if not, what is the value of the loc parameter when the mouse was clicked ouf of map?
> > ```
> 
> ~Always valid.~ ah hah! looks like I need to check `loc.valid()`

Actually I'm checking `loc == map_location::null_location()` which is better than `loc.valid()`, but neither assures we're clicking a valid hex tile. I presume it's possible to click a tile just on the edge of the map and `display::hex_clicked_on()` would return that logical tile coordinate, even though it was outside of the map's bounds?

daniel-santos:
> > I think I'll have to agree with you on that. I didn't read too deeply on `on_mouse_action`, but I would agree they could be merged. This will take a bit more consideration.
> 
> I think it does kinda make sense to have "low-level" mouse events (press, release, click, drag, etc) and "high-level" events (tile selected, unit selected, drag a unit, etc), but I haven't really given this a lot of thought…

My thought was to focus only on the logical mouse events that (I would think) a mod developer would be concerned with -- then again, I also wouldn't want to super-impose my ideas about that on a mod developer.

While I'm only focused on mouse events on the map, there could also be reason to care about such events on the mini map or gui component. My thought would be for those to be separate hooks. So perhaps this should be called `on_map_mouse_button` and any such hooks for the minimap or other GUI component given another name? Thinking it through, if I have a spell that a character casts on themselves, it might be neat to be able to right-click on that effect and remove it if it's something the character should be able to turn off at will.

As far as consuming an events, if we allow a mouse-down to be consumed, then we'll want to make sure we nuke the mouse up as well. Or simply restrict event consumption to logical clicks -- where the mouse went down and up in the same hex?

But I hadn't thought about passing on the mouse drag information. I can definitely see that being helpful in some cases!

CelticMinstrel:
> My thought was to focus only on the logical mouse events that (I would think) a mod developer would be concerned with -- then again, I also wouldn't want to super-impose my ideas about that on a mod developer.

No, that's fine. Even if there were separate "low-level" and "high-level" events like I mentioned, they are kind of unrelated and don't necessarily belong in the same pull request.

> So perhaps this should be called `on_map_mouse_button`

For a name I suppose I would lean towards something like `on_click_hex` or `on_hex_mouseup`. I don't know if it even makes sense to have such hooks for the GUI or the minimap.

> But I hadn't thought about passing on the mouse drag information. I can definitely see that being helpful in some cases!

You mean like adding an extra parameter to this callback which is the hex that the mouse was pressed down on? That sounds like it might be a bit difficult, but I could definitely see it being useful.

> Or simply restrict event consumption to logical clicks -- where the mouse went down and up in the same hex?

If not doing something like the above, then this is probably better, as a drag probably shouldn't trigger a click event.

gfgtdf:
> That was my thought on a reasonable implementation. Maybe this could go further? I would rather not ever give the lua later the low-level x,y pixel info, so I would rather translate any such events into a UI element and a logical target.

Only triggering when the actual map is clicked is fine.

> No, but should there be a separate hook for that?

hmm i don't think it needed, i mainly asked because iirc  on_mouse_action can actually trigger via the minimap since it seems like  you select units through the minimap iirc (not saying that this makes sense based on how tiny the tiles on the minimap are.) (i might be wrong here)

> Yoikes, I have no idea! But I'm following the mechanism used to determine if it's a _logical_ right, middle of left click -- so the logic in [`mouse_handler_base::is_left_click()`](https://github.com/wesnoth/wesnoth/blob/9183ee6791a4154e97b37e66a616e7e4f7d4a6ac/src/mouse_handler_base.cpp#L250C26), et. al., along with transforming a long-touch into a right click (or mouse down) event.

hmm sounds good so far, tbh i don't really know how the other input types work exactly.

> I'll have to test this case.

I'd be better if it didn't imo. 

> I think I'll have to agree with you on that. I didn't read too deeply on `on_mouse_action`, but I would agree they could be merged. This will take a bit more consideration.

I forgot that it's actually called on_mouse_action, i think that name is probably good enouth for now.

stevecotton:
How does it respond when the called function takes a long time to run, or even when the called function tries to show a dialog box? Does it avoid [the issue with on_mouse_action](https://wiki.wesnoth.org/LuaAPI/wesnoth/game_events#wesnoth.game_events.on_mouse_action), where the UI can respond as if the mouse was clicked on a different hex?

daniel-santos:
> How does it respond when the called function takes a long time to run, or even when the called function tries to show a dialog box? Does it avoid [the issue with on_mouse_action](https://wiki.wesnoth.org/LuaAPI/wesnoth/game_events#wesnoth.game_events.on_mouse_action), where the UI can respond as if the mouse was clicked on a different hex?

This is the type of issue that's important in cases like this. The documentation needs to inform users that handling this function should be done quickly or a coroutine/thread (or whatever lua uses) should be used. For my particular use case, I need to finish my code prior to the menu being shown. (I was actually trying to split my `[set_menu_item]` into two pieces so that I would have to update less when the event is run.)

Frankly, I wanted to submit a second patch to clean up the cases where we call `sdl::get_mouse_state()` right after we receive an event because it's wasteful, but further reading informed me why that problem was happening. But has somebody used to being in the kernel, it makes me shudder.

Also, we can at least generate a warning when mouse event handlers are taking too long -- not on the very first time, but if it happens more than once or twice.

daniel-santos:
I probably should have mentioned that I sure would hope to get this in before the feature freeze.

> I forgot that it's actually called on_mouse_action, i think that name is probably good enouth for now.

So you're suggesting to re-purpose the existing `on_mouse_action`? That will, of course, break existing mods, but I suppose that may be no big deal. Alternatively, we use a a new function name for callbacks if it is defined and only call `on_mouse_action` if it isn't defined. So for example, if we add a new `on_mouse_button` event handler and that is defined, then we only call `on_mouse_button`, but if it isn't defined, then we check for `on_mouse_action`. This would not break legacy code and also have a more accurate name for the functionality.

daniel-santos:
@stevecotton I think a better idea is to add a flag to detect such blocking infractions (like throwing up a message box) and ban the callback until the game is restarted, that'll learn 'em! :grin: It would be a tiny bit more code.

Another thought is a separate "notify" function that runs asynchronously and that the add-on can take forever with -- obviously, it won't be able to consume any events.

CelticMinstrel:
> So you're suggesting to re-purpose the existing `on_mouse_action`?

Definitely not. We can't reuse that name for a different purpose.

> So for example, if we add a new `on_mouse_button` event handler and that is defined, then we only call `on_mouse_button`, but if it isn't defined, then we check for `on_mouse_action`. This would not break legacy code and also have a more accurate name for the functionality.

I think `on_mouse_action` is essentially a more "high-level" event, so if `on_mouse_button` consumes the event then it should not be triggered. However, basing it on whether the event is defined at all is no good, as two different add-ons might define the two events and that situation would mean that one add-on breaks solely because the other one is active.

gfgtdf:
> So you're suggesting to re-purpose the existing on_mouse_action? That

No no, i was just refering to my first posts comment "Maybe we should rename the existing event" comment.

daniel-santos:
> If the feature is ready and complete, then I disagree.
> 
> It currently is not, however.

Almost done! I should have it in about 30 more minutes! I just want to make sure everything is correct. Also, I think I found a bug in the touch handling :)

daniel-santos:
So here's the revised patch. I had to do a lot of analysis and I have a list of things to fix for when ever the 1.19 branch is created, as the mouse code just needs a complete overhaul and clean up. Features have grown, as features do, and the incremental modifications have started to add up. I'll discuss that more in the other PR.

But I believe that what I'm doing here is the best that can be done for now. The "click" events for left, right and middle mouse buttons are occurring immediately following mouse down. This isn't the traditional behavior, but it'll do for now. These events can be consumed, which will inhibit the default behavior of Wesnoth. Buttons X1 and X2 will "click" after the button comes up, and only if it comes up in the same map hex it went down in.

The code will not discard any mouse activity when the GUI is locked or a map hex was clicked that isn't actually part of the loaded map.

src/mouse_events.cpp
```diff
@@ -597,6 +597,48 @@ void mouse_handler::mouse_motion(int x, int y, const bool browse, bool update, m
...
+	static const std::string events[3] = {
+		[SDL_RELEASED]		= "up",
+		[SDL_PRESSED]		= "down",
+		[2]					= "click",
+	};
+
+	if(game_lua_kernel* lk = pc_.gamestate().lua_kernel_.get()) {
+		lk->mouse_button_callback(loc, buttons[button], events[event.state]);
```
gfgtdf, line 627: It'd be better if this code wouldn't segfault (and either ignore it or throws an exception) when `button` or `event.state` is for whatever reason out of range.

gfgtdf: edit: okay i just see you check for it above.

src/mouse_events.cpp
```diff
@@ -597,6 +597,48 @@ void mouse_handler::mouse_motion(int x, int y, const bool browse, bool update, m
...
+
+		// Are we being asked to send a click event?
+		if (click) {
+			// Was both the up and down on the same map tile?
+			if (loc != drag_from_hex_) {
+				return false;
+			}
+			// We allow this event to be consumed, but not up/down
+			return lk->mouse_button_callback(loc, buttons[button], events[2]);
```
gfgtdf, line 636: Is `lk->mouse_button_callback`called twice when `click` is true?

gfgtdf: hmm okay i see

src/mouse_events.cpp
```diff
@@ -597,6 +597,48 @@ void mouse_handler::mouse_motion(int x, int y, const bool browse, bool update, m
 	}
 }
 
+// Hook for notifying lua game kernel of mouse button events. We pass button as
+// a serpaate argument than the original SDL event in order to manage touch
+// emulation (i.e., long touch = right click) and such.
```
CelticMinstrel, line 602: Very minor quibble, but this seems like it should be e.g., not i.e.

src/mouse_events.cpp
CelticMinstrel: This has still not been addressed.

src/mouse_events.cpp
daniel-santos: yup, it should be!

src/mouse_events.cpp
daniel-santos: Sorry I missed this! Hrm, I could always just use a plain array initializer and put the enums in comments. I like having the code prove that the mapping is correct. You can probably tell that I've done most of my code in C for the past few years. :)

```C++
static const std::string buttons[6] = {
	"",
	"left",		// SDL_BUTTON_LEFT
	"middle",	// SDL_BUTTON_MIDDLE
	"right",	// SDL_BUTTON_RIGHT
	"x1",		// SDL_BUTTON_X1
	"x2",		// SDL_BUTTON_X2
};
```
?

src/mouse_events.cpp
daniel-santos: So yes. You get one event for mouse down, one for mouse up and one for click, although not necessarily in that order. For left and middle buttons, you get down, click, then up. For the rest, you get down, up, then click.

src/mouse_events.cpp
CelticMinstrel: I definitely like the idea of using designated initializers here too, but I'm pretty certain it's going to fail to compile on some platform we support. (It would work if we enabled GNU extensions, but then that could cause problems on Windows.)

That proposal isn't how I'd do it, but I don't really have a problem with it.

Don't forget item 2 of my previous comment.

daniel-santos: Thanks, got the grammar too!

daniel-santos: "mouse-4" and "mouse-5"?

daniel-santos:
That last patch isn't necessary -- it can be added on later if need be.

src/mouse_events.cpp
gfgtdf: I think it'd be still better if this was more straightforward to see, especially now that theose constants are no longer used in the array initializer. maybe replace `std::string buttons[6]` with `std::array<std::string, 6> buttons` and compare the parameter then to `buttons.size()` instead of `SDL_BUTTON_X2` (same for `events`)

daniel-santos: okie dokie

daniel-santos: We'll just get rid of the events array. It was only ever there for a slight reduction in the construction of std::basic_string objects, which is super-light weight anyway.

PR #7956, opened 2023-10-16
Correct drag_distance calculation and adjust drag_threshold() to match

daniel-santos:
If the user drags 13 pixels on the x or y axis, but none on the other, then it will not be considered a drag event. However, dragging just 3 pixels in both the x and y direction, a total distance of 4.25 pixels, will be counted as a drag. This patch calculates the correct length and triggers a drag if the total distance is at least 4 pixels in any direction.


daniel-santos:
Oh, I didn't notice that the `drag_threshold()` was also squared, so as to not need the `sqrt` call. My mistake,

PR #7957, opened 2023-10-16
Replace wesnoth-map-diff with a submodule

cooljeanius:
I recently created a separate repository for the wesnoth-map-diff too, which is here: 
https://github.com/cooljeanius/wesnoth-map-diff
It preserves the existing history of the directory, while allowing it to develop separately, and also be used as a submodule in other UMC that want to make use of it as well.
cc @macabeus @Pentarctagon @stevecotton @soliton-


CelticMinstrel:
The only thing I would say is that the repository should probably be imported into the wesnoth organization and referenced from there, instead of a repository owned by you. Obviously that's not something you can do though…

cooljeanius:
> The only thing I would say is that the repository should probably be imported into the wesnoth organization and referenced from there, instead of a repository owned by you. Obviously that's not something you can do though…

If I add you as a collaborator to the repo, could you transfer the ownership? Actually let me try...

CelticMinstrel:
Do I even have that level of permission in the wesnoth organization…? I wouldn't expect to.

Pentarctagon:
I believe I'd need to do that, or at least someone with the Owner role.

cooljeanius:
> I believe I'd need to do that, or at least someone with the Owner role.

It says your invite to collaborate on the repo expired; do you know if there's a way that I could re-send it?
Screen Shot 2023-10-16 at 16 25 49

Pentarctagon:
Maybe remove me and re-add me? I don't recall ever getting an invite.

Pentarctagon:
Is this actually useful to do though, given git still treats submodules as a second class feature?

cooljeanius:
> Is this actually useful to do though, given git still treats submodules as a second class feature?

I mean, I think it is; it's already made maintenance of "A New Order" a lot simpler now that I've merged nemaara/A_New_Order#146
I was also thinking that I could set up Dependabot to keep the submodule up-to-date here, too, if that'd help simplify things

CelticMinstrel:
> this actually useful to do though, given git still treats submodules as a second class feature?

This particular submodule is also something where it wouldn't hurt anyone's development progress if it didn't get downloaded.

Pentarctagon:
Is there a reason it needs to be a submodule rather than just cloning it separately? I'd rather not have to deal with the submodule getting downgraded because it'll probably update more frequently than mariadbpp or lua.

cooljeanius:
> Is there a reason it needs to be a submodule rather than just cloning it separately? I'd rather not have to deal with the submodule getting downgraded because it'll probably update more frequently than mariadbpp or lua.

Well, if we're going to go the route of having Dependabot upgrade it, how often it updates would depend upon how often we configure Dependabot to check for updates. I currently have it set to "weekly" in my fork, but we could potentially bump it up to "monthly" for mainline Wesnoth, if "weekly" is too often...

CelticMinstrel:
> I'd rather not have to deal with the submodule getting downgraded

Do you mean whatever it is that has caused people to accidentally commit a Lua downgrade several times now? What _does_ cause that anyway? `git add -u` and `git commit -a` don't touch submodules.

soliton-:
> > Is this actually useful to do though, given git still treats submodules as a second class feature?
> 
> I mean, I think it is; it's already made maintenance of "A New Order" a lot simpler now that I've merged [nemaara/A_New_Order#146](https://github.com/nemaara/A_New_Order/pull/146)

This says nothing. What did it improve?

If there is an actual reason to put this one tool into a separate repository unlike all other tools then explain it clearly.

cooljeanius:
> > > Is this actually useful to do though, given git still treats submodules as a second class feature?
> > 
> > 
> > I mean, I think it is; it's already made maintenance of "A New Order" a lot simpler now that I've merged [nemaara/A_New_Order#146](https://github.com/nemaara/A_New_Order/pull/146)
> 
> This says nothing. What did it improve?
> 
> If there is an actual reason to put this one tool into a separate repository unlike all other tools then explain it clearly.

- language statistics are no longer confused
- dependabot updates no longer trigger a bunch of irrelevant CI runs 
- submodule can use its own separate set of GitHub Actions that don't interfere with the main set of actions
- development on it can proceed independently
- sources for it no longer have to be copied back and forth between all the other repositories that might want to use it, since they can all just use the same submodule

soliton-:
> * language statistics are no longer confused

Where do you see it show up in the language statistics? The statistics are "confused" anyway since pretty much no tool is implemented in C++ or lua or WML.

> * dependabot updates no longer trigger a bunch of irrelevant CI runs

dependabot is not even enabled.

> * submodule can use its own separate set of GitHub Actions that don't interfere with the main set of actions

What interferes with what?

> * development on it can proceed independently

How is that an advantage? Sounds like the opposite to me.

> * sources for it no longer have to be copied back and forth between all the other repositories that might want to use it, since they can all just use the same submodule

You can checkout a subdir of the wesnoth repo if you want to. It sounds like you want a github action for this though which could certainly be done and is unrelated to a separate repo.

cooljeanius:
> > * language statistics are no longer confused
> 
> Where do you see it show up in the language statistics? The statistics are "confused" anyway since pretty much no tool is implemented in C++ or lua or WML.
> 

I meant in the other campaigns that use the map-diff tool that can now remove its sources and switch it to a submodule.

> > * dependabot updates no longer trigger a bunch of irrelevant CI runs
> 
> dependabot is not even enabled.
> 

I have it enabled in my fork, and separately in the separate wesnoth-map-diff repo. Even if Wesnoth doesn't want to enable it for the mainline repo, I think it would still be worthwhile to allow it to be enabled specifically for the wesnoth-map-diff module, as that seems to kind of be the expectation for node.js modules.

> > * submodule can use its own separate set of GitHub Actions that don't interfere with the main set of actions
> 
> What interferes with what?

Check the additional workflows I added: 
https://github.com/cooljeanius/wesnoth-map-diff/tree/main/.github/workflows
We don't need things like eslint of njsscan looking in the rest of the repository, or causing failures for unrelated changes to unrelated parts.

> > * development on it can proceed independently
> 
> How is that an advantage? Sounds like the opposite to me.
> 

It means that proposed changes to it no longer have to get lost amongst the rest of the changes to rest of the repository. Keeping unrelated things separate from one another is better for organizational purposes.

> > * sources for it no longer have to be copied back and forth between all the other repositories that might want to use it, since they can all just use the same submodule
> 
> You can checkout a subdir of the wesnoth repo if you want to. It sounds like you want a github action for this though which could certainly be done and is unrelated to a separate repo.

Both checking out a subdir of the wesnoth repo, and creating a GitHub Action for this, would be significantly more difficult than going the submodule route.

macabeus:
I'm probably a bit later for the party, but...
I don't think we need a git submodule here *either* keeping it on this repository.

One of the advantages of JS ecosystem is that it's trivial to work using a CLI tool.
In the case of Node.js (the environment for `wesnoth-map-diff`) we can use [`npx`](https://docs.npmjs.com/cli/commands/npx).

The flow would be:
- We have a new repository, `wesnoth/wesnoth-map-diff`, used only to develop this CLI tool
- Then, whenever necessary, we publish a new version for this tool; it's as simple as running `npm run publish`
- It publishes a new NPM package, available on [`npmjs.com`](npmjs.com), and that can run as a CLI tool with `npx wesnoth-map-diff`
- Then, on GitHub Actions as [`map-diff.yml`](https://github.com/wesnoth/wesnoth/blob/master/.github/workflows/map-diff.yml), we don't need to install and build the project. We just need to run `npx wesnoth-map-diff`. It's possible to force a specific version (e.g., `npx wesnoth-map-diff@1.2.3`)

I have the above workflow in other private projects, and it works very well.
It avoids the issues when using git submodules, and makes it easier to work locally, since new users just need to run a single CLI command to use it on their projects.

cooljeanius:
> I'm a bit late for the party, but... I don't think we need a git submodule here _either_ keeping it on this repository.
> 
> One of the advantages of the JS ecosystem is that it's trivial to work using a CLI tool. In the case of Node.js (the environment used on `wesnoth-map-diff`) we can run [`npx`](https://docs.npmjs.com/cli/commands/npx).
> 
> The flow would be:
> 
>     * We have a new repository, `wesnoth/wesnoth-map-diff`, used only to develop this CLI tool
> 
>     * Then, whenever necessary, we publish a new version for this tool; it's as simple as running `npm run publish`
> 
>     * It publishes on NPM, available on [`npmjs.com`](npmjs.com). We can run this package as a CLI with `npx wesnoth-map-diff`
> 
>     * Then, on GitHub Actions like [`map-diff.yml`](https://github.com/wesnoth/wesnoth/blob/master/.github/workflows/map-diff.yml?rgh-link-date=2023-11-01T18%3A09%3A58Z), we don't need to install and build the whole project. Instead, we just need to run `npx wesnoth-map-diff`. If wanted, it's possible to force a specific version (e.g., `npx wesnoth-map-diff@1.2.3`)
> 
> 
> I have the above workflow in other private projects, and it works very well. It avoids the issues when using git submodules and makes it easier to reuse it in different projects since now we just need to run a single CLI command to run it anywhere.

ok, so which one of us would do the publishing step? You? Me? Pent? Someone else?

macabeus:
I think you can do it since you are the one leading this task

soliton-:
> Both checking out a subdir of the wesnoth repo, and creating a GitHub Action for this, would be significantly more difficult than going the submodule route.

For the record checking out a subdir from a repo amounts to writing a path in the checkout action. Creating a github action is pretty much equivalent to what everyone writing a workflow to use this is doing anyway.

If the actual goal here is to take over maintenance of wesnoth-map-diff then by all means do that. I wouldn't call it unrelated but I'm fine with ignoring it going forward.

cooljeanius:
> I think you can do it since you are the one leading this task. It's possible to create an NPM account for Wesnoth and publish it directly using their name.

I think I'd still want to get approval from @Pentarctagon first before doing this...

Pentarctagon:
What is meant by:
> create an NPM account for Wesnoth
?

macabeus:
@Pentarctagon [NPM](https://www.npmjs.com/) is the package manager for Node.js.
You can have an account to publish packages that will be available there.

It's common for individuals, companies, and open-source projects to have an account there to publish their packages under a name.

In the same way we have a `wesnoth` org on GitHub, we can have a `wesnoth` account on NPM.

It isn't mandatory, tho.
Since you can invite collaborators, or change the package owner.

cooljeanius:
> @cooljeanius
> 
> > I think I'd still want to get approval from @Pentarctagon first before doing this...
> 
> I still think you don't need to overthink it.
> 
> Just do it.
> 
> Creating a new `npx` package isn't rocket science.

OK, I published it: https://www.npmjs.com/package/wesnoth-map-diff

Pentarctagon:
Is there any reason for this to remain open?

cooljeanius:
@Pentarctagon I fixed the copyright failure with [4714dac](https://github.com/wesnoth/wesnoth/pull/7957/commits/4714dacf716d3fb4e0830c68ede651a89571ce45).

Pentarctagon:
I mean, I'm still not really in favor of this being a submodule.

Pentarctagon:
To be more explicit: dealing with submodules tends to be an annoyance, so I actively don't want to add more submodules to Wesnoth's repository. I would rather this be a separate repository and then cloned in as needed independently rather than add another submodule. So I'm leaning towards closing this PR.

Pentarctagon:
Well, I suppose not closed - ideally the map diff action would be updated to clone in and use the map-diff repository. If other people want to use it as a submodule, that's their decision.

cooljeanius:
OK I guess the annoyance of submodules has kinda been getting to me lately, too; I keep having to delete the wesnoth-map-diff subdirectory just to be able to switch branches, because apparently `git stash` doesn't work when you've got a submodule with dirty status... will close for now, but I still think it'd be worthwhile to find something to do here, and it might be worth revisiting the submodule approach if upstream `git` ever improves their handling of submodules...

PR #7940, opened 2023-10-04
EI Barrow Wight fix, licensing update - attempt #2

Dalas121:
Round 2 of the EI barrow wight fix and licensing update.

> Looks like EI's Barrow Wight unit configuration file was mistakenly deleted at some point, most likely when the Skeleton Rider/Bone Knight sprites were being merged into mainline. This breaks scenario 14_Drowned_Plains, since the Barrow Wight enemy leader doesn't spawn. This PR restores the Barrow Wight's cfg.
> 
> This PR also removes the Skeleton Rider2's cfg, since it was (I think?) mistakenly left in when its sprite was merged into mainline.
> 
> This PR also includes fixes to the attribution for pacifist.png and lamp-genie.png, and removes the unneeded barrow-wight.webp file (thanks Pentarctagon).

PR #7958, opened 2023-10-16
EI: Fix a double-space in the Horse Barrow Wight's description

stevecotton:
The text seems odd too, maybe there's more to do here.


Dalas121:
I think I meant to say "is shrouded behind cloak and bone", but accidentally deleted the "shrouded" at some point, leaving the double space behind.  Does that sound better?

stevecotton:
It was such a simple whitespace change that I assumed it worked and did not test.

This unit doesn't show up in the game, and using debug mode to skip to EI S14 prints `error engine/team_construction: game_error: unknown unit type: Barrow Wight`. This .cfg file is in the wrong directory, there's `images/units/` in the pathname.

@Dalas121, please would you do a PR with the full fix? I'll log a bug for it.

stevecotton:
On second thoughts, closing the PR is the wrong thing to do - I can at least get the right string in before the freeze, even if other stuff is wrong. The next push will make this a definite improvement, fixing the missing unit part of #7959, but I'd hesitate to mark that bug as fixed without proper testing.

Dalas121:
Well this is bizarre.  In the 1.17 version that I have from steam, the Barrow_Wight.cfg file looks to be correctly in the `/units` folder, but in my recent PR it looks like I accidentally moved it to `/images`.  Except IIRC, that PR was to fix the `Undead_Horse_Barrow_Wight.cfg` file being entirely missing, so I'm not sure how it's placed correctly on steam's 1.17 version.

Regardless, looks like the original mistake was mine, and this PR seems to fix it.

Wedge009:
I'm sorry I missed this review request. There are some spelling errors in this line (unrelated to the change) with delirium and essence. @Pentarctagon I can't remember: are spelling corrections sufficient to break string freeze?

Pentarctagon:
From https://forums.wesnoth.org/viewtopic.php?t=52785:
> For text containing full sentences, changing or adding strings is allowed when the text is plain wrong in a way that would cause issues for players relying on the text for something else (ie: if the objectives say to kill a named enemy unit that no longer exists in the scenario). This means there will be no minor fixes for punctuation, grammar, etc.

Though I wonder if that's really a good thing, since it's just delaying an inevitable change. Is this a brand new string for 1.17 or an existing string?

Wedge009:
It's a new file, only added in #7940 and updated in this PR.

Pentarctagon:
I'd say just fix it, and I can add a note about it in the email I still need to send to the translators mailing list about the freeze.

Wedge009:
Okay. I'm checking for anything else that needs correcting before doing so.

Dalas121:
Thanks for noticing and correcting both of those typos.

PR #7961, opened 2023-10-17
Add a "Teleport Unit (Debug!)" option to debug menu

### LINKED ISSUE #5649: Add a "Teleport Unit (Debug!)" option to debug menu
I'm on 1.14.16 downloaded from Steam for Mac.

Currently when enabling debug mode, it adds options to create units, kill units, and change unit side to the contextual menu (see screenshot below). It would be helpful if there were also an option to teleport an existing unit somewhere else to that menu in debug mode, to make it easier to test certain events that depend upon certain units triggering them. Using `unit moves=100` isn't as practical because there might be enemy units in the way, or impassable terrain. 

### Screenshots
Screen Shot 2021-03-31 at 02 04 51

TheHitsy:
resolves #5649 

Added a teleport debug option which works similarly to how move works, you right click a unit and select the teleport debug option (or hover over a unit and press shift + t) and then press left click on the tile you want to teleport to. There is no limitation over where or how far you can teleport as long as it is not on top of another unit.


gfgtdf:
the move.cpp code isn't that trivial so its not easy to review, but from what i saw if i didn't miss anything:
1) You call resources::undo_stack->clear(); in the replay code but not in the original code, the undo stack must be cleared in the original code too.
2) You fire the enter/exit_hex events but not the moveto event, this might be correct but i'd still like to hear the rationale for that.
3) There seem to be lines that set variables that are never used afterwards.

stevecotton:
We're already in string freeze, and this adds a new string.

knyghtmare:
> We're already in string freeze, and this adds a new string.

toss into the 1.19 bucket maybe?

cooljeanius:
Thanks for creating this PR! It'll probably need to be squashed when merging, however, due to how many commits it contains... (unless, of course, you'd like to rebase/squash it yourself first...)

TheHitsy:
Thanks for the responses! I did not know there was a string freeze currently inplace, so this can wait to be merged until a later more appropriate time. 
I agree that this contains a lot of commits and needs to be squashed but am a bit unsure of how this is done, if it can be completed when merging or if I need to squash it on the fork itself. 
Questions from @gfgtdf  
1 and 2: Toss that to me being not 100% sure about how every event and bookkeeping works, I discovered that I needed to use enter/exit_hex events to completely update everything about the units such as health bars and such. This should however be fixed similarly to the first question by calling post_move. 
3: If i'm not going completely blind I should now have removed the unnecessary variables.

CelticMinstrel:
> I agree that this contains a lot of commits and needs to be squashed but am a bit unsure of how this is done, if it can be completed when merging or if I need to squash it on the fork itself.

If it's to be squashed into a single commit, that can easily be done at merge time. If you wanted to be squashed into more than one commit, you should probably rebase it and do the squash yourself.

TheHitsy:
To me there is no point in not squashing everything into one commit at merge time so it depends on how everyone else wants it.

Pentarctagon:
Alright, coming back to this finally - @TheHitsy any chance you're still around?

src/modules/lua
cooljeanius, line 1: uh, is the lua submodule update necessary, or did that just slip in?

cooljeanius (approved):
LGTM, although I haven't actually tested it myself...

src/modules/lua
Pentarctagon: No, it shouldn't be in here. It's another example of how submodules are annoying though.

Pentarctagon:
Superseded by #8767 due to issues rebasing around a merge commit and the lua submodule issue.

Issue #7724, opened 2023-06-17
Review the English text in GUI.pyw


stevecotton:
PR #7627 made the strings in GUI.pyw translatable, but didn't change the English text. The current text probably isn't clear enough for an English-speaking audience, at least not to a wide enough audience to be ready as an input to translation.

For example:

* "Ignore refcount 0 on names matching regexp", should probably be described as unused functions and variables, however I haven't reviewed what exactly it does.
* "Ignore files matching the specified regular expression", should say filenames.

I think we should review the entire text before adding `wesnoth-tools` to the stats on https://gettext.wesnoth.org/ , and we should warn the translators that it's still being revised in the 1.17.18 release announcement.

(The "Prose" tag doesn't quite fit, but it's the closest for "text that will need translation".)

CelticMinstrel:
I don't think the prose label fits even slightly. Prose is a particular type of writing that is not present in the wesnoth-tools domain at all.

Slayer95:
> Ignore refcount 0 on names matching regexp

That text doesn't actually exist in master anymore. It currently reads ``"Allow unused macros with names matching regexp:"``

> "Ignore files matching the specified regular expression", should say filenames.

Not sure whether that actually improves the readability in English, but a translator note would help.

stevecotton:
We should apply the [UI Style Guide](https://wiki.wesnoth.org/UI_Style_Guide), for example dialog titles should be title case.

Placeholders in strings could be consistent. There's some using `{}`, and some using `{0}`. None are using named placeholders, but named placeholders would be consistent with our C++ and WML styles.

A lot of the strings seem to be strings from the individual tools, so maybe we need to translate them and then update the strings here.

These will need string changes (#7728 might work in English, but would need a translation hint)
* #7726
* #7727
* #7728 

Comments on strings:
```diff
+# Where is this displayed? If "open" is a verb, why is there an option in the program to open itself?
 #: data/tools/GUI.pyw:58
 msgid "Open a Graphical User Interface (GUI) to WML Tools"
 
+# What about the language? Why is this in caps?
 #: data/tools/GUI.pyw:59
 msgid "LANGUAGE"
 
+# More helpful would be to use the references that the translators are already looking at,
+# for example, to point to the list of languages on g.w.o., assuming that the languages that
+# use only two letters in their .po file names use the two_TWO format for this option.
 #. TRANSLATORS: ISO 15897 "Procedures for the registration of cultural elements" specifies use
[...]

PR #7964, opened 2023-10-19
Post-Freeze Corrections

Wedge009:
* Barrow Wight: discussed in #7958.
* UtBS abilities: looks like originally 'taunt' description had 'for a turn', then changed to 'for one turn' as per 'taunted' description. Also added punctuation for clarity.
* Roc: mentioned in #7902.
* Shadow Jumping Spider: spelling error.
* GUI descriptions: found these while looking for 'straight' apostrophes. I'm not sure if the user sees these, but the strings aren't marked for translation so I included them here.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
@@ -445,7 +445,7 @@ Any units adjacent to this unit will fight as if it were dusk when it is night, 
         name = _"nova"
         description= _ "This weapon, when used offensively, deals damage to all units adjacent to the caster when it hits." + "
 
-" + _ "This attack is the third verse of the Song of Sun Ascension. After using it, the first verse is availaibe in the next turn. Furthermore, invoking this attack will grant illumination for three turns."
+" + _ "This attack is the third verse of the Song of Sun Ascension. After using it, the first verse is available in the next turn. Furthermore, invoking this attack will grant illumination for three turns."
         special_note=_ "This unit has wide-area attacks centered on the caster."
         active_on = offense
     [/dummy]
@@ -1446,7 +1446,7 @@ Marksman attacks are only affected if the chance to hit is greater than 60%."
     [dummy]
         id = taunt
         name = _"taunt"
-        description= _ "Taunts the enemy making them try to attack Kaleh no matter what for a one turn."
+        description= _ "Taunts the enemy making them try to attack Kaleh, no matter what, for one turn."
     [/dummy]
 #enddef
 
@@ -1482,7 +1482,7 @@ Marksman attacks are only affected if the chance to hit is greater than 60%."
     [dummy]
         id = taunted
         name = _"taunted"
-        description= _ "This unit is taunted by Kaleh, it will try to attack him no matter what for one turn."
+        description= _ "This unit is taunted by Kaleh; it will try to attack him, no matter what, for one turn."
     [/dummy]
 #enddef
 
--- a/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
@@ -1446,7 +1446,7 @@ Marksman attacks are only affected if the chance to hit is greater than 60%."
     [dummy]
         id = taunt
         name = _"taunt"
-        description= _ "Taunts the enemy making them try to attack Kaleh, no matter what, for one turn."
+        description= _ "Taunts the enemy making them try to attack Kaleh for one turn."
     [/dummy]
 #enddef
 
@@ -1482,7 +1482,7 @@ Marksman attacks are only affected if the chance to hit is greater than 60%."
     [dummy]
         id = taunted
         name = _"taunted"
-        description= _ "This unit is taunted by Kaleh; it will try to attack him, no matter what, for one turn."
+        description= _ "This unit is taunted by Kaleh, it will try to attack him for one turn."
     [/dummy]
 #enddef
 
--- a/data/core/units/monsters/Roc.cfg
+++ b/data/core/units/monsters/Roc.cfg
@@ -17,7 +17,7 @@
     {AMLA_DEFAULT}
     cost=70
     usage=scout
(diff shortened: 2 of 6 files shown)


Pentarctagon:
I kind of wonder if it might make sense to delay the string freeze by a month at this point honestly. This and #7865 have string changes, and then #7965 and #7724 would really be good to get in so we don't create a bunch of rework for translators later.

Wedge009:
That sounds fair, but I think the judgement on that is yours as manager.

Sort-of-related: I wonder if there's a way to review recent string changes easily. Generally I only catch these issues when I review the English localisation and I think those only get updated with po/pot update (I'm not familiar with the process/terminology).

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1446,7 +1446,7 @@ Marksman attacks are only affected if the chance to hit is greater than 60%."
     [dummy]
         id = taunt
         name = _"taunt"
-        description= _ "Taunts the enemy making them try to attack Kaleh no matter what for a one turn."
+        description= _ "Taunts the enemy making them try to attack Kaleh, no matter what, for one turn."
```
nemaara, line 1449: Actually it doesn't need the no matter what if you don't mind changing that in this PR.

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -1482,7 +1482,7 @@ Marksman attacks are only affected if the chance to hit is greater than 60%."
     [dummy]
         id = taunted
         name = _"taunted"
-        description= _ "This unit is taunted by Kaleh, it will try to attack him no matter what for one turn."
+        description= _ "This unit is taunted by Kaleh; it will try to attack him, no matter what, for one turn."
```
nemaara, line 1485: Ditto.

nemaara (approved):

stevecotton:
> Sort-of-related: I wonder if there's a way to review recent string changes easily. Generally I only catch these issues when I review the English localisation and I think those only get updated with po/pot update (I'm not familiar with the process/terminology).

Other methods could be as good, but I doubt any will be better than these:
* Building the .po files and then diffing the .pot files
* Building the .po files and using Poedit on a non-English language that was at 100% before. It needs to be non-English so that Gettext doesn't automatically fill in the translations with the source text, but there's no need to read the target language. I generally use `zh_CN` if `de` wasn't at 100%.

The documentation is sparse, but https://wiki.wesnoth.org/WesnothTranslationsHowTo#Tools has the commands for CMake and SCons.

data/core/units/monsters/Roc.cfg
```diff
@@ -17,7 +17,7 @@
...
     usage=scout
-    description= _ "In remote desert mountains, far from any city or settlement, enormous raptors are rumoured to nest in the lofty crags.  Travellers and nomads claim to have witnessed livestock and grown men snatched into the sky without warning by these monsters.  While one should not place much faith in the accounts of the desert roamers, they are less fantastical than the drunken lies of sailors who claim to have witnessed full-rigged ships, attempting to navigate the narrow straights between the sheer cliffs where nothing can grow, sunk by giant eagles dropping large rocks.
+    description= _ "In remote desert mountains, far from any city or settlement, enormous raptors are rumoured to nest in the lofty crags.  Travellers and nomads claim to have witnessed livestock and grown men snatched into the sky without warning by these monsters.  While one should not place much faith in the accounts of the desert roamers, they are less fantastical than the drunken lies of sailors who claim to have witnessed full-rigged ships, attempting to navigate the narrow straits between the sheer cliffs where nothing can grow, sunk by giant eagles dropping large rocks.
```
stevecotton, line 20: Still some double-spaces to catch - there's one before Travellers. Also, Firefox says there should only be one L in the `en_US` spelling of Travelers.

data/core/units/monsters/Shadow_Jumping_Spider.cfg
```diff
@@ -26,9 +26,9 @@
     cost=58
     usage=fighter
     undead_variation=spider
-    description= _ "Unlike the feared Giant Spider, that uses poison and webs to catch prey, the Shadow Jumper relies on speed and the element of surprise.  These stealthy, keen-eyed hunters creep to within striking distance, then pounce with a piercing bite that is said to punch through steel.  Once they have pierced their victim's defenses, the spiders take what nourishment is available, feeding in a manner as fast and remorseless as befits their nature.
+    description= _ "Unlike the feared Giant Spider that uses poison and webs to catch prey, the Shadow Jumper relies on speed and the element of surprise.  These stealthy, keen-eyed hunters creep to within striking distance, then pounce with a piercing bite that is said to punch through steel.  Once they have pierced their victim's defenses, the spiders take what nourishment is available, feeding in a manner as fast and remorseless as befits their nature.
```
stevecotton, line 29: Double-spaces and typography in `victim's`. If you're making a change, may as well fix it all.

stevecotton, line 29: I'd take out the first `piercing`. The bite punches through steel, and `pierced` appears in the next sentence.

data/core/units/monsters/Roc.cfg
Wedge009: Already covered in #7965. This PR is strictly for spelling/punctuation changes - Pentarctagon suggested we could push those in before releasing for translation.

It's never been my aim to 'correct' for non-en_US spellings in the source text - I leave that for native en_US readers. I do notice these, though - I think it was in the review for WoF I recall some vacillation between en_US and non-en_US spellings, for example.

data/core/units/monsters/Shadow_Jumping_Spider.cfg
Wedge009: That's a bit more in the domain of modifying the text, which I was trying to avoid given all the blockers on string modification. But since nemaara has already requested some changes I may as well follow through with this one too.

Wedge009: Of course, just demonstrates how difficult this is, I don't get everything in one go either. I'll do this in #7965.

Wedge009:
>     * Building the .po files and using Poedit on a non-English language that was at 100% before. It needs to be non-English so that Gettext doesn't automatically fill in the translations with the source text, but there's no need to read the target language. I generally use `zh_CN` if `de` wasn't at 100%.

This sounds like what I'm doing already anyway. Even if the source text is copied, I still review anything marked fuzzy which is when I catch all those mistakes in the new/modified strings.

stevecotton (approved):

demario-w:
Really a bad idea to push this together with #7965 .
The reason is that changes in #7965 will certainly lead to no update in translations while these changes will most likely do.

The best is that the update of po files can be pushed in two different commits so that the fuzzies can be identified easily and cleared in a batch.

Issue #7891, opened 2023-09-03
``[store_reachable_locations]`` behaves weird if the player has shroud


Discontinuum:
### Game and System Information

- **Version:** 1.16.5
- **Downloaded from:** flatpak

### Description of the bug

Wiki https://wiki.wesnoth.org/InternalActionsWML#.5Bstore_reachable_locations.5D for ``[store_reachable_locations]`` says that "If ``viewing_side`` left unset then fog and shroud are ignored, hidden ambushers are not ignored, and the real reach of the units is stored." However it doesn't work in fact. If player has shroud and we call ``[store_reachable_locations]`` for the enemy leader, the output is weird. The test campaign 
[Store_Reachable_Bug.zip](https://github.com/wesnoth/wesnoth/files/12507542/Store_Reachable_Bug.zip)

### Steps to reproduce the behavior

1. Install the test campaign
2. Start scenario 1. Lift shroud with debug and see that there is only one unit spawned near the undead leader
3. Hit "end turn"  to proceed to the next scenario
4. Here you don't have shroud and all the hexes reachable by the enemy leader are filled with random undead

### Expected behavior

The expected behaviour is the behaviour in scenario 2

### Additional context

If you check "inspect" in the first scenario, you'll see the only hex retrieved by ``[store_reachable_locations]`` was the hex the undead leader stands on. As we tried to spawn there, the only unit spawned was placed in a left upper adjacent hex

gfgtdf:
Confirmed , the relevant code is:

https://github.com/wesnoth/wesnoth/blob/079892b12a316d88cc628e7c40712b8c0bd01216/src/pathfind/pathfind.cpp#L418-L420

and

https://github.com/wesnoth/wesnoth/blob/079892b12a316d88cc628e7c40712b8c0bd01216/src/pathfind/pathfind.cpp#L282-L285

Wedge009:
@Discontinuum Is this still an issue, given it's been marked as ready for testing?

Discontinuum:
I checked my test campaign in 1.18, seems it works now as intended

Wedge009:
Thanks for confirming.

PR #7970, opened 2023-10-20
fix  [store_reachable_locations] #7891

gfgtdf:


stevecotton:
Haven't tested, but looks good to me.

PR #7971, opened 2023-10-21
wmltools GUI review

Wedge009:
What started as a string / translation notes review for #7724 turned into a code review as well. I separated changes into commits - I advise looking at per-commit changes instead of trying to view the differences altogether because the auto-formatting causes a lot of changes.

While I could have skipped the auto-format, I honestly find compressed code (no spacing between commas, etc) very hard to read so I think it's worth including. If I had free reign, I'd add a lot new lines between code blocks too, to help with readability and logical grouping ('Sparse is better than dense'?).

The changes for #7724 are towards the end. This PR doesn't close that issue, however, as there are a lot of strings I don't really know the meaning of - mainly the ones displayed as options for each of the tools as shown in their own tab.

I can still run the program on my own system after these changes.


Pentarctagon:
@Slayer95 's comment from Discord:

> the Windows compat issue is that getlocale() may return bogus results, causing the next line to crash... So catching AttributeError at the getlocale() call is not enough

data/tools/GUI.pyw
```diff
@@ -56,38 +56,39 @@ WESNOTH_TRAN_DIR = os.path.join(WESNOTH_ROOT_DIR, "translations")
...
-        usage=("GUI.pyw [--lang=" + _("LANGUAGE") + "]")
+        description=_("Open a Graphical User Interface (GUI) to WML Tools")
     )
     parser.add_argument(
         '--lang',
-        # TRANSLATORS: ISO 15897 "Procedures for the registration of cultural elements" specifies use
-        # of underscores in the locale id. For example, use en_GB, not en-GB.
-        help=_("Launch GUI.pyw in the given language. Language code must follow ISO 15897 (e.g. en_GB).")
+        help=_("Launch GUI.pyw in the specified language. Language code is expected as a POSIX locale name, refer to "
+               "gettext.wesnoth.org for a full list.")
```
Pentarctagon, line 74: Having a single translatable string split across multiple lines works, right? I'd assume it does, but figured I'd ask.

Wedge009: I understand it resolves to a single line. I split text as per Python conventions, though I recognise parts of this code do not follow that convention.

data/tools/GUI.pyw
```diff
@@ -668,8 +675,9 @@ class WmllintTab(Frame):
                              sticky=W,
                              padx=10)
         self.tooltip_diff = Tooltip(self.radio_diff,
-                                    # TRANSLATORS: Explanation for Diff.
+                                    # TRANSLATORS: Tool-tip explanation for diff run mode.
```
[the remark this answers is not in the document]
Wedge009: Right, I'm not used to writing in en_US. I'll change that after work.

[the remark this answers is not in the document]
Wedge009: Done.

Wedge009:
I thought there was some urgency with #7724 for an upcoming release. Is that no longer the case?

Pentarctagon:
Mainly waiting on @Elvish-Hunter to review the python changes. The text changes could go in otherwise.

Pentarctagon:
If it wouldn't be a huge hassle to split out the text changes into another PR, then that could go in right away. Otherwise, I'll merge this prior to the next release regardless and Elvish_Hunter can get to reviewing it after that. Since yeah the string changes definitely need to be in before the next release.

Wedge009:
I thought about that, but there is a bit of overlap between them. In the end, the code changes aren't that significant and aside from the auto-formatting and spelling corrections (in comments) and a couple of variable renaming, I think it's just:
* The check for the presence of `WESNOTH_TRAN_DIR` happens whether setting user locale or system locale, because both rely on the translations it contains (line 77). Before it was only checking when setting the user-specified locale.
* Specifically catching OSError when a locale is not found instead of catching all exceptions including things like interrupts (lines 91/108).
* When removing the new-line character at line 1133, also had to remove the `columnspan` argument in the subsequent `grid()` call at line 1136 so that the text and input boxes align correctly. Seems like the forced new-line was a presentational hack, but it doesn't work well with large displays like mine (UHD). Better to let Tkinter handle the layout instead.
* Adding a missing tool-tip for Quiet Mode in the `wmlindent` tab (line 1159).
* Build the initial command line variables as a list literal instead of calling `append()`.
* Deleted unnecessary reference to `self` (line 619).
* Removed redundant brackets around `control_key` variable.
* Avoid name shadowing at lines 127/131 - probably could use a better name.

data/tools/GUI.pyw
```diff
@@ -56,38 +56,39 @@ WESNOTH_TRAN_DIR = os.path.join(WESNOTH_ROOT_DIR, "translations")
...
     if opts.lang is not None:
         if not os.path.isdir(WESNOTH_TRAN_DIR):
-            showerror(_("Error"),
-                      # TRANSLATORS: {0} is "translations", the folder where compiled translation files (.mo) are stored.
-                      _("`{0}` folder not found. Please run the GUI.pyw executable packaged with the Wesnoth installation.").format(
-                          "translations")
-                      )
+            showerror(
+                _("Error"),
+                # TRANSLATORS: {0} is "translations", the directory where compiled translation files (.mo) are stored.
```
CelticMinstrel: Huh? Why is format being used at all if it's just a constant string being substituted in?

data/tools/GUI.pyw
```diff
@@ -1004,7 +1049,9 @@ class WmlindentTab(Frame):
         self.mode_frame.grid(row=0,
                              column=0,
                              sticky=N + E + S + W)
+
         self.radio_normal = Radiobutton(self.mode_frame,
+                                        # TRANSLATORS: Normal run mode for wmlindent.
```
CelticMinstrel, line 1064: The presence of separate translator notes suggests that this `_("Normal")` and the other `_("Normal")` are supposed to be separate strings that can receive different translations. How are we going to handle this here? Or, do we just accept they're one string? The normal `tag^` syntax doesn't apply since this is being passed to Tk.

CelticMinstrel:
One last thing I'd like to mention is that gettext does not appear to recognize `{}` as a format argument – the strings using it appear be missing the `#, python-brace-format` line in the pot file. I'd suggest replacing all uses of `{}` wit `{0}` to work around this.

data/tools/GUI.pyw
Wedge009: Question for @Slayer95 in #7627? But if I had to guess, I'd say it's because we want the word 'translations' not to be translated. The 'translations' string is the directory name which remains fixed.

data/tools/GUI.pyw
Wedge009: I wondered about that. I think they're just meant to be one string - I'll update the notes to refer to the tools generically.

data/tools/GUI.pyw
Wedge009: At the very least, your highlighting of this string made me realise the quote characters needed correcting so thanks for that.

Elvish-Hunter (approved):
At a quick glance everything seems fine to me. Keep in mind that I just pushed a patch that adds support for a new library which I recently found, so there's this conflict to solve, but I'm approving it anyway.

Pentarctagon:
@Elvish-Hunter would you be able to resolve the conflict via the "Resolve conflicts" button here on github?

Elvish-Hunter:
@Pentarctagon I just did that; let me know if there are any issues.

PR #7972, opened 2023-10-21
Various terrain fixes and additions

doofus-01:
First commit fixes a bad Xor/Xo* transition, I'll add whatever other fixes or transition improvements I can do in the next day or two.


doofus-01:
Not sure if the windows TOD variants are really worth pursuing, since I don't see a way to keep the TOD darkness from bleeding through into a weird halo if you try to use a different time area on the wall from the hex in front of it.

doofus-01:
copyrights.csv continues to defy my understanding, I'm starting to regret approving that thing.

PR #7975, opened 2023-10-22
Change editor event from start to prestart.

### LINKED ISSUE #7969: scenario editor adds its content in a start event.
As said in the title the editor currently adds its content ([area]s, [unit]s etc) in a start event, this means they cannot be used form prestart events,

I suggest that the editor adds those in a prestart even with a higher `priority=`,  so that other prestart events can access it, a usecase for example would be a prestart event that randomly scatters a few items in an [area] that was defined in the editor.

Pentarctagon:
fixes #7969

### DIFF
--- a/src/editor/map/map_context.cpp
+++ b/src/editor/map/map_context.cpp
@@ -404,8 +404,8 @@ config map_context::convert_scenario(const config& old_scenario)
 	}
 
 	config& event = multiplayer.add_child("event");
-	event["name"] = "start";
-	event["id"] = "editor_event-start";
+	event["name"] = "prestart";
+	event["id"] = "editor_event-prestart";
 
 	// for all children that aren't [side] or [time], move them to an event
 	// for [side]:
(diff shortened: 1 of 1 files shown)


gfgtdf:
I still think it should also have a higher `priority=` so that other prestart events are guarenteed to be able to use the stuff  added by this event.

Pentarctagon:
How high a priority would be a good value?

sevu:
The id with both _ and - is a bit complicated … maybe something shorter?

Pentarctagon:
The point is to make it somewhat complicated/unique, so that someone doesn't by chance give one of their own events the same ID.

gfgtdf:
> How high a priority would be a good value?l

Hmm, not sure, since the prior value only gets meaning compared to others priority values the first one will always be chose a bit arbitrary. I'd probably put 1000.

Pentarctagon:
Alright, updated.

PR #7977, opened 2023-10-23
UtBS: Balance the "life and life's demise" text

stevecotton:
This line has already been altered in 94b48615f, so it's going to be fuzzy for the translators anyway.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/units/quenoth/Sun_Singer.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/quenoth/Sun_Singer.cfg
@@ -19,7 +19,7 @@
     [abilities]
         {UTBS_ABILITY_HEALS}
     [/abilities]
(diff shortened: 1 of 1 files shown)


Wedge009:
If I understand nemaara's intention in rewriting this back in 80a2aa0962612b6b50b61493c7436f5a04b919ca, I read the phrase as referring to how fire/heat, as represented in the twin suns, can bring both life/energy and death/destruction. Maybe nemaara was trying for something more than just 'life and death'. I can see what you mean by 'unbalanced', but it doesn't particularly bother me, given the poetic nature of the text. Additionally, kindling seems to refer more to fuel rather than the fire itself, to me.

I've had the term 'bike-shedding' thrown at me rather upsettingly in the past - while I wouldn't use the term against someone else, I wonder if that same sentiment applies here?

In other words, I'm neutral. I defer to nemaara's decision.

nemaara:
I get your point but I also don't really like the kindling phrasing. The wording here was a real struggle for me when I wrote the description, might need to think more about it later.

Pentarctagon:
It sounds like this PR can be closed?

PR #7981, opened 2023-10-24
Get the add-on ID from the file path if possible.

### LINKED ISSUE #7979: Editor ask for addon when opening a scenario
When i try to open a sacenario in the editor, it askes me which addon is associatged to this scenario.
The editor should be able to deduce this from the path of the scenario that was opened if it was inside an addon.

Pentarctagon:
Fixes #7979

### DIFF
--- a/src/editor/map/context_manager.cpp
+++ b/src/editor/map/context_manager.cpp
@@ -915,8 +915,15 @@ void context_manager::load_map(const std::string& filename, bool new_context)
 
 	if(filesystem::ends_with(filename, ".cfg")) {
 		if(editor_controller::current_addon_id_ == "") {
-			editor_controller::current_addon_id_ = editor::initialize_addon();
-			set_addon_id(editor_controller::current_addon_id_);
+			// if no addon id has been set and the file being loaded is from an addon
+			// then use the file path to determine the addon rather than showing a dialog
+			if(editor_controller::current_addon_id_ == "" && filename.find(filesystem::get_addons_dir()) == 0) {
+				editor_controller::current_addon_id_ = filesystem::get_addon_id_from_path(filename);
+				set_addon_id(editor_controller::current_addon_id_);
+			} else {
+				editor_controller::current_addon_id_ = editor::initialize_addon();
+				set_addon_id(editor_controller::current_addon_id_);
+			}
 		}
 
 		if(editor_controller::current_addon_id_ == "") {
--- a/src/editor/map/context_manager.cpp
+++ b/src/editor/map/context_manager.cpp
@@ -917,13 +917,13 @@ void context_manager::load_map(const std::string& filename, bool new_context)
 		if(editor_controller::current_addon_id_ == "") {
 			// if no addon id has been set and the file being loaded is from an addon
 			// then use the file path to determine the addon rather than showing a dialog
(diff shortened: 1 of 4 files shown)


src/filesystem.cpp
```diff
@@ -1776,4 +1776,23 @@ std::string get_localized_path(const std::string& file, const std::string& suff)
 	return "";
 }
 
+std::string get_addon_id_from_path(const std::string& location)
+{
+	bfs::path path = bfs::absolute(location);
+	bool use_next = false;
+	std::string addon_base = filesystem::base_name(filesystem::get_addons_dir());
```
soliton-, line 1783: Why does this not compare the absolute paths and if the addons dir is a prefix then pick the first dir after the prefix? Looking for an "add-ons" dir with no context seems like an unnecessary approximation.

Pentarctagon: `bfs::path` isn't indexed as far as I know, so there doesn't appear to be a way to say "get the next directory in the path after X".

src/editor/map/context_manager.cpp
```diff
@@ -915,8 +915,15 @@ void context_manager::load_map(const std::string& filename, bool new_context)
...
-			set_addon_id(editor_controller::current_addon_id_);
+			// if no addon id has been set and the file being loaded is from an addon
+			// then use the file path to determine the addon rather than showing a dialog
+			if(editor_controller::current_addon_id_ == "" && filename.find(filesystem::get_addons_dir()) == 0) {
+				editor_controller::current_addon_id_ = filesystem::get_addon_id_from_path(filename);
+				set_addon_id(editor_controller::current_addon_id_);
+			} else {
+				editor_controller::current_addon_id_ = editor::initialize_addon();
+				set_addon_id(editor_controller::current_addon_id_);
+			}
```
soliton-, line 926: I think it's better to leave the "from an addon" check to `filesystem::get_addon_id_from_path` now.

```suggestion
			if(editor_controller::current_addon_id_ == "") {
				editor_controller::current_addon_id_ = filesystem::get_addon_id_from_path(filename);
			}
			if(editor_controller::current_addon_id_ == "") {
				editor_controller::current_addon_id_ = editor::initialize_addon();
			}
			set_addon_id(editor_controller::current_addon_id_);
```

soliton- (approved):

src/editor/map/context_manager.cpp
```diff
@@ -915,7 +915,14 @@ void context_manager::load_map(const std::string& filename, bool new_context)
 
 	if(filesystem::ends_with(filename, ".cfg")) {
 		if(editor_controller::current_addon_id_ == "") {
-			editor_controller::current_addon_id_ = editor::initialize_addon();
+			// if no addon id has been set and the file being loaded is from an addon
+			// then use the file path to determine the addon rather than showing a dialog
+			if(editor_controller::current_addon_id_ == "") {
+				editor_controller::current_addon_id_ = filesystem::get_addon_id_from_path(filename);
+			}
```
soliton-, line 922: Sorry, just noticed the previous if already checks this...

```suggestion
			editor_controller::current_addon_id_ = filesystem::get_addon_id_from_path(filename);
```

soliton- (approved):

Issue #6988, opened 2022-08-28
Sometimes transparent UI elements are not drawn correctly 


grz0:
### Game and System Information

The Battle for Wesnoth version 1.17.7+dev (5155a557-Clean) x86_64
Running on Freedesktop.org 21.08.15 (Flatpak runtime) x86_64
Distribution channel: Default

### Description of the bug

Sometimes transparent UI elements are not drawn correctly and contain content of previously played map:
![image](https://user-images.githubusercontent.com/3605541/187075898-16197825-5013-4d53-a512-f60ad525e0ef.png)

### Steps to reproduce the behavior

1. Load some map, then quit to menu.

This issue happens very rarely and I don't know how to consistently reproduce it. So far, I've seen this three times.

Pentarctagon:
@mesilliac There was a really similar bug that go fixed previously, I think?

grz0:
> @mesilliac There was a really similar bug that go fixed previously, I think?

Do you mean #6836?

Wedge009:
Hmm, must be a specific set of circumstances, I haven't seen this yet. Still on 314f6f7c78d5fe1588e34acf0a73463197654b7f at the moment.

mesilliac:
It renders the background blur by taking a snapshot of the screen before drawing the panel widget. So for that to work it needs to render the title screen background once before drawing the UI panels on top. There must be some corner case in which this doesn't happen.

Wedge009:
@gfgtdf and others who experience this. Might be good to get some details on OS, graphics hardware, etc, because I still haven't encountered this.

gfgtdf:
I got this error on  an ubuntu 22.04, I actually don't know what graphics device it has, System information says "Unknown", it's a Dell Optiplex 9020.

grz0:
In my case, Fedora 35/36, integrated intel GPU

Wedge009:
> System information says "Unknown", it's a Dell Optiplex 9020.

https://www.dell.com/en-au/shop/desktops-workstations-all-in-ones/optiplex-9020-business-class-desktop/spd/optiplex-9020-desktop suggests it may also be using Intel iGPU (I can't imagine a 'business' machine using discrete GPUs.

gfgtdf:
Its lists as "Xeon E3-1200 v3/4th Gen Core Processor Integrated Graphics Controller" although i don't have an xeon cpu.

Wedge009:
[...]

Issue #7615, opened 2023-05-09
Opening windows or menus increases CPU usage


Pentarctagon:
This was reported on Discord a few days ago where a player was having a significant performance issue when opening the recruit dialog (or other windows/menus), to the point where the reported FPS dropped to only 5. Additionally:
* They're using Windows, with a GeForce GTX 1660 Super GPU and a "3 GHz i7" CPU. I can reproduce the issue of increased CPU usage on Linux, however my CPU is apparently powerful enough that it doesn't affect my FPS.
* This doesn't happen on 1.16 and only started happening on 1.17 with 1.17.6. 1.17.6 is when hardware accelerated rendering was first enabled.
* GPU usage doesn't increase when this is happening, the fps just drops drastically.
* Disabling map and unit idle/standing animations fixes the issue.

Pentarctagon:
@mesilliac if you're still around at all, do you have any idea what could be causing this?

mesilliac:
Something is probably marking an area for redraw every frame. If it doesn't happen with animations disabled i would guess somehow the animation code is calling for a redraw, but animations are paused so it does it every frame without changing anything. That's just a wild guess however, as i never really touched the animation code.

gfgtdf:
> but animations are paused

Actually they aren't (just tested wit the load game menu, standing animations still show), and i think this part is an improvement.

gfgtdf:
I tested this, and the amount of cup increase does depend on the type of dialog:

No dialog: 1.5%
Chatlog : 6.5%
LoadGame: 30%
Statistics: 19%
StatusTable: 18.5%


So it seems in particular that the amount of cpu increase is not really proportinal to the window size (chatlog is bigger than Statistics) but to the complexity of the dialog (chatlog is a really  simple dialog.)

I can also confirm that this only happens when the dialog covers something animated (in particular water)

gfgtdf:
I tested further with the load game dialog and i can for example also confirm that there is a small but consitent increase in cpu (26 to 30) usage whenever a savefile that shows a map is generated (as opposed to a start-of-scenario savefile that doesn't show a map).

gfgtdf:
So if understood correctly there are 3 options for fix this issue:
1) Disable map animations while dialogs show
2) Disable map animations under the dialog only.
3) Optimize the gui2 drawing code further.

(1) Is probably the easiest to do, but (3) also has the chance to be quite easy if it turns out that there is some obvious bottleneck thats easy to fix.

[...]

PR #7982, opened 2023-10-24
Offscreen window rendering

mesilliac:
I haven't done much (or any, really) testing on this, but it should solve #7615 

I verified that it does reduce CPU usage when opening the Load Game screen while in the tutorial.

I'm not sure if there might be rendering bugs in other areas. But the basic system is straightforward. It just renders windows to offscreen render textures when they change, and doesn't do any more work unless it needs to. When the window needs to be redrawn, it just blits from this texture without actually redrawing anything.

Renders should be minimal, and it's using all the same draw code so actually very little has changed.

### DIFF
--- a/data/gui/widget/panel_story_viewer.cfg
+++ b/data/gui/widget/panel_story_viewer.cfg
@@ -20,10 +20,7 @@
 
 			[draw]
 
-				# This works fine for normal messages. However, messages with options behave
-				# strangely, and messages with scrollbars perform badly.
-				# Uncomment once this has been fixed.
-				#{DEFAULT_BLUR}
+				{DEFAULT_BLUR}
 
 				[image]
 					x = 0
--- a/data/gui/widget/panel_wml_message.cfg
+++ b/data/gui/widget/panel_wml_message.cfg
@@ -20,10 +20,7 @@
 
 			[draw]
 
-				# This works fine for normal messages. However, messages with options behave
-				# strangely, and messages with scrollbars perform badly.
-				# Uncomment once this has been fixed.
-				#{DEFAULT_BLUR}
+				{DEFAULT_BLUR}
 
 				[image]
 					x = 0
--- a/src/draw.cpp
+++ b/src/draw.cpp
@@ -577,7 +577,7 @@ draw::render_target_setter::render_target_setter(const texture& t)
 	, viewport_()
 {
 	// Validate we can render to this texture.
-	assert(t.get_access() == SDL_TEXTUREACCESS_TARGET);
+	assert(!t || t.get_access() == SDL_TEXTUREACCESS_TARGET);
 
 	if (!renderer()) {
 		WRN_D << "can't set render target with null renderer";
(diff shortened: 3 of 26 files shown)


Hejnewar:
Might be totally unrelated but I guess it's fine to ask, is there any hope for offscreen rendering of terrain / unit animations? (Aka if unit / tile is offscreen its animation still shows on the on-screen part of the screen.) I believe there is issue opened for that as well.

mesilliac:
I remember looking into that but it was a bit outside anything i touched. There shouldn't be a problem with it theoretically, but i'm not really sure how the animations etc work. It might need special handling.

Pentarctagon:
Thanks for doing this - it definitely significantly decreases CPU usage for me. The only side effect I noticed when clicking around through various areas was that the Objectives popup is now completely opaque rather than translucent.

Also, the profiling results (Load Game dialog) for comparison:
perf-local

mesilliac:
Hmm it might have broken the blur effect. But i didn't notice a problem with the main menu blur, so maybe just in some cases...

gfgtdf:
Maybe just the area (that is specified by awaiting_rerender_) before redrawing on it

mesilliac:
Yeah it is erroneously grabbing from the render texture. But when i fixed it it broke the title screen panel blur. Looks like it will take some better handling to fix the blur properly.

That better handling would probably also fix #6988 so i'll look into it a bit more later.

mesilliac:
There are also some notes in `data/gui/widgets/panel_wml_message.cfg` and `panel_story_viewer.cfg` and they have disabled blur for some things... i'm not sure what that's about but maybe this would also fix whatever that issue was.

Git blame points out f5fbb3872a4d58e85029ad703ad733f93b652947 which looks very old.

mesilliac:
I implemented a solution which defers rendering of the blur effect for one frame. This should fix #6988 as well as #7615.

In addition, i reenabled the blur effect for dialogue messages, and in the other place it was disabled. I don't know what the original problem was with these blurs, but as they should be cached now it should probably be gone...

All the blurs i could find when briefly testing worked, and CPU usage remained pretty constant everywhere.

mesilliac:
I tested by playing all the way thru both tutorial missions and everything seems to work fine. If anyone knows what message windows have options or scrollbars (as mentioned in the two `cfg` files referenced above) let me know, but in general this seems stable.

The only issue i noticed was unrelated... and i just opened #7989 for it.

Pentarctagon:
I don't know of any in mainline off the top of my head. Adding it to a scenario temporarily would be pretty easy though:
```
    [event]
        name=start
        [message]
            [option]
                message=hi
            [/option]
        [/message]
    [/event]
```
and then just keep copy/pasting more options until you get a scrollbar.

mesilliac:
Seems to work fine with options and scrollbar. Verified that it only renders the blur once.

Pentarctagon:
Well, I'm good with this getting merged, at least.

Wedge009:
> I implemented a solution which defers rendering of the blur effect for one frame. This should fix #6988...

Looks like my Carcyn save is still reliably replicating #6988 so I can check this when it's merged. Nice to see you back, even if only for a bit.

mesilliac:
@Wedge009 cheers, let me know if it doesn't fix it and i'll see if there's something i can tweak.

Wedge009:
Confirmed at e24f60fd432 this appears to be resolved - at the very least my save that reliably replicated the issue is not producing it now. Since it was tricky to replicate, there's no guarantee some other scenario won't produce it but at least it's fixed for what I could reliably replicate.

PR #7983, opened 2023-10-24
[Units] update falcon and raven races to not get names

knyghtmare:
### Context:
- feral wolves do not get names (but wolf riders do, and that's supposed to be true)
- Ants do not get names
- scorpion do not get names
- bears do not get names
- bats do not get names
- Roc does not get name
- falcons and ravens get names which is unusual
- horses do not get names

### Expected behaviour
- Falcons and Ravens should not get names.

This pull request does this.

### DIFF
--- a/data/core/units.cfg
+++ b/data/core/units.cfg
@@ -141,7 +141,6 @@ In times of strife, the hierarchy of command becomes more adaptable, with the or
         num_traits=2
         undead_variation=falcon
         markov_chain_size=2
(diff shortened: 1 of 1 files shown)


knyghtmare:
Merging within 12 hours, unless anybody has any opinion to state.

Hejnewar:
The real difference is that they get traits but for me its like whatever. Just mentioning it.

PR #7987, opened 2023-10-25
Fix include_file= not being able to handle cfg files generated by the editor

gfgtdf:
including scenarios generated by the editor was the main usecase for this

### DIFF
--- a/src/saved_game.cpp
+++ b/src/saved_game.cpp
@@ -454,8 +454,9 @@ void saved_game::expand_mp_options()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


gfgtdf:
Hmm I don't have access to the wiki right now, will do it in around a week, if any else wants to do it before: it should be enough to remove the "prior to version 1.17.19" part.

soliton-:
That's definitely not enough. "the contents of this file is merged into the scenario tag" needs to be expanded to explain the new logic.

PR #7989, opened 2023-10-26
Fix UI update when changing pixel scale in preferences

mesilliac:
This just brings the custom resize event that can be fired in line with the change made in ac96df07bebbfc5d494f7325b89aa42ad4e567a9 .

The only thing that uses this event is the pixel scale slider which fires it manually when changed, as the window size doesn't actually change but the UI still needs to be regenerated.

This fixes the display updating incorrectly when changing pixel scale in prefs. (which i haven't reported because i just fixed it in stead)

### DIFF
--- a/src/events.cpp
+++ b/src/events.cpp
@@ -761,7 +761,7 @@ void raise_resize_event()
 	point size = video::window_size();
 	SDL_Event event;
 	event.window.type = SDL_WINDOWEVENT;
-	event.window.event = SDL_WINDOWEVENT_RESIZED;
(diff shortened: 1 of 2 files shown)


soliton-:
Yeah, you can ignore that. Building the windows dependencies unfortunately fails randomly and currently the cache got invalidated by a github runner update. So builds can fail randomly until the cache is up-to-date again.

Wedge009:
This makes sense to me and the change is an improvement.

Wedge009 (approved):

PR #7991, opened 2023-10-27
show error message if wml author don'tuse booelan value on bool filter call bool_matches_if_present

newfrenchy83:
if filter[attribute] has value other what boolean, it can't doing checking and should return true like if !filter.has_attribute(attribute).

name of function and bool_or_empty() changed.

### DIFF
--- a/src/utils/config_filters.cpp
+++ b/src/utils/config_filters.cpp
@@ -15,6 +15,8 @@
 #include <algorithm>
 #include <vector>
 
+#include "gettext.hpp"
+#include "gui/dialogs/message.hpp"      // for show_error_message
 #include "utils/config_filters.hpp"
 
 #include "serialization/string_utils.hpp" // for utils::split
@@ -24,6 +26,10 @@ bool utils::config_filters::bool_matches_if_present(const config& filter, const 
 {
 	if(!filter.has_attribute(attribute)) {
 		return true;
+	}
+	//check if filter is boolean, if not then show error message.
+	if(filter[attribute].to_bool(true) != filter[attribute].to_bool(false)) {
+		gui2::show_error_message(_("The filter ") + attribute +  _(" can only take the values yes/true or no/false."));
 	}
 
 	return filter[attribute].to_bool() == cfg[attribute].to_bool(def);


src/utils/config_filters.cpp
```diff
@@ -20,11 +20,16 @@
 #include "serialization/string_utils.hpp" // for utils::split
 #include "utils/math.hpp"                 // for in_ranges
 
-bool utils::config_filters::bool_matches_if_present(const config& filter, const config& cfg, const std::string& attribute, bool def)
+bool utils::config_filters::bool_matches_if_present_and_valid(const config& filter, const config& cfg, const std::string& attribute, bool def)
 {
 	if(!filter.has_attribute(attribute)) {
 		return true;
 	}
+	//check if filter is boolean, if not then checking can't be done and return true.
```
soliton-, line 28: What is the goal? This comment (or the code below) makes no sense to me.

Show a specific example of an ability and a filter that this is useful for.

CelticMinstrel:
Why doesn't `config_attribute_value` have `is_xxx` functions to detect the type? Well, whatever… either way this check seems overly complicated. Wouldn't this detect the same thing with less code?
```
if(filter[attribute].to_bool(true) != filter[attribute].to_bool(false)) {
    return true;
}
```

Since `to_bool` returns its argument if the value is not a boolean, `to_bool(true)` should be equal to `to_bool(false)` if and only if the value is a boolean.

newfrenchy83:
> Why doesn't `config_attribute_value` have `is_xxx` functions to detect the type? Well, whatever… either way this check seems overly complicated. Wouldn't this detect the same thing with less code?
> 
> ```
> if(filter[attribute].to_bool(true) != filter[attribute].to_bool(false)) {
>     return true;
> }
> ```
> 
> Since `to_bool` returns its argument if the value is not a boolean, `to_bool(true)` should be equal to `to_bool(false)` if and only if the value is a boolean.
> 
> Additionally, splitting on commas and checking for the presence of true/false strings doesn't even make any sense if the goal is to check that the value is a boolean.

fixed

src/utils/config_filters.hpp
```diff
@@ -67,6 +67,6 @@ bool int_matches_if_present_or_negative(
 bool string_matches_if_present(
 	const config& filter, const config& cfg, const std::string& attribute, const std::string& def);
 
-bool bool_or_empty(const config& filter, const config& cfg, const std::string& attribute);
+bool matches_if_attribute_boolean_or_missing(const config& filter, const config& cfg, const std::string& attribute);
```
soliton-, line 70: Remove this part from the PR. It is not in scope.

If you must open a different PR for it but I'd rather you didn't.

src/utils/config_filters.cpp
newfrenchy83: if someone want check affect_self attribute who is boolean with value true by default but write filtet like that:
`affect_self=myself` how sustem will understand that, check if false then what develloper search another value, or then like here declare filter invalid and don't check attribute.

soliton-: Why would this filter be "understandable"? Where is this documented?

src/utils/config_filters.hpp
newfrenchy83: removed

src/utils/config_filters.cpp
newfrenchy83: I believe this is the first time that a boolean attribute has been checked by a filter, so it seems normal to me to ensure that the filter used uses true/yes or false/no and not other things (the attribute affect_allies is an exception because its absence has a meaning other than true/false)

soliton-: You do not ensure only valid values by simply pretending anything else is not there. That's terrible. Raise an error and let the WML author fix it.

newfrenchy83: ok,, i done that.

soliton-:
Semantically that looks good to me. Implementation wise perhaps this should rather throw and display the error elsewhere. Also it seems like you expect `$attribute` to be expanded. I'm not going to look into that though.

newfrenchy83:
> Semantically that looks good to me. Implementation wise perhaps this should rather throw and display the error elsewhere. Also it seems like you expect `$attribute` to be expanded. I'm not going to look into that though.

function called for affect_self, affect_allies and cumulative, it is wise to said what filter is erroned.

src/utils/config_filters.cpp
```diff
@@ -25,6 +27,10 @@ bool utils::config_filters::bool_matches_if_present(const config& filter, const
 	if(!filter.has_attribute(attribute)) {
 		return true;
 	}
+	//check if filter is boolean, if not then show error message.
+	if(filter[attribute].to_bool(true) != filter[attribute].to_bool(false)) {
+		gui2::show_error_message(_("The filter $attribute can only take the values yes/true or no/false."));
```
Pentarctagon, line 32: Why would displaying `$attribute` in the error message be useful?

newfrenchy83: excuse me i would doing `show_error_message(_("The filter ") + attribute +  `

newfrenchy83: @soliton Yes, I was wrong, no need to lower your thumb.

Pentarctagon: That's also incorrect - for an example see https://github.com/wesnoth/wesnoth/blob/6b9a7627b90f90c0fc4f2065d19b179df4682280/src/chat_events.cpp#L57-L60

newfrenchy83: itry but that don't work

newfrenchy83: @Pentarctagon i pull that: `gui2::show_error_message(_("The filter ") + attribute +  _(" can only take the values yes/true or no/false, the check will therefore not be performed on this filter."));`

the method what you're suggest don't work.

Pentarctagon: `gui2::show_error_message(_("The filter ") + attribute +  _(" can only take the values yes/true or no/false, the check will therefore not be performed on this filter."));` is not acceptable since that creates to separate translatable strings.

Pentarctagon:
@newfrenchy83 as a general statement that might help make these reviews easier for everyone: please don't make code changes in response to a question being asked. Just answer the question, we'll discuss it, and then we'll tell you explicitly what should be done.

Issue #7992, opened 2023-10-27
[filter_ability] issues


soliton-:
https://github.com/wesnoth/wesnoth/blob/6b9a7627b90f90c0fc4f2065d19b179df4682280/src/utils/config_filters.cpp#L49-L51

Checks like this seem to be intended to test if an ability supports the attribute or not. This cannot be checked in these `*_matches_if_present` helper functions though since they have no idea what ability type they're checking.

For `value` there are some `tag_name` tests but `multiply` for example is checked regardless of what the `tag_name` is. Really the tests for `value` look wrong anyway since they also look at `cfg` as if it tells you what an ability supports.

So the logic in https://github.com/wesnoth/wesnoth/blob/6b9a7627b90f90c0fc4f2065d19b179df4682280/src/units/unit.cpp#L1424
must make sure to only call the `*_matches_if_present` helper functions on attributes the given `tag_name` actually supports. When checking an unsupported attribute the filter must fail to match.

Also all code that checks the given `cfg` for what an ability type supports must be removed since it is simply wrong.

stevecotton:
What should be the responsibility of the engine, and what should be the responsibility of the code using the engine?

I feel that it's the responsibility of the WML author to only filter for attributes that are supported by the ability that they're looking for.

newfrenchy83:
> What should be the responsibility of the engine, and what should be the responsibility of the code using the engine?
> 
> I feel that it's the responsibility of the WML author to only filter for attributes that are supported by the ability that they're looking for.

Sorry, but I have the distinct impression that my repairs will cause more problems than anything else; and if an author used an ability like tailwind with value or add, the repair will exclude value filtering for this ability because it does not have the right tags. Indeed the problem raised here cannot be solved in engine.

Pentarctagon:
Is this a valid issue then?

soliton-:
> What should be the responsibility of the engine, and what should be the responsibility of the code using the engine?
> 
> I feel that it's the responsibility of the WML author to only filter for attributes that are supported by the ability that they're looking for.

Btw, what should happen if you filter for a heals ability and the attribute `value = 0`? Does ABILITY_UNPOISON match?
What if you filter for drains and `value=50`, should WEAPON_SPECIAL_DRAIN when used as an ability match?
Worse, what should happen if you filter for `multiply = 1` on any ability that does not explicitly set multiply but does use value?

[...]

PR #7993, opened 2023-10-28
fix [filter_ability] issue

newfrenchy83:
Fix https://github.com/wesnoth/wesnoth/issues/7992

### DIFF
--- a/src/utils/config_filters.hpp
+++ b/src/utils/config_filters.hpp
@@ -15,6 +15,7 @@
 #pragma once
 
 #include "config.hpp"
+#include <set>
 
 /**
  * Utility functions for implementing [filter], [filter_ability], [filter_weapon], etc.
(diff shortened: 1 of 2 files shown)


src/utils/config_filters.hpp
```diff
@@ -69,4 +72,11 @@ bool string_matches_if_present(
 
 bool bool_or_empty(const config& filter, const config& cfg, const std::string& attribute);
 
+/**
+ * vector used for verify what attribute checked can be present in special/abilities.
+ */
+const std::set<std::string> abilities_tags{"tailwind", "hides", "skirmisher", "teleport", "leadership", "illuminates", "heals", "regenerate"};
```
stevecotton, line 78: The whole PR makes the logic feel more convoluted, more places where specific abilities are given odd edge cases.

stevecotton, line 78: Needing to put a campaign-specific ability (`tailwind`) in this list suggests that it shouldn't be engine-level.

newfrenchy83: i will perhaps close this PR then because between weapon andabilities not hard enginered, it si became impossible to sais if 'value' can be codedor not. I will await to see end checking.

PR #7995, opened 2023-10-28
Allow filter also apply_to in [resistance]

newfrenchy83:
In engine [resistance] is the only ability that also uses the apply_to attribute, but in a different way than for special weapons so I propose that in this case the filter applpy_to does not have a default value.

It seems unlikely to me that anyone will use this attribute in a custom ability (dummy, tailwind etc.) but if it is considered possible then I will abandon the idea

### DIFF
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -1451,8 +1451,12 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 	if(!string_matches_if_present(filter, cfg, "id", ""))
 		return false;
 
-	if(!string_matches_if_present(filter, cfg, "apply_to", "self"))
-		return false;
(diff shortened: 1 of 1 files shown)


stevecotton:
When would this have an effect? IIUC, it's only in the case that the filter WML includes the following, which seems like an error in the filter WML.
```
tag_name=resistance (or possibly a comma-separately list including resistance)
apply_to=self
```

newfrenchy83:
> When would this have an effect? IIUC, it's only in the case that the filter WML includes the following, which seems like an error in the filter WML.
> 
> ```
> tag_name=resistance (or possibly a comma-separately list including resistance)
> apply_to=self
> ```

i wiil not insist and let code like it is.

PR #7996, opened 2023-10-28
Multipass rendering

mesilliac:
I didn't like that the blur effect required two frames to render, so i implemented basic multipass rendering. That is, while rendering, anything can request that another rendering pass is performed before flipping to the screen. This means the blur can render the thing behind it, blur that, then render itself, all in the same frame.

Theoretically this would also work for multiple layered blurs, rendering in three or more passes. But try not to do that as it's expensive, and will probably drop frames on weak hardware.

I also noticed while testing that the story screen text panels now have blur, and it wasn't updating, so i forced it to regenerate the blur at each step. Without the multipass rendering, this causes a one-frame flicker which was pretty much unacceptable. It looks pretty good now.

### DIFF
--- a/src/draw_manager.cpp
+++ b/src/draw_manager.cpp
@@ -48,6 +48,7 @@ std::vector<rect> invalidated_regions_;
 bool drawing_ = false;
 bool tlds_need_tidying_ = false;
 uint32_t last_sparkle_ = 0;
+bool extra_pass_requested_ = false;
 } // namespace
 
 namespace draw_manager {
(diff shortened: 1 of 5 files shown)


CelticMinstrel:
Does this fix the blur on the title screen reading from the map you just played instead of from the title screen?

mesilliac:
That one should be fixed already by the commits i pushed to master earlier, so yes. This improves on the fix by not requiring an extra frame to do it.

Wedge009:
That was #6988 and at least for me it seemed to be resolved.

mesilliac:
As a side note, this basically completes the core graphics pipeline. A hardware blur would now be a drop-in replacement with shaders.

PR #7997, opened 2023-10-28
Start game with theme buttons hidden.

### LINKED ISSUE #6922: Game elements showing prematurely in story screen

- **Version:** Current master (6701542f633990ec14f9302d9e998062ad13d749)
- **Downloaded from:** Self-compiled
- **Build info:** SCons and MSVC
- **OS:** Linux/X11 and Windows

### Description of the bug

When making the transition from story screen to the main game screen, the top-left menu buttons and the mini-map buttons show prematurely. This happens whether using the skip button in the bottom-right or stepping through the story screens one-by-one.

Illustration:

1. I was just testing Northern Rebirth but I imagine this happens on all campaigns - start a campaign.
2. Step through story screens.
3. Observe the game elements showing prematurely in the transition from story to game.

### Expected behavior

Not a major issue, but ideally these elements shouldn't show until the story screen is gone.

### Additional context

_No response_

mesilliac:
Fixes #6922

Just hides the buttons until the game has started. There was already a function to cause the game map not to draw while the story screen was showing, so i just used that as the cue for unhiding the buttons.

There is some weirdness with the buttons under the minimap appearing at different times. But that seems to be a separate issue, and unrelated.

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -925,6 +925,11 @@ void display::create_buttons()
 		action_buttons_.push_back(std::move(b));
 	}
 
+	if (prevent_draw_) {
+		// buttons start hidden in this case
+		hide_buttons();
+	}
+
 	layout_buttons();
 	DBG_DP << "buttons created";
 }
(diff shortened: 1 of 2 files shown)


src/display.cpp
```diff
@@ -2211,6 +2236,21 @@ double display::turbo_speed() const
 		return 1.0;
 }
 
+void display::set_prevent_draw(bool pd)
+{
+	prevent_draw_ = pd;
+	if (!pd) {
+		// ensure buttons are visible
+		unhide_buttons();
+	}
```
gfgtdf, line 2245: Shouldn't the else branch then call`hide_buttons()`?

Wedge009:
I don't know about the `hide_buttons()` case, but the change as-is seems to resolve it for me. I just checked the intro of HttT.

src/display.cpp
mesilliac: At that point the buttons don't exist yet. I considered making it hide, but it might actually cause strange behaviour with disappearing buttons if someone uses it mid-game for some reason.

Are there any scenarios that trigger the story-viewer mid-scenario? I'm not sure how it would behave in that case.

mesilliac: I remembered there was an event in DiD that did. In this case they don't need to hide as the story screen correctly covers the buttons, it's only when the buttons are initially created that they jump on top of everything else.

PR #7998, opened 2023-10-28
add some logging to game history query

Pentarctagon:
---

Not sure if this makes sense to keep after the performance issue is resolved or not, but the goal is:
* log the actual SQL statement being run, to make sure it matches what I think it should.
* log between the individual steps so it's clear which is taking up too much time.


soliton-:
IMO this is definitely useful. Have to think about whether it should be info or debug level though.

soliton-:
Since we already have info level output for querying the game history I think all the added logging here should be debug level that we turn on if we want to investigate a specific issue.

Pentarctagon:
Updated.

soliton- (approved):

PR #7999, opened 2023-10-29
1.18 - Some loyalist animation updates

doofus-01:
Whatever I can get done this weekend.  No changes to anything but animations.


soliton-:
Just wondering why this appears to have been squashed. All commits seem perfectly fine on their own to me. I recommend using the rebase option in the future.

doofus-01:
No real reason.  Force of habit, I guess.

PR #8005, opened 2023-10-31
Update CONTRIBUTING.md

vatsalya-vyas:
# Fixed a typographical error in the repository

Revised the sentence for enhanced clarity and resolved a grammar issue by changing:

 should -> should use it


stevecotton:
The original seems better to me, it's the idiom "just because you can, doesn't mean you should".

CelticMinstrel:
The original sentence is using [ellipsis](https://en.wikipedia.org/wiki/Ellipsis_(linguistics)), which is a perfectly valid grammatical construction where a portion of the sentence is omitted because it can be understood from what remains.

I don't have any opinion on whether the original or revised sentences are better. The only thing I'd say is, if we're going to accept the choice not to use ellipsis, then only the word "should" should be emphasized, not the entire phrase "should use it".

Pentarctagon:
I'd also say the current wording is fine as-is.

Wedge009:
> The original seems better to me, it's the idiom "just because you can, doesn't mean you should".

As a native Anglophone, I understand the original as it stands.

> ...then only the word "should" should be emphasized, not the entire phrase "should use it".

Agreed on the emphasis change, in the case that the phrasing change is accepted..

stevecotton:
Judging by the +1 responses from @vatsalya-vyas, I'm closing this.

PR #8007, opened 2023-10-31
UtBS: Units with tailwind have green orbs and rest-heal

### LINKED ISSUE #7966: UtBS's tailwind buff prevents rest-healing, and causes orange orb

- **Version:** 1.17.22+dev 0945c30efd2cafff9f36ca328bcd5ffc591e6be9
- **Downloaded from:** built from Git

### Description of the bug

In UtBS, at the start of the player's turn any units that are next to a Quenoth Shaman get +1 move (or +2 for some of the L2/L3 shamans). The buffed units have an orange orb instead of the green unmoved orb. They also can't get the +2 hp bonus on the following turn for rest-healing.

If I didn't know what was causing it, I'd instead describe the bug like this: in UtBS S01, any unit that starts on the Great Tree's island gets an orange orb.

1. Play UtBS
2. In S01, get to the island or just debug-spawn a shaman
3. End turn

### Expected behavior

@Hejnewar Should they be able to rest-heal?

### Additional context

_No response_

stevecotton:
Fixes #7966

In UtBS, at the start of the player's turn any units that are next to a Quenoth Shaman get +1 move (or +2 for some of the L2/L3 shamans).

Without this fix, the buffed units have an orange orb instead of the green unmoved orb, and don't rest-heal next turn. This makes things look correct during the player's turn, at the expense of showing the unit's max_moves as the buffed amount, even if the unit won't get the buff next turn.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
@@ -587,6 +587,29 @@ Any units adjacent to this unit will fight as if it were dusk when it is night, 
                     find_in=utbs_tailwind2_sources
                 [/filter_adjacent]
             [/filter]
+            # Temporarily increase the max moves correspondingly, so the unit gets a green orb and
+            # can rest heal next turn, because units only rest heal if their moves and max moves are
+            # exactly equal at the start of the next turn.
+            #
+            # Without this object units only rest heal if the player moves them to reduce their
+            # movement down to match the maximum.
+            #
+            # Using `duration=turn` means that the wrong max moves will show during the AI turn.
+            # Rest healing mechanics work as normal - units rest heal if (and only if) they aren't
+            # moved the previous turn.
+            #
+            # Using `duration=turn end` allows the unit to move 1 or 2 hexes and still rest heal.
+            # When the object is removed at turn end, the moves remaining will be capped to the max,
+            # so rest healing works.
+            #
+            # Remember to change the tailwind+1 object to match this one.
+            [object]
+                duration=turn end
+                [effect]
+                    apply_to=movement
+                    increase=2
+                [/effect]
+            [/object]
             moves="$($this_unit.moves + 2)"
         [/modify_unit]
         [modify_unit]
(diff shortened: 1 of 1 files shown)


Hejnewar (approved):

data/campaigns/Under_the_Burning_Suns/utils/abilities.cfg
```diff
@@ -587,6 +587,17 @@ Any units adjacent to this unit will fight as if it were dusk when it is night,
                     find_in=utbs_tailwind2_sources
                 [/filter_adjacent]
             [/filter]
+            # Temporarily increase the max moves correspondingly, so the unit gets a green orb and
+            # rest-heals next turn. Using `duration=turn` means that the wrong max moves will show
+            # during the AI turn, but using `duration=turn end` allows the unit to move 1 or 2 hexes
+            # and still rest-heal.
```
soliton-, line 593: Is the idea here to not allow moving 1 or 2 hexes and still rest heal? Is this a workaround or the final fix?

To me allowing rest heal even if using the bonus moves seems more sensible.

stevecotton: @Hejnewar I think this is a question for you, about what you want the campaign to do.

My feeling is that the 2 extra hitpoints are for avoiding putting strain on the wounds by not moving at all. Okay, so it's also given if the unit defended themselves, but that's WINR.

Hejnewar: On one hand I really like the green orb. On the other hand it makes it harder to see which units got buffed. On the third hand, future proofing the game is good, unless we agree that abilities that increase movement should always also increase max moves as well but we can't enforce that in add-ons. I'm generally fine with either take, it is small enough difference that it's fine.  

When it comes to lore then for this particular unit it already heals as well so we can actually probably just assume that mystic fixed the wounds enough that nothing will happen if the unit moves a bit, but it only works in this case.

stevecotton: Latest push has changed it so that rest healing still happens even after using the bonus moves.

Issue #6959, opened 2022-08-12
Small black artifacts when map is zoomed in/out


grz0:
### Game and System Information

The Battle for Wesnoth version 1.17.6+dev (565b318c9f0-Modified) x86_64
Running on Fedora Linux 36 (Workstation Edition) x86_64
Distribution channel: Default

SDL video drivers:        wayland [x11] KMSDRM dummy
Window size:              1920x1021
Game canvas size:         1920x1021
Final render target size: 1920x1021
Screen refresh rate:      60
Boost:             1.76
Lua:               5.4.4
OpenSSL/libcrypto: 3.0.0e-dev (runtime 3.0.0e-dev)
Cairo:             1.17.6 (runtime 1.17.6)
Pango:             1.50.8 (runtime 1.50.8)
SDL:               2.0.22 (runtime 2.0.22)
SDL_image:         2.0.5 (runtime 2.0.5)
SDL_mixer:         2.0.4 (runtime 2.0.4)

### Description of the bug

This happens only for some zoom levels:
![image](https://user-images.githubusercontent.com/3605541/184409095-82a78d27-847a-42ed-a3b5-0c6286ef0fa8.png)
![image](https://user-images.githubusercontent.com/3605541/184409150-fbb90f22-2db8-49ff-8911-9b59e335873e.png)

### Steps to reproduce the behavior

1. Zoom in/out until you see black artifacts

Wedge009:
The 'artefacts' looks like a slight misalignment of the hex images at certain zoom levels. I observe the issue in current master (314f6f7c78d), but not in 1.16 branch.

Wedge009:
A git bisect suggests this regression started in 46750666b25704b97766bc691a199dee111b56f6, introduced with 1.17.5.

gfgtdf:
Happens to me too, the affected zoom levels are 0.33333333, 0.5 and 1.5

gfgtdf:
M first idea here would be to first assemble the map from it tiles, an then zoom the assembed map instead of zooming the tiles seperately and then placing them on the map, but i have no idea how easy that'd be.

Wedge009:
At an abstract level, likewise without knowing the technical details of the implementation, that does sound like a good approach.

@mesilliac Sorry to bother you, but noticed a few comments from you recently. Is that an accurate representation of what actually happens, that the tiles are being scaled separately?

mesilliac:
[...]

PR #8014, opened 2023-11-03
Fix overlapping fog texture at some zoom levels

mesilliac:
Zoom scales, which are described in draw-space pixels, were always required to be multiples of 4. This was violated at some point, resulting in graphical glitches when drawing fog at some zoom levels.

This both checks that zoom scales are divisible by four, and changes the zoom level specification to integers rather than multiplication factors so that they can be specified accurately.

This *also* fixes an auxiliary problem that the 1/3x zoom level specified as "0.33333333333333333" was incorrectly rounding to 23 in stead of 24, resulting in more graphical glitches.

This is related to #6959 but won't fix it entirely as that's a separate problem. The 1/3x zoom level is likely improved however.

These are the types of glitches fixed by this change. They are only obvious in fog, as that applies twice when hexes overlap:
fog_glitches

### DIFF
--- a/data/game_config.cfg
+++ b/data/game_config.cfg
@@ -33,8 +33,10 @@
     hp_bar_scaling=0.6
     xp_bar_scaling=0.5
 
-    # zoom factors must be a sorted list of ascending multipliers
-    zoom_levels = 0.25, 0.33333333333333333, 0.5, 0.75, 1.0, 1.25, 1.5, 2.0, 3.0, 4.0
+    # zoom_levels must be an ascending list of hex sizes in pixels.
+    # hex sizes must be divisible by 4 to prevent graphical glitches.
+    # sizes are in draw space, so standard 1x zoom is always 72.
+    zoom_levels = 16, 24, 36, 52, 72, 100, 144, 216, 288
 
     flag_rgb=flag_green
     unit_rgb=magenta
--- a/data/game_config.cfg
+++ b/data/game_config.cfg
@@ -34,8 +34,11 @@
     xp_bar_scaling=0.5
 
     # zoom_levels must be an ascending list of hex sizes in pixels.
-    # hex sizes must be divisible by 4 to prevent graphical glitches.
-    # sizes are in draw space, so standard 1x zoom is always 72.
+    # Hex sizes must be divisible by 4 to prevent graphical glitches.
+    # 1x zoom is always 72, as defined by tile_size in game_config.cpp.
+    # Sizes are in draw-space, so will be scaled for high-dpi automatically.
+    # As given here these correspond as closely as possible to zoom factors of:
+    #   1/4, 1/3, 1/2, 1/sqrt(2), 1, sqrt(2), 2, 3, 4
     zoom_levels = 16, 24, 36, 52, 72, 100, 144, 216, 288
 
     flag_rgb=flag_green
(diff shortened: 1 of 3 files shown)


data/game_config.cfg
```diff
@@ -33,8 +33,10 @@
     hp_bar_scaling=0.6
     xp_bar_scaling=0.5
 
-    # zoom factors must be a sorted list of ascending multipliers
-    zoom_levels = 0.25, 0.33333333333333333, 0.5, 0.75, 1.0, 1.25, 1.5, 2.0, 3.0, 4.0
+    # zoom_levels must be an ascending list of hex sizes in pixels.
+    # hex sizes must be divisible by 4 to prevent graphical glitches.
+    # sizes are in draw space, so standard 1x zoom is always 72.
```
Wedge009, line 38: Maybe worth highlighting that `tile_size = 72` is defined in `game_config.cpp` and commenting these sizes correspond to 1/4, 1/3, 1/2, 3/4, 1, 1.25, 1.5, 2, 3, 4 times magnification.

Just so it's clear to someone coming back to this later. It may be obvious to us now but easy to forget things months/years down the track.

Wedge009:
@grz0, @gfgtdf, and anyone else - can you see if this makes any difference for you? I was surprised to see how much this helps on my hardware, considering when I tried `int(round(stold(zoom) * tile_size))` previously it didn't seem to make much difference. I only get obvious artefacts at 0.5x zoom with this change.

grz0:
> can you see if this makes any difference for you?

I just compiled master with this commit, and it fixes all the weird fog scaling glitches that have been in Wesnoth forever. Regarding #6959 - without the fix, I see these black pixels on three zoom levels, with the fix only one zoom level has this issue.

gfgtdf:
> @grz0, @gfgtdf, and anyone else - can you see if this makes any difference for you?

I won't be on my main PC for a while, so I cannot test currently.

data/game_config.cfg
mesilliac: Good idea, i'll add those in.

Looking at it made me realize i missed a zoom level... will bring that up in the main thread

mesilliac:
Looking at it again i realized i missed a zoom level... but i'm not sure if i should readd it, as the new values scale much more evenly.

They came out nicely scaling by a factor of sqrt(2) between each level, giving

    36 = 0.5x
    52 ≈ 1/sqrt(2) down from 54 = 0.75
    72 = 1x
    100 ≈ sqrt(2) between 90 = 1.25 and 108 = 1.5
    144 = 2x

All the zoom factors scale very evenly like this. But if there was a reason for having 1.25x and 1.5x separately i can add another level in... but it has to be either at 88 ≈ 1.22x or 92 ≈ 1.28x as 90 doesn't divide by 4.

Wedge009:
I didn't realise the numbers you put in don't correspond to the old calculated zoom levels, but rather was adhering to the principle that the figures are divisible by 4. That probably explains why I see an improvement with these values.

In that case, you can make the commentary explain that's what you were doing and that the new zoom values are only an approximation of the old zoom factors.

mesilliac:
Yeah i actually fitted even steps on an exponential curve while rounding to multiples of 4... it came out so close to the existing values i didn't even notice i missed one. Was just thinking it was odd that the 1.5x value was comparitively quite far off 😅

Wedge009 (approved):

PR #8017, opened 2023-11-04
touchup to direwolf defense frames

doofus-01:


soliton-:
I would appreciate it if squashing is combined with editing the commit message to be sensible. The github default is not really useful in my opinion.

doofus-01:
What else should have gone in there?  It was a couple small image touch-ups, and then dealing with the copyrights.csv CI.

soliton-:
I expected less not more. Pretty much just the commit message of the first commit.

 The github default will use the PR title as the commit subject and then list all commit messages (or maybe just the subject) as the commit body. Often the PR title is the same as the first commits subject and if you want to squash it's probably because there are fixup or other commits in there that have no useful commit message. So the default is fine if you then edit it down to the useful bits but not really if you just leave it as is.

PR #8026, opened 2023-11-05
Update title screen widget blur when resizing the game window

### LINKED ISSUE #8023: Blurred elements in the title screen are not rendered correctly after resizing window

The Battle for Wesnoth version 1.17.22+dev (b8adf0cd-Clean) x86_64
Running on Freedesktop.org SDK 22.08 (Flatpak runtime) x86_64

Libraries
=========

Boost:             1.67
Lua:               5.4.6
OpenSSL/libcrypto: 3.0.0e-dev (runtime 3.0.0k-dev)
libcurl:           7.85.0 (runtime 7.86.0-DEV)
Pango:             1.50.9 (runtime 1.51.0)
SDL:               2.24.0 (runtime 2.24.0)
SDL_image:         2.6.2 (runtime 2.6.3)
SDL_mixer:         2.6.0 (runtime 2.6.3)

Current video settings
======================

SDL video drivers:        [x11] wayland dummy evdev
Window size:              800x540
Game canvas size:         800x540
Final render target size: 800x540
Screen refresh rate:      60
Screen dpi:               92.02x91.44
Renderer:                 opengl (hw)
Maximum texture size:     16384x16384
VSync:                    off

### Description of the bug

image

1. Resize the game window when on title screen the 

### Expected behavior

_No response_

### Additional context

_No response_

mesilliac:
This adds a check to regenerate blur whenever the screen region of the blur changes. This should catch most cases of moved and resized widgets.

It also explicitly regenerates the title screen blurs when the prefs dialog is closed. Otherwise the regenerated blurs will include the prefs dialog that's sitting on top.

Fixes #8023

### DIFF
--- a/src/gui/dialogs/title_screen.cpp
+++ b/src/gui/dialogs/title_screen.cpp
@@ -315,8 +315,24 @@ void title_screen::init_callbacks()
 	//
 	// Preferences
 	//
-	register_button(*this, "preferences", hotkey::HOTKEY_PREFERENCES, []() {
+	register_button(*this, "preferences", hotkey::HOTKEY_PREFERENCES, [this]() {
 		gui2::dialogs::preferences_dialog::display();
+
+		// Currently blurred windows don't capture well if there is something
+		// on top of them at the time of blur. Resizing the game window in
+		// preferences will cause the title screen tip and menu panels to
+		// capture the prefs dialog in their blur. This workaround simply
+		// forces them to re-capture the blur after the dialog closes.
+		panel* tip_panel = find_widget<panel>(this, "tip_panel", false, false);
+		if(tip_panel != nullptr) {
+			tip_panel->get_canvas(tip_panel->get_state()).queue_reblur();
+			tip_panel->queue_redraw();
+		}
+		panel* menu_panel = find_widget<panel>(this, "menu_panel", false, false);
+		if(menu_panel != nullptr) {
+			menu_panel->get_canvas(menu_panel->get_state()).queue_reblur();
+			menu_panel->queue_redraw();
+		}
 	});
 
 	//
--- a/data/gui/window/title_screen.cfg
+++ b/data/gui/window/title_screen.cfg
@@ -94,6 +94,7 @@ where
 
 #define _GUI_TIP_SECTION
 	[panel]
+		id = "tip_panel"
 		definition = "box_display"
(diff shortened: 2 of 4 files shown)


src/gui/dialogs/title_screen.cpp
```diff
@@ -315,8 +315,22 @@ void title_screen::init_callbacks()
...
 	// Preferences
 	//
-	register_button(*this, "preferences", hotkey::HOTKEY_PREFERENCES, []() {
+	register_button(*this, "preferences", hotkey::HOTKEY_PREFERENCES, [this]() {
 		gui2::dialogs::preferences_dialog::display();
+		// The screen may have changed size, so to be safe,
+		// ensure that blurred elements recalculate their blur.
+		// Better would be to update every styled_widget,
+		// but for now this just grabs the two we know have blur.
+		panel* tip_panel = find_widget<panel>(this, "tip_panel", false, false);
```
gfgtdf, line 324: The (wesnoth) window can also be resized without the preferences dialog (by dragging the window border with the mouse like any window) in that case this code is not called. So when the case that it changed through the preferences dialog needs special handling here, it's be useful to have a comment why that is.

mesilliac: You're right it's not very clear. I'll update the description there a bit

Pentarctagon:
Anything preventing this from being merged?

mesilliac:
Yeah i want to redo the info in the comments...

The solution isn't perfect. I slept on it a bit to see if a neat compromise magically popped up, but nope. It's just fundamentally the wrong thing doing the blurring so it will need to be tweaked in the future, probably with SDL3. So it'd be nice if i could make it clear what needs doing, i just haven't had the time.

src/gui/dialogs/title_screen.cpp
mesilliac: I made the description clearer about what it's doing in this case. If it turns out to be an issue that this doesn't cover resizing via dragging the window borders, this tweak may have to be adapted or moved elsewhere.

mesilliac:
I just updated the comment on the prefs dialog to be clearer about what it's doing. If it turns out to be a problem that this doesn't cover resizing via dragging window borders, that tweak can be adapted elswhere as a hack fix.

Issue #8009, opened 2023-10-31
UtBS: No documentation about the recall costs


stevecotton:
### Describe the desired feature

UtBS's equivalent of WoF's #8008

knyghtmare:
Nope. It was mentioned. Check:

https://github.com/wesnoth/wesnoth/blob/50a21a2c6c648b4ebcc2ab9fbfaed9b03f5d7f0f/data/campaigns/Under_the_Burning_Suns/utils/storytxt.cfg#L7C73-L7C73

```ini
            story= _ "Note: This campaign is probably not appropriate for beginners. It changes certain Wesnoth standards, such as the elves’ stats, the day/night cycle, recall costs and general approach to the difficulty levels. This campaign emphasizes role-playing elements and tends to have longer scenarios with objectives that change in the middle. For these reasons, we strongly suggest that you occasionally save your game mid-scenario, so you won’t lose all your progress if you get stuck and have to start over."
```

```txt
 cycle, recall costs and general
```

Hejnewar:
Turns out I just forgot i updated it. But yeah I kinda agree that this might not be visible enough but at least it is there.

Jonathan-Kelly:
It is a bit comical to only give players this information after they start the campaign and have no clear or quick way to return to the main or campaign selection menus. While not a new issue for UtBS, if now EI and WoF are adding similar warning messages then it might be worth considering whether to move this to the campaign description string instead.

doofus-01:
> It is a bit comical to only give players this information after they start the campaign and have no clear or quick way to return to the main or campaign selection menus. While not a new issue for UtBS, if now EI and WoF are adding similar warning messages then it might be worth considering whether to move this to the campaign description string instead.

I largely agree with this, but would go even further.  How many people play this, are experienced enough with Wesnoth to have certain expectations and not just roll with whatever, and will remember some message in passing or think to go back to campaign description?

If a campaign has non-standard mechanics, it should have a help dialog like World Conquest has.  This might be a larger issue (I'm thinking it might also apply to SoF runes?), and doesn't preclude updating the campaign descriptions, but it would help if there was a standard way to add custom help info.

PR #8027, opened 2023-11-05
UtBS: Add note on modified game mechanics to campaign description

Hejnewar:
Probably not good enough but good enough to start a discussion I hope.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/_main.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/_main.cfg
@@ -19,6 +19,8 @@
 
     description= _ "In the distant future a small band of elves struggles to survive amidst the ruins of fallen empires. Lead your people out of the desert on an epic journey to find a new home.
 
+" + _"Note: This campaign is probably not appropriate for beginners. It changes certain Wesnoth standards, such as the day/night cycle, recall costs and general approach to the difficulty levels, it is also making use of a custom faction of Quenoth elves as a replacement for standard elves.
+
 " + _"(Expert level, 10 scenarios.)"
 
     # Old difficulty levels are not selectable anymore, but for now loading saves using them should still work.
--- a/data/campaigns/Under_the_Burning_Suns/utils/storytxt.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/utils/storytxt.cfg
@@ -4,7 +4,7 @@
     [story]
         [part]
             # the note should be changed to better reflect the current state of this campaign and faction
-            story= _ "Note: This campaign is probably not appropriate for beginners. It changes certain Wesnoth standards, such as the elves’ stats, the day/night cycle, recall costs and general approach to the difficulty levels. This campaign emphasizes role-playing elements and tends to have longer scenarios with objectives that change in the middle. For these reasons, we strongly suggest that you occasionally save your game mid-scenario, so you won’t lose all your progress if you get stuck and have to start over."
+            story= _ "Note: This campaign is probably not appropriate for beginners. It changes certain Wesnoth standards, such as the day/night cycle, recall costs and general approach to the difficulty levels, it is also making use of a custom faction of Quenoth elves as a replacement for standard elves. This campaign emphasizes role-playing elements and tends to have longer scenarios with objectives that change in the middle. For these reasons, we strongly suggest that you occasionally save your game mid-scenario, so you won’t lose all your progress if you get stuck and have to start over."
         [/part]
         [part]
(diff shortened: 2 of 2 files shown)


CelticMinstrel:
The last clause should be a separate sentence.

stevecotton:
In the Git history, `Address #8009` isn't a useful commit message.

data/campaigns/Under_the_Burning_Suns/_main.cfg
```diff
@@ -19,6 +19,8 @@
 
     description= _ "In the distant future a small band of elves struggles to survive amidst the ruins of fallen empires. Lead your people out of the desert on an epic journey to find a new home.
 
+" + _"Note: This campaign is probably not appropriate for beginners. It changes certain Wesnoth standards, such as the day/night cycle, recall costs and general approach to the difficulty levels, it is also making use of a custom faction of Quenoth elves as a replacement for standard elves.
```
stevecotton, line 22: "... levels. The Quenoth Elves include units that can attack several times per turn, and others that can move after attacking."

Hejnewar: Others that use AoE attacks and change their attacks after attacking multiple times in a row. This campaign also includes heroes with often possibly not easy to use AMLAs. 

...is there even a need to go that far? I don't think so...

soliton-: If you're not going to mention anything about why the Quenoth elves are special I wouldn't mention them at all.

knyghtmare:
This looks too big:
image

Make it better, like this:
image

data/campaigns/Under_the_Burning_Suns/utils/storytxt.cfg
```diff
@@ -4,7 +4,7 @@
...
-            story= _ "Note: This campaign is probably not appropriate for beginners. It changes certain Wesnoth standards, such as the elves’ stats, the day/night cycle, recall costs and general approach to the difficulty levels. This campaign emphasizes role-playing elements and tends to have longer scenarios with objectives that change in the middle. For these reasons, we strongly suggest that you occasionally save your game mid-scenario, so you won’t lose all your progress if you get stuck and have to start over."
+            story= _ "Note: This campaign is probably not appropriate for beginners. It changes certain Wesnoth standards, such as the day/night cycle, recall costs and general approach to the difficulty levels, it is also making use of a custom faction of Quenoth elves as a replacement for standard elves. This campaign emphasizes role-playing elements and tends to have longer scenarios with objectives that change in the middle. For these reasons, we strongly suggest that you occasionally save your game mid-scenario, so you won’t lose all your progress if you get stuck and have to start over."
```
Jonathan-Kelly, line 7: I recommend removing this whole story `[part]` altogether, since the note in the campaign description makes it redundant.

Also, there is no easy or clear way for the player to back out of the campaign by the time they read this. They will have to skip the intro, wait for the scenario to load and then click through all the starting dialogue before they can access the interface to exit the campaign.

Hejnewar: Click what? You just press esc to skip everything.

Jonathan-Kelly: If players guess correctly that pressing `esc` skips through dialogue sequences then that will help speed up the process for them. But it is still a bit of a process to get out of the campaign beginning and they may not think to use this shortcut.

Hejnewar: Honestly I'm not really keen on modifying the campaign in any way other than balance at this point if I don't have to.

data/campaigns/Under_the_Burning_Suns/_main.cfg
soliton-: I would either mention custom units and why they're special or not mention them at all.

data/campaigns/Under_the_Burning_Suns/utils/storytxt.cfg
soliton-: An odd statement in a PR modifying UtBS with no balance implications.

Hejnewar: Eh.. I guess I also shouldnt be fix any issues and just leave them to others.

data/campaigns/Under_the_Burning_Suns/_main.cfg
stevecotton: While the AoE could be mentioned, I think that mentioning the multiple attacks and disengage ability is enough to set the player's expectations for what the campaign will be. So I've taken just my suggestion and put in it #8056.

Hejnewar: Thats kind of important because you can kill your own units with it.

Issue #7093, opened 2022-11-05
On inspect console, text on the right disappears when scrolling


Toranks:
### Game and System Information

The Battle for Wesnoth version 1.17.9 x86_64
Running on Microsoft Windows 7 Service Pack 1 (6.1.7601) x86_64
Distribution channel: Steam

Current video settings
======================

SDL video drivers:        [windows] dummy
Window size:              1920x1080
Game canvas size:         1920x1080
Final render target size: 1920x1080
Screen refresh rate:      60

### Description of the bug

When I try to scroll any section of inspect, the text disappears. That happens whether I use the mouse or the keyboard (arrows or page down). The only way I have to read it all is to use the copy button and read it with an external application. This did not happen in 1.16

### Steps to reproduce the behavior

1. Write :inspect (on debug mode)
2. Click on any section on the left
3. Try to scroll the text on the right if it fills more than the available space

### Expected behavior

It should allow me to scroll through all the text, like in 1.16

Wedge009:
For clarity, inspection of game variables requires debug-mode.

I wonder if this is specific to how Windows renders things (maybe from the big graphics overhaul in 1.17.x?) because I could only confirm this on Windows, not Linux. The issue occurs in current master, both Windows 7 and 10.

Wedge009:
The issue looks to have occurred between 1.17.5 and 1.17.6. I don't think I can do a bisect because during this period the vcpkg configuration was broken, relating to sdl2-mixer 'does not have required feature dynamic-load'.

Toranks:
This happens on plugin manager too, from a certain amount of text
![image](https://user-images.githubusercontent.com/40789364/200313834-971a934d-9f94-4544-a7d7-31812f3a8f1f.png)

Wedge009:
Yes, seems to be the same issue - I can only replicate on Windows, not Linux.

Wedge009:
The add-ons issue cited on 7th November 2022 seems to be what @knyghtmare was reporting in #6879.

Wedge009:
Just noting that while retesting #7201 on Linux, I also run into this.

Pentarctagon:
@Wedge009 so this isn't Windows-specific, then?

Wedge009:
[...]

PR #8029, opened 2023-11-06
Quick hack to get the gamestate inspector working again.

### LINKED ISSUE #7565: [BUG] [Gamestate Inspector] Viewing Unit Data in Inspector UI shows Nothing
### Screenshot
image

### Expected Behaviour
I see the unit data code

### Observed Case
This was HttT for me. Might be anything.

### Version Data
1. Wesnoth 1.17.16+dev (latest)
2. Github Download
3. Windows 10 (msys2)

### Extra Info
the vertical scroll is there, indicating the code is there...just not visible.

Pentarctagon:
This doesn't fix the underlying issue with labels trying to show text too long to be rendered, but it does at least make the text show up now.

Fixes #7565, not sure about the other inspector related issues.

### DIFF
--- a/src/gui/dialogs/gamestate_inspector.cpp
+++ b/src/gui/dialogs/gamestate_inspector.cpp
@@ -43,6 +43,7 @@
 
 #include "display_context.hpp"
 #include "filter_context.hpp"
+#include "video.hpp"
 
 #include <vector>
 #include <functional>
@@ -111,13 +112,13 @@ private:
 	{
 		pages.clear();
 		std::size_t start = 0;
-		while(start + max_inspect_win_len < data.size()) {
+		while(start + page_characters < data.size()) {
 			// This could search into data that's already on a previous page, which is why the result
 			// is then checked for end < start.
-			std::size_t end = data.find_last_of('\n', start + max_inspect_win_len);
+			std::size_t end = data.find_last_of('\n', start + page_characters);
 			int len;
 			if(end == std::string::npos || end < start) {
-				len = max_inspect_win_len;
+				len = page_characters;
 			} else {
 				len = end - start + 1;
 			}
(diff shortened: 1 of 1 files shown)


Wedge009:
I only reliably encounter #7565 in Windows. I used the same HttT starting scenario for testing. Reconfirmed the issue before change and no apparent difference after the change. As best as I can tell, aside from renaming the variable, this is halving the hard-coded limit, making it proportional to `pixel_scale()`, and removing the `static const`.

I'm not sure how this helps your case but it seems to make no difference for me...

Pentarctagon:
Well, before when I clicked a unit it wasn't showing any text and I got the error mentioned in the issue, after the change the text shows up when clicking a unit. If you keep decreasing the limit does text eventually show up?

CelticMinstrel:
Wasn't there also a crash sometimes if the text is too long?

Wedge009:
> I only reliably encounter #7565 in Windows.

Okay, so maybe #7565 and #7093 really are different in that the first shows no text at all and the text only disappears in the second after scrolling. So I can only observe #7093, I can't replicate #7565 at all.

> If you keep decreasing the limit does text eventually show up?

I tried length of 2000 and no difference. So probably different issue.

> Wasn't there also a crash sometimes if the text is too long?

Sounds like #7201.

src/gui/dialogs/gamestate_inspector.cpp
```diff
@@ -128,7 +129,7 @@ class gamestate_inspector::model
 			pages.emplace_back(start, data.size() - start);
 		}
 	}
-	static const unsigned int max_inspect_win_len = 20000;
+	unsigned int page_characters = 10000 / preferences::pixel_scale();
```
stevecotton, line 132: Needs to be `video::get_pixel_scale()`, as the `preferences` can also be set to automatic scaling.

stevecotton:
Agreed, it's a hack but it's better than the status quo.

CelticMinstrel:
So does this also fix the crash?

Pentarctagon:
I don't know if it fixes the crash - I was never able to reproduce that.

src/gui/dialogs/gamestate_inspector.cpp
Pentarctagon: Updated.

stevecotton:
I think it fixes the crash, explanation is in https://github.com/wesnoth/wesnoth/issues/7201#issuecomment-1343697267 .

PR #8031, opened 2023-11-06
WoF: list unique mechanics in campaign description

### LINKED ISSUE #8008: WoF: No documentation about the recall costs

- **Version:** 1.17.23

### Description of the bug

As pointed out by @max-torch on Discord, there's no documentation that WoF has customised the recall costs.
screenshot

_No response_

### Expected behavior

Probably should be stated plainly in the campaign description or in the campaign starting text, since it contradicts the in game help.

### Additional context

_No response_

Jonathan-Kelly:
Resolves #8008 

Any thoughts on the wording or formatting?

### DIFF
--- a/data/campaigns/Winds_of_Fate/_main.cfg
+++ b/data/campaigns/Winds_of_Fate/_main.cfg
@@ -17,8 +17,9 @@
     define=CAMPAIGN_WINDS_OF_FATE
     first_scenario="01_The_Hunt"
     image="data/core/images/portraits/drakes/inferno.webp~SCALE(350,350)"
-    description= _ "As the once plentiful hunting grounds of Morogor slip below the waves, war looms between drakes driven rogue by hunger and those still honorable. Yet the winds of fate may be shifting, for inspired by legends of vast far away lands, a pack of fledgling hunters makes ready to lead their kind on a journey towards salvation.
+    description= _ "As the once plentiful hunting grounds of Morogor slip below the waves, war looms between drakes driven rogue by hunger and those still honorable. Yet the winds of fate may be shifting, for inspired by legends of vast far away lands, a pack of fledgling hunters makes ready to lead their kind on a journey towards salvation." + "
 
+" + _ "<i>Features full recall costs and no gold carryover.</i>" + "
 " + _ "(Hard level, 11 scenarios.)"
 
     {CAMPAIGN_DIFFICULTY EASY      "units/drakes/burner-fly-1.png~CROP(0,0,72,72)~RC(magenta>red)"        ( _ "Aspirant")  ( _ "Easy")}


data/campaigns/Winds_of_Fate/_main.cfg
```diff
@@ -19,7 +19,8 @@
     image="data/core/images/portraits/drakes/inferno.webp~SCALE(350,350)"
     description= _ "As the once plentiful hunting grounds of Morogor slip below the waves, war looms between drakes driven rogue by hunger and those still honorable. Yet the winds of fate may be shifting, for inspired by legends of vast far away lands, a pack of fledgling hunters makes ready to lead their kind on a journey towards salvation.
 
-" + _ "(Hard level, 11 scenarios.)"
+" + _ "(Hard level, 11 scenarios.)
+Features full recall costs and zero gold carryover. Turn limits are rare."
```
soliton-, line 23: I assume full recall cost means recall cost equals recruit cost? Not sure if that is obvious.

knyghtmare (changes requested):
> Features full recall costs and zero gold carryover. Turn limits are rare.

- Better wording (a request to change):

```txt
In this campaign, recall costs are equivalent to recruitment costs, gold carryover for scenarios is zero, and turn limits are rarely used.
```
- Consider adding this in smaller font of a more greyish color.

- Not really a request, but I find enlarging campaign descriptions with such details rather unappealing. Wish we had a "Additional Info" button to add this stuff.

cooljeanius (approved):

Hejnewar:
Wait. You just went from 20g to 100%? I hope you playtested it on various difficulties.

cooljeanius:
> > Features full recall costs and zero gold carryover. Turn limits are rare.
> 
>     * Better wording (a request to change):
> 
> 
> "In this campaign, recall costs are equivalent to recruitment costs, gold carryover for scenarios is zero, and turn limits are rarely used."
> 
>     * Consider adding this in smaller font of a more greyish color.
> 
>     * Not really a request, but I find enlarging campaign descriptions with such details rather unappealing. Wish we had a "Additional Info" button to add this stuff.

I think the additional details are helpful, but would be fine with the suggestion of putting them behind an "Additional Info" button, but also don't think that that should hold up the merging of this PR.

data/campaigns/Winds_of_Fate/_main.cfg
Jonathan-Kelly: Your assumption is correct, though importantly it is the full cost of a unit at its current advancement level, not its original recruit cost (as a level 1). For example, a Drake Fighter recruited at 17 gold which has advanced to become a Drake Blademaster will cost 62 gold to recall. So the main impression players should be given here is that their recalls will generally cost much more than usual.

Also, if this note is longer than a single line of added text then a confusing (and ugly) scroll bar appears inside the description window. Cutting the sentence about turn limits frees up some space but not by too much, unfortunately.

Jonathan-Kelly:
> "In this campaign, recall costs are equivalent to recruitment costs, gold carryover for scenarios is zero, and turn limits are rarely used."

This is too long; it should be just one line or else the interface adds a scroll bar to the campaign description window. Brevity also prevents distracting the player's eye from more critical information, like the campaign difficulty rating and scenario count.

Also, mentioning recruit cost could terribly mislead players into assuming they will be recalling level 3 units at their _original_ (level 1) recruitment costs. Which is quite the opposite of what they will experience.

> Consider adding this in smaller font of a more greyish color.

That may be too difficult for many to read against a dark background?

> Wait. You just went from 20g to 100%? I hope you playtested it on various difficulties.

Tested heavily on the highest 'nightmare' difficulty (enemies have about 4x resource advantage) and the default 'normal' difficulty has players only facing enemies with no resource advantage. There is a 'hard' difficulty interpolated between those two with 2x enemy resource advantage. Plus a (super) easy mode with enemies having half the resources of the player.

How did you test the easier difficulties for UtBS ?

Hejnewar:
> Tested heavily on the highest 'nightmare' difficulty (enemies have about 4x resource advantage) and the default 'normal' difficulty has players only facing enemies with no resource advantage. There is a 'hard' difficulty interpolated between those two with 2x enemy resource advantage. Plus a (super) easy mode with enemies having half the resources of the player.

I played thru 2 random scenarios and that doesnt seem great. One was too easy and the other was both too easy and too hard at the same time. Need to check more but I can easily imagine unbeatable scenario somewhere along the way. 

> How did you test the easier difficulties for UtBS ?

I didnt have to, I set the recall cost % to one that on average gives ~20g / lvl 3 recall.

Jonathan-Kelly:
> I played thru 2 random scenarios and that doesnt seem great. One was too easy and the other was both too easy and too hard at the same time.

Could you post the replays so I can take a closer look at what happened?

Hejnewar:
I think there is no point. Just get to landfall (since there is no carryover there is no RNG you can just debug there), activate becon at like turn 2 and there isn't really much left after that, at least that was the case for me, where I just had to clean up. On the other hand if you ar too late you can just die cuz you are not winning Vs this force with 6 clashers and heroes. 

The other was the elv scenario which depends on flames and enemy movement RNG a bit, but I can easily imagine it not getting enough value and again on flat there isn't really a way for you to win Vs this force.

Jonathan-Kelly:
It sounds like those scenarios could use adjustments then, but I would need to see your replays to know what to adjust and by how much.

knyghtmare (approved):

cooljeanius (approved):

Jonathan-Kelly:
wof

Pentarctagon:
Can this be merged?

CelticMinstrel:
This may not be the best place to say it, but I think "full recall cost" is an extremely terrible idea and should be revoked.

data/campaigns/Winds_of_Fate/_main.cfg
```diff
@@ -19,7 +19,8 @@
     image="data/core/images/portraits/drakes/inferno.webp~SCALE(350,350)"
     description= _ "As the once plentiful hunting grounds of Morogor slip below the waves, war looms between drakes driven rogue by hunger and those still honorable. Yet the winds of fate may be shifting, for inspired by legends of vast far away lands, a pack of fledgling hunters makes ready to lead their kind on a journey towards salvation.
 
-" + _ "(Hard level, 11 scenarios.)"
+" + _ "(Hard level, 11 scenarios.)
+<i>Features full recall costs and no gold carryover.</i>"
```
stevecotton, line 23: I'd prefer this to be a separate translatable string, rather than combining it with the difficultly and scenario count. That would allow it to be moved above the scenario count without invalidating the translation, or removed if @CelticMinstrel's suggestion of lesser recall costs is taken.

Jonathan-Kelly:
wof

Jonathan-Kelly:
> This may not be the best place to say it, but I think "full recall cost" is an extremely terrible idea and should be revoked.

Why do you think it is a terrible idea? Or to put it another way, what do you think are the critical advantages of having cheaper recalls?

PR #8035, opened 2023-11-07
Help Browser: only show one image if male and female are the same

stevecotton:
If a unit type has male and female versions, then two images of the unit are displayed at the top-left of the help page. However, it did this even if the two images were the same, which made the duplication look like a bug.

The Naga Fighter is one of the affected unit types.

This was reported in the [WoF discussion about female drakes](https://r.wesnoth.org/p684546), and I was in the middle of logging a "good first issue" bug when I realised it was just a one-liner.

Screenshot **without** the fix:
image

### DIFF
--- a/changelog_entries/help_duplicate_images.md
+++ b/changelog_entries/help_duplicate_images.md
@@ -1,0 +1,2 @@
+ ### User interface
+   * The help browser, when looking at a unit type which uses the same image for male and female units, now only shows one image instead of showing a duplicate.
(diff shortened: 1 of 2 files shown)


Wedge009:
I don't really understand the logic of the original `&female_type != &male_type` - seems it has been this way since f969e60ca5937efaff016d1a8d81771c8e9a0da3 (nine years ago). But the change makes sense to me.

soliton-:
Been that way since 19 years 
https://github.com/wesnoth/wesnoth/commit/04ed3003f29e9d73ccc6f63a0dffee9c864de207#diff-f381f414e6e17a9351f331060ee29fea20ecd825006de601b7bfe74501d9367cR560

It checks if those types are actually different regardless if the images might be the same anyway.

Wedge009:
So the change is definitely an improvement then.

Issue #8037, opened 2023-11-09
UtBS: Mystic line doesnt always heal in S2 


Hejnewar:
### Game and System Information

- **Version: 1.17.22**
- **Downloaded from: Steam**

### Description of the bug

Mystic line unit (or any unit with UtBS heal) start to randomly heal or not heal after you have units with the debuff. 

### Steps to reproduce the behavior

1. Start UtBS S2
2. Play it
3. As you get units with the debuff (scenario specific one) your units without it also stop being healed seemingly at random, it seems the more debuffs you have the worse it is for units without it.

stevecotton:
I don't see a bug there - what prevents healing isn't dehydration itself, it's the /preventation/ of dehydration.

Looking at that photo, if the shaman wasn't there, Garak would become dehydrated. So the shaman prevents his dehydration instead of healing him.

Hejnewar:
So wait, the bug is actually that they heal when they shouldnt? 

![11111](https://github.com/wesnoth/wesnoth/assets/41817546/7cb5e4bf-8681-4d94-9469-483eb57ad0f9)
![2222222](https://github.com/wesnoth/wesnoth/assets/41817546/97ecbeb1-e387-43c9-a4d5-6e33c779024a)

stevecotton:
Hex 13,54 is terrain type `Ds`, "Beach Sands". That reminds me of #5149, which implies the bug has been in UtBS since 1.14.9.

Looks like the bug is https://github.com/wesnoth/wesnoth/blob/bf6b64be78550a6803c8f54eaa0b8e67f21c5429/data/campaigns/Under_the_Burning_Suns/utils/dehydration-utils.cfg#L6-L7

stevecotton:
Fixed (to make it heal on less terrains) in #8047.

PR #8039, opened 2023-11-11
New unit test for UtBS-style [heals] with a filter

stevecotton:
The ability's code in UtBS hasn't changed since 1.16, but there's now bug #8037 about healing sometimes not working in UtBS S02. The new test was written to check for an engine bug when a healer has two units next to it, only one of which passes the filter. Although the test is passing, it seems reasonable to keep the test.

Also, use `{ASSERT ...}` for better logging in the old healing test.

### DIFF
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/heal.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/heal.cfg
@@ -31,17 +31,10 @@
         variable=healer2
         kill=no
     [/store_unit]
-    [if]
-        [not]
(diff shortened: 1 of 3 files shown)


stevecotton:
@Pentarctagon the build failure is a network error, the MP Match History dialog has made the Boost unit tests depend on a server :(

Pentarctagon:
Build error?

stevecotton:
I reran it, the failure was https://github.com/wesnoth/wesnoth/actions/runs/6836419054/job/18591295993

Haven't logged a bug yet, it seems more complex than I thought. That test passes when completely blocked from network access (so DNS resolution fails in my local test run).

Pentarctagon:
Yeah, I'm not sure what would cause that - the test itself is hardly new.

PR #8041, opened 2023-11-12
Show an error when mainline campaigns are missing.

### LINKED ISSUE #7875: Check that mainline campaigns are present
On starting Wesnoth, it should check that all mainline campaigns are present and show a popup listing any that are missing with general instructions for how Linux users can get them installed as well as state that any add-ons which depend on a missing campaign will not work. This:
* Helps players who accidentally install the wrong package on Linux distros that package campaigns separately from the base game.
* Helps ensure that add-ons can rely on the mainline campaigns being present, if they decide to use content from them.

The popup should have a "don't show this again" option for players who do intentionally install only the base game and individual mainline campaigns.

Pentarctagon:
A Linux packaging issue has, for apparently multiple years now, resulted in Wesnoth getting installed with only the tutorial campaign available. So, check for that and alert the player that there's an issue so they don't think Wesnoth actually only has a tutorial available.

Fixes #7875

### DIFF
--- a/src/gui/dialogs/campaign_selection.cpp
+++ b/src/gui/dialogs/campaign_selection.cpp
@@ -360,7 +360,9 @@ void campaign_selection::pre_show(window& window)
 		widget_data data;
 		widget_item item;
 
-		item["label"] = _("Most mainline campaigns are missing. If you’re on Linux, then this most likely means the incorrect package was installed. To get the rest of the campaigns you will need to either: correctly install the wesnoth metapackage, install each campaign’s package one by one, or install Wesnoth via Steam or Flatpak.");
+		// TRANSLATORS: "more than 15" gives a little leeway to add or remove one without changing the translatable text.
+		// It's already ambiguous, 1.18 has 19 campaigns, if you include the tutorial and multiplayer-only World Conquest.
+		item["label"] = _("Wesnoth normally includes more than 15 mainline campaigns, even before installing any from the add-ons server. If you’ve installed the game via a package manager, there’s probably a separate package to install the complete game data.");
 		data.emplace("description", item);
 
 		pages.add_page(data);
--- a/src/gui/dialogs/campaign_selection.cpp
+++ b/src/gui/dialogs/campaign_selection.cpp
@@ -57,6 +57,12 @@ void campaign_selection::campaign_selected()
 	if(!tree.selected_item()->id().empty()) {
 		auto iter = std::find(page_ids_.begin(), page_ids_.end(), tree.selected_item()->id());
 
+		if(tree.selected_item()->id() == missing_campaign_) {
+			find_widget<button>(this, "ok", false).set_active(false);
+		} else {
+			find_widget<button>(this, "ok", false).set_active(true);
+		}
+
 		const int choice = std::distance(page_ids_.begin(), iter);
 		if(iter == page_ids_.end()) {
 			return;
@@ -347,7 +353,7 @@ void campaign_selection::pre_show(window& window)
 		missing["icon"] = "units/unknown-unit.png";
 		missing["name"] = _("Missing Campaigns");
 		missing["completed"] = false;
-		missing["id"] = "missing-campaign";
+		missing["id"] = missing_campaign_;
 
 		add_campaign_to_tree(missing);
 
@@ -358,7 +364,7 @@ void campaign_selection::pre_show(window& window)
 		data.emplace("description", item);
 
 		pages.add_page(data);
-		page_ids_.push_back("missing-campaign");
+		page_ids_.push_back(missing_campaign_);
 	}
 
 	//
--- a/src/gui/dialogs/campaign_selection.cpp
+++ b/src/gui/dialogs/campaign_selection.cpp
@@ -17,9 +17,10 @@
 
 #include "gui/dialogs/campaign_selection.hpp"
 
+#include "filesystem.hpp"
 #include "font/text_formatting.hpp"
-#include "gui/dialogs/campaign_difficulty.hpp"
 #include "gui/auxiliary/find_widget.hpp"
+#include "gui/dialogs/campaign_difficulty.hpp"
 #include "gui/widgets/button.hpp"
 #include "gui/widgets/image.hpp"
 #include "gui/widgets/listbox.hpp"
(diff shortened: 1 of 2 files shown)


src/gui/dialogs/campaign_selection.cpp
```diff
@@ -288,6 +290,12 @@ void campaign_selection::filter_text_changed(const std::string& text)
 
 void campaign_selection::pre_show(window& window)
 {
+	std::vector<std::string> dirs;
+	filesystem::get_files_in_dir(game_config::path + "/data/campaigns", nullptr, &dirs);
+	if(dirs.size() <= 1) {
+		gui2::show_error_message(_("Most mainline campaigns are missing. If you’re on Linux, then this means either you or the software center you’re using installed the incorrect package. To get the rest of the campaigns you will need to either: correctly install the wesnoth metapackage, install each campaign's package one by one, or install Wesnoth via Steam or Flatpak."));
```
Wedge009, line 296: Suggest using the typographic apostrophe for campaign.

Wedge009:
I suppose this addresses the bulk of your issue - I was wondering what was the simplest way to show a dialogue. I don't think this has the 'don't show me again' option you specified in #7875 but I reckon it's probably unnecessary. Even if it was intentional to only install the core packages, one might come back to the game after a long while and forget about the installation specifics.

stevecotton:
Do we have any feedback from Rhonda on this?

The text feels unnecessarily accusatory to me, even accepting that we don't want to support the separated campaign packages. I'll think about and suggest a different wording.

Pentarctagon:
Updated the wording.

Pentarctagon:
The motivation is rather that due to a packaging issue, *all* users who are installing Wesnoth on Ubuntu (potentially any Debian-based distro) are only getting the tutorial installed. Given that, we should inform players that Wesnoth does actually have more than just a tutorial campaign.

CelticMinstrel:
1. I would say "most of the campaigns are missing" (adding two words).
2. The details of this message seem like they're overly specific to a particular Linux distribution.

Pentarctagon:
Would stevecotton's wording in https://github.com/wesnoth/wesnoth/issues/7875 be better?

Pentarctagon:
Updated to the approach stevecotton showed in the linked issue.

CelticMinstrel:
No reaction to my comment at all…?

Pentarctagon:
> Would stevecotton's wording in https://github.com/wesnoth/wesnoth/issues/7875 be better?

Was my reply. I think it should specify mainline campaigns in the message though.

src/gui/dialogs/campaign_selection.cpp
```diff
@@ -339,6 +340,27 @@ void campaign_selection::pre_show(window& window)
 		page_ids_.push_back(campaign["id"]);
 	}
 
+	std::vector<std::string> dirs;
+	filesystem::get_files_in_dir(game_config::path + "/data/campaigns", nullptr, &dirs);
+	if(dirs.size() <= 1) {
```
stevecotton, line 345: Maybe 5 or 10 here, so the message still shows up if someone's told the software center to install Heir to the Throne, and `wesnoth-httt` has pulled in `wesnoth-core`. That would also cover some service dropping `Thumbs.db` files all over the filesystem.

stevecotton:
The "Play" button on the fake campaign currently launches the tutorial. That's a trivial detail, which can be left to a later PR as it wouldn't affect the string freeze; I'm only commenting to save anyone else the time for testing.

src/gui/dialogs/campaign_selection.cpp
Pentarctagon: If someone manually installed an additional campaign, then they would know that there are more to install already.

This also only gets a list of the directories, so a Thumbs.db file wouldn't get included in that count.

stevecotton: Not necessarily. They might believe that there's only the tutorial, the single official campaign, and the add-ons server.

Pentarctagon: How would they only find out about a single campaign package and not any of the others?

Pentarctagon:
Updated to disable the Play button.

src/gui/dialogs/campaign_selection.hpp
```diff
@@ -131,6 +131,8 @@ class campaign_selection : public modal_dialog
 	bool currently_sorted_asc_;
 
 	std::vector<std::string> last_search_words_;
+
+	inline const static std::string missing_campaign_ = "missing-campaign";
```
soliton-, line 135: Better would be something that is otherwise invalid.

stevecotton:
> Would stevecotton's wording in #7875 be better?
> 
> Was my reply. I think it should specify mainline campaigns in the message though.

```
// TRANSLATORS: "more than 15" gives a little leeway to add or remove one without changing the translatable text.
// It's already ambiguous, 1.18 has 19 campaigns, if you include the tutorial and multiplayer-only World Conquest.
item["label"] = _("Wesnoth normally includes more than 15 mainline campaigns, even before installing any from the add-ons server. If you’ve installed the game via a package manager, there’s probably a separate package to install the complete game data.");
```

src/gui/dialogs/campaign_selection.hpp
Pentarctagon: Updated.

stevecotton (approved):

src/gui/dialogs/campaign_selection.cpp
CelticMinstrel: I don't think you should underestimate the strange things that people might do. As a random example, maybe it's possible that their search in the package manager happened to match a keyword for one of the campaigns, and they installed that campaign without realizing it was just one of several for the base game. Obviously installing the campaign would also pull in the base game as a dependency, and just like in Stevecotton's example, they'd have a game with only one of the official campaigns plus the tutorial.

Why not just make the internal number match the number shown to the users? That is, if there are fewer than 16 campaigns (15 plus the tutorial), show the message.

Pentarctagon: Alright, updated then.

Wedge009: Regarding the 'updated' reply - maybe I'm mistaken but it looks like the final commit still says `<= 1` directories.

CelticMinstrel: It's an additional commit pushed separately to master after merging: d492a36019eaddd9ed3a664b25f31887615f0dc7

PR #8043, opened 2023-11-12
Favour OpenGL over DirectX on Windows (resolves #8038)

### LINKED ISSUE #8038: Screen redraw stops after Windows lock screen

- **Version: 1.17.22**
- **Downloaded from: Steam**
- **OS: Windows**

### Description of the bug

After afking with the game in the background its screen turns to black (which i believe is a known bug) after which it quickly gets updated as i move it around / animations play but one part of it no longer gets updated. 

hhhhuh

1. Wait till game screen turns black in the background
2. Mouse over unit descriptions  

### Expected behavior

_No response_

### Additional context

_No response_

Wedge009:
No idea if we really want to take this option but this works around the issue of #8038 by avoiding the specific renderer that exhibits that issue - namely DirectX.

### DIFF
--- a/src/sdl/window.cpp
+++ b/src/sdl/window.cpp
@@ -40,6 +40,13 @@ window::window(const std::string& title,
 		throw exception("Failed to create a SDL_Window object.", true);
 	}
 
+#ifdef _WIN32
+	// SDL usually selects DirectX renderer on Windows systems. However, returning from the Windows lock screen causes issues with rendering with DirectX.
+	// While it would be best to resolve this by handling the SDL_RENDER_TARGETS_RESET event, a straightforward work-around is to just switch to OpenGL rendering.
+	// See https://github.com/wesnoth/wesnoth/issues/8038 for details.
+	SDL_SetHint(SDL_HINT_RENDER_DRIVER, "opengl");
+#endif
+
 	if(sdl::runtime_at_least(2,0,10)) {
 		// Rendering in batches (for efficiency) is enabled by default from SDL 2.0.10
 		// The way Wesnoth uses SDL as of September 2019 does not work well with this rendering mode (eg story-only scenarios)


Pentarctagon:
Out of curiosity, does using "vulkan" as the option value work?

soliton-:
Would be nice to check that the corresponding env variable (i think SDL_RENDER_DRIVER) is not set so a user can still override this.

mesilliac:
Normal hints have lower priority than env vars, so it should override correctly if the user sets it explicitly.

Wedge009:
> Out of curiosity, does using "vulkan" as the option value work?

https://wiki.libsdl.org/SDL2/SDL_HINT_RENDER_DRIVER suggests there's no option for vulkan, so I never tried it. But I tried it just now and it appears the hint has no effect - #8038 returns presumably because the invalid hint is ignored and DirectX is used as normal.

> Normal hints have lower priority than env vars, so it should override correctly if the user sets it explicitly.

That's what https://wiki.libsdl.org/SDL2/SDL_SetHint states as well.

Pentarctagon:
Alright, looks good to me then.

Wedge009:
Just to be sure, I also reconfirmed #8038 when I added `set SDL_RENDER_DRIVER=directx`.

Wedge009:
Doing a bit more research on this: I noticed [this comment](https://github.com/libsdl-org/SDL/issues/4869#issuecomment-1007628224) and also tried `direct3d11` - that also seemed to avoid #8038. Do we still care to consider Direct3D for Windows users or should we just make it OpenGL to be consistent with Linux and macOS? (https://wiki.libsdl.org/SDL2/SDL_HINT_RENDER_DRIVER seems to suggest `opengl` would take precedence over `metal` by default)

The actual header shows the additional `direct3d` options that the wiki documentation does not:
https://github.com/libsdl-org/SDL/blob/SDL2/include/SDL_hints.h#L1410-L1430

`direct3d12` works around #8038 on Windows 10 but not Windows 7. `direct3d11` works on both. `direct3d` confirms the default behaviour (broken rendering after lock screen).

I also found [this discussion](https://discourse.libsdl.org/t/sdl-hint-render-batching-question/27060/3) that suggests (the [still undocumented](https://wiki.libsdl.org/SDL2/SDL_HINT_RENDER_BATCHING)) `SDL_HINT_RENDER_BATCHING` should go before `SDL_HINT_RENDER_DRIVER`. But I'm also checking whether we still need to disable batched renders in the first place - so far I can't replicate any of the issues that required the disabling.

Pentarctagon:
If it works, I'd say go with direct3d11, if only to use Microsoft's API on Microsoft's platform. I don't really have a strong opinion either way though.

Wedge009:
Wesnoth supports minimum Windows 7, right?

Pentarctagon:
It's currently `Microsoft Windows 10 1903 (64-bit) or later` for 1.17, though that's not a hard cutoff.

Wedge009:
Closing in favour of #8045.

Issue #4309, opened 2019-09-03
1.15.1 Switching virtual desktops with full-screen wesnoth causes graphics issues


vgaming:
Steps to reproduce:
* start wesnoth
* open preferences > Display > enable "full-screen" checkbox
* start any game
* switch to another application
* switch back to wesnoth

Expected at this point:
* battlefield is shown normally

Actually:
![2019 09 02_22:35:19](https://user-images.githubusercontent.com/33452702/64150848-667d9d00-ce29-11e9-9dcf-ca6baa04d0aa.png)

jyrkive:
Which operating system do you use? This may well be OS specific (and even depend on your window manager if you use GNU/Linux).

vgaming:
@jyrkive I'm using ArchLinux, XFCE desktop environment, but no compositor (window manager) at all. (To be clear: yes, really, no compositor. Neither xfwm4 nor compton.)

vgaming:
When I hover the tiles, they get back to normal.

jyrkive:
All right, I added the Linux label for now.

Pentarctagon:
I'm not able to reproduce this on Linux Mint 19.2.

vgaming:
I tried to reproduce locally again, but on the first attempt, I couldn't (read further).

The problem apparently only appears if you have "Preferences -> Didsplay -> Full screen" checkbox enabled. In non-fullscreen mode, there's no such bug for me as well.

Can anybody please confirm the existence or not existence of this bug if said checkbox is enabled? (BTW, it is the default preference when people install wesnoth anew.)

dvladf:
I couldn't reproduce this problem (with fullscreen mode) on Kubuntu 18.04. Probably, this is specific for xfce.

vgaming:
Very interesting. My system is XFCE indeed, but there's no window manager, so it should be identical to any other DE with no compositor. I'll check this bug with a compositor enabled in the evening.

Pentarctagon:
I'm still not able to reproduce the issue when switching between Wesnoth and TF2, whether either or both are fullscreen.

Does anything change if you do `:refresh`?

jostephd:
The minimap part looks similar to @Konrad22's screenshots on #4237/#4245.

And there's also #4196.

I use openbox on Debian, in case it matters.

Wedge009:
[...]

Issue #4245, opened 2019-08-17
Story scenes do not render properly (black screen)


Konrad22:
I'm stuck with a black screen. The window still reacts, I get my nice cursor. I just can't do anything with wesnoth except close it (which works as if everything else was fine as well).
![](https://cdn.discordapp.com/attachments/259976436490829825/612046914837414107/unknown.png)
Windows 10, 1.15.0
To reproduce: Load the savefile and try to proceed without using 'esc'.
https://cdn.discordapp.com/attachments/259976436490829825/612047813915705405/SotA-Mortality.gz
If you skip the talking by pressing esc when Ras-Tabahn speaks (for the first time, which is also the last things thats said) or earlier, then everthing works fine.

sigurdfdragon:
Can't reproduce with the attached save or my own using either esc or not.
I get the game moving on as expected tested against current master

Konrad22:
As you can see, this bug is still present in 1.15.1 and is not just about SotA. (E.g. HttT, Elvish Council)
![grafik](https://user-images.githubusercontent.com/39015360/63386874-51722a00-c3a4-11e9-9783-6aea9c615e77.png)
![grafik](https://user-images.githubusercontent.com/39015360/63386882-56cf7480-c3a4-11e9-89e0-128ac3cef1e9.png)

![grafik](https://user-images.githubusercontent.com/39015360/63386976-926a3e80-c3a4-11e9-84d2-f757cc973186.png)

Konrad22:
There is also this, which might be related.
![grafik](https://user-images.githubusercontent.com/39015360/63388166-c85cf200-c3a7-11e9-9d33-756ed310411e.png)
Mousing over hexes or tabbing out 'fixes' it though.

Konrad22:
I have exactly the same issues on a different pc of mine. @Wedge009 remembered encountering something like the last screenshot with a bad graphics driver bug.
I have Intel(R) HD Graphics 520 on one and NVIDIA GeForce 9600 GT on the other.

grz0:
I have an issue on Linux which may be related to this one
https://github.com/wesnoth/wesnoth/issues/4196

Wedge009:
Haven't yet had time to test my various platforms but would be interesting to see if there are any common patterns here in terms of graphics hardware, drivers, operating systems, SDL libraries, or even something else entirely.

Konrad22:
[...]

Issue #4237, opened 2019-08-15
Mini-map elements are displayed in non-game screens


Konrad22:
![colorpoints](https://user-images.githubusercontent.com/39015360/63128125-79344d00-bfb4-11e9-8293-81745ed42089.png)
![colorpoints2](https://user-images.githubusercontent.com/39015360/63128131-7b96a700-bfb4-11e9-9e98-7ecf3ba95eec.png)
![grafik](https://user-images.githubusercontent.com/39015360/63128527-6d955600-bfb5-11e9-8ba0-6d37e7dadadb.png)
I just tried starting HttT and THoT. No idea what this is, what's causing it. And it doesn't happen every time (e.g. nothing happened when going from HttT S3 to S4).
Windows 10, 1.15.0

EDIT:
Tabbing out 'fixed' the black background in the last picture.

jostephd:
Looks like the minimap. Probably another regression from #4188? There's also #4230 which is already open.

Konrad22:
I can confirm that this is the minimap. It matches the units present at the time of scenario start.

Konrad22:
I'd like to point out that this does not account for the 'black background'.

jostephd:
I can't reproduce this.

At this point, as I said on #3120, I think it would be best if someone reviewed #4188. It's known to have introduced two regressions and I imagine it may have introduced more.

jostephd:
#4188 has been reverted (#4238). @Konrad22 Would you test this again in 1.15.1 when it's released?

Konrad22:
Of course. Maybe it will also fix this:
![grafik](https://user-images.githubusercontent.com/39015360/63187157-7d1ca980-c05e-11e9-921b-f8a2050a347a.png)

Konrad22:
This bug is still present in 1.15.1.
![grafik](https://user-images.githubusercontent.com/39015360/63386780-24be1280-c3a4-11e9-9cee-a75072719140.png)

Konrad22:
I have exactly the same issues on a different pc of mine. @Wedge009 remembered encountering something like this with a bad graphics driver bug.
I have Intel(R) HD Graphics 520 on one and NVIDIA GeForce 9600 GT on the other.

Konrad22:
For some reason I do not encounter this bug when playing 1.15 on steam. But I am guaranteed to encounter it when using the sourceforge version.

Konrad22:
[...]

PR #8045, opened 2023-11-13
Update SDL render hints

### LINKED ISSUE #8038: Screen redraw stops after Windows lock screen

Wedge009:
Resolves #8038 by requesting Direct3D v11 over the default Direct3D v9 on Windows platforms.

According to https://discourse.libsdl.org/t/a-couple-of-questions-regarding-batching-in-sdl-2-0-10/26453/2, `SDL_HINT_RENDER_DRIVER` implies `SDL_HINT_RENDER_BATCHING=0`. I tested `SDL_HINT_RENDER_BATCHING=1` on Windows and Linux and was not able to replicate #4237 nor #4245 (I was never able to replicate the related #4309). Given the rendering work I think disabling batched renders is no longer necessary.

I suppose we could force `SDL_HINT_RENDER_BATCHING=1` on Windows as well just for consistency but Wesnoth isn't exactly a demanding application in terms of rendering. If someone with sufficiently low-end hardware can demonstrate a significant difference between batched and non-batched renders, then we can consider enforcing it for Windows too.

### DIFF
--- a/src/sdl/window.cpp
+++ b/src/sdl/window.cpp
@@ -40,11 +40,17 @@ window::window(const std::string& title,
 		throw exception("Failed to create a SDL_Window object.", true);
 	}
 
-	if(sdl::runtime_at_least(2,0,10)) {
-		// Rendering in batches (for efficiency) is enabled by default from SDL 2.0.10
-		// The way Wesnoth uses SDL as of September 2019 does not work well with this rendering mode (eg story-only scenarios)
-		SDL_SetHint(SDL_HINT_RENDER_BATCHING, "0");
-	}
+#ifdef _WIN32
+	// SDL uses Direct3D v9 by default on Windows systems. However, returning
+	// from the Windows lock screen causes issues with rendering. Resolution
+	// is either to rebuild render textures on the SDL_RENDER_TARGETS_RESET
+	// event or use an alternative renderer that does not have this issue.
+	// Suitable options are Direct3D v11+ or OpenGL.
+	// See https://github.com/wesnoth/wesnoth/issues/8038 for details.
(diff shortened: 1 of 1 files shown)


Wedge009:
Why does the GitHub bot like removing labels?

soliton-:
This is what is configured as the Graphics label currently:

https://github.com/wesnoth/wesnoth/blob/8b111a2a76ea8ef098fac911bad33ad000eeefca/.github/labeler.yml#L57-L60
We could add src/sdl there.

Also the sync-labels option is set here:
https://github.com/wesnoth/wesnoth/blob/8b111a2a76ea8ef098fac911bad33ad000eeefca/.github/workflows/labeler.yml#L14
which should mean it does not delete labels. I think the quotes are wrong there though and false is the default anyway so I'll remove that.

Pentarctagon:
Any reason not to go ahead and merge this?

Wedge009:
I don't think so. It's the same principle as #8043 but this time using `direct3d11` instead of `opengl` and also removes the disabling of batched renders that we had to introduce for SDL 2.0.10.

The only caution is that on non-Windows systems it will now use the default batched renders setting but I did not observe any of the artefacts that plagued us before, on the old software rendering.

PR #8046, opened 2023-11-14
verify what ''add'' 'sub' etc.. modify value by default in filter_ability checks 'value'(postpone1.19)

newfrenchy83:
As @soliton- in https://github.com/wesnoth/wesnoth/issues/7992#issuecomment-1795270245 noticed the presence of 'add' 'sub' etc.. is not enough to determine if the default value of an ability is changed (add=0, multiply=1); hence this repair to refine it to ensure the value is not changed.

### DIFF
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -1421,6 +1421,27 @@ void unit::remove_ability_by_id(const std::string& ability)
 	}
 }
 
+ /**
+ * Returns true if any of "add", "sub", "multiply" or "divide" are present and have a value
+ * other than their respective no-op value.
+ */
+static bool has_value_changing_modifier(const config& cfg)
+{
+	if(cfg.has_attribute("add") && cfg["add"].to_int(0) != 0){
+		return true;
+	}
+	if(cfg.has_attribute("sub") && cfg["sub"].to_int(0) != 0){
+		return true;
+	}
+	if(cfg.has_attribute("multiply") && cfg["multiply"].to_double(1) != 1){
+		return true;
+	}
+	if(cfg.has_attribute("divide") && cfg["divide"].to_double(1) != 1){
+		return true;
+	}
+	return false;
+}
+
 static bool matches_ability_filter(const config & cfg, const std::string& tag_name, const config & filter)
 {
 	using namespace utils::config_filters;
@@ -1472,8 +1493,7 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 		return false;
 
 	if(!filter["value"].empty()){
-		bool has_other_key = (!cfg["add"].empty() || !cfg["sub"].empty() || !cfg["multiply"].empty() || !cfg["divide"].empty());
-		if(!has_other_key){
+		if(!has_value_changing_modifier(cfg)){
 			if(tag_name == "drains"){
 				if(!int_matches_if_present(filter, cfg, "value", 50)){
 					return false;
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -1493,7 +1493,12 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 		return false;
 
 	if(!filter["value"].empty()){
-		if(!has_value_changing_modifier(cfg)){
+		if(has_value_changing_modifier(cfg)){
+			//if true, then default value modified, and can't be checked.
+			if(!int_matches_if_present(filter, cfg, "value")){
+				return false;
+			}
+		} else {
 			if(tag_name == "drains"){
 				if(!int_matches_if_present(filter, cfg, "value", 50)){
 					return false;
@@ -1506,10 +1511,11 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 				if(!int_matches_if_present(filter, cfg, "value" , 0)){
 					return false;
 				}
-			}
-		} else {
-			if(!int_matches_if_present(filter, cfg, "value")){
-				return false;
+				//if other tags, then ability can't have default value
+			} else {
+				if(!int_matches_if_present(filter, cfg, "value")){
+					return false;
+				}
 			}
 		}
 	}


src/utils/config_filters.hpp
```diff
@@ -69,4 +69,8 @@ bool string_matches_if_present(
 
 bool bool_or_empty(const config& filter, const config& cfg, const std::string& attribute);
 
+bool other_key_int(const config& cfg, const std::string& attribute);
+
+bool other_key_double(const config& cfg, const std::string& attribute);
```
stevecotton, line 74: This is unreadable code, there's no reason to expect "int" to mean "is being used for addition/subtraction" or "double" to mean "is being used for multiplication/division". What I'd normally say in a code review is "this needs documentation, without just repeating the words used in the function name", in the expectation that a better name will be found in the process of writing the documentation.

However, IMO the functions have so little code in them that it should either be in `matches_ability_filter` or in a single utility function in the same file as `matches_ability_filter`, for example
```
/**
 * Returns true if any of "add", "sub", "multiply" or "divide" are present and have a value
 * other than their respective no-op value.
 */
has_value_changing_modifier(const config& cfg);

Comment (newfrenchy83) [reply src/utils/config_filters.hpp]: it's done

File: src/units/unit.cpp [hunk 2]
```diff
@@ -1421,6 +1421,15 @@ void unit::remove_ability_by_id(const std::string& ability)
 	}
 }
 
+ /**
+ * Returns true if any of "add", "sub", "multiply" or "divide" are present and have a value
+ * other than their respective no-op value.
+ */
+static bool has_value_changing_modifier(const config& cfg)
+{
+	return  ((cfg.has_attribute("add") && cfg["add"].to_int(0) != 0) || (cfg.has_attribute("sub") && cfg["sub"].to_int(0) != 0) || (cfg.has_attribute("multiply") && cfg["multiply"].to_double(1) != 1) || (cfg.has_attribute("divide") && cfg["divide"].to_double(1) != 1));
```
stevecotton, line 1430: Please split this across multiple lines.

src/units/unit.cpp
```diff
@@ -1472,8 +1481,7 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 		return false;
 
 	if(!filter["value"].empty()){
-		bool has_other_key = (!cfg["add"].empty() || !cfg["sub"].empty() || !cfg["multiply"].empty() || !cfg["divide"].empty());
-		if(!has_other_key){
+		if(!has_value_changing_modifier(cfg)){
```
stevecotton, line 1484: Looking at the conditional block contained by this, I don't understand the `else` block. If there *is* a value-changing modifier, it checks whether the unchanged value matches.

newfrenchy83: when has_value_changing_modifier is true, then if not has_attribute_value, return false because default value of abilities will be changed, and absence of attribut eqals. It's only when has_value_changing_modifier is false what default value can be checked if not has_attribute("value")(in line 1484 has_value_changing_modifier() is false)

src/units/unit.cpp

src/units/unit.cpp
```diff
@@ -1421,93 +1421,146 @@ void unit::remove_ability_by_id(const std::string& ability)
...
-				if(!int_matches_if_present(filter, cfg, "value", 1)){
-					return false;
+		if(!string_matches_if_present(filter, cfg, "alternative_type", ""))
+			return false;
+
+		//for plague only
+		if(!string_matches_if_present(filter, cfg, "type", ""))
+			return false;
+
+		//if ability not return numerical value, and numerical attribute filtered, return false.
```
stevecotton, line 1523: What is this second commit for? Why should the engine know about this, instead of just assuming that the person writing the filter knew what they were filtering for?

Even if this should be here, it needs a comment in the code explaining the purpose, the reason for its existence; a comment that just translates C++ into English isn't saying *why* the code is there.

The list of tags shouldn't be in src/utils, because it's specific to units.

newfrenchy83: second commit removed

soliton-:
IMO this approach of approximately reimplementing ability logic is not useful. Either the actual ability logic/code should be reused or this should not be attempted and the limitation documented.

stevecotton:
Just realised that this could be done in WML, which would fit with the idea of making it an experimental API that could change in future.

Instead of
```
[filter_ability]
    value=10
    # engine will check for add=, sub=, multiply= or divide= affecting the value
[/filter_ability]
```
we could have
```
[filter_ability]
    value=10
    {FILTER_ABILITY_DO_NOT_MATCH_MODIFIED_VALUE}
[/filter_ability]
```

That means this logic would only be used when the WML author intends to use it, and a macro is much easier to remove than engine functionality.

newfrenchy83:
> Just realised that this could be done in WML, which would fit with the idea of making it an experimental API that could change in future.
> 
> Instead of
> 
> ```
> [filter_ability]
>     value=10
>     # engine will check for add=, sub=, multiply= or divide= affecting the value
> [/filter_ability]
> ```
> 
> we could have
> 
> ```
> [filter_ability]
>     value=10
>     {FILTER_ABILITY_DO_NOT_MATCH_MODIFIED_VALUE}
> [/filter_ability]
> ```
> 
> That means this logic would only be used when the WML author intends to use it, and a macro is much easier to remove than engine functionality.

I removed part of code for check existence of attrbute without specify value

newfrenchy83:
> Just realised that this could be done in WML, which would fit with the idea of making it an experimental API that could change in future.
> 
> Instead of
> 
> ```
> [filter_ability]
>     value=10
>     # engine will check for add=, sub=, multiply= or divide= affecting the value
> [/filter_ability]
> ```
> 
> we could have
> 
> ```
> [filter_ability]
>     value=10
>     {FILTER_ABILITY_DO_NOT_MATCH_MODIFIED_VALUE}
> [/filter_ability]
> ```
> 
> That means this logic would only be used when the WML author intends to use it, and a macro is much easier to remove than engine functionality.

Checking the presence or not of the add, sub etc attributes is only relevant in the case where the ability does not have a value= specified, otherwise we simply check if value has the desired value and if any other attribute digital are present within the same capacity, they must be checked separately. In the absence of value= we have two cases which arise, either the ability has a default value and then we check that other numerical attributes are not used with a value changing the default value; either the ability has no default value in the engine and the check returns a false value.

This can be done in wml, because otherwise the add, sub etc presence check would be done even if the ability has a specified value

stevecotton:
Sorry, I can't understand what that big block of text is saying.

Responding to [the comment from @soliton- that this PR is for](https://github.com/wesnoth/wesnoth/issues/7992#issuecomment-1795270245), I'm thinking that this filter:
```
[filter_ability]
    tag_name=drains
    value=50
    {FILTER_ABILITY_DO_NOT_MATCH_MODIFIED_VALUE}
[/filter_ability]
```
would match `[drains]` with default values, with explicit `value=50`, with `value,multiply=50,1` but would not match `value,multiply=50,2`. That set of what it matches seems like a useful set to me.

That wouldn't need an engine change in this PR, just a definition of the macro; and given the general issues in #7992 about labeling this API as experimental, I'm not even sure whether this macro should be added.
```
#define FILTER_ABILITY_DO_NOT_MATCH_MODIFIED_VALUE
[not]
    multiply=-infinity-0.99,1.01-infinity
[/not]
...same for add, sub, divide
#enddef
```

Pentarctagon:
Honestly at this point I'd say to just rename [filter_ability] to [experimental_filter_ability], leave it at that that for 1.18, and for 1.19 we need to create some sort of matrix or spreadsheet clearly defining what the expected result of various combinations are. It feels like we're spending a lot of time to make not a lot of progress currently without that.

newfrenchy83:
> Honestly at this point I'd say to just rename [filter_ability] to [experimental_filter_ability], leave it at that that for 1.18, and for 1.19 we need to create some sort of matrix or spreadsheet clearly defining what the expected result of various combinations are. It feels like we're spending a lot of time to make not a lot of progress currently without that.

done in https://github.com/wesnoth/wesnoth/pull/8050

PR #8047, opened 2023-11-14
UtBS S02: Beach Sand is a dehydrating terrain

newfrenchy83:
all terrain sand should be deshydrating terrain.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/utils/dehydration-utils.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/utils/dehydration-utils.cfg
@@ -4,7 +4,7 @@
 Ww*, *^V* #enddef
 
 #define DEHYDRATING_TILES
-Hd, Dd*, Dd^E*, Rd #enddef
+Hd, Dd*, Ds*, Dd^E*, Ds^E*, Rd #enddef
 
 #define DEHYDRATABLE_UNITS_FILTER
     [not]


stevecotton:
What testing have you done? It looks correct, but the PR description suggests that you're unsure whether it fixes the bug.

For the Git commit message, I suggest "UtBS S02: Beach Sand is a dehydrating terrain". Just putting a Github issue number in the commit message makes it hard to read the history in future.

newfrenchy83:
> What testing have you done? It looks correct, but the PR description suggests that you're unsure whether it fixes the bug.
> 
> For the Git commit message, I suggest "UtBS S02: Beach Sand is a dehydrating terrain". Just putting a Github issue number in the commit message makes it hard to read the history in future.

i changed message, if it is fix an issue,it's best.

newfrenchy83:
> What testing have you done? It looks correct, but the PR description suggests that you're unsure whether it fixes the bug.
> 
> For the Git commit message, I suggest "UtBS S02: Beach Sand is a dehydrating terrain". Just putting a Github issue number in the commit message makes it hard to read the history in future.

i not done testing, i just fixed terrain filter for what deshydrating effectapplied in sable sand because it seem what it is using of this terrain in map but no in filter who cause bug

stevecotton:
One of the reasons that I'm asking for testing is that, without testing, we don't know if there's another bug in addition to the one being fixed. If that other bug exists, then when someone starts debugging they'll think "this release must have been okay, otherwise the bug would have been found while testing, so it must have been introduced after that", which makes debugging much more time consuming and frustrating.

newfrenchy83:
> One of the reasons that I'm asking for testing is that, without testing, we don't know if there's another bug in addition to the one being fixed. If that other bug exists, then when someone starts debugging they'll think "this release must have been okay, otherwise the bug would have been found while testing, so it must have been introduced after that", which makes debugging much more time consuming and frustrating.

I just did a test, and once the filter is changed the units on the beach sand suffer the effects of thirst, and the healers cannot restore their life points during the day although they can outside the sand and the road (green plains), and before that this beach sand was not considered a dehydrating terrain and the trainers could do their job because this terrain did not aggravate thirst as it should. If the bug mentioned concerns the care taken when it is not supposed to happen then this repair is a success.

PR #8049, opened 2023-11-15
Solution to #7768 (No Label removing for other players in multiplayer)

ecomitz:
In issue #7768 we see an issue where if you delete a label by emptying its contents, it is deleted for you, but not other players in multiplayer. I attempted to remedy this issue this way:

I first examined the issue and discovered that the game currently has no way of deleting a single label - which, can be considered a separate issue per se. I resolved to then try to kill two birds with one stone - but first only resolving the first issue, and starting to carve the path for the second (An option to simply remove a single label)

First, I learned how the game maintains these things - a config file is created and maintained by the multiplayer server being run, and can accept commands to manipulate the file for the scenario maintaining the labels for everyone. 

I started first with labels themselves - their own ability to delete and find labels innately. 

I then looked at the recorder - the thing responsible for recording things that happen in the game, and told it how to record and send the command to the server, and implemented it into menu events

I taught the server to recognize "clear_label" (a single label) as a command. 

I tested this by running the boost test suite, and I used two computers with a locally built server, created labels, modified them without deletion to test whether or not that still works, cleared all labels to test whether or not that still works, and I emptied labels to see if the other player would still see it, and voila, the other player no longer sees the other label. 

Please feel free to ask any questions, comments, or if I need to change anything else, also feel free to let me know. If you folks would like a video of the aforementioned test, I can do it when I get home today from work. Thank you.

### DIFF
--- a/src/map/label.cpp
+++ b/src/map/label.cpp
@@ -223,6 +223,21 @@ void map_labels::clear(const std::string& team_name, bool force)
 	categories_dirty = true;
 }
 
+void map_labels::clear_label(const terrain_label* label)
+{
+	if (label == nullptr)
+	{
+		return;
+	}
+	team_label_map::iterator i = labels_.find(label->team_name());
+
+	label_map& p = i->second;
+
+	p.erase(label->location());
+
+
+}
+
 void map_labels::clear_map(label_map& m, bool force)
 {
 	label_map::iterator i = m.begin();
--- a/src/map/label.hpp
+++ b/src/map/label.hpp
@@ -71,6 +71,7 @@ public:
 	void enable(bool is_enabled);
 
 	void clear(const std::string&, bool force);
+	void clear_label(const terrain_label* label);
 
 	void recalculate_labels();
 	void recalculate_shroud();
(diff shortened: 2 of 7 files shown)


soliton-:
The flatpak check failure is not related to this PR but the trailing whitespace issues should be fixed.

On a brief look the code seems promising to me.

I'd leave the `menu_handler::clear_label()` stub out from this PR. That can be implemented later.

ecomitz:
Okay - I'll see to those back at my keyboard

ecomitz:
Installed a Ubuntu virtual box today,so I should be able to run the white space checker script independently now. Will push another commit when ready

ecomitz:
That last one was my bad - tried to update my own copy and didn't realize it would also send my cur copy over here

ecomitz:
Alright - pretty sure that should be the last of the whitespaces - if not, I might have to close this one and just go at it for a while before reopening

ecomitz:
Here's the video for this pull request: https://www.youtube.com/watch?v=lGcu1mCoqD4

PR #8050, opened 2023-11-15
rename [filter_ability(_active)] to experiemental_ability

newfrenchy83:
because it is experimental the tag of filter is changed except for [overwrite][filter_specials]

a bug is fixed also for case or ability diffrent of drains, berserk etc in checking 'value' attribute

### DIFF
--- a/data/schema/filters/unit.cfg
+++ b/data/schema/filters/unit.cfg
@@ -47,8 +47,8 @@
 	{FILTER_TAG "filter_adjacent" adjacent ()}
 	{FILTER_TAG "filter_location" location ()}
 	{FILTER_TAG "filter_side" side ()}
-	{FILTER_TAG "filter_ability" abilities ()}
-	{FILTER_TAG "filter_ability_active" abilities ()}
+	{FILTER_TAG "experimental_filter_ability" abilities ()}
+	{FILTER_TAG "experimental_filter_ability_active" abilities ()}
 	{FILTER_BOOLEAN_OPS unit}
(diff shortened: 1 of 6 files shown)


stevecotton:
> a bug is fixed also for case or ability diffrent of drains, berserk etc in checking 'value' attribute

Please make this a separate PR, instead of having feature-creep.

stevecotton:
I think this is ready to merge.

As mentioned in #7992, there's an open question about what to do with this deprecation message; however I think it's better to merge as-is and discuss what's deprecated in the issue ticket instead.
```
deprecated_message("[effect]apply_to=remove_ability [abilities]", DEP_LEVEL::INDEFINITE, "",
    "Use [filter_ability] instead in [effect]apply_to=remove_ability");
````

newfrenchy83:
> I think this is ready to merge.
> 
> As mentioned in #7992, there's an open question about what to do with this deprecation message; however I think it's better to merge as-is and discuss what's deprecated in the issue ticket instead.
> 
> ```
> deprecated_message("[effect]apply_to=remove_ability [abilities]", DEP_LEVEL::INDEFINITE, "",
>     "Use [filter_ability] instead in [effect]apply_to=remove_ability");
> ```

changed message for now, message changed anew when 'experimental' will be removed.

newfrenchy83:
@stevecotton deprecated message removed

newfrenchy83:
> Has the wiki been updated as well?

yeah, i already updated wiki for rename.

PR #8051, opened 2023-11-15
 remove checking for the presence of add, sub etc. attributes when check value.

newfrenchy83:
For me (and this is not the opinion of @soliton-) when we check the value of the value attribute,
 we seek to verify that the value encoded in ability (or possibly the default) corresponds,
 the value by default means that the value attribute is encoded with this default value, that's it.
 Apparently @soliton- understood that it was the returned value that was checked and
 I let myself get into this error which I am now correcting.

currently absence of default value is only detected when other numeric attributes are used, this commit is here to repair this rather unfortunate oversight.

### DIFF
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -1472,20 +1472,17 @@ static bool matches_ability_filter(const config & cfg, const std::string& tag_na
 		return false;
 
 	if(!filter["value"].empty()){
-		bool has_other_key = (!cfg["add"].empty() || !cfg["sub"].empty() || !cfg["multiply"].empty() || !cfg["divide"].empty());
-		if(!has_other_key){
-			if(tag_name == "drains"){
-				if(!int_matches_if_present(filter, cfg, "value", 50)){
-					return false;
-				}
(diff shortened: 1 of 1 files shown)


stevecotton (approved):
I think it's an improvement, and should be taken even while there's a discussion about the code surrounding this. It makes the current behavior better, and it's no extra effort to remove it if we decide to remove the surrounding code too.

stevecotton:
I know we complained when you opened new PRs instead of editing old ones, but in this case the force-pushed commit looks like a completely different change. Please either (make a new PR,) or (edit the title, and add a comment on the website), and then ask @soliton- for a review.

newfrenchy83:
> currently absence of default value is only detected when other numeric attributes are used, this commit is here to repair this rather unfortunate oversight.



> I know we complained when you opened new PRs instead of editing old ones, but in this case the force-pushed commit looks like a completely different change. Please either (make a new PR,) or (edit the title, and add a comment on the website), and then ask @soliton- for a review.

@soliton- could you review this PR?

soliton- (approved):

PR #8052, opened 2023-11-17
[filter_specials] -> [experimental_filter_specials]

Pentarctagon:
Since it calls into the same code as [filter_ability] did.

### DIFF
--- a/data/schema/units/specials.cfg
+++ b/data/schema/units/specials.cfg
@@ -57,7 +57,7 @@
 	{DEFAULT_KEY overwrite_specials ability_overwrite none}
 	[tag]
 		name="overwrite"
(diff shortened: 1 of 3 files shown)


newfrenchy83:
@Pentarctagon

newfrenchy83:
@Pentarctagon I hesitate to keep the term experiemental in the tags because once the debates are concluded, we will have to return to [filter_ability] and in the meantime certain wml authors will have started to use these filters which will force them to change the names, perhaps it would be worth It is better not to rename them and only indicate the experimental characters in the wiki and the changelog.

newfrenchy83:
Ok, then.

newfrenchy83:
@Pentarctagon i'm fine with it.

PR #8053, opened 2023-11-17
reverse change to filter_ability ->experimental_filter_ability

newfrenchy83:
By changing the names of filter_ability to experimental_filter_ability, I think I'm making a mistake. By doing this, we will force the authors who use these filters to change the names when we exit the "experimental" phase and I don't like that, it would be better to just indicate it in the changelog and the wiki.


soliton-:
Making it explicitly clear that someone using the experimental API will have to change something in the future is exactly the point.

IMO there is basically no chance filter_ability_active will stay for example or that filter_ability will behave the same as now. So that should be made very clear to anyone using these tags whether they check the documentation or not.

PR #8055, opened 2023-11-18
WoF S6: weaken player allies

Jonathan-Kelly:
A small balance patch based on feedback from @Hejnewar over [here.](https://github.com/wesnoth/wesnoth/pull/8031#issuecomment-1797917693)

### DIFF
--- a/data/campaigns/Winds_of_Fate/scenarios/06_Landfall.cfg
+++ b/data/campaigns/Winds_of_Fate/scenarios/06_Landfall.cfg
@@ -344,8 +344,8 @@ It will call in our reinforcements."
             [/have_location]
         [/filter_condition]
         # Karron’s Wing
-        {UNIT 3 (Sky Drake)      3 11 (facing=se)}
-        {UNIT 3 (Fire Drake)     4 11 (facing=se)}
+        {UNIT 3 (Sky Drake)      2 11 (facing=se)}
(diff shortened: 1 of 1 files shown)


max-torch:
I would like to share a replay of my playthrough of this scenario before this PR:
[WoF-Landfall replay 20231104-232032.gz](https://github.com/wesnoth/wesnoth/files/13406760/WoF-Landfall.replay.20231104-232032.gz)
The thing of note is that the allies take care of the objective without assistance from me. However, this is because i have sufficiently weakened the enemy force by calling the allies in late.

Jonathan-Kelly:
Thanks for the replay. Your tactics are good and essentially what was intended for the player to have to do at the higher difficulties. I think that after this PR you will get similar results as you did, but when playing on the second hardest difficulty.

The issue Hejnewar discovered is that setting off the reinforcements on turn 2 and rushing ahead with your own forces is (too) effective: [WoF-Landfall_replay_20231107-Hejnewar.gz](https://github.com/wesnoth/wesnoth/files/13450680/WoF-Landfall_replay_20231107-Hejnewar.gz)

Against this strategy the AI enemy is so surrounded that it cannot threaten either Karron or the player controlled characters sufficiently to make the scenario challenging for players who typically play on "Nightmare" difficulty. I think this PR might fix that and definitely improves it by making the AI allies a good deal weaker.

PR #8056, opened 2023-11-19
UtBS: Mention the campaign mechanics in the campaign description

stevecotton:
Let players see what to expect when looking through the campaign menu, instead of only giving players this information after they start the campaign and have no clear or quick way to return to the main or campaign selection menus.

The cherry-picks text from both the commits and the review comments of #8027, and fixes part (but not all) of #8009.

The preceding translatable string includes trailing newlines, but is the same as 1.16. So I've left that as-is, while using a separate non-translatable string for the new string's newlines.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/_main.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/_main.cfg
@@ -19,6 +19,8 @@
 
     description= _ "In the distant future a small band of elves struggles to survive amidst the ruins of fallen empires. Lead your people out of the desert on an epic journey to find a new home.
 
+" + _ "Note: This campaign changes some standard mechanics, such as the day/night cycle, recall costs and general approach to the difficulty levels. The Quenoth elves include units that can attack several times per turn, and others that can move after attacking." + "
+
 " + _"(Expert level, 10 scenarios.)"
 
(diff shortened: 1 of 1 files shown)


Jonathan-Kelly:
I think UtBS also [increases unit recruit costs](https://github.com/wesnoth/wesnoth/blob/master/data/campaigns/Under_the_Burning_Suns/utils/macros.cfg#L48-L75) as scenarios progress or based on a unit counter of some kind? Also there is a fairly significant RPG style experience system for a number of the main characters.

Not sure which unusual mechanics are worth mentioning in the description versus leaving vague, since this particular campaign has so many of them.

stevecotton:
While there are a lot of unusual mechanics, the multiple attacks are available in the first few turns, and a player who doesn't notice is at a significant disadvantage.

Also, at this point in the string freeze, I think the options are (post an exact proposed text) or (the status quo).

Jonathan-Kelly:
Fair points; do whatever you think is best. Further refinements can wait for 1.19.

stevecotton:
@Pentarctagon I think this is ready to merge, but who should review?

Pentarctagon:
That'd be @nemaara, but it looks fine to me at least.

stevecotton:
Merging now, sorry for adding a reviewer unreasonably late.

PR #8066, opened 2023-11-24
Add functionality to Time Schedule Editor

babaissarkar:
Changes :
1. Pressing the OK button in the Time Schedule Editor in the Map Editor now modifies the current schedule.
2. `Save Scenario As` menu item now writes the current assigned schedule to `utils/schedule.cfg`.
3. `color_slider_callback()` has new parameter to identify slider. Previously with one callback, the g and b sliders in the ui were not updating their values.
4. New `Description` textbox added in `custom_tod` dialog that allows to view and modify the `description` key of the `[time]` tag.
5. The Time Schedule button/hotkey now disabled unless the user is editing a scenario.
6. Preview button for color sliders. Map color changes can be previewed without closing and reopening the dialog.
7. WML translatable attributes written with leading underscore '_'.


src/editor/map/map_context.cpp
```diff
@@ -777,7 +777,46 @@ config map_context::to_config()
...
+	for(const config& time : times.child_range("time")) {
+		config& t = editor_times.add_child("time");
+		t.append(time);
+	}
+
+	/* Write to file */
+	try {
+		std::stringstream wml_stream;
+		wml_stream
+			<< "# This file was generated using the scenario editor.\n"
```
stevecotton, line 795: Too much copied from `save_scenario`?

stevecotton:
Note about translatable strings - this reuses an already-translatable string, `Description:` there's no new ones.
:+1:

src/editor/map/map_context.cpp
babaissarkar: Yes, I'm following how `save_scenario` is written to a file. How can I improve or simplify this?

stevecotton: Please ignore my comment - somehow I missed that this was writing a separate file, assumed that the schedule was being written to the same file as the scenario, and thus that this would duplicate the warning that was already in the same file.

babaissarkar: Okay.
A couple of questions.
1. Are the braces around the config writer necessary.
2. Is the if block really necessary? since wml_stream can never actually be empty due to the "#Editor" stuff.
`if(!wml_stream.str().empty()) {
			filesystem::write_file(filesystem::get_current_editor_dir(addon_id_) + "/utils/schedule.cfg", wml_stream.str());
}`

src/editor/map/map_context.cpp
```diff
@@ -777,7 +777,46 @@ config map_context::to_config()
...
+		wml_stream << "#endif";
+
+		if(!wml_stream.str().empty()) {
+			filesystem::write_file(filesystem::get_current_editor_dir(addon_id_) + "/utils/schedule.cfg", wml_stream.str());
+		}
+
+	} catch(const filesystem::io_exception& e) {
+		utils::string_map symbols;
+		symbols["msg"] = e.what();
+		const std::string msg = VGETTEXT("Could not save current schedule: $msg", symbols);
```
Pentarctagon, line 813: @stevecotton this would be a new string, wouldn't it?

Pentarctagon:
Out of curiosity, does this write out the translatable attributes with the leading underscore?

doofus-01:
> The Time Schedule button/hotkey now disabled unless the user is editing a scenario.

Just wondering, would it be difficult to have the schedule button either pull up existing schedules or temporarily show whatever tint was chosen in the dialog, but not save anything?  It would be convenient for testing maps at various times of day, either for the map design or terrain graphics drawing.

I could swear it used to load predefined schedules, but I'm probably misremembering.

babaissarkar:
> Out of curiosity, does this write out the translatable attributes with the leading underscore?

Probably not yet. I'll see if I can fix it.

babaissarkar:
> > The Time Schedule button/hotkey now disabled unless the user is editing a scenario.
> 
> Just wondering, would it be difficult to have the schedule button either pull up existing schedules or temporarily show whatever tint was chosen in the dialog, but not save anything? It would be convenient for testing maps at various times of day, either for the map design or terrain graphics drawing.
> 
> I could swear it used to load predefined schedules, but I'm probably misremembering.

Adding preview for the tints should be easy. I'm not so sure about schedule, but I'll look into it.

babaissarkar:
> I could swear it used to load predefined schedules, but I'm probably misremembering.
@doofus-01 It did, if you choose a schedule in the Assign Time Schedule item after clicking the Time button.

src/editor/map/map_context.cpp
CelticMinstrel: 1. It looks like it's technically unnecessary since you're writing without compression, but I think it's probably better to include them if you want to use `config_writer`. There is also the possibility of directly calling the global `write` function though, in which case it wouldn't make any sense to have braces.
2. Yeah, you seem to be correct that the stream will never be empty. However, I wonder if the intent of that check is more about whether the config writer wrote anything? After all, writing a file that just contains the warning and an empty `#ifdef` block isn't exactly useful.

src/gui/dialogs/editor/custom_tod.cpp
```diff
@@ -129,15 +129,15 @@ void custom_tod::pre_show(window& window)
...
 
 	connect_signal_notify_modified(
 			*(color_field_g_->get_widget()),
-			std::bind(&custom_tod::color_slider_callback, this));
+			std::bind(&custom_tod::color_slider_g_callback, this));
 
 	connect_signal_notify_modified(
 			*(color_field_b_->get_widget()),
-			std::bind(&custom_tod::color_slider_callback, this));
+			std::bind(&custom_tod::color_slider_b_callback, this));
```
CelticMinstrel, line 145: I get the feeling that these could all use the same callback, but with different bound parameters. Does that seem worthwhile? I guess it would require the use of a pointer to member, so it's not clear whether it's an improvement…

src/gui/dialogs/editor/custom_tod.cpp
```diff
@@ -278,13 +295,38 @@ void custom_tod::copy_to_clipboard_callback(tod_attribute_getter getter)
...
 void custom_tod::post_show(window& /*window*/)
 {
 	update_tod_display();
 
+	/*
 	if(get_retval() == retval::OK) {
 		// TODO: save ToD
-	}
+
+	}*/
```
CelticMinstrel, line 347: If this code is not needed anymore, it should be deleted.

src/editor/map/map_context.cpp
babaissarkar: 1. Okay, they can stay.
2. Maybe the check can be modified, so that it doesn't writes anything if there's no schedule? That is, the warning and empty `#ifdef`won't be written if there's no `[time]` block.

src/gui/dialogs/editor/custom_tod.cpp
babaissarkar: Maybe just add one more parameter to `color_slider_callback`? Like `std::bind(&custom_tod::color_slider_callback, this, "r")`? I'll test it. But it might fail if `connect_signal_notify_modified()` cannot attach the same function to multiple widgets.

CelticMinstrel: That wasn't precisely what I had in mind, but it's similar enough.

babaissarkar: Merged to one callback with parameter. Please see latest commit.

src/editor/map/map_context.cpp
babaissarkar: Probably a issue should be created as to why the "_" is not being written to the file.

data/schema/game_config.cfg
babaissarkar, line 1: Single tab replaced with 4 spaces, as per the convention inside the file.

src/editor/map/map_context.cpp
CelticMinstrel: Did my comments from the chat not help? With the current code I think it's correct that the _ is not being written.

babaissarkar: > Did my comments from the chat not help? With the current code I think it's correct that the _ is not being written.

No, you explained it well. But the problem is the `config_writer` and the related files are quite complicated for me to understand yet, so I haven't managed to figure out exactly what is stopping the '_' from being written.

babaissarkar: > so I haven't managed to figure out exactly what is stopping the '_' from being written.
Finally found it out. The right hand sides of assignments of keys (`cfg["name"] = name`) needs to be changed to (`cfg["name"] = t_string(name, textdomain)`). This problem exists in `map_context.cpp` and any other places where `config` objects are being created.

CelticMinstrel: I think I said the same thing in chat, didn't I? What are you using for the textdomain though?

babaissarkar: > I think I said the same thing in chat, didn't I?
Yes, took me a while to understand that `config_writer` isn't causing the problem and it's located elsewhere. Got confused.

CelticMinstrel:
Where's the code that writes the addon's _main.cfg? Does it write a `[textdomain]` tag?

babaissarkar:
> Where's the code that writes the addon's _main.cfg? Does it write a `[textdomain]` tag?
@CelticMinstrel  The code that writes the _main.cfg is in `editor_main.cpp`. The line you're looking for is
`*stream << "#textdomain wesnoth-" << addon_id << "\n"`
The textdomain is same as the one used by me.

CelticMinstrel:
I see it. Technically the lines I was looking for were the ones just after that line, but either way it looks like the `[textdomain]` is written so all is fine.

src/config_attribute_value.hpp
```diff
@@ -140,7 +140,7 @@ class config_attribute_value
 	config_attribute_value& operator=(const t_string &v);
 
 	/** Calls @ref operator=(const std::string&) if @a v is not empty. */
-	void write_if_not_empty(const std::string& v);
+	void write_if_not_empty(const t_string& v);
```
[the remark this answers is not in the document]
CelticMinstrel: I mean, I guess it works, but… the function is only 3 lines so I don't really see such a problem with just duplicating the same 3 lines, which will mean that it ends up stored as a string instead of a t_string in the variant… though I don't know if there's a real advantage to that or not. I suppose if there's an advantage it might be related to memory, since a t_string just is a pointer, whereas a sufficiently short string is stored inline.

data/gui/window/custom_tod.cfg
```diff
@@ -418,6 +446,19 @@
...
+							[column]
+								grow_factor = 0
+								border = "all"
+								border_size = 5
+								horizontal_alignment = "right"
+
+								[button]
+									id = "preview"
+									definition = "default"
+									label = _ "Preview"
```
stevecotton, line 459: That's a new translatable string :(

src/editor/map/map_context.cpp
stevecotton: Yes, it's a new string. Marking this as unresolved.

babaissarkar:
@Pentarctagon New Strings replaced with existing ones, see fcd7e8063b5e79411c1a328731a3749bc0c0399a.

babaissarkar:
Changes in fff021ab37b914bb8cbbfa670afa87f255f954e9
1. If you press OK on the schedule editor, the schedule is written and applied to the map. It also becomes available in the Assign Schedule menu.
2. If the user quits and reopens map editor, the new schedule is *not* visible in the Assign Schedule menu unless the user reloads the cache earlier with F5.
3. There's a new dialog (tod_new_schedule) to enter the new schedule's ID and name after pressing OK in the schedule editor.
4. If the user doesn't supply a new ID for the schedule, it will be written, but map editor refuses to load it because of duplicate ID clash with supplied schedules.
5. There are no new translatable strings (I reused "ID:" and "Name:")
6. Magnifying glass picture used for Display button in Time Schedule Editor.
7. Prompt Box used to take input form the user in New Scenario > New Add-on.

src/editor/controller/editor_controller.cpp
```diff
@@ -245,13 +246,29 @@ void editor_controller::custom_tods_dialog()
...
+		get_current_map_context().save_schedule(sch_id, sch_name);
+		gui_->update_tod();
+		context_manager_->refresh_all();
 	} else {
 		/* Restore old schedule */
 		update_map_schedule(prev_schedule);
 	}
+
+	//gui_->update_tod();
+	//context_manager_->refresh_all();
```
CelticMinstrel, line 271: Why is this commented code left in?

CelticMinstrel:
Is there other precedent for a widget definition being declared in the same file as the dialog that uses it?

src/editor/controller/editor_controller.cpp
babaissarkar: Will be removed.

babaissarkar:
> Would it be a huge amount of effort to fix (4) so that they can't click the ok button until they've entered an ID?

@Pentarctagon I'm fixing (4) by auto generating a unique custom id and name for the schedule, similar to the automatic Add-on id.

.gitignore
```diff
@@ -219,3 +219,58 @@ config.h
...
+/build/
+/build/
+/build/
+/build/
+/build/
+/build/
+/build/
+/build/
+/build/
+/build/
```
CelticMinstrel, line 276: What on earth is this?

babaissarkar: See 424a0cff3a65a201e5f7c449b43c0df0ae08274e. Should be cleaned up.

src/editor/map/map_context.cpp
babaissarkar: Replaced with existing string (fcd7e8063b5e79411c1a328731a3749bc0c0399a)

data/gui/window/custom_tod.cfg
babaissarkar: Replaced with picture in recent commit.

Pentarctagon:
Any other issues that need to be addressed before merging?

babaissarkar:
> Any other issues that need to be addressed before merging?

@Pentarctagon No, there shouldn't be anything else.

Pentarctagon:
An entry should be added to https://github.com/wesnoth/wesnoth/tree/master/changelog_entries

Other than that, I'll merge this in a couple days if there are no other comments.

CelticMinstrel:
The changelog file should probably be named differently.

babaissarkar:
@CelticMinstrel File renamed to 8066_time_schedule_editor.md

src/editor/controller/editor_controller.cpp
```diff
@@ -231,13 +232,67 @@ void editor_controller::custom_tods_dialog()
...
+				std::bind(
+						&editor::editor_controller::update_map_schedule,
+						this,
+						std::placeholders::_1));
+	tod_dlg.register_callback(update_func);
+
+	/* Autogenerate schedule id and name in case the user doesn't give any. */
+	std::int64_t current_millis = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+	std::string sch_id = "custom-"+std::to_string(current_millis);
+	std::string sch_name = "Custom Schedule "+std::to_string(current_millis);
```
stevecotton, line 252: While we can't add translatable strings at the moment, this ought to be one.

src/editor/controller/editor_controller.cpp
```diff
@@ -231,13 +232,67 @@ void editor_controller::custom_tods_dialog()
...
+			update_map_schedule(prev_schedule);
+			return;
+		}
+
+		/* In case the ID or Name field is blank and user presses OK */
+		if (sch_id.empty()) {
+			sch_id = "custom-"+std::to_string(current_millis);
+		} else {
+			/* Check if the id entered is same as any of the existing ids
+			 * If so, replace */
```
stevecotton, line 269: Should at least have a TODO for better error handling; given the string freeze, it's probably not possible to implement better handling.

src/gui/dialogs/editor/custom_tod.hpp
```diff
@@ -70,10 +80,17 @@ class custom_tod : public modal_dialog
...
+	void preview_schedule();
+
 	template<custom_tod::string_pair(*fptr)(const time_of_day&)>
 	void select_file(const std::string& default_dir);
 
-	void color_slider_callback();
+	/* Callback for color sliders */
+	void color_slider_callback(COLOR_TYPE type);
+
+    /* Update map and schedule in realtime */
```
stevecotton, line 92: Indentation is off here

src/gui/dialogs/editor/tod_new_schedule.hpp
```diff
@@ -0,0 +1,50 @@
...
+namespace gui2
+{
+
+namespace dialogs
+{
+
+/**
+ * @ingroup GUIWindowDefinitionWML
+ *
+ * Dialog that takes new schedule ID and name from the player.
```
stevecotton, line 28: Mentioning that it doesn't contain the ToD's settings  would be helpful.

src/gui/dialogs/editor/custom_tod.hpp
babaissarkar: Adjusted.

src/gui/dialogs/editor/tod_new_schedule.hpp
babaissarkar: Added more information.

src/editor/controller/editor_controller.cpp
babaissarkar: I've added TODO for error handling.  (76f27ac3a123cad94a813f97cfd8ff9325ad72f7)

src/editor/controller/editor_controller.cpp
babaissarkar: I've added a TODO to mark these as translatable strings. (76f27ac3a123cad94a813f97cfd8ff9325ad72f7)

src/editor/editor_main.cpp
```diff
@@ -43,8 +44,13 @@ std::string initialize_addon()
 	}
 
 	if(addon_id == "///newaddon///") {
+		std::string& addon_id_new = addon_id;
 		std::int64_t current_millis = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
 		addon_id = "MyAwesomeAddon-"+std::to_string(current_millis);
```
babaissarkar, line 49: @stevecotton Perhaps addon_id also needs to be a translatable string?

soliton-: Why would the addon id be translatable?

babaissarkar: Looks like I misunderstood something. I thought [this comment](https://github.com/wesnoth/wesnoth/pull/8066#discussion_r1423573578) considered both schedule ID and name to be translatable, but actually it was just the schedule name that has to be translatable. Please ignore this comment.

babaissarkar:
I've implemented changes suggested by @soliton-  and @CelticMinstrel in https://github.com/wesnoth/wesnoth/pull/8066/commits/fb6d8c96dbf84e99af79a90d6af761aef8364328.
In particular :
1. OK button is disabled unless user enters a name for the custom schedule. Name is no longer generated.
2. In case the `name` attribute is empty, the id of the schedule in used in the Assign Schedule menu.

Pentarctagon:
Alright, I'll be merging this tomorrow then if there are no additional comments.

src/editor/controller/editor_controller.cpp
```diff
@@ -130,8 +132,7 @@ void editor_controller::init_tods(const game_config_view& game_config)
 		tods_map::iterator times = tods_.find(schedule_id);
 		if (times == tods_.end()) {
 			std::pair<tods_map::iterator, bool> new_times =
-				tods_.emplace(schedule_id, std::pair(schedule_name, std::vector<time_of_day>()));
-
+			tods_.emplace(schedule_id, std::pair(schedule_name, std::vector<time_of_day>()));
```
soliton-, line 135: This looks like an accidental change?

src/gui/dialogs/prompt.hpp
```diff
@@ -30,6 +30,7 @@ namespace dialogs
  * ------------------|---------------|---------|-----------
  * prompt_box        | text_box      |yes      |The text box to enter the value into.
  */
+
```
soliton-, line 33: Looks like another accidental change.

babaissarkar: Fixed (8bcca43af0f6bcc3eaf9060940e54708a60c5159).

src/editor/controller/editor_controller.cpp
babaissarkar: Fixed (8bcca43af0f6bcc3eaf9060940e54708a60c5159).

Pentarctagon:
Thanks for sticking with this!

Issue #7894, opened 2023-09-06
`--log-none=deprecation` doesn't override a previous `--debug` flag on the command line


cooljeanius:
### Game and System Information

- **Version:** 1.16.10
- **Downloaded from:** Steam
- **OS:** macOS Big Sur

### Description of the bug

As described on Discord/IRC, I have tried starting Wesnoth with both the `--debug` and `--log-none=deprecation` flags, and still get spammed with deprecation messages, even though I have requested not to see them.

### Steps to reproduce the behavior

1. Start Wesnoth from the command line with both the `--debug` and `--log-none=deprecation` flags
2. Observe console output

### Expected behavior

`--log-none=deprecation` should override a previous `--debug` flag.

ZBugH1:
Confirmed on my end. I'll take a look when I get some free time later this week

OmarElfouly:
Hello :)
I just wanted to confirm this issue has been resolved in #8076 
(p.s. Im new to contributing and on the hunt for my first issue so you might see me elsewhere in the repo )

soliton-:
This issue is not resolved. One general problem is that our command line handling mostly doesn't care for the given argument order.

OmarElfouly:
> This issue is not resolved. One general problem is that our command line handling mostly doesn't care for the given argument order.

Oh that's interesting. Does that mean that the output of `--debug --log-none=deprecation` should be different than  `--log-none=deprecation --debug`

soliton-:
That is what this issue is about, yes.

gfgtdf:
> That is what this issue is about, yes.

i strongly disagree here, explicitly specifying  --log-none=deprecatio should always overrride --debug, unrelated to the other in that they appear.

OmarElfouly:
> i strongly disagree here, explicitly specifying --log-none=deprecatio should always overrride --debug, unrelated to the other in that they appear.

I agree with this. My initial understanding is that the issue was about --log-none=deprecation being ignored when --debug is passed which should not happen no matter the order.

soliton-:
> i strongly disagree here, explicitly specifying --log-none=deprecatio should always overrride --debug, unrelated to the other in that they appear.

Well, I strongly disagree that argument order should not matter. Our command line handling is pretty annoying in that it does not most of the time.

[...]

PR #8068, opened 2023-11-25
Avoid enabling deprecation logging after disabling

ZBugH1:
Fixes issue #7894

Setting "log-none deprecation" causes the deprecation severity to be set to -1 (disabled). The problem was that debug mode being enabled changed the deprecation severity afterwards hence
enabling deprecation messages.

### DIFF
--- a/src/game_config.cpp
+++ b/src/game_config.cpp
@@ -91,18 +91,28 @@ bool
 const bool& debug = debug_impl;
 
 void set_debug(bool new_debug) {
+    int severity;
+    lg::get_log_domain_severity("deprecation", severity);
+    int delta_severity; // change in severity
+    switch(severity) {
+        case -1: // LG_DISABLED, hence don't change
+            delta_severity = 0;
+            break;
+        case 0: // LG_ERROR
+        case 1: // LG_WARN
+        case 2: // LG_INFO
+        case 3: // LG_DEBUG
+            delta_severity = 2;
+            break;
+        default: // undefined case, hence don't change
+            delta_severity = 0;
+    }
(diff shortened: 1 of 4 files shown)


src/game_config.cpp
```diff
@@ -91,18 +91,28 @@ bool
 const bool& debug = debug_impl;
 
 void set_debug(bool new_debug) {
+    int severity;
+    lg::get_log_domain_severity("deprecation", severity);
+    int delta_severity; // change in severity
+    switch(severity) {
+        case -1: // LG_DISABLED, hence don't change
```
Wedge009, line 98: Wouldn't it be better to use the severity enumeration directly instead of the hard-coded values together with comments?

ZBugH1: yeah I 100% agree! The issue is the severity enumeration isn't used anywhere yet (according to my IDE). I was going to try to create a separate pull request that replaces int with the enum everywhere

ZBugH1: @Wedge009 I have just realised that an issue with my current solution is that if the severity goes above 3 (which it can't currently) then my code won't reset it properly. This wouldn't be an issue if everything was using the enum. I have [pushed](https://github.com/wesnoth/wesnoth/pull/8068/commits/b0fa883182272bb2c975ff5e4b0302433306bd4e) a more robust solution that is probably better. What do you think?

ZBugH1: Closing this pull request due to enum change (see pull request #8076). I'll make a new pull request (that is also more clear) for this issue that that uses the Enum once #8076 gets merged

Issue #8071, opened 2023-11-26
Screen size variables lag behind changes in window / screen size 


ZBugH1:
### Game and System Information

**Version:** v1.17.23+dev x86_64
**built from source:** from commit 1fa8946eb12f6b9897776975706afdf330cf9109 using GCC 11.4
**OS:** Ubuntu 22.04 

### Description of the bug

The variables in question are `gui2::settings::screen_height` and `gui2::settings::screen_width` in `settings.cpp`. There is a question if they should be renamed to  `window_height` and  `window_width`? (since that is what they seem to represent) 

**The bug:** 
`screen_height` and `screen_width` don't always get immediately updated when the window size or screen size changes (ie. changing the monitor that Wesnoth is on).
If this is the intended behaviour, then I reckon it should be documented as such.

### Steps to reproduce the behavior

1. Start Wesnoth
2. Watch the variables `gui2::settings::screen_height` and `gui2::settings::screen_width`  either through debugging or by printing them out at sufficient times 
3. Disable full screen via the usual menu 
4. Change the window size by dragging the corner of the window
5. Observe that  `gui2::settings::screen_height` and `gui2::settings::screen_width` don't get updated immediately. They seem to get updated when you click the menu 

### Expected behavior

I think the variables `gui2::settings::screen_height` and `gui2::settings::screen_width` should be a source of truth. 

### Additional context

Discovered this behaviour while attempting the TODO in `mouse_handler::drag_threshold()` in `mouse-events.cpp`.

ZBugH1:
I haven't checked this yet, but I think the problem is that we are using `SDL_VIDEO_RESIZE` from SDL1.2 which is equivalent to `SDL_WINDOWEVENT_RESIZED` in SDL2. This doesn't trigger for 'internal' events, see https://github.com/wesnoth/wesnoth/commit/ac96df07bebbfc5d494f7325b89aa42ad4e567a9. Hence, the changes in screen size that I tested are maybe 'internal' events...

ZBugH1:
I have just realised that the variables `gui2::settings::screen_height` and `gui2::settings::screen_width` are set in two different places in the code. 

This one uses the call SDL_VIDEO_RESIZE ([SDL_WINDOWEVENT_RESIZED](https://wiki.libsdl.org/SDL2/SDL_WindowEventID)):
[...]

Issue #7955, opened 2023-10-16
Clean up mouse event code, disginguish "click" from button up


daniel-santos:
### Describe the desired feature

In working issue #7949, I've noticed that mouse up events are treated as mouse clicks and I've seen comments of where this has caused issues. I propose that this be changed to define a "mouse click" is when a mouse button goes down and up on the same logical target. So if the mouse goes down on hex tile 43, 21, but comes up on 43, 22, it can be considered a drag event (if more than `drag_threshold()` pixels), but not a click. But if it goes down and up on the same button, even if dragged a fair distance, then it's customary to give it a click event.

PR #8073, opened 2023-11-27
Adaptive drag threshold

ZBugH1:
Implemented TODO* in `mouse_handler::drag_threshold()`

Changes the drag threshold from 14 to be based on window resolution. Scaling is performed so that a 1080p window has the original threshold of 14 (since 1080p is currently the most popular monitor size and I assume people are mostly playing Wesnoth in full screen mode). 

*Note that I haven't used the physical screen size because I imagine that it would be impractical to maintain cross-platform, and probably it isn't a good measure of user perceived drag distance anyway (eg. you hold a phone much closer to your face than a laptop). 

The main use case for this is ensuring that on very high resolution displays the drag distance is still reasonable. 

Tangentially related to issue #7955
Closely related to issue #8071

### DIFF
--- a/src/mouse_events.cpp
+++ b/src/mouse_events.cpp
@@ -93,8 +93,12 @@ void mouse_handler::set_side(int side_number)
 
 int mouse_handler::drag_threshold() const
 {
-	// TODO: Use physical screen size.
-	return 14;
+    // Function uses window resolution as an estimate of users perception of distance
+    // Tune this variable if necessary:
+    const unsigned threshold_1080p = 14; // threshold number of pixels for 1080p
+    double screen_diagonal = std::hypot(gui2::settings::screen_width,gui2::settings::screen_height);
+    const double scale_factor = threshold_1080p / std::hypot(1080,1920);
+    return static_cast<int>(screen_diagonal * scale_factor);
 }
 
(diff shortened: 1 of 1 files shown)


src/mouse_events.cpp
```diff
@@ -93,8 +93,15 @@ void mouse_handler::set_side(int side_number)
...
 int mouse_handler::drag_threshold() const
 {
-	// TODO: Use physical screen size.
-	return 14;
+    // Function uses window resolution as an estimate of users perception of distance
+    // Tune this variable if necessary:
+    const unsigned threshold_1080p = 14; // threshold number of pixels for 1080p
+    auto l2_norm = [](unsigned x,unsigned y) {
+        return std::sqrt(std::pow(x,2) + std::pow(y,2));
+    };
```
Pentarctagon, line 101: Any reason to not just use `std::hypot`?

ZBugH1: My own ignorance :)  I have updated now

stevecotton:
The code looks good to me, the idea looks good too, but as someone who always uses the keyboard, scroll wheel or middle-mouse to scroll, I can't check whether it feels right.

Issue #8072, opened 2023-11-26
Wesnothd - Cannot Connect Multiplayer LAN


Quin5:
### Game and System Information

- **Version:**    1.16.2
- **Downloaded from:**    Synaptic Package Manager
- **Build info:**     n/a
- **OS:**     LinuxMint 21.2

### Description of the bug

I launch Wesnoth from my GUI menu, select "Multiplayer", "Host Networked Game", and click "Ok".  I receive a message about Wesnoth server running in the background and click "Yes".  From a separate computer, attempting to connect using the host IP address and port 15000 gives the message, "connection failed operation timed out".

Upon investigation, I found the following error when wesnothd is ran:

```
Started wesnothd as a daemon with process id 7051
20231126 10:43:13 info server: Server configuration from file: '/home/me/.config/wesnoth-1.16/lan_server.cfg' read.
20231126 10:43:13 error server: could not make fifo at '/var/run/wesnothd/socket' (No such file or directory)
20231126 10:43:14 info server: 127.0.0.1	player joined using accepted version 1.16.2:	telling them to log in.
```

There is no '/var/run/wesnothd' directory.  The GUI appears to make a networked host, but is using localhost instead, which naturally does not allow LAN users to connect.

### Steps to reproduce the behavior

1. Run Wesnoth
2. Host networked game
3. Connect to host IP from separate computer on same LAN

### Expected behavior

Users on my LAN being able to connect to my host.

### Additional context

The computer I'm using to connect to my host is running Mac OSX 14.12.

loonycyborg:
Indeed if we listen on loopback interface you could only connect more clients on same pc. I'm not sure what's the best way to figure out the ip you should listen on for LAN case.

loonycyborg:
In general each host can have multiple network interfaces each on different networks, so would need to add an extra text field to input pc's ip address on desired network, unless there's some shortcut.

loonycyborg:
Could make it listen on 0.0.0.0 but in general it's considered insecure. Would it be acceptable for LAN server?

soliton-:
The fifo error is not an issue. It's not needed.

Is there a firewall running on the host running wesnothd? Then you need to allow connecting to port 15000.

loonycyborg:
Double check that wesnothd it starts actually listens on all interfaces with something like `netstat -plant`. For me it listens on :: and 0.0.0.0 so should already work on LAN.

Quin5:
[...]

PR #8075, opened 2023-11-27
Make mp server from "Host Networked Game" listen on all interfaces

loonycyborg:
Fixes issue #8072


soliton-:
I don't really see how this commit does what it says. This changes what host the client is trying to connect to. Presumably breaking connecting to the local wesnothd instance.
wesnothd is already listening on `0.0.0.0` and `[::]` for IPv4 and IPv6 respectively.
https://github.com/wesnoth/wesnoth/blob/320bed81aff8ccf4495f2794ecbf8e23db624d12/src/server/common/server_base.cpp#L83-L87

soliton-:
Looks like connecting to `[::]` or `0.0.0.0` somehow connects to localhost.

PR #8076, opened 2023-11-27
lg::severity Enum refactoring

ZBugH1:
Refactored all relevant code to use `lg::severity` Enum instead of `int` type. 
A nice-to-have prerequisite for issue #7894 (not technically required but would be nice).

### DIFF
--- a/src/game_config.cpp
+++ b/src/game_config.cpp
@@ -91,17 +91,20 @@ bool
 const bool& debug = debug_impl;
 
 void set_debug(bool new_debug) {
+    // TODO: remove severity static casts and fix issue #7894
 	if(debug_impl && !new_debug) {
 		// Turning debug mode off; decrease deprecation severity
-		int severity;
+		lg::severity severity;
 		if(lg::get_log_domain_severity("deprecation", severity)) {
-			lg::set_log_domain_severity("deprecation", severity - 2);
+            int severityInt = static_cast<int>(severity);
+			lg::set_log_domain_severity("deprecation", static_cast<lg::severity>(severityInt - 2));
 		}
 	} else if(!debug_impl && new_debug) {
 		// Turning debug mode on; increase deprecation severity
-		int severity;
+        lg::severity severity;
 		if(lg::get_log_domain_severity("deprecation", severity)) {
-			lg::set_log_domain_severity("deprecation", severity + 2);
+            int severityInt = static_cast<int>(severity);
+			lg::set_log_domain_severity("deprecation", static_cast<lg::severity>(severityInt + 2));
 		}
 	}
 	debug_impl = new_debug;
--- a/src/log.hpp
+++ b/src/log.hpp
@@ -72,13 +72,15 @@ const std::string log_file_suffix = ".log";
 // Note that this count does not include the current log file!
 const unsigned max_logs = 8;
 
-enum severity
+enum class severity
 {
+    LG_NONE=-1,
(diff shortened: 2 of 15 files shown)


src/log.hpp
```diff
@@ -74,6 +74,7 @@ const unsigned max_logs = 8;
 
 enum severity
 {
+    LG_NONE=-1,
```
CelticMinstrel, line 77: It should now be possible to make this type an `enum class`, right?

stevecotton:
It's failing to build the Boost-based unit tests, it needs this change in 3 places in src/tests/test_unit_map.cpp
```
-	lg::set_log_domain_severity("engine", lg::err().get_severity() - 1); // Don't log anything
+	lg::set_log_domain_severity("engine", lg::LG_NONE); // Don't log anything
```

src/log.hpp
ZBugH1: I'll change this, fix the other issues and push a new version to the same commit

Pentarctagon:
CI is still failing.

src/log.hpp
cooljeanius: > It should now be possible to make this type an `enum class`, right? Although `lg::severity::LG_ERROR` has a bit of redundancy to it…

Depends, remind me which version of the C++ standard Wesnoth is on again?

src/game_config.cpp
```diff
@@ -91,17 +91,20 @@ bool
...
 void set_debug(bool new_debug) {
+    // TODO: remove severity static casts and fix issue #7894
 	if(debug_impl && !new_debug) {
 		// Turning debug mode off; decrease deprecation severity
-		int severity;
+		lg::severity severity;
 		if(lg::get_log_domain_severity("deprecation", severity)) {
-			lg::set_log_domain_severity("deprecation", severity - 2);
+            int severityInt = static_cast<int>(severity);
+			lg::set_log_domain_severity("deprecation", static_cast<lg::severity>(severityInt - 2));
```
Pentarctagon, line 100: I know this is existing, but is there a reason to do this math here instead of setting it to a specific `severity`?

src/log.hpp
Pentarctagon: C++17 (mostly)

Pentarctagon:
Looks good to me at least.

src/game_config.cpp
```diff
@@ -91,17 +91,20 @@ bool
...
 		}
 	} else if(!debug_impl && new_debug) {
 		// Turning debug mode on; increase deprecation severity
-		int severity;
+        lg::severity severity;
 		if(lg::get_log_domain_severity("deprecation", severity)) {
-			lg::set_log_domain_severity("deprecation", severity + 2);
+            int severityInt = static_cast<int>(severity);
+			lg::set_log_domain_severity("deprecation", static_cast<lg::severity>(severityInt + 2));
```
stevecotton, line 107: On the one hand, there's no range check here, and I think that might be Undefined Behavior https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1766

OTOH, it's only reachable via command-line arguments, and I assume there will be a follow-up to this PR which addresses the TODO above. I'd be okay with merging as-is.

Pentarctagon: IMO rather than a range check, just hardcoding them to specific log levels rather than doing math would be preferable. Doesn't need to be with this PR though.

PR #8077, opened 2023-11-28
Add a 'special_id_active' attribute to [overwrite] in case the 'experimental_filter_specials' filter is not accepted(postpone 1.19)

newfrenchy83:
Because of this risk I duplicate special_id_active in overwrite to be sure to maintain a way to remain selective on the abilities/specials to overwrite.


stevecotton:
Until 1.18 releases, I only want to look at what the API will be in 1.18.

However, afterwards I'd like to look at documentation rather than code. Start with a use-case, tell me what the player-visible abilities are that could use this, and why this is needed.

PR #8078, opened 2023-11-30
Solution to #7768. Added a way for players to change an existing label to an empty label. 

alesk852:
Hello! 
This is my solution to #7768.

I made a statement that checks if the player wants to change the name of an existing map label to an empty label. And then creates a label that is visually empty but won't be deleted and thereby will still be recorded.

I tested my solution by running two instances of the game, entered multiplayer and logged in with two different accounts. Then I created a game and tried all the different permutations that one can create a label and change it to an empty text.

This is my first time contributing to Westnoth and would really appreciate the feedback!

### DIFF
--- a/src/map/label.cpp
+++ b/src/map/label.cpp
@@ -168,13 +168,19 @@ const terrain_label* map_labels::set_label(const map_location& loc,
 	{
 		// Found old checking if need to erase it
 		if(text.str().empty()) {
+
+			current_label->second.update_info(
+				text, creator, tooltip, team_name, color, visible_in_fog, visible_in_shroud, immutable, category);
+
+			res = &current_label->second;
+
 			// Erase the old label.
 			current_label_map->second.erase(current_label);
 
 			// Restore the global label in the same spot, if any.
 			if(terrain_label* global_label = get_label_private(loc, "")) {
-				global_label->recalculate();
-			}
+			 	global_label->recalculate();
+			 }
 		} else {
 			current_label->second.update_info(
 				text, creator, tooltip, team_name, color, visible_in_fog, visible_in_shroud, immutable, category);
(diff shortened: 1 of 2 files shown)


soliton-:
I don't think not actually removing the label is a solution. Presumably that also does not work as expected if there are global and team labels on the same hex.

alesk852:
> I don't think not actually removing the label is a solution. Presumably that also does not work as expected if there are global and team labels on the same hex.

Yeah, you are right.  Is my latest commit a better solution?
I fail a test, but I saw someone in the Discord had a similar problem.

soliton-:
Presumably this ends up with a `[label]` with an empty text attribute in replay/send to remote? That effect is exactly what we want but pretty sure that is undefined behavior since you use the label after it is erased.

We could either have some static label object with empty text for this or manually construct the `[label]` WML similar to `replay::add_label` but without the label pointer.

soliton-:
Well, a static label is probably not great since we would have to constantly change its location.

alesk852:
> Well, a static label is probably not great since we would have to constantly change its location.

Okay, thank you. Then I will manually construct the label.

alesk852:
> We could either have some static label object with empty text for this or manually construct the `[label]` WML similar to `replay::add_label` but without the label pointer.

Hello!
I might have confused myself a bit, but isn't possible just to use the add_label function already in the file label?

soliton-:
You mean https://github.com/wesnoth/wesnoth/blob/0157c245eaae5596b95ec92ae44c98a56f0a83ae/src/map/label.cpp#L202-L209?

Not sure what you want to use it for.

alesk852:
> Not sure what you want to use it for.
Yes, I was thinking of that function.

The thought was to add an empty label and assign it to res in set_label, which is later returned and recorded in menu_events when set_label is called.

soliton-:
I don’t understand what this is supposed to accomplish. Now instead of removing the label we remove it and add a new one? What’s the point?

PR #8082, opened 2023-12-02
use game_events.add in on_event.lua

gfgtdf:
previously on_event.lua and game_events.add had seperate priority lists, so that unrelated of the priority parameter (which both game_events.add and on_event.lua now support) on_event.lua events were always excuted first. 

The set_undoable(false) call is there for compatibility with the old on_event implementation.

### DIFF
--- a/data/lua/on_event.lua
+++ b/data/lua/on_event.lua
@@ -2,6 +2,20 @@
 -- so you have to call this function from a toplevel lua tag or from a preload event.
 -- It is also not possible to use this for first_time_only=yes events.
 
+-- This api used to be the default way to add events from lua (with its own implementation in lua
+-- based on game_events.on_event). Now its just calls game_evets.add, in particular because otherwise the
+-- priority parammter wouldn't work accord the differnt implmentaions..
+-- Still kept now for compatibility and because it has an easier to
+-- use interface meaning you can easily write in a lua file:
+--
+-- on_event("moveto", 10, function(ec)
+--   ...
+-- end)
+--
+-- which is imo more convenient than the interace wesnoth.game_events.add or wesnoth.game_events.add_repeating offers
+-- even though its at this point technicially equivalent to the later.
+
+
 return function(eventname, priority, fcn)
 	if type(priority) == "function" then
 		fcn = priority
--- a/data/lua/on_event.lua
+++ b/data/lua/on_event.lua
@@ -7,7 +7,7 @@ return function(eventname, priority, fcn)
 		fcn = priority
 		priority = 0
 	end
-	
+
 	wesnoth.game_events.add{
 		name = eventname,
 		priority = priority,
--- a/data/lua/on_event.lua
+++ b/data/lua/on_event.lua
@@ -4,7 +4,7 @@
 
 -- This api used to be the default way to add events from lua (with its own implementation in lua
 -- based on game_events.on_event). Now it just calls game_evets.add, because otherwise the
--- priority parammter wouldn't work across the differnt implmentaions.
+-- priority parameter wouldn't work across the different implementations.
 -- Still kept for compatibility and because it has an easier to
 -- use interface. Meaning you can easily write in a lua file:
 --
@@ -12,8 +12,8 @@
 --   ...
 -- end)
 --
--- which is imo more convenient than the interace wesnoth.game_events.add or wesnoth.game_events.add_repeating offers
--- even though its at this point technically equivalent to the later.
+-- which is imo more convenient than the interface wesnoth.game_events.add or wesnoth.game_events.add_repeating offers
+-- even though its at this point technically equivalent to the latter.
 
 
 return function(eventname, priority, fcn)
--- a/data/lua/on_event.lua
+++ b/data/lua/on_event.lua
@@ -25,6 +25,7 @@ return function(eventname, priority, fcn)
 	wesnoth.game_events.add{
 		name = eventname,
 		priority = priority,
+		first_time_only = false,
 		action = function()
 			context = wesnoth.current.event_context
 			wesnoth.experimental.game_events.set_undoable(true)
--- a/data/lua/on_event.lua
+++ b/data/lua/on_event.lua
@@ -13,7 +13,7 @@
 -- end)
 --
 -- which is imo more convenient than the interace wesnoth.game_events.add or wesnoth.game_events.add_repeating offers
--- even though its at this point technicially equivalent to the later.
+-- even though its at this point technically equivalent to the later.
 
 
 return function(eventname, priority, fcn)
--- a/data/lua/on_event.lua
+++ b/data/lua/on_event.lua
@@ -2,54 +2,19 @@
 -- so you have to call this function from a toplevel lua tag or from a preload event.
 -- It is also not possible to use this for first_time_only=yes events.
 
-if rawget(_G, "core_on_event") then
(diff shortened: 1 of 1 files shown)


CelticMinstrel (commented):
Shouldn't this also deprecate on_event? There isn't much point in using it in new code after this change. See 857d7e120a5999d1093399884886e13b03fbe7ee for what the deprecation should look like (that commit removed it).

I also don't think this should go in 1.18. Deprecating something probably counts as an API change, and rewriting its implementation after feature freeze is probably not a good idea either.

gfgtdf:
> Shouldn't this also deprecate on_event?, I also don't think this should go in 1.18. Deprecating something probably counts as an API change

Well if deprecation is a reason for not putting this into 1.18 then this pr shouldn't deprecate it.


> and rewriting its implementation after feature freeze is probably not a good idea either.

imo its pretty safe, (assuming that game_events.add has no bugs), and the thing is, currently someone who wants to stay compatible with on_event (for making sure that events get exceuted before/after other events that use on_event ) also have to use on_event, so by leaving the current implementation we are forcing people who want to hack into mainline content (that uses on_event like wc2, dark forecast) to also use on_event (which you want to depreacte eventually)

Pentarctagon:
> The set_undoable(false) call is there for compatibility with the old on_event implementation.

`set_undoable` is set to true though?

CelticMinstrel:
> the thing is, currently someone who wants to stay compatible with on_event (for making sure that events get exceuted before/after other events that use on_event ) also has to use on_event, so by leaving the current implementation we are forcing people who want to hack into mainline content (that uses on_event like wc2, dark forecast) to also use on_event (which you want to depreacte eventually)

I don't quite understand what you're saying here…?

gfgtdf:
> Presumably the code in game_lua_kernel.cpp that calls on_event should still make the call, in case UMC Lua code has hooked that call.

I'm not really sure what you mean here, which call? on_event.lua offers no way  to manually invoke events, afaik.

stevecotton:
> I'm not really sure what you mean here, which call? on_event.lua offers no way to manually invoke events, afaik.

If I understand the code correctly, there are two functions commonly called `on_event`, which were both defined in `on_event.lua`.
* The one to register new handlers, which is the one returned by `wesnoth.require("on_event")`
* The event dispatcher, `wesnoth.game_events.on_event`, for which this PR removes the Lua code but leaves the C++ in `game_lua_kernel::run_event`

That event dispatcher has code to add itself to a daisy-chain of functions - if there's already a `wesnoth.game_events.on_event` then it'll store it in `old_on_event` and call the old function from the new function. UMC code can do the same trick to redirect the `wesnoth.game_events.on_event` call to UMC code, optionally calling the existing code from the UMC code.

Era_of_High_Sorcery/lua_files/theme-item-ui.cfg does that and runs code *before* calling the mainline `w.g.on_event` function, however it does seem to call though to the mainline function without filtering any calls out.

gfgtdf:
> If I understand the code correctly, there are two functions commonly called on_event, which were both defined in on_event.lua.

hmm no, `wesnoth.game_events.on_event` existed long before `on_event.lua` (before i joined the team actually) , `game_events.on_event` was added as a lua way to hook arbitrary events , and then later `on_event.lua` was added on top of it (implemented using `w.g.on_event`), to make adding lua events easier. I'm also pretty sure that the code of EohS was written before `on_event.lua` was added so it should still work.

What we should maybe do is to add a dummy empty function in `w.g.on_event` so that umc that uses `w.g.on_event` doesn't have to deal with the case that its nil (which could totally be the case when `w.g.on_event` was added tho)

stevecotton:
I'm just suggesting to leave the C++ as-is, so no change to the PR.

data/lua/on_event.lua
```diff
@@ -2,54 +2,19 @@
...
 	end
-	eventname = string.gsub(eventname, " ", "_")
-	event_handlers[eventname] = event_handlers[eventname] or {}
-	local eh = event_handlers[eventname]
-	table.insert(eh, { h = handler, p = priority})
-	-- prioritize last entry
-	for i = #eh - 1, 1, -1 do
-		if eh[i].p < eh[i + 1].p then
-			eh[i], eh[i + 1] = eh[i + 1], eh[i]
+
```
stevecotton, line 10: The CI failure is because of whitespace on this line.

data/lua/on_event.lua
```diff
@@ -2,54 +2,33 @@
...
-	local context = nil
-	for _, entry in pairs(event_handlers[eventname] or {}) do
-		if context == nil then
-			context = wesnoth.current.event_context
-		end
-		entry.h(context)
-	end
-end
+-- This api used to be the default way to add events from lua (with its own implementation in lua
+-- based on game_events.on_event). Now its just calls game_evets.add, in particular because otherwise the
```
CelticMinstrel, line 6: Typo: game_evets

data/lua/on_event.lua
```diff
@@ -2,54 +2,33 @@
...
+-- priority parammter wouldn't work accord the differnt implmentaions..
```
CelticMinstrel, line 7: This line is full of typos and also doesn't quite make sense.

data/lua/on_event.lua
```diff
@@ -2,54 +2,33 @@
...
+-- Still kept now for compatibility and because it has an easier to
+-- use interface meaning you can easily write in a lua file:
+--
+-- on_event("moveto", 10, function(ec)
+--   ...
+-- end)
+--
+-- which is imo more convenient than the interace wesnoth.game_events.add or wesnoth.game_events.add_repeating offers
+-- even though its at this point technicially equivalent to the later.
```
CelticMinstrel, line 16: Typo: "technically" has only one i

CelticMinstrel:
Couldn't this be implemented using `wesnoth.game_events.add_repeating` instead of `wesnoth.game_events.add`? Sine the former is what will eventually replace this, I think it would make more sense to use it.

gfgtdf:
> Couldn't this be implemented using wesnoth.game_events.add_repeating instead of wesnoth.game_events.add? Sine the former is what will eventually replace this, I think it would make more sense to use it.

I usually prefer using a syntax like the one of game_events.add with its explicit parameters in codes like like this since its a little easier to understand for the reader imo. (and its not a repeating pattern where the shortening of game_events.add_repeating would have the advantage.)

CelticMinstrel:
It still _could_ be implemented with `add_repeating` though, right? If you used `add_repeating` the overall effect would be the same?

gfgtdf:
> It still could be implemented with add_repeating though, right? If you used add_repeating the overall effect would be the same?

I didn't look into the detail, but i think so.

data/lua/on_event.lua
CelticMinstrel: You fixed another typo on this line but not the specific one I identified.

data/lua/on_event.lua
CelticMinstrel: Using "across" made the intended meaning clear, but please also correct the other spellings:

parameter
different
implementations

data/lua/on_event.lua
CelticMinstrel: Also, that last word should be "latter" with two T's. And you misspelled "interface" on the line above.

CelticMinstrel:
I think the only typo left is `game_evets.add`.

PR #8085, opened 2023-12-02
WoF: fix macro argument in translatable string

### LINKED ISSUE #8084: WoF macros: BOOK_LEARNING puts a macro in a translatable string

- **Version:** 1.17.23

### Description of the bug

This translatable string has a macro argument in it, which [makes the translation fail](https://wiki.wesnoth.org/GettextForWesnothDevelopers#Things_to_avoid):

```
#. [floating_text]
#: data/campaigns/Winds_of_Fate/utils/macros.cfg:164
msgid "+{AMOUNT} experience"
```

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Jonathan-Kelly:
This simply puts the raw macro argument in a temporary variable.

Resolves  #8084


CelticMinstrel (commented):
Seems fine. I think I'd suggest using a less generic variable name though, like `book_learning_amount` or something. Just to be on the safe side. That said, if you're certain the name `amount` isn't used anywhere else in the campaign, it's probably fine like this?

stevecotton (approved):
As a bugfix, it should be accepted even in the string freeze.

Adding a changelog entry would be good though.

Jonathan-Kelly:
> if you're certain the name amount isn't used anywhere else in the campaign, it's probably fine like this?

Short names are only ever used for very temporary variables - essentially local variables. Whereas long lasting variables always get long serpent_case names to avoid accidental reuse.

> Adding a changelog entry would be good though.

Should it be under `Campaigns` or `Miscellaneous and Bug Fixes`?

Jonathan-Kelly:
Alright, added a change log entry. If you like its wording then I think this is ready to merge.

PR #8089, opened 2023-12-04
1.18 - Sprite updates to Ravens and Ants

doofus-01:
Reviewing the ants and ravens.  This should make the ants more visually distinct from each other.  This screenshot shows where we are when PR is opened.
ants-ravens


doofus-01:
This ended up being more ant-focused than bird-focused, will get to those later
Screenshot_20231210_ants

Issue #8092, opened 2023-12-08
[resistance] without max_value


Pentarctagon:
When `max_value` is not specified, instead it simply sets all the unit's resistances to 0%.

The test scenario which does this:
```
{GENERIC_UNIT_TEST "resistance_value" (
    [event]
        name = start

        [modify_unit]
            [filter]
            [/filter]
            [effect]
                apply_to = new_ability
                [abilities]
                    [resistance]
                        id=steadfast1
                        name="hi"
                        value=50
                        add=50
                        affect_self=yes
                        #max_value=90
                    [/resistance]
                [/abilities]
            [/effect]
        [/modify_unit]

        [end_turn][/end_turn]
    [/event]

    [event]
        name = side 2 turn 1
        [end_turn][/end_turn]
    [/event]

    [event]
        name =side 1 turn 2
        [end_turn][/end_turn]
    [/event]

    [event]
        name = side 2 turn 2

        [store_unit]
            [filter]
                type = "Elvish Archer"
            [/filter]
            variable = temp
        [/store_unit]
        
        #{SUCCEED}
    [/event]
)}
```

If you add a `max_value`, then it works. I assume the issue is that if it's not present then it's defaulting `max_value` to 0.

ProditorMagnus:
Seems max_value default should be 100.

newfrenchy83:
> Seems max_value default should be 100.

If max_value must have default value when not precised, please confirm what should be 100 or other value.

PR #8093, opened 2023-12-08
Fix [resistance] without max_value bug

newfrenchy83:
Fix https://github.com/wesnoth/wesnoth/issues/8092 issue.

If none [resistance] ability of the list contain max_avlue, on default value of 100 is used instead.

### DIFF
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -1798,10 +1798,18 @@ int unit::resistance_against(const std::string& damage_name,bool attacker,const 
 	if(!resistance_abilities.empty()) {
 		unit_abilities::effect resist_effect(resistance_abilities, 100-res);
 
-		res = 100 - std::min<int>(
-			resist_effect.get_composite_value(),
-			resistance_abilities.highest("max_value").first
-		);
+		unit_ability_list resistance_max_value;
+		resistance_max_value.append_if(resistance_abilities, [&](const unit_ability& i) {
+			return (*i.ability_cfg).has_attribute("max_value");
+		});
+		if(!resistance_max_value.empty()){
+			res = 100 - std::min<int>(
+				resist_effect.get_composite_value(),
+				resistance_max_value.highest("max_value").first
+			);
+		} else {
+			res = 100 - resist_effect.get_composite_value();
+		}
 	}
 
 	return res;
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/resistance_value.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/resistance_value.cfg
@@ -1,0 +1,144 @@
+# wmllint: no translatables
+
+#####
+# API(s) being tested: [resistance]abilitie witout max_value specified
+##
+# Actions:
+# Give both Alice and Bob 100% chance to hit.
+# Give Bob and Alice [resistance]add=20 without max_value attribute and give damage=10
+# Move Alice next to Bob, and have Alice attack Bob.
+##
+# Expected end state:
+# Alice and Bob fight with damage 10-(10*20/100)=8.
+#####
+
+{GENERIC_UNIT_TEST "resistance_value_test" (
+    [event]
+        name=start
+        [modify_unit]
+            [filter]
+            [/filter]
+            max_hitpoints=100
+            hitpoints=100
+            attacks_left=1
+        [/modify_unit]
+        [object]
+            silent=yes
+            [effect]
+                apply_to = new_ability
+                [abilities]
+                    [resistance]
+                        id=steadfast1
+                        add=20
+                        affect_self=yes
+                        #max_value=90
+                    [/resistance]
+                [/abilities]
+            [/effect]
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [attacks]
+                        value=1
+                    [/attacks]
+                    [damage]
+                        value=10
+                    [/damage]
+                    [chance_to_hit]
+                        value=100
+                    [/chance_to_hit]
+                [/set_specials]
+            [/effect]
+            [filter]
+                id=bob
+            [/filter]
+        [/object]
+        [object]
+            silent=yes
+            [effect]
+                apply_to = new_ability
+                [abilities]
+                    [resistance]
+                        id=steadfast1
+                        add=20
+                        affect_self=yes
+                        #max_value=90
+                    [/resistance]
+                [/abilities]
+            [/effect]
+            [effect]
+                apply_to=attack
+                [set_specials]
+                    mode=append
+                    [attacks]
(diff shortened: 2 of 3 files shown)


newfrenchy83:
@Pentarctagon could you review this PR please?

newfrenchy83:
@Pentarctagon I only see two possible solutions: either allow the ability to work without the max_value attribute by giving it the get_composite_value if no max_value attribute is present in the list, or make the ability invalid and remove it from the list . What would you prefer?

newfrenchy83:
@Pentarctagon PR reopen, i think have fixed testing

src/units/unit.cpp
```diff
@@ -1798,9 +1798,17 @@ int unit::resistance_against(const std::string& damage_name,bool attacker,const
...
 		unit_abilities::effect resist_effect(resistance_abilities, 100-res);
 
+		bool has_max_value = false;
+		for(auto& i : resistance_abilities) {
+			if((*i.ability_cfg).has_attribute("max_value")){
+				has_max_value = true;
+				break;
+			}
+		}
+		int max_value = has_max_value ? resistance_abilities.highest("max_value").first : 100;
```
ProditorMagnus, line 1808:
```
[resistance]
	id=steadfast1
	add=20
	affect_self=yes
	#max_value=90 - default 100
[/resistance]
[resistance]
	id=steadfast2
	add=20
	affect_self=yes
	max_value=50
[/resistance]
[resistance]
	id=steadfast3
	add=20
	affect_self=yes
	max_value=10
[/resistance]
# expected: steadfast3 limits resistances to 10
```
I think currently it limits to 50 instead.

newfrenchy83: Yes,if max_value  defined then default_value  not used, if value of highest attribute who is used for the choisi beteween composite and max_value

ProditorMagnus: It should use lowest max_value, and highest min_value.

newfrenchy83: what min_value? resistance can't have min_value in diffrence of [illuminates]

ProditorMagnus:
Resistance over 100% is treated same as 100%, while weakness of -200% takes more damage than weakness of -100%. Because of that it would make sense to have default min_value -infinity. -100% fits most units, but not as concept.

newfrenchy83:
I think what i preder https://github.com/wesnoth/wesnoth/pull/8094 concept.

newfrenchy83:
i will remove in_value attribute

newfrenchy83:
the actual implementation select the highest max_value like refrencenot the lowest, i don't thinf what an change like that will be validated.

newfrenchy83:
@Pentarctagon @ProditorMagnus i changed strategy for what max_value(and min_value) will be optionnal now.

ProditorMagnus:
I like that they are optional.

max_value = resistance_abilities.highest("max_value").first;
I think this needs to be
max_value = resistance_abilities.lowest("max_value").first;

Most restrictive value should apply.

newfrenchy83:
> I like that they are optional.
> 
> max_value = resistance_abilities.highest("max_value").first; I think this needs to be max_value = resistance_abilities.lowest("max_value").first;
> 
> Most restrictive value should apply.

It's been years (maybe since the beginning) that the highest value has been chosen, I don't think such a change will be approved

ProditorMagnus (approved):

newfrenchy83:
@Pentarctagon Could you review this PR, please?

Pentarctagon:
@ProditorMagnus is a `min_value` attribute needed for this? We *are* supposed to be in feature freeze after all.

ProditorMagnus:
Not needed.

newfrenchy83:
> @ProditorMagnus is a `min_value` attribute needed for this? We _are_ supposed to be in feature freeze after all.

i removed 2 last commits

ProditorMagnus (changes requested):
This reintroduces the issue of using highest max_value

newfrenchy83:
> This reintroduces the issue of using highest max_value

Corrected

src/units/unit.cpp
```diff
@@ -1786,6 +1786,19 @@ bool unit::resistance_filter_matches(const config& cfg, bool attacker, const std
 	return true;
 }
 
+static std::optional<int> ref_value(unit_ability_list resistance_abilities, const std::string& attribute)
```
ProditorMagnus, line 1789: when it only supports max_value then it may be better to include max in function name

ProditorMagnus (approved):

src/units/unit.cpp
newfrenchy83: @ProditorMagnus except what if a want introduce min_value later in 1.19, i prefer not to have change name anew, i propose to use "interval_value instead"

newfrenchy83: static function removed and code reintroduced directly in function because an attribute only to check

ProditorMagnus: looks good

gfgtdf:
> I like that they are optional.

> max_value = resistance_abilities.highest("max_value").first;
> I think this needs to be
> max_value = resistance_abilities.lowest("max_value").first;

> Most restrictive value should apply.

I disagree here, when i have two abilites a1, a2 where a1 can increase it to for example 50 and a2  can increase it to for example 70 then i certainly wouldn't expect  that a1 which is supposed to increase my resistance to actually lower it.

Pentarctagon:
The `value` attribute currently uses the highest value given (see the "*_cumulative_*" tests over in #8105), so I think the `max_value` attribute should also use the highest value given.

newfrenchy83:
> The `value` attribute currently uses the highest value given (see the "__cumulative__" tests over in #8105), so I think the `max_value` attribute should also use the highest value given.

it is fixed

src/units/unit.cpp
```diff
@@ -1798,10 +1798,22 @@ int unit::resistance_against(const std::string& damage_name,bool attacker,const
...
+			max_value = resistance_max_value.highest("max_value").first;
+		}
+		if(max_value){
+			res = 100 - std::min<int>(
+				resist_effect.get_composite_value(),
+				*max_value
+			);
+		} else {
+			res = 100 - resist_effect.get_composite_value();
+		}
```
Pentarctagon, line 1816: This could be simplified to:
```
if(!resistance_max_value.empty()){
	res = 100 - std::min<int>(
		resist_effect.get_composite_value(),
		resistance_max_value.highest("max_value").first
	);
} else {
	res = 100 - resist_effect.get_composite_value();
}
```
`std::optional<int> max_value;` can then be removed.

newfrenchy83: it's done

ProditorMagnus:
https://wiki.wesnoth.org/index.php?title=AbilitiesWML&type=revision&diff=71916&oldid=71872

PR #8095, opened 2023-12-08
EI S04b: fix scenario id, fix macro in translatable string

### LINKED ISSUE #7994: EI S04b: Scenario ID has en_GB spelling

- **Version:** 1.17.22+dev 1c51bd33885c6870cba276c34a19d14cd8df6eaf

### Description of the bug

EI S04b, "Ill Humours", is spelt in `en_GB`, and so is its scenario ID. As it's a scenario ID, it'll need more testing than a simple typo fix.

Within the file itself, both "humour" and "vapours" are spelt in en_GB.

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

### LINKED ISSUE #8086: EI S04b: puts a macro in a translatable string

- **Version:** 1.17.3

### Description of the bug

Similar to #8084, EI's S04b uses a macro in a translatable string: `"To travel east, collect {COUNT} alswdan herbs and pay Hahid al-Ali $gold_cost gold. <i>(challenging)</i>"`

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

stevecotton:
Two commits, related by the fact that one renames a file that the other one changes. Will split into two PRs if either needs to be significantly changed.

EI S04b: use en_US spelling for the scenario id
---

* S04b start-of-scenario saves since 1.17.21 will be broken
* S03 mid-scenario saves since 1.17.21 will show a warning,
  and players will only be able to go west
* Saves from earlier versions are already broken by the campaign rewrite

Fixes #7994, but does not change the translatable strings which are also mentioned in that issue.

EI S04b: Avoid using a macro in a translatable string
---

The required number is already in a WML variable, herbs_needed. However, the string is used in an `[objective]` tag, which means we're not necessarily in an ActionWML block when it's shown, and thus can't use the Lua API to look up plural strings. Instead, brute force it by making 6 separate strings.

Also, use the mainline `ALTERNATIVE_OBJECTIVE_CAPTION` macro instead of a campaign-specific translatable string.

Fixes #8086, which requires changing a translatable string.


stevecotton:
Hmm - the map-diff tool fails if a map file is renamed, but that shouldn't be a blocker for this.

Dalas121 (approved):

Issue #8065, opened 2023-11-23
(Re)move gzip/bzip2 command line compresion code from wesnoth.cpp 


grz0:
### Describe the desired feature

I was recently analysing Wesnoth initialisation code and stumbled upon command line (de)compression code. IMO, providing a compression tool in the game is out of scope and should be removed or moved to some other file than wesnoth.cpp.

knyghtmare:
I think the decompression is necessary for cache, save files, addons downloaded from the server so asking for it's removal sounds unnecessary.

Pentarctagon:
It's not really hurting anything, but yeah I've also wondered why someone bothered exposing this as command line options for wesnoth.

Pentarctagon:
Remove the gunzip/gzip and bunzip2/bzip2 code from wesnoth.cpp and commandline_options.cpp. Also remove any associated header includes.

Pentarctagon:
If you feel strongly about it then feel free to close the issue. Though I can't imagine anyone would actually need to use this functionality rather than just... using bzip2 or gzip.

soliton-:
While it is very common to have bzip2 or gzip installed on a UNIX-like OS that is not a requirement for running wesnoth and I doubt you can decompress either on windows for example without installing a thirdparty tool.

grz0:
@soliton-  How many players on Windows would come up with an idea to check command line arguments and use Wesnoth binary to decompress files instead of just installing 7-zip or other third-party tools?

soliton-:
No clue.

AnkushSingh0710:
I removed the code which is mentioned above and removed the header which is related to compression so should I go for pull request now

ProditorMagnus:
Is any of data/tools relying on this?

CelticMinstrel:
I did a quick search and didn't find anything in data/tools or utils that relies on this.

AnkushSingh0710:
the issue is solved or not?

Pentarctagon:
I think it depends on Soliton - I consider it to essentially be dead code, but it's definitely not something I feel strongly enough about that I'd want to force the issue.

CelticMinstrel:
Even if this is removed, is it not still possible to effectively decompress a saved game by disabling compression in Wesnoth preferences, opening the game, and resaving it? Although I'm not sure if that works for replays or start-of-scenario saves.

[...]

PR #8098, opened 2023-12-09
removed the compression code

AnkushSingh0710:


Wedge009:
Presumably this is trying to address #8065? I'm not sure the question of whether it's really a good idea to do this was properly addressed.

stevecotton:
I'm not sure if it's necessary either, but if it is done then it should make corresponding changes to:
* src/commandline_options.hpp
* src/tests/test_commandline_options.cpp (which is only built and run if the Boost-based tests are enabled in CMake or SCons)

I'll trigger a workflow run, because I want to find out if the Boost tests detect the removal of the functionality (I think they will).

Pentarctagon:
To me it's basically:
* the code doesn't hurt anything by being there
* the command line options aren't actually used by anyone or anything

So I think it makes sense to remove in the same way that removing dead code makes sense to remove, but I also don't have a strong opinion about it.

CelticMinstrel (changes requested):
If this is to be merged, the things mentioned by @stevecotton need to be addressed. In addition, the options need to be removed from the man page (wesnoth.6 under docs/).

grz0:
Also this line must be changed:
https://github.com/wesnoth/wesnoth/blob/79836c488325c2ed4071a2441fd8736968c1c71e/src/wesnoth.cpp#L983

AnkushSingh0710:
@grz0 i changed now can you please check it once and let me know if its
correct.

AnkushSingh0710:
can you guys please tell me what else i need to change.

Pentarctagon:
Nothing seems to have been pushed? It also does not remove anything from `commandline_options.cpp` as was mentioned in the issue.

AnkushSingh0710:
yaa i didn't pushed yet my bad. do i need to remove both bunzip2 and bzip2
or just b zip2?

Pentarctagon:
Both.

AnkushSingh0710:
Now i removed those codes from commandline_options.cpp also.

Pentarctagon:
Looks like the code no longer builds.

stevecotton:
> can you guys please tell me what else i need to change.

This question reminds me of something that I took a long time to understand about the transition from school to university - these "good first issues" are things where we'd like the work to be done, we're sure that it's an appropriately sized task, but we haven't done it ourselves yet. That is to say, we don't have an already-tested version of the answer to look at; instead, when you're done we'll be reviewing your work and saying "yes, that seems to be the right solution, we'll use it".

For this one, I'd suggest looking at all the places that `commandline_option::bunzip2` is used - does #8065 mean removing every place that that variable can be set to a non-`std::nullopt` value? If it does, the variable and every codepath that's only reached when it's not `std::nullopt` can be removed.

src/commandline_options.cpp
```diff
@@ -64,8 +64,7 @@ bad_commandline_tuple::bad_commandline_tuple(const std::string& str,
 
 
 commandline_options::commandline_options(const std::vector<std::string>& args)
-	: bunzip2()
-	, bzip2()
+	:
```
CelticMinstrel, line 67: This should look like

```
	: campaign()
	, campaign_difficulty()
```

src/wesnoth.cpp
```diff
@@ -980,7 +980,7 @@ int main(int argc, char** argv)
 		// the first = character, or in a subsequent argv entry which we don't
 		// care about -- we just want to see if the switch is there.
 		static const std::set<std::string> terminal_arg_switches = {
-			"--bunzip2", "--bzip2", "-D", "--diff", "--gunzip", "--gzip", "-p", "--preprocess", "-P", "--patch",
+		"-D", "--diff", "-p", "--preprocess", "-P", "--patch",
```
CelticMinstrel, line 983: There should be one more tab on this line.

AnkushSingh0710: i didn't understand.

src/commandline_options.cpp
AnkushSingh0710: okay.

src/wesnoth.cpp
CelticMinstrel: The indentation level is wrong.

AnkushSingh0710: okay i did.

AnkushSingh0710:
i changed both things please check it.

Pentarctagon:
Looks like there's some bits that need to be removed from https://github.com/wesnoth/wesnoth/blob/master/src/tests/test_commandline_options.cpp as well.

AnkushSingh0710:
yaa okay

AnkushSingh0710:
done.

stevecotton:
Please reread my comments, there's at least one more file that needs changes.

src/wesnoth.cpp
```diff
@@ -73,16 +73,13 @@
...
 #if defined(_MSC_VER)
 #pragma warning(push)
 #pragma warning(disable : 4456)
 #pragma warning(disable : 4458)
 #endif
 
-#include <boost/iostreams/filter/gzip.hpp> // for gzip_compressor, etc
-
 #if defined(_MSC_VER)
 #pragma warning(pop)
```
stevecotton, line 84: This one's learnt from experience, but these `#ifdef _MSC_VER #pragma ...` lines are turning two specific warnings off because they're triggered by the gzip.hpp file, and then the `pop` is turning them back on again.

All of these should be removed along with the `#include <.../gzip.hpp>` line.

AnkushSingh0710:
i'm not getting which lines should i remove.

AnkushSingh0710:
Which lines
cam anyone please tell me how to decide that which lines to remove.

stevecotton:
Sorry, but my point earlier about school vs university is that it's your responsibility to work out which lines need to be deleted, or your teachers' responsibility tell you where to learn from, and then your role to learn from those places. Our role should be to review a contribution, but it shouldn't require us to do all the work of writing a model answer in order to tell you exactly which lines to remove.

AnkushSingh0710:
yaa okay got it sorry sorry.

AnkushSingh0710:
Please check once.

Pentarctagon:
Some code that should be deleted is still missing.

CelticMinstrel:
As far as I can tell, the only thing still missing is to delete the declarations of the relevant variables in commandline_options.hpp? Though once you remove those it's possible you'll find one or two other things that were missed.

AnkushSingh0710:
removed the declarations form wesnoth.6

Pentarctagon:
That was not the only thing mentioned.

CelticMinstrel:
I don't understand why you suddenly changed doc/man/wesnothd.6 (note the extra D in there; the change to doc/man/wesnoth.6 is correct).

AnkushSingh0710:
> As far as I can tell, the only thing still missing is to delete the declarations of the relevant variables in commandline_options.hpp? 

Okay thanks

>Though once you remove those it's possible you'll find one or two other things that were missed.

I found gZip/gunZip and bZip/bunZip lines in comandline_option.hpp and i removed those lines. if something else should need to removed please let me know. 


> Oh right, there's also the man-page at doc/man/wesnoth.6 that needs to have the reference to these options removed.

There are some non English documentation also should i change them also.

stevecotton:
Regretfully, I think we should decline this PR, because I feel Ankush hasn't investigated how the code that he's changing interacts with the rest of the system. To work out whether the change in this PR is correct, someone needs to do that investigation, at which point they may as well be the author of a new PR, and take the credit for the PR; this would also make clear that a different person should be the reviewer.

For the change to wesnothd.6, although the `ban_save_file` option does involve a gzipped file, the PR includes no changes to the code that handles that option. That code shouldn't change for this, but finding that inconsistency would be a sign of doing the investigation.

PR #8099, opened 2023-12-09
automatically generate "load_addon_<addon_id>" [resource]

gfgtdf:
It is a common problem that when an addon1 that wants to  make use of another addon2s assets (units etc) it doesn't really have a good way to do so unless addon2 provides a [resource] that addon1 can load via [load_resource]. In particular it seems to be a not so uncommon practice to include other addons data via macro inclusion, however this is a very bad things to do since

* It parses the addons basicially twice (or even more times), resulting in much longer loading times.
* It evades the addon versions checks in mp (when the version of the included addon had changed for one player), leading to possible OOS erros during the game.
* It will can lead to events being exceuted twice when multiple addons include the same code that way.

To make it easier for addon1 to load addon2 without having to ask the author of addon2 to add a [resource] he can load, the game now automatically generates a [resource] for each addon.

### DIFF
--- a/src/game_config_manager.cpp
+++ b/src/game_config_manager.cpp
@@ -562,6 +562,7 @@ void game_config_manager::load_addons_cfg()
 			config umc_cfg;
 			cache_.get_config(main_cfg, umc_cfg, validator.get());
 
+			umc_cfg.add_child("resource", config{"id", "load_addon_" + addon_id});
 			static const std::set<std::string> tags_with_addon_id {
 				"era",
 				"modification",


stevecotton:
There should be some way for addon2 to override the engine-provided function. Otherwise, we'll have traps such as `[load_resource]id=load_addon_Ageless_Era` not enabling the events that `[load_resource]id={AE_REQUIRED_EVENTS_RESOURCE_NAME}` does.

See also (despite the name this is for 1.16 / 1.17) https://github.com/ProditorMagnus/Ageless-for-1-14/blob/master/data/resource.cfg

gfgtdf:
> There should be some way for addon2 to override the engine-provided function

Do you have an idea on how it should look in particular? My first idea was simply not to generate that [resource] when the addon already provides any [resource] but then i thought that a really bad idea because it could happen that the add-on generated a resource depending  of preproceccor defines

stevecotton:
Argh. I believe you, but at the moment my head isn't getting round the complexities of add-on loading mixed with preprocessor defines.

Maybe just have a way for addon2 to force an error if addon1 tries to load the automatically-added resource.

knyghtmare:
This issue also affects the UWO as it registers multiple entries

image

Currently addons loading LotI and all addons associated with "The Dark Master" UMC series are exhibiting such errors

CelticMinstrel:
So the automatically-generated resource doesn't actually contain anything and only serves to mark the add-on as "in use", right?

Logically it seems that if the add-on explicitly defines a resource with the same ID, then it shouldn't generate this resource. Which then makes me wonder, do we really need the `load_addon_` prefix on the ID? Can't it just use the add-on ID directly?

gfgtdf:
> Maybe just have a way for addon2 to force an error if addon1 tries to load the automatically-added resource.

I'm still a bit unsure tbh. in the end this will always be a feature that will work on some addons and not on others, even if addon2s authos don't actively disable it, it is currently still mostly the responsibility of the author of addon1 to check that it works (which usualyl means looking at the code of addon2). 


> So and done  the automatically-generated resource doesn't actually contain anything and only serves to mark the add-on as "in use", right?

yes

> Logically it seems that if the add-on explicitly defines a resource with the same ID, then it shouldn't generate this resource. Which then makes me wonder, do we really need the load_addon_ prefix on the ID? Can't it just use the add-on ID directly?

Hmm, i don't have a that strong opinion on that. i mostly thought having a more unique name like that makes it easier to for example to grep for it, or to change it later via wmlint if we need to.

CelticMinstrel:
Hmm… if the goal is specifically to make it greppable, the current ID is fine, but… it could also be done with a different attribute, like:
```
[load_resource]
  add_on=My_Addon_Id
[/load_resource]
```

soliton-:
With a different attribute it's impossible to get name clashes in existing resource usage so adding this and seeing how well it works should be safely doable.

I agree that there is likely no general solution to make this work properly with any addon but having an auto-generated resource available for the cases where it works without having to coordinate with the other addon's author seems useful.

PR #8100, opened 2023-12-09
add priority= attribute to [lua]  tags

gfgtdf:
When multiple [lua] tags from different [resource]s [era]s etc interact  with each other, the engine currently make no guarantee in which order they are executed, making it hard to reliably use (global) varialbes/functions in one [lua] tag that was defined in other [lua] tags

### DIFF
--- a/src/saved_game.cpp
+++ b/src/saved_game.cpp
@@ -418,6 +418,10 @@ void saved_game::expand_mp_events()
 				load_non_scenario("resource", id, pos);
 			}
 		}
+
+		auto lua_tags =  starting_point_.child_range("lua");
+		std::stable_sort(lua_tags.begin(), lua_tags.end(), [](const config& c1, const config& c2) { return c1["priority"].to_int() > c2["priority"].to_int(); });
+
 		starting_point_["has_mod_events"] = true;
 		starting_point_["loaded_resources"] = utils::join(loaded_resources);
 	}


stevecotton:
Could we instead have a dependency tree, where using `[load_resource]` guarantees that any Lua functions in the loaded resource will be run before those in the requesting resource?

CelticMinstrel:
This seems kind of confusing. Is it only operating on global Lua tags? Or is it operating on Lua tags placed in add-ons (eg the campaign tag)? Or does it combine them all first and then sort them? I'd say it's rather worthless if it's not the last one.

@stevecotton's suggestion seems to make a lot of sense, as long as you can determine the add-on that a global Lua tag belongs to. I'd say global Lua tags in an add-on should be treated "as if" they are a dependency of every resource, modification, etc defined in that add-on. So, the global tags run first, then the ones in the resource requested by your modification or era, then the ones in the modification or era itself. Multiple tags in a given location are simply run in order.

There is one possible case where this might be insufficient, though. Is there any case where a resource would want to run its Lua code _after_ the modification that depends on it? I'm not sure what's a good way to handle that. Priority could do it, but I don't think it would do it well.

stevecotton:
> There is one possible case where this might be insufficient, though. Is there any case where a resource would want to run its Lua code _after_ the modification that depends on it? I'm not sure what's a good way to handle that. Priority could do it, but I don't think it would do it well.

In that case, I think the resource should define a Lua function that the dependent can call when the dependent's Lua is run.

gfgtdf:
> This seems kind of confusing. Is it only operating on global Lua tags? Or is it operating on Lua tags placed in add-ons (eg the campaign tag)? Or does it combine them all first and then sort them? I'd say it's rather worthless if it's not the last one.

it combines ans sorts lua tags in various addon tags (campaign/resource) etc but not toplevel/global lua tags.

> Could we instead have a dependency tree, where using [load_resource] guarantees that any Lua functions in the loaded resource will be run before those in the requesting resource?

I think we could, i'm currently wondering whether there are usecases for wanting a different order though, even between  addons that are not mayndarotiry  dependent on each ohter.

CelticMinstrel:
> I think we could, i'm currently wondering whether there are usecases for wanting a different order though, even between addons that are not mayndarotiry dependent on each ohter.

To me it just seems natural to run code from the dependency before the dependent though? If you have a dependency diamond though, it doesn't matter which order the middle level is executed in. so we'd do those in whatever order they happen to be processed in.

> In that case, I think the resource should define a Lua function that the dependent can call when the dependent's Lua is run.

I don't know if that really satisfies the use-case. It's true that the add-on that depends on it knows that it depends on it and thus could call that function, but… if the depended-on resource won't work without calling that function, it seems a bit annoying for anyone who wants to use that resource…

But… if I think about it, maybe the best way to handle that is actually with a preload event? Or some other kind of generic loading hook that the resource can register with.

gfgtdf:
especially since its usually easier to add than to remove such a feature, i think its probably better for now if we just try to make sure dependencies are executed before dependees.

gfgtdf:
i currently have no intentions on working on this

soliton-:
@gfgtdf any point in keeping the branch?

Issue #2919, opened 2018-04-20
--log-<level>=all cannot be overridden piecewise


irydacea:
While trying to debug an issue with the MP lobby/server, I found that using either of these command lines (where the first should be the correct one, the second was just a futile experiment):

```
./wesnoth --log-info=all --log-error=gui/draw
./wesnoth --log-error=gui/draw --log-info=all
```

... still results in log noise from the `gui/draw` log domain, which is particularly noticeable if you mouse over buttons on the title screen, or (more egregiously) if you enter the MP lobby, where it'll spam messages about formula evaluations every second or so.

The expected behaviour ought to be that using the first command line resulted in all log domains but `gui/draw` being set to the `info` log level, with the latter domain being set to `error` instead.

PR #8102, opened 2023-12-11
Allow fine-tuning log levels for log domains from the command line

rfsc-mori:
Fixes the issue #2919.

Old behavior:
The log severities would always override the previous one, in this order: error < warning < info < debug < none.

Changes:
1) Whenever a log domain is specified more than once, the one with higher logging level will be applied, following this pattern: none < error < warning < info < debug.
2) The log domain severity is applied from the most generic directives to the more specific directives. This means that the "all" directive is applied before a "gui/*" wildcard directive, and then, finally, a "gui/draw" is applied last. By redefining this order, it is now possible to fine-tune log levels from the command line.

Tests added:
1) Adds a test case for the log domain severity override ordering, ensuring the domain and severity values parsed from the command line are applied correctly according to the new rules.

Decision context:
I've decided to add this change to commandline_options.cpp for these reasons:
1) It is in the closest context to command line parsing and is very close to the related function parse_log_domains_.
2) The code is not used anywhere else, so I believe it is fine to fix this issue only for the command line code path.
3) It is easily testable and already contains a test suite.
4) Requires the minimum number of changes. One alternative would be implementing this behavior in the log.cpp lg namespace, then passing the vector from wesnoth.cpp process_command_args to a new function that applies the severity to each log domain correctly. Another alternative is adding extra global state to log.cpp and/or set_log_domain_severity to enhance the handling for the "all" case.

This is still one of my first PRs for open source projects, so I'm not sure if I did everything correctly. If there are any requests, mistakes or anything at all, please let me know. I'll do my best to fix!


rfsc-mori:
It seems I missed the failing tests in the test_full_options test case after this change. I have fixed those now.

Pentarctagon:
New tests need to be added to https://github.com/wesnoth/wesnoth/blob/master/boost_test_schedule, otherwise they aren't run by the CI.

soliton-:
This is a good start but I'd rather that the logging options are applied in the order given on the commandline than sorting them a certain way.

As I understand it the current issue is that as you said they are applied in a fixed order no matter what order they were given which is the issue.

The sorting you implemented is one possibility but IMO the more intuitive behavior is to respect the order the options are given in.

soliton-:
I would also not mind if we allowed giving the log options multiple times and handle `--log-info=domain1 --log-info=domain2` the same way as `--log-info=domain1,domain2`.

It's mildly annoying that most options cannot be given multiple times when they could just add to or override previous ones like in most other programs.

rfsc-mori:
Yes, the issue can be seen here: https://github.com/wesnoth/wesnoth/blob/ff8d4d063f59a0aa54271184ddbc1e4fc608d13b/src/commandline_options.cpp#L388-L397

The command line is parsed for each log severity, and then each parse_log_domains_ splits the specified domains and adds them to the log vector. The log levels are always overridden by the next if block, and this is problematic when `all` or wildcards are processed.

I've changed the implementation and added support for multiple log options for the same severity.

According to the [variables_map documentation](https://www.boost.org/doc/libs/1_83_0/doc/html/boost/program_options/variables_map.html), it is derived from std::map, and this means the keys are stored in ascending order.

To work around this default ordering, I have to handle them somewhat manually:
https://github.com/rfsc-mori/wesnoth/compare/ff8d4d06..f4620161

Is this the right direction? I'm a lot more satisfied with this solution, thank you! It also avoids breaking the older test.

I made a separate branch and diff because I don't know how to proceed right now since the PRs are not squashed.

If my current changes are fine, should I rebase and erase these previous commits? Or simply add my current changes on top of this branch?

soliton-:
Excellent, this looks pretty much as I suspected. Feel free to force push the new changes over what's here right now.

rfsc-mori:
Done! Thank you so much! Again, please let me know if there are any more changes to make, if there's any I'll work on them as soon as my work hours ends today!

soliton-:
One thing I'm wondering is if that composing thing messes with the order again. Does this mean that `wesnoth --log-info=domain1 --log-error=domain1 --log-info=domain2`  would first apply the info=domain1,domain2 part, then error=domain1 and then info=domain1,domain2 again?

If so I think the composing feature does not work so well here unless we could somehow combine all the --log-... options into one. Or actually instead of the join we could pick the first `option.value` then the second etc.

rfsc-mori:
~~I believe that should happen if the vm["log-info"] method is used. I should test this case later today to confirm.~~ Actually, I decided to test right now.

Command line:
`wesnoth --logdomains --log-info=all,gui/draw,unit --log-error=all,gui/draw --log-none=all,gui/draw --log-info=all,gui/draw --log-none=all,gui/draw --log-warning=unit`

This is what I've used:
```c++
boost::algorithm::join(option.value, "|") // Using '|' to prevent ambiguity
```
```
log-info all,gui/draw,unit
log-error all,gui/draw
log-none all,gui/draw
log-info all,gui/draw
log-none all,gui/draw
log-warning unit
```

The vm["log-info"] method:
```c++
vm[option.string_key].as<std::vector<std::string>>()
boost::algorithm::join(vm[option.string_key].as<std::vector<std::string>>(), "|")
```
```
log-info all,gui/draw,unit|all,gui/draw
log-error all,gui/draw
log-none all,gui/draw|all,gui/draw
log-info all,gui/draw,unit|all,gui/draw
log-none all,gui/draw|all,gui/draw
log-warning unit
```

My first choice was to use the parsed options directly to prevent two lookups, since I'm already iterating over the command line. But it seems now that the join was not needed, and I should just pick the first vector value... I tried to double-check the documentation to confirm in what cases the option.value vector will have more than one element, and I cannot confirm without relying on the implementation details.

I want to change this to using the vm["log-info"] method and add a map (or an undordered_map) to track the current index for each log severity. The main reason is to have a more well-defined behavior at a small extra lookup cost.

Also, I made the test case with that in mind: https://github.com/wesnoth/wesnoth/blob/bac47753c2b7f9c73bf3e2f7cf321d42a8e96399/src/tests/test_commandline_options.cpp#L395-L415

The order for the --log-errors should be unchanged. I want to add a few more --log-* parameters as well on the next changes.

Thanks again! May I add these changes later today?

soliton-:
From what I can tell the composing makes it so you now get a list of for example log-info options. so when you iterate in command line order when you process the first log-info option you should only use the first element of that list. When you get to the next log-info option the second element etc. That way you would truly process in command line order and don't need to join anything.

I think that would be the simplest way to go.

soliton-:
Actually maybe I misunderstood your example output. It seems like `option.value` does already just contain the one current option and the join is simply not needed? Testing the current build a bit it does seem to already work as expected anyway.

rfsc-mori:
Yeah, `option.value` contains just one value from the vector during any iteration. I believe I should just remove the join, I misunderstood what it meant at first, and it has no effect.

My main worry is that I can't confirm this behavior from the parsed options' documentation alone, as it is a sort of intermediary class. It seems to be the only way to obtain the order aside from doing it manually or with another dependency, and it seems fine to rely on this.

But I'm not sure if it's correct to rely on the value vector always having one element, although it is very unlikely to change.

The alternative is keeping an iterator or index and fill the log vector from the `variable_map` on each iteration, but I feel I'm over-worrying over this.

Do you think it's best to just remove the join since it's likely to just work, or should I be a bit paranoid and implement it through vm["log-info"], vm["log-error"] etc., and access the first, then second elements and so on?

Also, I'm very sorry for overcomplicating things. I apologize. Thank you so much for your time and feedback!

rfsc-mori:
I went ahead and removed the join call since it has no effect, since it's likely to be a good change.

soliton-:
> Also, I'm very sorry for overcomplicating things. I apologize. Thank you so much for your time and feedback!

No worries, it was rather me making this more complicated by suggesting the addition of allowing to specify the same option multiple times.

Thank you for your contribution!

rfsc-mori:
Thank you so much! I'll search for something else to work on very soon!

PR #8109, opened 2023-12-12
allow units.remove_modifications to remove multiple types

gfgtdf:
previously it needed multiple calls to remove multiple types of modifications (which could be slow since each call involves a full rebuild of the unit).


CelticMinstrel:
Removing multiple types of modifications in one call should look like this:

```lua
u:remove_modifications({'object', 'trait'}, filter)
```

From the code it looks like your implementation doesn't achieve that.

gfgtdf:
the tagname has always been the third parameter of that function, i don't see enough reason to break compatibility by making it the second.

stevecotton:
Please add a unit test

gfgtdf:
push a unit test (and some other unrelated stuff)

CelticMinstrel:
> the tagname has always been the third parameter of that function, i don't see enough reason to break compatibility by making it the second.

That's fine, the key point is that multiple tag names should be specified as a list, not a string containing commas.

PR #8112, opened 2023-12-14
Fix out of sync errors with pick advance mod

gfgtdf:
local lua variables  are not persitent over reloading. moveto events that change the gamestate cannot be undoable unless on_undo code is provided.

### DIFF
--- a/data/modifications/pick_advance/main.lua
+++ b/data/modifications/pick_advance/main.lua
@@ -198,15 +198,14 @@ on_event("start", function()
 end)
 
 -- set "fresh_turn" for the moveto event at the start of each side turn
-local fresh_turn = false
 on_event("turn refresh", function()
-	fresh_turn = true
+	wml.variables.pa_fresh_turn = true
 end)
 
 -- the first time a unit moves at the start of each side's turn, check if there are any new units that need to be forced to make an advancement choice
 on_event("moveto", function()
-	if fresh_turn then
-		fresh_turn = false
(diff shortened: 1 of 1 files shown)


data/modifications/pick_advance/main.lua
```diff
@@ -220,6 +219,7 @@ on_event("moveto", function()
 				initialize_unit(unit)
 			end
 		end
+		wesnoth.allow_undo(false)
```
ProditorMagnus, line 222: > check if there are any new units that need to be forced to make an advancement choice

So if there are no units then no need for allow_undo()?

data/modifications/pick_advance/main.lua
```diff
@@ -198,15 +198,14 @@ on_event("start", function()
 end)
 
 -- set "fresh_turn" for the moveto event at the start of each side turn
-local fresh_turn = false
 on_event("turn refresh", function()
-	fresh_turn = true
+	wml.variables.pa_fresh_turn = true
 end)
 
 -- the first time a unit moves at the start of each side's turn, check if there are any new units that need to be forced to make an advancement choice
```
ProditorMagnus, line 205: why it needs to happen on movement at all?

gfgtdf: I have no idea

data/modifications/pick_advance/main.lua
gfgtdf: it still changes the `pa_fresh_turn` variable,

Issue #8115, opened 2023-12-15
segmentation fault on attack event (berserker)


andrzejskutela:
### Game and System Information

- **Version:** 1.16.10
- **Downloaded from:** github repo
- **Build info:** standard gentoo portage build
- **OS:** gentoo linux 6.4.3

### Description of the bug

```
[multiplayer]
    id="QQ_test"
    name="Crashing Wesnoth test"
    map_data="Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg
Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg
Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg, Gg
[...]

PR #8116, opened 2023-12-15
segmentation fault on attack event (berserker) #8115

gfgtdf:


stevecotton:
There's special handling for attacks that use the in-engine petrify ability on line 1187. Does this mean that that block could be removed, and replaced with a comment about how both the in-engine version and abilities implemented in Lua / WML both use the same code path?

gfgtdf:
I don't think petrifies automaticially sets the attack to nullptr, also then the "petrified" event wouldn't be fired. (if i unerstood your corectly)

stevecotton (approved):
Oh, now I see. So the crash isn't the petrification itself, it's that the WML event in question uses `apply_to=remove_attacks` too.

In that case, this looks good to me.

PR #8117, opened 2023-12-16
Fix ai error

gfgtdf:


knyghtmare:
What AI Error did this exactly fix? I see no linked issue nor do I see any summary of what the fix is

gfgtdf:
?, Its both there in the second commit

knyghtmare:
I meant the PR description where I see "no description provided".

gfgtdf:
Well yes, i assumes that people that are interested would just look in the commit, there werent that many that it really needed a summary.

Pentarctagon:
Is there any way to have it evaluate a formula instead of treating it like it doesn't exist?

CelticMinstrel:
Looks like this is the cause of #8297.

gfgtdf:
It's not the cause, the error was there before, it's just that the fix didn't work correctly.

PR #8118, opened 2023-12-16
1.18 SoF S1 - Victory Conditions

### LINKED ISSUE #8083: Unclear objectives in the first scenario of SoF

BfW 1.16

Check [forums thread](https://forums.wesnoth.org/viewtopic.php?t=57387) for more information

### Description of the bug

The objectives say you have to move Alanin and the caravans to the dwarvish citatel, like it was in older versions, but now you have to move them to the go-here-icon at the top of the map.

_No response_

### Expected behavior

_No response_

### Additional context

Proposal to fix the issue:
Change the objectives, or label the go-here-icon as dwarvish citatel

doofus-01:
This is to fix #8083 
1. Clarify the objectives of where the caravans and rider are supposed to go
2. make the target area a little bigger (could have also made it `enter_hex` instead of `moveto`, but I think this should do - it wasn't the main complaint anyway)

EDIT:  I'm really not a fan of this copyrights.csv CI, it gives me grief even when I don't touch images.
EDIT2: Never mind, I see it has nothing to do with this, it was failing after the recent translations update

### DIFF
--- a/data/campaigns/Sceptre_of_Fire/scenarios/1_A_Bargain_is_Struck.cfg
+++ b/data/campaigns/Sceptre_of_Fire/scenarios/1_A_Bargain_is_Struck.cfg
@@ -218,9 +218,9 @@
             side=1
             [objective]
 #ifdef HARD
-                description= _ "Transport 5 caravans full of silver back to the dwarvish castle"
+                description= _ "Transport 5 caravans full of silver to the dwarvish stronghold in the northeast corner"
 #else
-                description= _ "Transport 4 caravans full of silver back to the dwarvish castle"
+                description= _ "Transport 4 caravans full of silver to the dwarvish stronghold in the northeast corner"
 #endif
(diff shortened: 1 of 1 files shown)


stevecotton:
On the map itself it looks more like a mountain pass, although in the situation it would make far more sense if it was a stronghold or just doors into a mountain.

doofus-01:
Thanks.  Does anyone else need to approve this?

> On the map itself it looks more like a mountain pass, although in the situation it would make far more sense if it was a stronghold or just doors into a mountain.

Given the next scenario, that was supposed to be the very edge of the cave doors area.  Short of mushing the two scenarios together into a long UtBS-style scenario, I don't think we want to depict the doors here.  Would a couple more castle hexes help?  (Not that I plan to do that here, but it could be in a different PR.)

nemaara (approved):

PR #8119, opened 2023-12-16
Update the copyright info for the Finnish help images

stevecotton:
This assumes Styxnix is the right attribution, currently waiting for confirmation in IRC (in the Finnish language channel). I'm going AFK for the afternoon, so anyone please merge once it's confirmed.

Move the previous addition of the UtBS blowgun into sort-by-filename order.

Correct copyright of orb-two-color.png, as it was based on orb.png.


stevecotton:
Added the real name too.

PR #8122, opened 2023-12-16
[wip] add [status_type] toplevel tag

gfgtdf:
[status_type] describes a status (poisoned/unhealable etc), this intends to fix multiple problems:
1) States not being reset (see for example https://github.com/wesnoth/wesnoth/issues/5190). This has caused multiple
issues for example undead units not losing their undrainable etc statuses
when their type (and traits) is changed.
2) The previous way to show custom states (overwriting wesnoth.theme_items.
unit_status() is not very reusable since it is tied to theme items, in
particular it offers no way to show custom statuses in other places (like
unitlist, unit_preview_pane etc). With this it should be easier to implement
custom statuses also at those places.


Pentarctagon:
Adding the Postponed label since we're in feature freeze for 1.18.

gfgtdf:
> Adding the Postponed label since we're in feature freeze for 1.18.

we could restrict reading this tag to the mainline config /block addons from adding it. If this is a good way to fix the first issue them imo is a good thing to have it in 1.18.

CelticMinstrel:
I approve of this for 1.20. My only question is… what textdomain were those strings in? I can't find it from the C++ file that changed, but they are now in wesnoth-help. Is that a change or is it the same>

gfgtdf:
> I approve of this for 1.20. My only question is… what textdomain were those strings in? I can't find it from the C++ file that changed, but they are now in wesnoth-help. Is that a change or is it the same>

It's probably a change but thats unintentional. This pr is not finished yet (also it doesnt rally work yet becasue of the known_boolean_states_ thing)

soliton-:
Since this is specifically for units it'd make more sense to put it under `[units]` just like `[movetype]` for example.

CelticMinstrel:
Some thoughts on where this could go…

There are some obviously-related pairs here - poison / unpoisonable, slow / unslowable, petrify / unpetrifiable. It would be nice if we could encode that relationship into the tag. It could go in either the X or unXable status, but I think it's a bit better in the X status as that allows easily making a custom poison be blocked by unpoisonable, for example.

Some of these status types have associated effects – poison drains 8 HP per turn (unless you're affected by a "poison slow" ability), and slow halves damage. Is there any good way to encode these effects into the tag as well? Maybe something like supporting `[effect]` subtags?

gfgtdf:
> There are some obviously-related pairs here - poison / unpoisonable, slow / unslowable, petrify / unpetrifiable. It would be nice if we could encode that relationship into the tag

I thought so too at first, but then i thought its would probably only cover a small fraction of "un .. able" statuses, (not unhealable, unplaguable etc) so it'd probably rather make the code harder than easier.

> Some of these status types have associated effects – poison drains 8 HP per turn (unless you're affected by a "poison slow" ability), and slow halves damage. Is there any good way to encode these effects into the tag as well? Maybe something like supporting [effect] subtags?

I don't really see the point tbh, most of those things can be implemented with events. And such a feature will likely either be too so simple that it just covers the mainline abilities, or get really complicated so at least for me it'd probably not be worth the time implementing it.

CelticMinstrel:
> I thought so too at first, but then i thought its would probably only cover a small fraction of "un .. able" statuses, (not unhealable, unplaguable etc) so it'd probably rather make the code harder than easier.

True, it doesn't cover all of them, but it covers some of them, and it opens the door for new similar pairs.

> I don't really see the point tbh, most of those things can be implemented with events. And such a feature will likely either be too so simple that it just covers the mainline abilities, or get really complicated so at least for me it'd probably not be worth the time implementing it.

Can they really all be implemented with events? There's already an open issue about customizing poison, so even if it's possible there's clearly some hole in the system…

PR #8123, opened 2023-12-17
1.18 - bridge-end layering relative to castle walls

doofus-01:
This adjusts layering of bridges vs castles to put castle walls above.  The image below is a hybrid: 
1. NW-SE (new)
2. NE-SW (current)
Screenshot_20231216_bridges
My comment in #8121 on this issue may be wrong, I don't think new images are needed after all if "castles on top" are the desired outcome.  If "bridges on top" are desired, then lots of new images likely are needed to avoid weirdness like this:
Screenshot_20231216_bridge-glitches
(the jaggy shadow on the dwarf castle NW and the strange cut-off on the elf castle NW)

### DIFF
--- a/data/core/terrain-graphics/enduring-bridge.cfg
+++ b/data/core/terrain-graphics/enduring-bridge.cfg
@@ -79,7 +79,7 @@
         [/tile]
         [image]
             center=63,126
-            base=108,118
+            base=108,108
             layer={LAYER}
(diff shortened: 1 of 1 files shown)


doofus-01:
Merging now, if "castles-on-top" breaks something, better to figure that out sooner than later.  Status quo isn't great.

PR #8130, opened 2023-12-17
1.18 various ant image updates

doofus-01:
This is a minor update to ants:
1. Make some of the portraits a little more distinct
2. fix the giant ant attack/defend frames so they don't jump around


doofus-01:
The macos stuff has nothing to do with this, so I'll go ahead and merge before the copyrights.csv needs to be touched again.

PR #8131, opened 2023-12-18
1.18 - SoF Wyrms graphics

doofus-01:
Purpose of this PR:

- [x] different portrait for wyrm vs wyrmlet
- [ ] make wyrmlet sprite a bit smaller - probably wings unchanged, but body must shrink


doofus-01:
copyrights.csv strikes again.  I'll fix this conflict manually and open a new PR.

Pentarctagon:
Option (b) I mentioned in #7903 still stands btw.

doofus-01:
I'd forgotten about that.  I guess it's up to you, which is worse: my grousing or having to clean up after me?

There's probably some value in forcing me to think about copyrights and authorship of each image as it comes up.  If we get more consistent contributors, maybe they will have opinions.

Pentarctagon:
I'm fine either way. Mainly it's just that it can be hard to tell over the internet whether something is just a mild annoyance that isn't actually bothering the person vs something that they're actually getting more frustrated with, so I figured I'd bring option (b) up again in case it was the latter.

doofus-01:
Thanks for trying to accommodate and sorry for the confusion.  I'll learn to work with it, just not quite there yet.

PR #8133, opened 2023-12-18
Generalize the use of max_value and add the min_value attribute

newfrenchy83:
In the case of [leadership], the fact that the values of each ability can be added when cumulative=yes can justify the addition of limit values not to be exceeded, but this can also apply to heals and [regenerates] with the use of 'add' or 'multiply'.

As for [resistance], it already uses max_value, but if cumulative=yes, then all max_values can add up which could be problematic.

### DIFF
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/attacks/attacks_max_value.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/attacks/attacks_max_value.cfg
@@ -8,10 +8,9 @@
 # Have the side 1 leader attack the side 2 leader with both its weapons
 ##
 # Expected end state:
-# max_value does nothing.
-# The side 1 leader has 2 weapons both of which have 3 strikes
+# max_value works.
+# The side 1 leader has 2 weapons both of which have 1 strikes
 #####
-#ifndef SCHEMA_VALIDATION
 {COMMON_KEEP_A_B_UNIT_TEST "attacks_max_value" (
     [event]
         name = start
@@ -28,6 +27,5 @@
         [/modify_unit]
     [/event]
 
-    {CHECK_STRIKES "3,3" ({SUCCEED})}
+    {CHECK_STRIKES "1,1" ({SUCCEED})}
 )}
-#endif
--- a/src/units/abilities.cpp
+++ b/src/units/abilities.cpp
@@ -2031,7 +2031,7 @@ effect::effect(const unit_ability_list& list, int def, const_attack_ptr att, EFF
 			}
 		}
 
-		if(wham == EFFECT_CLAMP_MIN_MAX){
+		if(wham == EFFECT_DEFAULT || wham == EFFECT_CUMULABLE){
 			if(cfg.has_attribute("max_value")){
 				max_value = max_value ? std::min(*max_value, cfg["max_value"].to_int()) : cfg["max_value"].to_int();
 			}
--- a/data/schema/units/abilities.cfg
+++ b/data/schema/units/abilities.cfg
@@ -61,6 +61,8 @@
 	max=0
 	super="units/unit_type/abilities/~generic~"
 	{SIMPLE_KEY value f_int}
+	{SIMPLE_KEY max_value f_int}
+	{SIMPLE_KEY min_value f_int}
 	{SIMPLE_KEY add f_int}
 	{SIMPLE_KEY sub f_int}
 	{SIMPLE_KEY multiply f_int}
@@ -79,8 +81,6 @@
 	super="units/unit_type/abilities/~value~"
 	{SIMPLE_KEY apply_to string}
 	{SIMPLE_KEY active_on ability_context}
-	{SIMPLE_KEY max_value f_int}
-	{SIMPLE_KEY min_value f_int}
 	{FILTER_TAG "filter_weapon" weapon ()}
 	{FILTER_TAG "filter_second_weapon" weapon ()}
 [/tag]
@@ -95,12 +95,6 @@
 	name="illuminates"
 	max=infinite
 	super="units/unit_type/abilities/~value~"
-	{SIMPLE_KEY max_value f_int}
-	{SIMPLE_KEY min_value f_int}
-	{SIMPLE_KEY add f_int}
-	{SIMPLE_KEY sub f_int}
-	{SIMPLE_KEY multiply f_int}
-	{SIMPLE_KEY divide f_int}
 [/tag]
 [tag]
 	name="hides"
--- a/data/schema/units/specials.cfg
+++ b/data/schema/units/specials.cfg
@@ -51,6 +51,8 @@
 	{SIMPLE_KEY sub f_int}
 	{SIMPLE_KEY multiply f_real}
 	{SIMPLE_KEY divide f_real}
+	{SIMPLE_KEY max_value s_f_int}
+	{SIMPLE_KEY min_value s_f_int}
 	{SIMPLE_KEY cumulative bool}
 	{DEPRECATED_KEY backstab bool}
 	{FILTER_TAG "filter_base_value" base_value ()}
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/heals/heals_max_value.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/heals/heals_max_value.cfg
@@ -10,10 +10,9 @@
 # Wait a turn for healing to take place.
 ##
 # Expected end state:
-# max_value does nothing.
-# The Mage has 6 hp: 1 + 2 (rest healing) + 3 ([heals])
+# max_value works.
+# The Mage has 4 hp: 1 + 2 (rest healing) + 1 ([heals])
 #####
-#ifndef SCHEMA_VALIDATION
 {GENERIC_UNIT_TEST "heal_max_value" (
     [event]
         name = start
@@ -55,8 +54,7 @@
             [/filter]
             variable = temp
         [/store_unit]
-        {ASSERT {VARIABLE_CONDITIONAL temp.hitpoints numerical_equals 6}}
+        {ASSERT {VARIABLE_CONDITIONAL temp.hitpoints numerical_equals 4}}
         {SUCCEED}
     [/event]
 )}
-#endif
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/leadership/leadership_max_value.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/leadership/leadership_max_value.cfg
@@ -8,10 +8,9 @@
 # Attack each other
 ##
 # Expected end state:
-# max_value does nothing
-# The damage from the attack is increased by 50%
+# max_value works
+# The damage from the attack is increased by 10%
 #####
-#ifndef SCHEMA_VALIDATION
 {COMMON_KEEP_A_B_UNIT_TEST "leadership_max_value" (
     [event]
         name = start
@@ -27,8 +26,7 @@
             [/effect]
         [/modify_unit]
 
-        {ATTACK_AND_VALIDATE 150}
+        {ATTACK_AND_VALIDATE 110}
         {SUCCEED}
     [/event]
 )}
-#endif
--- a/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/regenerate/regenerate_max_value.cfg
+++ b/data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/regenerate/regenerate_max_value.cfg
@@ -9,10 +9,9 @@
 # Wait a turn for regenerating to take place.
 ##
 # Expected end state:
-# max_value does nothing.
-# The Mage has 6 hp: 1 + 2 (rest healing) + 3 ([regenerate])
+# max_value works.
+# The Mage has 4 hp: 1 + 2 (rest healing) + 1 ([regenerate])
 #####
-#ifndef SCHEMA_VALIDATION
 {GENERIC_UNIT_TEST "regenerate_max_value" (
     [event]
         name = start
@@ -52,8 +51,7 @@
             [/filter]
             variable = temp
         [/store_unit]
-        {ASSERT {VARIABLE_CONDITIONAL temp.hitpoints numerical_equals 6}}
+        {ASSERT {VARIABLE_CONDITIONAL temp.hitpoints numerical_equals 4}}
         {SUCCEED}
     [/event]
 )}
-#endif
--- a/src/tod_manager.cpp
+++ b/src/tod_manager.cpp
@@ -241,7 +241,7 @@ const time_of_day tod_manager::get_illuminated_time_of_day(
 			if(itor != units.end() && !itor->incapacitated()) {
 				unit_ability_list illum = itor->get_abilities("illuminates");
 				if(!illum.empty()) {
-					unit_abilities::effect illum_effect(illum, terrain_light);
+					unit_abilities::effect illum_effect(illum, terrain_light, nullptr, unit_abilities::EFFECT_WITHOUT_CLAMP_MIN_MAX);
 					const int unit_mod = illum_effect.get_composite_value();
 
 					// Record this value.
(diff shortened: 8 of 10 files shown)


Pentarctagon:
More generally, I would think that `max_value` and `min_value` should work the same for all abilities.

newfrenchy83:
> Why is there special handling for leadership and illuminates?

Illuminates has already own code for max_value and min_value and i fear to modify that, and for leadership i replace check bool by another with string comparaison

newfrenchy83:
> More generally, I would think that `max_value` and `min_value` should work the same for all abilities.

the illuminated abilities affect both the beneficiary unit as well as the units adjacent to it, the value depending on the nature of the terrain and the time of day which must therefore be calculated hexagon by hexagon, it is therefore impossible to make the comparison between value and max_value cat the luminosity of the terrain must be integrated into the calculation which is done after effect::effect is called hence the need to prevent the use of min_value and max_value in effect:::effect for [illuminates ]

Pentarctagon:
What I mean is that I'd expect they functionally work the same so that for any ability with a value attribute, the final value is restricted to the range specified by `min_value` and `max_value` (if those attributes are set).

newfrenchy83:
> What I mean is that I'd expect they functionally work the same so that for any ability with a value attribute, the final value is restricted to the range specified by `min_value` and `max_value` (if those attributes are set).

No, it was not the case, and that precicely the motice to this PR

Pentarctagon:
Alright. Can you rebase this against the current master branch?

newfrenchy83:
> Alright. Can you rebase this against the current master branch?

it's done, cheking begin.

newfrenchy83:
@Pentarctagon check work perfectly except Windows

Pentarctagon:
You can ignore the Windows failures. It's probably somehow related to the whole xz compromise:
`error: https://github.com/tukaani-project/xz/archive/v5.4.4.tar.gz: failed: status code 404`

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/attacks/attacks_max_value.cfg
```diff
@@ -8,8 +8,8 @@
 # Have the side 1 leader attack the side 2 leader with both its weapons
 ##
 # Expected end state:
-# max_value does nothing.
-# The side 1 leader has 2 weapons both of which have 3 strikes
+# max_value work.
```
Pentarctagon, line 11: **works**, plural

data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/attacks/attacks_max_value.cfg
```diff
@@ -8,8 +8,8 @@
...
+# The side 1 leader has 2 weapons both of which have 1 strikes
 #####
 #ifndef SCHEMA_VALIDATION
```
Pentarctagon, line 14: Is this ifndef still needed?

newfrenchy83: I don't  know

newfrenchy83: ifndef removed

src/units/abilities.cpp
```diff
@@ -1963,6 +1965,13 @@ effect::effect(const unit_ability_list& list, int def, const_attack_ptr att, boo
 			}
 		}
 
+		if((type != "illuminates") && cfg.has_attribute("max_value")){
+			max_value = max_value ? std::max(*max_value, cfg["max_value"].to_int()) : cfg["max_value"].to_int();
+		}
+		if((type != "illuminates") && cfg.has_attribute("min_value")){
+			min_value = min_value ? std::min(*min_value, cfg["min_value"].to_int()) : cfg["min_value"].to_int();
+		}
```
Pentarctagon, line 1973: This should be one outer if-statement for `type != "illuminates"`, and then checking for `max_value` and `min_value` done separately inside it.

src/units/abilities.cpp
```diff
@@ -2048,7 +2057,15 @@ effect::effect(const unit_ability_list& list, int def, const_attack_ptr att, boo
...
-	composite_value_ = static_cast<int>((value_set + addition) * multiplier / divisor);
+	if(max_value && min_value) {
+		composite_value_ = std::clamp(*min_value, *max_value, static_cast<int>((value_set + addition) * multiplier / divisor));
+	} else if(max_value) {
+		composite_value_ = std::min(*max_value, static_cast<int>((value_set + addition) * multiplier / divisor));
+	} else if(min_value) {
+		composite_value_ = std::max(*min_value, static_cast<int>((value_set + addition) * multiplier / divisor));
+	} else {
+		composite_value_ = static_cast<int>((value_set + addition) * multiplier / divisor);
+	}
```
Pentarctagon, line 2068: Can divisor ever be 0 here?

src/units/abilities.cpp
Pentarctagon: Also, what scenario would this be needed for? There's already a min/max check for illuminates in `tod_manager.cpp` isn't there?

newfrenchy83: I thought about allowing value limitations in abilities like leadership or damage but I realized that the simultaneous use of 2 chance_to_hit or 2 leadership, one with max_value and the other with cumulative=yes risks interfere with the operation of the latter; so I plan to close this PR and add min_value only to the resistance abilities

src/units/abilities.cpp
newfrenchy83: fix divisor for what if equals 0 because attribute then divisor value convert to 0.001 for don't cause crash in calculation

src/units/abilities.cpp
Pentarctagon: I'm not sure how that relates to my above question, but I have always wondered why only leadership is apparently affected by cumulative.

src/units/abilities.cpp
Pentarctagon: If it is possible to have a divisor value of 0 here, then that should be handled above when iterating over `values_div`.

src/units/abilities.cpp
newfrenchy83: Cumulative does not work the same way for leadership as it does for other abilities:

in leadership when cumulative=yes then each time and that several examples of the ability influence the same unit, the value will be the sum of each value=.

In special weapons like chance_to_hit, if cumulative=yes then the value used will be the highest between the base value and the value= of the special

src/units/abilities.cpp
newfrenchy83: error message exist cfg."divide" iteration

newfrenchy83: Dividing by 0.001 removed

newfrenchy83:
I will reformulate this PR to limit min and max_value to the resistance ability to reduce checking (for max_value you have to check the resistance list a second time to verify that one or more of the abilities use the attribute), for special weapons (with the abilities used as leadership) these attributes risk causing problems and I refuse to generalize their use

newfrenchy83:
close this PR, review https://github.com/wesnoth/wesnoth/pull/8722, please.

src/units/abilities.cpp
newfrenchy83: reopen this PR because someone add suggest in forum the idea of 2. Inacurrate
This would be an anti marksman, which allows for no more than 20% chance to hit in https://forums.wesnoth.org/viewtopic.php?p=689486#p689486 and i can be used with max_value and value=0 cumulative=yes

and another example would be a semi_magic weapon with a minimum like marksman (value=50) but also a maximum (70), so instead of an imposed fixed value, we would have a fairly easy to define value range, the reverse of the medal and that if a weapon adds a 10% chance of hitting to the whole, the maximum limit of 70% could not be exceeded in this case, but this kind of limitation could be precisely easier to encode than using [ filter_vase_value] and determining the base value as was the case in the Utbs 'training' ability before the quenoth lineage rework.

src/units/abilities.cpp
```diff
@@ -2028,7 +2028,7 @@ effect::effect(const unit_ability_list& list, int def, const_attack_ptr att, EFF
 			}
 		}
 
-		if(wham == EFFECT_CLAMP_MIN_MAX){
+		if(wham == EFFECT_CLAMP_MIN_MAX || wham == EFFECT_CUMULABLE){
```
Pentarctagon, line 2031: Why is this adding `EFFECT_CUMULABLE` when the three other C++ code changes are passing in `EFFECT_CLAMP_MIN_MAX`?

newfrenchy83: For what min_value and max_value  usable in leadership where cumulable is used

Pentarctagon:
Seems fine to me. @stevecotton any objection to merging?

stevecotton:
I've assumed that this is lower priority than the infinite recursion fixes.

My main question is, why are there still some abilities that aren't using this code to clamp the value? For example, illumination handles min_value and max_value, but doesn't use the codepath that's worked on here. `git grep` with this PR included:
```
src/actions/attack.cpp: unit_abilities::effect leader_effect(abil, 0, nullptr, unit_abilities::EFFECT_CUMULABLE);
src/actions/heal.cpp:                   unit_abilities::effect regen_effect(regen_list, 0, nullptr, unit_abilities::EFFECT_CLAMP_MIN_MAX);
src/actions/heal.cpp:           unit_abilities::effect heal_effect(heal_list, 0, nullptr, unit_abilities::EFFECT_CLAMP_MIN_MAX);
src/gui/dialogs/attack_predictions.cpp: unit_abilities::effect dmg_effect(dmg_specials, weapon->damage());
src/tod_manager.cpp:                                    unit_abilities::effect illum_effect(illum, terrain_light);
src/units/abilities.cpp:        return unit_abilities::effect(abil_list, base_value, shared_from_this(), unit_abilities::EFFECT_CLAMP_MIN_MAX).get_composite_value();
src/units/types.cpp:            unit_abilities::effect resist_effect(resistance_abilities, 100 - resistance);
src/units/unit.cpp:             unit_abilities::effect defense_effect(defense_abilities, def);
src/units/unit.cpp:             unit_abilities::effect resist_effect(resistance_list, 100-res, nullptr, unit_abilities::EFFECT_CLAMP_MIN_MAX);
```

newfrenchy83:
> I've assumed that this is lower priority than the infinite recursion fixes.
> 
> My main question is, why are there still some abilities that aren't using this code to clamp the value? For example, illumination handles min_value and max_value, but doesn't use the codepath that's worked on here. `git grep` with this PR included:
> 
> ```
> src/actions/attack.cpp: unit_abilities::effect leader_effect(abil, 0, nullptr, unit_abilities::EFFECT_CUMULABLE);
> src/actions/heal.cpp:                   unit_abilities::effect regen_effect(regen_list, 0, nullptr, unit_abilities::EFFECT_CLAMP_MIN_MAX);
> src/actions/heal.cpp:           unit_abilities::effect heal_effect(heal_list, 0, nullptr, unit_abilities::EFFECT_CLAMP_MIN_MAX);
> src/gui/dialogs/attack_predictions.cpp: unit_abilities::effect dmg_effect(dmg_specials, weapon->damage());
> src/tod_manager.cpp:                                    unit_abilities::effect illum_effect(illum, terrain_light);
> src/units/abilities.cpp:        return unit_abilities::effect(abil_list, base_value, shared_from_this(), unit_abilities::EFFECT_CLAMP_MIN_MAX).get_composite_value();
> src/units/types.cpp:            unit_abilities::effect resist_effect(resistance_abilities, 100 - resistance);
> src/units/unit.cpp:             unit_abilities::effect defense_effect(defense_abilities, def);
> src/units/unit.cpp:             unit_abilities::effect resist_effect(resistance_list, 100-res, nullptr, unit_abilities::EFFECT_CLAMP_MIN_MAX);
> ```

Like illumination choose value in function of TOD, i prefer what it's own code remain for don't have bug later.

newfrenchy83:
> I've assumed that this is lower priority than the infinite recursion fixes.
> 
> My main question is, why are there still some abilities that aren't using this code to clamp the value? For example, illumination handles min_value and max_value, but doesn't use the codepath that's worked on here. `git grep` with this PR included:
> 
> ```
> src/actions/attack.cpp: unit_abilities::effect leader_effect(abil, 0, nullptr, unit_abilities::EFFECT_CUMULABLE);
> src/actions/heal.cpp:                   unit_abilities::effect regen_effect(regen_list, 0, nullptr, unit_abilities::EFFECT_CLAMP_MIN_MAX);
> src/actions/heal.cpp:           unit_abilities::effect heal_effect(heal_list, 0, nullptr, unit_abilities::EFFECT_CLAMP_MIN_MAX);
> src/gui/dialogs/attack_predictions.cpp: unit_abilities::effect dmg_effect(dmg_specials, weapon->damage());
> src/tod_manager.cpp:                                    unit_abilities::effect illum_effect(illum, terrain_light);
> src/units/abilities.cpp:        return unit_abilities::effect(abil_list, base_value, shared_from_this(), unit_abilities::EFFECT_CLAMP_MIN_MAX).get_composite_value();
> src/units/types.cpp:            unit_abilities::effect resist_effect(resistance_abilities, 100 - resistance);
> src/units/unit.cpp:             unit_abilities::effect defense_effect(defense_abilities, def);
> src/units/unit.cpp:             unit_abilities::effect resist_effect(resistance_list, 100-res, nullptr, unit_abilities::EFFECT_CLAMP_MIN_MAX);
> ```

fixed except for illumination who shouldn't change of code.

stevecotton:
If almost all callers pass EFFECT_CLAMP_MIN_MAX, then it might as well be the default code path. Instead of changing the callers, I'd like that to be the default for this function.

In a very quick test, illumination seems to work OK with EFFECT_CLAMP_MIN_MAX added to the call too. My preference would be to remove EFFECT_CLAMP_MIN_MAX from the EFFECT enum, and have EFFECT_DEFAULT clamp to min and max. Given that that's feature creep on the PR, I'd accept if the PR was force-pushed back to 4c32a3b, merged, and then I'd do the next bit; having 4c32a3b in the history would also document the change to leadership.

newfrenchy83:
> If almost all callers pass EFFECT_CLAMP_MIN_MAX, then it might as well be the default code path. Instead of changing the callers, I'd like that to be the default for this function.
> 
> In a very quick test, illumination seems to work OK with EFFECT_CLAMP_MIN_MAX added to the call too. My preference would be to remove EFFECT_CLAMP_MIN_MAX from the EFFECT enum, and have EFFECT_DEFAULT clamp to min and max. Given that that's feature creep on the PR, I'd accept if the PR was force-pushed back to [4c32a3b](https://github.com/wesnoth/wesnoth/commit/4c32a3b03ff1bf77b983266bb9a27e5a22daf8a2), merged, and then I'd do the next bit; having [4c32a3b](https://github.com/wesnoth/wesnoth/commit/4c32a3b03ff1bf77b983266bb9a27e5a22daf8a2) in the history would also document the change to leadership.

in illuminates, this is highest "max_value" and lowest "min_value" who are used then what here it is lowes"max_value" and highest "min_value",same if a change encoding i must maiain exception for "illuminates" abilites

stevecotton:
> in illuminates, this is highest "max_value" and lowest "min_value" who are used then what here it is lowes"max_value" and highest "min_value"

Is there a reason for illuminates to be different like that? It would seem to prevent an ability such as "sprays fog into the air, limiting the maximum light to 10% lawful bonus" to override another illuminates ability with a 20% bonus.

There's a separate rule for terrains that illuminate, but that doesn't stack with abilities even in 1.18. Lava has `light,max_light=25,35`, but a drake hovering over a lava tile at dawn with a Mage of Light adjacent only gets a total of 25, not 35.

newfrenchy83:
> > in illuminates, this is highest "max_value" and lowest "min_value" who are used then what here it is lowes"max_value" and highest "min_value"
> 
> Is there a reason for illuminates to be different like that? It would seem to prevent an ability such as "sprays fog into the air, limiting the maximum light to 10% lawful bonus" to override another illuminates ability with a 20% bonus.
> 
> There's a separate rule for terrains that illuminate, but that doesn't stack with abilities even in 1.18. Lava has `light,max_light=25,35`, but a drake hovering over a lava tile at dawn with a Mage of Light adjacent only gets a total of 25, not 35.

i don't know reason but i don't want change existant behavior and remain careful.

stevecotton (approved):
Assuming the tests pass, I approve this; I then intend to do a follow-up PR changing the behavior of illuminates. @Pentarctagon, over to you

PR #8134, opened 2023-12-18
fix heals and regenrate test validation

newfrenchy83:
shema validation don't disabled for these test


data/test/scenarios/wml_tests/UnitsWML/AbilitiesWML/regenerate/regenerate_max_value.cfg
```diff
@@ -11,6 +11,7 @@
 # Expected end state:
 # The Mage has 6 hp: 1 + 2 (rest healing) + 3 ([regenerate])
```
stevecotton, line 12: Please add the same "max_value does nothing" comment that the other test has.

stevecotton:
Sorry for not merging - Pentarctagon has just pushed the same fix, also without an extra comment. :(

Pentarctagon:
Oh, I saw the comment on my commit but not this PR.

PR #8135, opened 2023-12-19
require_modification/campaign now defaults to yes

gfgtdf:
It's better to be defensive with those default values, especially now that the mp lobby code can automatically download addons, having require_modification set to true isn't inconvenient for the users either.

(The issue that the added comment is abotu isn't something that was added in this commit)

### DIFF
--- a/src/saved_game.cpp
+++ b/src/saved_game.cpp
@@ -343,7 +343,7 @@ void saved_game::load_non_scenario(const std::string& type, const std::string& i
 		non_scenario["name"] = cfg["addon_title"].str("mainline");
 		non_scenario["version"] = cfg["addon_version"].str(version_default);
 		non_scenario["min_version"] = cfg["addon_min_version"];
-		non_scenario["required"] = cfg[require_attr].to_bool(true);
+		non_scenario["required"] = cfg[require_attr].to_bool(!cfg["addon_id"].empty());
 		config& content = non_scenario.add_child("content");
 		content["id"] = id;
 		content["name"] = cfg["addon_title"].str(cfg["name"].str(""));
--- a/src/saved_game.cpp
+++ b/src/saved_game.cpp
@@ -322,6 +322,7 @@ void saved_game::check_require_scenario()
 	config& content = scenario.add_child("content");
 	content["id"] = starting_point_["id"];
 	content["name"] = starting_point_["name"];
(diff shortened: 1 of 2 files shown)


src/saved_game.cpp
```diff
@@ -333,15 +336,14 @@ void saved_game::load_non_scenario(const std::string& type, const std::string& i
 
 		// By default, eras have "require_era = true", and mods have "require_modification = false".
```
ProditorMagnus, line 337: review comments

src/saved_game.cpp
```diff
@@ -231,6 +231,9 @@ void saved_game::set_defaults()
 	};
 
 	if(auto campaign = game_config.find_child("campaign", "id", classification_.campaign)) {
+		// FIXME: The mp code could use `require_scenario` to check wheterh we ahve the addon in question installed.
+		//        But since [scenario]s are usually hidden behind `#ifdef CAMPAIGN_DEFINE ` it would be able to find them.
+		//        Investiage how this shoudl actually work.
```
soliton-, line 236:
```suggestion
		// FIXME: The mp code could use `require_scenario` to check whether we have the addon in question installed.
		//        But since [scenario]s are usually hidden behind `#ifdef CAMPAIGN_DEFINE` it would not be able to find them.
		//        Investigate how this should actually work.
```

gfgtdf: thx :)

soliton-:
I think this is a good idea. Not sure if we should do this for 1.18 or 1.20 though.

gfgtdf:
I see little reason now to have this in 1.18. Its not like addons are broken when hey require download when the didn't before.

gfgtdf:
I'm really sorry the comment was a typo s/now/not, i really didn't see it (but I also dont see how the sentence would make sense otherwise since the second part explains how it doesn't break anything) , and the comments in the code weren't directly related to this change (this is also mentioned in the first message) , just things I noticed while working on that code (and things that I didn't want to work on before 1.19)

The reason why I waited until now is mostly that I before I didn't test whether it worked correctly with mainline mods (as those are supposed to do no extra version Checks)

The  main reason why I merged now is that https://github.com/wesnoth/wesnoth/issues/8325 reminded me of it, (and that I didn't want to discuss that issue when it might actually change later)

Issue #7897, opened 2023-09-08
Schema ignores mandatory key constraint from super


soliton-:
Given WML schema test_schema.cfg:
```
[wml_schema]
    [type]
        name=id
        value="[a-zA-Z0-9_~$]+"
    [/type]
    [tag]
        name="root"
        min=1
        [tag]
            name="virtual~super"
            max=0
            [key]
                name=id
                type=id
                mandatory=true
            [/key]
        [/tag]
        [tag]
            name="mytag"
            min=1
            [tag]
                name="campaign"
                max=infinite
                super="virtual~super"
            [/tag]
        [/tag]
    [/tag]
[/wml_schema]
```

and test_schema_test.cfg:
```
[mytag]
    [campaign]
#        id=*
    [/campaign]
[/mytag]
```

`wesnoth -S test_schema.cfg --validate test_schema_test.cfg` outputs:
```
Validating test_schema_test.cfg against schema /home/USER/test_schema.cfg
validation succeeded
```

So the mandatory=true part is ignored. While the type part is not ignored as can be seen when uncommenting the id line.
```
error validation: Invalid value '*' in key 'id=' in tag [campaign]
 (expected value of type id) at test_schema_test.cfg:3
```

soliton-:
Also https://github.com/wesnoth/wesnoth/blob/e7252b212e4a1b6f92b9cbe8be67eef7295f112b/src/serialization/schema/tag.cpp#L367
should *really* be done. Any message is better than nothing.

soliton-:
I think we should try to get the schema as accurate as possible for 1.18.

Pentarctagon:
@CelticMinstrel can you look into this?

CelticMinstrel:
I don't think I can make any promises right now. If you're asking whether I can do it, the answer is of course yes, but it might not be until next year. I can't really predict a timeline.

rfsc-mori:
I've decided to try to fix to learn more about the project, and after a few days learning how the preprocessor, parser and schema validator worked I found a solution I was satisfied with. And then decided to open the PR in the case it was good enough fix. I apologize if I'm getting ahead of myself (since CelticMinstrel would be the one to look into this later on). I hope to not have caused two or more people to work on the same issue.

CelticMinstrel:
No worries. As mentioned in my previous comment, I had no specific plans to work on it.

PR #8136, opened 2023-12-20
Improve schema super validation (mandatory keys, cycles and missing super)

rfsc-mori:
Fixes issue #7897.

Before anything, I apologize if the PR is too big, I can split it if it's better for reviewing. All my changes were mostly in the same context, so I've decided to keep them together.

This PR aims to improve the validation of super tags, specifically to fix the mandatory keys and how that's affected by cycles in the super graph.

**Decision context:**
Implementing validation for mandatory super keys requires traversing the super tag tree and if there are any cycles this will lead to infinite loops. My first step was then to allow validating the schema itself to prevent super cycles. This was done using Boost Graph and DFS with back edge detection, treating the super dependency tree as a directed graph.

Now to the non-schema WML validation, I first had to change `wml_tag`'s `find_key` method, and decided to implement the same adjustment for `find_tag` despite not finding a way to reproduce the same issue. The issue was that if any key used in the WML file wasn't defined in the schema, it'd cause an infinite loop while traversing the super list. To solve this issue, I've added a bookkeeping vector to identify when a tag was processed twice. The non-schema cycle detection does not catch this because `find_key` calls happens before the final root tag validation.

Still in the `wml_tag` context, I've decided to add the super's path to the `super_refs_` member, and then expose its iterator in the same way the key and tag iterators are exposed. This allows for better diagnostics in the schema validator.

After the above preparation, I've implemented cycle detection for the schema validator as well, while parsing a non-schema WML file. After each tag is parsed, during its validation, I register them in the boost's graph list and search for cycles once we finally validate the root element.

Notes:
- This will catch super cycles even in the case of a schema and WML file that works normally, as some combinations may not lead to infinite loops.
- This won't catch super cycles for tags that aren't used in the WML file.
- The cycle is reported as an error following the same pattern as the other validation errors.

Following the reported issue this aims to solve, I've also decided to report an error when the WML schema specifies a super that does not exist. This is already handled by the schema_self_validator, but was not identified in a WML file validation. This has the same catch that the detection will only happen if the affected tag is used in the WML file.

Finally, I've addressed the mandatory keys validation in a way that includes all super keys, using the same bookkeeping method to prevent infinite loop in the case of a cycle. This also has the same issue where the super cycle cannot be detected beforehand, as all the other keys are validated before the root key is closed and validated.

**Other changes:**
During my experiments, I've decided to fix `tag.hpp`'s CRTP which was missing a parameter, causing it to pick a different template parameter (the default Map) in the base class context. This was done while experimenting to allow a `std::vector` to be used as the Map type parameter.

**Known risks:**
- I did not test the XCode project files as I don't have a system that supports XCode nor known an alternative way to validate if the file works. The script to add new files to the project didn't work very well for adding tests.
- This may be a breaking change for WML schemas that contains cycles or other conditions that would now cause an error.
- WML parsing performance may be affected due to the extra loops if there are _very_ big WML files.


soliton-:
Looks like this adds boost-graph as a dependency so needs more adjustments to the build system.

Presumably "boost-graph" needs to be added somwhere here:  https://github.com/wesnoth/wesnoth/blob/7921b685c3aa5632df8c57db626e2f80f12321ee/vcpkg.json#L22

CelticMinstrel:
This looks great! It appears to be causing CI failure due to some schema validation errors, though. We should probably fix those errors separately and merge the fixes before this PR.

rfsc-mori:
I'll check a bit later as well for my changes that caused the schema validation errors. I'm not too sure about the WML format in this aspect, but I was wondering if it's caused by the mandatory key validations and not handling links (?). I didn't have time yet to check the invalid key errors or the other "may not be used in" ones.

Also, I really apologize for not running the wml tests locally (I just found out they exist after submitting the PR). I should have noted that in the description. I'll run them locally (a bit later as well) and update here after I test the mandatory key constraint.

CelticMinstrel:
For what it's worth, if you're running on Linux you can validate _everything_ just by running this script:

https://github.com/wesnoth/wesnoth/blob/master/utils/CI/schema_validation.sh

Even if you're on Windows you can use the script as a guideline for what command-line paramaters to pass to the Wesnoth executable in order to validate everything.

Pentarctagon:
It looks like for CI failures it's:
* Ubuntu - new errors when running schema validation
* macOS - when building the unit test executable it has `error: Build input file cannot be found: '/Users/runner/work/wesnoth/wesnoth/src/tests/src/tests/wml/schema/test_schema_validator.cpp' (in target 'unit_tests' from project 'The Battle for Wesnoth')`. Just that somehow that doesn't get treated as a failure, so it only fails later once it tries to run the executable that failed to build. You should be able to use the `add_source_file` python script in the repo root to add new source files to xcode, though it may take a bit of trial and error to get it working.
* Windows - vcpkg.json needs to be updated, as was previously noted.

Otherwise though, thanks for this PR :)

Pentarctagon:
Also, new boost tests need to be added to `boost_test_schedule` in the repo root, otherwise they aren't actually run.

rfsc-mori:
I was testing this implementation against the schema.cfg and other WML schema files, and I'll need a few more days to correctly handle the special cases of schema_self_validator. The reason for the "missing key" errors was that there are a few cases that are not handled properly by my simpler mandatory key validation.

It isn't enough to just navigate the super dependency tree while checking if all mandatory keys are defined (e.g., "name" for most [tag]s) but to also follow links and handle [union]s to make a proper validation of the schema itself. I'm about halfway to get this working for links and unions by collecting all keys from the config, but it's a lot of special casing, so it's taking a bit long to reach a good solution.

The other two type of errors are also related to the find_key and find_tag changes' to keep track of cycles while validating the schema files. I still need to investigate those and come up with a proper solution, I've left them for later as the changes happens in tag implementation rather than on the schema's.

For now, I've marked the PR as a draft while I work on the changes.

CelticMinstrel:
I believe there may also be cases where the derived tag declares the same key or tag as the super. As I recall, that's supposed to completely override the declaration of that key or tag in the super.

rfsc-mori:
Thank you so much for mentioning the overrides! Handling that was the solution!

**I made a few changes today:**
1. Added boost-graph to vcpkg dependencies;
2. Identified what needs to be changed on the schema.cfg;
3. I ran the tests locally on my Linux machine and the results are the same as on CI;
4. macOS: I made a new attempt at fixing the build for the XCode project and re-added all my files again;
5. boost_test_schedule: I've double-checked and my tests are there!
6. I removed my `test_super_cycle_mandatory_missing` test as it is now impossible to trigger this behavior as the mandatory key validation will skip the tag when a cycle is found. This condition is already handled by super cycle detection and has its own test;
7. I changed the super cycle bookkeeping on tag.cpp find_key and find_tag methods;
8. Added handling for key overrides to fix mandatory keys validation.
9. Changed my uses of std::unordered_map to std::map to follow what we've been using in related code. This should also have better performance (?) as the sets are small.

Thank you for the help soliton-, Pentarctagon and CelticMinstrel!

**About the changes on the schema.cfg:** 
There were no overrides for some of the conditional tags such as if, else, elseif, then. I'll make a separate PR based off master with the fixes for review. This will solve all CI errors similar to `Missing key 'name=' in tag [case]`.

**About the changes on the XCode project:**
Running this command:
```bash
./add_source_file tests/wml/schema/test_schema_validator.cpp tests                                                                                                                                                 
```

Results in this error:
```
adding 'tests/wml/schema/test_schema_validator.cpp' to targets: ['tests']
 xcode targets: ['unit_tests']
Traceback (most recent call last):
  File "/home/USER/Repos/wesnoth/./add_source_file", line 244, in <module>
    add_to_xcode(filename, targets)
  File "/home/USER/Repos/wesnoth/./add_source_file", line 116, in add_to_xcode
    raise Exception(f"problem finding '{d}' group in '{groupname}'")
Exception: problem finding 'tests' group in 'src'
```

I could run the script properly after making this change:
https://github.com/rfsc-mori/wesnoth/compare/master..xcode_add_source_file_test

Which I did not include in this PR as I'm not sure if there's a reason for the tests group to be in the project root rather than in the src group.

**About the changes on tag.cpp:**
Cycles were wrongly detected in two different cases.

For find_key: When validating conditions, the tag may be enumerated more than once.
Solution: As the super tags are ignored in the first condition check, cycle detection is not needed.

For find_tag: Same as above. Additionally: Following links causes another tag traversal from the root tag, and any tag that has been visited before will trigger cycle detection.
Solution: Create a new container to reset cycle detection when following links.

**About the changes on mandatory keys validation:**
I've changed the routine to first collect all mandatory keys from the super tree and let any derivated tag override its super key.

This resulted in a cleaner validation and fixed most issues from the previous revision.

Thank you so much again, CelticMinstrel!

Note:
While splitting the codepath, I've decided to return an optional to signal a super cycle. As a result, this changes the behavior for when a mandatory key is missing in a schema containing a super cycle. Previously, I'd report that the key is missing, now this is not reported, and then the super cycle is reported later.

It is possible to re-add the diagnostic here, but it adds more complexity. I can work on it if you prefer it this way.

**About the still failing WML validations for ai/scenarios/scenario-lua_ai_unit_actions.cfg:142:**
This error is triggered:
```
Missing key 'guard_x=' in tag [micro_ai]
```

According to this schema:
https://github.com/wesnoth/wesnoth/blob/fcfc5dedf553e9ab1fa26a010f19b13586296380/data/schema/ai/micro.cfg#L145-L153

Caused here:
https://github.com/wesnoth/wesnoth/blob/fcfc5dedf553e9ab1fa26a010f19b13586296380/data/ai/scenarios/scenario-lua_ai_unit_actions.cfg#L142-L146

Related changelog entry:
https://github.com/wesnoth/wesnoth/blob/fcfc5dedf553e9ab1fa26a010f19b13586296380/changelog.md?plain=1#L5065-L5065

I'm not sure what's the correct approach here. I believe These keys should not be mandatory but the WML macro considers a glob/loc or x,y and I'm not sure what's the correct way to fix this situation.

rfsc-mori:
A MSVC-specific error was triggered on the CI. I'll look into it a bit later.

CelticMinstrel:
> It is possible to re-add the diagnostic here, but it adds more complexity. I can work on it if you prefer it this way.

It doesn't seem worth adding a diagnostic when it can only trigger if there's another, more serious error (the super cycle).

rfsc-mori:
In this last change, I fixed the error specific to MSVC mentioned above.

Our iterator class uses a reference type member, and this causes its [copy assignment to be deleted](https://en.cppreference.com/w/cpp/language/copy_assignment#Deleted_copy_assignment_operator). This affects all three iterators in the tag class: keys, tags and super (added with this PR).

On my current version of GCC and installed libraries, the `std::find_if` call simply increments the iterator copy obtained through copy constructor from the function parameter. This works because the [copy constructor is not deleted in this case](https://en.cppreference.com/w/cpp/language/copy_constructor#Deleted_copy_constructor).

On MSVC 16 2019 and 17 2022: The `std::find_if` implementation ends up calling `operator=(const T&);` when calling `_Seek_wrapped(our_iterator_copy, unwrapped_iterator);`.

Since my objective was always to check for the super existence, I've changed the call to `std::any_of` which has a clearer intent and doesn't trigger this error.

**Related:**
This also means that these iterators are breaking the requirements for being a [LegacyInputIterator](https://en.cppreference.com/w/cpp/named_req/InputIterator), because the required trait [LegacyIterator](https://en.cppreference.com/w/cpp/named_req/Iterator) needs it to be [CopyAssignable](https://en.cppreference.com/w/cpp/named_req/CopyAssignable).

I think the easiest fix would be to wrap the `const config& match` in a `std::reference_wrapper`, but this should probably be done in a separate PR. Do you agree with this solution?

**Regarding this PR:**
With this, I think the PR should be in a final-ish state, at least for review, but of course I'm open to changing anything!

I believe a small task list for what needs to be done before merging should be (unless I'm missing more things here, please let me know):
- Finish #8162.
- Fix Micro AI's stationed guard's guard_x and guard_y to be optional. I'm just really not sure about what's the correct approach here. There's also a .lua file with these keys as optional. Should these fields be optional along with guard_loc? Or only optional if guard_loc isn't specified? Is the lack of guard_x and guard_y on `scenario-lua_ai_unit_actions.cfg` intended?

Thanks again! Also, I apologize for the lengthy comments and for the number of revisions. I hope I'm contributing with good enough quality code, and I'm open to changes of any aspect!

CelticMinstrel:
> * Fix Micro AI's stationed guard's guard_x and guard_y to be optional. I'm just really not sure about what's the correct approach here. There's also a .lua file with these keys as optional. Should these fields be optional along with guard_loc? Or only optional if guard_loc isn't specified? Is the lack of guard_x and guard_y on `scenario-lua_ai_unit_actions.cfg` intended?

This is a bit complicated, huh? The keys are definitely optional, so one approach would just be to remove the special macro that's being used (the same one that's used for the station keys) and just specify the three keys as optional keys. That does accept some invalid combinations, though. I think it's invalid to specify only one of the x and y keys. It's probably also invalid to specify both loc and x+y. Maybe the logic could be "if guard_x or guard_y are found then declare both as required keys, otherwise declare guard_loc as an optional key"?

CelticMinstrel:
This is basically complete, right? The only blocker to merging is fixing new schema validation errors that were uncovered by these changes?

rfsc-mori:
Yes! But it might require changes if there is a way to cause an infinite loop still:

1) Confirming if there is any way to trigger a crash/infinite loop on `wml::find_tag` while defining links. The two main things I want to check in regards to this is writing a schema with list/elements that links to each other and trying to figure out if there's a way to somehow define links that would create a cycle for normal tags and links/super.

2) Fixing the new schema errors related to Micro AI.


I didn't manage to start working on these two things yet (also doesn't need to be me, of course). My NVMe died this week earlier during a crash while testing/stressing a new GPU, because the store is really pushing against my RMA request and my system was very unstable with it and throwing MCEs. And I'm also still organizing my time now that I'm back to work after holidays break.

If no one takes these two items and all goes well, I'm planning to start looking into this sometime on the next two weeks. But could also take a bit longer as I'm moving to another project at work and it might or not take more energy until I learn more about the platform. I'm sorry for the uncertainty and taking long on these last bits.

CelticMinstrel:
Okay. I think b9ff0c51fef should cover case 2. You could merge/rebase it into the PR to have CI run and double-check that it still works as expected with your changes.

Pentarctagon:
@rfsc-mori are you going to be able to come back to this?

rfsc-mori:
Should be able to soon. I've mostly finished my own things along with some other test assignments, there's only one test I'll take on next weekend, and I'm reviewing everything I can and also learning some more before I take it. I really need to prioritize this to make sure I don't waste a certain opportunity, I'll come back to this issue right after I'm done.

CelticMinstrel:
Regarding the Xcode project conflict, it's probably best to scrap all changes to the project and re-add whatever files need to be added using the add_source_file script.

Pentarctagon:
There were also some fixes made to the script in the mean time, so rebasing and using the current script would be a good idea.

pehjota:
> Looks like this adds boost-graph as a dependency so needs more adjustments to the build system.
> 
> Presumably "boost-graph" needs to be added somwhere here:
> 
> https://github.com/wesnoth/wesnoth/blob/7921b685c3aa5632df8c57db626e2f80f12321ee/vcpkg.json#L22

Also here:
https://github.com/wesnoth/wesnoth/blob/69fa5c199ee48aec5fb148d41222e07db0b42da0/CMakeLists.txt#L96

And here:
https://github.com/wesnoth/wesnoth/blob/69fa5c199ee48aec5fb148d41222e07db0b42da0/SConstruct#L373-L386

rfsc-mori:
I've rebased my branch against master:
1) Readded files to the Xcode project with the script. I had to keep the manual changes for the test file due script error: `Exception: problem finding 'tests' group in 'src'` when running `./add_source_file --target tests src/tests/wml/schema/test_schema_self_validator.cpp`.
Did I miss anything with the command? I can rebase again if this is not ok.

2) Added `graph` to `CMakeLists.txt` and `SConstruct`. Thank you!

I'll test for the other cases now.

CelticMinstrel:
Looks like the graph library isn't being installed in the CI script? At least, I can see the scons check for it concluded with "no".

rfsc-mori:
To add the graph library to the CI script, it is needed to edit `utils/dockerbuilds/CI/Dockerfile-base-2204-master` to update `wesnoth/wesnoth:2204-master` on docker hub. Is this done as a separate process or should be included on this PR?

Pentarctagon:
I can update and push the docker image if you can provide the library name.

rfsc-mori:
Of the places I found:
1) `.github/workflows/codeql.yml` possibly needs `libboost-graph1.74-dev` at line 56/57.

2) `utils/dockerbuilds/CI/Dockerfile-base-2204-master`:
line 8: `RUN apt install -y -qq ... libboost-graph1.74.0 libboost-graph1.74-dev`
line 30: `RUN tar -xvzf boost_1_74_0.tar.gz && cd boost_1_74_0 && ./bootstrap.sh --with-libraries=...,graph`

3) `utils/dockerbuilds/CI/Dockerfile-base-steamrt`:
line 11: `RUN cd /src/boost && ./bootstrap.sh --with-libraries=...,graph`

Both Dockerfiles still works after changes, for boost versions 1.74 and 1.66. Tested with:
- `docker build -f utils/dockerbuilds/CI/Dockerfile-base-2204-master .`
- `docker build -f utils/dockerbuilds/CI/Dockerfile-base-steamrt .`

I had to use another base image for the steamrt Dockerfile, but boost's bootstrap works with the graph library added.

Although the codeql build doesn't seem to fail currently, libboost-graph1.74-dev is not installed according to the CI log.

CelticMinstrel:
> To add the graph library to the CI script, it is needed to edit `utils/dockerbuilds/CI/Dockerfile-base-2204-master` to update `wesnoth/wesnoth:2204-master` on docker hub. Is this done as a separate process or should be included on this PR?

That may be true for the Linux builds, but I'm pretty sure the Windows and Mac builds both install all dependencies as a step in the CI.

Pentarctagon:
Windows had it added to the vcpkg dependencies list in this PR. Not sure about macOS, but if it's working then I assume Mac Compile Stuff just happens to already have it.

CelticMinstrel:
> if it's working then I assume Mac Compile Stuff just happens to already have it.

That's a good point – it probably needs to be added there too, right? I don't think the CI uses Mac Compile Stuff.

Pentarctagon:
@rfsc-mori I'll do (2) and (3), but (1) can just as easily be done in this PR, I think?

rfsc-mori:
I've added handling for link cycles. I had to add the check after building the `types_` list because there's currently no way to share the needed state between all related `wml_type` instances during the match call on `schema_validator::validate_key`.

I really cannot think of any way to trigger a cycle behavior on `wml_tag::find_tag` from a valid wml schema file, and the element list cycle ended up triggering a crash elsewhere. Unless an example of something I missed shows up, this point should be done? I've also added a unit test with an example schema of the link cycle scenario.

Also added `libboost-graph1.74-dev` to `codeql.yml`.

Thank you!

Sorry it took so long to come back to this. If there's any more required changes and/or suggestions of any nature, nitpicks etc please let me know and I'll try to address it as soon as possible!

Pentarctagon:
Can you rebase against current master and push again, to retrigger the CI? One thing was missed at least, and also it looks like something weird is going on - one Ubuntu job succeeded but the other wasn't able to find the graph dependency somehow.

SConstruct
```diff
@@ -381,7 +381,8 @@ if env["prereqs"]:
 	CheckAsio(conf) & \
 	conf.CheckBoost("thread") & \
         conf.CheckBoost("locale") & \
-        conf.CheckBoost("filesystem") \
+        conf.CheckBoost("filesystem") & \
+        conf.CheckBoost("graph") \
```
Pentarctagon, line 385: Should be part of `have_client_prereqs` below.

rfsc-mori: Done! Thank you!

Pentarctagon:
It looks like the issue is that the scons check doesn't work - it's failing even when the boost graph library is present.

rfsc-mori:
I've rebased against master to pull the wesnoth Dockerfile changes and added graph: `graph/graph_traits.hpp` to scons check, thank you!

I've attempted to run the scons build locally and it successfully detects boost-graph now. The build failed to run the mariadb_config script but I believe it should work on the properly configured CI.

CelticMinstrel:
No additional comments.

Pentarctagon:
@rfsc-mori thanks for working on this!

rfsc-mori:
Thank you for the help, for accepting and for the feedback!

CelticMinstrel:
This PR doesn't require any changes to [the documentation](https://wiki.wesnoth.org/SchemaWML), right?

Issue #2218, opened 2017-11-19
Horizontal scroll direction is reversed


dhardy:
Before the last system update, touchpad scrolling of the map worked perfectly. After the update (Wesnoth 1.13.10), horizontal scrolling works in the wrong direction (vertical scrolling is still correct). It's weird having only one axis reversed. I don't see any config for this.

sevu:
I recall this happening to me too.

Vultraz:
When I swipe left, the map goes to the right. Is that not the previous behavior?

dhardy:
It's not the previous behaviour I had. Now the map moves down when I swipe down (same direction) but left when I swipe right. It's pretty odd. (I never did get why we needed to reverse touchpad scrolling directions just because touchscreens work differently; one is a "scrolling" method the other is "drag a point".)

Vultraz:
That’s a good analogy, I guess. Now we just neee to decide which scrolling method to commit to.

dhardy:
Well, either stick to the old standard (and tweak your touchpad if you want different in all your apps) or add an option. Because all my other apps scroll in the same direction as I swipe.

AI0867:
Apple decided to make touchpads on macOS work the same as touchscreens on iOS at some point, and microsoft followed suit in windows 8 (even with scrollwheels I believe). Some linux software also does this, KDE (at least the versions I've run) does not.

It's all rather annoying.

dhardy:
I know, I just assumed the only sane way to implement the switch would be by swapping touchpad scroll output. (Although even that is not sane — queue *scroll down to zoom in* on maps interpreting wheel events the usual way.) Ugh, what a can of worms.

vgaming:
// Linux XFCE environment defines scrolling as "scroll down = text goes down", but there is an option to reverse scrolling behavior, called "Reverse scroll direction". This option is the approach that seems sane. It's not app-level, it's system-level.

I would be very happy if wesnoth would behave same as all others programs on my OS/environment. So that I could also reverse the behavior on all apps simultaneously by a single click (if I'd want to).

GregoryLundberg:
I don't know if it's such a good idea querying the DM for options unless it's explicitly an option supported by all DMs on Windows, macOS and every flavor of Linux out there.

vgaming:
[...]

PR #8137, opened 2023-12-20
Fix horizontal scrolling regression (fixes #7404)

### LINKED ISSUE #7404: With touchpad horizontal scrolling is inverted
When I use the touchpad, scrolling upwards scrolls upward, downward scrolls downwards. But scrolling right scrolls to left.

1.16 and c9eae7c affected.
When using the OS' settings to activate `natural scrolling`, all four directions are inverted, so the bug remains.

Using Linux with KDE on top of X11 with libinput.
Other programs are not affected by this problem.
Scrolling by moving the mouse pointer via the touchpad to a side or corner of the screen works correct as well.

pehjota:
The old issue #2218 was actually a bug in SDL, fixed on X11 in version 2.0.18 and on Wayland in version 2.0.20.  The hardcoded workaround in pull #2481 (commit 4bc4373) caused a regression in fixed SDL versions.

This fix is similar to the workaround in widelands/widelands#5394 committed as widelands/widelands@67db32a.

Tested with SDL 2.0.14 and 2.28.5.

### DIFF
--- a/src/controller_base.cpp
+++ b/src/controller_base.cpp
@@ -232,8 +232,7 @@ void controller_base::handle_event(const SDL_Event& event)
 		break;
 
 	case SDL_MOUSEWHEEL:
-		// Down and right are positive in Wesnoth's map, but
-		// mouse_handler_base::mouse_wheel() inverts both axes.
+		// Down and right are positive in Wesnoth's map.
 		// Up and right are positive in SDL_MouseWheelEvent on all
 		// platforms:
 		//     https://wiki.libsdl.org/SDL2/SDL_MouseWheelEvent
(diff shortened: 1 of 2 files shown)


src/controller_base.cpp
```diff
@@ -232,11 +232,32 @@ void controller_base::handle_event(const SDL_Event& event)
...
+		{
+			int xmul = -1;
+#if !defined(_WIN32) && !defined(__APPLE__)
+			const char* video_driver = SDL_GetCurrentVideoDriver();
+			SDL_version ver;
+			SDL_GetVersion(&ver);
+			if(video_driver != nullptr && ver.major <= 2 && ver.minor <= 0) {
+				if(std::strcmp(video_driver, "x11") == 0 && ver.minor < 18) {
+					xmul = 1;
+				} else if(std::strcmp(video_driver, "wayland") == 0 && ver.minor < 20) {
```
soliton-, line 254: Shouldn't this and above check `ver.patch`?

pehjota: Oops, yes, somehow I copied and pasted wrong from code I wrote elsewhere.  Fixed, thanks.

soliton-:
This looks good to me. If someone can confirm that it works as expected that'd be cool but even if not I think we should just merge this.

pehjota:
I see `mouse_handler_base::mouse_wheel()` calls virtual methods such as `mouse_wheel_left()` and `mouse_wheel_right()`, but the X-axis calls are incorrect.  It looks like the original intent in commit 821b009 was to follow SDL's coordinate signs (positive up and right), but this later changed to Wesnoth's map coordinate signs (positive down and right) in commit ea3bd56, and commit dfe2f33 reversed the X axis (positive down and left, opposite SDL) without updating the `mouse_wheel_*()` calls.  Although commit c912f7e removed the definitions of these methods, so this currently doesn't really matter.

I've pushed a second commit adding two more lines to the comment block, which should probably be squashed into the first commit.

And a third commit, which you can either merge or ignore, makes `mouse_handler_base::mouse_wheel()` use the same coordinate signs as the map itself and also fixes the `mouse_wheel_*()` virtual method calls.

sevu:
Thank you!
Tested it and works as expected.

Pentarctagon:
This can be merged then?

Pentarctagon:
Thanks!

pehjota:
Thanks for merging! Can/should this be cherry picked to 1.16?

Pentarctagon:
Up to you if you think it's worth it, given 1.18 is planned to be released in March and the next (final) 1.16 release that would include this fix will be released at the same time.

pehjota:
OK, I cherry picked, built, tested again with broken and fixed SDL versions, and sent a PR.  Too bad I missed 1.16.11 by just two days, but it'll be nice to at least get it into the final 1.16 release.

PR #8143, opened 2023-12-22
Modify shema validation for more consistency

newfrenchy83:
In the validation schema abilities-value in abilities.cfg contains not only the numerical attributes but also active_on and apply_to, while these two deniers are only used by [resistance], and the abilities used as weapons are validated by the schema validation of special weapons, and the abilities [leadership], [heals], [regenerate] and [illuminates] also use all numerical attributes.


Pentarctagon:
Makes sense to me.

PR #8146, opened 2023-12-23
1.18 - wooden-frame, non-village windmill

### LINKED ISSUE #8064: Windmills confuse players / Map makers dont want to use windmills because they confuse players
Windmills look like a villages, but arent villages in general, while being villages sometimes (because they look like villages and people give them village status). This is a problem because it often confuses players if they dont look at the tooltip and because they mostly dont look there (me included) map makers abstain from using that terrain on their maps thus making this terrain less useful than it could be.

doofus-01:
No API changes, just change the images for `^Wm`.
Fixes #8064 and 82e1e06 (it sure took a while, but I had always thought the ambiguity was intentional)
Not finished, but this is the first draft
Screenshot_20231223_windmills


stevecotton:
Maybe non-village ones should have the sails stationary, because the brake is on unless someone is using them.

Probably won't save any time on the artwork, but it's a lore reason to have less variations / less frames.

doofus-01:
Less frames would certainly be nice, but the animation is what the windmill has going for it, so I'd say we should leave the old windmill frames for the villages as-is.  Generally, 18 frames is annoying, but we don't want to make these windmills into turboprops, so a somewhat slow spin and lots of frames is required.

Speaking of reducing frames, the windmill frames are duplicated in `images/scenery`, as are the campfire and brazier (in `images/items`) frames.  It's silly, but I think it's because these were `[item]` decorations before they were terrains, and because they are such old images, deleting them will damage a lot of legacy content.  So, probably best to leave them alone for now, but maybe they can be removed for BfW 1.20.

CelticMinstrel:
> I had always thought the ambiguity was intentional

I really don't know if the ambiguity was intentional. However, I personally like the idea of them being differentiated, and this new windmill looks pretty nice as well.

I do notice that the sails are actually pretty much the same for both of them, which makes me think it'd be better to separate out the animated sails and place them on top of the base. Maybe it's not worth the extra effort now that they're already done though.

cooljeanius:
> > I had always thought the ambiguity was intentional
> 
> I really don't know if the ambiguity was intentional. However, I personally like the idea of them being differentiated, and this new windmill looks pretty nice as well.

I agree that having them differentiated is good, but I still think it would have been nice to resolve the question of whether the ambiguity was intentional or not before merging... I just tried checking `git blame` on [data/core/terrain-graphics.cfg](https://github.com/wesnoth/wesnoth/pull/8146/files#diff-4237ca2656ecf57e1035956306fa5dc332cbfa439d99ec88dfc58e0182b34617), but didn't really have much success tracking the changed macros through their reorganizations...

CelticMinstrel:
The commit that introduced the windmill village didn't really have any explanation. I got the impression that the committer just wanted it to be a village.

Issue #1394, opened 2017-05-08
transforming a unit with resilient/strong trait lowers its HP (GNA #22178)


wesnoth-bugs:
**[Original submission](https://gna.org/bugs/index.php?22178)** by vanagandr on 2014-06-11

When you transform a unit that has a certain combination of traits (among them resilient or strong, but quick,resilient behaves differently from resilient,quick) the unit might lose some HP.

I made a test scenario "transform right click test", you can transform 5 spearman with the right-click menu or hotkey t. For me the spearmen ended up with (from top to bottom):

40/41

37/37

40/42

39/39

34/39

HP respectively (note that the it only shows the updated HP once you move the mouse away from the unit and back over it). 

The spearmen just transform to their own unit type again (i.e. Transform_Spearman -> Transform_Spearman), so should retain their current HP, but don't.

(Reproduced on Windows 8)
Release: 1.11.15
Priority: 5 - Normal
Severity: 2 - Minor

wesnoth-bugs:
**Modified on 2014-06-11**

**vanagandr added attachment:** [transform_test.zip](http://gna.org/bugs/download.php?file_id=20981)

wesnoth-bugs:
**Modified on 2014-06-11**

**vanagandr wrote:**

> Sorry, I made a copy-paste mistake in my first attachment. The scenario uses Animate_Spearman from my earlier bug report, so if you haven't downloaded that it most likely won't load. I attached a fixed version.
> 
> (file #20982)

**vanagandr added attachment:** [transform_test_fixed.zip](http://gna.org/bugs/download.php?file_id=20982)

wesnoth-bugs:
**Modified on 2015-06-08**

**gfgtdf changed item group:** None of the others -> WML

Pentarctagon:
Can confirm this still happens on 1.15.16.

CelticMinstrel:
So, this is about _current_ hp, not max hp, right?

Pentarctagon:
Right.

PR #8147, opened 2023-12-25
 make sure transform_unit doesn't heal the unit

gfgtdf:
previously it could happen that removing an object applies the status healing, in particular of the default amlas (which are meant to only heal the unit once)

[tranform_unit], unit:transform(),  [remove_object]  and unit:remove_modifications() are required not to change a unit current hp see https://wiki.wesnoth.org/DirectActionsWML#.5Btransform_unit.5D , especially not in a unpredictable manner like this.

### DIFF
--- a/src/units/unit.cpp
+++ b/src/units/unit.cpp
@@ -183,18 +183,24 @@ namespace
 		}
 	}
 
-	auto stats_storage_resetter(unit& u)
+	auto stats_storage_resetter(unit& u, bool clamp = false)
 	{
 		int hitpoints = u.hitpoints();
 		int moves = u.movement_left();
+		int attacks = u.attacks_left(true);
 		int experience= u.experience();
 		bool slowed= u.get_state(unit::STATE_SLOWED);
 		bool poisoned= u.get_state(unit::STATE_POISONED);
 		return [=, &u] () {
-			//u.set_movement(std::min(u.total_movement(), moves));
-			u.set_movement(moves);
-			//u.set_hitpoints(std::min(u.max_hitpoints(), hitpoints));
-			u.set_hitpoints(hitpoints);
+			if(clamp) {
+				u.set_movement(std::min(u.total_movement(), moves));
+				u.set_hitpoints(std::min(u.max_hitpoints(), hitpoints));
+				u.set_attacks(std::min(u.max_attacks(), attacks));
+			} else {
+				u.set_movement(moves);
+				u.set_hitpoints(hitpoints);
+				u.set_attacks(attacks);
+			}
 			u.set_experience(experience);
 			u.set_state(unit::STATE_SLOWED, slowed && u.get_state("unslowable"));
 			u.set_state(unit::STATE_POISONED, poisoned && u.get_state("unpoisonable"));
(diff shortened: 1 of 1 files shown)


src/units/unit.cpp
```diff
@@ -182,6 +182,24 @@ namespace
 			++cur;
 		}
 	}
+
+	auto stats_storage_resetter(unit& u)
+	{
+		int hitpoints = u.hitpoints();
+		int moves = u.movement_left();
```
ProditorMagnus, line 189: attacks_left? more of similar kind?

gfgtdf: fixed

Vultraz:
Since all the variables are local to the lambda, could you declare them in the capture list?

gfgtdf:
I don't really see the advantage of doing so (if i understood your correctly).

CelticMinstrel:
I also don't see an advantage in doing so.

gfgtdf:
Also fixes #1394

src/units/unit.cpp
```diff
@@ -994,22 +1019,12 @@ void unit::advance_to(const unit_type& u_type, bool use_traits)
...
-	if(get_state("unpoisonable")) {
-		set_state(STATE_POISONED, false);
-	}
-	if(get_state("unslowable")) {
-		set_state(STATE_SLOWED, false);
-	}
+	// Make sure apply_modifications() didn't attempt to heal the unit (for example if the unit has a default amla.).
+	ss();
 	if(get_state("unpetrifiable")) {
 		set_state(STATE_PETRIFIED, false);
```
stevecotton, line 1025: Why is petrification treated differently than poison and slow?

gfgtdf: I just wasn't sure about petrification since i don't know how its usually used, with slow and poisoned i can easily say "it usually applied via attacks and not via objects", for petrification i cant. (I know that there exists a petrifies weapon special, bu i just never have seen it and by default is just compltely unbalances/gamebeaking)

ProditorMagnus: I think it makes sense to treat petrify and slow the same. And in 1.19 change default petrifying to limited turn duration.

CelticMinstrel: We can't just change petrification to limited duration. Several multiplayer scenarios rely on it being permanent.

stevecotton:
I've just marked my comment as resolved, is this waiting on anything else?

(Sorry about the time it took me to do that)

src/units/unit.cpp
```diff
@@ -182,6 +182,30 @@ namespace
...
+				u.set_movement(std::min(u.total_movement(), moves));
+				u.set_hitpoints(std::min(u.max_hitpoints(), hitpoints));
+				u.set_attacks(std::min(u.max_attacks(), attacks));
+			} else {
+				u.set_movement(moves);
+				u.set_hitpoints(hitpoints);
+				u.set_attacks(attacks);
+			}
+			u.set_experience(experience);
+			u.set_state(unit::STATE_SLOWED, slowed && u.get_state("unslowable"));
```
ProditorMagnus, line 205: shouldnt it be !u.get_state?

gfgtdf: indeed, i must have made a mistake when copying it.

Pentarctagon:
Could a unit test be added for this?

stevecotton:
Probably should have a suite of tests (transforming into / out of `unslowable`, `unpoisonable`, etc), not just one.

I'm dithering over whether it's better to merge tonight even without unit tests, so we get people alpha-testing for a month before RC1.

ProditorMagnus:
As bugfix instead of feature, burden of test creation is on the addon author who is unhappy with the way it happens to work now, as reasoning to get the development priority/approval. Since such convincing was not needed here, there are no test expected.

stevecotton:
I've done a quick test, and this seems to fix #8324.

Clarification of my previous comment: while a unit test would be nice, I think it's worth merging without.

Issue #8149, opened 2023-12-25
[Utbs]Is it normal that Kaleh or Nym's level never changes?


newfrenchy83:
### Game and System Information

- **Version:**1.17.24
- **Downloaded from:**build from source
- **Build info:**
- **OS:**widows7

### Description of the bug

When I try to evolve Kaleh, his stas increases (xp and hp) as well as his skills, but the level displayed remains 1, is this intended or not?

### Steps to reproduce the behavior

1Load game
2choose Utbs campaign or load savegame without external resource
3.evolve Kaleh several times with debug mode

### Expected behavior

We should expect the level to increase with each amla (every two amla) but this is no longer the case

Hejnewar:
It is not but it is on my todo list for when I finally have time and will to finish balancing UtBS (probably January).

newfrenchy83:
> It is not but it is on my todo list for when I finally have time and will to finish balancing UtBS (probably January).

[Youth.cfg.txt](https://github.com/wesnoth/wesnoth/files/13767075/Youth.cfg.txt)

this file contain an strategy for esolve that if it can help.

PR #8152, opened 2023-12-25
fix leveling for Kaleh and Nym in Utbs

newfrenchy83:
Fix https://github.com/wesnoth/wesnoth/issues/8149 issue

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/units/quenoth/Youth.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/quenoth/Youth.cfg
@@ -894,6 +894,11 @@ _"Nym is Kaleh’s childhood friend, a young, rebellious lady with quick wits an
     movement_type=quenoth_foot
     movement=6
     hitpoints=32
+    # The evolution of this unit is carried out only with AMLAs so that it takes two amlas
+    # to reach level two then two more for level 3.
+    # As the unit variables do not include the number of amla achieved,
+    # the maximum experience is the only one that can be directly used in advancement without resorting to an external lua type function.
+    # Do not modify this value without also modifying the max_experience checks in the custom AMLAs
     experience=16
     level=1
     profile=portraits/kaleh.webp
--- a/data/campaigns/Under_the_Burning_Suns/units/quenoth/Youth.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/quenoth/Youth.cfg
@@ -44,6 +44,38 @@ _"Nym is Kaleh’s childhood friend, a young, rebellious lady with quick wits an
         increase=8
 #endif
     [/effect]
+    [effect]
+        [filter]
+            level=1
+#ifdef EASY
+            formula="max_experience > 24"
+#endif
+#ifdef NORMAL
+            formula="max_experience > 28"
+#endif
+#ifdef HARD
+            formula="max_experience > 33"
+#endif
+        [/filter]
+        apply_to=level
+        set=2
+    [/effect]
+    [effect]
+        [filter]
+            level=2
+#ifdef EASY
+            formula="max_experience > 41"
+#endif
+#ifdef NORMAL
+            formula="max_experience > 46"
+#endif
+#ifdef HARD
+            formula="max_experience > 54"
+#endif
+        [/filter]
+        apply_to=level
+        set=3
+    [/effect]
 #enddef
 
 #define KALEH_ADVANCEMENTS


gfgtdf:
No reason to create a new type, just use apply_to=level

newfrenchy83:
execpt what i want use two advancement for increase level to 1, and i can done that what with change unit_type

Hejnewar:
You can filter by xp, I know it changes with difficulty but thats fine I can always balance / fix that latter.

newfrenchy83:
> You can filter by xp, I know it changes with difficulty but thats fine I can always balance / fix that latter.

i filtered with hp instead, sorry.

newfrenchy83:
> You can filter by xp, I know it changes with difficulty but thats fine I can always balance / fix that latter.

it's done

gfgtdf:
you cant to filter for `max_experience`instead of `experience`

newfrenchy83:
> you want to filter for `max_experience`instead of `experience`



> you want to filter for `max_experience`instead of `experience`

fixed

soliton-:
Leveling according to arbitrary XP values seems like a good way to cause future confusion when whatever this is based on is inevitably changed.

Hejnewar:
If only there was variable or stat that contained the number of AMLAs the unit took already (maybe it is possible in lua), or if setting unit variables using advancement or effect tag was possible, this would greatly simplify a lot of things. So until that is possible I will continue using xp especially since I'm already using it for unlocking one specific Amla.

gfgtdf:
> If only there was variable or stat that contained the number of AMLAs the unit took already (maybe it is possible in lua), or if setting unit variables using advancement or effect tag was possible

This is definitely possible and i use it a lot in my addons, but you have to write a custom effect (in lua), for example in WC2:
https://github.com/wesnoth/wesnoth/blob/master/data/campaigns/World_Conquest/lua/game_mechanics/effects.lua#L162
It's important that you use the `mods` container because that one is cleared whenever the unit is reset (so that the effects don't increase everytime the unit is reset.)

gfgtdf:
So a possible  lua code could look like 

```
function wesnoth.effects.utbs_set_level(u, cfg)
  local n_amla = u.variables["utbs_namla"] or 0
  n_amla = n_amla + 1
  u.variables["utbs_namla"] = n_amla
  if n_amla > 2 then
    u.level = 2
  end
  if n_amla > 4 then
    u.level = 3
  end
end
```

newfrenchy83:
If number of amla for increase level is know then value of max xp of unit just before lave-linge should could be calulated

newfrenchy83:
> So a possible lua code could look like
> 
> ```
> function wesnoth.effects.utbs_set_level(u, cfg)
>   local n_amla = u.variables["utbs_namla"] or 0
>   n_amla = n_amla + 1
>   u.variables["utbs_namla"] = n_amla
>   if n_amla > 2 then
>     u.level = 2
>   end
>   if n_amla > 4 then
>     u.level = 3
>   end
> end
> ```
@Hejnewar 2 amla for level 2 and 2other amla for level 3, will you agreed?

Hejnewar:
Its fine.

newfrenchy83:
@gfgtdf could you review this PR, please?

CelticMinstrel:
> If only there was variable or stat that contained the number of AMLAs the unit took already (maybe it is possible in lua)

Why would you need Lua for that? Isn't it just `$unit.modifications.advancement.length`? Assuming the unit is stored in `$unit` obviously.

newfrenchy83:
> > If only there was variable or stat that contained the number of AMLAs the unit took already (maybe it is possible in lua)
> 
> Why would you need Lua for that? Isn't it just `$unit.modifications.advancement.length`? Assuming the unit is stored in `$unit` obviously.

No need to count the number of amla, just determine the maximum experience of the unit before the advancement which should make it level up, this remains simpler for the code and with better integration.

gfgtdf:
> So a possible lua code could look like
> 
> ```
> function wesnoth.effects.utbs_set_level(u, cfg)
>   local n_amla = u.variables["utbs_namla"] or 0
>   n_amla = n_amla + 1
>   u.variables["utbs_namla"] = n_amla
>   if n_amla > 2 then
>     u.level = 2
>   end
>   if n_amla > 4 then
>     u.level = 3
>   end
> end
> ```

This code was wrong as it should use the mods container, corrected code

 ```
 function wesnoth.effects.utbs_set_level(u, cfg)
   local n_amla = u.variables["mods.utbs_namla"] or 0
   n_amla = n_amla + 1
   u.variables["mods.utbs_namla"] = n_amla
   if n_amla >= 2 then
     u.level = 2
   end
   if n_amla >= 4 then
     u.level = 3
   end
 end
 ```

gfgtdf:
> @gfgtdf could you review this PR, please?

I'd prefer if @Hejnewar does it, as i know very little about the utbs campaign, if you have specific question about lua/wml i can answer those though.

Hejnewar:
For me its fine if I have to review it form the campaign side, but I cant review code perspective because my lua knowledge is almost non existent. So as long as code is fine im fine as well.

Hejnewar (approved):

newfrenchy83:
> For me its fine if I have to review it form the campaign side, but I cant review code perspective because my lua knowledge is almost non existent. So as long as code is fine im fine as well.

@Pentarctagon could you review this PR,please?

Pentarctagon:
The current implementation looks like it's WFL rather than lua?

newfrenchy83:
> The current implementation looks like it's WFL rather than lua?

yes but it is appproved, ma_experience value don't choose by chance but after level up Youth for read values for each amla

Pentarctagon:
Well, as long as it's been tested to actually work, it looks fine to me.

newfrenchy83:
> Well, as long as it's been tested to actually work, it looks fine to me.

it's tested in hard mode and other and this work

Pentarctagon:
@gfgtdf @soliton- does this need to use the lua approach?

gfgtdf:
> For me its fine if I have to review it form the campaign side, but I cant review code perspective because my lua knowledge is almost non existent. So as long as code is fine im fine as well.

The current code doesn't use any lua though.

> @gfgtdf @soliton- does this need to use the lua approach?

The current approach is not how i would do it, but I don't have a problem with merging this.

soliton-:
Since the goal appears to be to change level after picking two AMLAs that’s what should be implemented. Manually calculating the XP that might be equivalent to this is just weird.

CelticMinstrel:
@gfgtdf's Lua looks frankly bizarre to me. I don't understand why unit variables need to be involved in this at all.

gfgtdf:
> @gfgtdf's Lua looks frankly bizarre to me. I don't understand why unit variables need to be involved in this at all.


Feel free to provide a better solution.

CelticMinstrel:
Well, if taking the same basic approach of using a custom "set level" effect, I think it could look something like this:

```lua
function wesnoth.effects.utbs_set_level(u, cfg)
  local n_amla = #u.advancements + 1
  if n_amla >= 4 then
    u.level = 3
  elseif n_amla >= 2 then
    u.level = 2
  end
end
```

gfgtdf:
> Well, if taking the same basic approach of using a custom "set level" effect, I think it could look something like this:

This approach has a problem, you modify the unit based in the state _after_ the modification (number of modifications), in practice  this causes problems for example in the following case:

1) a unit gets an amla that uses `apply_to=utbs_set_level` effect (at this point it doesn't gain a level because it only has one advancement)
2) a unit get an amla that doesnt have `apply_to=utbs_set_level` effect (maybe form some random [modification])

Now the unit is still level 1, but rebuilding (like removing an object) causes the unit to gain a level.

This example is a bit constructed because  in the campaign all advancement will have that effect, but for custom effects its still good practice to only make changes bases the previous state of the unit, to avoid running into bugs.

CelticMinstrel:
> This example is a bit constructed because in the campaign all advancement will have that effect

Kinda feels like you're overengineering it here. If it works in the campaign, isn't that good enough?

By the way, are you implying that the unit passed to an effect already has all its modifications added to the modifications array? If that's the case it actually simplifies things a bit – there's no need to add 1 to the number of advancements.

I'd also suggest that, rather than adding this effect to every advancement, it might be easier to instead add it to an object that's placed on the unit when it's first created. There's no real need to run it once per advancement when rebuilding the unit.

gfgtdf:
> By the way, are you implying that the unit passed to an effect already has all its modifications added to the modifications array?

I'm actually no sure on when the effect is applied for te first time, but i'm mostly talking about the point when a unit is rebuilt, because then it definitely has all modifications added to the array (even the ones not applied yes). 

> Although, that does assume that adding an advancement rebuilds the unit

It doesn't, but


imo it would be good if would give umc authors at least the change to do so, but probably out of scope for 1.18, i know that `apply_to=type` exists but it has other problems which are caused from the attempt to make [remove_object] work on these effects. In fact all suggestions except the original one one which used apply_to=type had a issue where the resiliant trait didn't work as most people would expect because the [advancement]s that set the level were applied after the trait, so at the time the trait is applied the unit is still level 1.

> Kinda feels like you're overengineering it here. If it works in the campaign, isn't that good enough?

I have, in the past dealt with various reports of "unit losing hp when removing an object" and similar issues of modifications, so i do think this is a common source of bugs.

If i have one solution that works by what i would call the default way effects currently work ("all modifications are applied in order so rebuilding the unit leads to the same result because all effects only depend on the previous unit stats") and one sulution that only works by the specific circumstances of the scenario that all amlas use that effect and that none of the other effects depend on the level, i would usually call the first solution simpler even if it involves one helper variables.

(There can of course be reasons to use the other implementation, but simplicity is not one of them)

CelticMinstrel:
> i'm mostly talking about the point when a unit is rebuilt, because then it definitely has all modifications added to the array (even the ones not applied yes).

I think in that case it would be good to ensure that this is true also when a modification is first applied. Or, alternatively, to ensure it's _not_ true when the unit is rebuilt (and also not true when first applied, obviously). So rebuilding a unit could be implemented by essentially removing all modifications and adding them back one by one. It would have to use the `all_children_range` order for that, of course, as an object and an advancement might have depended on each other in some way in order to obtain the final stats.

newfrenchy83:
From what I understand from this debate, is that no one seems to provide a satisfactory and simple alternative to my own method so how about you merge this PR and propose another solution later if you find one.

soliton-:
The implementation in this PR is obscure (there is no hint what the effect is trying to achieve or where the numbers come from). If this is truly the best that can be done (which I highly doubt since Kaleh has had AMLAs since years and I'm not aware it ever been done like this before) then it seems we have a serious problem with writing robust effects that should be resolved ASAP.

Pentarctagon:
If there is no better alternative yet, the please at least add a comment explaining the assumptions being made above https://github.com/wesnoth/wesnoth/blob/963029dd7cec3958f37e006b6a7cd1438b2d37bc/data/campaigns/Under_the_Burning_Suns/units/quenoth/Youth.cfg#L865

newfrenchy83:
> If there is no better alternative yet, the please at least add a comment explaining the assumptions being made above
> 
> https://github.com/wesnoth/wesnoth/blob/963029dd7cec3958f37e006b6a7cd1438b2d37bc/data/campaigns/Under_the_Burning_Suns/units/quenoth/Youth.cfg#L865

comment added

data/campaigns/Under_the_Burning_Suns/units/quenoth/Youth.cfg
```diff
@@ -862,6 +888,10 @@ _"Nym is Kaleh’s childhood friend, a young, rebellious lady with quick wits an
     movement_type=quenoth_foot
     movement=6
     hitpoints=32
+    # The evolution of this unit is carried out only with AMLAs so that it takes two amlas
+    # to reach level two then two more for level 3.
+    # As the unit variables do not include the number of amla achieved,
+    # the maximum experience is the only one that can be directly used in advancement without resorting to an external lua type function.
```
Pentarctagon, line 894: Please also add:
`# Do not modify this value without also modifying the max_experience checks in the custom AMLAs`

newfrenchy83: it's done

newfrenchy83:
@Pentarctagon PR updated to integrate modification expereience because difficulty.

Pentarctagon:
Opened #8179 for the question of how to determine the AMLA count more reliably.

PR #8158, opened 2023-12-27
[1.16] Backport #8137: Fix horizontal scrolling regression (fixes #7404)

### LINKED ISSUE #7404: With touchpad horizontal scrolling is inverted

pehjota:
The old issue #2218 was actually a bug in SDL, fixed on X11 in version 2.0.18 and on Wayland in version 2.0.20.  The hardcoded workaround in pull #2481 (commit 4bc4373) caused a regression in fixed SDL versions.

This fix is similar to the workaround in widelands/widelands#5394 committed as widelands/widelands@67db32a.

Also make mouse handler use same coordinate signs as map and fix mouse_wheel_*() virtual method calls, which have been wrong in the X axis since commit dfe2f33 (and unused since commit c912f7e).

Tested with SDL 2.0.14 and 2.28.5.

Cherry picked from commit ab4001d (pull #8137) in the master branch (1.17).

### DIFF
--- a/src/controller_base.cpp
+++ b/src/controller_base.cpp
@@ -230,11 +230,33 @@ void controller_base::handle_event(const SDL_Event& event)
 		break;
 
 	case SDL_MOUSEWHEEL:
-#if defined(_WIN32) || defined(__APPLE__)
-		mh_base.mouse_wheel(-event.wheel.x, event.wheel.y, is_browsing());
-#else
-		mh_base.mouse_wheel(event.wheel.x, event.wheel.y, is_browsing());
+		// Down and right are positive in Wesnoth's map.
+		// Up and right are positive in SDL_MouseWheelEvent on all
+		// platforms:
+		//     https://wiki.libsdl.org/SDL2/SDL_MouseWheelEvent
+		// Except right is wrongly negative on X11 in SDL < 2.0.18:
+		//     https://github.com/libsdl-org/SDL/pull/4700
+		//     https://github.com/libsdl-org/SDL/commit/515b7e9
+		// and on Wayland in SDL < 2.0.20:
+		//     https://github.com/libsdl-org/SDL/commit/3e1b3bc
+		// Fixes issue #7404, which is a regression caused by pull #2481
+		// that fixed issue #2218.
+		{
+			int xmul = 1;
+#if !defined(_WIN32) && !defined(__APPLE__)
+			const char* video_driver = SDL_GetCurrentVideoDriver();
+			SDL_version ver;
+			SDL_GetVersion(&ver);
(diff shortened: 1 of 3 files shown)


Pentarctagon:
Thanks! I'll leave this open for a few days in case there are any comments, otherwise I'll go ahead and merge it.

pehjota:
Sounds good, thanks.  I referenced this in the issue and original PR in the hopes of soliciting any such comments.

pehjota:
It occurred to me that I should add a changelog entry, so I added one to the commit here (no other changes) and opened a new PR to add one for the original PR in master.

src/controller_base.cpp
```diff
@@ -230,11 +230,33 @@ void controller_base::handle_event(const SDL_Event& event)
...
+		//     https://github.com/libsdl-org/SDL/pull/4700
+		//     https://github.com/libsdl-org/SDL/commit/515b7e9
+		// and on Wayland in SDL < 2.0.20:
+		//     https://github.com/libsdl-org/SDL/commit/3e1b3bc
+		// Fixes issue #7404, which is a regression caused by pull #2481
+		// that fixed issue #2218.
+		{
+			int xmul = 1;
+#if !defined(_WIN32) && !defined(__APPLE__)
+			const char* video_driver = SDL_GetCurrentVideoDriver();
```
vgaming, line 247: This this cheap enough to execute on any SDL_MOUSEWHEEL event? Probably "yes", but worth checking. You've probably ran it locally and didn't observe any noticeable slowdown, @pehjota ?

vgaming: UPD: yes, it's cheap enough. I've tested on 1.17 now and I'm not able to notice any slow-down. Thanks!

pehjota: Right, as mentioned in the commit message, I ran this locally with SDL 2.0.14 and 2.28.5.  No noticeable slowdown even on my 2.40-GHz Intel Core 2 Duo (GMA 4500MHD) laptop.

It's only two to four cheap function calls and a handful of assignments and comparisons on each SDL_MouseWheelEvent.  But it can be improved slightly by caching the version/driver checks result in a static variable.  Done, in PRs #8185 and #8186.

Thanks again!

PR #8160, opened 2023-12-27
EI - 04b_Ill_Humors.cfg "time over" bugfix

Dalas121:
04b's "time over" event triggers on turn 22, even after the turn limit gets extended (thanks for the bug report 'Emperor').  This is a quick fix for the issue.


nemaara (approved):

soliton-:
> 04b's "time over" event triggers on turn 22, even after the turn limit gets extended (thanks for the bug report 'Emperor'). This is a quick fix for the issue.

Something like that should be the commit message not a meaningless "Update file".

stevecotton:
Sorry for merging without noticing that. I've added a changelog entry in 356e20f3590bfa4f20bfa3403ec65e73935046e0 to improve the situation a bit.

PR #8162, opened 2023-12-27
Add optional tag name overrides to conditional tags in schema.cfg

rfsc-mori:
Fixes schema.cfg validation errors after #8136.

I've followed the same pattern of the other tags in this schema to allow the conditional tags `case`, `else`, `then`, `elseif` and `else` to be defined without a name key.

### DIFF
--- a/data/schema/schema.cfg
+++ b/data/schema/schema.cfg
@@ -191,7 +191,6 @@
 					[tag]
 						name="elseif"
 						max=infinite
-						super="wml_schema/tag"
 						any_tag=yes
 						{ANY_KEY string}
 						[tag]
--- a/data/schema/schema.cfg
+++ b/data/schema/schema.cfg
@@ -56,6 +56,13 @@
 		name="bool"
 		value="yes|no|true|false"
 	[/type]
+    # Used to override a required key into an optional key with no allowed values.
+    # This should be used for the `name` key in certain tags where it is not allowed,
+    # as the schema can't override a supertag key with its absence.
+    [type]
+        name="disallowed"
+        value="]^"
+    [/type]
 	[tag]
 		name="root"
 		min=1
@@ -78,10 +85,8 @@
 								name="type"
 								max=infinite
 								super="wml_schema/type"
-								# Override the required name with an optional name
-								# Technically name is not really allowed here at all,
-								# but the schema can't override a supertag key with its absence.
-								{SIMPLE_KEY name string}
+                                # The `name` key should not exist in a union tag.
+								{SIMPLE_KEY name disallowed}
 							[/tag]
 						[/tag]
 					[/then]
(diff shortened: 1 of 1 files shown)


CelticMinstrel:
Hmm. Though this seems good, we could perhaps document it a bit better by defining a "blank" type that accepts only an empty string as a valid value. That way, putting `name=foo` in those tags would still correctly be an error.

rfsc-mori:
Done! Thanks! I've decided to use an impossible pattern so even an empty "" string won't be matched. I've also removed the comments, but I can add them back if needed. Or maybe keep them only on the first occurrence?

rfsc-mori:
Hmm, it seems that some of these tags indeed have a name key. I'll look more into it later, sorry.

data/schema/schema.cfg
```diff
@@ -56,6 +56,10 @@
 		name="bool"
 		value="yes|no|true|false"
 	[/type]
+    [type]
```
CelticMinstrel, line 59: If you think it's a good idea, we could put the explanation of "can't override a super key with its absence" here? Just a thought, I don't strongly mind one way or the other.

rfsc-mori:
I tried to work a bit more on this and re-added some comments for documentation.

Also, please correct me if I'm wrong, I'm still learning about this part of the code (and of course in general too):
The `[elseif]` tag being a filter should be more of an `[if]` tag than a `[tag]` itself. I mean, in a sense that it should not specify `super="wml_schema/tag"` as it should not use the `[tag]` keys and tags, as this should go into its `[then]`.

Following the `[if]` pattern without the super, we can solve this by removing both the "disallowed" and the super, and then the intent feels clearer than overriding the name key with an optional value.

I feel this makes sense, but I'm not too sure if there are special cases for the `[elseif]` in this regard.

Also, if this is correct and accepted, do you think it's still a good idea to go ahead with the `[filter]` tag implementation?

data/schema/schema.cfg
rfsc-mori: It is a great idea! Thanks! Added!

CelticMinstrel:
> Also, please correct me if I'm wrong, I'm still learning about this part of the code (and of course in general too): The `[elseif]` tag being a filter should be more of an `[if]` tag than a `[tag]` itself. I mean, in a sense that it should not specify `super="wml_schema/tag"` as it should not use the `[tag]` keys and tags, as this should go into its `[then]`.
> 
> Following the `[if]` pattern without the super, we can solve this by removing both the "disallowed" and the super, and then the intent feels clearer than overriding the name key with an optional value.

Ah, you're right. I'd forgotten that `[elseif]` takes a `[then]` tag. So, it shouldn't have `[tag]` as its super at all.

stevecotton:
Should this be merged for 1.18, regardless of whether #8136 is merged for 1.18?

CelticMinstrel:
It's a pure schema change, so probably.

PR #8163, opened 2023-12-27
Deprecate abilites

gfgtdf:
the deprecated the `u.abilities` properties in formula and lua which returns a list of ability ids, and renames them to "ability_ids", the idea is that by doing this now we can eventually make `u.abilities` return a userdata that gives more information than just the ids.

### DIFF
--- a/src/formula/callable_objects.cpp
+++ b/src/formula/callable_objects.cpp
@@ -250,6 +250,9 @@ variant unit_callable::get_value(const std::string& key) const
 
 		return variant(std::move(res));
 	} else if(key == "abilities") {
+		deprecated_message("unit.abilities", DEP_LEVEL::FOR_REMOVAL, version_info("1.21"), "Use unit.ability_ids instead.");
+		return formula_callable::convert_vector(u_.get_ability_id_list());
+	} else if(key == "ability_ids") {
 		return formula_callable::convert_vector(u_.get_ability_id_list());
 	} else if(key == "hitpoints") {
 		return variant(u_.hitpoints());
@@ -381,6 +384,7 @@ void unit_callable::get_inputs(formula_input_vector& inputs) const
 	add_input(inputs, "objects_count");
 	add_input(inputs, "attacks");
 	add_input(inputs, "abilities");
+	add_input(inputs, "ability_ids");
 	add_input(inputs, "hitpoints");
 	add_input(inputs, "max_hitpoints");
 	add_input(inputs, "experience");
@@ -440,6 +444,9 @@ variant unit_type_callable::get_value(const std::string& key) const
 	} else if(key == "race") {
 		return variant(u_.race_id());
 	} else if(key == "abilities") {
+		deprecated_message("unittype.abilities", DEP_LEVEL::FOR_REMOVAL, version_info("1.21"), "Use unittype.ability_ids instead.");
+		return formula_callable::convert_vector(u_.get_ability_id_list());
+	} else if(key == "ability_ids") {
 		return formula_callable::convert_vector(u_.get_ability_id_list());
 	} else if(key == "traits") {
 		std::vector<variant> res;
@@ -493,6 +500,7 @@ void unit_type_callable::get_inputs(formula_input_vector& inputs) const
 	add_input(inputs, "race");
 	add_input(inputs, "alignment");
 	add_input(inputs, "abilities");
+	add_input(inputs, "ability_ids");
 	add_input(inputs, "traits");
 	add_input(inputs, "attacks");
 	add_input(inputs, "hitpoints");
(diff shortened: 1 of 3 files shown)


CelticMinstrel:
No objections to this as a whole. I'm not sure if it should be in 1.18 though.

gfgtdf:
Imo it makes sense to do such stuff rather sooner than later because  otherwise people who might want to start using this in 1.18 will have to change their code again. 
(Also i think this a pretty rarely used function)

Pentarctagon:
How long has `u.abilities` been available?

gfgtdf:
i don't really know, but i also don't see how that's relevant at all.

Pentarctagon:
If it's been this way for a while, there's probably plenty of people using it already. Is it something that's critical to be able to change in 1.19?

CelticMinstrel:
I think gfgtdf is probably right that there aren't many people using this, but there probably could be _some_ people. Still… this PR only deprecates it, so people using it wouldn't need to instantly update, right?

Pentarctagon:
Is there a reliable way to search for where it's being used? Either way though, while I suppose that deprecating something doesn't technically break the feature freeze, adding a new replacement API certainly does.

CelticMinstrel:
I don't think there's a good way to find uses. You could do a search for "abilities" but it would have lots of false positives.

And there's no point in deprecating without supplying a replacement, so from your words it seems like you could only deprecate something in feature freeze if the replacement already exists. For example, deprecating `on_event` would be fine (and I still think it should happen!), because its replacement already exists: `wesnoth.game_events.add_repeating`.

Pentarctagon:
> so from your words it seems like you could only deprecate something in feature freeze if the replacement already exists.

That's how I view it, yeah. If it's something only used by one or two add-ons then maybe it's a situation where in practice there's barely any impact so it doesn't really hurt anything to just go ahead and do it, but if we aren't able to reliably search for its usage then there's no way for us to know that.

gfgtdf:
> Is there a reliable way to search for where it's being used? Either way though, while I suppose that deprecating something doesn't technically break the feature freeze, adding a new replacement API certainly does.


i've rectly gotten around this, and i still think this is still a good thing to to, are there any opinions on this? if we ever get  to make an acceccor to actual ability objects i don't want to give it a bad name becasue the logical name is already taken by a rareley used and badly named feature.

gfgtdf:
updated

soliton-:
Just noting here that this is missing wmllint rules that are needed according to the new [deprecation policy](https://wiki.wesnoth.org/CompatibilityStandards#How_to_deprecate). Those should ideally be added for 1.20.

gfgtdf:
> Just noting here that this is missing wmllint rules that are needed according to the new [deprecation policy](https://wiki.wesnoth.org/CompatibilityStandards#How_to_deprecate). Those should ideally be added for 1.20.

As none of these changes effect wml keys/attributes i don't think our current wmlint can hande this.

soliton-:
> > Just noting here that this is missing wmllint rules that are needed according to the new [deprecation policy](https://wiki.wesnoth.org/CompatibilityStandards#How_to_deprecate). Those should ideally be added for 1.20.
> 
> As none of these changes effect wml keys/attributes i don't think our current wmlint can hande this.

Sounds like the new deprecation policy is not useful then.

gfgtdf:
> Sounds like the new deprecation policy is not useful then.


well, it says "in all cases where it's practical"  which just doesn't include this case as it would essesially require writing a new lua and formula parset in python that can deduce the types of the objects in question.

soliton-:
> well, it says "in all cases where it's practical" which just doesn't include this case as it would essesially require writing a new lua and formula parset in python that can deduce the types of the objects in question.

Sure and the policy is not useful if the result is not actually adding wmllint rules. Then we're back to UMC having no idea why stuff breaks and no tool that helps them either. Strictly worse than simply leaving the deprecation in.

Pentarctagon:
It should also be mentioned in the changelog as well as updated on https://forums.wesnoth.org/viewtopic.php?t=58532/https://wiki.wesnoth.org/CompatibilityBreakingChanges#Compatibility_breaking_changes_between_1.18_and_1.19.2F1.20

gfgtdf:
> It should also be mentioned in the changelog

will do
> as well as updated on https://forums.wesnoth.org/viewtopic.php?t=58532


It seems like i don't have permissiosn anymore to edit that.

> https://wiki.wesnoth.org/CompatibilityBreakingChanges#Compatibility_breaking_changes_between_1.18_and_1.19.2F1.20

Did that aleady

Pentarctagon:
Post a reply to https://forums.wesnoth.org/viewtopic.php?t=58532 and it will be added to the first post.

PR #8166, opened 2023-12-28
Unit Type Editor

babaissarkar:
New Unit Type Editor UI inside Map Editor. Allow the user to create a template that supports many of the common keys in UnitTypeWML, along with preview of the generated WML.
It can be accessed via new **Units menu > Edit**.

**Screenshots**
Screenshot from 2023-12-28 22-35-54

Screenshot from 2023-12-28 22-36-06

Screenshot from 2023-12-28 22-36-14

Screenshot from 2023-12-28 22-36-24

Screenshot from 2023-12-28 22-38-02

Screenshot from 2023-12-28 22-38-10

Screenshot from 2023-12-28 22-36-33

(Note : The screenshots are slightly old, so one or two texts might be a bit different.)


Pentarctagon:
Postponed as this isn't something that can go in until after 1.18.0 is released.

knyghtmare:
May I see some in-game screenshots of how it looks?

CelticMinstrel:
These are comments based entirely on the screenshots. I may review the code later as well.

* Layout issues. Both halves of the dialog should be aligned to the top of the dialog. I'm not sure if this is a blocker for merging, that depends on @Vultraz I think.
* I think the checkboxes for movement types and the like should be to the left of the slider, not to the right. Anyone else agree?
* I wonder if range should be a dropdown like type is? Since an add-on could define custom ranges… though I think it's uncommon.
* Should the WML view have a button to copy it to the clipboard? Maybe it's not that important since you're probably going to save it to a file, but…
* I find it weird that level is a text block when all the other numeric fields are sliders… though I also feel that the ideal state for all of them would be a spinbox. I don't think GUI2 has a spinbox as such, but it might have the tools needed to create one as a `[textbox_definition]`. There probably isn't anything for data validation though, so that would have to be done in the C++…?

For reference, this is what I mean by a spinbox:

babaissarkar:
>     I think the checkboxes for movement types and the like should be to the left of the slider, not to the right. Anyone else agree?
I'm happy both ways. I initially wanted to put them inside the menubutton, but looks like the menu button doesn't support getting individual state of the checkboxes. (?)

>     I wonder if range should be a dropdown like type is? Since an add-on could define custom ranges… though I think it's uncommon.
Depends on how complex defining a range is. The gui is supposed to not support complicated WML, which should be done manually. Please consider linking me to some documentation on how it's actually done.

>   Should the WML view have a button to copy it to the clipboard? Maybe it's not that important since you're probably going to save it to a file, but…
It's getting saved to a file, but I'll see if I can add it.

>   I find it weird that level is a text block when all the other numeric fields are sliders… though I also feel that the ideal state for all of them would be a spinbox. I don't think GUI2 has a spinbox as such, but it might have the tools needed to create one as a `[textbox_definition]`. There probably isn't anything for data validation though, so that would have to be done in the C++…?

Yes, somebody else on the chat also suggested this. Something like a textbox with left/right arrow buttons? the up/down arrows look too small. The logic needs to be done in C++, so probably we need a new Spinner widget. I'll adjust the level/damage/range.

CelticMinstrel:
> Depends on how complex defining a range is. The gui is supposed to not support complicated WML, which should be done manually. Please consider linking me to some documentation on how it's actually done.

The only documentation on defining a custom range or type is [here](https://wiki.wesnoth.org/LanguageWML).

> Yes, somebody else on the chat also suggested this.

Wasn't the someone else in chat also me? Or was there another person who mentioned it?

Left/right arrow buttons would be fine too, I guess. There'd also be the possibility of making diagonal arrow buttons kind of like this (please forgive the terrible ASCII art):

```
+-----+
|\    |
| \ ◥ |
|  \  |
| ◣ \ |
|    \|
+-----+
```

I think the buttons would be the same `[repeating_button]` used by scrollbars.

I wouldn't say making a spinbox is required to accept this pull request. It's more of a "nice to have" thing. If you're not making a spinbox I think the level should probably be a slider (maybe a max of 10?).

gfgtdf:
The `load_unit_type()` function its implemented /used so am i right to assume that this is not finished yet?


On the ui:
Imo next/previous pagers like the one in the attack section are a bit hard to use/require too many clicks, it woudl be better if this was either a scrollable list with all the attacks, or maybe make the attacks appear as single items in the left side list (for example by making a left side list a treeview and the single attacks subnodes of the attacks node)

babaissarkar:
> The `load_unit_type()` function its implemented /used so am i right to assume that this is not finished yet?

Yes, planned but not done yet. I was going to create it earlier, but ran into a confusion regarding whether to load it directly from a cfg file using the file chooser dialog, or load it from the unit types database inside wesnoth using some kind of unit selector dialog (something like the create unit dialog). In the second case another dialog needs to be created.

> On the ui: Imo next/previous pagers like the one in the attack section are a bit hard to use/require too many clicks, it woudl be better if this was either a scrollable list with all the attacks, or maybe make the attacks appear as single items in the left side list (for example by making a left side list a treeview and the single attacks subnodes of the attacks node)

Will a `[menubutton]` listing the attacks work? That way the user can just select the attack from it to go there in at least two clicks.

babaissarkar:
Preliminary design of Spinner widget :
Screenshot from 2023-12-30 11-58-12

Please comment/react on this whether you need/like it or not. It allows both manual and button controls over slider : you can directly enter a value or press the buttons to increase/decrease. Pressing and holding the buttons will increase the speed of the change (if that's possible for me to implement). Your feedback helps me to know if I can continue working on this or try some alternative solution.

babaissarkar:
Limitations when loading unit data :
1. Macros are not loaded in the UI, so the user still has to select the ability/weapon special from the list provided.
2. When loading unit, movement costs/defense/resistance are loaded, but the movetype ID cannot be loaded because it isn't recorded inside the game.

CelticMinstrel:
> Macros are not loaded in the UI, so the user still has to select the ability/weapon special from the list provided.

This is a good reason to refactor the ability specification system like I've pondered for awhile now. Maybe it's help motivate me to actually get started…



> When loading unit, movement costs/defense/resistance are loaded, but the movetype ID cannot be loaded because it isn't recorded inside the game.

It's not stored in the movetype itself, but it's usually available in the unit type config under the `movetype` key, isn't it? Or am I missing something?

babaissarkar:
> It's not stored in the movetype itself, but it's usually available in the unit type config under the `movetype` key, isn't it? Or am I missing something?

This doesn't looks like it holds the string : https://github.com/wesnoth/wesnoth/blob/7c4ac03be2dec337dbdc16b69c6efeaecc99be9e/src/units/types.hpp#L191, but perhaps this has : https://github.com/wesnoth/wesnoth/blob/7c4ac03be2dec337dbdc16b69c6efeaecc99be9e/src/units/types.hpp#L425, but it is private.

CelticMinstrel:
I looked again, and it seems to be true that even the unit type object doesn't track the movetype ID. It's definitely in the unit type file though. I misremembered the key – it's `movement_type=<ID of movetype>`. You wrote that information when saving the unit, right? So you should be able to read it when loading it, I would think? Even if you have to grab it straight from the config _before_ it's converted into a unit type…

I'd say the most important thing is that loading and resaving a unit doesn't result in anything significant changing (eg, different whitespace is fine).

babaissarkar:
> I looked again, and it seems to be true that even the unit type object doesn't track the movetype ID. It's definitely in the unit type file though. I misremembered the key – it's `movement_type=<ID of movetype>`. You wrote that information when saving the unit, right? So you should be able to read it when loading it, I would think? Even if you have to grab it straight from the config _before_ it's converted into a unit type…

I'm using wesnoth's unit create dialog (aka wesnoth's unit database) to load a movetype, and not loading it from the cfg file. Guess I can add an option to load a unit type directly from a cfg file.

CelticMinstrel:
Ah, right. Well, in that case, adding a getter to the unit type object just to store that information seems reasonable. Something like `get_base_movement_type()`.

babaissarkar:
> Ah, right. Well, in that case, adding a property and getter to the unit type object just to store that information seems reasonable. Something like `get_base_movement_type()` or `get_movement_type_base()`. I could imagine someone in Lua may like to know this in the future, too.

Yes, added a getter to the unit type. Will be in the next commit so you can look at it. I've also added a getter for the `movement_type_map`, since `unit_types` already has getters for races/traits.

CelticMinstrel:
Yeah, that's more or less fine. You can also compare the movetype from the unit type to the move type from the movement type map to decide on whether the "active" checkboxes are set for each of the terrain types.

babaissarkar:
Rebased this PR on top of PR #8199.

babaissarkar:
`[resistance]/[movement_costs]/[defense]` tags inside `[unit_type]` work now.

babaissarkar:
Somebody please remove the unnecessary labels please, came from the merge.

CelticMinstrel:
Would be nice to rebase this to get rid of the irrelevant commits in the diff.

babaissarkar:
@CelticMinstrel there should be just 1 commit now.

Pentarctagon:
I think this can be merged in a couple days, barring anything else coming up with 1.18.0 that needs to be focused on. There haven't been any major issues so far, so no point in waiting a long time to start merging things into master.

Wedge009:
I completely missed this addition, looks nice. I just have a question about image paths being marked for translation: is this intentional and if so, what is the reason for it?

https://github.com/wesnoth/wesnoth/blob/f7fea77603dc091396f640f96c70354c243f3147/data/gui/window/editor_edit_unit.cfg#L1245

https://github.com/wesnoth/wesnoth/blob/f7fea77603dc091396f640f96c70354c243f3147/data/gui/window/editor_edit_unit.cfg#L1257

https://github.com/wesnoth/wesnoth/blob/f7fea77603dc091396f640f96c70354c243f3147/data/gui/window/editor_edit_unit.cfg#L1269

https://github.com/wesnoth/wesnoth/blob/f7fea77603dc091396f640f96c70354c243f3147/data/gui/window/editor_edit_unit.cfg#L1281

CelticMinstrel:
I think it's incorrect. Was it done to satisfy the schema check, perhaps? That's still wrong though.

Wedge009:
#8729

babaissarkar:
@Wedge009 They were a typo, thanks for solving it.

Wedge009:
Thanks for confirming.

PR #8168, opened 2023-12-28
systemd: Run as _wesnoth:_wesnoth using sysusers.d

pehjota:
[systemd 246+ logs a warning message](https://github.com/systemd/systemd/blob/v246/NEWS#L106-L113), because running as "nobody" is unsafe.

This should be reviewed and tested by someone who knows and uses systemd, as I do not.  Is this worth escalating in the changelog to `### Security Fixes`?

See also #8169 for 1.16.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -786,7 +786,8 @@ if not access(fifodir, F_OK):
     env.Alias("install-wesnothd", fifodir)
 if env["systemd"]:
     env.InstallData("prefix", "wesnothd", "#packaging/systemd/wesnothd.service", "lib/systemd/system")
-    env.InstallData("prefix", "wesnothd", "#packaging/systemd/wesnothd.conf", "lib/tmpfiles.d")
+    env.InstallData("prefix", "wesnothd", "#packaging/systemd/wesnothd/tmpfiles.conf", "lib/tmpfiles.d")
+    env.InstallData("prefix", "wesnothd", "#packaging/systemd/wesnothd/sysusers.conf", "lib/sysusers.d")
 
 # Wesnoth campaign server
 env.InstallBinary(campaignd)
--- a/changelog_entries/systemd_nobody.md
+++ b/changelog_entries/systemd_nobody.md
@@ -1,0 +1,2 @@
+ ### Security Fixes
+   * Run wesnothd server as `_wesnoth:_wesnoth` instead of `nobody:users`, improving safety and fixing a warning message in systemd 246+
--- a/packaging/systemd/SConscript
+++ b/packaging/systemd/SConscript
@@ -1,4 +1,5 @@
 Import("env")
 
 env.ScanReplace("wesnothd.service", "wesnothd.service.scons.in")
-env.ScanReplace("wesnothd.conf", "wesnothd.tmpfiles.conf.in")
+env.ScanReplace("wesnothd.tmpfiles.conf", "wesnothd.tmpfiles.conf.in")
+env.ScanReplace("wesnothd.sysusers.conf", "wesnothd.sysusers.conf.in")
--- a/packaging/systemd/wesnothd.service.in
+++ b/packaging/systemd/wesnothd.service.in
@@ -23,8 +23,8 @@ ExecStopPost=/bin/rm -f @FIFO_DIR@/socket
 
 SyslogIdentifier=Wesnothd@BINARY_SUFFIX@
 WorkingDirectory=@FIFO_DIR@
-User=nobody
-Group=users
+User=_wesnoth
+Group=_wesnoth
 
 # Additional security-related features
 # (when using the -c option, do not use ProtectHome)
--- a/packaging/systemd/wesnothd.service.scons.in
+++ b/packaging/systemd/wesnothd.service.scons.in
@@ -4,6 +4,8 @@ After=network.target
 
 [Service]
 ExecStart=%bindir/wesnothd
+User=_wesnoth
+Group=_wesnoth
 
 [Install]
 WantedBy=multi-user.target
--- a/packaging/systemd/wesnothd.sysusers.conf.in
+++ b/packaging/systemd/wesnothd.sysusers.conf.in
@@ -1,0 +1,2 @@
+u _wesnoth -
+g _wesnoth -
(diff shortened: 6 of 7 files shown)


Pentarctagon:
Is `games:games` always going to exist?

Pentarctagon:
@sevu any thoughts on this?

sevu:
Archlinux has the group, but not the user.

Pentarctagon:
What happens if the group and/or user doesn't exist?

pehjota:
[So I see](https://gitlab.archlinux.org/archlinux/packaging/packages/filesystem/-/blob/main/sysusers).  Without digging through a bunch of unfamiliar code, I don't know what systemd does if the user doesn't exist, but I suppose it either fails to start the service or (hopefully/probably not) runs it as root (or maybe some default user like "daemon").

Worth nothing though is that Arch doesn't have a "nobody" user, so Wesnoth trying to run wesnothd as a non-existent user apparently hasn't been a problem there so far.

It's odd that Arch doesn't have a "games" user, so I wonder what user other game server packages in Arch use.  I see their minetest package [creates a "minetest" user](https://gitlab.archlinux.org/archlinux/packaging/packages/minetest/-/blob/main/sysusers.d), so maybe Arch would create a "wesnoth" user, if they even care at all (which like I said they don't seem to so far).  In which case, Wesnoth defaulting to "games" should be good enough for everyone else?

stevecotton:
Thanks for proposing this in Debian too, because it got a response from one of Debian's OpenArena maintainers: https://lists.debian.org/debian-devel-games/2023/12/msg00042.html

Well worth reading the full text, but the quick summary is:
> OpenArena uses the Debian-openarena user ID, which is part of an older naming convention - if I was packaging OpenArena today, it would be running as `_openarena` instead.

CelticMinstrel:
For what it's worth, MacOS doesn't have a "games" user.

pehjota:
Thanks, you quoted that here sooner than I got around to it.  So yes, his opinion is that wesnothd should have its own dedicated user and group, instead of reusing a system user/group, which would indeed be best.  I don't really think it's appropriate for a project's build system to go creating new users and groups on people's systems... so I'll have to try to find some examples of what users/groups other projects use in their provided systemd (or other) service files.

So I think this maybe shouldn't be merged yet, if at all, until figuring out a better alternative to the current nobody:users (another system user/group which shouldn't be reused for this).  Maybe the current systemd service file should be treated as just an example and not installed by the build system?

sevu:
> 

There’s a way to create users with systemd, I have some related changes and could make a PR next year.

> Maybe the current systemd service file should be treated as just an example and not installed by the build system?

That’s a good question

pehjota:
Simon and Alexandre of Debian [pointed](https://lists.debian.org/debian-devel-games/2023/12/msg00046.html) me to [sysusers.d](https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html) and the underscore convention.  I've updated this PR to use that to allocate `_wesnoth:_wesnoth`.

I thought about making the names configurable, but I'm not sure how to do that consistently between CMake and SCons.  For example, [packaging/systemd/wesnothd.tmpfiles.conf.in](https://github.com/wesnoth/wesnoth/blob/master/packaging/systemd/wesnothd.tmpfiles.conf.in) references `@FIFO_DIR@` and [CMakeLists.txt](https://github.com/wesnoth/wesnoth/blob/master/CMakeLists.txt) sets it, but [SConstruct](https://github.com/wesnoth/wesnoth/blob/master/SConstruct) and [packaging/systemd/SConscript](https://github.com/wesnoth/wesnoth/blob/master/packaging/systemd/SConscript) don't.

As before, I don't use systemd, so review and testing by someone who does would be appreciated.  Or maybe @sevu's changes do this better?

soliton-:
Using the underscore convention sounds good. Unless it is really trivial I don't think creating users is in scope.

pehjota:
> Unless it is really trivial I don't think creating users is in scope.

I'm not sure what this means.  The change in this PR (creating and running as user and group `_wesnoth` is out of scope?  Or making the name configurable would be out of scope?  If the latter, fair enough, that's easier anyway.  If the former, then the current behavior of running as `nobody:users` is flawed because:
1. It may cause an [information leak](https://github.com/systemd/systemd/blob/v246/NEWS#L106-L113) on systems with NFS, and
2. Arch [doesn't have](https://gitlab.archlinux.org/archlinux/packaging/packages/filesystem/-/blob/main/sysusers) a `nobody` user.

This change doesn't automatically create a `_wesnoth` user and group but rather just gives system administrators and distributors the option to do so with the systemd-sysusers command (if I'm correctly understanding how that utility works).

soliton-:
Sorry, didn't see that the sysusers stuff was already part of this PR. Looks good to me.

Pentarctagon:
Does this need a CVE created? I'm not familiar with when those are needed.

pehjota:
I don't know either, but I'm not sure this rises to that level.  It would take a vulnerability in wesnothd that allows arbitrary remote access to files owned by the same user/group as wesnothd for there to be a spillover into a more serious information leak involving NFS.  This change doesn't fix any actual known information leaks (as far as I know), but rather makes such a leak via shared users/groups impossible (for those who use systemd-sysusers).  Basically, it's a hardening improvement to sandbox the server a little better.

soliton-:
@sevu any news?

IMO this should be merged. I'll do so in a couple days if there are no further comments.

stevecotton:
@sevu the `_` prefix on the usernames is Debian policy, we learned about it in https://lists.debian.org/debian-devel-games/2023/12/msg00042.html .

The text you've quoted in the commit message of d6d2a4f9967be5cce600d581eabf252138ef79ac also explicitly recommends the underscore, so if there had been a PR for that change then I'd have been against it. I intend to revert that commit, unless there's a better reason to avoid having the underscore that your feeling that it looks better without.

pehjota:
> @sevu the `_` prefix on the usernames is Debian policy

It's Debian policy because more broadly/importantly it's a systemd strong recommendation (as you've both noted).

> commit message of d6d2a4f9967be5cce600d581eabf252138ef79ac

Also commit 8b6ef6aeeb80273a5826e2a25fffe5a2ca26a529 for 1.16.

pehjota:
The reason for the underscore, even if "wesnoth is unique enough to not clash" with people's names, is in case a fan of the game picked "wesnoth" as their user name when installing their system.  Underscore is a good namespace for system users/groups, leaving everything else for users.

PR #8169, opened 2023-12-28
[1.16] systemd: Run as _wesnoth:_wesnoth using sysusers.d

pehjota:
[systemd 246+ logs a warning message](https://github.com/systemd/systemd/blob/v246/NEWS#L106-L113), because running as "nobody" is unsafe.

This should be reviewed and tested by someone who knows and uses systemd, as I do not.  Is this worth escalating in the changelog to `### Security Fixes`?

See also #8168 for master.


pehjota:
Updated to match #8168.  Bumped changelog entry to "Security Fixes", since I see the Security label was added.  If you disagree, merge commit d8ea64d instead.

PR #8171, opened 2023-12-28
Fix font permissions to allow installable embedding

pehjota:
The TrueType "OS/2" table "fsType" (embedding permissions) field was set to 0x8 in the Droid Sans and Oldania ADF Std fonts, allowing only editable embedding.  This change marks the fonts as allowing permanent installation on a system, as is typically done with Wesnoth and allowed by the fonts' licenses.

I set the field using Tom Murphy 7's public domain and portable C "embed" program, now included at utils/font-embedding-fix.c (converted with dos2unix and fixed to `#include <string.h>`).  I verified with VBinDiff that only the "OS/2" table "checksum" and "fsType" fields were changed, so no other changes have been made to the fonts.  I confirmed using FontForge and the new utils/font-embedding-check.pl script that the fonts now allow installable embedding.

References:
http://carnage-melon.tom7.org/embed/
https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6.html https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6OS2.html https://learn.microsoft.com/en-us/typography/opentype/otspec140/os2#fst

See also #8172 for 1.16.


Pentarctagon:
Would this have any sort of impact in practice, or is this just to make them consistent with the other fonts?

pehjota:
It's a form of DRM that non-free OSes' (e.g. Windows and OS X) font installers might obey and refuse to install the fonts.  At least they're supposed to, according to Microsoft's OpenType spec, but I can't confirm.  So if a user of such an OS wanted to install the fonts system-wide for use with other applications (e.g. they like a font they find in Wesnoth and want to write documents with it), they might not be allowed to.

But within the context of Wesnoth, it has no particular impact, other than clarifying that Wesnoth is legally installing them to users systems.  It would, however, fix the [lintian](https://salsa.debian.org/lintian/lintian/-/blob/master/lib/Lintian/Check/Fonts/Truetype.pm) warning [tag](https://salsa.debian.org/lintian/lintian/-/blob/master/tags/t/truetype-font-prohibits-installable-embedding.tag) in the Debian package.

stevecotton:
That Debian Lintian tag is also triggering for [Debian's own packaged version](https://udd.debian.org/lintian/?packages=fonts-android), and on Debian Stable I confirm that `ttfdump -t OS/2 /usr/share/fonts/truetype/droid/DroidSansFallbackFull.ttf` reports that this flag is set in the Debian file too.

Is there any discussion with the Debian Fonts Task Force about changing it? I only checked the past 2 months on their mailing list and didn't spot it, however I feel that understanding what this flag means is outside the core scope of Wesnoth.

pehjota:
Indeed, it's a fairly recent lintian tag, added only [four years ago](https://lists.debian.org/debian-fonts/2019/12/msg00046.html) (about two weeks ago in Debian time), so packages are still catching up to it.  The most recent version of Debian's fonts-android package was released [almost three years *before*](https://tracker.debian.org/news/839009/accepted-fonts-android-1601r16-11-source-into-unstable/) the lintian tag was added.  I will file a bug report against the fonts-android package to bring the issue to the maintainers' attention and explain why and how to fix it.

But it was actually the Debian Fonts Task Force who [initially discussed](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=635068) checking the field in 2011 and noted the value of doing so, referring for example to the Ubuntu Font Family development: ["Not making this change may restrict the use of the fonts on certain software or systems which check the value of this field (eg. apparently MSIE9 with .ttf webfonts)"](https://bugs.launchpad.net/ubuntu-font-family/+bug/648406), ["Leaving the fstype DRM bits set we are depriving users of the very Freedoms that we are trying to offer users,"](https://bugs.launchpad.net/ubuntu-font-family/+bug/648406/comments/6) and ["Having a DRM'd Ubuntu font would not go down well with the press, no."](https://bugs.launchpad.net/ubuntu-font-family/+bug/648406/comments/6)  A somewhat long Debian Fonts Task Force mailing list thread in 2021 debated how to solve the problem.  [More examples of the DRM being enforced](https://lists.debian.org/debian-fonts/2021/10/msg00017.html) were found, and apparently the problem is tools like FontForge defaulting to enabling the DRM (for the benefit of proprietary font designers), so often designers who intend their fonts to be freely usable are unaware of the restrictions they're imposing.

In general, I imagine checking for this DRM field might help a distribution like Debian detect packages containing fonts that might actually be non-free/proprietary.  And it seems to me free software projects like Wesnoth ought to have DRM-free fonts that users can reuse wherever they like. :)  It could be a little off-putting if a user was forbidden from using a font they found in a free software open source project, of all places.  And again, this DRM is designed to prohibit the very thing that Wesnoth does: install fonts on a system; so it would be nice to clarify that such installation is being done legally.  Personally I don't think distributing DRM-free fonts is outside the scope of Wesnoth, if the goal is to provide an open-source game with assets that are free to be reused outside the game.

Regarding the technical details of this PR though, I wonder if it would be better to build and run the "embed" program in a CI script, to help automatically detect possibly proprietary or simply mistakenly DRM-encumbered fonts being added.  I'd have to spend some time playing around with that to figure it out.

soliton-:
While wesnoth includes fonts they were not made by wesnoth developers. It does seem more useful to discuss/address any issues with the font developers than with every project that includes them.

Pentarctagon:
I don't think there's really any need to have validation scripts/programs for this (especially in perl rather than python), but there shouldn't be any harm in updating the attributes of our existing fonts either?

pehjota:
I apologize for suggesting that Wesnoth be a "vanguard" on this.  Although I still believe Wesnoth should not be distributing DRM-encumbered files (regardless of whether Wesnoth embeds the fonts into PDF documents), I do still want to address this upstream as well.

To that end, I'm still putting some finishing touches on a message to the Debian Fonts Task Force, which I hope to send this weekend.

But to @Pentarctagon's point, yes, sorry for the Perl, but the SCons code in the system Lua PR is really the most Python I've ever written in my life.  Is a C program like utils/font-embedding-fix.c in this PR acceptable?  I learned after submitting this PR that this particular program actually sets the OS/2 table checksum incorrectly (so this PR is not ready to merge), which is fixed in a fork of the program.  I can push a new commit with the fixed fonts and/or a fixed program, which can also replace my Perl script.

Pentarctagon:
I'd really rather just update the current fonts and leave it at that, unless anyone else feels strongly otherwise.

stevecotton:
Don't merge yet. Seeing the discussion that's currently on the Debian mailing lists, we should wait for @pehjota to report back first.

pehjota:
Indeed, I said Saturday it's not ready to merge, and I marked it as a draft.

Pentarctagon:
Any update on this?

Pentarctagon:
Closing due to lack of response - feel free to reopen this later though.

PR #8172, opened 2023-12-28
[1.16] Fix font permissions to allow installable embedding

pehjota:
The TrueType "OS/2" table "fsType" (embedding permissions) field was set to 0x8 in the Droid Sans and Oldania ADF Std fonts, allowing only editable embedding.  This change marks the fonts as allowing permanent installation on a system, as is typically done with Wesnoth and allowed by the fonts' licenses.

I set the field using Tom Murphy 7's public domain and portable C "embed" program, now included at utils/font-embedding-fix.c (converted with dos2unix and fixed to `#include <string.h>`).  I verified with VBinDiff that only the "OS/2" table "checksum" and "fsType" fields were changed, so no other changes have been made to the fonts.  I confirmed using FontForge and the new utils/font-embedding-check.pl script that the fonts now allow installable embedding.

References:
http://carnage-melon.tom7.org/embed/
https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6.html https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6OS2.html https://learn.microsoft.com/en-us/typography/opentype/otspec140/os2#fst

See also #8171 for master.


Wedge009:
Noting that 1.18 has been branched since this was drafted.

Wedge009:
Perhaps this should be re-based against 1.18 branch? (I recognise #8171 is still in progress.)

Pentarctagon:
Closing since 1.16 isn't getting any further releases.

Issue #8176, opened 2023-12-30
Allow experience key in [harm_unit] tag to be set to kill, hit, both / yes or no


Hejnewar:
### Describe the desired feature

As in the title reasons for this are:
1. With aoe attacks getting xp for each harm unit trigger can end up in granting something like 42 xp per attack (depending on aoe, level of enemies etc.)
2. Attacks that use additional [harm_unit] procs per strike end up granting way more combat experience than inteneded
3. This allows for much more flexibility when designing units without standard attacks, that rely fully on non standard way of attacking thru [harm_unit] tag

ProditorMagnus:
https://github.com/inferno8/wesnoth-Era_of_Magic/issues/30 

> When attacker damages third party:
> 
> damager should NOT get fight xp
> damager should get kill xp
> third party should get fight xp
> When defender damages third party:
> 
> damager should NOT get fight xp
> damager should get kill xp
> third party should get fight xp

This use case is not covered by proposed values.

Hejnewar:
>This use case is not covered by proposed values.

Yes but it covers my use cases. Nor do I have any idea how to call them, it in best case i see defender / attacker / both kill, defender / attacker / both hit, defender / attacker / both both / yes or no, which is a lot.

ProditorMagnus:
experience= with list containing any number of `primary_kill,primary_hit,secondary_hit,secondary_kill`
With `hit` meaning `primary_hit,secondary_hit` and `kill` meaning `primary_kill,secondary_kill`
With `yes` and `no` meaning either all or none of them

Hejnewar:
If secondary kill is not possible then there is no need for primary kill to exist, you are not excluding anything, it also doesnt seem to be needed in the EoMa case.

PR #8178, opened 2023-12-30
Allow experience key in [harm_unit] tag to be set to kill, primary / secondary hit, hit or yes

### LINKED ISSUE #7776: UtBS: Attacks killing multiple enemies only awards XP for the first kill

- **Version:** 1.17.18+dev (at ae76ddff686)
- **Downloaded from:** GitHub
- **Build info:** SCons
- **OS:** Linux but presumably all

### Description of the bug

General case for, and related to, #7771.

#7349 introduced attacks targeting multiple enemies in 1.17.13 for the Quenoth Sun Singer/Sylph units. They may have worked at some point between 1.17.13 and 1.17.18, but as of 1.17.18+dev XP for enemies killed doesn't appear to be working correctly. The Sun Singer has a 'ray' attack that can hit a unit behind a target, while the Sun Sylph also has a 'nova' attack that can hit units adjacent to a target. An attack that kills more than one unit is supposed to award XP for all units killed but currently only XP from the first unit killed is gained.

The whole 'first/second/third verse' mechanic isn't explained at all. Eventually I worked out from the unit definitions that it seems to be related to how many attacks the Sun Singer/Sylph has made before it is allowed access to the subsequent 'verse' attacks, they are not available immediately from the start. So to simplify things I've prepared an XP test scenario: 
[xp-test.gz](https://github.com/wesnoth/wesnoth/files/11988805/xp-test.gz) (I'm guessing GUI Debug Tools may need to be installed.)

The Sun Singer at 13,24 is already primed for her 'ray' attack. Use it against the mudcrawlers - compare killing one at a time and killing both together. Observe her XP before and after the attacks and observe that killing two mudcrawlers together only counts the XP from the first mudcrawler killed.

### Expected behavior

In this particular set-up, the Sun Singer's XP should increase by 12: 8 from the Giant Mudcrawler, 4 from the Small Mudcrawler.

### Additional context

From the log:
```
ERROR: TYPE ERROR: expected string but found null (null())
  unit_at
  direction_from

formula type error: TYPE ERROR: expected string but found null (null())
```

Looks like something might have broken?

Hejnewar:
Discussed in issue #8176 

Needed for 7776  

Question:
There is no shema for this right? I tried to find it but didn't.


ProditorMagnus:
Trying to find better names for primary_hit/secondary_hit.

data/lua/wml/harm_unit.lua
```diff
@@ -160,6 +160,10 @@ function wml_actions.harm_unit(cfg)
 
 			wesnoth.interface.float_label( unit_to_harm.x, unit_to_harm.y, string.format( "<span foreground='red'>%s</span>", text ) )
 
+			local award_kill_xp = experience == "kill" or experience == true
+			local award_primary_hit_xp = experience == "hit" or experience == "primary_hit" or experience == true
+			local award_secondary_hit_xp = experience == "hit" or experience == "secondary_hit" or experience == true
```
ProditorMagnus, line 165: Does not seem to support experience=kill,secondary_hit

Hejnewar: Yeah I have no idea how to do that yet, gotta wait until I learn I guess.

gfgtdf:
this seems to change the default behavior (when experience= is not set) to not granting experience.

Hejnewar:
Isnt it yes by default?

CelticMinstrel:
If you want to support lists I suppose it would look something like this.

```lua
local xp_mode = {}
for _, opt in ipairs(stringx.split(experience)) do
  if opt == 'true' or opt == 'yes' then
    xp_mode.kill = xp_mode.primary_hit = xp_mode.secondary_hit = true
  elseif opt == 'hit' then
    xp_mode.primary_hit = xp_mode.secondary_hit = true
  elseif opt == 'primary_hit' then
    xp_mode.primary_hit = true
  elseif opt == 'secondary_hit' then
    xp_mode.secondary_hit = true
  elseif opt == 'kill' then
    xp_mode.kill = true
  elseif opt == 'no' or opt == 'false' then
    -- do nothing, just accept it as valid
  else
    -- put an error message here
  end
end
```

It means parsing the booleans yourself though.

stevecotton:
> Trying to find better names for primary_hit/secondary_hit.

Please would someone write a draft of the documentation that would go on the Wiki for this? Writing documentation is my primary way of finding better names for things.

It seems a reasonable change. If the PR included docs and unit tests, I'd support making an exception to the 1.18 feature-freeze.

Hejnewar:
I can try writing documentation later (perhaps after I figure out how to solve the last cases) but I have no idea how to make unit test.

Hejnewar:
```
			local xp_mode = {}
			for _, opt in ipairs(stringx.split(experience)) do
				if opt == 'true' or opt == 'yes' then
					xp_mode.kill = true
					xp_mode.primary_hit = true
					xp_mode.secondary_hit = true
				elseif opt == 'hit' then
					xp_mode.primary_hit = true
					xp_mode.secondary_hit = true
				elseif opt == 'primary_hit' then
					xp_mode.primary_hit = true
				elseif opt == 'secondary_hit' then
					xp_mode.secondary_hit = true
				elseif opt == 'kill' then
					xp_mode.kill = true
				elseif opt == 'no' or opt == 'false' then
					xp_mode.kill = false
					xp_mode.primary_hit = false
					xp_mode.secondary_hit = false
				else
					wml.error('Invalid [harm_unit] key experience - should be yes, kill, hit, primary_hit, secondary_hit or no')
				end
			end
```

I got this to this state, now it works but doesnt accept yes / no and doesnt work with coma separated list (or any list) right now. Maybe it would be better to make 2 more keys for the previous solution but I would need to think if I can do that anyway.

ProditorMagnus:
Need something like `if type(experience) ~= "string" then experience = tostring(experience) end` before stringx.split.

CelticMinstrel:
Judging from what you wrote there, you should probably consider following this suggestion:

> If you want no/false to only be accepted alone then the no/false branch could be removed and enclose the loop in an `if experience ~= false` check.

Hejnewar:
```
			local xp_mode = {}
			for _,opt in ipairs(tostring(experience):split()) do
				print(opt)
				if opt == 'true' or opt == 'yes' then
					xp_mode.kill = true
					xp_mode.primary_hit = true
					xp_mode.secondary_hit = true
				elseif opt == 'hit' then
					xp_mode.primary_hit = true
					xp_mode.secondary_hit = true
				elseif opt == 'primary_hit' or opt == 'kill_primary_hit' then
					xp_mode.primary_hit = true
				elseif opt == 'secondary_hit' or opt == 'kill_secondary_hit' then
					xp_mode.secondary_hit = true
				elseif opt == 'kill' or opt == 'kill_primary_hit' or opt == 'kill_secondary_hit' then
					xp_mode.kill = true
				elseif opt == 'no' or opt == 'false' then
					xp_mode.kill = false
					xp_mode.primary_hit = false
					xp_mode.secondary_hit = false
				else
					wml.error('Invalid [harm_unit] key, experience should equal yes, kill, hit, primary_hit, secondary_hit, kill_primary_hit, kill_secondary_hit or no')
				end
			end
```

Tbh you will probably not be very happy with this but it works, solves cases and is good enough for me, just gotta document it.

ProditorMagnus:
```
local xp_mode = {kill=false, attack=false, defend=false}
local experience_split = tostring(experience):split()
for _,opt in ipairs(experience_split) do
	if opt == 'true' or opt == 'yes' or opt == 'nil' then
		xp_mode.kill = true
		xp_mode.attack = true
		xp_mode.defend = true
	elseif opt == 'fight' then
		xp_mode.attack = true
		xp_mode.defend = true
	elseif opt == 'attack' then
		xp_mode.attack = true
	elseif opt == 'defend' then
		xp_mode.defend = true
	elseif opt == 'kill' then
		xp_mode.kill = true
	elseif opt == 'no' or opt == 'false' then
		-- preserve previous value
	else
		wml.error('Invalid [harm_unit] experience, list of values can contain: yes, kill, fight, attack, defend, no')
	end
end
if #experience_split == 0 then
	xp_mode = {kill=true, attack=true, defend=true}
end
```
Proposing this version. Notable changes: everything renamed. Removal of the strange combo-values. Better safety for not provided value, nil. Additive treatment of values: `fight,no,kill` = yes for all xp. When value is not provided at all, then default to previous yes for all.

Hejnewar:
I tried `experience = kill,attack` and `experience = kill,defend` and while it awards xp for killing it doesnt for attacing or defending, I will test it more later.

ProditorMagnus:
Renaming of primary_hit kind would need to be done in the code that reads from xp_mode.

Hejnewar:
Oh yeah I forgot sorry. I will test it later again then.

Hejnewar:
Wiki:
experience: (default yes) this can be set to yes, kill, fight, attack, defend, no. If there is a second unit, experience will be attributed like in regular combat while this is set to yes. Kill generates experience only when a unit is killed. Attack generates combat experience only for the unit attacking with defend only for the defending unit or units. No does not generate any additional experience and when used alone does not grant any experience. Keys can be combined in a comma separated list, using this will combine their effects.

ProditorMagnus:
Current PR description is set to close https://github.com/wesnoth/wesnoth/issues/7776.

ProditorMagnus:
Wiki:
experience: (default yes) If there is a second unit, experience will be attributed similar to regular combat. Can take list of values. When any value is provided, all unspecified experience types default to not giving experience. Supported values:
- no - none of involved units get any experience
- kill - the harming unit receives experience based on level of harmed unit in attack which kills harmed unit
- attack - the harming unit receives experience based on level of harmed unit in attack which does not kill harmed unit
- defend - the unit being harmed receives experience based on level of harming unit in attack which does not kill harmed unit
- fight - short form of experience=attack,defend
- yes - short form of experience=kill,attack,defend

data/lua/wml/harm_unit.lua
```diff
@@ -168,11 +194,15 @@ function wml_actions.harm_unit(cfg)
...
 				and wesnoth.sides.is_enemy( unit_to_harm.side, harmer.side )
 			then
-				if kill ~= false and unit_to_harm.hitpoints <= 0 then
+				if kill ~= false and unit_to_harm.hitpoints <= 0 and xp_mode.kill then
 					harmer.experience = harmer.experience + calc_xp( unit_to_harm.level )
 				else
-					unit_to_harm.experience = unit_to_harm.experience + harmer.level
-					harmer.experience = harmer.experience + wesnoth.game_config.combat_experience * unit_to_harm.level
+					if xp_mode.defend then
+						unit_to_harm.experience = unit_to_harm.experience + harmer.level
```
stevecotton, line 201: @gfgtdf Why is wesnoth.game_config.combat_experience applied to the attacker, but not the defender? I'm asking you because bc89512a5c000e0546676a3d0ae0f1be20b31731 seems to the be origin of it.

stevecotton:
The `attack` one wasn't clear, but this is getting too wordy:

* attack - the harming unit receives experience based on level of harmed unit, the same amount as a regular combat which does not kill the harmed unit. If `[harm_unit]` does kill the unit, then `experience=attack` means to award only the same amount as a non-kill.

ProditorMagnus:
> * If `[harm_unit]` does kill the unit, then `experience=attack` means to award only the same amount as a non-kill.

That is wrong. When experience=attack and unit dies, noone gets anything.

stevecotton:
> That is wrong. When experience=attack and unit dies, noone gets anything.

With the current code in this PR, the test below says otherwise. There are no unit tests yet for `[harm_unit]`, but I'll tidy up a bit and open a PR a test for the existing API, which this snippet can fit into. 
```
        {CHECK_HP_XP "alice" 10 1}
        {CHECK_HP_XP "bob" 1 1}

        [harm_unit]
            [filter]
                id=bob
            [/filter]
            [filter_second]
                id=alice
            [/filter_second]
            amount=5
            kill=yes
            experience=fight
        [/harm_unit]
        {CHECK_HP_XP "alice" 10 2}
        {ASSERT (
            [not]
                [have_unit]
                    id=bob
                [/have_unit]
            [/not]
        )}
```

That `CHECK_HP_XP` is currently defined in the same file, wondering about making it a general test macro (unless I find we already have one).

data/lua/wml/harm_unit.lua
```diff
@@ -168,11 +194,15 @@ function wml_actions.harm_unit(cfg)
 			if experience ~= false and harmer and harmer.valid
 				and wesnoth.sides.is_enemy( unit_to_harm.side, harmer.side )
 			then
-				if kill ~= false and unit_to_harm.hitpoints <= 0 then
+				if kill ~= false and unit_to_harm.hitpoints <= 0 and xp_mode.kill then
```
ProditorMagnus, line 197: this part is wrong then, xp_mode.kill needs to be wrapped in new if

stevecotton: The current behavior makes sense to me, that change doesn't.

* Get 2 XP for attacking a level 2 unit, regardless of whether the attack kills, vs
* Get 2 XP for attacking a level 2 unit, but get 0 XP instead if the attack kills.

stevecotton:
Unit tests for the existing API are in #8196.

The instructions for adding more are in data/test/scenarios/README.md, but it's easier to just create a PR as a worked example.

data/lua/wml/harm_unit.lua
ProditorMagnus: If there is no experience=attack/fight, then fight xp should not be given.

stevecotton: This is with `experience=attack`. If Bob has 5 hitpoints, and Alice does 4 damage, then Alice gets XP; the question is whether Alice should get 0 XP if she does more damage.

ProditorMagnus: From player perspective it would make sense to want xp, but non-harm_unit fight would not give xp.

stevecotton: I'm confused by the edit - I'm assuming we're discussing the situation with `wesnoth.game_config.kill_experience` at its default value. If someone reduced `kill_experience` to zero, why wouldn't they also reduce the attack experience to zero at the same time, thus avoiding an situation where doing more damage would result in less XP?

Hejnewar: A reason for that can be lore, if we have mind eater that earns xp only by eating living minds then killing shouldn't give it any xp, but attaching should.

ProditorMagnus:
Neither this or #8196 includes test for `experience=yes` - explicitly specifying default

Hejnewar:
Then I guess I should rename current yes tests to nil and make proper yes tests instead.

ProditorMagnus (approved):

ProditorMagnus:
False positive validation.

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/harm_unit_survivable_experience_nil.cfg
```diff
@@ -0,0 +1,61 @@
...
+# Both of these have Alice as the harmer.
+# Use [harm_unit] with Bob as the victim, and a non-fatal amount of damage.
+# Use [harm_unit] with Charlie and Dave as victims, and a non-fatal amount of damage.
+##
+# Expected end state:
+# The victims take the expected amount of damage.
+# Alice and Bob get 1 experience point each for the first fight.
+# Alice gets 2 experience points for the second fight, Charlie and Dave get 1 each.
+#####
+{COMMON_KEEP_A_B_C_D_UNIT_TEST "harm_unit_survivable_experience_nil" (
```
stevecotton, line 17: "unset" or "default" sound better than "nil" when the default maps to "yes"

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/harm_unit_kill_split_experience_defend_kill.cfg
```diff
@@ -0,0 +1,60 @@
...
+##
+# Actions:
+# Set Alice, Charlie and Dave to a known number of hitpoints.
+# Set Bob to a known but higher number of hitpoints.
+# Use [harm_unit] with Alice as harmer, the others as victims, and damage equal to Charlie's hitpoints.
+##
+# Expected end state:
+# Alice gets 16 experience points, Bob gets 1, Charlie and Dave die.
+#####
+{COMMON_KEEP_A_B_C_D_UNIT_TEST "harm_unit_kill_split_experience_defned_kill" (
```
stevecotton, line 14: Misspelling defned

data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/harm_unit_kill_split_experience_no_attack_defend_kill.cfg
```diff
@@ -0,0 +1,60 @@
...
+#####
+# API(s) being tested: [harm_unit]
+##
+# Actions:
+# Set Alice, Charlie and Dave to a known number of hitpoints.
+# Set Bob to a known but higher number of hitpoints.
+# Use [harm_unit] with Alice as harmer, the others as victims, and damage equal to Charlie's hitpoints.
+##
+# Expected end state:
+# Alice gets 17 experience points, Bob gets 1, Charlie and Dave die.
```
stevecotton, line 12: A comment here about "no" having no effect in the comma-separated list would help future readers.

data/lua/wml/harm_unit.lua
```diff
@@ -169,10 +195,15 @@ function wml_actions.harm_unit(cfg)
 				and wesnoth.sides.is_enemy( unit_to_harm.side, harmer.side )
 			then
 				if kill ~= false and unit_to_harm.hitpoints <= 0 then
-					harmer.experience = harmer.experience + calc_xp( unit_to_harm.level )
-				else
```
stevecotton, line 173: It's confusing to have both `kill` and `xp_mode.kill`, this `else` block is for "defending unit survives", not something based on the `xp_mode`.

stevecotton (changes requested):
The unit tests have caught something, the logic is broken.

data/lua/wml/harm_unit.lua
Hejnewar: Honestly I dont know what to change it to and for me kill ~= false pretty much means not dead unit but I understand why you can find that confusing to have kill two times.

Hejnewar:
I think I can confirm that at unset the unit didnt gain any xp for attacking or defending at least, whats interesting is that the unit seems to be getting xp for killing.

data/lua/wml/harm_unit.lua
```diff
@@ -169,10 +195,15 @@ function wml_actions.harm_unit(cfg)
 				and wesnoth.sides.is_enemy( unit_to_harm.side, harmer.side )
 			then
 				if kill ~= false and unit_to_harm.hitpoints <= 0 then
```
ProditorMagnus, line 197: fight xp is also wrapped in `unit_to_harm.hitpoints <= 0`

ProditorMagnus (changes requested):

data/lua/wml/harm_unit.lua
```diff
@@ -160,6 +160,32 @@ function wml_actions.harm_unit(cfg)
...
+					xp_mode.attack = true
+					xp_mode.defend = true
+				elseif opt == 'attack' then
+					xp_mode.attack = true
+				elseif opt == 'defend' then
+					xp_mode.defend = true
+				elseif opt == 'kill' then
+					xp_mode.kill = true
+				elseif opt == 'no' or opt == 'false' then
+					-- preserve previous value
```
ProditorMagnus, line 180: @CelticMinstrel what is the most proper way to get rid of luacheck of this branch? I guess it wants `elseif not (opt == 'no' or opt == 'false')`.

CelticMinstrel: I guess you'd either apply [DeMorgan's Law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws), or use an [inline comment](https://luacheck.readthedocs.io/en/stable/inline.html).

ProditorMagnus: I saw the inline comment page, but did not understand from it which comment would work.
`elseif opt == 'no' or opt == 'false' then -- luacheck: ignore` looks like way to ignore one line.

data/lua/wml/harm_unit.lua
stevecotton: There's *three* definitions of kill/death here, and from the test failures I guess you're missing significance of the third definition for this if...else block. The third definition is that the unit is only dead when `unit_to_harm.hitpoints <= 0`.

`kill ~= false` means the unit might be dead. Checking the value of it here is looking like redundant logic, because the code above this has already reduced the damage if `kill == false`, thus the logic should be able to just check `unit_to_harm.hitpoints <= 0`.

Hejnewar:
I'm closing this PR with Ravana taking it over.

data/lua/wml/harm_unit.lua
ProditorMagnus: I fixed it.

PR #8180, opened 2023-12-31
Add WML verification workflow

### LINKED ISSUE #7082: Add wmllint to Travis builds

czyzby:
Closes #7082.

The workflow steps are as follows:

* Wesnoth respository is checked out.
* Python 3.10 is set up.
* Dependencies for spellchecking are installed.
* `wmllint` is ran in dry-run mode against the `data/` folder.
  * Since the script does not define error codes, the messages are filtered instead to check if there are any issues.
  * Logs starting with `#` are filtered out, since they seem to be purely informational and do not indicate any errors.
  * If the filtered output logs are not empty, the pipeline step is considered to be failed.
  * The same output handling is applied to other tools.
* `wmlindent` is ran in dry-run mode against the `data/` folder.
* `wmlscope` is ran against the `data/` folder.
  * Currently no flags are selected. Let me know if that should be changed.
  * Trying to verify the WML files with `-u` (unresolved references) yields the following Python error:

```python
  File "/home/runner/work/wesnoth/wesnoth/data/tools/wmlscope", line 564, in <module>
    xref.incorrectlysized()
  File "/home/runner/work/wesnoth/wesnoth/data/tools/wmlscope", line 221, in incorrectlysized
    for (namespace, filename) in xref.filelist.generator():
AttributeError: 'list' object has no attribute 'generator'
```

The pipeline is executed on every push and pull request to the `master` branch, as well as by manual dispatch. Note that the workflow currently fails with 3000+ reported issues.

### DIFF
--- a/.github/workflows/wml-linter.yml
+++ b/.github/workflows/wml-linter.yml
@@ -1,0 +1,48 @@
+name: WML
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  lint:
+    name: Verify WML files
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: |
+          sudo apt-get update >/dev/null
+          sudo apt-get upgrade -y >/dev/null
+          sudo apt-get install -y enchant-2 hunspell hunspell-en-us
+          pip install pyenchant~=3.0
+      - name: Run wmllint
+        run: |
+          exec 5>&1
+          if [[ -n $(python data/tools/wmllint -d data/ 2>&1 | grep --line-buffered "^[^#].*" | tee >(cat - >&5)) ]]; then
+            echo "Found issues detected by wmllint."
+            exit 1
+          fi
+      - name: Run wmlindent
+        run: |
+          exec 5>&1
+          if [[ -n $(python data/tools/wmlindent -d data/ 2>&1 | grep --line-buffered "^[^#].*" | tee >(cat - >&5)) ]]; then
+            echo "Found issues detected by wmlindent."
+            exit 1
+          fi
+        if: always()
+      - name: Run wmlscope
+        run: |
+          exec 5>&1
+          if [[ -n $(python data/tools/wmlscope data/ 2>&1 | grep --line-buffered "^[^#].*" | tee >(cat - >&5)) ]]; then
+            echo "Found issues detected by wmlscope."
+            exit 1
+          fi
+        if: always()


czyzby:
[Sample output.](https://github.com/wesnoth/wesnoth/actions/runs/7372976293/job/20061782539?pr=8180)

ProditorMagnus:
This can probably be never added unless it can be configured to run only for changed files.

czyzby:
That's a bit more difficult to set up, but I can look into it next year. 😏

cooljeanius:
> This can probably be never added unless it can be configured to run only for changed files.

Yeah, there should be some sort of way to mark a certain set of warnings as a baseline of acceptable warnings, and then only fail the CI if the total number of warnings increases from the baseline...

Pentarctagon:
As mentioned in the issue, wmlindent shouldn't be here since it's already run as part of the CI checks:
https://github.com/wesnoth/wesnoth/blob/fb4ac0b5fb033581d9b544faee44a7860bb526f6/utils/CI/fix_whitespace.sh#L5

This also can't be merged until all the wmlint warnings are either addressed or removed. Running a CI check that outputs a bunch of warnings which are simply ignored isn't useful.

Likely this should also be run as part of the existing "checks" job rather than adding a new one.

czyzby:
> Yeah, there should be some sort of way to mark a certain set of warnings as a baseline of acceptable warnings, and then only fail the CI if the total number of warnings increases from the baseline...

That would be difficult to achieve, since you'd have to run it twice before and after the changes. The best I can think of right now would be to limit the output to the files modified by the PR or latest commit.

> As mentioned in the issue, wmlindent shouldn't be here since it's already run as part of the CI checks:

I think there are some errors that the tool cannot fix automatically. Even right now it reports hundreds of issues with the existing files. I can remove it though if that's preferable.

> Likely this should also be run as part of the existing "checks" job rather than adding a new one.

OK, got it, I'll look into it along with limiting the output to modified files.

Pentarctagon:
> I think there are some errors that the tool cannot fix automatically. Even right now it reports hundreds of issues with the existing files. I can remove it though if that's preferable.

It's that some directories (data/gui/ and data/schema/) don't follow regular indentation rules, and the feeling so far has been it's better to leave them as-is rather than fixing the indentation. Personally I wouldn't be opposed to making them be consistent with everything else, but that's a different question.

Pentarctagon:
Is this still being pursued?

cooljeanius:
> Is this still being pursued?

@czyzby mentioned returning to it later in the year, so I think it'd be ok to wait a bit before taking any further action on it

czyzby:
I can still look into this, but it isn't trivial to set up to detect errors only in specific lines modified by the commit. It would require something like diffing errors against the target branch/last commit, or perhaps limiting output to modified files (which isn't ideal). I'll set up a reminder for this weekend, if I get stuck or it ends up impractical due to too many errors reported I'll let you know.

Pentarctagon:
Closing as per my last comment - I still haven't gotten around to it but I think that's the correct path forward.

Issue #8179, opened 2023-12-31
Better way to determine the number of custom AMLAs a unit has received


Pentarctagon:
#8152 uses max_experience to determine when to increment the unit's level so that it doesn't stay as level 1 forever. However that's a fragile approach, so it would make sense for there to be a more reliable way to get that information.

CelticMinstrel:
A simple approach is to simply expose the modifications array to Lua and WFL. Then you could use `#unit.modifications.advancement` to get the number of AMLAs that have been applied. As a bonus you can also detect the number of traits or objects with this method, and even inspect the modifications to determine their effects or filter data.

gfgtdf:
you can already use __cfg for such a approach (and i explained in that other pr why i think it is at least in the  current form not a very good approach). unrelated to that it might be nice to have `unit.modifications` though to allow writing unit modifications.

PR #8189, opened 2024-01-01
Allow effects applied in [modifications] to be reported by [advancement], [object] or [trait]

newfrenchy83:
Can resolve https://github.com/wesnoth/wesnoth/issues/8179

This approach consists of using three different variables to perform the counting before making them accessible in the formulas, but not in the wml variables because they must remain read-only to avoid any modifications making their values incorrect.

### DIFF
--- a/data/campaigns/Under_the_Burning_Suns/units/quenoth/Youth.cfg
+++ b/data/campaigns/Under_the_Burning_Suns/units/quenoth/Youth.cfg
@@ -47,15 +47,7 @@ _"Nym is Kaleh’s childhood friend, a young, rebellious lady with quick wits an
     [effect]
         [filter]
             level=1
-#ifdef EASY
-            formula="max_experience > 24"
-#endif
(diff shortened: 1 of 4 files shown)


CelticMinstrel:
I'm not sure I like this implementation. Using a separate counter seems like it could run the risk of getting out of sync somehow.

newfrenchy83:
> I'm not sure I like this implementation. Using a separate counter seems like it could run the risk of getting out of sync somehow.

have you a suggestion?

ProditorMagnus:
Instead of
> formula="max_experience > 33"
there can be lua_function which takes what it needs from unit cfg.

CelticMinstrel:
We have to discuss whether it's reasonable to expose just the _number_ of modifications rather than providing information about those modifications. If we want to provide information, we need to decide which information to provide. It could be just the IDs, or it could be the entire config. Maybe there's a middle option too.

PR #8191, opened 2024-01-01
use lenght of modifications.advancement for leveling of Quenoth Youth

newfrenchy83:
Can resolve https://github.com/wesnoth/wesnoth/issues/8179

It is should be more consistent what my precedet commit


ProditorMagnus:
Looks correct.

PR #8195, opened 2024-01-02
1.18 - various minor monster updates

doofus-01:
These are some very small, random updates to a few monster sprites, just a tiny project I could get done in time for 1.18.0.

EDIT: Don't commit this unless I've disappeared before 1.18.0.  This is a holding pen for small images fixes I wish to get in before next stable release.


doofus-01:
I use GIMP (pencil, hard-edge eraser, no anti-alias lasso select).  One frame is one layer, though often the semitransparent shadow can be on a separate layer to avoid needing a 60% pencil.

PR #8196, opened 2024-01-02
Add tests for [harm_unit]

stevecotton:
This is testing experience=yes and experience=no, with the expectation that some new values for that attribute will be added afterwards.

Adds a new generic macro ASSERT_UNIT_HP_XP. Looking thorugh the ability tests, ASSERT_UNIT_HP (without the XP) could be a heavily-used macro, but it wouldn't be used by this particular commit, so I'm not adding it in this commit.

### DIFF
--- a/data/test/macros/wml_unit_test_macros.cfg
+++ b/data/test/macros/wml_unit_test_macros.cfg
@@ -136,3 +136,20 @@ bob#endarg
     {CLEAR_VARIABLE {ID1}}
     {CLEAR_VARIABLE {ID2}}
 #enddef
+
+# Takes a unit id, and checks that its hitpoints and experience are the expected values.
+# Clobbers the variable "temp".
+#define ASSERT_UNIT_HP_XP ID EXPECTED_HP EXPECTED_XP
+    [store_unit]
+        [filter]
+            id={ID}
+        [/filter]
+        variable=temp
+    [/store_unit]
+
+    {ASSERT ({VARIABLE_CONDITIONAL temp.length numerical_equals 1})}
+    {ASSERT ({VARIABLE_CONDITIONAL temp.hitpoints numerical_equals {EXPECTED_HP}})}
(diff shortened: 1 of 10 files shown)


data/test/scenarios/wml_tests/ScenarioWML/EventWML/ActionWML/DirectActionsWML/harm_unit_experience_yes.cfg
```diff
@@ -0,0 +1,135 @@
...
+# Use [harm_unit] with Charlie and Dave as victims, but a non-fatal amount of damage.
+# Heal Dave.
+# Use [harm_unit] with enough damage to kill Charlie but not Dave.
+##
+# Expected end state:
+# The victims take damage in every step.
+# Alice and Bob get 1 experience point each for the first fight.
+# Alice and Bob get 1 experience point each for the second fight.
+# Alice gets 8 experience points for the third fight, Bob dies.
+# Alice, Charlie and Dave get 1 experience point each for the fourth fight.
```
stevecotton, line 21: Doesn't match the test.

Just realised that I've documented what I expected it to do from the docs, and coded what I expected it to do from the code. On the next line, I've documented what it really does - Alice gets more XP depending on the number of victims.

Will fix the docs in the next push, but will be AFK for a bit, so waiting for more comments.

Pentarctagon:
I think these would be easier to read if they were split into separate tests/files. Or at least numbered the WML doing the testing to match the expected results documented.

stevecotton:
I could definitely number their parts. Will check about splitting them up, but I'm worried that it'll become 30 separate tests when #8178 is included.

stevecotton:
The most recent push bundles together "Alice does X to Bob" in the same test as "Alice does X to Charlie,Dave", but each different X gets a separate test.

stevecotton:
@Pentarctagon are you happy with the new file split?

For the interaction with #8178, my expectation is that this PR merges, and that PR gets a corresponding set of 4 tests for each value of `experience=`.

Pentarctagon (approved):

